PR #899, opened 2016-11-14
Convention payment tool

### LINKED ISSUE #860: Convention Payment via Journal Voucher
This issue is about to add tool for importing transaction rows for a convention payment. 

1. Click on the `convention payment` button in the toolbar for launching a modal : 

convention invoices list

The modal contains : 

- An input for selecting the convention 
- When the convention is selected, then : 
  - display in a grid the list of all unpaid invoices of the convention

2. Select invoices to pay and click on the `Select invoices for payment` button

When proceed step 2, you'll see another modal : 

invoices import confirm

In this modal, the user must set : 

- The amount to pay 
- The cashbox where the money must go

3. Click on the `import` button, which must : 

- Set the transaction type 
- Set the amount to debit for the selected cashbox 
- Credit convention for each selected invoices 
- Set correct entity and reference

voucher grid

In this step, we must be sure all credit must not exceed the amount to pay.

mbayopanda:
This PR is about : 

- [x] A tool for importing transaction rows in the complex voucher for a convention payment
- [ ] E2E test the importing tool for convention payment

Closes #860.


mbayopanda:
@jniles could you review this PR please.

client/src/js/services/DebtorGroupService.js
```diff
@@ -121,5 +123,17 @@ function DebtorGroupService($http, Modal, util, SessionService) {
     });
   }
 
+  /**
+   * @method invoices
```
jniles, line 127: :+1: nice documentation.

client/src/partials/vouchers/complex.js
```diff
@@ -19,7 +20,7 @@ ComplexJournalVoucherController.$inject = [
  * @todo - Implement Patient Invoices data and Cash Payment data for modal
  * @todo - Implement a mean to categorise transactions for cashflow reports
  */
-function ComplexJournalVoucherController(Vouchers, $translate, Accounts, Currencies, Session, FindEntity, FindReference, Notify, Cashbox) {
+function ComplexJournalVoucherController(Vouchers, $translate, Accounts, Currencies, Session, FindEntity, FindReference, Notify, Cashbox, Toolkit) {
```
jniles: Nice name!

client/src/partials/vouchers/toolkit.service.js
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * @function open
+   * @description This function initialise a modal for the selected tool
+   * @param {Object} option A detailed option object that should provide
+   * label, controller string, templateUrl string
+   */
+  function open(option) {
+    if (!option.controller || !option.templateUrl) { return; }
+
+    var controllerString = option.controller.concat(' as ToolCtrl');
```
jniles: Instead of doing this...

You can just use the property `controllerAs` in the options.  [Here is the documentation](https://angular-ui.github.io/bootstrap/#/-uibmodal-s-open-function).

``` js
var instance = Modal.open({
  controller : 'SomeController',
  controllerAs: '$ctrl'
});
```

client/src/partials/vouchers/toolkit/convention_payment.js
```diff
@@ -0,0 +1,176 @@
...
+  vm.import = submit;
+
+  // accounts from store
+  AccountStore.accounts().then(function (data) {
+    vm.accounts = data;
+  });
+
+  // debtors from store
+  // FIXME: need auto refresh
+   Debtors.store.then(function (data) {
```
jniles: This would be much better as a function.

client/src/partials/vouchers/toolkit/convention_payment.js
```diff
@@ -0,0 +1,176 @@
...
+     vm.debtors = data;
+   });
+
+  //  optimization with `Store` will be well
+   Invoices.read()
+   .then(function (data) {
+     vm.invoices = data;
+   });
```
jniles: All of these should have `Notify.handleError` to make sure the errors in network connections are reported.

client/src/partials/vouchers/toolkit/convention_payment.js
```diff
@@ -0,0 +1,176 @@
...
+      label: entity.text,
+      type: 'D',
+      uuid: entity.uuid
+    };
+  }
+
+  // get invoice reference
+  function getInvoiceReference(uuid) {
+    return vm.invoices.filter(function (item) {
+      item.document_type = 'VOUCHERS.COMPLEX.PATIENT_INVOICE';
```
jniles: Why is a filter function changing properties on the objects in the array?  That is confusing and not very pure.

client/src/partials/vouchers/toolkit/convention_payment.modal.html
```diff
@@ -0,0 +1,98 @@
...
+  <div class="modal-footer">
+    <button
+      type="button"
+      class="btn btn-default"
+      ng-click="ToolCtrl.close()"
+      data-method="close">
+      {{ "FORM.BUTTONS.CANCEL" | translate }}
+    </button>
+
+    <bh-loading-button loading-state="ToolCtrl.$loading"
```
jniles, line 93: :+1:

client/src/partials/vouchers/toolkit/convention_payment.modal.html
```diff
@@ -0,0 +1,98 @@
...
+      <div class="help-block" ng-messages="ToolForm.convention.$error" ng-show="ToolForm.$submitted">
+        <div ng-messages-include="partials/templates/messages.tmpl.html"></div>
+      </div>
+    </div>
+
+    <!-- select invoices  -->
+    <div ng-show="ToolCtrl.convention">
+      <div
+        id="invoiceGrid"
+        style="height: 300px; margin-bottom: 20px;"
```
jniles: Do you need to put a fixed height?  If so, I would base it off screen height.

For example, `height: 75vh` will put the modal at 75% of the current view height.

server/controllers/finance/debtors/groups/index.js
```diff
@@ -230,24 +230,68 @@ function list(req, res, next) {
...
 function invoices(req, res, next) {
   var options = req.query;
 
-  if (options.balanced) { options.balanced = Boolean(options.balanced); }
+  loadInvoices(req.params.uuid)
+  .then(function (rows) {
+    let data = rows;
 
-  res.status(500).send('Unimplemented..');
+    if (options.balanced) {
```
jniles: Could you make this in `options.balanced` filter using SQL?  Either a `WHERE` or a `HAVING` clause?

It would be much more efficient.

server/controllers/finance/debtors/groups/index.js
```diff
@@ -230,24 +230,68 @@ function list(req, res, next) {
...
+      data = rows.filter(item => {
+        return item.balance === 0;
+      });
+    }
 
-  /*
-  invoices(req.params.uuid, options)
-  .then(function (rows) {
-    res.status(200).json(rows);
+    if (options.unpaid) {
```
jniles: This should also do its filtering in SQL

server/controllers/finance/debtors/groups/index.js
```diff
@@ -230,24 +230,68 @@ function list(req, res, next) {
...
+  // get invoices balance for each debtor
+  let sqlInvoices =
+    `SELECT BUID(i.uuid) as uuid, i.date, CONCAT(project.abbr, i.reference) as reference,
+      debit, credit, (debit - credit) as balance, BUID(entity_uuid) as entity_uuid
+    FROM (
+      SELECT record_uuid as uuid, combined_ledger.date, SUM(debit) as debit, SUM(credit) as credit,
+        entity_uuid, invoice.reference, invoice.project_id
+      FROM combined_ledger
+      JOIN invoice ON combined_ledger.record_uuid = invoice.uuid OR combined_ledger.reference_uuid = invoice.uuid
+      WHERE entity_uuid IN (?) AND invoice.uuid NOT IN (SELECT voucher.reference_uuid FROM voucher WHERE voucher.type_id = 10)
```
jniles, line 277: Instead of saying `voucher.type_id = 10`, could you give the "10" variable some context?  Like:

``` js
const CASH_REVERSAL_TYPE = 10;

let sqlInvoices = `
  SELECT .....
   WHERE  .... AND invoice.uuid NOT IN (SELECT voucher.reference_uuid FROM voucher WHERE voucher.type_id = ${ CASH_REVERSAL_TYPE  })
  .... 
`;
```

I'm not sure, as a developer, what the magic number 10 means.

server/controllers/finance/debtors/index.js
```diff
@@ -260,14 +260,14 @@ function financialPatient(debtorUuid) {
       FROM posting_journal
       LEFT JOIN invoice ON invoice.uuid = posting_journal.record_uuid
       LEFT JOIN project ON invoice.project_id = project.id
-      WHERE posting_journal.entity_uuid = ? AND posting_journal.entity_type = 'D'
+      WHERE posting_journal.entity_uuid = ?
```
jniles, line 263: :+1:

jniles (changes requested):
@mbayopanda, I've completed an initial review.

The module works perfectly fine, from what I can tell.  Here are some important changes that need to be made before this can land in master:
1. Make sure every HTTP request has proper error handling with `Notify`.
2. Make sure all filtering is done in SQL - it is much more efficient!  A large request may block the server and lead to high latency.

Thanks!

client/src/partials/vouchers/toolkit.service.js
mbayopanda: 👍

server/controllers/finance/debtors/groups/index.js
mbayopanda: 👍

mbayopanda:
@jniles could you review this PR again please. I perform all your suggestions about this PR.

client/src/js/services/DebtorService.js
```diff
@@ -43,6 +43,9 @@ function DebtorService($http, util) {
   /** returns a list of invoices owed to a given debtor */
   service.invoices = invoices;
 
+  // load debtors
+  service.store = store;
```
jniles, line 47: That's better!

server/controllers/finance/debtors/groups/index.js
```diff
@@ -230,24 +230,66 @@ function list(req, res, next) {
...
+ * @description This method returns invoices of a debtor group
+ * @param {string} params A object which contains :
+ * {
+ *  debtor_uuid : ... // required
+ *  balanced: ... // for balanced invoices
+ * }
+ */
+function loadInvoices(params) {
+  // cancelled transaction type
+  const CANCELED_TRANSACTION_TYPE = 10;
```
jniles, line 262: Nice!

server/controllers/finance/debtors/groups/index.js
```diff
@@ -230,24 +230,66 @@ function list(req, res, next) {
...
+      JOIN invoice ON combined_ledger.record_uuid = invoice.uuid OR combined_ledger.reference_uuid = invoice.uuid
+      WHERE entity_uuid IN (?) AND invoice.uuid NOT IN (SELECT voucher.reference_uuid FROM voucher WHERE voucher.type_id = 10)
+      GROUP BY uuid
+    ) AS i
+    JOIN project ON i.project_id = project.id
+    WHERE i.uuid NOT IN (SELECT voucher.reference_uuid FROM voucher WHERE voucher.type_id = ${CANCELED_TRANSACTION_TYPE})
+    `;
+
+  // balanced or not
+  sqlInvoices += params.balanced ? ' HAVING balance = 0 ' : ' HAVING balance <> 0 ';
```
jniles, line 285: Nice!

jniles (approved):
@mbayopanda These changes LGTM.

jniles:
Can you do the end to end tests before this gets merged?

test/end-to-end/vouchers/complex.spec.js
```diff
@@ -124,4 +125,38 @@ describe('Complex Vouchers', function () {
...
+  it('Convention import invoices and payment via the tool', () => {
+
+    let detail = {
+      tool: 'Convention - Paiement factures',
+      cashbox: '$',
+      convention: 'Second Test',
+      invoices: [0, 1]
+    };
+
+    // click on the convention tool
```
jniles, line 137: This are really good comments!

test/end-to-end/vouchers/complex.spec.js
```diff
@@ -124,4 +125,38 @@ describe('Complex Vouchers', function () {
     components.notification.hasDanger();
   });
 
+  it('Convention import invoices and payment via the tool', () => {
+
+    let detail = {
+      tool: 'Convention - Paiement factures',
+      cashbox: '$',
+      convention: 'Second Test',
+      invoices: [0, 1]
```
jniles, line 134: If you never use `detail.invoices[1]`, you should probably just remove it.

client/src/partials/vouchers/toolkit/convention_payment.js
```diff
@@ -0,0 +1,190 @@
...
+    vm.rows = [];
+
+    var cashboxAccount = vm.accounts.get(result.cashbox.account_id);
+    var conventionAccount = vm.accounts.get(result.convention.account_id);
+    var entityAccount;
+    var invoiceReference;
+    var index;
+
+    // cashbox row
+    vm.rows.push(generateRow());
```
jniles, line 92: This is slightly inefficient.  It's also harder to read.  In the future, it would be best to do this:

``` js
var row = generateRow():

row.account = cashboxAccount;
row.debit = vm.totalSelected;
// etc...

vm.rows.push(row);
```

client/src/partials/vouchers/toolkit/convention_payment.js
```diff
@@ -0,0 +1,190 @@
...
+
+  function onRegisterApi(gridApi) {
+    vm.gridApi = gridApi;
+    vm.gridApi.selection.on.rowSelectionChanged(null, rowSelectionCallback);
+    vm.gridApi.selection.on.rowSelectionChangedBatch(null, rowSelectionCallback);
+  }
+
+  function rowSelectionCallback(row) {
+    vm.selectedRows = vm.gridApi.selection.getSelectedRows();
+    vm.totalSelected = vm.selectedRows.reduce(function (current, previous) {
```
jniles, line 173: Here's one future optimization.  `rowSelectionCallback()` will be called on every row click.  That means the `reduce()` function will be created on every click and then destroyed.

Since you use the `reduce()` function in several places, you could do this:

``` js
function getVoucherRowBalances(current, previous) {
  return current + previous.balance;
}

function rowSelectionCallback(row) {
  vm.selectedRows = vm.gridApi.selection.getSelectedRows();
  vm.totalSelected = vm.selectedRows.reduce(getVoucherRowBalances, 0);
}
```

This would prevent Chrome from having to recreate the same function over and over.

client/src/partials/vouchers/toolkit/convention_payment.js
```diff
@@ -0,0 +1,190 @@
...
+      debit         : 0,
+      credit        : 0,
+      document_uuid : undefined,
+      entity_uuid   : undefined
+    };
+  }
+
+  /* ================ Invoice grid parameters ===================== */
+  vm.gridOptions.appScopeProvider = vm;
+  vm.gridOptions.enableFiltering  = true;
```
jniles, line 153: You should use `vm.gridOptions.fastWatch = true` here.  It would be more efficient.

client/src/partials/vouchers/toolkit/convention_payment.js
```diff
@@ -0,0 +1,190 @@
...
+    // cashbox row
+    vm.rows.push(generateRow());
+    vm.rows[0].account = cashboxAccount;
+    vm.rows[0].debit = vm.totalSelected;
+    vm.rows[0].credit = 0;
+
+    // cashbox and invoices rows
+    for (var i = 0; i < result.invoices.length; i++) {
+      entityAccount = formatEntity(result.invoices[i].entity_uuid);
+      invoiceReference = getInvoiceReference(result.invoices[i].uuid);
```
jniles, line 100: This is `O(n^2)`.  You can easily make this `O(n)` like this:

``` js

var invoiceStore = new Store({ identifier : 'uuid', data : result.invoices });

for (var i = 0; i  < result.invoices.length; i++) {

  // this is now O(1)
  invoiceReference = invoiceStore.get(result.invoices[i].uuid).reference;

  // etc..
}
```

client/src/partials/vouchers/toolkit/convention_payment.js
```diff
@@ -0,0 +1,190 @@
...
+    var index;
+
+    // cashbox row
+    vm.rows.push(generateRow());
+    vm.rows[0].account = cashboxAccount;
+    vm.rows[0].debit = vm.totalSelected;
+    vm.rows[0].credit = 0;
+
+    // cashbox and invoices rows
+    for (var i = 0; i < result.invoices.length; i++) {
```
jniles, line 98: Did you know that `Array.prototype.forEach` has three parameters?  [Documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).

So, your code can be much more semantic:

``` js
result.invoices.forEach(function (invoice, index) {
  entityAccount = formatEntity(invoice);
  invoiceReference = getInvoiceReference(invoice);
});
```

You actually don't need `index` or `i`!  `Array.prototype.push()` returns the next index.  Like this:

``` js
var index = vm.rows.push(generateRow());
// index is now the length of rows.
```

jniles (approved):
:+1: LGTM.  I've made comments for your information, but none of them are strictly necessary for this PR.

Well done!

PR #922, opened 2016-11-23
Fix translation

lomamech:
- Complete translate file
- Update components bhPDFPrint, adding the using of current language key

@jniles  @sfount


sfount (approved):
LGTM :+1:

Issue #927, opened 2016-11-23
(refactor) Service configuration


sfount:
**Required Updates**
* `ui-router` latest standards `$state` management 
* Headercrumb standards
- [x] ID displaying for index in table - removed
- [x] [Bug] Overview crashses with 404 - routes not imlemented
* Optionally add cost and profit center instead of leaving blank
* Separate .route.js

jniles:
Closing this as superseded by https://github.com/IMA-WorldHealth/bhima-2.X/issues/1369.

PR #934, opened 2016-11-25
(Chore) Remove unsupported features in 2.X

sfount:
This pull request removes UI elements that have been added to 2.X but do not have features or designs to support them.

It also seperates `test/data.sql` into `bhima.sql` and and an enterprise data file. 
`bhima.sql` contains all system definitions that should be the base of all installations - modules, transaction types, languages etc. 
These can be added to by specific installations however the base should always be the same. This file is now run before the enterprise SQL (for example test/data.sql or lukonga.sql or tshikaji.sql). 

* Remove cost and profit center assignments in services - there is no way to create these in 2.x
* Remove IDs in services 
* Assign enterprises automatically for services - remove UI element 
* Remove distribute-able field from invoices - there is no stock to represent at this point

Additionally this contributes to but does not complete https://github.com/IMA-WorldHealth/bhima-2.X/issues/927


jniles:
Closing this in favor of https://github.com/Vanga-Hospital/bhima-2.X/pull/92.

Issue #865, opened 2016-11-09
(Feature) BHIMA Landing Proposal


sfount:
We currently have a placeholder landing page that demonstrates the ability of 2.x dashboards/ available information but does not implement anything useful to the enterprise. 

This mockup is a proposal for the initial landing implementation; this page is predominantly finance focused and shows:
* Current enterprise 
* Current project and project code 
* Enterprise currency 
* Exchange rate to other currencies 
* Currency fiscal year progression/ status
 
* Money that is owed from debtors broken down by groups
* Money that may be owed to creditors 

* Patient check ins
![home_mockup](https://cloud.githubusercontent.com/assets/2844572/20134708/b8c8ae98-a66c-11e6-975f-4f066e6c1788.png)

sfount:
My proposal is that I take this initially and then individual components (graphs, other queries) can be picked up by other developers.

jniles:
This looks like a really nice design.

If it were possible to display patient information to only medical users (doctors, receptionists, etc), that would be ideal.  Similarly, if it were possible to only show financial information to financial users (accountants, cashiers, etc) that would also be nice.  Potentially that requires more infrastructure than we want to implement at the moment.

jniles:
This was mostly closed in #938.  Another issue should be opened if subsequent changes are to be made.

PR #938, opened 2016-11-25
(Feature) Landing page

sfount:
This pull request implements the design https://github.com/IMA-WorldHealth/bhima-2.X/issues/865. 

It updates the landing page
* Removes activity ping
* Removes graph with random data

* Implements statistic view of enterprise information 
* Implements debtor overview graph
* Implements dashboard caching proof of concept 

landing

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -810,3 +810,27 @@ growl-notification.fading.ng-leave.ng-leave-active {
   margin-right: auto;
   margin-left: auto
 }
+
+/* @TODO Move all panel/ panel-body styles into segment */
+.segment {
+  margin-bottom : 5px;
+}
+
+.segment > .panel-body {
+  /* padding-top : 5px; */
+}
+.segment-title {
+  position : absolute;
+  top : 3px;
+  left : 10px;
+  text-align : left;
+  color : #a8a8a8;
+}
+
+.progress-bar {
+  transition : none;
+}
+
+.progress-bar.ima-blue {
+  background-color : #085484;
+}
--- a/client/src/js/app.js
+++ b/client/src/js/app.js
@@ -16,9 +16,19 @@ function bhimaConfig($stateProvider, $urlMatcherFactoryProvider) {
 
   $stateProvider
   .state('index', {
+    abstract : true,
     url : '/',
     controller : 'HomeController as HomeCtrl',
     templateUrl : 'partials/home/home.html'
+  })
+  .state('index.details', {
+    url : '',
+    views : {
+      'debtors@index' : {
+        templateUrl : 'partials/home/units/debtors.html',
+        controller : 'DashboardDebtorController as DebtorCtrl'
+      }
+    }
   })
   .state('login', {
     url : '/login',
--- a/client/src/partials/home/home.html
+++ b/client/src/partials/home/home.html
@@ -1,8 +1,8 @@
 <div class="flex-header">
   <div class="bhima-title">
     <ol class="headercrumb">
-      <li class="static">{{ ::HomeCtrl.enterprise.name }}</li>
-      <li class="title">{{ ::HomeCtrl.project.name }}</li>
+      <li class="static" style="text-transform : uppercase">{{ "HOME.BHIMA" | translate }}</li>
+      <li class="title">{{ ::HomeCtrl.enterprise.name }}</li>
     </ol>
   </div>
 </div>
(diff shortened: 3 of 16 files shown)


jniles:
Just FYI, this branch has conflicts.

client/src/css/structure.css
```diff
@@ -810,3 +810,27 @@ growl-notification.fading.ng-leave.ng-leave-active {
...
+}
+.segment-title {
+  position : absolute;
+  top : 3px;
+  left : 10px;
+  text-align : left;
+  color : #a8a8a8;
+}
+
+.progress-bar {
```
jniles, line 830: :+1:

client/src/js/app.js
```diff
@@ -16,10 +16,20 @@ function bhimaConfig($stateProvider, $urlMatcherFactoryProvider) {
...
+    abstract : true,
     url : '/',
     controller : 'HomeController as HomeCtrl',
     templateUrl : 'partials/home/home.html'
   })
+  .state('index.details', {
+    url : '',
+    views : {
+      'debtors@index' : {
+        templateUrl : 'partials/home/units/debtors.html',
```
jniles, line 28: I wonder if `units` is the best name for this...  It seems a bit too generic.

client/src/partials/home/home.html
```diff
@@ -1,8 +1,8 @@
 <div class="flex-header">
   <div class="bhima-title">
     <ol class="headercrumb">
-      <li class="static">{{ ::HomeCtrl.enterprise.name }}</li>
-      <li class="title">{{ ::HomeCtrl.project.name }}</li>
+      <li class="static" style="text-transform : uppercase">{{ "HOME.BHIMA" | translate }}</li>
+      <li class="title">{{ HomeCtrl.enterprise.name }}</li>
```
jniles, line 5: :+1:

client/src/partials/home/home.html
```diff
@@ -11,124 +11,94 @@
...
+        <div class="panel panel-default segment">
+          <div class="panel-body">
+            <div class="segment-title">{{ "FISCAL.FISCAL_YEAR" | translate }}</div>
 
-              <dt>{{ ::"SYSTEM.MACHINE_UPTIME" | translate }}</dt>
-              <dd>{{ HomeCtrl.system.machineUptime | amDurationFormat }}</dd>
+            <div style="margin-top : 12px">
+            <div class="progress" style="margin-bottom:0px;">
+              <div class="ima-blue progress-bar" role="progressbar" style="width:{{HomeCtrl.year.percentage}}%;">
+                {{ HomeCtrl.today | date }}
```
jniles, line 78: Nice!  This was the original idea.

client/src/partials/home/home.js
```diff
@@ -45,135 +45,18 @@ function HomeController(Currencies, Rates, Session, System, $interval, $translat
     .then(function () {
       vm.currencies.forEach(function (currency) {
         currency.rate = Rates.getCurrentRate(currency.id);
+        currency.formattedDate = new Moment(currency.date).format('LL');
       });
+
+      // @TODO Method for selecting primary exchange
+      vm.primaryExchange = vm.currencies[0];
```
jniles, line 52: This could always be the first element if the list is length 1.  Otherwise, it might be a setting on the settings page.

client/src/partials/home/units/debtors.js
```diff
@@ -0,0 +1,35 @@
+angular.module('bhima.controllers')
+.controller('DashboardDebtorController', DashboardDebtorController);
+
+DashboardDebtorController.$inject = ['DashboardService'];
+
+function DashboardDebtorController(Dashboard) {
+  var vm = this;
+
+  // @TODO move to service to deal with assigning application colours
```
jniles, line 9: We should totally create a `ColorService`.

client/src/partials/home/units/debtors.js
```diff
@@ -0,0 +1,35 @@
...
+      }
+    }
+  };
+
+  Dashboard.debtors()
+    .then(function (result) {
+      var debtorGroups = result.data.debtorGroups;
+      vm.reportCached = result.timeCached;
+
+      vm.owedGraph.data = debtorGroups.map(function (group) {
```
jniles, line 27: I feel obliged to comment that you could have done this in one loop with a reduce.  But this is fine :+1:

client/src/partials/settings/settings.html
```diff
@@ -50,6 +50,35 @@
           </span>
         </a>
       </div>
+      <!-- system information panel -->
+      <div class="col-md-6 col-md-offset-1">
```
jniles, line 54: The settings page is a much better place for this.  Glad to see it moved!

client/src/partials/home/home.js
```diff
@@ -13,21 +13,21 @@ HomeController.$inject = [
...
  * the infrastructure to power the dashboard completely.
  *
- * @todo - remove the fake graph generation code and replace with actual system
- * activity, such as patient registrations.
+ * @todo - implement actual system activity, such as patient registrations.
  * @todo - implement fiscal year client-side services to get relevant fiscal year
  * services and information.
  */
-function HomeController(Currencies, Rates, Session, System, $interval, $translate, $scope, Notify, Fiscal) {
+function HomeController(Currencies, Rates, Session, System, $translate, $scope, Notify, Fiscal, Moment, DashboardService) {
```
jniles: Do you still need System here?  Can it be removed?

server/controllers/dashboard/debtorGroups.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+const moment = require('moment');
+
+const report = require('../finance/reports/agedDebtors');
+
+// time until the cached value expires - currently set at 5 minutes
+const CACHE_TTL = 300000;
```
jniles, line 7: This is cool!  It could be way more generic.  Do you think that we should be using variables local to a module as the cache?  It seems rather amateur, but I cannot think of a better way at the moment...

jniles (approved):
This is an interesting idea - I think the cache could use some polish, but that will come with experience.

 Nothing is blocking in this PR.  LGTM.

mbayopanda:
@sfount , 

The landing page is very cool, i have some suggestions. 

In the Enterprise Debtors panel why not remove the panel heading (for having the same style as in other card) and display values in pies sections.

mbayopanda:
@sfount this branch has conflicts.

client/src/partials/home/home.js
sfount: Nope! No need for $translate, $scope or Moment either!

server/controllers/dashboard/debtorGroups.js
sfount: It could also be stored using redis or mysql but I think this is okay for now - the main consideration is that this does not scale well across different server machines or cores.

client/src/js/app.js
sfount: https://github.com/IMA-WorldHealth/bhima-2.X/issues/946

client/src/partials/home/units/debtors.js
sfount: My understanding is a reduce in JavaScript takes an array of values and reduces it to a single value? That is not what this map is used for in this case - do you have an example?

It could be done with a more generic loop as the the data indexed into is the same for both assignments. If there are any additional changes here I will include that.

mbayopanda:
👍

Issue #939, opened 2016-11-25
(Bug) Patient Invoices do not recalculate totals when an item is removed


sfount:
When line items are removed from an invoice on the Patient Invoice module the totals remain the same. If a new line is added the totals are recalculated correctly. 

This is very important as it is assumed the total always reflects the items in the invoice.

jniles:
This is probably due to a `fastWatch`.  Emitting a `uiGridConstants.dataChange.EDIT` will fix it.

PR #950, opened 2016-11-27
(Fix) Ensure totals digest is called on invoice item delete

sfount:
This commit call `this.digest()` when an invoice item is removed from
the invoice item service. This ensures that totals are calculated and kept
up to date. It also adds a check case for empty invoices which previously
send a `400 BAD REQUEST` to the server.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/939

---

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/client/src/partials/patient_invoice/patientInvoice.js
+++ b/client/src/partials/patient_invoice/patientInvoice.js
@@ -82,6 +82,12 @@ function PatientInvoiceController(Patients, PatientInvoices, PatientInvoiceForm,
     // update value for form validation
     detailsForm.$setSubmitted();
 
+    // make sure there are actually items to validate
+    if (vm.Invoice.store.data.length === 0) {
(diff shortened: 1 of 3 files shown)


client/src/partials/patient_invoice/patientInvoice.js
```diff
@@ -82,6 +82,12 @@ function PatientInvoiceController(Patients, PatientInvoices, PatientInvoiceForm,
     // update value for form validation
     detailsForm.$setSubmitted();
 
+    // make sure there are actually items to validate
+    if (vm.Invoice.store.data.length === 0) {
```
jniles, line 86: :+1: Nice catch!

jniles:
LGTM :+1:  Once tests pass, I'll merge it.

PR #953, opened 2016-11-27
feat(bhAccountSelect): implement uiSelect account

jniles:
This commit implements a generic account select for the app called
bhAccountSelect.  It uses a uiSelect, but introduces sane defaults.
This is tested on both the billing_services and Debtor Groups modules.

The defaults used are:
 1. debounce input by 200 milliseconds for better responsive behavior
 2. order accounts in hierarchical order
 3. disable title accounts

For performance reasons, this is the first component that uses one-way
bindings exclusively.  That means controllers will have to implement
callbacks to change the model data.

The usage is like this:
```html
<bh-account-select
  name="account_id"
  initial-value="$ctrl.model.account_id"
  on-update="$ctrl.updateAccount(account)"
  validation-trigger="Form.$submitted"
  required="true">
</bh-account-select>
```

The initial model value is the account id, and is set via the
`initial-value` parameter.  All updates are internal the component ...
to get the changed values, you must subscribe to the `on-update`
function callback.

To test, simply use the `components.accountSelect.set()` method.  Like this:
```js
// first param is the "name" binding
// second param is the thing to search for
components.accountSelect.set('account_id', '47001');
```

-----

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/client/src/partials/templates/bhAccountSelect.tmpl.html
+++ b/client/src/partials/templates/bhAccountSelect.tmpl.html
@@ -1,0 +1,28 @@
+<ng-form name="{{ $ctrl.name }}" data-bh-account-select>
+  <div class="form-group" ng-class="{ 'has-error' : $ctrl.validationTrigger && AccountForm.account.$invalid }">
+    <label class="control-label">
+      {{ $ctrl.label | translate }}
+    </label>
+
+    <ui-select
+     name="account"
+     ng-model="$ctrl.initialValue"
+     on-select="$ctrl.onSelect($item)"
+     ng-required="$ctrl.required">
+      <ui-select-match placeholder="{{ 'FORM.PLACEHOLDERS.ACCOUNT' | translate }}">
+       <strong>{{$select.selected.number}}</strong> <span>{{$select.selected.label}}</span>
+      </ui-select-match>
+      <ui-select-choices
+       ui-select-focus-patch
+       ui-disable-choice="account.type_id === $ctrl.TITLE_ACCOUNT"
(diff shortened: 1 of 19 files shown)


sfount:
@jniles There are conflicts with a recent re-factor on debtor groups

jniles:
@sfount, everything fixed!

I would appreciate an in-depth review...  this is something we will likely want to design once and touch as few times as possible in the future.

client/src/partials/templates/bhAccountSelect.tmpl.html
```diff
@@ -0,0 +1,29 @@
+<ng-form name="{{ $ctrl.name }}" data-bh-account-select>
+  <div class="form-group" ng-class="{ 'has-error' : $ctrl.validationTrigger && AccountForm.account.$invalid }">
+    <label class="control-label">
+      {{ $ctrl.label | translate }}
+    </label>
+
+    <ui-select
+     name="account"
+     ng-model="$ctrl.initialValue"
+     ng-model-options="{ 'debounce' : 200 }"
```
sfount: This currently uses a `debounce` which seems to cause issues described https://github.com/IMA-WorldHealth/bhima-2.X/issues/972. We should address this bug before allowing this everywhere in master.

sfount (changes requested):
These changes look good to me - no problem with the design or implementation of the component. This is very clean work.

There is one known issue with using a debounce that should be addressed before this component is used on the master branch.

sfount:
Best practices for filtering, tracking, displaying and debouncing on all account selects should be discussed and researched. 

This pull request will be re-opened or replaced with the implementation of `bhAccountSelect`.

Issue #960, opened 2016-11-27
(bug) landing page throw an error for the debtors panel


mbayopanda:
With Lukonga data, the debtor panel in the landing page throw an error : 

![error](https://cloud.githubusercontent.com/assets/5445251/20649492/428f616e-b476-11e6-9846-f707d523202b.png)

sfount:
@mbayopanda This report is driven by the debtor aged report. This likely means that the report either breaks or does not return an `aggregates` (totals) property if there is no data. That is where I would look.

mbayopanda:
@sfount Okay, i try BHIMA with `lukonga-latest.sql` file. Thanks

PR #961, opened 2016-11-27
(Fix) Handle no debtor information corner case

sfount:
This commit gracefully handles the server returning no information
on the aggregates of debtor groups. This is usually because there are
no transactions or even no debtor groups at all and should never occur
in a set up system.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/960


jniles:
LGTM.

Issue #963, opened 2016-11-28
(Bug) Exchange rate function precision


sfount:
The SQL function `GetExchangeRate` cannot return rates with any precision more than `INTEGER`. It is currently defined as `DECIMAL DETERMINISTIC`. The `DECIMAL` type must be defined with how a number of units before and after the decimal point. 

This causes exchange rates that are not integers to be rounded - providing incorrect or unknown values for exchange.

PR #964, opened 2016-11-28
(Fix) Ensure core GetExchangeRate function has precision

sfount:
This commit increases the number of decimal places returned by the
`GetExchangeRate` method. Currently `DECIMAL` is ued, this allows
no decimal places without specifying. This causes exchange rate values
to be rounded to the nearest integer.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/963


server/models/functions.sql
```diff
@@ -49,7 +49,7 @@ CREATE FUNCTION GetExchangeRate(
   currencyId INT,
   date TIMESTAMP
 )
-RETURNS DECIMAL DETERMINISTIC
+RETURNS DECIMAL(19, 8) DETERMINISTIC
```
sfount, line 52: This is the value specified for `rate` on `exchange_rate` in `schema.sql`

PR #974, opened 2016-11-28
(Fix) Account management UX

sfount:
This pull request refactor's the account management page, factoring feedback from module use during installation. 

**Filter Account Title/ Number**
_Filter disabled_
filter_off

_Filter enabled_
filter_on

_Filter results - maintains structure of accounts_
maintains_structure

**Update account type**
Account types had been made read only in the edit view to ensure that account types were never changed during actual use of the application - however this proved very impractical for application installation. This feature should be developed further in the future (allowing disabling once complete or introducing disabling with a more stable build).
edit_title

**Additional UX fixes**
* Account deletion simplified and updated - the grid is not forced to refresh on deletion and all bugs around this have been fixed
* Account update scrolls to the updated account - only if the parent account is updated is a grid redraw required (this allows movement of entire groups of accounts)
* Ensure title accounts that are created are displayed as title accounts (bold)
* Account selection UI-select correctly filters results on `hrlabel`

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -21,7 +21,7 @@
 }
 
 /**
- * IMA colors 
+ * IMA colors
  */
 .bg-ima-blue { background-color : #085484; color: #fff; }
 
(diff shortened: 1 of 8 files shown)


jniles:
LGTM.  These are awesome changes and make the page very usable.  :+1:

PR #977, opened 2016-11-29
(Fix) Account - Block titles with children from having their types updated

sfount:
This commit makes sure that a title account has no children before
it's type id can be changed. This ensures there are no other account
types with children on a well set up chart of accounts.

- [x] Requires French translation

@IMA-WorldHealth/bhima-core 
Can someone comment here with a good translation for:
```js
"TYPE_CHANGE_BLOCKED" : "You cannot change the type of a title account with children."
```

### DIFF
--- a/client/src/partials/accounts/edit/accounts.edit.js
+++ b/client/src/partials/accounts/edit/accounts.edit.js
@@ -136,6 +136,21 @@ function AccountEditController($rootScope, $state, AccountStore, Accounts, Notif
   }
 
 
+  vm.titleChangedValidation = titleChangedValidation;
+
+  // @todo form validation using validators on a component
+  function titleChangedValidation(newAccountType) {
+    var notTitleAccount = Number(newAccountType) !== Constants.accounts.TITLE;
+    var hasChildren = vm.account.children.length;
+
+    if (notTitleAccount && hasChildren) {
+      vm.invalidTitleAccount = true;
+    } else {
+      vm.invalidTitleAccount = false;
+    }
+  }
+
+
   /** @todo re-factor method - potentially these two actions should be split into two controllers */
   function updateAccount(accountForm) {
     // only require form to have changed if this is not the create state (no initial values)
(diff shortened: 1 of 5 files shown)


client/src/partials/accounts/edit/accounts.edit.js
```diff
@@ -136,6 +136,19 @@ function AccountEditController($rootScope, $state, AccountStore, Accounts, Notif
   }
 
 
+  vm.titleChangedValidation = titleChangedValidation;
+
+  // @todo form validation using validators on a component
+  function titleChangedValidation(newAccountType) {
+    console.log(newAccountType, Constants.accounts.TITLE);
```
jniles: Don't forget to clean this up!

client/src/partials/accounts/edit/accounts.edit.js
```diff
@@ -136,6 +136,19 @@ function AccountEditController($rootScope, $state, AccountStore, Accounts, Notif
...
+    if ((Number(newAccountType) !== Constants.accounts.TITLE) && vm.account.children.length) {
```
jniles: It would be easier to read if these were variables:
 ```js
var isTitleAccount = Number(newAccountType) === Constants.accounts.TITLE;
var hasChildren = vm.account.children.length;

if (!isTitleAccount && hasChildren) {
  // etc
}
```

**Review by jniles — APPROVED** [review]
Changes LGTM.  Once translated, these can be merged.

**Comment by jniles** [comment]
@sfount: My suggestion "Vous ne pouvez pas modifier le type d'un compte de titre qui a des sous-comptes."

File: client/src/partials/accounts/edit/accounts.edit.js (hunk 2 above)
Comment (sfount) [reply client/src/partials/accounts/edit/accounts.edit.js]: Good idea!

File: client/src/partials/accounts/edit/accounts.edit.js [hunk 3]
```diff
@@ -136,6 +136,21 @@ function AccountEditController($rootScope, $state, AccountStore, Accounts, Notif
   }
 
 
+  vm.titleChangedValidation = titleChangedValidation;
+
+  // @todo form validation using validators on a component
+  function titleChangedValidation(newAccountType) {
+    var notTitleAccount = Number(newAccountType) !== Constants.accounts.TITLE;
```
jniles, line 143: This is better than what I suggested :+1:

PR #981, opened 2016-11-29
fix(inventory): inventory patches

### LINKED ISSUE #930: (refactor): move Inventory.getInventoryItems() to Inventory.read()
There are quite a few modules (purchase, Patient Invoice) that use `Inventory.getInventoryItems()` instead of the standard route `Inventory.read()`.  These modules should be updated and `getInventoryItems()` removed from the client service.

### LINKED ISSUE #955: (Bug) Negative values can be set for price list items
Currently price lists can be set as values or as percentages. 

If set as a percentage this will be applied to the sale price of an inventory item, a negative percentage will reduce the price and a positive percentage will increase it.

If set as a value a negative value, this will be allowed by the client - this will be directly set at invoice. Although an invoice cannot be submitted with any value 0 or less this is still set. 

**Problem**
You can set a price list item as a value, less than 0 

**Expected Behavior**
Either the client or server will block a price list item being set as a value and less than 0

### LINKED ISSUE #978: (bug) Grouped Inventory still has "edit" on group header rows
This Inventory List allows a user to edit the headers of grouped rows.  This obviously doesn't make any sense.  The "edit" button should be hidden for cells in row headers.

See below:
inventoryeditableheaders
_Fig 1: Editable Header Rows in Inventory List_

jniles:
This commit removes the requirement for each inventory to have a
sale_account.  The requirement is motivated by needing a sale_account
for any gain from sales.  Since some medicines will not be sold, we 
still need to have a asset account without a sale account.  Also, any
physical inventory will not have a sale_account.

This commit fixes the inventory module to restrict users from putting in
negative prices.

Closes #955.

This commit hides the edit button on grouped inventory item cells.

Closes #978.  Closes #930.

----
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


client/src/partials/inventory/list/modals/actions.tmpl.html
```diff
@@ -44,19 +44,20 @@
       </div>
     </div>
 
-    <div class="form-group">
+    <div class="checkbox">
```
sfount, line 47: :+1:

sfount (approved):
Changes LGTM. I can confirm that it closes addressed issues.

Issue #980, opened 2016-11-29
(Bug) Posting Invoices ignore inventory items without sales account


sfount:
If an invoice is created using inventory items without a sales account, nothing is written for the sale item. This causes invoices to not balance. 

This behavior should be blocked by the posting process and reported to the invoice creation module. It should be clear that there is no sales account for the inventory group used. 

- [ ] Block invalid sales account on the server; Imbalanced transactions cannot be created
- [ ] Block the client from requesting badly configured items

jniles:
This was closed in #983.

PR #983, opened 2016-11-29
(Fix) Urgent: Block invalid sales account in PostingInvoice

sfount:
This commit ensures that imbalanced transactions are not posted by
the BHIMA Posting core. If an invalid (NULL) sales account is found
the server will throw a bad request based on the MySQL error thrown.

Contributes to https://github.com/IMA-WorldHealth/bhima-2.X/issues/980


jniles (approved):
This PR LGTM :+1:

PR #994, opened 2016-11-30
(Feature) Add optional flag to use the combined ledger; debtor debt

sfount:
This commit adds the `combinedLedger` flag to the aged debtor debt
report. This changes the 'source' of the debt query, using the
`general_ledger` by default and using the `combined_ledger` if the
flag is set.

It also updates the combined ledger to use the same column names as
the general ledger, querries can alias this after it is requested
if required.

### DIFF
--- a/server/controllers/finance/reports/debtors/index.js
+++ b/server/controllers/finance/reports/debtors/index.js
@@ -66,11 +66,13 @@ function queryContext(queryParams) {
   const params = queryParams || {};
   const havingNonZeroValues = ' HAVING total > 0 ';
   const includeZeroes = Boolean(Number(params.zeroes));
+  const useCombinedLedger = Boolean(Number(params.combinedLedger));
 
   // format the dates for MySQL escape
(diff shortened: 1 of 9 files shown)


server/controllers/finance/reports/debtors/index.js
```diff
@@ -66,11 +66,13 @@ function queryContext(queryParams) {
   const params = queryParams || {};
   const havingNonZeroValues = ' HAVING total > 0 ';
   const includeZeroes = Boolean(Number(params.zeroes));
+  const useCombinedLedger = Boolean(Number(params.combinedLedger));
 
   // format the dates for MySQL escape
   const dates = _.fill(Array(4), new Date(params.date));
 
   const data = {};
+  const source = useCombinedLedger ? 'combined_ledger' : 'general_ledger';
```
jniles, line 75: This is a really clean switch :+1:

jniles (approved):
Change LGTM.  This was a poor design decision at the beginning, glad to see it corrected.

Issue #971, opened 2016-11-28
FORM.ERROR.INVALID not translated


lomamech:
In the patient registration form, the FORM.ERROR.INVALID error message is not translated

jniles:
@lomamech, can you tackle this?

jniles:
Closed in #998.

Issue #970, opened 2016-11-28
Using Village and Sector not correct In the Urban town 


lomamech:
Modify in the management of the locations the use of the label Village and sector,
The village label may be replaced by "Commune / Village" and Sector by "Sector / City"

jniles:
This is a good idea!

Seems to just be a question of translation, can we add that easily?

jniles:
Closed in #998.

PR #985, opened 2016-11-29
small fixing in patient invoice page and translation

DedrickEnc:
This PR fixes #820 , #971 , #970 
Thank you for contributing!

Before submitting this pull request, please verify that you have:
 - [x] Run your code through JSHint.  [Check out our styleguide](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/docs/STYLEGUIDE.md).
 - [x] Run integration tests.
 - [x] Run end-to-end tests.
 - [x] Accurately described the changes your are making in this pull request.

For a more detailed checklist, [see the official review checklist](https://docs.google.com/document/d/1nupLVLRXgSZJQo_acLgrwvPnN8RukfSiwRhSToj81uU/pub) that this PR will be evaluated against.

Ensuring that the above checkboxes are completed will help speed the review process and help build a stronger application.  Thanks!

PR #998, opened 2016-11-30
small fixing in patient invoice page and translation

### LINKED ISSUE #820: Providing a default description in patient invoice
Our patient invoice User interface does not let you submit the bill if the description is not provided, so it can be boring for a user to provide a description every time, essentially if the hospital can records a big amount of record per day.
So the user can be lazy and provide a bad description just to be able to post their invoice, so I think a default description is worth.

DedrickEnc:
This PR fixes #820 , #971 , #970 
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/client/src/partials/patient_invoice/patientInvoice.html
+++ b/client/src/partials/patient_invoice/patientInvoice.html
@@ -37,6 +37,7 @@
                     name="service"
                     ng-model="PatientInvoiceCtrl.Invoice.details.service_id"
                     ng-options="service.id as service.name for service in PatientInvoiceCtrl.services"
+                    ng-change="PatientInvoiceCtrl.setService(PatientInvoiceCtrl.Invoice.details.service_id)"
                     required>
                     <option disabled="disabled" value="">{{ "FORM.LABELS.SERVICE" | translate }}</option>
                   </select>
--- a/client/src/partials/patient_invoice/patientInvoice.html
+++ b/client/src/partials/patient_invoice/patientInvoice.html
@@ -35,9 +35,9 @@
                   <select
                     class="form-control"
                     name="service"
-                    ng-model="PatientInvoiceCtrl.Invoice.details.service_id"
-                    ng-options="service.id as service.name for service in PatientInvoiceCtrl.services"
-                    ng-change="PatientInvoiceCtrl.setService(PatientInvoiceCtrl.Invoice.details.service_id)"
+                    ng-model="PatientInvoiceCtrl.Invoice.service"
+                    ng-options="service as service.name for service in PatientInvoiceCtrl.services"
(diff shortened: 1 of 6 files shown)


client/src/partials/patient_invoice/patientInvoice.js
```diff
@@ -157,6 +162,17 @@ function PatientInvoiceController(Patients, PatientInvoices, PatientInvoiceForm,
     }
   }
 
+  function handleServiceChange () {
```
sfount, line 165: :+1: this is clear.

client/src/partials/patient_invoice/patientInvoice.js
```diff
@@ -157,6 +162,17 @@ function PatientInvoiceController(Patients, PatientInvoices, PatientInvoiceForm,
...
+    vm.Invoice.setService(vm.Invoice.service);
+    updateDescription();
```
sfount, line 167: Significantly reduced dependencies, requests and lines of code. Great work for maintainability.

sfount (commented):
Code here LGTM! Confirmed that this addresses comments from the previous review.

client/src/partials/patient_invoice/patientInvoice.html
```diff
@@ -35,8 +35,9 @@
                   <select
                     class="form-control"
                     name="service"
-                    ng-model="PatientInvoiceCtrl.Invoice.details.service_id"
-                    ng-options="service.id as service.name for service in PatientInvoiceCtrl.services"
+                    ng-model="PatientInvoiceCtrl.Invoice.service"
```
sfount, line 38: The end to end tests use the model `service_id` to locate this component. They will need to be updated to pass with this model/ value change.

jniles:
Just a note - because this no longer uses the `service_id` as the selector, it doesn't play well with `AppCache`.  Services are no longer recoverable from cache.

Steps to reproduce:
 1. Create a bill for TPA1.
 2. Select *distributable*, *11/01/2016*, and *test service* as details.
 3. Make an invalid row (for example, with a negative cost)
 4. Submit.
 5. Refresh the browser.
 6. Select TPA1.
 7. Click "Recover Cache"

You will notice that *distributable* is set, *11/01/2016* is the date, and the old description is back.  But *test service* is not set.

We can fix this in a separate PR if necessary.

DedrickEnc:
Merci, I Weill see that tomorow

sfount:
This looks good to me - this pull request addresses the comments and review made in https://github.com/IMA-WorldHealth/bhima-2.X/pull/985. 

@jniles has pointed out that this breaks the cache, if a cached invoice is loaded the service in the description will be wrong. This feature can be made use of right now - @DedrickEnc Can you create an issue to track this bug?

PR #1001, opened 2016-12-01
Modify the load function, to avoid using $storage.paths and $storage.…

lomamech:
Modify the load function, to avoid using $storage.paths and $storage.project storaged during logging

@jniles


jniles:
@lomamech, this change works surprisingly well.  However, for some reason, it breaks the services page.  I think it's because `$storage.project` was deleted.

I like idea here - but I think a better option is to send an `$http` request to the server, and replace `$storage.project`, `$storage.paths`, etc, with the new data from the server.  That way, we are sure the client and server are in sync.

jniles:
@lomamech, this PR breaks the Karma tests for Session.  I can fork and fix those - they are a bit complicated.

Issue #1005, opened 2016-12-01
(feature request) User story: Show list of patients by patient group


jniles:
The users of 2.x have requested the ability to see all patients in a particular patient group.  This can be easily accomplished by filtering the patient registry if we add a new filter.

**Proposal**
 1. Add a new ui-select filter on the patient registry modal to allow filtering by one or many patient groups. (mulit-select)
 2. Add a link from the patient groups page that shows the number of patients in a particular patient group.  When clicked, it will go to the patient registry with the patient_group filter applied.

jniles:
Note: this should be implemented for debtor groups as well.

PR #1012, opened 2016-12-03
(Feature) Filter patient registry by patient and debtor groups

sfount:
This pull request implements two new filters on the patient registry, patient and debtor groups. 
Any patients belonging to the patient or debtor group selected will be shown in the results. 

It also adds a `filter` parameter to the patient registry page to allow other modules referencing the page to request filters to be applied. 

This allows patient and debtor groups to provide lists of patients that belong to those groups by passing in the UUID of the selected group. The patient registry will then override any previous filters and show the results requested by the linking module. 

**Debtor Groups - Number of subscribed debtors link**
debtors_link

**Results filtered directly from link**
filtered_patients

This commit also:
* Cleans up the registry search modal interface to bring it more in line with 2.x standards
* Removes legacy search modal template and controller from the `patient/registry` folder

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/1005


jniles:
LGTM :+1:

Hopefully we get more user feedback like this!

client/src/js/services/PatientService.js
```diff
@@ -230,14 +230,23 @@ function PatientService($http, util, Session, $uibModal, Documents, Visits) {
...
+
     // returns columns from filters
     return columns.filter(function (column) {
+      var LIMIT_UUID_LENGTH = 6;
       var value = params[column.field];
 
       if (angular.isDefined(value)) {
         column.value = value;
+
+        if (column.field === 'debtor_group_uuid' || column.field === 'patient_group_uuid') {
```
jniles, line 246: If we have many more of these, we should use an array to check for matches.

client/src/partials/debtors/groups.list.html
```diff
@@ -47,7 +47,7 @@ <h4 style="margin-top : 0px;"><span ng-if="debtorGroup.locked" class="fa fa-lock
                   </div>
 
                   <div class="col-md-6">
-                    <span class="fa fa-bars" aria-hidden="true"></span> <b>{{debtorGroup.total_debtors}}</b> {{ "DEBTOR_GROUP.DEBTOR.SUBSCRIBED" | translate }}
+                    <a ui-sref="patientRegistry({ filters : { debtor_group_uuid : debtorGroup.uuid }})"><span class="fa fa-bars" aria-hidden="true"></span> <b>{{debtorGroup.total_debtors}}</b> {{ "DEBTOR_GROUP.DEBTOR.SUBSCRIBED" | translate }}</a>
```
jniles, line 50: This is super cool!

client/src/partials/patients/registry/search.modal.html
```diff
@@ -5,17 +5,17 @@
   bh-form-defaults
   novalidate>
 
-  <div class="modal-header">
+  <div class="modal-header" style="box-shadow : 0px 2px 4px -4px black; z-index:1000">
```
jniles, line 8: ... why was this necessary?

client/src/partials/patients/registry/search.modal.html
```diff
@@ -5,17 +5,17 @@
...
     <ol class="headercrumb">
       <li class="static">{{ "FORM.INFO.PATIENTS" | translate }}</li>
       <li class="title">{{ "FORM.INFO.SEARCH" | translate }}</li>
     </ol>
   </div>
 
-  <div class="modal-body modal-tall" style="overflow: auto;">
+  <div class="modal-body" style="max-height : 400px; overflow: auto;">
```
jniles, line 15: I've been doing max-heights based on the viewport height using the `vh` unit.   No matter what we do, we should standardize it - which do you prefer?

client/src/partials/patients/registry/search.modal.html
```diff
@@ -83,7 +83,7 @@
 
     <fieldset>
       <legend>
-        <i class="fa fa-users"></i> {{ "FORM.LABELS.PATIENT_DETAILS" | translate }}
+        {{ "FORM.LABELS.PATIENT_DETAILS" | translate }}
```
jniles, line 86: :+1:  This is a cleaner look.

client/src/partials/patients/registry/search.modal.html
```diff
@@ -151,11 +154,58 @@
           {{ "FORM.LABELS.FEMALE" | translate }}
         </label>
       </div>
+
+      <div class="form-group">
+        <label class="control-label">
+          {{ "FORM.LABELS.DEBTOR_GROUP" | translate }}
+        </label>
+
+        <!-- @todo make this a component - it is used everywhere. Calls back method
```
jniles, line 163: I gotchu : https://github.com/IMA-WorldHealth/bhima-2.X/issues/1015

server/controllers/medical/patients/index.js
```diff
@@ -443,6 +444,12 @@ function find(options) {
     delete options.dateBirthTo;
   }
 
+  if (options.patient_group_uuid) {
+    conditions.statements.push('(SELECT COUNT(uuid) FROM assignation_patient where patient_uuid = q.uuid AND patient_group_uuid = ?) = 1');
```
jniles, line 448: This should have a comment :+1:

jniles (approved):
The code is clean and has some good comments.  The state-based routing is definitely a huge help at initializing filters.

client/src/partials/patients/registry/search.modal.html
sfount: **With this style**
with_shadow

**Without this style**
without_style

As it will always scroll it shows that there's depth between the header and the body. I think?

This method is also used on the primary application header bar on some pages - it could certainly be standardised as a class!

client/src/partials/patients/registry/search.modal.html
sfount: One resizes with the screen and the other doesn't - I think in this case it would be better for the modal to not stretch. 

I agree this should be standardized though - there could be two classes if both of these behaviours are required.

PR #1022, opened 2016-12-04
Complete translate file

lomamech:
- Complete translate file
- Using class text-danger for showing warnig in red

@jniles


jniles:
LGTM. :+1:

PR #1023, opened 2016-12-04
Prevent the choose of the date in the future

### LINKED ISSUE #916: (bug) Purchase Orders lets you choose a future date
The Purchase Order module allows a user to choose a date that is in the future.  This should not be the case.

**Steps to Reproduce**
 1. Go to the purchase page.
 2. Try selecting a future date from the `bhDateInput`.

**Expected Result**
The user is blocked with a disabled button and pointer.

**Actual Result**
The user is able to select whatever date they so desire.

lomamech:
Prevent the choose of the date in the future

@jniles 

Closes #916.


jniles:
LGTM. :+1:.

PR #1071, opened 2016-12-14
Fix balance report

### LINKED ISSUE #910: (bug) Confusing Balance Report
The balance report is confusing in its current form.  Here are some improvements that can be made: 
 - [x] Translate the report into French.  At the moment is only renders in English.
 - [ ] Put the dates in the column headers.  At the moment, it is hard to tell what the date range is.
 - [ ] The report says "all accounts" but only displays the accounts with data.  I think this should be an option flag in the modal - filter out accounts without transactions, filter out balanced accounts or show all accounts.
 - [ ] Should the user select a `period` instead of a date?  Or a fiscal year?

I've attached a copy of the present Balance Sheet on all accounts for your information:
balancereportconfusion
_Fig 1: The Balance Report for all accounts_.

mbayopanda:
This PR : 
- Add month information on balance columns
- Allow to display all accounts 

Close #910

### DIFF
--- a/client/src/partials/reports/modals/balance.config.js
+++ b/client/src/partials/reports/modals/balance.config.js
@@ -25,7 +25,7 @@ function BalanceReportConfigController($state, ModalInstance, Notify, Languages,
     { number : 6, name : 'ACCOUNT.COST' },
     { number : 7, name : 'ACCOUNT.REV' },
     { number : 8, name : 'ACCOUNT.EXP_PROD' },
-    { number : '*', name : 'ACCOUNT.ALL_ACCOUNT' }
+    { number : '*', name : 'ACCOUNT.ALL_CLASSES' }
   ];
 
   // init date
--- a/client/src/partials/reports/modals/balance.config.js
+++ b/client/src/partials/reports/modals/balance.config.js
@@ -53,7 +53,8 @@ function BalanceReportConfigController($state, ModalInstance, Notify, Languages,
     vm.$loading = true;
 
     var options = {
-      label : vm.label,
+      account_option: vm.accountOption,
+      label: vm.label,
       classe: vm.classe.number,
       classe_name: vm.classe.name,
       lang: Languages.key
--- a/client/src/partials/reports/modals/balance.modal.html
+++ b/client/src/partials/reports/modals/balance.modal.html
@@ -41,6 +41,19 @@
     </div>
   </div>
 
+  <!-- accounts option choosen -->
+  <fieldset class="form-group">
+    <legend>{{ 'BALANCE.ACCOUNT_CHOICE' | translate }}</legend>
+    <label>
+      <input type="radio" ng-model="ReportConfigCtrl.accountOption" ng-value="'all'">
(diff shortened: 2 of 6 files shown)


client/src/partials/reports/modals/balance.config.js
```diff
@@ -25,7 +25,7 @@ function BalanceReportConfigController($state, ModalInstance, Notify, Languages,
     { number : 6, name : 'ACCOUNT.COST' },
     { number : 7, name : 'ACCOUNT.REV' },
     { number : 8, name : 'ACCOUNT.EXP_PROD' },
-    { number : '*', name : 'ACCOUNT.ALL_ACCOUNT' }
+    { number : '*', name : 'ACCOUNT.ALL_CLASSES' }
```
jniles, line 28: I know you didn't write this code, but in the future, we should put these "Account Classes" in the `bhConstants` or just infer them from the first level title accounts.

client/src/partials/reports/modals/balance.modal.html
```diff
@@ -41,6 +41,19 @@
     </div>
   </div>
 
+  <!-- accounts option choosen -->
```
jniles, line 44: Just so you know, bootstrap has classes for radio buttons.

[Documentation](http://getbootstrap.com/css/#checkboxes-and-radios)

server/controllers/finance/reports/balance/index.js
```diff
@@ -223,5 +223,32 @@ function balanceReporting(params) {
   .then(function (rows) {
     data.middle = rows;
     return data;
+  })
+  .then(function (rows) {
+    // Manual mixing 
+    const TITLE_ACCOUNT_TYPE = 4;
```
jniles, line 229: This is a good declaration :+1:

In the future, when you have a constant like this, it would be better to put it outside the function, in the main body of the module.  That way, it is not re-defined every single HTTP request.  Like this:

```js
const TITLE_ACCOUNT_TYPE = 4;

function balanceReporting() {
  // something something batman...
}
```

server/controllers/finance/reports/balance/index.js
```diff
@@ -223,5 +223,32 @@ function balanceReporting(params) {
...
+    // fill with zero if all accounts 
+    sql = 
+      `SELECT a.number, a.id, a.label, a.type_id, a.is_charge, a.is_asset, '0' AS credit, '0' AS debit 
+       FROM account a WHERE a.type_id <> ${TITLE_ACCOUNT_TYPE} AND a.locked = 0;`;
+    
+    return query.accountOption === 'all' ? db.exec(sql) : false;
+  })
+  .then(function (rows) {
+
+    if (!rows) { return data; }
```
jniles, line 240: Cheeky :dancer:

jniles (approved):
LGTM. :+1:

In the future, we really must write as much of our report logic in SQL as possible.  Javascript is really bad at math.  I will merge this PR since there is no math done in JS.

Good code overall!  Very clear.

jniles:
@mbayopanda, another thing .. in the future, it would be best to include a screenshot with your report so that way we can all see it :)

PR #1077, opened 2016-12-15
perf(settings): use translate directive

### LINKED ISSUE #1074: (perf) Favor translate directive over the translate filter
According to the [angular-translate site](https://angular-translate.github.io/docs/#/guide/05_using-translate-directive), the translate directive is a better option than the filter.  It also may result in much more readable code, for example:
```html
<span>{{ 'SOME.KEY' | translate }}</span>
<!-- versus -->
<span translate>SOME.KEY</span>
```

The performance of these two should be compared.  In particular, we should pay attention to the number of `$watchers` on every page.  If we can drastically reduce the number of `$watchers`, we should do a major overhaul and refactor the entire application.

This should probably be investigated as soon as possible.  I'll look into it sometime this week.

jniles:
This commit modifies the tree and settings page to use the `translate`
directive instead of the translate filter.  This result in an average
decrease of `$watchers` by 50%.  The original page average was 648
`$watchers` with the tree fully expanded, but falls to 351 `$watchers`
by using the translate directive.

This commit determines the validity of #1074 and closes #1074.

----
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/client/src/index.html
+++ b/client/src/index.html
@@ -68,7 +68,7 @@
             <li>
               <a class="session" ng-click="AppCtrl.settings()">
                 <span class="fa fa-cog"></span>
-                <span class="link-label">{{ "SETTINGS.TITLE" | translate }}</span>
+                <span class="link-label" translate>SETTINGS.TITLE</span>
               </a>
             </li>
             <li>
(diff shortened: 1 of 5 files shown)


sfount (approved):
I did not even know this directive existed.

This is incredible analysis! Not using a filter to constantly watch for changes should have a significant impact on performance.

sfount:
I have tested this code and the functionality is equivalent to the filter method but with the benefits described.

Issue #1003, opened 2016-12-01
(bug) Cashflow report default date interval throws and error


jniles:
The cashflow report default date interval cannot be submitted with success.  The server throws an error 'Bad Date Interval'.  This should be fixed - having invalid defaults is a really poor user experience.

jniles:
Actually, the bug is that the cashflow report is not a default date.  The server sends an error if no data is received from the database.

jniles:
Closed in #1081.

PR #1081, opened 2016-12-15
Fix cashflow report

mbayopanda:
This PR fixes the cash-flow report date handling : 
- By setting up a default date, and
- Handling correctly dates interval with sql

Issue #1003

### DIFF
--- a/server/controllers/finance/reports/cashflow/index.js
+++ b/server/controllers/finance/reports/cashflow/index.js
@@ -55,6 +55,9 @@ function processingCashflowReport(params) {
     throw new BadRequest('Cashbox is missing', 'ERRORS.BAD_REQUEST');
   }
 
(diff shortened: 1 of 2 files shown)


server/controllers/finance/reports/cashflow/index.js
```diff
@@ -189,6 +192,9 @@ function processingWeekCashflow(params) {
     throw new BadRequest('Cashbox is missing', 'ERRORS.BAD_REQUEST');
   }
 
+  params.dateFrom = moment(params.dateFrom).format('YYYY-MM-DD').toString();
```
jniles, line 195: Why is this so complex?  All other date ranges seem to work with the simply JS and SQL parsing...

jniles (commented):
This seems like a really brute-force method to implement date ranges...  It should just work with `DATE()`.  We should figure out what is going on here in the future.  For the moment, this works :+1:

Issue #1080, opened 2016-12-15
Creating database schema and Implementing the server API


DedrickEnc:
This issue references #809, it is for creating the database schema and the server API as discussed in the referenced issue beyond here are the task to complete : 

- [x] updating the database structure 

- [x] Creating the server side API 

- [x] Test

sfount:
This looks like a good base for building cost centers - if you begin by creating the database structure we can build on that. 

It might be good to think about what the server API will have to accomplish. 

* Read - List all cost centers or detailed information about a specific cost center 
* Create 
* Update 
* Delete - Does it make sense to delete a cost center? 

* Assign costs? - This API must be designed

EDIT: This has been discussed in https://github.com/IMA-WorldHealth/bhima-2.X/issues/809#issuecomment-254763835, this just needs to be verified and implemented

PR #1101, opened 2017-01-13
Fee center schema and API

DedrickEnc:
This PR adds an initial schema and API to continue the feature developing
Here are things to notice in this PR : 

- The table ```cost_center``` and ```profit_center``` are deleted
-  A new table ```fee_center``` is created```
- The modules cost center and profit center are deleted
- The service module is updated to fit with the  changes
- The delete functionality is not yet implemented and the integration  test for this functionality is ignored, because we did not talk about that during the design. 
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


DedrickEnc:
Thank you @jniles , I did it.

jniles:
:+1:

DedrickEnc:
closes #1080

DedrickEnc:
Can this PR be reviewed?

jniles:
It looks like everything here was added in #1105 which has a review.  Is that not the case?

DedrickEnc:
This is the first PR, which must be reviewed.

client/src/js/services/FinancialService.js
```diff
@@ -8,33 +8,25 @@ function FinancialService ($http, util) {
...
   service.readCostCenter = readCostCenter;
   service.readProfitCenter = readProfitCenter;
-  service.getCost = getCost;
-  service.getProfit = getProfit;
+  service.getFeeValue = getFeeValue;
 
   // this will read all cost centers
   function readCostCenter(id, params) {
-     var url = '/cost_centers/';
+     var url = '/fee_centers?is_cost=1';
```
jniles: This is old code so it's not a necessary fix, but you could also just extend the `params` with `{ is_cost: 1}`.  Like this:

```js
var baseUrl = '/fee_centers';

function readCenter(id, params) {
  return $http.get(baseUrl, { params : params })
}

function readProfitCenter(id, params) {
  var profitParameters = angular.extend(params, { is_cost : 0 });
  return readCenter(id, profitParameters);
}

function readCostCenter(id, params) {
  var costParameters = angular.extend(params, { is_cost: 1 });
  return readCenter(id, costParameters);
}
```

server/controllers/admin/services.js
```diff
@@ -29,17 +29,18 @@ const Topic = require('../../lib/topic');
...
-      JOIN enterprise AS e ON s.enterprise_id = e.id
-      LEFT JOIN cost_center AS cc ON s.cost_center_id = cc.id
-      LEFT JOIN profit_center AS pc ON s.profit_center_id = pc.id`;
+      SELECT 
+s.id, s.name, s.enterprise_id, s.cc_id, s.pc_id, e.name AS enterprise_name, 
+e.abbr, cc.id AS cc_id, cc.cost_center_name, pc.id AS pc_id, pc.profit_center_name 
+
+FROM service AS s
+JOIN enterprise AS e ON s.enterprise_id = e.id
+LEFT JOIN (SELECT fee_center.label AS cost_center_name, id FROM fee_center WHERE fee_center.is_cost = 1) AS cc ON s.cc_id = cc.id
```
jniles, line 42: Do you need to do this complex subquery?  It looks like it would be harder for MySQL to optimize... was the other query too slow?

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,263 @@
...
+**/
+
+
+var db = require('../../lib/db');
+var NotFound = require('../../lib/errors/NotFound');
+
+/**
+* Returns an array of fee centers
+*
+* @param {object} req The express request object
```
jniles: I think at this point, all developers can be expected to know how `expressjs` works.  There is no need to document another library's parameters.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,263 @@
...
+*/
+
+function list (req, res, next) {
+  'use strict';
+
+  var sql =
+    'SELECT fc.id, fc.label, fc.is_cost FROM fee_center AS fc ';
+  var filter = '', ordering = 'ORDER BY fc.label;';
+
+  if (req.query.full === '1') {
```
jniles: I think `req.query.detailed` makes a lot more sense.  It will also be standardized once @sfount's PR lands in the filter library.  Could you use `req.query.detailed` instead?

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,263 @@
...
+    sql =
+      `SELECT fc.id, fc.label, fc.project_id, fc.is_cost, fc.note, fc.is_principal, p.name, p.abbr, p.enterprise_id, p.zs_id
+      FROM fee_center AS fc JOIN project AS p ON fc.project_id = p.id `;
+  }
+
+
+  if(req.query.available === '1') {
```
jniles, line 39: What does this parameter do?  What is an available fee center?

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,263 @@
...
+      filter += ' AND ';
+    }
+    filter += `fc.is_cost = ${req.query.is_cost} `;
+  }
+
+  if(req.query.is_principal){
+    if(filter !== ''){
+      filter += ' AND ';
+    }
+    filter += `fc.is_principal = ${req.query.is_cost} `;
```
jniles, line 54: What happens if `req.query.is_cost` is `(DELETE FROM general_ledger WHERE 1); -- ` ?

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  sql = filter === '' ? sql + ordering : sql + 'WHERE ' + filter + ordering;
```
jniles, line 57: If you used `db.exec(sql, [params])`, you could write your code so that you didn't have to check if `filter === ''` all the time.

server/controllers/finance/generalLedger/index.js
```diff
@@ -46,5 +45,5 @@ function list(req, res, next) {
     .then(function (rows) {
         res.status(200).json(rows);
     })
-    .catch(next);
+    .catch(function (err){console.log(err);});
```
jniles: This should have proper error handling.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,263 @@
...
+function update (req, res, next) {
+  'use strict';
+
+  var queryData = req.body;
+  var feeCenterId = req.params.id;
+  var updateFeeCenterQuery = 'UPDATE fee_center SET ? WHERE id = ?';
+
+  delete queryData.id;
+
+  lookupFeeCenter(feeCenterId)
```
jniles, line 114: This used to be best practices until we created our `errors/*.js` libraries for error handling.  You can now just do a UPDATE and then throw `NotFound()` if the `affectedRows` is 0 --- less database queries = easier for raspberry pi!

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,263 @@
...
+/**
+* Update a fee center in the database
+*
+* @param {object} req The express request object
+* @param {object} res The express response object
+* @param {object} next The express object to pass the control to the next middleware
+*
+* @example
+* // PUT /fee_centers : update a fee center
+* var feeCenters = require('finance/fee_centers');
```
jniles: This file is `server/controllers/finance/feeCenter.js`.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,263 @@
...
+*
+*/
+
+function lookupFeeCenter(id) {
+  'use strict';
+
+  var sql =
+    'SELECT fc.id, fc.label, fc.is_cost, fc.note, fc.is_principal, fc.project_id FROM fee_center AS fc WHERE fc.id = ?';
+
+  return db.exec(sql, id)
```
jniles: You can use `db.one()` to handle this case automatically.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,263 @@
...
+      sql =
+        'SELECT ac.id FROM account AS ac WHERE ac.fc_id = ? AND ac.is_title=0';
+
+      return db.exec(sql, req.params.id);
+    })
+    .then(function (rows){
+
+      if (rows.length > 0) {
+        rows = rows.map(function (row) { return row.id;});
+        optionalCondition = ' OR gl.account_id IN (' + rows.join(',') + ')';
```
jniles: It would be much, much better to use `db.exec()`'s parameters instead of doing string concatenation in JS.

server/models/schema.sql
```diff
@@ -317,51 +316,51 @@ CREATE TABLE `consumption_service` (
...
-
-CREATE TABLE `cost_center` (
-  `project_id` smallint(5) unsigned NOT NULL,
+CREATE TABLE `fee_center` (
   `id` smallint(6) NOT NULL AUTO_INCREMENT,
-  `text` varchar(100) NOT NULL,
-  `note` text,
+  `project_id` smallint(5) unsigned NOT NULL,
+  `label` varchar(100) NOT NULL,
+  `is_cost` tinyint(1) DEFAULT 0,
```
jniles, line 323: Since you do a lot of queries on `is_cost`, as a performance improvement, you could put an INDEX on it.

server/models/schema.sql
```diff
@@ -317,51 +316,51 @@ CREATE TABLE `consumption_service` (
...
-CREATE TABLE `cost_center_assignation` (
-  `project_id` smallint(5) unsigned NOT NULL,
+CREATE TABLE `cost_assignation` (
   `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
-  `auxi_cc_id` smallint(6) NOT NULL,
+  `auxiliary_cost_id` smallint(6) NOT NULL,
   `cost` float DEFAULT 0,
-  `date` date DEFAULT NULL,
+  `period_id` mediumint(8) UNSIGNED NOT NULL,
+  `auxi_cc_id` smallint(6) NOT NULL,
```
jniles: What is `auxi_cc_id`?  Can it be removed since we have `auxiliary_cost_id`?

test/integration/feeCenter.js
```diff
@@ -0,0 +1,160 @@
...
+        helpers.api.errored(res, 404);
+      })
+      .catch(helpers.handler);
+  });
+
+
+
+  /* @todo - should this return a simple number? */
+  it('GET /fee_centers/:id/value returns the value of a provided fee center', function () {
+    return agent.get('/fee_centers/:id/value'.replace(':id', FETCHABLE_FEE_CENTER_ID))
```
jniles: Since we are using Node v4+, you can just use string substitution:
```js
return agent.get(`/fee_centers/${FETCHABLE_FEE_CENTER_ID}/value`):
```

jniles (changes requested):
Most of these comments are copied from https://github.com/IMA-WorldHealth/bhima-2.X/pull/1105.  I'm pretty sure there are some SQL injection vulnerabilities, as well as outdated code.  Those should be fixed before a merge.

client/src/partials/services/services.html
```diff
@@ -117,7 +117,7 @@
               <div class="form-group" ng-class="{ 'has-error' : CreateForm.$submitted && CreateForm.cost_center_id.$invalid }">
                 <label class="control-label">{{ "FORM.LABELS.COST_CENTER" | translate }}</label>
```
sfount: I've not looked through this entire PR as a review but it would be best for our updated files/ new code to use the latest `translate` standards. https://github.com/IMA-WorldHealth/bhima-2.X/issues/1074

server/controllers/admin/services.js
DedrickEnc: This query is more adapted in our new schema where cost center and profit center are combined.

server/controllers/finance/feeCenter.js
DedrickEnc: ok

server/controllers/finance/feeCenter.js
DedrickEnc: a fee center which is not linked to a service

server/controllers/finance/feeCenter.js
DedrickEnc: The sql will be wrong formatted, but I think to avoid attack the best thing is to use `params` right?

DedrickEnc:
This branch has some not useful commits, I will reopen a new PR

PR #1105, opened 2017-01-19
Cost center client side

DedrickEnc:
This PR is implementing a client side of a cost center module and it is highly dependent to to #1080 .
He are tasks to perform : 

- [x] User Interface implementation

- [x] Testing 

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


DedrickEnc:
@sfount

client/src/i18n/en.json
```diff
@@ -262,6 +262,16 @@
     "SET_AS"              : "Set as",
     "ON"                  : "on"
   },
+  "FEE_CENTER" : {
+    "ADDING_FEE_CENTER" : "Adding Fee Center",
+    "COST_CENTER" : "Cost center",
+    "CREATE"        : "Create",
```
jniles, line 268: Can this just be `FORM.LABELS.CREATE`?

client/src/i18n/en.json
```diff
@@ -262,6 +262,16 @@
...
+    "PRINCIPAL"     : "Main",
+    "PROFIT_CENTER" : "Profit center",
+    "RECORD_SAME"   : "No edited data",
+    "UPDATE"        : "Update",
```
jniles, line 272: Can this just be `FORM.LABELS.UPDATE`?

client/src/js/services/FeeCenterService.js
```diff
@@ -0,0 +1,86 @@
+angular.module('bhima.services')
+  .service('FeeCenterService', FeeCenterService);
+
+FeeCenterService.$inject = ['$http', 'util', '$translate'];
+
+function FeeCenterService($http, util, translate) {
```
jniles, line 6: I think you can reduce this entire service to the following lines:

```js
function FeeCenterService(PrototypeApiService, $translate) {
  var service = new PrototypeApiService('/fee_centers');

  service.fullRead = function fullRead(id) {
    return this.read(id, { full : 1 });
  };

  service.formatRecord = function formatRecord(list) {
    list.forEach(function (item) {
      item.type = item.is_cost === 1 ? translate.instant('FEE_CENTER.COST_CENTER') : translate.instant('FEE_CENTER.PROFIT_CENTER');
      item.principalState = item.is_principal === 1 ? translate.instant('FORM.LABELS.YES') : translate.instant('FORM.LABELS.NO');
    });
    return list;
   }

  return service;
}
```
Doing this would help with maintainability/testing, reduce the size of the JS shipped to the client, and reduce the amount of time Chrome spends parsing the JS.

client/src/js/services/FeeCenterService.js
```diff
@@ -0,0 +1,86 @@
...
+   * @param {Integer} id, the id of the fee center (optional)
+   * @return {object} a promise object, with the response.body inside.
+   * @example
+   * service.fullRead()
+   * .then(function (feeCenter){
+  *   your code here
+  *  });
+   **/
+
+  function fullRead (id){
```
jniles, line 44: I highly recommend that you just use the standard `read()` method, passing the optional parameter as a JSON object.

client/src/js/services/FinancialService.js
```diff
@@ -8,33 +8,25 @@ function FinancialService ($http, util) {
...
   service.readCostCenter = readCostCenter;
   service.readProfitCenter = readProfitCenter;
-  service.getCost = getCost;
-  service.getProfit = getProfit;
+  service.getFeeValue = getFeeValue;
 
   // this will read all cost centers
   function readCostCenter(id, params) {
-     var url = '/cost_centers/';
+     var url = '/fee_centers?is_cost=1';
```
jniles, line 15: This is old code so it's not a necessary fix, but you could also just extend the `params` with `{ is_cost: 1}`.  Like this:

```js
var baseUrl = '/fee_centers';

function readCenter(id, params) {
  return $http.get(baseUrl, { params : params })
}

function readProfitCenter(id, params) {
  var profitParameters = angular.extend(params, { is_cost : 0 });
  return readCenter(id, profitParameters);
}

function readCostCenter(id, params) {
  var costParameters = angular.extend(params, { is_cost: 1 });
  return readCenter(id, costParameters);
}
```

client/src/partials/fee_center/feeCenter.modal.html
```diff
@@ -0,0 +1,96 @@
+<form name="FeeCenterForm" bh-submit="FeeCenterModalCtrl.submit(FeeCenterForm)" bh-form-defaults novalidate>
+  <div class="modal-header">
+    <ol class="headercrumb">
+      <li ng-if="FeeCenterModalCtrl.isCreating" class="title">
+        {{ "FEE_CENTER.ADDING_FEE_CENTER" | translate }}
```
jniles, line 5: These should all use the `translate` directive to reduce `$$watchers`.

[From the documentation](https://angular-translate.github.io/docs/#/guide/05_using-translate-directive)
>  Using translate filter is great, but using translate directive is better. It turned out that having too many filters in a view sets up too many watch expressions, which is why angular-translate also provides a directive to translate your contents in view layer.

client/src/partials/fee_center/feeCenter.modal.html
```diff
@@ -0,0 +1,96 @@
...
+
+    <div class="form-group"
+         ng-class="{ 'has-error' : FeeCenterForm.$submitted && FeeCenterForm.projects.$invalid }">
+
+      <label class="control-label">{{ "FORM.LABELS.PROJECT" | translate }}</label>
+
+      <ui-select
+        name="projects"
+        ng-model="FeeCenterModalCtrl.feeCenter.project_id"
+        theme="bootstrap"
```
jniles, line 32: Just as a note: the `bootstrap` theme is set globally in `app.js`, there is no need to include it in `ui-select` components.

client/src/partials/fee_center/feeCenter.modal.html
```diff
@@ -0,0 +1,96 @@
...
+  <div class="modal-footer">
+    <p id="feeCenter-same" ng-if="FeeCenterForm.$submitted && FeeCenterForm.$pristine && !FeeCenterModalCtrl.isCreating" class="text-warning">
+      <i class="fa fa-warning"></i> {{ "FEE_CENTER.RECORD_SAME" | translate }}
+    </p>
+    <p ng-if="FeeCenterForm.$submitted && FeeCenterForm.$invalid" class="text-danger">
+      <i class="fa fa-exclamation-triangle"></i> {{ "FORM.ERRORS.RECORD_ERROR" | translate }}
+    </p>
+
+    <button id="feeCenter-cancel" type="button" class="btn btn-default" ng-click="FeeCenterModalCtrl.closeModal()">
+      <i class="fa fa-ban"></i> {{ 'FORM.BUTTONS.CANCEL' | translate }}
```
jniles, line 89: Most of the other forms/modals in the application do not use icons on the buttons.  It would be best if we picked one style.  If you like icons on buttons, make an issue for discussion and we can standardize our practices.

DedrickEnc:
Now the grid has a new column for showing the fee center value like this : 

value in grid

client/src/partials/fee_center/feeCenter.modal.js
```diff
@@ -0,0 +1,59 @@
...
+        vm.feeCenter = fc;
+      })
+      .catch(Notify.handleError);
+  }
+  // submit the data to the server from all two forms (update, create)
+  function submit(feeCenterForm) {
+    var promise;
+
+    if (feeCenterForm.$invalid) { return; }
+    if (!feeCenterForm.$dirty) { return; }
```
jniles, line 37: Just so you know, this is equivalent to `if (feeCenterForm.$pristine) { return; }` which is slightly easier to read.  That is why angular gives you both `$dirty` and `$pristine` values.

client/src/partials/fee_center/feeCenter.modal.js
```diff
@@ -0,0 +1,59 @@
...
+      .then(function () {
+        var translateKey = (vm.isCreating) ?  'FEE_CENTER.CREATED' : 'FEE_CENTER.UPDATED';
+        Notify.success(translateKey);
+        $state.go('feeCenter.list', null, {reload : true});
+      })
+      .catch(Notify.handleError);
+  }
+
+  function closeModal () {
+    $state.transitionTo('feeCenter.list');
```
jniles, line 51: `$state.transitionTo()` is a pretty low-level method of `ui-router`.  [The wiki recommends $state.go() for transitions](http://stackoverflow.com/questions/21105528/difference-between-state-transitionto-and-state-go-in-angularjs).

client/src/partials/fee_center/fee_center.html
```diff
@@ -0,0 +1,30 @@
+<div class="flex-header">
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static">{{ "TREE.FINANCE" | translate }}</li>
```
jniles, line 4: These should use the `translate` directive.

client/src/partials/fee_center/fee_center.js
```diff
@@ -0,0 +1,93 @@
...
+FeeCenterController.$inject = ['FeeCenterService', 'NotifyService','$state', 'SessionService'];
+
+/** Transaction Type Controller  */
+function FeeCenterController(FeeCenterService, Notify, $state, Session) {
+  var vm = this;
+
+  vm.enterprise = Session.enterprise;
+
+  // edit button template
+  var actionTemplate = '<div class="ui-grid-cell-contents">' +
```
jniles, line 17: It is best to move these into their own HTML templates.  That way they can be cached, gzipped, and the JS minifier doesn't have to deal with these long strings.

client/src/partials/fee_center/fee_center.js
```diff
@@ -0,0 +1,93 @@
...
+    '<a href="" id="edit-link" title="{{ \'FORM.LABELS.EDIT\' | translate }}" ng-click="grid.appScope.edit(row.entity)"> ' +
+    '<i class="fa fa-edit"></i>{{ "FORM.LABELS.EDIT" | translate }}</a>|' +
+    '<a ng-if="row.entity.is_principal !==1" href="" title="{{ \'FORM.LABELS.ASSIGN\' | translate }}" ng-click="grid.appScope.editFeeCenter(row.entity)"> ' +
+    '<i class="fa fa-share-alt"></i>{{ "FORM.LABELS.ASSIGN" | translate }}</a>' +
+    '</div>';
+
+  var infoTemplate = '<div class="ui-grid-cell-contents">' +
+    '<a href="" uib-popover="{{grid.appScope.showNote(row.entity.note)}}" ' +
```
jniles, line 25: Why not ...
```html
<a href="" uib-popover="{{row.entity.note}}">
```
?

client/src/partials/templates/grid/feeCenterType.tmpl.html
```diff
@@ -0,0 +1,5 @@
+<div class="ui-grid-cell-contents">
+  <span ng-class="{'label label-info': row.entity.is_cost === 1, 'label label-warning': row.entity.is_cost !== 1}">
```
jniles, line 2: You could move the `label` class out of this `ng-class` to make it read better.

client/src/partials/templates/grid/fee_center_value.cell.html
```diff
@@ -0,0 +1,7 @@
+<div class="ui-grid-cell-contents"  title="TOOLTIP" ng-if="!row.groupHeader">
+  {{ row.entity.value | currency:grid.appScope.enterprise.currency_id}}
+</div>
+
+<div class="ui-grid-cell-contents" title="TOOLTIP" ng-if="row.groupHeader">
```
jniles, line 5: These is no need to separate these two out -- you aren't even using `ui-grid-grouping`.

client/src/partials/templates/grid/fee_center_value.cell.html
```diff
@@ -0,0 +1,7 @@
+<div class="ui-grid-cell-contents"  title="TOOLTIP" ng-if="!row.groupHeader">
```
jniles, line 1: You actually don't need to have a separate template for this. Just amend your gridOptions to be
```js
{ field : 'value', displayName : 'FORM.LABELS.VALUE',
        cellFilter: 'currency:' + Session.enterprise.currency_id
}
```
Since all you are doing is applying a filter, you can build it right into to `ui-grid` options!

server/controllers/admin/services.js
```diff
@@ -29,17 +29,18 @@ const Topic = require('../../lib/topic');
...
-      JOIN enterprise AS e ON s.enterprise_id = e.id
-      LEFT JOIN cost_center AS cc ON s.cost_center_id = cc.id
-      LEFT JOIN profit_center AS pc ON s.profit_center_id = pc.id`;
+      SELECT 
+s.id, s.name, s.enterprise_id, s.cc_id, s.pc_id, e.name AS enterprise_name, 
+e.abbr, cc.id AS cc_id, cc.cost_center_name, pc.id AS pc_id, pc.profit_center_name 
+
+FROM service AS s
+JOIN enterprise AS e ON s.enterprise_id = e.id
+LEFT JOIN (SELECT fee_center.label AS cost_center_name, id FROM fee_center WHERE fee_center.is_cost = 1) AS cc ON s.cc_id = cc.id
```
jniles, line 42: Can't you just LEFT JOIN on fee_center as cc?

server/controllers/finance/accounts/index.js
```diff
@@ -244,14 +242,12 @@ function getBalance(req, res, next) {
...
       a.classe, a.is_asset, a.reference_id, a.is_brut_link, a.is_charge,
       a.number, a.label, a.parent, a.type_id, a.is_title, at.type,
-      at.translation_key, cc.text AS cost_center_text, pc.text AS profit_center_text
-    FROM account AS a JOIN account_type AS at ON a.type_id = at.id
-    LEFT JOIN cost_center AS cc ON a.cc_id = cc.id
-    LEFT JOIN profit_center AS pc ON a.pc_id = pc.id
-    `;
+      at.translation_key, fc.label AS fee_center_text 
+      FROM account AS a JOIN account_type AS at ON a.type_id = at.id
+    LEFT JOIN fee_center AS fc ON a.fc_id = fc.id`;
```
jniles, line 250: :+1:

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,285 @@
...
+**/
+
+
+var db = require('../../lib/db');
+var NotFound = require('../../lib/errors/NotFound');
+
+/**
+* Returns an array of fee centers
+*
+* @param {object} req The express request object
```
jniles, line 19: At this point, I think that all developers know how expressjs works.  There is no need to document these parameters.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,285 @@
...
+* @param {object} res The express response object
+* @param {object} next The express object to pass the control to the next middleware
+*
+* @example
+* // GET /fee_centers : Get list of fee_centers
+* var feeCenters = require('finance/fee_centers');
```
jniles, line 25: This file is actually `finance/feeCenter`.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,285 @@
...
+					        general_ledger AS gl 
+				        JOIN
+					        fee_center AS f ON gl.fc_id = f.id 
+			        ) AS t
+	        ) AS tt ON tt.id = fc.id
+      `;
+  }
+
+
+  if(req.query.available === '1') {
```
jniles, line 63: What does this parameter do?  What is an available fee center?

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,285 @@
...
+    filter += ' fc.id NOT IN (SELECT s.cc_id FROM service AS s WHERE NOT ISNULL(s.cc_id)) AND fc.id NOT IN (SELECT s.pc_id FROM service AS s WHERE NOT ISNULL(s.pc_id)) ';
+  }
+
+  if(req.query.is_cost){
+    if(filter !== ''){
+      filter += ' AND ';
+    }
+    filter += `fc.is_cost = ${req.query.is_cost} `;
```
jniles, line 71: What happens if `req.query.is_cost` is `(DELETE FROM general_ledger WHERE 1); -- ` ?

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  sql = filter === '' ? sql + ordering : sql + 'WHERE ' + filter + ordering;
+
+  db.exec(sql)
+  .then(function (rows) {
+    res.status(200).json(rows);
+  })
+  .catch(function (err) {
+    console.log('******* erreur', err);
```
jniles, line 88: This should have proper error handling.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  if(req.query.is_principal){
+    if(filter !== ''){
+      filter += ' AND ';
+    }
+    filter += `fc.is_principal = ${req.query.is_cost} `;
+  }
+
+  sql = filter === '' ? sql + ordering : sql + 'WHERE ' + filter + ordering;
```
jniles, line 81: If you used `db.exec(sql, [params])`, you could write your code so that you didn't have to check if `filter === ''` all the time.

client/src/partials/fee_center/fee_center.js
```diff
@@ -0,0 +1,93 @@
+'use strict';
```
jniles, line 1: There is no need to `use strict` on the client.  Gulp will insert it for us.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,285 @@
...
+function update (req, res, next) {
+  'use strict';
+
+  var queryData = req.body;
+  var feeCenterId = req.params.id;
+  var updateFeeCenterQuery = 'UPDATE fee_center SET ? WHERE id = ?';
+
+  delete queryData.id;
+
+  lookupFeeCenter(feeCenterId)
```
jniles, line 144: This used to be best practices until we created our `errors/*.js` libraries for error handling.  You can now just do a UPDATE and then throw `NotFound()` if the `affectedRows` is 0 --- less database queries = easier for raspberry pi!

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,285 @@
...
+
+function lookupFeeCenter(id) {
+  'use strict';
+
+  var sql =
+    'SELECT fc.id, fc.label, fc.is_cost, fc.note, fc.is_principal, fc.project_id FROM fee_center AS fc WHERE fc.id = ?';
+
+  return db.exec(sql, id)
+    .then(function (rows) {
+      if (rows.length === 0) {
```
jniles, line 226: You can use `db.one()` to handle this case automatically.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,285 @@
...
+      sql =
+        'SELECT ac.id FROM account AS ac WHERE ac.fc_id = ? AND ac.is_title=0';
+
+      return db.exec(sql, req.params.id);
+    })
+    .then(function (rows){
+
+      if (rows.length > 0) {
+        rows = rows.map(function (row) { return row.id;});
+        optionalCondition = ' OR gl.account_id IN (' + rows.join(',') + ')';
```
jniles, line 261: It would be much, much better to use `db.exec()`'s parameters instead of doing string concatenation in JS.

server/controllers/finance/generalLedger/index.js
```diff
@@ -46,5 +45,5 @@ function list(req, res, next) {
     .then(function (rows) {
         res.status(200).json(rows);
     })
-    .catch(next);
+    .catch(function (err){console.log(err);});
```
jniles, line 48: This should be handled.

server/models/schema.sql
```diff
@@ -317,51 +316,51 @@ CREATE TABLE `consumption_service` (
...
-
-CREATE TABLE `cost_center` (
-  `project_id` smallint(5) unsigned NOT NULL,
+CREATE TABLE `fee_center` (
   `id` smallint(6) NOT NULL AUTO_INCREMENT,
-  `text` varchar(100) NOT NULL,
-  `note` text,
+  `project_id` smallint(5) unsigned NOT NULL,
+  `label` varchar(100) NOT NULL,
+  `is_cost` tinyint(1) DEFAULT 0,
```
jniles, line 325: Since you do a lot of queries on `is_cost`, as a performance improvement, you could put an INDEX on it.

server/models/schema.sql
```diff
@@ -317,51 +316,51 @@ CREATE TABLE `consumption_service` (
...
-CREATE TABLE `cost_center_assignation` (
-  `project_id` smallint(5) unsigned NOT NULL,
+CREATE TABLE `cost_assignation` (
   `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
-  `auxi_cc_id` smallint(6) NOT NULL,
+  `auxiliary_cost_id` smallint(6) NOT NULL,
   `cost` float DEFAULT 0,
-  `date` date DEFAULT NULL,
+  `period_id` mediumint(8) UNSIGNED NOT NULL,
+  `auxi_cc_id` smallint(6) NOT NULL,
```
jniles, line 341: What is `auxi_cc_id`?  Can it be removed since we have `auxiliary_cost_id`?

test/end-to-end/fee_center/feeCenter.page.js
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * This class is represents a fee center  page in term of structure and
+ * behaviour so it is a fee center page object
+ **/
+
+function FeeCenterPage() {
+  'use strict';
+
+  const page = this;
```
jniles, line 12: :+1:

test/end-to-end/fee_center/feeCenter.page.js
```diff
@@ -0,0 +1,77 @@
...
+   **/
+  function createFeeCenter () {
+    return addFeeCenterButon.click();
+  }
+
+  /**
+   * simulate a click to a link tailed to the fee center
+   *  listed in the grid to show the dialog for editing
+   **/
+  function editFeeCenter(n) {
```
jniles, line 36: This is a nice wrapper. :+1:

test/end-to-end/fee_center/feeCenterCU.page.js
```diff
@@ -0,0 +1,126 @@
+
+/* jshint expr:true */
+/* global element, by, browser */
+
+/**
+ * This class is represents a fee center creation page in terms of structure and
+ * behaviour so it is a fee center creation page object
+ **/
+function CreateUpdateFeeCenterPage() {
```
jniles, line 9: Many of these methods are already contained in helper libraries such as `FormUtils`.  Is there a reason to duplicate the effort here?

test/integration/feeCenter.js
```diff
@@ -0,0 +1,160 @@
+/* global expect, chai, agent */
+/* jshint expr : true */
+
+const helpers = require('./helpers');
+
+/*
+ * @todo - there are some tests missing:
```
jniles, line 7: Good comment :+1:

test/integration/feeCenter.js
```diff
@@ -0,0 +1,160 @@
...
+
+  it('GET /fee_centers returns a list of fee centers', function () {
+    return agent.get('/fee_centers')
+      .then(function (res) {
+        helpers.api.listed(res, 7);
+      })
+      .catch(helpers.handler);
+  });
+
+  /* @todo - make this route ?detailed=1 to conform to standards */
```
jniles, line 37: Good comment :+1:

test/integration/feeCenter.js
```diff
@@ -0,0 +1,160 @@
...
+        helpers.api.errored(res, 404);
+      })
+      .catch(helpers.handler);
+  });
+
+
+
+  /* @todo - should this return a simple number? */
+  it('GET /fee_centers/:id/value returns the value of a provided fee center', function () {
+    return agent.get('/fee_centers/:id/value'.replace(':id', FETCHABLE_FEE_CENTER_ID))
```
jniles, line 102: Since we are using Node v4+, you can just use string substitution:
```js
return agent.get(`/fee_centers/${FETCHABLE_FEE_CENTER_ID}/value`):
```

test/integration/feeCenter.js
```diff
@@ -0,0 +1,160 @@
...
+      .then(function (res) {
+        expect(res).to.have.status(200);
+        expect(res).to.be.json;
+        expect(res.body).to.not.be.empty;
+        expect(res.body).to.have.all.keys('value');
+        expect(res.body.value).to.satisfy(function (value) { return value >= 0;});
```
jniles, line 108: This is a nice test :+1:

jniles (changes requested):
@DedrickEnc, these are a lot of changes that represent a lot of work.

Many of the changes are based on old code that don't abide by best practices.  In general, good practices to follow are:
 1. Always pass parameters to the database by escaping them to prevent SQL injection attacks.  MySQL also handles translation of JS's BLOB --> BINARY, new DATE() ---> DATETIME, and other nice features.  You should always compose your SQL with `?` question mark syntax and pass the parameters as an array to `db.exec()`.
 2. Using tested libraries is almost always better than writing your own.  This applies to the `PrototypeApiService` which has Karma testes written for it, and `FormUtils`, which is used in almost every set of `e2e` tests in the application.
 3. Always use the `translate` directive where possible.
 4. Try to move HTML templates out of Javascript.  They cannot be cached efficiently and only cause the browser to more work.

Once these changes are made, the code will be much easier to audit for correctness.

jniles:
@DedrickEnc  does the PR supersede https://github.com/IMA-WorldHealth/bhima-2.X/pull/1101?

test/end-to-end/fee_center/feeCenterCU.page.js
DedrickEnc: I followed the page object as detailed from the official protractor tutorial, and now my buttons, radio button, and text field are privates members of my page. this let me represent a structure of my page by reading the code I can guess how many element I have on the page and I many actions or methods (interface) the page gives me.

jniles: In my view, the page object methods works best on really complicated pages - such as Complex Vouchers, Patient Invoice, Posting Journal...

Your current page is a simple form - in this case, it triples the amount of code needed to maintain and upgrade with marginal benefit.

PR #1126, opened 2017-02-06
refactor(filters) Server side `FilterParser` library

sfount:
This commit introduces a server side library for uniformly handling filters. Currently we have many server controllers that uniquely implement complex logic for parsing and compiling filters. The problem we currently have is that when our standards are updated on one filter the benefits or improvements are not applied to all of the modules. This also greatly increases maintenance complexity as you have to remember or learn how the logic for each different filter works. 

This library should standardise the way that filters are implemented and make it simple to add filters for the majority of queries or columns required when filtering data sets on the server. 

With `FilterParser` the patient entity search logic can be reduced to the following lines of code: 
```js 
// options is passed directly from the client's query
let filters = new FilterParser(options, { tableAlias : 'q' });

filters.fullText('display_name');
filters.dateFrom('dateRegistrationFrom', 'registration_date');
filters.dateTo('dateRegistrationTo', 'registration_date');
filters.dateFrom('dateBirthFrom', 'dob');
filters.dateTo('dateBirthTo', 'dob');
 
// more complex custom logic can be handled uniformly with the custom method
let patientGroupStatement = '(SELECT COUNT(uuid) FROM assignation_patient where patient_uuid = q.uuid AND patient_group_uuid = ?) = 1';
filters.custom('patient_group_uuid', patientGroupStatement);

let query = filters.applyQuery(sql);
let parameters = filters.parameters();
return db.exec(query, parameters);
```

Each of the filter methods shown here takes an attribute (`filterKey`) that can be passed through the query string from the client. It then uses this value to create the filter query for the sql passed in. 

The current utility methods implemented are: 
`fullText` - search for a string at the start, middle or end of any text column 
`dateFrom` - reject all data points before this date 
`dateTo` - reject all data points after this date 
`equals` - a way of overriding the default equals behaviour, with this method you can also provide column and table aliases 
`reversed` - a proof of concept toggle filter, show reversed records or exclude them

Any filter keys that are passed and don't have a specific rule set will attempt to do a direct match between the filter key (attribute name) and the value provided. This allows complex implicit filtering; for example the filtering for cash records only has to define two rules: 

```js
filters.dateFrom('dateFrom', 'date');
filters.dateTo('dateTo', 'date');
```

Everything else is matched on `attributeName=attributeValue`. 

There are still many search controllers on the server that could implement this behaviour; this is a proposal/ first implementation for standardising some of our more complex logic.


jniles:
LGTM :+1:

PR #1142, opened 2017-02-07
fix(creditNote): reverse all rows in a reversal

jniles:
This is a critical bug fix.

This commit changes the `ReverseTransaction()` method to using UNION ALL instead of a simple UNION to avoid only reversing unique rows.  The lack of the `ALL` term has resulted in multiple unbalanced credit notes in production.

**Description**
The problem was found while running the Trial Balance on ~5,000 unposted transactions containing over 100 credit notes.  Approximately 25% of the credit notes were unbalanced.  The problem was reproducible by removing the offending credit note and re-creating it, resulting in vouchers that looked like this one:

creditnoteunbalancedvouchers
_Fig 1: Unbalanced Credit Note Voucher_

Since the problem was reproducible for those vouchers, it appeared to be data-driven.  A quick audit of the `ReverseTransaction()` stored procedure revealed that it was using [UNION instead of UNION ALL](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/procedures.sql#L1141) to combine the results of the posting_journal and general_ledger.  [UNION removes identical results](http://stackoverflow.com/questions/49925/what-is-the-difference-between-union-and-union-all).

**FIX**
This problem could be solved in two ways:
 1. Include a unique value in the SELECT statement OR
 2. Use UNION ALL

This PR favors (2) because it requires that the database do less work.

------
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/client/src/js/services/VoucherService.js
+++ b/client/src/js/services/VoucherService.js
@@ -77,7 +77,8 @@ function VoucherService(Api, $http, util, TransactionTypeStore) {
    * This method facilitate annulling a transaction,
    * bhima should automatically be able to reverse
    * any transaction in the posting_journal by creating a
-   * new transaction that is an exact duplicate of the original transaction with sign minous.
+   * new transaction that is an exact duplicate of the original transaction with the
+   * debits and credits switched.
    */
   function reverse(creditNote) {
     return $http.post(baseUrl.concat(creditNote.uuid, '/reverse'), creditNote)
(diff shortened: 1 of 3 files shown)


sfount (approved):
Great analysis! The write up was very useful in understanding the context. 

I agree with the fix; looks good to me!

PR #1164, opened 2017-02-10
Using icon fa-filter instead fa-search

lomamech:
- Using icon fa-filter instead fa-search 
- Using btn-info when the filtering is activate


jniles:
LGTM. :+1:

PR #1133, opened 2017-02-06
Vanga desactivate

DedrickEnc:
This PR gives the possibility of locking a user. It closes https://github.com/Vanga-Hospital/bhima-2.X/issues/260.

----
Thank you for contributing!

Before submitting this pull request, please verify that you have:
 - [x] Run your code through JSHint.  [Check out our styleguide](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/docs/STYLEGUIDE.md).
 - [x] Run integration tests.
 - [x] Run end-to-end tests.
 - [x] Accurately described the changes your are making in this pull request.

For a more detailed checklist, [see the official review checklist](https://docs.google.com/document/d/1nupLVLRXgSZJQo_acLgrwvPnN8RukfSiwRhSToj81uU/pub) that this PR will be evaluated against.

Ensuring that the above checkboxes are completed will help speed the review process and help build a stronger application.  Thanks!

PR #1175, opened 2017-02-13
Login error messages

lomamech:
Login error messages  …
- Update auth.js to catch No Permissions and No Project
- Add two lib errors for No Permission and No Project

Update test integration for recognized user without a project


server/controllers/auth.js
```diff
@@ -52,12 +55,48 @@ function login(req, res, next) {
...
+    WHERE user.username = ? AND user.password = PASSWORD(?)`;
+
+  let sqlPermission = `
+    SELECT permission.id FROM permission
+    JOIN user ON user.id = permission.user_id
+    WHERE user.username = ? AND user.password = PASSWORD(?)`;  
+
+  transaction = db.transaction();
+
+  transaction
```
sfount: MySQL transactions are very useful for when we need to roll back changes, however they are expensive to perform as they lock MySQL tables during execution. 

As all of these operations only involve reads I recommend that a transaction is not used. 

Instead this section could read like this: 

```js
db.exec(sqlTestUser) 
  .then((result) => { 
    // comment describing was result was
   if (result.length) { throw new BadRequest() }
   return db.exec(sqlTestProject);
  })
  .then // ...
```

server/lib/errors/NoPermission.js
```diff
@@ -0,0 +1,52 @@
+const util = require('util');
+
+/**
+ * @class NoPermission
```
sfount: The `lib/BadRequest` class allows you to specify a key that will be sent to the user. In order to have less server code to maintain I suggest that these cases are handled as `BadRequest`s with unique error codes passed in the key field.

sfount (dismissed):
The functionality of this pull request looks exactly correct! I have made some comments about small changes that would help the maintainability of the codebase.

sfount:
I notice the `Unautharized` does not allow you to set a translation key for the user to see. This could be updated and used along with `BadRequest`

jniles:
`Unauthorized` gained that functionality in an [unmerged PR](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1133).  Should I merge that PR so it can be used here?

server/lib/errors/Unauthorized.js
```diff
@@ -29,7 +29,9 @@ const util = require('util');
  *
  * @requires util
  */
-function Unauthorized(description) {
+
+function Unauthorized(description, key) {
+  'use strict';
```
jniles, line 34: There is not need to add `use strict` anymore.  We are forcing the server to run in strict mode via an npm package.  It also shouldn't be necessary if you are running the latest NodeJS.

jniles (approved):
LGTM.  This functionality is important as users are getting confused when they are given generic error messages without any direction about what to do next.

:+1:

PR #1178, opened 2017-02-14
feature(report): Open Debtors

jniles:
This PR implements the Open Debtors Report.  It will close https://github.com/Vanga-Hospital/bhima-2.X/issues/188.

The open debtors report lists all debtors that have an outstanding debt, the date of their last invoice, the date of their last payment, and the totality of their debt.  It provides several options for sorting the data that may be useful to an accountant - both sorting by date and by amount of debt.

Some things to add in the future:
 1. Filter out "volatile" Posting Journal records and only use the General Ledger
 2. Sorting by debtor name

See below:
opendebtors
_Fig 1: Open Debtors Report_

----
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


client/src/partials/reports/modals/openDebtors.modal.html
```diff
@@ -0,0 +1,55 @@
+<form name="ConfigForm" bh-submit="ReportConfigCtrl.generate(ConfigForm)" bh-form-defaults novalidate autocomplete="none">
+
+  <div class="modal-header">
+    <ol class="headercrumb">
+      <li class="static">{{ ::ReportConfigCtrl.report.title_key | translate }}</li>
```
sfount: I'm not sure if this case cannot use the translate directive - if it is possible it would good to bring it in line with standards. 

Potentially this would work:
```html
<li class="static" translate>{{::ReportConfigCtrl.report.title_key}}</li> 
```

This should not block this pull request.

sfount (approved):
Code LGTM :+1:. This is a very clearly implemented report. 

In the future would it be useful to show the last invoice and cash payment reference next to the dates? You could look these values up quite easily using the registries so I don't think it's necessary.

jniles:
@sfount, thanks for your review.

The changes have been made and the conflicts fixed.

In regards to showing the latest invoice and cash payment, I don't think it is of that great use.  The dates are important to understand if the debtor has been barred from receiving care (hence the last invoice date) and how long they have gone without paying (hence the last payment date).  We actually probably need something like this report in the aged format as well to show debt over time.

sfount:
Changes look great. That makes complete sense, this is less about individual documents and more about debt and timing. 

LGTM.

PR #1114, opened 2017-01-24
Purchase order registry

mbayopanda:
This PR add an implementation of the purchase order registry and make some fixing.

- [x] Fix purchase order receipt
- [x] Add purchase order registry 
- [x] Add status for purchase order 
- [x] E2E test for the purchase order registry

PR #1186, opened 2017-02-15
feature: purchase order registry

jniles:
This commit leverages FilterParser for the purchase registry.  It also
implement integration tests for the `/search` route.  Finally, it
ensures that purchase orders abides by the new identifiers syntax.

Supersedes #1114.  Closes #1114.


server/controllers/finance/cash.js
```diff
@@ -273,7 +273,7 @@ function reference(req, res, next) {
       FROM cash JOIN project ON cash.project_id = project.id
     )c WHERE c.reference = ?;`;
 
-  db.one(sql, [ ref ])
+  db.one(sql, [ ref , 'cash'])
```
mbayopanda: What is the reason of `cash` in `db.one(sql, [ ref , 'cash'])` ?

mbayopanda (commented):
LGTM

Fix just travis errors and update the branch

server/controllers/finance/cash.js
jniles: Typo - nice catch.  It should read `db.one(sql, [ref], ref, 'cash');`.  That way it will print out a nice error message if this fails.

Although .. I'm actually not sure this route is ever used.

Issue #1076, opened 2016-12-15
(bug) Billing services aren't displayed correctly in invoices 


sfount:
We currently support both the application of billing services and subsidies to invoices. There are a number of places in invoices where we should verify that these are displayed correctly. 

Currently billing services aren't necesarily displayed in the correct currency and a number of the values are always 0.

![invoice_bug](https://cloud.githubusercontent.com/assets/2844572/21216798/15c7bcda-c2a9-11e6-89ae-2b7c3ee3bb37.PNG)

jniles:
To tackle at the same time: we should refer to the total of the items as "Subtotal", and then the grand total as "Total".  I've seen this on some invoicing software recently and thought it was a good idea.

lomamech:
#1196

PR #1196, opened 2017-02-16
Fix Display in Invoices

lomamech:
Fix Display in Invoices 
- Update the request in PatientInvoice invoiceBillingQuery adding total of invoice
- Update receip.handlebars

#1076


jniles (approved):
LGTM!  Great change.

Issue #1193, opened 2017-02-16
(feature request) Print Barcodes on A4 Receipts


jniles:
Just because a barcode is ideal for a thermal receipt doesn't mean we should restrict it to only thermal receipts.  Putting them on the larger A4 receipts gives us the possibility to use barcode scanners to look up any receipt at any time.  It also means that users without a thermal printer will be able to benefit from a barcode scanner immediately, without accepting the limited durability of the thermal receipt.

lomamech:
#1201

PR #1201, opened 2017-02-16
Update receipt A4 for cash paiement, invoices and voucher, Add proper…

lomamech:
Update receipt A4 for cash paiement, invoices and voucher, Add property Barcode

#1193

### DIFF
--- a/server/controllers/finance/reports/cash/receipt.handlebars
+++ b/server/controllers/finance/reports/cash/receipt.handlebars
@@ -1,6 +1,4 @@
 {{> head title="CASH.TITLE" }}
-<head>
-<script src="{{absolutePath}}/vendor/JsBarcode.all.min.js"></script>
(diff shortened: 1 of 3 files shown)


server/controllers/finance/reports/cash/receipt.handlebars
```diff
@@ -1,5 +1,8 @@
 {{> head title="CASH.TITLE" }}
-
+<head>
+  <meta charset="utf-8" />
```
jniles: This will result in two `<head>` tags.  It would be better to put this code in the `head.handlebars` partial that is added automatically.

jniles (commented):
You can reduce the amount of code here by adding the code you repeated to the `head.handlebars` partial.

jniles:
LGTM.

PR #1203, opened 2017-02-16
fix account report

mbayopanda:
This PR fix : 
- display correct transaction date

### DIFF
--- a/server/controllers/finance/reports/reportAccounts/index.js
+++ b/server/controllers/finance/reports/reportAccounts/index.js
@@ -61,10 +61,10 @@ function document(req, res, next) {
  * This feature select all transactions for a specific account
 */
 function queryReportAccount(accountId, source) {
-  source = parseInt(source);
+  const sourceId = parseInt(source, 10);
   let sql;
 
-  if (source === 1) {
+  if (sourceId === 1) {
     sql = `
       SELECT general_ledger.trans_id, BUID(general_ledger.entity_uuid) AS entity_uuid, general_ledger.description,
       general_ledger.trans_date, general_ledger.debit_equiv as debit, general_ledger.credit_equiv as credit
@@ -73,7 +73,7 @@ function queryReportAccount(accountId, source) {
       GROUP BY general_ledger.trans_id
       ORDER BY general_ledger.trans_date ASC;
     `;
-  } else if (source === 2) {
+  } else if (sourceId === 2) {
     sql = `
       SELECT posting_journal.trans_id, BUID(posting_journal.entity_uuid) AS entity_uuid, posting_journal.description,
       posting_journal.trans_date, posting_journal.debit_equiv as debit, posting_journal.credit_equiv as credit
@@ -82,7 +82,7 @@ function queryReportAccount(accountId, source) {
       GROUP BY posting_journal.trans_id
       ORDER BY posting_journal.trans_date ASC;
     `;
-  } else if (source === 3) {
+  } else if (sourceId === 3) {
     sql = `
       SELECT transaction.trans_id, transaction.entity_uuid, transaction.description, 
         transaction.trans_date, sum(transaction.credit_equiv) as credit, sum(transaction.debit_equiv) as debit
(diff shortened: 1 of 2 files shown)


jniles (approved):
LGTM.  Can you also migrate this to SQL instead of JS while you are at it?  Or should I merge this PR first?

mbayopanda:
I would like to migrate this to SQL but i have to duplicate sql code again.
Could you guide me if possible.

Thanks

2017-02-16 7:30 GMT-08:00 Jonathan Niles <notifications@github.com>:

server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -63,10 +61,10 @@ function document(req, res, next) {
  * This feature select all transactions for a specific account
 */
 function queryReportAccount(accountId, source) {
-  source = parseInt(source);
+  const sourceId = parseInt(source, 10);
   let sql;
 
-  if (source === 1) {
+  if (sourceId === 1) {
```
jniles: Here's an idea for removing the duplicate SQL:

```js
const sourceMap = {
  1 : 'general_ledger',
  2 : 'posting_journal',
  3 : 'combined_ledger'
};

// get the table name
const tableName = sourceMap[sourceId];

const sql = `
    SELECT trans_id, BUID(entity_uuid) AS entity_uuid, description, trans_date, debit_equiv as debit, credit_equiv as credit
    FROM ${tableName}
    WHERE account_id = ?
    GROUP BY record_uuid
    ORDER BY trans_date ASC;
`
```

Does that give you the results you need?

server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -22,35 +22,33 @@ function document(req, res, next) {
...
+      const title = {
         accountNumber : params.account_number,
-        accountLabel : params.account_label,
-        source       : params.sourceLabel
+        accountLabel  : params.account_label,
+        source        : params.sourceLabel,
       };
 
-      accounts.forEach(function (account) {
+      accounts.forEach((account) => {
```
jniles: This sum should be done in SQL as well.

jniles (commented):
@mbayopanda, I've left some suggestions.  Let me know if they work or do not work.  They were quick!

server/controllers/finance/reports/reportAccounts/index.js
mbayopanda: 👍 thanks

mbayopanda:

server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -1,8 +1,14 @@
...
-const TEMPLATE      = './server/controllers/finance/reports/reportAccounts/report.handlebars';
-const BadRequest    = require('../../../../lib/errors/BadRequest');
+
+const TEMPLATE = './server/controllers/finance/reports/reportAccounts/report.handlebars';
+const BadRequest = require('../../../../lib/errors/BadRequest');
+
+/**
+ * global constants
+ */
+const sourceMap = { 1: 'general_ledger', 2: 'posting_journal', 3: 'combined_ledger' };
```
jniles, line 11: Nice!

server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -63,47 +59,41 @@ function document(req, res, next) {
...
-      ) as transaction
-      GROUP BY transaction.trans_id
-      ORDER BY transaction.trans_date ASC;
-    `;
-  }
 
-  return db.exec(sql, [accountId, accountId]);
+  const bundle = {};
+
+  return db.exec(sql, [accountId, accountId])
```
jniles, line 88: This code is correct. :+1:

Since the promises do not depend on one another, you can actually do both at the same time to save space and make the code (slightly) faster.

Like this:
```js
// in the query function
const params = [ accountId, accountId ];
return q.all([ db.exec(sql, params), db.exec(sqlAggrega, params) ]);

// in the main function
 promise.spread((accounts, sum)) => {
  _.extend(bundle, { accounts, sum });
  // ... etc ...
 });
```

The code will be faster because MySQL can process both queries in parallel instead of one at a time.

jniles (commented):
Looks awesome :+1:

Issue #1170, opened 2017-02-11
(feature request) Account Report Suggestions


jniles:
The Account Report presents a decent overview of the transactions involving a particular account.  Here are some improvements based on feedback from @imawh and personal observations.

- [ ] Move the date column to the far left
- [ ] Keep a running balance on the far right side.
- [ ] All currency amount should be right aligned (`text-right`).
- [ ] Include some indication of the date range spanned and number of transactions in the dataset.
- [ ] The account number and account label can be on the same line to save space.

For reference, see below the current rendering:

![2017-02-1111 33 09](https://cloud.githubusercontent.com/assets/896472/22853165/348dc1b8-f04f-11e6-87ae-5144be86b6f2.png)
_Report of Accounts_

sfount:
There was also an issue with this report always showing the date of the report itself in each of the date columns. This should be addressed in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1203/files

sfount:
To add to this: Patrick has been reviewing a number of the reporting features in 2.x and has noted that a date range would be useful in this report. To be able to select only the month of January or only the previous year for example. 

This feedback should be formally compiled however this is relevant to this issue. 

- [ ] Date range added to report generation allowing the selection of a start date and an end date. 
- [ ] ~~Running balance should show previous balance of account before the date from begins for context~~
- [ ] Running balance should show the balance ONLY for the period selected for the report (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1170#issuecomment-280620654)

patsonjunior:
In the context of a specific date or period, the current balance (running balance) should only present the amount in relation to the date or period considered for drawing the report.

lomamech:
#1211

PR #1211, opened 2017-02-17
Suggestion account report

lomamech:
- Update strucutere of reportAccounts
- Adding specific date for the running balance  

Partially addresses #1170.


server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -58,42 +70,78 @@ function document(req, res, next) {
...
+        FROM ${tableName}
+        WHERE account_id = ? AND trans_date >= DATE(?) AND trans_date <= DATE(?)
+        GROUP BY trans_id 
+        ORDER BY trans_date ASC
+      ) AS t 
+      `;
+
+    const bundle = {};
+
+    return db.exec(sql, [accountId, dateFrom, dateTo])
```
sfount: There should not be two different branches to reach `db.exec` here. Testing this logic would be very difficult. 

Can the additional trans_date logic be templated in if `dateFrom` and `dateTo` are defined? 

For example: 
```js
const params = [accountId];
const dateCondition = '';

if (dateFrom && dateTo) { 
  dateCondition = 'AND trans_date >= DATE(?) AND trans_date <= DATE(?)';
  params.push(dateFrom, dateTo);
}

const sql = `
  //...
  WHERE account_id = ? {dateCondition}
`;
```

This should be well commented and unfortunately leads to a confusing state but I think it would be preferable to a large branching structure.

client/src/partials/reports/modals/report_accounts.modal.html
```diff
@@ -17,6 +17,22 @@
       </div>
     </div>
 
+    <!-- Inclure Date interval  -->
+    <div class="checkbox">
+      <label class="control-label">
+        <input type="checkbox" ng-model="ReportConfigCtrl.dateInterval" ng-true-value="1" ng-false-value="0">
+        <strong>{{ "FORM.LABELS.INCLUDE_DATE_INTERVAL" | translate }}</strong>
```
sfount: This should use the translate component syntax. 

```html 
<strong translate>FORM.LABELS.INCLUDE_DATE_INTERVAL</strong>
```

server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -40,8 +45,15 @@ function document(req, res, next) {
     return next(e);
   }
 
-  return queryReportAccount(params.account_id, params.sourceId)
+  return queryReportAccount(params.account_id, params.sourceId, params.dateFrom, params.dateTo)
     .then((result) => {
+
+      result.accounts.forEach(function (account) {
+        cumulBalance += account.debit - account.credit;
```
sfount: As much as possible all calculations for balances/ debits and credits should be done in MySQL and not in Javascript. As referenced in other issues/ pull requests Javascript has many issues when it comes to its Maths library that could cause issues in the long run. 

In this case I believe you can calculate a running total in MySQL by using a variable. Something like: 

```sql 
SET @cumulative_balance := 0;
SELECT 
debit, 
credit, 
(@cumulative_balance := @cumulative_balance + (debit - credit)) as cumulative_balance
FROM #...
```

sfount (dismissed):
This pull request well addresses the issues in the feature request. 

I have made a few comments to make sure that these features are using the same standards as other server code.

jniles:
@sfount @lomamech, this report really needs to be cleaned up and used in production.  It allows us to figure out what the account balances will be.

Some more feedback:
 1. The account select should also use `bhAccountSelect`.
 2. If possible, it would be nice to see the opening balances via the `period_total` table.  But this can wait as an advanced feature.

client/src/partials/reports/modals/report_accounts.config.js
```diff
@@ -31,25 +32,32 @@ function ReportAccountsConfigController($state, ModalInstance, Accounts, Notify,
     {id: 3, label : 'FORM.LABELS.ALL'}
   ];
 
-  Accounts.read()
-    .then(function (accounts) {
-      vm.accounts = accounts;
-    })
-    .catch(Notify.handleError);
+  function onAccountSelect(account) {
```
jniles, line 35: :+1:

client/src/partials/reports/modals/report_accounts.modal.html
```diff
@@ -1,61 +1,59 @@
...
         <div ng-messages-include="partials/templates/messages.tmpl.html"></div>
       </div>
     </div>
 
-    <!-- account selection  -->
-    <div class="form-group"
-      ng-class="{ 'has-error' : ConfigForm.$submitted && ConfigForm.account.$invalid }">
+    <!-- Inclure Date interval  -->
+    <div class="checkbox">
       <label class="control-label">
```
jniles, line 22: You don't need the "control-label" class for checkboxes.

client/src/partials/reports/modals/report_accounts.modal.html
```diff
@@ -65,11 +63,11 @@
 
   <div class="modal-footer">
     <button type="button" class="btn btn-default" ng-click="ReportConfigCtrl.cancel()" data-method="cancel">
-      {{ "FORM.BUTTONS.CANCEL" | translate }}
+      <span translate> FORM.BUTTONS.CANCEL </span>
```
jniles, line 66: :+1:

server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -63,42 +65,57 @@ function document(req, res, next) {
...
+function getAccountTransactions(accountId, source, dateFrom, dateTo) {
   const sourceId = parseInt(source, 10);
 
   // get the table name
   const tableName = sourceMap[sourceId];
+  const params = [accountId];
+  let dateCondition = '';
+
+  if(dateFrom && dateTo){
+    dateCondition = `AND DATE(trans_date) BETWEEN ? AND ?`
```
jniles, line 77: I think this should be `AND DATE(trans_date) BETWEEN DATE(?) AND DATE(?)`.  We've had errors where these dates haven't been parsed correctly previously.

jniles (commented):
I've made a few comments, none of them blocking.

This will be my last day working with the AG, so I'm merging this to teach him on.

Issue #1097, opened 2017-01-10
Display information in the General Ledger in the correct way


mbayopanda:
The general ledger is a book which contains sold of each accounts. The current General Ledger in BHIMA display information such as another Journal, 

This issue is about to display in the General ledger information that the General Ledger must display.

Here is a proposition : 

The grid in the General ledger must display accounts and their sold : 

![only gl](https://cloud.githubusercontent.com/assets/5445251/21806770/23be169e-d6f0-11e6-9a56-49dd8b5760d6.png)

To get more information about the sold of an account, we have just to click on the account row 

![gl](https://cloud.githubusercontent.com/assets/5445251/21806973/3a5f1046-d6f1-11e6-91f9-28de0c9c06c8.png)

In this way the General Ledger is more useful for the user for getting the sold of an account in the enterprise in a given date.

jniles:
Very nice issue :+1:

I agree that we need a view of the data like this.  I also think the current General Ledger is a useful tool for finding transactions.  Should we replace the current General Ledger, or make them two different modules/reports?

mbayopanda:
Normally, the journal must always keep information of transactions but in the current state of BHIMA transactions are removed from the journal when they are posted in the General Ledger.

The Journal must keep all transactions made in the enterprise, and the General ledger must have all accounts used in the enterprise and their sold.

jniles:
That sounds like a _lot_ of data to have to ship to the client on every request...  It seems to me, technically, like we should separate the posted and unposted transactions.

Vanga has made 1715 lines in the Posting Journal in one week of operation, without even installing expenses and stock!

mbayopanda:
The Journal is a daily registry of transactions, so we can just display by default only transactions of the day in the Journal and load others with tools. 

The idea here is to have a Journal for an accountant not just a module because when a transaction is posted into the General ledger the accountant must have the possibility to see the proof in the Journal (also with real "Pieces justificatives")

jniles:
Hmm...  In my view, mixing transactions that an accountant is allowed to edit (unposted transaction) and transactions that they cannot (posted transaction) would be confusing.  I think we should keep them separate as much as we can.

I think loading only the transactions of the day is a good idea :+1:

I like the account view shown in your balsamiq mockups.  If we went from the Trial Balance to a view like that, it would be very nice.

jniles:
@mbayopanda, what is missing from this issue?

jniles:
Closing this as completed in #1212.

PR #1212, opened 2017-02-17
General ledger and Posted Journal

mbayopanda:
This PR is about : 
- The implementation of new general ledger ui 
- Move old general ledger interface to the new Posted Journal

issue #1097


server/controllers/finance/reports/generalLedger/accountSlip.handlebars
```diff
@@ -0,0 +1,68 @@
+{{> head title="TGENERAL_LEDGER.ACCOUNT_SLIP"}}
```
jniles: Is this "T" supposed to be here?

jniles:
Could you provide a few screenshots of this change?  I'm quite curious to see what it looks like.

mbayopanda:
1. Journals and General Ledger in menu tree : 

- Posting Journal : contains transactions which are not yet posted
- Posted Journal : contains transactions which are posted
- General Ledger : contains accounts and their solds

journal_menus

2. The General Ledger 

The General Ledger contains accounts and their sold, the user has the possibility to export data in CSV and PDF, and can print the general ledger.

general_ledger

There is a possibility to display the account slip "Extrait de compte" for each account 

general_ledger_extrait

Colors in `debtor_sold` column and in `creditor_sold` column are used to focus the attention of users to a debtor sold or a creditor sold at first sight

extrait

extrait_real

The Posted Journal is just for having the possibility for seeing transactions which are posted, it is the previous General Ledger

journal_poste

server/controllers/finance/generalLedger/index.js
```diff
@@ -42,8 +46,46 @@ function list(req, res, next) {
...
+      IF(aggregator.balance >= 0, aggregator.balance, 0) AS debtor_sold,
+      IF(aggregator.balance < 0, -1 * aggregator.balance, 0) AS creditor_sold
+    FROM (
+      SELECT SUM(gl.debit_equiv - gl.credit_equiv) AS balance, 
+        a.id, a.number, a.label 
+      FROM general_ledger AS gl 
+        JOIN account a ON a.id = gl.account_id
+        JOIN project pro ON pro.id = gl.project_id
+        JOIN period per ON per.id = gl.period_id
+        JOIN user u ON u.id = gl.user_id
```
jniles: Do you really need all these JOINs?  It looks like you actually only need the `general_ledger` and the `account` table.

server/controllers/finance/generalLedger/index.js
```diff
@@ -42,8 +46,46 @@ function list(req, res, next) {
...
+}
+
+/**
+ * @function getlistAccounts
+ * get list of accounts
+ */
+function getlistAccounts() {
+  const sql =
+    `SELECT aggregator.id, aggregator.number, aggregator.label,
+      IF(aggregator.balance >= 0, aggregator.balance, 0) AS debtor_sold,
```
jniles: In my opinion, it is better to use `balance` instead of `sold`, to keep it in the same language.  This won't block this PR though.

server/controllers/finance/reports/generalLedger/index.js
```diff
@@ -0,0 +1,84 @@
...
+
+  try {
+    report = new ReportManager(REPORT_TEMPLATE, req.session, options);
+  } catch (e) {
+    return next(e);
+  }
+
+  return GeneralLedger.getlistAccounts()
+    .then((rows) => {
+      data = { rows };
```
jniles, line 40: I don't think you need this extra variable `data`.  It's not a problem to have it defined, but I kept looking for where it was needed.  Removing it would clarify the code:
```js
.then((rows) => return report.render({ rows }))
```

server/controllers/finance/reports/reportAccounts/report.handlebars
```diff
@@ -19,7 +19,7 @@
         </tr>
       </thead>
       <tbody>
-        {{#each accounts}}
+        {{#each transactions}}
```
jniles, line 22: :+1: Useful refactor.

client/src/partials/general_ledger/general_ledger_accounts.html
```diff
@@ -0,0 +1,63 @@
+<div class="flex-header">
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static">{{ "TREE.FINANCE" | translate }}</li>
```
jniles: These should use the `translate` directive.

client/src/js/services/receipts/ReceiptModal.js
```diff
@@ -229,6 +230,37 @@ function ReceiptModal(Modal, Receipts) {
...
+  /**
+   * @method accountSlip
+   *
+   * @description
+   * Invokes a general ledger account slip
+   *
+   * @param {number} id - Target account ii
+   * @param {Boolean} notifyCreated   Defines if a success message should be shown for entity creation
+   */
+  function accountSlip(id, notifyCreated) {
```
jniles, line 242: It seem inappropriate that this is a receipt.  Do we not have a report modal?

jniles (changes requested):
There isn't anything major blocking here.

Two comments it would be nice fix:
 1. Remove unused JOINs -- I'm pretty sure MySQL will optimize them away, but I am not positive.  Better remove them for readability.
 2. Remember to use the `translate` directive to keep our `$$watchers` count low.

Otherwise, run this interface by Patrick and Wayne.  If they agree, LGTM.

client/src/js/services/receipts/ReceiptModal.js
mbayopanda: No, we don't have a report modal

jniles (approved):

jniles:
@mbayopanda, I downloaded your latest changes and built them against Vanga's database.  It's awesome!  Really nice feature :+1:

PR #1225, opened 2017-02-20
Fix patient links from cash and Invoice Registry

### LINKED ISSUE #1188: (bug) Patient Inline Links no longer work
Apparently something in the database changed or in the filtering and now they throw "Column Not Found" errors.

columnnotfound
_Fig 1: Attempting to filter on a Patient Link_

I propose that these inline links be moved into the dropdown like the `View Invoices` and `View Payments` link on the Patients registry.  For variably sized names, that seems to work pretty well.

lomamech:
Fix patient links from cash and Invoice Registry

Closes #1188.


sfount (approved):
LGTM. 👍

Issue #1163, opened 2017-02-10
(refactor) Improvements to the Trial Balance


jniles:
The Trial Balance has never been tested with large amount of data or on older hardware.  Here are some improvements from this experience.

 1. ~~The `cancel` button should persist across every view to allow dismissing the modal. (#1540)~~
 2. ~The headercrumb label could use better formatting (do not bleed over the slash)~
 3. ~Refactor to behave better on smaller screen / lower resolutions (no pull right, use text-right)~
 4. ~Remove outdated warnings (entity_type_id)~
 5. ~Format amounts in the enterprise currency on the grid.~
 6. ~Display totals on the grid footer~
 7. Link to Posting Journal from "View Details" link.
 8. Use `TX.TPA.{reference}` for transactions (https://github.com/Vanga-Hospital/bhima-2.X/issues/245)
 9. Link errors to relevant transactions using state or hyper likes.

![lowrestrialbalance](https://cloud.githubusercontent.com/assets/896472/22826231/a03e5422-ef91-11e6-9799-41ae18e94dda.png)
_Fig 1: Trial Balance at low resolution_

jniles:
Linking prior art: #1042.

Issue #1042, opened 2016-12-10
(refactor) Trial Balance Improvements


jniles:
This issue concerns improvements to the Trial Balance that can be addressed in a single PR or multiple PRs.
 1. ~Change the icon to a `fa fa-balance-scale`~
 2. ~Make sure buttons do not overflow off of the modal.~
 3. Display on all pages of the Trial Balance:
   a. The number of transactions to be posted
   b. ~The total amount being debited~
   c. ~The total amount being credited~
   d. The number of lines involved.
 4. ~Make the headercrumb labels render correctly.~
 5. Every page should offer a "Post Transactions" button and a "Cancel" button
 6. ~There should be no grid grouping - this slows the grid down and is overly complex.~
 7. All transactions should be linked in the view, so that clicking one will close the modal and scroll the Posting Journal to that particular transaction.
 8.  ~~The Trial Balance should be a child state of the Journal.~~

jniles:
Closing this as superseded by  #1163.

PR #1233, opened 2017-02-21
refactor: Trial Balance

### LINKED ISSUE #868: (bug) The Trial Balance sometimes groups by account
The Trial Balance is a heavy module, since it has a ui-grid in a modal on a ui-grid.  Sometimes, when the trial balance modal opens, the UI grid groups by account initially, instead of removing the grouping. 

This happens randomly - there does not seem to be a way to reproduce the behavior predictably.

See below.
trialbalancegroupedaccounts
_Fig 1: Group Accounts_

### LINKED ISSUE #1171: (bug) Trial Balance Submit Needs Loading Indicator
The Trial Balance is one of the longest operations performed when working on Raspberry Pi with upwards of 3000+ transactions.  It must disable the entire modal and put the button in a loading state.  Otherwise, the user may navigate away or try to click multiple times leading to unpredictable behavior.

### LINKED ISSUE #1189: (feature request) Trial Balance needs Grid Totals
The Trial Balance currently does not show the total of movements - one of its primary functions.  To aid in analysis, this should be corrected as soon as possible.

trialbalanceneedsaggregationfunctions
_Fig 1: Trial Balance needs aggregations_

jniles:
**EDIT** See https://github.com/IMA-WorldHealth/bhima-2.X/pull/1233#issuecomment-281334242 for the most recent changes.

This PR refactors the Trial Balance to improve stability and usability.  It addresses a few of the issues cited in multiple issues.  It also fixes a bug with the Trial Balance calculation of account balance changes.

**Details**
This commit improves the Trial Balance UI and UX in the following ways:
 1. The submit button now disables and enters a loading state after
 posting to the GL.
 2. The grid now right aligns the currency amounts.
 3. The grid now uses currency rendering for all amounts.
 4. The buttons now behave responsively.
 5. The labels no longer bleed into the headercrumb.

There are also some performance improvements:
 1. Use flat entity access on the grids
 2. Use fastWatch
 3. Use translate directive where possible

Addresses some of the points in #1163.  Closes #1171.  Addresses some of the points in https://github.com/IMA-WorldHealth/bhima-2.X/issues/1042.

**Bug Fix**
This commit fixes a bug in the totals calculation for the trial balance
due to an incorrect LEFT JOIN.  This has been broken out into multiple
subqueries.  The performance of these queries should be investigated,
but there are no obvious optimizations to make.

**Bug Fix**
The `$timeout` hacks have been removed from `ui-grid` and replaced with events that redraw the grid when grouping changes.  These sometimes caused flaky behavior in rendering.

Closes #868.

**Feature**
Finally, this PR includes a grid footer on the Trial Balance to give an overview of the changes.  This is how the bug was discovered.  See below on Vanga's dataset.

Closes #1189.

**Feature**
The Trial Balance finally has a loading indicator!

trialbalanceimprovements
_Fig 1: Trial Balance Improvements_

---
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/client/src/partials/journal/modals/trialBalanceDetail.body.js
+++ b/client/src/partials/journal/modals/trialBalanceDetail.body.js
@@ -14,32 +14,39 @@ TrialBalanceDetailBodyController.$inject = [
 function TrialBalanceDetailBodyController(Session, trialBalanceService, Grouping, $state, $stateParams) {
   var vm = this;
   var cssClass = trialBalanceService.getCSSClass($stateParams.feedBack);
+
   var columns = [
     { field : 'trans_id', displayName : 'TABLE.COLUMNS.TRANSACTION', headerCellFilter: 'translate', headerCellClass : cssClass},
     { field : 'account_number', displayName : 'TABLE.COLUMNS.ACCOUNT', headerCellFilter: 'translate', headerCellClass : cssClass},
     {
-      field : 'debit_equiv',
-      displayName : 'TABLE.COLUMNS.DEBIT',
-      headerCellFilter: 'translate',
-      cellTemplate : '/partials/templates/grid/debit_equiv.cell.html',
-      headerCellClass : cssClass
+      field            : 'debit_equiv',
+      displayName      : 'TABLE.COLUMNS.DEBIT',
+      headerCellFilter : 'translate',
+      headerCellClass  : cssClass,
+      cellFilter       : 'currency:' + Session.enterprise.currency_id,
+      cellClass        : 'text-right',
     },
     {
-      field : 'credit_equiv',
-      displayName : 'TABLE.COLUMNS.CREDIT',
-      headerCellFilter: 'translate',
-      cellTemplate : '/partials/templates/grid/credit_equiv.cell.html',
-      headerCellClass : cssClass
+      field            : 'credit_equiv',
+      displayName      : 'TABLE.COLUMNS.CREDIT',
+      headerCellFilter : 'translate',
+      headerCellClass  : cssClass,
+      cellFilter       : 'currency:' + Session.enterprise.currency_id,
+      cellClass        : 'text-right',
     }
   ];
 
   vm.stateParams = $stateParams;
+
   vm.gridOptions = {
-    enableColumnMenus: false,
-    treeRowHeaderAlwaysVisible: false,
-    appScopeProvider: vm,
-    columnDefs : columns
+    enableColumnMenus          : false,
+    treeRowHeaderAlwaysVisible : false,
+    appScopeProvider           : vm,
+    columnDefs                 : columns,
+    flatEntityAccess           : true,
+    fastWatch                  : true,
   };
+
   vm.grouping = new Grouping(vm.gridOptions, false);
   vm.enterprise = Session.enterprise;
   vm.gridOptions.data = $stateParams.lines;
@@ -52,7 +59,7 @@ function TrialBalanceDetailBodyController(Session, trialBalanceService, Grouping
   function viewErrorList() {
     // FIX ME : what is the good way? storing original data in appcache?
     var lines = trialBalanceService.parseErrorRecord($stateParams.errors);
-    $state.go('trialBalanceErrors', {lines : lines, feedBack : $stateParams.feedBack, records : $stateParams.records}, {reload : false});
+    $state.go('trialBalanceErrors', { lines: lines, feedBack: $stateParams.feedBack, records: $stateParams.records }, { reload : false });
   }
 
   vm.viewErrorList = viewErrorList;
--- a/client/src/partials/journal/modals/trialBalanceMain.body.html
+++ b/client/src/partials/journal/modals/trialBalanceMain.body.html
@@ -1,49 +1,42 @@
-<p>
-  <div class="row">
-    <div class="col-xs-4 pull-left">
-      <button
-        type="button"
-        ng-show="TrialBalanceMainBodyCtrl.showErrorButton"
-        ng-class="
-          { 'btn btn-danger' : TrialBalanceMainBodyCtrl.feedBack.hasError,
-            'btn btn-warning': TrialBalanceMainBodyCtrl.feedBack.hasWarning,
-            'btn btn-primary' : TrialBalanceMainBodyCtrl.feedBack.hasSuccess}"
-        ng-click="TrialBalanceMainBodyCtrl.viewErrorList()"
-        data-method="show-error">
-        <i class="fa fa-bars"></i>
-        {{"FORM.BUTTONS.VIEW_ERRORS" | translate }}
-      </button>
-    </div>
-
-    <div class="col-xs-3 pull-right">
-      <button
-        ng-if="!TrialBalanceMainBodyCtrl.loading"
-        id="switchViewID"
-        type="button"
-        class="btn btn-default"
-        ng-click="TrialBalanceMainBodyCtrl.switchView()"
-        data-method="switch">
-        <i class="fa fa-retweet"></i>
-        {{TrialBalanceMainBodyCtrl.viewDetail.key | translate }}
-      </button>
-    </div>
+<div class="row" style="margin-bottom: 5px;">
+  <div class="col-xs-4">
+    <button
+      type="button"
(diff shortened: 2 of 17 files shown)


jniles:
Interested in @imawh's feedback as well.

jniles:
Received feedback from @imawh.  He proposes the following changes:
 1. Remove the total footer on `avant` and `apres` columns - these make no sense.
 2. Only show the footer if debits/credits imbalance.  They should be balanced normally.
 3. Make the `Voir les Details` text smaller/more compact.
 4. Include the account label with the account number.
 5. Remove the "Group by Transaction" button.

DedrickEnc:
We should also ask Patrick, what he is thinking.

jniles:
@sfount, can you see if Patrick has any feedback?

sfount:
Patrick is not at the office right now. I understand that he is not in good health.
@patsonjunior

jniles:
The following changes have landed:

This commit incorporates feedback from our accountant.
 1. Grid Footer is only shown when there are imbalances and is colored red.
 2. Totals are only ever shown for debit/credit and not for before/after.
 3. The grid is no longer padded - slightly more screen real estate for users.
 4. The account now includes the account text.
 5. Removed Group By Transaction
 6. Cancel button on detail page.
 7. "View Details" is translated just as "details".

**Screenshots**
trialbalancevalidstate
_Fig 1: Trial Balance Valid State_

trialbalanceerrorstate
_Fig 2: Trial Balance Error State_

client/src/partials/journal/modals/trialBalanceDetail.body.js
```diff
@@ -14,32 +14,39 @@ TrialBalanceDetailBodyController.$inject = [
...
   vm.gridOptions = {
-    enableColumnMenus: false,
-    treeRowHeaderAlwaysVisible: false,
-    appScopeProvider: vm,
-    columnDefs : columns
+    enableColumnMenus          : false,
+    treeRowHeaderAlwaysVisible : false,
+    appScopeProvider           : vm,
+    columnDefs                 : columns,
+    flatEntityAccess           : true,
```
sfount, line 46: :+1:

client/src/partials/journal/modals/trialBalanceMain.body.html
```diff
@@ -1,49 +1,29 @@
...
-        empty-state="TrialBalanceMainBodyCtrl.gridOptions.data.length === 0"
-        error-state="TrialBalanceMainBodyCtrl.hasError"
-      >
-      </bh-grid-loading-indicator>
-  </div>
+<div
+  id="main-grid"
+  ui-grid="TrialBalanceMainBodyCtrl.gridOptions"
+  ui-grid-auto-resize
+  ui-grid-grouping>
```
sfount, line 23: Does this UI grid still need `ui-grid-grouping` ?

client/src/partials/journal/modals/trialBalanceMain.body.js
```diff
@@ -2,8 +2,9 @@ angular.module('bhima.controllers')
   .controller('TrialBalanceMainBodyController', TrialBalanceMainBodyController);
 
 TrialBalanceMainBodyController.$inject = [
-  'SessionService', 'TrialBalanceService', 'GridGroupingService', 'GridColumnService',
-  'NotifyService', '$state', '$timeout', 'uiGridConstants',
+  'SessionService', 'TrialBalanceService', 'NotifyService',
+  '$state', 'uiGridConstants', 'uiGridGroupingConstants', '$filter',
```
sfount, line 6: Are grouping constants used any more? I think that functionality was removed with the view as transaction button. If they're no longer used all of these imports can be removed.

client/src/partials/journal/modals/trialBalanceMain.body.js
```diff
@@ -213,40 +218,44 @@ function TrialBalanceMainBodyController(Session, trialBalanceService, Grouping,
...
-      vm.loading = false;
-      vm.showErrorButton = vm.isInvalid && !vm.loading;
-    });
+        vm.feedBack = TrialBalance.getFeedBack(errorList); // getting a feedback object to customize the grid
+        vm.isInvalid = vm.feedBack.hasError || vm.feedBack.hasWarning;
 
-  fetchDataByAccount()
-    .then(function (data) {
-      vm.gridOptions.data = data;
+        cssClass = TrialBalance.getCSSClass(vm.feedBack);
```
sfount, line 233: Is there a reason these CSS classes can't be assigned in the template like in many other modules using `ng-class`? I haven't looked into the specifics but this logic is very specific and unknown to me. If it is required it would be a good thing to standardise for these cases.

client/src/partials/journal/modals/trialBalanceMain.body.html
```diff
@@ -1,49 +1,29 @@
...
+  <bh-grid-loading-indicator
```
sfount, line 24: :+1: 🎉

sfount (commented):
The code here LGTM. The changes to the SQL will make this a very robust procedure. 

I have made a few comments however I don't think any of these things should be blocking. 

I also have a number of comments on the UX experience for this module: 
1. Is there a reason that a different blue/ red colour is used for primary and error states? We have no other grids that have a blue header so when and where we use this should be considered. If these should have colours we should use the standard primary and danger colours used throughout the application. 
2. The submit button turns red if there are errors. A red button is used throughout the application for a serious action (usually deleting an entity) however this button can never be pressed and won't delete anything - I think this is misleading. 
3. The view errors button is also red, this is the same point as (2.) 

@jniles the code here looks to me like it can be merged into master. If you would like to make any comments/ changes based on this review there is no problem. If this has to be in as soon as possible let me know and we can address the additional comments in an issue thread.

client/src/partials/journal/modals/trialBalanceMain.body.html
jniles: Yes to make the `uiGridGroupingConstants` tree aggregation work.  I might be able to migrate this to `uiGridConstants` instead, but for the moment, that's the way it was set up.  The grouping aggregation doesn't work without the `ui-grid-grouping` directive.

client/src/partials/journal/modals/trialBalanceMain.body.js
jniles: It would be fairly straightforward.  You just have to make a header row template and switch the color based on that.

jniles: Header template:

https://github.com/angular-ui/ui-grid/blob/master/src/templates/ui-grid/ui-grid-header.html

client/src/partials/journal/modals/trialBalanceMain.body.js
jniles: They are used in the aggregation functions.

jniles:
@sfount, let's just get this in.  It will be useful to teach the team here without the previous bugs and with a loading button.  The others will just be UI issues and/or performance that we can address through more training when we make the upgrades.

Note: Travis CI isn't reporting it, but these tests pass.

patsonjunior:
1. A mon avis, il faudra d'abord revoir la configuration informatique de ce "Trial Balance" du fait qu'en principe toutes les écritures  qu'elles soient générées automatiquement ou passées manuellement dans le journal de saisie dévraient toujours être équilibrées.

2. Si après revision, l'erreur persiste, être alors capable de retouver le document ou la pièce qui soutient la transaction à partir de la colonne "Référence" qui se trouve dans le" journal de saisie" afin d'équilibrer la transaction du vrai montant, au lieu de le faire de façon imaginaire avec risque de surestimation ou de sousestimation de la valeurs du compte.

jniles:
@patsonjunior

 1. En principe, le system marche avec les transactions équilibrées.  Main il faut tous les temps faire la verification parce ce que le reseau, l'oridinateur, le courant, ou autres choses peuvent bouger les informations.  Dans le fonctionnement normale, on n'aura pas des déquilibrées.

 2. Vous avez raison!  Si l'utilisateur clique sur le bouton "Voir Erreurs", ca va lui montrer les peices.

sfount:
Okay, LGTM. We'll make issues to track UX improvements.

PR #1237, opened 2017-02-21
(fix) Registry filters

### LINKED ISSUE #1190: (bug) Credit Note filters do not commute
When you add a `reversal` filter to the FilterParser, it ignores all other filters.  See below:

creditnotesdonotrespectmultiplefilters
_Fig 1: Reversal does not respect additive filters_

### LINKED ISSUE #1191: (bug) Cannot clear "Client" filters when accessed from Patient Registry
The Patient Registry has lots of excellent links to payments and invoices.  When you click to see a client's invoices, you cannot clear the filter with the "Clear All Filters" button.  This is very annoying to say the least and breaks the page.

cannotclearclientfilter
_Fig 1: Clicking the "Clear Filters()" Link will do nothing in this case_

### LINKED ISSUE #1192: (bug) Credit Note submission refreshes registry without taking into account filters.
After successfully making a credit for a credit note for an invoice, the modal disappears and the registry is refreshed without taking into account the filters.  This leaves a dangling filter label at the top of the grid.  Ideally, this should either:
 1. Remove all filters and labels OR
 2. Leave the filters applied.

precreditnotefilters
_Fig 1: Before Credit Note - one filter applied that limits grid to a single item_.

postcreditnotefilters
_Fig 2: After Credit Note - Registry refreshes without respecting filters_.

sfount:
This pull request fixes a number of issues in the cash and invoices registry filters. 
* After reversing a record filters are ignored even though the filters are still shown in the module 
* Allow filters to be removed when passes in through a `$state.parameter` link 
* Ensure invoice `reversed` filter respects other filters 

For linking modules with filters: 
* Removes debtor_uuid filter from being passed in as this can't even be search in the registry 
* Uses patient reference to filter through the link

It also works to bring the cash payment registry inline with more up to date standards. 
* Support caching filters 
* Rename variables/ methods to match other modules 
* Update the loading of cached and state parameters to read like other modules
* Remove very specific logic for display text on filters (still requires a standard across module)
* Adds patient reference filter to cash payments

Closes #1192 
Closes #1191 
Closes #1190

### DIFF
--- a/client/src/partials/cash/payments/templates/search.modal.html
+++ b/client/src/partials/cash/payments/templates/search.modal.html
@@ -71,6 +71,11 @@
         </ui-select>
       </div>
 
+      <div class="form-group">
+        <label class="control-label" translate>FORM.LABELS.REFERENCE_PATIENT</label>
+        <input type="text" class="form-control" ng-model="$ctrl.bundle.patientReference"/>
+      </div>
+
       <bh-currency-select
         currency-id="$ctrl.bundle.currency_id"
         validation-trigger="CashVoucherForm.$submitted">
--- a/client/src/partials/patient_invoice/registry/registry.js
+++ b/client/src/partials/patient_invoice/registry/registry.js
@@ -90,17 +90,6 @@ function InvoiceRegistryController(Invoices, bhConstants, Notify, Session, Recei
     vm.hasError = false;
     toggleLoadingIndicator();
 
-    parameters = $state.params.filters ? $state.params.filters : parameters;
-
-    if($state.params.display){ 
-      var display = $state.params.display;
-      vm.filtersFmt = Invoices.formatFilterParameters(display);
-      // show filter bar as needed
-      vm.filterBarHeight = (vm.filtersFmt.length > 0) ?  FILTER_BAR_HEIGHT : {};
-
-    }
-
-
     // if we have search parameters, use search.  Otherwise, just read all
     // invoices.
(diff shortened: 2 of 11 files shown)


client/src/partials/cash/payments/templates/search.modal.html
```diff
@@ -71,6 +71,11 @@
         </ui-select>
       </div>
 
+      <div class="form-group">
```
jniles, line 74: :+1:  This will be useful.

client/src/partials/patient_invoice/registry/registry.js
```diff
@@ -162,10 +151,14 @@ function InvoiceRegistryController(Invoices, bhConstants, Notify, Session, Recei
 
   // startup function. Checks for cached filters and loads them.  This behavior could be changed.
   function startup() {
-    vm.filters = cache.filters;
+    // @TODO standardise loading/ caching/ assigning filters with a client service
```
jniles, line 154: We should create an issue tracking this.

jniles (approved):
Looks good to me.  We definitely need a standard way to do filters on the client.  I think @mbayopanda has started to make some good progress on that using the [SearchFilterFormatService](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/SearchFilterFormatService.js). 

This PR contains several bug fixes that are critical for production.  We can work on the refactors once a design has been submitted,

PR #1238, opened 2017-02-21
(fix) Debtor group deletion/ pagination

### LINKED ISSUE #1158: (bug) Debtor Groups limits view without paginating
Imported from https://github.com/Vanga-Hospital/bhima-2.X/issues/233.

The debtor groups module limits the display of groups to a certain number of groups ... however it does not provide the tools for pagination.  This makes it so that it looks like some debtor groups are not added and can lead to adding them twice.  The pagination tools should be added back in so that the module is properly usable.

### LINKED ISSUE #1227: (bug) Cancelling a 'Delete Debtor Group' modal reports 'Removed with success' message
On the Debtor Groups management page you can remove a debtor group with a modal interface. Cancelling the deletion results in a success message displaying 'successfully removed' 

Expected behaviour: 
No notification is given for cancelled deletion.

sfount:
This commit fixes the issues in pagination for debtor groups, for now pagination is removed given the user experience with filtering. All debtor groups are shown by default. 
* use `<ul uib-pagination>` vs. `<uib-pagination>`
* Include accurate page sizes

It also fixes the promise chain for deleting debtor groups which would previously incorrectly report a successful deletion even if cancel had been pressed. This also allows the request to correctly handle errors. 

It also resolves a bug in clearing the debtor groups sort allowing the user to clear the sort correctly. 

Closes #1158 
Closes #1227

### DIFF
--- a/client/src/partials/debtors/groups.js
+++ b/client/src/partials/debtors/groups.js
@@ -67,15 +67,19 @@ function DebtorGroupController($state, DebtorGroups, Accounts, Prices, $interval
    */
   function deleteGroup(groupUuid) {
     Modal.confirm()
-      .then(function (ans) {
-        if (!ans) { return false; }
-        return DebtorGroups.remove(groupUuid);
+      .then(function (confirmResponse) {
+        if (!confirmResponse) {
+          return false;
+        }
+
+        // user has confirmed removal of debtor group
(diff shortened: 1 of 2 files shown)


jniles:
LGTM

PR #1262, opened 2017-02-24
(feature) Posting Journal editing

jniles:
`//@TODO Full description`

**Default Posting Journal View**
screenshot 2017-02-23 at 13 44 10

**'Transaction' Mode**
screenshot 2017-02-23 at 13 44 59

**Edit Transaction Mode**
screenshot 2017-02-23 at 13 45 48

**Supported Editable Columns**
* Date - value written to database 
* Description - value written to database 
* Account - Number mapped to Account ID
* Debit - value written to database 
* Credit - value written to database 
* Recipient - Human readable reference mapped to UUID
* Reference - Human readable reference mapped to UUID

**Edit Toolbar Functions**
screenshot 2017-02-23 at 13 54 02
* Add row - adds a row to the current transaction, common data that is shared with each row is already populated on the transaction 
* Remove row(s) - Removes any row from the transaction that is selected. **Note: ** removing all the rows in a transaction and then saving is allowed in this version. 
* Save Transaction - Attempt to write changes to the database. If this is successful this will end edit mode. If there is an issue it will be reported in a Notification. 
* Cancel Edit - Stop editing and reset data to how it was.

**Filter + Edit**
Application logic - If you have filtered the server side data (reducing the posting journal to a subset), all of the rows for a transaction will be added if you click to edit it. Once you have 'Saved' or 'Cancelled' the edit the lines of the transaction that do not apply to the filter will be removed again. 
screenshot 2017-02-23 at 13 50 50

**Inline Filter**
screenshot 2017-02-23 at 13 53 08

**Tools Menu**
screenshot 2017-02-23 at 13 49 14

**Feature Requirements** 
* E2E for editing corner cases
* Integration tests for updating transaction rows 
* Refactor and comment editing services - split into two services

**Disclaimer** - This pull request skips journal end to end tests, these should be completely re-written for the new functionality

Supercedes #1253. Closes #1253.


jniles:
This PR fixes some corner cases (such as running a Trial Balance while editing) and makes cellNav disabled on the plugin columns.

PR #1263, opened 2017-02-24
Posting journal lets you hide all columns

### LINKED ISSUE #999: (bug) Posting Journal lets you hide all columns
The Posting Journal lets you hide every single column without complaint.  This yields a very broken looking Posting Journal.  Since columns are cached by default, it may confuse users who do not know this functionality exists.  See below:
postingjournalhidecolumns
_Fig 1: Posting Journal with all columns hidden._

I recommend we display a warning message that tells the user that at least one column should be selected when they are making their selections.

lomamech:
Posting journal lets you hide all columns 
Prevent the user from hiding all posting journal columns

closes #999

### DIFF
--- a/client/src/js/services/grid/GridColumns.js
+++ b/client/src/js/services/grid/GridColumns.js
@@ -2,7 +2,7 @@ angular.module('bhima.services')
 .service('GridColumnService', GridColumnService);
 
 GridColumnService.$inject = [
-  'uiGridConstants', 'AppCache', '$uibModal', 'util', '$timeout', 'NotifyService'
+  'uiGridConstants', 'AppCache', '$uibModal', 'util', '$timeout'
 ];
 
 /**
@@ -19,7 +19,7 @@ GridColumnService.$inject = [
  *
  * @todo - investigate using ui-grid-saveState for caching the column
  */
-function GridColumnService(uiGridConstants, AppCache, Modal, util, $timeout, Notify) {
+function GridColumnService(uiGridConstants, AppCache, Modal, util, $timeout) {
 
   /** @const cache alias for this service */
   var serviceKey = '-Columns';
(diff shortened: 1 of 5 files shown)


jniles (changes requested):
I don't think this solution goes in the right direction.  The code is okay, but the user experience isn't good.

Here's my proposal:
 1. Use a `text-danger` notification in the modal whenever the users hides all columns.  If the user has at least one column visible, the notification should be hidden again.
 2. Disable the "submit" button if there are fewer than 1 visible columns.

Currently, this code broadcasts a change event on every click.  This is inefficient.  You could make this more efficient by moving the 

```js
this.gridApi.core.notifyDataChange(uiGridConstants.dataChange.COLUMN);      
```
call into the `submit()` method.

client/src/js/services/grid/GridColumns.js
```diff
@@ -137,6 +137,36 @@ function GridColumnService(uiGridConstants, AppCache, Modal, util, $timeout) {
...
 
+
+  /**
+   * @method checkColumnMap
+   *
+   * @description
+   * This structure is used to count the number of selected columns
+   *
+   */
+  Columns.prototype.checkColumnMap = function checkColumnMap(columns) {
```
jniles: This is a well named and well described function.  Very nice!

client/src/partials/templates/modals/columnConfig.modal.js
```diff
@@ -44,7 +47,14 @@ function ColumnsConfigModalController(ModalInstance, Columns) {
     ModalInstance.dismiss();
   }
 
+  function checkVisible(){
+    var columnNumber = Columns.checkColumnMap(vm.map);
```
jniles: In the future, a more meaningful name would be better - `Columns.checkColumnMap()` does not return a number - it return a boolean.  A better name is `hasEnoughColumns` or something along those lines.

client/src/partials/templates/modals/columnConfig.modal.js
```diff
@@ -29,12 +29,15 @@ function ColumnsConfigModalController(ModalInstance, Columns) {
...
   function submit() {
+    if (vm.notification) { return; }
+
     Columns.setVisibleColumns(vm.map);
     ModalInstance.close();
   }
 
   // reset the column visibility to their default configuration
   function resetDefaults() {
+    vm.notification = false;
```
jniles: `vm.notification` could also use a nicer name, such as `vm.hasTooFewColumns`.

jniles (commented):
@lomamech, this code looks really good!

I cannot get this PR to work when I test it ... does it work on your machine?

jniles:
LGTM

PR #1285, opened 2017-02-28
fix(build) call asset hash/ compilation on JS watch

sfount:
This commit replaces direct Javascript compilation with re-versioning
and compiling assets on every javascript file change. This ensures that
the changes are factored into the build being used by the current
server.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -223,7 +223,9 @@ gulp.task('lint-i18n', (cb) => {
 // watches for any change and builds the appropriate route
 gulp.task('watch-client', () => {
   gulp.watch(paths.client.css, ['client-compile-css']);
-  gulp.watch(paths.client.javascript, ['client-compile-js']);
+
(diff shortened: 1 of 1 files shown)


jniles:
Can we just replace all the gulp-watches with compile assets?

sfount:
Right now I don't think all of the assets are versioned - as compile-assets does everything (taking the longest time) running individual jobs should be fine for now. As we update our build process this could change!

jniles:
That makes sense.  As long as changes are still reflected when HTML, CSS, or JS is changed.

Tests pass - merging this.

sfount:
Note: It seems like this continuously adds version names to the end of vendor files on every save of a Javascript file - this is resolved by running a full `gulp build`. It should be investigated.

jniles:
Adding gulp's 'clean' to be run before should solve it.  You might need to
break up cleaning the client and not the server.

PR #1287, opened 2017-02-28
feat(receipts) Add service line to invoice receipt

### LINKED ISSUE #1130: (refactor) Patient Invoice POS receipt should list service on it
Every invoice in bhima is assigned to a service for easy tracking.  Unfortunately, this does not appear on the POS receipt which makes it hard to compare the printed receipt and the computer record.  Adding the `service` in the header information of the POS invoice would resolve this problem.

sfount:
This commit adds the service that the invoice was created for to both
the invoice POS and standard receipt. On the standard receipt it
replaces the duplicated amount field, this is so that there is one point
of truth for the total amount on the receipt, it can be undone if
required.

Closes #1130 

**POS Receipt**
screenshot 2017-02-28 at 15 37 37

**Standard Receipt**
screenshot 2017-02-28 at 15 38 45


jniles:
LGTM.

PR #1301, opened 2017-03-02
Fix: report running balances

### LINKED ISSUE #1273: (bug) Running Balances do not work in Production
The running totals on the accounts report does not work in a multi-connection environment.  The `SET balance := 0` syntax only applies to individual MySQL connections which are taken randomly from a pool. Therefore, chaining `db.exec()` calls via `.then()` promise chains does not guarantee that you will receive the same connection with the initialized variable.

The running total should be redesigned with this consideration in mind - a transaction may yield the correct result, but a stored procedure or MySQL function would be the best solution to this problem.

jniles:
This commit implements the correct cumulative balance summation for the
accounts report that initializes and updates variables in the same SQL
statement. By maintaining all variables in the same statement, the
variables are ensured to be zeroes out prior to usage and do not spread
over multiple MySQL connections.

This solution, while complex, allows total control over filtering via
our standard filtering solutions.  Placing the query in a Stored
Procedure (another solution to localize the variable to a single MySQL
connection/context) would require that any further filters write a new
interface to the stored procedure.

Closes #1273.

----
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/server/controllers/finance/reports/reportAccounts/index.js
+++ b/server/controllers/finance/reports/reportAccounts/index.js
@@ -79,28 +79,28 @@ function getAccountTransactions(accountId, source, dateFrom, dateTo) {
     params.push(new Date(dateFrom), new Date(dateTo));
   }
 
-  const csum = 'SET @csum := 0;';
-
   const sql = `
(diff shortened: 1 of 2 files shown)


sfount (approved):
LGTM.

PR #1306, opened 2017-03-03
Schema fee center

DedrickEnc:
This PR adds an initial schema and API to continue the feature developing
Here are things to notice in this PR : 

- The table ```cost_center``` and ```profit_center``` are deleted
-  A new table ```fee_center``` is created```
- The modules cost center and profit center are deleted
- The service module is updated to fit with the  changes
- The delete functionality is not yet implemented and the integration  test for this functionality is ignored, because we did not talk about that during the design. 
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


DedrickEnc:
@sfount @jniles

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,230 @@
...
+* @example
+* // GET /fee_centers : Get list of fee_centers
+* var feeCenters = require('finance/fee_centers');
+* feeCenters.list(req, res, next);
+*/
+
+function list (req, res, next) {
+  'use strict';
+
+  let ordering = 'ORDER BY fc.label';
```
jniles: Just a note, since these values never change, `const` is more appropriate.  ESLint should alert you of this - if it isn't, be sure to configure your text editor to use it.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,230 @@
...
+* Returns an array of fee centers
+*
+* @example
+* // GET /fee_centers : Get list of fee_centers
+* var feeCenters = require('finance/fee_centers');
+* feeCenters.list(req, res, next);
+*/
+
+function list (req, res, next) {
+  'use strict';
```
jniles: We make the entire server globally `strict`.  No need to have this `use strict` directive.

See https://github.com/IMA-WorldHealth/bhima-2.X/pull/1105#discussion_r98899888 as well.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,230 @@
...
+
+  let ordering = 'ORDER BY fc.label';
+  let is_detailed = Boolean(Number(req.query.detailed));
+  let detailed = is_detailed ? ', fc.note, fc.is_principal, fc.project_id, fc.is_cost, p.name, p.abbr, p.enterprise_id, p.zs_id ' : '';
+
+  let sql =
+    `SELECT fc.id, fc.label, fc.is_cost ${detailed} FROM fee_center AS fc JOIN project as p on fc.project_id = p.id`;
+
+  let notInStatement = `(fc.id NOT IN (SELECT s.cc_id FROM service AS s WHERE NOT ISNULL(s.cc_id) UNION SELECT s.pc_id FROM service AS s WHERE NOT ISNULL(s.pc_id)) AND ?)`;
```
jniles: This statement would probably be more efficient if you used `SELECT DISTINCT`.  That way, MySQL could eliminate duplicates before searching through the array.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,230 @@
...
+  .catch(next)
+  .done();
+}
+
+/**
+* Create a fee center in the database
+*
+* @example
+* // POST /fee_centers : Insert a fee center
+* var feeCenters = require('finance/fee_centers');
```
jniles: Giving example code in your comments is an excellent idea.  It makes the usage clear, and it can potentially be extracted and run by a documentation program to make sure that the function behaves as expected.

However, this code will not work, because the file name is `finance/feeCenter.js`.  Could you either remove the misleading comment or update the name?   This problem has been raised in another review https://github.com/IMA-WorldHealth/bhima-2.X/pull/1105#discussion_r98898911.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,230 @@
...
+*/
+
+function getFeeValue (req, res, next){
+  var sql = null;
+
+
+  lookupFeeCenter(req.params.id)
+    .then(function (){
+
+      sql =
```
jniles: Since this variable binding never changes, it would be more appropriate to use `const`.

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,230 @@
...
+        `
+        SELECT
+          IFNULL(IF(t.is_cost = 1, SUM(t.debit_equiv - t.credit_equiv), SUM(t.credit_equiv - t.debit_equiv)), 0) AS value, t.id
```
jniles: This is a pretty complicated query to understand.  It would be more readable if you did the `SUM()` and `IF()` functions on different statements.

Alternatively, you could just do a multiplication by `-1` to inverse the amounts, with a comment explaining why it should be inverted.
```js
const sql = `
  SELECT id, IF(is_cost = 1, value, value * -1) AS value
  FROM (
    SELECT SUM(gl.debit_equiv - gl.credit_equiv) AS value, f.is_cost, f.id
    FROM general_ledger AS gl JOIN fee_center AS f
      ON gl.fc_id = f.id
    WHERE gl.fc_id = ?
    GROUP BY gl.fc_id
  )z;
`
```
This query may also be more performant by only using a single `SUM()`.

server/lib/filter.js
```diff
@@ -9,17 +9,17 @@ const DEFAULT_UUID_PARTIAL_KEY = 'uuid';
...
- * It implements a number of built in 'Filter Types' that allow column qurries
+ * It implements a number of built in 'Filter Types' that allow column queries
  * to be formatted for tasks that are frequently required.
  *
  * Supported Filter Types:
  * * equals - a direct comparison
  * * text - search for text contained within a text field
- * * dateFrom - limit the querry to records from a date
- * * dateTo - limit the querry to records up until a date
+ * * dateFrom - limit the query to records from a date
```
jniles, line 21: :+1:

server/lib/filter.js
```diff
@@ -43,7 +43,7 @@ class FilterParser {
   }
 
   /**
-   * @method text
+   * @method fullText
```
jniles, line 46: :+1:

server/models/schema.sql
```diff
@@ -316,51 +315,51 @@ CREATE TABLE `consumption_service` (
...
   `note` text,
   PRIMARY KEY (`id`),
-  KEY `project_id` (`project_id`),
-  KEY `auxi_cc_id` (`auxi_cc_id`),
-  FOREIGN KEY (`project_id`) REFERENCES `project` (`id`),
-  FOREIGN KEY (`auxi_cc_id`) REFERENCES `cost_center` (`id`)
+  UNIQUE KEY `cost_assignation_1` (`auxiliary_cost_id`, `period_id`),
+  KEY `auxiliary_cost_id` (`auxiliary_cost_id`),
+  KEY `period_id` (`period_id`),
+  FOREIGN KEY (`auxiliary_cost_id`) REFERENCES `fee_center` (`id`),
```
jniles, line 347: Will this always contain costs?  Or can we assign profits as well?  If so, we should find a better table name before we get too far.

test/integration/feeCenter.js
```diff
@@ -0,0 +1,159 @@
...
+
+  it('GET /fee_centers returns a list of fee centers', function () {
+    return agent.get('/fee_centers')
+      .then(function (res) {
+        helpers.api.listed(res, 7);
+      })
+      .catch(helpers.handler);
+  });
+
+  /* @todo - make this route ?detailed=1 to conform to standards */
```
jniles: This comment is no longer needed.

test/integration/feeCenter.js
```diff
@@ -0,0 +1,159 @@
...
+      .then(function (res) {
+        expect(res).to.have.status(200);
+        expect(res).to.be.json;
+        expect(res.body.id).to.equal(newFeeCenter.id);
+        expect(res.body.note).to.equal(updateInfo.note);
+      })
+      .catch(helpers.handler);
+  });
+
+  it.skip('DELETE /fee_centers/:id deletes a fee_center', function () {
```
jniles, line 171: Why are these skipped?  Does the delete route not work?

jniles (commented):
This PR has gone through a few iterations and this is the third or fourth review.  There are still a few issues that have carried over from previous reviews that should be addressed, either in code or in comments.

I've made a few suggestions for improvements to your code, but nothing is blocking.  Let me know what you think of the suggestions and we can merge or update the code.

@sfount has been following the design of Fee Centers more closely.   I really only looked at the code - he should take a look and make sure it fits with the design vision.

server/models/schema.sql
DedrickEnc: There is only auxiliary cost center and not auxiliary profit center, I will propose an issue to remove this concept of cost and profit. so we will keep only fee center that will help to restructure this table also.

jniles: Sounds good.  We'll get this in before that change occurs.

test/integration/feeCenter.js
DedrickEnc: We did not plan a deletion operation for fee center so they are there just in case there is a different decision. So there is no route at the server to delete a fee center for now.

DedrickEnc:
@jniles

jniles (approved):
Changes LGTM.

Issue #769, opened 2016-10-04
(bug) Patient Record does not handle missing picture errors


jniles:
The Patient Record page will throw a `404` error to the console when the patient picture is missing or deleted.  It will display a broken icon in the patient image spot.

**Steps to Reproduce**
1. Upload a patient image.
2. Manually remove the image from the server
3. Refresh the Patient Record page

**Expected Result**
1. Either: a) A nice error message should inform the user that the patient has an image and it has been misplaced or b) nothing should be visible to indicate that there was ever an error.

**Actual Result**
1. The browser's default broken image is shown
2. `GET http://localhost:8080/uploads/pics/:uuid 404 (Not Found)` is logged to the console.

PR #1307, opened 2017-03-03
Missing picture

lomamech:
Improve property of picture
- Prevent the browser show Default broken image 
- Adding a transparent image at png.format

#769


jniles:
Does this actually work?   I still get errors in the console:

patientpictureerror
_Errors on loading Patient Page_

lomamech:
I will find other way either with angular or JavaScript to avoid message 404 not found

jniles:
This is good enough for an initial attempt.  We should figure out the best user-interface way of taking/uploading patient images.

PR #1318, opened 2017-03-08
Bug journal vouchers

### LINKED ISSUE #1312: (bug) Journal vouchers - selecting a transaction type does not update validation
The journal voucher module provides a warning for users stating that a transaction referencing cashboxes must have a transaction type. 

> Transactions involving cash accounts (banks or cashboxes) must have a transaction type specified to differentiate them in the cashflow report. Please select a transaction type.

However when a transaction type is selected the validation is not checked and the message remains. Altering a value like debits or credits forces the validation to be re-run and this clears the message. 

This can be resolved by running validation on the selection of a new transaction type.

lomamech:
Bug journal vouchers 
- Clearing the warning when sectioning transaction type after selecting a
- cash accounts, banks and cashboxes

Closes #1312

### DIFF
--- a/client/src/partials/vouchers/VoucherForm.service.js
+++ b/client/src/partials/vouchers/VoucherForm.service.js
@@ -163,7 +163,11 @@ function VoucherFormService(Vouchers, Constants, Session, VoucherItem, Cashboxes
     }
 
     // validate that this uses multiple accounts in the transaction
-    if(uniqueAccountsArray.length !== 0) {
+    
+    // To prevent calling the validation function when selecting the transaction type before selecting accounts
+    var hasNullAccounts = (uniqueAccountsArray.length === 0);
+
+    if(!hasNullAccounts) {
       var hasUniqueAccounts = (uniqueAccountsArray.length > 1);
       if (!hasUniqueAccounts) {
         err = ERROR_SINGLE_ACCOUNT_TRANSACTION;
--- a/client/src/partials/vouchers/VoucherForm.service.js
+++ b/client/src/partials/vouchers/VoucherForm.service.js
@@ -163,27 +163,29 @@ function VoucherFormService(Vouchers, Constants, Session, VoucherItem, Cashboxes
     }
 
     // validate that this uses multiple accounts in the transaction
-    var hasUniqueAccounts = (uniqueAccountsArray.length > 1);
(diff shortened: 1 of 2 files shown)


client/src/partials/vouchers/VoucherForm.service.js
```diff
@@ -163,27 +163,29 @@ function VoucherFormService(Vouchers, Constants, Session, VoucherItem, Cashboxes
     }
 
     // validate that this uses multiple accounts in the transaction
-    var hasUniqueAccounts = (uniqueAccountsArray.length > 1);
-    if (!hasUniqueAccounts) {
-      err = ERROR_SINGLE_ACCOUNT_TRANSACTION;
-    }
+    if(uniqueAccountsArray.length !== 0) {
```
jniles: Good catch!

jniles (approved):
Good catch and good fix.  LGTM.

client/src/partials/vouchers/VoucherForm.service.js
jniles: Actually... this could all be avoided by correcting the `hasUnqiueAccounts` typo.  The code should read:
```js
var hasUniqueAccounts = (uniqueAccountsArray.length !== 0);
```
instead of 
```js
var hasUniqueAccounts = (uniqueAccountsArray.length > 1);
```

This is the typo causing this bug and you won't need to do the `if()`.

client/src/partials/vouchers/VoucherForm.service.js
```diff
@@ -163,7 +163,11 @@ function VoucherFormService(Vouchers, Constants, Session, VoucherItem, Cashboxes
     }
 
     // validate that this uses multiple accounts in the transaction
-    if(uniqueAccountsArray.length !== 0) {
+    
+    // To prevent calling the validation function when selecting the transaction type before selecting accounts
+    var hasNullAccounts = (uniqueAccountsArray.length === 0);
```
jniles: This is a better fix than mine.  Well done! :+1:

jniles:
@lomamech LGTM.  Great fix!

Let me know if it passes test locally and I'll merge.

lomamech:
All integrations and E2E tests pass

PR #1332, opened 2017-03-09
Any Zero be blank

### LINKED ISSUE #1297: (feature request) Any 0 amounts in a report should be blank
In our reports, the cells that do not have amounts are often rendered as $0 or 0FC.  This wastes printer ink, decreases readability and is not necessary.  In SanruTrakker, the amounts are simply not rendered.  I propose that we emulate this functionality by omitting values that are 0.

lomamech:
- Set any 0 amount in Balance Report
- Set any 0 amount in Cashflow Report
- Set any 0 in Report Aged Debtors
- Set any 0 in Accounts Reports
- Set any 0 in repors Income and Expense

closes #1297

### DIFF
--- a/client/src/i18n/en.json
+++ b/client/src/i18n/en.json
@@ -1515,6 +1515,7 @@
     "BALANCE"                    : "Balance",
     "BILLING_SERVICES"           : "Billing Services",
     "CASHBOX_MANAGEMENT"         : "Cashbox Management",
+    "CASHFLOW"                   : "Cashflow",
     "CASH_PAYMENT_REGISTRY"      : "Cash Payment Registry",
     "CASH_WINDOW"                : "Cash Window",
     "COMPLEX_JOURNAL_VOUCHER"    : "Journal Vouchers",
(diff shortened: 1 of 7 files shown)


client/src/i18n/fr.json
```diff
@@ -1243,7 +1243,7 @@
     "ACCOUNT"       : "Rapport du Compte",
     "BALANCE"       : "Rapport de la Balance",
     "AGED_DEBTORS"  : {
-      "TITLE": "Aged Debtors",
+      "TITLE": "Dettes des clients",
```
jniles, line 1246: @patsonjunior had a better name for this.  "Detttes Agee des Clients" or something like that.  You should check with him on the name.

client/src/i18n/en.json
```diff
@@ -1515,6 +1515,7 @@
     "BALANCE"                    : "Balance",
     "BILLING_SERVICES"           : "Billing Services",
     "CASHBOX_MANAGEMENT"         : "Cashbox Management",
+    "CASHFLOW"                   : "Cashflow",
```
jniles, line 1518: :+1: This has been annoying.  Glad you fixed it.

jniles (commented):
Code looks fine - can you include a screenshot so we can see how this works?

client/src/i18n/en.json
lomamech: rapportbalance

jniles:
LGTM :+1: Cleans up the report interface nicely.

PR #1337, opened 2017-03-10
View patient invoice cash

### LINKED ISSUE #1326: (feature request) View Patient in Registry from Invoice Registry and Cash Payments Registry
The Patient Registry has some really awesome links to the Invoice Registry and the Cash Payments Registry.  They help a user audit super quickly.  See below:

patientdropdown
_Fig 1: Patient Registry Dropdown_

It would be helpful to have these links for both the Invoice Registry and the Cash Payments Registry.

**Invoice Registry Links**
  1. View Patient _(concerning this invoice)_
  2. View Cash Payments _(against this invoice)_

**Cash Payment  Links**
 1. View Invoice _(concerning this payment)_
 2. View Patient _(concerning this patient)_

lomamech:
View Patient Registry and Cash Payment From Invoice Registry  …
 - Changing the payment list request to facilitate joining to the invoice table
 - Update cash Format Filter Parameters

View invoice Registry and Patient Registry from Cash Registry

closes #1326


server/controllers/finance/cash.js
```diff
@@ -152,41 +152,49 @@ function list(req, res, next) {
...
+        CONCAT_WS('.', '${identifiers.PATIENT.key}', project.abbr, p.reference) AS patientReference, 
+        do.text AS invoiceReference
+      FROM cash
+        LEFT JOIN voucher ON voucher.reference_uuid = cash.uuid
+        JOIN project ON cash.project_id = project.id
+        JOIN debtor d ON d.uuid = cash.debtor_uuid
+        JOIN patient p on p.debtor_uuid = d.uuid
+        JOIN cash_box cb ON cb.id = cash.cashbox_id
+        JOIN user u ON u.id = cash.user_id
+        LEFT JOIN cash_item ci ON ci.cash_uuid = cash.uuid
```
jniles: There are a lot of extra JOINs here...  As we saw last week on Friday, this makes the Raspberry Pi very slow.  Are you sure you need these?

Why `LEFT JOIN` on `cash` instead of `JOIN`?

jniles (dismissed):
I worry about the complexity of this change... it seems like it would result in a massive slowdown.

Can you figure out how to do SQL WHERE queries without the JOIN?  Something like:

```js
filters.custom('invoiceReference', 'uuid IN (SELECT cash_uuid FROM cash_item WHERE invoice_uuid = ?');
```

That way, there is no JOIN and queries that do not care about invoices never use invoices.

client/src/partials/cash/payments/templates/action.cell.html
```diff
@@ -23,5 +23,17 @@
         <span class="text-danger"><i class="fa fa-ban"></i> <span translate>FORM.LABELS.CANCELED</span></span>
       </a>
     </li>
+    <li class="divider"></li>
+    <li>
+      <a ui-sref="patientRegistry({ filters : { display_name : row.entity.patientName }})" data-method="invoiceReceipt" href>
```
sfount: Should this data method be `invoiceReceipt` ? I can't tell exactly from the code but it looks like this opens a patient card, not an invoice.

client/src/partials/patient_invoice/registry/templates/action.cell.html
```diff
@@ -23,5 +23,17 @@
...
       </a>
     </li>
+    <li class="divider"></li>
+    <li>
+      <a ui-sref="patientRegistry({ filters : { display_name : row.entity.patientName }})" data-method="invoiceReceipt" href>
+        <i class="fa fa-user"></i> <span translate>REPORT.VIEW_PATIENT</span>
+      </a>
+    </li>
+    <li>
+      <a data-method="card" ui-sref="cashRegistry({ filters : { identifiers : { invoice_uuid : row.entity.uuid }, display : { invoice_uuid : row.entity.uuid } }})" href>
```
sfount: The `data-method` does not line up here. This may cause an issue for developers writing tests.

server/lib/filter.js
```diff
@@ -75,6 +76,29 @@ class FilterParser {
...
+
+  /**
+  *@description
+  * Allows to filter the elements relative to the results of a subRequest
+  * @param {String} principale    The element to be compared with the results of the sub-request
+  * @param {String} table         table to be used in subRequests
+  * @param {String} column        Is the equivalence between the table `tableString` and `table` that of the subRequests
+  * @param {String} filterKey     key attribute on filter object to be used in filter
+  */
+  subRequest(principale, table, column, filterKey) {
```
sfount: This is a good idea and very well commented. 

However, calling:
```js
filters.subRequest('uuid', 'cash_item', 'invoice_uuid', 'cash_uuid');
```

creates the string:

```js
`cash.uuid IN (SELECT cash_item.invoice_uuid FROM cash_item WHERE cash_item.cash_uuid = HUID(?))`
```

Does this feature work? The cash UUID should never match the cash_item invoice uuid. 

This makes me think that the API for this function is too confusing, it uses the same parameter names as the other functions however they mean something different (tables and columns in a sub query). 

I would recommend that this behaviour is written using the `custom` method for now. 

Using this function and the function call, that would look like: 

```js
filters.custom('cash_uuid', 'cash.uuid IN (SELECT cash_item.cash_uuid FROM cash_item WHERE cash_item.invoice_uuid = HUID(?)');
```

I suggest this because the table, columns and filterKey provided here do not mean the same things as with the other functions.

jniles:
@lomamech, let me know when you need another review.

lomamech:
Yes, you can review the code

server/controllers/finance/patientInvoice.js
```diff
@@ -225,6 +225,8 @@ function find(options) {
   filters.dateFrom('billingDateFrom', 'date');
   filters.dateTo('billingDateTo', 'date');
 
+  filters.subRequest('uuid', 'cash_item', 'invoice_uuid', 'cash_uuid');
```
jniles: I don't think this is defined anymore.  It's breaking the tests.  Could you remove it?

server/controllers/finance/cash.js
```diff
@@ -177,6 +177,8 @@ function listPayment(options) {
 
   // @TODO Support ordering query (reference support for limit)?
   filters.setOrder('ORDER BY cash.date DESC');
+ 
+  filters.custom('invoice_uuid', 'cash.uuid IN (SELECT cash_item.cash_uuid FROM cash_item WHERE cash_item.invoice_uuid = HUID(?))');
```
jniles, line 181: :+1:  I haven't validated this for correctness, but it's a good re-use of tools.

jniles (changes requested):

jniles:
Excellent job @lomamech!

PR #1338, opened 2017-03-10
fix(invoices) apply patch to limit default invoice

sfount:
This commit applies a temporary patch to resolve an urgent installation
issue. Currently the requests for invoices is very expensive and we
always show all invoices by default. This commit applies a default
period to the invoice registry to only ever show the current date by
default.

Default filters:
* Cannot be removed
* Are only applied if there are no other filters

**Default filter**
screenshot 2017-03-10 at 16 19 17

**Removed on standard filter application**
screenshot 2017-03-10 at 16 20 27

### DIFF
--- a/client/src/partials/patient_invoice/registry/search.modal.js
+++ b/client/src/partials/patient_invoice/registry/search.modal.js
@@ -22,6 +22,9 @@ function InvoiceRegistrySearchModalController(ModalInstance, Users, Services, Da
   vm.periods = Dates.period();
   vm.today = new Date();
 
+  // @FIXME patch hack - this should be handled by FilterService
+  delete(vm.params.defaultPeriod);
+
   // set controller methods
   vm.submit = submit;
   vm.clear = clear;
--- a/server/lib/filter.js
+++ b/server/lib/filter.js
@@ -1,8 +1,16 @@
 const _ = require('lodash');
+const moment = require('moment');
 
 const RESERVED_KEYWORDS = ['limit', 'detailed'];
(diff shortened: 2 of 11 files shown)


client/src/partials/patient_invoice/registry/search.modal.js
```diff
@@ -22,6 +22,9 @@ function InvoiceRegistrySearchModalController(ModalInstance, Users, Services, Da
   vm.periods = Dates.period();
   vm.today = new Date();
 
+  // @FIXME patch hack - this should be handled by FilterService
+  delete(vm.params.defaultPeriod);
```
jniles, line 26: I didn't realize delete was a function!

server/lib/filter.js
```diff
@@ -1,8 +1,16 @@
 const _ = require('lodash');
+const moment = require('moment');
 
 const RESERVED_KEYWORDS = ['limit', 'detailed'];
 const DEFAULT_LIMIT_KEY = 'limit';
 const DEFAULT_UUID_PARTIAL_KEY = 'uuid';
+
+// @FIXME patch code - this could be implemented in another library
+const PERIODS = {
```
jniles, line 9: This is really clean :+1:

test/end-to-end/patient/invoice/registry.spec.js
```diff
@@ -45,4 +48,14 @@ describe('Invoice Registry', () => {
     FU.modal.close();
   });
 
+  function showAllTransactions() {
```
jniles, line 51: Ha!  This hack should be re-imagined.

PR #1353, opened 2017-03-14
Feature stock management

mbayopanda:
This PR implements the stock management modules : 
- [x] stock exit : patient, service, depot, loss
- [x] stock entry: purchase order
- [x] stock movement registry
- [x] stock lots registry
- [x] stock inventories registry
- [x] stock entry : purchase order and integration
- [x] stock adjustment
- [x] e2e tests for stock entry and stock exit modules
- [ ] e2e tests for stock adjustment
- [ ] e2e tests for stock origins registry


client/src/js/components/bhDateInterval.js
```diff
@@ -19,6 +19,7 @@ angular.module('bhima.components')
       validationTrigger : '<', // validation trigger action
       dateFrom : '=',          // date from
       dateTo : '=',            // date to
+      dateId : '@',            // date identifier
```
jniles, line 22: What is the `dateId` for?

client/src/js/components/bhDepotDropdown.js
```diff
@@ -0,0 +1,35 @@
...
+
+  var cache = new AppCache('bhDepotComponent');
+
+  $ctrl.$onInit = function $onInit() {
+    Depot.read()
+    .then(function (rows) {
+      if (!rows.length) { return; }
+      $ctrl.depots = rows;
+      $ctrl.selection = cache.selection || $ctrl.depots[0];
+      $ctrl.$loading = false;
```
jniles: Where did $ctrl.$loading come from?

client/src/js/services/receipts/ReceiptModal.js
```diff
@@ -269,5 +269,167 @@ function ReceiptModal(Modal, Receipts) {
...
+    var instance = Modal.open(configuration);
+    return instance.result;
+  }
+
+  /**
+   * @method stockEntryPurchaseReceipt
+   * @param {string} documentUuid
+   * @param {boolean} notifyCreated
+   */
+  function stockEntryPurchaseReceipt(documentUuid, notifyCreated) {
```
jniles, line 344: Many of these receipts look identical.  I wonder if we could make a factory that could reduce the amount of code here significantly.  Something like this:

```js
function ReceiptFactory(promise, options) {
  var defaults = {
    renderer: Receipts.renderers.PDF,
    notifyCreated: false
  };

  var parameters = angular.extend(defaults, options);
  var provider = {
     resolve :  {
       receipt : function receiptProvider() { return { promise: promise }; },
       options : function optionsProvider() { return parameters; },
     },
  };

  var configuration = angular.extend(modalConfiguration, reportProvider);
  var instance = Modal.open(configuration);
  return instance.result;
}
```

The usage would be something like this:
```js
function stockEntryPurchaseReceipt(documentUuid, notifyCreated) {
  var opts = { title : 'STOCK.RECEIPT.ENTRY_PURCHASE', notifyCreated : notifyCreated, renderer: Receipts.renderers.PDF };
  var promise = Receipts.stockEntryPurchaseReceipt(documentUuid, { renderer: opts.renderer });
  return ReceiptFactory(promise, opts);
}
```

Just a thought to try and reduce boiler plate.

client/src/partials/purchases/templates/cellStatus.tmpl.html
```diff
@@ -1,8 +1,10 @@
 <div class="ui-grid-cell-contents">
     <div ng-if="row.entity.is_confirmed === 1 && row.entity.is_cancelled === 0">
-        <div ng-if="row.entity.is_received === 0" class="label label-primary" translate>PURCHASES.STATUS.CONFIRMED</div>
+        <div ng-if="row.entity.is_received === 0 && row.entity.is_partially_received === 0" class="label label-primary" translate>PURCHASES.STATUS.CONFIRMED</div>
 
-        <div ng-if="row.entity.is_received === 1" class="label label-success" translate>PURCHASES.STATUS.RECEIVED</div>
+        <div ng-if="row.entity.is_received === 1 && row.entity.is_partially_received === 0" class="label label-success" translate>PURCHASES.STATUS.RECEIVED</div>
```
jniles, line 5: Can something be both received and partially received?

client/src/partials/stock/entry/entry.html
```diff
@@ -0,0 +1,145 @@
+<!-- fix style -->
+<style>
```
jniles: Why are these inline styles necessary?

client/src/partials/stock/entry/entry.js
```diff
@@ -0,0 +1,249 @@
...
+    try {
+      vm.Stock.store.data.forEach(function (item, index) {
+        item.inventory = findInventory(items[index].inventory_uuid);
+        item.unit_cost = items[index].unit_price;
+        item.quantity = items[index].balance;
+        item.cost = item.quantity * item.unit_cost;
+        configureItem(item);
+      });
+    } catch (err) {
+      Notify.errorHandler(err);
```
jniles: I don't think you need this catch - `Notify.errorHandler()` doesn't exist and so I assume this is never called.

However, in the rest of this module you should change `Notify.errorHandler()` to `Notify.handleError()`.

client/src/partials/stock/entry/entry.js
```diff
@@ -0,0 +1,249 @@
...
+    }
+  }
+
+  function findInventory(uuid) {
```
jniles: You can avoid this function function completely by making `inventories` into a `Store`.   You will then be able to do `inventories.get(uuid)` to return the appropriate record in `O(1)` time.

client/src/partials/stock/entry/entry.js
```diff
@@ -0,0 +1,249 @@
...
+          purchase_uuid    : vm.movement.entity.uuid,
+        };
+      }).concat(current);
+    }, []);
+
+    movement.lots = lots;
+
+    Stock.stocks.create(movement)
+    .then(function (document) {
+      Purchase.stockStatus(vm.movement.entity.uuid);
```
jniles: Why does this promise not have a `.then().catch()` chain?

client/src/partials/stock/entry/modals/findPurchase.modal.html
```diff
@@ -0,0 +1,55 @@
...
+          ng-click="$ctrl.toggleFilter()"
+          class="btn btn-sm btn-default"
+          ng-class="{ 'btn-info' : $ctrl.filterEnabled }">
+          <span class="fa fa-filter"></span>
+        </button>
+      </div>
+    </div>
+  </div>
+
+  <div class="modal-body clearfix">
```
jniles: You should only need clearfix if you are using `pull-right` or `pull-left`.  Do you need it here?

client/src/partials/stock/entry/modals/findPurchase.modal.html
```diff
@@ -0,0 +1,55 @@
...
+    <div class="row">
+      <div class="col-xs-12">
+        <div
+          id="PurchaseGrid"
+          ui-grid="$ctrl.gridOptions"
+          ui-grid-auto-resize
+          ui-grid-resize-columns
+          ui-grid-selection>
+        </div>
+        <bh-grid-loading-indicator
```
jniles: The loading indicator should go inside the grid.  That way it is correctly centered.

client/src/partials/stock/entry/modals/findPurchase.modal.html
```diff
@@ -0,0 +1,55 @@
+<form 
+  name="FindForm"
+  bh-submit="$ctrl.submit(FindForm)"
+  bh-form-defaults
+  novalidate>
+
+  <div class="modal-header space-between">
```
jniles, line 7: Does `space-between` work?

client/src/partials/stock/entry/modals/findPurchase.modal.html
```diff
@@ -0,0 +1,55 @@
...
+    <div class="headercrumb">
+      <i class="fa fa-search"></i> 
+      <span translate>FORM.LABELS.SEARCH</span> 
+      <span translate>STOCK.ENTRY_PURCHASE</span>
+    </div>
+
+    <div class="toolbar clearfix">
```
jniles: [Clearfix](http://learnlayout.com/clearfix.html) is considered a hack.  It some sometimes necessarily, but still ugly!

It is not necessary here - an easier way to achieve the same functionality is to use `text-right` and `<span>`s instead.  I think this should work:
```html
<div class="toolbar text-right">
  <span class="toolbar-item">
     <!-- your button here -->
   </span>
</div>
```

client/src/partials/stock/entry/modals/findPurchase.modal.js
```diff
@@ -0,0 +1,104 @@
...
+  .then(function (purchases) {
+    vm.gridOptions.data = purchases;
+  })
+  .catch(Notify.errorHandler);
+
+  // submit
+  function submit() {
+    if (!vm.selectedRow) { return; }
+
+    Purchase.stockBalance(vm.selectedRow.uuid)
```
jniles: For the loading indicator to work on the `bhLoadingButton`, you must `return` a promise.

This code should be `return Purchase.stockBalance()`

client/src/partials/stock/entry/modals/lots.modal.js
```diff
@@ -0,0 +1,142 @@
...
+  vm.inventory = { lots: [] };
+  vm.gridApi = {};
+
+  /* ======================= Grid configurations ============================ */
+  vm.gridOptions = {
+    appScopeProvider  : vm,
+    enableSorting     : false,
+    enableColumnMenus : false,
+    showColumnFooter  : true,
+    fastWatch         : true,
```
jniles, line 24: :+1:

client/src/partials/stock/entry/templates/quantity.tmpl.html
```diff
@@ -0,0 +1,3 @@
+<div class="ui-grid-cell-contents">
+  {{ row.entity.givenQuantity || row.entity.quantity }}
```
jniles, line 2: Why is this OR here?

client/src/partials/stock/exit/exit.html
```diff
@@ -0,0 +1,213 @@
...
+        </button>
+      </div>
+
+      <!--depot -->
+      <div class="col-xs-3">
+        <button class="btn-block panel panel-default segment" type="button" ng-class="{'bg-primary': StockCtrl.movement.exit_type === 'depot'}" ng-click="StockCtrl.selectExitType('depot')">
+          <div class="panel-body text-center">
+            <div class="ui lg ima-blue statistic">
+              <div class="value">
+                DEPOT
```
jniles: Should this be translated?

client/src/partials/stock/movements/registry.js
```diff
@@ -0,0 +1,206 @@
...
+  'SearchFilterFormatService', 'LanguageService', 'SessionService',
+  'FluxService', 'ReceiptModal', 'GridGroupingService',
+];
+
+/**
+ * Stock movements Controller
+ * This module is a registry page for stock movements
+ */
+function StockMovementsController($state, Stock, Notify,
+  uiGridConstants, $translate, Modal, SearchFilterFormat,
```
jniles: I don't think you use `$state` here.  We coul dalso check to see if other imports are not used

client/src/partials/stock/lots/registry.js
```diff
@@ -0,0 +1,135 @@
...
+  'uiGridConstants', '$translate', 'StockModalService',
+  'SearchFilterFormatService', 'LanguageService',
+  'GridGroupingService',
+];
+
+/**
+ * Stock lots Controller
+ * This module is a registry page for stock lots
+ */
+function StockLotsController($state, Stock, Notify,
```
jniles: I do not see where you used `$state` here.  If it isn't used, it can be removed.

client/src/partials/stock/lots/registry.js
```diff
@@ -0,0 +1,135 @@
...
+    { field: 'delay_expiration', displayName: 'STOCK.EXPIRATION', headerCellFilter: 'translate' },
+  ];
+
+  // options for the UI grid
+  vm.gridOptions = {
+    appScopeProvider  : vm,
+    enableColumnMenus : false,
+    columnDefs        : columns,
+    enableSorting     : true,
+    showColumnFooter  : true,
```
jniles, line 70: Can this grid use the performance improvements of `fastWatch` and `flatEntityAccess`?

client/src/partials/stock/movements/modals/search.modal.html
```diff
@@ -0,0 +1,114 @@
...
+      </bh-date-interval>
+    </fieldset>
+
+  </div>
+
+  <div class="modal-footer">
+    <button type="button" class="btn btn-default" ng-click="$ctrl.cancel()" data-method="cancel" translate> 
+      FORM.BUTTONS.CLOSE
+    </button>
+    <bh-loading-button loading-state="ModalForm.$loading">
```
jniles: The `bhLoadingButton` requires `bhSubmit` to work.  It also requires a Promise to be returned from the controller.  Like this:

```js
// in the controller
vm.submit = function () {
  return $http.post(/* ... */);
}
```
```html
<!-- in the HTML -->
<form name="XYZ" bh-submit="$ctrl.submit()">
  <bh-loading-button loading-state="XYZ.$loading">
   I will load!
  </bh-loading-button>
</form>
```

It will be useful to take a look at the code for `bhLoadingButton`.  [Here it is](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/components/bhLoadingButton.js).

And for [bhSubmit](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/directives/bhSubmit.js).

client/src/partials/stock/movements/modals/search.modal.js
```diff
@@ -0,0 +1,66 @@
...
+  .then(function (inventories) {
+      vm.inventories = inventories;
+  })
+  .catch(Notify.handleError);
+
+  // load flux 
+  Flux.read()
+  .then(function (rows) {
+      vm.fluxes = rows.map(function (row) {
+        row.label = $translate.instant(row.label);
```
jniles, line 42: I propose that you move this `$translate.instant()` into the `FluxService`.  Chance are, other controller will want to do this too.

client/src/partials/stock/movements/registry.html
```diff
@@ -0,0 +1,97 @@
...
+    <i class="fa fa-ban text-danger"></i>
+    {{ "FORM.INFO.CLEAR_FILTERS" | translate }}
+  </a>
+</div>
+
+<!--the grid -->
+<div class="flex-content">
+  <div class="container-fluid">
+
+    <div
```
jniles, line 87: This grid could use a `bhGridLoadingIndicator`

client/src/partials/stock/movements/registry.html
```diff
@@ -0,0 +1,97 @@
...
+  </bh-filters-applied>
+
+  <a
+    href
+    ng-click="StockCtrl.clearFilters()"
+    ng-show="StockCtrl.formatedFilters.length > 0"
+    class="text-danger"
+    data-method="clear">
+    <i class="fa fa-ban text-danger"></i>
+    {{ "FORM.INFO.CLEAR_FILTERS" | translate }}
```
jniles: These filters should use the `translate` directive.

client/src/partials/stock/movements/registry.js
```diff
@@ -0,0 +1,206 @@
...
+      cellTemplate     : 'partials/stock/movements/templates/flux.cell.html' },
+    { field           : 'action',
+      displayName     : '',
+      enableFiltering : false,
+      enableSorting   : false,
+      cellTemplate    : 'partials/stock/movements/templates/action.cell.html' },
+  ];
+
+  // options for the UI grid
+  vm.gridOptions = {
```
jniles, line 114: Can this grid take advantage of the performance options `flatEntityAccess` and `fastWatch`?

client/src/partials/stock/movements/registry.js
```diff
@@ -0,0 +1,206 @@
...
+    Stock.movements.read(null, filters || defaultDate).then((rows) => {
+      vm.gridOptions.data = rows;
+    })
+    .catch(Notify.handleError);
+  }
+
+  // search modal
+  function search() {
+    Modal.openSearchMovements()
+    .then((filters) => {
```
jniles: You cannot use fat arrows on the client.  It will break our `gulp` build in production.

client/src/partials/stock/inventories/templates/status.cell.html
```diff
@@ -0,0 +1,21 @@
+<div class="ui-grid-cell-contents" translate>
+    <div ng-if="row.entity.status === 'sold_out'" class="label label-danger" translate>
```
jniles: All of these `===` will be evaluated on every `$digest()` loop.  It is much better to cache these in a variable.  Like this:
```js
// in controller
function setStatusFlag(item) {
  item.isSoldOut = item.status === 'sold_out';
  item.isInStock = item.status === 'in_stock';
  item.hasSecurityWarning = item.status === 'security_reached';
  item.hasMinimumWarning = item.status === 'minimum_reached';
  item.hasOverageWarning = item.status === 'over_maximum';
}
```

This only needs to be done once on data load and then the template can just use `ng-if="row.entity.hasOverageWarning"` to check a boolean.

jniles: It also would be good to store the 'sold_out', 'in_stock', etc, flags in `bhConstants` in the future.

client/src/partials/stock/lots/modals/search.modal.html
```diff
@@ -0,0 +1,77 @@
...
+    </ol>
+  </div>
+
+  <div class="modal-body">
+
+    <!-- depot -->
+    <div class="form-group">
+      <label class="control-label" translate>STOCK.DEPOT</label>
+
+      <ui-select name="depot" ng-model="$ctrl.bundle.depot_uuid" theme="bootstrap">
```
jniles: Just a note: there is no need to use `theme="bootstrap"`.  We set it globally in `app.js`.

client/src/partials/stock/lots/registry.html
```diff
@@ -0,0 +1,96 @@
...
+    <i class="fa fa-ban text-danger"></i>
+    {{ "FORM.INFO.CLEAR_FILTERS" | translate }}
+  </a>
+</div>
+
+<!--the grid -->
+<div class="flex-content">
+  <div class="container-fluid">
+
+    <div
```
jniles, line 86: This could use a `bhGridLoadingIndicator`.

jniles (approved):
@mbayopanda this is a massive PR.

I'm sure I won't catch all the bugs, so I'll stop my review here.

Some things to look out for:
 1. Always use the `translate` directive.
 2. Use `flatEntityAccess` and `fastWatch` where possible.
 3. All registries should have `bhGridLoadingIndicator`s to provide a better UX on slow connections
 4. Make sure that you use `bhLoadingButton` correctly - it always needs `bhSubmit` on the form and a promise returned by the `submit()` method of the controller.

Let's get this in :+1:  It represents a lot of work!

server/controllers/stock/core.js
```diff
@@ -0,0 +1,344 @@
...
+
+  let queryExpiration;
+  let paramExpiration;
+  let queryEntry;
+  let paramEntry;
+  const queryArray = [];
+  const paramArray = [];
+
+  if (params.uuid) {
+    params['l.uuid'] = params.uuid;
```
sfount: I think `FilterParser` can save you ~200 lines of code here. 

I have not looked at this in depth but this is an idea of how it could be used: 

```js
const filters = new FilterParser(params);

filters.equals('uuid', 'uuid', 'l');

// this could be filters.fullText() to allow the search to be on any part of the string
filters.equals('depot_text', 'text', 'd');
filters.equals('text', 'text', 'i');
filters.equals('label', 'label', 'l');
filters.dateFrom('expiration_date_from', 'expiration_date', 'l');
filters.dateTo('expiration_date_to', 'expiration_date', 'l');
filters.dateFrom('entry_date_from', 'entry_date', 'l');
filters.dateTo('entry_date_to', 'entry_date', 'l');
filters.dateFrom('dateFrom', 'date', 'm');
filters.dateTo('dateTo', 'date', 'm');
```

mbayopanda:
:+1

2017-03-16 5:42 GMT-07:00 Steven Fountain <notifications@github.com>:

client/src/js/components/bhDateInterval.js
mbayopanda: It is for case where we have multiple `<bh-date-interval>` component in the same view for having a mean to perform tests on them.

client/src/js/services/receipts/ReceiptModal.js
mbayopanda: 👍 THANKS

client/src/js/components/bhDateInterval.js
jniles: I think you can just put an ID on the component.  Like this:
```html
<bh-date-interval id="hello-world"></bh-date-interval>
```
Since it doesn't affect the logic of the controller, it doesn't make sense to record a binding for it.

client/src/partials/stock/entry/entry.html
mbayopanda: They define the style of source buttons. We can move them into a specific style file

client/src/partials/stock/entry/entry.js
mbayopanda: 👍

client/src/partials/stock/entry/modals/findPurchase.modal.html
mbayopanda: yes

client/src/partials/stock/entry/templates/quantity.tmpl.html
mbayopanda: The quantity can come from a purchase order as default quantity or it can be calculate by quantities of lots

client/src/partials/stock/movements/modals/search.modal.js
mbayopanda: `$translate.instant` in the service FluxService doesn't translate keys

client/src/css/structure.css
```diff
@@ -807,6 +807,12 @@ growl-notification.fading.ng-leave.ng-leave-active {
   margin-left: 18px;
 }
 
+.text-ellipsis {
```
jniles, line 810: :+1:  This is a good addition.

client/src/js/components/bhDatePicker.js
```diff
@@ -0,0 +1,80 @@
+angular.module('bhima.components')
+.component('bhDatePicker', {
+  templateUrl : '/partials/templates/bhDatePickerAction.tmpl.html',
+  controller  : DatePickerController,
+  bindings    : {
+    date     : '=',
+    format   : '<',
+    mode     : '<',
+    required : '<',
+    onChange : '<',
```
jniles: @mbayopanda, many of these components do not properly set their inputs and outputs.  They rely on changing objects bound to one-way bindings.  This is both confusing and inefficient.

[Here is the Angular component documentation](https://docs.angularjs.org/guide/component#component-based-application-architecture).  They want all inputs to be either `<` or `@` (the `=` binding is less efficient) and all outputs to be _callbacks_ using the `&` binding.

Here's how it looks applied to your component:
```js
  .component('bhDatePicker', {
    templateUrl : '/some/template.tmpl.html',
    controller : DatePickerController,
    bindings : {
       date : '<',  // set the date once as the initial date and use callbacks to change it later
       format : '<',
       mode : '@', // will this ever change?  If so, we can use '<'
       required : '<',
       onChange : '&', // use a callback to notify for changes
    }
  });

function DatePickerController(Modal) {
  var $ctrl = this;

  // later on...
  function open() {
    openDatePicker({ mode : vm.mode })
      .then(function (res) {
        // notify the parent controller of a date change via a callback
       $ctrl.onChange({ date : res });
      });
}
```
Your HTML can then look like this:
```html
<bh-date-picker
  date="ParentCtrl.date"
  on-change="ParentController.onDateChangeCallback(date)">
</bh-date-picker>
```
And you parent controller:
```js
// in parent controller
vm.onDateChangeCallback = onDateChangeCallback;

// update the date when the component records a date change
function onDateChangeCallback(date) {
  vm.date = date;
}
```

One thing to notice that is tricky:
 1. If you use the following syntax in your HTML:
 ```html
<component-example
  on-change="controller.fn(paramA, paramB, paramC)">
</component-example>
```
 You _must_ use an object to set those parameters within you component.  The syntax is `this.fun({ paramA : 'hello', paramB : 'world', paramC : null });`.

 2.

File: client/src/js/components/bhDepotDropdown.js [hunk 32]
```diff
@@ -0,0 +1,37 @@
+angular.module('bhima.components')
+  .component('bhDepotDropdown', {
+    bindings : {
+      onSelect : '<',
```
jniles: Just like the comment above, these bindings should be `&`.  An example of this is the [bhCurrencySelect](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/components/bhCurrencySelect.js)

client/src/js/components/bhDepotDropdown.js
```diff
@@ -0,0 +1,37 @@
...
+
+  $ctrl.$onInit = function $onInit() {
+    $ctrl.loading = true;
+
+    Depot.read()
+    .then(function (rows) {
+      if (!rows.length) { return; }
+      $ctrl.depots = rows;
+      $ctrl.selection = cache.selection || $ctrl.depots[0];
+      $ctrl.loading = false;
```
jniles, line 26: By this logic, if the array of depots is 0 length, the component will keep loading for ever instead of reporting a problem.  

It's always a good idea to turn the $loading state off when the controller is no longer loading.  My preferred way of handling this is with [the Promise.finally() method](http://stackoverflow.com/questions/26030992/whats-the-difference-between-q-promise-librarys-finally-and-done).  Like this:

```js
Depot.read()
  .then(function () { 
    // successfully got none or more depots
  })
  .catch(function () {
     // HTTP Error, API Error, Database error ...
  })
  .finally(function () {
    $ctrl.loading = false;  // regardless of the result, we no longer need to pretend we are waiting for a result.
  });
```

client/src/js/services/SearchFilterFormatService.js
```diff
@@ -108,6 +121,11 @@ function SearchFilterFormatService() {
 
       if (angular.isDefined(value)) {
         column.value = value;
+
+        // @FIXME tempoarary hack for default period
```
jniles, line 125: :+1: Good comment

client/src/partials/stock/StockForm.service.js
```diff
@@ -0,0 +1,165 @@
...
+   * @param {Number} n - the number of items to add to the grid
+   */
+  StockForm.prototype.addItems = function addItems(n) {
+    var elt;
+    var idx;
+    var self = this;
+    while (n--) {
+      elt = new StockItem();
+      idx = self.store.data.length;
+      elt.index = idx ? idx + 1 : idx;
```
jniles, line 60: This isn't clear to me what happens when `idx` is 0.  Is `elt.index` always 0?

client/src/partials/stock/StockForm.service.js
```diff
@@ -0,0 +1,165 @@
...
+   * Checks to see if the invoice has cached items to recover.
+   */
+  StockForm.prototype.hasCacheAvailable =  function hasCacheAvailable() {
+    return Object.keys(this.cache).length > 0;
+  };
+
+  /**
+   * @class StockItem
+   */
+  function StockItem() {
```
jniles: It would be a good idea to put this in its own service.  It would make this file smaller and more focused.

client/src/partials/stock/adjustment/adjustment.html
```diff
@@ -0,0 +1,154 @@
...
+            required>
+          </textarea>
+          <div class="help-block" ng-messages="StockForm.description.$error" ng-show="StockForm.$submitted">
+            <div ng-messages-include="partials/templates/messages.tmpl.html"></div>
+          </div>
+        </div>
+      </div>
+
+      <div class="col-xs-6">
+        <div class=""  ng-if="StockCtrl.adjustmentOption === 'increase'" style="margin-top: 0px;">
```
jniles: Are these empty classes supposed to have something in them?

client/src/partials/stock/adjustment/adjustment.html
```diff
@@ -0,0 +1,154 @@
...
+  </div>
+</div>
+
+<!-- content -->
+<div class="flex-content">
+  <div class="container-fluid">
+
+    <form name="StockForm" bh-submit="StockCtrl.submit(StockForm)" bh-form-defaults novalidate>
+
+    <!-- date and note -->
```
jniles: Is this comment correct?  I don't see a note.

client/src/partials/stock/entry/entry.html
jniles: They should probably be put in the `less` file so that you can use the variables to set `.bg-primary`.

Did you try [the helper classes with bootstrap](http://getbootstrap.com/css/#helper-classes-backgrounds) before writing your own?

client/src/partials/stock/entry/entry.html
```diff
@@ -0,0 +1,187 @@
...
+                  {{ StockCtrl.movement.entity.instance.reference }}
+                </span>
+              </div>
+            </div>
+          </div>
+        </button>
+      </div>
+
+      <!-- stock from integration -->
+      <div class="col-md-4 col-xs-12">
```
jniles, line 68: :+1: Nice viewport scaling!

client/src/partials/stock/entry/entry.js
```diff
@@ -0,0 +1,281 @@
...
+ * @description This controller is responsible to handle stock entry module
+ */
+function StockEntryController(Depots, Inventory, Notify,
+  Session, util, bhConstants, ReceiptModal, Purchase,
+  StockForm, Stock, StockModal, uiGridGroupingConstants, uiGridConstants, Store) {
+  var vm = this;
+  var inventoryStore;
+
+  var mapEntry = {
+    purchase    : { find: findPurchase, submit: submitPurchase },
```
jniles, line 23: I assume these `noop()`s are just placeholders?  It would be good to write a comment so if so!

client/src/partials/stock/entry/modals/templates/lot.input.tmpl.html
```diff
@@ -0,0 +1,8 @@
+<div class="ui-grid-cell-contents">
+  <input type="text" 
+    class="form-control"
+    ng-model="row.entity.lot"
+    ng-model-options="{ 'debounce' : { 'default' : 150, 'blur' : 0 }}"
```
jniles, line 5: :+1:

client/src/partials/stock/movements/templates/flux.cell.html
```diff
@@ -0,0 +1,3 @@
+<div class="ui-grid-cell-contents" translate>
+    {{ grid.appScope.getFluxName(row.entity.flux_id) }}
```
jniles: Since this function will be called for every row in the grid anytime the grid changes ... can you simply map the fluxName to a property on the entity?  Like `row.entity._flux_name`?  If you need to submit it to the server you can then remove it in one loop.

client/src/partials/stock/origins/modals/search.modal.html
```diff
@@ -0,0 +1,77 @@
...
+
+    <!-- depot -->
+    <div class="form-group">
+      <label class="control-label" translate>STOCK.DEPOT</label>
+
+      <ui-select name="depot" ng-model="$ctrl.bundle.depot_uuid">
+        <ui-select-match>
+          <span>{{$select.selected.text}}</span>
+        </ui-select-match>
+        <ui-select-choices ui-select-focus-patch repeat="depot as depot in $ctrl.depots | filter:$select.search">
```
jniles: Just to let you know, `filter:$select.search` will search all keys/values in the `depot` object ... for each object in the list `$ctrl.depots`.  Since you are only showing `$select.selected.text`, the user will probably only want to search for the `text` name.

This can be made more efficient by using `filter:{ 'text' : $select.search }`.  For depots, it doesn't make a big difference, but for larger arrays (accounts, inventory, etc) it will make a large difference.  It's a good habit to develop :)

client/src/partials/stock/origins/modals/search.modal.html
```diff
@@ -0,0 +1,77 @@
...
+
+    <!-- inventory -->
+    <div class="form-group">
+      <label class="control-label" translate>STOCK.INVENTORY</label>
+
+      <ui-select name="inventory" ng-model="$ctrl.bundle.inventory_uuid">
+        <ui-select-match>
+          <span>{{$select.selected.label}}</span>
+        </ui-select-match>
+        <ui-select-choices ui-select-focus-patch repeat="inventory as inventory in $ctrl.inventories | filter:$select.search">
```
jniles: See the comment above.

client/src/partials/stock/origins/registry.html
```diff
@@ -0,0 +1,43 @@
...
+  </bh-filters-applied>
+
+  <a
+    href
+    ng-click="StockOriginCtrl.clearFilters()"
+    ng-show="StockOriginCtrl.formatedFilters.length > 0"
+    class="text-danger"
+    data-method="clear">
+    <i class="fa fa-ban text-danger"></i>
+    {{ "FORM.INFO.CLEAR_FILTERS" | translate }}
```
jniles: This could use the `translate` directive.

client/src/partials/stock/origins/registry.js
```diff
@@ -0,0 +1,170 @@
...
+    {
+      field           : 'action',
+      displayName     : '',
+      cellTemplate    : '/partials/stock/origins/templates/action.tmpl.html',
+      enableFiltering : false,
+      enableSorting   : false },
+  ];
+
+  // options for the UI grid
+  vm.gridOptions = {
```
jniles, line 83: Could this grid benefit from `fastWatch` or `flatEntityAccess`?

client/src/partials/stock/origins/registry.js
```diff
@@ -0,0 +1,170 @@
...
+    appScopeProvider  : vm,
+    enableColumnMenus : false,
+    columnDefs        : columns,
+    enableSorting     : true,
+    showColumnFooter  : true,
+  };
+
+  vm.grouping = new Grouping(vm.gridOptions, true, 'reference', vm.grouped, true);
```
jniles: Oh man, this API is getting out of hand...

client/src/partials/templates/bhDatePickerAction.tmpl.html
```diff
@@ -0,0 +1,10 @@
+<span class="text-action" ng-dblclick="$ctrl.open()">
+  <input 
+    uib-datepicker-popup="{{ $ctrl.format || 'dd/MM/yyyy' }}"
```
jniles: This default should happen in the controller and it should default to the date format in `bhConstants`.

package.json
```diff
@@ -14,6 +14,7 @@
     "build": "./node_modules/.bin/gulp build",
     "app_windows": "gulp build && cd bin && set NODE_ENV=production && node server/app.js",
     "dev_windows": "gulp build && cd bin && set NODE_ENV=development && node server/app.js",
+    "live_dev_windows": "gulp build && cd bin && set NODE_ENV=development && nodemon server/app.js",
```
jniles: Ha!

server/lib/util.js
```diff
@@ -210,3 +212,18 @@ function resolveObject(object) {
...
+ * @description returns a formated date string 
+ * 
+ * @param {date} date - A date 
+ * 
+ * @param {string} format - A date format 
+ * 
+ * @return {string} - A date string 
+ */
+function dateString(date, format) {
+  return moment(new Date(date)).format(format || 'YYYY-MM-DD').toString();
```
jniles: `moment.format()` returns a string... you don't need the extra `toString()` call.

jniles (commented):
I finally managed to do a full review (except the tests ... those will have to be later).

Overall, you've done a very good job of keeping up with changing standards!  Some things I noticed:
 1. The `translate` directive is preferred over `translate` filter :+1:
 2. Excellent use of filter parser :+1:
 3. You've made good use of bhConstants :+1:
 4. Components where possible :+1:
 5. Many of the `ui-grid`s have performance improvements of `fastWatch` and `flatEntityAccess`. :+1:
 6. Services are logically separated from controller code to manage validation and initialization.  :+1:

Some things that can be improved upon in the future:
 1. Moving custom styles into less.
 2. Using callback for components (`&`) instead of data-binding and dirty checking.
 3. Breaking services into smaller files to help test them, improve performance (by using prototypes) and clarify their responsibility for other developers.

You can choose to work on these comments here or have this merged at some point and the team can work on refactoring them as a sprint.

:+1:

client/src/partials/stock/StockForm.service.js
mbayopanda: `idx` is 0 when there is no data in the `StockForm.store.data`, so `elt` is the new element to add, it must have index 0 as the first element of `StockForm.store.data`.

client/src/partials/stock/origins/modals/search.modal.html
mbayopanda: 👍

package.json
mbayopanda: Hoops... Doesn't work well and it must be removed

sfount:
My comments here on use of `FilterParser` have been completely addressed :+1:. I will have a final look at the current state of this PR today.

sfount (approved):
I agree with @jniles that this pull request should be merged in, this represents a lot of great work. 

I have made a number of notes on potential performance difficulties with queries (joins, indexes, limiting required information). I also agree that the components should use the latest callback standard as we move forward. However these are large bodies of work in themselves and would be better tackled as issues on top of this solid base. 

:+1:

jniles:
@sfount, thanks for the approval.  Merging.

jniles:
@mbayopanda great work!

PR #1356, opened 2017-03-15
Loading patient financial

### LINKED ISSUE #1141: (refactor) Patient Financial Situation needs a loading indicator
The Patient Financial Situation panel needs a loading indicator.  It takes time to scan the `combined_ledger` and this is reflected by an empty financial activity panel until loaded. See below:

financialsituationloading
_Fig 1: Loading ..._

financialsituationloaded
_Fig 2: Loaded_

lomamech:
- Adding loading indicator in Patient Financial Situation
- Adding directive loading

closes #1141

### DIFF
--- a/client/src/js/components/bhLatestInvoice.js
+++ b/client/src/js/components/bhLatestInvoice.js
@@ -41,6 +41,9 @@ function LatestInvoice(Patient, moment, Notify, Session) {
       .then(function (balance) {
         vm.patientBalance = balance;
       })
-      .catch(Notify.handleError);
+      .catch(Notify.handleError)
+      .finally(function () {
+        vm.loading = false;
+      });      
   }
 }
--- a/client/src/js/components/bhLatestInvoice.js
+++ b/client/src/js/components/bhLatestInvoice.js
@@ -27,8 +27,11 @@ function LatestInvoice(Patient, moment, Notify, Session) {
   /** getting patient document */
   function startup() {
     if (!vm.debtorUuid) { return; }
+    vm.loading = true;
+
     Patient.latest(vm.debtorUuid)
       .then(function (patientInvoice) {
+        vm.loading = false;
         vm.patientInvoice = patientInvoice;
         vm.patientInvoice.durationDays = moment().diff(vm.patientInvoice.date, 'days');
       })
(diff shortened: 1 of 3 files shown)


client/src/partials/templates/bhLatestInvoice.tmpl.html
```diff
@@ -14,8 +14,11 @@
     </span>
   </div>
   <div class="panel-body">
+    <div ng-if="$ctrl.loading" class="text-center alert alert-info fade in">
```
jniles: I wouldn't use an alert here.  An `alert` should be very specific to get a user's attention, but this is just trying to mask a panel while it loads.

Check out the styles for [grid-msg-overlay](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/css/ui-grid.css#L20).  I think it does a good job of hiding the underlying contents without grabbing the user's attention.

client/src/js/components/bhLatestInvoice.js
```diff
@@ -27,8 +27,11 @@ function LatestInvoice(Patient, moment, Notify, Session) {
   /** getting patient document */
   function startup() {
     if (!vm.debtorUuid) { return; }
+    vm.loading = true;
+
     Patient.latest(vm.debtorUuid)
       .then(function (patientInvoice) {
+        vm.loading = false;
```
jniles, line 34: It would be better to put the `vm.loading = false` call in the `.finally()` method of the promise.  That way, even if the query errors, the loading indicator will be removed.

jniles (commented):
@lomamech, this is a good start!

A loading indicator should mask the underlying data without drawing a user's attention to much.  The `alert` that you used is very noticeable.  It would be a better user experience to use something more muted, like the `bhGridLoadingIndicator`.

jniles:
Here is an example of the grid loading indicator, for reference.  Notice how the underlying data is slightly visible.

exampleloadingindicator
_Fig 1: Grid Loading Indicator_

jniles:
@lomamech, what is the state of this PR?

client/src/js/components/bhLatestInvoice.js
```diff
@@ -38,6 +41,9 @@ function LatestInvoice(Patient, moment, Notify, Session) {
       .then(function (balance) {
         vm.patientBalance = balance;
       })
-      .catch(Notify.handleError);
+      .catch(Notify.handleError)
+      .finally(function () {
```
jniles, line 45: :+1:

jniles (approved):
LGTM

Issue #1347, opened 2017-03-13
(feature) Client side service for registry filters - FilterService


sfount:
Currently all of our registries implement custom search modals and controller logic. The leads to each registry being very difficult to maintain or fix. 

On the server we have a library called `FilterParser` that receives filters from the client and translates them into valid SQL. 

I propose that we add a client side service for interfacing with the library - a service that is used anywhere that filters are used and interfaces with the `bhFiltersApplied` component and the server `FilterParser`. 

It would be responsible for: 
* Defining a standard filter object that contains the information required by everything interfaced with (key, label, isDefault, timestamp, value, viewValue ngFilter)
* Formatting key/ value pair requests to be send to the server 
* Formatting viewValue's into a labels object (Array) to send to the server 
* Uniformly caching filters for each individual page
* Exposing the state of filters, number of filters, clear functionality, etc.

The `Filter` entity itself could be created with a constructor like:
```js
class Filter() { 

constructor (key, value, label=value, isDefault=false) { //... }
}
```

jniles:
@sfount, can this be closed as implemented in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1566?

sfount:
@jniles good catch!

Proposal closed with https://github.com/IMA-WorldHealth/bhima-2.X/pull/1566

PR #1360, opened 2017-03-16
fix(filters) Default cash registry filter by period (Today) 

sfount:
This commit updates the cash registry to include the hot fix patch
introduced in #1339. Only the invoice registry was limited to show daily
records however this is an issue with all commonly used registries.

This commit introduces temporary hot fix code that should be superseeded
by #1347.

This branch should be pulled and manually tested before merging with
master.


jniles:
Seems to work well in practice!

PR #1379, opened 2017-03-20
fix(invoice): disable buttons on submit

### LINKED ISSUE #1366: (refactor) Patient Invoice Form needs loading button
The Patient Invoice Form does not have a `bhLoadingButton`.  This causes users to click the button multiple times and query the database several times for the same invoice payment.  Most other forms use a loading button to both provide feedback to the user and to provide security by disabling submission.  A `bhLoadingButton` should be placed on the Patient Invoice module as soon as possible.

jniles:
This commit refactors the Patient Invoice module to disable the submit
button and clear button when the user submits an invoice.  It should
help reduce problems due to users absent-mindedly overloading the
database.

Closes #1366.

### DIFF
--- a/client/src/partials/patient_invoice/patientInvoice.html
+++ b/client/src/partials/patient_invoice/patientInvoice.html
@@ -11,237 +11,240 @@
 <div class="flex-content">
   <div class="container-fluid">
 
-    <div class="row">
-      <div class="col-xs-12">
-        <div class="panel panel-default">
-          <div class="panel-body">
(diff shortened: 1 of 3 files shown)


jniles:
This diff makes it hard to see what changed.

 1. I added a the `bh-submit` directive to the details form.
 2. I moved the details form higher in the DOM tree.
 3. I added `ng-disabled` to the clear button, listening on the `$loading` value.
 4. I added a `bh-loading-button` instead of a submit button.

PR #1381, opened 2017-03-20
Report multiple currencies

### LINKED ISSUE #1132: (feature request) Print reports in multiple currencies
We have a request from one of our clients to produce the financial reports (bilan, balance sheet, etc) in the multiple currencies based on the exchange rate of today.

I propose we repurpose the `bhReceiptCurrency` component to control the currencies rendered for all reports and receipts globally across the application where it makes sense to show multiple currencies.

lomamech:
Print Report in Multiple Currencies
- Income Expense
- Account
- Aged Debtors
- Cashflow
- Balance
- Cash flow by Service
- Open Debtor

closes #1132

### DIFF
--- a/client/src/i18n/fr/report.json
+++ b/client/src/i18n/fr/report.json
@@ -12,7 +12,7 @@
 "CASHFLOW":"Cashflow",
 "CASHFLOW_BY_SERVICE":"Cashflow par Service",
 "CHART_OF_ACCOUNTS":"Plan Comptable",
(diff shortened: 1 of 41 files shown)


jniles:
@lomamech, I noticed this uses a lot of `combined_ledger` queries.   It's clear the `combined_ledger` was a bad idea at this point.  See #1409 for more details.

Instead of addressing all those in this PR, I suggest we include them in our sprint on reports.

Issue #1373, opened 2017-03-20
(performance) Reduce huge query on searching for patient names (Query Rank #4)


sfount:
This is a profile of the most expensive queries currently running on an installation database. 
```
# Profile                                                                                                                                   
# Rank Query ID           Response time   Calls R/Call   V/M   Item                                                                         
# ==== ================== =============== ===== ======== ===== ===========                                                                  
#    1 0x86863BB0CF7F56D2 4834.8091 34.2%   106  45.6114 22.87 CALL CalculateCashInvoiceBalances                                            
#    2 0x172C9AD568506549 2271.6316 16.1%    17 133.6254 15.85 SELECT cash voucher project debtor patient cash_box user                     
#    3 0x7146126EBA8A45BB 1906.9753 13.5%    78  24.4484 13.47 CALL PostInvoice                                                             
#    4 0x79B3A6B3AFBA6A34 1073.9583  7.6%  1059   1.0141  0.19 SELECT patient project debtor debtor_group village sector user patient_visit 
#    5 0x70FBFAE700A0D0C7  879.2873  6.2%   106   8.2952 82.17 CALL PostCash                                                                
#    6 0x9A1C8B26FB9FDA98  737.8403  5.2%    95   7.7667  0.13 SELECT UNION combined_ledger invoice project                                 
#    7 0x006CA608F6F95299  665.7194  4.7%   165   4.0347  0.28 SELECT combined_ledger document_map                                          
#    8 0x3183E320043295DC  445.9682  3.2%   116   3.8446  0.06 SELECT combined_ledger                                                       
#    9 0x6F6D9FDC88980472  229.7403  1.6%     2 114.8702  2.17 SELECT combined_ledger invoice voucher project                               
#   10 0x4E40D93B961475C4  228.1232  1.6%    79   2.8876 28.06 CALL WriteInvoice                                                            
#   11 0x04BEBBE2B4CDCF36  196.6480  1.4%    49   4.0132  0.14 SELECT combined_ledger                                                       
#   12 0xB97BDB4EDF4B0656  179.1795  1.3%    12  14.9316  0.04 SELECT debtor_group debtor combined_ledger                                   
#   13 0xAAA3A0523A8998D1  176.1755  1.2%    12  14.6813  0.05 SELECT debtor_group debtor combined_ledger account                           
# MISC 0xMISC              309.3979  2.2%  2030   0.1524   0.0 <58 ITEMS>  
```

[...]

PR #1383, opened 2017-03-20
refactor(search) Add name search, minimal repsonse

sfount:
This commit introduces the patients/search/name API. This can be used to
search for a patient by name. It only returns the very minimal data
required to represent a patient, UUID, display name and reference.

This has been applied to the find patient directive when find by name is
selected, drastically reducing the complexity of the query to find a
patient by name.

Requirements: 
- [x] Implement `patients/search/name` route, minimal response for components 
- [x] Use new route in FindPatient directive 
- [ ] Integration tests for `patients/search/name` to work against future regressions 

**patients/search?display_name=nzala**
performance_after

**patients/search/name?display_name=nzala**
performance_before

The key metric to look at here is 'Waiting', the time it took for the server to respond with the required results. 
Full patient search route : **98.57ms**
Proposed patient search route : **17.49ms**

This response is significantly faster than doing a search on all patient information, on capable machines the difference is negligible however this can be crippling on lower end server hardware (as described https://github.com/IMA-WorldHealth/bhima-2.X/issues/1373)

Contributes to https://github.com/IMA-WorldHealth/bhima-2.X/issues/1373

### DIFF
--- a/client/src/js/components/bhFindPatient.js
+++ b/client/src/js/components/bhFindPatient.js
@@ -154,16 +154,7 @@ function FindPatientComponent(Patients, AppCache, Notify, SessionService, bhCons
       limit        : LIMIT,
     };
 
-    return Patients.search(options)
-      .then(function (patients) {
-
-        // loop through each
-        patients.forEach(function (patient) {
-          patient.label = formatPatient(patient);
-        });
-
-        return patients;
-      });
+    return Patients.searchByName(options);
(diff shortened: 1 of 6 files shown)


jniles:
This passes tests, so I'm merging it.  The benefits in production outweigh the immediate need for regression tests.  I'll make an issue tracking this (#1391).

PR #1385, opened 2017-03-21
Fix patient registry default records by entry date

mbayopanda:
This PR fix the issue #1377 

Also, in this PR these files are updated by comment all `.skip` function for avoiding error with Protractor
- end-to-end/journal/journal.config.js
- end-to-end/journal/trial_balance/trialBalance.config.js
- end-to-end/patient/invoice/registry.search.js


jniles:
Works as expected!

PR #1398, opened 2017-03-22
test(search) regression tests for name search

sfount:
This commit adds a basic integration test to guard against regressions
on the /patients/search/name API.


jniles (approved):
LGTM!

PR #1404, opened 2017-03-22
test(search) remove redundant find patient tests

sfount:
This commit resolves the current CI issues by removing tests that test
against old feature sets. These tests can be updated in the future.


sfount:
Merging this to give us CI for essential performance fixes.

Issue #1405, opened 2017-03-22
(performance) Remove all instances of `combined_ledger` in frequently used modules


sfount:
As researched in https://github.com/IMA-WorldHealth/bhima-2.X/issues/1388 and proven by https://github.com/IMA-WorldHealth/bhima-2.X/pull/1403, any queries that are built around the `combined_ledger` are extremely non-optimal, they are unacceptably slow on the standard hardware currently used at installations. 

In order to run the server on target hardware it is necessary to optimise all of these queries that are frequently used.
For a pull request to be accepted and close an entry in this issue it must demonstrate that the proposed query significantly addresses the issue demonstrated here. (For example https://github.com/IMA-WorldHealth/bhima-2.X/pull/1403#issuecomment-288388256 and https://github.com/IMA-WorldHealth/bhima-2.X/pull/1399#issuecomment-288383499)

These are ranked in order of highest to lowest and can be assigned to anyone who is free to work on this. 

- [x] **(Urgent)** `server/controllers/finance/debtors/index.js` **function invoiceBalances** - this method uses a `UNION` on two queries on the combined ledger and is called every time a cash OR invoice receipt is shown (@sfount)
- [x] **(Urgent)** `server/controllers/finance/debtors/index.js` **function balance** - this method is called every time a cash receipt is shown (@sfount)

- [x] `/server/controllers/finance/debtors/groups/index.js` **function loadInvoices()**
- [x] `/server/controllers/finance/debtors/index.js` **function financialActivity()**
- [x] `/server/controllers/medical/patients/index.js` **function latestInvoice()**

I have omitted reports for the time being as they are optional however they should be added to this list as progress is made.

jniles:
I'll take the `loadInvoices()`.

jniles:
The two urgent queries result in this performance in production:
![cashpayment](https://cloud.githubusercontent.com/assets/896472/24235871/7f78b902-0f9f-11e7-86f4-26484adc383d.png)
_Fig 1: Low performance loading cash receipts_

This is without making a payment - it takes 14 seconds to simply load the receipt.

I've commented out the debtor's balance query - it is nice to have, but throughput is the bottleneck at this point.  The result is this in production:
![cashcommenteddebtorbalances](https://cloud.githubusercontent.com/assets/896472/24236024/5536890c-0fa0-11e7-8394-6d170adac7cc.png)
_Fig 2: Commented Debtors.balance() Call_

sfount:
[...]

PR #1409, opened 2017-03-23
(perf) Break apart combined_ledger receipt queries

sfount:
These commits splits the debtor invoices and overall debtor balance query into multiple parts
to make use of MySQL's index. 
The combined_ledger is no longer used as
it does not index the required columns correctly.

Contributes to https://github.com/IMA-WorldHealth/bhima-2.X/issues/1405

### DIFF
--- a/server/controllers/finance/debtors/index.js
+++ b/server/controllers/finance/debtors/index.js
@@ -269,13 +269,20 @@ function balance(debtorUid) {
       SELECT SUM(debit - credit) AS balance, BUID(entity_uuid) as entity_uuid
       FROM (
         SELECT entity_uuid, record_uuid as uuid, debit_equiv as debit, credit_equiv as credit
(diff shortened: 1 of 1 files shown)


sfount:
@jniles I'm going to briefly verify that this performs as expected. This analysis should be done soon then this can be merged.

sfount:
### Performance Review

`debtors/index.js` **function invoiceBalances()**

Query Time | Lock Time | Rows Examined
--- | --- | --- |
0.002447 | 0.000335 | 33 |

`debtors/index.js` **function balance()**

Query Time | Lock Time | Rows Examined
--- | --- | --- |
0.000552 | 0.000223 | 14 |

sfount:
Merging this as @jniles does not have good internet access. This can be undone in the future if there is an issue.

PR #1411, opened 2017-03-23
(performance) Fix invoice lookup query

sfount:
This PR removes the relatively expensive voucher NOT IN query that
was used in the past to determine if invoices have an ascociated credit
note. This can now utilise the `reversed` column.

It also removes the implementation of the reversed filter type on
FilterParser. Our current standard is to use a `reversed` column for
performance reasons - this can be parsed by FilterParser as a default
equals filter.


jniles:
Awesome!  This is really good news.

jniles:
_ashamedly reopens_

jniles:
LGTM

PR #1416, opened 2017-03-24
Convert number to text

### LINKED ISSUE #175: Create a "Montant en Chiffres" Service/Filter
Most cash receipts have a section for expressing the amount in letters.  This issue is a feature request for the service that performs the conversion from a number to written words in either English or French.  For example:

``` js
Receipts.numberToText(100, currencyId);  // returns 'one hundred dollars'

Receipts.numberToText(750, currencyId); // returns 'sept cent cinquante francs'
```

The service should listen to the application's language locale and switch languages as necessary.  With a service such as the one described above, our receipts will look amazing.

lomamech:
Adding Numbers in translate key
Conversopm from a Number to Written Word  
  - Adding translate key for number
  - Adding lib NumberToText for the conversion
  - Adding function numberToText in helpers/finance
  - Adding property currencyName in Enterprise Property
  - Application of number to text in the patient's invoice
Complete comment

closes #175


server/lib/NumberToText.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * @module NumberToText
```
jniles, line 2: Creating a module for this was the right idea!

server/lib/NumberToText.js
```diff
@@ -0,0 +1,96 @@
...
+    '', _.get(dictionary, 'NUMBERS.THOUSAND'), _.get(dictionary, 'NUMBERS.MILLION'), _.get(dictionary, 'NUMBERS.BILLION'),
+    _.get(dictionary, 'NUMBERS.TRILLION'), _.get(dictionary, 'NUMBERS.QUADRILLION'),
+    _.get(dictionary, 'NUMBERS.QUINTILLION'), _.get(dictionary, 'NUMBERS.SEXTILLION'), _.get(dictionary, 'NUMBERS.SEPTILLION'),
+    _.get(dictionary, 'NUMBERS.OCTILLION'), _.get(dictionary, 'NUMBERS.NONILLION')
+  ];
+
+  let arr = x => Array.from(x);
+  let num = x => Number(x) || 0;
+  let str = x => String(x);
+  let isEmpty = xs => xs.length === 0;
```
jniles, line 61: Many of this functions are available in [lodash](https://lodash.com/docs/).  Is there any reason why those cannot be used?

server/lib/NumberToText.js
```diff
@@ -0,0 +1,96 @@
...
+      return [
+        num(huns) === 0 ? '' :
+          ((languageKey === 'fr') && (huns === 1)) ? ' ' + _.get(dictionary, 'NUMBERS.HUNDRED') + ' ' :  a[huns] + ' ' + _.get(dictionary, 'NUMBERS.HUNDRED') + ' ',
+        num(ones) === 0 ? b[tens] : b[tens] && b[tens] + '-' || '',
+        a[tens+ones] || a[ones]
+      ].join('');                    
+
+    };
+    
+    let thousand = (group,i) => group === '' ? group : ((group === a[1]) && (languageKey === 'fr') && (g[i] === g[1] )) ? ` ${g[i]}` :  `${group} ${g[i]}`;
```
jniles, line 86: This code is really complicated ... can you add some comments explaining it?  Or maybe break it out into multiple variable assignments?

server/controllers/finance/reports/invoices/receipt.pos.handlebars
```diff
@@ -36,6 +36,7 @@
 </table>
 {{> underline}}
 <h2 style="text-align : right;">{{translate 'FORM.LABELS.TOTAL'}}: <b>{{currency cost metadata.enterprise.currency_id}}</b></h2>
+<h4 style="text-align : right;"> <b>{{numberToText cost lang metadata.enterprise.currencyName}}</b></h>
```
jniles, line 39: Just like `translate` doesn't require the language, it would be nice to make this filter not require the language.  I believe that code is found in `html.js`, but I'm not certain.

jniles (commented):
This was a really old issue, and impressive work to complete it!

It looks like this code is from [here](http://stackoverflow.com/a/30524915) - is that correct?  If so, it is best to note it.

Also, the accepted answer to that post is [this one](http://stackoverflow.com/a/14767071), which might be easier to maintain (provided the programmer understands regular expressions).  I'm curious if you tried that code and if it works.

server/controllers/finance/reports/invoices/receipt.pos.handlebars
lomamech: If the language is not defined, English is used as the default language

jniles: But to use the currencyName, you need to specify a language, no?

lomamech: The currencyName is the contents of the column 'name' of the table 'currency'

jniles:
@lomamech, this gets the job done.  Thanks for adding a link to the source.

Once the conflicts are resolved, we can merge this PR.

PR #1463, opened 2017-04-03
Resolution of the problem related to the selection of accounts linked…

### LINKED ISSUE #1462: Problem with the API for the cashbox
The query should in principle return the id of the cash_box_account_currency.id table instead of cash_box.id because when selecting a cash box one always associates a currency and the linked account

Here is the illustration
image1

It can be seen that the query returns cashbox_id twice while the query should return cash_box_account_currency.id
image1

Here's how the data should be rendered

lomamech:
Resolution of the problem related to the selection of accounts linked to each cashbox for each currency

closes #1462

### DIFF
--- a/server/controllers/finance/cashboxes/index.js
+++ b/server/controllers/finance/cashboxes/index.js
@@ -43,7 +43,7 @@ function list(req, res, next) {
 
   if (req.query.detailed === '1') {
     sql = `
-      SELECT cash_box.id, label, account_id, is_auxiliary, transfer_account_id, symbol,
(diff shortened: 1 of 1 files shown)


jniles:
@lomamech, are you _sure_ this functionality isn't used somewhere important?  It may not be a mistake ... have you checked creating/updating a cashbox and changing cashboxes on the cash page?

lomamech:
master
@jniles Tries to visualize the id property you'll see that it's exactly the same thing, while this query should use the table id cash_box_account_currency

jniles:
@lomamech, okay.  Code is fine :+1:

PR #1465, opened 2017-04-04
Prevent Imbalanced transaction in Journal

### LINKED ISSUE #1256: Prevent any transaction imbalance in editing of Posting Journal
posting
-The publishing mode of the posting journal gives the possibility of unbalancing a transaction,

- It would be necessary for the user to have alerts to prevent any imbalance, but also to prevent the validation of such an operation.

lomamech:
Prevent Imbalanced transaction in Journal  …
- Adding function validate transaction who check if debit is equal to  credit
- Restructured catching error in controller of journal

closes #1256

### DIFF
--- a/client/src/js/services/grid/TransactionService.js
+++ b/client/src/js/services/grid/TransactionService.js
@@ -302,12 +302,12 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
    * @description
    * This method is used to check if a transaction is balanced by their record
    */
-  function transaction(entity) {
-    var dataEntity = entity.data.data;
+  function validateTransaction(entity) {
+    var transaction = entity.data.data;
     var debit = 0,
       credit = 0;
     
-    dataEntity.forEach(function (row) {
+    transaction.forEach(function (row) {
       debit += Number(row.debit_equiv);
       credit += Number(row.credit_equiv);
     });
@@ -561,7 +561,7 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
    * This function saves all transactions by
    */
   Transactions.prototype.save = function save() {
-    var clientErrors = transaction(this._entity);
+    var clientErrors = validateTransaction(this._entity);
 
     if(clientErrors){
       return $q.reject(clientErrors);
--- a/client/src/js/services/grid/TransactionService.js
+++ b/client/src/js/services/grid/TransactionService.js
@@ -2,7 +2,7 @@ angular.module('bhima.services')
   .service('TransactionService', TransactionService);
 
 TransactionService.$inject = [
-  '$timeout', 'util', 'uiGridConstants', 'bhConstants', 'NotifyService', 'uuid', 'JournalService', 'Store',
+  '$timeout', 'util', 'uiGridConstants', 'bhConstants', 'NotifyService', 'uuid', 'JournalService', 'Store', '$q'
 ];
 
 /**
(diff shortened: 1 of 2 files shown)


client/src/js/services/grid/TransactionService.js
```diff
@@ -295,6 +295,26 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
...
+    var dataEntity = entity.data.data;
+    var debit = 0,
+      credit = 0;
+    
+    dataEntity.forEach(function (journal) {
+      debit += Number(journal.debit_equiv);
+      credit += Number(journal.credit_equiv);
+    });
+
+    return debit !== credit ? true : false;
```
jniles: In the future, you can just `return debit !== credit`.  It's probably easier to read.

client/src/js/services/grid/TransactionService.js
```diff
@@ -532,6 +552,12 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
    * This function saves all transactions by
    */
   Transactions.prototype.save = function save() {
+    var clientErrors = validateTransaction(this._entity);
```
jniles, line 564: A better way than returning (and rejecting) a boolean would be to return the error code.  So, instead of returning `true` or `false`, return the string `POSTING_JOURNAL.ERRORS.UNBALANCED_TRANSACTIONS`. 

That way, anyone else who makes an error condition (like missing dates, missing values, etc) can simply add to the validate function.

jniles: The best way to do this would be to add a validation key at the top of this service, like:
```js
var ERR_IMBALANCED_TXN = 'POSTING_JOURNAL.ERRORS.UNBALANCED_TRANSACTIONS';

// later in validateTransaction()
var error;

if (debit !== credit) {
  error = ERR_UNBALANCED_TXN;
}

// another condition

// another condition ..

return error;
```

client/src/js/services/grid/TransactionService.js
```diff
@@ -295,6 +295,26 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
...
 
+
+  /**
+   * @method validateTransaction
+   *
+   * @description
+   * This method is used to check if a transaction is balanced by their record
+   */
+  function validateTransaction(entity) {
+    var dataEntity = entity.data.data;
```
jniles: A better name for this is `transaction`, since we are editing a transaction and the function is called `validateTransaction()`.

client/src/js/services/grid/TransactionService.js
```diff
@@ -295,6 +295,26 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
...
+    var debit = 0,
+      credit = 0;
+    
+    dataEntity.forEach(function (journal) {
```
jniles: A transaction is made of rows, not journals (a journal contains transactions), so a better name for this variable is `row`.  Or `line`.  Or `item`.  But `journal` is confusing.

jniles (changes requested):
This is a good start.  If we want to add more checks later, we will need to move all the logic into `validateTransacation()`.  Why don't you do that from the beginning?

client/src/js/services/grid/TransactionService.js
```diff
@@ -295,6 +295,35 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
...
   };
 
+
+  /**
+   * @method transaction
+   *
+   * @description
+   * This method is used to check if a transaction is balanced by their record
+   */
+  function transaction(entity) {
```
jniles: @lomamech, the name `validateTransaction()` was fine.  It's `dataEntity` that should be called `transaction`.  Because it _is_ a transaction, right?

client/src/js/services/grid/TransactionService.js
```diff
@@ -295,6 +295,35 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
...
 
+
+  /**
+   * @method transaction
+   *
+   * @description
+   * This method is used to check if a transaction is balanced by their record
+   */
+  function validateTransaction(entity) {
+    var transaction = entity.data.data;
```
jniles, line 306: :+1:  This is much clearer!

jniles (approved):

jniles:
@lomamech, can you rebase this branch to master so that we can be sure everything will work properly?

jniles:
@lomamech, thanks! :+1:

PR #1469, opened 2017-04-05
Update the translation of Aged Debtor in French

### LINKED ISSUE #1386: Rapport Dettes des clients
Il apparaît clairement que le contenu de ce rapport fait allusion à l'âge des créances sur clients et donc, je suggère que la dénomination change en "Balance agée clients"pour marier le contenu au nom du rapport.

lomamech:
Update the translation of Aged Debtor in French

closes #1386


jniles:
LGTM!

PR #1472, opened 2017-04-06
fix(grid): enable inline date filtering

jniles:
The inline date filters were all broken previously during an API change
that did not update the grids with the new API.  Additionally, the old
method did not work due to a syntax issue with the `moment().format()`
call.  This has been corrected by using an upper-cased version of the
bhConstant's date format.

Additionally, this commit bumps ui-grid's version to 4.0.4 to take
advantage of performance improvements.

**Example**
filterbydate
_Fig 1: Inline Filtering By Date_

---

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


client/src/partials/accounts/accounts.html
```diff
@@ -40,12 +40,9 @@
     <div class="row">
       <div class="col-xs-12">
         <!-- @todo Consider how the application is loaded and cached and make this functionality apply to the whole application -->
-        <!-- Hide broken UI-Grid on page load -->
-        <div ng-if="!AccountsCtrl.pageLoaded"><i class="fa fa-spin fa-circle-o-notch"></i></div>
         <div
           id="account-grid"
-          style="height : calc(100vh - 65px)"
-          ng-show="AccountsCtrl.pageLoaded"
+          class="grid-full-height"
```
mbayopanda, line 45: 👍

client/src/partials/accounts/accounts.js
```diff
@@ -28,13 +28,14 @@ function AccountsController($rootScope, $timeout, AccountGrid, Notify, Constants
...
-  vm.loading = true;
 
   vm.Accounts = new AccountGrid();
   vm.Accounts.settup()
     .then(bindGridData)
     .catch(Notify.handleError)
-    .finally(toggleLoadingIndicator);
+    .finally(function () {
+      toggleLoadingIndicator();
+    });
```
mbayopanda: What is the difference with `.finally(toggleLoadingIndicator);` syntax ? If there is no difference, `.finally(toggleLoadingIndicator);` seems more clear.

mbayopanda (approved):
LGTM

client/src/partials/accounts/accounts.js
jniles: They are logically equivalent.  I agree as well. :+1: I was trying to debug something and never switched it back.  Good catch :+1:

PR #1491, opened 2017-04-11
Standardize Create Buttons in 2.X

### LINKED ISSUE #1450: (refactor) Standardize "create" buttons in 2.x
We've done a good job standardizing submit buttons (blue, loading indicator, etc).  Another common pattern we have is a create button in the top right corner.  Here are three example:

create-3
create-2
create-1

Notice how all these are different in some way.  It looks rather unpolished and unprofessional.

I propose we standardize the following rules:
 1. Always use `fa fa-plus` for create buttons.
 2. Always use `text-capitalize` to make sure the text renders correctly.
 3. Always use `btn btn-default` for the class.
 4. Always use `data-method="create"` on the button as a test hook.

lomamech:
Standardize Create Buttons in 2.X  …
- Using fa fa-plus, text-capitalize, btn btn-default and data-method
  in All Buttons Create and Add
- Using component translate

closes #1450

### DIFF
--- a/client/src/modules/accounts/accounts.html
+++ b/client/src/modules/accounts/accounts.html
@@ -16,9 +16,9 @@
         <i class="fa fa-filter"></i>
        </button>
 
-       <a class="btn btn-default" data-method="create" ui-sref="accounts.create">
-         <i class="fa fa-plus-square"></i>
-         {{ "ACCOUNT.ADD" | translate }}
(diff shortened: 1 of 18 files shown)


jniles:
@lomamech, you'll have to rebase to master for this to get merged.  @sfount just merged a massive change to the client.

jniles (approved):
:+1: These are excellent changes.  It's also really nice that you updated using the component translate.

PR #1501, opened 2017-04-12
Complete validate Transaction with

### LINKED ISSUE #1498: Validate Transaction
Allow the validate transaction function to check whether a transaction is valid, and that it will not generate any errors if it is to be inserted into the general ledger

- Date in the wrong period
- Debit and Credit should not be Zero
- A transaction must have the same date
- A transaction must not have a single row

lomamech:
Complete validate Transaction with  …
- check single line transaction
- check mising accounts
- check missing dates
- check Date in wrong period
- check different date

closes #1498


client/src/js/services/grid/TransactionService.js
```diff
@@ -303,24 +303,81 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
...
+      }
+        
+
+      transaction.forEach(function (row) {
+        debit += Number(row.debit_equiv);
+        credit += Number(row.credit_equiv);
+
+
+        // Check if they are account number Null
+        if(!row.account_number){
```
jniles: A lot of these checks you can simplify by removing the "if" and setting the error equal to the check.  LIke this:
```js
accountNull = !row.account_number;
dateNull = !row.trans_date;
dateDifference = Dates.util.str(row.trans_date) !== Dates.util.str(initialDate);
```
and so on.  This  might reduce your code a bit.

The comments are nice!

client/src/js/services/grid/TransactionService.js
```diff
@@ -303,24 +303,81 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
...
+        }
+
+        // check if trans_date is in bed period
+        if(new Date(row.trans_date) < new Date(row.period_start) || new Date(row.trans_date) > new Date(row.period_end)){
+          dateWrongPeriod = true;
+        }
+      });
+
+      if(accountNull) {
+        error = 'POSTING_JOURNAL.ERRORS.MISSING_ACCOUNTS';
```
jniles: Instead of having these long text strings here, it would be nicer to list them all at the top of the module, like is done with `ROW_EDIT_FLAG` and others at the top of this file (https://github.com/IMA-WorldHealth/bhima-2.X/pull/1501/files#diff-2b8e7d8131d2079c84c7b41baa2361a5R25).  Something like this:
```js
var ERR_MISSING_ACCOUNTS = 'POSTING_JOURNAL.ERRORS.MISSING_ACCOUNTS';
var ERR_MISSING_DATES = 'POSTING_JOURNAL.ERRORS.MISSING_DATES';
//....

// much later...

if (accountNull) {
 error = ERR_MISSING_ACCOUNTS;
} else if { /// etc...
```

This will make it much easier to update translation keys in the future - ideally these keys will be shared with the complex vouchers module.

client/src/modules/journal/journal.js
```diff
@@ -313,11 +313,12 @@ function JournalController(Journal, Sorting, Grouping, Filtering, Columns, Confi
         load(vm.filters);
       })
       .catch(function (error) {
-        if(error === 'POSTING_JOURNAL.ERRORS.UNBALANCED_TRANSACTIONS'){
-          Notify.warn(error);
-        } else {
+        if(error.status){
```
jniles: You've mixed `if(` and `if (` in the PR a few times.  ESLint prefers that you use `if (` to distinguish it from a function call.  Could you use `if (` instead of `if(`?

client/src/js/services/grid/TransactionService.js
```diff
@@ -303,24 +303,81 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
...
+        dateNull =  false,
+        dateDifferent = false,
+        dateWrongPeriod = false;
+
+      if(transaction[0].trans_date){
+        initialDate = transaction[0].trans_date; 
+      }
+        
+
+      transaction.forEach(function (row) {
```
jniles: One loop is better than many! :+1:

client/src/js/services/grid/TransactionService.js
```diff
@@ -303,24 +303,81 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
...
-    // later in validateTransaction()
+    var numberOfLine = transaction.length;
     var error;
 
-    if (debit !== credit) {
-      error = ERR_UNBALANCED_TXN;
+    // If the transaction are single line transaction 
+    if(numberOfLine === 1){
+      error = 'POSTING_JOURNAL.ERRORS.SINGLE_LINE_TRANSACTION';
+    } else {
```
jniles: Instead of doing this giant `if () else {}`, you can just return here - if there is only one line, it is clearly a single line transaction and we don't need to do anything else.  It will save you one indentation.

Like this:
```js
// as suggested earlier...
var ERR_SINGLE_LINE_TRANSACTION = 'POSTING_JOURNAL.ERRORS.SINGLE_LINE_TRANSACTION';

// and then here....

if (numberOfLines === 1) {
  return ERR_SINGLE_LINE_TRANSACTION;
}

var debit = 0;
var credit = 0;
/// ... etc...
```

jniles (commented):
@lomamech looks pretty good!  Good work.

I've made a few suggestions to keep this code maintainable and readable.  The logic seems correct to me. 👍

jniles:
Looks good to me!

PR #1504, opened 2017-04-13
Add the column debit and credit source with the currency used in the …

### LINKED ISSUE #1502: Add the column debit and credit source with the currency used in the transaction
This will allow to analyze the posting journal by having a view on the debit and the credit with the monetary unit used during the transaction

lomamech:
Add the column debit and credit source with the currency used in the transaction
- Update journal grid structures
- adding templates for the format currency from debit and credit source

closes #1502


jniles:
LGTM :+1:

PR #1525, opened 2017-04-19
Filter Posting Journal on transaction Type

### LINKED ISSUE #1481: (feature request) Filter Posting Journal on Transaction Type
We've had the following feedback from HEV:

>  The Posting Journal responds very slow. It appeared that it is full of invoices. It would be necessary to get selectively rid of the Invoices (transfer them in the general ledger/grand livre.   This makes the posting journal hard to access.

We can relatively easily accommodate this with `FilterParser` and changes to the filter form -- the field to search on would be the transaction type.  We also should make sure that every one of our posted records has a transaction type.

@sfount is currently revising the default filters to speed up the initial load.  Once those are in, we can tackle this in a straightforward fashion.

lomamech:
Filter Posting Journal on transaction Type  …
- Adding element form for research by filter transaction type
- Update procedures SQL adding origin_id on cash_payment

closes #1481


jniles:
Looks good to me!

PR #1526, opened 2017-04-19
Select all rows (include group rows) in the journal grid

### LINKED ISSUE #1512: (bug) Make sure that the both tree header nodes and tree children are selected when global selection is applied.
When you click the "Select All" checkbox in the Posting Journal, only the children of parent nodes are selected.  This is a poor user experience and incorrect behavior.  The module should select all rows, including transaction headers.

mbayopanda:
This PR permits to select all rows in the journal grid, this selection include the group header rows.

select_all

Closes #1512.


client/src/js/services/grid/GridGrouping.js
```diff
@@ -21,6 +21,7 @@ function GridGroupingService(GridAggregators, uiGridGroupingConstants, Session,
   $timeout, util, uiGridConstants) {
   /** @const aggregators assigned by column ids */
   var DEFAULT_AGGREGATORS = GridAggregators.aggregators.tree;
+  var SELECTED_GROUP_HEADERS;
```
jniles: The ALL_CAPITAL_LETTERS syntax is usually for constants.  Does this variable need to be global to the entire service?  It looks like it is only used in `hadnleBatchSelection()`.

client/src/js/services/grid/GridGrouping.js
```diff
@@ -98,9 +99,32 @@ function GridGroupingService(GridAggregators, uiGridGroupingConstants, Session,
...
+        parentRow.isSelected = true;
+        parents[parentRow.uid] = parentRow;
+        SELECTED_GROUP_HEADERS = parents;
+      }
+    });
+
+    // handle deselect
+    if (hasSelections === false) {
+      angular.forEach(SELECTED_GROUP_HEADERS, function (node) {
+        node.isSelected = false;
```
jniles: What does `isSelected` do?  It looks like it is always set but never used.  Is it an internal `ui-grid` thing?

client/src/js/services/grid/GridGrouping.js
```diff
@@ -98,9 +99,32 @@ function GridGroupingService(GridAggregators, uiGridGroupingConstants, Session,
...
+    var parents = {};
+
+    var hasSelections = gridApi.selection.getSelectedRows().length > 0;
+
+    gridRows.forEach(function (row) {
+      var parentRow = row.treeNode.parentRow;
+      if (parentRow.treeLevel === 0 && !parents[parentRow.uid]) {
+        parentRow.isSelected = true;
+        parents[parentRow.uid] = parentRow;
+        SELECTED_GROUP_HEADERS = parents;
```
jniles: Do you need to set this every loop through the `gridRows`?  It looks like what you are really doing is:
```js
gridRows.forEach(function (row) {
  var parentRow = row.treeNode.parentRow;
  if (parentRow.treeLevel === 0 && !parents[parentRow.uid]) {
    parentRow.isSelected = true;
    parents[parentRow.uid] = parentRow;
  }
});

SELECTED_GROUP_HEADERS = parents;
```

If that is true, then it is easier to read (and might be more performant) to remove the `SELECTED_GROUP_HEADERS` variable assignment from the loop.

client/src/js/services/grid/GridGrouping.js
```diff
@@ -98,9 +99,32 @@ function GridGroupingService(GridAggregators, uiGridGroupingConstants, Session,
...
+  function handleBatchSelection() {
     var gridApi = this.gridApi;
+    var gridRows = gridApi.selection.getSelectedGridRows();
+    var parents = {};
+
+    var hasSelections = gridApi.selection.getSelectedRows().length > 0;
+
+    gridRows.forEach(function (row) {
+      var parentRow = row.treeNode.parentRow;
+      if (parentRow.treeLevel === 0 && !parents[parentRow.uid]) {
```
jniles: I know what this is because we've talked about it, but the code is hard to read as it is.  Could you move this check into a function?

```js
// this function identifies parent rows that we haven't seen yet
function isUnusedParentRow(row) {
  return row.treeLevel === 0 && !parents[row.uid];
}

gridRows.forEach(function (row) {
  var parentRow = row.treeNode.parentRow; 
  if (isUnusedParentRow(parentRow)) {
     // do something.
  }
});
```

client/src/js/services/grid/GridGrouping.js
```diff
@@ -98,9 +99,32 @@ function GridGroupingService(GridAggregators, uiGridGroupingConstants, Session,
     }
   }
 
-  function handleBatchSelection (){
+  function handleBatchSelection() {
     var gridApi = this.gridApi;
+    var gridRows = gridApi.selection.getSelectedGridRows();
+    var parents = {};
+
+    var hasSelections = gridApi.selection.getSelectedRows().length > 0;
```
jniles, line 107: Nice variable assignment 👍

jniles (changes requested):
@mbayopanda, I've made some comments for code clarity.

client/src/js/services/grid/GridGrouping.js
mbayopanda: The `isSelected` is responsible to set the row selected (checked) in the ui grid.

jniles: Cool - I was wondering if that was the case :+1:

client/src/js/services/grid/GridGrouping.js
mbayopanda: the `SELECTED_GROUP_HEADERS` is a global variable and must be updated only when there are new parent row (enter in the `if` clause). 

If it is out the `if` we cannot deselect parent rows which were previously selected because 
```javascript
var parents = {};

/* this code is not executed
gridRows.forEach(function (row) {
  var parentRow = row.treeNode.parentRow;
  if (parentRow.treeLevel === 0 && !parents[parentRow.uid]) {
    parentRow.isSelected = true;
    parents[parentRow.uid] = parentRow;
  }
}); */

SELECTED_GROUP_HEADERS = parents; // SELECTED_GROUP_HEADERS = {}
```

jniles: Aha, that makes sense.  I don't think there is a clearer way of expressing what you wrote.  LGTM :+1:

jniles (approved):
All my comments have been addressed :+1:

PR #1527, opened 2017-04-20
Implementation of cancel Button on the Debtor Groups Page

### LINKED ISSUE #1125: (feature request): implement a cancel button on the debtor groups page
For uniformity, I propose that implement a cancel button on the debtor groups page that will simply do go back to the `debtorGroups.list` page.  It will be functionally equivalent to the link on the top.

The motivation for this request is that it will be more familiar for users that are either (1) at the bottom of the page or (2) used to clicking cancel to stop the current action.

lomamech:
Implementation of cancel Button on the Debtor Groups Page

closes #1125

### DIFF
--- a/client/src/modules/debtors/groups.update.js
+++ b/client/src/modules/debtors/groups.update.js
@@ -20,6 +20,7 @@ function DebtorGroupsUpdateController($state, DebtorGroups, Accounts, Prices, Sc
 
   vm.$loading = true;
   vm.$loaded = false;
+  vm.cancel = cancel;
 
   vm.deleteGroup = deleteGroup;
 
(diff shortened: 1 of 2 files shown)


client/src/modules/debtors/groups.update.js
```diff
@@ -77,6 +78,10 @@ function DebtorGroupsUpdateController($state, DebtorGroups, Accounts, Prices, Sc
       .catch(Notify.handleError);
   }
 
+  function cancel(){
+    $state.go('debtorGroups.list', null, {reload : true});
```
jniles: Do you need the `reload:true` here?  Or does it work okay without it?

jniles:
LGTM!

PR #1554, opened 2017-04-24
Journal export csv with custom options

### LINKED ISSUE #1543: (feature request) Export the Trial Balance to CSV
As mentioned in #1541, 
> Besoin d'export a CSV pour le tableau raprochement.

We need to be able to export the Trial Balance to CSV.   I propose that we put this in a drop down menu at the top and right hand corner of the modal, allowing the user to download as CSV or as PDF.

mbayopanda:
This PR allows to: 

- Export to csv the trial balance
- Customize the name of the csv file
- Use the exportation tool with grids

Closes #1543


ghost:
:umbrella: The latest upstream changes (presumably 033d9cd628a4576dac6b6ea28b8b947b370ddbdc) made this pull request unmergeable. Please resolve the merge conflicts.

ghost:
:umbrella: The latest upstream changes (presumably ac06db3a2b0596ea5ac582832f05a4475b20714a) made this pull request unmergeable. Please resolve the merge conflicts.

ghost:
:umbrella: The latest upstream changes (presumably 9f17d54be7d91097fd7ee82df38c18fc7819710c) made this pull request unmergeable. Please resolve the merge conflicts.

mbayopanda:

client/src/i18n/fr/posting_journal.json
```diff
@@ -60,7 +60,7 @@
 "MISSING_DATES":"Date(s) manquante(s)",
 "MISSING_DOCUMENT_ID":"Ligne(s) sans ID Document",
 "MISSING_ENTITY":"Ligne(s) sans ID entite",
-"MISSING_FISCAL_OR_PERIOD":"Annee fiscale ou periode omise",
+"MISSING_FISCAL_OR_PERIOD":"Année fiscale ou periode omise",
```
jniles, line 64: :+1:

jniles:
Working on it!

client/src/js/app.js
```diff
@@ -5,7 +5,8 @@ var bhima = angular.module('bhima', [
   'tmh.dynamicLocale', 'ngFileUpload', 'ui.grid',
   'ui.grid.selection', 'ui.grid.autoResize', 'ui.grid.resizeColumns',
   'ui.grid.edit', 'ui.grid.grouping', 'ui.grid.treeView', 'ui.grid.cellNav',
-  'ui.grid.pagination', 'ui.grid.moveColumns', 'angularMoment', 'ngMessages',
+  'ui.grid.pagination', 'ui.grid.moveColumns', 'ui.grid.exporter',
```
jniles, line 8: Hmm... is this possible to do without the exporter service?  I don't see where it is used and I hate to introduce another dependency for something we can almost already do...

client/src/js/services/grid/GridExport.js
```diff
@@ -0,0 +1,56 @@
...
+
+function GridExportService(Modal, util) {
+
+  /**
+   * @constructor
+   */
+  function GridExport(gridOptions, defaultRows, defaultCols) {
+
+    this.gridOptions = gridOptions;
+    this.ROWS = defaultRows;
```
jniles: It took me a little while to realize that these aren't actually rows .. it's a string.  Could you name it something like `defaultRowKey` or something?

client/src/modules/journal/journal.js
```diff
@@ -50,6 +50,9 @@ function JournalController(Journal, Sorting, Grouping,
   // top level cache
   var cache = AppCache(cacheKey + '-module');
   var vm = this;
+
+  var gridApi;
```
jniles: I don't think you use `gridApi` here.

client/src/modules/journal/modals/trialBalanceError.body.js
```diff
@@ -15,10 +15,18 @@ TrialBalanceErrorBodyController.$inject = [
...
-    { field: 'transaction', displayName : 'TABLE.COLUMNS.TRANSACTION', headerCellFilter : 'translate', headerCellClass : cssClass}
-  ];
+
+  var columns = [{
+    field            : 'code',
+    displayName      : 'TABLE.COLUMNS.ERROR_TYPE',
+    headerCellFilter : 'translate',
+    headerCellClass  : cssClass,
+  }, {
+    field            : 'trans_id',
```
jniles: :+1:  Good catch.

client/src/modules/templates/modals/export.modal.js
```diff
@@ -0,0 +1,46 @@
...
+];
+
+function ExportGridModalController(Instance, uiGridConstants, $filter,
+  moment, bhConstants, Data) {
+  var vm = this;
+
+  var gridOptions = Data.options || {};
+  var gridApi = Data.api || {};
+  var filename = Data.filename || 'Export ' + moment().format('YYYY-MM-DD');
+  var ROWS = Data.rows || 'visible';
```
jniles, line 16: For simplicity, I propose we always render all rows in the grid.  That's how our PDFs work and it makes sense that the filters chosen will be the ones applied to the dataset.

ghost:
:umbrella: The latest upstream changes (presumably 4df53c2b697162cc4b8fa047984b52748fa300cc) made this pull request unmergeable. Please resolve the merge conflicts.

ghost:
:umbrella: The latest upstream changes (presumably 7cf8682555452bdd77fd84af2b8ff51561270f31) made this pull request unmergeable. Please resolve the merge conflicts.

ghost:
:umbrella: The latest upstream changes (presumably 6bc89ba7f86e7e1c38cae0d0161c85d1c135294e) made this pull request unmergeable. Please resolve the merge conflicts.

mbayopanda:

client/src/js/app.js
mbayopanda: Yes but with a lot of supplementary request to the server, also if we don't use the exporter service we cannot share the client service for exporting with custom name to other modules.

The first reason of using `exporter service` was because the trial balance request is so complex, and using this request again and again is not useful when we don't have a lot of data at client and data are already at the client we have just to export them.

Also `exporter service` give us more options for exportation which are flexible and we can use them in other modules with less effort

client/src/js/services/grid/GridExport.js
mbayopanda: 👍

client/src/modules/templates/modals/export.modal.js
mbayopanda: The current state of the `exporter` is just for `csv`, so we render what is visible to the client with the possibility of choosing other options (selected or all). This flexibility is useful because we have some grid which has grouped data and other not, for grid which has grouped data with the option `visible` the service export what is visible (ex. only grouping rows in the case of posting journal with grouped rows) and the user can select other options if he want, and developers can set the default option for grid according needs.

mbayopanda:

client/src/js/app.js
jniles: Your analysis is correct - for the trial balance, it is more work to export a CSV from the server side.

My biggest worry with this change is that, as a developer, I do not know when to use the server-side CSV rather than the client-side CSV.  It creates another question to ask during development and something else to test....

jniles (commented):
@mbayopanda, I need a second opinion on this.

I'm concerned about having two ways of doing things - one on the client and one on the server.  It feels like unnecessary code bloat and technical debt... kind of against both [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and [KISS](https://en.wikipedia.org/wiki/KISS_principle) principles.

@sfount, could you provide your perspective?

ghost:
:umbrella: The latest upstream changes (presumably 89a32725669d5bb541b80fcdfda3f996dedd39bf) made this pull request unmergeable. Please resolve the merge conflicts.

jniles:
Since the work has already been done for this, we'll merge it and see how it works in production.  If necessary, we can revert changes or implement something different/more.

PR #1556, opened 2017-04-25
Fix billing services modal

### LINKED ISSUE #1547: (bug) Clicking outside the Billing Services modal eliminates the ability to re-enter the create state
The Billing Services create modal allows a user to click outside of it to dismiss it.  The only problem is that you can never return into the create state again.  We should probably disallow click to dismiss the modal - same with the ESC key.

lomamech:
- Prevent to dismiss Billing Services modal when Clicking Outside
- Prevent dismiss when Click the ESC Key

closes #1547


jniles (approved):
LGTM.  Clean, simple fix.

sfount (approved):

Issue #1560, opened 2017-04-25
(bug) Can't edit transaction with new line in posting journal


mbayopanda:
When editing transaction by adding new line, there is an error with `account_name`

![error_account_name](https://cloud.githubusercontent.com/assets/5445251/25384074/b2a0b742-2973-11e7-8abc-97e11e362034.png)

Issue #1558, opened 2017-04-25
Delete transaction in posting journal is not implemented


mbayopanda:
The posting journal edit mode cannot remove a transaction i.e. delete all transactions rows and click on save  throw an error in the console, and we don't have another way for deleting a transaction in the journal in BHIMA.

![error_journal](https://cloud.githubusercontent.com/assets/5445251/25382941/a8b5fe18-296e-11e7-92d8-6cff2450c7dc.png)

Issue #1536, opened 2017-04-20
(bug) Cannot remove an Entity from the Posting Journal on a new line


jniles:
The  "Add Entity" feature only works on existing lines in the Posting Journal to remove, it seems.

Steps to reproduce:
 1. Load the Posting Journal with transactions already in it.
 2. Click "Edit" on a transaction
 3. Add a new line
 4. Double click the "Entity" column to add an entity.
 5. Add a debtor
 6. Submit the debtor
 7. Attempt to remove that debtor using the red trash bin.
 8. Nothing happens.

PR #1561, opened 2017-04-25
Fix journal edit errors

mbayopanda:
This PR fix these issues : #1536 #1560 #1558


client/src/js/services/grid/TransactionService.js
```diff
@@ -345,6 +345,11 @@ function TransactionService($timeout, util, uiGridConstants, bhConstants, Notify
       ERR_DEB_CRED_NOT_NULL = 'POSTING_JOURNAL.ERRORS.DEB_CRED_NOT_NULL',
       ERR_CREDITED_DEBITED = 'POSTING_JOURNAL.ERRORS.CREDITED_DEBITED';
 
+    // If the transaction has 0 line
+    if (numberOfLine === 0) {
```
jniles, line 349: :+1:  Good catch.

server/controllers/finance/journal/index.js
```diff
@@ -374,6 +374,10 @@ function transformColumns(rows, newRecord) {
       delete row.account_number;
     }
 
+    if (row.account_name) {
```
jniles, line 377: :+1: Good catch.

jniles (approved):
Really simple fixes, looks like an excellent PR.  Great work! :+1:

PR #1562, opened 2017-04-26
Changing the posting journal in transaction mode

### LINKED ISSUE #1557: Fix Posting Journal in Transaction Mode
Changing the posting journal in transaction mode
Adjust the display of total debits and Credits
Maintain the display of total debits and credits after using a filter

lomamech:
Changing the posting journal in transaction mode
- Adjust the display of total debits and Credits
- Maintain the display of total debits and credits after using a filter

closes #1557

### DIFF
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -7,7 +7,7 @@ JournalController.$inject = [
   'SessionService', 'NotifyService', 'TransactionService', 'GridEditorService',
   'bhConstants', '$state', 'uiGridConstants', 'ModalService', 'LanguageService',
   'AppCache', 'Store', 'uiGridGroupingConstants', 'ExportService', 'FindEntityService',
-  'FilterService', '$rootScope', '$filter', '$translate'
+  'FilterService', '$rootScope', '$filter'
 ];
 
 /**
@@ -33,7 +33,7 @@ JournalController.$inject = [
 function JournalController(Journal, Sorting, Grouping,
   Filtering, Columns, Config, Session, Notify, Transactions, Editors,
   bhConstants, $state, uiGridConstants, Modal, Languages, AppCache, Store,
-  uiGridGroupingConstants, Export, FindEntity, Filters, $rootScope, $filter, $translate) {
+  uiGridGroupingConstants, Export, FindEntity, Filters, $rootScope, $filter) {
 
   // Journal utilities
   var sorting;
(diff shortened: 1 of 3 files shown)


client/src/modules/journal/journal.js
```diff
@@ -321,6 +331,7 @@ function JournalController(Journal, Sorting, Grouping,
         vm.gridOptions.data = transactions.preprocessJournalData(records);
 
         transactions.applyEdits();
+        vm.gridApi.core.notifyDataChange(uiGridConstants.dataChange.ALL);
```
sfount: This looks like a good solution if it works! I wonder why the grid doesn't register that data has changed in this case? 

This is concerning as either this should happen automatically or it should be the responsibility of the service that has changed the data to make this notify. 

Could you leave an `@TODO` comment so that this can be investigated in the future? 
```//@TODO investigate why footer totals aren't updated automatically on data change```

This solution will work well for now.

**Comment by ghost** [comment]
:umbrella: The latest upstream changes (presumably ac06db3a2b0596ea5ac582832f05a4475b20714a) made this pull request unmergeable. Please resolve the merge conflicts.

File: client/src/modules/journal/journal.js [hunk 2]
```diff
@@ -340,6 +350,9 @@ function JournalController(Journal, Sorting, Grouping,
         vm.gridOptions.gridFooterTemplate = '<div><strong>' + $translate.instant('FORM.INFO.NUM_TRANSACTION') +
```
jniles: Would this also be solved if you just made the footer template into:
```html
<div>
  <strong>
    <span translate>FORM.INFO.NUM_TRANSACTION</span>:
    {{ grid.appScope.numberGridTransactions }} / {{ grid.appScope.numberTotalSystemTransctions }}
  </strong>
</div>
```

Then instead of adding in `gridApi`, `$translate`, and other variables in the journal's controller code, you can just write
```js
vm.numberCurrentGridTransactions = records.aggregate.length;
```

Maybe I'm missing something though ...

jniles (commented):
@lomamech, this looks like it could be solved a simpler way.  I've proposed doing all this in the grid footer, which I think is clearer and much less code.  Let me know what you think.

client/src/modules/journal/journal.js
lomamech: No problem for me

client/src/modules/journal/templates/grid.footer.html
```diff
@@ -0,0 +1,6 @@
+<div>
+  <strong>
+    <span translate>FORM.INFO.NUM_TRANSACTION</span>:
```
jniles, line 3: :+1:

jniles:
LGTM!

Issue #1533, opened 2017-04-20
Fix the "Extrait de compte" file name


mbayopanda:
This issue is about to fix the "extrait de compte" file name to a name which contains the concerned account number or label.

When we export the "extrait de compte" to csv file, the exported file has a name which cannot help to know the concerned account.

PR #1563, opened 2017-04-26
Fix account slip filename

mbayopanda:
This PR fix the issue #1533

### DIFF
--- a/server/controllers/finance/reports/generalLedger/index.js
+++ b/server/controllers/finance/reports/generalLedger/index.js
@@ -66,22 +66,23 @@ function renderAccountSlip(req, res, next) {
 
   let report;
 
-  try {
-    report = new ReportManager(ACCOUNT_SLIP_TEMPLATE, req.session, options);
-  } catch (e) {
-    return next(e);
-  }
-
   const data = {};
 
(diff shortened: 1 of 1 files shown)


server/controllers/finance/reports/generalLedger/index.js
```diff
@@ -66,22 +66,23 @@ function renderAccountSlip(req, res, next) {
...
-  }
-
   const data = {};
 
-  return AccountReport.getAccountTransactions(params.account_id, GENERAL_LEDGER_SOURCE)
+  return Accounts.lookupAccount(params.account_id)
+    .then((account) => {
+      _.extend(data, { account });
+
+      options.filename = 'Extrait'.concat('_', account.number, '_', account.label);
```
jniles, line 75: Do you think this should include a timestamp (or a date)?  That way a user can generate multiple without overriding the previous?

mbayopanda: The `ReportManager` add automatically a date at the end of the filename.

csv_export_name

jniles: Ah, I didn't know that.  LGTM then.

sfount (approved):
LGTM - addresses the referenced issue.

PR #1564, opened 2017-04-27
refactor(util): remove unused util.js fns

### LINKED ISSUE #1530: (refactor) Remove util.queryCondition() in favor of FilterParser()
The old (hacky) way to use filters in 2.x was to use `util.queryCondition()`.  We now have a well-designed and mature library for adding filters to SQL queries - `FilterParser()`.  To strengthen the codebase and reduce technical debt, we should remove `util.queryCondition()` from the repository, replacing all instances of it with `FilterParser()`.

### LINKED ISSUE #1531: (refactor) Remove unused lib/util.js methods
The `lib/util.js` functions on the server predate our usage of `lodash` and many of them are now unused.  Someone should go through and remove the ones that are no longer for clean-up reasons.

jniles:
This commit removes unused util.js function from the server library.
Many of these were never used; ~~however, to fully address some of the
changes required, we are required to updated `filters.js` to allow
arbitrary WHERE clause positioning.  I'm not confident in this method,
but it does allow the power of `FilterParser` to be used with `GROUP BY`
clauses, provided the developer is careful in crafting their queries.~~

**EDIT**
To fully address some of the changes required, I've extended the `FilterParser` library with a special method `setGroup()` that will encapsulate the GROUP BY clause of the SQL.  It turns out pretty cleanly.

I've also refactored the `FilterParser` class to be in conformance with ESLint.

Closes #1530.  Closes #1531.

----

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/server/lib/filter.js
+++ b/server/lib/filter.js
@@ -16,7 +16,6 @@ const DEFAULT_UUID_PARTIAL_KEY = 'uuid';
   // week : () => { return { start : moment().startOf('week').toDate(), end : moment().endOf('week').toDate() } },
   // month : () => {  return { start : moment().startOf('month').toDate(), end : moment().endOf('month').toDate() } }
 // };
-
 /**
  * @class FilterParser
  *
@@ -40,14 +39,12 @@ const DEFAULT_UUID_PARTIAL_KEY = 'uuid';
  */
 class FilterParser {
   // options that are used by all routes that shouldn't be considered unique filters
-  constructor(filters, filterOptions) {
-    let options = filterOptions || {};
-
+  constructor(filters = {}, options = {}) {
     // stores for processing options
     this._statements = [];
     this._parameters = [];
 
-    this._filters = _.clone(filters) || {};
+    this._filters = _.clone(filters);
 
     // configure default options
     this._tableAlias = options.tableAlias || null;
@@ -55,6 +52,7 @@ class FilterParser {
     this._order = '';
     this._parseUuids = _.isUndefined(options.parseUuids) ? true : options.parseUuids;
     this._autoParseStatements = _.isUndefined(options.autoParseStatements) ? true : options.autoParseStatements;
+    this._group = '';
   }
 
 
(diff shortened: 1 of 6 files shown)


server/lib/filter.js
```diff
@@ -195,12 +195,23 @@ class FilterParser {
...
     this._parseDefaultFilters();
-    let conditionStatements = this._parseStatements();
-    let order = this._order;
+    const conditionStatements = this._parseStatements();
+    const order = this._order;
+    let statement;
+
+    // allows a developer to specify the position of the WHERE injection
+    // by default, it will simply concat
+    const hasWhereInjection = sql.includes('{{ WHERE }}');
```
sfount: This is an interesting solution! It took me a minute to figure out exactly what was going on here when looking at it from an API perspective.

I think this is because `FilterParser` didn't anticipate correctly ordering statements like `ORDER BY`, `GROUP BY`, `LIMIT` etc. and so is limited. 

The quick solution for `ORDER BY` was the method `setOrder()`, limit is done automagically and somewhat confusingly with `_parseLimit()` and now group by can be handled by adding `{{ WHERE }}` into the template. 

I suggest we design the API here so that it can handle each of these in a uniform way, I think there are a number of different solutions. 

1. Use the `setX` method for each eventuality, `setOrder`, `setGroup`, `setLimit` etc. These methods will accept the SQL that should be templated and enter them in the correct order. 

2. Use the templating method for each eventuality, in this way the programmer will specify the SQL with a `{{ WHERE }}` line and add limits, groups and order manually. 

3. Limit the scope of filter parser to only deal with the `WHERE` condition; this would then leave it up to the programmer how the SQL is templated. For example: 

```js
// equivalent of filters.applyQuery but not concerned with the sql itself
const where = filters.parseFilterConditions();
const whereParams = filters.parameters();

// most complex example
const query = `${sql} WHERE ${where} ${groupSql} ${orderSql} ${limitSql};

db.exec(query, whereParams) //...
```


- Do you think that it's worth ensuring that these APIs are uniform or is the current method okay? 
- Do you have any thoughts on the proposals here?

sfount (commented):
This is a great refactor.

I've made a comment concerning the API for `FilterParser`, it would be good to discuss this however the code here is well written and can be merged at any time depending on your thoughts on the API.

server/lib/filter.js
jniles: In my mind, I would be in favor of leaving `LIMIT` handling in the `FilterParser` library... you can think of it as a type of filter that simple filters out result beyond X rows.

`filters.setGroup()` is fine as an API.  I wonder if we'll hit a corner case where the grouping isn't set in the correct location properly...  We can always try it an see.  I had started this PR that way, and then switched because I worried it wouldn't turn out correctly.

jniles:
@sfount I've made the API change.  It now looks like:
```js
filters.setOrder('ORDER BY table.column ASC;');
filters.setGroup('GROUP BY table.otherColumn');
```

It's much cleaner than the previous implementation, good suggestion: :+1:

ghost:
:umbrella: The latest upstream changes (presumably #1566) made this pull request unmergeable. Please resolve the merge conflicts.

sfount:
@jniles The more uniform API looks great. We can think this through fully in a future refactor. LGTM!

PR #1566, opened 2017-04-28
(refactor) Journal Search

sfount:
This pull request refactors the client/ server side APIs for filtering data. It introduces a standardised service (`FilterService`) for formatting filters for both local caching and server HTTP request/ labelling. 

It also introduces a first step towards selecting and caching commonly used periods with period select components and services - this allows an accountant to always be looking at the current/ previous day/ week/ month/ year. 

**Design concepts behind FilterService:**
The goal of the FilterService API is to drastically reduce the code and responsibility of the controller performing filtering logic. As it is right now each different controller implements filters in a custom way. The way this has been implemented in Journal shows how the controller can use filters without knowing anything about the logic - which has been divided between the `FilterService` and the `JournalService`.

**To demonstrate the responsibilities/ exposed API for each component in the proposed system:**
```js
// ModuleController.js
const ModuleService = require('./ModuleService');
const SearchModule = require('./SearchModule');

// make use of the currently active filters for HTTP requests
$http.get('...', ModuleService.filters.formatHTTP());

// pass filter values to search controller to have them updated 
SearchModule.open(ModuleService.filters)
  .then((changes) => ModuleService.filters.replaceFilters(changes));
```
```js
// ModuleService.js
const filters = require('./FilterService');
const cache = require('./AppCache');

// define default filters (always shown in orange) 
filters.registerDefaultFilters([
  { key : 'period', label : 'PERIOD', valueFilter : 'translate' },
  { key : 'limit', label : 'LIMIT' } ]);

// define custom filters - these are always optional and can be defined at any point
filters.registerCustomFilters([
  { key : 'description', label : 'DESCRIPTION' },
  { key : 'account_id', label : 'ACCOUNT' }]);

// load value from cache
filters.loadCache(cache.filters);

// assign values to cache 
cache.filters = filters.formatCache();
```

```js
// SearchModule.js
const changes = [];
const formValues = { 
  description : 'Description to search on', 
  account_id : 10030
};

// applying filters exposes an API that allows setting filters with the format:
// { key : 'xxx', value : 'yyy' }
changes = _.map(formValues, (key, value) => { key, value });

return changes;
```

Functionality Review: 

Filters - Default filters
screenshot 2017-04-28 at 09 47 34

Filters - Default and custom search queries 
screenshot 2017-04-28 at 10 17 39

Filters - Modal structure for defining defaults and search queries 
screenshot 2017-04-28 at 10 31 28
screenshot 2017-04-28 at 10 31 40

Filters - Period select Component 
screenshot 2017-04-28 at 10 12 31
screenshot 2017-04-28 at 10 12 52

Filters - Period select custom period 
screenshot 2017-04-28 at 10 14 02

Header - large screen 
screenshot 2017-04-28 at 10 24 06

Header - small screen 
screenshot 2017-04-28 at 10 24 50


client/src/css/structure.css
```diff
@@ -898,3 +898,22 @@ a[disabled] {
...
+.filter-list {
+  display : inline-block;
+}
+
+.filter-list > li {
+  padding : 0px;
+}
+
+.filter-list li:first-child {
+  padding-left : 5px !important;
```
jniles: Do you need the `!important` directive?

client/src/js/components/bhFilters.js
```diff
@@ -0,0 +1,35 @@
+angular.module('bhima.components')
+.component('bhFilters', {
```
jniles, line 2: This is a very clean component :+1:

client/src/js/components/bhFilters.js
```diff
@@ -0,0 +1,35 @@
...
+    filters.customFilters.forEach(mapDisplayValues);
+  };
+
+  function mapDisplayValues(filter) {
+    filter.displayValue = filter._displayValue || filter._value;
+
+    if (filter._valueFilter) {
+      filter.displayValue = $filter(filter._valueFilter)(filter.displayValue);
+    }
+    return filter;
```
jniles: None of these look like they are used as a `map`.  This `return` is not needed.

client/src/js/components/bhPeriodSelect.js
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  ctrl.selectCustomPeriod = function selectCustomPeriod(selection) {
+    var period = Periods.index.custom;
+
+    // alias start and
+    period.customPeriodStart = selection.from;
+    period.customPeriodEnd = selection.to;
+
+    ctrl.onSelectCallback({ key : period });
```
jniles: Perhaps this is nitpicky, but would it make better sense to have `{ period: period }`?  That way the HTML reads:
```html
<bh-period-select
  on-select-callback="SomeCtrl.onPeriodSelect(period)">
</bh-period-select>
```
It seems clearer to expect a period rather than a key.... unless we are actually returning a key, but it doesn't look like it.

client/src/js/services/FilterService.depricated.js
```diff
@@ -0,0 +1,53 @@
+angular.module('bhima.services')
+  .service('DepricatedFilterService', DepricatedFilterService);
```
jniles, line 2: _Technically_, it's [deprecated](https://en.wikipedia.org/wiki/Deprecation). :smile:

client/src/js/services/FilterService.js
```diff
@@ -4,50 +4,183 @@ angular.module('bhima.services')
...
+  this._label = label;
+  this._valueFilter = valueFilter;
+  this._value = null;
+  this._isDefault = null;
+  this._displayValue = null;
+
+  this.setDefault = function setDefault(value) {
+    this._isDefault = value;
+  }
+  this.assign = function assign(value, displayValue) {
```
jniles: I would recommend a different name for this method ... maybe `set` or `configure` or `setValue` or something.  I spent a long time trying to figure out how you were doing magic with [Object.assign()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) until I finally realized there was a custom method defined here!

client/src/js/services/FilterService.js
```diff
@@ -4,50 +4,183 @@ angular.module('bhima.services')
...
+  FilterList.prototype._resetCustomFilters = function resetCustomFilters() {
+    this._filterActiveFilters().forEach(function (filter) {
+      // only by default remove custom values
+      if (!filter._isDefault) {
+        this.resetFilterState(filter._key);
+      }
+    }.bind(this));
+  }
+
+  // @TODO registerDefaultFilter and registerCustomFilter could use the same underlying function
```
jniles, line 29: Nice comment :+1:

client/src/js/services/FilterService.js
```diff
@@ -4,50 +4,183 @@ angular.module('bhima.services')
...
+
+    // udpate index
+    this._indexList(this._filterIndex, formattedFilters);
+    this._defaultFilters = this._defaultFilters.concat(formattedFilters);
+  };
+
+  FilterList.prototype.registerCustomFilters = function registerCustomFilters(filterDefinitions) {
+    var formattedFilters = filterDefinitions.map(function (filterDefinition) {
+      var filter = new Filter(filterDefinition.key, filterDefinition.label, filterDefinition.valueFilter);
+      filter.setDefault(false);
```
jniles, line 50: Can `setDefault(boolean)` be assumed for one class of filters?  Or do you think it is better to be explicit about it?

sfount:
The E2E issues are a result of the automatic merge, I'll push an update for this.

kwilu:
@sfount: :key: Insufficient privileges: Not in reviewers

client/src/js/services/FilterService.js
```diff
@@ -4,50 +4,183 @@ angular.module('bhima.services')
...
+    this._resetCustomFilters();
+    this.assignFilters(valueList);
+  };
+
+  // return filters for the view - this method will always be compatible with the bhFilter component
+  FilterList.prototype.formatView = function formatView() {
+    var activeFilters = this._filterActiveFilters();
+    var activeKeys = activeFilters.map(function (filter) {  return filter._key });
+
+    function keysInActive(filter) { return activeKeys.includes(filter._key); }
```
jniles: [Array.prototype.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) is ECMAScript version 6.  I think it's probably best to avoid it until we can compile ES6 on the client (even if it is supported by Chrome).  What do you think?

client/src/js/services/PeriodService.js
```diff
@@ -0,0 +1,140 @@
...
+    OTHER : [periods.allTime]
+  };
+
+  service.index = periods;
+  service.categories = categories;
+  service.definition = definition;
+
+  service.dateFormat = 'DD/MM/YYYY';
+
+  function calculatePeriod(key) {
```
jniles: Is this explicitly left blank for a reason?

client/src/js/services/FilterService.js
```diff
@@ -4,50 +4,183 @@ angular.module('bhima.services')
...
+      defaultFilters : this._defaultFilters.filter(keysInActive),
+      customFilters : this._customFilters.filter(keysInActive)
+    };
+  };
+
+  // format filters for the server
+  // sendClientTimestamp - this will send an attribute hidden to the user
+  // returns a JSON object with active filters
+  FilterList.prototype.formatHTTP = function formatHTTP(sendClientTimestamp) {
+    var clientTimestamp = angular.isDefined(sendClientTimestamp) ? sendClientTimestamp : false;
```
jniles: It would be clearer to name this `hasClientTimestamp` or `needsClientTimestamp` or something similar to indicate this is a boolean and not an actual timestamp.

client/src/modules/journal/modals/search.modal.html
```diff
@@ -12,106 +12,153 @@
...
+          </div>
+        </div>
+
+        <!-- users ui-select -->
+        <div class="form-group" ng-class="{'has-error' : ModalCtrl.user.$invalid && ModalForm.$submitted}">
+          <label class="control-label" translate>
+            FORM.LABELS.USER
+          </label>
+          <bh-clear on-clear="ModalCtrl.clear('user_id')"></bh-clear>
+          <ui-select name="user" ng-change="ModalCtrl.onSelectUser(ModalCtrl.searchQueries.user_id)" ng-model="ModalCtrl.searchQueries.user_id">
```
jniles, line 46: We should make a `bh-user-select` component for this in the future.

client/src/modules/journal/modals/search.modal.js
```diff
@@ -2,17 +2,41 @@ angular.module('bhima.controllers')
...
-function JournalSearchModalController(Instance, Users, Projects, Notify, options, Vouchers, $translate) {
+function JournalSearchModalController(Instance, Users, Projects, Notify, Store, options, Periods, Vouchers, $translate) {
   var vm = this;
 
-  vm.options = options || {};
+  var changes = new Store({ identifier : 'key' });
 
-  // @FIXME patch hack - this should be handled by FilterService
-  delete vm.options.defaultPeriod;
+  var filters = options;
```
jniles: Can you just rename `options` to `filters`?

client/src/modules/journal/modals/search.modal.js
```diff
@@ -25,36 +49,58 @@ function JournalSearchModalController(Instance, Users, Projects, Notify, options
...
+    delete vm.searchQueries[key];
+  };
+
+  vm.cancel = Instance.dismiss;
+
   // returns the filters to the journal to be used to refresh the page
   vm.submit = function submit(form) {
-    if (form.$invalid) { return; }
+    // push all searchQuery values into the changes array to be applied
+    angular.forEach(vm.searchQueries, function (value, key) {
```
jniles, line 93: I wonder if we couldn't use `util.filterFormValues()` for this...  do you think it would be possible?

client/src/modules/templates/bhFilters.tmpl.html
```diff
@@ -0,0 +1,41 @@
+<div ng-hide="$ctrl.filters" class="text-muted">
+  <i class="fa fa-filter"></i> {{ "FORM.INFO.NO_FILTERS_APPLIED" | translate }}
```
jniles: This should use the `translate` directive.

client/src/modules/templates/bhFilters.tmpl.html
```diff
@@ -0,0 +1,41 @@
...
+    <span class="label label-warning">
+      <i class="fa fa-filter"></i>
+      {{ filter._label | translate }}
+      {{ filter.comparitor ? " " + filter.comparitor : ":" }}
+      {{ filter.displayValue }}
+    </span>
+  </li>
+</ul>
+
+<div ng-if="$ctrl.filters.customFilters.length" style="display : inline-block; width : 1px; height: 25px; vertical-align : top; background-color : #ccc;">
```
jniles: This looks unused/abandoned.  Can you remove it?

client/src/modules/templates/bhPeriodSelect.tmpl.html
```diff
@@ -0,0 +1,62 @@
...
+
+  <!-- Exception for ALL time - this is the only period that does not need to be defined -->
+  <span ng-if="$ctrl.period.key !== $ctrl.NO_PERIOD_LIMIT_KEY && $ctrl.period.key !== $ctrl.CUSTOM_PERIOD_KEY">
+  ({{$ctrl.period.start}}) - ({{$ctrl.period.end}})
+  </span>
+</p>
+
+<a href ng-click="$ctrl.toggleSelectionOptions()">
+  <i class="fa" ng-class="{ 'fa-calendar-plus-o' : !$ctrl.expanded, 'fa-calendar-minus-o' : $ctrl.expanded }"></i>
+  Select period
```
jniles: This needs to be translated.

client/src/modules/templates/bhPeriodSelect.tmpl.html
```diff
@@ -0,0 +1,62 @@
...
+  </div>
+  <div class="row">
+    <div class="col-md-12">
+      <ul>
+        <li ng-repeat="period in $ctrl.Periods.categories.OTHER"><a href ng-click="$ctrl.selectPeriod(period.key)"><span translate>{{period.translateKey}}</span></a></li>
+      </ul>
+
+      <!-- @TODO move these templates/ logic into a component that can be unit tested -->
+      <ul>
+        <li><a href ng-click="$ctrl.toggleCustomSelection()">Custom</a></li>
```
jniles: This should be translated.

jniles (commented):
This is a whole lot of code, and most of it is very readable.  There are a few complexities that are unavoidable.

I've made some comments for your feedback.  One of them involves an internal API change.  Let me know what you think.

client/src/js/services/FilterService.depricated.js
sfount: This is just further incentive to get the files/ code removed from the repository.

client/src/js/services/FilterService.js
sfount: Agreed, I fell for the trap of wanting the code to read well. I'll replace this with `indexOf` but one day we should drink the [Kool-Aid](www.koolaid.com) of ES6. (Good catch!)

client/src/modules/journal/modals/search.modal.html
sfount: 👍

client/src/modules/templates/bhFilters.tmpl.html
sfount: This is the separator between filter types - You're right though, I'll add a comment.

client/src/modules/journal/modals/search.modal.html
jniles: Relevant issue: #1567

client/src/js/services/FilterService.js
sfount: I would have no problem with this refactor - I think it can stay explicit for now and on the next round of development the internal methods can be streamlined.

client/src/modules/journal/modals/search.modal.js
sfount: This is certainly possible! The way that the changes array currently works expects all values to be used even if they haven't been changed ($dirty) - we would just have to make sure they're all picked up. This could be refined as this is implemented across modules.

sfount:
@jniles great suggestions - I have updated this pull request.

kwilu:
@sfount: :key: Insufficient privileges: Not in reviewers

jniles:
@kwilu are you still there?

ghost:
:cake: Target acquired

jniles:
@sfount, if you comment out [this line](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/edit_journal/edit_journal.spec.js#L68), your tests will pass.

Don't ask why, just believe.

jniles:
LGTM!  Glad it worked.

PR #1572, opened 2017-05-02
chore(translation) Add french .json for periods

sfount:
This commit adds french translations for the periods select feature
recently added to Journal filters. It also ensures that the correct
sizes are used for utility bar expand/ collapse.


jniles:
This looks good to me!  The height fixes are very helpful. :+1:

PR #1575, opened 2017-05-03
Create bh user select

### LINKED ISSUE #1567: (feature request) Create a bhUserSelect component
In almost every search form, we are using a `ui-select` for selecting users.  In the same way that we standardized `bhAccountSelect`, I think it would be profitable to encapsulate this into a component since we use it so often.  It would also give us some code to unit test.

lomamech:
Creation of component bhUserSelect
Using bh-user-select in:
- Journal Search
- Invoice Registry Modal
- Patient Registry
- Payment Registry
- Purchases Order 
Update test E2E using component UserSelect

closes #1567

### DIFF
--- a/client/src/js/components/bhUserSelect.js
+++ b/client/src/js/components/bhUserSelect.js
@@ -5,23 +5,25 @@ angular.module('bhima.components')
     transclude  : true,
     bindings    : {
       userId           : '<',
-      disable          : '<',
-      onSelectCallback : '&?',
+      disable          : '<?',
+      onSelectCallback : '&',
+      name             : '@?',
       required         : '<?',      
     },
   });
 
 UserSelectController.$inject = [
-  'UserService', '$scope'
+  'UserService'
 ];
 
 /**
  * User selection component
  *
  */
-function UserSelectController(Users, $scope) {
+function UserSelectController(Users) {
   var $ctrl = this;
-
+  // default for form name
+  $ctrl.name = $ctrl.name || 'UserForm';
 
   $ctrl.$onInit = function onInit() {    
     // fired when an user has been selected
--- a/client/src/js/components/bhUserSelect.js
+++ b/client/src/js/components/bhUserSelect.js
@@ -22,12 +22,13 @@ UserSelectController.$inject = [
  */
 function UserSelectController(Users) {
   var $ctrl = this;
-  // default for form name
-  $ctrl.name = $ctrl.name || 'UserForm';
 
   $ctrl.$onInit = function onInit() {    
     // fired when an user has been selected
     $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
+
+    // default for form name
+    $ctrl.name = $ctrl.name || 'UserForm';
 
     // load all User
     Users.read()
--- a/client/src/js/components/bhUserSelect.js
+++ b/client/src/js/components/bhUserSelect.js
@@ -1,0 +1,44 @@
+angular.module('bhima.components')
+  .component('bhUserSelect', {
+    templateUrl : 'modules/templates/bhUserSelect.tmpl.html',
+    controller  : UserSelectController,
+    transclude  : true,
+    bindings    : {
+      userId           : '<',
+      disable          : '<',
+      onSelectCallback : '&?',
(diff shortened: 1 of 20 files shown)


client/src/modules/templates/bhUserSelect.tmpl.html
```diff
@@ -0,0 +1,24 @@
...
+    ng-class="{ 'has-error' : UserForm.$$parent.$submitted && UserForm.user_id.$invalid }">
+
+    <label class="control-label" translate>
+      FORM.LABELS.USER
+    </label>
+    
+    <ng-transclude></ng-transclude>
+    <ui-select name="user_id" 
+      ng-model="$ctrl.userId"
+      on-select="$ctrl.onSelect($item, $model)"$onInit >
```
jniles: This looks like broken HTML.  I think you accidently pasted `$onInit` in there.

client/src/modules/vouchers/modals/search.modal.html
```diff
@@ -38,18 +38,14 @@
...
     </div>
 
-
-    <div class="form-group" ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.user.$invalid }">
-      <label class="control-label" translate>FORM.LABELS.USER</label>
+    <bh-user-select
+      user-id="ModalCtrl.params.user_id"
+      name="user_id"
+      on-select-callback="ModalCtrl.onSelectUser(user)"
+      required="0">
```
jniles: I don't think you need the "required" here.  It should work fine without it.

client/src/js/components/bhUserSelect.js
```diff
@@ -0,0 +1,44 @@
...
+    bindings    : {
+      userId           : '<',
+      disable          : '<',
+      onSelectCallback : '&?',
+      required         : '<?',      
+    },
+  });
+
+UserSelectController.$inject = [
+  'UserService', '$scope'
```
jniles: Why do you need $scope?

client/src/js/components/bhUserSelect.js
```diff
@@ -0,0 +1,44 @@
+angular.module('bhima.components')
+  .component('bhUserSelect', {
+    templateUrl : 'modules/templates/bhUserSelect.tmpl.html',
+    controller  : UserSelectController,
+    transclude  : true,
+    bindings    : {
+      userId           : '<',
+      disable          : '<',
```
jniles: This should probably be `<?`.

client/src/modules/templates/bhUserSelect.tmpl.html
```diff
@@ -0,0 +1,24 @@
+<div ng-form="{{ $ctrl.name }}" bh-user-select ng-model-options="{ updateOn: 'default' }">
```
jniles, line 1: The `$ctrl.name` isn't passed in.  It would be best then to name this "UserForm"

client/src/js/components/bhUserSelect.js
```diff
@@ -0,0 +1,44 @@
...
+      onSelectCallback : '&?',
```
jniles: This probably should be required.  `'&'` instead of `'&?'`.

jniles (changes requested):
@lomamech, this is a good initial implementation! 

The component looks like it was adapted from a few other components, and not all the parts were properly adapted.  I've made some comments for broken HTML, and unused parts that I think can be removed.

Once those are out, I think it will be a very useful component!

client/src/js/components/bhUserSelect.js
```diff
@@ -0,0 +1,46 @@
...
+];
+
+/**
+ * User selection component
+ *
+ */
+function UserSelectController(Users) {
+  var $ctrl = this;
+  // default for form name
+  $ctrl.name = $ctrl.name || 'UserForm';
```
jniles: I think this should be in the `$onInit` callback.

jniles (approved):
Code LGTM.  I'm going to try and debug these tests before merging such a large change.

jniles:
Tests now passing on master.  @lomamech, can you rebase this?

jniles:
Looks good to me!

PR #1598, opened 2017-05-04
Transactions shown can exceed transactions in Database

### LINKED ISSUE #1593: (bug) Transactions shown can exceed transactions in database
I attempted to load all data from HEV locally and got this in the grid footer:
somethingweirdwithnumberoftransactions
_Fig 1: Show Transactions > Total Transactions_

I'm not sure how this case could occur, but we should check it out.

lomamech:
Transactions shown can exceed transactions in Database  
- Group by trans_id instead by record_uuid because the column
  'record_uuid' in group statement is ambiguous

closes #1593

### DIFF
--- a/server/controllers/finance/journal/index.js
+++ b/server/controllers/finance/journal/index.js
@@ -278,7 +278,7 @@ function queryTransactionAggregates(journalRows) {
       SUM(debit_equiv) as debit_equiv, COUNT(uuid) as totalRows
     FROM posting_journal
     WHERE record_uuid in (?)
-    GROUP BY record_uuid;
+    GROUP BY trans_id;
   `;
 
   // if there are no record uuids to search on we can optimise by not running the query


jniles:
`trans_id` and `record_uuid` should be the same thing.  Are you sure this fixes it?  I'm curious about why this fixes it.

`trans_id` is much more expensive to operate on, since it is a string.  So this will slow down the operation.

jniles:
@lomamech, it may be that in a live system, the total number of transactions is not refreshed.  I haven't looked at the code to confirm this, but perhaps when someone creates a sale and the filters are refreshed to include it, the bottom number remains at the amount first loaded.  How often are the total number of transactions reloaded?

lomamech:
@jniles , I just did some investigations in the database of VANGA I realized that when one executes the request:
```js
SELECT COUNT(DISTINCT (trans_id)) FROM posting_journal
```
And 
```js
SELECT COUNT(DISTINCT (record_uuid)) FROM posting_journal
```
Nous avons de nombre des lignes differents, ce qui n'est pas normale, d'ou la seule hypothese propabable etait qu'il existe des transactions qui ont le meme trans_id et de record_uuid different

We have different lines number, which is not normal, hence the only propable hypothesis was that there are transactions that have the same trans_id and record_uuid different

The following query enabled me to detect these transactions

```js
SELECT b.trans_id, COUNT(b.record_uuid) AS chiffre
 FROM (
SELECT DISTINCT posting_journal.record_uuid, posting_journal.trans_id
FROM posting_journal) AS b
GROUP BY b.trans_id
HAVING chiffre > 1
```
duplicate

jniles:
@lomamech, good investigation.  We should figure out why the `trans_id` and `record_uuid`s may be different!  This may be a significant bug :(

Some questions to ask:
 1. Est-ce que le tous les transactions affectE par ce probleme sont des VO ou CP ou IV?  Si oui, quel est le probleme avec les PROCEDUREs qui peut rendrer ses resultats?
 2. Est-ce que ce probleme exist dedans le grand livre aussi?
 3. Est-ce que le probleme persist toujours ou bien c'etait seulment pendant quelque dates?

Happy hunting!

lomamech:
@jniles
- Errors appear when canceling payment of invoices which instead of creating a new transaction uses an ID transaction that is already in the database

- The problem is also found in the general ledger

- Only during the month of February and March

jniles:
@lomamech, I'll take a look at this.  I have a feeling it may be a bug in the data rather than a bug in the software...

jniles:
Closing this because I believe this is a data problem, not a logic problem.  We can always re-implement this fix if get more information.

Thanks for you investigation @lomamech !

Issue #992, opened 2016-11-30
(bug) Posting Journal sorts transactions as text instead of transaction ids


jniles:
The Posting Journal does not properly order Transaction IDs by their numeric ID.  Instead, it is sorted as text.  This leads to some bizarre sorting, such as shown below, where "10" is immediately under "1".

![journalsorttransids](https://cloud.githubusercontent.com/assets/896472/20749825/3870f0aa-b6f4-11e6-9b46-bfc618a1367c.png)
_Fig 1: Journal Sorting Trans IDs_

This should be fixed for maximum UX.

jniles:
It should be noted that all registries have this problem.

troyev:
I'll try to take a closer look at this in the next week or so. I've got the issue reproduced with a Vanga restored database. The code appears to already be parsing out the integer and sorting on that, so I suppose there's a subtle problem with it.

troyev:
It looks like the sorting code in client/src/js/services/grid/GridSorting.js is not even being run. 
The registration functions are being called, but the transactionIds/sortByReference functions aren't called.
Next step is to find out why.

jniles:
Hey @troyev, good investigation!

It looks like we are going to need a new sorting function.   The GridSorting function expects a string with two periods as separators - such as `IV.HEV.123`.  [Here is the failing line of code](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/grid/GridSorting.js#L104).

We use this type of identifier everywhere but the Posting Journal.  In the PJ, we simply use `HEV1234`.  It would be best to simply write a function that recognizes how to parse this string and sort it numerically.

troyev:
I found out the Angular ui-grid is in fact calling the sortingAlgorithm if supplied (I put in an inline function and saw some log outputs). When sortingAlgorithm is undefined, it falls back to a default built-in sort.

The problem is that general-ledger.ctrl.js is getting an undefined function when it tries to get sortByReference or transactionIds from the sorting service. I suppose the functions aren't exposed/fetched correctly.

troyev:
Proposed fix here:

https://github.com/IMA-WorldHealth/bhima-2.X/pull/1610

jniles:
Closed in #1610

PR #1610, opened 2017-05-07
Journal transaction ID sorting bug, issue #992 

troyev:
It appears there were two main issues: one being that the sorting functions were not exposed correctly, so Undefined was passed to grid-ui, and two being that the sorting function didn't seem to be correct for strings like "HEV142 (2)".

Let me know your thoughts on the changes and if they work for you.

### DIFF
--- a/client/src/js/services/grid/GridSorting.js
+++ b/client/src/js/services/grid/GridSorting.js
@@ -31,35 +31,11 @@ function GridSortingService(util) {
    */
   function transactionIds(a, b, rowA, rowB, direction) {
     var first, second;
-    var nullValuesResult = this.gridApi.core.sortHandleNulls(a, b);
+    // determine integer (reference) value by extracting it from the transaction ID
+    // match returns an array of mathces - take the first element
+    first = Number(/[a-z,A-Z]*([0-9]*)/g.exec(a)[1]);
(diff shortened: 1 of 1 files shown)


client/src/js/services/grid/GridSorting.js
```diff
@@ -83,10 +59,11 @@ function GridSortingService(util) {
     // global sorting configuration
     gridOptions.enableSorting = true;
 
+      this.transactionIds = transactionIds;
```
jniles, line 62: Much clearer, thanks! :+1:

client/src/js/services/grid/GridSorting.js
```diff
@@ -31,35 +31,11 @@ function GridSortingService(util) {
    */
   function transactionIds(a, b, rowA, rowB, direction) {
     var first, second;
-    var nullValuesResult = this.gridApi.core.sortHandleNulls(a, b);
+    // determine integer (reference) value by extracting it from the transaction ID
+    // match returns an array of mathces - take the first element
+    first = Number(/[a-z,A-Z]*([0-9]*)/g.exec(a)[1]);
```
jniles, line 36: The comments are highly appreciated!

jniles (approved):
@troyev, great work!  I've pulled this locally and the changes work well in both "transaction mode" (grouped) and ungrouped mode.  This is really helpful, thanks!

Issue #1429, opened 2017-03-27
(refactor) Eliminate lint errors on the server


jniles:
The server is currently full of lint errors after the transition to eslint.  I propose that we target the server for cleanup.  The benefits will be:
 1. More syntactically similar code throughout the application
 2. An opportunity to remove and deprecate old code/libraries that are unused
 3. An opportunity to catch any bugs caught by a stricter linter
 4. The ability to lint as a test on Travis CI or other integration test suites.

The current state is:
```sh
✖ 3532 problems (1725 errors, 1807 warnings)
```

jniles:
There are still
```
✖ 351 problems (351 errors, 0 warnings)
```

I'll take a look at polishing these off while building a test dataset.

Issue #1519, opened 2017-04-19
(refactor/proposal) Use debug for logging


jniles:
There have been a few issues on logging.  For reference: https://github.com/IMA-WorldHealth/bhima-2.X/issues/210, https://github.com/IMA-WorldHealth/bhima-2.X/issues/565, https://github.com/IMA-WorldHealth/bhima-2.X/issues/1026.

We have previously used `winston` for logging, but I don't think it suits our needs.  [Winston](https://www.npmjs.com/package/winston) is a very large catch-all error handler, not suited to only logging errors from particular modules (for example, `db` or `invoices`.  This loses signal for all the noise.  Additionally, our HTTP errors should _probably_ be handled by `nginx`.

I propose we move our logging to [debug](https://www.npmjs.com/package/debug).  This was first proposed in a comment [here](https://github.com/IMA-WorldHealth/bhima-2.X/issues/1026#issuecomment-293283344).  We are needing to turn on logging for specific submodules, such as database errors, or only `db.transaction()` errors.  We haven't yet started using log files from `winston` (plus, we can use `pm2` to gather the logs if we need).

For all these reasons, I recommend switching to `debug` for better debugging in production.

jniles:
I'll be tackling this issue this week.

PR #1612, opened 2017-05-08
Linting the server side code

DedrickEnc:
This PR request amends the code by correcting errors of syntax in most of files in the server side based on the ```eslint``` rules.
Some rule was edited directly in the ```.eslintrc.js```

references #1429 

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


sfount:
This pull request contributes to https://github.com/IMA-WorldHealth/bhima-2.X/issues/1429

ghost:
:umbrella: The latest upstream changes (presumably 0856853eee7b7d87eb7fd9d75c5420629d748f6c) made this pull request unmergeable. Please resolve the merge conflicts.

.eslintrc.js
```diff
@@ -35,7 +35,15 @@ module.exports = {
...
       "allowAfterThis": true
-    }]
+    }],
+    "quotes" : ["error", "single", { "allowTemplateLiterals": true }],
+    "arrow-parens" : "off",
+    "arrow-body-style" : "off",
+    "func-names" : "off",
+    "prefer-arrow-callback" : "off",
+    "no-underscore-dangle" : "off",
+    "no-unused-vars" : "off",
```
jniles: In my experience, this rule is very helpful in cleaning up my code.  The only example I can think of that makes it annoying is ExpressJS error handlers, where functions must have arity 4 (and so you might not use `next`, but are required to pass it in).

What cases are you finding that it would be useful to turn it off?

.eslintrc.js
```diff
@@ -35,7 +35,15 @@ module.exports = {
     "func-names" : ["warn", "as-needed"],
     "no-underscore-dangle": ["error", {
       "allowAfterThis": true
-    }]
+    }],
+    "quotes" : ["error", "single", { "allowTemplateLiterals": true }],
+    "arrow-parens" : "off",
+    "arrow-body-style" : "off",
+    "func-names" : "off",
```
jniles: In my opinion, we should enable `func-names`, perhaps with the "[as-needed](http://eslint.org/docs/rules/func-names#options)" parameter.  Naming a function is better for logs and debugging, which is super helpful in production -- or at least until we have a better logger (like #1519).

I think this should take care of all places that the name cannot be assumed.  To understand how Javascript figures out a function name, Dr. Axel Rauschmayer has a r[eally detailed article on it](http://2ality.com/2015/09/function-names-es6.html).

.eslintrc.js
DedrickEnc: In the express, when passing next as parameter but not use that. that why I disabled this.

jniles: `next` isn't required.  The following code is perfectly valid expressjs middleware:
```js
app.get('/route', (req, res) => res.redirect('/other/route'));
```

If code doesn't use `next`, it is confusing to pass it into a function.  A better fix would be to simply remove `next` from the parameters.

As mentioned about, if it is error handling middleware (which we should only have one), we can just put comment telling `ESlint` to ignore that line.

.eslintrc.js
DedrickEnc: @jniles
I agree, but this rule lead me to even a function which process a promise 
```
.then(function (){
});
```
produces an error

```
.then(function someName(){
});
```
passes but produces an error in the execution

So do you think, I should declare those function on the top of their scopes?

jniles: Interesting ... I have never tried `.then(function someName() { /* ... */})` before.

It seems to me that callbacks do not need access to `this`, so they can be written more easily as arrow functions:
```js
.then((data) => {
  // do something without declaring this
});
```

That way the function's scope is the parent's scope and it behaves like any other code block.

DedrickEnc:
@jniles
Conflicts are fixed

server/config/interceptors.js
```diff
@@ -53,7 +53,8 @@ const SQL_STATES = {
  *
  * This error handler interprets all errors and sends them to the client.
  */
-exports.handler = function handler(error, req, res, next) {
+exports.handler = function handler(err, req, res, next) {
+  let error = err;
```
jniles, line 57: What's going on here?

jniles (commented):
I still get 456 lint errors, so the server work is not completed, but this is already a huge help!

Code changes look good to me.  :+1:

jniles:
@kwilu r+

ghost:
:pushpin: Commit 171343a has been approved by `jniles`

ghost:
:hourglass: Testing commit 171343a0190cbc8541211cee18b5bd0c05993787 with merge 5e5a3c8fee16e35469c206e6b8c3f6c71c7235de...

Issue #1609, opened 2017-05-05
(feature request) Dropdown to Edit Transaction Types in Posting Journal


jniles:
The Posting Journal does not allow one to view or edit the transaction type (`origin_id`).  This is critical for proper classification of transactions, research, and generation of meaningful reports (such as the cash flow report).

We need to be able to:
 1. Show the transaction's type in the Posting Journal
 2. Edit it via a <select> box in the ui-grid.

mbayopanda:
Completed by #1616

PR #1616, opened 2017-05-08
Fix journal transaction type

mbayopanda:
This PR fixes the issue #1609

### DIFF
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -527,11 +527,8 @@ function JournalController(Journal, Sorting, Grouping,
 
       vm.typeList = list.map(function (item) {
         item.hrText = $translate.instant(item.text);
+        vm.mapOrigins[item.id] = item.hrText;
         return item;
-      });
-
-      vm.typeList.forEach(function (type) {
-        vm.mapOrigins[type.id] = type.text;
       });
     })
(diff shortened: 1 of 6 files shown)


client/src/modules/journal/journal.js
```diff
@@ -493,7 +502,41 @@ function JournalController(Journal, Sorting, Grouping,
...
+    TransactionType.read()
+    .then(function (list) {
+      vm.mapOrigins = {};
+
+      vm.typeList = list.map(function (item) {
+        item.hrText = $translate.instant(item.text);
+        return item;
+      });
+
+      vm.typeList.forEach(function (type) {
```
jniles: You do two loops here where one will do:
```js
list.forEach(function (item) {
  item.hrText = $translate.instant(item.text);
  vm.mapOrigins[item.id] = item.text;
});
```

client/src/modules/journal/templates/transaction_type.html
```diff
@@ -0,0 +1,3 @@
+<div class="ui-grid-cell-contents" translate>
```
jniles: Do you need translate here?  Can you not use `hrText` instead and avoid translate?

jniles (approved):
LGTM!

PR #1622, opened 2017-05-09
Using the transanction type 'Invoicing' 

### LINKED ISSUE #1592: (refactor) Mark all Patient Invoices with an "Invoice" transaction type
At the moment, all incoming patient invoices are marked as "Generic Income".  It probably would be more useful to classify these in reports are income from patients, as opposed to the sale of other products (rent, gas, etc) and any other income the hospital generates.

lomamech:
Using the transanction type 'Invoicing' 
- Update procedure.sql using the id for 'Invoicing'
- Update test integration for transaction type

closes #1592


jniles (approved):
Looks good to me!

PR #1623, opened 2017-05-09
Refinements and Improvements of text of journal footer

### LINKED ISSUE #1581: (refactor) Journal Footer Improvements
The Posting Journal's footer shows the number of transactions in the view and the amount unposted.  This is really helpful in practice.  Here are some refinements from my experience (briefly) using it in production:
 1. The text should read "Showing {X} / {Y} Unposted Transactions"  At the moment, it's pretty confusing what it means at first glance.
 2. The text would be much nicer centered vertically.  It would be easier to quickly read.
 3. There is no need to use `$translate.instant()`.  Just put the `translate` filter in the HTML template.

lomamech:
Refinements and Improvements of text of journal footer

closes #1581


jniles (approved):
Looks good to me!

PR #1639, opened 2017-05-11
Propagate the changes in some transactions columns through the rest of

### LINKED ISSUE #1618: (proposal) Propagate changes in some transactions columns through the rest of the transaction
There are a couple columns in the database that we enforce always being the same across a transaction's rows.  Those are:
 1. `trans_date`
 2. `origin_id`
 3. `fiscal_year_id`
 4. `period_id`
... etc ...

I propose that we change our logic so that when a user edits a transaction date, all dates change.  Similarly, when they edit a transaction type, all transaction types change.  This will reduce the burden on a user of having to make sure every single column in a transaction is correct, and will remove the need for these client-side checks.

lomamech:
Propagate the changes in some transactions columns through the rest of  transaction
- Implentation of function propagate
- Improve date in data in bhima_test Database

closes #1618

### DIFF
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -285,24 +285,23 @@ function JournalController(Journal, Sorting, Grouping,
     });
   }
 
+  function updateSharedPropertyOnRow(rows, column, value){
+    rows.forEach(function (row) {
+      transactions.editCell(row, column, value, row[column]);
+      row[column] = (column === 'trans_date') ? new Date(value) : value;
+    });
+
+  }
+
   function propagate(column, value){
-    var propagateColumn = ['trans_date', 'entity_uuid', 'origin_id'];
-    
+    var propagateColumn = ['trans_date', 'entity_uuid', 'origin_id'];    
     //Check if the column updated must be propragated in all transaction
-    var checkPropagate = (propagateColumn.indexOf(column));
+    var hasSharedProperty = propagateColumn.indexOf(column) !== -1;
 
-    if(checkPropagate !== -1){
-      // For the old row
-      vm.transactions._entity.data.data.forEach(function (row) {
-        transactions.editCell(row, column, value, row[column]);
-        row[column] = (column === 'trans_date') ? new Date(value) : value;
-      });
-
-      // For the new row
-      vm.transactions._entity.newRows.data.forEach(function (row) {
-        transactions.editCell(row, column, value, row[column]);
-        row[column] = (column === 'trans_date') ? new Date(value) : value;
-      });
+    if (hasSharedProperty) {
+      // pass updates on to both the original rows and the new (pending) rows
+      updateSharedPropertyOnRow(vm.transactions._entity.data.data, column, value);
+      updateSharedPropertyOnRow(vm.transactions._entity.newRows.data, column, value);
     }
   }
 
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -7,7 +7,7 @@ JournalController.$inject = [
   'SessionService', 'NotifyService', 'TransactionService', 'GridEditorService',
   'bhConstants', '$state', 'uiGridConstants', 'ModalService', 'LanguageService',
   'AppCache', 'Store', 'uiGridGroupingConstants', 'ExportService', 'FindEntityService',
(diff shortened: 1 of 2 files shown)


client/src/modules/journal/journal.js
```diff
@@ -277,6 +277,33 @@ function JournalController(Journal, Sorting, Grouping,
   // API register function
   function onRegisterApi(gridApi) {
     vm.gridApi = gridApi;
+    
+    vm.gridApi.edit.on.afterCellEdit($scope, function(rowEntity, colDef, newValue, oldValue){
```
jniles, line 281: :+1:

client/src/modules/journal/journal.js
```diff
@@ -512,6 +539,9 @@ function JournalController(Journal, Sorting, Grouping,
   function editTransactionType(row) {
```
jniles, line 538: We should eventually make this a `ui-grid` dropdown editor - it would remove some code, and you wouldn't have to call `propogate` (because it would be caught by the `afterCellEdit` method you created).

I'll make an issue about it - no need to address it in the PR.

client/src/modules/journal/journal.js
```diff
@@ -277,6 +277,33 @@ function JournalController(Journal, Sorting, Grouping,
...
+        propagate(colDef.field,newValue);
+      }
+    });
+  }
+
+  function propagate(column, value){
+    var propagateColumn = ['trans_date', 'entity_uuid', 'origin_id'];
+    
+    //Check if the column updated must be propragated in all transaction
+    var checkPropagate = (propagateColumn.indexOf(column));
```
jniles: A clearer way to do write this is to do the ` value === -1` check here.  Like this:
```js
var hasSharedProperty = propagateColumn.indexOf(column) !== -1;

if (hasSharedProperty) {
  // propagate changes to all rows!
}
```

client/src/modules/journal/journal.js
```diff
@@ -277,6 +277,33 @@ function JournalController(Journal, Sorting, Grouping,
...
+
+    if(checkPropagate !== -1){
+      // For the old row
+      vm.transactions._entity.data.data.forEach(function (row) {
```
jniles: Lines 296-299 is the same code as lines 302-305.  Let's make it into the same funtion:

```js
function updateSharedPropertyOnRow(row) {
  transactions.editCell(row, column, value, row[column]);
  row[column] = (column === 'trans_date') ? new Date(value) : value;
}

// pass updates on to both the original rows and the new (pending) rows
vm.transactions._entity.data.data.forEach(updateSharedPropertyOnRow);
vm.transactions._entity.newRows.data.forEach(updateSharedPropertyOnRow);
```

jniles: ~~If you move this function outside of the `propagate` function, it will reduce the number of memory allocations that are performed.   V8 will not garbage collect it as often.~~
```js
function updateSharedPropertyOnRow(row) {
 // function body
}

function propagate(column, value) {
  // function body
}
```

*Edit*: this won't work since you depend on variables local to `propagate()`.  Ignore this comment.

jniles (approved):
@lomamech this PR is exactly correct.

I've made some comments to improve the clarity of the code (and a tiny bump to memory performance).  Let me know what you think of these suggestions - the code could be merged without them if you think they are not improvements.

client/src/modules/journal/journal.js
```diff
@@ -285,24 +285,23 @@ function JournalController(Journal, Sorting, Grouping,
...
-        row[column] = (column === 'trans_date') ? new Date(value) : value;
-      });
+    var hasSharedProperty = propagateColumn.indexOf(column) !== -1;
 
-      // For the new row
-      vm.transactions._entity.newRows.data.forEach(function (row) {
-        transactions.editCell(row, column, value, row[column]);
-        row[column] = (column === 'trans_date') ? new Date(value) : value;
-      });
+    if (hasSharedProperty) {
```
jniles: Beautiful!  This is a better proposal than my suggestion. :+1:

jniles:
@lomamech, once tests pass I'll merge this.

client/src/modules/journal/journal.js
jniles: Issue created: https://github.com/IMA-WorldHealth/bhima-2.X/issues/1643

PR #1644, opened 2017-05-11
feat(account): report opening balances

### LINKED ISSUE #1611: (feature request): Account reports need the real account opening balance
The account reports have a running balance that always begins at zero.  While this gives a nice total of how each line changes over time, we really need to understand the values in context of the total account balance.  It will also motivate our users to post from the Posting Journal to the General Ledger to update the account balances.

To be precise:
 1. The first line in the report should give the accounts opening balance.
 2. The running balance should begin at this balance and continue to the end.
 3. The final balance of the account should take into account the opening balance.

### LINKED ISSUE #1636: (proposal) Server utility for finding beginning balance for an account.
To completed the work on #1635, we need to have a way to get the beginning balance for any give date.

**Proposal**
I propose we write a server utility that can be tested as a standalone.  It will be in `server/controllers/finance/accounts.js` called `getOpeningBalanceForDate(accountId, date)`.  It should return a number - the balance of the account.  The usage pseudocode would be like this:
```js
const Accounts = require('../finance/accounts');

const date = new Date(2017, 03, 01);

Accounts.getOpeningBalanceForDate(accountId, date)
  .then(balance => {
   console.log(`The account opening balance is: ${balance}`);
  })
  .catch(next);
```

**Implementation Algorithm**
Internally, this function must do two things:
 1. Look up the period's opening balance via the `period_total` table.  First, look up the fiscal year:
```sql
SELECT id FROM fiscal_year WHERE start_date >= DATE(?) AND end_date <= DATE(?);
```
Then, find all the period totals for this fiscal year and account.
```sql
SELECT SUM(debit - credit) AS balance FROM period_total
WHERE period_total.fiscal_year_id = ? 
  AND period_total.end_date <= DATE(?)
  AND period_total.account_id = ?;
```

 2. Afterwards, we must figure out the balance up until that date.  Start by getting the period current period:
```sql
SELECT id FROM period WHERE start_date >= DATE(?) AND end_date <= DATE(?);
```
Then we can find the balance for this period up to today:
```sql
SELECT SUM(debit - credit) AS balance FROM general_ledger WHERE trans_date <= DATE(?) AND period_id = ? and account_id = ?;
```

Finally, we return the result of 1) + 2) as the final result!

jniles:
This commit builds a new feature for accounts that can discern their opening balance (from the
General Ledger) given a date and account id.  The major changes have been added in an "AccountExtra"
file that should be merged into accounts/index.js as soon as the major refactoring is finished.

This PR consists of two parts: the updated account report that now
contains an opening balance and the utility to find the opening balance
give a particular date and account_id.

Closes #1611.  Closes #1636.

-----

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


jniles:
@mbayopanda, can you give me a review of this code?

sfount:
@jniles I will review this code as soon as possible today.

server/controllers/finance/accounts/extra.js
```diff
@@ -0,0 +1,127 @@
...
+      getPeriodAccountBalanceUntilDate(accountId, date, fiscalYearId)
+    )
+    .then((previousPeriodClosingBalance) => {
+      balance += previousPeriodClosingBalance;
+      return getPeriodForDate(date);
+    })
+    .then(periodId =>
+      getComputedAccountBalanceUntilDate(accountId, date, periodId)
+    )
+    .then(runningPeriodBalance => balance + runningPeriodBalance);
```
sfount: These calculations are used here in financial reports - are they ever rounded? It looks like they would be subject to Javascript's rounding floating point issues?

utilities/translation/tfcomp.js
```diff
@@ -31,6 +31,8 @@ let frFileMissList = [];
 
 const jsonFiles = buildJsonFileArray();
 
+const notSwapFile = (fname) => !fname.includes('.swp');
```
sfount, line 34: Nice fix 👍 . Truly professional swap file abusers end up with .sw* to the ends of the earth. This should work well though.

sfount (commented):
These are very useful calculations that will be used by all reports. I've made some comments on the arithmetic used to combine balances, it would be useful to discuss the best practices for doing this going forward.

utilities/translation/tfcomp.js
jniles: I don't actually think this fix works, actually.  You can test it out.  I had limited success.

server/controllers/finance/accounts/extra.js
jniles: In theory, the currency filter takes care of that.  However, I could make them into fixed numbers if you think it's best.

server/controllers/finance/accounts/extra.js
```diff
@@ -0,0 +1,127 @@
...
+      getPeriodAccountBalanceUntilDate(accountId, date, fiscalYearId)
+    )
+    .then((previousPeriodClosingBalance) => {
+      balance += previousPeriodClosingBalance;
+      return getPeriodForDate(date);
+    })
+    .then(periodId =>
+      getComputedAccountBalanceUntilDate(accountId, date, periodId)
+    )
+    .then(runningPeriodBalance => (balance + runningPeriodBalance).toFixed(4));
```
sfount, line 124: This should work well for now! It would be good to research a library like `BigNumber` to make this a standard for reports going forward.

PR #1648, opened 2017-05-12
Remove DateService by the using of moment.js

### LINKED ISSUE #1633: (refactor/deprecate) Remove DateService
We developed `DateService` to have a single, tested location to do date manipulation.  This has been eclipsed by `moment` as discussed in #440.  It's time to remove `DateService` and get rid of that technical debt.  All instances could be easily replaced by `moment`'s API.

lomamech:
Remove DateService by the using of moment.js
- Update components bhDateInverval.js
- Remove all using of DateService
- Update the test E2E by the using of component date interval

closes #1633


client/src/js/components/bhDateInterval.js
```diff
@@ -59,22 +59,23 @@ function bhDateInterval(Dates, moment, bhConstants) {
...
 
   function day() {
-    vm.dateFrom = Dates.current.day();
+    vm.dateFrom = new Date();
     vm.dateTo = new Date();
   }
 
   function week() {
-    vm.dateFrom = Dates.current.week();
+    // Fix me if is necessary the first day of week is Sunday or Monday
```
jniles, line 67: Good point :+1:

client/src/modules/vouchers/modals/search.modal.js
```diff
@@ -40,11 +40,11 @@ function VoucherRegistrySearchModalController(ModalInstance, Dates, filters, Not
 
     // convert dates to strings
     if (parameters.dateFrom) {
-      parameters.dateFrom = Dates.util.str(parameters.dateFrom);
+      parameters.dateFrom = moment(parameters.dateFrom).format('YYYY-MM-DD');
```
jniles: It would be best if you used `bhConstants.date.format` instead of the string `'YYYY-MM-DD'`.  That way, we can easily change the date format if we need to in the future, and we can be guaranteed that all date formats are the same.

client/src/modules/patients/registry/search.modal.js
```diff
@@ -94,37 +89,6 @@ function PatientRegistryModalController(ModalInstance, Dates, params, DebtorGrou
     }
   }
 
-  // sets the date range for date inputs
-  function setDateRange(type, range) {
```
jniles, line 98: Excellent refactor!

jniles (commented):
@lomamech, this is a very good clean up! :+1:

There are a few places where you use plain strings 'YYYY-MM_DD'.  It would be best to use `bhConstants` in those locations so that we are guaranteed to be using the same date format everywhere.

Other than that, this is an excellent PR!

client/src/js/app.js
```diff
@@ -156,6 +156,7 @@ function constantConfig() {
     dates : {
       minDOB : new Date('1900-01-01'),
       format : 'dd/MM/yyyy',
+      formatDB : 'YYYY-MM-DD',
```
jniles, line 159: :+1:

jniles:
Looks good to me!

PR #1653, opened 2017-05-15
Remove combined_ledger view from the database

### LINKED ISSUE #1645: (refactor) Remove combined_ledger view from the database
The `combined_ledger` was a good idea until we discovered that MySQL doesn't use INDEXes on views.  This makes it a no-go for the two largest tables in our database.

Keeping `combined_ledger` around is just going to allow developers to continue to write non-performant code.  Someone should go through, refactor, and remove this view from all our queries in order to alleviate that technical debt.

lomamech:
Remove combined_ledger view from the database
- Using the union of combined ledger and posting journal instead combined ledger
- Remove view combined_ledger

closes #1645

### DIFF
--- a/server/controllers/dashboard/stats.js
+++ b/server/controllers/dashboard/stats.js
@@ -57,7 +57,7 @@ function invoiceStat(req, res, next) {
       );`;
 
   // query invoices
-  let sqlBalance =
+  const sqlBalance =
     `SELECT (debit - credit) as balance, project_id, cost
      FROM (
       (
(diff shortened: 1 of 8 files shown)


server/controllers/dashboard/stats.js
```diff
@@ -58,28 +58,27 @@ function invoiceStat(req, res, next) {
...
-        invoice.uuid NOT IN (
-          SELECT 
-            voucher.reference_uuid 
-          FROM voucher 
-          WHERE voucher.type_id = ${CANCELED_TRANSACTION_TYPE})
+      (
+        SELECT SUM(debit_equiv) as debit, SUM(credit_equiv) as credit, invoice.project_id, invoice.cost
+        FROM posting_journal
+        JOIN invoice ON posting_journal.record_uuid = invoice.uuid OR posting_journal.reference_uuid = invoice.uuid
+        WHERE invoice.uuid NOT IN (SELECT voucher.reference_uuid FROM voucher WHERE voucher.type_id = ${CANCELED_TRANSACTION_TYPE})
```
jniles: You should replace this `WHERE invoice.uuid NOT IN()` with `WHERE invoice.reversed = 0`.  It will be much faster :+1:

jniles (approved):
These changes are good!  I've made one comment to optimize the stats lookup.

After this is fixed, LGTM and we can merge :+1:

PR #1654, opened 2017-05-15
Personalized periods in journal search modal do not show date range

### LINKED ISSUE #1620: (bug) Personalized Periods in Journal Search Modal do not show date range
The date range feature is very useful to validate that a user has put in the correct date range, however, it does not show up when a user has customized their period selection.  It would be useful to render the dates (start and end) so that a user can see their selection before clicking submit to filter on those values.

lomamech:
Personalized periods in journal search modal do not show date range

closes #1620

### DIFF
--- a/client/src/modules/templates/bhPeriodSelect.tmpl.html
+++ b/client/src/modules/templates/bhPeriodSelect.tmpl.html
@@ -7,6 +7,10 @@
   <span ng-if="$ctrl.period.key !== $ctrl.NO_PERIOD_LIMIT_KEY && $ctrl.period.key !== $ctrl.CUSTOM_PERIOD_KEY">
   ({{$ctrl.period.start}}) - ({{$ctrl.period.end}})
   </span>
+
+  <span ng-if="$ctrl.period.customPeriodStart && $ctrl.period.customPeriodEnd">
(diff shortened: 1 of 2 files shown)


client/src/modules/templates/bhPeriodSelect.tmpl.html
```diff
@@ -7,6 +7,10 @@
   <span ng-if="$ctrl.period.key !== $ctrl.NO_PERIOD_LIMIT_KEY && $ctrl.period.key !== $ctrl.CUSTOM_PERIOD_KEY">
   ({{$ctrl.period.start}}) - ({{$ctrl.period.end}})
   </span>
+
+  <span ng-if="$ctrl.period.customPeriodStart && $ctrl.period.customPeriodEnd">
+    ({{ $ctrl.period.customPeriodStart | date: 'dd/mm/yyyy' }}) - ({{ $ctrl.period.customPeriodEnd | date: 'dd/mm/yyyy' }})
```
jniles: This is a good start!  Can you use the `bhConstants` date format to render this date?

jniles: This is a good start!  Can you use the `bhConstants` date format to render this date?

jniles:
LGTM!

PR #1657, opened 2017-05-16
Improvement to the Block User Feature

### LINKED ISSUE #1580: (refactor) Improvements to the "Block User" feature
The "block user" feature works correctly to block users who should not have access to the application from using it.  We can improve it further:
 1.  Move the block/unblock toggle from the users form into the dropdown.
 2. Grey out (using [text-muted](http://getbootstrap.com/css/#helper-classes-colors)) the users name.
 3. Remove the checkbox from the form, and instead put a text warning at the top: "This user has had their system access revoked"

This would make using the feature easier and clearer.

lomamech:
- Adding activate/Deactivate in DropDown
- Remove the checkBox from the form
- Adding test E2E for deactivate User system Access

closes #1580


jniles (commented):
@lomamech, the changes here are good.

The original issue (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1580) asks for the user name to be grayed out with `text-muted`.

It should be pretty easy to accomplish this with either a row template or a cellClass on the grid options.  Just set `class="text-muted"` on the user's name.  If it doesn't show up, you could also add a `strikethrough` class to show that the user no longer has access.

client/src/modules/users/templates/user.name.cell.html
```diff
@@ -0,0 +1,3 @@
+<div class="ui-grid-cell-contents text-muted" ng-class="{'strike' : row.entity.deactivated }">
```
jniles: It looks like `text-muted` will always be applied to this cell. I think it's probably better to only apply it on deactivation, similar to how our buttons work.

```html
<div class="ui-grid-cell-contents" ng-class="{'strike text-muted' : row.entity.deactivated }">

```

jniles (approved):
Good enough for me!

Issue #1628, opened 2017-05-09
(refactor) Implement FilterService across all Registries


jniles:
A new FilterService with a standard API was merged into `master` in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1566.  We should roll all these filters out over every registry in BHIMA, so that we have a standardized development and user experience.

The registries that need to be tackled are:
 - [x]  Cash Payments Registry (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1664)
 - [x] Patient Registry ( #1771)
 - [x] Invoice Registry (#1760)
 - [x] Voucher Registry (#1777)
 - [x] General Ledger ()
 - [x] Inventory Registry (#1789)
 - [x] Purchase Order Registry (#1907)
 - [x] Inventories in Stock ()
 - [x] Lots in Stock (#1938)
 - [x] Stock Movements (#1987)

Once these are complete, the maintenance burden on any one registry should be a lot lighter.

jniles:
I'm going to attempt to update the Cash Payments Registry to see how easy this is.

jniles:
This can be closed once #2020 and #2021 land.

jniles:
They have landed.

PR #1664, opened 2017-05-19
Refactor: Cash Payments use clientside filter service

jniles:
This commit migrates the cash page to use the filter service.  The
following changes have been made:
 1. The /cash route now uses FilterParser without autoParseStatements.
 2. The Cash Payments page now uses a search modal with tabs, default
 filters, and the like.
 3. All clientside code for filters has been migrated to leveraging the
 FilterService.

This commit re-writes the cash registry tests so that they use the new
filters, search modal.  Two new test harnesses had to be created:
 1. The `FilterService` and its display
 2. The `SearchModal` and its display

The current test syntax uses a giant SearchModal class that contains
some custom code for the Cash Search Modal.  In the future, it might be
prudent to use a mechanism that extends a base SearchModal class.

The FilterServer on the client is now ready for prime-time and can be
leveraged across all registries.

It also would be best to remove the `cash/search` route (and all similar
routes) to make sure that we only have a single GET route.

Partially addresses #1628.

-----
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


client/src/modules/cash/cash.service.js
```diff
@@ -123,60 +124,67 @@ function CashService(Modal, Api, Exchange, Session, moment, $translate) {
...
+    var periodDefined =
+      service.util.arrayIncludes(assignedKeys, ['period', 'custom_period_start', 'custom_period_end']);
+
+    if (!periodDefined) {
+      cashFilters.assignFilters(Periods.defaultFilters());
+    }
+
+    // assign default limit filter
+    if (assignedKeys.indexOf('limit') === -1) {
+      cashFilters.assignFilter('limit', 100);
```
sfount, line 171: It would be good to define a system wide default for some of these filters so that every module uses the same value (potentially through `bhConstants`). We can create an issue tracking the best practice for this.

client/src/modules/cash/payments/registry.js
```diff
@@ -5,7 +5,7 @@ angular.module('bhima.controllers')
 CashPaymentRegistryController.$inject = [
   'CashService', 'bhConstants', 'NotifyService', 'SessionService', 'uiGridConstants',
   'uiGridGroupingConstants', 'LanguageService', 'appcache', 'ReceiptModal', 'ModalService',
-  'GridSortingService', '$state', 'DepricatedFilterService'
```
sfount, line 8: 👍

client/src/modules/cash/payments/templates/search.modal.js
```diff
@@ -3,28 +3,35 @@ angular.module('bhima.controllers')
...
-  vm.submit = submit;
-  vm.cancel = Instance.close;
+  // @TODO ideally these should be passed in when the modal is initialised
+  //       these are known when the filter service is defined
+  var searchQueryOptions = [
+    'is_caution', 'reference', 'cashbox_id', 'user_id', 'reference_patient', 'currency_id', 'reversed',
+  ];
+
+  // assign already defined custom filters to searchQueries object
+  vm.searchQueries = util.maskObjectFromKeys(filters, searchQueryOptions);
```
sfount, line 27: This is a much cleaner abstraction 👍

server/controllers/finance/cash.js
```diff
@@ -149,7 +149,7 @@ function search(req, res, next) {
  * @description list all payment made
  */
 function listPayment(options) {
-  const filters = new FilterParser(options, { tableAlias : 'cash' });
+  const filters = new FilterParser(options, { tableAlias : 'cash', autoParseStatements : false });
```
sfount, line 152: Explicitly stating what will be filtered reads much clearer👍

server/lib/filter.js
```diff
@@ -149,7 +149,12 @@ class FilterParser {
     const tableString = this._formatTableAlias(tableAlias);
 
     if (this._filters[filterKey]) {
-      const preparedStatement = `${tableString}${columnAlias} = ?`;
+      let valueString = '?';
+      if (filterKey.includes(DEFAULT_UUID_PARTIAL_KEY)) {
```
sfount, line 153: There is a flag somewhere in this library that allows you to do this. I think in other controllers I've used the db (or util) parse UUIDs before passing the options to FilterParser. I'm not sure which one is better, as long as we clearly document that any key containing `uuid` will be converted it should be okay. 

I think this can be decided as the other controllers are converted to explicitly defining what should be filtered/ searched on.

sfount (approved):
The code here LGTM! moving other registries to standardised client filters will help maintenance a lot. 

I've made some comments on how these filter implementations will scale across other modules but for this PR there are no issues.

I have not fetched to test the feature but the tests look like they prove this feature.

jniles:
@mbayopanda, you can check out the tests for the clientside filter service here.

Issue #1638, opened 2017-05-10
Account Statement - basic use interface


mbayopanda:
This issue is a proposal for the account statement #1635  basic user interface

1. The user interface seems like the posting journal with those differences : 
- Display single line of accounts, there is not any grouping by transaction
- The default filter is based on : an account, a date range and a limit
![grid](https://cloud.githubusercontent.com/assets/5445251/25905228/4a1426b6-3556-11e7-98de-6db81e5e4d5f.png)

2. There is also a column filter such as in the posting journal
![grid filter](https://cloud.githubusercontent.com/assets/5445251/25905277/6c56b220-3556-11e7-9d88-9b7efd9300ad.png)

3. Comment line for analysis 
- The user can comment lines for need of analysis by selecting rows and clicking on the `comment` button
![grid comment](https://cloud.githubusercontent.com/assets/5445251/25905357/aea4e8e0-3556-11e7-945e-44f9bbc61bf2.png)

4. There is also a search menu like in the posting journal but with a new field for account as default filter
![grid search](https://cloud.githubusercontent.com/assets/5445251/25905507/0aaed6d2-3557-11e7-90be-542f9ddd02a6.png)

5. Utils drop down for exporting to PDF, CSV and printing
![grid utils](https://cloud.githubusercontent.com/assets/5445251/25905564/337f673e-3557-11e7-9957-e7f0327fea4d.png)

NOTE: The UI for details of a row is not described here

jniles:
Closed in #1672.

PR #1667, opened 2017-05-22
Account period totals

lomamech:
Build Account by Period Totals Report  …
- Update uigrid of genereal ledger, Added columns for each monthly total periods
- Create template for the balance cell for eachn period
- Adding a function who returns all period of current fiscal Year

closes #1637 .  Closes #1676.

@jniles I will wait for your fead back finally to be able to restore the old presentation ( @mbayopanda  ) of the u grid of general ledger

PR #1672, opened 2017-05-23
Account statement - basic user interface

mbayopanda:
This PR is about the issue #1638


client/src/js/services/FilterService.js
```diff
@@ -122,29 +122,29 @@ function FilterService(Store) {
...
-  }
+  };
 
   // replaces current filters with filters from cache
   FilterList.prototype.loadCache = function loadCache(storedCache) {
     Object.keys(storedCache).forEach(function (key) {
       var cached = storedCache[key];
       var currentFilter = this._filterIndex[key];
-      if(currentFilter) {
+      if (currentFilter) {
```
jniles, line 136: :+1:

client/src/modules/account_statement/account_statement.ctrl.js
```diff
@@ -0,0 +1,288 @@
...
+  // expose to the view
+  vm.selectedRows = [];
+  vm.enterprise = Session.enterprise;
+
+  // grid definition ================================================================
+  vm.gridApi = {};
+
+  vm.gridOptions = {
+    enableColumnMenus        : false,
+    enableCellEdit           : false,
```
jniles: We don't need the `enableCellEdit` option for this module.

client/src/modules/account_statement/account_statement.ctrl.js
```diff
@@ -0,0 +1,288 @@
...
+      footerCellClass  : 'text-right',
+    },
+
+    { field              : 'description',
+      displayName        : 'TABLE.COLUMNS.DESCRIPTION',
+      headerCellFilter   : 'translate',
+      visible            : false,
+      footerCellTemplate : '<i></i>' },
+
+    { field            : 'uuid',
```
jniles: Do we ever need to see the uuid?

client/src/modules/account_statement/account_statement.html
```diff
@@ -0,0 +1,94 @@
...
+
+      <div class="toolbar-item">
+        <button data-method="search" class="btn btn-default" ng-click="AccountStateCtrl.openSearchModal()">
+          <i class="fa fa-search"></i> <span class="hidden-xs" translate>FORM.BUTTONS.SEARCH</span>
+        </button>
+      </div>
+
+      <div class="toolbar-item">
+        <button data-method="comment" class="btn btn-default" ng-disabled="!AccountStateCtrl.selectedRows.length" ng-click="AccountStateCtrl.commentRows()">
+          <i class="fa fa-sticky-note-o"></i> <span class="hidden-xs" translate>FORM.LABELS.COMMENT</span>
```
jniles, line 45: :+1:

client/src/modules/account_statement/account_statement.html
```diff
@@ -0,0 +1,94 @@
...
+          id="account-statement-grid"
+          class="grid-journal"
+          ng-style="AccountStateCtrl.filterBarHeight"
+          ui-grid="AccountStateCtrl.gridOptions"
+          ui-grid-auto-resize
+          ui-grid-resize-columns
+          ui-grid-move-columns
+          ui-grid-selection
+          ui-grid-cellNav
+          ui-grid-edit
```
jniles: Do we need ui-grid edit?

client/src/modules/account_statement/account_statement.html
```diff
@@ -0,0 +1,94 @@
...
+        <div
+          id="account-statement-grid"
+          class="grid-journal"
+          ng-style="AccountStateCtrl.filterBarHeight"
+          ui-grid="AccountStateCtrl.gridOptions"
+          ui-grid-auto-resize
+          ui-grid-resize-columns
+          ui-grid-move-columns
+          ui-grid-selection
+          ui-grid-cellNav
```
jniles: Do we need ui-grid cellNav?

client/src/modules/account_statement/account_statement.service.js
```diff
@@ -0,0 +1,113 @@
...
+  service.filters = accountStatementFilters;
+  service.openCommentModal = openCommentModal;
+  service.commentAccountStatement = commentAccountStatement;
+
+  /**
+   * @method commentAccountStatement
+   * @param {object} params - { uuids: An array of uuid, string: comment }
+   */
+  function commentAccountStatement(params) {
+    return $http.put(baseUrl.concat('comment'), { params : params })
```
jniles: This should probable be plural (`comments`) so that it's clear that multiple comments can be submitted.

client/src/modules/account_statement/account_statement.service.js
```diff
@@ -0,0 +1,113 @@
...
+      .then(util.unwrapHttpResponse);
+  }
+
+  /**
+   * @method openCommentModal
+   * @param {object} request
+   */
+  function openCommentModal(request) {
```
jniles, line 37: In the future, we could make this a component that would plug into every grid.  This is fine for a first implementation.

client/src/modules/account_statement/account_statement.ctrl.js
```diff
@@ -0,0 +1,288 @@
...
+  }
+  // end grid defintion =============================================================
+
+  // comment selected rows
+  vm.commentRows = function commentRows() {
+    AccountStatement.openCommentModal({ rows : vm.selectedRows })
+    .then(function (res) {
+      if (!res) { return; }
+      Notify.success('ACCOUNT_STATEMENT.SUCCESSFULLY_COMMENTED');
+      load(AccountStatement.filters.formatHTTP(true));
```
sfount: Is it necessary to reload all of the data in the account statement just because comments are added? I propose that on confirmation from the server we update the data locally so that it reflects these changes. 

With editing financial information I think it is important to ensure the client and the server have the same data, however with comments I think an optimistic local update should work well. 

What do you think?

client/src/modules/account_statement/account_statement.service.js
```diff
@@ -0,0 +1,113 @@
...
+    // assign default period filter
+    var periodKeys = ['period', 'custom_period_start', 'custom_period_end'];
+    var periodDefined = util.arrayIncludes(assignedKeys, periodKeys);
+
+    if (!periodDefined) {
+      accountStatementFilters.assignFilters(Periods.defaultFilters());
+    }
+
+    // assign default limit filter
+    if (assignedKeys.indexOf('limit') === -1) {
```
jniles, line 97: @Sfount, what was your idea about these limits?  Should this be done automatically?

client/src/modules/account_statement/account_statement.ctrl.js
```diff
@@ -0,0 +1,288 @@
...
+
+  // startup
+  function load(options) {
+    vm.loading = true;
+    vm.hasErrors = false;
+
+    GeneralLedger.read(null, options)
+    .then(function (data) {
+      vm.gridOptions.data = data;
+      vm.loading = false;
```
sfount: This should not be needed as it is included in the `.finally`

client/src/modules/account_statement/modals/comment.modal.js
```diff
@@ -0,0 +1,41 @@
+angular.module('bhima.controllers')
+  .controller('CommentAccountStatementController', CommentAccountStatementController);
+
+// DI
+CommentAccountStatementController.$inject = [
+  '$uibModalInstance', 'data', 'AccountStatementService', 'NotifyService',
```
sfount: What does data mean here? 

I see from the context of the controller calling it that it contains the selected rows from the grid - I think the name could better reflect what is happening here! 

Something like `gridSelectionContext`? Whatever makes sense to you.

client/src/modules/account_statement/modals/comment.modal.js
```diff
@@ -0,0 +1,41 @@
...
+
+    var params = {
+      uuids : uuids(vm.rows),
+      comment : vm.comment,
+    };
+
+    AccountStatement.commentAccountStatement(params)
+      .then(function (res) {
+        if (!res) { return; }
+        Instance.close(res);
```
sfount: Again I think `.finally` if this promise if resolved or rejected. `Instance.close` will be called twice in the case of success.

client/src/modules/account_statement/modals/comment.modal.js
```diff
@@ -0,0 +1,41 @@
...
+      })
+      .catch(Notify.handleError)
+      .finally(Instance.close);
+  }
+
+  // get rows uuid
+  function uuids(rows) {
```
jniles: A better name for this function would be something like "getRowIds()" or something.  It's confusing since `uuid()` on the server.

server/controllers/finance/generalLedger/index.js
```diff
@@ -6,46 +6,87 @@
...
+ * @function find
+ *
+ * @description
+ * This function filters the general ledger by query parameters passed in via
+ * the options object.  If no query parameters are provided, the method will
+ * return all items in the general ledger
  */
-function list(req, res, next) {
+function find(options) {
+  const filters = new FilterParser(options, { tableAlias : 'gl', autoParseStatements : false });
```
sfount, line 42: 👍

test/end-to-end/shared/GridUtils.js
```diff
@@ -111,6 +111,28 @@ function selectRow(gridId, rowNum) {
...
+  * that you want to check the value of
+  * @param {string} value a regex or string of the value you expect in that cell
+  *
+  * @example
+  * <pre>
+  *   gridTestUtils.expectCellValueMatch('myGrid', 0, 2, 'CellValue');
+  * </pre>
+  *
+  */
+function expectCellValueMatch(gridId, row, col, value) {
```
jniles, line 131: :+1:

client/src/modules/journal/modals/search.modal.html
```diff
@@ -30,14 +30,16 @@
...
 
-        <bh-account-select
-          account-id="ModalCtrl.options.account_id"
-          name="account_id"
-          on-select-callback="ModalCtrl.onSelectAccount(account)"
-          exclude-title-accounts="true"
-          required="0">
-          <bh-clear on-clear="ModalCtrl.clear('account_id')"></bh-clear>
-        </bh-account-select>
+        <div ng-if="!ModalCtrl.hasDefaultAccount">
```
sfount: A comment explaining when this switch is used would be useful for developers working on either the Journal or the Account Statement module.

server/controllers/finance/reports/account_statement/report.handlebars
```diff
@@ -0,0 +1,67 @@
...
+            </tr>
+          {{/each}}
+        </tbody>
+      </table>
+    </div>
+  </div>
+
+  <div class="row">
+    <div class="col-xs-4">
+      {{translate 'TABLE.COLUMNS.TOTAL'}} {{translate 'TABLE.COLUMNS.DEBIT'}}: <strong>{{currency (sum rows 'debit_equiv')}}</strong>
```
jniles: Could you move this `sum` function to SQL?  It would be much more efficient.

server/controllers/finance/reports/account_statement/report.handlebars
```diff
@@ -0,0 +1,67 @@
...
+    </div>
+
+    <div class="col-xs-4">
+      {{translate 'TABLE.COLUMNS.TOTAL'}} {{translate 'TABLE.COLUMNS.CREDIT'}}: <strong>{{currency (sum rows 'credit_equiv')}}</strong>
+    </div>
+
+    <div class="col-xs-4">
+      {{translate 'TABLE.COLUMNS.RESULT'}}: <strong>{{currency (substract (sum rows 'debit_equiv') (sum rows 'credit_equiv'))}}</strong>
```
jniles: Could you move this `sum` function to SQL?  It would be much more efficient.

client/src/modules/journal/modals/search.modal.js
```diff
@@ -22,11 +27,30 @@ function JournalSearchModalController(Instance, Projects, Notify, Store, filters
   // assign already defined custom filters to searchQueries object
   vm.searchQueries = util.maskObjectFromKeys(filters, searchQueryOptions);
 
+  // has default account
+  if (options.hasDefaultAccount) {
```
sfount: It would be useful to document the behaviour of `hasDefaultAccount` somewhere in a comment. The branching logic here could be difficult to follow if you didn't see that it was used by different modules with different needs.

server/controllers/finance/generalLedger/index.js
```diff
@@ -6,46 +6,87 @@
...
       JOIN user u ON u.id = gl.user_id
-    ORDER BY gl.trans_date DESC;
-    `;
+      JOIN currency c ON c.id = gl.currency_id
+      LEFT JOIN entity_map em ON em.uuid = gl.entity_uuid
+      LEFT JOIN document_map dm1 ON dm1.uuid = gl.record_uuid
+      LEFT JOIN document_map dm2 ON dm2.uuid = gl.reference_uuid
+  `;
+
+  filters.period('period', 'trans_date');
```
jniles, line 65: Nice improvement with FilterParser :+1:

sfount (commented):
This looks like a great start for the Account Statement module! I've made comments with some minor suggestions but none of them should be blocking if this needs to be merged.

jniles (commented):
@mbayopanda, there are some very nice features in this PR.

 1. Sharing the same filter modal between the Posting Journal and the Account Statement is a really good idea.  Your implementation looks fine to me.
 2. Migrating the GL to FilterParser is a really nice improvement.

Some things to improve in the future, but are not necessarily blocking: 
 1. We already know that this component will be used on both the Posting Journal and the Account Statement.  It would be ideal if we could design this feature so that it is easily transposed to each module, potentially as a component or directive.

I've made some other comments that may need to be addressed.

client/src/modules/account_statement/account_statement.service.js
sfount: I think this whole process can be improved, my comment on the previous PR was the the hard-coded default value `100` could be shared across modules with `bhConstant`.

client/src/modules/account_statement/modals/comment.modal.js
mbayopanda: In this section the value returned is not false, in the finally clause we have `.finally(Instance.close)` which return a falsy value

client/src/modules/account_statement/account_statement.service.js
jniles: Ah, gotcha.  Thanks!

client/src/modules/journal/modals/search.modal.js
```diff
@@ -22,11 +27,38 @@ function JournalSearchModalController(Instance, Projects, Notify, Store, filters
   // assign already defined custom filters to searchQueries object
   vm.searchQueries = util.maskObjectFromKeys(filters, searchQueryOptions);
 
+  /**
+   * hasDefaultAccount is used to set a default account selection behavior
+   * if the search modal need to set account selection in default query panel we can send it
+   * as parameters
+   * @example
+   * <pre>
```
jniles, line 35: Very nice comments :+1:

client/src/modules/journal/modals/search.modal.js
```diff
@@ -22,11 +27,38 @@ function JournalSearchModalController(Instance, Projects, Notify, Store, filters
...
   if (filters.limit) {
     vm.defaultQueries.limit = filters.limit;
   }
 
+  // assing default account
+  if (filters.account_id) {
+    vm.defaultQueries.account_id = filters.account_id;
+  }
+
+  Account.read()
```
jniles, line 53: You don't need this .. I don't think.  The `onSelectAccount()` callback should return the entire account object.  Just do this:

```js
vm.onSelectAccountCallback = function (account) {
  vm.currentAccountLabel = account.hrLabel;
  // other logic
});

angular.forEach(vm.searchQueries, function (value, key) {
    if (key === 'account_id') {
      changes.post({ key : key, value : value, displayValue : vm.currentAccountLabel });
    }
    // ... etc ...
  }
});
```

This way, you don't have to download the entire account list.

jniles (approved):
LGTM.

I've made one final comment to avoid an HTTP request and improve performance.  Let me know what you think.

jniles (approved):
LGTM.

I've made one final comment to avoid an HTTP request and improve performance.  Let me know what you think.

jniles:
@mbayopanda, the merge of #1667 seems to have broken this :(

client/src/modules/journal/modals/search.modal.js
jniles: Another way to reduce the data transfer is to do `Account.read(filters.account_id)` if it exists.  That way, we only load one account if we need it's `hrlabel` and otherwise we do not load any accounts at all.

I'll merge this and try and implement it to see if it works.

PR #1685, opened 2017-05-26
Refactor db one

### LINKED ISSUE #1670: (refactor) Use db.one() for identity checks
On the server, we have a database query method `db.one()` for retrieving single rows from the database.  This is useful for lookups that are on a single identifier (.e.g. `where id =?`).   There are a number of modules that could benefit from this on the server, including, but not limited to, `creditor_groups`, `debtor`, `cashboxes.js`, etc...

We should either commit entirely to `db.one()` or find a better way that that we practice DRY programming.

lomamech:
Refactor use dbOne for identity checks

closes #1670

### DIFF
--- a/server/controllers/admin/functions.js
+++ b/server/controllers/admin/functions.js
@@ -13,10 +13,7 @@ function lookupFunction(id) {
     `SELECT id, fonction_txt FROM fonction
     WHERE fonction.id = ?`;
 
-  return db.one(sql, [id])
-  .then(function (row) {
-    return row;
-  });
+  return db.one(sql, [id]);
 }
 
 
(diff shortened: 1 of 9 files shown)


jniles (changes requested):
@lomamech, all the checks that look like this:
```js
return db.one(sql, [id])
  .then(function (row) {
    if (row.length === 0) {
      throw new NotFound('...');
    }
  })
```
... no longer need the `if (row.length === 0) {` check.  That's what `db.one()` does for you.  You can remove all those `if ()` statements and their errors.

For reference, [here is the db.one() source code](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/db/index.js#L112).

jniles (commented):
This is closer!

```js
return db.one(sql, [id])
  .then(function (row) {
    return row; 
  });
```
is the same thing as
```js
return db.one(sql, [id]);
```
Go ahead and remove these extra promise calls.

jniles (approved):
Excellent refactor :+1:

PR #1689, opened 2017-05-26
Adding Notify warning in Reporting Modules When document not found

### LINKED ISSUE #854: (fix) Reporting Modules must notify for file not found
Reporting modules must notify user when document are not found on the disk. 

The deletion of reports must also be implemented.

lomamech:
Adding Notify warning in Reporting Modules When document not found

closes #854


jniles:
I'm moving the review to @mbayopanda since he was the original author on the issue.

jniles (approved):
LGTM.  :+1:

PR #1690, opened 2017-05-26
refactor(users): make activate/deactivate red

jniles:
This commit changes the user activate/deactivate to a red color and puts
it under a divider to indicate that it's a stronger action than the
other options.

---

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/client/src/modules/users/templates/grid/action.cell.html
+++ b/client/src/modules/users/templates/grid/action.cell.html
@@ -1,17 +1,34 @@
-<div class="ui-grid-cell-contents text-right" uib-dropdown dropdown-append-to-body uib-dropdown-toggle>
-  <a href>
+<div class="ui-grid-cell-contents text-right" uib-dropdown dropdown-append-to-body>
+  <a uib-dropdown-toggle href>
     <span data-method="action" translate>FORM.BUTTONS.ACTIONS</span>
     <span class="caret"></span>
   </a>
 
   <ul data-action="{{ rowRenderIndex}}"  class="dropdown-menu-right" uib-dropdown-menu>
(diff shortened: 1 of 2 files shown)


client/src/modules/users/templates/grid/action.cell.html
```diff
@@ -1,17 +1,34 @@
...
+    <li class="divider"></li>
+    <li ng-if="!row.entity.deactivated">
+      <a data-method="activated" ng-click="grid.appScope.activatePermissions(row.entity, 1, 'FORM.DIALOGS.CONFIRM_DEACTIVATION')" href>
+        <span class="text-danger">
+          <i class="fa fa-lock"></i> <span translate>FORM.LABELS.DEACTIVATE</span>
+        </span>
+      </a>
+    </li>
+    <li ng-if="row.entity.deactivated">
+      <a data-method="activated" ng-click="grid.appScope.activatePermissions(row.entity, 0, 'FORM.DIALOGS.CONFIRM_ACTIVATION')" href>
```
sfount, line 27: Should these have the same `data-method`? Or is the attribute not used here?

jniles: I think the idea was the tests could use the same tag then (the `ngIf` removes the other tag when the conditions aren't met).  It does make for less precise testing.

Would you like me to change it to be more semantically correct?

sfount: Seems fine to me - if it's needed by a test in the future we can update it then.

jniles:
:(

Issue #1631, opened 2017-05-10
(feature request) Automatically create a Patient when registering an Employee


jniles:
When an employee is registered in the BHIMA software, it would be ideal to create a Patient entity for them in the `patient` table.  This implies that the `patient` information is a _subset_ of the `employee` information.

To accomplish this goal, the following changes needs to be made:
 1. `POST /employee` must result in a `db.transaction()` to ensure that either the patient or employee are not created without the other.
 2. The employee registration page must contain both Origin Location and Current Location location selects.
 3. The employee registration page must contain a hospital number - this is likely going to be their medical license number in practice.
 4. The `employee` table should be modified to contain a `patient_uuid` foreign key.

DedrickEnc:
I think, it is a good idea. But the all employee page should be refactored to follow our current standard

DedrickEnc:
already addressed by #1691

PR #1674, opened 2017-05-24
Remove search end points

lomamech:
Remove all entity/search endpoints with  …
- Patients
- Inventory
- vouchers modals
- Patients invoices
- Update some integration and E2E test
Remove all entity search endpoints in Cash Registry and update test integration

closes #1665

Issue #1665, opened 2017-05-19
refactor: remove all {entity}/search endpoints


jniles:
The cash payments registry uses a switch to determine whether to send requests `GET cash/search` or `GET cash/`.  [Here it is](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/cash/payments/registry.js#L120).  This switch should not exist - they both map to the [listPayment](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/cash.js#L125) function on the server.

I assume this pattern has been used in many places.  It was helpful before we standardized filters on the server + client.  As a way of removing technical debt, we should remove this forked route and only support a single endpoint to handle all filtering (or lack of filtering) requests.

Issue #1447, opened 2017-03-29
(refactor) Deprecate usage of bhFormDefaults


jniles:
The [bhFormDefaults](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/directives/bhFormDefaults.js) directive was intended to turn off autocomplete on all `<input>` fields of a form as described in the [MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion).  Chrome seems [bent on ignoring developers](https://bugs.chromium.org/p/chromium/issues/detail?id=587466) so this trick no longer really works.  We should deprecate and remove `bhFormDefaults` as it doesn't do anything.

PR #1691, opened 2017-05-27
Dedrickenc employee patient

DedrickEnc:
# Summary

This PR is refactoring the employee registration page and his test (integration and end to end)

# Details
here are some tasks performed : 

1. Update the database schema

2. Update  the server Interface

3. Refactor the user Interface

4. Add a registry for an employees

5. Clean the code

6. Integration test

7. End to End test

# More details

- The employee controller was fixed to create a debtor and a creditor properly

- The ```debtor_uuid``` reference was removed from the ```employee``` table 

- Now creating an **employee** will create a related **patient** directly  

- The UI interface for employee management was completely updated to follow the new standard

- **Employee** will have a card like a **patient** when he is created in the system

- Employee folder restructured as patient folder

- Some duplicated keys are removed from the translation files 

- The employee registry will keep ```update```, ```invoices```, ```payments``` 

- The old interface of employee for searching was removed and his test of integration 

- The all employee end-to-end test was skip now it is not

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

#1631


client/src/modules/employees/employee.service.js
```diff
@@ -1,81 +1,75 @@
 angular.module('bhima.services')
   .service('EmployeeService', EmployeeService);
 
-EmployeeService.$inject = ['$http', 'util'];
+EmployeeService.$inject = ['DepricatedFilterService', '$uibModal', 'PrototypeApiService'];
```
jniles: Just a note - `DepricatedFilterService` is name "deprecated" because new modules shouldn't be using it.  A much better filter service (`FilterService`) was introduced about a month ago in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1566 that supports stronger filtering.

client/src/modules/employees/registration/employees.html
```diff
@@ -0,0 +1,513 @@
+<div class="flex-header">
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static">{{ "TREE.PAYROLL" | translate }}</li>
```
jniles: Please use the `translate` directive instead of the `translate` filter.  The repository has adopted this practice since https://github.com/IMA-WorldHealth/bhima-2.X/issues/1074.

client/src/modules/employees/registration/employees.html
```diff
@@ -0,0 +1,513 @@
...
+                        ng-model="EmployeeCtrl.employee.nb_enfant"
+                        autocomplete="off">
+                     <div class="help-block" ng-messages="OptionalForm.nb_enfant.$error" ng-show="EmployeeRegistrationForm.$submitted">
+                      <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+                     </div>
+                    </div>
+                  </div>
+
+                  <!-- Employee service-->
+                  <div class="form-group"
```
jniles, line 332: Does an employee always have to have a service?  Where do guards and cleaning staff go?

client/src/modules/employees/registration/employees.html
```diff
@@ -0,0 +1,513 @@
...
+                        name="daily_salary"
+                        ng-model="EmployeeCtrl.employee.daily_salary"
+                        autocomplete="off">
+                      <div class="help-block" ng-messages="DetailForm.daily_salary.$error" ng-show="EmployeeRegistrationForm.$submitted">
+                        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+                      </div>
+                    </div>
+                  </div>
+
+                  <!-- Employee bank -->
```
jniles, line 456: What is the difference between `bank` and `bank_account`?  How are they meant to be used?

client/src/modules/employees/registration/employees.html
```diff
@@ -0,0 +1,513 @@
...
+                        name="bank_account"
+                        ng-model="EmployeeCtrl.employee.bank_account"
+                        autocomplete="off">
+                      <div class="help-block" ng-messages="DetailForm.bank_account.$error" ng-show="EmployeeRegistrationForm.$submitted">
+                        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+                      </div>
+                    </div>
+                  </div>                  
+
+                  <!-- Employee locked state-->
```
jniles, line 492: What application logic should happen if an employee is locked?  Can their patient entity still receive treatment?

client/src/modules/employees/registration/employees.js
```diff
@@ -0,0 +1,76 @@
...
+
+  // Expose employee to the scope
+  vm.employee = {};
+
+  // Expose methods to the scope
+  vm.submit = submit;
+
+  // Loading Grades
+  Grades.read(null, { detailed : 1 }).then((data) => {
+    data.forEach((g) => {
```
jniles: Using ES2015+ is excellent!  However, we cannot use it on the client ... it breaks the production build.  To do this, we'll need to use a transpiler like [babeljs](https://babeljs.io/) or [TypeScript](https://www.typescriptlang.org/).

Please switch these back to ES5 until we have a transpiler included.

Here is the tracking issue: https://github.com/IMA-WorldHealth/bhima-2.X/issues/1439

client/src/modules/employees/registration/employees.html
```diff
@@ -0,0 +1,513 @@
...
+                      <span ng-show="DetailForm.hospitalNumber.$error.unique" class="glyphicon glyphicon-remove form-control-feedback" id="unique-error-icon"></span>
+                      <span ng-show="DetailForm.hospitalNumber.$pending.unique" class="glyphicon glyphicon-hourglass form-control-feedback"></span>
+
+                      <div class="help-block" ng-messages="DetailForm.hospitalNumber.$error" ng-show="DetailForm.$submitted || DetailForm.hospitalNumber.$error.unique">
+                        <p ng-messages-include="modules/templates/messages.tmpl.html"></p>
+                      </div>
+                    </div>
+                  </div>
+
+                  <!-- Employee creditor group-->
```
jniles, line 168: Why does an employee need a creditor group?

client/src/modules/employees/registration/employees.js
```diff
@@ -0,0 +1,76 @@
...
+    vm.functions = data;
+  }).catch(Notify.handleError);
+
+
+  // submit the data to the server
+  function submit(employeeForm) {
+    var promise;
+
+    if (employeeForm.$invalid) { return Notify.danger('FORM.ERRORS.INVALID');}
+    if (!employeeForm.$dirty) { return Notify.danger('FORM.ERRORS.NO_DATA_PROVIDED'); }
```
jniles: I think this condition should already be caught in the `employeeForm.$invalid` case.  Are you seeing something different?

client/src/modules/employees/registry/registry.html
```diff
@@ -0,0 +1,80 @@
+<div class="flex-header static">
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static" translate>TREE.PAYROLL</li>
```
jniles, line 4: :+1: translate directive

client/src/modules/employees/registry/registry.js
```diff
@@ -0,0 +1,245 @@
...
+];
+
+/**
+ * Employee Registry Controller
+ *
+ * This module is responsible for the management of Employe Registry.
+ */
+function EmployeeRegistryController($state, Employees, Notify, AppCache,
+  util, Receipts, uiGridConstants, $translate,
+  Columns, Sorting, bhConstants, Filters) {
```
jniles: It doesn't look like you actually use `GridSorting`.  Could you remove all imports that you don't need?

client/src/modules/employees/registry/registry.js
```diff
@@ -0,0 +1,245 @@
...
+      visible          : false 
+    },
+    { name          : 'actions',
+      displayName   : '',
+      cellTemplate  : '/modules/employees/templates/action.cell.html',
+      enableSorting : false 
+    },
+  ];
+
+  /** TODO manage column : last_transaction */
```
jniles: I don't think this column exists.

client/src/modules/employees/registry/registry.js
```diff
@@ -0,0 +1,245 @@
...
+    { field            : 'text',
+      displayName      : 'TABLE.COLUMNS.GRADE',
+      headerCellFilter : 'translate',
+    },
+    { field            : 'nb_spouse',
+      displayName      : 'FORM.LABELS.NB_SPOUSE',
+      headerCellFilter : 'translate',
+      type             : 'number',
+    },
+    { field            : 'nb_enfant',
```
jniles: Do we need to see the `nb_enfant` by default?   It seems to me that some of these columns aren't needed on the main page and can be hidden by default with the option to expose them later.

client/src/modules/employees/registry/registry.js
```diff
@@ -0,0 +1,245 @@
...
+  // this function loads employees from the database with search parameters
+  // if passed in.
+  function load(parameters) {
+    // flush error and loading states
+    vm.hasError = false;
+    toggleLoadingIndicator();
+
+    // if we have search parameters, use search.  Otherwise, just read all
+    // employees.
+    var request = angular.isDefined(parameters) && !isEmpty(parameters) ?
```
jniles: In all other modules, these routes were merged into a single route to simplify the logic.  This was proposed in https://github.com/IMA-WorldHealth/bhima-2.X/issues/1665 and implemented in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1674.

It shouldn't be too hard to update this PR before merge.

client/src/modules/employees/registry/registry.js
```diff
@@ -0,0 +1,245 @@
...
+    load(vm.filters);
+  }
+
+  // toggles the loading indicator on or off
+  function toggleLoadingIndicator() {
+    vm.loading = !vm.loading;
+  }
+
+  // employee patient card
+  function employeeCard(uuid) {
```
jniles: Since the employee card is just a patient card, why not call this `openPatientCard()`?

client/src/modules/employees/registry/search.modal.html
```diff
@@ -0,0 +1,184 @@
+<form
+  name="ModalForm"
+  ng-submit="ModalCtrl.submit(ModalForm)"
+  data-modal="employee-search"
+  bh-form-defaults
+  novalidate>
+
+  <div class="modal-header" style="box-shadow : 0px 2px 4px -4px black; z-index:1000">
```
jniles: Out of curiosity, where did this style come from?

client/src/modules/employees/registry/search.modal.html
```diff
@@ -0,0 +1,184 @@
+<form
+  name="ModalForm"
+  ng-submit="ModalCtrl.submit(ModalForm)"
+  data-modal="employee-search"
+  bh-form-defaults
```
jniles: `bhFormDefaults` is deprecated.  Please do not use it in new modules.

Here is the tracking issue: https://github.com/IMA-WorldHealth/bhima-2.X/issues/1447

client/src/modules/employees/registry/search.modal.html
```diff
@@ -0,0 +1,184 @@
...
+        date-to="ModalCtrl.params.dateEmbaucheTo"
+        mode="clean">
+      </bh-date-interval>
+    </fieldset>
+
+    <br />
+
+    <fieldset>
+      <legend>
+        {{ "FORM.LABELS.EMPLOYEE_DETAILS" | translate }}
```
jniles: Please use the translate directive for all of these keys.

client/src/modules/employees/registry/search.modal.html
```diff
@@ -0,0 +1,184 @@
...
+      </legend>
+
+      <div class="form-group" ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.name.$invalid }">
+        <label class="control-label">
+          {{ 'FORM.LABELS.NAME' | translate }}
+        </label>
+        <span style="display:inline-block;" class="pull-right">
+          <a href ng-click="ModalCtrl.clear('display_name')" tabindex="-1">
```
jniles: We have a component that performs this action.  It was described and implemented in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1298.  Using it will clean up this code a lot.

client/src/modules/employees/registry/search.modal.html
```diff
@@ -0,0 +1,184 @@
...
+
+        <label>
+          <input type="radio" name="sexe" id="male" value="M" ng-model="ModalCtrl.params.sexe">
+          <span translate>FORM.LABELS.MALE</span>
+        </label>
+      </div>
+
+      <div class="radio">
+        <label>
+          <input type="radio" name="sexe" id="female" value="F" ng-model="ModalCtrl.params.sexe">
```
jniles: Can you make this module `sex`?  It will line up with the `patient` table much better and keep everything in English.

client/src/modules/employees/registry/search.modal.html
```diff
@@ -0,0 +1,184 @@
...
+          <span translate>FORM.LABELS.FEMALE</span>
+        </label>
+      </div>
+
+      <div class="form-group">
+        <label class="control-label" translate>
+          FORM.LABELS.GRADE
+        </label>
+
+        <!-- @todo make this a component - it is used everywhere. Calls back method
```
jniles: This was tackled in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1298.

client/src/modules/employees/registry/search.modal.js
```diff
@@ -0,0 +1,99 @@
+angular.module('bhima.controllers')
+.controller('EmployeeRegistryModalController', EmployeeRegistryModalController);
+
+EmployeeRegistryModalController.$inject = [
+  '$uibModalInstance', 'params', 'FunctionService', 'GradeService',
+  'bhConstants', 'moment', 'ServiceService'
+];
+
+/**
+ * @class PatientRegistryModalController
```
jniles: These comments should probably be updated or removed.

client/src/modules/employees/templates/action.cell.html
```diff
@@ -0,0 +1,31 @@
...
+      </a>
+    </li>
+    <li>
+      <a data-method="card" href>
+        <span class="fa fa-id-card-o"></span> <span translate>PATIENT_REGISTRY.CARD</span>
+      </a>
+    </li>
+    <li class="divider"></li>
+    <li>
+      <a data-method="card" href>
```
jniles: These methods should refer to the action being performed.

server/controllers/finance/debtors/index.js
```diff
@@ -36,8 +36,8 @@ function list(req, res, next) {
   const sql = `
     SELECT BUID(d.uuid) AS uuid, BUID(d.group_uuid) AS group_uuid,
       d.text, map.text as hr_entity
-    FROM debtor d
-    JOIN entity_map map ON map.uuid = d.uuid;
+    FROM debtor d
```
jniles: Out of curiosity, why did this change to a LEFT JOIN?

server/controllers/medical/patients/index.js
```diff
@@ -16,6 +16,11 @@
  * @requires lib/node-uuid
  * @requires lib/errors/BadRequest
  * @requires lib/errors/NotFound
+ * @requires lib/barcode
+ * @requires lib/filter
```
jniles, line 20: :+1: Nice updates to comments.

server/models/schema.sql
```diff
@@ -612,24 +612,21 @@ CREATE TABLE `employee` (
...
-  `debtor_uuid`   BINARY(16) DEFAULT NULL,
   `locked`        TINYINT(1) DEFAULT NULL,
+  `patient_uuid`  BINARY(16) DEFAULT NULL,
   PRIMARY KEY (`id`),
   UNIQUE KEY `employee_1` (`code`),
   KEY `fonction_id` (`fonction_id`),
   KEY `service_id` (`service_id`),
-  KEY `location_id` (`location_id`),
   KEY `creditor_uuid` (`creditor_uuid`),
-  KEY `debtor_uuid` (`debtor_uuid`),
```
jniles, line 625: Since most financial reports use the `debtor_uuid`, it makes sense to keep it in the table, I think.  That's why https://github.com/IMA-WorldHealth/bhima-2.X/issues/1631 only asked that that `patient_uuid` be added.

server/controllers/payroll/employees/index.js
```diff
@@ -387,3 +393,65 @@ function search(req, res, next) {
...
+ * matching parameters provided in the options parameter.
+ *
+ * @param {Object} options - a JSON of query parameters
+ * @returns {Promise} - the result of the promise query on the database.
+ */
+function find(options) {
+  // ensure epected options are parsed appropriately as binary
+  db.convert(options, ['grade_id', 'creditor_uuid', 'patient_uuid']);
+
+  const filters = new FilterParser(options, { tableAlias : 'e' });
```
jniles: For best practice, set [autoParseStatements](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/filter.js#L56) to `false`.  This option was added later to FilterParser (https://github.com/IMA-WorldHealth/bhima-2.X/commit/be9ebdcd68c22e0df178837d13baefd400b163d4) to ensure that the library didn't do anything the programmer had not explicitly requested.  It also helps the code read better.

test/end-to-end/employees/registration.page.js
```diff
@@ -0,0 +1,170 @@
...
+        FU.validation.ok('EmployeeCtrl.employee.phone');
+        FU.validation.ok('EmployeeCtrl.employee.email');
+        FU.validation.ok('EmployeeCtrl.employee.bank');
+        FU.validation.ok('EmployeeCtrl.employee.bank_account');
+        FU.validation.ok('EmployeeCtrl.employee.service_id');
+        FU.validation.ok('EmployeeCtrl.employee.fonction_id');
+        FU.validation.ok('EmployeeCtrl.employee.adresse');
+    }
+
+    page.createEmployee = createEmployee;
```
jniles: This file has 22 lines of `page.X = X`.  In ES6, you can use a [class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) to make the code read better.  Instead of this:
```js
function SomePage() {
  const page = this;

  setSomeProperty(value) {
    FU.input('$ctrl.key', value);
  }

  page.setSomeProperty = setSomeProperty;
}
```
... your code can be written as:
```js
class SomePage {
  setSomeProperty(value) {
    FU.input('$ctrl.key', value);
  }
}
```

In this case, the end behavior will be the same.

test/end-to-end/employees/registry.spec.js
```diff
@@ -0,0 +1,88 @@
...
+      name : 'Dedrick',
+      oneFilter : 1,
+      twoFilters : 2,
+      threeFilters : 3,
+      fourFilters : 4,
+  };
+
+
+
+  before(() => {return helpers.navigate(path)});
```
jniles: Just FYI:
```js
before(() => { return helpers.navigate(path)});
```
is the same as
```js
before(() => helpers.navigate(path));
```

jniles (changes requested):
@DedrickEnc this represents an impressive amount of work.  Well done!

There are a number of parts of your code that are out of date.  I've commented on them to bring them up to date.  In general, here are the broad categories:
 1. Use the `translate` directive instead of the `translate` filter.
 2. Use the new `FilterService` instead of `DepricatedFilterService`.
 3. Use `bhClear` instead of `<span style="display: 'inline-block' " ...></span>` everywhere.
 4. Use ES5 on the client.

We've been using the `translate` directive for the last four months, you should change your code to use it here before this comes into master.  Also, using ES6 on the client will break the `gulp` build and prevent us from upgrading Vanga.  These should be fixed before this is merged into `master` so that we don't break our client's installation.

In the future, I think it would have been best to do one PR for creating patients when creating an employee, and a second PR for the employee registry.  It would be easier to rebase to master (since less code on each branch) and easier to review (since less code to read).

client/src/modules/employees/registration/employees.html
DedrickEnc: From the hospital perspective, a employee is a creditor so we need to know what will be his group of creditor because every creditor must have a group.

client/src/modules/employees/registration/employees.html
DedrickEnc: A service is not required for an employee in BHIMA, but in reality every employee belongs to a service. (for example at IMCK cleaners and guards belong to SERVICES GENERAUX ).

client/src/modules/employees/registration/employees.html
DedrickEnc: Bank is the name of the bank which is hosting the employee bank account, the bank account his a string of characters representing the employee bank account. There are not required

client/src/modules/employees/registration/employees.html
DedrickEnc: Honestly, I don't know. This should be discussed to help BHIMA find the right thing to do.

server/models/schema.sql
DedrickEnc: For the hospital, we have creditors and debtors. 
- Creditors are providers of services or drugs in this group we have (employees and suppliers)
- Debtors consumed services and drugs in this group we have (patients and other clients like visitors of guest house for the case of IMCK).
If you want we will discuss it, but I think it makes more sense to remove the ```debtor_uuid``` reference from the ```employee``` table. As the **employee** must have a correspondent **patient** that why there is a ```patient_uuid``` reference in the employee table.

client/src/modules/employees/registry/search.modal.html
DedrickEnc: Patient registry I think

server/controllers/finance/debtors/index.js
DedrickEnc: Some debtors have not an entity_map so ```LEFT JOIN``` gave me every row I had also this problem with creditor and I did the same. I saw the triggers every thing seems well.

client/src/modules/employees/registry/search.modal.html
DedrickEnc: Thanks, there is an other directive to use in place?

DedrickEnc: Ok, I see

jniles:
No, unfortunately.  In the linked issue, you will see that the Chrome
Browser actively prevents this type of directive from being created.  It
was a nice idea, but not really feasible at the moment without some serious
hacks.

server/controllers/finance/debtors/index.js
jniles: If some debtors are missing an entity map, than that is a bug.  A trigger should create an an entity_map record anytime a debtor-type entity is created.

Do you have a reproducible test case?

server/models/schema.sql
jniles: Okay - the `patient_uuid` will still link to a `debtor`, so employees are still both debtors/creditors in BHIMA after this change.  I was just thinking that getting the `debtor_uuid` for an employee will now require that we do an extra JOIN on `patient`.

client/src/modules/employees/registry/search.modal.html
jniles: Normally when CSS is used in multiple locations, it is good practice to make it into a CSS `class`.  That way this element, and future elements, can benefit from sharing the same code base, the same bug fixes, and the whole application becomes smaller.

This is a non-blocking comment.

client/src/modules/employees/registration/employees.html
jniles: Okay, that makes sense.

We may need to separate the idea of services at Patient Invoice from these services.  We wouldn't want Patient Invoices to go to the SERVICES GENERAUX .. that data would be messy to clean up.

DedrickEnc:
@jniles
Changes are landed

client/src/modules/employees/employee.service.js
```diff
@@ -1,81 +1,93 @@
 angular.module('bhima.services')
   .service('EmployeeService', EmployeeService);
 
-EmployeeService.$inject = ['$http', 'util'];
+EmployeeService.$inject = ['FilterService', '$uibModal', 'PrototypeApiService', 'appcache'];
```
jniles, line 4: :+1:

client/src/modules/employees/employee.service.js
```diff
@@ -1,81 +1,93 @@
...
-   * .then(function (employees){
-   *   your code here
-   *  });
-   */
-  function read(id) {
-    var url = baseUrl.concat(id || '');
-    return $http.get(url)
-      .then(util.unwrapHttpResponse);
-  }
+  employeeFilters.registerDefaultFilters([{ key : 'limit', label : 'FORM.LABELS.LIMIT' }]);
```
jniles, line 25: In the future, it might be useful to have a default filter on `period` as well.

The feedback from Vanga is that they preferred the period selection over the date selection because it includes useful links like "yesterday"  and "last year".

client/src/modules/employees/registration/employees.html
```diff
@@ -0,0 +1,513 @@
...
+    </ol>
+  </div>
+</div>
+
+<div class="flex-content">
+  <div class="container-fluid">
+    <form 
+      name="EmployeeRegistrationForm"
+      bh-submit="EmployeeCtrl.submit(EmployeeRegistrationForm)"
+      bh-form-defaults
```
jniles: At some point, we should remove this directive.  I'll make an issue about it for later.

**EDIT**: Issue added: #1739.

client/src/modules/employees/registration/employees.html
```diff
@@ -0,0 +1,513 @@
...
+                      </div>
+                    </div>
+                  </div>
+
+                  <!-- Employee grade-->
+                  <div class="form-group"
+                    ng-class="{'has-error' : DetailForm.grade.$invalid && EmployeeRegistrationForm.$submitted}">
+
+                    <!-- Warning for incomplete data - the page will not be able to continue -->
+                    <uib-alert class="text-danger" type="danger" ng-if="EmployeeCtrl.grades.length === 0">
```
jniles, line 119: Could we add a link to where to add a grade here?  It might help users get set up faster.

client/src/modules/employees/registration/employees.html
```diff
@@ -0,0 +1,513 @@
...
+                    <label class="col-md-3 control-label" translate>FORM.LABELS.GRADE</label>
+                    <div class="col-md-9">
+                      <ui-select
+                        name="grade"
+                        ng-model="EmployeeCtrl.employee.grade_id"
+                        required>
+                        <ui-select-match placeholder="{{ 'FORM.SELECT.GRADE' | translate }}">
+                          <span>{{$select.selected.format}}</span>
+                        </ui-select-match>
+                        <ui-select-choices ui-select-focus-patch repeat="g.uuid as g in EmployeeCtrl.grades | filter:{format : $select.search} | orderBy:'text'">
```
jniles: Are you sure you want to `orderBy: 'text'` here, considering that you are only showing the `format` to the user?

client/src/modules/employees/registration/employees.html
```diff
@@ -0,0 +1,513 @@
...
+                    </div>
+                  </div>
+
+                  <!-- Employee daily salary -->
+                  <div
+                    class="form-group"
+                    ng-class="{'has-error' : DetailForm.daily_salary.$invalid && EmployeeRegistrationForm.$submitted}">
+
+                    <label class="col-md-3 control-label" translate>FORM.LABELS.DAILY_SALARY</label>
+                    <div class="col-md-9">
```
jniles, line 440: It would probably be good to show the currency on this input, so that the users do not get confused.

client/src/modules/employees/registration/employees.js
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  // Expose employee to the scope
+  vm.employee = {};
+
+  // Expose methods to the scope
+  vm.submit = submit;
+
+  // Loading Grades
+  Grades.read(null, { detailed : 1 }).then((data) => {
```
jniles: This ES6 still needs to be cleaned up.

client/src/modules/employees/registry/registry.js
```diff
@@ -0,0 +1,218 @@
+angular.module('bhima.controllers')
+  .controller('EmployeeRegistryController', EmployeeRegistryController);
+
+EmployeeRegistryController.$inject = [
+  '$state', 'EmployeeService', 'NotifyService', 'AppCache',
+  'util', 'ReceiptModal', 'uiGridConstants', '$translate',
```
jniles: I don't think you used $translate.

client/src/modules/employees/registry/registry.js
```diff
@@ -0,0 +1,218 @@
...
+    return load(Employees.filters.formatHTTP(true));
+  }
+
+  // toggles the loading indicator on or off
+  function toggleLoadingIndicator() {
+    vm.loading = !vm.loading;
+  }
+
+  // employee patient card
+  function openPatientCard(uuid) {
```
jniles, line 206: 👍

client/src/modules/employees/registry/search.modal.js
```diff
@@ -0,0 +1,83 @@
...
+ * @description
+ * This controller is responsible for setting up the filters for the employee
+ * search functionality on the employee registry page.
+ */
+function EmployeeRegistryModalController(ModalInstance, Functions, Grades, bhConstants, moment, Services, Store, util, filters) {
+  var vm = this;
+  var changes = new Store({identifier : 'key'});
+
+  vm.filters = filters;
+  vm.serchQueries = {};
```
jniles: This variable isn't ever used.

server/controllers/payroll/employees/index.js
```diff
@@ -38,33 +38,12 @@ exports.search = search;
...
-  .then(rows => {
-    res.status(200).json(rows);
-  })
-  .catch(next)
-  .done();
+  find(req.query)
+    .then((rows) => {
+      res.status(200).json(rows);
+    })
+    .catch(function (err){console.log(err);})
```
jniles: This should use `next` to handle the error.

jniles (commented):
This is much better - the classes for page objects make much cleaner code, and the new filters are excellent. 👍   Good work.

The app still doesn't build with gulp-uglify to compress the files due to ES6 on the client.  I believe I found one place where an arrow function remains, there may be others.  You can build a minified version locally to test whether your code will run in production by changing the `isProduction` variable in our [gulpfile](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/gulpfile.js#L30)
```js
// old
const isProduction = (process.env.NODE_ENV === 'production');

// new
const isProduction = true;
```
This will help you catch any other errors here and in future pull requests.

There's also another spot where you could clear up your imports (`$translate`) since you don't use them.

After that, the code is good enough to be merged.

client/src/modules/employees/registration/employees.html
DedrickEnc: Yes but I think not for this PR I think. An other PR will follow for editing and other functionalities and I can implement it over there.
What do you think?

jniles: Sure - no problem.  Just a comment for the future.

DedrickEnc:
@jniles
I have tried the build on production, every thing pass.

DedrickEnc:
@jniles
I have tried the build on production, every thing pass.

DedrickEnc:
All Es6 was removed from the client

jniles (approved):

jniles:
@DedrickEnc, this PR passes code review.  All the comments have been responded to.  Well done :+1:

Since this is a large feature (2K+ line changes!), I propose that you present the module to the team on a projector Monday.  That way you can walk us through it, and the team can give feedback as we did with the report module last Thursday.

I'll defer to bors to merge it.

PR #1695, opened 2017-05-29
Using Menus instead all Util to Menu

### LINKED ISSUE #1693: (bug) general ledger and balance report are corrupted
[ ] General ledger :
	- Bug: The amount of the balance doesn't correspond to the sold of the account
	- Fix: the french translation "Grande Livre" to "Grand Livre"

[ ] Report / Balance : 
	- Bug: The report is crashed

[ ]  Report / Balance agee des creances :
	- Bug: Error in the console

lomamech:
Using Menus instead all Util to Menu

closes #1693

### DIFF
--- a/client/src/i18n/en/posting_journal.json
+++ b/client/src/i18n/en/posting_journal.json
@@ -7,7 +7,6 @@
 "CREDIT":"Credit",
 "ROW":"Row",
 "ROWS":"Row(s)",
-"MENUS":"Menus",
 "CREDIT_LETTER":"C",
 "DEBIT_LETTER":"D",
 "DEBIT":"Debit",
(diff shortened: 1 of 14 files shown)


client/src/i18n/en/posting_journal.json
```diff
@@ -7,7 +7,7 @@
 "CREDIT":"Credit",
 "ROW":"Row",
 "ROWS":"Row(s)",
-"TOOLS":"Tools",
+"MENUS":"Menus",
```
jniles: Since you have many keys "MENUS", I propose to simple make one key in `form.json` that says "MENU".

Also, in English, "MENUS" should be singular - "MENU".  Since we only have one menu, with multiple items.

jniles:
LGTM.  As soon as test pass, this will be merged :+1:

PR #1698, opened 2017-05-30
Prevent $state.current to become empty afer refresh

### LINKED ISSUE #1345: (bug) User Management does not load user on page refresh
When you navigate to the user management page and click "Edit" on a user, the state+url changes to `/users/:id/edit` and a modal appears with all the user's data.  However, if you refresh from this page, the form loads empty, but still with the "update" label.  Submitting results in form validation rejecting the changes.

**Expected Result** 
The user's information should fill the form.

**Actual Result**
A blank form is loaded.

**Steps to reproduce**
 1. Navigate to `/users`
 2. Click `Edit` on any user.
 3. Wait for the modal to appear.
 4. Refresh the page using F5 or the refresh button on the browser.
 5. Observe an empty form appears.

lomamech:
Prevent $state.current to become empty afer refresh
- User Management
- Account Edit
- Billing Sercice Updare
- Inventory actions modal

closes #1345

### DIFF
--- a/client/src/modules/inventory/inventory.routes.js
+++ b/client/src/modules/inventory/inventory.routes.js
@@ -30,6 +30,9 @@ angular.module('bhima.routes')
       })
       .state('inventory.create', {
         url : '/create',
+        params : {
+          creating : { value : true }
+        },        
         onEnter :['$state', 'ModalService', onEnterFactory('create')],
         onExit : ['$uibModalStack', closeModal]
       })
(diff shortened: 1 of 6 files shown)


client/src/modules/inventory/inventory.routes.js
```diff
@@ -30,6 +30,9 @@ angular.module('bhima.routes')
       })
       .state('inventory.create', {
         url : '/create',
+        params : {
+          creating : { value : true }
+        },
```
mbayopanda, line 35: What is the need of value ? What do you think about :
```
params : {
  creating : true,
},
```

mbayopanda (commented):
LGTM

client/src/modules/inventory/inventory.routes.js
lomamech: It makes it possible to differentiate the creation of the update, to consider only the present of the $state.params.uuid was not sufficient especially when one refreshed the browser

jniles:
Seems like this got the green light!  Merging...

PR #1700, opened 2017-05-30
feature(grids): Grid save state

sfount:
This commit introduces a new feature to manually save and restore a
grid's state using `ui-grid-save-state`. The feature is prototyped in
the Journal module and adds an entry to the utils dropdown.

State is added as an additional Grid service that can be imported and
used by any module that uses a UI-Grid.

This feature adds two options to the 'Tools' menu on the Journal - these actions allow the user to save the current state of the UI Grid. The state is also saved by showing/ hiding columns as well as grouping the gird. 

Saving the grid state allows the user to save many different options like column width, column order and others. When the page is refreshed the saved grid configuration will be used.

screenshot 2017-05-30 at 13 34 02

Allows grid configuration that can be saved - for example a posting journal interface with moved and resized columns. 

screenshot 2017-05-30 at 13 37 23

**Future Development**
Saving and loading a grids state can be added to other modules by importing the `GridSaveState` service. Once the feature has been improved/ approved it could be applied to all modules with a UI-grid. 

---

This commit updates the journal module, replacing all instances of
custom local storage caching with the GridState save state method. Save
state is simply invoked to capture the current configuration set by each
component. Column configuration and the Grouping toggle were updated to
use save state.


client/src/js/services/grid/GridState.js
```diff
@@ -0,0 +1,57 @@
...
+
+    util.after(gridOptions, 'onRegisterApi', function onRegisterApi(api) {
+      this._gridApi = api;
+
+      this._gridApi.core.on.rowsRendered(null, util.once(function () {
+        this.restoreGridState();
+      }.bind(this)));
+    }.bind(this));
+
+    this.saveGridState = saveGridState.bind(this);
```
jniles, line 28: It would probably be more efficient to attach these to `StateInstance.prototype`.  Is there a reason not to?

client/src/js/services/grid/GridState.js
```diff
@@ -0,0 +1,57 @@
...
+    this.clearGridState = clearGridState.bind(this);
+  }
+
+  function saveGridState(notifyFlag) {
+    var shouldNotify = angular.isDefined(notifyFlag) ? notifyFlag : true;
+
+    if (this._gridApi) {
+      this._cache.gridState = this._gridApi.saveState.save();
+
+      if (shouldNotify) {
```
jniles, line 39: :+1: Cool flag/feature interface

jniles (commented):
Code looks fine to me.

I'm curious about the road map for this feature?  What functionality does this enable us to use, how can we incorporate it into new grids?

Also, a screenshot or two would be awesome. :+1:

client/src/js/services/grid/GridState.js
sfount: I did this initially. With the current structure of `journal.js` the scope for `this` wasn't available - it would be a useful re-factor to look at the Grid utilities and favour this method though.

sfount:
@jniles Updated description.

jniles:
Cheers, thanks!

jniles:
Looks good to me. :+1:

PR #1701, opened 2017-05-30
fix(trialBalance): align exporter/error buttons

### LINKED ISSUE #1680: (bug) Trial Balance UI Errors + Exporter
The Trial Balance "Exporter" button and "View Errors" buttons do not play well together.  See below:
screenshot from 2017-05-25 11 40 10
_Fig 1: View Errors and Exporter Buttons shift modal size_

This should be corrected so that the two are on the same line.

jniles:
This commit fixes the alignment bugs in the exporter and error buttons.
It also makes the exporter button a normal button, instead of a
"success" button.

Closes #1680.

------

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/client/src/modules/journal/modals/trialBalanceMain.body.html
+++ b/client/src/modules/journal/modals/trialBalanceMain.body.html
@@ -1,22 +1,9 @@
-<!-- exportation -->
-<div class="row text-right" style="margin: 5px;">
-  <div class="col-xs-12">
-    <button
-      id="export"
-      type="button"
-      class="btn btn-success"
-      ng-click="TrialBalanceMainBodyCtrl.exportGrid()">
-      <i class="fa fa-file-excel-o"></i> 
-      <span translate>FORM.BUTTONS.EXPORT</span>
-    </button>
-  </div>
-</div>
-
-<div class="row" style="margin: 5px;" ng-if="TrialBalanceMainBodyCtrl.showErrorButton">
-  <div class="col-xs-4">
+<!-- tools for data validation and export -->
+<div class="row " style="margin: 5px;">
+  <div class="col-xs-6">
     <button
       type="button"
-      class="btn"
+      class="btn btn-default"
       ng-class="{
         'btn-danger' : TrialBalanceMainBodyCtrl.feedBack.hasError,
         'btn-warning': TrialBalanceMainBodyCtrl.feedBack.hasWarning,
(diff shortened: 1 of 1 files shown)


client/src/modules/journal/modals/trialBalanceMain.body.html
```diff
@@ -1,22 +1,9 @@
...
-      <i class="fa fa-file-excel-o"></i> 
-      <span translate>FORM.BUTTONS.EXPORT</span>
-    </button>
-  </div>
-</div>
-
-<div class="row" style="margin: 5px;" ng-if="TrialBalanceMainBodyCtrl.showErrorButton">
-  <div class="col-xs-4">
+<!-- tools for data validation and export -->
+<div class="row " style="margin: 5px;">
```
sfount, line 2: At some point we should go through these hard coded margins and see if there is a pattern that we can resolve in the less styles.

sfount (approved):
Changes LGTM.

client/src/modules/journal/modals/trialBalanceMain.body.html
```diff
@@ -28,6 +15,17 @@
       <span translate>FORM.BUTTONS.VIEW_ERRORS</span>
     </button>
   </div>
+
+  <div class="col-xs-6 text-right">
+    <button
+      id="export"
+      type="button"
```
sfount, line 22: I wanted to comment on this type being redundant on a button but: 
> Tip: Always specify the type attribute for the <button> element. Different browsers may use different default types for the <button> element.
[source](https://www.w3schools.com/tags/att_button_type.asp)

jniles: Good reference!

I've also run into the case where `<button>Cancel</button>` is interpreted as `<button type="submit">Cancel</button>` and submits the form instead of canceling it (if in a `<form>` element).  I'm not sure how often this occurs, but it is mighty annoying and confusing.

client/src/modules/journal/modals/trialBalanceMain.body.html
jniles: Sounds good to me.  ~~I'll put it in an issue.~~  See https://github.com/IMA-WorldHealth/bhima-2.X/issues/1711.

PR #1710, opened 2017-05-31
refactor(journal): Remove unused legacy imports

sfount:
This commit removes the `FilterService` and `$scope` imports on the
client side Journal module. `FilterService` was used during
development and is not used by the module. `$scope` is not currently required
given the one ui-grid used by the page.


jniles (approved):
LGTM.  Good clean up!

jniles:
Hmmm... I get the feeling Travis CI is having hiccups.  Either that or `master` is broken.  I'll investigate later.

Issue #1709, opened 2017-05-30
chore: add new maintainer to package.json


jniles:
We recently gained a new developer: @jeremielodi!  Welcome!

I propose that @jeremielodi implement a Pull Request (PR) that adds his name to the [package.json](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/package.json#L38) file under the key "maintainers".  This PR will be reviewed and promptly merged.

@jeremielodi how does this sound?  Can you tackle it?

jeremielodi:
I am ready to do it

jniles:
Closed in #1726.

PR #1726, opened 2017-05-31
adding my name in the mainteners list from git

jeremielodi:
This PR will update the file package.js  by adding jeremielodi as a new maintener


jniles:
Closes #1709.

PR #1735, opened 2017-06-01
Fix (Research DebtorGroup)

### LINKED ISSUE #1715: Problem of the link between the list of debtor groups and patients
Problem with the link between the list of lending groups and the patients belonging to this group
imagg

lomamech:
Fix the link between the list of debtor groups and Patient
- Update the filter equals remove the alternative who check if there are a
  Default Uuid Partial Key
- Using db.convert for cash.debtor_uuid

closes #1715

### DIFF
--- a/server/controllers/finance/cash.js
+++ b/server/controllers/finance/cash.js
@@ -135,6 +135,8 @@ function read(req, res, next) {
  * @description list all payment made
  */
 function listPayment(options) {
+  // ensure epected options are parsed appropriately as binary
+  db.convert(options, ['debtor_uuid']);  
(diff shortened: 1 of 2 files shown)


server/controllers/finance/cash.js
```diff
@@ -135,6 +135,8 @@ function read(req, res, next) {
  * @description list all payment made
  */
 function listPayment(options) {
+  // ensure epected options are parsed appropriately as binary
+  db.convert(options, ['debtor_uuid']);
```
jniles, line 139: I am sorry to say that I introduced this bug :(

Good catch!

jniles (approved):

Issue #1738, opened 2017-06-01
(refactor) Journal module grid state


sfount:
After meeting and discussing the refactor required in https://github.com/IMA-WorldHealth/bhima-2.X/issues/1737 it is proposed that we should remove all implicit state saving as a standard on grid modules. This behaviour is currently implemented in the Journal module so this should be updated. 

Proposed behaviour: 
Grid state should only ever be saved if the user select 'Save grid configuration'

jniles:
@sfount, what is the status of this issue?

jniles:
This was closed some time ago.

PR #1753, opened 2017-06-02
refactor(grids): Remove implicit grid save state

sfount:
This commit removes the implicit save state from the Journal grid,
called when the columns are changed and groups are updated. This
improves the UX by ensuring the grid settings will only ever be cached
if the user has selected to save the grid state.

It also updates the grouping utility to check if the grid is in a
grouped state. This should be optimised in the future.

Contributes to https://github.com/IMA-WorldHealth/bhima-2.X/issues/1738

### DIFF
--- a/client/src/modules/journal/journal.html
+++ b/client/src/modules/journal/journal.html
@@ -65,7 +65,7 @@
           </button>
         </div>
         <div class="toolbar-item">
-          <button ng-class="{ 'btn-info' : JournalCtrl.grouped }" class="btn btn-default" data-method="grouping" ng-click="JournalCtrl.toggleTransactionGroup()">
+          <button ng-class="{ 'btn-info' : JournalCtrl.gridGroupedState() }" class="btn btn-default" data-method="grouping" ng-click="JournalCtrl.toggleTransactionGroup()">
             <i class="fa fa-object-group"></i> <span class="hidden-xs" translate>FORM.LABELS.TRANSACTION</span>
           </button>
         </div>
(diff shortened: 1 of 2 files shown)


client/src/modules/journal/journal.html
```diff
@@ -65,7 +65,7 @@
           </button>
         </div>
         <div class="toolbar-item">
-          <button ng-class="{ 'btn-info' : JournalCtrl.grouped }" class="btn btn-default" data-method="grouping" ng-click="JournalCtrl.toggleTransactionGroup()">
+          <button ng-class="{ 'btn-info' : JournalCtrl.gridGroupedState() }" class="btn btn-default" data-method="grouping" ng-click="JournalCtrl.toggleTransactionGroup()">
```
jniles, line 68: :+1:

client/src/modules/journal/journal.js
```diff
@@ -475,25 +473,24 @@ function JournalController(Journal, Sorting, Grouping,
...
-      vm.grouping.changeGrouping('trans_id');
-
-      // save grids state to keep track of this change
-      state.saveGridState(false);
+    var transactionColumnKey = 'trans_id';
 
-      // @FIXME temporary cahced variable to track the grouped state - this should be encapsulated in a component
-      vm.grouped = cache.grouped = true;
+    // column grouping is permenantly saved if the user saves the current grid state
+    if (vm.grouping.getCurrentGroupingColumn()) {
```
jniles, line 479: Nice cleanup!

jniles (approved):
LGTM

PR #1341, opened 2017-03-12
(refactor) Report Client Infrastructure

sfount:
This pull request refactors how reports are generated on the client and how the report saving/ archive feature works. 

Users have found it very inconvenient for every report to be saved by default - there is also a significant wait time for generating a PDF on every report attempt. https://github.com/IMA-WorldHealth/bhima-2.X/issues/1260

This has been addressed by: 
* Setting the flag to save the report to false unless the user explicitly requests 'Save As' 
* 'Preview' the form as HTML to increase response time, this removes the requirement to compile every report as PDF 
* Allow the user to save and view the archive page independently of report creation

This pull request will also help form the standards for report UX requirements in https://github.com/IMA-WorldHealth/bhima-2.X/issues/1278

**Account Transactions Report Page**
screenshot 2017-03-09 at 19 09 45

**Report Preview - 1100 Test Capital Account**
screenshot 2017-03-09 at 19 10 25

**Explicitly Save Report**
screenshot 2017-03-09 at 19 11 12

**View Report Archive**
screenshot 2017-03-09 at 19 17 37

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/1260

PR #1765, opened 2017-06-05
refactor(report) Account report

mbayopanda:
This PR is about the refactoring of bhima 2.X reports according the standard of preview and archive.

This PR concerns the account report by update it according the new standard.

This PR is an extension of https://github.com/IMA-WorldHealth/bhima-2.X/pull/1341


mbayopanda:
@sfount @jniles  could you review this PR please

jniles:
@mbayopanda, I'll take a look at this.

client/src/js/components/bhReportPreview.js
```diff
@@ -0,0 +1,20 @@
+angular.module('bhima.components')
+.component('bhReportPreview', {
+  bindings : {
+    sourceDocument : '<',
+    onClearCallback : '&',
```
jniles: It would be useful to put a comment here saying that these methods are called in the template - I was confused for a second.

client/src/js/components/bhReportSource.js
```diff
@@ -0,0 +1,47 @@
+angular.module('bhima.components')
+.component('bhReportSource', {
+  bindings : {
+    onSelectCallback : '&?',
```
jniles: I don't think this should be marked as optional - the component will likely error if you don't pass it in.

client/src/js/components/bhReportSource.js
```diff
@@ -0,0 +1,47 @@
...
+  controller : bhReportSourceController,
+});
+
+bhReportSourceController.$inject = [];
+
+function bhReportSourceController() {
+  var ctrl = this;
+  var DEFAULT_SOURCE = 1;
+
+  ctrl.label = ctrl.label || 'FORM.SELECT.SOURCE';
```
jniles: Initialization code like this should go in the `ctrl.$onInit` method.  From the [AngularJS documentation](https://docs.angularjs.org/guide/component):
> $onInit() - Called on each controller after all the controllers on an element have been constructed and had their bindings initialized (and before the pre & post linking functions for the directives on this element). This is a good place to put initialization code for your controller.

By my reading of this, the `ctrl.label` assignment should happen in the `$onInit` function so that you are sure the bindings are enabled.

client/src/js/components/bhReportSource.js
```diff
@@ -0,0 +1,47 @@
...
+  ctrl.name = ctrl.name || 'source';
+
+  /* @const */
+  ctrl.sources = [
```
jniles: Can/should this support a combination of the two?

client/src/modules/reports/baseReports.service.js
```diff
@@ -63,16 +45,14 @@ function BaseReportService($http, Modal, util) {
...
    */
-  function requestPDF(url, report, reportOptions) {
-    var pdfParams = {
+  function saveReport(url, report, reportOptions) {
+    var params = {
       // @TODO This should be known by the server
       reportId : report.id,
-      saveReport : true,
-      renderer : 'pdf'
+      saveReport : '1',
```
jniles, line 53: Does `true` not work?  The string `'1'` seems really messy.

client/src/modules/reports/deprecated-modals/accounts_chart.modal.html
```diff
@@ -5,15 +5,15 @@
   </ol>
 </div>
 
-<form name="ConfigForm" bh-submit="ReportConfigCtrl.generate()" novalidate>
+<form name="ConfigForm" bh-submit="ReportConfigCtrl.generate()" bh-form-defaults novalidate>
```
jniles, line 8: `bhFormDefaults` has been deprecated in https://github.com/IMA-WorldHealth/bhima-2.X/issues/1447 and completely removed in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1755.  It should be removed from these modules.

client/src/modules/reports/deprecated-modals/accounts_chart.modal.html
```diff
@@ -5,15 +5,15 @@
...
 <div class="modal-body">
   <div class="form-group"
     ng-class="{ 'has-error' : ConfigForm.$submitted && ConfigForm.label.$invalid }">
     <label class="control-label">{{ "FORM.LABELS.LABEL" | translate }}</label>
     <input class="form-control" autocomplete="off" id="label" name="label" ng-model="ReportConfigCtrl.label" required />
 
     <div class="help-block" ng-messages="ConfigForm.label.$error" ng-show="ConfigForm.$submitted">
-      <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      <div ng-messages-include="partials/templates/messages.tmpl.html"></div>
```
jniles, line 16: I don't think this route exists anymore - "partials" was renamed to "modules".

client/src/modules/reports/deprecated-modals/income_expense.modal.html
```diff
@@ -5,15 +5,15 @@
   </ol>
 </div>
 
-<form name="ConfigForm" bh-submit="ReportConfigCtrl.generate(ConfigForm)" novalidate>
+<form name="ConfigForm" bh-submit="ReportConfigCtrl.generate(ConfigForm)" bh-form-defaults novalidate>
 <div class="modal-body">
   <div class="form-group"
     ng-class="{ 'has-error' : ConfigForm.$submitted && ConfigForm.label.$invalid }">
     <label class="control-label">{{ "FORM.LABELS.LABEL" | translate }}</label>
```
jniles, line 12: These labels should be translated with the `translate` directive.

client/src/modules/reports/generate/report_accounts/report_accounts.config.js
```diff
@@ -0,0 +1,75 @@
+angular.module('bhima.controllers')
+  .controller('report_accountsController', ReportAccountsConfigController);
```
jniles: The naming convention here is vary confusing... The controller is `report_accounts`, but the AppCache is `reports_account` and the route is `report/finance/account`.

I propose that we change all these to `account_report` - no plurals.  That way, we won't have to remember which word the `s` goes on.  All variables should be updated accordingly.

client/src/modules/reports/generate/report_accounts/report_accounts.html
```diff
@@ -0,0 +1,63 @@
...
+
+  <div class="row" style="margin-top : 10px">
+    <div class="col-md-6">
+      <div class="panel panel-default">
+        <div class="panel-heading">
+          <span translate>REPORT.UTIL.OPTIONS</span>
+        </div>
+
+        <div class="panel-body">
+          <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" bh-form-defaults novalidate>
```
jniles: `bhFormDefaults` no longer exists.  You can remove it here.

client/src/modules/reports/reports.html
```diff
@@ -3,19 +3,23 @@
 
     <ol class="headercrumb">
       <li class="static" translate>TREE.REPORTS</li>
-      <li class="title" translate>{{ReportCtrl.report.title_key}}</li>
+      <li ng-hide="ReportCtrl.isArchive()" class="title" translate>{{ ReportCtrl.report.title_key }}</li>
+      <li ng-show="ReportCtrl.isArchive()">
+        <a data-method="report-config" ng-href="#!/reports/{{ReportCtrl.report.report_key}}" translate>{{ ReportCtrl.report.title_key }}</a>
```
jniles: Can this use `ui-sref` instead of `ng-href`, by any chance?  It would be more in line with the rest of our modules.

client/src/modules/reports/reports.js
```diff
@@ -2,76 +2,17 @@ angular.module('bhima.controllers')
   .controller('ReportsController', ReportsController);
 
 ReportsController.$inject = [
-  '$state', 'BaseReportService', 'NotifyService'
+  '$state', 'NotifyService', 'reportData',
 ];
 
-function ReportsController($state, SavedReports, Notify) {
+function ReportsController($state, Notify, reportData) {
```
jniles: Do you use `Notify` here?  If not, please remove it.

client/src/modules/reports/reports.routes.js
```diff
@@ -1,11 +1,40 @@
 angular.module('bhima.routes')
   .config(['$stateProvider', function ($stateProvider) {
+    // a list of all supported reported and their respective keys, this allows
+    // the ui-view to be populated with the correct report configuration form
+    /* @const */
+    var SUPPORTED_REPORTS = [
+      'report_accounts',
```
jniles: In the future, we should figure out some way to make `SUPPORTED_REPORTS` data-driven.

client/src/modules/templates/modals/report.save.html
```diff
@@ -0,0 +1,37 @@
+<div class="modal-header">
+  <ol class="headercrumb">
+    <li class="static" translate>{{SaveCtrl.report.title_key}}</li>
+    <li class="title" translate>REPORT.UTIL.SAVE_AS</li>
+  </ol>
+</div>
+
+<form name="SaveForm" bh-submit="SaveCtrl.submit(SaveForm)" bh-form-defaults autocomplete="off" novalidate>
```
jniles: No need to `bhFormDefaults`.

jniles: I'm also not sure the autocomplete="off" will work - I think the `<input>`'s autocomplete="off" is sufficient.

server/controllers/finance/reports/clientsReport/index.js
```diff
@@ -26,35 +25,25 @@ function document(req, res, next) {
...
-    session,
-    {
-      dateFrom : new Date(params.dateFrom),
-      dateTo : new Date(params.dateTo),
-      detailPrevious : params.detailPrevious,
-      ignoredClients : params.ignoredClients,
-      enterprise : req.session.enterprise,
-    });
-
+  _.defaults(session, { dateFrom : new Date(params.dateFrom), dateTo : new Date(params.dateTo), detailPrevious : params.detailPrevious, ignoredClients : params.ignoredClients, enterprise : req.session.enterprise });
```
jniles, line 28: Just so you know, you can use ECMAScript2016 destructuring to assign variables inline from an object.  If you used it above, it would look like this:
```js
// destructure the parameters we care about
const { dateFrom, dateTo, detailPrevious, ignoredClients } = params;

// add them to the defaults
_.defaults(session, { dateFrom: new Date(dateFrom), dateTo: new Date(dateTo), detailPrevious, ignoredClients, enterprise = req.session.enterprise });
```
Just an FYI - I don't necessarily think it is appropriate here. :)

jniles: [Relevant MDN Documentation](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

server/controllers/finance/reports/reportAccounts/report.handlebars
```diff
@@ -27,20 +27,6 @@
         </tr>
       </thead>
       <tbody>
-
-        {{! Opening Balance Line }}
-        <tr>
```
jniles: :(.  What happened to the opening balance?

test/end-to-end/agedCreditors/agedCreditors.spec.js
```diff
@@ -10,7 +10,7 @@ const GU = require('../shared/GridUtils');
 
 helpers.configure(chai);
 
-describe('Aged Creditors Report', () => {
+describe.skip('Aged Creditors Report', () => {
```
jniles: Probably best to just remove these files.

server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -30,42 +38,27 @@ function document(req, res, next) {
...
+
   try {
     report = new ReportManager(TEMPLATE, req.session, params);
   } catch (e) {
     return next(e);
   }
 
-  const dateFrom = (params.dateFrom) ? new Date(params.dateFrom) : new Date();
-
-  return AccountsExtra.getOpeningBalanceForDate(params.account_id, dateFrom)
```
jniles: This may be the result of a bad merge - the opening balance should be included here.

test/end-to-end/reports/report_accounts/report_accounts.page.js
```diff
@@ -0,0 +1,58 @@
...
+const chai = require('chai');
+const helpers = require('../../shared/helpers');
+
+helpers.configure(chai);
+
+const FU = require('../../shared/FormUtils');
+const ReportPage = require('../page');
+const components = require('../../shared/components');
+
+class AccountReportPage {
```
jniles: Cool :+1:

Considering that this page will look identical for every report, I propose that you refactor this so that the common elements can be shared in a generic `ReportPage`.  For example - saving and printing will be the same for all reports.

jniles (dismissed):
@mbayopanda, this is a large amount of work.  Well done bringing the branch up to date!

I've made a number of comments on syntax and style.  The ones in the `deprecated-modals` folder don't matter too much - that is old code and we will have to update them as we re-implement them.

Here are the changes I believe need to be made:
 1. The directive `bhFormDefaults` is no longer implemented and should be removed.
 2. Removing the Opening Balances is a significant regression - it was a feature request from our clients and should be returned to the Account Report.

Otherwise, this looks great on the code side.  I haven't pulled the code to see how it works yet - after these changes are made, we can take a look at the functionality and just make sure it behaves as expected.

Good work :+1:

client/src/modules/reports/deprecated-modals/accounts_chart.modal.html
mbayopanda: all these files which are in the `deprecated-modals` (old `modals`) are not used, and they must be removed, they are still there because this a first implementation of the new design of reports.

client/src/modules/reports/deprecated-modals/accounts_chart.modal.html
mbayopanda: deprecated code

client/src/modules/reports/deprecated-modals/income_expense.modal.html
mbayopanda: deprecated code and not in use

client/src/modules/reports/generate/report_accounts/report_accounts.config.js
mbayopanda: 👍

client/src/modules/reports/reports.js
mbayopanda: 👍

sfount: :+1:

client/src/modules/reports/deprecated-modals/accounts_chart.modal.html
jniles: No problem - just noting.  Nothing blocking about this in my mind.

mbayopanda:
@jniles @sfount can review this PR because i made some changes that you recommend

sfount:
I will review this to be merged this morning.

sfount:
Note that this PR supersedes and fulfils the requirements of https://github.com/IMA-WorldHealth/bhima-2.X/pull/1341

sfount:
bors, r+

PR #1767, opened 2017-06-06
Using Bcrypt to save password and hash strings

### LINKED ISSUE #1754: best practice: use bcrypt for storing user passwords
We currently use MySQL's `PASSWORD()` function for hashing passwords.  This is lazy security practice - the recommended methods of storing passwords all involve using a [salt](https://en.wikipedia.org/wiki/Salt_(cryptography)) and a [strong algorithm](https://crackstation.net/hashing-security.htm).  The npm ecosystem has a secure [bcrypt](https://www.npmjs.com/package/bcrypt) implementation which includes salt generation.  You can see an example if it's usage in our [emplois.bhi.ma](https://github.com/IMA-WorldHealth/emplois.bhi.ma/blob/master/server/students.js#L44) application.

It would be good practice to introduce a stronger hash function that is cryptographically secure for storing user passwords.  It shouldn't be difficult to implement.

See also this list of [acceptable algorithms](https://paragonie.com/blog/2016/02/how-safely-store-password-in-2016#acceptable-algorithms).

DedrickEnc:
This PR introduces ```bcrypt``` to hash our password and other string.
The mysql function **PASSWORD** was replaced by ```bcrypt``` library

closes #1754 
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


server/controllers/auth.js
```diff
@@ -54,36 +56,53 @@ function login(req, res, next) {
...
+
+      return util.checkString(password, param.user[0].password);
+    })
+    .then((resp) => {
       const connect = param.connect.length;
       const permission = param.permission.length;
-      const user = param.user.length;
 
+      // Will be false if the password does not much
+      if (!resp) {
```
sfount, line 101: This variable has a good comment however it would be useful to give it a meaningful name, that way even if comments get misplaced or removed the code can be read well. 

Something like: 
```js
.then((loginPasswordHasMatched) => { 

  if (loginPasswordHasMatched) // ...
})

File: server/controllers/auth.js [hunk 2]
```diff
@@ -54,36 +56,53 @@ function login(req, res, next) {
...
+        throw new Unauthorized('Bad password, Please Try again', 'FORM.ERRORS.PASSWORD_MATCH');
```
sfount, line 102: In the past the server has always responded with 'Bad username and password combination', now it responds with a specific message about the password. 

This isn't directly an issue however it may be seen as weaker security as now you know if you have found a valid username, you can try and break the password. 

Given our target users and install base I do think that it is a major issue either way. 

@IMA-WorldHealth/bhima-core Do you think we should stick with our current system or change to giving the user specific feedback?

server/controllers/admin/users/index.js
```diff
@@ -145,10 +148,19 @@ function create(req, res, next) {
...
   let sql = `
     INSERT INTO user (username, password, email, display_name) VALUES
-    (?, PASSWORD(?), ?, ?);
+    (?, ?, ?, ?);
   `;
 
-  db.exec(sql, [data.username, data.password, data.email, data.display_name])
+  if (!data.password) {
+    next(new BadRequest(`You cannot create a user without a password.`,
+        `ERRORS.NO_PASSWORD`));
```
sfount: This should not be merged into master with missing translation keys. `ERRORS.NO_PASSWORD`

server/controllers/auth.js
```diff
@@ -54,36 +56,53 @@ function login(req, res, next) {
...
       param.user = rows;
-      return db.exec(sqlPermission, [username, password]);
+      return db.exec(sqlPermission, [username]);
     })
     .then((rows) => {
       param.permission = rows;
+
+      // check if username is correct
+      if (param.user.length === 0) {
+        throw new Unauthorized('The user name provided is bad.');
```
sfount, line 86: Without keys, these messages will display the default unauthorized message to the user, is that the intended behavior?

server/controllers/auth.js
sfount, line 102: This should not be merged into master with missing translation keys. `FORM.ERRORS.PASSWORD_MATCH`

sfount (commented):
This looks like a great start towards using a more secure password hashing system. 

The code looks good to me with useful comments in places, I have made some comments on areas that might make it easier to read. 

I have some questions about what this PR means for current installations: 
* How will current databases work with this new encryption system? All of our current installation databases are encrypted with the MySQL `PASSWORD` function. They will no longer match if they are being checked against a bcrypt string. 
* What is the performance of `bcrypt` like on the Raspberry Pi? I noticed that the dependency is built using node-gyp when running `npm install`, it would be good to check this on the Pi's ARM architecture.

sfount:
@DedrickEnc I've updated the base branch to `IMA-WorldHealth/feature-bcrypt`. That will allow us to think about how to formally update current installation sites, keep this feature up to date with master but without accidentally pushing it to hospitals.

server/controllers/auth.js
jniles: Hmm.. I can see the usability argument.

In my view, the reason we want to use `bcrypt` over `PASSWORD()` is to start getting serious about enterprise security, in case we ever actually do have to conform to something like HIPPA requirements.  So, in my view, we should try and adopt security best practices.

According to [The Open Web Application Security Project](https://www.owasp.org/index.php/Authentication_Cheat_Sheet), you should always give a generic error message.  Here's the quote:
> An application should respond with a generic error message regardless of whether the user ID or password was incorrect. It should also give no indication to the status of an existing account.

For this reason, I'm a fan of using a uniform authentication failure message.  However, I can be swayed if our users can't figure it out.

DedrickEnc:
- Yes, to roll out this feature in our running system (Vanga for example) without writing additional script, we should encrypt a common password with bcrypt, for example **1234** to get a hashed string **X** and then update every user password to **X**. So every user can change his password after being logged in by using **1234** as password. 
- About comment, I will address them when necessary, because we are focused on report now
- Thank you @sfount to make this branch a feature branch.
- @jniles I will have a look to the link you shared to see how can I amend this pull request.

sfount:
Closing this PR until it can be tackled with a direct upgrade path for current institutions - the code here will still be referenced from https://github.com/IMA-WorldHealth/bhima-2.X/issues/1754

jniles:
Both this and https://github.com/IMA-WorldHealth/bhima-2.X/pull/2249 will have breaking changes to users and permissions.  We should do both upgrades at the same time.

jniles:
I'm currently at Vanga, and can assist users with password changes as needed.  Let's see if we can get this in.

It looks like [some folks are having luck with bcyrpt on the raspberry pi](https://github.com/TryGhost/Ghost/issues/1207#issuecomment-28681332).  The pure JS implementation is slow, but the implementation here seems to work fine for those reported use cases.

PR #1774, opened 2017-06-08
fix(filters): Invoice filter custom period start

sfount:
This commit resolves a typo in the custom period start filter for the
invoice registry.

It also introduces the comparitor alias for the FilterService, allowing
custom comparitors to be assigned to period start and end dates.

### DIFF
--- a/client/src/js/services/FilterService.js
+++ b/client/src/js/services/FilterService.js
@@ -30,7 +30,7 @@ function FilterService(Store) {
   //       with a toggle between the array to populate and the default value
   FilterList.prototype.registerDefaultFilters = function registerDefaultFilters(filterDefinitions) {
     var formattedFilters = filterDefinitions.map(function (filterDefinition) {
-      var filter = new Filter(filterDefinition.key, filterDefinition.label, filterDefinition.valueFilter);
+      var filter = new Filter(filterDefinition.key, filterDefinition.label, filterDefinition.valueFilter, filterDefinition.comparitor);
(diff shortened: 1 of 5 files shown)


client/src/js/services/FilterService.js
```diff
@@ -30,7 +30,7 @@ function FilterService(Store) {
   //       with a toggle between the array to populate and the default value
   FilterList.prototype.registerDefaultFilters = function registerDefaultFilters(filterDefinitions) {
     var formattedFilters = filterDefinitions.map(function (filterDefinition) {
-      var filter = new Filter(filterDefinition.key, filterDefinition.label, filterDefinition.valueFilter);
+      var filter = new Filter(filterDefinition.key, filterDefinition.label, filterDefinition.valueFilter, filterDefinition.comparitor);
```
jniles, line 33: Yay!  @mbayopanda and I had a discussion about this ... really great to see it fixed!

server/controllers/finance/patientInvoice.js
```diff
@@ -232,7 +232,7 @@ function find(options) {
   );
 
   filters.period('period', 'date');
-  filters.dateFrom('custion_period_start', 'date');
+  filters.dateFrom('custom_period_start', 'date');
```
jniles, line 235: This is the disadvantage of not using copy and paste... 👎

jniles (approved):
Great fixes 👍

PR #1775, opened 2017-06-08
feature(implement inline filtering on users)

### LINKED ISSUE #1750: feature: implement inline filtering on users
The users grid could probably use an inline filtering toggle to allow an administrator to quickly find disabled users, or users by their username.  For a large enterprise, you can image that they would have many users.

lomamech:
Implement inline filtering on users
closes #1750

### DIFF
--- a/client/src/js/services/grid/GridColumns.js
+++ b/client/src/js/services/grid/GridColumns.js
@@ -2,7 +2,7 @@ angular.module('bhima.services')
 .service('GridColumnService', GridColumnService);
 
 GridColumnService.$inject = [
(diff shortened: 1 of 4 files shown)


jniles (approved):
LGTM.  Not that this also fixes translation - the module didn't appear to be translated before.  Great work! 👍

jniles:
@lomamech your commit messages are a good step in the right direction!  The best way, is to write one or two words for the module concerned.  In this case, it would look like:
```
feature(users): implement inline filtering
```

That way our changelog will be able to parse your message.

sfount (approved):
LGTM. Useful refactor!

PR #1776, opened 2017-06-08
fix(Column configuration limits length of visible columns)

### LINKED ISSUE #1759: bug: Column Configuration limits length of visible columns to 3
The Column Configuration modal isn't very portable since it doesn't know how to filter out special ui-grid plugin columns.  It should filter those out first, and then warn the user about selecting less than 1 column.

lomamech:
Update column hasEnoughColumns
- change default value column set 1
- Update structure alternative check if the firth and second element of ui
  grid are visible for the incrementation of defaultValueColumn

closes #1759

### DIFF
--- a/client/src/js/services/grid/GridColumns.js
+++ b/client/src/js/services/grid/GridColumns.js
@@ -114,18 +114,19 @@ function GridColumnService(uiGridConstants, AppCache, Modal, util, $timeout) {
   Columns.prototype.hasEnoughColumns = function hasEnoughColumns(columns) {
     var grid = this.gridApi.grid;
     var visibleColumn = 0;
-    var defaultValueColumn = 3;
+    var defaultValueColumn = 1;
     var totalColumn = 0;
 
     angular.forEach(columns, function (visible, field) {
       var column = grid.getColumn(field);
-
+      
       /**
-        *This alternative structure checks if the first element of the grid is visible or not,
-        *finally to reinitialize the number of column by default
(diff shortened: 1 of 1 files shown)


client/src/js/services/grid/GridColumns.js
```diff
@@ -114,18 +114,19 @@ function GridColumnService(uiGridConstants, AppCache, Modal, util, $timeout) {
...
       /**
-        *This alternative structure checks if the first element of the grid is visible or not,
-        *finally to reinitialize the number of column by default
+        *This alternative structure checks if selectionRowHeaderCol or treeBaseRowHeaderCol are visible or not,
+        *finally to incremente the number of column by default
       */
-      if((column.name === "treeBaseRowHeaderCol") && !visible){
-        defaultValueColumn--;
+
+      if((column.name === "selectionRowHeaderCol" || column.name === "treeBaseRowHeaderCol") && visible){
```
jniles, line 128: This is the correct way to do this :+1:

jniles: Do you have eslint installed? I think it asks for a space between the `if` and the `(`.  Like this:
```js
if (x) {
  // something
}
```

The reason is to try and separate function calls and control flow code.  It makes it easy for a programmer (without syntax highlighting) to realize at a glance that `if` is not a function call - it is a control flow statement.

jniles (approved):
LGTM. I haven't tested this, but it's definitely the correct approach.  :+1:

Issue #1766, opened 2017-06-05
(feat) Email pushing report


jeremielodi:
This issue concerne pushing report by email to a groupe of people (Administrator, suppliers, donnators,.. ) every day, every week or every month.

We will need first of all a report group where we can specify each category of report

![1_report group](https://cloud.githubusercontent.com/assets/25838121/26790786/e7d18ed2-4a0c-11e7-8d9f-11791ca7ebed.png)

After, we'll create a email groupe, each email should be associated to a report group
![2_new email for reporting copy](https://cloud.githubusercontent.com/assets/25838121/26790910/57714372-4a0d-11e7-8fee-fe62609b229b.png)

If we want to pick a email from an existing user or employee we can do it like this
![3_select email origin](https://cloud.githubusercontent.com/assets/25838121/26790961/89711a14-4a0d-11e7-82cd-93cef79f8c06.png)

![4_select an employee](https://cloud.githubusercontent.com/assets/25838121/26790982/9a452ace-4a0d-11e7-9154-e50c36a5c840.png)

this issue linked to #1688

jniles:
Those are fantastic markups @jeremielodi!  Linking the old issue is also excellent :+1:

Here is so quick feedback:
 1. I like the idea of optionally allowing a user to import emails.  I think the link is best placed next to the email field, instead of next to the name.
 2. A small thing - the "create" button should be blue to signify an action to the user.

Again, this is a really great first issue.  We can talk about this in person tomorrow.

jeremielodi:
Okay, thanks

jniles:
Okay, now that I've had a chance to look at this more:

It seems to me the "code" field for a report_group is unnecessary - it seems more natural to give it a name.  For example "Monthly Cashflow Report".  This name should be unique though.

On your mockups, you include a checkbox column. We generally try to avoid using checkboxes for adding things, unless we are going to do a multi-select.  It doesn't look like you need a multiple select in this case.  You can probably just embed a link in an "actions" column that allows a user to both edit or delete the row - kind of like the report archive page.

Obviously these mockups are just examples and not the final version.  But! Just adding my feedback.

Issue #150, opened 2016-03-01
Create a "Type Text to Confirm Delete" modal


jniles:
One several pages, we allow uses to delete cashboxes, projects, enterprises, or others.  Currently, we create a `$window.confirm()` instance such as the one shown below
##### $window.confirm()

![window confirm](https://cloud.githubusercontent.com/assets/896472/13419746/ebc21cac-df80-11e5-9383-7ee56ada5ac9.png)

There are a few small problems with this:
1. After some number of alerts, the user can click the checkbox "prevent this window from making any more dialogs", disabling some functionality in the application.
2. A click-to-confirm is better than no click at all.  However, for processes that have huge implications, we might want more security -- typing in the name of the cashbox/project/entity would ensure the use knew exactly which entity they were removing.
3. From a developer perspective, this isn't a great solution:
   1. A developer has no control over the style of the `$window.confirm()` window.
   2.  A developer must manually translate the text using the `$translate` filter in the controller before display
   3. The dialog blocks the view without the choice of making a "click somewhere else to close".
   4. The user can disabled this feature by accident.
4. From a user perspective, this isn't a great solution:
   1. `$window.confirm()` windows look very different from the application, potentially leading to confusion.
   2. The text size and font is independent of the application's size and font, making this choice less accessible.
   3. The user could disabled this feature without realizing it by simply clicking a checkbox.  It is unclear how to re-enable the delete process without shutting down the browser and restarting.

I propose that we create a confirm modal similar to Github's.  See below:
##### Confirm Modal

![confirmmodal](https://cloud.githubusercontent.com/assets/896472/13419940/690dcb56-df82-11e5-95c7-ddb435044f37.png)

This could be imported into any module and used similar to how we use `$window.confirm()`, except with promises.  For example:

``` js
angular.module('bhima.controllers')
.controller('ExampleController', ExampleCtrl);

ExampleCtrl.$inject = [ 'ConfirmModalService' ];

/**
 * Modal.open() looks like this:
 *  service.open = function openModal(match) { 
 *  var instance = $uibModal.open({
 *     controller : 'ConfirmModalController as ConfirmModalCtrl',
 *    templateUrl : 'path/to/the/template.tmpl.html',
 *    resolve : { match : function () { return match; } }
 *    size : 'md'
 *  });
 *  return instance.result;
 * };
 */

function ExampleCtrl(Modal) {
  vm.confirm = function confirm() {
    Modal.open(vm.textToMatch)
    .then(function () {
[...]

PR #1778, opened 2017-06-09
feat(emailing report) : add new modules for emailing report

jeremielodi:
in this pull request I try to program the mockups I did in the issue #1766


client/src/modules/report-group/reportGroup.ctrl.js
```diff
@@ -0,0 +1,182 @@
...
+
+
+/**
+ * Report group Controller
+ *
+ * @description
+ 
+ */
+
+function ReportGroupController($state, ReportGroupSvc, Session, util, Notify, ScrollTo, bhConstants, uiGridConstants) {
```
sfount: This file should be updated to meet the project whitespace standard.

Full reference for the style we (generally) stick to: https://github.com/airbnb/javascript#whitespace

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,116 @@
+<div class="flex-header static">
```
sfount, line 1: This file should be updated to meet the project whitespace standard.

Full reference for the style we (generally) stick to: https://github.com/airbnb/javascript#whitespace

client/src/modules/email-report/modal/categories-people-list-ctrl.js
```diff
@@ -0,0 +1,124 @@
...
+
+
+/**
+ * Report group Controller
+ *
+ * @description
+ 
+ */
+
+function CategoriesPeopleListController($state, EmailReportSvc, CategoriesPeople, Session, util, Notify, ScrollTo, bhConstants, uiGridConstants) {
```
sfount: This file should be updated to meet the project whitespace standard.

Full reference for the style we (generally) stick to: https://github.com/airbnb/javascript#whitespace

client/src/modules/email-report/modal/categories-people-list-service.js
```diff
@@ -0,0 +1,45 @@
...
+    'DocumentService', 'VisitService', 'DepricatedFilterService',
+];
+
+/**
+ * @module EmailReportService
+ *
+ * This service is responsible for providing an interface between angular
+ * module controllers and the server /tools/bed/ API.
+ */
+function CategoriesPeopleService($http, util, Session, $uibModal,
```
sfount: This file should be updated to meet the project whitespace standard.

Full reference for the style we (generally) stick to: https://github.com/airbnb/javascript#whitespace

client/src/modules/email-report/modal/liste-categories-people-modal.html
```diff
@@ -0,0 +1,30 @@
+<!-- billing service form -->
+<form name="BillingServicesForm" bh-submit="BillingServicesFormCtrl.submit(BillingServicesForm)" novalidate>
```
sfount: This file should be updated to meet the project whitespace standard.

Full reference for the style we (generally) stick to: https://github.com/airbnb/javascript#whitespace

client/src/modules/report-group/templates/report-group-action-cell.html
```diff
@@ -0,0 +1,20 @@
+<div class="ui-grid-cell-contents" data-row-item="{{ row.entity.code}}" uib-dropdown dropdown-append-to-body>
```
sfount: This file should be updated to meet the project whitespace standard.

Full reference for the style we (generally) stick to: https://github.com/airbnb/javascript#whitespace

server/controllers/email-report/weeklySummaryReport.handlebars
```diff
@@ -0,0 +1,105 @@
+{{> head title="PATIENT_RECORDS.FINANCIAL_ACTIVITY.REPORT_TITLE" }}
```
sfount, line 1: This file should be updated to meet the project whitespace standard.

Full reference for the style we (generally) stick to: https://github.com/airbnb/javascript#whitespace

server/controllers/report-group/mail.js
```diff
@@ -0,0 +1,145 @@
+/**
+ * @module email-report/report-group
```
sfount: This file should be updated to meet the project whitespace standard.

Full reference for the style we (generally) stick to: https://github.com/airbnb/javascript#whitespace

server/controllers/report-group/mail.js
```diff
@@ -0,0 +1,145 @@
...
+
+    //formation the attach name it should include the date
+
+    var dat = moment().format('YYYY MM DD HH[h]');
+
+    var attch = new mailgun.Attachment({ data: file, filename: dat + ' ' + fileName });
+
+    // structuring all infomations about an email  when using mailgun API
+    var data = {
+        from: 'Mailgun Sandbox <postmaster@sandbox666c0ed4d4d24f0a98a503fc5fe16b0e.mailgun.org>',
```
sfount: These details should not be hard coded into the servers design. Would it be possible to use the environment files for this ? 

We current use .env files to load in our MySQL and server configurations. Maybe something similar could be done with Mailgun details.

server/controllers/email-report/weeklySummaryReport.handlebars
```diff
@@ -0,0 +1,105 @@
...
+        {{#each servicesIncome}}
+        <tr>
+            <td style="width: 50%">{{this.serviceName}}</td>
+            <td class="text-right">{{this.sumCost}}</td>
+
+        </tr>
+        {{/each}} 
+        {{#unless servicesIncome}}
+        <tr colspan='2'>
+            <td>No income found in services.</td>
```
sfount: All text should be translated so that it can be displayed both in French and English. This should be done using the translate directive, for example: 

```
{{translate SERVICES.NO_INCOMES}}
```

server/controllers/email-report/weeklySummaryReport.handlebars
```diff
@@ -0,0 +1,105 @@
...
+        {{#each MaxMinInvoiceDate}}
+        <tr>
+            <td>{{timestamp this.minDate}}</td>
+            <td>{{timestamp this.maxDate}}</td>
+
+        </tr>
+        {{/each}}
+        {{#unless MaxMinInvoiceDate}}
+        <tr colspan='2'>
+            <td>There was no invoice this week.</td>
```
sfount: This should be translated.

server/controllers/report-group/mail.js
```diff
@@ -0,0 +1,145 @@
...
+        to: 'Jeremie Lodi<jeremielodi@gmail.com>',
+        bcc: bcc_emails,
+        subject: 'BHIMA reports',
+        text: 'The report is attached',
```
sfount: This text should be translated.

server/controllers/email-report/weeklySummaryReport.js
```diff
@@ -0,0 +1,213 @@
...
+                JOIN project ON project.id = invoice.project_id
+            WHERE DATE(invoice.date) >= DATE(?) AND 
+                  DATE(invoice.date) <= DATE(?)  
+            GROUP BY service.id `;
+
+
+  var period = new Period(new Date());
+  var week = period.periods.week.limit;
+
+  var monday = '2016-01-01 00:00:00.000';// week.start();
```
sfount: These hardcoded values should not be included in the actual code base! I think these were used to test.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,116 @@
...
+                                <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+                            </div>
+                        </div>
+                    </div>
+                    <!-- name -->
+                    <div class="form-group" ng-class="{'has-error' : ReportGroupRegistrationForm.name.$invalid && ReportGroupRegistrationForm.$submitted}">
+
+
+                        <label class="control-label ng-binding">
+                            Name
```
sfount: All text here should be translated using 

```html
<label translate>EXAMPLE.NAME</label>
```

server/controllers/report-group/index.js
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * @method create
+ *
+ * @description
+ 
+ * POST /email-report API
+ */
+function update(req, res, next) {
+
+  const sql = 'UPDATE report_group set code=?, name=?, description=? WHERE code=?;';
```
sfount, line 73: The MySQL library `?` syntax allows you to do `SET` with an object for key:value pairs. 

For example: 

```js
const sql = 'UPDATE report_group SET ?';

const values = { 
   code : // ...
   name : // ... 
   description : //...
};

db.exec(sql, values)
//...
```

server/controllers/email-report/weeklySummaryReport.js
```diff
@@ -0,0 +1,213 @@
...
+
+
+//one of the reports that bhima send by email
+function weeklySummaryReport(req, res, next) {
+
+
+  const options = {
+    renderer: 'pdf',
+    saveReport: false,
+    filename: 'test_summary_report.pdf',
```
sfount, line 183: The file name here could be formatted as the current weeks time stamp - it looks like the filename here was just for testing purposes.

server/controllers/email-report/weeklySummaryReport.js
```diff
@@ -0,0 +1,213 @@
...
+  var period = new Period(new Date());
+  var week = period.periods.week.limit;
+
+  var monday = week.start();
+  var sunday = week.end();
+
+  return db.exec(sql, [monday, sunday]);
+}
+
+function view1(req, res, next) {
```
sfount: This function name has no meaning - it would be better to give this function a meaningful name and maybe a comment describing what the purpose of this function is.

server/controllers/email-report/weeklySummaryReport.js
```diff
@@ -0,0 +1,213 @@
...
+
+  const options = {
+    renderer: 'pdf',
+    saveReport: false,
+    filename: 'test_summary_report.pdf',
+    orientation: 'portrait',
+  };
+
+  //patient registered
+  return loadPatientData().then(_patientsData => {
```
sfount: This promise chain is set up using promises like callbacks and is very difficult to follow. 

Currently this code works like this 
```js
return firstFunction()
.then(() => { 
    return secondFunction()
          .then(() => { 

             return thirdFunction()
             .then(() => { 
                //...
              });
          })
}
```

However using promises allows us to do all of these calls in a linear flat structure, one after the other - this makes the code much easier to read and allows error handling to propagate correctly. 

For example: 

```js
return firstFunction()
  .then((firstResult) => { 
    
    return secondFunction()
  })
  .then((secondResult) => { 
    
    return thirdFunction();
  })
  .then((thirdResult) => { 
   
    //...
  });
```

sfount (commented):
This looks like a great first attempt at integrating emails into the system! I have seen the functionality and it works well. 

I have made a number of comments on the code at places I think it could be improved, there are two main things to do here: 
1. Many of these files are created using tabs and 4 spaces. The repositories style is defined as using 2 spaces. It would be good to make sure these files fit with that style. (https://github.com/airbnb/javascript#whitespace)
2. There are a number of places where test data and hardcoded values are used. These should be updated so that the feature will work as expected. 

Great work so far! A lot of the Angular and server side code is looking good.

server/controllers/report-group/mail.js
jniles: To follow up:

We use a the file `.env.development` to hold our environment variables.  If you write a variable in that file like
```sh
# demonstrate a variable
SOME_VAR=1
```
... it will be automatically imported and added to the `process.env` global object in NodeJS.  The example above can be accessed from any node module like this:

```js
if (process.env.SOME_VAR === 2) {
  console.log('This isn\'t right!');
} else if (process.env.SOME_VAR === 1) {
  console.log('that\'s better');
} else {
  return null;
}
```

----- 

In your case, it would be best to move the mailgun credentials into the `.env.development` file.

jniles: See the NPM module we use for this functionality for more information:
https://www.npmjs.com/package/dotenv

jniles:
@jeremielodi do you need any help on these changes?  How's it going?

jeremielodi:
Hi !, I have almost  finished, i think that I will send a PR today. I had
somme trouble with the  computer


Le 13 juil. 2017 22:06, "Jonathan Niles" <notifications@github.com> a
écrit :

jniles:
@jeremielodi let me know when you need/want another review.

jniles:
@jeremielodi what is the status of this PR?

.env.development
```diff
@@ -12,12 +12,18 @@ SESS_SECRET='XopEn BlowFISH'
...
 UPLOAD_DIR='client/upload'
 
 # this will toggle all MYSQL errors to be reflected to the client.
 LOG_ALL_MYSQL_ERRORS=0
 
 # control Redis Pub/Sub
 ENABLE_EVENTS=false
+
+# MAILGUN API LOGGIN PARAMS
+MAILGUN_API_KEY = key-c30213afb78d0d0e53a0f973c3db7021
```
jniles: This is the correct location for these :+1:

client/src/modules/email-report/modal/liste-categories-people-ctrl.js
```diff
@@ -0,0 +1,30 @@
+angular.module('bhima.controllers')
+    .controller('CategoriesPeopleModalController', CategoriesPeopleModalController);
+
+CategoriesPeopleModalController.$inject = [
```
jniles, line 4: It looks like you don't use `bhConstants`.   Could you look through all the `Controller.$injects` you define and remove unused imports?  They will confuse future developers.

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,216 @@
...
+
+
+  }
+
+
+  // reset the form state
+  function resetForm(RegistrationForm) {
+
+    vm.emailReport = {};
+    vm.selectedEmailReport = {};
```
jniles, line 99: You could actually do this in the same line:
```js
vm.selectedEmailReport = { selected : false };
```

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,216 @@
...
+    RegistrationForm.$setUntouched();
+    ScrollTo('anchor');
+  }
+
+
+  //inserting a new email for a  reporting in the database
+  function save(RegistrationForm) {
+
+    // end propagation for invalid state - this could scroll to an $invalid element on the form
+    if (RegistrationForm.$invalid) {
```
jniles: :+1:

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,216 @@
...
+      return Notify.danger('FORM.ERRORS.INVALID');
+    }
+
+    if (vm.selectedEmailReport.selected === false) {
```
jniles: This type of logic is hard to read.  You can create a variable that provides some more information in the name like this:
```js
var isCreateReport = vm.selectedEmailReport.selected === false;

if (isCreateReport) {
  // do specific create code
} else {
  // do other things
}
```

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,216 @@
...
+
+      //calling the EmailReportService create method
+      return EmailReportSvc.create(vm.emailReport)
+        .then(function (confirmation) {
```
jniles: In both branches of this `if/else`, you use almost the exact some code!   Following the principle [Don't Repeat Yourself](http://deviq.com/don-t-repeat-yourself/) (DRY), we should try to only write this logic once.  Fortunately, this is easy!

Instead of writing:
```js
if (isCreateReport) {
  return EmailReportSvc.create(vm.emailReport)
    .then( /* ... */ );
}  else {
  return EmailReportSvc.update(vm.emailReport
    .then( /* ... */ );
}
```
... you could reuse the code without branching like this:

```js
// check to see if we are creating or updating a report
var isCreateReport = vm.selectedEmailReport.selected === false;

// Change the HTTP request depending on the module state
var request = isCreateReport ? EmailReportSvc.create(vm.emailReport) : EmailReportSvc.update(vm.emailReport);

// change the `alert()` method according to the type of request
var message = isCreateReport ? 'Created Successfully' : 'Updated Successfully';

// now we can use the same logic
return promise
  .then(function () {
    alert(message);
    load();
    resetForm(RegistrationForm);
  })
  .catch(Notify.handleError);
 ```

This style of code is easier to read since there are fewer nested branches.  It also clearly shows that the function will only return one thing - a promise.

File: client/src/modules/email-report/email-report.ctrl.js [hunk 25]
```diff
@@ -0,0 +1,216 @@
...
+
+  // options for the UI grid
+  vm.gridOptions = {
+    appScopeProvider: vm,
+    enableColumnMenus: false,
+    columnDefs: columns,
+    enableSorting: true,
+    showColumnFooter: true,
+    fastWatch: true,
+    flatEntityAccess: true,
```
jniles, line 217: :+1: These are good performance best practices!  Well done.

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,216 @@
...
+      displayName: 'FORM.LABELS.REPORT_GROUP',
+      headerCellFilter: 'translate',
+    },
+    {
+      field: 'action',
+      displayName: '',
+      cellTemplate: '/modules/email-report/templates/email-report-action-cell.html',
+      enableFiltering: false,
+      enableSorting: false,
+      enableColumnMenu: false,
```
jniles: Since you set `enableColumnMenu` on the gridOptions, you don't need to have this line.

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+
+  //delete a record
+  function remove(RegistrationForm) {
+
+    //calling the EmailReportService create method
+    return EmailReportSvc.remove(vm.selectedEmailReport.id)
+      .then(function (confirmation) {
+        alert('deleted successfully');
```
jniles: The native `browser.alert()` can be blocked or disabled by users.  This leads to a really bad user experience (it makes it look like BHIMA is not working).  See the discussion on https://github.com/IMA-WorldHealth/bhima-2.X/issues/150.

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,216 @@
...
+    fastWatch: true,
+    flatEntityAccess: true,
+  };
+
+
+  //fill the grid
+  function load(filters) {
+
+    EmailReportSvc.read().then(function (reportGroups) {
+      vm.loading = false;
```
jniles: The grid will load forever if this encounters an error, since `vm.loading` will still be true after the `.catch()` method fires.  As a best practice, you should use `.finally()` to toggle error handling on and off.  See the [Patient Registry](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/patients/registry/registry.js#L154) as an example of this behavior.

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,216 @@
...
+      vm.gridOptions.data = reportGroups;
+
+    })
+      .catch(Notify.handleError);
+  }
+
+  load();
+
+  return vm;
```
jniles: You don't need the `return vm;` statement.  You only need to `return this` from angular.services.

client/src/modules/email-report/email-report.routes.js
```diff
@@ -0,0 +1,83 @@
...
+
+  return function onEnter($state, Modal, Notify) {
+    Modal.open({
+      templateUrl: 'modules/email-report/modal/liste-categories-people-modal.html',
+      controller: ctrl,
+      backdrop: 'static',
+      keyboard: false,
+    }).result
+      .then(function (id) {
+        Notify.success(message);
```
jniles: Where is message defined?

client/src/modules/email-report/email-report.routes.js
```diff
@@ -0,0 +1,83 @@
...
+      .catch(function (error) {
+        if (error) {
+          Notify.handleError(error);
+        }
+      });
+  };
+}
+
+
+function onSelectProfileCategoryFactory(stateType) {
```
jniles: You could combine these two functions - `onSelectProfileFactory` and `onSelectProfileCategoryFactory`.  It would look something like this:
```js
function onSelectFactory(stateType) {
  var isCategoryState = (stateType === 'category');
  
  var ctrl = isCategoryState ?
    'CategoriesPeopleListController as catPeopleListCtrl' :
    'CategoriesPeopleModalController as catPeopleModalListCtrl';

   var templateUrl = isCategoryState ?
     'modules/email-report/modal/categories-people-list-modal.html' :
     'modules/email-report/modal/liste-categories-people-modal.html';

  return function onEnter($state, Modal, Notify) {
    Modal.open({
      templateUrl: 'modules/email-report/modal/categories-people-list-modal.html',
      controller: ctrl,
      backdrop: 'static',
      keyboard: true,
   }).result
    .then(function () {
      Notify.success(message);
    })
    .catch(function (error) {
      Notify.handleError(error);
    });
  };
}

File: client/src/modules/email-report/email-report.service.js [hunk 32]
```diff
@@ -0,0 +1,95 @@
...
+ * @module EmailReportService
+ *
+ * This service is responsible for providing an interface between angular
+ * module controllers and the server /email-report/ API.
+ */
+function EmailReportService($http, util, Session, $uibModal, Documents, Visits, Filters) {
+
+  var service = this;
+
+  service.frequencies = ["Dayly", "Weekly", "Monthly"];
```
jniles: This is a nice way to store data!

The English spelling should be `Daily` instead of `Dayly`.  The rest look good :+1:

client/src/modules/email-report/email-report.service.js
```diff
@@ -0,0 +1,95 @@
...
+
+  var baseUrl = '/email-report/';
+  var baseUrl_list_people = '/email-report/list-people';
+  var baseUrl = '/email-report/';
```
jniles: You have repeated the `baseUrl` here.  You can remove it.

client/src/modules/email-report/email-report.service.js
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * This method accepts information recorded by a controllers form, formats it
+   * for submission and forwards it to the server /email-report API. It can
+   * be used for creating new email for reporting record in the database.
+   *
+   * @param {Object} email_report   email for reporting information.
+   * @returns {Object}          Promise object returning success/failure confirmation.
+   */
+  function create(email_report) {
+    var formatBedRequest = {
```
jniles: Why is this called format bed request?

client/src/modules/email-report/email-report.service.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /*
+   reading data from the server, 
+   here we load a groupe of people(users, patients,donators,..),
+   the group is detected once the user select an item in the modal
+   , also the query if formed by the item's infomation 
+   Those items come from CategoriesPeopleService in the modal folder
+   */
+  function read_people(query) {
```
jniles: A function called `read` should never use an HTTP POST request.  In general, you should have:
 - js read() --> HTTP GET --> SQL SELECT
 - js create() --> HTTP POST --> SQL CREATE
 - js update() --> HTTP PUT --> SQL UPDATE
 - js remove() --> HTTP DELETE --> SQL DELETE

 See the [API guidelines in the wiki](https://github.com/IMA-WorldHealth/bhima-2.X/wiki/Design:-Server-CRUD-APIs) for more details.

client/src/modules/email-report/email-report.service.js
```diff
@@ -0,0 +1,95 @@
...
+    return $http.post(baseUrl_list_people, { sql: query })
+      .then(util.unwrapHttpResponse);
+  }
+
+  /*
+   delete an email report (a profile)
+   */
+  function remove(_id) {
+    return $http.delete(baseUrl + _id, {})
```
jniles: You don't need the final `{}` here.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
+<div class="flex-header static">
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static">{{ "TREE.ADMIN" | translate }}</li>
```
jniles: Please use the `translate` directive.  This should be:
```html
<li class="static" translate>TREE.ADMIN</li>
```

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
...
+    </ol>
+
+
+    <div class="toolbar">
+      <div class="toolbar-item">
+        <a class="btn btn-default pull-right" 
+          ui-sref="reportGroup"
+          ui-sref-opts="{notify:false}">
+            <i class="fa fa-plus"></i>
+            <span translate="" class="ng-scope">FORM.LABELS.REPORT_GROUP</span>
```
jniles: Please remove all `class="ng-scope"`.  These will only confuse debugging tools.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
...
+
+    <div class="col-lg-12">
+      <!-- start new Reporting group form -->
+      <div class="col-lg-5">
+        <h4 translate>FORM.LABELS.NEW_PROFILE</h4>
+
+        <form
+          name="EmailReportRegistrationForm"
+          bh-submit="EmailReportCtrl.save(EmailReportRegistrationForm)"
+          bh-form-defaults
```
jniles: Please remove this `bh-form-defaults`.  The `bhFormDefaults` directive was removed.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
...
+                  FORM.LABELS.SELECT_PROFILE
+              </a>
+            </div>
+
+
+            <!-- name -->
+            <div class="form-group" 
+              ng-class="{'has-error':EmailReportRegistrationForm.name.$invalid && EmailReportRegistrationForm.$submitted}">
+
+              <label class="control-label ng-binding" translate>
```
jniles: Please remove all `class="ng-binding"`.  It will only confuse debugging tools.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
...
+
+              <div class="input-group col-lg-12">
+
+                <select
+                  required=""
+                  name="frequency"
+                  ng-model="EmailReportCtrl.emailReport.frequency"
+                  class="form-control">
+                    <option  
+                      ng-repeat="item in EmailReportCtrl.frequencies"
```
jniles: `ng-repeat` has really bad performance.  In general, you should avoid it where possible.  You can see some discussion of this a long time ago in https://github.com/IMA-WorldHealth/bhima-1.X/issues/143.

In this case, you can use `ng-options` on the `<select>` element.  It would look like this:
```html
<select
  name="frequency"
  ng-model="EmailReportCtrl.emailReport.frequency"
  ng-options="item as item in EmailReportCtrl.frequencies"
  class="form-control"
  required>
</select>
```
It is also best practice to order all options in alphabetical order.  You can do this pretty easily by just adding an `| orderBy: item`.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
...
+             ng-show="EmailReportCtrl.selectedEmailReport.selected==false"
+             class="pull-right" loading-state="BedRegistrationForm.$loading">
+              {{ "FORM.BUTTONS.SAVE" | translate }}
+            </bh-loading-button>
+
+            <span class="btn btn-danger" 
+              ng-click="EmailReportCtrl.remove(EmailReportRegistrationForm)"
+              ng-show="EmailReportCtrl.selectedEmailReport.selected==true"
+              class="pull-right">
+                {{ "FORM.BUTTONS.DELETE" | translate }}
```
jniles: Please use the `translate` directive.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
...
+                    </div>
+                </div>
+              </div>
+            </div>
+
+
+
+            <bh-loading-button 
+             ng-show="EmailReportCtrl.selectedEmailReport.selected==false"
+             class="pull-right" loading-state="BedRegistrationForm.$loading">
```
jniles: Instead of having multiple `pull-right` classes, it would be best to wrap all buttons in an element that is `pull-right`.  Something like this:

```html
<div class="pull-right">
   <button></button>
   <button></button>
   <button></button>
</div>
```

That way the browser does less calculation when the layout changes.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
...
+
+      <div class="col-lg-7">
+
+        <!--the grid -->
+        <div 
+          id="report-groups-grid"
+          ui-grid-auto-resize
+          ui-grid-resize-columns
+          ui-grid-move-columns
+          ui-grid-cellNav
```
jniles: Do you really need cellNav?

It seems like a lot of these are not used - could you remove all the ones you do not use?  They will only slow down the page.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
...
+
+            <bh-loading-button 
+             ng-show="EmailReportCtrl.selectedEmailReport.selected==false"
+             class="pull-right" loading-state="BedRegistrationForm.$loading">
+              {{ "FORM.BUTTONS.SAVE" | translate }}
+            </bh-loading-button>
+
+            <span class="btn btn-danger" 
+              ng-click="EmailReportCtrl.remove(EmailReportRegistrationForm)"
+              ng-show="EmailReportCtrl.selectedEmailReport.selected==true"
```
jniles: You don't need the equality here.  You can actually just write:
```html
ng-show="EmailReportCtrl.selectedEmailReport.selected"
```

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,217 @@
...
+            </bh-loading-button>
+
+        </form>
+
+      </div>
+      <!-- end new Reporting group form -->
+
+      <!-- start grid of emails -->
+
+      <div class="col-lg-7">
```
jniles: The rest of the application uses `col-md-*`.  To make sure everything works the same way across multiple screen sizes, you should use `col-md-7`.

You should also update all places where this is used.

client/src/modules/email-report/modal/categories-people-list-ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+  vm.loading = false;
+  vm.hasError = false;
+
+  function load(filters) {
+    /*
+     category people index is save in the localStorage as expained after
+     if there's no index, we can use it, in the most case when the user refresh the page
+     */
+    if (!$state.params.index_current_category) {
+      $state.params.index_current_category = localStorage.getItem('index_current_category');
```
jniles: Aha, I see you are using `localStorage`!  We actually have that built into the application, called `AppCache`.  You can use it like this:

```js
SomeCtrl.$inject = ['appcache'];

function SomeCtrl(AppCache) {
  var cache = new AppCache('SomeKey');

   // anything you bind to cache will be saved across sessions
   console.log('cache.name', cache.name);

    // this is how you save something - simple JS assignment. 
   cache.name = 'Hello World';
}
```

Therefore, you can modify your code to use this:
```js
// assuming you have injected AppCache
var cache = AppCache('ReportGroupCache');

// this is the same as your code:
$state.params.index_current_category = $state.params.index_current_category || cache.index_current_category;
```

Much simpler!

jniles: If you want to see the implementation of AppCache, you can find it in our services: [AppCache.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/AppCache.js).

client/src/modules/email-report/modal/categories-people-list-ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    //if a categry of peoplo has been selected from the list,
```
jniles: These are really nice comments :+1:

client/src/modules/email-report/modal/categories-people-list-ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+      //saving the index in the localstorage because the user can refresh the page
+      // and state can be deleted
+      localStorage.setItem('index_current_category', index);
+
+      var currentCategory = CategoriesPeople.list[index];
+      vm.selectedCategory=currentCategory;
+      /*
+       we know now in what table in the database we request
+       */
+      var sql = "SELECT " + currentCategory.Columns[0] + " , " + currentCategory.Columns[1] + " FROM " + currentCategory.table;
```
jniles: It is really dangerous to put SQL on the client, since an attacker can modify the string and do anything they want.  Please move this code to the server.

client/src/modules/email-report/modal/categories-people-list-ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+
+    //if a categry of peoplo has been selected from the list, 
+    //we can use that index in order to create an sql request whitch return a list of people
+    if ($state.params.index_current_category) {
+
+      var index = $state.params.index_current_category;
+
+      //saving the index in the localstorage because the user can refresh the page
+      // and state can be deleted
+      localStorage.setItem('index_current_category', index);
```
jniles: Using `AppCache` this would be:
```js
cache.index_current_category = index;
```

client/src/modules/email-report/modal/categories-people-list-ctrl.js
```diff
@@ -0,0 +1,126 @@
...
+        columnDefs: columns,
+        enableSorting: true,
+        showColumnFooter: true,
+        fastWatch: true,
+        flatEntityAccess: true,
+      };
+
+      vm.gridOptions.multiSelect = false;
+
+      vm.gridOptions.onRegisterApi = function (gridApi) {
```
jniles, line 97: Nice :+1:

client/src/modules/email-report/modal/categories-people-list-modal.html
```diff
@@ -0,0 +1,50 @@
+<!-- billing service form -->
+<form
+    name="BillingServicesForm"
+    bh-submit="BillingServicesFormCtrl.submit(BillingServicesForm)"
+    novalidate>
+
+    <div class="modal-header">
+        <ol class="headercrumb">
+            <li class="static">
+                {{"FORM.LABELS.PROFILE_FROM" | translate}}
```
jniles: Please use the `translate` directive in this HTML file.

client/src/modules/email-report/modal/categories-people-list-modal.html
```diff
@@ -0,0 +1,50 @@
...
+    <div class="modal-header">
+        <ol class="headercrumb">
+            <li class="static">
+                {{"FORM.LABELS.PROFILE_FROM" | translate}} 
+                {{catPeopleListCtrl.selectedCategory.table_display | translate}}
+            </li>
+        </ol>
+    </div>
+
+    <div class="modal-body" style="height: 100%;border: none !important">
```
jniles: Why do you need to put `height:100%`?  Do these `style` tags actually do anything?

client/src/modules/email-report/modal/categories-people-list-modal.html
```diff
@@ -0,0 +1,50 @@
...
+        </div>
+
+
+    </div>
+
+    <div class="modal-footer">
+        <a class="btn btn-default"
+            data-method="selectProfile"
+            ui-sref="emailreport"
+            data-method="back"
```
jniles: Since the button is labeled "CLOSE", this `data-method` should probably be `data-method="close"` or `data-method="cancel"`.

client/src/modules/email-report/modal/categories-people-list-modal.html
```diff
@@ -0,0 +1,50 @@
...
+            translate>
+            FORM.BUTTONS.CLOSE
+        </a>
+
+
+        <bh-loading-button
+            ng-click="catPeopleListCtrl.getSelectedRows()"
```
jniles: In general, `bhLoadingButton`s are used for firing `submit` methods.  If this doesn't submit anything, just use a basic button.

client/src/modules/email-report/modal/categories-people-list-service.js
```diff
@@ -0,0 +1,41 @@
...
+    Documents, Visits, Filters) {
+
+    var service = this;
+    service.list = [];
+
+    init();
```
jniles: There is no need for this `init` method.  We use them in our controllers to group the startup functions together (since a controller is initialized each page navigation in angular).

[AngularJS services are singletons](https://stackoverflow.com/questions/32677694/angular-js-services-are-singleton-objects-means), which means they are only created once.  Since you aren't doing any complex logic, just bind everything to the service directly, like this:
```js
var service = this;

service.list = [{
   table: 'employee',
   // etc ...
}];

return service;
```

client/src/modules/email-report/modal/categories-people-list-service.js
```diff
@@ -0,0 +1,41 @@
...
+];
+
+/**
+ * @module EmailReportService
+ *
+ * This service is responsible for providing an interface between angular
+ * module controllers and the server /tools/bed/ API.
+ */
+function CategoriesPeopleService($http, util, Session, $uibModal,
+    Documents, Visits, Filters) {
```
jniles: Most of these imports you never use!  Remove all unused imports.

client/src/modules/email-report/modal/liste-categories-people-ctrl.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Report group Controller
+ *
+ * @description
+  The controller that collect every definied group of people 
+  where the user can select a profile
+ */
+
+function CategoriesPeopleModalController(CategoriesPeople, Session, util, Notify, ScrollTo, bhConstants, uiGridConstants) {
```
jniles: The comment says this is `Report Group Controller`, but the name says `CategoriesPeopleModalController`.  What is it?

client/src/modules/email-report/modal/liste-categories-people-modal.html
```diff
@@ -0,0 +1,40 @@
+<!-- billing service form -->
+<form
+    name="BillingServicesForm"
```
jniles: Do you actually use this?   If so, please name it correctly...

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
+<div class="flex-header static">
+    <div class="bhima-title">
+        <ol class="headercrumb">
+            <li class="static">{{ "TREE.ADMIN" | translate }}</li>
```
jniles: Please use the `translate` directive.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+        <div class="col-lg-12">
+            <!-- start new Reporting group form -->
+            <div class="col-lg-5">
+                <h4 translate=""> FORM.LABELS.NEW_REPORT_GROUP</h4>
+
+
+                <form 
+                    name="ReportGroupRegistrationForm"
+                    bh-submit="ReportGroupCtrl.save(ReportGroupRegistrationForm)"
+                    bh-form-defaults
```
jniles: Remove all references to `bh-form-defaults`.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+
+                <div 
+                    id="report-groups-grid"
+                    ui-grid-auto-resize
+                    ui-grid-resize-columns
+                    ui-grid-move-columns
+                    ui-grid-cellNav
+                    ui-grid-grouping
+                    ui-grid-exporter
+                    ui-grid-save-state
```
jniles: Please make sure you only use the plugins you need - anything extra will slow the page down.

client/src/modules/report-group/reportGroup.service.js
```diff
@@ -0,0 +1,69 @@
...
+    'DocumentService', 'VisitService', 'DepricatedFilterService',
+];
+
+/**
+ * @module ReportGroupService
+ *
+ * This service is responsible for providing an interface between angular
+ * module controllers and the server /tools/bed/ API.
+ */
+function ReportGroupService($http, util, Session, $uibModal,
```
jniles: Please remove all imports that you do not need.

server/config/routes.js
```diff
@@ -84,10 +84,39 @@ const financialPatient = require('../controllers/finance/patient');
...
+const weeklySummaryReport=require('../controllers/email-report/weeklySummaryReport');
+const mail=require('../controllers/report-group/mail');
+const scheduler=require('../controllers/report-group/scheduler');
 // expose routes to the server.
 exports.configure = function configure(app) {
   winston.debug('Configuring routes');
 
+
+//this is the scheduler config for send reports emailing 
+scheduler.setScheduler();
```
jniles: This should be in the `server/config/express.js` file.  The `server/config/routes.js` file should only contain routes.

server/controllers/email-report/index.js
```diff
@@ -0,0 +1,189 @@
...
+const db = require('../../lib/db');
+const Topic = require('../../lib/topic');
+
+const BadRequest = require('../../lib/errors/BadRequest');
+const NotFound = require('../../lib/errors/NotFound');
+const Period = require('../../lib/period');
+
+exports.create = create;
+exports.list = list;
+exports.list_people = list_people;
```
jniles: Please use `camelCase` instead of `underscore_case`.  This should be called `listPeople`.

server/controllers/email-report/weeklySummaryReport.handlebars
```diff
@@ -0,0 +1,118 @@
+{{> head title="PATIENT_RECORDS.FINANCIAL_ACTIVITY.REPORT_TITLE" }}
+
+<body>
+    <style type="text/css">
```
jniles, line 4: Please do not define new styles in HTML.  If you want to change styles, do it in `bhima-bootstrap.less`.

server/controllers/email-report/weeklySummaryReport.js
```diff
@@ -0,0 +1,217 @@
...
+
+    //services incomes
+    return loadServicesIncome().then((_servicesIncome) => {
+
+      //max and min date of invoice each day for a week
+      return MaxMinInvoiceDate().then(_MaxMinInvoiceDate => {
+        //rendering the report
+        var report = new ReportManager(TEMPLATE, currentSession, options);
+
+        if (_patientsData.lenght === 0) {
```
jniles: This is a syntax error - it should say `length`.

server/controllers/email-report/weeklySummaryReport.js
```diff
@@ -0,0 +1,217 @@
...
+  //patient registered
+  return getRegisteredPatientsNumber().then( (_patientsData) => {
+
+    //services incomes
+    return loadServicesIncome().then((_servicesIncome) => {
+
+      //max and min date of invoice each day for a week
+      return MaxMinInvoiceDate().then(_MaxMinInvoiceDate => {
+        //rendering the report
+        var report = new ReportManager(TEMPLATE, currentSession, options);
```
jniles: This should be in a `try/catch` block at the top of the function like all other reports.  It makes sure we don't make any database requests if the user has sent a malformed HTTP request.

server/controllers/email-report/weeklySummaryReport.js
```diff
@@ -0,0 +1,217 @@
...
+
+  const options = {
+    renderer: 'pdf',
+    saveReport: false,
+    filename: 'test_summary_report.pdf',
+    orientation: 'portrait',
+  };
+
+  //patient registered
+  return getRegisteredPatientsNumber().then( (_patientsData) => {
```
jniles: This should be a promise chain to avoid making a giant tree:
```js
const data = {};
return getRegisteredPatientsNumber()
  .then(patientsData => {
     data.patientsData = patientsData;
     return loadServicesIncome();
  .then(servicesIncome => {
     data.servicesIncome = servicesIncome;
     return MaxMinInvoiceDate();
  })
  .then(maxMinInvoiceDate => {
    data.maxMinInvoiceDate = maxMinInvoiceDate;
    return report.render(data);
   });
```

jniles (changes requested):
The code is slowly improving!

There are a lot of non-standard issues that need to be cleaned up before the logic can be addressed.

 1. Make sure you are always using the `translate` directive; never use the translate filter.
 2. Make sure you never `$inject` or `require()` something and then not use it.  Everything you import _must_ be used in the module.  Eslint should catch this on the server.
 3. Remove all `class="ng-scope"` and `class="ng-binding"`.  These do not do anything for programmers and will only break debugging tools like batarang.
 4. Remove all instances of `bh-form-defaults`. 

Once these are done, please leave a comment and @sfount or I will give you another review.

.env.development
jniles: So, it turns out that these should be strings, or else bash will attempt to execute them as a function.  This breaks the integration tests on Travis.

Please make this a string:
```sh
MAILGUN_API_KEY="key-c30213afb78d0d0e53a0f973c3db7021"
```

server/controllers/email-report/index.js
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * @method list_people
+ *
+ * @description
+ * API /email-report/list-people
+ */
+function listPeople(req, res, next) {
+
+  const sqlParams = req.params;
+  const sql = `SELECT ${sqlParams.column1} , ${sqlParams.column2} FROM ${sqlParams.table}`;
```
jniles, line 130: :frowning_face: This is a really bad idea.  This API can be abused to reveal sensitive information.   For example, I was able to list all the user names and passwords in the system.

emailreportsqlinjectionattack
_Fig 1: Abusing the email-report API for evil purposes_

The solution to this is removing the client's ability to specify the columns and table in the database.  This essentially gives them super user privileges.  Since our code is open-source, anyone can inspect it and discover this bug, then use it to attack an instutition.

Another problem with this code is that the values are not escaped.  This means that it is vulnerable to an [SQL Injection (SQLi) attack](https://www.acunetix.com/websitesecurity/sql-injection/).  In general, you should always pass your values into the database library like this:
```js
db.exec(sql, [ parameterA, parameterB /*, ... */ ])
```
... to prevent SQL injection attacks.  The parameters in the array will be escaped using [mysqljs's query escape features](https://github.com/mysqljs/mysql#escaping-query-values).

In general, you should never trust the client-side code - anyone could write a separate client for BHIMA and use this API to pull data they are not supposed to have access to.  Your server code should always try to be attack-proof.

jniles: My proposed solution is to either:
 1. Make a URL for each table/parameter you need.

OR

 2. Use a query string to switch between SQL statements.  For example the code would like something like this (pseudo-js-code):
  ```js
function listPeople(req, res, next) {
    const tableASql = 'SELECT columnA, columnB from tableA;'
    const tableBSql = 'SELECT columnX, columnY from tableB;'

    let query;

    // when the client send a GET with query string ?tableB=1, this will be triggered
    if (req.query.tableB) {
       query = tableBSql;
    } else {
       query = tableASql;
    }

    return db.exec(query)
       // etc...
}

server/controllers/email-report/index.js
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * @method update
+ *
+ * @description
+
+ * POST /email-report/:id
+ */
+function update(req, res, next) {
+
+  const sql = 'UPDATE email_report SET name=?, email=?,frequency=?,code_report_group=? WHERE id=?';
```
jniles, line 75: [This comment by @sfount](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1778/files#r126668627) will help clean this code up greatly.

The mysqljs library allows us to use `SET ?` to specify the columns and values.  [From the documentation](https://github.com/mysqljs/mysql#escaping-query-values):
```js
var post  = {id: 1, title: 'Hello MySQL'};
var query = connection.query('INSERT INTO posts SET ?', post, function (error, results, fields) {
  if (error) throw error;
  // Neat!
});
console.log(query.sql); // INSERT INTO posts SET `id` = 1, `title` = 'Hello MySQL'
```

Using db.js, this means your query can look like this:
```js
const sql = 'UPDATE email_report SET ? WHERE id = ?;'

// make sure we do not update the ID
const emailReport = req.body.emailReport;
delete req.body.emailReport.id;

// execute the query
db.exec(sql, [emailReport, req.params.id]);
// etc ...
```

This makes the code much more compact and easier for a developer to read.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,162 @@
...
+        <div class="col-md-12">
+            <!-- start new Reporting group form -->
+            <div class="col-md-5">
+                <h4 translate=""> FORM.LABELS.NEW_REPORT_GROUP</h4>
+
+
+                <form
+                    name="ReportGroupRegistrationForm"
+                    bh-submit="ReportGroupCtrl.save(ReportGroupRegistrationForm)"
+                    bh-form-defaults
```
jniles, line 38: Please do not use `bh-form-defaults`.  It is no longer used in BHIMA.  I recommend looking for all files and removing it.

[This has been previously mentioned in a review](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1778#discussion_r130862026).

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,162 @@
...
+                                  <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+                            </div>
+                        </div>
+                    </div>
+
+
+                    <bh-loading-button
+                        ng-show="ReportGroupCtrl.selectedReportGroup.selected == false"
+                        class="pull-right" loading-state="BedRegistrationForm.$loading">
+                          {{ "FORM.BUTTONS.SAVE" | translate }}
```
jniles: Please use the `translate` directive.  [This has been previously mentioned in a review](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1778#discussion_r130861928).

server/controllers/email-report/weeklySummaryReport.handlebars
jniles: This still needs to be addressed.

client/src/modules/email-report/modal/categories-people-list-modal.html
```diff
@@ -0,0 +1,52 @@
+<!-- billing service form -->
+<form
+    name="BillingServicesForm"
```
jniles, line 3: This form does not contain billing services.  Please rename this form to the proper name or remove the code if you do not need it.

[This has been previously mentioned in a review](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1778#discussion_r130861839).

client/src/modules/email-report/modal/categories-people-list-ctrl.js
```diff
@@ -0,0 +1,124 @@
...
+  vm.listCategoriesPeople = CategoriesPeople.list;
+  vm.selectedCategory = {};
+
+  vm.loading = false;
+  vm.hasError = false;
+
+  function load() {
+
+    // we will use cash to save the current category index
+    const cash = AppCache('current_category');
```
jniles, line 37: This is much better!  :+1:

A better name for this is `cache`.  "Cash" means "money" in English.  "Cache" means "container".

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,211 @@
...
+
+              <div class="input-group col-md-12">
+
+                <select
+                  required=""
+                  name="frequency"
+                  ng-model="EmailReportCtrl.emailReport.frequency"
+                  ng-options="item.key as (item.label | translate) for item in EmailReportCtrl.frequencies"
+                  class="form-control">
+                    <option>
```
jniles, line 113: This option should be translated.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,211 @@
+<div class="flex-header static">
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static" translate> TREE.ADMIN</li>
+      <li class="static" translate> FORM.LABELS.EMALING_REPORT</li>
```
jniles, line 5: This class should be "title" not "static".

client/src/modules/email-report/email-report.routes.js
```diff
@@ -0,0 +1,81 @@
+angular.module('bhima.routes')
+  .config(['$stateProvider', function ($stateProvider) {
+
+    $stateProvider
+      .state('emailreport', {
+        url: '/email-report/',
+        controller: 'EmailReportController as EmailReportCtrl',
+        templateUrl: 'modules/email-report/index.html',
+        params: {
+          data: null,
```
jniles: If you change this line to:
```js
params : {
  data : {},
}
```

This will make data default to an object if it is not defined.  Then you will not need all the `try/catch` statements in your controllers.

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,230 @@
...
+
+   */
+
+
+  function init() {
+    /* loading frequencies*/
+    vm.frequencies = EmailReportSvc.frequencies;
+
+    /* If the user has selected a profile, fill name and email field */
+    try {
```
jniles: If you default `data` to an empty object, you will not need this `try/catch` block.

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,230 @@
...
+    if (isCreateReport) {
+      operation = EmailReportSvc.create(vm.emailReport);
+    } else {
+      operation = EmailReportSvc.update(vm.emailReport);
+    }
+
+    operation.then(function (confirmation) {
+
+      if (confirmation) {
+        alert('operation successeded');
```
jniles: Please do not use `alert()` on the client.  A better option is to use [ModalService](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/ModalService.js#L35)'s alert() method instead.

[This was previously mentioned in a review](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1778#discussion_r130851350).

client/src/modules/email-report/email-report.ctrl.js
```diff
@@ -0,0 +1,230 @@
...
+    // end propagation for invalid state - this could scroll to an $invalid element on the form
+    if (RegistrationForm.$invalid) {
+      return Notify.danger('FORM.ERRORS.INVALID');
+    }
+
+    const isCreateReport = vm.selectedEmailReport.selected === false;
+
+    let operation;
+
+    if (isCreateReport) {
```
jniles: Much cleaner!

Could you used the other places where you use this pattern?  [Like here](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1778/files#diff-615fa2cf70ad0d290df4a2590b8c2fd7R56)

client/src/modules/email-report/email-report.routes.js
```diff
@@ -0,0 +1,81 @@
...
+  var ctrl = 'CategoriesPeopleModalController as catPeopleModalListCtrl';
+
+  return function onEnter($state, Modal, Notify) {
+    Modal.open({
+      templateUrl: 'modules/email-report/modal/liste-categories-people-modal.html',
+      controller: ctrl,
+      backdrop: 'static',
+      keyboard: false,
+    }).result
+      .then(function () {
```
jniles: If you don't use the `.then()` block, you don't need it!  Just delete it.

client/src/modules/email-report/index.html
```diff
@@ -0,0 +1,211 @@
...
+                ng-show="EmailReportCtrl.selectedEmailReport.selected"
+                translate
+                >
+                FORM.BUTTONS.DELETE
+              </span>
+
+              <bh-loading-button
+                ng-show="EmailReportCtrl.selectedEmailReport.selected"
+                loading-state="BedRegistrationForm.$loading">
+                  {{ "FORM.BUTTONS.UPDATE" | translate }}
```
jniles: Remember to use the `translate` directive.

jniles (commented):
A couple more comments to clean up -- there were a few things from the last review that were not addressed and should be fixed before we focus on the functionality.

client/src/modules/email-report/index.html
jeremielodi: It gives an error when I use it with bh-loading-button

jniles: Simple - just add an extra `<span>`.
```html
<bh-loading-button
  ng-show="EmailReportCtrl.selectedEmailReport.selected"
  loading-state="BedRegistrationForm.$loading">
   <span translate>FORM.BUTTONS.UPDATE</span>
</bh-loading-button>
```

jniles:
@jeremielodi, there are still some issues here to be addressed.  Let me know if you need clarification.

jniles:
This PR is really out of date and has a few design flaws.   It will likely never get merged into master.  We can break these components up into smaller pieces and merge them as needed.  This PR was useful practice though!

PR #1782, opened 2017-06-09
Test development data

sfount:
This commit introduces development data that spans two fiscal years and
has correctly configured accounts as well as randomly generated patients
and transactions. The development data set should be used to verify the
output of financial reports.

Replaces #1779 and #1780 to resolve CI issues.


sfount:
Merging... Circle CI integration was broken so this failure was ignored.

PR #1787, opened 2017-06-13
feature(bhDebtorGroupSelect) 

### LINKED ISSUE #1741: feature request: implement bhDebtorGroupSelect
We use debtor groups a lot (since we are a financial system).  It's almost as ubiquitous as an account select.

For this reason, I propose we create a `bhDebtorGroupSelect` component that has an API similar to the [bhAccountSelect](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/components/bhAccountSelect.js).  It would allow us to reduce the number of lines of code in most controllers/templates and leave only the logic that is needed for using that debtor group.  It would also give us another component to test that is currently not easily testable.

lomamech:
feature(bhDebtorGroupSelect)  …
- Implement component bhDebtorGroupSelect
- Using the bhDebtorGroupSelect in Modules: Patients, Employees, Invoice
  Registry, Cash registry,
- Update test End to End

closes #1741


client/src/modules/cash/payments/templates/search.modal.html
```diff
@@ -53,21 +53,11 @@
...
-            <ui-select
-              name="debtor_group_uuid"
-              ng-model="$ctrl.searchQueries.debtor_group_uuid">
-              <ui-select-match placeholder="{{ 'FORM.PLACEHOLDERS.DEBTOR_GROUP' | translate }}"><span>{{$select.selected.name}}</span></ui-select-match>
-              <ui-select-choices ui-select-focus-patch repeat="group.uuid as group in ($ctrl.debtorGroups | filter:$select.search | orderBy:'name') track by group.uuid">
-                <span ng-bind-html="group.name | highlight:$select.search"></span>
-              </ui-select-choices>
-            </ui-select>
-          </div>
+          <bh-debtor-group-select
```
sfount, line 56: :+1: This is a huge improvement for maintenance.

client/src/js/components/bhDebtorGroupSelect.js
```diff
@@ -0,0 +1,51 @@
+angular.module('bhima.components')
+  .component('bhDebtorGroupSelect', {
+    templateUrl : 'modules/templates/bhDebtorGroupSelect.tmpl.html',
+    controller  : DebtorGroupSelectController,
+    // transclude  : true,
```
sfount: I think it is necessary to allow this component to `transclude` to allow you to use the `bh-clear` component inside (in search modals for example).

client/src/modules/cash/payments/templates/search.modal.html
```diff
@@ -53,21 +53,11 @@
...
+            debtor-group-uuid = "$ctrl.searchQueries.debtor_group_uuid"
+            on-select-callback = "$ctrl.onSelectDebtor(debtorGroup)">
+			<bh-clear on-clear="$ctrl.clear('debtor_group_uuid')"></bh-clear>
```
sfount: This inner component (`bh-clear`) is not currently displayed. I think you have to allow the component (`bh-debtor-group-select`) to `transclude`

sfount (commented):
This is a great refactor! I think there is a problem with using `transclude` on the component that should be looked at.

client/src/js/components/bhDebtorGroupSelect.js
```diff
@@ -0,0 +1,51 @@
+angular.module('bhima.components')
+  .component('bhDebtorGroupSelect', {
+    templateUrl : 'modules/templates/bhDebtorGroupSelect.tmpl.html',
+    controller  : DebtorGroupSelectController,
+    transclude  : true,
+    bindings    : {
+      debtorGroupUuid   : '<',
+      disable           : '<?',
```
jniles, line 8: Does disable do anything?

client/src/js/components/bhDebtorGroupSelect.js
```diff
@@ -0,0 +1,51 @@
...
+      onSelectCallback  : '&',
+      name              : '@?',
+      label             : '@?',
+      required          : '<?',
```
jniles, line 12: I don't think required does anything here.

client/src/modules/employees/registration/employees.js
```diff
@@ -4,11 +4,11 @@ angular.module('bhima.controllers')
...
 EmployeeController.$inject = [
   'EmployeeService', 'ServiceService', 'GradeService', 'FunctionService',
-  'CreditorGroupService', 'DebtorGroupService', 'util', 'NotifyService',
-  'bhConstants', 'ReceiptModal', 'SessionService'
+  'CreditorGroupService', 'util', 'NotifyService',
+  'bhConstants', 'ReceiptModal', 'SessionService',
 ];
 
-function EmployeeController(Employees, Services, Grades, Functions, CreditorGroups, DebtorGroups, util, Notify, bhConstants, Receipts, Session) {
+function EmployeeController(Employees, Services, Grades, Functions, CreditorGroups, util, Notify, bhConstants, Receipts, Session,) {
```
jniles, line 11: ... are you sure this works?  It looks like there should be a syntax error - the trailing comma is allowed on arrays, but not functions.

client/src/modules/templates/bhDebtorGroupSelect.tmpl.html
```diff
@@ -0,0 +1,22 @@
...
+    ng-class="{ 'has-error' : $ctrl.validationTrigger && DebtorForm.debtor_group_uuid.$invalid }">
+    <label class="control-label" translate>
+      {{ $ctrl.label }}
+    </label>
+    <ng-transclude></ng-transclude>
+    <ui-select name="debtor_group_uuid" 
+      ng-model="$ctrl.debtorGroupUuid"
+      on-select="$ctrl.onSelect($item, $model)" required>
+      <ui-select-match placeholder="{{ 'FORM.SELECT.DEBTOR_GROUP' | translate }}">{{$select.selected.name}}</span></ui-select-match>
+      <ui-select-choices ui-select-focus-patch repeat="group.uuid as group in ($ctrl.debtorGroups | filter:$select.search | orderBy:'name') track by group.uuid">
```
jniles, line 13: when Instead of using `| orderBy: 'name'`, it would be better to simply do the sort in the controller when you download the data - that way Angular will not have to recompute the sort on every change in the view.

jniles (approved):
@lomamech this is a good implementation.

I think there are a few pieces of code that are copied from other modules that do not belong here.  They are not blocking though.  If tests pass, LGTM :+1:

PR #1789, opened 2017-06-13
refactor(inventory registry): use filter, export and save grid state

### LINKED ISSUE #1747: refactor: Improvements to the Inventory Registry
This issue is to track improvements to the Inventory Registry to be tackled while the team is working all registries.

Parent Issue #1737.

-------

The follow changes should be made to the Inventory Registry:
 - [ ] Make a "Menu" dropdown with the following options:
   1. Column Configuration
   2. Saving the Grid Configuration
   3. Resetting the Grid Configuration
   4. Download View as CSV
   5. Download View as PDF
 - [ ] Implement the Grid State work on the grid (See #1738)
 - [ ] Implement the latest FilterService, replacing the old `bhFiltersApplied` with the new `FilterService`
 - [ ] Implement downloading as CSV using the `ui-grid` exporter (See https://github.com/IMA-WorldHealth/bhima-2.X/pull/1554)
 - [ ] Make sure the search form is up to the latest standards (see #1702)
 - [ ] Sort by both the code and group by default

**Note**: There is already an issue concerning the Inventory Registry in https://github.com/IMA-WorldHealth/bhima-2.X/issues/1721

mbayopanda:
This PR refactor the inventory registry accoriding the new standard for search filters, data exportation, saving grid state and update the UI.

Closes #1747


mbayopanda:

sfount (commented):
This looks like a great refactor. The creation of `SearchFilterSerivce` is very interesting - potentially this would be a good way to remove a lot of the duplication with code that is written to cache filters etc. 

I have also tested the functionality of this PR and I think there are one or two things that can be addressed before this is merged: 
1. When the search modal is opened it does not remember the current filters, I believe the expected behaviour is that filters that are already assigned should be filled in
2. The 'Remove all filters' link has been removed on other grids, I don't think it is needed here. 
3. Exporting to PDF does not apply the filters but instead exports everything - I think it is quite a small change to pass the filters object to the PDF render to limit the results.

client/src/modules/inventory/list/list.html
```diff
@@ -8,6 +8,47 @@
...
+
+            <li role="menuitem">
+              <a href data-method="clear-state" ng-click="InventoryCtrl.clearGridState()">
+                <i class="fa fa-close"></i> <span translate>FORM.BUTTONS.CLEAR_GRID_CONFIGURATION</span>
+              </a>
+            </li>
+
+            <li role="separator" class="divider"></li>
+            <li role="menuitem">
+              <a ng-href="/reports/inventory/items/?{{ InventoryCtrl.download('pdf') }}" download="{{ 'CASH_REGISTRY.TITLE' | translate }}">
```
sfount: This should be updated to use the Inventory registry TITLE instead of cash.

jniles:
@mbayopanda it looks like point 2 of @sfount's remarks have been addressed.  As soon as the other two are addressed, this can be merged.  Code LGTM. :+1:

mbayopanda:
@jniles could you review this PR ?

jniles (approved):
LGTM.

There is a minor bug that exists here, but I don't think it is blocking.  Basically, if you search for a label, then remove the label filter in the list view, it is still cached on the form.  That's slightly annoying and different from other modules.  However, the changes here outweigh that bug.  I'll make an issue about it.

PR #1790, opened 2017-06-13
fix(reports): Ensure period 0 is considered in balance calculations

sfount:
This commit ensures that period 0 is factored into opening balance
calculations. Currently as no start or end date is provided for period 0
it is ignored in opening balance calculations. This adds a general
exception for period 0.


mbayopanda (approved):
LGTM

PR #1796, opened 2017-06-14
fix(reports): Use correct end period comparison for account balance

sfount:
This commit ensures that values the current period are not included in
the period total calculation for determining an accounts opening
balance. This is essential as transactions from the current period (up
to the current date) are then summed and added to the total value. If
the last date in a period is selected this would then result in counting
the current period twice.


sfount:
Note that this has passed test - an issue with branches on the `ima` repository caused a travis test to permanently error here. (reference: https://travis-ci.org/IMA-WorldHealth/bhima-2.X/builds/242848514)

PR #1797, opened 2017-06-15
refactor(balance report): update the balance report

### LINKED ISSUE #1293: (bug) Balance Sheet date ranges do not make sense
The current balance sheet report doesn't really make sense.  Here are some confusing issues:
 1. I can select a report range, yet the report will give me opening balances and the movements for a single month.
 2. Account classes do not necessarily have to exist ... they only exist in OHADA.  We should either allow a balance sheet for a particular set of title accounts or simply give the entire balance sheet.
 3. Also, I think the account class filtering doesn't work.

Here is a balance sheet produced by BHIMA with the "this year" option set:
buggybalancesheet
_Fig 1: Buggy Balance Sheet Report_

I recommend that we remove the option to select only particular account classes and propose a better solution.  It's broken anyway.

I also recommend that we only allow you to select periods instead of dates.  These would come from the period table and would select the current period by default.

### LINKED ISSUE #1791: (refactor) Balance sheet report
Re: the new report client infrastructure merged in #1672.

Step 1.
This issue tracks updating the balance report to work with the new infrastructure. The client module should be very similar but updated to use the preview/ save architecture.

Step 2.
Write tests to verify the outcome of the report given the new development database

Step 3.
Update the report to the latest standards defined by the SCRUM Master (TODO)

mbayopanda:
This PR refactor the balance report according the new standard of reports.

Closes #1791   Closes #1293.


mbayopanda:

jniles:
@mbayopanda this has a lot of merge conflicts.  Can you rebase to master?

mbayopanda:
balance report

client/src/modules/reports/generate/balance_report/balance_report.config.js
```diff
@@ -0,0 +1,75 @@
...
+
+  vm.requestSaveAs = function requestSaveAs() {
+    var options = {
+      url : reportUrl,
+      report : reportData,
+      reportOptions : angular.copy(vm.reportDetails),
+    };
+
+    return SavedReports.saveAsModal(options)
+      .then(function callBack() {
```
jniles: This function doesn't need a name :)

client/src/modules/reports/generate/balance_report/balance_report.html
```diff
@@ -0,0 +1,43 @@
...
+        <div class="panel-heading">
+          <span translate>REPORT.UTIL.OPTIONS</span>
+        </div>
+
+        <div class="panel-body">
+
+        <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate autocomplete="off">
+
+          <!--select the period by a date-->
+          <bh-date-editor
```
jniles, line 28: I feel like a period selection ("today", "this week", "this month", "this year") would be useful as well to quickly set the date to the beginning of this month (for example).

Additionally, this allows you to set future dates - I'm not sure if you wanted that.  If not, use the `maxDate` directive.

server/controllers/finance/reports/balance/index.js
```diff
@@ -34,10 +34,10 @@ function document(req, res, next) {
 
   // date option
   if (params.dateFrom && params.dateTo) {
-    session.dateFrom = new Date(params.dateFrom);
-    session.dateTo = new Date(params.dateTo);
+    session.dateFrom = moment(params.dateFrom).format('YYYY-MM-DD');
```
jniles, line 37: Are you sure this was necessary?  I thought `node-mysql` parsed these for us ... has something changed?

server/controllers/finance/reports/balance/index.js
```diff
@@ -172,76 +173,52 @@ function balanceReporting(params) {
...
-    const accounts = rows;
-    const touched = data.beginning.map(item => item.id);
-
-    accounts.forEach(item => {
-      if (touched.indexOf(item.id) === -1) {
-        data.beginning.push(item);
+  return db.one(sqlFiscalYear, [query.date])
+    .then((fiscalYear) => {
+      if (!fiscalYear.id) {
+        throw new NotFound(`Could not find a Fiscal Year for the date ${query.date}.`);
```
jniles: I believe that `db.one()` makes this error unnecessary.  If no record is found, it will throw and error.

server/models/procedures.sql
```diff
@@ -1304,4 +1304,10 @@ BEGIN
   SELECT DISTINCT BUID(record_uuid) AS record_uuid, trans_id, code FROM stage_trial_balance_errors ORDER BY code, trans_id;
 END $$
 
-DELIMITER ;
+-- compute account class
+CREATE PROCEDURE ComputeAccountClass()
```
jniles: Is this procedure ever called from the code?   If not, it probably should be in `admin.sql`.

jniles (commented):
@mbayopanda I've made a few comments below.

I'm getting a "Show Exploitation" checkbox on my machine that I cannot find in the github diff.  I'll attach a screenshot below, but do you know what that is?  It doesn't seem to be formatted properly.

Run this by @WayneNiles before it is merged.  The code is fine.

jniles:
showexploitationcheckbox
_Fig 1: Show Exploitation Checkbox_

client/src/modules/reports/generate/balance_report/balance_report.html
mbayopanda: you're right, may be we have to update the `bhDateEditor` component.

client/src/modules/reports/generate/balance_report/balance_report.config.js
mbayopanda: there is an `eslint` rule which doesn't allow function without name...

jniles: Aha - I thought it was turned of in our repository...

client/src/modules/reports/generate/balance_report/balance_report.html
jniles: You should be able to simply add it.  The cash module uses it.  [Here's the line](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/cash/cash.html#L64).

mbayopanda:
About the picture `Fig 1`, the Balance Sheet report is a kind of Bilan (in the French context) it means it doesn't contains income and expense accounts, the option `show exploitation` is there for showing the exploitation's result for being sure that the equation `ASSETS = LIABILITIES + CAPITAL` is verified.

For the Balance Report, all touched accounts must be there (it is not an option)

In other words `Balance Sheet Statement` is different to `Balance Report`

mbayopanda:
@jniles this report has Mr Wayne's authorization

jniles:
@mbayopanda, excellent!  I'll make sure there are no last minute things to do.

@IMA-WorldHealth/bhima-core, if there are any other reports that have approval, please comment on them.

PR #1807, opened 2017-06-20
refactor(Cash flow report)

### LINKED ISSUE #1805: (Refactor) cashflow Report
Re: the new report client infrastructure merged in #1672.

Step 1.
This issue tracks updating the aged debtor report to work with the new infrastructure. The client module should be very similar but updated to use the preview/ save architecture.

Step 2.
Write tests to verify the outcome of the report given the new development database

Step 3.
Update the report to the latest standards (TODO)
(See #1784 for an example of some of the report requirements)

lomamech:
- Implement Cashflow Report
- feat(bhCashboxSelect) implement the componant Cashbox Select
- Implement the test E2E for Cashflow Report

closes #1805


client/src/js/components/bhCashboxSelect.js
```diff
@@ -0,0 +1,55 @@
...
+
+    // load all Cashbox
+    Cashbox.read(null, { detailed: 1 })
+      .then(function (cashboxes) {
+        cashboxes.forEach(function (cashbox) {
+          cashbox.hrlabel = cashbox.label + ' ' + cashbox.symbol;
+        });
+        
+        $ctrl.cashboxes = cashboxes;
+      });
```
jniles: This should probably have a `.catch()` statement.

client/src/modules/templates/bhCashboxSelect.tmpl.html
```diff
@@ -0,0 +1,21 @@
...
+  <div
+    class="form-group"
+    ng-class="{ 'has-error' : $ctrl.validationTrigger && CashboxForm.cashbox_id.$invalid }">
+    <label class="control-label" translate>
+      {{ $ctrl.label }}
+    </label>
+    <ui-select name="cashbox_id" 
+      ng-model="$ctrl.cashboxId"
+      on-select="$ctrl.onSelect($item, $model)" required>
+      <ui-select-match placeholder="{{ 'FORM.SELECT.CASHBOX' | translate }}">{{$select.selected.hrlabel}}</span></ui-select-match>
```
jniles: This looks like you have a broken `</span>`.  You should remove it.

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -66,11 +69,12 @@ function processingCashflowReport(params) {
         throw new BadRequest('Periods not found due to a bad date interval', 'ERRORS.BAD_DATE_INTERVAL');
       }
       glb.periods = periods;
-      return closingBalance(params.account_id, glb.periods[0].start_date);
+      return AccountsExtra.getOpeningBalanceForDate(params.account_id, glb.periods[0].start_date);
     })
     .then((balance) => {
-      if (!balance.length) { balance[0] = { balance : 0, account_id : params.account_id }; }
-      glb.openningBalance = balance[0];
+      if (!balance) { balance = { balance : 0, account_id : params.account_id }; }
```
jniles: `AccountsExtra.getOpeningBalanceForDate()` will return an object.  See [this line](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/accounts/extra.js#L152).  That line does not contain a `account_id`.

Is there ever a case where `if(!balance) {}` is triggered?  Do you ever use the `account_id`?

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -99,51 +103,33 @@ function queryIncomeExpense(params, dateFrom, dateTo) {
...
+      t.debit, t.credit, t.currency_id, t.description, t.comment,
+      BUID(t.record_uuid) AS record_uuid, t.origin_id, u.display_name,
+      x.text AS transactionType
+    FROM (        
+      SELECT gl.project_id, gl.uuid, gl.record_uuid, gl.trans_date, gl.period_id,
+        gl.debit_equiv, gl.credit_equiv, gl.debit, gl.credit,
+        gl.account_id, gl.entity_uuid, gl.currency_id, gl.trans_id,
+        gl.description, gl.comment, gl.origin_id, gl.user_id
+      FROM general_ledger gl
+      WHERE gl.account_id IN (?) AND
```
jniles: It would be more efficient to say `account_id = ?`.  It doesn't look like you use multiple accounts.

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -223,15 +216,21 @@ function processingWeekCashflow(params) {
...
-    });
+  return AccountsExtra.getOpeningBalanceForDate(params.account_id, glb.periods[0].start_date)
+  .then((balance) => {
+    if (!balance) { balance = { balance : 0, account_id : params.account_id }; }
+    glb.openningBalance = balance;
+
+    // get all periods for the the current fiscal year
+    return queryIncomeExpense(params);
+  })
+  .then((result) => {
```
jniles: Just FYI, the following two are equivalent:
```js
.then((result) => {
   return groupByPeriod(glb.periods, result, params.weekly);
});

// ... is the same as ...

.then(result =>  groupByPeriod(glb.periods, result, params.weekly));
```

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -553,11 +551,16 @@ function document(req, res, next) {
     if (incomes) {
       incomes.forEach((item) => {
         tempIncome[item.origin_id] = typeof tempIncome[item.origin_id] !== 'undefined';
+        // FIX ME
+        // typeof tempIncome[item.origin_id] !== 'undefined' It's TRUE
+        // But this affectation " tempIncome[item.origin_id] = typeof tempIncome[item.origin_id] !== 'undefined' " is FALSE ???
+        // if (tempIncome[item.origin_id] === true) {  >> Not tested TRUE
 
-        if (tempIncome[item.origin_id] === true) {
+        if (typeof tempIncome[item.origin_id] !== 'undefined') {
```
jniles: Why don't you store this in a temporary variable?

```js
var hasIncomeValue= typeof tempIncome[item.origin_id] !== 'undefined';
tempIncome[item.origin_id] = hasIncomeValue;
if (hasIncomeValue) {
  // do something
}
```

If the `undefined` check is giving you problems, you can try `_.isUndefined(tempIncome[item.origin_id])`.  [Here is the documentation](https://lodash.com/docs/4.17.4#isUndefined).

jniles (changes requested):
@lomamech This looks pretty good!

I've made a few comments to try and improve the code. From just the code, I think it could probably be merged after this.

server/controllers/finance/reports/cashflow/index.js
lomamech: I just noticed that the data in the cashflow report differs from the account report, this difference is not obtained with the test data, I tried to add another transaction, a payment and a receipt with The main case in Dollars, I will go there to discover the origin of these errors

client/src/modules/reports/generate/cashflow/cashflow.html
```diff
@@ -0,0 +1,60 @@
...
+          <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate>
+            <!-- date interval  -->
+            <bh-date-interval
+              date-from="ReportConfigCtrl.reportDetails.dateFrom"
+              date-to="ReportConfigCtrl.reportDetails.dateTo"
+              required="true">
+            </bh-date-interval>
+
+            <!-- by week  -->
+            <div class="checkbox">
```
jniles, line 33: It might be good to include some explanation text about what a weekly checkbox does.  You can use a `<p class="help-block"></p>` tag to add this text under the `<div class="checkbox></div>`.

For reference, [here is how the Open Debtors report](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/reports/generate/open_debtors/open_debtors.html#L89) accomplishes this.

test/temporary-financial-reports/cashFlow.js
```diff
@@ -0,0 +1,171 @@
...
+    saveReport: '0',
+    weekly: '1' 
+  };
+
+  var periodicBalance = {
+    jan : 1131.42,
+    feb : 1061.45,
+    mar : 991.54,
+    apr : 924.43,
+    may : 855.9599999999999,
```
jniles: This is not a valid currency amount that we can display - if the server returns this, it means that we are doing some mathematics in JS.  We should make sure all the logic is in SQL.

I'm not sure if this should be blocking - @sfount, what do you think?

jniles (commented):
@lomamech, I made one comment about putting help text for users on the report and another about JS rounding bugs.

It looks like most of the Cashflow report is done in JS ... maybe we can schedule a refactor of this later.  It might be something I can work on in the US.  Don't worry about that comment until we figure out what the plan is for refactoring that report.  The other comment should be addressed.

jniles (approved):
Looks good to me.  Good work :+1:

PR #1813, opened 2017-06-21
Income expense report implementation

DedrickEnc:
This PR implement the income expense report 
The income expense report list the expense account and income account with their debit and credit
The report is implement following the new standard.

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


DedrickEnc:
@sfount
Changes are landed, so this PR can get a review. I am waiting for feedback

DedrickEnc:
@sfount 
Can you review this pull request?

sfount:
@DedrickEnc Changes look great, the report client is now very simple to use!

Final comments on usability: 
* It would be useful for 'Incomes and expenses' to be the default option for the 'Report Type', I think this is the option people will select the most and this would mean less clicks 
* The period selections should have the year just to confirm that they are correct (for example January 2016, February 2016) - this will show they are clearly different when you change the Fiscal year selection
* If there are no incomes or expenses at all the report is currently just a title with no contents. It might be useful to leave a message saying there are no transactions within this time period.

client/src/modules/reports/generate/income_expense/income_expense.config.js
```diff
@@ -0,0 +1,70 @@
...
+
+function IncomeExpenseConfigController($sce, Notify, SavedReports, AppCache, reportData, $state) {
+  var vm = this;
+  var cache = new AppCache('configure_income_expense');
+  var reportUrl = 'reports/finance/income_expense';
+  vm.reportDetails = {};
+  vm.previewGenerated = false;
+
+  vm.reportTypes = [
+    { id: 1, label: 'FORM.LABELS.INCOME_EXPENSE' },
```
sfount, line 16: It would be useful for this to be the default option - I think this is what users want when they come to create this report.

server/controllers/finance/fiscalPeriod.js
```diff
@@ -0,0 +1,116 @@
...
+function isInSameFiscalYear(opt){
+  if(opt.periods.length !== 2){
+    return false;
+  }
+  const sql = `
+  SELECT 
+    p.fiscal_year_id 
+  FROM 
+    period AS p`
+  const filters = new FilterParser(opt, { tableAlias : 'p', autoParseStatements : false });
```
sfount: It is good to see FilterParser in use, however in this case the code is quite complex for a simple query! 

This might be clearer to understand the condition was included directly in the SQL. This logic could also use SQL `GROUP BY` so that the reduce was not required in Javascript, something like: 

```javascript
const sql = `
  SELECT fiscal_year_id FROM period WHERE id IN ? GROUP BY fiscal_year_id 
`;

// query for fiscal years given period ids, return true if there is only one fiscal year
return db.exec(sql, [opt.periods])
   .then((fiscalYears) => fiscalYears.length === 1);
```

server/controllers/finance/fiscalPeriod.js
```diff
@@ -0,0 +1,116 @@
...
+  return db.exec(query)
+    .then(function (res){
+      const value = res.reduce(function (x, y){
+        return x.fiscal_year_id === y.fiscal_year_id;
+      });
+      return value;
+    });
+}
+
+function getPeriodDiff (periodIdA, periodIdB){
```
sfount: I think this function could be done as one MySQL query! I have not tested this though: 

```sql 
 SELECT DATEDIFF(
   (SELECT start_date FROM period where id = ?), (SELECT start_date FROM period WHERE id = ?)
 ) as nb
```

server/controllers/finance/periodTotal/index.js
```diff
@@ -0,0 +1,104 @@
...
+         IFNULL(SUM(pt.credit), 0) AS credit, IFNULL(SUM(pt.debit), 0) AS debit, 
+         ac.number, ac.type_id, ac.is_asset
+        FROM 
+         period_total AS pt
+        JOIN
+         account ac ON ac.id = pt.account_id
+        JOIN 
+         period p ON p.id = pt.period_id `;
+
+    // fix me the filter logic for that seems complicate
```
sfount: @DedrickEnc You are correct! For specific report logic like this it might be good  to not use `FilterParser` and to just write a specific query and then use '?' syntax for the values. 

That would also make it very clear what the function accepts/ requires.  When a generic options object is used it is difficult to see exactly what a function will do.

For example: 

```js
function getAccountsBalance (account, period) { 
  // ... 
}
```

server/controllers/finance/reports/incomeExpense/index.js
```diff
@@ -17,282 +17,141 @@ const _ = require('lodash');
...
+            return getDateRange(options.periodFrom, options.periodTo);
+        })
+        .then((range) => {
+            _.merge(options, { dateFrom : new Date(range.dateFrom), dateTo : new Date(range.dateTo) });
+            return fiscalPeriod.isInSameFiscalYear({ periods : [options.periodFrom, options.periodTo] })
+        })
+        .then((ans) => {
+
+            if(!ans){
+                throw new BadRequest(`The two period selected must be in the same fiscal year`, 'FORM.ERRORS.PERIOD_DIFF_FISCAL');
```
sfount: :+1: Great error handling!

server/controllers/finance/reports/incomeExpense/index.js
```diff
@@ -17,282 +17,141 @@ const _ = require('lodash');
...
-      } else {
-        incomeExpense.accountNumber = incomeExpense.accountNumber;
-      }
-
-      session.reportType = parseInt(session.reportType, 10);
-
-      if (session.reportType === 1 || session.reportType === 2) {
-        incomeExpense.incomes.forEach((income) => {
-          sumIncome += income.debit;
+function fetchIncome(options) {
```
sfount: I do not think there is any different between the fetch income and fetch expense functions - could these be written as one function that accepts a `type_id` ? 

For example:

```js
// get the total balances for accounts given a specific account type
function fetchBalances(typeId) { 
  // ...
}
```

Reducing the number of places that things could go wrong or require maintenance here would be useful!

test/temporary-financial-reports/incomeExpense.js
```diff
@@ -0,0 +1,116 @@
...
+        periodTo: 30,
+        type: 1,
+        lang : 'en',
+        renderer: 'json',
+        reportId: 3,
+        saveReport: 0
+    };
+
+    // independently tested by summing the general ledger
+    it('Returns correct data for a one period january 2016', () => {
```
sfount, line 54: 👍 Nice regression tests!

sfount (commented):
This is a very refined report! There are a few places I think the code could be reduced to make it more maintainable. As you mentioned in your comments there some places that using `FilterParser` creates very complicated code, in those places it is fine to just write the query directly! Code is looking good to me.

server/controllers/finance/periodTotal/index.js
DedrickEnc: It was difficult to implement (because the table columns was not necessarily contained in the ```options``` object) but not to understand, because the code is reduced with ```filterParser``` and very clear.
the function ```dateFrom``` and ```dateTo``` parse the date very well, this can not be done without adding more code in the report controller.

server/controllers/finance/reports/incomeExpense/index.js
DedrickEnc: These two functions build the final object (the object which will be used by the handlebar) by adding properties. Separating them seems clear. If I use one function for both of them, I will be supposed to use ```if(type_id === x)``` in many places and the function will be difficult to read.
Feel free to propose me your implementation may be it will work.

client/src/modules/reports/generate/income_expense/income_expense.config.js
DedrickEnc: ok

DedrickEnc:
@sfount

sfount:
@DedrickEnc to clear up some of my comments about duplicated code. My thinking is that these methods introduce quite a lot of complexity for a query that should just be summing the period totals table, they are good in that they are very flexible but I'm not sure the income/expense report needs this flexibility. 

I have tried to condense the information that is required into an SQL query, it would be possible to just query the accounts balance and then SUM these in Javascript but it is also possible to just change the grouping to have MySQL do the summing. 

This function 
1. Groups by account_id to get the individual account balances 
2. Groups by type_id to get the overall account type sums
3. Groups by nothing to get the balance of everything 

```js
function sumIncomeExpenseAccounts(fiscalYearId, periodFromId, periodId) {
    // @TODO These should be defined in a system shared file
    const types = [1, 2];
    let reportContext = {};

    // get all of the period IDs between the first periods number and the second periods number (within a fiscal year)
    const periodCondition = `
        (SELECT id FROM period
        WHERE fiscal_year_id = ${fiscalYearId}
        AND number BETWEEN (SELECT number FROM period WHERE id = ${periodFromId}) AND (SELECT number FROM period WHERE id = ${periodToId}))
    `;

    // Get the absolute value of the balance, if the value is negative a positive value will be returned
    const balanceQuery = `
        SELECT type_id, SUM(credit) as credit, SUM(debit) as debit, ABS(SUM(credit) - SUM(debit)) as balance
        FROM period_total JOIN account
        ON period_total.account_id = account.id
        WHERE
        period_id IN ${periodCondition}
        AND type_in IN ?
        GROUP BY ?
    `;

    // grouping by account_id gives us the individual account line items
    db.exec(balanceQuery, [types, 'account_id'])
        .then((accountBalances) => {
            reportContext.accounts = accountBalances;

            // grouping by type_id gives us total income/ expense types balance
            return db.exec(balanceQuery, [types, 'type_id']);
        })
        .then((typeBalances) => {
            reportContext.incomeBalance = typeBalances[0];
            reportContext.expenseBalance = typeBalances[1];

            // grouping by nothing gives us the overall balance of all types
            return db.one(balanceQuery, [types, 'NULL']);
        })
        .then((overallBalance) => {
            reportContext.overallBalance = overallBalance;
        });
}
```

This could maybe be split into two functions to reduce the length however I think it replaces 6 methods from the current code and reduces duplication. 

This codes does not work so it shouldn't be used directly but I think the benefits of this idea could be: 
* 100s of lines less code 
* No use of FilterParser for a simple case, just a direct query
* No duplicated code that would have to be updated in multiple places

DedrickEnc:
@sfount

client/src/modules/reports/generate/income_expense/income_expense.html
```diff
@@ -0,0 +1,60 @@
...
+              on-select-period-from-callback="ReportConfigCtrl.onSelectPeriodFrom(period)"
+              on-select-period-to-callback="ReportConfigCtrl.onSelectPeriodTo(period)">
+            </bh-fiscal-period-select>
+
+            <!-- report type selection  -->
+            <div 
+                class="form-group"
+                ng-class="{ 'has-error' : ConfigForm.$submitted && ConfigForm.type.$invalid }">
+                <label class="control-label">
+                    <i class="fa fa-briefcase"></i> {{ "FORM.SELECT.REPORT_TYPE" | translate }}
```
sfount, line 38: This should use the translate directive and should be changed in the future - this will not block the feature for now.

client/src/modules/reports/generate/income_expense/income_expense.html
```diff
@@ -0,0 +1,60 @@
...
+                </label>
+                <select
+                    class="form-control"
+                    name="type"
+                    ng-model="ReportConfigCtrl.reportDetails.type"
+                    ng-options="type.id as (type.label | translate) for type in ReportConfigCtrl.reportTypes"
+                    required>
+                    <option value="" disabled>{{ "FORM.SELECT.REPORT_TYPE" | translate }}<option>
```
sfount, line 46: Translate directive

server/controllers/finance/reports/incomeExpense/index.js
```diff
@@ -17,282 +17,142 @@ const _ = require('lodash');
...
-
-      return documentReport.render({ incomeExpense });
-    })
-    .then(result => {
-      res.set(result.headers).send(result.report);
-    })
-    .catch(next)
-    .done();
-}
+function getDateRange(periodIdFrom, periodIdTo) {
```
sfount, line 147: Very clean method 👍

sfount (approved):
This PR looks good to me. This report has passed all of the requirements according the accountants at IMA and BHIMA style. 

This can be merged, great work!

Issue #1812, opened 2017-06-21
(bug) Account statement form validation


sfount:
The account statement report client module has a number of bugs to do with which fields are required for creating the report. 

- [x] BUG - Submitting the report without an account gives no user feedback. It should highlight the account select in red with text saying that this is a required field 
- [x] The period select is no longer optional for this report - it should not be hidden behind a flag 
- [x] If a date is missing or invalid the report should not be submitted

sfount:
Closed in #1812

PR #1815, opened 2017-06-22
fixing the 1812 issue about highlighting fields in red when the form …

jeremielodi:


client/src/modules/reports/generate/account_report/account_report.html
```diff
@@ -22,30 +22,51 @@ <h3 class="text-capitalize" translate>REPORT.REPORT_ACCOUNTS.TITLE</h3>
...
-              on-select-callback="ReportConfigCtrl.selectAccount(account)">
-            </bh-account-select>
 
-            <!-- included Date interval  -->
-            <div class="checkbox">
-              <label>
-                <input type="checkbox" ng-model="ReportConfigCtrl.dateInterval" ng-true-value="1" ng-false-value="0">
-                <span translate>FORM.LABELS.INCLUDE_DATE_INTERVAL</span>
-              </label>
+            <div class="form-group"
```
sfount: The `bh-account-select` component already has a form-group error class as well as an `ng-messages` block. We do not need a second version here. It would be good to investigate why the ng-message and error validation are not working for the component in this context.

client/src/modules/reports/generate/account_report/account_report.html
```diff
@@ -22,30 +22,51 @@ <h3 class="text-capitalize" translate>REPORT.REPORT_ACCOUNTS.TITLE</h3>
...
+            
 
             <!-- Date interval  -->
             <!-- @TODO this should use a component that callsback with well defined dates -->
             <div ng-if="ReportConfigCtrl.dateInterval === 1">
-              <bh-date-interval
-                date-from="ReportConfigCtrl.reportDetails.dateFrom"
-                date-to="ReportConfigCtrl.reportDetails.dateTo">
-              </bh-date-interval>
+               <div class="form-group">
```
sfount: I think we can remove this `form-group` now that the validation is done using `validation-trigger`.

client/src/modules/reports/generate/account_report/account_report.html
```diff
@@ -22,30 +22,51 @@ <h3 class="text-capitalize" translate>REPORT.REPORT_ACCOUNTS.TITLE</h3>
...
+                
+                    <div class="input-group col-lg-12">
+                  <bh-date-interval
+                  
+                    date-from="ReportConfigCtrl.reportDetails.dateFrom"
+                    date-to="ReportConfigCtrl.reportDetails.dateTo"
+                    required
+                    validation-trigger="ConfigForm.$submitted"
```
sfount: 👍 Good catch! This will help the user see when they have not formatted the date correctly.

client/src/modules/templates/bhDateInterval.tmpl.html
```diff
@@ -1,6 +1,9 @@
 <ng-form name="DateIntervalForm" novalidate>
 
-  <div class="form-group" style="margin-bottom:0">
+  <div 
+    class="form-group"
+    ng-class="{'has-error': $ctrl.validationTrigger && (DateIntervalForm.dateFrom.$invalid || DateIntervalForm.dateTo.$invalid)}"
```
sfount, line 5: 👍

sfount (commented):
This looks like a great fix! 

I have made some comments to make sure that we are not duplicating code that is already written inside the components here. 

If the validation for the `bh-account-select` does not work using the `$$parent.$submitted` potentially this can be updated to use an attribute `validation-trigger` on the component.

sfount:
@jeremielodi It is worth noting that the error validation on the `bh-account-select` component in other places in BHIMA (like Cashbox management) also doesn't work. This component might need to be fixed/ updated.

sfount (approved):
This is the perfect fix, LGTM 👍 . Once tests pass this can be merged.

sfount (approved):
Changes LGTM :+1: Great fix!

sfount:
Closes #1812

PR #1816, opened 2017-06-22
Feat bilan report

### LINKED ISSUE #1801: (feature) Enterprise Balance Statement
This issue will track the development of the core enterprise balance statement. This is a standard accounting summary and should provide an overview a verification of a enterprises finances. 

Currently in BHIMA we have an `account_type` that allows us to categorise accounts based on if they are income, expense or balance accounts. 

In order to correctly categorise all accounts according to standard accounting practices we need finer granularity. The proposal is to add a table `account_category` that will allow selection from the 5 main account types:
* Asset 
* Liability 
* Equity 
* Revenue 
* Expense 

In theory the balance statement should demonstrate that: 
* Assets = Liabilities + Equity 
* Net Income = Revenues - Expenses 

The `account_category` table will have a reference to the `account_type` - an account will now be assigned an account category which will implicitly give it an account type. 

Requirements for balance statement: 

1. Add 'Account Category' table to schema (this requires update SQL for current supported databases)
2. Update account management module to select category in place of type 
3. Update development data set with correct categories 
4. Create balance sheet report using account categories 

In the future we will look at how nested accounts are in the chart of accounts for displaying more or less detailed views into these categories.

mbayopanda:
This PR is about the balance sheet report #1801 

Closes #1801


server/controllers/finance/reports/balance_sheet/index.js
```diff
@@ -0,0 +1,256 @@
...
+
+const _ = require('lodash');
+const moment = require('moment');
+const db = require('../../../../lib/db');
+const ReportManager = require('../../../../lib/ReportManager');
+
+// report template
+const TEMPLATE = './server/controllers/finance/reports/balance_sheet/report.handlebars';
+
+const ASSET = 1;
```
sfount, line 23: In the future these should be defined as system wide variables - can you create an issue to track this feature?

server/controllers/finance/reports/balance_sheet/report.handlebars
```diff
@@ -0,0 +1,138 @@
...
+      {{!-- liabilities --}}
+      <h4 style="margin: 2px;"><strong>{{translate "ACCOUNT.TYPES.LIABILITY"}}</strong></h4>
+      <table class="report-condensed-table">
+        <tbody>
+          {{#each liabilities.accounts as |account|}}
+            {{#if account.balance}}
+            <tr class="text-right">
+              <td style="width: 70%;"  class="text-left"><b>{{ account.number }}</b> - {{ account.label }}</td>
+              <td style="width: 15%;"  class="text-right">{{#gt account.balance 0}}{{currency account.balance metadata.enterprise.currency_id}}{{/gt}}</td>
+              <td style="width: 15%;"  class="text-right">{{#lt account.balance 0}}{{currency (multiply account.balance -1) metadata.enterprise.currency_id}}{{/lt}}</td>
```
sfount: It is best to reduce the amount logic that templates have to perform - it is much easier for another developer to understand what is happening if the display layer (this HTML template) does not manipulate the values directly. 

This value should be calculated in the controller.

server/controllers/finance/reports/balance_sheet/index.js
```diff
@@ -0,0 +1,256 @@
...
+  // gets the amount up to the current period
+  const sql = `
+    SELECT a.number, a.id, a.label, a.type_id,
+      SUM(pt.credit) AS credit, SUM(pt.debit) AS debit, SUM(pt.debit - pt.credit) AS balance 
+    FROM period_total AS pt JOIN account AS a ON pt.account_id = a.id
+    JOIN period AS p ON pt.period_id = p.id
+    WHERE pt.enterprise_id = ?  
+      AND (DATE(p.start_date) <= DATE(?) OR (p.start_date IS NULL OR p.end_date IS NULL))
+      AND pt.fiscal_year_id = (SELECT f.id FROM fiscal_year f WHERE DATE(?) BETWEEN DATE(f.start_date) AND DATE(f.end_date) LIMIT 1)
+    GROUP BY a.id `;
```
sfount, line 222: It would be best to do a lot of the sum/ totals in MySQL if it is possible. 

I suggest adding a variable to change the grouping. 

For example 
```sql
GROUP BY account_id #returns account balances
```
```sql 
GROUP BY type_id #returns category balances 
```
```sql 
GROUP BY NULL #returns flat balance (all accounts) 
```

You can also use ```ABS()``` to return a positive balance.

sfount (commented):
This report looks excellent! All code LGTM. 

It would be good to remove the Javascript Maths logic and implement this with MySQL. 

Another small thing is the chart of accounts report uses a hardcoded type id (server/controllers/finance/reports/accounts/index.js). This should be updated to 6 instead of 4!

mbayopanda:

PR #1818, opened 2017-06-25
refactor(General Ledger Module)

### LINKED ISSUE #1810: (refactor) General Ledger Module
We currently have a module that shows an accounts balance across each period in a fiscal year. By default this module shows the current fiscal and there is no way of changing this. 

In order to verify the data in other reports it is necessary to update this report to allow the accountant to view data in different fiscal years, as well as review account transactions accurately. 

This refactor should include: 
- [ ] Add a dropdown (for now) to the top of the module that allows the user to select which fiscal year is being dispalyed 
- [ ] Use the latest account statement report to show all of the transactions against the report (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1809) 
- [ ] ONLY show transactions for the account within the current fiscal year (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1809)

lomamech:
refactor(General Ledger Module)
- Add modal for the select of fiscal year
- View general ledger by fiscal year
- Use the selected fiscal year for the selected account statement

closes #1810


client/src/modules/fiscal/fiscal.list.html
```diff
@@ -36,7 +36,7 @@ <h4 style="margin-top : 0px;">
             <div class="row">
               <span class="fa fa-calendar"></span>
               {{ fiscal.start_date | date }} - {{ fiscal.end_date | date }} ({{ fiscal.number_of_months }}
-              {{ "FORM.LABELS.PERIODS" | translate }})
+              {{ "FORM.LABELS.PERIODES" | translate }})
```
jniles: This should use the `translate` directive.
```html
<span translate>FORM.LABELS.PERIODS</span>
```

client/src/modules/general-ledger/general-ledger-accounts.ctrl.js
```diff
@@ -172,6 +178,11 @@ function GeneralLedgerAccountsController(GeneralLedger, Session, Notify,
     vm.loading = !vm.loading;
   }
 
+  // API register function
+  function onRegisterApi(gridApi) {
```
jniles, line 182: Do you actually use this?  I cannot seem to find where `gridApi` is used.  If not, you should remove it.

client/src/modules/general-ledger/general-ledger-accounts.ctrl.js
```diff
@@ -209,28 +210,58 @@ function GeneralLedgerAccountsController(GeneralLedger, Session, Notify,
...
-    var params = formatExportParameters('csv');
+  vm.download = GeneralLedger.download;
+  vm.slip = GeneralLedger.slip;
+
+  // open search modal
+  vm.openFiscalYearConfiguration = function openFiscalYearConfiguration() {
+    Modal.openSelectFiscalYear()
+      .then(function (filters) {
+        if (!filters) { return; }
+          vm.fiscalYearLabel = filters.fiscal_year.label;
```
jniles: Your indentation is really poor here - can you follow eslint's indentation guidelines?  It should look like:
```js
if (!filters) { return; }
vm.fiscalYearLabel = filters.fiscal_year.label;
```
instead of
```js
if (!filters) { return; }
  vm.fiscalYearLabel = filters.fiscal_year.label;
```

server/controllers/finance/fiscal.js
```diff
@@ -622,24 +622,23 @@ function closing(req, res, next) {
 }
 
 /**
- * @method getPeriodCurrent
+ * @method getPeriodByFiscal
  *
  * @description
- * This function returns all period of current fiscal Year
- * the Date provided.  If no record is found, it throws a NotFound error.
+ * This function returns all Fiscal Year's periods
```
jniles, line 634: This is a nice update to the documentation :+1:

server/controllers/finance/generalLedger/index.js
```diff
@@ -100,9 +100,10 @@ function list(req, res, next) {
  * list accounts and their solds
  */
 function listAccounts(req, res, next) {
-  const currentDate = new Date();
+  let fiscal_year_id;
+  fiscal_year_id = req.query.fiscal_year_id;
```
jniles: These two should be on the same line and the `let` should be a `const`.

server/controllers/finance/reports/generalLedger/index.js
```diff
@@ -35,20 +35,22 @@ function renderReport(req, res, next) {
...
-  const currentDate = new Date();
 
   try {
     report = new ReportManager(REPORT_TEMPLATE, req.session, options);
   } catch (e) {
     return next(e);
   }
 
-  return Fiscal.getPeriodCurrent(currentDate)
+  let fiscal_year_id = options.fiscal_year_id;
```
jniles: This should be camelCase (`fiscalYearId`) and should be a `const`.  I think ESLint should catch this.

jniles (changes requested):
@lomamech, I think @sfount's original idea was to simply have a `<ui-select>` at the top of the module.  I think that would be a lot easier - it wouldn't require you to implement a modal.

I've mad some other comments on the PR - make sure you are following ESLint!  Let me know if you have any other questions.

client/src/modules/general-ledger/general-ledger-accounts.ctrl.js
lomamech: Yes it used in 
````
  function toggleFilter() {
    vm.filterEnabled = !vm.filterEnabled;
    vm.gridOptions.enableFiltering = vm.filterEnabled;
  }
````

server/controllers/finance/reports/generalLedger/index.js
lomamech: @jniles Let me know, if it is still necessary to use a ui-select instead modal for the select of Fiscal Year

jniles: Let's ask @sfount - he made the original issue and has worked with the team at Vanga most recently.

@sfount what is your opinion?

client/src/modules/general-ledger/general-ledger-accounts.ctrl.js
jniles: Ah, I see it now.  Thanks!

server/controllers/finance/reports/generalLedger/index.js
sfount: I don't think the staff at the hospital would notice either way - it would be good to make a decision as BHIMA though. 

We will see a similar behavior I think on the account statement module selecting a single account to drive the module. 

My opinion is that the modal just adds a lot of boilerplate that we wouldn't otherwise need but I don't think it's a bad user experience.

jniles (approved):

PR #1820, opened 2017-06-27
Clients Report Implementation

DedrickEnc:
This PR implement the clients report implementation

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


DedrickEnc:
@sfount
Can I get a review?

sfount:
@DedrickEnc I have tested the functionality of this PR and it seems to work well. Can you update it to the latest `master` branch? It can then be merged.

DedrickEnc:
~~Will open an other one, more file are broken and vanished~~

client/src/i18n/en/report.json
```diff
@@ -20,7 +20,7 @@
     "CASHFLOW":"Cashflow",
     "CASH_INCOME":"Incomes",
     "CHART_OF_ACCOUNTS":"Chart of Accounts",
-    "CLIENTS_REPORT":{"TITLE":"Clients Report", "CURRENT_MVT":"Current Movement", "PREVIOUS_MVT":"Previous Movement"},
+    "CLIENTS_REPORT":{"TITLE":"Clients Report", "CURRENT_MVT":"Current Movement", "PREVIOUS_MVT":"From biginning to :"},
```
jniles: This should read "From Beginning To:".

client/src/js/components/bhDebtorGroupSelect.js
```diff
@@ -45,7 +45,7 @@ function DebtorGroupSelectController(Debtors) {
   };
 
   // fires the onSelectCallback bound to the component boundary
-  $ctrl.onSelect = function ($item) {
+  $ctrl.onSelect = function ($item, model) {
```
jniles: Do you need the model here?

client/src/js/components/bhMultipleDebtorGroupSelect.js
```diff
@@ -0,0 +1,51 @@
+angular.module('bhima.components')
```
jniles, line 1: I think you could pretty easily make this the same component as the `bhDebtorGroupSelect`.  Just use a `transclude` that puts in the `<ui-select multiple>` directive.  It would let you reuse the code and tests from the previous component.

If you need some guidance on this, I can take a look at it.

client/src/js/components/bhMultipleDebtorGroupSelect.js
```diff
@@ -0,0 +1,51 @@
...
+  .component('bhMultipleDebtorGroupSelect', {
+    templateUrl : 'modules/templates/bhMultipleDebtorGroupSelect.tmpl.html',
+    controller  : MultipleDebtorGroupSelectController,
+    transclude  : true,
```
jniles: I don't see where you used `transclude` here.  Do you need it?

client/src/js/components/bhMultipleDebtorGroupSelect.js
```diff
@@ -0,0 +1,51 @@
...
+    bindings    : { 
+      label            : '@?',
+      onSelectCallback : '&',
+      onRemoveCallback : '&',
+      formName         : '@?',
+      required         : '<?'
```
jniles: I don't see where you needed `required`.  Can it be removed?

client/src/js/components/bhMultipleDebtorGroupSelect.js
```diff
@@ -0,0 +1,51 @@
...
+
+    // fired when an debtor group has been selected
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
+
+    // default for form name
+    $ctrl.formName = $ctrl.formName || 'DebtorGroupForm';
+
+    // load all Debtor Group
+    DebtorGroups.read()
+      .then(function (dgs) {
```
jniles, line 36: I think this should have a `.catch()` method to handle errors.

client/src/modules/reports/generate/clients_report/clients_report.html
```diff
@@ -0,0 +1,64 @@
...
+        <div class="panel-heading">
+          <span translate>REPORT.UTIL.OPTIONS</span>
+        </div>
+
+        <div class="panel-body">
+          <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate>
+             <!-- show a simple or complex report  -->
+            <div class="form-group">
+              <label class="text-action">
+                <input type="checkbox" name="simple_preview" ng-model="ReportConfigCtrl.reportDetails.simplePreview" ng-true-value="true" ng-false-value="false">
```
jniles: By default, `true` is `true` and `false` is `false`.  There is no need to use `ngTrueValue` and `ngFalseValue`.  See the [input[checkbox] documentation](https://docs.angularjs.org/api/ng/input/input%5Bcheckbox%5D).

client/src/modules/reports/generate/clients_report/clients_report.html
```diff
@@ -0,0 +1,64 @@
...
+            <!-- detail the previous year flag  -->
+            <div class="form-group" ng-if="!ReportConfigCtrl.reportDetails.simplePreview">
+              <label class="text-action">
+                <input type="checkbox" name="detail_previous" ng-model="ReportConfigCtrl.reportDetails.detailPrevious">
+                <span translate>FORM.LABELS.DETAIL_PREVIOUS</span>
+              </label>
+            </div>
+
+            <!-- Client to ignore  -->
+            <bh-multiple-debtor-group-select
```
jniles, line 49: This is a really clean implementation.

client/src/modules/templates/bhMultipleDebtorGroupSelect.tmpl.html
```diff
@@ -0,0 +1,22 @@
+<div bh-multiple-debtor-group-select>
+  <div class="form-group">
+    <label class="control-label" translate>{{$ctrl.label}}</label>
+      <ui-select 
+        multiple
+        name="clients"
+        ng-model="$ctrl.selectedDebtorGroups"
+        on-select="$ctrl.onSelect($ctrl.selectedDebtorGroups)"
+        on-remove="$ctrl.onRemove($ctrl.selectedDebtorGroups)"
+        theme="bootstrap"
```
jniles: You don't need to set the theme - it is set in `app.js` for the entire application to ensure a uniform look.

client/src/modules/templates/bhMultipleDebtorGroupSelect.tmpl.html
```diff
@@ -0,0 +1,22 @@
...
+        close-on-select="false">
+          <ui-select-match placeholder="{{ 'FORM.SELECT.DEBTOR_GROUP' | translate }}">
+            <span>{{$item.name}}</span>
+          </ui-select-match>
+
+          <ui-select-choices repeat="dg.uuid as dg in $ctrl.debtorGroups| filter:{name: $select.search}">
```
jniles, line 15: Nice filter!

test/end-to-end/shared/components/bhDebtorGroupSelect.js
```diff
@@ -8,8 +8,6 @@ module.exports = {
     const locator = (uuid) ? by.id(uuid) : by.css(this.selector);
     const target = element(locator);
 
-    // hack to make sure previous 'blur' event fires if we are using
```
jniles: Why did you remove this comment?

utilities/translation/tfcomp.js
```diff
@@ -55,7 +55,7 @@ jsonFiles.forEach(function (jsonFile) {
   // Report items in french translation but missing from english translation
   if (enMissList.length > 0) {
     errMsg += '\nMissing from ' + jsonFile.en + ': \n';
-    enMIssList.sort();
```
jniles, line 58: Nice catch! 👍

client/src/js/components/bhMultipleDebtorGroupSelect.js
```diff
@@ -0,0 +1,51 @@
...
+      required         : '<?'     
+    },
+  });
+
+MultipleDebtorGroupSelectController.$inject = [
+  'DebtorGroupService'
+];
+
+/**
+ * User selection component
```
jniles: This documentation should be updated.

client/src/js/components/bhMultipleDebtorGroupSelect.js
```diff
@@ -0,0 +1,51 @@
...
+
+  $ctrl.$onInit = function onInit() {
+    //label to display
+    $ctrl.label = $ctrl.label || 'FORM.LABELS.DEBTOR_GROUP';
+
+    // fired when an debtor group has been selected
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
+
+    // default for form name
+    $ctrl.formName = $ctrl.formName || 'DebtorGroupForm';
```
jniles, line 32: I don't think you use the `formName` variable.

client/src/modules/reports/generate/clients_report/clients_report.config.js
```diff
@@ -0,0 +1,77 @@
...
+      })
+      .catch(Notify.handleError);
+  };
+
+  vm.preview = function preview(form) {
+    if(form.$invalid){ return; }
+    cache.reportDetails = angular.copy(vm.reportDetails);
+
+    var sendDetails = sanitiseDateStrings(vm.reportDetails);
+    sendDetails.dateTo = Moment(sendDetails.dateTo).format('YYYY-MM-DD');
```
jniles: These date formats could be stored in `bhConstant`s.

client/src/modules/reports/generate/clients_report/clients_report.config.js
```diff
@@ -0,0 +1,77 @@
...
+        // update cached configuration
+        vm.previewGenerated = true;
+        vm.previewResult = $sce.trustAsHtml(result);
+      })
+      .catch(Notify.handleError);
+  };
+
+  function sanitiseDateStrings(options) {
+    var sanitisedOptions = angular.copy(options);
+    sanitisedOptions.dateTo = Moment(sanitisedOptions.dateTo).format('YYYY-MM-DD');
```
jniles: It looks like you performed this operation twice.  Putting it in this function is a good idea!  You can remove the other date parsing.

client/src/modules/reports/generate/clients_report/clients_report.config.js
```diff
@@ -0,0 +1,77 @@
...
+    sanitisedOptions.dateFrom = Moment(sanitisedOptions.dateFrom).format('YYYY-MM-DD');
+    return sanitisedOptions;
+  }
+
+  function checkCachedConfiguration() {
+    if (cache.reportDetails) {
+      vm.reportDetails = angular.copy(cache.reportDetails);
+    }
+    //FIX ME : We don't need the ignored clients list
```
jniles: Can we remove the variable then?

client/src/i18n/fr/report.json
```diff
@@ -56,6 +56,12 @@
       "TITLE" : "Rapport des Comptes d'exploitation",
       "DESCRIPTION" : "Montre les profits et charges qui se sont produit au sein de l'entreprise"
     },
+    "CLIENTS_REPORT" : {
+      "TITLE" : "Rapport Clients",
+      "CURRENT_MVT" : "Mouvement Courant",
+      "PREVIOUS_MVT" : "Du debut Au :",
+      "DESCRIPTION" : "Ce rapport affiche tous les groupes de debiteurs, leurs situations precedent l'annee en cours, leurs situations de l'annee en cours ainsi que la balance finale"
```
jniles, line 79: I think the English description is missing.  It should be 
> This report displays all the debtor groups, their account balances at the beginning of the fiscal year, the current balances, and their ending balances.

client/src/modules/reports/generate/clients_report/clients_report.html
```diff
@@ -0,0 +1,64 @@
...
+    <div class="col-md-6">
+      <div class="panel panel-default">
+        <div class="panel-heading">
+          <span translate>REPORT.UTIL.OPTIONS</span>
+        </div>
+
+        <div class="panel-body">
+          <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate>
+             <!-- show a simple or complex report  -->
+            <div class="form-group">
```
jniles: In bootstrap, the correct way to display a checkbox is outside of a `form-group` class like this:
```html
  <div class="checkbox">
    <label>
      <input type="checkbox"> Check me out
    </label>
  </div>
```

For more information, see the [bootstrap documentation for forms](http://getbootstrap.com/css/#forms-example).

server/models/bhima.sql
```diff
@@ -98,7 +98,7 @@ INSERT INTO `report` (`id`, `report_key`, `title_key`) VALUES
   (6, 'account_report', 'REPORT.REPORT_ACCOUNTS.TITLE'),
   (7, 'cashflowByService', 'REPORT.CASHFLOW_BY_SERVICE'),
   (8, 'openDebtors', 'REPORT.OPEN_DEBTORS.TITLE'),
-  (9, 'clients_report','REPORT.CLIENTS_REPORT'),
+  (9, 'clients_report','REPORT.CLIENTS'),
```
jniles: I think the translation key should be `REPORT.CLIENTS_REPORT`.  The i18n key `REPORT.CLIENTS` hasn't been defined.

test/end-to-end/reports/clients_report/clients_report.spec.js
```diff
@@ -0,0 +1,48 @@
+/* global browser, element, by */
+
+const chai = require('chai');
+const helpers = require('../../shared/helpers');
+
+helpers.configure(chai);
+
+const ReportClientsPage = require('./clients_report.page');
+
+// @FIXME skip end to end tests until form validation and components are finalised
```
jniles: Does this fixme still apply?

test/end-to-end/reports/clients_report/clients_report.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    end_date : '31/12/2016',
+    clients : ['First Test Debtor Group'],
+    report_name : 'Clients Report Saved by E2E',
+    renderer : 'PDF'
+  };
+
+  before(() => {
+    helpers.navigate(`#!/reports/${key}`);
+    Page = new ReportClientsPage(key);
+    browser.refresh();
```
jniles, line 25: Out of curiosity, why do you need to refresh the browser?

test/end-to-end/shared/components/bhMultipleDebtorGroupSelect.js
```diff
@@ -0,0 +1,19 @@
...
+
+module.exports = {
+  selector : '[bh-multiple-debtor-group-select]',
+  set      : function set(debtorGroups, id) {
+    const locator = (id) ? by.id(id) : by.css(this.selector);
+    const target = element(locator);
+
+    target.click();
+
+    debtorGroups.forEach(function (dg){
```
jniles, line 13: :+1:

jniles (changes requested):
@DedrickEnc, this is a pretty good start.  There are a few mistranslations and unused variables in the code base that should be cleaned up before merging.

I've suggested that you simply re-use the debtor groups component, instead of writing a new one (it would be simpler to test and maintain that way).  If you find it too difficult to extend the original component, the new one can stay.

client/src/modules/reports/generate/clients_report/clients_report.config.js
DedrickEnc: how? by using ```delete```?

jniles: No - just removing the line.  The comment sounds like it is never used, and I couldn't find a place where the code used it.

test/end-to-end/reports/clients_report/clients_report.spec.js
DedrickEnc: Our end to end test see,s strange, it needs an investigation. For now we are refreshing the page to let test pass otherwise only the first report test will work.

jniles: It sounds like an action isn't occurring that needs to.  Make sure that if you are showing a success message or something similar, you correctly dismiss the message in the test before attempting to execute a second test.

client/src/js/components/bhMultipleDebtorGroupSelect.js
DedrickEnc: @jniles 
I have noticed this : 
* The two components interfaces are : 

**For bhDebtorGroupSelect**

```
bindings    : {
      debtorGroupUuid   : '<',
      disable           : '<?',
      onSelectCallback  : '&',
      name              : '@?',
      label             : '@?',
      required          : '<?',
      validationTrigger : '<',
    }
```
**For bhMultipleDebtorGroup**

```

bindings    : { 
      label            : '@?',
      onSelectCallback : '&',
      onRemoveCallback : '&',
      formName         : '@?'
    }

```

* The controller can be combined without problem
* The views (html templates) are pretty different, the ```bhMultipleDebtorGroupSelect``` does not close the select after a selection, and the ```on-select``` event is firing a function which receive a model (not ```$item``` like bhDebtorGroupSelect).

The easiest solution is to create an issue so I can do it separately or I will need some guidance.
What do you think?

jniles: Go ahead and make another issue.  This report should get in as soon as possible.

client/src/modules/reports/generate/clients_report/clients_report.config.js
DedrickEnc: Actually, It is used at the server side. In the comment I mean we don't need the ignored clients from cache.
So the the ```ignoredClients``` array should be empty when the page loads

test/end-to-end/reports/clients_report/clients_report.spec.js
DedrickEnc: Yes, this bug is in every report it should be tackled separately and once at all for every report.

DedrickEnc:
@jniles 
The test passed

client/src/modules/reports/generate/clients_report/clients_report.html
```diff
@@ -0,0 +1,64 @@
...
+              <label class="text-action">
+                <input type="checkbox" name="detail_previous" ng-model="ReportConfigCtrl.reportDetails.detailPrevious">
+                <span translate>FORM.LABELS.DETAIL_PREVIOUS</span>
+              </label>
+            </div>
+
+            <!-- Client to ignore  -->
+            <bh-multiple-debtor-group-select
+              label="FORM.LABELS.CLIENT_TO_IGNORE"
+              formName="ConfigForm"
```
jniles: Out of curiosity, what do you need this for?

server/controllers/finance/reports/clientsReport/index.js
```diff
@@ -24,25 +26,47 @@ function document(req, res, next) {
...
     report = new ReportManager(TEMPLATE, req.session, params);
   } catch (e) {
     return next(e);
   }
 
-  // Getting data to be rendered
-  return fetchClientsData(session)
+  // if there is some client to ignore
+  if (session.ignoredClients) {
+    const ignoredClients = (Array.isArray(session.ignoredClients)) ? session.ignoredClients : [session.ignoredClients];
```
jniles: This is kind of cool:
```js
const ignoredClients = [].concat(session.ignoredClients);
```
and will produce the same result!

(Please don't do this - your implementation is much better.  I just thought it was cool and wanted to share it.)

server/controllers/finance/reports/clientsReport/index.js
```diff
@@ -226,4 +223,176 @@ function escapeItems(list) {
   }, []);
 }
 
+/**
+ * @function getClientsData
+ * @param {object} options The object for filtering data using filterParser lib
+ * @param {boolean} isFromPeriodTotal is true if the data will be fetched from the period_total
+ * @param {boolean} isIniial is true if the data will be considered as initial
```
jniles: This parameters should probably spelled correctly - `isInitial`.

server/controllers/finance/reports/clientsReport/index.js
```diff
@@ -52,167 +76,140 @@ function document(req, res, next) {
...
- * @param {object} date The date in which we want to get the fiscal year
+ * @param {date} date The date in which we want to get the fiscal year
+ * @param {integer} periodZeroNumber the period zero number
  * @description
- * This function is responsible of returning a correct fiscal year
+ * This function is responsible of returning a correct fiscal year and its period zero Id
  * according a date given
  */
-function getFiscalYear(date) {
+function getFiscalYear(date, periodZeroNumber) {
```
jniles: Why do we need `periodZeroNumber`?

server/controllers/finance/reports/clientsReport/index.js
```diff
@@ -226,4 +223,176 @@ function escapeItems(list) {
...
+ * @param {boolean} isIniial is true if the data will be considered as initial 
+ * @description
+ * This function is responsible of returning a list of debtor and their 
+ * debit, credit, balance
+ **/
+function getClientsData(options, isFromPeriodTotal, isIniial) {
+  let finalQuery;
+  let parameters;
+
+  if (isFromPeriodTotal) {
```
jniles: This function looks like this:
```js
function getClientsData(options, isFromPeriodTotal, isIniial) {
  if (isFromPeriodTotal) {
    // do something
  } else {
   // do something else
  }
}
```

It would probably be better to just make a series of unique functions, since the only repeating code is the `db.exec()` call.  Probably something like:
```js

// computes the client balances using the period total's table
getClientBalancesFromPeriodTotal()

getClientBalancesFromGeneralLedger()
```

This would make the code easier to read by:
 1. Removing an `if/else` block, so that the programmer doesn't have to remember too many logical layers of code
 2. Make the name of the function more descriptive to what the data source is
 3. `isIniial` only applies to the `isFromPeriodTotal` branch; the parameter is never used when `isFromPeriodTotal` is false.  Since we don't need it, we can only include it as a parameter in a function when it is needed.

server/controllers/finance/reports/clientsReport/index.js
```diff
@@ -226,4 +223,176 @@ function escapeItems(list) {
...
+          t.number AS accountNumber, t.name, IFNULL(t.debit, 0) AS debit, IFNULL(t.credit, 0) AS credit, 
+          IFNULL(t.balance, 0) AS balance
+        FROM 
+          (${query}) AS t`;
+  }
+
+  return db.exec(finalQuery, parameters);
+}
+
+function getClientsTotal(options, isFromPeriodTotal, isIniial) {
```
jniles: This function could be similarly broken up for maintainability.

jniles (commented):
@DedrickEnc, the changes have been well made. :+1:

Now that the report has been approved, I've made a few suggestions for maintaining the codebase on the server.  You have some pretty complex functions that could be made simpler by breaking them into smaller blocks, and by naming variables in a way that describes their purpose.

client/src/modules/reports/generate/clients_report/clients_report.html
DedrickEnc: For this purpose, it can be removed

DedrickEnc: Actually, nothing, it can be removed

server/controllers/finance/reports/clientsReport/index.js
DedrickEnc: Why? I like it!

server/controllers/finance/reports/clientsReport/index.js
DedrickEnc: We need it to get the period zero ID

server/controllers/finance/reports/clientsReport/index.js
jniles: It just seems like a hack and Array.isArray() is semantically clearer.  I suppose you can add it if you want, but it would probably needed a comment like "Ensure ignored clients is an array" or something.

DedrickEnc:
@jniles 
All the changes are landed

test/temporary-financial-reports/clientsReport.js
```diff
@@ -0,0 +1,136 @@
...
+  it('Returns expected data for a simple report preview', () => {
+    const finalBalance = 57.3137;
+    const totalLines = 4;
+
+    return agent.get(reportURL)
+      .query(simpleReportCurrentYear)
+      .then((result) => {
+        const report = result.body;
+
+        expect(Number(report.totalBalance)).to.equal(finalBalance);
```
jniles, line 118: In the future, I don't think the `Number()` call is necessary ... in fact, it would be good to make sure everything returned is an actual number, instead of parsing it to a number first.

jniles (approved):
LGTM.

There's a lot of complexity in this report that looks like it could be moved into SQL. I'm not able to build the report locally (I'm on a chromebook since I'm travelling), but I assume it works as expected.  We can take a look at reducing the complexity when I'm back in the office.

PR #1830, opened 2017-07-09
fix(build) Catch networking issue for WSL

sfount:
This commit is a temporary patch fix for developers using
Windows Subsystem for Linux. The networking interfaces are
not fully exposed prior to the creators update. Once this
is addressed this patch should be removed.


jniles:
This is cool!  Are we supporting WSL now?

PR #1835, opened 2017-07-10
perf(Data base for regression test)

### LINKED ISSUE #1834: (perf) Upgrade Database for regression test
There is some problem with the database for the regression test, particularly with regard to the creditor group,

- Hence the need to modify the addition of creditor groups
- To modify certain accounting entries, in particular those relating to the payment of the company's receivables
- Addition of the invoices concerning the invoicing of the company to its creditors
- To update the elements of the period_total table

lomamech:
perf(Data base for regression test) 
- Add creditors groups
- Update transactions for the invoicing by creditors
- Update data in general_ledger and period_total

closes #1834


jniles:
@lomamech, it looks like there are conflicts here. Could you attempt to fix those?

jniles:
@lomamech, I've tested this PR and it seems to pass tests ... should it still be merged?  I wasn't sure what the original fix was for.

lomamech:
@jniles This issue corrects a good number of transactions to the turn of the invoicing of the hospital to its suppliers as well as the payment of the debts, in the database development-build.sql there exist several transactions which simulates the payment of the invoices Which is presented in the following manner

compta

According to the principle of accounting, it is not possible to write directly between the fund and a charge account, but also before paying a debt it would have to be possible beforehand that this debt could already exist, that is why this PR Corrects problems by adding writings between a creditor's account and the expense account and for the payment of the debt a writing between the account of a creditor and the cash account

correc

We also add two accounts for creditors, this will allow us to have data to finalize the aging report of the receivables

jniles:
@lomamech, great analysis.  Thanks for helping me understand.

LGTM!

PR #1836, opened 2017-07-10
Cash boxes report

DedrickEnc:
This pull request updated the report call initially **Income expense report** to look like other report (fallowing new report standard). 

The report name has changed from **Income expense report** to **Cash Report**

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


DedrickEnc:
@sfount
This PR can get a review.

DedrickEnc:
@jniles
Can I get a review?

client/src/modules/reports/generate/cash_report/cash_report.html
```diff
@@ -0,0 +1,80 @@
...
+
+        <div class="panel-body">
+          <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate>
+            
+            <!-- report type selection  -->
+            <div 
+                class="form-group"
+                ng-class="{ 'has-error' : ConfigForm.$submitted && ConfigForm.type.$invalid }">
+                <label class="control-label">
+                    <i class="fa fa-briefcase"></i> {{ "FORM.SELECT.REPORT_TYPE" | translate }}
```
jniles: This text should use the `translate` directive.

server/controllers/finance/reports/cashReport/index.js
```diff
@@ -0,0 +1,188 @@
...
+const BadRequest = require('../../../../lib/errors/BadRequest');
+
+const TEMPLATE = './server/controllers/finance/reports/cashReport/report.handlebars';
+
+// expose to the API
+exports.document = document;
+
+
+function getRecordQuery (token){
+  const query =
```
jniles: This is a well-written query :+1:

server/controllers/finance/reports/cashReport/index.js
```diff
@@ -0,0 +1,188 @@
...
+  if (!params.dateFrom || !params.dateTo) {
+    throw new BadRequest('Date range should be specified', 'ERRORS.BAD_REQUEST');
+  }
+
+  if (!params.account_id) {
+    throw new BadRequest('account of cash box not specified', 'ERRORS.BAD_REQUEST');
+  }
+
+  params.user = req.session.user;  
+  let sumEntry = 0;
```
jniles: I don't think these variables are ever used again.

client/src/modules/reports/generate/cash_report/cash_report.html
```diff
@@ -0,0 +1,80 @@
...
+                    required>
+                    <option value="" disabled>{{ "FORM.SELECT.CASHBOX" | translate }}</option>
+                </select>
+                <div class="help-block" ng-messages="ConfigForm.cashbox.$error" ng-show="ConfigForm.$submitted">
+                <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+                </div>
+            </div>
+
+            <!-- Date interval  -->
+            <bh-date-interval
```
jniles, line 88: It would be really nice if this had a default set of dates.  It may just be on a slow internet connection, but when I load this on `staging.bhi.ma`, it doesn't have default dates.

client/src/modules/reports/generate/cash_report/cash_report.config.js
```diff
@@ -0,0 +1,74 @@
...
+  vm.clearPreview = function clearPreview() {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.preview = function preview(form) {
+    if (form.$invalid) { return; }
+
+    vm.reportDetails.account_id = vm.reportDetails.cashbox.account_id;
+    delete vm.reportDetails.cashbox;
```
jniles: This `delete` statement makes the validation get triggered for the form.  If you fill out a report, then cancel the previous, it will be red and angry.

server/models/bhima.sql
```diff
@@ -89,7 +89,8 @@ INSERT INTO `report` (`id`, `report_key`, `title_key`) VALUES
   (7, 'cashflowByService', 'REPORT.CASHFLOW_BY_SERVICE'),
   (8, 'openDebtors', 'REPORT.OPEN_DEBTORS.TITLE'),
   (9, 'clients_report','REPORT.CLIENTS_REPORT'),
-  (10, 'agedCreditors','REPORT.AGED_CREDITORS');
+  (10, 'agedCreditors','REPORT.AGED_CREDITORS'),
+  (11, 'cash_report', 'REPORT.CASH_REPORT');
```
jniles: `REPORT.CASH_REPORT` needs to be translated.  It currently is untranslated.

test/integration/reports/finance/cash_report.js
```diff
@@ -0,0 +1,35 @@
+/* global expect, chai, agent */
```
jniles, line 1: In the future, you can just write a single report test for this - that way all our updates to the report schema are made in the same place.

test/temporary-financial-reports/cashReport.js
```diff
@@ -0,0 +1,115 @@
...
+        return agent.get(reportURL)
+            .query(noAccountRequest)
+            .then((result) => {
+                const report = result.body;
+                expect(Number(report.status)).to.equal(400);
+            });
+    });
+
+    it('Throw an error when date are not provided', () => {
+        // see server/controllers/finance/reports/cashReport/index.js, document method
```
jniles, line 179: Nice comments!

server/controllers/finance/reports/cashReport/index.js
```diff
@@ -0,0 +1,188 @@
...
+      reportContext.expenses = expenses;
+
+      // Getting sum entries 
+      return db.one(aggregateRecordQuery('t.debit > 0'), [accountId, dateFrom, dateTo, accountId, dateFrom, dateTo]);      
+    })
+    .then((totalEntry) => {
+      reportContext.totalEntry = totalEntry.balance;
+
+      // Getting sum expenses 
+      return db.one(aggregateRecordQuery('t.credit > 0'), [accountId, dateFrom, dateTo, accountId, dateFrom, dateTo]);
```
jniles: There ~5 calls to posting_journal UNION general_ledger.  Since you always look at `account_id` , you could replace some of the sums on the `general_ledger` with `period_total`s?

jniles (commented):
@DedrickEnc, this is a pretty good PR!  I've made some UX comments.

Your tests are nice!

Please get this approved by @WayneNiles before it is merged.  We should be sure the logic is correct before this is deployed.

server/controllers/finance/reports/cashReport/index.js
DedrickEnc: I could if we were dealing with periods but the user can choose a custom date.

server/controllers/finance/reports/cashReport/index.js
DedrickEnc: You are rigth

client/src/modules/reports/generate/cash_report/cash_report.html
DedrickEnc: I don't catch very well, can you be more clear

DedrickEnc:
@jniles

DedrickEnc:
@jniles 
@WayneNiles approved this report, so can I get a review?

server/controllers/finance/reports/cashReport/index.js
```diff
@@ -0,0 +1,294 @@
...
+}
+
+
+function getRecordQuery(token, format, openingBalance) {
+  let query;
+
+  if (format === 1) {
+    query =
+      `
+    SET @cbal := ${openingBalance};
```
jniles: MySQL user-defined variables like `@cbal` are shared across connections.  This was the reason for bug https://github.com/IMA-WorldHealth/bhima-2.X/issues/1273.  If two reports are created simultaneously, the `@cbal` of one will destroy the `@cbal` of the other and a user will get an unpredictable result.  I worry this will have the same results.

The way to guarantee this will not occur is to initialize the logic defining `@cbal` _inside_ the SQL statement.  That way, `@cbal` is defined at the same time as it is used.  See https://github.com/IMA-WorldHealth/bhima-2.X/issues/1273 that fixed the above issue.

server/controllers/finance/reports/cashReport/index.js
```diff
@@ -0,0 +1,294 @@
...
+const BadRequest = require('../../../../lib/errors/BadRequest');
+
+const TEMPLATE_1 = './server/controllers/finance/reports/cashReport/report_combined.handlebars';
+const TEMPLATE_2 = './server/controllers/finance/reports/cashReport/report_separated.handlebars';
+
+// expose to the API
+exports.document = document;
+
+function getOpeningBalance(accountId, maxDate) {
+  const sql = `
```
jniles: For performance reasons, this would be much more efficient to use `period_total` to pick up the balances from period periods.  In fact, this function already exists in [server/controllers/finance/accounts/extra.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/accounts/extra.js#L124).  It has the signature `function getOpeningBalanceForDate(accountId, date)`.  It would be best to use that function since it will use `period_total` over `general_ledger` where possible.

server/controllers/finance/reports/cashReport/index.js
```diff
@@ -0,0 +1,294 @@
...
+ * @requires lib/ReportManager
+ * @requires lib/errors/BadRequest
+ */
+
+const _ = require('lodash');
+const db = require('../../../../lib/db');
+const ReportManager = require('../../../../lib/ReportManager');
+const BadRequest = require('../../../../lib/errors/BadRequest');
+
+const TEMPLATE_1 = './server/controllers/finance/reports/cashReport/report_combined.handlebars';
```
jniles: Can these be named something more meaningful?  Like `TEMPLATE_COMBINED`?  The name `TEMPLATE_1` doesn't tell a developer a lot about what it does.

server/controllers/finance/reports/cashReport/index.js
```diff
@@ -0,0 +1,294 @@
...
+
+      // Getting intermediate balance of cash account 
+      return db.one(aggregateRecordQuery(), [accountId, dateFrom, dateTo, accountId, dateFrom, dateTo]);
+    })
+    .then((intermediateTotal) => {
+      reportContext.intermediateTotal = intermediateTotal.algebricBalance;
+      // getting final balance of cash account
+      return db.one(aggregateRecordQuery(1, reportContext.openingBalance), [accountId, dateFrom, dateTo, accountId, dateFrom, dateTo]);
+    })
+    .then((finalTotal) => {
```
jniles: This last `.then()` doesn't do anything.  You can go ahead and remove it.

server/controllers/finance/reports/cashReport/report_separated.handlebars
```diff
@@ -0,0 +1,133 @@
...
+      <h6 class="text-center text-uppercase">
+        {{date this.dateFrom }} - {{date this.dateTo }}
+      </h6>
+
+      {{#if this.isEmpty}}
+        <div class="text-center">
+          {{translate 'FORM.INFO.NO_RECORDS_DATABASE'}}
+        </div>        
+      {{else}}
+        <div style="width: 100%" class="text-right">
```
jniles: Does this "width:100%" do anything?

server/lib/db/index.js
```diff
@@ -28,6 +28,7 @@ class DatabaseConnector {
   /** @constructor */
   constructor() {
     const params = {
+      multipleStatements: true,
```
jniles: I wondered how you had executed multiple statements above.  Please don't add this option - it slows down `node-mysql`'s query processing significantly.  It also opens us up to more types of SQL injection vulnerabilities, as warned in [the documentation](https://github.com/mysqljs/mysql#connection-options).

test/integration/reports/rendering.js
```diff
@@ -11,7 +11,7 @@ module.exports = function LazyTester(target, keys, params) {
     params = _.clone(params || {});
 
     // renders
-    const invalid = { renderer : 'unknown' };
+    const invalid = _.merge({ renderer : 'unknown' }, params);
```
jniles, line 14: Why do you need to specify parameters here?

client/src/modules/reports/generate/cash_report/cash_report.html
jniles: When I first loaded the page, the date fields were empty.  It would be cool if they defaulted to today.

jniles (changes requested):
@DedrickEnc, I've given the server side code a much closer look this time.

The running balance calculation has two problems: it should make sure the user-defined variable will not collide with another one, and it should be a single statement.  In my previous tests, `mysqljs`'s multipleStatements option will slow down all SQL queries because it is trying to figure out how many statements you are passing in.  Additionally, there are some security considerations to keep in mind.

I've also made some suggestions for reusing code to compute opening balances.

Let me know what you think.

server/controllers/finance/reports/cashReport/index.js
DedrickEnc: ~It is there to resolve the ```finalTotal``` promise, can you be more clear ?~

server/lib/db/index.js
DedrickEnc: Yes I agree

test/integration/reports/rendering.js
DedrickEnc: Because, I want to make sure, it is testing for an invalid renderer, if I don't specify parameters, cash report (which test the existence of parameters at the beginning) will throw an error for missing parameters but not for render.

client/src/modules/reports/generate/cash_report/cash_report.html
DedrickEnc: ok, I fixed it

DedrickEnc:
@jniles
Changes has been applied

server/controllers/finance/reports/cashReport/report_separated.handlebars
```diff
@@ -0,0 +1,133 @@
...
+                <th style="width: 10%"> {{ translate 'TABLE.COLUMNS.BY' }} </th>                            
+                <th style="width: 10%"> {{ translate 'TABLE.COLUMNS.AMOUNT' }} </th>
+              </tr>            
+            </thead>
+            <tbody>
+              {{#each this.expenses}}
+                <tr>
+                  <td class="text-left">{{ trans_id }}</td>
+                  <td class="text-left">{{ date trans_date }}</td>              
+                  <td class="text-left" style="white-space : nowrap;  text-overflow : ellipsis; overflow : hidden; max-width : 500px;">
```
jniles, line 91: At some point we should make this into a class - something like `text-overflow-hidden`.  You use it a lot.

server/controllers/finance/reports/cashReport/report_separated.handlebars
```diff
@@ -0,0 +1,133 @@
...
+
+      {{#if this.isEmpty}}
+        <div class="text-center">
+          {{translate 'FORM.INFO.NO_RECORDS_DATABASE'}}
+        </div>        
+      {{else}}
+        <div class="text-right">
+          <strong>{{ translate 'FORM.LABELS.OPENING_BALANCE' }} : {{currency this.openingBalance metadata.enterprise.curency_id}}</strong>
+        </div>
+        {{#if (isIncomeViewable this.type_id)}}
```
jniles, line 27: This is a nice boolean statement.  :+1:

server/controllers/finance/reports/cashReport/index.js
```diff
@@ -0,0 +1,280 @@
...
+                    SELECT
+                      g.trans_date, g.debit_equiv, g.credit_equiv, g.trans_id, g.record_uuid,
+                      g.description, g.account_id, (g.debit_equiv - g.credit_equiv) AS balance
+                    FROM
+                      general_ledger AS g 
+                    WHERE 
+                      g.account_id= ? AND 
+                      (g.trans_date >= DATE(?) AND g.trans_date <= DATE(?))
+                  )
+              ) AS u, (SELECT @cbal := ${openingBalance}
```
jniles, line 66: :+1: Well done!  This is a pretty complicated thing to do, and you've done it well.

jniles (approved):
Look good to me for an initial PR!

PR #1838, opened 2017-07-11
fix(Latest invoice)

### LINKED ISSUE #1837: (bug) Latest invoice statement has no loading indicator
At a number of installations with limited hardware the latest invoice statement (on the patient record module) can take up to 20 seconds - this query will need to be optimised. 

Right now a partially filled in component is displayed without any indication of loading. It is important to hide the components display until the content is loaded, displaying a loading indicator would also help the user know that the data is not yet ready.

lomamech:
- hide the components display until the content is loaded
- refactor the function startup With the combination of the search of the
  balance as well as the information on the last invoice

closes #1837

### DIFF
--- a/client/src/modules/patients/record/bh-latest-patient-invoice.component.js
+++ b/client/src/modules/patients/record/bh-latest-patient-invoice.component.js
@@ -9,13 +9,13 @@ angular.module('bhima.components')
 });
 
 LatestInvoice.$inject = [
-  'PatientService', 'moment', 'NotifyService', 'SessionService',
+  'PatientService', 'moment', 'NotifyService', 'SessionService', '$q'
 ];
 
 /**
  * This component is responsible for displaying the Latest Invoice
  */
-function LatestInvoice(Patient, moment, Notify, Session) {
+function LatestInvoice(Patient, moment, Notify, Session, $q) {
   var vm = this;
 
   this.$onInit = function $onInit() {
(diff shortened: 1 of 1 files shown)


jniles (commented):
This is a pretty good start!

In my view, if we are worried about requests taking a long time to load, we should run them in parallel.  Like this:

```js
var requests = $q.all([
  Patient.latest(vm.debtorUuid),
  Patient.balance(vm.debtorUuid)
]);

requests.then(function (results) {
  vm.patientInvoice = results[0];  // returned from the latest() call
  vm.balance = results[1]; // returned from balance() call
  // etc ...
})
 .catch(Notify.handleError)
 .finally(function () {
   vm.loading = false;
  });
```

The advantage here is that we can then fire off two requests at once.  If the problem is that this request takes a long time to load, we should make it as soon as possible.

@sfount, what do you think?

client/src/modules/patients/record/bh-latest-patient-invoice.component.js
```diff
@@ -29,21 +29,21 @@ function LatestInvoice(Patient, moment, Notify, Session) {
...
-
-    Patient.balance(vm.debtorUuid)
-      .then(function (balance) {
-        vm.patientBalance = balance;
-      })
-      .catch(Notify.handleError)
-      .finally(function () {
-        vm.loading = false;
-      });      
+    var requests = $q.all([
```
jniles, line 32: :+1:

jniles:
I'll test this locally and, if it all looks good, I'll merge it.

jniles:
LGTM!  bors r+

PR #1844, opened 2017-07-12
fix(Update user permissions)

### LINKED ISSUE #1819: (bug) Updated permissions do not refresh until the session is destroyed
When permissions to new modules are granted to a user, the permissions will not be applied until the user has destroyed their current session and opened a new one (new tab or logged out logged back in). 

**Example:** 
User 'user' is currently logged in, he is granted access to 'Module' 

'user' clicks on the 'Module' in the tree. 

**Expected behaviour:** 
User currently has access to 'Module', module should load.

**Current behaviour:** 
User currently has access to 'Module', 403 page restricted message shows because the permissions have been recently updated. 

This is relevant to work on the users and permissions module found https://github.com/IMA-WorldHealth/bhima-2.X/issues/166

lomamech:
- Enable instant updating of user access rights
closes #1819

### DIFF
--- a/client/src/modules/users/userPermission.modal.js
+++ b/client/src/modules/users/userPermission.modal.js
@@ -2,11 +2,11 @@ angular.module('bhima.controllers')
   .controller('UserPermissionModalController', UserPermissionModalController);
 
 UserPermissionModalController.$inject = [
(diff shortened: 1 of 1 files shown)


jniles (approved):
LGTM.  Great re-use of code!

I'll let @sfount merge - I'm not sure of the deployment status at HEV.

PR #1855, opened 2017-07-13
refactor the report to render amount with enterprise currency

DedrickEnc:
This P.R fix an issue relative to the currency, now the income expense can render amount in the enterprise currency and not a default one.
inc-exp-currency

### DIFF
--- a/server/controllers/finance/reports/incomeExpense/report.handlebars
+++ b/server/controllers/finance/reports/incomeExpense/report.handlebars
@@ -38,8 +38,8 @@
                 <tr>
                   <td>{{number}}</td>
                   <td>{{label}}</td>
-                  <td class="text-right">{{currency debit}}</td>
-                  <td class="text-right">{{currency credit}}</td>
+                  <td class="text-right">{{currency debit ../enterprise.currency_id}}</td>
+                  <td class="text-right">{{currency credit ../enterprise.currency_id}}</td>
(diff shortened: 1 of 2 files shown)


DedrickEnc:
@sfount 
The changes are landed

sfount (approved):
Code LGTM 👍 Great quick fix.

DedrickEnc:
@sfount
can you merge it?

server/controllers/finance/reports/incomeExpense/report.handlebars
```diff
@@ -38,16 +38,16 @@
                 <tr>
                   <td>{{number}}</td>
                   <td>{{label}}</td>
-                  <td class="text-right">{{currency debit}}</td>
-                  <td class="text-right">{{currency credit}}</td>
+                  <td class="text-right">{{currency debit ../enterprise.currency_id}}</td>
```
jniles, line 41: Just FYI, I believe the enterprise is always available on the `metadata` object.  For example, 
```html
<td class="text-right">{{currency debit metadata.enterprise.currency_id}}</td>
```

You can see this usage in [the invoices report](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/reports/invoices/report.handlebars#L45).

jniles (approved):

PR #1870, opened 2017-07-16
test(lint): only lint changed files

jniles:
This commit implements a selective-linting feature that only shows the
lint results for the changes made since master.  It assumes that the
local master is up to date with the remote master.

Once all lint errors are fixed on master, we can enforce linting as a test on
Travis CI.

### DIFF
--- a/sh/lint.sh
+++ b/sh/lint.sh
@@ -1,12 +1,20 @@
 #!/bin/bash
+##
+## Lint Test
+##
+
+# bash 'strict mode'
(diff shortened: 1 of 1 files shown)


DedrickEnc:
@jniles the test does not pass

jniles:
Closing this as probably better for a precommit hook (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1904) rather than Travis.

PR #1876, opened 2017-07-18
fix(reports): hack to refresh reportData

### LINKED ISSUE #1821: (bug) Reports module does not update report definition on transition
Currently the reports module structure (title and archive button) is in an external `reports.html`. This allows the structure to share the report definition details with the report generation template (in a `ui-view`). 

The report details are not updated or refreshed when the user transitions to another report. This leaves invalid archive links as well as the incorrect title displayed for the report.

jniles:
~~This commit implements a $rootScope hack to refresh the $state's reportData.~~  [See updated description](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1876#issuecomment-318047514)

Closes #1821.

### DIFF
--- a/client/src/modules/reports/reports.js
+++ b/client/src/modules/reports/reports.js
@@ -2,10 +2,10 @@ angular.module('bhima.controllers')
   .controller('ReportsController', ReportsController);
 
 ReportsController.$inject = [
-  '$state', 'reportData',
+  '$state', 'reportData', '$scope', 'BaseReportService'
 ];
 
-function ReportsController($state, reportData) {
+function ReportsController($state, reportData, $scope, SavedReports) {
   var vm = this;
   var archiveState = 'reportsBase.reportsArchive';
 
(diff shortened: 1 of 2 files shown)


jniles:
@sfount, what are your thoughts on this solution to get the report name to change with $state routing?

sfount:
I've tested this fix branch locally and it seems that the report title still lags one behind the selection? I'm not sure if this is just on my machine or if something is logically one step behind. 

I'll have a closer look at the code later to see if I can find what this would be.

jniles:
@sfount, noted.  I'll take a look.

jniles:
@sfount, implemented what is probably a cleaner solution.  This one actually works as expected.  Sorry for not thoroughly testing. :(

The `onEnter` event was not firing with updated parameters.  In this implementation, I've opted to listen on `$stateChangeSuccess` using `$scope`.

Advantages:
 1. `$scope` is automatically cleaned up when the controller is destroyed, and all events/listeners are unbound - `$rootScope` is not (I had a memory leak in the previous implementation).
 2. No need to edit the `routes.js` file.
 3. Only fires on `$state` changes that are for child reports.
 4. Less magic, less code.

Downsides:
 1. I think `$stateChangeSuccess` is going to be removed in upcoming ui-router implementations.

jniles:
Ping @sfount

sfount:
This will help reduce confusion for users significantly and is needed right now, interesting patch! In the next few weeks we should investigate how this can be incorporated into the routing logic the code is easier to follow.

PR #1880, opened 2017-07-19
refactor(Debtor group list display)

### LINKED ISSUE #1857: refactor: debtor group list display
In my mind, the debtor group list should reverse the order of display of the "Locked" and "Color" properties.  See below.

debtorgroupmodifyingflags
_Fig 1: Flags Modifying the Debtor Groups_

In my mind, this doesn't look very clean.  Some group names are shifted Since each Debtor Group has a color by default, I would list these first, followed by the name, then the locked status.
 1. The first letter of each name will line up regardless of locked status.
 2. The color is then seen as a fundamental part of the group
 3. The locked status is more of a flag applied to the group.

lomamech:
- Change the display of Debtor Group, display colour followed by the name,
  then the status

closes #1857

### DIFF
--- a/client/src/modules/debtors/groups.list.html
+++ b/client/src/modules/debtors/groups.list.html
@@ -40,8 +40,8 @@
             <div class="row">
               <div class="col-md-6">
                 <h4 style="margin-top : 0px;">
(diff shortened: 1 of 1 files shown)


jniles (approved):
LGTM!

jniles:
bors  r+

PR #1881, opened 2017-07-19
fix(build) Resolve uglify syntax error in employee

sfount:
This commit removes a trailing comma in a function definition
that forces the uglify (minify) JS step of the build process
to crash on production servers.


jniles:
LGTM

jniles:
Bors r+

jniles:
bors  r+

PR #1885, opened 2017-07-21
Feat(stock consumption): Tracking stock consumption and compute the average consumption

mbayopanda:
This PR is about : 

- Tracking stock consumption 
    - for client distributions
    - for service distributions

- Enhancing the stock inventories report

- Fix the `bhDepotDropdown` component


mbayopanda:
@sfount, @jniles could you review this PR ?

client/src/js/components/bhDepotDropdown.js
```diff
@@ -35,6 +36,13 @@ function bhDepotController(Depot, AppCache, Notify) {
   $ctrl.select = function select(option) {
     $ctrl.selection = option;
     cache.selection = $ctrl.selection;
-    $ctrl.onSelect({ depot: $ctrl.selection });
+    $ctrl.onSelect({ depot : $ctrl.selection });
   };
+
+  function cachedDepotExist() {
+    if (!cache.selection) { return false; }
+    return $ctrl.depots.map(function (item) {
```
jniles: Hmm... while this looks cool, it is _really_ hard to read.  You probably want a `filter`.  It may even be faster.
```js
// returns true if the cached uuid matches a depot uuid
return $ctrl.depots.filter(function (depot) {
  return depot.uuid === cache.selection.uuid;
});
```

client/src/js/components/bhDepotDropdown.js
```diff
@@ -22,9 +22,10 @@ function bhDepotController(Depot, AppCache, Notify) {
     .then(function (rows) {
       if (!rows.length) { return; }
       $ctrl.depots = rows;
-      $ctrl.selection = cache.selection || $ctrl.depots[0];
+      $ctrl.depotsAvailable = !!rows.length;
```
jniles: This is really nit-picky, but it is helpful to name boolean variables with `is` or `has` as the prefix to show they are boolean.  In this case, it would be nice to call this `$ctrl.hasDepotsAvailable`.

The reason to do this is because `is` or `has` expects a true/false response.  `isOkay?` can only be a "yes" or "no".  Similarly, `hasDepotsAvailable?` is either a "yes" or "no".  However, `depotsAvailable` could easily be an array of depots which are available to use.

Just best practice :)

client/src/js/components/bhDepotDropdown.js
```diff
@@ -22,9 +22,10 @@ function bhDepotController(Depot, AppCache, Notify) {
...
+      $ctrl.selection = cachedDepotExist() ? cache.selection : $ctrl.depots[0];
       $ctrl.loading = false;
```
jniles: I know this is old code, but you do not need the `$ctrl.loading` line - it will be executed in the `.finally()` method.

server/controllers/stock/core.js
```diff
@@ -233,7 +240,7 @@ function stockManagementProcess(inventories) {
     inventory.S_SEC = CM * inventory.delay; // stock de securite
     inventory.S_MIN = inventory.S_SEC * 2; // stock minimum
     inventory.S_MAX = (CM * inventory.purchase_interval) + inventory.S_MIN; // stock maximum
-    inventory.S_MONTH = inventory.quantity / CM; // mois de stock
+    inventory.S_MONTH = Math.round(inventory.quantity / CM); // mois de stock
```
jniles: ... I think this should be `Math.floor()`.  I'm not sure how it is used, but we do not want to tell an administrator that they have 8 months of stock, if they only have 7.8 months of stock.  Patients could die!

server/controllers/stock/core.js
```diff
@@ -259,3 +266,135 @@ function stockManagementProcess(inventories) {
...
+ * @function getStockConsumptionAverage
+ *
+ * @description returns average of stock consumption (CMM)
+ *
+ * @param {number} periodId - the base period
+ *
+ * @param {number} numberOfMonths - the number of months for calculating the average
+ */
+function getStockConsumptionAverage(periodId, numberOfMonths) {
+  const baseAvgNumberOfMonths = numberOfMonths || 6;
```
jniles: You could probably move this `6` into a global variable
```js
const BASE_NUMBER_OF_MONTHS = 6;

function getStockConsumptionAvg() { /* ... */ }
```

server/controllers/stock/core.js
```diff
@@ -259,3 +266,135 @@ function stockManagementProcess(inventories) {
...
+    SELECT ROUND(AVG(s.quantity)) AS quantity, BUID(i.uuid) AS uuid, i.text, i.code, BUID(d.uuid) AS depot_uuid, d.text AS depot_text
+    FROM stock_consumption s
+    JOIN inventory i ON i.uuid = s.inventory_uuid 
+    JOIN depot d ON d.uuid = s.depot_uuid
+    JOIN period p ON p.id = s.period_id
+    WHERE p.id IN (?) 
+    GROUP BY i.uuid, d.uuid
+  `;
+
+  return db.one(queryPeriodId, [periodId])
```
jniles: Are you sure you want `db.one()` here?  It looks like it might return multiple rows.  Double check that this still works the way you want it with multiple consumption events.

server/controllers/stock/core.js
```diff
@@ -259,3 +266,135 @@ function stockManagementProcess(inventories) {
...
+    .then((period) => {
+      const beginingPeriod = period.id - baseAvgNumberOfMonths;
+      const paramPeriodRange = beginingPeriod > 0 ? [beginingPeriod + 1, period.id] : [1, period.id];
+      return db.exec(queryPeriodRange, paramPeriodRange);
+    })
+    .then((rows) => {
+      const ids = rows.map(row => {
```
jniles: You can actually write this on one line!
```js
const ids = rows.map(row => row.id);
```

server/controllers/stock/core.js
```diff
@@ -259,3 +266,135 @@ function stockManagementProcess(inventories) {
...
+        return row.id;
+      });
+
+      return db.exec(queryStockConsumption, [ids]);
+    })
+    .then((rows) => {
```
jniles: This `.then()` doesn't do anything.  You can remove it.

server/controllers/stock/core.js
```diff
@@ -259,3 +266,135 @@ function stockManagementProcess(inventories) {
...
+          }
+        }
+      }
+
+      return bundle.inventories;
+    })
+    .then(stockManagementProcess)
+    .then((rows) => {
+      if (status) {
+        return rows.filter((row) => {
```
jniles: This could also be written on a single line!
```js
return rows.filter(row => row.status === status);
```

server/controllers/stock/index.js
```diff
@@ -153,8 +158,18 @@ function normalMovement(document, params) {
...
     transaction.addQuery(createMovementQuery, [createMovementObject]);
+
+    isDistributable = !!(parameters.flux_id === core.flux.TO_PATIENT || parameters.flux_id === core.flux.TO_SERVICE);
+
+    // track distribution to patient
+    if (parameters.is_exit && isDistributable) {
+      const consumptionParams = [
+        db.bid(lot.inventory_uuid), db.bid(parameters.depot_uuid), document.date, lot.quantity,
+      ];
+      transaction.addQuery('CALL ComputeStockConsumptionByDate(?, ?, ?, ?)', consumptionParams);
```
jniles, line 170: This is cool 👍

server/controllers/stock/reports/stock_inventories.report.handlebars
```diff
@@ -28,71 +32,93 @@
...
+                  {{#equal status 'minimum_reached'}}{{translate 'STOCK.STATUS.MINIMUM'}}{{/equal}}
+                  {{#equal status 'over_maximum'}}{{translate 'STOCK.STATUS.OVER_MAX'}}{{/equal}}
+                </td>
+                <td class="text-right">{{avg_consumption}}</td>
+                <td class="text-right">{{S_MONTH}}</td>
+                <td class="text-right">{{S_SEC}}</td>
+                <td class="text-right">{{S_MIN}}</td>
+                <td class="text-right">{{S_MAX}}</td>
+                <td class="text-right">
+                  {{#gt S_Q 0}}<span>{{S_Q}}</span>{{else}} 0 {{/gt}}
```
jniles: This is kind of hard to read - can you move this logic into the JS?  That way you can just write:
```html
{{ S_Q }}
```

server/controllers/stock/reports/stock_inventories.report.handlebars
```diff
@@ -28,71 +32,93 @@
...
+                </td>
+              </tr>
+            {{/each}}
+
+             <!-- blank row  -->
+             {{#unless @last }}
```
jniles: 👍

jniles (commented):
@mbayopanda, I've made a few comments.  I haven't pulled this to test the functionality yet - but you can run that by @sfount and @WayneNiles.

Let me know what you think.

client/src/js/components/bhDepotDropdown.js
mbayopanda: `$ctrl.depots.filter()` will return an array.

you want say : 
```javascript
// returns true if the cached uuid matches a depot uuid
return $ctrl.depots.filter(function (depot) {
  return depot.uuid === cache.selection.uuid;
}).length;
```

server/controllers/stock/core.js
mbayopanda: 👍 you're right...

server/controllers/stock/core.js
mbayopanda: We need just one period, this query is for verifying that the period exit really. Probably i have to add the limit `LIMIT 1` to the query.

server/controllers/stock/core.js
mbayopanda: 👍

server/controllers/stock/core.js
mbayopanda: Thanks 👍

client/src/js/components/bhDepotDropdown.js
jniles: Ah, I misunderstood the code.  You are right ... but a Array.some() is even better!

mbayopanda:
You're right... I had used Array.filter().length

2017-07-23 7:35 GMT-07:00 Jonathan Niles <notifications@github.com>:

jniles:
@mbayopanda, that's fine.  It's a non-blocking comment.

I don't have time to test the logic here ... you should run this by @WayneNiles to make sure it all works as expected.

mbayopanda:
I did show him the report, he's okay with that and he suggest me to show it
to a stock manager such as Jean Claude.

2017-07-23 9:06 GMT-07:00 Jonathan Niles <notifications@github.com>:

jniles:
@mbayopanda, let me know when you need another review.

mbayopanda:
@jniles could you review this again please

client/src/i18n/en/report.json
```diff
@@ -78,14 +78,26 @@
...
     },
+    "STOCK":{
+      "TITLE":"Stock Report",
+      "DESCRIPTION":"This report show stock of products, their average consumptions and their status",
+      "ALL_DEPOTS":"All Depots",
+      "ONE_DEPOT":"One Specific Depot",
+      "ALL_INVENTORIES":"All Inventories",
+      "ONE_INVENTORY":"One Specific Inventory",
+      "INVENTORY_REPORT":"Inventory Report",
+      "INVENTORY_REPORT_DESCRIPTION":"This report show all movement of a product and it stock status and value",
```
jniles: Just a few typos.  This should read:
> This report shows all movements of a product and its stock status and value.

client/src/i18n/en/report.json
```diff
@@ -78,14 +78,26 @@
...
-      "TITLE": "Income Expense Report",
-      "DESCRIPTION": "This report show all income and expense occured in the enterprise"
-    },
     "CASH_REPORT": {
       "TITLE": "Cash Report",
       "DESCRIPTION": "This report show all entries and exits of cash inside the enterprise"
     },
+    "STOCK":{
+      "TITLE":"Stock Report",
+      "DESCRIPTION":"This report show stock of products, their average consumptions and their status",
```
jniles: Just a few typos. This should read:
> This report shows the stock of products, their average consumption and their status.

client/src/i18n/en/stock.json
```diff
@@ -46,7 +46,8 @@
     "MAXIMUM"         : "Maximum Stock",
     "MINIMUM"         : "Minimum Stock",
     "MONTH"           : "Stock Months",
-    "MONTHLY_CONSUM"  : "Month Consumptions",
+    "MONTH_OF_STOCK"  : "Stock Months - Months before sold out",
```
jniles: A better way of saying this is 
> Stock Months - Months until Stock Out

client/src/i18n/en/stock.json
```diff
@@ -46,7 +46,8 @@
...
+    "MONTHLY_CONSUM"  : "Monthly Average Consumption",
```
jniles: In English, it makes more sense to reverse the order of these adjectives:
> Average Monthly Consumption

client/src/js/components/bhDepotSelect.js
```diff
@@ -0,0 +1,30 @@
+angular.module('bhima.components')
+  .component('bhDepotSelect', {
+    bindings : {
+      onSelect : '&',
+    },
+    templateUrl  : 'modules/templates/bhDepotSelect.tmpl.html',
+    controller   : bhDepotSelectController,
+    controllerAs : '$ctrl',
```
jniles: All components are automatically bound with `controllerAs: '$ctrl'`.  You do not need to specify it.  [For reference, here is the documentation for Angular's component() method](https://docs.angularjs.org/guide/component#comparison-between-directive-definition-and-component-definition).

client/src/js/components/bhDepotSelect.js
```diff
@@ -0,0 +1,30 @@
...
+
+function bhDepotSelectController(Depot, Notify) {
+  var $ctrl = this;
+
+  $ctrl.$onInit = function $onInit() {
+    Depot.read()
+    .then(function (rows) {
+      if (!rows.length) { return; }
+      $ctrl.depots = rows;
+      $ctrl.onSelect({ depot : $ctrl.selection });
```
jniles: When does `$ctrl.selection` get defined?

client/src/modules/reports/generate/inventory_report/inventory_report.config.js
```diff
@@ -0,0 +1,95 @@
+angular.module('bhima.controllers')
+  .controller('inventory_reportController', InventoryReportConfigController);
+
+InventoryReportConfigController.$inject = [
```
jniles, line 4: I can't find where you use `$timeout` in this module.  Could you make sure you clean out all unused dependencies?  They make it harder to understand what the code is supposed to do.

client/src/modules/reports/generate/inventory_report/inventory_report.config.js
```diff
@@ -0,0 +1,95 @@
...
+
+    return SavedReports.saveAsModal(options)
+      .then(function () {
+        $state.go('reportsBase.reportsArchive', { key : options.report.report_key });
+      })
+      .catch(Notify.handleError);
+  };
+
+  function checkCachedConfiguration() {
+    vm.reportDetails = cache.reportDetails ? angular.copy(cache.reportDetails) : {};
```
jniles: Just FYI, I think you could actually do this same thing like this:
```js
vm.reportDetails =  angular.copy(cache.reportDetails || {});
```

client/src/modules/reports/generate/inventory_report/inventory_report.html
```diff
@@ -0,0 +1,101 @@
...
+          <div class="form-group col-xs-12 col-sm-6">
+            <label class="control-label" translate>FORM.LABELS.PURCHASE_INTERVAL</label>
+            <div class="input-group">
+              <input class="form-control" type="number" step="1" name="purchase_interval" ng-model="ReportConfigCtrl.purchaseInterval">
+              <span class="input-group-addon" translate>FORM.LABELS.MONTH</span>
+            </div>
+          </div>
+
+          <!-- depot options -->
+          <div class="form-group">
```
jniles: The correct way to specify a bootstrap checkbox  is demonstrated in the [documentation](http://getbootstrap.com/css/#checkboxes-and-radios).

```html
<div class="checkbox">
  <label>
    <input type="checkbox" value="">
    Option one is this and that&mdash;be sure to include why it's great
  </label>
</div>
```

This way causes the `form-group`'s label styling to by applied to the `<label>` and text doesn't always line up correctly.

client/src/modules/reports/generate/inventory_file/inventory_file.config.js
```diff
@@ -0,0 +1,87 @@
...
+  vm.clearPreview = function clearPreview() {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.preview = function preview(form) {
+    var options;
+    var params;
+
+    if (form.$invalid || !vm.depot || !vm.inventory) { return 0; }
```
jniles: Can you not make `vm.depot` or `vm.inventory` required, and then just use `form.$invalid`?

client/src/modules/templates/bhDepotDropdown.tmpl.html
```diff
@@ -1,4 +1,4 @@
-<div class="btn-group" uib-dropdown>
+<div ng-if="$ctrl.hasDepotsAvailable" class="btn-group" uib-dropdown>
```
jniles, line 1: :+1:

server/controllers/stock/core.js
```diff
@@ -259,3 +271,231 @@ function stockManagementProcess(inventories) {
...
+        if (line.is_exit) {
+          stockQuantity -= line.quantity;
+          stockValue = stockQuantity * stockUnitCost;
+
+          // exit
+          movement.exit.quantity = line.quantity;
+          movement.exit.unit_cost = stockUnitCost;
+          movement.exit.value = line.quantity * line.unit_cost;
+
+          // stock status
```
jniles, line 473: It looks like you have repeated these lines of code (after `// stock status`).  Could you move those out of the `if/else` block?  It would make the code easier to read to only putting logic that depends on the `if/else` in the `if/else` block.

server/controllers/stock/reports/index.js
```diff
@@ -594,7 +595,12 @@ function stockLotsReport(req, res, next) {
 
   const data = {};
   let report;
-  const optionReport = _.extend(req.query, { filename : 'TREE.STOCK_LOTS', orientation : 'landscape' });
+  const optionReport = _.extend(req.query, {
+    filename : 'TREE.STOCK_LOTS',
+    orientation : 'landscape',
+    footerRight : '[page] / [toPage]',
```
jniles, line 601: :+1:

server/controllers/stock/reports/index.js
```diff
@@ -663,6 +689,16 @@ function stockMovementsReport(req, res, next) {
       data.csv = rows;
       data.display = display;
 
+      // group by depot
+      let depots = _.groupBy(rows, d => d.depot_text);
```
jniles, line 693: At some point, we will likely have move this into SQL for performance reasons.  It would be good to put a `// TODO` here so that we can easily locate it.

server/controllers/stock/reports/index.js
```diff
@@ -663,6 +689,16 @@ function stockMovementsReport(req, res, next) {
...
+
+      // make sure that they keys are sorted in alphabetical order
+      depots = _.mapValues(depots, lines => {
```
jniles, line 696: I've never done it, but you could probably use `_.chain()` to make the code look fluid!  Something like this:
```js
const depots = _.chain(depots)
  .groupBy(rows, d => d.depot_text)
  .mapValues(depots, lines => _.sortBy(lines, 'depot_text'))
  .value();
```

Just FYI. :)

server/controllers/stock/reports/index.js
```diff
@@ -685,31 +721,56 @@ function stockInventoriesReport(req, res, next) {
...
+    footerFontSize : '8',
+  });
 
   // set up the report with report manager
   try {
     if (req.query.identifiers && req.query.display) {
       options = JSON.parse(req.query.identifiers);
       display = JSON.parse(req.query.display);
       hasFilter = Object.keys(display).length > 0;
+    } else if (req.query.params) {
```
jniles, line 742: Why do you have a separate `req.query.params`?  Can we not put this all in `req.query`?

jniles (commented):
@mbayopanda, looks pretty good!  I've made a few final comments to touch up the translations and code a bit more.  Let me know what you think.

client/src/js/components/bhDepotSelect.js
mbayopanda: the selection is defined in the view by a `ng-model="$ctrl.selection"`

jniles:
@mbayopanda, there is a conflict in the translation files.  Could you fix these?

mbayopanda:

jniles:
@mbayopanda, I already did [thirteen days ago](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1885#pullrequestreview-52941046).  Did you make changes based on that?

mbayopanda:
@jniles I rebase the branch according the last version of master

client/src/i18n/en/report.json
```diff
@@ -76,15 +76,31 @@
...
+      "DESCRIPTION": "This report shows all incomes and expenses of the enterprise"
     },
     "CASH_REPORT": {
       "TITLE": "Cash Report",
-      "DESCRIPTION": "This report show all entries and exits of cash inside the enterprise"
+      "DESCRIPTION": "This report shows all entries and exits of cash inside the enterprise"
+    },
+    "STOCK":{
+      "TITLE":"Stock Report",
+      "DESCRIPTION":"This report shows the stock of products, their average consumption and their status.",
```
jniles, line 91: :+1:  Much better!

client/src/js/components/bhDepotSelect.js
```diff
@@ -21,20 +21,20 @@ DepotSelectController.$inject = [
...
-      .then(function (depots) {        
+      .then(function (depots) {
         $ctrl.depots = depots;
       })
       .catch(Notify.handleError);
   };
 
   // fires the onSelectCallback bound to the component boundary
-  $ctrl.onSelect = function ($item, $model) {
+  $ctrl.onSelect = function ($item) {
```
jniles, line 37: :+1:

client/src/modules/reports/generate/inventory_file/inventory_file.config.js
```diff
@@ -0,0 +1,89 @@
...
+  var reportUrl = 'reports/stock/inventory';
+
+  vm.previewGenerated = false;
+
+  vm.dateTo = new Date();
+
+  // chech cached configuration
+  checkCachedConfiguration();
+
+  vm.setupDepot = function setupDepot(depot) {
```
jniles: In the future, since these are callbacks to components, it would be useful to label them `onChangeDepot` or `onSelectDepot` or something similar.  It signals to the programmer that this is called when an event is fired (when the depot changed, for example).

client/src/modules/reports/generate/inventory_file/inventory_file.config.js
```diff
@@ -0,0 +1,89 @@
...
+      depot_uuid : vm.depot.uuid,
+      inventory_uuid : vm.inventory.uuid,
+      dateTo : vm.dateTo,
+    };
+
+     // update cached configuration
+    cache.reportDetails = angular.copy(params);
+
+    // format date for the server
+    params.dateTo = moment(params.dateTo).format('YYYY-MM-DD');
```
jniles, line 55: In the future, if we are going to stringify our dates before sending them to the server, we should do it all in the same spot.  `$httpParamSerializer` is the best place to detect if the parameter is a date and then stringify it appropriately.

This should be discussed first though.

client/src/modules/reports/generate/inventory_file/inventory_file.html
```diff
@@ -0,0 +1,60 @@
...
+            inventory-uuid="ReportConfigCtrl.inventory.uuid"
+            on-select-callback="ReportConfigCtrl.setupInventory(inventory)"
+            required="true"
+            validate-trigger="ConfigForm.$submitted">
+            <bh-clear on-clear="ReportConfigCtrl.clear('inventory')"></bh-clear>
+          </bh-inventory-select>
+
+          <!-- choose date until  -->
+          <bh-date-editor
+            label="FORM.LABELS.UNTIL_PERIOD"
```
jniles: Why does this say "UNTIL_PERIOD" but allow a user to select a date?  Should it not allow a user to selected a period?

client/src/modules/reports/generate/inventory_report/inventory_report.html
```diff
@@ -0,0 +1,93 @@
...
+            <div class="input-group">
+              <input class="form-control" type="number" step="1" name="purchase_interval" ng-model="ReportConfigCtrl.purchaseInterval">
+              <span class="input-group-addon" translate>FORM.LABELS.MONTH</span>
+            </div>
+          </div>
+
+          <!-- depot options -->
+          <div class="checkbox">
+            <label>
+              <input type="checkbox" ng-model="ReportConfigCtrl.chooseOneDepot" ng-true-value="1" ng-false-value="0">
```
jniles: Since you aren't sending these parameters back to the database, you can just use `true` and `false` without needing to make it 1 or 0.

client/src/modules/reports/generate/inventory_report/inventory_report.html
```diff
@@ -0,0 +1,93 @@
...
+              <span translate>REPORT.STOCK.ONE_DEPOT</span>
+            </label>
+          </div>
+
+          <!-- select depot -->
+          <div ng-if="!!ReportConfigCtrl.chooseOneDepot">
```
jniles, line 74: I think both the `<bh-depot-select>` and the `<bh-inventory-select>` should be required if they are shown.

Without them required, this bug exists:
 1. Navigate to this page.
 2. Check both the checkboxes to show the `select`s.
 3. Click Submit 
 4. Observe that the button loads forever.  The console reveals:
> TypeError: Cannot read property 'uuid' of undefined

client/src/modules/reports/generate/inventory_report/inventory_report.html
```diff
@@ -0,0 +1,93 @@
...
+              <input class="form-control" type="number" step="1" name="delay" ng-model="ReportConfigCtrl.delay">
+              <span class="input-group-addon" translate>FORM.LABELS.MONTH</span>
+            </div>
+          </div>
+
+          <!-- purchase interval -->
+          <div class="form-group col-xs-12 col-sm-6">
+            <label class="control-label" translate>FORM.LABELS.PURCHASE_INTERVAL</label>
+            <div class="input-group">
+              <input class="form-control" type="number" step="1" name="purchase_interval" ng-model="ReportConfigCtrl.purchaseInterval">
```
jniles: The `step=1` doesn't do anything for non-range `<input>`s.   You can remove it.

client/src/modules/stock/inventories/templates/appro.cell.html
```diff
@@ -1,9 +1,9 @@
 <div class="ui-grid-cell-contents">
-    <span ng-show="row.entity.S_Q < 0" class="label label-success" translate>
+    <span ng-show="row.entity.S_Q <= 0" class="label label-success" translate>
```
jniles, line 2: Nice catch :+1:

server/controllers/stock/core.js
```diff
@@ -269,3 +281,231 @@ function stockManagementProcess(inventories) {
...
+
+  const baseDate = periodDate ? moment(periodDate).format(DATE_FORMAT) : moment().format(DATE_FORMAT);
+
+  const queryPeriodRange = `
+    SELECT id FROM period WHERE id BETWEEN ? AND ?;
+  `;
+
+  const queryPeriodId = periodId ?
+    'SELECT id FROM period WHERE id = ? LIMIT 1;' :
+    'SELECT id FROM period WHERE DATE(?) BETWEEN DATE(start_date) AND DATE(end_date) LIMIT 1;';
```
jniles, line 327: If you want to make this bullet proof, it should use the Enterprise ID to make sure BHIMA works correctly with multiple enterprises  ... but I guess this is okay for now.

server/controllers/stock/core.js
```diff
@@ -269,3 +281,231 @@ function stockManagementProcess(inventories) {
...
+      bundle.inventories = rows;
+      return getStockConsumptionAverage(null, params.dateTo);
+    })
+    .then((rows) => {
+      var sameInventory;
+      var sameDepot;
+
+      bundle.consumptions = rows;
+
+      for (let i = 0; i < bundle.consumptions.length; i++) {
```
jniles, line 404: Is there any possible way this could be done in SQL?

It seems to me "same inventory" and "same depot" is English for "SQL JOIN".

server/controllers/stock/index.js
```diff
@@ -48,13 +52,13 @@ function createStock(req, res, next) {
...
     uuid : uuid.v4(),
-    date : moment(new Date(params.date)).format('YYYY-MM-DD').toString(),
+    date : moment(new Date(params.date)).format('YYYY-MM-DD HH:mm:ss').toString(),
     user : req.session.user.id,
   };
 
   params.lots.forEach((lot) => {
         // lot expiration date
-    date = moment(new Date(lot.expiration_date)).format('YYYY-MM-DD').toString();
+    date = moment(new Date(lot.expiration_date)).format('YYYY-MM-DD HH:mm:ss').toString();
```
jniles: Why can't this code use MySQL's date functions?

On my machine:
```js
const mysql = require('mysql');
mysql.escape(new Date());
// --> '\'2017-08-11 08:17:57.930\''
```

In my mind, the code is a lot clearer if you write:
```js
date = new Date(lot.expiration_date);
```
then if you write:
```js
date = moment(new Date(lot.expiration_date)).format('YYYY-MM-DD HH:mm:ss').toString();
```
since they will almost map to the same thing.

For example, on my machines:
```js
const mysql = require('mysql');
const moment = require('moment');

// some random date - this would be lot.expiration_date in your case
const date = new Date();

// log the Moment version
console.log(moment(new Date(date)).format('YYYY-MM-DD HH:mm:ss').toString());
// -> 2017-08-11 08:22:38

console.log(mysql.escape(date));
// -> '2017-08-11 08:22:38.245'
```

From the above, there are only two differences:
 1. MySQL's function includes apostrophes (`'`) to escape the query
 2. MySQL includes µs and Moment only includes milliseconds.

test/integration/stock.js
```diff
@@ -133,9 +137,9 @@ describe('(/stock/) The Stock HTTP API', () => {
 
   // ==============================================================================
 
-  const depot_principal_uuid = 'f9caeb16-1684-43c5-a6c4-47dbac1df296';
-  const depot_principal_mvt = 14; // 10 initial plus 4 distributions
-  const lot_quinine_uuid = 'ae735e99-8faf-417b-aa63-9b404fca99ac';
+  const depotPrincipalUuid = 'f9caeb16-1684-43c5-a6c4-47dbac1df296';
```
jniles, line 140: :+1:

jniles (commented):
@mbayopanda, I've made some additional comments.  I think this can be merged soon.

None of the "in the future" comments really need to be addressed, but there is a bug in the report validation for `inventory_report`.

It is also best practice to write a report test ... I'm not sure if @sfount wants these to be blocking or not.  But it would be nice to have a test for these reports to make sure they produce accurate information based on the test database.

client/src/modules/reports/generate/inventory_file/inventory_file.html
mbayopanda: 👍

server/controllers/stock/core.js
mbayopanda: `bundle.consumptions` contains consumptions data which come from a function used by  others functions, the goal here is to get consumption information and current quantities for inventories; both come from two complex functions : `getLots()` and `getStockConsumptionAverage()`

jniles (approved):
These changes look good to me :+1:

Issue #1739, opened 2017-06-01
refactor: remove bhFormDefaults


jniles:
We have deprecated `bhFormDefaults` for a while now - see https://github.com/IMA-WorldHealth/bhima-2.X/issues/1447.  It will be easier to erase it once and for all from the repository rather than enforce developers not using it at code review.  Someone should go through the entire repository and remove this directive.

PR #1886, opened 2017-07-21
Link entities from report

jeremielodi:
feat(journal) Entity link components

 This PR add a link on each reference in the journal
 It allow to quickly see the referenced report (invoice,cash,..)


client/src/modules/journal/templates/references.link.html
```diff
@@ -0,0 +1,11 @@
+
+<div class="ui-grid-cell-contents">
+  <a 
+    style="text-align:center !important"
+    href="journal/refenrece_transaction/{{row.entity.hrReference}}/{{grid.appScope.languages.key}}"
```
jniles: The correct way to write this is using `ngHref`.  See the [angularjs documenation for ngHref](https://docs.angularjs.org/api/ng/directive/ngHref).

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+      <li class="static">{{ "Reporting Group" | translate }}</li>
+    </ol>
+
+    <div class="toolbar">
+      <div class="toolbar-item">
+        <a  class="btn btn-default pull-right"
+          ui-sref="emailreport"
+          ui-sref-opts="{notify:false}">
+            <i class="fa fa-plus"></i>
+            <span  translate=""  class="ng-scope">
```
jniles: You should just write a single, empty `translate` tab to indicate the directive does not take any parameters.  Like this:
```html
<span translate>
  SOME.I18N.KEY
</span>
```

Also, `class="ng-scope"` should never be added by you - it is added automatically by Angular when the page is uploaded.  By adding it yourself, you may confuse debugging tools.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
+<div class="flex-header static">
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static">{{ "TREE.ADMIN" | translate }}</li>
```
jniles: These should use the `translate` directive; you have used the `translate` filter.  This code should look like this:
```html
<li class="static" translate>TREE.ADMIN</li>
```

You can read about the differences between the two by comparing the [angular-translate filter](https://angular-translate.github.io/docs/#/guide/04_using-translate-filter) and the [angular-translate directive](https://angular-translate.github.io/docs/#/guide/05_using-translate-directive).

jniles: This change should be made everywhere you have used the `translate` filter.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+
+
+
+<div class="flex-content">
+  <div class="container-fluid" id="anchor">
+
+  <div class="col-lg-12">
+    <!-- start new Reporting group form -->
+    <div class="col-lg-5">
+    <h4> {{"New reporting group"}}</h4>
```
jniles: There is no difference between the following two lines in output:
```html
<h4>{{ "New reporting Group" }}</h4>
<!-- versus -->
<h4>New reporting Group</h4>
```

When you write the `{{ }}` angular syntax, the browser attempts to evaluate what is between the brackets.  Since it's a string, it will simply return the string.

However, any time anything changes on the page, angular will a "digest" cycle.  It will re-evaluate everything in curly braces (`{{ }}`) on the page.  The more curly braces you have, the slower the performance will be.

Since you do not need the braces here, you should remove them.

jniles: The correct way to write this in is using the `translate` directive.  First, you must make a key in the `i18n` JSONs.  Try this:
 1. Create a file called `report-group.json` in the folder `client/src/i18n/en/`.
 2.  Put the following code inside it:
  ```json
{
  "REPORT" : {
    "NEW_GROUP" : "New Reporting Group"
   }
}
```
 3. Do the same thing in a new file, `report-group.json` in the folder `client/src/i18n/fr/`, but with french translations.
 4. Change the above HTML to be:
```html
<h4 translate>REPORT.NEW_GROUP</h4>
```

You have completed a translation in BHIMA!

If you need any help or clarity, anyone on the local team can help you out.  We are all very good at this now :dancer:

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+
+
+    <form
+      name="ReportGroupRegistrationForm"
+      bh-submit="ReportGroupCtrl.save(ReportGroupRegistrationForm)"
+      bh-form-defaults
```
jniles: You do not need to include `bhFormDefaults`.  It was deprecated (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1447) and then removed (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1739).

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+      novalidate>
+      <!-- code -->
+      <div
+      class="form-group"
+      ng-class="{'has-error' : ReportGroupRegistrationForm.code.$invalid && ReportGroupRegistrationForm.$submitted}">
+
+      <label class="control-label ng-binding">
```
jniles: There is no need to put the class `ng-binding`.  It is added automatically by angular.  You may confuse debugging tools by including it.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+        ng-messages="ReportGroupRegistrationForm.description.$error"
+        ng-show="ReportGroupRegistrationForm.$submitted">
+          <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+        </div>
+      </div>
+      </div>
+
+
+      <bh-loading-button
+      ng-show="ReportGroupCtrl.selectedReportGroup.selected == false"
```
jniles: You should always show submit buttons.  If you need to make it inactive, use the `disabled` directive.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+            class="pull-right" loading-state="BedRegistrationForm.$loading">
```
jniles: The `loading-state` should say "ReportGroupRegistrationForm" instead of "BedRegistrationForm".

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+              {{ "Save"}}
+          </bh-loading-button>
+
+          <span
+            class="btn btn-danger"
+            ng-click="ReportGroupCtrl.remove(ReportGroupRegistrationForm)"
+            ng-show="ReportGroupCtrl.selectedReportGroup.selected == true"
```
jniles: There is no need to compare to `true` here.  You can just write:
```html
ng-show="ReportGroupCtrl.selectedReportGroup.selected"
```
with the same effect.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+      <!-- end new Reporting group form -->
+
+      <!-- start grid of Reports group -->
+
+      <div class="col-lg-7">
+        <!--the grid -->
+
+        <div
+          id="report-groups-grid"
+          ui-grid-auto-resize
```
jniles: Do you need all these options on the `uiGrid`?!  Remember that every option you add will slow the module down!

server/controllers/finance/journal/index.js
```diff
@@ -36,6 +40,64 @@ exports.journalEntryList = journalEntryList;
...
+   const query = `
+    SELECT BUID(uuid) as uuid FROM ${documentDefinition.table}  tb, project p 
+    WHERE tb.project_id=p.id AND p.abbr="${project_name}" AND tb.reference="${refence}"
+  `;
+  
+  // search for full UUID
+  db.one(query)
+    .then(result =>
+      {
+        var uuid=result.uuid;
```
jniles: On  the server, you should prefer `let` and `const` over `var`.

server/controllers/finance/journal/index.js
```diff
@@ -36,6 +40,64 @@ exports.journalEntryList = journalEntryList;
...
+  const refence=codeRef[2];
+  const documentDefinition = identifiersIndex[code];
+   
+   const query = `
+    SELECT BUID(uuid) as uuid FROM ${documentDefinition.table}  tb, project p 
+    WHERE tb.project_id=p.id AND p.abbr="${project_name}" AND tb.reference="${refence}"
+  `;
+  
+  // search for full UUID
+  db.one(query)
```
jniles: It would be nicer to follow the repository standards for indentation.  We generally write our code like this:
```js
db.one(query)
  .then(row => {
     console.log(row.id);
     return 1;
  })
  .then(integer => {
     // do something else in this block
   })
   .catch(err => {
      // include some error handling code
   })
   .done();
```

server/controllers/finance/journal/index.js
```diff
@@ -36,6 +40,64 @@ exports.journalEntryList = journalEntryList;
...
+    .then(result =>
+      {
+        var uuid=result.uuid;
+        var url= reportsTypes[code]+uuid+"?lang="+language+"&posReceipt=0&renderer=pdf";
+        res.redirect(url);
+      }
+    ).catch((error)=>{
+      res.send({'message':'an error occured'});
+    });
```
jniles: On the server, it's a good idea to end promise chains with `.done()` to make sure all the errors are thrown in the promise block.

server/controllers/finance/journal/index.js
```diff
@@ -36,6 +40,64 @@ exports.journalEntryList = journalEntryList;
...
+    IV:'/reports/finance/invoices/',
+    CP:'/reports/finance/cash/'
+  };
+
+  var codeRef =  req.params.codeRef.split(".");
+  const language= req.params.language;
+
+  const code = codeRef[0];
+  const project_name=codeRef[1];
+  const refence=codeRef[2];
```
jniles: Please include a space between the name of the variable, the equals sign, and the assignment.  Like this:
```js
const refence = codeRef[2];
```

I also think this should probably be called `reference` instead of `refence`.

server/controllers/finance/journal/index.js
```diff
@@ -36,6 +40,64 @@ exports.journalEntryList = journalEntryList;
...
+    VO:'/reports/finance/vouchers/',
+    IV:'/reports/finance/invoices/',
+    CP:'/reports/finance/cash/'
+  };
+
+  var codeRef =  req.params.codeRef.split(".");
+  const language= req.params.language;
+
+  const code = codeRef[0];
+  const project_name=codeRef[1];
```
jniles: We prefer to use `camelCase` for our variable names instead of `underscores_syntax`.  The variable should be called
```js
const projectName = codeRef[1];
```

server/controllers/finance/journal/index.js
```diff
@@ -36,6 +40,64 @@ exports.journalEntryList = journalEntryList;
...
+  const refence=codeRef[2];
+  const documentDefinition = identifiersIndex[code];
+   
+   const query = `
+    SELECT BUID(uuid) as uuid FROM ${documentDefinition.table}  tb, project p 
+    WHERE tb.project_id=p.id AND p.abbr="${project_name}" AND tb.reference="${refence}"
```
jniles: This should make sure to use SQL template strings to avoid SQL injection attacks.  You can read about it [here](https://github.com/mysqljs/mysql#escaping-query-values).

Every SQL statement that you write that takes in user input should use question marks (`?`) to escape values sent from the client.  Like this:
```js
 const query = `
    SELECT BUID(uuid) as uuid FROM ${documentDefinition.table}  tb, project p 
    WHERE tb.project_id = p.id AND p.abb r= ? AND tb.reference = ?
`;
```

Then, to execute the query, you pass in the parameters in an array for each question mark to `db.exec()` or `db.one()`.  Like this:
```js
db.one(query, [project_name, refence])
  .then(result => {
    // ... etc ...
   });
```

client/src/modules/journal/templates/references.link.html
jniles: Additionally, please change this URL to be `reference_transaction` instead of `refenrece_transaction`.  There is just a minor typo in the spelling of `reference` :+1:

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
...
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static">{{ "TREE.ADMIN" | translate }}</li>
+      <li class="static">{{ "Reporting Group" | translate }}</li>
+    </ol>
+
+    <div class="toolbar">
+      <div class="toolbar-item">
+        <a  class="btn btn-default pull-right"
+          ui-sref="emailreport"
```
jniles: Does this state exist?

jniles (changes requested):
@jeremielodi Great attempt at this issue!  I know you do not have a lot of experience contributing, but this represents a lot of work on both the client and the server.  Well done!

I believe that your logic is mostly correct.  I've made a lot of comments to improve the quality of your code - right now it does not look anything like we have in our repository.  We'll continue to work on this until you cannot tell the difference between each member's code!

The main issues here are:
 1. Learning how to do translations.  Please ask someone like @DedrickEnc or @mbayopanda for help on this if you need it.  We cannot merge your code until the text is represented in both English and French.  Do not worry about the correctness of the English translations ... I can help you with that when the time comes!
 2. Never include `ng-scope`, `ng-binding` or another other  classes that include the prefix `ng-*`.  This is used by angular internally and we shouldn't include it.
 3. Please make sure that your code passes ESLint.  I'm not sure if it does right now.  That should help you get things like `camelCase` and indentation correct.

Let me know if you need more feedback or if the changes are made, and I'll give you another review.

client/src/modules/report-group/index.html
```diff
@@ -0,0 +1,170 @@
+<div class="flex-header static">
```
sfount: As I understand it these files should **not** be included in this pull request - they belong to https://github.com/IMA-WorldHealth/bhima-2.X/pull/1778.

server/controllers/finance/journal/index.js
```diff
@@ -36,6 +40,64 @@ exports.journalEntryList = journalEntryList;
...
+//The reference code is a combination of table_key.project_abbr.reference
+//The table name is variable, it can be :invoice, cash or voucher
+
+function refenreceTransaction(req,res,next){
+   
+  //loading identifiers keyies, used for defining the table name
+  indexIdentifiers();
+
+  //reports urls
+  var reportsTypes={
```
sfount: This map already exists in the `identifiers.js` object that you import - it would be best not to duplicate this behaviour.

I think it would be best to add these as `redirectPath` attributes to the already existing object. 

This code could then run without defining report types or indexing them.

server/controllers/finance/journal/index.js
```diff
@@ -36,6 +40,57 @@ exports.journalEntryList = journalEntryList;
...
+  // search for full UUID
+  db.one(query,[projectName,refence])
+    .then(result =>
+      {
+        let uuid=result.uuid;
+        let url= documentDefinition.documentPath + uuid + "?lang="+language+"&posReceipt=0&renderer=pdf";
+        res.redirect(url);
+      }
+    ).catch((error)=>{
+      res.send({'message':'an error occured'});
```
sfount: With this default error handling this error can just be passed to `next` - if a specific error message must be sent catch can be used like this. 

Next can be used as: 
```js
db.one()
  .then((result) => //...
  .catch(next)
  .done()
```

`next` will pass this error on to the next method in the middleware stack - at the end of this stack there is an error handler that will send a generic error message.

server/controllers/finance/journal/index.js
```diff
@@ -36,6 +40,57 @@ exports.journalEntryList = journalEntryList;
...
+    SELECT BUID(uuid) as uuid FROM ${documentDefinition.table}  tb, project p 
+    WHERE tb.project_id=p.id AND p.abbr=? AND tb.reference=?
+  `;
+  
+  // search for full UUID
+  db.one(query,[projectName,refence])
+    .then(result =>
+      {
+        let uuid=result.uuid;
+        let url= documentDefinition.documentPath + uuid + "?lang="+language+"&posReceipt=0&renderer=pdf";
```
sfount: This string concatenation could be done quite clearly with the new String syntax

For example: 
```js
const url = `${documentDefinition.documentPath}${uuid}?lang=${language}&renderer=pdf`
```

sfount (commented):
I've now looked fully through this - this is great work. 

I've made some final comments on error handling and stylistic comments. 

The last thing I think should be done is to move this logic from the `journal` code into a unique library in `server/lib` that accepts a human readable reference and turns it into a path for the report. 

Having this code in the journal code makes it very specific for the client module - if this was moved to a server library it could be used by all of the server as well as a generic point for all client modules.

jniles:
@jeremielodi, when you need another review, just leave a comment!  I think we can probably merge this soon :+1:

jeremielodi:
okay

2017-07-28 10:07 GMT+01:00 Jonathan Niles <notifications@github.com>:

jniles:
Closed in https://github.com/IMA-WorldHealth/bhima-2.X/commit/9b690a1bfa683000b091d5dc70885fa8caa2ac95

jniles:
@jeremielodi, I fixed the `next()` request that @sfount wanted, and it works well!  Great work!

PR #1892, opened 2017-07-25
Fix data period total

### LINKED ISSUE #1891: (bug) Problem with Account Report
There is a difference in value when looking for the account statement using a date range (2015-01-01 and 2015-12-31) with (2015-01-01 and 2016-01-01) ,

NB: these tests were carried out with the development-build.sql database

lomamech:
fix(Data Development) 
- Update data for the table period_total in development-build.sql 

test(Update Regression test) 
- Update and Reset data for regression test - Expected the balance sheet report formule
 - Restaure the request in extra.js

closes #1891


jniles:
LGTM :+1:

jniles:
All tests pass!

PR #1906, opened 2017-07-30
fix(reports) Account statement period total currency

sfount:
This commit ensures that the period total displayed at the bottom
of the account statement report respects the correct currency.
The enterprise currency ID is now correctly passed to the
currency helper.


jniles:
Good catch!

Issue #1843, opened 2017-07-12
(refactor) Purchase Order Registry refactoring functionality


DedrickEnc:
The purchase order registry is reviewed and is in master branch, after trial here is points to address

1. Applying registry standard #1749

2. **MODIFIER ETAT** seems not useful and can not see how this can work with a purchase order of type **PAYER LORS DE L'ACHAT** and it is also easy to forget updating it!

These point can be discussed to have final point to address.

jniles:
See as well https://github.com/IMA-WorldHealth/bhima-2.X/issues/1746 which discusses point 1 in detail.

DedrickEnc:
Can be closed, if there is a point not tackled, a separated issue can be created

PR #1907, opened 2017-07-31
Refactor(Purchase Order Registry)

### LINKED ISSUE #1746: refactor: Improvements to Purchase Order Registry
This issue is to track improvements to the Purchase Order Registry to be tackled while the team is working all registries.

Parent Issue #1737.

-------

The follow changes should be made to the Purchase Order Registry:
 - [ ] Make a "Menu" dropdown with the following options:
   1. Column Configuration
   2. Saving the Grid Configuration
   3. Resetting the Grid Configuration
   4. Download View as CSV
   5. Download View as PDF
 - [ ] Implement the Grid State work on the grid (See #1738)
 - [ ] Implement the latest FilterService, replacing the old `bhFiltersApplied` with the new `FilterService`
 - [ ] Implement downloading as CSV using the `ui-grid` exporter (See https://github.com/IMA-WorldHealth/bhima-2.X/pull/1554)
 - [ ] Remove GridGrouping until we can properly use it
 - [ ] Make sure the search form is up to the latest standards (see #1702)
 - [ ] Use an `actions` dropdown menu in line of the grid.

lomamech:
Refactor(Purchase Order Registry) 
- Integrate grid save State
- Add dropdaown for column configuration, Download report on CSV and PDF 
- Implement report Purchase Order Handlebars

test(Purchase Order Registry) 
- Implement E2E test for check Purchase order registry 
- Fix some property and Identification
 - Remove old E2E test

closes #1746


client/src/modules/purchases/list/list.js
```diff
@@ -89,37 +91,34 @@ function PurchaseListController ($translate, PurchaseOrder, Notify, uiGridConsta
...
+  /** TODO manage column : last_transaction */
+  vm.uiGridOptions = {
     appScopeProvider  : vm,
-    enableFiltering   : vm.filterEnabled,
     showColumnFooter  : true,
-    fastWatch         : true,
+    enableSorting     : true,
+    enableColumnMenus : false,
     flatEntityAccess  : true,
+    fastWatch         : true,
```
jniles, line 99: :+1:

DedrickEnc:
references #1843

client/src/modules/purchases/modals/search.modal.js
```diff
@@ -1,24 +1,55 @@
...
 
 // dependencies injections
 SearchPurchaseOrderModalController.$inject = [
-  'SupplierService', 'NotifyService', '$uibModalInstance', 'SearchFilterFormatService',
+  '$uibModalInstance', 'params', 'SupplierService', 'bhConstants', 'moment', 'Store', 'util', 'PeriodService', 'NotifyService'
 ];
 
-function SearchPurchaseOrderModalController(Suppliers, Notify, Instance, SearchFilterFormat) {
+/**
+ * @class PatientRegistryModalController
```
jniles: This should probably talk about the `SearchPurchaseOrderModalController()` instead.

client/src/modules/purchases/modals/search.modal.js
```diff
@@ -27,27 +58,52 @@ function SearchPurchaseOrderModalController(Suppliers, Notify, Instance, SearchF
     })
     .catch(Notify.handleError);
 
-  function init() {
-    vm.bundle = {
-      dateFrom : new Date(),
-      dateTo   : new Date(),
-    };
+  // custom filter user_id - assign the value to the params object
+  vm.onSelectUser = function onSelectUser(user) {
```
jniles, line 55: :+1:  Nice comments!

server/controllers/finance/purchases.js
```diff
@@ -288,7 +289,20 @@ function search(req, res, next) {
  * filter the purchase orders.
  */
 function find(options) {
-  const filters = new FilterParser(options, { tableAlias : 'p' });
+  // ensure epected options are parsed appropriately as binary
+  db.convert(options, ['supplier_uuid']);
+
+  const filters = new FilterParser(options, { tableAlias : 'p', autoParseStatements : false  });
```
jniles, line 295: :+1:

server/controllers/finance/reports/purchases/report.handlebars
```diff
@@ -0,0 +1,83 @@
...
+          {{#each rows}}
+            <tr>
+              <td>{{reference}}</td>
+              <td class="text-right">{{date date}}</td>
+              <td>{{author}}</td>
+              <td>{{note}}</td>
+              <td class="text-right">{{currency cost currency_id}}</td>
+              <td>
+                <strong>
+                  {{#if is_confirmed}}
```
jniles: Wow, this is a really confusing series of logic.  Can you move this into JS?  Something like:
```js
row.isPartiallyRecieved = (row.is_confirmed && !row.is_canceled && row.is_received && row.is_partially_received);
```
and then you can just 
```hbs
<strong>
  {{#if row.isPartiallyReceived }} {{translate 'PURCHASES.STATUS.PARTIALLY_RECIEVED' }} {{/if}}
  {{! etc ...}}
</strong>
```

client/src/modules/purchases/list/list.html
```diff
@@ -1,40 +1,92 @@
...
+      id="purchase-registry"
+      class="grid-full-height-with-filters"
+      ng-style="PurchaseListCtrl.filterBarHeight"
+      ui-grid="PurchaseListCtrl.uiGridOptions"
       ui-grid-auto-resize
       ui-grid-resize-columns
-      ui-grid-grouping>
+      ui-grid-move-columns
+      ui-grid-save-state
+      ui-grid-selection>
```
jniles: You never actually use `ui-grid-selection`.  It can be removed.

client/src/modules/purchases/modals/search.modal.js
```diff
@@ -1,24 +1,55 @@
...
+  }
+
+  // @TODO ideally these should be passed in when the modal is initialised
+  //       these are known when the filter service is defined
+  var searchQueryOptions = [
+    'reference', 'user_id', 'supplier_uuid', 'defaultPeriod', 'is_confirmed', 'is_received', 'is_cancelled'
+  ];
+
+  // assign already defined custom params to searchQueries object
+  vm.params = util.maskObjectFromKeys(params, searchQueryOptions);
```
jniles, line 40: :+1:

client/src/modules/purchases/modals/search.tmpl.html
```diff
@@ -1,88 +1,111 @@
 <form
   name="ModalForm"
-  ng-submit="$ctrl.submit(ModalForm)"
+  ng-submit="ModalCtrl.submit(ModalForm)"
   data-modal="purchase-search"
   novalidate>
 
-  <div class="modal-header">
+  <div class="modal-header" style="box-shadow : 0px 2px 4px -4px black; z-index:1000">
```
jniles: You can just use `class="search-modal-header" and remove hard coded styles.

client/src/modules/purchases/modals/search.tmpl.html
```diff
@@ -1,88 +1,111 @@
...
+          </bh-user-select> 
+
+
+          <!-- status -->
+          <div class="form-group">
+            <p class="control-label" style="margin-bottom:5px;">
+              <strong translate>FORM.LABELS.STATUS</strong>
+            </p>
+
+            <label class="checkbox-inline">
```
jniles: Are you sure these should be checkboxes?  It looks like they do not work as expected.

For example, in the test dataset, search of "confirmed" yields 1 result.  Searching for both "confirmed" and "Received In Stock" should show _at least_ one, but nothing matches.
confirmedandreceivedinstock
_Fig 1: Confirmed and Received Selection_

server/controllers/finance/purchases.js
```diff
@@ -288,7 +289,20 @@ function search(req, res, next) {
...
+
+  // default purchase date
+  filters.period('period', 'date');
+  filters.dateFrom('custion_period_start', 'date');
+  filters.dateTo('custom_period_end', 'date');
+  filters.equals('user_id');
+  filters.equals('is_confirmed');
```
jniles, line 302: Since you are using checkboxes, you would expect this to make an SQL like:
```sql
is_confirmed = ? OR is_received = ? OR is_canceled = ?
```
That way at least one matches when you search for both being checked.

server/controllers/finance/purchases.js
```diff
@@ -302,16 +316,12 @@ function find(options) {
       JOIN user AS u ON u.id = p.user_id
   `;
 
-  filters.dateFrom('dateFrom', 'date');
-  filters.dateTo('dateTo', 'date');
-
   const referenceStatement = `CONCAT_WS('.', '${identifiers.PURCHASE_ORDER.key}', pr.abbr, p.reference) = ?`;
   filters.custom('reference', referenceStatement);
 
-  filters.setOrder('ORDER BY p.date DESC');
```
jniles: You should probably keep the `setOrder()` around.  We want to make sure only recent records are displayed at the top.

client/src/modules/purchases/modals/search.modal.js
```diff
@@ -1,24 +1,55 @@
...
-function SearchPurchaseOrderModalController(Suppliers, Notify, Instance, SearchFilterFormat) {
+/**
+ * @class PatientRegistryModalController
+ *
+ * @description
+ * This controller is responsible for setting up the filters for the patient
+ * search functionality on the patient registry page.  Filters that are already
+ * applied to the grid can be passed in via the params inject.
+ */
+function SearchPurchaseOrderModalController(ModalInstance, params, Suppliers, bhConstants, moment, Store, util, Periods, Notify) {
```
jniles: Do you use `moment`?  Please remove any unused imports.

test/end-to-end/purchases/registry.spec.js
```diff
@@ -0,0 +1,13 @@
+const helpers = require('../shared/helpers');
+
+const PurchaseOrderSearch = require('./registry.search');
+
+// purchase order registry tests
+describe('Purchase Order Registry', function () {
+  'use strict';
+
+  before(() => helpers.navigate('#/purchases/list'));
```
jniles: It would be nice to make this URL `/purchases`.  That way it would match the other registries:
 1. `/patients` -> patient registry
 2. `/payments` -> cash payments registry
 3. `/invoices` -> invoice registry.

client/src/modules/purchases/list/list.html
```diff
@@ -1,40 +1,92 @@
...
+
+        </div>
+      </div>
+        
+      <div class="toolbar-item">
+        <button
+          ng-click="PurchaseListCtrl.search()"
+          data-method="search"
+          class="btn btn-default">
+          <span class="fa fa-search"></span> <span translate>FORM.BUTTONS.SEARCH</span>
```
jniles: If you put `class="hidden-xs"` on the FORM.BUTTONS.SEARCH element, it will be hidden on small screens.  That makes a much nicer user experience :+1:

jniles (changes requested):
@lomamech, this is a really nice refactor.  It's beginning to look like a standard registry!

Some comments:
 1. The `is_confirmed`, `is_received`, and `is_canceled` do not behave correctly.  When combined, they will never match anything.  The solution is to write a custom filter (`filters.custom()` that will allow these to be matched with an OR.
 2. Please remove all unused imports (`require()` and `$injects`).
 3. The `amount` column does not respect sorting.  You need to put `type: 'number'` on the column `ui-grid`.
 4. Instead of using `Document`, it would be nice if this looked like the other registries.  For example, you could simply link the receipt from the `Reference` column, like the Patients Registry does.

After these edits are made, this will be completely up to date with 2.x standards.

jniles:
I'll tackle updating these.

lomamech:
@jniles I am also working on this

jniles:
@lomamech, aren't you supposed to be on vacation?!

Okay, it is all yours 👍

lomamech:
I think it would be necessary to review the Purchase Order status modification tool, because in my a purchase order that was received can no longer be canceled

jniles:
@lomamech, if that is the case, it doesn't seem like checkboxes are a good way to select POs that have multiple statuses.  The reason being is that there are three states you can think about:
 1. Default - do not consider the field
 2. Cancelled/Confirmed/Received - search for this field as true.
 3. NOT Cancelled/Confirmed/Received - search for this field as false.

This isn't ideally suited for a checkbox.  A better solution would be to implemented either radio buttons for each input (that would make it a really long search modal) or a ui-select multiple like in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1941.

You've used radio buttons for now, which only allows filtering on one field.  That is a fine first solution until we can implement a better select option.

I'll merge this manually later today.

jniles (approved):

jniles:
Finally figured out how to push to remote branches. :+1:

PR #1910, opened 2017-07-31
fix(patients): responsive search button

jniles:
This commit makes the "SEARCH" text get hidden on smaller screens,
allowing the registry to behave more responsively on smaller screens.  It 
also removes the ui-grid selection plugin which was never used.


sfount (approved):
LGTM - defining this in a class clears up the template well 👍

PR #1913, opened 2017-08-01
refactor(billing services): update to standards

jniles:
This commit updates the billing services module in the following ways:
 1. It uses inline filtering like the other registries.
 2. Fixes a bug in rendering the account field - now only the number is
 displayed to avoid server API changes.
 3. Account dropdown filters out title accounts
 4. Sorting is enabled on the registry
 5. Use performance improvements for the ui-grid

### DIFF
--- a/client/src/modules/billing-services/billing-services.ctrl.js
+++ b/client/src/modules/billing-services/billing-services.ctrl.js
@@ -1,8 +1,8 @@
 angular.module('bhima.controllers')
-.controller('BillingServicesController', BillingServicesController);
+  .controller('BillingServicesController', BillingServicesController);
 
 BillingServicesController.$inject = [
-  '$state', 'BillingServicesService', 'AccountService', 'NotifyService', 'bhConstants', '$timeout'
+  '$state', 'BillingServicesService', 'NotifyService', 'bhConstants', '$timeout', 'uiGridConstants',
 ];
 
 /**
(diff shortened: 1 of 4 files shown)


client/src/modules/billing-services/billing-services.ctrl.js
```diff
@@ -11,55 +12,63 @@ BillingServicesController.$inject = [
...
+    displayName: '',
+    enableSorting : false,
+    cellTemplate : actionTemplate,
+  }];
 
   // these options are for the ui-grid
   vm.options = {
     appScopeProvider: vm,
     enableSorting : true,
+    flatEntityAccess  : true,
```
sfount, line 66: :+1: incrementally making sure our older grids are using the latest standards

package.json
```diff
@@ -56,7 +56,7 @@
     "express-handlebars": "^3.0.0",
     "express-session": "^1.15.4",
     "gulp-merge-json": "^1.1.0",
-    "helmet": "^3.8.0",
+    "helmet": "^3.8.1",
```
sfount: I don't think these version updates are related to this PR - could they be separated?

jniles: Cheers, I'll remove them.

jniles:
@sfount, updates have landed!

client/src/modules/billing-services/billing-services.ctrl.js
```diff
@@ -94,17 +102,17 @@ function BillingServicesController($state, BillingServices, Accounts, Notify, bh
...
 
     // retrieve a detailed list of the billing services in the application
     BillingServices.read(null, { detailed : 1 })
       .then(function (billingServices) {
 
         // make a pretty human readable account label
+        // TODO(@jniles) - make this use the account_label.  Requires changing
+        // the server
         billingServices.forEach(function (service) {
-          service.account = Accounts.label(service);
```
sfount, line 107: Is the `Accounts` library used anymore here? I think the import can be removed.

sfount (approved):
Changes LGTM. This is a great re-factor! I think you make one of the controller imports redundant in this PR so it would be good to remove that. This can be merged any time.

client/src/modules/billing-services/billing-services.ctrl.js
jniles: Aha - I'll get right on that.

jniles:
@sfount second round of changes have landed!

jniles:
bors r=sfount

PR #1916, opened 2017-08-02
(fix) Reference links corner cases

sfount:
This commit introduces checks to ensure that the server does not
get stuck in a redirect loop if invalid information is passed
to the reference lookup method.

This commit refactors the reference lookup library according to
eslint standards. It also renames the library fixing a language
typo.


jniles:
LGTM :+1:

jniles:
This seems to work as expected.  Great refactor!

PR #1925, opened 2017-08-02
imporing the inventory search modal

jeremielodi:
updating inventory search modal by adding more seach fields

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -212,6 +212,7 @@
             "COST": "Cost",
             "COST_CENTER": "Cost Center",
             "CONSUMABLE": "Consumable",
+            "UNCONSUMABLE": "Unconsommable",
             "CONVENTION": "Convention",
             "COUNTRY": "Country",
             "CREATE": "Create",
--- a/client/src/js/services/SearchFilterService.js
+++ b/client/src/js/services/SearchFilterService.js
@@ -33,6 +33,9 @@ function SearchFilterService(Filters, AppCache) {
       { key : 'invoice_uuid', label : 'FORM.LABELS.INVOICE' },
       { key : 'group_uuid', label : 'FORM.LABELS.GROUP' },
       { key : 'text', label : 'FORM.LABELS.LABEL' },
+      { key : 'code', label : 'FORM.LABELS.CODE' },
+      { key : 'price', label : 'FORM.LABELS.UNIT_PRICE' },
+      { key : 'type_id', label : 'FORM.LABELS.TYPE' },
     ];
 
     this._filters.registerCustomFilters(this._availableFilters);
--- a/client/src/js/services/SearchFilterService.js
+++ b/client/src/js/services/SearchFilterService.js
@@ -36,6 +36,7 @@ function SearchFilterService(Filters, AppCache) {
       { key : 'code', label : 'FORM.LABELS.CODE' },
       { key : 'price', label : 'FORM.LABELS.UNIT_PRICE' },
       { key : 'type_id', label : 'FORM.LABELS.TYPE' },
+      { key : 'consumable', label : 'FORM.LABELS.CONSUMABLE' },
     ];
 
     this._filters.registerCustomFilters(this._availableFilters);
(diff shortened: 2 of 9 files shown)


client/src/modules/inventory/list/list.js
```diff
@@ -143,6 +143,7 @@ function InventoryListController($translate, Inventory, Notify, uiGridConstants,
       vm.latestViewFilters = search.latestViewFilters();
       vm.hasCustomFilters = search.hasCustomFilters();
       vm.parameters = JSON.stringify(params);
+      console.log(vm.parameters);
```
sfount: Console logs that were used for debugging should be removed.

client/src/modules/inventory/list/modals/search.modal.html
```diff
@@ -43,6 +43,62 @@
...
+        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      </div>
+    </div>
+
+
+      <div class="form-group" ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.consumable.$invalid }">
+        <label class="control-label" translate>
+          FORM.LABELS.CONSUMABLE
+        </label>
+        <input type="checkbox"   name="consumable" ng-model="ModalCtrl.param.consumable">
```
sfount: Does this flag for consumable actually work? 

Here it is assigned to a `param` object but I think the Modal controller uses `params` to filter and send parameters.  I am not sure if an exception is made in the case.

client/src/js/services/SearchFilterService.js
```diff
@@ -33,6 +33,9 @@ function SearchFilterService(Filters, AppCache) {
       { key : 'invoice_uuid', label : 'FORM.LABELS.INVOICE' },
       { key : 'group_uuid', label : 'FORM.LABELS.GROUP' },
       { key : 'text', label : 'FORM.LABELS.LABEL' },
+      { key : 'code', label : 'FORM.LABELS.CODE' },
+      { key : 'price', label : 'FORM.LABELS.UNIT_PRICE' },
+      { key : 'type_id', label : 'FORM.LABELS.TYPE' },
```
sfount, line 38: Is the `consumable` field factored in to this object? As I understand it if there is no definition for `consumable` it will not be displayed as an active filter.

sfount (commented):
This PR is looking good! This extends the functionality of inventory search and sticks to the current best practices. 

I have made some comments on code that may be broken or needs cleaning up. I may have misunderstood on these; if it works as expected you can leave a comment - otherwise it is good to briefly test all new functionality before submitting a PR.

client/src/modules/inventory/list/modals/search.modal.html
```diff
@@ -43,8 +43,62 @@
...
+    </div>
+
+
+      <div class="form-group" ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.consumable.$invalid }">
+        <label class="control-label" translate>
+          FORM.LABELS.CONSUMABLE
+        </label>
+         <bh-clear on-clear="ModalCtrl.clear('consumable')"></bh-clear>
+
+         <input type="checkbox" ng-true-value='1' ng-false-value='0' name="consumable" ng-model="ModalCtrl.params.consumable"/>
```
sfount: As discussed today in the office; I think a clearer user experience would be using 2 radio buttons with the options :
* Exclude all inventory items that are consumable 
* Only include inventory items that are consumable 

With a check box there is no way of excluding consumable items.

sfount (commented):
Looks like almost everything is addressed. I think there is one issue left with user experience and then this can be merged!

client/src/i18n/en/form.json
```diff
@@ -212,6 +212,7 @@
             "COST": "Cost",
             "COST_CENTER": "Cost Center",
             "CONSUMABLE": "Consumable",
+            "UNCONSUMABLE": "Unconsommable",
```
sfount, line 215: This translation should be 'unconsumable' I think!

sfount (commented):
Radio buttons are looking good! I think there's a translation issue but I think this PR can be merged.

PR #1933, opened 2017-08-03
updating zRecalculatePeriodTotals procedure

### LINKED ISSUE #1927: feature request: ignore opening balances in zRecalculatePeriodTotals()
We have a special function for recomputing the `period_total` table called [zRecalculatePeriodTotals()](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/admin.sql#L107).  Unfortunately, the function destroys the opening balances (all periods with `period.number = 0`) since it truncates the table.

A developer should go through and modify this function so that it removes only the period_total values where the period.number is not 0.  This will allow us to use the function on production data.

jeremielodi:
This PR allows to update zRecalculatePeriodTotals() procedure by specify witch row must be deleted while the period_total is being recomputed instead of truncating the all table

closes #1927

### DIFF
--- a/server/models/admin.sql
+++ b/server/models/admin.sql
@@ -9,7 +9,7 @@ DELIMITER $$
 CREATE PROCEDURE zRecomputeEntityMap()
 BEGIN
   DELETE FROM entity_map;
-  
+
   -- patient
   INSERT INTO entity_map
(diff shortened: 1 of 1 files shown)


jniles (approved):
Looks good to me! :+1:

Issue #1909, opened 2017-07-31
(feature request) show default filters on registry PDFs


jniles:
Some of our registry PDFs include the search filters sent back to the server on them.  However, these are limited to only the custom filters and do not include default filters.   Default filters (particularly the LIMIT) provide important contextual information.  We should include them on the printed PDF registries.  See below:

![patientregistrationnodefaultfilters](https://user-images.githubusercontent.com/896472/28783058-fff0685a-7606-11e7-917d-5df7021aa7f9.png)
_Fig 1: Patient Registrations without any custom filters_

jniles:
@jeremielodi, can you look into this?

The basic idea is to put the filters that we see on the client like this:

![defaultfilters](https://user-images.githubusercontent.com/896472/29116032-7b79be18-7cf1-11e7-944e-cded8fd9af93.png)
_Fig 1: Default Filters_

.. on the server PDF reports.

To render on the PDF report.  A good example of how filters can be rendered on the server is seen in [the Patient Registrations Report](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/medical/reports/registrations.handlebars#L9).

It should be pretty straight-forward.  Let me know if you need any help!

jeremielodi:
Yes, I can look into it but my internet connection is very slow now

jniles:
Closed in #1947.

PR #1947, opened 2017-08-09
Report rendering including filters

jeremielodi:
This PR help to add default filter parameters in the report

### DIFF
--- a/server/controllers/medical/reports/registrations.js
+++ b/server/controllers/medical/reports/registrations.js
@@ -41,7 +41,7 @@ function formatFilters(qs) {
     { field : 'patient_group_uuid', displayName : 'PATIENT_GROUP.PATIENT_GROUP' },
     { field : 'user_id', displayName : 'FORM.LABELS.USER' },
     { field : 'limit', displayName : 'FORM.LABELS.LIMIT' },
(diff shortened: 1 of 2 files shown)


server/controllers/medical/reports/registrations.js
```diff
@@ -39,13 +40,22 @@ function formatFilters(qs) {
     { field : 'debtor_group_uuid', displayName : 'FORM.LABELS.DEBTOR_GROUP' },
     { field : 'patient_group_uuid', displayName : 'PATIENT_GROUP.PATIENT_GROUP' },
     { field : 'user_id', displayName : 'FORM.LABELS.USER' },
+    { field : 'limit', displayName : 'FORM.LABELS.LIMIT' },
+    { field : 'period', displayName : 'FORM.LABELS.PERIODES', isPeriod : true },
```
jniles: I think we already have a key for this - `TABLE.COLUMNS.PERIOD`.  You should use this one since it will render the same as the one on the client.

jniles (approved):
Looks good to me!  Great job!

I've made one comment on the key to make sure it looks the same as the client.  That should be fixed and then I will merge it.

jniles (approved):
Looks good to me!   Great work.

jniles:
Note: references #1909.

PR #1962, opened 2017-08-10
fix component validation triggering

DedrickEnc:
This P.R fixes the ```bhDepotSelect``` and ```bhInventorySelect``` component validation to behave properly if they are required

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ nbproject/
 
 # ignore istanbul generated folders
 .nyc_output/
+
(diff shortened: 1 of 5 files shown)


jniles (approved):
LGTM :+1:

jniles:
So, it turns out the reason validation was failing is because `$$parent` doesn't exist - the variable pointing to the parent form is called [$$parentForm](https://github.com/angular/angular.js/blob/master/src/ng/directive/form.js#L158).

:(

PR #1963, opened 2017-08-10
(fix) Voucher tool select

### LINKED ISSUE #1915: (bug) Patient support tool does not work when patient is selected by ID
The patient support tool works as expected when a patient is selected by name search - it returns two lines in a voucher debiting the new account and crediting the patient's account. 

However if a patient is selected by ID this process fails and only one line is shown on the voucher. 

This is a critical bug. 

Additional refactors: 
* Voucher account select shows title accounts and account number and text are not aligned 
* Patient support tool - submitting without an account selecting does nothing when the account should be highlighted in red and shown as required data

sfount:
This commit ensures that both methods of searching for a patient
return a detailed patient object.
Finding by reference uses the /patients/ route and find by name uses
/patients/:uuid. By passing detailed : 1 to the generic route we
can make sure both objects returned have the same information.

This resolves the voucher tools for covering patients not having
enough information to create all rows.

This PR also removes two large GET requests in the patient support tool. 

Contributes to https://github.com/IMA-WorldHealth/bhima-2.X/issues/1915
Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/1915

### DIFF
--- a/client/src/js/components/bhFindPatient.js
+++ b/client/src/js/components/bhFindPatient.js
@@ -122,6 +122,7 @@ function FindPatientComponent(Patients, AppCache, Notify, SessionService, bhCons
 
     options = {
       reference : isValidNumber ? [bhConstants.identifiers.PATIENT.key, SessionService.project.abbr, reference].join('.') : reference,
+      detailed : 1,
       limit     : 1
     };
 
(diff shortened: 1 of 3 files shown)


jniles (approved):
LGTM. :+1:

PR #1966, opened 2017-08-10
feat(deps): use yarn on Travis

### LINKED ISSUE #880: (build) use yarn instead of NPM for packages
A faster new alternative to NPM was developed by Facebook called [yarn](https://github.com/yarnpkg/yarn).  Probably the largest advantage is speed - [the benchmarks](https://www.berriart.com/blog/2016/10/npm-yarn-benchmark/) place it at 2-3 times faster than npm.

You can try it out by installing it with [windows](https://yarnpkg.com/en/docs/install#windows-tab) or [linux](https://yarnpkg.com/en/docs/install#linux-tab).

jniles:
In order to get reproducible builds, we need a deterministic build
system like yarn.  Note that this change allows programmers to continue
to use npm alongside yarn without any issue.  However, yarn will be used
on Travis CI, respecting the yarn.lock file.

Closes #880.

For more information, see [Installing Yarn](https://yarnpkg.com/en/docs/install) and [Migrating from NPM to Yarn](https://yarnpkg.com/lang/en/docs/migrating-from-npm/)


sfount (approved):

sfount:
This looks like a step in the right direction! All tests pass so this should be good to go - if any developers have problems with their builds we can review the migration process for this.

jniles:
Hmm... bors seems to have halted.

jniles:
bors r=sfount

jniles:
bors r=sfount

Issue #1942, opened 2017-08-08
Deprecate and remove DepricatedFilterService


jniles:
As part of our sprint on registries, we have attempted to remove all instances of `DepricatedFilterService` in favor of the new `FilterService`.  We should go through and clean out all instances of `DepricatedFilterService`, replacing them with the latest standard to ensure a uniform user experience.

jniles:
@dedrickenc, what is the status of this issue?

DedrickEnc:
@jniles
I was lost, I have one more file to refactor before closing this issue.
I do it today or tomorrow morning. So it still remain opened

jniles:
Cheers, sounds good!  Just trying to clean up old issues :)

PR #1967, opened 2017-08-11
Purchase Order Registry Refactor

### LINKED ISSUE #1964: refactor: improvements to the Purchase Order Registry
The Purchase Order Registry was refactored extensively in #1907 that brought it partially up to standards with the other registries.  Here are some further improvements to bring it completely in line with standards:
 1. Use a dropdown menu to view the document
 2. Make the "Search" text hidden on small screens -- not the icon.
 3. Add inline filtering to the registry
 4. Make the PO.PROJ.XXX identifier into a link to the receipt.

DedrickEnc:
This pull request addresses points listed at #1964. So it fixes : 

- The CSS issue to make search button adapted to screen resize

- Use a dropdown menu to present registry actions

- Set a link between the purchase order and its receipt through the reference column

- A component to choose a supplier on the search modal has been added and its test
closes #1964 
It removes deprecated filter into the purchase order registry also

references #1942 
  
Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


client/src/js/components/bhSupplierSelect.js
```diff
@@ -0,0 +1,44 @@
...
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
+
+    // default for label
+    $ctrl.label = $ctrl.label || 'FORM.LABELS.SUPPLIER';
+
+    // load all Suppliers
+    Suppliers.read()
+      .then(function (suppliers) {        
+        $ctrl.suppliers = suppliers;
+      });
```
jniles: This should have a `catch()` to handle any errors.

client/src/modules/purchases/list/list.html
```diff
@@ -55,15 +55,16 @@
...
         </button>
       </div>
     </div>
   </div>
 </div>
 
-<div class="flex-util bh-filter-bar">
+<div class="flex-util" style="min-height : 35px; padding-top : 7px; max-height: initial">
   <bh-filters
+    style="max-width:90%"
```
jniles: Why do we need to limit the width?

client/src/modules/purchases/list/list.html
```diff
@@ -55,15 +55,16 @@
...
-          <span class="hidden-xs"></span> <span translate>FORM.BUTTONS.SEARCH</span>
+          <span class="fa fa-search"></span> <span class="hidden-xs" translate>FORM.BUTTONS.SEARCH</span>
         </button>
       </div>
     </div>
   </div>
 </div>
 
-<div class="flex-util bh-filter-bar">
+<div class="flex-util" style="min-height : 35px; padding-top : 7px; max-height: initial">
```
jniles: It looks like this style is [already defined in the bh-filter-bar class](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/css/structure.css#L886).  Is there a reason you replaced the `bh-filter-bar` class with a hard-coded style?

client/src/modules/purchases/list/list.js
```diff
@@ -3,7 +3,7 @@ angular.module('bhima.controllers')
 
 PurchaseListController.$inject = [
   '$state', 'PurchaseOrderService', 'NotifyService', 'ReceiptModal',
-  'uiGridConstants', 'GridColumnService', 'DepricatedFilterService',
+  'uiGridConstants', 'GridColumnService', 'SearchFilterFormatService',
```
jniles: You can just remove this -  it is unused.

client/src/modules/purchases/list/list.js
```diff
@@ -147,11 +140,10 @@ function PurchaseListController($state, PurchaseOrder, Notify, Receipts, uiGridC
     PurchaseOrder.openSearchModal(filtersSnapshot)
       .then(function (changes) {
         PurchaseOrder.filters.replaceFilters(changes);
-
         PurchaseOrder.cacheFilters();
         vm.latestViewFilters = PurchaseOrder.filters.formatView();
         return load(PurchaseOrder.filters.formatHTTP(true));
-      });
+      }).catch(angular.noop);
```
jniles: This `.catch()` will never be called since `load()` has its own error handling already.

client/src/modules/purchases/modals/search.tmpl.html
```diff
@@ -15,94 +15,80 @@
...
+            supplier-uuid="$ctrl.searchQueries.supplier_uuid"
+            on-select-callback="$ctrl.onSelectSupplier(supplier)"
+            required="false">
+            <bh-clear on-clear="$ctrl.clear('supplier_uuid')"></bh-clear>
+          </bh-supplier-select>
+    
           <!-- user -->
           <bh-user-select
-            user-id="ModalCtrl.bundle.user_id"
+            user-id="$ctrl.searchQueries.user_id"
```
jniles, line 39: Nice catch :+1:

client/src/modules/purchases/modals/search.tmpl.html
```diff
@@ -15,94 +15,80 @@
...
 
           <!-- status -->
           <div class="form-group">
             <p class="control-label" style="margin-bottom:5px;">
               <strong translate>FORM.LABELS.STATUS</strong>
             </p>
 
-            <label class="radio-inline">
-              <input type="radio" id="is_confirmed" name="state" ng-model="ModalCtrl.setState" value="is_confirmed">
+            <label class="checkbox-inline">
```
jniles: We had a discussion about this [in this comment](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1907#discussion_r131340490) and in [this review](https://github.com/IMA-WorldHealth/bhima-2.X/pull/1907#issuecomment-320877860).

For checkboxes to work, you would expect there to be a case where a purchase order could be both `is_received` and `is_cancelled` (since the inputs are not exclusive).  If this is the case, the SQL should be updated to reflect that.  Otherwise, they should be radio buttons.

As discussed yesterday, we might be removing these states altogether.  I propose you leave these are radio buttons until we figure out how we want to use these properties.

client/src/modules/purchases/templates/action.cell.html
```diff
@@ -0,0 +1,29 @@
...
+  </a>
+
+  <ul data-action="{{ rowRenderIndex}}"  class="dropdown-menu-right" uib-dropdown-menu>
+    <li>
+        <a
+          href
+          data-method="document"
+          ng-click="grid.appScope.getDocument(row.entity.uuid)"
+          data-edit-metadata="{{ row.entity.uuid }}">
+          <i class="fa fa-file-pdf-o"></i> <span translate>FORM.LABELS.DOCUMENT</span>
```
jniles, line 14: :+1: Good use of the translate filter.

server/controllers/finance/purchases.js
```diff
@@ -194,37 +194,10 @@ function create(req, res, next) {
...
-    `;
-  }
-
-  db.exec(sql)
-    .then((rows) => {
-      res.status(200).json(rows);
-    })
-    .catch(next)
-    .done();
+  find(req.query)
```
jniles, line 197: Really nice refactor! This is a win for maintainability. :+1:

jniles (commented):
The code seems pretty good.  I haven't tested the module itself  yet, but will leave a comment if it seems like something is missing alter when I am in a place to test it.

I've made a few comments about cleaning up error handling and styles.  These are also shared with #1969.

It doesn't seem like changing the search from a checkbox to a radio button is appropriate unless we figure out how to account for all types of searches.  This should be:
 1. Look for only canceled purchase orders.
 2. Look for only non-canceled purchase orders
 3. Do not consider the `is_canceled` field.

With checkboxes you can only consider two states - off and on.  For this reason, using checkboxes isn't appropriate.  Additionally, it seems like the "Status" is radio button - so a PO cannot hold more than one state.  For these reasons, a radio button seems appropriate unless the SQL is modified.

What do you think?

client/src/modules/purchases/list/list.html
DedrickEnc: I thought it was a fix to a problem as it is in a file recently reviewed. I will remove it!

client/src/modules/purchases/modals/search.tmpl.html
DedrickEnc: Yes, you are right, these states are exclusive. I am okay to change them.

jniles (approved):
LGTM!

PR #1971, opened 2017-08-12
(refactor) Entity Reference Links

sfount:
This commit removes the current implementation of a UI grid cell template
link and replaces it with a component. This component allows the programmer
to set a reference and a display value that can be shown to the user.

Advantages of a component design in this case:
* Can be used interchangeably in and out of UI grids
* Every controller that wants to make use of reference links does not have to expose languages
* Can factor in global settings to link internally or redirect (documents
  vs. search
* Reads more naturally in markup
* Optionally display different text to the user than the reference
* Link behaviours can be updated in one central file rather than at every
  implementation

**Usage example**
```html
<!-- link to a transactions entity document but show the transaction description -->
<bh-reference-link
  reference="transaction.entity"
  display="transaction.description">
</bh-reference-link>

<!-- link to a hard-coded known reference -->
<bh-reference-link
  reference="IV.HSP.200">
</bh-reference-link>
```


jniles (approved):
LGTM. This will definitely speed up the process of linking entities everywhere. :+1:

PR #1973, opened 2017-08-14
Remove the SearchFilterFormatService form lot registry

DedrickEnc:
In order to remove the deprecated filter, this P.R removes the use of ```SearchFilterFormatService``` in the lot registry ans search modal of journal ledger.
This service reference the deprecated filter.


jniles (approved):
LGTM!  I haven't tested it, but assume it all still works.

PR #1977, opened 2017-08-14
Unifying date components

jeremielodi:
Unifying date components in patient registry


jniles:
Ping @sfount.

jniles (approved):
Looks good to me!  I've tried it an it works well.

PR #1981, opened 2017-08-14
refactor(bhDateEditor): use onChange() callback

### LINKED ISSUE #1960: refactor: modernize bhDateEditor
The [bhDateEditor](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/components/bhDateEditor.js) component is one of the core components used in every single page.  It was one of the first components completed, and the design has aged.  Here are some refactors to bring it up to date with standards:
 1. Use the translate directive instead of the translate filter.
 2. Initialize label in the `$onInit` method.
 3. BREAKING CHANGE: remove the two-way bound data and use an `onChange()` to update the parent controller's date input.
 4. Use the `$$parent` form property to validate the date input.
 5. Use font-awesome instead of glyphicons.

jniles:
BREAKING CHANGE

This commit migrates the bhDateEditor component to an onChange() style
of callbacks.  This is the recommended Angular 2-style syntax, and
allows the component bindings to dictate the flow of data from `<` and
`@` bindings into the component to `&` for returning data via callbacks.

Additionally, the bhDateEditor has been modernized with `translate`
directives instead of the filter, and everything has been moved into the
$onInit method.

Closes #1960.

----

If this style of component gets reviewed and approved, _all_ instances of the `bhDateEditor` will need to be updated.  I've put this PR on a feature branch so that anyone can contribute to its progress.

### DIFF
--- a/client/src/js/components/bhDateEditor.js
+++ b/client/src/js/components/bhDateEditor.js
@@ -1,19 +1,19 @@
 angular.module('bhima.components')
-.component('bhDateEditor', {
-  templateUrl : '/modules/templates/bhDateEditor.tmpl.html',
-  controller  : dateEditorController,
-  bindings    : {
-    dateValue         : '=', // two-way binding
-    minDate           : '<', // one-way binding
-    maxDate           : '<', // one-way binding
-    validationTrigger : '<', // one-way binding
-    disabled          : '<', // one-way binding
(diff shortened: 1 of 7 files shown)


jniles:
@sfount, can you glance at this code and see if you think it is worth moving in this direction?

PR #1982, opened 2017-08-14
altering price_list_item.value to Double instead of Integer

jeremielodi:
This PR help to accept double price values


jniles:
@jeremielodi good work!

For a change like this, we should include a test.  Could you write an integration test for this?  You will need to add it to [test/integration/priceList.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/integration/priceList.js).

It should look something like this:
```js

const somePriceListWithFloatValues = {
  // ... add your own values here
  items : [{
    inventory_uuid : 'some-inventory-uuid',
    value : 3.14
    // other properties.
};

it('will register a float as a price list item', () => {
  return agent.post('/prices')
    .send({ list : somePriceListWithFloatValues })
    .then(res => {
      // assert that the records was successfully created
      helpers.api.created(res);

      // look up the record to make sure the price list's value is actually a float
      return agent.get('/prices/'.concat(res.body.uuid));
    })
    .then(res => {
       // ... do some checks ...
       expect(res.body.items[0].value).to.equal(3.14);
     });
});
```

[Read more about creating integration tests here.](https://github.com/IMA-WorldHealth/bhima-2.X/wiki/Testing:-Integration-Testing-Server-APIs)

test/integration/priceList.js
```diff
@@ -12,70 +12,105 @@ describe('(/prices ) Price List', function () {
...
   }, {
     inventory_uuid : 'c48a3c4b-c07d-4899-95af-411f7708e296',
     label : 'Test 13% reduction on an item',
     is_percentage : true,
-    value : 12
+    value : 12,
   }];
+  const floatPriceListItems = [
+    {
+      inventory_uuid : '289cc0a1-b90f-11e5-8c73-159fdc73ab02',
```
jniles, line 31: :+1:

jniles (approved):
Looks like tests passed!  Could you update the description of this so it automatically closes the issue?

After that, I'll merge this.

jniles:
Well done!

PR #1983, opened 2017-08-14
fixing patient registration cashflow report pdf when non record match

### LINKED ISSUE #1953: (bug) Patient Registrations PDF crashes when no matching records
If you try to print the Patient Registry without any data matching the filters, the PDF report is returned to the client, but does not render as a PDF.  This leads to a really confusing user experience.  It also means that if we ever allow someone to download the PDF without first going to the registry, they may be thoroughly confused about what is going on since they will not realize that there are not records matching their criteria.

patientregistrationsfailedrequest
_Fig 1: A Failed Request for the Patient Registry_

The bug exists in [the aggregation SQL](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/medical/reports/registrations.js#L117), which fails with the following syntax error:
```sql
-- this is executed
SELECT COUNT(patient.uuid) AS numPatients, MIN(patient.created_at) AS minDate, MAX(patient.created_at) AS maxDate,
      COUNT(DISTINCT(DATE(patient.created_at))) AS numDays,
      SUM(sex = 'F') AS numFemales, ROUND(SUM(sex = 'F') / COUNT(patient.uuid) * 100) AS percentFemales,
      SUM(sex = 'M') AS numMales, ROUND(SUM(sex = 'M') / COUNT(patient.uuid) * 100) AS percentMales,
      COUNT(DISTINCT(patient.current_location_id)) AS numDistinctResidences,
      COUNT(DISTINCT(debtor.group_uuid)) AS numDebtorGroups
    FROM patient JOIN debtor ON patient.debtor_uuid = debtor.uuid
    WHERE patient.uuid IN ();
```

> debug:  Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')' at line 7

We should make sure that ever registry can render even with no data in it to provide a decent user experience.

jeremielodi:
This PR help to fixe the patient registration cashflow report pdf when non record match.

Closes #1953.

### DIFF
--- a/server/controllers/medical/reports/registrations.js
+++ b/server/controllers/medical/reports/registrations.js
@@ -118,7 +118,7 @@ function build(req, res, next) {
 
       // if no patients matched the previous query, set the promise value to false
       // and skip rendering aggregates in the handlbars view
-      if (!patients) { return false; }
+      if (patients.length === 0) { return false; }
 
       // gather the uuids for the aggregate queries
       const uuids = patients.map(p => db.bid(p.uuid));


jniles (approved):
Looks good to me!

jniles:
Added "Closes #1953" to make sure the merge closes the old issue.

PR #1989, opened 2017-08-16
updating medical module by using translate directives

jeremielodi:
This PR helps to mirate the medical module by using translate directives instead of  translate filters

### DIFF
--- a/client/src/modules/patients/record/units/checkin.html
+++ b/client/src/modules/patients/record/units/checkin.html
@@ -1,7 +1,7 @@
 <div class="panel panel-default">
 
-  <div class="panel-heading clearfix">
-    {{ "PATIENT_RECORDS.CHECK_IN.TITLE" | translate }}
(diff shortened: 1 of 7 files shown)


jniles:
@jeremie, could you please include a "Closes #" tag to close the relevant issue?

client/src/modules/patients/record/units/checkin.html
```diff
@@ -1,7 +1,7 @@
 <div class="panel panel-default">
 
-  <div class="panel-heading clearfix">
-    {{ "PATIENT_RECORDS.CHECK_IN.TITLE" | translate }}
+  <div class="panel-heading clearfix" translate>
```
jniles: I don't think this will work.  Try something like this:

```html
<span translate>PATIENT_RECORDS.CHECK_IN.TITLE</span>
```

jniles:
This looks good to me!

PR #1991, opened 2017-08-16
Fix(Invetory): adding loading indicator into the inventory list

jeremielodi:
This PR helps to add a loading indicator to the inventory list

### DIFF
--- a/client/src/modules/inventory/list/list.js
+++ b/client/src/modules/inventory/list/list.js
@@ -31,7 +31,8 @@ function InventoryListController($translate, Inventory, Notify, uiGridConstants,
   vm.filterEnabled = false;
   vm.gridOptions = {};
   vm.gridApi = {};
-  vm.loading = false;
+
+  vm.loading = true;
 
 
   // grid default options
@@ -162,8 +163,6 @@ function InventoryListController($translate, Inventory, Notify, uiGridConstants,
       .finally(function () {
         vm.loading = false; // this will execute after the data is downloaded.
       });
-
-    vm.loading = false;
   }
 
 
--- a/client/src/modules/inventory/list/list.js
+++ b/client/src/modules/inventory/list/list.js
@@ -155,15 +155,18 @@ function InventoryListController($translate, Inventory, Notify, uiGridConstants,
       .then(function (rows) {
         vm.gridOptions.data = rows;
       })
-      .catch(Error);
+      .catch(function (exception) {
+        vm.hasError = true;
+        Notify.handleError(exception);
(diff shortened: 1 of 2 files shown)


client/src/modules/inventory/list/list.js
```diff
@@ -138,20 +140,30 @@ function InventoryListController($translate, Inventory, Notify, uiGridConstants,
...
+    vm.hasError = false;
+
     Inventory.read(null, params)
       .then(function (rows) {
         vm.gridOptions.data = rows;
       })
-      .catch(Notify.handleError);
+      .catch(Error);
+
+    vm.loading = false;
```
jniles: Does the loading indicator work?  It seems like this will not wait for the data to download.

The correct way to wait for the data to download is to use the .finally() Promise method.  Like this:
```js
.catch(Notify.handleError)
.finally(function () {
  vm.loading = false; // this will execute after the data is downloaded.
});
```

client/src/modules/inventory/list/list.js
```diff
@@ -138,20 +140,30 @@ function InventoryListController($translate, Inventory, Notify, uiGridConstants,
...
   }
 
+  function Error() {
```
jniles: [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) is a reserved word in JavaScript.  Could you name this function something else?

client/src/modules/inventory/list/list.js
```diff
@@ -138,20 +140,30 @@ function InventoryListController($translate, Inventory, Notify, uiGridConstants,
...
+    vm.hasError = true;
+    Notify.handleError();
```
jniles: `Notify.handleError()` takes in a parameter - the error sent from the server.  It should be passed to thus function.

jniles (commented):
@jeremielodi, this is a good first attempt.  The component is exactly what the issue asked for! :+1:

Unfortunately, this will never show - the loading indicator should turn on, wait for the data to load, and turn off.  You can test this by setting the network tab in your chrome developer tools to us 'slow 3G' connections.  The correct way to turn off the loading is shown above - either when the server has sent the data or an error.

Additionally, the handle error method should be renamed to make sure you don't break JS's error class.

jniles:
It would also be good to put "fixes #" and reference the issue so that this closes the issue when merged....

jniles:
@jeremielodi, let me know when you ready for another review.

jniles (approved):
LGTM.

PR #1992, opened 2017-08-16
refactor: Trial Balance

jniles:
This PR refactors the Trial Balance to simplify the API, separate JS and SQL, speed up SQL queries, and improve the UI of the module.  The major features that land in this are:

**Client Changes**
 1. Link errors and accounts from the Trial Balance into the Posting Journal
 2. Only download the Trial Balance data once, and ensure Error/Overview states can be rapidly toggled.
 3. Remove the $state-based passing of records, data is now stored in TrialBalanceService
 4. Remove the Details Trial Balance state - this will eventually be shown in an expanded row.
 5. Use bootstrap colors for success/error states
 6. Remove multiple header/footer templates, combine into a single parent `TrialBalance` controller and children `TrialBalanceOverview` and `TrialBalanceErrors` controllers.
 7. Moved the exporter link into modal header.
 8. Introduce `bhJournalLink` for automatically linking to record_uuids in the Journal.

**Server Changes**
 1. Move all Trial Balance operations into SQL stored procedures:
   a. `StageTrialBalanceTransaction()`  - moves data into a temporary table.
   b. `TrialBalanceSummary()` - produces the before, debit, credit, after summary table
   c. `TrialBalanceErrors()` - produces an error list of problematic transactions
   c `PostToGeneralLedger()` - posts data to the GL, removing it from the PJ
 2. Allow filtering on Record UUIDs in the posting journal
 3. Refactor the API to be `/journal/trialbalance` and `/journal/transactions` for posting.  These can be altered as required in the future.

**TODOs**
 - [ ] Integration tests still need to be updated.
 - [ ] Translations need to be finalized.

**Screen Shots**
trialbalancesuccess
_Fig 1: Trial Balance Balanced State_

trialbalancewarningoverview
_Fig 2: Trial Balance Error Overview_

trialbalanceerrorview
_Fig 3: Trial Balance Error State_


client/src/js/services/FilterService.js
```diff
@@ -73,12 +73,18 @@ function FilterService(Store) {
   };
 
   // alias for `assignFilters`, clears the currently active filters before
-  // calling the erferenced method
```
sfount, line 76: 👍

client/src/less/bhima-bootstrap.less
```diff
@@ -262,6 +262,20 @@ input[type=number] {-moz-appearance: textfield;}
   animation-duration: 2s;
 }
 
+/* These classes need to be "!important" because ui-grid's header cell template */
```
sfount, line 265: Very useful clarification!

client/src/modules/journal/trial-balance/overview.js
```diff
@@ -0,0 +1,208 @@
...
+
+  vm.subGridOptions = {
+    enableColumnMenus : false,
+    showColumnFooter  : false,
+    fastWatch : true,
+    flatEntityAccess : true,
+    columnDefs : subGridColumns,
+  };
+
+  // FIXME(@jniles) - this is kind of hacky
```
sfount, line 145: 🚢

server/controllers/finance/trialBalance/index.js
```diff
@@ -1,126 +1,99 @@
...
 
-  transactions.forEach((transaction) => {
-    txn.addQuery('CALL StageTrialBalanceTransaction(?);', transaction);
-  });
+  // stage all trial balance transactions
+  stageTrialBalanceTransactions(txn, transactions);
 
-  txn.addQuery('CALL TrialBalance()');
+  // compute any trial balance errors
+  txn.addQuery('CALL TrialBalanceErrors()');
```
sfount, line 49: This is a really clear API 👍

server/controllers/finance/trialBalance/index.js
```diff
@@ -1,126 +1,99 @@
...
 
   return txn.execute()
-    .then((errors) => {
-      const lastIndex = errors.length - 1;
-      res.status(201).json(errors[lastIndex][0]);
+    .then((results) => {
+      // because we do not know the number of stageTrialBalance() calls, we must index back by two
+      // to get to the CALL TrialBalanceErrors() query and one for the Call TrialBalanceSummary()
+      // query.
+      const errorsIndex = results.length - 2;
```
sfount, line 59: Depending on what you need to do with the array after this; if the values you need are guaranteed at the end this could just use `pop`, for example: 
```js
// remove the last two elements, summary and errors queries 
const data = { 
  summary : results.pop()
  errors : results.pop()
};
```

Having read this I'm not even sure it's clearer - just a thought!

server/models/procedures.sql
```diff
@@ -1227,17 +1222,42 @@ END $$
 
 -- this writes transactions to a temporary table
 CREATE PROCEDURE StageTrialBalanceTransaction(
-  IN trans_id TEXT
+  IN record_uuid BINARY(16)
```
sfount, line 1225: This should help reduce corner cases/ ambiguity in the future.

sfount (approved):
This PR greatly improves UX for the trial balance significantly, as well as cleaning up the database code, this is great work. 

There are some hanging features and a few lines that need to be cleaned up but this should be merged to continue this feature branch.

server/controllers/finance/trialBalance/index.js
jniles: That's interesting, and certainly more terse!  I think that, for the sake of clarify, we should keep the long form.  It takes a careful set of eyes to realize that `results` is mutating in your example with ever function call!

PR #1993, opened 2017-08-17
chore(tests): connect to sauce labs

jniles:
This commit attempts to connected to Sauce Labs to run e2e tests.  It is thoroughly experimental at this point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+
 dist: trusty
 
 before_install:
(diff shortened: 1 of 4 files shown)


jniles:
I'll continue working on this PR over the next couple weeks (it is fairly self-contained).  But I'm pretty sure we will need less brittle tests (less inter-dependent) to make sure this works properly.

There are a few cool advancements unrelated to Sauce Labs here - log folding with Travis, using container images for faster boots, etc.  I'll break those out into a separate PR which can be merged today.

jniles:
This got so close.  We'll give it another push in a couple of months.

PR #1994, opened 2017-08-17
fix(tests): update Trial Balance integration tests

jniles:
This commit updates the Trial Balance integration tests to the latest API.


sfount (approved):
LGTM

Issue #1721, opened 2017-05-31
feat: Inventory search modal requires more parameters


jniles:
The Inventory Search Modal is quite limited.  This is the current version on master:
![inventorysearchform](https://cloud.githubusercontent.com/assets/896472/26630578/418f966a-45ff-11e7-939f-e0f9260015bc.png)
_Fig 1: The most recent inventory search modal form_

We are missing the following fields:
 1. `code` - the unique inventory item code
 2. `price` - the sales price
 3. `inventory_type` - the type of inventory
 4. `inventory_unit` - the unit of the inventory
 5. `consumable` - a flag to see if it is consumable

These inputs should be implemented with the latest `FilterService` on both the client and server.  They should also take into account #1702.

jeremielodi:
Closed in #1996.

PR #1996, opened 2017-08-17
adding filters parameter into stock inventories report

jeremielodi:
This PR helps to add filter parameters into the stock inventories report, in order to discribe the result

### DIFF
--- a/server/controllers/stock/reports/index.js
+++ b/server/controllers/stock/reports/index.js
@@ -13,6 +13,7 @@ const identifiers = require('../../../config/identifiers');
 const NotFound = require('../../../lib/errors/NotFound');
 const ReportManager = require('../../../lib/ReportManager');
 
+const PeriodService = require('../../../lib/period');
 const Stock = require('../core');
 
 const BASE_PATH = './server/controllers/stock/reports';
@@ -35,6 +36,42 @@ const STOCK_INVENTORIES_REPORT_TEMPLATE = `${BASE_PATH}/stock_inventories.report
 const STOCK_INVENTORY_REPORT_TEMPLATE = `${BASE_PATH}/stock_inventory.report.handlebars`;
 
 // ===================================== receipts ========================================
+
+/*
+* This function help to format filter display name
+* Whitch must appear in the report
+*/
+function formatFilters(qs) {
+
+  const columns = [
+    { field : 'depot_uuid', displayName : 'STOCK.DEPOT' },
+    { field : 'inventory_uuid', displayName : 'STOCK.INVENTORY' },
+    { field : 'status', displayName : 'FORM.LABELS.STATUS' },
+    { field : 'defaultPeriod', displayName : 'TABLE.COLUMNS.PERIOD', isPeriod : true },
+    { field : 'period', displayName : 'TABLE.COLUMNS.PERIOD', isPeriod : true },
+    { field : 'limit', displayName : 'FORM.LABELS.LIMIT' },
+
+    { field : 'entry_date_from', displayName : 'STOCK.ENTRY_DATE', comparitor : '>', isDate : true },
+    { field : 'entry_date_to', displayName : 'STOCK.ENTRY_DATE', comparitor : '<', isDate : true },
+  ];
+
+  return columns.filter(column => {
+
+    const value = qs[column.field];
+
+    if (!_.isUndefined(value)) {
+
+      if (column.isPeriod) {
+        const service = new PeriodService(new Date());
+        column.value = service.periods[value].translateKey;
+      } else {
+        column.value = value;
+      }
+      return true;
+    }
+    return false;
+  });
+}
 
 /**
  * @method stockExitPatientReceipt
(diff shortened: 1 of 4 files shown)


server/controllers/stock/reports/index.js
```diff
@@ -592,9 +592,29 @@ function stockLotsReport(req, res, next) {
...
   const data = {};
   let report;
+
+// getting params
+  var defaultParam = req.query;
+  try {
+
+    const keys = Object.keys(defaultParam);
+    keys.forEach((key) => {
+      if ((key !== 'lang') && (key !== 'renderer')) {
```
jniles: Hmm.... this isn't really extensible.  Also, this method doesn't allow keys to be translated.

Why don't you try using something like like the `formatFilters()` function from the patient registration report?  It lists every key and it's translation value ... like this:
```js
// translation key mappings for dynamic filters
// Basically, to show a pretty filter bar, this will translate URL query params
// into human-readable text to be placed in the report, showing the properties
// filtered on.
function formatFilters(qs) {
  const columns = [
    { field : 'name', displayName : 'FORM.LABELS.NAME' },
    { field : 'sex', displayName : 'FORM.LABELS.GENDER' },
    { field : 'hospital_no', displayName : 'FORM.LABELS.HOSPITAL_NO' },
    { field : 'reference', displayName : 'FORM.LABELS.REFERENCE' },
    { field : 'dateBirthFrom', displayName : 'FORM.LABELS.DOB', comparitor : '>', isDate : true },
    { field : 'dateBirthTo', displayName : 'FORM.LABELS.DOB', comparitor : '<', isDate : true },
    { field : 'dateRegistrationFrom', displayName : 'FORM.LABELS.DATE_REGISTRATION', comparitor : '>', isDate : true },
    { field : 'dateRegistrationTo', displayName : 'FORM.LABELS.DATE_REGISTRATION', comparitor : '<', isDate : true },
    { field : 'debtor_group_uuid', displayName : 'FORM.LABELS.DEBTOR_GROUP' },
    { field : 'patient_group_uuid', displayName : 'PATIENT_GROUP.PATIENT_GROUP' },
    { field : 'user_id', displayName : 'FORM.LABELS.USER' },
    { field : 'limit', displayName : 'FORM.LABELS.LIMIT' },
    { field : 'period', displayName : 'TABLE.COLUMNS.PERIOD', isPeriod : true },
  ];

  return columns.filter(column => {

    const value = qs[column.field];

    if (!_.isUndefined(value)) {
  
      if (column.isPeriod) {
        const service = new PeriodService(new Date());
        column.value = service.periods[value].translateKey;
      } else {
        column.value = value;
      }
      return true;
    }
    return false;
  });
} 
```

This format has the following advantages:
 1. It's clear which filters are supported and which are not.
 2. The display name can be anything found in the translation dictionary.
 3. No need to `try/catch` since the code will ignore anything that isn't in the supported filters object.  
 4. Returns an array as required.
 5. Allows any additional properties to be added to a given filter (like the `isDate` property or the `comparitor` property).

server/controllers/stock/reports/index.js
```diff
@@ -592,9 +592,29 @@ function stockLotsReport(req, res, next) {
...
+        filters.push({ displayName : key, value : defaultParam[key] });
+      }
+    }, keys);
+    // filters paramter found
+    hasFilter = filters.length > 0;
+
+  } catch (error) {
```
jniles: Errors should almost always be handled in some way in production.  There has to be a really good reason to have a blank `catch`.  At minimum, we should log something using `winston.error()` to let the application developers know that something when wrong.

jniles:
@jeremielodi, let me know when you need a second review.

jeremielodi:
Hi! Try just to run branch and see the result, if it good I'll just remove
unimportant codes in handlebar files and push the last changes

jniles:
@jeremielodi, please push the latest changes for me to analyse.  It doesn't do any good for me to look at old material.  If those changes are not ready, we can wait on the review until they are ready.

server/controllers/stock/reports/index.js
```diff
@@ -677,17 +720,25 @@ function stockMovementsReport(req, res, next) {
...
   } catch (e) {
     return next(e);
   }
 
+
   return Stock.getLotsMovements(null, options)
     .then((rows) => {
       data.rows = rows;
       data.hasFilter = hasFilter;
+      data.filters = filters;
```
jniles: This code is no longer necessary.  Please remove it.

server/controllers/stock/reports/index.js
```diff
@@ -677,17 +720,25 @@ function stockMovementsReport(req, res, next) {
       hasFilter = Object.keys(display).length > 0;
     }
 
+    filters = Object.keys(display).map(function (key) {
```
jniles: This code is no longer necessary.  Please remove it.

server/controllers/stock/reports/index.js
```diff
@@ -592,9 +629,11 @@ function stockLotsReport(req, res, next) {
   let options = {};
   let display = {};
   let hasFilter = false;
+  let filters = [];
```
jniles: Do you need this array?  It looks like it is never used.

jniles (changes requested):
It looks like there is still some legacy code in here that should be removed.  After that, LGTM.

jeremielodi:
closes #1721

PR #1999, opened 2017-08-17
(Feature) Journal posted transactions

sfount:
This commit updates the queries for the journal on the server using
a union between two known methods to combine substets. It also removes
aggregation logic that was used in legacy journal editing.

This commit visually differentiates between posted and non-posted transactions. It also adds a footer row that displays the following totals: 
* number of rows 
* number of posted transactions 
* number of non-posted transactions 
* number of remaining system non-posted transactions not shown.


client/src/modules/journal/modals/search.modal.js
```diff
@@ -22,11 +22,13 @@ function JournalSearchModalController(Instance, Projects, Notify,
 
   // @TODO ideally these should be passed in when the modal is initialised
   //       these are known when the filter service is defined
-  var searchQueryOptions = ['description', 'user_id', 'account_id', 'project_id', 'amount', 'trans_id', 'origin_id'];
+  var searchQueryOptions = ['description', 'user_id', 'account_id', 'project_id', 'amount', 'trans_id', 'origin_id', 'includeNonPosted'];
 
   // assign already defined custom filters to searchQueries object
   vm.searchQueries = util.maskObjectFromKeys(filters, searchQueryOptions);
 
+  window.search = vm.searchQueries;
```
jniles, line 30: This should be removed before we forget it happened.

client/src/modules/journal/templates/grid.footer.html
```diff
@@ -1,4 +1,6 @@
-<div style="padding: 0.5rem; font-weight: bold;">
-  <span translate>FORM.INFO.SHOWING</span>
-  {{ grid.appScope.numberCurrentGridTransactions }} <span translate>FORM.INFO.OF</span> {{ grid.appScope.numberTotalSystemTransactions }} <span translate>FORM.INFO.UNPOSTED_TRANSACTION</span>
+<div class="ui-grid-cell-contents">
+  <span>({{grid.rows.length}} <span translate>FORM.INFO.ROWS</span>)</span>
+  <span style="color : #FDBA64" class="fa fa-circle"></span> {{grid.appScope.rowsDetails.totalPosted}} <span translate>FORM.LABELS.POSTED_TRANSACTIONS</span>
```
jniles, line 3: Can this color be a class?  That way we can change it in a single place if we never need to change it in the future ... such as `bhima-bootstrap.less`.

Also, since it's shared in multiple places (and presumably will be shared in the Account Statement and others), putting in a class early will make sure we have consistency.

server/controllers/finance/journal/index.js
```diff
@@ -38,70 +38,68 @@ exports.editTransaction = editTransaction;
...
+  const options = {
+    record_uuid : recordUuid,
+    includeNonPosted : true
+  };
+
+  return find(options)
     .then((result) => {
       // if no records matching, throw a 404
-      if (result.journal.length === 0) {
+      if (result.length === 0) {
```
jniles, line 56: :dancer:

jniles:
Jotted some notes to remind us to fix stuff before merge into `master`.

client/src/modules/journal/modals/search.modal.js
sfount: 😞

PR #2002, opened 2017-08-18
(Fix) Reference links component

sfount:
This commit ensures that the reference link component works in a UI grid
when underlying data is changed. UI Grid optimises the creation of components by
updating existing components instead of replacing them. This requires the
component to act and change the display label on change - it cannot assume
the reference value passed will be static.

### DIFF
--- a/client/src/js/components/bhReferenceLink.js
+++ b/client/src/js/components/bhReferenceLink.js
@@ -2,7 +2,7 @@ angular.module('bhima.components')
 .component('bhReferenceLink', {
   bindings : {
     reference : '<',
-    display : '@?' 
+    display : '@?'
   },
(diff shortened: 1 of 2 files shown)


client/src/js/components/bhReferenceLink.js
```diff
@@ -2,22 +2,36 @@ angular.module('bhima.components')
...
-  
+
   $ctrl.languageKey = Languages.key;
+  $ctrl.displayLabel = '-';
+
+  $ctrl.$onInit = function onInit() {
+    $ctrl.displayLabel = $ctrl.display || $ctrl.reference;
+  };
+
+  $ctrl.$onChanges = function onChanges(changes) {
```
jniles, line 25: Nice :+1:

jniles (approved):
Good catch!  Looks good to me!

PR #2006, opened 2017-08-18
(refactor) Journal transaction editing API: delete all rows case

sfount:
This commit handles the delete all rows case for the server transaction
editing API. If the client tools sent removed rows that are greater
than the number of initial rows this will now result in a 400.

This should eventually encourage developers and then users to use the
proposed delete transaction API.


jniles (approved):
LGTM

PR #2022, opened 2017-08-21
fix(journal): exchange debit/credit fields on edit

jniles:
This commit updates the debit/credit fields to compute their values
based on the exchange rate of the day.  This makes sure the debit/credit
values are always up to date with their debit_equiv/credit_equiv
counterparts.

### DIFF
--- a/server/controllers/finance/journal/index.js
+++ b/server/controllers/finance/journal/index.js
@@ -62,8 +62,8 @@ function lookupTransaction(recordUuid) {
     });
 }
 
-// @TODO(sfount) find a more effecient way of combining multiple table sets than a union all on the final results
-//               - new method should be proven as more effecient on large data sets before being accepted
+// @TODO(sfount) find a more efficient way of combining multiple table sets than a union all on the final results
+//               - new method should be proven as more efficient on large data sets before being accepted
 //
 // Current merge logic : subset 1 UNION ALL subset 2 ORDER
 // 1. select all from the posting journal including all joins, conditions etc.
@@ -71,8 +71,9 @@ function lookupTransaction(recordUuid) {
 // 3. UNION ALL between both complete sets of data
 // 4. Apply date order
 function naiveTransactionSearch(options, includeNonPosted) {
-  // hack to ensure only the correct amount of rows are returned - this should be improved in the more effecient method of selection
-  var limitCondition = '';
+  // hack to ensure only the correct amount of rows are returned - this should be improved
+  // in the more efficient method of selection
+  let limitCondition = '';
   if (options.limit) {
     limitCondition = ` LIMIT ${Number(options.limit)}`;
   }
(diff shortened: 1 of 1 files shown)


server/controllers/finance/journal/index.js
```diff
@@ -300,10 +305,19 @@ function editTransaction(req, res, next) {
...
   const ACCOUNT_NUMBER_QUERY = 'SELECT id FROM account WHERE number = ?';
   const ENTITY_QUERY = 'SELECT uuid FROM entity_map WHERE text = ?';
   const REFERENCE_QUERY = 'SELECT uuid FROM document_map  WHERE text = ?';
+  const EXCHANGE_RATE_QUERY = `
+    SELECT ? * IF(enterprise.currency_id = ?, 1, GetExchangeRate(enterprise.id, ?, ?)) AS amount FROM enterprise
+    JOIN project ON enterprise.id = project.enterprise_id WHERE project.id = ?;
+  `;
+
+  // these are global/shared properties of the current transaction
+  const projectId = currentTransaction[0].project_id;
```
sfount: It looks like both the client and the server logic are making use of which properties are shared on each row in a transaction - potentially it would good to define these columns in a constant definition that can be used by both the client and server. (transaction shared vs. transaction unique columns).

server/controllers/finance/journal/index.js
```diff
@@ -251,6 +253,9 @@ function editTransaction(req, res, next) {
     .then((currentTransaction) => {
       const posted = currentTransaction[0].posted;
 
+      // bind the current transaction under edit as "oldTransaction"
+      oldTransaction = currentTransaction;
```
sfount: It would be good to use one name here with a comment description of what this actually represents - throughout this code `oldTransaction` and `currentTransaction` are the same entity but it is hard to follow the logic. 

I think `currentTransaction` is misleading because it ambiguously refers to either the current journal transaction or the current edited transaction. `oldTransaction` may be more appropriate.

sfount (commented):
Code LGTM. This is a good patch, simply assigning an enterprise currency amount to transactions made in foreign currencies as before was a serious bug. 

In the future it would be good to allow the user to edit the actual debit and credit columns from the editing tools - this edit could be to represent that 10,000FC was paid instead of 5,000FC - given this system the value would have to be back calculated to the enterprise currency. At least now it is possible!

server/controllers/finance/journal/index.js
jniles: I'll update references to make use of `oldTransaction` and `_oldTransaction` for when it just needs to be exposed to a larger scope.

server/controllers/finance/journal/index.js
jniles: Do you want these done in this PR or can it be a `// TODO`?

jniles:
@sfount, changes made!

server/controllers/finance/journal/index.js
sfount: This can be done in a separate PR, I think it can be applied to many places so should be carefully laid out. A `TODO` would be ideal

sfount:
@jniles LGTM

PR #2023, opened 2017-08-21
(test) Fix transaction editing tests

sfount:
This commit updates the E2E that validate transaction editing to use the
latest modal editing structure. It also adds required IDs or data-methods to
transaction editing modal to allow tests to interact with components.


jniles:
@sfount, looks like everything is green!  I'm merging.

PR #2026, opened 2017-08-21
(feat): using user.display_name in pdf_report instead of user.username

### LINKED ISSUE #2014: refactor: Use real user name in the "Produced By" field of PDF headers
Our PDFs generally contain headers that give information about the date it was created and the person who initiated it.  If we intend these to be official documentation, we should use the user's first and last names (`user.display_name`) instead of their username --- which often times is meaningless.

producedbyusername
_Fig 1: Who was this document produced by?_

jeremielodi:
This PR helps to use user display_name in pdf report that he has initiated instead of username
 closes #2014


jniles:
Nice commit message!  I'll pull this and check it out locally. :+1:

jniles (approved):
LGTM :+1:

PR #2030, opened 2017-08-22
fix(journal): block Trial Balance for posted txns

jniles:
This commit blocks the Trial Balance with a warning to the user if the user has selected any posted records to perform a Trial Balance.  The Trial Balance only makes sense on unposted records.


sfount (approved):
LGTM.

PR #2033, opened 2017-08-22
refactor(depot): Apply 2.X standards for depot management

mbayopanda:
This PR refactor the depot management module by applying the 2.X standards relatives to Grid and Modal.

### DIFF
--- a/client/src/modules/depots/depots.js
+++ b/client/src/modules/depots/depots.js
@@ -1,7 +1,10 @@
 angular.module('bhima.controllers')
 .controller('DepotManagementController', DepotManagementController);
 
-DepotManagementController.$inject = ['$translate', 'DepotService', 'SessionService', 'util'];
+DepotManagementController.$inject = [
+  'DepotService', 'SessionService', 'ModalService',
+  'NotifyService', 'uiGridConstants', '$state',
+];
 
 /**
  * Depot Management Controller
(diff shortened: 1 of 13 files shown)


mbayopanda:
@jniles @DedrickEnc 
Could you review this PR ?

client/src/modules/depots/depots.js
```diff
@@ -1,154 +1,106 @@
...
-    }
+  // update an existing depot
+  function editDepot(depotObject) {
+    $state.go('depots.edit', { depot : depotObject });
   }
 
-  function refreshValidation() {
-    vm.state.errored = vm.depot.text ? false : true;
+  // create a new depot
+  function createDepot() {
```
jniles, line 101: Instead of having a `createDepot()` function, you could easily just have a `<a ui-sref="depots.create"></a>`.  It would reduce the amount of JS code ... but I'm not sure the best practice here.

client/src/modules/depots/depots.js
```diff
@@ -1,154 +1,106 @@
...
 /**
  * Depot Management Controller
  *
  * This controller is about the depot management module in the admin zone
  * It's responsible for creating, editing and updating a depot
  */
-function DepotManagementController($translate, DepotService, SessionService, util) {
-  'use strict';
-
+function DepotManagementController(Depots, Session, ModalService,
```
jniles: Do you still need `Session`?

client/src/modules/depots/modals/depot.modal.js
```diff
@@ -0,0 +1,42 @@
+angular.module('bhima.controllers')
+  .controller('DepotModalController', DepotModalController);
+
+DepotModalController.$inject = [
+  '$state', 'ServiceService', 'DepotService',
+  'SessionService', 'ModalService', 'NotifyService',
+];
+
+function DepotModalController($state, Services, Depots,
+  SessionService, ModalService, Notify) {
```
jniles: Do you use SessionService here?

client/src/modules/depots/depots.routes.js
```diff
@@ -1,9 +1,42 @@
...
+      url         : '/depots',
+      controller  : 'DepotManagementController as DepotCtrl',
+      templateUrl : 'modules/depots/depots.html',
+    })
+
+    .state('depots.create', {
+      url : '/create',
+      params : {
+        depot : { value : null },
+        creating : { value : true },
```
jniles, line 14: Just FYI, I think this would work equally well as
```js
params : {
  depot : null,
  creating : true,
}
```

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -0,0 +1,48 @@
...
+
+    <div class="form-group" ng-class="{ 'has-error' : DepotForm.$submitted && DepotForm.text.$invalid }">
+      <label class="control-label" translate>DEPOT.LABEL</label>
+      <input name="text" ng-model="DepotModalCtrl.depot.text" autocomplete="off" class="form-control" required>
+      <div class="help-block" ng-messages="DepotForm.text.$error" ng-show="DepotForm.$submitted">
+        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      </div>
+    </div>
+
+    <div class="form-group">
```
jniles: This HTML should follow [Bootstrap's standard markup for checkboxes and radiobuttons](https://getbootstrap.com/docs/3.3/css/#checkboxes-and-radios).  It should look something like this:
```html
<div class="checkbox">
  <label>
    <input type="checkbox" .... > 
    <span translate>SOME.KEY</span>
   </label>
</div>
```
Since there is no validation, you don't need the "form-group" styles.

client/src/modules/depots/modals/depot.modal.js
```diff
@@ -0,0 +1,42 @@
...
+
+  // exposed methods
+  vm.submit = submit;
+  vm.closeModal = closeModal;
+
+  // submit the data to the server from all two forms (update, create)
+  function submit(depotForm) {
+    var promise;
+
+    if (depotForm.$invalid || !depotForm.$dirty) { return 0; }
```
jniles: FYI, an alias for `!FormController.$dirty` is `FormController.$pristine`.  It makes the code easier to read.

client/src/modules/depots/modals/depot.modal.js
```diff
@@ -0,0 +1,42 @@
...
+      .then(function () {
+        var translateKey = (vm.isCreating) ? 'DEPOT.CREATED' : 'DEPOT.UPDATED';
+        Notify.success(translateKey);
+        $state.go('depots', null, { reload : true });
+      })
+      .catch(Notify.handleError);
+  }
+
+  function closeModal() {
+    $state.transitionTo('depots');
```
jniles: It is much better to use `$state.go()` that `$state.transitionTo()`.  The first one uses the second one internally.  To ensure we have the same behavior across all our modules, let us continue to use `$state.go()`.

server/controllers/inventory/depots/index.js
```diff
@@ -44,12 +44,15 @@ function create(req, res, next) {
   // prevent missing uuid by generating a new one
   req.body.uuid = db.bid(req.body.uuid || uuid.v4());
 
+  // enterprise for the depot
+  req.body.enterprise_id = req.session.enterprise.id;
```
jniles, line 48: :+1:

test/end-to-end/depots/depots.page.js
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * forbid deletion of used depot
+   */
+  function errorOnDeleteDepot(n) {
+    GA.clickOnMethod(n, actionLinkColumn, 'delete', gridId);
+    components.modalAction.confirm();
+    components.notification.hasError();
+  }
+
+  page.getDepotCount = getDepotCount;
```
jniles: If you made this a JS class, you could reduce all these bindings.  Classes export their methods by default.  [Here is a link to the MDN documentation for Class](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Classes).

jniles (commented):
@mbayopanda, this is a good refactor.

I've made some comments on code that looks like it is a couple months old.  It should be cleaned up before we merge this into master.  After that, LGTM.

test/end-to-end/depots/depots.page.js
mbayopanda: 👍

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -0,0 +1,48 @@
...
+
+    <div class="form-group" ng-class="{ 'has-error' : DepotForm.$submitted && DepotForm.text.$invalid }">
+      <label class="control-label" translate>DEPOT.LABEL</label>
+      <input name="text" ng-model="DepotModalCtrl.depot.text" autocomplete="off" class="form-control" required>
+      <div class="help-block" ng-messages="DepotForm.text.$error" ng-show="DepotForm.$submitted">
+        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      </div>
+    </div>
+
+    <div class="checkbox">
```
jniles, line 25: :dancer:

jniles (approved):
Looks like everything here has been addressed.

Issue #1955, opened 2017-08-09
proposal(patient registry) : Add link to patient financial statement


mbayopanda:
In the patient registry, add a link to the financial statement of a patient such as for invoices or cash payments.

It will be useful for generating the patient statement quickly instead of entering into the patient details

jeremielodi:
closed in  #2151

PR #2035, opened 2017-08-22
(refactor) Journal reference links

sfount:
This commit adds an additional check to the reference link server
controller; a redirect is not requested unless the document type is
defined and it specifies a documentPath - this stops an infinite
redirect in certain cases.

It also updates the patient document ('PA') to link to the patient
financial history document given a reference lookup.


server/controllers/medical/patients/index.js
```diff
@@ -575,7 +576,7 @@ function loadLatestInvoice(inv) {
   const combinedLedger = `
     (
       (SELECT record_uuid, debit_equiv, credit_equiv, reference_uuid, entity_uuid FROM posting_journal)
-      UNION 
+      UNION
```
jniles, line 579: I wonder if this should be UNION ALL ...

jniles (approved):
LGTM!

jniles:
Note: contributes to #1955.

PR #2036, opened 2017-08-22
fix(cashboxes): display all cashboxes in selection

### LINKED ISSUE #1976: (bug) Cash window gives no indication of only loading cash boxes from current project
The cash window currently selects only cash boxes from the users current project (chosen at sign in time). It can be very confusing to users to not see any cash boxes even though the configuration is set up with multiple cash boxes. 

I think the two options are to clearly demonstrate that these are only the cash boxes that exist in the currently signed in to project, for example 'Hospital's cash boxes:'  or to just show all cash boxes with the project shown next to the name. 

I propose that for now we just show all cash boxes that exist with the project name next to the box - if hospital's become overwhelmed with multiple projects and cash boxes we can look at changing this in the future. 

For example: 
* Cash Window (Hospital) 
* Primary Cash Box (Hospital) 
* Cash Window (Pharmacy)

jniles:
This commit allows a user to select any project's cashbox from the
cashbox selection page.  As designed, BHIMA assumes that users will have
separate projects and use their own cashboxes, so those are cleanly
separated from the remaining cashboxes.  However, it does give the user
the possibility of using any registered cashbox.

Closes #1976.


sfount (approved):
This is a huge improvement. Code LGTM 👍 
Fully addresses the referenced issue.

PR #2040, opened 2017-08-23
(feat): using a loading button instead of a simple button in purchase…

### LINKED ISSUE #2013: (bug) Purchase Order Submit button does not load.
On the purchase order creation page, after a user is completely filled out the form, they click the "submit" button.  The button appears to correctly disable, but does not provide any indication of loading.  On a slow connection, this is confusing and may result in the user navigating away.  We should update this button to listen and show the loading state.

jeremielodi:
For improving the user experience, this PR helps to use  a loading button instead of a simple button in purchase order creation form.

Closes #2013.

### DIFF
--- a/client/src/modules/purchases/create/create.js
+++ b/client/src/modules/purchases/create/create.js
@@ -18,6 +18,7 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Suppliers, Notify, Se
   vm.enterprise = Session.enterprise;
   vm.maxLength = util.maxLength;
   vm.maxDate = new Date();
+  vm.loagingState = false;
 
   // make sure we have all the suppliers we need.
   Suppliers.read()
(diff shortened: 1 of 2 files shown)


client/src/modules/purchases/create/create.js
```diff
@@ -18,6 +18,7 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Suppliers, Notify, Se
   vm.enterprise = Session.enterprise;
   vm.maxLength = util.maxLength;
   vm.maxDate = new Date();
+  vm.loagingState = false;
```
jniles, line 21: This variable should be spelled "loadingState" instead of "loagingState".

jniles (commented):
Looks pretty good!  If you update the variable name, I'll merge it in.

You should be able to do this without creating another commit.  Use `git commit --amend` and then a force push (`git push --force`) to update the branch.

It would also be useful to include the comment "Fixes #..." to specify which issue this is fixing.  That way it will be automatically closed when it gets merged.

Your commit message is very good :+1:

jniles:
Closes #2013.

PR #2043, opened 2017-08-23
(refactor) Inverse Journal default transaction sort

sfount:
This commit sorts transactions from smallest to largest by default.
Usually transactions with a smaller ID will have an earlier timestamp -
this change will implicitly (roughly) order transactions by date so that
the ordering is not swapped when moving to the grouped view.


jniles (approved):
LGTM

PR #2045, opened 2017-08-23
chore(periods): make period_ids readable (YYYYMM)

### LINKED ISSUE #1625: (proposal) Make Period IDs YYMM instead of AUTO INCREMENT
The application currently generates a `period_id` from the database using an autoincrement system.  This means that there is no association between the period date and the ID of the period.

I propose that we make `period_id`s uniformly four digits:
 1. The first two digits will be the last two digits of the current year - 17 for the current year.
 2. The last two digist will be the current month, indexed at 1 _not zero_.  This is how MySQL indexes months, but not JavaScript.

This will make the `period_id` predictable.  That way, if we have a date on the client, we can generate the period ID without even a trip to the database.  This is advantageous for numerous reasons:
 1. We can automatically compute the `period_id` for the posting journal.
 2. In the future, a filter could turn "this month" into `period_id = XXXX` for a faster lookup.
 3. `period_id`s will be human-readable in database dumps or Excel exports.

I can't see any downsides to this proposal besides marginally larger space used storing integers instead of SMALLINTs.

### LINKED ISSUE #1871: Convert period keys to human readable format
While poking around the db I noticed the keys in Period, GL, Budget, and PeriodTotals are 1,2,3 ... whereas 201701, 201702 are human readable and also no bigger than the mediumint size used for the key .  Furthermore the tests to see if a date falls in a given period  require two tests: xdate <= period.minimumdate or xdate >= period.maxdate  whereas I assume there's function that formats a date into the yyyymm form which can be directly compared with the period in question.    I would also use the sql between function to see if a date falls within a range.  Not urgent.

jniles:
This commit migrates the period_ids in the period table to be human
readable by using YYYYMM.  This was proposed twice, but should help
greatly increase the legibility of database when working with raw SQL.
As a side effect, it is also much easier to check if dates are in the
correct period, as well as select all records from a given period.

Closes #1871.  Closes #1625.

----

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/server/models/procedures.sql
+++ b/server/models/procedures.sql
@@ -1049,7 +1049,7 @@ CREATE PROCEDURE CreatePeriods(
 )
 BEGIN
   DECLARE periodId MEDIUMINT(8);
-  DECLARE periodNumber SMALLINT(5);
+  DECLARE periodNumber SMALLINT(5) DEFAULT 0;
   DECLARE periodStartDate DATE;
   DECLARE periodEndDate DATE;
   DECLARE periodLocked TINYINT(1);
@@ -1066,7 +1066,6 @@ BEGIN
   DECLARE fyUserId MEDIUMINT(5);
   DECLARE fyNote TEXT;
 
-  DECLARE v_i INT DEFAULT 0;
 
   -- get the fiscal year informations
   SELECT
@@ -1078,24 +1077,24 @@ BEGIN
   FROM fiscal_year WHERE id = fiscalYearId;
 
   -- insert N+2 period
-  WHILE v_i <= fyNumberOfMonths + 1 DO
+  WHILE periodNumber <= fyNumberOfMonths + 1 DO
 
-    IF v_i = 0 OR v_i = fyNumberOfMonths + 1 THEN
+    IF periodNumber = 0 OR periodNumber = fyNumberOfMonths + 1 THEN
       -- Extremum periods 0 and N+1
-      -- Insert periods with null dates
-      INSERT INTO period(`fiscal_year_id`, `number`, `start_date`, `end_date`, `locked`)
-      VALUES (fiscalYearId, v_i, NULL, NULL, 0);
+      -- Insert periods with null dates - period id is YYYY00
+      INSERT INTO period (`id`, `fiscal_year_id`, `number`, `start_date`, `end_date`, `locked`)
+      VALUES (CONCAT(YEAR(fyStartDate), periodNumber), fiscalYearId, periodNumber, NULL, NULL, 0);
     ELSE
       -- Normal periods
       -- Get period dates range
-      CALL GetPeriodRange(fyStartDate, v_i, periodStartDate, periodEndDate);
+      CALL GetPeriodRange(fyStartDate, periodNumber, periodStartDate, periodEndDate);
 
-      -- Inserting periods
-      INSERT INTO period(`fiscal_year_id`, `number`, `start_date`, `end_date`, `locked`)
-      VALUES (fiscalYearId, v_i, periodStartDate, periodEndDate, 0);
+      -- Inserting periods -- period id is YYYYMM
+      INSERT INTO period(`id`, `fiscal_year_id`, `number`, `start_date`, `end_date`, `locked`)
+      VALUES (DATE_FORMAT(periodStartDate, '%Y%m'), fiscalYearId, periodNumber, periodStartDate, periodEndDate, 0);
     END IF;
 
-    SET v_i = v_i + 1;
+    SET periodNumber = periodNumber + 1;
   END WHILE;
 END $$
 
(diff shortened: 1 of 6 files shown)


server/models/procedures.sql
```diff
@@ -1082,17 +1082,17 @@ BEGIN
 
     IF v_i = 0 OR v_i = fyNumberOfMonths + 1 THEN
       -- Extremum periods 0 and N+1
-      -- Insert periods with null dates
-      INSERT INTO period(`fiscal_year_id`, `number`, `start_date`, `end_date`, `locked`)
-      VALUES (fiscalYearId, v_i, NULL, NULL, 0);
+      -- Insert periods with null dates - period id is YYYY00
+      INSERT INTO period (`id`, `fiscal_year_id`, `number`, `start_date`, `end_date`, `locked`)
+      VALUES (CONCAT(YEAR(fyStartDate), v_i), fiscalYearId, v_i, NULL, NULL, 0);
```
DedrickEnc: What is ```v_i```?

test/data.sql
```diff
@@ -459,6 +459,6 @@ INSERT INTO `stock_movement` (`uuid`, `lot_uuid`, `document_uuid`, `depot_uuid`,
...
-  (HUID('6b2b9882-3ff3-11e7-b0b2-507b9dd6e004'), 1, 3, 34, 'TPB11', CURRENT_DATE(), HUID('3b2754c2-969f-4767-8a10-bb0bca06a370'), 'Facture a Test 2 Patient (PA.TPA.2) pour 1 items dans le service Administration. ', 3642, 0.0000, 25.0000, 0.0000, 25.0000, 2, '', '', NULL, 11, 1, NULL, NULL);
+INSERT INTO `general_ledger` VALUES
+  (HUID('6b2b8aa9-3ff3-11e7-b0b2-507b9dd6e004'), 1, 3, DATE_FORMAT(NOW(), '%Y%m'), 'TPB10', CURRENT_DATE(), HUID('3b2754c2-969f-4767-8a10-bb0bca06a370'), 'Facture a Test 2 Patient (PA.TPA.2) pour 1 items dans le service Administration. ', 3631, 25.0000, 0.0000, 25.0000, 0.0000, 2, HUID('3be232f9-a4b9-4af6-984c-5d3f87d5c107'), '', NULL, 11, 1, NULL, NULL),
+  (HUID('6b2b9882-3ff3-11e7-b0b2-507b9dd6e004'), 1, 3, DATE_FORMAT(NOW(), '%Y%m'), 'TPB11', CURRENT_DATE(), HUID('3b2754c2-969f-4767-8a10-bb0bca06a370'), 'Facture a Test 2 Patient (PA.TPA.2) pour 1 items dans le service Administration. ', 3642, 0.0000, 25.0000, 0.0000, 25.0000, 2, '', '', NULL, 11, 1, NULL, NULL);
```
DedrickEnc, line 464: Well done! so we will not be supposed to changed the ```period_id``` in the future

DedrickEnc (approved):
I agree with two issues which are the origin of this P.R
LGTM

DedrickEnc:
A rebase is needed and then it will be merged

sfount:
The code here LGTM and I think the idea is very interesting but I have a few thoughts/ questions on the implications of this change: 
* As there is no database rule to ensure that periods and their ids are correctly set up, what would happen if the logic expected ids to be in this format and they weren't? (During migration potentially)
* Will this require running an update script at current institutions to make sure that their periods are now correctly defined?

server/models/procedures.sql
jniles: It is just an increment variable, like the `i` in `for i in range(10):`.  It could be renamed as needed.  Do you think this should be renamed before merge? 

For clarity, it corresponds to the `period_number`.

jniles:
@DedrickEnc, I believe bors automatically merges with the latest master.  So if you type "bors r+" and it fails, then something has gone wrong between this PR and master and we can debug it.  But if it succeeds it would be the same as rebase + merge.

@sfount, this change is purely cosmetic - it will make the CSV print out be 201501 instead of 1 in the period_id column.  At this point, there is no logic hinging on this change.  So, if a database has not been migrated, the previous years should behave the same.  All new years will simply have a massively incremented FY ID (unless the procedures were also not rebuilt).

To back-port this change to running institutions, a migration script would be required.

server/models/procedures.sql
DedrickEnc: If possible yes.

jniles: Cheers, will do so.

jniles:
@DedrickEnc updates have landed.

DedrickEnc:
@jniles
Something goes wrong with the test, almost every integration test fails with the code 401.

jniles:
@DedrickEnc now it should work.  Looks like the name change collided with an unused variable in the same procedure.

PR #2050, opened 2017-08-24
(feat) removing the locked option when creating a supplier

### LINKED ISSUE #2012: (refactor) Remove the "locked" flag from the Supplier creation page
It doesn't make sense for a user to create a supplier that is automatically blocked.  This checkbox should be removed from the supplier creation page.  We do still need it on the update/edit page though.

jeremielodi:
closes #2012

### DIFF
--- a/client/src/modules/suppliers/suppliers.html
+++ b/client/src/modules/suppliers/suppliers.html
@@ -116,12 +116,7 @@
                 </div>
               </div>
 
-              <div class="checkbox">
-                <label>
-                  <input type="checkbox"  name="locked" ng-true-value="1" ng-false-value="0" ng-model="SupplierCtrl.supplier.locked">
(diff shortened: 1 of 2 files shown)


jniles:
@jeremielodi this looks perfect!

Unfortunately, this PR breaks our tests ... which us why we have tests!  The tests are found in the `test/end-to-end` folder.  They should be in a file called `suppliers.spec.js` or similar.  They check that there is an error message when a supplier is not locked - you should update these tests to remove any checks for `locked` field.

jniles:
Aha, found it - [this line](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/suppliers/suppliers.spec.js#L78) will need to be removed.

jniles:
Looks correct!

jeremielodi:
Okay

PR #2027, opened 2017-08-22
chore(ci): migrate Travis to containers

jniles:
This commit upgrades Travis tests to run on a container-based
infrastructure.  This should result in faster startup time and removes
the need for 'sudo' privileges.  It also allows cleans up the Travis
logs with folds to hide irrelevant details.

----

Thank you for contributing!

Before submitting this pull request, please verify that you have:
 - [x] Run your code through eslint.
 - [x] Run integration tests.
 - [x] Run end-to-end tests.
 - [x] Accurately described the changes your are making in this pull request.

For a more detailed checklist, [see the online review checklist](https://docs.google.com/document/d/1nupLVLRXgSZJQo_acLgrwvPnN8RukfSiwRhSToj81uU/pub) that this PR will be evaluated against.

Ensuring that the above checkboxes are completed will help speed the review process and help build a stronger application.  Thanks!

PR #2055, opened 2017-08-25
fix(stock): fix end to end tests

jniles:
This commit fixes the end to end tests in master that are blocking a few PRs: #2027 and #2033.


jniles:
Hmm... this will take more work.  It seems like this makes tests pass locally, but not on master.

PR #2057, opened 2017-08-29
refactor(Editing transaction Dates)

### LINKED ISSUE #2052: (refactor) Editing transaction dates
Currently during transaction editing, dates can be changed for every row in the transaction directly. However this does not have any effect on the period_id or fiscal_year_id that are also included in transaction information. This will definitely cause upstream logic to fail as it assumes that dates and period assignments are well formed. 

It is necessary to update the period_id and fiscal_year_id appropriately when changing the date, this should be done on the server (similarly to how debit and credit equivalents are updated) so that any contextual information can be used to block inappropriate dates from being accepted.

lomamech:
- This PR makes it possible to put the period as well as the year whenever
the date changes and prevent the modification of the date for closed or
inexistent fiscal years
- Implement of function lookupFiscalYearByDate for check the FiscalYear
  and period for one Date
closes #2052

### DIFF
--- a/server/controllers/finance/fiscal.js
+++ b/server/controllers/finance/fiscal.js
@@ -655,11 +655,7 @@ function getPeriodByFiscal(fiscalYearId) {
  *
  * @description
  * This function returns a single record from the fiscal year table matching
- * the DATE provided.  If no record is found, it throws a NotFound error.
  *
- * @returns {Promise} - a promise resolving to the fiscal record
- *
- * @private
  */
 function lookupFiscalYearByDate(transDate) {
   const sql = `
--- a/server/controllers/finance/fiscal.js
+++ b/server/controllers/finance/fiscal.js
@@ -30,6 +30,7 @@ exports.detail = detail;
 exports.update = update;
 exports.remove = remove;
 exports.getPeriodByFiscal = getPeriodByFiscal;
+exports.lookupFiscalYearByDate = lookupFiscalYearByDate;
 
 /**
  * @method lookupFiscalYear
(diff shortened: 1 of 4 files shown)


server/controllers/finance/fiscal.js
```diff
@@ -648,3 +649,25 @@ function getPeriodByFiscal(fiscalYearId) {
...
+/**
+ * @method lookupFiscalYearByDate
+ *
+ * @description
+ * This function returns a single record from the fiscal year table matching
+ * the DATE provided.  If no record is found, it throws a NotFound error.
+ *
+ * @returns {Promise} - a promise resolving to the fiscal record
+ *
+ * @private
```
jniles: This isn't a private method if you expose it :)

server/controllers/finance/fiscal.js
```diff
@@ -648,3 +649,25 @@ function getPeriodByFiscal(fiscalYearId) {
 
   return db.exec(sql, [fiscalYearId]);
 }
+
+/**
+ * @method lookupFiscalYearByDate
+ *
+ * @description
+ * This function returns a single record from the fiscal year table matching
+ * the DATE provided.  If no record is found, it throws a NotFound error.
```
jniles: I don't think this function throws a `NotFound` error.  If you used `db.one()` it would.

server/controllers/finance/journal/index.js
```diff
@@ -276,7 +280,28 @@ function editTransaction(req, res, next) {
...
+      if (transDate) {
+        if (!result.length) {
+          throw new BadRequest('Invalid Fiscal Year', 'POSTING_JOURNAL.ERRORS.NO_FISCAL_FOR_DATE');
+        }
+
+        if (result[0].locked) {
+          throw new BadRequest('Closed Fiscal Year', 'POSTING_JOURNAL.ERRORS.CLOSED_FISCAL_YEAR');
+        }
 
+        _.each(rowsChanged, (row) => {
```
jniles: This function alters data in the transaction.   @sfount attempted to avoid [spaghetti code](https://en.wikipedia.org/wiki/Spaghetti_code) by putting all logic that changes the transaction in the function `transformColumns()`.   This code belongs there.

server/controllers/finance/journal/index.js
```diff
@@ -452,6 +477,17 @@ function transformColumns(rows, newRecord, oldTransaction) {
     .then(() => rows);
 }
 
+//Check the Valid Update of Date
+function selectTransDate(rows){
```
jniles: I don't think this function does any validation.  If anything, it should just do this:
```js
const selectTransDate = (rows) => rows[0].trans_date;
```

jniles (changes requested):
@lomamech, while this accomplishes the original goal, it should really conform to the expectations of the module.

Any code that alters or _transforms_ the transaction should appear in the function `transformColumns` or be called from `transformColumns`.  That makes it much easier to refactor/maintain in the future.

jniles (approved):

Issue #2000, opened 2017-08-18
fix(stock) : entry and exit disfunctions


mbayopanda:
* Stock Entry
  ===========
  - [x] Select operation doesn't display the entity concerned. for example, when you select a transfer in the modal, there is no reference of the selected element
  - [ ] The donation operation is not completed. must be removed for now, for avoiding confusing
  - [x] Depot Transfer :
    - [x] There is no reference as document ID of transfer, it will be useful for finding quickly a transfer
    - [x] The filter doesn't work in the transfer selection modal
    - [x] Multiple entry for a transfer is allowed
    - [x] Transfers in the modal must be related to the current depot for avoiding errors

* Stock Exit
  ==========
  - [x] Distribution to a given service doesn't display the selected service
  - [x] Distribution to a given depot doesn't display the selected depot

PR #2061, opened 2017-08-31
refactor(stock transfer): update the modal ui and ux for entry

mbayopanda:
This PR : 
- fix the issue #2000 points 1 and 3
- add a document identifier in the document map table
- add a toggle button for showing received transfers

stock transfer


client/src/modules/stock/entry/modals/findTransfer.modal.html
```diff
@@ -19,6 +19,16 @@
           <span class="fa fa-filter"></span>
         </button>
       </div>
+
+      <div class="toolbar-item">
+        <button type="button"
+          ng-click="$ctrl.toggleReceived()"
+          class="btn btn-sm btn-default"
+          ng-class="{ 'btn-warning' : $ctrl.filterReceived }">
```
jniles: The other toggle we have (filters) uses `btn-info`.  Does it not make sense to also use `btn-info` there?

server/controllers/stock/index.js
```diff
@@ -107,12 +112,12 @@ function createStock(req, res, next) {
...
  */
-function createMovement(req, res, next) {  
+function createMovement(req, res, next) {
   const params = req.body;
-  
+
   const document = {
     uuid : params.document_uuid || uuid.v4(),
-    date : moment(new Date(params.date)).format('YYYY-MM-DD').toString(),
+    date : new Date(params.date),
```
jniles, line 131: :dancer:

test/integration/stock.js
```diff
@@ -166,7 +166,7 @@ describe('(/stock/) The Stock HTTP API', () => {
       .catch(helpers.handler));
 
   // list all movement relatives to 'Depot Principal'
-  it('GET /stock/lots/movements?depot_uuid=... returns movements for Depot Principal (13: 9 IN + 2 OUT)', () => agent.get(`/stock/lots/movements?depot_uuid=${depotPrincipalUuid}`)
+  it('GET /stock/lots/movements?depot_uuid=... returns movements for Depot Principal (13: 10 IN + 3 OUT)', () => agent.get(`/stock/lots/movements?depot_uuid=${depotPrincipalUuid}`)
```
jniles: This is a really long line.  Could you make it two?

server/controllers/stock/reports/index.js
```diff
@@ -560,11 +568,11 @@ function getDepotMovement(documentUuid, enterprise, isExit) {
         description        : line.description,
         date               : line.date,
         document_uuid      : line.document_uuid,
-        document_reference : line.document_reference,
+        document_reference : line.documentReference,
```
jniles: :dancer:

mbayopanda:
@DedrickEnc could you review this PR ?

DedrickEnc:
@mbayopanda
Be sure to check the point 1 and 3 of the issue #2000 when this pull request will be merged in, I can also do it.

client/src/modules/stock/entry/entry.js
```diff
@@ -202,15 +202,16 @@ function StockEntryController(Depots, Inventory, Notify,
     });
   }
 
-  function initSelectedEntity (){
-    vm.reference = vm.displayName = '';
+  function initSelectedEntity() {
+    vm.displayName = '';
+    vm.reference = vm.displayName;
```
DedrickEnc: This is the same thing, I propose to do it explicitly like this : 
```
vm.reference = '';
vm.displayName = '';
```
Good catch though 👍

client/src/modules/stock/entry/modals/findTransfer.modal.html
DedrickEnc: There is only one css class here, so what will happen if ```$ctrl.filterReceived``` takes respectively these values : ```false```, ```true``` and ```false```?

client/src/modules/stock/entry/modals/findTransfer.modal.js
```diff
@@ -9,36 +9,42 @@ StockFindTransferModalController.$inject = [
...
-  var columns = [
+  columns = [
     {
-      field: 'date',
-      cellFilter: 'date',
-      filter: { condition: filtering.filterByDate },
-      displayName: 'TABLE.COLUMNS.DATE',
-      headerCellFilter: 'translate',
-      sort: { priority: 0, direction: 'desc' },
+      field : 'status',
```
DedrickEnc: 👍

client/src/modules/stock/entry/modals/findTransfer.modal.js
```diff
@@ -48,15 +54,16 @@ function StockFindTransferModalController(Instance, StockService, Notify,
   vm.gridOptions.enableColumnMenus = false;
   vm.gridOptions.fastWatch = true;
   vm.gridOptions.flatEntityAccess = true;
-  vm.toggleFilter = toggleFilter;
+  vm.gridOptions.rowTemplate = 'modules/stock/entry/modals/templates/transfer.row.tmpl.html';
```
DedrickEnc: You can go a head and combine all gridOptions configuration once at time if possible, like this 
: 
```
vm.gridOptions = {
  enableColumnMenus : false,
  fastWatch : true,
  flatEntityAccess : true,
  rowTemplate : 'modules/stock/entry/modals/templates/transfer.row.tmpl.html'  
}
```
at line 18

client/src/modules/stock/entry/modals/findTransfer.modal.js
```diff
@@ -74,42 +81,46 @@ function StockFindTransferModalController(Instance, StockService, Notify,
     vm.gridApi.core.notifyDataChange(uiGridConstants.dataChange.COLUMN);
   }
 
+  /** toggle received */
+  function toggleReceived() {
```
DedrickEnc, line 79: 👍

client/src/modules/stock/entry/modals/findTransfer.modal.js
```diff
@@ -74,42 +81,46 @@ function StockFindTransferModalController(Instance, StockService, Notify,
...
 
-  function load (){
+  function load() {
     vm.loading = true;
 
-    StockService.movements.read(null, {
-      entity_uuid: data.depot_uuid,
-      is_exit: 1,
-      groupByDocument : 1,
+    StockService.transfers.read(null, {
```
DedrickEnc, line 93: This is much clear but now we have more code added, in the current system (**bhima-2-X**), a transfer is not an entity and there is no table to store transfer so it is just a movement with some characteristics, we have already a good service at the server side to filter entities so I suggest you to use it to filter the ```stock_movement``` table in order to get records which are transfers since a transfer is not really a entity in the system.
This will preserve us to add more code (which sometimes are unnecessary).
It is like getting conventions in debtor group table, you do not need to create a new route and new function in the server controller, you just need to filter the debtor group table.

server/controllers/stock/core.js
```diff
@@ -61,18 +61,18 @@ function getLots(sqlQuery, parameters, finalClauseParameter) {
...
-      l.expiration_date, BUID(l.inventory_uuid) AS inventory_uuid, i.delay, l.entry_date,
-      i.code, i.text, BUID(m.depot_uuid) AS depot_uuid, d.text AS depot_text, iu.text AS unit_type
-    FROM lot l
-    JOIN inventory i ON i.uuid = l.inventory_uuid
-    JOIN inventory_unit iu ON iu.id = i.unit_id
-    JOIN stock_movement m ON m.lot_uuid = l.uuid AND m.flux_id = ${flux.FROM_PURCHASE}
-    JOIN depot d ON d.uuid = m.depot_uuid
-`;
-
-  // make sure all uuids are properly converted to binary
```
DedrickEnc, line 75: Why you removed this comment? I think we should provide a good comment like : 
```
// convert uuids listed in the array (second parameter) from hexadecimal to binary
```
or something better to help future developers

server/controllers/stock/index.js
```diff
@@ -195,29 +204,25 @@ function normalMovement(document, params) {
...
+  const fluxId = isExit ? core.flux.TO_OTHER_DEPOT : core.flux.FROM_OTHER_DEPOT;
 
   parameters.lots.forEach((lot) => {
-
     record = {
-      depot_uuid,
-      entity_uuid,
-      is_exit,
-      flux_id,
+      depot_uuid    : depotUuid,
```
DedrickEnc, line 216: Using ```{depot_uuid}``` is valid like using ```{depot_uuid : depotUuid}``` so why to remove it?

server/controllers/stock/reports/index.js
```diff
@@ -98,14 +98,16 @@ function stockExitPatientReceipt(req, res, next) {
       u.display_name AS user_display_name, p.display_name AS patient_display_name,
       CONCAT_WS('.', '${identifiers.DOCUMENT.key}', m.reference) AS document_reference,
       CONCAT_WS('.', '${identifiers.PATIENT.key}', proj.abbr, p.reference) AS patient_reference, p.hospital_no,
-      l.label, l.expiration_date, d.text AS depot_name
+      l.label, l.expiration_date, d.text AS depot_name,
+      dm.text AS documentReference
```
DedrickEnc: At some point we should decide as a team how to name identifiers in SQL. Like here we have ```AS depot_name``` and ```AS documentReference``` very confusing.

server/controllers/stock/reports/index.js
```diff
@@ -547,40 +559,36 @@ function stockEntryIntegrationReceipt(req, res, next) {
  * getDepotMovement
  * @param {string} documentUuid
  * @param {object} enterprise
- * @param {boolean} isExit 
+ * @param {boolean} exit
```
DedrickEnc: I think a good name for a boolean is ```isExit```

server/models/procedures.sql
```diff
@@ -1334,6 +1334,20 @@ BEGIN
   ON DUPLICATE KEY UPDATE `quantity` = `quantity` + movementQuantity;
 END $$
 
+-- stock movement document reference
```
DedrickEnc, line 1336: You should add more comment to help other understand what is done in this stored procedure

DedrickEnc (changes requested):
@mbayopanda 
I have looked to the pull request, and I am okay with the goal of the pull request.
I suggested some changes and let me know if you are ready.

client/src/modules/stock/entry/modals/findTransfer.modal.html
mbayopanda: The default css class is `btn-default` on the line `26`.
- false : `btn-default` css class will be applied
- true : `btn-default btn-warning` css classes applied, the `btn-warning` will override the `btn-default` class
- false : `btn-default` css class will be applied

client/src/modules/stock/entry/modals/findTransfer.modal.js
mbayopanda: On the line `32` the filter `filtering.filterByDate` is used in the column definition of the grid, so it is not possible to combine all gridOptions configuration once at time with an object definition.

```javascript
vm.gridOptions = {};

var filtering = new Filtering(vm.gridOptions);

var columnDefs = [{
  field : 'date',
  cellFilter : 'date',
  filter : { condition : filtering.filterByDate },
  ...
}]

vm.gridOptions.columnDefs = columnDefs;
```
the `filtering` object need the `vm.gridOptions`,

client/src/modules/stock/entry/modals/findTransfer.modal.js
mbayopanda: We don't have an API route for having the transfer movements to a depot and also the number of these movements which are received by the depot, there is a particular query which has been made for having the transfer movements and the number of received.

For this particular query => we needed an API route and for this API route => we need client side service (a method).

server/controllers/stock/index.js
mbayopanda: By rules define by eslint we must use `camelCase` practice for all variable name. And `depotUuid` is a variable name.

server/controllers/stock/reports/index.js
mbayopanda: You're right, it is confusing here because we already has `document_reference` on the line 99.

server/controllers/stock/reports/index.js
mbayopanda: You're right, but in the `getDepotMovement` function there is another `isExit` variable which is used. So we must have a distinction between `isExit` as parameter and 'isExit' as variable in the function.

server/models/procedures.sql
mbayopanda: 👍

client/src/modules/stock/entry/modals/templates/transfer.row.tmpl.html
```diff
@@ -0,0 +1,9 @@
+<div 
+  ng-hide="row.entity.countedReceived > 0 && !grid.appScope.filterReceived"
```
jniles, line 2: Why do the filter on a row template?  Why not just filter the data in the controller?

This seems to me overly complicated.

jniles (approved):
My changes were addressed :+1:

PR #2062, opened 2017-08-31
Financial writting for stock entry by purchase order

DedrickEnc:
This P.R enhances the stock entry from purchase order process by writing finance records into the posting journal.
Note that the ```db.convert``` method has been augmented to support items array converting. Now you can pass an array uuid to the ```db.convert``` , no change to the interface but if a key is an array, every item will be converted into binary uuid.


server/lib/db/index.js
```diff
@@ -200,6 +200,13 @@ class DatabaseConnector {
       if (data[key] && typeof data[key] === 'string') {
         data[key] = this.bid(data[key]);
       }
+
+      if(data[key] && Array.isArray(data[key])) { 
+        var that = this;
```
sfount: This looks like a useful case to handle! 

I have only briefly scanned the code here but I think that this can be written as: 
```js 
// map the bid convert method across all elements in the value array
data[key] = data[key].map(this.bid);
```

The way the closure should be respected and it reads exactly as it performs - map this function to these values. 
It would also be useful to have a comment here showing what case this is handling (like the previous `string` case has).

DedrickEnc:
@mbayopanda @jniles 

Can I get a review ?

server/lib/db/index.js
```diff
@@ -200,6 +200,14 @@ class DatabaseConnector {
       if (data[key] && typeof data[key] === 'string') {
         data[key] = this.bid(data[key]);
       }
+
+      // the key exists on the object and value is an array
+      if(data[key] && Array.isArray(data[key])) { 
+        var that = this;
```
jniles: It doesn't look you ever use `that`.  Could you remove it?

server/controllers/stock/index.js
```diff
@@ -103,6 +82,60 @@ function createStock(req, res, next) {
     .done();
 }
 
+function processMovements (document, lots) {
+  
+  const items = lots || [];
+  const movements = [];
+
+  items.forEach((item) => {
```
jniles: If you are doing a pattern like this:
```js
var x = [];

row.forEach((r) => {
  // do something
  x.push(r);
});
```
what you are really doing is a `map()`.

Your code could also be expressed  like this:
```js
function processMovements(document, lots = []) {
  const movements = lots.map((item) => {
      uuid       : db.bid(uuid.v4()),
      lot_uuid   : item.uuid,
      depot_uuid : db.bid(document.depot_uuid),
      document_uuid : db.bid(document.uuid),
      flux_id    : document.flux_id,
      date       : document.date,
      quantity   : item.quantity,
      unit_cost  : item.unit_cost,
      is_exit    : 0,
      user_id    : document.user ,
      entity_uuid : null,
      description : null
  });

  // do your filter down here to organize your code
}
```

server/lib/db/index.js
```diff
@@ -200,6 +200,14 @@ class DatabaseConnector {
       if (data[key] && typeof data[key] === 'string') {
         data[key] = this.bid(data[key]);
       }
+
+      // the key exists on the object and value is an array
+      if(data[key] && Array.isArray(data[key])) {
```
jniles, line 205: This is a cool improvement. :+1:

At some point, we should start writing unit tests for these server-side functions.

server/models/procedures.sql
```diff
@@ -1334,4 +1334,152 @@ BEGIN
...
+BEGIN
+  INSERT INTO 
+    `lot` (`uuid`, `label`, `initial_quantity`, `unit_cost`, `expiration_date`, `inventory_uuid`, `origin_uuid`, `delay`)
+  VALUES
+    (uuid, label, initial_quantity, unit_cost, expiration_date, inventory_uuid, origin_uuid, delay);
+END $$
+
+
+-- This stored procedure, insert a record into the table stock movement
+CREATE PROCEDURE CreateStockMovement (
```
jniles: Why do these need to be procedures?  It looks like all they do is `INSERT` to a regular table...

Both `CreateCash` and `CreateInvoice` only exist because they use `TEMPORARY` tables that need to be shared between stored procedures.  Since these procedures don't use TEMPORARY TABLEs, I don't think they need to exist .. they will just be one more thing to update when we try to do a database migration.

server/models/procedures.sql
```diff
@@ -1334,4 +1334,152 @@ BEGIN
...
+  INSERT INTO posting_journal 
+  (
+    uuid, project_id, fiscal_year_id, period_id, trans_id, trans_date,
+    record_uuid, description, account_id, debit, credit, debit_equiv,
+    credit_equiv, currency_id, origin_id, user_id
+  ) 
+  SELECT
+    HUID(UUID()), project_id, current_fiscal_year_id, current_period_id, transaction_id, 
+    date, p.uuid, sm.description, ig.stock_account, pi.total, 0, pi.total, 0, currency_id,
+    9, user_id
```
jniles: Where did this 9 come from?

jniles (commented):
@DedrickEnc this is a good start!

I've made a few comments to clean up the code a bit.  It doesn't look like a stored procedures is needed for some of the preliminary INSERTs.

The extension to the `db.convert()` method is cool :+1:

server/models/procedures.sql
DedrickEnc: In the transaction type table.

jniles: It would be good to put a variable that describes why this is 9 and not 8 or 10.  For example.
```sql
SOME_RANDOM_ID = 9;

INSERT INTO some_table.some_column SET id = SOME_RANDOM_ID
```

Otherwise, it makes it hard to know why this is 9 and not any other number.

DedrickEnc: I agree, but I am not understanding the code example that you written can you be more clear?
For example : if the code must be a variable, how can you edit the followed statement : 
``` INSERT INTO user (id, name, code) VALUES (1, 'some name', 5)```

Comment (jniles) [reply server/models/procedures.sql]: Like this:
```sql
DECLARE USER_MANAGER_CODE INT;

SET USER_MANAGER_CODE = 5;

INSERT INTO user (id, name, code) SELECT 1, 'some name', USER_MANAGER_CODE;
INSERT INTO user (id, name, code) SELECT 2, 'some name', USER_MANAGER_CODE;
```

In this case, instead of a magic number "5", any other programmers editing the code will know what the number "5" signifies.  And, if a test fails (because someone inserted a new code, making the code for the manager 6), a programmer can easily diagnose and correct the error because they will realize that "5" is no longer what the original programmer intended it to be.

jniles: Here is an example when the [cost of invoice items](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/procedures.sql#L1250) is computed and the [here is where it is inserted into a table](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/procedures.sql#L129).

DedrickEnc:
@jniles
Changes are landed

server/controllers/stock/index.js
```diff
@@ -103,6 +119,28 @@ function createStock(req, res, next) {
     .done();
 }
 
+// this function process lot list in order to insert them to the database
```
jniles: It doesn't look like you use this code - can it be removed?

server/models/procedures.sql
```diff
@@ -1334,4 +1334,116 @@ BEGIN
...
+  DECLARE InvalidInventoryAccounts CONDITION FOR SQLSTATE '45006';
+  DECLARE current_fiscal_year_id MEDIUMINT(8) UNSIGNED;
+  DECLARE current_period_id MEDIUMINT(8) UNSIGNED;
+  DECLARE current_exchange_rate DECIMAL(19, 4) UNSIGNED;
+  DECLARE transaction_id VARCHAR(100);
+  DECLARE verify_invalid_accounts SMALLINT(5);
+  DECLARE TRANSACTION_TYPE TINYINT(3) UNSIGNED;
+
+  -- should not be reassigned during the execution, to know why 9 please see the transaction_type table
+  SET TRANSACTION_TYPE = 9;
```
jniles: A much better name for this would be "PURCHASE_TRANSACTION_TYPE".

jniles (approved):
Looks good to me!  This has a nice separation between SQL logic and JS logic.

I've made some comments if you want to remove the old code.  Otherwise, you can do "bors r+".

jniles:
bors delegate+

Issue #1936, opened 2017-08-04
refactor: Make exchange rate display human-readable units


jniles:
I believe we had a good idea that was poorly execute with the exchange rate display on the home page.  Since we force the ratio to be `enterprise_currency : foreign_currency`, it makes for very bizarre display when the enterprise is run in a currency that is roughly 1000 : 1, as is the FC / USD ratio.

This looks nonsensical and broken to a hospital administrator:
![exchangenonsense](https://user-images.githubusercontent.com/896472/28962330-48350708-78fd-11e7-8d1c-c93ceecc5a57.png)
_Fig 1: Nonsense display when the Enterprise is run in FC_

![exchangesomesense](https://user-images.githubusercontent.com/896472/28962331-4849f41a-78fd-11e7-97fa-aea0655ed973.png)
_Fig 2: This makes sense; We should show this._

Comparing the two, the second is meaningful and the former looks broken.  Even though it is _technically_ useful to show the ratio between the two, it would be better if we showed the largest number possible.  So, we could compute `enterprise / foreign` and `foreign / enteprise` and simply show the largest one to the user.

jeremielodi:
closed  #1936

jniles:
Closed in #2063.

PR #2063, opened 2017-08-31
feat(Home page) Make exchange rate display human-readable

jeremielodi:
This PR solves the problem we had at #1936  abaut making the exchange rate display human-readable

The result looks like this
screem_short

Instead of
screem2


jniles (approved):
Looks good to me!

PR #2064, opened 2017-08-31
fix(registry): Fix the search filter in the payment registry

### LINKED ISSUE #2059: bug: Payments Registry "Exclude Reversed Payments" doesn't work
On the Cash Payments Registry, the "reversed" search parameter can either be:
 1. Include uniquely reversed records.
 2. Exclude all reversed records.

While (1) works perfectly, the (2) option doesn't seem to register at all.  This should be fixed as soon as possible since it is helpful to remove records that have been reversed from the analysis.  This bug does not  affect the Patient Invoice Registry.

DedrickEnc:
This P.R fixes the bug on the registry of payment when filtering by reversed equal zero
closes #2059

### DIFF
--- a/client/src/js/services/FilterService.js
+++ b/client/src/js/services/FilterService.js
@@ -150,8 +150,8 @@ function FilterService(Store) {
 
     Object.keys(this._filterIndex).forEach(function (key) {
       var filter = this._filterIndex[key];
-
-      if (filter._value) {
+      
+      if (filter._value !== null && filter._value !== undefined && filter._value.length !== '') {
         filtered.push(angular.copy(filter));
       }
     }.bind(this));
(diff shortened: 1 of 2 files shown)


client/src/js/services/FilterService.js
```diff
@@ -150,8 +150,8 @@ function FilterService(Store) {
 
     Object.keys(this._filterIndex).forEach(function (key) {
       var filter = this._filterIndex[key];
-
-      if (filter._value) {
+      
+      if (filter._value !== null && filter._value !== undefined && filter._value.length !== '') {
```
jniles, line 154: I think you can just do `if (angular.isDefined(filter._value)) {`.

DedrickEnc: ```angular.isDefined``` will condider ```null``` and empty string as value and will load every filter. This what I have got : 
filtres

jniles: Fascinating.  I wonder how a user is able to enter `null`...

jniles:
LGTM 

bors  r+

Issue #1951, opened 2017-08-09
fix(posting journal) : mark transactions as comment


mbayopanda:
A comment tool like in the account statement module is necessary for marking transactions

jniles:
@sfount, what do you think - should we include this in `feature-journal-tools`?

sfount:
@jniles I think this feature can be included in `feature-journal-tools`. It should introduce very little new code but adds a lot of functionality to working with the Journal.

jniles:
I'll take it then.

jniles:
Whoops - the mobile UI doesn't show assignments.  It looks like @jeremielodi is already assigned.  It's all yours!

jeremielodi:
closes #2065

jeremielodi:
closed in #2065

PR #2065, opened 2017-08-31
feat(posting journal) marking transactions as comment

jeremielodi:
This PR solves this problem #1951

### DIFF
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -57,6 +57,8 @@ function JournalController(Journal, Sorting, Grouping,
   var cache = AppCache(cacheKey.concat('-module'));
   var vm = this;
 
+  vm.selectedRows = [];
+
   // number of all of the transactions in the system
   Journal.count()
     .then(function (data) {
@@ -77,8 +79,31 @@ function JournalController(Journal, Sorting, Grouping,
     flatEntityAccess           : true,
     enableGroupHeaderSelection : true,
     enableRowHeaderSelection   : true,
-    onRegisterApi : function (api) { vm.gridApi = api; }
+    onRegisterApi :  function onRegisterApi(api) {
+      vm.gridApi = api;
+    },
   };
+
+  // comment selected rows
+  vm.commentRows = function commentRows() {
+
+    vm.selectedRows = vm.gridApi.selection.getSelectedGridRows();
+
+    Journal.openCommentModal({ rows : vm.selectedRows })
+    .then(function (comment) {
+      if (!comment) { return; }
+      updateGridComment(vm.selectedRows, comment);
+      Notify.success('ACCOUNT_STATEMENT.SUCCESSFULLY_COMMENTED');
+    })
+    .catch(Notify.handleError);
+  };
+
+   // update local rows
+  function updateGridComment(rows, comment) {
+    rows.forEach(function (row) {
+      row.entity.comment = comment;
+    });
+  }
 
   vm.grouped = angular.isDefined(cache.grouped) ? cache.grouped : false;
 
@@ -265,6 +290,9 @@ function JournalController(Journal, Sorting, Grouping,
       displayName      : 'TABLE.COLUMNS.RESPONSIBLE',
       headerCellFilter : 'translate',
       visible          : false },
+    { field            : 'comment',
+      displayName      : 'TABLE.COLUMNS.COMMENT',
+      headerCellFilter : 'translate' },
   ];
   vm.gridOptions.columnDefs = columns;
 
--- a/client/src/modules/journal/journal.service.js
+++ b/client/src/modules/journal/journal.service.js
@@ -2,7 +2,7 @@ angular.module('bhima.services')
   .service('JournalService', JournalService);
 
 // Dependencies injection
-JournalService.$inject = ['PrototypeApiService', 'AppCache', 'FilterService', 'PeriodService', '$uibModal'];
+JournalService.$inject = [ 'PrototypeApiService', 'AppCache', 'FilterService', 'PeriodService', '$uibModal'];
 
 /**
  * Journal Service
@@ -15,8 +15,10 @@ function JournalService(Api, AppCache, Filters, Periods, Modal) {
   service.grid = grid;
   service.saveChanges = saveChanges;
   service.openSearchModal = openSearchModal;
+  service.openCommentModal = openCommentModal;
   service.openTransactionEditModal = openTransactionEditModal;
   service.mapTransactionIdsToRecordUuids = mapTransactionIdsToRecordUuids;
+  service.commentPostingJournal = commentPostingJournal;
 
   /**
    * Standard API read method, as this will be used to drive the journal grids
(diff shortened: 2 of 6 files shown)


client/src/i18n/en/table.json
```diff
@@ -23,6 +23,7 @@
             "CHARGE": "Charge",
             "CLIENT": "Client",
             "CODE": "Code",
+            "COMMENT" :"Commentaire",
```
jniles: This key should already be translated somewhere, right?  Can we just reuse the old key?

client/src/modules/journal/journal.service.js
```diff
@@ -180,5 +202,12 @@ function JournalService(Api, AppCache, Filters, Periods, Modal) {
     }).result;
   }
 
+  // updating the posting journal by adding comments in transactions
+  function commentPostingJournal(params) {
+    return service.$http.put(URL.concat('comments'), { 'params' : params })
+      .then(util.unwrapHttpResponse);
```
jniles: The looks correct :+1:

You actually don't need to inject `util` here.  It is available on the `service` instance.  Like this:

```js
```

client/src/modules/journal/modals/comment.modal.html
```diff
@@ -0,0 +1,33 @@
+<form name="ModalForm"
```
jniles: It would be really cool if this would use the exact same modal HTML as the account statement.  Do you think that is possible?

client/src/modules/journal/journal.service.js
jniles: Then you can remove the extra `$inject` for util.

server/controllers/finance/journal/index.js
```diff
@@ -514,3 +514,26 @@ function count(req, res, next) {
...
+
+
+/**
+ * PUT /journal/comments
+ * @param {object} params - { uuids: [...], comment: '' }
+ */
+function commentPostingJournal(req, res, next) {
+  const params = req.body.params;
+  const uuids = params.uuids.map((uuid) => {
+    return db.bid(uuid);
```
jniles: Just FYI: this can also be  
```
const uuids = params.uuids.map(db.bid);

// or

const uuids = params.uuids.map(uuid => db.bid(uuids));
```

server/controllers/finance/journal/index.js
```diff
@@ -514,3 +514,26 @@ function count(req, res, next) {
...
+  });
+
+  const sql = 'UPDATE posting_journal SET comment = ? WHERE uuid IN ?';
+  db.exec(sql, [params.comment, [uuids]])
+    .then((rows) => {
+      if (!rows.affectedRows || rows.affectedRows !== uuids.length) {
+        throw new BadRequest('Error on update general ledger comment');
```
jniles: This should say "journal" not "general ledger".

It also needs to have a second parameter which is a key to be translated on the client

client/src/modules/journal/modals/comment.modal.html
jeremielodi: I think it can be a litle bit complicated because there is a submit function on the button click, this function has the API url witch is not the same for both (Account statement and posting journal)

client/src/modules/journal/journal.js
```diff
@@ -77,9 +79,37 @@ function JournalController(Journal, Sorting, Grouping,
     flatEntityAccess           : true,
     enableGroupHeaderSelection : true,
     enableRowHeaderSelection   : true,
-    onRegisterApi : function (api) { vm.gridApi = api; }
+    onRegisterApi :  function onRegisterApi(api) {
+      vm.gridApi = api;
+      vm.gridApi.selection.on.rowSelectionChanged(null, rowSelectionChanged);
+      vm.gridApi.selection.on.rowSelectionChangedBatch(null, rowSelectionChanged);
```
jniles: This is unnecessary - the data is already captured in the `SelectionService`.

client/src/modules/journal/journal.js
```diff
@@ -77,9 +79,37 @@ function JournalController(Journal, Sorting, Grouping,
...
   };
 
+  // row selection changed
+  function rowSelectionChanged() {
+    vm.selectedRows = vm.gridApi.selection.getSelectedGridRows();
+  }
+
+  // comment selected rows
+  vm.commentRows = function commentRows() {
+    Journal.openCommentModal({ rows : vm.selectedRows })
```
jniles: Instead of using `vm.selectedRows`, try using the data in the `SelectionService`.  It should already be captured in the `JournalCtrl.selection.selected.groups` variable.  That way, you don't create expensive listeners.

client/src/modules/journal/journal.js
```diff
@@ -77,9 +79,37 @@ function JournalController(Journal, Sorting, Grouping,
...
+    .then(function (comment) {
+      if (!comment) { return; }
+      updateGridComment(vm.selectedRows, comment);
+      Notify.success('ACCOUNT_STATEMENT.SUCCESSFULLY_COMMENTED');
+    })
+    .catch(Notify.handleError);
+  };
+
+   // update local rows
+  function updateGridComment(rows, comment) {
```
jniles: :+1:

client/src/modules/journal/modals/comment.modal.html
jniles: The JS can be different - just bind a different controller.  That way, the `submit()` method will be different, but the HTML will be the same.  But the HTML can be reused from [comment.modal.html](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/account_statement/modals/comment.modal.html).

This way we don't have to maintain two HTML files and fix bugs in two different places :+1:  It also respects the principle of [Don't Repeat Yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

client/src/modules/journal/modals/comment.modal.html
```diff
@@ -0,0 +1,33 @@
...
+  ng-submit="$ctrl.submit(ModalForm)"
+  novalidate>
+
+  <div class="modal-header" translate>
+    ACCOUNT_STATEMENT.COMMENT
+  </div>
+
+  <div class="modal-body">
+    <div class="form-group">
+      {{ $ctrl.rows.length || 0 }}
```
jniles: Shouldn't this always be zero or greater?  Why is the `|| 0` necessary?

jniles (commented):
@jeremielodi this is almost finished.

I still think we can use the same HTML file.  It shouldn't be too difficult - we've done it in a couple other places with edit/update files.  You'll just need to bind a different  controller in the `Modal.open()` call.

Otherwise, looks pretty good!  I've made some comments to improve the performance of the Journal page.  The fewer listeners the better.

jniles:
@jeremielodi, what is the status of this PR?

jeremielodi:
Every problems was solved,

jniles:
It would have been good to leave a comment to that affect.

It looks like there are still some listeners on the grid ... It was recommended to remove them in [this comment](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2065#discussion_r136570774).

These listeners are fired on every selection change.  We actually only need to know if there are items selected at any given time ... which you can get from `JournalCtrl.selection.selected.groups`.  If this is empty, nothing is selected.

You can then get the rows in the `commentRows` function by `vm.gridApi.selection.getSelectedRows()` once.  By using this strategy, you will reduce the number of listeners on the grid and increase the overall performance of the module.

jeremielodi:
I have reduced the number of listener by use you advice Jniles. Normaly every thing should be good now

jniles (approved):
Looks much better!

PR #2066, opened 2017-08-31
fix(patients): fix invalid path to file

jniles:
This commit fixes an invalid path to the file.virus upload.  It should
fix our continuous integration tests.

... Let's try and get Travis fixed, shall we?

-----

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.

### DIFF
--- a/test/end-to-end/patient/record.spec.js
+++ b/test/end-to-end/patient/record.spec.js
@@ -1,56 +1,57 @@
-/* global element, by, browser */
-'use strict';
+/* global element, by */
 
-const path    = require('path');
-const chai    = require('chai');
-const expect  = chai.expect;
+const path = require('path');
+const chai = require('chai');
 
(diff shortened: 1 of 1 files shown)


jniles:
Note that I am running the e2e tests locally, but they take some time and I may have to board my flight before they finish :small_airplane: 

If they do not pass, I'll debug them from ATL. :baggage_claim:

jniles:
Looks like tests pass locally.  Now to see if :robot: is happy about it.

jniles:
Looks like this is going to solve it.  The issue is likely that webdriver updated such that an invalid file is no longer treated the same.  Since the test was for an invalid file, and invalid _path_ was never checked and the author never updated the path when they added this test. :(

I've updated the path so that the test actually tests what it says it will.

PR #2084, opened 2017-09-07
Refactor break up stock

### LINKED ISSUE #2068: refactor: break up stock/reports/index.js
The [stock/reports/index.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/stock/reports/index.js) file is about 890 lines long.  I believe it handles rendering multiple PDF reports - we should break it up into separate files that can easily be updated.  Any shared functions can be moved into a shared javascript file that is imported by each smaller report file.

ghost:
This is a series of simple changes to refactor server/controllers/stock/reports/index.js, which was requested in issue #2068.

Closes #2068.


server/controllers/stock/reports/stock/inventory_report.js
```diff
@@ -0,0 +1,59 @@
...
+      return report.render(data);
+    })
+    .then((result) => {
+      res.set(result.headers).send(result.report);
+    })
+    .catch(next)
+    .done();
+}
+
+exports.stockInventoyReport = stockInventoriesReport;
```
jniles: This is where you are running into errors - this should be `stockInventoryReport`, not `stockInventoriesReport`.

We are a multilingual team and sometimes have a few confusing variable names. :)

server/controllers/stock/reports/stock/inventory_report.js
```diff
@@ -0,0 +1,59 @@
...
+ * GET /reports/stock/inventory
+ */
+function stockInventoryReport(req, res, next) {
+  const data = {};
+  let options;
+  let report;
+
+  const optionReport = _.extend(req.query, {
+    filename : 'TREE.STOCK_INVENTORY_REPORT',
+    orientation : 'landscape',
```
jniles: It looks like almost every report uses the following object:
```js
{
  orientation : 'landscape',
  footerRight : '[page] / [toPage]',
  footerFontSize : '8',
}
```

Could you put this object in `common.js`?  Then it will be available to use with `_.extend()`, [which can take multiple objects](https://lodash.com/docs/4.17.4#assignIn).  It would look something like this:
```js
const optionReports = _.extend(req.query, pdfOptions, {
  filename : 'TREE.STOCK_INVENTORY_REPORT',
});
```

jniles (commented):
This is looking really good!  Great work.  :+1:

I think I've pinpointed the source of the trouble and made a subsequent recommendation for refactoring.

Let me know if you need any help!

server/controllers/stock/reports/stock/inventory_report.js
ghost: I see this typo now :) That is fixed in commit 16a6eecd3e8e29ecee92d0355e7594765f07904a

server/controllers/stock/reports/stock/inventory_report.js
ghost: You are right -  I can see it in a lot of reports. So I have refactored it out like you say and put it in common.js :)
974cc6bfd44e4e4f85a98653ce301ca6bd187a80

server/controllers/stock/reports/index.js
```diff
@@ -7,868 +7,18 @@
...
-      res.set(result.headers).send(result.report);
-    })
-    .catch(next)
-    .done();
-}
+const stockLotsReport = require('./stock/lots_report');
+const stockMovementsReport = require('./stock/movements_report');
+const stockInventoriesReport = require('./stock/inventories_report');
+const stockInventoryReport = require('./stock/inventory_report');
+const stockExitPatientReceipt = require('./stock/exit_patient_receipt');
```
jniles, line 14: It looks like you have a typo in the file name which makes this `require()` fail.  The file is called `exit_patent_receipt` instead of `exit_patient_receipt`.

jniles (commented):
Changes are looking very good!

To debug the tests, you can take a look at the Travis log.  For example, I deduced the filename error from [this line failing](https://travis-ci.org/IMA-WorldHealth/bhima-2.X/jobs/273240971#L861) in our test suite.  Unfortunately, the tests take a while to run because we lump everything together ... we'll be trying to optimize our opts in the future, but it isn't a priority currently.

Thanks for all your help!

server/controllers/stock/reports/index.js
ghost: Ah great thanks for the heads up. I see the travis page now - I will commit and check the output make sure there isn't another typo.

ghost: Ok I can see a different error now regarding Route.get:
https://travis-ci.org/IMA-WorldHealth/bhima-2.X/jobs/273610354

jniles: Sweet!   This means that there is an `app.get()` somewhere that isn't being given a function.  We'll have to track that down...

ghost: All the app.get( lines for stockReports seems to be okay in terms of referencing existing functions. I can see it is failing on this line in travis: server/config/routes.js:583, which is the first line like this:
```
app.get('/reports/stock/lots', stockReports.stockLotsReport);
```

It is the same exports as before so shouldn't be an issue with that:
https://github.com/jambroo/bhima-2.X/blob/89dbd443b3a5bbdfde386ab5f5cf2f97343e924e/server/controllers/stock/reports/index.js#L24

Does anything look wrong to you there?

jniles: Nothing looks wrong online - let me pull your branch and play around with it.  I'll report back when I'm able to figure something out!

jniles: Aha - gotcha.  Each of the `require()`s are nested one too deep.  In other words, you export the functions using `exports.fn = fn`, and then pass them up to the top via another `exports.fn = fn`.  Specifically, each of [these lines](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2084/files#diff-a51ce60ba2265ea47425f240a42b96cdL874) should either read:
```js
 exports.stockLotsReport = stockLotsReport.stockLotsReport;
// etc ...
```
OR

You can change [the exports line to using module.exports](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2084/files#diff-37cc91ea7119fcc956060bf5ae71be36R70).

I've gone ahead and fixed this while I was messing around with you code base.  I chose the `module.exports` pattern.  I think it should all pass - let's see if Travis is happy with it!

jniles:
Yay!  Got past the integration test stage!

But now our [end to end tests are not happy](https://travis-ci.org/IMA-WorldHealth/bhima-2.X/jobs/273645508#L1977).  Which means that our integration test coverage is probably not great.  I'll try tracking this down locally again and let you know what the issue is.

ghost:
Really good catch with the module.exports fix :) I can look at the end to end tests results tomorrow and see what is wrong.

ghost:
Not sure if this will help but I can see this line in the end-to-end test output, which references growl-notification.
https://travis-ci.org/IMA-WorldHealth/bhima-2.X/jobs/273645508#L1978

Might be worth mentioning too it references the following functions in StockCtrl.search(), StockLotsCtrl.search(), SubsidyCtrl.create() and SupplierCtrl.create().

jniles:
@jambroo yeah, I got a bit sidelined today from coding 🤷‍♂️ 

The failing test is a bit misleading - the bug occurs somewhere beforehand in the stock exit tests.   However, those tests don't check for our failure notification (hence the growl message), and when the search tests try to click a button to search it is inaccessible.

So, it will be a matter of back-tracking to find the offending test ☹️

server/controllers/stock/reports/stock/exit_service_receipt.js
```diff
@@ -0,0 +1,69 @@
+const { _, ReportManager, Stock, identifiers, NotFound, STOCK_EXIT_SERVICE_TEMPLATE } = require('../common');
```
jniles: Aha!  The bug is here - `db` is not defined, and should be imported.

This kind of error can be caught fairly easily if you have [eslint](https://eslint.org/) installed in your editor.  It has a ton of [integrations](https://eslint.org/docs/user-guide/integrations), so your preferred editor is probably supported.

Guess that's why we have tests ;)

ghost:
Great catch again :) I stopped being lazy and opened up the project in WebStorm and it ran an eslint check. I fixed the structure of the includes as well because it was complaining there as well.
I will keep an eye on Travis now and see how it goes with the build.

ghost:
Looks good now :) https://travis-ci.org/IMA-WorldHealth/bhima-2.X/builds/274339215

jniles (approved):
Awesome!  Code LGTM.

jniles:
@jambroo, great work!  Your contributions are much appreciated.  I know some of the internals are not pretty, but these modules will be much more maintainable after your changes :+1:

ghost:
Thanks for your kind words @jniles. And thanks for being so patient with me on this :)

jeremielodi:
Hi! I was running the end to end test and I got an error while closing the fiscal year. I try to debug manualy then a got this result in the log
errro fiscal
 can any help me to solve this please.

Issue #1432, opened 2017-03-28
(design) Journal Module


sfount:
This proposal follows on from https://github.com/IMA-WorldHealth/bhima-2.X/issues/1241 with specific details on how this could be accomplished. 

Objectives of the Journal: 
* View all financial transactions in the system 
* 'Post' records that are verified/ complete (view only unposted records to keep on top of this process)
* Correct mistakes on transactions that have not been posted/ finalised
* Provide a view into any of the billing entities within BHIMA 
  * Patients 
  * Invoices
  * Cash Payments 
  * Vouchers 
  * Accounts 
  * Periods
* **(optional)** Add a comment/ note to any line in the journal, filter on these notes to have custom views into the data
* Ability to quickly export any group of transactions/ rows to CSV to be manipulated in excel

Technical objectives of the Journal: 
* Remember your exact configuration settings
  The journal is a complicated module and it is frustrating to constantly need to resize, move and show/ hide columns according to your required configuration 
* Responsive load times
* Tested filter/ grouping functionality
* Default and custom filters
  * Transaction date default filter 
  * Posted/ not posted default filter
  * Other custom filters
* Allow optimised custom filters 
  * Only show transactions relating to invoices 
  * Only show transactions relating to cash payments
  * Only show transactions relating to vouchers

Work focus: 
* Performance - ensuring that features/ refactors applied to the journal can be run on low end hardware 
* Functionality - common questions asked by an accountant should be able to be researched using the journal module

Required steps to implement the Journal: 
* Remove additional modules that require maintenance 
  * Remove posted journal - this module will require exactly the same 
  * Change structure of current General Ledger (all research for transactions should be through the journal)
* Formally propose migration of General Ledger to report structure 
  * Propose design for reports interactively linking to each other (Through HTML preview?)
* **[Requires Issue]** Research saving ui-grid state
* **[Requires Issue]** Refactor posting journal export to CSV 
* **[Requires Issue]** Validate posting journal edits on save
* **[Requires Issue]** Split journal filters into default and custom filters

jniles:
1. Research ui-grid save state - #1050.
 2. Validating Posting Journal Edits - #1501.
 3. Refactoring Posting Journal Export to CSV - #1506.

jniles:
Fascinatingly, I think all of these are actually completed in the system.  Let me know if anything is missing and I'll re-open.

Issue #1824, opened 2017-06-27
(feature) Show invoice details on Cash Receipt


sfount:
This bug was submitted to the BHIMA `support@bhi.ma` thread. 

**Proposal** 
> Cash payment of patient 2 000FC against TF bill for c-section of 200 000FC
I can see the bill was from maternity department and the Facture number.
"Paiement par Sedi Lubutu Marie (PA.HEV.4555) contre les facture(s) IV.HEV.20724. -- Maternite"
Could this line also include the following text from the Facture/Invoice Description
In this example the description of the item is: Cesariene TF3.03 (Tarif Forfetaire)
This would help make it clear what the patient is paying towards on the printed quittance.  

* Add the invoice description to the cash payments receipts - this should consider both the POS simplified receipts as well as the more complex PDF receipts

jniles:
@lomamech what is the progress on this?

lomamech:
Sorry, before I had not understood the problematic of this issue (Translation problem). Therefore the only problem is only to add the descriction to the cash payments receipts

jniles:
Okay, just curious :)

lomamech:
I'm working on

sfount:
As I understand it they are asking for the description of the **invoice item** on the invoice here. If we displayed the description for every invoice item on every cash payment that has the potential to become a very large invoice... 

In his example here I believe they will always have only one invoice item, however that is not something we can guarantee. 

@jniles 
Do you think this is worth pursuing further with an optional flag to show all invoice items in the cash payment or would that be an additional level of complexity we do not need to support?

jniles:
We could probably support this with an additional flag - I can't see it as detrimental.  However, I would make is a rendering flag ... so that the receipt fetched all the details of the invoice, rather than storing any extra information in the database.  That way we don't have to worry about updating data if the user toggles the flag - just behavior.

jniles:
Relabeling as "Feature Request" since that is more appropriate here.

PR #2085, opened 2017-09-09
refactor: Posting Journal and Associated Tools

### LINKED ISSUE #1034: (Proposal) Link receipts/ documents from posting journal rows
Currently recipients (Debtors/ Creditors) and References (Invoices/ Cash Payments etc.) are shown in the posting journal as well formatted strings. 

I propose we make these links to documents and entities within BHIMA. This would provide valuable context for the transaction. 

**Current rows** 
link documents

Each of these reference patient invoices or patients themselves. Linking to the invoice receipts or patient cards would allow an accountant to directly view relevant notes/ users/ inventory items processed etc. in a transaction.

### LINKED ISSUE #1163: (refactor) Improvements to the Trial Balance

### LINKED ISSUE #1402: (bug) Posting Journal suffers JS rounding flaws
The Posting Journal's aggregation functions seem to suffer from the addition flaws that JavaScript has when adding two floating point numbers.  MySQL's Trial Balance thinks this transaction balances, but the Posting Journal is displaying a different view.

unbalancedbalancedtxn
_Fig 1: Unbalanced (but actually balanced)  transaction_

### LINKED ISSUE #1500: (refactor) Posting Journal Editing
Editing in the Posting Journal is fairly immature.  The following changes are needed to make sure that users can easily, effectively, and correctly use the Posting Journal:
 1. A single click to edit the Posting Journal
 2. A typeahead for editing accounts - filtered on account number and text
 3. A modal entry for entity linking.
 4. A modal entry for document linking.

### LINKED ISSUE #1640: About changing the date of a transaction in the posting journal
My preocupation is the following, what may be the reasons that may need to change a date in the posting journal,

In my opinion the Modules Journal voucher ansi that Simple Voucher already gives the possibility of anti date a transaction

And the modification of the date of a transaction does not repercute on the documents that generated this transaction

### LINKED ISSUE #1659: Bug in the display of data in posting journal,
Bug in the display of data in posting journal,
fileerror

Some transactions are mixed in the posting, which disrupts the operation of updating the data

### LINKED ISSUE #1716: proposal: remove the grouped transaction view from the TrialBalance
The Trial Balance allows a user to view all the transactions that are currently selected.  This makes the page excessively heavy, and the view doesn't give enough information to resolve several classes of errors (dates in the wrong period, etc) since it only shows ~4 columns.  Furthermore, to resolve these issues, a user will need to go back to the Journal anyway.

I propose that we remove this functionality.  I cannot find an issue proposing it to understand it's advantages.  #585 only describes the error modal and an initial design for #1559.

Remove it has the following advantages:
 1. Easier `$state` management - no need to pass Mb of transactions back and forth.
 2. Cleaner TB API - The Trial Balance only needs to know the `trans_id` for display and `record_uuid` for processing.
 3. Lighter weight ui-grid - we can remove the `ui-grid` grouping functionality, plus all imports needed by it since this is the functionality which uses it.
 4. Simpler UX - For a user, it greatly simplifies the procedure.  One page is for looking at account balances, another is for errors.  Both can be used to clear the Trial Balance and return to the Journal for edits.

### LINKED ISSUE #1717: refactor: use ng-class and standard colors for Trial Balance coloring
The current Trial Balance modal uses different colors for success/errors than the rest of the application.  You can see them below:
trialbalancedanger
_Fig 1: Trial Balance Danger State_

trialbalancesuccess
_Fig 2: Trial Balance Success State_

As part of standardizing out user experience, we should switch these to [bootstrap's contextual background colors](http://getbootstrap.com/css/#helper-classes-backgrounds).

The reasons are as follows: 
 1. We can configure them with less, along with the rest of the application in one place.
 2. A uniform UX ... the danger color will always be the same.  You can see the subtle difference between the "View Errors" button and the `ui-grid` header in the image above.
 3. Code reduction - no need for bespoke classes.
 4. The classes currently come from JavaScript, using the following line:
 ```js
cssClass = TrialBalance.getCSSClass(vm.feedBack); 
 ```
 This style of programming is very different from anything else we've written, making it hard to know where to go to change the behavior.  It also pollutes the JS with styling-related code.

### LINKED ISSUE #1724: bug(journal): debit/credit column grouping does not render currencies properly.
The Journal's debit/credit currency display can sometimes result in what look like imbalanced transactions.  This can be fixed with a simple call to `Number.prototype.toFixed(4)` as was previously done for the logical comparisons.  Now we need to do it for the display values.

See below:
bugdebitcreditgrouprounding
_Fig 1: Some values are very, very long, making it look unbalanced._

**Solution**
In the journal's `columnDefs`, their is a property `customTreeAggregationFinalizerFn`, which can take a function.  For example, [here is the function that renders dates]().  The programmer should simply create this function to the `debit_equiv`, `debit`, `credit_equiv`, and `credit` columns, and return `value.toFixed()`.

### LINKED ISSUE #1832: (proposal) Move Journal editing into a discrete modal
Currently our posting journal allows you to edit transactions in place. In transaction view you can click edit in a transaction, it will be highlighted blue with a number of editing functions (save, add line, remove line etc.) presented at the top. All rows in a transaction are downloaded whatever the data set to ensure you are not editing without context. 

The reason for this decision is to allow the user to watch the totals of their current query (journal data) change as they edit. 

However there are many UX issues with the current implementation, if a user has filtered on any values, the filter will be removed scrolling the view unpredictably. 

My proposal is that the edit button is removed from in line transaction and is instead placed as a function in the menu or the toolbar. The flow would be:
1. User selects one or more rows from a transaction 
2. User selects 'Edit' from the menu 
3. A modal is presented with all of the rows in that transaction in a UI grid, the experience for editing is the same as now. User can add, remove lines, save or cancel.
4. Once the user has saved they are returned to the exact same view they had before, with the underlying values updated 

The same code for editing lines in a UI grid would be used however it would be a discrete action that happens aside from the main grid. 

**Pros**
* No requirement on being in a certain view 
* Far fewer corner cases with in line scrolling and downloading of data and custom editing / rendering (this constitutes thousands of lines in the journal code) 
* discrete states, no ambiguous mode or rendering 
* No concerns about current filters or queries 

**Cons**
* This method would not be suited for editing many transactions at once. Editing would be limited to 1 transaction by 1. 
* Significant development time transitioning system 

Watching users use the current system with changes of filters and scrolling beyond their control it is clear that it would take a long time to master and also has some required fixes - potentially now is a good time to make a call on if we think the number of corner cases here is ever worth the convenience of in line editing. 

Potentially this would also allow other modules like the account statement to request a transaction for editing. This might be a very convenient tool if you are using other modules to validate data.

### LINKED ISSUE #1839: (bug) Trial Balance initialises with no data
I think this bug report may have been made before. Sometimes the trial balance loads with a blank interface displaying no transactions. When this happens it is necessary to close the trail balance and open it again. This is a very bad user experience, especially when the trial balance takes a long time to load.

### LINKED ISSUE #1921: (proposal) Make the patient financial history easily available
Our patient financial history report ties together all of the financial entities in BHIMA that relate to a patient. Primarily invoices and cash payments but also vouchers that reference the patient. 

Right now it is hidden in the patient overview module: I propose we make this easily accessible to the user in a number of ways: 
1. Directly link the patient financial report _anywhere_ that a patient human readable ID is used. This would use the work done in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1886 and https://github.com/IMA-WorldHealth/bhima-2.X/pull/1916.
(the patient card could also be linked in a multi-part component. For example [PA.HSP.100](https://patient-link) [[C]](https://card-link)
2. Register it a BHIMA report just like the account statement  - the user would be able to select a patient and preview and save the report. 

Additionally it would be good to treat this as an up to date report and test the data values produced by the financial history.

### LINKED ISSUE #1928: refactor: Post to General Ledger
We have received reports from HEV that posting to the General Ledger takes a very long time.  This is part of the motivation for a hardware upgrade that was installed recently.

On [analysis of the code](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/procedures.sql#L489), it appears that the SQL is formed using a large string - the one below:
```sql
"INSERT INTO stage_journal_transaction 
  SELECT DISTINCT record_uuid FROM posting_journal WHERE trans_id IN (", transactions, ");"
```

I believe this string may be the source of many of our problems - it requires a search on `trans_id` which is a string and does not leverage indexes properly.

This should be investigated - I believe if we use the `StageTransaction()` method such as the Trial Balance, we could gain some massive speed ups.

### LINKED ISSUE #1934: (refactor) Remove custom group selection logic
As of the latest version of UI grid, selecting a group header will select all individual rows. We currently assign a method to validate if a selected row is a tree level header and manually select rows. This should be replaced/ removed with the recent upgrade of UI grid.

### LINKED ISSUE #1943: fix(posting journal) scroll to edited transaction
After editing a transaction it is difficult to find the edited  transaction line directly.

This issue is about to scroll or add highlight to the edited transaction

### LINKED ISSUE #1944: fix(posting journal) : entity column linked to entity informations
Activate the link for the entity column like for the reference colulmn

### LINKED ISSUE #1948: fix(posting journal) : show account list in edit mode
In edit mode, most of the time, the account's section doesn't display the list of accounts when we are in account area and typing account information in the typeahead

### LINKED ISSUE #1950: The posted transactions must be accessible
By the fact that users can't use the posted journal (performance issue) there is no possibility to see if posted transactions are really in the general ledger

### LINKED ISSUE #1961: (bug) Journal editing fails on new row validation
The client side posting journal validation currently validates a transactions date is inside the period start and period end sent from the server.

This never works on rows that have been added because the period start and end attributes are not copied to these new rows. 

Effectively this means rows can never be added as client side validation always fails.

### LINKED ISSUE #2031: proposal: rename "Posting Journal" into "Journal"
Since we have deprecated the "Posted Journal" and moved it's view into a filter on the regular "Posting Journal", I think it might reduce a lot of confusion by renaming the "Posting Journal" to the "Journal".  It will be easier to refer to, and more semantically correct since we now allow viewing of posted records in the journal.

References work done in [refactor-journal-tools](https://github.com/IMA-WorldHealth/bhima-2.X/tree/refactor-journal-tools).

\cc @IMA-WorldHealth/bhima-core what do you all think?

### LINKED ISSUE #2041: discussion: Issue with UI-Grid Selection
By default, `ui-grid`'s selection doesn't seem to behave well with grouping.  When a group header is clicked, it does not select the group header.  See below:

bugingroupingselection
_Fig 1: Problematic Bug in Grouping Selection_

This can be overcome by keeping track of the transactions that are grouped, but it has to be a hack.  The previous version of the Journal (outside of `refactor-journal-tools`) did this in a complicated fashion - see the  [GridGrouping Service](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/grid/GridGrouping.js#L72).

The question is: what behavior do we want?  I see several possibilities:
 - [ ] Check the header row when _any_ element in the group is selected.
 - [ ] Only check the header row if all elements in the group are selected.
 - [ ] Default behavior (never check the header row).

To get the first two behaviors, we will need to manually track selection in a service.  There are some optimizations we can do, but it will be ugly no matter what.

What behavior do we think is the best?  Our initial feedback from users suggested they really wanted the header row to be checked at some point.

/cc @IMA-WorldHealth/bhima-core what do you think?

jniles:
This Pull Request finalizes the refactor of the Posting Journal and associated modules.  The commit history contains a full list of changes.  In brief:
 1. Editing has been moved out into a modal.  By breaking out the editing code, the complexity of the Posting Journal is reduced.  It also enforces editing a single transaction at a time.
 2. The Trial Balance has moved into Stored Procedures.  This separates the JS code from the SQL code and speeds up the execution of the Trail Balance slightly.
 3. The Trial Balance no longer reloads data between `$state` changes.  The modal is much more snappy than previous.
 4. Links to the receipts/documents/patients associated with each transaction are embedded in the Posting Journal.  This makes it easy to find the details associated with a given transaction.
 5. The Posting Journal is able to load "posted" transactions.  To accommodate this, the Posting Journal footer has been improved to show the number of unposted or posted transactions.  Additionally, the "Posted Journal" module has been removed.

There have been a number of miscellaneous bug fixes addressed as well.  As with any change, this may introduce new bugs specific to the changes listed above.

Partially addresses #1432.

Closes #1034.
Closes #1163.
Closes #1402.
Closes #1500.
Closes #1640.
Closes #1659.
Closes #1716.
Closes #1717.
Closes #1724.
Closes #1832.
Closes #1839.
Closes #1921.
Closes #1928.
Closes #1934.
Closes #1943.
Closes #1944.
Closes #1948.
Closes #1950.
Closes #1961.
Closes #2031.
Closes #2041.

---

### TODO Before Merge
 - [x] Ensure all keys are translated into both French and English
 - [x] Ensure tests pass
 - [x] Make sure all filters/links work on Posting Journal page.
 - [x] Make sure all filters/links work on the Trial Balance
 - [x] Ensure all editors work as expected.
 - [x] Ensure that new code passes lint checks.


DedrickEnc:
@jniles
This PR needs a rebase first

lomamech:
**Error analysis in the new version of the Input Journal**
1.  When you want to edit a transaction, the check box that precedes the transaction ID does not activate the Edit button, but only if you click on a line in the transaction
journal

2.  Validation functions do not seem to work well

-  Non-balanced transactions are validated

-   It is possible to edit on one line of a transaction at the same time the debit as well as the credit

-   It is also possible to validate a line of a transaction from the null values to the debit but also to the credit

jniles:
I believe the lack of validation was by design from @sfount.  I think the motivation was to make it really easy for a user to edit a transaction, but the posting process would still ensure that the transaction's rules were respected.

That being said, I would personally be more comfortable with stricter validation.  I'll write some more checks before this is merged.

Your first critique is a bug.  Good catch!  I never tested that path.  It may be related to #2041.

mbayopanda:
**Refactor Journal Tools – Assessment**

**1.	The comment tool has bug :**
- Generate an error when click on submit
- The comment doesn’t succeed

image

image

**2.	Selecting  a group of transactions doesn’t select children transactions**

image

**3. The text  “Transactions Vérifiées” must clearly refer to the posted transactions**
- The text can just be “Transactions Postées”

image

**4. The cash payment receipt has bug**
- Generate an error

image

**5. The edit tool is not strict in terms of validation**

image

The good thing is that the trial balance modal blocks well bad transactions

jniles:
@mbayopanda, thanks for the feedback!

I cannot reproduce the cash payment error.  Can you give me steps to reproduce?

The problem with UI Grid selection was reported in https://github.com/IMA-WorldHealth/bhima-2.X/issues/2041.  The main problem is that we are fighting ui-grid on this one - we will be changing default behavior if we try to hack this.  What do you think the best option is?

lomamech:
I think that the description of a transaction should be the same for all the lines of the transposition, so this property should be defined only once
journal

lomamech:
I discovered an error with respect to updating TPA18, TPA19, TPA20, TPA21 and TPA22 transactions with the bhima_test database, all of these transactions were only initializing with a single Accounts number, but the problem is that when trying to edit any of these transactions, the modal displays 60lines whatever transaction you choose
journal

But also, if one modifies for example the date, this modification will repercute on the other transactions

jniles:
@lomamech, that actually isn't an issue with this module - it was an issue with the posting procedure for Purchase Orders.  It was fixed in #2104, but I haven't updated that branch to master yet.

jniles:
@lomamech, @mbayopanda thanks for your feed back!

I have pushed some fixes that should address everything except validation - could you check out the branch and see if those fixes work as you expect?

lomamech:
@jniles, somes issues are fixed, but in regard to the description, it would have to be defined at once as for the date, or else it should be maintained like this

jniles:
@lomamech, you are right.   The description is allowed to be different on every line by design.  We actually change the description for things like billing services in invoices.  There's also a suggestion to put the inventory item description on each line of an invoice or cash payment (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1824).  For this reason, it is best to keep the ability to change each individual line's description.

jniles:
Alright, all feedback including validation has been addressed.

@DedrickEnc @mbayopanda can I get a code review?

mbayopanda:
@jniles

There are some new bugs found : 

1. Cannot post data into the general ledger

image

image

2. Cannot remove a comment

This behavior is also in account statement, we can only add and update a comment to a row, but we cannot remove completely a comment.

image

image

jniles:
@mbayopanda thanks for these catches!  I guess the comment PR wasn't properly vetted :(

I'll put some patches in by this afternoon.

jniles:
@dedrickenc, @mbayopanda this branch is ready for a code review.  All usability bugs reported by @lomamech and @mbayopanda have been addressed.  It looks like #2119 has resolved the bugs in Travis CI for now.

client/src/js/services/grid/GridSelection.js
```diff
@@ -0,0 +1,71 @@
...
+    var isHeaderSelected;
+    var children;
+
+    var getRowChildren = this._gridApi.treeBase.getRowChildren.bind(this);
+    var toggleRowSelectionFn = this._gridApi.selection.toggleRowSelection.bind(this);
+
+    // special treatment to header rows:
+    // if the header is selected, select all children.
+    // if the header is unselected, unselect all children.
+    if (isHeaderRow(row)) {
```
mbayopanda, line 47: Good comments and it works well 👍

client/src/modules/account_statement/modals/comment.modal.html
```diff
@@ -12,9 +9,8 @@
       <label translate>ACCOUNT_STATEMENT.SELECTED_ROWS</label>
     </div>
 
-    <div class="form-group"
-      ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.comment.$invalid }">
-      <input type="text" class="form-control" name="comment" ng-model="$ctrl.comment" required>
+    <div class="form-group" ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.comment.$invalid }">
+      <input type="text" class="form-control" name="comment" ng-model="$ctrl.comment">
```
mbayopanda, line 13: Good fix 👍

client/src/modules/general-ledger/general-ledger.ctrl.js
```diff
@@ -1,239 +0,0 @@
...
-  // export data into csv file
-  vm.exportFile = function exportFile() {
-    var url = '/reports/finance/posted_journal';
-    var params = formatExportParameters('csv');
-
-    if (!params) { return; }
-
-    Export.download(url, params, 'POSTING_JOURNAL.TITLE');
-  };
-}
```
mbayopanda, line 239: :+1

client/src/modules/general-ledger/general-ledger.html
```diff
@@ -1,80 +0,0 @@
...
-            error-state="GeneralLedgerCtrl.hasError"
-          >
-          </bh-grid-loading-indicator>
-
-        </div>
-      </div>
-    </div>
-  </div>
-</div>
-
```
mbayopanda, line 80: :+1; That is a refactor

client/src/modules/journal/journal.html
```diff
@@ -60,13 +55,30 @@
...
+        </div>
+        <div class="toolbar-item">
+          <button
+            ng-class="{ 'btn-success' : JournalCtrl.selection.selected.groups.length === 1 }"
+            ng-disabled="JournalCtrl.selection.selected.groups.length === 0"
+            ng-click="JournalCtrl.editTransactionModal()"
+            class="btn btn-default" data-method="edit">
+            <i class="fa fa-pencil-square-o"></i>
+              <span ng-if="JournalCtrl.selection.selected.groups.length === 1">({{JournalCtrl.selection.selected.groups[0]}})</span>
+              <span ng-if="JournalCtrl.selection.selected.groups.length > 1">(...)</span>
```
mbayopanda, line 76: :+1;

client/src/modules/journal/journal.service.js
```diff
@@ -133,5 +167,47 @@ function JournalService(Api, AppCache, Filters, Periods, Modal) {
     }).result;
   }
 
+   /**
+   * @method openCommentModal
+   * @param {object} request
+   */
+  function openCommentModal(request) {
```
mbayopanda, line 174: Is accepted that posted transactions cannot be commented ? The main reason for the comment tool is for analysis, so i think it will be good to be able to comment also posted transactions for analysis to.

client/src/modules/journal/modals/editTransaction.modal.html
```diff
@@ -0,0 +1,116 @@
...
+                </span>
+                </div>
+              </div>
+            </div>
+          </div>
+        </form>
+      </div>
+    </div>
+
+    <div class="alert alert-danger" id="validation-errored-alert" ng-if="ModalCtrl.validation.errored"  style="border-radius : 0px; padding-left : 5px; padding-right : 5px; padding-top : 10px; padding-bottom : 10px; margin-bottom : 10px;">
```
mbayopanda: The style 
```css
border-radius : 0px;
padding-left : 5px; 
padding-right : 5px;
padding-top : 10px;
padding-bottom : 10px;
margin-bottom : 10px;
```
Can be set in the `bhima-bootstrap.less` file because it is used more than one place.

client/src/modules/journal/modals/editTransaction.modal.js
```diff
@@ -0,0 +1,386 @@
...
+      if (!hasTransactionType) {
+        return ERROR_MISSING_TRANSACTION_TYPE;
+      }
+
+      var hasNegativeNumbers = (row.debit_equiv < 0 || row.credit_equiv < 0);
+      if (hasNegativeNumbers) {
+        return ERROR_NEGATIVE_NUMBERS;
+      }
+
+      window.util = util;
```
mbayopanda: What is the use of this `window.util = util` here ?

client/src/modules/journal/trial-balance/errors.js
```diff
@@ -0,0 +1,71 @@
+angular.module('bhima.controllers')
+  .controller('TrialBalanceErrorsController', TrialBalanceErrorsController);
+
+TrialBalanceErrorsController.$inject = ['TrialBalanceService', 'NotifyService'];
+
+/**
+ * @module TrialBalanceErrorsController
```
mbayopanda: Is it appropriate to describe a controller as  a module ?
```js
/**
* @module TrialBalanceErrorsController
*/
```

client/src/modules/journal/trial-balance/overview.js
```diff
@@ -0,0 +1,212 @@
...
+        vm.subGridOptions.data = data;
+      })
+      .catch(function () {
+        vm.hasSubGridErrors = true;
+      })
+      .finally(function () {
+        vm.loadingSubGrid = false;
+      });
+  }
+  */
```
mbayopanda: can this commented block be removed ?

client/src/modules/journal/trial-balance/overview.js
```diff
@@ -0,0 +1,212 @@
...
+    expandableRowScope : vm,
+    enableColumnResizing : true,
+    onRegisterApi : onRegisterApi,
+  };
+
+  function onRegisterApi(api) {
+    vm.gridApi = api;
+
+    // TODO - should we really cache the trial balance transactions?
+    // api.expandable.on.rowExpandedStateChanged(null, loadSubGridRecords);
```
mbayopanda: if the `loadSubGridRecords` function is removed, these commented lines can also be removed.

server/controllers/finance/reports/journal/report.handlebars
```diff
@@ -57,15 +52,15 @@
...
+      {{translate 'TABLE.COLUMNS.TOTAL'}} {{translate 'TABLE.COLUMNS.DEBIT'}}: <strong>{{currency totals.debit }}</strong>
     </div>
 
     <div class="col-xs-4">
-      {{translate 'TABLE.COLUMNS.TOTAL'}} {{translate 'TABLE.COLUMNS.CREDIT'}}: <strong>{{currency (sum rows 'credit_equiv')}}</strong>
+      {{translate 'TABLE.COLUMNS.TOTAL'}} {{translate 'TABLE.COLUMNS.CREDIT'}}: <strong>{{currency totals.credit }}</strong>
     </div>
 
     <div class="col-xs-4">
-      {{translate 'TABLE.COLUMNS.RESULT'}}: <strong>{{currency (substract (sum rows 'debit_equiv') (sum rows 'credit_equiv'))}}</strong>
```
mbayopanda, line 68: 👍

mbayopanda (approved):
This is an awesome refactoring job.

There are some non-blocking comments but this PR LGTM

mbayopanda:
@jniles could you update this branch with the master branch ?

client/src/modules/journal/modals/editTransaction.modal.js
jniles: Ha!  Good catch.

jniles:
@mbayopanda, thanks for the review.  I'll fix those comments and ping you back.

client/src/modules/journal/journal.service.js
jniles: Oh man, that's a bug.

client/src/modules/journal/trial-balance/errors.js
jniles: Updated to `@overview`.

jniles:
@mbayopanda, changes landed!

PR #2088, opened 2017-09-11
feat(cashbox_management) proposing the current project to user 

jeremielodi:
This PR help to  propose to the user the current project instead of obliging him to seclect one in the dropdown list. In this case hi can, if he wants to create a cash in another project, choose it in the dropdown list.

### DIFF
--- a/client/src/modules/cash/cashboxes/update/update.js
+++ b/client/src/modules/cash/cashboxes/update/update.js
@@ -3,10 +3,10 @@ angular.module('bhima.controllers')
 
 CashboxUpdateController.$inject = [
   '$state', '$uibModal', 'ModalService', 'NotifyService',
-  'CashboxService', 'CurrencyService',
+  'CashboxService', 'CurrencyService', 'SessionService',
 ];
 
-function CashboxUpdateController($state, Modal, ModalService, Notify, Boxes, Currencies) {
+function CashboxUpdateController($state, Modal, ModalService, Notify, Boxes, Currencies, Session) {
   var vm = this;
 
   var CREATE_STATE = 'cashboxes.create';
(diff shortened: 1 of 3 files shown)


jniles (approved):
This is exactly right. :+1:

jniles:
@jeremielodi, it seems this broke the end to end tests.  Can you try to debug it?

client/src/modules/cash/cashboxes/update/update.js
```diff
@@ -17,7 +17,7 @@ function CashboxUpdateController($state, Modal, ModalService, Notify, Boxes, Cur
   var cashboxUuid = $state.params.uuid;
 
   vm.state = $state;
-  vm.box = {};
+  vm.box = { project_id : Session.project};
```
jniles: Taking another look at this when I am less busy .... why isn't this
```js
vm.box = { project_id : Session.project.id };
```
?

jeremielodi:
The selected project should be a object, not a simple value

Le 16 sept. 2017 10:18, "Jonathan Niles" <notifications@github.com> a
écrit :

jniles:
Alright, we'll put it in.

bors r+

jniles:
It [looks like these changes break tests](https://travis-ci.org/IMA-WorldHealth/bhima-2.X/jobs/277225378#L1940).  Specifically, the end-to-end tests for cashboxes do not work.  This will take some debugging.

Could you investigate?  It should all be contained in the [cashboxes.spec.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/cashboxes/cashboxes.spec.js) file.  If you need references, here is [a link to the Protractor API](http://www.protractortest.org/#/api).

PR #2092, opened 2017-09-12
Accounting writing for stock integration

DedrickEnc:
This pull request add a functionality to pass some writing in posting journal for the stock integration.

### DIFF
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -341,7 +341,7 @@ function StockEntryController(
     var entry = {
       integration : { date : vm.movement.date, description : vm.movement.description },
       lots : lots,
-      movement : movement,
+      movement : movement
     }
 
     Stock.integration.create(entry)
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -324,23 +324,10 @@ function StockEntryController(
       flux_id     : bhConstants.flux.FROM_INTEGRATION,
       user_id     : Session.user.id,
     };
-
-    var lots = vm.Stock.store.data.reduce(function (current, previous) {
-        return previous.lots.map(function (lot) {
-          return {
-            label            : lot.lot,
-            initial_quantity : lot.quantity,
-            quantity         : lot.quantity,
-            unit_cost        : previous.unit_cost,
-            expiration_date  : lot.expiration_date,
-            inventory_uuid   : previous.inventory.uuid
-          };
-        }).concat(current);
-      }, []);    
     
     var entry = {
       integration : { date : vm.movement.date, description : vm.movement.description },
-      lots : lots,
+      lots : processLotsFromStore(vm.Stock.store.data, movement.entity_uuid),
       movement : movement
     }
 
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -326,7 +326,6 @@ function StockEntryController(
     };
     
     var entry = {
-      integration : { date : vm.movement.date, description : vm.movement.description },
       lots : processLotsFromStore(vm.Stock.store.data, movement.entity_uuid),
       movement : movement
     }
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -333,14 +333,13 @@ function StockEntryController(
             quantity         : lot.quantity,
             unit_cost        : previous.unit_cost,
             expiration_date  : lot.expiration_date,
-            inventory_uuid   : previous.inventory.uuid,
-            origin_uuid      : uuid,
+            inventory_uuid   : previous.inventory.uuid
           };
         }).concat(current);
       }, []);    
     
     var entry = {
-      integration : { description : vm.movement.description },
+      integration : { date : vm.movement.date, description : vm.movement.description },
       lots : lots,
       movement : movement,
     }
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -315,7 +315,7 @@ function StockEntryController(
   }
 
   // submit integration
-  function submitIntegration() {
+  function submitIntegration() {    
     var movement = {
       depot_uuid  : vm.depot.uuid,
       entity_uuid : null,
@@ -325,17 +325,33 @@ function StockEntryController(
       user_id     : Session.user.id,
     };
 
-    Stock.integration.create({ description : vm.movement.description })
-      .then(function (uuid) {
-        movement.lots = processLotsFromStore(vm.Stock.store.data, uuid);
-        return Stock.stocks.create(movement);
-      })
-      .then(function (document) {
-        vm.Stock.store.clear();
-        vm.movement = {};
-        ReceiptModal.stockEntryIntegrationReceipt(document.uuid, bhConstants.flux.FROM_INTEGRATION);
-      })
-      .catch(Notify.handleError);
+    var lots = vm.Stock.store.data.reduce(function (current, previous) {
+        return previous.lots.map(function (lot) {
+          return {
+            label            : lot.lot,
+            initial_quantity : lot.quantity,
+            quantity         : lot.quantity,
+            unit_cost        : previous.unit_cost,
+            expiration_date  : lot.expiration_date,
+            inventory_uuid   : previous.inventory.uuid,
+            origin_uuid      : uuid,
+          };
+        }).concat(current);
+      }, []);    
+    
+    var entry = {
+      integration : { description : vm.movement.description },
+      lots : lots,
+      movement : movement,
+    }
+
+    Stock.integration.create(entry)
+    .then(function (document) {
+      vm.Stock.store.clear();
+      vm.movement = {};
+      ReceiptModal.stockEntryIntegrationReceipt(document.uuid, bhConstants.flux.FROM_INTEGRATION);
+    })
+    .catch(Notify.handleError);
   }
 
   // submit donation
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -102,12 +102,13 @@ function createStock(req, res, next) {
     transaction.addQuery(createMovementQuery, [createMovementObject]);
   });
 
-  // An arry of common info, to send to the store procedure in order to insert to the posting journal
+  // An array of common info, to send to the store procedure in order to insert to the posting journal
   const commonInfos = [
     db.bid(document.uuid), document.date,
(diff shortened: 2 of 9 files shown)


server/controllers/stock/index.js
```diff
@@ -329,21 +336,73 @@ function listLotsOrigins(req, res, next) {
...
+      db.bid(documentUuid),
+      new Date(params.movement.date),
+      req.session.enterprise.id,
+      req.session.project.id,
+      req.session.enterprise.currency_id,
+      req.session.user.id
+    ];
+
+    // writting all records relative to the movement in the posting journal table
+    transaction.addQuery('CALL PostIntegration(?)', [commonInfos]);
```
jniles, line 398: Does this produce any kind of financial document a user can print out?  It looks like it just writes straight into the Posting Journal.  Should it link to some form of document somewhere?  Could it create an intermediary voucher?

server/models/procedures.sql
```diff
@@ -1445,4 +1445,113 @@ BEGIN
...
+  );
+
+  -- getting the period id
+  SET current_period_id = (
+    SELECT id FROM period AS p
+    WHERE DATE(date) BETWEEN DATE(p.start_date) AND DATE(p.end_date) AND p.fiscal_year_id = current_fiscal_year_id
+  );
+
+  -- getting the transaction number
+  SET transaction_id = GenerateTransactionId(project_id);
```
jniles: This should probably be done after the check for invalid accounts.  That way, it is declared closer to where it is used (and easier to find), and we are less likely to generate duplicate transaction IDs.

server/models/procedures.sql
```diff
@@ -1445,4 +1445,113 @@ BEGIN
...
+  IN enterprise_id SMALLINT(5) UNSIGNED,
+  IN project_id SMALLINT(5) UNSIGNED,
+  IN currency_id TINYINT(3) UNSIGNED,
+  IN user_id SMALLINT(5) UNSIGNED
+)
+BEGIN
+  DECLARE InvalidInventoryAccounts CONDITION FOR SQLSTATE '45006';
+  DECLARE current_fiscal_year_id MEDIUMINT(8) UNSIGNED;
+  DECLARE current_period_id MEDIUMINT(8) UNSIGNED;
+  DECLARE current_exchange_rate DECIMAL(19, 4) UNSIGNED;
```
jniles: It doesn't look like you ever use `current_exchange_rate`.  Could you remove it and any other declarations you don't need?

server/models/bhima.sql
```diff
@@ -139,7 +139,8 @@ INSERT INTO `transaction_type` (`id`, `text`, `type`, `prefix`, `fixed`) VALUES
   (8, 'VOUCHERS.SIMPLE.CASH_RETURN', 'expense', 'PAYBACK', 1),
   (9, 'VOUCHERS.SIMPLE.PURCHASES', 'expense', 'ACHAT', 1),
   (10,'VOUCHERS.SIMPLE.CREDIT_NOTE', 'other', 'CREDIT NOTE', 1),
-  (11,'VOUCHERS.SIMPLE.INVOICING', 'income', 'INV', 1);
+  (11,'VOUCHERS.SIMPLE.INVOICING', 'income', 'INV', 1),
+  (12, 'VOUCHERS.SIMPLE.STOCK_INTEGRATION', 'stock integration', 'STOCK. INT', 1);
```
jniles, line 143: At some point in the future, we should generate a custom JSON translation file for these.

jniles (commented):
@DedrickEnc, just a few comments on this - we'll need to fix the bug on #2094 before this can be successfully merged.

server/controllers/stock/index.js
DedrickEnc: There is a document produced if the integration is done with success but I can not say if it is a financial document or not.

DedrickEnc:
@jniles
Can I have a review, it will be easy to write into voucher once everything is in master

client/src/modules/stock/entry/entry.js
```diff
@@ -312,17 +312,32 @@ function StockEntryController(
...
-        movement.lots = processLotsFromStore(vm.Stock.store.data, uuid);
-        return Stock.stocks.create(movement);
-      })
-      .then(function (document) {
-        vm.Stock.store.clear();
-        vm.movement = {};
-        ReceiptModal.stockEntryIntegrationReceipt(document.uuid, bhConstants.flux.FROM_INTEGRATION);
-      })
-      .catch(Notify.handleError);
+    var lots = vm.Stock.store.data.reduce(function (current, previous) {
```
jniles: Can this use `processLotsFromStore()`?  It would clean up the code a lot.

server/controllers/stock/index.js
```diff
@@ -102,13 +102,14 @@ function createStock(req, res, next) {
     transaction.addQuery(createMovementQuery, [createMovementObject]);
   });
 
-  // An arry of common info, to send to the store procedure in order to insert to the posting journal
+  // An array of common info, to send to the store procedure in order to insert to the posting journal
```
jniles, line 105: :dancer:

server/controllers/stock/index.js
jniles: From the code, it looks like this may suffer from the same bug as #2102.  Can you verify that multiple lots will not result in multiple transactions?

jniles (commented):
@DedrickEnc, just a few more things to check.  Otherwise, LGTM.

server/controllers/stock/index.js
```diff
@@ -336,21 +337,76 @@ function listLotsOrigins(req, res, next) {
...
   const identifier = uuid.v4();
+  const documentUuid = uuid.v4();
+  let commonInfos;
+
   const integration = {
     uuid        : db.bid(identifier),
     project_id  : req.session.project.id,
-    description : params.description || 'INTEGRATION',
-    date        : new Date(),
+    description : params.integration.description || 'INTEGRATION',
```
jniles: Hmm.. it seems like we should either make this required or do this on the client.  It seems like magic that this word is written into the database.  I know you didn't write this code originally, so we should just schedule it for a later refactor.  Maybe when we introduce vouchers.

server/controllers/stock/index.js
```diff
@@ -336,21 +337,76 @@ function listLotsOrigins(req, res, next) {
...
+      date          : new Date(params.movement.date),
+      quantity      : lot.quantity,
+      unit_cost     : lot.unit_cost,
+      is_exit       : 0,
+      user_id       : params.movement.user_id,
+    });
+  });
+
+    // An arry of common info, to send to the store procedure in order to insert to the posting journal
+    commonInfos = [
```
jniles: Since `commonInfos` never changes, it could be  a`const`.

server/models/procedures.sql
```diff
@@ -1463,4 +1463,115 @@ BEGIN
...
+  SET current_period_id = (
+    SELECT id FROM period AS p
+    WHERE DATE(date) BETWEEN DATE(p.start_date) AND DATE(p.end_date) AND p.fiscal_year_id = current_fiscal_year_id
+  );
+
+  CALL PostingJournalErrorHandler(enterprise_id, project_id, current_fiscal_year_id, current_period_id, 1, date);
+
+ -- Check that all every inventory has a stock account and a variation account - if they do not the transaction will be Unbalanced
+  SELECT 
+    COUNT(l.uuid) INTO verify_invalid_accounts
```
jniles, line 1504: This is a good idea!  I wonder if we can also do this on the client when we download the inventory for the first time...  That way someone doesn't spend a lot of time configuring lots to realize that their inventory is not set up properly...

server/controllers/stock/index.js
```diff
@@ -336,21 +337,76 @@ function listLotsOrigins(req, res, next) {
...
+  const documentUuid = uuid.v4();
+  let commonInfos;
+
   const integration = {
     uuid        : db.bid(identifier),
     project_id  : req.session.project.id,
-    description : params.description || 'INTEGRATION',
-    date        : new Date(),
+    description : params.integration.description || 'INTEGRATION',
+    date        : new Date(params.integration.date),
```
jniles: In some places you use `params.integration.date` and other places you use `params.movement.date`.  Is there a reason for this?  Are they always the same date?

server/models/procedures.sql
```diff
@@ -1463,4 +1463,115 @@ BEGIN
...
+-- Crediting cost of good sale account, by inserting a record to the posting journal table
+  INSERT INTO posting_journal 
+  (
+    uuid, project_id, fiscal_year_id, period_id, trans_id, trans_date,
+    record_uuid, description, account_id, debit, credit, debit_equiv,
+    credit_equiv, currency_id, origin_id, user_id
+  ) 
+  SELECT
+    HUID(UUID()), project_id, current_fiscal_year_id, current_period_id, transaction_id, 
+    date, i.uuid, sm.description, ig.cogs_account, 0, l.quantity * l.unit_cost, 0, l.quantity * l.unit_cost, currency_id,
```
jniles, line 1561: :+1: Doing math in SQL.

jniles (commented):
This code is here mostly looks good.  It can be merged anytime.

Could look into the `params.integration.date` and make sure that it is behaving correctly? That code looks like it may have a bug in it somewhere.  If not, you can "bors r+" this PR.

bors delegate+

server/controllers/stock/index.js
DedrickEnc: I agree, I propose to tackle it in voucher introduction

server/controllers/stock/index.js
DedrickEnc: When initializing the integration object I used ```params.integration.date```, when inserting the movement object I used ```params.movement.date```. In the commonInfos I have just chosen one of them. SO in theory it is the same date but they are sent by the client so we should be careful and process them as they are provided.

jniles: Hmm...  in my view, the client should only send one date if they are going to be the same - otherwise it is a [code smell](https://en.wikipedia.org/wiki/Code_smell).

server/models/procedures.sql
DedrickEnc: So what do you propose ?
A filter to fetch only inventories which have all accounts ? or a logic to check if a lot contains an inventory which has every account ?

jniles: I'll make an issue about it and link it here.

server/controllers/stock/index.js
DedrickEnc: Oops! you are right, fixed now!

server/models/procedures.sql
jniles: See https://github.com/IMA-WorldHealth/bhima-2.X/issues/2115.

Issue #1986, opened 2017-08-15
bug: bhDepotSelect shows "no depot" message on slow network


jniles:
The bhDepotSelect component shows "No Depots" error message when the network is slow.  This can be very confusing since there are depots available, but they are simply not loaded.  It should show a loading indicator instead.

![bugnodepotwhendepotactuallyexists](https://user-images.githubusercontent.com/896472/29318687-13a4385e-81c9-11e7-9b91-253dad660566.png)
_Fig 1: Misleading Error when Depots Actually are Loading_

PR #2098, opened 2017-09-14
fix: stock depot selection issues

jniles:
This PR addresses multiple issues associated with depot selection in the stock management modules.  It implements the suggestion in #2003 and removes the `bhDepotDropdown`.

The basic idea is to make depot selection like cashbox selection, where a user is presented a modal if they have no depot selected. This modal is shown below:
depotselectionfirsttime
_Fig 1: Depot Selection if no depot is selected_

Depots are cached between selections, so a page will always use the selected depot.  They are not (yet) cached between pages ... although this would be as simple as changing the cache key.

Changing the depot is accessible via the dropdown menu as on most pages:
depotselectionmenu
_Fig 2: Depot Selection Menu_

Any subsequent selections bring up the following modal:
depotselectionallothertimes
_Fig 3: Depot Selection Modal All Other Times_

Closes #2003.  Closes #2094.  Closes #1986.

PR #2099, opened 2017-09-14
feat(bhDepotDropDown) using a loading indicator

jeremielodi:
This PR add a loading indicator in the bhDepotDropDown component
It solve this issue #1986


jniles:
This is going to be removed in https://github.com/IMA-WorldHealth/bhima-2.X/pull/2098.

Issue #975, opened 2016-11-29
(refactor) Price List allows duplicate items to be selected, but not submitted


jniles:
The Price Lists have a confusing user interface that allows a user to select the same item to add twice to a list, but will not allow duplicates to be submitted.  This can be solved several ways:
 1. Do not even show the duplicates in the ui-select 
 2. Disable the duplicates in the ui-select.

No matter the case, the error message is every generic "You are not allowed to perform this operation."  We should improve these errors to be more specific - "The price list cannot contain duplicates."  "This price list is assigned to a group, you cannot delete it."

jniles:
Closed in #2103.

PR #2103, opened 2017-09-18
feat(price list) avoid duplicated items in the list

jeremielodi:
This PR help to avoid duplicated items to be selected ,  you can look at this issue for more details #975

### DIFF
--- a/client/src/modules/prices/modal.js
+++ b/client/src/modules/prices/modal.js
@@ -2,10 +2,10 @@ angular.module('bhima.controllers')
 .controller('PriceListModalController', PriceListModalController);
 
 PriceListModalController.$inject = [
-  '$uibModalInstance', 'InventoryService', 'util', 'NotifyService'
+  '$uibModalInstance', 'InventoryService', 'util', 'NotifyService', 'appcache',
 ];
 
-function PriceListModalController( $uibModalInstance, Inventory, util, Notify) {
+function PriceListModalController( $uibModalInstance, Inventory, util, Notify, AppCache) {
   var vm = this;
 
   // bind variables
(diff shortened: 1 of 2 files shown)


client/src/modules/prices/modal.js
```diff
@@ -14,10 +14,27 @@ function PriceListModalController( $uibModalInstance, Inventory, util, Notify) {
   vm.action = null;
   vm.isValidItem = isValidItem;
 
-  // load Inventory
+  // Here we load the All items from the list
+  const cache = new AppCache('selecedItems');
```
jniles: We cannot use `const` or `let` or any ES2015+ features on the client.  It will break our production builds.  These should be `var`s until we start transpiling to ES5.

client/src/modules/prices/modal.js
```diff
@@ -14,10 +14,27 @@ function PriceListModalController( $uibModalInstance, Inventory, util, Notify) {
...
+  const selectedItems = cache.items;
+
   Inventory.read()
     .then(function (inventory) {
-      vm.inventories = inventory;
+      // Let filter the list,
+      // all items in from the list should not appear in the select dropdown list
+
+      vm.inventories = inventory.map(function (item) {
```
jniles: This is an interesting idea!

I think what you want to do here is actually an [Array.prototype.map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) and then an [Array.prototype.filter()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter).  Here's how you can make this really simple:
```js
// gather all used uuids in a flat array
var selectedUuids = selectedItems.map(function (item) {
  return item.inventory_uuid;
});

// filter the inventory to only show ones that are not used.
vm.inventories = inventory.filter(function (item) {
  return selectedUuids.indexOf(item) === -1;
});
```

client/src/modules/prices/prices.js
```diff
@@ -23,6 +25,9 @@ function PriceListController(PriceListService, $uibModal, Inventory, ModalServic
   vm.length250 = util.length250;
   vm.maxLength = util.maxTextLength;
 
+  // Here we create a cache, all items in the list will be in it
+  var cache = new AppCache('selecedItems');
```
jniles: This cache should be called `selectedItems`.

jniles (changes requested):
@jeremielodi this idea is good!  The implementation can be improved.  I've made a few suggestions down below.  Let me know what you think.

jniles (approved):
LGTM

jniles:
Seems like we still have a flaky test in `master`.

bors r+

PR #2105, opened 2017-09-19
feat(bh-add-item)

### LINKED ISSUE #2095: feature request: make the "add items" into a component
We've used a pattern in most creation modules - Invoices, Purchase Orders, Vouchers, etc.  It involves a "add items" button to add items to an underlying grid.  This is ripe for encapsulation.  See the button below:

addbutton
_Fig 1: Add Button could be a component_

lomamech:
Implementation of bh-add-item component, using this component in all module who used Add Bouton
- Patient Invoice
- Stock Entry
- Stock Exit
- Purchase Order
And Complexe Voucher

closes #2095


client/src/js/components/bhAddItem.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+function addItemController() {
+  var $ctrl = this;
+
+  $ctrl.$onInit = function onInit() {    
+    // default for form name
+    $ctrl.name = $ctrl.name || 'AddItemForm';
+  };
+
+  // fires the onSelectCallback bound to the component boundary
```
jniles: I don't think there is a `onSelectCallback`.  Could you remove this comment or update it?

client/src/modules/invoices/patientInvoice.html
```diff
@@ -110,25 +110,13 @@
...
-              type="number"
-              class="form-control"
-              ng-model="PatientInvoiceCtrl.itemIncrement"
-              style="max-width : 40px;"
-              ng-disabled="!PatientInvoiceCtrl.Invoice.recipient">
-          </div>
+         <!-- "Component Add number of grid rows" input-group -->
+          <bh-add-item
+            item-increment = "PatientInvoiceCtrl.itemIncrement"
+            name = "add_item_invoice"
```
jniles: Since this isn't ever going to be submitted as part of a form, there is no need for a `name` binding.  `name` is only useful for form validation or submission to servers.

client/src/js/components/bhAddItem.js
```diff
@@ -0,0 +1,35 @@
+angular.module('bhima.components')
+  .component('bhAddItem', {
+    templateUrl : 'modules/templates/bhAddItem.tmpl.html',
+    controller  : addItemController,
+    transclude  : true,
+    bindings    : {
+      itemIncrement    : '<',
+      recipient        : '<',
```
jniles: I don't see the value of `recipient` ... it looks like just another way to `disable` the input.  Since you already have the `disable` parameter, I propose you remove `recipient`.

client/src/modules/purchases/create/create.html
```diff
@@ -134,23 +134,13 @@
...
-          </span>
-          <input
-            type="number"
-            class="form-control"
-            ng-model="PurchaseCtrl.itemIncrement"
-            style="max-width : 40px;"
-            ng-disabled="!PurchaseCtrl.supplier">
-          </div>
+        <bh-add-item
+          item-increment = "PurchaseCtrl.itemIncrement"
```
jniles: In my view, the component should handle this itself - there is no reason for the parent controller to care about the item increment.  That will clean up the component bindings and the parent controllers significantly.

client/src/modules/templates/bhAddItem.tmpl.html
```diff
@@ -0,0 +1,20 @@
+<div ng-form="{{ $ctrl.name }}" bh-add-item ng-model-options="{ updateOn: 'default' }">
```
jniles: There's no need to make an `ngForm` here ... we don't perform any form validation.  `ng-model-options` can also be removed (since it only makes sense to define on an `ngForm` or `ngModel`.

jniles (changes requested):
@lomamech, this is about half way there.

The skeleton of this component is good, but we can go further.  For example, the `recipient` and `disable` bindings are essentially the same: I propose you keep `disable` since it describes what will happen to the component if the binding is `true`.

Ideally, the basic component should look something like this:
```html
<bh-add-item callback="SomeCtrl.onChange(numItems)"></bh-add-item>
```

That is all we need for most modules.  The basic implementation can be optionally extended by specifying an initial `item-increment` and/or `disable` parameter.  The item increment can be a simple `@` binding that you parse with a call to `Number()`.  It will default to 1 if nothing is passed in.

This will make the component much simpler.  At the moment, it looks quite complicated!

client/src/js/components/bhAddItem.js
```diff
@@ -0,0 +1,35 @@
+angular.module('bhima.components')
+  .component('bhAddItem', {
+    templateUrl : 'modules/templates/bhAddItem.tmpl.html',
+    controller  : addItemController,
+    transclude  : true,
```
jniles: Transclusion is not necessary for this component.

client/src/js/components/bhAddItem.js
```diff
@@ -0,0 +1,35 @@
...
+    bindings    : {
+      disable          : '<',
+      callback         : '&?',
+    },
+  });
+
+addItemController.$inject = [];
```
jniles: You can go ahead and remove this line if we never `$inject` everything.

client/src/js/components/bhAddItem.js
```diff
@@ -0,0 +1,35 @@
+angular.module('bhima.components')
+  .component('bhAddItem', {
+    templateUrl : 'modules/templates/bhAddItem.tmpl.html',
+    controller  : addItemController,
+    transclude  : true,
+    bindings    : {
+      disable          : '<',
+      callback         : '&?',
```
jniles: This should be required (`&` instead of `&?`).  The module will break without it!

jniles (approved):
@lomamech, just a few clean up issues.  Otherwise, this looks perfect :+1:

client/src/js/components/bhAddItem.js
```diff
@@ -0,0 +1,32 @@
+angular.module('bhima.components')
+  .component('bhAddItem', {
+    templateUrl : 'modules/templates/bhAddItem.tmpl.html',
+    controller  : addItemController,
+    bindings    : {
+      disable     : '<',
+      Callback    : '&',
```
jniles: I think this should not be capital - that may be what is breaking your tests.

jniles:
bors r+

... try again.

jniles:
?

bors r+

Issue #1342, opened 2017-03-12
(deployment) Set up continuous deployment on low-end hardware


jniles:
Our recent issues with the production server at HEV show that we need to be continuously testing on low-resource environments with high data loads.  I propose that we tackle this in two steps:
 1. Set up an office Raspberry Pi to test out the loads on a no-network environment.  This should be able to automatically deploy with a push from git.
 2. Set up a low-resource cloud instance that can test our loads in a high-latency network environment.  This should also be able to be deployed via a github hook.

jniles:
Whoops, wrong reference.

jniles:
This is likely to high effort for what we are currently aiming for.  We no longer target the raspberrypi and most hardware is fast enough that these concerns are not really relevant anymore.

PR #2106, opened 2017-09-19
feat(exporting file) downloading reports in excel 

jeremielodi:
This PR brings a new way to export reports.
Instead of exporting reports in PDF  , CSV or JSON we can now do the same thing in xlsx ( Excel extension). I used the json2xls library to make it possible

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/1342.

### DIFF
--- a/client/src/modules/general-ledger/general-ledger-accounts.html
+++ b/client/src/modules/general-ledger/general-ledger-accounts.html
@@ -12,22 +12,22 @@
           <a class="btn btn-default" uib-dropdown-toggle>
             <span class="fa fa-bars"></span> <span class="hidden-xs" translate>FORM.LABELS.MENU</span> <span class="caret"></span>
           </a>
-          
+
           <ul uib-dropdown-menu role="menu" class="dropdown-menu-right">
             <li role="menuitem">
               <a href data-method="configure" ng-click="GeneralLedgerAccountsCtrl.openFiscalYearConfiguration()">
                 <i class="fa fa-calendar"></i> <span translate>FISCAL.FISCAL_YEAR</span>
               </a>
-            </li> 
+            </li>
 
-            <li role="separator" class="divider"></li>          
+            <li role="separator" class="divider"></li>
             <li role="menuitem">
               <a href data-method="configure" ng-click="GeneralLedgerAccountsCtrl.openColumnConfiguration()">
                 <i class="fa fa-columns"></i> <span translate>FORM.LABELS.COLUMNS</span>
               </a>
-            </li> 
+            </li>
 
-            <li role="separator" class="divider"></li>  
+            <li role="separator" class="divider"></li>
             <li role="menuitem">
               <a href data-method="save-state" ng-click="GeneralLedgerAccountsCtrl.saveGridState()">
                 <i class="fa fa-save"></i> <span translate>FORM.BUTTONS.SAVE_GRID_CONFIGURATION</span>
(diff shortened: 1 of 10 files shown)


client/src/modules/general-ledger/general-ledger-accounts.html
```diff
@@ -46,11 +46,18 @@
                 <span class="fa fa-file-pdf-o"></span> <span translate>DOWNLOADS.PDF</span>
               </a>
             </li>
+
+            <li role="menuitem">
+              <a ng-href="/reports/finance/general_ledger/?{{ GeneralLedgerAccountsCtrl.download('xlsx', GeneralLedgerAccountsCtrl.filters) }}" download="{{GeneralLedgerAccountsCtrl.fiscalYearLabel | translate }}">
+                <span class="fa fa-file-excel-o"></span> <span translate>EXCEL</span>
```
jniles: This key should probably be something like "DOWNLOADS.EXCEL".

server/controllers/finance/reports/generalLedger/index.js
```diff
@@ -57,7 +57,17 @@ function renderReport(req, res, next) {
       return report.render(data);
     })
     .then((result) => {
+      //if(result.Content-Type : 'application/xls',)
+
+      console.log(result.headers);
```
jniles: This should be removed before a merge.

server/app.js
```diff
@@ -20,6 +20,8 @@
  * @license GPL-2.0
  * @copyright IMA World Health 2016
  */
+// json2xls haldles an error in strict mode so we require it before
+var json2xls = require('json2xls');
```
jniles: :+1:  This should use `const` since it never changes.

server/lib/renderers/xlsx.js
```diff
@@ -0,0 +1,36 @@
...
+ * JSON Render Method
+ * @param {Object} data      Contains data in any format that will be used to
+ *                            drive the final report - this will be exposed to
+ *                            the view by the renderer.
+ * @param {Object} options   (Optional) parameters that can be passed to switch
+ *                            render features on/off.
+ * @returns {Object}          JSON Object representing a report that can be sent
+ *                            to the client.
+ */
+function renderJSON(data) {
```
jniles: Do you need this method?  If so, the name and comment should probably be updated.

server/controllers/finance/reports/generalLedger/index.js
```diff
@@ -57,7 +57,17 @@ function renderReport(req, res, next) {
       return report.render(data);
     })
     .then((result) => {
+      //if(result.Content-Type : 'application/xls',)
```
jniles: This comment should be removed before merge.

jniles (commented):
Looks pretty good!  A few clean up issues and then it could be merged.

server/lib/renderers/xlsx.js
```diff
@@ -0,0 +1,36 @@
...
+  'Content-Type' : 'application/xlsx',
+  'type' : 'xlsx',
+};
+
+exports.render = renderXLSX;
+exports.extension = '.xlsx';
+exports.headers = headers;
+
+/**
+ * JSON Render Method
```
jniles, line 25: This comment should be updated or removed.

jniles (approved):
Looks good to me.  I'll give this a local test and see how it works.  If everything passes, I'll give it a merge.

jniles:
*sigh*

bors r+

PR #2107, opened 2017-09-19
Fix stock transfer date

### LINKED ISSUE #2081: Bad date for stock entry from transfert
When we enter a stock from transfer, the date of the entry is not considered; the movement of entry still has the date of exit.

The normal behavior is to have exit distinct to entry also in terms of dates.

mbayopanda:
This PR : 
- fix the reception date for a stock transfer
- set a default description for a stock entry by a transfer

close #2081

### DIFF
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -331,7 +331,7 @@ function StockEntryController(
       from_depot : vm.movement.entity.instance.depot_uuid,
       to_depot : vm.depot.uuid,
       document_uuid : vm.movement.entity.instance.document_uuid,
-      date : vm.movement.entity.instance.date,
+      date : vm.movement.date,
       description : vm.movement.entity.instance.description,
       isExit : false,
       user_id : Session.user.id,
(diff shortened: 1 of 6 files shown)


server/controllers/stock/index.js
```diff
@@ -236,7 +236,7 @@ function depotMovement(document, params) {
 
     isWarehouse = !!(parameters.from_depot_is_warehouse);
 
-    // track distribution to patient
+    // track distribution to other depot from a warehouse
```
jniles, line 239: :dancer:

client/src/modules/stock/entry/entry.js
```diff
@@ -331,7 +333,7 @@ function StockEntryController(
       from_depot : vm.movement.entity.instance.depot_uuid,
       to_depot : vm.depot.uuid,
       document_uuid : vm.movement.entity.instance.document_uuid,
-      date : vm.movement.entity.instance.date,
+      date : vm.movement.date,
```
jniles, line 336: :dancer:

jniles: Are you sure this bug doesn't affect other modules (exit)?  Or other entry types?

jniles (approved):
Looks fine to me :+1:

PR #2114, opened 2017-09-21
Refact stock entry purchase

### LINKED ISSUE #2015: (refactor) Improvements to the Stock Entry from Purchase Order
Here are some ideas to improve the Stock Entry modal when importing stock from a purchase order:
 1. Use the headercrumb class to reduce space in the modal header
 2. Add "Importing Stock" or a similar text to the modal header to give context
 3. Make the inventory item inputs into [form-control-static tags](https://getbootstrap.com/docs/3.3/css/#forms-controls-static) instead.  Since the user will never be able to edit these, let's just show it like a static display.  This will also reduce space.
 4. Use `aggregationHideLabel: true` on the colDefs to remove the untranslated footer text.
 5. Make sure validation messages appear ... I tried to submit with the expiration dates as the current date and nothing happened. It was confusing.
 6. You are allowed to add floats as quantity ... not sure if this is a good thing or not.
 7. Turn off all validation until the submit() method is called.

See below a screenshot:
stockentryfrompurchaseordermodal
_Fig 1: Stock Entry From Purchase Order_

This is related to #2000 .

lomamech:
refactor(Improvements Stock Entry) 
- Improve modal header and give text
- Use form-control-static in non editable mode
- Appear all Validation messages
- Turn off all validation until the submit method called
- Add the validation for excessive quatity
- Remove the untranslate label by aggregationHideLabel

closes #2015


client/src/i18n/en/form.json
```diff
@@ -90,6 +91,7 @@
             "MISSING_DEBTOR_ID": "Identifying missing Debtor",
             "RECORD_ERROR": "Please resolve any errors on the form (marked in red).",
             "REQUIRED": "This value is required.",
+            "PLEASE_CHECK_EXPIRY_DATE": "Please check the expiry date",
```
jniles: This should say "Please check the expiration date."

client/src/modules/stock/entry/modals/lots.modal.html
```diff
@@ -4,17 +4,23 @@
...
 
   <div class="modal-header">
-    <div class="headercrumb">
-      <i class="fa fa-tags"></i> 
-      <h3 style="margin: 0px;">{{ $ctrl.inventory.inventory.label }}</h3>
-    </div>
+    <ol class="headercrumb">
+      <li class="static" translate>TREE.STOCK</li>
+      <li class="static" translate>STOCK.ENTRY</li>
+      <li class="title">{{ $ctrl.inventory.inventory.label }}</li>
```
jniles, line 10: :+1:

client/src/modules/stock/entry/modals/lots.modal.html
```diff
@@ -69,6 +79,12 @@ <h3 style="margin: 0px;">{{ $ctrl.inventory.inventory.label }}</h3>
       <i class="fa fa-warning"></i> 
       <span translate>FORM.INFO.MAX_LOT_REACHED</span>
     </p>
+
+    <p ng-if="$ctrl.submitError" class="alert alert-danger" style="margin: 5px 0px;">
```
jniles, line 83: :+1:

client/src/modules/stock/entry/modals/lots.modal.html
```diff
@@ -4,17 +4,23 @@
...
-      <label translate>FORM.LABELS.GLOBAL_QUANTITY</label>
+      <label class="control-label" translate>FORM.LABELS.GLOBAL_QUANTITY</label>
+      
+      <div ng-if="$ctrl.entryType === 'purchase' || $ctrl.entryType === 'transfer_reception'">
+        <span class="form-control-static">{{ $ctrl.inventory.quantity }}</span>
+      </div>
+
       <input 
-        ng-disabled="$ctrl.entryType === 'purchase' || $ctrl.entryType === 'transfer_reception'" 
+        ng-if="$ctrl.entryType !== 'purchase' && $ctrl.entryType !== 'transfer_reception'"
```
jniles, line 23: Instead of repeating this check multiple times, a better way to express this would be in JS.  Something like this:
```js
// determine if the inventory quantity and cost should be editable or not.
$ctrl.hasEditableInventory = ($ctrl.entryType !== 'purchase' && $ctrl.entryType !== 'transfer_reception');
```
... later in the HTML...
```html
<div ng-if="$ctrl.hasEditableInventory">
   <!--- put all of the editable code here -->
</div>
<div ng-if="!$ctrl.hasEditableInventory">
  <!-- put all form-static code here -->
</div>
```

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -34,13 +34,15 @@ function StockDefineLotsModalController(Instance, Notify, uiGridConstants, Data,
         displayName      : 'TABLE.COLUMNS.LOT',
         headerCellFilter : 'translate',
         aggregationType  : uiGridConstants.aggregationTypes.count,
+        aggregationHideLabel : true,
```
jniles, line 37: :dancing_men:

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -34,13 +34,15 @@ function StockDefineLotsModalController(Instance, Notify, uiGridConstants, Data,
...
         cellTemplate     : 'modules/stock/entry/modals/templates/lot.input.tmpl.html' },
 
       { field            : 'quantity',
         width            : 150,
         displayName      : 'TABLE.COLUMNS.QUANTITY',
         headerCellFilter : 'translate',
         aggregationType  : uiGridConstants.aggregationTypes.sum,
+        aggregationHideLabel : true,
```
jniles, line 45: :dancing_women:

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -124,6 +126,19 @@ function StockDefineLotsModalController(Instance, Notify, uiGridConstants, Data,
     var hasExpiration = (new Date(inventory.expiration_date) >= new Date());
     inventory.is_valid = (hasQuantity && hasLotLabel && hasExpiration);
 
+    // set the value if the date is not correct
+    vm.isExpiration = hasExpiration ? false : true;
+    vm.isExcessiveQuatity = hasQuantity ? false : true;
+
+    vm.submitError = vm.isExpiration;
+    if (vm.isExpiration) {
+      vm.errorText = 'FORM.ERRORS.PLEASE_CHECK_EXPIRY_DATE';
```
jniles: Since these errors are only related to Stock, it probably makes sense to put them in the stock en/fr JSON.  Something like `STOCK.ERRORS.PLEASE_CHECK_EXPIRY_DATE`.

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -124,6 +126,19 @@ function StockDefineLotsModalController(Instance, Notify, uiGridConstants, Data,
     var hasExpiration = (new Date(inventory.expiration_date) >= new Date());
     inventory.is_valid = (hasQuantity && hasLotLabel && hasExpiration);
 
+    // set the value if the date is not correct
+    vm.isExpiration = hasExpiration ? false : true;
```
jniles: Hmm... this binds a lot of variables to `vm` that you don't actually need.

How about your write a validation function for an inventory item (or lot, whichever makes sense)?  Something like this:
```js
function hasInvalidInventory(inventory) {
   // returns a string error like 'FORM.ERRORS.PLEASE_CHECK_EXPIRY_DATE' if there is an error.
  // otherwise, returns false.
}
```
You could then just call this function on every inventory item and it would return the appropriate message.
```js
var error = hasInvalidInventory(inventory);
if (error) {
  vm.submitErrror = true;
  vm.errorText = error;
}
```

This would reduce the logic needed to know the error type.  What do you think?

jniles (commented):
This is a step in the right direction.  The logic for error handling looks a bit scattered, I've suggested a function wrapper for it.  Let me know if you can find a way to clean this up.

jniles (approved):
This is a good step in the right direction! :+1:

bors r+

Issue #1769, opened 2017-06-07
refactor: standardize CSV export


jniles:
The current CSV export functionality supplied with BHIMA comes in two flavors: 
 1. Server-side CSV export, via changing the "renderer" flag to "csv".
 2. Client-side CSV export, via a ui-grid

There server-side method has some distinct advantages over the client-side at this point, which leads to developers favoring it - see https://github.com/IMA-WorldHealth/bhima-2.X/issues/1744#issuecomment-306726111 as an example discussion.   Here are those advantages:
 1. Because the switch is only an HTTP query parameter, developers can attach it to an `<a download>` tag easily, as long as they correctly serialize parameters.
 2. Because of (1), you can put it in a bootstrap dropdown menu without CSS complications, and without inflating your controller code.
 3. The CSV is identical in every way to the PDF produced with a renderer flag.  This make the UX nice - it feels like the same report, represented differently.  Because of the ui-grid exporter's options, it can feel like a different data set.

The clientside CSV exporter was developed to avoid expensive SQL queries re-running and has the distinct advantage of knowing how to $translate all columns, since it has access to the gridOptions object.

We should pick on way and standardize across modules.  @bhima-core, what are your opinions?

PR #2118, opened 2017-09-21
feat(journals export) Download Journals as Excel

### LINKED ISSUE #1348: (feature request) Download Journals as Excel
As pointed out by @patsonjunior, we cannot download the Posting (or Posted) Journal as Excel.  This would be helpful in finding transactions that have gone astray.  We should support downloading as Excel from those modules like we do the General Ledger.

jeremielodi:
This PR helps to export posting journal as excel in order to solve this issue #1348

Closes #1348.

### DIFF
--- a/client/src/modules/journal/journal.html
+++ b/client/src/modules/journal/journal.html
@@ -46,7 +46,7 @@
                 </a>
               </li>
               <li role="menuitem">
-                <a target='_blank' ng-href="/reports/finance/journal?{{JournalCtrl.downloadExcel()}}" translate-attr="{'download': 'INVOICE_REGISTRY.TITLE'}" >
+                <a target='_blank' ng-href="/reports/finance/journal?{{JournalCtrl.downloadExcel()}}" translate-attr="{'download': 'DOWNLOADS.EXCEL'}" >
                   <span class="fa fa-file-excel-o"></span> <span translate>DOWNLOADS.EXCEL</span>
                 </a>
               </li>
--- a/client/src/modules/journal/journal.html
+++ b/client/src/modules/journal/journal.html
@@ -46,6 +46,11 @@
                 </a>
               </li>
               <li role="menuitem">
+                <a target='_blank' ng-href="/reports/finance/journal?{{JournalCtrl.downloadExcel()}}" translate-attr="{'download': 'INVOICE_REGISTRY.TITLE'}" >
+                  <span class="fa fa-file-excel-o"></span> <span translate>DOWNLOADS.EXCEL</span>
+                </a>
+              </li>
+              <li role="menuitem">
                 <a href data-method="export" id="export" ng-click="JournalCtrl.exportFile()">
                   <i class="fa fa-file-excel-o"></i> <span translate>FORM.BUTTONS.CSV_EXPORT</span>
                 </a>
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -5,7 +5,7 @@ JournalController.$inject = [
   'JournalService', 'GridSortingService', 'GridGroupingService', 'GridFilteringService', 'GridColumnService',
   'SessionService', 'NotifyService', 'bhConstants', '$state', 'uiGridConstants', 'ModalService', 'LanguageService',
   'AppCache', 'Store', 'uiGridGroupingConstants', 'ExportService', '$filter', '$translate', 'GridExportService',
-  'GridStateService', 'GridSelectionService', 'TrialBalanceService',
+  'GridStateService', 'GridSelectionService', 'TrialBalanceService', '$httpParamSerializer',
 ];
 
 /**
@@ -31,7 +31,7 @@ JournalController.$inject = [
 function JournalController(
   Journal, Sorting, Grouping, Filtering, Columns, Session, Notify, bhConstants,
   $state, uiGridConstants, Modal, Languages, AppCache, Store, uiGridGroupingConstants,
-  Export, $filter, $translate, GridExport, GridState, GridSelection, TrialBalance
+  Export, $filter, $translate, GridExport, GridState, GridSelection, TrialBalance, $httpParamSerializer
 ) {
   var sorting;
   var grouping;
@@ -358,15 +358,22 @@ function JournalController(
   }
 
   // display the journal printable report of selected transactions
-  vm.openJournalReport = function openJournalReport() {
+  vm.openJournalReport = function openJournalReport(type) {
     var url = '/reports/finance/journal';
-    var params = formatExportParameters('pdf');
-
+    var params = formatExportParameters(type);
     if (!params) { return; }
-
-    Modal.openReports({ url : url, params : params });
+    Modal.openReports({ url: url, params: params });
   };
 
+  vm.downloadExcel = function () {
+    var filterOpts = Journal.filters.formatHTTP();
+    var defaultOpts = { renderer : 'xlsx', lang : Languages.key };
+
+    // combine options
+    var options = angular.merge(defaultOpts, filterOpts);
+    // return  serialized options
+    return $httpParamSerializer(options);
+  }
   // export data into csv file
   vm.exportFile = function exportFile() {
     exportation.run();
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -358,9 +358,9 @@ function JournalController(
   }
 
   // display the journal printable report of selected transactions
-  vm.openJournalReport = function openJournalReport(type) {
+  vm.openJournalReport = function openJournalReport() {
     var url = '/reports/finance/journal';
-    var params = formatExportParameters(type);
+    var params = formatExportParameters('pdf');
     if (!params) { return; }
     Modal.openReports({ url: url, params: params });
   };
--- a/server/controllers/finance/reports/journal/index.js
+++ b/server/controllers/finance/reports/journal/index.js
@@ -10,6 +10,7 @@ const _ = require('lodash');
 const ReportManager = require('../../../../lib/ReportManager');
 const db = require('../../../../lib/db');
 const Journal = require('../../journal');
+const util = require('../../../../lib/util');
 
 const REPORT_TEMPLATE = './server/controllers/finance/reports/journal/report.handlebars';
 
@@ -57,7 +58,13 @@ function postingJournalExport(req, res, next) {
       return report.render({ rows, totals });
     })
     .then((result) => {
-      res.set(result.headers).send(result.report);
+
+      if (result.headers.type === 'xlsx') {
+        res.xls(result.headers.filename, util.dateFormatter(result.report.rows));
+      } else {
+        res.set(result.headers).send(result.report);
+      }
+
     })
     .catch(next)
     .done();
--- a/server/lib/util.js
+++ b/server/lib/util.js
@@ -10,11 +10,12 @@
 
 const _ = require('lodash');
 const q = require('q');
+const moment = require('moment');
 
 /** The query string conditions builder */
 module.exports.take = take;
 module.exports.loadModuleIfExists = requireModuleIfExists;
-
+exports.dateFormatter = dateFormatter;
 exports.resolveObject = resolveObject;
 
 /**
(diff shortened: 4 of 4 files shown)


jniles:
Let's wait until https://github.com/IMA-WorldHealth/bhima-2.X/pull/2085 is reviewed and merged for this one.  It will be easier to merge that one first.

mbayopanda (approved):
LGTM
The exported file can have a date column in this format `YYYY-MM-DD` instead of `Fri Sep 22 2017 10:01:45 GMT+0100 (W. Central Africa Standard Time)`

server/controllers/finance/reports/journal/index.js
```diff
@@ -42,12 +43,40 @@ function postingJournalExport(req, res, next) {
...
+        res.xls(result.headers.filename, filter(result.report.rows));
+      } else {
+        res.set(result.headers).send(result.report);
+      }
     })
     .catch(next)
     .done();
 }
 
+function dateFormat(date) {
```
jniles: The [csv module already has this feature](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/renderers/csv.js#L19), with a comment from @sfount.

```js
// @TODO Discuss if this should be moved into its own library
const DATE_FORMAT = 'DD/MM/YYYY H:mm:s';
```

Since you are using the same date format here, I propose you make a new library for dealing with formats like dates and include it in both the CSV renderer and the XLSX renderer.  Otherwise, this is a duplication of code (and not very [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)).

jniles (commented):
@jeremielodi, this is a good start.  It will need to be updated to `master` to get rid of these conflicts.  I've made a suggestion for encapsulating the date format code and sharing it between server-side libraries.

jniles:
Looks like you have a broken merge somewhere that is breaking everything.  See if you can diagnose it.  [Test results](https://travis-ci.org/IMA-WorldHealth/bhima-2.X/jobs/280374501).

client/src/modules/journal/journal.js
```diff
@@ -358,13 +362,17 @@ function JournalController(
...
+  vm.openJournalReport = function openJournalReport(type) {
     var url = '/reports/finance/journal';
-    var params = formatExportParameters('pdf');
+    var params = formatExportParameters(type);
 
     if (!params) { return; }
 
-    Modal.openReports({ url : url, params : params });
+    if (type === 'xlsx') {
+      window.location = url + '?'+ $.param(params);
```
jniles: Hmm... this looks like a hack.  Why do you have to do this?  Can't you do the same thing with an `ng-href` and never call `openJournalReport()`?

server/controllers/finance/reports/journal/index.js
```diff
@@ -57,7 +107,12 @@ function postingJournalExport(req, res, next) {
       return report.render({ rows, totals });
     })
     .then((result) => {
-      res.set(result.headers).send(result.report);
+      if(result.headers.type ==='xlsx'){
+        res.xls(result.headers.filename, result.report.rows);
+      }else
```
jniles: This could is really sloppy.  Please respect ESLint.

server/controllers/finance/reports/journal/index.js
```diff
@@ -32,15 +33,64 @@ function postingJournalExport(req, res, next) {
...
+    .catch(next)
+    .done();
+}
+
+function dateFormat(date) {
+
+  if (date) {
+    const dateTime = new Date(date);
+    return  moment(dateTime).format('YYYY-MM-DD HH:mm:ss');
+  } else {
```
jniles: This function could be much simpler:
```js
const DATE_FORMAT = 'YYYY-MM-DD HH:mm:ss';
function dateFormat(date) {
  return date ? moment(date).format(DATE_FORMAT) : date;
}
```
If you wanted to be _really_ fancy, you could use arrow functions, but I don't prefer them here, personally.

server/controllers/finance/reports/journal/index.js
```diff
@@ -32,15 +33,64 @@ function postingJournalExport(req, res, next) {
...
+        res.xls(result.headers.filename, filter(result.report.rows));
+      } else {
+        res.set(result.headers).send(result.report);
+      }
+    })
+    .catch(next)
+    .done();
+}
+
+function dateFormat(date) {
```
jniles: Can you please combine this function with the date format function in the CSV renderer?  It will make it much easier to maintain.  This was mentioned in a previous review.

client/src/modules/journal/journal.js
```diff
@@ -121,6 +121,8 @@ function JournalController(
 
   vm.onRemoveFilter = onRemoveFilter;
 
+  vm.cancelEdit = cancelEdit;
```
jniles: Where did this code come from?

client/src/modules/journal/journal.html
```diff
@@ -41,10 +41,17 @@
...
               </li>
               <li role="menuitem">
-                <a href data-method="print" id="print" ng-click="JournalCtrl.openJournalReport()">
+                <a href data-method="print" id="print" ng-click="JournalCtrl.openJournalReport('pdf')">
                   <i class="fa fa-print"></i> <span translate>FORM.BUTTONS.PRINT</span>
                 </a>
               </li>
+
+            <li role="menuitem">
+              <a href data-method="print" ng-click="JournalCtrl.openJournalReport('xlsx')">
```
jniles: You should remove `data-method="print"` here.

server/controllers/finance/reports/journal/index.js
```diff
@@ -32,15 +33,64 @@ function postingJournalExport(req, res, next) {
...
+
+  if (date) {
+    const dateTime = new Date(date);
+    return  moment(dateTime).format('YYYY-MM-DD HH:mm:ss');
+  } else {
+    return null;
+  }
+}
+
+function filter(rows) {
```
jniles: You normally expect a filter function to filter things out of an array.  All this does is map.  A better name for the function is therefore something like `convertDatesToStrings` or something descriptive of what this does.

Also, I think ESLint will get angry about all these empty lines.   Go head and clean this up according to ESLint's rules.

jniles (changes requested):
@jeremielodi, this is really sloppy code.  Please take your time and make sure that the code passes ESLint and is nicely formatted.  I think you also have some broken merge artifacts in here - those need to be cleaned up.

client/src/modules/journal/journal.js
jeremielodi: @jniles I tried to see how you export the journal as  CSV, the operation is done at the client side as can see right?

jeremielodi: I have a problem, I should know this value 
 var params = formatExportParameters(type);
before sending the request to the server, do you think it is  also possible  even if I use ng-href?

jniles: Not in all places - that's why ~~@sfount~~ I opened #1769.  [Here is an example of the ng-href version](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/invoices/registry/registry.html#L43).

jeremielodi: ok, let me try it out

server/controllers/finance/reports/journal/index.js
```diff
@@ -32,15 +33,64 @@ function postingJournalExport(req, res, next) {
...
-
   try {
     report = new ReportManager(REPORT_TEMPLATE, req.session, options);
   } catch (e) {
     return next(e);
   }
 
-  const unposted = Journal.buildTransactionQuery(options, false);
-  const posted = Journal.buildTransactionQuery(options, true);
+  return Journal.journalEntryList(options)
```
jeremielodi: I don't Know if some think has been changed here, this function Journal.journalEntryList(options) is not recognised

jeremielodi: Does it work for you. you can try to export the journal as pdf and see if it is okay

jniles: It was removed a couple days ago in https://github.com/IMA-WorldHealth/bhima-2.X/commit/67c3b7e17cc01696980fd818d8a1b67069a0a0b3#diff-e98d808c974246550596644bed19f516.

jeremielodi:
I create just a new branch with the same name

jeremielodi:
This part works  without date filters

jniles:
By the way, it looks like we might be able to get rid of the `middleware` part to make the API look the same.  [It is only a couple lines](https://github.com/rikkertkoppes/json2xls/blob/master/lib/json2xls.js#L89).

jeremielodi:
Date columns are formatted now
closes #1348

jniles (approved):
LGTM!

jniles:
This code is much improved!  Well done.

As noted offline, there is a tiny bug where, if no rows are selected, everything is exported.  This will crash lower-end hardware.  It should be fixed in a subsequent PR.

bors r+

client/src/modules/journal/journal.html
```diff
@@ -46,6 +46,11 @@
                 </a>
               </li>
               <li role="menuitem">
+                <a target='_blank' ng-href="/reports/finance/journal?{{JournalCtrl.downloadExcel()}}" translate-attr="{'download': 'INVOICE_REGISTRY.TITLE'}" >
```
jniles: Why does this say "Invoice Registry"?

client/src/modules/journal/journal.js
```diff
@@ -358,15 +358,22 @@ function JournalController(
   }
 
   // display the journal printable report of selected transactions
-  vm.openJournalReport = function openJournalReport() {
+  vm.openJournalReport = function openJournalReport(type) {
```
jniles: You changed this function to require a type, but in the dropdown menu, you never pass it a type - see Line 44 of the journal.html file.  That may be what is breaking your tests.

server/lib/util.js
```diff
@@ -95,3 +96,23 @@ function resolveObject(object) {
       return settled;
     });
 }
+
+/**
+ * @method dateFormatter
```
jniles, line 101: Nice comments!

server/lib/util.js
```diff
@@ -95,3 +96,23 @@ function resolveObject(object) {
...
+ *
+ * @description
+ * Accepts an object of key/value pairs. Returns the same object with all values
+ * that are dates converted to a standard format.
+ */
+function dateFormatter(Rows, dateFormat) {
```
jniles: It is generally good practice to only use capital letters with constructors - functions that you will call `new` on to invoke a new instance.  We occasionally break this rule, but it isn't necessary to break it here.  `Rows` is just a flat array.

server/lib/util.js
```diff
@@ -95,3 +96,23 @@ function resolveObject(object) {
...
+  const DATE_FORMAT = dateFormat || 'YYYY-MM-DD HH:mm:ss';
+  return Rows.map(function (element) {
+    Object.keys(element).forEach(function (currentKey) {
```
jniles: A cleaner way to do this would be with [lodash](https://lodash.com/docs/4.17.4#forEach).  It allows you to use functional programming on objects.  For example:
```js
function dateFormatter(rows, dateFormat) {
  return rows.map((element) => {
    return _.forEach(element, (value, key) => {
      if (_.isDate(key)) {
         element[key] = moment(value).format(DATE_FORMAT);
      }
    });
  });
}
```

jniles (commented):
I'm not entirely sure why tests are breaking, but there are a few mistakes here that I found on a second debug review.  That might help fix them.

jniles:
@jeremielodi, can you look into these issues?

client/src/modules/journal/journal.html
jniles: This shouldn't say invoice registry.  This is not the invoice registry.

client/src/modules/journal/journal.js
jniles: This function shouldn't take any parameters.  It isn't called with any parameters.

server/lib/util.js
jniles: This shouldn't be capitalized.

jniles (changes requested):

jniles (approved):
LGTM.

bors r+

PR #2126, opened 2017-09-22
chore(bhBreadcrumb): remove component

### LINKED ISSUE #1923: proposal: deprecate bhBreadcrumb
I propose that we deprecate `bhBreadcrumb`.  What bhBreadcrumb mostly does is allow developers to write JS instead of HTML.  This is demonstrated below:
```js
// using bhBreadcrumb, the header looks like this:
  vm.paths = [{                                                                                                                      
    label : $translate.instant('TREE.FINANCE'),                                                                                      
  }, {                                                                                                                                                                                                                                                                    
    label   : $translate.instant('VOUCHERS.COMPLEX.TITLE'), 
    current : true, 
  }]; 
```
```html
<!-- instead of like this -->
     <ol class="headercrumb">                                                                                                                                                                                                                                              
      <li class="static" translate>TREE.FINANCE</li>                                                                                
      <li class="title" translate>VOUCHERS.COMPLEX.TITLE</li>
    </ol>
```

There are a few things to notice here:
 1. The `bhBreadcrumb` version is actually two lines longer.
 2. The `bhBreadcrumb` version requires an additional import of `$translate` ... even if the developer does not use it elsewhere in the controller.
 3. The `bhBreadcrumb` doesn't add anything special - it simply moves the configuration from HTML to JS.

These reasons apply to the other components within `bhBreadcrumb` as well - buttons, the dropdown, etc.  Ultimately, it was a good idea at one time, but it no longer makes sense as we are trying to customize our dropdown menus and title labels.

/cc @mbayopanda would you be okay with this proposal?

jniles:
This commit removes the last vestiges of bhBreadcrumb.  The component
introduced more complexity than it solved and was much less efficient
than desirable.

Closes #1923.

----

Thank you for contributing!

 - [x] Run integration tests.
 - [x] Run end-to-end tests.


mbayopanda (approved):
LGTM

mbayopanda:
@jniles could you rebase this branch master and fix conflict ?

PR #2127, opened 2017-09-22
fix(login): remove untranslated Notify.warn()

jniles:
This commit removes an annoying `Notify.warn()` call that happens when the user is trying to access a substate of the application, but is not logged in.  Because it is one of the first things loaded, it can never
be translated correctly, making the application look broken.


DedrickEnc (approved):
LGTM
As soon as the test pass it could be merged

PR #2134, opened 2017-09-25
perf(link Inventory Invoice) 

### LINKED ISSUE #2128: Link Invoices made with an Inventory Item from Inventory List
One of the the most common questions our clients have is "How much money has X service/item made?"  One step towards that analysis is giving them the number of times that item has been used in an invoice.  Since we already have the capability of linking filters from one registry to another, I propose that we implement a link from the Inventory list (`/inventory/list`) to the Invoice Registry (`/invoices`).  That way, a user can keep track of the invoices that involve a particular item if they so desire.

lomamech:
perf(link Inventory Invoice)  …
- Complete translate key
- Add option in Inventory Registry for lookup invoice registry par
  Inventory item
- Add filter Option for inventory text in API for patient invoices

test(link inventory to Invoice)
- Add test E2E to verificate link between inventory to Invoice Registry

closes #2128


client/src/i18n/fr/vouchers.json
```diff
@@ -29,6 +29,7 @@
         "GLOBAL": {
             "CASHFLOW_OPTION": "Option pour cashflow",
             "CONVENTION_INVOICES": "Convention - Paiement factures",
+            "CONVENTION_PAYMENT_DESCRIPTION": "Paiement des Conventions: paiement de {{numInvoices}} facture(s) par {{debtorGroupName}}.",
```
jniles, line 32: :+1:

server/controllers/finance/patientInvoice.js
```diff
@@ -231,6 +231,11 @@ function find(options) {
     'invoice.uuid IN (SELECT cash_item.invoice_uuid FROM cash_item WHERE cash_item.cash_uuid = ?)'
   );
 
+  filters.custom(
```
jniles, line 234: Hmm... it is probably better to do this filter on the `uuid` itself.  That way it will be super fast (the `uuid` column is INDEXed).

Something like this would be better:
```js
filters.custom('inventory_uuid', 'invoice.uuid IN (SELECT invoice_item.invoice_uuid FROM invoice_item WHERE invoice_item.inventory_uuid = ?');
```

client/src/modules/inventory/list/templates/action.cell.html
```diff
@@ -14,5 +14,13 @@
           <i class="fa fa-edit"></i> <span translate>FORM.BUTTONS.EDIT</span>
       </a>
     </li>
+
+    <li class="divider"></li>
+    <li>
+      <a data-method="invoice" ui-sref="invoiceRegistry({ filters : { key : 'inventoryLabel', value : row.entity.label }})" href>
```
jniles: If you use the `uuid`, the query will be more efficient.  To show a nice label instead of a giant UUID, try passing in a display value like this:
```html
ui-sref="invoiceRegistry({ filters : { key : 'inventory_uuid', value : row.entity.uuid, displayValue : row.entity.label }})"
```
Then, in the invoice registry, you can bind the `displayValue` to the filter and it should filter correctly.

test/end-to-end/verificationLinks/verificationLinks.spec.js
```diff
@@ -58,5 +58,14 @@ describe('Check links', () => {
     GU.expectRowCount('payment-registry', 0);
   });
 
+  it('Check link betwen Inventory Registry -> Invoice Registry', () => {
```
jniles: :+1:  This should say "between" instead of "betwen"

jniles (changes requested):
@lomamech, this is a really good start!

I'm concerned about the performance of this query.  Could you migrate this to using the `inventory_uuid`?  I've suggested how to do this.  If you need help making the label appear, you can ask me.

jniles (approved):
bors r+

This works really well!  Well done!

PR #2135, opened 2017-09-25
refractor(ui-grid Action link) add or display a caret

### LINKED ISSUE #2123: refactor: make sure all "actions" links have a caret
We have two types behavior for links in BHIMA.
 1. Navigate the page to some other location (such as filtering a registry on a value or opening a modal/tab with a receipt).
 2. Open a drop down menu to perform an action.

Most of the (2) links are labeled "actions", but not all of them have a caret ( :arrow_down_small: ) that informs the user what will happen after they click it.  Two examples are shown below:

niceactionslinks
_Fig 1: Nice Actions Links with Carets_

badactionslinks
_Fig 2: Bad Actions Links without Carets_

It is a small change, but very helpful for usability.  Someone should go through the entire application and make sure that every dropdown has a `<span class="caret"></span>` on it.

jeremielodi:
This PR helps to display caret for each action link in the ui-grid as recommanded in this issue #2123

Fixes #2123.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,12 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.7.1
+ignore: {}
+# patches apply the minimum changes required to fix a vulnerability
+patch:
+  'npm:debug:20170905':
+    - compression > debug:
+        patched: '2017-09-27T03:37:07.471Z'
+    - helmet > connect > debug:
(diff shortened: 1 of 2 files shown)


jniles:
@jeremielodi, before this is merged, can you update `billing_services` as well?

Also, this needs a "Fixes #" comment to make sure it closes the original issue.

jniles (approved):
bors r+

PR #2136, opened 2017-09-25
Fix: Inventory group item links

### LINKED ISSUE #2122: bug: Inventory Groups Configuration no longer filters inventory list
The inventory groups configuration page shows a link in the "contains" column, allowing a user to click on the link and filter the Inventory List by this value.  Unfortunately, this link is broken and the Inventory list is no longer filtered.

**Expected Behavior**:
A user will be navigated to `/#!/inventory` and be able to see all the items filter on that group.

**Actual Result**
A user is navigated to the inventory list, but the list remains unfiltered.

See below:
buginventorycontainslinkbroken
_Fig 1: Bug in Inventory Contains Link_

jniles:
This commit completely refreshes the inventory list fixing many small bugs.  These bugs include, but are not limited to:
 1. Filters are now cached and loaded correctly. Filters can be linked  from $states.
 2. There are no more column menus.
 3. The filter now uses FilterParser on the server side.
 4. Unit Prices are rendered in the enterprise currency.

This commit fixes the inventory registry by linking the inventory groups to the inventory registry.

Closes #2122.

### DIFF
--- a/client/src/modules/inventory/list/modals/search.modal.html
+++ b/client/src/modules/inventory/list/modals/search.modal.html
@@ -1,7 +1,4 @@
-<form
-  name="ModalForm"
-  ng-submit="ModalCtrl.submit(ModalForm)"
-  novalidate>
+<form name="ModalForm" ng-submit="ModalCtrl.submit(ModalForm)" novalidate>
 
   <div class="modal-header">
     <ol class="headercrumb">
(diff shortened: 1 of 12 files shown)


client/src/modules/inventory/list/modals/search.modal.html
```diff
@@ -10,106 +7,109 @@
...
+        <div class="tab-body">
+
+          <div class="form-group">
+            <label class="control-label" translate>
+              FORM.LABELS.GROUP
+            </label>
+            <bh-clear on-clear="ModalCtrl.clear('group_uuid')"></bh-clear>
+
+            <!-- @todo make data driven selects components to handle loading and error states -->
+            <ui-select
```
DedrickEnc, line 22: We should create a component to manage inventory group separately, so that the controller will not care about inventory group data downloading for focusing on scope model only.
May be we have it already.
If not possible, we can create a separate issue.

client/src/modules/inventory/list/modals/search.modal.html
```diff
@@ -10,106 +7,109 @@
...
+              ng-model="ModalCtrl.searchQueries.group_uuid">
+              <ui-select-match placeholder="{{ 'FORM.SELECT.INVENTORY_GROUP' | translate }}"><span>{{$select.selected.name}}</span></ui-select-match>
+              <ui-select-choices ui-select-focus-patch repeat="group.uuid as group in (ModalCtrl.inventoryGroups | filter:$select.search | orderBy:'name')">
+                <span ng-bind-html="group.name | highlight:$select.search"></span>
+              </ui-select-choices>
+            </ui-select>
+
+          </div>
+
+          <div class="form-group" ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.text.$invalid }">
```
DedrickEnc, line 33: Is it possible to use the inventory select here? so that we pass a ```uuid``` instead of text

server/controllers/inventory/inventory/core.js
```diff
@@ -110,8 +110,8 @@ function getIds() {
 * @return {Promise} Returns a database query promise
 */
 function getItemsMetadata(params) {
-  db.convert(params, ['inventory_uuids']);
-  const filters = new FilterParser(params, { tableAlias : 'inventory' });
+  db.convert(params, ['inventory_uuids', 'uuid', 'group_uuid']);
+  const filters = new FilterParser(params, { tableAlias : 'inventory', autoParseStatements : false });
```
DedrickEnc, line 114: 👍

DedrickEnc (commented):
This P.R can be merged any time, I have just asked some questions to see if a change can happen in the current pull request or in a separate issue. 👍 
LGTM

DedrickEnc:
This branch needs a rebase

client/src/modules/inventory/list/modals/search.modal.html
jniles: Good idea.  I don't think we have it yet.

client/src/modules/inventory/list/modals/search.modal.html
jniles: I think this translates to a `fullText` on the server - the idea being that we can find everything that matches "computer" or "hi" instead of selecting a particular inventory item.

jniles:
@DedrickEnc  rebased.

Issue #2144, opened 2017-09-26
Link Patient Financial Activity in the Patient Registry dropdown


jniles:
According to the Vanga MD, the Patient Financial Activity report is one of the most used reports in the system.  He proposes that we link the Patient Financial Activity from the Patient Registry.  This will save them a few clicks and lots of loading.

The link should open a PDF in a separate tab, similar to what happens in the present Journal.

jniles:
Closed in https://github.com/IMA-WorldHealth/bhima-2.X/pull/2151.

PR #2151, opened 2017-09-26
feat(patient register) Link Patient Financial Activity

jeremielodi:
This PR links a Patient Financial Activity in the Patient Registry dropdown  as described in this issue #2144

### DIFF
--- a/client/src/modules/patients/templates/action.cell.html
+++ b/client/src/modules/patients/templates/action.cell.html
@@ -23,6 +23,12 @@
     </li>
     <li class="divider"></li>
     <li>
+      <a data-method="financialActivities" ng-href="reports/finance/financialPatient/{{row.entity.uuid}}" target="_blank">
+        <span class="fa fa-file-text-o"></span> <span translate>FORM.LABELS.VIEW_FINANCIAL_ACT</span>
+      </a>
+    </li>
+    <li class="divider"></li>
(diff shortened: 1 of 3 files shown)


client/src/modules/patients/templates/action.cell.html
```diff
@@ -23,6 +23,11 @@
     </li>
     <li class="divider"></li>
     <li>
+      <a data-method="financialActivities" ng-href="reports/finance/financialPatient/{{row.entity.uuid}}" target="_blank">
+        <span class="fa fa-file-text-o"></span> <span translate>FORM.BUTTONS.PRINT_FINANCIAL_ACT</span>
```
jniles: Since this opens a PDF report, let's make the icon a PDF file.  Also, since this isn't opening a print modal, a better name would be "View Financial Activity" instead of "Print".

jniles (commented):
@jeremielodi this works well.

Since its behavior is different from the other links right below it, I propose that you move it to the bottom of the list and put a `<li class="divider"></li>` between it and the other links.

jniles (approved):
bors r+

jeremielodi:
closed #2144

jniles:
Unfortunately, you have to comment "closes #2144" before it gets merged.  You'll have to manually close that issue now.

PR #2154, opened 2017-09-26
docs: remove PR template

jniles:
We have robots and automated testing suites to enforce these policies for us.  No need to keep this checklist around.  It has fallen into disuse a while ago.


DedrickEnc:
You are right

DedrickEnc (approved):
LGTM

Issue #2100, opened 2017-09-15
chore: migrate `modules/` from $translate filter to $translate directive


jniles:
This issue has been discussed many times before, however, we are nearing the end of this task.  A quick scan of the `modules/` directory shows:
```sh
$ grep -r "| translate" | wc -l
380
```

This means we only have 380 lines left to change in the modules directory!  Someone could easily go through and clean all these references up.

jeremielodi:
closed in #2155

Issue #2149, opened 2017-09-26
Disabled Editing Account Types after Account Creation


jniles:
Our account management module lets a user edit and change the account type whenever they desire.  From an accounting standpoint, this does not make any sense.  If an account has been operating as income and expense for several months, it cannot be suddenly changed to a balance account.  We should remove the ability to edit the account type (with `form-static` class) from the account management page.

DedrickEnc:
@jniles
Are you sure this is a good solution? So what will happen if an account was registered with mistake and the account type must be changed?
These kind of mistakes are frequent at the field.

jniles:
It would be much better for them to lock and forget about the account than try to change it.  If you make a debtor account into an expense account, and then later into an asset account, it will completely confuse the reports at the end of the year.

PR #2155, opened 2017-09-26
refractor(translate directives) replace translate filters by directives

jeremielodi:
This PR replace all  translate filters by translate directives

### DIFF
--- a/client/src/i18n/en/posting_journal.json
+++ b/client/src/i18n/en/posting_journal.json
@@ -5,6 +5,7 @@
     "ROWS" : "Row(s)",
     "EDITING" : "Editing",
     "NO_TRANSACTION_TYPE" : "No assigned transaction type",
+    "HAS_PREVIOUS_EDITS" : "Last edit by {{ user }} on {{ date }}",
     "JUSTIFICATION" : "Justification",
     "REFRESH" : "Refresh",
     "SAVE_TRANSACTION" : "Save Transaction",
--- a/client/src/i18n/fr/posting_journal.json
+++ b/client/src/i18n/fr/posting_journal.json
@@ -4,6 +4,7 @@
     "REMOVE_ROW" : "Supprimer Ligne(s)",
     "ROWS" : "Ligne(s)",
     "SAVE_TRANSACTION_SUCCESS" : "La transaction etait enregistré avec succés",
+    "HAS_PREVIOUS_EDITS" : "Dernière édition par {{ user }} le {{ date }}",
     "JUSTIFICATION" : "Justification",
     "REFRESH" : "Rafraichir",
     "SAVE_TRANSACTION" : "Sauvegarder la transaction",
--- a/client/src/modules/journal/journal.service.js
+++ b/client/src/modules/journal/journal.service.js
@@ -19,6 +19,7 @@ function JournalService(Api, AppCache, Filters, Periods, Modal) {
   service.openTransactionEditModal = openTransactionEditModal;
   service.mapTransactionIdsToRecordUuids = mapTransactionIdsToRecordUuids;
   service.commentPostingJournal = commentPostingJournal;
+  service.getTransactionEditHistory = getTransactionEditHistory;
 
   /**
    * Standard API read method, as this will be used to drive the journal grids
@@ -204,7 +205,14 @@ function JournalService(Api, AppCache, Filters, Periods, Modal) {
 
   // updating the posting journal by adding comments in transactions
   function commentPostingJournal(params) {
-    return service.$http.put(URL.concat('comments'), { 'params' : params })
+    return service.$http.put(URL.concat('comments'), { params : params })
+      .then(service.util.unwrapHttpResponse);
+  }
+
+  // load the edit history of a particular transaction
+  function getTransactionEditHistory(uuid) {
+    var url = '/transactions/:uuid/history'.replace(':uuid', uuid);
+    return service.$http.get(url)
       .then(service.util.unwrapHttpResponse);
   }
 
--- a/client/src/modules/journal/modals/editTransaction.modal.html
+++ b/client/src/modules/journal/modals/editTransaction.modal.html
@@ -1,11 +1,12 @@
 <div class="modal-header">
-  <ol class="headercrumb">
+  <ol class="headercrumb clearfix">
     <li class="static" translate>POSTING_JOURNAL.TRANSACTION</li>
     <li class="title" translate>{{ModalCtrl.shared.trans_id}}</li>
     <span class="badge badge-warning" ng-if="ModalCtrl.readOnly"><i class="fa fa-eye"></i> <span translate>POSTING_JOURNAL.VIEWING</span></span>
     <span class="badge badge-primary" ng-if="!ModalCtrl.readOnly"><i class="fa fa-pencil"></i> <span translate>POSTING_JOURNAL.EDITING</span></span>
   </ol>
 </div>
+
 <div class="modal-body">
   <!-- warn user that standard users cannot edit posted transactions -->
   <div class="alert alert-warning pad-alert" ng-if="ModalCtrl.validation.blockedPostedTransactionEdit">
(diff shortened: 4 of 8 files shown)


client/src/modules/accounts/accounts.html
```diff
@@ -13,20 +13,20 @@
           </a>
           <ul uib-dropdown-menu role="menu" class="dropdown-menu-right">
             <li role="menuitem">
-              <a ng-href="/reports/finance/accounts/chart/?{{ AccountsCtrl.Accounts.download('pdf') }}" download="{{ 'ACCOUNT.TITLE' | translate }}">
+              <a ng-href="/reports/finance/accounts/chart/?{{ AccountsCtrl.Accounts.download('pdf') }}" translate-attr="{ download: 'ACCOUNT.TITLE' }">
```
jniles: :+1:

client/src/modules/accounts/edit/accounts.edit.modal.html
```diff
@@ -50,7 +50,7 @@
     <!-- @TODO Re-disable account type editing after accounts are stable -->
     <!-- Account type read only -->
     <!-- <div ng-if="!AccountEditCtrl.isCreateState" class="form-group"> -->
-      <!-- <label class="control-label">{{ "FORM.LABELS.ACCOUNT_TYPE" | translate }}</label> -->
+      <!-- <label class="control-label">{{ "FORM.LABELS.ACCOUNT_TYPE"  }}</label> -->
```
jniles: Eventually we should re-enable this: #2149.  Glad to see the code is still here!

client/src/modules/cash/modals/invoice-modal.html
```diff
@@ -30,7 +30,7 @@
   </div>
 
   <div ng-show="CashInvoiceModalCtrl.empty" class="text-warning">
-    <i class="fa fa-info-circle"></i> {{ ::"FORM.WARNINGS.EMPTY_SELECTION" | translate }}
+    <i class="fa fa-info-circle"></i> <span translate> {{ ::"FORM.WARNINGS.EMPTY_SELECTION" }} </span>
```
jniles: You can remove the `{{ :: }}`.  The two colons `::` was trying to make the binding more efficient - they are angular's one-time binding.  But since we are moving to the directive, no need for binding at all.

client/src/modules/cash/modals/invoice-modal.html
```diff
@@ -14,7 +14,7 @@
 
   <!-- warn the client if not debtor id has been passed in -->
   <p ng-show="CashInvoiceModalCtrl.missingId" class="text-danger text-center">
-    <i class="fa fa-warning"></i> {{ ::"FORM.ERRORS.MISSING_DEBTOR_ID" | translate }}
+    <i class="fa fa-warning"></i> <span translate>{{ ::"FORM.ERRORS.MISSING_DEBTOR_ID" }}</span>
```
jniles: This also does not need the `::`.  It can just be `<span translate>FORM.ERRORS.MISSING_DEBTOR_ID</span>`.

client/src/modules/creditors/groups/creditor-groups-list.html
```diff
@@ -19,7 +19,7 @@ <h5 style="margin-top : 0px;"></span><span translate> FORM.LABELS.ACCOUNT_NUMBER
               </div>
             </div>
 
-            <span class="fa fa-bars"></span> <b>{{ cg.total_creditors }}</b> {{ 'FORM.LABELS.CREDITOR_SUBCRIBED' | translate }}
+            <span class="fa fa-bars"></span> <b>{{ cg.total_creditors }}</b> <span transalet>FORM.LABELS.CREDITOR_SUBCRIBED</span>
```
jniles: It looks like there is a typo here - this should be `translate` not `transalet`.

client/src/modules/debtors/groups.list.html
```diff
@@ -23,7 +23,7 @@
           </div>
         </div>
 
-        <input ng-model="GroupCtrl.filter" ng-show="GroupCtrl.filterActive" class="form-control" placeholder="{{'FORM.PLACEHOLDERS.FILTER_NAME' | translate }}" style="border-radius: 0"/>
+        <input ng-model="GroupCtrl.filter" ng-show="GroupCtrl.filterActive" class="form-control" translate-attr="{ alt: 'FORM.PLACEHOLDERS.FILTER_NAME' }" style="border-radius: 0"/>
```
jniles: This attribute should be `placeholder`, not `alt`.

client/src/modules/debtors/groups.update.html
```diff
@@ -129,50 +130,50 @@
...
+                <h5><b translate>DEBTOR_GROUP.POLICIES.DISCOUNTS.TITLE</b></h5>
                 <p ng-if="GroupUpdateCtrl.group.apply_discounts===false" class="text-info">
-                  <span class="fa fa-info-circle" aria-hidden="true"></span> {{"DEBTOR_GROUP.POLICIES.DISCOUNTS.INFO" | translate }}
+                  <span class="fa fa-info-circle" aria-hidden="true"></span><span translate>DEBTOR_GROUP.POLICIES.DISCOUNTS.INFO</span>
                 </p>
                 <div class="checkbox">
                   <label>
                     <input ng-model="GroupUpdateCtrl.group.apply_discounts" name="apply_discounts" type="checkbox">
-                    {{"DEBTOR_GROUP.POLICIES.DISCOUNTS.LABEL" | translate }}
+                    <span translate>{{"DEBTOR_GROUP.POLICIES.DISCOUNTS.LABEL</span>
```
jniles: This is broken - it should be `<span translate>DEBTOR_GROUP.POLICIES.DISCOUNTS.LABEL</span>`.

client/src/modules/debtors/subscriptions.modal.html
```diff
@@ -1,14 +1,14 @@
...
-    <li class="title">{{ "DEBTOR_GROUP.SUBSCRIPTIONS" | translate }}</li>
+    <li class="title" translate>DEBTOR_GROUP.SUBSCRIPTIONS</li>
   </div>
 </div>
 
 <div class="modal-body">
   <form name="subscriptionForm">
     <div class="form-group">
-      <label class="control-label">{{ SubCtrl.entityKey | translate }}</label>
+      <label class="control-label" translate>{{ SubCtrl.entityKey}}</label>
```
jniles: :+1:

client/src/modules/fiscal/fiscal.manage.html
```diff
@@ -41,7 +41,7 @@
             <select class="form-control"
               ng-model="FiscalManageCtrl.fiscal.previous_fiscal_year_id"
               ng-options="fy.id as fy.hrLabel for fy in FiscalManageCtrl.previous_fiscal_year">
-              <option value="" disabled>--{{ 'FORM.SELECT.FISCAL_YEAR' | translate }}--</option>
+              <option value="" disabled>--<span translate>FORM.SELECT.FISCAL_YEAR</span>--</option>
```
jniles: Unfortunately, this will not work.  The HTML `<option>` element doesn't let you render HTML within it (that is part of the reason libraries like `ui-select` exist).  The correct way to implement this will be:
```html
 <option value="" disabled translate>FORM.SELECT.FISCAL_YEAR</option>
```

client/src/modules/general-ledger/templates/action.cell.html
```diff
@@ -7,18 +7,18 @@
 
     <ul class="dropdown-menu-right" uib-dropdown-menu>
       <li>
-        <a ng-href="/reports/finance/account_report/?{{ grid.appScope.slip('pdf', grid.appScope.filtersSlip, row.entity.id) }}" download="{{GENERAL_LEDGER.ACCOUNT_SLIP | translate }}">
+        <a ng-href="/reports/finance/account_report/?{{ grid.appScope.slip('pdf', grid.appScope.filtersSlip, row.entity.id) }}" translate-attr="{ download: '{{GENERAL_LEDGER.ACCOUNT_SLIP}}' }">
```
jniles: This can just be `{ download : 'GENERAL_LEDGE.ACCOUNT_SLIP' }`.  There is no need to use the extra `{{ }}`s.

client/src/modules/invoices/templates/grid/code.tmpl.html
```diff
@@ -14,7 +14,7 @@
     <span
       class="text-info"
       ng-show="row.entity._hasPriceList"
-      uib-popover="{{ 'PATIENT_INVOICE.BASE_PRICE_ALTERED' | translate }}"
+      translate-attr="{ 'uib-popover': 'PATIENT_INVOICE.BASE_PRICE_ALTERED' }"
```
jniles: I don't think this will work, unfortunately,  `uib-popover` is a directive, not an attribute.  Can you confirm that it works?  If not, we'll have to leave it the way it was.

client/src/modules/invoices/templates/grid/status.tmpl.html
```diff
@@ -7,10 +7,10 @@
   class="ui-grid-cell-contents">
 
   <!-- warning state: invalid, but not initialised yet -->
-  <span ng-hide="row.entity._initialised" class="fa fa-circle-o" uib-tooltip="{{ row.entity._message | translate }}" tooltip-placement="right" tooltip-append-to-body="true"></span>
+  <span ng-hide="row.entity._initialised" class="fa fa-circle-o" translate-attr="{ 'uib-tooltip': 'row.entity._message' }"  tooltip-placement="right" tooltip-append-to-body="true"></span>
```
jniles: The same observation goes for this one:  I'm not sure the `uib-tooltip` will work since it is a directive.  Additionally, the `row.entity._message` should not be wrapped in `'`.

client/src/modules/stock/exit/templates/lot.tmpl.html
```diff
@@ -3,7 +3,7 @@
     name="lot"
     ng-model="row.entity.lot"
     required>
-    <ui-select-match placeholder="{{ ::'FORM.PLACEHOLDERS.LOT' | translate }}">
+    <ui-select-match translate-attr="{ placeholder: 'FORM.PLACEHOLDERS.LOT' }">
```
jniles: :+1:

client/src/modules/templates/modals/transactionType.modal.html
```diff
@@ -4,36 +4,36 @@
   novalidate>
 
   <div class="modal-header">
-    {{ 'TREE.TRANSACTION_TYPE' | translate }}
-    <span ng-if="$ctrl.action === 'edit'" class="label label-info text-uppercase">{{ 'FORM.LABELS.EDIT' | translate }}</span>
-    <span ng-if="$ctrl.action === 'create'" class="label label-success text-uppercase">{{ 'FORM.LABELS.CREATE' | translate }}</span>
+    <span translarte>TREE.TRANSACTION_TYPE</span>
```
jniles: Typo here.

client/src/modules/templates/modals/transactionType.modal.html
```diff
@@ -4,36 +4,36 @@
...
 
     <div class="form-group" ng-class="{ 'has-error' : TransactionTypeForm.$submitted && TransactionTypeForm.type.$invalid }">
-      <label class="control-label">{{ "FORM.LABELS.TYPE" | translate }}</label>
+      <label class="control-label" translate> FORM.LABELS.TYPE</label>
       <select class="form-control"
         name="type"
         ng-model="$ctrl.transactionType.type" required>
-        <option value="" disabled>-- {{ 'FORM.SELECT.TYPE' | translate }} --</option>
-        <option value="{{opt.value}}" ng-repeat="opt in $ctrl.optionType">{{opt.label | translate}}</option>
+        <option value="" disabled>-- <span translate> FORM.SELECT.TYPE </span>--</option>
```
jniles: Same problem with the other `<option>`.  This should not have HTML inside it.

client/src/modules/transaction-type/transaction-type.ctrl.js
```diff
@@ -18,9 +18,9 @@ function TransactionTypeController(TransactionType, TransactionTypeStore, Notify
   // @TODO - move this into it's own HTML template.
   var editTemplate =
     '<div class="ui-grid-cell-contents">' +
-    '<a href title="{{ \'FORM.LABELS.EDIT\' | translate }}" ' +
+    '<a href translate-attr="{ title: \'FORM.LABELS.EDIT\' }"' +
     'ng-click="grid.appScope.editType(row.entity)" ' +
-    'uib-popover="{{grid.appScope.notAllowed(row.entity.fixed) | translate }}" ' +
+    'translate-attr="{ uib-popover : \'{{grid.appScope.notAllowed(row.entity.fixed) }} \'"' +
```
jniles: Does this work?

jniles (commented):
@jeremielodi this was a ton of work!  Well done!

I've made a few comments - there are some typos in the code, which is to be expected with a change this size.

I'm concerned that the `translate-attr` doesn't always work on directives, but I have not tested it recently.  Does this code:
```html
<ui-select-match translate-attr="{ placeholder: 'FORM.PLACEHOLDERS.ACCOUNT' }">
```
actually render a placeholder that is translated?  Does it also work for `uib-tooltip`s?  Can you quickly check to make sure?

jniles:
So ... it doesn't look like the `translate-attr` works at all for directives.  These should remain using the translate filter.

jeremielodi:
closed #2100

Issue #2046, opened 2017-08-23
(proposal) Journal transaction editing history


sfount:
We currently have no way of differentiating transactions that have been edited and transactions that were directly written by the system. 

If an accountant or developer is trying to audit the system to find out where something has gone wrong it would be very useful to know if something has been changed to know where to look for the issue. 

I propose we introduce a new table `transaction_history` that we will write in every time an editing session is successfully run on a transaction. 

The columns would be 
* uuid 
* transaction_uuid
* timestamp 
* user_id 
* (optional) Operations ['EDITING', 'DELETING', 'ADDING']

Whilst this doesn't give us a history of the full transaction it will be a good first step towards answering questions: 

Has a transaction been edited? 
```sql
SELECT COUNT(uuid) as numberOfEdits FROM transaction_history WHERE transaction_uuid = ?
```

Who has edited a specific transaction? 
```sql
SELECT user_id, timestamp FROM transaction_histroy JOIN user ON user_id = user.id WHERE transaction_uuid = ?
```

jniles:
To clarify, `transaction_uuid` should be `record_uuid`, correct?

This proposal LGTM.  Just knowing if a transaction has been messed with can inform us what next questions to ask.

jniles:
Note - this is tangentially related to https://github.com/IMA-WorldHealth/bhima-2.X/issues/1559

jniles:
I'll tackle this one.

The optional fields don't make much sense to me - this kind of resolution needs to be available on a line-by-line basis, not a transaction basis.

The only update I'll make is renaming `transaction_uuid` -> `record_uuid` since that is the column name found in the `posting_journal` and `general_ledger`.

jniles:
Closed in https://github.com/IMA-WorldHealth/bhima-2.X/pull/2156.

PR #2156, opened 2017-09-26
Feature: Transaction History

jniles:
This PR implements transaction histories as described in #2046.  A new table has been added - `transaction_history`.  This table contains the `record_uuid` of a transaction, the user id, and a timestamp for when it was edited.  The table is updated during the transaction edit database transaction.

Using this simplistic implementation, I've also included a small text on the Edit Transaction modal to show when the last edit was and by whom.  See below:

transactionhistorydisplay
_Fig 1: Transaction History Display in the Bottom Left of the Modal._

### DIFF
--- a/client/src/i18n/en/posting_journal.json
+++ b/client/src/i18n/en/posting_journal.json
@@ -5,6 +5,7 @@
     "ROWS" : "Row(s)",
     "EDITING" : "Editing",
     "NO_TRANSACTION_TYPE" : "No assigned transaction type",
+    "HAS_PREVIOUS_EDITS" : "Last edit by {{ user }} on {{ date }}",
     "JUSTIFICATION" : "Justification",
     "REFRESH" : "Refresh",
     "SAVE_TRANSACTION" : "Save Transaction",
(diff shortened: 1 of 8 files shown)


client/src/i18n/en/posting_journal.json
```diff
@@ -5,6 +5,7 @@
     "ROWS" : "Row(s)",
     "EDITING" : "Editing",
     "NO_TRANSACTION_TYPE" : "No assigned transaction type",
+    "HAS_PREVIOUS_EDITS" : "Last edit by {{ user }} on {{ date }}",
```
DedrickEnc, line 8: Why not attach it from the controller? so that we will avoid digest issue
This is a no blocking comment

DedrickEnc (commented):
@jniles 
I have made one comment, if you do not find it worth for a code change, I will merge it directly. But I think it is important to answer to the question may be.
LGTM

client/src/i18n/en/posting_journal.json
jniles: I don't think this should do more than one digest.  Since it uses the `translate` directive, it will only change if the `user` and `date` values change, which they never do.

Issue #1488, opened 2017-04-11
(feature) Accounting for stock


mbayopanda:
This issue is about the crucial functionality of stock management : Accounting for stock

Add accounting entries for all stock operations : 
- [x] Stock Entry : 
  - [x] Purchase
  - [x] Donation
  - [x] Integration

- [x] Stock Exit : 
  - [x] Distribution to patient
  - [x] Distribution to service
  - [x] Loss

jniles:
@mbayopanda @DedrickEnc could you review these issues and close the ones with are completed?

PR #2166, opened 2017-09-28
feat(stock accounting) : Stock exit accounting

mbayopanda:
This PR add : 
- A procedure `PostStockMovement` for writing in vouchers and journal
- A reference lookup for stock movement
- Transaction types for stock entry and stock exit
- Write into vouchers and journal for : 
  - stock exit to patient
  - stock exit to service
  - stock loss

related to #1488 about stock exit


server/lib/referenceLookup.js
```diff
@@ -26,6 +26,25 @@ function getEntity(req, res, next) {
...
+  const fluxId = codeRef[1];
+
+  if (code === STOCK_MOVEMENT_PREFIX) {
+    const type = stockMovementType(fluxId);
+    const queryDocument = `SELECT BUID(uuid) as uuid FROM document_map WHERE text = ?`;
+
+    return db.one(queryDocument, [req.params.codeRef])
+      .then(entity => {
+        const uuid = entity.uuid;
+        const path = ''.concat('/receipts/stock/', type.path, '/');
```
DedrickEnc: or ```['/receipts/stock/', type.path, '/'].join()the use ;``` to avoid the use of empty string 

 ``` `/receipts/stock/${type.path}/` ```

server/lib/referenceLookup.js
```diff
@@ -26,6 +26,25 @@ function getEntity(req, res, next) {
...
+        const url = `${path}${uuid}?lang=${language}&renderer=pdf`;
```
DedrickEnc: 👍

server/lib/referenceLookup.js
```diff
@@ -59,3 +78,31 @@ function indexIdentifiers() {
   });
 }
 
+// hack: handle special case of stock movements
+function stockMovementType(id) {
+  /**
```
DedrickEnc: I can not understand theses comments, why they are there?

server/controllers/stock/reports/stock/exit_patient_receipt.js
```diff
@@ -38,6 +38,7 @@ function stockExitPatientReceipt(req, res, next) {
     JOIN patient p ON p.uuid = m.entity_uuid
     JOIN project proj ON proj.id = p.project_id
     JOIN user u ON u.id = m.user_id
+    JOIN document_map dm ON dm.uuid = m.document_uuid
```
jniles, line 41: Personally, I worry this will have a larger performance impact than you expect, for no real gain in functionality.  The table `document_map` has ever single reference in the system.  That may make the server work really hard ... an SQL performance analysis would be needed to see exactly how it composes this query.

We'll have to keep an eye on this in the future to make sure it doesn't slow down execution.

server/lib/referenceLookup.js
jniles: Or, in ESMAScript 2015+:
```js
const path = `receipts/stock/${type.path}/`;
```

server/lib/referenceLookup.js
```diff
@@ -26,6 +26,25 @@ function getEntity(req, res, next) {
   const reference = codeRef[2];
   const documentDefinition = identifiersIndex[code];
 
+  // hack: handle special case of stock movements
```
jniles: Why does stock movements require a hack?

mbayopanda:
@DedrickEnc  could you review this PR ?

server/lib/referenceLookup.js
```diff
@@ -59,3 +78,31 @@ function indexIdentifiers() {
...
+function stockMovementType(id) {
+  /**
+   * Stock Receipt API
+   * /receipts/stock/{{name}}/:document_uuid
+   *
+   * the {{name}} is what we define for example in { key : 'FROM_PURCHASE', path :'entry_purchase' },
+   *
+   * empty {{name}} means that there is no API entry for this name.
+   */
+  const flux = {
```
jniles: Since this never changes, it would be easier to put it outside the function `stockMovementType`.  That way, the V8 engine will not allocate memory, remove it, allocate it, remove it, etc, on every HTTP request that hits this route.  If we make it a module-level constant, then it will be allocated at server start up and never freed.

```js
const flux = {
 // .... etc ...
};

function stockMovementType(id) {
  return flux[id];
}
```

server/lib/referenceLookup.js
```diff
@@ -59,3 +78,31 @@ function indexIdentifiers() {
   });
 }
 
+// hack: handle special case of stock movements
+function stockMovementType(id) {
```
jniles: A better name would be `getStockMovementType()` since this is going to return something.

server/models/procedures/stock.sql
```diff
@@ -34,10 +34,137 @@ CREATE PROCEDURE ComputeMovementReference (
...
+  DECLARE v_finished INTEGER DEFAULT 0;
+  
+  DECLARE stage_stock_movement_cursor CURSOR FOR 
+  	SELECT temp.stock_account, temp.cogs_account, temp.unit_cost, temp.quantity, temp.document_uuid, temp.is_exit 
+	FROM stage_stock_movement as temp;
+  
+  DECLARE CONTINUE HANDLER FOR NOT FOUND SET v_finished = 1;
+
+  -- temporarise the stock movement
+  CREATE TEMPORARY TABLE stage_stock_movement (
```
DedrickEnc: The ```stage_stock_movement``` can be null so this situation must be handled properly.
Bellow a blockof code I used to check that I am about to deal with inventories which have ```sale_account``` and ```cogd_account```

```
DECLARE InvalidInventoryAccounts CONDITION FOR SQLSTATE '45006';
  DECLARE verify_invalid_accounts SMALLINT(5);

  -- Check that every inventory has a stock account and a variation account 
  -- if they do not, the transaction will be Unbalanced, so the operation will not continue
  SELECT
    COUNT(l.uuid) INTO verify_invalid_accounts
  FROM
    stock_movement AS sm
  JOIN
    lot l ON l.uuid = sm.lot_uuid
  JOIN
  	inventory i ON i.uuid = l.inventory_uuid
  JOIN
  	inventory_group ig ON ig.uuid = i.group_uuid
  WHERE
  	ig.stock_account IS NULL AND
    ig.cogs_account IS NULL AND
    sm.document_uuid = document_uuid;

  IF verify_invalid_accounts > 0 THEN
    SIGNAL InvalidInventoryAccounts
    SET MESSAGE_TEXT = 'Every inventory should belong to a group with a cogs account and stock account.';
  END IF;
```
You should adapt this code

server/models/procedures/stock.sql
```diff
@@ -34,10 +34,137 @@ CREATE PROCEDURE ComputeMovementReference (
...
+  SET flux = (SELECT flux_id FROM stock_movement WHERE document_uuid = documentUuid LIMIT 1);
 
   INSERT INTO `document_map` (uuid, text)
-  VALUES (documentUuid, CONCAT('MVT.', reference))
+  VALUES (documentUuid, CONCAT_WS('.', 'SM', flux, reference))
   ON DUPLICATE KEY UPDATE uuid = uuid;
 END $$
+
+-- post stock movement into vouchers
+DROP PROCEDURE IF EXISTS PostStockMovement;
```
DedrickEnc, line 48: Very good stored procedure, it process once at all the voucher and voucher item inserting.

server/models/procedures/stock.sql
```diff
@@ -34,10 +34,137 @@ CREATE PROCEDURE ComputeMovementReference (
...
+      JOIN lot l ON l.uuid = m.lot_uuid 
+      JOIN inventory i ON i.uuid = l.inventory_uuid 
+      JOIN inventory_group ig 
+        ON ig.uuid = i.group_uuid AND (ig.stock_account IS NOT NULL AND ig.cogs_account IS NOT NULL) 
+      WHERE m.document_uuid = documentUuid AND m.is_exit = isExit
+    );
+
+  -- define voucher variables
+  SET voucher_uuid = (SELECT HUID(UUID()));
+  SET voucher_date = (SELECT date FROM stage_stock_movement LIMIT 1);
```
jniles: What you probably want to do is this:
```js
SELECT 
  date, project_id, currency_id
INTO
  voucher_date, voucher_project_id, voucher_currency_id
 FROM stage_stock_movement;
```

server/models/procedures/stock.sql
```diff
@@ -34,10 +34,137 @@ CREATE PROCEDURE ComputeMovementReference (
...
+      FROM stock_movement m  
+      JOIN lot l ON l.uuid = m.lot_uuid 
+      JOIN inventory i ON i.uuid = l.inventory_uuid 
+      JOIN inventory_group ig 
+        ON ig.uuid = i.group_uuid AND (ig.stock_account IS NOT NULL AND ig.cogs_account IS NOT NULL) 
+      WHERE m.document_uuid = documentUuid AND m.is_exit = isExit
+    );
+
+  -- define voucher variables
+  SET voucher_uuid = (SELECT HUID(UUID()));
```
DedrickEnc: All your select are in one table, instead of having many select like this you can do
``` SELECT @var1 = column1, @var2 = column2 FROM table```

DedrickEnc (changes requested):
@mbayopanda 
Your PR is very good particularly your stored procedure, I have made some comments to improve. I ma waiting for your change to adapt my stored procedure.
Also I don't know if you really merge, but it will be good to rebase next time instead of merging.

server/controllers/stock/reports/stock/exit_patient_receipt.js
mbayopanda: 👍 you're right

server/lib/referenceLookup.js
mbayopanda: the `referenceLookup.js` has a structure for handling reference like `CP.HBB.001` (Prefix.Project.Reference) but the stock a reference like `SM.1.1` (Prefix.FluxId.ComputedReference).

This file was not created for handling reference such as for stock movement, so we just put the code here for having the same result also for stock movement.

server/lib/referenceLookup.js
mbayopanda: Normally this file was for handling just reference of this format : `Prefix.Project.Reference` but stock movement has the format : `Prefix.FluxId.Reference`.

server/lib/referenceLookup.js
jniles: Hmm..  then why does stock use `referenceLookup` if the reference is not similar?  Maybe it should use a different function?

mbayopanda:
@DedrickEnc could you review this PR ?

DedrickEnc (approved):
All my suggestions and comments have been addressed.
LGTM

PR #2167, opened 2017-09-28
Default autoParseStatement to false

### LINKED ISSUE #2163: Turn autoParseStatements off by default in FilterParser
The `FilterParser` library has been extremely useful in making complex registry queries easy to use.  One of the beginning design flaws was that it magically assigned every query string value to the search query.  This was a Bad Idea (TM).  It was fixed with the flag `autoParseStatements`, introduced later and set "on" by default to preserve default behavior.

Most of our FilterParsers have been migrated to setting `autoParseStatements` to `false`.  We should now preserve this as default behavior by making it off by default in the `FilterParser` library.  Every module can then be cleaned up by removing that flag.

DedrickEnc:
This PR set the ```autoParseStatement``` to false by default and update every filter in the server.
closes #2163

### DIFF
--- a/server/controllers/finance/cash.js
+++ b/server/controllers/finance/cash.js
@@ -137,7 +137,7 @@ function read(req, res, next) {
 function listPayment(options) {
   // ensure epected options are parsed appropriately as binary
   db.convert(options, ['debtor_uuid', 'debtor_group_uuid']);
-  const filters = new FilterParser(options, { tableAlias : 'cash', autoParseStatements : false });
+  const filters = new FilterParser(options, { tableAlias : 'cash' });
 
   const sql = `
     SELECT BUID(cash.uuid) as uuid, cash.project_id,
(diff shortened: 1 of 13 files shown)


server/controllers/finance/cash.js
```diff
@@ -137,7 +137,7 @@ function read(req, res, next) {
 function listPayment(options) {
   // ensure epected options are parsed appropriately as binary
   db.convert(options, ['debtor_uuid', 'debtor_group_uuid']);
-  const filters = new FilterParser(options, { tableAlias : 'cash', autoParseStatements : false });
+  const filters = new FilterParser(options, { tableAlias : 'cash' });
```
mbayopanda, line 140: 👍

mbayopanda (approved):
LGTM

mbayopanda:
@DedrickEnc could you rebase this branch with master

bors r+

PR #2169, opened 2017-09-29
feat(developpement build sql) merge all files in one

### LINKED ISSUE #2080: chore: merge all development build SQL files
We currently have several files in the [test directory](https://github.com/IMA-WorldHealth/bhima-2.X/tree/master/test).  Specifically:
 1. `update_dev_build.sql`
 2. `development-build.sql`
 3. `update-development-build.sql`

It's unclear why these three are separated and which should be built/executed before which other one.  We should merge all these files into a single SQL file that is simply built once ... ideally the test build should not be altered often.

jeremielodi:
This PR gathers all development sql files in one
closes #2080


jniles (approved):
bors r+

Issue #2139, opened 2017-09-25
Troubleshoot ID Display Problems in Some Filters


lomamech:
We note that with several filters and for some registers, the filter directly displays a very long UUID or an ID that are generally incomprehesive for a user

![image](https://user-images.githubusercontent.com/8035588/30814740-53f38dfc-a211-11e7-9ea8-ec3c46cbdf7b.png)

jniles:
This is related to https://github.com/IMA-WorldHealth/bhima-2.X/issues/2130. :+1:

This is an excellent thing to work on!

jniles:
Closed in a series of PRs: #2186, #2184, #2182, #2170, #2178.

PR #2170, opened 2017-09-29
feat(transaction links in journal)

### LINKED ISSUE #2120: Feature Request: Link to Transactions in Journal from Registries
We have awesome links between registries.  This would be complimented by a link to the Journal.  Since the Journal can now display posted/unposted records, this style of link is trivial ... one would just need to put two filters - one to show posted records and the other to filter on the `record_uuid` to link this data.

If we implement this, we will need to address https://github.com/IMA-WorldHealth/bhima-2.X/issues/520 as well, since many users are no allowed to see the Journal.  It will need to say something like "You are not allowed to view this resource" as an error page instead of a giant 403 which is not comprehensible to a user.

jeremielodi:
This PR adds a Link to Transactions in Journal from Registries 
closes #2120

### DIFF
--- a/client/src/modules/cash/payments/templates/action.cell.html
+++ b/client/src/modules/cash/payments/templates/action.cell.html
@@ -36,7 +36,7 @@
     </li>
     <li class="divider"></li>
     <li>
-      <a data-method="viewInvoice" ui-sref="journal({ filters : [{ key: 'record_uuid', value : row.entity.uuid, displayValue: row.entity.reference}]})" href>
+      <a data-method="viewInvoice" ui-sref="journal({ filters : [{ key: 'record_uuid', value : row.entity.uuid, displayValue: row.entity.reference}, { key : 'period', value : 'allTime' }]})" href>
         <span class="fa fa-file-text-o"></span> <span translate>TRANSACTIONS.VIEW_TRANSACTIONS</span>
       </a>
     </li>
--- a/client/src/modules/cash/payments/templates/action.cell.html
+++ b/client/src/modules/cash/payments/templates/action.cell.html
@@ -34,5 +34,12 @@
         <span class="fa fa-file-text-o"></span> <span translate>REPORT.VIEW_INVOICE</span>
       </a>
     </li>
+    <li class="divider"></li>
+    <li>
+      <a data-method="viewInvoice" ui-sref="journal({ filters : [{ key: 'record_uuid', value : row.entity.uuid, displayValue: row.entity.reference}]})" href>
+        <span class="fa fa-file-text-o"></span> <span translate>TRANSACTIONS.VIEW_TRANSACTIONS</span>
+      </a>
+    </li>
+
   </ul>
 </div>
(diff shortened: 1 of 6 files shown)


client/src/modules/vouchers/templates/action.cell.html
```diff
@@ -14,5 +14,10 @@
         <i class="fa fa-file-pdf-o"></i> <span translate>TABLE.COLUMNS.RECEIPT</span>
       </a>
     </li>
+    <li>
+      <a data-method="viewInvoice" ui-sref="journal({ filters : [{ key: 'record_uuid', value : row.entity.uuid}]})" href>
```
jniles: This method probably shouldn't be `viewInvoice`, since you are in the voucher registry.

client/src/modules/invoices/registry/templates/action.cell.html
```diff
@@ -34,6 +34,11 @@
         <span class="fa fa-money"></span> <span translate>REPORT.VIEW_PAYMENTS</span>
       </a>
     </li>
+    <li>
+      <a data-method="viewInvoice" ui-sref="journal({ filters : [{ key: 'record_uuid', value : row.entity.uuid}]})" href>
```
jniles: At the moment, all of these will show a giant uuid at the top ... which isn't very pretty.  @lomamech was working on fixing this in #2139.

You can prevent introducing new ones in your PR by using `displayValue` in the filters.  This will pass a value to be displayed.  The `value` will be used to filter.

client/src/modules/cash/payments/templates/action.cell.html
```diff
@@ -34,5 +34,12 @@
         <span class="fa fa-file-text-o"></span> <span translate>REPORT.VIEW_INVOICE</span>
       </a>
     </li>
+    <li class="divider"></li>
+    <li>
+      <a data-method="viewInvoice" ui-sref="journal({ filters : [{ key: 'record_uuid', value : row.entity.uuid}]})" href>
```
jniles: The `method` here should talk about cash.

jniles:
@jeremielodi, this is an excellent start.  The behavior is nice, the code is clean.

One problem is if the transaction is posted or is not within the search parameters that the user specified.  For example, if the posting journal is set to "today", it will never show the transactions that happened yesterday.

I propose you add the filters for "all time" and "include posted records" to make sure that the transaction is displayed.

jniles:
@jeremielodi, just one more thing as mentioned above - this should also toggle the "Include Posted Records" checkbox.

client/src/modules/journal/journal.js
```diff
@@ -504,8 +505,8 @@ function JournalController(
   function startup() {
     var hasStateFilters = $state.params.filters.length > 0;
     if (hasStateFilters) {
+      $state.params.filters.push({ key : 'includeNonPosted', value : '1' });
```
jniles: You can just put this in the link the same as the rest of the filters.  There's no reason to put some filters in the controller and others in the links.

jniles: Like this:
```html
<a data-method="viewInvoice" ui-sref="journal({ filters : [{ key: 'record_uuid', value : row.entity.uuid, displayValue: row.entity.reference}, { key : 'period', value : 'allTime' }, { key : 'includeNonPosted', value : 1, displayValue : 'true' }]})" href>
```

jniles (approved):
This looks much better :+1:

PR #2173, opened 2017-09-29
Posting cleanning Data into posting journal

### LINKED ISSUE #2168: Remove Posting Journal and General Ledger from test/data.sql
We have a few lines of the Posting Journal (e.g. https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/data.sql#L382) and the General Ledger (e.g. https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/data.sql#L461).  Let's get rid of these.  They break links between the Journal and each module that creates them.  It also makes the period totals behave strangely since not all transaction that happened are reflected in the ledgers.

Instead, we should do this - after every invoice, voucher, or cash payment is defined in the `test/data.sql`, we use `CALL PostMethod(uuid)` to post the value into the posting journal.  Then, we can use `PostToGeneralLedger()` to post the transactions that we wish to the General Ledger.  In this way, all tables will be in sync.

DedrickEnc:
This P.R removes records in the posting journal and general ledger and post only data from invoice, cash and voucher table to make sure data remains clean.

It closes #2168

### DIFF
--- a/test/data.sql
+++ b/test/data.sql
@@ -340,6 +340,11 @@ INSERT INTO `patient_group_subsidy` VALUES
 INSERT INTO `debtor_group_subsidy` VALUES
   (1, HUID('4de0fe47-177f-4d30-b95f-cff8166400b4'), 1, CURRENT_TIMESTAMP);
 
+-- exchange rate for the current date
+INSERT INTO `exchange_rate` VALUES
+  (1,1,1,900.0000, DATE('2016-01-01')),
+  (2,1,1,930.0000, NOW());
+
 -- voucher sample data
 SET @first_voucher = HUID('a5a5f950-a4c9-47f0-9a9a-2bfc3123e534');
 SET @second_voucher = HUID('304cfa94-0249-466c-9870-95eb3c221b0a');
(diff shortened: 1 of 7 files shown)


test/data.sql
```diff
@@ -439,6 +438,8 @@ INSERT INTO `depot` VALUES
   (@depot_uuid, 'Depot Principal', 1, 1),
   (@second_depot_uuid, 'Depot Secondaire', 1, 0);
 
+-- TODO : As soon as the stored Procedure for Stock accounting landed, stock movement records should be posted also
```
jniles, line 441: :+1:

test/integration/accounts.js
```diff
@@ -124,9 +124,9 @@ describe('(/accounts) Accounts', function () {
         expect(res).to.be.json;
         expect(res.body).to.not.be.empty;
         expect(res.body).to.have.all.keys('account_id', 'debit', 'credit', 'balance');
-        expect(res.body.debit).to.equal(125);
+        expect(res.body.debit).to.equal(600);
```
jniles, line 127: :+1:

jniles (approved):
Looks awesome!  This has been needed for a long time.

bors r+

PR #2176, opened 2017-10-02
Block Inventory Items from being Invoiced

### LINKED ISSUE #2140: Block Inventory Items from being Invoiced
At Vanga, there are a few inventory items that should not be invoiced.  It is not possible to delete them because they were used in the past.  However, they should no longer be invoiced to patients because they have been retired or combined into a service.

I propose we implement a `locked` flag, like in many of our other modules.  This `locked` flag would prevent those from being loading into the invoice module.  As usual, the "locked" flag should only be available in the editing state (not the create state), and should cross out the item in the grid row.

DedrickEnc:
This P.R addresses #2140 

It add the functionality of invoicing only unlocked inventory and let user filter inventory based on the inventory locked state.
The enventory can be locked during the editing phase.

closes #2140

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -341,6 +341,7 @@
             "LAST_PAGE": "Last Page Visited",
             "LOCATION_REGISTER": "Location Register",
             "LOCKED": "Locked",
+            "UNLOCKED": "Unlocked",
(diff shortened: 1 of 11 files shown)


jniles (approved):
Code LGTM!

bors r+

Issue #2130, opened 2017-09-25
refactor: make all links use a human-readable name


jniles:
Many of our links between modules/registries do not show human-readable strings in the filter bar.  An example is shown below:
![nonhumanreadablelinkstart](https://user-images.githubusercontent.com/896472/30800449-ecb3ed6e-a1d7-11e7-918d-4b1cb765e44d.png)
_Fig 1: Click this link to show all debtors_
![nonhumanreadablelink](https://user-images.githubusercontent.com/896472/30800448-eca55eb6-a1d7-11e7-855a-3d203cf7f3b4.png)
_Fig 2: The Debtor Group is unreadable!_

We have the tools to make sure that these are shown using [the `displayValue` property](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/FilterService.js#L60) in Filter Service.  This can be set by the sending module as the filters are passed through `$state` parameters.  By implementing this, it would significantly increase the usability of the registry system.

lomamech:
I want to close both issues at the same time

jniles:
@lomamech, no problem.  Close whichever one you feel is best.

jniles:
Closed in a series of PRs: #2186, #2184, #2182, #2170, #2178.

PR #2177, opened 2017-10-02
refactor(Make all links use a human readable): In posting Journal

lomamech:
- Implement component bhTransactionType
- Implement test E2E for the component bhTransactionType
- Update journal search Modal and display readable link


client/src/js/components/bhTransactionType.js
```diff
@@ -0,0 +1,50 @@
+angular.module('bhima.components')
```
jniles: There is already a [bhTransactionTypeSelect](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/components/bhTransactionTypeSelect.js).  Can you use that instead?

jniles (commented):
It looks like many of the comments from https://github.com/IMA-WorldHealth/bhima-2.X/pull/2178 would be applicable here.  Let me know if you can adapt the comments into something nice.  If not, I'll be happy to work with you in person on it!

client/src/js/components/bhTransactionType.js
lomamech: The reason I had created another component and that in the old bhTransactionTypeSelect. we had the possibility of choosing several type transactions whereas in the bhTransactionType we have the option of choosing one

client/src/modules/journal/modals/search.modal.js
```diff
@@ -80,8 +58,18 @@ function JournalSearchModalController(Instance, Notify,
   // custom filter account_id - assign the value to the searchQueries object
   vm.onSelectAccount = function onSelectAccount(account) {
     vm.searchQueries.account_id = account.id;
+    displayValues.account_id = String(account.number).concat(' - ', account.label);
```
jniles, line 80: There is no need to use `String` - what you are looking for is:
```js
displayValues.account_id = account.hrLabel;
```

The `AccountService` formats labels after they are downloaded - see [this line](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/accounts/accounts.service.js#L45).

client/src/js/components/bhProjectSelect.js
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Project Select component
+ *
+ */
+function projectSelectController(Projects, Notify) {
+  var $ctrl = this;
+
+  $ctrl.$onInit = function onInit() {    
+    // fired when a project has been selected
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
```
jniles: You don't actually need this - angular will automatically assign a `noop()` to it since it so not required (`<?`).

client/src/js/components/bhProjectSelect.js
```diff
@@ -0,0 +1,47 @@
...
+    // default for form name
+    $ctrl.name = $ctrl.name || 'FORM.LABELS.PROJECT';
+
+    Projects.read()
+      .then(function (projects) {
+        $ctrl.projects = projects;
+      })
+      .catch(Notify.handleError);
+
+    $ctrl.valid = true;
```
jniles: Where did `valid` come from?

client/src/js/components/bhProjectSelect.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+function projectSelectController(Projects, Notify) {
+  var $ctrl = this;
+
+  $ctrl.$onInit = function onInit() {    
+    // fired when a project has been selected
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
+
+    // default for form name
+    $ctrl.name = $ctrl.name || 'FORM.LABELS.PROJECT';
```
jniles: There's no need for this child form.  We can remove it.

client/src/js/components/bhProjectSelect.js
```diff
@@ -0,0 +1,47 @@
...
+
+projectSelectController.$inject = [
+  'ProjectService', 'NotifyService'
+];
+
+/**
+ * Project Select component
+ *
+ */
+function projectSelectController(Projects, Notify) {
```
jniles: Controllers should begin with a capital letter.

server/controllers/finance/journal/index.js
```diff
@@ -138,9 +144,10 @@ function buildTransactionQuery(options, posted) {
   filters.equals('account_id');
   filters.equals('project_id');
   filters.equals('trans_id');
-  filters.equals('origin_id');
   filters.equals('record_uuid');
 
+  filters.custom('origin_id', 'p.origin_id IN (?)', [typeIds]);
```
jniles, line 149: :+1:

client/src/modules/templates/bhProjectSelect.tmpl.html
```diff
@@ -0,0 +1,23 @@
+<div ng-form="{{ $ctrl.name }}" bh-project-select ng-model-options="{ updateOn: 'default' }">
```
jniles: You can remove this `ng-form` call.  it doesn't actually do anything.

client/src/js/components/bhProjectSelect.js
```diff
@@ -0,0 +1,47 @@
...
+  .component('bhProjectSelect', {
+    templateUrl : 'modules/templates/bhProjectSelect.tmpl.html',
+    controller  : projectSelectController,
+    transclude  : true,
+    bindings    : {
+      projectId        : '<',
+      disable          : '<?',
+      onSelectCallback : '&',
+      name             : '@?',
+      required         : '<?',
```
jniles: Go ahead and remove `required` and `disable`.  You never use them.

client/src/modules/journal/modals/search.modal.js
```diff
@@ -52,42 +57,62 @@ function JournalSearchModalController(Instance, Projects, Notify,
...
 
   // handle component selection states
   // custom filter account_id - assign the value to the searchQueries object
   vm.onSelectAccount = function onSelectAccount(account) {
     vm.searchQueries.account_id = account.id;
+    displayValues.account_id = String(account.number).concat(' - ', account.label);
   };
 
+  // Set displayLabel if the filters account is defined
+  if (filters.account_id) {
```
jniles: I still don't understand why this is even necessary.  Can'y we just do:
```js
displayValues.account_id = account.hrLabel;
```
and that be it?

All of these `if` statements are really ugly.  Let me fork this PR and see if I can get a cleaner version.

jniles (commented):
A few clean up things here.

All the `if (filters.key) {` statements look really ugly.  I'm going to try and fork this PR and make it cleaner.  My gut feels like it should be easier than this.

client/src/modules/journal/modals/search.modal.js
```diff
@@ -117,16 +137,16 @@ function JournalSearchModalController(Instance, Projects, Notify,
...
-      }
+        // default to the original value if no display value is defined
+        var displayValue = displayValues[key] || lastDisplayValues[key] || value;
+        changes.post({ key: key, value: value, displayValue: displayValue });
+       }
     });
 
     var loggedChanges = changes.getAll();
 
+    console.log('changes', loggedChanges);
```
jniles: This `console.log()` should be removed before merging.

client/src/js/components/bhProjectSelect.js
```diff
@@ -0,0 +1,36 @@
...
+  $ctrl.$onInit = function onInit() {    
+    Projects.read()
+      .then(function (projects) {
+        $ctrl.projects = projects;
+      })
+      .catch(Notify.handleError);
+  };
+
+  // fires the onSelectCallback bound to the component boundary
+  $ctrl.onSelect = function ($item, $model) {
```
jniles: If you don't use the `$model`, you can leave it out.

jniles (commented):
Looks good to me.  There's a console log that should be removed and then this can land.  Great work!

jniles:
@lomamech, it looks like there are conflicts here - could you take a look at this?

jniles (approved):
Looks good to me!

bors r+

jniles:
That's weird.

bors r+

PR #2178, opened 2017-10-02
refactor(Make all links readable in Voucher)

lomamech:
- Update search modal, make select User in readable format
- Update API for finance/voucher fixe the filter in select multiple
  transaction type
- Update select transanction type in readable format

### DIFF
--- a/client/src/modules/vouchers/modals/search.modal.js
+++ b/client/src/modules/vouchers/modals/search.modal.js
@@ -30,6 +30,12 @@ function VoucherRegistrySearchModalController(ModalInstance, filters, Notify, mo
 
   var lastViewFilters = Vouchers.filters.formatView().customFilters;
 
+  // map key to last display value for lookup in loggedChange
+  var lastDisplayValues = lastViewFilters.reduce(function (object, filter) {
+    object[filter._key] = filter.displayValue;
+    return object;
+  }, {});  
+
   // displayValues will be an id:displayValue pair
   var displayValues = {};
   
@@ -65,29 +71,11 @@ function VoucherRegistrySearchModalController(ModalInstance, filters, Notify, mo
     displayValues.type_ids = typeText;
   };
 
-  // Set displayLabel if the filters user is defined
-  if (filters.type_ids) {
-    lastViewFilters.forEach(function (filter) {
-      if (filter._key === 'type_ids') {
-        displayValues.type_ids = filter._displayValue;
-      }
-    });
-  }
-
   // custom filter user_id - assign the value to the params object
   vm.onSelectUser = function onSelectUser(user) {
     vm.searchQueries.user_id = user.id;
     displayValues.user_id = user.display_name;
   };
-
-  // Set displayLabel if the filters user is defined
-  if (filters.user_id) {
-    lastViewFilters.forEach(function (filter) {
-      if (filter._key === 'user_id') {
-        displayValues.user_id = filter._displayValue;
-      }
-    });
-  }
 
   // default filter period - directly write to changes list
   vm.onSelectPeriod = function onSelectPeriod(period) {
@@ -124,7 +112,7 @@ function VoucherRegistrySearchModalController(ModalInstance, filters, Notify, mo
     angular.forEach(vm.searchQueries, function (value, key) {
       if (angular.isDefined(value)) {
         // default to the original value if no display value is defined
-        var displayValue = displayValues[key] || value;
+        var displayValue = displayValues[key] || lastDisplayValues[key] || value;
         changes.post({ key: key, value: value, displayValue: displayValue });
        }
     });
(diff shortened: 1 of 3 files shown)


server/controllers/finance/vouchers.js
```diff
@@ -115,7 +115,7 @@ function find(options) {
   let typeIds = [];
 
   if (options.type_ids) {
-    typeIds = typeIds.concat(options.type_ids);
+    typeIds = options.type_ids.toString();
```
jniles: Why was this change necessary?  The previous version escaped an array of values, but this one turns them into a string without escaping.  In theory, I could send `?type_ids=1 OR 1=1` and bypass all filtering via an SQL injection.

client/src/modules/vouchers/voucher-registry.ctrl.js
```diff
@@ -116,6 +116,9 @@ function VoucherController(
   function search() {
     var filtersSnapshot = Vouchers.filters.formatHTTP();
 
+    // Keep the last filters made during the last search and keeps the display in a readable format
```
jniles: This is an interesting way to pass filters in ... but it would probably be better to use the service to get the filters in the modal controller.  It should be as simple as importing `VoucherService` into the `ModalController` and then doing `var filters = VoucherService.filters.formatView();`

client/src/modules/vouchers/modals/search.modal.js
```diff
@@ -74,8 +98,21 @@ function VoucherRegistrySearchModalController(ModalInstance, filters, Notify, mo
 
     // push all searchQuery values into the changes array to be applied
     angular.forEach(vm.searchQueries, function (value, key) {
-      if (angular.isDefined(value)) {
-        changes.post({ key : key, value : value });
+      if (angular.isDefined(value)) {
```
jniles: This is pretty complicated code.

Why don't you track display values in a separate object.  It doesn't even need to be tied to the view - something like this:
```js
// searchQueries is the same id:value pair
vm.searchQueries = {};
// displayValues will be an id:displayValue pair
var displayValues = {};
```
... then, each individual method can define how to display it's display value.  For example, for `user_id`, you can just do this:
```js
// search on user.id, but show user.display_name in the filters
vm.onSelectUser = function onSelectUser(user) {
  vm.searchQueries.user_id = user.id;
  displayValues.user_id = user.display_name;
};
```
... and then in the submit method ...
```js
angular.forEach(vm.searchQueries, function (value, key) {
    var displayValue = displayValues[key] || value;
    changes.post({ key: key, value: value, displayValue: displayValue });
   }
});
```

client/src/modules/vouchers/voucher-registry.ctrl.js
jniles: Having read all the code now, I actually don't think you need to share the view values at all.  It should all be driven by the search modal.

jniles (changes requested):
@lomamech this is on the right track, but there is a much easier way to handle these changes.  I've proposed it in the comments below.

server/controllers/finance/vouchers.js
lomamech: Because with typeIds = typeIds.concat (options.type_ids); the query filters only the first selected element
error

jniles: Try using ES6 then - typeIds = [...options.type_ids];

Chances are `type_ids` is no longer being sent as an array, which is what the server API expects.

client/src/modules/vouchers/modals/search.modal.js
```diff
@@ -13,32 +14,67 @@ VoucherRegistrySearchModalController.$inject = [
...
+
   if (filters.limit) {
     vm.defaultQueries.limit = filters.limit;
   }
 
   vm.onTransactionTypesChange = function onTransactionTypesChange(transactionTypes) {
     vm.searchQueries.type_ids = transactionTypes;
+    var typeText = '/';
+
+    transactionTypes.forEach(function (typeId) {
```
jniles, line 63: This solution is okay for now.

What we should probably do is modify the `onTransactionTypeChange()` callback to send back the list of transaction types, rather than a list of IDs.  That way we can simply use their labels instead of downloading transaction types twice.

server/controllers/finance/vouchers.js
```diff
@@ -144,7 +144,7 @@ function find(options) {
   // @todo - could this be improved
   filters.custom('entity_uuid', 'v.uuid IN (SELECT DISTINCT voucher_uuid FROM voucher_item WHERE entity_uuid = ?)');
 
-  filters.custom('type_ids', 'v.type_id IN (?)', typeIds);
+  filters.custom('type_ids', 'v.type_id IN (?)', [typeIds]);
```
jniles, line 147: 💃

jniles (approved):
Looks good to me!

bors r+

PR #2185, opened 2017-10-05
feat(debtorgroup color) using a dropdown menu to pick a color

### LINKED ISSUE #2121: refactor: Make color picker into a dropdown menu of colors
The Debtor Groups module allows users to set a color for a debtor group to aid accessibility (https://github.com/IMA-WorldHealth/bhima-2.X/issues/484).  It hasn't proven its worth yet, so we may remove it in the long run.  However, in the short run, we can improve it to make it easier to use.

**Recommendation**
Make the color wheel into a simple dropdown with ~10 color options.  Chances are a user will not more than 10 options since there will not be many debtor groups.  This is similar to what Github does with labels - only a limited number of colors are presented to the user to make it easier for the them to pick.

One additional factor is that the `<input type="color">` renders differently across operating systems and browers.  If we make the color input a simple drop down, we can guarantee its styling across browsers/oses.

jeremielodi:
This PR replace a color input by a drop drown menu color
closes #2121

### DIFF
--- a/client/src/modules/debtors/groups.update.html
+++ b/client/src/modules/debtors/groups.update.html
@@ -23,9 +23,19 @@
 
           <div
             class="form-group has-feedback"
-            ng-class="{'has-error' : debtorGroup.color.$invalid && debtorGroup.$submitted}">
+            ng-class="{'has-error' : !GroupUpdateCtrl.group.color && debtorGroup.$submitted}">
             <label class="control-label" translate>FORM.LABELS.COLOR</label>
-            <input name="color" id="title-color" type="color" ng-model="GroupUpdateCtrl.group.color" class="form-control" ng-value="GroupUpdateCtrl.group.color">
+
+            <ui-select tagging ng-model="GroupUpdateCtrl.group.color">
+              <ui-select-match placeholder="Pick one...">
+                <i class="fa fa-circle" style="color: {{$select.selected.value}}; font-sze:14px"></i>
+                &nbsp;<span translate>{{$select.selected.name}}</span>
+              </ui-select-match>
+              <ui-select-choices repeat="val.value as val in GroupUpdateCtrl.colors | filter: {'name' :$select.search}">
+                <div>
+                  <span><i class="fa fa-circle" style="color: {{val.value}}; font-sze:14px"></i>&nbsp; <span translate>{{val.name}} </span> </span>
+                </div>
+              </ui-select>
           </div>
 
           <bh-account-select
@@ -135,10 +145,10 @@
             <p translate>DEBTOR_GROUP.CONVENTION.DESCRIPTION</p>
             <div class="checkbox">
               <label>
-                <input 
+                <input
                   type="checkbox"
                   name="is_convention"
-                  ng-model="GroupUpdateCtrl.group.is_convention" 
+                  ng-model="GroupUpdateCtrl.group.is_convention"
                   ng-true-value="1"
                   ng-false-value="0">
                 <span translate>DEBTOR_GROUP.CONVENTION.TITLE</span>
@@ -160,7 +170,7 @@
 
                 <!-- Only show if the subsidies option is unchecked -->
                 <p ng-if="GroupUpdateCtrl.group.apply_subsidies===false" class="text-info">
-                  <span class="glyphicon glyphicon-info-sign"></span> 
+                  <span class="glyphicon glyphicon-info-sign"></span>
                   <span translate>DEBTOR_GROUP.POLICIES.SUBSIDIES.INFO</span>
                 </p>
                 <div class="checkbox">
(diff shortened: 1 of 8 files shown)


client/src/modules/debtors/groups.update.html
```diff
@@ -23,9 +23,19 @@
 
           <div
             class="form-group has-feedback"
-            ng-class="{'has-error' : debtorGroup.color.$invalid && debtorGroup.$submitted}">
+            ng-class="{'has-error' : !GroupUpdateCtrl.group.color && debtorGroup.$submitted}">
             <label class="control-label" translate>FORM.LABELS.COLOR</label>
-            <input name="color" id="title-color" type="color" ng-model="GroupUpdateCtrl.group.color" class="form-control" ng-value="GroupUpdateCtrl.group.color">
+
+            <ui-select tagging ng-model="GroupUpdateCtrl.group.color" theme="bootstrap">
```
jniles: You don't need `theme="bootstrap"` here.  We already define it in our `app.js` configuration.  See [this file](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/app.js#L176).

client/src/modules/debtors/groups.update.html
```diff
@@ -23,9 +23,19 @@
...
+              <ui-select-match placeholder="Pick one...">
+                  <i class="fa fa-circle" style="color: {{$select.selected.value}}; font-sze:14px"></i>&nbsp;
+                   <span translate>{{$select.selected.name}}</span>
+              </ui-select-match>
+              <ui-select-choices repeat="val.value as val in GroupUpdateCtrl.colors | filter: {'name' :$select.search}">
+                  <div ng-click="GroupUpdateCtrl.group.color = val.value">
```
jniles: Do this `ngClick` need to be defined?  You already set the `ng-model` to this value.

The proper way to get a selection from a `uiSelect` is using the `onSelect` callback anyway.  [Here is the documentation](https://github.com/angular-ui/ui-select/wiki/ui-select#events).

client/src/modules/debtors/groups.update.html
```diff
@@ -23,9 +23,19 @@
...
+                      <h5><i class="fa fa-circle" style="color: {{val.value}}; font-sze:14px"></i>&nbsp; <span translate>{{val.name}} </span> </h5>
```
jniles: The `<h5>` element makes this really large.  Can you replace it with a span?

client/src/modules/debtors/groups.update.html
```diff
@@ -23,9 +23,19 @@
...
+                  </div>
+              </ui-select-cfhoices>
```
jniles: This has a typo.  I'm surprised it works!

test/end-to-end/debtors/debtorgroups.spec.js
```diff
@@ -29,8 +29,8 @@ describe('Debtor Groups Management', function () {
     FU.input('GroupUpdateCtrl.group.name', 'E2E Debtor Group');
 
     // Fix me : What is the best way to use color picker with protractor
```
jniles: You can remove this comment now.  It is out of date.

jniles (changes requested):
A few cleanup things to address here.

client/src/modules/debtors/groups.update.html
```diff
@@ -23,9 +23,19 @@
 
           <div
             class="form-group has-feedback"
-            ng-class="{'has-error' : debtorGroup.color.$invalid && debtorGroup.$submitted}">
+            ng-class="{'has-error' : !GroupUpdateCtrl.group.color && debtorGroup.$submitted}">
             <label class="control-label" translate>FORM.LABELS.COLOR</label>
-            <input name="color" id="title-color" type="color" ng-model="GroupUpdateCtrl.group.color" class="form-control" ng-value="GroupUpdateCtrl.group.color">
+
+            <ui-select tagging ng-model="GroupUpdateCtrl.group.color">
```
jniles, line 29: I don't think you need the `tagging` directive here.  I'm not even sure what it does.

client/src/modules/debtors/groups.update.html
```diff
@@ -23,9 +23,19 @@
...
+              <ui-select-match placeholder="Pick one...">
+                <i class="fa fa-circle" style="color: {{$select.selected.value}}; font-sze:14px"></i>
+                &nbsp;<span translate>{{$select.selected.name}}</span>
+              </ui-select-match>
+              <ui-select-choices repeat="val.value as val in GroupUpdateCtrl.colors | filter: {'name' :$select.search}">
+                <div>
+                  <span><i class="fa fa-circle" style="color: {{val.value}}; font-sze:14px"></i>&nbsp; <span translate>{{val.name}} </span> </span>
```
jniles, line 36: In the future, it would be better to use [a ngStyle tag](https://docs.angularjs.org/api/ng/directive/ngStyle) if you will interpolate on a style element.  Something like this:
```html
<i class="fa fa-circle" ng-style="{ color : val.value }">
```

jniles (approved):
LGTM!

bors r+

PR #2189, opened 2017-10-09
Some general fix and enhancements

DedrickEnc:
This set of commits fix and add some functionalities in the inventory module, in order to make it more clear.


jniles (approved):
LGTM.  We need to remember to write a SQL migration script for our installations when we next update.

bors r+

PR #2195, opened 2017-10-10
refractor(user module) password's strength validation and unique username

jeremielodi:
This PR provides a password's strength validation directive and notify the user if the current user name he has typed to create a new user is already used because it should be unique

### DIFF
--- a/client/src/js/components/bhStrengthMeter.js
+++ b/client/src/js/components/bhStrengthMeter.js
@@ -1,0 +1,21 @@
+angular.module('bhima.components')
+.component('bhStrengthMeter', {
+  template :
+    "<div class='strength-meter'>"+
+    " <div class='strength-meter-fill' data-strength='{{$ctrl.counter()}}'></div>"+
+    "</div>",
+  bindings : {
+    password : '<',
+  },
+  controller: StrengthMeterController,
+});
+
+StrengthMeterController.$inject = ['PasswordMeterService'];
+
+function StrengthMeterController(PasswordMeterService){
+
+  var $ctrl = this;
+  $ctrl.counter = function () {
+    return PasswordMeterService.counter($ctrl.password);
+  }
+}
--- a/client/src/js/directives/bhPasswordMeter.js
+++ b/client/src/js/directives/bhPasswordMeter.js
@@ -1,0 +1,33 @@
+angular.module('bhima.directives')
+.directive('bhPasswordMeter', PasswordValidatorDirective);
+
+PasswordValidatorDirective.$inject = ['PasswordMeterService'];
+
+/**
+ * Passord Validator Directive
+ *
+ * This directive provides a generic password validatation.
+ * the validation condition is defined at  PasswordMeterService
+ *
+ * @example
+ * <input name="pwd" type="passord" bh-password-meter>
+ *
+ * <div ng-messages="form.pwd.$error">
+ *   <p ng-messages-include="modules/templates/messages.tmpl.html"></p>
+ * </div>
+ *
+ * @module directives/PasswordValidatorDirective
+ */
+function PasswordValidatorDirective(PasswordMeterService) {
+  return {
+    restrict : 'A',
+    require : 'ngModel',
+
+    link : function passwordLink(scope, element, attrs, ctrl) {
+
+      ctrl.$validators.password = function (modelValue, viewValue) {
+        return PasswordMeterService.validate(viewValue);
+      };
+    }
+  };
+}
--- a/client/src/js/services/PasswordMeterService.js
+++ b/client/src/js/services/PasswordMeterService.js
@@ -1,0 +1,52 @@
+angular.module('bhima.services')
+.service('PasswordMeterService', PasswordMeterService);
+
+function PasswordMeterService() {
+  var service = this;
+  service.validate = validate;
+  service.counter = counter;
+
+  var strongRegularExp = new RegExp("^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#\$%\^&\*])(?=.{8,})");
+  var mediumRegularExp = new RegExp("^(((?=.*[a-z])(?=.*[A-Z]))|((?=.*[a-z])(?=.*[0-9]))|((?=.*[A-Z])(?=.*[0-9])))(?=.{6,})");
+
+  // both Strong and Weak password are accepted
+  function validate(viewValue) {
+
+    if (strongRegularExp.test(viewValue)) {
+      // Strong password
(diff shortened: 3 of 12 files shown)


client/src/js/components/bhStrengthMeter.js
```diff
@@ -0,0 +1,21 @@
+angular.module('bhima.components')
+.component('bhStrengthMeter', {
+  template :
+    "<div class='strength-meter'>"+
+    " <div class='strength-meter-fill' data-strength='{{$ctrl.counter()}}'></div>"+
+    "</div>",
+  bindings : {
+    password : '=',
```
jniles: Is there any reason for this to be a [two-way binding](https://docs.angularjs.org/guide/component#component-based-application-architecture)?  Can it be a one-way binding (`<`)?

client/src/js/directives/bhPasswordMeter.js
```diff
@@ -0,0 +1,33 @@
...
+
+PasswordValidatorDirective.$inject = ['PasswordMeterService'];
+
+/**
+ * Passord Validator Directive
+ *
+ * This directive provides a generic password validatation.
+ * the validation condition is defined at  PasswordMeterService
+ *
+ * @example
```
jniles, line 12: Nice comments!

client/src/js/services/PasswordMeterService.js
```diff
@@ -0,0 +1,54 @@
+angular.module('bhima.services')
+.service('PasswordMeterService', PasswordMeterService);
+
+PasswordMeterService.$inject = [];
```
jniles: If you don't inject anything, there is no need to put this empty array up here.

client/src/modules/users/user.modal.html
```diff
@@ -12,16 +12,28 @@
...
+  <div class="modal-body" style="overflow-y: scroll; max-height:600px; ">
+
 
     <div class="form-group" ng-class="{ 'has-error' : UserForm.$submitted && UserForm.display_name.$invalid }">
-      <label class="control-label" translate>FORM.LABELS.USERNAME</label>
-      <input name="display_name" ng-model="UserModalCtrl.user.display_name" autocomplete="off" ng-maxlength="UserModalCtrl.maxLength" class="form-control" required>
-      <div class="help-block" ng-messages="UserForm.display_name.$error" ng-show="UserForm.$submitted">
-        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      <label class="control-label col-md-3" translate>FORM.LABELS.USERNAME</label>
+      <div class="col-md-12">
```
jniles: Having a `col-md-3` and `col-md-12` on the same line doesn't make any sense ... bootstrap only has 12 columns.  Which do you mean to put here - `col-md-3` and `col-md-9`?

client/src/modules/users/user.modal.html
```diff
@@ -12,16 +12,28 @@
...
+
+
+        <div class="help-block" ng-messages="UserForm.display_name.$error" ng-show="UserForm.$submitted">
+          <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+        </div>
       </div>
     </div>
 
+    <ng-messages for="UserForm.display_name.$error" ng-if="UserForm.display_name.$dirty">
+      <ng-message when="uniqueUserName"> user already exists</ng-message>
```
jniles: This should be translated.

client/src/modules/users/user.modal.html
```diff
@@ -73,14 +85,21 @@
...
+        <input
+          name="password"
+          ng-model="UserModalCtrl.user.password"
+          class="form-control"
+          bh-password-meter
+          type="password" required>
 
-        <span ng-show="UserForm.password.$dirty && !UserModalCtrl.validPassword()" class="glyphicon glyphicon-warning-sign form-control-feedback text-danger" aria-hidden="true"></span>
-        <span ng-show="UserForm.password.$dirty && UserModalCtrl.validPassword()" class="glyphicon glyphicon-ok form-control-feedback text-success" aria-hidden="true"></span>
+          <bh-strength-meter password="UserForm.password.$viewValue"></bh-strength-meter>
```
jniles, line 93: :+1:

client/src/modules/users/user.modal.js
```diff
@@ -11,21 +14,22 @@ function UserModalController($state, Projects, Users, Notify, AppCache) {
   // the user object that is either edited or created
   vm.user = {};
   vm.stateParams = {};
+  vm.passwordValidationStrength = '';
```
jniles: You never use this - it can be removed.

client/src/modules/users/user.modal.js
```diff
@@ -11,21 +14,22 @@ function UserModalController($state, Projects, Users, Notify, AppCache) {
...
-
-
   vm.isCreating = vm.stateParams.creating;
 
-  //exposed methods
-  vm.submit = submit;
-  vm.closeModal = closeModal;
-  vm.validPassword = validPassword;
-  vm.editPassword = editPassword;
+  vm.passwordValidationCount = -1;
```
jniles: It doesn't look like you use this either.

server/config/routes.js
```diff
@@ -438,6 +438,7 @@ exports.configure = function configure(app) {
   app.get('/users', users.list);
   app.post('/users', users.create);
   app.get('/users/:id', users.detail);
+  app.get('/users/:username/exists', users.exists);
```
jniles, line 441: :+1:

server/controllers/admin/users/index.js
```diff
@@ -124,6 +125,17 @@ function detail(req, res, next) {
   .done();
 }
 
+function exists(req, res, next) {
+  const sql='SELECT count(id) as nbr FROM user WHERE username=?';
```
jniles, line 129: A more efficient SQL query might be `SELECT true FROM user WHERE username = ?`.

jniles (commented):
This is really cool!  There are a few clean up things to take care of.

client/src/modules/users/user.modal.html
jeremielodi: the curent disposition of the form is : the input is underneath the lable that is why it should have this class 'col-md-12', I'm change the label class col-md-3 to col-md-12 too.
 The 'col-md-12' class in the username input help to set the seach icon visible

jniles: It doesn't work without the `col-md-12`?  That's interesting.

jeremielodi: It works but the seach icon position is bad

jniles:
LGTM.  bors r+

PR #2202, opened 2017-10-15
refactor: use bhConstants.defaultFilters

### LINKED ISSUE #2133: Use bhConstants.defaultFilters everywhere that use/assign default filters
In https://github.com/IMA-WorldHealth/bhima-2.X/issues/1988, we proposed putting the default filters in `bhConstants`.  This was accomplished in https://github.com/IMA-WorldHealth/bhima-2.X/commit/2e1f72fe875fcc6d1d02e1d501b61a458ddbb932, and they are now available in [bhConstants](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/constants/bhConstants.js#L115).  The registries that assign default filters should all be updated to use this object.

jniles:
This commit replaces all instances of hand-defined default filters with
the shared instance found in `bhConstants`.  This allows a significant
reduction in code for all services that use the FilterService to limit
records.

Additionally, a few typos were fixed in the affected services, as well
as pruning functions that are never called.  Where appropriate, further
documenting comments were added.

Closes #2133.

### DIFF
--- a/client/src/modules/patients/patients.service.js
+++ b/client/src/modules/patients/patients.service.js
@@ -1,9 +1,10 @@
 angular.module('bhima.services')
-.service('PatientService', PatientService);
+  .service('PatientService', PatientService);
 
 PatientService.$inject = [
-  '$http', 'util', 'SessionService', '$uibModal',
-  'DocumentService', 'VisitService', 'FilterService', 'appcache', 'PeriodService', 'PrototypeApiService', '$httpParamSerializer', 'LanguageService'
+  'SessionService', '$uibModal', 'DocumentService', 'VisitService',
+  'FilterService', 'appcache', 'PeriodService', 'PrototypeApiService',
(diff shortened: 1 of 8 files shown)


jniles:
@dedrickenc, could I get a review?

client/src/modules/patients/patients.service.js
```diff
@@ -151,17 +137,17 @@ function PatientService($http, util, Session, $uibModal,
...
 
-    return $http.post(path, options)
-      .then(util.unwrapHttpResponse);
+    return service.$http.post(path, options)
+      .then(service.util.unwrapHttpResponse);
   }
 
   function searchByName(options) {
-    options = angular.copy(options || {});
+    var opts = angular.copy(options || {});
```
DedrickEnc, line 145: Nice!

DedrickEnc (approved):
LGTM

PR #2203, opened 2017-10-15
refactor(reports): remove browser.refresh()

### LINKED ISSUE #1884: Fixing the report e2e test
Currently, we are supposed to refresh the page before every test case to make our report e2e test passing, since it is not correct logically (because the user will not refresh the page after every use case). 
These ~`browser.pause()`~ `browser.refresh()` should be removed and the test should be updated for best practice.

jniles:
This commit removes all browser.refresh() calls in the reports end to end tests.  This ensures that the tests behave consistently and do not need a reset.  It also fractionally decreases the execution time since the browser does not need to load and parse all the JS code and recompute layouts multiple times.

Closes #1884.

### DIFF
--- a/test/end-to-end/reports/open_debtors/open_debtors.spec.js
+++ b/test/end-to-end/reports/open_debtors/open_debtors.spec.js
@@ -1,5 +1,3 @@
-/* global browser */
-
 const chai = require('chai');
 const helpers = require('../../shared/helpers');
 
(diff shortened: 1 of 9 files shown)


jniles:
@mbayopanda, could I get a review?

test/end-to-end/reports/open_debtors/open_debtors.spec.js
```diff
@@ -20,7 +18,6 @@ describe('Open Debtors Report', () => {
   before(() => {
     helpers.navigate(`#!/reports/${key}`);
     Page = new ReportOpenDebtorsPage(key);
-    browser.refresh();
```
mbayopanda, line 23: 👍

mbayopanda (approved):
LGTM

PR #2210, opened 2017-10-16
feat(client report) debtor accounts balance report

### LINKED ISSUE #636: (design) Client Report
This issue concerns the design of the client report.  It should look like the mockup below.

client report
_Fig 1: Client Report_

The data from this report should only come from the `period_total` table.  The SQL statement will look up the debtor group accounts and retrieve the opening balances from the period 0 of the `period_total` table.  It will then sum the `period_total` grouped by account_id where the fiscal year is 
the selected fiscal year.

As in 1.x, the client report could have two versions: a "detailed" version that shows debits/credits for opening balances and a "plain" version that only includes the combined balance.  The

jeremielodi:
This PR add  a new client report as needed in this issue #636 

closes #636


server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,76 @@
...
+ * @param {Object} params Parameters passed in to customise the report - these
+ *                        are usually passed in through the query string
+ * @description
+ * The HTTP interface which actually creates the report.
+ */
+function queryContext(fiscalYearId) {
+
+  // aggregates the data above as totals into columns of 30, 60, 90, and >90
+  const sql = `
+    SELECT ac.id as Account, ac.label as Name, p.number as period,
```
jniles: Please try to avoid doing changes in capitalization where you can ... database behave strangely with uppercase/lowercase.  [Here is an interesting answer exploring the quirks in database behavior with uppercase and lowercase letters](https://stackoverflow.com/a/12077352/2496633).

A good rule to follow (which is enforced by ESLint) is to:
 1. Use `UPPER_CASE` only for global constants that will never change.  This practice is inspired by C programmers.
 2. Use `CapitalLettersOnFunctions` only if that function/object is a class.  Classes are things you can call `new` on.
 3. Use `lowercase` for everything else.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,76 @@
...
+      SUM(pt.debit) as Debit, SUM(pt.credit) as Credit,
+      SUM(IF(p.number = 0,1,0))  as OpeningBalance,
+      SUM(IF(p.number <> 0, (pt.debit-pt.credit) ,0))  as closingBalance
+    FROM  period_total  pt
+    JOIN debtor_group d ON pt.account_id = d.account_id
+    JOIN account ac  ON ac.id = pt.account_id
+    JOIN fiscal_year f ON f.id = pt.fiscal_year_id
+    JOIN period p ON p.id = pt.period_id
+    WHERE pt.fiscal_year_id = ${fiscalYearId}
```
jniles: This is vulnerable to an SQL injection attack.  Please use `?` to escape the query.  Remember that the syntax for `db.exec()` is `db.exec(sql, params)`.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * @method agedDebtorReport
+ *
+ * @description
+ * The HTTP interface which actually creates the report.
+ */
+function debtorAccountValence(req, res, next) {
+  const options = _.extend(req.query, {
+    filename : 'TREE.GENERAL_LEDGER',
```
jniles: This filename should not be `GENERAL_LEDGER`.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,76 @@
...
+  } catch (e) {
+    return next(e);
+  }
+
+  // fire the SQL for the report
+
+  const fiscalYearId = req.params.fiscalYearId;
+  queryContext(fiscalYearId)
+  .then(
+    function(data) {
```
jniles: This will probably give you a ESLint error - do you have ESLint installed?

Since you already defined `let data` above, you cannot use it again in the same scope.

A better way to do this would be
```js
.then(data => report.render({ rows : data }))
.then(result => res.set(result.headers).send(result.report))
.catch(next)
.done();
```

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,76 @@
+const _ = require('lodash');
+const ReportManager = require('../../../../lib/ReportManager');
+const db = require('../../../../lib/db');
+
+// path to the template to render
+const TEMPLATE = './server/controllers/finance/reports/debtors/balance.handlebars';
+
+exports.debtorAccountValence = debtorAccountValence;
```
jniles: What is `Valence`?

server/controllers/finance/reports/debtors/balance.handlebars
```diff
@@ -0,0 +1,47 @@
...
+          </tr>
+        </thead>
+        <tbody>
+
+          <!-- print a row for each debtor -->
+          {{#each rows}}
+            <tr>
+              <td>{{this.Account}}</td>
+              <td>{{this.Name}}</td>
+              <td>{{this.OpeningBalance}}</td>
```
jniles: Balances should always be expressed with a currency, and be right-aligned.  Like this:
```hbs
<td class="text-right">{{ currency this.openingBalance metadata.enterprise.currency_id }}</td>
```
In this case, the `currency` is the currency helper, the `this.openingBalance` is the number and `metadata.enterprise.currency_id` tells the handlebars renderer what currency to use.  It should always use the enterprise currency.

client/src/modules/reports/generate/debtorBalanceReport/debtorBalanceReport.js
```diff
@@ -0,0 +1,53 @@
...
+  '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'BaseReportService', 'reportData',
+];
+
+function DebtorBalanceReportController($sce, Notify, BaseReportService, AppCache, SavedReports, reportData) {
+  var vm = this;
+  var cache = new AppCache('configure_debtorAccountBalence');
+  var reportUrl = 'debtorAccountBalence';
+
+  vm.reportDetails = {};
+  vm.currentFiscalYear = 0;
```
jniles: Do you need to set this property?

server/config/routes.js
```diff
@@ -84,6 +84,8 @@ const financialPatient = require('../controllers/finance/patient');
 const dashboardDebtors = require('../controllers/dashboard/debtorGroups');
 const stats = require('../controllers/dashboard/stats');
 
+var reportDebutor = require('../controllers/finance/reports/debtors/debtorAccountBalance');
```
jniles: This should probably be called `reportDebtor`.

client/src/modules/reports/generate/debtorBalanceReport/debtorBalanceReport.js
```diff
@@ -0,0 +1,56 @@
...
+  vm.reportDetails = {};
+  vm.currentFiscalYear = {
+    label : 'Test Fiscal Year 2015',
+  };
+
+  var _url = '';
+
+  vm.onSelectFiscal = function (fiscal) {
+    vm.currentFiscalYear.id = fiscal;
+    _url = reportUrl + '/' + vm.currentFiscalYear.id + '/'+ vm.currentFiscalYear.label;
```
jniles: Why do you need to send back the label?

client/src/modules/reports/generate/debtorBalanceReport/debtorBalanceReport.html
```diff
@@ -0,0 +1,45 @@
...
+      <div class="panel-heading">
+        <span translate>REPORT.UTIL.OPTIONS</span>
+      </div>
+
+      <div class="panel-body">
+
+      <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate autocomplete="off">
+
+        <!--select the period by a date-->
+        <bh-fiscal-period-select
```
jniles: As discussed, this should only have the fiscal year.

jniles (changes requested):
@jeremielodi, this is a good first attempt.

The SQL injection attack vectors are really important to catch before code gets submitted for a review.  Could you try to review all your SQL before submitting a PR to make sure there are using `db.exec()` properly?  It would save  a lot of time.

Most reports use query parameters to send back their data - could you use that here?  That way you can take advantage of some of our testing tools instead of having to write your own.

This PR will need tests written for it on both the client and the server.  The server integration tests can use [the LazyTester](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/integration/reports/rendering.js#L8) module if you wish.  The end to end tests should build their own report test like [the cashflow report test](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/reports/cashflow/cashflow.spec.js).

client/src/modules/reports/generate/debtorBalanceReport/debtorBalanceReport.js
jeremielodi: It should appear in the report, as I can already have it from  a fiscal year component. I don't think it's important to execute another query to get the fiscal year label from it id

server/config/routes.js
```diff
@@ -366,6 +368,7 @@ exports.configure = function configure(app) {
   app.get('/reports/archive/:uuid', report.sendArchived);
   app.delete('/reports/archive/:uuid', report.deleteArchived);
 
+  app.get('/debtorAccountBalence/:fiscalYearId', reportDebtor.debtorAccountBalance);
```
jniles: This url should read `debtorAccountBalance`.

Almost all of our reports use the url `reports/....`.  Could you use that style of URL?  It will make things easier to debug when bugs occur.

server/controllers/finance/reports/debtors/balance.handlebars
```diff
@@ -0,0 +1,47 @@
...
+    <h3 class="text-center text-uppercase">
+      <strong>{{translate "REPORT.CLIENT_DEBTOR_ACCOUNT_BALANCE_REPORT"}}</strong>
+    </h3>
+
+    <h5 class="text-center">
+      <strong>
+        <span class="text-capitalize"> {{ fiscalLabel }}</span>
+      </strong>
+    </h5>
+    <!-- margin is the cell size -->
```
jniles: What does this cell size comment mean?

server/controllers/finance/reports/debtors/balance.handlebars
```diff
@@ -0,0 +1,47 @@
...
+            <th>{{translate "TABLE.COLUMNS.NAME"}}</th>
+            <th>{{translate "REPORT.OPENING_BALANCE"}}</th>
+            <th>{{translate "TABLE.COLUMNS.DEBIT"}}</th>
+            <th>{{translate "TABLE.COLUMNS.CREDIT"}}</th>
+            <th>{{translate "REPORT.CLOSING_BALANCE"}}</th>
+          </tr>
+        </thead>
+        <tbody>
+
+          <!-- print a row for each debtor -->
```
jniles: :+1:

server/controllers/finance/reports/debtors/balance.handlebars
```diff
@@ -0,0 +1,47 @@
...
+          {{#each rows}}
+            <tr>
+              <td>{{this.account}}</td>
+              <td>{{this.name}}</td>
+              <td class="text-right">{{currency this.openingBalance metadata.enterprise.currency_id}}</td>
+              <td class="text-right">{{currency this.cebit  metadata.enterprise.currency_id}}</td>
```
jniles: I think this should say "debit"

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,91 @@
...
+      SUM(IF(p.number <> 0, (pt.debit-pt.credit) ,0))  as closingBalance
+    FROM  period_total  pt
+    JOIN debtor_group d ON pt.account_id = d.account_id
+    JOIN account ac  ON ac.id = pt.account_id
+    JOIN fiscal_year f ON f.id = pt.fiscal_year_id
+    JOIN period p ON p.id = pt.period_id
+    WHERE pt.fiscal_year_id = ?
+    GROUP BY pt.account_id
+  `;
+  return db.exec(sql, fiscalYearId);
```
jniles, line 79: Much better!

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function fiscalyerQuery (id){
```
jniles: This should probably be spelled `fiscalYearQuery`.

One of the most well known computer scientists, [Donald Knuth](https://en.wikipedia.org/wiki/Donald_Knuth), once remarked:
> Programs are meant to be read by humans and only incidentally for computers to execute.

This conveys the idea that your code should always be high quality for others to read, modify, and extend.  That is the reason we tend to be pretty strict on typos, even if the functionality is not affected by the typo.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,90 @@
...
+      .done();
+  })
+  .catch(next)
+  .done();
+}
+
+/**
+ * @method queryContext
+ *
+ * @param {Object} params Parameters passed in to customise the report - these
```
jniles: These comments should probably talk about `fiscalYearId` and not `params`, which isn't defined.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,90 @@
...
+    JOIN account ac  ON ac.id = pt.account_id
+    JOIN fiscal_year f ON f.id = pt.fiscal_year_id
+    JOIN period p ON p.id = pt.period_id
+    WHERE pt.fiscal_year_id = ?
+    GROUP BY pt.account_id
+  `;
+  return db.exec(sql, fiscalYearId);
+}
+
+function fiscalYearQuery (id){
```
jniles: Make sure this respects eslint!  I'm pretty sure this will throw an eslint error.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,90 @@
...
+ * The HTTP interface which actually creates the report.
+ */
+function queryContext(fiscalYearId) {
+
+  // aggregates the data above as totals into columns of 30, 60, 90, and >90
+  const sql = `
+    SELECT ac.id as account, ac.label as name, p.number as period,
+      SUM(pt.debit) as debit, SUM(pt.credit) as credit,
+      SUM(IF(p.number = 0,1,0))  as openingBalance,
+      SUM(IF(p.number <> 0, (pt.debit-pt.credit) ,0))  as closingBalance
```
jniles: Did you test this?

It looks like the closing balance should be `SUM(pt.debit) - SUM(pt.credit)` instead of `pt.debit - pt.credit`.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,90 @@
...
+ * @method queryContext
+ *
+ * @param {Object} params Parameters passed in to customise the report - these
+ *                        are usually passed in through the query string
+ * @description
+ * The HTTP interface which actually creates the report.
+ */
+function queryContext(fiscalYearId) {
+
+  // aggregates the data above as totals into columns of 30, 60, 90, and >90
```
jniles: This comment should be removed.  It is not about the query below.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,90 @@
...
+    JOIN period p ON p.id = pt.period_id
+    WHERE pt.fiscal_year_id = ?
+    GROUP BY pt.account_id
+  `;
+  return db.exec(sql, fiscalYearId);
+}
+
+function fiscalYearQuery (id){
+  const sql =`SELECT label FROM fiscal_year WHERE id =?`
+  return db.exec(sql, id);
```
jniles: If you use `db.one()`, this would return the first row as an object.  It would clean up your report code a bit by removing the need to check for `length`.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,90 @@
...
+
+  const fiscalYearId = req.params.fiscalYearId;
+  let label = '';
+  queryContext(fiscalYearId)
+  .then(
+    function(data) {
+
+      fiscalYearQuery(fiscalYearId).then(function (response){
+
+        if(response.length>0){
```
jniles: If you use `db.one()` in the `fiscalYearQuery`, you could then just do this:
```js
fiscalYearQuery(fiscalYearId)
  .then(fiscalYear => {
   const label = fiscalYear.label;
   // .. etc ...
  });
```

server/controllers/finance/reports/debtors/debtorAccountBalance.js
```diff
@@ -0,0 +1,90 @@
...
+        var _data = {rows : data, fiscalLabel : label };
+        report.render(_data)
+        .then(result => {
+          res.set(result.headers).send(result.report);
+        })
+      })
+      .catch(next)
+      .done();
+  })
+  .catch(next)
```
jniles: Your error handling would be better if you made this into a promise chain.  Like this:
```js
const data = {};
queryContext(fiscalYearId)
  .then(rows => {
    _.extend(data, { rows });
   return fiscalYearQuery(fiscalYearId);
  })
  . then(fiscalYear => {
    // pretend we used db.one() above.
    const fiscalLabel = fiscalYear.label;
    _.extend(data, { fiscalLabel });
    return report.render(data);
  })
  .then(result => {
    res.set(result.headers).send(result.report);
  })
  .catch(next)
  .done();
```
In this way, you make sure that each function has its own context, and the only shared context is shared via the `data` object.  Additionally, there is only one error handler, which will always be fired if an error occurs.

jniles (changes requested):
This is getting better!

Please make sure you have eslint installed and your code passes eslint.  It will give you lots of help cleaning your code up.  If you are confused about a rule, look it up [on the eslint site](https://eslint.org/docs/rules/).

I've proposed some modifications to clean up your promises and error handling.  Additionally, I don't think your closing balance calculation is correct.

server/controllers/finance/reports/debtors/debtorAccountBalance.js
jeremielodi: It was tested ,  if I try this : SUM(IF(p.number <> 0, ( SUM(pt.debit) - SUM(pt.credit) ) ,0)) ,an error occurres

jeremielodi: Note that we are summing in a certain condition, try the query if you have data and tell me if the result is correct. It's good for me

client/src/js/components/bhFiscalYearSelect.js
```diff
@@ -0,0 +1,27 @@
+angular.module('bhima.components')
+.component('bhFiscalYearSelect', {
+    bindings: {
+      validationTrigger: '<',
+      onSelectFiscalCallback: '&?',
+      formName: '@',
```
jniles: It doesn't look like you ever used `formName`.  Go ahead and remove it.

client/src/js/components/bhFiscalYearSelect.js
```diff
@@ -0,0 +1,27 @@
+angular.module('bhima.components')
+.component('bhFiscalYearSelect', {
+    bindings: {
+      validationTrigger: '<',
```
jniles, line 4: It would be ideal to include `fiscalYearId` in these bindings.  That way other modules can set the `fiscalYearId` if they already have it.

This would allow you to set the fiscal year to the current fiscal year by default.  You should do that in your report controller though, not in the component controller.

client/src/js/components/bhFiscalYearSelect.js
```diff
@@ -0,0 +1,27 @@
...
+      onSelectFiscalCallback: '&?',
```
jniles: I think this should be required.  (Remove the `?`).

client/src/modules/reports/generate/debtorBalanceReport/debtorBalanceReport.js
```diff
@@ -0,0 +1,53 @@
...
+.controller('debtorBalanceReportController', DebtorBalanceReportController);
+
+DebtorBalanceReportController.$inject = [
+  '$state', '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'BaseReportService', 'reportData',
+];
+
+function DebtorBalanceReportController($state, $sce, Notify, BaseReportService, AppCache, SavedReports, reportData) {
+  var vm = this;
+  var cache = new AppCache('configure_debtorAccountBalence');
+  vm.reportUrl = 'reports/debtorAccountBalence';
```
jniles: No need to have this attached to `vm`.  It can just be `var reportUrl`.  Attaching it to `vm` makes me think it will appear somewhere in the view, but when I looked, I couldn't find it.

server/config/routes.js
```diff
@@ -363,8 +365,8 @@ exports.configure = function configure(app) {
   app.get('/reports/archive/:uuid', report.sendArchived);
   app.delete('/reports/archive/:uuid', report.deleteArchived);
 
+  app.get('/reports/debtorAccountBalence/:fiscalYearId', reportDebtor.debtorAccountBalance);
```
jniles: This url should say `Balance`.

server/controllers/finance/reports/debtors/balance.handlebars
```diff
@@ -0,0 +1,46 @@
+{{> head title="REPORT.AGED_DEBTORS.TITLE"}}
```
jniles: I think this should be `CLIENT_DEBTOR_ACCOUNT_BALANCE_REPORT` and not `AGED_DEBTORS`.

test/data.sql
```diff
@@ -462,3 +465,10 @@ INSERT INTO `stock_movement` (`uuid`, `lot_uuid`, `document_uuid`, `depot_uuid`,
   (HUID('d8c83ad9-a3ea-4f9f-96f9-456a435f480d'), HUID('ef24cf1a-d5b9-4846-b70c-520e601c1ea6'), HUID('682e11c0-93a7-49f8-b79b-a4bc8e3e6f47'), HUID('f9caeb16-1684-43c5-a6c4-47dbac1df296'), '', 1, '2017-02-02', 50, 2.0000, 0, 1),
   (HUID('f9aa33f1-65e2-4e37-89cb-843d27b2c586'), HUID('064ab1d9-5246-4402-ae8a-958fcdb07b35'), HUID('682e11c0-93a7-49f8-b79b-a4bc8e3e6f47'), HUID('f9caeb16-1684-43c5-a6c4-47dbac1df296'), '', 1, '2017-02-02', 100, 1.2000, 0, 1),
   (HUID('e8502c3e-7483-11e7-a8de-507b9dd6de91'), HUID('064ab1d9-5246-4402-ae8a-958fcdb07b35'), HUID('0cc6c435-7484-11e7-a8de-507b9dd6de91'), HUID('f9caeb16-1684-43c5-a6c4-47dbac1df296'), HUID('d4bb1452-e4fa-4742-a281-814140246877'), 8, '2017-02-02', 75, 1.2000, 1, 1);
+
+
+-- admin user width all permissions
+
+INSERT INTO permission (unit_id, user_id)
```
jniles, line 309: :+1:  Could you delete all the permissions for user 1 above?  That way, no one will be confused about trying to add/remove permissions in the future and it not taking any effect.

test/end-to-end/shared/components/index.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
   userSelect     : require('./bhUserSelect'),
   reportSource   : require('./bhReportSource'),
   fiscalPeriodSelect          : require('./bhFiscalPeriodSelect'),
+  fiscalYearSelect          : require('./bhFiscalYearSelect'),
```
jniles, line 27: :+1:

test/end-to-end/shared/components/bhFiscalYearSelect.js
```diff
@@ -0,0 +1,19 @@
...
+module.exports = {
+  mainSelector : '[bh-fiscal-year-select]',
+  set : function set(fiscal_year_id, id) {
+      var opts;
+    /** return selector for an option element */
+    function getValue(id) {
+      return by.css('option[value="number:?"]'.replace('?', id));
+    }
+
+    const bhFiscalPeriod = (id) ? element(by.id(id)) : element(by.css(this.mainSelector));
```
jniles: This should just be called `bhFiscal` - there are no more periods here.

test/end-to-end/reports/debtor_account_balance/balanceReport.spec.js
```diff
@@ -0,0 +1,46 @@
...
+helpers.configure(chai);
+
+const DebtorClientAccountBalanceReportPage = require('./balance.page');
+
+describe('Debtor Clients Account Balance Report', () => {
+  let Page;
+  const key = 'debtorBalanceReport';
+
+  const dataset = {
+    fiscal_id : 2,
```
jniles, line 15: The goals of an end-to-end test is to try and mimic the behavior of a user.  A user will not select a fiscal year by the id - they will select it by text.  Try putting the text here and updating your `bhFiscalSelect` component to handle that case.

test/end-to-end/shared/components/bhFiscalYearSelect.js
```diff
@@ -0,0 +1,19 @@
+/* global browser, element, by */
+
+const FU = require('../FormUtils');
+
+module.exports = {
+  mainSelector : '[bh-fiscal-year-select]',
+  set : function set(fiscal_year_id, id) {
+      var opts;
+    /** return selector for an option element */
+    function getValue(id) {
```
jniles, line 10: Just FYI - you can use `FormUtils` to make this all one line.  The method you are looking for is [select()](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/shared/FormUtils.js#L82).

```js
set: function (fiscalYearText, id) {
  const locator = id ? by.id(id) : by.css(this.mainSelector);
  FU.select('$ctrl.selectedFiscal', fiscalYearText, element(locator));
};
```

jniles (commented):
@jeremielodi, this is shaping up nicely.  Hopefully this will be the last round of reviews.

jniles:

jniles:
I'm going to fork this report and see if we can get it landed in master.

jeremielodi:
Ok

PR #2211, opened 2017-10-17
refactor(Stock Registry)

lomamech:
Make stock registry link Search Readable
- in Stock Inventory, Stock lots and Stock Registry
- Implement component bh-flux-select for Inventory Reason
- Implement test E2E for bh-flux-select

### DIFF
--- a/client/src/js/components/bhFluxSelect.js
+++ b/client/src/js/components/bhFluxSelect.js
@@ -4,7 +4,7 @@ angular.module('bhima.components')
     controller  : FluxSelectController,
     bindings    : {
       onChange : '&',
-      fluxId : '<?',
+      fluxIds : '<?',
       label : '@?',
       required : '<?',
       validationTrigger : '<?',
@@ -30,7 +30,7 @@ function FluxSelectController(Flux, Notify, $translate) {
     $ctrl.onChange = $ctrl.onChange || angular.noop;
 
     // init the model
-    $ctrl.selectedFlux = $ctrl.fluxId || [];
+    $ctrl.selectedFlux = $ctrl.fluxIds || [];
 
     // load all Flux
     Flux.read()
--- a/client/src/js/components/bhFluxSelect.js
+++ b/client/src/js/components/bhFluxSelect.js
@@ -1,0 +1,50 @@
+angular.module('bhima.components')
+  .component('bhFluxSelect', {
+    templateUrl : 'modules/templates/bhFluxSelect.tmpl.html',
+    controller  : FluxSelectController,
+    bindings    : {
+      onChange : '&',
+      fluxId : '<?',
+      label : '@?',
+      required : '<?',
+      validationTrigger : '<?',
+    },
+  });
+
+FluxSelectController.$inject = [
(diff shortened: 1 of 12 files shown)


client/src/modules/stock/inventories/modals/search.modal.js
```diff
@@ -60,10 +73,27 @@ function SearchInventoriesModalController(data, Notify, Instance, Store, Periods
   vm.cancel = function cancel() { Instance.close(); };
 
   vm.submit = function submit() {
+    // Set the label of status Value
+    if (vm.searchQueries.status) {
```
jniles, line 77: I can imagine that other developers will want to use this too.

Why don't you make an object map and put it in a service (like the `StockService`)?  That way, you could do this:
```js
// this is in a stock service
service.statusLabelMap = {
  'sold_out' : 'STOCK.STATUS.SOLD_OUT',
  // ... etc ...
};

// this is in your submit() function.
if (vm.searchQueries.status) {
  displayValues.status = service.statusLabelMap[vm.searchQueries.status];
}
```
That way, if multiple controllers need to have the status --> i18n key mapping, they can just import your service!

client/src/js/components/bhFluxSelect.js
```diff
@@ -0,0 +1,50 @@
+angular.module('bhima.components')
+  .component('bhFluxSelect', {
+    templateUrl : 'modules/templates/bhFluxSelect.tmpl.html',
+    controller  : FluxSelectController,
+    bindings    : {
+      onChange : '&',
+      fluxId : '<?',
```
jniles: It looks like this is a `ui-select multiple`.  Should this name be `fluxIds` then (in plural)?  Since it could take multiple flux ids?

jniles (commented):
Just a few comments here.

client/src/modules/stock/inventories/modals/search.modal.js
lomamech: Sorry for the few Comment, some times when I update my branch after resolving a conflict, I lose the message of my former Commit

jniles:
@lomamech, let me know when you need another review.

lomamech:
Yes, I need another review

client/src/js/services/StockService.js
```diff
@@ -219,6 +219,20 @@ function StockService(Api, Filters, AppCache, Periods, $httpParamSerializer, Lan
     }, []);
   }
 
+    /** Get label for purchase Status */
+  function statusLabelMap(status) {
```
jniles, line 223: :+1:

client/src/modules/stock/inventories/modals/search.modal.js
```diff
@@ -60,10 +73,17 @@ function SearchInventoriesModalController(data, Notify, Instance, Store, Periods
...
+      displayValues.status = Stock.statusLabelMap(vm.searchQueries.status);
```
jniles, line 78: Beautiful!

jniles (approved):
LGTM!

bors r+

PR #2212, opened 2017-10-17
feature(Implement Link vouchers from Invoices and Cash Payments)

### LINKED ISSUE #2146: Link Vouchers from Invoices and Cash Payments
On both the Cash Payments and Invoices Registries, we link the patient and either the payments against the invoice (for the Cash Payment Registry) or the payments (for the Invoice Registry).

An additional enhancement to this would be linking vouchers from both.  This is useful because vouchers form a kind of payment, particularly for Conventions.  In this way, if a patient pays in a voucher, it will be linked from both their invoices and their record.

lomamech:
- Update drop down On Invoice and Cash Payment Registry, add voucher
  Registry link
- Update API for find voucher and filter for invoice_uuid and cash_uuid

closes #2146

### DIFF
--- a/client/src/modules/cash/payments/templates/action.cell.html
+++ b/client/src/modules/cash/payments/templates/action.cell.html
@@ -21,12 +21,12 @@
       </a>
     </li>
     <li ng-show="!row.entity.is_caution">
-      <a data-method="view-invoice"  href ui-sref="invoiceRegistry({ filters : { key : 'cash_uuid', value : row.entity.uuid, displayValue: row.entity.reference }})">
+      <a data-method="view-invoice" href ui-sref="invoiceRegistry({ filters : { key : 'cash_uuid', value : row.entity.uuid, displayValue: row.entity.reference }})">
         <span class="fa fa-file-text-o"></span> <span translate>REPORT.VIEW_INVOICE</span>
       </a>
     </li>
-    <li ng-show="!row.entity.is_caution">
-      <a data-method="view-invoice"  href ui-sref="vouchers({ filters : [{ key : 'cash_uuid', value : row.entity.uuid, displayValue: row.entity.reference }]})">
+    <li>
+      <a data-method="view-voucher" href ui-sref="vouchers({ filters : [{ key : 'cash_uuid', value : row.entity.uuid, displayValue: row.entity.reference }]})">
         <span class="fa fa-file-text-o"></span> <span translate>REPORT.VIEW_VOUCHER</span>
       </a>
     </li>
--- a/client/src/modules/cash/payments/templates/action.cell.html
+++ b/client/src/modules/cash/payments/templates/action.cell.html
@@ -21,10 +21,16 @@
       </a>
     </li>
     <li ng-show="!row.entity.is_caution">
-      <a data-method="view-invoice"  href ui-sref="invoiceRegistry({ filters : { key : 'cash_uuid', value : row.entity.uuid }})">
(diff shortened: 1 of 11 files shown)


client/src/modules/invoices/registry/templates/action.cell.html
```diff
@@ -22,11 +22,17 @@
       </a>
     </li>
     <li>
-      <a data-method="view-payment" href ui-sref="cashRegistry({ filters : { key : 'invoice_uuid', value : row.entity.uuid }})">
+      <a data-method="view-payment" href ui-sref="cashRegistry({ filters : { key : 'invoice_uuid', value : row.entity.uuid, displayValue: row.entity.reference }})">
         <span class="fa fa-money"></span> <span translate>REPORT.VIEW_PAYMENTS</span>
       </a>
     </li>
     <li>
+      <a data-method="view-invoice"  href ui-sref="vouchers({ filters : [{ key : 'invoice_uuid', value : row.entity.uuid, displayValue: row.entity.reference }]})">
```
jniles: I think this should be 'view-voucher'.

client/src/modules/cash/payments/templates/action.cell.html
```diff
@@ -21,10 +21,16 @@
       </a>
     </li>
     <li ng-show="!row.entity.is_caution">
-      <a data-method="view-invoice"  href ui-sref="invoiceRegistry({ filters : { key : 'cash_uuid', value : row.entity.uuid }})">
+      <a data-method="view-invoice"  href ui-sref="invoiceRegistry({ filters : { key : 'cash_uuid', value : row.entity.uuid, displayValue: row.entity.reference }})">
         <span class="fa fa-file-text-o"></span> <span translate>REPORT.VIEW_INVOICE</span>
       </a>
     </li>
+    <li ng-show="!row.entity.is_caution">
+      <a data-method="view-invoice"  href ui-sref="vouchers({ filters : [{ key : 'cash_uuid', value : row.entity.uuid, displayValue: row.entity.reference }]})">
```
jniles: This `data-method` should be 'view-voucher'.

jniles (commented):
@lomamech, just a couple clean up items.

Could you implement an end to end test for these?  I've added some tools to make it much easier (and safer) to write end to end tests.  You can see an example [here](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/vouchers/vouchers.spec.js#L21).

Basically, you should pick an invoice with a credit note, and write something like this:
```js
 // this is an example invoice with a credit note
const row = new GridRow('IV.TPA.4');

// this line opens the drop down menu on that grid row.
row.dropdown().click(); 

// this line you will have to implement.
row.goToVoucher().click();
```

The `goToVoucher()` method should be a method on [GridRow](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/shared/GridRow.js#L37) and look like this:
```js
goToVoucher() {
  return this.menu.$('[data-method="view-voucher"]');
}
```

The same thing can go for payments and that link.

test/end-to-end/shared/GridRow.js
```diff
@@ -49,6 +49,11 @@ class GridRow {
   openReverseReceipt() {
     return this.menu.$('[data-method="reverse-receipt"]');
   }
+
+  goToVoucher() {
+    return this.menu.$('[data-method="view-voucher"]');
```
jniles, line 54: :+1:

jniles (approved):
LGTM.

bors r+

PR #2215, opened 2017-10-18
Fixing functions bugs and some refactoring

DedrickEnc:
This P.R fixes the stock entry module to improve the user experience, here are some consequences about tasks performed : 

- Now the grid lines are updating properly when switching from one purchase order to an other

- Removing unnecessary ```self``` syntax and replace it with ```this```

- Stock entry is loading only consumable items

- ```StockItemService``` is cleaned

- The checking lot code is synplified

- The stock transfer feature is refactored.

### DIFF
--- a/client/src/modules/stock/StockForm.service.js
+++ b/client/src/modules/stock/StockForm.service.js
@@ -1,5 +1,5 @@
 angular.module('bhima.services')
-.service('StockFormService', StockFormService);
+  .service('StockFormService', StockFormService);
 
 StockFormService.$inject = [
   'StockItemService', 'Store', 'AppCache', 'SessionService',
@@ -36,10 +36,10 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
    * two lines are always present in the form.
    */
   StockForm.prototype.setup = function setup() {
-    var self = this;
-    self.details = {};
-    self.details.date = new Date();
-    self.details.user_id = Session.user.id;
+    this.details = {
+      date: new Date(),
+      user_id: Session.user.id
+    };
   };
 
   /**
@@ -52,12 +52,15 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
    */
   StockForm.prototype.addItems = function addItems(n) {
     var elt;
-    var idx;
-    var self = this;
+    // var idx;
+    // var self = this;
+
+    // will repeat will n > 0
     while (n--) {
       elt = new StockItem();
-      idx = self.store.data.length;
-      elt.index = idx ? idx + 1 : idx;
+      // idx = this.store.data.length;
+      // elt.index = idx ? idx + 1 : idx;
+      elt.index = this.store.data.length;
       this.store.post(elt);
     }
   };
@@ -120,7 +123,7 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
    * @description
    * Checks to see if the invoice has cached items to recover.
    */
-  StockForm.prototype.hasCacheAvailable =  function hasCacheAvailable() {
+  StockForm.prototype.hasCacheAvailable = function hasCacheAvailable() {
     return Object.keys(this.cache).length > 0;
   };
 
--- a/client/src/modules/stock/StockForm.service.js
+++ b/client/src/modules/stock/StockForm.service.js
@@ -52,14 +52,10 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
    */
   StockForm.prototype.addItems = function addItems(n) {
     var elt;
-    // var idx;
-    // var self = this;
 
     // will repeat will n > 0
     while (n--) {
       elt = new StockItem();
-      // idx = this.store.data.length;
-      // elt.index = idx ? idx + 1 : idx;
       elt.index = this.store.data.length;
       this.store.post(elt);
     }
--- a/client/src/modules/stock/StockForm.service.js
+++ b/client/src/modules/stock/StockForm.service.js
@@ -25,7 +25,7 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
 
     this.cache = AppCache(cacheKey);
     this.details = {};
-    this.store = new Store({ identifier: 'index', data: [] });
+    this.store = new Store({ data: [] });
   }
 
   /**
@@ -56,7 +56,7 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
     // will repeat will n > 0
     while (n--) {
       elt = new StockItem();
-      elt.index = this.store.data.length;
+      elt.id = this.store.data.length;
       this.store.post(elt);
     }
   };
--- a/client/src/modules/stock/StockItem.service.js
+++ b/client/src/modules/stock/StockItem.service.js
@@ -1,5 +1,5 @@
 angular.module('bhima.services')
-.service('StockItemService', StockItemService);
+  .service('StockItemService', StockItemService);
 
 StockItemService.$inject = [];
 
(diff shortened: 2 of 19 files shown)


client/src/modules/stock/StockItem.service.js
```diff
@@ -11,37 +11,21 @@ function StockItemService() {
...
+    this.quantity = 0;
+    this.unit_cost = 0;
+    this.expiration_date = new Date();
+    this.lots = [];
+    this.isValid = false;
+
+    this.validate();
   }
 
+  StockItem.prototype.validate = function validate() {
```
jniles, line 26: :+1:  Nice addition!

client/src/modules/stock/StockForm.service.js
```diff
@@ -36,10 +36,10 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
    * two lines are always present in the form.
    */
   StockForm.prototype.setup = function setup() {
-    var self = this;
-    self.details = {};
-    self.details.date = new Date();
-    self.details.user_id = Session.user.id;
+    this.details = {
```
jniles, line 39: :+1:

client/src/modules/stock/StockForm.service.js
```diff
@@ -52,12 +52,11 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
...
     var elt;
-    var idx;
-    var self = this;
+
+    // will repeat will n > 0
     while (n--) {
       elt = new StockItem();
-      idx = self.store.data.length;
-      elt.index = idx ? idx + 1 : idx;
+      elt.index = this.store.data.length;
```
jniles: I'm curious - what is the index used for?  Usually with a store, we want to index items by their IDs, not by their position in an array.

client/src/modules/stock/entry/entry.js
```diff
@@ -20,120 +20,137 @@ function StockEntryController(
...
-
-  var mapEntry = {
-    purchase    : { find : findPurchase, submit : submitPurchase },
-    donation    : { find : handleDonationSelection, submit : submitDonation },
-    integration : { find : handleIntegrationSelection, submit : submitIntegration },
-    transfer_reception : { find : findTransfer, submit : submitTransferReception },
-  };
-
-  vm.Stock = new StockForm('StockEntry');
+  vm.stockForm = new StockForm('StockEntry');
```
jniles, line 24: This is a good idea!  However, in the stock entry view, the form controller is called [StockForm](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/stock/entry/entry.html#L40).  That could get really confusing later.

I propose that you rename the "StockForm" in [entry.html](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/stock/entry/entry.html#L40) to be "StockEntryForm" to avoid this.

client/src/modules/stock/entry/entry.js
```diff
@@ -205,21 +226,38 @@ function StockEntryController(
...
-      item.inventory = inventoryStore.get(items[index].inventory_uuid);
-      item.unit_cost = items[index].unit_price || items[index].unit_cost;
-      item.quantity = items[index].balance || items[index].quantity;
+    vm.stockForm.store.data.forEach(function (item, index) {
+      var inventory = inventoryStore.get(items[index].inventory_uuid);
+
+      item.code = inventory.code;
+      item.inventory_uuid = inventory.uuid;
+      item.label = inventory.label;
+      item.unit_cost = items[index].unit_price || items[index].unit_cost; // transfer comes with unit_cost
```
jniles: Nice comment.

client/src/modules/stock/entry/entry.js
```diff
@@ -235,119 +273,116 @@ function StockEntryController(
...
+  /**
+   * @function processLotsFromStore
+   *
+   * @description
+   * This function loops through the store's contents mapping them into a flat array
+   * of lots.
+   *
+   * @returns {Array} - lots in an array.
+   */
+  function processLotsFromStore(data, uuid) {
```
jniles: I think this is used in a couple controllers now.  We should probably put it in a `StockService` eventually.

client/src/modules/stock/entry/modals/lots.modal.html
```diff
@@ -75,15 +75,22 @@
       </div>
     </div>
 
-    <p ng-if="$ctrl.maxLotReached" class="alert alert-warning" style="margin: 5px 0px;">
-      <i class="fa fa-warning"></i> 
-      <span translate>FORM.INFO.MAX_LOT_REACHED</span>
-    </p>
+    <div ng-if="!$ctrl.isAllValid">
```
jniles: Instead of having `!` in front of every `ngIf` here, it might be better to name compute these as values and name them appropriately.  For example, `!$ctrl.isAllValue` -> `$ctrl.hasInvalidEntries`, `!$ctrl.isLotExist` -> `$ctrl.hasNoLots`, `!$ctrl.isEveryLineCorrect` -> `$ctrl.hasIncorrectLines`, etc...

It makes it easier to read the HTML, since there isn't a `!` that's hiding in the expression.

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -9,196 +9,142 @@ StockDefineLotsModalController.$inject = [
...
-    var hasLotLabel = inventory.lot;
-    var hasExpiration = (new Date(inventory.expiration_date) >= new Date());
-    inventory.is_valid = (hasQuantity && hasLotLabel && hasExpiration);
-
-    vm.error = hasInvalidInventory(hasQuantity, hasExpiration);
-    vm.submitError = vm.error ? vm.error : false;
+  function checkAll() {
+    vm.isLotExist = vm.stockLine.lots.length > 0;
 
+    vm.isEveryLineCorrect = vm.stockLine.lots.every(function (lot) {
```
jniles: Instead of looking for each line being correct, you could also assert the opposite - that some line is not correct.  It might be quicker too, if `Array.prototype.some` breaks early (I'm not sure if it does).  Like this:
```js
vm.hasInvalidLines = vm.stockLine.lots.some(function (lot) {
  return lot.isValid === false;
});
```

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -9,196 +9,142 @@ StockDefineLotsModalController.$inject = [
...
+      return lot.isValid === true;
+    });
 
-    vm.remainingQuantity = (vm.inventory.quantity - sum >= 0) ? vm.inventory.quantity - sum : 0;
-    vm.sum = sum;
+    vm.sum = vm.stockLine.lots.reduce(function (x, y) {
```
jniles: Depending how often this is called (it looks like it might be called a lot), you could move this function out of the `checkAll()` function.  That way the V8 engine doesn't keep allocating a new function and discarding it when `checkAll()` completes.  Like this:
```js
function summer(x, y) {
  return x + y.quantity
}

function checkAll() {
  vm.sum = vm.stockLine.lots.reduce(summer, 0);

   // etc...
}

File: client/src/modules/stock/entry/modals/lots.modal.js [hunk 10]
```diff
@@ -9,196 +9,142 @@ StockDefineLotsModalController.$inject = [
...
-
-      if (!validLots()) { 
-        vm.submitError = vm.error ? vm.error : false;
+  function checkLine(line, date) {
+    if (date) { line.expiration_date = date; }
 
-        return;         
-      }
-    }
+    var isPosterior = new Date(line.expiration_date).getTime() >= new Date().getTime();
```
jniles: I think the `getTime()` call here is unnecessary.  It should work properly with just `new Date() > new Date()`.

jniles (commented):
Nice refactor.  I've made a few comments, but nothing really blocking.  Let me know what you think.

client/src/modules/stock/StockForm.service.js
DedrickEnc: You are right, this code was already there.
I think index is the name chosen to define a key of the object to store and the value does not matter because it is just an integer.

DedrickEnc: I decided to use ```id```

DedrickEnc:
@jniles 
Changes landed

client/src/modules/stock/entry/templates/code.tmpl.html
```diff
@@ -2,14 +2,14 @@
   <div ng-if="!row.entity._initialised">
     <input
       class="form-control"
-      ng-model="row.entity.inventory"
+      ng-model="row.entity.inventory_uuid"
```
jniles, line 5: :+1:

client/src/modules/stock/entry/modals/templates/lot.expiration.tmpl.html
```diff
@@ -3,7 +3,7 @@
   >
   <bh-date-picker
     date="row.entity.expiration_date"
-    on-change="grid.appScope.onDateChangeCallback(date, row.entity)"
+    on-change="grid.appScope.checkLine(row.entity, date)"
```
jniles, line 6: :dancer:

jniles (approved):
LGTM.

bors r+

PR #2220, opened 2017-10-19
fix(vouchers): undo reversals on delete

### LINKED ISSUE #2207: Undo reversals when vouchers are deleted
With https://github.com/IMA-WorldHealth/bhima-2.X/pull/2188, we gained the ability to delete records from the database tables.  However, when a user deletes a voucher that is a reversal, it does not toggle the "reversed" flag on the underlying table.  This means that cash payments and invoices are still marked as "reversed", even if the voucher has been deleted.

This is a corner case bug.  It should be fixed as soon as possible.

jniles:
This commit updates the voucher delete method to undo the "reversal"
flag on the reversed record if the voucher has been removed.  It is
currently a naive method - simply attempt to remove the value if it is
present.  This can be improved in the future by checking the transaction
types first.

Closes #2207.

### DIFF
--- a/server/controllers/finance/vouchers.js
+++ b/server/controllers/finance/vouchers.js
@@ -109,6 +109,14 @@ function lookupVoucher(vUuid) {
     });
 }
 
+// NOTE(@jniles) - this is used to find references for both vouchers and credit notes.
+const REFERENCE_SQL = `
+  v.uuid IN (
+    SELECT DISTINCT voucher.uuid FROM voucher JOIN voucher_item
+      ON voucher.uuid = voucher_item.voucher_uuid
+    WHERE voucher_item.document_uuid = ? OR voucher.reference_uuid = ?
+  )`;
+
 function find(options) {
   db.convert(options, ['uuid', 'reference_uuid', 'entity_uuid', 'cash_uuid', 'invoice_uuid']);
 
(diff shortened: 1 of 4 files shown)


jniles:
Hmm... it looks like the vouchers aren't correctly linked.  I'll follow up on this.

DedrickEnc (approved):
LGTM

jniles:
bors r-

This doesn't work yet due to https://github.com/IMA-WorldHealth/bhima-2.X/issues/2222. :(

jniles:
Okay, now this ready for a second look.

There have been two bugs fixed:
 1. I used the voucher after deleting it.  That was silly.
 2. The link between invoices and vouchers (and cash payments) was only searching on one of two possible fields (`document_uuid` and `reference_uuid`).  This is has been corrected in the voucher's `find()` method.

jniles:
@DedrickEnc, this can be looked at anytime.

server/controllers/finance/vouchers.js
```diff
@@ -266,15 +276,40 @@ function safelyDeleteVoucher(guid) {
     DELETE FROM document_map WHERE uuid = ?;
   `;
 
+  // NOTE(@jniles) - this is a naive way of undoing reversals.  If no value is
```
DedrickEnc, line 279: Next time it will be better to use appropriate comment instruction for multi-line. it is fancy.

PR #2226, opened 2017-10-20
Inline an account typeahead for Complex Vouchers

### LINKED ISSUE #1289: (proposal) use typeahead for defining row account(s) in complex vouchers
I propose that we re-factor the complex vouchers module to use typeaheads to select accounts. This would be similar in funcionality to the patient invoice module where inventory items can be searched for and then 'locked in' by selecting the item or pressing enter. 

Pros: 
* The current `ui-select` that is used is too tall for our grid rows. This is a bad user experience with some of the select missing from the bottom. A typeahead avoids this by being used on a simple native input.

* The DOM for the `ui-select` is much larger and more complex. This means that if there are a large number of rows in a voucher every row will have to render this complex element. It will also have to be rendered and removed when scrolling with `ui-grid`. A typeahead would simply be an input and it would be removed once the user has 'locked in' their choice. 
* A uniform `ui-grid` entity selection (accounts, inventory items). 
* `ui-select` overrides the tab key and forces you to tab through all of the options, in this case this is not helpful as it is just a small element on a complex page.
* If there are many accounts it can be much more convenient to just type in the number vs. given the display options in the first place. 

Cons: 
* `ui-select` is used for account select throughout the application, this would be a different experience however it would be inline with how selects work in grids. 
* If you have completely forgotten an account, both the name and the account number you will have to go and look it up in the chart of accounts.

jniles:
This commit adds the bhAccountTypeaheadInlineComponent for use in grids,
such as the Complex Voucher grid.  It is prototyped and demonstrated in
the Complex Vouchers module.

Closes #1289.

### DIFF
--- a/client/src/js/components/bhAccountTypeaheadInline.js
+++ b/client/src/js/components/bhAccountTypeaheadInline.js
@@ -1,0 +1,73 @@
+angular.module('bhima.components')
+  .component('bhAccountTypeaheadInline', {
+    templateUrl : 'modules/templates/bhAccountTypeaheadInline.html',
+    controller  : AccountTypeaheadInlineController,
+    bindings    : {
+      accountId        : '<',
+      onSelectCallback : '&',
+      name             : '@?',
+    },
+  });
+
+AccountTypeaheadInlineController.$inject = [
+  'AccountService', '$timeout', '$scope', 'Store',
+];
+
+/**
+ * Inline Account Typeahead
+ *
+ * This component is much more limited in options compared to the bhAccountSelect.
+ * It is intended to be used in ui-grids to facilitate entering accounts easily
+ * without having a heavy uiSelect component.
+ */
+function AccountTypeaheadInlineController(Accounts, $timeout, $scope, Store) {
+  var $ctrl = this;
+  var store = new Store();
+
+  // fired at the beginning of the account select
+  $ctrl.$onInit = function $onInit() {
+    // fired when an account has been selected
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
+
+    // default for form name
+    $ctrl.name = $ctrl.name || 'AccountForm';
+
+    // load accounts
+    loadAccounts();
+
+    // alias the name as AccountForm
+    $timeout(aliasComponentForm);
+  };
+
+  // this makes the HTML much more readable by reference AccountForm instead of the name
+  function aliasComponentForm() {
+    $scope.AccountForm = $scope[$ctrl.name];
+  }
+
+  // loads accounts from the server
+  function loadAccounts() {
+    Accounts.read()
+      .then(function (elements) {
+        // bind the accounts to the controller
+        var accounts = Accounts.order(elements);
+        $ctrl.accounts = Accounts.filterTitleAccounts(accounts);
+
+        store.setData($ctrl.accounts);
+      });
+  }
+
+  $ctrl.$onChanges = function $onChanges(changes) {
+    var accountId = changes.accountId && changes.accountId.currentValue;
+    if (accountId) {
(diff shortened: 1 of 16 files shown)


client/src/js/components/bhAccountTypeaheadInline.js
```diff
@@ -0,0 +1,73 @@
+angular.module('bhima.components')
+  .component('bhAccountTypeaheadInline', {
```
mbayopanda, line 2: 👍

client/src/modules/vouchers/complex-voucher.ctrl.js
```diff
@@ -174,9 +174,9 @@ function ComplexJournalVoucherController(
   }
 
   /** Get the selected currency symbol */
-  function currencySymbol(currency_id) {
-    if (!currency_id) { return; }
-    return Currencies.symbol(currency_id);
+  function currencySymbol(currencyId) {
```
mbayopanda, line 177: 👍

client/src/modules/vouchers/templates/account.grid.tmpl.html
```diff
@@ -1,20 +1,7 @@
...
-     class="ui-grid-ui-select"
-     ui-select-focus-patch
-     ui-disable-choice="account.type_id === grid.appScope.bhConstants.accounts.TITLE"
-     repeat="account.id as account in grid.appScope.Voucher.accounts | filter: { 'hrlabel' : $select.search } track by account.id">
-     <strong ng-bind-html="account.number | highlight:$select.search"></strong>
-     <span ng-bind-html="account.label | highlight:$select.search"></span>
-   </ui-select-choices>
- </ui-select>
+<div class="ui-grid-cell-contents">
+  <bh-account-typeahead-inline
```
mbayopanda, line 2: 👍

test/data.sql
```diff
@@ -14,7 +14,7 @@ INSERT INTO `project` VALUES (1,'Test Project A','TPA',1,1,0),(2,'Test Project B
 
 -- Accounts
 INSERT INTO `account` VALUES
-  (3626,6,1,1000,'Test Capital Account',0,0,NULL,NULL,'2015-11-04 13:25:12',1,NULL,NULL,NULL,1,1),
+  (3626,3,1,1000,'Test Capital Account',0,0,NULL,NULL,'2015-11-04 13:25:12',1,NULL,NULL,NULL,1,1),
```
mbayopanda, line 17: 💃

test/end-to-end/vouchers/complex.spec.js
```diff
@@ -6,30 +6,42 @@ const ComplexVoucherPage = require('./complex.page');
 const FU = require('../shared/FormUtils');
 const GU = require('../shared/GridUtils');
 
-describe('Complex Vouchers', function () {
+describe('Complex Vouchers', () => {
   before(() => helpers.navigate('vouchers/complex'));
 
-  it('creates a complex voucher', () => {
+  it('creates a complex voucher', function () {
```
mbayopanda, line 12: I'm just curious to know why you replaced the `() => ` by a `function ()`, is there any reason ?

mbayopanda (approved):
LGTM

test/end-to-end/vouchers/complex.spec.js
jniles: Yeah - arrow functions don't have a `this` object inside them, and on line 16 I use `this.timeout()`.

[According to stack overflow](https://stackoverflow.com/a/15982893), you are supposed to be able to use `it('something', () => { /* ... */ }).timout(45000);`.   That version didn't work for me though... I think our mocha version might not be up to date.

For more information on setting the timeout inside a test, [you can check out Mocha's documentation on it](http://mochajs.org/#test-level).

PR #2227, opened 2017-10-20
Add permission for stock entry/exit operations to depots

mbayopanda:
This PR add permissions to depots for stock entry/exit operations


mbayopanda:

client/src/i18n/en/stock.json
```diff
@@ -35,6 +35,7 @@
     "EXIT_PATIENT"    : "Exit to a patient",
     "EXIT_SERVICE"    : "Exit to a service",
     "EXIT_LOSS"       : "Loss",
+    "EXIT_INDIVIDIOUS" : "Exit to Individious",
```
jniles: We say "Individuals" in English.

client/src/js/components/bhStockEntryExitType.js
```diff
@@ -1,58 +1,64 @@
...
+  var $ctrl = this;
+
+  $ctrl.$onInit = function onInit() {
+    $ctrl.isEntry = $ctrl.isEntry === 'true';
+
+    reloadEntryExitTypes();
+  };
+
+  $ctrl.$onChanges = function onChanges() {
+    reloadEntryExitTypes();
```
jniles: It doesn't look like you need to reload the list every single time ... only when the depot changes.  A performance improvement would come from doing something like this:
```js
function onChanges(changes) }
  if (changes.depot) {
    reloadEntryExitTypes();
  }
}
```
That way, changes to the `name` and `reference` do not result in rebinding everything.

client/src/modules/depots/modals/depot.modal.js
```diff
@@ -19,6 +19,13 @@ function DepotModalController($state, Depots, ModalService, Notify) {
   function submit(depotForm) {
     var promise;
 
+    var stockEntryCache = new AppCache('StockEntry');
+    var stockExitCache = new AppCache('StockExit');
+
+    // remove depot from caches
+    delete stockEntryCache.depot;
```
jniles: It looks like this is a hack to make sure old cache values are not preserved.  The best way to do this is always cache only the id of the depot ... that way, the depot will be redownloaded each time and the new text will be presented.

client/src/modules/stock/StockEntryExitType.service.js
```diff
@@ -0,0 +1,100 @@
+angular.module('bhima.services')
+.service('StockEntryExitTypeService', StockEntryExitTypeService);
+
+StockEntryExitTypeService.$inject = [];
+
+function StockEntryExitTypeService() {
+  var $ctrl = this;
```
jniles: Since this is a service, could you call this `service` instead of `$ctrl`?  It will confuse future developers for sure.

client/src/modules/stock/StockEntryExitType.service.js
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * @function filterByEntry
+   *
+   * @description
+   * filter the Entry/Exit types list according the isEntry attribute
+   * is true or false
+   *
+   * @param {boolean} isEntry
+   */
+  function filterByEntry(isEntry) {
```
jniles, line 95: :+1:

client/src/modules/stock/StockEntryExitType.service.js
```diff
@@ -0,0 +1,100 @@
+angular.module('bhima.services')
+.service('StockEntryExitTypeService', StockEntryExitTypeService);
+
+StockEntryExitTypeService.$inject = [];
+
+function StockEntryExitTypeService() {
+  var $ctrl = this;
+
+  $ctrl.entryExitTypeList = [
```
jniles: Do you need to bind this to the service?  It actually looks like you could keep this list a private variable, and only expose it through the "getAllowedTypes()" api.

client/src/modules/stock/StockEntryExitType.service.js
```diff
@@ -0,0 +1,100 @@
...
+   * @description
+   * return allowed entry/exit feature for a given depot
+   *
+   * @param {object} depot
+   */
+  function getAllowedTypes(depot) {
+    if (!depot || !depot.uuid) { return []; }
+
+    return $ctrl.entryExitTypeList.map(function (item) {
+      item.allowed = depot[item.allowedKey];
```
jniles: Because this is a boolean, it would be nice to call it `isAllowed`.

client/src/modules/stock/StockEntryExitType.service.js
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * @function getAllowedTypes
+   *
+   * @description
+   * return allowed entry/exit feature for a given depot
+   *
+   * @param {object} depot
+   */
+  function getAllowedTypes(depot) {
+    if (!depot || !depot.uuid) { return []; }
```
jniles, line 78: Should this just throw an error instead?

test/end-to-end/account_statement/account_statement.spec.js
```diff
@@ -35,14 +35,15 @@ describe('Account Statement Core', () => {
     AccountStatement.expectRowCount(0);
   });
 
-  it.skip(`comment the rows for account ${sample.account} with ${sample.comment}`, () => {
-    // select the first row
-    AccountStatement.selectRow(0);
+  // skip throw error : Test `title` should be a "string" but "function" was given instead.
```
jniles, line 38: :dancer: Any particular reason why this had to be commented?

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -22,16 +22,81 @@
...
+        </div>
+        <div class="checkbox">
+          <label>
+            <input type="checkbox" name="allow_exit_service" ng-true-value="1" ng-false-value="0" ng-model="DepotModalCtrl.depot.allow_exit_service">
+            <span translate>STOCK.EXIT_SERVICE</span>
+          </label>
+        </div>
+        <div class="checkbox">
+          <label>
+            <input type="checkbox" name="allow_exit_transfer" ng-true-value="1" ng-false-value="0" ng-model="DepotModalCtrl.depot.allow_exit_transfer">
```
jniles, line 72: If the depot is not a warehouse, should this be shown?

jniles (commented):
Pretty good first implementation.  I've made a few clean up comments.  Let me know what you think.

client/src/modules/depots/modals/depot.modal.html
mbayopanda: I don't think so in the context of bhima (warehouse only can perform transfer to other depots) but the application must be flexible.

client/src/modules/stock/StockEntryExitType.service.js
mbayopanda: I don't think so, we are just handling the case where there is no depot it's doesn't mean that it is an error.

test/end-to-end/account_statement/account_statement.spec.js
mbayopanda: On Windows, the `skip` function in `it.skip(` break the e2e test down by throwing this error : **Test `title` should be a "string" but "function" was given instead**

jniles: Is your mocha version up to date?

mbayopanda:
@jniles could you review it and bors it ?

client/src/modules/depots/modals/depot.modal.html
jniles: Alright - I feel like there might be a hidden corner case here.  The code is fine - I'll merge this and we can discuss it.

client/src/modules/stock/StockEntryExitType.service.js
jniles: You never updated this!

jniles (commented):
@mbayopanda it doesn't look like you completed your changes.  Please updated `$ctrl` to be `service`.

jniles (approved):
Looks like all the changes have landed!  LGTM.

bors r+

PR #2230, opened 2017-10-22
Create filter handlebars helper

### LINKED ISSUE #1979: (feature request) a handlbars helper for rendering filters
It would save an awful lot of space of we made a [handlebars partial](http://handlebarsjs.com/#partials) for rendering filters in the handlebars PDF reports.  This way, we could do something like:
```hbs
{{!-- this will render the filters in a partial call "filter bar" --}}
{{> filterbar filters=filters }}
```

The code for the `filterbar` should look something like the ones found in the [patient registry handlebars file](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/medical/reports/registrations.handlebars#L10).
```hbs
  <p class="pills">
    {{#each filters}}
      <span class="label label-primary text-capitalize">
        <i class="fa fa-filter"></i> {{translate this.displayName}}

        {{#if this.comparitor}} {{this.comparitor}}
        {{else}}: {{/if}}

        {{#if this.isDate}} {{date this.value}}
        {{else}} {{translate this.value}} {{/if}}
      </span>
    {{/each}}
  </p>
```

### LINKED ISSUE #1980: chore(reports): ensure that all PDFs registries print filters at the top
We have filters clearly visible using the client-side `bhFilters` component linked to a filter service for each registry.  This make it clear to a user that the data in the view is filtered.

However, on the server side, after printing out a registry, we sometimes do not show the filters!  This is a big problem since the filters explain the data in the table.  It should be a requirement of every registry PDF report to show their filters above the table of data.

Both default filters and optional filters should be displayed - default filters will always be displayed.  See #1909 for more discussion of default filters.

The work will be made a lot easier if https://github.com/IMA-WorldHealth/bhima-2.X/issues/1979 is completed before this issue is tackled.

### LINKED ISSUE #2229: Filtering does not affect PDF report for employee registrations.
The PDF (and presumably CSV) reports for employees does not respond to filters.  Normally, when a registry is filtered, the downloads should reflect the view, as much as possible.  For the employee registry, this is not the case.

jniles:
This commit implements a filterbar partial to enable rendering filters on the server.  It implements a giant shared filter to render the correctly translated display value based on the filter passed in.  The basic usage pattern is like this:
```js
const shared = require('reports/shared');

function build(req, res, next) {
  const filters = shared.formatFilters(req.query);

  const data = { filters };

  // ... later ...
  report.render(data);
}
```
.. and in handlebars template ...
```hbs
{{> filterbar filters=filters }}
```

The following reports now use this partial for rendering filters:
 1. The Invoices Registry
invoiceregistryfilterbar

 2. The Cash Payments Registry
cashpaymentsregistryfilterbar

 3. The Patients Registry
patientregistryfilterbar

 4. The Voucher Registry
voucherregistryfilterbar

 5. The Employees Registry
employeemanagementfilterbar

Closes #2229.  Closes #1979.  Closes #1980.


client/src/modules/inventory/list/list.html
```diff
@@ -35,7 +35,7 @@
 
             <li role="separator" class="divider"></li>
             <li role="menuitem">
-              <a ng-href="/reports/inventory/items/?params={{ InventoryCtrl.parameters }}&lang={{ InventoryCtrl.lang }}"
+              <a ng-href="/reports/inventory/items/?params={{ InventoryCtrl.filters.formatHTTP(true) }}&lang={{ InventoryCtrl.lang }}"
```
DedrickEnc, line 38: Nice!

server/controllers/finance/reports/shared.js
```diff
@@ -0,0 +1,85 @@
...
+  { field : 'status', displayName: 'STOCK.STATUS.LABEL' },
+  { field : 'hospital_no', displayName : 'FORM.LABELS.HOSPITAL_NO' },
+  { field : 'patient_group_uuid', displayName : 'PATIENT_GROUP.PATIENT_GROUP' },
+
+  { field : 'entry_date_from', displayName: 'STOCK.ENTRY_DATE', comparitor: '>', isDate : true },
+  { field : 'entry_date_to', displayName: 'STOCK.ENTRY_DATE', comparitor: '<', isDate : true },
+  { field : 'expiration_date_from', displayName: 'STOCK.EXPIRATION_DATE', comparitor: '>', isDate : true },
+  { field : 'expiration_date_to', displayName: 'STOCK.EXPIRATION_DATE', comparitor: '<', isDate : true },
+  { field : 'is_exit', displayName : 'STOCK.OUTPUT' },
+  { field : 'label', displayName : 'STOCK.LOT' },
```
DedrickEnc, line 36: Now all filter field are in the same place, some of them lead to confusion, we should rename them appropriately even in an other pull request. For example ```label``` should be something like **lot** or **better**

server/controllers/finance/reports/shared.js
```diff
@@ -0,0 +1,85 @@
...
+  { field : 'display_name', displayName : 'FORM.LABELS.NAME' },
+  { field : 'sex', displayName : 'FORM.LABELS.GENDER' },
+  { field : 'code', displayName : 'FORM.LABELS.CODE' },
+  { field : 'dateBirthFrom', displayName : 'FORM.LABELS.DOB', comparitor : '>', isDate : true },
+  { field : 'dateBirthTo', displayName : 'FORM.LABELS.DOB', comparitor : '<', isDate : true },
+  { field : 'dateEmbaucheFrom', displayName : 'FORM.LABELS.DATE_EMBAUCHE', comparitor : '>', isDate : true },
+  { field : 'dateEmbaucheTo', displayName : 'FORM.LABELS.DATE_EMBAUCHE', comparitor : '<', isDate : true },
+  { field : 'grade_id', displayName : 'FORM.LABELS.GRADE' },
+  { field : 'fonction_id', displayName : 'FORM.LABELS.FUNCTION' },
+  { field : 'service_id', displayName : 'FORM.LABELS.SERVICE' },
```
DedrickEnc: Some filter field are repeated such as ```service_id``` you can just choose one.

server/controllers/finance/reports/shared.js
```diff
@@ -0,0 +1,85 @@
...
+
+  { field : 'reference', displayName : 'FORM.LABELS.REFERENCE' },
+  { field : 'reversed', displayName : 'CASH.REGISTRY.REVERSED_RECORDS' },
+  { field : 'limit', displayName : 'FORM.LABELS.LIMIT' },
+  { field : 'period', displayName : 'TABLE.COLUMNS.PERIOD', isPeriod : true },
+  { field : 'custom_period_start', displayName : 'PERIODS.START', isDate : true, comparitor : '>' },
+  { field : 'custom_period_end', displayName : 'PERIODS.END', isDate : true, comparitor : '<' },
+];
+
+_.sortBy(columns, c => c.field);
```
DedrickEnc: I can not see the benefit of sorting these items, you can remove it and avoid an unnecessary operation.

server/controllers/finance/reports/shared.js
```diff
@@ -0,0 +1,85 @@
...
+// translation key mappings for dynamic filters
+// Basically, to show a pretty filter bar, this will translate URL query string
+// into human-readable text to be placed in the report, showing the properties
+// filtered on.
+function formatFilters(qs) {
+  return columns.filter(column => {
+    const value = qs[column.field];
+
+    if (!_.isUndefined(value)) {
+      if (column.isPeriod) {
```
DedrickEnc: It will be better the handler service situation also. I am seeing something like ```Service : 2``` on the report filter. can you replace it with the service label?

server/controllers/finance/reports/shared.js
```diff
@@ -0,0 +1,85 @@
...
+
+// translation key mappings for dynamic filters
+// Basically, to show a pretty filter bar, this will translate URL query string
+// into human-readable text to be placed in the report, showing the properties
+// filtered on.
+function formatFilters(qs) {
+  return columns.filter(column => {
+    const value = qs[column.field];
+
+    if (!_.isUndefined(value)) {
```
DedrickEnc: Hmmm ... the flexibility of this logic is problematic. It seems like no error will be thrown if the client send an unknown filter key. silently ignoring exception is so ... unsafe don't you think?

server/controllers/inventory/reports/prices.js
```diff
@@ -23,7 +23,10 @@ const TEMPLATE = './server/controllers/inventory/reports/prices.handlebars';
 function prices(req, res, next) {
   const params = req.query.params ? JSON.parse(req.query.params) : {};
 
-  if (params && params.group_uuid) { params.group_uuid = db.bid(params.group_uuid); }
+  if (params && params.group_uuid) {
+    db.convert(params)
+    params.group_uuid = db.bid(params.group_uuid);
```
DedrickEnc: This operation is doing that : 

1. HUID --> BUID
2. BUID --> HUID

So why?

server/controllers/payroll/reports/registrations.js
```diff
@@ -107,7 +79,7 @@ function build(req, res, next) {
       // and skip rendering aggregates in the handlbars view
       if (!employees) { return false; }
 
-      // gather the uuids for the aggregate queries
+      // gather the ids for the aggregate queries
```
DedrickEnc, line 82: 👍

server/lib/template/partials/filterbar.handlebars
```diff
@@ -0,0 +1,15 @@
+{{#if filters}}
```
DedrickEnc, line 1: Very good idea to isolate this common logic 👍

DedrickEnc (commented):
@jniles 
I am pleased with this refactor, the server side filtering was implemented in many places, now it is more clear to read and understand. I have made some comments. 

I very good idea. 👍

server/controllers/finance/reports/shared.js
jniles: That is actually how filter parser works.  For example "renderer" is not a filter, but still exists on the query string.  So, will ignore the non-filter keys.  It's like setting `autoParseStatements` to false for the renderer.

server/controllers/inventory/reports/prices.js
jniles: Whoops, that is a typo.  I'll fix it.  Good catch :+1:

server/controllers/finance/reports/shared.js
jniles: :+1:

server/controllers/finance/reports/shared.js
jniles: I agree - most of the registry keys are unique, since many of them share the same filters.  However, the stock registries have many keys that overlap with other registries.  We could address this in another PR, since it will require changing both client and server.

server/controllers/finance/reports/shared.js
jniles: Nice catch!  It looks like that was the only one.

server/controllers/finance/reports/shared.js
jniles: We could do that if we had the correct translation for the service on the server.  The same goes for all the UUIDs that show up in these filters.  We do have the translated for periods because they are found in [lib/period.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/period.js#L15).

This PR is a first step to implementing what @lomamech just finished doing on the client - having human readable translation text for the filters.  It will require an extension of the `FilterService` on the client first, so that the `displayValue` is also back with every filter.  I'll propose a design in an issue.

client/src/modules/inventory/list/list.html
jniles: Unfortunately, this doesn't fix the inventory list :(.  I've made an issue to be tackled separately https://github.com/IMA-WorldHealth/bhima-2.X/issues/2232.

jniles:
@dedrickenc, thank you for the review.  I've addressed the points that you've made here.

I'll make an issue about sending `displayValue`s to the server with the client's FilterService and link it here when it is finished.

EDIT: See https://github.com/IMA-WorldHealth/bhima-2.X/issues/2233 .

DedrickEnc (approved):
LGTM

PR #2236, opened 2017-10-23
chore: use uniform interface for $state filters

### LINKED ISSUE #1935: proposal: passing filters to registries through $state.params
We have begun to experiment with filters that are applied to a registry through $state parameters.  For example, the Cash Registry implements them by:
 1. [Adding filters and display to the route](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/cash/cash.routes.js#L9).
 2. [Assigning the values to Cash.filters at startup](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/cash/payments/registry.js#L187).

This was a good first implementation, but has a number of drawbacks:
 1. The API of `filters` and `display` isn't obvious what goes where.  It turns out, `display` is never used.  I believe this was a hold-over from the old way of specifying filters.
 2. It isn't clear which filters take precedence - the cached ones or the $state ones?  Can the $state ones replace default filters?
 3. Only one filter can ever be used at a time from $state (in the current implementation).

**Design**

Building off the current implementation, I propose we adopt a new way of passing filters from `$state`.  Every registry that accepts external filters will provide a top-level `filters` array in the parameters, defined in this way:
```js
$stateProvider.state('someState', {
  url : '/xyz',
  controller : 'SomeCtrl',
  templateUrl :'some-template.html',
  params : { filters : [ ] }
});
```

In the startup method, we can simply check if the filters exist like this:
```js
```

If they do, we load them into the current view:
```js
if (hasStateFilters) {
  ControllerService.filters.replaceFiltersFromState($state.params.filters);
  ControllerService.cacheFilters();
}
```

The only difference between `replaceFilters` and `replaceFiltersFromState` is an extra `angular.copy()`.

A filter would be defined in the usual way:
```js
filters = [
  { key : 'account_id', value :1234 },
  { key : 'patient_uuid', value : 'some-long-uuid' }
];
```

**Other Thoughts**
 1. We could optionally specify if the filters should be merged, overwritten, or temporary (not cached).  I'm not sure what the best implementation should be.

jniles:
This commit implements the recommendations of #1935 on all registries. This allows all linked registries to send multiple filters through $state parameters as necessary.

Closes #1935.

### DIFF
--- a/client/src/modules/cash/payments/registry.js
+++ b/client/src/modules/cash/payments/registry.js
@@ -185,13 +185,12 @@ function CashPaymentRegistryController(
   }
 
   function startup() {
-    if ($state.params.filters) {
-      // fix me, generate change dynamically
-      var change = [{ key : $state.params.filters.key, value : $state.params.filters.value }];
-
-      Cash.filters.replaceFilters(change);
+    if ($state.params.filters.length) {
+      Cash.filters.replaceFiltersFromState($state.params.filters);
       Cash.cacheFilters();
     }
+
+    vm.latestViewFilters = Cash.filters.formatView();
 
     load(Cash.filters.formatHTTP(true));
     vm.latestViewFilters = Cash.filters.formatView();
--- a/client/src/modules/cash/cash.routes.js
+++ b/client/src/modules/cash/cash.routes.js
@@ -7,8 +7,7 @@ angular.module('bhima.routes')
         controller  : 'CashPaymentRegistryController as CPRCtrl',
         templateUrl : 'modules/cash/payments/registry.html',
         params      : {
-          filters : null,
-          display : null,
+          filters : [],
         },
       })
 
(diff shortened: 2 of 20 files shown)


client/src/modules/cash/payments/registry.js
```diff
@@ -185,14 +185,13 @@ function CashPaymentRegistryController(
   }
 
   function startup() {
-    if ($state.params.filters) {
-      // fix me, generate change dynamically
-      var change = [{ key : $state.params.filters.key, value : $state.params.filters.value }];
-
-      Cash.filters.replaceFilters(change);
+    if ($state.params.filters.length) {
+      Cash.filters.replaceFiltersFromState($state.params.filters);
```
mbayopanda, line 189: 👍 You have used `replaceFiltersFromState` in registries in a clearly way so the code looks uniforms.

mbayopanda (approved):
LGTM

Issue #2237, opened 2017-10-23
The depot selection modal can be dismissed before a depot is selected


jniles:
The Depot Selection Modal provides a way for a user to select a depot before they can use any of the stock/inventory modules.  However, it is possible to get around it and into an undefined state, simply by selecting a depot and clicking "cancel" instead of "Submit".  See an example below:

![depotselectioncancelable](https://user-images.githubusercontent.com/896472/31886888-1b563420-b7ee-11e7-9e7d-7a37bfaabf82.png)
_Fig 1: Cancel-able Depot Selection Modal_

**Steps to Reproduce**
 1. Refresh your cache so you do not have a depot selected.
 2. Navigate to a stock module (say, stock/adjustment).
 3. Observe that the modal comes up.  The cancel button is disabled.
 4. Click any depot to highlight it.
 5. Observe that the cancel button is now enabled.
 6. Click Cancel.

**Expected Result**
You shouldn't be able to enter a stock module without a depot.

**Actual Result**
You are now in a stock module, without a depot.

**Proposed Fix**
Do not allow the stock modal to be canceled without a depot id stored in the cache.  If there isn't a cached depot id, it should not be able to be canceled and the cancel button should be disabled.  If there is, cancelling should be an option.

PR #2238, opened 2017-10-23
chore: turn on unhandledRejectionErrors

### LINKED ISSUE #1686: refactor: handle all $q promise rejections
In AngularJS 1.6.x, the Angular team turned on console errors for unhandled promise rejections, similar to NodeJS 6.  In order to prevent console errors, [we turned them off](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/app.js#L301) using this little bit of code:
```js
function qConfig($qProvider) {
  $qProvider.errorOnUnhandledRejections(false);
}
```

This is pretty bad practice.  We may have unhandled errors that are being swallowed and we'll never know about it.  We should make a roadmap to enabling this class of errors to strengthen the entire application.

jniles:
This commit re-enables unhandled rejection errors for the $q promise library.  It looks like most modules are clean, without errors.  The only errors are corner cases caused by bugs (see #2237).  These can be
handled progressively as they are discovered.

Closes #1686.

### DIFF
--- a/client/src/js/app.js
+++ b/client/src/js/app.js
@@ -180,11 +180,6 @@ function uiSelectConfig(config) {
   config.theme = 'bootstrap';
 }
 
-// TODO - remove this
-function qConfig($qProvider) {
-  $qProvider.errorOnUnhandledRejections(false);
-}
-
 // configure services, providers, factories
 bhima.config(['$urlMatcherFactoryProvider', bhimaConfig]);
 bhima.config(['$translateProvider', translateConfig]);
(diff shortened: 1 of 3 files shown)


jniles:
Aha, this breaks our unit tests.  Interesting.

jniles:
@dedrickenc, this is ready for a review.

DedrickEnc:
Thanks @jniles, I will have a look after the break

DedrickEnc (approved):
LGTM

PR #2243, opened 2017-10-24
small fixes and refactors on stock exit module

DedrickEnc:
This pull request refactors the stock exit modules in terms of functionality : 
here are tasks performed : 

1. One of the changes made on the ```stockForm.service.js``` (replacing ```index``` by ```id``` in the store) leads to a regression in the stock exit module. It was not able to remove a line from the grid.

2. The ```bhGridLoadingIndicator.js``` is updated to accept a customized message

3. Now if the depot has no inventory to distribute, the grid will not be accessible and a message will be shown.

4. Some small refactors on comments and codes.

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -16,7 +16,7 @@
     "DETAILS"         : "Details",
     "DOCUMENT"        : "Document",
     "DONATION"        : "Donation",
-    "EMPTY"           : "No stock found",
+    "EMPTY"           : "You do not have any stock in your depot",
     "ENTRY"           : "Stock Entry",
     "ENTRY_DATE"      : "Entry Date",
     "ENTRY_DOCUMENT"  : "Entry Document",
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -16,6 +16,7 @@
     "DETAILS"         : "Details",
     "DOCUMENT"        : "Document",
     "DONATION"        : "Donation",
+    "EMPTY"           : "No stock found",
     "ENTRY"           : "Stock Entry",
     "ENTRY_DATE"      : "Entry Date",
     "ENTRY_DOCUMENT"  : "Entry Document",
--- a/client/src/js/components/bhGridLoadingIndicator.js
+++ b/client/src/js/components/bhGridLoadingIndicator.js
@@ -4,6 +4,7 @@ angular.module('bhima.components')
     loadingState : '<',
     emptyState   : '<',
     errorState   : '<',
+    messageState : '<?',
   },
   templateUrl : 'modules/templates/bhGridLoadingIndicator.tmpl.html',
 });
--- a/client/src/js/components/bhGridLoadingIndicator.js
+++ b/client/src/js/components/bhGridLoadingIndicator.js
@@ -1,10 +1,12 @@
 angular.module('bhima.components')
 .component('bhGridLoadingIndicator', {
   bindings : {
-    loadingState : '<',
-    emptyState   : '<',
-    errorState   : '<',
-    messageState : '<?',
+    loadingState         : '<',
+    loadingStateMessage  : '@?',
+    emptyState           : '<',
+    emptyStateMessage    : '@?',
+    errorState           : '<',
+    errorStateMessage    : '@?',
   },
   templateUrl : 'modules/templates/bhGridLoadingIndicator.tmpl.html',
 });
--- a/client/src/modules/stock/exit/exit.html
+++ b/client/src/modules/stock/exit/exit.html
@@ -118,6 +118,12 @@
         style="height: 300px; width: 100%;"
         ui-grid-auto-resize
         ui-grid-resize-columns>
+        <bh-grid-loading-indicator
+          loading-state="StockCtrl.loading"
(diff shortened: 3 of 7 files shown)


client/src/js/components/bhGridLoadingIndicator.js
```diff
@@ -4,6 +4,7 @@ angular.module('bhima.components')
     loadingState : '<',
     emptyState   : '<',
     errorState   : '<',
+    messageState : '<?',
```
jniles: It looks like this message only applies to the `emptyState`.  If so, a nicer binding would be `emptyStateMessage`.  That way, we could also have an `errorStateMessage` and `loadingStateMessage` if we need them.  The bindings would look something like this:
```html
<bh-grid-loading-indicator

  loading-state="$ctrl.loading"
  loading-state-message="I'm thinking about it ... "

  error-state="$ctrl.error"
  error-state-message="Something is terribly wrong."

  empty-state="$ctrl.rows.length === 0"
  empty-state-message="I have nothing to show you, my friends">
</bh-grid-loading-indicator>
```

jniles: You could also then make it a `@` binding and remove the requirement for a listener.

client/src/modules/stock/exit/exit.js
```diff
@@ -47,169 +38,173 @@ function StockExitController(
...
         item.lots = lots;
       })
       .catch(Notify.handleError);
   }
 
   function startup() {
+    // setting params for grid loading state
+    vm.loading = true;
+    vm.hasError = false;
+    vm.message = 'STOCK.EMPTY';
```
jniles: It doesn't look like this ever changes.  Could you define this in the HTML?  Like this:
```html
<bh-grid-loading-indicator
  ...
  message-state="'STOCK.EMPTY'">
</bh-grid-loading-indicator>
```

client/src/modules/stock/exit/exit.js
```diff
@@ -22,17 +22,8 @@ function StockExitController(
...
-    service : { description : 'STOCK.EXIT_SERVICE', find : findService, submit : submitService },
-    depot : { description : 'STOCK.EXIT_DEPOT', find : findDepot, submit : submitDepot },
-    loss : { description : 'STOCK.EXIT_LOSS', find : configureLoss, submit : submitLoss },
-  };
-
-  var cache = new AppCache('StockExit');
 
-  vm.util = util;
-  vm.Stock = new StockForm('StockExit');
+  vm.stockForm = new StockForm('StockExit');
```
jniles, line 26: It looks like the `StockForm` is [already defined in the scope](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/stock/exit/exit.html#L39).  This means that `$scope.StockForm` and `vm.stockForm` are both defined and different things.  To avoid confusion, could you rename this (or the view) as `StockExitForm`?

client/src/i18n/en/stock.json
```diff
@@ -16,6 +16,7 @@
     "DETAILS"         : "Details",
     "DOCUMENT"        : "Document",
     "DONATION"        : "Donation",
+    "EMPTY"           : "No stock found",
```
jniles: Should this say something more specific?  Like "You do not have any stock in your depot" or something?

jniles (commented):
A few suggestions here.  Let me know what you think.

DedrickEnc:
@jniles
I have pushed changes

client/src/js/components/bhGridLoadingIndicator.js
```diff
@@ -1,9 +1,12 @@
 angular.module('bhima.components')
 .component('bhGridLoadingIndicator', {
   bindings : {
-    loadingState : '<',
-    emptyState   : '<',
-    errorState   : '<',
+    loadingState         : '<',
+    loadingStateMessage  : '@?',
+    emptyState           : '<',
+    emptyStateMessage    : '@?',
```
jniles, line 7: :+1:

client/src/modules/stock/exit/exit.html
```diff
@@ -36,7 +36,7 @@
 <div class="flex-content">
   <div class="container-fluid">
 
-    <form name="StockForm" bh-submit="StockCtrl.submit(StockForm)" novalidate>
+    <form name="StockExitForm" bh-submit="StockCtrl.submit(StockExitForm)" novalidate>
```
jniles, line 39: Better! 👍

client/src/modules/templates/bhGridLoadingIndicator.tmpl.html
```diff
@@ -17,7 +17,7 @@
   <!-- indicate that some error occurred -->
   <div class="msg" ng-show="$ctrl.errorState">
     <span class="text-danger">
-      <span class="fa fa-warning"></span> <span translate>ERRORS.UNKNOWN</span>
+      <span class="fa fa-warning"></span> <span translate>{{ $ctrl.errorStateMessage || 'ERRORS.UNKNOWN' }}</span>
```
jniles, line 20: In the future, this operation would be best done in your `$onInit` method.  It would allow you to easily use one-time binding (`::`), as well as clarify that the message will never change after the first initialization.

jniles (approved):
LGTM.

bors r+

PR #2244, opened 2017-10-24
feat(Dropdown behavior) turns up dropdown if the page can fall off

### LINKED ISSUE #1124: (bug) Users page dropdown falls off page
The `ui-grid`'s dropdown in the `users` configuration page pushes the entire grid up as you get to the bottom of the page.  It also does not appear below the link in the page.  I have a feeling that it is not appended to the body, but someone should investigate.

weirddropdownbehavior
_Fig 1: Users Dropdown Behavior_

jeremielodi:
This PR helps to turn up the dropdown  if it is at the page bottom, it solve is problem described at #1124 
closes #1124


client/src/js/services/dropdownBehavior.js
```diff
@@ -0,0 +1,29 @@
...
+  //this function check if html element's position is on the bottom
+  // if true it changes the element's position in order to display it conviniently
+
+  // it function is usefull for carets for exemple
+  // the Users page dropdown won't falls off page thanks to this function
+  // it takes the dropdown id as parameter
+
+  service.setPosition = function(id_dropdown){
+    
+    var menu =  $('#'+ id_dropdown);
```
jniles: In general, you should try not use JQuery when you use AngularJS.  They have two different ways of tackling problems.

Instead of JQuery, AngularJS provides [jqLite (also known as angular.element())](https://docs.angularjs.org/api/ng/function/angular.element) to manipulate the DOM.  In general, AngularJS also asks that you do DOM manipulations in a directive.

I propose that you move this into a directive that extends Bootstrap's dropdown menu.  It would listen on the [isOpen](https://github.com/angular-ui/bootstrap/blob/master/src/dropdown/dropdown.js#L239) variable, and then position the dropdown menu appropriately.  You can still use a service to do the calculation, but all DOM manipulation should happen in a directive.

It would something like this:
```html
<ul uib-dropdown-menu  bh-dropdown-menu-auto-dropup>
</ul>
```

client/src/js/services/dropdownBehavior.js
```diff
@@ -0,0 +1,29 @@
+angular.module('bhima.services')
+.service('DropdownBehaviorService', DropdownBehaviorService);
+
+function DropdownBehaviorService() {
+
+  var service = this;
+
+  //this function check if html element's position is on the bottom
```
jniles: Nice comments!

jniles (changes requested):
@jeremielodi, I honestly didn't think this would get fixed anytime soon.  I'm glad you chose to tackle it!

The behavior looks pretty good.  It would be best to move this functionality into a directive, as proposed below.  That would give you three things:
 1. You would get the jQuery element for free (it would be injected as `$element` into your controller). No need for excess `id=""` in the DOM.
 2. You could listen in on the dropdown menu's scope for changes.
 3. No need to involve the parent controller - everything would be self-contained.

Looks like an excellent start!  :+1:

client/src/modules/users/users.js
```diff
@@ -7,7 +7,7 @@ UsersController.$inject = ['$state', 'UserService', 'NotifyService', 'ModalServi
  * Users Controller
  * This module is responsible for handling the CRUD operation on the user
  */
-function UsersController($state, Users, Notify, Modal, uiGridConstants) {
+function UsersController($state, Users, Notify, Modal, DropdownBehavior, uiGridConstants) {
```
jniles: It doesn't look like you use `DropdownBehavior` anymore.  Go ahead and remove it.

client/src/js/directives/bhDropdownMenuAutoDropup.js
```diff
@@ -0,0 +1,45 @@
...
+angular.module('bhima.directives')
+.directive('bhDropdownMenuAutoDropup', ['$document',
+  function($document) {
+    return {
+      restrict: 'AC',
+      link: function(scope, iElement, iAttrs) {
+
+        var iElementWrapper = iElement.parent();
+
+        iElementWrapper.on('click', function() {
```
jniles, line 25: Nice!

jniles (commented):
This looks excellent.  There is one unused import which breaks the user controller.

jniles (approved):
Works pretty well for me.  Great work!

bors r+

PR #2245, opened 2017-10-25
deps: replace node-uuid with uuid

### LINKED ISSUE #1395: (deps): Replace node-uuid with uuid
The `node-uuid` package [has been deprecated by its maintainers](https://www.npmjs.com/package/node-uuid) in favor of the `uuid` package.  Both come from the same source code (`uuid` was a fork of `node-uuid`), so they have migrated to the `uuid` package to not duplicate efforts.

We should upgrade whenever convenient.  The `node-uuid` package will likely stop receiving updates in the future.

jniles:
This commit replaces the deprecated node-uuid with the uuid npm module. The uuid library is supported through the community, and has much smaller implementation.  It removes support for `parse()` and
`unparse()` methods, which are now implemented in the `uuid-parse` library.  Since we use these methods upon occasion ( e.g. `db.bid()`), I have added that as a dependency as well.

Closes #1395.

### DIFF
--- a/server/controllers/admin/grades.js
+++ b/server/controllers/admin/grades.js
@@ -5,7 +5,7 @@
  * This controller exposes an API to the client for reading and writing Grade
  **/
 const db = require('../../lib/db');
-const uuid = require('node-uuid');
+const uuid = require('uuid/v4');
 const NotFound = require('../../lib/errors/NotFound');
 
 // GET /Grade
@@ -56,16 +56,17 @@ function detail(req, res, next) {
 // POST /grade
 function create(req, res, next) {
   const data = req.body;
+  const recordUuid = data.uuid || uuid();
 
   // Provide UUID if the client has not specified
-  data.uuid = db.bid(data.uuid || uuid.v4());
+  data.uuid = db.bid(recordUuid);
 
   const sql =
     'INSERT INTO grade SET ? ';
 
   db.exec(sql, [data])
     .then(() => {
-      res.status(201).json({ uuid : uuid.unparse(data.uuid) });
+      res.status(201).json({ uuid : recordUuid });
     })
     .catch(next)
     .done();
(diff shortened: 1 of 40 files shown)


server/controllers/admin/grades.js
```diff
@@ -56,16 +56,17 @@ function detail(req, res, next) {
 // POST /grade
 function create(req, res, next) {
   const data = req.body;
+  const recordUuid = data.uuid || uuid();
 
   // Provide UUID if the client has not specified
```
mbayopanda, line 61: The comment is now about convert into binary, for example : Convert record string uuid into a binary uuid

server/controllers/admin/grades.js
```diff
@@ -56,16 +56,17 @@ function detail(req, res, next) {
...
-  data.uuid = db.bid(data.uuid || uuid.v4());
+  data.uuid = db.bid(recordUuid);
 
   const sql =
     'INSERT INTO grade SET ? ';
 
   db.exec(sql, [data])
     .then(() => {
-      res.status(201).json({ uuid : uuid.unparse(data.uuid) });
+      res.status(201).json({ uuid : recordUuid });
```
mbayopanda, line 69: 👍

server/controllers/medical/patients/documents.js
```diff
@@ -68,21 +68,21 @@ function create(req, res, next) {
...
+      // publish a patient update event
+      Topic.publish(Topic.channels.MEDICAL, {
+        event : Topic.events.UPDATE,
+        entity : Topic.entities.PATIENT,
+        user_id : req.session.user.id,
+        id : req.params.uuid,
+      });
+
+      res.status(201).json({
+        uuids : req.files.map(file => file.filename),
```
mbayopanda, line 81: it is not obvious that `filename` is a uuid, a comment on that will be not bad.

Ex. 
```js
// document file name is in the string uuid format
// return a set of these files names as an array of uuid
```

server/controllers/medical/patients/index.js
```diff
@@ -90,8 +90,7 @@ exports.latestInvoice = latestInvoice;
 function create(req, res, next) {
   const createRequestData = req.body;
 
-  let medical = createRequestData.medical;
-  let finance = createRequestData.finance;
+  let { medical, finance } = createRequestData;
```
mbayopanda, line 93: Maybe `const` can be used instead of `let`.

but your code is very nice 👍

mbayopanda (approved):
LGTM

jniles:
@mbayopanda, thanks for the review.  I'll make the changes today.

PR #2247, opened 2017-10-25
feat(depot select modal) avoid canceling is the depot is required

### LINKED ISSUE #2237: The depot selection modal can be dismissed before a depot is selected

jeremielodi:
This PR helps to fixe this issue #2237
closes #2237

### DIFF
--- a/client/src/modules/depots/depots.service.js
+++ b/client/src/modules/depots/depots.service.js
@@ -20,13 +20,18 @@ function DepotService(Api, Modal) {
    * Opens the selection modal to allow a user to select a depot.
    *
    * @returns Promise - a promise containing the selected depot.
+   * 
+   * @param isDepotRequired helps to keep the modal displayed as long as no depot was submitted
    */
-  service.openSelectionModal = function openSelectionModal(depot) {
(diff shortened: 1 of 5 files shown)


jniles (approved):
Looks good to me!  Great fix.

bors r+

PR #2248, opened 2017-10-26
feat(employee registry) display stored date interval

### LINKED ISSUE #2228: Employee Registry Search form does not preserve hiring date
The Employee Registry's search form does not remember the "hiring dates" when the user closes and opens the search form.  Normally, any filter applied should be automatically loaded into the search modal form when the modal is loaded.

**Steps to Reproduce**
 1. Navigate to the Employee Registry.
 2. Click the "Search" button
 3. Set a hiring date range.  Also set a name.
 4. Submit.
 5. Click the "search" button again.

**Expected Result**
Both the name field and the hiring date should be populated with the previous values.

**Actual Result**
Only the name field is restored.

jeremielodi:
This PR solves this problem #2228
closes #2228

### DIFF
--- a/client/src/js/components/bhDateInterval.js
+++ b/client/src/js/components/bhDateInterval.js
@@ -7,7 +7,7 @@
  *
  * @example
  * ```html
- * <bh-date-interval date-from="$MyCtrl.dateFrom" date-to="$MyCtrl.dateTo" >
+ * <bh-date-interval date-from="$MyCtrl.dateFrom" date-to="$MyCtrl.dateTo">
  * </bh-date-interval>
  * ```
  */
@@ -15,15 +15,16 @@ angular.module('bhima.components')
   .component('bhDateInterval', {
     templateUrl: '/modules/templates/bhDateInterval.tmpl.html',
     controller: bhDateInterval,
-    bindings: {
-      validationTrigger: '<', // validation trigger action
-      dateFrom: '=',          // date from
-      dateTo: '=',            // date to
-      dateId: '@',            // date identifier
-      required: '<',          // true or false
-      onChange: '<',          // on change action
-      mode: '@'               // the date mode (day|month|year)
-    }
+    bindings : {
+      validationTrigger : '<', // validation trigger action
+      dateFrom : '=', // date from
+      dateTo : '=', // date to
+      dateId : '@?', // date identifier
+      required : '<?', // true or false
+      onChange : '<?', // on change action
+      label : '@?',
+      mode : '@?', // the date mode (day|month|year)
+    },
   });
 
 // dependencies injection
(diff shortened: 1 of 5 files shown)


jniles:
Did you test this?  It doesn't look like it works for me.  It just defaults to the current date.

jeremielodi:
Oh! I didn't check very well, for me the dates are displayed but are default ones. it  doesn't get the strored dates yet

jeremielodi:
The component it self modifies the stored value by the default one. This should be fixed

client/src/js/components/bhDateInterval.js
```diff
@@ -48,11 +49,13 @@ function bhDateInterval(moment, bhConstants) {
     vm.dateFormat = bhConstants.dayOptions.format;
 
     vm.pickerOptions = { showWeeks: false };
-
-    // start up the modal
     startup();
   };
 
+  vm.onChange = function(){
```
jniles: This should be removed before merged into master.

client/src/modules/employees/registry/registry.js
```diff
@@ -182,9 +182,11 @@ function EmployeeRegistryController(
 
   function search() {
     var filtersSnapshot = Employees.filters.formatHTTP();
-
     Employees.openSearchModal(filtersSnapshot)
       .then(function (changes) {
+        //This is very important if changes is undefined, a cache problem occurs
```
jniles, line 187: Nice comment. :+1:

client/src/modules/employees/registry/search.modal.js
```diff
@@ -90,7 +91,7 @@ function EmployeeRegistryModalController(ModalInstance, bhConstants, moment, Ser
       if (angular.isDefined(value)) {
         // default to the original value if no display value is defined
         var displayValue = displayValues[key] || lastDisplayValues[key] || value;
-        changes.post({ key: key, value: value, displayValue: displayValue });
+        changes.post({ key: key, value: value, displayValue: value });
```
jniles: Is this change really necessary?  It looks like it undoes the work by @lomamech on displaying human-readable values instead of IDs.

client/src/modules/templates/bhDateInterval.tmpl.html
```diff
@@ -1,4 +1,4 @@
-<ng-form name="DateIntervalForm" novalidate>
+<ng-form name="DateIntervalForm{{ $ctrl.dobDate }}" novalidate>
```
jniles: This looks like a typo.  Can you remove it?

client/src/js/components/bhDateInterval.js
```diff
@@ -79,27 +82,42 @@ function bhDateInterval(moment, bhConstants) {
...
+
   function clear() {
     delete vm.dateFrom;
     delete vm.dateTo;
   }
 
   function startup() {
     var option;
-
+    /*
```
jniles: If these are no longer needed, just delete them.

client/src/js/components/bhDateInterval.js
```diff
@@ -79,27 +82,42 @@ function bhDateInterval(moment, bhConstants) {
     vm.dateTo = moment().endOf('year').toDate();
   }
 
+ 
+  function ajustDateType(myvar){
```
jniles: Are you sure this function is necessary?  It looks like it will just do `new Date()` if `myvar` is defined.  what happens if you don't use it?

jniles (dismissed):
There's a quite a few clean up things to take care of before this gets into `master`.  Let me know what you think.

client/src/js/components/bhDateInterval.js
jeremielodi: dates are not displayed misfortunatelly

client/src/modules/employees/registry/search.modal.js
jeremielodi: I don't understand what is happening if I use  @lomamech  method here,  if I let  things like this
changes.post({ key: key, value: value, displayValue: displayValue });
the first store works but no the second.
When I try to change  doing things like this 
changes.post({ key: key, value: value, displayValue: value });
every thing works flawlessly

jniles: Alright, I'll fork this PR and see if I can get it to work.  I don't want to destroy @lomamech's work.

jeremielodi: I have already used @lomamech's  storing style in another issue, it worked without any problem. But I don't know why it poses problem here

jniles:
@jeremielodi, I forked this PR and made some changes.  Let me know what you think of them.

PR #2249, opened 2017-10-26
 feature(Cashbox for User)

### LINKED ISSUE #2242: Add Cashbox Permissions By User
Like #2171 
This issues proposal to  limit access cashboxes By User. The user must be able to perform Cashbox Management operations.

lomamech:
- Implement component bh-multiple-cashbox-select
- Update User registry add in template Action Cell Management Cashbox
- Implement modal to Set Management Cashboxes to User
- Implement API for Set cashbox to User
- Implement Integration test for Cashbox permissions
- Update and fix integration test for Depot Permission
- Implement test E2E and Integration

closes #2242

### DIFF
--- a/client/src/i18n/en/cash.json
+++ b/client/src/i18n/en/cash.json
@@ -22,7 +22,9 @@
       "CURRENT_PROJECT_CASHBOXES" : "Current Project Cashboxes",
       "OTHER_PROJECT_CASHBOXES" : "Other Project Cashboxes",
       "NO_OTHER_PROJECT_CASHBOXES" : "No cashboxes found registered to other projects.",
-      "NO_CURRENT_PROJECT_CASHBOXES" : "No cashboxes found registered to the current project."
+	  "NO_OTHER_PROJECT_CASHBOXES_ACCESS" : "No cashboxes found registered to other projects that you have access right to.",
+      "NO_CURRENT_PROJECT_CASHBOXES" : "No cashboxes found registered to the current project.",
+	  "NO_CURRENT_PROJECT_CASHBOXES_ACCESS" : "There are no cashbox in your project that you have access right to"
     },
     "TITLE":"Cash Window",
     "RECEIPT":{
--- a/client/src/js/components/bhCashboxSelect.js
+++ b/client/src/js/components/bhCashboxSelect.js
@@ -53,4 +53,4 @@ function CashboxSelectController(Cashbox, Notify) {
   $ctrl.onSelect = function ($item) {
     $ctrl.onSelectCallback({ cashbox : $item });
   };
-}    
+}
--- a/client/src/js/components/bhCashboxSelect.js
+++ b/client/src/js/components/bhCashboxSelect.js
@@ -36,7 +36,7 @@ function CashboxSelectController(Cashbox, Notify) {
     $ctrl.label = $ctrl.label || 'FORM.SELECT.CASHBOX';
 
     // load all Cashbox
-    Cashbox.read(null, { detailed: 1, user_id : true })
+    Cashbox.read(null, { detailed: 1, only_user : true })
       .then(function (cashboxes) {
         cashboxes.forEach(function (cashbox) {
           cashbox.hrlabel = cashbox.label + ' ' + cashbox.symbol;
--- a/client/src/js/components/bhCashboxSelect.js
+++ b/client/src/js/components/bhCashboxSelect.js
@@ -15,14 +15,14 @@ angular.module('bhima.components')
   });
 
 CashboxSelectController.$inject = [
-  'CashboxService', 'NotifyService'
+  'CashboxService', 'NotifyService', 'SessionService'
 ];
 
 /**
  * Cashbox selection component
  *
  */
-function CashboxSelectController(Cashbox, Notify) {
+function CashboxSelectController(Cashbox, Notify, Session) {
   var $ctrl = this;
 
   $ctrl.$onInit = function onInit() {
@@ -36,7 +36,7 @@ function CashboxSelectController(Cashbox, Notify) {
     $ctrl.label = $ctrl.label || 'FORM.SELECT.CASHBOX';
 
     // load all Cashbox
-    Cashbox.read(null, { detailed: 1 })
+    Cashbox.read(null, { detailed: 1, user_id : Session.user.id })
       .then(function (cashboxes) {
         cashboxes.forEach(function (cashbox) {
           cashbox.hrlabel = cashbox.label + ' ' + cashbox.symbol;
--- a/client/src/js/components/bhCashboxSelect.js
+++ b/client/src/js/components/bhCashboxSelect.js
@@ -15,19 +15,19 @@ angular.module('bhima.components')
   });
 
 CashboxSelectController.$inject = [
-  'CashboxService', 'NotifyService', 'SessionService'
+  'CashboxService', 'NotifyService',
 ];
 
 /**
  * Cashbox selection component
  *
  */
-function CashboxSelectController(Cashbox, Notify, Session) {
+function CashboxSelectController(Cashbox, Notify) {
   var $ctrl = this;
 
   $ctrl.$onInit = function onInit() {
     // fired when a Cashbox has been selected
-    $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback;
 
     // default for form name
     $ctrl.name = $ctrl.name || 'CashboxForm';
@@ -36,7 +36,7 @@ function CashboxSelectController(Cashbox, Notify, Session) {
     $ctrl.label = $ctrl.label || 'FORM.SELECT.CASHBOX';
 
     // load all Cashbox
-    Cashbox.read(null, { detailed: 1, user_id : Session.user.id })
+    Cashbox.read(null, { detailed: 1, user_id : true })
       .then(function (cashboxes) {
         cashboxes.forEach(function (cashbox) {
           cashbox.hrlabel = cashbox.label + ' ' + cashbox.symbol;
--- a/client/src/js/components/bhMultipleCashboxSelect.js
+++ b/client/src/js/components/bhMultipleCashboxSelect.js
@@ -33,12 +33,8 @@ function MultipleCashboxSelectController(Cashbox, Notify) {
     $ctrl.cashboxesIds = $ctrl.cashboxesIds || [];
 
     // load all Cashbox
-    Cashbox.read(null, { detailed: 1 })
+    Cashbox.read()
       .then(function (cashboxes) {
-        cashboxes.forEach(function (cashbox) {
-          cashbox.hrlabel = cashbox.label + ' ' + cashbox.symbol;
-        });
-
         $ctrl.cashboxes = cashboxes;
       })
       .catch(Notify.handleError);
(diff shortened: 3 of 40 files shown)


client/src/i18n/en/cash.json
```diff
@@ -22,7 +22,9 @@
       "CURRENT_PROJECT_CASHBOXES" : "Current Project Cashboxes",
       "OTHER_PROJECT_CASHBOXES" : "Other Project Cashboxes",
       "NO_OTHER_PROJECT_CASHBOXES" : "No cashboxes found registered to other projects.",
-      "NO_CURRENT_PROJECT_CASHBOXES" : "No cashboxes found registered to the current project."
+	  "NO_OTHER_PROJECT_CASHBOXES_ACCESS" : "No cashboxes found registered to other projects that you have access right to.",
+      "NO_CURRENT_PROJECT_CASHBOXES" : "No cashboxes found registered to the current project.",
+	  "NO_CURRENT_PROJECT_CASHBOXES_ACCESS" : "There are no cashbox in your project that you have access right to"
```
jniles, line 27: Nice! :+1:

client/src/js/components/bhCashboxSelect.js
```diff
@@ -36,7 +36,7 @@ function CashboxSelectController(Cashbox, Notify) {
     $ctrl.label = $ctrl.label || 'FORM.SELECT.CASHBOX';
 
     // load all Cashbox
-    Cashbox.read(null, { detailed: 1 })
+    Cashbox.read(null, { detailed: 1, user_id : Session.user.id })
```
jniles: Hmm... the user could in theory alter this, since it's on the client.  I wonder what we should do about that.  We could use `req.session.user.id` on the server, but that would limit ability to download all cashboxes.

We need a better security model in general.  I do not have specific proposals for this route, so this is a non-blocking comment.

client/src/js/components/bhMultipleCashboxSelect.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+function MultipleCashboxSelectController(Cashbox, Notify) {
+  var $ctrl = this;
+
+  $ctrl.$onInit = function onInit() {
+    // label to display
+    $ctrl.label = $ctrl.label || 'FORM.LABELS.CASHBOX';
+
+    // fired when a Cashbox has been selected or removed from the list
+    $ctrl.onChange = $ctrl.onChange || angular.noop;
```
jniles: Since this function is required in your bindings (`&`), there isn't any need to alias it with a `noop` function.

client/src/modules/cash/modals/select-cashbox-modal.ctrl.js
```diff
@@ -43,6 +43,27 @@ function SelectCashboxModalController(Session, Instance, Cashboxes, $stateParams
       })
       .catch(Notify.handleError)
       .finally(toggleLoadingIndicator);
+
+    /**
+    * This section makes it possible to check if the user does not have permissions to a cash register or that it does not exist
+    */ 
+    Cashboxes.read(undefined, { is_auxiliary : 1 })
```
jniles: It would be best not to download the cashboxes twice.  Is there any way to send something like an HTTP parameter `includeUserPrivileges=1`  that would do a `cash_box LEFT JOIN cashbox_permission`?  That way you could loop through the list twice and group by project and user?

client/src/modules/reports/generate/cash_report/cash_report.html
```diff
@@ -65,24 +65,11 @@ <h3 class="text-capitalize" translate>REPORT.CASH_REPORT.TITLE</h3>
...
-                    ng-options="cash.account_id as cash.hrlabel for cash in ReportConfigCtrl.cashboxes track by cash.id"
-                    required>
-                    <option value="" disabled>{{ "FORM.SELECT.CASHBOX" | translate }}</option>
-                </select>
-                <div class="help-block" ng-messages="ConfigForm.cashbox.$error" ng-show="ConfigForm.$submitted">
-                <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
-                </div>
-            </div>
+            <bh-cashbox-select
+              cashboxId = "ReportConfigCtrl.reportDetails.account_id"
```
jniles, line 69: I think this should be `cashbox-id`.  Also, why is this account id?

client/src/modules/users/UserCashBoxManagementModal.html
```diff
@@ -0,0 +1,39 @@
...
+        <label class="badge badge-warning" translate>USERS.SET_CASHBOX</label>
+      </li>
+    </ol>
+  </div>
+
+  <div class="modal-body" style="overflow-y: scroll; max-height:600px; ">
+
+
+    <div class="form-group" ng-class="{ 'has-error' : UserForm.$submitted && UserForm.display_name.$invalid }">
+      <label class="control-label col-md-12" translate>FORM.LABELS.USERNAME</label>
```
jniles: Why do you need the `col-md-12` on these lines?

client/src/modules/users/UserCashBoxManagementModal.html
```diff
@@ -0,0 +1,39 @@
...
+      <div class="col-md-12">
+        <h4><i> {{ UsersCashBoxModalCtrl.user.display_name }} </i></h4>
+      </div>
+    </div>
+
+    <span class="help-block" translate> FORM.INFO.USER_CASHBOX </span>
```
jniles: I think `help-block`s often go within the `form-group`.  Could you move this up into the form group?

client/src/js/components/bhMultipleCashboxSelect.js
```diff
@@ -0,0 +1,47 @@
+angular.module('bhima.components')
+  .component('bhMultipleCashboxSelect', {
+    templateUrl : 'modules/templates/bhMultipleCashBoxSelect.tmpl.html',
+    controller  : MultipleCashboxSelectController,
+    bindings    : {
+      onChange : '&',
+      cashboxesIds : '<?',
```
jniles: This should probably be `cashbox-ids` to make it plural.

client/src/modules/users/users.routes.js
```diff
@@ -47,7 +47,15 @@ angular.module('bhima.routes')
         },
         onEnter :['$uibModal', depotManagementModal],
         onExit : ['$uibModalStack', closeModal]
-      })      
+      })
+      .state('users.cashBoxManagement', {
+        url : '/:id/cashBoxManagement',
```
jniles: Since your server URL is ` $http.get('/users/' + id + '/cashboxes')`, it would be nice to make this `/:id/cashboxes`.  That way, it is easier for developers to find it.

server/controllers/admin/users/cashboxes.js
```diff
@@ -0,0 +1,67 @@
...
+  })
+  .catch(next)
+  .done();
+}
+
+
+/**
+ * POST /users/:id/cashboxes
+ *
+ * Creates and updates a user's cashboxes.  This works by completely deleting
```
jniles, line 42: 👍

server/controllers/admin/users/depots.js
```diff
@@ -49,7 +49,7 @@ function create(req, res, next) {
     .addQuery('DELETE FROM depot_permission WHERE user_id = ?;', [req.params.id]);
 
   // if an array of permission has been sent, add them to an INSERT query
-  if (req.body.depots) {
+  if (req.body.depots.length) {
```
jniles, line 52: 💃

server/controllers/finance/cashboxes/index.js
```diff
@@ -54,6 +54,18 @@ function list(req, res, next) {
   filters.equals('project_id');
   filters.equals('is_auxiliary');
 
+  if (req.query.detailed === '1') {
```
jniles, line 62: Does just `filters.equals('user_id')` not work for both cases?

test/data.sql
```diff
@@ -242,6 +242,13 @@ INSERT INTO `cash_box_account_currency` VALUES
   (3,1,2,3626,3627),
   (4,2,2,3626,3627);
 
+-- Set Cashbox Management By User
+INSERT INTO cashbox_permission (id, user_id, cashbox_id) VALUES
```
jniles: Instead of inserting NULLs, you can just do this:
```sql
INSERT INTO cashbox_permission (user_id, cashbox_id) VALUES (1,1), (1,2);
```

test/end-to-end/cash/cash.spec.js
```diff
@@ -31,6 +35,18 @@ describe('Cash Payments', () => {
   }
 
   describe('Cashbox Select Interface', () => {
+    it(`Set Cashbox ${cashboxC.text} Manage Right to RegularUser `, () => {
```
jniles: This belongs in the `users` tests.  It isn't a test tied to cash payments.

test/end-to-end/depots/depots.spec.js
```diff
@@ -16,12 +16,11 @@ describe('Depots Management', () => {
 
   const Page = new DepotPage();
   /**
-  /* The implementation of the E2E test of the assignment of a repository
-  /* to a user is added in the E2E test of the repository and not in the 
-  /* user test because the E2E test on the users runs after the repository test. 
+  /* The implementation of the E2E test of the assignment of a Depot
+  /* to a user is added in the E2E test of the Depot and not in the 
+  /* user test because the E2E test on the users runs after the Depot test.
```
jniles, line 21: 💃

test/integration/users.js
```diff
@@ -243,4 +244,37 @@ describe('(/users) Users and Permissions', function () {
...
+      .then(function (res) {
+        helpers.api.listed(res, 3);
+        expect(res).to.have.status(200);
+        expect(res.body).to.not.be.empty;
+      })
+      .catch(helpers.handler);
+  });
+
+  // Reset cashbox permissions for user
+  it('POST /users/:id/cashboxes will reset user cashboxes', function () {
```
jniles, line 265: 👍

test/integration/users.js
```diff
@@ -243,4 +244,37 @@ describe('(/users) Users and Permissions', function () {
       .catch(helpers.handler);
   });
 
+
+  // Add cashbox permissions for user
+  it('POST /users/:id/cashboxes will create user cashboxes', function () {
```
jniles, line 249: These are really nice tests!

jniles (commented):
@lomamech, this represents a lot of work.  I'm a bit concerned about the security model employed here ... if a user changes their user_id locally, it will affect the cashboxes they have access to.   I don't have a better proposal right now, but we'll need to come up with one long-term.

You tests are excellent!  They should all be `users` since they are tied to the `users` module.

I've made some code clean up comments.  Let me know what you think.

client/src/modules/reports/generate/cash_report/cash_report.html
lomamech: @jniles, the value to be sent is indeed the account Id, but after analysis this report does not work and there is no End to End test to verify the proper functioning of the report / reports / cash_report

server/controllers/finance/cashboxes/index.js
lomamech: I used an alternative structure just because in the first case we use the table cash_box_account_currency and in the second it's cash_box

server/controllers/finance/cashboxes/index.js
```diff
@@ -36,14 +37,18 @@ exports.currencies = currencies;
  * optional parameters:
  */
 function list(req, res, next) {
+  if (req.query.user_id) {
```
jniles: This is pretty confusing.  Why don't you use a variable that isn't a filter property -  Like `showOnlyUserCashboxes`?  That way, developers don't get confused thinking that `user_id = true` is a real user ID somewhere.

jniles (commented):
@lomamech, one final piece to update, and then this can be merged.

jniles (approved):
LGTM!

bors r+

PR #2250, opened 2017-10-26
feat(enterprise) edit location

jeremielodi:
This PR enable editing the enterprise location, there was a missing attribute in the bh-location-select, the name.


jniles:
Good catch!

jniles (approved):
bors r+

PR #2252, opened 2017-10-26
fix(vouchers): sort receipt items

### LINKED ISSUE #2221: Sort the Voucher Receipt's lines by Account Number
The receipt produced by any voucher has the tendency to be quite complex.  It is hard to read as well, because the lines tend to be printed in a nonsensical order.  See below for an example:

voucherreceipt
_Fig 1: Voucher Receipt with Multiple Lines_

This can easily be solved by sorting by account.  This will (at minimum) group the debtor lines by arranging them with the 41 and 42s clumped together.

Note that you should sort by account number, and not account_id.  The account_id is not guaranteed to be in order.

jniles:
This commit fixes the voucher receipt by sorting the lines first by account, then by amounts, and finally by the entity affected.  This should ensure a fairly stable appearance to all receipts, no matter how the accountant ordered the transaction initially or in what order the database decided to enter them.

Closes #2221.

**Unordered Voucher Items (prior to this change)**
unorderedvouchers

**Ordered Voucher Items (after this change)**
orderedvouchers

### DIFF
--- a/server/controllers/finance/vouchers.js
+++ b/server/controllers/finance/vouchers.js
@@ -95,7 +95,8 @@ function lookupVoucher(vUuid) {
     JOIN account a ON a.id = vi.account_id
     LEFT JOIN entity_map ON entity_map.uuid = vi.entity_uuid
     LEFT JOIN document_map ON document_map.uuid = vi.document_uuid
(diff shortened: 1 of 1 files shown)


DedrickEnc (approved):
LGTM

PR #2256, opened 2017-10-26
Feature:  bhReceiptComponent for displaying receipts

jniles:
This PR implements a receipt-rendering component that opens the receipt modal when a link is clicked.  It supports three values:
 1. value - the ID or UUID for the receipt
 2. displayValue - whatever you would like to show.
 3. type - a string denoting what kind of receipt it is.  This will call the corresponding method in the ReceiptModal service.

With the bhReceipt component on the financial registries, there is no need to import the ReceiptModal into every single controller.  I've updated the vouchers, purchases, invoices, and cash payments accordingly.

### DIFF
--- a/client/src/js/components/bhReceipt.js
+++ b/client/src/js/components/bhReceipt.js
@@ -21,7 +21,7 @@ function bhReceiptController(ReceiptModal, $log) {
     // make sure the receipt type exists before it is clicked
     var hasCallbackFn = ReceiptModal[$ctrl.type];
     if (!hasCallbackFn) {
-      $log.warn('Warning: Cannot find '
+      $log.error('Warning: Cannot find '
         .concat($ctrl.type)
         .concat(' in ReceiptModalService.'));
     }
--- a/client/src/js/components/bhReceipt.js
+++ b/client/src/js/components/bhReceipt.js
@@ -1,0 +1,34 @@
+var bhReceiptTemplate =
+  '<a ng-click="$ctrl.open()" href="">{{ $ctrl.displayValue }}</a>';
+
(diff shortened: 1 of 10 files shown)


jniles:
@DedrickEnc fixed the bug with opening receipts on invoices. This can be reviewed anytime.

client/src/js/components/bhReceipt.js
```diff
@@ -0,0 +1,34 @@
...
+bhReceiptController.$inject = ['ReceiptModal', '$log'];
+
+function bhReceiptController(ReceiptModal, $log) {
+  var $ctrl = this;
+
+  $ctrl.$onInit = function $onInit() {
+    // make sure the receipt type exists before it is clicked
+    var hasCallbackFn = ReceiptModal[$ctrl.type];
+    if (!hasCallbackFn) {
+      $log.warn('Warning: Cannot find '
```
DedrickEnc: Ha ha this is actually an error. This is a no blocking comment.

DedrickEnc (approved):
LGTM

client/src/js/components/bhReceipt.js
jniles: Yeah, you are right.  I'll make it an error.

jniles:
@DedrickEnc changes landed.  I also updated the commit message to reflect that the ReceiptModal service is still required in InvoiceRegistry.

DedrickEnc:
Thanks @jniles 
As soon as the test pass, it will be merged.

jniles:
Looks like the test failures were a fluke.

bors r+

PR #2258, opened 2017-10-27
refactor(Inventory Purchasable)

### LINKED ISSUE #2251: (refactor) stock Select Inventory Purchasable
Add a filter to select products that have reached a purchase order point,

lomamech:
- Implement the way to filter inventory who requires a Purchase Order
- Update API for stock core filter the inventorys who Quantity is sup to
  zero when require_po is true
- Update E2E test for inventory stock add test for the Inventories who
  require purchase order and use property resetFilters for improve some
  test

closes #2251

### DIFF
--- a/client/src/modules/stock/inventories/modals/search.modal.html
+++ b/client/src/modules/stock/inventories/modals/search.modal.html
@@ -31,7 +31,7 @@
             <bh-clear on-clear="$ctrl.clear('inventory_uuid')"></bh-clear>
           </bh-inventory-select>
 
-          <div class="form-group checkbox">
+          <div class="checkbox">
             <label>
               <input type="checkbox"
                 ng-model="$ctrl.searchQueries.require_po"
--- a/client/src/modules/stock/inventories/modals/search.modal.html
+++ b/client/src/modules/stock/inventories/modals/search.modal.html
@@ -31,6 +31,15 @@
             <bh-clear on-clear="$ctrl.clear('inventory_uuid')"></bh-clear>
           </bh-inventory-select>
 
+          <div class="form-group checkbox">
+            <label>
+              <input type="checkbox"
+                ng-model="$ctrl.searchQueries.require_po"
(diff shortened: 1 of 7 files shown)


server/controllers/stock/core.js
```diff
@@ -450,6 +456,11 @@ function getInventoryQuantityAndConsumption(params) {
       if (status) {
         return rows.filter(row => row.status === status);
       }
+
+      if(require_po) {
```
jniles: It looks like `status and `require_po` are exclusive.  Can both options be set?  If so, these shouldn't `return rows.filter()`, they should do something like:
```js
var filteredRows = rows;

if (status) {
  filteredRows = filteredRows.filter(row => row.status === status);
}

if (require_po) {
  filteredRows = filteredRows.filter(row => row.S_Q > 0);
}

return filteredRows;
```

client/src/modules/stock/inventories/modals/search.modal.html
```diff
@@ -31,6 +31,15 @@
             <bh-clear on-clear="$ctrl.clear('inventory_uuid')"></bh-clear>
           </bh-inventory-select>
 
+          <div class="form-group checkbox">
```
jniles: There's no need for the `form-group` class here.

jniles (commented):
@lomamech, this code is pretty good!

I'm curious - shouldn't almost all the inventory be needing purchase as soon as one stock item is consumed?  It looks like this basically filters on `quantity != S_MAX`.

If so, is it really that useful to have a clientside filter?  The server-side filter is useful because we can then populate a PO for a client, but I'm not so sure the checkbox is useful.  What do you think?

lomamech:
The client-side filter was implemented just to allow inventory managers to evaluate the list of products requiring a purchase order, I will create another issue this issue will select through the Purchase Order module the inventories that requires a purchase order (Automatic Purchase Order)

jniles (approved):
LGTM!

bors r+

PR #2259, opened 2017-10-30
refractor(registries) use bh-dropdown-menu-auto-dropup in all registries

### LINKED ISSUE #2246: Implement bh-dropdown-menu-auto-dropup on all registries
Once #2244 is merged, it will greatly ease usability of the registries.  For registries that contain lots of information, the default behavior of the dropdown is to drop off the bottom of the page.  It's quite difficult to use, even for experienced computer users.

We should go through all the registries and add the `bh-dropdown-menu-auto-dropup` directive to our dropdown menus.  Since it is only a directive, all we have to do is touch the actions.html templates that contain the dropdown menus.

jeremielodi:
This PR helps to use the  bh-dropdown-menu-auto-dropup directive in all registries, the directive created since #2244 landed
closes #2246

### DIFF
--- a/client/src/modules/billing-services/templates/action.cell.html
+++ b/client/src/modules/billing-services/templates/action.cell.html
@@ -4,7 +4,7 @@
     <span class="caret"></span>
   </a>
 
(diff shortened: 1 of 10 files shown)


jniles (approved):
LGTM.

bors r+

PR #2260, opened 2017-10-30
feat(inventory) edit item

### LINKED ISSUE #1822: (bug) Editing inventory items
This bug was submitted to the BHIMA `support@bhi.ma` thread. 

There has been a reported issue with editing inventory items in BHIMA. After a successful edit, editing a second inventory item will not correctly populate the `/update/` modal.

**Problem Description**
> Problems editing of data on List view of inventory
Got to  Inventory/List view in BHIMA
Click on item.  Select edit.  Make changes.  Save edits.
When you click on the second item the Inventory / Edit Inventory pop
up window has ***no information***.  Each of the field simply says
"Enter code" and is empty.  This behavior is exactly the same as the
Add Inventory Item pop up window functions.
The only way to access to edit another item in this view is to refresh
the page and then the second time is available.

> priority: high (not a crisis but the level below)  The system works but not as expected resulting in extra work.

I have not been able to reproduce this error. I think this is a combination of things resulting in overloading the low end server hardware so that it cannot server the `/update` modal in a timely fashion. 

Proposed methods for resolving issues: 
* Continue to try to reproduce - using both the installation build version and data set 
* Remove the need to re-download every single inventory item on every edit 
* Reduce the load of this module for low end hardware to try to reduce overloading the server causing the client to not receive values

This issue should be mostly addressed by the inventory registry work in the pending pull request found https://github.com/IMA-WorldHealth/bhima-2.X/pull/1789

jeremielodi:
Some requirements of this issue : #1822 was already solved, as I noticed, But the only problem That I have remarked is when the submit the form without doing any change, a sql syntaxe occures. So this PR solves it

closes #1822

### DIFF
--- a/client/src/modules/inventory/list/modals/actions.modal.js
+++ b/client/src/modules/inventory/list/modals/actions.modal.js
@@ -44,9 +44,9 @@ function InventoryListActionsModalController(Account, Inventory, Notify, Instanc
 
     // if no changes were made, simply dismiss the modal
     if (util.isEmptyObject(record)) {
-      cancel();
+      return cancel();
     }
-
+ 
     var promise = vm.isCreateState ?
       Inventory.create(record) :
       Inventory.update(vm.identifier, record);
--- a/server/controllers/inventory/inventory/core.js
+++ b/server/controllers/inventory/inventory/core.js
@@ -8,7 +8,7 @@
 const uuid = require('node-uuid');
 const db = require('../../../lib/db');
 const FilterParser = require('../../../lib/filter');
-
+const _ = require('lodash');
 // this should be a const in future ES versions
 const errors = {
   MISSING_PARAMETERS : {
@@ -78,14 +78,22 @@ function updateItemsMetadata(record, identifier) {
   if (record.group_uuid) {
     record.group_uuid = db.bid(record.group_uuid);
   }
+ 
 
   const sql = 'UPDATE inventory SET ? WHERE uuid = ?;';
+  // if there is no property to update this query won't work
+
+  if (_.isEmpty(record)) { //there is no change, but user has submitted
+    return getItemsMetadataById(identifier);
+  }
+
   /*
    * return a promise which can contains result or error which is caught
    * in the main controller (inventory.js)
    */
   return db.exec(sql, [record, db.bid(identifier)])
-    .then(() => getItemsMetadataById(identifier));
+  .then(() => getItemsMetadataById(identifier));
+
 }
 
 /**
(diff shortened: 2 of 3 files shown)


server/lib/util.js
```diff
@@ -118,3 +119,15 @@ function dateFormatter(rows, dateFormat) {
   return rows;
 }
 
+//check if an object has no key
+function isEmpty(map) {
```
jniles: There's a lodash function for this [_.empty()](https://lodash.com/docs/4.17.4#isEmpty).  No need to re-implement it.

server/controllers/inventory/inventory/core.js
```diff
@@ -78,14 +78,28 @@ function updateItemsMetadata(record, identifier) {
   if (record.group_uuid) {
     record.group_uuid = db.bid(record.group_uuid);
   }
+ 
+
+  let sql = 'UPDATE inventory SET ? WHERE uuid = ?;';
+  // if there is no property to update this query won't work
+
+  if(util.isEmpty(record)) { //there is no change, but user has submitted
+    sql = 'UPDATE inventory SET uuid = ? WHERE uuid = ?;';
```
jniles: Instead of executing this query, why not just return `getItemsMetadataById()`?

Also, we should probably block a user from sending an empty request on the client.

client/src/modules/inventory/list/modals/actions.modal.js
```diff
@@ -44,9 +44,9 @@ function InventoryListActionsModalController(Account, Inventory, Notify, Instanc
 
     // if no changes were made, simply dismiss the modal
     if (util.isEmptyObject(record)) {
-      cancel();
+      return cancel();
```
jniles, line 47: Nice! :+1:

server/controllers/inventory/inventory/core.js
```diff
@@ -78,14 +78,24 @@ function updateItemsMetadata(record, identifier) {
...
     record.group_uuid = db.bid(record.group_uuid);
   }
+ 
 
   const sql = 'UPDATE inventory SET ? WHERE uuid = ?;';
+  // if there is no property to update this query won't work
+
+  if(_.isEmpty(record)) { //there is no change, but user has submitted
+    return getItemsMetadataById(identifier);
+  }else{
```
jniles: This is much better.    ESLint should tell you that you do not need to `else` statement.  Please make sure you have ESLint installed in your editor.

```js
if (_.isEmpty(record)) {
  return getItemsMetadataById(identifier);
}

return db.exec(sql, [record, db.bid(identifier)])
  .then(() => getItemsMetadataById(identifier));
```

PR #2264, opened 2017-10-30
tests: remove multiple element warnings

### LINKED ISSUE #2193: Remove "W/element - more than one element found for locator" warnings from end to end tests
Our end to end tests are riddled with errors that look like the following:
```sh
[15:06:55] W/element - more than one element found for locator by.model("EmployeeCtrl.employee.sex") - the first result will be used
```
These warnings mean that multiple elements have matched a CSS selector - in this case, a selector like:
```js
element(by.model('EmployeeCtrl.employee.sex'))
```
was used and multiple values were found for it.  This makes sense - [the `sex` input is a radio button](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/employees/registration/employees.html#L78) and so there are multiple `ng-model`s defined.  However, the test informs us that it will only use the first result, giving us little control over what actually is happening.  This defeats the purpose of a test.

We should go through all these and clean them up.  We have already seen the [painful results](https://github.com/IMA-WorldHealth/bhima-2.X/issues/2094) of [inaccurate tests](https://github.com/IMA-WorldHealth/bhima-2.X/issues/2110).

jniles:
This commit addresses as many duplicate element warnings as possible. There remains warnings on the open debtors report, since the uiSelect uses an icon to differentiate the sort order, which cannot be selected upon easily.

Closes #2193.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,10 +15,10 @@
     "src"
   ],
   "dependencies": {
-    "angular": "1.6.5",
-    "angular-animate": "1.6.5",
-    "angular-sanitize": "1.6.5",
-    "angular-messages": "1.6.5",
+    "angular": "1.6.6",
+    "angular-animate": "1.6.6",
+    "angular-sanitize": "1.6.6",
+    "angular-messages": "1.6.6",
     "angular-bootstrap": "2.5.0",
     "angular-ui-router": "1.0.0-rc.1",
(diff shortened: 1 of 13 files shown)


jniles:
@DedrickEnc can I get a review?

DedrickEnc (approved):
LGTM

PR #2271, opened 2017-11-01
Pruning i18n files

### LINKED ISSUE #2187: Prune i18n translation keys
We have many translation keys that may or may not be out of date.  It would be helpful to write a tool to find out which keys are not used in the application, or if there are any keys that are missing, rather than simply relying on a programmer doing the right thing.

If you want to work on it, the basic outline would be this:
 1. Import the `en.json` file.  This will have you a list of keys that should be used in the application.
 2. Use a HTML parser - probably something like [cheerio](https://www.npmjs.com/package/cheerio) to load all the HTML files in `client/src/modules/`.  You may need some tools that will load files based on [globs](https://www.npmjs.com/package/glob).
 3. Get all the `translate` directives and their contents.  In pseudocode, you will want to do something like this:
```js
const keys = $.querySelectorAll('[translate]').map(element => element.getInnerText());
```
 4. Either write your own [or use an existing tool](https://www.npmjs.com/package/dot-wild) to check if each key is in the object of keys.
 5. For each translation found in the application, mark the corresponding key as "found".
 6. Anything left at the end of the loop should be removed.

DedrickEnc:
This pull request is about writting a script to look for keys which are not used in the template and removing them.
Element with the attribute ```translate``` are extracted form template thanks to ```cheerio``` and they are removed form the ```en.json``` by scanning each i18n file.

this script is ignoring keys with angular format, so it can be addressed in a separated pull request, it is also running for English translation only but it can be easily extend to french.

You need the npm module ```cheerio``` to run the script and run ```npm run clean-i18n``` to run the script

closes #2187


utilities/translation/clean.js
```diff
@@ -0,0 +1,185 @@
...
+    let keys;
+    keys = keyString.split('.');
+
+    return keys.reduce((current, next) => {
+        if (baseObejct[next]) {
+            return current[next] = baseObejct[next];
+        } else {
+            if (typeof current[next] === 'string') {
+                delete current[next];
+                return current;
```
jniles, line 162: You can move this `return current` out of the `if` for simplicity.  It doesn't depend on the `if` logic.
```js
if (typeof current[next] === 'string') {
  delete current[next];
}
return current;
```

utilities/translation/clean.js
```diff
@@ -0,0 +1,185 @@
...
+    }
+});
+
+function getAllKeyString(obj, path, res) {
+    res = res || [];
+
+    keys = Object.keys(obj).sort();
+
+    keys.forEach((key) => {
+        if (typeof obj[key] === 'object') {
```
jniles, line 123: There are a couple places where you use `typeof`.  The code would be easier to read if you used `lodash`'s wrappers:
```js
if (_.isString(object[key])) {
  // do something
}

if (_.isObject(object[key])) {
  /// do something
}
```

utilities/translation/clean.js
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+});
+
+function getAllKeyString(obj, path, res) {
+    res = res || [];
+
+    keys = Object.keys(obj).sort();
+
+    keys.forEach((key) => {
```
jniles, line 122: It looks like you might be able to do:
```js
return keys.map(key => {
});
```
Then instead of `return res.concat(x)` it would just be `return x`.

utilities/translation/clean.js
```diff
@@ -0,0 +1,185 @@
...
+
+// Now the en object will contain only used key, the next step is trying to remove them safely and through this script
+let remainKeys = Object.keys(en).sort();
+
+remainKeys.forEach((key) => {
+    // transforming to lower case every key and removing unuseful spaces
+    key = key.toLowerCase().trim();
+
+    // Evoiding dealing with value parsed in angular format and other
+    if (key.indexOf('{') === -1 &&
```
jniles, line 80: Nice!

utilities/translation/clean.js
```diff
@@ -0,0 +1,185 @@
...
+            //loading the small json file used for developpement
+            let obj = require(jsonRoute);
+
+            // for every i18n json file, this function will return an array of keyString with the format : KEY1.KEY2.KEY3
+            var list = getAllKeyString(obj, '', []);
+
+            //processing every single key string
+            list.forEach((item) => {
+                // return true if the key string is present in the object containing data to remove (en.json)
+                if (isPresent(item)) {
```
jniles, line 97: Nice function name 👍

utilities/translation/clean.js
```diff
@@ -0,0 +1,185 @@
...
+
+                    //if true, the last key is removed from the small json file obj
+                    obj[item.split('.')[0]] = removeLastKey(item, obj);
+                }
+            });
+
+            // stringify the object in order to write with utf8 encoding
+            const objAsString = JSON.stringify(obj);
```
jniles, line 105: Most of the JSON files written out this way end up really difficult to modify - they are all compressed on one line.

I propose you use something like [json-prettify](https://www.npmjs.com/package/json-prettify) to write out more easily manipulated JSON files.

utilities/translation/clean.js
```diff
@@ -0,0 +1,185 @@
...
+            const f = fs.openSync(jsonRoute, 'w');
+
+            // Writting  a new content into the small json file i18n
+            fs.writeFileSync(f, objAsString, 'utf8');
+        }
+        catch (e) {
+            //FIX ME : in case require can not find the file, how can I handle it properly?
```
jniles, line 112: How often does this occur?

utilities/translation/clean.js
```diff
@@ -0,0 +1,185 @@
...
+            }
+        }
+
+    }, {});
+}
+
+function isPresent(keyString) {
+    let found = true;
+    let keys = keyString.split('.');
+    let currentObj = en;
```
jniles, line 174: I propose you make this function take in the object as well as the key string.  That way we can more easily extend this code to French/other languages.

Like:
```js
function isPresent(object, keyString) {
  // gimme a boolean
}
```

jniles (commented):
@DedrickEnc this is a really good start.  The code is really clean.

A couple of corner cases we might consider:
 1. What about the values defined in JS?  like `$ctrl.key = 'x.y.z'`?
 2. What about the handlebars files?

I think we could pretty easily extend the parser you wrote to do the handlebars files. 

I propose that you add a "white list" of keys that we can use for values defined in JS.  The white list ensure that they are not removed from the JSON files.  For example, you could add:
```js
// These values are used in the application, but are not found in the HTML. 
// For example, the server may send error codes for the Trial Balance, 
// NotifyService often has custom error translation keys, or labels are passed
// into components.  If an item is added to this list, it will not be removed 
// from the JSON files.
const keyWhiteList = [
  'SOME.COOL.KEY'.=,
];
```

What do you think?

Edit: It is also worth noting that a _lot_ of our [translation keys](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/receipts/modal/receiptModal.tmpl.html#L6) are still using the $translate filter syntax.   PR https://github.com/IMA-WorldHealth/bhima-2.X/pull/2155 attempted to fix this, but was closed an never reopened.  Before we run this script on the i18n JSONs, we should update all our translation keys.

utilities/translation/clean.js
DedrickEnc: One time with ```color.json``` because the key is colors.

DedrickEnc:
Thank you very much @jniles for the review. I will see as soon as possible

PR #2273, opened 2017-11-03
feat(filters): implement getDisplayValueMap()

### LINKED ISSUE #2209: Implement FilterService.getDisplayValueMap()
We're recently undertaken a large refactor of our filters to ensure that human readable ids are presented to the user when they filter the registries instead of large uuids or irrelevant numbers.  One of the last PRs to land will be https://github.com/IMA-WorldHealth/bhima-2.X/pull/2206.

Each search modal controller now contains the following code to create a map of `id:displayValue`:
```js
var lastViewFilters = SOME_SERVICE.filters.formatView().customFilters;
var lastDisplayValues = lastViewFilters.reduce(function (object, filter) {
    object[filter._key] = filter.displayValue;
    return object;
}, {});
```

This code could easily be included in the `FilterService` on the client.  Instead, of the above 5+ lines of code, a search modal should just do this:
```js
var lastDisplayValues = SOME_SERVICE.filters.getDisplayValueMap();
```
to the same effect.  It would reduce the complexity in our search modals and the total lines of code overall.

jniles:
This commit implements the getDisplayValueMap() method on the
FilterService as a wrapper for code that cached display values for
lookup later.  This code was used in every single search modal file.  It
has now been replace with a method on the FilterService itself.

Closes #2209.

### DIFF
--- a/client/src/js/services/FilterService.js
+++ b/client/src/js/services/FilterService.js
@@ -163,6 +163,24 @@ function FilterService() {
     return filtered;
   };
 
+  /**
+   * @method getDisplayValueMap
+   *
+   * @description
+   * Turns stored filters into a map of filter key -> display value.  This
+   * allows the search modals to get the display values in case any alterations
+   * need to be made.
+   *
+   * @returns {Object} - { key : displayValue }
+   */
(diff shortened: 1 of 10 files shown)


jniles:

PR #2274, opened 2017-11-05
Use debug for debugging

### LINKED ISSUE #1519: (refactor/proposal) Use debug for logging

jniles:
This commit adds debug symbols to the database scripts.  This allows us to show debug logs for specific modules based on their module keys.

The default debug level is "app" to allow developers to see when the server has started.

Closes #1519.


jniles:
bors  r+

PR #2275, opened 2017-11-05
feat(journal): add showFullTransaction filter

### LINKED ISSUE #2266: Add new default filter for Journal: Show Full Transactions
I propose that we add a new default filter to the journal "Show Full Transactions".  This would be a radio button with a value of "Yes" or "No".

The effect of this radio button would be to always show the full transaction record for all transactions matching the filter.  For example, if you search for account 123 instead of just showing a single line with account 123 from each transaction, toggling `Show Full Transaction` to `Yes` would download the entire transaction for that has account 123 used in it.

We could accomplish this by having a post-processing filter.  After the regular filters have been applied, the Journal controller would get a unique list of record_uuids and simply filter on those transactions.  It would be a two step process.  If the user has not set the `Show Full Transactions`, this step would be skipped and the original filter would be sent.

This has been requested by our clients in Vanga.

jniles:
The journal filters have been extended with the option to show the entire transaction following a search on the Journal. This extension gives the user the ability to not only find an account view, but also a
full transaction view of all the data in the journal.

Note that the implementation is quite heavy (requires two passes on the Journal).  A warning message has been placed underneath the option to make the user aware of this.

Closes #2266.

showfulltransactionoption
_Fig 1: The modal for selecting transactions_

showfulltransactiongrid
_Fig 2: After filtering, the full transactions containing the query are shown_

### DIFF
--- a/server/controllers/finance/journal/index.js
+++ b/server/controllers/finance/journal/index.js
@@ -100,7 +100,7 @@ function naiveTransactionSearch(options, includeNonPosted) {
 // if posted ONLY return posted transactions
 // if not posted ONLY return non-posted transactions
 function buildTransactionQuery(options, posted) {
-  db.convert(options, ['uuid', 'record_uuid', 'uuids']);
+  db.convert(options, ['uuid', 'record_uuid', 'uuids', 'record_uuids']);
 
   const filters = new FilterParser(options, { tableAlias : 'p' });
 
@@ -149,6 +149,7 @@ function buildTransactionQuery(options, posted) {
   filters.custom('origin_id', 'p.origin_id IN (?)', [typeIds]);
 
   filters.custom('uuids', 'p.uuid IN (?)', [options.uuids]);
+  filters.custom('record_uuids', 'p.record_uuid IN (?)', [options.record_uuids]);
   filters.custom('amount', '(credit_equiv = ? OR debit_equiv = ?)', [options.amount, options.amount]);
 
   return {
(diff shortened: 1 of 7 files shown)


server/controllers/finance/journal/index.js
```diff
@@ -177,18 +178,42 @@ function find(options) {
...
+  return find({ record_uuids : records, includeNonPosted });
+}
+
 /**
  * @method list
  *
  * @description
  * This function simply uses the find() method to filter the posting journal and
- * (optionally) the general ledger.
+ * (optionally) the general ledger.  If the
```
DedrickEnc: Ha ha it seems like this comment is not completed

test/integration/journal.js
```diff
@@ -76,6 +76,24 @@ function SearchTests() {
       .catch(helpers.handler);
   });
 
+  it(`GET /journal?account_id=${account_id}&showFullTransaction should find complete transactions`, () => {
```
DedrickEnc: it is not ```&showFullTransaction=${showFullTransaction}``` ?

DedrickEnc (approved):

DedrickEnc:
@jniles 
LGTM
if you want, you can correct some comment and run bors

test/integration/journal.js
jniles: Yup!  I will fix that.

jniles:
@dedrickenc, I've fixed the remarks.  I've also included some screenshots to explain how the filtering works.

PR #2276, opened 2017-11-05
Feat: edit flag on transactions

### LINKED ISSUE #2047: (proposal) Underlying `EDITED` flag on financial entities
Currently there is no way of knowing if a BHIMA document has had its underlying transaction edited or changed during the posting process. There currently is not a use case for this however it may be useful to show on an invoice that the financial details of this document may have changed. 

I propose that when an editing session is complete on a transaction it then uses the `identifiers.js` map to lookup the reference table and set an `EDITED` flag to `TRUE` .

The `EDITED` flag would be added to: 
* invoices
* cash payments 
* vouchers

For example: 
```js
const transactionIdentifier = 'IV';
const documentTable = Identifiers[transactionIdentifier].table;

const sql = `UPDATE ${documentTable} SET edited = TRUE WHERE uuid = ?`;
db.exec(sql, transactionUuid)
  //...
```

jniles:
This PRimplements the EDITED flag on records of transactions that have been modified via the posting journal edit tool.  This signals the user that the transaction has been edited and the data might be out of
date.

This PR also improves the bhAutoDropup in some corner cases by accurately estimating the window size on startup and any resize events that occur. Previously, the first click might have an inaccurate window size. Furthermore, it implements debounce to ensure that the event loop is not blocked and the DOM is only manipulated a single time.

Finally, a client-side flag is added to the registries and it looks like this:
editedrecordflag
_Fig 1: Edited Record Flag_

Closes #2047.


jniles:

jniles:

DedrickEnc:
@jniles
I am sorry, I will have look this evening

DedrickEnc (approved):
LGTM

PR #2278, opened 2017-11-05
feat: implement bhHiddenField

### LINKED ISSUE #933: (Feature) Optional (important) field selection
Currently in service management cost and profit centers can be assigned using a normal select - they are optional fields but are essential to the correct functioning of the accounting. 
* Once a cost or profit center is allocated they can not be removed as the select has no null value, this is a bad user experience 
* No logic or rules exist considering if you can change a cost or profit center and the implications of that 

I suggest a component for important but optional data that allows the user to 'Add' the field, making the choice to add or remove the configuration makes it clear that it is optional. 

**Service management - optional fields not added**
service

**Service management - add cost center configuration**
service cost center

jniles:
This commit implements the `bhHiddenField` component which hides any transcluded HTML behind a toggle and lets the user decide when to show it and to hide it after.  The use case is for option fields in forms that do not need to be considered by default.  For example, many reports have complex options that should not be presented on the outset, but can instead by added in later.

Closes #933.

clientsreportclosed
_Fig 1: Closed Hidden Field_

clientsreportopen
_Fig 2: Open Hidden Field_

### DIFF
--- a/client/src/js/components/bhHiddenField.js
+++ b/client/src/js/components/bhHiddenField.js
@@ -1,0 +1,42 @@
+/**
+ * @overview bhHiddenField
+ *
+ * @description
+ * This component allows a programmer to optionally hide a field in a form at
+ * startup that may toggled open by the user if they need it.
+ */
+
+var template =
+ '<p>' +
+   '<a ng-click="$ctrl.toggleVisibility()" bh-hidden-field-toggle href>' +
+      '<span ng-hide="$ctrl.visible"> + <span translate>{{ $ctrl.showText }}</span></span>' +
+      '<span ng-show="$ctrl.visible"> - <span translate>{{ $ctrl.hideText }}</span></span>' +
+   '</a>' +
+ '</p>' +
+ '<div ng-if="$ctrl.visible">' +
+  '<ng-transclude></ng-transclude>' +
+ '</div>';
+
+
+angular.module('bhima.components')
+  .component('bhHiddenField', {
+    template : template,
+    transclude : true,
+    controller : bhHiddenFieldController,
+    bindings : {
+      showText : '@',
+      hideText : '@',
+    },
+  });
+
+function bhHiddenFieldController() {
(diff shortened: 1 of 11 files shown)


jniles:

DedrickEnc:
@jniles
Will do it this evening

DedrickEnc (commented):
@jniles

This feature is very important, it avoids information pollution in the system which very crucial for novices in computer system using (Vanga and Lukonga users). 
Information pollution can easily unoriente users.

However, it will be good to see a test for this component. It should expose a method to toggle the view for viewing/hiding components which are inside.

Great idea!

client/src/js/components/bhHiddenField.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * @overview bhHiddenField
+ *
+ * @description
+ * This component allows a programmer to optionally hide a field in a form at
+ * startup that may toggled open by the user if they need it.
+ */
+
+var template =
+ '<p>' +
```
DedrickEnc: This is so small but since we are not using ECMA 6 syntax to make string templating at the client side, you can create a separate file to make it easy to read.

jniles:
@dedrickenc, good idea.  Tests are always a good thing :+1:

I'll work on this in the next couple days - my internet connection is giving me a hard time at the moment!

jniles:
@dedrickenc, thank you for your review.

You will need to run `yarn install` or `npm install` to get `chai-dom` to run the unit tests locally.

I've implemented unit tests in Karma for the bhHiddenField component.  We should do this for all directives and components in the future.  I've also moved the template into `modules/templates`.  We should probably figure out how to put the JS and component templates side by side in the future.

Let me know what you think!

DedrickEnc:
Thanks for the feedback, I am working on the stock at Pax. Will review it this evening.

DedrickEnc (approved):
LGTM

PR #2285, opened 2017-11-14
chore: enable strict component bindings

### LINKED ISSUE #2223: Turn on strictComponentBindingsEnabled()
With the Angular 1.6.6. release, the option to turn on [strictComponentBindings](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#new-features) was added.  This means that if a component requires a parameter and it is not supplied, the application will break at runtime.   Using this option would enforce the component's API contract and ensure developers are properly using components the way the component designer intended.

We should experiment to see if we can get away with turning this option on.  It may be that our underlying libraries will break (e.g. ui-bootstrap).  Still, if possible, it would be one extra check to make sure developers are respecting the component APIs.

jniles:
This commit turns on parsing-time checks for bindings that are required
by angular components but not provided.  These checks ensure that a
component's API is respected.  Since they provide a slight performance
hit in production, they probably can be turned off in releases.

Closes #2223.

### DIFF
--- a/client/src/js/app.js
+++ b/client/src/js/app.js
@@ -171,6 +171,8 @@ function compileConfig($compileProvider) {
     $compileProvider.commentDirectivesEnabled(false);
     $compileProvider.cssClassDirectivesEnabled(false);
   }
+
+  $compileProvider.strictComponentBindingsEnabled(true);
 }
 
 /**


jniles:
Looks like this won't be possible since `uiGrid` doesn't respect the strict components binding option.

Issue #2217, opened 2017-10-19
Make a client TransactionService for actions on transactions


jniles:
At the moment, we have two methods that are unique to transactions.  These methods do not depend on whether the transaction is in the Posting Journal or General Ledger.  They are:
 1. `comment()` - attaches a comment to the transaction or set of transactions.
 2. `remove()` - deletes the transaction and associated rows in other tables.  Note that this will only succeed if the transaction is in the Posting Journal.

We could make a instead of having duplicated code, we could make a client-side service that holds these methods.  This service would the be shared with the Posting Journal, the General Ledger, and any other modules that need the shared functionality.

jniles:
I suppose this could also include `reverse()`.

PR #2286, opened 2017-11-14
Unify transaction functions into TransactionService

jniles:
This  PR centralizes many of the shared functions that apply to transactions into a TransactionService.  This includes:
 1. Commenting - previously scattered between the Account Statement, Posting Journal, and General Ledger.
 2. Remove - previously each financial service (Invoice, Voucher, and Cash) implemented their own removal code.  This unifies it under the Transaction Service.

Additionally, a default modal configuration has been applied to reduce the amount of configuration code we have to write.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/2217.


client/src/js/services/TransactionService.js
```diff
@@ -0,0 +1,76 @@
...
+      resolve : {
+        params :  function paramsProvider() { return rows; },
+      },
+    };
+
+    return Modal.open(config).result;
+  }
+
+  /**
+   * @method history
```
DedrickEnc: Actually the method is ```historyFn```

DedrickEnc (commented):
@jniles

LGTM

You can fix the comment mistake as mentioned in the review and run bors

jniles:
Changes landed.

bors r+

PR #2288, opened 2017-11-17
feature(Unificate Employee Patient)

### LINKED ISSUE #2282: (feature) Unification of employee and patient data
In the most recent updates, the employee management module allowed adding a patient each time an employee was registered, but the limitation of this approach was that when changing a patient's information these changes do not occur. it is also the reason why updating an employee's information was not easy to implement, it is an issue that the employee table can only keep the specific information of the employee :
- code
- date_embauche
- grade_uuid
- nb_spouse
- nb_enfant
- daily_salary
- bank
- bank_account
- fonction_id
- service_id
- creditor_uuid
- locked
- patient_uuid
- is_medical

lomamech:
- Update structure of employee table, remove all common between patient
  and employee table on Employee Table
- Update employee API on server side
- Implement the possibility to Update Employee Data
- Fix Update data on debtor and Creditor table On update Patient Data
- Update directive bhUnique add property origin to not to trigger the
  function in case you update the property hospital number

- Update integration and end to end test

closes #2282


client/src/i18n/fr/tree.json
```diff
@@ -17,11 +17,11 @@
     "COMPLEX_JOURNAL_VOUCHER":"Ajout de transactions",
     "COST_CENTER_MGMT":"Centres de coûts",
     "CREDIT_NOTE":"Note de Crédit",
-    "CREDITOR_GROUP":"Gestion de groupes crediteurs",
+    "CREDITOR_GROUP":"Gestion des groupes crediteurs",
```
jniles, line 20: :+1:

client/src/js/directives/bhUnique.js
```diff
@@ -49,7 +50,15 @@ function UniqueDirective($q, UniqueValidator) {
 
         UniqueValidator.check(validationUrl, viewValue)
           .then(function (valueExists) {
+            // Check if hospital_no change
+            var originUpdated = viewValue !== attrs.origin;
```
jniles, line 54: This is pretty cool!

I think you might be able to do this without adding the `origin` part.  Could you try putting this at the top of your scope and seeing if it will do the same thing?
```js
var originValue = ctrl.$viewValue;

// now you can just compare to originValue
```
This will make it easier on programmers since they will not be obligated to put `origin` in the html.

client/src/modules/employees/registration/employees.html
```diff
@@ -3,7 +3,16 @@
     <ol class="headercrumb">
       <li class="static" translate>TREE.PAYROLL</li>
       <li class="title" translate>EMPLOYEE.TITLE</li>
+      <li ng-if="EmployeeCtrl.isUpdating" class="title">{{ EmployeeCtrl.employee.name }}</li>
     </ol>
+
+    <!-- ng-show to stop flickering on load -->
```
jniles: I don't think `ng-show` is used here.  Maybe you mean `ng-if`?

client/src/modules/employees/registration/employees.html
```diff
@@ -3,7 +3,16 @@
...
+    <span
+        ng-if="EmployeeCtrl.isUpdating" 
+        style="margin-left : 5px"
+        class="label label-warning">
+      <span style="text-transform : uppercase" translate>FORM.LABELS.EDIT</span>
```
jniles: There's a class for this: `text-uppercase`.

client/src/modules/employees/registration/employees.js
```diff
@@ -63,15 +118,27 @@ function EmployeeController(Employees, Services, Grades, Functions, CreditorGrou
...
+      Employees.update(referenceId, vm.employee);
 
+    return promise
+      .then(function (feedBack) {
         // reset form state
         employeeForm.$setPristine();
         employeeForm.$setUntouched();
         vm.employee = {};
+
+        if (!referenceId) {
```
jniles, line 136: This is cool!

client/src/modules/employees/templates/action.cell.html
```diff
@@ -7,24 +7,14 @@
   <ul data-action="{{ rowRenderIndex}}"  class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
     <li class="dropdown-header">{{row.entity.code}}</li>
     <li>
-      <a data-method="edit" href>
+      <a data-method="edit" ui-sref="employeeEdit({id : row.entity.id})" href>
```
jniles, line 10: :+1:

server/controllers/medical/patients/index.js
```diff
@@ -246,6 +247,59 @@ function lookupPatient(patientUuid) {
...
+      const updateDebtor = `UPDATE debtor SET ? WHERE debtor.uuid = ?`;
+
+      const transaction = db.transaction();
+
+      transaction
+        .addQuery(updateDebtor, [debtorText, debtorUuid])
+        .addQuery(updateCreditor, [creditorText, creditorUuid]);
+
+      transaction.execute()
+      .then((res) => {
```
jniles: No need for this `.then` statement.  You should probably just do `return transaction.execute()`.

server/controllers/payroll/employees/index.js
```diff
@@ -221,18 +228,21 @@ function update(req, res, next) {
 
   const updateCreditor = `UPDATE creditor SET ? WHERE creditor.uuid = ?`;
   const updateDebtor = `UPDATE debtor SET ? WHERE debtor.uuid = ?`;
+  const updatePatient = `UPDATE patient SET ? WHERE uuid = ?`;
```
jniles, line 231: Nice!

server/controllers/payroll/employees/index.js
```diff
@@ -408,20 +427,21 @@ function find(options) {
...
+  filters.equals('sex', 'sex', 'patient');
   filters.equals('code', 'code', 'employee');
   filters.equals('service_id', 'service_id', 'employee');
   filters.equals('fonction_id', 'fonction_id', 'employee');
   filters.equals('grade_uuid', 'grade_uuid', 'employee');
   filters.equals('is_medical', 'is_medical', 'employee');
 
   // @TODO Support ordering query
-  filters.setOrder('ORDER BY employee.display_name DESC');
+  filters.setOrder('ORDER BY patient.display_name DESC');
```
jniles, line 443: 👍

server/models/schema.sql
```diff
@@ -618,19 +618,13 @@ DROP TABLE IF EXISTS `employee`;
 CREATE TABLE `employee` (
   `id`            INT(11) UNSIGNED NOT NULL AUTO_INCREMENT,
   `code`          VARCHAR(20) NOT NULL,
-  `display_name`  TEXT NOT NULL,
```
jniles, line 621: :+1:

jniles (commented):
Looks pretty good to me.  Great work @lomamech.

I've made a few comments.  The logic looks good though.  Could you write an end to end test for editing the employee page?

client/src/js/directives/bhUnique.js
lomamech: There is a problem, when initializing this variable, ctrl.$viewValue is empty

jniles: Hmm... I wondered if that might happen... Okay, that is fine then.  We can leave it as it is for now. Thanks for trying!

jniles (approved):
Looks good to me!

bors r+

PR #2289, opened 2017-11-17
feat(registry) standardize text sizes and labels for pdf

### LINKED ISSUE #2231: Standardize Text Sizes and Labels for Registry PDFs
As you can see from https://github.com/IMA-WorldHealth/bhima-2.X/pull/2230, each registry will often have its own size of text and style of rendering.  To look professional, we should really have a uniform output.

I propose the following:
 1. All registries should be called "X Registry".
 2. No registry should have `table-striped`.  This will waste ink when printed, and is much more useful for screens.
 3. Every registry should use `table-report` to make the font size smaller.
 4. Every registry should include "Number of Xs" in the bottom left hand corner instead of the "Total" in the bottom right.

jeremielodi:
This PR solves #2231 
closes #2231

### DIFF
--- a/client/src/i18n/en/patient_registry.json
+++ b/client/src/i18n/en/patient_registry.json
@@ -11,7 +11,7 @@
     "PAYMENTS":"View Cash Payments",
     "YEAR_OLD":"year old",
     "YEARS_OLD":"years old",
-    "TITLE":"Patients",
+    "TITLE":"Patient Registry",
     "RECORD":"Record"
   }
 }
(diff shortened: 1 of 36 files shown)


client/src/i18n/en/patient_registry.json
```diff
@@ -11,7 +11,7 @@
     "PAYMENTS":"View Cash Payments",
     "YEAR_OLD":"year old",
     "YEARS_OLD":"years old",
-    "TITLE":"Patients",
+    "TITLE":"Patients Registry",
```
jniles: Since this is a now singular, it should read "Patient Registry".

jniles (commented):
@jeremielodi looks pretty good!

There's one point in the issue that was ignored:
> Every registry should include "Number of Xs" in the bottom left hand corner instead of the "Total" in the bottom right.

An example of this is the Voucher Registry, which is missing a total completely.  See below:
voucherregistry
_Fig 1: Voucher Registry Missing Count_

jniles:
I guess that the count can be addressed in another PR.

bors r+

jeremielodi:
@jniles , I'm working on it, would you mind waiting a little please.

Issue #2292, opened 2017-11-17
(feature) Complete Payroll Module


lomamech:
This issue suggests the restoration of the old elements of the Payroll module of Bhima 1.x in Bhima 2.x while respecting the standards and principles of development of Bhima 2.x, this module are:

- Contribution
- Function
- Grade
- Periods
- Topics
- Taxes
- Holidays
- Off days

PR #2298, opened 2017-11-22
feature(Job Title Management)

lomamech:
- Implement Job Title(Function) Client Side,
- Implement test E2E for Job title management module

reference #2292


client/src/i18n/en/form.json
```diff
@@ -312,7 +312,7 @@
             "FROM_ACCOUNT": "From Account",
             "FROM": "From",
             "FOR": "For",
-            "FUNCTION": "Function",
+            "FUNCTION": "Job Title",
```
jniles: This is a much nicer name!  :+1:

client/src/i18n/fr/function.json
```diff
@@ -0,0 +1,18 @@
+{
+  "FUNCTION" : {
```
jniles: Since this is no longer translated as "Function", we should make all the keys "PROFESSION" or something similar.  This can be a separate PR if you wish.

client/src/modules/functions/functions.js
```diff
@@ -0,0 +1,98 @@
...
+
+  function loadFunctions() {
+    vm.loading = true;
+
+    Functions.read(null, { detailed : 1 })
+    .then(function (data) {
+      vm.gridOptions.data = data;
+    })
+    .catch(Notify.handleError)
+    .finally(function () {
```
jniles, line 67: Nice :+1:

client/src/modules/functions/functions.js
```diff
@@ -0,0 +1,98 @@
...
+      vm.loading = false;
+    });
+  }
+
+  // switch to delete warning mode
+  function deleteFunction(title) {
```
jniles: Instead of "title" which sounds like it might be a `String`, try calling this `profession`.  I thought it was a text string at first.

... alternatively, just pass in the `id`.  It looks like all you actually need is the `id`.

test/end-to-end/functions/functions.spec.js
```diff
@@ -0,0 +1,39 @@
+const helpers = require('../shared/helpers');
+const FunctionPage = require('./functions.page');
+const chai = require('chai');
+
+
+/** configuring helpers**/
+helpers.configure(chai);
+
+describe('Functions Management', () => {
```
jniles: This should be called `Job Titles Management` now, right?

jniles (approved):
@lomamech, this is pretty good code!

I've made a few comments - none of them are blocking.  Would you like to work on them before a merge or would you like me to merge these and make an issue tracking changes?

Great work!

jniles (approved):
LGTM.

bors r+

Issue #1145, opened 2017-02-08
(refactor) Cashflow report logic


sfount:
Our current cashflow report was the first report that was created in 2.x. There are number of standards that have been developed since then that should be applied to the cashflow report. 

The current cashflow report is calculated using Javascript for totalling and grouping. All of this logic should be converted to MySQL to bring it in line with the other reports as well as to leverage the strong Maths library in MySQL. 
An example of logic being transferred to MySQL can be found in https://github.com/IMA-WorldHealth/bhima-2.X/pull/651

This report logic should be carefully tested with the latest installation dataset before closing this issue.

jniles:
Closed in https://github.com/IMA-WorldHealth/bhima-2.X/pull/2578

PR #2302, opened 2017-11-23
feat(voucher registry) add total amount in the report

jeremielodi:
I added an important information in the voucher report.
Now it's possible to know how many voucher's currency is $ or Fc ...

### DIFF
--- a/server/controllers/finance/reports/vouchers/index.js
+++ b/server/controllers/finance/reports/vouchers/index.js
@@ -17,6 +17,8 @@ const pdf = require('../../../../lib/renderers/pdf');
 const barcode = require('../../../../lib/barcode');
 const identifiers = require('../../../../config/identifiers');
 
+const util = require('../../../../lib/util');
+
 const entityIdentifier = identifiers.VOUCHER.key;
 
 const RECEIPT_TEMPLATE = './server/controllers/finance/reports/vouchers/receipt.handlebars';
@@ -107,9 +109,13 @@ function report(req, res, next) {
 
   const data = { filters };
 
-  return Vouchers.find(options)
+    return Vouchers.find(options)
     .then(rows => {
       _.extend(data, { rows });
+      return Vouchers.totalAmountByCurrency(options);
+    })
+    .then((sumAmount) => {
+      _.extend(data, { totals : sumAmount });
       return reportInstance.render(data);
     })
     .then((result) => {
@@ -118,3 +124,4 @@ function report(req, res, next) {
     .catch(next)
     .done();
 }
+
--- a/server/controllers/finance/reports/vouchers/report.handlebars
+++ b/server/controllers/finance/reports/vouchers/report.handlebars
@@ -37,7 +37,20 @@
               <td>{{voucher.display_name}}</td>
             </tr>
           {{/each}}
+           
+          
         </tbody>
+      </table>
+ 
+      <table 
+        style='width:400px; float:right; margin-top:10px'
+        class="table table-condensed table-bordered table-report">
+        {{#each totals as |total|}}
+            <tr>
+              <td> {{total.numVouchers}} {{translate 'FORM.LABELS.VOUCHERS'}} ({{total.currencySymbol}})</td>
+              <td class="text-right">{{total.totalAmount}} {{total.currencySymbol}}</td>
+            </tr>
+          {{/each}}
       </table>
     </div>
   </div>
--- a/server/lib/util.js
+++ b/server/lib/util.js
@@ -21,7 +21,6 @@ module.exports.take = take;
 module.exports.loadModuleIfExists = requireModuleIfExists;
 exports.dateFormatter = dateFormatter;
 exports.resolveObject = resolveObject;
-exports.sumAmounts = sumAmounts;
 
 /**
  * @function take
@@ -118,20 +117,4 @@ function dateFormatter(rows, dateFormat) {
   });
 
   return rows;
-}
-
-// this function separate amounts by curreny_id
-
-function sumAmounts(rows) {
-  let tb = {};
-  for (let i = 0; i < rows.length; i++) {
-    let row = rows[i];
-    if (tb[row.currency_id]) {
-      tb[row.currency_id].amount += row.amount;
-      tb[row.currency_id].iteration += 1;
-    } else {
-      tb[row.currency_id] = { amount : row.amount, symbol : row.symbol, iteration : 1 };
-    }
-  }
-  return tb;
 }
--- a/server/lib/util.js
+++ b/server/lib/util.js
@@ -21,6 +21,7 @@ module.exports.take = take;
 module.exports.loadModuleIfExists = requireModuleIfExists;
 exports.dateFormatter = dateFormatter;
 exports.resolveObject = resolveObject;
+exports.sumAmounts = sumAmounts;
 
 /**
  * @function take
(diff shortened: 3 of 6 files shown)


server/controllers/finance/reports/vouchers/report.handlebars
```diff
@@ -37,8 +37,21 @@
...
+          
         </tbody>
       </table>
+ 
+      <table 
+        style='width:400px; float:right; margin-top:10px'
+        class="table table-condensed table-bordered table-report">
+        {{#each totals as |total|}}
+            <tr>
+              <td> {{total.iteration}} {{translate 'FORM.LABELS.VOUCHERS'}} ({{total.symbol}})</td>
```
mbayopanda: 👍 good

server/lib/util.js
```diff
@@ -118,3 +119,19 @@ function dateFormatter(rows, dateFormat) {
 
   return rows;
 }
+
+// this function separate amounts by curreny_id
+
+function sumAmounts(rows) {
```
jniles: Could you do this in SQL?  Javascript has problems calculation math correctly.

For reference, see the following issues: https://github.com/Vanga-Hospital/bhima-2.X/issues/204, https://github.com/IMA-WorldHealth/bhima-2.X/issues/1145.

For other reports that do totaling in SQL instead of JS for this reason, see [the cash report](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/reports/cash/index.js#L164) and the [invoice report](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/reports/invoices/index.js#L65).

For more thoughts on JS being bad at math, [here is an article on the subject](https://gooroo.io/GoorooTHINK/Article/16305/Is-Math-Broken-in-JavaScript-Part-1/18866)/

jniles (changes requested):

server/lib/util.js
```diff
@@ -118,3 +119,19 @@ function dateFormatter(rows, dateFormat) {
 
   return rows;
 }
+
+// this function separate amounts by curreny_id
```
mbayopanda: Could you add in the comment that this function requires that each row must contain : amount, currency_id and symbol

server/lib/util.js
```diff
@@ -118,3 +119,19 @@ function dateFormatter(rows, dateFormat) {
...
+
+function sumAmounts(rows) {
+  let tb = {};
+  for (let i = 0; i < rows.length; i++) {
+    let row = rows[i];
+    if (tb[row.currency_id]) {
+      tb[row.currency_id].amount += row.amount;
+      tb[row.currency_id].iteration += 1;
```
mbayopanda: Could rename `iteration` by something which more meaningful in this context such as `count`

mbayopanda (approved):
LGTM

server/lib/util.js
jeremielodi: You're right @jniles, but we have already a json array with all information we need for summing, doing another sql query won't be easy because it shoul take care all fiters or options that the previous query had.
When seaching the solution, I come across a libary witch can help to solve this probem, alasql
I'm investigating to see if good

jniles: Just use [FilterParser](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/filter.js).  It is a library @sfount wrote for this purpose.  [Here's a pretty complex example of how to use it](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/journal/index.js#L81). 

In this case, you'll just need to do this:
```js
const aggregateQuery = `
SELECT c.id as currencyId, c.symbol as currencySymbol, SUM(v.amount) as totalAmount, COUNT(c.symbol) AS numVouchers
FROM voucher JOIN currency ON voucher.currency_id = currency.id
GROUP BY currency.id`;

// options here is req.query
const filters = new FilterParser(options);
  filters.period('period', 'date');
  filters.dateFrom('custom_period_start', 'date');
  filters.dateTo('custom_period_end', 'date');
  filters.equals('user_id');
  filters.equals('edited');

  filters.custom('reference', referenceStatement);

  filters.fullText('description');

  // @todo - could this be improved
  filters.custom('entity_uuid', 'v.uuid IN (SELECT DISTINCT voucher_uuid FROM voucher_item WHERE entity_uuid = ?)');

  filters.custom('type_ids', 'v.type_id IN (?)', [typeIds]);

  // @todo - could this be improved
  filters.custom('account_id', 'v.uuid IN (SELECT DISTINCT voucher_uuid FROM voucher_item WHERE account_id = ?)');

  filters.custom('invoice_uuid', REFERENCE_SQL, [options.invoice_uuid, options.invoice_uuid]);
  filters.custom('cash_uuid', REFERENCE_SQL, [options.cash_uuid, options.cash_uuid]);

  // @TODO Support ordering query (reference support for limit)?
  filters.setOrder('ORDER BY v.date DESC');

  const query = filters.applyQuery(aggregateQuery);

  return db.exec(aggregateQuery, parameters);
```

Since all this code already exists in [the voucher controller](https://github.com/jeremielodi/bhima-2.X/blob/7fe5c121ec16a1d17465a7bcf224e3fc9b273fda/server/controllers/finance/vouchers.js), I would suggest you make a function that simply takes in the base SQL and the options.  This function would be called `buildVoucherQuery`.  It would then return the query and parameters after applying all the filters to it.

Like this:
```js
const aggregateQuery = `
SELECT c.id as currencyId, c.symbol as currencySymbol, SUM(v.amount) as totalAmount, COUNT(c.symbol) AS numVouchers
FROM voucher JOIN currency ON voucher.currency_id = currency.id
GROUP BY currency.id`;

const { sql, params } = vouchers.buildVoucherQuery(aggregateQuery, options);
return db.exec(sql, params);
```

server/controllers/finance/reports/vouchers/index.js
```diff
@@ -107,9 +109,13 @@ function report(req, res, next) {
...
   const data = { filters };
 
-  return Vouchers.find(options)
+    return Vouchers.find(options)
     .then(rows => {
       _.extend(data, { rows });
+      return Vouchers.totalAmountByCurrency(options);
+    })
+    .then((sumAmount) => {
+      _.extend(data, { totals : sumAmount });
```
jniles, line 118: This is much better!  Since our math is done is SQL now, it has the following advantages:
 1. It will not block the server - this code will allow other HTTP requests to be handled as fast as possible.  Since NodeJS is [single-threaded](https://stackoverflow.com/a/17959746/6251968), the previous JS function would block the event loop while it figured out what the totals were.  This would slow the system down for every user in the system.  Using MySQL is the most optimal solution.
 2. Since JS is pretty bad at math, this code is guaranteed to be correct.  You've eliminated two classes of bugs: bugs due to type incompatibilities (instead of concatenating strings, MySQL will throw an error.  JS would never tell you something went wrong) and bugs due to schema changes (JS would simply add `undefined + undefined = NaN` while SQL will throw an error). 

Just FYI, the code can be made to run faster (on multiprocessor machines) by running these two queries in parallel.  Since `Vouchers.find()` and `Vouchers.totalAmountByCurrency()` do not depend on each other, you could actually do this:
```js
return q.all([
  Vouchers.find(options),
  Vouchers.totalAmountByCurrency(options),
]).spread((rows, totals) => {

  // concurrency win!
  _.extend(data, { rows, totals});
});
```

This is a pretty advanced performance concept though and probably won't net a lot of gain on a slower machine.

jniles (approved):
@jeremielodi this code is much better than the first version!  You've eliminated several classes of bugs and wrote optimized code.  Well done!

We'll figure out how to combine the Voucher filters into a single function as @sfount did for transactions in the future.  It will keep our code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  This code is already very good as it is.

PR #2303, opened 2017-11-24
 feature(Cotisation Management)

lomamech:
- Implement API for Cotisation Management
- Implement test Integration for Cotisation Management
- Implement client side for cotisation Management
- Update bhAccountSelect to allow the filter of account by Classe
- Fix error in Account Management Set the property classe in Registred and
  Updated Account
- Implement test E2E for Cotisation Management

reference #2292


client/src/i18n/en/cotisation.json
```diff
@@ -0,0 +1,18 @@
+{
+  "COTISATION" : {
+    "ADD_COTISATION" : "Add a Cotisation",
```
jniles: It looks like cotisation is [membership fee](https://translate.google.com/#auto/en/cotisation) in English.  I would propose calling it an "employment fee" so it doesn't get confused with other types of fees.

client/src/js/components/bhAccountSelect.js
```diff
@@ -37,6 +38,9 @@ function AccountSelectController(Accounts, AppCache, $timeout, bhConstants, $sco
     // cache the title account ID for convenience
     $ctrl.TITLE_ACCOUNT_ID = bhConstants.accounts.TITLE;
 
+    // To filter accounts by class of accounts
+    $ctrl.classe = $ctrl.classe || '';
```
jniles, line 42: The account class should be the same thing as the account type.  Can you call this account type and use that?

client/src/modules/cotisations/cotisations.routes.js
```diff
@@ -0,0 +1,42 @@
...
+      params : {
+        cotisation : { value : null },
+        creating : { value : true },
+      },
+      onEnter : ['$uibModal', cotisationModal],
+      onExit : ['$uibModalStack', closeModal],
+    })
+
+    .state('cotisations.edit', {
+      url : '/:id/edit',
```
jniles, line 21: :+1:

server/controllers/finance/accounts/index.js
```diff
@@ -134,19 +137,17 @@ function list(req, res, next) {
...
-  if (req.query.locked) {
-    locked = Number(req.query.locked);
-  }
+  filters.equals('classe');
 
-  // if locked is a number, filter on it
-  if (!isNaN(locked)) {
-    sql += ` WHERE a.locked = ${locked}`;
-  }
+  filters.equals('locked');
```
jniles, line 142: Nice!

test/end-to-end/cotisations/cotisations.page.js
```diff
@@ -0,0 +1,133 @@
...
+   */
+  cancelDeleteCotisation(n) {
+    GA.clickOnMethod(n, this.actionLinkColumn, 'delete', this.gridId);
+    components.modalAction.dismiss();
+  }
+
+  /**
+   * forbid deletion of used function
+   */
+  errorOnDeleteCotisation(n) {
```
jniles: Can you make this refer to a text instead of an index?  It will make the test much easier to modify in the future.

jniles (changes requested):
@lomamech, a few comments here.

I'm pretty sure account `classe` and account type_id are the same thing - especially after the modifications by @mbayopanda which added in Actif and Passif and so on.  Could you use account type_id instead?

Otherwise, the code looks pretty good!

lomamech:
@DedrickEnc @jniles

In the optics that must be added, the four account and the six account at the table "rubric", it seems good to me to gather all these different elements in one table, so I want to block all the PR in course (Rubric, Tax and Contribution) and merge them into a single Table payroll rubric

Issue #2299, opened 2017-11-22
Installation feature for bhima


mbayopanda:
This issue is about to have an installation feature in bhima, when we want to initialize bhima for a particularly institution this feature will help by :

- Allowing to create a super user for bhima, define enterprise information (name, abbreviation, currency) and define also the initial project of the enterprise.

For having this feature, the proposal is : 

1. To have a route `/install` which help us to check if the application is already installed and also to run other installation scripts

2. On the client side, we define a particular route for the installation page. When there is no installation we can access to installation page, but when the application is already installed this route on client will redirect us to the login page or display a message (need discussion).

Having a page in the client side is important because it can help us to enhance this feature with accounts import, inventories import, etc.

This issue is about a first implementation, which suppose that before install the application via the installation module the database  must be built with basic information (the database will allow us to check also if bhima is already installed)

Feel free to give your suggestions 😃

jniles:
It seems to me this can be broken down into the following steps:
 1. Create the Enterprise (and defines the default currency)
 2. Create one or more projects.
 3. Create one or more users.

One question is ... should the app simply cache these values until the very end of the process and then click "Write values and complete" or should it write the values at the end of each step?  I'm in favor of the first, since it will be easier to detect when the database is in an undefined state.  It's an "all or nothing" approach.

jniles:
Closed in #2304.

PR #2304, opened 2017-11-27
Installation script

mbayopanda:
This PR concerns the first implementation of the installation process related to the issue #2299 

- [ ] tests : 
  - [ ] end-to-end
  - [x] integration


client/src/modules/install/install.js
```diff
@@ -0,0 +1,51 @@
...
+// controller definition
+function InstallApplicationController(InstallService, $state, Notify) {
+  var vm = this;
+
+  vm.setup = { enterprise : {}, project : {}, user : {} };
+
+  // expose methods to the view
+  vm.submit = submit;
+
+  InstallService.checkBasicInstallExist()
```
jniles: This is pretty cool!

It looks like what this code will do is basically this:
navigate to /install -> load InstallApplicationController -> fire HTTP GET /install -> if already exist -> $state.go('login').

You could probably remove the step that loads the install controller if the `alreadyExist` flag is false in `onEnter` part of the `$state` change.  Something like this:
```js
onEnter: ['InstallService', '$state', function (InstallService, $state) {
  return InstallService.checkBasicInstallExist()
     .then(function (isInstalled) { 
       if (isInstalled) { $state.go('login'); }
     });
}]
```

Trying to catch the `isInstalled` flag before navigating to the route should make sure that there isn't a flash of content before re-routing.

client/src/modules/install/install.js
```diff
@@ -0,0 +1,51 @@
...
+
+  function notifyInstallSucess() {
+    Notify.success('INSTALL.SUCCESS_INSTALL');
+    $state.go('login');
+  }
+
+  function checkPassword() {
+    if (!vm.setup.user.password) { return false; }
+
+    if (vm.setup.user.password !== vm.setup.user.repassword) {
```
jniles: Do we not have a service for this?  Or a directive/component?

client/src/modules/install/install.js
```diff
@@ -0,0 +1,51 @@
...
+    return true;
+  }
+
+  function submit(form) {
+    if (form.$invalid) { return 0; }
+
+    if (checkPassword() === false) { return Notify.warn('USERS.PASSWORD_MISMATCH'); }
+
+    return InstallService.proceedInstall(vm.setup)
+      .then(notifyInstallSucess)
```
jniles: Nice function names 👍

server/app.js
```diff
@@ -25,11 +25,16 @@
 // Workaround - required before 'use-strict'
 const json2xls = require('json2xls');
 
+
 require('use-strict');
+const dotEnv = require('dotenv');
```
jniles, line 30: :+1:

server/controllers/install.js
```diff
@@ -0,0 +1,123 @@
...
+ */
+function basicInstallExist() {
+  // check users, enterprise, projects
+  const userExist = 'SELECT COUNT(*) AS total_users FROM user;';
+  const enterpriseExist = 'SELECT COUNT(*) AS total_enterprises FROM enterprise;';
+  const projectExist = 'SELECT COUNT(*) AS total_projects FROM project;';
+
+  const dbPromise = [db.one(userExist), db.one(enterpriseExist), db.one(projectExist)];
+
+  return Q.all(dbPromise)
```
jniles: Nice!

server/controllers/install.js
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @description
+ * check if the basic information for installation exist, in the case
+ * it doesn't exist we can perform a new installation
+ */
+function basicInstallExist() {
+  // check users, enterprise, projects
+  const userExist = 'SELECT COUNT(*) AS total_users FROM user;';
+  const enterpriseExist = 'SELECT COUNT(*) AS total_enterprises FROM enterprise;';
+  const projectExist = 'SELECT COUNT(*) AS total_projects FROM project;';
```
jniles: I know JavaScript will treat 1 as `true`, but why don't you just go ahead and make the SQL return a boolean?  Like:
```js
const userExist = 'SELECT COUNT(*) > 1 AS hasUsers FROM user;'
```
or something?

server/controllers/install.js
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * POST /install
+ *
+ * proceed to the application installation
+ */
+exports.proceedInstall = (req, res, next) => {
+  const { enterprise, project, user } = req.body;
+
+  basicInstallExist()
+    .then(ans => {
```
jniles: Instead of `ans` a better name would be `isInstalled`.

server/controllers/install.js
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    // prepare transactions
+    const transaction = db.transaction();
+    const sqlEnterprise = 'INSERT INTO enterprise SET ? ';
+    const sqlProject = 'INSERT INTO project SET ? ';
+    const sqlUser = `
+      INSERT INTO user (username, password, display_name) VALUES (?, PASSWORD(?), ?);`;
+
+    const sqlUnitPermission = `
```
jniles, line 108: :+1:

jniles (commented):
Awesome work!  The code is very readable, you have nice variable names, nice refactoring to help ESLint along.  Overall, really great work.

I've made a few comments to try and improve the logic and get rid of that flash of unrendered content.  Let me know if it works or not.  We might also explore a method using `resolve` instead of `onEnter` to resolve a promise before navigating and change the behavior based on that.

It doesn't look like `test/data-init.sql` is used ... if you are going to use it for integration tests or end to end tests, that is good.

mbayopanda:
@jniles could you review this PR ?

client/src/modules/install/install.js
jniles: It looks like [UserService.validatePassword()](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/users/users.service.js#L124) will do this for you.  Could you use that?

test/end-to-end/locations/provinces.spec.js
```diff
@@ -1,25 +1,26 @@
...
+describe('Provinces Management', () => {
 
   const path = '#!/locations/province';
 
   before(() => helpers.navigate(path));
 
   const province = {
-    country : 'Test Hook Country',
-    name : 'New Province'
+    country : 'République Démocratique du Congo',
```
jniles, line 19: Nice cleanup :+1:

test/integration/locations.js
```diff
@@ -262,8 +250,8 @@ describe('(/locations) Locations Interface', function () {
   });
 
   it('PUT /locations/countries/:uuid should update a Country', function () {
-    return agent.put('/locations/countries/' + country.uuid)
-      .send({ name : 'Update New Country'})
+    return agent.put(`/locations/countries/${country.uuid}`)
```
jniles, line 253: Nice cleanup! 👍

jniles (approved):
Code looks pretty good!  I've made a few non-blocking comments.

How do I test this locally?

jniles:
Looks like everything is addressed and ready to work here.

bors r+

PR #2309, opened 2017-11-28
feat(unit test) unit test for the server

### LINKED ISSUE #1442: (tests) Implement unit tests for the server
We have a few libraries are the server that we should implement stricter testing standards for.  These include:
 1. `/errors/*.js`
 2. `topic.js`
 3. `db/*.js`
 4. `util.js`

There may be others, but these are core libraries used in most server operations.

**Overview**
Basically, you should make a command called `npm run test:server-unit`.  We already have `npm run test:client-unit` which runs karma.  This command will use [mocha](https://mochajs.org/) and [chai](http://chaijs.com/) which are already included in the repository.

You should make a test suite in `test/server-unit/` that will be a single file for every file tested.  It will `require()` the file to test at the top of the function and run a series of tests on it.  For example:
```js
const util = require('../../server/lib/util');
const { expect } = require('chai');

describe('util.js', function () {
  it('#take() should take values from one key of each object in an array of objects', () => {
    const objects = [{ id: 1 }. { id: 2 }, { id: 3}];
    const expected = [1,2,3];
    expect(util.take(objects, 'id')).to.equal(expected);
  });
});
```

jeremielodi:
This PR attempt to implement the server side unit test as suggested in #1442

Closes #1442.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test:ends": "./sh/test-ends.sh",
     "test:lint": "./sh/lint.sh",
     "test:client-unit": "./node_modules/.bin/karma start --single-run --no-auto-watch --concurrency 1 karma.conf.js",
-    "test:server-unit": "./node_modules/.bin/mocha --recursive test/server-unit",
+    "test:server-unit": "./node_modules/.bin/mocha --recursive --exit test/server-unit",
     "build": "./node_modules/.bin/gulp build",
     "build:db": "./sh/build-database.sh",
     "app_windows": "gulp build && cd bin && set NODE_ENV=production && node server/app.js",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test:ends": "./sh/test-ends.sh",
     "test:lint": "./sh/lint.sh",
     "test:client-unit": "./node_modules/.bin/karma start --single-run --no-auto-watch --concurrency 1 karma.conf.js",
-    "test:server-unit": "echo 'Not yet implemented.'",
+    "test:server-unit": "./node_modules/.bin/mocha --recursive test/server-unit",
     "build": "./node_modules/.bin/gulp build",
     "build:db": "./sh/build-database.sh",
     "app_windows": "gulp build && cd bin && set NODE_ENV=production && node server/app.js",
--- a/sh/test.sh
+++ b/sh/test.sh
@@ -32,6 +32,11 @@ startfold "Running Client Unit Tests..." "test-client-unit";
 ./node_modules/.bin/karma start --single-run --no-auto-watch --concurrency 1 karma.conf.js
 endfold "test-client-unit" ;
 
+#run server-unit test
+startfold "Running server Unit Tests ......" "server-unit"
+./node_modules/.bin/mocha --recursive test/server-unit
+endfold "server-unit" ;
+
 # run end to end tests
 startfold "Running Client Unit Tests..." "test-end-to-end";
 ./sh/test-ends.sh
--- a/sh/test.sh
+++ b/sh/test.sh
@@ -34,7 +34,7 @@ endfold "test-client-unit" ;
 
 #run server-unit test
 startfold "Running server Unit Tests ......" "server-unit"
-./node_modules/.bin/mocha --recursive test/server-unit
+./node_modules/.bin/mocha --recursive --exit test/server-unit
 endfold "server-unit" ;
 
 # run end to end tests
--- a/test/server-unit/db.spec.js
+++ b/test/server-unit/db.spec.js
@@ -1,0 +1,44 @@
+const path = require('path');
+
+const env = path.resolve(__dirname, '../../.env.development');
+
+require('dotenv').config({ path : env });
+
+const db = require('../../server/lib/db');
+
+const { expect } = require('chai');
+
+const _ = require('lodash');
+
+// console.log(process.env);
+
+const dbTest = function () {
+
+  it('Should check the connection to mysql', function (done) {
+    var fx = db.pool.getConnection(function (err, result) {
+      if (err) {
+        done(err);
+        return;
+      }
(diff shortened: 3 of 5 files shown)


test/server-unit/db.spec.js
```diff
@@ -0,0 +1,43 @@
+const db = require('../../server/lib/db');
+const { expect } = require('chai');
+var _ = require('lodash');
+
+var dbTest = function () {
+
+  before(function () {
+    var params = {
```
jniles: Ideally, we should use the values found in `.env.development`.  You can do that like this:
```js
require('dotenv').config({ path : '../../.env.development' });

// process.env.DB_NAME and other environmental variables are now defined. 
// We can just load db with a require() and everything will work as expected.

const db = require('../../server/lib/db');
```

test/server-unit/db.spec.js
```diff
@@ -0,0 +1,43 @@
...
+      done();
+    });
+
+  });
+
+  it('#Should try to retrieve data from a specific table (unit)', () => {
+    return db.exec('SELECT * FROM unit', {})
+      .then((rows) => {
+				const test = rows.length > 0;
+        expect(test).to.equal(true);
```
jniles: One of the advantages of chai is that it makes common operations like this easy - using [lengthOf](http://chaijs.com/api/bdd/#method_lengthof) you can chain this together.  Like this:
```js
.then(rows => expect(rows).to.have.lengthOf(2));
```

As a rule, tests should be as specific as possible.  That way, you know each function is performing exactly the way you expect it to at all times.  If any change happens, a test will break and you will have to change either the test or fix the function.

test/server-unit/db.spec.js
```diff
@@ -0,0 +1,43 @@
...
+      })
+      .catch((error) => {
+        return error;
```
jniles: This code will ignore the error, but we want tests to fail if there is an error that occurs.  It looks like [mocha supports promise errors](http://tobyho.com/2015/12/16/mocha-with-promises/), so you can just remove the `.catch()` statement and it should handle the error automatically.

test/server-unit/util.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    var ids = _.flatMap(objects, filter);
+    expect(ids).to.deep.equal(expected);
+  });
+
+  it('#requireModuleIfExists() should require module if it exists', () => {
+    var exists = util.loadModuleIfExists('chai');
+    expect(exists).to.equal(true);
+  });
+
+  it('#dateFormatter() should format each javascript datetime value in a array of objects', () => {
```
jniles, line 19: This is a really nice test! :+1:

package.json
```diff
@@ -12,7 +12,7 @@
     "test:ends": "./sh/test-ends.sh",
     "test:lint": "./sh/lint.sh",
     "test:client-unit": "./node_modules/.bin/karma start --single-run --no-auto-watch --concurrency 1 karma.conf.js",
-    "test:server-unit": "echo 'Not yet implemented.'",
+    "test:server-unit": "./node_modules/.bin/mocha --recursive test/server-unit && set NODE_ENV=development node server/app.js",
```
jniles: Do you need the `&& set NODE_ENV=development node server/app.js` part of this?  Unit tests usually do not need an entire server.

jniles: Also, could you add this command to the `./sh/test.sh` command?  It should come after the "lint tests" and before the "integration tests".

test/server-unit/db.spec.js
```diff
@@ -0,0 +1,43 @@
...
+      user     : 'bhima',
+      password : 'HISCongo2013',
+      database : 'bhima_test',
+    };
+    db.init(params);
+
+  });
+
+  it('Should check the connection to mysql', function (done) {
+    var fx = db.pool.getConnection(function (err, result) {
```
jniles: For callbacks, the best way to test them is with "spies".  There's a chai library for it called [chai-spies](https://github.com/chaijs/chai-spies).

I'll work on this once the branch gets merged as an example.

jniles (commented):
@jeremielodi, this is a great start!  Your tests for `util.js` look very good.  Once this lands, we can work on all the library modules on the server.

If you add these tests to `./sh/test.sh` they will be automatically run by Travis on every PR.  That will ensure that they always pass.

sh/test.sh
```diff
@@ -32,6 +32,11 @@ startfold "Running Client Unit Tests..." "test-client-unit";
 ./node_modules/.bin/karma start --single-run --no-auto-watch --concurrency 1 karma.conf.js
 endfold "test-client-unit" ;
 
+#run server-unit test
+startfold "Running server Unit Tests ......" "server-unit"
```
jniles, line 36: Nice :+1:

test/server-unit/db.spec.js
```diff
@@ -0,0 +1,44 @@
...
+      expect(true).to.equal(true);
+      done();
+    });
+
+  });
+
+  it('#Should try to retrieve data from a specific table (unit)', function (done) {
+     db.exec('SELECT * FROM unit LIMIT 2', {})
+      .then((rows) => {
+        expect(rows).to.have.lengthOf(2);
```
jniles, line 32: Nice 👍

test/server-unit/util.spec.js
```diff
@@ -0,0 +1,38 @@
+const util = require('../../server/lib/util');
+const { expect } = require('chai');
+var _ = require('lodash');
+
+const newLocal = function () {
```
jniles, line 5: In the future, it's probably a good idea to call this function "UtilTests" or something like that.  That way the function name has meaning.

jniles (approved):
@jeremielodi great work!  This is an excellent start, and should give us a good starting point to test all aspects of the server.

Code LGTM.

bors r+

Issue #2308, opened 2017-11-28
Implement a scheduling library


jniles:
There are many actions we might want to schedule:
 1. Database backups (https://github.com/IMA-WorldHealth/bhima-2.X/pull/2305)
 2. File synchronization (https://github.com/IMA-WorldHealth/bhima-2.X/issues/2071) and uploads.
 3. Emailing reports on a regular basis (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1766)
 4. Do system checks (such as the "is installed" check in #2299) on reboot.

All of these features require some sort of scheduler that can run at regular intervals such as daily, weekly, and monthly. 

A similar feature is already proposed in https://github.com/IMA-WorldHealth/ima-data/issues/61.  I'll copy the relevant information here:

**Scheduler**
We will write a library called "scheduler" that will contain the schedules (monthly, weekly, daily, real time) in cron syntax.  This will be stored in an object `schedules`.  Basically:

| id  | schedule | label | description |
|:-: |:-:|:-:|:-:|
| 1 | 0 2 * * 0 | weekly | Weekly - this runs once a week at 2AM on Sunday |

We would find a library for Node that handles schedules in a CRON syntax - like [node-cron](https://www.npmjs.com/package/node-cron).  We would write a small wrapper that allows any modules to use a schedule like this:
```js
const scheduler = require('./lib/scheduler');

scheduler.daily(() => {
  console.log('I will be logged daily');
});

scheduler.weekly(() => {
  console.log('I get fired weekly.');
});

scheduler.custom('* * * * *', () > {
  console.log('I do something every minute');
});
```

Our backup scripts, email scripts, and anything else we want to schedule would be able to just import the scheduler and then use it as it.

jniles:
This was taken care of by adding a dependency on [TaskList](https://github.com/IMA-WorldHealth/TaskList).

Issue #1688, opened 2017-05-26
design: weekly email report


jniles:
We've talked about email reporting for a long time, but we'll need to start somewhere to actually refine the procedure.  I propose that we begin with a weekly email report that we hand-craft ourselves.  Then, following client feedback, we can remove fields, add fields, or change the frequency as needed.

**Prior Art**
For reference and inspiration, see an email from bhima 1.x below:
![daily email report - bhima 1 x](https://cloud.githubusercontent.com/assets/896472/26497975/4af15172-4225-11e7-8c4a-2e079d534437.png)
_Fig 1: Daily Email Report from 1.X_

**Technical Design**
In 1.x, we used `child_process.exec()` calls to the mail command to send email.  This meant that all error handling, retrying etc, was handled ourselves.

This time, I suggest we leverage [mailgun](http://mailgun.com/) for emailing, retries, bounces, etc.  It has the following advantage:
 1. Allows us to monitor the progress of sending mail from afar.  In 1.x, we had to physically be on site to update the recipients.  With mailgun, we won't have to.
 2. Manages mailing lists for us.
 3. Both @sfount and I have used it extensively.
 4. Has a nice [nodejs library](https://www.npmjs.com/package/mailgun-js) that is actively maintained.

**Report Design**
These are the following questions the report should address.  The can be altered as the report is developed if the information is not present or shown to be irrelevant.

 1. How many users used the application this week?
 2. How many patients were registered this week, categorized by age range?
 3. How many payments were made this week?  How does that compare to the average number of payments per week?
 4. How much money entered the main coffres?  How much exited?
 5. How many invoices were made versus how many were closed (paid in full)?  This should filter out credit notes.
 6.  What were some of the most purchased items this week?   This is defined as sold to the most number of patients.  What was the average quantity and price of those sales?

sfount:
This is a great jumping off point for integrating emails. I second the recommendation to use mailgun for this feature. 

Another thing that may be useful to see on a weekly basis is how many patients are being invoiced only once or to surface patients that seem very similar to previous patients. One of the major practical problems during installation has been the re-registering of patients so a metric that might help reduce that may be useful to provide.

jniles:
Hey, that's a good idea.  Or even patients that are never invoiced.  Both those would provide insight into how the posts are relating to each other.

DedrickEnc:
Yes, I think it is time to have this report but I have a question : 
[...]

PR #2310, opened 2017-11-28
feat(reports): implement report emailing

jniles:
This commit implements the email dropdown on reports as a first step
towards automatic email reporting.  The automatic email reports will
depend on a scheduler proposed in #2308.

The email functionality depends on having a mailgun account. You should
put your mailgun API key and domain in the .env.development file.  Some
defaults for this project have been filled out.

Partially addresses #1688 and #1766.

emailactivatedindropdown
_Fig 1: The Email Link Dropdown_

emailmodal
_Fig 2: The Email Modal.  It defaults to the current user's email address_

emailreceivedwithattachment
_Fig 3: Email in my inbox_


client/src/i18n/fr/report.json
```diff
@@ -131,7 +131,8 @@
       "BACK":"Revenir en arriere",
       "OPTIONS":"Options",
       "PREVIEW":"Aperçu",
-      "VIEW_ARCHIVE":"Archives"
+      "VIEW_ARCHIVE":"Archives",
+      "EMAIL_HELP_TXT" : "L'e-mail ci-dessus recevra le rapport en pièce jointe. Le serveur doit avoire une connexion Internet pour que l'e-mail soit livré avec succès."
```
DedrickEnc: Aie ... **avoir** surelly a typo ~avoire~

client/src/modules/reports/baseReports.service.js
```diff
@@ -63,6 +65,25 @@ function BaseReportService($http, Modal, util, Languages) {
       .then(util.unwrapHttpResponse);
   }
 
+  function emailReport(uuid, email) {
+    var url = '/reports/archive/'.concat(uuid, '/email');
+    return $http.put(url, { address : email })
```
DedrickEnc: Why a `PUT` method? We are not updating the report (resource) right?

client/src/modules/reports/modals/reports.email.js
```diff
@@ -0,0 +1,31 @@
...
+  'SessionService',
+];
+
+function EmailReportController(ModalInstance, Notify, SavedReports, options, Session) {
+  var vm = this;
+
+  vm.reportName = options.reportName;
+
+  vm.params = {
+    email : Session.user.email,
```
DedrickEnc, line 15: 👍

client/src/modules/reports/modals/reports.email.js
```diff
@@ -0,0 +1,31 @@
...
+  };
+
+  vm.submit = function submit(EmailForm) {
+    if (EmailForm.$invalid) { return 1; }
```
DedrickEnc, line 19: Why not just return?

server/controllers/report.js
```diff
@@ -158,10 +166,74 @@ function deleteArchived(req, res, next) {
...
+ */
+function emailArchived(req, res, next) {
+  const { uuid } = req.params;
+  const { address } = req.body;
+
+  debug(`#emailArchived(): Received email request for ${address}.`);
+
+  lookupArchivedReport(uuid)
+    .then(report => {
+      debug(`#emailArchived(): sending ${report.label} to ${address}.`);
```
DedrickEnc, line 202: Is there a difference between `debug` and `winston`?

server/controllers/report.js
```diff
@@ -158,10 +166,74 @@ function deleteArchived(req, res, next) {
...
+
+      const attachments = [
+        { filename, path : report.link },
+      ];
+
+      // template parameters for the email
+      const parameters = {
+        filename,
+        date,
+        user : report.display_name,
```
DedrickEnc, line 215: What is the difference between the user and the requester?

server/lib/mailer.js
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @description
+ * This function implements emailing with nicer handling of file names than the
+ * default mailgunjs library.  If attachments are provided, they are converted
+ * into mailgun attachments with either their filename as the attachment name or
+ * a default.
+ */
+exports.email = function email(address, subject, message, options = {}) {
+  debug(`#email() sending email ${subject} to ${address} with options %j`, options);
+  return processAttachments(options.attachments)
```
DedrickEnc, line 79: No method handle error, so it will be handled outside?

DedrickEnc (commented):
@jniles 

I made some comments and asked some questions to know more about the code. I waiting for your clarification.

Well done!
Thanks!

client/src/modules/reports/baseReports.service.js
jniles: Yeah, I needed a `body` in the request and only `PUT` or `POST` requests have a body.  I just chose PUT.  Do you think POST is better?

client/src/modules/reports/modals/reports.email.js
jniles: It was to satisfy ESLint.  We should probably turn that rule off - it causes more harm than good, I think.

server/controllers/report.js
jniles: Good question!  Winston uses "error levels" (winston.log(), winston.error(), winston.warn()), which debug simply namespaces logs by module.  It's like having `console.log()` statements that you can turn on and off.

This is advantageous because you can turn all logs off except the ones you care about, or you can filter logs based on what module it came from.  For example, here we log only email logs by using `DEBUG=mailer` and it will only print the `debug()` calls that are in the `mailer` module.  That way, if there are errors in the email script, we will be able to easily track them down in production by either filtering through the logs or turning off all logs except the email logs.

[This issue](https://github.com/IMA-WorldHealth/bhima-2.X/issues/1519) gives more of a detailed overview about why debug() works better for our needs than winston.

server/controllers/report.js
jniles: The user created the report originally and archived it.  The requester is the person asking for the email to be sent.   The AG might create the report, but the MD might request it to be sent somewhere.

server/lib/mailer.js
jniles: Yes, it is handled [here](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2310/files#diff-4e6ab4c187c9d8c8b986b3d4f8ba3b94R229).

I think we will want to pass back errors to the module that requests the email to be sent.  That way, the parent module can decide what to do - throw the error or retry the email or take another action.

jniles:
@dedrickenc, Thanks for the review.  I've fixed the typo - good catch!

I need your feedback on the POST vs PUT issue.  Either one works - which makes the most sense to you?

client/src/modules/reports/baseReports.service.js
DedrickEnc: Yes, let use POST if possible

client/src/modules/reports/modals/reports.email.js
DedrickEnc: Yes you are right

server/controllers/report.js
DedrickEnc: Thank you for the explanation

server/lib/mailer.js
DedrickEnc: Good idea

DedrickEnc:
@jniles
Every thing is fine, let just change the PUT method to POST one. And you can run bors.

jniles:
Thanks!

bors r+

PR #2318, opened 2017-11-30
feat(server unit test) tests for dates lib

### LINKED ISSUE #2317: Write tests for helpers/dates.js

### date.js
The following test cases should be covered in the unit tests for [lib/template/helpers/dates.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/template/helpers/dates.js):
 - [x] `date()` should format a date is 'DD/MMYYYY' format.
 - [x] `date()` should return an empty string (`''`) if the date is null.
 - [x] `date()` should allow you to specify a custom format.

 - [ ] `timestamp()` should return an empty string (`''`) if the value is null.
 - [x] `timestamp()` should format a date as `DD/MM/YYYY HH:mm:ss`.

 - [x] `age()` should return 0 for the current date.
 - [x] `age()` should return 3 for a date three years ago.

 - [x] `month` should return the full month name in English for a given month.

jeremielodi:
This PR write tests for helpers/dates.js  as described in this issue => #2317
closes #2317

### DIFF
--- a/test/server-unit/dates.spec.js
+++ b/test/server-unit/dates.spec.js
@@ -1,0 +1,66 @@
+const dates = require('../../server/lib/template/helpers/dates');
+const { expect } = require('chai');
+
+
+const DateHelperUnitTests = function () {
+  it('#date() should format a date is "DD/MM/YYYY" format.', () => {
+    const dat = new Date('2015-03-25');
+    const expected = '25/03/2015';
+    const formated = dates.date(dat);
+    expect(formated).to.equal(expected);
+  });
+
+  it('#date() should return an empty string ("") if the date is null.', () => {
+    const dat = null;
+    const expected = '';
+    const formated = dates.date(dat);
+    expect(formated).to.equal(expected);
+  });
+
+  it('#date() should allow you to specify a custom format.', () => {
+    const dat = new Date('2015-03-25');
+    const expected = '03/2015';
+    const format = 'MM/YYYY';
+    const formated = dates.date(dat, format);
+    expect(formated).to.equal(expected);
+  });
+
+  it('#timestamp() should return an empty string ("") if the date is null.', () => {
+    const dat = null;
+    const expected = '';
+    const formated = dates.timestamp(dat);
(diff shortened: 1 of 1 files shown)


test/server-unit/dates.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    let format =  'MM/YYYY';
+    var formated = dates.date(dat, format);
+    expect(formated).to.equal(expected);
+  });
+
+  it('#timestamp() should return an empty string ("") if the date is null.', () => {
+    let dat =  null;
+    var expected = '';
+    var formated = dates.timestamp(dat);
+    expect(formated).to.deep.equal(expected);
```
jniles: For strings, you don't need `deep`.  The `deep` call will recursively descend into objects.

test/server-unit/dates.spec.js
```diff
@@ -0,0 +1,68 @@
+const dates = require('../../server/lib/template/helpers/dates');
+const { expect } = require('chai');
+
+
+const newLocal = function () {
```
jniles: Could you give this meaning with a name?  Something like
```js
function DateHelperUnitTests() {

}
```

test/server-unit/dates.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+  it('#age() should return 0 for the current date.', () => {
+    let dat =  new Date();
+    var expected = 0;
+    var formated = dates.age(dat);
+    expect(formated).to.equal(expected);
+  });
+
+  
+  it('#age() should return 3 for the current date.', () => {
```
jniles: I think this should read `#age() should return 3 for a date three years ago`.

jniles (commented):
@jeremielodi this looks really good!  The tests are well defined and will make sure that we detect any hidden API changes in the future.  Well done!  :+1:

I've made a few comments to clean up the code and better explain for future coders.  Could you make sure this passes `ESLint`?  After that, this can be merged no problem.

jniles (approved):
LGTM.

bors r+

PR #2319, opened 2017-11-30
feat(server unit test) write test for translate.js

### LINKED ISSUE #2316: Write unit tests for server/lib/helpers/translate.js

### translate.js
The following test cases should be covered in the unit tests for [translate.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/helpers/translate.js):
 - [ ] The `getTranslationHelper()` should return a function.
 - [ ] Using an unknown language key should default to English.
 - [ ] The function returned by `.getTranslationHelper('en')` should translate a string in English.
 - [ ] The function returned by `.getTranslationHelper('fr')` should translate a string in French.

jeremielodi:
This PR write tests for helpers/translate.js  as described in this issue => #2316
closes #2316

### DIFF
--- a/server/lib/helpers/translate.js
+++ b/server/lib/helpers/translate.js
@@ -1,8 +1,13 @@
 /* eslint import/no-unresolved:off */
 const _ = require('lodash');
+const path = require('path');
 
-const en = require('../../../client/i18n/en.json');
-const fr = require('../../../client/i18n/fr.json');
+const frPath = '../../../bin/client/i18n/fr.json';
+const enPath =  '../../../bin/client/i18n/en.json';
+
+const en = require(enPath);
+
+const fr = require(frPath);
 
 /**
  * @function getTranslationHelper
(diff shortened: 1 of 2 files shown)


jeremielodi:
Hi @jniles, the travis test  fails because of the required module in translate.js.
It seems that it uses a compiled file "en.json" or "fr.json".
I have tried to have a look at rewrite library but I notice it rewrite the path only in the test file. In their examples   the path to rewrite is called within test But our case we have : 
require('../lib/translate');
and the paths to rewrite are in translate.js

jniles:
@jeremielodi, this looks pretty complicated.  I think there should be a way to make this work with [rewire](https://github.com/jhnns/rewire)  or another module to aid in testing.  I'll see if I can figure this out at some point this week if you don't find a solution first!

jeremielodi:
Okay, it will help a lot even in other tests

PR #2332, opened 2017-12-05
fix(inventory): account select validation

jniles:
This commit fixes the validation on the `bhAccountSelect`s for inventory group accounts.  Only the sales account is required.  If it is left blank, a validation message is shown.  A test has been written to make sure this bug never happens again.

It also incorporates various ESLint fixes.


jniles:
~~Actually, let me finish an end to end test for this before it gets reviewed and merged to prevent a regression.~~

Done.

.env.development
```diff
@@ -10,7 +10,7 @@ DB_NAME='bhima_test'
 # session variables
 SESS_SECRET='XopEn BlowFISH'
 
-DEBUG=db
+DEBUG=app
```
DedrickEnc, line 13: What is the meaning of this? can you comment?

client/src/modules/inventory/configuration/groups/groups.js
```diff
@@ -26,18 +27,24 @@ function InventoryGroupsController($translate, InventoryGroup, Account, Notify,
   // startup
   startup();
 
+  function handler(err) {
+    if (err) {
```
DedrickEnc, line 31: Why not using `Notify` even if ```err``` is ```undefined``` and print a general error message?

DedrickEnc (commented):
@jniles 

No blocking comment for this P.R, just some questions to understand more your logic.

.env.development
jniles: This changes the log level to only show when the app has started and loaded environmental variables.  Previously, @mbayopanda set it to print every db execution making the environmental variable "db".  The problem with the "db" value is that it overflows Travis's logs and makes them unreadable.  So I was switching it back to its default behavior.

The way debug works is described in its README.  You can read all about it [here](https://www.npmjs.com/package/debug).  It's worth reading for when we have to debug errors in production!

client/src/modules/inventory/configuration/groups/groups.js
jniles: Because when `err` does not exist, it means the modal has been canceled by the user.  The way it works is like this:
 1. User clicks the [cancel button](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/inventory/configuration/groups/modals/actions.tmpl.html#L72) which calls [cancel()](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/inventory/configuration/groups/modals/actions.modal.js#L64) in the controller.
 2. This fires the [ModalInstance.dismiss()](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/inventory/configuration/groups/modals/actions.modal.js#L65) method.  According to [UI Bootstrap](https://angular-ui.github.io/bootstrap/#!#modal) this `dismiss()` method rejects the promise returned by the modal.
 3. The rejection makes the promise chain terminate in [this .catch() statement](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/inventory/configuration/groups/groups.js#L40).  However, since the `ModalInstance.dismiss()` method wasn't called with any parameters, the `err` parameter is empty.  This tells us that it wasn't an actual error from the HTTP request - those all throw real errors.  It was just a cancellation by the user.

DedrickEnc:
LGTM
bors r+

PR #2338, opened 2017-12-05
feat(fiscal year module) display message when no fiscalyear registered so far

### LINKED ISSUE #2327: Fiscal Year looks broken when no fiscal years registered
Since our[installation script (https://github.com/IMA-WorldHealth/bhima-2.X/pull/2304) landed, we can get a look at what the application looks like when there is very little data in the application.  One problem area is an empty fiscal year page looks broken - there is no indication that the application is loading or needs more information.  See below:

Here's what you should do:
 1. Put a loading indicator as a small circle with `fa-spin` on the page.
 2. Wait until the fiscal years are downloaded.  If there are no fiscal years, hide the loading indicator and display a message like:
 > There are no fiscal years registered so far!  Click "Create a Fiscal Year" to add your first fiscal year.

jeremielodi:
This PR helps to display message to inform the user when no fiscal year was registered so far.
The he can create a new one.
closes #2327

### DIFF
--- a/client/src/modules/fiscal/fiscal.js
+++ b/client/src/modules/fiscal/fiscal.js
@@ -11,11 +11,11 @@ function FiscalController($state, Fiscal, ModalService, Notify, $window) {
 
   // pagination configuration
   /** @todo this should all be moved to a component */
-  vm.pageSize     = 5;
-  vm.currentPage  = 1;
-  vm.fiscalYears  = [];
-  vm.maxSize      = 5;
-  vm.del          = del;
+  vm.pageSize = 5;
+  vm.currentPage = 1;
+  vm.fiscalYears = [];
+  vm.maxSize = 5;
+  vm.del = del;
 
   vm.sort = sort;
   vm.state = $state;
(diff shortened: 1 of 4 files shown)


client/src/modules/fiscal/fiscal.js
```diff
@@ -25,17 +25,26 @@ function FiscalController($state, Fiscal, ModalService, Notify, $window) {
...
+  vm.loadingError = false;
+
+
   // refresh Fiscal Year
   function refreshFiscalYear() {
-    return Fiscal.read(null,{ detailed: 1 })
+    return Fiscal.read(null, { detailed : 1 })
       .then(function (fiscalYears) {
         vm.fiscalYears = fiscalYears;
+      }).catch(() => {
```
jniles: Unfortunately, we can't use arrow functions on the client - they break our minified build.  Could you change all these to `function () {` calls?

client/src/modules/fiscal/fiscal.js
```diff
@@ -25,17 +25,26 @@ function FiscalController($state, Fiscal, ModalService, Notify, $window) {
...
+        vm.loadingError = true;
+        Notify.handleError();
```
jniles: Notify expects to be passed an error object.  You should use the error that is returned from the server.  Like this:
```js
.catch(function (err) {
  vm.loadingError = true;
  Notify.handleError(err);
})
```

client/src/modules/fiscal/fiscal.js
```diff
@@ -45,19 +54,18 @@ function FiscalController($state, Fiscal, ModalService, Notify, $window) {
...
-      Fiscal.delete(fiscal.id)
-        .then(function () {
-          Notify.success('FORM.INFO.DELETE_SUCCESS');
-          return refreshFiscalYear();
-        })
-        .catch(function (error) {
-          Notify.danger('FISCAL.CAN_NOT_DELETE_FY');
-        });
-    });
+      .then((bool) => {
```
jniles: All of these arrow functions will need to be migrated to `function () {` calls to avoid breaking our production builds.

jniles (commented):
Looks pretty good!

Unfortunately, the arrow functions can only be used on the server for the moment.  We'll need to figure out how to compile to ES5 before we use fat arrow functions on the client :(

The rest looks pretty good!  Nice work.

client/src/modules/fiscal/fiscal.js
jniles: This still needs to be addressed, and then it can be merged.

jniles (approved):
Everything LGTM!

bors r+

PR #2339, opened 2017-12-05
fix(vouchers): bhAccountSelect on Simple Vouchers

jniles:
This commit adds the bhAccountSelect to the simple vouchers module. This comes with all the benefits of the bhAccountSelect, including excluding title accounts and vastly simplified code.


DedrickEnc (approved):
LGTM
bors r+

PR #2343, opened 2017-12-05
feat(creditor group) hide locked option at registration

### LINKED ISSUE #2335: Hide the "Locked" option on Creditor Group creation
The Creditor Group create form (`/#!/creditors/groups/create`) has an option for locking a creditor group on creation.  See below:

lockedcreditorgrouponcreate
_Fig 1: Lock a newly created creditor group_

It's hard to imagine a use case for locking a creditor group while it is being created.  Normally a creditor is only blocked after they fall out of use.  We should adopt the same practice proposed in https://github.com/IMA-WorldHealth/bhima-2.X/issues/2012 and hide this option on the create page.

jeremielodi:
This PR hide the locked option at registration, as suggested at #2335
closes #2335

### DIFF
--- a/client/src/modules/creditors/groups/creditor-groups-manage.html
+++ b/client/src/modules/creditors/groups/creditor-groups-manage.html
@@ -30,7 +30,7 @@
             </bh-account-select>
 
             <!-- is locked ?  -->
-            <div class="checkbox">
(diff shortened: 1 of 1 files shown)


jniles:
LGTM.

bors r+

PR #2344, opened 2017-12-05
fix(patient groups): require description

jniles:
This commit fixes the patient groups form making the description a required field.  It is a required field in the database so this catches the error on the client before it is even submitted.

Additionally, the client now uses Notify to alert the any errors on data download or group submission.

### DIFF
--- a/client/src/modules/patients/groups/groups.html
+++ b/client/src/modules/patients/groups/groups.html
@@ -31,7 +31,6 @@
               <tr>
                 <th translate> TABLE.COLUMNS.NAME </th>
                 <th translate> TABLE.COLUMNS.PRICE_LIST </th>
-                <th translate> TABLE.COLUMNS.NOTE </th>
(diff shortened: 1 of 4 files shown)


client/src/modules/patients/groups/groups.html
```diff
@@ -108,7 +106,7 @@ <h4 translate> PATIENT_GROUP.TITLE </h4>
                   <label class="control-label" translate>
                     FORM.LABELS.NOTES
                   </label>
-                  <textarea class="form-control" ng-maxlength="PatientGroupCtrl.maxLength" name="note" ng-model="PatientGroupCtrl.patientGroup.note"></textarea>
+                  <textarea class="form-control" ng-maxlength="PatientGroupCtrl.maxLength" name="note" ng-model="PatientGroupCtrl.patientGroup.note" required></textarea>
```
mbayopanda, line 109: 👍

mbayopanda (approved):
LGTM

PR #2345, opened 2017-12-05
feat(patient registration) correct message for date of birth input

### LINKED ISSUE #2336: Year of Birth option does not say "Enter only Year of Birth" when first rendered
On the Patient Registration page, we allow a user to put in either a date of birth or a year of birth.  Unfortunately, the link that allows a user to toggle between the year of birth and the date of birth does not render correctly and states "Enter Full Date of Birth" when it should say "Enter only Year of Birth".  See below:

bugdateofbirthshouldsayyearofbirth
_Fig 1: Bug: This is should say "enter year of birth"_

As you can see above, it says is already on the date of birth input.  The link is false.  When you click it, it will allow you to enter the year of birth.  It should like this:

correctdateofbirth
_Fig 2: The correct rendering of the Year Of Birth toggle_.

This bug should be fixed as soon as possible.  It is misleading and confusing.

jeremielodi:
When first rendered in the patient registration page, the Year of Birth option does not say "Enter only Year of Birth" . this PR solves that problem.
closes  #2336

### DIFF
--- a/client/src/modules/patients/registration/registration.js
+++ b/client/src/modules/patients/registration/registration.js
@@ -24,7 +24,7 @@ function PatientRegistrationController(Patients, Session, util, Notify,
   vm.submit = submit;
   vm.toggleFullDate = toggleFullDate;
   vm.calculateYOB = calculateYOB;
-  vm.dateIndicatorLabel = 'FORM.LABELS.ENTER_BIRTH_DAY';
+  vm.dateIndicatorLabel = 'FORM.LABELS.ENTER_BIRTH_YEAR';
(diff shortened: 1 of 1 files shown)


jniles (changes requested):
This doesn't fix the problem.  The problem looks like it comes from the data initialized on [line 27](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/patients/registration/registration.js#L27).  This should be initialized with `ENTER_BIRTH_YEAR` instead of `ENTER_BIRTH_DAY`.  That should fix the issue.

jniles:
LGTM.  

bors r+

PR #2347, opened 2017-12-05
fix account list after adding first account

### LINKED ISSUE #2330: Account Management no longer refreshes the data on insert
The Account Management module used to respond to `create()` by refreshing the grid and showing the newly added accounts.  For some reason this no longer happens and forces the user to refresh the page to see the changes made. 

To fix this, you should take a look at the update code.  It's likely that the account grid simply needs to register the change with a `dataChange` event.  This is much more preferable to re-downloading all accounts like we do for other `search()` actions on the registries.

### LINKED ISSUE #2331: Account Management will not remove the last account from the view when deleted
This may be related to #2330.

When the last account is deleted in the Account Management grid, it remains in the view though it is deleted in the database.  The only way to get it to disappear is by refreshing the page.  This leads to some confusing errors and user experience.  See below:
deletelastaccount
_Fig 1: Error - Account Already Deleted!_

**Expected Result**
The account should be removed from the view when it is deleted.

**Actual Result**
The account remains in the view but is removed from the database.

mbayopanda:
This PR fix the account list after adding the first account.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/2330.
Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/2331.


client/src/modules/accounts/accounts.service.js
```diff
@@ -38,34 +38,39 @@ function AccountService(Api, $http, util, bhConstants) {
...
+    return accounts;
+  }
+
+  function humanReadableLabel(account) {
+    account.hrlabel = label(account);
   }
 
   function label(account) {
-    return account.number + ' - ' + account.label;
+    return String(account.number).concat(' - ', account.label);
```
jniles, line 59: :+1:

client/src/modules/accounts/accounts.service.js
```diff
@@ -38,34 +38,39 @@ function AccountService(Api, $http, util, bhConstants) {
...
   }
 
   function filterTitleAccounts(accounts) {
-    return accounts.filter(function (account) {
-      return account.type_id !== bhConstants.accounts.TITLE;
-    });
+    return accounts.filter(handleFilterTitleAccount);
+  }
+
+  function handleFilterTitleAccount(account) {
```
jniles, line 72: 👍

client/src/modules/accounts/edit/accounts.edit.js
```diff
@@ -196,58 +205,71 @@ function AccountEditController($rootScope, $state, AccountStore, Accounts,
     }
 
     if (vm.isCreateState) {
+      handleAccountCreateState();
+    } else {
+      handleAccountElseState();
```
jniles: This isn't a very descriptive name.  If you aren't in the `CREATE` state, you must be in the `UPDATE` state. So a better name for this is `handleAccountUpdateState()`.

client/src/modules/accounts/edit/accounts.edit.js
```diff
@@ -196,58 +205,71 @@ function AccountEditController($rootScope, $state, AccountStore, Accounts,
...
-          vm.fetchError = null;
-          $rootScope.$broadcast('ACCOUNT_UPDATED', result);
-          Notify.success('ACCOUNT.UPDATED');
-          close();
-        })
+        .then(handleAccountElseResult)
         .catch(handleModalError);
     }
+
+    function handleAccountElseResult(result) {
```
jniles: This should mention something about updating in its name.

test/data.sql
```diff
@@ -1,3 +1,6 @@
+SET names 'utf8';
```
jniles, line 1: Nice 👍

jniles (approved):
LGTM.  I've made a few comments, but they are all non-blocking.

jniles:
You can "bors r+" this when ready 👍

PR #2352, opened 2017-12-06
feat(patient invoice) warn if no invetory items available

### LINKED ISSUE #2340: Patient Invoice should warn if no Inventory Items are available
If you try to create a patient invoice without any inventory items, the module will stubbornly refuse to add lines to the Patient Invoice grid without alerting the user to a problem.  This is quite confusing.  Since the Patient Invoice Form downloads the inventory items, we should use that to detect if the application is missing inventory items and alert the user if there are no inventory times present.

jeremielodi:
This PR handle a notification if no inventory items are available, more description about the problem here #2340.

close #2340


jniles:
There are a lot of lines changed here ... I thought this would be about four lines to change.  What happened?

jeremielodi:
I was trying to detect the right place to solve this problem, that is why. the inventory service is a little bit complexe....

client/src/modules/invoices/patientInvoiceForm.service.js
```diff
@@ -102,20 +101,23 @@ function PatientInvoiceFormService(Patients, PriceLists, Inventory, AppCache, St
...
       .then(function (data) {
-
+        if (data.length === 0) {
+          Notify.danger('INVENTORY.NO_INVENTORY_REGISTERED');
+          return;
+        }
         // make sure both the label and code is searchable
         data.forEach(function (item) {
-          item.hrlabel = item.code + ' ' + item.label;
+          item.hrlabel = `${item.code  } ${  item.label}`;
```
jniles, line 112: :+1:

jniles (approved):
LGTM!  In the future, we will probably just want to put a permanent message in the grid as a warning, but this requires a proposal.

bors r+

PR #2356, opened 2017-12-06
Fix: misc registry fixes

### LINKED ISSUE #2355: The link from the Payments Registry -> Patient Registry is broken
It looks like the link from that Payment Registry is filtering the Patient Registry on patient group instead of on `patient_uuid`.  This is a bug.

jniles:
This PR fixes two bugs.  The first was a duplicated item in the dropdown of users.  See below:

usersdropdownbroken
_Fig 1: Duplicated Dropdown Item Disactivate_

usersdropdownfixed
_Fig 2: Fixed it!_

And the second is the link from the Cash Registry to the Patient Registry which previously did nothing.

Closes #2355.

### DIFF
--- a/client/src/modules/cash/payments/templates/action.cell.html
+++ b/client/src/modules/cash/payments/templates/action.cell.html
@@ -22,12 +22,12 @@
 
     <!-- view linked records -->
     <li>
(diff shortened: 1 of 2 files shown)


jniles:

PR #2358, opened 2017-12-07
feat(debtor group) cancel at registration

### LINKED ISSUE #2334: "Cancel" button on create page does not return to Debtor Groups List
On the Debtor Groups Management Page, when creating a new group the "cancel" button doesn't do anything.  It should take the user back to the list view.  The editing view's "cancel" button works - you might refer to that code if you want to fix this.

jeremielodi:
The cancel action was not working at the debtor group registration. this PR solves that problem.
closes #2334


jniles (approved):
bors r+

PR #2359, opened 2017-12-07
feat(invetory item) add currency symbol next to the price label

### LINKED ISSUE #2341: Show the currency of the price when adding an Inventory Item
The Inventory Item create form has an input for the default price.  For someone not familiar with BHIMA, it is unclear what currency value the price will be in.  We should put the enterprise currency symbol somewhere to let the user know it will be in the enterprise's currency.

See below:
showcurrencyoninventorypriceinput
_Fig 1: What currency will be used?_

jeremielodi:
For more understanding, the currency symbol should be display some where in the inventory item registration. So this PR helps to do it.
You can see, at the bellow picture, how the modal looks like now
symbolecurrency

closes #2341


jniles (approved):
Looks great!

bors r+

PR #2360, opened 2017-12-07
Warning for no debtor groups

### LINKED ISSUE #2324: Patient Registration no longer alerts the user if there are no debtor groups
Previously, the debtor group select would have a red error message underneath it if there were no debtor groups.  Unfortunately, when we switched to the component version some time ago, it appears that we did not bring along this functionality.  We should re-implement a nice error message so that the user is aware they need to create some debtor groups.

I suggest this be a flag: `warn-on-no-groups = true`.  This will allow the component to `ngShow` an error if there are no debtor groups downloaded.  By default, this flag is false.

mbayopanda:
This PR add a warning message on the `bhDebtorGroupSelect` for no debtor groups

close #2324


client/src/js/components/bhDebtorGroupSelect.js
```diff
@@ -40,15 +40,18 @@ function DebtorGroupSelectController(DebtorGroup) {
...
     DebtorGroup.read(null, filters)
-      .then(function (debtorGroups) {
-        $ctrl.debtorGroups = debtorGroups;
-      });
+      .then(handleDebtorGroups);
 
     $ctrl.valid = true;
   };
 
+  function handleDebtorGroups(debtorGroups) {
```
jniles, line 49: Could you make this configurable?  We use debtor groups selects in a couple places (reports, search modals) where they aren't necessarily required.  The original issue proposes a flag on the component, kind of like `excludeTitleAccounts` on `bhAccountSelect`.  What do you think?

mbayopanda: We can put the message at the bottom : 
image

The alert message can be shown only if the attribute `warn-no-group` is true. The input box is always displayed

client/src/js/components/bhDebtorGroupSelect.js
```diff
@@ -12,6 +12,7 @@ angular.module('bhima.components')
       required          : '<?',
       validationTrigger : '<',
       filter            : '<?',
+      warnNoGroup       : '<?',
```
jniles, line 15: 👍 Beautiful.

client/src/modules/templates/bhDebtorGroupSelect.tmpl.html
```diff
@@ -19,4 +19,9 @@
       <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
     </div>
   </div>
+  <div ng-if="$ctrl.warnNoGroup && $ctrl.noDebtorGroups" class="alert alert-danger">
```
jniles, line 22: Excellent!

jniles (approved):
Looks great to me!  Good work!

bors r+

PR #2361, opened 2017-12-07
fix(db, UI): make transfer account not required

### LINKED ISSUE #2357: Do not require "transfer account" for a cashbox
The "transfer account" doesn't make sense to have in every case.  If an enterprise (like a small clinic) only has a single caisse, what is the transfer account for that?  Instead it should be optional.

DedrickEnc:
This pull request update the db to make transfer account null by default and update the UI in order to let the user creating a cash-box without being required to provide a transfer account.
closes #2357


jniles (approved):
LGTM.

bors r+

PR #2362, opened 2017-12-07
feat(cashbox configuration) fix  error message position

### LINKED ISSUE #2337: Sloppy Error on the Cashbox Currency Configuration Modal
The Cashbox Management page allows a user to configure accounts used for the each currency associated with a cashbox.  The error handling on this page is really sloppy.  See below:
sloppyerroroncashboxmodal
_Fig 1: This error pushes the button out of the way_

This is a really sloppy way to handle errors.  It would be much better if the error appeared above the buttons at least, or maybe even in the modal body itself.  For example:

bettererroroncashboxmodal
_Fig 2: A better error positioning_

You don't have to implement the proposal above ... but it should avoid moving the buttons too much.  It's distracting when an error message suddenly replaces the button you clicked on!

jeremielodi:
This PR fixes the error message position at the cashbox configuration modal as recommended at #2337. 
closes #2337

### DIFF
--- a/client/src/modules/cash/cashboxes/configure_currency/modal.html
+++ b/client/src/modules/cash/cashboxes/configure_currency/modal.html
@@ -3,7 +3,8 @@
     <ol class="headercrumb">
       <li class="static" translate>CASHBOX.TITLE</li>
       <li class="title">
-        <span translate>CASHBOX.CONFIG_ACCOUNTS</span> <small>{{ CashboxModalCtrl.cashbox.text }}</small>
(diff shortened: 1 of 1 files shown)


jniles:
Hey @jeremielodi, I think you text editor keeps rewriting most of the spacing in the file.  Could you fix this?  It is probably a configuration on your editor.

It makes it hard to review the code not knowing what actually changed.

jniles (approved):
LGTM.

bors r+

PR #2363, opened 2017-12-07
feat(debtor group) alert message when no group registered

### LINKED ISSUE #2324: Patient Registration no longer alerts the user if there are no debtor groups

jeremielodi:
This PR add an alert message where there is no debtor group registered from so far.
closes #2324


jniles:
Duplicate of #2360.

PR #2372, opened 2017-12-09
fix(currency): use decimal point 1000s place FC

### LINKED ISSUE #2371: Use decimal points as 1000s separators for FC
We currently have two ways of rendering currency:
 1. Using the `currency` filter on the client, which uses spaces to render the 1000s place.  Like this: `1 000 000,00FC`.
 2. Using the handlebars currency helper on the server, based on `accounting-js`.  This one renders 1000s with a decimal point.  Like this: `1.000.000,00FC`.

We should use one or the other.  In my view, the second option is much more readable.  It wouldn't break the currency over two lines if there is a lack of space, and it doesn't look like two different numbers sitting close together.  See below:
decimalcurrency
_Fig 1: Decimal Point 1000s separator_

Unfortunately, I cannot find any guidance online as to which is the correct one for the CDF.  Could we just use the second option?  What do you think @IMA-WorldHealth/bhima-core ?

jniles:
This commit uses a decimal point instead of white space to separate the
FCs thousands place.  It makes the currency much more readable on the
client.

Closes #2371.

### DIFF
--- a/client/src/i18n/currency/fc.json
+++ b/client/src/i18n/currency/fc.json
@@ -1,7 +1,7 @@
 {
   "CURRENCY_SYM": "Fc",
   "DECIMAL_SEP": ",",
-  "GROUP_SEP": "\u00a0",
+  "GROUP_SEP": ".",
   "PATTERNS": [
     {
       "gSize": 3,


DedrickEnc (approved):
LGTM

Issue #2364, opened 2017-12-07
Add an option for "account-type-filter" on bhAccountSelect


jniles:
We have implemented some finer grained account types: actif, passif, capital, recette, depense.  See https://github.com/IMA-WorldHealth/bhima-2.X/issues/1607 for a discussion on going further with these ideas.

At the moment, they are just theoretical labels added to an account.  We can make them more powerful by using them as a filter for the `bhAccountSelect`, by specifying an account type to filter the list on.  This would enable us to only show expense accounts (for example) on a report that requires expense accounts.  Alternatively, we could only show actif accounts for debtor groups.  This should help the user make the right choice of accounts.

Issue #2240, opened 2017-10-23
Update APIs to use FilterParser


jniles:
[PR #2239](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2239/files#diff-0d44e4623388e794b089cc7db3d361e4R115) made me realize that not all server side code is using FilterParser.  We should do a scan of all server APIs that use filters and migrate them to using `FilterParser`.  This will enable us to easily update them with any future filters as needed.

Issue #1607, opened 2017-05-05
(proposal) Make finer-grained account types


jniles:
We could catch a lot of errors early by knowing if a certain account is supposed to contain debtors or not.  Or if a certain account should have suppliers in it.  We could do this with finer-grained account types, but it would require some infrastructure.

I propose we create a system where certain account types can be marked as `debtor` accounts or `cash` accounts, or a variety of different account types.  This would drastically reduce errors by preventing a patient from being assigned a stock account type for example - we could filter the accounts a debtor group may be attached to make sure that they are debtor accounts.

The advantage of this is that an enterprise must only set up their chart of accounts correctly and they can avoid a large class of user errors.  Furthermore, the errors will be contained to "invoiced the incorrect debtor account" rather than "invoiced the stock account".

There are a couple ways to implement this - either by creating different account types with an "income/ expense/balance" label (e.g. the cash account is always a balance account, the sale account is always an income account).  Or we could make two different account types (more complex from a user perspective).  I don't think either should be considered unless we believe that this proposal is a good idea.

Interesting in @IMA-WorldHealth/bhima-core's feedback.

mbayopanda:
The main thing that we must have in mind is for what purpose (or usage) we have to finer-grained account types. If there are concrete needs in term of modules or reports that need that, i think it is crucial to have that `finer-grained account types`. 

- What modules or reports need that ?

- There is a column in the table `account` named `classe`, this column cannot help to that needs ?

I think it is good to have a finer grained account types but it must answer concretes needs for modules or reports for avoiding waste time.

jniles:
There are two use cases that I can see being useful:
 1. Limiting a dropdown to only `debtor` accounts based on an account type `DEBTORS`.
 2. Making a report on all your bank accounts (but not your caisses)

These are classically `BALANCE` accounts, but you cannot separate them at a finer level without parsing account numbers.

We could use the `classe` column.   I think it used to simply be the larger account types (`Assets`, `Income`, `Expense`, `Debtor/Creditor`, etc...).  That would need to change.

Basically, this would be a middle ground between account tags (proposed in #209) and account types (which are too large for the scenarios proposed above).

jniles:
As we move more and more towards trying to make BHIMA easier to use, this seems more and more necessary.  It could be done in conjunction with https://github.com/IMA-WorldHealth/bhima-2.X/issues/2225.

[...]

PR #2386, opened 2017-12-11
Payroll all rubrics

lomamech:
refactor(Payroll) 
-Remove all old payroll table for bhima 1.x (Tax, Rubric, Cotisation)
- Implement API and client side for Payroll Rubric
- Implement Integration test for Payroll Rubrics
- Implement end to end test for Payroll Rubrics

reference #2292


client/src/i18n/en/form.json
```diff
@@ -314,6 +320,7 @@
             "FINANCIAL_DETAIL": "Current balance, the posting journal is included:",
             "FINANCIAL_STATUS": "Financial Status",
             "FIRST_NAME": "First Name",
+            "FOUR_ACCOUNT": "Third parties accounts",
```
jniles: Instead of calling this "Four Account", can we say "debtor_account" or something like that?   That way it isn't hard-coded OHADA.  "Four account" doesn't mean anything in English.

client/src/i18n/en/form.json
```diff
@@ -454,6 +464,7 @@
             "SEX": "Sex",
             "SHOW_SIMPLE_PREVIEW": "Show Simple Preview",
             "SIGNATURE": "Signature",
+            "SIX_ACCOUNT": "Costs Accounts",
```
jniles: Instead of "Six Account", can we call this "Expense Account"?  That way, we will be able to explain it to people who are not familiar with OHADA.  Six accounts are just expense accounts.

server/controllers/finance/accounts/index.js
```diff
@@ -117,6 +118,8 @@ function remove(req, res, next) {
  * GET /accounts
  */
 function list(req, res, next) {
+  const filters = new FilterParser(req.query, { tableAlias : 'a' });
```
jniles, line 129: Nice :+1:  Contributes to #2240.

client/src/js/components/bhAccountSelect.js
```diff
@@ -36,6 +37,9 @@ function AccountSelectController(Accounts, AppCache, $timeout, bhConstants, $sco
     // cache the title account ID for convenience
     $ctrl.TITLE_ACCOUNT_ID = bhConstants.accounts.TITLE;
 
+    // To filter accounts by class of accounts
+    $ctrl.classe = $ctrl.classe || '';
```
jniles, line 41: I really think this is a duplication of `account_type_id`.  Let's hold off on this dependency until we are certain we want to use `classe` instead of `account_type`.  After all, users can view and modify the `type_id` on the account page, but they cannot modify the `classe`.

client/src/modules/payroll/rubrics/rubrics.js
```diff
@@ -0,0 +1,110 @@
...
+  function toggleFilter() {
+    vm.filterEnabled = !vm.filterEnabled;
+    vm.gridOptions.enableFiltering = vm.filterEnabled;
+    vm.gridApi.core.notifyDataChange(uiGridConstants.dataChange.COLUMN);
+  }
+
+  function loadRubrics() {
+    vm.loading = true;
+
+    Rubrics.read(null, { detailed : 1 })
```
jniles: Does rubrics have a detailed option?

client/src/modules/payroll/rubrics/rubrics.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  // update an existing Rubric
+  function editRubric(title) {
+    $state.go('rubrics.edit', { id : title.id });
+  }
+
+  // create a new Rubric
+  function createRubric() {
+    $state.go('rubrics.create');
```
jniles, line 107: These state links should just be directives using `ui-sref`.  It will be easier to migrate to future versions of angular that way, I think.

client/src/modules/payroll/rubrics/templates/ipr.tmpl.html
```diff
@@ -0,0 +1,5 @@
+<div class="ui-grid-cell-contents text-left">
+  <div ng-if="row.entity.is_ipr">
+  	<i class="fa fa-check-square"></i>
```
jniles: In my experience at HEV, users don't really know what the icons are.  Plus they have resizing issues in the grid (many users with poor eyesight increase the screen size and the icons then look very distorted).  It would be better to use "oui/non" or "true/false" for boolean values, I think.

test/end-to-end/rubrics/rubrics.spec.js
```diff
@@ -0,0 +1,48 @@
...
+  before(() => helpers.navigate('#!/payroll/rubrics'));
+
+  const Page = new RubricPage();
+
+  const rubric = {
+    label : 'Rubric Syndical',
+    abbr  : 'CoSynd',
+    is_employee : 1,
+    is_percent : 1,
+    third_party_account_id : '40111002',
```
jniles: Could you include a comment with the account name here?  Like:
```
 third_party_account_id : '40111002', // SNEL
```
or something?  That will help future developers know what account's balance is changing.

test/integration/rubrics.js
```diff
@@ -0,0 +1,90 @@
...
+  it('GET /RUBRICS/:ID should not be found for unknown id', function () {
+    return agent.get('/rubrics/unknownRubric')
+    .then(function (res) {
+      helpers.api.errored(res, 404);
+    })
+    .catch(helpers.handler);
+  });
+
+  it('PUT /RUBRICS  should update an existing Rubric ', function () {
+    return agent.put('/rubrics/' + cotisation.id)
```
jniles: Could you make these tests respect eslint?

server/models/schema.sql
```diff
@@ -1687,43 +1542,6 @@ CREATE TABLE `supplier` (
   FOREIGN KEY (`creditor_uuid`) REFERENCES `creditor` (`uuid`) ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
-
-DROP TABLE IF EXISTS `tax`;
```
jniles, line 1693: Wee!!

client/src/modules/payroll/rubrics/modals/rubric.modal.html
```diff
@@ -0,0 +1,172 @@
...
+      classe="4"
+      name="third_party_account_id"
+      on-select-callback="RubricModalCtrl.selectAccountFour(account)"
+      exclude-title-accounts="true"
+      required="true"
+      validation-trigger="RubricForm.$submitted">
+    </bh-account-select>
+
+    <bh-account-select
+      id="costs_account_id"
```
jniles: A better name is "Expense Account".  Tax is an expense, for example.  The account type it will use is an expense account as well.

client/src/modules/payroll/rubrics/modals/rubric.modal.js
```diff
@@ -0,0 +1,77 @@
...
+    vm.rubric.third_party_account_id = account.id;
+  };
+
+  vm.selectAccountSix = function selectAccountSix(account) {
+    vm.rubric.costs_account_id = account.id;
+  };
+
+  vm.setDiscount = function setDiscount(value) {
+    vm.setting = true;
+    vm.rubric.is_discount = value === 1 ? true : false;
```
jniles: Can you use `ng-true-value` and `ng-false-value` to do this automatically in the view?

jniles (commented):
@lomamech this represents a large amount of work.   I've tested this and the user experience of creating a rubric is pretty good!  You might consider using `inline` checkboxes and radio buttons instead of the vertical ones to save space.

I'm not certain we should go towards using account `classe` yet.  We've had it for historical reasons, but it really is better to use account types.  We have pretty fine-grained account types now that roughly correspond with the `classe` in OHADA (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1607).

Filtering by account type is a good idea!  We should implement it (https://github.com/IMA-WorldHealth/bhima-2.X/issues/2364).

In general, we probably shouldn't use icons as the only piece of text in a grid column.  They look terrible when printed out, and they don't convey a lot of information to users who cannot recognize what they are.  For true/false values, let's just say that.

I'm not very experienced with the rules for apply taxes and payrolls, so I recommend you run this by @mbayopanda or @DedrickEnc who previously worked on the payroll forms in 1.X to see what they think.

Overall, this is a heap of work and pretty good code!  Nice!  :+1:

jniles:
Also, this is something weird - I made this tax.  Is it valid?  If not, can we add rules to the form to prevent it?

weirdtax

lomamech:
@jniles

jniles (approved):
LGTM.  A lot of work went into this!

bors r+

PR #2388, opened 2017-12-11
Complete account statement module

### LINKED ISSUE #1762: refactor: Account Statement Feedback
We've analyzed the Account Statement as a team, and came up with the following suggestions:
 - [x] Remove the word "totals" from the grid footer.
 - [x] Order the records by date by default
 - [x] By default, no need to show the account label/number
 - [x] Need variance/difference between credits and debits in the footer
 - [x] Need to the beginning balance in the top of the grid.
 - [x] Put `type: number` on grid columns definition to enable proper sorting.
 - [x] Implement SaveState

jniles:
This PR completes all the work for the Account Statement module set out in the #1762.

 1. SaveState has been implemented on the grid.
 2. A grid footer shows:
  a. The number of records in the grid.
  b. The opening balance before the first record shown in the grid.
  c. The debits and credits summed.
  d. The difference between debits and credits.

Closes #1762.


server/controllers/finance/accounts/index.js
```diff
@@ -226,6 +234,83 @@ function getBalance(req, res, next) {
     .done();
 }
 
+/**
```
DedrickEnc, line 238: This logic should be in the fiscal year controller file.

server/config/routes.js
```diff
@@ -154,6 +154,7 @@ exports.configure = function configure(app) {
   app.get('/accounts', accounts.list);
   app.get('/accounts/:id', accounts.detail);
   app.get('/accounts/:id/balance', accounts.getBalance);
+  app.get('/accounts/:id/openingBalance', accounts.getOpeningBalanceForPeriod);
```
DedrickEnc, line 157: Nice!

client/src/modules/accounts/accounts.service.js
```diff
@@ -35,9 +50,7 @@ function AccountService(Api, $http, util, bhConstants) {
    *   an array of JSONs.
    */
   function read(id, options) {
-    var url = baseUrl.concat(id || '');
-    return $http.get(url, { params : options })
-      .then(util.unwrapHttpResponse)
+    return Api.read.call(this, id, options)
```
DedrickEnc, line 53: Hmmm ... do you really need this? LOL

client/src/modules/account_statement/account_statement.html
```diff
@@ -78,7 +94,8 @@
           ui-grid-resize-columns
           ui-grid-move-columns
```
DedrickEnc, line 95: Do we need `ui-grid-move-columns`?

DedrickEnc (commented):
@jniles

This is not related to your pull request but I can not see why we should continue to keep `extra.js` and `accounts.js` separated.

I have made some comment also

client/src/modules/accounts/accounts.service.js
jniles: Unfortunately, I do, since I need to do post-processing of the data.  This lets us call `.then()` after the `read()` call.

There might be a prettier way of doing this ... but I'm not sure what it is currently.  Happy to have suggestions if you know it!

server/controllers/finance/accounts/index.js
jniles: Will do! 👍

client/src/modules/account_statement/account_statement.html
jniles: I think it's helpful if a user wants to customize their view.  In the long term, we'll make our PDFs print out in the same order as the ui-grid, and that will give users a lot of power to customize their application.

jniles:
@DedrickEnc thanks for the review!  I've made the suggested changes.

The column movements helps a lot in auditing, comparing the paper to the computer.  I always drag the "record" column to the far left so that I can go line by line down a paper registry and make sure the users have properly recorded all the information.  Since I haven't noticed a significant performance impact from it, I propose we keep it, at least for now.

We need to address the accounts `extra.js` file sooner or later.  Next week I'll be working on Comptes d'Exploitation and will tackle it then.  I'm hesitant to put it in accounts/index.js since it will make a ~550 line file, which is a lot to work with.

This is ready for a second review!

jniles:
Try again.  Seems like a flake.

bors r+

jniles:
Fixed an test that breaks #2418.  Basically, if you update protractor, [this change is applied](https://github.com/angular/protractor/commit/a62efc6e401bc1aa7408e3008ccdaa219b528636#diff-89f02ac1eed840a7f24ba7f0c2411178R687) and protractor rejects numbers in the `.findByCssContainingText()` function.  We use a number in the [account_statement.spec.js test](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/account_statement/account_statement.spec.js#L17), which breaks.   I've switched it to a string.

PR #2393, opened 2017-12-12
chore(stock movements registry): filter movement by user

mbayopanda:
This PR add a filter for users, so that we can perform also search stock movements by users.

### DIFF
--- a/client/src/modules/stock/movements/modals/search.modal.js
+++ b/client/src/modules/stock/movements/modals/search.modal.js
@@ -8,10 +8,13 @@ SearchMovementsModalController.$inject = [
 
 function SearchMovementsModalController(data, Notify, Instance, Flux, $translate, Periods, Store, util, Stock) {
   var vm = this;
+  var lastViewFilters;
+  var lastDisplayValues;
+  var displayValues = {};
   var changes = new Store({ identifier : 'key' });
 
   var searchQueryOptions = [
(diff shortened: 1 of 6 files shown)


client/src/modules/stock/movements/modals/search.modal.js
```diff
@@ -8,46 +8,52 @@ SearchMovementsModalController.$inject = [
...
       changes.post(filterChange);
     });
   };
 
-  var lastViewFilters = Stock.filter.movement.formatView().customFilters;
+  lastViewFilters = Stock.filter.movement.formatView().customFilters;
 
   // map key to last display value for lookup in loggedChange
-  var lastDisplayValues = lastViewFilters.reduce(function (object, filter) {
+  lastDisplayValues = lastViewFilters.reduce(handleLastViewFilter, {});
```
jniles: Most other registries put this on the service - see [the Patients search modal](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/invoices/registry/search.modal.js#L24) for example.  The method is `getDisplayValueMap()` on the FilterService.

jniles (commented):
Code looks fine to me.  Could you include an integration test to make sure the API always works?

jniles:
I force-pushed a new test!  And I also used the `getDisplayValueMap()`.  This is ready for merge.

bors r+

PR #2398, opened 2017-12-13
fix(inventory) : refresh correctly the grid after an update

### LINKED ISSUE #2377: (Inventory) Inventory list doesn't refresh the grid after adding  a new inventory when the option `add another` is activated
When we add a new inventory with the option `add another` activated, the grid which contains list of inventory is not updated with the new inventory that we have added.

mbayopanda:
This PR fix the refresh issue on the inventory grid after the creation of a new inventory (with the option add other activated)

Close #2377

### DIFF
--- a/client/src/modules/inventory/list/list.js
+++ b/client/src/modules/inventory/list/list.js
@@ -5,7 +5,7 @@ angular.module('bhima.controllers')
 InventoryListController.$inject = [
   'InventoryService', 'NotifyService', 'uiGridConstants', 'ModalService',
   '$state', 'FilterService', 'appcache', 'GridColumnService', 'GridStateService',
-  'GridExportService', 'LanguageService', 'SessionService',
+  'GridExportService', 'LanguageService', 'SessionService', '$rootScope',
 ];
 
 /**
@@ -14,11 +14,9 @@ InventoryListController.$inject = [
  */
 function InventoryListController(
   Inventory, Notify, uiGridConstants, Modal, $state, Filters, AppCache, Columns, GridState,
-  GridExport, Languages, Session
+  GridExport, Languages, Session, $rootScope
 ) {
   var vm = this;
-  vm.download = Inventory.download;
-
   var cacheKey = 'InventoryGrid';
   var cache = new AppCache(cacheKey);
 
@@ -28,6 +26,7 @@ function InventoryListController(
   var state;
 
   // global variables
+  vm.download = Inventory.download;
   vm.lang = Languages.key;
   vm.filterEnabled = false;
   vm.gridOptions = {};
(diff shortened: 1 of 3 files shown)


client/src/modules/inventory/list/list.js
```diff
@@ -183,7 +196,7 @@ function InventoryListController(
 
   function startup() {
     // if parameters are passed through the $state object, use them.
-    if ($state.params.filters.length) {
+    if ($state.params.filters && $state.params.filters.length) {
```
jniles, line 200: Hmm ...  could we make this default to an empty array?  That would be in `inventory.routes.js`.  That way we would be sure this is always defined.

```js
{
  params :  {
    filters : [],
  }
}
```

client/src/modules/inventory/list/modals/actions.modal.js
```diff
@@ -39,37 +45,42 @@ function InventoryListActionsModalController(Account, Inventory, Notify, Instanc
...
+      Notify.success(message);
 
-        // pass back the uuid
-        Instance.close(res.uuid);
-      })
-      .catch(Notify.handleError);
+      // if we are supposed to create another item, just refresh the state
+      if (vm.createAnotherItem) {
+        vm.item = {};
+        form.$submitted = false;
```
jniles: It's probably best to use AngularJS's FormController functions.  Like [$setPristine()](https://docs.angularjs.org/api/ng/type/form.FormController#$setPristine).  That way, it cleans up all the previous submission data (`$touched`, etc).

client/src/modules/inventory/list/modals/actions.modal.js
```diff
@@ -39,37 +45,42 @@ function InventoryListActionsModalController(Account, Inventory, Notify, Instanc
...
-        Notify.success(message);
+    function handleAction(res) {
+      message = vm.isCreateState ? 'FORM.INFO.CREATE_SUCCESS' : 'FORM.INFO.UPDATE_SUCCESS';
 
-        // if we are supposed to create another item, just refresh the state
-        if (vm.createAnotherItem) {
-          vm.item = {};
-          return;
-        }
+      $rootScope.$broadcast('INVENTORY_UPDATED');
```
jniles, line 70: This is an interesting idea ... I wonder if we can make this more general with services, like the Accounts controller.

I'll think about it and propose a design in an issue!

client/src/modules/inventory/list/modals/actions.modal.js
```diff
@@ -81,29 +92,29 @@ function InventoryListActionsModalController(Account, Inventory, Notify, Instanc
...
-      .then(function (types) {
+      .then(function handleType(types) {
         vm.types = types;
       })
       .catch(Notify.handleError);
 
     // Inventory Unit
     Inventory.Units.read()
-      .then(function (units) {
+      .then(function handleUnit(units) {
```
jniles, line 109: You seem to like the prefix `handle` 😄

jniles (approved):
Code looks fine to me.  There are a few conflicts, feel free to merge anytime.

jniles:
@mbayopanda whenever you feel it is appropriate, you can "bors r+" this PR.

PR #2399, opened 2017-12-13
refactor(service): Translating the transaction type key inside the se…

### LINKED ISSUE #2396: Translate Voucher Types in Voucher Type Service
We do this in a couple places:
```js
    Vouchers.transactionType()                                                                                                                                                                                     
    .then(function (list) {                                                                                                                                                                                      
      vm.types = list.data.map(function (item) {                                                                                                                                                                 
          item.hrText = $translate.instant(item.text);                                                                                                                                                             
          return item;                                                                                                                                                                                             
        });                                                                                                                                                                                                        
      })                                                                                                                                                                                                           
      .catch(Notify.handleError);
```

We should move this logic into the VoucherService, so that we only need to do this once and do not repeat ourselves.

DedrickEnc:
This commit moves the transaction type translation logic from controllers to `voucher.service` 
closes #2396

### DIFF
--- a/client/src/modules/vouchers/complex-voucher.ctrl.js
+++ b/client/src/modules/vouchers/complex-voucher.ctrl.js
@@ -2,7 +2,7 @@ angular.module('bhima.controllers')
   .controller('ComplexJournalVoucherController', ComplexJournalVoucherController);
 
 ComplexJournalVoucherController.$inject = [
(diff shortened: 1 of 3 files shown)


jniles (approved):
Looks great to me!  Awesome!

bors r+

PR #2404, opened 2017-12-14
chore: migration from v0.5.0 to v0.6.0

### LINKED ISSUE #2400: BHIMA is broken on MySQL 5.7.5+
MySQL 5.7.5+ has begun solving early [implementation errors](https://mysqlserverteam.com/mysql-5-7-only_full_group_by-improved-recognizing-functional-dependencies-enabled-by-default/) that affect BHIMA.  In particular, invoicing breaks due to `ONLY_FULL_GROUP_BY` not being respected.

You can debug this locally on earlier MySQL installations by setting `SET @@GLOBAL.sql_mode="ONLY_FULL_GROUP_BY"`.

This problem makes BHIMA incompatible with future versions of MySQL.  The solution appears to be to redesign many of our aggregate queries to be compliant with the pitfalls of group by.

**For Example**
db_errors
_Fig 1: Group By Errors on the Invoice Module_

jniles:
This PR adds a migration script to upgrade the v0.5.0 BHIMA installations to v0.6.0.  It also makes sure BHIMA is forward compatible with future MySQL versions by fixing `ONLY_FULL_GROUP_BY` errors.

In details:
 1. The inventory list controller used ES2015+.
 2. The interceptors file did not log the full error from a database SQL
 error.
 3. The cashboxes server API hard-coded the name of the test database
 `bhima_test`.  This prevented using any other database.
 4. Added extra migration scripts to update units.
 5. Fixed four instances of Group By rules breaking.  All integration
 tests now pass.
 6. The `migrate.sql` script was added.  It is kind of lazy at this point.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/2400.


jniles:
It looks like this has worked all today without any hiccups.  Bumping for release.

PR #2409, opened 2017-12-15
feat(build): Use to cross-env module to stop deal with the operation …

### LINKED ISSUE #2394: Replace `npm run dev` and `npm run dev_windows` with `cross-env`
There's an awesome module for teams that work cross platform called [cross-env](https://www.npmjs.com/package/cross-env).  We should try to use it on both Windows and Linux environments to see if it could replace our need to have multiple scripts in our package.json.  It's getting kind of messy!

DedrickEnc:
This commit use `cross-env` to make the operating system difference transparent to the developer. The use of `cross-env` leads to drop two commands with which we got used to use such as `npm run dev_windows` and `npm run app_windows`. 
Now, no matter you are in linux or windows, use : 
* `npm run dev` to build the application in a development mode 
* `npm run app` to build the application in a production mode

Run : ```npm install cross-env --save-dev```

closes #2394


DedrickEnc:
I have just finished to test in linux, it works!

package.json
```diff
@@ -4,8 +4,8 @@
   "description": "A rural hospital information management system.",
   "main": "npm run build",
   "scripts": {
-    "app": "NODE_ENV=production ./node_modules/.bin/gulp build && cd bin && NODE_ENV=production node server/app.js",
-    "dev": "NODE_ENV=development ./node_modules/.bin/gulp build && cd bin && NODE_ENV=development node server/app.js",
+    "app": "cross-env gulp build && cd bin && set NODE_ENV=production && node server/app.js",
```
jniles, line 7: Could you put `NODE_ENV=production` in front of the gulp build here?  That way it will minify the code using [this line in the gulpfile](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/gulpfile.js#L38) which activates the minification [here](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/gulpfile.js#L174).

jniles (commented):
Looks awesome!

Could we put `cross-env` in the dependencies and not just in the `devDependencies`?  I think it would be useful in production too!

PR #2418, opened 2017-12-17
Use to cross-env module to normalize package scripts

### LINKED ISSUE #2394: Replace `npm run dev` and `npm run dev_windows` with `cross-env`

jniles:
These commits use `cross-env` to make the operating system difference transparent to the developer. The use of `cross-env` leads to drop two commands with which we got used to use such as `npm run dev_windows` and `npm run app_windows`. 
Now, no matter you are in linux or windows, use : 
* `npm run dev` to build the application in a development mode 
* `npm run app` to build the application in a production mode

Run : `npm install` or `yarn install` to get update your dependencies, then use `npm run ${script}` to run your chosen build.

Closes #2394.

Previous PR: #2409.


jniles:
Yay!

bors r+

PR #2420, opened 2017-12-18
chore(reports): add account filter to shared.js

jniles:
This commit adds the "account_id" filter to the shared filter library in the reports to allow accounts to be rendered on the heads of PDF reports.  Note that this still has account ids instead of count numbers.


jniles:
Well ... that was cool.

PR #2422, opened 2017-12-18
fix(reports): order Income/Expense by account

jniles:
This commit makes sure the Income and Expense report is ordered by account number.  This makes it easy to confirm that all the expected accounts are accounted for and that their balances are correct.


jniles:
Curious.

PR #2426, opened 2017-12-19
feat(login form) add loading indicator

### LINKED ISSUE #2406: Do not flash install state while waiting for login.
On slower connections, the "install" button appears before the page changes to the login screen.  This is a bit confusing.

There may not be a way to prevent the flash, but we can at least show a "loading" sign instead of the "install" button.  That way, a user will think they are waiting for the page load rather than thinking the application is broken or that they are seeing things.

jeremielodi:
This PR add a loading indicator in the login page, it solves many problems detected at #2406 
closes #2406

### DIFF
--- a/client/src/modules/login/login.html
+++ b/client/src/modules/login/login.html
@@ -3,15 +3,26 @@
     <div class="row">
       <div class="col-md-4 col-md-offset-4 col-xs-12">
 
-        <div ng-if="!LoginCtrl.installationExist" class="text-center" style="margin-top : calc(50vh - 190px); max-width: 400px; margin-right: auto; margin-left: auto;">
+        <div 
+          ng-if="!LoginCtrl.installationExist && LoginCtrl.finishInstallationChecking && !LoginCtrl.installationCheckingError" 
+          class="text-center" style="margin-top : calc(50vh - 190px); max-width: 400px; margin-right: auto; margin-left: auto;">
           <p>
             <span translate>INSTALL.NO_INSTALLATION</span><br>
             <a class="btn btn-primary" ui-sref="install" translate>FORM.BUTTONS.INSTALL</a>
           </p>
         </div>
-
+        <!--loading indicator-->
+        <center>
+          <span 
+            class="fa fa-circle-o-notch fa-spin fa-4x"
+            ng-show="!LoginCtrl.finishInstallationChecking"
+            style="margin-top : calc(50vh - 190px); max-width: 400px; margin-right: auto; margin-left: auto">
+          </span> 
+        </center>
+        <!-- /loading indicator-->
+                
          <!--@fixme - style hacks for content centering-->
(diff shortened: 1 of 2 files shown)


client/src/modules/login/login.html
```diff
@@ -3,15 +3,26 @@
     <div class="row">
       <div class="col-md-4 col-md-offset-4 col-xs-12">
 
-        <div ng-if="!LoginCtrl.installationExist" class="text-center" style="margin-top : calc(50vh - 190px); max-width: 400px; margin-right: auto; margin-left: auto;">
+        <div 
+          ng-if="!LoginCtrl.installationExist && LoginCtrl.finishInstallationChecking && !LoginCtrl.installationCheckingError" 
+          lass="text-center" style="margin-top : calc(50vh - 190px); max-width: 400px; margin-right: auto; margin-left: auto;">
```
jniles: It looks like this has broken the "class" attribute.  It now reads `lass="..."`.

client/src/modules/login/login.html
```diff
@@ -3,15 +3,26 @@
     <div class="row">
       <div class="col-md-4 col-md-offset-4 col-xs-12">
 
-        <div ng-if="!LoginCtrl.installationExist" class="text-center" style="margin-top : calc(50vh - 190px); max-width: 400px; margin-right: auto; margin-left: auto;">
+        <div 
+          ng-if="!LoginCtrl.installationExist && LoginCtrl.finishInstallationChecking && !LoginCtrl.installationCheckingError"
```
jniles, line 7: Do you need the `LoginCtrl.installationCheckingError`?  It looks like this code would work fine without it.

client/src/modules/login/login.js
```diff
@@ -25,7 +25,16 @@ function LoginController(AppCache, Session, Languages, Projects, Notify, Install
   // contains the values from the login form
   vm.credentials = {};
   vm.login = login;
-  vm.languageService = Languages;
+  vm.languageService = Languages;  
+
+  // vm.finishInstallationChecking help to :
```
jniles, line 30: Nice comments!

client/src/modules/login/login.js
```diff
@@ -34,7 +43,13 @@ function LoginController(AppCache, Session, Languages, Projects, Notify, Install
   // check basic installation information exist
   Install.checkBasicInstallExist()
     .then(handleCheckInstallExist)
-    .catch(Notify.handleError);
+    .catch(function(err){
```
jniles: This function should have spaces around the parens:
```js
function (err) {
```

client/src/modules/login/login.html
jeremielodi: When I try to stop redis and launch bhima, I have the installation button displayed.
The checking operation in this case can't succeed because redis is down.
@jniles , try to create an error (network,...) and see what will happen

jniles: Aha.  Interesting.  Okay 👍

jniles:
LGTM.

bors r+

PR #2437, opened 2018-01-06
deps: bump versions

jniles:
This commit bumps the bower versions of modules and updates yarn's
lockfile.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,10 +15,10 @@
     "src"
   ],
   "dependencies": {
-    "angular": "1.6.8",
-    "angular-animate": "1.6.8",
-    "angular-sanitize": "1.6.8",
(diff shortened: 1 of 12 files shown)


jniles:
Hmm... seemed to move the problem to the clientside unit tests.

EDIT: I think this is caused by the bump to Karma 2.0.  I'm investigating.

jniles:
So, the problem was Angular 1.6.7 now rejects the GET for our i18n.  This was reflected in the [CHANGELOG](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#167-imperial-backstroke-2017-11-24).  Clearly our karma proxy isn't working as expected.

I've rolled back the angular version for now, until we figure out how to make karma behave.

jniles:
Yay!

PR #2439, opened 2018-01-09
feat(date formatting) Make all dates render in DD/MM/YYYY

### LINKED ISSUE #2413: Make all dates render in DD/MM/YYYY
A few of our registries you a long-form date.  These dates aren't recognized by Excel and so we have to show two different versions of dates between the client and server.  This is not ideal.  For example:
daterendering

We should go through and make sure all dates are rendered in the French locality DD/MM/YYYY.  This will make it easier for users to recognize dates and compare them.

jeremielodi:
This PR helps to render all registry's date column in DD/MM/YYYY
closes #2413

### DIFF
--- a/client/src/js/services/util.js
+++ b/client/src/js/services/util.js
@@ -1,7 +1,7 @@
 angular.module('bhima.services')
-.service('util', UtilService);
+  .service('util', UtilService);
 
-UtilService.$inject = ['moment' ];
+UtilService.$inject = ['moment'];
 
 /**
  * @class util
@@ -17,6 +17,17 @@ function UtilService(moment) {
   service.unwrapHttpResponse = function unwrapHttpResponse(response) {
     return response.data;
   };
+
+  service.formatDate = function (date, format) {
+    var f = format || 'DD/MM/YYYY HH:mm:ss';
+    if (date) {
+      return moment(date).format(f);
+    } else {
+      return null;
+    }
+  }
+
+
 
   /** @todo comments showing usage */
   service.filterFormElements = function filterFormElements(formDefinition, requireDirty) {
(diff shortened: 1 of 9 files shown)


client/src/js/services/util.js
```diff
@@ -13,11 +13,40 @@ UtilService.$inject = ['moment' ];
...
+  service.formatDate = function(date, format) {
+    var f = format || 'DD/MM/YYYY HH:mm:ss';
+    if(date){
+      return moment(date).format(f);
+    }else{
+      return null;
+    }
+  }
+
+  service.FormatArraysDates = function (rows, dateFormat) {
```
jniles: It doesn't look like you use this ... do you need it?  Or can it be removed?

client/src/js/services/util.js
```diff
@@ -13,11 +13,40 @@ UtilService.$inject = ['moment' ];
...
   var service = this;
-
+  
   service.unwrapHttpResponse = function unwrapHttpResponse(response) {
     return response.data;
   };
 
+  service.formatDate = function(date, format) {
+    var f = format || 'DD/MM/YYYY HH:mm:ss';
+    if(date){
```
jniles: Could you make sure this respects eslint?  There should be a space between the `if` and all brackets.

````js
if (date) {
// etc ...
```

jniles (commented):
Just a few comments.  Looks pretty good!

jniles (approved):
Code LGTM.

By the way, I think this could have been done with a single line of code on each column:
```js
cellFilter: 'date:DD/MM/YYYY HH:mm:ss'
```
This probably is more performant as well.  However, since you've done all the work to implement these functions, we'll use them until we hit a performance bottleneck.

Good work!

bors r+

PR #2441, opened 2018-01-09
chore: use "Management" suffix on names

### LINKED ISSUE #2435: Use "Management" suffix to denote management modules
When the application was originally conceived, we had a folder called "Admin" in the tree for all administrative/management modules.  These modules performed overarching functions, like add users, depots, accounts, fiscal years, employees, etc, to the system.

However, with the increasing complexity of the application, we have started to move some of the administrative modules into their respective categories.  In this way, we centralize concepts and roles - users who are dealing with inventory shouldn't need access to other administrative modules and so on.  However, suddenly the naming scheme matters more.

I propose that we append the suffix "Management" to all management modules.  Despite being a longer name, it should bring clarity.  After all "Finance > Account" and "Admin > User" doesn't mean much anyway.  However, saying "Finance > Account Management" is much clearer and makes the tree easier to navigate for the inexperienced user.

jniles:
This commit adds the "management" suffix to module names in English to ensure clarity when a user is selecting modules.

Closes #2435


jniles (commented):
bors r+

PR #2442, opened 2018-01-09
fix(account statement): remove posted toggle

### LINKED ISSUE #2429: Hide the "Show Posted Transactions" in the Account Statement search.
The Account Statement module by definition shows only posted transactions.  However, we have a checkbox on the module for including posted transactions ... and if checked, it breaks the page.  Many users got tripped up on this.

I believe this bug stems from using the same search modal on the Journal and Account Statement.  We should make sure that only the Journal has the "Show Posted Transactions" option.

jniles:
The posted toggle should not be shown on the account statement since all records are posted.  Having this option on the page can cause unexpected behavior in production, so it is removed.

Closes #2429.


jniles:

PR #2443, opened 2018-01-09
fix(purchase): ui glitches in purchase orders

### LINKED ISSUE #1842: (refactor) Purchase order functionality refactoring
The purchase order is on master now, so here are some noticed problems when using the module

1. The payment mode should be default on **LORS DE L'ACHAT** 

2.  Change ```FORM.INFO.INVALID_ITEMS``` into ```FORM.ERRORS.INVALID_ITEMS``` and translate it

3. Remove the button **RETROUVER UN ARCTICLE** or implement it

These points can be discussed to find the final task list

jniles:
This commit fixes the following issues:
 1. Payment defaults to "on purchase" instead of "on delivery"
 2. The button to fetch the cache has been removed.

Closes #1842.


jniles (commented):
This is just straight forward refactoring.

bors r+

PR #2446, opened 2018-01-12
feat(account) prevent the user to set 0 as account number

### LINKED ISSUE #2433: Creating an account with "0" account number breaks account management.
If a user makes an account with the account number "0", account management breaks.  Specifically, grouping breaks and collapsing at the root does not work anymore.  See below:

weirdaccounts
_Fig 1: All of these are parent accounts except 100_.

It also makes for a confusing "parent account" dropdown.  See below:
createaccountparentaccount
_Fig 1: Two "0" Accounts?_

There are two workarounds to this - either allow the parent account number to be NULL _or_ simply disallow "0" as an account number the user can select.  I'm in favor of the second option personally, since it requires the least work and allows the database to enforce its NOT NULL constraint.  But I am open to any suggestion.

You also cannot delete an account with account number "0".

jeremielodi:
This PR  oblige the user to enter an account number different to 0.
closes #2433

### DIFF
--- a/client/src/modules/accounts/edit/accounts.edit.js
+++ b/client/src/modules/accounts/edit/accounts.edit.js
@@ -51,8 +51,8 @@ function AccountEditController(
   // states that are available as sibling states - these can be used to show and hide
   // relevent components
   vm.states = {
-    create : 'accounts.create',
-    edit : 'accounts.edit',
+    create: 'accounts.create',
+    edit: 'accounts.edit',
   };
 
   vm.state = angular.copy(vm.stateCurrent.name);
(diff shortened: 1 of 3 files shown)


client/src/modules/accounts/edit/accounts.edit.js
```diff
@@ -196,6 +196,12 @@ function AccountEditController(
       return;
     }
 
+    var number = parseInt(vm.account.number);
```
jniles: The `parseInt()` function should always be called with a radix - in this case, it should be 10.  See the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Parameters) for more information.

PR #2451, opened 2018-01-12
Feature aggregate stock purchase

### LINKED ISSUE #2385: Quantity to order
Display problem with the Inventories in Stock module just because there should not be the quantity to order as long as the initial stock is not yet distributed
issue2

lomamech:
- Implementing the calculation of the agreement time between the order and
  the delivery, initially the value of the delay is nullm then takes the
  wait value expressed in months and subsequently this valus is deatained
  by the average value saved with the last delivery
- Implement automatic of Purchase Interval Average
- Update type for delay use decimal

close #2385


server/controllers/finance/purchases.js
```diff
@@ -9,12 +9,14 @@
  * @requires db
  * @requires NotFound
  * @requires BadRequest
+ * @requires moment
```
jniles, line 12: 👍

server/controllers/finance/purchases.js
```diff
@@ -180,6 +182,46 @@ function create(req, res, next) {
...
+
+      return db.exec(getInventory, [data.uuid]);
+    })
+    .then((rows) => {
+      const datePurchase = new Date(data.date);
+
+      const transaction = db.transaction();
+      rows.forEach((row) => {
+        /**
+          * Normally purchase interval is calculated by deducting the gap in months of
```
jniles, line 204: These are excellent comments!

For those who do not understand how this code works, it is essentially a running average.  However, it will not always be the average.  Here's some code you can run to test it out for yourself:
```js
// first make a list of random integers between 0 and 100
const numbers = [];
let i = 100;
while (i--) { numbers.push(Math.floor(Math.random() * 101)); }

// numbers is now a list o length 100 with random values between 0 and 100

// this is @lomamech's average function.  Take the previous value, add it to the next value and divide by two.
// we will use Array.prototype.reduce() to apply it to the whole list of random numbers and console log the result.
const avg = (prev, next) => (prev + next) / 2;
const runningAverage = numbers.reduce(avg);
console.log(runningAverage);

// Now let's compute a traditional average.  This is just SUM(numbers) / LENGTH(numbers)
const sum = (prev, next) => prev + next;
const traditionalAverage = numbers.reduce(sum) / numbers.length;
console.log(traditionalAverage);

// Notice that these values are not equivalent!
console.log(`${runningAverage} !== ${traditionalAverage}`);
```

These results can be quite different.  For example, on my machine, I ran a few experiments comparing the difference between the traditional average and the running average.  Here's what I did:
```js
function runner() {
  // copy the above code here and add:
  return traditionalAverage - runningAverage;
}

let experiments = 1000;
const avgs = [];
while (experiments--) {
  // record the absolute value of the difference here.
  avgs.push(Math.abs(runner()));
}

// calculate the maximum value of the difference:
const maxDifference = Math.max.apply(null, avgs);
const minDifference = Math.min.apply(null, avgs);

console.log('maximum difference:', maxDifference);
console.log('minimum difference:', minDifference);
```

On my machine, I got 39.84 as the largest difference between the running average and the traditional average.  That is almost 40% spread!

The flaw with the running average is that it doesn't take into account the number of values.  [Here is a StackOverflow formula that accounts for this](https://stackoverflow.com/questions/12636613/how-to-calculate-moving-average-without-keeping-the-count-and-data-total/23493727#23493727).

I'm worried that this algorithm, while really efficient, can get very inaccurate results.  I propose that we store the "num_purchases" as a variable as well so that we can use it to calculate the average.  We can then use the Stack Overflow formula to calculate a more accurate running average with the same efficiency.

server/controllers/finance/purchases.js
```diff
@@ -180,6 +182,46 @@ function create(req, res, next) {
...
+          * Normally purchase interval is calculated by deducting the gap in months of 
+          * all purchase orders for a product, this form is also very expensive in terms of resources, 
+          * so let's store the date of the last orders and store the results in months in the column Purchase interval, 
+          * and in the following we would calculate each time the average value between the value purchase interval 
+          * and the difference in months between last_purchase and the date of the current purchase Order
+        */
+        let purchaseInterval = row.purchase_interval;
+        if (row.last_purchase) {
+          let diff = moment(datePurchase).diff(moment(row.last_purchase));
+          let duration = moment.duration(diff, 'milliseconds');
```
jniles: Can you not use `moment.duration(diff, 'months')` here to skip the next step?

server/controllers/finance/purchases.js
jniles: Alternatively, we can keep a separate table called `purchase_total` that has the following values:

| inventory_uuid | total_duration| num_purchases |
|----------------|-------------|---------------|
|    ...            |        3     |      12         | 
|   ...          |    5         |      2         | 

This way, we won't have to update the inventory table on every purchase/stock entry, just the `purchase_total`, which might make more sense.  We could then calculate the average by simply doing:
```js
const newNumberOfPurchases = row.num_purchases + 1;
const currentDuration = 2; // whatever the delay is using the date math @lomamech already implemented.
const newTotalDuration = row.total_duration + currentDuration;
const avgDuration = newTotalDuration / newNumberOfPurchases;  // this is the traditional average
```

server/controllers/finance/purchases.js
```diff
@@ -346,7 +389,39 @@ function purchaseStatus(req, res, next) {
...
+
+      rows.forEach((row) => {
+        /**
+          * Normally the delay agreement time between the order and the delivery is calculated 
+          * by finding the average duration of agreement of orders of last six months for a product 
+          * this calculation is very expensive in terms of memory, which is the reason why we keep 
+          * this information in the inventory table for article shovel
+        */
+
+        let delay = row.delay ? ((status.delay + row.delay) / 2) : status.delay;
```
jniles: The same idea can be applied as above.  Just add a `total_delay` column to the `purchase_total` table.

jniles (commented):
@lomamech, this was a really slick PR.  I had to do a lot of tests to see how it worked and make sure it performs correctly!  Your implementation is clear and your comments are excellent.

From my tests, it looks like this method of averaging can become very wrong in certain scenarios.  We don't want to be very wrong for our reorder intervals.  Fortunately, it looks like we can make this work properly by dividing by the number of samples instead of dividing by two.

You have two options:
 1. Add the column `num_purchase` to the `inventory` table and update your formula to be: `New average = old average * (n-1)/n + new value /n`.  [This was proposed on [StackOverflow](https://stackoverflow.com/questions/12636613/how-to-calculate-moving-average-without-keeping-the-count-and-data-total/23493727#23493727).
 2. Implement the `purchase_total` table proposed in my second comment.  I prefer this option since it will limit the number of times we change the `inventory` table.

Let me know what you think!  And @mbayopanda if you have any feedback, feel free to comment it.

client/src/i18n/en/form.json
```diff
@@ -93,7 +93,7 @@
             "GENERIC": "The form contains errors.",
             "HAS_ERROR": "Contains an error",
             "HAS_ERRORS": "The form contains errors.  Please correct them and resubmit.",
-            "LOT_QUANTITY": "Lot Quantity",
+            "LOT_QUANTITY": "Batch number Quantity",
```
mbayopanda: 'Batch' means `Lot`, so you don't need to add 'number' because 'Batch number' means 'Numero de lot'.

The good translation of `Quantité du lot` is 'Batch Quantity' or 'Lot Quantity'

client/src/i18n/en/form.json
```diff
@@ -138,7 +138,7 @@
             "ITEMS_FULL": "There are no additional inventory items that be assigned to this invoice.",
             "LOADING": "Loading",
             "MANY_TRYS": "Many tries",
-            "MAX_LOT_REACHED": "Max lot reached",
+            "MAX_LOT_REACHED": "Max Batch Number reached",
```
mbayopanda: `Max Batch Quantity Reached` or `Max Lot Quantity Reached`

client/src/i18n/en/form.json
```diff
@@ -591,6 +591,7 @@
             "DOCUMENT_NAME": "Enter document name",
             "PATIENT_GROUP": "Enter patient group",
             "EMAIL": "Enter email",
+            "LOT":"Enter Batch Number",
```
mbayopanda, line 595: 👍

jniles (approved):
All changes have been addressed!

bors r+

PR #2452, opened 2018-01-12
feat(patient) add health zone and health area fields

### LINKED ISSUE #2370: Add Health Zone and Health Area to Patient Registration Form
We've been asked by our HEV clients to include health zone and health area as optional fields on the Patient Registration form.  This will help make sure they conform to PBF (financing based on performance) standards.  They should be two simple text fields to begin with.

jeremielodi:
This PR add more informations for a patient
closes #2370


server/controllers/medical/patients/index.js
```diff
@@ -225,7 +225,7 @@ function lookupPatient(patientUuid) {
   // @FIXME(sfount) ALL patient queries should use the same column selection and gaurantee the same information
   const sql = `
     SELECT BUID(p.uuid) as uuid, p.project_id, BUID(p.debtor_uuid) AS debtor_uuid, p.display_name, p.hospital_no,
-      p.sex, p.registration_date, p.email, p.phone, p.dob, p.dob_unknown_date,
+      p.sex, p.registration_date, p.email, p.phone, p.dob, p.dob_unknown_date, p.health_zone, p.health_area,
```
jniles: You should probably add these to the `find()` method as well so that they can be downloaded via CSV from the patient registry.

jniles (changes requested):
@jeremielodi, I'm not sure why the tests have been failing, could you look into them?

This puts the information in, but we need a way to view the information once it is in the DB.  Could you add two (hidden by default) columns to the Patient Registry so that the user can confirm that they correctly added the data?  This will enable our clients to easily see the health zone/health areas of patients by simply browsing the Patient Registry.

client/src/modules/patients/edit/edit.html
```diff
@@ -396,6 +396,26 @@ <h4><span translate>FORM.LABELS.CURRENT_LOCATION</span></h4>
                   </div>
                 </div>
 
+                <div class="form-group" ng-class="{'has-error' : patientDetails.employer.$invalid && PatientRegistrationForm.$submitted}">
```
jniles: What is `patientDetails.employer`? 

It should say `patientDetails.health_zone`.  The `has-error` class should match the `name` of the `<input>` that it wraps.

jniles (commented):
@jeremielodi , I'll fork this PR and see if I can get tests to pass.

client/src/modules/patients/edit/edit.html
jniles: This still needs to addressed, afterwards we can merge it.

jniles (approved):
LGTM.  bors r+

PR #2453, opened 2018-01-12
chore: replace translate filter with directive

jniles:
This commit replaces all the translate filters in the application with the translate directive.  I've also partially replaced the translate attributes with translate-attrs.


jniles (commented):
bors r+

PR #2457, opened 2018-01-18
feat(inventory) add broken and note columns

### LINKED ISSUE #2379: (stock) Handle correctly individual inventory state
The current state of stock in `bhima` cannot handle individual inventory data such state of the inventory (good/broken)

jeremielodi:
This PR handles correctly individual inventory state
closes  #2379

### DIFF
--- a/client/src/modules/inventory/list/list.js
+++ b/client/src/modules/inventory/list/list.js
@@ -97,7 +97,22 @@ function InventoryListController(
     cellClass : 'text-right',
     type : 'number',
     visible : false,
-  }, {
+  },
+  {
+    field : 'is_broken',
+    displayName : 'INVENTORY.IS_BROKEN',
+    headerCellFilter : 'translate',
+    cellTemplate : '/modules/inventory/list/templates/isbroken.cell.html',
+    visible : false,
+  },
+  {
+    field : 'note',
+    displayName : 'FORM.INFO.NOTE',
+    headerCellFilter : 'translate',
+    cellClass : 'text-right',
+    visible : false,
+  },
+  {
     field : 'action',
     displayName : '',
     cellTemplate : '/modules/inventory/list/templates/action.cell.html',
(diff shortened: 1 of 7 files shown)


client/src/modules/inventory/list/list.js
```diff
@@ -97,7 +97,22 @@ function InventoryListController(
     cellClass : 'text-right',
     type : 'number',
     visible : false,
-  }, {
+  },
+  {
+    field : 'is_brocken',
```
jniles: The proper way to spell this is `is_broken`.  Could you make this change?

server/controllers/inventory/inventory/core.js
```diff
@@ -143,6 +144,7 @@ function getItemsMetadata(params) {
   filters.equals('consumable');
   filters.equals('locked');
   filters.equals('label');
+  filters.equals('is_brocken');
```
jniles: You should include `filters.fullText('note')` here, so that filtering on the `note` column works.

server/models/schema.sql
```diff
@@ -764,6 +764,8 @@ CREATE TABLE `inventory` (
   `stock_min` INT(10) UNSIGNED NOT NULL DEFAULT 0,
   `type_id` TINYINT(3) UNSIGNED NOT NULL DEFAULT 0,
   `consumable` TINYINT(1) NOT NULL DEFAULT 0,
+  `is_brocken` TINYINT(1)  NULL DEFAULT 0,
```
jniles: This should be NOT NULL with a DEFAULT, to make sure that the default is always used.

client/src/modules/inventory/list/templates/isbrocken.cell.html
```diff
@@ -0,0 +1,12 @@
+<div class="ui-grid-cell-contents">
+  <span 
+    class="text-danger text-right"
+    ng-show='row.entity.is_brocken==1'
+    translate>
+    FORM.LABELS.YES
+  </span>
+  <span
```
jniles: You could save some space by moving the "text-right" class to the parent `<div>`.

jniles (changes requested):
@jeremielodi just a few comments, and then this looks LGTM.

server/controllers/inventory/inventory/core.js
```diff
@@ -143,6 +144,8 @@ function getItemsMetadata(params) {
   filters.equals('consumable');
   filters.equals('locked');
   filters.equals('label');
+  filters.equals('is_broken');
+  filters.equals('note');
```
jniles, line 148: In the future, this should use `fullText()` to do a `LIKE %value%` comparison.

jniles (approved):
LGTM.

bors r+

PR #2461, opened 2018-01-23
refractor(user permission) implement role for users

jeremielodi:
This PR extends the user permission management by creating roles and after affect permissions to them

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -89,7 +89,8 @@
             "SURE_CONTINUE": "Are you ABSOLUTELY sure?",
             "UNDERSTAND_ACTION": "I accept the consequences of this action",
             "WARNING": "Warning",
-            "CONFIRM_MODAL": "Confirmation Needed"
+            "CONFIRM_MODAL": "Confirmation Needed",
+            "DELETE_ROLE": "Do you really want to delete this role?"
         },
         "ERRORS": {
             "AT_LEAST_1_COL": "At least one column must be selected",
@@ -241,6 +242,7 @@
             "CASH_JOURNAL": "Cash Journal",
             "CASH_SPLITED": "Revenu depenses splited",
             "CASH_PAYMENTS": "Cash Payments",
+            "CHECK_ALL":"Check all",
             "CAUTION": "Caution",
             "CHARGE": "Charge",
             "CLIENT": "Client",
@@ -599,7 +601,10 @@
             "SEPTEMBER": "September",
             "OCTOBER": "October",
             "NOVEMBER": "November",
-            "DECEMBER": "December"
+            "DECEMBER": "December",
+            "PERMISSION_ASSIGNED_SUCCESS":"Permissions successfully affected",
+            "EDIT_ROLE":"Assign role(s)",
+            "ROLES_FOR":"Role(s) for"
         },
         "LEGENDS": {
             "CREATION": "Creation",
--- a/client/src/i18n/en/tree.json
+++ b/client/src/i18n/en/tree.json
@@ -95,6 +95,7 @@
     "USERS": "User Management",
     "VOUCHER_REGISTRY": "Voucher Registry",
     "AGED_DEBTORS": "Aged Debtors",
-    "WEEKEND_CONFIGURATION":"Weekend Configuration"
+    "WEEKEND_CONFIGURATION":"Weekend Configuration",
+    "ROLE_MANAGEMENT":"Role management"
   }
 }
--- a/client/src/modules/roles/create.html
+++ b/client/src/modules/roles/create.html
@@ -1,0 +1,32 @@
+<form name='roleForm' bh-submit="RolesAddCtrl.submit(roleForm)" novalidate>
+
+  <div class="modal-header">
+    <ol class="headercrumb">
+      <li>
+        <span translate>TREE.ROLE_MANAGEMENT</span>
+        <label class="label label-warning text-uppercase" translate>{{RolesAddCtrl.action}}</label>
+      </li>
+    </ol>
+  </div>
+
+  <div class="modal-body">
+    <div class="form-group" ng-class="{'has-error' : roleForm.label.$invalid && roleForm.$submitted}">
+      <label class="control-label" translate>FORM.LABELS.NAME</label>
+      <input type="text" class="form-control col-md-9" name="label" ng-model="RolesAddCtrl.role.label" required>
+      <div class="help-block" ng-messages="roleForm.label.$error" ng-show="roleForm.$submitted">
+        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      </div>
+    </div>
+  </div>
+
+  <div class="modal-footer">
+    <button data-method="cancel" type="button" class="btn btn-default" ng-click="RolesAddCtrl.close()">
+      <span translate>FORM.BUTTONS.CANCEL</span>
+    </button>
+
+    <bh-loading-button loading-state="roleForm.$loading">
+      <span translate>FORM.BUTTONS.SUBMIT</span>
+    </bh-loading-button>
+  </div>
+
+</form>
--- a/client/src/modules/roles/create.html
+++ b/client/src/modules/roles/create.html
@@ -10,11 +10,13 @@
   </div>
 
   <div class="modal-body">
-    <div class="form-group" ng-class="{'has-error' : roleForm.label.$invalid && roleForm.$submitted}">
-      <label class="control-label" translate>FORM.LABELS.NAME</label>
-      <input type="text" class="form-control col-md-9" name="label" ng-model="RolesAddCtrl.role.label" required>
-      <div class="help-block" ng-messages="roleForm.label.$error" ng-show="roleForm.$submitted">
-        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+    <div class="row">
+      <div class="form-group" ng-class="{'has-error' : roleForm.label.$invalid && roleForm.$submitted}">
+        <label class="control-label" translate>FORM.LABELS.NAME</label>
+        <input type="text" class="form-control col-md-9" name="label" ng-model="RolesAddCtrl.role.label" required>
+        <div class="help-block" ng-messages="roleForm.label.$error" ng-show="roleForm.$submitted">
+          <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+        </div>
       </div>
     </div>
   </div>
@@ -24,7 +26,7 @@
       <span translate>FORM.BUTTONS.CANCEL</span>
     </button>
 
-    <bh-loading-button loading-state="roleForm.$loading">
+    <bh-loading-button data-method="submit" loading-state="roleForm.$loading">
       <span translate>FORM.BUTTONS.SUBMIT</span>
     </bh-loading-button>
   </div>
--- a/client/src/modules/roles/create.js
+++ b/client/src/modules/roles/create.js
@@ -7,7 +7,7 @@ RolesAddController.$inject = [
 
 function RolesAddController(data, $state, RolesService, Notify, $uibModalInstance) {
 
-  var vm = this;
+  const vm = this;
   vm.close = close;
   vm.submit = submit;
   vm.role = angular.copy(data);
@@ -21,15 +21,15 @@ function RolesAddController(data, $state, RolesService, Notify, $uibModalInstanc
     if (form.$invalid) {
       return false;
     }
-    var operation = (!data.uuid)?RolesService.create(vm.role) : RolesService.update(data.uuid, vm.role);
+    const operation = (!data.uuid) ? RolesService.create(vm.role) : RolesService.update(data.uuid, vm.role);
 
-    return operation.then(function () {
+    return operation.then(() => {
       Notify.success('FORM.INFO.OPERATION_SUCCESS');
       $state.reload();
       vm.close();
       return operation;
     })
-    .catch(Notify.handleError);
+      .catch(Notify.handleError);
 
   }
 
--- a/client/src/modules/roles/create.js
+++ b/client/src/modules/roles/create.js
@@ -1,0 +1,42 @@
+angular.module('bhima.controllers')
+  .controller('RolesAddController', RolesAddController);
+
+RolesAddController.$inject = [
+  'data', '$state', 'RolesService', 'NotifyService', '$uibModalInstance',
+];
+
+function RolesAddController(data, $state, RolesService, Notify, $uibModalInstance) {
+
+  var vm = this;
+  vm.close = close;
+  vm.submit = submit;
+  vm.role = angular.copy(data);
+  vm.isCreate = (!vm.role.uuid);
+  vm.action = vm.isCreate ? 'FORM.LABELS.CREATE' : 'FORM.LABELS.UPDATE';
+
+
+  function submit(form) {
+
+    form.$setSubmitted();
+    if (form.$invalid) {
+      return false;
+    }
+    var operation = (!data.uuid)?RolesService.create(vm.role) : RolesService.update(data.uuid, vm.role);
+
+    return operation.then(function () {
+      Notify.success('FORM.INFO.OPERATION_SUCCESS');
+      $state.reload();
+      vm.close();
+      return operation;
+    })
+    .catch(Notify.handleError);
+
+  }
+
+
+  function close() {
+    // transition to the overall UI grid state - this modal will be cleaned up on state change
+    $uibModalInstance.close();
+  }
+
+}
--- a/client/src/modules/roles/modal/rolesPermissions.html
+++ b/client/src/modules/roles/modal/rolesPermissions.html
@@ -1,0 +1,54 @@
+<form
+name="permissionForm"
+bh-submit="RolesPermissionsCtrl.getAffected()"
+novalidate>
+<div class="modal-header">
+
+  <span translate>USERS.PERMISSIONS</span> :
+  <span class="text-primary"> {{RolesPermissionsCtrl.role.label}}</span>
+</div>
+<div class="modal-body" style="max-height: 70vh; overflow:auto;">
+
+  <div class='row'>
+    <div class="col-lg-12">
+
+      <div>
+        <label class="radio-inline">
+          <input type='checkbox' ng-change="RolesPermissionsCtrl.selectAll()" ng-model='RolesPermissionsCtrl.allPage' ng-true-value="1"
+            ng-false-value="0" />
+          <span translate>FORM.LABELS.CHECK_ALL</span>
+        </label>
+      </div>
+      <div ng-repeat="module in RolesPermissionsCtrl.tree" class="treeview">
+
+        <h4>
+          <label class="radio-inline">
+            <input type='checkbox' ng-model='module.affected' ng-true-value="1" ng-false-value="0" />
+            <b translate>{{module.key}}</b>
+          </label>
+        </h4>
+
+        <ul class="list-group">
+          <li ng-repeat="page in module.pages" class="list-group-item">
+            <label class="radio-inline">
+              <input type='checkbox' ng-model='page.affected' ng-true-value="1" ng-false-value="0" />
+              <span translate>{{page.key}}</span>
+            </label>
+          </li>
+        </ul>
+
+      </div>
+    </div>
+
+  </div>
+</div>
+<div class="modal-footer">
+
+  <div pull-right>
+    <button class='btn btn-default' ng-click='RolesPermissionsCtrl.close()' translate>FORM.BUTTONS.CANCEL</button>
+
+    <bh-loading-button loading-state="PatientRegistrationForm.$loading">
+      <sapn translate>FORM.BUTTONS.SAVE</sapn>
+    </bh-loading-button>
+  </div>
+</div>
--- a/client/src/modules/roles/modal/rolesPermissions.html
+++ b/client/src/modules/roles/modal/rolesPermissions.html
@@ -16,7 +16,7 @@ novalidate>
         <label class="radio-inline">
           <input type='checkbox' ng-change="RolesPermissionsCtrl.selectAll()" ng-model='RolesPermissionsCtrl.allPage' ng-true-value="1"
             ng-false-value="0" />
-          <span translate>FORM.LABELS.CHECK_ALL</span>
+          <span id="checkall" translate>FORM.LABELS.CHECK_ALL</span>
         </label>
       </div>
       <div ng-repeat="module in RolesPermissionsCtrl.tree" class="treeview">
--- a/client/src/modules/roles/modal/rolesPermissions.js
+++ b/client/src/modules/roles/modal/rolesPermissions.js
@@ -1,0 +1,65 @@
+angular.module('bhima.controllers')
+  .controller('RolesPermissionsController', RolesPermissionsController);
+
+RolesPermissionsController.$inject = [
+  'data', '$state', '$uibModalInstance', 'AppCache', 'RolesService', 'NotifyService',
+];
+
+function RolesPermissionsController(data, $state, $uibModalInstance, AppCache, RolesService, Notify) {
+  const vm = this;
+  vm.close = close;
+  vm.role = angular.copy(data);
+  vm.tree = [];
+  vm.selectAll = selectAll;
+  vm.allPage = 0;
+
+  // affeted pages(permission) to this role
+  vm.getAffected = getAffected;
+
+  // vm.role.uuid
+  RolesService.unit(vm.role.uuid)
+    .then(res => {
+      vm.tree = res.data;
+    });
+
+  // close modal
+  function close() {
+    $uibModalInstance.close();
+  }
+
+  function selectAll() {
+    vm.tree.forEach(_module => {
+      _module.affected = vm.allPage;
+      _module.pages.forEach(page => {
+        page.affected = vm.allPage;
+      });
+    });
+  }
+
+  function getAffected() {
+
+    const ids = [];
+    vm.tree.forEach(_module => {
+      if (_module.affected === 1) {
+        ids.push(_module.id);
+      }
+      _module.pages.forEach(page => {
+        if (page.affected === 1) {
+          ids.push(page.id);
+        }
+      });
+    });
+
+    const params = {
+      role_uuid : vm.role.uuid,
+      unit_ids : ids,
+    };
+
+    RolesService.affectPages(params)
+      .then(() => {
+        Notify.success('FORM.LABELS.PERMISSION_ASSIGNED_SUCCESS');
+        vm.close();
+      })
+      .catch(Notify.handleError);
+  }
+}
--- a/client/src/modules/roles/modal/userRole.html
+++ b/client/src/modules/roles/modal/userRole.html
@@ -25,7 +25,7 @@
   <div class=" pull-right">
   <button class='btn inverse' ng-click='UsersRolesCtrl.close()'>Fermer</button>
  
-  <button data-method="cancel" type="button" class="btn btn-primary" ng-click="UsersRolesCtrl.assignRolesToUser()">
+  <button data-method="submit" type="button" class="btn btn-primary" ng-click="UsersRolesCtrl.assignRolesToUser()">
       <span translate>FORM.BUTTONS.SAVE</span>
     </button>
   </div>
--- a/client/src/modules/roles/modal/userRole.html
+++ b/client/src/modules/roles/modal/userRole.html
@@ -1,0 +1,32 @@
+<div class="modal-header">
+  <h4><span translate>FORM.LABELS.ROLES_FOR</span>
+    : <span class="text-primary">
+      {{UsersRolesCtrl.user.display_name}}
+    </span>
+  </h4>
+</div>
+<div class="modal-body">
+
+  <div class="row">
+    <div class="col-lg-12">
+      <ul class="list-group">
+        <li  class="list-group-item" ng-repeat="role in UsersRolesCtrl.roles">
+          <label class="radio-inline">
+            <input type='checkbox' ng-model='role.affected' ng-true-value="1" ng-false-value="0" />
+            <span title="{{role.label}}">{{role.label}}</span>
+          </label>
+        </li>
+      </ul>
+    </div>
+  </div>
+</div>
+
+<div class="modal-footer">
+  <div class=" pull-right">
+  <button class='btn inverse' ng-click='UsersRolesCtrl.close()'>Fermer</button>
+ 
+  <button data-method="cancel" type="button" class="btn btn-primary" ng-click="UsersRolesCtrl.assignRolesToUser()">
+      <span translate>FORM.BUTTONS.SAVE</span>
+    </button>
+  </div>
+</div>
--- a/client/src/modules/roles/modal/userRole.js
+++ b/client/src/modules/roles/modal/userRole.js
@@ -1,0 +1,75 @@
+angular.module('bhima.controllers').controller('UsersRolesController', UsersRolesController);
+UsersRolesController.$inject = ['data', '$uibModal', '$rootScope',
+  '$uibModalInstance', 'RolesService', 'SessionService', 'NotifyService',
+];
+
+function UsersRolesController(data, $uibModal, $rootScope, $uibModalInstance, RolesService, session, Notify) {
+  const vm = this;
+  vm.close = close;
+  vm.user = angular.copy(data);
+  vm.loadRoles = loadRoles;
+  vm.assignRolesToUser = assignRolesToUser;
+  vm.roles = [];
+
+  // loa all roles
+  function loadRoles() {
+    RolesService.userRoles(vm.user.id, session.project.id)
+      .then(response => {
+        delete vm.gridOptions.data;
+        vm.roles = response.data;
+      })
+      .catch(Notify.handleError);
+  }
+
+  // assigned role to he user
+  function assignRolesToUser() {
+    const codes = vm.roles
+      .filter(role => role.affected === 1)
+      .map(role => role.uuid);
+
+    const param = {
+      user_id : vm.user.id,
+      role_uuids : codes,
+    };
+    RolesService.assignToUser(param)
+      .then(() => {
+        Notify.success('FORM.INFO.OPERATION_SUCCESS');
+        vm.close();
+      }).catch(Notify.handleError);
+  }
+
+  // ui-grid
+  const columns = [{
+    field : 'label',
+    displayName : 'Label',
+  },
+  {
+    field : '-',
+    width : 100,
+    displayName : 'Affecté',
+    enableFiltering : false,
+    cellTemplate : 'modules/roles/templates/userAssignedRole.cell.html',
+  },
+  ];
+  // ng-click="
+  vm.gridOptions = {
+    appScopeProvider : vm,
+    enableColumnMenus : false,
+    columnDefs : columns,
+    enableSorting : true,
+    data : [],
+    fastWatch : true,
+    flatEntityAccess : true,
+  };
+
+
+  vm.gridOptions.onRegisterApi = function intApi(gridApi) {
+    vm.gridApi = gridApi;
+  };
+
+  function close() {
+    $uibModalInstance.close();
+  }
+
+  vm.loadRoles();
+}
--- a/client/src/modules/roles/roles.ctrl.js
+++ b/client/src/modules/roles/roles.ctrl.js
@@ -1,0 +1,104 @@
+angular.module('bhima.controllers')
+  .controller('RolesController', RolesController);
+
+RolesController.$inject = [
+  '$uibModal', 'RolesService', 'SessionService', 'ModalService', 'NotifyService',
+];
+
+function RolesController($uibModal, RolesService, session, Modal, Notify) {
+  const vm = this;
+  vm.loadRoles = loadRoles;
+  vm.loading = false;
+
+  vm.add = function add(role) {
+    const _role = role || {
+      project_id : session.project.id,
+    };
+
+    $uibModal.open({
+      keyboard : false,
+      backdrop : 'static',
+      templateUrl : 'modules/roles/create.html',
+      controller : 'RolesAddController as RolesAddCtrl',
+      resolve : {
+        data : function dataProvider() {
+          return _role;
+        },
+      },
+    });
+  };
+
+  // pages to affect to this role
+  vm.pages = function pages(role) {
+    $uibModal.open({
+      keyboard : false,
+      backdrop : 'static',
+      templateUrl : 'modules/roles/modal/rolesPermissions.html',
+      controller : 'RolesPermissionsController as RolesPermissionsCtrl',
+      resolve : {
+        data : function dataProvider() {
+          return role;
+        },
+      },
+    });
+  };
+
+  vm.remove = function remove(uuid) {
+    const message = 'FORM.DIALOGS.DELETE_ROLE';
+    Modal.confirm(message)
+      .then(confirmResponse => {
+        if (!confirmResponse) {
+          return;
+        }
+        RolesService.delete(uuid).then(() => {
+          Notify.success('FORM.INFO.DELETE_SUCCESS');
+          vm.loadRoles();
+        })
+          .catch(Notify.handleError);
+      });
+  };
+
+  function loadRoles() {
+    vm.loading = true;
+    RolesService.list(session.project.id)
+      .then(response => {
+        vm.gridOptions.data = response.data;
+      })
+      .catch(Notify.handleError)
+      .finally(() => {
+        vm.loading = false;
+      });
+  }
+
+  const columns = [{
+    field : 'label',
+    displayName : 'FORM.LABELS.NAME',
+    headerCellFilter : 'translate',
+  },
+  {
+    field : 'actions',
+    width : 100,
+    displayName : '',
+    headerCellFilter : 'translate',
+
+    cellTemplate : `modules/roles/templates/action.tmpl.html`,
+  },
+  ];
+
+  // ng-click="
+  vm.gridOptions = {
+    appScopeProvider : vm,
+    enableColumnMenus : false,
+    columnDefs : columns,
+    enableSorting : true,
+    data : [],
+    fastWatch : true,
+    flatEntityAccess : true,
+  };
+
+
+  vm.gridOptions.onRegisterApi = function intApi(gridApi) {
+    vm.gridApi = gridApi;
+  };
+  loadRoles();
+}
--- a/client/src/modules/roles/roles.html
+++ b/client/src/modules/roles/roles.html
@@ -4,12 +4,12 @@
       <li class="static" translate> TREE.ADMIN </li>
       <li class="title" translate>TREE.ROLE_MANAGEMENT</li>
     </ol>
-
+    
     <div class="toolbar">
       <div class="toolbar-item">
-        <button class="btn btn-default text-capitalize" id="create" ng-click="RolesCtrl.add()" data-method="create">
+        <button data-method="add" class="btn btn-default text-capitalize" ng-click="RolesCtrl.add()" data-method="create">
           <span class="fa fa-plus"></span>
-          <span translate>FORM.LABELS.ADD</span>
+          <span  translate>FORM.LABELS.ADD</span>
         </button>
       </div>
     </div>
@@ -24,6 +24,7 @@
   <div class="row">
     <div class="col-lg-12">
       <div class="grid-full-height" 
+        id = 'roles-grid'
         ui-grid-auto-resize 
         ui-grid-empty-base-layer 
         ui-grid="RolesCtrl.gridOptions"
@@ -31,7 +32,6 @@
         ui-grid-resize-columns
         ui-grid-move-columns>
 
-        
         <bh-grid-loading-indicator
         loading-state="RolesCtrl.loading"
         empty-state="RolesCtrl.gridOptions.data.length === 0">
--- a/client/src/modules/roles/roles.html
+++ b/client/src/modules/roles/roles.html
@@ -1,0 +1,46 @@
+<div class="flex-header">
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static" translate> TREE.ADMIN </li>
+      <li class="title" translate>TREE.ROLE_MANAGEMENT</li>
+    </ol>
+
+    <div class="toolbar">
+      <div class="toolbar-item">
+        <button class="btn btn-default text-capitalize" id="create" ng-click="RolesCtrl.add()" data-method="create">
+          <span class="fa fa-plus"></span>
+          <span translate>FORM.LABELS.ADD</span>
+        </button>
+      </div>
+    </div>
+  </div>
+</div>
+
+<!-- main content -->
+<div class="flex-content">
+
+
+  <!-- Row start -->
+  <div class="row">
+    <div class="col-lg-12">
+      <div class="grid-full-height" 
+        ui-grid-auto-resize 
+        ui-grid-empty-base-layer 
+        ui-grid="RolesCtrl.gridOptions"
+        ui-grid-selection
+        ui-grid-resize-columns
+        ui-grid-move-columns>
+
+        
+        <bh-grid-loading-indicator
+        loading-state="RolesCtrl.loading"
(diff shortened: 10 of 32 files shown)


client/src/i18n/en/form.json
```diff
@@ -566,7 +567,8 @@
             "SEPTEMBER": "September",
             "OCTOBER": "October",
             "NOVEMBER": "November",
-            "DECEMBER": "December"
+            "DECEMBER": "December",
+            "PERMISSION_AFFECT_SUCCESS":"Permissions succefully affected"
```
jniles: "Affect" in French is "Assign" in English.  So the proper translation for this would be "Permissions successfully assigned".

client/src/i18n/en/tree.json
```diff
@@ -88,6 +88,7 @@
     "TRANSACTION_TYPE": "Transaction Type Management",
     "USERS": "User Management",
     "VOUCHER_REGISTRY": "Voucher Registry",
-    "AGED_DEBTORS": "Aged Debtors"
+    "AGED_DEBTORS": "Aged Debtors",
+    "ROLE":"Role"
```
jniles: Let's name this "Role Management" to be like most other management modules.  See #2435.

client/src/modules/roles/create.html
```diff
@@ -0,0 +1,36 @@
+<form name='formRole' ng-submit="RolesAddCtrl.submit(formRole)" novalidate>
```
jniles: To make the loading indicator work, this needs to be `bh-submit` instead of `ng-submit`.

client/src/modules/roles/create.js
```diff
@@ -0,0 +1,45 @@
...
+  function submit(form) {
+
+    form.$setSubmitted();
+    if (form.$invalid) {
+      return false;
+    }
+
+    var operation = (!data.uuid)?RolesService.create(vm.role) : RolesService.update(data.uuid, vm.role);
+
+    operation.then(function (response) {
```
jniles: To make the loading indicator work, this should be `return operation`.

client/src/modules/roles/create.js
```diff
@@ -0,0 +1,45 @@
...
+
+function RolesAddController(data, $state, $rootScope, $uibModalStack, RolesService, Notify) {
+
+  var vm = this;
+  vm.close = close;
+  vm.submit = submit;
+  vm.role = angular.copy(data);
+  vm.action = (!vm.role.uuid) ? 'FORM.LABELS.CREATE' : 'FORM.LABELS.UPDATE';
+  vm.title = (!vm.role.uuid) ? 'Nouveau rôle' : 'Modification du rôle';
+  vm.isCreate = (!vm.role.uuid);
```
jniles: To make this more readable, I suggest reordering these variables like this:
```js
var isCreateState = (!vm.role.uuid);
vm.isCreate = isCreateState;
vm.action = isCreateState ? 'FORM.LABELS.CREATE' : 'FORM.LABELS.UPDATE';
vm.title = isCreateState ?  'Nouveau rôle' : 'Modification du rôle';
```

That way other programmers will understand what `!vm.role.uuid` means ... it means you are in the create state.

client/src/modules/roles/create.js
```diff
@@ -0,0 +1,45 @@
...
+];
+
+function RolesAddController(data, $state, $rootScope, $uibModalStack, RolesService, Notify) {
+
+  var vm = this;
+  vm.close = close;
+  vm.submit = submit;
+  vm.role = angular.copy(data);
+  vm.action = (!vm.role.uuid) ? 'FORM.LABELS.CREATE' : 'FORM.LABELS.UPDATE';
+  vm.title = (!vm.role.uuid) ? 'Nouveau rôle' : 'Modification du rôle';
```
jniles: This should be translated.

client/src/modules/roles/create.js
```diff
@@ -0,0 +1,45 @@
...
+    form.$setSubmitted();
+    if (form.$invalid) {
+      return false;
+    }
+
+    var operation = (!data.uuid)?RolesService.create(vm.role) : RolesService.update(data.uuid, vm.role);
+
+    operation.then(function (response) {
+      Notify.success('FORM.INFO.OPERATION_SUCCESS');
+      var r = response;
```
jniles: Do you ever use `r`?  If not, can you remove this line?

client/src/modules/roles/create.js
```diff
@@ -0,0 +1,45 @@
...
+      $state.reload();
```
jniles: Why do you need to reload the state here?  If you are trying to refresh the underlying grid, it is much better to do that in the parent's controller.

client/src/modules/roles/create.js
```diff
@@ -0,0 +1,45 @@
+angular.module('bhima.controllers')
+  .controller('RolesAddController', RolesAddController);
+
+RolesAddController.$inject = [
+  'data', '$state', '$rootScope', '$uibModalStack', 'RolesService', 'NotifyService',
+];
+
+function RolesAddController(data, $state, $rootScope, $uibModalStack, RolesService, Notify) {
```
jniles: Using `$uibModalStack` directly is a bit of a hack - we use it on `$state` changes to ensure that all modals are closed, but it's not proper.

A better thing to use here is `$uibModalInstance`, which will be injected into this controller by default.  You can then use `$uibModalInstance.dismiss()` to reject the modal's promise or `$uibModalInstance.close()` to end the promise with success.  In both cases, the modal is closed.

See the [UI Bootstrap documentation](https://angular-ui.github.io/bootstrap/#!#modal) for more usage information.

client/src/modules/roles/create.js
```diff
@@ -0,0 +1,45 @@
...
+    var operation = (!data.uuid)?RolesService.create(vm.role) : RolesService.update(data.uuid, vm.role);
+
+    operation.then(function (response) {
+      Notify.success('FORM.INFO.OPERATION_SUCCESS');
+      var r = response;
+      $state.reload();
+      vm.close();
+    })
+      .catch(function (err) {
+        Notify.danger('echec');
```
jniles: This should use a translate key.  Something like `FORM.ERRORS.XXX`.

client/src/modules/roles/create.js
```diff
@@ -0,0 +1,45 @@
...
+      });
+
+  }
+
+
+  function close() {
+    // transition to the overall UI grid state - this modal will be cleaned up on state change
```
jniles, line 38: This is a nice comment!  As mentioned above, the best way to do this is with `$uibModalInstance` instead of with the `$uibModalStack`.

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -0,0 +1,58 @@
+<div class="modal-header">
+   
+  <span translate>USERS.PERMISSIONS</span> :
+  <span class="text-primary"> {{RolesPermissionsCtrl.role.label}}</span>
+</div>
+<div class="modal-body" style="max-height: 70vh; overflow:auto;">
```
jniles: Just a note: at some point we need to standardize the heights/overflow of all modals.  We currently do it several different ways.

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -0,0 +1,58 @@
...
+
+  <div class='row'>
+    <div class="col-lg-12">
+    
+      <div
+        ng-repeat="module in RolesPermissionsCtrl.tree"
+        class="treeview col-lg-12">
```
jniles: What is the `treeview` class?

Also, you already have a `col-lg-12` above, so this one is unnecessarily.  All `col-*-*` classes should be direct children of `row` classes.

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -0,0 +1,58 @@
...
+      
+          <h4>  
+            <label class="radio-inline">
+              <input type='checkbox' ng-model = 'module.affected' ng-true-value="1" ng-false-value="0"/>
+              <b translate>{{module.key}}</b> 
+            </label>
+          </h4>
+          
+          <ul class="list-group"  style="max-height: 70vh; overflow:auto;">
```
jniles: Does this need a `max-height`?  You already have a height limit on the parent, do you need it on the child?

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -0,0 +1,58 @@
...
+  </div>
+      
+    </div>
+
+  </div>
+
+</div>
+
+<div class="modal-footer">
+  <div class="pull-left">
```
jniles: Do you need this to be `pull-left`?  I think it should be left-aligned by default, no?

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -0,0 +1,58 @@
...
+    <label class="radio-inline">
```
jniles: In general, we put the "select all" at the top of the selection.  That is how the [the current permissions modal works](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/users/userPermission.modal.html#L17) and how ui-grid structures its selection.  In my view, we should keep the same design pattern for the whole application.  Could you move this checkbox to the first value in the list?

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -0,0 +1,58 @@
...
+      <input type='checkbox' 
+        ng-change="RolesPermissionsCtrl.selectAll()"
+        ng-model = 'RolesPermissionsCtrl.allPage'
+        ng-true-value="1" 
+        ng-false-value="0"/>
+      <span translate>FORM.LABELS.CHECK_ALL</span>
+    </label>
+  </div>
+  <div pull-right>
+    <button class='btn inverse ' ng-click='RolesPermissionsCtrl.close()' translate>FORM.BUTTONS.CANCEL</button>
```
jniles: Is `inverse` really a class?  I can't find where it is defined.

In general, we use `btn btn-default` for our cancel buttons.  Can you use this style for this cancel button?

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -0,0 +1,58 @@
...
+  
+    <button data-method="cancel" type="button" class="btn btn-primary" ng-click="RolesPermissionsCtrl.getAffected()">
```
jniles: This `data-method` should not be "cancel".  It should be "save" or "submit".

client/src/modules/roles/modal/rolesPermissions.js
```diff
@@ -0,0 +1,69 @@
...
+    $uibModalStack.dismissAll();
+  }
+
+  function selectAll() {
+
+    for (var i = 0; i < vm.tree.length; i++) {
+      var _module = vm.tree[i];
+      _module.affected = vm.allPage;
+      
+      for (var j = 0; j < _module.pages.length; j++) {
```
jniles: Another way to write this is to use functional programming (FP).  FP makes it easier to reason about code, since there are strict rules about what should change and what shoudn't change.

Here is a functional solution to this problem:
```js
function selectAll() {
  vm.tree.forEach(function (module) {
    module.pages.forEach(function (page) {
      page.affected = vm.allPage;
    });
    module.affected = vm.allPage;
  });
}
```

In English, this reads very nicely: "for each module in the tree, and for each page in the module, set affected to be vm.allPage for the pages and then set affected to be vm.allPage for the module."

client/src/modules/roles/modal/rolesPermissions.js
```diff
@@ -0,0 +1,69 @@
...
+        var page = _module.pages[j];
+        page.affected = vm.allPage;
+      }
+    }
+  }
+  function getAffected() {
```
jniles: This function is another great use of functional programming!  (Functional programming works really nicely on trees and recursive functions).  Here is the functional version of what you've written:

```js
function getAffected() {
  return vm.tree.reduce(function (ids, module) {
     if (module.affected === 1) { ids.push(module.id); }

     var pageIds = module.pages.reduce(function (list, page) {
       return list.concat(page.affected ? page.id : []);
     }, []);

     return ids.concat(pageIds);
  }, []);
}
```

If you want to be really fancy, you can reduce the code further, but this is potentially unclearer:
```js
function getAffected() {
  return vm.tree.reduce(function (ids, module) {
     var pageIds = module.pages.reduce(function (list, page) {
       return list.concat(page.affected ? page.id : []);
     }, []);

     return ids.concat(module.affected? module.id : [])
       .concat(pageIds);
  }, []);
}
```

client/src/modules/roles/roles.html
```diff
@@ -0,0 +1,44 @@
...
+    </div>
+  </div>
+</div>
+
+<!-- main content -->
+<div class="flex-content">
+
+
+  <!-- Row start -->
+  <div class="row top10">
```
jniles: I can't find the class "top10".  Could you point me to it?

client/src/modules/roles/roles.html
```diff
@@ -0,0 +1,44 @@
...
+    <div class="col-lg-12">
+      <div class="grid-full-height" ui-grid-auto-resize ui-grid-empty-base-layer ui-grid="RolesCtrl.gridOptions" ui-grid-selection
+        ui-grid-resize-columns ui-grid-move-columns>
+
+        <div class="grid-msg-overlay" ng-show='RolesCtrl.gridOptions.data.length==0'>
+          <!-- indicate that some error occurred -->
+          <div class="msg">
+            <span class="text-danger">
+              <span class="fa fa-warning"></span>
+              <span>La liste est vide</span>
```
jniles: This should be translated.

client/src/modules/roles/roles.html
```diff
@@ -0,0 +1,44 @@
...
+<div class="flex-content">
+
+
+  <!-- Row start -->
+  <div class="row top10">
+    <div class="col-lg-12">
+      <div class="grid-full-height" ui-grid-auto-resize ui-grid-empty-base-layer ui-grid="RolesCtrl.gridOptions" ui-grid-selection
+        ui-grid-resize-columns ui-grid-move-columns>
+
+        <div class="grid-msg-overlay" ng-show='RolesCtrl.gridOptions.data.length==0'>
```
jniles: You don't you use `bhGridLoadingIndicator`?  It should cover every state (loading, success, error) and be translated for you.

client/src/modules/roles/roles.service.js
```diff
@@ -0,0 +1,43 @@
...
+    return service.$http.get(url);
+  };
+
+  service.affectToUser = function (data) {
+    var url = '/roles/affectTo/user';
+    return service.$http.post(url, data);
+  }
+  service.userRoles = function (userId, projectId) {
+
+    var url = '/roles/user/'.concat(userId) + '/'.concat(projectId);
```
jniles: Just FYI: you can pass multiple value to [String.prototype.concat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat).  Like this:
```js
var url = '/roles/user/'.concat(userId, '/', projectId);
```
It removes the need to use '+'.

client/src/modules/roles/roles.ctrl.js
```diff
@@ -0,0 +1,106 @@
...
+
+  vm.remove = function (uuid) {
+    var message = 'Voulez-vous vraiment supprimer ce rôle?';
+    Modal.confirm(message)
+      .then(function(confirmResponse) {
+        if (!confirmResponse) {
+          return false;
+        }
+        RolesService.delete(uuid).then(function (response) {
+          var r = response;
```
jniles: Do you use `r` anywhere?   If not, go ahead and remove it.

client/src/modules/roles/roles.ctrl.js
```diff
@@ -0,0 +1,106 @@
...
+      resolve: {
+        data: function dataProvider() {
+          return role;
+        },
+      },
+    });
+  }
+
+  vm.remove = function (uuid) {
+    var message = 'Voulez-vous vraiment supprimer ce rôle?';
```
jniles: This message should be translated.

server/models/schema.sql
```diff
@@ -1606,6 +1606,41 @@ CREATE TABLE `user` (
...
+  FOREIGN KEY (`project_id`) REFERENCES `project` (`id`) ON UPDATE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+DROP TABLE IF EXISTS `user_role`;
+CREATE TABLE `user_role` (
+  `uuid` binary(16) NOT NULL,
+  `user_id` SMALLINT(5) UNSIGNED NOT NULL,
+  `role_uuid` binary(16) NOT NULL,
+  PRIMARY kEY(`uuid`),
+  UNIQUE `role_for_user` (`user_id`,`role_uuid`),
```
jniles, line 1677: These are nice constraint names!

server/config/routes.js
```diff
@@ -640,4 +642,18 @@ exports.configure = function configure(app) {
   app.post('/install', install.proceedInstall);
 
   app.get('/diagnoses', diagnoses.list);
-};
+
+
+  // roles
+  app.get('/roles/project/:project_id', rolesCtrl.list);
```
jniles: Instead of having `roles/project/:project_id`, why don't you make a url that lists all roles by default called `/roles`?  It would then use `project_id` as an optional HTTP query parameter (`?project_id=XYZ`).

That way you can use `RolesService.read()` on the client, and `FilterParser` on the server.  The route will also be flexible to have multiple conditions simply by adding them to `FilterParser`, like `?user_id=XYZ` to get all rules assigned to a user.  It will eliminate the need for having two different routes.

server/controllers/admin/unit.js
```diff
@@ -0,0 +1,79 @@
...
+      rows.forEach(row => {
+        queries[i] = db.exec(sql2, [roleUuid, roleUuid, row.id]);
+        i++;
+      });
+      return q.all(queries);
+    })
+    .then((permissions) => {
+
+      i = 0;
+      permissions.forEach(permission => {
```
jniles: The `forEach()` function can take in three parameters: the next value in the array, the index of that value in the array, and then a reference to the array itself.  There is no need to use the `i` variable here, this will work just fine:
```js
permissions.forEach((permission, idx) => {
  modules[idx].pages = permission;
});
```

For more detail, please see the MDN documentation on [Array.prototype.forEach()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).

server/lib/db/index.js
```diff
@@ -165,6 +166,14 @@ class DatabaseConnector {
     return Buffer.from(uuidParse.parse(hexUuid));
   }
 
+  /**
+   * @function uuid
+   * generate a uuid
+   */
+  uuid(){
```
jniles: Cool!  Can you include a space in between the `()` and the `{`?

server/controllers/admin/unit.js
```diff
@@ -0,0 +1,79 @@
+const db = require('../../lib/db');
+var q = require('q');
```
jniles: This should be `const`.

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,223 @@
...
+
+// affect roles to a user
+// roles ares permissions
+function affectToUser(req, res, next) {
+
+  const data = req.body;
+  const roles_uuids = data.role_uuids;
+  const userId = data.user_id;
+
+  var msg = {
```
jniles: This should be `const`.

server/config/routes.js
```diff
@@ -640,4 +642,18 @@ exports.configure = function configure(app) {
...
+  app.get('/roles/:uuid', rolesCtrl.detail);
+  app.post('/roles', rolesCtrl.create);
+  app.put('/roles/:uuid', rolesCtrl.update);
+  app.delete('/roles/:uuid', rolesCtrl.remove);
+  app.post('/roles/affect/units', rolesCtrl.affectPages);
```
jniles: Instead of `/roles/affect/units/`, it would be better to follow the same syntax as `permissions`.  For assigning permissions to a user, we use `POST /users/:id/permissions`.  This is read "assign to user id `:id` the following permissions".

So, here you could do `app.post(/roles/:uuid/units', roles.affectPages)`.  This would be read "assign the following units to the role with uuid `:uuid`".

server/config/routes.js
```diff
@@ -640,4 +642,18 @@ exports.configure = function configure(app) {
...
+  app.post('/roles/affectTo/user', rolesCtrl.affectToUser);
```
jniles: See the above comment - the same goes here.

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,223 @@
...
+}
+
+
+
+function detail(req, res, next) {
+
+  const sql = `
+   SELECT BUID(uuid) as uuid, label, project_id 
+   FROM role
+   WHERE  uuid = HUID(?) LIMIT 1
```
jniles: This works, but makes it harder to modify the SQL in the future.  The better way to do this operation is to convert the `req.params.uuid` to a binary array in JS and then do the comparison in SQL.  Like this:
```js
const binaryUuid = db.bid(req.params.uuid);

db.one(sql, binaryUuid)
```

jniles: Also, you shouldn't need `LIMIT 1` if you are really querying a uuid.   It should be unique, no?

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,223 @@
...
+module.exports.detail = detail;
+module.exports.create = create;
+module.exports.update = update;
+module.exports.remove = remove;
+module.exports.affectPages = affectPages;
+module.exports.affectToUser = affectToUser;
+module.exports.listForUser = listForUser;
+
+function list(req, res, next) {
+  const sql = `
```
jniles: If you use `FilterParser` here (see #2240), it will make this API much easier to extend in the future.

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,223 @@
...
+      res.json(rows);
+    })
+    .catch(next)
+    .done();
+}
+
+function update(req, res, next) {
+  const role = req.body;
+  delete role.uuid;
+  const sql = ` UPDATE role SET ? WHERE uuid = HUID(?)`;
```
jniles: See the above comments on using `HUID()`.

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,223 @@
...
+  const unitIds = data.unit_ids;
+  const roleUuid = data.role_uuid;
+  var msg = {
+    state : 1,
+    message : 'success',
+  };
+  const deleteFromRole = `DELETE FROM role_unit WHERE role_uuid = HUID(?) `;
+  const affectPage = `
+    INSERT INTO  role_unit(uuid, unit_id, role_uuid)
+    VALUES( HUID(uuid()), ?, HUID(?))`;
```
jniles: The SQL `uuid()` function is a version 1 uuid, while the uuid generated by our JS functions are version 4 uuids.  For broad overview of the differences, see [this SO answer](https://stackoverflow.com/questions/20342058/which-uuid-version-to-use/20342413#20342413).  It's best that we keep using the same version 4 uuid.

For a more general overview, [Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier) has a good summary.

client/src/modules/roles/modal/userRole.js
```diff
@@ -0,0 +1,94 @@
...
+      exportWidth: 250
+    },
+    {
+      field: '-',
+      width: 100,
+      displayName: 'Affecté',
+      enableFiltering: false,
+      cellTemplate: `
+      <div class="ui-grid-cell-contents">
+          <input id="checkbox3" type="checkbox"
```
jniles: The `id` attribute of an HTML node should be unique on the page, but this will be repeated multiple times.  It shouldn't have an id.

client/src/modules/roles/modal/userRole.js
```diff
@@ -0,0 +1,94 @@
...
+      displayName: 'Label',
+      exportable: 1,
+      exportWidth: 250
+    },
+    {
+      field: '-',
+      width: 100,
+      displayName: 'Affecté',
+      enableFiltering: false,
+      cellTemplate: `
```
jniles: Please don't use template strings on the client.  They will break our production build.

client/src/modules/roles/modal/userRole.html
```diff
@@ -0,0 +1,33 @@
+<div class="modal-header">
+  <h4><span translate>TREE.ROLE</span>
+    : <span class="text-primary">
+      {{UsersRolesCtrl.user.display_name}}
+    </span>
+  </h4>
+</div>
+<div class="modal-body">
+
+  <div class="row top10">
```
jniles: Where is `top10` defined?

client/src/modules/roles/modal/userRole.js
```diff
@@ -0,0 +1,94 @@
...
+    var data = {
+      user_id: vm.user.id,
+      role_uuids: codes
+    }
+    RolesService.affectToUser(data)
+      .then(function (response) {
+        Notify.success('FORM.INFO.OPERATION_SUCCESS');
+        vm.close();
+      }).catch(function (err) {
+        alert(erreur)
```
jniles: Please use Notify instead of `console` or `alert`.

client/src/modules/roles/modal/userRole.js
```diff
@@ -0,0 +1,94 @@
...
+  function loadRoles() {
+
+    RolesService.userRoles(vm.user.id, session.project.id)
+      .then(function (response) {
+        delete vm.gridOptions.data;
+        vm.roles = response.data;
+
+      })
+      .catch(function (err) {
+        console.log(err)
```
jniles: Please use Notify instead of `console` or `alert`.

jniles (changes requested):
@jeremielodi, I've not used this module fully yet... but it is a step in the right direction.

I've made a lot of client-side comments.  The server side also needs work, but we'll clean that up in an additional review.  There's already plenty to work on here.

In general:
 1. Please use `let`/`const` on the server instead of `var`.  They will help you identify values that should change and those that should not be reassigned.
 2. Please don't use any version higher than ES5 on the client - this includes arrow functions, template strings, or any other newer concepts.  We currently cannot compile them to older browsers in our build process.
 3. When you are using modals, the `ui-bootstrap` library provides the `$uibModalInstance` $inject to control your modal.  There is no need to use `$uibModalStack` unless you are outside the modal controller.
 4. The work "affect" should be "assign".  "Affect" means to change in English, rather than to assign, which is what you mean here.
 5.  Try as much as possible to respect ESLint on the server.
 6. Please make sure that the CSS classes you use, or the HTML parameters that you put on your nodes are actually needed and used.
 7.  In the future, please make sure all values are translated when you make a pull request.  It makes it much easier to review.
 8. Please make sure all errors are handled with `Notify` and not `alert` or `console`.
 9.  Please make sure all variables that you assign are actually used.  It's quite confusing to read code that doesn't actually do anything - it makes your peers wonder if you've a mistake somewhere.  Remember, [programs are meant to be read by humans, and only incidentally for computers to execute](https://www.quora.com/How-true-is-Programs-are-meant-to-be-read-by-humans-and-only-incidentally-for-computers-to-execute).

Let me know when you've gotten a chance to take a look at this and I'll give you another review of the server side code.

client/src/modules/roles/create.js
jeremielodi: you're right, I don't use it any more howere

client/src/modules/roles/modal/rolesPermissions.js
jeremielodi: I know but it seem to be an array function, right? 
I usually use this structure at the server side

jniles: Ah, array functions are fine.  They are in ES5.  It's "fat arrow" functions that you don't want to use until #2472 is fulfilled.  So `Array.prototype.forEach()` is fine and `Array.prototype.map()` and the rest are okay.

The thing that _doesn't_ work is `var x = (y) => y + 1;`.  That is an "arrow function" since it is a function that doesn't use the word `function`.  Those are ES2015+.

client/src/modules/roles/create.html
```diff
@@ -0,0 +1,39 @@
+<form 
+  name='formRole'
```
jniles: Almost all our forms are named `XxxxxForm`.  For example, [PatientRegistrationForm](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/patients/registration/registration.html#L15), [InvoiceForm](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/invoices/patientInvoice.html#L14), and [EmployeeRegistrationForm](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/employees/registration/employees.html#L22).  Could keep that standard and name this `RolesForm` or something similar?

client/src/modules/roles/create.html
```diff
@@ -0,0 +1,39 @@
...
+    <ol class="headercrumb">
+      <li>
+        <span translate>TREE.ROLE_MANAGEMENT</span>
+        <label class="label label-warning text-uppercase" translate>{{RolesAddCtrl.action}}</label>
+      </li>
+    </ol>
+  </div>
+
+  <div class="modal-body">
+    <div class='row'>
```
jniles: Since you don't have more than one row or one giant column, I propose you remove the `<div class="row">` and `<div class="col-log-12">`.  They don't do anything.

client/src/modules/roles/create.html
```diff
@@ -0,0 +1,39 @@
...
+          <div class="help-block" ng-messages="OptionalForm.health_area.$error" ng-show="formRole.$submitted">
+            <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+
+  <div class="modal-footer">
+    <button data-method="cancel" type="button" class="btn btn-default" ng-click="RolesAddCtrl.close()">
```
jniles, line 25: 👍

client/src/modules/roles/create.html
```diff
@@ -0,0 +1,39 @@
...
+    </ol>
+  </div>
+
+  <div class="modal-body">
+    <div class='row'>
+      <div class="col-lg-12">
+        <div class="form-group" ng-class="{'has-error' : formRole.label.$invalid && formRole.$submitted}">
+          <label class="control-label" translate>FORM.LABELS.NAME</label>
+          <input type="text" class="form-control col-md-9" name="label" ng-model="RolesAddCtrl.role.label" required>
+          <div class="help-block" ng-messages="OptionalForm.health_area.$error" ng-show="formRole.$submitted">
```
jniles: This `help-block` needs to be updated to refer to your current form and not `OptionalForm`.

client/src/modules/roles/create.js
```diff
@@ -0,0 +1,42 @@
+angular.module('bhima.controllers')
+  .controller('RolesAddController', RolesAddController);
+
+RolesAddController.$inject = [
+  'data', '$state', '$rootScope', 'RolesService', 'NotifyService', '$uibModalInstance',
```
jniles: It doesn't look like you use `$rootScope`. Could you remove all unused $injects?

client/src/modules/roles/modal/userRole.js
```diff
@@ -0,0 +1,83 @@
+angular.module('bhima.controllers').controller('UsersRolesController', UsersRolesController);
+UsersRolesController.$inject = ['data', '$uibModal', '$rootScope',
+  '$uibModalInstance', 'RolesService', 'SessionService', 'NotifyService',
+];
+
+function UsersRolesController(data, $uibModal, $rootScope, $uibModalInstance, RolesService, session, Notify) {
```
jniles, line 6: It doesn't look like you use `$rootScope` or `$uibModal`.  Could you remove this and all other unused imports?

client/src/modules/roles/modal/userRole.js
```diff
@@ -0,0 +1,83 @@
...
+        vm.roles = response.data;
+      })
+      .catch(Notify.handleError);
+  }
+
+  // assigned role to he user
+  function getAffected() {
+    var codes = [];
+    var list = vm.roles;
+    list.forEach(role => {
```
jniles: Cool!  Just FYI, you could also do this:
```js
const codes = vm.roles
  .filter(role => role.affected === 1)
  .map(role => role.uuid);
```
to accomplish the same thing :)

client/src/modules/roles/modal/userRole.js
```diff
@@ -0,0 +1,83 @@
...
+    RolesService.userRoles(vm.user.id, session.project.id)
+      .then(function (response) {
+        delete vm.gridOptions.data;
+        vm.roles = response.data;
+      })
+      .catch(Notify.handleError);
+  }
+
+  // assigned role to he user
+  function getAffected() {
```
jniles: Normally, a function called `get` would return a value, or send a `GET` HTTP request.  In this case, it sends a POST and doesn't return anything, which is kind of confusing to read.

Could you rename this function `assignRolesToUser()` or `submitRole()` or something like that?

client/src/modules/roles/modal/userRole.js
```diff
@@ -0,0 +1,83 @@
...
+        vm.close();
+      }).catch(Notify.handleError);
+
+  }
+  
+  // ui-grid
+  var columns = [{
+      field: 'label',
+      displayName: 'Label',
+      exportable: 1,
```
jniles: What does exportable do? Do you ever use it?  If not, go ahead and remove it.

client/src/modules/roles/templates/userAssignedRole.cell.html
```diff
@@ -0,0 +1,8 @@
+<div class="ui-grid-cell-contents">
+	<input 
+		type="checkbox" 
+    ng-model='row.entity.affected'
+    ng-true-value="1" 
+    ng-false-value="0" 
+  	ng-checked="{{row.entity.affected==1}}">
```
jniles: It looks like this would work without the `ngChecked` directive.  Have you tried that?

Otherwise, you could try `ng-checked="row.entity.affected"`, since the true/false values are 1/0.

client/src/modules/users/templates/grid/action.cell.html
```diff
@@ -19,6 +19,12 @@
         <i class="fa fa-briefcase"></i> <span translate>FORM.BUTTONS.CASHBOX_MANAGEMENT</span>
       </a>
     </li>
+    <li>
+        <a data-method="cashbox" ng-click="grid.appScope.editRoles(row.entity)" href>
```
jniles: I think this `data-method` should be `edit-roles` instead of `cashbox`.

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,211 @@
...
+
+  const filters = new FilterParser(req.query, { tableAlias : 'r' });
+
+  let sql = `
+    SELECT BUID(r.uuid) as uuid, r.label, r.project_id 
+    FROM role  r
+  `;
+
+  filters.equals('project_id');
+  sql = filters.applyQuery(sql);
```
jniles, line 26: Nicer 👍

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,211 @@
...
+    }).catch(next)
+    .done();
+}
+
+
+function detail(req, res, next) {
+  const sql = `
+   SELECT BUID(uuid) as uuid, label, project_id 
+   FROM role
+   WHERE  uuid = ?  LIMIT 1
```
jniles: Do you need LIMIT 1 here?

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,211 @@
...
+    VALUES( ?, ?, ?)`;
+
+  db.exec(deleteFromRole, roleUuid)
+    .then(() => {
+      const t = [];
+
+      if (!unitIds) {
+        return;
+      }
+      if (_.isArray(unitIds)) {
```
jniles: You can eliminate this `if` by simply forcing `unitIds` to be an array.  Like this:
```js
// this works if unitIds is a number or an array, converting both into an array.
const ids = [].concat(unitIds);
```

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,211 @@
...
+        unitIds.forEach((id, idx) => {
```
jniles: Instead of using an index to make a new array, just use [Array.prototype.map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).  It will return a new array of whatever return values you specify.  Like this:
```js
const promises = unitIds.map(id => db.exec(affectPage, [db.uuid(), id, roleUuid]));

return q.all(promises);
```

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,211 @@
...
+
+// affect permission to a specific role
+
+function affectPages(req, res, next) {
+  const data = req.body;
+
+  const unitIds = data.unit_ids;
+  const roleUuid = db.bid(data.role_uuid);
+
+  const msg = {
```
jniles: It doesn't look like you ever use this on the client.  If you want to send a success status without a message, use ExpressJS's [res.sendStatus()](https://expressjs.com/en/4x/api.html#res.sendStatus) function.

So, you code would change from:
```js
const msg = {
  state : 1,
  message : 'success'
};

// .... later ...

res.json(message);
```
to 
```js
res.sendStatus(201); // send HTTP's create success code
```

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// affect roles to a user
+// roles ares permissions
+function affectToUser(req, res, next) {
+  const data = req.body;
+  const roles_uuids = data.role_uuids;
+  const userId = data.user_id;
+
+  const msg = {
```
jniles: This could also use `sendStatus()`

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,211 @@
...
+    state : 1,
+    message : 'success',
+  };
+
+
+  const deleleUserRoles = `DELETE FROM user_role WHERE user_id=? `;
+  const addRole = ` INSERT INTO user_role SET ?`;
```
jniles: This will generate an INSERT statement for every single role.  A more efficient query could use `INSERT INTO user_role (uuid, role_uuid, user_id) VALUES ?` to insert multiple items.  This is because of how [mysqljs escapes values](https://github.com/mysqljs/mysql#user-content-escaping-query-values).  Specifically (from the documentation):

> Nested arrays are turned into grouped lists (for bulk inserts), e.g. [['a', 'b'], ['c', 'd']] turns into ('a', 'b'), ('c', 'd')

This means that you can then pass an array of arrays to `db.exec()` and it will turn it into `(valueA, valueB), (valueC, valueD)`, which will allow you to use only one INSERT statement.  I'll show you below.

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,211 @@
...
+
+  db.exec(deleleUserRoles, userId)
+    .then(() => {
+      if (!roles_uuids) {
+        return msg;
+      } else if (_.isArray(roles_uuids)) {
```
jniles: This could also use the array trick that I showed you up above.  Like this:
```js
const roleUuids = [].concat(roles_uuids);

// this will be an array of arrays.  Each array will contain the 
// [uuid, role_uuid, and user_id] in that order.
const parameters => rolesUUids.map(roleUuid => [db.uuid(), db.bid(roleUuid), userId]);

return db.exec(addRoles, parameters);
```

server/models/bhima.sql
```diff
@@ -76,8 +76,8 @@ INSERT INTO unit VALUES
   (184, 'Job Title Management','TREE.PROFESSION','', 1,'/modules/functions/','/functions'),
   (185, 'Payroll Rubric Management','TREE.PAYROLL_RUB_MANAGEMENT','', 57,'/modules/payroll/rubrics','/payroll/rubrics'),
   (186, 'Holidays Management','TREE.HOLIDAYS_MANAGEMENT','Holidays Management',57,'/modules/holidays/','/holidays'),
-  (187, 'Offdays Management','TREE.OFFDAYS_MANAGEMENT','Offdays Management', 1,'/modules/offdays/','/offdays');
-
+  (187, 'Offdays Management','TREE.OFFDAYS_MANAGEMENT','Offdays Management', 1,'/modules/offdays/','/offdays'),
+  (188, 'Role management','TREE.ROLE_MANAGEMENT','Roles Management', 1,'/modules/role/','/role');
```
jniles: All other routes are plural - `/invoices`, `/patients`, `/users`, etc...  Could you name this `/roles` to keep that pattern?

server/controllers/admin/roles.js
```diff
@@ -0,0 +1,211 @@
...
+}
+
+
+// retireves affected and not affected role by a user id
+function listForUser(req, res, next) {
+  const userId = req.params.user_id;
+  const projectId = req.params.project_id;
+
+  const sql = `
+    SELECT DISTINCT q.uuid, q.label, q.project_id , q.affected
```
jniles, line 123: If I understand your query, you want to have every role in the database, and a boolean column "affected" that will tell if it is assigned to the user or not.  This is the ideal situation for a LEFT JOIN and a GROUP BY.  Here is how it will work:
 1. The LEFT JOIN will ensure that we get every role, even if there is no corresponding value in the user_role table for that role.  (If you need practice remembering what LEFT JOINs do, [this is a good article about them](http://www.mysqltutorial.org/mysql-left-join.aspx)).
 2. The GROUP BY will make sure that we only get one value for each role, and allow us to check if the user is assigned to it.
Like this:
```sql
SELECT BUID(r.uuid) AS uuid, r.label, r.project_id, MAX(IF(ur.user_id = ?, 1, 0)) AS affected
FROM role AS r
LEFT JOIN user_role AS ur ON r.uuid = ur.role_uuid
GROUP BY role.uuid
WHERE project_id = ?;
```

jniles: MySQL version 5.7+ may complain because GROUP BY statements should specify an aggregation value for each column.  If that happens to you, add MAX() or MIN() to each column (besides the grouping column).

server/controllers/admin/unit.js
```diff
@@ -0,0 +1,72 @@
...
+ * each modules and his pages
+ */
+function list(req, res, next) {
+  const roleUuid = db.bid(req.params.roleUuid);
+
+  // permission(pages) in a module
+  // the column affect will inform is the user has acces to the permission
+  const sql = `
+    SELECT DISTINCT rqt.id,  rqt.name, rqt.key , rqt.parent, rqt.affected FROM (
+      SELECT u.*,  1 as affected
```
jniles: This could also use the GROUP BY trick above.

server/controllers/admin/unit.js
```diff
@@ -0,0 +1,72 @@
...
+  const sql1 = `${sql}\n WHERE rqt.parent=0`;
+  // children
+  const sql2 = `${sql}\n WHERE rqt.parent=?`;
+
+  let modules = [];
+  const queries = [];
+
+  db.exec(sql1, [roleUuid, roleUuid]).then((rows) => {
+    modules = rows;
+    rows.forEach((row, idx) => {
```
jniles: A good rule to remember is that every time you do this kind of statement:
```js
const list = []
array.forEach(x => {
  list.push(x);
}):
return list
```
.. you are actually doing a `map()`.  In the above case, you can do this:
```js
return q.all(rows.map(row => db.exec(sql2, [roleUuid, roleUuid, row.id])));
```

server/controllers/admin/unit.js
```diff
@@ -0,0 +1,72 @@
...
+        modules[idx].pages = permission;
+      });
+
+      res.json(modules);
+    })
+    .catch(next)
+    .done();
+}
+
+function userAffectedUnit(req, res, next) {
```
jniles: It doesn't look like you ever use this function ... can it be removed?

client/src/modules/roles/modal/rolesPermissions.js
```diff
@@ -0,0 +1,66 @@
...
+
+  // vm.role.uuid
+  RolesService.unit(vm.role.uuid)
+    .then(
+      function (res) {
+        vm.tree = res.data;
+      });
+
+  // close modal
+  function close() {
```
jniles: This should use `$uibModalInstance` instead of the ModalStack.

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -0,0 +1,50 @@
...
+    </div>
+
+  </div>
+</div>
+<div class="modal-footer">
+
+  <div pull-right>
+    <button class='btn btn-default' ng-click='RolesPermissionsCtrl.close()' translate>FORM.BUTTONS.CANCEL</button>
+
+    <button data-method="save" type="button" class="btn btn-primary" ng-click="RolesPermissionsCtrl.getAffected()">
```
jniles: It would make a much better user experience if you included a loading indicator.  You can get one for free if you use the `bhLoadingButton`.

jniles (commented):
@jeremielodi, I've completed a second review and made some more server-side suggestions.  Let me know what you think!

Feel free to use whatever version of JS you want on the client now, since @mbayopanda submitted typescript PR.

server/controllers/admin/roles.js
jeremielodi: Your query is good if a role has been already assigned to a user else it returns an empty array.
also ``` GROUP BY role.uuid ```  must come after  ```WHERE project_id = ?```
So I keep mine for the moment and think of an efficient one that we can use instead

jniles: You're right ... it was too late at night when I wrote that.  I'll play around with it on the command line and see if I can come up with something better.

server/controllers/admin/roles.js
jeremielodi: this doesn't work for my Mysql version(MariaDb), I will install mysql server  to get it work

jniles: It shouldn't depend on the MySQL version.  It's a part of the npm module.  What code are you trying?

jeremielodi: inserting multiple rows together

jniles (commented):
@jeremielodi, it doesn't look like there are any tests in this PR.  Could you add them?  Anytime we write a new module, we need to include end to end tests to ensure they work.  If you need help, @lomamech is our expert as writing tests.

jniles (commented):
@jeremielodi these end to end tests look really good!

End to end tests are critical to catching errors that break the clientside, but they take a long time to complete.  To help us catch server-side errors, and catch general errors faster, we also usually write some integration tests.  You can find some documentation on how to do that [in this wiki](https://github.com/IMA-WorldHealth/bhima-2.X/wiki/Testing:-Integration-Testing-Server-APIs).

Could you add some integration tests for the `/roles` API route?

After that, we will merge this and make progressive refinements, like all modules.

test/integration/role.js
```diff
@@ -0,0 +1,62 @@
...
+  const roleAdmin = {
+    label : 'Administrator',
+  };
+  const userRole = {
+    user_id : 2,
+    role_uuids : adminUuid,
+  };
+
+  it('GET /roles returns a list of roles', () => {
+    return agent.get('/roles?project_id='.concat(roles.project_id))
```
jniles: Just FYI - you can use chai-http's `.query()` method to pass queries as an object.  Like this:
```js
return agent.get('/roles')
  .query({ project_id : 1 });
 // this becomes HTTP GET /roles?project_id=1
```
It makes it easier to update and change values, since you can then do `agen.get('/roles').query(roles);` for example.

test/integration/role.js
```diff
@@ -0,0 +1,62 @@
...
+        expect(res.body).to.not.be.empty;
+      })
+      .catch(helpers.handler);
+  });
+
+  it('POST /roles add a new role', () => {
+    return agent.post('/roles')
+      .send(roles)
+      .then((res) => {
+        expect(res).to.have.status(200);
```
jniles: Normally, POST should send back 201 if it was created.  Check out the [w3 spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) on it for detailed information, or our own [API Guidelines](https://github.com/IMA-WorldHealth/bhima-2.X/wiki/Design:-Server-CRUD-APIs).

jniles (commented):
One quick change: the POST should return a 201.  Didn't notice this until the test.  After that, LGTM.

jniles (approved):
LGTM.  Well done!

bors r+

PR #2462, opened 2018-01-24
Refactor General Ledger

### LINKED ISSUE #1706: refactor: Restructure account slip structure
Currently the account slip is used to show detailed information on transactions that have hit an account. It is very similar as the 'Account Report'. I suggest that we restructure these reports to minimise the amount of code to be maintained and simplify future development. 

File structure: 
- [ ] Move account slip under the `accounts` report folder so that it can be found at `finance/reports/accounts/slip`
- [ ] Separate the account slip code from the general ledger controller so that it can be found at `finance/reports/accounts/slip.js`

UX structure: 
* Use the exact same template/ controller for both the report and the slip that can be displayed within modules
* Allow easy filtering of the slip to only show data from a specific time period

Report logic: 
- [ ] Use MySQL or well defined JavaScript to calculate totals so that the template is not responsible for this - contributes to https://github.com/IMA-WorldHealth/bhima-2.X/issues/1668
- [ ] Separate `getAccountTransactions` from the `reportAccounts.js` controller so that it can be used by other services/ controllers

### LINKED ISSUE #2373: Implement option to show title accounts and totals on the Balance Sheet
It would be really cool to show the title accounts along with their children on the Balance Sheet report.  This would provide a reason for the title accounts to exist since we cut them out of most account selection views.  They would provide an overview of the total balance of an individual group.

### LINKED ISSUE #2459: Add links from the General Ledger to the Account Statement
The General Ledger currently only has a single link "Extrait de Compte".  This links to a PDF report of the account transactions for the month.

Another link that would be helpful is a link to the Account Statement filtered on that account, allowing a user to explore the transactions hitting that account, with the power of viewing all the details directly from this link.  Since the GL only contains posted values, it does not need to link to the Journal as well.

### LINKED ISSUE #2463: Improvements to the General Ledger
Once #2462 lands, we'll have a General Ledger that provides a view into the account totals by month, completely with summing of title accounts.  However, there is still more work to be done!

 1. Implement a sum of the entire chart of accounts.  This should always be zero, for every column.  If it is not, we have a rounding error.
 2. If the Fiscal Year does not have any data, you cannot change fiscal years (the `<select>` element is under the `<bh-grid-loading-indicator>`.  We should find a way to make it accessible.

jniles:
This PR refactors the General Ledger to make it more useful from an accounting perspective.

**Code Refactor**
 1. The server-side SQL query has been rewritten to be more performant and easier to read.  In particular, the WHERE query filters on fiscal year instead of first looking up the periods in the fiscal year.
 2. Removed the Account Slip code - it was essentially a duplication of effort from the Account Statement code.
 3. Renamed all variables references `general-ledger-accounts` to `general-ledger`.  We had folded the previous "General Ledger" into the Journal, so only this one remains.

**Features**
 1. A `<select>` element to the grid footer to use to switch the fiscal year and removes the Fiscal Year modal.  This was requested by @WayneNiles and improves the UX by allowing users to switch fiscal years on the fly.  By removing the modal, we also alleviate the maintenance burden on developers.
 2. A "check row" at the bottom of the General Ledger has been implemented, showing the sum of each column with period balances.  _These columns should always sum to zero_.  If not, it means that there is an imbalanced transaction somewhere.
 3. Title accounts are now show and hold the sum of the balances of their children.  This allows a really quick comparison of bulk amounts, such as Income versus Expense.  This has been implemented using a `Tree` class, available to be used in other reports.
 4. Added title account highlighting and indentation to both the client and server-side generated reports for the General Ledger.  The indentation is set to `15px` by default.
 5. Add a link to the Account Statement module.  This allows a user to directly inspect the values from the General Ledger in the Account Statement.

**Screenshots**
newgl
_Fig 1: New General Ledger Client_

newglreport
_Fig 2: New General Ledger Report w/ Indentation_

 Closes #1706.  Closes #2459.  Closes #2463. Closes #2373.


client/src/modules/general-ledger/general-ledger-accounts.ctrl.js
```diff
@@ -212,52 +235,27 @@ function GeneralLedgerAccountsController(
     vm.gridApi.core.notifyDataChange(uiGridConstants.dataChange.COLUMN);
   }
 
-  function labelTitleAccounts(account) {
-    account.isTitleAccount = account.type_id === bhConstants.accounts.TITLE;
+  function removeZeroes(account) {
+    // remove zero values from the matrix to render as empty cells.
+    fields.forEach(function (field) {
```
mbayopanda: 👍

server/lib/Tree.js
```diff
@@ -0,0 +1,138 @@
+/**
+ * @class Tree
```
mbayopanda, line 2: 👍 awesome... Good job

test/server-unit/Tree.spec.js
```diff
@@ -0,0 +1,119 @@
+const Tree = require('../../server/lib/Tree');
+const { expect } = require('chai');
+
+function TreeUnitTests() {
```
mbayopanda, line 4: 💃 Excellent

mbayopanda (commented):
LGTM

I just noticed that there is an error when click on `action -> Extrait de compte` to get the `extrait de compte` report.

mbayopanda:
@jniles this is the error thrown when we click on `action -> Extrait de compte` : 

error_action_general_ledger

jniles:
@mbayopanda, thanks for the bug report!  Clearly `undefined` is not a currency... :(

I should probably address all the problems with the GL in this PR to make sure that we get it right.  Here's my current plan of action:
 1. Fix #2463.  This should take care of the select bug you mentioned.
 2. Fix the bug you just found.  Whoops!
 3. Implement #2459.  This should give us a better way to validate the GL.

Thanks for the review!  It's good to have you back 👍

jniles:
@mbayopanda, I've completed all the work I said I would do.  When you get a chance, could I have another review?

mbayopanda (approved):
It works very well. :+1

PR #2467, opened 2018-01-25
feat(journal view) auto remove no cachable filters

### LINKED ISSUE #2367: "View in Journal" link should not cache filters
The "view in the journal" link from the trial balance caches the filters and saves them in the grid.  This behavior is confusing if the user doesn't expect it, and probably isn't all that helpful.  We should remove the filter cache.

jeremielodi:
This PR helps to automatically remove some filters whitch are not cachable.
closes  #2367

### DIFF
--- a/client/src/js/services/FilterService.js
+++ b/client/src/js/services/FilterService.js
@@ -87,14 +87,18 @@ function FilterService() {
   // return filters for the view - this method will always be compatible with the bhFilter component
   FilterList.prototype.formatView = function formatView() {
     var activeFilters = this._filterActiveFilters();
-    var activeKeys = activeFilters.map(function (filter) { return filter._key; });
+    var activeKeys = activeFilters.map(function (filter) {
+      return filter._key;
+    });
 
-    function keysInActive(filter) { return activeKeys.indexOf(filter._key) !== -1; }
+    function keysInActive(filter) {
+      return activeKeys.indexOf(filter._key) !== -1;
+    }
 
     // parse into two lists
     return {
-      defaultFilters : this._defaultFilters.filter(keysInActive),
-      customFilters : this._customFilters.filter(keysInActive)
+      defaultFilters: this._defaultFilters.filter(keysInActive),
+      customFilters: this._customFilters.filter(keysInActive),
     };
   };
 
@@ -130,6 +134,14 @@ function FilterService() {
   FilterList.prototype.formatCache = function formatCache() {
     return angular.copy(this._filterIndex);
   };
+
+  FilterList.prototype.removeUnCachableFilters = function removeUnCachableFilters(filters) {
+    filters.forEach(filter => {
+      if (filter.cachable === 0) {
+        delete this._filterIndex[filter.key];
+      }
+    });
+  }
 
   // replaces current filters with filters from cache
   FilterList.prototype.loadCache = function loadCache(storedCache) {
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -44,7 +44,9 @@ function JournalController(
   var columns;
 
   // store journal data
-  var journalStore = new Store({ identifier : 'uuid' });
+  var journalStore = new Store({
+    identifier : 'uuid',
+  });
   var transactionIdToRecordUuidMap;
 
   /** @const the cache alias for this controller */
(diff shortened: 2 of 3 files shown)


client/src/modules/journal/journal.js
```diff
@@ -70,13 +72,13 @@ function JournalController(
   // options, it is also used by the grid to expose the API
   // FIXME(@jniles) - why does this not have fastWatch?
   vm.gridOptions = {
-    enableColumnMenus          : false,
-    showColumnFooter           : true,
-    appScopeProvider           : vm,
-    flatEntityAccess           : true,
+    enableColumnMenus : false,
```
jniles, line 75: :+1:

client/src/modules/journal/journal.js
```diff
@@ -473,10 +494,10 @@ function JournalController(
     }
 
     return {
-      totalRows : totalRows,
-      totalPosted : totalPosted,
-      totalNonPosted : totalNonPosted,
-      remainingSystemTransactions : remainingSystemTransactions,
+      totalRows,
```
jniles, line 497: This will break our production build. :(.  Could you switch it back?

client/src/modules/journal/journal.js
```diff
@@ -510,19 +531,33 @@ function JournalController(
...
 
+  // initial loading in startup() function
+  function initLoading() {
+    load(Journal.filters.formatHTTP(true));
+    vm.latestViewFilters = Journal.filters.formatView();
+  }
   // runs on startup
   function startup() {
     var hasStateFilters = $state.params.filters.length > 0;
+    var noCachaleKeys = [];
```
jniles: It doesn't look like you use this.  Could you remove it?

client/src/modules/journal/journal.js
```diff
@@ -510,19 +531,33 @@ function JournalController(
...
     if (hasStateFilters) {
       Journal.filters.replaceFiltersFromState($state.params.filters);
+
+      initLoading();
+      for (var i = 0; i < $state.params.filters.length; i++) {
```
jniles: Hmm... this works, but we could use this functionality across the entire application if we made it a feature of the the `FilterService`.

I propose that you instead modify the [FilterService.prototype.formatCache()](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/FilterService.js#L130) function to only cache filters that are `cacheable`.  I imagine the code will look something like this:
```js
FilterList.prototype.formatCache = function formatCache() {
  var filters = {};

   // loop through each filter and only include it if it is cacheable
  angular.forEach(this._filterIndex, function (filter, key) {
     if (filter.cacheable === 0) { return; }
     filters[key] = filter;
  });

  return filters;
}
```

jniles: That way, you actually don't have to change any code in the `Journal` and other modules will immediately benefit by simply adding `cacheable: 0` to their filter parameters.

jniles (commented):
The `cacheable` key is an excellent solution!  I propose that you build it into the `FilterService` so that it can be used everywhere.

client/src/modules/journal/journal.js
jeremielodi: cool, but 
```js 
this._filterIndex  
``` 
does not save cachable index,
Another problem,  data should be loaded with all parameters for the first time and we can remove uncachable filters after
I created this function instead
```js
 // uncachable filters should not remain for a long time, so they can be remove if needed
  FilterList.prototype.removeUnCachableFilters = function removeUnCachableFilters(filters) {
    filters.forEach(filter => {
      if (filter.cachable === 0) {
        delete this._filterIndex[filter.key];
      }
    });
  };

File: client/src/modules/journal/journal.js (hunk 2 above)
Comment (jniles) [reply client/src/modules/journal/journal.js:497]: Nevermind!  #2473

File: client/src/js/services/FilterService.js [hunk 5]
```diff
@@ -131,6 +131,15 @@ function FilterService() {
     return angular.copy(this._filterIndex);
   };
 
+  // uncachable filters should not remain for a long time, so they can be remove if needed
+  FilterList.prototype.removeUnCachableFilters = function removeUnCachableFilters(filters) {
+    filters.forEach(filter => {
```
jniles: This is good code!

Instead of making a separate API call, I propose that you put this in the function `FilterService.prototype.formatCache()`.  Since that is where the cache is saved, if you put this first, they will never be cached.  Then you won't need to called `initLoading()` twice and can remove the `Journal.filters.removeUnCachableFilters($state.params.filters)` call.

jeremielodi: where  formatCache is called?

jniles: On [line 151 of the JournalService](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/journal/journal.service.js#L151).

To be clear, you would change the `formatCache()` function, not the `JournalService`.  Basically, the rule would be "if `uncacheable` is set, do not return this filter from the `formatCache()` function call".

jeremielodi: but when we have uncachable and cachable params in the filter we will have to load

jniles: Excuse me, I meant `cacheable`.  If `cacheable === 0`, as your code shows, it shouldn't return that filter from the `formatCache()` call.

I think that if you just change the formatCache, it will not have any affect on the HTTP request .. just on the ones that are saved.

jeremielodi:
@jniles, I tried to reduce lines of code in the startup function in journal.js 
I hope it's  the good way to solve this problem

jniles (approved):
Good enough.  Thanks!

bors r+

PR #2471, opened 2018-01-25
refactor(fiscal year): close FYs by period totals

### LINKED ISSUE #1651: (design) Closing/Opening a Fiscal Year via Period Totals
It doesn't seem to me that there is any reason to write transactions into the Posting Journal or GL.  See  this source for a detailed overview: [Démarche pour clôturer un exercice Comptable.pdf](https://github.com/IMA-WorldHealth/bhima-2.X/files/995805/Demarche.pour.cloturer.un.exercice.Comptable.pdf)

I propose that, when closing a fiscal year and opening a new one, we operate entirely on the `period_total`s table without introducing any transactions into the `posting_journal` or `general_ledger`. 
  1. This prevents the issues encountered in https://github.com/IMA-WorldHealth/bhima-2.X/issues/761. 
  2. It also prevents double-counting in reports that pull their opening and closing balances from period totals ... no need to filter the GL on the opening/closing transaction.

**Design: Fiscal Year Closing Algorithm**
When closing a Fiscal Year, the following actions occur:
 1. The application checks that there are no unposted transactions against the Fiscal Year.  If so, it must reject the procedure.  This directly prevents #1647.
 2. If (1) passes, _this means all balances are already in the `period_total`'s table_.  The user should confirm those balances.  At this point, the user should also specify the gain or loss account that will hold the balance of the `INCOME - EXPENSE` accounts.
 3. If the user confirms (2), the Fiscal Year should be closed by:
    a. Sum the balance of all balance accounts from Period 0 through Period 12 in the `period_total` table.  These numbers are then written to Period 13 by account.
    b. Write 0 for the `period_total`s of all `INCOME` and `EXPENSE` accounts in Period 13.
    c. Write the result of the `INCOME - EXPENSE` into Period 13 of the user-defined balance account.
    d. Mark the Fiscal Year as closed.

After this procedure is done, Period 13 will have 0s for all INCOME and EXPENSE accounts.  All other accounts will contain the sum of Period 0 through Period 12 in their Period 13.

**Design: Opening a Fiscal Year**
If we adopt the above procedure, the way to open a new Fiscal Year is to copy Period 13 from the previous Fiscal Year into Period 0 of the new Fiscal Year.  That is all.

### LINKED ISSUE #2464: Migrate Fiscal Year Closing to a Stored Procedure
The server-side fiscal year code for closing a fiscal year is written mostly in JavaScript.  It would make sense to migrate this to a stored procedure, since the creation of a fiscal year takes place in SQL as well.

The basic design API should be `CALL CloseFiscalYear(fiscalYearId, closingAccountId);` which will cause the income/expense accounts of the previous fiscal year to totaled and written into the opening balance of the following fiscal year's `closingAccountId`.

For more discussion on the design of Fiscal Year opening/closing, please see https://github.com/IMA-WorldHealth/bhima-2.X/issues/1651.

jniles:
This commit implements the closing fiscal year code as a stored
procedure `CloseFiscalYear(fiscalYearId, closingAccountId)`.  This
replaces the previous several hundred lines of JavaScript code with a
single database call, and optimises the process significantly. It also
removes the need for a period number 13 and importing the previous
year's balances.

The following things still need to be done:
 1. Remove the import code for the previous fiscal year balances.
 2. Remove the generation of period number 13.
 3. Document the FY creation procedures with comments.

Closes #1651.  Closes #2464.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "test:server-unit": "./node_modules/.bin/mocha --recursive --exit test/server-unit",
     "build": "./node_modules/.bin/gulp build",
     "build:db": "./sh/build-database.sh",
+    "build:clean" : "./sh/build-init-database.sh",
     "watch": "./node_modules/.bin/gulp watch-client",
     "release": "./node_modules/.bin/standard-version",
     "snyk-protect": "snyk protect",
--- a/server/controllers/finance/fiscal.js
+++ b/server/controllers/finance/fiscal.js
@@ -11,11 +11,11 @@
 
 const q = require('q');
 const _ = require('lodash');
-const uuid = require('uuid/v4');
 const db = require('../../lib/db');
 const Transaction = require('../../lib/db/transaction');
 const NotFound = require('../../lib/errors/NotFound');
-const BadRequest = require('../../lib/errors/BadRequest');
+
+const debug = require('debug')('FiscalYear');
 
 // Account Service
 const AccountService = require('./accounts');
@@ -162,7 +162,9 @@ function create(req, res, next) {
  * Returns the detail of a single Fiscal Year
  */
 function detail(req, res, next) {
-  const id = req.params.id;
+  const { id } = req.params;
+
+  debug(`#detail() looking up FY${id}.`);
 
   lookupFiscalYear(id)
     .then((record) => {
@@ -176,7 +178,7 @@ function detail(req, res, next) {
  * Updates a fiscal year details (particularly id)
  */
 function update(req, res, next) {
-  const id = req.params.id;
+  const { id } = req.params;
   const sql = 'UPDATE fiscal_year SET ? WHERE id = ?';
   const queryData = req.body;
 
@@ -187,6 +189,8 @@ function update(req, res, next) {
 
   // remove the id before updating (if the ID exists)
   delete queryData.id;
+
+  debug(`#update() updating column ${Object.keys(queryData)} on FY${id}.`);
 
   lookupFiscalYear(id)
     .then(() => db.exec(sql, [queryData, id]))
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "test:server-unit": "./node_modules/.bin/mocha --recursive --exit test/server-unit",
     "build": "./node_modules/.bin/gulp build",
     "build:db": "./sh/build-database.sh",
+    "build:clean" : "./sh/build-init-database.sh",
```
mbayopanda, line 18: 👍

server/models/procedures/time_period.sql
```diff
@@ -96,3 +96,97 @@ BEGIN
...
+subtracting the total income from the total expense.
+ 4. Write the balances of every account that isn't an income or expense account
+to the period 0 of the subsequent year.
+ 5. Write the balance of income and expense accounts (step 3) to the period 0
+value of the account provided. If this account had a previous value, add the two
+to get the final opening balance.
+
+TODO - check that there are no unposted records from previous years.
+*/
+CREATE PROCEDURE CloseFiscalYear(
```
mbayopanda, line 118: 👍 very good

server/controllers/finance/fiscal.js
```diff
@@ -307,9 +315,14 @@ function lookupBalance(fiscalYearId, periodNumber) {
  * set the opening balance for a specified fiscal year
  */
 function setOpeningBalance(req, res, next) {
-  const id = req.params.id;
+  const {
+    id,
+  } = req.params;
```
mbayopanda, line 320: Is it possible to write this in one line, such as : 
```js
const { id } = req.params;
```
It will be simply clear

mbayopanda (approved):
I tried to close the 2018 fiscal year but i cannot see the any balance in the closing modal ?

image

PR #2473, opened 2018-01-26
Compile js with typescript into es5

### LINKED ISSUE #2472: Use gulp-typescript to compile our client code to es5
Several members of the team (@mbayopanda @jeremielodi ) has expressed interest in using TypeScript to ensure stricter typings.  In the long run, we likely want to evolve our codebase that way anyway, but it will take a few steps to get there.

As  first step, I propose that we use the typescript compiler to compile ES2015+ --> ES5 on the client.  I've made a lot of comments on PRs because programmers keep using ES2015+ features on the client and it breaks our builds at HEV (since `gulp-uglify` can't understand ES2015+).

I propose therefore that we add [gulp-typescript](https://www.npmjs.com/package/gulp-typescript) in to our workflow.  According to [this answer](https://stackoverflow.com/questions/12682268/is-it-possible-to-compile-typescript-into-minified-code/12682308#12682308), we will still need to keep `gulp-uglify` around to minify our codebase.

This proposal has the following advantages:
 1. Allows us to use ES2015+ on the client (here we come arrow functions!)
 2. Provides the way forward in case we want to use TypeScript in the future for either client or server code.
 3. Allows us to enforce ESLint rules across both client and server.

mbayopanda:
This PR allow to write our `.js` code in `ESNEXT` format and compile it into `ES5` via the `Typescript` compiler.

So it is now possible to write :  
```js
const greetings = (name) => {
 return `hello ${name}`;
}
```
instead of 
```js
function greetings (name) {
 return 'hello ' + name
}
```

close #2472

### DIFF
--- a/client/src/modules/cash/payments/registry.js
+++ b/client/src/modules/cash/payments/registry.js
@@ -16,7 +16,7 @@ CashPaymentRegistryController.$inject = [
  */
 function CashPaymentRegistryController(
   Cash, bhConstants, Notify, Session, uiGridConstants, Modal, Sorting, $state,
-  Filters, Columns, GridState, Modals, util,
+  Filters, Columns, GridState, Modals, util
 ) {
   var vm = this;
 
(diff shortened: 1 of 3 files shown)


jniles (approved):
Works for me!  Awesome!

bors r+

jniles:
@IMA-WorldHealth/bhima-core you will need to run `npm install` or `yarn` to get the latest dependencies to build the application.

PR #2474, opened 2018-01-26
chore(accounts): rename revenue -> income

### LINKED ISSUE #2465: Rename "revenue" account type to "income"
Technically, they mean the same thing.  But we refer to "income" accounts _way_ more than we refer to "revenue" accounts.  We should change the word "revenue" to "income".

jniles:
This commit simply renames "revenue" to the more commonly used "income".

Closes #2465.


jniles:
Silly tests.

bors r+

PR #2478, opened 2018-01-29
feat(bh-account-select) add account-type-id filter

### LINKED ISSUE #2364: Add an option for "account-type-filter" on bhAccountSelect

jeremielodi:
This PR add a new binding parameter to  bh-account-select component  for filtering accounts
closes #2364

### DIFF
--- a/client/src/js/components/bhAccountSelect.js
+++ b/client/src/js/components/bhAccountSelect.js
@@ -5,6 +5,7 @@ angular.module('bhima.components')
     transclude  : true,
     bindings    : {
       accountId        : '<',
+      accountTypeId :  '<',
       onSelectCallback : '&',
       disable          : '<?',
       required         : '<?',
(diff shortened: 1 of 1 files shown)


jniles (approved):
LGTM!  As of this PR, we will deprecate `classe` to see if we can make everything work with `type_id` instead.

PR #2485, opened 2018-01-31
fix stock lot registry sorting

### LINKED ISSUE #2479: Order "Lots in Stock" Registry by Lot Number
The Lots in Stock registry does not have an "ORDER BY" clause.  This means that it will print out in whatever order it was inserted it, which isn't very useful.  We should put an ORDER BY clause on this query.

Please add `ORDER BY inventory.code, lot.number` to the query somewhere.

mbayopanda:
This PR fix the stock lot registry : 
- By fixing the sorting columns
- By enhancing the report template

close #2479

### DIFF
--- a/client/src/modules/inventory/configuration/groups/modals/actions.modal.js
+++ b/client/src/modules/inventory/configuration/groups/modals/actions.modal.js
@@ -89,8 +89,8 @@ function InventoryGroupsActionsModalController(InventoryGroups, Notify, Instance
 
     if (vm.identifier) {
       InventoryGroups.read(vm.identifier)
-        .then(function (groups) {
-          vm.session = groups[0];
+        .then((groups) => {
+          vm.session = groups;
         })
         .catch(Notify.handleError);
     }
--- a/server/controllers/stock/reports/stock_lots.report.handlebars
+++ b/server/controllers/stock/reports/stock_lots.report.handlebars
@@ -41,8 +41,10 @@
           <tr>
             <th>{{translate 'STOCK.CODE'}}</th>
             <th>{{translate 'STOCK.INVENTORY'}}</th>
+            <th>{{translate 'STOCK.INVENTORY_GROUP'}}</th>
             <th>{{translate 'STOCK.LOT'}}</th>
             <th>{{translate 'STOCK.QUANTITY'}}</th>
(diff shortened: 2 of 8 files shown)


client/src/modules/inventory/configuration/groups/modals/actions.modal.js
```diff
@@ -89,8 +89,8 @@ function InventoryGroupsActionsModalController(InventoryGroups, Notify, Instance
 
     if (vm.identifier) {
       InventoryGroups.read(vm.identifier)
-        .then(function (groups) {
-          vm.session = groups[0];
+        .then((groups) => {
+          vm.session = groups;
```
jniles, line 93: Are you sure this doesn't break inventory?  What changed that it should be `groups` instead of `groups[0]` now?

server/controllers/stock/reports/stock_lots.report.handlebars
```diff
@@ -41,8 +41,10 @@
           <tr>
             <th>{{translate 'STOCK.CODE'}}</th>
             <th>{{translate 'STOCK.INVENTORY'}}</th>
+            <th>{{translate 'STOCK.INVENTORY_GROUP'}}</th>
```
jniles, line 44: 👍

server/controllers/stock/reports/stock_lots.report.handlebars
```diff
@@ -68,11 +70,18 @@
...
                 <td>{{code}}</td>
                 <td>{{text}}</td>
+                <td>{{group_name}}</td>
                 <td>{{label}}</td>
                 <td class="text-right">{{quantity}}</td>
+                <td class="text-right">{{unit_type}}</td>
                 <td class="text-right">{{date entry_date}}</td>
-                <td class="text-right">{{date expiration_date}}</td>
-                <td class="text-right">{{delay_expiration}}</td>
+                {{#if item.expires}}
```
jniles, line 78: Nice 👍

jniles (approved):
LGTM.

There's one question I have - if you are sure it doesn't produce a new bug, we can merge this.

client/src/modules/inventory/configuration/groups/modals/actions.modal.js
mbayopanda: Yes, there was a bug with `group[0]`

PR #2487, opened 2018-02-01
feat(installation) use password validation

### LINKED ISSUE #2468: Use the password validation component on the installation page
The installation page allows a user to put in a password for the first time installation.  Since we have a rather nice password validation directive (introduced in #2204), we should use it here to make sure the first password is a strong one.

jeremielodi:
closes #2468


jniles (approved):
LGTM.

bors r+

PR #2488, opened 2018-02-02
feat(fiscal Year) prevent a Fiscal year to be itself previous one

### LINKED ISSUE #2470: Fiscal Years can reference themselves as the previous fiscal year.
Each fiscal year should have a `previous_fiscal_year_id` set on creation.  However, when you edit the year, you can choose to self-reference, resulting in a year being its own previous fiscal year!  This is clearly not what we want.  See below an example:

bugcanchoosecurrentyearaspreviousyear
_Fig 1: Self-Referencing Fiscal Year_

The solution is to filter out the current year from the `<select>` element to avoid this kind of behavior.

jeremielodi:
This PR prevent fiscal years to be themselves previous year
close #2470


jniles (approved):
LGTM

bors r+

PR #2491, opened 2018-02-02
feat(installation) set the project name the enterprise one if not provided

### LINKED ISSUE #2475: Create a project for the enterprise by default on installation
Refer to the following screenshot:
projectduplicateenterprise
_Fig 1: Create an Enterprise and Project_

BHIMA is able to handle multiple projects for each enterprise, corresponding to multiple installation locations/servers.  Most setups will not need this functionality, since they will have a single server and keep track of a single project.  Therefore your enterprise will be the same as your project, and the above form is a duplication.

I propose that we hide the project configuration behind a `bhHiddenField` component.  By default, if a project's name and abbreviation are not provided, they will take the enterprise's configuration.  A user has the option to toggle the visibility of the project form and set a custom the name and abbreviation if they desire.

This should simplify the setup for most use cases.

jeremielodi:
This PR solves this problem #2475 
closes #2475

### DIFF
--- a/client/src/modules/install/install.html
+++ b/client/src/modules/install/install.html
@@ -43,6 +43,7 @@
   </div>
 
   <!-- project -->
+  <bh-hidden-field show-text='FORM.LABELS.PROJECT' hide-text ='FORM.BUTTONS.CLOSE'>
   <div class="panel panel-default">
     <div class="panel-heading">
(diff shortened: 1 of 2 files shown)


jniles:
LGTM.

bors r+

PR #2498, opened 2018-02-02
Add group in the inventory group registry

mbayopanda:
This PR add the group column in the stock inventory registry


jniles (approved):
LGTM.

bors r+

PR #2504, opened 2018-02-02
fix(inventory): misc ui improvements.

### LINKED ISSUE #2493: Inventory "Delete" Needs Icon
In the inventory list, the "delete" dropdown does not have an icon.  For consistency, let's put an icon.

See below:
inventorydeleteneedsicon
_Fig 1: This should have a trash can_

### LINKED ISSUE #2499: Make Inventory List ui-grid fill screen and no more
The Inventory List ui-grid overflows the page resulting in two scroll bars.  This is really annoying, since it hides the headers if you scroll to the bottom of the page.  We should be able to fix this with a simply change of CSS classes on the grid itself.

See below:
twoscrollbarsinventory
_Fig 1: What good are two scroll bars?!_

jniles:
This commit:
 1. Adds a red trash icon to the delete option in the inventory list.
 2. Moves the "delete" option to the last option in the inventory list.
 3. Makes the inventory ui-grid always fit on the page.

Closes #2499. Closes #2493.

### DIFF
--- a/client/src/modules/inventory/list/list.html
+++ b/client/src/modules/inventory/list/list.html
@@ -87,8 +87,7 @@
   <div class="container-fluid">
     <div
       id="inventoryListGrid"
-      class="grid-full-height"
-      ng-class="{'grid-full-height-with-filters': InventoryCtrl.hasCustomFilters}"
+      class="grid-full-height-with-filters"
       ui-grid="InventoryCtrl.gridOptions"
       ui-grid-save-state
       ui-grid-exporter
--- a/client/src/modules/inventory/list/templates/action.cell.html
+++ b/client/src/modules/inventory/list/templates/action.cell.html
@@ -8,26 +8,28 @@
   <ul class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
     <li>
       <a href
-          data-method="edit"
-          data-edit-metadata="{{ row.entity.code }}"
-          ng-hide="row.groupHeader" ui-sref="inventory.update({uuid : row.entity.uuid })">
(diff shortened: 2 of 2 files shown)


client/src/modules/inventory/list/list.html
```diff
@@ -87,8 +87,7 @@
   <div class="container-fluid">
     <div
       id="inventoryListGrid"
-      class="grid-full-height"
-      ng-class="{'grid-full-height-with-filters': InventoryCtrl.hasCustomFilters}"
+      class="grid-full-height-with-filters"
```
mbayopanda, line 90: 👍

client/src/modules/inventory/list/templates/action.cell.html
```diff
@@ -8,26 +8,28 @@
...
-          data-edit-metadata="{{ row.entity.code }}"
-          ng-hide="row.groupHeader" ui-sref="inventory.update({uuid : row.entity.uuid })">
-          <i class="fa fa-edit"></i> <span translate>FORM.BUTTONS.EDIT</span>
+        data-method="edit"
+        data-edit-metadata="{{ row.entity.code }}"
+        ui-sref="inventory.update({uuid : row.entity.uuid })">
+        <i class="fa fa-edit"></i> <span translate>FORM.BUTTONS.EDIT</span>
       </a>
     </li>
+    <li class="divider"></li>
```
mbayopanda, line 17: Having a `divider` here is for separating concepts ?

mbayopanda (approved):
LGTM

client/src/modules/inventory/list/templates/action.cell.html
jniles: Yup!  It also adds more space so that we don't accidentally click it when we try to click "edit".

PR #2505, opened 2018-02-02
feat(stock): allow moving lots registry columns

### LINKED ISSUE #2502: Implement move columns on stock lots registry
The stock lots registry could use column movement.  I'd like to re-position the "lot" column to the far left and save it in grid state.  However, I cannot.

jniles:
This commit implements the ui-grid plugin to move columns on the stock lots registry.

Closes #2502.


client/src/modules/stock/lots/registry.html
```diff
@@ -104,6 +104,7 @@
       ui-grid="StockLotsCtrl.gridOptions"
       class="grid-util-full-height"
       ui-grid-resize-columns
+      ui-grid-move-columns
```
mbayopanda, line 107: 👍

mbayopanda (approved):
LGTM

PR #2508, opened 2018-02-03
feat(stock entry) add default line in grid after type selection

### LINKED ISSUE #2497: By default, add a new line to the Stock Entry grid after the user selects entry type
In the Patient Invoice page, after a patient is selected, the ui-grid is populated with a single, empty row.  This is because the user will _definite_ try to add a line ... the module will not work without it.

We should do the same thing on the Stock Entry page.  As soon as the user has selected an entry type from the above values, they should immediately be able to start entering items without having to click "add an item".

jeremielodi:
This PR add a new line to the Stock Entry grid after the user selects entry type 
closes #2497

### DIFF
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -115,6 +115,9 @@ function StockEntryController(
     mapEntry[entryType.label].find();
 
     vm.entryOption = entryType && entryType.label !== 'purchase';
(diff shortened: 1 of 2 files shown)


jniles (approved):
LGTM.

Bors r+

jniles:
Looks like these tests need to be updated.

PR #2509, opened 2018-02-03
feact(stock entry) delete grid row

### LINKED ISSUE #2492: Stock Entry does not let you delete rows
There is a bug in stock entry (`#!/stock/entry`) that does not let a user delete rows that they just created with the "add a row" button.  No error is echoed to the client.

jeremielodi:
closes #2492


jniles:
There should be a way to do this using the store in one or two lines of code.  I'll try investigating.  It used to work at one time...

jniles:
So, the bug here was the call to `removeItem()` using `index` instead of `id`.  It was probably copy and pasted from somewhere and never updated.

PR #2514, opened 2018-02-05
feat(journal): add debit/credit source sums

### LINKED ISSUE #2512: Sum the debit (source) and credit (source) columns in the Journal
In the Journal, the debit/credit columns are summed and their values displayed both on the transaction headers and the footer of the grid.   We should also sum the debit (source) and credit (source) columns, but only render there values on the transaction header where it makes sense.

To do this, you will likely have to hide the grid footer column's value.  This is easy - just set the footer template to some empty HTML, like `<i></i>`.  Other than that, it should be the same as the regular debit/credit columns.

jniles:
This commit adds summation on the debit/credit source columns of the Journal and Account Statement.  It also makes both files pass ESLint.

fixedaggregategroupingjournal
_Fig 1: End Result on Journal_

Closes #2512.

### DIFF
--- a/client/src/js/services/FilterService.js
+++ b/client/src/js/services/FilterService.js
@@ -137,11 +137,11 @@ function FilterService() {
 
   FilterList.prototype.removeUnCachableFilters = function removeUnCachableFilters(filters) {
     filters.forEach(filter => {
-      if (filter.cachable === 0) {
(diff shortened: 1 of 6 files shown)


mbayopanda (approved):
👍 LGTM

jniles:
Looks like there are conflicts :).  I'll fix them and then `bors r+` myself.  Thanks for the review!

PR #2515, opened 2018-02-05
fix exchange rate bug

mbayopanda:
This PR fix a bug with variables of the exchange rate in procedures which had the type `DECIMAL(19, 4)` instead of `DECIMAL(19, 8)`.


jniles:
Beautiful.  Well done!

bors r+

PR #2518, opened 2018-02-07
fix(journal): Allow deletion of transaction in the journal

### LINKED ISSUE #2510: Allow deleting a transaction from the transaction edit modal
There needs to be a button on the "edit transaction" form that allows you to delete the transaction.  Since we support deletion through a [uniform API](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2286), it should be trivial to add this.

The basic design is this:
 1. Add a button `btn-danger` that says "delete transaction" to the TransactionEditModal.  It should probably be below the transaction's grid, but not on the footer.
 2. Call `TransactionService.remove(uuid)` with the `record_uuid` of the transaction.
 3. Close the modal when it is successfully removed.

### LINKED ISSUE #2511: If a transction is made in a currency, use that currency for the editing
Currently, we allow a user to edit a transactions `debit_equiv` and `credit_equiv` values directly in the Posting Journal.  This changes the _value_ of the transaction in the enterprise currency.  We then [convert those values on the server](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/journal/index.js#L451) into the correct debit/credit values.

As our clients in HEV noted, this is backwards.  If a user makes a mistake in a non-enterprise currency, they want to correct _that value_ and not the enterprise's currency.

For example, imagine that you are running an enterprise in FC and you have a bill for $50.  By accident, you type $500 in the voucher.  When you try to edit this, you will be editing the FC value of `exchange_rate * 500` instead of what you originally mis-typed of $500.   This makes it really confusing to fix.

I propose the following:
 1. If you made a transaction in a particular currency, it should be displayed for editing in that currency.  If the currency is not the enterprise currency, you should also show the exchange rate.
 2. Once the transaction is saved, you should convert it into the enterprise currency.

In this way, it is the application itself that does the exchange rate math, not the user.  That's just silly.

mbayopanda:
This PR : 
- Allow to delete a transaction in the journal via the edit transaction modal
- Edit transaction debit and credit, not the converted amounts

Close #2510 
Close #2511

### DIFF
--- a/client/src/js/components/bhExchangeRate.js
+++ b/client/src/js/components/bhExchangeRate.js
@@ -1,0 +1,72 @@
+/**
+ * @overview bhExchangeRate
+ *
+ * @description
+ * This component allows to display the exchange rate
+ */
+angular.module('bhima.components')
+  .component('bhExchangeRate', {
+    templateUrl : 'modules/templates/bhExchangeRate.tmpl.html',
+    controller : bhExchangeRateController,
+    controllerAs : '$ctrl',
+  });
+
+bhExchangeRateController.$inject = [
+  'CurrencyService', 'ExchangeRateService', 'SessionService',
+  'NotifyService', 'moment',
+];
+
+function bhExchangeRateController(Currencies, Rates, Session, Notify, Moment) {
+  var $ctrl = this;
+
+  $ctrl.today = new Date();
+  $ctrl.primaryExchange = {};
+  $ctrl.enterprise = Session.enterprise;
+  $ctrl.EXCHANGE_RATE_DISPLAY_SIZE = 6;
+
+  /*
+    $ctrl.isFirstCurencyLabel is used to check the exchange Rate
+    is lower then 1  the program display something
+    mutch better for reading
+  */
+  $ctrl.isFirstCurencyLabel = false;
+
+  // load exchange rates
+  function loadExchangeRates() {
+    Currencies.read(true)
+      .then((currencies) => {
+        $ctrl.currencies = currencies.filter((currency) => {
+          return currency.id !== Session.enterprise.currency_id;
+        });
+
+        // load supported rates
+        return Rates.read(true);
+      })
+      .then(() => {
+        $ctrl.currencies.forEach((currency) => {
+          currency.rate = Rates.getCurrentRate(currency.id);
+
+          /*
+            Let check if the currency rate is lower the 1
+            so that we could format it in a readable way
+          */
+          if (currency.rate < 1) {
+            currency.rate = (1 / currency.rate);
+            $ctrl.isFirstCurencyLabel = true;
+          }
+
+          currency.formattedDate = new Moment(currency.date).format('LL');
+        });
+
+        // get the first element of $ctrl.currencies ($ctrl.currencies[0])
+        [$ctrl.primaryExchange] = $ctrl.currencies;
+
+        console.log('currency : ', $ctrl.primaryExchange);
+      })
+      .catch(Notify.handleError);
+  }
+
+  $ctrl.$onInit = function onInit() {
+    loadExchangeRates();
+  };
+}
--- a/client/src/js/components/bhExchangeRate.js
+++ b/client/src/js/components/bhExchangeRate.js
@@ -54,14 +54,10 @@ function bhExchangeRateController(Currencies, Rates, Session, Notify, Moment) {
             currency.rate = (1 / currency.rate);
             $ctrl.isFirstCurencyLabel = true;
           }
-
-          currency.formattedDate = new Moment(currency.date).format('LL');
         });
 
         // get the first element of $ctrl.currencies ($ctrl.currencies[0])
         [$ctrl.primaryExchange] = $ctrl.currencies;
-
-        console.log('currency : ', $ctrl.primaryExchange);
       })
       .catch(Notify.handleError);
   }
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -531,7 +531,6 @@ function JournalController(
   //       updated to keep track of the current grid grouped state
   vm.gridGroupedState = vm.grouping.getCurrentGroupingColumn;
 
- 
   // runs on startup
   function startup() {
     var params = $state.params.filters;
@@ -551,8 +550,8 @@ function JournalController(
       return;
     }
 
-    var selectedTransaction = selection.selected.groups[0];
-    var transactionUuid = lookupIntermediateRecordUuid(selectedTransaction);
+    const selectedTransaction = selection.selected.groups[0];
+    const transactionUuid = lookupIntermediateRecordUuid(selectedTransaction);
 
     // Journal module rules for optimistic updating:
     // 1. If a row in the current dataset has been removed - remove this row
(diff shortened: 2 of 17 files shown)


client/src/js/components/bhExchangeRate.js
```diff
@@ -0,0 +1,68 @@
...
+ * @overview bhExchangeRate
+ *
+ * @description
+ * This component allows to display the exchange rate
+ */
+angular.module('bhima.components')
+  .component('bhExchangeRate', {
+    templateUrl : 'modules/templates/bhExchangeRate.tmpl.html',
+    controller : bhExchangeRateController,
+    controllerAs : '$ctrl',
```
jniles, line 11: Just FYI - I believe all components default to `$ctrl` for their `controllerAs`.

client/src/modules/journal/modals/editTransaction.modal.js
```diff
@@ -335,28 +356,49 @@ function JournalEditTransactionController(
...
+        const deleteTransactionResult = {
+          deleted : true,
+          removed : getGridRowsUuid(),
+        };
+        Modal.close(deleteTransactionResult);
+      });
+  };
+
+  function getGridRowsUuid() {
+    return vm.rows.data.map((row) => {
```
jniles, line 384: Alternatively: `vm.rows.data.map(row => row.uuid);`.

client/src/modules/journal/modals/editTransaction.modal.js
```diff
@@ -335,28 +356,49 @@ function JournalEditTransactionController(
...
+      return row.uuid;
+    });
+  }
+
   // rows - array of rows
   // uuids - array of uuids
   function filterRowsByUuid(rows, uuids) {
-    var result = rows.filter(function (row) {
+    const result = rows.filter((row) => {
```
jniles, line 392: Alternatively, `return rows.filter(row => contains(uuids, row.uuid));`.

client/src/modules/templates/bhExchangeRate.tmpl.html
```diff
@@ -0,0 +1,12 @@
+<span >
+  <span translate>FORM.LABELS.EXCHANGE_RATE</span> : 
+  <b ng-show="$ctrl.isFirstCurencyLabel">
+    1{{$ctrl.primaryExchange.symbol}} = {{ $ctrl.primaryExchange.rate | limitTo:$ctrl.EXCHANGE_RATE_DISPLAY_SIZE }}
+    ({{ $ctrl.enterprise.currencySymbol }})
+  </b>
+
+  <b ng-show="!$ctrl.isFirstCurencyLabel">
```
jniles, line 8: Currency should have two "r"s. :)

client/src/modules/templates/modals/confirmDialog.modal.html
```diff
@@ -1,20 +1,16 @@
 <form name="ConfirmDeletion" ng-submit="$ctrl.accept(ConfirmDeletion)" novalidate>
 
   <div class="modal-header">
-    <button ng-click="$ctrl.close()" type="button" class="close">&times;</button>
-    <h4>
-      {{ "FORM.DIALOGS.SURE_CONTINUE" | translate }}
```
jniles, line 6: Nice refactors!

client/src/modules/templates/modals/confirmDialog.modal.js
```diff
@@ -9,35 +9,33 @@ ConfirmDialogModalController.$inject = ['$uibModalInstance', '$translate', 'data
  * This controller is responsible for check a match text given to continue
  * with an action
  */
-function ConfirmDialogModalController (Instance, $translate, Data) {
-  var vm = this;
+function ConfirmDialogModalController(Instance, $translate, Data) {
+  const vm = this;
```
jniles, line 13: 👍

client/src/modules/templates/modals/confirmDialog.modal.js
```diff
@@ -9,35 +9,33 @@ ConfirmDialogModalController.$inject = ['$uibModalInstance', '$translate', 'data
...
     // make sure the modal isn't accidentally called with empty values
     if (!Data.pattern || !Data.patternName) {
-      throw new Error(
-        'ConfirmDialogModal requires both a pattern and an element name to be defined, but received:' +
-        'data.pattern = ' + Data.pattern + ' ' +
-        'data.patternName = ' + Data.patternName
-      );
+      throw new Error(`
+        ConfirmDialogModal requires both a pattern and an element name to be defined, but received: 
+        data.pattern = ${Data.pattern} and data.patternName = ${Data.patternName}
```
jniles, line 33: 💃

client/src/modules/templates/modals/confirmDialog.modal.js
```diff
@@ -55,13 +53,13 @@ function ConfirmDialogModalController (Instance, $translate, Data) {
...
 
-    var result = isMatching(vm.pattern, vm.text);
+    const result = isMatching(vm.pattern, vm.text);
 
     if (!form.$invalid && result) {
       Instance.close(result);
     } else if (!form.$invalid && !result) {
       vm.noCorrespondancy = $translate.instant('FORM.DIALOGS.NO_CORRESPONDANCY')
-        .concat(' "', vm.text, '"');
+        .concat(` "${vm.text}" `);
```
jniles, line 62: 👯‍♂️

server/controllers/finance/journal/index.js
```diff
@@ -469,6 +474,27 @@ function transformColumns(rows, newRecord, transactionToEdit, setFiscalData) {
       });
     }
 
+    if (row.debit) {
```
jniles, line 477: Nice!

client/src/js/components/bhExchangeRate.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+bhExchangeRateController.$inject = [
+  'CurrencyService', 'ExchangeRateService', 'SessionService',
+  'NotifyService', 'moment',
+];
+
+function bhExchangeRateController(Currencies, Rates, Session, Notify, Moment) {
+  var $ctrl = this;
```
jniles, line 20: This could be a `const` :)

client/src/js/components/bhExchangeRate.js
```diff
@@ -0,0 +1,68 @@
...
+
+  $ctrl.today = new Date();
+  $ctrl.primaryExchange = {};
+  $ctrl.enterprise = Session.enterprise;
+  $ctrl.EXCHANGE_RATE_DISPLAY_SIZE = 6;
+
+  /*
+    $ctrl.isFirstCurencyLabel is used to check the exchange Rate
```
jniles, line 28: Great comment!

client/src/js/components/bhExchangeRate.js
```diff
@@ -0,0 +1,68 @@
...
+        $ctrl.currencies = currencies.filter((currency) => {
+          return currency.id !== Session.enterprise.currency_id;
+        });
+
+        // load supported rates
+        return Rates.read(true);
+      })
+      .then(() => {
+        $ctrl.currencies.forEach((currency) => {
+          currency.rate = Rates.getCurrentRate(currency.id);
```
jniles, line 47: Is this the same code used on the home page?  If so, we should probably put it in a service ... something like `Exchange.getHumanReadableRate()`.

client/src/modules/journal/journal.js
```diff
@@ -551,50 +550,71 @@ function JournalController(
...
+      // rows have been added, we have no guarantees on filters so display a warning
+      vm.unknownTransactionEditState = true;
+    }
+    vm.gridApi.grid.notifyDataChange(uiGridConstants.dataChange.ALL);
+
+    Notify.success('POSTING_JOURNAL.SAVE_TRANSACTION_SUCCESS');
+  }
+
+  // Handle Delete Transaction Result
+  function handleDeleteTransactionResult(deleteSessionResult) {
```
jniles, line 607: Nice separation.  Very readable. 👍

client/src/modules/journal/modals/editTransaction.modal.html
```diff
@@ -109,13 +115,16 @@
...
 
 <div class="modal-footer clearfix">
-  <!-- informational: this record has been edited x times  -->
-  <span ng-if="ModalCtrl.hasPreviousEdits" class="text-muted pull-left" style="margin-top:5px;">
-    <i class="fa fa-info-circle"></i>
-    <span translate translate-values="ModalCtrl.lastEditValues">POSTING_JOURNAL.HAS_PREVIOUS_EDITS</span>
-  </span>
+  <button
+    class="btn btn-danger pull-left" 
+    ng-if="!ModalCtrl.readOnly"
```
jniles, line 120: Nice 👍

client/src/modules/journal/modals/editTransaction.modal.js
```diff
@@ -231,14 +252,14 @@ function JournalEditTransactionController(
 
 
   function verifyEditableTransaction(transaction) {
-    var posted = transaction[0].posted;
+    const { posted } = transaction[0];
```
jniles, line 255: 💃

jniles (approved):
Works really well for me!  Great work!

A couple of things for us to work on in the future:
 1. Simultaneously updating the `debit` and `debit_equiv` of a transaction doesn't make sense ... we should block it with an error on the server.  Same goes for `credit` and `credit_equiv`.
 2. If we use the same blocks of code for exchange rate, we should move it into a service.  That way we reuse the maximum code possible.

These are important changes.  The code looks really good!

bors r+

PR #2519, opened 2018-02-07
feat(report) operating account

### LINKED ISSUE #2425: Implement the MOH's Comptes d'Exploitation Report
We've been requested by HEV to implement the Ministry of Health's Operating Accounts Report ("Comptes d'Exploitation").  A sample version of it is attached to this issue as a XLSX.

The basic format is as follows:
 1. The expense accounts (represented in BHIMA as `account_type.label === 'ACCOUNT.TYPES.EXPENSE'`) are featured on the left side of the document, ordered by account number.
 2. The income/revenue accounts (represented in BHIMA as `account_type.label === 'ACCOUNT.TYPES.REVENUE'`) are featured on the right side of the document, ordered by account number.
 3. The titles of the income/expense accounts are featured on the report.  For a first implementation, this should be the direct parents of the income/expense accounts only.  Future versions of the report may allow determining the depth of the parent/child relationships.
 4. The accounts should include their `period_total` balance up to the current date.  Future reports may include a range of dates.
 5. The titles should contain the sum of all the child account balances.
 6. In addition to balances, a column will contain a percent breakdown of the total value within its group,  For expenses, the account's % is calculated as `(account.balance / total_expenses)* 100`.  A similar calculation is done for the income.  Note that the parent account should also have the same calculation, which will naturally be the sum of their child percentages.
 7.  The report should have the total columns for all expense and all income.  It should also have a "General Total" which takes the total expense and subtracts it from the total income to arrive at the profitability of the institution.

[COMPTE D'EXPLOITATION.xlsx](https://github.com/IMA-WorldHealth/bhima-2.X/files/1571173/COMPTE.D.EXPLOITATION.xlsx)

jeremielodi:
This PR Implement the MOH's Comptes d'Exploitation Report,
closes #2425

### DIFF
--- a/client/src/modules/reports/generate/operating/operating.config.js
+++ b/client/src/modules/reports/generate/operating/operating.config.js
@@ -1,0 +1,69 @@
+angular.module('bhima.controllers')
+  .controller('operatingController', OperatingConfigController);
+
+  OperatingConfigController.$inject = [
+  '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state',
+];
+
+
+function OperatingConfigController($sce, Notify, SavedReports, AppCache, reportData, $state) {
+  var vm = this;
+  var cache = new AppCache('configure_operating');
+  var reportUrl = 'reports/finance/operating';
+  vm.reportDetails = {};
+  vm.previewGenerated = false;
+  checkCachedConfiguration();
+  
+  vm.onSelectFiscal = function onSelectFiscal(fiscal) {
+    vm.reportDetails.fiscal = fiscal;
+  };
+
+  vm.onSelectPeriodFrom = function onSelectPeriodFrom(period) {
+    vm.reportDetails.periodFrom = period;
+  };
+
+  vm.onSelectPeriodTo = function onSelectPeriodTo(period) {
+    vm.reportDetails.periodTo = period;
+  };
+
+  vm.preview = function preview(form) {
+    if (form.$invalid) { return; }
+
+    // update cached configuration
+    cache.reportDetails = angular.copy(vm.reportDetails);
+
+    return SavedReports.requestPreview(reportUrl, reportData.id, angular.copy(vm.reportDetails))
+      .then(function (result) {
+        vm.previewGenerated = true;
+        vm.previewResult = $sce.trustAsHtml(result);
+      })
+      .catch(Notify.handleError);
+  };
+
+  vm.clearPreview = function clearPreview() {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.requestSaveAs = function requestSaveAs() {
+
+    var options = {
+      url : reportUrl,
+      report : reportData,
+      reportOptions : angular.copy(vm.reportDetails),
+    };
+
+    return SavedReports.saveAsModal(options)
+      .then(function () {
+        $state.go('reportsBase.reportsArchive', { key : options.report.report_key });
+      })
+      .catch(Notify.handleError);
+  };
+
+  function checkCachedConfiguration() {
+    if (cache.reportDetails) {
+      vm.reportDetails = angular.copy(cache.reportDetails);
+    }
+    vm.reportDetails.type = 1;
+  }
+}
--- a/client/src/modules/reports/generate/operating/operating.config.js
+++ b/client/src/modules/reports/generate/operating/operating.config.js
@@ -7,9 +7,9 @@ angular.module('bhima.controllers')
 
 
 function OperatingConfigController($sce, Notify, SavedReports, AppCache, reportData, $state) {
-  var vm = this;
-  var cache = new AppCache('configure_operating');
-  var reportUrl = 'reports/finance/operating';
+  let vm = this;
+  let cache = new AppCache('configure_operating');
+  const reportUrl = 'reports/finance/operating';
   vm.reportDetails = {};
   vm.previewGenerated = false;
   checkCachedConfiguration();
--- a/server/controllers/finance/fiscal.js
+++ b/server/controllers/finance/fiscal.js
@@ -33,7 +33,9 @@ exports.getPeriodByFiscal = getPeriodByFiscal;
 exports.lookupFiscalYearByDate = lookupFiscalYearByDate;
 exports.getFirstDateOfFirstFiscalYear = getFirstDateOfFirstFiscalYear;
 exports.getNumberOfFiscalYears = getNumberOfFiscalYears;
-
+exports.getDateRangeFromPeriods = getDateRangeFromPeriods;
+exports.getPeriodIdsFromDateRange = getPeriodIdsFromDateRange;
+exports.accountBanlanceByTypeId = accountBanlanceByTypeId;
 /**
  * @method lookupFiscalYear
  *
@@ -550,4 +552,46 @@ function getNumberOfFiscalYears(dateFrom, dateTo) {
   `;
 
   return db.one(sql, [dateFrom, dateTo]);
+}
+
+function getPeriodIdsFromDateRange(startDate, endDate) {
+  const sql = `SELECT id FROM period WHERE start_date>= ? AND end_date<= ?`;
+  return db.one(sql, [startDate, endDate]);
+}
+
+function getDateRangeFromPeriods(periods) {
+  const sql = `
+    SELECT
+      MIN(start_date) AS dateFrom, MAX(end_date) AS dateTo
+    FROM
+      period
+    WHERE
+      period.id IN (?, ?)`;
+
+  return db.one(sql, [periods.periodFrom, periods.periodTo]);
+}
+
+/**
+ * return a query for retrieving account'balance by type_id and periods
+ */
+function accountBanlanceByTypeId() {
+  return `
+    SELECT ac.id, ac.number, ac.label, ac.parent, IFNULL(s.amount, 0) AS amount, s.type_id
+    
+    FROM account as ac LEFT JOIN (
+    SELECT SUM(pt.credit - pt.debit) as amount, pt.account_id, act.id as type_id
+    FROM period_total as pt
+    JOIN account as a ON a.id = pt.account_id
+    JOIN account_type as act ON act.id = a.type_id
+    JOIN period as p ON  p.id = pt.period_id
+    JOIN fiscal_year as fy ON fy.id = p.fiscal_year_id
+    WHERE fy.id = ? AND 
+      pt.period_id IN (
+        SELECT id FROM period WHERE start_date>= ? AND end_date<= ?   
+      )
+      AND act.id = ?
+    GROUP BY pt.account_id 
+    )s ON ac.id = s.account_id
+    ORDER BY ac.number
+  `;
 }
--- a/server/controllers/finance/reports/operating/index.js
+++ b/server/controllers/finance/reports/operating/index.js
@@ -1,0 +1,120 @@
+const q = require('q');
+const _ = require('lodash');
+const db = require('../../../../lib/db');
+const util = require('../../../../lib/util');
+const ReportManager = require('../../../../lib/ReportManager');
+
+const TEMPLATE = './server/controllers/finance/reports/operating/report.handlebars';
+
+exports.document = document;
+
+function document(req, res, next) {
+
+  const params = req.query;
+
+  let docReport;
+  const options = _.extend(req.query, {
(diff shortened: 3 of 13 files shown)


jniles:
Looks like this has some conflicts, can you try to resolve them?

jeremielodi:
@jniles, Can you give me a review of this PR because today I plan to work on HGR Financial Report, it's fairly the same with this one

client/src/modules/reports/generate/operating/operating.config.js
```diff
@@ -0,0 +1,69 @@
+angular.module('bhima.controllers')
+  .controller('operatingController', OperatingConfigController);
+
+  OperatingConfigController.$inject = [
+  '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state',
+];
+
+
+function OperatingConfigController($sce, Notify, SavedReports, AppCache, reportData, $state) {
+  var vm = this;
```
jniles: Since we have `gulp-typescript`, you can go ahead and start using `let`/`const` everywhere.  ESLint will let you know which one is most appropriate.

server/controllers/finance/reports/operating/index.js
```diff
@@ -0,0 +1,137 @@
...
+    const totalRevenue = `SELECT SUM(r.amount) as total FROM (${getQuery(typeRevenue, id, range.dateFrom, range.dateTo)}) as r`;
+    const totalExpense = `SELECT SUM(r.amount) as total FROM (${getQuery(typeExpense, id, range.dateFrom, range.dateTo)}) as r`;
+
+    queries.push(db.exec(getQuery(typeExpense, id, range.dateFrom, range.dateTo)));
+    queries.push(db.exec(getQuery(typeRevenue, id, range.dateFrom, range.dateTo)));
+    queries.push(db.one(totalExpense));
+    queries.push(db.one(totalRevenue));
+
+    q.all(queries)
+      .then(results => {
```
jniles: You could avoid one level of indentation by using `return q.all(...)` here.  Like this:
```js
  return q.all(queries);
})
.then(results => {
```
It would have the same effect, I think.

jniles: Another cool trick with promises is the [q.spread()](https://github.com/kriskowal/q/wiki/API-Reference#promisespreadonfulfilled-onrejected) method.  It makes it really easy to name parameters clearly, and might apply here:
```js
  // some code
  return q.all(queries);
})
.spread((expense, revenue, totalExpense, totalRevenue) => {
  const context = {
    expense,
    revenue,
    totalExpense : totalExpense.total,
    totalRevenue: totalRevenue.total,
    // etc ...
  }
```
The method basically takes all the values of the `Promise.all()` call and "spreads" them across the next function's parameters.  The the first index of the `Promise.all()` array is the first parameter of the `.spread()` function, the second index is the second parameter., etc ...

server/controllers/finance/reports/operating/index.js
```diff
@@ -0,0 +1,137 @@
...
+          dateFrom : range.dateFrom,
+          dateTo : range.dateTo,
+        };
+
+        let revenueTree = new Tree(context.revenue);
+        revenueTree.filterByLeaf('translation_key', typeRevenue);
+        revenueTree.sumOnProperty('amount');
+        context.revenue = revenueTree.toArray();
+
+        let expenseTree = new Tree(context.expense);
```
jniles: Nice use of the tree!  I'm glad you got it working :)

server/controllers/finance/reports/operating/index.js
```diff
@@ -0,0 +1,137 @@
...
+        expenseTree.sumOnProperty('amount');
+        context.expense = expenseTree.toArray();
+       
+
+        formatData(context.expense, context.totalExpense);
+        formatData(context.revenue, context.totalRevenue);
+
+        // diff is the result in the report
+        const diff = util.roundDecimal(context.totalRevenue - context.totalExpense);
+        const isExpenseHigher = context.totalExpense < context.totalRevenue;
```
jniles: Nice variable name!

server/controllers/finance/reports/operating/index.js
```diff
@@ -0,0 +1,137 @@
...
+    return;
+  }
+
+
+  const queries = [];
+  // fiscal year ID
+  const id = params.fiscal;
+
+  getDateRange(params).then(range => {
+    const typeExpense = `ACCOUNT.TYPES.EXPENSE`;
```
jniles: Instead of using the translation key, go ahead and use the `type_id`.  We consider those constants.  A mapping is already fixed [on the client](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/constants/bhConstants.js#L13) and [in the database](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/bhima.sql#L92).

Like this:
```js
const EXPENSE_ACCOUNT_TYPE = 5;
const INCOME_ACCOUNT_TYPE = 4;
```
It will also make your SQL and JS comparisons a little more efficient (comparing numbers is O(1), strings are higher).

server/controllers/finance/reports/operating/index.js
```diff
@@ -0,0 +1,137 @@
...
+    FROM account as ac LEFT JOIN (
+    SELECT SUM(pt.credit - pt.debit) as amount, pt.account_id, act.translation_key
+    FROM period_total as pt
+    JOIN account as a ON a.id = pt.account_id
+    JOIN account_type as act ON act.id = a.type_id
+    JOIN period as p ON  p.id = pt.period_id
+    JOIN fiscal_year as fy ON fy.id = p.fiscal_year_id
+    WHERE fy.id = ${fiscalId} AND 
+      pt.period_id IN (
+        SELECT id FROM period WHERE start_date>= ${db.escape(startDate)} AND end_date<=${db.escape(endDate)}
```
jniles: I think it should be pretty easy to make this use MySQL's escaping automatically.  If you return the query like this:
```js
  return `
    SELECT ac.id, ac.number, ac.label, ac.parent, IFNULL(s.amount, 0) AS amount, s.translation_key    
   FROM account as ac LEFT JOIN (
    SELECT SUM(pt.credit - pt.debit) as amount, pt.account_id, act.translation_key
    FROM period_total as pt
    JOIN account as a ON a.id = pt.account_id
    JOIN account_type as act ON act.id = a.type_id
    JOIN period as p ON  p.id = pt.period_id
    JOIN fiscal_year as fy ON fy.id = p.fiscal_year_id
    WHERE fy.id = ? AND 
      pt.period_id IN (
         SELECT id FROM period WHERE start_date >= DATE(?) AND end_date <= DATE(?)   
      )
      AND act.translation_key = ?
    GROUP BY pt.account_id 
    )s ON ac.id = s.account_id
    ORDER BY ac.number
  `;
```
and then in calling function:
```js
 queries.push(db.exec(getQuery(),[typeExpense, id, range.dateFrom, range.dateTo]));
```

This makes three things easier:
 1. It is easier to copy and paste the code into MySQL to test it (you just have to change the `?` to real values).
 2. It makes the SQL a bit easier to read.
 3. Most importantly, if you want to change the SQL in the future, you don't need change the function's signature.  The only thing you change is the SQL itself and the array of parameters.

There is only one thing you must be careful for: that `range.dateFrom` and `range.dateTo` are actual JS dates.  So you may need to do: `const dateFrom = new Date(range.dateFrom);` and use that instead.

server/controllers/finance/reports/operating/index.js
```diff
@@ -0,0 +1,137 @@
...
+      )
+      AND act.translation_key = ${db.escape(type)}
+    GROUP BY pt.account_id 
+    )s ON ac.id = s.account_id
+    ORDER BY ac.number
+  `;
+}
+
+
+function getDateRange(params) {
```
jniles: This looks like a good function to use in multiple places.  It's not necessary for this PR, but we might want to move it to `fiscal.js` and call it `getPeriodIdsFromDateRange(startDate, endDate)`.

server/controllers/finance/reports/operating/report.handlebars
```diff
@@ -0,0 +1,138 @@
...
+                  <th> {{ translate 'FORM.LABELS.AMOUNT'}}</th>
+                  <th>%</th>
+                </thead>
+                <tbody>
+                  {{#each this.expense}}
+                  <tr>
+                    <td class="text-right">{{number}}</td>
+                    <td>
+                      {{#if title}}
+                        <center><b>{{label}}</b></center>
```
jniles: [The `<center>` element is obsolete](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center).  You can accomplish the same thing by doing this:
```handlebars
<td {{#if title}}style="font-weight:bold; text-align:center;"{{/if}}>
  {{label}}
</td>
```

server/lib/Tree.js
```diff
@@ -133,6 +133,46 @@ class Tree {
...
+  setPropertyToParent(node, prop, value) {
+    node[prop] = value;
+    if (node.parentNode) {
+      this.setPropertyToParent(node.parentNode, prop, value);
+    }
+  }
+
+  // walk arround the tree
+  // search the node by property value
+  interate(node, prop, value, parent) {
```
jniles: Nice 👍

server/lib/util.js
```diff
@@ -179,3 +181,14 @@ function unlinkp(file) {
...
   return deferred.promise;
 }
+
+/**
+ * arround a decima
+ * @param {*} number 
+ * @param {*} precision 
+ */
+function roundDecimal(number, precision) {
+  var _precision = precision || 4;
```
jniles: This is a cool function!

Just FYI, with new JS functions, you can set default parameters like this:
```js
function roundDecimal(number, precision = 4) {
  const base = Math.pow(10, precision);
  return Math.round(number * base) / base;
}
```

jniles (commented):
@jeremielodi this is a really good looking PR!  Well done!

I've made a few suggestions to make the code easier to read.  The logic looks sound.  Your extension of Tree.js is really cool!

The introduction of `util.roundDecimal()` is great idea.  It will help us make sure our reports are uniform and their presentation clean.

The major points in this are:
 1. It's better to use the account_type_id instead of the translation key.  Even though the translation key is human readable, we don't expect the IDs to ever change.  If you put them as constants at the top of your controller, they will be easy to find and modify in the future if we ever need to.  [This is a good example from the Balance Sheet report](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/reports/balance_sheet/index.js#L23).
 2. Using `db.exec(sql, [...params])` will help in the long run instead of doing manual parameter parsing.

In general, this is a really nice pull request. 👍

server/controllers/finance/reports/operating/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+  const periods = {
+    periodFrom : params.periodFrom,
+    periodTo : params.periodTo,
+  };
+  
+  fiscal.getDateRangeFromPeriods(periods).then(dateRange => {
+    range = dateRange;
+
+    const totalIncome = `SELECT SUM(r.amount) as total FROM (${getQuery()}) as r`;
```
jniles, line 48: You could actually use the same template here if you wanted.  But this works as well.

server/controllers/finance/reports/operating/index.js
```diff
@@ -0,0 +1,124 @@
...
+    ];
+
+    queries.push(db.exec(getQuery(), expenseParams));
+    queries.push(db.exec(getQuery(), incomeParams));
+    queries.push(db.one(totalExpense, expenseParams));
+    queries.push(db.one(totalIncome, incomeParams));
+
+    return q.all(queries);
+  })
+    .spread((expense, revenue, totalExpense, totalIncome) => {
```
jniles, line 71: Nice :+1:

server/controllers/finance/fiscal.js
```diff
@@ -551,3 +553,45 @@ function getNumberOfFiscalYears(dateFrom, dateTo) {
...
+
+  return db.one(sql, [periods.periodFrom, periods.periodTo]);
+}
+
+/**
+ * return a query for retrieving account'balance by type_id and periods
+ */
+function accountBanlanceByTypeId() {
+  return `
+    SELECT ac.id, ac.number, ac.label, ac.parent, IFNULL(s.amount, 0) AS amount, s.type_id
```
jniles, line 579: Nice :+1:

jniles (approved):
Looks really good!  We'll have to run it by AG Neron to review the functionality, but the code is good enough to merge.

bors r+

PR #2520, opened 2018-02-07
Payroll account configuration

lomamech:
- Implement client side for Payroll Accounting Configuration
- Implement API for Payroll Accounting Configuration
- Implement integration test for Payroll Account Configuration
- Implement E2E test for Payroll Account Configuration

reference #2292

### DIFF
--- a/client/src/i18n/en/tree.json
+++ b/client/src/i18n/en/tree.json
@@ -2,6 +2,7 @@
   "TREE": {
     "ACCOUNT": "Account Management",
     "ACCOUNT_STATEMENT": "Account Statement",
+    "ACCOUNT_CONFIGURATION":"Account Configuration",
     "ADMIN": "Admin",
     "AGED_CREDITORS": "Aged Creditors",
     "BALANCE": "Balance",
--- a/client/src/i18n/en/tree.json
+++ b/client/src/i18n/en/tree.json
@@ -2,7 +2,6 @@
   "TREE": {
     "ACCOUNT": "Account Management",
     "ACCOUNT_STATEMENT": "Account Statement",
-    "ACCOUNT_CONFIGURATION":"Account Configuration",
     "ADMIN": "Admin",
     "AGED_CREDITORS": "Aged Creditors",
     "BALANCE": "Balance",
@@ -60,6 +59,7 @@
     "PATIENT_REGISTRATIONS":"Patient Registrations",
     "PATIENT_REGISTRY": "Patients Registry",
     "PAYROLL":"Payroll",
+    "PAYROLL_ACCOUNT_CONFIGURATION":"Account Configuration",
     "PAYROLL_RUB_MANAGEMENT" : "Payroll Rubric Management",    
     "PHARMACY": "Pharmacy Management",
     "POSTING_JOURNAL": "Journal",
--- a/client/src/modules/payroll/account_configuration/configuration.js
+++ b/client/src/modules/payroll/account_configuration/configuration.js
@@ -1,0 +1,97 @@
+angular.module('bhima.controllers')
+.controller('ConfigurationAccountController', ConfigurationAccountController);
+
+ConfigurationAccountController.$inject = [
+  'ConfigurationAccountService', 'ModalService',
+  'NotifyService', 'uiGridConstants', '$state', 'SessionService',
+];
+
+/**
+ * Account Management Controller
+ *
+ * This controller is about the Account management module in the admin zone
+ * It's responsible for creating, editing and updating a Account
+ */
+function ConfigurationAccountController(Configs, ModalService,
+  Notify, uiGridConstants, $state, Session) {
+  var vm = this;
+
+  // bind methods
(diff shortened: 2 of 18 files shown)


client/src/modules/payroll/account_configuration/configuration.service.js
```diff
@@ -0,0 +1,17 @@
+angular.module('bhima.services')
+  .service('ConfigurationAccountService', ConfigurationAccountService);
+
+ConfigurationAccountService.$inject = ['PrototypeApiService', '$uibModal'];
+
+/**
+ * @class RubricService
```
jniles: These comments need to be updated.

client/src/modules/payroll/account_configuration/modals/account.modal.js
```diff
@@ -0,0 +1,59 @@
+angular.module('bhima.controllers')
+  .controller('AccountConfigModalController', AccountConfigModalController);
+
+AccountConfigModalController.$inject = [
+  '$state', 'ConfigurationAccountService', 'ModalService', 'NotifyService', 'appcache',
+];
+
+function AccountConfigModalController($state, Config, ModalService, Notify, AppCache) {
```
jniles: It doesn't look like you use `ModalService` here, could you remove it?

client/src/modules/payroll/account_configuration/templates/account.tmpl.html
```diff
@@ -0,0 +1,3 @@
+<div class="ui-grid-cell-contents text-left">
```
jniles: The ui-grid templates are usually pretty heavy to load.  Could you refactor this to simply render a value?  It would be much more performant.   Like this:
```js
.then(rows => {
  rows.forEach(row => row.hrAccountText = `[${row.account_number}] ${row.account_label}`);
  // etc...
})


// In your column configuration, you can now just use the `hrAccountText` column!
```

client/src/modules/payroll/account_configuration/templates/action.tmpl.html
```diff
@@ -0,0 +1,22 @@
+<div class="ui-grid-cell-contents text-center" uib-dropdown dropdown-append-to-body>
+  <a uib-dropdown-toggle href>
+    <span data-method="action" translate>FORM.BUTTONS.ACTIONS</span>
+    <span class="caret"></span>
+  </a>
+
+  <ul data-action="{{rowRenderIndex}}"  class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
```
jniles, line 7: Nice autodropup 👍

server/controllers/payroll/accounts/index.js
```diff
@@ -0,0 +1,112 @@
+/**
+* Account Controller
+*
+* This controller exposes an API to the client for reading and writing Account
+*/
+
+var db = require('../../../lib/db');
```
jniles: For these variables, it is much better to use `const`.

client/src/i18n/en/tree.json
```diff
@@ -2,6 +2,7 @@
   "TREE": {
     "ACCOUNT": "Account Management",
     "ACCOUNT_STATEMENT": "Account Statement",
+    "ACCOUNT_CONFIGURATION":"Account Configuration",
```
jniles: Could you call the key `PAYROLL_ACCOUNT_CONFIGURATION`?  We might get confused about what it is for otherwise...

client/src/modules/payroll/account_configuration/configuration.js
```diff
@@ -0,0 +1,97 @@
...
+    ];
+
+  // options for the UI grid
+  vm.gridOptions = {
+    appScopeProvider  : vm,
+    enableColumnMenus : false,
+    fastWatch         : true,
+    flatEntityAccess  : true,
+    enableSorting     : true,
+    onRegisterApi     : onRegisterApiFn,
```
jniles, line 51: 👍

jniles (commented):
Looks pretty good to me!  A few suggestions to increase clarity and performance.

Since we now compile a lot of our code to ES2015+, please use the new JS methods as much as possible.  Things like arrow functions and `let`/`const` should be used wherever you have an anonymous function or use a `var`.

client/src/modules/payroll/account_configuration/configuration.js
```diff
@@ -0,0 +1,96 @@
...
+  vm.currencySymbol = Session.enterprise.currencySymbol;
+
+  // global variables
+  vm.gridApi = {};
+  vm.filterEnabled = false;
+
+  var gridColumn =
+    [
+      { field : 'label', displayName : 'FORM.LABELS.DESIGNATION', headerCellFilter : 'translate' },
+      { field : 'hrAccountText', displayName : 'FORM.LABELS.ACCOUNT', headerCellFilter : 'translate' },
```
jniles, line 33: Nice!

jniles (approved):
Looks good to me.

bors r+

PR #2525, opened 2018-02-08
feat(bhPeriodSelect): preserve custom dates

### LINKED ISSUE #2516: The custom date selection should default to the last dates the user put in
Currently, the date selection does not remember the last values that it undertook.  See below:

periodselectioncustomizeddoesnotrememberpreviousvalues
_Fig 1: These two should have the same dates_

This is a really annoying user experience.  We should fix this as soon as possible.

jniles:
This commit ensures that the bhPeriodSelect remembers custom dates if they are set in the same session.

Closes #2516.

### DIFF
--- a/client/src/js/components/bhPeriodSelect.js
+++ b/client/src/js/components/bhPeriodSelect.js
@@ -1,18 +1,18 @@
 angular.module('bhima.components')
-.component('bhPeriodSelect', {
-  bindings : {
-    defaultPeriod : '@',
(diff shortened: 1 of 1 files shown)


jniles:
Merging this before the Vanga upgrade this evening.  It should be fairly uncontroversial.

bors r+

jniles:
Looks like it was a fluke.  These asynchronous redis tests seem to be having issues...

bors r+

PR #2527, opened 2018-02-08
fix(stock): better error message in lot modal

### LINKED ISSUE #2494: Confusing Stock Entry Error Messages
May it is just the English translation, but I do not understand the errors below:
confusingstockerrormessage
_Fig 1: Errors on Stock Entry Modal_

My guess is that they could be some of the following:
 1. Every row must have a quantity
 2. Every row must have an expiration date in the future
 3. Etc ...

These should extremely clear to the user how to fix any mistakes they make.  Currently, it's not at all clear to me what I should do.

### LINKED ISSUE #2495: Use the previous quantity to populate the next quantity in Stock Entry of lots
On the lots modal, the default value is "0" for the next line whenever a line is added.  I propose we either:
 1. Default to 1, which is valid.
 2. Default to the previous line's value for the quantity.

It is extremely frustrating to put in singular items into the stock entry modal since you have to set the quantity 1 every single time.  And the error message is persistent, making you feel guilty every time you ignore it.

We can solve this with the above proposals.  Since you need a lot number anyway, it isn't likely that the user will submit an extra row by accident.

### LINKED ISSUE #2496: Do not show stock error messages until after the user clicks submit
If I open a blank stock entry form, it starts of angry at me.  Without any input, it informs me that I've already committed an error.  See below:
stockentryerrorswithoutinput
_Fig 1: What did I do wrong?_

This is a poor user experience.  Errors are meant to help the user do the right thing - in this case, I haven't done anything at all.  It would be better to show the error message after the user clicks "submit" if there are problems with the form.

mbayopanda:
This PR : 
- Set the default quantity to 1 for stock lots in the modal
- Define a better message for errors on lots entry
- Show error message only when click on submit
- Remove data binding with quantity in the modal and in the grid

Close #2494 
Close #2495 
Close #2496

### DIFF
--- a/client/src/modules/stock/entry/modals/lots.modal.html
+++ b/client/src/modules/stock/entry/modals/lots.modal.html
@@ -25,7 +25,6 @@
         type="number" 
         ng-model="$ctrl.stockLine.quantity" 
         ng-model-options="{ 'debounce' : { 'default' : 150, 'blur' : 0 }}"
-        ng-change="$ctrl.checkAll()"
         min="0">
     </div>
 
@@ -74,21 +73,14 @@
       </div>
     </div>
 
-    <div ng-if="$ctrl.hasInvalidEntries">
-      <div ng-if="$ctrl.hasNoLot" class="alert alert-warning">
-        <i class="fa fa-warning"></i> 
-        <span translate>FORM.WARNINGS.NO_LOT</span>
-      </div>
-
-      <div ng-if="$ctrl.isSomeLineIncorrect" class="alert alert-danger">
-        <i class="fa fa-warning"></i> 
-        <span translate>FORM.ERRORS.STOCK_LINE</span>
-      </div>
-
-      <div ng-if="$ctrl.isQuantityIncorrect" class="alert alert-danger">
-        <i class="fa fa-warning"></i> 
-        <span translate>FORM.ERRORS.LOT_QUANTITY</span>
-      </div>
+    <div ng-if="$ctrl.hasInvalidEntries" class="alert alert-danger clearfix" style="margin-top: 10px;">
+      <ul>
+        <li ng-if="$ctrl.hasMissingLotIdentifier" translate>STOCK.ERRORS.MISSING_LOT_NAME</li>
+        <li ng-if="$ctrl.hasInvalidLotQuantity" translate>STOCK.ERRORS.INVALID_LOT_QUANTITY</li>
+        <li ng-if="$ctrl.hasInvalidLotExpiration" translate>STOCK.ERRORS.INVALID_LOT_EXPIRATION</li>
+        <li ng-if="$ctrl.isQuantityIncorrect" translate>STOCK.ERRORS.INVALID_GLOBAL_QUANTITY</li>
+        <li ng-if="$ctrl.isSomeLineIncorrect" translate>STOCK.ERRORS.LOT_ROW_WITH_ERRORS</li>
+      </ul>
     </div>
 
   </div>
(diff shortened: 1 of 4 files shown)


client/src/modules/stock/entry/modals/lots.modal.html
```diff
@@ -74,21 +73,14 @@
...
-      <div ng-if="$ctrl.isSomeLineIncorrect" class="alert alert-danger">
-        <i class="fa fa-warning"></i> 
-        <span translate>FORM.ERRORS.STOCK_LINE</span>
-      </div>
-
-      <div ng-if="$ctrl.isQuantityIncorrect" class="alert alert-danger">
-        <i class="fa fa-warning"></i> 
-        <span translate>FORM.ERRORS.LOT_QUANTITY</span>
-      </div>
+    <div ng-if="$ctrl.hasInvalidEntries" class="alert alert-danger clearfix" style="margin-top: 10px;">
```
jniles, line 76: Do you need `clearfix` here as a class?

jniles: Also, 👍 combining all the alerts into a single alert!

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -7,10 +7,15 @@ StockDefineLotsModalController.$inject = [
 ];
 
 function StockDefineLotsModalController(Instance, Notify, uiGridConstants, Data, Session, bhConstants) {
-  var vm = this;
-  var current = new Date();
+  const vm = this;
```
jniles, line 10: Nice :)

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -7,10 +7,15 @@ StockDefineLotsModalController.$inject = [
...
+  const current = new Date();
+
+  vm.hasMissingLotIdentifier = false;
+  vm.hasInvalidLotExpiration = false;
+  vm.hasInvalidLotQuantity = false;
+
   vm.enterprise = Session.enterprise;
-  vm.stockLine = Data.stockLine;
+  vm.stockLine = angular.copy(Data.stockLine);
```
jniles, line 18: :dancer:

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -69,7 +74,7 @@ function StockDefineLotsModalController(Instance, Notify, uiGridConstants, Data,
       },
     ],
     data : vm.stockLine.lots,
-    onRegisterApi : onRegisterApi,
+    onRegisterApi,
```
jniles, line 77: 👍

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -111,11 +106,15 @@ function StockDefineLotsModalController(Instance, Notify, uiGridConstants, Data,
   function checkAll() {
     vm.hasNoLot = !vm.stockLine.lots.length > 0;
 
-    vm.isSomeLineIncorrect = vm.stockLine.lots.some(function check(lot) {
-      return lot.isValid === false;
-    });
+    vm.isSomeLineIncorrect = vm.stockLine.lots.some(lot => lot.isValid === false);
+
+    vm.hasMissingLotIdentifier = vm.stockLine.lots.some(lot => lot.hasMissingLotIdentifier === true);
```
jniles, line 111: 👍

jniles (approved):
LGTM! 

bors r+

PR #2529, opened 2018-02-08
Notify errors on delete transaction

mbayopanda:
This PR allow to see errors thrown when we delete a transaction.


jniles (approved):
Good catch!

bors r+

Issue #2533, opened 2018-02-09
feat(stock): Count items in stock entry/exit grid


mbayopanda:
This issue is about to know how many items are defined in the grid for a stock entry/exit.

A proposal is to have a label in the bottom of the grid which display the number of items

jniles:
@mbayopanda, should this be closed?  It looks like #2539 might have fixed it.

jniles:
Closed in #2539.

PR #2539, opened 2018-02-13
export stock exit grid and enhancements

mbayopanda:
In this PR : 
- We can export the content of the grid into a csv file
- We can count rows of the grid
- We apply some in style sheet files for responsive behavior and color

This PR is relate to #2533

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -23,7 +23,12 @@
 /**
  * IMA colors
  */
-.bg-ima-blue { background-color : #085484; color: #fff; }
+:root {
+  --ima-blue : #085484;
+}
+
+.ima-blue { color: var(--ima-blue); }
+.bg-ima-blue { background-color : var(--ima-blue); color: #fff; }
 
 /*
  * Sidebar
@@ -757,9 +762,9 @@ growl-notification.fading.ng-leave.ng-leave-active {
 }
 
 .notification-info {
-  color : #085484;
+  color : var(--ima-blue);
   background-color : #64BDEE;
-  border-left : 5px solid #085484;
+  border-left : 5px solid var(--ima-blue);
 }
 
 .notification-warn {
@@ -870,7 +875,7 @@ growl-notification.fading.ng-leave.ng-leave-active {
 }
 
 .progress-bar.ima-blue {
-  background-color : #085484;
+  background-color : var(--ima-blue);
 }
 
 /* properly deal with ui-selects inside ui-grids */
@@ -929,3 +934,50 @@ a[disabled] {
   z-index : 9999;
 }
 
+.show-on-small-screen {
+  display: none;
+}
+
+@media (max-width: 500px) {
+
+  /*---------------------------
+      show only on small screen
+  -----------------------------*/
+
+  .show-on-small-screen {
+    display: block;
+    color: white;
+    background-color: var(--ima-blue);
+    padding: 10px;
+  }
+}
+
+/* -------------------------------------
+    Grid Toolbar
+--------------------------------------- */
+.grid-toolbar {
+  display: flex;
+  align-items: flex-start;
+}
+
+.grid-toolbar-item {
+  display: inline-flex;
+}
+
+/* -------------------------------------
+    Segment Statistic Card
+--------------------------------------- */
+.ima-stat-card .ui.statistic > .value {
+  color: var(--ima-blue);
+}
+.ima-stat-card .ui.statistic > .ui-hidable-label {
+  color: black;
+}
+
+.ima-stat-card-reversed {
+  background-color: var(--ima-blue);
+}
+.ima-stat-card-reversed .ui.statistic > .value,
+.ima-stat-card-reversed .ui.statistic > .ui-hidable-label {
+  color: white;
+}
--- a/client/src/js/services/StockService.js
+++ b/client/src/js/services/StockService.js
@@ -43,10 +43,18 @@ function StockService(Api, Filters, AppCache, Periods, $httpParamSerializer, Lan
     { key : 'inventory_uuid', label : 'STOCK.INVENTORY' },
     { key : 'group_uuid', label : 'STOCK.INVENTORY_GROUP' },
     { key : 'label', label : 'STOCK.LOT' },
-    { key : 'entry_date_from', label : 'STOCK.ENTRY_DATE', comparitor : '>', valueFilter : 'date' },
-    { key : 'entry_date_to', label : 'STOCK.ENTRY_DATE', comparitor : '<', valueFilter : 'date' },
-    { key : 'expiration_date_from', label : 'STOCK.EXPIRATION_DATE', comparitor : '>', valueFilter : 'date' },
-    { key : 'expiration_date_to', label : 'STOCK.EXPIRATION_DATE', comparitor : '<', valueFilter : 'date' },
+    {
+      key : 'entry_date_from', label : 'STOCK.ENTRY_DATE', comparitor : '>', valueFilter : 'date',
+    },
+    {
+      key : 'entry_date_to', label : 'STOCK.ENTRY_DATE', comparitor : '<', valueFilter : 'date',
+    },
+    {
+      key : 'expiration_date_from', label : 'STOCK.EXPIRATION_DATE', comparitor : '>', valueFilter : 'date',
+    },
+    {
+      key : 'expiration_date_to', label : 'STOCK.EXPIRATION_DATE', comparitor : '<', valueFilter : 'date',
+    },
   ]);
 
   StockMovementFilters.registerCustomFilters([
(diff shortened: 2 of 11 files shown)


client/src/js/services/grid/GridExport.js
```diff
@@ -24,16 +24,13 @@ function GridExportService(Modal, util) {
    * @description run the export tool
    */
   GridExport.prototype.run = function run() {
-    var gridApi = this.api;
-    var gridOptions = this.gridOptions;
-    var rows = this.ROWS;
-    var cols = this.COLS;
+    const { api } = this;
+    const { options } = this;
```
jniles: Alternatively,
```js
const { api, options, rows, cols } = this;
const request = { api, options, rows, cols };
```

client/src/modules/stock/exit/exit.js
```diff
@@ -117,16 +126,57 @@ function StockExitController(
...
+   * this function is helpful for handling the export of lot row which are in the grid to a csv file
+   * in a correct format.
+   */
+  vm.exportGrid = () => {
+    const columns = vm.gridOptions.columnDefs
+      .filter(col => col.displayName && col.displayName.length)
+      .map(col => ({ displayName : $translate.instant(col.displayName), width : col.width }));
+
+    const rows = vm.gridOptions.data.map(row => {
+      const code = row.inventory && row.inventory.code ? row.inventory.code : null;
```
jniles: Technically, this will work as well:
```js
 // code if exists, else null
const code = row.inventory && row.inventory.code || null;
```

But this is a [code smell](https://en.wikipedia.org/wiki/Code_smell) - seems like a poor design.  Why can't we guarantee that all these items exist?  Should we be exporting data that can change like this?

client/src/modules/templates/modals/export.modal.js
```diff
@@ -12,7 +12,7 @@ function ExportGridModalController(Instance, uiGridConstants, $filter,
 
   var gridOptions = Data.options || {};
   var gridApi = Data.api || {};
-  var filename = Data.filename || 'Export ' + moment().format('YYYY-MM-DD');
+  var filename = Data.filename || `Export ${moment().format('YYYY-MM-DD')}`;
```
jniles, line 15: 👍

client/src/modules/stock/exit/exit.html
```diff
@@ -82,33 +101,43 @@
...
+            <!-- Other Actions -->
+            <div class="grid-toolbar-item">
+              <div uib-dropdown dropdown-append-to-body data-action="grid-lots-tools">
+                  <a class="btn btn-default" uib-dropdown-toggle>
+                    </span> <span class="hidden-xs" translate>FORM.BUTTONS.ACTIONS</span> <span class="caret"></span>
+                  </a>
+                  <ul uib-dropdown-menu role="menu" class="dropdown-menu-right">
+                    <li role="menuitem">
+                      <a href ng-click="StockCtrl.exportGrid()">
+                        <i class="fa fa-file-excel-o"></i> <span translate>FORM.BUTTONS.EXPORT</span>
```
jniles, line 123: Hmm...  should we allow people to export from the grid before the stock exit is even made?  What is the use-case?

client/src/modules/stock/exit/exit.js
```diff
@@ -117,16 +126,57 @@ function StockExitController(
...
   // exposing the grid options to the view
   vm.gridOptions = gridOptions;
 
+  /**
+   * @method exportGrid
+   * @description export the content of the grid to csv.
+   * this function is helpful for handling the export of lot row which are in the grid to a csv file
+   * in a correct format.
+   */
+  vm.exportGrid = () => {
```
jniles: Considering how much manipulation is going into this function, it probably is better to put it in a service.  Maybe with a signature like:
```js
function (columns, rows) { }
```

jniles (commented):
Just a few comments/questions.

client/src/modules/stock/exit/exit.js
mbayopanda: The user can add rows in the grid without defines inventories, in this case the grid is populate but it doesn't have the object that we need (`inventory`), the user is free to export in any time.

jniles: So, instead of having a long list of boolean checks, can we just do something like [the PatientInvoiceItemService](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/invoices/patientInvoiceItem.service.js#L39) where we set a variable to track whether the row should have data or not?  Something like:
```js
  if (row.initialized) {
     // export this row, assume all values are there, throw errors if any values are missing
  }  else {
     // skip this row or refuse to export or something else 
  }
```

It just makes it easier to read instead of a ton of `&& ? :` logic.

client/src/modules/stock/exit/exit.html
mbayopanda: Suppose that we have to get out 100 inventories from our stock, we fill the grid with the 100. And we want to check before the submission that all inventories defined and their lot are correct, we decide to check row by row.

With the ui-grid : 
- It is difficult to move cell by cell and perform search as with Excel
- Share inventories and their lot defined in the grid to other people for an extra checking before the submission (Note : submission cannot be updated)

One of solution found is to export those data

jniles: Okay, let's try it out.  You have more experience with manipulating the stock modules than I do 👍 

We probably need to figure out some way to update submission, honestly.  I can see a lot of things going wrong that the users will need to correct.

client/src/modules/stock/exit/exit.js
mbayopanda: This case is particular to StockItem object, because each rows of the grid contains a StockItem object, we have to convert these rows in a simple type that ui-grid can easily and correctly export.

This function get columns and rows from the grid, and convert them in a good format (a specific format that we cannot generalize), and export these data with the `uiGridExportService`.

If we have to create a service `X` for that, this service will contains just two line of true process : 
- `uiGridExportService.formatAsCsv(rows, cols)`
- `uiGridExportService.downloadFile()`

Because we have to give columns and rows to this service `X` and this case is not frequent and because of these true line of true process that we can call in the controller, so we decide to do not use a service for that job.

Maybe we can correctly rename this function in something like `formatRowsAndColumns`, and we can add another function `exportGridToCsv` which contains `uiGridExportService.formatAsCsv(rows, cols)` and `uiGridExportService.downloadFile()`

jniles: It's actually AngularJS best practice to try and make controllers as small as possible.  You can find mention of this in [the AngularJS documentation for controllers](https://docs.angularjs.org/guide/controller).

> In general, a Controller shouldn't try to do too much. It should contain only the business logic needed for a single view.

For more discussion on why this is considered AngularJS best practice, you can [check out this StackOverflow thread](https://stackoverflow.com/questions/31454488/angularjs-why-not-write-logic-in-controller).

Services are the ideal place to put logic, no matter how small they are - it will make it easier for us to test.  For example, see our [VoucherFormTests](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/client-unit/services/VoucherForm.spec.js).  Basically, we can test creating a voucher without ever having HTML.

If you really believe this is the best place for it, I'll merge it in.  But it does feel like the extra ~25 lines of code could be abstracted a bit.

mbayopanda:
@jniles could you review again my PR

client/src/css/structure.css
```diff
@@ -23,7 +23,12 @@
 /**
  * IMA colors
  */
-.bg-ima-blue { background-color : #085484; color: #fff; }
+:root {
+  --ima-blue : #085484;
+}
+
+.ima-blue { color: var(--ima-blue); }
+.bg-ima-blue { background-color : var(--ima-blue); color: #fff; }
```
jniles, line 31: Looks like CSS variables have [good browser support](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables).  I had no idea!  Cool!

client/src/js/services/StockService.js
```diff
@@ -186,9 +198,9 @@ function StockService(Api, Filters, AppCache, Periods, $httpParamSerializer, Lan
       return {};
     }
 
-    var keys = ['name', 'text', 'display_name'];
+    const keys = ['name', 'text', 'display_name'];
```
jniles, line 201: 👍

client/src/js/services/grid/GridExport.js
```diff
@@ -50,7 +53,35 @@ function GridExportService(Modal, util) {
...
+   *
+   * @param {string} filename - [optional] The name of the csv file
+   * @param {function} rowsFormatterFn - [optional] callback fn to apply to rows
+   * @param {function} columnsFormatterFn - [optional] callback fn to apply to columns
+   */
+  GridExport.prototype.exportToCsv = (filename, columnsFormatterFn, rowsFormatterFn) => {
+    var columns = service.options.columnDefs || [];
+    var rows = service.options.data || [];
+
+    if (columnsFormatterFn) {
```
jniles, line 72: This is really clean 👍

client/src/modules/stock/exit/exit.js
```diff
@@ -117,16 +126,75 @@ function StockExitController(
...
+   *
+   * @param {array} rows - refer to the grid data array
+   * @return {array} - return an array of array with value as an object in this format : { value : ... }
+   */
+  function formatExportRows(rows) {
+    return (rows || []).map(row => {
+      const code = row.inventory && row.inventory.code ? row.inventory.code : null;
+      const description = row.inventory && row.inventory.text ? row.inventory.text : null;
+      const lot = row.lot && row.lot.label ? row.lot.label : null;
+      const price = row.inventory && row.inventory.unit_cost ? row.inventory.unit_cost : null;
```
jniles, line 183: I still feel like we could do much better than this, but I'll have to take a look at the code to see.

jniles (approved):
Getting better!  Good enough to merge.

We should include some tests in a future PR to make sure this doesn't break.

bors r+

PR #2540, opened 2018-02-13
feat(report)implement HGR financial report

### LINKED ISSUE #2448: Implement the HGR Treasury Report
To help hospitals adopt BHIMA, we will need to provide reports they are familiar with.  I propose that we implement the "HGR Treasure Report", which is recommended by the Ministry of Health for hospitals to adopt.  A model of the report is attached to this issue.

The report is basically all the income and expense accounts placed side by side.  The title accounts will contain the values from all the child accounts summed up.  The following guidelines should hold:
 1. Accounts should be ordered by their account number.
 2. Title accounts should be bold and contain the sum of all their children's balances.
 3. All data should come out of the `period_total` table.
 4. Income Accounts (and their titles) should be on the left hand side.
 5. Expense Accounts (and their titles) should be on the right hand side.
 6. The totals at the bottom should total all accounts, but not their title accounts to avoid double-counting. 

See the attached document for more details.

[HGR Treasory Report.xlsx](https://github.com/IMA-WorldHealth/bhima-2.X/files/1626037/HGR.Treasory.Report.xlsx)

jeremielodi:
this PR implement HGR financial report
closes #2448

### DIFF
--- a/server/controllers/finance/fiscal.js
+++ b/server/controllers/finance/fiscal.js
@@ -36,6 +36,8 @@ exports.getNumberOfFiscalYears = getNumberOfFiscalYears;
 exports.getDateRangeFromPeriods = getDateRangeFromPeriods;
 exports.getPeriodIdsFromDateRange = getPeriodIdsFromDateRange;
 exports.accountBanlanceByTypeId = accountBanlanceByTypeId;
+exports.receivables = receivables;
+exports.debts = debts;
 /**
  * @method lookupFiscalYear
  *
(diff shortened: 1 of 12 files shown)


server/controllers/finance/fiscal.js
```diff
@@ -595,3 +597,56 @@ function accountBanlanceByTypeId() {
     ORDER BY ac.number
   `;
 }
+
+// client debts by filscal year id and date range
+function receivables() {
+  return `
+    SELECT
```
jniles, line 604: Can you get this from the `period_total`'s table?  It would make the query faster.

server/controllers/finance/fiscal.js
```diff
@@ -595,3 +597,56 @@ function accountBanlanceByTypeId() {
...
+      GROUP BY gl.account_id 
+    )s ON a.id = s.account_id
+    ORDER BY a.number
+  `;
+}
+
+function debts() {
+  return `
+    SELECT 
+      a.id, a.number, a.label, a.parent, IFNULL(s.balance, 0) AS amount,
```
jniles, line 631: This should also use the period totals table, if possible.

server/controllers/finance/reports/hgr_report/index.js
```diff
@@ -0,0 +1,142 @@
...
+// set the percentage of each amoun's row,
+// round amounts
+function formatData(result, total, decimalPrecision) {
+  const _total = (total === 0) ? 1 : total;
+  return result.forEach(row => {
+    if (row.depth < 2) {
+      row.percent = util.roundDecimal(Math.abs((row.amount / _total) * 100), decimalPrecision);
+    }
+    row.amount = util.roundDecimal(row.amount, decimalPrecision);
+    row.title = row.depth <= 1;
```
jniles: Can you use account types to determine title accounts instead of depth?

server/controllers/finance/reports/hgr_report/index.js
```diff
@@ -0,0 +1,142 @@
...
+
+    tree.sumOnProperty(summableProp);
+    return tree.toArray();
+  }
+  return [];
+}
+
+// set the percentage of each amoun's row,
+// round amounts
+function formatData(result, total, decimalPrecision) {
```
jniles: This code looks almost identical to the code you wrote for the last report.  Can you just use that function?

jniles (commented):
@jeremielodi, just a few suggestions here.

Could you run this by AG Neron when you are done?

server/controllers/finance/reports/hgr_report/index.js
jeremielodi: In this context the title is just the parent account, it has children so it label should be highlighted

server/controllers/finance/reports/hgr_report/index.js
jeremielodi: okay

jniles:

Issue #2397, opened 2017-12-13
Update fiscal periods


jeremielodi:
Yesterday I was studying the process for creating a fiscal Year, I notice That there are procedures that create a period for each month of the fiscal year, but when I tried to update the interval (start date -- end date)   the periods remained the same. Is this a problem? or I just misunderstood.

jniles:
This is a bug.  Other details are https://github.com/IMA-WorldHealth/bhima-2.X/issues/2434.

jniles:
Closed In https://github.com/IMA-WorldHealth/bhima-2.X/pull/2542

PR #2542, opened 2018-02-14
Refactor: Fiscal Year

### LINKED ISSUE #2349: Do not compute period totals for title accounts
By default it looks like the fiscal year module creates period totals for title accounts if they are submitted to it.  This can be seen in #2348, where the opening balances for various title accounts are $0.00.

I propose that we always filter out title accounts before calculating period totals.  This will save us time computing on posting operations and setting opening/closing balances.  If you need a title account's balance, the server can compute it on the fly using the period totals of all its children.

### LINKED ISSUE #2535: Remove Income/Expense accounts from opening balance view
Currently, we should the income and expense accounts in the opening balance view.  These accounts are periodic - they are always annulled at the end of a fiscal year by writing their balances to a balance account (see #1651).

Since the values will always be zero, we should just remove them from the view.

### LINKED ISSUE #2536: Use a TreeService to calculate the title account totals for opening and closing balances
In the Fiscal Year controllers, we frequently display the list of accounts for either editing or viewing, along with their title accounts in a tree.  See below for an example:

needstreeforcalculatiokn
_Fig 1: Opening Balance Tree_

This could really use a summing calculation based on our "Tree" library to calculate the totals of the title accounts by summing their children.  This will help users better understand the data being presented to them.

jniles:
This PR essentially rewrites a lot of the Fiscal Year's core components.  The code was over a year old and a bit brittle.  Plus, a few more corner cases are considered (though not all of them #2397).  Here are the major changes:

 1. The user interface now include text instead of just buttons.  This makes it easier to click and see what the values of each button are.  They are hidden on small screen sizes, though.
 2. A complete rewrite of the core features of the `Tree` library.  This allows much greater flexibility without changing the tree's API at all.  This library has also been ported to the client as the `TreeService`.  In the future, we should figure out how to compile the same code to both the client and the server.
 3. Updated the opening balance modal to show totals on the title accounts and include a grid footer with the balances in the footer.  This feature uses the new `TreeService`.
 4. Opening balances can only be edited for the first year, so I've removed the button editing the fiscal year's opening balances.  If you are in the first fiscal year, it is open for edit.  Otherwise, it is closed.
 5. Removed the Income/Expense accounts from the opening balance modal.  Also ensured we are not posting balances for Title Accounts.
 6. Made all fiscal controllers and services pass ESLint checks on the client.
 7. Used "filter" icons instead of "search" icons for filtering the ui-grid (we do this in all other places in the application, it's confusing to do it differently here).
 8. A closed fiscal year can no longer be closed again.  The button is hidden/removed.  

Here are some screenshots:

fiscalyearlist
_Fig 1: Fiscal Year List_

beginningbalancemodalthatcanbeedited
_Fig 2: First Fiscal Year's Opening Balance can be edited_

beginningbalancemodalthatcannotbeedited
_Fig 3: Other Fiscal Year's Opening Balances cannot be edited_

Closes #2349. 
Closes #2535. 
Closes #2536.


client/src/modules/fiscal/fiscal.closingBalance.html
```diff
@@ -12,7 +12,7 @@
           </span>
 
           <span class="text-action" ng-click="FiscalCBCtrl.toggleAccountFilter()">
-            <span translate>FORM.BUTTONS.SEARCH</span> <i class="fa fa-search"></i>
+             <i class="fa fa-filter"></i> <span translate>FORM.BUTTONS.SEARCH</span>
```
mbayopanda, line 15: 👍

client/src/modules/fiscal/fiscal.list.html
```diff
@@ -40,9 +40,9 @@ <h4 style="margin-top : 0px;">
             </div>
 
             <div class="row text-lowercase">
-              <i class="fa fa-clock-o"></i> {{ "FORM.INFO.CREATED" | translate }} <span am-time-ago="fiscal.created_at"></span>
+              <i class="fa fa-clock-o"></i> <span translate>FORM.INFO.CREATED</span> <span am-time-ago="fiscal.created_at"></span>
               <span translate> FORM.INFO.BY </span>
-              <a href="">
+              <a href="" ui-sref="users.list({ id : fiscal.user_id })" class="text-capitalize">
```
mbayopanda, line 45: 👍

mbayopanda (changes requested):
@jniles i did try to run the application but i have an error : Error: $injector:unpr Unknown Provider related to Tree.

Is it work correctly on your computer ?

jniles:
That's curious ... end to end tests would usually break if there is an $injector error.

On what controller does it happen?  I've tested most of the controllers in the fiscal year page.  Did you rebuild the client side?

client/src/js/services/TreeService.js
```diff
@@ -0,0 +1,217 @@
...
+ * @description
+ * This file contains the generic class definition of a tree. A tree is defined
+ * as an array of JSON objects having a parent key referring to another member
+ * of the array.  The only exception is the root node, which does not need to be
+ * in the tree.
+ *
+ * This code is also found (in a similar form) on the server in /lib/Tree.js
+ */
+
+class TreeService {
```
mbayopanda, line 13: 👍

mbayopanda (approved):
LGTM

mbayopanda:
@jniles could you fix the conflict and push again this branch

jniles:
Note: the import previous year balance is no longer necessary and should be removed in a subsequent PR.

jniles:
Hmm... weird build timeout issues.  I've increased the timeout, we'll see if Travis is happy.

bors r+

jniles:
So, it was code climate causing bors to fail since I had marked `master` as a "protected branch" and codeclimate as a required check.  Oops!  It looks like, since I changed the ESLint definitions, code climate marked all problems in the repository as problems with this branch...

Trying again w/o requiring codeclimate.

bors r+

PR #2544, opened 2018-02-15
feat(stock exit) avoid lot multi-selection

### LINKED ISSUE #2534: feat(stock): message for multi-usage of a lot in a grid
This issue is about to show a message if a lot is used more than one time in the stock entry/exit grid

jeremielodi:
closes #2534

### DIFF
--- a/client/src/i18n/en/inventory.json
+++ b/client/src/i18n/en/inventory.json
@@ -25,6 +25,7 @@
         "REGISTRY": "Inventories Registry",
         "EXPIRES" : "Expires",
         "UNIQUE_ITEM" : "Unique item",
-        "IS_BROKEN" : "is broken"
+        "IS_BROKEN" : "is broken",
+        "LOT_ALREAD_SELECTED": "This lot has already been selected"
     }
 }
--- a/client/src/i18n/fr/inventory.json
+++ b/client/src/i18n/fr/inventory.json
@@ -25,6 +25,8 @@
         "NO_INVENTORY_REGISTERED": "Il n'y aucun inventaire enregistré jusque là",
         "EXPIRES" : "Expire",
         "UNIQUE_ITEM" : "Un seul item",
-        "IS_BROKEN" : "est endommagé"
+        "IS_BROKEN" : "est endommagé",
+        "IS_BROKEN" : "is broken",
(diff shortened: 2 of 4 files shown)


jniles:
@jeremielodi you should check out the [Pool](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/Pool.js).  It was made to solve this exact problem.

The [unit tests](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/client-unit/services/Pool.spec.js) give some information on its usage.

jeremielodi:
okay

jeremielodi:
okay but stockItem's structure is not easy to be combined with poolService

jniles:
It wouldn't actually be `stockItem`, it would be the `StockForm`.   Currently the [StockForm uses a Store](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/stock/StockForm.service.js#L28).   You can find the Store's code [here](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/Store.js).  It is an elegant solution for simple data stores that we adapted from [The Dojo Toolkit](https://dojotoolkit.org/).

The idea is to replace the `this.store = new Store` with `this.store = new Pool`.  Then there are three API changes to make:
 1. Change all references to `this.store.data` to `this.store.available`.
 2. Change all references to `this.store.add` to `this.store.use`.
 3. Change all references to `this.store.remove` to `this.store.release`.

And that should be it.  Since a Pool uses two stores underneath, the constructor is the same.  To avoid confusing other developers, you might change `this.store` to `this.pool`.

[The PatientInvoiceForm service uses a pool to ensure the same inventory item is not used twice](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/invoices/patientInvoiceForm.service.js#L92).  You might gain inspiration from its code.

The purpose of a pool is to make adding/removing and keeping track of used elements super efficient.  Adding an element, removing an element, and checking if it is available all take `O(1)` instead of `O(n)`, as in the `for ... ` loop solution presented here.

jniles:
Closed in #3194.

PR #2553, opened 2018-02-20
Unit test PatientInvoiceFormService

jniles:
This PR adds unit tests using Karma to the PatientInvoiceForm.  The tests are not complete, but major steps have been tested, including the initialization step, setting a patient, and digests. It also rewrites the PatientInvoiceFormService to comply with ESLint.  Finally, a new service has been added to Karma to allow sharing mocked data between tests called `bhima.mocks`.

### DIFF
--- a/client/src/modules/invoices/patientInvoiceForm.service.js
+++ b/client/src/modules/invoices/patientInvoiceForm.service.js
@@ -2,8 +2,9 @@ angular.module('bhima.services')
   .service('PatientInvoiceForm', PatientInvoiceFormService);
 
 PatientInvoiceFormService.$inject = [
-  'PatientService', 'PriceListService', 'InventoryService', 'AppCache', 'Store', 'Pool',
-  'PatientInvoiceItemService', 'bhConstants', 'ServiceService', '$q', '$translate', 'NotifyService',
+  'PatientService', 'PriceListService', 'InventoryService', 'AppCache', 'Store',
+  'Pool', 'PatientInvoiceItemService', 'bhConstants', 'ServiceService', '$q',
+  '$translate', 'NotifyService',
 ];
 
 /**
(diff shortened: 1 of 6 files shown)


test/client-unit/services/PatientInvoiceForm.spec.js
```diff
@@ -0,0 +1,224 @@
...
+    const subsidies = Mocks.subsidies();
+
+    expect(form.prices.data).to.have.deep.members(priceList.items);
+
+    // test that all ids of invoice fees are in the form.
+    const invoiceFeeIds = form.invoicingFees.map(fee => fee.id);
+
+    // console.error(form.invoicingFees);
+    // console.error(' === ');
+    // console.error(invoiceFees);
```
mbayopanda: these comments are still here

client/src/modules/invoices/patientInvoiceForm.service.js
```diff
@@ -465,15 +456,15 @@ Patients, PriceLists, Inventory, AppCache, Store, Pool,
...
-      var item = this.addItem();
+    cp.items.forEach((cacheItem) => {
+      const item = this.addItem();
 
       item.inventory_uuid = cacheItem.inventory_uuid;
       this.configureItem(item);
 
       item.quantity = cacheItem.quantity;
       item.transaction_price = cacheItem.transaction_price;
-    }.bind(this));
```
mbayopanda, line 476: By removing this `.bind(this)`, the function inside the foreach can still see the parent `this` and call it like ``this.addItem(); ?

mbayopanda (approved):
LGTM
I run the test locally on my machine it works fine.
👍

client/src/modules/invoices/patientInvoiceForm.service.js
jniles: Yup!  That is the power of arrow functions in JavaScript.  [Here is the MDN documentation on it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_separate_this)

test/client-unit/services/PatientInvoiceForm.spec.js
jniles: Whoops, you are right. :(

jniles:
@mbayopanda changes have landed!

PR #2554, opened 2018-02-21
test(handlebars) : unit test for handlebars rendering

mbayopanda:
This PR tests the `lib/template` based on handlebars for seeing if values displayed by the library are correct and are in the correct format


jniles (approved):
LGTM.  Great tests!

PR #2555, opened 2018-02-21
deps(ci): bump minimal node to 8+

jniles:
This commit bumps to latest tested node version on Travis to node version 8.


jniles:
@mbayopanda once this is merged, you should rebase #2554 to master.  It's currently breaking because of `async/await` support in Node 6.

PR #2563, opened 2018-02-23
Refactor: Account Statement Report

### LINKED ISSUE #1919: refactor: standardize display of credit amounts in BHIMA
In BHIMA, we have two ways that we display negative/credit amounts.  Either with a negative sign or with parentheses in red.  Here are two typical examples:

creditsongeneralledger
_Fig 1: Negative Numbers have a Preceding minus sign_

creditsonpatientfinancialrecord
_Fig 2: Negative Number are in Red with Parens_

 Typically, spreadsheet systems (like Excel or Sanru Trakker) choose a standard way of displaying these amounts.  BHIMA should adopt a standard and use it everywhere.

In my opinion, the preferable way to display these numbers is as in **Fig 2**, with a **red highlight and surrounding parentheses**.  That way, even in small cells that cut off part of the text you will always know that it is a credit balance.

@IMA-WorldHealth/bhima-core what do you guys think?

/cc @WayneNiles

### LINKED ISSUE #2557: Improvements to the Account Statement Report
We've been given feedback on our Account Statement report.  The current presentation is confusing since it mixes currencies without an exchange rate column. The following proposal is attached as an Excel.  It involves:
 1. Adding a column for the Exchange Rate to the table
 2. Separating out the amounts to the include a "Amount" for the original amount and "Value {{currency}}" to give the valuation in that currency.
 3. Providing the running balance in the selected currency as "Total {{ currency }}".

[Rapports devise et normal.xlsx](https://github.com/IMA-WorldHealth/bhima-2.X/files/1747392/Rapports.devise.et.normal.xlsx)

jniles:
This commit implements multi-currency support on the Account Statement
Report. It is derived from a template by @WayneNiles.

The logic is somewhat complex.  If a user produces a report in the
enterprise currency, then the values should convert from a foreign
currency into the enterprise currency.  This is done through _division_
with the formula `amount / rate`.  However, if the user produces a
report in a different currency, all values must be converted from the
enterprise currency to that foreign currency.  This happens by
_multiplication_ with the formula `amount * rate`.

The report has been clarified by:
 1. Combining the two footers into one and naming it `footer`.
 2. Renaming the `openingBalance` property to `header`, since that is
 what is it.
 3. Removing the multiple options on the report page to simply give
 "currency" instead of having different total currencies.
 4. Renaming `accountDetails` to just `account`.

Finally, a new HBS helper has been made for rendering debits/credits
nicely.  It will automatically convert negative values into red currency
values.

Closes #1919.  Closes #2557.

See the attached example report: 
[ExampleReport.pdf](https://github.com/IMA-WorldHealth/bhima-2.X/files/1752025/ExampleReport.pdf)

### DIFF
--- a/client/src/modules/reports/generate/account_report/account_report.html
+++ b/client/src/modules/reports/generate/account_report/account_report.html
@@ -43,55 +43,12 @@
               validation-trigger="ConfigForm.$submitted">
             </bh-date-interval>
 
-            <bh-hidden-field
-              show-text="REPORT.OPTIONS.MORE_OPTIONS"
-              hide-text="REPORT.OPTIONS.MORE_OPTIONS">
-
-              <div class="radio" ng-class="{ 'has-error': ConfigForm.$submitted && ConfigForm.useOriginalTransactionCurrency.$invalid }">
-                <p class="control-label" style="margin-bottom:5px;">
-                  <strong translate>REPORT.OPTIONS.USE_ORIGINAL_TRANSACTION_CURRENCY</strong>
-                </p>
-
-                <label
-                  class="radio-inline"
-                  translate-attr="{ title : 'FORM.LABELS.YES' }">
-                  <input
-                    name="useOriginalTransactionCurrency"
-                    type="radio"
-                    ng-model="ReportConfigCtrl.reportDetails.useOriginalTransactionCurrency"
-                    ng-value="true"
-                    required>
-                  <span translate>FORM.LABELS.YES</span>
-                </label>
-
-                <label
-                  class="radio-inline"
-                  translate-attr="{ title : 'FORM.LABELS.NO' }">
-                  <input
-                    name="useOriginalTransactionCurrency"
-                    type="radio"
-                    ng-model="ReportConfigCtrl.reportDetails.useOriginalTransactionCurrency"
-                    ng-value="false"
-                    required>
-                  <span translate>FORM.LABELS.NO</span>
-                </label>
-
-                <div class="help-block" ng-messages="ConfigForm.$error" ng-show="ConfigForm.$submitted">
-                  <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
-                </div>
-
-                <p class="help-block" translate>REPORT.OPTIONS.USE_ORIGINAL_TRANSACTION_CURRENCY_HELP</p>
-              </div>
-
-              <!-- the currency to be used in the footer -->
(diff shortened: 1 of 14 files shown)


jniles:
AG Neron has indicated that he would like the totals of the debit/credit at the bottom of the list of accounts.  If the account contains transactions in multiple currencies, we will not display a balance (what is 100FC + 0.35USD?), but for those with a single currency, the balance will be displayed.

client/src/modules/reports/generate/account_report/account_report.html
```diff
@@ -43,55 +43,12 @@ <h3 class="text-capitalize" translate>REPORT.REPORT_ACCOUNTS.TITLE</h3>
               validation-trigger="ConfigForm.$submitted">
             </bh-date-interval>
 
-            <bh-hidden-field
```
mbayopanda, line 46: The form is more clear now 👍

server/lib/template/helpers/finance.js
```diff
@@ -45,6 +46,22 @@ function indentAccount(depth) {
   return number ? number * INDENTATION_STEP : 0;
 }
 
+function debcred(value = 0, currencyId) {
```
mbayopanda, line 49: 🕺 🥇 really useful and well done

mbayopanda (approved):
The code and behaviors looks good to me.

I just noticed that i have zero display as `Taux` for `1$=930FC` when i choose Fc as currency

account report

jniles:
That exchange rate issue is a bug - I never tested USD as an enterprise currency,  Let me work on it a bit more...

jniles:
@mbayopanda, good catch!  I've now implemented a fix.  LMK what you think.

server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -248,7 +248,7 @@ function getAccountTransactions(options, openingBalance = 0) {
         exchangedBalance : totals.balance * totals.rate,
         exchangedCumSum : lastCumSum,
         exchangedDate : new Date(),
-        invertedRate : util.roundDecimal(1 / totals.rate, 2),
+        invertedRate : Exchange.formatExchangeRateForDisplay(totals.rate),
```
mbayopanda: 👍

mbayopanda (approved):

mbayopanda:
Now it is fixed 
account report 2

PR #2564, opened 2018-02-23
ci: remove lint checks

jniles:
This commit removes lint checks from the repository.  ESLint should be installed in your editor and CodeClimate should ensure PRs are brought in with passing lint.


jniles:
These flaky tests are getting annoying.

PR #2567, opened 2018-02-23
fix stock entry purchase label for supplier

mbayopanda:
This PR set the supplier name in the stock entry ui for purchase instead of user name.

purchase_supplier


client/src/modules/stock/entry/entry.js
```diff
@@ -317,9 +316,7 @@ function StockEntryController(
 
   // validation
   function hasValidInput() {
-    return vm.stockForm.store.data.every(function checkLot(line) {
-      return line.lots.length > 0;
-    });
+    return vm.stockForm.store.data.every(line => line.lots.length > 0);
```
jniles, line 319: :+1:

jniles (approved):
Looks good to me!

PR #2569, opened 2018-02-24
fix(Trial Balance): limit lookup to FY

### LINKED ISSUE #2568: Trial Balance does not limit to most recent fiscal year
The Trial Balance groups the entire period totals table when calculating the `balance_before` column in the Trial Balance view.  While is acceptable if all data is correct, if previous fiscal years have corrupted data, it perpetuates the data corruption.

This causes the `balance_before` to be different from the balance reported by both the General Ledger and the Account Statement.  We should bring the Trial Balance in line with those two reports.

jniles:
This commit limits the Trial Balance lookup to only oldest fiscal year.

Closes #2568.

### DIFF
--- a/server/models/procedures/trial_balance.sql
+++ b/server/models/procedures/trial_balance.sql
@@ -190,6 +190,16 @@ CREATE PROCEDURE TrialBalanceSummary()
 BEGIN
   -- this assumes lines have been staged using CALL StageTrialBalanceTransaction()
 
+  -- fiscal year to limit period_total search
+  DECLARE fiscalYearId MEDIUMINT;
+
+  -- get the fiscal year of the oldest record to limit period_total search
+  SET fiscalYearId = (
+    SELECT MIN(fiscal_year_id)
+    FROM posting_journal JOIN stage_trial_balance_transaction
+      ON posting_journal.record_uuid = stage_trial_balance_transaction.record_uuid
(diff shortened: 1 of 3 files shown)


server/models/procedures/trial_balance.sql
```diff
@@ -190,23 +190,36 @@ CREATE PROCEDURE TrialBalanceSummary()
...
+      WHERE period_total.fiscal_year_id = fiscalYearId
+    ) totals ON u.account_id = totals.account_id
     GROUP BY u.account_id;
 
   SELECT account_id, account.number AS number, account.label AS label,
     balance_before, debit_equiv, credit_equiv,
     balance_before + debit_equiv - credit_equiv AS balance_final
   FROM (
-    SELECT posting_journal.account_id, SUM(totals.balance_before) AS balance_before, SUM(debit_equiv) AS debit_equiv,
+    SELECT posting_journal.account_id, MAX(totals.balance_before) AS balance_before, SUM(debit_equiv) AS debit_equiv,
```
mbayopanda, line 222: Could you fix this `MAX` if it is required, in my test i cannot see the error with `MAX`

mbayopanda (commented):
I did a test locally with the test dataset the trial balance shows correct values. 

it looks good to me 👍

Issue #2538, opened 2018-02-13
Export Receipts/Reports as Excel


jniles:
We had a request for a stock receipt as an Excel document.  This probably isn't too hard to do, but will need some careful designing.  Ideally, the same format should be used for every single receipt/invoice in the system.  It will definitely make it easier to export data from BHIMA to use in other contexts.

jniles:
This looks like a library worth investigating: https://github.com/dtjohnson/xlsx-populate

Demo: https://gitcdn.xyz/cdn/dtjohnson/xlsx-populate/32921c0b357828cf4c9065c60aa5ab9ad5aefd62/examples/browser/index.html

jeremielodi:
I have discover this <a href='https://www.npmjs.com/package/excel4node'>excel4node</a>
I'm tring to use it

PR #2572, opened 2018-02-27
feat(receipt) xlsx export

jeremielodi:
This PR helps to render receipts as Excel,  it tries to solve this issue #2538

### DIFF
--- a/server/controllers/finance/reports/invoices/index.js
+++ b/server/controllers/finance/reports/invoices/index.js
@@ -29,6 +29,8 @@ const RECEIPT_TEMPLATE = './server/controllers/finance/reports/invoices/receipt.
 const REPORT_TEMPLATE = './server/controllers/finance/reports/invoices/report.handlebars';
 const CREDIT_NOTE_TEMPLATE = './server/controllers/finance/reports/invoices/creditNote.handlebars';
 
+const xlsx = require('./receipt.xlsx');
+
 exports.report = report;
 exports.receipt = receipt;
 exports.creditNote = creditNote;
@@ -116,6 +118,8 @@ function receipt(req, res, next) {
 
   let template = RECEIPT_TEMPLATE;
 
+  const renderer = req.query.renderer;
+
   if (Boolean(Number(options.posReceipt))) {
     template = POS_RECEIPT_TEMPLATE;
     _.extend(options, pdf.posReceiptOptions);
@@ -162,9 +166,16 @@ function receipt(req, res, next) {
         invoiceResponse.exchangedTotal = _.round(invoiceResponse.cost * invoiceResponse.exchange);
       }
 
+      // render the excel worksheet
+      if (renderer === 'xlsx') {
+        return xlsx.render(invoiceResponse);
+      }
+
       return receiptReport.render(invoiceResponse);
+
     })
     .then(result => {
+
       res.set(result.headers).send(result.report);
     })
     .catch(next)
--- a/server/controllers/finance/reports/invoices/index.js
+++ b/server/controllers/finance/reports/invoices/index.js
@@ -29,8 +29,6 @@ const RECEIPT_TEMPLATE = './server/controllers/finance/reports/invoices/receipt.
 const REPORT_TEMPLATE = './server/controllers/finance/reports/invoices/report.handlebars';
 const CREDIT_NOTE_TEMPLATE = './server/controllers/finance/reports/invoices/creditNote.handlebars';
 
-const xlsx = require('./receipt.xlsx');
-
 exports.report = report;
 exports.receipt = receipt;
 exports.creditNote = creditNote;
@@ -118,8 +116,6 @@ function receipt(req, res, next) {
 
   let template = RECEIPT_TEMPLATE;
 
-  const renderer = req.query.renderer;
-
   if (Boolean(Number(options.posReceipt))) {
     template = POS_RECEIPT_TEMPLATE;
     _.extend(options, pdf.posReceiptOptions);
@@ -164,11 +160,6 @@ function receipt(req, res, next) {
       invoiceResponse.dateFormat = (new Moment()).format('L');
       if (invoiceResponse.exchange) {
         invoiceResponse.exchangedTotal = _.round(invoiceResponse.cost * invoiceResponse.exchange);
-      }
-
-      // render the excel worksheet
-      if (renderer === 'xlsx') {
-        return xlsx.render(invoiceResponse);
       }
 
       return receiptReport.render(invoiceResponse);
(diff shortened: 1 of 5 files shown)


server/controllers/finance/reports/invoices/index.js
```diff
@@ -162,9 +166,16 @@ function receipt(req, res, next) {
         invoiceResponse.exchangedTotal = _.round(invoiceResponse.cost * invoiceResponse.exchange);
       }
 
+      // render the excel worksheet
+      if (renderer === 'xlsx') {
+        return xlsx.render(invoiceResponse);
```
jniles: Can you put the `xlsx` into the `receiptReport` renderer?  That way it can all use the same interface?  I think that already controls json/html/pdf versions.  It would make sense to put it there.

server/controllers/finance/reports/invoices/receipt.xlsx.js
```diff
@@ -0,0 +1,194 @@
+// Require library
+const xl = require('excel4node');
```
jniles: This is pretty cool!

Could you write a small unit test to show that this works?  That way we can be sure changes in the future do not break this renderer.

jniles (commented):
@jeremielodi this looks really cool!  I'm glad you got it working!

Since this introduces a new library/dependency, it would be helpful to have a unit test to make sure it produces an XLSX document correctly.  That way, when we adapt it for our reports, we'll be able to be sure nothing breaks.

Could you add that?

server/controllers/finance/reports/invoices/index.js
jeremielodi: okay, but we have already an old xlsx renderer, I'm trying to rewrire it

jeremielodi: or we can just give this one another name such as xlsx2

jniles: Aha, good thought.  Hmm...

For now, let's give it another name like xlsx2.  That way, we can get this PR merged and then gradually combine them.

server/lib/renderers/xlsxReceipt.js
```diff
@@ -0,0 +1,207 @@
...
+  };
+
+  // use translate lib fro translating labels
+  const t = translate(data.lang);
+  const { currencySymbol, currencyName } = data.enterprise;
+
+  // let specify information for each excel cell
+
+  // Header's informations
+  // cell merge , cell(1,1) =A1; cell(1,3) = C1
```
jniles, line 112: These are really great comments!

server/lib/renderers/xlsxReceipt.js
```diff
@@ -0,0 +1,207 @@
...
+  // Create a new instance of a Workbook class
+  const wb = new xl.Workbook();
+
+  // Add Worksheets to the workbook
+  const ws = wb.addWorksheet('Sheet 1');
+
+  // Create a reusable style
+  // bold,hightlighted
+
+  const style1 = wb.createStyle({
```
jniles: It would be best to name this "bold" or something that indicates what it actually does.  That way we can read:
```js
ws.cell(1, 1, 1, 3, true).string(data.enterprise.name).style(bold); 
```
and it tells use exactly what it is doing!

server/lib/renderers/xlsxReceipt.js
```diff
@@ -0,0 +1,207 @@
...
+  ws.cell(12, 1).string(t('FORM.LABELS.DESCRIPTION')).style(styleUndeLine);
+  ws.cell(13, 1, 13, 7, true).string(data.description).style(style2);
+  ws.cell(13, 1, 13, 7, true).style(styleMultiLine);
+  ws.row(13).setHeight(40); // description might be pretty long
+
+  // detail
+  ws.cell(15, 1).string(t('FORM.LABELS.INVOICES_DETAILS')).style(styleUndeLine);
+
+  // titles
+  let line = 16;
```
jniles: Nice comments and variable names.

server/lib/renderers/xlsxReceipt.js
```diff
@@ -0,0 +1,207 @@
...
+    ws.cell(line, 2, line, 4, true).string(row.text);
+    ws.cell(line, 5).number(row.transaction_price);
+    ws.cell(line, 6).number(row.quantity);
+
+    const total = (row.transaction_price * row.quantity);
+    ws.cell(line, 7).string(''.concat(total, '', currencySymbol));
+
+    ws.cell(line, 1, line, 7).style(style2);
+    ws.cell(line, 1, line, 7).style(styleAllBorders);
+    line++;
```
jniles: Instead of doing `line++`, you could also use an offset:
```js
const offset = 17;
data.items.forEach((row, idx) => {
  const line = idx + offset;
  // other code...
});
```

server/lib/renderers/xlsxReceipt.js
```diff
@@ -0,0 +1,207 @@
...
+  ws.cell(line, 3, line, 7, true).string(numberToText.convert(data.cost, data.lang, currencyName)).style(style1);
+  ws.cell(line, 3, line, 7, true).style(styleAllBorders);
+  line++;
+
+  const deferred = q.defer();
+
+  wb.writeToBuffer().then(xlsxBuffer => {
+    deferred.resolve(xlsxBuffer);
+  });
+  return deferred.promise;
```
jniles: There is no need to do this - in fact, you will lose errors from this if `wb.writeToBuffer()` errors since you have no catch statement.

Just return the `wb.writeToBuffer()`.  Like this:
```js
return wb.writeToBuffer();
```

jniles (commented):
Look pretty good!

The only blocking comment here is to `return wb.writeToBuffer()`.  That will make it so that the `catch()` is caught by an upstream module.

Can you add a test or two?

server/lib/renderers/xlsxReceipt.js
jeremielodi: ```line``` is used in many places, not only here

server/lib/renderers/xlsxReceipt.js
```diff
@@ -0,0 +1,202 @@
...
+  ws.cell(line, 3, line, 4, true).string(t('FORM.LABELS.TOTAL')).style(bold);
+  ws.cell(line, 5, line, 7, true).number(data.cost).style(bold);
+  ws.cell(line, 3, line, 7, true).style(styleAllBorders);
+  line++;
+
+  ws.cell(line, 3, line, 7, true).string(numberToText.convert(data.cost, data.lang, currencyName)).style(bold);
+  ws.cell(line, 3, line, 7, true).style(styleAllBorders);
+  line++;
+
+  return wb.writeToBuffer();
```
jniles, line 199: :dancer:

jniles (approved):
LGTM!

Could you rebase this PR to master and see if that fixes the snyk error?  It also looks like you have one more codeclimate error to fix.

jniles:
Guess we can merge this without a test.  We'll do a test later.

bors r+

PR #2574, opened 2018-02-27
fix(package): remove amzn backups

jniles:
This commit removes the amazon backup script.  It didn't work in production (too slow internet connection) and has snyk complaining about errors.  It also bumps the versions of other packages.


jniles:
Snyk is complaining about PRs https://github.com/IMA-WorldHealth/bhima-2.X/pull/2572 and https://github.com/IMA-WorldHealth/bhima-2.X/pull/2569, both of which are blocked by it.  This will enable them to be unblocked.

bors r+

PR #2578, opened 2018-03-01
refactor(cashflow) : use a simplified cashflow report by transaction type

mbayopanda:
This PR refactor the cashflow report by : 
- Removing not necessary codes
- Grouping cashflow information according transaction type
- Showing details of each transaction type
- Extracting the transaction type `Transfer` from incomes


client/src/i18n/en/report.json
```diff
@@ -25,7 +25,13 @@
     "CASHFLOW" : {
       "TITLE":"Cash Flow Physical Report",
       "DESCRIPTION":"This report shows changes in the accounts and revenues of the balance sheet on cash",
-      "SHOW_WEEKLY_REPORT_HELP":"Enabling this option will allow to view the cash flow report per week"
+      "SHOW_WEEKLY_REPORT_HELP":"Enabling this option will allow to view the cash flow report per week",
```
jniles: Does this option still apply?

client/src/modules/reports/generate/cashflow/cashflow.config.js
```diff
@@ -6,13 +6,13 @@ CashFlowConfigController.$inject = [
 ];
 
 function CashFlowConfigController($sce, Notify, SavedReports, AppCache, reportData, $state) {
-  var vm = this;
-  var cache = new AppCache('configure_cashflow');
-  var reportUrl = 'reports/finance/cashflow';
+  const vm = this;
```
jniles, line 9: 👍

client/src/modules/reports/generate/cashflow/cashflow.html
```diff
@@ -29,26 +29,11 @@ <h3 class="text-capitalize" translate>REPORT.CASHFLOW.TITLE</h3>
...
-              </label>
-            </div>
-            <p class="help-block" translate>REPORT.CASHFLOW.SHOW_WEEKLY_REPORT_HELP</p>
-
-            <bh-cashbox-select
-              cashboxId = "ReportConfigCtrl.reportDetails.account_id"
-              name = "cashbox"
-              on-select-callback = "ReportConfigCtrl.onSelectCashbox(cashbox)">
-            </bh-cashbox-select>
+            <bh-multiple-cashbox-select
```
jniles, line 32: 👍

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -7,610 +7,39 @@
...
-//      WHERE ? BETWEEN p.start_date AND p.end_date`;
-//   return db.exec(query, [date]);
-// }
-
 /**
  * @function getPeriods
  * @param {date} dateFrom A starting date
  * @param {date} dateTo A stop date
  */
 function getPeriods(dateFrom, dateTo) {
```
jniles: Can you use [Fiscal.getPeriodIdsFromDateRange](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/fiscal.js#L592)?  Or does it need to be a custom query?

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +182,226 @@ function reportByService(req, res, next) {
...
+ * with their balance for each transaction type
+ */
+function report(req, res, next) {
+  let serviceReport;
+  const dateFrom = new Date(req.query.dateFrom);
+  const dateTo = new Date(req.query.dateTo);
+  const options = _.clone(req.query);
+  const data = {};
+
+  const cashboxesIds = _.values(req.query.cashboxesIds);
```
jniles: You should probably add a comment here for future developers to know why this exists.
```js
// ensure cashboxIds is an array
const cashboxesIds = _.values(req.query.cashboxesIds);
```

test/end-to-end/shared/components/bhMultipleCashBoxSelect.js
```diff
@@ -10,8 +10,8 @@ module.exports = {
 
     target.click();
 
-    cashboxes.forEach(function (cashboxes){
-    	FU.uiSelect('$ctrl.cashboxIds', cashboxes);
+    (cashboxes || []).forEach(cashbox => {
```
jniles: You could also set the default value in the function parameter, like this:
```js
set : function set(cashboxes = [], id) {
 // ... etc ...
```

server/controllers/finance/reports/cashflow/report.handlebars
```diff
@@ -1,220 +1,159 @@
...
+                  {{#each ../../periods as |period| }}
+                    <td class="text-right">{{debcred (look value period) ../../../metadata.enterprise.currency_id}}</td>
+                  {{/each}}
+                </tr>
               {{/each}}
-            </tr>
+
+              {{!-- blank line --}}
+              <tr>
+                <td colspan="{{add ../periods.length 1}}">&nbsp;</td>
```
jniles: Could you calculate this number in the JS?  It makes it easier to debug in the future.

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +182,226 @@ function reportByService(req, res, next) {
...
+
+  getCashboxesDetails(cashboxesIds)
+    .then(rows => {
+      data.cashboxes = rows;
+
+      data.cashAccountIds = data.cashboxes.map(cashbox => cashbox.account_id);
+
+      data.cashLabels = data.cashboxes.map(cashbox => `${cashbox.label}`);
+
+      data.cashLabels = _.chain(data.cashLabels).uniq().join(' | ').value();
```
jniles: You could actually make this a giant chain with the line above:
```js
data.cashLabels = _.chain(data.cashboxes)
  .map(cashbox => `${cashbox.label}`)
  .uniq()
  .join(' | ')
  .value();
```

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +182,226 @@ function reportByService(req, res, next) {
...
+            gl.period_id, gl.account_id
+          FROM general_ledger gl
+          JOIN account a ON gl.account_id = a.id
+          JOIN transaction_type tt ON gl.origin_id = tt.id
+          WHERE gl.uuid IN ? 
+          GROUP BY a.id, tt.id
+        ) AS source 
+        GROUP BY transaction_type, account_id;
+      `;
+      return data.detailsIdentifiers.length ? db.exec(query, [[data.detailsIdentifiers]]) : [];
```
jniles: Why don't you do this early in the function?

```js
.then(periods => {
  data.periodDates = periods.map(p => p.start_date);
  data.periods = periods.map(p => p.id);

  // skip period matrix if not transactions were identified
  if (data.detailsIdentifiers.length === 0) {
     return [];
  }
  
  // ... etc...
  return db.exec(query, [[data.detailsIdentifiers]]);
```

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +182,226 @@ function reportByService(req, res, next) {
...
+function getDetailsIdentifiers(cashboxesAccountIds, dateFrom, dateTo) {
+  const ids = cashboxesAccountIds;
+  const queryTransactions = `
+      SELECT gl.trans_id 
+      FROM general_ledger gl 
+      WHERE gl.account_id IN ? 
+        AND 
+        (DATE(gl.trans_date) >= DATE(?) AND DATE(gl.trans_date) <= DATE(?))
+        AND
+        (gl.origin_id <> 10 AND gl.record_uuid NOT IN (SELECT v.uuid FROM voucher v WHERE v.type_id = 10));
```
jniles: What is GL origin_id 10?  Could you leave a comment?

jniles: Out of curiosity, do you need both these conditions?  Won't one of them work?

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +182,226 @@ function reportByService(req, res, next) {
...
+    `;
+
+  return db.exec(queryTransactions, [[ids], dateFrom, dateTo])
+    .then(rows => {
+      const transIds = rows.map(row => row.trans_id);
```
jniles: Can you use `record_uuid` instead of `trans_id`?  MySQL will compares binary, which is faster than comparing strings.

jniles (commented):
@mbayopanda this code is much easier to read!  Well done!

I would really prefer the code use SQL for totaling.  It should be much faster and safer.  Is there any way we can transition this code to using  SQL for the totals?

I'm made a few supplementary comments - let me know what you think!

server/controllers/finance/reports/cashflow/index.js
mbayopanda: `Fiscal.getPeriodIdsFromDateRange` returns one element (`db.one`).

jniles: Oh, look at that.  Nevermind!

jniles:
@mbayopanda, let me know when you need another review.

mbayopanda:
Could you review this PR ?

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +169,241 @@ function reportByService(req, res, next) {
...
+ * with their balance for each transaction type
+ */
+function report(req, res, next) {
+  let serviceReport;
+  const dateFrom = new Date(req.query.dateFrom);
+  const dateTo = new Date(req.query.dateTo);
+  const options = _.clone(req.query);
+  const data = {};
+
+  // convert cashboxesIds parameters in array format ['', '', ...]
```
jniles, line 185: Nice! 👍

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +169,241 @@ function reportByService(req, res, next) {
...
+  // this parameter can be sent as a string or an array we force the conversion into an array
+  const cashboxesIds = _.values(req.query.cashboxesIds);
+
+  _.extend(options, { orientation : 'landscape' });
+
+  // catch missing required parameters
+  if (!dateFrom || !dateTo || !cashboxesIds.length) {
+    throw new BadRequest('ERRORS.BAD_REQUEST');
```
jniles: In the future, it would be good to include a text for our logs describing why the error happened.  I think `BadRequest` accepts multiple parameters:
```js
throw new BadRequest('ERRORS.BAD_REQUEST', 'You must have a start and end date defined and provide at least one cashbox id.');
```
This is purely to help us debug any errors - I don't think it is shown to the user.

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +169,241 @@ function reportByService(req, res, next) {
...
+      data.cashLabelDetails = data.cashboxes.map(cashbox => `${cashbox.account_number} - ${cashbox.account_label}`);
+
+      // get oposite lines to cashbox accounts in transactions as details
+      return getDetailsIdentifiers(data.cashAccountIds, dateFrom, dateTo);
+    })
+    .then(rows => {
+      data.detailsIdentifiers = rows.map(row => row.uuid);
+
+      // build periods columns from calculated period
+      return Fiscal.getPeriodsFromDateRange(data.dateFrom, data.dateTo);
```
jniles, line 231: Nice 👍

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +169,241 @@ function reportByService(req, res, next) {
...
+    })
+    .then(periods => {
+      data.periodDates = periods.map(p => p.start_date);
+
+      data.periods = periods.map(p => p.id);
+
+      data.colspan = data.periods.length + 1;
```
jniles, line 238: Good name 👍

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +169,241 @@ function reportByService(req, res, next) {
...
+
+      // skip period matrix if not transactions were identified
+      if (data.detailsIdentifiers.length === 0) { return []; }
+
+      // build periods string for query
+      const periodString = data.periods.length ? data.periods.map(periodId => {
+        return `SUM(CASE WHEN source.period_id = ${periodId} THEN source.balance ELSE 0 END) AS "${periodId}"`;
```
jniles: This is correct and good SQL.  Just FYI - I like to use `IF()` because it makes the code more readable like this:
```js
const periodString = data.periods.length ? data.periods.map(periodId => {
  return `SUM(IF(source.period_id = ${periodId}, source.balance, 0)) AS ${periodId}`;
});
```

I know this comes from the database, but I would encourage you to always use `?` syntax to pass in parameters to SQL.  At some point, we'll make a mistake and have a terrible SQL injection error that [ends us up on international news](http://www.irinnews.org/news/2018/01/18/exclusive-audit-exposes-un-food-agency-s-poor-data-handling).

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +169,241 @@ function reportByService(req, res, next) {
...
+            a.number AS account_number, a.label AS account_label,
+            SUM(gl.debit_equiv - gl.credit_equiv) AS balance,
+            tt.id AS transaction_id, tt.type AS transaction_type,
+            MAX(tt.text) AS transaction_text,
+            gl.period_id, gl.account_id
+          FROM general_ledger gl
+          JOIN account a ON gl.account_id = a.id
+          JOIN transaction_type tt ON gl.origin_id = tt.id
+          WHERE gl.uuid IN ? 
+          GROUP BY a.id, tt.id
```
jniles, line 265: Do you need this GROUP BY?  It looks like you only need to do one GROUP BY - the outer one.  What happens if you remove the `SUM(gl.debit_equiv - gl.credit_equiv) as balance` and just use `(gl.debit_equiv - gl.credit_equiv) as balance`?   Would that give you the same result?

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +169,241 @@ function reportByService(req, res, next) {
...
+        ) AS source 
+        GROUP BY transaction_type, account_id;
+      `;
+      return db.exec(query, [[data.detailsIdentifiers]]);
+    })
+    .then(rows => {
+      // split incomes from expenses
+      const incomes = _.chain(rows).filter({ transaction_type : 'income' }).groupBy('transaction_text').value();
```
jniles: I haven't tried this yet, but will `_.partition` work ([docs](https://lodash.com/docs/#partition))?   It looks like you should be able to do something like this:
```js
const groups = _.partition(rows, row => transaction_type);
// this gives me an array of arrays [[...transactions with one type], [transactions with another type], ...]
// now we need to figure out what transaction type is in which array (they could be in any order).

// let's loop through the groups and convert them to a map of transaction type -> group
const mapping = {}
groups.forEach(group => {
   const [firstElementOfGroup] = group;
   mapping[firstElementOfGroup.transaction_type] = group;
});

// now we can do this:
const incomes = mapping.income.groupBy('transaction_text');
const expenses = mapping.expense.groupBy('transaction_text');
const transfers = mapping.transfer.groupBy('transaction_text');
```

It might also be more performant to use `groupBy('transaction_type_id')` instead of the `transaction_text`.  Comparing numbers is faster than comparing strings.

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -753,3 +169,241 @@ function reportByService(req, res, next) {
...
+     * aggregateTotalByKeys
+     *
+     * this function process totals for incomes or expense by transaction type
+     * @param {*} source
+     * @param {*} sourceTotalByTextKeys
+     */
+  function aggregateTotalByTextKeys(source = {}) {
+    const sourceTotalByTextKeys = {};
+
+    _.keys(source).forEach((index) => {
```
jniles, line 324: Just FYI, this should work:
```js
_.forEach(source, (currentTransactionText, index) => {
  // code...
});
```

jniles (approved):
LGTM.  I made a few suggestions, you can either respond or "bors r+" yourself, depending on how tired you are of working on this PR!

Great Work! 👍

server/controllers/finance/reports/cashflow/index.js
mbayopanda: Yes, it is required because the query have to split accounts by transaction type (transaction_type.text : purchase, sale, cash_payment, ...).

We have to notice that the `transaction_type` which is outside correspond to transaction_type.type which can be income, expanse or transfer.

If we remove this `GROUP BY a.id, tt.id` we will get different results.

server/controllers/finance/reports/cashflow/index.js
mbayopanda: The idea here is to throw an error message if `dateFrom`, `dateTo` or `cashboxesIds` are missing.

server/controllers/finance/reports/cashflow/index.js
mbayopanda: Compare it to this : 
```js
const incomes = _.chain(rows).filter({ transaction_type : 'income' }).groupBy('transaction_text').value();
const expenses = _.chain(rows).filter({ transaction_type : 'expense' }).groupBy('transaction_text').value();
const transfers = _.chain(rows).filter({ transaction_type : 'transfer' }).groupBy('transaction_text').value();
```
It seems like there are a lot of code with your proposal.

jniles: But much less loops for the poor Rsapberry Pi!

test/integration/reports/rendering.js
```diff
@@ -14,38 +14,38 @@ module.exports = function LazyTester(target, keys, options = {}) {
...
+    it(`GET ${target} should return HTML data for 'html' rendering target`, () => {
       return agent.get(target)
         .query(html)
         .then(expectHTMLReport)
         .catch(helpers.handler);
     });
 
-    it(`GET ${target} should return PDF data for 'pdf' rendering target`, function () {
+    it(`GET ${target} should return PDF data for 'pdf' rendering target`, () => {
       this.timeout(5000);
```
jniles: Remember that `this` isn't defined in a "fat arrow" function.  This is what is breaking your two tests.

You should keep the `function () {` instead of going to `() => {`.

mbayopanda: 👍

jniles (approved):
Code LGTM.  Some very nice improvements!  Good variable naming, and well done dropping the code from 922 lines to 464.  That's awesome!

bors r+

PR #2579, opened 2018-03-02
feat(General Ledger): add totals on PDF printout

jniles:
This commit adds totals to the PDF print out, as well as ensuring that the depth is correctly computed.  Due to API changes with the tree library, the depth was removed - this restores the calculation.  It also styles credits following the standard credit style.

See the attached document for a printout of what it looks like.
[General Ledger 2018-3-2.pdf](https://github.com/IMA-WorldHealth/bhima-2.X/files/1774519/General.Ledger.2018-3-2.pdf)

### DIFF
--- a/server/controllers/finance/reports/generalLedger/index.js
+++ b/server/controllers/finance/reports/generalLedger/index.js
@@ -9,6 +9,7 @@
 const _ = require('lodash');
 const ReportManager = require('../../../../lib/ReportManager');
 const GeneralLedger = require('../../generalLedger');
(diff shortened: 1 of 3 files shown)


jniles:

jniles:
This shouldn't be too controversial.

bors r+

jniles:
Looks like we are being hit by https://bugzilla.mozilla.org/show_bug.cgi?id=1425706.

jniles:
Seems to be fixed now!  #2603 

bors r+

PR #2582, opened 2018-03-05
feat(debtor group) make color field optional

### LINKED ISSUE #2333: Move "Color" option lower in the Debtor Group Create Form
We implemented a nice color dropdown for picking colors to assign to a debtor group.  Unfortunately, it is the second option on the form, making it seem really important.  Even more important than an account!  See below:
movecolorlowerprioritty
_Fig 1: Color Is Number 2!_

This isn't a good user experience - if they wish to skip the color, they must fill out first field, then skip another field, then fill out the required account field.  We should move the "color" option to be one of the last in the form to emphasize that it is optional.

jeremielodi:
This PR move the color field to bottom because it is no required
closes #2333

### DIFF
--- a/client/src/modules/debtors/groups.update.html
+++ b/client/src/modules/debtors/groups.update.html
@@ -19,20 +19,6 @@
             </div>
           </div>
 
-          <div
-            class="form-group has-feedback"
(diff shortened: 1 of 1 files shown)


jniles (approved):
LGTM.


bors r+

PR #2583, opened 2018-03-05
refractor(home page) exchange rate

### LINKED ISSUE #2562: Home Page Exchange Rate sometimes shows incorrect date
The exchange rate component on the main page sometimes shows the incorrect date.  It may also show the incorrect rate at times too. See the screenshot below:

badexchangerate
_Fig 1: Exchange Rate Date?_

This is likely due to a request which does not order the exchange rate values by date, but instead just queries them in the order they were written to the database.

jeremielodi:
This PR solves the issue related at #2562
closes #2562

### DIFF
--- a/client/src/modules/home/home.ctrl.js
+++ b/client/src/modules/home/home.ctrl.js
@@ -3,7 +3,7 @@ angular.module('bhima.controllers')
 
 HomeController.$inject = [
   'CurrencyService', 'ExchangeRateService', 'SessionService',
-  'NotifyService', 'FiscalService', 'DashboardService', 'moment'
+  'NotifyService', 'FiscalService', 'DashboardService', 'moment',
 ];
 
 /**
@@ -52,18 +52,18 @@ function HomeController(Currencies, Rates, Session, Notify, Fiscal, DashboardSer
     })
     .then(() => {
       vm.currencies.forEach((currency) => {
-        currency.rate = Rates.getCurrentRate(currency.id);
-
-
+        const exchange = Rates.getCurrentExchange(currency.id);
+        currency.rate = exchange.rate;
+        currency.date = exchange.date;
         /*
           Let check is the currency rate is lower the 1
           so that we could format it in a readable way
         */
+
         if (currency.rate < 1) {
           currency.rate = (1 / currency.rate);
           vm.isFirstCurencyLabel = true;
         }
-
 
         currency.formattedDate = new Moment(currency.date).format('LL');
       });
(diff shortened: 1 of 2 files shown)


client/src/modules/home/home.ctrl.js
```diff
@@ -39,45 +40,40 @@ function HomeController(Currencies, Rates, Session, Notify, Fiscal, DashboardSer
...
-
     })
     .catch(Notify.handleError);
 
   Fiscal.fiscalYearDate({ date : vm.today })
-    .then(function (year) {
+    .then((year) => {
       vm.year = year[0];
-      vm.year.percentage = vm.year.percentage * 100;
+      vm.year.percentage *= 100;
```
jniles: :dancer:

server/controllers/finance/exchange.js
```diff
@@ -73,6 +84,33 @@ function getExchangeRateList(enterpriseId, opts) {
...
+    FROM exchange_rate ex_r
+    JOIN enterprise ON enterprise.id = ex_r.enterprise_id
+    WHERE ex_r.enterprise_id = ? and ex_r.currency_id = ?
+    ORDER BY date DESC
+    LIMIT 1
+  `;
+
+  return db.exec(sqlCurrency, enterpriseId).then(IDs => {
+    return q.all(IDs.map(currency => {
+      return db.one(sqlExchangeRate, [enterpriseId, currency.id]);
```
jniles: Did you know you can use the SQL function [GetExchangeRate](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/functions.sql#L47) from here?  It might be easier.  I think you can do it in a single SQL statement.
```sql
SELECT currency.id, GetExchangeRate(?, currency.id, NOW()) as rate, NOW()) as date
FROM currency WHERE currency.id NOT IN (SELECT currency_id FROM enterprise WHERE id = ?);
````
You would then call it from JS like this:
```js
return db.exec(sql, [enterpriseId, enterpriseId]);
```

server/config/routes.js
```diff
@@ -123,6 +123,7 @@ exports.configure = function configure(app) {
 
   // exchange rate modules
   app.get('/exchange', exchange.list);
+  app.get('/exchange_rates', exchange.rates);
```
jniles: Is this ever used?  It doesn't look like it is ever called from the client?

jniles (approved):
LGTM.

bors r+

PR #2584, opened 2018-03-06
Unit test for pdf rendering

mbayopanda:
This PR test the PDF generation with `wkhtmltopdf` and `pdf.render`.


test/server-unit/translate.spec.js
```diff
@@ -41,6 +41,11 @@ function TranslateUnitTests() {
     const compiled = translate('en')(key);
     expect(compiled).to.equal(key);
   });
+
+  it('#translate() should return undefined for undefined value', () => {
```
jniles, line 45: :+1:

test/server-unit/pdf.spec.js
```diff
@@ -0,0 +1,101 @@
...
+
+    // give the giant string to the render method
+    const rendered = await pdf.render({}, tmpl, {});
+    const cached = await fs.readFile(generatedFile);
+
+    expect(isBufferInstance(rendered)).to.equal(true);
+    expect(isBufferInstance(cached)).to.equal(true);
+
+    // pdf DateCreation must be ignored when comparing
+    const slicedRendered = sliceOutCreationDate(rendered);
```
jniles, line 56: 👍

test/server-unit/pdf.spec.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const fixturesPath = path.resolve('test/fixtures');
+const htmlFile = path.join(fixturesPath, '/pdf-sample.html');
+const generatedFile = path.join(fixturesPath, '/pdf-sample.pdf');
+
+
+function PDFRenderUnitTest() {
+  it('#wkhtmltopdf() creates correctly a PDF file from an HTML', async () => {
+    const wk = await exec(`wkhtmltopdf ${htmlFile} ${generatedFile}`);
```
jniles: Can you make this a random file name and then remove it at the end of the test?

jniles (commented):
Looks excellent!

I think the PDF file will be overwritten multiple times and have git will try to track those changes.  Could you make the generated file be a random file name and then clean it up?

Alternatively, you can make it write to a temporary directory using [tempy](https://www.npmjs.com/package/tempy) or something similar. :)

jniles:
I think the most likely reason this is failing is a different screen resolution between machines.  I wonder if there is any way of changing that ...   maybe by specifying the page size?

We'll keep working on the Travis upload to see how it works.

jniles:
I'm going to move this to a shared branch on the upstream repo.

PR #2587, opened 2018-03-07
fix(bhNavigation): update tree on navigation delta

### LINKED ISSUE #2585: Update the Tree highlight on navigation
There is a really annoying bug in the system that doesn't update the Tree highlight when navigated.  This is really confusing for users who are automatically navigated to different modules by links in registries.  See below:

errorintreenavigation
_Fig 1: This should say Journal_

jniles:
This commit causes the tree to be updated when the user navigates between states.

Closes #2585


jniles (commented):
bors r+

PR #2592, opened 2018-03-08
fix(Trial Balance): show loading indicator

jniles:
This commit fixes a bug in the Trial Balance where the loading indicator was not shown on submit.


jniles:

PR #2594, opened 2018-03-09
feat(journal) Remove the Period column from the Journal

### LINKED ISSUE #2586: Remove the Period column from the Journal
The journal has a "period" column that is heavy to render, but of limited value.  See the image below:
removeperiodfromjournal
_Fig 1: Period Column_

I don't think anyone uses this in production.  I propose we remove it.

jeremielodi:
closes #2586

### DIFF
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -169,13 +169,8 @@ function JournalController(
     headerCellFilter : 'translate',
     visible : false,
     enableCellEdit : false,
-  }, {
-    field : 'period_end',
-    displayName : 'TABLE.COLUMNS.PERIOD',
-    headerCellFilter : 'translate',
-    cellTemplate : 'modules/templates/bhPeriod.tmpl.html',
-    visible : false,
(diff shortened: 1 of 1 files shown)


client/src/modules/journal/journal.js
```diff
@@ -169,13 +169,8 @@ function JournalController(
     headerCellFilter : 'translate',
     visible : false,
     enableCellEdit : false,
-  }, {
-    field : 'period_end',
-    displayName : 'TABLE.COLUMNS.PERIOD',
-    headerCellFilter : 'translate',
-    cellTemplate : 'modules/templates/bhPeriod.tmpl.html',
```
jniles, line 176: Does any module still use this template?  If not, can it be removed?

jniles: Ah, looks like it is used [here](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/account_statement/account_statement.ctrl.js#L126).

PR #2597, opened 2018-03-13
Replace latest invoice with Financial Activity on patient record page

### LINKED ISSUE #2506: Rewrite loadLatestInvoice() code in patients/index.js
The SQL code to load a patient's latest invoice is quite old and isn't up to our regular standards.  I believe it's origins are from when we used `combined_ledger` before learning the MySQL's views were not INDEXed.  

Someone should go through and rewrite this code.  It can probably be combined with filtering on transaction_type and some smarter parsing to increase the performance robustly.

For reference, [here is a link to the code](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/medical/patients/index.js#L626).

jniles:
This PR implements the Patient Financial Activity component, a replacement for the previous `bhLatestInvoice` component that had ceased to work.  This version pulls from the same data source as the Financial Activity report.

**Interface**
The component used to look like this:
previousfinancialactivity
_Fig 1: Old version_

And now it looks like this:
financialactivityinterface
_Fig 2: New version (with irregularities shown)_

normalfinancialstatus
_Fig 5: Normal Financial Status_

The idea is to present two things:
 1. General statistics + links about the patient's activity.
 2. Warnings/advice about actions that need to be taken if the patient has an irregular financial history.

**Other Features**
To make this look nice, I've also implemented the `bhPanelOverlay` to put a loading indicator.  It looks like this while loading:
loadingbhpaneloverlay
_Fig 3: bhPanelOverlay for loading_

I also use the `bhPanelOverlay` for when there is no data.  See below:
noactivity
_Fig 4: bhPanelOverlay for no activity_.

Also, I fixed a bug with exporting to CSV due to a misuse of `this` in an arrow function.

Closes #2506.


jniles:
It looks like these code climate issues are related to TODO and FIXME.  I'll turn that engine off..

jniles:
Hmm.. silly code climate.

jniles:
@mbayopanda, when you get a chance, can you take a look at this?

sh/test.sh
```diff
@@ -16,16 +16,16 @@ function endfold {
...
 # run karma (client unit) tests
 startfold "Running Client Unit Tests..." "test-client-unit";
 ./node_modules/.bin/karma start --single-run --no-auto-watch --concurrency 1 karma.conf.js
 endfold "test-client-unit" ;
 
+# run integration tests
+startfold "Running Integration Tests..." "test-integration";
+./sh/integration-tests.sh
+endfold "test-integration" ;
+
```
mbayopanda, line 28: by curiosity, is there any reason why the integration test must run after the client unit test (karma) ?

client/src/i18n/fr/patient_records.json
```diff
@@ -1,31 +1,51 @@
...
+    "INVOICE_COVERED_BY":"Cette facture est couverte par",
+    "TITLE":"Patients",
+    "UPDATE_PICTURE":"Mettre à jour la Photo",
+    "UPLOAD_PICTURE":"Télécharger Photo Patient",
+    "NOT_FOUND":"Impossible de trouver le patient ayant ce ID. Veuillez contacter votre administrateur systeme",
+    "DOB_NOT_SPECIFIED":"Seulement une année de naissance a été spécifiée lors de l'enregistrement du patient. C'est une date approximative.",
+    "FINANCES" : {
+      "IN_GOOD_STANDING" : "Tout a l'air normal!",
+      "NOT_IN_GOOD_STANDING" : "Il y a quelques irrégularités dans l'activité financière.",
+      "HAS_CREDITOR_BALANCE" : "Le patient a un solde créditeur (négatif) de <b>({{ balance | currency:currency_id}})</b>. Cela peut être intentionnel, mais peut indiquer un problème.",
```
mbayopanda: Instead of having `créditeur (négatif)` we can just have `créditeur` or `négatif`; because if we read just `créditeur négatif` it is a bit confusing, and we know that `solde créditeur` =  `solde negatif`.

Example : 
`Le patient a un solde créditeur de <b>({{ balance | currency:currency_id}})</b>. Cela peut être intentionnel, mais peut aussi indiquer un problème.`

mbayopanda (approved):
LGTM 👍

sh/test.sh
jniles: In general, unit tests tend to be faster, so I moved up the unit tests as far as I could so they "fail fast" and give us an error as soon as possible if there is a unit test error.

jniles:
@mbayopanda, changes have landed!

jniles:
@mbayopanda, can I get another review?

jniles:
Presumably this is okay.  I'm merging this since it has been a week since review and I'd like to use the balance code for cautions.

PR #2598, opened 2018-03-13
refactor(fiscal year) part2

### LINKED ISSUE #2561: Refactor Fiscal Year Part 2
We did a first refactor of the fiscal year in #2542.  This PR implemented a TreeService that is used to calculate the totals of groups on the client.  It also used a stored procedure to create a fiscal year.

The fiscal year module still needs to be improved!  Here are the things that still need to be updated:
 1. Remove the "Import Beginning Balance from Previous Year" button.  This operation is no longer necessary as closing a Fiscal Year now writes the opening balances of the next fiscal year in the same step.
 2. Remove the code that creates and uses Period Number 13.
 3. Remove the "Clear Button" on the Fiscal Year creation/update form.
fiscalyearclearbutton
_Fig 1: Remove clear button_

**Closing Balance Modal**
The following improvements are specific to the Closing Balance Modal:
 1. Do **not** allow users to select an Income/Expense account as the "resultat" account.
 2. [Feature Request] Allow multiple accounts to be used for the closing procedure.
 3. Right align all currencies/numbers.
 4. Remove the footer text "total".
 5. Use the currency filter to show that the values are currencies.
 6. Remove the green/red coloring.  Some people are green/red color blind.
 7. Order the values in the grid by account number.
 8. Remove the `ui-grid-grouping` tag.
 9. Implement a loading indicator on the submit buttons.
 10. Remove the need to type the fiscal year name on closing.
 8. Does the Closing Balance really need to be in a modal?  Could we make it like the other steps where we use different `ui-router` states?
closingbalancemodal
_Fig 2: Closing Balance Modal Improvements_

**Closing Balance View**
 1. Use the TreeService to calculate totals of parent accounts on load.
 2. Make the account numbers of parent accounts bold.
 3. Implement a footer that sums the values of the entire grid.
 4. If the Fiscal Year is closed, inform the user of this fact.  Show the balances as plain numbers, not disabled text inputs.  Inform the user that to edit the closing balance, they will need to re-open the fiscal year.
closingbalanceview
_Fig 3: Closing Balance View_

jeremielodi:
This PR solves differents problems detailed at #2561 by @jniles for refractoring fiscal year module

closes #2561

### DIFF
--- a/client/src/js/components/bhAccountSelect.js
+++ b/client/src/js/components/bhAccountSelect.js
@@ -5,11 +5,11 @@ angular.module('bhima.components')
     transclude  : true,
     bindings    : {
       accountId        : '<',
-      accountTypeId :  '<',
       onSelectCallback : '&',
       disable          : '<?',
       required         : '<?',
       classe           : '@?',
+      accountTypeId :  '<?',
       label            : '@?',
       name             : '@?',
       excludeTitleAccounts : '@?',
@@ -29,18 +29,11 @@ function AccountSelectController(Accounts, AppCache, $timeout, bhConstants, $sco
   var hasCachedAccounts = false;
   var cache = new AppCache('bhAccountSelect');
 
-  // cache accounts locally for three seconds
-  var CACHE_TIMEOUT = 3000;
-
   // fired at the beginning of the account select
   $ctrl.$onInit = function $onInit() {
 
-    $ctrl.classe = $ctrl.accountTypeId || '';
     // cache the title account ID for convenience
     $ctrl.TITLE_ACCOUNT_ID = bhConstants.accounts.TITLE;
-
-    // To filter accounts by class of accounts
-    $ctrl.classe = $ctrl.classe || '';
 
     // translated label for the form input
     $ctrl.label = $ctrl.label || 'FORM.LABELS.ACCOUNT';
@@ -95,17 +88,22 @@ function AccountSelectController(Accounts, AppCache, $timeout, bhConstants, $sco
 
   // loads accounts from the server
   function loadHttpAccounts() {
-    var detailedRequest = $ctrl.classe ? 1 : 0;
-    var params = { detailed : detailedRequest, classe : $ctrl.classe };
-    if ($ctrl.accountTypeId) {
-      params.type_id = $ctrl.accountTypeID;
+    const detail = $ctrl.accountTypeId || $ctrl.classe;
+    const detailedRequest = detail ? 1 : 0;
+    const params = { detailed : detailedRequest };
+
+    if ($ctrl.classe) {
+      params.classe = $ctrl.classe.split(',').map(num => { return parseInt(num, 10); });
     }
-    
+    if ($ctrl.accountTypeId) {
+      params.type_id = $ctrl.accountTypeId.split(',').map(num => { return parseInt(num, 10); });
+    }
+
     // load accounts
     Accounts.read(null, params)
       .then(function (elements) {
         // bind the accounts to the controller
-        var accounts = Accounts.order(elements);
+        let accounts = Accounts.order(elements);
 
         if ($ctrl.excludeTitleAccounts) {
           accounts = Accounts.filterTitleAccounts(accounts);
--- a/client/src/js/components/bhDateInterval.js
+++ b/client/src/js/components/bhDateInterval.js
@@ -22,6 +22,7 @@ angular.module('bhima.components')
       dateId : '@?', // date identifier
       required : '<?', // true or false
       onChange : '<?', // on change action
+      canClear : '<?', // flag for displaying clear button
       label : '@?',
       mode : '@?', // the date mode (day|month|year)
     },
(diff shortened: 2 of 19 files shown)


client/src/js/components/bhAccountSelect.js
```diff
@@ -5,11 +5,11 @@ angular.module('bhima.components')
     transclude  : true,
     bindings    : {
       accountId        : '<',
-      accountTypeId :  '<',
       onSelectCallback : '&',
       disable          : '<?',
       required         : '<?',
       classe           : '@?',
+      accountTypeId :  '<',
```
jniles: If this isn't required in all cases, it should be `<?`

client/src/js/components/bhAccountSelect.js
```diff
@@ -29,19 +29,15 @@ function AccountSelectController(Accounts, AppCache, $timeout, bhConstants, $sco
...
   $ctrl.$onInit = function $onInit() {
 
-    $ctrl.classe = $ctrl.accountTypeId || '';
     // cache the title account ID for convenience
     $ctrl.TITLE_ACCOUNT_ID = bhConstants.accounts.TITLE;
 
     // To filter accounts by class of accounts
-    $ctrl.classe = $ctrl.classe || '';
-
+    $ctrl.classe = $ctrl.classe || undefined;
```
jniles: If both of these are optional, they should be set to `undefined` by default.  What happens if you remove these two lines?

client/src/js/components/bhDateInterval.js
```diff
@@ -22,6 +22,7 @@ angular.module('bhima.components')
       dateId : '@?', // date identifier
       required : '<?', // true or false
       onChange : '<?', // on change action
+      canClear : '<', // flag for displaying clear button
```
jniles: This should be optional with the `<?` flag.

client/src/js/components/bhDateInterval.js
```diff
@@ -38,8 +39,10 @@ function bhDateInterval(moment, bhConstants) {
   vm.search = search;
   vm.clear = clear;
 
-
   vm.$onInit = function $onInit() {
+    // specify if clear button can be displayed
```
jniles, line 43: Nice comment 👍

client/src/js/components/bhDateInterval.js
```diff
@@ -38,8 +39,10 @@ function bhDateInterval(moment, bhConstants) {
...
+    vm.canClear = !angular.isUndefined(vm.canClear) ? Boolean(vm.canClear) : true;
```
jniles: Also, I don't think the `Boolean` call does anything.  Try this in your console:
```js
var f = 'false';
console.log(Boolean(f)); // => true
```

Since you use `false` later, that means it is passing it in by value, not as a string to parse.  You can probably just do:
```js
vm.canClear = (vm.canClear || true);
```
to the same effect.

client/src/modules/fiscal/templates/credit.cell.tmpl.html
```diff
@@ -0,0 +1 @@
+<div class="ui-grid-cell-contents text-right">{{row.entity.credit}}</div>
```
jniles: There is no need to make a separate template for this - it will only slow down rendering.  Just add `cellClass : 'text-right'` to your columDef in the controller.

server/models/procedures/time_period.sql
```diff
@@ -75,8 +75,8 @@ BEGIN
     fyPreviousFYId, fyLocked, fyCreatedAt, fyUpdatedAt, fyUserId, fyNote
   FROM fiscal_year WHERE id = fiscalYearId;
 
-  -- insert N+2 period
-  WHILE periodNumber <= fyNumberOfMonths + 1 DO
+  -- insert N+1 period
+  WHILE periodNumber <= fyNumberOfMonths DO
 
     IF periodNumber = 0 OR periodNumber = fyNumberOfMonths + 1 THEN
```
jniles: You can remove the second part of this OR.  It will never get reached now.

jniles (commented):
@jeremielodi pretty good!

I've made some comments below.  In general, it is better to use `cellClass` and `cellFilter` on ui-grids instead of making a new html template.  This speeds things up a bit.

Otherwise, it all looks pretty good to me.  Let me know if you have any questions about my feedback.

test/end-to-end/fiscalYears/fiscalYears.spec.js
```diff
@@ -71,52 +71,6 @@ describe('Fiscal Year', () => {
     components.notification.hasSuccess();
   });
 
-  it('set the opening balance for the first fiscal year', () => {
```
jniles: Out of curiosity, why did this test get removed?  Is there a replacement test for it?

jniles (commented):
@jeremielodi, this is a step in the right direction.  It doesn't address all of the issues in #2561, though so I will removed the "closes" tag if we wish to merge  We still need to:
 1. Use the TreeService to calculate the balance of parent accounts based on their children.
 2. Make the closing balance modal use a tree.
 5. Remove the ui-grid-grouping tag on the closing balance modal.
 3. Remove the disabled inputs from the "view closing balance" page.
 4. Put a footer on the closing balance page grid.

etc ...

Would you like to continue working on those issues, or do you want me to merge this as-is?

jniles:
Looks okay to me.  We'll get these in and then continue work.

bors r+

PR #2607, opened 2018-03-20
feat(enterprise): add settings panel

### LINKED ISSUE #2174: Adding Enterprise Customizations
This issue is to track settings that it would be ideal for enterprises to be able to customize.  It will be an ongoing list:
 1. Report Headers - They should be able to upload their own HTML header to the reports and not use the BHIMA default.
 2. Lock price changes - An administrator should be able to enforce stricter rules around pricing - that price changes cannot occur without the administrators approval at invoicing time.
 3. ~~Require Long Passwords - For security reasons, longer passwords are better than shorter (in general).  Sites like Github give enterprises the ability to enforce two factor authentication.  We could optionally let the administrator enforce long passwords.~~

This list will probably be evolving for some time as ideas are generated.

jniles:
This commit adds a new panel and CRUD interface for operation on
enterprise settings.  The settings table is linked 1-1 to the enterprise
table.  Tests have been updated accordingly.  The projects service has
also been updated to use the PrototypeApiService.

Closes #2174.

### DIFF
--- a/client/src/i18n/en/enterprise.json
+++ b/client/src/i18n/en/enterprise.json
@@ -1,8 +1,17 @@
-{"ENTERPRISE":{"ADD_ENTERPRISE":"Add Enterprise.",
-"ALL_ENTERPRISES":"All Enterprises",
-"CONFIGURATION_ENTERPRISE":"Configuration Enterprise",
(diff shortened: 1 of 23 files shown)


jniles:
~~Note, this PR only sets up the infrastructure for enterprise customization.  It doesn't add the business logic for price change locks yet.~~

Edit: the price lock functionality has been implemented.

PR #2608, opened 2018-03-21
refractor(stok managent) improvements

### LINKED ISSUE #2600: Improvements to Stock Management
I did a bit of stock management on my local machine and ran into the following minor usability issues. 

**Stock Entry and Exit**
 1. The "can't find your depot?" message should mention that you must have _permission_ to use the depot.  It can exist, but the user not have permission.
 2. After submitting, do not clear the date!  It should reset to the current date, or use the previously used date.
 3. The "cancel" button doesn't work at all.  It should clear everything in the form or reset it to the defaults.
 4. After submitting, the form should be reset using `$setPristine()` so that errors do not show.  For example, the description lights up red as an error, even though no error occurred.

**Stock Lots Registry**
 1. Add group of inventory to "lots in stock" search.  It would be helpful to print out all computers, for example.
 2.  Is the "total" at the bottom of the quantity column helpful?  I propose we remove it to reduce confusion - it is not the total number of lots in the system, but the sum of all quantities across all lots.  Do you really have 21 items if you have two bottles of 10 pills and one computer?

jeremielodi:
closes #2600

### DIFF
--- a/client/src/i18n/en/depot.json
+++ b/client/src/i18n/en/depot.json
@@ -15,6 +15,7 @@
       "TITLE" : "Depot Management"
     },
     "NO_DEPOT" : "No Depot",
+    "NO_DEPOT_ASSIGNED_TO_USER" : "You are not allowed to access any depot or there's no depot registered yet. Please contact the administrateur if needed",
     "NOT_ALLOWED_ACTIONS" : "This depot is protected.  You cannot modify its properties.",
     "TITLE" : "Depot Management",
     "UPDATED" : "Depot updated with success",
@@ -25,7 +26,7 @@
       "NO_DEPOTS" : "Do not see any depots?",
       "SELECT_A_DEPOT" : "Select a depot",
       "NO_DEFAULT_DEPOT" : "You need to choose a depot.",
-      "SELECTION_EXPLANATION" : "It looks like you don't have a default depot set!  Please select your depot from the list below.  You will only need to do this once.  If your depot is not listed below, you can add it with this link:"
+      "SELECTION_EXPLANATION" : "It looks like you don't have a default depot set!  Please select your depot from the list below.  You will only need to do this once."
     }
   }
 }
--- a/client/src/modules/stock/depot-selection.modal.html
+++ b/client/src/modules/stock/depot-selection.modal.html
@@ -7,46 +7,41 @@
 
 <div class="modal-body">
 
-  <!-- link to depots page if depots not found -->
-  <p class="text-info">
-    <i class="fa fa-question-circle-o"></i>
-    <span translate>DEPOT.MODAL.NO_DEPOTS</span>
-
-    <a href ui-sref="depots.create">
-      <i class="fa fa-link"></i> <span translate>DEPOT.ADD_DEPOT</span>
-    </a>
-  </p>
-
-  <div uib-alert class="alert-warning" ng-if="$ctrl.hasNoDefaultDepot">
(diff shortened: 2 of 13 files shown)


client/src/modules/stock/depot-selection.modal.html
```diff
@@ -7,46 +7,41 @@
...
+  <div uib-alert class="alert-warning" ng-if="$ctrl.hasNoDefaultDepot && $ctrl.depots.length>0">
     <p>
       <strong> <i class="fa fa-warning"></i> <span translate>DEPOT.MODAL.NO_DEFAULT_DEPOT</span></strong>
       <br />
       <span translate>DEPOT.MODAL.SELECTION_EXPLANATION</span>
-      <strong><a ui-sref="depots.create"><b><span translate>DEPOT.ADD_DEPOT</span></b></a></strong>.
     </p>
   </div>
 
+  <div uib-alert class="alert-warning" ng-if="$ctrl.depots.length===0">
```
jniles, line 18: Note that, if there are no depots in the enterprise, this will give a slightly misleading message.  There is also the problem that you cannot exit the modal if you trigger it anymore - the link used to be a hack to work around that.  We should probably just fix the origin of the problem them - presenting links only to people who can use them.

client/src/modules/stock/entry/entry.html
```diff
@@ -101,8 +101,8 @@
       <div class="row" style="margin-top: 5px;">
         <div class="col-xs-6 col-xs-offset-6">
           <div class="text-right">
-            <button class="btn btn-default" ng-click="StockCtrl.suspend(StockEntryForm)" type="button" translate>
-            FORM.BUTTONS.SUSPEND
+            <button class="btn btn-default" ng-click="StockCtrl.reset(StockEntryForm)" type="button" translate>
+            FORM.BUTTONS.CLEAR
```
jniles, line 105: 👍

client/src/modules/stock/lots/modals/search.modal.js
```diff
@@ -61,6 +61,12 @@ function SearchLotsModalController(data, Inventory, Notify, util, Store, Instanc
     }
   };
 
+  // custom filter group_uuid - assign the value to the params object
+  vm.onSelectGroup = (group) => {
```
jniles, line 65: 👍

jniles (approved):
Code looks good to me.  Could you write an end-to-end test that filters on `group_uuid` to make sure it continues working for all time?

client/src/modules/stock/depot-selection.modal.html
jeremielodi: in that case I can just correct the message to help the user to know that he has no permission on a depot or there is no depot enregistered yet, he can cantact the administrator

jniles: That sounds like a good idea :)

client/src/i18n/en/depot.json
```diff
@@ -15,6 +15,7 @@
       "TITLE" : "Depot Management"
     },
     "NO_DEPOT" : "No Depot",
+    "NO_DEPOT_ASSIGNED_TO_USER" : "You are not allowed to access any depot or there's no depot registered yet. Please contact the administrateur if needed",
```
jniles, line 18: :+1:

jniles (approved):
LGTM.

bors r+

jniles:
Seems like a fluke. 

bors r+

PR #2610, opened 2018-03-21
refactor(vouchers): polish registry

### LINKED ISSUE #2412: Improvements to the Voucher Registry
The voucher registry will be used to validate pretty much all writings for the main cashboxes, the bank accounts, and other account to account transfers.  Here are some improvements we should add:
 1. Filter by account number
 2. Filter by Entity Reference.
 3. Filter by Document Reference.

In my experience, the grouping on the column headers is confusing.  A much more straightforward way to do grouping is like that in the Posting Journal and Account Statement with an activated button.  I propose we remove all column menus for simplicity.

Also, the action column should be hidden in the column configuration page.

jniles:
This commit polishes the registry by removing the grid grouping values, updating the code to ES6 and fixing the voucher receipt link.

Closes #2412.


jniles:
Hmm... I should probably skip that topic test.  We don't use Topic anyway.

bors r+

PR #2612, opened 2018-03-21
Refactor Cashflow By Service

### LINKED ISSUE #2524: Improvements to the Cashflow by Service report
The "Cashflow by Service" report is in need of a refresh.  Here are some suggestions for improvements:
 1. Slant the totals so that they take up less horizontal space.
 2. Ensure the report has page numbers and prints out in landscape mode, rather than portrait.
 3. Use the enterprise currency for the cells.
 4. Allow display of the totals in multiple currencies.

jniles:
It seems like our clients want the Cashflow by Service report.  This PR brings it into the modern age.
 1. Headers are now slightly slanted to save space. It isn't perfect, but it works well enough for me.  Anyone else can render an opinion if they like.
 2. Client codes has been migrated to ES6 and passes ESLint.
 3. Bugs with the cash reference number fixed by using `document_map`.
 4. The FC currency no longer breaks across multiple lines.
 5. The currency is correctly set to the currency of the cash transaction.
 6. The report is renamed in French to be "Journal de Ventilation".

Closes #2524.


jniles:
This should be reviewed by AG Neron before merge.

server/lib/template/helpers/finance.js
```diff
@@ -9,15 +9,15 @@ const FC_FMT = {
...
+  format : '%v&nbsp;%s', // value before symbol
 };
 
 /** @todo use the currency filter fork written for the client to perform the same behaviour here */
 function currency(value = 0, currencyId) {
   // if currencyId is not defined, defaults to USD.
   // @TODO - super-hardcoded values for the moment.  Can we do better?
   const fmt = (Number(currencyId) === 1) ? FC_FMT : USD_FMT;
-  return accountingjs.formatMoney(value, fmt);
+  return new Handlebars.SafeString(accountingjs.formatMoney(value, fmt));
```
mbayopanda, line 20: 👍

mbayopanda (approved):
LGTM

bors r+

PR #2615, opened 2018-03-23
feat(account statement) download as excel

### LINKED ISSUE #2591: Add option to export Account Statement Report to Excel
We have received the following request from a client:

The Account Statement Report needs to be available as an export to Excel.  Currently, you can only export it to PDF.  This would be under the "save as" feature of the reports page.

jeremielodi:
Adds export to XLS to the Account Statement module.


jniles (approved):
Technically, the issue was for the Account Statement Report, not the account statement module.  This a good PR so we will merge it, but it doesn't address Dr. Rice's concern, since the Account Statement Report is different from the Account Statement module (it has exchange rates on each line and a running balance).

Issue #2595, opened 2018-03-09
Filters not updating displayValues on multiple searches


jniles:
It looks like most filters in the filterbar are not updating their values when the value has been set, then changed.  For example:
 1. Go to the Patient Registry.
 2. Search for "bob".  The filter will be set to "bob".
 3. Search for "jane".  The module will load "jane", but the display value will still be "bob"

This affects most modules I have tested.

PR #2616, opened 2018-03-26
fix: correct filter display values

jniles:
The search filters must update their display values if the underlying value changes.  Previously, the default case was to preserve display values, but it didn't take into account the changes in the underlying value, leading to display values being out of sync with the values.

This was reported in #2595.

### DIFF
--- a/client/src/js/services/util.js
+++ b/client/src/js/services/util.js
@@ -27,8 +27,6 @@ function UtilService(moment) {
     }
   }
 
-
-
   /** @todo comments showing usage */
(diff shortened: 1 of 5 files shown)


jniles:
I've tested manually the following conditions:
 1. Add a filter, submit, update it, submit.  The filter's display value should change.
 2. Add a filter A, submit, add a filter B, submit.  Filter A's display value should always stay the same.
 3. Add a filter A and B.  Refresh the page.  Both A and B should restore their display values.
 4. Navigating via a ui-sref should use the provided display value.

bors r+

PR #2622, opened 2018-03-27
feat(report) layout orientation configuration

### LINKED ISSUE #2294: Allow users to specify landscape or portrait registry printing
Our registries have been developed by different developers for different purposes and all have a different default set of configurations.  One configuration is the landscape/portrait mode of printing.  We should put this option in the hands of our users so that they can decide how they want their registries printed.

jeremielodi:
This PR help users to specify the report's layout orientation 
closes #2294

### DIFF
--- a/client/src/js/components/bhRenderOption.js
+++ b/client/src/js/components/bhRenderOption.js
@@ -1,19 +1,18 @@
 angular.module('bhima.components')
   .component('bhRenderOption', {
     templateUrl : '/modules/templates/bhRenderOption.tmpl.html',
-    controller  : RenderOptionController,
-    bindings    : {
-      orientation         : '=', // two-way binding
+    controller : RenderOptionController,
+    bindings : {
+      orientation : '<', // two-way binding
+      onChange : '&',
     },
   });
 
 
 /**
- * bhDateEditor Component
+ * bhRenderOption Component
  *
- * A component to deal with date, it lets a user choose a date by either typing
- * into an <input> or clicking a calendar dropdown.  It wraps the
- * uib-date-picker to provide the dropdown calendar functionality.
+ * used for specifying the documents properties (layout, ...)
  *
  * @example
  * <bh-render-option
@@ -27,6 +26,9 @@ function RenderOptionController() {
 
   this.$onInit = function $onInit() {
     ctrl.orientation = ctrl.orientation || 'portrait';
+    ctrl.onChange = ctrl.onChange || angular.noop();
+    ctrl.setOrientation = () => ctrl.onChange({ orientation : ctrl.orientation });
   };
 
+   
 }
--- a/client/src/js/components/bhRenderOption.js
+++ b/client/src/js/components/bhRenderOption.js
@@ -3,7 +3,7 @@ angular.module('bhima.components')
     templateUrl : '/modules/templates/bhRenderOption.tmpl.html',
     controller : RenderOptionController,
     bindings : {
-      orientation : '<', // two-way binding
+      orientation : '<', // one-way binding
       onChange : '&',
     },
   });
@@ -23,12 +23,9 @@ angular.module('bhima.components')
  */
 function RenderOptionController() {
   const ctrl = this;
-
   this.$onInit = function $onInit() {
     ctrl.orientation = ctrl.orientation || 'portrait';
     ctrl.onChange = ctrl.onChange || angular.noop();
     ctrl.setOrientation = () => ctrl.onChange({ orientation : ctrl.orientation });
   };
-
-   
 }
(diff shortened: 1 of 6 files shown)


jniles:
@jeremielodi this is an okay start.

This should be implemented on all reports.  Could you make it a component so that the code is shared an then put it on every report?

We also shouldn't really merge code without tests.  Could you write a test or two to show that this works?

jeremielodi:
@jniles, all reports use ```save as``` functionality, right?

jniles:
Aha, I didn't test it in the browser.  Yes you are right.  This should also affect the "print" button though - it should be on the report's main page.  I didn't notice that it was only on the save form.

jeremielodi:
@jniles, how about this PR

client/src/js/components/bhRenderOption.js
```diff
@@ -0,0 +1,32 @@
...
+    templateUrl : '/modules/templates/bhRenderOption.tmpl.html',
+    controller  : RenderOptionController,
+    bindings    : {
+      orientation         : '=', // two-way binding
+    },
+  });
+
+
+/**
+ * bhDateEditor Component
```
jniles: You should probably remove all the comments about `bhDateEditor`.  They don't apply here.

client/src/js/components/bhRenderOption.js
```diff
@@ -0,0 +1,32 @@
+angular.module('bhima.components')
+  .component('bhRenderOption', {
+    templateUrl : '/modules/templates/bhRenderOption.tmpl.html',
+    controller  : RenderOptionController,
+    bindings    : {
+      orientation         : '=', // two-way binding
```
jniles: In the future, it would be better to use a combination of a one-way binding `<` and a callback `&`, as recommended by the AngularJS component documentation.  This is fine for this PR since you didn't know at the start.

client/src/modules/templates/bhRenderOption.tmpl.html
```diff
@@ -0,0 +1,14 @@
+<div id="orientation" class="form-group">
+  <label class="control-label" translate>FORM.LABELS.PRINT_ORIENTATION</label>
+  <div>
+    <label class="radio-inline">
+      <input type="radio" name="portrait" id="portrait" ng-true-value='portrait' value="portrait" ng-model="$ctrl.orientation" required>
```
jniles: According to the [AngularJS documentation](https://docs.angularjs.org/api/ng/input/input%5Bradio%5D), `ngTrueValue` is only defined for `input[type="checkbox"]`.  You should remove it from these `input[type="radio"]` buttons.

jniles (commented):
Sorry, didn't realize work had finished on it.  Thanks for letting me know!

client/src/js/components/bhRenderOption.js
```diff
@@ -0,0 +1,38 @@
...
+      onChange : '&',
+    },
+  });
+
+
+/**
+ * bhRenderOption Component
+ *
+ * A component to deal with date, it lets a user choose a date by either typing
+ * into an <input> or clicking a calendar dropdown.  It wraps the
```
jniles: This still talks about dates :(

client/src/js/components/bhRenderOption.js
```diff
@@ -0,0 +1,38 @@
+angular.module('bhima.components')
+  .component('bhRenderOption', {
+    templateUrl : '/modules/templates/bhRenderOption.tmpl.html',
+    controller : RenderOptionController,
+    bindings : {
+      orientation : '<', // two-way binding
```
jniles: This comment should now read "one-way binding"

client/src/modules/templates/modals/report.save.html
```diff
@@ -6,32 +6,38 @@
...
-    <select class="form-control" ng-model="SaveCtrl.documentOptions.renderer" ng-options="type.key as (type.label | translate) for type in SaveCtrl.supportedRenderTypes">
-    </select>
+    <div class="form-group">
+      <label class="control-label" translate>REPORT.UTIL.SAVE_AS</label>
+      <select class="form-control" ng-model="SaveCtrl.documentOptions.renderer" ng-options="type.key as (type.label | translate) for type in SaveCtrl.supportedRenderTypes">
+      </select>
+    </div>
+    <!-- layout orientation -->
+    <bh-render-option
+      on-change="SaveCtrl.setDocumentorientation"
```
jniles: If this is a function, it should be called in the HTML.  I'm not sure why AngularJS does this, but it's how all callbacks seem to work with AngularJS.  Maybe it is for clarity.

The best way to structure this is:
```html
<bh-render-option
  on-change="SaveCtrl.setDocumentOrientation(orientation)">
</bh-render-option>
```

... then your JS in the component can be:
```js
$ctrl.setOrientation = () => $ctrl.onChange({ orientation : $ctrl.orientation });
```
And I think that should work.

jniles (commented):
This is the right idea.  Just one more adjustment to the callback and it should be good to go.

jniles (approved):
Code LGTM.

bors r+

PR #2627, opened 2018-03-28
refractor(role management) select permision

### LINKED ISSUE #2621: Role Management should subselect all permission of a parent node
The Role Management module contains a tree of modules that a user should set permissions on.  However, the tree is not really usable.  If you wanted to make an accountant role with permissions to all the finance and reporting modules, you would need to click every single permission in the tree.

We can make this easier by allowing a user to click the top module of a tree to subselect all children in the tree.

Compare the two following user experiences:

**User Management Permissions Modal**

_Fig 1: User Management Selection_

**Role Management Permission Modal**

Fig 2: Role Management Selection_

**Proposed Solution**
I propose we use the same modal for both the user management and role management.  Probably the best way to do this is create a component for setting permission on individual modules.   It would be called `<bh-permissions-tree></bh-permissions-tree>` or something similar.  That way, we could write unit tests for it to unsure that the functionality is always preserved.

jeremielodi:
closes #2621


client/src/js/components/bhPermissionTree.js
```diff
@@ -0,0 +1,134 @@
+angular.module('bhima.components')
+  .component('bhPermissionsTree', {
+    bindings : {
+      assignedUnits : '=', // array [{unit_id :1}, {unit_id : 3}, ...]
```
jniles: The comment here is good!

In general, components shouldn't ever use two-way data binding (the `=` sign).  They should use one-way bindings (`<`) and then update the parent view with a callback `onChange()`.

The reason for this is two-fold:
 1. It is more efficient for Angular - once it knows the value can only be updated from the outside, all it needs to do is check the value of the parent scope without descending into the child scope.  This makes the application slightly faster.
 2. It enforces a [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) so that the child cannot change the variables of the parent controller except by explicitly calling a `callback()`.

AngularJS has a long discussion and examples of this works in their [component documentation](https://docs.angularjs.org/guide/component#creating-and-configuring-a-component).  I encourage you to read it over and over.  I took me a long time to understand how to structure components.

One way data binding was first introduced in AngularJS 1.5, but we started writing the application with AngularJS 1.2.   Therefore, we still have some old components, like the bhDateEditor that use the old style.  [But there are issues to try and fix those](https://github.com/IMA-WorldHealth/bhima-2.X/issues/1960).

client/src/js/components/bhPermissionTree.js
```diff
@@ -0,0 +1,134 @@
...
+  '$translate', '$http', 'util', 'NodeTreeService', 'NotifyService',
+];
+
+function PermissionsTreeController($translate, $http, util, NT, Notify) {
+  const ctrl = this;
+
+
+  ctrl.$onInit = function onInit() {
+    // bind methods
+    ctrl.editPermissions = editPermissions;
```
jniles: The `$onInit` function is designed for initialization code.  These are typically values that are passed into the component.  The `editPermissions` function is defined within the controller and will never change.  It would be best to put it outside of the `$onInit()` function to keep the function is small and testable.

jniles: [This is a good article](https://blog.thoughtram.io/angularjs/2016/03/29/exploring-angular-1.5-lifecycle-hooks.html) on how to use the "lifecycle" hooks of a component.

client/src/js/components/bhPermissionTree.js
```diff
@@ -0,0 +1,134 @@
...
+    ctrl.checkboxOffset = checkboxOffset;
+    ctrl.toggleUnitChildren = toggleUnitChildren;
+    ctrl.toggleSuperUserPermissions = toggleSuperUserPermissions;
+    ctrl.toggleParents = toggleParents;
+    ctrl.loading = true;
```
jniles: `$ctrl.loading` is the only item of this list that should be in the `$onInit()` method.

client/src/js/components/bhPermissionTree.js
```diff
@@ -0,0 +1,134 @@
...
+    // loads the permissions tree
+    function editPermissions() {
+      let units;
+
+      // load the tree units
+      loadUnits()
+        .then((data) => {
+
+          // unit value comparison function
+          function cmp(nodeA, nodeB) {
```
jniles: Because this function is in the `.then()` callback, the JS interpreter will need to create it and throw it away ever time `loadUser()` is called and then completes. 

We can reduce the load on the client's browser by moving this out of the `.then()` callback.  After all, it doesn't depend on the `data` variable.  So we can do this instead:
```js
function cmp(nodeA, nodeB) {
  // logic
}

loadUnits()
  .then(data => {
    // initialization code ...

    NT.flattenInPlace(tree, cmp);

    // ... etc ...
  });
```

client/src/js/components/bhPermissionTree.js
```diff
@@ -0,0 +1,134 @@
...
+        if (node.children) {
+          toggleUnitChildren(node, node.children);
+        }
+      });
+      updateAssignedUnits();
+    }
+
+
+    function updateAssignedUnits() {
+      ctrl.assignedUnits = ctrl.units.filter((u) => {
```
jniles: Alternatively, `ctrl.units.filter(u => u.checked);`.

client/src/js/components/bhPermissionTree.js
```diff
@@ -0,0 +1,134 @@
...
+        return u.checked;
+      })
+        .map((unit) => {
+          const u = { unit_id : unit.id };
```
jniles: Alternatively,
```js
.map(unit => ({ unit_id : unit.id }));
```

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -9,43 +9,14 @@
 </div>
 <div class="modal-body" style="max-height: 70vh; overflow:auto;">
 
-  <div class='row'>
-    <div class="col-lg-12">
+<bh-permissions-tree assigned-units='RolesPermissionsCtrl.tree'>
```
jniles: Nice 👍

server/controllers/admin/unit.js
```diff
@@ -15,34 +14,15 @@ function list(req, res, next) {
...
-    modules = rows;
-    return q.all(rows.map(row => { return db.exec(sql2, [roleUuid, row.id]); }));
-  })
-    .then((permissions) => {
-      permissions.forEach((permission, idx) => {
-        modules[idx].pages = permission;
-      });
-
-      res.json(modules);
+  db.exec(sql, [roleUuid])
```
jniles: 👍

client/src/js/components/bhPermissionTree.js
jniles: So, the best API for this is probably to have the following bindings:
```js
bindings : {
  assignedUnits : '<?',  // this could be either an array of JSONs or a simply an array of integers
  onChange : '&', // callback for when a user checks or unchecks a permission.
}
```

jniles (commented):
@jeremielodi this is a good start.

All new components should use AngularJS's recommendations for data binding: one-way bindings for inputs and callbacks for outputs.  This allows us to reason about the components more easily, as well as have the benefits discussed below.

Once you are done with these changes, we can work together on tests for this component.  If you want to get a head start over the weekend, we'll probably need the following tests:
 1. Only the `assignedUnits` should be checked/assigned by default.
 2. The `assignedUnits` should not be required in case a user doesn't have any permissions (as a new user would).  It should have all values unchecked.
 3. `checkboxOffset(2) should return `{ 'padding-left' : '60px' }`.
 4. `loadUnits()` should be called when `$onInit()` is called.
 5. `loadUnits()` should send a `GET` reqest to `/units`.
 6. `toggleParents()` should toggle the parent of a node in the tree.
 7. `toggleUnitChildren()` should check and uncheck all permissions on a child node.

... that should be a good start.  We might think of more to add as the component evolves.

jniles:
@jeremielodi, what happened to this?

jeremielodi:
I don't know, did you notice something?

jniles:
You closed it.  Was there a reason for that?

jeremielodi:
Oh! it seems that it is a mistake, I did not realise that i closed it

2018-04-06 10:16 GMT+01:00 Jonathan Niles <notifications@github.com>:

PR #2628, opened 2018-03-28
Import inventories from csv

mbayopanda:
This PR is about to : 
- Add a feature in the inventory module for import inventories from CSV file
- Add a feature in the inventory module for download a template file for inventories import
- Add an integration test for the `/inventory/import` API


client/src/js/services/util.js
```diff
@@ -73,19 +85,19 @@ function UtilService(moment) {
   // utility function
   service.clean = function clean(o) {
     // clean off the $$hashKey and other angular bits and delete undefined
-    var cleaned = {};
+    const cleaned = {};
 
-    for (var k in o) {
+    Object.keys(o).forEach(k => {
```
jniles, line 90: Alternatively, we could have used `angular.forEach()`.

client/src/js/services/util.js
```diff
@@ -100,12 +112,11 @@ function UtilService(moment) {
...
   service.once = function once(fn, context) {
-    var result;
-
-    return function () {
+    let result;
 
+    return function out() {
       if (!fn) {
-        return;
+        return null;
```
jniles, line 119: 👍

client/src/js/services/util.js
```diff
@@ -264,7 +275,7 @@ function UtilService(moment) {
   service.debounce = debounce;
 
   service.arrayIncludes = function arrayIncludes(array, values) {
-    return values.some(function (value) {
+    return values.some((value) => {
       return array.indexOf(value) !== -1;
```
jniles, line 279: Just a note -  we can use [array.includes(value)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) as a substitute for `array.indexOf(value) !== -1`.  It makes things slightly more readable.

client/src/modules/inventory/list/list.html
```diff
@@ -44,6 +44,14 @@
                 <span class="fa fa-file-excel-o"></span> <span translate>DOWNLOADS.CSV</span>
               </a>
             </li>
+
+            <!-- data import section -->
+            <li role="separator" class="divider"></li>
+            <li role="menuitem">
+              <a ng-click="InventoryCtrl.openImportInventoriesModal()" class="text-success" href>
+                <span class="fa fa-cloud-upload"></span> <span translate>INVENTORY.IMPORT_INVENTORIES</span>
```
jniles, line 52: 👍

client/src/modules/inventory/list/list.js
```diff
@@ -128,14 +126,14 @@ function InventoryListController(
...
-    onRegisterApi : onRegisterApi,
+    columnDefs,
+    onRegisterApi,
   };
 
   // configurations
-  gridColumns = new Columns(vm.gridOptions, cacheKey);
-  exportation = new GridExport(vm.gridOptions, 'visible', 'visible');
-  state = new GridState(vm.gridOptions, cacheKey);
+  const gridColumns = new Columns(vm.gridOptions, cacheKey);
```
jniles, line 134: 👍

client/src/modules/inventory/list/list.js
```diff
@@ -236,20 +234,31 @@ function InventoryListController(
...
   function exportCsv() {
     exportation.run();
   }
+
+  /**
+   * start the import of inventories from a csv file
+  */
+  function openImportInventoriesModal() {
+    Inventory.openImportInventoriesModal()
+      .then(() => {
```
jniles, line 259: Is this block supposed to be empty...?

server/controllers/inventory/import/index.js
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * formatCsvToJson
+ * @param {object} file the csv file sent by the client
+ */
+function formatCsvToJson(file) {
+  const filePath = path.resolve(file.path);
+  const defer = Q.defer();
+  const rows = [];
+
+  csvtojson()
```
jniles, line 99: Note - you could also use [stream to promise](https://www.npmjs.com/package/stream-to-promise) as a wrapper here.  We use it already in [lib/renderers/pdf.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/renderers/pdf.js#L91).

I believe it would look like this:
```js
const csvStream = csvtojson().fromFile(filePath).on('json', data => rows.push(data));

return streamToPromise(csvStream)
  .then(() => rows);
```

jniles: It also looks like [csvtojson emits a 'data' event](https://www.npmjs.com/package/csvtojson#data).  This means you might be able to clean the above code up a bit by setting [objectMode](https://nodejs.org/api/stream.html#stream_new_stream_readable_options) on the readable stream.

Like this:
```js
// up at the top, we can set the stream options
// see: https://www.npmjs.com/package/csvtojson#parameters for more information
const csvtojson = require('csvtojson');

// objectMode will make the 'data' event emit JSON objects, I think.
// See: https://nodejs.org/api/stream.html#stream_new_stream_readable_options
// and: https://www.npmjs.com/package/csvtojson#data
const converter = csvtojson(null, { objectMode: true });

// now I think we can just wrap the stream in stream-to-pdf
return streamtopdf(converter().fromFile(filepath));
```

This might not work .. you might still need to push JSON objects onto an array manually.  It depends on how streamtopdf treats each 'data' event.  But it might be worth a shot!

server/lib/db/index.js
```diff
@@ -32,6 +32,7 @@ class DatabaseConnector {
       user     : process.env.DB_USER,
       password : process.env.DB_PASS,
       database : process.env.DB_NAME,
+      charset  : 'utf8_unicode_ci',
```
jniles, line 35: 👍

... we should probably do this ima-data as well!

server/models/procedures/inventory.sql
```diff
@@ -0,0 +1,81 @@
...
+  DECLARE inventoryUnitId SMALLINT(5);
+
+  CALL SetupDebugLog();
+
+  SET existInventoryGroup = (SELECT IF((SELECT COUNT(`name`) AS total FROM `inventory_group` WHERE `name` = inventoryGroupName COLLATE utf8_unicode_ci) > 0, 1, 0));
+  SET existInventory = (SELECT IF((SELECT COUNT(`text`) AS total FROM `inventory` WHERE `code` = inventoryCode COLLATE utf8_unicode_ci OR `text` = inventoryText COLLATE utf8_unicode_ci) > 0, 1, 0));
+  SET existInventoryType = (SELECT IF((SELECT COUNT(*) AS total FROM `inventory_type` WHERE `text` = inventoryType COLLATE utf8_unicode_ci) > 0, 1, 0));
+  SET existInventoryUnit = (SELECT IF((SELECT COUNT(*) AS total FROM `inventory_unit` WHERE `text` = inventoryUnit COLLATE utf8_unicode_ci) > 0, 1, 0));
+
+  CALL DebugLog(CONCAT('Exist Inventory Group : ', existInventoryGroup));
```
jniles: You should remove all these `DebugLog` statements.  They are similar to `console.log()`, and I don't think we build `debug.sql` in production.

server/resources/templates/import-inventory-template.csv
```diff
@@ -0,0 +1 @@
+inventory_group_name, inventory_code, inventory_text, inventory_type, inventory_unit, inventory_unit_price
```
jniles, line 1: Oh, this is cool 👍

sh/build-database.sh
```diff
@@ -31,6 +31,7 @@ mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/functions.sql &> /dev/null
 echo "[build] procedures"
 mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/procedures.sql &> /dev/null
 mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/admin.sql &> /dev/null
+mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/debug.sql &> /dev/null
```
jniles, line 34: This is for debugging only.  Let's leave it out of the build process for now.

test/client-unit/services/util.spec.js
```diff
@@ -24,12 +22,12 @@ describe('util', function () {
 
     // the result of the function shouldn't change
     expect(fn(1)).to.be.equal(2);
-    expect(fn(10)).to.be.undefined;
-    expect(fn('some string')).to.be.undefined;
+    expect(fn(10)).to.be.equal(null);
```
jniles, line 25: 👍

test/fixtures/inventory-to-import.csv
```diff
@@ -0,0 +1,3 @@
+inventory_group_name, inventory_code, inventory_text, inventory_type, inventory_unit, inventory_unit_price
```
jniles, line 1: Beautiful 👍

test/integration/inventory/group.js
```diff
@@ -0,0 +1,66 @@
+/* global expect, agent */
+
+const helpers = require('../helpers');
```
jniles, line 3: Splitting this file up was a great improvement!  Well done!

test/integration/inventory/import.js
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * test the /inventory/import API for importing
+   * inventories from a csv file and comparing totals before and after the import
+   */
+  it('POST /inventory/import/ upload the filled template file as new import for inventories', () => {
+
+    // get the number of inventory before the import
+    return agent.get('/inventory/metadata')
+      .then(res => {
+        totalInventoriesBeforeImport = res.body.length;
```
jniles, line 39: 👍

test/integration/inventory/import.js
```diff
@@ -0,0 +1,61 @@
...
+
+        // import inventories from a csv file
+        return agent.post('/inventory/import')
+          /**
+           * to attach file into req.files please use fs.createReadStream
+           * fs.readFileSync doesn't work because it insert the file into req.body.file
```
jniles, line 45: So much pain in the comment. 💃

jniles (commented):
Code looks pretty good to me.

I made a lot of comments about the CSV stream - none of those are required, I just find streams interesting and curious if it could be more clean/compact.

You should remove all the debug code from the SQL though.  You are doing unnecessary MySQL writes, as well as triggering `COMMIT`s to the transaction.  That means if a CSV file fails half-way through, it will likely leave behind inventory groups and items.

One more thing: could you add an integration test for an error case?  Something like a CSV file with one correct line and one incorrect line?

I'm pulling this and testing it manually.

mbayopanda:
Thank you @jniles for your review 👍

mbayopanda:
@jniles could you review this PR again.

jniles (approved):
LGTM!

bors r+

PR #2629, opened 2018-03-28
feat(installation) assign the superuser role to the first user

### LINKED ISSUE #2625: BHIMA Installation script broken
I'm pretty sure the roles management merge broke the installation script.  You try to install a new enterprise, you will not have permission to log in with full units.

jeremielodi:
closes #2625

### DIFF
--- a/server/controllers/auth.js
+++ b/server/controllers/auth.js
@@ -63,9 +63,12 @@ function login(req, res, next) {
     SELECT user.id FROM user
     WHERE user.username = ? AND user.password = PASSWORD(?)`;
 
+  // a role should be assigned to the user
+  // each role has some units(paths or urls) that the user is allowed to access(permissions)
   const sqlPermission =
-    `SELECT permission.id FROM permission
-    JOIN user ON user.id = permission.user_id
+    `SELECT  user.id 
+    FROM  user_role
+    JOIN user ON user.id =  user_role.user_id
     WHERE user.username = ? AND user.password = PASSWORD(?)`;
 
   q.all([
@@ -75,7 +78,6 @@ function login(req, res, next) {
   ])
     .spread((connect, user, permission) => {
       _.merge(param, { connect, user, permission });
-
       const hasAuthorization = param.connect.length > 0;
       const isMissingPermissions = param.permission.length === 0;
       const isUnrecognizedUser = param.user.length === 0;
@@ -177,13 +179,16 @@ function loadSessionInformation(user) {
       [session.user] = rows;
 
       // next make sure this user has permissions
+      // we use now roles for assigning permissions to users
       sql = `
-        SELECT IF(permission.user_id = ?, 1, 0) authorized, unit.path
-        FROM unit LEFT JOIN permission
-          ON unit.id = permission.unit_id;
+        SELECT IF(user_role.user_id = ?, 1, 0) authorized, unit.path
+        FROM unit 
+        LEFT JOIN role_unit ON unit.id = role_unit.unit_id
+        LEFT JOIN user_role ON user_role.role_uuid = role_unit.role_uuid
+        WHERE user_role.user_id=?
       `;
 
-      return db.exec(sql, [session.user.id]);
+      return db.exec(sql, [session.user.id, session.user.id]);
     })
     .then(modules => {
       const unauthorized = modules.every(mod => !mod.authorized);
--- a/server/models/procedures.sql
+++ b/server/models/procedures.sql
@@ -52,4 +52,9 @@ SOURCE server/models/procedures/stock.sql
 */
 SOURCE server/models/procedures/inventory.sql
 
+/*
+  role management procedures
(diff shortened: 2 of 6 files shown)


test/end-to-end/errors/errors.spec.js
```diff
@@ -8,7 +8,7 @@ helpers.configure(chai);
 
 const FU = require('../shared/FormUtils');
 
-describe('Errors', () => {
+describe.only('Errors', () => {
```
jniles: This should be removed :)

sh/build-init-database.sh
```diff
@@ -36,4 +36,6 @@ echo "[ build ] default data"
 mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/icd10.sql &> /dev/null
 mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/bhima.sql &> /dev/null
 
+echo "[build] role management procedures"
+mysql -u $DB_USER -p$DB_PASS $DB_NAME <  server/models/procedures/roles.sql > /dev/null
```
jniles: All of our procedures are built from [procedures.sql](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/procedures.sql).  Could you add `source` there instead of building from the .sh file?

server/controllers/auth.js
```diff
@@ -177,13 +179,17 @@ function loadSessionInformation(user) {
...
       // next make sure this user has permissions
+      // we use now roles for assigning permissions to users
       sql = `
-        SELECT IF(permission.user_id = ?, 1, 0) authorized, unit.path
-        FROM unit LEFT JOIN permission
-          ON unit.id = permission.unit_id;
+        SELECT IF(user_role.user_id = ?, 1, 0) authorized, unit.path
+        FROM unit 
+        LEFT JOIN role_unit ON unit.id = role_unit.unit_id
+        LEFT JOIN role ON role.uuid = role_unit.role_uuid
```
jniles: Do you need the `LEFT JOIN role`?   It looks like this will work equally well:
```sql
SELECT IF(user_role.user_id = ?, 1, 0) authorized, unit.path
FROM unit
LEFT JOIN role_unit ON unit.id = role_unit.unit_id
LEFT JOIN user_role ON user_role.role_uuid =role_unit.role_uuid
WHERE user_role.user_id = ?;
```

It would save MySQL a bit of work if it does!

server/controllers/auth.js
```diff
@@ -63,9 +63,12 @@ function login(req, res, next) {
     SELECT user.id FROM user
     WHERE user.username = ? AND user.password = PASSWORD(?)`;
 
+  // a role should be assigned to the user
+  // each role has some unit(path or url) that the user is allowed to access(permissions)
```
jniles: Nice comment 👍

test/data.sql
```diff
@@ -823,15 +823,27 @@ INSERT INTO `payroll_configuration` (`id`, `label`, `dateFrom`, `dateTo`, `confi
...
+-- superuser
 INSERT INTO role_unit
  SELECT HUID(uuid()) as uuid,@roleUUID, id FROM unit;
 
 INSERT INTO `user_role`(uuid, user_id, role_uuid) 
 VALUES(HUID(uuid()), 1, @roleUUID);
 
+-- regular user
+INSERT INTO role_unit
+ VALUES(HUID(uuid()), @regularRoleUUID , 0 ),
```
jniles, line 841: Nice!

jniles (commented):
Just a few more tweaks here.

server/controllers/auth.js
```diff
@@ -177,13 +179,16 @@ function loadSessionInformation(user) {
...
 
       // next make sure this user has permissions
+      // we use now roles for assigning permissions to users
       sql = `
-        SELECT IF(permission.user_id = ?, 1, 0) authorized, unit.path
-        FROM unit LEFT JOIN permission
-          ON unit.id = permission.unit_id;
+        SELECT IF(user_role.user_id = ?, 1, 0) authorized, unit.path
+        FROM unit 
+        LEFT JOIN role_unit ON unit.id = role_unit.unit_id
```
jniles, line 186: 💃

server/models/procedures.sql
```diff
@@ -47,4 +47,10 @@ SOURCE server/models/procedures/trial_balance.sql
 */
 SOURCE server/models/procedures/stock.sql
 
+
+/*
+  role management procedures
+*/
+SOURCE server/models/procedures/roles.sql > /dev/null
```
jniles: The `> /dev/null` is a bash command, but this is run in SQL.  I'm surprised it works!  Can you remove it?

jniles (approved):
Except for the bash redirect, code LGTM.

jniles:
@jeremielodi once the conflicts here are fixed, we can merge this branch.

PR #2630, opened 2018-03-28
Implement bhBarcodeScanner Component

jniles:
This PR implements the `bhBarcodeScanner` component and uses it to replace the barcode scanning functionality on the Cash Payments page.

**Component Signature**
```html
<bh-barcode-scanner on-scan-callback="callback(record)"></bh-barcode-scanner>
```

**Description**
The `bhBarcodeScanner` uses the underlying `BarcodeService` to query the barcode API and determine which record to return to the client.  If the API returns a success HTTP code, it will call the `onScanCallback` function.

The key difference between this component and the previous implementation is the ability to re-focus the hidden barcode input if focus is lost.  This is accomplished by clicking a button that is only visible once focus is lost from the hidden input.

**In Action**


client/src/i18n/en/barcode.json
```diff
@@ -1,5 +1,11 @@
...
-"READ_ERROR":"Unreadable Barcode!  Please enter the barcode manually."}}
\ No newline at end of file
+{
+  "BARCODE" : {
+    "SCAN" : "Scan Document Barcode",
+    "AWAITING_INPUT" : "Waiting for Input",
+    "AWAITING_HTTP" : "Barcode Read!  Looking up Record...",
+    "READ_SUCCESS" : "Found Record",
+    "READ_ERROR" : "Unreadable Barcode!  Please enter the barcode manually.",
+    "LOST_FOCUS" : "The barcode input has lost focus.  Please click \"Read Barcode\" to ready the barcode component.",
```
mbayopanda: 👍

client/src/modules/cash/cash-form.service.js
```diff
@@ -121,7 +121,7 @@ function CashFormService(AppCache, Session, Patients, Exchange) {
   CashForm.prototype.configure = function configure(config) {
 
     if (config.patient) {
-      this.setPatient(config.invoices);
+      this.setPatient(config.patient);
```
mbayopanda, line 124: 👍 good catch

test/client-unit/components/bhBarcoderScanner.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    // Mock the document.getElementById call in the setFocusOnHiddenInput() call.
+    _$window_.document.getElementById = (id) => find(element, `#${id}`);
+
+    // spy on the onScanCallback
+    $scope.callback = chai.spy();
+
+    $scope.$digest();
+  }));
+
+  // TODO - how do you find a focused element?
```
mbayopanda, line 48: The TODO message is about to find a hidden input box or a focused element ? Because the test below is about a hidden input.

test/client-unit/components/bhBarcoderScanner.spec.js
```diff
@@ -0,0 +1,139 @@
...
+
+  it('shows the reset button if focus is lost from the input', () => {
+    const btn = find(element, '[data-method="reset"]');
+    expect(isHidden(btn)).to.equal(true);
+
+    const input = find(element, '#hidden-barcode-input');
+    angular.element(input).triggerHandler('blur');
+    $scope.$digest();
+
+    expect(isHidden(btn)).to.equal(false);
```
mbayopanda, line 68: 👍

client/src/i18n/fr/barcode.json
```diff
@@ -1,5 +1,11 @@
...
-"AWAITING_HTTP":"Code à barres lu! Recherche de factures ....",
-"READ_SUCCESS":"Facture Trouvée",
-"READ_ERROR":"Code à barres illisible! Entrez le code à barres manuellement."}}
\ No newline at end of file
+{
+  "BARCODE" : {
+    "SCAN" : "Scanner le Code à Barres",
+    "AWAITING_INPUT" : "En Attente d'Entrée",
+    "AWAITING_HTTP" : "Code à barres lu! Recherche l'enregistrement....",
+    "READ_SUCCESS" : "Trouvée",
```
mbayopanda: Trouvé et non Trouvée (parce que le mot enregistrement est au masculin).

A good message can be : `Enregistrement trouvé` instead of simply `Trouvé`

client/src/i18n/fr/barcode.json
```diff
@@ -1,5 +1,11 @@
...
+    "READ_ERROR" : "Code à barres illisible! Entrez le code à barres manuellement.",
```
mbayopanda: for the "READ_ERROR" a good key and message can be :
```
"READ_FAILED" : "Enregistrement non trouvé"
```
and for any errors during the reading : 
```
"READ_ERROR" : "Une erreur est survenue lors de la lecture du code barre"
```

client/src/js/components/bhBarcodeScanner.js
```diff
@@ -0,0 +1,69 @@
...
+    $ctrl.currentStep = steps.AWAIT_HTTP;
+
+    Barcode.search($ctrl.barcode)
+      .then(record => {
+        $ctrl.currentStep = steps.READ_SUCCESS;
+        $ctrl.record = record;
+        $ctrl.onScanCallback({ record });
+      })
+      .catch(() => {
+        $ctrl.currentStep = steps.READ_ERROR;
```
mbayopanda: If the record is not found this code falls in the .catch section, and set the step to an error.

I think if the record is not found the step must be READ_FAILED.

On the server side if we use `db.exec`, on the client we can have : 
```js
if (records.length === 0) {
  $ctrl.currentStep = steps.READ_FAILED;
}
```

and the READ_ERROR must be set only if we have a real error.

client/src/modules/templates/bhBarcodeScanner.html
```diff
@@ -0,0 +1,37 @@
...
+      <span translate>BARCODE.AWAITING_HTTP</span>
+    </span>
+
+    <span class="text-success" ng-if="$ctrl.currentStep === $ctrl.steps.READ_SUCCESS">
+      <i class="fa fa-check-circle-o"></i> <span translate>BARCODE.READ_SUCCESS</span> {{ $ctrl.record.reference }}
+    </span>
+
+    <span class="text-danger" ng-if="$ctrl.currentStep === $ctrl.steps.READ_ERROR">
+      <i class="fa fa-danger-sign"></i> <span translate>BARCODE.READ_ERROR</span>
+    </span>
```
mbayopanda: Add also a  case for a READ_FAILED

mbayopanda (commented):
The code is excellent and the test also, i just made some comments about messages according the step.

test/client-unit/components/bhBarcoderScanner.spec.js
jniles: Ah, I can see how that isn't clear.

I originally tried to write a test to find a "hidden, focused" input, but never got the focus to work.  The CSS selector `$(':focus')` is supposed to work, but I never got that working...

So it's currently a TODO.  I should probably just make a skipped test for it.

client/src/i18n/fr/barcode.json
jniles: That is an excellent idea.  I'll expand the API for it.

jniles:
@mbayopanda, this is ready for a second review!

client/src/i18n/en/barcode.json
```diff
@@ -1,5 +1,12 @@
...
-"READ_ERROR":"Unreadable Barcode!  Please enter the barcode manually."}}
\ No newline at end of file
+{
+  "BARCODE" : {
+    "SCAN" : "Scan Document Barcode",
+    "AWAITING_INPUT" : "Waiting for Input",
+    "AWAITING_HTTP" : "Barcode Read!  Looking up Record...",
+    "READ_SUCCESS" : "Found Record",
+    "READ_ERROR" : "An error occurred during lookup.  Please use another method to find the record.",
+    "NOT_FOUND" : "Cannot find a record with that code.  Please search for the document manually.",
```
mbayopanda, line 8: 👍 good

client/src/js/components/bhBarcodeScanner.js
```diff
@@ -0,0 +1,71 @@
...
+  const $ctrl = this;
+
+  // steps in the search space
+  const steps = {
+    AWAIT_READ :  2,
+    AWAIT_HTTP : 4,
+    READ_SUCCESS : 8,
+    READ_ERROR : 16,
+    LOST_FOCUS : 32,
+    NOT_FOUND : 64,
```
mbayopanda, line 24: 💃

client/src/js/components/bhBarcodeScanner.js
```diff
@@ -0,0 +1,71 @@
...
+
+    Barcode.search($ctrl.barcode)
+      .then(record => {
+        $ctrl.currentStep = steps.READ_SUCCESS;
+        $ctrl.record = record;
+        $ctrl.onScanCallback({ record });
+      })
+      .catch(err => {
+        const isNotFound = (err.status === 404);
+        $ctrl.currentStep = isNotFound ? steps.NOT_FOUND : steps.READ_ERROR;
```
mbayopanda, line 62: 👍

server/lib/barcode.js
```diff
@@ -36,7 +36,7 @@ function generate(receiptIdentifier, uuid) {
 // YYYYYYYY - First characters of the entity UUID
 // - returns the full UUID of the entity
 function reverseLookup(barcodeKey) {
-  const code = barcodeKey.substr(0, 2);
+  const code = barcodeKey.substr(0, 2).toUpperCase();
```
mbayopanda, line 39: great 👍

mbayopanda (approved):
LGTM

PR #2632, opened 2018-03-29
feat(user management) remove permission

### LINKED ISSUE #2620: Remove permissions from user management
We recently replaced individual module permissions with user roles.  However, the old links to permissions exist and no longer make sense to users who are used to the previous way of assigning a permission to each user.  See for example the workflow below that I tried:

_Fig 1: Trying to create a user the old way_

There are a couple of problems with this:
 1. It gets the user into a really confusing state that they never could have gotten into previously.  We blocked users who didn't have permissions previously.
 2. It's unclear what the user did wrong.  What good are permission if there are no roles?

I propose that we remove the ability to set permissions per user completely.  We should also list the roles of the user in the user list.

### LINKED ISSUE #2624: Improvements to Role Management
The following improvements could be made to Role Management:
 1. Remove the ui-grid-selection plugin on the grid.  It server no purpose that I can find.
 2. Use the `.headercrumb` class in the headers of all modals to make them appear consistent with the rest of the application.
 3. Allow a user to delete a role that has permissions assigned to it.  You should _not_ be able to remove a role that has been assigned to a user.
 4. Do not show a success message when the user cancels the permissions modal.

jeremielodi:
closes #2620


jniles (approved):
Code LGTM.  This is the best PR we've received all week. Less code is almost always better than more! 💃

PR #2633, opened 2018-03-29
feat(role management) improvements

### LINKED ISSUE #2624: Improvements to Role Management

jeremielodi:
closes #2624

### DIFF
--- a/client/src/modules/roles/modal/rolesPermissions.html
+++ b/client/src/modules/roles/modal/rolesPermissions.html
@@ -2,53 +2,58 @@
 name="permissionForm"
 bh-submit="RolesPermissionsCtrl.getAffected()"
 novalidate>
-<div class="modal-header">
+  <div class="modal-header">
+    <ol class="headercrumb">
+      <li class="static" translate>USERS.PERMISSIONS</li>
+      <li class="title">
+        <span>{{RolesPermissionsCtrl.role.label}}</span>
+      </li>
+    </ol>
+  </div>
+  <div class="modal-body" style="max-height: 70vh; overflow:auto;">
 
-  <span translate>USERS.PERMISSIONS</span> :
-  <span class="text-primary"> {{RolesPermissionsCtrl.role.label}}</span>
-</div>
-<div class="modal-body" style="max-height: 70vh; overflow:auto;">
+    <div class='row'>
+      <div class="col-lg-12">
 
-  <div class='row'>
-    <div class="col-lg-12">
-
-      <div>
-        <label class="radio-inline">
-          <input type='checkbox' ng-change="RolesPermissionsCtrl.selectAll()" ng-model='RolesPermissionsCtrl.allPage' ng-true-value="1"
-            ng-false-value="0" />
-          <span id="checkall" translate>FORM.LABELS.CHECK_ALL</span>
-        </label>
-      </div>
-      <div ng-repeat="module in RolesPermissionsCtrl.tree" class="treeview">
-
-        <h4>
+        <div>
           <label class="radio-inline">
-            <input type='checkbox' ng-model='module.affected' ng-true-value="1" ng-false-value="0" />
-            <b translate>{{module.key}}</b>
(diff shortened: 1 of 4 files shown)


client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -2,7 +2,7 @@
 name="permissionForm"
 bh-submit="RolesPermissionsCtrl.getAffected()"
 novalidate>
-<div class="modal-header">
+<div class="modal-header headercrumb">
```
jniles: The `headercrumb` should be used with an ordered list.  See [the cash invoice modal](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/cash/modals/invoice-modal.html#L2) for how to do it.

The end result should look something like this:
headercrumbexample

client/src/modules/roles/modal/rolesPermissions.html
```diff
@@ -45,7 +45,7 @@ <h4>
 <div class="modal-footer">
 
   <div pull-right>
-    <button class='btn btn-default' ng-click='RolesPermissionsCtrl.close()' translate>FORM.BUTTONS.CANCEL</button>
+    <span class='btn btn-default' ng-click='RolesPermissionsCtrl.close()' translate>FORM.BUTTONS.CANCEL</span>
```
jniles: If you are going to click an element, it should really be either a `button` or and `a` element.  Why was this change necessary?

client/src/modules/roles/modal/userRole.html
```diff
@@ -23,7 +23,7 @@ <h4><span translate>FORM.LABELS.ROLES_FOR</span>
 
 <div class="modal-footer">
   <div class=" pull-right">
-  <button class='btn inverse' ng-click='UsersRolesCtrl.close()'>Fermer</button>
+  <span class='btn inverse' ng-click='UsersRolesCtrl.close()'>Fermer</span>
```
jniles: Oh - this should be translated.  I never noticed it before.  Could you make this a button as well?

server/models/schema.sql
```diff
@@ -1696,7 +1696,7 @@ CREATE TABLE `role_unit` (
   `role_uuid`  binary(16) NOT NULL,
   `unit_id` SMALLINT(5) UNSIGNED DEFAULT NULL,
   PRIMARY kEY(`uuid`),
-  FOREIGN KEY (`role_uuid`) REFERENCES `role` (`uuid`) ON UPDATE CASCADE,
+  FOREIGN KEY (`role_uuid`) REFERENCES `role` (`uuid`) ON UPDATE CASCADE ON DELETE CASCADE,
```
jniles, line 1699: 👍

client/src/modules/roles/roles.html
```diff
@@ -28,7 +28,6 @@
         ui-grid-auto-resize 
         ui-grid-empty-base-layer 
         ui-grid="RolesCtrl.gridOptions"
-        ui-grid-selection
```
jniles, line 31: 💃

jniles (commented):
This needs a bit more work.

We should always use either `a` or `button` elements to click, unless there is a really good reason not to use them.  My guess for why you removed the `button` tags is that [an HTML button's default type is "submit"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type), which results in your form being submitted even if you click "cancel".  Set the `<button type="button">` and you should be fine.

The headercrumb's purpose is to make a nice rendering cascade as pictured.  See if you can get it working and if you need more help, I'll be glad to help.

jniles (approved):
The code has improved!  Well done!

bors r+

PR #2635, opened 2018-04-02
feat(patient document) preview only images when uploading

### LINKED ISSUE #2590: Document Upload shouldn't show preview if not an image/pdf
The Patient Document Upload feature attempts to show a preview of the uploaded document after it has been uploaded.  However, this is not necessarily intuitive if the upload isn't an image.  For example, see the two cases below:

goodimageupload
_Fig 1: Upload is an image.  It is nice to get a preview._

badpreviewupload
_Fig 2: Upload is a .docx format.   Why is there so much whitespace?_

I propose that we only show the preview if it is an image or PDF being uploaded.  I believe this will be a better user experience than what currently exists.

jeremielodi:
closes #2590

### DIFF
--- a/client/src/modules/patients/documents/modals/documents.modal.html
+++ b/client/src/modules/patients/documents/modals/documents.modal.html
@@ -28,7 +28,7 @@
 
     <!-- thumbnail  -->
     <div ng-if="$ctrl.file && $ctrl.isImage" class="upload-thumbnail">
-      <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}">
+      <img id="upload_thumbnail" ngf-thumbnail="{{ '$ctrl.thumbnail' }}">
     </div>
     
 
--- a/client/src/modules/patients/documents/modals/documents.modal.html
+++ b/client/src/modules/patients/documents/modals/documents.modal.html
@@ -27,9 +27,10 @@
     </div>
 
     <!-- thumbnail  -->
-    <div ng-if="$ctrl.file" class="upload-thumbnail">
+    <div ng-if="$ctrl.file && $ctrl.isImage" class="upload-thumbnail">
       <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}">
     </div>
+    
 
     <!-- progress bar  -->
     <div class="progress" ng-show="$ctrl.file.progress >= 0">
--- a/client/src/modules/patients/documents/modals/documents.modal.js
+++ b/client/src/modules/patients/documents/modals/documents.modal.js
@@ -50,6 +50,7 @@ function PatientDocumentsModalController(Instance, Upload, patientUuid, PatientS
   function setThumbnail(file) {
     vm.thumbnail = file || 'assets/placeholder.gif';
     vm.documentError = !file ? true : false;
+    vm.isImage = file.type.includes('image/');
   }
 
   /** upload the file to server */
--- a/test/end-to-end/patient/record.spec.js
+++ b/test/end-to-end/patient/record.spec.js
@@ -82,9 +82,11 @@ describe('Patient Record', () => {
     FU.input('$ctrl.file', absolutePath);
 
     FU.modal.submit();
+
     components.notification.hasSuccess();
   });
 
+ 
   // upload patient documents
   it('upload a PDF document', () => {
     const title = '[e2e] New Document';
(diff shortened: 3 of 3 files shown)


client/src/modules/patients/documents/modals/documents.modal.html
```diff
@@ -1,3 +1,4 @@
+<script src="//wzrd.in/standalone/blob-util@latest"></script>
```
jniles: What is this?  Why is it necessary?  Where it is used?

jeremielodi: oh! I should remove it

client/src/modules/patients/documents/modals/documents.modal.js
```diff
@@ -50,6 +50,12 @@ function PatientDocumentsModalController(Instance, Upload, patientUuid, PatientS
   function setThumbnail(file) {
     vm.thumbnail = file || 'assets/placeholder.gif';
     vm.documentError = !file ? true : false;
+    vm.isImage = false;
+    vm.isPdf = false;
```
jniles: Where do you use `vm.isPdf`?

client/src/modules/patients/documents/modals/documents.modal.js
```diff
@@ -50,6 +50,12 @@ function PatientDocumentsModalController(Instance, Upload, patientUuid, PatientS
...
+    const imageCheck = file.type.search('image/');
```
jniles: Is `file.type` a string?  If so, you can reduce this to a single line:
```js
vm.isImage = file.type.includes('image/');
```

jniles (changes requested):
I think you can do this in two lines if `file.type` is a string.

Could you write a quick unit test to show that the function `setThumbnail()` will properly set `vm.isImage` if the filetype contains `image/`?

test/end-to-end/patient/record.spec.js
```diff
@@ -123,6 +125,34 @@ describe('Patient Record', () => {
     FU.exists(by.css('[data-view="list"]'), true);
   });
 
+
+
+  it('Should check if upload_thumbnail is not displayed', () => {
```
jniles: A better description here would tell why it isn't displayed.  Something like "A thumbnail should not be shown if the upload is not an image".

test/end-to-end/patient/record.spec.js
```diff
@@ -123,6 +125,34 @@ describe('Patient Record', () => {
...
+    const title = '[e2e] New Image As Document';
+    const fileToUpload = 'file.pdf';
+    const absolutePath = path.resolve(fixtures, fileToUpload);
+
+    element(by.css('[data-document-action="add"]')).click();
+
+    FU.input('$ctrl.title', title);
+    FU.input('$ctrl.file', absolutePath);
+    FU.exists(by.css('[id="upload_thumbnail"]'), false);
```
jniles: Note that you can also use `FU.exists(by.id('upload_thumbnail'))` and it should be faster, since it can use `document.getElementById()`.

test/end-to-end/patient/record.spec.js
```diff
@@ -123,6 +125,34 @@ describe('Patient Record', () => {
     FU.exists(by.css('[data-view="list"]'), true);
   });
 
+
+
+  it('Should check if upload_thumbnail is not displayed', () => {
+    const title = '[e2e] New Image As Document';
```
jniles: Technically, this isn't an image.   Maybe say "New PDF"?

test/end-to-end/patient/record.spec.js
```diff
@@ -123,6 +125,34 @@ describe('Patient Record', () => {
...
+    FU.exists(by.css('[id="upload_thumbnail"]'), false);
+    FU.modal.close();
+  });
+
+  it('Should check if upload_thumbnail is displayed', () => {
+    const title = '[e2e] New Image As Document';
+    const fileToUpload = 'file.jpg';
+    const absolutePath = path.resolve(fixtures, fileToUpload);
+
+    element(by.css('[data-document-action="add"]')).click();
```
jniles: Just FYI - `$` is an alias for `element(by.css())`.  So you can do this as well:
```js
$('[data-document-action="add"]').click();
```

jniles (commented):
Nice end to end test!

The text should probably tell us what the test is testing.  If it breaks in the future, we'll have a better idea how to fix it if it is better described.  Otherwise, LGTM.

jniles (approved):
Beautiful!

bors r+

PR #2636, opened 2018-04-02
fix tree navigation console error

### LINKED ISSUE #2623: Unit Tree throws errors when module key not matched
We have several modules/paths that are not explicitly represented in the tree.  This causes the algorithm that tries to select an individual unit fail with an error.  This happens in the test dataset on the modules **Report > [Stock] Inventories** and **Report > [Stock] Inventory Report**.

erroronunittree
_Fig 1: Example error thrown on tree_

mbayopanda:
This PR fix the error in the console when navigating on some nodes of the unit tree.

close #2623


jniles (approved):
LGTM.

bors r+

PR #2640, opened 2018-04-03
deps: use @ima-worldhealth/topic

### LINKED ISSUE #2634: Move Topic.js into its own repository
The [Topic](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/topic.js) library used in most modules to pubish events, even if these events are not subscribed to.  The original vision was to have an activity stream of events happening on the client that could be displayed on a dashboard somewhere.

Topic is becoming more trouble than it is worth - because it uses Redis of asynchronous tests, it can be flaky on Travis CI.  This has crashed a number of PRs early, including https://github.com/IMA-WorldHealth/bhima-2.X/pull/2633.

I propose that we move it out into its own repository.  We will gain the following advantages:
 1. Remove a few dependencies from the main bhima-2.x instance, including `ioredis`.
 2. Remove the tests from 2.x, instead putting them on their own CI where we can debug them much more easily.
 3. Be able to write extensive tests for Topic to ensure that it actually works correctly when we plan to start using.
 4. Make it easier to decouple and turn off.

jniles:
This commit removes the custom topic library and imports the NPM published version.  This reduces the testing load on the 2.x application as well as makes sure flaky tests are less likely to occur.

Closes #2634


server/controllers/admin/locations.js
```diff
@@ -15,9 +15,10 @@
  * */
 
 
-const db = require('../../lib/db');
 const uuid = require('uuid/v4');
-const Topic = require('../../lib/topic');
+const Topic = require('@ima-worldhealth/topic');
```
mbayopanda, line 19: 👍 it is official

server/lib/topic.js
```diff
@@ -1,277 +0,0 @@
...
-  }
-
-  /* entities that could be affected by the events */
-  get entities() {
-    return entities;
-  }
-}
-
-/* export a singleton Event Emitter */
-module.exports = new Topic();
```
mbayopanda, line 277: 💃

mbayopanda (approved):
LGTM

PR #2643, opened 2018-04-04
Fix stock movement today filter

### LINKED ISSUE #2641: Do not show lots with 0 Quantity in the Stock Registry
In the "lots in stock" registry, we sometimes shows lots with 0 quantity.  This can happen if a transfer is occurring between one depot and another, but hasn't arrived in the end depot.

**Steps to Reproduce**
 1. Register a stock entry to some depot.  For example, 100 pills of penicillin.
 2. Immediately exit that stock to another depot via the Stock Exit module.
 3. View the lot in the Stock Registry.  It will look something like this:
zeroquantitylots
_Fig 1: Lot with 0 Quantity in a Depot_

**Proposed Solution**
By default, we should hide all quantities that are 0 quantity (probably via a SQL `HAVING` statement).  If the user wants to see them, we can add a toggle on the Search Modal that allows the user to select "show 0 quantity lots".

mbayopanda:
This PR fix the stock movement registry which was not able to display correct data for period `today` and also the lot registry which was not able to display the correct entry date of a lot in a given depot.

close #2641


server/controllers/stock/core.js
```diff
@@ -103,14 +103,18 @@ function getLots(sqlQuery, parameters, finalClauseParameter) {
 
   filters.period('defaultPeriod', 'date');
   filters.period('defaultPeriodEntry', 'entry_date', 'l');
-  filters.period('period', 'entry_date');
+  filters.period('period', 'date');
```
jniles, line 106: 💃

client/src/modules/stock/movements/modals/search.modal.js
```diff
@@ -105,20 +104,19 @@ function SearchMovementsModalController(data, Notify, Instance, Flux, $translate
...
 
     // push all searchQuery values into the changes array to be applied
-    angular.forEach(vm.searchQueries, function handleDefinedValue(_value, key) {
-      var _displayValue;
+    angular.forEach(vm.searchQueries, (_value, key) => {
+      let _displayValue;
 
       if (angular.isDefined(_value)) {
         // default to the original value if no display value is defined
         _displayValue = displayValues[key] || lastDisplayValues[key] || _value;
```
jniles: Unfortunately, this has the bug reported in https://github.com/IMA-WorldHealth/bhima-2.X/issues/2595.  This commit https://github.com/jniles/bhima-2.X/commit/826d7292b5c5832753539f9b9600ffad3d374ec7#diff-a8511c3ae241c3b5dd78e90697b17bc3L113 shows you how to change it so that the bug doesn't exist.

jniles (commented):
Code looks fine.

Note that there is still a filter bug that we fixed on other registries but not on this one.  I've linked a solution, let me know if you need help fixing it.  It would be good to fix it in this PR.

PR #2645, opened 2018-04-04
feat(stok lot)csv download considers filters

### LINKED ISSUE #2638: Downloading CSV of stock registry should limit to only the searched items
When you download a CSV from the stock registry, the entire registry is printed out.  If an enterprise has been functioning a long time, it will have a ton of stock and we should allow them to only download part of the list.  Additionally, most other registries function this way.

Steps to reproduce:
 1. Navigate to the Stock Registry.
 2. Search on a depot.
 3. Click "Download as CSV".
 4. Note that the entire list is downloaded.

jeremielodi:
closes #2638

### DIFF
--- a/test/integration/stock.js
+++ b/test/integration/stock.js
@@ -183,6 +183,28 @@ describe('(/stock/) The Stock HTTP API', () => {
       .catch(helpers.handler)
   );
 
+    // (report) render all stock exit
+  it(
+    `GET /reports/stock/lots?renderer=json
+    returns exits for all depots (12 OUT)`,
+    () => agent.get(`/reports/stock/lots?renderer=json`)
+      .then((res) => {
+        expect(res.body.rows.length).to.equal(12);
+      })
+      .catch(helpers.handler)
+  );
+  
+  // (report) render all stock exit relatives to 'Depot Principal'
+  it(
+    `GET /reports/stock/lots?renderer=json
+    returns exits for Depot principal(10 OUT)`,
+    () => agent.get(`/reports/stock/lots?renderer=json&depot_uuid=${depotPrincipalUuid}`)
+      .then((res) => {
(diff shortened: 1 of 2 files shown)


jeremielodi:
@jniles, what test I should write for this issue, it should concern the csv file content or grid

jniles:
@jeremielodi, good question!

The easiest thing: You can write an integration test that pull data using the `json` renderer.  The code you have changed affects all renderers, not just the CSV.  So you can do two tests:
 1. The first will look at `/reports/stock/lots?renderer=json` and make sure it downloads the full set of rows.
 2. The second test will look at `/reports/stock/lots?depot_uuid=XXXX?renderer=json` and make sure it downloads only part of the dataset.

Since all renderers begin with the JSON renderer, we can assume that, if the JSON renderer limits the date, the rest of them will also limit the data.

test/integration/stock.js
```diff
@@ -183,6 +183,28 @@ describe('(/stock/) The Stock HTTP API', () => {
...
+    () => agent.get(`/reports/stock/lots?renderer=json`)
+      .then((res) => {
+        expect(res.body.rows.length).to.equal(12);
+      })
+      .catch(helpers.handler)
+  );
+  
+  // (report) render all stock exit relatives to 'Depot Principal'
+  it(
+    `GET /reports/stock/lots?renderer=json
```
jniles, line 199: This is the right idea!

The text should actually reflect what the test is testing.  So, in this case, it should read:
```js
it(`GET /reports/stock/lots?renderer=json&depot_uuid=${depotPrincipalUuid} returns exists for depot principal (10 out)`)
```

jniles (approved):
Really good!  One last update to the text so that we all know what the test is doing and we can merge this.

PR #2658, opened 2018-04-04
feat(patient registry) add location filter

### LINKED ISSUE #2601: Missing Patient Registry Filters
It would make sense to be able to filter a patient's registration based on any of the required information on the patient page.  We are currently missing filters for the location information of a patient.

I propose that we use a single text field for a user to type one part of the location information (e.g. "DRC" or "Kimpese").  This would then be used to look up their corresponding information in the database.

jeremielodi:
closes #2601

### DIFF
--- a/client/src/modules/patients/registry/search.modal.html
+++ b/client/src/modules/patients/registry/search.modal.html
@@ -102,6 +102,20 @@
               mode="clean">
             </bh-date-interval>
           </fieldset>
+
+          <!-- location -->
+
+          <div class="form-group" ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.originLocationLabel.$invalid }">
+            <label class="control-label" translate>
+                FORM.LABELS.ORIGIN_LOCATION
+            </label>
+            <bh-clear on-clear="$ctrl.clear('originLocationLabel')"></bh-clear>
+
+            <input type="text" class="form-control" name="originLocationLabel" ng-model="$ctrl.searchQueries.originLocationLabel">
+            <div class="help-block" ng-messages="ModalForm.originLocationLabel.$error" ng-show="ModalForm.$submitted">
+              <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+            </div>
+          </div>
         </div>
       </uib-tab>
 
--- a/server/controllers/medical/patients/index.js
+++ b/server/controllers/medical/patients/index.js
@@ -437,6 +437,11 @@ function find(options) {
   filters.dateTo('dateBirthTo', 'dob');
   filters.equals('health_zone');
   filters.equals('health_area');
+
+  // filters for location
+  const orignSql = `(originVillage.name LIKE ?) OR (originSector.name LIKE ?) OR (originProvince.name LIKE ?)`;
+  const params1 = _.fill(Array(3), `%${options.originLocationLabel || ''}%`);
+  filters.customMultiParameters('originLocationLabel', orignSql, params1);
   // default registration date
   filters.period('period', 'registration_date');
   filters.dateFrom('custom_period_start', 'registration_date');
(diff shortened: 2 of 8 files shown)


server/controllers/medical/patients/index.js
```diff
@@ -437,6 +437,12 @@ function find(options) {
   filters.dateTo('dateBirthTo', 'dob');
   filters.equals('health_zone');
   filters.equals('health_area');
+
+  // filters for location
+  const orignSql = `(originVillage.name LIKE ?) OR (originSector.name LIKE ?) OR (originProvince.name LIKE ?)`;
+  let village = options.originVillageName || ``;
+  village = ''.concat('%', village, '%');
+  filters.customMultiParameters('originVillageName', orignSql, [village, village, village]);
```
jniles: This parameter can be more than just the village, correct?  Would it make more sense to call it "originLocationLabel" or something of that nature?

server/controllers/medical/patients/index.js
```diff
@@ -437,6 +437,12 @@ function find(options) {
   filters.dateTo('dateBirthTo', 'dob');
   filters.equals('health_zone');
   filters.equals('health_area');
+
+  // filters for location
+  const orignSql = `(originVillage.name LIKE ?) OR (originSector.name LIKE ?) OR (originProvince.name LIKE ?)`;
+  let village = options.originVillageName || ``;
```
jniles: Just FYI, you could do this in one line with a combination of [lodash's `_.fill()` method](https://lodash.com/docs/4.17.5#fill) and [JavaScript Template Strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) like so:
```js
const params = _.fill(3, `%${options.originVillageName || ''}%`);
filters.customMultiParameters('originVillageName', originSQL, params);
```

test/end-to-end/patient/registry.search.js
```diff
@@ -102,6 +103,13 @@ function PatientRegistrySearch() {
     expectNumberOfGridRows(NUM_MATCHING);
   });
 
+  it(`should find patients with origin location "${parameters.originVillageName}" `, () => {
```
jniles, line 106: Nice test!

For APIs, it is easier to test them by writing an integration test.  They are both faster to execute and easier to debug.  The tests for the patients is in [this file](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/integration/patients.js).

jniles (commented):
The logic is good!  Code is pretty clear as well.  Well done!

Since this text can contain the village, province, or sector I propose that you update the name to help future developers understand the code a bit better.

It would also be helpful to write an integration test for this.  The end to end test is valuable, but it might be difficult to debug.  If we compliment it with an integration test, we'll be sure this keep working clearly into the future.

server/controllers/medical/patients/index.js
jeremielodi: the server filter service  crached when I used a name that doesn't exist in the sql result, that is why I was obliged to use originVillage

jniles: I believe the trick is to make sure you have the same name in three places:
 1. The client-side filter service [here](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2658/files#diff-a1aef484cc7f9a5f88d0d8555acc93fdR223)
 2. The "searchQueries" array in the search modal [here](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2658/files#diff-c12093d0d555baf8f446bd24643f1dd1R26)
 3. The server-side label that is passed to FilterParser [here](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2658/files#diff-4737dbdec30279777bd8d18c425035f8R445).

We have used non-column names before for our filters.  If you run into trouble, you can check out the [JournalService's filters](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/journal/journal.service.js#L117) which includes the non-column filters `includeNonPosted` and `hrEntity`.

client/src/modules/patients/registry/search.modal.html
```diff
@@ -102,6 +102,20 @@
...
+          <!-- location -->
+
+          <div class="form-group" ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.originLocationLabel.$invalid }">
+            <label class="control-label" translate>
+                FORM.LABELS.ORIGIN_LOCATION
+            </label>
+            <bh-clear on-clear="$ctrl.clear('originLocationLabel')"></bh-clear>
+
+            <input type="text" class="form-control" name="originLocationLabel" ng-model="$ctrl.searchQueries.originLocationLabel">
+            <div class="help-block" ng-messages="ModalForm.reference.$error" ng-show="ModalForm.$submitted">
```
jniles: I'm pretty sure this should be `ng-messages="ModalForm.originLocationLabel.$error"` instead of using `reference`.  That portion should match the `name` on the input.

jniles (commented):
One more tiny change to make sure validation works, and then we can merge this.

jniles (approved):
LGTM.

bors r+

Issue #2350, opened 2017-12-06
Add support for FireFox


jniles:
Mozilla has been the traditional champions of the free web and we were originally planning on using [FireFox in Kiosk mode](http://openkiosk.mozdevgroup.com/) to limit a user's ability to mess with their machine.  However, the last few years of FireFox's growth were not encouraging.

Good news though, [FireFox is now cool again](https://blog.mozilla.org/blog/2017/11/14/introducing-firefox-quantum/)!  It looks like Mozilla is channeling a lot of work into the browser, and we can expect speed and stability in the future.

For this reason, I propose that we investigate adding support for FireFox.  Here's how we would make that happen:
 1. Go through most functionality on FireFox and document breaking issues.  We already know that the `print()` method is broken on embedded frames ... but what else works/doesn't work?  Each issue we find should be reported on Github with the "FireFox" label.
 2. Fix the reported issues.
 3. Enable protractor builds using FireFox.
 4. Enable Travis CI testing using FireFox.

Once all these are done, we'll be able to reliably say we support both major browers.  Chances are, if the application works on both FF and Chrome, it will likely work in MS Edge too!

jniles:
Well, it looks like we won't have a lot of luck with this without severely modifying our server, which is a shame.

This boils down to two issues:
 1. Allowing scripts to load content from cross domains.  We can do this with the `cors` module in expressjs.
 2. A [pretty significant bug in FireFox](https://stackoverflow.com/a/33824109).  There's no work around for this besides embedding JS in the PDF and hacking on it a bit.

Guess we'll let another six months go by before trying again.

Scholldier:
@jniles Just curious -- I use FireFox on my Ubuntu VM and haven't experienced any issues that I would attribute to Firefox. Although, to be fair, I haven't done any printing or heavy use. It's also been close to 6 months since your last post -- has the issue you linked been fixed?

jniles:
Hi @Scholldier!  Welcome back :)

The system is usable on FF for the most part, but the print issues make training hard.  Basically, the bug revolves around the inability to tell embedded PDFs that they should be printed.  Typically we have a large button that says "Print" (since we can translate this) which we tell all users to click to print.  In FF, this functionality is broken ([bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=911444) here).  This makes training unacceptably confusing for users.

[...]

PR #2664, opened 2018-04-05
fix(bhPDFPrint): anticipate memory leak

jniles:
This PR cleans up the bhPDFPrint by removing unused bits and migrating to modern JS.

This commit uses the $onDestroy() hook of angularjs components to clean up the event handlers for printing.  It also defers the load() event to prevent FireFox from automatically attempting to print the window before any PDFs are loaded.

Note that the print() function is still broken in FF, but this allows you to preview modules without print() calls suddenly hijacking the session.

Partially addresses #2350.

### DIFF
--- a/client/src/js/components/bhPDFPrint.js
+++ b/client/src/js/components/bhPDFPrint.js
@@ -2,15 +2,15 @@ angular.module('bhima.components')
   .component('bhPdfPrint', {
     bindings : {
       pdfUrl       : '@',
-      disableCache : '@',
-      options      : '<',
-      disabled     : '<',
+      disableCache : '@?',
+      options      : '<?',
+      disabled     : '<?',
     },
     template :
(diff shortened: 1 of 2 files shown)


client/src/js/components/bhPDFPrint.js
```diff
@@ -64,93 +66,95 @@ bhPDFPrintController.$inject = [
...
           Notify.danger('FORM.WARNINGS.DOC_NOT_FOUND');
         }
 
       });
   }
 
+  // ensure the blob is cleared when this $scope is cleaned up
+  $ctrl.$onDestroy = () => {
+    if ($ctrl.blobFileURL) {
+      URL.revokeObjectURL($ctrl.blobFileUrl);
```
mbayopanda, line 142: 👍 good catch

mbayopanda (approved):
LGTM

PR #2672, opened 2018-04-09
feat(inventory registry) search modal remembers filters

### LINKED ISSUE #2667: Inventory Search Modal does not remember filters
Our search modals typically "remember" filters, meaning that after they are closed, they will re-open with the same filters selected.  On the Inventory Search Modal, this is not the case.  Furthermore, it appears that clicking "x" on a single filter will remove multiple of the filters.

jeremielodi:
closes #2667

### DIFF
--- a/client/src/modules/inventory/list/modals/search.modal.js
+++ b/client/src/modules/inventory/list/modals/search.modal.js
@@ -14,18 +14,18 @@ InventorySearchModalController.$inject = [
  * applied to the grid can be passed in via the filters inject.
  */
 function InventorySearchModalController(ModalInstance, Notify, filters, Inventory, Store, util) {
-  var vm = this;
+  const vm = this;
 
   // @TODO ideally these should be passed in when the modal is initialised these are known when the filter service is defined
-  var searchQueryOptions = [
-    'code', 'group_uuid', 'consumable', 'locked', 'label', 'type_id', 'price',
+  const searchQueryOptions = [
+    'code', 'group_uuid', 'consumable', 'text', 'locked',
+    'label', 'type_id', 'price',
   ];
 
-  var changes = new Store({ identifier : 'key' });
+  const changes = new Store({ identifier : 'key' });
 
   // displayValues will be an id:displayValue pair
-  var displayValues = {};
-  var lastDisplayValues = Inventory.filters.getDisplayValueMap();
+  const displayValues = {};
 
   vm.filters = filters;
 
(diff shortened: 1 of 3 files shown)


client/src/modules/inventory/list/modals/search.modal.js
```diff
@@ -68,17 +66,17 @@ function InventorySearchModalController(ModalInstance, Notify, filters, Inventor
...
     // push all searchQuery values into the changes array to be applied
-    angular.forEach(vm.searchQueries, function (value, key) {
-      if (angular.isDefined(value)) {
+    angular.forEach(vm.searchQueries, (val, _key) => {
+      if (angular.isDefined(val)) {
         // default to the original value if no display value is defined
-        var displayValue = displayValues[key] || lastDisplayValues[key] || value;
-        changes.post({ key: key, value: value, displayValue: displayValue });
+        const displayVal = displayValues[_key] || val;
+        changes.post({ key : _key, value : val, displayValue : displayVal });
```
jniles: ESLint will throw an error here because newer JavasScript allows you to combine the key and the value if they are named the same.  Like this:
```js
// before:
var key = '1';
var o =  { key : key };
// o is { 'key' : '1' }

// after
var key = 2;
var o = { key };
// o is now { 'key' : 2 }
```

jniles (changes requested):
Looks pretty good.

There's one bug here, caused by the removal of `lastDisplayValues`.  This is used to make sure that value are saved even if the filter modal is re-submitted.  See the GIF for a demonstration of this.




The fix is described [here](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/patients/registry/search.modal.js#L93).

client/src/modules/inventory/list/modals/search.modal.js
jeremielodi: I wonder if I should update my ESlint library in this  case

jniles: Ah - I meant the previous code.  Not the code that you updated :)

I'm just explaining what the error used to be - the code you wrote is fine.

jniles:
@jeremielodi how is work coming on this?

jniles (approved):

jniles:
LGTM.  Well done!

bors r+

jeremielodi:
@jniles, Depots' test problem affects also this PR

jniles:
Yeah.  See if you can figure it out while I work on a review of https://github.com/IMA-WorldHealth/bhima-2.X/pull/2703,

PR #2684, opened 2018-04-13
refactor(depot): add location and link to stock

### LINKED ISSUE #2663: refactor(depot): Move the depot in its own folder in the navigation tree
This issue is about : 
- To move the depot module from `administration` to its own folder in the navigation tree
- Update the depot registry with these information : 
  - name
  - location
  - number of articles : link to stock registry
  - is_transit (optional)
  - is_warehouse

mbayopanda:
This PR : 
- Move the depot in its own folder in the navigation tree
- Add location to depot
- Add link to stock registry from depot

close #2663


client/src/modules/depots/depots.js
```diff
@@ -12,9 +12,8 @@ DepotManagementController.$inject = [
  * This controller is about the depot management module in the admin zone
  * It's responsible for creating, editing and updating a depot
  */
-function DepotManagementController(Depots, ModalService,
-  Notify, uiGridConstants, $state) {
-  var vm = this;
+function DepotManagementController(Depots, ModalService, Location, Notify, uiGridConstants, $state) {
```
jniles: It would be better to call this `Locations`, since that has been our practice with most other services (for example, `Depots`), and `location` is a reserved JS word that may confuse future developers.  GitHub's syntax highlighting certainly thinks so!

client/src/modules/depots/templates/label.tmpl.html
```diff
@@ -0,0 +1,10 @@
+<div class="ui-grid-cell-contents">
+  <a data-method="stock-inventories"
+    ui-sref="stockInventories({ filters : [
```
jniles, line 3: This is cool 👍

client/src/modules/depots/templates/label.tmpl.html
```diff
@@ -0,0 +1,10 @@
...
+      { key : 'depot_uuid', value : row.entity.uuid, displayValue : row.entity.text },
+      { key : 'period', value : 'allTime', displayValue : 'PERIODS.ALL_TIME' },
+      { key : 'limit', value : 1000 }
```
jniles: I think you can put `cacheable:false` on all these filters to make sure that the FilterService cdoesn't automatically save them and override whatever the previous user's values were.

client/src/modules/depots/templates/location.tmpl.html
```diff
@@ -0,0 +1,4 @@
+<div ng-if="row.entity.location_uuid"  class="ui-grid-cell-contents">
+  <i class="fa fa-globe"></i> 
+  <b>{{ row.entity.village_name }}</b> / {{ row.entity.sector_name }} / {{ row.entity.province_name }} ({{ row.entity.country_name }})
```
jniles: This is a lot of listeners!  Do we really need all this information?

client/src/modules/depots/templates/label.tmpl.html
jniles: See [FilterService.js#L149](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/FilterService.js#L149) for implementation.

server/controllers/inventory/depots/index.js
```diff
@@ -126,18 +149,24 @@ function list(req, res, next) {
...
+    LEFT JOIN village v ON v.uuid = d.location_uuid
+    LEFT JOIN sector s ON s.uuid = v.sector_uuid 
+    LEFT JOIN province p ON p.uuid = s.province_uuid
+    LEFT JOIN country c ON c.uuid = p.country_uuid 
   `;
 
   filters.custom(
     'user_id',
-    'depot.uuid IN (SELECT depot_permission.depot_uuid FROM depot_permission WHERE depot_permission.user_id = ?)'
+    'd.uuid IN (SELECT depot_permission.depot_uuid FROM depot_permission WHERE depot_permission.user_id = ?)'
```
jniles, line 149: 💃

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -99,6 +99,24 @@
       </div>
     </div>
 
+    <div class="form-group">
+      <div class="checkbox">
```
jniles, line 103: Would a `bhHiddenField` be better here?  It seems like this is what it was designed for.

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -99,6 +99,24 @@
...
+            name="join_location"
+            ng-true-value="1"
+            ng-false-value="0"
+            ng-model="DepotModalCtrl.depot.joinLocation">
+          <span translate>DEPOT.JOIN_LOCATION</span>
+        </label>
+      </div>
+
+      <div ng-if="DepotModalCtrl.depot.joinLocation">
+        <bh-location-select location-uuid="DepotModalCtrl.depot.location_uuid"></bh-location-select>
```
jniles: 👍

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -99,6 +99,24 @@
...
 
+    <div class="form-group">
+      <div class="checkbox">
+        <label>
+          <input
+            type="checkbox"
+            name="join_location"
+            ng-true-value="1"
+            ng-false-value="0"
+            ng-model="DepotModalCtrl.depot.joinLocation">
```
jniles: It looks like `joinLocation` is just a flag for the view to show/hide the depot's location select.  Therefore, you don't need to attach it to the `depot` object - this causes you to reference it on the server.

I propose that you make another variable such as this:
```js
// this will be "true" if the uuid is set
vm.hasDepotLocation = !!(vm.depot.location_uuid);
```
Then you can just toggle `hasDepotLocation` without changing properties on `vm.depot`.

If on the server you need to know if it has a location, you can do the same thing!

test/end-to-end/depots/depots.page.js
```diff
@@ -35,13 +35,21 @@ class DepotPage {
 
   /**
    * simulate the create depot button click to show the dialog of creation
+   * @param {string} name the name of the depot
+   * @param {boolean} hasWarehouse
+   * @param {boolean} joinLocation if true it will enable the option of adding location
+   * @param {array} location an array of location as [country_uuid, province_uuid, sector_uuid, village_uuid]
    */
-  createDepot(name, hasWarehouse) {
+  createDepot(name, hasWarehouse, joinLocation, location) {
```
jniles: Instead of `joinLocation`, I would call this variable `hasLocation`.

test/integration/depots.js
```diff
@@ -53,93 +53,83 @@ describe('(/depots) The depots API ', () => {
...
-    .then((res) => {
-      expect(res).to.have.status(200);
-      expect(res.body[0].text).to.exist;
-      expect(res.body[0].text).to.be.equal(editDepot.text);
-      expect(res.body[0].is_warehouse).to.exist;
-      expect(res.body[0].is_warehouse).to.be.equal(editDepot.is_warehouse);
+      .send(editDepot)
+      .then((res) => {
+        expect(res).to.have.status(200);
+        expect(res.body[0].text).to.be.equal(editDepot.text);
```
jniles, line 89: One way to improve this code (not saying you need to do it in the PR), is to do the following:
```js
const [updated] = res.body;

expect(updated.text).to.equal(editDpot.text); 
// .... etc ....
```

It just makes it easier to read.

jniles (commented):
Some general comments here.  Let me know what you think!

client/src/modules/depots/templates/label.tmpl.html
mbayopanda: I tried to put `cacheable:false` but it seems like there is no effect 😢

client/src/modules/depots/modals/depot.modal.html
mbayopanda: the `joinLocation` flag is required for depot on the server side if we want to remove a location on a depot

mbayopanda:
@jniles could you review this PR again ?

client/src/modules/depots/depots.js
```diff
@@ -68,28 +79,34 @@ function DepotManagementController(Depots, ModalService,
...
-      vm.gridOptions.data = data;
-    })
-    .catch(Notify.handleError)
-    .finally(function () {
-      vm.loading = false;
-    });
+      .then(data => {
+        // format location
+        vm.gridOptions.data = data.map(item => {
+          item.location = item.location_uuid ?
```
jniles, line 85: This is better!

client/src/modules/depots/modals/depot.modal.js
```diff
@@ -6,28 +6,32 @@ DepotModalController.$inject = [
 ];
 
 function DepotModalController($state, Depots, ModalService, Notify) {
-  var vm = this;
+  const vm = this;
 
-  vm.depot = $state.params.depot;
+  vm.depot = $state.params.depot || {};
```
jniles: This default value should probably go in the `routes` file.

client/src/modules/depots/modals/depot.modal.js
```diff
@@ -37,4 +41,12 @@ function DepotModalController($state, Depots, ModalService, Notify) {
   function closeModal() {
     $state.go('depots');
   }
+
+  function clean(depot) {
```
jniles: In the future, this would be better placed in the `DepotService`.  It doesn't really have anything to do with the view, and is only called when `DepotSerivce.create()` or `DepotService.update()` are called.

jniles: You could also delete `hasLocation` here, and that would save you the work of putting `delete hasLocation` on the server.

server/controllers/inventory/depots/index.js
```diff
@@ -88,6 +96,21 @@ function update(req, res, next) {
   // prevent updating the uuid by accident
   if (req.body.uuid) { delete req.body.uuid; }
 
+  // handle location
+  if (req.body.hasLocation === 0) {
```
jniles: Why can't we just send back `location_uuid : null` and that be it?  Why do we need to know `hasLocation` on the server?

jniles (commented):
Just a few more clarifying comments.  It doesn't seem to me like we need `hasLocation` on the server.  In my mind we can either set `depot.location_uuid = 'some-uuid'` or `depot.locaiton_uuid = null;` and avoid the need for the `hasLocation` flag.  Is there a good reason for it?

mbayopanda:

mbayopanda:
@jniles changes landed

client/src/modules/depots/depots.routes.js
```diff
@@ -1,32 +1,32 @@
...
-        depot : { value : null },
-        creating : { value : true },
-      },
-      onEnter : ['$uibModal', depotModal],
-      onExit : ['$uibModalStack', closeModal],
-    })
+      .state('depots.create', {
+        url : '/create',
+        params : {
+          depot : { value : {} },
```
jniles, line 13: :dancer:

client/src/modules/depots/modals/depot.modal.js
```diff
@@ -6,28 +6,36 @@ DepotModalController.$inject = [
...
 
 function DepotModalController($state, Depots, ModalService, Notify) {
-  var vm = this;
+  const vm = this;
 
   vm.depot = $state.params.depot;
   vm.isCreating = !!($state.params.creating);
 
+  if (vm.depot.location_uuid) {
+    vm.hasLocation = 1;
```
jniles, line 15: 🎉

jniles (approved):
LGTM.  All my previous comments have been addressed.  Good work!

bors r+

PR #2688, opened 2018-04-16
feat(role management) assign actions to roles

jeremielodi:
The role management module allow us to assign just permissions to particular pages but not their actions.
Some actions can be sensitives so we should be able to manage them.
This PR tries to solve that problem

### DIFF
--- a/client/src/js/constants/bhConstants.js
+++ b/client/src/js/constants/bhConstants.js
@@ -10,6 +10,9 @@ function constantConfig() {
   var JOURNAL_UTIL_HEIGHT = '150px';
 
   return {
+    actions : {
+      CAN_EDIT_ROLES : 1,
+    },
     accounts : {
       ROOT  : 0,
       ASSET : 1,
@@ -119,8 +122,8 @@ function constantConfig() {
     },
     defaultFilters : [
       { key : 'period', label : 'TABLE.COLUMNS.PERIOD', valueFilter : 'translate' },
-      { key : 'custom_period_start', label : 'PERIODS.START', comparitor: '>', valueFilter : 'date' },
-      { key : 'custom_period_end', label : 'PERIODS.END', comparitor: '<', valueFilter : 'date' },
+      { key : 'custom_period_start', label : 'PERIODS.START', comparitor : '>', valueFilter : 'date' },
+      { key : 'custom_period_end', label : 'PERIODS.END', comparitor : '<', valueFilter : 'date' },
       { key : 'limit', label : 'FORM.LABELS.LIMIT' }],
     weekDays : [
       { id : 0, label : 'FORM.LABELS.WEEK_DAYS.SUNDAY' },
--- a/client/src/modules/roles/modal/roleActions.html
+++ b/client/src/modules/roles/modal/roleActions.html
@@ -1,0 +1,31 @@
+<div class="modal-header">
+    <ol class="headercrumb">
+      <li class="static" translate>FORM.BUTTONS.ACTIONS</li>
+      <li class="title">
+        <span translate>{{RoleActionsCtrl.role.label}}</span>
+      </li>
+    </ol>
+</div>
+<div class="modal-body">
+
+  <div class="row">
+    <div class="col-lg-12">
+      <ul class="list-group">
+        <li  class="list-group-item" ng-repeat="action in RoleActionsCtrl.actions">
+          <label class="radio-inline">
+            <input type='checkbox' ng-model='action.affected' ng-true-value="1" ng-false-value="0" />
+            <span id="{{action.id}}" translate>{{action.description}}</span>
+          </label>
+        </li>
+      </ul>
+    </div>
+  </div>
+</div>
+
+<div class="modal-footer text-right">
+  <button type="button" class='btn btn-default' ng-click='RoleActionsCtrl.close()' translate>FORM.BUTTONS.CLOSE</button>
+ 
+  <button data-method="submit" type="button" class="btn btn-primary" ng-click="RoleActionsCtrl.assignActionToRole()">
+      <span translate>FORM.BUTTONS.SAVE</span>
+  </button>
+</div>
--- a/client/src/modules/roles/modal/roleActions.html
+++ b/client/src/modules/roles/modal/roleActions.html
@@ -1,4 +1,6 @@
-<div class="modal-header">
+<form name="RoleActionForm" bh-submit="RoleActionsCtrl.assignActionToRole()">
+  <div class="modal-header">
+
     <ol class="headercrumb">
       <li class="static" translate>FORM.BUTTONS.ACTIONS</li>
       <li class="title">
@@ -24,8 +26,6 @@
 
 <div class="modal-footer text-right">
   <button type="button" class='btn btn-default' ng-click='RoleActionsCtrl.close()' translate>FORM.BUTTONS.CLOSE</button>
- 
-  <button data-method="submit" type="button" class="btn btn-primary" ng-click="RoleActionsCtrl.assignActionToRole()">
-      <span translate>FORM.BUTTONS.SAVE</span>
-  </button>
+  <bh-loading-button loading-state="RoleActionForm.$loading"></bh-loading-button>
 </div>
+</form>
--- a/client/src/modules/roles/modal/roles.actions.js
+++ b/client/src/modules/roles/modal/roles.actions.js
@@ -1,0 +1,46 @@
+angular.module('bhima.controllers').controller('RoleActionsController', RoleActionsController);
+RoleActionsController.$inject = [
+  'data', '$uibModal', '$uibModalInstance',
+  'RolesService', 'SessionService', 'NotifyService',
+];
+
+function RoleActionsController(data, $uibModal, $uibModalInstance, RolesService, session, Notify) {
+  const vm = this;
+  vm.close = close;
+  vm.role = angular.copy(data);
+  vm.loadRoles = loadRoles;
+  vm.assignActionToRole = assignActionToRole;
+  vm.actions = [];
+
+  // loa all roles
+  function loadRoles() {
+    RolesService.actions(vm.role.uuid)
+      .then(response => {
+        vm.actions = response.data;
+      })
+      .catch(Notify.handleError);
+  }
+
+  // assigned actions to a role
+  function assignActionToRole() {
+    const ids = vm.actions
+      .filter(action => action.affected === 1)
+      .map(action => action.id);
+
+    const param = {
+      role_uuid : vm.role.uuid,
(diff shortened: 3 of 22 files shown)


client/src/modules/roles/modal/roles.actions.js
```diff
@@ -0,0 +1,46 @@
...
+    RolesService.actions(vm.role.uuid)
+      .then(response => {
+        vm.actions = response.data;
+      })
+      .catch(Notify.handleError);
+  }
+
+  // assigned role to he user
+  function assignActionToRole() {
+    const ids = vm.actions
```
jniles: :+1:

client/src/modules/roles/modal/roles.actions.js
```diff
@@ -0,0 +1,46 @@
...
+      .filter(action => action.affected === 1)
+      .map(action => action.id);
+
+    const param = {
+      role_uuid : vm.role.uuid,
+      action_ids : [].concat(ids),
```
jniles: Note that you can also use `action_ids : [...ids]` to the same effect.

client/src/modules/roles/modal/roleActions.html
```diff
@@ -0,0 +1,33 @@
+<div class="modal-header">
+    <ol class="headercrumb">
+      <li class="static" translate>FORM.BUTTONS.ACTIONS</li>
+      <li class="title">
+        <span translate>{{RoleActionsCtrl.role.label}}</span>
```
jniles: Does this need to be translated?  Or is it a user-provided parameter?

client/src/modules/roles/modal/roleActions.html
```diff
@@ -0,0 +1,33 @@
...
+          </label>
+        </li>
+      </ul>
+    </div>
+  </div>
+</div>
+
+<div class="modal-footer">
+  <div class=" pull-right">
+  <button type="button" class='btn inverse' ng-click='RoleActionsCtrl.close()' translate>FORM.BUTTONS.CLOSE</button>
```
jniles: I don't think the `btn inverse` class exists.  Your choices are in the [bootstrap documentation](https://getbootstrap.com/docs/3.3/css/#buttons).  I think we've been using `btn btn-default` for all our "cancel" buttons.

client/src/modules/roles/modal/roleActions.html
```diff
@@ -0,0 +1,33 @@
...
+            <span id="{{action.id}}" translate>{{action.description}}</span>
+          </label>
+        </li>
+      </ul>
+    </div>
+  </div>
+</div>
+
+<div class="modal-footer">
+  <div class=" pull-right">
```
jniles: You can accomplish this same thing without the extra `<div>`.  Just put `text-right` next to the `modal-footer` class.  This will align the buttons to the right of the modal.

client/src/modules/roles/modal/roleActions.html
```diff
@@ -0,0 +1,33 @@
...
+</div>
+<div class="modal-body">
+
+  <div class="row">
+    <div class="col-lg-12">
+      <ul class="list-group">
+        <li  class="list-group-item" ng-repeat="action in RoleActionsCtrl.actions">
+          <label class="radio-inline">
+            <input type='checkbox' ng-model='action.affected' ng-true-value="1" ng-false-value="0" />
+            <span id="{{action.id}}" translate>{{action.description}}</span>
```
jniles, line 19: To avoid collisions with other element ids (since ids are supposed to be unique per page), why don't you add a prefix here?  Something like `id="action-{{action.id}}"`.  That way it will render as `id="action-1"` instead of `id="1"`.

client/src/modules/roles/roles.ctrl.js
```diff
@@ -28,6 +30,20 @@ function RolesController($uibModal, RolesService, session, Modal, Notify) {
...
   };
 
+  vm.editActions = function editActions(role) {
+    $uibModal.open({
+      keyboard : false,
+      backdrop : 'static',
+      templateUrl : 'modules/roles/modal/roleActions.html',
+      controller : 'RoleActionsController as RoleActionsCtrl',
+      resolve : {
+        data : function dataProvider() {
```
jniles: Just FYI, this could be one line:
```js
data : () => role,
```

server/controllers/admin/roles.js
```diff
@@ -151,6 +141,71 @@ function listForUser(req, res, next) {
...
+
+  const roleUuid = db.bid(data.role_uuid);
+
+
+  const deleleUserRoles = `DELETE FROM role_actions WHERE role_uuid=? `;
+  const addAction = `INSERT INTO role_actions SET ?`;
+
+  db.exec(deleleUserRoles, roleUuid)
+    .then(() => {
+      const promisses = actionIds.map(actionId => {
```
jniles: If you are going to add a ton of promises here, it might be a good idea to wrap this all in a `db.transaction()`.  That way, if connection to the database is lost for some reason we do not end up with half the data in an undefined state.

server/controllers/admin/roles.js
```diff
@@ -151,6 +141,71 @@ function listForUser(req, res, next) {
...
+  const actionId = req.params.action_id;
+  const userId = req.session.user.id;
+
+  const sql = `
+    SELECT count(ra.uuid) as nbr FROM role_actions ra
+    JOIN user_role as ur ON ur.role_uuid = ra.role_uuid
+    WHERE actions_id =? AND ur.user_id=?`;
+
+  db.one(sql, [actionId, userId]).then(row => {
+    res.json(row.nbr > 0);
```
jniles: Always include a status code with your responses.  I'm surprised express doesn't throw a warning on this route...

Something like:
```js
res.status(200).send(row.nbr > 0);
```

client/src/js/constants/bhConstants.js
```diff
@@ -10,6 +10,9 @@ function constantConfig() {
   var JOURNAL_UTIL_HEIGHT = '150px';
 
   return {
+    actions : {
+      CAN_EDIT_ROLES : 1,
```
jniles, line 14: Why do you have actions both here and in the database?  It seems like they should be in one place, not in two.

client/src/modules/roles/templates/action.tmpl.html
```diff
@@ -1,4 +1,4 @@
-<div class="ui-grid-cell-contents text-right"
+<div class="ui-grid-cell-contents text-right" ng-show="grid.appScope.canEditRoles"
```
jniles, line 1: 👍

client/src/js/constants/bhConstants.js
jeremielodi: your right but to my mind I thought that using action's id in the controllers is not a good thing that is why I created constants instead

client/src/modules/roles/modal/roles.actions.js
jeremielodi: the comment is not good here

client/src/modules/roles/modal/roleActions.html
jeremielodi: Yes it should be translated, an action should be described and every user have to understand what it means

server/controllers/admin/roles.js
jeremielodi: okay, good idea

jeremielodi:
codeclimate has crashed

client/src/js/constants/bhConstants.js
jniles: Okay, that makes sense.  We'll see how easy it is to continue working this way.

client/src/modules/roles/modal/roleActions.html
```diff
@@ -0,0 +1,31 @@
...
+        </li>
+      </ul>
+    </div>
+  </div>
+</div>
+
+<div class="modal-footer text-right">
+  <button type="button" class='btn btn-default' ng-click='RoleActionsCtrl.close()' translate>FORM.BUTTONS.CLOSE</button>
+ 
+  <button data-method="submit" type="button" class="btn btn-primary" ng-click="RoleActionsCtrl.assignActionToRole()">
```
jniles: It would be really cool if this could be a `bh-loading-button`.  That way you could get a loading indicator on a slow connection.  For example, see the GIF below of what this modal looks like on a slow connection:


_Fig 1: Slow behavior on a 3G connection_

You can simulate this in your Chrome Developer Tools by changing the network speed ([docs](https://developers.google.com/web/tools/chrome-devtools/network-performance/network-conditions)).

jniles: To put a `bh-loading-button` in this, you will need to do a couple of things:
 1. Use the `bh-submit` directive on a form that wraps the modal.  Make sure that the `bh-loading-button` is inside the form.  It should look something like this:
```html
<form name="SomeForm" bh-submit="$ctrl.submit()">

  <!-- put HTML as the body of the form -->

  <bh-loading-button loading-state="SomeForm.$loading"></bh-loading-button>
</form>
```
 2. Make sure `<bh-loading-button>` has the `loading-state` listening on the `$loading` property of the form.
 3. Make sure that the `bh-submit` function returns a promise.

And it should work!

jniles (changes requested):
The code is looking better!

I've made some comments inline.  We try to use loading indicators wherever possible since network latency can be a problem at our installation sites.

It doesn't look like there has been any tests updated - we probably should have both integration tests and end to end tests this functionality.

The integration tests should test the responses for the URLs you added here:
 1. `GET /roles/actions/:roleUuid`
 2. `GET /roles/actions/user/:action_id`
 3. `GET /roles/user/:user_id/:project_id`
 4.  `POST /roles/actions`

The end to end tests should test the functionality of opening the modal and assigning actions.

jeremielodi:
@jniles, I added tests but depot's bug keeps troubling every PR

jniles:
@jeremielodi, thanks.  We should figure out what is going on with that depot test suite.

client/src/modules/roles/modal/userRole.html
```diff
@@ -25,9 +26,7 @@
 <div class="modal-footer">
   <div class=" pull-right">
   <button type="button" class='btn inverse' ng-click='UsersRolesCtrl.close()' translate>FORM.BUTTONS.CLOSE</button>
- 
-  <button data-method="submit" type="button" class="btn btn-primary" ng-click="UsersRolesCtrl.assignRolesToUser()">
-      <span translate>FORM.BUTTONS.SAVE</span>
-    </button>
+  <bh-loading-button loading-state="RolesForm.$loading"></bh-loading-button>
```
jniles, line 29: :+1:

jniles (approved):
Let's see if bors passes now that we have those dept tests worked out.

bors r+

PR #2690, opened 2018-04-16
feat(unit test) test for uuid service

### LINKED ISSUE #2646: Write unit tests for uuid service
The [uuid](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/uuid.js) service would be a quick win for anyone wanting to practice writing unit tests.  The following tests need to be written:

 1. The service should return a function.
 2. The function should return a string with the number `4` in position `14`.
 3. Calling the function multiple times should returns unique values.

jeremielodi:
closes #2646

### DIFF
--- a/test/client-unit/services/uuid.spec.js
+++ b/test/client-unit/services/uuid.spec.js
@@ -1,0 +1,29 @@
+/* global inject, expect */
+describe('uuid', () => {
+
+  let UUID;
+  beforeEach(module('bhima.services'));
+
+  beforeEach(inject(_uuid_ => {
+    UUID = _uuid_;
+  }));
+
+  it('UUID() should be a function', () => {
+    expect(UUID).to.be.a('function');
+  });
+
+  it('The forteen carator must be equal to 4', () => {
+    const uuid = UUID();
+    expect(uuid.slice(14, 15)).to.be.equal('4');
+  });
+
+  it('Should compare generated uuids, they should be unique', () => {
+    const uuid1 = UUID();
+    const uuid2 = UUID();
+    const uuid3 = UUID();
+    expect(uuid1).to.not.equal(uuid2);
+    expect(uuid2).to.not.equal(uuid3);
+    expect(uuid1).to.not.equal(uuid3);
+  });
+
+});


test/client-unit/services/uuid.spec.js
```diff
@@ -0,0 +1,20 @@
...
+describe('Store', () => {
+
+  let UUID;
+  beforeEach(module('bhima.services'));
+
+  beforeEach(inject(_uuid_ => {
+    UUID = _uuid_;
+  }));
+
+  it('Should compare generated uuids, they should be unique', () => {
```
jniles: This is the right idea!

test/client-unit/services/uuid.spec.js
```diff
@@ -0,0 +1,20 @@
+/* global inject, expect */
+describe('Store', () => {
```
jniles: The `describe()` block exists to describe what is being tested.   You should always update it to say what component is under test.  In this case, it is the 'Uuid' service.

jniles (changes requested):
The code so far looks pretty good!

There are two more tests required by https://github.com/IMA-WorldHealth/bhima-2.X/issues/2646.

The first one is to check that `UUID()` is a function.  You can do this really simply:
```js
expect(UUID).to.be.a('function');
```
You find the methods you can use with `expect()` in the [chai.js documentation](http://www.chaijs.com/api/bdd/#atype-msg).

The second test is to see if this is really a version 4 uuid.  You can read about what a version 4 uuid is [on wikipedia](https://fr.wikipedia.org/wiki/Universal_Unique_Identifier#Version_4_.28random.299).  The [English version](https://en.wikipedia.org/wiki/Universal_Unique_Identifier#Version_4_.28random.299) has a bit more detail.  Basically, we need to check that the uuid creates a string with the number `4` at index `14`.  The test will look something like this:
```js
const uid = UUID();
expect(uid[14]).to.equal('4');
```

jniles (approved):

jniles:
Nice!  See?  Tests are easy :)

💃 


bors r+

PR #2696, opened 2018-04-17
refractor(bhDateEditor component) add allow-future-date flag 

### LINKED ISSUE #2677: Proposal: allow-future-date flag on bhDateEditor
I'm currently working on https://github.com/IMA-WorldHealth/bhima-2.X/issues/1960, and noticed that we use the following pattern in a lot of places:
```js
// in the main module's controller:
vm.timestamp = new Date();  // or vm.today = new Date();
```
```html
<!-- and then in the HTML -->
<bh-date-editor
  max-date="ModuleCtrl.timestamp">
</bh-date-editor>
```
The purpose of this pattern is to prevent future dates from being registered. 

Since we use this pattern so much (and `vm.timestamp` is never used again in the controller), we should just move it into the component and put it behind a flag.  Technically, it is giving the directive instructions.

I propose we put another optional binding (in addition to `max-date`) that is `allow-future-date`, which is false by default.  This will remove a lot of cruft from our controllers and enable them to focus on logic instead of putting variables that are not used.

jeremielodi:
closes #2677

### DIFF
--- a/client/src/js/components/bhDateEditor.js
+++ b/client/src/js/components/bhDateEditor.js
@@ -7,6 +7,7 @@ angular.module('bhima.components')
       onChange : '&',
       minDate : '<?',
       maxDate : '<?',
+      allowFutureDate : '<?',
       validationTrigger : '<?',
       disabled : '<?',
       dateFormat : '@?',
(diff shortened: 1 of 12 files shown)


client/src/js/components/bhDateEditor.js
```diff
@@ -7,6 +7,7 @@ angular.module('bhima.components')
       onChange : '&',
       minDate : '<?',
       maxDate : '<?',
+      allowFutureDate : '<',
```
jniles: This should be optional `<?`.

client/src/js/components/bhDateEditor.js
```diff
@@ -44,12 +45,15 @@ function bhDateEditorController(bhConstants) {
...
   $ctrl.$onInit = () => {
     $ctrl.label = $ctrl.label || 'FORM.LABELS.DATE';
-
+    $ctrl.allowFutureDate = $ctrl.allowFutureDate || false;
     // options to be passed to datepicker-option
     $ctrl.options = {
       minDate : $ctrl.minDate,
       maxDate : $ctrl.maxDate,
     };
+    if ($ctrl.allowFutureDate) {
```
jniles: Instead of deleting this, why not just add it in if `$ctrl.allowFutureDate` doesn't exit?  Like this:
```js
$ctrl.options = {
  minDate : $ctrl.minDate,
};

if ($ctrl.allowFutureDate) {
  $ctrl.options.maxDate = $ctrl.maxDate;
}
```

client/src/modules/cash/cash.html
```diff
@@ -66,14 +66,14 @@
                   <span class="text-center" translate>DEBTOR_GROUP.CONVENTION.CANNOT_PAY</span>
                   <strong class="text-center">{{CashCtrl.Payment.messages.patientConventionName}}</strong>
                 </div>
-
+ 
                 <!-- allowed for cash payment -->
                 <fieldset ng-disabled="CashCtrl.Payment.messages.isNonCashPatient">
                   <bh-date-editor
+                    allow-future-date = "false"
```
jniles: If this defaults to `false`, there isn't any need to put it.

jniles (changes requested):
Looks clean.  Well done!

Could you write a couple tests to show how this flag works?

jniles (approved):

jniles:
@jeremielodi, well done.  These are very nice tests!

Could you rebase this PR to master to get rid of those snyk errors?

It looks like the depot tests are causing errors on all PRs today.  I'm not sure what's going on with them...

PR #2698, opened 2018-04-17
Fix the cross-env NODE_ENV for windows OS

### LINKED ISSUE #2697: Fix scripts in package.json for running the application on windows
This issue is about to update the package.json for running correctly the application on Windows.

mbayopanda:
This PR fix the issue #2697 

Close #2697

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "description": "A rural hospital information management system.",
   "main": "npm run build",
   "scripts": {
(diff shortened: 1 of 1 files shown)


jniles:
This will work!

It looks like we could also use [cross-env-shell](https://www.npmjs.com/package/cross-env#cross-env-vs-cross-env-shell) to do it.

PR #2702, opened 2018-04-18
Add description in integration test for stock

### LINKED ISSUE #2293: Stock entry and exit do not create a description in the voucher table.
When the stock modules record an exit or entry, they do not appear to write a description in the voucher table.  It would be nice to include a description with the following elements:
 1. The user who authorized it.
 2. For stock exits - the thing it went to (patient, pharmacy, etc)
 3. For stock entry - the thing it came from (pharmacy, supplier, etc).
 4. The number of inventory items affected ("distributed X items")

That way, it is easy for a user to audit our voucher registry.

mbayopanda:
This PR add description in the integration test for the stock module.

close #2293

### DIFF
--- a/test/integration/stock.js
+++ b/test/integration/stock.js
@@ -83,6 +83,7 @@ describe('(/stock/) The Stock HTTP API', () => {
     date       : new Date(),
     flux_id    : 1,
     user_id    : 1,
+    description : 'Initial Lot Movement',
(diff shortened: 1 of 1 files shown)


jniles:
LGTM.  This is going to be super helpful for my work 💃 

bors r+

PR #2703, opened 2018-04-19
Payroll process manager

### LINKED ISSUE #2558: Payroll Multiple
This issue proposes to redo the module Payroll Multiple of Bhima 1.X whose main interface is presented in the following way
payroll1

in the first version of Bhima it was question of selection the Period as well as the currency to be used for the payment of the employees

The Multiple Payroll module displayed a list of all employees who are not yet paid for the pay period select
payroll2
payroll3

By default all employees were considered as present, and for those who had periods of leave, their number of days of their attendance was deducted from the number of working days of their vacation period, the IPR tax was calculated automatically and the other items was either deducted or calculated automatically

lomamech:
- Update bhCurrencyInput add property disabled
- Update bhEmployeeSelect use EmployeeUuid instead EmployeeId
- Implement component Payroll Period Select
- Implement component Payroll Status Select
- Remove not necessary table config_paiment_period
- Uptable employee use uuid instead id for Employee Identification
- Create table employee advantage for manage rubrics payroll defined by
  Employee
- Change all employee_id by employee_uuid
- Add table offday_paiement for track all value for OffDay duiring Payroll
- Update table paiement
- Add Table paiement status
- Add Table Rubric Paiement for track all value for Rubrics duiring
  Payroll
- Implement Service for Receipt of Payslip
- Update employee module use uuid instead id
- Remove employee daily salary
- Add Property individual_salary
- Update Holiday management use employee_uuid instead employee_id
- Update employees route use uuid instead id
- Update Multiple Payroll API route
- Update Holidays management use employee_uuid
- Update and Upgrade offday management
- Implement API for Multi Payroll Management
- Update Employee Management
- Update holidays management
- Update templates for bhEmployeeSelect
- Implement API for Payroll Process Manager
- Update integration tests for Employee and Holidays
- Update tax ipr configuration select set property not required
- Update Payroll configuration service add service paiementStatus
- Update Rubric Cretion add new property Is defined by Employee
- Update bhima.sql add node for Multiple Payroll Management
- Add in bhima.sql data for paiement status
- Update complex voucher, add toolkit for Employee Paiement
- Upgrate complexe voucher creat, update paiement status if voucher
  paiement type is Salary Paiement
- Implement integration test for Payroll Configuration
- Implement integration test for Payroll MultiConfiguration
- Implement integration test for Payroll Commitment
- Update data.sql
- Update integration test for Employees, Debtors, Patient, voucher
- Improve rubrics Payroll Api integrate filters for rubrics defined by Employee
- Sanitaze code for Holidays management
- Update voucher creation for Salary Paiement, set element Paiement Date
- Updata initial data in Database test, set new element
- Update test integration Employee set uuid for easy integration on
  Payroll Paiement
- ReImplement Data for Payroll Configuration, MultiConfiguration and
  Commitment of Paiement
- Implement test integration for Paiement salary by Voucher ID
- Update the calcul of Holiday valide days, Check if in a holiday period
  there is not a offday
- Update E2E test for Employee, set Value for Rubric Payroll defined By
  Employee
- Update form Employee Registration adding ID in employee.payroll Element
- Update number of data in Employee Registry, Test
- Update E2E test for Update Employee
- Update E2E test for manage holiday for Employee use uuid instead Id
- Skip E2E test between Patient Registry and Invoice Registry

closes #2558


jniles:
@lomamech, let me know once the lint issues are fixed and I'll give you a review.  It will be easier for both of us that way :)

client/src/i18n/en/form.json
```diff
@@ -590,6 +608,7 @@
             "WEEKEND_CONFIGURATION" : "Weekend Configuration",
             "WEIGHT": "Weight",
             "WIDTH": "Width",
+            "WORKING_DAY" : "Working days",
```
jniles: A better translation in English would be "Days of Work"

client/src/js/components/bhCurrencyInput.js
```diff
@@ -7,6 +7,7 @@ angular.module('bhima.components')
       model : '=', // two way binding
       validationTrigger : '<', // one-way binding
       label : '@?',
+      disabled : '<',
```
jniles: This binding should be optional `<?`.

client/src/js/components/bhEmployeeSelect.js
```diff
@@ -4,7 +4,7 @@ angular.module('bhima.components')
     controller  : EmployeeSelectController,
     transclude  : true,
     bindings    : {
-      employeeId           : '<',
+      employeeUuid     : '<',
```
jniles, line 7: 💃

lomamech:
@jniles, In pending review...

jniles:
Yes sir!  It's a lot of code to review :).  I'm working on it.

client/src/js/components/bhPayrollPeriodSelect.js
```diff
@@ -0,0 +1,63 @@
...
+function PayrollPeriodSelectController(PayrollConfigs, $timeout, $scope, Notify) {
+  const $ctrl = this;
+
+  // fired at the beginning of the Payroll Configuration select
+  $ctrl.$onInit = function $onInit() {
+    // translated label for the form input
+    $ctrl.label = $ctrl.label || 'FORM.LABELS.PERIOD_PAYMENT';
+
+    // fired when an Payroll Period configuration has been selected
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
```
jniles: Since this is a required binding, you shouldn't put a default value.  If the developer forgets to pass in the callback, that is an error.

client/src/js/components/bhPayrollPeriodSelect.js
```diff
@@ -0,0 +1,63 @@
...
+
+    // default for form name
+    $ctrl.name = $ctrl.name || 'PayrollPeriodForm';
+
+    if (!angular.isDefined($ctrl.required)) {
```
jniles, line 37: Nice 👍

client/src/js/components/bhPayrollPeriodSelect.js
```diff
@@ -0,0 +1,63 @@
...
+  function aliasComponentForm() {
+    $scope.PayrollPeriodForm = $scope[$ctrl.name];
+  }
+
+  // fires the onSelectCallback bound to the component boundary
+  $ctrl.onSelect = function onSelect($item) {
+    $ctrl.onSelectCallback({ payrollConfig : $item });
+
+    // alias the PayrollPeriodForm name so that we can find it via filterFormElements
+    $scope[$ctrl.name].$bhValue = $item.id;
```
jniles, line 61: 👍

client/src/js/components/bhPayrollStatusSelect.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+function PayrollStatusSelectController(Payroll, Notify, $translate) {
+  const $ctrl = this;
+
+  $ctrl.$onInit = function onInit() {
+    // label to display
+    $ctrl.label = $ctrl.label || 'FORM.LABELS.STATUS';
+
+    // fired when a purchase status has been selected or removed from the list
+    $ctrl.onChange = $ctrl.onChange || angular.noop;
```
jniles: Since this is required, we shouldn't set a default value for it.

client/src/js/components/bhPayrollStatusSelect.js
```diff
@@ -0,0 +1,50 @@
...
+
+    // init the model
+    $ctrl.selectedPayrollStatus = $ctrl.statusId || [];
```
jniles: Is the `statusId` an array or a single value?  If it is an array, it should probably be called "statusIds" (in plural) to show that it can be multiple.

client/src/modules/accounts/accounts.js
```diff
@@ -140,6 +144,10 @@ function AccountsController($rootScope, $timeout, AccountGrid, Notify, Constants
   }
 
   function bindGridData() {
+    vm.Accounts.data.forEach(account => {
+      account.translation_key = $translate.instant(account.translation_key);
```
jniles: Why was this necessary...?  If we need to use the translated label, it should be available in the [hrLabel](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/accounts/accounts.service.js#L68) property.

client/src/modules/employees/employee.service.js
```diff
@@ -57,6 +58,11 @@ function EmployeeService(Filters, $uibModal, Api, AppCache, Languages, $httpPara
     }
   }
 
+  function advantage(uuid) {
+    var url = ''.concat(uuid, '/advantage');
```
jniles: This should be `const`.

client/src/modules/employees/employees.routes.js
```diff
@@ -7,7 +7,7 @@ angular.module('bhima.routes')
         templateUrl : 'modules/employees/registration/employees.html',
       })
       .state('employeeEdit', {
-        url         : '/employees/:id/edit',
+        url         : '/employees/:uuid/edit',
```
jniles, line 10: 💃

client/src/modules/employees/registration/employees.html
```diff
@@ -459,25 +458,43 @@ <h4 translate>FORM.LABELS.CURRENT_LOCATION</h4>
                   <!-- Employee daily salary -->
                   <div
                     class="form-group"
-                    ng-class="{'has-error' : DetailForm.daily_salary.$invalid && EmployeeRegistrationForm.$submitted}">
+                    ng-class="{'has-error' : DetailForm.individual_salary.$invalid && EmployeeRegistrationForm.$submitted}">
```
jniles: This is a nicer name!  Good improvement 👍

client/src/modules/employees/registration/employees.html
```diff
@@ -459,25 +458,43 @@ <h4 translate>FORM.LABELS.CURRENT_LOCATION</h4>
...
+                    <div class="form-group">
+                      <label class="col-md-3 control-label" translate>{{ rubric.label}}</label>
+                      <div class="col-md-9">
+                        <div class="input-group">
+                          <span class="input-group-addon text-uppercase">{{ EmployeeCtrl.enterprise.currencySymbol || '...' }}</span>
+                          <input
+                            type="number"
+                            class="form-control"
+                            name="payroll"
+                            id="{{rubric.abbr}}"
```
jniles: Are rubric abbreviations unique?  If not, we probably shouldn't use an `id` here.

client/src/modules/employees/registration/employees.js
```diff
@@ -1,43 +1,55 @@
...
+  'bhConstants', 'ReceiptModal', 'SessionService', 'RubricService',
 ];
 
-function EmployeeController(Employees, Services, Grades, Functions, CreditorGroups, util, Notify, $state, bhConstants, Receipts, Session) {
-  var vm = this;
-  var referenceId = $state.params.id;
-
+function EmployeeController(Employees, Services, Grades, Functions, CreditorGroups, util, Notify, $state, bhConstants, Receipts, Session, Rubrics) {
+  const vm = this;
+  const referenceId = $state.params.uuid;
```
jniles: A better name for this variable would be `referenceUuid`.

client/src/modules/employees/registration/employees.js
```diff
@@ -1,43 +1,55 @@
...
-      vm.employee = employee;
-    })
-    .catch(function (error) {
+      .then((employee) => {
+        formatEmployeeAttributes(employee);
+        vm.origin = employee.hospital_no;
+        vm.employee = employee;
+        vm.employee.payroll = {};
+
+        return Employees.advantage(referenceId);
```
jniles: What is an advantage?  Can you leave a comment in the code explaining what this is getting?

client/src/modules/holidays/modals/holiday.modal.html
```diff
@@ -15,7 +15,7 @@
   <div class="modal-body">
     <div class="form-group">
       <bh-employee-select
-        employee-id="HolidayModalCtrl.holiday.employee_id"
+        employee-uuid="HolidayModalCtrl.holiday.employee_uuid"
```
jniles, line 18: 👍

client/src/modules/multiple_payroll/modals/config.modal.html
```diff
@@ -0,0 +1,78 @@
...
+      <li class="title" translate>TREE.MULTI_PAYROLL</li>
+      <label class="badge badge-success">{{ ConfigPaiementModalCtrl.employee.display_name }}</label>
+      <label class="badge badge-primary"><span translate> {{ ConfigPaiementModalCtrl.label }} </span> : {{ ConfigPaiementModalCtrl.displayValue }}</label>
+    </ol>
+  </div>
+
+  <div class="modal-body" style="max-height: 70vh; overflow:auto;">
+    <div class="form-group">
+      <label class="control-label" translate>FORM.LABELS.GRADE</label>
+      <input disabled="disabled" name="code_grade" ng-model="ConfigPaiementModalCtrl.employee.code_grade" autocomplete="off" class="form-control">
```
jniles: Is this always disabled?  If so, it might be better to make this [a static input](http://getbootstrap.com/docs/3.3/css/#forms-controls-static).

client/src/modules/multiple_payroll/modals/config.modal.html
```diff
@@ -0,0 +1,78 @@
...
+
+    <!-- currency selection -->
+    <bh-currency-select
+      currency-id="ConfigPaiementModalCtrl.currencyId"
+      validation-trigger="ConfigPaiementForm.$submitted"
+      on-change="ConfigPaiementModalCtrl.setCurrency(currency)">
+    </bh-currency-select>
+
+    <div ng-repeat="rubConfigured in ConfigPaiementModalCtrl.rubConfigured" class="form-group">
+      <bh-currency-input
```
jniles, line 57: This is cool. 👍

client/src/modules/employees/registration/employees.html
```diff
@@ -459,25 +458,43 @@ <h4 translate>FORM.LABELS.CURRENT_LOCATION</h4>
...
                         <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
                       </div>
                     </div>
                   </div>
 
+                  <div ng-repeat="rubric in EmployeeCtrl.rubrics">
+                    <div class="form-group">
+                      <label class="col-md-3 control-label" translate>{{ rubric.label}}</label>
+                      <div class="col-md-9">
+                        <div class="input-group">
```
jniles: It looks like [you used bh-currency-input](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2703/files#diff-f96951b2e4b3ca0f2ccce186159382acR57) in the payment configuration modal.  Can you use it here too?  If not, why not?

client/src/modules/multiple_payroll/modals/config.modal.html
```diff
@@ -0,0 +1,78 @@
...
+        currency-id="ConfigPaiementModalCtrl.currencyId"
+        model="ConfigPaiementModalCtrl.payroll.value[rubConfigured.abbr]"
+        validation-trigger="ConfigPaiementForm.$submitted"
+        min = 0>
+      </bh-currency-input>
+    </div>
+  </div>
+
+  <div class="modal-footer">
+    <button data-method="cancel" type="button" class="btn btn-default" ui-sref="multiple_payroll">
```
jniles, line 70: Interesting use of `ui-sref` to cancel the modal. :D

client/src/modules/vouchers/toolkit/payment_employee/payment_employee.js
```diff
@@ -0,0 +1,174 @@
...
+  // submission
+  function submit(form) {
+    if (form.$invalid) { return; }
+
+    const selected = vm.gridApi.selection.getSelectedRows();
+    const bundle = generateTransactionRows({
+      account_id : vm.account_id,
+      paiements  : selected,
+    });
+    const msg = `${$translate.instant('VOUCHERS.GLOBAL.PAYMENT_EMPLOYEES')} [ ${vm.dateFrom}]`;
```
jniles: Instead of using a template string, why don't you use [angular-translate's variable replacement](https://angular-translate.github.io/docs/#/guide/06_variable-replacement)?

The basic idea is that you put in your template:
```js
{
  "PAYMENT_EMPLOYEES" : "I want to substitute {{variable}}",
}
```
and then call it in the controller like this:
```js
const msg = $translate.instant('PAYMENT_EMPLOYEES', { variable : 'HI' });
// renders => "I want to substitute HI"
```

The code you have written is a bit hard to read, since it has lots of `[]`. `$`, and `{}` in it.  It could look like this:
```js
const msg= $translate.instant('VOUCHERS.GLOBAL.PAYMENT_EMPLOYEES', { date : vm.dateFrom });
```
... which is quite a bit easier to read!

server/controllers/payroll/reports/multipayroll.js
```diff
@@ -0,0 +1,80 @@
+
+/**
+ * @overview reports/payroll/multipayroll
+ *
+ * @description
+ * This file contains code to create a PDF report of all MultiPayroll registrations,
+ * matching query conditions passed from the multi Payroll UI grid.
+ *
+ * @requires lodash
+ * @requires moment
```
jniles: Do you use moment here?

client/src/modules/employees/registration/employees.js
```diff
@@ -82,52 +94,48 @@ function EmployeeController(Employees, Services, Grades, Functions, CreditorGrou
...
-
+  }
 
   // Loading Grades
-  Grades.read(null, { detailed : 1 }).then(function (data) {
-    data.forEach(function (g) {
-      g.format = g.code + ' - ' + g.text;
+  Grades.read(null, { detailed : 1 }).then((data) => {
+    data.forEach((g) => {
+      g.format = `${g.code} - ${g.text}`;
```
jniles, line 113: 👍

client/src/modules/multiple_payroll/modals/config.modal.html
jniles: The same goes for the rest of the inputs on this modal that are `disabled`.

server/models/bhima.sql
```diff
@@ -173,14 +178,22 @@ INSERT INTO `flux` VALUES
   (12, 'STOCK_FLUX.TO_ADJUSTMENT'),
   (13, 'STOCK_FLUX.FROM_INTEGRATION');
 
--- transaction type
+-- Purchase Status
```
jniles: 💃

test/integration/payoll.js
```diff
@@ -0,0 +1,382 @@
...
+  it('POST /MULTI_PAYROLL/:ID/CONFIGURATION should Set Configuration of Paiement', () => {
+    return agent.post('/multiple_payroll/'.concat(params.periodPaie, '/configuration'))
+      .send(dataConfiguration)
+      .then((res) => {
+        expect(res).to.have.status(201);
+      })
+      .catch(helpers.handler);
+  });
+
+  it('POST /MULTI_PAYROLL/:ID/COMMITMENT should Set Configuration of Paiement', () => {
```
jniles: The URL here should be the same as the one in the test, to help us debug failures.  It should read `POST /multiple_payroll/:id/commitment)`.   The same should apply to all tests in this test suite.

client/src/modules/multiple_payroll/modals/config.modal.js
```diff
@@ -0,0 +1,131 @@
...
+
+  vm.setCurrency = function setCurrency(currencyId) {
+    vm.payroll.currency_id = currencyId;
+    const sameCurrency = currencyId === vm.lastExchangeRate.currency_id;
+
+    const rateCurrency = sameCurrency ? vm.lastExchangeRate.rate : (1 / vm.lastExchangeRate.rate);
+
+    vm.employee.basic_salary *= rateCurrency;
+
+    if (vm.payroll.value.length) {
```
jniles: Is `vm.payroll.value` an array or an object?  If it is an array, we can just use `vm.payroll.value.forEach()` instead of using the `Object.keys()` call.

client/src/modules/multiple_payroll/modals/config.modal.js
```diff
@@ -0,0 +1,131 @@
...
+  vm.submit = submit;
+  vm.closeModal = closeModal;
+
+  vm.setCurrency = function setCurrency(currencyId) {
+    vm.payroll.currency_id = currencyId;
+    const sameCurrency = currencyId === vm.lastExchangeRate.currency_id;
+
+    const rateCurrency = sameCurrency ? vm.lastExchangeRate.rate : (1 / vm.lastExchangeRate.rate);
+
+    vm.employee.basic_salary *= rateCurrency;
```
jniles: To make it clear what is happening here, I propose that you move this code (and the code beneath it) to a separate function.  You can then call that function here.  That way, we can test the logic of setting the currency apart from the logic of changing the salary based on the exchange rate.

Something like this:
```js
vm.setCurrency = function setCurrency(currencyId) {
  vm.payroll.currency_id = currencyId;
   const isSameCurrency = currencyId === vm.lastExchangeRate.currency_id;
   const rate = isSameCurrency ? vm.lastExchangeRate.rate : (1 / vm.lastExchangeRate.rate);
  calculatePaymentWithExchangeRate(rate);
}

function calculatePaymentWithExchangeRate(rate) {
  vm.employee.basic_salary *= rate;
  vm.payroll.value = vm.payroll.value.map(value => value * rate);
}
```
}

client/src/modules/multiple_payroll/modals/config.modal.js
```diff
@@ -0,0 +1,131 @@
...
+      Object.keys(vm.payroll.value).forEach((key) => {
+        vm.payroll.value[key] *= rateCurrency;
+      });
+    }
+  };
+
+  Employees.read(vm.stateParams.uuid)
+    .then((employee) => {
+      vm.employee = employee;
+      vm.latestViewFilters = MultiplePayroll.filters.formatView();
```
jniles: None of the "latestViewFilters" code is related to the `Employees.read()`  function.  You should probably move it out of this callback.

client/src/modules/multiple_payroll/modals/config.modal.js
```diff
@@ -0,0 +1,131 @@
...
+
+      vm.label = vm.latestViewFilters.defaultFilters[0]._label;
+      vm.displayValue = vm.latestViewFilters.defaultFilters[0]._displayValue;
+      vm.idPeriod = vm.latestViewFilters.defaultFilters[0]._value;
```
jniles: Hmm...  we should figure out a different way to get this information.  Passing it through the filters display value isn't a good idea, since it ties this module's behavior to the the `FilterService` API.

Could you just pass the values you need into the controller when you create the modal?

client/src/modules/multiple_payroll/modals/config.modal.js
```diff
@@ -0,0 +1,131 @@
...
+      vm.currencyId = vm.latestViewFilters.defaultFilters[1]._value;
+      vm.payroll.currency_id = vm.latestViewFilters.defaultFilters[1]._value;
+      vm.employee.basic_salary = vm.employee.individual_salary ?
+        vm.employee.individual_salary : vm.employee.basic_salary;
+
+      return Exchange.read();
+    })
+    .then((exchangeRate) => {
+      vm.lastExchangeRate = exchangeRate.slice(-1)[0];
```
jniles: Why not `exchangeRate.pop()`?

client/src/modules/multiple_payroll/modals/config.modal.js
```diff
@@ -0,0 +1,131 @@
...
+      return Configuration.read(vm.idPeriod);
+    })
+    .then((period) => {
```
jniles: Since the `Configuration.read()` does not depend on the result of the `Exchange.read()` call, you could call these in parallel.  Like this:

```js
  .then((employee) => {
    // logic and code!
    // ... etc ...
 
    return [
      Exchange.read(),
      Configuration.read(vm.idPeriod),
    ];
  })
  .then(([exchangeRate, period]) => {
    vm.lastExchangeRate = exchangeRate.pop();
    const params = { /* etc.. */ };
    // ... etc ...
  })
```

It would speed the calculation up a bit.

client/src/modules/multiple_payroll/modals/config.modal.js
jniles: Also, will [Exchange.getCurrentRate(currencyId)](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/ExchangeRateService.js#L163) work?  What happens if there are multiple currencies?

client/src/modules/multiple_payroll/templates/cellStatus.tmpl.html
```diff
@@ -0,0 +1,7 @@
+<div class="ui-grid-cell-contents">  
+  <div ng-if="row.entity.status_id == 1" class="label label-default text-uppercase" translate> PAYROLL_STATUS.WAITING_FOR_CONFIGURATION </div>
```
jniles: Instead of having multiple `class="text-uppercase`, try moving this out onto the top level `<div>`
```html
<div class="ui-grid-cell-contents text-uppercase">
```
does that work?

client/src/modules/multiple_payroll/templates/cellStatus.tmpl.html
```diff
@@ -0,0 +1,7 @@
...
+  <div ng-if="row.entity.status_id == 2" class="label label-info text-uppercase" translate> PAYROLL_STATUS.CONFIGURED </div>
```
jniles: these would probably be better as `<span>` elements instead of `<div>`s.

client/src/modules/payroll/payroll.service.js
```diff
@@ -13,5 +13,13 @@ PayrollConfigurationService.$inject = ['PrototypeApiService'];
 function PayrollConfigurationService(Api) {
   var service = new Api('/payroll_config/');
 
+  service.paiementStatus = paiementStatus;
+
+  function paiementStatus() {
+    var url = ''.concat('paiementStatus');
```
jniles: This should be `const`.

server/models/schema.sql
```diff
@@ -755,15 +756,15 @@ DROP TABLE IF EXISTS `holiday`;
 
 CREATE TABLE `holiday` (
   `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
-  `employee_id` int(11) unsigned NOT NULL,
+  `employee_uuid` binary(16) NOT NULL,
```
jniles, line 762: 💃

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -0,0 +1,230 @@
...
+        });
+      });
+
+      return transactionOffDay.execute();
+    })
+    .then(results => {
+      employeeData.forEach(emp => {
+        emp.employee.offDaysPaid = [];
+        results.forEach(offDayEmployee => {
+          offDayEmployee.forEach(offDay => {
```
jniles, line 192: To eliminate one level of depth here, you could also do this:
```js
.then(results => {                                                                                                                                                                                                 
  employeeData.forEach(emp => {                                                                                                                                                                                    
    emp.employee.offDaysPaid = [];                                                                                                                                                                                 
    results.forEach(offDayEmployee => {                                                                                                                                                                            
                                                                                                                                                                                                                   
      const offdays = offDayEmployee                                                                                                                                                                               
        .filter(offDay => emp.employee.uuid === offDay.paiement_uuid);                                                                                                                                             
                                                                                                                                                                                                                   
      emp.employee.push(...offdays);                                                                                                                                                                              
    });                                                                                                                                                                                                            
  });                                                                                                                                                                                                              
}); 
```

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -0,0 +1,230 @@
...
+ *
+ * @description
+ * This method builds the Payslips for employees payed to be shipped back to
+ * the client.
+ *
+ * GET /reports/payroll/employees
+ */
+function build(req, res, next) {
+  const options = _.clone(req.query);
+  const typeVar = Array.isArray(options.employees);
```
jniles, line 32: Hmm.  I assume that you want an array of employees here?

@mbayopanda had a similar issue, where values would come as either a string (`'1'`) or an array (`['1', '2', '3']`).  We can solve this pretty easily y combining the values into an array with `concat()`.  Like this:
```js
const employees = [].concat(options.employees);

// now employees is an array, no matter what kind of values it was previously.
```

Will that work?

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -0,0 +1,230 @@
...
+
+  options.employees = !typeVar ? [JSON.parse(options.employees)] : options.employees;
+
+  // options.employees = [JSON.parse(options.employees)];
```
jniles: This can be removed now.

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -0,0 +1,230 @@
...
+
+  _.extend(options, {
+    filename : 'FORM.LABELS.PAYSLIP',
+    csvKey : 'payslipGenerator',
+    orientation : 'portrait',
```
jniles, line 43: Would page numbers be useful on this report?

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -0,0 +1,230 @@
...
+  const data = {};
+
+  data.enterprise = req.session.enterprise;
+  data.user = req.session.user;
+  data.lang = options.lang;
+
+  // set up the report with report manager
+  try {
+    report = new ReportManager(TEMPLATE, req.session, options);
+    delete options.orientation;
```
jniles: Why delete `options.orientation`?

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -0,0 +1,230 @@
...
+  } catch (e) {
+    next(e);
+    return;
+  }
+
+  options.employees.forEach(emp => {
+    const employee = typeVar ? JSON.parse(emp) : emp;
```
jniles: I think this conversion is unnecessary.  If you have already converted employees into an array above, we shouldn't need to do this .. unless I am missing something somewhere.

server/controllers/finance/creditors.js
```diff
@@ -28,7 +28,7 @@ function list(req, res, next) {
     FROM creditor AS c 
     JOIN creditor_group AS cg 
     JOIN account AS a ON c.group_uuid = cg.uuid AND cg.account_id = a.id 
-    JOIN entity_map map ON map.uuid = c.uuid;
+    LEFT JOIN entity_map map ON map.uuid = c.uuid;
```
jniles, line 31: 💃

server/controllers/finance/vouchers.js
```diff
@@ -260,6 +265,30 @@ function create(req, res, next) {
   // preprocess the items so they have uuids as required
   items.forEach((value) => {
     let item = value;
+    // Only for Employee Salary Paiement
+    if (voucherType === 7) {
```
jniles: Let's move this into it's own function.  It doesn't really belong in the middle of the `processItems` logic.

Something like this:

```js

items.forEach(value) => {
  // etc ...
});

if (voucherType === 7) {
  processPaymentInformation(items);
}
```

That way the logic is a bit more focused.

client/src/modules/vouchers/toolkit/payment_employee/payment_employee.modal.html
```diff
@@ -0,0 +1,58 @@
...
+    <bh-payroll-period-select
+      id="payroll_configuration_id"
+      payroll-configuration-id="ToolCtrl.searchQueries.payroll_configuration_id"
+      name="PayrollPeriodForm"
+      on-select-callback="ToolCtrl.onSelectPayrollPeriod(payrollConfig)"
+      required="true"
+      validation-trigger="ToolForm.$submitted">
+    </bh-payroll-period-select>
+ 
+    <!-- select invoices  -->
```
jniles: Is this really for selecting invoices?

client/src/modules/vouchers/voucher-registry.ctrl.js
```diff
@@ -35,6 +35,7 @@ function VoucherController(
   vm.clearGridState = clearGridState;
   vm.download = Vouchers.download;
   vm.deleteVoucher = deleteVoucherWithConfirmation;
+  vm.showReceipt = showReceipt;
```
jniles, line 38: 👍 Good bug fix!

package.json
```diff
@@ -76,6 +76,7 @@
     "multer": "^1.1.0",
     "mysql": "^2.14.0",
     "mz": "^2.7.0",
+    "protractor": "^5.3.1",
```
jniles: It looks like this moved from `devDependencies` to `dependencies`.  It should only be in `devDependencies` since we don't need it in production.  Could you move it back?

server/controllers/finance/vouchers.js
```diff
@@ -260,6 +265,30 @@ function create(req, res, next) {
...
+      paiementRows.forEach((paiement) => {
+        if (paiement.entity && item.entity) {
+          if (paiement.entity.uuid === item.entity.uuid) {
+            item.document_uuid = paiement.document_uuid;
+
+            const statusID = item.debit === paiement.debit ? 5 : 4;
+            const updatePaiement = `UPDATE paiement SET 
+              amount_paid = amount_paid + '${item.debit}',
```
jniles: This `item.debit` should be passed in using SQL parameter substitution (`?`).

client/src/modules/vouchers/complex-voucher.ctrl.js
```diff
@@ -262,6 +268,19 @@ function ComplexJournalVoucherController(
...
+    /**
+    * FIXE ME
+    * With the use of Journal Vouchers for the payment of the salaries
+    * of the agents, it was question of knowing the identifier
+    * of the payment to facilitate the update of the payment table finally to determine
+    * if the payment is partial or total, and in case of partial payment
+    * it would be impossible to determine the indetifier of the payment,
+    * reason for which we had used the property document_uuid to preserve the identifier of the table
+    */
+    if (voucher.type_id === 7) {
```
jniles: This code will be easier to modify in the future if you name this number:
```js
if (voucher.type_id === bhConstants.transactionType.SALARY_PAYMENT) {
 // etc..
}
```

Also great comment!

client/src/modules/purchases/list/list.js
```diff
@@ -15,11 +15,8 @@ function PurchaseListController(
   $state, PurchaseOrder, Notify, uiGridConstants, Columns, GridState, Session,
   Modal
 ) {
-  var vm = this;
-
-  var cacheKey = 'PurchaseRegistry';
-  var state;
-  var columnDefs;
+  const vm = this;
```
jniles, line 18: Nice refactoring 👍

client/src/modules/payroll/rubrics/modals/rubric.modal.html
```diff
@@ -33,7 +33,6 @@
       id="debtor_account_id"
       account-id="RubricModalCtrl.rubric.debtor_account_id"
       label="FORM.LABELS.DEBTOR_ACCOUNT"
-      classe="4"
```
jniles, line 36: 💃

jniles (commented):
@lomamech, this is a lot of code.

I'll leave this initial review, and will return for another review tomorrow.  Let me know if you have any questions!

client/src/modules/multiple_payroll/modals/config.modal.js
lomamech: @jniles  The use of the getCurrentRate function (currencyId) or the getCurrentExchange service ExchangeRateService generates errors and after several analysis and test the conclusion is that the Bhima system will not be able to support more than two currencies I tried to add a third currency and the ExchangeRate service did not work well

server/controllers/payroll/reports/payslipGenerator.js
lomamech: I do not think so, just because there is not the page number on the payroll sheet of an employee

server/controllers/payroll/reports/payslipGenerator.js
lomamech: If the options.employees  is an Array (When selecting multiple employees)  
    each element of the array must be converted to Json because they are string type

server/controllers/payroll/reports/payslipGenerator.js
lomamech: This is a good idea but will not solve the problem because by using a filter we will add in the array 'emp.employee.offDaysPaid' the filtered offDayEmployee while we only need to progressively add the elements offDayEmployee which respects the condition

lomamech:
@jniles  need a second review

jniles:
@lomamech gladly.  I'll begin working on it!

client/src/modules/multiple_payroll/modals/config.modal.html
```diff
@@ -0,0 +1,78 @@
...
+      <li class="title" translate>TREE.MULTI_PAYROLL</li>
+      <label class="badge badge-success">{{ ConfigPaiementModalCtrl.employee.display_name }}</label>
+      <label class="badge badge-primary"><span translate> {{ ConfigPaiementModalCtrl.label }} </span> : {{ ConfigPaiementModalCtrl.displayValue }}</label>
+    </ol>
+  </div>
+
+  <div class="modal-body" style="max-height: 70vh; overflow:auto;">
+    <div class="form-group">
+      <label class="control-label" translate>FORM.LABELS.GRADE</label>
+      <input disabled name="code_grade" ng-model="ConfigPaiementModalCtrl.employee.code_grade" autocomplete="off" class="form-control">
```
jniles: In general, `disabled` inputs are a bit difficult to read since they are greyed out.  It might be better to use either a `<p class="form-control-static">` tag or use `<input readonly ...>` in the future.

client/src/modules/multiple_payroll/modals/config.modal.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  // exposed methods
+  vm.submit = submit;
+  vm.closeModal = closeModal;
+
+  vm.latestViewFilters = MultiplePayroll.filters.formatView();
+
+  //FIXE ME: 
+  //Dont use index but use the property to found label, display value and value for each filter (@lomamech)
```
jniles, line 36: Nice comment 👍

client/src/modules/multiple_payroll/multiple_payroll.js
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * Multiple Payroll Registry Controller
+ *
+ * @description
+ * This controller is responsible for display all vouchers in the voucher table as a
+ * registry.  The registry supports client-side filtering, server-side searching, column
+ * reordering, and many more features.
+ */
+function MultiplePayrollController(
+  MultiplePayroll, Notify, uiGridGroupingConstants, uiGridConstants,
```
jniles: It doesn't look like you use `uiGridGroupingConstants`.  This could be removed.

client/src/modules/multiple_payroll/multiple_payroll.html
```diff
@@ -0,0 +1,111 @@
...
+    <div
+      id="multipayroll-grid"
+      class="grid-full-height-with-filters"
+      ui-grid="MultiplePayrollCtrl.gridOptions"
+      ui-grid-auto-resize
+      ui-grid-resize-columns
+      ui-grid-save-state
+      ui-grid-selection
+      ui-grid-move-columns
+      ui-grid-grouping>
```
jniles: Do you need `ui-grid-grouping`?

client/src/modules/multiple_payroll/multiple_payroll.js
```diff
@@ -0,0 +1,277 @@
...
+  }, {
+    field : 'net_salary',
+    displayName : 'FORM.LABELS.NET_SALARY',
+    headerCellFilter : 'translate',
+    cellFilter : 'currency:row.entity.currency_id',
+  }, {
+    field : 'balance',
+    displayName : 'FORM.LABELS.BALANCE',
+    headerCellFilter : 'translate',
+    cellFilter : 'currency:row.entity.currency_id',
```
jniles, line 56: Nice 👍

client/src/modules/payroll/rubrics/modals/rubric.modal.html
```diff
@@ -69,6 +67,20 @@
       </div>
     </div>
 
+    <div class="form-group">
+      <div class="checkbox">
+        <label>
+          <input
+            type="checkbox"
+            name="is_percent"
```
jniles: I think this `name` should be `is_defined_employee`.

client/src/modules/accounts/accounts.js
```diff
@@ -3,7 +3,7 @@ angular.module('bhima.controllers')
 
 AccountsController.$inject = [
   '$rootScope', '$timeout', 'AccountGridService', 'NotifyService', 'bhConstants',
-  'LanguageService', 'uiGridConstants',
+  'LanguageService', 'uiGridConstants', '$translate',
```
jniles: It doesn't look like you need `$translate` here.

test/integration/vouchers.js
```diff
@@ -98,6 +103,60 @@ describe('(/vouchers) The vouchers HTTP endpoint', () => {
...
+      entity_uuid: '42d3756a-7770-4bb8-a899-7953cd859892',
+      entity: { 
+        label : 'TEST 2 PATIENT',
+        type  : 'C',
+        uuid  : '42d3756a-7770-4bb8-a899-7953cd859892' 
+      },
+    }],
+  };
+
+  const voucher2 = {
```
jniles: Do you ever use `voucher2`?  If not, it could be removed.

test/data.sql
```diff
@@ -799,24 +864,43 @@ INSERT INTO `taxe_ipr` (`id`, `label`, `description`, `currency_id`) VALUES
...
+(1, 'Février 2018', '2018-02-01', '2018-02-28', 1, 1, 1, 1);
+
+-- Paiement DATA 
+
+INSERT INTO `paiement` (`uuid`, `employee_uuid`, `payroll_configuration_id`, `currency_id`, `paiement_date`, `total_day`, `working_day`, `basic_salary`, `daily_salary`, `base_taxable`, `gross_salary`, `net_salary`, `amount_paid`, `status_id`)
+VALUES (HUID('2a3f17b0ae3242bb9333a760825fd257'), HUID('75e0969465f245a1a8a28b025003d793'), 1, 2, NULL, 20, 20, 500.0000, 25.0000, 550.0000, 730.0000, 614.0700, 0.0000, 2);
+
+-- rubric_paiement DATA
+INSERT INTO `rubric_paiement` (`id`, `paiement_uuid`, `rubric_payroll_id`, `value`, `posted`) VALUES 
+(1, HUID('2a3f17b0ae3242bb9333a760825fd257'), 5, 20, NULL),
```
jniles: In the future, it would be better to use a temporary variable to make this statement a bit more readable.  Like this:
```sql
SET @paymentUuid = HUID('2a3f17b0ae3242bb9333a760825fd257');

INSERT INTO `rubric_paiement` (`id`, `paiement_uuid`, `rubric_payroll_id`, `value`, `posted`) VALUES 
(1, @paymentUuid, 5, 20, NULL)
-- etc..
```

server/controllers/payroll/reports/payslipGenerator.handlebars
```diff
@@ -0,0 +1,219 @@
...
+            <td style="background-color: #efefef;" colspan="7"></td>
+          </tr>
+          <tr style="background-color: #ddd;" >
+            <td class="text-center" colspan="7"> <strong> {{translate 'TABLE.COLUMNS.CHARGE'}} </strong></td>
+          </tr>          
+          {{#each employee.rubricDiscount}}
+            <tr>
+              <td class="text-left"> {{ label }} </td>
+              <td class="text-left"> {{ ratePercentage }} </td>              
+              <td class="text-left"> --- </td>
```
jniles: Do you need `text-left` classes...?  Shouldn't it be default?

server/controllers/payroll/reports/payslipGenerator.handlebars
```diff
@@ -0,0 +1,219 @@
...
+          <tr>
+            <td colspan="4" valign="top" class="text-right"> </td>
+            <td colspan="3"><h4><strong> ({{translate 'TABLE.COLUMNS.GROSS_SALARY'}}) - ({{translate "TABLE.COLUMNS.CONTRIBUTION_DEDUCTIONS"}})</strong></h4></td>
+          </tr>
+          <tr>
+            <td colspan="4" valign="top" class="text-right">&nbsp;</td>
+            <td colspan="3"><h4><strong> {{currency employee.gross_salary employee.currency_id}} - {{currency employee.somChargeEmployee employee.currency_id}} </strong></h4></td>
+          </tr>
+          <tr>
+            <td colspan="4" valign="top" class="text-right"><h3>{{translate 'TABLE.COLUMNS.NET_SALARY'}}</h3></td>
```
jniles: It is better to use `style` tags to define styling.  Like this:
```html
<td style="vertical-align : 'top'" class="text-right">
```

server/controllers/payroll/reports/payslipGenerator.handlebars
```diff
@@ -0,0 +1,219 @@
+{{> head title="FORM.LABELS.PAYSLIP"}}
+<body>
+  <!-- body  -->
+  {{#each elementsPayslip}}
+
+    <div style="page-break-after: always">
```
jniles, line 6: 👍

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
+/**
+ * Multiple Payroll Controller
```
jniles, line 2: I think this is the longest file in the application.  It would be helpful to break it up into smaller pieces.

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+        paiement.net_salary, paiement.working_day, paiement.total_day, paiement.daily_salary, paiement.amount_paid, 
+        paiement.status_id, paiement_status.text AS status
+        FROM employee 
+        JOIN creditor ON creditor.uuid = employee.creditor_uuid  
+        JOIN creditor_group ON creditor_group.uuid = creditor.group_uuid 
+        JOIN patient ON patient.uuid = employee.patient_uuid 
+        JOIN grade ON employee.grade_uuid = grade.uuid
+        JOIN paiement ON paiement.employee_uuid = employee.uuid
+        JOIN paiement_status ON paiement_status.id = paiement.status_id
+        WHERE paiement.payroll_configuration_id = '${options.payroll_configuration_id}'
```
jniles: This should use `?` syntax to avoid SQL injection attacks.

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+  filters.fullText('code');
+
+  // Company currency filtering is optional only if you want to
+  // know the currency for which the employees have been configured for payment
+  if (options.filterCurrency) {
+    filters.equals('currency_id');
+  }
+
+  filters.custom('status_id', 'payroll.status_id IN (?)', [statusIds]);
+  filters.setOrder('ORDER BY payroll.display_name');
```
jniles: Good ordering 👍

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+  if (validHolidays.length) {
+    rows.push(validHolidays);
+  } else {
+    rows.push([]);
+  }
+
+  // Get Working Days
+  let workingDay = 0;
+
+  for (let day = periodFrom; day <= periodTo; day.setDate(day.getDate() + 1)) {
```
jniles: It looks like you have made two loops to loop through a date range.  I propose that you make a function to create a date range from a `startDate` and `endDate`.  Like this:
```js
function createDateRange(startDate, endDate) {
  // returns an array of dates between startDate and endDate
}
```

This will allow you to simplify the code by then using a `forEach()` loop to loop through the array.  Like this:
```js
const range = createDateRange(periodFrom, periodTo);
range.forEach(day => {
  // add loop logic
});
```

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+        weekEndDays.forEach(days => {
+          if (holidayIndice === days.indice) {
+            invalidHoliday = true;
+          }
+        });
+
+        // Check if in a holiday period there is a offDay
+        validOffDays.forEach(off => {
+
+          const offDayCheck = moment(off.date).format('YYYY-MM-DD');
```
jniles: This works... but moment has more powerful tools to compare dates!  If you want to check if two dates are the same day, try this:
```js
// invalidHoliday is true if "off.date" is the same day as "day"
invalidHoliday = moment(off.date).isSame(day, 'day');
```

The [documentation for moment's isSame() method](http://momentjs.com/docs/#/query/is-same/) has more information.

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+  const periodFrom = new Date(params.dateFrom);
+  const periodTo = new Date(params.dateTo);
+
+  const validOffDays = [];
+  const validHolidays = [];
+
+  offDays.forEach(offDay => {
+    let invalidOffDays = false;
+    const offdayIndice = new Date(offDay.date).getDay();
+    weekEndDays.forEach(days => {
```
jniles: This code:
```js
let invalidOffDays = false;
weekEndDays.forEach(days => {
  if (offdayIndice === days.indice) {
    invalidOffDays = true;
  }
});
```
is actually an [Array.prototype.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)!  This method returns true if a condition is matched for at least one of the elements in an array.

You can rewrite your code more simply as:
```js
const offdayIndice = new Date(offDay.date).getDay();
const invalidOffdays = weekEndDays.some(days => {
  if (offdayIndice === days.indice) {
    return true;
  }
});
```
or even more simply as:
```js
const invalidOffdays = weekEndDays.some(weekEndDay => offdayIndice === weekEndDay.indice);
```
One line of code is much easier to read than six ;)

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+function manageConfigurationData(rows, params) {
+  const offDays = rows[1];
+  const weekEndDays = rows[3];
+  const periodFrom = new Date(params.dateFrom);
+  const periodTo = new Date(params.dateTo);
+
+  const validOffDays = [];
+  const validHolidays = [];
+
+  offDays.forEach(offDay => {
```
jniles: It looks like this structure is:
```js
validOffDays =  []
offDays.forEach(offday => {
  // do some comparison
  if (someComparison) {
    validOffDays.push(offday);
  }
});
```
This patter is perfect for [Array.prototype.filter()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)!

The idea is to only include offdays that are valid.  If you use the code I wrote below using Array.prototype.some(), you could transform this code into a `filter()` like this:
```js
const validOffDays = offDays.filter(offDay => {
  const offdayIndice = new Date(offDay.date).getDay());
  const invalidOffdays = weekEndDays.some(weekEndDay => offdayIndice === weekEndDay.indice);
  return !invalidOffdays;
});

// now validOffdays is an array of only days that are not weekends.
```

jniles: You can make this code even simpler by reversing the condition.  Instead of checking that there are no invalid off days, you can check that every offday is valid.  Instead of using [Array.prototype.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) which returns true if one element in an array satisfies the condition, you could use [Array.prototype.every()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) which returns true if _every_ element in the array satisfies the condition.

The code looks like this:
```js
const validOffDays = offDays.filter(offDay => {
  const offdayIndice = new Date(offDay.date).getDay());
  const isValidOffDay = weekEndDay.every(weekEndDay => offdayIndice !== weekEndDay.indice);
  return isValidOffDay;
});
```

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+  const getRubricPayroll = `
+    SELECT config_rubric_item.id, config_rubric_item.config_rubric_id, config_rubric_item.rubric_payroll_id, 
+    payroll_configuration.label AS PayrollConfig, rubric_payroll.* 
+    FROM config_rubric_item
+    JOIN rubric_payroll ON rubric_payroll.id = config_rubric_item.rubric_payroll_id
+    JOIN payroll_configuration ON payroll_configuration.config_rubric_id = config_rubric_item.config_rubric_id
+    WHERE payroll_configuration.id = ?;
+  `;
+
+  transaction
```
jniles: If you are only doing `SELECT` sql queries, you do not need to wrap them in a transaction.  It would be better to put these in a `q.all()` or `Promise.all()` to execute them at the same time.  Like this:
```js
const queries = q.all([
  db.exec(getPeriodData, [payrollConfigurationId]),
  db.exec(getRubricPayroll, [payrollConfigurationId]),
]);

queries.then(rows => {
  // etc..
});
```

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+
+        let advantagesEmployee = [];
+
+        const option = {
+          dateFrom,
+          dateTo,
+          employeeUuid : employee.employee_uuid,
+        };
+
+        return Exchange.getExchangeRate(enterpriseId, employee.currency_id, new Date())
```
jniles: It doesn't look like these first three `.then()` methods depend on one another.  Let's execute them all at the same time:
```js
const queries = [
  Exchange.getExchangeRate(enterpriseId, employee.currency_id, new Date()),
  Exchange.getExchangeRate(enterpriseId, iprCurrencyId, new Date()),
  EmployeeData.lookupEmployeeAdvantages(employee.employee_uuid),
];

q.all(queries)
  .then(([exchange, exchangeIpr, advantages]) => {
    enterpriseExchangeRate = currencyId === parseInt(employee.currency_id, 10) ? 1 : exchange.rate;
    iprExchangeRate = exchangeIpr.rate;
    advantagesEmployee = advantages;
    return getConfigurationData(payrollConfigurationId, option);
  })
  .then(/*  etc ... */);
```

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+          .then((exchangeIpr) => {
+            iprExchangeRate = exchangeIpr.rate;
+            return EmployeeData.lookupEmployeeAdvantages(employee.employee_uuid);
+          })
+          .then((advantages) => {
+            advantagesEmployee = advantages;
+            return getConfigurationData(payrollConfigurationId, option);
+          })
+          .then((dataConfiguration) => {
+            const dataManaged = manageConfigurationData(dataConfiguration, option);
```
jniles: What is `dataManaged[0]`, `dataManaged[1]` and `dataManaged[3]`?  Why are they not used?

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+            const workingDayCost = dailySalary * (daysPeriod.working_day - (nbHolidays + nbOffDays));
+
+            const nbChildren = employee.nb_enfant;
+
+            /**
+            * Some institution allocates a percentage for the offday and holiday payment,
+            * the calculation of this rate is found by calculating the equivalence of the daily wage with
+            * the percentage of the offday or holiday.
+            */
+            if (offDays.length) {
```
jniles: I don't think you need the `if (offDays.length) {` check.  If it's an array, `offDays.forEach()` will not be called on an array of 0 length.

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+            // Annual cumulation of Base IPR
+            const annualCumulation = baseIpr * 12;
+
+            let ind = -1;
+            let iprValue = 0;
+            let scaleIndice;
+
+            if (iprScales.length) {
+              iprScales.forEach(scale => {
+                ind++;
```
jniles: If you are just looking for the index in the array, you can use the second parameter of the [Array.prototype.forEach()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) function.  It gives you access to the index in the second parameter.  Like this:
```js
let index;

iprScales.forEach((scale, idx) {
  index = idx;
});
```

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+    })
+    .catch(next)
+    .done();
+}
+
+/**
+ * POST /multiple_payroll/:id/configuration
+ */
+function setConfiguration(req, res, next) {
+  const data = req.body.data;
```
jniles: Alternatively
```js
const { data  } = req.body;
```

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+  const uid = db.bid(paiementUuid);
+
+  const allRubrics = [];
+  const holidaysElements = [];
+  const offDaysElements = [];
+
+  let enterpriseExchangeRate = 0;
+  let iprExchangeRate = 0;
+
+  Exchange.getExchangeRate(enterpriseId, data.currency_id, new Date())
```
jniles: These two can also be executed at the same time for a small speed up.
```js
q.all([
  Exchange.getExchangeRate(enterpriseId, data.currency_id, new Date()),
  Exchange.getExchangeRate(enterpriseId, iprCurrencyId, new Date()),
])
  .spread((exchange, exchangeIpr) => {
   // etc
  });

File: server/controllers/payroll/multiplePayroll/index.js [hunk 77]
```diff
@@ -0,0 +1,1177 @@
...
+
+      /**
+     * Some institution allocates a percentage for the offday and holiday payment,
+     * the calculation of this rate is found by calculating the equivalence of the daily wage with
+     * the percentage of the offday or holiday.
+    */
+
+      data.offDays.forEach(offDay => {
+        const offDayValue = ((dailySalary * offDay.percent_pay) / 100);
+        offDaysCost += offDayValue;
```
jniles: I'm pretty worried about doing lots of math in JavaScript, but I'm not sure of a better solution.  It would be nice to do as much of this in SQL as possible.  Maybe we can migrate it into a stored procedure eventually.
```js
console.log(1.3 + 0.4)
// => 1.7000000000000002
```

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+          let taxesContributions = [];
+
+          if (rubrics.length) {
+            rubrics.forEach(rubric => {
+            // Initialize values for rubrics that are not automatically calculated
+              rubric.result = util.roundDecimal(data.value[rubric.abbr], DECIMAL_PRECISION);
+            });
+
+            // Filtering non-taxable Rubrics
+            nonTaxables = rubrics.filter(item => item.is_social_care);
```
jniles: Nice names 👍

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+            const baseIpr = ((baseTaxable - membershipFeeEmployee) * (iprExchangeRate / enterpriseExchangeRate));
+
+            // Annual cumulation of Base IPR
+            const annualCumulation = baseIpr * 12;
+
+            let ind = -1;
+            let iprValue = 0;
+            let scaleIndice;
+
+            if (iprScales.length) {
```
jniles: This code looks identical to [line 796](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2703/files#diff-7562bcd0f3ac8c515205012f560eaebdR796).   Is there a way we could make little functions that could be called twice instead of writing the code twice?

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+            daily_salary : dailySalary,
+            total_day : totalDayPeriod,
+            working_day : data.working_day,
+            gross_salary : grossSalary,
+            net_salary : netSalary,
+            amount_paid : 0,
+            status_id : 2,
+          };
+
+          const deletePaiementData = 'DELETE FROM paiement WHERE employee_uuid = ? AND payroll_configuration_id = ?';
```
jniles: Why are we deleting payment data ..?

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+
+          if (holidaysElements.length) {
+            transaction.addQuery(setHolidayPaiement, [holidaysElements]);
+          }
+
+          if (offDaysElements.length) {
+            transaction.addQuery(setOffDayPaiement, [offDaysElements]);
+          }
+
+          transaction.execute()
```
jniles: You should `return transcation.execute()` here and move the `.then().catch()` code outside the promise chain.

server/controllers/payroll/multiplePayroll/index.js
```diff
@@ -0,0 +1,1177 @@
...
+          FROM paiement
+          JOIN rubric_paiement ON rubric_paiement.paiement_uuid = paiement.uuid
+          JOIN rubric_payroll ON rubric_payroll.id = rubric_paiement.rubric_payroll_id
+          WHERE paiement.uuid = ? AND rubric_paiement.value > 0
+          `;
+
+        transac
+          .addQuery(sqlGetRubricPayroll, [paiementUuid]);
+
+        return transac.execute()
```
jniles: There is no need to put SELECT queries in a transaction.

server/controllers/payroll/reports/multipayroll.handlebars
```diff
@@ -0,0 +1,53 @@
...
+            <tr class="text-left">
+              <td>{{ add @key 1}}</td>
+              <td>{{display_name}}</td>
+              <td>{{code}}</td>
+              <td>{{currency net_salary currency_id}}</td>
+              <td>{{currency balance currency_id}}</td>
+              <td><strong>{{translate status}} </strong></td>
+            </tr>
+          {{else}}
+            {{> emptyTable columns=3}}
```
jniles: This should read `columns=6`

server/controllers/payroll/reports/payslipGenerator.handlebars
```diff
@@ -0,0 +1,219 @@
...
+            <th style="width: 15%;" class="text-center"> {{translate "TABLE.COLUMNS.CONTRIBUTION_DEDUCTIONS"}}</th>
+            <th style="width: 15%;" class="text-center"> {{translate "TABLE.COLUMNS.EMPLOYER_CONTRIBUTIONS"}}</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr style="background-color: #ddd;" >
+            <td class="text-center" colspan="7"> <strong> {{translate 'TABLE.COLUMNS.BASIC_SALARY'}} </strong></td>
+          </tr>
+          <tr>
+            <td class="text-left">  {{translate 'TABLE.COLUMNS.DAY_WORKED'}} </td>
```
jniles: I don't think you need to add `text-left` on any of these.

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -0,0 +1,234 @@
...
+
+  getHolidays.forEach(item => {
+    transactionHoliday.addQuery(item.query, item.params);
+  });
+
+  getOffDays.forEach(item => {
+    transactionOffDay.addQuery(item.query, item.params);
+  });
+
+  transaction.execute()
```
jniles: If you are just using SELECTs a transaction will be a lot slower.  Go ahead and use an array of promises with `q.all()`.

server/controllers/payroll/reports/payslipGenerator.js
jniles: It won't produce the same result?  I thought it would. :(

jniles (commented):
@lomamech, this was a lot of code!  I've made some specific comments, but in general the code is quite readable.

If you only have SELECT statements, there is no need to put them in a `db.transaction()`.  It will make the database do a lot of extra work for nothing.  A transaction should only be used with multiple INSERT/UPDATE/DELETE statements that need happen together as a block.  A better way to structure the code is to use an array of promises and wrap it with `q.all()`.

Remember to use `?` to escape all parameters sent from the client!  It is important to ensure we don't suffer SQL injection attacks.

I've made a few comments about the array processing.  If you need clarification, feel free to ask!

Again, this was a huge amount of work and it is impressive how clear the code actually is.  Well done!

server/controllers/payroll/multiplePayroll/index.js
lomamech: I wanted to do it but the problem is that payment.payroll_configuration_id is in a subquery and we also use a FilterParser

server/controllers/payroll/multiplePayroll/index.js
lomamech: allows you to learn the data set up for payment, this function is used when setting up an employee's payment with the modal and when configuring for several employees at once, this double use is the cause of the non-use of certain vauleurs returned including dataManaged[0], dataManaged[1] and dataManaged[3] which are used against when we do the manual configuration

server/controllers/payroll/multiplePayroll/index.js
lomamech: The payment configuration can be done several times and in each configuration it replaces the old Employee Payment and Payroll configuration data.

lomamech:
@jniles  all test pass, awaiting review...

client/src/i18n/en/tree.json
```diff
@@ -50,6 +50,7 @@
     "IPR_TAX_CONFIGURATION": "IPR Tax Configuration",
     "JOURNAL_VOUCHER": "Journal Voucher",
     "LOCATION": "Location Management",
+    "MULTI_PAYROLL" : "Multiple Payroll",
```
jniles, line 53: Out of curiosity - is there a single payroll?  Why do we keep calling this multiple payroll?

client/src/js/components/bhPayrollPeriodSelect.js
```diff
@@ -0,0 +1,63 @@
...
+function PayrollPeriodSelectController(PayrollConfigs, $timeout, $scope, Notify) {
+  const $ctrl = this;
+
+  // fired at the beginning of the Payroll Configuration select
+  $ctrl.$onInit = function $onInit() {
+    // translated label for the form input
+    $ctrl.label = $ctrl.label || 'FORM.LABELS.PERIOD_PAYMENT';
+
+    // fired when an Payroll Period configuration has been selected
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback;
```
jniles, line 32: Just FYI - this doesn't do anything.  I've seen it in a couple pull requests, but it is basically like writing:
```js
var x = 2;
x = x;  // not necessary
```

client/src/js/components/bhPayrollPeriodSelect.js
```diff
@@ -0,0 +1,63 @@
...
+
+    // default for form name
+    $ctrl.name = $ctrl.name || 'PayrollPeriodForm';
+
+    if (!angular.isDefined($ctrl.required)) {
+      $ctrl.required = true;
+    }
+
+    PayrollConfigs.read()
```
jniles, line 41: The `read()` code should be in the `$onInit` method - it will make it easier to test this component in isolation.

client/src/modules/multiple_payroll/modals/search.modal.js
```diff
@@ -0,0 +1,126 @@
...
+  const lastValues = {};
+
+  let statusText = '/';
+
+  // displayValues will be an id:displayValue pair
+  const displayValues = {};
+  const lastDisplayValues = MultiplePayroll.filters.formatView().defaultFilters;
+
+
+  lastDisplayValues.forEach((last) => {
```
jniles, line 35: Just FYI - I think this is already implemented in [FilterService.prototype.getDisplayValueMap()](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/FilterService.js#L200).

client/src/modules/vouchers/toolkit/payment_employee/payment_employee.modal.html
```diff
@@ -0,0 +1,58 @@
...
+        style="height: 40vh; margin-bottom: 20px;"
+        ui-grid="ToolCtrl.gridOptions"
+        ui-grid-filtering
+        ui-grid-auto-resize
+        ui-grid-selection>
+      </div>
+    </div>
+
+    <div ng-show="ToolCtrl.gridDisplay" class="text-right">
+      <span translate>FORM.LABELS.TOTAL</span>: <strong class="text-danger">{{ ToolCtrl.totalNetSalary || 0 | currency: ToolCtrl.currencyId }}</strong><br>
```
jniles, line 39: In the future, these would be easier to read with parentheses:
```html
<strong class="text-danger">{{ (ToolCtrl.totalNetSalary || 0) | currency: ToolCtrl.currencyId }}</strong/
```
It just makes it easier to parse.

server/controllers/finance/vouchers.js
```diff
@@ -260,6 +263,23 @@ function create(req, res, next) {
   // preprocess the items so they have uuids as required
   items.forEach(value => {
     let item = value;
+    // Only for Employee Salary Paiement
+    if (voucherType === 7) {
+      if (item.document_uuid) {
+        const updatePaiement = `
```
jniles, line 269: This SQL statement does not have anything to do with the loop - it could easily be moved out, like this:
```js
const updateSQL = ` some sql`

items.forEach(value => {
  updatesPaiementData.push({query : updateSQL, params : [ /* ... */ ] });
});
```

jniles (approved):
@lomamech thanks for sticking with it this long.

I can't guarantee that everything is perfect, but we should just get this into master to start working with it and improving it iteratively.

bors r+

Issue #2695, opened 2018-04-17
Proposal: make Prepayments/Caution payments an enterprise setting


jniles:
We added "enterprise settings" in https://github.com/IMA-WorldHealth/bhima-2.X/pull/2607 to toggle features on and off behaviors and features in the BHIMA application as a whole.  So far, we've only used this for allowing price changes in the Patient Invoice module.

I propose that we add an additional setting to disable caution payments.  One of our install points (HEV) does not use cautions at all.  We could save some logic, rendering, and avoid mistakes by disabling them completely if they are not used.

The basic idea would be to put a flag "enable_prepayments" in the `enterprise_setting` table.  This flag would control the following things:
 1. The "caution" option would not be available on the Cash page.  The page would default to "invoice payment".
 2. The alerts for creditor balances on the Patient Invoice page would never show, even if there was a creditor balance.
 3. The server would not check a creditor balance and skip logic described in #2687.

Issue #2687, opened 2018-04-16
Design: Caution "Links" in 2.x


jniles:
This issue describes as design for balancing caution payments in 2.x

Previous issues on cautions: https://github.com/IMA-WorldHealth/bhima-2.X/issues/808, https://github.com/IMA-WorldHealth/bhima-2.X/issues/1324.

A caution payment is any payment that happens _before_ goods or services are rendered.  These are often required as a down-payment or collateral to ensure the hospital can recuperate some of its investment before engaging in an expensive operation.

**The Problem of Cautions in 2.x**
Cautions are tricky in 2.x because of how we track balances in invoices and patients.  A patient's balance is computed by looking at the `entity_uuid` column.  This does not take into account what kinds of transactions hit the debtor's account - it only knows the debtor account was it.

However, 2.x also tracks the amount due on an invoice by patient.  This is different from the patient balance owed to the enterprise, since it groups by `reference_uuid`.  All "unbalanced" invoices are shown for payment at the Cash Payment windows, even if the debtor's overall balance is different from the sum of their invoices if they paid a caution.

**Linking Cautions**
The problem is clear - cautions are never linked to the invoices, making them unbalanced, from BHIMA's perspective, regardless of the debtor's situation.

I propose that we implement a caution "link" that associates one (or more) caution(s) with an invoice at the time of invoice creation.  This "caution link" will be a voucher created _after_ the invoice is created that associates the patient's invoice with their previous payment.  It will look like this:

| **Record** | **Account** | **Debit** | **Credit** | **Entity** | **Description** | **Reference** |
| --------- | ----- | -----:| ----:| ---- | ---- | --- |
| **CP.TPA.1** | 41XXXX | 0 | $50.00 | PA.TPA.2 | A caution payment of $50.00 for PA.TPA.2 | |
| **CP.TPA.1** | 57XXXX | $50.00 | 0 |   | A caution payment of $50.00 for PA.TPA.2 | |
| **IV.TPA.1** | 41XXXX | $15.00 | 0 | PA.TPA.2 | An invoice to PA.TPA.2 for quinine and penicillin | |
| **IV.TPA.1** | 75XXXX | 0 | $7.00 |   | An invoice to PA.TPA.2 for quinine and penicillin | |
| **IV.TPA.1** | 75XXXX | 0 | $8.00 |   | An invoice to PA.TPA.2 for quinine and penicillin | |
[...]

Issue #1324, opened 2017-03-09
(feature request) Caution refactor and tests


sfount:
There are a number of issues tracking the current implementation of cautions in 2.x. This must be corrected before releasing a solid financial base - many institutions rely on maintaining a line of credit with patients. 

Relevant discussions: 
https://github.com/IMA-WorldHealth/bhima-2.X/issues/808
https://github.com/Vanga-Hospital/bhima-2.X/issues/121
https://github.com/Vanga-Hospital/bhima-2.X/issues/34

jniles:
I'll be tackling all caution-related work.

jniles:
To summarize other issues here:

--------
Vanga-Hospital#121
Cautions have been a concept in 2.x for some time however there has only ever been facility to pay into a patients caution, crediting the patient and debiting a cash box. 

This proposal aims to: 
* Allow cautions to pay towards invoices 
* Reduce the number of complex corner cases to do with using cautions 
* Ensure users are very clearly notified about the state of a patients cautions 

The proposal is that the system implements the following rules: 
* **You cannot pay into a caution if you have remaining debt to the hospital** (This should be blocked with a clear modal explanation)
* **Any invoice made towards a patient with a caution remaining is directly linked and paid for by the caution** (This should be made very clear both on the invoice receipt as well as the user creating the invoice)

By implementing these two clear rules we guarantee that unless the accountant creates a specific voucher the system will not pay into cautions if the client has debt. This greatly reduces corner cases.

Required Infrastructure: 
- [ ] (MySQL) Patient invoice posting process uses cautions during the posting process (potentially as a separate voucher transaction)
- [ ] (Client UX) Invoices paid with a caution automatically very clearly mark that they have been fully or partially paid on creation
- [ ] (Client UX) Cash window blocks user from paying into a caution if the patient has debts with the hospital
- [ ] (Client UX) Reports/ views into caution payments, clear description in financial activity 

---

Required decision: 
Currently we can make this assumption because we do not allow the cash payment page to overpay an invoice. However this may restrict the user from recording what actually happened at cash payment. 

The current system: 
* Patient cannot pay more than the minimum currency unit above an invoice 

A proposal for how to deal with this is to automatically pay against all remaining debts given an over payment, after all debts have been paid any additional payment would go into the patients caution. 

The steps would be as follows: 
* Patient pays $20 to a $10 invoice, patient has another unrelated invoice for $5
* System puts $10 to the bill
[...]

PR #2704, opened 2018-04-19
Allow caution payments for invoices

### LINKED ISSUE #808: (feat) Handling automatically the CAUTION when invoicing a patient
This issue is about to handle CAUTION of a patient automatically when invoicing the patient.

We must have a mean to signal to the user of BHIMA who makes invoice that the patient has a CAUTION and if we have to use this CAUTION for an invoice.

### Caution Steps: 

1. Client pay a caution via the `cash payment module`

    - The client will be `credited`
    - The cashbox will be `debited`
    - A caution receipt is generated as proof 

2. Bill the patient : 

    - Display the remaining caution amount, in reality the balance of the client which is `< 0`
    - In this step, the caution displayed is for `just notify` the user that the enterprise must money to the client
    - The caution amount doesn't affect the invoice, invoices are independent to cautions

3. Patient Financial Activities : 

    - Notify also for a remaining caution amount of the client

jniles:
This PR presents the initial step in supporting cautions/prepayments in BHIMA.  The following components have been implemented:

 1. A reliable calculation of the patient's balance with the enterprise.  If a patient has a creditor balance with the hospital, the Patient Invoice module will signal the user that the patient has a creditor balance with an alert.
 2. Automatic "linking" of the patient's previous cash prepayment/caution when they are invoiced.  This is done by creating a third transaction that both debits and credits the same accountant.  Interestingly, this rule is not caught by the Trial Balance, which it probably should.

This PR is now complete!  It includes the option of toggling on/off prepayments, which will be particularly useful to institutions that do not accept prepayments.  The following issues have been closed:

 1. Closes #808 
 2. Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/1324
 3. Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/2687
 4. Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/2695


jniles:
@mbayopanda, can you give this a quick review?  I'd like to break this into smaller components, so that you don't have a ton to review at the end.

server/models/procedures/invoicing.sql
```diff
@@ -17,8 +16,7 @@ CREATE PROCEDURE StageInvoice(
   IN date DATETIME,
   IN cost DECIMAL(19, 4) UNSIGNED,
   IN description TEXT,
-  IN service_id SMALLINT(5) UNSIGNED,
-  IN debtor_uuid BINARY(16),
+  IN service_id SMALLINT(5) UNSIGNED, IN debtor_uuid BINARY(16),
```
mbayopanda: Why this in one line instead of multiple lines.

server/models/procedures/invoicing.sql
```diff
@@ -580,5 +578,196 @@ BEGIN
...
+  DECLARE enterpriseCurrencyId SMALLINT(5);
+  DECLARE linkTransactionTypeId SMALLINT(5);
+  DECLARE debtorAccountId INT(10);
+
+
+  DECLARE curse CURSOR FOR
+    SELECT payment.uuid, payment.balance FROM stage_payment_balances AS payment;
+  DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;
+
+  CALL SetupDebugLog();
```
mbayopanda: Is this call necessary ?

sh/build-database.sh
```diff
@@ -29,8 +29,9 @@ echo "[build] functions"
 mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/functions.sql &> /dev/null
 
 echo "[build] procedures"
-mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/procedures.sql &> /dev/null
+mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/procedures.sql
 mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/admin.sql &> /dev/null
+mysql -u $DB_USER -p$DB_PASS $DB_NAME < server/models/debug.sql &> /dev/null
```
mbayopanda: `debug.sql` is now a part of development environment.

server/models/procedures/invoicing.sql
```diff
@@ -580,5 +578,196 @@ BEGIN
...
+
+  -- TODO(@jniles) - decide on this for real
+  SET linkTransactionTypeId = 3;
```
mbayopanda: `TransactionTypeId = 3` is for convention payment, so maybe we can add a new transaction type for Caution

caution link transaction type

mbayopanda (commented):
I made a comment on `transactionType=3` apart from that the caution payment and its link to invoice work well.

mbayopanda:
@jniles ping me when it will be possible to `bors` this PR

server/models/procedures/invoicing.sql
jniles: Hehehe nope.  Good catch.

jniles:
@mbayopanda, this is ready for another review.

client/src/modules/cash/cash-form.service.js
```diff
@@ -29,10 +29,16 @@ function CashFormService(AppCache, Session, Patients, Exchange) {
   CashForm.prototype.setup = function setup() {
     this.details = { date : new Date(), description : '', invoices : [] };
     this.details.currency_id = Session.enterprise.currency_id;
+    this.hasPrepaymentSupport = Session.enterprise.settings.enable_prepayments;
```
mbayopanda, line 32: 👍

mbayopanda (approved):
LGTM

Here are some things that i noticed : 
- In the enterprise module we can have a button for updating enterprise information in the section "Parametre de l'entreprise" for an easy use

- Translate are missing in the message of caution in the invoicing page.

jniles:
@mbayopanda, thanks for the review!

I'll put in a patch with the translations soon.

PR #2710, opened 2018-04-23
feat(enterprise exchage rate) display currency symbol

### LINKED ISSUE #2707: fix(enterprise exchange rate): no currency symbol
There is no currency symbol for a fresh instance of the application (after running `./sh/build-init-database`).

We create an enterprise with $ as currency with the default project.

no currency symbol

The currency symbol is not displayed correctly

Also the feature `Exchange Rate` doesn't work if we want to set an exchange rate.

jeremielodi:
closes #2707

### DIFF
--- a/client/src/modules/enterprises/exchange/exchange.modal.js
+++ b/client/src/modules/enterprises/exchange/exchange.modal.js
@@ -41,7 +41,7 @@ function ExchangeRateModalController(ModalInstance, Exchange, Currencies, Sessio
         .filter(currency => currency.id !== Session.enterprise.currency_id);
 
       // use the first currency in the list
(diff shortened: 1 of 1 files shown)


client/src/modules/enterprises/exchange/exchange.modal.js
```diff
@@ -41,7 +41,7 @@ function ExchangeRateModalController(ModalInstance, Exchange, Currencies, Sessio
         .filter(currency => currency.id !== Session.enterprise.currency_id);
 
       // use the first currency in the list
-      [vm.rate.currency] = vm.currencies0;
```
jniles, line 44: :dancer:

jniles (approved):
bors r+

PR #2713, opened 2018-04-24
feat(bhReportPeriodSelect) display current period by default

### LINKED ISSUE #2671: Aged Debtors and Creditors Reports should default to current period
The Aged Debtors and Aged Creditors report defaults to an empty form.  It would be useful if these defaulted to the current period.  It might also be helpful to group by the fiscal year in the `ui-select` to provide a bit more context.

_Argh... hang on, I got this.  No wait.  Just gimme today_

jeremielodi:
closes #2671

### DIFF
--- a/client/src/js/components/bhReportPeriodSelect.js
+++ b/client/src/js/components/bhReportPeriodSelect.js
@@ -28,7 +28,7 @@ function ReportPeriodSelectController(FiscalPeriod, $filter) {
   $ctrl.$onInit = function onInit() {
     // fired when a Period has been selected
     $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
-
+    $ctrl.displayCurrentPeriod = !($ctrl.periodId);
     // default for form name
     $ctrl.name = $ctrl.name || 'PeriodForm';
 
@@ -43,6 +43,10 @@ function ReportPeriodSelectController(FiscalPeriod, $filter) {
         });
 
         $ctrl.periodes = periodes;
+        // console.log($ctrl.periodes);
+        if ($ctrl.displayCurrentPeriod) {
+          setCurrentperiod();
+        }
       });
 
     $ctrl.valid = true;
@@ -50,4 +54,10 @@ function ReportPeriodSelectController(FiscalPeriod, $filter) {
 
   // fires the onSelectCallback bound to the component boundary
   $ctrl.onSelect = period => $ctrl.onSelectCallback({ period });
+
+  function setCurrentperiod() {
+    FiscalPeriod.currentPeriod().then(id => {
+      $ctrl.periodId = id;
+    });
+  }
 }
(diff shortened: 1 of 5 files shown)


client/src/js/components/bhReportPeriodSelect.js
```diff
@@ -5,6 +5,7 @@ angular.module('bhima.components')
     // transclude  : true,
     bindings    : {
       periodId          : '<',
+      displayCurrentPeriod : '<?',
```
jniles: This doesn't need to be an option passed in.  Just check if `periodId` is defined, and if not, default to the current period.

server/controllers/finance/fiscalPeriod.js
```diff
@@ -103,3 +105,12 @@ function getPeriodDiff(periodIdA, periodIdB) {
   return db.one(sql, [periodIdA, periodIdB]);
 }
 
+function currentPeriod(req, res, next) {
+  const sql = `SELECT id FROM period WHERE start_date <=? AND end_date >=?`;
+  const currentDate = util.dateFormatter(new Date());
```
jniles: It's probably easier to let SQL format these.  Like this:
```js
const sql = `
  SELECT id FROM period WHERE DATE(start_date) <= DATE(?) AND DATE(end_date) >= DATE(?);
`;

db.one(sql, [new Date(), new Date()])
  .then(period => res.status(200).json(period.id));
  .catch(next)
  .done();
```

[According to the documentation for mysqljs](https://github.com/mysqljs/mysql#escaping-query-values)
> Date objects are converted to `'YYYY-mm-dd HH:ii:ss'` strings

This is a valid [DATETIME representation](https://dev.mysql.com/doc/refman/5.7/en/datetime.html) in MySQL.  However, since we don't care about the hour/minute/second parts, passing them to the SQL DATE() function will remove those parts, leaving just the DATE part for comparison.

server/lib/util.js
```diff
@@ -92,16 +92,19 @@ function requireModuleIfExists(moduleName) {
...
+        }
+      });
     });
-  });
 
-  return rows;
+    return rows;
+  }
+
+  return moment(rows).format(DATE_FORMAT);
```
jniles, line 111: Hmm... instead of having one function with different behaviors, it would be better to write two functions and have the programmer choose which one to use - one for arrays and one for single values.

Also, remember unit tests :)

client/src/js/services/FiscalPeriodService.js
```diff
@@ -14,6 +14,20 @@ FiscalPeriodService.$inject = [ 'PrototypeApiService' ];
...
+  const service = new Api('/periods');
+  service.currentPeriod = currentPeriod;
+
+  /**
+   *
+   * @method currentPeriod
+   * @description
+   * return the current period'id
+   */
+  function currentPeriod(params) {
```
jniles: Do you ever use `params`?  If not, go ahead and remove it.

jniles:
@jeremielodi, in thinking about it... we probably want to send the date from the client.  That way, if two people are in two different timezones, they will always see the correct period for their particular timezone.

PR #2714, opened 2018-04-25
feat(inventory) add sellable property

### LINKED ISSUE #2642: Add inventory property "sellable"
As we begin to mix stock management for invoicing and stock management for fixed assets, we will need some way to hide items from the invoicing form.  For example, a computer should not be available for sale through the Patient Invoice window, but might be tracked in stock in BHIMA.

**Proposal**
I propose that we put a "sellable" property on the `inventory` table that will be read by the Patient Invoice form.  When a user registers an item, they will have the option to choose "For Sale to Patients" or not.

The original proposal (https://github.com/IMA-WorldHealth/bhima-2.X/issues/1064) proposed this flag to be attached to the `inventory_group` table, however I now see this as incorrect.  Suppose the hospital had reserved medication for surgery - would it need to create a separate group for that just to ensure no one accidentally billed it?  Instead, it makes sense to me to limit the ability to sell a product to the product level.

Supersedes https://github.com/IMA-WorldHealth/bhima-2.X/issues/1064.

jeremielodi:
closes #2642


client/src/modules/inventory/list/templates/sellable.cell.tmpl.html
```diff
@@ -0,0 +1,4 @@
+<div class="ui-grid-cell-contents text-center">
+  <i ng-show="row.entity.sellable === 1" class="fa fa-check-circle-o"></i>
```
jniles: Alternatively:
```html
<i class="fa" ng-class="{ 'fa-check-circle-o' : row.entity.sellable, 'fa-times-circle-o' : row.entity.sellable }"></i>
```

client/src/i18n/en/inventory.json
```diff
@@ -37,6 +37,8 @@
         "INVENTORY_TYPE_INFO" : "inventory_type : the type of inventory such as Article, Service, Assembly or Other",
         "INVENTORY_UNIT_INFO" : "inventory_unit : the unit of inventory such as Piece, Tablet, Box, Bottle or Other",
         "INVENTORY_PRICE_INFO" : "inventory_price : the sale price of the inventory in the enterprise currency",
-        "COLUMNS_INFO" : "Note about some columns"
+        "COLUMNS_INFO" : "Note about some columns",
+        "SELLABLE" : "Sellable",
+        "UNSELLABLE" : "Unsellable"
```
jniles: It doesn't look like you use the key "unsellable".  You could remove it.

jniles (commented):
Code looks decent.

Could you modify the integration tests for inventory to check that the `sellable` property is able to be filtered on?   The PatientInvoice module will probably send a request like "GET /inventory/list?sellable=1" to figure out what items it should display.  We should make sure that "sellable=1" and "sellable=0" filter the inventory items correctly.

test/integration/inventory/metadata.js
```diff
@@ -80,6 +81,15 @@ describe('(/inventory/metadata) The inventory metadata http API', () => {
       .catch(helpers.handler);
   });
 
+  it('GET /inventory/metadata?sellable=1 returns the list of sellable inventories', () => {
+    return agent.get('/inventory/metadata?sellable=1')
+      .then(res => {
+        expect(res).to.have.status(200);
+        expect(res.body).to.not.be.empty;
```
jniles: Hmm... this test only checks that the route doesn't break.   Let's make sure the logic is correct.

How about you change one of the default inventory items to `sellable = 0` and then check that `agent.get('/inventory/metadata?sellable=1')` does not include that item, but `agent.get('/inventory/metadata')` does?

jniles:
@jeremielodi, the code the the sellable work looks fine to me.  It seems like you wrote a GridExportService test which got in here that shouldn't be here.  Could you remove it?

jniles (approved):
LGTM.  Excellent work.

bors r+

PR #2716, opened 2018-04-26
feat(inventory) remove is_broken property

### LINKED ISSUE #2679: Remove the is_broken column for the inventory item
We have an `is_broken` column on our inventory items (proposed: https://github.com/IMA-WorldHealth/bhima-2.X/issues/2379, implemented: https://github.com/IMA-WorldHealth/bhima-2.X/pull/2457).  However, these properties are actually tied to individual stock items, not to the inventory items.  So far, we haven't really been able to use this column.

I propose that we remove the `is_broken` column from the inventory table.

jeremielodi:
closes #2679


client/src/modules/inventory/list/list.js
```diff
@@ -96,13 +96,6 @@ function InventoryListController(
     type : 'number',
     visible : false,
   },
-  {
-    field : 'is_broken',
```
jniles, line 100: :dancer:

server/models/migrations/v0.6.0-v0.7.0/migrate.sql
```diff
@@ -55,8 +55,6 @@ CREATE TABLE `rubric_payroll` (
 -- For some reason, MariaDB is being a pain with this.  Skipping it for now.
 -- ALTER TABLE `offday` DROP CONSTRAINT `offday_1`;
 -- ALTER TABLE `offday` ADD UNIQUE `offday_1` (`date`);
-
-ALTER TABLE `inventory` ADD COLUMN `is_broken` TINYINT(1) NOT NULL DEFAULT 0;
```
jniles, line 59: While it doesn't hurt anything currently, the `migrations/*` scripts are for updating old databases and usually shouldn't be edited.  It's to help us upgrade old versions of BHIMA in production (if we are ever running more than 2)

jniles (approved):
LGTM.

bors r+

PR #2717, opened 2018-04-26
feat(report): adding stock exit report

mbayopanda:
This PR is about adding a stock exit report, this report concerns : 
- stock exit to patient
- stock exit to service
- stock exit to depot
- stock exit as loss

stock exit report

stock exit print


.env.development
```diff
@@ -10,7 +10,7 @@ DB_NAME='bhima_test'
 # session variables
 SESS_SECRET='XopEn BlowFISH'
 
-DEBUG=app
+DEBUG=app,errors
```
jniles: It comes and goes 👯

client/src/js/components/bhDateInterval.js
```diff
@@ -100,9 +100,7 @@ function bhDateInterval(moment, bhConstants) {
   }
 
   function startup() {
-    var option;
-
-    option = ['day', 'week', 'month', 'year'].indexOf(vm.mode);
+    const option = ['day', 'week', 'month', 'year'].indexOf(vm.mode);
```
jniles, line 103: 💃

client/src/modules/reports/generate/stock_exit/stock_exit.html
```diff
@@ -0,0 +1,90 @@
...
+
+        <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate autocomplete="off">
+
+          <!-- select depot -->
+          <bh-depot-select
+            depot-uuid="ReportConfigCtrl.depot.uuid"
+            on-select-callback="ReportConfigCtrl.onSelectDepot(depot)"
+            required="true"
+            validate-trigger="ConfigForm.$submitted">
+            <bh-clear on-clear="ReportConfigCtrl.clear('depot')"></bh-clear>
```
jniles: Do you need this `clear()` method?  To me, if you cannot submit this report without a depot then there is no need for supporting clearing it.

server/controllers/stock/reports/stock/exit/exitToDepot.js
```diff
@@ -0,0 +1,38 @@
+const db = require('../../../../../lib/db');
```
jniles, line 1: This is the most `..` I've ever seen in an import :P

server/controllers/stock/reports/stock/exit/exitToDepot.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @function fetch
+ * @description fetch stock exit for patient
+ */
+function fetch(depotUuid, dateFrom, dateTo, showDetails) {
+  const IS_EXIT = 1;
+  const EXIT_TO_DEPOT_ID = 8;
```
jniles: Both of these `const` could go outside the `fetch()` function, since they don't depend on the function itself.

server/controllers/stock/reports/stock/exit/exitToDepot.js
```diff
@@ -0,0 +1,38 @@
...
+    JOIN inventory_unit iu ON iu.id = i.unit_id 
+    JOIN depot d ON d.uuid = m.depot_uuid
+    JOIN depot dd ON dd.uuid = m.entity_uuid
+    JOIN user u ON u.id = m.user_id
+    JOIN document_map dm ON dm.uuid = m.document_uuid
+  WHERE m.is_exit = ${IS_EXIT} AND m.flux_id = ${EXIT_TO_DEPOT_ID} AND d.uuid = ? 
+    AND (DATE(m.date) BETWEEN DATE(?) AND DATE(?))
+  GROUP BY i.uuid`;
+
+  const details = ', m.uuid, dd.uuid';
```
jniles: Hmm... this may be picky, but it might be good to call this `grouping` or something like that.  I was expect this to be extra details in the `SELECT ... FROM` part of the query, but it looks like it adds to the `GROUP BY` portion.

client/src/modules/reports/generate/stock_exit/stock_exit.config.js
```diff
@@ -0,0 +1,83 @@
...
+      includeDepotExit : vm.includeDepotExit,
+      includeLossExit : vm.includeLossExit,
+      showDetails : vm.showDetails,
+    };
+
+    // update cached configuration
+    cache.reportDetails = angular.copy(params);
+
+    const options = {
+      params,
```
jniles: Why do you include the params as a JSON instead of just using HTTP's parameter serializing?  Is there a particular reason?

If not, I propose you do this:
```js
const options = angular.merge(params, { lang : Lanuages.key });
```
... which means that you can remove the `JSON.parse(req.query.params)` command from server because it will automatically be parsed by express's middleware.

client/src/modules/reports/generate/stock_exit/stock_exit.html
```diff
@@ -0,0 +1,90 @@
...
+          <!-- date interval -->
+          <bh-date-interval 
+            date-from="ReportConfigCtrl.dateFrom" 
+            date-to="ReportConfigCtrl.dateTo">
+          </bh-date-interval>
+
+          <!-- stock exit type -->
+          <div class="checkbox">
+            <label>
+              <input type="checkbox" ng-true-value="1" ng-false-value="0" value="1" ng-model="ReportConfigCtrl.includePatientExit">
```
jniles: Did you mean to have both `ng-*-value` and `value="1"`?  If you want to make this "checked" by default, I recommend you just set the model to `1` and let AngularJS do the work for you :)

server/controllers/stock/reports/stock/exit/exitToLoss.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * @function fetch
+ * @description fetch stock exit for patient
+ */
+function fetch(depotUuid, dateFrom, dateTo, showDetails) {
+  const IS_EXIT = 1;
+  const EXIT_TO_LOSS_ID = 11;
+  const sql = `
+  SELECT 
+    i.code, i.text, iu.text AS unit_text, BUID(m.document_uuid) AS document_uuid,
```
jniles: Hmm... Besides the `EXIT_TO_LOSS_ID` variable, what has changed between this `fetch()` function and the others?  Can they be combined?

client/src/modules/reports/generate/stock_exit/stock_exit.html
```diff
@@ -0,0 +1,90 @@
...
+          <bh-depot-select
+            depot-uuid="ReportConfigCtrl.depot.uuid"
+            on-select-callback="ReportConfigCtrl.onSelectDepot(depot)"
+            required="true"
+            validate-trigger="ConfigForm.$submitted">
+            <bh-clear on-clear="ReportConfigCtrl.clear('depot')"></bh-clear>
+          </bh-depot-select>
+
+          <!-- date interval -->
+          <bh-date-interval
```
jniles: Is there some way to make this required?  It would be nice - otherwise it kind of looks broken.

client/src/modules/reports/generate/stock_exit/stock_exit.config.js
```diff
@@ -0,0 +1,83 @@
...
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.preview = form => {
+    if (form.$invalid) { return 0; }
+
+    const params = {
+      depotUuid : vm.depot.uuid,
+      dateFrom : vm.dateFrom,
```
jniles: I propose that you set `vm.dateFrom = new Date();` as a default value early in the controller to prevent the following:


_Submission w/o date or checkboxes is interesting_

client/src/modules/reports/generate/stock_exit/stock_exit.config.js
```diff
@@ -0,0 +1,83 @@
...
+    delete vm[key];
+  };
+
+  vm.clearPreview = () => {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.preview = form => {
+    if (form.$invalid) { return 0; }
```
jniles: You might want to put custom logic here to make sure that at least one checkbox is checked.  Something like:
```js
const sum = vm.includePatientExit + vm.includeServiceExit + vm.includeDepotExit + vm.includeLossExit;
const hasOneChecked = sum > 0;
if (!hasOneChecked) {
  return 0;
}
```

You might also set a message that says at least one of the options should be checked.

server/controllers/stock/reports/stock/exit_report.js
```diff
@@ -0,0 +1,166 @@
...
+ * @function groupCollection
+ * @description group collected data by inventory
+ */
+function groupCollection(exitCollection) {
+  const collection = {};
+
+  // exit to patient
+  collection.exitToPatient = _.groupBy(exitCollection.exitToPatient, row => row.text);
+  collection.exitToPatient = _.map(collection.exitToPatient, formatExit);
+  collection.exitToPatientEmpty = (Object.keys(collection.exitToPatient).length === 0);
```
jniles: Alternatively, `_.size(collection.exitToPatient) === 0` using lodash's [_.size()](https://lodash.com/docs/4.17.5#size).

server/controllers/stock/reports/stock/exit_report.js
```diff
@@ -0,0 +1,166 @@
...
+  collection.exitToLoss = _.map(collection.exitToLoss, formatExit);
+  collection.exitToLossEmpty = (Object.keys(collection.exitToLoss).length === 0);
+
+  return collection;
+}
+
+/**
+ * @function formatExit
+ */
+function formatExit(value, index) {
```
jniles: Since this is called on an object, `key` is a better parameter name than `index`, which makes you think of an array.

server/controllers/stock/reports/stock/exit_report.js
```diff
@@ -0,0 +1,166 @@
...
+function groupCollection(exitCollection) {
+  const collection = {};
+
+  // exit to patient
+  collection.exitToPatient = _.groupBy(exitCollection.exitToPatient, row => row.text);
+  collection.exitToPatient = _.map(collection.exitToPatient, formatExit);
+  collection.exitToPatientEmpty = (Object.keys(collection.exitToPatient).length === 0);
+
+  // exit to service
+  collection.exitToService = _.groupBy(exitCollection.exitToService, row => row.text);
```
jniles: You could avoid repeating code by sacrificing your `collection` object structure.  For example, if you used nested objects, you could do this:
```js
function formatAndCombine(data) {
  return _.chain(data)
    .groupBy('text')
    .map(formatExit)
    .value();
}

collection.toServices = formatAndCombine(exitCollection.exitToService);
collection.toServicesEmpty = _.size(collection.toServices) === 0;
```

You could even reduce it more:
```js
function formatAndCombine(data) {
  const aggregate = _.chain(data)
    .groupBy('text')
    .map(formatExit)
    .value();

  return { data : aggregate, isEmpty : _.size(aggregate) === 0 };
}

// note - this is now an object with properties .data and .isEmpty
collection.toServices = formatAndCombine(exitCollection.exitToService);
collection.toPatients = formatAndCombine(exitCollection.exitToPatient);
```

Just a suggestion!

server/controllers/stock/reports/stock/exit_report.js
```diff
@@ -0,0 +1,166 @@
...
+    if (includeDepotExit) {
+      data.exitToDepot = await StockExitToDepot.fetch(depotUuid, dateFrom, dateTo, showDetails);
+    }
+
+    // get stock exit to loss
+    if (includeLossExit) {
+      data.exitToLoss = await StockExitToLoss.fetch(depotUuid, dateFrom, dateTo, showDetails);
+    }
+
+  } catch (error) {
```
jniles: Actually, there isn't a need for this `try/catch` block.  If any of the code throws an error, the whole function will throw an error and you can catch it in your top-level `try/catch` block.

server/controllers/stock/reports/stock/exit_report.js
```diff
@@ -0,0 +1,166 @@
...
+    inventory_unit : value && value[0] ? value[0].unit_text : '',
+    inventory_stock_exit_data : value,
+    inventory_stock_exit_quantity : _.sumBy(value, 'quantity'),
+  };
+}
+
+/**
+ * @function collect
+ * @param {object} params query parameters
+ * @return {object} { exitToPatient: [], exitToService: [], exitToDepot: [], exitToLoss: [] }
```
jniles: [An async function always returns a promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Return_value).  This means that whatever calls this function should call it like this:
```js
collection(params).then().catch()
```

server/controllers/stock/reports/stock/exit_report.js
```diff
@@ -0,0 +1,166 @@
...
+  try {
+    data = collect(params);
+  } catch (e) {
+    return next(e);
+  }
+
+  return fetchDepotDetails(params.depotUuid)
+    .then(depot => {
+      params.depotName = depot.text;
+      return data;
```
jniles: This is really interesting!  It works ... but I don't think it works the way you think it does ;)

The `collect()` function returns a promise, since it is an [async function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/async_function).  This means that a clearer representation of what you are doing is shown by renaming `data` to `promise`.
```js
try {
  promise = collect(params);
} catch (e) {
   // This catch() will  never get called since the thing in the try block is a promise
  return next(e);
}

return  fetchDepotDetails(params.depotUuid)
  .then(depot => {
    params.depotName = depot.text;
    return promise;
   })
   .then(/* etc... */);
```
Since the `collect()` returns a promise, the errors will be thrown in the promise chain and cannot be caught by `try/catch`.  This makes your `try/catch` unnecessary.

You could rewrite it like this:
```js
data = collect(params);

return  fetchDepotDetails(params.depotUuid)
  .then(depot => {
    params.depotName = depot.text;
    return data;
   })
   .then(/* etc... */);
```
without changing the logic of the function.  And this is almost equivalent to the following:
```js
return  fetchDepotDetails(params.depotUuid)
  .then(depot => {
    params.depotName = depot.text;
    return collect(params);
   })
   .then(/* etc... */);
```
It's just the `collect()` function fires a bit earlier.

So!  It works, but the `try/catch` is probably unnecessary. :)

server/controllers/stock/reports/stock/exit_report.js
```diff
@@ -0,0 +1,166 @@
...
+    includeDepotExit,
+    includeLossExit,
+  } = params;
+
+  const data = {};
+
+  try {
+    // get stock exit to patient
+    if (includePatientExit) {
+      data.exitToPatient = await StockExitToPatient.fetch(depotUuid, dateFrom, dateTo, showDetails);
```
jniles: This is cool!  However, it will also be kind of slow.  It says that we will wait for each one of these values in series.

You can speed it up by executing these queries in parallel.  Just push each value into an array and use `Promise.all()`!  Like this:
```js
const promises = [];
if (includePatientExit) {
  promises.push(StockExitToPatient.fetch(depotUuid, dateFrom, dateTo, showDetails))
}

 // etc, do that for the rest..

 return q.all(promises)

   // convert from array to data object
  .spread(exitToPatient, exitToService /* .. etc */) => ({
    exitToPatient,
    exitToService,
    // etc...
  }));
```
[Here's a fun link that shows cool things you can do with promises](http://exploringjs.com/es6/ch_promises.html#_forking-and-joining-computations-via-promiseall).  Also, [this SO answer](https://stackoverflow.com/a/31414472) gives some really good rules about how to think about promises.

test/end-to-end/reports/stock_exit/stock_exit.spec.js
```diff
@@ -0,0 +1,50 @@
+const chai = require('chai');
+const helpers = require('../../shared/helpers');
+
+helpers.configure(chai);
```
jniles: I only discovered this recently but ... we don't need to call `helpers.configure(chai)` in every module.  It will be configured globally in our protractor.conf.js.  So you can remove this line :)

jniles (commented):
I've made a few suggestions for improvements.

I like the direction you are going with async functions, but they will take a while to learn and get used to.  We can talk about these tomorrow - I think there might be some confusion about how they work.

mbayopanda:

client/src/modules/reports/generate/stock_exit/stock_exit.config.js
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  function checkCachedConfiguration() {
+    vm.reportDetails = angular.copy(cache.reportDetails || {});
+  }
+
+  function onExitTypeChange() {
+    // be sure at least one checkbox is checked
+    const sum = vm.includePatientExit + vm.includeServiceExit + vm.includeDepotExit + vm.includeLossExit;
+    vm.hasOneChecked = sum > 0;
```
jniles, line 94: :+1:

server/lib/util.js
```diff
@@ -151,3 +153,32 @@ function loadDictionary(key, dictionaries = {}) {
...
+
+/**
+ * @function convertStringToNumber
+ *
+ * @description
+ * look into an object and convert each string number to number if it is possible
+ *
+ * @param {object} obj An object in which we want to convert value of each property into the correct type
+ */
+function convertStringToNumber(obj) {
```
jniles, line 179: A better name for this is probably `parseObjectValuesToNumbers()` or something similar.  I know it is long, but something should tell the programmer that this takes in an object and not a string..

server/lib/util.js
```diff
@@ -151,3 +153,32 @@ function loadDictionary(key, dictionaries = {}) {
...
+  _.keys(obj).forEach(property => {
```
jniles, line 180: Alternatively:
```js
return _.mapValues((value) => stringToNumber(value));
```

jniles (approved):
Looks good to me!

bors r+

PR #2718, opened 2018-04-27
feat(journal grouping) include Date and Record in the headers

### LINKED ISSUE #2692: Journal should include "Date" and "Record" in the headers of transactions
At Tshikaji, they are used to the previous 1.x grid (SlickGrid) which was locked in the "transaction mode", meaning that transactions are grouped with a header and subsequent rows.  In 2.x, we give you the possibility to remove this.

In grouped mode, all the unchanging properties of the transaction should be found in the header.  Currently, we are missing two:
 1. Record
 2. Date

See below:
adddateandrecordgroup
_Fig 1: No Date And Record_

Your task is to put those values in the transaction header so that they are visible even when the transaction is collapsed.

jeremielodi:
closes #2692

### DIFF
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -200,10 +200,7 @@ function JournalController(
     displayName : 'TABLE.COLUMNS.RECORD',
     headerCellFilter : 'translate',
     visible : true,
-    cellTemplate :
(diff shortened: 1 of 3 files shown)


jniles (approved):
LGTM.  Well done!


bors r+

PR #2719, opened 2018-04-27
chore(db): ignore duplicate keys in triggers

jniles:
~~DO NOT MERGE.  This is experimental.~~  It works!

When using bidirectional data sync, the triggers for INSERTs are fired on the rows that are synced, sometimes causing PRIMARY KEY errors if the table has a trigger on it that writes metadata elsewhere.  The most common error is in the `entity_map` and `document_map` tables.

The triggers for creating these maps now ignore the row if a key already exists for it.


jniles:
This is needed for the bidirectional sync that I'm working on today.

bors r+

PR #2720, opened 2018-04-30
feat(registry) rendering displayname on server

jeremielodi:
Some time we export reports (Excel) with the database columns' labels
witch is not understandable for the end user, This PR tries to solve that issue.
Other cases, as described at #2333 will also be treated in this PR.

### DIFF
--- a/client/src/js/services/grid/GridColumns.js
+++ b/client/src/js/services/grid/GridColumns.js
@@ -1,8 +1,8 @@
 angular.module('bhima.services')
-.service('GridColumnService', GridColumnService);
+  .service('GridColumnService', GridColumnService);
 
 GridColumnService.$inject = [
-  'uiGridConstants', 'AppCache', '$uibModal', 'util', '$timeout',
+  'uiGridConstants', 'AppCache', '$uibModal', 'util', '$translate',
 ];
 
 /**
@@ -19,9 +19,9 @@ GridColumnService.$inject = [
  *
  * @todo - investigate using ui-grid-saveState for caching the column
  */
-function GridColumnService(uiGridConstants, AppCache, Modal, util, $timeout) {
+function GridColumnService(uiGridConstants, AppCache, Modal, util, $translate) {
   /** @const cache alias for this service */
-  var serviceKey = '-Columns';
+  // const serviceKey = '-Columns';
 
   /**
    * @method cacheDefaultColumnVisibility
@@ -39,15 +39,14 @@ function GridColumnService(uiGridConstants, AppCache, Modal, util, $timeout) {
     // this.defaults is set.
     if (Object.keys(this.defaults).length > 0) { return; }
 
-    var defaults = this.defaults;
-    var cache = this.cache;
+    const { defaults } = this;
 
-    angular.forEach(this.gridOptions.columnDefs, function (defn) {
-      var field = defn.field;
+    angular.forEach(this.gridOptions.columnDefs, (defn) => {
+      const { field } = defn;
 
       // only use the fields have usable names
       if (field) {
-        var column = api.grid.getColumn(field);
+        const column = api.grid.getColumn(field);
 
         // cache the default visible value
         defaults[field] = column.visible;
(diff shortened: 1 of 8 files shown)


client/src/modules/journal/journal.js
```diff
@@ -382,13 +382,22 @@ function JournalController(
     Modal.openReports({ url, params });
   };
 
+  function getDisplayNames() {
```
jniles: It would be nice to attach this to a service - maybe the [GridColumnService](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/grid/GridColumns.js#L22).

It could take in the column definitions as a parameter and return the new object.

client/src/modules/journal/journal.js
```diff
@@ -382,13 +382,22 @@ function JournalController(
...
+    });
+    return displayNames;
+  }
   vm.downloadExcel = () => {
+
     const filterOpts = Journal.filters.formatHTTP();
     const defaultOpts = {
       renderer : 'xlsx',
       lang : Languages.key,
+      renameKeys : true,
```
jniles, line 391: Nice addition 👍

server/controllers/finance/reports/journal/index.js
```diff
@@ -58,8 +66,10 @@ function postingJournalExport(req, res, next) {
       return report.render({ rows, totals });
     })
     .then((result) => {
+
       if (result.headers.type === 'xlsx') {
-        res.xls(result.headers.filename, util.dateFormatter(result.report.rows));
+        const rows = (renameKeys) ? util.renameKeys(result.report.rows, displayNames) : result.report.rows;
```
jniles: It would be cool if we could do this for the CSV and XLS reporters.  Could we make this part of the [ReportManager.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/ReportManager.js#L64) class?

server/lib/util.js
```diff
@@ -182,3 +184,27 @@ function convertStringToNumber(obj) {
...
   return obj;
 }
+
+/*
+ * rename an object's keys
+ */
+
+function renameKeys(objs, newKeys) {
+  const stringType = 'string';
+  const formatedKeys = (typeof stringType) ? JSON.parse(newKeys) : newKeys;
```
jniles: Just FYI - this `typeof` code isn't complete.  I think you meant to write `typeof newKeys === stringType`.

server/lib/util.js
```diff
@@ -182,3 +184,27 @@ function convertStringToNumber(obj) {
   });
   return obj;
 }
+
+/*
+ * rename an object's keys
+ */
+
+function renameKeys(objs, newKeys) {
```
jniles, line 192: I think you can use [lodash](https://lodash.com/docs/4.17.10) quite effectively here.  I can rewrite this to be two lines:

```js
const keyMap = _.isString(newKeys) ? JSON.parse(newKeys) : newKeys;
return objs.map(row => _.mapKeys(row, (value, key) => keyMap[key] || key));
```

It works like this:
 1. Parses the JSON if necessary to get the `{ column_id : translatedKey }` map.
 2. Loops through the database rows with `objs.map()`.
 3. For each row object, it uses [lodash's mapKeys](https://lodash.com/docs/4.17.10#mapKeys) function to loop through the keys and replaces the key with a new value.  The new value will either be the translated key found in `keyMap[key]` or itself.

jniles (commented):
@jeremielodi, the end result is exactly correct.  I've made some suggestions for improving the code on both the client and server.

Additionally, this code could use some server-unit tests.  Specifically, we should test the `server/lib/util` functions added here.

If you use my version of the code, you only need to test the `renameKeys` function (since it is the only one).  If you use the two separated functions they will both need to be tested.

You will want to test the following conditions:
 1.  That the code works with an array of objects.  You should have something like this:
```js
const a = [{ id : 1}];
const keyMap = { id : 'hello' };
const result = util.renameKeys(a, keyMap);
expect(result).to.deep.equal([{ hello : 1 }]);
```
 2. The code should work with an empty array, returning an empty array back.

server/lib/util.js
jeremielodi: your're right

client/src/js/services/grid/GridColumns.js
```diff
@@ -63,20 +62,32 @@ function GridColumnService(uiGridConstants, AppCache, Modal, util, $timeout) {
...
-    }.bind(this));
+    });
   }
 
+
+  /**
+   * returns [{field1 : displayName1}, {field2 : displayName2}, ...]
+   * this function is useful for renaming keys
+   */
+  Columns.prototype.getDisplayNames = function getDisplayNames() {
```
jniles, line 84: Awesome 👍

jniles (approved):
LGTM.  Good work.  

bors r+

PR #2722, opened 2018-05-01
Added minor comment to INSTALL.md file

jmcameron:
Where the .env.distribution file is mentioned, added a note that the DB entries correspond to MySQL.


jniles (approved):
Thanks for the clarification!

bors r+

PR #2727, opened 2018-05-03
feact(xlsx report)Remove the need for res.xls() to send an XLS file

### LINKED ISSUE #2723: Remove the need for res.xls() to send an XLS file
The XLS reporter hacks the `res` object to add a new method `res.xls()` for sending XLS files.  This makes all of our code that wants to send XLS have to be changed to include the following check:
```js
if (result.headers.type === 'xlsx') {
  res.xls(result.headers.filename, data);
} else {
  res.set(result.headers).send(result.report);
}
```

The underlying code _must_ use `res.send()` at some point.  We should figure out how to rewrite the XLS reporter to avoid this check.  That way, the HTML, JSON, CSV, PDF, and XLS reporters will all use the same interface.

jeremielodi:
This PR rewrire xlsx.js renderer, we were using json2excel that is why we were obliged to have res.xls() method in our code. Now we switch xlsx.js to excel4node witch is much better than json2excel.
closes #2723

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -21,10 +21,6 @@
  * @copyright IMA World Health 2016
  */
 
-// json2xls has a strict-mode octal error in underlying modules.
-// Workaround - required before 'use-strict'
-const json2xls = require('json2xls');
-
 require('use-strict');
 const dotEnv = require('dotenv');
 
@@ -35,11 +31,6 @@ const express = require('express');
 const debug = require('debug')('app');
 
 const app = express();
-/*
-  * json2xls middelware used to send xlsx file as http response
-  *example : res.xls('data.xlsx', jsonArray);
-*/
-app.use(json2xls.middleware);
 
 /**
  * @function configureEnvironmentVariables
--- a/server/controllers/finance/reports/account_statement/index.js
+++ b/server/controllers/finance/reports/account_statement/index.js
@@ -9,7 +9,6 @@
 const _ = require('lodash');
 const db = require('../../../../lib/db');
 const ReportManager = require('../../../../lib/ReportManager');
-const util = require('../../../../lib/util');
 const generalLedger = require('../../generalLedger');
 
 const REPORT_TEMPLATE = './server/controllers/finance/reports/account_statement/report.handlebars';
@@ -66,11 +65,7 @@ function report(req, res, next) {
       });
     })
     .then((result) => {
-      if (result.headers.type === 'xlsx') {
-        res.xls(result.headers.filename, util.dateFormatter(result.report.rows));
-      } else {
-        res.set(result.headers).send(result.report);
-      }
+      res.set(result.headers).send(result.report);
     })
     .catch(next)
     .done();
--- a/server/controllers/finance/reports/journal/index.js
+++ b/server/controllers/finance/reports/journal/index.js
@@ -10,7 +10,6 @@ const _ = require('lodash');
 const ReportManager = require('../../../../lib/ReportManager');
 const db = require('../../../../lib/db');
 const Journal = require('../../journal');
-const util = require('../../../../lib/util');
 
 const REPORT_TEMPLATE = './server/controllers/finance/reports/journal/report.handlebars';
 
@@ -22,6 +21,13 @@ exports.postingReport = postingJournalExport;
  * @method postingJournalExport
  */
 function postingJournalExport(req, res, next) {
+
+  /*
+   theses below properties are used for rename the result keys
+    some time export report with the database columns' labels
+    whitch is not understandable for the end user
+  */
+
   const options = _.extend(req.query, {
     filename                 : 'POSTING_JOURNAL.TITLE',
     orientation              : 'landscape',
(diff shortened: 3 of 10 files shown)


server/app.js
```diff
@@ -21,10 +21,6 @@
  * @copyright IMA World Health 2016
  */
 
-// json2xls has a strict-mode octal error in underlying modules.
-// Workaround - required before 'use-strict'
-const json2xls = require('json2xls');
```
jniles, line 26: 🕺

server/app.js
```diff
@@ -35,11 +31,6 @@ const express = require('express');
 const debug = require('debug')('app');
 
 const app = express();
-/*
-  * json2xls middelware used to send xlsx file as http response
-  *example : res.xls('data.xlsx', jsonArray);
-*/
-app.use(json2xls.middleware);
```
jniles, line 42: 👯‍♂️

server/controllers/finance/reports/account_statement/index.js
```diff
@@ -66,11 +65,7 @@ function report(req, res, next) {
       });
     })
     .then((result) => {
-      if (result.headers.type === 'xlsx') {
-        res.xls(result.headers.filename, util.dateFormatter(result.report.rows));
-      } else {
-        res.set(result.headers).send(result.report);
-      }
+      res.set(result.headers).send(result.report);
```
jniles, line 68: 👍

server/controllers/finance/reports/journal/index.js
```diff
@@ -22,6 +21,13 @@ exports.postingReport = postingJournalExport;
  * @method postingJournalExport
  */
 function postingJournalExport(req, res, next) {
+
+  /*
+   theses below properties are used for rename the result keys
```
jniles: I think this comment is old and doesn't apply anymore.

server/lib/renderers/xlsx.js
```diff
@@ -3,34 +3,103 @@
...
+  // Add Worksheets to the workbook
+  const ws = wb.addWorksheet('Sheet 1');
+  const { rows } = data;
+  const firstObject = rows[0] || {};
+
+  // writing columns
+  const firstLineCols = Object.keys(firstObject);
+  let line = 1;
+  firstLineCols.forEach((key, index) => {
+    ws.cell(line, index + 1).string(key)
```
jniles: For consistency, could you put the `.string(key)` on the next line?  It should look like this:
```js
ws.cell(line,index + 1)
  .string(key)
  .style(headerStyle)
  .style(styleAllBorders);
```

server/lib/renderers/xlsx.js
```diff
@@ -3,34 +3,103 @@
...
+  if (_.isNumber(value)) {
+    return ws.cell(x, y).number(value);
+  }
+
+  if (_.isDate(value)) {
+    return ws.cell(x, y).date(value).style({
+      numberFormat : 'yyyy-mm-dd hh:mm:ss',
+    });
+  }
+  return ws.cell(x, y).string(''.concat(value || ''));
```
jniles: Personally, I think this following is easier to read:
```js
return ws.cell(x,y).string(`${value || ''}`);
```

jniles (commented):
@jeremielodi this is beautiful 👍  I really like where this is going.

The only blocking part of this PR is that we should add a unit test or two.  These should be server unit tests.  Specifically, it would be nice to see that:
 1. Calling `render([ /* .. json objects .. */ ]);` should return a buffer.
 2. Calling `render([])` should not error.

Excellent work so far!

test/server-unit/xlsx.js
```diff
@@ -0,0 +1,33 @@
...
+const { expect } = require('chai');
+const _ = require('lodash');
+
+describe('util.js', () => {
+
+  it('Shoul return a xlsx buffer', (done) => {
+    const data = {
+      rows : [{ Firstname : 'Alice', Lastname : 'Bob' }],
+    };
+    xlsx.render(data)
```
jniles, line 12: There isn't any need to use `done()` in these callbacks.  It works, but promises are prettier.  All you need to do is return the `xlsx.render()` call.  Like this:
```js
it('Should return a xlsx buffer', () => {
  const data = { /* some data */ };
  return xlsx.render(data).then((reportStream) => {
     expect(_.isBuffer(reportStream)).to.be.equal(true);
   });
});
```

jniles: [This is a good article](http://tobyho.com/2015/12/16/mocha-with-promises/) about it.

test/server-unit/xlsx.js
```diff
@@ -0,0 +1,33 @@
...
+      .catch((error) => {
+        done(error);
+      });
+  });
+
+  it('Shoul works for an empty object', (done) => {
+    const data = {};
+    xlsx.render(data)
+      .then(reportStream => {
+        expect(_.isBuffer(reportStream)).to.be.equal(true);
```
jniles, line 26: Just FYI, I think you can test for a buffer using [the `a()` method](http://www.chaijs.com/api/bdd/#atype-msg):
```js
expect(reportStream).to.be.a('buffer');
```

server/lib/renderers/xlsx.js
```diff
@@ -101,5 +103,5 @@ function setValue(ws, x, y, value) {
       numberFormat : 'yyyy-mm-dd hh:mm:ss',
     });
   }
-  return ws.cell(x, y).string(''.concat(value || ''));
+  return ws.cell(x, y).string(value || '');
```
jniles: 👍 much prettier

jniles (approved):
Assuming tests pass, this LGTM.  Great work.

bors r+

PR #2735, opened 2018-05-07
fix(reports): show 0 on footer

jniles:
A bug crept into the Account Statement report if the account was completely balanced (0.00).  The report would interpret this as not having totals and would default to the sum of the debit/credit column instead of the running balance.

### DIFF
--- a/server/controllers/finance/reports/reportAccounts/index.js
+++ b/server/controllers/finance/reports/reportAccounts/index.js
@@ -235,7 +235,9 @@ function getAccountTransactions(options, openingBalance = 0) {
       // if there is data in the transaction array, use the date of the last transaction
       const lastTransaction = bundle.transactions[bundle.transactions.length - 1];
       const lastDate = (lastTransaction && lastTransaction.trans_date) || options.dateTo;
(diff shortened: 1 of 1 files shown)


server/controllers/finance/reports/reportAccounts/index.js
```diff
@@ -235,7 +235,9 @@ function getAccountTransactions(options, openingBalance = 0) {
       // if there is data in the transaction array, use the date of the last transaction
       const lastTransaction = bundle.transactions[bundle.transactions.length - 1];
       const lastDate = (lastTransaction && lastTransaction.trans_date) || options.dateTo;
-      const lastCumSum = (lastTransaction && lastTransaction.cumsum) || (totals.balance * totals.rate);
+
+      const hasLastCumSum = !_.isUndefined(lastTransaction && lastTransaction.cumsum);
+      const lastCumSum = hasLastCumSum ? lastTransaction.cumsum : (totals.balance * totals.rate);
```
mbayopanda, line 240: 👍

mbayopanda (approved):
bors r+

PR #2738, opened 2018-05-08
feat(enterprise): toggle password policy rules

### LINKED ISSUE #2730: Add Enterprise Setting to block deletions from the registries
We may have hit our first case of data loss due to someone deleting a record from the registry.  We should definitely allow the enterprise to turn this behavior on and off via a flag in the enterprise settings.

jniles:
The password policy currently is not practical for our users in the field.  This feature allows an administrator to toggle off the password enforcement feature.

Closes #2730.

I'll be adding some tests for the password checking component before this should be merged.

_Fig: The changes in action_


jniles:

jeremielodi:
I'm still on odk, I will do it this afternoon

jniles:
Sounds good!  ODK is the priority. 🕺

jniles:

client/src/i18n/fr/enterprise.json
```diff
@@ -15,7 +15,9 @@
       "ENABLE_PREPAYMENTS_LABEL" : "Activer les Cautions",
       "ENABLE_PREPAYMENTS_HELP_TEXT" : "L'activation de cette fonctionnalité permet aux patients de déposer des fonds à la fenêtre de paiement avant d'être facturé. Si le patient a un solde créditeur, ses factures seront automatiquement payées après facturation.",
       "ENABLE_DELETE_RECORDS_LABEL" : "Activer la Suppression d'Enregistrements",
-      "ENABLE_DELETE_RECORDS_HELP_TEXT" : "Cette fonctionnalité permet aux utilisateurs de supprimer des transactions comme une alternative à l'inversion des transactions directement à partir des registres."
+      "ENABLE_DELETE_RECORDS_HELP_TEXT" : "Cette fonctionnalité permet aux utilisateurs de supprimer des transactions comme une alternative à l'inversion des transactions directement à partir des registres.",
+      "ENABLE_PASSWORD_VALIDATION_LABEL" : "",
```
jeremielodi: these translate keys have no value, do you need them? if not, go ahead and remove them.
I see, you use it when calling the component on the view, but you have to provide values, right?

client/src/js/components/bhStrengthMeter.js
```diff
@@ -1,21 +1,28 @@
...
+/**
+ * @function StrengthMeterController
+ *
+ * @description
+ * This is a display only component that evaluates a password's strength as the
+ * user types it.  It displays these values in a meter underneath the password
+ * input.
+ */
+function StrengthMeterController(PasswordMeterService, Session) {
+  this.showStrengthMeter = Session.enterprise.settings.enable_password_validation;
```
jeremielodi, line 26: okay

client/src/i18n/fr/enterprise.json
jniles: Whoops, you are right.  Guess I forgot to translate these values :(

PR #2739, opened 2018-05-08
feat(inventory seach modal) add labels to ratio buttons

### LINKED ISSUE #2666: Inventory Search Modal radio buttons need labels
There are a couple of options that are radio buttons on the inventory's search modal.  They need labels - currently they look like they might be part of the same set of options, just overflowing onto several lines.

See below:
inventoryradiobuttonsshouldhavelabels
_Fig 1: How many options are here?  One or Two?_

jeremielodi:
closes #2666


jniles (approved):
LGTM.  

Remember we have the `bhYesNoRadios` that could be used as well!

bors r+

PR #2743, opened 2018-05-09
Stock distribution to patient linked with invoice

mbayopanda:
This PR : 
- Add link between stock movement with invoice for stock exit to patient
- Add a `bhFindInvoice` component for finding invoice for a given patient
- Update the stock exit report for displaying also invoice information
- Update `End 2 End` tests for testing also the stock exit to patient with linked invoice
- Add lodash on the client side, so we can use for instance `_.groupBy()` on the client side
- Remove unused dependencies `Chart.js` and `Angular-chart`


client/src/js/components/bhFindInvoice.js
```diff
@@ -0,0 +1,96 @@
+angular.module('bhima.components')
+  .component('bhFindInvoice', {
+    controller : FindInvoiceComponent,
+    templateUrl : 'modules/templates/bhFindInvoice.tmpl.html',
+    bindings : {
+      patientUuid :       '<', // patient uuid - to restrict search to this patient
+      onSearchComplete :  '&', // bind callback to call when data is available
+      disabled :          '<', // bind disable behavior
```
jniles: This should probably be optional: `<?`

bower.json
```diff
@@ -32,14 +32,13 @@
...
-    "Chart.js": "2.7.1",
-    "angular-chart.js": "1.1.1",
     "components-font-awesome": "4.7.0",
     "angular-ui-select": "0.17.1",
     "JsBarcode": "jsbarcode#3.8.0",
     "angular-touch": "1.6.6",
     "angular-growl-notifications": "2.6.0",
-    "webcam-directive": "3.1.1"
+    "webcam-directive": "3.1.1",
+    "lodash": "^4.17.10"
```
jniles: I don't actually see where you use `_.groupBy()`.  Where can I find it?

client/src/modules/templates/bhFindInvoice.tmpl.html
```diff
@@ -0,0 +1,43 @@
...
+          translate>
+          FORM.BUTTONS.SEARCH
+        </button>
+      </span>
+    </div>
+    <p 
+      ng-if="!$ctrl.invoiceFound && FindInvoiceForm.$submitted" 
+      class="help-block" 
+      translate="FORM.INFO.PATIENT_INVOICE_NOT_FOUND" 
+      translate-values="{{ $ctrl.translateValues }}">
```
jniles: I can't figure out where this is defined.  Did you mean to write `$ctrl.translate` to refer to [this line](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2743/files#diff-e3fb58cdaea9a73c062994612f047c19R67)?

client/src/js/components/bhFindInvoice.js
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+FindInvoiceComponent.$inject = [
+  'PatientService', 'PatientInvoiceService', 'AppCache', 'NotifyService',
+];
+
+/**
+ * The Find Invoice Component
+ */
+function FindInvoiceComponent(Patients, PatientInvoice, AppCache, Notify) {
```
jniles: Do you need AppCache here?  It doesn't like you ever use it.

client/src/i18n/en/stock.json
```diff
@@ -49,10 +49,13 @@
     "FLUX"            : "Reason",
     "FROM"            : "From",
     "GROUPING"        : "Grouping",
+    "JOIN_INVOICE"    : "Join an invoice", 
     "INCREASE"        : "Increase",
     "INCREASE_HELP"   : "Adjustment of stock by increasing the quantity of lots of products in stock",
     "INVENTORY"       : "Inventory",
     "INVENTORY_GROUP" : "Inventory Group",
+    "INVENTORIES_NOT_AVAILABLE" : "Inventories not available",
```
jniles: In English, this would be written as "Inventory not available".  The plural here is the same as the singular. 👍

client/src/js/app.js
```diff
@@ -1,7 +1,7 @@
 var bhima = angular.module('bhima', [
   'bhima.controllers', 'bhima.services', 'bhima.directives', 'bhima.filters',
   'bhima.components', 'bhima.routes', 'bhima.constants', 'ui.bootstrap',
-  'pascalprecht.translate', 'ngStorage', 'chart.js',
+  'pascalprecht.translate', 'ngStorage',
```
jniles, line 4: 👍

bower.json
```diff
@@ -32,14 +32,13 @@
     "moment": "2.20.1",
     "angular-dynamic-locale": "0.1.32",
     "ngstorage": "0.3.11",
-    "Chart.js": "2.7.1",
```
jniles, line 35: 💃

client/src/modules/stock/exit/exit.js
```diff
@@ -285,6 +285,13 @@ function StockExitController(
       .then(inventories => {
         vm.loading = false;
         vm.selectableInventories = angular.copy(inventories);
+
+        // map of inventories by inventory uuid
+        vm.mapSelectableInventories = _.reduce(vm.selectableInventories, (aggregate, value) => {
```
jniles: I'm not sure it's worth adding the entire lodash library just for a _.reduce().  Isn't `vm.selectedInventories` an array?

client/src/modules/stock/exit/exit.js
```diff
@@ -318,10 +325,43 @@ function StockExitController(
...
       .catch(Notify.handleError);
   }
 
+  // load inventories from an invoice
+  function loadInvoiceInventories(patient) {
+    vm.inventoryNotAvailable = [];
+    vm.stockForm.clear();
+    vm.gridApi.core.notifyDataChange(uiGridConstants.dataChange.EDIT);
+
+    if (!patient || !patient.invoice) { return; }
```
jniles: Should this `return` go before `vm.stockForm.clear()`?

client/src/modules/stock/exit/exit.js
```diff
@@ -318,10 +325,43 @@ function StockExitController(
...
+    const nbItems = patient.invoice.items.length;
+
+    if (!nbItems) { return; }
+
+    let inventoryAvailableIndex = 0;
+
+    patient.invoice.items.forEach((item) => {
+      if (vm.mapSelectableInventories[item.inventory_uuid]) {
+        addItems(1);
+        vm.stockForm.store.data[inventoryAvailableIndex].inventory = vm.mapSelectableInventories[item.inventory_uuid];
```
jniles: Hmm..  it looks like you need to have the last item inserted into the store.  The [PatientInvoiceFormService](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/invoices/patientInvoiceForm.service.js#L375) deals with this by simply returning the last element inserted into the store. I propose we do the same thing here.

Modify [the StockFormService](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/stock/StockForm.service.js#L52) to `return elt;` as the last line.  Then you can write this code, which will be easier to read:
```js
patient.invoice.items.forEach((item) => {
  if (vm.mapSelectableInventories[item.inventory_uuid]) {
    const row = vm.stockForm.addItems(1);
    row.inventory = vm.mapSelecteableInventories[item.inventory_uuid];
    row.inventory_uuid = item.inventory_uuid;
    // etc..
  }
});

File: client/src/modules/stock/exit/exit.js (hunk 8 above)
Comment (jniles) [reply client/src/modules/stock/exit/exit.js]: Also, you might consider making this a `new Store()`.  I think it's probably what you want - an ability to look up the entire inventory by its `uuid`.  Like this:
```js
let selectableInventoriesStore = new Store({ key : 'inventory_uuid', data : vm.selectableInventories });
```
and then later you can use it like:
```js
patient.invoice.items.forEach((item) => {
  const inventory = selectableInventoriesStore.get(item.inventory_uuid);
  if (inventory) {
    addItems(1);
    // do something
  }
});

File: server/models/schema.sql [hunk 11]
```diff
@@ -1812,6 +1812,7 @@ CREATE TABLE `stock_movement` (
   `is_exit`         TINYINT(1) NOT NULL,
   `user_id`         SMALLINT(5) UNSIGNED NOT NULL,
   `reference`       INT(11) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `invoice_uuid`    BINARY(16) NULL,
```
jniles, line 1854: I thought this is what the `document_uuid` was for?

client/src/js/components/bhFindInvoice.js
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * @method selectInvoice
+   *
+   * @param {object} invoice The invoice object
+   */
+  function selectInvoice(invoice) {
+    vm.invoiceFound = true;
+
+    const elementId = 'search';
+    const searchButton = document.getElementById(elementId);
```
jniles: Would it be possible to use `$document` instead of `document` here?  It will make it easier to test.

client/src/js/components/bhFindPatient.js
```diff
@@ -65,16 +66,21 @@ function FindPatientComponent(Patients, AppCache, Notify, SessionService, bhCons
...
     loadDefaultOption(cache.optionKey);
 
     // call the onRegisterApi() callback with the
     vm.onRegisterApi({
-      api : { reset: vm.reset, searchByUuid: searchByUuid },
+      api : { reset : vm.reset, searchByUuid },
     });
   };
 
+  vm.$onChanges = function onChanges(changes) {
```
jniles, line 78: This change should be accompanied by a test in the [bhFindPatient](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/client-unit/components/bhFindPatient/controller.spec.js) controller tests.

client/src/modules/invoices/patientInvoice.service.js
```diff
@@ -184,5 +190,24 @@ function PatientInvoiceService(
     return $httpParamSerializer(options);
   };
 
+  /**
+   * findInvoice
+   */
+  function findInvoice(invoiceUuid) {
```
jniles: Does this ever get used?  It doesn't look like it.

client/src/modules/stock/exit/exit.js
```diff
@@ -18,8 +18,8 @@ StockExitController.$inject = [
  * @todo Implement caching data feature
  */
 function StockExitController(
-  Depots, Inventory, Notify, Session, util, bhConstants, ReceiptModal, StockForm, Stock,
-  StockModal, uiGridConstants, $translate, AppCache, moment, GridExportService
+  Depots, Inventory, Notify, Session, util, bhConstants, ReceiptModal, StockItem, StockForm, Stock,
+  StockModal, uiGridConstants, $translate, AppCache, moment, GridExportService, _,
```
jniles: Good use of dependency injection 👍

jniles (commented):
A few comments here.  I'll follow up with more specifics in person today.

server/models/schema.sql
mbayopanda: I don't think so, `document_uuid` was for the document generated for a stock movement (SM.X.Y).

client/src/modules/templates/bhFindInvoice.tmpl.html
mbayopanda: You're right 👍

client/src/modules/stock/exit/exit.js
```diff
@@ -132,7 +132,10 @@ function StockExitController(
       },
     ],
     data : vm.stockForm.store.data,
-    fastWatch : true,
+
+    // fastWatch to false is required for updating the grid correctly for
```
jniles, line 136: Nice comment :+1:

client/src/js/components/bhFindInvoice.js
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * @method selectInvoice
+   *
+   * @param {object} invoice The invoice object
+   */
+  function selectInvoice(invoice) {
+    vm.invoiceFound = true;
+
+    const elementId = 'search';
+    const searchButton = $window.document.getElementById(elementId);
```
jniles: :+1:

client/src/js/components/bhFindPatient.js
```diff
@@ -260,10 +265,10 @@ function FindPatientComponent(Patients, AppCache, Notify, SessionService, bhCons
   function loadDefaultOption(optionKey) {
 
     // default to findById
-    optionKey = optionKey || 'findById';
+    const _optionKey = optionKey || 'findById';
```
jniles: Just FYI, you can set default options in the function parameters.  Like this:
```js
loadDefaultOption(optionKey = 'findById') {
  findBy(optionKey);
}
```

client/src/modules/reports/generate/stock_exit/stock_exit.config.js
```diff
@@ -1,12 +1,12 @@
 angular.module('bhima.controllers')
-  .controller('stock_exitController', InventoryFileConfigController);
+  .controller('stock_exitController', StockExitConfigController);
```
jniles, line 2: 💃

gulpfile.js
```diff
@@ -72,6 +72,10 @@ const paths = {
       // jquery
       'client/vendor/jquery/dist/jquery.min.js',
       'client/vendor/cropper/dist/cropper.js',
+
+      // lodash
+      'client/vendor/lodash/dist/lodash.min.js',
```
jniles: I think this should be removed.

server/controllers/finance/invoice/lookupConsumableInvoice.js
```diff
@@ -0,0 +1,64 @@
...
+ * this function return an invoice details and all consumable inventories
+ * related to this invoice
+ */
+function lookupConsumableInvoicePatient(req, res, next) {
+  const params = req.query;
+  const record = {};
+
+  let invoiceDetailQuery =
+    `SELECT
+      BUID(invoice.uuid) as uuid, CONCAT_WS('.', '${identifiers.INVOICE.key}',
```
jniles: In the case of multiple projects, this might produce the incorrect result.  I can't remember why though - just that we found it at HEV.

To retrieve the identifier, it's better to just JOIN on document_map.

server/models/schema.sql
jniles: The document meaning the `voucher`?

jniles (commented):
Test finally pass! 👍 Well done.

It would be really good to have unit tests for `bhFindInvoice`.  They should be pretty similar to a mix between the [bhBarcodeScanner tests](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/client-unit/components/bhBarcoderScanner.spec.js) and the [bhFindPatient tests](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/client-unit/components/bhFindPatient/controller.spec.js).  Let me know if you need help with these.

test/client-unit/components/bhFindInvoice.spec.js
```diff
@@ -0,0 +1,217 @@
...
+    $httpBackend = _$httpBackend_;
+    PatientInvoice = _PatientInvoiceService_;
+    Notify = _NotifyService_;
+    Q = _$q_;
+
+    element = angular.element(template);
+    element = $compile(element)($scope);
+
+    // overwrite the getElementById function
+    _$window_.document.getElementById = (id) => find(element, `#${id}`);
```
jniles, line 68: 👍

test/client-unit/components/bhFindInvoice.spec.js
```diff
@@ -0,0 +1,217 @@
...
+  let Notify;
+
+  // find element function without jquery dependencies
+  const find = (elm, selector) => elm[0].querySelector(selector);
+
+  // inject dependencies
+  beforeEach(inject((_$rootScope_, _$compile_, _$httpBackend_, _$q_, _PatientInvoiceService_, _$window_, _NotifyService_) => {
+    $compile = _$compile_;
+    $scope = _$rootScope_.$new();
+    $httpBackend = _$httpBackend_;
```
jniles: If you never use $httpBackend for a test, you probably don't need to import it.

test/client-unit/components/bhFindInvoice.spec.js
```diff
@@ -0,0 +1,217 @@
...
+    const input = find(element, '#search-input');
+    expect(angular.element(input).length).to.equal(1);
+
+    const button = find(element, '#search-button');
+    expect(angular.element(button).length).to.equal(1);
+  });
+
+  it('Runs the search method when the ENTER key is pressed', () => {
+    const ENTER_KEY = 13;
+    const FormController = $scope.$$childTail.FindInvoiceForm || {};
```
jniles: Why would this ever be empty..?

test/client-unit/components/bhFindInvoice.spec.js
```diff
@@ -0,0 +1,217 @@
...
+
+  it('#bhFindInvoice component exists and it is compiled', () => {
+    const input = find(element, '#search-input');
+    expect(angular.element(input).length).to.equal(1);
+
+    const button = find(element, '#search-button');
+    expect(angular.element(button).length).to.equal(1);
+  });
+
+  it('Runs the search method when the ENTER key is pressed', () => {
```
jniles, line 86: These are really great tests, @mbayopanda .  The descriptions are clear, each test is small and self-contained.  Really great work!

test/client-unit/components/bhFindInvoice.spec.js
```diff
@@ -0,0 +1,217 @@
...
+    Q = _$q_;
+
+    element = angular.element(template);
+    element = $compile(element)($scope);
+
+    // overwrite the getElementById function
+    _$window_.document.getElementById = (id) => find(element, `#${id}`);
+
+    // overwrite the notification errorhandler and track it
+    Notify.handleError = () => {};
```
jniles: Why do you need to make this an empty function before making it a spy?

jniles (approved):
Code looks pretty good.  Your unit tests are awesome!

There are 19 commits in the PR, most of which are from other PRs or merge commits or random bug fixes unrelated to the functionality here.  Could you try squashing them all into a single commit?  (If there is any confusion, [this SO answer](https://stackoverflow.com/a/5189600) talks about squashing).  That way we have a more focused git history when we eventually do our next release.

test/client-unit/components/bhFindInvoice.spec.js
mbayopanda: Here we want just to be notified when the `Notify.handleError` is called, and not to test the `Notify.handleError`.

mbayopanda:
@jniles it seems like the result is still the same...

PR #2744, opened 2018-05-10
fix(cash): restore show receipt feature

### LINKED ISSUE #2733: Cash Registry's openReceiptModal() method was removed
It appears that the Cash Registry's `openReceiptModal()` was removed so that a user can not longer bring up a PDF receipt.   I think if we just restore the method, everything should work as planned.

jniles:
This commit fixes the cash payments registry link to receipt in the action dropdown.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/2733

### DIFF
--- a/client/src/js/components/bhReceipt.js
+++ b/client/src/js/components/bhReceipt.js
@@ -1,5 +1,7 @@
-var bhReceiptTemplate =
-  '<a ng-click="$ctrl.open()" href="">{{ $ctrl.displayValue }}</a>';
+const bhReceiptTemplate = `
+  <a ng-click="$ctrl.open()" href="">
+    {{ $ctrl.displayValue }}
+  </a>`;
 
 angular.module('bhima.components')
   .component('bhReceipt', {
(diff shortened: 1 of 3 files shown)


jniles:

client/src/js/components/bhReceipt.js
```diff
@@ -1,34 +1,34 @@
-var bhReceiptTemplate =
-  '<a ng-click="$ctrl.open()" href="">{{ $ctrl.displayValue }}</a>';
+const bhReceiptTemplate = `
+  <a ng-click="$ctrl.open()" href="">
+    {{ $ctrl.displayValue }}
+  </a>`;
```
mbayopanda, line 4: 👍

client/src/js/components/bhReceipt.js
```diff
@@ -1,34 +1,34 @@
...
 
 angular.module('bhima.components')
   .component('bhReceipt', {
     template : bhReceiptTemplate,
     controller  : bhReceiptController,
     bindings    : {
       value : '<',
-      displayValue : '<',
       type : '@',
+      displayValue : '<',
```
mbayopanda, line 13: Why is it restored in this PR ? Because in this PR i don't see where it is used.

mbayopanda (approved):
LGTM 👍

client/src/js/components/bhReceipt.js
jniles: It wasn't restored, just moved from the second position to the third.  It is used in the template on line 1.

PR #2748, opened 2018-05-10
feat(enterprise): disable record deletion

### LINKED ISSUE #2730: Add Enterprise Setting to block deletions from the registries

jniles:
The enterprise now has the option to disable the deletion of records
from the registries.  Deletion transaction from the posting journal is
still supported.

Closes #2730.

_Fig 1: The setting in action_.

### DIFF
--- a/server/controllers/admin/enterprises.js
+++ b/server/controllers/admin/enterprises.js
@@ -22,7 +22,8 @@ exports.list = function list(req, res, next) {
     sql = `
       SELECT id, name, abbr, email, po_box, phone,
         BUID(location_id) AS location_id, logo, currency_id,
-        gain_account_id, loss_account_id, enable_price_lock, enable_prepayments
+        gain_account_id, loss_account_id, enable_price_lock, enable_prepayments,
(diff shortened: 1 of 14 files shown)


jniles:

server/controllers/admin/enterprises.js
```diff
@@ -22,7 +22,8 @@ exports.list = function list(req, res, next) {
     sql = `
       SELECT id, name, abbr, email, po_box, phone,
         BUID(location_id) AS location_id, logo, currency_id,
-        gain_account_id, loss_account_id, enable_price_lock, enable_prepayments
+        gain_account_id, loss_account_id, enable_price_lock, enable_prepayments,
+        enable_delete_records
```
mbayopanda, line 26: 👍

mbayopanda (approved):
LGTM 

bors r+

PR #2755, opened 2018-05-15
feat(reports): Profit & Loss Statement

### LINKED ISSUE #810: Variation Income Expense report Implementation
## Description

The Income Expense Variation report, presents all income expense accounts and their balances in the current and the previous fiscal year, there is also a variation number in percentage and in amount to show how each account  behaves through fiscal years (previous and current).
## Technical Info

The data will be fetched from the general ledger or period total (a decision can be made after discussion).
The report belongs to a financial group and must be fully configurable.
- It must give the possibility of subtracting an amount to a specific account
- It must give the possibility of viewing the result with/without an account group (subtracting the value of the account group from the result value)
- It should give possibility of ignoring a specific transaction in the current or previous fiscal year
- It should list incomes accounts in one side and expenses accounts in an other side 

img_20161017_130120_1

@IMA-WorldHealth/bhima-core

### LINKED ISSUE #1781: (refactor) Income expense variance report
With the work done by @mbayopanda the new report client infrastructure has been merged in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1672. 

This issue tracks updating the Income vs. Expense variance report being implemented with the same structure. 

The client code should be very similar but it should be updated to use the new preview/ archive module infrastructure. 

The report should be fully tested using the new data set that will land in https://github.com/IMA-WorldHealth/bhima-2.X/pull/1780. An example of testing report outcomes using this data will be submitted soon. This issue will be updated with requirements for testing the report.

jniles:
This PR rewrites the bhFiscalPeriod selection component to be easier
to read and more user friendly.  In particular:
  1. The periods are now sorted by date in ascending order.
  2. The periods rely on moment for translation.
  3. The periodFrom selection limits the periodTo selection to only
  the same or greater dates.  This avoids a corner case where periodTo
  could be before periodFrom.
  4. The component uses the FiscalService to get periods instead of a
  custom service.

This PR refactors the income and expense report to be a profit and
loss statement.  The income & expense accounts are displayed as a tree
with the totals for each sub-group in the header.  The variation between
the present fiscal year and the previous year provided in subsequent
columns.

Additionally, the bhReportPeriodSelect has been removed and replaced
with the combined bhFiscalSelect and a new bhPeriodSelection component
that allows a user to first select a fiscal year, then select a period.

Closes #1781.  Closes #810.  Closes #2713.


jniles:
@mbayopanda could this get a review as well?

client/src/js/components/bhFiscalPeriodSelect.js
```diff
@@ -1,66 +1,66 @@
...
 
-    $ctrl.onSelectPeriodTo = function onSelectPeriodTo(selectedItem) {
-        $ctrl.onSelectPeriodToCallback({ period: selectedItem });
+        // sure the period are ordered in an ASC fashion
+        $ctrl.periods.sort((a, b) => {
+          if (a.start_date > b.start_date) {
+            return 1;
+          }
+          return -1;
+        });
```
mbayopanda, line 50: What about : 
```js
$ctrl.periods.sort((a, b) => a.start_date > b.start_date ? 1 : -1)
```

server/controllers/finance/reports/income_expense/index.js
```diff
@@ -0,0 +1,274 @@
...
+    })
+    .then(result => {
+      res.set(result.headers).send(result.report);
+    })
+    .catch(next);
+}
+
+function variance(current, previous) {
+  const difference = (previous - current);
+  return -1 * (difference / previous);
```
mbayopanda, line 118: Why not : 
```js
const difference = (current - previous);
return difference / previous;
```

server/controllers/finance/reports/income_expense/index.js
```diff
@@ -0,0 +1,274 @@
...
+    parentNode.isExpenseAccount = (node.isExpenseAccount || node.type_id === EXPENSE_TYPE_ID);
+  }, false);
+
+  // label depths of nodes
+  tree.walk(Tree.common.computeNodeDepth);
+
+  return tree;
+}
+
+const MAX_ITERATIONS = 25;
```
mbayopanda, line 246: Could comment about max iteration, because i am wondering why just 25

server/controllers/finance/reports/income_expense/report.handlebars
```diff
@@ -0,0 +1,83 @@
...
+  <!-- body -->
+  <div class="row">
+    <div class="col-xs-12">
+      <!-- page title  -->
+      <h3 class="text-center">
+        <strong>{{translate "REPORT.INCOME_EXPENSE"}}</strong>
+      </h3>
+
+      <h5 style="margin:15px;" class="text-center">
+        {{date periodFrom.start_date "MMMM"}} - {{date periodTo.end_date "MMMM"}}
```
mbayopanda, line 15: There no need for knowing the year ? because we display just months.

server/controllers/finance/reports/index.js
```diff
@@ -13,7 +13,7 @@ exports.vouchers = require('./vouchers');
 exports.cash = require('./cash');
 exports.cashflow = require('./cashflow');
 exports.patient = require('./financial.patient').report;
-exports.incomeExpense = require('./incomeExpense');
+exports.income_expense = require('./income_expense');
```
mbayopanda, line 16: We don't have a naming convention for file name `./income_expense`, but for method i think that it was camelCase convention.

server/lib/template/helpers/finance.js
```diff
@@ -61,7 +59,15 @@ function debcred(value = 0, currencyId) {
   `);
 }
 
+function percentage(value = 0, precision = 2) {
+  if (!value || !Number.isFinite(value) || Number.isNaN(value)) { return ''; }
+
+  const str = (value * 100).toFixed(precision);
```
mbayopanda, line 65: 👍

mbayopanda (approved):
LGTM 👍

client/src/js/components/bhFiscalPeriodSelect.js
jniles: I tried that, but eslint complains :(.  They call it "[no confusing arrow](https://eslint.org/docs/rules/no-confusing-arrow)".  I guess it makes the code look confusing?

server/controllers/finance/reports/income_expense/index.js
jniles: Yeah, that's much better.   I'll do that.

server/controllers/finance/reports/income_expense/report.handlebars
jniles: _Technically_ you shouldn't look at income/expense reports over more than one year.  They should be made 0 at the opening of the fiscal year.

The client side also doesn't let you use more than one fiscal year.

PR #2758, opened 2018-05-16
feat(voucher) link between voucher and stock movement

### LINKED ISSUE #2694: Stock Movement documents should have either a Voucher or Transaction Reference
Our stock movements link the stock movement document with an identifier `SM.${type_id}.${increment}`.  A typical document from a purchase order is shown below:
stockmovementdocument
_Fig 1: Stock Movement from Purchase Order_

This document has a lot of information on it.  The link from the voucher to the stock movement is good, however it would be nice to have links back to the voucher/transaction.  Since the movement isn't the voucher, we should include a reference the voucher for future reference.

jeremielodi:
closes #2694

### DIFF
--- a/server/models/procedures/stock.sql
+++ b/server/models/procedures/stock.sql
@@ -73,6 +73,7 @@ BEGIN
   
   -- variables
 	DECLARE v_stock_account INT(10);
+  DECLARE v_stock_movement_reference INT(10);
 	DECLARE v_cogs_account INT(10);
 	DECLARE v_unit_cost DECIMAL(19, 4);
 	DECLARE v_quantity INT(11);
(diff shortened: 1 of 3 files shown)


mbayopanda:
@jniles the issue was not about the _Stock Movement documents should have either a Voucher or Transaction Reference_ #2694 

Because this PR doesn't set either a Voucher or Transaction Reference in the stock movement document.

@jeremielodi could you discuss about that with @jniles

jniles:
@mbayopanda, thanks for the review.

Yes, the original issue describes putting the identifier of the voucher (VO.TPA.XXXX) on the stock transfer document.  It should only change the stock receipt handlebars file and maybe the stock receipt JS, but not the database.

PR #2766, opened 2018-05-22
refactor(vouchers): require transaction type

### LINKED ISSUE #2323: Make "Transaction Type" a required field
In order to facilitate reporting, I propose we make "transaction type" a required field.  Since we already have "generic income" , "generic expense" and "transfer", a user should be able to label any transaction with one of those.  However, if they wish to increase the resolution of their reporting, they will be able to create transaction types such as "Advances sur Salaires", "Rent Payments", etc, for common payment types.  These could then be used in reports such as cashflow.

This will introduce some useful guarantees on our reports - in particular, that every transaction can be grouped by it's transaction type.

jniles:
The behavior of both the transaction type inputs on the voucher modules have changed - we now require them explicitly.  This simplifies the code greatly and forces users to provide a designation for their transactions.

Closes #2323.

### DIFF
--- a/client/src/modules/vouchers/complex-voucher.html
+++ b/client/src/modules/vouchers/complex-voucher.html
@@ -99,20 +99,15 @@
               <div class="col-md-6">
 
                 <div class="form-group" ng-class="{ 'has-error' : ComplexVoucherForm.$submitted && ComplexVoucherForm.type_id.$invalid }">
-                  <label class="control-label">
-                    <span translate>FORM.LABELS.TRANSACTION_TYPE</span>
-                    <span uib-popover="{{ 'VOUCHERS.GLOBAL.TRANSACTION_TYPE_INFO' | translate }}"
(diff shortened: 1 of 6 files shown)


client/src/modules/vouchers/complex-voucher.html
```diff
@@ -99,20 +99,15 @@
...
+                  <label class="control-label" translate>
+                    FORM.LABELS.TRANSACTION_TYPE
                   </label>
 
                   <!-- Select a Transaction Type -->
                   <ui-select
                     name="type_id"
                     ng-model="ComplexVoucherCtrl.Voucher.details.type_id"
-                    ng-required="ComplexVoucherCtrl.Voucher.hasCashboxAccount"
+                    required="true"
```
mbayopanda, line 110: 👍

mbayopanda (approved):
LGTM 

bors r+

PR #2770, opened 2018-05-23
Refactor transaction types

### LINKED ISSUE #1054: (discussion): Group all transaction types under "income", "expense", or "custom"
The current behavior of the transaction types module is to allow a user to specify a custom name for the transaction type and a custom type.  I cannot see a good use for this behavior - as it is dynamic, we cannot easily build reports off of it.  Our "CREDIT NOTE" transaction type has the type `creditNote` ... and I assume all transaction types in the future will just duplicate their name for the "type" field.

We should have `income`, `expense`, and `custom` or `other` as our transaction classes and leave it at that.  @IMA-WorldHealth/bhima-core what do you think?

### LINKED ISSUE #1311: (refactor) Reconsider 'Transaction Types'
Currently most of our financial transactions have to be tagged with a transaction type. This defines a transactions purpose and marks it as one of the following 'income', 'expense' or 'other'. 

This system should be reconsidered looking at real use cases and reports. 

Corner cases to consider: 
* Transferring money between cash boxes 
* Transaction types use in report generation
* Complexity of selecting a transaction type on different modules - list of rules for where these are automatically set 
* An 'undefined' transaction type state - is this acceptable, what does it mean?

Additionally I propose that we consider transaction types as either of the following options:
* Income
* Expense 
* Transfer

jniles:
Transaction types have been simplified.  Only three different types are supported - income, expense and other.  This simplifies the explanations of how the module works.  The prefix has been removed since it was never used - we tend to put any text in our `i18n/*.json` files.  The server side has been updated to the latest standards of comments and ESLint.

Closes #1054.  Closes #1311.


client/src/modules/templates/modals/transactionType.modal.html
```diff
@@ -4,66 +4,48 @@
...
-      <select class="form-control"
+      <label class="control-label" translate>FORM.LABELS.TYPE</label>
+      <select
+        class="form-control"
         name="type"
-        ng-model="$ctrl.transactionType.type" required>
-        <option value="" disabled>-- {{ 'FORM.SELECT.TYPE' | translate }} --</option>
-        <option value="{{opt.value}}" ng-repeat="opt in $ctrl.optionType">{{opt.label | translate}}</option>
+        ng-model="$ctrl.transactionType.type"
+        ng-options="type.value as (type.label | translate) for type in $ctrl.types"
```
mbayopanda, line 32: 👍

client/src/modules/transaction-type/transaction-type.ctrl.js
```diff
@@ -55,54 +69,54 @@ function TransactionTypeController(TransactionType, TransactionTypeStore, Notify
...
+        list.forEach(type => {
+          type._popover = type.fixed ? $translate.instant('TRANSACTION_TYPE.FIXED_INFO') : '';
+        });
       })
       .catch(Notify.handleError);
   }
 
+  function startup() {
+    load();
+  }
```
mbayopanda, line 118: Why not to use just the  `load()` method ?

server/models/bhima.sql
```diff
@@ -146,24 +146,24 @@ INSERT INTO `fonction` VALUES
...
-  (14, 'VOUCHERS.SIMPLE.STOCK_ENTRY', 'stock reception', 'STOCK_ENTRY', 1),
-  (15, 'VOUCHERS.SIMPLE.COMMITMENT', 'other', 'COMMIT', 1),
-  (16, 'VOUCHERS.SIMPLE.EMPLOYEE_WITHHOLDINGS', 'other', 'WITHHOLDING', 1),
-  (17, 'VOUCHERS.SIMPLE.CHARGE_REMUNERATION', 'other', 'CHARGE_REMUNERATION', 1);
+INSERT INTO `transaction_type` (`id`, `text`, `type`, `fixed`) VALUES
+  (1, 'VOUCHERS.SIMPLE.GENERIC_INCOME', 'income', 1),
+  (2, 'VOUCHERS.SIMPLE.CASH_PAYMENT', 'income', 1),
+  (3, 'VOUCHERS.SIMPLE.CONVENTION_PAYMENT', 'income', 1),
+  (4, 'VOUCHERS.SIMPLE.SUPPORT_INCOME', 'income', 1),
+  (5, 'VOUCHERS.SIMPLE.TRANSFER', 'other',  1),
```
mbayopanda, line 154: 👍

mbayopanda (approved):
LGTM

bors r+

PR #2784, opened 2018-05-29
feat(voucher registry) link from cash payment registry

### LINKED ISSUE #2781: Voucher link from cash registry does not reset the search filters
When you click the "View Vouchers" link in the cash registry, it should remove all previous filtering on the voucher registry to show only the specific value the user is searching for.  [This can be seen on the "View Transactions" link](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/cash/payments/templates/action.cell.html#L42), but it should be applied to every single link in the actions registry.

jeremielodi:
closes #2781


jniles:
Looks good!  Could you apply this to every link in the dropdown menu?

Issue #2786, opened 2018-05-29
Issue related to Advanced Journal Vouchers


lomamech:
There is a problem with the add transaction with Advanced Journal Vouchers
 module, when using the Menu for posting entries, because Journal Vouchers

 module only has two default rows and when the menu used returns more than two rows l we notice that they miss the numbers of the accounts on these lines
![example2lines](https://user-images.githubusercontent.com/8035588/40655300-341645fa-6341-11e8-92f8-f404a52b1ab8.png)

In the picture above, we are not trying to support two employee invoices from a third party account

![example2lines](https://user-images.githubusercontent.com/8035588/40655422-96b19a7a-6341-11e8-8299-19fd67be0be5.png)
Interface preview with the missing account number on the third line

jniles:
Closed in #2794.

PR #2794, opened 2018-05-30
display correctly complex journal voucher

jeremielodi:
This PR attempts to solve #2786


jniles:
It would be ideal to find a better way to do this.  I'll merge this, but if you have free time, can you continue to research why ui-grid isn't rendering the right accounts?

PR #2797, opened 2018-05-30
feat(Aged Debtor Report) useMonthGrouping takes into account years

### LINKED ISSUE #2795: Aged Debtors Report "useMonthGrouping" does not take into account years
The group by months columns in the debtor groups is wrong.  Specifically [this line](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/controllers/finance/reports/debtors/index.js#L83) only uses `MONTH()`, not `MONTH()` and `YEAR()`.  This means that _all_ fiscal years are grouped if the date matches that month.  This bug should be fixed.

jeremielodi:
closes #2795

### DIFF
--- a/client/src/modules/reports/generate/aged_debtors/aged_debtors.config.js
+++ b/client/src/modules/reports/generate/aged_debtors/aged_debtors.config.js
@@ -25,6 +25,9 @@ function AgedDebtorsConfigController($sce, Notify, SavedReports, AppCache, repor
   };
 
   vm.onSelectPeriod = (period) => {
+    if (!period) {
+      return;
+    }
     vm.reportDetails.period_id = period.id;
   };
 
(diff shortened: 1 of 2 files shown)


server/controllers/finance/reports/debtors/index.js
```diff
@@ -94,7 +95,7 @@ function queryContext(queryParams) {
   `;
 
   const columns = useMonthGrouping ? groupByMonthColumns : groupByRangeColumns;
-
+  const filterByFiscalId = useMonthGrouping ? `AND gl.fiscal_year_id = ${fiscalId} ` : ``;
```
jniles: This should use `?` syntax to avoid SQL injection attacks.

client/src/modules/reports/generate/aged_debtors/aged_debtors.config.js
```diff
@@ -25,6 +25,9 @@ function AgedDebtorsConfigController($sce, Notify, SavedReports, AppCache, repor
   };
 
   vm.onSelectPeriod = (period) => {
+    if (!period) {
```
jniles, line 28: Does this ever fire without a period? It isn't supposed to :/

PR #2802, opened 2018-05-31
feat(balance report) use absolute values for credit

### LINKED ISSUE #2796: Use absolute value for credits in balance report
Technically, credits can be seen as negative numbers.  If we have two columns "debit" and "credit", there is no need to render the credits as negative - this is understood.  On the balance report (seen below), we render credits as negative values.  We should make them positive using absolute value before displaying them.
useabsolutevaluesforcredits
_Fig 1: Credits are negative_

jeremielodi:
closes #2796

### DIFF
--- a/server/controllers/finance/reports/balance/index.js
+++ b/server/controllers/finance/reports/balance/index.js
@@ -126,20 +126,33 @@ function getBalanceSummary(periodId, currencyId, shouldPrune) {
         });
       };
 
+      // specify debit and credit, make credit absolute
+      const makeBalanceKeysAbs = (currentNode) => {
+        balanceKeys.forEach(key => {
+          const amount = _.clone(currentNode[key]);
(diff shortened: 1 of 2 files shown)


server/controllers/finance/reports/balance/index.js
```diff
@@ -126,20 +126,33 @@ function getBalanceSummary(periodId, currencyId, shouldPrune) {
         });
       };
 
+      // specify debit and credit, make credit absolute
+      const makeBalanceKeysAbs = (currentNode) => {
+        balanceKeys.forEach(key => {
+          const amount = _.clone(currentNode[key]);
```
jniles, line 132: Why do you need to use `_.clone()`?

jniles (commented):
Looks fine to me.  Good work.

bors r+

PR #2804, opened 2018-05-31
refractor(profit lost statement) improvement, add fiscal label, totals

### LINKED ISSUE #2798: Improvements to the Profit and Loss Statement
The Profit and Loss Statement is already a pretty solid report.  The following improvements would be helpful:

 1. The report needs to have the Fiscal Year in the title (such as 2018, 2017, 2016, etc..).  Currently the only way to figure out the year is by looking at the columns.
 2. Each table - both income and expense - should include the total sum of their values at the bottom of the grid.
 3. The report needs a final section that compares the total profit and the total loss and gives a net result.

jeremielodi:
closes #2798

### DIFF
--- a/server/controllers/finance/reports/income_expense/index.js
+++ b/server/controllers/finance/reports/income_expense/index.js
@@ -79,8 +79,8 @@ function document(req, res, next) {
 
       const isIncomeFirstElement = root.children[0].isIncomeAccount;
 
-      let income;
-      let expense;
+      let income = {};
+      let expense = {};
       if (isIncomeFirstElement) {
         [income, expense] = root.children;
       } else {
@@ -91,8 +91,14 @@ function document(req, res, next) {
       const losses = [];
 
       tree.walk(node => profits.push(node), true, income);
-
       tree.walk(node => losses.push(node), true, expense);
+
+      // calculate totals and profit
+      const emptyTotal = { balance : 0, previousBalance : 0, difference : 0 };
+      const totals = {
+        income :  profits[0] || emptyTotal,
+        expense : losses[0] || emptyTotal,
+      };
 
       // computes the variance on the income/expense
       profits.forEach(account => {
@@ -103,7 +109,9 @@ function document(req, res, next) {
         account.variance = variance(account.balance, account.previousBalance);
       });
 
-      _.extend(data, { profits, losses, previousFiscalYear });
+      _.extend(data, {
+        profits, losses, previousFiscalYear, totals,
+      });
 
       return report.render(data);
     })
--- a/server/controllers/finance/reports/income_expense/report.handlebars
+++ b/server/controllers/finance/reports/income_expense/report.handlebars
@@ -12,7 +12,7 @@
       </h3>
 
       <h5 style="margin:15px;" class="text-center">
-        {{date periodFrom.start_date "MMMM"}} - {{date periodTo.end_date "MMMM"}}
+        {{date periodFrom.start_date "MMMM"}} - {{date periodTo.end_date "MMMM"}} ({{fiscalYear.label }})
       </h5>
 
       {{#if profits}}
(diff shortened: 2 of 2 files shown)


server/controllers/finance/reports/income_expense/report.handlebars
```diff
@@ -75,8 +82,34 @@
...
         </tbody>
       </table>
+
+      <br/>
+      <table style="width:400px;background-color:#ddd;" class="pull-right table table-condensed table-report">
+        <tr>
+          <th colspan="2" class="text-center"><b>{{translate 'FORM.LABELS.PROFIT'}}</b></th>
+        </tr>
+        <tr>
+          <td>{{translate 'FORM.LABELS.EXPENSE'}}</td>
```
jniles: Since the report has "income" then "expense", these totals should have "income" then "expense".

server/controllers/finance/reports/income_expense/report.handlebars
```diff
@@ -75,8 +82,34 @@
...
+          <td class="text-right">{{debcred totals.expense.balance ../metadata.enterprise.currency_id}}</td>
+        </tr>
+        <tr>
+          <td>{{translate 'FORM.LABELS.INCOME'}}</td>
+          <td class="text-right">{{debcred totals.income.balance ../metadata.enterprise.currency_id}}</td>
+        </tr>
+        <tr>
+          <td>{{translate 'FORM.LABELS.BALANCE'}}</td>
+          <td class="text-right">{{debcred totals.profits ../metadata.enterprise.currency_id}}</td>
```
jniles: I think all these `../metadata.enterprise.currency_id` should just be `metadata.enterprise.currency_id`.  Have you tried this report with multiple enterprise currencies?  Do they change?

server/controllers/finance/reports/income_expense/report.handlebars
```diff
@@ -41,6 +41,13 @@
               <td class="text-right">{{percentage account.variance}}</td>
             </tr>
           {{/each}}
+          <tr>
+            <td colspan="2" class="text-right"><b>{{translate 'FORM.LABELS.TOTAL'}}</b></td>
+            <td class="text-right">{{debcred totals.income.previousBalance ../metadata.enterprise.currency_id}}</td>
```
jniles: Same comment about enterprise currency.  I think it should be just `metadata.enterprise.currency_id`.  The reason the others are `../metadata` is because they are in an `#each` loop.

server/controllers/finance/reports/income_expense/index.js
```diff
@@ -89,10 +89,29 @@ function document(req, res, next) {
...
       const profits = [];
       const losses = [];
+      const totals = {
+        income : { balance : 0, previousBalance : 0 },
+        expense : { balance : 0, previousBalance : 0 },
+      };
 
       tree.walk(node => profits.push(node), true, income);
+      // sum income
+      tree.walk(node => {
```
jniles: This code is correct, but the solution is even easier!  The `income` here is the root of the `income` account tree, and `expense` is the root of the expense tree.  Every node in the tree has the following properties: `balance`, `previousBalance`, `difference`, `variance`.  They are computed [here](
https://github.com/IMA-WorldHealth/bhima-2.X/pull/2804/files#diff-950ab90f8fd78e6083caa9787d57fac4L219).

So the total of all income and all expense accounts are already calculated!  You can just grab them from the root node of each tree:
```js
const totals = {
  income: { balance : income.balance, previousBalance: income.previousBalance, difference: income.difference },
  expense : { balance : expense.balance, expense.previousBalance, difference: expense.difference }
};
```

server/controllers/finance/reports/income_expense/index.js
```diff
@@ -89,10 +89,29 @@ function document(req, res, next) {
...
+      // sum expences
+      tree.walk(node => {
+        totals.expense.balance += Math.abs(node.balance);
+        totals.expense.previousBalance += Math.abs(node.previousBalance);
+      }, true, expense);
+
+      // calculate totals and profit
+      totals.income.difference = Math.abs(totals.income.balance - totals.income.previousBalance);
+      totals.expense.difference = Math.abs(totals.expense.balance - totals.expense.previousBalance);
+      totals.profits = Math.abs(totals.expense.balance - totals.income.balance);
```
jniles: A better name for this is `totals.net` or `totals.result` because it could be negative.  You usually only call positive numbers "profit" and negative numbers "loss" (which is why this report is called Profit & Loss).

server/controllers/finance/reports/income_expense/index.js
```diff
@@ -103,7 +122,9 @@ function document(req, res, next) {
         account.variance = variance(account.balance, account.previousBalance);
       });
 
-      _.extend(data, { profits, losses, previousFiscalYear });
+      _.extend(data, {
+        profits, losses, previousFiscalYear, totals,
```
jniles, line 113: 👍

server/controllers/finance/reports/income_expense/report.handlebars
```diff
@@ -12,7 +12,7 @@
       </h3>
 
       <h5 style="margin:15px;" class="text-center">
-        {{date periodFrom.start_date "MMMM"}} - {{date periodTo.end_date "MMMM"}}
+        {{date periodFrom.start_date "MMMM"}} - {{date periodTo.end_date "MMMM"}} ({{fiscalYear.label }})
```
jniles, line 15: 💃

jniles (changes requested):
This is the right idea.  Three points of feedback:
 1. Income must come before expense in the footer.  An accountant would require that.
 2. The currency id probably doesn't work in another enterprise currency.  Try setting the enterprise currency to FC in the database and see if your totals show up in FC or USD.
 3. The totals are 100% correct.  However, you can reduce the amount of processing and the amount of code by using the pre-computed totals from the tree.

server/controllers/finance/reports/income_expense/index.js
jeremielodi: ```income``` is a tree, so ```income.balance``` is not the total of all nodes

jniles (approved):
Great work!

bors r+

PR #2806, opened 2018-06-01
fix(reports): misc fixes

jniles:
HACK.  This commit fixes the aged reports by making them look up the final date in the period provided.  In the future, this should filter on period ids instead of on dates.


mbayopanda (approved):
👍 LGTM

PR #2809, opened 2018-06-01
Add employee on entity map

### LINKED ISSUE #2790: Difficult to know if a transaction is linked to a debtor or creditor
With the Journal Voucher module, it is possible to associate a debtor or creditor with a transaction, but it is impossible to determine via the Posting Journal or even the Voucher, transactions related to a debtor or creditor, it is for this I propose to add as possible in the description of the transaction this information

lomamech:
- Add the Path of EmployeeStanding Report in config/identifiers
- Update the API for Employee Allow the Possibility to set creditor and
  debitor uuid from Client
- Update referenceLookup track when the code is for Employee for redirect
  to Employee Standing Reports
- Add New Trigger to set employee on entity_map table
- Update Data fore database of test

test(Employee Reference)  …
- Implement E2E test for Research Reference Of Employee
- Update Api for Employee And Filtering of reference of Employee

Closes #2790


client/src/i18n/en/table.json
```diff
@@ -23,7 +23,7 @@
             "CANCEL_CASH": "Cancel Cash Payment",
             "CHARGE": "Charge",
             "CLIENT": "Client",
-            "CODE": "Code",
+            "CODE": "Registration number",
```
jniles: It looks like this is used in other modules (like [stock](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/stock/entry/entry.js#L72)) for the word "Code".  Could you make a new key for this translation?

client/src/modules/employees/registry/registry.js
```diff
@@ -38,6 +38,12 @@ function EmployeeRegistryController(
       aggregationHideLabel : true, headerCellFilter     : 'translate',
       footerCellClass      : 'text-center',
     },
+    { field                : 'reference',
+      displayName          : 'TABLE.COLUMNS.REFERENCE',
+      aggregationType      : uiGridConstants.aggregationTypes.count,
+      aggregationHideLabel : true, headerCellFilter     : 'translate',
```
jniles: This should be on two lines:
```js
aggregationHideLabel: true,
```

server/config/identifiers.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   EMPLOYEE : {
     key   : 'EM',
     table : 'employee',
+    documentPath : '/reports/finance/employeeStanding/',
```
jniles, line 23: Nice! 👍

test/data.sql
```diff
@@ -615,9 +615,9 @@ INSERT INTO `creditor_group` VALUES
 
 -- Creditors
 INSERT INTO `creditor` VALUES
-  (HUID('42d3756a-7770-4bb8-a899-7953cd859892'), HUID('b0fa5ed2-04f9-4cb3-92f7-61d6404696e7'), 'Personnel'),
+  (HUID('42d3756a-7770-4bb8-a899-7953cd859892'), HUID('b0fa5ed2-04f9-4cb3-92f7-61d6404696e7'), 'Crediteur[Test 2 Patient]'),
```
jniles, line 618: Did these need to change...?

test/end-to-end/employees/registry.spec.js
```diff
@@ -57,6 +58,14 @@ describe('Employees Registry', () => {
     employeeRegistryPage.clearFilter();
   });
 
+  it(`should find One employee With reference "${parameters.reference}"`, () => {
```
jniles, line 61: Nice 👍

jniles (commented):
Looks pretty good to me!

It looks like multiple other modules use `CODE` for translation where the new text wouldn't work.  Could you make a new key?

test/data.sql
lomamech: Just to respect the formatting of the generate text when creating a creditor

jniles: Ah, okay.  Makes sense.

jniles (approved):
LGTM.  All my remarks were addressed.

bors r+

PR #2810, opened 2018-06-01
feat(bhAccountSelect): filter out locked accounts

### LINKED ISSUE #2808: Add "hidden" and "locked" options to accounts
~~We haven't really implemented https://github.com/IMA-WorldHealth/bhima-2.X/issues/490, but we can still use the column "locked" in the database to filter out locked accounts.  Eventually, we'll need to implement the ability to lock accounts properly, but for the moment, the only way to properly use IMCK's accounts is to lock the unused ones from their pre-ohada days.~~

---

**Design**
This proposal supersedes #490 and the above suggestion.  Instead, it separates two ideas (based on @lomamech's feedback on #2810): _hidden_ accounts and _locked_ accounts.  The difference is this:

  1. _Hidden_ accounts are accounts that will not show up in any account dropdown.  The only place where hidden accounts will be displayed is the accounts module, to allow an administrator to know they are there.  (icon: `fa-eye-slash` if hidden)
 2. _Locked_ accounts are visible in all account dropdowns.  However, they cannot be posted to the GL, since the Trial Balance will block them. (icon: `fa-locked` if locked)

If a user wants to make an account completely unavailable (such as the pre-OHADA accounts described above) they should both lock the account and hide the account.

In order to accomplish these complex options, I propose that we make an actions dropdown on the account module.  It will have the following options:
 1. **Actions > Edit Account** - brings up the edit modal as the current link does.
 2. **Actions > Unlock/Lock Account** - brings up a modal to specifically lock or unlock the account depending on the account's current state.  It will contain an explanation of what locking an account does if the user is trying to lock the account.
 3. **Actions > Hide/Unhide Account** - brings up a modal to specifically hide or unhide the account depending on the current state.
 3. **Actions > Delete Account** - allows the user to delete the account if possible.

jniles:
This commit implements the locked account filter on the bhAccountSelect. It also cleans up the report controller and removes the `classe` variable since it isn't used in any modules that I can find.

Closes #2808.


lomamech:
This PR managing the display on behalf of accounts, blocked, but the account management module does not give the possibility to lock or unlock an account, it should also allow to hide the accounts (Invalides) which should no longer be displayed to users

jniles:
Thanks for the review @lomamech!

As @lomamech points out, "locked" and "hidden" should probably be different options.  In this PR, I made them the same thing.  I'll update this PR with both options to allow administrators to still view reports on "locked" accounts, but have any number of "hidden" accounts that will not show up in the dropdown.

client/src/i18n/en/account.json
```diff
@@ -47,6 +47,10 @@
     "SUBMIT_DELETE":"Delete Account",
     "SUBMIT_UPDATE":"Update Account",
     "TYPE_CHANGE_BLOCKED":"You cannot change the type of a title account with children.",
+    "CONFIRM_HIDE" : "Hiding an account removes the account from all account selection elements.  It will not affect any pending transactions using the account.  Make it hidden?",
+    "CONFIRM_UNHIDE" : "This account is currently hidden. Are you sure you want to reveal it for use?",
+    "CONFIRM_LOCK" : "Locking an account restrict the ability to post transactions using the account.  Are you sure you want to lock it?",
+    "CONFIRM_UNLOCK" : "The account is locked.  Unlocking it will allow transactions to be posted containing this account.  Are you sure you want to unlock it?",
```
mbayopanda, line 53: There are missing french translations.
```js
"CONFIRM_HIDE" : "Masquer un compte supprime le compte de tous les éléments de sélection de compte. Cela n'affectera pas les transactions en attente utilisant le compte. Le rendre caché ?",
"CONFIRM_UNHIDE" : "Ce compte est actuellement caché. Êtes-vous sûr de vouloir le révéler pour l'utiliser ?",
"CONFIRM_LOCK" : "Le verrouillage d'un compte limite la possibilité de publier des transactions en utilisant le compte. Êtes-vous sûr de vouloir le verrouiller ?",
"CONFIRM_UNLOCK" : "Le compte est bloqué. Le débloquer permettra d'afficher des transactions contenant ce compte. Êtes-vous sûr de vouloir le débloquer ?",
```

client/src/modules/accounts/templates/grid.actionsCell.tmpl.html
```diff
@@ -1,8 +1,64 @@
-<div class="ui-grid-cell-contents">
-  <a
-    id="account-edit-{{row.entity.id}}"
-    ui-sref="accounts.edit({id:row.entity.id})">
-    <i class="fa fa-edit"></i>
-    <span translate>TABLE.COLUMNS.EDIT</span> {{row.entity.number}}
+<div class="ui-grid-cell-contents" uib-dropdown dropdown-append-to-body data-action="{{ row.entity.id }}">
+  <a href uib-dropdown-toggle>
```
mbayopanda, line 2: 👍 nice

server/controllers/finance/accounts/index.js
```diff
@@ -32,11 +32,10 @@ const types = require('./types');
 const categories = require('./categories');
 const Periods = require('../../../lib/period');
 const AccountExtras = require('./extra.js');
```
mbayopanda: `./extra.js` could be also `./extra`

mbayopanda (approved):
LGTM

server/controllers/finance/accounts/index.js
jniles: Will do.

client/src/i18n/en/account.json
jniles: Ah, beautiful.  This is exactly what I needed!

jniles:
@mbayopanda, changes landed!

jniles:
Seems like node 10 + lzma-native are having a disagreement on Travis.  I may need to revert to an older version of node 10.

PR #2816, opened 2018-06-04
improve(Patient Support)

### LINKED ISSUE #2789: Improving the Patient Support Interface
- It is important to allow the selection of an employee, when using the module Support, I propose to add a radio button that will display according to the choice the employee who support patient

- I also propose to modify the description message in the case of supported, to use Support instead of transfere

lomamech:
- Add Possibility to select an Employee or Not duiring a Patient Support
  Interface

Closes #2789


jniles (approved):
LGTM.  Great refactoring!

bors r+

PR #2819, opened 2018-06-07
fix(caution): rename variable

jniles:
There was an old/unknown variable named "scost" that blocks payments if you have a caution.  It has been renamed.


jniles:
This is a bug.  No review needed.

bors r+

PR #2831, opened 2018-06-09
Toggle title account in balance report

### LINKED ISSUE #2828: feature(balance report): option for hiding title accounts
This issue is about to have an option in the balance report form for hiding title accounts if needed.

mbayopanda:
This PR add an option for showing or hiding title accounts in the balance report module.

Close #2828


server/controllers/finance/reports/balance/index.js
```diff
@@ -86,16 +90,27 @@ function getBalanceSummary(periodId, currencyId, shouldPrune) {
     SELECT a.id, a.number, a.label, a.type_id, a.label, a.parent,
       a.type_id = ${TITLE_ID} AS isTitleAccount,
       IFNULL(s.before, 0) AS "before",
+      IFNULL(s.before_debit, 0) AS "before_debit",
```
jniles, line 93: Beautiful!

server/controllers/finance/reports/balance/report.handlebars
```diff
@@ -81,31 +81,35 @@
           {{#each accounts as |account|}}
             <tr class="text-right" {{#if account.isTitleAccount}}style="font-weight:bold;"{{/if}}>
               <td class="text-left">
-                <span style="padding-left: calc(10px * {{account.depth}})">{{ account.number }} - {{ account.label }}</span>
+                <span {{#unless ../shouldHideTitleAccounts}}style="padding-left: calc(10px * {{account.depth}})"{{/unless}}>{{ account.number }} - {{ account.label }}</span>
```
jniles, line 84: 💃

server/controllers/finance/reports/balance/report.handlebars
```diff
@@ -81,31 +81,35 @@
...
+                {{/gt}}
+
+                {{!-- creditor sold --}}
+                {{#lt account.before.amount 0}}
                   <td></td>
                   <td>{{currency account.before.credit account.currencyId }}</td>
-                  {{else}}
+                {{/lt}}
+
+                {{!-- void cells in other case --}}
```
jniles, line 100: Great comments!

server/controllers/finance/reports/balance/index.js
```diff
@@ -58,6 +59,9 @@ function document(req, res, next) {
 
   getBalanceSummary(params.period_id, req.session.enterprise.currency_id, context.shouldPruneEmptyRows)
     .then(data => {
+      if (context.shouldHideTitleAccounts) {
+        data.accounts = data.accounts.filter(account => account.isTitleAccount === 0);
```
jniles, line 63: This is the easiest way of doing it, but it is also the slowest.  Ideally, it would be nice to filter the title accounts in SQL to make things faster.

jniles (approved):
The logic LGTM.  Looks like the tests need to be appeased and then we can merge.

server/controllers/finance/reports/balance/index.js
mbayopanda: I am trying to use any condition like `WHERE a.type_id <> 6` in the sql inside the `getBalanceSummary` function but there is nothing as returned value.

jniles: It's probably because that gets passed to the tree next.  It would need to skip all the `Tree()` functions as well (since if there are not parents, there isn't a tree!).  We can work this out later :+1:

jniles:
Looks like a fluke.

bors r+

PR #2847, opened 2018-06-12
fix(invoice): order invoice items by code

jniles:
This commit orders the invoice items by code to make them correspond better to the 1.x invoices.


jniles:
:(

bors r+

PR #2862, opened 2018-06-14
bug(Journal Transanction Type)

### LINKED ISSUE #2851: Journal "transaction type" column should contain the transaction type, not the ID
The Journal's "transaction type" column contains the ID of the transaction type, not the text of the transaction type.  The user has no idea what "10" is, but they do know what "transfer" is.  I propose that we make the transaction type column be filled with the transaction type, instead of the id.

lomamech:
- Set transaction type text instead transaction type Id

Closes #2851

### DIFF
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -294,7 +294,7 @@ function JournalController(
     visible : true,
   }, {
     field : 'transaction_type_text',
-    displayName : 'FORM.LABELS.TRANSACTION_TYPE',  
+    displayName : 'FORM.LABELS.TRANSACTION_TYPE',
     headerCellFilter : 'translate',
     cellTemplate :
       `<div class="ui-grid-cell-contents" translate>
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -293,9 +293,13 @@ function JournalController(
     headerCellFilter : 'translate',
     visible : true,
   }, {
-    field : 'transaction_type_id',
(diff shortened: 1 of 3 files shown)


client/src/modules/journal/journal.js
```diff
@@ -293,9 +293,13 @@ function JournalController(
     headerCellFilter : 'translate',
     visible : true,
   }, {
-    field : 'transaction_type_id',
+    field : 'transaction_type_text',
     displayName : 'FORM.LABELS.TRANSACTION_TYPE',
     headerCellFilter : 'translate',
+    cellTemplate :
```
jniles, line 299: This works - a better solution is to use the translate filter:
```js
cellFilter : 'translate',
```

This will avoid creating a new template.

server/controllers/finance/journal/index.js
```diff
@@ -120,12 +120,13 @@ function buildTransactionQuery(options, posted) {
       BUID(p.entity_uuid) AS entity_uuid, em.text AS hrEntity,
       BUID(p.reference_uuid) AS reference_uuid, dm2.text AS hrReference,
       p.comment, p.transaction_type_id, p.user_id, p.cc_id, p.pc_id, pro.abbr,
-      pro.name AS project_name, per.start_date AS period_start,
+      pro.name AS project_name, per.start_date AS period_start, tp.text AS transaction_type_text,
```
jniles, line 123: 👍

jniles (approved):
LGTM.  Solves the issue!

bors r+

PR #2863, opened 2018-06-14
feat(role management) alphabetic order role's list

### LINKED ISSUE #2856: Do not restrict roles to `project_id`
There doesn't seem to be a good reason to restrict roles to a particular project.  Project permissions are set already on a per-user basis.  Roles are meant to be re-used across the enterprise, so we don't want to have the same role twice for two different projects - it should just be a single role and the users have restricted access.

jeremielodi:
This PR sorts the roles's list.

Closes #2856

### DIFF
--- a/client/src/modules/roles/modal/rolesPermissions.html
+++ b/client/src/modules/roles/modal/rolesPermissions.html
@@ -32,7 +32,7 @@ novalidate>
           </h4>
 
           <ul class="list-group">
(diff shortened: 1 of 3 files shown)


jniles (approved):
LGTM.

bors r+

jniles:
Note: this will also close #2856

jniles:
@jeremielodi it looks like there are some errors here.  Can you take a look?

Issue #2805, opened 2018-06-01
Cashflow by Service report should filter by cashbox


jniles:
The Cashflow by Service ("Journal De Ventilation") report does not give the user the possibility to filter by cashbox.  In an institution with multiple cashboxes (such as IMCK), this doesn't make any sense.  See below:

![nofilterbycashbox](https://user-images.githubusercontent.com/896472/40830776-2ed69c76-657f-11e8-9ed0-25e2a764a5b4.PNG)
_Fig 1: Current options on Cashflow By Service_

This should be corrected by adding the `<bh-cashbox-select>` component to the page.

jniles:
Note: this report should also present the option of showing values in all currencies supported by the enterprise, at the exchange rate of the day.

jniles:
@mbayopanda, since the team here seems to be really interested in the "Ventillation" report, could you look at making these changes?  That way we can deploy it and test it pretty quickly.

jniles:
I'll grab this since I have more free time.

sfount:
Closed by #2866

PR #2866, opened 2018-06-14
feat(cashflowByService): filter by cashbox

### LINKED ISSUE #2807: Budget vs Balance Report
An important report for managers and financial analysts is the budget versus balance report.  Note that the report depends upon the implementation of #267.

### Budget versus Balance
The budget versus balance report is very similar to the Profit and Loss Statement.  It concerns only income and expense accounts.  The idea is to compare the income and expense accounts against what the budget has for the income and expense accounts.  Importantly, the report should show the variance between what has been budgeted and what was actually received.

jniles:
This commit implements cashbox filtering on the cashflow by service
report.  To accomadate this, the bh-cashbox-select component has been
refreshed to send the user_id if the component is restricted to the
user.

Closes #2807.

### DIFF
--- a/client/src/js/components/bhCashboxSelect.js
+++ b/client/src/js/components/bhCashboxSelect.js
@@ -45,7 +45,7 @@ function CashboxSelectController(Cashbox, Notify, Session) {
     Cashbox.read(null, params)
       .then(cashboxes => {
         cashboxes.forEach(cashbox => {
-          cashbox.hrlabel = `${cashbox.label}  ${cashbox.symbol}`;
+          cashbox.hrlabel = `${cashbox.label} ${cashbox.symbol}`;
         });
 
         $ctrl.cashboxes = cashboxes;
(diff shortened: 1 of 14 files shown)


lomamech:
Hi jonathan, I just tested this module and I have only two comments, 
- It would be necessary to add on the head of the report the name of cashbox the selected 

cashflowservice

- In Cash Flow Physical Report it is possible to select more cashboxes at the same time for the report, consider that for Cashbox By Service you can only use one Cashbox at a time.

jniles:
@lomamech, thanks for the review!  I will make those changes and let you know when they are done.

jniles:
@lomamech, the Cashflow by Service is supposed to model the ventillation which only looks at a single cashbox.  Does IMCK want us to try and support multiple cashboxes?

jniles:
Changes have landed!  Here's what it looks like:
newcashflowbyservicepresentation
_Fig 1: New presentation of Cashflow By Service_

lomamech:
@jniles  Travis CI build failed

lomamech:
@jniles I try again but The Travis CI Buil Failed again, Try to discover the problem

sfount:
I'm not 100% but it looks like this PR accidentally references and closes #2807 instead of #2805. I will update these issues accordingly, please revert this if this was intended.

PR #2868, opened 2018-06-14
feat(stock registries) download as excel

jeremielodi:
all stock registries can now be downloaded as excel


jniles (approved):
LGTM.

bors r+

PR #2871, opened 2018-06-15
chore: remove Modal.confirm() from Credit Note

### LINKED ISSUE #2836: Remove the "Modal.confirm()" call on credit note creation
Since we already have a modal for it, we don't need to make a second confirmation modal before the detailed one.  The detailed modal requires the user give an explanation, so they will not do this by accident.

jniles:
This commit removes the Modal.confirm() call, instead putting a danger text on the credit note page.

newpresentationofcreditnote

Closes #2836

### DIFF
--- a/client/src/modules/invoices/registry/modalCreditNote.html
+++ b/client/src/modules/invoices/registry/modalCreditNote.html
@@ -7,6 +7,11 @@
   </div>
 
   <div class="modal-body">
+    <div class="alert alert-danger" ng-if="ModalCtrl.hasPaymentsAgainstInvoice">
+      <i class="fa fa-exclamation-triangle"></i>
+      <span translate>FORM.DIALOGS.CONFIRM_CREDIT_NOTE</span>
+    </div>
+
     <div class="alert alert-warning">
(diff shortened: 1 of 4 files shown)


jniles:
@mbayopanda , one way we might avoid the problem in https://github.com/IMA-WorldHealth/bhima-2.X/pull/2923 is by putting this message up for caution links too.  Should I do that?

jniles:
@mbayopanda, could I get a review?

client/src/modules/invoices/registry/modalCreditNote.html
```diff
@@ -7,6 +7,11 @@
   </div>
 
   <div class="modal-body">
+    <div class="alert alert-danger" ng-if="ModalCtrl.hasPaymentsAgainstInvoice">
```
mbayopanda, line 10: :+1

mbayopanda (approved):
There is no modal, when me make a credit note for an invoice.

LGTM 👍 

bors r+

mbayopanda:
@jniles the modal doesn't care about the payment via caution (caution link)

PR #2872, opened 2018-06-15
Filter out caution links from patient activity

### LINKED ISSUE #2830: Create an "adjustment" transaction type
For the folks at IMCK, a lot of their writings are using "accounting" or "journal" transaction types.  These are for simple adjustments that don't quite make sense to put anywhere else (e.g. they are correcting errors).  I propose we make a new transaction type called "adjustment" to assign these transactions.

### LINKED ISSUE #2852: Filter out caution links from patient financial activity
The Patient Financial Activity report provides an overview of the invoices made to the patient, the payments and transfers against their account.  Unfortunately, our design of caution payments requires that we "link" the invoice to the payment via a dummy transaction which debit and credits that patient's account for a total movement value of 0.00.  This shows up in the Patient Financial Activity as $0.00, which is kind of confusing.  Even more confusing, those values are added up in the bottom total as "Total Billed" and "Total paid".  This is double-counting.

We need to make the following changes:
 1. Create a custom transaction type for caution payment links.  We'll call it "Link to Caution".
  2. Filter out the "Link to Caution" transaction type from the debtor financial history.

This will give us the correct balances, as well as allow us to tell the financial story of caution links.

jniles:
This commit adds new voucher types for adjustments and links to
cautions.

Closes #2830.

This fixes a critical bug in the cash payments where the receipt was not
rendered correctly in a different project.

This commit filters the caution link from the patient's financial
activity report so that the payments and invoices balance.  This removes
the confusion of having a $0 transaction in the mix.

Closes #2852


jniles:
@mbayopanda, this PR fixes the cash payments bug.

server/controllers/admin/enterprises.js
```diff
@@ -117,7 +117,7 @@ function lookupByProjectId(id) {
       JOIN village ON e.location_id = village.uuid
       JOIN sector ON village.sector_uuid = sector.uuid
       JOIN province ON sector.province_uuid = province.uuid
-    WHERE p.enterprise_id = ?
+    WHERE p.id = ?
```
mbayopanda, line 120: Great :+1

server/models/procedures/invoicing.sql
```diff
@@ -631,10 +631,9 @@ BEGIN
     SELECT payment.uuid, payment.balance FROM stage_payment_balances AS payment;
   DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;
 
-  -- This is marked as a "transfer" transaction type.
   -- TODO(@jniles): propose issue for how to deal with this kind of transaction
   -- type that is automatically generated by the system.
-  SET linkTransactionTypeId = 5;
+  SET linkTransactionTypeId = 19;
```
mbayopanda, line 636: What is meaning of these numbers ? 19 instead of 5

mbayopanda (approved):
LGTM 👍 

bors r+

server/models/procedures/invoicing.sql
jniles: Ah, it's the new transaction type ID for caution links.  I supposed I should have named the variable "cautionLinkTransactionTypeId", since "link" could be anything... :/

PR #2875, opened 2018-06-15
Fix stock and purchase module

### LINKED ISSUE #2874: bug(purchase_registry): the document button under the action link doesn't work
This issue is about the document link which doesn't work into the purchase registry : 

document link

mbayopanda:
This PR fixes :
- purchase registry by fixing the document link (#2874)
- stock entry from purchase editable cost and quantity
- stock entry description for purchase
- stock entry loading button
- stock entry calculation of purchaseStatus
- stock entry balance of entered inventories in the grid

Close #2874


client/src/i18n/en/stock.json
```diff
@@ -81,7 +81,7 @@
     "ORIGINS"         : "Lots Origins",
     "ORIGIN_DOCUMENT" : "Origin Document",
     "PATIENT_DISTRIBUTION" : "Patient Distribution",
-    "PURCHASE_DESCRIPTION" : "Purchaser order description",
+    "PURCHASE_DESCRIPTION" : "Stock from a purchase from {{supplier}} ({{purchase_order}})",
```
jniles, line 84: :+1:

client/src/modules/purchases/list/list.js
```diff
@@ -103,6 +104,10 @@ function PurchaseListController(
     Notify.handleError(error);
   }
 
+  vm.getDocument = (uuid) => {
```
jniles, line 107: 💃

client/src/modules/reports/generate/inventory_report/inventory_report.config.js
```diff
@@ -31,6 +31,10 @@ function InventoryReportConfigController(
     vm.inventory = inventory;
   };
 
+  vm.onSelectDate = (date) => {
```
jniles, line 34: 💃

client/src/modules/stock/entry/entry.js
```diff
@@ -341,8 +359,19 @@ function StockEntryController(
...
+      .then(toggleLoadingIndicator)
+      .finally(forceReload);
+  }
+
+  function toggleLoadingIndicator() {
+    vm.$loading = !vm.$loading;
+  }
+
+  function forceReload() {
+    $state.reload();
```
jniles: Oof.  Why do we need to force reload?  Why not just `form.$setPristine()` and delete the state data?

server/controllers/finance/purchases.js
```diff
@@ -361,58 +362,82 @@ function find(options) {
...
-    SELECT IFNULL(SUM(m.quantity * m.unit_cost), 0) AS movement_cost, p.cost, CEIL(DATEDIFF(m.date, p.date)) AS delay
+
+  const sqlQuantities = `
+    SELECT SUM(t.purchase_quantity) AS purchase_quantity, SUM(t.movement_quantity) AS movement_quantity
+    FROM (
+      (
+        SELECT IFNULL(SUM(pi.quantity), 0) AS purchase_quantity, 0 AS movement_quantity FROM purchase p
+        JOIN purchase_item pi ON pi.purchase_uuid = p.uuid
+        WHERE p.uuid = ?
+      ) UNION (
```
jniles: Just to be safe, this should probably be `UNION ALL`.  It might remove (duplicate) rows from the plain `UNION` statement.

server/controllers/finance/purchases.js
```diff
@@ -361,58 +362,82 @@ function find(options) {
...
+    WHERE purchase_item.purchase_uuid = ?
+  `;
+
+  const dbPromise = [
+    db.one(sqlQuantities, [purchaseUuid, purchaseUuid, FROM_PURCHASE_ID]),
+    db.one(sqlPurchaseDelay, [purchaseUuid, FROM_PURCHASE_ID]),
+    db.exec(sqlPurchaseInventories, [purchaseUuid]),
+  ];
+
+  q.all(dbPromise)
```
jniles, line 411: Good performance improvement 👍

server/controllers/finance/purchases.js
```diff
@@ -361,58 +362,82 @@ function find(options) {
...
+    .spread((resultQuantities, resultDelay, resultPurchaseInventories) => {
+      const rq = resultQuantities;
+      const rd = resultDelay;
+
+      glb.delay = Math.round((rd.delay / 30) * 10) / 10;
```
jniles, line 416: Where did these magic numbers come from?  Are you converting months -> days?  Do you really need the ` * 10) / 10`?

Could you put a comment above this line explaining it?

jniles (commented):
@mbayopanda nice bug fixes in here!

I'm concerned about the `$state.reload()` call - it seems like a really lazy solution because the module is either too complex to be reset or there isn't a way to clear the $state data without refreshing the browser.  Ideally, it should just remove the values that need to be reset to their default values.  For example, [here is the reset code for the patientInvoice.js controller](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/invoices/patientInvoice.js#L176).  Cash payments is even simpler - [just three lines](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/cash/cash.js#L162).  Could you see if we can adapt the stock controllers to this kind of reset without doing a redesign?

In the long run, we will probably want an SQL stored procedure to calculate the delay on purchase.  I can imagine bugs where the delay is not recalculated and the user/developers want to force a recalculation.  I'll make an issue about it.

Moving from `db.exec()` to `q.all([ /* ... */ ])` is a cool performance improvement!  Well done!

server/controllers/finance/purchases.js
mbayopanda: They are magic !!!

client/src/modules/stock/entry/entry.html
```diff
@@ -85,7 +86,7 @@
           </p>
 
           <!-- "Add number of grid rows" input-group -->
-          <bh-add-item disable="StockCtrl.entryOption" callback="StockCtrl.addItems(numItem)">
+          <bh-add-item disable="StockCtrl.movement.entry_type" callback="StockCtrl.addItems(numItem)">
```
jniles: :+1:

client/src/js/components/bhStockEntryExitType.js
```diff
@@ -29,6 +30,10 @@ function StockEntryExitTypeController(StockEntryExitType) {
     if (changes.depot) {
       reloadEntryExitTypes();
     }
+
+    if (changes.reset && changes.reset.currentValue === true) {
```
jniles, line 34: :dancer:

client/src/modules/stock/entry/entry.js
```diff
@@ -109,46 +100,94 @@ function StockEntryController(
...
+  function onDateChange(date) {
+    vm.movement.date = date;
+  }
+
+  /**
+   * @method reset
+   * @param {object} form
+   * @description reset the form after submission or on clear
+   */
+  function reset(form = mapEntry.form) {
```
jniles, line 130: Much better :+1:

client/src/modules/stock/entry/entry.js
```diff
@@ -109,46 +100,94 @@ function StockEntryController(
...
+   * @description [grid] add n items (rows) in the grid and call a validation function on each rows
+   */
   function addItems(n) {
     vm.stockForm.addItems(n);
     hasValidInput();
   }
 
+  /**
+   * @method removeItem
+   * @param {number} index
```
jniles, line 179: Nice documentation :+1:

server/controllers/finance/purchases.js
```diff
@@ -361,58 +362,82 @@ function find(options) {
...
+  `;
+
+  const sqlPurchaseDelay = `
+    SELECT
+      p.cost, CEIL(DATEDIFF(m.date, p.date)) AS delay
     FROM stock_movement m
     JOIN lot l ON l.uuid = m.lot_uuid
     JOIN purchase p ON p.uuid = l.origin_uuid
-    WHERE p.uuid = ? AND m.flux_id = ? AND m.is_exit = 0;
+    WHERE p.uuid = ? AND m.flux_id = ? AND m.is_exit = 0 LIMIT 1;
```
jniles, line 395: If you are only taking one of these, should you have a ORDER BY so that we know which one you are taking...?

jniles (commented):
Changes look good to me.  This can be merged whenever.

bors delegate+

PR #2877, opened 2018-06-16
chore: general cleanup

jniles:
This commit does the following cleanup actions:
 1. The SEARCH_QUERRIES i18n key is now spelled correctly (QUERIES).
 2. Removed the date-interval modal.  I couldn't find where it was used.
 3. Made some `translate` filters into `translate` directives.


mbayopanda (approved):
LGTM

bors r+

PR #2878, opened 2018-06-16
Payroll automatic

### LINKED ISSUE #2873: Automatic calculation of  Seniority bonus as well as the family allowance because
Automatic calculation of the seniority bonus as well as the family allowance because the calculation of its two elements are fixed by the Labor Code,

This issue is also intended to make the Payroll module grid Configurable Topics

lomamech:
refactor(Payroll Automatic)  …
- Add Property for Manage Seniority Bonus and Family Allowances

refactor(Calcul of Seniority and Family Allowances)  …
- Improve rubric Modal grid set Configuration of Columns and Saving Columns
- Add Seniority and Family allowances in Rubrics API
- Add function in automatic Rubric on Calculation
- Get seniority and Family allowances in configuration and
  MultiConfiguration of Payroll

bug(Rubric Defined with enterprise currency)  …
- Resolve problem with the rubric defined with Enterprise Currency,

Closes #2873

### DIFF
--- a/client/src/modules/payroll/rubrics/rubrics.html
+++ b/client/src/modules/payroll/rubrics/rubrics.html
@@ -7,6 +7,35 @@
 
     <div class="toolbar">
       <div class="toolbar-item">
+        <div uib-dropdown dropdown-append-to-body data-action="open-tools">
+          <a class="btn btn-default" uib-dropdown-toggle>
+            <span class="fa fa-bars"></span> <span class="hidden-xs" translate>FORM.LABELS.MENU</span> <span class="caret"></span>
+          </a>
+
+          <ul uib-dropdown-menu role="menu" class="dropdown-menu-right">
+            <li role="menuitem">
+              <a href data-method="configure" ng-click="RubricCtrl.openColumnConfiguration()">
+                <i class="fa fa-columns"></i> <span translate>FORM.LABELS.COLUMNS</span>
+              </a>
+            </li>
+
+            <li role="separator" class="divider"></li>
+            <li role="menuitem">
+              <a href data-method="save-state" ng-click="RubricCtrl.saveGridState()">
+                <i class="fa fa-save"></i> <span translate>FORM.BUTTONS.SAVE_GRID_CONFIGURATION</span>
+              </a>
+            </li>
+
+            <li role="menuitem">
+              <a href data-method="clear-state" ng-click="RubricCtrl.clearGridState()">
+                <i class="fa fa-close"></i> <span translate>FORM.BUTTONS.CLEAR_GRID_CONFIGURATION</span>
+              </a>
+            </li>
+          </ul>
+        </div>
+      </div>
+
+      <div class="toolbar-item">
         <button class="btn btn-default text-capitalize" id="create" ui-sref="rubrics.create" data-method="create">
           <span class="fa fa-plus"></span> <span translate>PAYROLL_RUBRIC.ADD_PAYROLL_RUBRIC</span>
         </button>
@@ -22,6 +51,7 @@
           <i class="fa fa-filter"></i>
         </button>
       </div>
+
     </div>
   </div>
 </div>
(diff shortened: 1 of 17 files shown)


client/src/modules/payroll/rubrics/rubrics.html
```diff
@@ -33,7 +63,9 @@
       ui-grid="RubricCtrl.gridOptions"
       class="grid-full-height"
       ui-grid-auto-resize
-      ui-grid-resize-columns>
+      ui-grid-resize-columns
+      ui-grid-move-columns
```
jniles, line 67: Nice additions :+1:

server/controllers/payroll/multiplePayroll/calculation.js
```diff
@@ -25,4 +25,11 @@ function iprTax(annualCumulation, iprScales) {
   return iprValue;
 }
 
+function automaticRubric(coefficient, variable1, variable2) {
+  const value = variable2 ? coefficient * variable1 * variable2 : coefficient * variable1;
```
jniles: This is cool!

One thing that you can do with higher versions of JavsaScript is use rest parameters to make this function smaller (but it might look unclear at first).  It looks like this:
```js
function automaticRubric(coefficient, ...variables) {
  const value = variables.reduce((total, varN) =>total * varN, coefficient);
  return value;
}
```
You can even go smaller by just using:
```js
function automaticRubric(coefficient, ...variables) {
  return variables.reduce((total, next) =>total * next, coefficient);
}
```
What this does is the following:
 1. `...variables` takes all the following function parameters and puts them in an array.  This is called [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters).
 2. `reduce()` will take in a function and a starting number, then return a single number.
 3. `(total, next) => total * next` says that we should take in the total so far, and multiply it by the next number, and return that value.

That means, you can call the function above with:
```js
automaticRubric(1) // => 1
automaticRubric(1, 2) // => 2
automaticRubric(1, 2, 3) // => 6
automaticRubric(1, 23, 7, 8, 12, 30) // => 463680
```
Much more flexible!

(You don't have to do this, just wanted to show you it exists)

server/controllers/payroll/multiplePayroll/setConfiguration.js
```diff
@@ -69,6 +73,11 @@ function config(req, res, next) {
       const workingDayCost = dailySalary * data.working_day;
       const nbChildren = employee.nb_enfant;
 
+      // Calcul of Seniority date Between date_embauche and the end date of Period
+      const diff = moment(periodDateTo).diff(moment(employee.date_embauche));
+      const duration = moment.duration(diff, 'milliseconds');
+      const yearOfSeniority = parseInt(duration.asYears(), 10);
```
jniles, line 79: Nice 👍

server/controllers/payroll/rubrics/index.js
```diff
@@ -25,16 +25,18 @@ function list(req, res, next) {
...
+    r.expense_account_id, a6.number AS six_number, a6.label AS six_label, r.is_ipr, r.value, r.is_tax,
+    r.is_membership_fee, r.is_associated_employee, r.is_seniority_bonus, r.is_family_allowances
     FROM rubric_payroll AS r
     JOIN account AS a4 ON a4.id = r.debtor_account_id
-    JOIN account AS a6 ON a6.id = r.expense_account_id   
+    JOIN account AS a6 ON a6.id = r.expense_account_id
   `;
 
   filters.equals('is_defined_employee');
+  filters.setOrder('ORDER BY r.label');
```
jniles, line 39: 👍

jniles (approved):
@lomamech this is nice code.  As soon as tests pass, it could be merged.

lomamech:
@jniles  Is it possible to redefine the number of lines of code "File setMultiConfiguration.js has (271 lines of code (exceeds 250 allowed). Consider refactoring."

server/controllers/payroll/multiplePayroll/calculation.js
```diff
@@ -25,4 +25,9 @@ function iprTax(annualCumulation, iprScales) {
   return iprValue;
 }
 
+function automaticRubric(coefficient, variables) {
```
jniles, line 28: Just FYI - if you use `...variables`, it will automatically create an array for you.

jniles:
@lomamech, we could, but it is generally a good reminder to refactor code.  Seems like you got it solved though.

jniles (approved):
LGTM.

bors r+

PR #2879, opened 2018-06-16
feat(receipts): allow HTML printing

### LINKED ISSUE #2876: Allow receipts to be rendered in HTML
On slower hardware, `wkhtmltopdf` can delay the rendering of the PDF, even if the data is already registered.  Furthermore, on slower clients, displaying the PDF in the browser can be difficult.

I propose we allow a client-side option that allows a user to request either HTML or PDF for their Receipts.  This would allow users who do not need to print to rapidly gain register information without taxing the server too much.

jniles:
Receipts can now be displayed (and printed, with some conditions) in HTML.  PDF rendering can take a long time on slower hardware - by displaying the receipt as HTML, users can reduce the load on the server and significantly speed up their experience.

The drawback is that receipts printed with HTML are not properly formatted.  If a user needs a nicely formatted receipt, they are far better off using PDF for printing.

Closes #2876.

_GIF1: Showing how it works in practice_


client/src/css/structure.css
```diff
@@ -573,6 +573,11 @@ ol.headercrumb {
     top : 0;
   }
 
+  .bh-print-expand {
+    overflow: auto;
+    height: initial;
+  }
```
mbayopanda, line 579: code-climate suggests that properties must be in alphabetical order 😄

mbayopanda (approved):
LGTM 👍 

The big challenge for HTML renderer is when we print the content to print has the size of the iframe window.

bors r+

PR #2885, opened 2018-06-18
fix(patient): lookup patient group price lists

jniles:
This commit makes the patient lookup check for patient group price lists if a debtor group one does not exist.  This logic should be validated by our clients at HBB.

### DIFF
--- a/server/controllers/medical/patients/index.js
+++ b/server/controllers/medical/patients/index.js
@@ -216,8 +216,10 @@ function update(req, res, next) {
  * @returns {Promise} - the result of the database query
  */
 function lookupPatient(patientUuid) {
-  // convert uuid to database usable binary uuid
+  // convert uuid to patientbase usable binary uuid
   const buid = db.bid(patientUuid);
+
+  let patient;
 
   // @FIXME(sfount) ALL patient queries should use the same column selection and guarantee the same information
   const sql = `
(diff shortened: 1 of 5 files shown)


server/controllers/medical/patients/index.js
```diff
@@ -216,9 +216,11 @@ function update(req, res, next) {
  * @returns {Promise} - the result of the database query
  */
 function lookupPatient(patientUuid) {
-  // convert uuid to database usable binary uuid
+  // convert uuid to patientbase usable binary uuid
```
jniles, line 219: :dancer:

server/controllers/medical/patients/index.js
```diff
@@ -228,25 +230,55 @@ function lookupPatient(patientUuid) {
...
-      AND a.id = dg.account_id
+      dg.account_id, BUID(dg.price_list_uuid) AS price_list_uuid, dg.is_convention,
+      BUID(dg.uuid) as debtor_group_uuid, dg.locked, dg.name as debtor_group_name, u.username,
+      u.display_name AS userName, a.number
+    FROM patient AS p
+      JOIN project AS proj ON p.project_id = proj.id
+      JOIN debtor AS d ON p.debtor_uuid = d.uuid
+      JOIN debtor_group AS dg ON d.group_uuid = dg.uuid
+      JOIN user AS u ON p.user_id = u.id
+      JOIN account AS a ON a.id = dg.account_id
```
mbayopanda, line 241: :+1 it is more clear now

mbayopanda (approved):
LGTM

It seems like there is no change in client side

bors r+

PR #2895, opened 2018-06-20
Fix stock duplicated movement and entry validation

mbayopanda:
This PR fixes the validation in stock entry grid and duplicated data inside stock registry.

### DIFF
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -171,7 +171,7 @@ function StockEntryController(
    */
   function addItems(n) {
     vm.stockForm.addItems(n);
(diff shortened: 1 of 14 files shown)


client/src/modules/stock/entry/entry.js
```diff
@@ -171,7 +171,7 @@ function StockEntryController(
    */
   function addItems(n) {
     vm.stockForm.addItems(n);
-    hasValidInput();
+    vm.hasValidInput = hasValidInput();
```
jniles, line 174: :+1:

jniles (approved):
This is an important fix.  LGTM.

bors r+

PR #2913, opened 2018-06-21
fix(operating report): update presentation

### LINKED ISSUE #2907: Report Operating Account
There are signs (minus) on the Report Level OPERATING ACCOUNT, It must not allow the SIGN (-) and change the interface of the Balance Remove the line Balance

jniles:
This commit updates the operating account report presentation with the fixes desired by IMCK.

Closes #2907.

This is what the new format looks like:
newoperatingpresentation
_Fig 1: New Presentation of Operating Account_

### DIFF
--- a/client/src/modules/reports/generate/operating/operating.html
+++ b/client/src/modules/reports/generate/operating/operating.html
@@ -8,7 +8,7 @@
 <div ng-show="!ReportConfigCtrl.previewGenerated">
   <div class="row">
     <div class="col-md-12">
-      <h3 class="text-capitalize" translate>TREE.OPERATING_ACCOUNT</h3> 
+      <h3 class="text-capitalize" translate>TREE.OPERATING_ACCOUNT</h3>
     </div>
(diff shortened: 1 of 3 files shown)


lomamech:
@jniles  This is a good modification, but we could only display the values without the () or even in Red color, because this Report is exactly a section of Canvas Snis, the Director of the IMCK really appreciated it

jniles:
@lomamech, how should we handle the income accounts that have a debit balance?  For example, in the screenshot above, look at both `60310010` and `60310011`.   Account `60310010` has a debit balance (which is not normal accounting) and account `60310011` has a credit balance (which is normal).  How should we tell those two apart?

lomamech:
Thanks @jniles  I will explain to them the reason for these parentheses

lomamech:
@jeremielodi  can you review and merge this PR

jniles:
@jeremielodi @lomamech what is the status of this PR?

PR #2914, opened 2018-06-21
feat(price list) print to pdf

### LINKED ISSUE #2906: [Reports] pricing list of items and services
Add a report that displays the pricing list of items and services versus cash paying patients as well as conventioned

jeremielodi:
This PR helps to print a price list
closes #2906

### DIFF
--- a/client/src/modules/prices/prices.html
+++ b/client/src/modules/prices/prices.html
@@ -44,6 +44,13 @@
                     <i class="fa fa-pencil"></i> <span translate>FORM.BUTTONS.EDIT</span>
                   </a>
                 </td>
+                <td width='50'>
+                  <a class="btn btn-xs btn-default"
+                    target="_blank"
+                    href="/prices/report/{{list.uuid}}?{{PriceListCtrl.download()}}">
+                    <i class="fa fa-print"></i> <span translate>FORM.BUTTONS.PRINT</span>
+                  </a>
+                </td>
               </tr>
               <tr ng-if="PriceListCtrl.priceList.data.length===0">
                 <td colspan="3"><i translate> PRICE_LIST.NO_RECORDS </i></tr>
--- a/server/controllers/finance/reports/priceList/index.js
+++ b/server/controllers/finance/reports/priceList/index.js
@@ -1,0 +1,105 @@
+/**
+ * @overview server/controllers/finance/reports/priceList/index.js
+ *
+ * @description
+ * This file contains code to create a PDF report for a price list
+ *
+ * @requires db
+ * @requires lodash
+ * @requires util
+ * @requires ReportManager
+ */
+const _ = require('lodash');
+const ReportManager = require('../../../../lib/ReportManager');
+const db = require('../../../../lib/db');
+const util = require('../../../../lib/util');
+const q = require('q');
+
+
+const TEMPLATE = './server/controllers/finance/reports/priceList/report.handlebars';
+
+const PDF_OPTIONS = {
+  filename : 'FORM.LABELS.PRICE_LIST',
+};
+
+exports.report = (req, res, next) => {
+  const options = req.query;
+
+  let report;
+
+  _.defaults(options, PDF_OPTIONS);
+
+  // set up the report with report manager
+  try {
+    report = new ReportManager(TEMPLATE, req.session, options);
+
+    const priceListUUID = db.bid(req.params.uuid);
+    return lookupPriceList(priceListUUID).then(([row, rows]) => {
+      const _priceList = row;
+      _priceList.items = rows;
+      const { items } = _priceList;
+
+      items.forEach(item => {
+        if (item.priceListValue) {
+          if (item.is_percentage === 1) {
+            item.price += util.roundDecimal((item.price * item.priceListValue) / 100, 2);
+          } else {
+            item.price = util.roundDecimal(item.priceListValue, 2);
+          }
+        }
+      });
+      // group by inventory group
+      let groups = _.groupBy(items, i => i.groupName);
+
+      // make sure that they keys are sorted in alphabetical order
+      groups = _.mapValues(groups, lines => {
+        _.sortBy(lines, 'label');
+        return lines;
+      });
+      return report.render({ groups });
+    })
+      .then(result => {
+        res.set(result.headers).send(result.report);
+      })
+      .catch(next)
+      .done();
+
+  } catch (e) {
+    return next(e);
+  }
+};
+
+
+function lookupPriceList(uuid) {
+  const priceListSql =
+    `SELECT BUID(uuid) AS uuid, label, description, created_at, updated_at
+    FROM price_list WHERE uuid = ?;`;
+
+  const inventorySql = `
+    SELECT BUID(inventory.uuid) as uuid, inventory.code, inventory.text AS label, inventory.price, 
+      iu.abbr AS unit,
+      it.text AS type, ig.name AS groupName, BUID(ig.uuid) AS group_uuid, ig.expires, 
+      ig.unique_item, inventory.consumable,inventory.locked, inventory.stock_min,
+      inventory.stock_max, inventory.created_at AS timestamp, inventory.type_id, inventory.unit_id,
+      inventory.note,  inventory.unit_weight, inventory.unit_volume, 
+      ig.sales_account, ig.stock_account, ig.donation_account, inventory.sellable, inventory.note, 
+      inventory.unit_weight, inventory.unit_volume, ig.sales_account, ig.stock_account, ig.donation_account,
+      ig.cogs_account, inventory.default_quantity, s1.priceListValue, s1.is_percentage
+    FROM inventory 
+      JOIN inventory_type AS it ON inventory.type_id = it.id  
+      JOIN inventory_unit AS iu  ON  inventory.unit_id = iu.id
+      JOIN inventory_group AS ig ON inventory.group_uuid = ig.uuid
+    LEFT JOIN(
+      SELECT pl.inventory_uuid, pl.value as priceListValue, pl.is_percentage,
+        pl.created_at, iv.text as inventoryLabel, iv.price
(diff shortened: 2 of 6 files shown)


server/controllers/finance/reports/priceList/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+  return db.one(sql, [uuid])
+    .then((row) => {
+      _priceList = row;
+
+      sql =
+        `SELECT BUID(pl.uuid) as uuid, pl.label, pl.value, pl.is_percentage, pl.created_at, iv.text as inventoryLabel
+        FROM price_list_item pl
+        JOIN inventory as iv ON iv.uuid = pl.inventory_uuid
+        WHERE pl.price_list_uuid = ?;`;
```
mbayopanda: You have to enhance the query for having all inventory that can be sold. 

The idea behind the price list is a list of article or service that the enterprise sale and their cost computed according enterprise rate.

You could see a kind of price list without calculation when you export the inventory registry to pdf.

server/controllers/finance/reports/priceList/report.handlebars
```diff
@@ -0,0 +1,49 @@
+{{> head title="{{translate 'FORM.LABELS.PRICE_LIST' }}" }}
```
mbayopanda: This template must be updated to be conform with the `tarification` model, you can remove columns like `consumable` and `default quantity`

price list correct

client/src/modules/prices/prices.html
```diff
@@ -44,6 +44,13 @@
                     <i class="fa fa-pencil"></i> <span translate>FORM.BUTTONS.EDIT</span>
                   </a>
                 </td>
+                <td width='50'>
+                  <a class="btn btn-xs btn-default" id="price_list_print_{{ $index+1 }}"
+                    target="_blank"
+                    href="/prices/report/{{list.uuid}}?{{PriceListCtrl.download()}}">
```
mbayopanda: 👍

mbayopanda (changes requested):
The ideal thing to do is to use 2.x standard for the ui (use of grid and modals) and allow to print or export a price list as `Tarification`.

server/controllers/finance/reports/priceList/index.js
```diff
@@ -0,0 +1,109 @@
+/**
+ * @overview server/controllers/finance/reports/financial.employee.js
```
mbayopanda: Could you fix this to : `@overview server/controllers/finance/reports/priceList/index.js`

server/controllers/finance/reports/priceList/index.js
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @description
+ * This file contains code to create a PDF report for financial activities of an employee
+ *
+ * @requires Employee
+ * @requires ReportManager
+ */
```
mbayopanda: These information are not correct would you fix them.

server/controllers/finance/reports/priceList/index.js
```diff
@@ -0,0 +1,109 @@
...
+    const priceListUUID = db.bid(req.params.uuid);
+    return lookupPriceList(priceListUUID).then(data => {
+      const { items } = data;
+
+      items.forEach(item => {
+        if (item.priceListValue) {
+          if (item.is_percentage === 1) {
+            item.price += util.roundDecimal((item.price * item.priceListValue) / 100, 2);
+          } else {
+            item.price += util.roundDecimal(item.priceListValue, 2);
```
mbayopanda: I think you must fix this line with `=` instead of `+=` for not percentage values

server/controllers/finance/reports/priceList/index.js
```diff
@@ -0,0 +1,109 @@
...
+        ig.unique_item, inventory.consumable,inventory.locked, inventory.stock_min,
+        inventory.stock_max, inventory.created_at AS timestamp, inventory.type_id, inventory.unit_id,
+        inventory.note,  inventory.unit_weight, inventory.unit_volume, 
+        ig.sales_account, ig.stock_account, ig.donation_account, inventory.sellable, inventory.note, 
+        inventory.unit_weight, inventory.unit_volume, ig.sales_account, ig.stock_account, ig.donation_account,
+        ig.cogs_account, inventory.default_quantity, s1.priceListValue, s1.is_percentage
+      FROM inventory JOIN inventory_type AS it
+        JOIN inventory_unit AS iu JOIN inventory_group AS ig ON
+        inventory.type_id = it.id AND inventory.group_uuid = ig.uuid AND
+        inventory.unit_id = iu.id
```
mbayopanda: Could you write JOINs like this : 
```sql
FROM inventory 
JOIN inventory_group ig ON ig.uuid = inventory.group_uuid
JOIN inventory_type it ON it.id = inventory.type_id
JOIN inventroy_unit iu ON iu.id = inventory.unit_id
```
It is more clear to read

mbayopanda (changes requested):
The price list report now is perfect 👍 

I made some comments on things you could change before we merge this PR.

server/controllers/finance/reports/priceList/index.js
jeremielodi: @mbayopanda last time you told me that if a price list is for example 100% and the inventory price is 0.5$, the patient will pay 0.5 + 0.5 = 1$, right?

server/controllers/finance/reports/priceList/index.js
jeremielodi: I used on of sql query in the stock core

server/controllers/finance/reports/priceList/index.js
jniles: If the price isn't a percentage, we just replace the price.  That way the person responsible for the price list doesn't have to do a lot of calculations to try and arrive at his desired price.

client/src/modules/prices/prices.html
```diff
@@ -44,6 +44,13 @@
                     <i class="fa fa-pencil"></i> <span translate>FORM.BUTTONS.EDIT</span>
                   </a>
                 </td>
+                <td width='50'>
+                  <a class="btn btn-xs btn-default" id="price_list_print_{{ $index+1 }}"
```
jniles: Do you need the `id` here?  If you do, it would be much better to use the price list's `uuid` instead of the `$index`.  It will make testing easier in the future, since if we add more price lists, this ID might change.  If it's the price list's `uuid`, it will not.

server/controllers/finance/reports/priceList/index.js
```diff
@@ -0,0 +1,111 @@
...
+        ig.cogs_account, inventory.default_quantity, s1.priceListValue, s1.is_percentage
+      FROM inventory 
+        JOIN inventory_type AS it ON inventory.type_id = it.id  
+        JOIN inventory_unit AS iu  ON  inventory.unit_id = iu.id
+        JOIN inventory_group AS ig ON inventory.group_uuid = ig.uuid
+      LEFT JOIN(
+        SELECT pl.inventory_uuid, pl.value as priceListValue, pl.is_percentage,
+          pl.created_at, iv.text as inventoryLabel, iv.price
+        FROM price_list_item pl
+        JOIN inventory as iv ON iv.uuid = pl.inventory_uuid
```
jniles: It doesn't look like you use `inventory` in the `LEFT JOIN()` statement.  Can you remove it?

server/controllers/finance/reports/priceList/index.js
```diff
@@ -0,0 +1,111 @@
...
+};
+
+
+function lookupPriceList(uuid) {
+  let _priceList;
+  let sql =
+    `SELECT BUID(uuid) AS uuid, label, description, created_at, updated_at
+    FROM price_list WHERE uuid = ?;`;
+
+  return db.one(sql, [uuid])
```
jniles: It doesn't look like the second SQL statement (looking up price list items) depends on the first.   That means you could execute them at the same time to speed the query up.  Like this:
```js
const sql = `SELECT BUID(uuid), ....`;
const itemSql = `SELECT BUID(inventory.uuid) ....`;

return Promise.all([
  db.one(sql, [uuid]),
  db.exec(sql, [uuid]),
])
  .then(([row, rows]) => {
    // do something with the result
  });
```

server/controllers/finance/reports/priceList/index.js
```diff
@@ -0,0 +1,111 @@
...
+  let report;
+
+  _.defaults(options, PDF_OPTIONS);
+
+  // set up the report with report manager
+  try {
+    report = new ReportManager(TEMPLATE, req.session, options);
+
+    const priceListUUID = db.bid(req.params.uuid);
+    return lookupPriceList(priceListUUID).then(data => {
```
jniles: Having a `return` statement within a `try {} catch{}` is a bit confusing.  What code are you afraid will error?  Is there any way to isolate the code in a `try...catch` and then continue with the rest of the code outside of the `try...catch` blocks?

mbayopanda:
@jeremielodi What is the state of this PR ?

server/controllers/finance/reports/priceList/index.js
jeremielodi: it is used at this point  ``` iv.text as inventoryLabel, iv.price```

server/controllers/finance/reports/priceList/index.js
```diff
@@ -0,0 +1,105 @@
...
+      const _priceList = row;
+      _priceList.items = rows;
+      const { items } = _priceList;
+
+      items.forEach(item => {
+        if (item.priceListValue) {
+          if (item.is_percentage === 1) {
+            item.price += util.roundDecimal((item.price * item.priceListValue) / 100, 2);
+          } else {
+            item.price = util.roundDecimal(item.priceListValue, 2);
```
mbayopanda, line 47: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #2915, opened 2018-06-21
Display sold and prepayment on an invoice

mbayopanda:
This PR add an option in the enterprise to allow : 
- to display prepayment on an invoice
- to display sold of the invoice inside the invoice

## Enterprise Settings
image

## Caution (20000fc) less than the invoice
image

## Caution (100$) more than the invoice
image


server/models/Update/add_prepayment_inside_invoice.sql
```diff
@@ -0,0 +1,11 @@
+/*
+  UPDATE ENTERPRISE_SETTING TABLE
+  ===============================
+
+  ADD PREPAYMENT INSIDE INVOICE OPTION
+  If yes, the prepayment will be displayed on the invoice as proof that
+  there was a prepayment
+*/
+
+ALTER TABLE `enterprise_setting` ADD COLUMN `enable_prepayment_inside_invoice` TINYINT(1) NOT NULL DEFAULT 0;
```
jniles: 👍

server/controllers/finance/reports/invoices/receipt.handlebars
```diff
@@ -132,32 +132,68 @@
...
         </tr>
         <tr>
           <td colspan='2' class="text-right">
-            <h4><b> {{numberToText cost lang metadata.enterprise.currencyName}}</b></h4>
+            <em>{{numberToText cost lang metadata.enterprise.currencyName}}</em>
           </td>
         </tr>
 
+        <!-- Prepayment inside invoice -->
+        {{#if prepaymentInsideInvoice}}
```
jniles: It looks like we have enough space - could we put both currencies on the same line?  That way, we don't waste vertical space.  Something like:

| | | | 
|:---:| ----:| -----:|
| total | 86.00$ | 139320FC |

jniles (commented):
@mbayopanda this doesn't look like it has anything to do with prepayments - it just displays the balance of the invoice on the invoice.  So it would apply to both prepayments and regular payments, right?

The enterprise setting here is probably unnecessary.  Let's just show this by default.  It's similar to what we do on Cash Payment receipts anyway.  However, we should encourage our users to use the Patient Financial Activity to check the balance of a patient before they are discharged from the hospital.

mbayopanda:
@jniles you right, this PR just display the balance and the paid amount on an invoice

We define an option in enterprise because we want to keep bhima flexible to any other use case. we let users to decide if they want this information on invoice or not

server/controllers/finance/reports/invoices/receipt.handlebars
mbayopanda: @jniles do we need the total amount in full letter (e.g. Dix mille dollars) ?

jniles:
Alright, if we keep this option as an enterprise setting, then we need to change the name of the setting.  Something like "Show balance on Invoice Receipt" or something similar.

server/models/schema.sql
```diff
@@ -628,6 +628,7 @@ CREATE TABLE `enterprise_setting` (
   `enable_delete_records` TINYINT(1) NOT NULL DEFAULT 0,
   `enable_prepayments` TINYINT(1) NOT NULL DEFAULT 1,
   `enable_password_validation` TINYINT(1) NOT NULL DEFAULT 1,
+  `enable_prepayment_inside_invoice` TINYINT(1) NOT NULL DEFAULT 0,
```
jniles: This is the most important label to change (since it will be the most difficult to change later).  Let's call it `enable_balance_on_invoice_receipt`.

mbayopanda:
@jniles would you review this PR please

jniles (approved):
LGTM.  Very cleanly implemented.

bors r+

PR #2890, opened 2018-06-18
navigation exchange rate

mbayopanda:
This PR move the exchange rate view into its own module in the administration folder.

close #2889

PR #2918, opened 2018-06-22
Remove filter action

### LINKED ISSUE #2911: Invoicing Fees Inline Filter
The Invoicing Fees module's inline filter also allows a user to filter on "actions".  This isn't standard BHIMA practice and should be removed.  See the figure below:

buginvoicingfeesfilter
_Fig 1: Invoicing Fees Actions_

lomamech:
bug(Remove filter on Action)
- Set enableFiltering false in column action on Invoicing fees

feature(Filtering)  
- Add inline filter in Employees and Multipayroll Registry
- Complete translate file key

closes #2911

### DIFF
--- a/client/src/i18n/en/vouchers.json
+++ b/client/src/i18n/en/vouchers.json
@@ -53,7 +53,7 @@
       "PAYMENT_EMPLOYEES": "Payment of wages {{date}}"
     },
     "SIMPLE": {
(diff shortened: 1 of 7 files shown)


jniles (approved):
LGTM.  This might have conflicts with #2890, but we'll try to merge it anyway!

bors r+

Issue #2901, opened 2018-06-20
Prevent Billing and Ordering


lomamech:
- Prevent billing of items that are missing profit accounts
- Prevent the ordering of items that are missing expense and inventory accounts

jniles:
We handled the first case a long time ago at Lukonga ... I wonder what happened?

lomamech:
There was a case at HBB where an Inventory belonged to an Inventory group that did not have a profit account, we had to go to Accounting to add a Profit Account to this inventory group

jniles:
I'll look into it and try to fix it.

PR #2919, opened 2018-06-22
fix(invoice): block sales_account if id is zero

jniles:
This commit fixes a bug at IMCK where the unknown sales_accounts have id 0 instead of being NULL.

Partially addresses #2901.


lomamech:
Good
inventory

PR #2923, opened 2018-06-23
Fix CalculatePrepaymentBalance Procedure

mbayopanda:
This PR fix the error that occurs when we have these steps :
- 1. Patient A paid a caution for X amount
- 2. Invoicing patient A for Y amount for the invoice IV.1
- 3. Cancel the invoice IV.1
- 4. Invoicing patient A for Z amount for the invoice IV.2 throw an error in the procedure `LinkPrepaymentsToInvoice`

The solution adopted in this PR is to ignore cancelled invoices

### DIFF
--- a/server/models/procedures/invoicing.sql
+++ b/server/models/procedures/invoicing.sql
@@ -749,14 +749,16 @@ BEGIN
     SELECT p.reference_uuid AS uuid, p.debit_equiv as debit, p.credit_equiv as credit, p.trans_date as date
     FROM posting_journal AS p
       JOIN cash c ON c.uuid = p.reference_uuid
-    WHERE c.debtor_uuid = debtor_uuid AND c.reversed = 0 AND c.is_caution = 1;
+    WHERE c.debtor_uuid = debtor_uuid AND c.reversed = 0 AND c.is_caution = 1
+      AND p.trans_id NOT IN (SELECT pj.trans_id FROM posting_journal pj JOIN invoice i ON i.uuid = pj.reference_uuid WHERE pj.trans_id = p.trans_id AND i.reversed = 1);
 
   -- pull in general_ledger references
   INSERT INTO stage_payment_references
     SELECT p.reference_uuid AS uuid, p.debit_equiv as debit, p.credit_equiv as credit, p.trans_date as date
     FROM general_ledger AS p
       JOIN cash c ON c.uuid = p.reference_uuid
-    WHERE c.debtor_uuid = debtor_uuid AND c.reversed = 0 AND c.is_caution = 1;
+    WHERE c.debtor_uuid = debtor_uuid AND c.reversed = 0 AND c.is_caution = 1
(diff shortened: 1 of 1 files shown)


mbayopanda:
For tshikaji update we have to : 
1. Drop the procedure `CalculatePrepaymentBalances`
```sql
DROP PROCEDURE CalculatePrepaymentBalances;
```

2. Recreate the procedure `CalculatePrepaymentBalances`.

server/models/procedures/invoicing.sql
```diff
@@ -749,14 +749,16 @@ BEGIN
     SELECT p.reference_uuid AS uuid, p.debit_equiv as debit, p.credit_equiv as credit, p.trans_date as date
     FROM posting_journal AS p
       JOIN cash c ON c.uuid = p.reference_uuid
-    WHERE c.debtor_uuid = debtor_uuid AND c.reversed = 0 AND c.is_caution = 1;
+    WHERE c.debtor_uuid = debtor_uuid AND c.reversed = 0 AND c.is_caution = 1
+      AND p.trans_id NOT IN (SELECT pj.trans_id FROM posting_journal pj JOIN invoice i ON i.uuid = pj.reference_uuid WHERE pj.trans_id = p.trans_id AND i.reversed = 1);
```
jniles, line 753: For these, it's better to use `record_uuid`, since it is binary and faster to compare.

jniles (commented):
@mbayopanda, I'm not sure this is the right solution.

It seems like the problem is this:
 1. Patient pays a caution of $100.
 2. Patient is invoiced $30.
 3. Patient _automatically_ pays $30 from their caution via the Caution Link.
 4. User realizes a mistake happened, tries to reverse the invoice.

What is step 3 linked to now?

To think about this another way, imagine what BHIMA does in this scenario:
 1. Patient is invoiced $30.
 2. Patient pays $30.
 3. User realizes a mistake happened, tries to reverse invoice.

In the above case, BHIMA will block the reversal of the invoice, since it has a payment against it.  Instead you must _first reverse the cash payment_ and then you can reverse the invoice.

We should do the same thing here.  We must find a way to reverse the caution link before reversing the invoice.  That can either be:
 1. Manual - an interface to reverse the voucher.
 2. Automatic - if you are reversing an invoice with a caution link against it, it will automatically reverse the caution link.

In my mind, these are the only two options if we continue to automatically pay for invoices with caution links.  What do you think?

server/models/procedures/invoicing.sql
mbayopanda: You're right the query must be optimized because with real dataset there is a `ER_LOCK_WAIT_TIMEOUT'`

mbayopanda: @jniles the purpose of this procedure is to get a balance, so why not use a simple query for getting the balance of the debtor

mbayopanda:
@jniles you're right, i think the second option is maybe necessary : reverse the caution link transaction automatically.

server/models/procedures/invoicing.sql
jniles: I'm pretty sure the purpose is to get the balance on this invoice, but I'll have to read the code to make sure.  I'll look into it.

jniles:
@mbayopanda, if we implement reversals, and if this is to slow in production, can we close this PR?

PR #2925, opened 2018-06-24
 refactor(Search Project)

### LINKED ISSUE #2924: Adding search items in some registers
- It was found that the modal windows of some Registers did not allow searching on some columns of the Grid, which is why this PR will strengthen the search capability in most Registers

lomamech:
- Adding the reasearch Project in Patient Registry
- Add Column Project in Patient Invoice Registry and also the research
  Invoice By Project
- Update and Improve Multipayroll Registry, Add Aggregation Type for Count
  Row and For Sum Net Salary and Balance

- Update Voucher Registry add the Research By Project and By Currency
- And Fix Format view for TransactionTypeMap when a voucher don't have
  transaction Type

Closes #2924

### DIFF
--- a/client/src/modules/invoices/patientInvoice.service.js
+++ b/client/src/modules/invoices/patientInvoice.service.js
@@ -135,6 +135,7 @@ function PatientInvoiceService(
     { key : 'reversed', label : 'FORM.INFO.CREDIT_NOTE' },
     { key : 'defaultPeriod', label : 'TABLE.COLUMNS.PERIOD', valueFilter : 'translate' },
     { key : 'debtor_group_uuid', label : 'FORM.LABELS.DEBTOR_GROUP' },
+    { key : 'project_id', label : 'FORM.LABELS.PROJECT' },
     { key : 'cash_uuid', label : 'FORM.INFO.PAYMENT' },
   ]);
 
--- a/client/src/modules/invoices/registry/search.modal.html
+++ b/client/src/modules/invoices/registry/search.modal.html
@@ -50,6 +50,14 @@
           </div>
         </div>
 
+        <!-- filter by project -->
+        <bh-project-select
+          project-id="$ctrl.searchQueries.project_id"
+          name="project_id"
+          on-select-callback="$ctrl.onSelectProject(project)">
+          <bh-clear on-clear="$ctrl.clear('project_id')"></bh-clear>
+        </bh-project-select>
+
         <bh-service-select
           service-id="$ctrl.searchQueries.service_id"
           name="service_id"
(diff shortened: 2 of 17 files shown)


client/src/modules/invoices/patientInvoice.service.js
```diff
@@ -135,6 +135,7 @@ function PatientInvoiceService(
     { key : 'reversed', label : 'FORM.INFO.CREDIT_NOTE' },
     { key : 'defaultPeriod', label : 'TABLE.COLUMNS.PERIOD', valueFilter : 'translate' },
     { key : 'debtor_group_uuid', label : 'FORM.LABELS.DEBTOR_GROUP' },
+    { key : 'project_id', label : 'FORM.LABELS.PROJECT' },
```
jniles, line 138: Do we really need to put the project names here?  Shouldn't it be obvious from the identifiers? `PA.HBB.XXX`?

client/src/modules/invoices/registry/search.modal.html
```diff
@@ -50,6 +50,14 @@
           </div>
         </div>
 
+        <!-- filter by project -->
+        <bh-project-select
```
jniles, line 54: 👍

client/src/modules/multiple_payroll/multiple_payroll.js
```diff
@@ -50,11 +52,17 @@ function MultiplePayrollController(
     displayName : 'FORM.LABELS.NET_SALARY',
     headerCellFilter : 'translate',
     cellFilter : 'currency:row.entity.currency_id',
+    aggregationType : uiGridConstants.aggregationTypes.sum,
+    aggregationHideLabel : true,
+    footerCellFilter : 'currency:'.concat(Session.enterprise.currency_id),
```
jniles: Can you pay employees in multiple currencies?  What happens if the grid contains FC + USD?

jniles (commented):
@lomamech, search functionality works well.  Good job!

Do we really need to include the project name in all these registries?  Shouldn't it be clear from the identifiers?  It isn't necessary to search on it.

client/src/modules/multiple_payroll/multiple_payroll.js
lomamech: You're right, I'm going to delete it

client/src/modules/invoices/patientInvoice.service.js
lomamech: It's true it's more information, but it can still help other users like the Administrator and the Director

jniles (approved):
LGTM.

bors r+

PR #2941, opened 2018-06-25
fix stock loading button, transfer validation and depot selection

mbayopanda:
This PR fixes : 
- loading button in the stock adjustment module
- load all depot with bhDepotSelect
- run validation after select a transfer from the transfer modal

### DIFF
--- a/client/src/modules/stock/adjustment/adjustment.js
+++ b/client/src/modules/stock/adjustment/adjustment.js
@@ -227,9 +227,11 @@ function StockAdjustmentController(
 
     movement.lots = lots;
 
+    vm.$loading = true;
     return Stock.movements.create(movement)
       .then(document => {
         vm.Stock.store.clear();
+        vm.$loading = false;
         ReceiptModal.stockAdjustmentReceipt(document.uuid, fluxId);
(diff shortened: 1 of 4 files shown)


client/src/modules/stock/adjustment/adjustment.js
```diff
@@ -227,9 +227,11 @@ function StockAdjustmentController(
 
     movement.lots = lots;
 
+    vm.$loading = true;
```
jniles: A better solution is to change [this line](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/stock/adjustment/adjustment.html#L143) from
```html
<bh-loading-button loading-state="StockCtrl.$loading" disabled="!StockCtrl.validForSubmit">
```
to
```html
<bh-loading-button loading-state="StockForm.$loading" disabled="!StockCtrl.validForSubmit">
```

Basically, the author meant to type "StockForm", but they typed "StockCtrl" originally.  This should fix it without introducing new code in the controller.

jniles (approved):
One comment, otherwise looks good.

PR #2942, opened 2018-06-25
Feature: Reverse Vouchers

### LINKED ISSUE #1204: (feature request) Reverse Voucher Record
All our other modules allow a user to quickly fix an error in a record by quickly reversing the transaction - either a cash payment or an invoice record.  It would be good to provide similar functionality on the vouchers page.

### LINKED ISSUE #2927: Credit note must reverse also caution link associated to an invoice
When is reversed by the credit note tool, automatically if there are some caution link associated, these caution link also must be reversed.

If not, the user cannot invoicing the patient again. That is a **bug**

jniles:
This PR implements the reverse voucher feature.  It is functionally identical to the other reversals, using the `CREDIT_NOTE` transaction type (which should probably now be called correction).  When a row is reversed, it turns red like the other registries and a call to `ReverseTransaction()` is made.

Additionally, a bug is fixed with cash reversals that show up with empty descriptions.  This was caused by an incorrect binding.

Closes #1204.  Closes #2927.

**Screenshots**
voucherreverserecorddropdown
_Fig 1: Reverse Record in the Dropdown_

voucherreverserecordmodal
_Fig 2: Reverse Record Modal_

voucherreverserecordrow
_Fig 3: A reversed row in the Voucher Registry_


client/src/modules/vouchers/vouchers.service.js
```diff
@@ -142,14 +143,16 @@ function VoucherService(
...
-   * any transaction in the posting_journal by creating a
-   * new transaction that is an exact duplicate of the original transaction with the
-   * debits and credits switched.
+   * @method reverse
+   *
+   * @description
+   * This method reverses a transaction.
+   * bhima should automatically be able to reverse any transaction in the
+   * posting_journal by creating a new transaction that is an exact duplicate of
+   * the original transaction with the debits and credits switched.
```
mbayopanda, line 152: 👍

server/controllers/finance/vouchers.js
```diff
@@ -370,6 +371,10 @@ function safelyDeleteVoucher(guid) {
       WHERE voucher.uuid = ?;
   `;
 
+  const TOGGLE_VOUCHER_REVERSAL = `
+      UPDATE voucher SET voucher.reversed = 0 WHERE voucher.uuid = ?;
+  `;
```
mbayopanda, line 376: Is is correct to have `SET voucher.reversed = 0` for the `TOGGLE_VOUCHER_REVERSAL` instead of `SET voucher.reversed = 1`

client/src/i18n/en/form.json
```diff
@@ -60,6 +60,7 @@
             "RESET": "Reset",
             "RESET_DEFAULTS": "Reset Defaults",
             "RETURN": "Return",
+            "REVERSE_RECORD" : "Reverse Record",
```
mbayopanda, line 63: The french translation is missing

jniles: Could you give me one?

jniles:
Note: to update IMCK's database, run:
```sql
ALTER TABLE voucher ADD COLUMN  `reversed`        TINYINT NOT NULL DEFAULT 0;
```

client/src/i18n/en/form.json
mbayopanda: We can use `Inverser l'enregistrement`

server/controllers/finance/vouchers.js
jniles: Yeah, I can see how this is confusing.

The code here is trying to flip the `reversed` flag if the reversing record (i.e. the voucher that made the reverse) is deleted.  So, if you delete the credit note, it will automatically remove the `reversed=1` from the invoice.   The same thing works here for vouchers.

jniles:
@mbayopanda, this is ready for a second review.

I caught a bug in the route that checks if the invoice has been paid - it was using a JOIN, but never said ON.  This has been fixed.

server/controllers/finance/cash.js
```diff
@@ -253,19 +255,28 @@ function update(req, res, next) {
...
 
-  const sql = `
-    SELECT cash_item.cash_uuid, cash_item.invoice_uuid
-    FROM cash JOIN cash_item
-    WHERE cash_item.invoice_uuid = ? AND cash.reversed <> 1
-    GROUP BY cash_item.invoice_uuid;
+
+  const getInvoicePayment = `
+    SELECT DISTINCT BUID(cash_item.cash_uuid) cash_uuid, BUID(cash_item.invoice_uuid) invoice_uuid, cash.reversed
+    FROM cash JOIN cash_item ON cash.uuid = cash_item.cash_uuid
```
mbayopanda, line 261: 👍 good catch

mbayopanda (approved):
LGTM :+1

And it works fine

bors r+

PR #2948, opened 2018-06-27
fix(General Ledger): ignore empty rows correctly

jniles:
This commit fixes the General Ledger presentation so that it ignores empty rows correctly.  Previously we filtered out accounts whose balance was 0, even if they had movements during the exercise.  Now we only
remove accounts which are zero in every period.

You will find attached an upgrade script that will remove the REPORT transactions from the IMCK dataset and make them into period 0 transactions.

[period-zero-imck.sql.txt](https://github.com/IMA-WorldHealth/bhima-2.X/files/2141720/period-zero-imck.sql.txt)

### DIFF
--- a/server/controllers/finance/generalLedger/index.js
+++ b/server/controllers/finance/generalLedger/index.js
@@ -23,12 +23,13 @@ const PERIODS = [
   0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12,
 ];
 
-// expose to the api
+// expose to the API
 exports.list = list;
 exports.getAggregates = getAggregates;
 
 // expose to server controllers
 exports.getAccountTotalsMatrix = getAccountTotalsMatrix;
+exports.getAccountTotalsMatrixAggregates = getAccountTotalsMatrixAggregates;
 
 /**
  * @function list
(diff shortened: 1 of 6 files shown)


jniles:
~~So, this code is correct, but I'm not sure the SQL script is correct yet.  Let me work on it more today and I'll figure it out.~~

Okay, I think this is pretty good.

server/controllers/finance/generalLedger/index.js
```diff
@@ -104,7 +105,22 @@ function getAccountTotalsMatrix(fiscalYearId) {
   `;
 
   //  returns true if all the balances are 0
-  const isEmptyRow = (row) => row.balance === 0;
+  const isEmptyRow = (row) => (
```
mbayopanda, line 108: :+1:

mbayopanda (commented):
The code LGTM and the feature too,

but i just tried to see the `Relevé de compte` for a random account by i got nothing

jniles:
@mbayopanda, it looks like this bug exists in master.

I'm probably not going to get a chance to address this before leaving the country.  However, it is very important that the period zero sql is run at IMCK.  Patrick and I validated the numbers today for 2014, 2015, and 2016.

mbayopanda (approved):
LGTM 👍 

bors r+

Issue #2741, opened 2018-05-09
Redesign Cash Transfers between Cashboxes


jniles:
BHIMA currently halfheartedly supports cash transfers via a cashbox menu item.  A screenshot of this tool appears below:
![cashtransfer](https://user-images.githubusercontent.com/896472/39820627-2861944c-539e-11e8-85e3-7b56bbe2ff25.PNG)
_Fig 1: Cash Transfer_

There are a few issues with this tool:
 1. It only completes one side of the transaction.  Currently, it exports to a "transfer account" but someone will still need to take the money from a transfer account to the final account.
 2. There is no indication of the amount of money in the cashbox.  It would be ideal if we could include this number for the user.
 3. Transfer accounts are tied to a cashbox, but OHADA wants to have a different transfer account for a _combination_ of cashboxes.  That means that Caisse Aux -> Caisse Principale A is a different account from Caisse Aux -> Caisse Principale B.  BHIMA doesn't currently allow this feature.

I imagine we could take two approaches to refining this tool:
 1. Add the ability for a user to select a different cashbox to send money to.  If the accounting practices of their country require that they use a transfer account as OHADA does, they could check a box to "use a transfer account" and specify it.
 2. Add a "transfer account" matrix to the database.  For every cashbox in the system, we will need to configure the transfer accounts between them.  The user can then select a cashbox and the transfer account will be specified for them.

PR #2956, opened 2018-06-28
feature(Using Transfer Account)

### LINKED ISSUE #1417: Outil Transfert d'argent dans le module d'application Paiement
Je constate que le tranfert d'argent de la petite caisse vers la grande caisse ne génère qu'une seule écriture comptable automatiquement et la seconde à passer manuellement via le module "Bordereau de Transfert" ou "Ajout de Transaction": Débit: 588 ou 591 par Crédit 571 "petite caisse"  alors que ce transfert devait normalement généré automatiquement deux écritures: Dt 588 ou 591 par Cr 571 et Dt 572 "grande caisse" par Cr 588 ou 591.

Je suggère donc la revision du code relatif au transfert d'argent pour la générer désormais les deux écritures automatiquement au risque d'omettre de passer la seconde et de surcharger le chargé de la grande caisse et d'avoir par effet d'entrainement, des comptes à soldes éronnés.
outil transfert d argent

lomamech:
- Modify the Simple Voucher module by adding a check box that appears only
  if the transaction type is transfer then calls a new service that
  triggers two transactions for transfers from one Treasury Account to
  another through a transfer account

Closes #1417


jniles:
@lomamech, this is a huge amount of work.

My understanding of transfers is that BHIMA models transfers incorrectly, according to OHADA. See https://github.com/IMA-WorldHealth/bhima-2.X/issues/2741.

See the following screen shot:
accountsfortransfer

The problem is that OHADA uses a different account for every combination of cashboxes, but in BHIMA, we limit you to a single transfer account.  To implement what OHADA wants, we need to implement a matrix of cashbox -> cashbox  with an account for each.  #2741 goes into more detail about it.

What are your thoughts?

lomamech:
Thank you @jniles for this clarification, this PR will remain pending until we go  implement a matrix of cashbox -> cashbox with an account for each

jniles:
As this has been open for a number of months, I'm closing it in favor of implementing https://github.com/IMA-WorldHealth/bhima-2.X/issues/2741 more directly.  We just landed the transaction types to do it (https://github.com/IMA-WorldHealth/bhima-2.X/pull/3455).

PR #2957, opened 2018-06-29
fix(balance): update balance calculations

jniles:
The PR rewrites the balance report.  Here's what it look like now:
balancereportnoew
_Fig 1: New Balance Presentation_

The following changes have been made:
 1. The `before` section is now replaced with the period 0 balance.
 2. The `movements` section is replaced with all movements up through the selected period.
 3. The `after` section is replace by `before` + `after`.
 4. The Income/Expense accounts are removed from the report.
 5. The title reflects the fact that the report only goes from January through the period selected.

### DIFF
--- a/server/controllers/finance/reports/balance/index.js
+++ b/server/controllers/finance/reports/balance/index.js
@@ -31,6 +31,8 @@ const DEFAULT_PARAMS = {
 };
 
 const TITLE_ID = 6;
+const INCOME_ID = 4;
+const EXPENSE_ID = 5;
 
 /**
  * @function document
@@ -38,6 +40,8 @@ const TITLE_ID = 6;
  * @description
  * This function renders the Balance report.  The balance report provides a view
  * of the balances of any account used during the period.
+ *
+ * NOTE(@jniles): This file corresponds to the "Balance Report" on the client.
  */
 function document(req, res, next) {
   const params = req.query;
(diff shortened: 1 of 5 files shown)


server/controllers/finance/reports/balance/index.js
```diff
@@ -78,57 +82,96 @@ function document(req, res, next) {
...
+        SUM(IF(p.number = 0, pt.debit - pt.credit, 0)) AS "before",
+        SUM(IF(p.number BETWEEN 1 AND ?, pt.debit - pt.credit, 0)) AS "during",
+        SUM(IF(p.number BETWEEN 1 AND ?, pt.debit, 0)) AS "during_debit",
+        SUM(IF(p.number BETWEEN 1 AND ?, pt.credit, 0)) AS "during_credit",
+        SUM(IF(p.number <= ?, pt.debit - pt.credit, 0)) AS "after"
       FROM period_total AS pt
+        JOIN period p ON pt.period_id = p.id
+        JOIN account ac ON pt.account_id = ac.id
       WHERE pt.fiscal_year_id = ?
+        AND ac.type_id NOT IN (${INCOME_ID}, ${EXPENSE_ID})
```
mbayopanda, line 126: Why do we have this restriction ? I was thinking that those values (incomes and expenses) become null automatically when data in period 0 of the fiscal year

jniles: Yes and no - the opening balance value is always 0, but this is for the movements throughout the year.  That's one advantage - since we never pass a transaction to make them 0 during the fiscal year, it allows us to always generate income/expense reports without worrying that the values become 0.

I just talked to Patrick about this.  We should hide these transactions if the fiscal year is closed, but we should keep it if the fiscal year is still open.  So, I need to add that code.

mbayopanda (commented):
I tried to close the fiscal year 2016, and look what i get for the balance 2017.

image

Data are duplicated

jniles:
Isn't fiscal year 2016 already closed?  That is like passing the "rapport" transaction twice, so it will definitely double the values!

I'm pretty sure the only fiscal year that is not closed at IMCK is 2017 and 2018.

mbayopanda:
@jniles but in BHIMA, all fiscal year are not closed it means we can still close them with fiscal year module

jniles:
@mbayopanda, I think that's a problem with the data migration.  For some reason, we didn't close the fiscal years.  Can you run:
```sql
UPDATE fiscal_year SET locked = 1 WHERE id < 3;  -- I think this is 2016
```
on the live server?

mbayopanda:
The totals sum is correct 👍 

We can improve this report by changing the column title `Mouvements du mois` => `Mouvements`.

How can we explain accounts starting with 9 : 

image

jniles:
Ah, the accounts starting with 9 are wrong.  They are because IMCK has non-unique account numbers and BHIMA 2.x requires them to be unique.  So, during data migration, we attempted to set the duplicate accounts to have "9" as a prefix if they are indeed duplicates.

You can try to look and see if we can remove the other `23140000` account if it is not used.  If it is ... we need to figure out how to combine them in some way.

Yeah, the "month movement" thing needs to change.

jniles:
@mbayopanda, I don't think I will have time to work on this since I leave tomorrow.  Could either you or @lomamech try to clean this up?

jniles:
Also, thank you for your reviews!

mbayopanda:
bors r+

Close #2962

jniles:
You figured it out!  Well done :)

PR #2960, opened 2018-06-29
Add toggle inline filter for stock registries

mbayopanda:
This PR add inline filters for stock registries


jniles (approved):
LGTM.

bors r+

jniles:
Round 2!

bors r+

PR #2961, opened 2018-06-30
Fix account statement registry

mbayopanda:
This PR fixes the account statement registry


client/src/modules/account_statement/account_statement.ctrl.js
```diff
@@ -158,13 +152,22 @@ function AccountStatementController(
...
+      </div>`,
+    visible              : true,
   }, {
     field                : 'hrEntity',
     displayName          : 'TABLE.COLUMNS.RECIPIENT',
     headerCellFilter     : 'translate',
     cellTemplate :
       `<div class="ui-grid-cell-contents">
-        <bh-reference-link ng-if="row.entity.hrRecord" reference="row.entity.hrRecord" />
+        <bh-reference-link ng-if="row.entity.hrEntity" reference="row.entity.hrEntity" />
```
jniles, line 170: Good catch 👍

server/controllers/finance/generalLedger/index.js
```diff
@@ -26,11 +27,44 @@ const PERIODS = [
...
+exports.findTransactions = findTransactions;
 exports.getAccountTotalsMatrix = getAccountTotalsMatrix;
 exports.getAccountTotalsMatrixAggregates = getAccountTotalsMatrixAggregates;
 
+/**
+ * @function findTransactions
+ * @description returns general ledger transactions
+ * @param {object} options
+ */
+function findTransactions(options) {
```
jniles, line 42: We should probably figure out what API the journal is using and just use that in the long run, but since we are short on time ... this is a good fix 👍

client/src/modules/account_statement/account_statement.ctrl.js
```diff
@@ -75,6 +75,11 @@ function AccountStatementController(
     filter               : { condition : filtering.filterByDate },
     editableCellTemplate : 'modules/journal/templates/date.edit.html',
     footerCellTemplate   : '<i></i>',
+  }, {
+    field              : 'description',
+    displayName        : 'TABLE.COLUMNS.DESCRIPTION',
+    headerCellFilter   : 'translate',
+    footerCellTemplate : '<i></i>',
```
jniles, line 82: I think the `footerCellTemplate` is a hack to override the totals in the footer if the column has `aggregation` on it.  So it shouldn't be needed here since there is no `aggregation` on this column.

jniles (approved):
Looks great to me!  Quite a few bug fixes here!

bors r+

PR #2963, opened 2018-06-30
Optimize debtor group invoices search

mbayopanda:
This PR optimize the search of debtor group invoices.

Optimization needed for convention that has a huge lot of invoices for paying them via journal voucher.

### DIFF
--- a/server/controllers/finance/debtors/groups/index.js
+++ b/server/controllers/finance/debtors/groups/index.js
@@ -272,8 +272,6 @@ function invoices(req, res, next) {
  * do we need to look at each individual invoice?
  */
 function loadInvoices(params) {
-  // cancelled transaction type
-  const CANCELED_TRANSACTION_TYPE = 10;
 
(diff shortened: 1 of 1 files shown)


server/controllers/finance/debtors/groups/index.js
```diff
@@ -314,9 +314,8 @@ function loadInvoices(params) {
     ) AS i
     JOIN project ON i.project_id = project.id
     JOIN entity_map ON i.entity_uuid = entity_map.uuid
-    WHERE i.uuid NOT IN (
```
jniles, line 317: This query does need to be optimized!

Can we just do
```sql
WHERE i.reversed <> 1;
```
?

mbayopanda: 👍 Much better 

NOT IN : 50seconds 
LEFT JOIN : 872ms
i.reversed <> 1 : 473ms

PR #2964, opened 2018-06-30
Optimize invoice find function

mbayopanda:
This PR optimize the invoice find function by forcing to use index on patient table

### DIFF
--- a/server/controllers/finance/patientInvoice.js
+++ b/server/controllers/finance/patientInvoice.js
@@ -235,6 +235,7 @@ function find(options) {
       user.display_name, invoice.user_id, invoice.reversed, invoice.edited
     FROM invoice
     JOIN patient FORCE INDEX(debtor_uuid) ON invoice.debtor_uuid = patient.debtor_uuid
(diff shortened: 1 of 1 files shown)


server/controllers/finance/patientInvoice.js
```diff
@@ -234,8 +234,7 @@ function find(options) {
       em.text AS patientReference, service.name as serviceName, proj.name AS project_name,
       user.display_name, invoice.user_id, invoice.reversed, invoice.edited
     FROM invoice
-    LEFT JOIN patient ON invoice.debtor_uuid = patient.debtor_uuid
```
jniles, line 237: I've never seen FORCE INDEX before.  From what I am reading, this looks like the right solution!

jniles (approved):
@mbayopanda does this speed up the query on your machine?  How long does it take now?

server/controllers/finance/patientInvoice.js
mbayopanda: @jniles it's now very fast... 460ms instead of 11seconds

jniles:
bors r+

Let's try it then!

PR #2965, opened 2018-07-01
feat(complex voucher)currency changing

jeremielodi:
This PR solve the currency change problem whatever is the selected date

### DIFF
--- a/client/src/modules/vouchers/complex-voucher.ctrl.js
+++ b/client/src/modules/vouchers/complex-voucher.ctrl.js
@@ -53,17 +53,21 @@ function ComplexJournalVoucherController(
 
   vm.currencyChange = function currencyChange(vCurrencyId) {
     const entCurrencyId = vm.enterprise.currency_id;
-    let exchange = {};
+    let exchangeRate = 1;
+    const currentDate = vm.Voucher.details.date;
+
     Rates.read(true).then(() => {
       if (vCurrencyId !== entCurrencyId) {
-        exchange = Rates.getCurrentExchange(vCurrencyId);
-        vm.currentCurrency = { id : vCurrencyId, rate : exchange.rate };
+        exchangeRate = Rates.getExchangeRate(vCurrencyId, currentDate);
+        vm.currentCurrency = { id : vCurrencyId, rate : exchangeRate };
         vm.gridOptions.data = vm.gridOptions.data.map(row => {
-          row.credit *= exchange.rate;
-          row.debit *= exchange.rate;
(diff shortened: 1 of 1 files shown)


client/src/modules/vouchers/complex-voucher.ctrl.js
```diff
@@ -53,17 +53,21 @@ function ComplexJournalVoucherController(
 
   vm.currencyChange = function currencyChange(vCurrencyId) {
     const entCurrencyId = vm.enterprise.currency_id;
-    let exchange = {};
+    let exchangeRate = 1;
+    const currentDate = vm.Voucher.details.date;
```
mbayopanda, line 62: 👍

client/src/modules/vouchers/complex-voucher.ctrl.js
```diff
@@ -53,17 +53,21 @@ function ComplexJournalVoucherController(
...
+
     Rates.read(true).then(() => {
       if (vCurrencyId !== entCurrencyId) {
-        exchange = Rates.getCurrentExchange(vCurrencyId);
-        vm.currentCurrency = { id : vCurrencyId, rate : exchange.rate };
+        exchangeRate = Rates.getExchangeRate(vCurrencyId, currentDate);
```
mbayopanda, line 66: What was the problem with the previous implementation, because i didn't see any difference between both implementation when i change dates

mbayopanda (commented):
LGTM

jeremielodi:
The date, the exchange rate is now selected by date

mbayopanda:
I tried with an old date which have defined exchange rate but i cannot see change related with the exchange rate of this old date

jeremielodi:
It works perfectly for me, you have to sexe how calculations are done. Try
a transaction of 10$ then change the currency

mbayopanda:
When we change the date, we still have the exchange rate with the daily rate instead of the given date rate.

If that is the behavior that you expect so i can merge it like this.

jniles:
In my opinion, this should set the exchange rate to the date of the voucher.  It doesn't make a lot of sense to me that a user would want an incorrect exchange rate.

jeremielodi:
It what I have done. const  currentDate=vm.Voucher.details.date

Le lun. 2 juil. 2018 08:44, Jonathan Niles <notifications@github.com> a
écrit :

mbayopanda:
The commit works perfectly, we have just to enhance the exchange rate list by displaying last exchange rate by date.

bors r+

PR #2972, opened 2018-07-10
refractor(patient invoice)set focus on quantity after selecting the code

### LINKED ISSUE #2837: Automatically go to "quantity" after "code" on Patient Invoice
The Patient Invoice module is hard to use (compared to 1.x) because after selecting the code in the grid, everything blurs.  An ideal way to improve this would be to have application select "quantity" after entering the "code" automatically.  This would allow a user to rapidly progress through the module.

jeremielodi:
closes #2837

### DIFF
--- a/client/src/modules/invoices/patientInvoiceForm.service.js
+++ b/client/src/modules/invoices/patientInvoiceForm.service.js
@@ -4,7 +4,7 @@ angular.module('bhima.services')
 PatientInvoiceFormService.$inject = [
   'PatientService', 'PriceListService', 'InventoryService', 'AppCache', 'Store',
   'Pool', 'PatientInvoiceItemService', 'bhConstants', 'ServiceService', '$q',
-  '$translate', 'NotifyService',
+  '$translate', 'NotifyService', '$timeout',
 ];
 
 /**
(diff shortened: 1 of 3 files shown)


client/src/modules/invoices/patientInvoiceForm.service.js
```diff
@@ -433,7 +433,10 @@ function PatientInvoiceFormService(
     if (angular.isDefined(price)) {
       item.applyPriceList(price);
     }
-
+    // set focus on quantity field
+    setTimeout(() => {
+      document.getElementById(item.uuid).focus();
+    }, 100);
```
mbayopanda: Could you use the angular version `$timeout`.

mbayopanda (commented):
LGTM

There is a comment on an Angularjs wrapper for setTimeout, which is more natural for an Angular application.

jniles (commented):
Code looks okay to me.  The change from `setTimeout` -> `$timeout` was a good one.  You might consider doing the same thing with `document` -> `$document`.

I'm curious how you plan to write a test for this?

jniles:
The behavior here works well.  👍 

I'll try to write a test for this.

jniles:
Alright, test added.  @mbayopanda can you give us a review?

mbayopanda (approved):
LGTM 👍 

bors r+

PR #2974, opened 2018-07-11
refractor(balance_sheet) OHADA balance sheet

jeremielodi:


mbayopanda (commented):
As a first implementation, the report looks good.

But i noticed that you've forgot to insert equity accounts in the report

server/controllers/finance/reports/balance_sheet/report.handlebars
```diff
@@ -1,158 +1,106 @@
...
+            </td>
+            <td>
+              <table class="table table-condensed table-report table-bordered">
+                <thead>
+                  <th>{{translate 'FORM.LABELS.ACCOUNT_NUMBER'}}</th>
+                  <th>{{translate 'FORM.LABELS.LABEL'}}</th>
+                  <th>{{translate 'FORM.LABELS.AMOUNT'}}</th>
+                </thead>
+                <tbody>
+                  {{#each this.liability}}
```
mbayopanda, line 58: Equity accounts must appear above liabilities.

server/controllers/finance/reports/balance_sheet/index.js
```diff
@@ -1,240 +1,153 @@
...
+  let docReport;
+  const options = _.extend(req.query, {
+    filename : 'TREE.OPERATING_ACCOUNT',
+    csvKey : 'rows',
+    user : req.session.user,
+  });
 
   try {
-    report = new ReportManager(TEMPLATE, req.session, params);
+    docReport = new ReportManager(TEMPLATE, req.session, options);
```
mbayopanda, line 33: A curiosity question, is there any reason to rename this variable to `docReport` ?

server/models/Update/equity_accounts.sql
```diff
@@ -0,0 +1,2 @@
+ UPDATE account SET type_id = 3
+ WHERE number LIKE '1%' AND type_id <> 6;
```
mbayopanda, line 2: Could you do the same operation with data for assets and liabilities, to be sure that all accounts have the correct type.

It is important for correctly test what we have in our Bilan

mbayopanda (commented):
It's look much better.

Could you fix the problem with tests

server/models/Update/equity_accounts.sql
jeremielodi: I need the assets and liabilities conditions for updating those accounts

mbayopanda: You can use these document for seeing which account is for assets (actifs) and liabilities (passifs)

## Assets : 
actifs

## Liabilities
passifs

jniles:
As discussed, closing this PR.

PR #2980, opened 2018-07-13
Fix operating account

### LINKED ISSUE #2979: Problem with the operating account report
The operating account report shows hidden and blocked accounts and the balance of this report is not well calculated.

lomamech:
- Remove not necessary column Balance in Operating Report
- Use Charge and Profit instead Expense and Income
- use Signe plus because totalExpense have negative (-) * (-) give (+), In
  this case the good thing is to Use (+),
- Prevent the Using of locked Account for Operating Account Report

Closes #2979


jeremielodi (commented):
everything looks perfect but you can if necessary rename also some variables such as ```Expense, Income, totalExpense, totalIncome```

jeremielodi (commented):
@jeremielodi approved this PR

PR #2981, opened 2018-07-18
Write into the general ledger when closing a fiscal year

mbayopanda:
In this PR we write also in the voucher and general_ledger during the closing fiscal year process.

We have updated `CloseFiscalYear` procedure and created a new one `PostGeneralVoucher`.


jniles:
Please don't do this.  It makes it impossible to close and reopen and correct issues with previous balances.  We had a really difficult time migrating 1.x's database since multiple transactions closed a fiscal year.

What is wrong with just writing into `period_total`?  What errors are you running into with that?

mbayopanda:
@jniles we still write into the `period_total`, for a need of proof we must see transaction which sold exploitation accounts, so for having that we create a voucher for that which write directly into the general ledger.

The previous code did nothing as transaction, it did just a movement balance accounts into the period 0 of the next year and let exploitation account untouched.

jniles:
@mbayopanda we don't write to the `general_ledger` on purpose.  See https://github.com/IMA-WorldHealth/bhima-2.X/issues/1651 for the discussion about why we only use the `period_total` table.

What kind of proof do we need?  Who is asking for the proof?  We could make an interface at the end of the closing process to print that out of we need to, but writing to the General Ledger is problematic for the reasons discussed in the previous issue.

mbayopanda:
@jniles once again changes that you made are outstanding, so we still continue with your logic with period zero.

@jniles could you look at the procedure that i made, feel free to comment and give your point of vue.

Thank you very much for your feedback

PR #2987, opened 2018-07-27
Accounts References Report

mbayopanda:
This PR add a report for accounts references.


server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,184 @@
...
+  const queryAccounts = `
+    SELECT includeTable.account_id, includeTable.account_number FROM (
+      SELECT DISTINCT 
+        account.id AS account_id, account.number AS account_number, t.is_exception FROM account
+        JOIN (
+          SELECT a.id, a.number, ar.is_amo_dep, ari.is_exception FROM account a 
+          JOIN account_reference_item ari ON ari.account_id = a.id
+          JOIN account_reference ar ON ar.id = ari.account_reference_id
+          WHERE ar.abbr LIKE ? AND ar.is_amo_dep = ? AND ari.is_exception = 0
+        ) AS t ON LEFT(account.number, CHAR_LENGTH(t.number)) LIKE t.number
```
jniles: Why are we joining on `LEFT account.number`?  Can't we make this an account type?

mbayopanda: Because we want to have all accounts from the account table starting with account number of accounts in account_reference_item table

mbayopanda:
@jniles could review this PR please.

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,184 @@
...
+  // get all references
+  const queryAccountReferences = `
+    SELECT id, abbr, description, is_amo_dep FROM account_reference;
+  `;
+
+  return db.one(queryFiscalYear, [periodId])
+    .then(fiscalYear => {
+      glb.fiscalYear = fiscalYear;
+
+      return db.exec(queryAccountReferences);
```
jniles: It doesn't look like this query depends on the previous query.  You could speed it up by combining both queries and wrapping them in `Q.all()`.

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,184 @@
...
+    })
+    .then(accountReferences => {
+      const dbPromises = accountReferences.map(ar => {
+        return getValueForReference(
+          ar.abbr,
+          ar.is_amo_dep,
```
jniles, line 47: What is `is_amo_dep`?

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,184 @@
...
+      const accountIds = accounts.map(a => a.account_id);
+      const parameters = [
+        abbr,
+        isAmoDep,
+        referenceDescription,
+        fiscalYearId,
+        periodNumber,
+        accountIds.length ? accountIds : null,
+      ];
+      return db.exec(queryTotals, parameters);
```
jniles: Your function is called `getValueForReference`, but it returns an array!  It's clear that all of the functions that call this one want a single value, could you either do:
```js
const [value] = await db.exec(queryTotals, parameters);
return value;
```
...or...
```js
return db.exec(queryTotals, parameters).then(values => values[0]);
```

It would probably make this function easier to use.

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,184 @@
...
+ * Returns all accounts concerned by the reference without exception accounts
+ *
+ * @param {string} abbr - the reference of accounts. ex. AA or AX
+ * @param {boolean} isAmoDep - the concerned reference is for amortissement, depreciation or provision
+ */
+function getAccountsForReference(abbr, isAmoDep = 0) {
+  /**
+   * Get the list of accounts of the reference without excepted accounts
+   * mysql implementation of minus operator
+   * @link http://www.mysqltutorial.org/mysql-minus/
```
jniles: This is a nice comment, but the MINUS operator is never used.

server/controllers/finance/accounts/references.compute.js
jniles: Can we use `=` instead of `LIKE`?

Instead of linking the account number, would it be possible to use the `account_id`?  Then we could have a foreign key relationship which would be more reliable than parsing numbers, I think.

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,184 @@
...
+   */
+  const queryAccounts = `
+    SELECT includeTable.account_id, includeTable.account_number FROM (
+      SELECT DISTINCT 
+        account.id AS account_id, account.number AS account_number, t.is_exception FROM account
```
jniles: You do not use `t.is_exception` in the `includeTable`.  You only pull out `account_id` and `account_number`.  Do you need `t.is_exception` in the query?

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,184 @@
...
+        JOIN (
+          SELECT a.id, a.number, ar.is_amo_dep, ari.is_exception FROM account a 
+          JOIN account_reference_item ari ON ari.account_id = a.id
+          JOIN account_reference ar ON ar.id = ari.account_reference_id
+          WHERE ar.abbr LIKE ? AND ar.is_amo_dep = ? AND ari.is_exception = 0
```
jniles: Shouldn't a `LIKE` cal be surrounded by `%`?  Otherwise, it should be an `=`, right?

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,184 @@
...
+ * @param {boolean} isAmoDep - the concerned reference is for amortissement, depreciation or provision
+ */
+function getAccountsForReference(abbr, isAmoDep = 0) {
+  /**
+   * Get the list of accounts of the reference without excepted accounts
+   * mysql implementation of minus operator
+   * @link http://www.mysqltutorial.org/mysql-minus/
+   */
+  const queryAccounts = `
+    SELECT includeTable.account_id, includeTable.account_number FROM (
```
jniles: This is a really complicated query.  Can you walk me through the logic we are trying to get?  It looks like we could simplify it, but I'm not sure what the goal is.

server/controllers/finance/accounts/references.js
```diff
@@ -12,13 +12,11 @@
...
  * @requires db
- * @requires NotFound
- *
- * @todo HANDLE ACCOUNT REFERENCE ITEMS
+ * @requires util
  */
-const Q = require('q');
 const util = require('../../../lib/util');
 const db = require('../../../lib/db');
+const compute = require('./references.compute');
```
jniles, line 19: Moving large queries into a separate file is a good implementation step 👍

server/controllers/finance/reports/account_reference/index.js
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @requires lodash
+ * @requires lib/db
+ * @requires lib/ReportManager
+ * @requires lib/errors/BadRequest
+ */
+
+const _ = require('lodash');
+const AccountReference = require('../../accounts').references;
+// const Tree = require('../../../../lib/Tree');
```
jniles: Go ahead and remove this line - it doesn't like you ever use it.

server/controllers/finance/reports/account_reference/index.js
```diff
@@ -0,0 +1,76 @@
...
+    WHERE p.id = ?;
+  `;
+
+  db.one(getFiscalYearSQL, [params.period_id])
+    .then(period => {
+      _.merge(context, { period });
+      return AccountReference.computeAllAccountReference(params.period_id);
+    })
+    .then(data => {
+      _.merge(context, { data, currencyId : req.session.enterprise.currency_id });
```
jniles: Just FYI - the `currency_id` is bound to `metadata` in the view along with all other `session` parameters.  You can see that here: https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/ReportManager.js#L115

So you could easily remove this and change all `../currency_id` calls to `../metadata.enterprise.currency_id`

jniles (commented):
I haven't run the code yet to see how it works, but I've made some code comments to follow up on my previous comments.  There are a few decisions made in the SQL that aren't clear to me why they were made.

The steps you took to refactor the computations into their own files are a good improvement.

Let me know your thoughts!

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,179 @@
...
+          ar.abbr,
+          ar.is_amo_dep,
+          ar.description,
+          glb.fiscalYear.period_number,
+          glb.fiscalYear.id
+        );
+      });
+      return Q.all(dbPromises);
+    })
+    .then(data => data);
```
jniles: You could remove this `.then()`.  It doesn't do anything.  It's like writing:

```js
function ident(obj) {
  return obj;
}

ident({ id : 1 }) // returns { id : 1}
ident(ident(ident({ id : 1 }))) // returns { id : 1 }
```

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,179 @@
...
+ * @param {string} abbr - the reference of accounts. ex. AA or AX
+ * @param {number} periodId - the period needed
+ * @param {boolean} isAmoDep - the concerned reference is for amortissement, depreciation or provision
+ */
+function computeSingleAccountReference(abbr, isAmoDep = 0, periodId) {
+  const glb = {};
+
+  // get fiscal year information for the given period
+  const queryFiscalYear = `
+    SELECT fy.id, p.number AS period_number FROM fiscal_year fy
```
jniles: You use the `queryFiscalYear` SQL twice.  It might be useful to make that its own function.  Like:
```js
function getFiscalYearByPeriodId(periodId) {
  const sql =  `
    SELECT fy.id, p.number AS period_number FROM fiscal_year fy
    JOIN period p ON p.fiscal_year_id = fy.id
    WHERE p.id = ?
  `;
  return db.one(sql, [periodId];
}
```

Then you could call it wherever you needed it.  Eventually we would refactor it into either `fiscal.js` or `period.js`.

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * @description
+ * Returns all accounts concerned by the reference without exception accounts
+ *
+ * @param {string} abbr - the reference of accounts. ex. AA or AX
+ * @param {boolean} isAmoDep - the concerned reference is for amortissement, depreciation or provision
+ */
+function getAccountsForReference(abbr, isAmoDep = 0) {
+  /**
+   * Get the list of accounts of the reference without excepted accounts
```
jniles, line 142: You should remove this MINUS comment if you never use MINUS.

server/controllers/finance/accounts/references.compute.js
```diff
@@ -0,0 +1,179 @@
...
+   * mysql implementation of minus operator with left join
+   * @link http://www.mysqltutorial.org/mysql-minus/
+   */
+  const queryAccounts = `
+    SELECT includeTable.account_id, includeTable.account_number FROM (
```
jniles: I wonder if we do the filtering on `is_exception` as the last step, if that would simplify the query ... I can try and work on this in my spare time and see if I can come up with something.

jniles (approved):
Changes look fine to me.  This can be merged whenever you are ready for it.

bors delegate+

Issue #2747, opened 2018-05-10
Remove the "ID" column from the invoicing fees ui-grid


jniles:
The Invoicing Fees ui-grid has an ID column corresponding to the internal database ID.  The user has no control over the id, and doesn't need to see it.   Think of it this way: should we show uuids to the client?  It should be removed.

![invoicingfeesid](https://user-images.githubusercontent.com/896472/39866765-e4a6c194-5449-11e8-8ccb-fb164ac95555.PNG)
_Fig 1: This column must go_

Scholldier:
I think I've got this one. I simply remove the following code from /bhima-2.X/client/src/modules/invoicing-fees

{
field : 'id',
displayName : 'TABLE.COLUMNS.ID',
headerCellFilter: 'translate',
width: 45,
},

Question is, should I also remove the bs.id value from the select statement in /bhima-2.X/server/controllers/finance ?

Also, Jonathan, how do I commit this change?

jniles:
Hey @Scholldier,

Thanks for helping!  You are right, that change would do the trick!

The "id" is assuredly used in multiple places internally to track changes.  Passing around the id to look up is more efficient than passing the whole record on the client, and we try to use that pattern on the client.  It would be good to keep it around in the SQL to facilitate this.

To commit the change, the best option you have is to clone the repository, make the change locally, then make a git commit, push to your clone on github, then open a Pull Request (PR) from your forked repository to this one.  Your PR would be reviewed by one of the team, automatically tested online, and then finally merged into the main repository.  Afterwards, we rinse and repeat!

[Here is an article](https://akrabat.com/the-beginners-guide-to-contributing-to-a-github-project/) that could help in a step by step process for forking, cloning and contributing.  Let me know if you need additional guidance!

Scholldier:
Thanks Jonathan!

I tried building a fork / pull request directly from the GitHub site, but I don't think I did it right as two of the checks failed. I'll re-read the article you sent more in-depth later on and try to issue the pull request I have locally from my test install.

jniles:
Of course.  You actually did exactly right, but it looks like an end-to-end test or two of ours depends on the displayed "ID" column.  This is wrong from a testing perspective - the end-to-end tests should reflect user behavior and we users will always refer to the invoicing fees by their _name_ not their _ID_.

We'll have to update those tests.  Let me see if I can pull your code and figure out the easiest way to do that and walk you through it?

Scholldier:
[...]

PR #2988, opened 2018-07-28
Update invoicing-fees.ctrl.js for Issue 2747

Scholldier:
This pull request is for issue #2747 to remove the ID column from the invoicing fees UI grid. I removed the following code from /bhima-2.X/client/src/modules/invoicing-fees/invoicing-fees.ctrl.js

{
field : 'id',
displayName : 'TABLE.COLUMNS.ID',
width: 45,
},

### DIFF
--- a/client/src/modules/invoicing-fees/invoicing-fees.ctrl.js
+++ b/client/src/modules/invoicing-fees/invoicing-fees.ctrl.js
@@ -24,11 +24,6 @@ function InvoicingFeesController($state, InvoicingServices, Notify, bhConstants,
   vm.toggleFilter = toggleFilter;
 
   columns = [{
-    field : 'id',
-    displayName : 'TABLE.COLUMNS.ID',
-    headerCellFilter: 'translate',
-    width: 45,
-  }, {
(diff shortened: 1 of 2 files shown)


jniles:
@Scholldier, it turns out it was much simpler than I thought, so I just updated it directly.  You can look at my change directly - it's https://github.com/IMA-WorldHealth/bhima-2.X/pull/2988/commits/04d5392ba151677234ae718d38ca0a4ea48416be#diff-3b1bdebcb35d86356cdeec54efcbda53L38.

The test clicks on the "action" column, which is the last column of the grid at index 6.  Since you removed a column, so we needed to update the column index to 5.  Now that it is done, tests _should_ pass ... although we have encountered some flakey-ness in the past.

jniles:
Looks like we got it solved.

@mbayopanda could you give @Scholldier a review?

jniles (approved):
Finally fixed the PID issue!

bors r+

jniles:
Thanks so much @Scholldier!  Great contribution!

PR #3000, opened 2018-08-06
feat(general ledjer) toggle hide title accounts

### LINKED ISSUE #2947: Allow users to hide title accounts in the General Ledger
The General Ledger provides a nice grouping via title accounts.  However, it is sometimes not obvious how to sum these accounts, since the sums are repeated on the headers.  To aid this quick and dirty analysis, we should support hiding the title accounts via a flag to filter them out.

jeremielodi:
closes #2947

### DIFF
--- a/client/src/modules/general-ledger/general-ledger.ctrl.js
+++ b/client/src/modules/general-ledger/general-ledger.ctrl.js
@@ -22,6 +22,9 @@ function GeneralLedgerController(
   const cacheKey = 'GeneralLedger';
   let columns = [];
   vm.year = {};
+  vm.accounts = [];
+  vm.toggleHideTitleAccount = toggleHideTitleAccount;
+  vm.hideTitleAccount = false;
   const fields = [
     'balance',
     'balance0',
@@ -46,6 +49,7 @@ function GeneralLedgerController(
   vm.aggregates = {};
 
   vm.indentTitleSpace = 15;
+  const isNotTitleAccount = (account) => account.type_id !== bhConstants.accounts.TITLE;
 
   const tmpl = `
     <div class="ui-grid-cell-contents">
@@ -162,12 +166,29 @@ function GeneralLedgerController(
     });
   }
 
+  // specify if titles accounts should be hidden
+  function toggleHideTitleAccount() {
+    vm.hideTitleAccount = !vm.hideTitleAccount;
+    hideTitles();
+  }
+
+  // Hide when possible title account
+  function hideTitles() {
+    if (vm.hideTitleAccount) {
+      vm.gridOptions.data = vm.accounts.filter(isNotTitleAccount);
+    } else {
+      vm.gridOptions.data = vm.accounts;
+    }
+  }
+
   function loadData(accounts = []) {
     accounts.forEach(preProcessAccounts);
     Accounts.order(accounts);
 
     renameGidHeaders(vm.year);
     vm.gridOptions.data = accounts;
+    vm.accounts = accounts;
+    hideTitles();
   }
 
   vm.download = GeneralLedger.download;
--- a/client/src/modules/general-ledger/general-ledger.html
+++ b/client/src/modules/general-ledger/general-ledger.html
@@ -65,6 +65,17 @@
           ng-class="{ 'btn-info' : GeneralLedgerCtrl.filterEnabled }">
           <i class="fa fa-filter"></i>
         </button>
+        <button
+          type="button"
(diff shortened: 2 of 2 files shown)


client/src/modules/general-ledger/general-ledger.html
```diff
@@ -65,6 +65,14 @@
           ng-class="{ 'btn-info' : GeneralLedgerCtrl.filterEnabled }">
           <i class="fa fa-filter"></i>
         </button>
+        <button
+          type="button"
+          ng-click="GeneralLedgerCtrl.toggleHideTitlesAccounts()"
+          data-method="hide_titles"
+          class="btn btn-default"
+          ng-class="{ 'btn-info' : GeneralLedgerCtrl.hideTitlesAccounts }">
+          <i class="fa fa-eye-slash"></i>
```
jniles: It might be nice to have the `ng-class` here toggle between `fa-eye` (if title accounts are visible) and `fa-eye-slash` (if title accounts are hidden).

client/src/modules/general-ledger/general-ledger.ctrl.js
```diff
@@ -330,10 +331,28 @@ function GeneralLedgerController(
...
+  function toggleHideTitlesAccounts() {
+    vm.hideTitlesAccounts = !vm.hideTitlesAccounts;
+    hideTitles();
+  }
+
+  // Hide when possible title accounts
+  function hideTitles() {
+    if (vm.hideTitlesAccounts) {
+      vm.gridOptions.data = vm.accounts.filter(t => {
+        return t.type_id !== 6;
```
jniles: The `6` is the ID of the title account type, but that is unclear.  Thankfully, we have put this mapping in [`bhConstants`](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/constants/bhConstants.js#L16).  Please use the following:
```js
vm.gridOptions.data = vm.accounts.filter(account => account.type_id !== bhConstants.accounts.TITLE);
```

client/src/modules/general-ledger/general-ledger.ctrl.js
```diff
@@ -20,7 +20,8 @@ function GeneralLedgerController(
 ) {
   const vm = this;
   const cacheKey = 'GeneralLedger';
-
+  vm.accounts = [];
+  vm.toggleHideTitlesAccounts = toggleHideTitlesAccounts;
```
jniles: The `Titles` doesn't need to be plural.  You can call it `hideTitleAccounts`.

client/src/modules/general-ledger/general-ledger.ctrl.js
jniles: You could actually make the `filter()` function into its own function for a performance boost:

```js
const isNotTitleAccount = (account) => account.type_id !== bhConstants.accounts.TITLE;

function hideTitles() {
  // ... etc ...
  vm.gridOptions.data = vm.accounts.filter(isNotTitleAccount);

}
```

The advantage here is that we don't make the function `isNotTitleAccount` every time `hideTitles()` is called.

.env.development
```diff
@@ -3,9 +3,9 @@ PORT=8080
 
 # database configuration
 DB_HOST='localhost'
-DB_USER='bhima'
+DB_USER='bhima12'
```
jniles: This should probably be changed back.  I'm actually surprised it passes tests.

client/src/modules/general-ledger/general-ledger.ctrl.js
```diff
@@ -330,10 +332,27 @@ function GeneralLedgerController(
...
+  // specify if titles accounts should be hidden
+  function toggleHideTitleAccount() {
+    vm.hideTitleAccount = !vm.hideTitleAccount;
+    hideTitles();
+  }
+
+  // Hide when possible title account
+  function hideTitles() {
+    if (vm.hideTitleAccount) {
+      const isNotTitleAccount = (account) => account.type_id !== bhConstants.accounts.TITLE;
```
jniles: For performance reasons, you should declare this function outside of the `hideTitles()` function.  We don't need to re-create it every single time that we call `hideTitles()`.

jniles (commented):
The `.env.development` changes should be reverted.

jeremielodi:
@jniles, This PR can be reviewed

.env.development
jniles: The DB_USER hasn't changed.  Could you change it?

jniles (approved):
LGTM.

bors r+

PR #3001, opened 2018-08-06
Improve payroll modules

### LINKED ISSUE #2968: Improved data Sending mode in the Payroll Multiple module
- Improved data Sending mode in the Payroll Multiple module to increase server performance, payroll billing, payment setup, payroll commitment and other operations when selecting employees individually on the Payroll Multiple grid

- Add condensed pay slips, payroll taxes report

lomamech:
- Improve the way to send selected employee, send a array who contains the
  reference of selected Employees
- Deleting the request that runs in a loop,
- Implementation of function commitment who manage Data that will be used
  to execute the transaction to support the employees
- Update Integration Test for Payroll, Update sending data for Make
  Commitment
- Update payslip Generator API, Improve the request Perfomance
- Update Payslip Design of Payslip
- Improve the generated of report of Payroll
- Add two report of Payroll Report and Report of social Charge
- Add new function commitmentFunction for Generate data by Employees to
  make Commitment and Split the function Commitments
- Improved Payroll module, added the ability to display payroll bulletins
reports configured in multiple currencies into a single currency,
- Correction of conversion errors

Closes #2968


lomamech:
@jniles

client/src/modules/multiple_payroll/multiple_payroll.js
```diff
@@ -256,6 +266,47 @@ function MultiplePayrollController(
     let invalid = false;
     let employeeStatusId;
 
+    // get All Employees Reference
+    const employeesRef = employees.map(emp => emp.reference);
+
+    if (employees.length) {
```
jniles: You probably don't need to check `employees.length` here.  `employees.forEach()` will skip an empty array.

client/src/modules/multiple_payroll/multiple_payroll.js
```diff
@@ -256,6 +266,47 @@ function MultiplePayrollController(
...
+      employees.forEach(employee => {
+        employeeStatusId = parseInt(employee.status_id, 10);
+
+        if (employeeStatusId === 1) {
```
jniles: It looks like you want to conditionally set `invalid` if some `employeeStatusId` is 1.  A clearer way to write this is to use [`Array.prototype.some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some).  It will return true if a condition is met for one element in the array.  Like this:

```js
// returns true if one employee has status_id of 1
const invalid = employees.some(employee => parseInt(employee.status_id, 10) === 1);
```

It would probably make it clearer to rename the function something that other programmers can read.  Like:
```js
const isAwaitingPayment =employee => parseInt(employee.status_id, 10) === 1;
const invalid = employees.some(employee => isAwaitingPayment(employee));
// alternatively: const invalid = employees.some(isAwaitingPayment);
```

client/src/modules/multiple_payroll/multiple_payroll.js
```diff
@@ -256,6 +266,47 @@ function MultiplePayrollController(
...
+  /*
+   * The PayrollReport function allows to display two reports related to the Payroll,
+   * the first to display a condensed report of the bulletins of payrolls of the employees
+   * and the second a report of the payroll taxes on remuneration (Charges of the company)
+  */
+  vm.viewPaySlipReport = function viewPaySlipReport(socialCharge) {
+    const employees = vm.gridApi.selection.getSelectedRows();
+
+    // get All Employees Reference
+    const employeesRef = employees.map(emp => emp.reference);
```
jniles: Is it possible for employees to share references?  In most modules (cash, invoice, patient, etc), we assume references are shared between projects.  Do employees behave the same way?

server/controllers/finance/exchange.js
```diff
@@ -143,3 +144,29 @@ exports.delete = function del(req, res, next) {
     .catch(next)
     .done();
 };
+
+// This query returns the current exchange rate of all currencies
+function getCurrentExchangeRateByCurrency(date) {
```
jniles: This function is called `getCurrentExchangeRateByCurrency()`, but it doesn't take a currency.  It should be renamed to something that reflects what it actually does.

server/controllers/finance/exchange.js
```diff
@@ -143,3 +144,29 @@ exports.delete = function del(req, res, next) {
...
+  const exchangeDate = date || new Date();
```
jniles: By the way, in the newest JavaScript, you can set default parameters for functions with a new syntax.  Instead of:
```js
function getCurrentExchangeRateByCurrency(date) {
  const exchangeDate = date || new Date();
  /// etc...
```
... you can do the same thing with this syntax....
```js
function getCurrentExchangeRateByCurrency(date = new Date()) {
  // now use `date` instead of `exchangeDate`
```

server/controllers/payroll/multiplePayroll/makeCommitment.js
```diff
@@ -6,274 +6,103 @@
...
+
+  configurationData.find(options)
+    .then(dataEmployees => {
+      data.employees = dataEmployees;
+
+      const queries = q.all([
+        db.exec(sqlGetRubricPayroll, [referenceEmployees, payrollConfigurationId]),
+        db.exec(sqlGetRubricConfig, [payrollConfigurationId]),
+        db.exec(sqlGetAccountPayroll, [payrollConfigurationId]),
+        Exchange.getCurrentExchangeRateByCurrency(),
```
jniles: This function may return multiple exchange rates.  Why?  Shouldn't we use only one?

server/controllers/payroll/multiplePayroll/makeCommitment.js
```diff
@@ -6,274 +6,103 @@
...
+      ]);
+
+      return queries;
+    })
+    .spread((rubricsEmployees, rubricsConfig, account, exchangeRates) => {
```
jniles: Cool use of `.spread()` 👍

server/controllers/payroll/multiplePayroll/setConfiguration.js
```diff
@@ -94,17 +94,23 @@ function config(req, res, next) {
...
+          holiday.label,
+          holidayValue,
+        ]);
       });
 
       /*
      * Recalculation of base salary on the basis of any holiday or vacation period,
      * where the percentages are respectively equal to 100% of the basic salary will
-     * remain equal to that defined at the level of the grade table BB
+     * remain equal to that defined at the level of the Holiday table
```
jniles, line 110: 💃

server/controllers/payroll/multiplePayroll/setMultiConfiguration.js
```diff
@@ -7,16 +7,20 @@
...
  */
 const db = require('../../../lib/db');
+const Exchange = require('../../finance/exchange');
 const q = require('q');
 const payrollSettings = require('./payrollSettings');
 
 function config(req, res, next) {
-  const dataEmployees = req.body.data;
+  const dataEmployees = req.body.data.employees;
+  const currencyId = req.body.data.currencyId;
```
jniles: Just FYI, you can rewrite this as:
```js
const { employees, currencyId } = req.body.data;
```

You'd have to rename `dataEmployees` as `employees` however.

server/controllers/payroll/reports/payrollReportSocialCharge.handlebars
```diff
@@ -0,0 +1,71 @@
...
+          </thead>
+
+          <tbody>
+            {{#each dataEmployees}}
+              <tr class="text-capitalize">
+                <td style="text-align: left;"> {{add @index 1}} </td>
+                <td style="text-align: left;"><strong> {{display_name}} </strong></td>
+                {{#each rubricsChargeEnterprise}}
+                  <td style="text-align: right;">
+                    <div>
```
jniles: Do you need these extra `<div>` and `<span>`s?

server/controllers/payroll/reports/payrollReportSocialCharge.handlebars
```diff
@@ -0,0 +1,71 @@
...
+                      <span >
+                        {{#if result}} 
+                          {{currency (multiply result ../../payrollPeriod.exchangeRate) ../../payrollPeriod.currency}}
+                        {{/if}}
+                      </span>
+                    </div>
+                  </td>
+                {{/each}}
+                <td>
+                  {{currency (multiply somChargeEnterprise ../payrollPeriod.exchangeRate) ../payrollPeriod.currency}}
```
jniles: Is there anyway to compute these values in JS instead of in the Handlebars template?

server/controllers/payroll/reports/payrollReportSocialCharge.handlebars
```diff
@@ -0,0 +1,71 @@
...
+                <th><div><span >{{ abbr }}</span></div></th>
+              {{/each}}
+              <th style="width: 4%;"><div><span >{{translate 'TABLE.COLUMNS.TOTAL' }}</span></div></th>
+            </tr>
+          </thead>
+
+          <tbody>
+            {{#each dataEmployees}}
+              <tr class="text-capitalize">
+                <td style="text-align: left;"> {{add @index 1}} </td>
```
jniles: It doesn't look like `text-align:left` will actually do anything here.  Do you need to have it?

jniles: Also, is this data sorted? I worry that the `@index` will change depending on the sort order of the data. Can we use the employee `reference` instead?

server/controllers/payroll/multiplePayroll/makeCommitment.js
```diff
@@ -6,274 +6,103 @@
...
+  const options = {
+    payroll_configuration_id : payrollConfigurationId,
+    reference : referenceEmployees,
+  };
+
+  configurationData.find(options)
+    .then(dataEmployees => {
+      data.employees = dataEmployees;
+
+      const queries = q.all([
```
jniles: Instead of declaring `queries` and immediately returning it, it makes more sense to me to immediately `return q.all([])`.

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -1,56 +1,57 @@
...
+
+  if (options.currency && options.socialCharge) {
+    TEMPLATE = templateSocialCharge;
+    orientation = 'portrait';
+    options.footerRight = footerRight;
+
+  } else if (options.currency && !options.socialCharge) {
+    TEMPLATE = templatePayrollReport;
+
+    options.footerRight = footerRight;
```
jniles: You repeat `options.footerRight` several times here.  Could you set a default set of options, then modify them if they need to be modified?  Like:
```js

const DEFAULT_OPTS = {
  orientation : 'landscape',
  footerRight : '[page] / [toPage]',
  filename : 'FORM.LABELS.PAYSLIP',
  // all other default opts...
};

function build(...) {
  _.extend(options, DEFAULT_OPTS);

  if (options.currency_id && !options.socialCharge) {
    options.orientation = 'portrait';
  }
}
```

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -64,153 +65,171 @@ function build(req, res, next) {
...
-  const queriesHolidays = q.all(getHolidays);
-
-  const queriesOffDays = q.all(getOffDays);
+  PayrollConfig.lookupPayrollConfig(options.idPeriod)
+    .then(payrollPeriod => {
+      data.payrollPeriod = payrollPeriod;
+      return Exchange.getExchangeRate(data.enterprise.id, options.currency, new Date(data.payrollPeriod.dateTo));
+    })
+    .then(exchange => {
+      data.payrollPeriod.exchangeRate = parseInt(options.currency, 10) === parseInt(data.enterprise.currency_id, 10) ?
```
jniles: Is the `parseInt()` here really necessary? Does the database give you strings for the id?

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -64,153 +65,171 @@ function build(req, res, next) {
...
+    .then(dataEmployees => {
+      // Set Aggregate of Rubrics
+      let totalNetSalary = 0;
+      let totalBasicSalary = 0;
+      let totalBaseTaxable = 0;
+      let totalGrossSalary = 0;
+
+      dataEmployees.forEach(employee => {
+        data.exchangeRatesByCurrency.forEach(exchange => {
+          employee.net_salary_equiv = parseInt(exchange.currency_id, 10) === parseInt(employee.currency_id, 10) ?
```
jniles: Man ... if you are doing `parseInt() === parseInt()` every line, let's just do it once:
```js
const exchangeCurrencyId = parseInt(exchange.currency_id, 10);
const employeeCurrencyId = parseInt(employee.currency_id, 10);
const isSameCurrency = exchangeCurrencyId  === employeeCurrencyId ;

employee.net_salary_equiv = isSameCurrency ? employee.net_salary / exchange.rate : employee.net_salary;
```

Even better, we could set the value to 1 if the same currency:
```js
const exchangeCurrencyId = parseInt(exchange.currency_id, 10);
const employeeCurrencyId = parseInt(employee.currency_id, 10);
const isSameCurrency = exchangeCurrencyId  === employeeCurrencyId ;

// set the denominator to 1 if they are the same currency.
const base = isSameCurrency ? exchange.rate : 1;

employee.net_salary_equiv =  employee.net_salary / base;
employee.daily_salary_equiv = employee.daily_salary / base;
// etc...
```

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -64,153 +65,171 @@ function build(req, res, next) {
...
+        employee.somRubNonTaxable = somRubNonTaxable;
+        employee.somChargeEnterprise = somChargeEnterprise;
+        employee.somChargeEmployee = somChargeEmployee;
+        data.exchangeRatesByCurrency.forEach(exchange => {
+          employee.somRubTaxable_equiv = parseInt(exchange.currency_id, 10) === parseInt(employee.currency_id, 10) ?
+            employee.somRubTaxable / exchange.rate : employee.somRubTaxable;
+          employee.somRubNonTaxable_equiv = parseInt(exchange.currency_id, 10) === parseInt(employee.currency_id, 10) ?
+            employee.somRubNonTaxable / exchange.rate : employee.somRubNonTaxable;
+          employee.somChargeEnterprise_equiv =
+            parseInt(exchange.currency_id, 10) === parseInt(employee.currency_id, 10) ?
```
jniles: You could do a similar thing to that `parseInt()` trick here.

jniles (commented):
@lomamech, this was a lot of changes, and I'm not sure I've gotten all of them.  Let me know when you are done with this review and I'll make an effort to take a look at them again.

client/src/modules/multiple_payroll/multiple_payroll.js
lomamech: @jniles it's just to warn the user that he has not selected any Employee
````
 if (employees.length) {
 ....
} else {
      Notify.danger('FORM.WARNINGS.NO_EMPLOYE_SELECTED');
}
````

File: client/src/modules/multiple_payroll/multiple_payroll.js (hunk 3 above)
Comment (lomamech) [reply client/src/modules/multiple_payroll/multiple_payroll.js]: No, employees are not related to the Project, only to the Enterprise

File: client/src/modules/multiple_payroll/multiple_payroll.js (hunk 1 above)
Comment (jniles) [reply client/src/modules/multiple_payroll/multiple_payroll.js]: Ah, I see.  Maybe you should put that as the first thing checked? 

EDIT: as in:
```js
if (!employees.length) {
  Notify.warn(/*...*/);
}

// other code...

```

client/src/modules/multiple_payroll/multiple_payroll.js
jniles: Perfect 👍

server/controllers/finance/exchange.js
lomamech: To test this function I had to add in my test database a third currency '€' Euro, and the function sent the exchange rate for all currencies in relation to the currency of the company

server/controllers/payroll/multiplePayroll/makeCommitment.js
lomamech: This function returns the last exchange rate configured per currency, if the system outside the company currency use several other currencies, this function would return the exchange rate for each of the currencie

server/controllers/payroll/reports/payrollReportSocialCharge.handlebars
lomamech: The employees are displayed by alphabetical order, the employees who appear on this report are those who have been selected, the index only allows the order of employees and it can also allow to quickly know the number of employees select

lomamech:
@jniles I need another Review for this pull Request

jniles:
Excellent work @lomamech.  I'll take a look.

test/end-to-end/vouchers/vouchers.spec.js
```diff
@@ -6,12 +6,12 @@ const GridRow = require('../shared/GridRow');
...
 
 describe('Voucher Registry', () => {
-  const NUM_VOUCHERS = 28;
+  const NUM_VOUCHERS = 23;
   const gridId = 'voucher-grid';
 
   before(() => helpers.navigate('vouchers'));
 
-  it(`displays ${NUM_VOUCHERS} vouchers on the page`, () => {
+  it(`displays Above ${NUM_VOUCHERS} vouchers on the page`, () => {
```
jniles, line 14: 💃

test/end-to-end/vouchers/vouchers.spec.js
```diff
@@ -6,12 +6,12 @@ const GridRow = require('../shared/GridRow');
 const components = require('../shared/components');
 
 describe('Voucher Registry', () => {
-  const NUM_VOUCHERS = 28;
+  const NUM_VOUCHERS = 23;
```
jniles, line 9: We should probably figure out a better way to test the number of records downloaded in the future.

server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -1,216 +1,238 @@
...
-  const employeeData = [];
+  const params = {
+    payroll_configuration_id : options.idPeriod,
+    reference : options.employees,
+  };
+
+  let template;
+  _.extend(options, DEFAULT_OPTS);
+
+  if (options.currency && options.socialCharge) {
```
jniles, line 40: Much cleaner!

jniles (approved):
Code looks fine to me.  I've also tried it out and it seems to work alright.  Good work!

bors r+

jniles:
Seems like a fluke...

bors r+

PR #3007, opened 2018-08-07
feat(server fiscal controller) get fiscal year by period id function

### LINKED ISSUE #2995: Create Fiscal.getFiscalYearByPeriodId()
We need a function to retrieve a fiscal year by it's period_id.  We often need to get data about the Fiscal Year for report headers.  We could then replace a series of custom functions with this new function to simplify report creation.

jeremielodi:
closes #2995


jniles (approved):
Perfect 👍

PR #3008, opened 2018-08-07
refractor(general ledjer) display dynamically periods' names

jeremielodi:


client/src/modules/general-ledger/general-ledger.ctrl.js
```diff
@@ -365,6 +199,34 @@ function GeneralLedgerController(
       });
   }
 
+  function renameGidHeaders() {
+    const actions = angular.copy(vm.columns[vm.columns.length - 1]);
+    const newColumns = [];
+    for (let i = 0; i < 3; i++) {
```
jniles: Just FYI, you can do this in one line:
```js
const newColumns = vm.columns.slice(0, 3);
```

client/src/modules/general-ledger/general-ledger.ctrl.js
```diff
@@ -365,6 +199,34 @@ function GeneralLedgerController(
...
+      newColumns.push(vm.columns[i]);
+    }
+
+    vm.year.periods.forEach((p, index) => {
```
jniles: Whenever you write a `array.forEach()` with a `newArray.push()` method inside it, you should ask yourself if you actually mean `array.map()`.  This code could be better rewritten as:
```js

const header = {
  cellFilter       : 'currency:'.concat(Session.enterprise.currency_id),
  // all other common properties...
};

const translatedColumns = vm.year.periods.map((period, index) => {
  return angular.extend({}, header, {
    field : `balance${index}`,
    displayName : p.translate_key,
  });
});
```

client/src/modules/general-ledger/general-ledger.ctrl.js
```diff
@@ -365,6 +199,34 @@ function GeneralLedgerController(
       });
   }
 
+  function renameGidHeaders() {
+    const actions = angular.copy(vm.columns[vm.columns.length - 1]);
```
jniles: Just FYI, you can also get this value via [`Array.prototype.slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice).
```js
const actions = vm.columns.slice(-1);
```

server/controllers/finance/fiscal.js
```diff
@@ -83,6 +83,15 @@ function list(req, res, next) {
     JOIN user AS u ON u.id = f.user_id
   `;
 
+  const periodsSql = `
```
jniles, line 88: Instead of changing the API for fiscal years, could we just do an additional call to `FiscalService.getPeriods()` on the client?

server/controllers/finance/fiscal.js
```diff
@@ -98,13 +107,21 @@ function list(req, res, next) {
...
-
   const query = filters.applyQuery(sql);
   const parameters = filters.parameters();
 
+  let fiscals = [];
   db.exec(query, parameters)
     .then((rows) => {
-      res.status(200).json(rows);
+      fiscals = rows;
+      return q.all(fiscals.map(fiscal => {
```
jniles: This has the potential to send back a lot of data!  If we do this, it would be better to put it behind a query string flag - like `?includePeriods=1`.

That being said, I much prefer the solution using `FiscalService.getPeriods()` as a second HTTP call.

jniles (changes requested):
@jeremielodi,  it looks like we already have a method to get the periods for a fiscal year on the client in `FiscalService.getPeriods()`.  Do you think we could just use `moment(date).format('MMMM')` to get the month on the client?  It would make the code less complex, as well as ensure we don't have to write new tests for the fiscal year route.

Otherwise, if you change an API, you should write tests for it.

server/controllers/finance/fiscal.js
jeremielodi: The problem is : I will have to send another quest for getting periods Once I have a fiscal year. I can just write a test for the changes I made to the API

server/controllers/finance/fiscal.js
jeremielodi: Ok! adding a parameter is much better

server/controllers/finance/fiscal.js
jniles: But everyone else who requests a fiscal year will also get all the periods, even if they don't want them.  In the long run, that is probably the more expensive behavior.

client/src/modules/general-ledger/general-ledger.ctrl.js
```diff
@@ -365,6 +199,36 @@ function GeneralLedgerController(
...
+
+    year.periods.forEach((p, index) => {
+      newColumns.push(angular.extend({}, header, {
+        field            : `balance${index}`,
+        displayName      : p.translate_key || 'FORM.LABELS.OPENING_BALANCE',
+      }));
+    });
+
+    newColumns.push(actions);
+    vm.gridOptions.columnDefs = angular.copy(newColumns);
```
jniles: Not a huge issue, but you don't need a second `angular.copy()` here.   You have already copied `actions`, `newColumns` is a new array, and `Array.prototype.slice()` returns a shallow copy.

> The slice() method returns a shallow copy of a portion of an array into a new array object selected from begin to end (end not included). The original array will not be modified.

[MDN Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)

A more concise format might be:
```js
vm.gridOptions.columnDefs = [...newColumns, actions];
```

test/integration/fiscalYear.js
```diff
@@ -51,6 +51,17 @@ describe('(/fiscal) Fiscal Year', () => {
       .catch(helpers.handler);
   });
 
+  it('GET /fiscal returns a list of fiscal_years width their periods', () => {
+    return agent.get('/fiscal?includePeriods=1')
+      .then(res => {
+        helpers.api.listed(res, 5);
+        const firstYearperiods = res.body[0].periods;
+        expect(firstYearperiods).to.not.be.empty;
```
jniles: Could you test some values to see that it returns the correct results?

jniles (commented):
This is almost ready.

It is important to also test that `GET /fiscal` only returns the fiscal year without `fiscal.periods`.  That way, we aren't changing the old API behavior.

client/src/modules/general-ledger/general-ledger.ctrl.js
```diff
@@ -20,7 +20,8 @@ function GeneralLedgerController(
 ) {
   const vm = this;
   const cacheKey = 'GeneralLedger';
-
+  vm.columns = [];
```
jniles: I am also curious why you stuck this on the `vm`?  Was there a problem with `let columns`?

jniles (approved):

jniles:
Looks good to me.  All changes addressed!

bors r+

PR #3014, opened 2018-08-09
report(ohada): bilan report

mbayopanda:
This PR add the ohada balance sheet (bilan) report.

Before use the report, it is necessary to configure correclty accounts references which are used in the ohada balance sheet.

The report use these references : 

actif
passif

To produce : 

image

image


client/src/i18n/en/report.json
```diff
@@ -176,6 +176,81 @@
...
+      "REF" : "Ref",
+      "ASSET" : "Asset",
+      "LIABILITY" : "Liability",
+      "BRUT" : "Brut",
+      "ADP" : "ADP",
+      "ADP_LONG" : "Amortissements, Dépréciations, Provisions",
+      "NET" : "Net",
+      "EXERCISE" : "Fiscal Year",
+      "REF_DESCRIPTION" : {
+        "AD" : "Immobilisations incorporelles",
```
jniles, line 191: At some point, we should encode these in our database.

client/src/i18n/en/tree.json
```diff
@@ -104,6 +106,8 @@
     "VOUCHER_REGISTRY" : "Voucher Registry",
     "AGED_DEBTORS" : "Aged Debtors Report",
     "WEEKEND_CONFIGURATION" : "Weekend Configuration",
-    "ROLE_MANAGEMENT" : "Role Management"
+    "ROLE_MANAGEMENT" : "Role Management",
+    "OHADA_BALANCE_SHEET":"[OHADA] Balance Sheet",
```
jniles, line 110: 👍 Good namespacing

server/controllers/finance/reports/ohada_balance_sheet/index.js
```diff
@@ -0,0 +1,299 @@
...
+// default report parameters
+const DEFAULT_PARAMS = {
+  csvKey : 'accounts',
+  filename : 'TREE.BALANCE',
+  orientation : 'landscape',
+  footerRight : '[page] / [toPage]',
+};
+
+const balanceSheetAssetTable = [
+  { ref : 'AD', is_title : 1 },
```
jniles, line 33: 👍 I appreciate that these are in alphabetical order!

server/controllers/finance/reports/ohada_balance_sheet/index.js
```diff
@@ -0,0 +1,299 @@
...
+
+  getFiscalYearDetails(params.fiscal_id)
+    .then(fiscalYear => {
+      _.merge(context, { fiscalYear });
+
+      const dbPromises = [];
+      const usePrevious = fiscalYear.previous.period_id ?
+        AccountReference.computeAllAccountReference(fiscalYear.previous.period_id) : [];
+
+      dbPromises.push(AccountReference.computeAllAccountReference(fiscalYear.current.period_id));
```
jniles: Do you need to make a `dbPromises` array here?  It disguises the fact there there are only two elements.

I propose you just give both results a name and put them in the array one at a time.  It will save you a line of code as well:
```js
const currentPeriodReferences = AccountReference.computeAllAccountReference(fiscalYear.current.period_id);
const previousPeriodReferences = fiscalYear.previous.period_id ?
        AccountReference.computeAllAccountReference(fiscalYear.previous.period_id) : [];

return Q.all([currentPeriodReferences, previousPeriodReferences]);
```

In my mind it makes it a bit more clear.

server/controllers/finance/reports/ohada_balance_sheet/index.js
```diff
@@ -0,0 +1,299 @@
...
+    item.currentNet += currentDb[ref] ? currentDb[ref].net.balance : 0;
+    item.previousNet += previousDb[ref] ? previousDb[ref].net.balance : 0;
+  });
+
+  return item;
+}
+
+function formatReferences(references) {
+  const values = {};
+  Object.keys(references).forEach(key => {
```
jniles: If you are already using `lodash`, you could also do:
```js
_.forEach(references, (value, key) => {
  // etc...
});
```

I recommend it in this case because instead of doing `references[key].filter()`, you can instead do `value.filter()`.  It would also be nice to name `value` something - like `accounts` or `balances` or `whateverItRepresents`.

server/controllers/finance/reports/ohada_balance_sheet/index.js
```diff
@@ -0,0 +1,299 @@
...
+    const [brut] = references[key].filter(elt => elt.is_amo_dep === 0);
+    let [amortissement] = references[key].filter(elt => elt.is_amo_dep === 1);
```
jniles: `is_amo_dep` must be either 0 or 1 always, right?  Can we use [`_.partition()`](https://lodash.com/docs/4.17.10#partition) for this?
```js
let [brut, amortissement] = _.partition(references[key], elt => elt.is_amo_dep === 0);
```

jniles: Wait... what are the types of `brut` and `amortissment`?  Are they always objects?  Or can they be arrays?

server/controllers/finance/reports/ohada_balance_sheet/index.js
```diff
@@ -0,0 +1,299 @@
...
+    next(e);
+    return;
+  }
+
+  getFiscalYearDetails(params.fiscal_id)
+    .then(fiscalYear => {
+      _.merge(context, { fiscalYear });
+
+      const dbPromises = [];
+      const usePrevious = fiscalYear.previous.period_id ?
```
jniles: Can you help me understand the logic of checking `period_id`?  Why do we need to consider it at all?

server/controllers/finance/reports/ohada_balance_sheet/index.js
```diff
@@ -0,0 +1,299 @@
...
+  const query = `
+    SELECT 
+      p.id AS period_id, p.end_date,
+      fy.id, fy.label, fy.previous_fiscal_year_id 
+    FROM fiscal_year fy 
+    JOIN period p ON p.fiscal_year_id = fy.id 
+      AND p.number = (SELECT MAX(period.number) FROM period WHERE period.fiscal_year_id = ?)
+    WHERE fy.id = ?;
+  `;
+  return db.one(query, [fiscalYearId, fiscalYearId])
```
jniles, line 291: I think this would be easier to understand if you just did two calls to `getFiscalYearDetails()`.  That way, the function does what we expect - just gets the fiscal year info for one fiscal year identified by the `fiscalYearId`.  It also will make it easier to change in the future if we need to factor in multiple years or something.

The pseudocode would look like this:

```js
let bundle = {};
getFiscalYearDetails(params.fiscal_id)
  .then(fiscalYear => {
    bundle.current = fiscalYear;
    
    return fiscalYear.previous_fiscal_year_id ? getFiscalYearDetails(fiscalYear.previous_fiscal_year_id) : {};
  })
  .then(previousFiscalYear => {
   // etc...
  });
```

What do you think?

server/controllers/finance/reports/ohada_balance_sheet/index.js
```diff
@@ -0,0 +1,299 @@
...
+      description : brut.description,
+      balance : brut.balance - amortissement.balance,
+    };
+
+    values[key] = { brut, amortissement, net };
+  });
+  return values;
+}
+
+function getFiscalYearDetails(fiscalYearId) {
```
jniles, line 279: By the way, this is a perfect place to use `async/await` if you want to.

```js
function async getFiscalYearDetails(fiscalYearId) {
  const bundle = {};
  const query = ` ... `;
  bundle.fiscalYear = await db.one(query, [ /* params */ ]);

  if (bundle.fiscalYear.previous_fiscal_year_id) {
    bundle.previousYear = await db.one(query, [ /* params */ ]);
  }

  return bundle;
}
```

server/models/Update/add_ohada_balance_sheet.sql
```diff
@@ -0,0 +1,8 @@
+/* Add report in the navigation tree */
```
jniles, line 1: Nice upgrade script :+1:

server/models/Update/add_ohada_balance_sheet.sql
```diff
@@ -0,0 +1,8 @@
...
+INSERT INTO unit VALUES 
+  (206, '[OHADA] Bilan','TREE.OHADA_BALANCE_SHEET','',144,'/modules/reports/ohada_balance_sheet_report','/reports/ohada_balance_sheet_report');
```
jniles, line 3: Can you remove the `id` here?

I can imagine that other developers will put in there own units with conflicting IDs.  Since we aren't sure which will get merged first, it makes sense to let the database assign the IDs if possible.

server/models/Update/add_ohada_balance_sheet.sql
```diff
@@ -0,0 +1,8 @@
...
+
+/* Record the report */
+INSERT INTO `report` (`id`, `report_key`, `title_key`) VALUES
```
jniles, line 6: Same comment for the `id` here.

server/controllers/finance/reports/ohada_balance_sheet/index.js
```diff
@@ -0,0 +1,299 @@
...
+      return Q.all(dbPromises);
+    })
+    .spread((current, previous) => {
+      let list = [];
+      const currentReferences = formatReferences(_.groupBy(current, 'abbr'));
+      const previousReferences = formatReferences(_.groupBy(previous, 'abbr'));
+
+      const assetTable = balanceSheetAssetTable.map(item => {
+        item.label = 'REPORT.OHADA.REF_DESCRIPTION.'.concat(item.ref);
+        if (currentReferences[item.ref]) {
```
jniles: Let's give `currentReferences[item.ref]` an alias.  Like:
```js
const row = currentReferences[item.ref];
if (row) {
  item.currentBrut = row.brut.balance;
  // etc..
}
```

It just makes it easier to read.  Choose a name for `row` that is true to what it actually is.

server/controllers/finance/reports/ohada_balance_sheet/index.js
```diff
@@ -0,0 +1,299 @@
...
+          item.currentBrut = currentReferences[item.ref].brut.balance;
+          item.currentAmo = currentReferences[item.ref].amortissement.balance;
+          item.currentNet = currentReferences[item.ref].net.balance;
+          item.previousNet = previousReferences[item.ref] ? previousReferences[item.ref].net.balance : 0;
+        }
+
+        // process manually totals
+        if (item.ref === 'AD') {
```
jniles: 👍 

Technically, you could do this all more efficiently with `switch`/`case` but this is really easy to read.  Good decision.

jniles (approved):
In general, the code is nice and readable.  I get zeros when I run it for everything, but I assume you've done the work to make sure it works.

I've lots of little cosmetic suggestions.  The output report looks good.

We will probably want to store all the references in their own table somewhere with labels and descriptions so that a BHIMA developer does not have to change everything every time OHADA decides it wants to update a name.

Looks good to me.  A solid PR.  You can incorporate any suggestions you find useful and then bors r+ it.

bors delegate+

server/models/Update/add_ohada_balance_sheet.sql
mbayopanda: The `id` column is not auto increment in the unit table

jniles: Oh ... I suppose that makes sense.

PR #3018, opened 2018-08-10
Add barcode scanning to bhFindPatient component

### LINKED ISSUE #2668: Implement barcode scanning on bhFindPatient component.
It would be really cool if we could use a barcode to quickly identify patients.  That was one of the original goals of the barcode scanning feature - reducing the number of errors made in the patient selection by scanning barcodes.

The basic idea is this:
 1. Add an option to the bhFindPatient component that allows a user to scan a barcode.  Something like so:
 patientbarcode
 2. Use the barcode scanner modal to look up [the patient identifier](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/barcode.js#L11).
 3. Set the `bhFindPatient` to success on a successful lookup and call the callback.

jniles:
This pull request adds the barcode scanning component to the `bhFindPatient` component.  Barcodes are far more reliable and easy to use than typing in a patient's name.  The goal is to increase registration of patients and improve accuracy of all operations affecting patients.

Closes #2668.

Depends on #3013.

See below for the component in action.

### DIFF
--- a/client/src/modules/patients/registry/registry.html
+++ b/client/src/modules/patients/registry/registry.html
@@ -48,6 +48,12 @@
                 <span class="fa fa-file-excel-o"></span> <span translate>DOWNLOADS.EXCEL</span>
               </a>
             </li>
+            <li role="separator" class="divider"></li>
+            <li role="menuitem">
+              <a href ng-click="PatientRegistryCtrl.openBarcodeScanner()" data-action="scan-barcode">
+                <span class="fa fa-barcode"></span> <span translate>BARCODE.SCAN</span>
+              </a>
+            </li>
           </ul>
 
         </div>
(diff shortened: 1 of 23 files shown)


client/src/modules/patients/registry/registry.html
```diff
@@ -48,6 +48,12 @@
                 <span class="fa fa-file-excel-o"></span> <span translate>DOWNLOADS.EXCEL</span>
               </a>
             </li>
+            <li role="separator" class="divider"></li>
+            <li role="menuitem">
+              <a href ng-click="PatientRegistryCtrl.openBarcodeScanner()" data-action="scan-barcode">
+                <span class="fa fa-barcode"></span> <span translate>BARCODE.SCAN</span>
```
mbayopanda, line 54: The translation here is about the patient not the document, so it will be better to have a good translation for the patient.

Ex. 
```json
{ "BARCODE" : {
  "SCAN_PATIENT" : "Scanner le code bar du patient"
  }
}
```

mbayopanda (approved):
LGTM 👍 

bors r+

client/src/modules/patients/registry/registry.html
jniles: Yeah, that is true.  We can add that 👍

PR #3023, opened 2018-08-13
feat(invoice registry) optimize sql query

jeremielodi:
Joining invoice with debtor makes the request very slow, this PR solves that issue.

### DIFF
--- a/server/controllers/finance/patientInvoice.js
+++ b/server/controllers/finance/patientInvoice.js
@@ -227,6 +227,12 @@ function find(options) {
   // @FIXME Remove this with client side filter design
   delete options.patientNames;
 
+  let debtorJoin = ``;
+  const debtorKey = 'debtor_group_uuid';
+  const hasDebtorKey = options[debtorKey];
+  if (hasDebtorKey) {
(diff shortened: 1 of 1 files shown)


server/controllers/finance/patientInvoice.js
```diff
@@ -227,6 +227,12 @@ function find(options) {
   // @FIXME Remove this with client side filter design
   delete options.patientNames;
 
+  let debtorJoin = ``;
+  const debtorKey = 'debtor_group_uuid';
```
jniles, line 231: Are you sure this shouldn't be `debtor_uuid`?  Why filter on the `debtor_group_uuid`?

jeremielodi: I have not changed the join query to debtor. The only thing I tried to do is to add a flag that test if the user sent ```debtor_group_uuid ``` in ```req.query```, if true : the join to debtor table is useful else we can skip it, that makes the request more faster.

jniles: Ah, I understand now.  LGTM.

jniles (approved):
bors r+

PR #3024, opened 2018-08-13
accounts references and report script

mbayopanda:
This PR add a separate file for adding accounts references and report unit

### DIFF
--- a/server/models/Update/add_account_reference_module.sql
+++ b/server/models/Update/add_account_reference_module.sql
@@ -4,6 +4,30 @@
   NOTA : Please create `account_reference` and `account_reference_item` tables first
 */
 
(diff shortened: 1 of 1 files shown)


server/models/Update/add_account_reference_module.sql
```diff
@@ -0,0 +1,12 @@
+/*
+  ACCOUNT REFERENCE MODULE AND REPORT
+  ===================================
+  NOTA : Please create `account_reference` and `account_reference_item` tables first
```
lomamech, line 4: @mbayopanda I think it would be better to include also the scripts of the creations of the tables that were added in the file schema.sql this will facilitate the update of the databases already (Tshikaji and Vanga)

PR #3029, opened 2018-08-13
feat(journal search) fillter by amount

### LINKED ISSUE #3025: Change "Amount" search filter to look at both "amount" fields and "amount_equiv" fields
In auditing an account statement, it came to our attention that we should look at all the `debit`, `credit`, `debit_equiv`, and `credit_equiv` fields for the "amount" filter.  For example, if an account statement is given in a different currency, the "amount" filter will not be able to locate the amount.  You could exchange it on the fly, but what happens if the exchange rate is unknown or incorrect?

jeremielodi:
closes  #3025

### DIFF
--- a/server/controllers/finance/journal/index.js
+++ b/server/controllers/finance/journal/index.js
@@ -157,7 +157,11 @@ function buildTransactionQuery(options, posted) {
 
   filters.custom('uuids', 'p.uuid IN (?)', [options.uuids]);
   filters.custom('record_uuids', 'p.record_uuid IN (?)', [options.record_uuids]);
-  filters.custom('amount', '(credit_equiv = ? OR debit_equiv = ?)', [options.amount, options.amount]);
+  const { amount } = options;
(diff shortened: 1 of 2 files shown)


server/controllers/finance/journal/index.js
```diff
@@ -157,7 +157,11 @@ function buildTransactionQuery(options, posted) {
 
   filters.custom('uuids', 'p.uuid IN (?)', [options.uuids]);
   filters.custom('record_uuids', 'p.record_uuid IN (?)', [options.record_uuids]);
-  filters.custom('amount', '(credit_equiv = ? OR debit_equiv = ?)', [options.amount, options.amount]);
+  const { amount } = options;
+  filters.custom(
```
jniles, line 161: This is right!  Looks like the tests need to be updated.

jniles (commented):
Code is good, looks like the test needs to be updated to work on the new behavior.

PR #3031, opened 2018-08-13
Allow users to disable barcodes for enterprise

### LINKED ISSUE #2820: Add Enterprise Setting to disable barcodes
Some enterprises may never use barcodes.  It would useful to have an Enterprise Setting that completely "removes" barcodes from the system by:
 1. Remove the barcode-related options, such as "Scan a barcode" from the cash payments module.
 2. Remove the rendering of barcodes from the receipts (this would avoid calling the `jsbarcode.init()` and produce a small speed increase).
 3. Remove automatic options such as opening another barcode modal from the cash payments page.

The enterprise setting would be **Enable Barcodes** and be in our classic yes/no format.

jniles:
This PR implements the ability to remove barcodes from the application with a single click.  For institutions that will never use barcodes, they are a waste of ink and space.  Now, the option to have barcodes is an enterprise-level setting to either show them or remove them.

A new component `bhRequireEnterpriseSetting` has been added to make life easier when hiding elements based on the enterprise setting.  You just have a stick it to an element and it will be removed if the enterprise does not support that setting.  Such as this:
```html
<div bh-require-enterprise-setting="prepayments">
 <span> I will only be shown if "session.enterprise.settings.enable_prepayments" is true! </span>
</div>
```

Closes #2820.

### DIFF
--- a/client/src/js/directives/bhRequireEnterpriseSetting.js
+++ b/client/src/js/directives/bhRequireEnterpriseSetting.js
@@ -1,0 +1,26 @@
+angular.module('bhima.directives')
+  .directive('bhRequireEnterpriseSetting', bhRequireEnterpriseSetting);
+
+bhRequireEnterpriseSetting.$inject = ['SessionService'];
+
+/**
+ * @function bhRequireEnterpriseSetting
+ *
+ * @description
+ * A sort directive to remove the underlying HTML if the Enterprise Setting is
+ * not set.
+ */
+function bhRequireEnterpriseSetting(Session) {
+  return {
+    restrict : 'A',
+    scope : { bhRequireEnterpriseSetting : '@bhRequireEnterpriseSetting' },
+    link : (scope, element, attrs) => {
+      const hasRequiredSetting = Session.isSettingEnabled(attrs.bhRequireEnterpriseSetting);
+
+      // remove setting if not available
+      if (!hasRequiredSetting) {
+        element.remove();
+      }
+    },
+  };
+}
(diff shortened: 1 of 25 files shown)


client/src/modules/enterprises/enterprises.js
```diff
@@ -170,30 +165,28 @@ function EnterpriseController(Enterprises, util, Notify, Projects, Modal, Scroll
...
-    $touched = true;
+  /**
+   * @function proxy
+   *
+   * @description
+   * Proxies requests for different enterprise settings.
+   *
+   * @returns {function}
+   */
+  function proxy(key) {
```
mbayopanda, line 177: 👍

server/controllers/finance/reports/cash/receipt.handlebars
```diff
@@ -27,7 +27,9 @@
           </span> <br>
           <strong>{{payment.reference}}</strong> <br>
           <small>{{date payment.date}}</small> <br>
-          <small>{{> barcode value=details.barcode}}</small> <br>
+          {{#if metadata.enterprise.settings.enable_barcodes}}
```
mbayopanda, line 30: 👍

server/controllers/finance/reports/cash/receipt.pos.handlebars
```diff
@@ -1,13 +1,17 @@
 <head>
   <meta charset="utf-8" />
-  <script src="{{absolutePath}}/vendor/JsBarcode.all.min.js"></script>
+  {{#if metadata.enterprise.settings.enable_barcodes}}
```
mbayopanda, line 3: Good catch for performances 👍

client/src/js/directives/bhRequireEnterpriseSetting.js
```diff
@@ -0,0 +1,26 @@
...
+bhRequireEnterpriseSetting.$inject = ['SessionService'];
+
+/**
+ * @function bhRequireEnterpriseSetting
+ *
+ * @description
+ * A sort directive to remove the underlying HTML if the Enterprise Setting is
+ * not set.
+ */
+function bhRequireEnterpriseSetting(Session) {
```
mbayopanda, line 13: 👍

mbayopanda (approved):
LGTM

bors r+

PR #3036, opened 2018-08-14
Process stock lot lifetime

mbayopanda:
This PR add useful information (lot lifetime, lot risk by month and lot risk by quantity) for stock manager into the stock lots registry.


server/controllers/stock/core.js
```diff
@@ -494,6 +475,59 @@ function getInventoryQuantityAndConsumption(params) {
...
+ *
+ * @description
+ * the goals of this function is to give the risk of peremption for each lots for
+ * a given inventory
+ */
+function processMultipleLots(inventories) {
+  const flattenLots = [];
+  const inventoryLots = _.groupBy(inventories, 'inventory_uuid');
+
+  _.each(inventoryLots, (inventory) => {
```
jniles: Can we do a `_.map()` here?  It looks like that is what you are wanting to do.

server/controllers/stock/core.js
```diff
@@ -494,6 +475,59 @@ function getInventoryQuantityAndConsumption(params) {
...
+    // order lots by ascending lifetime
+    const orderedInventoryLots = _.orderBy(inventory, 'lifetime', 'asc');
```
jniles: Is there any way we can do the order by in SQL?  I worry this will be pretty slow.

client/src/modules/stock/lots/registry.js
```diff
@@ -92,6 +92,40 @@ function StockLotsController(
       displayName : 'STOCK.EXPIRATION',
       headerCellFilter : 'translate',
     },
+    {
+      field : 'avg_consumption',
+      displayName : 'STOCK.CMM',
```
jniles, line 97: Are these numbers?  If so, could we put `type: "number"` on the field?  It will make sure we sort these columns in the right way.

jniles (approved):
LGTM.

bors r+

PR #3040, opened 2018-08-16
feat(services) color service

### LINKED ISSUE #3027: Move colors into their own service
We added support for colors to easily identify different debtor groups.  The code currently resides directly in the debtor groups service ([here](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/debtors/groups.service.js#L27)) as `DebtorGroups.colors`.

It would make more sense to move this into its own service or constant.  That way the same color scheme could be used anywhere else that we expect to use colors.  It will also make it easier to change/remove without introducing side-effects.

jeremielodi:
closes #3027

### DIFF
--- a/client/src/js/services/color.js
+++ b/client/src/js/services/color.js
@@ -1,0 +1,22 @@
+angular.module('bhima.services').service('ColorService', ColorService);
+ColorService.$inject = [
+  '$translate',
+];
+
+function ColorService($translate) {
+  const service = this;
+  service.list = [
+    { name : $translate.instant('COLORS.AQUA'), value : '#00ffff' },
+    { name : $translate.instant('COLORS.GRAY'), value : '#808080' },
+    { name : $translate.instant('COLORS.FORESTGREEN'), value : '#228b22' },
+    { name : $translate.instant('COLORS.LIME'), value : '#00ff00' },
+    { name : $translate.instant('COLORS.RED'), value : '#ff0000' },
+    { name : $translate.instant('COLORS.YELLOW'), value : '#ffff00' },
+    { name : $translate.instant('COLORS.YELLOWGREEN'), value : '#9acd32' },
+    { name : $translate.instant('COLORS.SLATEBLUE'), value : '#6a5acd' },
+    { name : $translate.instant('COLORS.MAROON'), value : '#800000' },
+    { name : $translate.instant('COLORS.CRIMSON'), value : '#dc143c' },
+    { name : $translate.instant('COLORS.BLUEVIOLET'), value : '#8A2BE2' },
+  ];
+  return service;
+}
--- a/client/src/modules/debtors/groups.create.js
+++ b/client/src/modules/debtors/groups.create.js
@@ -37,12 +37,8 @@ function DebtorGroupCreateController(
   vm.onSelectAccountCallback = onSelectAccount;
   vm.cancel = cancel;
 
-  Accounts.read()
-    .then((accounts) => {
-      vm.accounts = accounts;
-      return Prices.read();
-    })
   /* @todo This controller should not be concerned about individual price lists */
+  Prices.read()
     .then((priceLists) => {
       vm.priceLists = priceLists;
       vm.$loaded = true;
--- a/client/src/modules/debtors/groups.create.js
+++ b/client/src/modules/debtors/groups.create.js
@@ -2,8 +2,9 @@ angular.module('bhima.controllers')
   .controller('DebtorGroupCreateController', DebtorGroupCreateController);
 
 DebtorGroupCreateController.$inject = [
-  '$state', 'ScrollService', 'SessionService', 'DebtorGroupService',
-  'PriceListService', 'uuid', 'NotifyService',
+  '$state', 'ScrollService', 'SessionService',
+  'DebtorGroupService', 'AccountService', 'PriceListService',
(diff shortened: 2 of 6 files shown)


client/src/modules/debtors/groups.service.js
```diff
@@ -24,19 +24,7 @@ function DebtorGroupService($http, Modal, util, SessionService, $translate) {
...
-    { name : $translate.instant('COLORS.LIME'), value : '#00ff00' },
-    { name : $translate.instant('COLORS.RED'), value : '#ff0000' },
-    { name : $translate.instant('COLORS.YELLOW'), value : '#ffff00' },
-    { name : $translate.instant('COLORS.YELLOWGREEN'), value : '#9acd32' },
-    { name : $translate.instant('COLORS.SLATEBLUE'), value : '#6a5acd' },
-    { name : $translate.instant('COLORS.MAROON'), value :'#800000' },
-    { name : $translate.instant('COLORS.CRIMSON'), value :'#dc143c' },
-    { name : $translate.instant('COLORS.BLUEVIOLET'), value: '#8A2BE2' },
-  ];
+  service.colors = Color.list;
```
jniles: Instead of importing Colors into the debtor groups module, could you put it in the controller that needs `DebtorGroups.colors`?  That way this service focuses only on DebtorGroups.

client/src/js/services/color.js
```diff
@@ -0,0 +1,22 @@
+angular.module('bhima.services').service('ColorService', ColorService);
```
jniles, line 1: Nice!

Can you write a quick unit test to make sure that `service.list` returns an array?

jniles:
If you need help with client-side unit testing, this is a great link:
http://www.bradoncode.com/tutorials/angularjs-unit-testing/

jniles:
@jeremielodi, could you write a quick unit tests to make sure these colors are correctly exported?  After that and the ESLint fix, LGTM.

jniles:
Looks like just the test remains!

jeremielodi:
ready for review

client/src/modules/debtors/groups.update.js
```diff
@@ -6,13 +6,13 @@ angular.module('bhima.controllers')
   .controller('DebtorGroupUpdateController', DebtorGroupsUpdateController);
 
 DebtorGroupsUpdateController.$inject = [
-  '$state', 'DebtorGroupService', 'PriceListService',
-  'ScrollService', 'util', 'NotifyService', 'ModalService',
+  '$state', 'DebtorGroupService', 'AccountService', 'PriceListService',
+  'ScrollService', 'util', 'NotifyService', 'ModalService', 'ColorService',
 ];
 
 function DebtorGroupsUpdateController(
```
jniles: Why do we need accounts here?

test/client-unit/services/ColorsService.spec.js
```diff
@@ -0,0 +1,20 @@
...
+    'pascalprecht.translate'
+  ));
+
+  beforeEach(inject(_ColorService_ => {
+    ColorService = _ColorService_;
+  }));
+
+  it('ColorService.list should return an array', () => {
+    const colorList = ColorService.list;
+    expect(Array.isArray(colorList)).to.be.equal(true);
```
jniles: A better assertion is `expect(colorList).to.be.a('array');`.

client/src/modules/debtors/groups.service.js
```diff
@@ -24,20 +24,20 @@ function DebtorGroupService(Modal, Session, $translate, Api) {
...
-    { name : $translate.instant('COLORS.BLUEVIOLET'), value : '#8A2BE2' },
-  ];
+
+  /**
+  * @method read
+  * @param {string} uuid The debtor group uuid
+  * @param {object} parameters The query string object
+  * @description This function is responsible for getting debtor groups
+  */
+  function read(uuid, parameters) {
```
jniles: Why do you need this function?  Since the service is an Api (line 15), it should already have it.  Does it not work for some reason?

client/src/modules/debtors/groups.js
```diff
@@ -27,7 +27,7 @@ function DebtorGroupController($state, DebtorGroups) {
   vm.setOrder = setOrder;
 
   vm.state = $state;
-  vm.colors = DebtorGroups.colors;
+  vm.colors = Color.list;
```
jniles, line 30: Better 👍

client/src/modules/debtors/groups.create.js
```diff
@@ -24,17 +29,21 @@ function DebtorGroupCreateController($state, ScrollTo, Session, DebtorGroups, Pr
...
+  vm.colors = Color.list;
 
   vm.$loading = true;
   vm.$loaded = false;
 
   vm.onSelectAccountCallback = onSelectAccount;
   vm.cancel = cancel;
 
-  Prices.read()
+  Accounts.read()
```
jniles: Why do you need accounts here?  Where is it used?

jniles (commented):
A few things to clean up.  I'm not sure why some of the changes were necessary, and think it will work without them.  Could you take a look?

jeremielodi:
I think I found it in this  service

Le dim. 2 sept. 2018 à 4:45 AM, Jonathan Niles <notifications@github.com> a
écrit :

PR #3041, opened 2018-08-16
refractor(session service) using $state instead of $location

### LINKED ISSUE #3030: Update SessionService to use $state instead of $location
The `SessionService` manages the user's session on the client.  When the session is logged out, the service redirects the user to the `/login` URL using $location.  Since we've mostly lifted navigation from angular's `$location` service to use ui-router's `$state` service, it would make sense that we also switch these calls to using `$state` for uniformity.

jeremielodi:
closes #3030

### DIFF
--- a/client/src/js/services/Session.js
+++ b/client/src/js/services/Session.js
@@ -2,7 +2,7 @@ angular.module('bhima.services')
   .service('SessionService', SessionService);
 
 SessionService.$inject = [
-  '$sessionStorage', '$http', '$location', 'util', '$rootScope', '$q',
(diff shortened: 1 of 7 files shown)


jniles:
Looks like the SessionService tests need to be updated before this will pass.  Could you take look at them?

jniles (approved):
Looks good to me!

bors r+

jniles:
@jeremielodi, looks like the most recent changes to the master branch have broken these tests.  Can you try to resolve these?

jeremielodi:
ok, let me give it a try

PR #3045, opened 2018-08-16
fix(reports): polish P&L Statement

### LINKED ISSUE #3044: Profit and Loss Statement doesn't calculate the "balance" correctly
It looks like the "balance" on the Profit and Loss statement doesn't correctly compute the balance.  Here it is below:
profitandlossbug
_Fig 1: Balance should not be zero_

This might be a data-binding issue or a miscalculation.  I would guess the first.

jniles:
This commit fixes a few bugs in the profit and loss statement:
 1. Renders the enterprise currency properly.
 2. Calculates the balance/difference between the profit and loss.
 3. No longer renders variance totals as a currency.
 4. Throws an error if there isn't a matching set of data.

Closes #3044.

### DIFF
--- a/server/controllers/finance/reports/income_expense/index.js
+++ b/server/controllers/finance/reports/income_expense/index.js
@@ -59,7 +59,7 @@ function document(req, res, next) {
   ])
     .then(([periodFrom, periodTo, fiscalYear]) => {
       if (periodFrom.start_date > periodTo.start_date) {
-        throw new BadRequest('The date range is inverted.');
+        throw new BadRequest('The date range is inverted.', 'ERRORS.BAD_DATE_INTERVAL');
       }
 
       _.extend(data, { periodFrom, periodTo, fiscalYear });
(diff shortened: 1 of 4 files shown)


jniles:

server/controllers/finance/reports/income_expense/index.js
```diff
@@ -77,6 +77,10 @@ function document(req, res, next) {
 
       const root = tree.getRootNode();
 
+      if (!Array.isArray(root.children) || root.children.length < 2) {
+        throw new BadRequest('Could not find both income and expense accounts for the time period');
+      }
```
mbayopanda: Could you add a key as second parameter of BadRequest() for having a meaningful message for the user.

Ex.
```js
throw new BadRequest('Could not find both income and expense accounts for the time period', 'ERRORS.NO_TRANSACTION_FOUND');
```

mbayopanda (approved):
LGTM 👍

server/controllers/finance/reports/income_expense/index.js
jniles: Or 'ERRORS.EVERYTHING_BAD' 😄 

Will do!

Issue #3039, opened 2018-08-15
UI for Reports/Balance of Accounts a little unfriendly


jmcameron:
When you bring up Reports/Balance of Accounts (/reports/balance_report), it has a dialog to tell it what period you want the report for.  However, if you only enter data for part of the 5 fields and press the **[Preview]** button, nothing happens.  And there is no helpful hint telling you to select the other items.    I am sure that Angular supports all kinds of form error warnings, so I am guessing that this not too difficult to fix.    

BTW, when you start the app, the first time you go to this page none of the 3 yes/no questions at the bottom of the form have anything selected.  Does it make sense to provide a default for these radio fields?

Although an experienced user or users who have a lot of computer experience will be able to figure this, out some of the more naive users this SW is designed for might scratch there heads on this for a while.

Not sure if this is in the scope of the kinds of things that are worth spending time on...

jniles:
Not having an error message sounds like a bug.  We typically try to catch those, but clearly one slipped through.

I agree!  We should have pre-selected the three options by default.  It's definitely something worth spending time on - confusion for inexperienced users is not acceptable.  Here are my proposed preselections:


![balanceofaccountsissue](https://user-images.githubusercontent.com/896472/44152718-2ab76594-a09e-11e8-8e6d-6cc63e95dc24.PNG)

_Fig 1: Preselections for the Balance of Accounts_

jmcameron:
Ok, I'll work on this (more getting up to speed).   I'll do it in two phases.  Fix the defaults, then handle errors.    Can you give me a suggestion of what form handles errors well to look at?

PR #3047, opened 2018-08-16
Fixed defaults on radio items on 'Balance des comptes' report

jmcameron:
This fix makes sure that the values for the radio items on this report respect the cached value (if defined), otherwise use default values per Jonathan Nile's suggestions.  Note that it not necessary to define reportDetails = {} in the else clause of the cache check since the containing function defines it already.   

Partial fix for Issue https://github.com/IMA-WorldHealth/bhima-2.X/issues/3039


jniles:
Beautiful.  Looks good to me!

bors r+

PR #3048, opened 2018-08-17
feat(accounts) export as Excel

### LINKED ISSUE #3042: Account Management XLSX export broken
At some point, the XLSX export for account management was broken - the user now downloads an empty file.  We know it is specific to the XLSX renderer since both the PDF and the CSV exporters work.  The XLSX exporter seems to work fine on the Journal.

It may have something to do with the addition of the "hidden" option for accounts.  Whoever fixes this should include a test to make sure we catch any bugs like this in the future.

jeremielodi:
closes #3042

### DIFF
--- a/server/lib/renderers/xlsx.js
+++ b/server/lib/renderers/xlsx.js
@@ -18,6 +18,7 @@ const headers = {
 };
 
 exports.render = render;
+exports.find = find;
 exports.extension = '.xlsx';
 exports.headers = headers;
 
@@ -26,7 +27,7 @@ exports.headers = headers;
  * @param {Object} data   { rows : []}
  */
 
-function render(data) {
+function render(data, template, options) {
   // Create a new instance of a Workbook class
   const wb = new xl.Workbook();
 
@@ -65,9 +66,7 @@ function render(data) {
 
   // Add Worksheets to the workbook
   const ws = wb.addWorksheet('Sheet 1');
-  let { rows } = data;
-  rows = rows || [];
-
+  const rows = find(data, options); // get all rows to set in the sheet
   const firstObject = rows[0] || {};
 
   // writing columns
@@ -93,6 +92,16 @@ function render(data) {
   return wb.writeToBuffer();
 }
 
+
+/*
+ find available data to write in the excel file
+ rows is the default key, if nothing found, this function get look at rowsDataKey
+ it return an empty array if no data found
+*/
+function find(data, _options) {
+  const options = _options || {};
+  return data.rows || data[options.rowsDataKey] || [];
+}
 // set value to a paticular cell
 function setValue(ws, x, y, value) {
   if (_.isNumber(value)) {
--- a/server/lib/renderers/xlsx.js
+++ b/server/lib/renderers/xlsx.js
@@ -84,7 +84,7 @@ function render(data) {
   // writing rows
   rows.forEach((row) => {
     firstLineCols.forEach((key, index) => {
-      setValue(ws, line, index + 1, row[key])
+      setValue(ws, line, index + 1, row[key] || '')
         .style(styleAllBorders)
         .style(normaFontSize);
     });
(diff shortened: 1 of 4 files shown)


jeremielodi:
@jniles there is already a test  <a href="https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/server-unit/xlsx.js">here</a> for an empty object case.

There problem was, there was no rowsDataKey in the options for telling xlsx renderer where to get data, by default it get them in ```data['rows']```

jniles:
That is a good catch!

Is there a test we could write to make sure the Excel rendering works for the accounts report?  Maybe a server integration test that makes sure the downloaded Excel has at least one row? 

Clearly the tests that exist failed us, since this was a real bug.

test/server-unit/xlsx.js
```diff
@@ -30,4 +30,26 @@ describe('util.js', () => {
...
+    const data = {
+      students : [
+        { name : 'Alice' },
+        { name : 'Bob' }],
+    };
+    // rowsDataKey is the specific key where the renderer will get data to write in the expected file
+    // it is used when the provided array doesn't have this key "rows"
+    const options = { rowsDataKey : 'students' };
+    const result = xlsx.find(data, options);
+    expect(result.length).to.be.equal(2);
```
jniles, line 52: Good tests!

You can probably check directly the values, since `result` should be `students`, right?
```js
expect(result).to.deep.equal(data.students);
```

test/server-unit/xlsx.js
```diff
@@ -30,4 +30,26 @@ describe('util.js', () => {
         done(error);
       });
   });
+
+  it('Shoul check the number of rows to write in the xlsx file', () => {
```
jniles: Small typo: "Shoul" should be "Should"

server/lib/renderers/xlsx.js
```diff
@@ -93,6 +92,16 @@ function render(data) {
   return wb.writeToBuffer();
 }
 
+
+/*
+ find available data to write in the excel file
+ rows is the default key, if nothing found, this function get look at rowsDataKey
+ it return an empty array if no data found
+*/
+function find(data, _options) {
```
jniles, line 101: This is an interesting function.  To avoid needless variables, let's just use [JavaScript's notation for default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters):
```js
function find(data, options = {}) {
```

server/lib/renderers/xlsx.js
```diff
@@ -93,6 +92,16 @@ function render(data) {
...
+  const options = _options || {};
+  return data.rows || data[options.rowsDataKey] || [];
```
jniles, line 103: Should the check for `data[options.rowDataKey]` come first?  What should happen if we do this:
```js
const data = {
  rows : 'This is a cool example',
  students : [{ id : 1, name : 'josh' }, { id : 3, name : 'jose' }]
};

const values = find(data, { rowsDataKey: 'students' });
// values is 'This is a cool example'
```

Is this the expected result?

jniles (commented):
Nice!  The tests are quite helpful.  I've made a few comments on the logic, but overall, this is looking good.

test/server-unit/xlsx.js
jeremielodi: a lot of shoul in the sentence

jeremielodi: ok

server/lib/renderers/xlsx.js
jeremielodi: Check the Reportmanager.js for this case, ```data.rows``` will be equal to ```data.students```
The man problem is, we have several final results, the soft option for me is to specify for all reports the rowsDataKey option when rendering into excel

jeremielodi:
@jniles, can this PR be merged?

jniles:
@jeremielodi almost.  You need to fix the following issues:
 1. Change all sentences with "Shoul" to "Should".
 2. Add the following to your server-unit test: `expect(result).to.deep.equal(data.students);`

After that, should be fine to merge.

jeremielodi:
ready!

jniles (approved):
LGTM.

bors r+

PR #3053, opened 2018-08-19
Update INSTALL.md resolves #3038

### LINKED ISSUE #3038: Change install documentation to use wkhtmltopdf version 0.12.4
The installation documentation for BHIMA does not include a specific version since wkhtmltopdf did not update itself in ~2 years.  But [the released version 0.12.5](https://wkhtmltopdf.org/downloads.html) earlier this summer and it breaks our PDF rendering unit test (since the report renders slightly differently).

In the long run, it would be nice to transition over to the new wkhtmltopdf version, but we would have to ensure that every report continued to behave the same.  In the short run, the most practical thing is to explicitly list 0.12.4 as the required version.

@Scholldier would you be willing to include this in your installation documentation changes?

Scholldier:
Hello! I have updated the installation documentation. Please let me know what changes need to be made, or if it looks good. One consideration to make is that I am specifying to grant all privileges to the bhima account, which probably isn't a 'best practice' for environments using any other database alongside bhima. What are your thoughts? Since this is just for the installation documentation, I suppose that will be self-evident enough.

Fixes #3038.

Thanks!

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -50,7 +50,6 @@ sudo rm wkhtmltox-0.12.4_linux-generic-amd64.tar.xz  && rm -rf wkhtmltox
 #Installs yarn without re-installing NodeJS
 curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
 echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
-
 sudo apt-get update && sudo apt-get install yarn --no-install-recommends
 
 #Run the following command to install git:
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -48,7 +48,10 @@ sudo mv wkhtmltox/bin/wkhtmltopdf /usr/bin
 sudo rm wkhtmltox-0.12.4_linux-generic-amd64.tar.xz  && rm -rf wkhtmltox
 
 #Installs yarn without re-installing NodeJS
-sudo apt-get install yarn --no-install-recommends
+curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
+echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+
+sudo apt-get update && sudo apt-get install yarn --no-install-recommends
 
 #Run the following command to install git:
 sudo apt-get install git
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -9,12 +9,11 @@ This guide will get you up and running with bhima locally.  Please note that bhi
 Before you begin the installation process, please make sure you have all the bhima dependencies installed locally.  We only test on Linux, so your best bet is to use a Linux flavor you are familiar with.  Please make sure you have recent version of:
  1. [MySQL](http://dev.mysql.com/downloads/) (5.6 or newer)
  2. [Redis](redis.io)
- 3. [NodeJS](https://nodejs.org/en/) (we recommend using [node version manager](https://github.com/creationix/nvm) on linux.  Note that we only test on stable and edge).
- 4. [WKHTMLtoPDF](http://wkhtmltopdf.org/downloads.html) (use the compiled binaries, even if it is distributed with your package manager.  The binaries come with patched Qt).
- 5.	npm
- 6.	curl
- 7.	yarn
- 8.	git
+ 3. curl
+ 4. [NodeJS](https://nodejs.org/en/) (we recommend using [node version manager](https://github.com/creationix/nvm) on linux.  Note that we only test on stable and edge).
+ 5. [WKHTMLtoPDF](http://wkhtmltopdf.org/downloads.html) (use the compiled binaries, even if it is distributed with your package manager.  The binaries come with patched Qt).
+ 6.	yarn
+ 7.	git
  
 ###### Detailed dependency installation instructions for Ubuntu (verified / installed specifically using VirtualBox)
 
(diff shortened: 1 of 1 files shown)


docs/INSTALL.md
```diff
@@ -11,9 +11,50 @@ Before you begin the installation process, please make sure you have all the bhi
...
+sudo apt-get update
+
+#Install MySQL with the following command:
+sudo apt-get install mysql-server
+
+#Run the following commands to install Redis:
+sudo apt-get install redis-server
+
+#Run the following command to install NodeJS:
+sudo apt-get install nodejs
```
jniles: In point 3, we recommend using [nvm](https://github.com/creationix/nvm), but these instructions do not use it.  My preference is to encourage users to use node version manager, since that guarantees what version of node they get when the install (their distro's package repositories may be terribly out of date).

I propose you update these instructions by moving the nodejs installation instructions below the wkhtmltopdf and use nvm to install node version 8 (the highest version we support without warnings).  Note that the installation uses `curl`, so curl will need to be installed before node.

```sh
#Install node version manager locally
curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash

#Set up the environmental variables for node version manager
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" # This loads nvm

#Download NodeJS version 8
nvm install 8
```

docs/INSTALL.md
```diff
@@ -11,9 +11,50 @@ Before you begin the installation process, please make sure you have all the bhi
...
+
+#Run the following commands to install WKHTMLtoPDF:
+sudo apt-get install xvfb
+wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
+tar xvf wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
+sudo mv wkhtmltox/bin/wkhtmltopdf /usr/bin 
+sudo rm wkhtmltox-0.12.4_linux-generic-amd64.tar.xz  && rm -rf wkhtmltox
+
+#Run the following commands to install npm:
+sudo apt-get install npm
```
jniles: If we use the `nvm` instructions, `npm` will automatically be downloaded and usable from the command line.

docs/INSTALL.md
```diff
@@ -11,9 +11,50 @@ Before you begin the installation process, please make sure you have all the bhi
...
+
+#Run the following commands to install curl:
+sudo apt-get install curl
+
+#Run the following commands to install yarn:
+curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
+echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+sudo apt-get update && sudo apt-get install yarn
```
jniles: Finally, if we go to the `nvm` instructions, the command to install yarn should be updated to:
```sh
#Installs yarn without re-installing NodeJS
sudo apt-get install yarn --no-install-recommends
```

This will force yarn to use the nvm version of node and not install its own copy.

docs/INSTALL.md
```diff
@@ -37,6 +78,8 @@ yarn install
 # bower is now installed in ./node_modules/.bin/bower
 # install client-side dependencies with bower
 ./node_modules/.bin/bower install -f
+#If this command gives you an error (I.E. if you’re running Parallels), try running the following command:
+git config -global url.”https://“.insteadOf git://
```
jniles, line 85: 👍

docs/INSTALL.md
```diff
@@ -50,8 +93,24 @@ the default node instance, `NODE_ENV="development"`.  Please set this globally,
...
-connection parameters.  Their variables should be self-explanatory. Then,
-build the app with
+connection parameters.  Their variables should be self-explanatory. 
+
+Use the following command to edit the .env.development file if desired (make your changes and then type ctrl + x to exit and save):
+```bash
+nano .env.development
+```
+
+###### Configure the bhima user in MySQL and build the app
```
jniles, line 106: Great catch.  👍  I hadn't realized we never documented the mysql user process.

jniles (commented):
@Scholldier super good work!  Having blow by blow instructions definitely will reduce the startup time for most users.  The documentation reads nicely as well.

I've made a few comments, mostly just suggesting the change to node version manager and the subsequent changes to support that.  Let me know what you think!

As soon as this gets in, I'll mirror it on the https://manual.bhi.ma site! 👍

Scholldier:
Great feedback Jonathan, thanks! I had googled all but nvm, I thought it was a misspelling of npm and didn't realize it was its own thing! I've added in your suggestions. Let me know how it looks!

Scholldier:
...ahh, one issue. It looks like yarn isn't installed correctly with the method utilized in the documentation. When I execute the next command, yarn install, it fails saying there is no such file or directory as 'install'. Does this mean we have to use the former code for installing yarn with curl?

jniles:
Hey @Scholldier, yeah the method before was perfect, save the addition of the "--no-install-recommends" flag.  You still need to have:
```sh
curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list

sudo apt-get update && sudo apt-get install yarn --no-install-recommends
```

Scholldier:
Right on! I just updated the file, and ran through the setup again. It worked great! There was one warning about an airbnb dependency not being linked or set properly (during yarn install), but it seemed to work fine and push past that. It was probably there before and I just saw it.

Should we leave the documentation as is regarding the MySQL user being given full privileges? If someone is just blindly installing the application with no experience they may not realize what that does, but again I think it's pretty self-evident. Just want to ask and make sure I'm following the proper train of thought.

jniles (approved):
Thanks! 

Regarding MySQL, let us just leave it as it is for now.  There are a few "production hacks", such as using gzip compression on nginx, setting up HTTPS with letsencrypt, etc, that would could add as a separate concern.  We could put all these best practice elements together for deployment in production as a separate step.  For now, the full privileges will help folks get up and running to tweak the system fast.

Everything looks great to me!  Thank you, this is a big help!

bors r+

jniles:
[Live now on the manual.bhi.ma site](https://manual.bhi.ma/for-developers/installing-bhima.html)

PR #3060, opened 2018-08-20
chore(price list): clean up code

jniles:
This commit removes unused HTML tags, properties, and directives that did not have bindings in either tests or the controller.  It also ensures that the rendering of the module looks like others: namely, the headercrumb of the modal is the same as well as the actions dropdown header.  Additionally, the loading indicator now works on "submit".


jniles:
This really deserves a much larger refactor.  I'll tackle it eventually.

PR #3061, opened 2018-08-20
feat(accounts) : importing accounts from csv file

### LINKED ISSUE #3034: feat(account): import accounts from a csv file
This issue is about to add a way to import accounts into bhima from an external csv file

### LINKED ISSUE #3076: Disable upload of Travis CI screenshot/artifacts
We currently upload from Travis CI to Amazon's S3 using [the artifacts addon](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/.travis.yml#L55).  We don't really use it enough to merit the cost of maintaining the S3 account (though it is cheap).  Let's get rid of it.

@mbayopanda can you handle this with the rest of your Travis CI changes?

mbayopanda:
This PR add the feature of importing accounts to use into the application from a csv file.

Close #3076 
Close #3034


client/src/modules/accounts/modals/import.js
```diff
@@ -0,0 +1,66 @@
...
+      Instance.close();
+    }
+
+    function handleError(err) {
+      Notify.handleError(err);
+      Instance.close();
+    }
+
+    // progress handler
+    function handleProgress(evt) {
```
jniles, line 65: How many places to we use this logic?  It looks like we use it in `src/modules/inventory/list/modals/import.modal.js` as well as `src/modules/patients/documents/modals/documents.modal.js`.

Since we use it in three place, I propose you make an issue for a directive that would do this calculation and rendering for you.  Something like:
```html
<bh-progress-meter current="currentProgress" total="totalProgress">
</bh-progress-meter>
```
That way, we can remove all this custom logic and just put the directive in the HTML.  We can also test it to make sure it works the way we hope it to work.

server/controllers/finance/accounts/import.js
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * @method importAccounts
+ *
+ * @description this method allow to do an import of accounts
+ */
+function importAccounts(req, res, next) {
+  const params = util.convertStringToNumber(req.query);
+
+  if (params.option !== IMPORT_DEFAULT_OHADA_ACCOUNTS && (!req.files || req.files.length === 0)) {
+    next(new BadRequest('Expected at least one file upload but did not receive any files.'));
```
jniles: We should add a translated error here - it is the second parameter on the `BadRequest()` constructor.

So you would do something like
```js
new BadRequest('Something broke', 'ERRORS.EVERYTHING_BAD');
```
and then translate the `ERRORS.EVERYTHING_BAD` in the `i18n` files.

server/controllers/finance/accounts/import.js
```diff
@@ -0,0 +1,137 @@
...
+
+  const file = req.files[0];
+  const dbPromises = [];
+
+  if (params.option === IMPORT_DEFAULT_OHADA_ACCOUNTS || params.option === IMPORT_CUSTOM_OHADA_ACCOUNTS) {
+    const basicAccountFile = path.join(__dirname, '../../../resources/templates/ohada-main-accounts.csv');
+    dbPromises.push(importAccountFromFile(basicAccountFile, req.session.enterprise.id, params.option));
+  }
+
+  if (params.option === IMPORT_CUSTOM_OHADA_ACCOUNTS || params.option === IMPORT_OTHER_ACCOUNTS) {
```
jniles, line 58: Why do we need `params.option === IMPORT_CUSTOM_OHADA_ACCOUNTS` twice?  Could you leave a comment in the code?  It looks like a bug from just reading the code.

server/controllers/finance/accounts/import.js
```diff
@@ -0,0 +1,137 @@
...
+ * @method importAccountFromFile
+ * @description import accounts from a given file
+ * @param {object} file the file object
+ * @param {number} enterpriseId the enterprise id
+ * @param {number} option the option (1 | 2 | 3) sent by the client
+ */
+function importAccountFromFile(filePath, enterpriseId, option) {
+  let query;
+  let queryParams;
+  return formatCsvToJson(filePath)
```
jniles: This is an ideal place for an `async`/`await` function!  You could do this:
```js
async function importAccountFromFile(filePath, enterpriseId, option) {
  const data = await formatCsvToJson(filePath);

  if (!hasValidDataFormat(data)) {
    throw new BadRequest('The given file has a bad data format for accounts', 'ERRORS.BAD_DATA_FORMAT');
  }

  const transaction = db.transaction();

  data.forEach(item => {
    query = 'CALL ImportAccount(?, ?, ?, ?, ?, ?);';
    queryParams = [
      enterpriseId,
      item.account_number,
      item.account_label,
      item.account_type,
      item.account_parent || null,
      option,
    ];
    transaction.addQuery(query, queryParams);
  });
  
  return transaction.execute();
}
```

server/controllers/finance/accounts/import.js
```diff
@@ -0,0 +1,137 @@
...
+  return data.every(item => {
+    return item.account_number && item.account_label && item.account_type;
+  });
+}
+
+/**
+ * formatCsvToJson
+ * @param {object} file the csv file sent by the client
+ */
+function formatCsvToJson(file) {
```
jniles: Looks like this function is a copy and paste of `controllers/inventory/import/index.js`'s `formatCsvToJson()` function.  Instead of copying and pasting the code, could we make a library for this and re-use the same code?  It would also allow us to unit test this function.

server/models/functions.sql
```diff
@@ -91,4 +91,53 @@ BEGIN
   );
 END $$
 
+/*
+  PredictAccountTypeId(accountNumber)
+
+  Returns the account type id of the given account number
+*/
+CREATE FUNCTION PredictAccountTypeId(accountNumber INT(11))
```
jniles, line 99: Why does this need to be an SQL function?  It doesn't look like it uses the database at all...

Can we do this in JavaScript?  SQL functions are a pain to upgrade in production.

server/controllers/finance/accounts/import.js
mbayopanda: Twice in this condition or in this line and the line 52 ?

We have three options for loading accounts into the system : 
- 1. Load default ohada accounts : this option doesn't require to send a csv file, so the system load the file directly from the server side.

- 2. Load custom ohada accounts : in this option, the user must add ohada accounts (by uploading a csv file) to default ohada accounts which are already loaded into the system.

- 3. Load accounts : this option is more generic, if the user need to load any kind of accounts (not ohada) into the system. the user must upload a csv file of accounts.

In the line 52, we test the first or the second option, in this case we must load ohada default accounts anyway.

In the line 57, we test the second or the third option, in this case we must load accounts from uploaded csv file of accounts.

server/models/functions.sql
mbayopanda: The simple reason is this function is just used when we are in the option 1 and the account number has four digit, and the most important we cannot send this value (the predicted account type id) as parameters because it is not needed at anytime we call the ImportAccount procedure

server/controllers/finance/accounts/import.js
mbayopanda: it is not totally a copy pasta, the line 120 is different (file.path versus file).

You are right, we can put this function in a library for unit test it.

mbayopanda:

server/controllers/finance/accounts/import.js
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * @method importAccountFromFile
+ * @description import accounts from a given file
+ * @param {object} file the file object
+ * @param {number} enterpriseId the enterprise id
+ * @param {number} option the option (1 | 2 | 3) sent by the client
+ */
+function importAccountFromFile(filePath, enterpriseId, option) {
+  let query;
```
jniles: It's probably better practice to move these to `const` within the `forEach()` loop.

sh/travis.sh
```diff
@@ -1,14 +1,40 @@
...
-mysql -h $DB_HOST -u root -e "SET GLOBAL sql_mode='STRICT_ALL_TABLES';"
-
 # setup usernames and permissions
 mysql -h $DB_HOST -u root -e "GRANT ALL PRIVILEGES ON *.* TO '$DB_USER'@'$DB_HOST' IDENTIFIED BY '$DB_PASS' WITH GRANT OPTION;"
 mysql -h $DB_HOST -u root -e "FLUSH PRIVILEGES;"
+
+## install and configure softwares
+export CHROME_BIN=/usr/bin/google-chrome
+/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16
+sudo apt-get update -qq
```
jniles: Since we are no longer installing MySQL, this line is useless.  Let's remove it.

sh/travis.sh
```diff
@@ -1,14 +1,40 @@
...
+wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
```
jniles: Let's try removing Google Chrome from out build completely.  It looks like [Travis supports google chrome as an addon](https://docs.travis-ci.com/user/chrome).  I propose you remove all references to Chrome here and then add:
```yml
addons:
  chrome: stable
```
to the `.travis.yml` file.

test/server-unit/util.spec.js
```diff
@@ -63,4 +64,42 @@ describe('util.js', () => {
     const result = util.renameKeys(a, keyMap);
     expect(result).to.deep.equal([]);
   });
+
+  it('#formatCsvToJson should return a json from a csv file', () => {
+    /**
+     * The structure of the sample csv file (ohada-accounts.csv)
```
jniles, line 70: Fantastic comment :)

test/server-unit/util.spec.js
```diff
@@ -63,4 +64,42 @@ describe('util.js', () => {
...
+     * "account_number",  "account_label",    "account_type", "account_parent"
+     * "10",              "CAPITAL",          "title",        "1"
+     * "12",              "REPORT A NOUVEAU", "title",        "1"
+     */
+    const filePath = 'test/fixtures/ohada-accounts.csv';
+    const promise = util.formatCsvToJson(path.resolve(filePath));
+    return promise
+      .then(csvObjectArray => {
+        const [first] = csvObjectArray;
+        const second = csvObjectArray[1];
```
jniles: Did you know you could do this?
```js
const [first, second] = csvObjectArray;
```

.travis.yml
```diff
@@ -2,20 +2,8 @@ sudo: required
...
-  - wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
-  - sudo dpkg -i google-chrome*.deb
-  - wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
-  - tar -xf wkhtmltox-0.12.4_linux-generic-amd64.tar.xz
-  - sudo mv ./wkhtmltox/bin/wkhtmltopdf  /usr/bin/wkhtmltopdf
-  - yarn global add greenkeeper-lockfile@1
-  - export GK_LOCK_YARN_OPTS="--ignore-engines"
-
-
+  - ./sh/travis.sh
```
jniles, line 6: This is awesome :+1:

client/src/modules/accounts/accounts.routes.js
```diff
@@ -19,19 +19,30 @@ function accountStateProvider($stateProvider) {
...
       params : {
         id : { squash : true, value : null },
       },
       onEnter : ['$uibModal', accountsModal],
       onExit : ['$uibModalStack', closeModal],
+    })
+
+    .state('accounts.import', {
+      url : '/import',
+      params : {
```
jniles: It looks like this is unused.  Let's remove it.

client/src/modules/accounts/accounts.routes.js
```diff
@@ -44,6 +55,15 @@ function accountsModal($modal) {
   });
 }
 
+function importAccountsModal($modal) {
+  $modal.open({
+    keyboard : false,
```
jniles: Just FYI - the `keyboard: false` and `backdrop: 'static'` are set by default [in app.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/app.js#L208).

client/src/modules/accounts/accounts.service.js
```diff
@@ -2,17 +2,17 @@ angular.module('bhima.services')
...
 
 /**
  * Account Service
  *
  * A service wrapper for the /accounts HTTP endpoint.
  */
-function AccountService(Api, bhConstants) {
-  var baseUrl = '/accounts/';
-  var service = new Api(baseUrl);
+function AccountService(Api, bhConstants, util) {
```
jniles: There isn't any need to inject `util` here.  It's already available on `service` as `service.util`.

client/src/modules/accounts/accounts.service.js
```diff
@@ -2,17 +2,17 @@ angular.module('bhima.services')
   .service('AccountService', AccountService);
 
 AccountService.$inject = [
-  'PrototypeApiService', 'bhConstants',
+  'PrototypeApiService', 'bhConstants', 'util',
 ];
 
 /**
  * Account Service
```
jniles: Good cleanup of ESLint here :+1:

client/src/modules/accounts/modals/import.html
```diff
@@ -0,0 +1,97 @@
...
+            ng-value="1" 
+            ng-change="ImportAccountsCtrl.onChange(ImportAccountsCtrl.option)">
+          <span translate>ACCOUNT.IMPORT.LOAD_DEFAULT_OHADA_ACCOUNTS</span>
+        </label>
+      </div>
+      <div class="radio">
+        <label>
+          <input type="radio" name="importingOption" id="import_ohada_accounts_option" 
+            ng-model="ImportAccountsCtrl.option" 
+            ng-value="2"
```
jniles: To make this a bit more readable, it would be nice to make these numbers into variables attached to the controller.  That way, we can give them a name.  Like this:
```js
vm.LOAD_OHADA_ACCOUNTS = 2
```
then you can use it in your HTML like:
```html
<input type="radio" name="importingOption" id="import_ohada_accounts_option" 
            ng-model="ImportAccountsCtrl.option" 
            ng-value="ImportAccountsCtrl.LOAD_OHADA_ACCOUNTS">
```

It will just make it easier to read.

client/src/modules/accounts/modals/import.js
```diff
@@ -0,0 +1,68 @@
...
+  const IMPORT_DEFAULT_OHADA_ACCOUNTS = 1;
+
+  vm.downloadTemplate = Accounts.downloadAccountsTemplate;
+  vm.cancel = Instance.close;
+
+  vm.select = (file) => {
+    vm.noSelectedFile = !file;
+  };
+
+  vm.submit = () => {
```
jniles, line 22: The `bhLoadingButton` will only show a loading indicator if the `submit()` method returns a promise.  I propose you add `return uploadFile(vm.file)` to this method, and then add `return Upload.upload()` to the `uploadFile()` method to get a nice loading indicator.

jniles: Also, should this submit method check for `form.$invalid`?

server/controllers/finance/accounts/import.js
```diff
@@ -0,0 +1,112 @@
...
+    const basicAccountFile = path.join(__dirname, '../../../resources/templates/ohada-main-accounts.csv');
+    dbPromises.push(importAccountFromFile(basicAccountFile, req.session.enterprise.id, params.option));
+  }
+
+  if (params.option === IMPORT_CUSTOM_OHADA_ACCOUNTS || params.option === IMPORT_OTHER_ACCOUNTS) {
+    dbPromises.push(importAccountFromFile(file.path, req.session.enterprise.id, params.option));
+  }
+
+  Q.all(dbPromises)
+    .then(() => res.sendStatus(200))
```
jniles: Since this is CREATE-ing accounts via a POST request, the best response code here is 201.

server/controllers/inventory/import/index.js
```diff
@@ -35,15 +34,15 @@ function downloadTemplate(req, res, next) {
  */
 function importInventories(req, res, next) {
   if (!req.files || req.files.length === 0) {
-    next(new BadRequest('Expected at least one file upload but did not receive any files.'));
+    next(new BadRequest('Something broke', 'ERRORS.EVERYTHING_BAD'));
```
jniles, line 37: You took what I said a bit too literally ;).  The 'Expected at least one file upload but did not receive any files.' is a perfect translation, but we need to send a key with that translation back to the client.  So, something like:
```js
next(new BadRequest('Expected at least one file upload but did not receive any files.', 'ERRORS.MISSING_UPLOAD_FILES'));
```
where `ERRORS.MISSING_UPLOAD_FILES` is "Expected at least one file upload but did not receive any files." in English.  Does that makes sense?

test/server-unit/util.spec.js
```diff
@@ -63,4 +64,42 @@ describe('util.js', () => {
     const result = util.renameKeys(a, keyMap);
     expect(result).to.deep.equal([]);
   });
+
+  it('#formatCsvToJson should return a json from a csv file', () => {
+    /**
+     * The structure of the sample csv file (ohada-accounts.csv)
+     * =========================================================
+     * "account_number",  "account_label",    "account_type", "account_parent"
```
jniles, line 72: Can we call this column (and the others) "account number" instead of "account_number"?  It would be easier for our users to type.

jniles (commented):
@mbayopanda, great work on Travis!  This is pretty much ready.  I've made a few quick comments, but it can be merged pretty much anytime.  Let me know when your fixes have landed.

mbayopanda:

server/controllers/inventory/import/index.js
jniles: I think you forgot to correct this!

jniles (approved):
Looks fine to me!

bors r+

PR #3064, opened 2018-08-21
Fixed fiscal year/period selection input fields to improve error handling

### LINKED ISSUE #3039: UI for Reports/Balance of Accounts a little unfriendly

jmcameron:
fixes #3039 

This fixes the error handling in "Reports/Balance of Accounts" report when the user presses the [Preview] button without selecting the fiscal year and/or period.   

**Notes:**  I had to update the <bh-select-fiscal-year> input field replacement html in the templates file to get this to work.  There were several fixes that were needed in the two inputs files:

- $ctrl.validationTrigger does not seem to work.  When I replaced them with  xyzForm.$submitted, they worked.
- The period select form did not have the ng-form angular setup at the top so even xyzForm.$submitted did not work until I added it.
- The fiscal year selection field did not have the error-only message beneath it so I added it.  The period selection one did.

I tried these fields out on on several other reports such as the aged debtors/creditors and they seem to work okay.   I did not do an exhaustive check on all reports that use these input fields.

It is likely that all of these input field templates need review may need some TLC.

-Jonathan C

### DIFF
--- a/client/src/modules/reports/generate/balance_report/balance_report.config.js
+++ b/client/src/modules/reports/generate/balance_report/balance_report.config.js
@@ -40,7 +40,7 @@ function BalanceReportConfigController($sce, Notify, SavedReports, AppCache, rep
   vm.preview = function preview(form) {
     if (form.$invalid) {
       Notify.danger('FORM.ERRORS.RECORD_ERROR');
-      return;
+      return 0;
     }
 
     // update cached configuration
--- a/client/src/modules/reports/generate/balance_report/balance_report.config.js
+++ b/client/src/modules/reports/generate/balance_report/balance_report.config.js
@@ -38,7 +38,10 @@ function BalanceReportConfigController($sce, Notify, SavedReports, AppCache, rep
   };
 
   vm.preview = function preview(form) {
-    if (form.$invalid) { return 0; }
+    if (form.$invalid) {
+      Notify.danger('FORM.ERRORS.RECORD_ERROR');
+      return;
+    }
 
     // update cached configuration
     cache.reportDetails = angular.copy(vm.reportDetails);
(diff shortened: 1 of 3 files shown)


client/src/modules/reports/generate/balance_report/balance_report.config.js
```diff
@@ -38,7 +38,10 @@ function BalanceReportConfigController($sce, Notify, SavedReports, AppCache, rep
   };
 
   vm.preview = function preview(form) {
-    if (form.$invalid) { return 0; }
+    if (form.$invalid) {
+      Notify.danger('FORM.ERRORS.RECORD_ERROR');
```
jniles, line 42: :+1:

jniles (approved):
This looks right to me.

The likely issue with the `$ctrl.validationTrigger` is that it wasn't being passed in to the component correctly.  The reason we use `validationTrigger` is that, until recently, AngularJS did not submit child forms when the parent was submitted.  However, [they fixed this behavior in v1.7.0](https://github.com/angular/angular.js/issues/10071) ([release notes](https://github.com/angular/angular.js/blob/master/CHANGELOG.md#170-nonexistent-physiology-2018-05-11)).  I've been hoping to move in this direction (checking `$submitted` on child forms), so I am glad to see this fix!

PR #3065, opened 2018-08-22
test(ExchangeRate): write client unit tests

jniles:
This commit writes unit tests for the exchange rate service.  Only a few have been written so far, but more coverage will be added later.

@lomamech

### DIFF
--- a/test/client-unit/services/ExchangeRateService.spec.js
+++ b/test/client-unit/services/ExchangeRateService.spec.js
@@ -1,0 +1,117 @@
+/* global expect inject */
+describe('ExchangeRateService', () => {
+  let Exchange;
+  let $httpBackend;
+  let Session;
+  let Mocks;
+
+  beforeEach(module(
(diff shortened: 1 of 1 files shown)


test/client-unit/services/ExchangeRateService.spec.js
```diff
@@ -0,0 +1,130 @@
...
+
+    // currency_id 1 is the enterprise currency, so it is always 1
+    let rate = Exchange.getCurrentRate(1);
+    expect(rate).to.equal(1);
+
+    rate = Exchange.getCurrentRate(2);
+    expect(rate).to.equal(2000);
+  });
+
+  it('#getCurrentRate() should call getExchangeRate()', () => {
```
jniles: I'm pretty sure this doesn't work.  I've tried making the test a bit more specific and it doesn't seem to behave.  I'll have to look into it.

Issue #3067, opened 2018-08-22
Remove uuids from Excel/CSV exports


jniles:
Every registry that we download usually comes with a couple of columns of `uuid`s.  These are not useful for a user, and our users might now be technically savvy enough to hide or remove them themselves.  Let's remove all uuid columns from the Excel and CSV exports!

jniles:
Closed in #3070

PR #3070, opened 2018-08-23
feat(xlsx renderer) specify ignored columns

jeremielodi:
trying to solve this #3067

### DIFF
--- a/server/lib/renderers/xlsx.js
+++ b/server/lib/renderers/xlsx.js
@@ -99,8 +99,22 @@ function render(data, template, options) {
  it return an empty array if no data found
 */
 function find(data, _options) {
+  let IGNORED_COLUMNS = [
+    'uuid', 'invoice_uuid', 'entity_uuid',
+    'reference_uuid', 'record_uuid', 'debtor_uuid',
+  ];
   const options = _options || {};
-  return data.rows || data[options.rowsDataKey] || [];
+  const result = data.rows || data[options.rowsDataKey] || [];
+  // check if it's important to remove some unseful columns for the user
+  if (options.ignoredColumns) {
+    IGNORED_COLUMNS = IGNORED_COLUMNS.concat(options.ignoredColumns);
+  }
+  result.forEach(row => {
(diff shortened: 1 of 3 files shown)


test/server-unit/xlsx.js
```diff
@@ -53,4 +53,21 @@ describe('util.js', () => {
     expect(data.students).to.deep.equal(result);
   });
 
+  it('Should remove unuseful columns for the user such as uuid', () => {
```
jniles, line 56: Beautiful test 👍

server/lib/renderers/xlsx.js
```diff
@@ -100,7 +100,17 @@ function render(data, template, options) {
 */
 function find(data, _options) {
   const options = _options || {};
-  return data.rows || data[options.rowsDataKey] || [];
+  const result = data.rows || data[options.rowsDataKey] || [];
+  // check if it's important to remove some unseful columns for the user
+  if (options.ignoredColumns) {
```
jniles, line 109: This works!

jniles (approved):
LGTM.

jniles:
@jeremielodi, this is good.  Could you implement a global `IGNORED_COLUMNS` array that always ignores uuids?  You can then combine that with any columns sent from the client.  I'm imagining like this:
```js
const IGNORED_COLUMNS = ['uuid', 'invoice_uuid', 'entity_uuid', 'reference_uuid' /* etc... */];

// inside your find() method
const columnsToIgnore = IGNORED_COLUMNS.concat(ignoredColumns);
rows.forEach(row => {
  // filter out columns, etc ...
});
```

That way, most registries will automatically filter out useless junk.

jeremielodi:
This is a good idea I see but it is important to add a flag saying that uuid's colums should be removed.
Maybe we can at some point need to use the xlsx file for another action

jniles:
Good point - eventually we'll need a way for the user to control it.  But for the moment, let's just aggressively filter them out.

jniles (approved):
LGTM.

bors r+

PR #3071, opened 2018-08-23
feat(user creation) display_name validation error

### LINKED ISSUE #3058: Async validation for user display name is unclear
We have an async validator for the display name of a user on the user creation form.  I had forgotten it exists because it is so discrete.  Here it is below:

erroronduplicateusername
_Fig 1: Async validator for display name_

The above indicator shows an error for a duplicate display name - "admin" has already been used.  The following changes need to be made:
 1. Include text that says "this display name is already in use"
 2. Color the input, and the "X" and any accompanying text, in red.

jeremielodi:
closes #3058

### DIFF
--- a/client/src/modules/users/user.modal.html
+++ b/client/src/modules/users/user.modal.html
@@ -16,21 +16,26 @@
 
   <div class="modal-body" style="overflow-y: scroll; max-height:600px; ">
 
-    <div class="form-group" ng-class="{ 'has-error' : UserForm.$submitted && UserForm.display_name.$invalid }">
+    <div class="form-group has-feedback" ng-class="{ 'has-error' : (UserForm.$submitted || UserForm.display_name.$error.unique) && UserForm.display_name.$invalid }">
       <label class="control-label" translate>FORM.LABELS.USERNAME</label>
-      <div>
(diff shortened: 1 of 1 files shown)


jniles:
Could you include a screenshot?

jeremielodi:
val

client/src/modules/users/user.modal.html
```diff
@@ -21,7 +21,10 @@
       <div>
         <input name="display_name" bh-unique="/users/{{UserModalCtrl.user.display_name}}" ng-model="UserModalCtrl.user.display_name" autocomplete="off" ng-maxlength="UserModalCtrl.maxLength" class="form-control" required>
 
-        <span ng-show="UserForm.display_name.$error.unique" class="glyphicon glyphicon-remove form-control-feedback" id="unique-error-icon"></span>
+        <div ng-show="UserForm.display_name.$error.unique" translate class="text-danger">
```
jniles: From your screenshot, looks like you need to make this a `<span>` and have [class="help-block"](https://getbootstrap.com/docs/3.3/css/#forms-help-text) on it, to make it behave the same as the other help blocks.

jniles:
I've switched up the classes a bit based on [bootstrap's documentation](http://getbootstrap.com/docs/3.3/css/#forms-control-validation).  It looks like the above.

PR #3075, opened 2018-08-23
Fixes to error handling for Cash and CashFlowByService reports

jmcameron:
Fixed error handling for cash-box selection and date interval input fields.   Updated cash_report and cashflowByService reports to use the improved error handling.

NOTE:  
All uses of the **bh-date-interval** input field must insert
        required="true"'
in the directive to get the error handling to properly handle missing From/To dates -- only do this if both the fields are both required.     By putting required="true" in the directive, the missing field data check is done on the client side rather than the server side (which gives error messages that are not very helpful).    The checks on the server side are still there in server/controllers/finance/reports/cashReport/index.js and should be kept as a final check.

-Jonathan C


client/src/modules/templates/bhCashboxSelect.tmpl.html
```diff
@@ -1,23 +1,25 @@
...
+	       required 
+	       on-select="$ctrl.onSelect($item, $model)">
+      <ui-select-match placeholder="{{ 'FORM.SELECT.CASHBOX' | translate }}"><span translate>{{$select.selected.hrlabel}}</span></ui-select-match>
       <ui-select-choices ui-select-focus-patch repeat="cashbox.id as cashbox in ($ctrl.cashboxes | filter:$select.search) track by cashbox.id">
         <span ng-bind-html="cashbox.hrlabel | highlight:$select.search"></span>
       </ui-select-choices>
     </ui-select>
 
-    <div class="help-block" ng-messages="CashboxForm.$error" ng-show="$ctrl.validationTrigger">
+    <div class="help-block" ng-messages="CashboxForm.$error" ng-show="CashboxForm.$submitted">
```
jniles, line 21: Beautiful 👍  I've been waiting for this change for a while.

client/src/modules/reports/generate/cashflowByService/cashflowByService.config.js
```diff
@@ -40,7 +40,10 @@ function CashFlowByServiceConfigController($sce, Notify, SavedReports, AppCache,
   };
 
   vm.preview = function preview(form) {
-    if (form.$invalid) { return 0; }
+    if (form.$invalid) {
+      Notify.danger('FORM.ERRORS.RECORD_ERROR');
```
jniles, line 44: 👍

client/src/modules/reports/generate/cash_report/cash_report.html
```diff
@@ -66,18 +64,17 @@ <h3 class="text-capitalize" translate>REPORT.CASH_REPORT.TITLE</h3>
...
             <!-- Date interval  -->
             <bh-date-interval
-                date-from="ReportConfigCtrl.reportDetails.dateFrom"
-                date-to="ReportConfigCtrl.reportDetails.dateTo"
-                required
-                validation-trigger = 'ConfigForm.$submitted'
-                >
+              date-from="ReportConfigCtrl.reportDetails.dateFrom"
+              date-to="ReportConfigCtrl.reportDetails.dateTo"
+              required="true"
```
jniles, line 76: Good catch!  🐛

jniles (commented):
All these changes look excellent!

It appears our bower doesn't fetch the latest versions of AngularJS.  I'm going to do a quick version bump and then see how this works out locally.  If all goes well, should be merged by this evening.  Thanks!

jmcameron:
FYI, It looks like I'm running Angular 1.7.0 on my laptop where I did these fixes.

jniles:
Works well on my machine!

bors r+

jniles:
Okay, now that I am finally in good enough internet to look at this, it seems like this changes the behavior of the fiscal year page.  I'll have to download it and poke around to see if I can figure out what is going on.

client/src/modules/templates/bhCurrencySelect.tmpl.html
```diff
@@ -1,5 +1,11 @@
-<ng-form name="$ctrl.form">
-  <div class="radio" ng-class="{ 'has-error' : ($ctrl.validationTrigger && $ctrl.form.$invalid) || !$ctrl.valid }" data-bh-currency-select>
+<fieldset
+  ng-form="CurrencySelectForm"
+  novalidate
+  bh-currency-select>
```
jniles: Looks like this needs to have "data-" appended to it for the cash tests to pass!

jmcameron: Ok, pushed a fix to this.  Lets see if it works!

client/src/modules/templates/bhDateInterval.tmpl.html
```diff
@@ -17,16 +19,16 @@
...
-    <li class="pull-right" ng-if = "$ctrl.canClear">
+    <li class="pull-right" ng-if="$ctrl.canClear">
       <a href ng-click="$ctrl.clear()" tabindex="-1">
         <i class="fa fa-eraser"></i> <span translate>FORM.BUTTONS.CLEAR</span>
       </a>
     </li>
   </ul>
 
-  <div class="row" date-id="$ctrl.dateId" data-bh-date-interval>
+  <div class="row" date-id="$ctrl.dateId">
```
jniles: Aha!  Located it - the tests are failing because the [data-bh-date-interval tag is used to locate this element on the page](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/shared/components/bhDateInterval.js#L13).  I think if you restore this tag, it should pass tests.

jniles:
@jmcameron thanks for your work on this!

Since you've been messing in the `bhDateInterval` component, I discovered a bug that you could fix in this PR.  In the component controller, [line 44](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/components/bhDateInterval.js#L44), the `canClear` isn't set properly.  It can never be false!  Could you update the line:
```js
vm.canClear = (vm.canClear || true);
```
to be 
```js
if (!angular.isDefined(vm.canClear)) {
  vm.canClear = true;
}
```

jniles:
This is awesome!   Looks good to me.

bors r+

PR #3078, opened 2018-08-24
feat(bhMaxInteger): error on large integers

### LINKED ISSUE #3072: Block large numbers from being submitted as the search limit
We hit a very interesting bug in production where the LIMIT was set to an extremely high value, causing the database to break.  We must block this on two fronts: the clientside and the serverside.

First, we must set a `ngMax` directive on the `limit` input of all search modals.  This should be set to the maximum size of a MySQL integer (`18446744073709552000`).  Any larger should produce an error message telling the user to lower their value.

Second, we must either throw an error or truncate large values passed into the LIMIT field in FilterParser.

jniles:
This commit adds the bhMaxIntegerValidator and applies it to all limit inputs in the application.  The bhMaxIntegerValidator blocks numbers that are larger than MySQL's MEDIUMINT.

Closes #3072.

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -774,7 +774,7 @@
             "EXCEPTION": "There was a problem validating this field. Please check your internet connection.",
             "INTEGER": "This value must be a whole number (integer).",
             "INVALID_DATE": "Invalid date",
-            "MAX": "Maximum value not respected",
+            "MAX": "This value is too large!  Choose a lower value.",
             "MAXLENGTH": "This value is too long!",
             "MIN": "This value is too small.",
             "MINLENGTH": "This value is too short",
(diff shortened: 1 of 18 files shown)


jniles:

jniles:

client/src/i18n/en/form.json
```diff
@@ -774,7 +774,7 @@
             "EXCEPTION": "There was a problem validating this field. Please check your internet connection.",
             "INTEGER": "This value must be a whole number (integer).",
             "INVALID_DATE": "Invalid date",
-            "MAX": "Maximum value not respected",
+            "MAX": "This value is too large!  Choose a lower value.",
```
sfount: I think this is much clearer feedback for the user! Could this be applied to `fr/form.json` also? 

Current value:
```
"MAX": "Valeur maximale non respectée"
```

sfount:
I think ` "MAX": "This value is too large!  Choose a lower value.",` in favour of ` "MAX": "Maximum value not respected",` is much clearer feedback for the user! 

Could this be applied to `fr/form.json` also? 

Current value:
```
"MAX": "Valeur maximale non respectée"
```

Edit: Commenting inline in the code counted as a PR review - moved to discussion for clarity!

jniles:
@sfount thanks for your review!  Very helpful suggestions.

I've made the change to the French form.  We should probably go through out errors and make sure they make sense to our users.

jniles:
bors r+

(let's see if this gets it in one try...)

jniles:
... not even close.

PR #3079, opened 2018-08-26
Scholldier issue 2116

### LINKED ISSUE #2116: Feature Request: Link Cashflow Report from Cash Pages
One way to get our reports easily used is to directly link them from their respective modules.  One module that could really benefit from this is the Cash Payments module and the Cash Registry.  We could include a direct link to the Cashflow Report the current selected cashbox like this:

dailycashflowlink
_Fig 1: Example Cashflow Link_

In my view, this would put the report into the hands of people who create the data - not just administrators.  If the accountant did something strange and moved money in or out of the Caisse account without the Caissier knowing, they can see that in the report and inform the administration of the mistake.

Scholldier:
This resolves issue 2116 -- adding a link to the Cashflow Report from the Cash Window module.

Closes #2116.

### DIFF
--- a/client/src/i18n/en/cash.json
+++ b/client/src/i18n/en/cash.json
@@ -8,6 +8,7 @@
     "CONFIRM_PAYMENT_WHEN_CAUTION":"The debtor has a positive balance with the hospital.  Accepting cash from the debtor will increase their credit with the hospital.  Are you sure you want to accept payment?",
     "CURRENT_CASHBOX":"Current Cashbox",
     "DEBTOR_INVOICES":"Debtor Invoices",
+    "LINK_TO_CASHFLOW":"Link to Cashflow Report",
     "PAYMENT":"Payment",
     "PAYMENT_DESCRIPTION":"Cash payment by {{patientName}} ({{patientReference}}) against invoice(s) {{invoiceReferences}}.",
     "PREPAYMENT_DESCRIPTION":"Prepayment by {{patientName}} ({{patientReference}}).",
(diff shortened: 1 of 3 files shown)


jniles (commented):
Looks good to me and the behavior works according to spec.  Well done!

For the last thing we need, we'll want to have a translated text.  The way you accomplish translations in BHIMA is by using the `translate` directive from the excellent [angular-translate](https://angular-translate.github.io/) project.  Any text put within a translation tag will be used as the key to lookup a translation in the current rendered language.

An example of this is directly about the change you made:
```html
<span translate>BARCODE.SCAN</span>
```
The above HTML will look up "BARCODE.SCAN" from [our i18n files](https://github.com/IMA-WorldHealth/bhima-2.X/tree/master/client/src/i18n) and substitute whatever matching text it finds into the `<span>` element.  Specifically, it matches [this line in English](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/i18n/en/barcode.json#L3) and [this line in French](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/i18n/fr/barcode.json#L3).

Notice that the "." corresponds to a hierarchy - the "SCAN" key is located within the "BARCODE" object.  By convention, we capitalize all keys. You'll probably want to add a new key to the [cash.json](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/i18n/en/cash.json) called "LINK_TO_CASHFLOW" and add this translated text to it.

While it isn't necessary that you do the French translations yourself (though you are welcome to try!), we should at least set up the infrastructure for the translations by doing the English one.  Someone else can go through and touch up the translations later.

If you need any further assistance, feel free to give us a holler!

Scholldier:
Awesome write-up Jonathan, thank you so much! I really appreciate the detail you provide. 👍 

I'm also very impressed with how the translation works. It's seamless and so cool! Hopefully I translated the text correctly...!

jniles:
Looks excellent!

bors r+

Issue #3057, opened 2018-08-20
User Management should be sorted by display_name


jniles:
Our user management comes out in the order of insertion into the database.  This is quite annoying when trying to find a user (although having an inline filter helps).  It would be nice if the users came out in alphabetical order on `display_name`.

sfount:
This issue was closed with https://github.com/IMA-WorldHealth/bhima-2.X/pull/3081.

PR #3081, opened 2018-08-26
Update users.js

Scholldier:
This resolves issue #3057 and sorts the display name column alphabetically.


jniles (approved):
Excellent!  Looks good to me.  I think I may have to dig into the tests to make sure all is well, which I will do after church.  But great work!

jniles:
Okay, after a brief investigation, my current hypothesis is that our users are assigned permissions in the test based on their ordering in the user grid.  After the order shifted alphabetically, the correct user int' being assigned permission to access the depot - which is why the stock tests failed.

This is a larger issue with the way the user protractor test was designed.  Rather than giving a permission to an index in the grid, we should be clicking on the user based on their name ... after all, this is how a user would locate the user.  I'll nail it down and submit a patch for this issue to the PR.  I have a feeling it will be fairly extensive.

jniles:
Alright, unless I've made a lint mistake, I'm pretty sure I've finally cracked this one.  Took a while though.

PR #3086, opened 2018-08-27
feat(report)exporting report

### LINKED ISSUE #2538: Export Receipts/Reports as Excel

jeremielodi:
This PR help to export our reports as an excel or a word document.
In order to perform these functionalities, the application get the  report html content with all it css styles and download it as ```xls``` or ```doc```

Closes #2538


client/src/i18n/en/report.json
```diff
@@ -156,6 +156,8 @@
       "PDF": "PDF",
       "HTML": "HTML",
       "CSV": "CSV",
+      "WORD" : "WORD",
```
jniles: We should probably call these "MS Word" and "MS Excel"

jeremielodi:
Oups! Getting all styles takes too much time when the reports is long I have to optimize the process

jniles:
Note: this will close #2538

client/src/modules/reports/reports.save.js
```diff
@@ -17,12 +17,15 @@ function SaveReportController(ModalInstance, Notify, SavedReports, options, Lang
   //       it should also be possible to override this depending on the support of each individual report
   vm.supportedRenderTypes = [
     { key : 'pdf', label : 'REPORT.UTIL.PDF' },
+    { key : 'doc', label : 'REPORT.UTIL.WORD' },
+    { key : 'xls', label : 'REPORT.UTIL.EXCEL' },
```
jniles, line 21: Nice :+1:

server/lib/renderers/doc.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @description
+ * This library is just a shim to ensure API uniformity with other renderers.
+ * it renders an Word report from html.
+ * @module lib/renderers/xls
+ * @requires juice
+ * @requires q
+ * @requires lodash
```
jniles: It doesn't look like you use `lodash` here.  Can you remove this comment?

server/lib/renderers/doc.js
```diff
@@ -0,0 +1,30 @@
...
+const headers = {
+  'Content-Type' : 'application/vnd.ms-word;charset=utf-8',
+};
+
+exports.render = render;
+exports.extension = '.doc';
+exports.headers = headers;
+
+
+async function render(data, template, options) {
```
jniles, line 21: Nice use of async/await!

server/lib/renderers/doc.js
```diff
@@ -0,0 +1,30 @@
...
+  const htmlStream = await html.render(data, template, options);
+  const htmlString = juice(htmlStream);
+  const deferred = q.defer();
+  const buffer = Buffer.from(htmlString, 'utf8');
+  deferred.resolve(buffer);
+  return deferred.promise;
```
jniles: There shouldn't be any need to create a promise here.  Since you are using `async`/`await`, it will automatically return a promise.  I think you can probably just do:
```js
return Buffer.from(htmlString, 'utf8');
```

server/lib/renderers/xls.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @description
+ * This library is just a shim to ensure API uniformity with other renderers.
+ * it renders an Excel report from html.
+ * @module lib/renderers/xls
+ * @requires juice
+ * @requires q
+ * @requires lodash
```
jniles: The `@requires lodash` isn't needed.

server/lib/renderers/xls.js
```diff
@@ -0,0 +1,30 @@
...
+exports.render = render;
+exports.extension = '.xls';
+exports.headers = headers;
+
+
+async function render(data, template, options) {
+  const htmlStream = await html.render(data, template, options);
+  const htmlString = juice(htmlStream);
+  const deferred = q.defer();
+  const buffer = Buffer.from(htmlString, 'utf8');
```
jniles: Same comment here about the need for `q`.

server/lib/template/partials/head.handlebars
```diff
@@ -1,23 +1,38 @@
 <!-- this partial requires binding the variable 'title' -->
 <head>
+  <meta charset="utf-8">
```
jniles, line 3: Good addition.  👍

jniles (commented):
Just a few suggestions here.  Looks pretty good!

server/lib/renderers/doc.js
jeremielodi: the ReportManager is waiting for a promise

server/lib/renderers/doc.js
jeremielodi: thanks

server/lib/renderers/doc.js
jniles: You are right.  According to [the MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Return_value), an asynchronous function returns:

>A Promise which will be resolved with the value returned by the async function, or rejected with an uncaught exception thrown from within the async function.

jniles:
Looks like CodeClimate is complaining that your yarn.lockfile is broken.  Did you merge it properly?  It might be easiest to simply delete it and run yarn again.

jeremielodi:
ok

jniles:
Looks good to me!

bors r+

jniles:
... looks like it might have been a fluke.

bors r+

jniles:
A second fluke!

bors r+

PR #3087, opened 2018-08-28
Basic unit tests for stock service

mbayopanda:
This PR add unit tests for some functions of the stock service and fix some export feature in stock registries.


client/src/js/services/StockService.js
```diff
@@ -241,7 +248,7 @@ function StockService(Api, Filters, AppCache, Periods, $httpParamSerializer, Lan
 
   /** Get label for purchase Status */
   function statusLabelMap(status) {
-    var keys = {
+    const keys = {
```
jniles: Since these are constants, can we move them outside of this function?  I feel like you structured your test similar to this code, but the code isn't well structured in this case.

client/src/modules/stock/lots/registry.html
```diff
@@ -35,12 +35,12 @@
             </li>
             <li role="separator" class="divider"></li>
             <li role="menuitem">
-              <a ng-href="/reports/stock/lots/?{{ StockLotsCtrl.download('lot', 'pdf') }}" download="{{ 'VOUCHER_REGISTRY.TITLE' | translate }}">
+              <a ng-href="/reports/stock/lots/?{{ StockLotsCtrl.getQueryString('lot', 'pdf') }}" download="{{ 'VOUCHER_REGISTRY.TITLE' | translate }}">
```
jniles: Why is this download called 'VOUCHER_REGISTRY.TITLE'?  I know if pre-existed your changes, but can you fix that?

test/client-unit/services/StockService.spec.js
```diff
@@ -0,0 +1,126 @@
...
+          {
+            uuid : 'd03e7870-0c8e-47d4-a7a8-a17a9924b124', lot : 'B', quantity : 30, expiration_date : '2018-12-01',
+          },
+        ],
+      },
+    ];
+
+    const expectedLotsArray = [
+      {
+        uuid : mockStockFormStoreData[0].lots[0].uuid,
```
jniles: An important part of tests is readability - and this isn't very readable to me.  I propose that we clean this up by creating some extra variables.  Like so:

```js
const lotA = {
  uuid : 'd03e7870-0c8e-47d4-a7a8-a17a9924b123',
  lot : 'A',
  quantity : 50,
  expiration_date : '2019-01-01',
};

const lotB = {
  uuid : 'd03e7870-0c8e-47d4-a7a8-a17a9924b124',
  lot : 'B',
  quantity : 30,
  expiration_date : '2018-12-01',
};
```
Now that you have those, you can do this:
```js
// a singe inventory with two lots
const mockStockFormStoreData = [{
  inventory_uuid : 'd03e7870-0c8e-47d4-a7a8-a17a9924b3f4',
  unit_cost : 0.5,
  lots : [ lotA, lotB ],
}];

const processed = Stock.processLotsFromStore(mockStockFormStoreData, entityUuid);
const [first] = processed;

const result = {
  uuid : 'd03e7870-0c8e-47d4-a7a8-a17a9924b123',
  lot : 'A',
  quantity : 50,
  expiration_date : '2019-01-01',
  inventory_uuid : 'd03e7870-0c8e-47d4-a7a8-a17a9924b3f4',
  unit_cost : 0.5,
  origin_uuid : entityUuid,
};

expect(first).to.deep.equal(result);
```

test/client-unit/services/StockService.spec.js
```diff
@@ -0,0 +1,126 @@
...
+
+    expect(uniformText).to.have.property('displayName', entityText.text);
+    expect(uniformName).to.have.property('displayName', entityName.name);
+    expect(uniformDisplayName).to.have.property('displayName', entityDisplayName.display_name);
+  });
+
+  it('#getQueryString() returns a query string with parameters for a request', () => {
+    const FILTER_KEY_MOVEMENT = 'movement';
+    const FILE_TYPE_CSV = 'csv';
+    const query = Stock.getQueryString(FILTER_KEY_MOVEMENT, FILE_TYPE_CSV);
```
jniles: It will be helpful for future programmers to have a comment telling us what string this should produce.  Something like:
```js
// returns the string "/stock/lots/?type=csv&limit=100"
const query = Stock.getQueryString(FILTER_KEY_MOVEMENT, FILE_TYPE_CSV);
```

test/client-unit/services/StockService.spec.js
```diff
@@ -0,0 +1,126 @@
...
+        fileTypeRendererExists = true;
+      }
+      if (params[i].indexOf('period=') > -1) {
+        periodParamExists = true;
+      }
+      if (params[i].indexOf('limit=') > -1) {
+        limitParamExists = true;
+      }
+    }
+    expect(fileTypeRendererExists).to.be.equal(true);
```
jniles: You can actually just use `chai` to do this - specifically the [include() method](http://www.chaijs.com/api/bdd/#method_include).  You would something like this work and save you from doing the `for` loop?

```js
expect(query).to.include(`renderer=${FILE_TYPE_CSV}`);
expect(query).to.include('period=');
expect(query).to.include('limit=');
```

test/client-unit/services/StockService.spec.js
```diff
@@ -0,0 +1,126 @@
+/* global inject, expect */
+
+describe('StockService', StockServiceTests);
```
jniles, line 3: This is a good first attempt!

I propose you make a few lots/inventory items in the [MockDataService](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/client-unit/mocks/data.js) (if they don't exist already) and use those in your tests.  That way, this code can focus on the test itself, and spend less time setting up the tests.

jniles (commented):
@mbayopanda, nice tests!  Your testing logic is pretty good, but there is a lot of set up which makes the test a bit hard to read.  I propose you set up all your data, either in the top of the test file or in a MockService, and then begin executing test logic.

It's also typically better to check against new objects/values instead of the original objects/values when you write a test.  Imagine the following function:

```js
const o = {
  id : 1,
  data : { hello : 'world' }
};

function getData(o) {
  return o.data;
}
```

Now imagine you wrote the following test:
```js
const exampleO = {
  id : 2,
  data : { food : 'sandwich' },
}

const result = getData(exampleO);
expect(results).to.equal(exampleO.data);
```

This test _looks_ okay, but in reality, it may not catch all the changes I could make to the object.  Imagine that I begin refactoring and change the function `getData` to be:
```js
function getData(o) {
  delete o.data;
  return undefined;
}
```

I'm pretty sure the tests above would still pass, even though the behavior is completely different!  That is because `delete` will remove `o.data`, which is equal to `undefined`.  I wouldn't need to update the tests at all, and that is a _bad thing_.

We want our tests to tell us when the API changes in important ways.  For this reason, it is better to repeat yourself and say:
```js
expect(result).to.equal({ food : 'sandwich' });
```
... which will yield the error `expected undefined to equal {food : 'sandwich' }`.  It immediately tells you that I made a mistake somewhere in my code instead of silently breaking the behavior of the module.

client/src/js/services/StockService.js
```diff
@@ -8,31 +8,40 @@ StockService.$inject = [
...
   // API for stock integration
-  var integration = new Api('/stock/integration');
+  const integration = new Api('/stock/integration');
 
   // API for stock transfer
-  var transfers = new Api('/stock/transfers');
+  const transfers = new Api('/stock/transfers');
+
+  // stock status label keys
+  const stockStatusLabelKeys = {
```
jniles, line 29: Nice 👍

jniles (approved):
Looks good to me!  Nice and clean.  

bors r+

Issue #3085, opened 2018-08-27
Tests for Filter Parser


lomamech:
Proposal to Implement Server-unit test for the library Filter

jniles:
See #2313 as well!

jniles:
Closed in #3088.

PR #3088, opened 2018-08-28
Test filter library

### LINKED ISSUE #2313: Write unit tests for FilterParser

### filter.js
The following test cases should be covered in the unit tests for [FilterParser](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/filter.js).
 - [ ] Test that FilterParser can construct the query `"SELECT 1 + 1 WHERE id = 1"`.
 - [ ] Test a query that uses `.fullText()` to add a "LIKE" statement to the query.
 - [ ] Test a query that uses `.dateFrom()` without a `.dateTo()` to format a query.
 - [ ] Test a query that uses `.dateTo()` without a `.dateFrom()` to format a query.
 - [ ] Test a query that uses both `.dateFrom()` and `.dateTo()` together to format a query.
 - [ ] Test a query that uses `limit`.
 - [ ] Test a query that uses `.custom()` to set a custom query text.
 - [ ] Test a query that uses `.custom()` with an array to set an "IN (...)" query.
 - [ ] Test a query that uses `.setOrder()` to set the ORDER BY parameter.
 - [ ] Test a query that uses `.setGroup()` to set the GROUP BY parameter.
 - [ ] Test a query that uses `.period()` to set the period id.
 - [ ] Test a query with `autoParseStatments` and make sure it works.
 - [ ] Test a query with a UUID to make sure it is automatically parsed.

lomamech:
- Implement Unit test fror Library Filter
- Update API for patient, Use filters.custom instead
  filters.customMultiParameters

improve(library Filter)  
- Remove filter customMultiParameters

Complete unit test for Filter Library to :
  - test a query that uses both dateFrom() and dateTo()
  - test a query with autoParseStatments
  - test a query with a Uuid is automatically parsed

closes #2313  #3085

### DIFF
--- a/server/lib/filter.js
+++ b/server/lib/filter.js
@@ -182,18 +182,6 @@ class FilterParser {
     }
   }
 
-  customMultiParameters(filterKey, preparedStatement, preparedValues) {
-    if (this._filters[filterKey]) {
-      const parameters = preparedValues || this._filters[filterKey];
-
-      // add the filters to the custom query, destructing parameters
-      this._statements.push(preparedStatement);
-      this._parameters.push(...parameters);
-
-      delete this._filters[filterKey];
-    }
-  }
-
   /**
    * @method setOrder
    *
--- a/test/server-unit/filter.spec.js
+++ b/test/server-unit/filter.spec.js
@@ -3,7 +3,7 @@ const _ = require('lodash');
 
 const filter = require('../../server/lib/filter');
 
-const objects = {
+const object1 = {
   id : 1,
   name : 'Unit Test',
   date_object : '1960-06-30',
@@ -15,9 +15,11 @@ const objects = {
   town : 'Kinshasa',
   street : 'Sgt Moke',
   location : 'Safricas / Sgt Moke/ Kinshasa / DR Congo',
+  dateFrom : '1960-06-30',
+  dateTo : '1997-05-17',
 };
 
-const accounts = {
+const object2 = {
   id : 1,
   label : 'PROFIT',
   number : 7001010,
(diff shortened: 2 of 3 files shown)


server/lib/filter.js
```diff
@@ -182,18 +182,6 @@ class FilterParser {
     }
   }
 
-  customMultiParameters(filterKey, preparedStatement, preparedValues) {
```
jniles, line 185: :dancer:

test/server-unit/filter.spec.js
```diff
@@ -0,0 +1,169 @@
...
+};
+
+describe('filter.js', () => {
+  
+  it('#fullText Format the sql query when filtered by full text', () => {
+    const filters = new filter(object1, { tableAlias : 't' });
+    filters.fullText('name');
+    const sql = `SELECT t.id, t.name, t.date_object, t.value, t.country_id FROM tables AS t`;
+
+    const formated = filters.applyQuery(sql).trim();
```
jniles: Just a small note - "formated" should be spelled "formatted" :)

test/server-unit/filter.spec.js
```diff
@@ -0,0 +1,169 @@
...
+
+const object4 = {
+  grade_uuid : 'c8a406a953d6429f84d8fc497875a580',
+  grade : 'Lieutenant-général',
+};
+
+describe('filter.js', () => {
+  
+  it('#fullText Format the sql query when filtered by full text', () => {
+    const filters = new filter(object1, { tableAlias : 't' });
```
jniles: To make these tests a bit simpler/cleaner, I propose you move all the `new filter(object1, { tableAlias : 't' })` calls into a `beforeEach()` method.  Like this:
```js

let firstFilters;
let firstSql;

beforeEach(() => {
  firstFilters = new Filter(object1, { tableAlias : 't' });
  firstSql = `SELECT t.id, t.name, t.date_object, t.value, t.country_id FROM tables AS t`;
});
```

This will clean up each block of code by focusing only on the test instead of the setup portion.  For example, you first test would then be:
```js
it('#fullText Format the sql query when filtered by full text', () => {

  // fullText matches the SQL "LIKE" filter.
  filters.fullText('name');

  const formatted = filters.applyQuery(sql).trim();
  const expected = `SELECT t.id, t.name, t.date_object, t.value, t.country_id FROM tables AS t WHERE LOWER(t.name) LIKE ?`;

  // assert that the SQL is formatted correctly.
  expect(formatted).to.equal(expected);
});
```

jniles (commented):
@lomamech, these are really good tests!

One important thing to add to these tests: you should also check that the `filters.parameters()` returns the correct values after a filter is applied.  You already check that the SQL statement is correct - let's just add something like:
```js
const params = filters.parameters();
expect(params).to.deep.equal([1, 2, 3]);
```

This is particularly important when we combine queries, like the `dateFrom()` and `dateTo()` filters ...  if the FilterParser returned them in the wrong order, we wouldn't get the right results!

jniles:
Looks good to me!  These are useful tests 👍 

bors r+

PR #3089, opened 2018-08-29
Update INSTALL.md

Scholldier:
I realized that there was a bit of a typo / difficult to read section around the MySQL strict no tables setup that should be fixed in the install documentation.


jniles:
Nice! 

bors r+

PR #3094, opened 2018-08-29
fix: migrate to utf8mb4

### LINKED ISSUE #2989: Remove collations from SQL queries
We have eight queries that manage character set collations themselves instead of using the database's globally defined collation.  This is making the upgrade process hard from MySQL's broken `utf8` to `utf8mb4`.   It would be ideal to remove these individual collation in favor of datbase, table, and column level collations. 

These queries probably ran into some issue with collations that resulted in errors.  The developer may have to debug these queries to see what the issue is.

Here they are, with line numbers:
managedcollations
_Fig 1: Queries with COLLATE_

(Found via `grep -nr "COLLATE"`)

jniles:
This commit migrates the database to UTF8MB4 in preparation for MySQL 8 which defaults to this.  It also produces better results for diverse character sets.

Closes #2989.

**Overview**
The motivation for this Pull Request is two-fold:
 1. Make sure no one had [break the server](https://medium.com/@adamhooper/in-mysql-never-use-utf8-use-utf8mb4-11761243e434) by putting valid characters that [cannot be represented in MySQL's UTF8 character set](https://stackoverflow.com/questions/30074492/what-is-the-difference-between-utf8mb4-and-utf8-charsets-in-mysql#30074553).
 2. Ensure we are set up for future releases of MySQL and future languages by migrating to UTF8MB4.  MySQL 8 will support this out of the box.

And additional benefit continues the work started by @mbayopanda in his work on Travis CI changes in #3061 to document and standardize the character sets and collations in our database.  Previously, we just relied on having "the right set up", with some of our Stored Procedures in the server's default encoding, functions called in the mysqljs's default encoding, etc...  Now, we define at creation time and connection time the character sets and collations we are working with.

**Notes**
I had to [skip one test](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/end-to-end/verificationLinks/verificationLinks.spec.js#L19).  It was the link between debtor groups and patients.  I'm pretty sure this is due to the fact that the order data was inserted in changed somehow.  Either way, the test is non-specific and should be re-written to reference the group by its name.

### DIFF
--- a/server/lib/db/index.js
+++ b/server/lib/db/index.js
@@ -3,13 +3,12 @@ const q = require('q');
 const mysql = require('mysql');
 const uuidParse = require('uuid-parse');
 const uuidv4 = require('uuid/v4');
-const Transaction = require('./transaction');
 const _ = require('lodash');
+const debug = require('debug')('db');
 
 const BadRequest = require('../errors/BadRequest');
 const NotFound = require('../errors/NotFound');
-
-const debug = require('debug')('db');
+const Transaction = require('./transaction');
 
 /**
  * @class DatabaseConnector
@@ -32,7 +31,11 @@ class DatabaseConnector {
       user     : process.env.DB_USER,
       password : process.env.DB_PASS,
       database : process.env.DB_NAME,
-      charset  : 'utf8_unicode_ci',
+
+      // NOTE(@jniles): the MySQL character set variable must be uppercase.  To
+      // see the full list of check out:
+      // https://github.com/mysqljs/mysql/blob/master/lib/protocol/constants/charsets.js
+      charset  : 'UTF8MB4_UNICODE_CI',
     };
 
     this.pool = mysql.createPool(params);
@@ -115,8 +118,8 @@ class DatabaseConnector {
   one(sql, params, id, entity = 'record') {
     return this.exec(sql, params)
       .then(rows => {
-        const errorMessage =
-          `Expected ${entity} to contain a single record with id ${id}, but ${rows.length} were found!`;
+        // eslint-disable-next-line max-len
+        const errorMessage = `Expected ${entity} to contain a single record with id ${id}, but ${rows.length} were found!`;
 
         if (rows.length < 1) {
           debug(`#one(): Found too few records!  Expected 1 but ${rows.length} found.`);
--- a/server/models/functions.sql
+++ b/server/models/functions.sql
@@ -134,7 +134,7 @@ BEGIN
     SET accountType = 'expense';
   END IF;
 
-  SET accountTypeId = (SELECT id FROM account_type WHERE `type` = accountType COLLATE utf8_unicode_ci LIMIT 1);
+  SET accountTypeId = (SELECT id FROM account_type WHERE `type` = accountType LIMIT 1);
 
   RETURN accountTypeId;
 END
--- a/server/models/procedures/account.sql
+++ b/server/models/procedures/account.sql
@@ -28,23 +28,23 @@ BEGIN
   DECLARE IMPORT_DEFAULT_OHADA_ACCOUNT_OPTION TINYINT(1) DEFAULT 1;
 
   SET existAccount = (SELECT IF((SELECT COUNT(`number`) AS total FROM `account` WHERE `number` = accountNumber) > 0, 1, 0));
-  SET existAccountType = (SELECT IF((SELECT COUNT(*) AS total FROM `account_type` WHERE `type` = accountType COLLATE utf8_unicode_ci) > 0, 1, 0));
-  SET accountTypeId = (SELECT id FROM `account_type` WHERE `type` = accountType COLLATE utf8_unicode_ci LIMIT 1);
+  SET existAccountType = (SELECT IF((SELECT COUNT(*) AS total FROM `account_type` WHERE `type` = accountType) > 0, 1, 0));
+  SET accountTypeId = (SELECT id FROM `account_type` WHERE `type` = accountType LIMIT 1);
   SET existAccountParent = (SELECT IF((SELECT COUNT(*) AS total FROM `account` WHERE `number` = accountParent) > 0, 1, 0));
 
   SET accountLength = (SELECT CHAR_LENGTH(accountNumber));
 
-  /* 
+  /*
     Handle parent account for importing ohada list of accounts
     We assume that ohada main accounts are already loaded into the system
   */
   IF (existAccountParent = 1) THEN
     SET accountParentId = (SELECT id FROM `account` WHERE `number` = accountParent);
   END IF;
-  
 
-  /* 
-    Create account if it doesn't exist 
+
+  /*
+    Create account if it doesn't exist
 
     if the account already exist skip because we are in a loop and
     we have to continue importing other accounts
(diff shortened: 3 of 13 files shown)


jniles:

jmcameron:
If a separate manual step is necessary to upgrade the database (apart from those done automatically by 'yarn dev'), please make a note of that somewhere so I'll know what to do.   Thanks.

-Jonathan C

jniles:
@jmcameron good point!  I'm not confident it works on anyone else's machine but my own (plus Travis), so I'm hoping @mbayopanda can help contribute those insights when he tests on Windows.  Once we are confident that it works with `yarn dev`, I'll jot down some notes to include with this PR.

server/lib/db/index.js
```diff
@@ -32,7 +31,7 @@ class DatabaseConnector {
       user     : process.env.DB_USER,
       password : process.env.DB_PASS,
       database : process.env.DB_NAME,
-      charset  : 'utf8_unicode_ci',
+      charset  : 'UTF8MB4_UNICODE_CI',
```
mbayopanda: 👍 it is important to comment that the charset `UTF8MB4_UNICODE_CI` must be in uppercase.

server/lib/db/index.js
```diff
@@ -115,8 +114,8 @@ class DatabaseConnector {
   one(sql, params, id, entity = 'record') {
     return this.exec(sql, params)
       .then(rows => {
-        const errorMessage =
-          `Expected ${entity} to contain a single record with id ${id}, but ${rows.length} were found!`;
+        // eslint-disable-next-line max-len
```
mbayopanda, line 121: 😄 is it allowed ?

server/models/functions.sql
```diff
@@ -134,7 +134,7 @@ BEGIN
     SET accountType = 'expense';
   END IF;
 
-  SET accountTypeId = (SELECT id FROM account_type WHERE `type` = accountType COLLATE utf8_unicode_ci LIMIT 1);
+  SET accountTypeId = (SELECT id FROM account_type WHERE `type` = accountType COLLATE utf8mb4_unicode_ci LIMIT 1);
```
mbayopanda: Is it possible to get rid also of all these `COLLATE utf8mb4_unicode_ci` ?

server/models/procedures/account.sql
```diff
@@ -28,31 +28,31 @@ BEGIN
   DECLARE IMPORT_DEFAULT_OHADA_ACCOUNT_OPTION TINYINT(1) DEFAULT 1;
 
   SET existAccount = (SELECT IF((SELECT COUNT(`number`) AS total FROM `account` WHERE `number` = accountNumber) > 0, 1, 0));
-  SET existAccountType = (SELECT IF((SELECT COUNT(*) AS total FROM `account_type` WHERE `type` = accountType COLLATE utf8_unicode_ci) > 0, 1, 0));
-  SET accountTypeId = (SELECT id FROM `account_type` WHERE `type` = accountType COLLATE utf8_unicode_ci LIMIT 1);
+  SET existAccountType = (SELECT IF((SELECT COUNT(*) AS total FROM `account_type` WHERE `type` = accountType COLLATE utf8mb4_unicode_ci) > 0, 1, 0));
+  SET accountTypeId = (SELECT id FROM `account_type` WHERE `type` = accountType COLLATE utf8mb4_unicode_ci LIMIT 1);
```
mbayopanda: If we can remove all these `COLLATE utf8mb4_unicode_ci` it will be very nice

sh/build-database.sh
```diff
@@ -16,38 +16,38 @@ set +a
 TIMEOUT=${BUILD_TIMEOUT:-8}
 
 # build the test database
-mysql -u $DB_USER -p$DB_PASS -e "DROP DATABASE IF EXISTS $DB_NAME ;" &> /dev/null
```
mbayopanda: Why don't we redirect the output to `/dev/null` now ?

test/end-to-end/verificationLinks/verificationLinks.spec.js
```diff
@@ -16,7 +16,7 @@ describe('Check Inter-Registry Links', () => {
     GU.expectRowCount('patient-registry', 2);
   });
 
-  it('Checks the link between Debtor Groups -> Patient Registry', () => {
+  it.skip('Checks the link between Debtor Groups -> Patient Registry', () => {
```
mbayopanda: Could you leave a comment about the reason of skipping this test ?

mbayopanda (approved):
LGTM 👍 

I leave some comments which cannot block the merge this PR

server/lib/db/index.js
jniles: True.  Will do.

server/models/functions.sql
jniles: Hmm... good question.  Let me try!

server/lib/db/index.js
jniles: Eh, it's either that or figure our a shorter representation ;)

sh/build-database.sh
jniles: Ah - this was a feature requested by @jmcameron.  We basically eliminate all ability to debug by sending all errors into `/dev/null/`.  I figured this was a bit too much.  Do you think we should put it back?

jmcameron: Can we set up the script to have an argument that determines whether the output goes to /dev/null or to regular output?

jmcameron: If you want, I can take a cut at setting up the script with an optional argument for debugging:

  sh/build-database.sh debug

would force output to output to go to stdout (if I can figure out how to do that...)

-Jonathan C

jmcameron: Got a test bash script working for the debug option here.

jmcameron: [testo.txt](https://github.com/IMA-WorldHealth/bhima-2.X/files/2336072/testo.txt)

Download, remove .txt and chmod +x and try it with:
   ./testo
   ./testo debug

jniles: Hey, that is kind of nice!  I'll incorporate it. 👍

jniles:
@mbayopanda, this is ready for a second review.

jniles:
@sfount want to weigh in here?  Did you test it locally?

sfount:
@jniles I've tested this branch locally and it passes all tests. Running the most up to date version of MySQL 5.6 on linux.

jniles:
@sfount, thanks for you feedback.  Glad this didn't break anything!

jniles:
Alright, merging!

bors r+

jniles:
@jmcameron - to answer your earlier question: it's a long an laborious process and depends a lot on the database.  [This SO answer is a good overview](https://dba.stackexchange.com/questions/8239/how-to-easily-convert-utf8-tables-to-utf8mb4-in-mysql-5-5#21684) of the basic procedure, to be adapted to whatever databases are needed.  Additionally, the Stored Procedures and Functions may have to be recreated in a connection whose CHARSET is utf8mb4 for the parameters to be correct.

For our production sites, we'll need to write a custom script to migrate their datasets.  Since we don't have a standard way of pushing updates, we'll need to write scripts for each site's database in particular.

EDIT:
[Percona](https://www.percona.com/blog/2018/04/10/migrating-database-charsets-to-utf8mb4/) has a great article on migrating existing databases in a sane fashion.  Depends on our hardware though.

PR #3097, opened 2018-08-29
Fix bh-currency-select unhandled error logging 

### LINKED ISSUE #3096: bh-currency-select throws an unhandled error on initialise (low priority)
During testing of the cash module in #3095 it was noted that the `bh-currency-select` component always throws an undhandled error. 

```
TypeError: Cannot read property 'currency' of undefined
```

This issue has never been reported as a bug so it doesn't look like it has an impact on the end users experience however it comes up frequently as this component is used in multiple places. It would be good to clean this up so that future changes can be built with a clean log. 

The validation method on the component references a `form` attribute on the component that does not exist, it looks like this was potentially introduced as a work around to a previous version of angular's `ng-form` 

As I understand the code there are two possible solutions: 
1. Remove the code directly, I will follow up with tests but I believe the component works without directly setting the $valid property of the underlying form elements. 
- [x] Test the component with different inputs to ensure it works without the legacy code 
2. The sub form elements *can* be exposed to the controllers scope by using an `ng-form` wrapper around the currency select inputs. This would expose the `$setValidity` method that the current controller expects to exist. for example: 
```html 
<!-- Attach 'form' to the components scope -->
<div ng-form="$ctrl.form">
   <!-- Currency select HTML inputs -->
</div>
```

sfount:
This commit removes a method forcing validation on form elements that
are updated by changes propegated in `ng-form` and closes #3096. 

It also updates the relevant controller code to repository standards.

### DIFF
--- a/client/src/js/components/bhCurrencySelect.js
+++ b/client/src/js/components/bhCurrencySelect.js
@@ -119,6 +119,5 @@ function bhCurrencySelect(Currencies) {
     // if the two array lengths are equal, it means every currency is disabled.
     // there is no possible $valid state.
     $ctrl.valid = ($ctrl.currencies.length !== disabledIds.length);
-    $ctrl.form.currency.$setValidity('currency', $ctrl.valid);
   }
 }
--- a/client/src/js/components/bhCurrencySelect.js
+++ b/client/src/js/components/bhCurrencySelect.js
@@ -63,15 +63,15 @@ bhCurrencySelect.$inject = ['CurrencyService'];
  * @requires services/CurrencyService
  */
 function bhCurrencySelect(Currencies) {
-  var $ctrl = this;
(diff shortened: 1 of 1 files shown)


client/src/js/components/bhCurrencySelect.js
```diff
@@ -119,6 +119,5 @@ function bhCurrencySelect(Currencies) {
     // if the two array lengths are equal, it means every currency is disabled.
     // there is no possible $valid state.
     $ctrl.valid = ($ctrl.currencies.length !== disabledIds.length);
-    $ctrl.form.currency.$setValidity('currency', $ctrl.valid);
```
jniles, line 122: Good catch!

jniles (approved):
Looks good to me.  Nice clean up of legacy code as well.

bors r+

PR #3098, opened 2018-08-29
Fix Journal Search: Full Transactions (3055)

### LINKED ISSUE #3055: "Show Full Transactions" option in Journal Search Modal is broken
It looks like the "Show Full Transactions" option no longer works in the Posting Journal.  Changes are not reflected on the client, nor are they cached in AppCache.

Steps to Reproduce:
 1. Go to the Journal
 2. Click "Search"
 3. Click "Default Filters" in the tabs
 4. Change the "Show Full Transactions" option
 5. Submit

The value should not have changed in the filters above.

sfount:
This commit changes the value that it is passed into the `onChange`
callback used with the `bh-yes-no` component for the 'Show Full
Transactions' option with the Posting Journal.

Previously the value 0 was always being set which meant the setting
would never reflect a change.

The `default-value` flag was removed from the component as it is
meaningless and not supported, however the current behaviour in
`journal.service.js` is to default `showFullTransactions` to 0 if it has not
yet been defined. A final decision should be made as to the default non-set
value for this flag.

It also includes a test case to catch this case with further repository changes. 

Closes #3055

### DIFF
--- a/client/src/modules/journal/modals/search.modal.html
+++ b/client/src/modules/journal/modals/search.modal.html
@@ -160,10 +160,12 @@
 
           <!-- Full Transaction applies to the Journal, not the Account Statement -->
           <div ng-if="!ModalCtrl.hasDefaultAccount">
+
+            <!-- Given this components architecture, `value` on onChangeCallback -->
+            <!-- refers to the ng-value of the currently selcted radio button -->
             <bh-yes-no-radios
-              default-value="true"
               value ="ModalCtrl.defaultQueries.showFullTransactions"
-              on-change-callback="ModalCtrl.toggleFullTransaction(ModalCtrl.defaultQueries.showFullTransactions)"
+              on-change-callback="ModalCtrl.toggleFullTransaction(value)"
               label="POSTING_JOURNAL.SHOW_FULL_TRANSACTION_RECORDS"
               help-text="POSTING_JOURNAL.SHOW_FULL_TRANSACTION_RECORDS_HELP">
             </bh-yes-no-radios>
--- a/client/src/modules/journal/modals/search.modal.html
+++ b/client/src/modules/journal/modals/search.modal.html
@@ -14,7 +14,7 @@
   <div class="modal-body" style="padding : 0px;">
 
     <uib-tabset>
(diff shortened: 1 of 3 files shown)


jniles:
Hey @sfount - great to have you back!  I'll give this a whirl back at work, but looks good on a quick glance. 👍

client/src/modules/journal/modals/search.modal.html
```diff
@@ -160,10 +160,12 @@
...
           <!-- Full Transaction applies to the Journal, not the Account Statement -->
           <div ng-if="!ModalCtrl.hasDefaultAccount">
+
+            <!-- Given this components architecture, `value` on onChangeCallback -->
+            <!-- refers to the ng-value of the currently selcted radio button -->
             <bh-yes-no-radios
-              default-value="true"
               value ="ModalCtrl.defaultQueries.showFullTransactions"
-              on-change-callback="ModalCtrl.toggleFullTransaction(ModalCtrl.defaultQueries.showFullTransactions)"
+              on-change-callback="ModalCtrl.toggleFullTransaction(value)"
```
jniles, line 168: Good catch 👍

test/end-to-end/journal/journal.page.js
```diff
@@ -55,15 +58,41 @@ function JournalCorePage() {
...
+    // reset custom filters
+    filters.resetFilters();
+
+    // reset default filter negative value
+    showFullTransactions(false);
+  }
+
+  // @TODO(sfount) A significantly better selector should be used here but
+  //               the model on the component is `$ctrl.value`. Something
+  //               less ambiguos should be used.
```
jniles, line 73: Yeah, that's a good point.  Although, since this is just a DOM selector, we could probably stick an `id` onto the bh-yes-no-radios and then subselect the containing HTML with it.  Like this:
```html
<bh-yes-no-radios
  id="showFullTransactionsToggle"
 ...>
```
```js
const radios = $('#showFullTransactionsToggle');
radios.$(`[data-choice="${selection}"]`).click();
```

jniles (approved):
Looks and behaves as described.  Great bug fix, complete with a test!

bors r+

PR #3101, opened 2018-08-30
Updates to employee selection component view

jmcameron:
Updated the employee selection component view like other recent updates.   Tested it in Reports/Employee Standing and it seems to work.

@jniles I have two questions/uncertainties about my changes:

1. I changed **EmployeeForm.employee_id.$invalid** to **EmployeeForm.employee_uuid.$invalid**.  It appears to me that the name of the field in the form is **employee_uuid**.  Not sure if this is a misunderstanding on my part or an unknown bug due to renaming the field.   Let me know if I need to revert this.  I did verify that the error message is invoked it the form is submitted without choosing an employee so I think my fix is correct.

2. In the original version of bhPeriodSelection.html, there were two instances of the bh-employee-select tag (in the outer and inner html scopes).  It did not seem necessary to me to have both so I removed it from the inner scope.   If that breaks something (eg, in testing), I can put it back.

-Jonathan C


jniles:
Hey @jmcameron , thanks for the notes!

 1. Your changes are exactly right.  We migrated from `employee_id` to `employee_uuid` a while ago and this apparently was forgotten.  Thanks for catching that bug!
 2. We'll have to let the tests complete to answer the second one.  I can't recall of the top of my head.

The changes look pretty good though.  I'll take a look once we have test results in.

client/src/modules/reports/generate/employeeStanding/employeeStanding.js
```diff
@@ -42,7 +42,10 @@ function EmployeeStandingController($state, $sce, Notify, BaseReportService, App
   };
 
   vm.preview = function preview(form) {
-    if (form.$invalid) { return 0; }
+    if (form.$invalid) {
+      Notify.danger('FORM.ERRORS.RECORD_ERROR');
```
jniles, line 46: Awesome!

client/src/modules/templates/bhEmployeeSelect.tmpl.html
```diff
@@ -19,8 +24,8 @@
       </ui-select-choices>
     </ui-select>
 
-    <div class="help-block" ng-messages="EmployeeForm.employee_uuid.$error" ng-show="$ctrl.validateTrigger">
+    <div class="help-block" ng-messages="EmployeeForm.employee_uuid.$error" ng-show="EmployeeForm.$submitted">
```
jniles, line 27: 💃

jniles (approved):
Looks good to me!  Thanks!

bors r+

PR #3104, opened 2018-08-30
fix fiscal year list pagination

mbayopanda:
This PR fix the pagination in the fiscal year list module by using the uib-pagination attribute instead of the element.


jniles (approved):
Works as expected.  Glad you got this working!

PR #3106, opened 2018-08-30
Order Transaction Types by 'fixed' and then 'text' (3059)

### LINKED ISSUE #3059: Transaction Type should be ordered by 'fixed', then by "text"
Another client-side interface that could benefit from sorting is the Transaction Type module.  However, we always want the "fixed" values to appear on top.  We should first order by "fixed", then by the label of the transaction type, in order to preserve this hierarchy, but sort alphabetical order within the groups.

sfount:
This commit orders the Transaction Type Management `ui-grid`. It first
orders by the 'fixed' property (if the transaction type is part of the
system of a custom user type) and then alphabetically.

It also updates the `ui-grid` according to a number of other similar UI patterns found in the application.

Addresses a user facing bug resulting in the Transaction Type 'types' rendering hard-coded database values. 

**Current master:**
screenshot from 2018-08-30 17-47-08

Note: French translation with English hard-coded values 

**Updated branch:**
screenshot from 2018-08-30 17-35-57

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/3059

### DIFF
--- a/client/src/js/constants/bhConstants.js
+++ b/client/src/js/constants/bhConstants.js
@@ -144,7 +144,7 @@ function constantConfig() {
       { id : 5, label : 'FORM.LABELS.WEEK_DAYS.FRIDAY' },
       { id : 6, label : 'FORM.LABELS.WEEK_DAYS.SATURDAY' },
     ],
-    transactionTypesMap : {
+    transactionTypeMap : {
       income : { label : 'VOUCHERS.SIMPLE.INCOME', value : 'income' },
       expense : { label : 'VOUCHERS.SIMPLE.EXPENSE', value : 'expense' },
       other : { label : 'FORM.LABELS.OTHER', value : 'other' },
--- a/client/src/js/constants/bhConstants.js
+++ b/client/src/js/constants/bhConstants.js
@@ -6,8 +6,8 @@ angular.module('bhima.constants')
  * populated according to the enterprise configuration.
  */
 function constantConfig() {
-  var UTIL_BAR_HEIGHT = '106px';
-  var JOURNAL_UTIL_HEIGHT = '150px';
+  const UTIL_BAR_HEIGHT = '106px';
+  const JOURNAL_UTIL_HEIGHT = '150px';
 
   return {
     actions : {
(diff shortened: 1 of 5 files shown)


client/src/js/constants/bhConstants.js
```diff
@@ -134,6 +144,10 @@ function constantConfig() {
       { id : 5, label : 'FORM.LABELS.WEEK_DAYS.FRIDAY' },
       { id : 6, label : 'FORM.LABELS.WEEK_DAYS.SATURDAY' },
     ],
+    transactionTypesMap : {
+      income : { label : 'VOUCHERS.SIMPLE.INCOME', value : 'income' },
```
jniles: It probably is better English to call this "transactionTypeMap".

client/src/modules/transaction-type/transaction-type.ctrl.js
```diff
@@ -96,25 +84,32 @@ function TransactionTypeController(TransactionType, TransactionTypeStore, Notify
...
       .catch(Notify.handleError);
   }
 
+  // Assign translatable labels to each transaction type based on the hardcoded
+  // database strings
+  function assignTransactionTypeLabels(transactionType) {
+    // @TODO(sfount) The `transaction_type` database currently hard codes 'income',
+    //               'expense' and 'other' transaction types. When this is updated
+    //               with a more data driven approach it should include translatable
+    //               labels
```
jniles, line 115: Good comment!  Mind making an issue about it?  I can't recall if one exists or not..

jniles (commented):
@sfount great work!  I hadn't realized that the transaction types weren't respecting i18n.

Could we get one more level of sort on here?  It would be nice to sort the left-most values (`field: text`) in alphabetical order.   You'll probably need to translate the values programmatically and then sort them.  That way, it is easy to locate a particular transaction type in the grid.

sfount:
@jniles What do you know, the screenshot shows it's not sorted... At least that wasn't the explicit reason for the pull request! 

I'll have a look at this now. It is sorting but it looks like it's sorting on the underlying translate key.

sfount:
@jniles Thanks for the comments! 

Now with _actual_ sorting: 
ordered

jniles (approved):

bors r+

jniles:
hmm... import accounts seems to be failing ~25% of the time.   I'll look into this.

bors r+

jniles:
Yay! 💃

Issue #3108, opened 2018-08-30
Proposal: Add Integer column to journal tables to improve Transaction ID generation performance


sfount:
This issue follows on from https://github.com/IMA-WorldHealth/bhima-2.X/issues/3002 and the following discussion https://github.com/IMA-WorldHealth/bhima-2.X/issues/3002#issuecomment-417288766. 

The proposed solution to the `GenerateTransactionId()` SQL function performance issues is to add the column `trans_id_reference_integer` to both the `posting_journal` and the `general_ledger`. Having this indexed and sortable Integer will allow GenerateTransactionId to move from **O(n)** complexity to **O(1)**, as shown in https://github.com/IMA-WorldHealth/bhima-2.X/issues/3002#issuecomment-417288766.

### Requirements: 
-  [x] Database schema changed to add an additional column and index to the `posting_journal` and `general_ledger`
- [x]  Upgrade script for current databases to add the new column and index, as well as populate the new column
- [x] Update all controllers writing to the posting_journal and general_ledger to also write the integer (reference) part of the transaction ID
- [x] Update the GenerateTransactionID function to take advantage of the new indexed column
- [x] Measure performance improvement 
- [x] Thoroughly test before merging with production code 

### Pros: 
* Order of magnitude performance improvement (~200% during prototype testing) 
* Reduce time required to create all financial entities 

### Cons: 
* Increased database storage requirements with an additional indexed column

sfount:
Performance and tests were reviewed in https://github.com/IMA-WorldHealth/bhima-2.X/pull/3110. This may require further evaluation regarding table size/ performance however https://github.com/IMA-WorldHealth/bhima-2.X/pull/3110 closes the requirements of this issue.

Issue #3002, opened 2018-08-06
Improve GenerateTransactionId() function performance


jniles:
@lomamech points out that our SQL function [`GenerateTransactionId()`](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/functions.sql#L73) function is very inefficient.  We could improve this easily enough by rewriting it to store the latest transaction ID for each project in a small table.  The table would look like this:

| **project_id** | **transaction_id** |
|:--:|:--:|
| 1 | 254 |
| 2 | 576 |

The logic is that that every time a transaction ID is successfully used/generated, the `transaction_id` column is updated.  This means that instead of a full scan of the `posting_journal` or `general_ledger`, it will just pluck the latest version from the table.

As much as possible, this call should not change the signature of the `GenerateTransactionId()` function.

jmcameron:
Note that this proposal has the potential of being brittle since the table may be updated and used in different places, possibly by multiple sections of code.  Therefore it will be extra important to have good testing of this feature to make sure the table us updated EVERY time a transaction ID is generated or used anywhere in the code.

sfount:
@lomamech @jniles @jmcameron improving something that is called so frequently will make a huge difference on low end hardware. 

**In brief:** With an indexed integer representation of the transaction ID we could improve this query by ~200% (0.157624s vs. 0.000883s as briefly tested)

I reviewed the query this morning  (`SHOW PROFILE;`)  and it looks like the element of the profile that is most time consuming is interestingly **'Sending data'**. My understanding here is that MySQL has to actually load (very temporarily) every single `trans_id` into memory to then perform the `SUBSTR`, `FLOOR`, and `MAX`. This could easily be 200,000+ strings.

The inefficiency is most likely in transporting these strings around rather than any joins or index issues. 

One solution that addresses the multiple sections of code consistency pointed out by @jmcameron and the performance would be to store the integer part of the ID in both the `posting_journal` and the `general_ledger` as an indexed column. This would allow us to write this query with a more succinct limit order.

```SQL 
ORDER BY trans_id_integer LIMIT 1;
```
In theory this should improve performance but guard against allowing non-transactional updates from different areas of the code. 

A quick analysis on ~200,000 records and averaged across multiple queries:

Original query (copied from repo):

| Roundtrip Query Duration | Query Profile Duration | Sending Data Duration  |
| ------------- | ------------- | ----- |
| 0.516s    | 0.157624s | 0.154132 |

[...]

PR #3110, opened 2018-08-31
Performance Refactor on GenerateTransactionID SQL (3108)

sfount:
This PR represents the initial implementation of the performance improvement proposal https://github.com/IMA-WorldHealth/bhima-2.X/issues/3108 as discussed in https://github.com/IMA-WorldHealth/bhima-2.X/issues/3002. 

The referenced issues do a good job of explaining the details and analysis, in brief this PR: 
* Updates the GenerateTransactionID SQL function for improved performance 
* Implements an upgrade script to consider current production systems 
* Updates server controllers responsible for populating the additional data column required

I am not 100% on the latest SQL migration standards, what is included is trying to replicate the standard I have seen. 

**Please note: This pull request changes a fundamental method that is relied on by every financial entity in the system, it should be thoroughly vetted before integrating with the production code base.**


sfount:
Looks like end to end tests are failing consistently, they passed on the last local check. I'll look into this now.

jniles:
Yeah, we have some spotty tests.  Same with the BUID() pull request.  My next push should be on sorting those out.

server/models/functions.sql
```diff
@@ -66,28 +66,43 @@ END $$
...
 CREATE FUNCTION GenerateTransactionId(
-  project_id SMALLINT(5)
+  target_project_id SMALLINT(5)
 )
 RETURNS VARCHAR(100) DETERMINISTIC
 BEGIN
   DECLARE trans_id_length TINYINT(1) DEFAULT 4;
   RETURN (
-    SELECT CONCAT(abbr, IFNULL(MAX(increment), 1)) AS id
+    SELECT CONCAT(
```
jniles: Great rewrite - this is clear than the previous version as well!

I think you can remove `trans_id_length`, since it has no more bearing on this function.  Note that this will need to be done in the migration script as well for completeness.

client/src/modules/journal/journal.service.js
```diff
@@ -74,6 +74,8 @@ function JournalService(Api, AppCache, Filters, Periods, Modal, bhConstants, Tra
       .then(service.util.unwrapHttpResponse);
   }
 
+  // @TODO(sfount) new rows will need to access the transaction shared attribute
```
jniles, line 77: Why will new rows need access to this attribute?

jniles (commented):
Looks good to me!  One quick thing to clean up and then I'm satisfied with it.

A future optimization we might do is to remove the INDEX on trans_id, and instead parse the incoming transaction id, then match on trans_id_reference_number, then match on the result of that.  It would probably be faster since it would be looking at an integer and save use a heap of space by avoiding the trans_id index.

jniles:
I can reproduce this locally.  Looking into the test error.

jniles:
So ... this PR is revealing a lot of cool errors.  For example, in our [test dataset](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/data.sql#L727), we have a cash payment marked "is_caution = 1" that has a linked invoice.  This doesn't make any sense.

I'm not sure why this PR is surfacing those errors, but it's nice to try and clean them up.  This may take a while to trace down.

jniles:
Okay, finally got these tests working.

Let's see if bors is happy.

bors try

jniles:
bors try

jniles:
Looks like this seems to work well.

@sfount, I'm going to try and get this running on an ARM board this evening, do you have any thoughts on the code changes I've made?  I'll post the results here for performance analysis.

sfount:
@jniles your code changes look good to me, this represents some really great research :+1: . 

Because this is so fundamental to generating transactions I would like to do one last test to make sure there are no discrepancies that we have missed comparing this branch and master but that shouldn't involve any code changes or interrupt your performance analysis. 

Great work! I'll update this thread when I've double checked for untested corner cases.

jniles:
Some benchmarks on the function `time mysql imck -e "SELECT GenerateTransactionId(1)"`


  | Run 1 | Run 2 | Run 3 | Run 4 | Run 5 | Average
-- | -- | -- | -- | -- | -- | --
before | 4.665 | 5.564 | 4.738 | 5.085 | 4.872 | 4.9848
after | 0.236 | 0.214 | 0.228 | 0.211 | 0.223 | 0.2224

It looks like we got a 95% performance increase in this one function.  Great work @sfount !

_Note: all measurements are in seconds_

PR #3111, opened 2018-08-31
Perf: implement HttpCache service to cache Accounts.read() call

### LINKED ISSUE #2892: Accounts seems to be hit a lot!
It looks like our HTTP requests seems to be hitting the `/accounts/` route a lot!  We should see if there is any way to reduce these requests - from my naive guess, the following screenshot was taken from the stock module.

accountshitalot

jniles:
**Original Text**
~~This PR introduces a new factory function for debouncing function calls.  The suggested ng-debounce library did not use promises and was cumbersome to use, thus I have implemented my own based on a SO issue.  The `Accounts.read()` method has been debounced by 2000 milliseconds to allow for things like `bhAccountSelects` to benefit from the same HTTP request instead of making one each.~~

**Updated Information**
This Pull request implements a new service called `HttpCache` that should transparently wrap any HTTP requests and ensure that the same function call will return the same result.  This is useful for smaller, more frequently called routes.

_Basic Usage:_
```js
// a GET HTTP request to some url
function read(id, options) {
  return $http.get(/*...*/);
}

// to proxy, simply wrap the function in HttpCache
const proxy = HttpCache((id, options) => read(id, options));

// You can now call proxy as
proxy(1, { limit : 10 })
  .then(() => /*... */);
```

For uncached requests, the behavior is the same a raw HTTP request.  For those that have been called multiple times, the web page no longer needs to send an extra HTTP request.

If you want to skip the cache (and bust it), simply pass `true` as the third parameter.

Closes #2892.


jniles:
So, on further analysis, this isn't going to work as I want it to.  The main reason is the `debounce()` doesn't take into account argument parameters.  So, `accounts.read(1)` and `accounts.read(2)` will return the same debounced result.

What we really need is a cross between [memoization](https://en.wikipedia.org/wiki/Memoization) and debounce.  As mentioned by @sfount (https://github.com/IMA-WorldHealth/bhima-2.X/issues/2892), there was an effort to use stores on the client for this kind of thing.  However, with all of our complex server-side filters, we'll either need to transition to filtering on the client in stores or figure out another option.

One potential hack would be to serialize the parameters passed on and use those as keys to an object caching the result of the function call with those parameters, similar to how memoization works.  However, these cache's would expire after a period of time (a number of seconds).  I'll work on this and see if I can get decent results.

client/src/js/services/HttpCacheService.js
```diff
@@ -0,0 +1,75 @@
...
+ * The HttpCacheService is a generic wrapper for asynchronous requests to ensure
+ * they requests with identical parameters are only called once. It works by
+ * serializing the request's parameters and caching the response to those parameters
+ * for a short time.  If any other requests are made during the cached time, the
+ * previous response is returned directly.
+ *
+ * NOTE(@jniles) - we use $interval here to avoid slowing down protractor tests.
+ * Thanks to @sfount who debugged this in Dec 2016.
+ */
+function HttpCacheService($interval) {
```
sfount, line 19: This is a super clean implementation! Really great comments and use of ES6 :ok_hand:.

client/src/modules/accounts/accounts.service.js
```diff
@@ -52,8 +58,7 @@ function AccountService(Api, bhConstants) {
    *   an array of JSONs.
    */
   function read(id, options) {
-    return Api.read.call(this, id, options)
-      .then(handleAccounts);
+    return fetcher(id, options);
```
sfount: It took me a minute to follow this logic because all of our other `read` methods follow the pattern: 
```
return fetchSomeData().then(processSomeData);
```

With this I had to follow the callback which then involved a good understanding of how `Function.prototype.call` worked. 

Could you move the `.then(handleAccounts)` out of the callback and into the `read` method? Then the callback and HttpCache are only responsible for loading accounts and the read method handles reading and pipes into processing. 

For example: 
```js
return fetcher(id, options)
  .then(handleAccounts);
```

sfount (commented):
Performance kicks :soccer:, this is awesome. 
 
I haven't had time to test this but should the HttpCacheService provide an option to bust the cache if required? I may be mistaken in this logic but because angularjs service follow the [singleton pattern](https://docs.angularjs.org/guide/services), the same instance of HttpCache will be used by any module using the AccountsService. 

As we are serialising the `...args` could it possibly return the same list of accounts even if a module has updated one of them? 

For example (I know this is obscure): 
* Accounts management pages loads all accounts
* Accounts management page updates account_1 to have text 'this is new text'
* A report module loads all accounts, 10 seconds later. As I understand it account_1 still have text 'this is old test'

client/src/modules/accounts/accounts.service.js
jniles: Yeah, good call.  I'll do that.

jniles:
Hmmm... I wonder what the best way is implement the cachebust.  We could do it by either: 
1. Putting a third parameter OR
2. Putting a reserved word on the `options` object as the second parameter.

I'm going explore the first implementation since I think it makes more sense from a user perspective, but it may result in some ... inelegant code.

jniles:
Hey, it seemed like that worked!

@sfount, ready for another review.

sfount (approved):
This is a good clean implementation of the feature described. 

All of the tests pass so it doesn't seem to break any functionality but I can think of a few corner cases we may have to address in the future. If this cache wrapper works well going forward we could look to add to other services that could benefit from similar performance improvements.

PR #3113, opened 2018-08-31
refactor(Trial Balance Procedures)

### LINKED ISSUE #3112: refactor Trial Balance
Refactor trial balance, accepts the balance of a transaction up to two ranks after the decimal point

lomamech:
- Update procedure for Trial Balance by adding the validation of the
  balance of a transaction up to two ranks after the decimal point

closes #3112


jniles (approved):
Looks good to me.  On the long run, we will probably want to make "2" into a variable to ensure that we can change it easily.

bors r+

Issue #3043, opened 2018-08-16
Changing roles does not reset the user's permissions in real time


jniles:
When a user updates permissions for their own role, nothing happens.  Previously, when we had a mapping of user->unit, it would refresh the list of units in the tree and reset their session on the server.  This made it so they could begin to use the application right away without needing to log out and log back in again.  It would be great to have this functionality again.

This was the original issue from before we had roles:  https://github.com/IMA-WorldHealth/bhima-2.X/issues/1819

jeremielodi:
This issue seems to be very easy to be solved but I'm afraid about the end to end test, the user will be disconnected and the test will restart or some thing like that

jniles:
It shouldn't have any effect - we did the same thing for permissions previously without any issues.

jeremielodi:
I see.

PR #3115, opened 2018-09-01
Reload user session on successful role permissions update (3043)

sfount:
This PR follows the discussion in https://github.com/IMA-WorldHealth/bhima-2.X/issues/3043.

Specifically it uses `SessionService.reload` to ensure the users current session information is up to date following a successful role permissions update. 

It also refactors client code for updating role permissions to improve legibility.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/3043

### DIFF
--- a/client/src/modules/roles/roles.ctrl.js
+++ b/client/src/modules/roles/roles.ctrl.js
@@ -6,12 +6,12 @@ RolesController.$inject = [
   'NotifyService', 'bhConstants',
 ];
 
-function RolesController($uibModal, Roles, session, Modal, Notify, bhConstants) {
+function RolesController($uibModal, Roles, Session, Modal, Notify, bhConstants) {
   const vm = this;
 
   vm.canEditRoles = false;
 
-  vm.add = (role = { project_id : session.project.id }) => {
+  vm.add = (role = { project_id : Session.project.id }) => {
     $uibModal.open({
       templateUrl : 'modules/roles/create.html',
       controller : 'RolesAddController as RolesAddCtrl',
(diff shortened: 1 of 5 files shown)


client/src/modules/roles/roles.ctrl.js
```diff
@@ -6,34 +6,44 @@ RolesController.$inject = [
   'NotifyService', 'bhConstants',
 ];
 
-function RolesController($uibModal, Roles, session, Modal, Notify, bhConstants) {
+function RolesController($uibModal, Roles, Session, Modal, Notify, bhConstants) {
```
jniles, line 9: Nice :+1:

client/src/modules/roles/roles.ctrl.js
```diff
@@ -6,34 +6,44 @@ RolesController.$inject = [
...
-      resolve : { data : () => role },
-    });
+      resolve : { data : () => selectedRole },
+    }).result
+      .then(() => {
+        // refresh the application session to ensure the latest versions of roles
+        // and permissions are applied, this will only run on submission
+        // @TODO(sfount) if the session information kept track of the current users
+        //               role, this method could only update if the current role is changed
+        Session.reload();
```
jniles, line 45: This is perfect 👍

client/src/modules/roles/roles.ctrl.js
```diff
@@ -85,7 +95,7 @@ function RolesController($uibModal, Roles, session, Modal, Notify, bhConstants)
     width : 100,
     displayName : '',
     headerCellFilter : 'translate',
-    cellTemplate : 'modules/roles/templates/action.tmpl.html',
+    cellTemplate : 'modules/roles/templates/action.cell.html',
```
jniles, line 98: 👍

client/src/modules/roles/roles.html
```diff
@@ -7,7 +7,7 @@
 
     <div class="toolbar" ng-show="RolesCtrl.canEditRoles">
       <div class="toolbar-item">
-        <button class="btn btn-default" ng-click="RolesCtrl.add()" data-method="create">
+        <button class="btn btn-default" ng-click="RolesCtrl.createUpdateRoleModal()" data-method="create">
```
jniles, line 10: We should probably standardize this nomenclature.  My former Python self agrees "explicit is better than implicit".

jniles (approved):
Good cleanup, nice comments.  Goes above and beyond the original issue to fix a number of issues.  Looks good to me!

I like the proposal to change actions that do both creation and update/edit actions to the name "createUpdate".  I think it makes a lot of sense and is clearer than our previous mix of "manage", "config", "edit", etc.  I'll make an issue about it.

bors r+

jniles:
Wow!  A successful merge! 💃

PR #3116, opened 2018-09-01
Fix partial path matching on client side authorisation (3114)

### LINKED ISSUE #3114: Client side authorisation allows partial routes without correct permissions
Client side authorisation is currently implemented by intercepting the `$stateChangeStart` event. If the target of the state change is deemed valid the change is allowed to happen. 

Currently the method that checks authorisation is as follows: 
```js
function checkUserAuthorization(data) { 
  return path.indexOf(data.path) === 0 && data.authorized;
}
```
Working on client side roles and permissions it was noted that this allows subsets of routes to pass even if the user does not have that specific permission. 

An example of why: 

```js 
// bhima route - /cash
// bhima route - /cashboxes 
const permittedRoute = '/cash';
const restrictedRoute = '/cashboxes';

const path = '/cashboxes';

// as permittedRoute is included in the users permissions - this route _will_ be checked
// against the invalid route

// returns 0 - allowing the route, even though this route is restricted to the user
path.indexOf(permittedRoute)
```

sfount:
N.B The mechanisms for client and server side authorisation haven't been significantly reviewed since the beginning of the project. All of this core code should be brought up to the modern project standards.

This commit fixes an issue with client side authorisation allowing
partial routes that the user does not have permission to, under this
system '/cashboxes' would be a valid path because the user has access to
'/cash'. 

As described in #3114. 

This is fixed by adding a second step checking each section of the
required path against the target path, if the base of the target path
match each required section the routing is approved.

My understanding is that the reason

```js 
return validPath === targetPath;
``` 

was not used is that client side routing relies heavily on passing parameters through the browser path. For example `/cash` and `/cash/2` are both perfectly valid paths and both should be allowed access to the `/cash` page. However '/cash/2' does not === '/cash' and this simple method would reject this.

This proposed solution, until a significant rework of the codebase, is to split the paths up and ensure that the **base** of the target path matches the permission path **until** the permission path is complete, any number of variables can then follow.

Example - partial route blocking: 
```js 
// in this example the user does NOT have access to the cashboxes module
const targetPath = '/cashboxes'; 

// the user DOES have access to the cash module 
const validPath = '/cash';

// returns 0 - this is where the previous algorithm considered this an authorised route
// it is clear however that cashboxes !== cash
validPath.indexOf(targetPath) === 0;

// target path split routes 
// ['', 'cashboxes']

// valid path split routes 
// ['', 'cash']

// because cash !== cashboxes the proposed algorithm will reject this authorisation
```

Example - allowing variable passing beyond the base path: 
```js
const targetPath = '/cash/2';

// in this case the user DOES have permission to access the module
// however they are accessing the second cashbox module by passing a variable 
const validPath = '/cash';

// this still returns 0 - the original algorithm would authorise this route
// however note that validPath === targetPath would NOT authorise this route
validPath.indexOf(targetPath) === 0

// target path split routes 
// ['', ''cash', '2']

// valid path split routes 
// ['', 'cash'] 

// because '' === '' AND 'cash' === 'cash' the proposed algorithm will
// authroise this route it is only the base of the path that needs to exactly 
// match the permission route in order to be accepted. 
```

This can be easily tested, removing the cashbox module permission and navigating to /cashbox will still allow the user access. Note until the inclusion of https://github.com/IMA-WorldHealth/bhima-2.X/pull/3115 you will have to log out and back in to properly update local user permissions, https://github.com/IMA-WorldHealth/bhima-2.X/pull/3115 ensures roles and permissions are reloaded on any change. 

Note: This method doesn't consider `?` params however I don't believe they are used on the client, this should definitely be addressed before merging if there is an example of this. However end to end tests covering a majority of client functionality pass this method.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/3114

### DIFF
--- a/client/src/js/app.js
+++ b/client/src/js/app.js
@@ -105,8 +105,8 @@ function startupConfig(
 
     // check if we are going to an error state;
     const isErrorState = (
-      next.name.indexOf('404') !== -1 ||
-      next.name.indexOf('403') !== -1
+      next.name.indexOf('404') !== -1
+      || next.name.indexOf('403') !== -1
     );
 
     const isSettingsState = next.name.indexOf('settings') !== -1;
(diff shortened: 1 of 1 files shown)


jniles (approved):
Nice catch!

No issues here, and this fixes an import bug.  The documentation is great!

PR #3121, opened 2018-09-01
Show users with access to cash box feature (2402)

### LINKED ISSUE #2402: Show users with access to a Cashbox on the Cashboxes page
We implemented User-Caisse assignment in #2242.  It's very difficult to know who has permission to a given cashbox once permissions have been assigned.

I propose that we list the users with access to the cashbox on cashboxes page.  This will give administrators a clear view of who has access and allow them to add or remove access as needed.  For larger institutions, this will become critical.

sfount:
This PR implements the feature requested in https://github.com/IMA-WorldHealth/bhima-2.X/issues/2402. 

It shows a list of users that have access to a given cash box on the Cash Box Management page, allowing quick review of which employees can use different cash windows. 

It also implements the component `bh-user`. This is a component that should be useful in a number of other places in the application, it shows a system user's display name, username, active status and time since they have last logged in. It can be used as follows: 

```html
<!-- Directly provide the user details object --> 
<bh-user details="user"></bh-user>

<!-- This component uses an experimental 'fetch' binding which -->
<!--  if found useful could be standardised. Including fetch in the --> 
<!-- component definition instructs the component to download the details -->
<bh-user id="5" fetch></bh-user>
```

Example of the `bh-user` component: 

screenshot from 2018-09-01 23-38-28

Example of cash box management, users with permission feature: 

screenshot from 2018-09-01 23-22-39

Closes #2402

### DIFF
--- a/client/src/i18n/en/cashbox.json
+++ b/client/src/i18n/en/cashbox.json
@@ -13,12 +13,11 @@
 
     "USERS": { 
       "TITLE": "Cashbox Users",
-      "SUBTITLE": "Users",
-      "INFO_1": "To update users with access to this cash box, use the",
-      "INFO_2": "tool.",
-      "MANAGEMENT_LINK": "Users Management",
-      "NO_PERMISSIONS": "No users have permission to user this cashbox",
-      "LAST_SEEN": "Last seen"
+      "SUBTITLE": "Users with access to this cash box:",
+      "INFO_LINK": "To update users with acess to this cash box, use the <a ui-sref='users.list'>Users Management</a> tool.",
+      "NO_PERMISSIONS": "No users have permission to use this cashbox",
+      "LAST_SEEN": "Last seen",
+      "DEACTIVATED": "This user is deactivated."
     }
   }
 }
--- a/client/src/i18n/en/cashbox.json
+++ b/client/src/i18n/en/cashbox.json
@@ -9,6 +9,16 @@
     "DESCRIPTION":"This module is responsible for cashbox management",
     "TITLE":"Cashbox Management",
     "NO_SUPPORTED_CURRENCIES":"This cashbox does not have any supported currencies.",
-    "SET_CURRENCY_ACCOUNTS":"Set currency accounts for this cashbox."
+    "SET_CURRENCY_ACCOUNTS":"Set currency accounts for this cashbox.",
+
+    "USERS": { 
+      "TITLE": "Cashbox Users",
+      "SUBTITLE": "Users",
+      "INFO_1": "To update users with access to this cash box, use the",
+      "INFO_2": "tool.",
+      "MANAGEMENT_LINK": "Users Management",
+      "NO_PERMISSIONS": "No users have permission to user this cashbox",
+      "LAST_SEEN": "Last seen"
+    }
   }
 }
--- a/client/src/i18n/en/cashbox.json
+++ b/client/src/i18n/en/cashbox.json
@@ -11,10 +11,10 @@
     "NO_SUPPORTED_CURRENCIES":"This cashbox does not have any supported currencies.",
     "SET_CURRENCY_ACCOUNTS":"Set currency accounts for this cashbox.",
 
-    "USERS": { 
+    "USERS": {
       "TITLE": "Cashbox Users",
       "SUBTITLE": "Users with access to this cash box:",
-      "INFO_LINK": "To update users with acess to this cash box, use the <a ui-sref='users.list'>Users Management</a> tool.",
+      "INFO_LINK": "To update users with access to this cash box, use the <a ui-sref='users.list'>Users Management</a> tool.",
       "NO_PERMISSIONS": "No users have permission to use this cashbox",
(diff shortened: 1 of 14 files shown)


client/src/i18n/en/cashbox.json
```diff
@@ -9,6 +9,17 @@
     "DESCRIPTION":"This module is responsible for cashbox management",
     "TITLE":"Cashbox Management",
     "NO_SUPPORTED_CURRENCIES":"This cashbox does not have any supported currencies.",
-    "SET_CURRENCY_ACCOUNTS":"Set currency accounts for this cashbox."
+    "SET_CURRENCY_ACCOUNTS":"Set currency accounts for this cashbox.",
+
+    "USERS": { 
+      "TITLE": "Cashbox Users",
+      "SUBTITLE": "Users",
+      "INFO_1": "To update users with access to this cash box, use the",
```
jniles: Just FYI, it's possible to include HTML tags in these translation values, as well as variable substitutions.  An example of that is [in patient records](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/i18n/en/patient_records.json#L43).  I'm pretty sure you could even include a `ui-sref` with the addition of a [translate-compile](https://angular-translate.github.io/docs/#/guide/05_using-translate-directive#using-translate-directive_post-compiling) tag.

Personally, I prefer this method (even though it seems impure) since it allows you to preserve the context of the statement and allow developers to infer from the string what is going to happen.  It would look something like this for your code:
```js
"INFO" : "To update users with access to this cash box, use the <a href ui-sref="users.list">Users Management</a> link."
```

What do you think?

jniles: Of course you'll need to then have `translate-compile` on the `translate` directive tag in your HTML.

client/src/js/components/bhUser.js
```diff
@@ -0,0 +1,59 @@
...
+    },
+  });
+
+bhUser.$inject = ['UserService'];
+
+/**
+ * @component bhUser
+ *
+ * @description
+ * Displays a sysyem users details.
```
jniles: This should probably be `system` :D

client/src/js/components/bhUser.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * If a `details` object is provided, this will be used by the component, otherwise
+ * a `id` can be provided in combination with `fetch`, the component will be
+ * responsible for getting the latest data.
+ *
+ * @example
+ * <bh-user details="user"></bh-user>
```
jniles, line 25: Nice examples!

client/src/i18n/en/cashbox.json
jniles: I would also add "The following users have access to this cashbox:" at the very end, just to be explicit.

test/integration/cashboxes.js
```diff
@@ -154,6 +154,21 @@ describe('(/cashboxes) The Cashboxes API endpoint', () => {
       .catch(helpers.handler);
   });
 
+  it('GET /cashboxes/:id/users should return users subscribed to a cashbox', () => {
```
jniles, line 157: :dancer

server/controllers/admin/users/cashboxes.js
```diff
@@ -26,10 +26,10 @@ function list(req, res, next) {
   `;
 
   db.exec(sql, [req.params.id])
-    .then((rows) => {
-      rows = rows.map(row => row.cashbox_id);
+    .then((cashboxes) => {
```
jniles, line 29: Helpful!

jniles (commented):
Code in general looks good to me.  Appreciate the lint updates!

My comment about angular-translate is at your discretion.  Personally, I think it is easier to reason about, but have never had feedback on it.  Let me know your thoughts.

client/src/i18n/en/cashbox.json
sfount: This is good to know about! I will follow this suggestion for this example; I think the context of what is being translated is lost slightly when broken into 3 pieces. 

In general though I think that it's concerning to put client application logic in i18n translation files. In this case it's fairly simple, I think any use of `<b>` or `<u>` or general style changes should be fine. Potentially we should be more careful when it comes to including elements that change the way the application works.

```html
<span translate translate-compile translate-sanitize-strategy="'sce'">
    CASHBOX.USERS.INFO_LINK
</span> 
``` 

Does read clearer than the previous implementation, however.

sfount:
@jniles Thanks for the feedback! I've commented on the angular-translate thread here https://github.com/IMA-WorldHealth/bhima-2.X/pull/3121#discussion_r214709502. The suggestion has been applied to this PR. 

In general, compiling style and markup seems reasonable for brevity. Including application logic feels concerning with the number of places you would have to look to potentially try and hunt down an issue.

jniles:
Awesome.  I'll add the translation tags and we'll be set for a merge.

jniles:
Translations added.

bors r+

PR #3123, opened 2018-09-02
Refactor gulp build process, address file name versioning development issue

### LINKED ISSUE #1328: (bug) Repeated building break the file system
By repeatedly building the server, it is possible to create a file long enough the file system resulting in a  message like the one below.

hashminificationbuildbreaking
_Fig 1: File Name too long to Store_

This is due to the minifaction/revision pipeline in gulp naively appending the filename to the already hashed filename.  We should make sure that the versioned files are not used as inputs into the gulp pipeline and are successfully cleaned up on rebuild.

sfount:
This pull request updates `gulpfile.js` to accomplish four main goals: 
### 1. Developement `watch` fix

**Fix issue that causes version file names to grow too large and break the `gulp watch` process** 
:tada: :+1: :1st_place_medal: :fireworks: :champagne: 

### 2. Simplify process 

```js
gulp.task('client', ['js', 'css', 'less', 'i18n', 'vendor', 'static'], templateHTML);
```

Make the process easier to understand, the current version in master includes many different styles and quirks of gulp. This refactor follows the linting process and attempts to do everything in the same `gulp.src` `pipe` style.

There is nothing special about the client build, it just processes javascript, css, less, translation files, vendor files and static assets. It then links all versioned build files by templating them into HTML. 

### 3. Improve versioning system 

We use `gulp-rev` to generate hashes of files and include these in filenames in order to 'cache-bust' in development and production. This has been updated and standardised so that every part of the build process can version itself and update the collective manifest file. 

It also updates the cleaning system so that versioned files are removed in favour of the most up to date copy.

### 4. Standardise process 

Move towards standardising the build process to allow future upgrades. [(upcoming gulp 4 release)](https://github.com/gulpjs/gulp/tree/4.0)

### Dependency changes: 
* `rimraf` - library deprecated in favour of `del` [(github)](https://github.com/robrich/gulp-rimraf)
* `gulp-rev-replace` - library deprecated in favour of `gulp-rev-rewrite` [(github)](https://github.com/jamesknelson/gulp-rev-replace)
* `merge-stream` - library added [(github)](https://github.com/grncdr/merge-stream)

This change updates dependencies so a `yarn install` is required. 

This PR should be thoroughly tested by the development team and preferably in production scenarios before being merged with production code.

(The offending `watch` problem:)
screenshot from 2018-09-02 19-29-42

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/1328


jniles:
Hey, this is awesome!  I like it.  Let me spin it up on my dev environment and see what it does to performance.

jniles (approved):
Looks good to me! We've needed a good re-write of the gulpfile for a while.  It seems to have about the same performance on my local machine.

bors r+

sfount:
This might be worth testing on the core teams development machines as soon as possible, it doesn't affect too many files so shouldn't be too difficult to revert.

jniles:
Yeah, you are probably right.  We'll do that.

sfount:
I've just thought the `watch` task will need to watch and run the 'static' method, otherwise template HTML files won't be reflected after changes. 

I'll update this PR shortly.

jniles:
I thought we were gonna have a phone call ☎️ ?

jniles:
@sfount give this a rebase and see if it all works out?

sfount:
@jniles Looks like it still passes after rebasing :+1: . I've also included the changes mentioned here https://github.com/IMA-WorldHealth/bhima-2.X/pull/3123#issuecomment-417950608.

jniles:
@mbayopanda when you get a chance, can you pull this PR and see if it passes tests on your machine?

client/src/index.html
```diff
@@ -10,10 +10,10 @@
 
   <!-- Compiled bootstrap style definition -->
   <link rel="stylesheet" href="vendor/angular-ui-select/dist/select.css">
-  <link rel="stylesheet" href="css/bhima-bootstrap.css">
+  <link rel="stylesheet" href="bhima-bootstrap.css">
   <link rel="stylesheet" href="vendor/angular-ui-grid/ui-grid.min.css">
   <link rel="stylesheet" href="vendor/components-font-awesome/css/font-awesome.min.css">
-  <link rel="stylesheet" href="css/style.min.css">
+  <link rel="stylesheet" href="css/bhima.min.css">
```
mbayopanda: It seems like we have some bhima css files inside `/css` folder and other outside the `/css` folder, it will be good to have all bhima css files inside one folder such as `/css` or `/css/bhima/` or flattened into one file `bhima.min.css`.

client/src/index.html
```diff
@@ -104,8 +104,8 @@ <h4><a class="title-content" ui-sref="index">{{ AppCtrl.project.name }}</a></h4>
 </div>
 
 <!-- scripts loaded in body to prevent page from 'hanging'-->
-<script src="js/vendor.min.js"></script>
-<script src="js/bhima.min.js"></script>
+<script src="vendor.min.js"></script>
+<script src="js/dist/bhima.min.js"></script>
```
mbayopanda: `bhima.min.js` is the endpoint of all angular code of bhima, so we can put it in the same level (the root folder) as `vendor.min.js`.

All we need in `/bin/client` as javascript code are `vendor.min.js` and `bhima.min.js`, so there is no need to have different folders.

gulpfile.js
```diff
@@ -58,86 +38,61 @@ const paths = {
       'client/src/**/*.js',
       '!client/src/i18n/**/*.js',
     ],
-    css         : ['client/src/css/*.css'],
-    vendorStyle : [
```
mbayopanda, line 62: 👍

client/src/index.html
sfount: @mbayopanda thanks for the comment! I will update these to be clearer for the build process. 

The organisation of these files isn't as important as `client/src` but should still be considered.

The reason for the different folder is because of the 'cleanup' process. The last build process simply deleted the entire `bin/client` folder on the `client-build` and didn't delete anything on the `watch` method. This iteration tries to delete smaller folders only when they are used, for example `clean-js` will only delete the `bin/client/js` folder when a new `bhima.min.js` is being generated.  For this reason nothing else was but in the `bin/client/js` folder because it would have been deleted at the time. 

I propose to try and keep the targetted folder clean style but factor in your comments we make these files build into seperate js subfolder, for example: 

```
- bin/ 
- - bin/client 
- - - bin/client/js
- - - - bin/client/js/bhima/
- - - - - bhima.min.js
- - - - bin/client/js/vendor/
- - - - - vendor.min.js
```

The same goes for CSS as per your previous comment.

client/src/index.html
sfount: This should be addressed with the same logic as https://github.com/IMA-WorldHealth/bhima-2.X/pull/3123#discussion_r215196781.

sfount:
@jniles @mbayopanda 

I've standardised the build folder structure to clear up the `index.html` references as per your comments: 
```html
<link rel="stylesheet" href="css/bootstrap/bhima-bootstrap.css">
<link rel="stylesheet" href="css/bhima/bhima.min.css">

<script src="js/vendor/vendor.min.js"></script>
<script src="js/bhima/bhima.min.js"></script> 
```

This folder structure is simpler to understand but allows the targeted file cleanup tasks to just remove folders relevant to file changes.

PR is rebased with the latest `master` resolving `yarn.lock` conflicts.

sfount:
Just noticed this closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/1328

mbayopanda (approved):
LGTM 👍 

bors  r+

PR #3128, opened 2018-09-03
Issue 3056 PR

### LINKED ISSUE #3056: Use bhInlineFilter on Role Management
If you ever have a long list of roles, locating a single one can be a nightmare.  We need to have a `bhInlineFilter` button on these pages.  You can see an example of one in the accounts management module code.

Scholldier:
This PR adds filtering to the Roles Management module.

Closes #3056


Scholldier:
I didn't load this PR correctly, so I'm going to redo it.

PR #3129, opened 2018-09-03
Update roles.ctrl.js

Scholldier:
This is to edit the second necessary file (roles.ctrl.js), adding the inline filtering code, that resolves issue #3056.


Scholldier:
I didn't load this PR correctly, so I'm going to redo it.

PR #3130, opened 2018-09-03
PR for Issue 3056

### LINKED ISSUE #3056: Use bhInlineFilter on Role Management

Scholldier:
This PR is to update two files to add inline filtering to the Roles Management module.

Closes #3056

### DIFF
--- a/client/src/modules/roles/roles.ctrl.js
+++ b/client/src/modules/roles/roles.ctrl.js
@@ -3,7 +3,7 @@ angular.module('bhima.controllers')
 
 RolesController.$inject = [
   '$uibModal', 'RolesService', 'SessionService', 'ModalService',
-  'NotifyService', 'bhConstants', 'uiGridConstants'
+  'NotifyService', 'bhConstants', 'uiGridConstants',
 ];
 
 function RolesController($uibModal, Roles, Session, Modal, Notify, bhConstants, uiGridConstants) {
--- a/client/src/modules/roles/roles.ctrl.js
+++ b/client/src/modules/roles/roles.ctrl.js
@@ -120,8 +120,8 @@ function RolesController($uibModal, Roles, Session, Modal, Notify, bhConstants, 
    * @description
    * Switches the inline filter on and off.
    */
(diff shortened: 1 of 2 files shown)


Scholldier:
So this code tests fine, EXCEPT that there is also a filter field above the 'Actions' -- which shouldn't be the case. Can you point me in the right direction to resolve that? I'll keep digging on my side but I don't see anything obvious yet.

Scholldier:
I found it -- I added code to make it so that the actions column can't be filtered on and it tested correctly.

Scholldier:
There are some things I need to fix according to the code climate test, which I'll review and fix tomorrow.

client/src/modules/roles/roles.ctrl.js
```diff
@@ -92,6 +92,7 @@ function RolesController($uibModal, Roles, Session, Modal, Notify, bhConstants)
     headerCellFilter : 'translate',
   }, {
     field : 'actions',
+    enableFiltering : false,
```
jniles, line 95: Glad you figured this out!  I suppose I should have mentioned it 😄

jniles:
@Scholldier, no problem!  The code climate fixes should be straightforward.

Our tests on `master` are pretty flakey, but I've just landed a fix that should help... if it helps the other pull requests that are failing, I'll rebase this PR onto master while merging and we'll see if that solves the issue.

Scholldier:
Haha, no worries -- it's good for me to put my troubleshooting hat on for learning sake. 👍

client/src/modules/roles/roles.ctrl.js
Scholldier: Haha, no worries! It's good for me to learn through searching, trial and error, and such. 👍

jniles:
Code looks good to me.  Downloaded and tested this and it behaves as excepted.  Good work.

bors r+

PR #3133, opened 2018-09-03
Fix `bhFindPatient` expose API exception

### LINKED ISSUE #3132: bhFindPatient optional parameter `onRegisterAPI` throws error if not defined
This behaviour can be observed on the 'Patient Standing Report'. 

The report uses the `bh-find-patient` component and does not require the API exposed, it does not set the `on-register-api` parameter and an error is thrown in the console every time the module is loaded. 

As `onRegisterApi` is bound to optional `&?` the component behaviour should respect this and not throw errors if it is not defined. 

To reproduce: 
1. Load the Patient Standing Report

```
TypeError: vm.onRegisterApi is not a function
```

sfount:
This commit includes an additional check on the `$onInit` method for the
`bh-find-patient` component. This check ensures the API expose flag has
been set before executing the callback, this is inline with the optional
binding used by the component.

It also updates `vm` to `$ctrl` in favour of Angular component naming
standards.

Closes #3132

### DIFF
--- a/client/src/js/components/bhFindPatient.js
+++ b/client/src/js/components/bhFindPatient.js
@@ -38,7 +38,7 @@ FindPatientComponent.$inject = [
  *     should be run
  */
 function FindPatientComponent(Patients, AppCache, Notify, Session, bhConstants, Barcode) {
-  const vm = this;
+  const $ctrl = this;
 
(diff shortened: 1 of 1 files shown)


jniles (approved):
Looks good to me.  Nice bug fix!

bors r+

PR #3136, opened 2018-09-04
PR for Issue 3068

### LINKED ISSUE #3068: Remove the "pin" column from the user table
We used to use pins for high-privileged actions in the first version of our software, but we no longer use them.  It's just one more thing to remember!  Let's remove this column from the user table.

Scholldier:
This PR is for issue #3068 -- @jniles are there other call-outs to user that I also need to account for? I don't see any others...

Closes #3068


jniles:
Nope!  It was just that one vestigial column to my knowledge.  Thanks!

bors r+

Issue #3135, opened 2018-09-03
Management module card list component


sfount:
A number of management modules make use of a 'card list' to display entities with a collection of useful attributes. 

This pattern can currently be found on: 
* Creditor group management 
* Debtor group management 
* Fiscal Year (Slight variation on the theme) 

This pattern makes sense where there are entities with a lot of meta data but not too many records. Many records would best suit a grid but this view provides a middle ground of clearly legible but also able to provide a lot of information. Potential other candidates for this pattern are: 
* Cash box management 
* Patient group management 

Example: 
![screenshot from 2018-09-03 17-37-01](https://user-images.githubusercontent.com/2844572/44996477-021ec800-afa0-11e8-8402-09bc4f881470.png)

This issue proposes a design for a component to simplify, test and reduce duplication implementing this pattern across modules. 

Component requirements: 
* Display a generic list passed through binding 
* Allow some method of sorting and filtering on the list in a user friendly way
* Include basic tests 
* Allow templating for card layout, not all entities have exactly the same attributes

Potential usage: 
```html
<!-- name - sorted A-Z and filtered -->
<!-- age - sorted oldest, newest -->
<!-- size - sorted largest, smallest -->
<bh-card-list 
  data="debtors" 
  name="debtorName"
  age="debtorDate"
  size="debtorSize"
  template="/my/debtor/template.card.html">
</bh-card-list>
```

mbayopanda:
Hi @sfount 

What do you think to have a pagination on this `bh-card-list`, sometime the list is too long a pagination can be useful for avoiding to scroll down.

sfount:
Hey @mbayopanda great suggestion! I have not included pagination with the initial pull request for two reasons: 
* the use case in this PR does not expect to have many entities
* the UX for pagination in combination with filtering should also be carefully designed - it can be confusing for pagination to change when you filter for less results. 

If this Component is merged and implemented then pagination would be a great feature to support on top of the base filtering and sorting! 

(This comment relates to https://github.com/IMA-WorldHealth/bhima-2.X/pull/3137#discussion_r214874542)

Issue #3122, opened 2018-09-01
UI Improvements on Cashbox Management module


sfount:
Similar to https://github.com/IMA-WorldHealth/bhima-2.X/issues/3120 there are a number of tweaks to the Cashbox Management module that will make it simpler to use for people that already understand the BHIMA UX. 

1. - [x] Create label should be applied to the Cashbox Management segment of the headercrumb, not in it's own segment
![screenshot from 2018-09-01 23-44-10](https://user-images.githubusercontent.com/2844572/44950504-0b455300-ae41-11e8-8f75-91e4c1ed2f26.png)

(For example)
![screenshot from 2018-09-01 23-45-36](https://user-images.githubusercontent.com/2844572/44950509-2617c780-ae41-11e8-89c5-127860c94ceb.png)

2. - [x] A standard should be proposed for 'danger' actions on management pages and then followed, Cashbox Management here could potentially follow Debtor Group Management (Danger Zone may need to be renamed to something users can clearly understand - this could be researched)
![screenshot from 2018-09-01 23-47-42](https://user-images.githubusercontent.com/2844572/44950526-c5d55580-ae41-11e8-986b-2a622f44bb66.png)
![screenshot from 2018-09-01 23-49-36](https://user-images.githubusercontent.com/2844572/44950527-c5d55580-ae41-11e8-91e9-725a90a6c35b.png)

3. - [x] (See https://github.com/IMA-WorldHealth/bhima-2.X/pull/3121) The 'Edit' button on management pages like this should be changed to 'Manage' or 'Configure' to reflect what you are able to do. For example on cash boxes you can do many things other than edit, for example edit the name, change currency accounts, view users with permissions etc. One suggestion would be 'Manage Cash Box'. 
![screenshot from 2018-09-01 23-52-18](https://user-images.githubusercontent.com/2844572/44950542-50b65000-ae42-11e8-9c7e-b95e7cafa5e9.png)

3.a. 
- [x] This change should also be made on the Debtor Group Management module 
Issue to track https://github.com/IMA-WorldHealth/bhima-2.X/issues/3148

jniles:
There are some good ideas here.  These pages are used so infrequently that it probably isn't worth _too much_ time trying to get right, but we can definitely do better.

I think we might want to do away completely with the current cashbox configuration page - it's very similar to the old 1.x - style of configuration.  It would probably be nicer to have them in a debtor-groups style layout without an extra modal to configure each currency.  What do you think of going down that route?

[...]

PR #3137, opened 2018-09-04
`bh-card-list` and `bh-card` components

sfount:
This pull request implements the proposed `bh-card-list` API found in https://github.com/IMA-WorldHealth/bhima-2.X/issues/3135. 

This component is then applied to the Cashbox module, significantly contributing to https://github.com/IMA-WorldHealth/bhima-2.X/issues/3122. 

The `bh-card-list` component displays a list of cards provided a list of entities, it provides sorting and filtering functionality on these entities. 

**Example usage:** 
```html
<bh-card-list
  data = "arrayOfThings"
  name = "thingName"
  template = "/path/to/my/thing/template.card.html">
</bh-card-list>
```

This will generate the following (example the cashbox implementation in this PR):
screenshot from 2018-09-04 08-20-39

Note a benefit of the larger card layout is making actionable information very clear to the user - in this example cashboxes that no users can access are highlighted with a red exclamation point, administrators will have to provide permissions before these can be used in the application. 

In order to significantly reduce boiler plate for implementing this design pattern in other modules, this PR also implements the `bh-card` component, this uses [multiple transclusion](https://docs.angularjs.org/api/ng/directive/ngTransclude) introduced in angular 1.5x to template in data at different points in the cards layout. 

**Example usage:** 
```html
<bh-card>
  <bh-card-title>{{card.label}}</bh-card-title>
  <bh-card-subtitle>Project: {{card.project_name}}</bh-card-sutitle>
  <bh-card-edit><a ui-sref="card.edit">Edit Entity</bh-card-edit>
  <bh-card-detail>{{card.category_label}}</bh-card-detail>
  <bh-card-statistic>{{card.number_of_users}} total users</bh-card-statistic>
</bh-card>
```

Sorting and filtering are automatically applied based on the information passed through the `name` and optional `age` and `size` bindings. Any properties that aren't bound will simply not be shown to the user. 

**Filtering**
screenshot from 2018-09-04 08-21-35

**Sorting**
screenshot from 2018-09-04 09-14-04

Contributes to https://github.com/IMA-WorldHealth/bhima-2.X/issues/3135.

### DIFF
--- a/client/src/js/components/bhCardList.js
+++ b/client/src/js/components/bhCardList.js
@@ -13,27 +13,81 @@ angular.module('bhima.components')
     ,
   });
 
-function bhCardList() {
+bhCardList.$inject = ['$timeout'];
+
+/**
+ * @description
+ * Component for displaying cards of information, providing sorting and filtering
+ * components.
+ *
+ * Requires:
+ * `data` : the array to iterate over
+ * `name` : the entity display name that should be sorted and filtered
+ * `template` : path to an HTML template that should be displayed for each element
+ */
+function bhCardList($timeout) {
   const $ctrl = this;
+  const DEFAULT_IDENTIFIER = 'uuid';
 
   // name, age and size options will be added to this _if_ they are defined
   // by through bindings
-  $ctrl.orderOptions = {};
-  $ctrl.filterOptions = {};
+  $ctrl.orderOptions = { active : false, available : [] };
+  $ctrl.filterOptions = { active : false };
 
-  $ctrl.currentOrderAttribute = $ctrl.name;
-  $ctrl.currentOrderReversed = false;
-  $ctrl.currentFilterValue = undefined;
-
-  $ctrl.identifier = $ctrl.id || 'uuid';
+  // keep internal copy of data so that it is only exposed to the view once
+  // an identifer has been determined
 
   $ctrl.$onInit = function onInit() {
 
+    // setup defaults
+    $ctrl.currentOrderAttribute = $ctrl.name;
+    $ctrl.currentFilterAttribute = $ctrl.name;
+    $ctrl.currentOrderReversed = false;
+
+    $ctrl.identifier = $ctrl.id;
+
+    // name is required
+    $ctrl.orderOptions.available.push({ attribute : $ctrl.name, key : 'TABLE.COLUMNS.SORTING.NAME_ASC', reverse : false });
+    $ctrl.orderOptions.available.push({ attribute : $ctrl.name, key : 'TABLE.COLUMNS.SORTING.NAME_DSC', reverse : true });
+
     if (angular.isDefined($ctrl.age)) {
+      $ctrl.orderOptions.available.push({ attribute : $ctrl.age, key : 'TABLE.COLUMNS.SORTING.CREATED_ASC', reverse : false });
+      $ctrl.orderOptions.available.push({ attribute : $ctrl.age, key : 'TABLE.COLUMNS.SORTING.CREATED_DSC', reverse : true });
+    }
+
+    if (angular.isDefined($ctrl.size)) {
+      $ctrl.orderOptions.available.push({ attribute : $ctrl.size, key : 'TABLE.COLUMNS.SORTING.TOTAL_ASC', reverse : true });
     }
   };
 
   $ctrl.$onChanges = function onChanges(changes) {
     console.log(changes);
+    if (changes.data && angular.isDefined(changes.data.currentValue)) {
+      if (!$ctrl.identifier) {
+        // no identifier override found
+        $ctrl.identifier = defaultIdentifier(changes.data.currentValue);
+        console.log('set up identifier', $ctrl.identifier);
+      }
+    }
+
+    $ctrl.toggleFilter = function toggleFilter() {
+      $ctrl.filterOptions.active = !$ctrl.filterOptions.active; $ctrl.filterOptions.value = {};
+      console.log($ctrl.filterOptions);
+    }
+
+    $ctrl.setOrder = function setOrder(order) {
+      $ctrl.orderOptions.current = order;
+      $ctrl.orderOptions.active = angular.isDefined(order);
+
+      console.log('setting order', $ctrl.orderOptions);
+    }
+
+    // attempt to pick an identifier based on common ids
+    // this is overriden by binding a specific `id`
+    // supported default identifiers are 'id' and 'uuid'
+    function defaultIdentifier(data) {
+      const sampleData = data[0] || {};
+      return angular.isDefined(sampleData.uuid) ? 'uuid' : 'id';
+    }
   }
 }
--- a/client/src/js/components/bhCardList.js
+++ b/client/src/js/components/bhCardList.js
@@ -1,0 +1,39 @@
+angular.module('bhima.components')
+  .component('bhCardList', {
+    templateUrl : 'modules/templates/bhCardList.tmpl.html',
+    controller : bhCardList,
+    bindings : {
+      data : '<',
+      template : '@',
+      name : '@', // name attribute to be sorted alphabetically and filtered
+      age : '@?', // date attribute to be sorted by date
+      size : '@?', // size attribute to be sorted largest to smallest
+      id : '@?', // optional data override
+    }
+    ,
+  });
+
+function bhCardList() {
+  const $ctrl = this;
+
+  // name, age and size options will be added to this _if_ they are defined
+  // by through bindings
+  $ctrl.orderOptions = {};
+  $ctrl.filterOptions = {};
+
+  $ctrl.currentOrderAttribute = $ctrl.name;
+  $ctrl.currentOrderReversed = false;
+  $ctrl.currentFilterValue = undefined;
+
+  $ctrl.identifier = $ctrl.id || 'uuid';
+
+  $ctrl.$onInit = function onInit() {
+
+    if (angular.isDefined($ctrl.age)) {
+    }
+  };
+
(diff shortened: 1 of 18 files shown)


client/src/js/components/bhCardList.js
```diff
@@ -0,0 +1,100 @@
...
+    $ctrl.orderOptions.available.push(
+      { attribute : name, key : 'TABLE.COLUMNS.SORTING.NAME_ASC', reverse : false }
+    );
+    $ctrl.orderOptions.available.push(
+      { attribute : name, key : 'TABLE.COLUMNS.SORTING.NAME_DSC', reverse : true }
+    );
+
+    // parse optional bindings
+    if (angular.isDefined(age)) {
+      $ctrl.orderOptions.available.push(
```
jniles, line 81: Just FYI, you can pass multiple parameters into `Array.prototype.push`, so you could write:
```js
$ctrl.orderOptions.available.push(
  { attribute : age, key : 'TABLE.COLUMNS.SORTING.CREATED_ASC', reverse : false },
  { attribute : age, key : 'TABLE.COLUMNS.SORTING.CREATED_DSC', reverse : true }
);
```
to achieve the same result.

client/src/modules/cash/cashboxes/cashboxes.js
```diff
@@ -13,7 +13,7 @@ CashboxController.$inject = [
  * A valid cashbox must have accounts defined for each enterprise currency, for
  * ease of use thought the application.
  */
-function CashboxController(Session, Projects, Boxes, util, Notify, $state) {
+function CashboxController(Session, Projects, CashBoxes, util, Notify, $state) {
```
jniles, line 16: Good choice of renaming 👍

client/src/modules/templates/bhCard.tmpl.html
```diff
@@ -0,0 +1,23 @@
+<div class="panel panel-default bh-card" data-group-entry="{{card.id}}">
+  <div class="panel-body">
+    <div class="row">
+      <div class="col-md-6">
+        <h4 style="margin-top : 0px;" ng-transclude="title"></h4>
+        <h5 style="margin-top : 0px;" ng-transclude="subtitle"></h5>
+      </div>
+      <div class="col-md-6 text-float-util">
```
jniles: What is `text-float-util`?  I think this is first time I've seen it!

client/src/modules/templates/bhCardList.tmpl.html
```diff
@@ -0,0 +1,45 @@
...
+    class="form-control"
+    placeholder="{{'FORM.PLACEHOLDERS.FILTER_NAME' | translate }}"
+    style="border-radius: 0"/>
+</div>
+
+<div
+  ng-repeat="card in $ctrl.data
+    | filter : $ctrl.filterOptions.value : strict
+    | orderBy : 'label':$ctrl.orderOptions.current.reverse">
+  <ng-include src="$ctrl.template"></ng-include>
```
jniles: Should we support pagination?  I know our Fiscal Year controller already has support for it via `uib-pagination`.  Do you think we could put it in here?

client/src/modules/templates/bhCardList.tmpl.html
```diff
@@ -0,0 +1,45 @@
...
+        </ul>
+      </span>
+    </div>
+  </div>
+
+  <input
+    ng-model="$ctrl.filterOptions.value[$ctrl.currentFilterAttribute]"
+    ng-show="$ctrl.filterOptions.active"
+    class="form-control"
+    placeholder="{{'FORM.PLACEHOLDERS.FILTER_NAME' | translate }}"
```
jniles: Just FYI - with any plain HTML element, you can use `angular-translate` attributes for a slight performance increase, instead of using the expensive pipe.  However, this does not work for custom components (read: `ui-select`), which is why you still see this pattern in those components.

To use the translate attributes, you can use the [translate-attr](https://angular-translate.github.io/docs/#/api/pascalprecht.translate.directive:translate-attr) directive to do it in one pass:
```html
<input
  translate-attr="{ placeholder : 'FORM.PLACEHOLDERS.FILTER_NAME' }">
```

server/controllers/finance/cashboxes/index.js
```diff
@@ -36,9 +36,11 @@ exports.privileges = privileges;
 function list(req, res, next) {
   const filters = new FilterParser(req.query);
 
-  let sql =
-    'SELECT id, label, is_auxiliary, project_id FROM cash_box ';
+  let sql = 'SELECT id, label, is_auxiliary, project_id FROM cash_box ';
 
+  // @TODO(sfount) this isn't a detailed query for cashboxes because it returns
```
jniles, line 43: Yeah, good call!  We can change the API for this down the road.

jniles (commented):
@sfount, great work on this.  It's a very nice prototype already!

I'd be interested in writing some unit tests for this code before merging. It will serve as a really good example of how to unit test logic (sorting, filtering) in both the controller and the DOM.  If I write the tests, it would also go to validate your API design (unless it seems super confusing, in which case, it would prompt reforms).  What do you think?  Shall I make an attempt or do you think it isn't stabilized yet?

sfount:
@jniles Thanks! 

I think you writing unit tests to cover the functionality would be excellent, having a clean standard to follow would be very worth while. 

Regarding testing the API I think it's especially important to test the identifier and optional parameter binding parsing logic, as these are the behaviours that could be most easily interrupted by other changes, for example: 
* Providing the `size` binding make the component provide sorting by size, the same goes for `age`
* Providing a custom `id` binding overrides the default parsed ID 

It would be great to ensure with a test that the sorting behaviours are all linked together correctly, i.e the click event on the sort menu does set the correct `orderOptions` and this is reflected in the DOM however as all of the actual sorting behaviour is handed over to angularjs we can just cover the linking of all the elements in our tests.

If you can take the tests, I will follow up with one more `chore()` commit that addresses your comments and cleans things up a little bit, it won't change the API in any way!

client/src/modules/templates/bhCard.tmpl.html
sfount: Apparently and bizarrely it sets `text-align: right`. `text-float-util` is just a ruse to make the programmer think this is more important than it is :disappointed_relieved:. 

It looks like this is a carry over from older code, I will _consider_ using the more aptly named bootstrap `text-right`.

client/src/modules/templates/bhCardList.tmpl.html
sfount: Good to know!

client/src/modules/templates/bhCardList.tmpl.html
sfount: I did notice pagination on the fiscal year controller. I think with all the modules in line to support this pattern none of them will have an overwhelming number of entities that would require both filtering and paginating. 

Definitely something to consider for future improvement though! The nice thing about setting up like this will be updating the Component to support pagination across all modules that implement this.

sfount:
I've updated this branch with master to solve conflicts caused by this recent merge https://github.com/IMA-WorldHealth/bhima-2.X/pull/3121. I also responded to comments on the code. This rebase is the last commit I'll force push to allow contributions to the unit tests.

jniles:
@sfount, I have implemented a few unit tests for this.

I'm having some trouble with the filtering tests - mind taking a second glance at it for me?  It is the only skipped test.  It might just be something I'm misunderstanding about the API.

Based on my experience here, I propose the following changes: 
 1. Rename the SORTING.NAMES_DSC variable to SORTING.NAMES_DESC.  It makes more sense coming from SQL.
 2. The binding names are a bit confusing, since I wasn't sure what "size" meant.  So, I propose we rename it to: 
    1. `size` => `sort-size`
     2. `age` => `sort-age`
    3. `name` => `sort-name`

I think that will help developers understand what the API is doing when binding variables to it.

test/client-unit/components/bhCardList.spec.js
```diff
@@ -0,0 +1,175 @@
...
+    expect(filterInput).not.to.have.class('ng-hide');
+
+    angular.element(toggle).click();
+    $scope.$digest();
+
+    filterInput = find(element, '[data-filter-input]');
+    expect(filterInput).to.have.class('ng-hide');
+  });
+
+  it.skip('should support filtering the values in the DOM', () => {
```
sfount: It looks like this is behaving as it should! I haven't tried this yet but you're setting the filter value (name, as you have well suggested changing) to be `title`. 

It looks like all three titles contain the word 'Card' and we are then filtering on 'd' so all 3 should be rendered. If the `name` attribute was set to `text` I think the results would be as expected, with only 1 rendered.

sfount (commented):
These are really great unit tests, very clear and easy to read! The skipped test shouldn't cause too much issue. 

I'm not sure why the yarn.lock file was included? Would you like to fix this test or should I pull it, rebase and then push a version to be merged?

EDIT: I agree that the name change suggestions you made should be applied, they can be made along with the last test and resolving yarn lock issues.

jniles:
Either way works for me.  If you are in the middle of something I can do it.

test/client-unit/components/bhCardList.spec.js
sfount: To update this: It looks like the offending line is on the input:
```html
ng-model = "$ctrl.filterOptions[$ctrl.name]"
```

In the browser this correctly evaluates as follows: 
screenshot from 2018-09-10 16-37-32

But in the client test runner the $ctrl.name is not respected: 
screenshot from 2018-09-10 16-37-15

It looks like a quirk with the test runner but I will look into it a bit more.

jniles:
I added translations.  This is good to go.

bors r+

PR #3127, opened 2018-09-02
Date components - limit min date fiscal year feature (2785)

sfount:
This PR implements the functionality described in https://github.com/IMA-WorldHealth/bhima-2.X/issues/2785. 

The flag `limit-min-fiscal` has been applied to the `bh-date-interval` and `bh-date-editor` components, these are the two primarily used for reports. 

This flag will limit the date editor so that it can only select up to the beginning of the enterprises first fiscal year. 

This PR applies the `limit-min-fiscal` flag to the following modules: 
* Cash report 
* Cashflow by service 
* Client report 
* Cashflow 
* Stock exit 
* Balance sheet 
* Stock inventory

Dates in 2015, this is the earliest fiscal year so these are valid
screenshot from 2018-09-02 19-51-53

Dates in 2014 are before the opening fiscal year, these are invalid
screenshot from 2018-09-02 19-52-06

Year view 
screenshot from 2018-09-02 19-52-25

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/2785.

PR #3142, opened 2018-09-05
refractor(fiscal year) set end date not earlier than the start date

### LINKED ISSUE #3140: Fiscal Year allows you to set the end date earlier than the beginning
On the Fiscal Year creation page, you are allowed to set the end date earlier than the beginning, resulting in a negative length Fiscal Year.  We have the ability to set a min date in our uib-datepicker, let's block out dates in the ending input that are earlier than the beginning.  This way, users can't accidentally shoot themselves in the foot.

screenshot 2018-09-04 at 19 28 03
_Fig 1: Fiscal Year Date Error_

jeremielodi:
closes #3140

### DIFF
--- a/client/src/js/components/bhDateInterval.js
+++ b/client/src/js/components/bhDateInterval.js
@@ -53,11 +53,12 @@ function bhDateInterval(moment, bhConstants) {
     ];
 
     vm.label = vm.label || 'FORM.SELECT.DATE_INTERVAL';
-
     vm.dateFormat = bhConstants.dayOptions.format;
-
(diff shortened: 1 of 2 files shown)


jniles:
Nice bug fix!  Looks good to me.

bors r+

sfount:
Quick note! https://github.com/IMA-WorldHealth/bhima-2.X/pull/3127 will have to be updated to factor in the splitting of picker options logic. It might be useful to name them `pickerFromOptions` and `pickerToOptions` to clearly differentiate them.

jniles:
Good call!

PR #3143, opened 2018-09-05
Test(fiscal year) : integration test for opening balance

mbayopanda:
This PR add integration tests for opening balance API and fix some not necessary codes.

### DIFF
--- a/client/src/modules/fiscal/fiscal.service.js
+++ b/client/src/modules/fiscal/fiscal.service.js
@@ -93,12 +93,11 @@ function FiscalService(Api) {
     const url = service.url.concat(id, '/periods');
     return service.$http.get(url)
       .then(service.util.unwrapHttpResponse)
-      .then(periods =>
-        periods.map(p => {
-          p.start_date = new Date(p.start_date);
-          p.end_date = new Date(p.end_date);
-          return p;
-        }));
+      .then(periods => periods.map(p => {
+        p.start_date = new Date(p.start_date);
+        p.end_date = new Date(p.end_date);
+        return p;
+      }));
   }
 
   return service;
--- a/server/controllers/finance/fiscal.js
+++ b/server/controllers/finance/fiscal.js
@@ -16,6 +16,7 @@ const _ = require('lodash');
 const db = require('../../lib/db');
 const Transaction = require('../../lib/db/transaction');
 const NotFound = require('../../lib/errors/NotFound');
+const BadRequest = require('../../lib/errors/BadRequest');
 const FilterParser = require('../../lib/filter');
 
 const Tree = require('../../lib/Tree');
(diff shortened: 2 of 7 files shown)


client/src/modules/fiscal/fiscal.service.js
```diff
@@ -93,12 +93,11 @@ function FiscalService(Api) {
     const url = service.url.concat(id, '/periods');
     return service.$http.get(url)
       .then(service.util.unwrapHttpResponse)
-      .then(periods =>
-        periods.map(p => {
-          p.start_date = new Date(p.start_date);
-          p.end_date = new Date(p.end_date);
-          return p;
-        }));
+      .then(periods => periods.map(p => {
```
jniles, line 96: :+1:

server/controllers/finance/fiscal.js
```diff
@@ -396,17 +396,12 @@ function setOpeningBalance(req, res, next) {
...
     .then((hasPrevious) => {
-      let promise;
-
       if (hasPrevious) {
-        // load from the period N+1 of the year N-1 into period 0 of the year N
-        promise = loadOpeningBalance(fiscalYear);
-      } else {
-        // set new opening balance
-        promise = newOpeningBalance(fiscalYear, accounts);
+        const msg = `The fiscal year with id ${id} is not the first fiscal year`;
```
jniles, line 400: Good catch!

test/integration/fiscalYear.extra.js
```diff
@@ -0,0 +1,144 @@
...
+        const dec = res.body[12];
+        expect(flatDate(jan.start_date)).to.be.equal('2017-01-01');
+        expect(flatDate(jan.end_date)).to.be.equal('2017-01-31');
+        expect(flatDate(dec.start_date)).to.be.equal('2017-12-01');
+        expect(flatDate(dec.end_date)).to.be.equal('2017-12-31');
+      })
+      .catch(helpers.handler);
+  });
+
+  function flatDate(_date_) {
```
jniles, line 136: Just FYI - you could also use moment for this without worry.  That way, you could do
```js
function flatDate(data) {
  return moment(data).format('YYYY-MM-DD');
}
```

test/integration/fiscalYear.extra.js
```diff
@@ -0,0 +1,144 @@
...
+        const [fy] = res.body;
+        expect(fy).to.be.an('object');
+        expect(fy).to.have.property('fiscal_year_id', 3);
+        expect(fy).to.have.property('previous_fiscal_year_id', 2);
+        expect(fy).to.have.property('label', 'Test Fiscal Year 2017');
+
+        /**
+         * dates are returned with timezone information
+         * we extract just the date information
+         * FIXME: timezone issue : we have 2017-12-31 in the database
```
jniles: Looks like this is an old comment that should be removed.  The test below works without a timezone issue 👍

test/integration/fiscalYear.extra.js
```diff
@@ -0,0 +1,144 @@
...
+        expect(res).to.have.status(201);
+      })
+      .catch(helpers.handler);
+  });
+
+  it('GET /fiscal/:id/opening_balance returns the opening balance of a given fiscal year', () => {
+    return agent.get(url.concat(`/${fiscalYear2015.id}/opening_balance`))
+      .then(res => {
+        const updatedAccounts = res.body.filter(item => {
+          return [81, 83, 174, 190].indexOf(item.id) > -1;
```
jniles: You can also use `includes()`!  It would make it more readable to name the array inside as well.  Like this:

```js
const accountSubset = [81, 83, 174, 190];
const updatedAccounts = res.body.filter(account => accountSubset.includes(account.id));
```

jniles: It might be a good idea to sort it as well after downloading it - so that any changes in ordering do not break your test.

```js
updatedAccounts.sort();
// then
const [first, second, third, fourth] = updatedAccounts;
```

test/integration/fiscalYear.extra.js
```diff
@@ -0,0 +1,144 @@
...
+      })
+      .catch(helpers.handler);
+  });
+
+  it('GET /fiscal/:id/periods returns periods of the given fiscal year', () => {
+    return agent.get(url.concat(`/${fiscalYear2017.id}/periods`))
+      .then(res => {
+        expect(res.body).to.have.length(fiscalYear2017MonthsNumber + 1);
+        for (let i = 0; i < res.body.length; i++) {
+          expect(res.body[i].number).to.be.equal(i);
```
jniles: This took me a minute to understand what it was doing.  Could you leave a comment?  It seems to me it is checking that the fiscal year periods are returned in order.

jniles (commented):
Looks like some good improvements!

Nothing wrong with the logic from what I can tell.  I've made a few comments to improve the readability of the code.

test/integration/fiscalYear.extra.js
mbayopanda: You're right 👍

jniles:
Changes made!

bors r+

Issue #3118, opened 2018-09-01
Cashbox access permissions - no changes results in Server 500 Error


sfount:
Submitting the cashbox access permissions modal without making any changes results in the 500 error returned from the server. 

To replicate: 
1. Load the User Management module 
2. Edit the first users cashbox permissions with `Actions` -> `Cashbox Management` 
3. Click Submit

A number of things should be happening here: 
* Ideally - The client recognises no changes are made and makes no GET requests to the server 
* [x] Required - The server handles a no changes state and returns 200 success or 400 bad request depending on this specific API end point 
* [x] Required - A test is written to ensure this is covered, sending no changes should not break the server

The client also does not respond to the 500 error with a user facing notification, this failure is silently logged to the console. 
* [x] Required - Client responds to server error for cashbox permissions submission

Bonus:
- [x] The client side modal 'rejects' even on submission - this needs to be standardised across the client.

jniles:
Yeah, I just ran into this as well.  Good catch.

To round this out:
 1. If no changes occur, we can use the same pattern as on the patient record page and just flash "no changes detected".

I feel that the server should send a 400 if the update API is called without giving a list of cashboxes.  That way, the client can handle it properly with a message.

sfount:
I had code ready to resolve this issue but didn't get around to writing the PR - sorry for confusion.

PR #3144, opened 2018-09-05
Fix cash box access permission update

sfount:
This commit introduces two checks on the server `/users/:id/cashboxes`
route that ensure that the correct data is provided before continuing,
this lets the client know if a request is badly formed in order to tell
the user.

It also updates the client controller to remove the default case where
it would send a badly formed request if no changes were made.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/3118

### DIFF
--- a/test/integration/users.js
+++ b/test/integration/users.js
@@ -279,4 +279,13 @@ describe('(/users) Users and Permissions', () => {
       })
       .catch(helpers.handler);
   });
+
+  it('POST /users/:id/cashboxes will reject invalid data formats', () => {
+    return agent.post(`/users/${newUser.id}/cashboxes`)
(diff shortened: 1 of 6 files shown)


test/integration/users.js
```diff
@@ -279,4 +279,13 @@ describe('(/users) Users and Permissions', () => {
       })
       .catch(helpers.handler);
   });
+
+  it('POST /users/:id/cashboxes will reject invalid data formats', () => {
```
jniles, line 283: Perfect 👍

jniles (approved):
Looks good to me!  Thanks for including a test 👍 

bors r+

PR #3154, opened 2018-09-06
fix(Complex Voucher): do not break on exchange

### LINKED ISSUE #3152: Changing the currency in Complex Vouchers locks up the grid
For some reason, if you change currency whilst manipulating the Complex Vouchers grid, the application refuses to add or remove lines.  You can see a GIF reproduced below off this issue.

**Steps to Reproduce**
 1. Start a complex voucher, filling out a few lines of the grid.
 2. Set the transaction type, description and date, but do not set the currency.
 3. Add a line and fill it out.
 4. Change currency.
 
Observe that you can no longer add lines or remove them.  If the values are not correctly set, then it will lock up the complex voucher module completely.

_GIF 1: Lock up the Vouchers for fun and profit_

jniles:
This commit improves the exchange function on the Complex Voucher grid by removing the asynchronous code to refresh the exchange rate and by ensuring the grid is updated and not broken when updated.  It also fixes lint issues in the Exchange Rate Service.

Closes #3152.

### DIFF
--- a/client/src/i18n/en/transactions.json
+++ b/client/src/i18n/en/transactions.json
@@ -5,7 +5,7 @@
     "INCLUDE_POSTED_TRANSACTIONS_DETAIL": "Include posted transactions in search",
     "INCLUDE_POSTED_TRANSACTIONS_SHORT": "Posted Transactions",
     "SINGLE_ACCOUNT_TRANSACTION":"This transaction only concerns a single account.  Please select multiple accounts to complete the transaction.",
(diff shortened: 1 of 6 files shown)


jniles:

lomamech:
@jniles I just tested the change, while trying to change the currency several times, adding lines, deleting lines, and so on, except that it has a message that should only appear if '
we choose a treasury account or banks 
image

jniles:
@lomamech thanks for the review!

It looks like we made transaction types a required field (https://github.com/IMA-WorldHealth/bhima-2.X/issues/2323 ; https://github.com/IMA-WorldHealth/bhima-2.X/pull/2766), but the message is misleading.  I'll re-word the message to say that a transaction type is required.  Thanks!

jniles:
Fixed:
voucherfix

PR #3156, opened 2018-09-07
Fix simple vouchers state transition to patient debt transfer tool

### LINKED ISSUE #3141: Simple vouchers 'Patient Invoice Debt Transfer' tool has no action
It isn't clear if this is a bug or a deprecated feature that hasn't been correctly removed, this should be investigated and either fixed or cleaned up. 

No errors are thrown in the console, a 'Menu' is displayed with an actionable item however there is no action.

sfount:
Addresses lack of functionality issue found in https://github.com/IMA-WorldHealth/bhima-2.X/issues/3141. If another developer could verify this is an issue on `master` before merging this patch that would be great. I can't think of a reason this behaviour should change.

This commit fixes an issue with the state transition on the Simple
Vouchers module. The Simple Vouchers barcode scanning route does not
define any `params` but the transition was not being respected unless an
empty object was bound through the `ui-sref` definition.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/3141


jniles (approved):
Tested and works for me.

bors r+

PR #3158, opened 2018-09-07
use date string for avoiding timezone info on fiscal year date range

mbayopanda:
This PR remove timezone information in fiscal year for avoiding timezone influence on the fiscal year date range.

Issue : Dates are depending on the timezone

Fix : Remove dependency to timezone by using date string in the format `YYYY-MM-DD`

### DIFF
--- a/client/src/modules/fiscal/fiscal.manage.js
+++ b/client/src/modules/fiscal/fiscal.manage.js
@@ -109,6 +109,10 @@ function FiscalManagementController($state, Fiscal, Notify, Modal, util, moment)
     // get the number of months
     numberOfMonths();
 
+    // remove timezone information by considering just date
+    vm.fiscal.start_date = moment(vm.fiscal.start_date).format('YYYY-MM-DD');
+    vm.fiscal.end_date = moment(vm.fiscal.end_date).format('YYYY-MM-DD');
+
     const promise = isUpdate ? Fiscal.update(id, vm.fiscal) : Fiscal.create(vm.fiscal);
 
     return promise


sfount:
This is a great fix! Date processing is in need of review.

A quick question, if anyone has an idea about this, is it possible for `node` servers or `MySQL` databases to interpret 'YYYY-MM-DD' differently based on the locale of the server/ database machine? I know this is a fairly standard format but this is up for interpretation on the server. Potentially if the server and client were both configured to read and write the same format this would avoid that issue, if it exists.

jniles:
So, one confusing thing about the whole date/datetime issue is that JavaScript by default represents all dates relative to UTC.  [Here is the ECMAScript specification for it](http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15).  Specifically, it says:
> Z	is the time zone offset specified as “Z” (for UTC) or either “+” or “-” followed by a time expression HH:mm

That means that when we have a string that looks like this:
```js
const s = new Date():
console.log(s.toISOString()); //=> 2018-09-10T07:06:26.697Z
```
It is saying that I generated that string at 7:06:26 UTC time... which happens to be 8:06 WAT time (my current timezone).  So _technically_ the date is correctly capturing the correct time.

The confusing issue is when we don't need the time component of it.  For example, if we want to say this happened on midnight at 2018-01-01, but we don't want to say _whose timezone that midnight was in_, really, that means we do not need the time component.

Unfortunately, by default `mysqljs` read out strings into JS Date objects (which automatically assign the time to it), but [we can turn this off in the connection properties](https://github.com/mysqljs/mysql#connection-options).  That means we might be able to get away with just writing into a DATE field in MySQL (which only keeps the date portion, and then read something like:
```js
db.exec('SELECT start_date FROM fiscal_year WHERE id = ?', [1], { dateStrings : false })
  .then(rows => { 
    console.log(rows[0].start_date); // => this should now be a string, not a date
  });
```

That might give us some control over the date issues faced here?

jniles:
This solves the immediate problem!  Good catch @mbayopanda and awesome gifs :)

bors r+

sfount:
@jniles the `dateStrings` flag is really good to know about. I think something that would go towards fixing this issue is being explicit rather than implicit. Right now many of our dates just spin up a `new Date()` or a `new Date(some_string)` and take the default behaviour. 

I think a good common practice is to lean into the UTC behaviour and essentially serialise all dates as UTC time, only ever taking it out of UTC to **display** to the end user, in this case the client nor the server would ever process anything in any other timezone. 

If we are trying to specify on the client midnight on 2018-01-01 it shouldn't matter according to the data model what timezone you're in as the specified time is UTC 2018-01-01 00:00:00, this would be sent from the client to the server and interpreted as a UTC date before writing. 

The only time a date would be taken out of UTC time is when the client is showing what time something happened in which case it should take the UTC timezone and format it according to their locale. To do this correctly the client will have to pass it's timezone locale details to the server to generate reports etc.

If both the client and the server were explicit about sending and receiving dates in UTC that may contribute towards this also.

Issue #3134, opened 2018-09-03
Replace validationTrigger binding in components with child form$submitted


jniles:
As discussed in https://github.com/IMA-WorldHealth/bhima-2.X/pull/3064, with Angular 1.7, parent forms now submit their child forms.  This means that we are able to forgo the `validationTrigger` call to check if the parent form is submitted... instead, we can just listen directly on the `form.$submitted` event like we would in any other case.

PR #3159, opened 2018-09-07
feat(accounts reports) report on transtions for multiple accounts

### LINKED ISSUE #3103: Report on Transactions in Multiple Accounts
A useful report would combine data from multiple accounts.  It would provide the same options as the Account Statement Report, but allow you to select more than one account to report on via `<ui-select multiple>` for the accounts.  This is a feature requested by our clients.

Ideally, it should be available on two presentations: **combined** and **separated**.  The "combined" presentation will put all the transactions in the same `<table>` and add an extra column for the "account" so that we can determine from which account the transaction came from.  The "separated" view would basically repeat the "Account Statement Report" multiple times for each account selected, one `<table>` after the other.

For example:
p_20180903_081525

jeremielodi:
closes #3103


server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,275 @@
...
+    })
+    .then((result) => {
+      res.set(result.headers).send(result.report);
+    })
+    .catch(next)
+    .done();
+}
+
+/**
+ * @function getGeneralLedgerSQL
```
sfount: It looks like this method already exists in `reportAccounts` https://github.com/IMA-WorldHealth/bhima-2.X/blob/59a150ec96c18fa053e877e047761d03f34311d1/server/controllers/finance/reports/reportAccounts/index.js#L106

There is a useful CS principle known as [Don't repeat yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) (or DRY). Keeping code in one place means that when we update features, text or fix bugs we can just change the core code and not have to search around.

Can this be moved to a shared library so that all accounts that use this logic can share the code? (Something like `AccountsExtra`)

server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,275 @@
...
+
+  const query = filters.applyQuery(sql);
+  const parameters = filters.parameters();
+
+  return { query, parameters };
+}
+
+// @TODO define standards for displaying and rounding totals, unless numbers are rounded
+//       uniformly they may be displayed differently from what is recorded
+function getTotalsSQL(options) {
```
sfount: https://github.com/IMA-WorldHealth/bhima-2.X/blob/59a150ec96c18fa053e877e047761d03f34311d1/server/controllers/finance/reports/reportAccounts/index.js#L172

(Same principle as comment on `getGeneralLedgerSQL`)

server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,275 @@
...
+}
+
+
+/**
+ * @function getAccountTransactions
+ *
+ * @description
+ * This function returns all the transactions for an account,
+ */
+function getAccountTransactions(options, openingBalance = 0) {
```
sfount: https://github.com/IMA-WorldHealth/bhima-2.X/blob/59a150ec96c18fa053e877e047761d03f34311d1/server/controllers/finance/reports/reportAccounts/index.js#L205

(Same principle as comment on `getGeneralLedgerSQL`)

server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,258 @@
...
+const Currency = require('../../currencies');
+const Fiscal = require('../../fiscal');
+const FilterParser = require('../../../../lib/filter');
+
+const TEMPLATE = './server/controllers/finance/reports/reportAccountsMultiple/report.handlebars';
+
+/**
+ * @method document
+ *
+ * @description
```
sfount, line 18: It would be good to update this comment description to explain the difference between this and the accounts report.

server/models/bhima.sql
```diff
@@ -98,7 +98,8 @@ INSERT INTO unit VALUES
   (204, 'Exchange Rate','TREE.EXCHANGE','',1,'/modules/exchange/exchange','/exchange'),
   (205, 'Account Reference Management','TREE.ACCOUNT_REFERENCE_MANAGEMENT','',1,'/modules/account_reference','/account_reference'),
   (206, '[OHADA] Bilan','TREE.OHADA_BALANCE_SHEET','',144,'/modules/reports/ohada_balance_sheet_report','/reports/ohada_balance_sheet_report'),
-  (207, 'Account Reference Report','TREE.ACCOUNT_REFERENCE_REPORT','',144,'/modules/reports/account_reference','/reports/account_reference');
+  (207, 'Account Reference Report','TREE.ACCOUNT_REFERENCE_REPORT','',144,'/modules/reports/account_reference','/reports/account_reference'),
```
sfount, line 101: @jniles How should changes to core BHIMA tables be tracked for upgrading production servers? 

Is `bhima.sql` run whenever and upgrade is performed? It does include things that may have changed like `transaction_type`. Otherwise is there a method of tracking changes that don't require a schema update?

sfount (commented):
This looks like a good first pass on the multiple accounts report, great work! 

My main feedback here is on code that can be shared between this report and the account statement report. It looks like around ~200 lines of code could be removed from this and shared by the reports in a library.

The issue also mentions a 'combined' or 'separated' flag for the reports presentation: 
> Ideally, it should be available on two presentations: combined and separated. The "combined" presentation will put all the transactions in the same table and add an extra column for the "account" so that we can determine from which account the transaction came from. The "separated" view would basically repeat the "Account Statement Report" multiple times for each account selected, one table after the other

This feature can be discussed in the issue proposal thread if it is not appropriate, otherwise it would be good to implement it to properly close the feature.

jeremielodi:
@sfount This PR is ready for a review

.env.development
```diff
@@ -5,7 +5,7 @@ PORT=8080
 DB_HOST='localhost'
 DB_USER='bhima'
 DB_PASS='HISCongo2013'
-DB_NAME='bhima_test'
+DB_NAME='imck'
```
jniles: This should be reset to `bhima_test`.

client/src/i18n/en/report.json
```diff
@@ -109,6 +109,12 @@
...
       "WARN_CURRENCY": "Attention! You are using a currency other than the enterprise currency.  Values on this report are subject to change with changes to the exchange rate and should not be considered final.  Please save a report in the enterprise currency for long term reference."
     },
+    "REPORT_ACCOUNTS_MULTIPLE": {
+      "TITLE": "Accounts Report",
+      "DESCRIPTION": "This report shows all transactions for a group of accounts given a set time period.",
+      "WARN_MULTIPLE": "N.B. This report concerns an income/ expense account and spans multiple fiscal years. Income/ expense accounts are cleared to 0 at the beginning of every fiscal year, the cumulative balance displayed is for the transactions within this period of time. It is recommended to use this statement within a single fiscal year.",
+      "WARN_CURRENCY": "Attention! You are using a currency other than the enterprise currency.  Values on this report are subject to change with changes to the exchange rate and should not be considered final.  Please save a report in the enterprise currency for long term reference."
```
jniles: If these keys are the same as the previous report, you can just re-use them.  There is no need to repeat ourselves.

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.config.js
```diff
@@ -0,0 +1,107 @@
...
+function AccountReportMultipleConfigController(
+  $sce, Notify, SavedReports, AppCache, reportData, $state,
+  Moment, Session
+) {
+  const vm = this;
+  const cache = new AppCache('configure_account_report_multiple');
+  const reportUrl = 'reports/finance/account_report_multiple';
+  vm.previewGenerated = false;
+  vm.accountIds = [];
+  delete cache.reportDetails;
```
jniles: Why do we need to delete the `cache.reportDetails`?  Doesn't that defeat the purpose of a cache?

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.config.js
```diff
@@ -0,0 +1,107 @@
...
+  vm.reportDetails = {
+    currency_id : Session.enterprise.currency_id,
+    selectedAccounts : [],
```
jniles: Shouldn't we save the accounts so that when a user comes back to this screen, they find the same accounts?  It's how the other account report works.  See below.

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.config.js
```diff
@@ -0,0 +1,107 @@
...
+
+function AccountReportMultipleConfigController(
+  $sce, Notify, SavedReports, AppCache, reportData, $state,
+  Moment, Session
+) {
+  const vm = this;
+  const cache = new AppCache('configure_account_report_multiple');
+  const reportUrl = 'reports/finance/account_report_multiple';
+  vm.previewGenerated = false;
+  vm.accountIds = [];
```
jniles: Why do we need both `vm.accountIds` and `vm.reportDetails.selectedAccounts`?  We should keep all the data in one array.

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.config.js
```diff
@@ -0,0 +1,107 @@
...
+    currency_id : Session.enterprise.currency_id,
+    selectedAccounts : [],
+  };
+
+  vm.dateInterval = 1;
+
+  checkCachedConfiguration();
+
+  vm.selectAccount = function selectAccount(account) {
+    vm.reportDetails.selectedAccounts.push({ type_id : account.type_id, id : account.id });
```
jniles: Do you ever use `type_id`?  Why keep it around?

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.html
```diff
@@ -0,0 +1,62 @@
...
+            <bh-account-select-multiple
+              id="account-id"
+              account-ids="ReportConfigCtrl.accountIds"
+              label="FORM.SELECT.ACCOUNTS"
+              name="account"
+              on-select-callback="ReportConfigCtrl.selectAccount(account)"
+              on-change="ReportConfigCtrl.onChangeAccounts(id)"
+              exclude-title-accounts="true"
+              required="true"
+              validation-trigger="ConfigForm.$submitted">
```
jniles: It looks like this is the only place `bh-account-select-multiple` is used.  Could you implement https://github.com/IMA-WorldHealth/bhima-2.X/issues/3134 while you are here?

Basically, you need to remove `validation-trigger` from here, the component `bindings` and the component template itself.  You can also remove the `name` binding.

When you have done both of those, change `ng-form="{{ $ctrl.name }}"` in the [first line of the bhAccountSelectMultiple template](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/templates/bhAccountSelectMultiple.tmpl.html#L1) to `ng-form="AccountSelectMultiple"`.

Then you can change everywhere in the template that said `$ctrl.validationTrigger` and replace it with `AccountSelectMultiple.$submitted`.

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.html
```diff
@@ -0,0 +1,62 @@
...
+            </bh-account-select-multiple>
+
+            <!-- Date interval  -->
+            <!-- @TODO this should use a component that callback with well defined dates -->
+            <bh-date-interval
+              date-from="ReportConfigCtrl.reportDetails.dateFrom"
+              date-to="ReportConfigCtrl.reportDetails.dateTo"
+              required="true"
+              validation-trigger="ConfigForm.$submitted">
```
jniles, line 42: It looks like [validationTrigger was removed from bh-date-interval](https://github.com/IMA-WorldHealth/bhima-2.X/commit/2e8ebe0910ae7df77231c464c09d63dbcb10f59d#diff-33e4c895023c285d3bf777adbe56d375).  Please remove it here.

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.html
```diff
@@ -0,0 +1,62 @@
...
+            </bh-date-interval>
+
+            <!-- the currency to be used in the footer -->
+            <bh-currency-select
+              currency-id="ReportConfigCtrl.reportDetails.currency_id"
+              on-change="ReportConfigCtrl.setCurrency(currency)"
+              validation-trigger="ConfigForm.$submitted">
```
jniles: [It looks like `validationTrigger` was removed from bh-currency-select](https://github.com/IMA-WorldHealth/bhima-2.X/commit/e1d4778077facfee296cdc7a3005e046e43aabe4#diff-a67ac182eb7399f3660336328199d452).  Please remove it here.

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.config.js
```diff
@@ -0,0 +1,107 @@
...
+    if (form.$invalid) {
+      Notify.danger('FORM.ERRORS.RECORD_ERROR');
+      return 0;
+    }
+
+    parseDateInterval(vm.reportDetails);
+
+    vm.reportDetails.selectedAccounts = vm.reportDetails.selectedAccounts.filter(ac => {
+      return vm.accountIds.indexOf(ac.id);
+    });
```
jniles: What is this statement doing?  It isn't clear to me.

server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,97 @@
...
+        rate,
+        invertedRate,
+      };
+
+      _.extend(bundle, { header });
+
+      const { selectedAccounts } = params;
+      const accounts = [].concat(selectedAccounts);
+      return q.all(accounts.map(account => {
+        const parseAccount = JSON.parse(account);
```
jniles: This statement tells me that you are not sending account IDs to the server, but JSON objects.  You should be sending an array of one or more account IDs to the server.  Like this:
```
GET some/report?account_ids=1,2,3
```
That way, there won't be any need to call `JSON.parse` on each value.

server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,97 @@
...
+        params.account_id = parseAccount.id;
+        params.account = parseAccount;
+
+        return getAccountTransactions(params, bundle.header.exchangedBalance);
+      }));
+    })
+    .then(results => {
+      _.extend(bundle, { Alltransactions : results }, { params });
+      return Fiscal.getNumberOfFiscalYears(params.dateFrom, params.dateTo);
```
jniles: It doesn't look like you ever use `Fiscal.getNumberOfFiscalYears()`.  Why did you call it?

server/models/bhima.sql
jniles: Ugh.  Right now we just manually upgrade things.  I'm not sure of a better way, or how to make sure people respect the better way when writing new code.  As a reviewer, I find it hard to remember to remind people to make upgrade scripts.

jniles (changes requested):
There is a lot of code here, and most looks good.  I've made some comments on some confusion that needs to be cleared up.

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.config.js
jeremielodi: It's already like that

client/src/i18n/en/report.json
jeremielodi: okay

client/src/modules/reports/generate/account_report_multiple/account_report_multiple.config.js
jeremielodi: Here we are using a select multiple; when a selected account is removed , this snipet of code will  filter  ```vm.reportDetails.selectedAccounts```  then it will be reduced.
You can see at the screen shot bellow how accounts are displayed

selected accounsts

jeremielodi: You can notice that I need in the ui-select both then label and the account number but when the user remove the selected account, the on-remove callback return just the id of the account

jniles:
Looks like this isn't quite done:

The account dropdown should have the account numbers on it.  This is how all other account dropdowns work.
fixmultipleaccountreport
_Fig 1: Account Dropdown Needs Numbers_

The currency symbol should be rendered in the account header:
needscurrencyaccount

The opening balance needs to be in the same currency as what the user chose.
openingbalanceshouldbeincorrectcurrency

jniles:

jeremielodi:
Ready to be merged

jniles:
Could you fix the conflicts first?  Also, are you sure you fixed the issues in the last review? It doesn't look like any changes were made.

client/src/i18n/en/report.json
```diff
@@ -109,6 +109,10 @@
       "WARN_MULTIPLE": "N.B. This report concerns an income/ expense account and spans multiple fiscal years. Income/ expense accounts are cleared to 0 at the beginning of every fiscal year, the cumulative balance displayed is for the transactions within this period of time. It is recommended to use this statement within a single fiscal year.",
       "WARN_CURRENCY": "Attention! You are using a currency other than the enterprise currency.  Values on this report are subject to change with changes to the exchange rate and should not be considered final.  Please save a report in the enterprise currency for long term reference."
     },
+    "REPORT_ACCOUNTS_MULTIPLE": {
+      "TITLE": "Multiple Accounts Report",
+      "DESCRIPTION": "This report shows all transactions for a group of accounts given a set time period."
```
mbayopanda, line 114: What about : This report displays all transactions of a group of accounts for a given period

client/src/i18n/fr/tree.json
```diff
@@ -83,6 +83,7 @@
     "PURCHASE_REGISTRY":"Registre des achats",
     "REPORTS":"Rapports",
     "REPORT_ACCOUNTS":"Rapport de comptes",
+    "REPORTS_MULTIPLE_ACCOUNTS" : "Rapport de comptes mutiple",
```
mbayopanda: You would say `multiples` instead of `mutiple`.

Also, i think the correct phrase could just be : "Rapport des comptes"

client/src/i18n/fr/tree.json
```diff
@@ -83,6 +83,7 @@
     "PURCHASE_REGISTRY":"Registre des achats",
     "REPORTS":"Rapports",
     "REPORT_ACCOUNTS":"Rapport de comptes",
```
mbayopanda: This translation could be updated to "Rapport de compte" in singular

mbayopanda: Also seeing that `Rapport de compte` and `Rapport des comptes` are close in the navigation tree, it can be a bit confusing for users, so we can add a suffixes (simple and multiple) to differentiate them : 

- "Rapport de compte" => "Rapport de compte (simple)", in the report page we just keep "Rapport de compte"
- "Rapport des comptes" => "Rapport des comptes (multiple)", in the report page we just keep "Rapport des comptes"

mbayopanda (commented):
The list of accounts doesn't display the account number 

image

The translation here could just be "Rapport des comptes"

image

🔥 **There is not opening balance**

The "Rapport de compte (simple)" has the opening balance for the selected account but not the "Rapport des comptes (multiple)"

- "Rapport de compte (simple)" : for 57110011 
image

- "Rapport de compte (multiple)" : for 57110011 
image

server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,111 @@
...
+    })
+    .then(_accountTransactions => {
+      accountTransactions = _accountTransactions;
+      // let get the opening balance of each account
+      return q.all(accountTransactions.map(transaction => {
+        return accountExtra.getOpeningBalanceForDate(transaction.account.id, params.dateFrom, false);
+      }));
+    })
+    .then(openginBalances => {
+      openginBalances.forEach((balance, index) => {
```
jniles: It's easier to read if you change this variable name from `openginBalances` to `openingBalances`.

server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+
+      return q.all(accounts.map(account => {
+        params.account_id = account.id;
+        params.account = account;
+        return getAccountTransactions(params, bundle.header.exchangedBalance);
+      }));
+    })
+    .then(_accountTransactions => {
+      accountTransactions = _accountTransactions;
```
jniles, line 94: This variable is useless.  Do ahead and just call the parameter `accountTransactions`.

server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,111 @@
...
+      // let get the opening balance of each account
+      return q.all(accountTransactions.map(transaction => {
+        return accountExtra.getOpeningBalanceForDate(transaction.account.id, params.dateFrom, false);
+      }));
+    })
+    .then(openginBalances => {
+      openginBalances.forEach((balance, index) => {
+        accountTransactions[index].header = balance;
+      });
+      _.extend(bundle, { Alltransactions : accountTransactions }, { params });
```
jniles: Usually only class names start with a capital letter.  Could you rename this variable `allTransactions`?

server/controllers/finance/reports/reportAccountsMultiple/index.js
```diff
@@ -0,0 +1,111 @@
...
+  } catch (e) {
+    return next(e);
+  }
+
+
+  params.dateFrom = (params.dateFrom) ? new Date(params.dateFrom) : new Date();
+  let accountTransactions = []; // list of all account containing transactions
+  const accountIDs = [].concat(params.accountIds);
+
+  const accounts = await q.all(accountIDs.map(id => {
```
jniles: Nice use of `await`.  We could apply it to the entire function, but it isn't necessary here.

server/controllers/finance/reports/reportAccountsMultiple/report.handlebars
```diff
@@ -0,0 +1,129 @@
...
+          </th>
+
+          <th class="text-right">
+            {{currency header.credit ../metadata.enterprise.currency_id }}
+          </th>
+
+          <th></th>
+
+          <th class="text-right">
+            {{debcred header.exchangedBalance ../params.currency_id }}
```
jniles: For some reason, this doesn't reflect the same as the normal account statement in the "balance" column.  Can you check what is going on?

simpleaccountstatement
_This version works_

multipleaccountsbalance
_This version isn't correct_

jniles (commented):
Just a few more things to clean up.

server/controllers/finance/reports/reportAccountsMultiple/index.js
jeremielodi: No it's used in several places.

jniles: Ah, I see it now.  I wasn't expecting it to be used outside of it's context.

jniles (approved):

jniles (approved):
Good enough!

bors r+

Issue #3161, opened 2018-09-07
(Patient Invoice Module) Invalid invoice rows "Fall off the screen" on submission


sfount:
This issue can best be described with a demonstration: 

![fall](https://user-images.githubusercontent.com/2844572/45226719-91d1b880-b2b7-11e8-9cbd-a90ee51ff5cc.gif)

This behaviour shouldn't be possible in the `master` branch, it looks like wires are being crossed between the BHIMA `less` bootstrap build and the BHIMA `structure.css` when validation on the `ui-grid` rows is applied.

Unfortunately is is [impossible](https://github.com/IMA-WorldHealth/bhima-2.X/pull/364) to track down exactly [where](https://github.com/IMA-WorldHealth/bhima-2.X/pull/364/files#diff-50d9fdb1995374404c8adddf2b356a4dR658) or [when](https://github.com/IMA-WorldHealth/bhima-2.X/commit/2492afad647848af06e43f46f1f943c4350934d6) the possibility of this animation being introduced was [merged](https://github.com/IMA-WorldHealth/bhima-2.X/pull/370).

sfount:
Note that this issue has only very recently been observed with the change in build process https://github.com/IMA-WorldHealth/bhima-2.X/pull/3123

PR #3162, opened 2018-09-07
Fix invoice item validation CSS animation

sfount:
This commit removes the possibility of the `.hinge` animation being
applied in BHIMA. It is never intentionally used but seems to be applied
to ui-grid rows in a quirk of validation.

After investigating it looks like this quirk was introduced in the build process change on https://github.com/IMA-WorldHealth/bhima-2.X/pull/3123. It has been noted that glyphicons are not properly copied during this process, this style issue must be related but I cannot connect the two. 

The style deprecated here should be removed anyway but I will continue to investigate.

Expected patient invoice item validation behaviour: 
notfall


jniles:
Well... I thought this was a fluke of our end to end tests when I first saw it.  It's a shame we can't figure out what the root cause of it is.

Great work catching this and tracking it down.

bors r+

sfount:
Note that while this does fix the problem it looks like this problem was only introduced in https://github.com/IMA-WorldHealth/bhima-2.X/pull/3123. That means that fixing whatever is happening the build process will officially close https://github.com/IMA-WorldHealth/bhima-2.X/issues/3161. 

Then we could merge this as a refactor to prevent against this in the future!

bors r-

sfount:
bors test

jniles:
Do you mean "bors try"?

sfount:
:man_facepalming: I prefer to use `test` to disambiguate from trying the pull request for crimes. 

This time though, 
bors try

PR #3163, opened 2018-09-07
Fix build process: Glyphicon path issue and corrupt css minification

### LINKED ISSUE #3161: (Patient Invoice Module) Invalid invoice rows "Fall off the screen" on submission

sfount:
This commit resolves two issues that were introduced by changes in the
build process:
1. The bootstrap less font paths are relative, that means the
bhima-bootstrap file must be located relative to the `../fonts` folder.
This is moved.

2. Minifying the bootstrap CSS with out current minifier `cssnano` seems
to override some styles and animations causing issues on `ui-grid` row
validation. This has been reverted until it can be determined why this
minification causes this.

This pull request actually closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/3161

https://github.com/IMA-WorldHealth/bhima-2.X/pull/3162 should still be merged to ensure users are never accidentally met with an entire application shifting experience through a quirk of building CSS.

### DIFF
--- a/client/src/index.html
+++ b/client/src/index.html
@@ -10,7 +10,7 @@
 
   <!-- Compiled bootstrap style definition -->
   <link rel="stylesheet" href="vendor/angular-ui-select/dist/select.css">
-  <link rel="stylesheet" href="css/bootstrap/bhima-bootstrap.css">
+  <link rel="stylesheet" href="css/bhima-bootstrap.css">
   <link rel="stylesheet" href="vendor/angular-ui-grid/ui-grid.min.css">
   <link rel="stylesheet" href="vendor/components-font-awesome/css/font-awesome.min.css">
   <link rel="stylesheet" href="css/bhima/bhima.min.css">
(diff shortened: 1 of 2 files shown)


jniles:
We could always try a different minifier - I see to recall there being quite a few to choose from.

For example: [csso](https://www.npmjs.com/package/csso)

PR #3168, opened 2018-09-10
feat(weekend configuration) add help message

### LINKED ISSUE #3102: Weekend Configuration needs explanatory text
The configuration of weekends seems to be confusing for our users - with good reason.  There isn't any text explaining how to use the module.  We should add it so that our users do not forget how the system behaves.  See below:

weekendconfigurationconfusion
_FIg 1: Weekend Configuration_

The user is supposed to click the days _that are weekend days_ or days that are not worked during the week.  We need a text that reads "Mark only the days that are not worked.  Unchecked days will be considered as working days in the payroll calculation."

jeremielodi:
closes #3102

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -762,6 +762,7 @@
             "USER": "Select User",
             "VILLAGE": "Select a Village/Township",
             "WEEKEND_CONFIGURATION": "Select a weekend Configuration",
+            "WEEKEND_DAYS" : "Select the day(s) of the weekend, (example : saturday and sunday) ",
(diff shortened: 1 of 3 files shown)


jniles:
Can you include a screenshot since this is just a UI change?

jeremielodi:
weekend

jniles:
Thanks!  Instead of putting the text in the modal-header, could you put it in the modal-body?  That way the modal looks like most other modals in the application.

jniles:
This is what it looks like now:
weekendconfigurationconfusionresolved

bors r+

PR #3170, opened 2018-09-10
chore: collapse migrations scripts into one

jniles:
This commit collapses all migration scripts into a single migration script, making it easier to apply them in bulk.


sfount (approved):
Good clean up! I haven't verified through PRs that all of these upgrades happened within this database version but I'm confident that you would know.

jniles:
I've tested it on the HEV dataset, and it seems to work.

bors r+

PR #3171, opened 2018-09-10
fix(user): add clarity to user fields

### LINKED ISSUE #3119: Proposal: rename user creation fields to match data model
The user creation modal currently has fields for **'username'** and **'login'**. Even as someone who is familiar with the web and computers I am not sure how to differentiate these two fields. 

screenshot from 2018-09-01 21-30-22

These inputs correspond with database values: 
* Login - `username`
* Username - `display_name` 

I propose that everywhere we show user information with a field (this create modal and the users table for example) we use 'Username' and 'Display Name' labels. 

It may even be useful to put a short text description like

Username: 
_This is the user's identifier for the system, it should be short and memorable and will be used every time they log in._ 

Display Name:
_The user's full name, this is displayed on any formal documents or receipts processed by the system._

jniles:
This commit adds clarifying text to the fields on the user creation page.

Closes #3119


client/src/modules/users/user.modal.html
```diff
@@ -16,35 +16,38 @@
 
   <div class="modal-body" style="overflow-y: scroll; max-height:600px; ">
 
-    <div class="form-group has-feedback" ng-class="{ 'has-error' : (UserForm.$submitted || UserForm.display_name.$error.unique) && UserForm.display_name.$invalid }">
-      <label class="control-label" translate>FORM.LABELS.USERNAME</label>
+    <div class="form-group" ng-class="{ 'has-error' : UserForm.$submitted && UserForm.display_name.$invalid }">
+      <label class="control-label" translate>FORM.LABELS.DISPLAY_NAME</label>
       <input
         name="display_name"
-        bh-unique="/users/"
```
sfount, line 23: Good catch on the unique field too!

sfount (approved):
This addresses all of the points made in the issue and also catches `bh-unique` on the wrong field, great work!

PR #3174, opened 2018-09-12
Import stock from a csv file

mbayopanda:
This PR a the option of importing stock into a depot from a csv file.


mbayopanda:
@jniles

client/src/i18n/en/stock.json
```diff
@@ -57,6 +57,15 @@
     "INVENTORY_GROUP" : "Inventory Group",
     "INVENTORY_NOT_AVAILABLE" : "Inventories not available",
     "PRODUCT_NOT_AVAILABLE" : "Products not available",
+    "IMPORT" : {
+      "STOCK_FROM_FILE" : "Import stock from a file",
+      "ABOUT_DESCRIPTION" : "You are about to import the stocks for the following deposit",
```
jniles: Small typo: `deposit` -> `depot`.

client/src/modules/stock/import/stockImport.js
```diff
@@ -0,0 +1,108 @@
...
+  /** upload the file to server */
+  function uploadFile(file) {
+    vm.uploadState = 'uploading';
+
+    const parameters = {
+      url : '/stock/import/',
+      data : { file, depot_uuid : vm.depot.uuid },
+    };
+
+    const filters = [
```
jniles: Could you move these `filters` to outside the function?  Since they never change, there is no reason to have them in the `uploadFile()`.

client/src/modules/stock/import/stockImport.js
```diff
@@ -0,0 +1,108 @@
...
+];
+
+/**
+ * @class StockImportController
+ *
+ * @description
+ * This module helps to import stock from a file
+ */
+function StockImportController(
+  Depots, Inventory, Notify, Session, util, bhConstants, ReceiptModal, StockForm,
```
jniles: It looks like there is a lot of imports here that you never use.  Can you clean this up?

server/controllers/inventory/import/index.js
```diff
@@ -34,7 +34,9 @@ function downloadTemplate(req, res, next) {
  */
 function importInventories(req, res, next) {
   if (!req.files || req.files.length === 0) {
-    next(new BadRequest('Something broke', 'ERRORS.EVERYTHING_BAD'));
+    const errorDescription = 'Expected at least one file upload but did not receive any files.';
```
jniles, line 37: This kind of error handling looks like it would be best handled by a middleware.  Could we change the `upload.middleware()` function, or create our own middleware, that simply throws an error if a route does not have `req.files` set?

server/controllers/stock/core.js
```diff
@@ -496,6 +496,7 @@ function processMultipleLots(inventories) {
       lot.S_LOT_LIFETIME = lot.lifetime - lotLifetime;
       lot.S_RISK = lot.S_LOT_LIFETIME - lot.S_MONTH;
       lot.S_RISK_QUANTITY = lot.S_RISK * lot.avg_consumption;
+      lot.S_RISK_QUANTITY = Math.round(lot.S_RISK_QUANTITY);
```
jniles: It's slightly confusing to set `lot.S_RISK_QUANTITY` twice.  Could you just wrap the first line with`Math.round()`?

server/controllers/stock/import.js
```diff
@@ -0,0 +1,117 @@
...
+    next(errorDetails);
+    return;
+  }
+
+  let query;
+  let queryParams;
+
+  const filePath = req.files[0].path;
+  const depotUuid = db.bid(req.body.depot_uuid);
+  const documentUuid = db.bid(uuid.v4());
```
jniles: You use use [util.uuid()](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/util.js#L248) to generate your UUIDs.

server/controllers/stock/import.js
```diff
@@ -0,0 +1,117 @@
...
+    })
+    .then(data => {
+      if (!hasValidDataFormat(data)) {
+        throw new BadRequest('The given file has a bad data format for stock', 'ERRORS.BAD_DATA_FORMAT');
+      }
+
+      const transaction = db.transaction();
+
+      data.forEach(item => {
+        query = 'CALL ImportStock(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);';
```
jniles: This query never changes - you could move it out of the `data.forEach()` loop.

server/controllers/stock/import.js
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @description check if data has a valid format for stock
+ *
+ * @param {object} data
+ */
+function hasValidDataFormat(data = []) {
+  return data.every(item => {
+    return item.inventory_group_name
+      && item.inventory_text && item.inventory_type && item.inventory_unit
+      && item.inventory_unit_price && item.stock_lot_label
```
jniles: If `inventory_unit_price` is 0 ... is that allowed?

server/models/bhima.sql
```diff
@@ -98,7 +98,8 @@ INSERT INTO unit VALUES
   (204, 'Exchange Rate','TREE.EXCHANGE','',1,'/modules/exchange/exchange','/exchange'),
   (205, 'Account Reference Management','TREE.ACCOUNT_REFERENCE_MANAGEMENT','',1,'/modules/account_reference','/account_reference'),
   (206, '[OHADA] Bilan','TREE.OHADA_BALANCE_SHEET','',144,'/modules/reports/ohada_balance_sheet_report','/reports/ohada_balance_sheet_report'),
-  (207, 'Account Reference Report','TREE.ACCOUNT_REFERENCE_REPORT','',144,'/modules/reports/account_reference','/reports/account_reference');
+  (207, 'Account Reference Report','TREE.ACCOUNT_REFERENCE_REPORT','',144,'/modules/reports/account_reference','/reports/account_reference'),
+  (208, 'Import Stock From File','TREE.IMPORT_STOCK_FROM_FILE','',160,'/modules/stock/import','/stock/import');
```
jniles, line 102: Could you add a script to the [`migrate.sql`](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/migrations/v0.8.0-v0.9.0/migrate.sql) file to make these changes for us?

server/controllers/stock/import.js
```diff
@@ -0,0 +1,117 @@
...
+          item.inventory_text,
+          item.inventory_type,
+          item.inventory_unit,
+          item.inventory_unit_price,
+          item.inventory_cmm || 0,
+          item.stock_lot_label,
+          item.stock_lot_quantity,
+          moment(item.stock_lot_expiration).format('YYYY-MM-DD'),
+        ];
+        transaction.addQuery(query, queryParams);
```
jniles, line 74: I just thought of this - but most of our other modules that write multiple lines in the database use `Staging` tables ... such as [invoicing](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/models/procedures/invoicing.sql#L15), cash, and vouchers.  In theory, this gives us both a performance improvement (since we can import all the data at one time) and better error handling (since we can say that inventory XYZ doesn't exist before we even start to import the data).

It would be worth writing a second version of this function that uses temporary tables to compare with this version.  That doesn't need to be done for this PR, but we should do some performance analysis just to see.

jniles (commented):
@mbayopanda just made a few notes here.  I haven't completed a full review, but will do so later this evening after my meetings.

server/controllers/stock/import.js
mbayopanda: Zero must be allowed, so i think we have to handle that for allowing zero or to just be sure that `item.inventory_unit_price` is not undefined

mbayopanda:

server/controllers/stock/reports/stock_lots.report.handlebars
```diff
@@ -46,8 +46,14 @@
             <th>{{translate 'STOCK.QUANTITY'}}</th>
             <th>{{translate 'STOCK.INVENTORY_UNIT'}}</th>
             <th>{{translate 'STOCK.ENTRY_DATE'}}</th>
-            <th>{{translate 'STOCK.EXPIRATION_DATE'}}</th>
+            <th>{{translate 'STOCK.EXPIRATION_DATE'}}</th>`
```
jeremielodi: Is this cote important, if not, go head and remove it

jeremielodi (commented):
Everything looks correct!

PR #3175, opened 2018-09-12
feat(transaction type) translate grouping type

### LINKED ISSUE #3153: `bh-transaction-type-select` uses hard-coded category values for headers
Our system transaction type categories are currently split into 3 options, income, expense and other. These are identified by and ID:String match in the database. 

The Transaction Type select conveniently shows these 3 types separately however it uses the string directly from the database to display the header, this means that the values aren't translated and don't respect the language of the user.

This screenshot from the simple vouchers shows this in action: 
screenshot from 2018-09-06 17-48-54

Here the system language is french but the headers are still 'income', 'expense' and 'other'. 

This should be updated to use the `bhConstants` transaction type map to convert these database values into translatable keys. See https://github.com/IMA-WorldHealth/bhima-2.X/pull/3106 for an example implementation of this process.

jeremielodi:
closes #3153

### DIFF
--- a/client/src/modules/vouchers/complex-voucher.ctrl.js
+++ b/client/src/modules/vouchers/complex-voucher.ctrl.js
@@ -26,6 +26,9 @@ function ComplexJournalVoucherController(
 ) {
   const vm = this;
 
+  // help to group transaction_types
+  vm.groupTransactionByType = Vouchers.groupTransactionByType;
+
   // bind constants
   vm.bhConstants = bhConstants;
   vm.enterprise = Session.enterprise;
--- a/client/src/modules/vouchers/complex-voucher.html
+++ b/client/src/modules/vouchers/complex-voucher.html
@@ -117,7 +117,7 @@
                     <ui-select-choices
                       ui-select-focus-patch
                       repeat="item.id as item in ComplexVoucherCtrl.types | filter: { hrText : $select.search }"
-                      group-by="'type'">
+                      group-by= "ComplexVoucherCtrl.groupTransactionByType">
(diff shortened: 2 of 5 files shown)


sfount (commented):
This looks like it correctly addresses the issue referenced! It would be good if the type translate label could be taken from the shared application constants. 

I haven't checked if the `$translate.instant` causes any inconsistencies if the page loads slowly but given this group function will be called whenever Angular calls `$digest` this shouldn't be visible to the user.

client/src/modules/vouchers/complex-voucher.ctrl.js
```diff
@@ -323,5 +323,11 @@ function ComplexJournalVoucherController(
       .catch(Notify.handleError);
   }
 
+  // help to group transaction_types
+  vm.groupByType = (item) => {
+    const type = `FORM.LABELS.${item.type.toUpperCase()}`;
```
sfount: A shared definition for this mapping can be found in `bhConstans` here: 
https://github.com/IMA-WorldHealth/bhima-2.X/blob/c94a1b4dbd1dac5d476b364c60f3d361dba906ae/client/src/js/constants/bhConstants.js#L150

Using this would mean that if the JSON translation structure or the database changes we only have to update the application in one place. 

It can be used by requiring the constants library: 
```js
Controller.$inject = ['bhConstants'];

function Controller(bhConstants) { 
  const type = bhConstants.transactionTypeMap[item.type];
}
```

jeremielodi: okay

jniles (approved):
Looks good to me!

bors r+

PR #3177, opened 2018-09-16
PR for issue #3169

### LINKED ISSUE #3169: Proposal: switch enterprise_settings lookups to a "SELECT *" statement
Normally I'm opposed to `SELECT * FROM table;` statements since they are imprecise.  However, every time we add a new enterprise setting in development, we tend to break our client datasets.  For example, this just happened to me while I was trying to run an HEV dataset on `master`:
```sh
  errors #interceptor() { Error: ER_BAD_FIELD_ERROR: Unknown column 'enable_auto_stock_accounting' in 'field list'

```

While this isn't a huge issue, we could avoid it easily if we just had `SELECT * FROM enterprise_settings` instead of naming each column by name.

Scholldier:
This PR is to resolve issue #3169.

@jniles did I fix the right section of code? Did I miss any select statements?

I used the following command to search for them:

grep -r 'from enterprise_setting' * -i

Resolves #3169


jniles:
Hey @Scholldier, this looks right!  However, it looks like you last branched off of master a long time ago (in BHIMA world, approx two weeks ago).  This has led to a few collision in trying to merge.

Since this is rather a small change, I propose you re-branch off of BHIMA's upstream master and reintroduce the changes to fix these collision.

But yeah, this looks like it!

jniles:
@Scholldier what is the status of this PR?  Are you able to work on it, or shall I have another member of my team get it working?

PR #3181, opened 2018-09-18
fix async validation in bhFindInvoice

### LINKED ISSUE #3145: Stock Exit to patient does not handle asynchronous validation well
On slow network connections, the Stock Exit to patient will replace any user-set patient with a cached one surprisingly.  Additionally, the validation for the invoice defaults to an error while the $http request is still pending.  See below:

mbayopanda:
This PR close the issue #3145 

Close #3145


jniles:
Looks good to me!

bors r+

PR #3194, opened 2018-09-24
Enhance stock lots exit module

### LINKED ISSUE #3188: fix(stock exit): load just articles and lots which have stock available for the given date
To fix the problem of exit stock before entry in the stock file report, we must calculate for each defined date in the stock exit module, the stock available according the date and load just inventory which are available for the given date.

If there is no stock available for the given date shows a message with an information like "There is no stock available for the date DD-MM-YYYY".

### LINKED ISSUE #3189: Fix(stock exit) : Allow only one lot in the grid of stock exit
If a lot is already selected, forbid to select the same lot again in another line of the grid for the stock exit.

For avoiding things like : 
image

In the picture above, we have two `amoxycilline 250 mg` exited by accident

mbayopanda:
This PR : 
- Prevents more than one lot from multiple lines
- Prevents submission of duplicated lots
- Fix custom date filter at the server side
- Warn the user about the use of an anterior date
- Fix the export to csv function

Close #3188 
Close #3189

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -141,7 +141,7 @@
     "SUCCESS"         : "Successfully created a movement",
     "TO"              : "Destination",
     "UNIT_COST"       : "Unit cost",
-    "ANTERIOR_DATE_QUANTITY_MESSAGE" : "You have chosen a previous date, only the lots available on the chosen date will be visible as well as the quantity on this date"
+    "ANTERIOR_DATE_QUANTITY_MESSAGE" : "You have chosen a past date, only the lots available on the chosen date will be visible as well as the quantity on this date"
   },
   "SETTINGS" : {
     "ORDER_BY_CREATED_AT" : "Sort lots movements by their true creation date",
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -140,7 +140,8 @@
     },
     "SUCCESS"         : "Successfully created a movement",
     "TO"              : "Destination",
-    "UNIT_COST"       : "Unit cost"
+    "UNIT_COST"       : "Unit cost",
+    "ANTERIOR_DATE_QUANTITY_MESSAGE" : "You have chosen a previous date, only the lots available on the chosen date will be visible as well as the quantity on this date"
   },
   "SETTINGS" : {
     "ORDER_BY_CREATED_AT" : "Sort lots movements by their true creation date",
--- a/client/src/js/services/grid/GridExport.js
+++ b/client/src/js/services/grid/GridExport.js
@@ -69,7 +69,7 @@ function GridExportService(Modal, util, bhConstants, uiGridExporterService, mome
    * @param {function} rowsFormatterFn - [optional] callback fn to apply to rows
    * @param {function} columnsFormatterFn - [optional] callback fn to apply to columns
    */
-  GridExport.prototype.exportToCsv = (filename, columnsFormatterFn, rowsFormatterFn) => {
+  GridExport.prototype.exportToCsv = function exportToCsv(filename, columnsFormatterFn, rowsFormatterFn) {
     let columns = this.options.columnDefs || [];
     let rows = this.options.data || [];
 
--- a/client/src/modules/stock/exit/exit.js
+++ b/client/src/modules/stock/exit/exit.js
@@ -25,6 +25,7 @@ function StockExitController(
   vm.stockForm = new StockForm('StockExit');
   vm.movement = {};
   vm.gridApi = {};
+  vm.selectedLots = [];
   vm.reset = reset;
 
   vm.onDateChange = date => {
@@ -44,6 +45,7 @@ function StockExitController(
   vm.submit = submit;
   vm.changeDepot = changeDepot;
   vm.checkValidity = checkValidity;
+  vm.onLotSelect = onLotSelect;
 
   const mapExit = {
     patient : { description : 'STOCK.EXIT_PATIENT', find : findPatient, submit : submitPatient },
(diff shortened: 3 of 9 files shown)


client/src/i18n/en/stock.json
```diff
@@ -140,7 +140,8 @@
     },
     "SUCCESS"         : "Successfully created a movement",
     "TO"              : "Destination",
-    "UNIT_COST"       : "Unit cost"
+    "UNIT_COST"       : "Unit cost",
+    "ANTERIOR_DATE_QUANTITY_MESSAGE" : "You have chosen a previous date, only the lots available on the chosen date will be visible as well as the quantity on this date"
```
jniles: Instead of "previous date", it makes more sense in English to say "an old date" or "a past date".    I propose you change this to say: "You have chosen a past date.  Only lots available on the chosen ...." (the `...` should be the rest of the statement you have written).

client/src/js/services/grid/GridExport.js
```diff
@@ -69,7 +69,7 @@ function GridExportService(Modal, util, bhConstants, uiGridExporterService, mome
    * @param {function} rowsFormatterFn - [optional] callback fn to apply to rows
    * @param {function} columnsFormatterFn - [optional] callback fn to apply to columns
    */
-  GridExport.prototype.exportToCsv = (filename, columnsFormatterFn, rowsFormatterFn) => {
+  GridExport.prototype.exportToCsv = function exportToCsv(filename, columnsFormatterFn, rowsFormatterFn) {
```
jniles, line 72: Aha!  This is an important change. 🐛

client/src/modules/stock/exit/exit.js
```diff
@@ -299,6 +316,46 @@ function StockExitController(
...
+      .filter(item => item.lot && item.lot.uuid)
+      .map(item => item.lot.uuid);
+  }
+
+  // detect the presence of duplicated lots
+  function hasDuplicatedLots() {
+    refreshSelectedLotsList();
+
+    let doublonDetected = false;
+    for (let i = 0; i < vm.selectedLots.length; i++) {
```
jniles: It looks like `vm.selectedLots` is an array of `uuid`s.  That means, if you check the first value, you never have to check it again.  You can increase the performance by making `j` start at `i + 1`.  Like this:

```js
for (let i = 0; i < vm.selectedLots.length; i++) {
 let found = 0;
  for (let j = i + 1; j < vm.selectedLots.length; j++) {
    if (lot === vm.selectedLots[j]) {
      found++;
      if (found > 1) { break; }
    }
   // etc..
  }
}
```

I still wish we had solved this [with a datastructure though](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2544#issuecomment-366499969).  It would have been easier to read and more performant.

client/src/modules/stock/exit/exit.js
```diff
@@ -299,6 +316,46 @@ function StockExitController(
...
+      let found = 0;
```
jniles: As an alternative to the above suggestion, you can also make this easier to read, but maybe at a loss of performance.  It uses [lastIndexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/lastIndexOf):

```js
let doubleIndex;
const doublonDetected = vm.selectedLots.some((lot, idx) => {
  const hasDoubles = vm.selectedLots.lastIndexOf(lot) !== idx;
  if (hasDoubles) { doubleIndex = idx; }
  return hasDoubles;
});
```

jniles: 6 lines instead of 18!

client/src/modules/stock/exit/exit.js
```diff
@@ -412,19 +469,36 @@ function StockExitController(
...
     return mapExit[vm.movement.exit_type].submit()
       .then(toggleLoadingIndicator)
       .catch(Notify.handleError)
       .finally(() => reinit(form));
   }
 
+  function errorLineHighlight() {
+    // set and unset error flag for allowing to highlight again the row
+    // when the user click again on the submit button
+    vm.stockForm.store.data[vm.doublonDetectedLine][vm.ROW_ERROR_FLAG] = true;
```
jniles: `vm.doubleDetectedLine` is a global variable and if we ever change the name, we will need to update this function twice!

In general, [global variables are](https://stackoverflow.com/questions/1841916/how-to-avoid-global-variables-in-javascript) [considered](http://wiki.c2.com/?GlobalVariablesAreBad) [bad practice](https://gist.github.com/hallettj/64478).

You can avoid using the global variable here by just passing it in as a function parameter:
```js
function errorLineHighlight(rowIdx) {
  const row = vm.stockForm.store.data[rowIdx];

  row[vm.ROW_ERROR_FLAG] = true;
  $timeout(() => {
    row[vm.ROW_ERROR_FLAG] = false;
   }, 1000);
}
```

Now you can just do:
```js
if (hasDuplicatedLots()) {
  errorLineHighlight(vm.doublonDetectedLine);
  // etc..
}
```

client/src/modules/stock/exit/exit.js
```diff
@@ -412,19 +469,36 @@ function StockExitController(
     if (!vm.movement.entity.uuid && vm.movement.entity.type !== 'loss') {
       return Notify.danger('ERRORS.ER_NO_STOCK_DESTINATION');
     }
+
+    if (hasDuplicatedLots()) {
+      // notify on the concerned row
+      errorLineHighlight();
+      return Notify.danger('ERRORS.ER_DUPLICATED_LOT', 20000);
```
jniles, line 463: Why is this such a long notification? 20 seconds?

client/src/modules/stock/exit/exit.js
```diff
@@ -231,7 +238,14 @@ function StockExitController(
   // remove item
   function removeItem(item) {
     vm.stockForm.removeItem(item.id);
+    if (item.lot && item.lot.uuid) {
```
jniles: Why do you manually remove the selected lot and then call `refreshSelectedLotsList()`?  Shouldn't you only have to do one?

client/src/modules/stock/exit/exit.js
mbayopanda: I would like to show the message for a long time to be sure that the user read that and he can close the message himself

jniles (approved):
Looks good to me.

bors r+

PR #3197, opened 2018-09-25
Extend bhInventorySelect with consumable filter

### LINKED ISSUE #3186: fix(stock file) : display only inventory which can be in stock
The stock file report displays also inventory which are not article in stock (not consumable) in the inventory selection list.

image

mbayopanda:
This PR add the attribute `only-consumable` to the `bh-inventory-select` component so that we can get or not only inventory which are consumable.

Close #3186


jniles (approved):
Seems fine to me.

Bors r+

PR #3198, opened 2018-09-26
Remove custom date filter for 'allTime' period selected

### LINKED ISSUE #3195: Setting the registry filters to "allTime" does not remove a custom date filter
See below:
filteralltimedoesnotremovecustomfilter

The issue is that we have set "allTime", but that didn't remove the previously customized filter.

mbayopanda:
This PR fix the issue #3195 by removing `custom_period_start` and `custom_period_end` filters when the filter period with value `allTime` is selected.

Close #3195


jniles (approved):
Looks good to me.

bors r+

PR #3205, opened 2018-09-28
Quick insert lots

### LINKED ISSUE #3204: Fast stock lots insert
This issue is about to add checkbox in the lot modal for helping the user inserting quickly lot label with a codebar scanner for example

mbayopanda:
This PR fix the issue #3204 by adding fast lots insertion feature with a simple enable/disable checkbox

Close #3204


.env.development
```diff
@@ -5,7 +5,7 @@ PORT=8080
 DB_HOST='localhost'
 DB_USER='bhima'
 DB_PASS='HISCongo2013'
-DB_NAME='bhima_test'
+DB_NAME='imck'
```
jniles: This should be switched back to `bhima_test`

client/src/i18n/en/stock.json
```diff
@@ -164,7 +164,9 @@
     "TO_ADJUSTMENT"    : "Adjustment (Negative)"
   },
   "LOTS" : {
-    "REGISTRY" : "Lots Registry"
+    "REGISTRY" : "Lots Registry",
+    "ENABLE_FAST_INSERT": "Enable fast lots insertions",
+    "ENABLE_FAST_INSERT_DESCRIPTION": "This option allows you to quickly insert lots one after the other. It can be useful if you use barcode reader"
```
jniles: Instead of describing its purpose, let's describe what it does: `This option will automatically select the next blank row's serial/lot number in the entry table.  If necessary, it adds a new row.  Useful for quickly inserting with a barcode reader.`

client/src/js/directives/bhBlur.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * This directive helps to give trigger an action() when the Enter or Tab key is pressed
+ *
+ * <input type="text" bh-blur="action()">
+ */
+angular.module('bhima.directives')
+  .directive('bhBlur', () => {
+    return (scope, elem, attrs) => {
+      elem.bind('keydown', (event) => {
```
jniles: I'm concerned this might have a memory leak since the event might never be cleaned up - see [this long SO discussion](https://stackoverflow.com/questions/26983696/angularjs-does-destroy-remove-event-listeners) about the different ways to remove events in AngularJS.

Did you try with `element.on('keydown', () => ...)`?  If it works, that is probably the best solution, because then you can use `element.off()` to remove the event when the scope is destroyed.

client/src/js/directives/bhBlur.js
```diff
@@ -0,0 +1,19 @@
...
+        // 13 represents enter button and 9 represents tab button
+        if (event.which === 13 || event.which === 9) {
```
jniles: Could you make these constants to improve readability?  Like this:
```js
const ENTER_KEY = 13;
const TAB_KEY = 9;

// later...
if (event.which === ENTER_KEY) { // etc... }
```

client/src/js/directives/bhFocus.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * This directive helps to give focus to an element when an expression is true
+ *
+ * <input type="text" bh-focus="true">
+ */
+angular.module('bhima.directives')
```
jniles: I also had to figure out how to set the focus when I was working on the barcode input.  [This SO article helped me a lot](https://stackoverflow.com/questions/25596399/set-element-focus-in-angular-way).  Basically, just create a service that will focus on the element for you based on the ID.  We might have one already - you'll have to investigate.

client/src/modules/stock/entry/modals/lots.modal.html
```diff
@@ -79,8 +85,8 @@
       FORM.BUTTONS.CLOSE
     </button>
 
-    <bh-loading-button loading-state="FindForm.$loading">
-      <span translate>FORM.BUTTONS.SUBMIT</span>
-    </bh-loading-button>
+    <button type="submit" class="btn btn-primary" ng-click="$ctrl.submit(FindForm, $event)" data-method="submit" translate>
```
jniles, line 95: Hmm... this removes the loading component of it.  Should we be loading here?  Is there a possibility that this action is asynchronous?

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -93,6 +94,30 @@ function StockDefineLotsModalController(Instance, Notify, uiGridConstants, Data,
...
+   * @description
+   * if the fast insert option is enable do this :
+   * - add new row automatically on blur
+   * - set the focus in the new row
+   * @param {string} rowLot the row.entity.lot string
+   */
+  function onLotBlur(rowLot) {
+    if (vm.enableFastInsert && rowLot) {
+      // remove focus on previous rows
+      vm.form.rows.forEach(lot => {
```
jniles: Yeah, I think using a focus service is probably your best option here.  It would reduce the code to:
```js
const item = vm.form.addItem();
focus(item.uuid)
```
Or something very similar.

test/end-to-end/stock/stock.entry.page.js
```diff
@@ -137,6 +134,38 @@ function StockEntryPage() {
...
+  /**
+   * fast insert lots rows
+   * @param {array} lots an array of strings
+   */
+  page.fastLotsInsert = (lots = []) => {
+    lots.forEach((lot, index) => {
+      const lotCell = GU.getCell(lotGridId, index, 1);
+      const input = FU.input('row.entity.lot', lot, lotCell);
+
+      input.sendKeys(protractor.Key.TAB);
```
jniles, line 161: Nice 👍

jniles (commented):
I've made a few comments.

You've added two directives in this PR.  If you transform `bhFocus` into a service, then it will only be one.  Directives should be unit tested.  Thankfully, this one is really easy and simple to verify that it works.  All you need to do is create an element, blur it, and check that the function is called.

The functionality is cool and seems to work fairly well to me.  It always focuses the last row - should it instead focus the next empty row?  Would that make more sense?

client/src/modules/stock/entry/modals/lots.modal.html
mbayopanda: No... the action here is not asynchronous, the submit function return an array of json to the exit stock module, there is no request to server in this modal.

Also, the main reason why i remove the `bh-loading-button` was when i define one lot and type Enter the form was submitted automatically.

jniles: Alright, then no problem.  I wonder why the 'submit' even was called on the whole form when the enter key is detected...

mbayopanda:
@jniles

jniles:
@mbayopanda,

Code is much improved!  Two things to work on:
 1. This PR needs unit tests for both directives/services you created.  These should be made with Karma and have a couple of test cases for each.
 2. It would be nice if the "quick insert" jumped to the next empty row instead of always creating a new row.  Do you think you could work on that?

mbayopanda:
@jniles

test/client-unit/directives/bhBlur.spec.js
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  it('should not call a given method on blur for other key apart from Tab and Enter', () => {
+    const input = find(element, '#row');
+    for (let i = 0; i < 128; i++) {
+      if (i === ENTER_KEY || i === TAB_KEY) { continue; }
+
+      triggerKeyDown(angular.element(input), i);
+      $scope.$digest();
+      expect($scope.models.callback).to.not.have.been.called();
```
jniles, line 64: You could probably move this out of the loop and only check it at the end.  If it is never called in the loop, it shouldn't be called outside the loop.

test/client-unit/directives/bhBlur.spec.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+  });
+
+  it('should call a given method with a given parameter on blur by Enter key', () => {
```
jniles, line 68: Hey, this is a really nice test 👍

test/client-unit/directives/bhBlur.spec.js
```diff
@@ -0,0 +1,75 @@
...
+  let element;
+
+  const ENTER_KEY = 13;
+  const TAB_KEY = 9;
+
+  // utility function
+  const find = (elm, selector) => elm[0].querySelector(selector);
+
+  const triggerKeyDown = (_element, _keyCode) => {
+    const e = angular.element.Event('keydown');
```
jniles, line 16: Interesting.  I had no idea this was a thing 👍

test/client-unit/services/Focus.spec.js
```diff
@@ -0,0 +1,65 @@
+/* global inject, expect, chai */
+describe('Focus', () => {
+
+  let $scope;
+  let $compile;
```
jniles, line 5: Just FYI, since you only use `$compile` in the `beforeEach()` block, you don't need to make it a variable.  Just import it in the `beforeEach()` and use it there.

jniles (approved):
@mbayopanda, well done!  These are some of the best tests in the repository at the moment.  They are clearly laid out, they test all the important logic, and are a pleasure to read.  Fantastic work!

My only comment is that `bhBlur` doesn't actually have anything to do with `blur`.  It only has to do with hitting the ENTER or TAB keys.  It might be better to break this into on to something like `on-enter-key` and `on-tab-key`.  Or even something more generic like `bh-keypress="callback"  bh-keypress-keys="ENTER_KEY,TAB_KEY"`.

Everything looks good here though.  Well done!

bors r+

PR #3208, opened 2018-10-01
fix(fiscal): fix fiscal year loading error

### LINKED ISSUE #3207: Fiscal year shows "no fiscal years found" while it is loading
On a slow connection, the Fiscal Year page shows "no fiscal years found", even while it is loading.  See below:

fiscalyearerror
_Fig 1: Confusing Fiscal Year Message_

This is confusing.  We should only show the message after the module has finished loading and the values are all settled.

jniles:
This commit hides the "no fiscal years found" error case until it is certain that no fiscal years are actually located in the database.

Closes #3207

### DIFF
--- a/client/src/modules/fiscal/fiscal.list.html
+++ b/client/src/modules/fiscal/fiscal.list.html
@@ -56,7 +56,7 @@
           <span class="fa fa-circle-o-notch fa-spin"></span>
           <span translate>TABLE.COLUMNS.LOADING</span>
         </p>
-        <p ng-show="(FiscalCtrl.fiscalYears.length == 0) && !FiscalCtrl.loadingError"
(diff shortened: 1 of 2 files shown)


jniles:

client/src/modules/fiscal/fiscal.list.html
```diff
@@ -56,7 +56,7 @@ <h4 style="margin-top : 0px;">
           <span class="fa fa-circle-o-notch fa-spin"></span>
           <span translate>TABLE.COLUMNS.LOADING</span>
         </p>
-        <p ng-show="(FiscalCtrl.fiscalYears.length == 0) && !FiscalCtrl.loadingError"
+        <p ng-show="(FiscalCtrl.fiscalYears.length == 0) && !FiscalCtrl.loadingError && !FiscalCtrl.loadingState"
```
jeremielodi, line 59: 👍

jeremielodi (commented):
Everything looks correct, well done!

PR #3214, opened 2018-10-02
chore(inventory): remove dead code

jniles:
This commit removes all the dead code from the inventory/stock routes.  Many of these are no longer used since stock was re-written.

I'll keep pruning until tests start failing 👍


jniles:
Tests pass!  Looks like none of this code was ever used.

bors r+

Issue #2652, opened 2018-04-04
Write unit tests for the bhFilterToggle component


jniles:
We haven't used it as much as we should, but it has the potential to make many modules much easier to maintain: [bh-filter-toggle](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/components/bhFilterToggle.js).  This component basically just calls a callback on toggle.

The following things needs to be tested:
 1. By default, the filter should not have the class `btn-info`.
 2. If the controller has `filterEnabled` set, the button should have the class `btn-info`.
 3. Clicking the button should call the `toggle()` function on the controller.
 4. If `filterEnabled = true`, calling the `toggle()` function on the controller should set the `filterEnabled` value to false.
 5. Calling the `toggle()` function should call the `onToggleCallback()` function.

PR #3216, opened 2018-10-02
test(bhFilterToggle): write client-unit test

jniles:
This commit adds a client unit tests for the bhFilterToggle to ensure the onToggle() callback is always fired.  It also tightens up the code slightly by adding doc comments and moving all callbacks to the component border.

Closes https://github.com/IMA-WorldHealth/bhima-2.X/issues/2652

### DIFF
--- a/client/src/js/components/bhFilterToggle.js
+++ b/client/src/js/components/bhFilterToggle.js
@@ -1,33 +1,35 @@
 angular.module('bhima.components')
   .component('bhFilterToggle', {
-    templateUrl: '/modules/templates/bhFilterToggle.tmpl.html',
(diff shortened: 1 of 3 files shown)


jeremielodi (commented):
@jniles your code for this PR looks good so it can be merged
I have a question, do you think the filter action on the ui-grid is implemented differently in the controllers?
If yes, having a callback function make sense else we will have less codes if we include the unique logic in this component

jniles:
@jeremielodi, you are right, we could incorporate the ui-grid options and save a callback.  In the long run, we might want to do this.

bors r+

PR #3218, opened 2018-10-02
fix: add index on reference_uuid

jniles:
Welp.  We indexed `record_uuid` twice instead of `reference_uuid` once.  Speeds up most reporting queries that use it.

@sfount, guess we both rushed this one! 😞


jniles:
It's a fluke, but I forgot to add the migrate.sql changes.  I'll do that.

PR #3227, opened 2018-10-04
feat(report) rename archive file

### LINKED ISSUE #3226: Report file doesn't have a readable name
I notice that In the Report module, when a use downloads a report he has archived, the file name is really unreadable. Normally it should be the same as the archive label.

jeremielodi:
closes #3226


jniles:
Looks good to me!  Great find and fix.

bors r+

PR #3238, opened 2018-10-09
feat(voucher receipt) signal if it's canceled

### LINKED ISSUE #3225: Vouchers do not signal that they are canceled on the receipt
Unlike Cash Payments and Invoices, the Vouchers do not have any identifier on their receipt.  For example, a canceled invoice looks like this:
chrome_2018-10-04_11-38-32
_Fig 1: A reversed invoice clearly showing the reversal status_

Whereas, a reversed voucher doesn't show it:

_Gif 1: No reversal status on the voucher_

It would be ideal if both these documents showed a "canceled" status on them.  The only current way to see that a voucher is canceled is to check in the voucher registry.

jeremielodi:
closes #3225

### DIFF
--- a/server/controllers/finance/reports/vouchers/index.js
+++ b/server/controllers/finance/reports/vouchers/index.js
@@ -12,7 +12,7 @@ const shared = require('../shared');
 const ReportManager = require('../../../../lib/ReportManager');
 const Vouchers = require('../../vouchers');
 const pdf = require('../../../../lib/renderers/pdf');
-
+const db = require('../../../../lib/db');
 // dependencies for barcode translation
 const barcode = require('../../../../lib/barcode');
 const identifiers = require('../../../../config/identifiers');
@@ -59,6 +59,7 @@ function receipt(req, res, next) {
 
   return Vouchers.lookupVoucher(req.params.uuid)
     .then((voucher) => {
+      voucher.isCreditNoted = voucher.reversed === 1;
       voucher.barcode = barcode.generate(entityIdentifier, voucher.uuid);
 
       // voucher details
@@ -73,6 +74,14 @@ function receipt(req, res, next) {
       // populate data for the view
       _.extend(data, record, metadata);
 
+      if (voucher.reversed === 1) {
+        return creditNotedRef(db.bid(voucher.uuid));
+      }
+      return null;
+    }).then(creditNoted => {
+      if (creditNoted) {
+        record.details.creditNoteVoucher = creditNoted;
+      }
       return receiptReport.render(data);
     })
     .then((result) => {
@@ -126,3 +135,12 @@ function report(req, res, next) {
     .done();
 }
 
+function creditNotedRef(uuid) {
+  const sql = `
+    SELECT dm.text as reference
+    FROM voucher v
+    JOIN  document_map dm ON v.uuid = dm.uuid 
+    WHERE v.reference_uuid = ?
+  `;
+  return db.one(sql, db.bid(uuid));
+}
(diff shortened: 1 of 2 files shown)


server/controllers/finance/reports/vouchers/receipt.handlebars
```diff
@@ -4,6 +4,15 @@
   <header>
     <!-- headings  -->
     <div class="row">
+
+      {{#if details.isCreditNoted}}
+        <div class="text-center" style="position: absolute; border: 5px solid red; -webkit-transform: rotate(10deg); transform: rotate(10deg); left: 35%; top: 20px;">
```
jniles, line 9: It would be better to use `FORM.LABELS.CANCELED` since a voucher can be either a credit or a debit.

server/controllers/finance/reports/vouchers/index.js
```diff
@@ -126,3 +135,18 @@ function report(req, res, next) {
...
+    SELECT BUID(v.uuid) as uuid, v.date, v.project_id, v.currency_id, v.amount,
+      v.description, v.user_id, v.type_id,  u.display_name, transaction_type.text,
+      CONCAT_WS('.', '${entityIdentifier}', p.abbr, v.reference) AS reference, reversed
+    FROM voucher v
+    JOIN project p ON p.id = v.project_id
+    JOIN user u ON u.id = v.user_id
+    LEFT JOIN transaction_type ON v.type_id = transaction_type.id
+    WHERE v.reference_uuid = ?;
+`;
+  return db.one(sql, db.bid(uuid)).then(voucher => {
```
jniles: The `.then()` here doesn't do anything.  In general, if you have a function that simply returns it's input, it probably can be removed, unless it is needed for a hack.

server/controllers/finance/reports/vouchers/index.js
```diff
@@ -126,3 +135,18 @@ function report(req, res, next) {
     .done();
 }
 
+function creditNotedRef(uuid) {
+  const sql = `
+    SELECT BUID(v.uuid) as uuid, v.date, v.project_id, v.currency_id, v.amount,
+      v.description, v.user_id, v.type_id,  u.display_name, transaction_type.text,
+      CONCAT_WS('.', '${entityIdentifier}', p.abbr, v.reference) AS reference, reversed
+    FROM voucher v
+    JOIN project p ON p.id = v.project_id
```
jniles: All these JOINs are expensive, but I'm not sure we need them.  In fact, I'm pretty sure we could just look this up in the document_map table.  Could we do this instead:
```js
const sql = `
  SELECT dm.text FROM document_map dm WHERE dm.uuid = ?;
`
return db.one(sql, db.bid(uuid));

Comment (jniles) [reply server/controllers/finance/reports/vouchers/index.js]: Otherwise, we could just do a `Vouchers.lookupVoucher(uuid)`.

**Review by jniles — COMMENTED** [review]
@jeremielodi, great PR!   The behavior is correct and the format looks good.  For those curious, it looks like this:

voucherreversal


I've made a few comments to optimize the code a bit.  Let me know what you think.

File: server/controllers/finance/reports/vouchers/index.js (hunk 3 above)
Comment (jeremielodi) [reply server/controllers/finance/reports/vouchers/index.js]: this sql query is not correct, here we are looking for the voucher whitch was created in order to cancel the current one(the one we know it uuid)

this correct sql is :
```sql
SELECT dm.text as reference
FROM voucher v
JOIN  document_map dm ON v.uuid = dm.uuid 
WHERE v.reference_uuid = ?
```

jniles (approved):
Looks good to me!

bors r+

PR #3246, opened 2018-10-10
feat(stock entry and exit)moving alert message to the bottom

### LINKED ISSUE #3242: Move alerts to the bottom of the stock entry/exit modules
This diagram will better explain it:
stock - move no destination defined to a central place

**Rationale**
We should put all our error handling in a single, well defined spot.  Currently, it's a bit all over the place, making the user have to learn a ton of places to check for errors.  It would be best to put this in a single location.

jeremielodi:
closes #3242


jniles:
LGTM.

bors r+

PR #3253, opened 2018-10-15
chore: add i18n keys

jniles:
This commit adds a few missing i18n keys to the JSON files.


jniles:
This should be non-controversial.

bors r+

PR #3282, opened 2018-10-24
Add unit cost in lots registry

mbayopanda:
This PR add the unit cost column in the lots registry


jniles (approved):
Should this have a currency filter?  You probably should also include `type: 'number'` to make sure it sorts correctly.

Otherwise, LGTM.

PR #3287, opened 2018-10-25
Ohada balance sheet enhancement

mbayopanda:
This PR enhance the ohada balance sheet for getting data from the period zero of the next fiscal year.


.env.development
```diff
@@ -5,7 +5,7 @@ PORT=8080
 DB_HOST='localhost'
 DB_USER='bhima'
 DB_PASS='HISCongo2013'
-DB_NAME='bhima_test'
+DB_NAME='imck'
```
lomamech: @mbayopanda restore 'bhima_test' like Db name and not IMCK

lomamech:
image

The display of the 2017 Balance Sheet data is not correct

.env.development
jniles: Also, did you know that you can change the environmental variables globally by just prefixing the script?

```sh
# This should make the DB_NAME the variable
DB_NAME="imck" npm run dev
```

lomamech:
@mbayopanda According to @patsonjunior 'Patrick Kianzuaku' the Passive Part should be displayed in absolute value as for the report operating account with the exception of CI 
"Resultat Net de l'exercice" which can be either debtor or creditor

mbayopanda:
@lomamech could you review this PR please

PR #3293, opened 2018-10-29
fix(journal): edit transaction exchange bug

### LINKED ISSUE #3292: Posting Journal does not update debit_equiv or credit_equiv on transaction edit
It seems like we are accidentally leaving around the old debit_equiv and credit_equiv values if they are changed to zero on edit.  This is probably an `if (value)` error, where `value` is false when it is 0.

See below:
bugineditingdebitcredits
_Fig 1: Unbalanced Txn_

jniles:
This commit fixes the exchange rate bug in the editing the Journal.

Fixes #3292.


mbayopanda (approved):
LGTM :+1

bors r+

PR #3295, opened 2018-10-30
feat(fiscal year) : Message on the result account

mbayopanda:
This PR add a message if the selected account in the closing fiscal year modal contains value

image


jniles (approved):
Looks good to me.

On the long run, we'll probably want to make this query optional.  It is only OHADA that requires the result account to be empty.

bors r+

PR #3303, opened 2018-11-01
fix: clean up nodejs PIDs

jniles:
This commit cleans up the node PIDS in the bash script by killing them if an error occurred.


jniles:
Looks like this is a yarn bug, actually.

https://github.com/yarnpkg/yarn/issues/6608

PR #3304, opened 2018-11-01
Feature: Close Fiscal Years in last period

### LINKED ISSUE #3288: Write closing transaction to Period 13 of current fiscal year
We have come across an interesting case: if you sum the fiscal year N, you will not get the opening balance of fiscal year N + 1.  This is because we write a "virtual" transaction that sums all income / expense accounts into a single account at the end of the fiscal year.

We need both sets of the data - the summed number (period 0) and the closing transaction to show on the balance sheet.  If we do not include the final transaction, we lose the detailed information on movements.

This proposal is to write the closing transaction information to the last period of the fiscal year, another version of period 0.  This will not contain the balances of the sum of the fiscal year, it will contain a single transaction that closes the fiscal year.

jniles:
This PR introduces a much-debated feature for closing fiscal years by writing all closing transactions (any movements of money that need to take place before the next fiscal year) into a final, virtual period.

**What is a "virtual period"?**
It's a period that exists without a date.  We already have one - period 0 - which contains our opening balances.  This final period will contain any additional writings to move money around before transferring the balance of all accounts into the next fiscal year's period 0.  In a twelve-month year, this period has the number 13.

**Why make a closing period?**
We need a closing period to house transactions that are purely conceptual - such as writing of income/expense into a balance account.  Previously, we did this magically by simply dumping the result of Income - Expense into Period 0 of the next year on a particular account.  By putting this writing in a closing period, we are able to:
 1. Show all movements of the year _including the final closing operation_ without looking at the next fiscal year's period 0.
 2. Show all the movements of the year _excluding the final closing operation_.
 3. Reproduce in fine detail the closing operation(s) that the accountant did to close the fiscal year.  This might be important for auditing.
 4. The sum of all values in fiscal year N is guaranteed to be the values in fiscal year N + 1's opening balance.

Closes #3288


.travis.yml
```diff
@@ -11,8 +11,8 @@ services:
 
 language: node_js
 node_js:
-  - "8"
-  -  10
+  - lts
+  - node
```
mbayopanda, line 15: 👍

mbayopanda (approved):
LGTM 👍 

The code works fine, i was wondering the migration script doesn't have code for updating an existing periods of an existing fiscal year.

And also, the ohada balance sheet that i made must be updated after the merge of this code.

jniles:
@mbayopanda, thanks for the review!

I have low confidence in how this PR will go before the IMCK board meetings, so I'm leaving it out of their upgrade.  @lomamech did some tests today on master after @jeremielodi's PR and we are upgrading IMCK based on that.

After the board meetings, we can schedule an additional upgrade when we have confidence in all the code and reports 🕺

jniles:
@mbayopanda, it seems like we haven't heard anything about the board meetings.  When do you think we should merge this branch?

PR #3308, opened 2018-11-05
feat(account) refresh the grid after importation

### LINKED ISSUE #3291: Importing Accounts should refresh the account grid
When you import accounts, the account grid is not refreshed, leading to a blank looking account chart.  We should make sure we call for a refresh after updating the account list.  See below:

jeremielodi:
closes #3291


jniles (commented):
@jeremielodi looks great!  We spoke in person how to re-use the request variable to reduce the amount of code.  Really good work so far!

jniles:
Flaky tests.

bors r+

PR #3311, opened 2018-11-05
feat(account management) add account type column in the account grid

### LINKED ISSUE #3279: Add account type to the Account Management interface
The Account Management interface doesn't tell you what account types any account is at a glance.  This is annoying when trying to assess whether the enterprise has done the right thing or not.  It would be useful to add an additional column to the Account Management interface with the account type listed inline.

jeremielodi:
closes #3279


jeremielodi:
gridc

jniles:
Well, that was easy.

bors r+

PR #3315, opened 2018-11-06
Add search by movement reference in stock movement registry

mbayopanda:
This PR fix the issue #3314 

Closes #3314.  Closes #3318.

PR #3318, opened 2018-11-07
feat(stock movement search modal) add reference filter

### LINKED ISSUE #3314: feat(stock): add the search by stock movement reference
The stock movement registry doesn't allow to search a movement by its reference, this issue is to add this feature in the search modal of the stock movement registry.

jeremielodi:
closes #3314


jeremielodi:
gridc

mbayopanda:
it seems like we have worked on the same issue, there is already a PR #3315  about this issue #3314

PR #3322, opened 2018-11-08
feat(fiscal year) notify while loading data from the server

### LINKED ISSUE #3256: Closing fiscal year does not notify the user of loading
When you try to close a fiscal year, it sometimes says "no data to display" instead of showing a loading indicator.  This is confusing on a slow connection.  See below:

_Gif 1: Flash of "no data" before data shows up_

This should be fixed with a loading indicator.

jeremielodi:
closes #3256

### DIFF
--- a/client/src/modules/fiscal/fiscal.closingBalance.html
+++ b/client/src/modules/fiscal/fiscal.closingBalance.html
@@ -23,6 +23,14 @@
             id="account-balance-grid"
             style="height : 800px;"
             ui-grid="FiscalCBCtrl.gridOptions">
+
+            
+            <bh-grid-loading-indicator
+              loading-state="FiscalCBCtrl.loading"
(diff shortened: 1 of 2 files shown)


client/src/modules/fiscal/fiscal.closingBalance.html
```diff
@@ -23,6 +23,14 @@
             id="account-balance-grid"
             style="height : 800px;"
             ui-grid="FiscalCBCtrl.gridOptions">
+
+            
+            <bh-grid-loading-indicator
+              loading-state="FiscalCBCtrl.loading"
+              empty-state="FiscalCBCtrl.gridOptions.data.length === 0"
+              error-state="FiscalCBCtrl.hasError">
```
jniles, line 31: Will `hasError` ever be set?

jniles (approved):
bors r+

PR #3323, opened 2018-11-08
feat(module) stock tags

### LINKED ISSUE #3201: CRUD for Stock Tags
This issue is about to add a module for create, read, update and delete tags in BHIMA.

Tags can be assigned to many kind of element in bhima for example stock lots.

As tag, we can have : 
- Broken
- Repaired
- Any name we want

At the end, we must have a module for creating, editing or deleting tags; and also a tags registry.

Having a component for selecting tags will be also better.

jeremielodi:
closes #3201

### DIFF
--- a/client/src/i18n/fr/tree.json
+++ b/client/src/i18n/fr/tree.json
@@ -107,6 +107,7 @@
     "STOCK_VALUE" : "[Stock] Valeur de stock",
     "STOCK_MOVEMENTS":"Movements de stock",
     "TRANSACTION_TYPE":"Types de Transaction",
+    "TAGS" : "Gestion des étiquettes",
     "SUBSIDY":"Gestion des subventions",
     "USERS":"Utilisateurs",
     "VOUCHER_REGISTRY":"Registre des Bordereaux",
--- a/client/src/modules/tags/tags.ctrl.js
+++ b/client/src/modules/tags/tags.ctrl.js
@@ -1,0 +1,95 @@
+angular.module('bhima.controllers')
+  .controller('TagsController', TagsController);
+
+TagsController.$inject = [
+  '$uibModal', 'TagsService', 'ModalService',
+  'NotifyService', 'bhConstants', 'uiGridConstants', '$rootScope',
+];
+
+function TagsController($uibModal, Tags, Modal,
+  Notify, bhConstants, uiGridConstants, $rootScope) {
+  const vm = this;
+
+  vm.canEditTags = false;
+
+  vm.createUpdateTagsModal = (tag) => {
+    $uibModal.open({
+      templateUrl : 'modules/tags/modal/createUpdate.html',
+      controller : 'TagsModalController as ModalCtrl',
+      resolve : { data : () => tag },
+    });
+  };
+
+  vm.remove = function remove(uuid) {
+    const message = 'FORM.DIALOGS.CONFIRM_ACTION';
+    Modal.confirm(message)
+      .then(confirmResponse => {
+        if (!confirmResponse) {
+          return;
+        }
+        Tags.delete(uuid)
+          .then(() => {
+            Notify.success('FORM.INFO.DELETE_SUCCESS');
+            loadTags();
+          })
+          .catch(Notify.handleError);
+      });
+  };
+
+  function loadTags() {
+    vm.loading = true;
+    vm.errorState = false;
+    Tags.read()
+      .then(tags => {
+        vm.gridOptions.data = tags;
+      })
+      .catch(err => {
+        vm.errorState = true;
+        Notify.handleError(err);
+      })
+      .finally(() => {
+        vm.loading = false;
+      });
+  }
+
+  $rootScope.$on('TAGS_CHANGED', loadTags);
+
+  const columns = [{
+    field : 'name',
+    displayName : 'FORM.LABELS.NAME',
+    headerCellFilter : 'translate',
+  }, {
+    field : 'actions',
+    enableFiltering : false,
+    width : 100,
+    displayName : '',
+    headerCellFilter : 'translate',
+    cellTemplate : '/modules/tags/templates/action.cell.html',
+  }];
+
+  // ng-click="
+  vm.gridOptions = {
+    appScopeProvider : vm,
+    enableColumnMenus : false,
+    columnDefs : columns,
+    enableSorting : true,
+    data : [],
+    fastWatch : true,
+    flatEntityAccess : true,
+    onRegisterApi : (gridApi) => {
+      vm.gridApi = gridApi;
+    },
+  };
+
+  loadTags();
+  /**
+   * @function toggleInlineFilter
+   *
+   * @description
+   * Switches the inline filter on and off.
+   */
+  vm.toggleInlineFilter = function toggleInlineFilter() {
+    vm.gridOptions.enableFiltering = !vm.gridOptions.enableFiltering;
+    vm.gridApi.core.notifyDataChange(uiGridConstants.dataChange.COLUMN);
+  };
+}
--- a/client/src/modules/tags/tags.html
+++ b/client/src/modules/tags/tags.html
@@ -1,0 +1,43 @@
+<div class="flex-header">
+    <div class="bhima-title">
+      
+      <ol class="headercrumb">
+        <li class="static" translate>TREE.ADMIN</li>
+        <li class="title" translate>TREE.TAGS</li>
+      </ol>
+  
+      <div class="toolbar">
+        <div class="toolbar-item">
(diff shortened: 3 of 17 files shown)


client/src/i18n/fr/tree.json
```diff
@@ -107,6 +107,7 @@
     "STOCK_VALUE" : "[Stock] Valeur de stock",
     "STOCK_MOVEMENTS":"Movements de stock",
     "TRANSACTION_TYPE":"Types de Transaction",
+    "TAGS" : "Gestion des étiquette",
```
jniles: Should this be plural? `des étiquettes`?

client/src/modules/tags/tags.html
```diff
@@ -0,0 +1,40 @@
...
+      </ol>
+  
+      <div class="toolbar">
+        <div class="toolbar-item">
+          <button class="btn btn-default" ng-click="TagsCtrl.createUpdateTagsModal()" data-method="create">
+            <span class="fa fa-plus"></span>
+            <span translate class="text-capitalize">FORM.LABELS.ADD</span>
+          </button>
+        </div>
+        <bh-filter-toggle on-toggle="TagsCtrl.toggleInlineFilter()">
```
jniles: Nice use of `bhFilterToggle`.

I think it needs to be in a `<div class="toolbar-item">` though.  It doesn't include that part itself.

client/src/modules/tags/tags.html
```diff
@@ -0,0 +1,40 @@
...
+      <ol class="headercrumb">
+        <li class="static" translate>TREE.ADMIN</li>
+        <li class="title" translate>TREE.TAGS</li>
+      </ol>
+  
+      <div class="toolbar">
+        <div class="toolbar-item">
+          <button class="btn btn-default" ng-click="TagsCtrl.createUpdateTagsModal()" data-method="create">
+            <span class="fa fa-plus"></span>
+            <span translate class="text-capitalize">FORM.LABELS.ADD</span>
```
jniles: If you include `class="hidden-xs"` next to the `text-capitalize` class, the text will disappear on smaller screens, which is nice.

client/src/modules/tags/tags.html
```diff
@@ -0,0 +1,40 @@
...
+        id="tags-grid"
+        class="grid-full-height"
+        ui-grid="TagsCtrl.gridOptions"
+        ui-grid-auto-resize
+        ui-grid-resize-columns
+        ui-grid-move-columns>
+  
+        <bh-grid-loading-indicator
+          loading-state="TagsCtrl.loading"
+          empty-state="TagsCtrl.gridOptions.data.length === 0">
```
jniles, line 38: Can you include an `error` state here in case something goes wrong?  Often, things like HTTP errors will happen and the user won't know what went wrong.

client/src/modules/tags/tags.service.js
```diff
@@ -0,0 +1,14 @@
...
+  .service('TagsService', TagsService);
+
+TagsService.$inject = ['PrototypeApiService'];
+
+/**
+ * Tags Service
+ *
+ * A service wrapper for the /tags HTTP endpoint.
+ */
+function TagsService(Api) {
```
jniles, line 11: Nice use of `PrototypeApiService`. 👍

server/controllers/admin/tags.js
```diff
@@ -0,0 +1,70 @@
...
+exports.detail = detail;
+exports.delete = remove;
+// add a new tag
+function create(req, res, next) {
+  const sql = `INSERT INTO tags SET ?`;
+  const data = req.body;
+  data.uuid = data.uuid ? db.bid(data.uuid) : db.uuid();
+  db.exec(sql, data)
+    .then(() => {
+      res.sendStatus(201);
```
jniles, line 15: This should send back the ID/uuid of the newly created record.

server/controllers/admin/tags.js
```diff
@@ -0,0 +1,70 @@
...
+
+// get a tag detail
+function remove(req, res, next) {
+  const sql = `
+    DELETE FROM tags WHERE uuid =?
+  `;
+  const uuid = db.bid(req.params.uuid);
+  db.exec(sql, uuid)
+    .then(rows => {
+      res.status(204).json(rows);
```
jniles, line 68: No rows will be returned (it's an empty array).  Instead, use `res.sendStatus(204)` to reflect this fact.

test/end-to-end/tags/tags.page.js
```diff
@@ -0,0 +1,53 @@
...
+
+const FU = require('../shared/FormUtils');
+const GA = require('../shared/GridAction');
+const GridRow = require('../shared/GridRow');
+
+function RolesPage() {
+  const page = this;
+
+  // the grid id
+  const gridId = 'tagss-grid';
```
jniles, line 11: I don't think you ever really use this, since `tagss-grid` doesn't exist.

test/integration/tags.js
```diff
@@ -0,0 +1,83 @@
...
+  const tagsUpdate = {
+    uuid,
+    name : 'Repaired',
+  };
+
+  it('POST /tags add a new tags', () => {
+    return agent.post('/tags')
+      .send(tags)
+      .then((res) => {
+        expect(res).to.have.status(201);
```
jniles, line 32: Can you check the number of tags after this to make sure it got added.

test/integration/tags.js
```diff
@@ -0,0 +1,83 @@
...
+      })
+      .catch(helpers.handler);
+  });
+
+  it('POST /tags add another tags', () => {
+    return agent.post('/tags')
+      .send(tags2)
+      .then((res) => {
+        expect(res).to.have.status(201);
```
jniles, line 51: You can check the number of tags again to mkae sure this works.

test/integration/tags.js
```diff
@@ -0,0 +1,83 @@
...
+  it('PUT /tags update a tags', () => {
+    return agent.put(`/tags/${uuid}`)
+      .send(tagsUpdate)
+      .then((res) => {
+        expect(res).to.have.status(200);
+        return agent.get(`/tags/${uuid}`);
+      })
+      .then(res => {
+        expect(res).to.have.status(200);
+        expect(res.body.name).to.equal(tagsUpdate.name);
```
jniles, line 84: Nice test 👍

server/models/schema.sql
```diff
@@ -2052,4 +2052,12 @@ CREATE TABLE `department`(
   FOREIGN KEY (`enterprise_id`) REFERENCES `enterprise` (`id`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
+
+CREATE TABLE `tags`(
+  `uuid` BINARY(16) NOT NULL PRIMARY KEY,
+  `name` VARCHAR(100) NOT NULL,
+  UNIQUE KEY  (`name`)
```
jniles, line 2088: It looks like tags are supposed to be unique - could you add a test to make sure you cannot add the same name tag twice?
It would be best as an integration test.

client/src/modules/tags/tags.ctrl.js
```diff
@@ -0,0 +1,92 @@
+angular.module('bhima.controllers')
+  .controller('TagsController', TagsController);
+
+TagsController.$inject = [
+  '$uibModal', 'TagsService', 'SessionService', 'ModalService',
+  'NotifyService', 'bhConstants', 'uiGridConstants', '$rootScope',
+];
+
+function TagsController($uibModal, Tags, Session, Modal,
```
jniles: It doesn't look like you use `Session`.  Please check all your imports and remove the ones you do not use.

jniles (commented):
@jeremielodi this is a really great PR!  Tests are well made, it includes both integration and end to end tests, and the code is clean.

I've made a few comments to clean up a few issues, but afterwards it should be able to be quickly merged.  Great work!

server/models/migrations/v0.8.0-v0.9.0/migrate.sql
```diff
@@ -837,3 +837,10 @@ BEGIN
     UPDATE voucher SET reversed = 0 where voucher.uuid = uuid;
   END IF;
 END $$
+
+
+
+CREATE TABLE `tags`(
+  `uuid` BINARY(16) NOT NULL PRIMARY KEY,
+  `name` VARCHAR(100) NOT NULL
```
jniles: In the schema, the `name` is unique, but in the migrate script, you do not have `UNIQUE`. You also don't have a test for uniqueness.  Which should it be?

jniles (approved):
Good enough.  Well done!

bors r+

PR #3337, opened 2018-11-12
feat(general ledger) rename columns while exporting as excel

### LINKED ISSUE #3331: Exporting the General Ledger to Excel should translate title names
The General Ledger export functionality has two bugs in it:
 1. (non-critical) The file name is simply "Grand".  It should probably include the date as a somewhat unique identifier.
 2. (critical) The columns are not translated.  Our users will have no idea what "balance0" is!  We should fix this.

See below for a reproduction of the error:

jeremielodi:
closes #3331


jniles (approved):
Tested and works.

bors r+

PR #3338, opened 2018-11-12
chore(accounts): rewrite end to end tests

jniles:
This commit reworks the account tests to remove unused code and use the `class` pattern for the page objects.  It also standardizes the grid to use `GridRow` instead of the custom test scripts and uses the human readable account number instead of the account id.


jniles:

jeremielodi (commented):
LGTM

PR #3339, opened 2018-11-12
 feature(Fee Center Management)

### LINKED ISSUE #3138: Distribution of Expenses and Profits of Auxiliary Fees Centers
This issues proposes to design a module that will make it possible to distribute the charges or profits of the auxiliary fees centers towards the principals fees centers.

This module will analyze all the transactions on the posting-journal and the general ledger, which contain accounts referring to the auxiliary fees centers, in order to assign them to the main fees centers.

````
Here is a table structure that can allow the distribution
- trans_id : the identifier of the transaction
- account_id : the account identiant
- principal_fee_center_id : The identifier of the main expense center
- auxiliary_fee_center_id : The identifier of the auxiliary cost center
- amount : The amount of the distribution
- currency_id : The identifier of currency
- date_distribution : The Date
- user_id : The User responsible of Distribution
````

### PR DESCRIPTION
**Pull request by lomamech**
Implement Distribution Auxiliaries Fees Centers to the Main Centers
-Implement Component Account Reference Select
-Implement component Fee Center Select
-Implement component Percentage Input
-Implement component Multiple Services Select
- Add function Round Decimal in Util Service
- Display of the messages of the different invalidation messages in the
  different cases the distribution
- Improved expense center report, added optional option to include manual
  distributions,
-Implement distribution key integration test
- Update and Refactor some E2E test
- Restaure Module Fee Center Management

closes #3138

### CODE REVIEW

File: client/src/js/components/bhFeeCenterSelect.js [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+
+  $ctrl.$onInit = function onInit() {
+    $ctrl.loading = true;
+
+    FeeCenters.read(null)
+      .then((feeCenters) => {
+        $ctrl.feeCenters = feeCenters;
+
+        if ($ctrl.filter) {
+          if ($ctrl.principal) {
```
jniles: You can simplify this logic by moving the `$ctrl.prinicpal` check into the filter function.  Like this:
```js
if ($ctrl.filter) {
  $ctrl.feeCenters = feeCenters.filter(center => {
     return $ctrl.principal ? item.is_principal : !item.is_principal;
  });
}
```

client/src/js/components/bhFeeCenterSelect.js
```diff
@@ -0,0 +1,53 @@
...
+          } else {
+            $ctrl.feeCenters = feeCenters.filter(item => {
+              return !item.is_principal;
+            });
+          }
+        }
+      })
+      .catch(Notify.handleError)
+      .finally(() => {
+        $ctrl.loading = false;
```
jniles: Does anything actually use the `loading` property?

client/src/i18n/fr/form.json
```diff
@@ -8,7 +8,10 @@
             "ADD_PAYROLL_CONFIGURATION": "Configurer une période des paiements",
             "ADD_RUBRIC_CONFIGURATION" : "Ajouter une configuration des rubriques",
             "ADD_WEEKEND_CONFIGURATION": "Ajouter une configuration des weekend",
+            "AUTOMATIC_BREACKDOWN": "Ventilation automatique des factures",
```
jniles: `BREACKDOWN` should be spelled `BREAKDOWN` 👍

client/src/i18n/fr/form.json
```diff
@@ -497,7 +515,7 @@
             "PERCENTAGE": "Pourcentage",
             "PERCENT": "Pourcent",
             "PERIODES": "Périodes",
-            "PERIOD": "Périod",
+            "PERIOD": "Période",
```
jniles, line 518: 👍

client/src/js/components/bhPercentageInput.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This is a Percentage input form based on <input type="number">, with specific
+ * validation based on the min and max value.
+ */
+function PercentageInputController() {
+  const $ctrl = this;
+  const isDefined = angular.isDefined;
+
+  // translated label for the form input
+  $ctrl.label = $ctrl.label || 'FORM.LABELS.VALUE';
```
jniles: This binding should be in the `$ctrl.onInit()` lifecycle method.

client/src/modules/templates/bhPercentageInput.tmpl.html
```diff
@@ -0,0 +1,42 @@
...
+    <label class="control-label" translate>{{ ::$ctrl.label }}</label>
+
+    <div class="input-group">
+      <span class="input-group-addon text-uppercase">{{ $ctrl.symbol || '%' }}</span>
+      <input
+        class="form-control"
+        type="number"
+        name="percentageInputValue"
+        ng-model="$ctrl.model"
+        ng-min="$ctrl.min || 0"
```
jniles: It would be better to put these in the `$ctrl.onInit`.  Like this:
```js
$ctrl.$onInit = () => {
  $ctrl.symbol = $ctrl.symbol || '%';
  $ctrl.min = $ctrl.min || 0;
  // etc...
};
```
That way, this is only defined a single time.

client/src/modules/templates/bhPercentageInput.tmpl.html
```diff
@@ -0,0 +1,42 @@
+<ng-form name="bhPercentageInputForm">
+  <div class="form-group"
+    ng-class="{ 'has-error' : $ctrl.validationTrigger && bhPercentageInputForm.percentageInputValue.$invalid }" bh-percentage-input>
```
jniles: Instead of using `validationTrigger`, you should use `bhPercentageInputForm.$submitted`.

client/src/modules/templates/bhServicesMultipleSelect.tmpl.html
```diff
@@ -0,0 +1,29 @@
+<div ng-form="ServicesMultipleForm" bh-services-multiple-select ng-model-options="{ updateOn: 'default' }">
+  <div
+    class="form-group"
+    ng-class="{ 'has-error' : $ctrl.validationTrigger && ServicesMultipleForm.services.$invalid }">
```
jniles: Instead of using `validationTrigger`, you should use `ServicesMultipleForm.$submitted`.

client/src/i18n/en/form.json
```diff
@@ -8,7 +8,10 @@
             "ADD_PAYROLL_CONFIGURATION": "Set up a payment period",
             "ADD_RUBRIC_CONFIGURATION" : "Add a Rubric configuration",
             "ADD_WEEKEND_CONFIGURATION": "Add Weekend Configuration",
+            "AUTOMATIC_BREACKDOWN": "Automatic Breakdown of invoices",
```
jniles: All these "BREAKDOWN"s should be spelled "BREAKDOWN".

client/src/i18n/en/form.json
```diff
@@ -8,7 +8,10 @@
...
             "BACK": "Back",
+            "BREAKDOWN_AUTOMATIQUE": "Breakdown Automatique",
+            "BREAKDOWNS_PERCENTAGES": "Breakdowns in Percentages",
```
jniles: A better way to say this is "Breakdown by percentage". 👍

client/src/modules/templates/bhPercentageInput.tmpl.html
```diff
@@ -0,0 +1,42 @@
...
+        ng-min="$ctrl.min || 0"
+        ng-max="$ctrl.max || 100"
+        required>
+    </div>
+
+    <!-- validation messages -->
+    <p class="help-block" ng-messages="bhPercentageInputForm.percentageInputValue.$error" ng-show="$ctrl.validationTrigger">
+
+      <!-- displays: "This value is required" -->
+      <span ng-message="required">
```
jniles: It looks like all these errors are already [messages.tmpl.html](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/modules/templates/messages.tmpl.html).

Could we just use our regular validation for this?  Like:
```html
<div class="help-block" ng-messages="CreateForm.name.$error" ng-show="CreateForm.$submitted">                                                                                                    
</div>
```

server/controllers/finance/distributionFeeCenter/automatic.js
```diff
@@ -0,0 +1,74 @@
...
+*/
+const db = require('../../../lib/db');
+const NotFound = require('../../../lib/errors/NotFound');
+
+function automatic(req, res, next) {
+  const { data } = req.body;
+  const transUuids = data.map(item => db.bid(item.uuid));
+
+  const sql = `
+    SELECT BUID(gl.uuid) AS trans_uuid, gl.trans_id, gl.debit_equiv, gl.credit_equiv,
```
jniles: Hmm... `trans_uuid` is a confusing name because transactions don't have a `uuid`.  They have an `id`.  The `uuid` is for each row in the general ledger/posting journal.  Could we name this field something different?  Why not just `uuid`?  Or `row_uuid`?

server/controllers/finance/distributionFeeCenter/breackDown.js
```diff
@@ -0,0 +1,69 @@
...
+          creditValuePercent,
+          transaction.currency_id,
+          new Date(),
+          userId,
+        ]);
+      }
+    });
+  });
+
+  const sql = `INSERT INTO fee_center_distribution (
```
jniles: Is `fee_center_distribution` going to be the same size of the `general_ledger`?  It seems like we are repeating the following information between the two tables:
 1. account_id
 2. debit_equiv
 3. credit_equiv
 3. trans_uuid -> general_ledger.uuid
 4. trans_id
 5. currency_id

Is there any way we can reduce the number of repeated columns here?

server/controllers/finance/distributionFeeCenter/getDistributed.js
```diff
@@ -0,0 +1,55 @@
...
+*
+* This function searches all the distributions of the costs and profits made, in order to allow possible updates.
+*/
+const db = require('../../../lib/db');
+const FilterParser = require('../../../lib/filter');
+
+function getDistributed(req, res, next) {
+  const options = req.query;
+  options.is_cost = options.typeFeeCenter;
+  options.limit = 100;
```
jniles, line 12: Why is this `limit` forced to be 100?

server/controllers/finance/reports/fee_center/index.js
```diff
@@ -0,0 +1,127 @@
...
+const AccountReference = require('../../accounts').references;
+const setting = require('./setting');
+const getDistributionKey = require('../../distributionFeeCenter/getDistributionKey');
+// expose to the API
+exports.report = report;
+
+// default report parameters
+const DEFAULT_PARAMS = {
+  csvKey : 'account_reference',
+  filename : 'TREE.ACCOUNT_REFERENCE_REPORT',
```
jniles: I think this renders the `FEE_CENTER_REPORT`, right?  You should change all references of `ACCOUNT_REFERENCE_REPORT` to `FEE_CENTER_REPORT`, if so.

server/controllers/finance/reports/fee_center/report.handlebars
```diff
@@ -0,0 +1,515 @@
...
+              </th> 
+            </tr>
+          </thead>
+          <tbody>
+            <tr style="background-color:#dfdfdf; font-size: 15px"> 
+              <td colspan="3" class="text-uppercase"> <strong> {{translate "TABLE.COLUMNS.PRINCIPAL_CENTERS" }} </strong></td> 
+            </tr>
+            {{#each principalProfitCenters}}
+              <tr>
+                <td class="text-right"> {{add @index 1}} </td>
```
jniles, line 110: Is this array sorted?  What determines the order of the `@index`?

Generally we try to avoid putting `@index` in the rendering since it can change based on the sort order.  If there is a real `id` would could use, that would be preferable.

server/models/migrations/v0.8.0-v0.9.0/migrate.sql
```diff
@@ -847,3 +840,107 @@ CREATE TABLE `tags`(
 
 INSERT INTO unit VALUES
 (217, 'Tags','TREE.TAGS','', 1,'/modules/tags/tags','/tags');
+=======
```
jniles: These `===` are because of a broken merge.  Go ahead and remove them.

server/models/schema.sql
```diff
@@ -2081,12 +2081,91 @@ CREATE TABLE `department`(
...
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
+DROP TABLE IF EXISTS `fee_center`;
+CREATE TABLE `fee_center` (
+  `id` MEDIUMINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `label` VARCHAR(200) NOT NULL,
+  `is_principal` tinyint(1) UNSIGNED DEFAULT 0,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `fee_center_1` (`label`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
jniles: We are using the charset `utf8mb4`.  Please use `utf8mb4` for all your tables.

jniles: You should also change your server to use `utf8mb4` as well, to make sure your environment is the same as the ones in production.

client/src/js/components/bhPercentageInput.js
```diff
@@ -0,0 +1,30 @@
+angular.module('bhima.components')
+  .component('bhPercentageInput', {
+    templateUrl : 'modules/templates/bhPercentageInput.tmpl.html',
+    controller : PercentageInputController,
+    bindings : {
+      currencyId : '<', // one-way binding
+      model : '=', // two way binding
```
jniles, line 7: In general, we try to avoid two way bindings since they unpredictable.  The new angular best practice (of two years or so) is to use one-way bindings, with a callback when things change.  Like this:
```js
bindings : {
  model : '<',
  onChange : '&',
}
```

That way, you know exactly when the data is changed and who changed it.

You don't have to fix this in this PR, but in the future, please use one way bindings.  We can work on this together if you need help creating one.

client/src/js/components/bhServicesMultipleSelect.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+function ServicesMultipleSelectController(Services, Notify) {
+  const $ctrl = this;
+
+  $ctrl.$onInit = function onInit() {
+    // label to display
+    $ctrl.label = $ctrl.label || 'FORM.LABELS.SERVICES';
+
+    // fired when the Services has been selected or removed from the list
+    $ctrl.onChange = $ctrl.onChange;
```
jniles: Please remove this line.  It doesn't do anything.  It's like saying:
```js
1 = 1;
```

client/src/modules/distribution_center/modals/distribution.modal.js
```diff
@@ -0,0 +1,99 @@
...
+
+  const path = vm.transaction.updating ? 'update_distribution_center' : 'distribution_center';
+
+  // exposed methods
+  vm.submit = submit;
+  vm.latestViewFilters = DistributionCenter.filters.formatView();
+
+  FeeCenters.read()
+    .then((feeCenter) => {
+      vm.principalFeeCenter = feeCenter.filter(item => {
```
jniles: Could you make this plural since it is an array?  `vm.principalFeeCenters`.

client/src/modules/distribution_center/modals/distribution.modal.js
```diff
@@ -0,0 +1,99 @@
...
+          vm.transaction.distributionValues.forEach(values => {
+            if (item.id === values.id) {
+              vm.transaction.values[item.id] = values.debit_equiv || values.credit_equiv;
+            }
+          });
+        });
+      }
+
+      if (!vm.principalFeeCenter.length) {
+        vm.noPrincilFeeCenter = true;
```
jniles: This variable should be `vm.noPrincipalFeeCenters`.

jniles: Actually... I can't figure out where this is used.  Do you ever use it?  If not, let's remove it.

client/src/modules/distribution_center/modals/distribution.modal.js
```diff
@@ -0,0 +1,99 @@
...
+      sumDistributed += vm.transaction.values[key];
+    });
+
+    sumDistributed = util.roundDecimal(sumDistributed, 2);
+    const amountEquiv = util.roundDecimal(vm.transaction.amount_equiv, 2);
+    const diffAmount = util.roundDecimal((amountEquiv - sumDistributed), 2);
+
+    vm.invalidDistribution = sumDistributed !== amountEquiv;
+
+    vm.diffAmount = (diffAmount > 0)
```
jniles: Isn't this the same as `vm.diffAmount = Math.abs(diffAmount)`?  If so, I think that `Math.abs()` is clearer about what you are doing.

client/src/modules/distribution_center/modals/distribution_key.modal.html
```diff
@@ -0,0 +1,45 @@
...
+      <br>
+      <label class="badge badge-warning"> {{ DistributionKeyModalCtrl.errorMessage }} </label>
+    </div>    
+  </div>
+
+  <div class="modal-body" style="max-height: 70vh; overflow:auto;">
+    <div ng-repeat="principal in DistributionKeyModalCtrl.principalFeeCenter" class="form-group">
+      <bh-percentage-input
+        label="{{principal.label}}"
+        symbol="%"
```
jniles: Won't percentage inputs always have `%` as a symbol?  If so, can we just move `%` into the component and not have that binding?

client/src/modules/distribution_center/distribution_center.service.js
```diff
@@ -0,0 +1,176 @@
...
+  function openDistributionKeyModal(data) {
+    return Modal.open({
+      templateUrl : 'modules/distribution_center/modals/distribution_key.modal.html',
+      size : 'md',
+      animation : false,
+      keyboard : false,
+      backdrop : 'static',
+      controller : 'DistributionKeyModalController as DistributionKeyModalCtrl',
+      resolve : {
+        seetings : () => data,
```
jniles: Could you spell this as `settings`?  `seetings` -> `settings`

client/src/modules/distribution_center/distribution_center.service.js
```diff
@@ -0,0 +1,176 @@
...
+  /**
+   * @function openSettingModal
+   * @description
+   * This functions opens the setting modal form for Setting of distribution.
+   */
+  function openSettingModal(filters) {
+    return Modal.open({
+      templateUrl : 'modules/distribution_center/modals/setting_distribution.modal.html',
+      size : 'md',
+      animation : false,
```
jniles, line 80: The properties `size`, `keyboard`, and `backdrop` are all set to the [default values](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/app.js#L242).  You can go ahead and remove them.

test/data.sql
```diff
@@ -965,3 +965,74 @@ INSERT INTO `user_role`(uuid, user_id, role_uuid)
...
+INSERT INTO `reference_fee_center` (`id`, `fee_center_id`, `account_reference_id`, `is_cost`) VALUES (15, 3, 3, 0);
+INSERT INTO `reference_fee_center` (`id`, `fee_center_id`, `account_reference_id`, `is_cost`) VALUES (21, 6, 4, 0);
+INSERT INTO `reference_fee_center` (`id`, `fee_center_id`, `account_reference_id`, `is_cost`) VALUES (22, 4, 9, 1);
+
+-- GENERAL LEDGER FOR DISTRIBUTION FEE CENTER TEST
+INSERT INTO `general_ledger` (`uuid`, `project_id`, `fiscal_year_id`, `period_id`, `trans_id`, `trans_id_reference_number`, `trans_date`, `record_uuid`, `description`, `account_id`, `debit`, `credit`, `debit_equiv`, `credit_equiv`, `currency_id`, `entity_uuid`, `reference_uuid`, `comment`, `transaction_type_id`, `user_id`, `cc_id`, `pc_id`, `created_at`, `updated_at`) VALUES (0xE70109CCE0DC11E89F4F507B9DD6DEA5, 1, 4, 201811, 'TPA37', 37, '2018-11-05 10:26:05', 0x79B0393553C54498A5ECA8CA6DFEA7AC, 'Facture de Test 2 Patient (PA.TPA.2) pour 2 items dans le service Medecine Interne. ', 174, 256.6200, 0.0000, 256.6200, 0.0000, 2, 0x3BE232F9A4B94AF6984C5D3F87D5C107, NULL, NULL, 11, 1, NULL, NULL, '2018-11-05 10:29:21', NULL);
```
jniles: Are there any vouchers/invoices/cash payments for these records?  In BHIMA it isn't possible to have records in the posting journal without making invoices, cash payments, or invoices. 

It would be better to create records for these using the use `PostVoucher()`, `PostInvoice`, etc...

jniles (commented):
@lomamech, I've completed an initial review.  This is a ton of code!

I've made some comments, let me know if you have any questions.  We should see if Larry can review this at some point.

lomamech:
@jniles Please leave me two hours to complete this pull request before doing another review

jniles:
Of course!  No problem.  Let me know when you are ready 👍

server/controllers/finance/distributionFeeCenter/breackDown.js
lomamech: If you are in an institution that does not use the distribution keys in this case, the distribution_fee_center will have a lot more lines than the general_ledger

Each of this column is needed, 
image

In this illustration, we see that a single line of the general ledger distributed to three main centers gives 3 lines

server/controllers/finance/distributionFeeCenter/getDistributed.js
lomamech: Since this table may have multiple lines, at the risk of having overflow, I have limited the row numbers to 100 if no filter is enabled

server/controllers/finance/reports/fee_center/report.handlebars
lomamech: It's just to display the number on a list

client/src/modules/distribution_center/modals/distribution.modal.js
lomamech: It's Used in distribution.modal.html

test/data.sql
lomamech: Yes, I have set Its, 
````
-- INVOICE FOR DISTRIBUTION FEE CENTER
INSERT INTO `invoice` (`project_id`, `uuid`, `cost`, `debtor_uuid`, `service_id`, `user_id`, `date`, `description`, `reversed`, `edited`, `created_at`) 
VALUES (1, 0x79B0393553C54498A5ECA8CA6DFEA7AC, 256.6200, 0x3BE232F9A4B94AF6984C5D3F87D5C107, 3, 1, '2018-11-05 10:26:05', 'Facture de Test 2 Patient (PA.TPA.2) pour 2 items dans le service Medecine Interne. ', 0, 0, '2018-11-05 10:26:32');
INSERT INTO `invoice_item` (`invoice_uuid`, `uuid`, `inventory_uuid`, `quantity`, `inventory_price`, `transaction_price`, `debit`, `credit`) 
VALUES (0x79B0393553C54498A5ECA8CA6DFEA7AC, 0x3B456F6FAA594CE5BD80C91647EB6CF9, 0xDCFBCD6DB50F48B385614A0B8A4EFBA5, 100, 1.9000, 1.9000, 0.0000, 190.0000);
INSERT INTO `invoice_item` (`invoice_uuid`, `uuid`, `inventory_uuid`, `quantity`, `inventory_price`, `transaction_price`, `debit`, `credit`) 
VALUES (0x79B0393553C54498A5ECA8CA6DFEA7AC, 0xC150EFE416144C428643BFEA00608800, 0xC560E79CE57A4BC9BA53406AA861F858, 40, 5.1200, 5.1200, 0.0000, 204.8000);
````

**Comment by lomamech** [comment]
@jniles You can now continue with the code review

File: client/src/js/components/bhAccountReferenceSelect.js [hunk 28]
```diff
@@ -0,0 +1,64 @@
...
+function AccountReferenceSelectController(AccountReferences, $timeout, $scope, Notify) {
+  const $ctrl = this;
+
+  // fired at the beginning of the account Reference select
+  $ctrl.$onInit = function $onInit() {
+    // translated label for the form input
+    $ctrl.label = $ctrl.label || 'FORM.LABELS.REFERENCE';
+
+    // fired when an account reference has been selected
+    $ctrl.onSelectCallback = $ctrl.onSelectCallback || angular.noop;
```
jniles: It looks like this is required (`&`) - if so, there is no need to have the `onSelectCallback` default.

client/src/modules/distribution_center/distribution_center.html
```diff
@@ -0,0 +1,95 @@
...
+        <button
+          ng-click="DistributionCenterCtrl.setting()"
+          data-method="setting"
+          class="btn btn-default">
+          <span class="fa fa-cogs"></span> <span class="hidden-xs" translate> FORM.BUTTONS.SETTING </span>
+        </button>
+      </div> 
+      
+      <div class="toolbar-item"> 
+        <bh-filter-toggle on-toggle="DistributionCenterCtrl.toggleFilter()">
```
jniles, line 62: Good use of `bhFilterToggle` 👍

client/src/modules/distribution_center/modals/setting_distribution.modal.js
```diff
@@ -0,0 +1,98 @@
+angular.module('bhima.controllers')
+  .controller('SettingDistributionModalController', SettingDistributionModalController);
+
+SettingDistributionModalController.$inject = [
+  '$state', 'DistributionCenterService', 'filters', 'ModalService',
+  'NotifyService', 'bhConstants', 'Store', 'util', '$uibModalInstance', '$translate'];
+
+function SettingDistributionModalController($state, DistributionCenter, filters, ModalService, Notify,
```
jniles: It looks like `ModalService` is never used here.  Could you remove it?

client/src/modules/distribution_center/distribution_key/distribution_key.js
```diff
@@ -0,0 +1,146 @@
+angular.module('bhima.controllers')
+  .controller('DistributionKeyController', DistributionKeyController);
+
+DistributionKeyController.$inject = [
+  'DistributionCenterService', 'ModalService', 'NotifyService', 'uiGridConstants',
```
jniles: It looks like `ModalService` isn't used here - could you remove that?

client/src/modules/distribution_center/distribution_key/distribution_key.js
```diff
@@ -0,0 +1,146 @@
...
+      field : 'auxiliary_label',
+      displayName : 'TABLE.COLUMNS.AUXILIARY_CENTER',
+      headerCellFilter : 'translate',
+    }, {
+      field : 'principal_label',
+      displayName : 'TABLE.COLUMNS.PRINCIPAL_CENTER',
+      headerCellFilter : 'translate',
+    }, {
+      field : 'rate',
+      type : 'number',
```
jniles, line 47: Nice! 👍

client/src/modules/distribution_center/distribution_center_update.service.js
```diff
@@ -0,0 +1,49 @@
...
+
+DistributionCenterUpdateService.$inject = ['PrototypeApiService', 'FilterService', 'appcache'];
+
+/**
+ * @class DistributionCenterUpdateService
+ *
+ * @description
+ * Encapsulates common requests to the /distribution_fee_center/ URL.
+ */
+function DistributionCenterUpdateService(Api, Filters, AppCache) {
```
jniles, line 12: Nice service 👍 

It is well laid out and very easy to read.  Good code!

server/controllers/finance/distributionFeeCenter/breackDown.js
jniles: Okay... we'll see what Larry thinks.  Can we at least remove the currency_id?  If we are using the `debit_equiv` and `credit_equiv`, it should always be the enterprise currency, correct?

server/controllers/finance/distributionFeeCenter/getDistributionKey.js
```diff
@@ -0,0 +1,42 @@
...
+  `;
+
+  db.exec(sql)
+    .then((rows) => {
+      res.status(200).json(rows);
+    })
+    .catch(next);
+}
+
+function allDistibutionKey() {
```
jniles: Looks like there is a small typo here: `Distibution` -> `Distribution`

server/controllers/finance/distributionFeeCenter/proceed.js
```diff
@@ -0,0 +1,71 @@
...
+* Distribution Fee Center Controller
+*
+* This function makes it possible to proceed to a basic distribution
+* of a profit or a cost, of an auxiliary center towards the main centers
+*/
+const db = require('../../../lib/db');
+
+function proceed(req, res, next) {
+  const { data } = req.body;
+  const isDebtor = !!data.debit_equiv;
```
jniles: Would this check also work: `data.debit_equiv > 0`?  It might be clearer - the debit value must be positive instead of a boolean.

server/controllers/finance/distributionFeeCenter/proceed.js
```diff
@@ -0,0 +1,71 @@
...
+  const dataToDistribute = [];
+
+  data.user_id = req.session.user.id;
+  data.uuid = data.row_uuid || data.uuid;
+
+  Object.keys(dataValues).forEach((principalCenterId) => {
+    const debitEquivDistributed = isDebtor ? dataValues[principalCenterId] : 0;
+    const creditEquivDistributed = isDebtor ? 0 : dataValues[principalCenterId];
+
+    if (debitEquivDistributed || creditEquivDistributed) {
```
jniles: It seems like this check will always be true.  Will it ever be false?  What happens if we remove it?

server/controllers/finance/distributionFeeCenter/proceed.js
```diff
@@ -0,0 +1,71 @@
...
+    auxiliary_fee_center_id, 
+    principal_fee_center_id, 
+    debit_equiv, 
+    credit_equiv, 
+    currency_id, 
+    date_distribution, user_id) VALUES ?`;
+
+  const transaction = db.transaction();
+
+  if (dataToDistribute.length) {
```
jniles, line 52: If `dataToDistribute.length` is 0, should we be throwing an error?  It looks like we will call `transaction.execute()` without adding any queries in that case.

server/controllers/finance/distributionFeeCenter/referenceAccount.js
```diff
@@ -0,0 +1,115 @@
...
+function auxilliary(params) {
+  let sqlIsCost = '';
+  const typeFeeCenter = parseInt(params.typeFeeCenter, 10);
+
+  if (typeFeeCenter === 1 || typeFeeCenter === 0) {
+    sqlIsCost = ` AND reference_fee_center.is_cost = ${typeFeeCenter}`;
+  }
+
+  // fee_center_id
+  const filter = params.fee_center_id ? ` AND fee_center.id = ${params.fee_center_id} ` : '';
```
jniles: It looks like this is vulnerable to an SQL injection attack.  In theory, we could set `fee_center_id` to `1 OR 1=1` and get everything, right?

The best course of action would be to use `FilterParser`.  If that is too complicated in this case, you should instead use `db.escape(params.fee_center_id)` ([docs](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/db/index.js#L232)) and it should be fine.  Like this:
```js
const filter = params.fee_center_id ? ` AND fee_center.id = ${db.escape(params.fee_center_id)} ` : '';

File: server/controllers/finance/distributionFeeCenter/configuration.js [hunk 39]
```diff
@@ -0,0 +1,67 @@
...
+
+      if (query.fee_center_id || query.trans_id) {
+        delete options.limit;
+      }
+
+      return generalLedger.findTransactions(options);
+    })
+    .then((rows) => {
+
+      if (rows.length) {
```
jniles: I don't think there is a need for this check.  If `rows.length` is 0, `rows.forEach()` will just skip it anyway.  So you can probably just remove the `if (rows.length) {` check.

server/controllers/finance/distributionFeeCenter/referenceAccount.js
```diff
@@ -0,0 +1,115 @@
...
+      // Get the child of all accounts title definied in Account Reference Items
+      accountsValids.forEach(item => {
+        tabAccountsValids.push(`(SELECT ${item.id} AS id, '${item.label}' AS label,
+      ${item.is_principal} AS is_principal, ${item.is_cost} AS is_cost,
+      ${item.account_reference_id} AS account_reference_id, account.id AS account_id, account.number,
+      ${item.is_exception} AS is_exception, account.type_id
+      FROM account
+      WHERE account.number LIKE '${item.number}%' AND account.type_id IN (4,5))`);
+      });
+      const sqlGetAllAccountsValids = tabAccountsValids.join(' UNION ');
```
jniles, line 76: This seems really inefficient.  Normally, we get accounts using a `WHERE` query.  Is there anyway we can avoid using a `UNION` here?

server/controllers/finance/distributionFeeCenter/referenceAccount.js
```diff
@@ -0,0 +1,115 @@
...
+
+      if (accountsExceptions.length) {
+      // Get the child of all accounts title definied in Account Reference Items
+        accountsExceptions.forEach(item => {
+          tabAccountsExceptions.push(`(SELECT ${item.id} AS id, '${item.label}' AS label,
+        ${item.is_principal} AS is_principal,
+        ${item.account_reference_id} AS account_reference_id, account.id AS account_id,
+        account.number, ${item.is_exception} AS is_exception, account.type_id
+        FROM account
+        WHERE account.number LIKE '${item.number}%' AND account.type_id IN (4,5))`);
```
jniles, line 89: Can we use account ids instead of `LIKE`?  The `LIKE` statement will perform a regex on the account number, which is less efficient than just finding by ID.  Why don't we know the account IDs here?

server/controllers/finance/distributionFeeCenter/setting.js
```diff
@@ -0,0 +1,54 @@
...
+
+const db = require('../../../lib/db');
+
+function setting(req, res, next) {
+  const { data } = req.body;
+  const dataValues = data.values;
+  data.user_id = req.session.user.id;
+  const distributionKey = [];
+
+  Object.keys(dataValues).forEach((principalCenterId) => {
```
jniles: Just FYI, you have a couple of different tools you can use to loop through objects besides `Object.keys(o).forEach(fn)`.  Here are a couple of them:

```js
// imagine we have this object
const o = { id : 1, name : 'jniles', rate : 1.15, ill : true };

// we can do this to loop through only keys
Object.keys(o).forEach(key => console.log(key));
// => prints 'id', 'name', 'rate', 'ill'

// or we can loop through only values:
Object.values(o).forEach(value => console.log(values));
// => prints '1', 'jniles', '1.15', 'true'

// or we can loop through both keys and values using entries()
Object.entries(o).forEach(([key, value]) => console.log(key, value));
// => prints 'id' 1, 'name' 'jniles', 'rate' '1.15', 'ill' true'
```

jniles: The documentation has more examples: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

server/controllers/finance/feeCenter.js
```diff
@@ -0,0 +1,229 @@
...
+      };
+
+      return data;
+    });
+}
+
+// Lists
+function list(req, res, next) {
+  const sql = `
+    SELECT f.id, f.label, f.is_principal, GROUP_CONCAT(' ', LOWER(ar.abbr)) AS abbrs,
```
jniles, line 47: Wow, I had never heard of `GROUP_CONCAT` before!  Awesome - this is really good to know about.

server/models/schema.sql
jniles: This still needs to be addressed.

jniles (commented):
This is probably as far as I can go just looking at the code.  We'll need to get feedback from out client to understand how to improve this next, if it needs to be improved.

client/src/i18n/en/form.json
```diff
@@ -8,7 +8,10 @@
             "ADD_PAYROLL_CONFIGURATION": "Set up a payment period",
             "ADD_RUBRIC_CONFIGURATION" : "Add a Rubric configuration",
             "ADD_WEEKEND_CONFIGURATION": "Add Weekend Configuration",
+            "AUTOMATIC_BREAKDOWN": "Automatic Breakdown of invoices",
             "BACK": "Back",
+            "BREAKDOWN_AUTOMATIQUE": "Breakdown Automatique",
```
mbayopanda: `AUTOMATIC_BREAKDOWN` and `BREAKDOWN_AUTOMATIQUE` are so close, since the first is related to invoices, could you rename the first to `AUTO_INVOICES_BREAKDOWN`, and the second to `AUTO_BREAKDOWN`

client/src/i18n/en/form.json
```diff
@@ -8,7 +8,10 @@
...
+            "BREAKDOWNS_PERCENTAGES": "Breakdown by percentage",
```
mbayopanda, line 13: `BREAKDOWNS_PERCENTAGES` is in plural but the definition doesn't contains plural elements.

If `BREAKDOWNS_PERCENTAGES` is not for plural elements, could you rename to a singular version some thing like `BREAKDOWN_PERCENTAGE`.

client/src/i18n/en/form.json
```diff
@@ -198,6 +212,7 @@
             "UPLOAD_PICTURE_FAILED": "The selected file is not an picture",
             "USER_CASHBOX": "These are the cashboxes that you have permission for manage",
             "USER_DEPOT": "These are the depots that you have permission for manage",
+            "VALIDATE_DISTRIBUTION": "Please click here to validate the distribution of all Ancillary Cost Centers to the Main Cost Centers according to the respective distribution keys",
```
mbayopanda, line 214: I learned that `Ancillary` is similar to `Auxiliary`  👍

client/src/i18n/en/form.json
```diff
@@ -515,6 +533,7 @@
             "PRICE": "Price",
             "PRICE_LIST": "Price List",
             "PRINCIPAL": "Principal",
+            "PRINCIPAL_2": "Principal",
```
mbayopanda, line 536: Why do you have `PRINCIPAL` and `PRINCIPAL_2` with the same definition or value ? Could you just use `PRINCIPAL`.

client/src/i18n/en/form.json
```diff
@@ -799,15 +824,20 @@
         "WARNINGS": {
             "ATTENTION_CONFIGURED": "Attention, only employees waiting for configuration can be configured",
             "ATTENTION_PAYSLIPS": "Attention, there are no payslips records for employees waiting for configuration",
-            "ATTENTION_WAITING_LIST": "Attention, only employees configured can be placed in the waiting list of payroll",            
+            "ATTENTION_WAITING_LIST": "Attention, only employees configured can be placed in the waiting list of payroll",
+            "ATTENTION_SELECTED": "Warning you have not selected any element",
+            "ATTENTION_WAITING_LIST": "Attention, only employees configured can be placed in the waiting list of payroll",
```
mbayopanda: This entry `ATTENTION_WAITING_LIST` is duplicated in lines 827 and 829

client/src/js/components/bhPercentageInput.js
```diff
@@ -0,0 +1,36 @@
...
+      validationTrigger : '<', // one-way binding
```
mbayopanda: Does the `validationTrigger` is necessary since it is not used ?

client/src/js/components/bhServicesMultipleSelect.js
```diff
@@ -0,0 +1,44 @@
+angular.module('bhima.components')
+  .component('bhServicesMultipleSelect', {
+    templateUrl : 'modules/templates/bhServicesMultipleSelect.tmpl.html',
+    controller  : ServicesMultipleSelectController,
+    bindings    : {
+      onChange : '&',
+      servicesIds : '<?',
+      label : '@?',
+      required : '<?',
+      validationTrigger : '<?',
```
mbayopanda, line 10: Could you remove the `validationTrigger` since it is not used

client/src/modules/distribution_center/distribution_center.js
```diff
@@ -0,0 +1,298 @@
...
+    DistributionCenters.cacheFilters();
+    vm.latestViewFilters = DistributionCenters.filters.formatView();
+
+    load(DistributionCenters.filters.formatHTTP(true));
+  }
+
+  function load(filters) {
+    // flush error and loading states
+    vm.hasError = false;
+    toggleLoadingIndicator();
```
mbayopanda, line 249: In this line the vm.loading must be set to true, instead of just toggle it.

client/src/modules/distribution_center/distribution_center.service.js
```diff
@@ -0,0 +1,176 @@
...
+  const service = new Api('/distribution_fee_center/');
+  const distributionFilters = new Filters();
+  const filterCache = new AppCache('distribution-center-filters');
+
+  service.filters = distributionFilters;
+  service.openSettingModal = openSettingModal;
+  service.openDistributionModal = openDistributionModal;
+  service.cacheFilters = cacheFilters;
+  service.proceedDistribution = proceedDistribution;
+  service.proceedBreackDownPercent = proceedBreackDownPercent;
```
mbayopanda: If it is possible could you replace all Breack in names of variables or any other elements by Break

client/src/modules/distribution_center/distribution_key/distribution_key.js
```diff
@@ -0,0 +1,146 @@
...
+        if (changes) {
+          load();
+        }
+      });
+  }
+
+  function load() {
+    // flush error and loading states
+    vm.hasError = false;
+    toggleLoadingIndicator();
```
mbayopanda, line 112: Could you set the vm.loading to true instead of calling the function.

client/src/modules/distribution_center/modals/distribution_key.modal.js
```diff
@@ -0,0 +1,77 @@
+angular.module('bhima.controllers')
+  .controller('DistributionKeyModalController', DistributionKeyModalController);
+
+DistributionKeyModalController.$inject = [
+  '$state', 'NotifyService', 'DistributionCenterService', 'PayrollConfigurationService',
+  'ExchangeRateService', 'SessionService', '$uibModalInstance', 'PatientInvoiceService',
+  'FeeCenterService', 'util', 'settings', '$translate',
+];
```
mbayopanda, line 7: Could you remove : PayrollConfigurationService, ExchangeRateService, PatientInvoiceService, and util since they are not used

client/src/modules/distribution_center/modals/distribution_key.modal.js
```diff
@@ -0,0 +1,77 @@
...
+
+function DistributionKeyModalController(
+  $state, Notify, DistributionCenter, Configuration,
+  Exchange, Session, ModalInstance, Invoices, FeeCenters, util, settings, $translate,
+) {
```
mbayopanda, line 12: These parameters must be removed : Configuration, Exchange, Invoices and util because they are not used.

client/src/modules/fee_center/modals/fee_center.modal.js
```diff
@@ -0,0 +1,150 @@
+angular.module('bhima.controllers')
+  .controller('FeeCenterModalController', FeeCenterModalController);
+
+FeeCenterModalController.$inject = [
+  '$state', 'FeeCenterService', 'ModalService', 'NotifyService', 'appcache',
+];
```
mbayopanda, line 6: The `ModalService` is never used, could you remove it ?

client/src/modules/fee_center/modals/fee_center.modal.js
```diff
@@ -0,0 +1,150 @@
...
+    vm.auxiliaryCenter = value ? 1 : 0;
+    if (value) {
+      vm.hasProfitCenter = !value;
+      vm.hasCostCenter = !value;
+    }
+
+    if (vm.auxiliaryCenter) {
+      vm.isCostCenter = 0;
+      vm.isProfitCenter = 0;
+    }
```
mbayopanda, line 90: Instead of creating two `if` statement, you could just write one `if` statement with all the code inside : 
```js
if (value) {
  vm.hasProfitCenter = !value;
  vm.hasCostCenter = !value;
  vm.isCostCenter = 0;
  vm.isProfitCenter = 0;
}
```

server/controllers/finance/distributionFeeCenter/referenceAccount.js
lomamech: @jniles This section allows to generate a dynamic query, sqlGetAllAccountsValids will generate a request of this kind 
image

client/src/modules/templates/bhAccountReferenceSelect.tmpl.html
```diff
@@ -0,0 +1,34 @@
+<div ng-form="{{ $ctrl.name }}" bh-account-reference-select ng-model-options="{ updateOn: 'default' }">
+  <div
+    class="form-group"
+    ng-class="{ 'has-error' : $ctrl.validationTrigger && AccountReferenceForm.account_reference_id.$invalid }">
```
mbayopanda: Is it possible to remove the $ctrl.validationTrigger, by using the form name.

You can perform that by replacing the `ng-form="{{ $ctrl.name }}"` by something like `ng-form="AccountReferenceForm"`.

And replace all `$ctrl.validationTrigger` by `AccountReferenceForm.$submitted`

client/src/modules/templates/bhAccountReferenceSelect.tmpl.html
```diff
@@ -0,0 +1,34 @@
...
+    </label>
+    <ng-transclude></ng-transclude>
+    <div class="text-danger" ng-if="!$ctrl.accountReferences.length">
+      <i class="fa fa-exclamation-triangle"></i> <span translate> FORM.WARNINGS.NO_CONFIGURATION_FOUND </span>  
+    </div>       
+
+    <ui-select
+      name="account_reference_id"
+      ng-model="$ctrl.accountReferenceId"
+      on-select="$ctrl.onSelect($item, $model)"
```
mbayopanda: If $model is not used, you could remove it

client/src/modules/templates/bhFeeCenterSelect.tmpl.html
```diff
@@ -0,0 +1,28 @@
+<div ng-form="FeeCenterForm" bh-fee-center-select ng-model-options="{ updateOn: 'default' }">
+  <div
+    class="form-group"
+    ng-class="{ 'has-error' : $ctrl.validateTrigger && FeeCenterForm.fee_center_id.$invalid }">
```
mbayopanda: Could you replace `$ctrl.validateTrigger` by `FeeCenterForm.$submitted`

client/src/modules/templates/bhFeeCenterSelect.tmpl.html
```diff
@@ -0,0 +1,28 @@
...
+      ng-required="$ctrl.required">
+      <ui-select-match placeholder="{{ 'FORM.SELECT.FEE_CENTER' | translate }}">
+        <span>{{$select.selected.label}}</span>
+      </ui-select-match>
+      <ui-select-choices ui-select-focus-patch repeat="fee_center.id as fee_center in $ctrl.feeCenters | filter: { 'label': $select.search }">
+        <span ng-bind-html="fee_center.label | highlight:$select.search"></span>
+      </ui-select-choices>
+    </ui-select>    
+
+    <div class="help-block" ng-messages="FeeCenterForm.fee_center_id.$error" ng-show="$ctrl.validateTrigger">
```
mbayopanda: Could you replace `$ctrl.validateTrigger` by `FeeCenterForm.$submitted`

client/src/modules/templates/bhFiscalPeriodSelect.tmpl.html
```diff
@@ -25,7 +25,7 @@
       class="form-control"
       ng-model="$ctrl.selectedPeriodFrom"
       name="periodFrom"
-      ng-options="periodFrom as periodFrom.hrLabel for periodFrom in $ctrl.periods"
+      ng-options="periodFrom as periodFrom.hrLabel for periodFrom in $ctrl.periods track by periodFrom.id"
```
mbayopanda, line 28: 👍

client/src/modules/templates/bhServicesMultipleSelect.tmpl.html
```diff
@@ -0,0 +1,29 @@
+<div ng-form="ServicesMultipleForm" bh-services-multiple-select ng-model-options="{ updateOn: 'default' }">
+  <div
+    class="form-group"
+    ng-class="{ 'has-error' : ServicesMultipleForm.$submitted && ServicesMultipleForm.services.$invalid }">
```
mbayopanda, line 4: 👍

mbayopanda (commented):
I made some comments about codes for the client side

server/controllers/finance/distributionFeeCenter/automatic.js
```diff
@@ -0,0 +1,74 @@
...
+
+  db.exec(sql, [transUuids])
+    .then((rows) => {
+      const dataToDistribute = [];
+      const userId = req.session.user.id;
+
+      rows.forEach((row) => {
+        data.forEach((item) => {
+          if (row.row_uuid === item.uuid) {
+            dataToDistribute.push([
```
mbayopanda, line 31: Is it possible to just use the `rows` array like : 
```js
rows.forEach(row => {
  dataToDistribute.push([
    ...
  ]);
});
```
Because it seems like the `rows` array contains already data defined by this condition : 
```js
if (row.row_uuid === item.uuid)
```

server/controllers/finance/distributionFeeCenter/configuration.js
```diff
@@ -0,0 +1,67 @@
...
+        rows.forEach(item => {
+          refAccounts.forEach(ref => {
+            if (ref.account_id === item.account_id) {
+              item.fee_center_id = ref.id;
+              item.fee_center_label = ref.label;
+            }
+          });
+
+          item.amount = item.credit ? item.credit : item.debit;
```
mbayopanda: Does this scenario can happen, having item.credit with 0 as value, and item.debit = undefined ?

Assume that, we just define value for the profit (`item.credit`) and this value is `0`, the code above will affect automatically `undefined` (`item.debit`) to `item.amount`.

server/controllers/finance/feeCenter.js
lomamech: @mbayopanda  use this sometimes

client/src/i18n/en/form.json
lomamech: The difference is in the French translation, one is masculine and the other is feminine
````
"PRINCIPAL": "Principale",
"PRINCIPAL_2": "Principal",
````

client/src/modules/distribution_center/distribution_center.js
lomamech: vm.loading is initially assigned by false at the beginning of the script
````
  vm.loading = false;
````
and it becomes true when calling this function

server/controllers/finance/distributionFeeCenter/automatic.js
lomamech: After analyzing the codes, the first element ```` rows ```` Get the data from the request, linking trans_uuid to billing for related services

and for the automatic distribution there is place of combined some elements of ```` data```` and ```` rows```` like ````account_id, is_cost, fee_center_id````

server/controllers/finance/distributionFeeCenter/configuration.js
lomamech: This scenario can only happen if in the table general ledger has a line with the field debit_equiv or credit_equiv with undefined values,

lomamech:
@jniles  @mbayopanda, I finished with the code review

lomamech:
@jniles @mbayopanda all things are Okey

jniles:
It looks like Larry is happy with the progress so far.  To avoid to many more conflicts, I'll merge this in.

bors r+

PR #3342, opened 2018-11-14
Add validation on service selection

### LINKED ISSUE #3244: Stock: Service selection should warn you when there are no services selected
If a user chooses to select a service for entry/exit, they should not be able to click the "submit" button until a service is selected.  Either that, or there should be an error when they click the "submit" button without a service.  See below:

stock - warn when no services in select

mbayopanda:
This PR resolve the issue #3244

close #3244


jniles (approved):
Tested and works.

bors r+

PR #3344, opened 2018-11-14
fix(transaction type): rm async/await from client

jniles:
The async/await is broken in production builds since AngularJS is not aware of it and uglify changes the scope of the compressed JS.  Unfortunately, we missed the transaction types async/await.  This commit removes it.


mbayopanda (approved):
LGTM 👍 

bors r+

PR #3345, opened 2018-11-14
chore(roles): remove project_id from roles

jniles:
This commit removes the `project_id` from the roles.  We don't have good error handling to inform the user what went wrong if they don't have the write project permissions, and we don't have the client-side interface
to give user roles permissions to projects.

Finalizes #2856.


jeremielodi (commented):
@jniles removing project_id is a good idea, normally the project_id column is
not useful.
This PR can be merged once all the tests pass.

jniles:
@jeremielodi thanks for the review!

PR #3346, opened 2018-11-14
refactor(Account References): rewrite tests

jniles:
This commit rewrites the account reference tests to better support the page object syntax.  It also makes it clear which account references is being clicked by using the GridRow syntax to select rows in the uiGrid.

### DIFF
--- a/test/end-to-end/account-reference/account-reference.page.js
+++ b/test/end-to-end/account-reference/account-reference.page.js
@@ -1,55 +1,42 @@
 /* global element, by */
+/* eslint class-methods-use-this:off */
 
 /**
- * This class is represents a accountReference page in term of structure and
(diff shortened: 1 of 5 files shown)


test/end-to-end/account-reference/account-reference.page.js
```diff
@@ -1,55 +1,42 @@
...
  * behaviour so it is a accountReference page object
  */
 
-/* loading grid actions */
-const GA = require('../shared/GridAction');
+const GridRow = require('../shared/GridRow');
 
-function AccountReferencePage() {
-  const page = this;
+class AccountReferencePage {
```
jeremielodi, line 11: 👍 using POO is a good idea

jeremielodi (commented):
This PR looks good to me

PR #3348, opened 2018-11-15
xlsx set boolean value

### LINKED ISSUE #3343: bug(General Ledger): curious console.log()s in Excel Export
When you export to Excel on the General Ledger, we get the following `console.log()`s.  These are clearly generated from the library, but it would be userful to know why they are made and how to suppress them or correct whatever mistake we are making.

See below:
curiousconsolelogs
_Fig 1: Curious Console Logs in GL Export to Excel_

jeremielodi:
closes #3343


jniles:
Looks fine to me.

bors r+

PR #3355, opened 2018-11-16
test(Job Titles): rewrite end to end tests

jniles:
This commit rewrites the Job Titles/Functions/Professions end to end tests so that they are simpler and easier to debug.  We've had a few errors with this test suite, so hopefully this helps solve the issue.


lomamech:
This is a complete refactoring

jniles:
Looks like this is failing due to https://github.com/IMA-WorldHealth/bhima-2.X/pull/3369.  We'll try this again.

bors r+

_Edit: looks like @lomamech beat me to it ;)_

PR #3356, opened 2018-11-16
test(weekend): rewrite weekend end to end tests

jniles:
The weekend tests have posed us issues in the past, but they have been updated for simplicity and clarity.  The grid uses GridRow to find entries by their labels and renames variables that were spelled in
correctly throughout the weekend configuration.  Additionally, the client-side has been updated to the latest ES2015+ standards.

### DIFF
--- a/client/src/modules/payroll/modals/payroll.modal.html
+++ b/client/src/modules/payroll/modals/payroll.modal.html
@@ -13,7 +13,7 @@
   </div>
 
   <div class="modal-body">
-    <div class="modal-body" ng-class="{ 'has-error' : PayrollModalForm.$submitted && PayrollModalForm.label.$invalid }">
+    <div class="form-group" ng-class="{ 'has-error' : PayrollModalForm.$submitted && PayrollModalForm.label.$invalid }">
       <label class="control-label" translate>FORM.LABELS.DESIGNATION</label>
       <input name="label" ng-model="PayrollConfigModalCtrl.payroll.label" autocomplete="off" class="form-control" required>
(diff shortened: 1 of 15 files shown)


client/src/modules/payroll/modals/payroll.modal.html
```diff
@@ -13,80 +13,66 @@
...
-      </bh-date-interval>
-    </div>
+    <bh-date-interval
+      label="FORM.LABELS.PERIODES"
+      date-id="periodes-date"
+      date-from="PayrollConfigModalCtrl.payroll.dateFrom"
+      date-to="PayrollConfigModalCtrl.payroll.dateTo"
+      required="true"
+      validation-trigger="PayrollModalForm.$submitted">
+    </bh-date-interval>
```
mbayopanda, line 31: 👍 good catch and more clean

test/end-to-end/payrollConfig/payroll_config.spec.js
```diff
@@ -6,7 +6,7 @@ const chai = require('chai');
 /** configuring helpers* */
 helpers.configure(chai);
 
-describe('Payroll Configuration Management', () => {
+describe.only('Payroll Configuration Management', () => {
```
mbayopanda: I don't think we need the .only, could you remove it ?

mbayopanda (commented):
Tests passes 👍 

Could you remove the `.only` in describe of the test and after that we will bors it

test/end-to-end/payrollConfig/payroll_config.spec.js
jniles: Whoops!  You are right.

PR #3357, opened 2018-11-16
feat(user registry) add roles column

### LINKED ISSUE #3333: Add column to the users grid interface with roles
It would be nice if we could have a column on the users grid.  This column would list at most two roles.  If there were more roles than two, the system would substitue the text _,+{{numRoles}} more_.  For example:

user | roles
-- | --
jniles | admin
mbayopanda | accountant, receptionist, +3 more
lomamech | superuser, admin

The rest of the roles would be found by clicking on "roles".

jeremielodi:
closes #3333

### DIFF
--- a/server/controllers/admin/users/index.js
+++ b/server/controllers/admin/users/index.js
@@ -81,12 +81,34 @@ function list(req, res, next) {
   const sql = 'SELECT user.id, display_name, user.username, user.deactivated FROM user;';
 
   db.exec(sql)
-    .then((rows) => {
-      res.status(200).json(rows);
+    .then((users) => {
+      return setRoles(users);
+    }).then(users => {
+
+      res.status(200).json(users);
     })
     .catch(next);
 }
 
+async function setRoles(users) {
+  const sql = `
+    SELECT ur.user_id, r.label
+    FROM user_role ur
+    JOIN role r ON r.uuid = ur.role_uuid`;
+
+  const userMap = {};
+  users.forEach(user => {
+    user.roles = '';
+    userMap[user.id] = user;
+  });
+
+  const roles = await db.exec(sql);
+  roles.forEach(role => {
+    userMap[role.user_id].roles += `${role.label}, `;
+  });
+
+  return users;
+}
 
 /**
  * @function detail
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -663,7 +663,8 @@
             "DECEMBER": "December",
             "PERMISSION_ASSIGNED_SUCCESS":"Permissions successfully affected",
             "EDIT_ROLE":"Assign role(s)",
-            "ROLES_FOR":"Role(s) for"
+            "ROLES_FOR":"Role(s) for",
+            "ROLES" : "Roles"
         },
         "LEGENDS": {
             "CREATION": "Creation",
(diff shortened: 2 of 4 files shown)


server/controllers/admin/users/index.js
```diff
@@ -81,12 +82,34 @@ function list(req, res, next) {
...
 
+async function setRoles(users) {
+  const sql = `
+    SELECT r.label
+    FROM user_role ur
+    JOIN role r ON r.uuid = ur.role_uuid
+    WHERE ur.user_id = ?`;
+
+  return q.all(users.map(r => db.exec(sql, r.id))).then(roles => {
+    users.forEach((user, index) => {
```
jniles: Does this really make an SQL query for every user in the database?

A more efficient way would be to do a LEFT JOIN on roles and then use [`GROUP_CONCAT()`](https://dev.mysql.com/doc/refman/5.5/en/group-by-functions.html#function_group-concat) (which https://github.com/IMA-WorldHealth/bhima-2.X/pull/3339 just taught me about).  Could you try that way?

It should be just a couple extra lines of code instead of a whole new function.

jniles (changes requested):

server/controllers/admin/users/index.js
jeremielodi: this function ```js setRoles``` is called in the list where every users are selected

jeremielodi: the setRoles is useful because it convert the roles array for a user into string

jeremielodi:
users grid bhi q

server/controllers/admin/users/index.js
jniles: We don't want to spin up a new database connection for each user.  That is a lot of overhead, and makes everyone else trying to use the application wait for their connections.  Why use N SQL queries when you can use 1?

jniles: For reference, at both IMCK and Vanga, we only allow 5 simultaneous database connections.

server/controllers/admin/users/index.js
```diff
@@ -81,12 +81,34 @@ function list(req, res, next) {
...
+    }).then(users => {
+
+      res.status(200).json(users);
     })
     .catch(next);
 }
 
+async function setRoles(users) {
+  const sql = `
+    SELECT ur.user_id, r.label
```
jniles, line 95: :+1:  Uses one SQL query

server/controllers/admin/users/index.js
```diff
@@ -81,12 +81,34 @@ function list(req, res, next) {
...
+    FROM user_role ur
+    JOIN role r ON r.uuid = ur.role_uuid`;
+
+  const userMap = {};
+  users.forEach(user => {
+    user.roles = '';
+    userMap[user.id] = user;
+  });
+
+  const roles = await db.exec(sql);
```
jniles, line 105: This code might be a bit cleaner by looping through the roles and grouping by user.  Something like this:
```js
const roles = await db.exec(sql);

users.forEach(user => {
  user.roles = roles
    .filter(role => role.user_id === user.id)
    .map(role => role.label)
    .join(', ');
});

return users;
```

jniles (approved):
Solves the issue and uses a single SQL query.  Well done. 👍 

bors r+

server/controllers/admin/users/index.js
joshuahedlund: A benefit of using .join is that it keeps a comma from trailing at the end of each Roles field.

joshuahedlund: FYI, I submitted the suggested update above as a commit. https://github.com/IMA-WorldHealth/bhima-2.X/commit/ed876e8f69c4a8f50552764e2a90135497012ee6

PR #3358, opened 2018-11-16
test(grades): rewrite grade management e2e test

jniles:
This commit rewrites the Grade Management end to end tests to use GridRow syntax and removes unused assertions.


jniles:

lomamech:
@jniles  i will

lomamech:
@jniles I have tested it in local, LGTM

PR #3359, opened 2018-11-16
Rewrite End to End tests for Rubrics, Holidays, Rubric Configuration, and IPR Tax

jniles:
This PR rewrites the tests for holiday management, rubric configuration, rubrics, and IPR taxes.  They all now use `GridRow()` to identify which row to click, instead of the GridAction/GridUtil, which was both inefficient and imprecise.  The goal is to reduce the number of flakey tests in `master`, or at least provide more information when they crash to debug.

All of these use page objects, with standard `create()`, `update()` and `remove()` methods.  A few use `count()` to check the number of rows in the grid.  We probably should do `count()` for all registries to check the number of rows in their grids.

Finally, a few modules have seen HTML improvements, including:
 1. Code lint fixes
 2. Better use of `headercrumb`.
 3. Migrating to `bhFilterToggle`
 4. Transforming `class='modal-body"` repeats into `class="form-group".
 5. Pruning unused ids/data-* tags.

This is following the current trend of refactoring our tests to be more standard, simpler, and more predictable.

### DIFF
--- a/client/src/modules/holidays/holidays.html
+++ b/client/src/modules/holidays/holidays.html
@@ -7,21 +7,15 @@
 
     <div class="toolbar">
       <div class="toolbar-item">
-        <button class="btn btn-default text-capitalize" id="create" ui-sref="holidays.create" data-method="create">
+        <button class="btn btn-default text-capitalize" ui-sref="holidays.create" data-method="create">
           <span class="fa fa-plus"></span> <span translate>HOLIDAY.ADD_HOLIDAY</span>
         </button>
       </div>
 
-      <div class="toolbar-item">
-        <button
-          type="button"
-          ng-click="HolidayCtrl.toggleFilter()"
-          data-method="filter"
-          class="btn btn-default"
-          ng-class="{ 'btn-info' : HolidayCtrl.filterEnabled }">
-          <i class="fa fa-filter"></i>
-        </button>
-      </div>
+      <bh-filter-toggle
+        on-toggle="HolidayCtrl.toggleFilter()">
+      </bh-filter-toggle>
+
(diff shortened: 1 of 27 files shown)


client/src/modules/holidays/holidays.html
```diff
@@ -7,31 +7,26 @@
...
-          ng-click="HolidayCtrl.toggleFilter()"
-          data-method="filter"
-          class="btn btn-default"
-          ng-class="{ 'btn-info' : HolidayCtrl.filterEnabled }">
-          <i class="fa fa-filter"></i>
-        </button>
-      </div>
+      <bh-filter-toggle
+        on-toggle="HolidayCtrl.toggleFilter()">
+      </bh-filter-toggle>
```
mbayopanda, line 17: 👍

test/end-to-end/rubrics/rubrics.page.js
```diff
@@ -1,48 +1,29 @@
...
-
-helpers.configure(chai);
-
-/* loading grid actions */
-const GA = require('../shared/GridAction');
-const GU = require('../shared/GridUtils');
+const GridRow = require('../shared/GridRow');
 const FU = require('../shared/FormUtils');
-const components = require('../shared/components');
+const { notification, accountSelect } = require('../shared/components');
```
mbayopanda, line 6: 👍

mbayopanda (approved):
The code looks good and there are a lot of enhancements, and also tests passes 👍 

bors r+

jniles:
Well..  Looks like I need to rewrite our Trial Balance tests to make sure this works.

PR #3362, opened 2018-11-19
Fix trailing comma on user roles concatenation

joshuahedlund:
Clean up building of user roles values


jniles (approved):
Thanks for your contribution @joshuahedlund!  It looks good to me.

bors r+

(Note: this was proposed in https://github.com/IMA-WorldHealth/bhima-2.X/pull/3357#discussion_r234480314)

PR #3370, opened 2018-11-21
feat(nativation) authentification

### LINKED ISSUE #3334: Invoice Registry accessible even if a user does not have the right to access it.
Our permissions are broken when navigating between Patient Registry -> Invoice Registry.  For some reason, a user with rights to only see the Patient Registry can see the Invoice Registry via the grid link.

See example:

_Gif 1: Escaping the permissions restrictions via the ui-grid link_

This might happen to more modules - it should be thoroughly investigated.

jeremielodi:
This PR helps to make sure the user is allowed to access a specific page on the client side
close #3334

### DIFF
--- a/client/src/js/app.js
+++ b/client/src/js/app.js
@@ -122,13 +122,13 @@ function startupConfig(
 
     const publicRoutes = ['/', '/settings', '/login', '/landing/stats', '/install'];
 
-    const isPublicPath = publicRoutes.indexOf(path) > -1;
+    const isPublicPath = publicRoutes.includes(path);
 
     // pass through
     if (!paths || isPublicPath) { return; }
 
     // check if the user is authorized to access this route.
-    const authorized = paths.some(checkUserAuthorization);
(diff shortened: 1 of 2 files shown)


jeremielodi:
I thought this is the right way to solve the client authentification problem but I realise that, in bhima, we have several type of urls. Some of the them are egal to the unit.path in the database but others are weird, like ``` url : /create```.  This last case are not eased to compare with unit.path

jeremielodi:
To my mind it will be good to add a column in unit table named  ```state_name```, it will contain the angular route state for each unit record. This will allow us to check if a user as access to every path. The last think to check will be the sub states, for those, role_action may be used

jniles:
@jeremielodi, that seems like a reasonable proposal.  I would remove the `/path` from the unit table if we are adding the state_name.  I also think the `url` field isn't used and you could remove that as well.

jniles:
Superseded by #3380.

PR #3374, opened 2018-11-21
fix(accounts): do not rebind account data

jniles:
This commit fixes the bindings with the AccountGrid store that was broken with the filter to hide accounts.

The problem came from rebinding the `vm.gridOptions.data` to a new array via the `Array.prototype.filter()` function.  By using `this.data.splice()` we perform the filtering in-place and do not break any bindings.

### DIFF
--- a/client/src/modules/accounts/AccountGrid.js
+++ b/client/src/modules/accounts/AccountGrid.js
@@ -144,5 +144,23 @@ function AccountGridService(AccountStore, Accounts, Store, Languages, $httpParam
     return this._store.get(id);
   };
 
+  AccountGrid.prototype.filterHiddenAccounts = function filterHiddenAccounts(showHiddenAccounts) {
+    const arr = this._store.data;
+
+    // either filter out hidden accounts or passthrough
+    const passthrough = () => true;
+    const filterFn = showHiddenAccounts
+      ? passthrough
+      : account => account.hidden === 0;
+
+    // we want to replace the whole array, so choose the max
+    // if we use a length that is too short, we will leave some behind
+    // if we pick a length that is greater, JS just expands the array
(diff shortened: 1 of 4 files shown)


client/src/modules/accounts/AccountGrid.js
```diff
@@ -144,5 +144,23 @@ function AccountGridService(AccountStore, Accounts, Store, Languages, $httpParam
...
+    // either filter out hidden accounts or passthrough
+    const passthrough = () => true;
+    const filterFn = showHiddenAccounts
+      ? passthrough
+      : account => account.hidden === 0;
+
+    // we want to replace the whole array, so choose the max
+    // if we use a length that is too short, we will leave some behind
+    // if we pick a length that is greater, JS just expands the array
+    const len = Math.max(arr.length, this.data.length);
```
jniles, line 159: I actually think this can be simply `this.data.length`.  The `splice()` function parameters is `startIndex`, `numOfElementsToRemove`, and then none or more elements to insert.  The number of elements to remove will always be the length of the data array - `this.data.length`.

jniles: I'll make the relevant changes to simplify the code.

client/src/modules/accounts/AccountGrid.js
```diff
@@ -144,5 +144,23 @@ function AccountGridService(AccountStore, Accounts, Store, Languages, $httpParam
...
+
+    // cheekily filters the data in place - thanks stack overflow
+    this.data.splice(0, len, ...arr.filter(filterFn));
```
mbayopanda, line 162: 👍 nicely done

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3380, opened 2018-11-26
client side authenticication

jeremielodi:
new approach for this pull #3370


jniles:
Looks like this got merged with another commit somehow.  Could you try and separate them out?

jeremielodi:
Yes I can

jeremielodi:
A this point, authentification is applied to 90% of bhima's routes.
10% left are sub routes (when using abstract state in angularJS) that we should rewrite in order to complete this hard work

jniles (commented):
So, this is an interesting approach.  It's on the right track, but I'm not sure the solution of making a bunch of abstract states is the best idea.  [Abstract states are a very specific concept](https://github.com/angular-ui/ui-router/wiki/Nested-States-%26-Nested-Views#abstract-states) in ui-router.  I'm not sure using them for authentication/authorization is the best way forward.

Using states instead of paths is excellent though 👍 .

I think the problem is that we have more states than we have paths.  We probably need to redesign our permissions to better model this, rather than trying to hack in a permissions structure that we will have to change eventually anyway.

mbayopanda:
@jeremy what is the state of this PR ?

kwilu:
Heads up @jeremielodi, we just merged some commits that conflict with the changes your made in this pull request! You can review this repository's [recent commits](https://github.com/IMA-WorldHealth/bhima/commits/master) to see where the conflicts occur. Please rebase your feature branch against the `upstream/master` branch and [resolve](https://zulip.readthedocs.io/en/latest/git/troubleshooting.html#recover-from-a-git-rebase-failure) your pull request's merge conflicts accordingly.

PR #3382, opened 2018-11-26
Refactor stock service

mbayopanda:
This PR move some code from StockService into the class StockFilterer.

The goal is to reduce complexity of filters for stock registries

### DIFF
--- a/client/src/modules/stock/StockFilterer.service.js
+++ b/client/src/modules/stock/StockFilterer.service.js
@@ -1,0 +1,145 @@
+angular.module('bhima.services')
+  .service('StockFilterer', StockFiltererService);
+
+StockFiltererService.$inject = [
+  'FilterService', 'appcache', 'PeriodService',
+  '$httpParamSerializer', 'LanguageService', 'bhConstants', 'util',
+];
+
+/**
+ * @class StockFilterer
+ */
+function StockFiltererService(Filters, AppCache, Periods, $httpParamSerializer, Languages, bhConstants, Util) {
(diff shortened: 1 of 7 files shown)


mbayopanda:
@jniles Could you review this PR, i need it to be merged for my Stock Assign PR

client/src/modules/stock/StockFilterer.service.js
```diff
@@ -0,0 +1,145 @@
...
+    {
+      key : 'expiration_date_from', label : 'STOCK.EXPIRATION_DATE', comparitor : '>', valueFilter : 'date',
+    },
+    {
+      key : 'expiration_date_to', label : 'STOCK.EXPIRATION_DATE', comparitor : '<', valueFilter : 'date',
+    },
+  ];
+
+  class StockFilterer {
+    constructor(cacheKey = 'stock-filterer-cache') {
```
jniles, line 45: Do you actually want to set a default `cacheKey`?  It seems like this should be required and throw an error if it doesn't exist.

jniles (approved):
Looks fine to me.  Some unit tests would be nice in the future, so that we can easily tell if it works.

client/src/modules/stock/StockFilterer.service.js
mbayopanda: Since this Class is for Stock, if we omit to specify the `cacheKey` the class will just use this default `cacheKey` of stock for its appcache instance.

Another thing is we must be able to use different caches for stock registries.

PR #3388, opened 2018-11-26
Puppeteer wrapper for pdf rendering

mbayopanda:
This PR add puppeteer for pdf rendering as replacement of wkhtmltopdf.


jniles:
I'm kind of curious why the `setContent()` requires styles and scripts bundled separately.  I put together a little test script myself - see how it works on your machine.  The script is below in full:
```js
const pptr = require('puppeteer-core');

const tmpl = `
  <!doctype html>
  <html>
    <head>
      <title>Test</title>
      <style>
        .red {
          color : red;
        }
      </style>
    </head>
    <body>
      <p class="red">Is this okay?</p>
    </body>
      <script>
        setInterval(() => {
          const body = document.querySelector('body');
          body.appendChild(document.createTextNode('This was added programmatically!'));
          body.appendChild(document.createElement('br'));
        }, 3000);
      </script>
  </html>
`;

(async () => {
  try {
    const browser = await pptr.launch({ executablePath: '/usr/bin/chromium-browser', headless: false });
    const page = await browser.newPage();
    page.setContent(tmpl);

    console.log('Check out the page');
  } catch (e) {
    console.log('e:', e);
  }
})();
```
You will need to change the `excuteablePath` (I didn't want to download chromium again).

It produces this result:


It might just be something amiss with your setup that the script tags and rendering tags don't work as expected.  We can debug this when we have a chance to look over the same code base in person.

jniles:
Another advantage with this method of rendering is that we can function more seamlessly on ARM devices. WKHTMLTOPDF on ARM doesn't exist, and all ports we've found suffer from webkit/qt rendering issues.

mbayopanda:
@jniles could you review this PR again, please.

package.json
```diff
@@ -114,6 +114,8 @@
     "mysql": "^2.14.0",
     "mz": "^2.7.0",
     "ngstorage": "gsklee/ngStorage#0.3.11",
+    "puppeteer": "^1.10.0",
+    "puppeteer-core": "^1.10.0",
```
jniles: Do we need both puppeteer and puppeteer-core?

server/lib/ReportManager.js
```diff
@@ -236,4 +235,12 @@ class ReportManager {
   }
 }
 
+// function to encode file data to base64 encoded string
+function base64Encode(file) {
+  // read binary data
+  const bitmap = fs.readFileSync(file);
```
jniles: This might be part of the problem with the speed - we don't want to write synchronous code on the server!  Can you make this into an asynchronous operation?

server/lib/renderers/pdf.js
```diff
@@ -2,96 +2,160 @@
...
+// load scripts and styles
+const { dbStyles, dbScripts } = environmentStyleAndScripts();
+
+// pptr options
+const pptrOptions = {
+  headless : true,
+  args : [
+    '--no-sandbox',
+    '--disable-web-security',
+    '--disable-gpu',
```
jniles: Out of curiosity, do we actually need all these flags?  What happens if we remove them?

server/lib/renderers/pdf.js
```diff
@@ -2,96 +2,160 @@
...
+  }
+
+  // include scripts
+  for (let i = 0; i < dbScripts.length; i++) {
+    const content = dbScripts[i];
+    // eslint-disable-next-line no-await-in-loop
+    await page.addScriptTag({ content : content.toString() });
+  }
+
+  const pdf = await page.pdf(options);
```
jniles, line 98: Let's just `return page.pdf(options);`

server/lib/renderers/pdf.js
```diff
@@ -2,96 +2,160 @@
...
+async function pdfGenerator(htmlString, options = {}) {
+  const browser = await browserPromise;
+  const page = await browser.newPage();
+  await page.setContent(htmlString);
+
+  // include styles
+  for (let i = 0; i < dbStyles.length; i++) {
+    const content = dbStyles[i];
+    // eslint-disable-next-line no-await-in-loop
+    await page.addStyleTag({ content : content.toString() });
```
jniles: Since we switched from using `path:` to using `content:`, we now control the order of these scripts.  I think you can write this more efficiently by simply:
```js
await Promise.all(dbStyles);
```

Note that this will only work if you have called the `readFile` operation with 'utf8' to avoid needing to call the `.toString()` method.

server/lib/template/partials/enterpriseDetails.handlebars
```diff
@@ -1,12 +1,13 @@
 {{#if metadata.enterprise.logo}}
   <div class="col-xs-1">
-    <img src="{{absolutePath}}/{{slashed metadata.enterprise.logopath}}" width="65" height="65" alt="">
+    {{!-- <img src="{{absolutePath}}/{{slashed metadata.enterprise.logopath}}" width="65" height="65" alt=""> --}}
+    <img src="data:image/jpg;base64,{{metadata.enterprise.binaryLogo}}" width="65" height="65" alt="">
```
jniles, line 3: One thing to consider ... this is always `image/jpg`.  Should we limit all uploads to be `jpg` or should we try to convert pngs -> jpgs, or is there a way to figure out the image type ...?

server/lib/renderers/pdf.js
```diff
@@ -2,96 +2,160 @@
...
+}
+
+function handleOption(options) {
+  options.landscape = !!((options.orientation && options.orientation === 'landscape'));
+  return options;
+}
+
+function environmentStyleAndScripts() {
+  const styles = resources.cssPaths.map(cssPath => {
+    return fs.readFileSync(cssPath);
```
jniles: Can we do this in asynchronously?  Also, these should read the files as 'utf8'.

I propose that we use the `mz/fs` (which is already used in the backup scripts) to make this into a promise.  so that we are basically doing:
```js
function environmentStyleAndScripts() {
  const styles = resources.cssPaths.map(path => fs.readFile(path, 'utf8'));
  const scripts = resources.jsPaths.map(path => fs.readFile(path, 'utf8'));
  return { styles, scripts };
}
```

... then later you will be able to do:
```js
const stylesheets = await Promise.all(styles);
```

server/lib/renderers/pdf.js
```diff
@@ -2,96 +2,160 @@
...
+    '--disable-web-security',
+    '--disable-gpu',
+    '--hide-scrollbars',
+    '--disable-setuid-sandbox',
+  ],
+};
+const browserPromise = pptr.launch(pptrOptions)
+  .then(brwsr => {
+    // notify about the pdf service
+    debug('PDF service is running');
```
jniles: 👍

jniles (commented):
@mbayopanda this is an awesome start!

We have a lot of synchronous code introduced, which will slow NodeJS down.  Let's see if we can deal with that by loading up our files asynchronously.  It's not critical for those files loaded at the start of the process, but it is better practice.

server/lib/template/partials/enterpriseDetails.handlebars
mbayopanda: It seems like pngs are also handled by `image/jpg`

jniles: :dancer:

mbayopanda:
@jniles could you review it again please

.travis.yml
```diff
@@ -3,6 +3,8 @@ dist: trusty
 
 before_install:
   - jdk_switcher use oraclejdk8
+  - "curl -o- -L https://yarnpkg.com/install.sh | bash -s --"
```
jniles: :+1:

client/src/js/components/bhRenderOption.js
```diff
@@ -26,6 +26,8 @@ function RenderOptionController() {
   this.$onInit = function $onInit() {
     ctrl.orientation = ctrl.orientation || 'portrait';
     ctrl.onChange = ctrl.onChange || angular.noop();
-    ctrl.setOrientation = () => ctrl.onChange({ orientation : ctrl.orientation });
+    ctrl.setOrientation();
   };
+
+  ctrl.setOrientation = () => ctrl.onChange({ orientation : ctrl.orientation });
```
jniles, line 32: This is a good change.  👍

client/src/modules/templates/bhRenderOption.tmpl.html
```diff
@@ -2,13 +2,13 @@
   <label class="control-label" translate>FORM.LABELS.PRINT_ORIENTATION</label>
   <div>
     <label class="radio-inline">
-      <input type="radio" name="portrait" id="portrait"  value="portrait" ng-click ="$ctrl.setOrientation()" ng-model="$ctrl.orientation" required>
+      <input type="radio" name="portrait" id="portrait"  value="portrait" ng-change ="$ctrl.setOrientation()" ng-model="$ctrl.orientation" required>
```
jniles, line 5: 💃

gulpfile.js
```diff
@@ -32,13 +32,24 @@ const supportedLanguages = {
...
   client : {
+    // this section contains custom js scripts of plugins
+    plugins : [
+      'client/src/js/plugins/*.js',
+    ],
     javascript : [
       'client/src/js/define.js',
       'client/src/js/app.js',
       'client/src/**/*.js',
+      '!client/src/js/plugins/*.js',
```
jniles, line 50: Do barcodes still render on the client-side in receipts with HTML format?

package.json
```diff
@@ -123,8 +124,7 @@
     "uuid": "^3.1.0",
     "uuid-parse": "^1.0.0",
     "webcam": "jonashartmann/webcam-directive#3.1.2",
-    "winston": "^3.1.0",
-    "wkhtmltopdf": "^0.3.3"
```
jniles: 💃

package.json
```diff
@@ -114,6 +114,7 @@
     "mysql": "^2.14.0",
     "mz": "^2.7.0",
     "ngstorage": "gsklee/ngStorage#0.3.11",
+    "puppeteer": "^1.10.0",
```
jniles: Have you tried using puppeteer-core on windows?  Does it work?

server/lib/renderers/pdf.js
```diff
@@ -2,96 +2,160 @@
...
-      const pdfStream = wkhtmltopdf(htmlStringResult, pdfOptions);
+async function pdfGenerator(htmlString, options = {}) {
+  let browser = await browserPromise;
 
-      // this promise will only be resolved once the stream 'end' event is fired - with this implementation this will
-      // not allow the client to receive chunks of data as they are available
-      return streamToPromise(pdfStream);
-    });
+  // check if browser is closed
+  browser._process.once('close', () => {
```
jniles: Looks like this might have a memory leak :)

jniles (commented):
Just a few clarification comments.  Looks like Travis is giving us trouble for some reason :/

package.json
mbayopanda: No, it doesn't works, it require to download chrome : 
image

`puppeteer-core` requires also the path to chromium executable : 
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteer-vs-puppeteer-core

server/lib/template/partials/enterpriseDetails.handlebars
```diff
@@ -1,12 +1,12 @@
...
   <div class="col-xs-1">
-    <img src="{{absolutePath}}/{{slashed metadata.enterprise.logopath}}" width="65" height="65" alt="">
+    <img src="data:image/jpg;base64,{{metadata.enterprise.binaryLogo}}" width="65" height="65" alt="">
   </div>
 {{/if}}
 <div class="col-xs-6">
   <h3 style="margin: 0px;">{{metadata.enterprise.name}}</h3>
   <strong>{{metadata.project.name}} ({{metadata.project.abbr}})</strong> <br>
-  <span class="text-capitalize">{{translate 'FORM.LABELS.ADDRESS'}}</span>: {{enterprise.location}} <br>
+  <span class="text-capitalize">{{translate 'FORM.LABELS.ADDRESS'}}</span>: {{metadata.enterprise.location}} <br>
```
jniles, line 9: :dancer:

jniles (commented):
So...

It looks like my advice was wrong.  According to [this blog post](https://docs.browserless.io/blog/2018/06/04/puppeteer-best-practices.html), they don't want you to spin up a browser unless it is absolutely necessary and to close it each time.

Everything I've seen seems to indicate the same thing.  Let's see what happens if we change course and try to get it working by just rendering with the browser each time...

jniles:
@mbayopanda, I think I may have gotten this working.  With a bit more cleanup, I think we can probably start doing rigorous tests on it.

The main issue was that nodejs wasn't shutting down since chromium would never be closed.  We have this same problem with MySQL as well when a stream is open and not closed properly.  However, given the article I cited above, instead of figuring out how to listen for an exit event and closing Chrome, I opted to just spin up a chrome instance per request. 

In the long run, we'll need to figure out some rate limiting to make sure people don't accidentally (or purposely) overwhelm the server.  I'm not quite sure how to handle that just yet, but we'll think of something 👯‍♂️

jniles:
Okay, the steps to merge this:
 1. Let's build this on a Raspberry Pi and see if it solves that problem.  That was my main goal for it.
 2. ~~We need to rewrite the unit tests to use puppeteer.  That way, we have an idea of changes between different chrome versions.~~
 2. Let's see if we can modify this to use `puppeteer-core` instead of `puppeteer`.  That way, we don't have to download a whole chrome browser at every installation site :(
 3. We should see if we can take some memory profiles and see how affects the memory consumption of the system.
 5. Do parallel requests leak into each other?

jniles:
@kwilu claim

jniles (commented):
@mbayopanda, can you give this a test on your machine to make sure it works on a windows machine?

kwilu:

kwilu:

PR #3389, opened 2018-11-26
Fix annual clients report rendering

### LINKED ISSUE #3383: Allow users to specify currency for annual clients report
The Annual Clients Report (`debtorBalanceReport`) needs a refresh.  There is a bug in it which only renders the row values in the enterprise currency and not the footer values.  It would be nice to have the user be able to select the currency like the balance report and use that currency to render the report.  We can fix the bug and add the feature at the same time.

annualclientsreportbug
_Fig 1: An Example of the Bug in Action_

### LINKED ISSUE #3384: Merge the Clients Report and the Annual Clients Report
We have two reports - the annual clients report and the clients report.  They both contain the same values except that the annual clients report is much faster since it leverages period totals.  Interestingly, for Vanga, the two reports are identical.  However, for IMCK, they contain different values.  We need to trace the root cause of this problem, then merge the reports into a single report.  It's confusing to have both reports.

jniles:
This commit fixes a bug in the Annual Clients Report related to the
currency rendering. All rows now use the correct currency id.

Partially addresses #3383.

This commit adds the ability to show the client report in multiple
currencies using the exchange rate of the day.  This gives further
flexibility to the report.

It also fixes a bug with the debits and credits, ensuring they respect
the correct date ranges.

Closes #3383.

This commit merges all client reports into a single report called the
Annual Clients Report.  This both reduces code and confusion since both
existing reports produced the same information in a slightly different
way with vastly different performance characteristics.

Closes #3384.


jniles:
@mbayopanda, can this get reviewed?

server/models/bhima.sql
```diff
@@ -89,7 +87,7 @@ INSERT INTO unit VALUES
   (195, 'Role management','TREE.ROLE_MANAGEMENT','Roles Management', 1,'/modules/role/','/roles'),
   (196, 'Depot Registry','TREE.DEPOTS_REGISTRY','',20,'/modules/depots/','/depots'),
   (197, 'Stock Exit Report','TREE.STOCK_EXIT_REPORT','Stock Exit Report', 144,'/modules/reports/generated/stock_exit','/reports/stock_exit'),
-  (199, 'Client Debtor Account Balance', 'REPORT.CLIENT_DEBTOR_ACCOUNT_BALANCE_REPORT', 'Client Debtor Account Balance', 144, '/modules/reports/debtor_accounts_balance', '/reports/debtorBalanceReport'),
+  (199, 'Annual Clients Report', 'REPORT.CLIENTS.TITLE', 'Annual Clients Report', 144, '/modules/reports/clients', '/reports/annual-clients-report'),
```
mbayopanda, line 90: Is it possible to add this update in the migration file ?

server/models/bhima.sql
```diff
@@ -141,15 +139,14 @@ INSERT INTO `report` (`id`, `report_key`, `title_key`) VALUES
...
-  (9, 'clients_report','REPORT.CLIENTS'),
   (10, 'aged_creditors','TREE.AGED_CREDITORS'),
   (11, 'balance_sheet_report', 'REPORT.BALANCE_SHEET.TITLE'),
   (12, 'cash_report', 'REPORT.CASH_REPORT.TITLE'),
   (13, 'inventory_report', 'REPORT.STOCK.TITLE'),
   (14, 'inventory_file', 'REPORT.STOCK.INVENTORY_REPORT'),
   (15, 'operating', 'TREE.OPERATING_ACCOUNT'),
   (16, 'stock_exit', 'REPORT.STOCK.EXIT_REPORT'),
-  (17, 'debtorBalanceReport', 'REPORT.CLIENT_DEBTOR_ACCOUNT_BALANCE_REPORT'),
+  (17, 'annual-clients-report', 'REPORT.CLIENTS.TITLE'),
```
mbayopanda, line 149: This update can also be in the migration file

server/controllers/finance/reports/debtors/annual-clients-report.handlebars
```diff
@@ -30,10 +36,10 @@
...
               <td>{{row.accountNumber}}</td>
               <td>{{row.groupName }}</td>
-              <td class="text-right">{{currency row.openingBalance ../metadata.enterprise.currency_id}}</td>
-              <td class="text-right">{{currency row.debit  ../metadata.enterprise.currency_id}}</td>
-              <td class="text-right">{{currency row.credit  ../metadata.enterprise.currency_id}}</td>
-              <td class="text-right">{{currency row.closingBalance ../metadata.enterprise.currency_id}}</td>
+              <td class="text-right">{{debcred row.openingBalance row.currencyId}}</td>
+              <td class="text-right">{{debcred row.debit row.currencyId}}</td>
+              <td class="text-right">{{debcred row.credit row.currencyId}}</td>
+              <td class="text-right">{{debcred row.closingBalance row.currencyId}}</td>
```
mbayopanda, line 44: Is it possible to have also the movement column here, just after the credit column : 
```html
<td class="text-right">{{debcred row.movement  row.currencyId}}</td>
```

mbayopanda (commented):
I have gotten a warning message when i use the enterprise currency, i think this warning message must be showed only if the selected currency is not the enterprise currency.

image

server/models/bhima.sql
jniles: Will do!

jniles:
@mbayopanda, changes made!

annualdebtorsreport

mbayopanda:
Is it correct that the footer of the difference column is empty : 

image

jniles:
Yeah, I did that on purpose.  Not sure why though.  Do you think it should be filled in?

jniles:
@mbayopanda, I added the footer :)

mbayopanda:
LGTM 👍 

bors r+

jniles:
bors r-

It seems like this might be an issue with a merge.  Let me follow it up.

jniles:
Figured it out. 

bors r+

PR #3390, opened 2018-11-27
feat(price list) export as pdf, excel, csv

### LINKED ISSUE #3349: Add exporters for price lists
The Price List module lacks the options to export the grid like our other modules.  See below for the current presentation of the Price List module:
needexportersforpricelist
_Fig 1: Current Price List Menu_

It would be nice to have the following options in the price list menu:
 1. Export to Excel
 2. Export to CSV
 3. Export to PDF.

jeremielodi:
closes #3349

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -5,7 +5,7 @@ PORT=8080
 DB_HOST='localhost'
 DB_USER='bhima'
 DB_PASS='HISCongo2013'
-DB_NAME='hbbdata'
+DB_NAME='bhima_test'
 
 # session variables
 SESS_SECRET='XopEn BlowFISH'
--- a/client/src/modules/prices/prices.html
+++ b/client/src/modules/prices/prices.html
@@ -31,6 +31,23 @@
                 <i class="fa fa-close"></i> <span translate>FORM.BUTTONS.CLEAR_GRID_CONFIGURATION</span>
               </a>
             </li>
+
+            <li role="separator" class="divider"></li>
+            <li role="menuitem">
+              <a data-method="print" id="print" href="/prices/download/list?{{PriceListCtrl.download('pdf')}}"  translate-attr="{'download': 'DOWNLOADS.PDF'}">
+                <i class="fa fa-print"></i> <span translate>FORM.BUTTONS.PRINT</span>
+              </a>
+            </li>
+            <li role="menuitem">
+              <a data-method="export" id="export" href="/prices/download/list?{{PriceListCtrl.download('csv')}}" translate-attr="{'download': 'DOWNLOADS.CSV'}" >
+                <i class="fa fa-file-excel-o"></i> <span translate>DOWNLOADS.CSV</span>
+              </a>
+            </li>
+            <li role="menuitem">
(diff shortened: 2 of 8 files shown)


.env.development
```diff
@@ -5,7 +5,7 @@ PORT=8080
 DB_HOST='localhost'
 DB_USER='bhima'
 DB_PASS='HISCongo2013'
-DB_NAME='hbbdata'
+DB_NAME='bhima_test'
```
jniles, line 8: 💃

client/src/modules/prices/prices.html
```diff
@@ -31,6 +31,30 @@
                 <i class="fa fa-close"></i> <span translate>FORM.BUTTONS.CLEAR_GRID_CONFIGURATION</span>
               </a>
             </li>
+
+            
+            <li role="menuitem">
+              <a href data-method="clear-state" ng-click="PriceListCtrl.clearGridState()">
+                <i class="fa fa-close"></i> <span translate>FORM.BUTTONS.CLEAR_GRID_CONFIGURATION</span>
```
jniles: It looks like you repeated this option?   Isn't it the same button as on line 31?

client/src/modules/prices/prices.js
```diff
@@ -135,11 +134,17 @@ function PriceListController(
...
+  function download(renderer) {
+    const displayNames = columnConfig.getDisplayNames();
     const options = {
-      renderer : 'pdf',
+      renderer,
       lang : Languages.key,
+      detailed : 1,
     };
+
+    if (renderer === 'xlsx') {
```
jniles: We should probably rename keys if option is not `pdf`.  CSV will want renamed keys as well.

client/src/modules/prices/prices.service.js
```diff
@@ -72,5 +73,10 @@ function PriceListService(Api) {
       .then(service.util.unwrapHttpResponse);
   }
 
+  function download(params) {
+    const url = service.url.concat('download/list');
+    return service.$http.post(url, params)
```
jniles: It looks like this will send a `POST` request, but the download list will receive a `GET` request.  If you are running into bugs, this might be where it is coming from.

server/controllers/finance/priceList.js
```diff
@@ -321,3 +322,35 @@ exports.delete = function del(req, res, next) {
 function isEmptyObject(object) {
   return Object.keys(object).length === 0;
 }
+
+exports.download = (req, res, next) => {
```
jniles: Most other controllers have a specialized controllers under `finance/reports`.  It makes it easier to read and ensure we don't have really long files.  Would it be possible to move this code under reports?

jniles (commented):
A few changes to make here.  Let me know if you have any questions.

client/src/modules/prices/prices.js
jeremielodi: okay

client/src/modules/prices/prices.html
jeremielodi: you're right! let me remove it then.

jniles (approved):
Looks good to me!  Good work.

bors r+

PR #3391, opened 2018-11-27
feat(client report) ignore closed accounts option

### LINKED ISSUE #3386: Allow users to filter out locked clients from the clients report
At IMCK, they will close some client accounts since they assume they will never pay.  We could add an option to the clients report to ignore closed accounts.  This might be useful to them.

jeremielodi:
closes #3386


jniles:
It looks like this got blown apart by the combination of the annual clients report (https://github.com/IMA-WorldHealth/bhima-2.X/pull/3389).  It's probably easier to just start over :/

PR #3405, opened 2018-11-29
Tests rewrite employees config end to end tests

jniles:
This commit rewrites the end to end tests for the payroll employee configuration
page.  These tests now use the GridRow method of testing for better
precision.


jniles:

PR #3410, opened 2018-11-29
fix(stock): respect stock lots registry groups

jniles:
This commit fixes the stock lots registry grouping by ordering the inventory items by depot first.

Closes #3409.

The problem fixed here is that ui-grid wasn't grouping values properly unless they were sorted first.  It's not clear to me why this happens - the documentation doesn't mention it.

In the long run, we should probably deprecate GridGrouping service, and use only TreeBase for our grouping.  This will give us a more uniform way to interact with ui-grid.

Finally, we also need to rewrite the client sort to be done in the server.  The Stock Lots API is overloaded and can be done more efficiently in SQL.

**Before**
stocklotsoriginal

**After**
stocklotsafter

It also makes these changes for the stock inventories and stock movements registries.

### DIFF
--- a/client/src/modules/stock/lots/registry.js
+++ b/client/src/modules/stock/lots/registry.js
@@ -29,122 +29,95 @@ function StockLotsController(
   ];
 
   // grid columns
-  const columns = [
-    {
-      field : 'depot_text',
-      displayName : 'STOCK.DEPOT',
-      headerCellFilter : 'translate',
-    },
-
-    {
-      field : 'code',
-      displayName : 'STOCK.CODE',
-      headerCellFilter : 'translate',
-    },
-
-    {
-      field : 'text',
-      displayName : 'STOCK.INVENTORY',
-      headerCellFilter : 'translate',
-    },
-
-    {
-      field : 'group_name',
-      displayName : 'TABLE.COLUMNS.INVENTORY_GROUP',
-      headerCellFilter : 'translate',
-    },
-
-    {
-      field : 'label',
-      displayName : 'STOCK.LOT',
-      headerCellFilter : 'translate',
-    },
-
-    {
-      field : 'quantity',
-      displayName : 'STOCK.QUANTITY',
-      headerCellFilter : 'translate',
-    },
-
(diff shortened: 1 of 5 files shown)


jniles:
@mbayopanda, try to deploy this branch online and see if it works with the latest IMCK dataset.  I tested it, but you can't be too careful!

client/src/modules/stock/lots/registry.js
```diff
@@ -245,8 +221,15 @@ function StockLotsController(
 
     Stock.lots.read(null, filters)
       .then((lots) => {
+
+        // FIXME(@jniles): we should do this ordering on the server via an ORDER BY
+        lots
+          .sort(orderByDepot);
```
mbayopanda, line 227: Could you add this change also in the stock/inventory and stock/movement registries.

client/src/modules/stock/lots/registry.js
```diff
@@ -235,6 +207,10 @@ function StockLotsController(
     vm.loading = !vm.loading;
   }
 
+  function orderByDepot(rowA, rowB) {
+    return rowA.depot_text > rowB.depot_text ? 1 : -1;
```
mbayopanda, line 211: This is a non-blocking comment, in terms of performances, if the use of $timeout works what can be acceptable in the of performances : sort data or use of $timeout ?

mbayopanda (approved):

client/src/modules/stock/lots/registry.js
jniles: Timeout didn't work for me.  Did it work for you?

jniles:
@mbayopanda, thanks for the review.  I've updated the changes on both registries.

client/src/modules/stock/lots/registry.js
jniles: I am indifferent to whether we use sorting or $timeout in the short run.  In the long run, I think we probably need to sort on the server in SQL first, then we'll figure out what is going on with the grouping issues :/

PR #3419, opened 2018-11-30
fix sort order in the lot registry

mbayopanda:
This PR fix the sorting order in the lot registry for correct lifetime risk calculation.

Before : 
The second row must appear at the top so that we can easily interpret columns : duree lot, risk, ...
image

After : 
image


jniles (approved):
Looks good to me.  Tested it online and it seems like the previous sort issue hasn't returned.

bors r+

PR #3421, opened 2018-12-01
Chore clean up validation triggers

### LINKED ISSUE #3134: Replace validationTrigger binding in components with child form$submitted

jniles:
This commit replaces the validateTrigger/validationTrigger across the entire application with listeners on the parent form.  This simplifies things for developers and new contributors.

See #3134 for why this is possible now and the motivation behind it.

Closes #3134.


client/src/modules/templates/bhAccountConfigSelect.tmpl.html
```diff
@@ -1,32 +1,34 @@
-<div ng-form="{{ $ctrl.name }}" bh-account-config-select ng-model-options="{ updateOn: 'default' }">
+<div ng-form="AccountConfigForm" bh-account-config-select ng-model-options="{ updateOn: 'default' }">
```
mbayopanda, line 1: 👍

client/src/modules/templates/bhUserSelect.tmpl.html
```diff
@@ -1,23 +1,26 @@
...
+      <ui-select-match placeholder="{{ 'FORM.SELECT.USER' | translate }}">
+        <span>{{$select.selected.display_name}}</span>
+      </ui-select-match>
       <ui-select-choices ui-select-focus-patch repeat="user.id as user in $ctrl.users | filter: { 'display_name': $select.search }">
         <span ng-bind-html="user.display_name | highlight:$select.search"></span>
       </ui-select-choices>
     </ui-select>
 
-    <div class="help-block" ng-messages="UserForm.user_id.$error" ng-show="UserForm.$$parent.$submitted">
+    <div class="help-block" ng-messages="UserForm.user_id.$error" ng-show="UserForm.$submitted">
```
mbayopanda, line 23: 👍

mbayopanda (approved):
LGTM 👍

PR #3428, opened 2018-12-03
fix(users): fix user name validation on update

### LINKED ISSUE #3402: Username unique validator blocks updates to the user details
On the user edit page, you can't change any of the user details since the async validator for unique names fires on entry to the modal.  Here it is in action:
donotvalidateupdateusers
_Fig 1: Why is username automatically an error?_

We need to fix this by not validating the uniqueness of the username until it has been touched.  This will allow use to update other fields if the username is constant as well as check if the username is breaking the uniqueness constraint before submission.

jniles:
This commit fixes the username validation when the user exists already.
It should not validate the user's own username against itself.  Ideally,
we'll build this property into the unique component in the future.

Closes #3402.

### DIFF
--- a/client/src/js/components/bhProjectSelect.js
+++ b/client/src/js/components/bhProjectSelect.js
@@ -16,7 +16,6 @@ ProjectSelectController.$inject = [
 
 /**
  * Project Select component
- *
  */
 function ProjectSelectController(Projects, Notify) {
   const $ctrl = this;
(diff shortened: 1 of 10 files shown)


jniles:

jniles:
@jeremielodi, this is ready for a final review.

mbayopanda (approved):
LGTM :+1

bors r+

Issue #3202, opened 2018-09-27
Stock Assignation module for assigning lots to person or organization


mbayopanda:
This issue is about to add a module for assigning stock lots to person so that we can know where is each lots supposed to be in a depot.

- The assignment is based on lot (stock lot), we can have a registry like : 
![assign grid](https://user-images.githubusercontent.com/5445251/48540633-e8143680-e8ba-11e8-95e8-b0c9f8c9c02d.png)

- The modal for adding a new assignment can be : 
![assing modal add](https://user-images.githubusercontent.com/5445251/48540656-fbbf9d00-e8ba-11e8-8bdf-b52676c6508b.png)

- The modal for editing an assignment can be : 
![assign modal edit](https://user-images.githubusercontent.com/5445251/48540672-09752280-e8bb-11e8-8af2-e60f73799f4f.png)

By adding this feature we must update the lot table for having a flag for assignment which will help us to do not consider assigned lots for stock exits.

@jniles what do you think about that ?

jniles:
I think this has already been done.  Is that correct @mbayopanda ?

jniles:
Confirmed offline!

PR #3429, opened 2018-12-03
Stock Assignment Module

mbayopanda:
This PR add the stock assignment feature in bhima.


client/src/modules/stock/assign/modals/action.modal.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /**
+   * @function groupBy
+   * @description group an array of objects according a property
+   * @param {array} array
+   * @param {string} property
+   * @returns {object}
+   */
+  function groupBy(array, property) {
```
jniles: This looks like it would be useful in other modules.  Can you move it into `services/util.js`?  We also should write a unit test for it that shows what it does.

It looks like you already use `util` in this controller, so it should work well.

client/src/i18n/en/stock.json
```diff
@@ -173,5 +173,17 @@
   },
   "INVENTORIES" : {
     "REGISTRY" : "Inventories Registry"
+  },
+  "ASSIGN" : {
+    "TITLE" : "Assignment",
+    "STOCK_ASSIGN" : "Stock Assignment",
+    "EDIT_ASSIGN" : "Edit Assignment",
+    "REMOVE_ASSIGN" : "Remove Assignment",
+    "CREATE_SUCCESS" : "Successfull created",
```
jniles: These are missing a `y`.  It should read `Successfully created`.

test/end-to-end/stock/stock.assign.spec.js
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  const record2 = {
+    depot : 'Depot Secondaire',
+    inventory : 'Quinine',
+    lot : 'VITAMINE-B',
+    entity : 'Wayne Enterprise',
+  };
+
+  it('Create a new stock assignment', () => {
```
jniles, line 33: This are nice looking tests :+1:

server/models/migrations/v0.8.0-v0.9.0/migrate.sql
```diff
@@ -964,3 +964,35 @@ WHERE id = 199;
...
+
+-- author: @mbayopanda
+-- stock assignment feature
+-- the stock assign table
+DROP TABLE IF EXISTS `stock_assign`;
+CREATE TABLE `stock_assign` (
+  `uuid`              BINARY(16) NOT NULL,
+  `lot_uuid`          BINARY(16) NOT NULL,
+  `entity_uuid`       BINARY(16) NOT NULL,
+  `depot_uuid`        BINARY(16) NOT NULL,
```
jniles, line 985: Is the `depot_uuid` where the stock has come from when it is assigned?

jniles (commented):
@mbayopanda code looks clear, and it seems like you respected the design in #3202  Well done.

I'm curious why we need an `is_assigned` column.  Can't we just check for the presence of the `uuid` in the `stock_assign` table?

It seems to me the confusions comes from this:  If the item is a "unique item" (e.g. a lenovo computer), it should not be assigned to more than one entity at a time.  However, if the item is a box of five VGA cables, you can imagine that multiple people might have those VGA cables, and the lots could be assigned to multiple people.

It looks like you strictly enforce that lots should be assigned to a single entity at a time.  Do you think this is the right decision?  Or can we have some lots that must be assigned to one entity at a time, and some lots that can be assigned to multiple entities at a time?

My personal preference is to allow users to assign lots to multiple people.  What do you think?

server/models/migrations/v0.8.0-v0.9.0/migrate.sql
mbayopanda: Yes

mbayopanda:
`My personal preference is to allow users to assign lots to multiple people. What do you think?`

This implementation is focused on unique item first, assume that we have laptops but each laptop is unique and we can just assign it to one entity at time.

jniles:
@mbayopanda, if that is the case, we need to make it really clear the difference between unique items and non-unique items.  For example, we should probably use serial numbers instead of lot numbers and allow the expiration date column to be NULL.

client/src/modules/stock/assign/modals/action.modal.js
```diff
@@ -0,0 +1,107 @@
...
+   * Since data contains inventories and lots that we need, we do not want to
+   * perform others queries to the server, so we extract inventories and lots
+   * from the data given
+   * @param {array} data
+   */
+  function computeAvailableInventories(data) {
+    vm.globalAvailableLots = data;
+    vm.groupedInventories = Util.groupBy(data, 'inventory_uuid');
+    const uniqueInventoriesUuids = Util.uniquelize(data.map(item => item.inventory_uuid));
+    vm.availableInventories = uniqueInventoriesUuids.map(inventoryUuid => vm.groupedInventories[inventoryUuid][0]);
```
jniles, line 103: Can this just be `Object.values(availableInventories)`?

mbayopanda:

jniles:
@jeremielodi could you give this a review?

client/src/js/services/StockService.js
```diff
@@ -27,6 +27,15 @@ function StockService(Api, StockFilterer) {
   // API for stock import
   const importing = new Api('/stock/import');
 
+  // API for stock assignment
+  const stockAssign = new Api('/stock/assign/');
+
+  // Overide the stock assign api
+  stockAssign.remove = uuid => {
+    return stockAssign.$http.put(`/stock/assign/${uuid}/remove`)
```
jeremielodi, line 35: You should  use  ``` $http.delete ``` method

client/src/js/services/receipts/ReceiptService.js
```diff
@@ -165,13 +165,20 @@ function ReceiptService($http, util, Language, AppCache, Session) {
...
 
   // stock exit patient receipt
   function stockExitPatientReceipt(uuid, options) {
     const route = '/receipts/stock/exit_patient/'.concat(uuid);
     return fetch(route, options);
   }
 
+  // stock assign receipt
+  function stockAssignReceipt(uuid, options) {
+    const route = '/receipts/stock/assign/'.concat(uuid);
```
jeremielodi, line 182: This can be  betterly written like this ```const route = `/receipts/stock/assign/${uuid}`;  ```

client/src/modules/stock/stock.routes.js
```diff
@@ -53,10 +53,52 @@ angular.module('bhima.routes')
...
   }]);
 
 
 function closeModals($uibModalStack) {
   $uibModalStack.dismissAll();
 }
+
+// creates both the create and update states
+function onEnterFactory(stateType) {
+  const isCreateState = (stateType === 'create');
```
jeremielodi: parentheses are not useful here, go ahead and remove them

jeremielodi (commented):
The code looks correct, try to fixe theses few problems below!

client/src/js/services/StockService.js
mbayopanda: The remove method here is about removing assignment, it doesn't delete the record but just change the value of the column `is_active` in the `stock_assign` table.

client/src/js/services/receipts/ReceiptService.js
mbayopanda: You're right, it looks pretty but i will be obliged to update others lines of code in this file for having the same style, the other thing is `code-climate` allow this style so i think we can keep it like this.

client/src/js/services/StockService.js
jeremielodi: okay

client/src/js/services/receipts/ReceiptService.js
jeremielodi: okay

jeremielodi:
LGTM good job @mbayopanda

PR #3430, opened 2018-12-03
feat(invoice regisry) set project after search

### LINKED ISSUE #3414: Invoices Registry does not set the Project after search
Most search filters re-populate the search modal when you search again.  This is a handy feature for users, since they can modify only a few filters and get similar results.

We have a bug: the invoice registry project search does not repopulate itself after a search.  You can see this in the gif below:

jeremielodi:
closes #3414


jniles (approved):
Looks right!

bors r+

jniles:
@jeremielodi can you check the other registries to see if they need the same fix?

PR #3431, opened 2018-12-03
lifetime has a high priority than quantity

mbayopanda:
This PR fix priority issue in stock lots between lifetime and quantity : The lifetime has a higher priority than the quantity

**Before :** 
image

**After :** 
image

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -492,11 +492,12 @@ function processMultipleLots(inventories) {
   const inventoryLots = _.groupBy(inventories, 'inventory_uuid');
 
   _.map(inventoryLots, (lots) => {
-    // order lots by ascending lifetime
-    let orderedInventoryLots = _.orderBy(lots, 'lifetime', 'asc');
     // order lots also by ascending quantity
     // assuming the lot with lowest quantity is consumed first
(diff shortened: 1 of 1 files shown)


server/controllers/stock/core.js
```diff
@@ -492,11 +492,12 @@ function processMultipleLots(inventories) {
   const inventoryLots = _.groupBy(inventories, 'inventory_uuid');
 
   _.map(inventoryLots, (lots) => {
-    // order lots by ascending lifetime
-    let orderedInventoryLots = _.orderBy(lots, 'lifetime', 'asc');
     // order lots also by ascending quantity
     // assuming the lot with lowest quantity is consumed first
-    orderedInventoryLots = _.orderBy(orderedInventoryLots, 'quantity', 'asc');
+    let orderedInventoryLots = _.orderBy(lots, 'quantity', 'asc');
```
jniles, line 497: You know, you can actually combine these two calls, right?  [documentation](https://lodash.com/docs/4.17.11#orderBy).

```js
const orderedInventoryLots = _.orderBy(lots, ['quantity', 'asc'], ['lifetime', 'asc']);
```

It might be simpler to make changes in the future like this.

jniles (approved):
LGTM.

bors r+

PR #3432, opened 2018-12-04
number of groups

### LINKED ISSUE #3352: Add a column to the price list grid to show number of groups it modifies
It would also be nice to note how many groups are currently using a price list.  Similar to how the patient groups, debtor groups, and others link the lists from their main page, it would be nice to have "{{ x }} groups using this list" as a column.

jeremielodi:
closes #3352

### DIFF
--- a/client/src/modules/patients/groups/groups.html
+++ b/client/src/modules/patients/groups/groups.html
@@ -26,11 +26,12 @@
           <div class="panel-heading">
             <span translate> PATIENT_GROUP.ALL_PATIENT_GROUP </span>
           </div>
-          <table class="table table-condensed">
+          <table class="table table-condensed table-bordered">
             <thead>
               <tr>
                 <th translate> TABLE.COLUMNS.NAME </th>
                 <th translate> TABLE.COLUMNS.PRICE_LIST </th>
+                <th translate> FORM.LABELS.PATIENTS</th>
                 <th translate> TABLE.COLUMNS.ACTION </th>
               </tr>
             </thead>
@@ -38,6 +39,7 @@
               <tr ng-repeat="group in PatientGroupCtrl.groups track by group.uuid" ng-class="{ 'bg-primary' : group.uuid == PatientGroupCtrl.patientGroup.uuid }">
                 <td>{{ group.name}}</td>
                 <td>{{ group.priceListLabel }}</td>
+                <td class="text-right">{{group.patientNumber}}</td>
                 <td>
                   <a class="btn btn-xs btn-default" ng-click="PatientGroupCtrl.update(group.uuid)" id="group-{{group.uuid}}">
                     <span class="glyphicon glyphicon-pencil"></span>
--- a/client/src/modules/prices/prices.js
+++ b/client/src/modules/prices/prices.js
@@ -49,6 +49,12 @@ function PriceListController(
       cellTemplate : `/modules/prices/templates/itemsNumber.cell.html`,
     },
     {
+      field : 'subcribedGroupsNumber',
+      displayName : 'PATIENT_GROUP.PATIENT_GROUP',
+      headerCellFilter : 'translate',
+      cellTemplate : `/modules/prices/templates/subscribedGroups.cell.html`,
+    },
+    {
       field : 'FORM.BUTTONS.ACTIONS',
       width : 100,
       enableFiltering : false,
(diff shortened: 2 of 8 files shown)


client/src/modules/prices/prices.js
```diff
@@ -48,6 +48,12 @@ function PriceListController(
       headerCellFilter : 'translate',
       cellTemplate : `/modules/prices/templates/itemsNumber.cell.html`,
     },
+    {
+      field : 'subcribedGroupsNumber',
+      displayName : 'PATIENT_GROUP.PATIENT_GROUP',
+      headerCellFilter : 'translate',
+      cellTemplate : `/modules/prices/templates/subscribedGroups.cell.html`,
```
jniles, line 55: Why did you need to make a cell for this?

server/controllers/finance/priceList.js
```diff
@@ -36,9 +36,16 @@ function lookup(req) {
 
   if (req.query.detailed === '1') {
     sql = `
-      SELECT BUID(pl.uuid) as uuid, pl.label, pl.created_at, pl.description, COUNT(pl_it.uuid) as itemsNumber
+      SELECT BUID(pl.uuid) as uuid, pl.label, pl.created_at, pl.description,
+        COUNT(pl_it.uuid) as itemsNumber, groups.nbr as subcribedGroupsNumber
       FROM price_list pl
       LEFT JOIN price_list_item pl_it ON pl_it.price_list_uuid = pl.uuid
+      LEFT JOIN (
```
jniles: Why does this need to be a subquery?  Can't we do this in the same format as the `itemsNumber`?

client/src/modules/patients/groups/groups.html
```diff
@@ -26,18 +26,20 @@
           <div class="panel-heading">
             <span translate> PATIENT_GROUP.ALL_PATIENT_GROUP </span>
           </div>
-          <table class="table table-condensed">
+          <table class="table table-condensed table-bordered">
             <thead>
               <tr>
                 <th translate> TABLE.COLUMNS.NAME </th>
                 <th translate> TABLE.COLUMNS.PRICE_LIST </th>
+                <th translate> FORM.LABELS.PATIENT</th>
```
jniles: Looks like this should be `PATIENTS`.

server/controllers/medical/patientGroups.js
```diff
@@ -30,8 +30,15 @@ function list(req, res, next) {
   const filters = new FilterParser(req.query, { tableAlias : 'pg' });
   const sql = `
     SELECT BUID(pg.uuid) as uuid, pg.name, BUID(pg.price_list_uuid) AS price_list_uuid,
-      pg.note, pg.created_at, pl.label AS priceListLabel, pl.description
-    FROM patient_group AS pg LEFT JOIN price_list AS pl ON pg.price_list_uuid = pl.uuid
+      pg.note, pg.created_at, pl.label AS priceListLabel, pl.description,
```
jniles, line 33: Is this needed for the Price List group count...?

jniles (commented):
It looks like you've combined two pull requests into one.  What needed to change in patient groups?

client/src/modules/prices/prices.js
jeremielodi: Hum..., I was trying to add this subscribers number where possible to end with this problem

server/controllers/finance/priceList.js
jeremielodi: There is an important ```GROUP BY``` in the subquery

client/src/modules/prices/prices.js
jniles: But do you need a template?  What happens if you remove the cell template?

server/controllers/medical/patientGroups.js
jeremielodi: It's used in the patient group grid, we also need to know how many patients belong to a group

server/controllers/finance/priceList.js
jniles: Try
```sql
LEFT JOIN patient_group ON patient_group.price_list_uuid = pl.uuid
```

And then put `COUNT(patient_group.uuid)  as subcribedGroupsNumber` in the SELECT part.  I think ti should behave the same, but save quite a bit of work.  JOINs are typically more standard than writing subqueries.

jniles (approved):
LGTM.

bors r+

PR #3444, opened 2018-12-06
Fix lots registry

mbayopanda:
This PR enhance calculation in the lots registry according stock management : 
- Unique CMM for each lots of an inventory
- Lots with zero MSD are not considered
- Expired lots are not considered

Before : 
image

After : 
image


server/controllers/stock/core.js
```diff
@@ -489,25 +490,41 @@ function getInventoryQuantityAndConsumption(params) {
...
+    const inventoryLots = _.groupBy(depotInventories, 'inventory_uuid');
+
+    _.map(inventoryLots, (lots) => {
+      // if we don't have the default CMM (avg_consumption) use the
+      // defined or computed CMM for each lots
+      const cmm = _.max(lots.map(lot => lot.avg_consumption));
+
+      // order lots also by ascending quantity
+      // assuming the lot with lowest quantity is consumed first
+      let orderedInventoryLots = _.orderBy(lots, 'quantity', 'asc');
```
jniles, line 506: These are some nice comments 👍

jniles (approved):
I'm not sure I understand it all, but the code is clear.  We should work towards writing tests for this registry and/or stock calculations so it is easier to tell what the logic is and what the corner cases are.

bors r+

Issue #2648, opened 2018-04-04
Additional unit tests for client services/util.js


jniles:
We already have some unit tests for [services/util.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/util.js) ([here](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/test/client-unit/services/util.spec.js)).  These only cover the Dojo-style before(), after(), once() functions.  We need to extend our tests coverage for this file.

**Tests To Write**
 1. `formatDate()` should take a date and transform it into a string.
 2. `formatDate()` should take a date and transform it into a format passed in as a second argument.
 3. `formatDate()` should default to `null` if nothing is passed in.
 4. `filterFormElements()` should return an empty object if provided an empty object.
 5. `filterFormElements()` should return `$modelValue` or `$bhValue` properties of sub-objects.
 6. `filterFormElements()` should filter out values that are not `$dirty` if `requireDirty = true` is passed in as a second parameter.
 7. `clean()` should filter out all properties that have empty values or that begin with `$`.
 8. `getMomentAge()` should return a string.
 9. `defaultBirthMonth` should be defined.
 10. `uniquelize` should turn an array into unique values.
 11. `isEmptyObject()` should return true for `{}`.
 12. `isEmptyObject()` should return true for `[]`.
 13. `isEmptyObject()` should return false for an object that has keys.
 14. The `xor` versus each entry in the [XOR Truth Table](https://en.wikipedia.org/wiki/Exclusive_or).
 15.  `maskObjectFromKeys()` should return an object with only the properties specified as a second value.
 16. `maskObjectFromKeys()` should return an empty object if given empty keys.
 17. `debounce()` should fire a function in the future.

We also have `arrayIncludes`, but this is just a mask for `array.includes()`.  Since we use ES2015+, we could remove it and change all references to `array.includes()`.

jniles:
@jeremielodi, can you work on this?

jeremielodi:
okay

jniles:
Closed in #3448.

PR #3448, opened 2018-12-06
feat(unit test) more tests for client/util.js service

jeremielodi:
implements all features related at #2648

### DIFF
--- a/test/client-unit/services/util.spec.js
+++ b/test/client-unit/services/util.spec.js
@@ -1,4 +1,4 @@
-/* global inject, expect */
+/* global inject, expect, chai */
 describe('util', () => {
   let util;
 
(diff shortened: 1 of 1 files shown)


test/client-unit/services/util.spec.js
```diff
@@ -74,4 +74,130 @@ describe('util', () => {
...
     expect(input).to.be.equal(2);
   });
+
+  it('#formatDate() should take a date and transform it into a string.', () => {
+    const date = new Date('2018-01-31');
+    const expected = 'string';
+
+    const formatedDate = util.formatDate(date);
+
+    expect(typeof (formatedDate)).to.be.equal(expected);
```
jniles: The works!  You can also do `expect(formatDate).to.be.a('string')`.

test/client-unit/services/util.spec.js
```diff
@@ -74,4 +74,130 @@ describe('util', () => {
...
+  });
+
+  it('#filterFormElements() should return an empty object if provided an empty object', () => {
+    const form = {};
+    const expected = {};
+    const elements = util.filterFormElements(form);
+    expect(elements).to.deep.equal(expected);
+  });
+
+  it.skip('#filterFormElements() should return $modelValue or $bhValue properties of sub-objects', () => {
```
jniles, line 109: I'll work on this one. 👍

test/client-unit/services/util.spec.js
```diff
@@ -74,4 +74,130 @@ describe('util', () => {
...
+    const data = [];
+    const expected = true;
+    const formatedData = util.isEmptyObject(data);
+    expect(formatedData).to.be.equal(expected);
+  });
+
+  it('#xor should returns the logical XOR of two booleans.', () => {
+    const expected = true;
+    const formatedData = util.xor(true, false);
+    expect(formatedData).to.be.equal(expected);
```
jniles: We should probably write several tests here, since XOR is defined very strictly:
```js
expect(xor(true, true)).to.equal(false);
expect(xor(false, true)).to.equal(true);
expect(xor(true, false)).to.equal(true);
expect(xor(false, false)).to.equal(false);
````

File: test/client-unit/services/util.spec.js [hunk 4]
```diff
@@ -74,4 +74,130 @@ describe('util', () => {
...
+  });
+
+  it('#maskObjectFromKeys should return an object with only the properties specified as a second value', () => {
```
jniles, line 172: Nice 👍

test/client-unit/services/util.spec.js
```diff
@@ -74,4 +74,130 @@ describe('util', () => {
...
+      age : 12,
+    };
+    const mask = [];
+    const expected = {};
+    const formatedData = util.maskObjectFromKeys(data, mask);
+    expect(formatedData).to.deep.equal(expected);
+  });
+
+
+  it.skip('debounce() should fire a function in the future.', () => {
```
jniles: The way to test this is with a [spy](https://www.chaijs.com/plugins/chai-spies/).  We need to add a function to debounce, and check that the function is called later.

Since [debounce() uses a timeout](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/util.js#L270), we will need to do the same thing here:
```js
it('#debounce() should fire a function in the future', (done) => {
  const spy = chai.spy();
  const debounced = debounce(spy, 200);  // debounce spy for 200 milliseconds
  expect(spy).to.not.have.been.called();
  setTimeout(() => {
    expect(spy).to.have.been.called();
    done();
   }, 250);
});
```

The basic idea is:
  1. We create a spy function that tells us when it is called or not.
  2. We pass in the `done()` function so that [mocha knows that this test is asynchronous](https://stackoverflow.com/a/21175342).
   3. We make sure the function is not called immediately (otherwise it might a problem with the function).
   4. We make sure the function is called after the timeout.

jniles (commented):
These are good tests!  I've made a suggestion for the `debounce()` test.  You should see if it works, so that you have some experience testing asynchronous code with spies.  Once you are done, I'll work on the filterFormElements function.

test/client-unit/services/util.spec.js
```diff
@@ -74,4 +74,140 @@ describe('util', () => {
...
+  it('#isEmptyObject() should turn true for []', () => {
+    const data = [];
+    const expected = true;
+    const formatedData = util.isEmptyObject(data);
+    expect(formatedData).to.be.equal(expected);
+  });
+
+  it('#xor should returns the logical XOR of two booleans.', () => {
+    const { xor } = util;
+    expect(xor(true, true)).to.equal(false);
```
jniles, line 166: Much better!

jniles (approved):
Looks good to me.  Great job.

I'll add the other tests once this is merged.

bors r+

PR #3451, opened 2018-12-07
fix: parse project abbr to get transaction length

### LINKED ISSUE #3450: Project abbreviations should not be limited to three letters
If we make project abbreviations longer than three letters, the system breaks.  This is due to a hard `substring(transId, 4)` command when generating transaction identifiers.  It is undocumented and completely unclear to the user that this is happening or why it should be the case.

buginprojectidgeneration
_Fig 1: Bug in Trans ID generation with project abbr longer than 3_

jniles:
This commit adds a new function to the SQL to parse the project
abbreviation and come up with the correct transaction identifier length.

Closes #3450.

### DIFF
--- a/server/models/functions.sql
+++ b/server/models/functions.sql
@@ -103,6 +103,24 @@ BEGIN
   );
 END $$
 
+/**
+ GetTransactionNumberPart(transId, projectId)
+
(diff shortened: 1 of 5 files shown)


jeremielodi (commented):
LGTM

PR #3459, opened 2018-12-10
chore(db): add DELIMITERs to all *.sql

jniles:
This commit adds DELIMITER queries to all SQL files (in particular, the
stored procedures) so that we can use them in various contexts without
assuming a particular environment.


mbayopanda (approved):
LGTM :+1

bors r+

PR #3466, opened 2018-12-12
feat(account) remove classe column

### LINKED ISSUE #3453: Remove the column "classe" from the account table
We have fully adopted the account type fashion of distinguishing the account's use.  This means we can finally drop the "classe" column of the accounts table.

jeremielodi:
closes #3453


jniles:
We'll need to add an entry to the migration sql script for this.

jniles (approved):
LGTM.  

bors r+

PR #3467, opened 2018-12-12
refactor(bhStockEntryExitType): rewrite component

jniles:
This commit rewrites the `bhStockEntryExitType` component to make it simpler and more performant.  It also adds a special warning message if the user forgot to add entry/exit types for the depot - instead of appearing blank, it will now inform the user of the error.

_GIF 1:  Here's how it looks_

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -91,6 +91,8 @@
     "MOVEMENTS"       : "Movements",
     "NO_ORDER"        : "No Orders",
     "NO_DATA"         : "No Data",
+    "NO_ENTRY_TYPES"  : "There are no entry types defined for {{text}}.  Add at least one entry type in the Depot Management module.",
+    "NO_EXIT_TYPES"  : "There are no exit types defined for {{text}}.  Add at least one exit type in the Depot Management module.",
     "OUTPUT"          : "Output",
     "ORDERS"          : "Orders",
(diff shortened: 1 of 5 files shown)


client/src/i18n/en/stock.json
```diff
@@ -91,6 +91,8 @@
     "MOVEMENTS"       : "Movements",
     "NO_ORDER"        : "No Orders",
     "NO_DATA"         : "No Data",
+    "NO_ENTRY_TYPES"  : "There are no entry types defined for {{text}}.  Add at least one entry type in the Depot Management module.",
+    "NO_EXIT_TYPES"  : "There are no exit types defined for {{text}}.  Add at least one exit type in the Depot Management module.",
```
mbayopanda, line 95: 👍

client/src/modules/templates/bhStockEntryExitType.tmpl.html
```diff
@@ -1,24 +1,23 @@
-<div ng-if="type.isAllowed" class="col-md-3 col-xs-6" ng-repeat="type in $ctrl.entryExitTypeList | filter:{isEntry : $ctrl.isEntry}">
```
mbayopanda, line 1: Good catch by defining `entryTypes` or `exitTypes` in the controller instead of filtering in the view 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3468, opened 2018-12-12
Edit lot feature

### LINKED ISSUE #2489: Allow editing lots after stock entry
Lots are created on stock entry, but as we all have experienced, users make many mistakes.  Imagine a simple mistake to make - entering the wrong expiration date.  This will fundamentally affect the functioning of the stock modules and all other reports/alerts generated off of the database.

We need a (privileged) way to allow a user to edit a lot.  A lot is mostly just a lot number and an expiration date.  This should be regarded as a "super user" style action.

### LINKED ISSUE #2661: feat(lot) : Add an edit page for lots
This issue is about to add a page for editing lot information (label, expiration_date, entry_date and also unit_cost)

mbayopanda:
This PR is about the issue #2661 

In this PR we can now edit lot details : label, expiration date and the unit cost.

Closes #2489.   Closes #2661.

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -168,7 +168,12 @@
   "LOTS" : {
     "REGISTRY" : "Lots Registry",
     "ENABLE_FAST_INSERT": "Enable fast lots insertions",
-    "ENABLE_FAST_INSERT_DESCRIPTION": "This option will automatically select the next blank row's serial/lot number in the entry table. If necessary, it adds a new row. Useful for quickly inserting with a barcode reader"
+    "ENABLE_FAST_INSERT_DESCRIPTION": "This option will automatically select the next blank row's serial/lot number in the entry table. If necessary, it adds a new row. Useful for quickly inserting with a barcode reader",
+    "DETAILS":"Lot Information",
+    "EXPIRATION_DATE":"Expiration Date",
+    "LOT_LABEL":"Lot label",
+    "UNIT_COST":"Unit cost",
+    "SUCCESSFULLY_EDITED":"Lot updated with success"
   },
   "MOVEMENTS" : {
     "REGISTRY" : "Movements Registry"
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -170,9 +170,7 @@
     "ENABLE_FAST_INSERT": "Enable fast lots insertions",
     "ENABLE_FAST_INSERT_DESCRIPTION": "This option will automatically select the next blank row's serial/lot number in the entry table. If necessary, it adds a new row. Useful for quickly inserting with a barcode reader",
     "DETAILS":"Lot Information",
-    "EXPIRATION_DATE":"Expiration Date",
     "LOT_LABEL":"Lot label",
-    "UNIT_COST":"Unit cost",
     "SUCCESSFULLY_EDITED":"Lot updated with success"
   },
   "MOVEMENTS" : {
--- a/client/src/js/services/LotService.js
+++ b/client/src/js/services/LotService.js
@@ -1,0 +1,17 @@
+angular.module('bhima.services')
+  .service('LotService', LotService);
+
+LotService.$inject = ['PrototypeApiService'];
+
+function LotService(Api) {
+  const lots = new Api('/lots/');
+
+  lots.read = (uuid) => {
+    return Api.read.call(lots, uuid)
+      .then(res => {
+        res.expiration_date = new Date(res.expiration_date);
+        return res;
+      });
+  };
+  return lots;
+}
--- a/client/src/modules/stock/lots/modals/edit.modal.js
+++ b/client/src/modules/stock/lots/modals/edit.modal.js
@@ -23,7 +23,6 @@ function EditLotModalController(Data, Session, Lots, Notify, Instance) {
   }
 
   function onDateChange(date) {
-    if (date <= new Date()) { return; }
     vm.model.expiration_date = date;
   }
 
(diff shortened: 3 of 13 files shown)


client/src/i18n/en/stock.json
```diff
@@ -166,7 +166,12 @@
   "LOTS" : {
     "REGISTRY" : "Lots Registry",
     "ENABLE_FAST_INSERT": "Enable fast lots insertions",
-    "ENABLE_FAST_INSERT_DESCRIPTION": "This option will automatically select the next blank row's serial/lot number in the entry table. If necessary, it adds a new row. Useful for quickly inserting with a barcode reader"
+    "ENABLE_FAST_INSERT_DESCRIPTION": "This option will automatically select the next blank row's serial/lot number in the entry table. If necessary, it adds a new row. Useful for quickly inserting with a barcode reader",
+    "DETAILS":"Lot Information",
+    "EXPIRATION_DATE":"Expiration Date",
```
jniles: Don't we have the text for expiration date somewhere else we could use?  And unit cost?

client/src/js/services/LotService.js
```diff
@@ -0,0 +1,17 @@
+angular.module('bhima.services')
+  .service('LotService', LotService);
+
+LotService.$inject = ['PrototypeApiService'];
+
+function LotService(Api) {
+  const lots = new Api('/lots/');
+
+  lots.read = (uuid) => {
```
jniles, line 9: Nice 👍

jniles: Note that one disadvantage of using `=>` is that we can't use `.bind()` on arrow functions.  But it shouldn't matter here.

client/src/modules/stock/lots/modals/edit.modal.js
```diff
@@ -0,0 +1,41 @@
...
+  function startup() {
+    Lots.read(Data.uuid)
+      .then(lot => {
+        vm.model = lot;
+      })
+      .catch(Notify.handleError);
+  }
+
+  function onDateChange(date) {
+    if (date <= new Date()) { return; }
```
jniles: I think we should let a user a set any date here - no need for this restriction.

client/src/modules/stock/lots/modals/search.modal.js
```diff
@@ -3,11 +3,10 @@ angular.module('bhima.controllers')
 
 // dependencies injections
 SearchLotsModalController.$inject = [
-  'data', 'InventoryService', 'NotifyService',
-  'util', 'Store', '$uibModalInstance', 'PeriodService', 'StockService',
+  'data', 'util', 'Store', '$uibModalInstance', 'PeriodService', 'StockService',
 ];
 
-function SearchLotsModalController(data, Inventory, Notify, util, Store, Instance, Periods, Stock) {
+function SearchLotsModalController(data, util, Store, Instance, Periods, Stock) {
```
jniles, line 9: Nice 👍

client/src/modules/stock/lots/registry.service.js
```diff
@@ -0,0 +1,139 @@
...
+
+  service.gridFooterTemplate = `
+    <div>
+      <b>{{ grid.appScope.countGridRows() }}</b>
+      <span translate>TABLE.AGGREGATES.ROWS</span>
+    </div>
+  `;
+
+  service.orderByDepot = (rowA, rowB) => {
+    return rowA.depot_text > rowB.depot_text ? 1 : -1;
```
jniles: Remember that strings without ASCII characters should be compared using [`String.prototype.localeCompare()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Sorting_non-ASCII_characters).

server/controllers/stock/lots.js
```diff
@@ -0,0 +1,62 @@
...
+const db = require('../../lib/db');
+
+exports.update = update;
+exports.details = details;
+
+/**
+ * GET /stock/lots/:uuid
+ * Get details of a lots
+ */
+function details(req, res, next) {
```
jniles, line 23: Nice API 👍

server/controllers/stock/lots.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/**
+ * PUT /stock/lots/:uuid
+ * Edit a stock lot
+ */
+function update(req, res, next) {
+  const bid = db.bid(req.params.uuid);
+  const allowedToEdit = ['label', 'expiration_date', 'unit_cost'];
+  const params = _.pick(req.body, allowedToEdit);
```
jniles, line 50: Nice 👍

test/integration/stock/lots.js
```diff
@@ -0,0 +1,46 @@
...
+    return agent.put(`/lots/${shared.lotQuinineUuid}`)
+      .send(update)
+      .then((res) => {
+        expect(res).to.have.status(200);
+        return agent.get(`/lots/${shared.lotQuinineUuid}`);
+      })
+      .then(res => {
+        const details = res.body;
+        expect(details.label).to.be.equal(update.label);
+        expect(formatDate(details.expiration_date)).to.be.equal(update.expiration_date);
```
jniles, line 38: As an alternative to using `formatDate()`, I'm pretty sure this will work:
```js
const update = {
       expiration_date : new Date('2020-12-15'),
};

// later ....
      .then(res => {
        const details = res.body;
        expect(details.label).to.be.equal(update.label);
        expect(details.expiration_date).to.be.equal(update.expiration_date);
      });
```

Could you try it and see if it works?

jniles (commented):
Code looks really good!  I've made some comments for improvements.

test/integration/stock/lots.js
mbayopanda: The server return the date in the string format, here is the result of the test : 
image

jniles: Alright, thank you for trying! :+1:

jniles (approved):
LGTM.

Bors r+

PR #3474, opened 2018-12-13
feat(server unit test) numberToText test

### LINKED ISSUE #2312: Write unit tests for NumberToText.js

### NumberToText.js
The following test cases should be covered in the unit tests of [NumberToText.js](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/NumberToText.js):
 - [ ] Test that the number 1 in English and USD correctly renders one dollar.
 - [ ] Test that number 1 in English and FC correctly renders one congolese franc
 - [ ] Test that the number 1 in French and USD correctly renders the french text.
 - [ ] What happens when a negative number is passed to the library?   This should be tested.
 - [ ] What happens when an unknown currency name is used?
 - [ ] What happens when the number is null?

jeremielodi:
closes  #2312

### DIFF
--- a/test/server-unit/numberToText.spec.js
+++ b/test/server-unit/numberToText.spec.js
@@ -1,0 +1,77 @@
+const { expect } = require('chai');
+const numberToText = require('../../server/lib/NumberToText');
+const en = require('../../client/src/i18n/en/numbers.json');
+const fr = require('../../client/src/i18n/fr/numbers.json');
+const util = require('../../server/lib/util');
+
+const loadDictionary = util.loadDictionary;
+
+describe('NumberToText convert number to text', () => {
+
+   // use local dictionary
+    before(() => {
+     util.loadDictionary = lang => lang === 'fr' ? fr : en;
+    });
+
+  
+  // mock translation dictionaries
+  const dictionaries = { en, fr };
+
+  const currencies = {
+    fr : { usd : 'Dollars', fc : 'Franc Congolais' },
+    en :  { usd : 'Dollars', fc : 'Congolese francs' },
+  };
+
+  it(`#Test that the number 1 in English and USD correctly renders 'One dollar'.`, () => {
+    const input = 1;
+    const expected = `One  ${currencies.en.usd}`;
+    const convertedNumber = numberToText.convert(input, 'en', currencies.en.usd, dictionaries.en);
+
+    expect(convertedNumber).to.be.equal(expected);
+  });
+
+  it(`#Test that the number 1 in French and FC correctly renders 'Un Franc Congolais'.`, () => {
+    const input = 1;
+    const expected = `Un  ${currencies.fr.fc}`;
+    const convertedNumber = numberToText.convert(input, 'fr', currencies.fr.fc, dictionaries.fr);
+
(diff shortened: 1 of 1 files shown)


server/lib/NumberToText.js
```diff
@@ -22,12 +22,12 @@ exports.convert = convert;
 *  More about those after we take a look at numToWords itself …
 */
 
-function convert(input, lang, currencyName) {
+function convert(input, lang, currencyName, _dictionary) {
```
jniles: Instead of modifying this function, let's just overwrite the `util.loadDictionary()` function.  Like this:

```js
const numberToText = require('../../server/lib/NumberToText'); 
const util = require('../../server/lib/util');

util.loadDictionary = lang => lang === 'fr' ? fr : en;
```

I think that should be enough work to make this test work without modifying the underlying code.

test/server-unit/numberToText.spec.js
```diff
@@ -0,0 +1,66 @@
...
+  it(`#Test when a negative number is passed, it absolute value will be used instead'.`, () => {
+    const input = -10;
+    const expected = `Dix  ${currencies.fr.fc}`;
+    const convertedNumber = numberToText.convert(input, 'fr', currencies.fr.fc, dictionaries.fr);
+
+    expect(convertedNumber).to.be.equal(expected);
+  });
+
+  /* @TODO : propose a default curreny when an undefied value is defined */
+  it(`#Test when a unknow currency is passed, undefied will be used as default value'.`, () => {
```
jniles, line 50: There is  a typo here: `unknow` should be spelled `unknown`.

test/server-unit/numberToText.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    const input = -10;
+    const expected = `Dix  ${currencies.fr.x}`;
+    const convertedNumber = numberToText.convert(input, 'fr', currencies.fr.x, dictionaries.fr);
+
+    expect(convertedNumber).to.be.equal(expected);
+  });
+
+  it(`#Test when a unknow currency is passed, undefied will be used as default value'.`, () => {
```
jniles, line 58: Same typo here.  `unknow` -> `unknown`.

jniles (changes requested):
The tests are well written.  Normally, tests shouldn't modify the underlying code.  I've proposed a method to try and prevent this.  Try it and see if it works.

test/server-unit/numberToText.spec.js
```diff
@@ -0,0 +1,78 @@
+const { expect } = require('chai');
+const numberToText = require('../../server/lib/NumberToText');
+const en = require('../../client/src/i18n/en/numbers.json');
+const fr = require('../../client/src/i18n/fr/numbers.json');
+const util = require('../../server/lib/util');
+const _ = require('lodash');
+
+const loadDictionary = _.clone(util.loadDictionary);
```
jniles: I don't think you need to clone this.  Just replace it like:
```js
const loadDictionary = util.loadDictionary; 
```

It looks like lodash doesn't support cloning functions.

jniles (approved):

jniles:
LGTM.

bors r+

PR #3476, opened 2018-12-14
deps: upgrade bootstrap 3.4.0

jniles:
This commit upgrades our bootstrap version to 3.4.0 to avoid a critical attack path.

See https://github.com/IMA-WorldHealth/bhima-2.X/pull/3436 for more information.


jniles:
Seems like a fluke.

bors r+

PR #3477, opened 2018-12-14
Enterprise logo and header with logo

mbayopanda:
This PR add a way of adding an enterprise logo :


client/src/i18n/en/errors.json
```diff
@@ -49,6 +49,7 @@
     "ER_DUP_FIELDNAME" : "Duplicate column name",
     "ER_DUP_UNIQUE" : "A unique key is duplicated",
     "ER_ACCESS_DENIED_ERROR" : "Access denied for the current user",
-    "ER_NO_DB_ERROR" : "No database selected"
+    "ER_NO_DB_ERROR" : "No database selected",
+    "ER_TOO_MANY_FIELDS" : "There are many more columns"
```
jniles, line 53: 💃

client/src/index.html
```diff
@@ -39,7 +39,7 @@
               <div class="header-image">
                 <a class="app-logo" id="logo">
                   <span class="logo-container">
-                    <img src="assets/icon.png" width="40" height="40">
+                    <img ng-src="{{ AppCtrl.enterprise.logo || 'assets/icon.png' }}" width="40" height="40">
```
jniles: You can try this to see if we can get the logo automatically updating:
```html
ng-src="AppCtrl.getEnterpriseLogo()"
```

And then define the function to be:
```js
vm.getEnterpriseLogo = () => Session.enterprise && Session.enterprise.logo || 'assets/icon.png';
```

See if that works!

server/controllers/admin/enterprises.js
```diff
@@ -171,3 +172,21 @@ exports.update = function update(req, res, next) {
...
+    next(BadRequest('Expected at least one file upload but did not receive any files.'));
+    return;
+  }
+
+  const logo = req.files[0].link;
+  const sql = 'UPDATE enterprise SET logo = ? WHERE id = ?';
+
+  db.exec(sql, [logo, req.params.id])
+    .then(() => {
+      res.status(200).json({ link : logo });
```
jniles: I think this should just be `res.status(200).json({ logo })`.  Since we call the property "logo" in the table, we should return that name here.

jniles (commented):
Just a few comments here.

jniles (approved):
Looks good to me!

bors r+

PR #3484, opened 2018-12-23
refactor: clean up old reports

jniles:
This PR renames, refactors, and removes old reports.  The goal is to standardize report names, remove duplicated reports, and make sure the names makes sense in French and English.


jniles:
I wonder why codeclimate isn't working anymore.  I think it might have to do with us renaming the repo.  I'm looking into it.

PR #3497, opened 2018-12-30
feat(report): rework/rename unpaid inv w/ Pivot

jniles:
This commit renames the unbalanced invoices as unpaid invoices and makes it use a pivot table.  This is based on the feedback we received from HEV.


jniles:
Yes!  Hound now works :)

PR #3525, opened 2019-01-07
feat(patient group) set column note not required

### LINKED ISSUE #3516: Do not make "note" a required field on Patient Groups
When creating a patient group, you must specify a note, since it is required.  Let's remove this requirement, as it is annoying and rarely useful.

jeremielodi:
closes #3516


jniles (approved):
LGTM.

bors r+

PR #3527, opened 2019-01-08
chore(vouchers): lint toolkit and reduce code

jniles:
This commit refactors the voucher toolkits through linting the JS code. I've also used the bh-cashbox-select on the convention payment page to reduce the overall size of the file.


jniles:
@lomamech or @jeremielodi , can I get a review?

jniles:
Looks like the internet crisis is limiting abilities to review.

bors r+

PR #3528, opened 2019-01-08
feat(journal search modal) add currency filter

### LINKED ISSUE #3488: Add search filter on currency_id to the Journal
If we want to find all transactions made in USD or FC, there isn't a good way to do this.  We should add a filter to the client that allows you to search on the currency id.  We already have the infrastructure on the server.

jeremielodi:
closes #3488


client/src/modules/templates/bhCurrencySelect.tmpl.html
```diff
@@ -26,7 +26,7 @@
         ng-change="$ctrl.onChange({ currency : $ctrl.currencyId })"
         data-currency-option="{{ currency.id }}"
         ng-disabled="currency.disabled"
-        required>
+        ng-required="$ctrl.required">
```
jniles, line 29: Nice 👍

jniles (commented):
LGTM.

bors r+

PR #3531, opened 2019-01-09
fix(report): add debtor text to unpaid invoices

jniles:
This commit makes the unpaid invoices report more useful as it adds the full debtor name to the report.  Based on feedback from HEV.


jniles:
Looks like it is affected by: https://github.com/IMA-WorldHealth/bhima/issues/3529.

bors r+

PR #3540, opened 2019-01-15
feat(operating report) change currency

jeremielodi:


server/controllers/finance/reports/operating/index.js
```diff
@@ -48,9 +49,8 @@ function getQueryExpense() {
   `;
 }
 
-function document(req, res, next) {
+async function document(req, res, next) {
```
jniles: It seems like this `async` is out of place.  Normally, you would use `await` with `async`, but it doesn't look like you've gone that far.  Could you remove it?

server/controllers/finance/reports/operating/index.js
```diff
@@ -77,9 +78,13 @@ function document(req, res, next) {
 
   fiscal.getDateRangeFromPeriods(periods).then(dateRange => {
```
jniles, line 52: So, I never noticed this before, but it looks like we send the periods to the server, convert them into dates in this line, then later do:
```sql
      pt.period_id IN (
        SELECT id FROM period WHERE start_date>= ? AND end_date<= ?
      )
```
... which basically undoes the whole operation.  We probably could make this code cleaner by just using the period ids we have from the client directly.

This change could be another PR.

server/controllers/finance/reports/operating/index.js
```diff
@@ -112,7 +117,7 @@ function document(req, res, next) {
         totalIncome : totalIncome.total,
         dateFrom : range.dateFrom,
         dateTo : range.dateTo,
-        currencyId : req.session.enterprise.currency_id,
+        currencyId : params.currency_id,
```
jniles, line 93: 👍

server/controllers/finance/reports/operating/index.js
```diff
@@ -25,9 +26,9 @@ const DECIMAL_PRECISION = 2; // ex: 12.4567 => 12.46
  * This function does exactly the same thing except the value of amount will be
  * debit - credit this will know the expense account
  */
-function getQueryExpense() {
+function getQueryExpense(rate = 1) {
```
jniles: It looks like this is the same code as `getAccountBalancesByTypeId`.  Could we just delete this code and replace it with the `getAccountBalancesByTypeId` call?

jniles (commented):
Just a few clean up comments.  I haven't tested this locally yet, but will do so once the code is clean.

jniles (approved):
Looks good to me.

bors r+

PR #3544, opened 2019-01-21
feat(report) set compte de resultat to the enterprise currency

### LINKED ISSUE #3508: Compte de Resultat is not in the enterprise currency
See below:
comptederesultat should be in fc

The enterprise currency in this case is FC, but the Compte de Resultat is in USD.  This is likely because the currency_id is not correctly passed to the currency filter.

jeremielodi:
closes #3508


jniles (approved):
LGTM.

bors r+

PR #3545, opened 2019-01-21
feat(patient edit) display the reference

### LINKED ISSUE #3518: Give the Patient identifier (PA.XXX.123) on the Patient Edit page
For some reason, we don't specify the patient identifier on the patient edit page.  It would be helpful to specify it on this page, since it is one of the core pieces of information identifying the patient in the system.

jeremielodi:
closes #3518

### DIFF
--- a/client/src/modules/patients/edit/edit.html
+++ b/client/src/modules/patients/edit/edit.html
@@ -108,6 +108,19 @@
                   </div>
                 </div>
 
+                <!-- Reference Name -->
+                <div class="form-group">
+
+                 <label class="col-md-3 control-label" translate> FORM.LABELS.REFERENCE </label>
+                 <div class="col-md-9">
+                   <p class="form-control-static">
+                      <span>{{PatientEditCtrl.medical.reference}}</span>
+                   </p>
+                 
+                 </div>
+               </div>
+
+
                 <!-- Display Name -->
                 <div
                   class="form-group has-feedback"


jniles:
Can we put this in the form instead of in the headercrumb?  Like in the "Required Information" section as a non-editable field.

client/src/modules/patients/edit/edit.html
```diff
@@ -108,6 +108,18 @@
                   </div>
                 </div>
 
+                 <!-- Reference Name -->
+                 <div
+                 class="form-group has-feedback"
+                 ng-class="{'has-error' : patientDetails.display_name.$invalid && patientDetails.$submitted}">
```
jniles: This is better!

Since the user can't edit it, we don't need the `has-feedback` or `has-error` classes.  go ahead and remove those.

client/src/modules/patients/edit/edit.html
```diff
@@ -108,6 +108,18 @@
...
+
+                 <label class="col-md-3 control-label" translate> FORM.LABELS.REFERENCE </label>
+                 <div class="col-md-9">
+                   <input class="form-control" name="reference" disabled="disabled" ng-model="PatientEditCtrl.medical.reference">
```
jniles: Instead of a disabled input (which seems like the user should be able to edit at some point), bootstrap gives us another class called `form-control-static` if the input will never be editable.  We used it right above this input [in the registration date section](https://github.com/IMA-WorldHealth/bhima/pull/3545/files#diff-d1558a6f91b42148dcbbe4563ad75eaeR107).

Could you change this to `<p class="form-control-static">` and put the reference inside it?

jniles (commented):
Improving, but we can do even better :)

client/src/modules/patients/edit/edit.html
jniles: This still needs to be addressed, and then we can merge it.

jniles: @jeremielodi are you going to address these?

jeremielodi:
@jniles this PR can be merged!

jniles:
Yes it can!

Looks good to me.

bors r+

PR #3547, opened 2019-01-21
Improve account reference

### LINKED ISSUE #3523: improve Account Reference
The module of Account Reference is a very capital module, which makes it possible to make the regrouping of the accounts, this module is used in several reports/ratios in particular the report/ratio of the centers of the expenses, the report/ratio of the assessment, the income statement report/ratio, this issues proposes to add an identifier to identify them and gathered

lomamech:
- Remove of the property debit_balance and credit_balance on
the reference of Account
- Give the possibility to be able to delete the parent property in the
  registration form of the reference accounts
- Implementing the Account Reference Type module
- Implementing the server API
- Implementation of the integration test as well as the E2E test for the
management of the account reference type
- Complete Integration test and E2E test for Account Reference
- Implement Component Account Reference type Select
- Set Default value for Account Reference Type
- Add the component for the select of Account Reference Type
- Update the report of Account Reference Set the Account Reference type if
  is selected

closes #3523


client/src/modules/account_reference/account_reference.modal.js
```diff
@@ -52,14 +52,22 @@ function AccountReferenceModalController($state, Accounts, AccountReferences, No
...
 
+  // callback for Account Reference Type
+  function onSelectAccountReferenceType(referenceType) {
+    vm.accountReference.reference_type_id = referenceType.id;
+  }
+
   // submit the data to the server from all two forms (update, create)
   function submit(accountReferenceForm) {
     if (accountReferenceForm.$invalid) { return null; }
+    // Fixe me @lomamech : Give the possibility to validate the modification
```
jniles, line 63: :+1:

client/src/modules/account_reference_type/account_reference_type.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  function loadGrid() {
+    toggleLoadingIndicator();
+    vm.hasError = false;
+
+    AccountReferenceType.read()
+      .then((accountReferenceType) => {
+        accountReferenceType.forEach((item) => {
+          item.label = $translate.instant(item.label);
```
jniles: It looks like every time we download account reference types, we want to translate them.  Let's move this operation into the `AccountReferenceTypeService` so that after every `read()` call, we translate these labels.

client/src/modules/account_reference_type/modals/account_reference_type.modal.html
```diff
@@ -0,0 +1,47 @@
...
+        <label class="badge badge-warning" translate>FORM.LABELS.UPDATE</label>
+      </li>
+    </ol>
+  </div>
+  
+  <div class="modal-body" style="overflow-y: scroll; max-height:600px; ">
+    <!-- label -->
+    <div class="form-group" ng-class="{ 'has-error' : AccountReferenceTypeForm.$submitted && AccountReferenceTypeForm.description.$invalid }">
+      <label class="control-label" translate>FORM.LABELS.DESIGNATION</label>
+      <div>
```
jniles: Why do we need this extra div..?

client/src/modules/account_reference_type/modals/account_reference_type.modal.html
```diff
@@ -0,0 +1,47 @@
+<form name="AccountReferenceTypeForm" bh-submit="AccountReferenceTypeModalCtrl.submit(AccountReferenceTypeForm)" novalidate>
+  <div class="modal-header">
+    <ol class="headercrumb">
+      <li class="static" translate>TREE.ADMIN</li>
+
+      <li ng-if="AccountReferenceTypeModalCtrl.isCreating" class="title">
```
jniles: It looks like the only thing that changes here is the `badge`.  Could you move the `ngIf` to put it on the badge?

```html
<label ng-if="AccountReferenceTypeModalCtrl.isCreating" class="badge badge-warning" translate>FORM.LABELS.CREATE</label>
```

client/src/modules/account_reference_type/modals/account_reference_type.modal.html
```diff
@@ -0,0 +1,47 @@
...
+        <label class="badge badge-warning" translate>FORM.LABELS.CREATE</label>
+      </li>
+      <li ng-if="!AccountReferenceTypeModalCtrl.isCreating" class="title">
+        <span translate>TREE.ACCOUNT_REFERENCE_TYPE</span>
+        <label class="badge badge-warning" translate>FORM.LABELS.UPDATE</label>
+      </li>
+    </ol>
+  </div>
+  
+  <div class="modal-body" style="overflow-y: scroll; max-height:600px; ">
```
jniles, line 14: NOTE: We should try to standardize our modal sizes somewhere.

client/src/modules/account_reference_type/templates/action.cell.html
```diff
@@ -0,0 +1,36 @@
+<div class="ui-grid-cell-contents text-right" uib-dropdown dropdown-append-to-body  data-row="{{row.entity.label}}">
+  <span ng-if="!row.entity.fixed">
+    <a href uib-dropdown-toggle data-action="open-dropdown-menu">
+      <span data-method="action" translate>FORM.BUTTONS.ACTIONS</span>
+      <span class="caret"></span>
+    </a>
+
+    <ul data-row-menu="{{row.entity.label}}" class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>	
+      <li>
```
jniles, line 15: Could you add a header to the dropdown to give the user context about which row they have clicked?  It should look like [the one found on the account action dropdown](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/accounts/templates/grid.actionsCell.tmpl.html#L9).

client/src/modules/account_reference_type/templates/action.cell.html
```diff
@@ -0,0 +1,36 @@
...
+        <a data-method="edit-record" ng-click="grid.appScope.edit(row.entity)" href>
+          <span>
```
jniles: Do we need this extra `<span>`?

server/controllers/finance/accounts/accountReferenceType.js
```diff
@@ -0,0 +1,109 @@
...
+      res.status(201).json({ id : row.insertId });
+    })
+    .catch(next)
+    .done();
+}
+
+
+// PUT /account_reference_type /:id
+function update(req, res, next) {
+  const sql = `UPDATE account_reference_type SET ? WHERE id = ?;`;
```
jniles, line 69: Let's delete `req.body.id` before doing this action to make sure the user cannot change the id.

server/controllers/finance/accounts/accountReferenceType.js
```diff
@@ -0,0 +1,109 @@
...
+    .then((record) => {
+    // all updates completed successfull, return full object to client
+      res.status(200).json(record);
+    })
+    .catch(next)
+    .done();
+}
+
+// DELETE /account_reference_type/:id
+function del(req, res, next) {
```
jniles: Let's cal this function `remove` instead of `del` since `delete` is a keyword.

jniles (commented):
Just a few comments here.  Looks pretty good!

lomamech:
@jniles, I need another review

client/src/js/components/bhAccountReferenceTypeSelect.js
```diff
@@ -0,0 +1,33 @@
...
+        accountReferenceTypes.forEach((item) => {
+          item.label = $translate.instant(item.label);
+        });
+
+        $ctrl.accountReferenceTypes = accountReferenceTypes;
+      });
+  };
+
+  $ctrl.onSelect = accountReferenceType => $ctrl.onSelectCallback({ accountReferenceType });
+}
```
mbayopanda, line 33: 👍

client/src/modules/account_reference_type/account_reference_type.js
```diff
@@ -0,0 +1,99 @@
...
+      });
+  }
+
+  function handleError(error) {
+    vm.hasError = true;
+    Notify.handleError(error);
+  }
+
+  function loadGrid() {
+    toggleLoadingIndicator();
```
mbayopanda: We don't know the value of the `vm.loading` here, it will be more clear to assign the value directly to the variable instead of calling the function. For example : 
```js
vm.loading = true;
```

client/src/modules/account_reference_type/modals/account_reference_type.modal.html
```diff
@@ -0,0 +1,42 @@
...
+      <label class="control-label" translate>FORM.LABELS.DESIGNATION</label>
+      <input name="label" ng-model="AccountReferenceTypeModalCtrl.types.label" autocomplete="off" ng-maxlength="90" class="form-control" required>
+      <div class="help-block" ng-messages="AccountReferenceTypeForm.label.$error" ng-show="AccountReferenceTypeForm.$submitted">
+        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      </div>
+    </div>
+  </div>
+
+  <div class="modal-footer">
+    <p id="account-reference-type-same" ng-if="AccountReferenceTypeForm.$submitted && AccountReferenceTypeForm.$pristine && !AccountReferenceTypeModalCtrl.isCreating" class="text-warning">
```
mbayopanda: Is it possible to have a variable for example `vm.hasNoChange` in the controller like this : 
```js
vm.hasNoChange = form.$submitted && form.$pristine && !vm.isCreating;
```
So that we can reduce the code in the view like this : 
```html
<p id="account-reference-type-same" ng-if="AccountReferenceTypeModalCtrl.hasNoChange" class="text-warning">
  <i class="fa fa-warning"></i> <span translate>FORM.INFO.RECORD_SAME</span>
</p>
```

mbayopanda: Also, the translation entry key 'FORM.INFO.RECORD_SAME' can have a meaningful name which reflect the message : 
image

Something like : `FORM.INFO.NO_CHANGE` can be not bad.

client/src/modules/account_reference_type/modals/account_reference_type.modal.js
```diff
@@ -0,0 +1,72 @@
...
+  AccountReferenceType.read()
+    .then(ReferencesType => {
+      vm.ReferencesType = ReferencesType;
+    })
+    .catch(Notify.handleError);
+
+  // submit the data to the server from all two forms (update, create)
+  function submit(accountReferenceTypeForm) {
+    if (accountReferenceTypeForm.$invalid) { return null; }
+    if (!accountReferenceTypeForm.$dirty) { return null; }
```
mbayopanda: lines above can be written like this : 
```js 
if (accountReferenceTypeForm.$invalid || !accountReferenceTypeForm.$dirty) { return null; }
```

client/src/modules/reports/generate/account_reference/account_reference.html
```diff
@@ -34,6 +34,14 @@ <h3 class="text-capitalize" translate>REPORT.ACCOUNT_REFERENCE.TITLE</h3>
               on-select-callback="ReportConfigCtrl.onSelectPeriod(period)">
             </bh-period-selection>
 
+            <bh-account-reference-type-select
+              reference-type-id="ReportConfigCtrl.reportDetails.reference_type_id"
+              id="reference_type_id"
+              name="reference_type_id"
+              on-select-callback="ReportConfigCtrl.onSelectAccountReferenceType(accountReferenceType)">
+              <bh-clear on-clear="ReportConfigCtrl.clear('reference_type_id')"></bh-clear>
```
mbayopanda, line 42: 👍 well done

client/src/modules/templates/bhAccountReferenceTypeSelect.tmpl.html
```diff
@@ -0,0 +1,27 @@
...
+
+    <label class="control-label" translate>
+      FORM.LABELS.ACCOUNT_REFERENCE_TYPE
+    </label>
+
+    <ng-transclude></ng-transclude>
+    <ui-select
+      name="reference_type_id"
+      ng-model="$ctrl.referenceTypeId"
+      on-select="$ctrl.onSelect($item, $model)">
```
mbayopanda: Could you add the property `append-to-body="true"` so that the ui-select list be correctly displayed in the modal.

image

server/controllers/finance/accounts/accountReferenceType.js
```diff
@@ -0,0 +1,113 @@
...
+
+// DELETE /account_reference_type/:id
+function remove(req, res, next) {
+  const sql = `DELETE FROM account_reference_type WHERE id = ?;`;
+
+  db.exec(sql, [req.params.id])
+    .then((row) => {
+    // if nothing happened, let the client know via a 404 error
+      if (row.affectedRows === 0) {
+        throw new NotFound(`Could not find a function with id ${req.params.id}`);
```
mbayopanda: Does it correct the error message talk about function ?

mbayopanda: It is about preference may be, did you try the `lookupAccountReferenceType(req.params.id)` function for looking if the account reference exists, this function will thrown an 404 error automatically if the record doesn't exists.

If you decide to use the `lookupAccountReferenceType` function the code can be : 
```js
const sql = `DELETE FROM account_reference_type WHERE id = ?;`;
lookupAccountReferenceType(req.params.id)
  .then(() => db.exec(sql, [req.params.id]))
  .then(() => res.sendStatus(204))
  .catch(next)
  .done();

```

server/controllers/finance/reports/account_reference/report.handlebars
```diff
@@ -16,6 +16,12 @@
         {{translate 'REPORT.ACCOUNT_REFERENCE.TITLE'}}
       </h2>
 
+      <h4 class="text-center text-uppercase">
+        <strong>
+          {{translate referenceTypeLabel}}
```
mbayopanda, line 21: 👍

server/models/migrations/v0.8.0-v0.9.0/migrate.sql
```diff
@@ -1021,3 +1021,26 @@ INSERT INTO report VALUES  (23, 'unpaid-invoice-payments', 'REPORT.UNPAID_INVOIC
...
+) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+-- update the account_reference table
+ALTER TABLE account_reference ADD COLUMN `reference_type_id` MEDIUMINT(8) UNSIGNED NULL;
+
+-- Default Account Reference Type
+INSERT INTO `account_reference_type` (`id`, `label`, `fixed`) VALUES
+(1, 'FORM.LABELS.FEE_CENTER', 1),
+(2, 'FORM.LABELS.BALANCE_SHEET', 1),
+(3, 'FORM.LABELS.PROFIT_LOSS', 1);
```
mbayopanda, line 1042: It is a good idea to have translation key as label, but the bad thing with that is if `FORM.LABELS.BALANCE_SHEET` is translated  `Bilan` we can create a new account reference type with the label `Bilan` the user will see duplicated data : 

image

Even if in the database we have : 
image

We can just set fixed account reference type with real label not with translation key so that we have uniform value in the database (not translation key mixed with real account reference type name).

```sql
INSERT INTO `account_reference_type` (`id`, `label`, `fixed`) VALUES
(1, 'Centre de cout', 1),
(2, 'Bilan', 1),
(3, 'Centre de profit', 1);
```

mbayopanda (commented):
The code LGTM 👍  

I made some comments for clarity and better user experience.

server/models/migrations/v0.8.0-v0.9.0/migrate.sql
lomamech: If I set the default values in French, what will happen if we install Bhima in an English speaking country? I kept the same logic that exists on the transaction type table

lomamech:
@mbayopanda  I have finish the review

lomamech:
@mbayopanda  I'm waiting for this branch, to make the PR conserning the breakeven report

PR #3551, opened 2019-01-22
fix(journal) amount filter update

### LINKED ISSUE #3546: The amount search field in the Journal causes all other filters to be ignored
Bug report from Vanga:

> The limit function for Finances  Journal de Saisie does not exclude other dates properly

> So I want to see what the items are for this month at a value of 99000
(21 enregistrements)  11 Transactions postés  0 Transaction(s) non postée(s)
The result comes back with items posted as far back as 13/2/2017
The limit to a certain number works.  For example I only get 10 transactions if I say limit to 10 transactions.

It seems like this field somehow causes all other fields to be ignored in the WHERE clause.  I'm guessing it is an issue with the parentheses or with an OR clause.

jeremielodi:
closes #3546

### DIFF
--- a/server/controllers/finance/journal/index.js
+++ b/server/controllers/finance/journal/index.js
@@ -170,7 +170,7 @@ function buildTransactionQuery(options, posted) {
   filters.custom('accounts_id', 'p.account_id IN (?)', [options.accounts_id]);
   const { amount } = options;
   filters.custom(
-    'amount', '(credit = ? OR debit = ?) OR (credit_equiv = ? OR debit_equiv = ?)',
+    'amount', '(credit = ? OR debit = ? OR credit_equiv = ? OR debit_equiv = ?)',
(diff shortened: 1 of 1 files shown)


server/controllers/finance/journal/index.js
```diff
@@ -170,7 +170,7 @@ function buildTransactionQuery(options, posted) {
   filters.custom('accounts_id', 'p.account_id IN (?)', [options.accounts_id]);
   const { amount } = options;
   filters.custom(
-    'amount', '(credit = ? OR debit = ?) OR (credit_equiv = ? OR debit_equiv = ?)',
```
jniles, line 173: :+1:  Good catch!

jniles (approved):
Looks like this got hit by https://github.com/IMA-WorldHealth/bhima/issues/3529.

bors r+

PR #3552, opened 2019-01-22
deps: bump dependencies to latest

jniles:
This PR bumps all dependencies to their latest version, except ui-router, which has major breaking changes for both an upgrade and a downgrade.

The major changes are:
 1. We removed all `@bower_components`  and replaced them with their native NPM counterparts.  This allows green-keeper to keep us up to date with different versions.  The one exception to this is angular-ui-router, as mentioned above.
 2. `gulp` has been upgraded to version 4!  If there was a performance increase, it was small.  However, this allows us to rewrite our gulpfile for better clarity.
 3. A ton of unused CSS has been pruned.  We don't use it anymore, so let us remove it.
 4. All `vendor` references have been removed.  We no longer use a vendor folder without bower.

Super awesome changes:
 1. When using `NODE_ENV=development`, the application bundles the unminified dependencies.  That means we have much better insight into what bugs are happening and less cryptic error message.


jniles:
I jumped the gun a bit, and we are being blocked now by https://github.com/angular/angular.js/issues/16814.  We'll see if they release a new version with the bug patched soon.

jniles:
@mbayopanda, this is also finally ready for review.

client/src/js/app.js
```diff
@@ -203,6 +203,9 @@ function httpConfig($httpProvider) {
 
   // register error handling interceptor
   $httpProvider.interceptors.push('ErrorInterceptor');
+
+  // perf - applyAsync
+  $httpProvider.useApplyAsync(true);
```
mbayopanda, line 208: Do we really need this ?

I found this [comment](https://stackoverflow.com/questions/41458545/what-are-the-disadvantages-of-http-useapplyasync)

client/src/js/services/NotifyService.js
```diff
@@ -10,43 +10,43 @@ NotifyService.$inject = ['$translate'];
  * @todo  unit tests
  */
 function NotifyService($translate) {
-  var service = this;
+  const service = this;
 
   // default time to live of 3 seconds
-  var TTL = 3000;
+  const TTL = 1000000;
```
mbayopanda: :-) i think the comment must also be updated.

A curiosity question, why this decision of using 1000 seconds now ?

package.json
```diff
@@ -41,129 +41,109 @@
...
-    "@bower_components/ng-file-upload": "danialfarid/angular-file-upload-bower#12.2.13",
-    "@bower_components/ngstorage": "gsklee/ngStorage#0.3.11",
-    "@bower_components/webcam-directive": "jonashartmann/webcam-directive#3.1.2",
-    "@types/angular": "^1.6.41",
+    "@types/angular": "^1.6.53",
     "accounting-js": "^1.1.1",
-    "angular": "angular/bower-angular#1.7.5",
-    "angular-animate": "angular/bower-angular-animate#1.7.5",
-    "angular-dynamic-locale": "lgalfaso/angular-dynamic-locale#0.1.37",
-    "angular-file-upload-bower": "danialfarid/angular-file-upload-bower#12.2.13",
```
mbayopanda, line 72: Don't we need the `angular-file-upload-bower` anymore ?

server/controllers/finance/reports/ohada_profit_loss/index.js
```diff
@@ -296,7 +296,6 @@ function document(req, res, next) {
         return item;
       });
 
-      // console.log(assetTable);
```
mbayopanda, line 299: 👍

sh/test.sh
```diff
@@ -32,7 +32,7 @@ startfold "Running server Unit Tests ......" "server-unit"
 endfold "server-unit" ;
 
 # run end to end tests
-startfold "Running Client Unit Tests..." "test-end-to-end";
+startfold "Running Client End to End Tests..." "test-end-to-end";
```
mbayopanda, line 35: 👍 good catch

mbayopanda (commented):
I made a comment about the `angular-file-upload-bower` dependency.

And also, i have got this error with font-awesome : 

image

client/src/js/services/NotifyService.js
jniles: Yup, you are right.  This is a mistake, and I'll revert it.

client/src/js/app.js
jniles: It's not clear if this will help or not, but I can imagine three places where this will help:
  1. Typeaheads/validators (such as the hospital_no) will trigger less digest calls.  We use these quite a bit in our application, shipping lots of small requests rapidly over a local network.
  2. Modules where we add new grid rows to Stock/Vouchers/Transactions that prompt HTTP requests.  For example, when we add voucher tools, those often trigger several HTTP requests from the Voucher module.  This will stack those requests to make the browser do fewer $digest calls.
  3. Each ui-grid cellTemplate seems to send an HTTP request.  I'm not sure why - maybe we need to debug this somewhere, but for the moment, this might help.

Again, this might be a bad idea, but we have to try it to find out... from the documentation, it doesn't look like it will _hurt_ anything, just might not help as much as we hope.

jniles:
Thanks for the thorough review @mbayopanda. 

About snyk - I'm not sure what is going on there, although I remember some issue with snyk in the past that was similar.  Do the font-awesome icons load when you launch BHIMA?

package.json
jniles: Good question.  It has been renamed `ng-file-upload` and is still included in the package.json file.

[This is the old repository](https://github.com/danialfarid/ng-file-upload-bower).  You can in the README that they link to "angular-file-upload", but that it redirects to [ng-file-upload](https://github.com/danialfarid/ng-file-upload).

And [ng-file-upload is in line 92 of this file](https://github.com/IMA-WorldHealth/bhima/pull/3552/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R92).

mbayopanda:
I cannot run the application correctly : 
image

mbayopanda:
With `yarn` there is not any errors during the `yarn` command

image

mbayopanda:
Everything seems to work correctly and also the file upload

mbayopanda:
Could you fix the mergin conflict so that i can merge it

PR #3569, opened 2019-02-04
Record changes in Debtor Groups via the Debtor Group History table

### LINKED ISSUE #3517: Record changes in Debtor Groups via the Debtor Group History table
We have a table called `debtor_group_history` that describes the debtor group changes for each debtor.  It would be nice to use this table to track changes in a debtor's group when they are moved from one group to another.

Ideally, we would like to record the following information:
 1. `previous_debtor_group` - where they were before
 2. `next_debtor_group` - where they ended up
 3. `timestamp` - the time the request was initiated
 4. `user_id` - the person who initiated it.

This would give us some perspective on the movement of patients/debtors between groups.  For example, patients could be transferred based on their employment from Patients Payant Cash to an employee group and back, if they were taken up by the hospital.

jeremielodi:
closes #3517

### DIFF
--- a/server/controllers/finance/debtors/index.js
+++ b/server/controllers/finance/debtors/index.js
@@ -17,6 +17,7 @@
 */
 
 const q = require('q');
+const _ = require('lodash');
 const db = require('../../../lib/db');
 const NotFound = require('../../../lib/errors/NotFound');
 const identifiers = require('../../../config/identifiers');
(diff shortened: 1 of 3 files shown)


server/controllers/finance/debtors/index.js
```diff
@@ -64,28 +65,43 @@ function detail(req, res, next) {
...
   // escape the group_uuid if it exists
   if (req.body.group_uuid) {
-    req.body.group_uuid = db.bid(req.body.group_uuid);
+    data.group_uuid = db.bid(req.body.group_uuid);
+  }
+  const transaction = db.transaction();
+  transaction.addQuery(sql, [data, uid]);
+
+  // check if we should update the historic table
+  if (req.body.group_uuid !== previousGroup.group_uuid) {
```
jniles: Instead of just checking the previous debtor_group_uuid, can we just assume that if they sent the `group_uuid` they are changing the debtor group?  That would make this simpler.

server/controllers/finance/debtors/index.js
```diff
@@ -64,28 +65,43 @@ function detail(req, res, next) {
...
   }
 
-  db.exec(sql, [req.body, uid])
-    .then(() => lookupDebtor(req.params.uuid))
-    .then((debtor) => {
-      res.status(200).json(debtor);
-    })
-    .catch(next)
-    .done();
+  return transaction.execute().then(() => {
```
jniles: If you have `async`/`await`, why not just use `async`/`await`?  You could rewrite this code as:
```js
await transaction.execute();

const debtor = await lookupDebtor(req.params.uuid);
res.status(200).json(debtor);
```

Remember that every time you use async/await, you need to wrap it in a `try/catch` block.

server/models/migrations/v0.8.0-v0.9.0/migrate.sql
```diff
@@ -1044,3 +1044,23 @@ INSERT INTO `account_reference_type` (`id`, `label`, `fixed`) VALUES
...
 (225, 'Account Reference Type','TREE.ACCOUNT_REFERENCE_TYPE','Account Reference Type', 1,'/modules/account_reference_type','/account_reference_type');
+
+
+DROP TABLE IF EXISTS `debtor_group_history`;
+CREATE TABLE `debtor_group_history` (
+  `uuid` BINARY(16) NOT NULL,
+  `debtor_uuid` BINARY(16) DEFAULT NULL,
+  `previous_debtor_group` BINARY(16) DEFAULT NULL,
+  `next_debtor_group` BINARY(16) DEFAULT NULL,
+  `income_date` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
jniles: I know you were not the original author, but what is `income_date`?  Can we just make it called `created_at`, like most timestamps?

server/controllers/finance/debtors/index.js
```diff
@@ -64,28 +65,46 @@ function detail(req, res, next) {
...
+      next_debtor_group : data.group_uuid,
+    });
+
+
+    await transaction.execute();
+    const debtor = await lookupDebtor(req.params.uuid);
+    res.status(200).json(debtor);
+
+  } catch (error) {
+    next();
```
jniles: This should be called with `next(error)`.  You can read more about it in the [expressjs documentation](http://expressjs.com/en/guide/routing.html).

jniles (commented):
One last modification with the error and then we are good to go!

jniles:
Looks good to me.  We'll merge this, then I will make the relevant changes in https://github.com/IMA-WorldHealth/bhima/pull/3573

PR #3572, opened 2019-02-05
feat(bh-account-select) add on blur action

### LINKED ISSUE #3533: Add "select on blur" to bh-account-inline-typeahead
One issue I've run into with Complex Vouchers: it gets difficult to select the accounts if we need to click or hit "enter" every single time to select the account.  UI Select provides a way around this by allowing us to "select on blur", which will automatically select the highlighted item if the user blurs the input.  This would be helpful for dealing with giant transactions on the complex vouchers page.

jeremielodi:
closes #3533


client/src/modules/templates/bhAccountTypeaheadInline.html
```diff
@@ -12,6 +12,8 @@
       uib-typeahead="account as account.hrlabel for account in $ctrl.accounts | filter:{ hrlabel: $viewValue } | limitTo:10"
       typeahead-template-url="bhAccountTypeaheadItem.html"
       typeahead-on-select="$ctrl.onSelect($item)"
+      typeahead-select-on-exact="true"
```
jniles: Do we need the on exact part?  Does it add anything?

PR #3573, opened 2019-02-05
Migration script for deployed apps to version 9

mbayopanda:
This PR add some scripts to execute for migrating deployed app to the version v.0.9 of bhima.

- the `setup-migration-script` build the migration-`database_name`.sql which make the migration
- `migrate.revised.sql` is a cleaned version of `migrate.sql` without functions, procedures and not necessary queries.


server/models/migrations/v0.8.0-v0.9.0/setup-migration-script.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+# this script rebuild function and procedure 
+
+DATABASE="imck"
```
jniles: By the way - a trick that will enable us to use this script in multiple sites and not just IMCK is to have the variable default to whatever is passed in.  It works like this:
```sh
DATABASE=${DATABASE:-"imck"}
```

What this will do is set DATABASE to whatever is passed in.  If nothing is passed in, it will default to "imck".  [Here is the documentation for these shell variables](https://bash.cyberciti.biz/guide/Default_shell_variables_value).

jniles: That way, we can call this script:
```sh
DATABASE="pepkit" ./setup-migration-script.sh
```
and it will switch to operating on the database called "pepkit".

server/models/migrations/v0.8.0-v0.9.0/setup-migration-script.sh
```diff
@@ -0,0 +1,46 @@
...
+FILENAME="migration"
+
+# path to the cleaned migration script
+CLEANED_MIGRATION_SCRIPT="./migrate.revised.sql"
+
+# path to bhima root from ./server/models/migrations/v0.8.0-v0.9.0
+BHIMA_PATH="$HOME/IMA/bhima-2.x"
```
jniles: I propose we use default variables to allow us to set the BHIMA_PATH and DATABASE variables.  This will make it easier to work on multiple different environments.

server/models/migrations/v0.8.0-v0.9.0/migrate.revised.sql
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * =============================================================================
+ * STRUCTURE
+ * =============================================================================
+*/
+
+-- Fix reference_uuid index bug
+ALTER TABLE `general_ledger` DROP INDEX `reference_uuid`;
+ALTER TABLE `posting_journal` DROP INDEX `reference_uuid`;
```
jniles: I am pretty sure all our installation sites already have this properly set. I remember upgrading IMCK with this previously, but we never published it as a release :(.

jniles (approved):
@mbayopanda this looks really good!  I like the idea of having a "revised" migrate script that cleans everything up.  Once we publish this, we can probably just remove the old migration script and keep the revised script.

Go ahead and test this on the local server and see how far you get.  I think if you implement my suggestions for default variables it will help a lot.

mbayopanda:
@jniles

jniles:
@mbayopanda, looks good to me.  Let's leave this open until Friday, when we actually do the upgrades. :)

PR #3587, opened 2019-02-11
feat(invoice receipt) display created_at label

### LINKED ISSUE #3505: Need to display created_at somewhere on the receipts
We store the creation date timestamp `created_at` for invoices, vouchers, and cash payments.  This is useful information to have when records are created with previous dates.  The timestamp allows us to triangulate exactly when the record was created.

It would be ideal if we could find some place on the receipt to show this information or on the receipt modal.  Currently, I have to go into the back end to discover the true creation date of a record.

jeremielodi:
closes #3505

### DIFF
--- a/server/controllers/finance/patientInvoice.js
+++ b/server/controllers/finance/patientInvoice.js
@@ -100,7 +100,7 @@ function lookupInvoice(invoiceUuid) {
       project.abbr, invoice.reference) AS reference, invoice.cost,
       invoice.description, BUID(invoice.debtor_uuid) AS debtor_uuid,
       patient.display_name AS debtor_name,   BUID(patient.uuid) as patient_uuid,
-      invoice.user_id, invoice.date, user.display_name, invoice.service_id,
+      invoice.user_id, invoice.date, invoice.created_at, user.display_name, invoice.service_id,
       service.name AS serviceName, enterprise.currency_id
     FROM invoice
     LEFT JOIN patient ON patient.debtor_uuid = invoice.debtor_uuid
--- a/server/controllers/finance/reports/invoices/receipt.handlebars
+++ b/server/controllers/finance/reports/invoices/receipt.handlebars
@@ -40,7 +40,8 @@
       <div class="col-xs-6">
         <span class="text-capitalize">{{translate 'FORM.LABELS.INVOICE'}}</span>: <strong>{{reference}}</strong> <br>
         <span class="text-capitalize">{{translate 'TABLE.COLUMNS.SERVICE'}}</span>: {{serviceName}} <br>
-        <span class="text-capitalize">{{translate 'FORM.LABELS.DATE'}}</span>: {{date date}} <br>
+        <span class="text-capitalize">{{translate 'FORM.LABELS.DATE'}}</span>: {{date date}}<br>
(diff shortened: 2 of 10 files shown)


server/controllers/finance/reports/invoices/receipt.handlebars
```diff
@@ -40,8 +40,11 @@
       <div class="col-xs-6">
         <span class="text-capitalize">{{translate 'FORM.LABELS.INVOICE'}}</span>: <strong>{{reference}}</strong> <br>
         <span class="text-capitalize">{{translate 'TABLE.COLUMNS.SERVICE'}}</span>: {{serviceName}} <br>
-        <span class="text-capitalize">{{translate 'FORM.LABELS.DATE'}}</span>: {{date date}} <br>
+        <span class="text-capitalize">{{translate 'FORM.LABELS.DATE'}}</span>: {{date date}}<br>
+        {{#if delay}}
+        <span class="text-capitalize">{{translate 'FORM.LABELS.CREATED_AT'}}</span>: {{date created_at}}<br>
         <span class="text-capitalize">{{translate "FORM.LABELS.CREATED_BY"}}</span>: {{display_name}}
+        {{/if}}
```
jniles: We don't want to hide the CREATED_BY if the delay is different :/

server/controllers/finance/patientInvoice.js
```diff
@@ -100,7 +100,8 @@ function lookupInvoice(invoiceUuid) {
       project.abbr, invoice.reference) AS reference, invoice.cost,
       invoice.description, BUID(invoice.debtor_uuid) AS debtor_uuid,
       patient.display_name AS debtor_name,   BUID(patient.uuid) as patient_uuid,
-      invoice.user_id, invoice.date, user.display_name, invoice.service_id,
+      invoice.user_id, invoice.date, invoice.created_at, user.display_name, invoice.service_id,
+      DATEDIFF(invoice.date, invoice.created_at) as date_difference,
```
jniles: To save modifying the JS, you could just do something like;
```
IF(DATEDIFF(invoice.date, invoice.created_at), TRUE, FALSE) AS has_delay
```
or something similar.  Then you would just need to change the handlebars templates to do `{{if has_delay}}`.

jniles (commented):
This is on the right track!

Can we do this to all receipts?  (voucher/cash/invoice/stock)?

That would be excellent.

server/controllers/finance/patientInvoice.js
```diff
@@ -100,7 +100,8 @@ function lookupInvoice(invoiceUuid) {
       project.abbr, invoice.reference) AS reference, invoice.cost,
       invoice.description, BUID(invoice.debtor_uuid) AS debtor_uuid,
       patient.display_name AS debtor_name,   BUID(patient.uuid) as patient_uuid,
-      invoice.user_id, invoice.date, user.display_name, invoice.service_id,
+      invoice.user_id, invoice.date, invoice.created_at, user.display_name, invoice.service_id,
+      IF(DATEDIFF(invoice.date, invoice.created_at), TRUE, FALSE) AS has_delay,
```
jniles: It looks like this is the only place you list "has_delay".  Want to remove it?

jniles (approved):

jniles (approved):
LGTM.

bors r+

PR #3589, opened 2019-02-11
feat(aged debtor) change currency

### LINKED ISSUE #3210: Allow Aged Debtors to show values in a non-enterprise currency
The Aged Debtor Report only supports the enterprise currency at this point.  It would be helpful to show the values in a non-enterprise currency.  It should function similarly to the Account Statement Report currency select, among others.

jeremielodi:
closes  #3210

### DIFF
--- a/server/controllers/finance/reports/debtors/index.js
+++ b/server/controllers/finance/reports/debtors/index.js
@@ -57,6 +57,7 @@ function agedDebtorReport(req, res, next) {
   db.one(sql, [qs.period_id])
     .then(period => {
       qs.date = period.end_date;
+      qs.enterprise_id = metadata.enterprise.id;
       // fire the SQL for the report
       return queryContext(qs);
     })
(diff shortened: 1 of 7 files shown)


server/models/functions.sql
```diff
@@ -56,6 +57,23 @@ BEGIN
...
+  projectId INT,
+  currencyId INT,
+  date TIMESTAMP
+)
+RETURNS DECIMAL(19, 8) DETERMINISTIC
+BEGIN
+  DECLARE _enterprise_id smallint(5) unsigned;
+  SELECT enterprise_id INTO _enterprise_id FROM project WHERE id = projectId;
+
+  RETURN (SELECT GetExchangeRate(_enterprise_id, currencyId, date ));
```
jniles: Just FYI, you could simplify this by putting it in one statement:
```sql
RETURN (SELECT GetExchangeRate(p.enterprise_id, currencyId, date) FROM project p WHERE p.id = projectId);
```

server/controllers/finance/reports/debtors/index.js
```diff
@@ -85,19 +86,35 @@ async function queryContext(params = {}) {
   // format the dates for MySQL escape
   const dates = _.fill(Array(5), params.date);
   const data = {};
+  const currencyId = db.escape(params.currency_id);
```
jniles, line 89: Nice escaping! 👍

server/controllers/finance/reports/debtors/index.1.js
```diff
@@ -0,0 +1,171 @@
+/**
```
jniles: Why does this file exist..?

jniles (approved):
This looks really good!  It looks like you accidentally pushed an extra file - could you remove that?  Then it can be merged!

server/controllers/finance/reports/debtors/index.1.js
jeremielodi: it should be removed

jniles (approved):
Excellent work!

bors r+

PR #3594, opened 2019-02-12
fix(prices): show % only when is_percentage

jniles:
The price list items now only show percentage when it actually is a percentage.


jniles:

jeremielodi:
Yes of course

jeremielodi (commented):
@jeremielodi approved this PR!

PR #3596, opened 2019-02-15
refractor(currency select componet) update to angular new style

### LINKED ISSUE #3387: Update bhCurrencySelect to latest Angular best practices
We rely on the [`bhCurrencySelect`](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/components/bhCurrencySelect.js) in several places, but we didn't follow best practices when designing it.  We need to update it so that it is up to date with our latest component standards.  To fix this issue, we'll need to:
 1. Remove the [validateTrigger](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/components/bhCurrencySelect.js#L8) binding and replace it with a `childForm.$submitted` listener.
 2. Make the currencyId binding one-way.
 3. Use the `$onChanges` component binding to listen for changes from the parent controller and update appropriately.
 4. Add unit tests

jeremielodi:
closes #3387

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -316,6 +316,7 @@
             "CREDITOR_GROUP": "Creditor Group",
             "CREDITOR_SUBCRIBED": "Subscribed Creditors",
             "CURRENCIES": "Currencies",
+            "CONVERTION_RATE" : "Conversion rate",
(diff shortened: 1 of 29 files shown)


jniles (approved):
This looks excellent!

bors r+

PR #3597, opened 2019-02-16
Add inline filter for inventory configuration

mbayopanda:
This PR add inline filters in the inventory configuration module 

image

### DIFF
--- a/client/src/modules/inventory/configuration/groups/groups.html
+++ b/client/src/modules/inventory/configuration/groups/groups.html
@@ -30,12 +30,11 @@
             <div class="input-group">
               <span class="input-group-addon"><i class="fa fa-filter"></i></span>
               <input
-                type="text" 
                 ng-model="GroupsCtrl.searchValue"
                 type="search"
                 class="form-control input-sm"
                 style="display: inline"
-                placeholder="{{ 'FORM.LABELS.SEARCH' | translate }}">
+                translate-attr="{ placeholder: 'FORM.LABELS.SEARCH' }">
             </div>
           </th>
           <th></th>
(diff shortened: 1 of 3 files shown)


client/src/modules/inventory/configuration/groups/groups.html
```diff
@@ -22,18 +22,31 @@
...
+          <th>
+            <div class="input-group">
+              <span class="input-group-addon"><i class="fa fa-filter"></i></span>
+              <input
+                type="text" 
+                ng-model="GroupsCtrl.searchValue"
+                type="search"
+                class="form-control input-sm"
+                style="display: inline"
+                placeholder="{{ 'FORM.LABELS.SEARCH' | translate }}">
```
jniles: Just so you know, when you are using regular HTML inputs (`<input>`), you can get a performance increase by using [`translate-attr`](https://angular-translate.github.io/docs/#/api/pascalprecht.translate.directive:translate-attr).  This is similar to using the directive instead of the filter, as it reduces `$watch` statements.

In this case, it would be:
```html
translate-attr="{ placeholder: 'FORM.LABELS.SEARCH' }"
```

client/src/modules/inventory/configuration/types/types.html
```diff
@@ -19,12 +19,27 @@
...
+        </tr>
+        <tr>
+          <td></td>
+          <td>
+            <div class="input-group">
+              <span class="input-group-addon"><i class="fa fa-filter"></i></span>
+              <input
+                type="text" 
+                ng-model="TypesCtrl.searchValue"
+                type="search"
```
jniles, line 31: Looks like you have `type` here twice - once as "text" and another as "search".

jniles (commented):
Two comments that are shared across all three inputs.  After they are addressed, it could be merged.

What do you think about (in the future) making the groups configuration into a `uiGrid`?  I think it would make more sense than the current layout.

mbayopanda:
```
What do you think about (in the future) making the groups configuration into a `uiGrid`? I think it would make more sense than the current layout.
```
You are talking about having more than one `uiGrid` on a page ?

jniles (approved):
LGTM.

bors r+

PR #3604, opened 2019-02-18
feat(report) add patient's age to unpaid report

### LINKED ISSUE #3555: Add patient age to Unpaid Invoices report
We've received a request from our clients in the field to add the age of the patient to the unpaid invoices report.  This will help them better identify the patients in the hospital.

jeremielodi:
closes #3555

### DIFF
--- a/server/lib/util.js
+++ b/server/lib/util.js
@@ -28,6 +28,7 @@ exports.dateFormatter = dateFormatter;
 exports.execp = execp;
 exports.format = require('util').format;
 
+exports.calcualteAge = calcualteAge;
 exports.renameKeys = renameKeys;
 
 exports.roundDecimal = roundDecimal;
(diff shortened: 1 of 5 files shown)


server/lib/util.js
```diff
@@ -235,6 +236,13 @@ function formatCsvToJson(filePath) {
   return defer.promise;
 }
 
+// calcualte age function
+function calcualteAge(dateString) {
+  const birthday = +new Date(dateString);
```
jniles: Let's use `moment` for this.
```js
return moment().diff(dob, 'years');  
```

We already do this on the client, so it will make the presentation consistent between the two.

jeremielodi: okays, it's more simple

jeremielodi:
aged

jniles (approved):
LGTM.

bors r+

PR #3610, opened 2019-02-19
Improve Cashbox Report

### LINKED ISSUE #3595: Rewrite Cash Report
The Cash Report provides a simplistic view of the cashbox as a series of entries and exits.  This is helpful for non-accountants when reviewing their daily reports to compare to the system.

However, the report is very outdated.  [Here is the server-side controller](https://github.com/IMA-WorldHealth/bhima/blob/master/server/controllers/finance/reports/cashReport/index.js).  There are a number of issues with it:
 1. It should use period totals to establish an opening balance and running balance.
 2. The opening balance is not in the cashbox's or enterprise's currency - it is in USD.
 3. The report should _optionally_ include unposted records, and should provide an italics representation like the account report.

There is also a bug where the dates are not respected.

jniles:
This PR improves the cashbox report in the following ways:
 1. The dates (from - to) are now respected in the search query.
 2. The currencies are rendered in the correct cashbox currency.
 3. We include a real opening balance and closing balance.
 4. The user can identify the difference between posted and unposted records via italics applied to unposted records.

We've also centralized the `getAccountTransactions()` call used between multiple reports into a single shared file `accounts/transactions`.  We've also rewritten the query for the combined posting_journal and general_ledger such that the WHERE statement is executed in subqueries, greatly improving the performance of the combined query.

Finally, the `getAccountTransactions()` function now sorts first based on `created_at` date, then on `trans_date`.  This makes back-dated transactions show up later in the rendering.  For example, users frequently back-date the transfer from the cashbox into the main coffre, and it makes sense to show these at the bottom of the report.

Closes #3595.

### DIFF
--- a/server/controllers/finance/accounts/transactions.js
+++ b/server/controllers/finance/accounts/transactions.js
@@ -1,0 +1,207 @@
+/**
+ * @overview AccountTransactions
+ *
+ * @description
+ * This file provides a common tool for reading the transations associated with an account.
+ * It provides a re-usable interface to access this information, and supports:
+ *  1. Inclusion of non-posted records
+ *  2. Exchange Rate Calculation
+ *  3. Running Balance Calculation
+ *
+ */
+
+const _ = require('lodash');
+const db = require('../../../lib/db');
+const FilterParser = require('../../../lib/filter');
+const Exchange = require('../exchange');
+const Accounts = require('.');
+
+/**
+ * @function getGeneralLedgerSQL
+ *
+ * @description
+ * Used by the getAccountTransactions() function internally.  The internal SQL
+ * just pulls out the values tied to a particular account.
+ *
+ * The exchange rate logic is complicated.  Here is the basic idea:
+ *  1. If you are in the enterprise currency, you want the calculation to
+ *    use ${amount} / rate to calculate the values.
+ *  2. If the you not in the enterprise currency, you want to use ${amount} * rate
+ *    to convert "back" to the enterprise currency.
+ *
+ */
+function getGeneralLedgerSQL(options) {
+  const filters = new FilterParser(options);
+
+  // return the proper way to convert the values depending on if we are in the
+  // exchange rate currency or not.
+  const enterpriseCurrencyColumns = `
+    (debit / rate) AS exchangedDebit, (credit / rate) AS exchangedCredit,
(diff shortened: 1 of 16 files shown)


jniles:

server/controllers/finance/accounts/transactions.js
```diff
@@ -0,0 +1,225 @@
...
+ *  2. Exchange Rate Calculation
+ *  3. Running Balance Calculation
+ *
+ */
+
+const _ = require('lodash');
+const db = require('../../../lib/db');
+const FilterParser = require('../../../lib/filter');
+const Exchange = require('../exchange');
+const Accounts = require('.');
```
mbayopanda, line 17: What does it means ? The current file or all files of the current directory.

If it is the current file, why do we need to import it ?

server/controllers/finance/accounts/transactions.js
```diff
@@ -0,0 +1,225 @@
...
+  const filters = new FilterParser(options);
+
+  options.isEnterpriseCurrency = options.isEnterpriseCurrency || false;
+
+  // return the proper way to convert the values depending on if we are in the
+  // exchange rate currency or not.
+  const enterpriseCurrencyColumns = `
+    (debit / rate) AS exchangedDebit, (credit / rate) AS exchangedCredit,
+    (balance / rate) AS exchangedBalance, @cumsum := (balance / rate) + @cumsum AS cumsum
+  `;
```
mbayopanda, line 43: The request is so clear 👍

mbayopanda (commented):
LGTM 👍 

I was not able to test it with real data of imck because the cashbox selection did not work

server/controllers/finance/accounts/transactions.js
jniles: Aha.  That is confusing.  It actually imports `./index.js`.  This file is `./transactions.js`.

jniles:
@mbayopanda, thank you for thoroughly testing this module.  It appears that the `bhCashboxSelect` component [defaults to `restrict-to-user="true"`](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/components/bhCashboxSelect.js#L36) and so we couldn't see the data due to not having the correct permissions.

I've fixed this by turning the restriction off.

PR #3611, opened 2019-02-19
feat(component) bh-bebtor-group-history

### LINKED ISSUE #3609: Make display component for debtor group history (bhDebtorGroupHistory)
We already added the logic to track the historical values of debtor group history changes in the database.  However, this information is not presented to the client anywhere.  It would be ideal to have a component that displays the changes to the debtor group's history.

The call signature would look something like this:
```html
<bh-debtor-group-history
  debtor-uuid="vm.debtorUuid"
  limit="5">
</bh-debtor-group-history>
```

This would display the last 5 movements of the debtor between debtor groups.  We could then use it on the patient record page, the patient edit page, and other places to display the history of the debtor.

jeremielodi:
closes #3609

### DIFF
--- a/client/src/js/components/bhDebtorGroupHistory.js
+++ b/client/src/js/components/bhDebtorGroupHistory.js
@@ -1,0 +1,38 @@
+angular.module('bhima.components')
+  .component('bhDebtorGroupHistory', {
+    templateUrl : 'modules/templates/bhDebtorGroupHistory.tmpl.html',
+    controller : bhDebtorGroupHistoryController,
+    transclude : true,
+    bindings : {
+      debtorUuid : '<',
+      limit : '@?',
+    },
+  });
+
+bhDebtorGroupHistoryController.$inject = ['DebtorGroupService', 'NotifyService'];
+/**
+ * Debtor group history component
+ *
+ */
+function bhDebtorGroupHistoryController(DebtorGroup, Notify) {
+  const $ctrl = this;
+
+  // fired at the beginning
+  $ctrl.$onInit = () => {
+    $ctrl.limit = $ctrl.limit || 5;
+    loadHistory();
+  };
+
+  $ctrl.$onChanges = () => {
+    loadHistory();
+  };
+
+  function loadHistory() {
+    const parameters = { limit : $ctrl.limit };
+    DebtorGroup.history($ctrl.debtorUuid, parameters)
+      .then(groupChanges => {
+        $ctrl.groupChanges = groupChanges;
+      })
+      .catch(Notify.handleError);
+  }
+}
--- a/client/src/modules/debtors/groups.service.js
+++ b/client/src/modules/debtors/groups.service.js
@@ -20,6 +20,7 @@ function DebtorGroupService(Modal, Session, $translate, Api) {
   service.updateInvoicingFees = updateInvoicingFees;
   service.updateSubsidies = updateSubsidies;
   service.invoices = invoices;
+  service.history = history;
   service.remove = service.delete;
 
   service.manageInvoicingFees = manageInvoicingFees;
@@ -114,6 +115,12 @@ function DebtorGroupService(Modal, Session, $translate, Api) {
   function invoices(uuid, parameters) {
     const url = `${baseUrl}${uuid}/invoices`;
     return service.$http.get(url, { params : parameters })
+      .then(service.util.unwrapHttpResponse);
+  }
+
+  function history(debtorUuid, parameters) {
+    const url = `${baseUrl}history/${debtorUuid}`;
+    return service.$http.get(url, { params : parameters || {} })
       .then(service.util.unwrapHttpResponse);
   }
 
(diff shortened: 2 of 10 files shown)


client/src/js/components/bhDebtorGroupHistory.js
```diff
@@ -0,0 +1,33 @@
+angular.module('bhima.components')
+  .component('bhDebtorGroupHistory', {
+    templateUrl : 'modules/templates/bhDebtorGroupHistory.tmpl.html',
+    controller : bhDebtorGroupHistoryController,
+    transclude : true,
+    bindings : {
+      debtorUuid : '@?',
+      limit : '<',
```
jniles: Looks like this should be optional.

client/src/js/components/bhDebtorGroupHistory.js
```diff
@@ -0,0 +1,33 @@
+angular.module('bhima.components')
+  .component('bhDebtorGroupHistory', {
+    templateUrl : 'modules/templates/bhDebtorGroupHistory.tmpl.html',
+    controller : bhDebtorGroupHistoryController,
+    transclude : true,
+    bindings : {
+      debtorUuid : '@?',
```
jniles: Looks like the `debtorUuid` should be required.

client/src/js/components/bhDebtorGroupHistory.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ */
+function bhDebtorGroupHistoryController(DebtorGroupService, Notify) {
+  const $ctrl = this;
+
+  // fired at the beginning
+  $ctrl.$onInit = () => {
+    $ctrl.limit = $ctrl.limit || 5;
+    loadHistory();
+  };
```
jniles, line 24: Let's also put a [`$onChanges`](https://docs.angularjs.org/guide/component#component-based-application-architecture) binding so that we reload the history when the `debtorUuid` changes.

client/src/modules/debtors/groups.service.js
```diff
@@ -117,5 +118,11 @@ function DebtorGroupService(Modal, Session, $translate, Api) {
       .then(service.util.unwrapHttpResponse);
   }
 
+  function history(debtorUuid, limit) {
+    const url = `${baseUrl}history/${debtorUuid}/${limit}`;
```
jniles: Could you make the `limit` into a query string?  That way, it would be option and you could either see the entire history with `/history/:uuid` or a slice of the history with `/history/:uuid?limit=X`.

client/src/modules/patients/registration/registration.html
```diff
@@ -101,7 +92,7 @@
                       <div class="help-block text-right">
                         <p>
                           <a ng-click="PatientRegCtrl.toggleFullDate()" href>
-                            <i class="fa fa-circle-o" ng-class="PatientRegCtrl.dateIndicatorIcon"></i>
+                            <i class="fa fa-calendar-o"></i>
```
jniles: I think this might change the display @mbayopanda was going for - in one state, it is supposed to be a circle that is filled in and in the other it is supposed to be a circle that is empty.

client/src/modules/patients/registration/registration.html
```diff
@@ -2,17 +2,8 @@
   <div class="bhima-title">
     <ol class="headercrumb">
       <li class="static" translate>TREE.HOSPITAL</li>
-      <li class="title" translate>PATIENT_REG.PAGE_TITLE</li>
+      <li class="title" translate> PATIENT_REG.PAGE_TITLE</li>
     </ol>
-
-    <div class="toolbar">
```
jniles, line 8: I think all the changes to this page must be a mistake.  Can you revert them?

client/src/modules/templates/bhDebtorGroupHistory.tmpl.html
```diff
@@ -0,0 +1,16 @@
+<div class="panel panel-default">
+    <p><u><span translate>DEBTOR_GROUP.CHANGE</span></u></p>
+
+    <ul class="list-unstyled">
+      <li ng-repeat="change in $ctrl.groupChanges | orderBy:'-created_at'">
```
jniles: Since you do the ORDER BY in SQL, there isn't a need for this here.

client/src/modules/templates/bhDebtorGroupHistory.tmpl.html
```diff
@@ -0,0 +1,16 @@
...
+        <i class="fa fa-clock-o"></i>
+        <b><span translate>FORM.LABELS.FROM</span></b> {{change.group_prev}}
```
jniles: Let's make this into an arrow to make it super clear:  Something like
```html
{{change.group_prev}} &#8594; {{change.group_next}}
```
which will render &#8594; as the arrow.

jniles (commented):
@jeremielodi this is a good start! 

It looks like a mistake happened in the Patient Registration page, let's restore it back to the old values.

Could you put this component on the patient edit page?  Right under this:

screenshot 2019-02-19 at 18 33 36


That way we can see how it looks!

jeremielodi:
group change

jniles (approved):
LGTM.  Good work.

bors r+

PR #3616, opened 2019-02-21
Ward configuration

mbayopanda:
This PR add the ward configuration module which allow to perform CRUD on wards, rooms and beds

image


client/src/js/components/bhRoomSelect/bhRoomSelect.js
```diff
@@ -0,0 +1,48 @@
...
+    Room.read(null, { ward_uuid : $ctrl.wardUuid })
+      .then(rooms => {
+        $ctrl.rooms = rooms;
+      })
+      .catch(Notify.handleError);
+  };
+
+  $ctrl.$onChanges = (changes) => {
+    if (changes.wardUuid) {
+      Room.read(null, { ward_uuid : $ctrl.wardUuid })
```
jniles: This works which is the most important part.

I am a bit worried that `$ctrl.wardUuid` might not be updated before the `$onChanges` call, and it would use the old value instead of the newly changed value.  In our tests this does happen, but we can make sure that we are definitely using the new value by calling `read()` with the `wardUuid` that came from outside the component.  Like this:

```js
if (changes.wardUuid && changes.wardUuid.currentValue) {
  Room.read(null, { ward_uuid : changes.wardUuid.currentValue })
   // ... etc
}
```

test/integration/ward/ward.js
```diff
@@ -1,14 +1,14 @@
 /* global expect, agent */
-
-const helpers = require('./helpers');
+const helpers = require('../helpers');
 
 /*
  * The /wards API endpoint
  *
- * This test suite implements full CRUD on the /projects HTTP API endpoint.
+ * This test suite implements full CRUD on the /wards HTTP API endpoint.
```
jniles, line 7: :dancer:

server/models/schema.sql
```diff
@@ -1679,6 +1679,37 @@ CREATE TABLE `ward`(
...
+ `room_type_id` SMALLINT(5) UNSIGNED NULL,
+  PRIMARY KEY(`uuid`),
+  UNIQUE KEY `room_label_0` (`label`, `ward_uuid`),
+  FOREIGN KEY (`ward_uuid`) REFERENCES ward (`uuid`),
+  FOREIGN KEY (`room_type_id`) REFERENCES room_type (`id`)
+)ENGINE=InnoDB  DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+DROP TABLE IF EXISTS `bed`;
+CREATE TABLE `bed`(
+ `id` SMALLINT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
```
jniles, line 1704: As long as our hospital doesn't have more than 65535 beds, we are good 💃.

server/models/schema.sql
```diff
@@ -1679,6 +1679,37 @@ CREATE TABLE `ward`(
...
+ `label` VARCHAR(20) NOT NULL,
+ `room_uuid` BINARY(16) NOT NULL,
+ `is_occupied` TINYINT(1) NOT NULL DEFAULT 0,
+  PRIMARY KEY(`id`),
+  UNIQUE KEY `bel_label_0` (`label`, `room_uuid`),
```
jniles: Should this key be called `bed_label_0`?

server/models/schema.sql
```diff
@@ -1679,6 +1679,37 @@ CREATE TABLE `ward`(
...
+  PRIMARY KEY(`uuid`),
+  UNIQUE KEY `room_label_0` (`label`, `ward_uuid`),
+  FOREIGN KEY (`ward_uuid`) REFERENCES ward (`uuid`),
+  FOREIGN KEY (`room_type_id`) REFERENCES room_type (`id`)
+)ENGINE=InnoDB  DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+DROP TABLE IF EXISTS `bed`;
+CREATE TABLE `bed`(
+ `id` SMALLINT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
+ `label` VARCHAR(20) NOT NULL,
```
jniles: Let's make this VARCHAR a bit bigger - something like 120 should be good.

server/models/schema.sql
```diff
@@ -1679,6 +1679,37 @@ CREATE TABLE `ward`(
...
+ `room_uuid` BINARY(16) NOT NULL,
+ `is_occupied` TINYINT(1) NOT NULL DEFAULT 0,
```
jniles, line 1707: Can we add `user_id`, `created_at`, and `updated_at` to this table?  All of these would be set automatically on the server when the bed is created.  It would help us answer questions in production if a bed randomly appears and no one remembers making it.

server/models/schema.sql
```diff
@@ -1679,6 +1679,37 @@ CREATE TABLE `ward`(
...
+CREATE TABLE `room_type`(
+ `id` SMALLINT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
+ `label` VARCHAR(50) NOT NULL,
+  PRIMARY KEY(`id`)
+)ENGINE=InnoDB  DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+DROP TABLE IF EXISTS `room`;
+CREATE TABLE `room`(
+ `uuid` BINARY(16) NOT NULL,
+ `label` VARCHAR(50) NOT NULL,
```
jniles: Let's make all these VARCHARs a bit bigger.

server/controllers/medical/ward/bed.js
```diff
@@ -0,0 +1,92 @@
...
+
+  db.exec(sql, [id])
+    .then(() => {
+      res.sendStatus(204);
+    })
+    .catch(next);
+}
+
+// get all beds
+function read(req, res, next) {
```
jniles, line 55: We will eventually want to do more complex queries on these beds (for example, list all beds in the ward X).

Let's make a common function called `lookupBeds(options)` that will use `FilterParser` to filter this beds.  Like this:
```js
function lookupBeds(options) {
  db.convert(options, ['ward_uuid', 'room_uuid']);
  const filters = new FilterParser(options);
  const sql = `
    SELECT b.id, b.label, 
      BUID(r.uuid) as room_uuid, r.label AS room_label,
      BUID(w.uuid) AS ward_uuid, w.name AS ward_name, w.description,
      s.name as service_name
    FROM bed b
    JOIN room r ON r.uuid = b.room_uuid
    JOIN ward w ON w.uuid = r.ward_uuid
    LEFT JOIN service s ON s.id = w.service_id
  `;

   filters.equals('id');
   filters.equals('ward_uuid');
   // etc ...
}
```

That way, our work will be much easier in the future.

client/src/modules/ward_module/ward_module.routes.js
```diff
@@ -0,0 +1,26 @@
+angular.module('bhima.routes')
+  .config(['$stateProvider', wardModuleStateProvider]);
+
+function wardModuleStateProvider($stateProvider) {
+  $stateProvider
+    .state('wardModuleConfiguration', {
+      url : '/ward_module/configuration',
```
jniles: I think that `/wards/configuration` is cleaner than `ward_module/configuration`.  How do you feel about switching this URL to `wards`?

client/src/modules/ward_module/ward_module.service.js
```diff
@@ -0,0 +1,9 @@
+angular.module('bhima.services')
+  .service('WardModuleService', WardModuleService);
+
+WardModuleService.$inject = [];
+
+function WardModuleService() {
```
jniles: Do you ever use WardModuleService?  If not, let's go ahead and remove it.

server/controllers/medical/ward/room.js
```diff
@@ -0,0 +1,96 @@
...
+
+  db.exec(sql, uuid)
+    .then(() => {
+      res.sendStatus(204);
+    })
+    .catch(next);
+}
+
+// get all rooms
+// todo: add nb of beds
```
jniles: I think you did this :)

jniles (commented):
Just a few comments here.  Looks pretty good!

client/src/modules/ward_module/ward_module.routes.js
mbayopanda: 😄 there was already the previous `/ward`, i will fix that 👍

mbayopanda:
@jniles could you review this PR again please...

server/models/schema.sql
jniles: This still needs to be addressed 🕺

jniles (commented):
Looks good!   Looks like you missed one thing, then we can merge this.

mbayopanda:
@jniles could look at that again please

jniles (approved):
LGTM.

bors r+

PR #3617, opened 2019-02-21
feat(cash credit not) up to date

### LINKED ISSUE #3269: Cash cancellation interface looks out of date
This is the interface to cancel a cash payment:
better debit note interface
_Fig 1: Debit Note Interface_

Compare this interface to the Credit Note interface:
creditnoteinterface
_Fig 2: Credit Note Interface_

The Credit Note interface is a lot richer, providing details about the invoice that is about to be canceled in a familiar way.  It also clearly explains the history of the invoice.  Finally, it includes the human readable identifiers of the invoice and of the patient.  

We could do a similar thing for canceling cash payments.  It would be nice to see:
 1. The patient's reference
 2. The invoice(s) that was being paid
 3. A warning that this action is permanent and irreversible
 4. A link to the patient.

jeremielodi:
closes #3269

### DIFF
--- a/client/src/modules/cash/modals/modal-cancel-cash.ctrl.js
+++ b/client/src/modules/cash/modals/modal-cancel-cash.ctrl.js
@@ -2,12 +2,14 @@ angular.module('bhima.controllers')
   .controller('ModalCancelCashController', ModalCancelCashController);
 
 ModalCancelCashController.$inject = [
-  '$filter', '$state', '$uibModalInstance',
+  '$filter', '$state', '$uibModalInstance', 'bhConstants',
   'CashService', 'data', 'VoucherService', 'NotifyService',
 ];
 
-function ModalCancelCashController($filter, $state, Instance, Cash, data, Vouchers, Notify) {
+function ModalCancelCashController($filter, $state, Instance, Constants, Cash, data, Vouchers, Notify) {
   const vm = this;
+
+  vm.Constants = Constants;
 
   vm.cancelCash = {};
   vm.submit = submit;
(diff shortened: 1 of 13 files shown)


jeremielodi:
capture

jeremielodi:
fina

client/src/modules/cash/modals/modal-cancel-cash.ctrl.js
```diff
@@ -2,36 +2,58 @@ angular.module('bhima.controllers')
...
     .catch(Notify.handleError);
 
   function submit(form) {
-     // stop submission if the form is invalid
+    // stop submission if the form is invalid
     if (form.$invalid) { return; }
 
-    return Vouchers.reverse(vm.cancelCash)
-      .then(function () {
+    Vouchers.reverse(vm.cancelCash)
```
jniles: By removing the `return` of the promise, this breaks the loading indicator on the submit button.  It would be better to put this back and add a "0" to the other `return`.

jniles (commented):
@jeremielodi this addresses the majority of the points in #3269.  It also introduces a bug on the loading button which should be fixed.

We still need to show the invoices that were paid as requested in the issue.  Ideally, It should be the invoice that goes in the table, not the cash payment (we can only reverse a single cash payment at a time, so we'll never have more than one in a table).  This would be similar to the credit note, where the credit note shows each inventory item that the invoice was for.  In this case, we should show each invoice that the payment was for.  It was be nice to link them as well with something like `bhReferenceLink`.

Also, while you are here, could you remove the  `<div class="bhima-title">`?  It makes the modal head much larger than any of our other modals.

Great first step!

kwilu:
Hello @jeremielodi, it seems like you have referenced #3269 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #3269.`.

```
Author: zulipbot

    Fixes #51.
```

Thank you for your contributions to Zulip!

jniles:
I have made a few improvements here to try and clean both the code and presentation up:
 1. The original code was cut and paste, so sometimes it mentioned "invoice" when it actually meant "cash payment".  I've removed all those references and changed them to "payment" or "cash".
  2. The invoice details and the cash details have been separated, so that the cashbox, date, patient, etc, are all put in the top section of the modal, and then the list of invocies is presented at the bottom.
 3. We hide the table if there is a caution and just show "caution amount" to avoid a broken-looking modal.
 4. There are now totals on the bottom of the table.

See it below in action:

jniles:
@lomamech could you review this PR?

PR #3618, opened 2019-02-22
feat(depot management) move to stock module

### LINKED ISSUE #3614: Move "Depots Registry" into Stock
Supersedes https://github.com/IMA-WorldHealth/bhima/issues/725.

While working on the documentation for Stock Management, it occurred to me that depots are simply containers for stock.  BHIMA does not use the depot concept in any other context but stock management.  And all stock in BHIMA is associated with a depot.

For this reason, I propose that we move the Depots Registry into the Stock folder.

jeremielodi:
closes #3614


jniles (approved):
LGTM.

bors r+

PR #3619, opened 2019-02-25
feat(patient invoice) add posted label

### LINKED ISSUE #3473: Add "reversed" and "posted" labels to the receipt modal
If a receipt (cash payment, invoice, voucher) has been posted or reversed, we should add those pieces of information to the receipt modal.  It comes automatically with the record, so why not?

I'm thinking `<span class="label label-primary" translate>FORM.LABELS.POSTED</span>` and `label-danger` for the reversed flag.  They would appear in the top right of the receipt modal pictured below:
addnewpostedreversedtoreceiptmodal

jeremielodi:
closes #3473


jeremielodi:
capture

jeremielodi:
capture

jniles:
@jeremielodi, please read the discussion on #3473.  The label needs to be on the modal, not on the PDF in the modal.

jeremielodi:
posted

client/src/js/services/receipts/ReceiptModal.js
```diff
@@ -41,7 +41,9 @@ function ReceiptModal(Modal, Receipts) {
    * @param {String} uuid             Target invoice UUID
    * @param {Boolean} notifyCreated   Defines if a success message should be shown for entity creation
    */
-  function invoice(uuid, notifyCreated) {
+  async function invoice(uuid, notifyCreated) {
```
jniles: Unfortunately, we can't use async/await on the client without some hacks, due to the way AngularJS's $digest loop works.  For reference, you can see [this post](https://medium.com/@alSkachkov/using-async-await-function-in-angular-1-5-babel-6-387f7c43948c),

server/controllers/finance/patientInvoice.js
```diff
@@ -352,3 +353,16 @@ function safelyDeleteInvoice(guid) {
       return transaction.execute();
     });
 }
+
+function isPosted(req, res, next) {
```
jniles: Oof.  This is a huge query.  We should probably store the "posted" value in the invoice table - I thought it already existed on there.

Let's do this:
 1. Update the schemas of the `invoice`, `cash` and `voucher` tables to include `posted`.  It should be a NOT NULL 1 or 0 with a default to 0.
 2. Add queries to [the server-side trialbalance controller](https://github.com/IMA-WorldHealth/bhima/blob/master/server/controllers/finance/trialBalance/index.js#L93) to update the `posted` status of the invoices, cash payments, and/or vouchers where their `uuid` is found in the `record_uuid` of the posted records.
 3. We can just use the `posted` value directly from the invoice table then.

That way, we aren't locking the entire general ledger every time we do a lookup of an invoice.

client/src/js/services/receipts/ReceiptModal.js
```diff
@@ -43,31 +43,34 @@ function ReceiptModal(Modal, Receipts) {
...
-      resolve : {
-        receipt : function receiptProvider() { return { promise : invoiceRequest }; },
-        options : function optionsProvider() { return options; },
-      },
-    };
-
-    const configuration = angular.extend(modalConfiguration, invoiceProvider);
-    const instance = Modal.open(configuration);
-    return instance.result;
+    return Receipts.invoicePosted(uuid).then(isPosted => {
```
jniles: If we add a `posted` column to the invoice table, we can use the `InvoiceService.read()` method to get both the `posted` and `reversed` values for this invoice.

We should also put this into the `invoiceProvider` instead of firing it before the modal is opened.  It will improve the application's performance.
```js
const invoiceProvider = {
  resolve : {
    receipt : () => ({ promise : invoiceRequest }),
    options : () => options,
    invoice : () => Invoice.read(uuid),
  }
};
```

That way, the modal can open, then we can just check the `data.reversed` and `data.posted` values to figure out if we should display the `posted` and `reversed` signs.

server/models/procedures/trial_balance.sql
```diff
@@ -296,8 +299,24 @@ BEGIN
   GROUP BY fiscal_year_id, period_id, account_id
   ON DUPLICATE KEY UPDATE credit = credit + VALUES(credit), debit = debit + VALUES(debit);
 
+  SET _record_uuid = (SELECT record_uuid FROM stage_trial_balance_transaction LIMIT 1);
```
jniles: This is the right location to make changes to the code.

The Trial Balance allows a user to post multiple transactions at a singe time, but this will only mark a single record as posted because of the LIMIT 1.  Much better to do:
```sql
UPDATE invoice JOIN stage_trial_balance_transaction ON invoice.uuid = stage_trial_balance_transaction.record_uuid SET invoice.posted = 1;
```
... for both cash, invoices, and vouchers.

server/models/schema.sql
```diff
@@ -1556,6 +1557,7 @@ CREATE TABLE `invoice` (
   `reversed`            TINYINT NOT NULL DEFAULT 0,
   `edited`              TINYINT NOT NULL DEFAULT 0,
   `created_at`          TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `posted`            TINYINT(1) UNSIGNED NOT NULL DEFAULT 0,
```
jniles, line 1560: Please add a `posted` flag to the` voucher` table as well.

jniles (commented):
This is really good!   We are almost there.  Just a few bugs to fix in the trial balance limit and corresponding queries.

kwilu:
Hello @jeremielodi, it seems like you have referenced #3473 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #3473.`.

```
Author: zulipbot

    Fixes #51.
```

Thank you for your contributions to Zulip!

jniles:
Interesting, @kwilu.    I didn't know you had that functionality.

server/models/procedures/trial_balance.sql
```diff
@@ -296,8 +299,24 @@ BEGIN
...
+
+  -- Let specify that this invoice or the cash payment is posted
+  SELECT COUNT(uuid) INTO isInvoice  FROM invoice  WHERE invoice.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+  SELECT COUNT(uuid) INTO isCash  FROM cash  WHERE cash.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+
+  IF isInvoice > 0 THEN
+    UPDATE invoice SET posted = 1 WHERE uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+  END IF;
+
+  IF isCash > 0 THEN
```
jniles, line 316: Let's add "isVoucher" in here as well.

jniles (approved):
Just need to add "isVoucher" and then we'll be good!

jniles (approved):
LGTM.

bors r+

PR #3623, opened 2019-03-02
feat(patient record): patient visits management

mbayopanda:
This PR upgrades the patient visits management in BHIMA by : 
- Adding a new registry for visits 

image

- Allow to specify the patient in the visit modal

image

- Adding new information in the admission visit modal : 
  - New Case or not
  - Pregnant
  - Refered
  - In the health zone or not
  - The service

- Upgrading the discharge modal : 

image

- Upgrading the visit view in the patient details : 

image


mbayopanda:
@jniles

client/src/js/components/bhDischargeSelect/bhDischargeSelect.js
```diff
@@ -0,0 +1,28 @@
+angular.module('bhima.components')
+  .component('bhDischargeSelect', {
+    bindings : {
+      onChangeCallback : '&',
```
jniles, line 5: Let's add a parameter to pass in the value here.  That way, we can preset the value in future functionality.  It would be something like:
```js
value : '<?',
```
and then later:
```js
.then(data => {
  const hasDefaultValue = angular.isDefined($ctrl.value);
  if (!hasDefaultValue && data.length) {
    $ctrl.value = data[0].id;
  }
})
```

client/src/modules/patients/visits/templates/duration.cell.html
```diff
@@ -0,0 +1,3 @@
+<div class="ui-grid-cell-contents">
+  {{ row.entity.durationAge }}
```
jniles: Do you need to make a new cell for this?  Why not just let ui-grid render this normally?

Also, remember that we have [angular-moment](https://github.com/urish/angular-moment), so instead of creating a new function in `util.js`, we could just make a grid column with:
```js
'cellFilter': 'amDurationFormat:"day"',
```
and that should be all that is necessary.

server/config/routes.js
```diff
@@ -859,4 +863,7 @@ exports.configure = function configure(app) {
   app.put('/account_reference_type/:id', accountReferenceType.update);
   app.delete('/account_reference_type/:id', accountReferenceType.delete);
 
+  // API for discharge type
+  app.get('/discharge_type', dischargeType.list);
```
jniles: Can we make this URL plural?  I think we've gotten out of the habit of doing it, but [most REST folks seem to like naming resources as plural](https://stackoverflow.com/q/6845772) and we've mentioned it in our [API design guide](https://github.com/IMA-WorldHealth/bhima/wiki/Design:-Server-CRUD-APIs).

client/src/i18n/en/patient_records.json
```diff
@@ -40,7 +42,39 @@
...
+      "VISITS_INFO": "Visits Information",
+      "ADMISSION_REGISTRY":"Admissions Registry",
+      "ONLY_HOSPITALIZED":"Only hospitalized patients",
+      "NEW":"New Visit",
+      "ALREADY_ADMITTED":"The patient is already admitted, and his visit is not discharged",
+      "PREGNANT":"Pregnant",
+      "HEALTH_ZONE":"health Zone",
+      "CASE":"Type of case",
+      "REFERED":"Refered",
+      "IS_PREGNANT":"Does she pregnant ?",
```
jniles: Better English translation would be "Is the patient pregnant?"

client/src/i18n/en/form.json
```diff
@@ -133,6 +133,10 @@
             "ASSIGNED_COST_CENTER" : "Assigned a cost center",
             "ASSIGNED_PROFIT_CENTER" : "Assigned a profit center",
             "ASSIGNED_PROJECTS" : "Assigned to projects",
+            "VISIT_RECORDED_SUCCESSFULLY":"Visit Successfully Recorded",
+            "VISIT_RECORDED_UNSUCCESSFULLY":"Visite Not Recorded",
```
jniles: This should be "visit", not "visite".

client/src/i18n/en/patient_records.json
```diff
@@ -40,7 +42,39 @@
...
+      "IS_IN_HEALTH_ZONE":"In the health zone (ZS)",
+      "IS_NOT_IN_HEALTH_ZONE":"Not in the health zone (HZ)",
+      "IS_NEW_CASE":"New Case",
+      "IS_NOT_NEW_CASE":"Old Case",
+      "IS_REFERED":"Does he or she a refered patient ?"
```
jniles: A better translation is "Is the patient a referral?"

client/src/i18n/en/patient_records.json
```diff
@@ -40,7 +42,39 @@
...
+      "ONLY_HOSPITALIZED":"Only hospitalized patients",
+      "NEW":"New Visit",
+      "ALREADY_ADMITTED":"The patient is already admitted, and his visit is not discharged",
+      "PREGNANT":"Pregnant",
+      "HEALTH_ZONE":"health Zone",
+      "CASE":"Type of case",
+      "REFERED":"Refered",
+      "IS_PREGNANT":"Does she pregnant ?",
+      "IS_IN_HEALTH_ZONE":"In the health zone (ZS)",
+      "IS_NOT_IN_HEALTH_ZONE":"Not in the health zone (HZ)",
```
jniles: A better translation is "Outside of the health zone (HZ)".

client/src/i18n/en/patient_records.json
```diff
@@ -40,7 +42,39 @@
...
-      "VISITS_INFO": "Visits Information"
+      "VISITS_INFO": "Visits Information",
+      "ADMISSION_REGISTRY":"Admissions Registry",
+      "ONLY_HOSPITALIZED":"Only hospitalized patients",
+      "NEW":"New Visit",
+      "ALREADY_ADMITTED":"The patient is already admitted, and his visit is not discharged",
+      "PREGNANT":"Pregnant",
+      "HEALTH_ZONE":"health Zone",
+      "CASE":"Type of case",
+      "REFERED":"Refered",
```
jniles: Referred has two "r"s in English :)

client/src/i18n/en/patient_records.json
```diff
@@ -40,7 +42,39 @@
...
+      "TITLE":"Discharge",
+      "TYPE":"Type of discharge",
+      "REGULAR":"Regular Discharge",
+      "ON_PATIENT_WILL":"The patient having left the hospital on his own will",
+      "EMERGENCY":"Emergency Discharge",
+      "SERVICE_CHANGE":"Service Change",
+      "DEATH":"Death of the patient",
+      "EVASION":"The patient escaped from the hospital",
+      "DISCHARGE_BUT_ON_BED":"Medical discharge but the patient stays in the bed",
+      "STATUQUO_CLINIC":"Statuquo Clinic",
```
jniles, line 76: What is this?

client/src/i18n/en/patient_records.json
```diff
@@ -40,7 +42,39 @@
...
+    },
+    "DISCHARGE":{
+      "TITLE":"Discharge",
+      "TYPE":"Type of discharge",
+      "REGULAR":"Regular Discharge",
+      "ON_PATIENT_WILL":"The patient having left the hospital on his own will",
+      "EMERGENCY":"Emergency Discharge",
+      "SERVICE_CHANGE":"Service Change",
+      "DEATH":"Death of the patient",
+      "EVASION":"The patient escaped from the hospital",
```
jniles: This is _technically_ correct, however it makes it sound like the hospital is a prison 😆 .  A better way to say this is "The patient fled the hospital".

client/src/i18n/en/ward.json
```diff
@@ -14,6 +14,11 @@
     "TITLE":"Bed",
     "LIST":"Bed List",
     "ADD":"Add Bed",
-    "NB_BEDS":"Number of beds"
+    "NB_BEDS":"Number of beds",
+    "AUTO_ASSIGN_AVAILABLE_BED":"Assign the available bed automatically",
+    "MANUAL_ASSIGN_AVAILABLE_BED":"Assign the available bed manually",
+    "NO_BED_AVAILABLE":"No Bed Available",
+    "PATIENT_AUTO_BED":"The patient will be on the bed : {{BED_LABEL}}",
```
jniles: Awesome use of interpolation 👍 

A better way to phrase this is: "The patient will be assigned to {{ BED_LABEL }}".

client/src/js/components/bhBedSelect/bhBedSelect.html
```diff
@@ -0,0 +1,42 @@
+<div ng-form="BedForm" bh-bed-select ng-model-options="{ updateOn: 'default' }">
+  <div
+    class="form-group"
+    ng-class="{ 'has-error' : BedForm.$submitted && BedForm.bed_uuid.$invalid }">
```
jniles: It doesn't look like this should be `bed_uuid` - shouldn't it be `bed_id`?

client/src/js/components/bhBedSelect/bhBedSelect.html
```diff
@@ -0,0 +1,42 @@
...
+      <i class="fa fa-bed"></i> 
+      <span
+        translate-values="{ BED_LABEL : $ctrl.bed.hrLabel }"
+        translate="{{ 'BED.PATIENT_AUTO_BED' }}">
+      </span>
+    </p>
+
+    <p class="text-info" ng-if="$ctrl.noBedAvailable" translate>BED.NO_BED_AVAILABLE</p>
+
+    <div class="help-block" ng-messages="BedForm.bed_uuid.$error" ng-show="BedForm.$submitted">
```
jniles: Same comment: `bed_uuid` -> `bed_id`.

client/src/js/components/bhBedSelect/bhBedSelect.js
```diff
@@ -0,0 +1,79 @@
...
+    controller  : BedSelectController,
+    transclude  : true,
+    bindings    : {
+      uuid      : '<',
+      onSelectCallback : '&',
+      required : '<?',
+      label    : '@?',
+      autoSelect : '<?',
+      roomUuid : '<?',
+      occupied : '<?',
```
jniles: Let's rename this for clarity - `show-occupied-beds={true|false|undefined}`

client/src/js/components/bhBedSelect/bhBedSelect.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  function loadParams() {
+    // load only beds for a given room
+    const allBeds = { room_uuid : $ctrl.roomUuid || 'x' };
+    const occupiedBeds = { room_uuid : $ctrl.roomUuid || 'x', is_occupied : 1 };
+    const notOccupiedBeds = { room_uuid : $ctrl.roomUuid || 'x', is_occupied : 0 };
+
+    // eslint-disable-next-line no-nested-ternary
+    return $ctrl.occupied === true ? occupiedBeds
```
jniles: This function is pretty confusing.  I understand what it is doing, but I don't understand why it is doing it.

It looks like what is happening is this:
```js
const defaultRoomUuid = 'x';  // why is this an x?
const parameters = { room_uuid : $ctrl.roomUuid || defaultRoomUuid };
switch ($ctrl.occupied) {
   case true:
      return angular.extend(parameters, { is_occupied : 1 });
    case false:
      return angular.extend(parameters, { is_occupied : 0 });
    default:
      return parameters;
}
```

Just reading the code, I wonder:
  1. Is a `room_uuid` required for this component?  What does the 'x' mean?
  2. Why can "$ctrl.occupied" be in three states?  Do we really need to support all three states?

client/src/js/components/bhRoomAssign/bhRoomAssign.html
```diff
@@ -0,0 +1,38 @@
...
+  <bh-bed-select
+    id="$ctrl.bed.id"
+    room-uuid="$ctrl.bed.room_uuid"
+    on-select-callback="$ctrl.onSelectBed(bed)"
+    auto-select="!$ctrl.bed.isManualAssign"
+    occupied="false"
+    required="true">
+  </bh-bed-select>
+
+  <div class="checkbox">
```
jniles: I personally don't think a checkbox works well here.  In my view, we should do either one of two things:
 1. Have "This patient will be assigned to bed X" and then a button that says "Change bed assignment".
 2. Have the automatic assignment to optionally added via link.  For example, this is what github does for new repos:
image
  In this case, you can click the text "improve-happiness" is clickable and will automatically fill in the repository name with that text if we choose it.

To be honest, I prefer the first version much more, but figured I should site them both for completeness.

client/src/js/services/grid/GridFilterer.js
```diff
@@ -0,0 +1,189 @@
...
+    {
+      key : 'expiration_date_from', label : 'STOCK.EXPIRATION_DATE', comparitor : '>', valueFilter : 'date',
+    },
+    {
+      key : 'expiration_date_to', label : 'STOCK.EXPIRATION_DATE', comparitor : '<', valueFilter : 'date',
+    },
+  ];
+
+  class GridFilterer {
+    constructor(cacheKey = 'grid-filterer-cache') {
```
jniles: We definitely do not want a default cacheKey here - that will cause multiple grid's filters to collide.  Let's make this required and throw an error if the developer does not specify it.

client/src/js/services/grid/GridFilterer.js
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * @method getQueryString
+     * @description
+     * returns a query string with parameters with the consideration
+     * of the current applied filters
+     * @param {string} type
+     */
+    getQueryString(type) {
```
jniles: Where is "getQueryString" used?  Can it be removed?

client/src/js/services/grid/GridFilterer.js
```diff
@@ -0,0 +1,189 @@
...
+
+GridFiltererService.$inject = [
+  'FilterService', 'appcache', 'PeriodService',
+  '$httpParamSerializer', 'LanguageService', 'bhConstants', 'util',
+];
+
+/**
+ * @class GridFilterer
+ */
+function GridFiltererService(Filters, AppCache, Periods, $httpParamSerializer, Languages, bhConstants, Util) {
```
jniles: I like this idea.  It will may save a lot of duplication of work.

client/src/js/services/grid/GridFilterer.js
```diff
@@ -0,0 +1,189 @@
...
+  const customFiltersList = [
+    { key : 'display_name', label : 'FORM.LABELS.NAME' },
+    { key : 'sex', label : 'FORM.LABELS.GENDER' },
+    { key : 'hospital_no', label : 'FORM.LABELS.HOSPITAL_NO' },
+    { key : 'depot_uuid', label : 'STOCK.DEPOT' },
+    { key : 'inventory_uuid', label : 'STOCK.INVENTORY' },
+    { key : 'group_uuid', label : 'STOCK.INVENTORY_GROUP' },
```
jniles, line 22: We need to figure out what to do with filters like this.  For example, `group_uuid` can be referring to the debtor group uuid, the inventory group uuid, the creditor group uuid, etc...  We can either:
 1. Rename our filters such that we look up `inventory_group_uuid` to match the `group_uuid` column in the database.
 OR
 2. Make the translation key for `group_uuid` just "Group" so that we can reuse it for multiple registries without conflict.

What do you think?  Which should we choose?

client/src/js/services/grid/GridFilterer.js
```diff
@@ -0,0 +1,189 @@
...
+      key : 'custom_date_start', label : 'FORM.LABELS.DATE', comparitor : '>', valueFilter : 'date',
+    },
+    {
+      key : 'custom_date_end', label : 'FORM.LABELS.DATE', comparitor : '<', valueFilter : 'date',
+    },
+    {
+      key : 'dateBirthFrom', label : 'FORM.LABELS.DOB', comparitor : '>', valueFilter : 'date',
+    },
+    {
+      key : 'dateBirthTo', label : 'FORM.LABELS.DOB', comparitor : '<', valueFilter : 'date',
```
jniles, line 64: It looks like we mix camelCase and underscore_case in our filter keys quite a bit.  We should probably figure out which style we want and just adopt it for all of the keys.

client/src/modules/patients/record/bh-patient-visit.component.js
```diff
@@ -36,6 +36,7 @@ function VisitsController(Patients, Notify, Moment) {
     return Patients.Visits.read($ctrl.patientUuid)
       .then(visits => {
         visits.forEach(calculateDays);
```
jniles: Let's combine these two into a single loop:
```js
visits.forEach(visit => {
  calculateDays(visit);
  hrBedLocation(visit);
});

File: client/src/modules/patients/record/bh-patient-visit.component.js [hunk 23]
```diff
@@ -60,6 +61,10 @@ function VisitsController(Patients, Notify, Moment) {
     visit.totalDays = endDate.diff(startDate, 'days');
   }
 
+  function hrBedLocation(visit) {
+    visit.hrBedLocation = (visit.ward_name || '').concat('/', visit.room_label, '/', visit.bed_label);
```
jniles, line 67: Why wouldn't the visit have a ward_name?

client/src/modules/patients/visits/modals/search.modal.html
```diff
@@ -0,0 +1,192 @@
...
+    </ol>
+  </div>
+
+  <div class="modal-body search-modal">
+    <uib-tabset>
+      <uib-tab index="0" heading="{{ 'FORM.LABELS.SEARCH_QUERIES' | translate }}" data-custom-filter-tab>
+        <div class="tab-body">
+          <!-- hospitalized -->
+          <div class="checkbox">
+            <label>
```
jniles: Checkboxes seem to me confusing on search modals.  If this is checked, then hospitalized=1, but if it is unchecked, then hospitalized is either undefined or 0 depending on if it was the user that touched it.

So if I see a screenshot of a form without the check box marked, I have no idea what query it will send to the server.  In my view, this is unacceptably confusing.

Could we use radio buttons to clean this up?  Something like "Hospitalized Patients", with the choices being "Show only hospitalized patients" and "Show only ambulatory patients" and the empty state being interpreted as "show all patients" .

server/controllers/medical/patients/visits.js
```diff
@@ -178,13 +229,175 @@ function admission(req, res, next) {
...
+    .then(availableBed => {
+      glb.newHospitalizationUuid = uuid();
+      const paramInsertHospitalization = {
+        uuid : db.bid(glb.newHospitalizationUuid),
+        patient_visit_uuid : patientVisitUuid,
+        patient_uuid : patientUuid,
+        room_uuid : params.room_uuid,
+        bed_id : availableBed.id,
+      };
+      const transaction = db.transaction();
```
jniles, line 394: Good use of a transaction here 👍

server/controllers/medical/patients/visits.js
```diff
@@ -206,6 +419,11 @@ function discharge(req, res, next) {
   const visitUuid = data.uuid;
   delete data.uuid;
 
+  if (!data.discharge_type_id) {
```
jniles, line 430: I think we should probably make discharge_type_id a required field and not have any ambiguity in what kind of discharge it is.  It seems like a really important statistic to get right!

client/src/modules/patients/visits/modals/search.modal.html
```diff
@@ -0,0 +1,192 @@
...
+              <span translate>PATIENT_RECORDS.VISITS.IS_REFERED</span>
+            </label>
+            <bh-clear on-clear="$ctrl.clear('is_refered')"></bh-clear>
+          </div>
+
+          <!-- pregnant -->
+          <div class="checkbox">
+            <label>
+              <input type="checkbox" ng-true-value="1" ng-false-value=0 ng-model="$ctrl.searchQueries.is_pregnant">
+              <span translate>PATIENT_RECORDS.VISITS.IS_PREGNANT</span>
```
jniles: I'm curious why we have a special field for pregnant patients.  It feels to me like that information belongs in another field - perhaps in diagnosis?

Because we could easily add fields for `is_hiv_positive`, `is_youth`, etc, but that seems like it shouldn't be infinite numbers of fields in that database...  Do you know why we are singling out pregnancy as a particular field, even though ~40% of our patients will never be pregnant?

client/src/modules/patients/visits/modals/search.modal.html
```diff
@@ -0,0 +1,192 @@
...
+            <bh-clear on-clear="$ctrl.clear('hospital_no')"></bh-clear>
+
+            <input type="text" class="form-control" name="hospital_no" ng-model="$ctrl.searchQueries.hospital_no">
+            <div class="help-block" ng-messages="ModalForm.reference.$error" ng-show="ModalForm.$submitted">
+              <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+            </div>
+          </div>
+
+          <!-- service -->
+          <bh-service-select on-select-callback="$ctrl.onSelectService(service)">
```
jniles, line 152: Should this have a `service-id` attribute?

client/src/modules/patients/visits/modals/visits.modal.html
```diff
@@ -0,0 +1,209 @@
...
+    <bh-room-assign on-select-callback="AdmitCtrl.onBedRoomSelect(bed)"></bh-room-assign>
+  </div>
+</div>
+
+<div class="modal-footer">
+  <button type="button" class="btn btn-default" ng-click="AdmitCtrl.cancel()" data-method="cancel">
+    <span translate>FORM.BUTTONS.CLOSE</span>
+  </button>
+
+  <bh-loading-button loading-state="AdmitCtrl.$loading">
```
jniles: Out of curiosity, why did you choose to manage `$loading` yourself instead of using `bh-submit`?

client/src/modules/patients/visits/modals/transfer.modal.js
```diff
@@ -0,0 +1,34 @@
...
+  vm.transfer = () => {
+    if (!vm.bed) { return; }
+
+    Visits.transfer(Params.patient_uuid, Params.patient_visit_uuid, vm.bed)
+      .then(transfered => {
+        if (!transfered) { return; }
+
+        Notify.success('PATIENT_RECORDS.TRANSFER.SUCCESSFULLY_TRANSFERED');
+        Modal.close(true);
+      });
```
jniles, line 32: Should we do something if an error occurs?

Also, is there a loading indicator for this operation?  It looks like it might cause the user to wait for a bit.

server/controllers/medical/patients/visits.js
```diff
@@ -27,44 +29,91 @@ exports.detail = detail;
...
+  patient_visit.hospitalized,
+  z.bed_label, z.room_label, z.ward_name,
+  patient.display_name, patient.hospital_no, em.text AS reference,
+  s.name AS service_name,
+  dt.label AS discharge_label,
+  inside_health_zone, is_new_case, is_refered, is_pregnant, last_service_id, discharge_type_id
+`;
+
+const TABLES = `
+  patient_visit
```
jniles, line 66: What is the goal of this SQL query?  I can't seem to figure out what it does.

Let's talk this through tomorrow.

server/controllers/medical/patients/visits.js
```diff
@@ -178,13 +229,175 @@ function admission(req, res, next) {
...
+  transaction.addQuery(visitService.query, [visitService.parameters]);
+  return transaction.execute();
+}
+
+function createHospitalization(data) {
+  const visitService = getVisitServiceQuery(data);
+  const { bed, service } = data;
+  const visit = _.omit(data, ['bed', 'service']);
+
+  return Q.fcall(() => {
```
jniles: Why do we need to use `Q.fcall`?

server/controllers/medical/patients/visits.js
```diff
@@ -178,13 +229,175 @@ function admission(req, res, next) {
...
     INSERT INTO patient_visit SET ?;
   `;
+  const transaction = db.transaction();
+  visit.last_service_id = service.id;
+  transaction.addQuery(sqlInsertVisit, [visit]);
+  transaction.addQuery(visitService.query, [visitService.parameters]);
+  return transaction.execute();
+}
+
+function createHospitalization(data) {
```
jniles: This function is complicated enough - let's use `async`/`await` here.

server/controllers/medical/patients/visits.js
```diff
@@ -178,13 +229,175 @@ function admission(req, res, next) {
...
+  const visitService = getVisitServiceQuery(data);
+  const { bed, service } = data;
+  const visit = _.omit(data, ['bed', 'service']);
+
+  return Q.fcall(() => {
+    return lookupAutoAvailableBed(bed);
+  })
+    .then(b => {
```
jniles: Can we rename this variable "nextBed" instead of "b"?

server/controllers/medical/patients/visits.js
```diff
@@ -178,13 +229,175 @@ function admission(req, res, next) {
...
+    patient_visit_uuid : visit.uuid,
+  };
+  return {
+    query : sqlInsertVisitService,
+    parameters : paramInsertService,
+  };
+}
 
-  db.exec(sql, [data])
+function lookupAutoAvailableBed(bedOptions) {
```
jniles: Can we name this "lookupNextAvailableBed"?  We don't really need to know if it was automatic or not in the function name.

server/controllers/medical/patients/visits.js
```diff
@@ -178,13 +229,175 @@ function admission(req, res, next) {
...
+  const sql = `
+    SELECT b.id FROM bed b 
+    JOIN room r ON r.uuid = b.room_uuid
+    WHERE r.uuid = ? AND b.is_occupied = 0
+    ORDER BY r.label, b.label
+    LIMIT 1;
+  `;
+  return bedOptions.id ? { id : bedOptions.id } : db.one(sql, [db.bid(bedOptions.room_uuid)]);
```
jniles: Hmm...  This function will either return a promise or an object.  It seems like this is a [code smell](https://en.wikipedia.org/wiki/Code_smell).  If you ever have this scenario in the future, just wrap the object in `Promise.resolve()`, like this:
```js
return bedOptions.id ? Promise.resolve({id : bedOptions.id}) : db.one(sql, [ ... ]);
```

However, it seem like if we have `bedOptions.id`, this function doesn't do anything.  But if that is the case, why did we call it in the first place?  We should have done something like this:
```js
if (options.id) {
  bed = { id : options.id };
}  else {
  bed = await lookupNextAvailableBed(options).
}
```

jniles (commented):
@mbayopanda this represents a lot of work, and for the most part, it is rock solid.  I've made comments where I think the code can be improved.  There are three major areas that we should discuss in person though: 
 1. The SQL is way too complicated, particularly where we have the `const TABLES` declaration.  It is a query in a query in a join a query, and I can't even read it - I'm sure the other developers will have a horrible time with it as well.  Let's try to simplify it together.
 2. It seems like we are writing to the patient_hospitalization table multiple times if a patient is transferred.  I'm not entirely sure how the transfer logic works - let's discuss it together tomorrow.
  3. I'm not entirely sure what all the different discharge types and their logic might be.  For example, what is an emergency discharge?  How is it different from a regular discharge?  Can you have a regular discharge where the patient was not healed?  Etc..

Considering this PR has over 2.5K new lines of code, I'm sure I've missed something.  I'll take a second look after the first round of changes have been made.

mbayopanda:
@jniles

client/src/js/components/bhBedSelect/bhBedSelect.html
jniles: This still needs to be addressed.  It looks like a typo.

client/src/js/components/bhDischargeSelect/bhDischargeSelect.js
jniles: It looks like you need to complete the second part of this - I feel that it might replace the old value with the first one in the dechargeTypes list without it.

client/src/js/components/bhRoomAssign/bhRoomAssign.js
```diff
@@ -0,0 +1,45 @@
...
+      label    : '@?',
+    },
+  });
+
+RoomAssignController.$inject = [];
+
+/**
+ * RoomAssign selection component
+ */
+function RoomAssignController() {
```
jniles, line 18: This is a really clean controller 👍

client/src/js/services/grid/GridFilterer.js
jniles: Can this be removed?

jniles (commented):
@mbayopanda, could you go back and look at some of the questions I had in the first review?  They still stand.

mbayopanda:
@jniles

client/src/i18n/en/patient_records.json
mbayopanda: The status of the patient doesn't change

client/src/js/components/bhBedSelect/bhBedSelect.js
mbayopanda: 1. The room_uuid is required, if it missing we must not have beds listed. I set 'x' as default because if it null or undefined the filter on the server side returns all beds.

2. "$ctrl.occupied" is in three states because it works like a filter, if true it will show occupied beds, if false it will show just free beds (not occupied), if undefined it will show all beds

client/src/js/components/bhRoomAssign/bhRoomAssign.html
mbayopanda: image

client/src/js/services/grid/GridFilterer.js
mbayopanda: Look at that : 
```js
class GridFilterer {
    constructor(cacheKey) {
      if (!cacheKey) {
        Notify.danger('FORM.LABELS.GRID_CACHE_KEY_MISSING');
        return;
      }
```

client/src/js/services/grid/GridFilterer.js
mbayopanda: The first option seems good to me

client/src/js/services/grid/GridFilterer.js
mbayopanda: I do not change some filter keys because they are used elsewhere

client/src/modules/patients/record/bh-patient-visit.component.js
mbayopanda: In the case the visit is for hospitalized patient, we must be able to see the assigned location (ward, room, bed) of the patient

server/controllers/medical/patients/visits.js
mbayopanda: You're right, but it is recorded in the visit (in the visit table), during the start of the visit we cannot set the discharge type.

client/src/modules/patients/visits/modals/search.modal.html
mbayopanda: No

mbayopanda:
@jniles

lomamech:
@mbayopanda  I spent my time analyzing the features of the different modules, I suggested a small modification regarding the translation "Room" should be translated "Salle" and not by "Chambre"

mbayopanda:
@lomamech

PR #3624, opened 2019-03-03
feat(reports): filter locked clients

### LINKED ISSUE #3386: Allow users to filter out locked clients from the clients report

jniles:
This commit adds a filter for locked clients to the Annual Clients
Report, allowing administrators to prevent the display of debtor groups
in the report.

A bug was also discovered with the currency select and is now fixed.

Closes #3386.

filterlockedclients


jeremielodi (commented):
LGTM

PR #3631, opened 2019-03-07
fix(barcode): restore barcode functionality

### LINKED ISSUE #3629: jsBarcode not imported correctly
The barcode doesn't work because it is not correctly imported : 

image

I got the error below with the visit report : 
image

jniles:
The barcode functionality was destroyed in a merge somewhere, and so this is a critical fix.

Closes #3629.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -96,6 +96,9 @@ const paths = {
       'node_modules/moment/min/moment.min.js',
       'node_modules/moment/locale/fr.js',
       'node_modules/angular-moment/angular-moment.min.js',
+
+      // barcode
+      'node_modules/jsbarcode/dist/JsBarcode.all.min.js',
     ],
 
     // these must be globs ("**" syntax) to retain their folder structures
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "helmet": "^3.15.0",
     "ioredis": "^4.5.1",
     "jquery": "^3.3.1",
-    "jsbarcode": "^4.0.0-alpha.5",
+    "jsbarcode": "^3.11.0",
     "json-2-csv": "^3.2.0",
     "juice": "^5.1.0",
     "lodash": "^4.17.11",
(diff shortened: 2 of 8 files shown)


jniles:
@mbayopanda, can this get a quick review?

mbayopanda (commented):
I got this error : 

image

I think it is relative to Windows

mbayopanda (changes requested):
The server crashes after generating a pdf

jniles:
Interesting.  Thanks for trying it out! :+1:

jniles:
Hmm... looks like we won't be able to get this to work without:
 1. Prefixing a `file:///` url (which will break the client-side version of this).
 2. Making a special route to server the barcode.min.js file
 3. Upgrade wkhtmltopdf
 4. Adding in the entire vendor.js file into the PDF.

Since none of these options are really great on the long run, I'm going to pick backup our experiments with puppeteer (https://github.com/IMA-WorldHealth/bhima/pull/3388) and see if I can get that up to date.  It's definitely a longer term solution,

kwilu:
Heads up @jniles, we just merged some commits that conflict with the changes your made in this pull request! You can review this repository's [recent commits](https://github.com/IMA-WorldHealth/bhima/commits/master) to see where the conflicts occur. Please rebase your feature branch against the `upstream/master` branch and [resolve](https://zulip.readthedocs.io/en/latest/git/troubleshooting.html#recover-from-a-git-rebase-failure) your pull request's merge conflicts accordingly.

mbayopanda:
Hi @jniles what do you about this PR are you working on that, or can i assign the task to another developper since you are in your vacation

jniles:
@mbayopanda, sure.  It should be noted that this fixes the problem in everything but Windows.  Unfortunately, I don't have a windows install to test it on though.

mbayopanda:
@jniles i fix the problem with windows but i cannot add a commit in your PR, so i created a new one based on yours.

jniles:
@mbayopanda, no problem.  I'll close it and continue on that one.

I _think_ to you should be able to commit in my PR.  Check that you have my remote repository on your github registered with the `git@github.com:` prefix and not the `https://` prefix.   That will be the reason you cannot commit if it is the case.

PR #3634, opened 2019-03-11
fix: update bhima.sql with appropriate report keys

jniles:
It appears bhima.sql got out of date with our migrate.sql file and the reports keys were not being translated correctly.  This commit corrects that issue.

**Previous Values**
report_before_fix

**New Fixed Values**
report_fix_after


mbayopanda (approved):
LGTM 👍 

bors r+

Issue #3633, opened 2019-03-10
Tracking Hospital Management Indicators


mbayopanda:
This issue is about to add a way of inserting each month some basic indicator variables so that we can produce hospital management indicators which can be viewed from a dashboard.

1. Indicator Variables Registry : this registry page allows to add each months indicator variables which will help to produce indicators values.

![Registre des indicateurs](https://user-images.githubusercontent.com/5445251/54082377-1956bd80-4315-11e9-9403-a4fd380ddee0.png)

There are three types of indicators, so we must have also three types of indicator variables for : 
- Hospitalization
- Personnel (Human Resources)
- Finance

Each months users must fill a monthly indicator variables, if everything is ok then status of this indicator variables must be completed. If administrators validate it then the status can be changed to validated.

NOTE: We must have a unique constraint for period and type

1.1. Hospitalization file (Monthly Indicators Variables)

![Indicateurs des hospi](https://user-images.githubusercontent.com/5445251/54082076-5f5d5280-4310-11e9-871b-35f2218c824a.png)

1.2. Personnel file

![Indicateur des Personels](https://user-images.githubusercontent.com/5445251/54082091-a2b7c100-4310-11e9-8ed1-07b5d4f042d9.png)

1.3. Finance file

![Indicateurs financier](https://user-images.githubusercontent.com/5445251/54082094-b531fa80-4310-11e9-966c-10de5c9e665b.png)

If all files (Hospitalization, Personnel and Finance) are filled then we can see hospital management indicators produced through a dashboard page.

2. Indicators Dashboards

![Dashboard](https://user-images.githubusercontent.com/5445251/54082122-15c13780-4311-11e9-88c9-78ff7497ba00.png)

A click on an indicator must show more details in a modal

![Indicator details](https://user-images.githubusercontent.com/5445251/54082145-51f49800-4311-11e9-8d02-97d8e6704e5e.png)

2.1. Hospitalization indicators

![Hospi Indicators](https://user-images.githubusercontent.com/5445251/54082165-96803380-4311-11e9-9fbe-6447284c156a.png)

2.2. Personnel Indicators

![Personel](https://user-images.githubusercontent.com/5445251/54082230-5a999e00-4312-11e9-95bf-b3717fc8c1fe.png)

2.3. Finance Indicators

![Finance](https://user-images.githubusercontent.com/5445251/54082178-c4657800-4311-11e9-84e9-3897332f45b0.png)

[...]

PR #3637, opened 2019-03-11
feat(indicators) hospitalization

jeremielodi:
#3633


client/src/modules/indicators/indicator.ctrl.js
```diff
@@ -0,0 +1,46 @@
+angular.module('bhima.controllers')
```
mbayopanda: Could you remove all these client code and keep only services which provide communication to the API at the server

client/src/modules/indicators/indicator.service.js
```diff
@@ -0,0 +1,18 @@
...
+ * A service wrapper for the /indicators HTTP endpoint.
+ */
+function IndicatorService(Api) {
+  const service = this;
+  service.hospitalization = new Api('/indicators/hospitalization');
+  service.personel = new Api('/indicators/personel');
+  service.fincances = new Api('/indicators/fincances');
+  service.status = new Api('/indicators/status');
+  return service;
+}
```
mbayopanda, line 21: 👍

mbayopanda (commented):
👍 Excellent

Could you remove just some controllers and views at the client side so that the code will be merge

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3649, opened 2019-03-23
chore: refactor gulpfile into individual files

jniles:
We take advantage of gulp's ability to break build scripts into multiple
files to ease the maintenance of the build scripts.  Some positive
benefits of this include:
 1. No longer compiling/minifying for development builds
 2. Karma is able to read from static file instead of parsing a JSON to
 figure out where to look.
 3. Karma is in headless mode.
 4. Only transfer server files that we need, instead of everything.
 5. We have unminified dependencies when we are trying to debug development builds.   Only in production do we have minified builds.

I'm not sure if this increases the speed of the build at all.

TODO:
  - [x] Implement the `watch()` statements so that we can rebuild the client on changes.


jniles:
@mbayopanda this is ready for a review.  You can probably just download it, and try to develop for a day using the `watch()` functionality and see how it works out.

Building the app in development mode from scratch takes 14s on my computer, which is way too long, but it is a rather large application.  I suppose something like webpack would help solve that.

Building for production takes 24 seconds.  Hopefully this means we can optimize both production and development at the same time.  Clearly TypeScript is our largest time consumer.

Let me know if you run into any issues.

mbayopanda (approved):
LGTM 👍 

It seems like the watch feature works now.

bors r+

PR #3653, opened 2019-03-26
fix(reports): handle unpaid balance errors

### LINKED ISSUE #3575: Unpaid invoice report is broken
The issue is about the unpaid invoice report which doesn't works properly.

When we define entries, we get an error relate to collate : 
image

jniles:
The Unpaid Invoices errors would throw SQL errors if there was no data
for the range (since that is what the Pivot() function does).  The
report has been updated to catch and gracefully handle certain data
errors so that it simply informs the user that there is no data for
that range.

Closes #3575.

**Demo**


mbayopanda (approved):
LGTM 👍 

bors r+

PR #3657, opened 2019-03-27
feat(journal): allow more modifications to posted transactions

jniles:
We have implemented the ability to modify the transaction date of
transactions to allow users to change the date if it was a date error
that was made on a posted record.  Additionally, the transaction type
can also be modified.  Note that both of these options were available
for unposted records - these changes simply extend those options to
  posted records.


mbayopanda (commented):
It seems like changes on date and transaction type doesn't take effect : 

edit posted transaction

jniles:
I think that is the way the Posting Journal works in general.  I think you need to refresh the page for the changes to take effect.  I'll test it out to see if that is the case, but if so, I'll add a note to the documentation.

We probably should change this behavior, but it would be better to add it to another pull request instead of trying to shove it all in this one.

jniles:
@mbayopanda, I can't reproduce the condition you indicated.  See if you can reproduce it on your local instance.

For the record, it seems like all edits are automatically applied - both these ones and the normal ones to description/account, etc.

jniles:
Okay ... this needs work.  Clearly none of the server is included in this PR, so I'll need to go back and figure out what happened to it.

jniles:
bors try

jniles:
bors try

jniles:
bors try

jniles:
bors try

jniles:
Tests finally pass!  @mbayopanda, can you see if the changes you remarked work now?

mbayopanda (approved):
Works fine now, the description text changes to an auto text description.

LGTM :+1

bors r+

mbayopanda:
bors try

PR #3665, opened 2019-04-03
feat(report) indicators report

jeremielodi:


client/src/i18n/en/report.json
```diff
@@ -130,6 +130,10 @@
       "TITLE": "Multiple Accounts Report",
       "DESCRIPTION": "This report shows all transactions for a group of accounts given a set time period."
     },
+    "REPORT_INDICATORS": {
+      "TITLE": "Indicators Report",
+      "DESCRIPTION": "This report sum up  indications in a hospital"
```
mbayopanda: What do you think about this translation : `This report summarizes the indicators for a hospital`

Indicators is more meaningful than indications in this context.

server/lib/template/helpers/logic.js
```diff
@@ -23,6 +22,11 @@ function lt(a, b, options) {
   return options.inverse(this);
 }
 
+function setdefault(val, _default) {
```
mbayopanda: Could you rename this function to something like `ignoreNan` or `hideNan`, and also instead of adding ' ' in the handlebars you could just add it from this function like : 
```js
function ignoreNan(val) {
  return (val >= 0 || val < 0) ? val : '';
}
```

mbayopanda (commented):
That's great, i made a comment about the setDefault helper.

mbayopanda:
LGTM 👍 
bors r+

PR #3669, opened 2019-04-09
fix profit and loss references details

mbayopanda:
This PR fix calculation in the profit and loss report


lomamech:
LGTM 👍

bors r+

PR #3671, opened 2019-04-10
Fix(barcode): restore barcode functionnality

### LINKED ISSUE #3629: jsBarcode not imported correctly

mbayopanda:
Closes #3629.


mbayopanda:
@jniles

jniles:
Okay, I don't anticipate this having any test issues since we are out of the integration/unit test phases.

@mbayopanda great work.  I've tried to improve it by writing a test that will fail (in theory) if the barcode library breaks in the future by breaking the PDF rendering test.  Ideally, this should be done in an HTML rendering test, but we don't have those tests yet.  Instead of strapping an HTML test suite onto this PR, I've just implemented a single unit test for the PDF rendering with a barcode.

I discovered that the handlebars wasn't actually turned into HTML since we had a "dumb" rendering function (`noop()`).  I swapped it for a "dumb" call to `handlebars`.  This slows the tests down (since we actually have to use handlebars now), but it makes the tests a lot more reflective of reality.  Also, the first test didn't actually load the file, so I made it load the file.

Your modifications look good to me.  I've tested it and it works on my machine.  Well done :)

bors r+

PR #3678, opened 2019-04-15
improvement(Balance Account and Ohada Balance sheet)

### LINKED ISSUE #3677: Balance Sheet et Balance Accounts
There is a change in the display of the Ohada Balance Sheet just because in the new Ohada revised social debts and tax debts are now in the same reference

Account Balance Report Shows Hidden and Blocked Counts

lomamech:
- Remove reference for Balance DL and merge DL reference in DK
- Prevent the display of blocked and blocked accounts in the
  balance of accounts report

closes #3677


lomamech:
@mbayopanda

server/controllers/finance/reports/break_even/break_even_project.js
```diff
@@ -0,0 +1,284 @@
+function configuration(data) {
```
mbayopanda: Why do we have break-even code here ?

server/controllers/finance/reports/balance/index.js
```diff
@@ -154,6 +154,7 @@ async function getBalanceForFiscalYear(period, currencyId) {
       WHERE pt.fiscal_year_id = ? AND p.fiscal_year_id = ?
       GROUP BY pt.account_id
     )s ON a.id = s.account_id
+    WHERE a.locked <> 1 AND a.hidden <> 1
```
mbayopanda: AND or OR ?

server/controllers/finance/reports/balance/index.js
```diff
@@ -154,6 +154,7 @@ async function getBalanceForFiscalYear(period, currencyId) {
       WHERE pt.fiscal_year_id = ? AND p.fiscal_year_id = ?
       GROUP BY pt.account_id
     )s ON a.id = s.account_id
+    WHERE a.locked <> 1 OR a.hidden <> 1
```
mbayopanda, line 157: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

server/controllers/finance/reports/balance/index.js
jniles: Do we have rules that prevent someone from locking or hiding an account with a balance?  If not, we need to implement them with this change.

The risk is that someone will hide an account with a balance and we'll never know why their balance sheet was incorrectly balanced.

Issue #3689, opened 2019-04-23
Harmonize administration modules according 2.X style


mbayopanda:
This issue is about to harmonize some modules of Administration folder in BHIMA according 2.x style : Grid and Modals

- [x] Supplier Management
- [x] Subsidies Management
- [ ] Patient Group Management
- [x] Localisation

kwilu:
Hello @jeremielodi, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@kwilu claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

jniles:
I can confirm that this is done thanks to the hard work of @jeremielodi .

PR #3691, opened 2019-04-24
use of grid and modal in subsidies management module

mbayopanda:
This PR is about to use 2.X grid and modal in the subsidies management module.

see #3689

### DIFF
--- a/client/src/modules/subsidies/subsidies.routes.js
+++ b/client/src/modules/subsidies/subsidies.routes.js
@@ -1,9 +1,42 @@
 angular.module('bhima.routes')
-  .config(['$stateProvider', function ($stateProvider) {
+  .config(['$stateProvider', $stateProvider => {
     $stateProvider
       .state('subsidies', {
         url         : '/subsidies',
         controller  : 'SubsidyController as SubsidyCtrl',
         templateUrl : 'modules/subsidies/subsidies.html',
+      })
+
+      .state('subsidies.create', {
+        url : '/create',
+        params : {
+          subsidy : { value : {} },
+          creating : { value : true },
+        },
+        onEnter : ['$uibModal', subsidyModal],
+        onExit : ['$uibModalStack', closeModal],
+      })
+
+      .state('subsidies.edit', {
+        url : '/edit',
+        params : {
+          subsidy : { value : {} },
+          creating : { value : false },
+        },
+        onEnter : ['$uibModal', subsidyModal],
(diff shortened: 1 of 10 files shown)


client/src/modules/subsidies/subsidies.routes.js
```diff
@@ -1,9 +1,42 @@
...
+        params : {
+          subsidy : { value : {} },
+          creating : { value : true },
+        },
+        onEnter : ['$uibModal', subsidyModal],
+        onExit : ['$uibModalStack', closeModal],
+      })
+
+      .state('subsidies.edit', {
+        url : '/edit',
```
jniles: I think we typically include the id of the object we are editing in the URL, and then download it in the controller.  This ensures that we can link to it from someplace else in the application.

What do you think of changing this URL to being `/subsidies/:id/edit`, removing the subsidy parameter, and then downloading the subsidy in the modal controller?

client/src/modules/subsidies/templates/action.tmpl.html
```diff
@@ -0,0 +1,23 @@
...
+      <li>
+        <a data-method="edit-record" ng-click="grid.appScope.update(row.entity)" href>
+          <i class="fa fa-edit"></i> <span translate>FORM.BUTTONS.EDIT</span>
+        </a>
+      </li>
+      <li class="divider"></li>
+      <li>
+        <a data-method="delete-record" ng-click="grid.appScope.remove(row.entity)" href>
+          <span class="text-danger">
+            <i class="fa fa-trash"></i> <span translate>DEPOT.DELETE</span>
```
jniles: Hmm.. is there a `FORM.BUTTONS.DELETE`?  Depot looks out of place!

jniles (commented):
Just a few comments about design decision.  Code is clean!

client/src/modules/subsidies/subsidies.routes.js
```diff
@@ -1,9 +1,41 @@
...
+        url : '/create',
+        params : {
+          creating : { value : true },
+        },
+        onEnter : ['$uibModal', subsidyModal],
+        onExit : ['$uibModalStack', closeModal],
+      })
+
+      .state('subsidies.edit', {
+        url : '/edit/:id',
```
jniles: This is a small nitpick, but could you make this `:id/edit` or does that not work?

jniles (approved):
LGTM!

bors r+

PR #3716, opened 2019-05-09
Chore rewrite tests with async await

### LINKED ISSUE #2559: Turn off control flow for Protractor Tests
Our protractor tests can no longer be debugged NodeJS removed the debugger in version 8.  In order to make our tests work with a debugger, we'll need to use [disable control flow](https://github.com/angular/protractor/blob/master/docs/debugging.md#disabled-control-flow) in our tests.  This will let us use a debugger call to launch a chrome debugger.

To disable control flow, we need to start using `async`/`await` and then turn off protractor's promise manager.   Unfortunately, we cannot mix `async`/`await` with the old control flow structures (see [this document](https://github.com/angular/protractor/blob/master/docs/async-await.md)).  This means that we'll need to rewrite all our tests using async/await.  Thankfully, it shouldn't be hard, but it will take a day of upgrading!

### LINKED ISSUE #3655: Run Protractor tests with fixed screen size
There are a number of issues with our protractor tests, but most cumbersome is the lack of a fixed screen size.  This affects the visibility of elements on the page, since on some screens they would be visible, yet on other sizes, they might be hidden (either because of our responsive design or because the screen is too small).

The solution is to set a `window-size` in our Chrome instance that is running the test.  Luckily, with protractor, we can just pass this to `chromeOpts` when instantiating the browser.  I'm sure a few of our tests will need to be rewritten to make sure that they account for the chosen size.

My initial though is 1280x1024 (what we set xvfb to on Travis), but we should play around with it to see what values work best.

### LINKED ISSUE #3710: Rewrite E2E test with async/await instead of SELENIUM_PROMISE_MANAGER
Well, we've finally been forced to that point - [protractor has updated to drop support for the promise manager](https://github.com/angular/protractor/blob/master/CHANGELOG.md).  We've discussed this internally a bit, but never pulled the trigger.  We need to remove the promise manager and begin writing all our tests with async/await.

Basically, we need to go through all our old E2E tests and use `await` everywhere that a promise is executed.  This process is documented in the [protractor documentation](https://www.protractortest.org/#/async-await).  For a discussion of the previous control flow, [see this page](https://www.protractortest.org/#/control-flow).

It's going to be a long process, but hopefully not too painful.  On the plus side, it will remove some of the magic from our tests and hopefully speed them up a bit.

jniles:
This PR re-touches all our end to end tests, migrating them from the previous promise syntax (supported by protractor <6) to the newer async/await syntax (supported by protractor 6+).  This amounts to turning the selenium promise manager off, and managing ourselves the promises generated by each line of code with async/await.

Closes #3710.  Closes #3655. Closes #2559.


kwilu:
Hello @jniles, it seems like you have referenced #3710 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #3710.`.

```
Author: kwilu

    Fixes #51.
```

jniles:
@jeremielodi, can you pull this, update your `node_modules` and try to run the end to end tests to see if they work?  They work on my machine, and on Travis, but I want to get a third opinion.

jeremielodi:
ok let try

jeremielodi:
I think these Conflicting files must be fixed before running the test

jniles:
The tests should work fine without the conflicts.  The conflicts will just determine if we could merge these changes into the others without manual rebasing.

jniles:
Alright.  I think this massive overhaul is ready.  @jeremielodi confirmed it works on his machine, it works on mine, it works on Travis.  I think it's ready for prime time.

bors r+

PR #3720, opened 2019-05-09
fix(Cash Report)

### LINKED ISSUE #3719: Problem with the cash report
There is a difference when consulting the report of the accounts with the reports of the cashbox, this issue proposes to correct this problem

lomamech:
- Use enterprise currency and Use only enterprise currency in Cash report 
- Add column Value for display the debit or credit convert in enterprise   currency 
- Update Report Separated display the difference between incomes and   expense and the balances
closes #3719


lomamech:
@jniles it's fixed

jniles:
Hi @lomamech,

Thanks for the signal!  I'll take a look.

You asked me about adding labels to this PR - you can use @kwilu to do it, following [this documentation](https://zulip.readthedocs.io/en/latest/contributing/zulipbot-usage.html).

Like this:

@kwilu add "bug fix"

kwilu:
**ERROR:** Label "bug fix" does not exist and was thus not added to this pull request.

jniles:
... looks like it is case-sensitive:

@kwilu add "Bug Fix"

server/controllers/finance/reports/cashReport/index.js
```diff
@@ -101,8 +101,12 @@ function document(req, res, next) {
       _.merge(context, { cashbox });
 
       // determine the currency rendering
-      params.currency_id = cashbox.currency_id;
-      params.isEnterpriseCurrency = params.currency_id === req.session.enterprise.currency_id;
+      // Update By @lomamech
```
jniles, line 104: 👍

jniles (approved):
Looks good to me.

bors r+

PR #3726, opened 2019-05-14
Add currency in unpaid invoice report

mbayopanda:
This PR add the currency in the unpaid invoices report.

unpaind invoice


mbayopanda:
The export of this report into an `xls` file for `Fc` currency doesn't allow to perform calculation correctly, since this report is in enterprise currency we decide to let it as it was (just values without currency formatting).

PR #3733, opened 2019-05-18
fix: account reference dropdown header links

jniles:
Fixes the link rendering on the dropdown headers for the account references and the account reference type.  Previously, it was displayed like a link, but didn't have an action when clicked.  Now, it looks like a title, the way that bootstrap wishes.

**Demo**

_Fig 1: Showing the improvements to the dropdown header_

### DIFF
--- a/client/src/modules/account_reference/templates/action.cell.html
+++ b/client/src/modules/account_reference/templates/action.cell.html
@@ -6,9 +6,7 @@
   </a>
 
   <ul data-row-menu="{{row.entity.abbr}}" class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
-    <li class="dropdown-header text-ellipsis"  style="max-width:250px;">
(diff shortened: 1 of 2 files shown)


client/src/modules/account_reference/templates/action.cell.html
```diff
@@ -6,9 +6,7 @@
   </a>
 
   <ul data-row-menu="{{row.entity.abbr}}" class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
-    <li class="dropdown-header text-ellipsis"  style="max-width:250px;">
-      <a>{{row.entity.abbr}}</a>
-    </li>
+    <li class="dropdown-header text-ellipsis" style="max-width:250px;">{{row.entity.abbr}}</li>
```
mbayopanda, line 9: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3734, opened 2019-05-18
fix: dismiss fiscal loading indicator once loaded

jniles:
The loading indicator on the fiscal year closing balances page would load indefinitely due to a swapped boolean.  The variables have been changed so that the loading indicator is reset once the grid is populated.


client/src/modules/fiscal/fiscal.closingBalance.js
```diff
@@ -193,7 +193,7 @@ function FiscalClosingBalanceController(
         Notify.handleError(err);
       })
       .finally(() => {
-        vm.loading = true;
+        vm.loading = false;
```
mbayopanda, line 196: 👍 good catch

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3735, opened 2019-05-18
fix: display correct currency in account footer

jniles:
Fixes a display issue with the currency in the footer of the account statement.  Previously it rendered an unknown currency, now it renders the enterprise currency.


mbayopanda (approved):
LGTM 👍 

bors r+

PR #3736, opened 2019-05-18
fix: parse <head> contents from XLS renderer

jniles:
This removes are links from the <head> so they don't give us errors when we open the report in XLS.

### DIFF
--- a/server/lib/renderers/xls.js
+++ b/server/lib/renderers/xls.js
@@ -16,8 +16,16 @@ exports.render = render;
 exports.extension = '.xls';
 exports.headers = headers;
 
-
 async function render(data, template, options) {
   const htmlString = juice(await html.render(data, template, options));
-  return Buffer.from(htmlString, 'utf-8');
+
+  // NOTE(@jniles)
+  // This code replaces the <head></head> contents to avoid CSS <link> errors when
+  // opening in MS Excel.
+  const start = htmlString.substring(0, htmlString.search('<head>'));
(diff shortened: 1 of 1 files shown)


server/lib/renderers/xls.js
```diff
@@ -16,8 +16,16 @@ exports.render = render;
...
-
 async function render(data, template, options) {
   const htmlString = juice(await html.render(data, template, options));
-  return Buffer.from(htmlString, 'utf-8');
+
+  // NOTE(@jniles)
+  // This code replaces the <head></head> contents to avoid CSS <link> errors when
+  // opening in MS Excel.
+  const start = htmlString.substring(0, htmlString.search('<head>'));
+  const end = htmlString.substring(htmlString.search('</head>') + 7);
```
mbayopanda, line 26: Very good, i would like to know what means the `7`, is it possible to use a variable for that.

mbayopanda:
@jeremielodi what is the state of the review of this PR ?

server/lib/renderers/xls.js
jniles: Aha.  7 is the length of the string `'</head>'`.  Since without the 7, it would leave the word `</head>` in the template.

jniles: Do you have a suggestion for how to do this cleaner?

mbayopanda (approved):
LGTM 👍 

bors r+

jniles:
Clearly bors likes me better that @mbayopanda :D :D

PR #3749, opened 2019-05-27
feat(payroll) paying by indices

jeremielodi:
This PR helps to create indices for grade and function, those indices will be used later on employee payment calculations.


jniles:
@jeremielodi, what is the status of this PR?

lomamech (commented):
Hello @jeremielodi try to update this PR and I'll start with the Review

jniles:
bors try

kwilu:
Heads up @jeremielodi, we just merged some commits that conflict with the changes your made in this pull request! You can review this repository's [recent commits](https://github.com/IMA-WorldHealth/bhima/commits/master) to see where the conflicts occur. Please rebase your feature branch against the `upstream/master` branch and [resolve](https://zulip.readthedocs.io/en/latest/git/troubleshooting.html#recover-from-a-git-rebase-failure) your pull request's merge conflicts accordingly.

jeremielodi:
@lomamech, this PR is ready for the review

jniles:
@jeremielodi, some notes from the presentation today:

 1. Updating rubrics doesn't seem to work - if an error occurred it needs to be highlighted in red and shown as a `Notify.handleError()` call.
 2. The accounts in the rubric configuration grid shows "[null] null" when there is no account defined. Since the accounts are not required for indice, this should just show a blank celll.
 3. In the `staffing_indices` modal, replace the "actions" dropdown with just a simple "delete" link.  Since the only action is delete, it makes sense to just have it directly accessible.

lomamech:
@jeremielodi There are major advancements,
- We need to improve the business management module, when we select the option, Activate the system of payment by index, the check boxes do not mention anything when we consult Enterprise Settings

- It is necessary to improve the scripts to import in the file migrate.sql, it is not important to have another file

- For the creation headings (indices), it is not necessary to have to inform the elements such as (expressed in percentage, is related to Emploiyee,

image

server/controllers/payroll/rubrics/rubricsIndex.js
```diff
@@ -0,0 +1,105 @@
+module.exports = rubircs();
+
+function rubircs() {
```
lomamech, line 3: @jniles @jeremielodi @mbayopanda  In the future, this approach may have to be revised just because this condemns the logic of index management to that of the CBCO hospitals, including VANGA and KITAMBO.

server/models/migrations/next/migrate.sql
```diff
@@ -13,3 +13,348 @@ INSERT INTO unit VALUES
 
 INSERT INTO `report` (`id`, `report_key`, `title_key`) VALUES
```
lomamech: @jeremielodi fix this error INSERT INTO `report` (`report_key`, `title_key`) VALUES

lomamech (commented):
@jeremielodi Missing the script that must add the property enable_index_payment_system

lomamech:
@jniles  @mbayopanda  All module tests were successful,

jniles:
Woot!  Congratulations @jeremielodi and @lomamech !  🎆

PR #3751, opened 2019-05-28
Merge patients feature

mbayopanda:
This PR add a way of merge patients into one by selecting the one to keep. All transactions of patients to remove will be affected to patient to keep.

qJtEQ8gGTt

### DIFF
--- a/server/controllers/medical/patients/merge.js
+++ b/server/controllers/medical/patients/merge.js
@@ -1,0 +1,101 @@
+/**
+ * This controller is responsible of handling merge of patients into
+ * one patient.
+ *
+ * The merge consists to :
+ * - get debtor_uuids of the selected patients and others
+ * - replace all occurences of debtor_uuid of patient to remove by the debtor_uuid
+ *   of the selected patient in journal and general ledger
+ * - remove the not selected patient
+ * - recalculate period total
+ */
+const db = require('../../../lib/db');
+
+exports.mergePatients = mergePatients;
+
+function mergePatients(req, res, next) {
+  const params = req.body;
+  const glb = {};
+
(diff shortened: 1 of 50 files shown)


server/controllers/medical/patients/merge.js
```diff
@@ -0,0 +1,102 @@
...
+
+      transaction.addQuery(replacePatientInEmployee, [selectedPatientUuid, [otherPatientUuids]]);
+      transaction.addQuery(replacePatientInPatientAssignment, [selectedPatientUuid, [otherPatientUuids]]);
+      transaction.addQuery(replacePatientInPatientDocument, [selectedPatientUuid, [otherPatientUuids]]);
+      transaction.addQuery(replacePatientInPatientHospitalization, [selectedPatientUuid, [otherPatientUuids]]);
+      transaction.addQuery(replacePatientInPatientVisit, [selectedPatientUuid, [otherPatientUuids]]);
+
+      transaction.addQuery(removeOtherPatients, [otherPatientUuids]);
+      transaction.addQuery(removeOtherDebtors, [glb.otherDebtorUuids]);
+      return transaction.execute();
```
mbayopanda, line 109: Since we merge just two patients not more, i think i can replace all these codes by a MySQL procedure.

The reason of these staff are just in the case we will have to merge more than one patients into a one, it is not obvious to send an array of uuid to a MySQL procedure.

lomamech (commented):
@mbayopanda,  I am analyzing the features of the patient merge procedure, I found that there is a problem when trying to merge two patients who are also employees there is a duplication of key on the employee table

lomamech:
@mbayopanda I will suggest that he may have restrictions on merging patients for the patient who are registered as employees.

PR #3753, opened 2019-05-30
fix(invoice): primary key collision invoicing_fees

jniles:
The invoice creation threw errors when two invoicing_fees of the same
_value_ were applied to the same invoice.   This is obvious incorrect.
This change allows you to apply any value any number of times to the
same invoice, provided you are not reusing the same invoicing fee.

### DIFF
--- a/server/models/migrations/next/migrations.sql
+++ b/server/models/migrations/next/migrations.sql
@@ -109,7 +109,7 @@ CREATE PROCEDURE UnbalancedInvoicePaymentsTable(
 
   DROP TEMPORARY TABLE IF EXISTS tmp_invoices_2;
   CREATE TEMPORARY TABLE tmp_invoices_2 AS SELECT * FROM tmp_invoices_1;
-  
+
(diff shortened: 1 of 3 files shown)


server/models/migrations/next/migrations.sql
```diff
@@ -187,4 +187,10 @@ CREATE PROCEDURE UnbalancedInvoicePaymentsTable(
   );
 END$$
 
-DELIMITER ;
\ No newline at end of file
+-- author: jniles
+-- date: 30/05/2019
+-- Fix primary key collisions at HEV.
+ALTER TABLE invoice_invoicing_fee DROP PRIMARY KEY;
+ALTER TABLE invoice_invoicing_fee ADD PRIMARY KEY (invoice_uuid, invoicing_fee_id);
```
mbayopanda, line 194: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3755, opened 2019-05-31
Update improve payroll

### LINKED ISSUE #3754: Update of the Payroll module,
this issue proposes to ensure that when entering accounting entries for wage commitments, the closing date of the payment period is used.
 
Add tax configuration option that is not expressed as a percentage

lomamech:
- Sanitaze code remove unused require employees
- Use the date to instead the current days for commitment of paiement in
  accounting
- Update the max length of abbreviation of rubrics
- remove unused require
- Before we had to prevent the edition of all the headings which were
  taxes, but it can be made that there are taxes which is not calculated
  expressed as a percentage, to allow the edition of this kind of tax, it
  would be enough to exclude the only tax which is expressed as a
  percentage neither and which is not nor éditable reason why we had
  excluded the IPR
- closes #3754

### DIFF
--- a/client/src/modules/payroll/rubrics/modals/rubric.modal.html
+++ b/client/src/modules/payroll/rubrics/modals/rubric.modal.html
@@ -23,7 +23,7 @@
 
     <div class="form-group" ng-class="{ 'has-error' : RubricForm.$submitted && RubricForm.label.$invalid }">
       <label class="control-label" translate>FORM.LABELS.ABBREVIATION</label>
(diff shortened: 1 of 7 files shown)


lomamech:
@jniles  i finish with this pull request, can you make a review

lomamech:
@jeremielodi @mbayopanda  can you review this Pull Request,

mbayopanda (approved):
Since we have to consider the date of the payment's period instead of the current date, the code LGTM

bors r+

PR #3760, opened 2019-06-07
feat(component) bhperiod, use period translate_key as label

jeremielodi:
This PR fixes the fiscal period label translation issue


jniles (approved):
LGTM.

bors r+

PR #3767, opened 2019-06-19
Auto email reports by using cron job

mbayopanda:
This PR add a new feature of sending report automatically by email.

In this PR, we have :
- In the enterprise settings an option to enable or disable this feature of emailing
- Create a component to add in reports modules for having this feature on the report module
- Added this feature in : 
  - Balance Report
  - Bilan Report
  - Compte Resultat Report
  - Compte d'exploitation Report

PCfDwQk18n


client/src/js/components/bhCronEmailReport/bhCronEmailReport.js
```diff
@@ -0,0 +1,114 @@
...
+
+  $ctrl.$onInit = init;
+
+  function init() {
+    $ctrl.cron = {
+      report_id : $ctrl.reportId,
+      report_url : $ctrl.reportUrl,
+      has_dynamic_dates : 0,
+    };
+    $ctrl.isFeatureEnabled = Session.enterprise.settings.enable_auto_email_report;
```
jniles: Let's name this `isEmailFeatureEnabled` or something a little more precise.  Even though we are in the $scope of a component, it will be helpful in the future to know what feature is enabled if the application throws an error.

client/src/js/components/bhCronEmailReport/bhCronEmailReport.html
```diff
@@ -0,0 +1,85 @@
...
+
+        <!-- dates behaviour -->
+        <bh-yes-no-radios
+          label="CRON.DYNAMIC_DATES"
+          value="$ctrl.cron.has_dynamic_dates"
+          on-change-callback="$ctrl.onChangeDynamicDates(value)">
+        </bh-yes-no-radios>
+  
+        <div class="text-right">
+          <button data-method="save-cron-report" type="submit" class="btn btn-primary" translate>
```
jniles, line 41: Nice `data-method` name!

client/src/js/components/bhCronEmailReport/bhCronEmailReport.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  function remove(id) {
+    CronEmailReports.delete(id)
+      .then(() => load($ctrl.reportId))
+      .catch(Notify.handleError);
+  }
+
+  function submit(cronForm, reportForm) {
+    if (reportForm.$invalid) {
```
jniles: I don't think you need to add this as a parameter.  Isn't it available as `$ctrl.reportForm`?  If so, is there a reason not to just use `$ctrl.reportForm`?

server/models/schema.sql
```diff
@@ -1369,6 +1388,32 @@ CREATE TABLE `entity` (
...
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+DROP TABLE IF EXISTS `cron_email_report`;
+CREATE TABLE `cron_email_report` (
+  `id` SMALLINT(5) NOT NULL AUTO_INCREMENT,
+  `entity_group_uuid` BINARY(16) NOT NULL,
+  `cron_id` SMALLINT(5) NOT NULL,
+  `report_id` SMALLINT(5) NOT NULL,
+  `report_url` VARCHAR(200) NOT NULL,
```
jniles: Do we need to store the `report_url`?  Can't we just look it up from the `report` table?

I recommend we just look it up - it will be _much_ easier to make database migrations that way.  We'll only need to change the URL in one place.

server/lib/moment.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * this class helps to get start date and end date of a period
+ */
+const moment = require('moment');
```
jniles: I advise that we do two things with this library:
 1. rename it to something like `bhmoment` or something similar.  The name `moment` will be misleading to inexperienced developers.
 2. Write some unit tests!  If we call it with the date '2015-01-01', what should `.day()` return?  What about `.week()`?  etc...

server/lib/mailer.js
```diff
@@ -37,15 +37,26 @@ function processAttachments(attachments = []) {
       // default to the name of the file if the name has not been specified
       attach.filename = attach.filename || path.parse(attach.path).base;
 
-      debug(`#processAttachments() loading ${attach.path}`);
+      if (!attach.stream) {
```
jniles, line 40: Just FYI, you could combine this using an empty promise to save on having multiple return values:
```js
return q()
.then(() => {
  if (attach.stream) { return attach.stream; }
  return fs.readFile(attach.path);
})
.then(file =>  new mailgun.Attachment({ filename : attach.filename, data : file }));
```

client/src/i18n/en/cron.json
```diff
@@ -0,0 +1,25 @@
...
+  "CRON":{
+    "TITLE":"Cron",
+    "AUTO_EMAIL_REPORT":"Auto sending report by email",
+    "FREQUENCY":"Frequency",
+    "RECEIVERS":"Receivers",
+    "DAILY":"Daily",
+    "WEEKLY":"Weekly",
+    "MONTHLY":"Monthly",
+    "YEARLY":"Yearly",
+    "EVERY_MINUTE":"Every Minutes",
```
jniles: Should this be "every minute"?  Or "each minute"?

client/src/i18n/en/cron.json
```diff
@@ -0,0 +1,25 @@
...
+    "SAVE":"Save",
+    "SAVED_AUTO_REPORT":"Saved report for mailing",
+    "LAST":"Last Send",
```
jniles: This should be "last sent".

client/src/i18n/en/cron.json
```diff
@@ -0,0 +1,25 @@
...
+    "NEXT":"Next Send",
+    "PLEASE_FILL_REPORT_FORM":"Please fill correctly the report form",
```
jniles: It looks like this is because of an error in the report form.   If so, we should say that.  Something like: "Oops!  It looks like there is an error in the report configuration form.  Please fill it in correctly before saving the email configuration."

client/src/i18n/en/cron.json
```diff
@@ -0,0 +1,25 @@
...
+    "PLEASE_FILL_CRON_FORM":"Please fill correctly the email form",
```
jniles: A similar comment to the one above: "The email form isn't filled out correctly.  Please check the values and resubmit."

client/src/i18n/en/cron.json
```diff
@@ -0,0 +1,25 @@
...
+    "DATES_BEHAVIOR":"Dates behaviors",
+    "FIXED_DATES":"Fixed dates",
+    "DYNAMIC_DATES":"Dynamic dates",
+    "EMAIL_SENT_SUCCESSFULLY":"Email sent successfully",
+    "SEND":"Send",
+    "DELETE":"Delete"
```
jniles: Doesn't DELETE exist somewhere already?  Like `FORM.BUTTONS.DELETE`?  Can we just use that?

client/src/i18n/en/enterprise.json
```diff
@@ -24,7 +24,9 @@
       "ENABLE_BARCODES_LABEL" : "Enables barcodes throughout the application",
       "ENABLE_BARCODES_HELP_TEXT" : "Enable to place barcodes on most printed records and enable options to scan barcodes for document inputs.",
       "ENABLE_AUTO_STOCK_ACCOUNTING_LABEL" : "Enables realtime stock accounting",
-      "ENABLE_AUTO_STOCK_ACCOUNTING_HELP_TEXT" : "Enabling this feature will write stock movement transactions into the posting journal in real time. It requires all inventory accounts to be correctly configured."
+      "ENABLE_AUTO_STOCK_ACCOUNTING_HELP_TEXT" : "Enabling this feature will write stock movement transactions into the posting journal in real time. It requires all inventory accounts to be correctly configured.",
+      "ENABLE_AUTO_EMAIL_REPORT_LABEL" : "Enable sending automatic reports by email",
+      "ENABLE_AUTO_EMAIL_REPORT_HELP_TEXT" : "Enabling this option gives access to the reporting configuration for automatic emails"
```
jniles: This text is probably better: "Enabling this option allows users to configure schedules to automatically send reports to a list of email addresses."

client/src/js/components/bhCronEmailReport/bhCronEmailReport.html
```diff
@@ -0,0 +1,85 @@
...
+        <bh-input-text
+          key="label"
+          text-value="$ctrl.cron.label"
+          label="FORM.LABELS.LABEL"
+          autocomplete="off"
+          required="true">
+        </bh-input-text>
+  
+        <!-- group -->
+        <div class="form-group">
```
jniles: I don't think you need this `form-group`.  Looks like you have included it in the `bhEntityGroupSelect` component.

client/src/js/components/bhCronEmailReport/bhCronEmailReport.html
```diff
@@ -0,0 +1,85 @@
...
+          <bh-entity-group-select
+            uuid="$ctrl.cron.entity_group_uuid"
+            on-select-callback="$ctrl.onSelectEntityGroup(entityGroup)"
+            required="true">
+          </bh-entity-group-select>
+        </div>
+  
+        <!-- frequency -->
+        <div class="form-group">
```
jniles: It doesn't look like you need this `form-group` `<div>`.  You have included it in the `bhCronSelect` component.

jniles (commented):
@mbayopanda, I've not gotten a chance to test this, but have written some of my thoughts below.

mbayopanda:

test/server-unit/bhMoment.spec.js
```diff
@@ -0,0 +1,48 @@
+const { expect } = require('chai');
+const BhMoment = require('../../server/lib/bhMoment');
+
+const dateString = '2019-08-24';
+const startTime = '00:00:00.000';
+const endTime = '23:59:59.999';
+
+const date = new BhMoment(dateString);
```
jniles, line 8: :+1:

test/integration/cronEmailReport/cronEmailReport.js
```diff
@@ -0,0 +1,71 @@
...
+      .then((res) => {
+        helpers.api.created(res);
+        record.id = res.body.id;
+      })
+      .catch(helpers.handler);
+  });
+
+  it('GET /cron_email_reports list registered cron email reports', () => {
+    return agent.get('/cron_email_reports')
+      .send(record)
```
jniles, line 39: Do you really need this `.send()`?  I don't think it applies to GET requests.

jniles (approved):
Looks good to me!  I'll fix the migration script, then we can merge it.

PR #3787, opened 2019-07-08
chore: add migration script generator

jniles:
A migrations script generator has been added directly to repository and to the npm scripts section of the `package.json`.  The idea is that it will read the configuration found in the `.env` file and create a migration script that can be run to upgrade the files to the next version.

**Usage**
```sh
yarn migrate
```

Here is what it looks like running with the "imck" database.

_GIF 1: Running with the IMCK database_


jniles:
@mbayopanda can you take a look at this and give me your feedback?

jniles:
@mbayopanda?

mbayopanda (approved):
LGTM 👍 

bors r+

sh/setup-migration-script.sh
```diff
@@ -0,0 +1,70 @@
...
+# NOTE: the file used to determined variables is the .env.development file.  Please make sure
+# this file is up to date.
+
+# TODO(@jniles) - look up current version, and pull in all versions until present version.
+
+echo "[migrate] Migrating BHIMA database"
+
+set -a
+source .env.development
+set +a
```
mbayopanda, line 19: On most of our servers, bhima is running in productioin mode `.env.production`, with this `.env.development` above, we will be oblige to handle migration in production mode in another way than `yarn migrate` (and the mysql DATABASE < MIGRATION_SCRIPT)

Issue #3819, opened 2019-07-25
(bug) bad data displayed in the Cashflow by service (Journal de ventilation)


mbayopanda:
In the Cashflow by service report (Journal de ventilation).

It seems like the cashflow by service report is buggy, we found a case where we have :
- Two projects : project_1 and project_2
- Two cashboxes : cash_project_1 and cash_project_2
- The cashflow by service report for the cash_project_1 shows data of the cash_project_2

mbayopanda:
Fixed by #3820

PR #3820, opened 2019-07-25
Fix Cashflow by service report and search modal for caution in payment registry

mbayopanda:
This PR fix the issue #3819 

And enhance the search modal in payment registry for caution by allowing to filter payments which are caution or payments which are not caution.

### DIFF
--- a/client/src/modules/templates/bhYesNoRadios.tmpl.html
+++ b/client/src/modules/templates/bhYesNoRadios.tmpl.html
@@ -14,7 +14,7 @@
         ng-value="1"
         ng-model="$ctrl.value"
         ng-change="$ctrl.onChange($ctrl.value)"
-        required>
+        ng-required="$ctrl.required">
       <span translate>FORM.LABELS.YES</span>
     </label>
 
@@ -26,7 +26,7 @@
         ng-value="0"
         ng-model="$ctrl.value"
         ng-change="$ctrl.onChange($ctrl.value)"
-        required>
+        ng-required="$ctrl.required">
       <span translate>FORM.LABELS.NO</span>
     </label>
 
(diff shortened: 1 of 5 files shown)


server/controllers/finance/reports/cashflow/index.js
```diff
@@ -117,7 +117,7 @@ function reportByService(req, res, next) {
   db.one(cashboxDetailsSql, cashboxId)
```
jniles: !!!  Can we rename this variable from `cashboxId` to `cashboxAccountId`?  The origin of this bug seems to be from a bad name.

jniles (approved):
Looks good to me.  Great catch!

bors r+

jniles:
Welp.

bors r+

client/src/modules/templates/bhYesNoRadios.tmpl.html
```diff
@@ -12,7 +14,7 @@
         ng-value="1"
         ng-model="$ctrl.value"
         ng-change="$ctrl.onChange($ctrl.value)"
-        required>
+        ng-required="$ctrl.required">
```
jniles: Why do we need to toggle the required if we have a `bhClear`?  It seems like both of them are addressing the same solution.

mbayopanda: You're right @jniles

jniles (approved):
Looks good to me!

mbayopanda:
@jniles could review this PR please

jniles:
I did!  And I approved it.  You can merge anytime.

mbayopanda:
Thanks @jniles

bors r+

PR #3825, opened 2019-07-30
fix (Invoicing Bug)

### LINKED ISSUE #3822: Problem with the call of the function Debtors.balance
Problem with the call of the function Debtors.balance, indeed this function verifies if the customers has a creditor balance by checking if the sum of the credits and superior to the sums of debits, it posed a problem concerning the transactions whose difference is expressed in terms of 4 digits after the decimal point, for this it is necessary to round the values of credits and debits before making the comparison

````
      const hasCreditorBalance = hasPrepaymentSupport && pBalance && (pBalance.credit > pBalance.debit);
      const preparedTransaction = createInvoice(invoice, hasCreditorBalance, prepaymentDescription);
      return preparedTransaction.execute();

```` 
 /server/controllers/finance/patientInvoice.js

### PR DESCRIPTION
**Pull request by lomamech**
- resolution of the problem when calling the Debtors.balance function
with rounding to two ranks after the decimal point of the total credit and
debit values

closes #3822

### DIFF
--- a/server/controllers/finance/debtors/index.js
+++ b/server/controllers/finance/debtors/index.js
@@ -261,9 +261,15 @@ function balance(debtorUuid, excludeCautionLinks = false) {
 
   const excludeCautionLinkStatement = `AND transaction_type_id <> ${CAUTION_LINK_TYPE_ID}`;
 
+  /**
+   * resolution of the problem when calling the Debtors.balance function with
+   * rounding to two ranks after the decimal point of the total credit and
+   * debit values
+   *
+   */
   const sql = `
-    SELECT IFNULL(SUM(ledger.debit_equiv), 0) AS debit, IFNULL(SUM(ledger.credit_equiv), 0) AS credit,
-      IFNULL(SUM(ledger.debit_equiv - ledger.credit_equiv), 0) AS balance, MIN(trans_date) AS since,
+    SELECT ROUND(IFNULL(SUM(ledger.debit_equiv), 0), 2) AS debit, ROUND(IFNULL(SUM(ledger.credit_equiv), 0),2) AS credit,
+    ROUND(IFNULL(SUM(ledger.debit_equiv - ledger.credit_equiv), 0),2) AS balance, MIN(trans_date) AS since,
       MAX(trans_date) AS until
     FROM (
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by lomamech** [comment]
@jniles  @mbayopanda  this PR resolve the probleme of patient KALOMBO MUKENGE in IMCK database, 
````
 SELECT IFNULL(SUM(ledger.debit_equiv), 0) AS debit, IFNULL(SUM(ledger.credit_equiv), 0) AS credit,
      IFNULL(SUM(ledger.debit_equiv - ledger.credit_equiv), 0) AS balance, MIN(trans_date) AS since,
      MAX(trans_date) AS until
    FROM (
      SELECT debit_equiv, credit_equiv, entity_uuid, trans_date FROM posting_journal
        WHERE entity_uuid = X'6dacb561ff174212b16cfda1cffd26c4'
      UNION ALL
      SELECT debit_equiv, credit_equiv, entity_uuid, trans_date FROM general_ledger
        WHERE entity_uuid = X'6dacb561ff174212b16cfda1cffd26c4'
    ) AS ledger
    GROUP BY ledger.entity_uuid;
````
The balance of this patient was 0.003, this value caused the billing module to be in error and gave the impression that this patient had paid a deposit, which is why I modified the SQL query by rounding up the two-digit values after the decimal point
image

**Comment by jniles** [comment]
@lomamech it seems like the problem comes from here:

```js
const hasCreditorBalance = hasPrepaymentSupport && pBalance && (pBalance.credit > pBalance.debit);
```

Instead of changing the SUM() function, could we just change the check `pBalance.credit > bBalance.debit` to something like `(pBalance.credit - pBalance.debit) > 0.01`?

I'm concerned that the `ROUND()` function will not work well when we deal with really cheap items.  For example, if 1 pill is 5FC, it is `0.003` USD, which would be `ROUND(0.003, 2)` = 0.  For this reason, instead of adjusting the debits/credits via a `ROUND()` function, let's just change the display logic for when we show the warning.

jniles:
Interesting - it looks like you tried something like that here: https://github.com/IMA-WorldHealth/bhima/pull/3823.  What went wrong with that PR?

lomamech:
@jniles  @mbayopanda  all things are okay with this PR

jniles (approved):
Looks good to me.  :+1:

bors r+

PR #3829, opened 2019-08-06
ci: add initial appveyor support

### LINKED ISSUE #887: (test) Run integration and end-to-end tests on separate CIs
I propose that we shard our tests over multiple continuous integration services.  This would allow our tests to speed up overall, as well as limit the affect of a single network outage on our testing procedures.

I recommend Semaphore CI, as I've heard good things about its speed.  But we could explore our options.

I see the ideal testing like this:
 1. Travis -> installs dependencies and runs end-to-end tests
 2. Other Service -> runs Lint, Karma, and integration tests.

jniles:
Start running a few tests using appveyor to take the pressure of TravisCI.  The end goal is to run our tests in two CI environments to speed up the tests and also give us better coverage.  For example, right now we only test on Travis on trusty.  It would be nice to move over to Bionic or even Windows support as referenced in https://github.com/IMA-WorldHealth/bhima/issues/2477.

Closes https://github.com/IMA-WorldHealth/bhima/issues/887.


jniles:
@mbayopanda looks like this is working on appveyor now.

I've only migrated the server-unit tests so far, but there is no reason we can't make it work for the integration tests as well.  I propose that we try to see if it respects bors, then continue adding in more tests.

What do you think?

test/server-unit/pdf.spec.js
```diff
@@ -53,7 +53,7 @@ function PDFRenderUnitTest() {
     expect(isBuffer && hasValidVersion).to.be.equal(true);
   });
 
-  it('#pdf.render() renders an identical PDF given an HTML template', async () => {
+  it.skip('#pdf.render() renders an identical PDF given an HTML template', async () => {
```
mbayopanda: I think there is a good reason for this `.skip`, i would like to know it was running on Travis why you skipped it here.

sh/test.sh
```diff
@@ -27,9 +27,9 @@ startfold "Running Integration Tests..." "test-integration";
 endfold "test-integration" ;
 
 #run server-unit test
-startfold "Running server Unit Tests ......" "server-unit"
-./node_modules/.bin/mocha --recursive --exit test/server-unit
-endfold "server-unit" ;
+# startfold "Running server Unit Tests ......" "server-unit"
+# ./node_modules/.bin/mocha --recursive --exit test/server-unit
+# endfold "server-unit" ;
```
mbayopanda: Is it possible to add a comment about the fact that server-unit tests are running on `appveyor` or just remove these commented codes.

mbayopanda (approved):
LGTM 👍

sh/test.sh
jniles: Good point.  I'll remove this code.

test/server-unit/pdf.spec.js
jniles: I skipped this because I didn't want to set up XVFB.  Let me see if I can reproduce it with XVFB like on Travis CI.

jniles: Well, it looks like the PDF output isn't deterministic: https://github.com/wkhtmltopdf/wkhtmltopdf/issues/4437.  I'm going to just remove this test since it won't ever work properly.

jniles:
@mbayopanda, changes made!

jniles:
I'm going to just roll this in to see if bors respects the appveyor restrictions.  I've already gotten approval.

bors r+

jniles:
bors r+

... round 2!

PR #3832, opened 2019-08-07
fix(Calcul Sum Distributed)

### LINKED ISSUE #3831: Calculating the sum of keys when the distribution key values are decimal
For auxiliary center distribution to the main centers using the distribution keys, there is a problem when the values are decimal, which is why this issue corrects the problem by rounding the integer key sum
image

lomamech:
- Around the sum distributed for escape error when percent as decimal

closes #3831


lomamech:
@mbayopanda  check this PR

mbayopanda (approved):
bors r+

PR #3834, opened 2019-08-08
Handle negative values in stock reports

mbayopanda:
This PR is about to fix disorder caused by negative quantities in stock reports.

Negative quantity come when we distribute a stock lot more than what we have.

In the inventory file report, this negative value affect the calculated cost the inventory which cause biases in reports.

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -602,6 +602,9 @@ function getInventoryMovements(params) {
         if (line.is_exit) {
           stockQuantity -= line.quantity;
           stockValue = stockQuantity * stockUnitCost;
+          // fix negative value disorder
+          // ignoring negative stock value by setting them to zero for entry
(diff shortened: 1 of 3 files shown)


jniles:
@mbayopanda, do you know why stock value would be negative?  What scenario makes it negative?

mbayopanda:
@jniles negative values was due to the fact that in the stock exit module, within the grid, people filled the grid a lot of rows for the same inventory, and after that they set lots, this behavior allowed to define  the same lot in multiple line.

If the available quantity for a lot ABC is for example 100, they can do these exits :  
- ABC : 100
- ABC : 20
Which will produce -20 as quantity of ABC. We have already fixed that in the stock exit module, but we have to do the same for stock adjustment

jeremielodi (approved):
LGTM

PR #3844, opened 2019-08-15
fix(grid): rendering bug on reversed records

### LINKED ISSUE #3816: The red highlight on ui-grid rows isn't redrawn often enough
_GIF 1: The red highlight doesn't get redrawn on a large grid, leading to eventual coloration of the entire grid_

The above GIF illustrates the problem.  For some reason, the red highlight on the rows doesn't refresh as often as the data within the individual cells, causing the red highlight to stay on rows that it shouldn't be associated with.

jniles:
Fixes the rendering bug on reversed records that caused the red highlight to hang around on non-reversed records due to template caching.  The solution also removes an `ngStyle` tag.

Closes #3816.

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -875,6 +875,11 @@ a[disabled] {
   text-decoration : line-through;
 }
 
+.reversed {
+  text-decoration : line-through;
+  background-color: #ffb3b3 !important;
+}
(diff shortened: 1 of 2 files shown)


mbayopanda (approved):
bors r+

PR #3846, opened 2019-08-15
refactor(Make Exchange Rates human readable)

### LINKED ISSUE #3838: Make Exchange Rates human readable
At one of our client hospitals, we have a different default enterprise currency.  The result is a very hard to read exchange rate page.  We should improve this so that the numbers are never smaller than 0.1 by showing the inverse (instead of 1USD to X money, show 1X money to YYYUSD).

MakeExchangeRatesMoreHumanReadable

lomamech:
- Addition of the minimum value of the exchange rate to prevent the
  exchange rate equal to zero
- Show exchange rates less than 0.1 as a fraction for example 1/1630
  instead of 0.0006134

closes #3838

### DIFF
--- a/client/src/modules/exchange/exchange.modal.html
+++ b/client/src/modules/exchange/exchange.modal.html
@@ -53,7 +53,8 @@
         </div>
         <div class="col-xs-10">
           <div class="input-group">
-            <input class="form-control" type="number" name="rate" ng-model="ModalCtrl.rate.rate" required>
+            <!-- Improve me Prevent exchange rate equal 0 -->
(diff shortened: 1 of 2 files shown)


lomamech:
image

@jniles  all things are okay

client/src/modules/exchange/exchange.modal.html
```diff
@@ -53,7 +53,8 @@
         </div>
         <div class="col-xs-10">
           <div class="input-group">
-            <input class="form-control" type="number" name="rate" ng-model="ModalCtrl.rate.rate" required>
+            <!-- Improve me Prevent exchange rate equal 0 -->
```
jniles, line 56: Great idea. 👍

jniles (approved):
Looks good to me.

bors r+

Issue #3157, opened 2018-09-07
Remove influence of timezone on some dates values which doesn't need it


mbayopanda:
This issue is about to remove timezone information on some dates which doesn't need timezone information.

This issue concerns for example the fiscal year module, during the creation step when we define the date interval, the date interval doesn't need timezone additional information. The current state of the fiscal year module use timezone info on the dates ranges values, which has as consequence, for West Africa timezone when we choose 2018-12-31 we got 2019-01-01.

the proposal is to send date as just a simple string "2018-12-31" instead of a date object.

sfount:
This is slightly related to https://github.com/IMA-WorldHealth/bhima-2.X/issues/1274 and https://github.com/IMA-WorldHealth/bhima-2.X/issues/1793 specifically

> Proposed solution:
> Do not send a full timestamp from the client including time and time zone. Uniformly format period dates as 'DD/MM/YYYY'.

It would be great to have a documented standard for processing and transferring dates on the client and the server.

kwilu:
Hello @jniles, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@kwilu claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

PR #3847, opened 2019-08-15
fix(bhDateInterval): send dates in UTC format

jniles:
This commit sends the dates to the server in UTC format, by
adding/subtracting the UTC format. It is a stopgap until we implement
the #3157.


client/src/js/components/bhDateInterval.js
```diff
@@ -95,20 +98,27 @@ function bhDateInterval(moment, bhConstants, Fiscal, Session) {
     $ctrl.dateTo = new Date();
   }
 
+  function momentStartHelper(interval) {
+    return moment().startOf(interval).add(tzOffset, 'minutes').toDate();
```
mbayopanda, line 102: Good 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3848, opened 2019-08-15
fix(cash): broken barcode binding

jniles:
The barcodes on the main cash receipt all read 'null'.  This is due to an improper binding in the handlebars that has been fixed.


mbayopanda (approved):
LGTM :+1

bors r+

PR #3850, opened 2019-08-16
fix(cash): prevent multiple barcode read()s

### LINKED ISSUE #3839: Barcode reader launches multiple HTTP requests
The barcode scanner launches a series of HTTP requests to the server.  I thought we had fixed this issue, but it appears to have either reappeared or never been addressed.  Below, you can see the log of a single barcode scan observed in production:
BarcodeScanForCashResultsinLotsofHTTPRequests
_Fig 1: Barcode Scan for Cash Payment_

We need to limit these successive requests.  We have the [HTTPCacheService ](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/services/HttpCacheService.js) that might help limit subsequent queries.  Otherwise, we need to play with `$timeout`s to make sure we don't send multiple queries on a single `$digest`.

jniles:
Prevents multiple HTTP requests from being sent after each barcode
scan to read in invoices.  Should increase server performance.

Steps to reproduce:
 1. Launch the server
 2. Find a (valid) invoice barcode to scan.  
 3. Go the the cash module (`/cash`).
 4. Find the invoice via the "scan from barcode" dropdown option.
 5. Observe that only two HTTP requests are sent to the server instead of multiple.

Closes #3839.

### DIFF
--- a/client/src/modules/cash/cash-form.service.js
+++ b/client/src/modules/cash/cash-form.service.js
@@ -207,8 +207,7 @@ function CashFormService(AppCache, Session, Patients, Exchange) {
 
     // bind the enterprise currency value
     this.totals.enterpriseCurrencyTotal = total;
-    this.totals.currentExchangeRate =
-      Exchange.getExchangeRate(this.details.currency_id, this.details.date);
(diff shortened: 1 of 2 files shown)


mbayopanda (approved):
LGTM 👍 

bors r+

PR #3854, opened 2019-08-20
chore: clean up old code, tests, and references

jniles:
Removes 1.x tables and references from the database and test data.  Adds a migration script to drop these in production since all current production sites are running with these tables.  Finally, cleans up residual inventory/depots.js code that stuck around from 1.x and is unused in the current bhima codebase.


mbayopanda (approved):
LGTM 👍 

bors r+

PR #3864, opened 2019-08-28
Bhima data collector

### LINKED ISSUE #3863: Feature creation of a Bhima data collection system
This issue proposes the creation of a complex system of data collection in BHIMA, this system will collect data, configure the survey form of a
dynamic and simple way,

The data to be collected can be of numeric type, text, choice list, text box, date, time, image (Photo), calculations with a particular synthax,

The use of this system will include the realization of three actions: development of a Data Collector Manager, lists of choices, as well as the elements of the survey form,
   Once these three conditions are met, data collection can begin.
  
   Some forms will have a link to patient management, it must also have a report with a mini search engine for data from surveys

lomamech:
- Implente Client side for Data Collector
- Implete API for Data Collector Management
- Extend service color js adding new color
- Implement integration test for dataCollectorManagement
- Implement E2E test for dataCollectorManagement

This module makes it possible to configure the different types of the
choice lists, the lists of choices can be in the form of trees.
- Implement client side of choice list management
- Implement API for choice list Management
- Add integration and E2E for choice List Management

This module allows to add the elements of the form of a data collector,
the elements of a form have a type, some elements can have filter
compared to other elements
- Add new component for Data Collector
- Add new component for select Survey Form Type Select
- Add new component for Select Survey List Select
- Implement client side for Survey Form management
- Implement API for survey form manangement
- Add Integration and End to End test for Survey Form management

- Implementation of APIs related to the filling of forms as well as the
  display of survey data,

- Implementation of links between patient-related forms on the patient
  management page,

- Implementing the data collection report with a small search engine

- Implementation of components for Choicelist Multipke Select, Patient
  Medical Sheet, Survey Form Select,

- Adding Integration and End to End test for Displaying Metadata and
  Filling of survey Form

closes #3863


lomamech:
@jniles  @mbayopanda  @jeremielodi All the tests pass for this PR, do not hesitate to consult me for the review of this PR

jeremielodi:
The data collector module works perfectly, there are just few things that we have to fix

- We can put the module "Collecte des données" and "Kit de collecte de données" together for better 
   understanding of users, we can just use "Collecte des données" as the unique module. So inside we 
    will have the setup and filling of surveys.

- /#!/survey_form 
 the add field button should be disabled if the survey is not selected

- #!/data_collector_management
 There should be a link between a survey on this grid with his configuration in  /#!/survey_form page

- #!/display_metadata
 Add the Export data into Ms Excel  feature
 Add the number of submissions for each servey

client/src/modules/survey_form/modals/survey_form.modals.html
```diff
@@ -0,0 +1,206 @@
...
+          required
+          rows="4">
+        </textarea>
+        <div class="help-block" ng-messages="SurveyFormForm.calculation.$error" ng-show="SurveyFormForm.$submitted">
+          <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+        </div>
+      </div>
+    </div>
+
+    <div ng-if="SurveyFormModalCtrl.selectList" style="background-color:#ededed; padding-left: 10%; padding-right: 3%">
```
jniles: I'm not a fan of this padding - it looks a bit off.  There is a bootstrap class that is pretty close to what you want though - it is called `.well`.  See how it works below.

8fW4Iki2dZ

Would you be willing to change these `styles` to `class="well"`?

client/src/modules/display_metadata/display_metadata.html
```diff
@@ -0,0 +1,121 @@
...
+                      'pdf', 
+                      DisplayMetadataCtrl.changes, 
+                      DisplayMetadataCtrl.collectorId,
+                      DisplayMetadataCtrl.filterElements,
+                      DisplayMetadataCtrl.patient.uuid,
+                      DisplayMetadataCtrl.patient) }}">
+                    <span class="fa fa-file-pdf-o"></span> <span translate>DOWNLOADS.PDF</span>
+                  </a>
+                </li>
+                <li role="separator" class="divider"></li>
```
jniles: Let's comment out the "separator" until the rest are done.  It makes the dropdown menu look a bit funny:

DisplayMetadataExtraButton

client/src/modules/display_metadata/display_metadata.html
```diff
@@ -0,0 +1,121 @@
...
+        <div class="toolbar-item">
+            <div uib-dropdown dropdown-append-to-body data-action="open-tools">
+              <a class="btn btn-default" uib-dropdown-toggle>
+                <span class="fa fa-bars"></span> <span class="hidden-xs" translate>FORM.LABELS.MENU</span> <span class="caret"></span>
+              </a>
+    
+              <ul uib-dropdown-menu role="menu" class="dropdown-menu-right">
+                <li role="menuitem">
+                  <a target='_blank'
+                    ng-href="/data_kit/report/?{{ DisplayMetadataCtrl.download(
```
jniles: Instead of having this large `DisplayMetadataCtrl.download()` call, it would be better to move this into the controller in a function called `downloadPDF()`.  This href will become `ng-href="/data_key/report?{{DisplayMetadataCtrl.downloadPDF()"`.

And then in your controller you can make a function:
```js
ctrl.downloadPDF = () => {
  return ctrl.download('pdf', ctrl.changes, ctrl.collectorId, /* etc ... */ );
};
```

client/src/modules/display_metadata/display_metadata.html
```diff
@@ -0,0 +1,121 @@
...
+            </div>
+          </div>
+
+      <div ng-if="DisplayMetadataCtrl.patientData" class="toolbar-item">
+        <button class="btn btn-default text-capitalize" id="create" ui-sref="display_metadata.patientfill({id : DisplayMetadataCtrl.collector.id, patient: DisplayMetadataCtrl.patient.uuid})" data-method="create">
+          <span class="fa fa fa-file-text"></span> <span class="hidden-xs" translate>FORM.BUTTONS.FILL_FORM</span>
+        </button>
+      </div>
+
+      <div ng-if="DisplayMetadataCtrl.patientData" class="toolbar-item">
```
jniles: You have two `ng-if="DisplayMetadataCtrl.patientData"` next to each other.  Can you merge them?

client/src/modules/display_metadata/display_metadata.service.js
```diff
@@ -0,0 +1,116 @@
...
+  service.openSearchModal = openSearchModal;
+  service.displayFilters = displayFilters;
+  service.download = download;
+
+  function download(type, changes, id, filterClient, patientUuid, patient) {
+    const options = {
+      renderer : type,
+      changes,
+      lang : Languages.key,
+      patient_uuid : patientUuid || null,
```
jniles: Why do you specify `patient` and `patientUuid` as two different parameters?

client/src/modules/display_metadata/display_metadata.service.js
```diff
@@ -0,0 +1,116 @@
...
+      filterClient,
+      patient,
+    };
+
+    // return  serialized options
+    return $httpParamSerializer(options);
+  }
+
+  function listSurveyformtype() {
+    const url = ''.concat('listSurveyformtype');
```
jniles: `const url = ''.concat('listSurveyformtype');` is the same as `const url = 'listSurvveyformType'`.  Go ahead and just use a plain string.

client/src/modules/display_metadata/display_metadata.service.js
```diff
@@ -0,0 +1,116 @@
...
+ * Encapsulates common requests to the /display_metadata/ URL.
+ */
+function DisplayMetadataService(Api, Modal, moment, $translate, $httpParamSerializer, Languages) {
+  const service = new Api('/display_metadata/');
+  service.listSurveyformtype = listSurveyformtype;
+  service.openSearchModal = openSearchModal;
+  service.displayFilters = displayFilters;
+  service.download = download;
+
+  function download(type, changes, id, filterClient, patientUuid, patient) {
```
jniles: This function has a lot of parameters.  Can you just make it receive an `opts` object instead?  It would look something like this?

```js
function download(type, opts) {
  const options = Object.assign({}, opts,  { renderer : type });
}
```

This would be easier for other developers to work with since they will not have to remember the order of each parameter.  It will also be more flexible if you need to change something in the future.

client/src/modules/display_metadata/display_metadata.service.js
```diff
@@ -0,0 +1,116 @@
...
+          }
+        });
+      });
+    }
+
+    if (search.multipleChoice) {
+      const multipleChoiceLength = Object.keys(search.multipleChoice).length;
+      if (multipleChoiceLength) {
+        Object.keys(search.multipleChoice).forEach((key) => {
+          survey.forEach(item => {
```
jniles: You are using `survey.forEach()` a lot in this function!  Let's make lookups more efficient by making a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) from the `survey` array.

Basically:
```js
const surveyMap = new Map(survey.map(item => ([item.name, item]));
```

Now, instead of writing:
```js
survey.forEach(item => {
  if (item.name === key) {
     // do something 
  }
});
```
you can write:
```js
const item = surveyMap.get(key);
```

If the item doesn't exist, `surveyMap.get()` will return `undefined`.

client/src/modules/display_metadata/display_metadata.service.js
```diff
@@ -0,0 +1,116 @@
...
+      backdrop : 'static',
+      controller : 'DisplayMetadataSearchModalController as $ctrl',
+      resolve : {
+        filters : function filtersProvider() { return filters; },
+      },
+    }).result;
+  }
+
+  function displayFilters(survey, search) {
+    let filters = ``;
```
jniles: I'm not a big fan of using `+=` to add to a string.  Could you instead use an array, do `array.push()` for each string, and then later to `array.join('')` to turn it into a larger string?

My logic is that it lets you use array functions (`forEach`, `map`, etc) on each item if you need them later.  It also looks a bit cleaner.

client/src/modules/display_metadata/display_metadata.html
```diff
@@ -0,0 +1,121 @@
...
+        <button
+          ng-click="DisplayMetadataCtrl.search()"
+          data-method="search"
+          class="btn btn-default">
+          <span class="fa fa-search"></span> <span class="hidden-xs" translate>FORM.BUTTONS.SEARCH</span>
+        </button>
+      </div>
+
+      <div class="toolbar-item">
+        <button
```
jniles, line 63: We have a component for this - [bhFilterToggle](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/components/bhFilterToggle.js).  You could use it here to reduce code:

```html
 <bh-filter-toggle on-toggle="DisplayMetadataCtrl.toggleFilter()">
 </bh-filter-toggle>
```

client/src/modules/display_metadata/display_metadata.html
```diff
@@ -0,0 +1,121 @@
...
+      </div>
+    </div>
+  </div>
+</div>
+
+<!-- main content -->
+<div class="flex-util" style="overflow: auto" ng-if="DisplayMetadataCtrl.collector">
+  <span data="filter">
+    <span class="label" style="text-shadow: {{ DisplayMetadataCtrl.collector.color }} 5px 3px 8px; font-size: 15px; color: black">
+      <span> {{ DisplayMetadataCtrl.collector.label }} </span>
```
jniles: Do you need these extra `<span>`s?

client/src/modules/display_metadata/display_metadata.html
```diff
@@ -0,0 +1,121 @@
...
+    </span>
+  </span>
+
+  <span ng-if="DisplayMetadataCtrl.filterElements" data="filter">
+    <span class="label" ng-class="{ 'label-primary' : true, 'label-emphasize' : true }">
```
jniles: Do you need an `ngClass` like this `ng-class="{ 'label-primary' : true, 'label-emphasize' : true }"`?   Why not just make these regular HTML classes?
```html
<span class="label label-primary label-emphasize">
```

client/src/modules/display_metadata/display_metadata.html
```diff
@@ -0,0 +1,121 @@
...
+      {{ DisplayMetadataCtrl.filterElements }}
+      <a href ng-click="DisplayMetadataCtrl.onRemove()" class="filter-text-link">
+        <i class="fa fa-close"></i>
+      </a>
+    </span>
+  </span>
+
+  <span ng-if="DisplayMetadataCtrl.patientData" data="filter">
+    <span class="label"  style="font-size: 15px; color: black; text-shadow: gray 4px 1px 10px;">
+      <span>:{{ DisplayMetadataCtrl.patient.display_name }} / {{ DisplayMetadataCtrl.patient.reference}} </span>
```
jniles: It looks like you have extra spans here as well.

client/src/modules/display_metadata/display_metadata.js
```diff
@@ -0,0 +1,208 @@
...
+  // options for the UI grid
+  vm.gridOptions = {
+    appScopeProvider  : vm,
+    enableColumnMenus : false,
+    fastWatch         : true,
+    flatEntityAccess  : true,
+    enableSorting     : true,
+    showColumnFooter  : true,
+    onRegisterApi     : onRegisterApiFn,
+    enableHorizontalScrollbar : uiGridConstants.scrollbars.WHEN_NEEDED,
```
jniles, line 71: This is cool 👍

client/src/modules/display_metadata/display_metadata.js
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  // options for the UI grid
+  vm.gridOptions = {
+    appScopeProvider  : vm,
+    enableColumnMenus : false,
+    fastWatch         : true,
+    flatEntityAccess  : true,
+    enableSorting     : true,
+    showColumnFooter  : true,
```
jniles, line 69: Do you do anything with the column footer?  I don't see it being used in my tests.

client/src/modules/display_metadata/display_metadata.js
```diff
@@ -0,0 +1,208 @@
...
+
+  ChoicesList.read()
+    .then(choicesLists => {
+      vm.choicesLists = choicesLists;
+    })
+    .catch(Notify.handleError);
+
+  if ($state.params.id && $state.params.patient) {
+    // Prevent a non-patient form from being used to collect non-patient data
+    vm.patientData = true;
```
jniles: Since this is a boolean, let's call it `hasPatientData`.

client/src/modules/display_metadata/display_metadata.js
```diff
@@ -0,0 +1,208 @@
...
+    vm.gridOptions.enableFiltering = vm.filterEnabled;
+    vm.gridApi.core.notifyDataChange(uiGridConstants.dataChange.COLUMN);
+  }
+
+  // bind methods
+  vm.edit = edit;
+  vm.remove = remove;
+
+  function edit(data) {
+    if (!$state.params.patient) {
```
jniles: We already checked for this in the startup code with `hasPatientData`.  Could we use that variable instead of `$state`?

```js
if (hasPatientData) {
  // do x
} else {
  // do y
}
```

client/src/modules/display_metadata/display_metadata.js
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  // load user grid
+  function loadGrid() {
+    toggleLoadingIndicator();
+    // Reuse cash unless patient data is consulted
+    if (cache.collector && !vm.collectorId && !$state.params.patient) {
+      vm.collectorId = cache.collector.id;
+    }
+    const changesLength = Object.keys(vm.changes).length;
```
jniles: Let's make this into a boolean:
```js
const hasNoChanges = Object.keys(vm.changes).length === 0;
```
Then we can use it in a nice check:
```
if (cache.changes && hasNoChanges) {
  // do x
}
```

client/src/modules/display_metadata/display_metadata.js
```diff
@@ -0,0 +1,208 @@
...
+      search();
+    } else if (vm.collectorId && !$state.params.id) {
+      vm.hasError = false;
+      vm.loading = true;
+
+      SurveyForm.read(null, { data_collector_management_id : vm.collectorId })
+        .then((survey) => {
+          vm.filterElements = DisplayMetadata.displayFilters(survey, vm.changes);
+
+          vm.params = {
```
jniles: Does `vm.params` depend on the result of `SurveyForm.read()`?  If not, let's move it out:
```js
vm.params = {
  data_collector_management_id : vm.collectorId,
  changes : vm.changes,
};

SurveyForm.read(/* etc ...*/)
```

client/src/modules/display_metadata/display_metadata.js
```diff
@@ -0,0 +1,208 @@
...
+            changes : vm.changes,
+          };
+
+          return DisplayMetadata.read(null, vm.params);
+        })
+        .then((data) => {
+          vm.gridOptions.columnDefs = data.columns;
+          vm.gridOptions.data = data.surveyData;
+
+          return DataCollectorManagement.read(vm.collectorId);
```
jniles: You have a nice promise chain here that sends a request for:
 1. `SurveyForm.read()`
 2. `DisplayMetadata.read()`
 3. `DataCollectorManagement.read()`

However, it doesn't look like any of these actually depend on one another.  So, let's send them in parallel using `Promise.all()`!

```js
vm.params = {/* etc  */};
return Promise.all([
  SurveyForm.read(null, { data_collector_management_id : vm.collectorId }),
  DisplayMetadata.read(null, vm.params),
  DataCollectorManagement.read(vm.collectorId),
])
.then(([survey, data, collector]) => {
  // process all queries...
});
```

client/src/modules/display_metadata/display_metadata.js
```diff
@@ -0,0 +1,208 @@
...
+      vm.loading = true;
+      vm.collectorId = $state.params.id;
+      // View Patient Form Data
+      vm.params = {
+        data_collector_management_id : $state.params.id,
+        patient_uuid : $state.params.patient,
+        changes : vm.changes,
+      };
+
+      DisplayMetadata.read(null, vm.params)
```
jniles, line 216: See if you can do a similar `Promise.all()` system here.

jniles (commented):
@lomamech,

This PR is incredible - it represents a lot of work, and it performs very well.  Well done!

As discussed, I've limited my review to the `DisplayMetaData` modules for the moment, and mostly the graphical user interface.

Let me know if you have any questions!

client/src/modules/display_metadata/display_metadata.service.js
lomamech: The parameter 'patient'  contains patient information for the purpose of no longer querying the database for its information

jniles: Can you just use `patient.uuid`?

client/src/modules/display_metadata/display_metadata.html
lomamech: I could not use the bh-filter-toggle component just because in normal use the elements that use this component come from the elements defined in advance in the FilterService service, whereas the elements of the structure of the grid comes from the database

client/src/modules/display_metadata/display_metadata.js
lomamech: Thanks, I add the number of rows 
image

client/src/modules/display_metadata/display_metadata.js
lomamech: @jniles, i try this options but the displaying became too slow

client/src/modules/display_metadata/display_metadata.js
lomamech: image
The displaying became very very too slow

client/src/js/components/bhChoiceListMultipleSelect.js
```diff
@@ -0,0 +1,56 @@
...
+    controller  : ChoiceListMultipleSelectController,
+    bindings    : {
+      onChange : '&',
+      lists : '<?',
+      label : '@?',
+      required : '<?',
+      listLabel        : '@?',
+      isTitle          : '<?',
+      isGroup          : '<?',
+      parentId         : '<',
```
jniles: It looks like the `parentId` is optional (line 36).  This should be `<?`.

client/src/js/components/bhChoiceListSelect.js
```diff
@@ -0,0 +1,64 @@
...
+    transclude  : true,
+    bindings    : {
+      onSelectCallback : '&',
+      list             : '<?',
+      disable          : '<?',
+      required         : '<?',
+      listLabel        : '@?',
+      isTitle          : '<?',
+      isGroup          : '<?',
+      parentId         : '<',
```
jniles: This should also be optional `<?`.

kwilu:
Hello @lomamech, it seems like you have referenced #3863 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #3863.`.

```
Author: kwilu

    Fixes #51.
```

kwilu:
Hello @lomamech, it seems like you have referenced #3863 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #3863.`.

```
Author: kwilu

    Fixes #51.
```

lomamech:
@jniles I have finish with the first review

client/src/modules/choices_list_management/modals/choices_list_management.modals.js
```diff
@@ -0,0 +1,84 @@
...
+  function onSelectGroup(group) {
+    vm.choice.group_label = group.id;
+  }
+
+  function clear(value) {
+    vm.choice[value] = 0;
+  }
+
+  function closeModal() {
+    $state.transitionTo('choices_list_management');
```
jniles: Most of our `$state` changes use `$state.go()`.  Can you use that there?

client/src/modules/choices_list_management/templates/action.cell.tmpl.html
```diff
@@ -0,0 +1,37 @@
+<div class="ui-grid-cell-contents text-right" uib-dropdown dropdown-append-to-body data-row="{{row.entity.label}}">
+  <span ng-if="!row.entity.fixed">
+    <a href uib-dropdown-toggle data-action="open-dropdown-menu">
+      <span data-method="action" translate>FORM.BUTTONS.ACTIONS</span>
+      <span class="caret"></span>
+    </a>
+
+    <ul data-row-menu="{{row.entity.label}}" class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
+      <li class="dropdown-header text-ellipsis" style="max-width:250px;">
+        <a> {{row.entity.label}} </a>
```
jniles: Why does this need to be wrapped in a `<a>` tag?  Can you just put it in plain `<span>` tag instead?

client/src/modules/data_collector_management/modals/data_collector_management.modals.js
```diff
@@ -0,0 +1,74 @@
...
+      })
+      .catch(Notify.handleError);
+  }
+
+  function clear(value) {
+    vm.dataCollectors[value] = null;
+  }
+
+  function closeModal() {
+    $state.transitionTo('data_collector_management');
```
jniles: `$state.go()` should be sued here as well.

client/src/modules/choices_list_management/templates/gridName.tmpl.html
```diff
@@ -0,0 +1,8 @@
+<div class="ui-grid-cell-contents" data-title-row="{{ row.entity.label }}">
+  <span ng-if="row.entity.is_title" style="text-transform: uppercase">
+    {{ row.entity.label}}
+  </span>
+  <span ng-if="!row.entity.is_title" style="font-weight:normal; font-style: italic; text-align: right">
```
jniles: Do you need the `font-style:italic` if you have wrapped the element in `<i>`?

client/src/modules/data_collector_management/modals/data_collector_management.modals.html
```diff
@@ -0,0 +1,98 @@
...
+        <input name="label" ng-model="DataCollectorManagementModalCtrl.dataCollector.label" autocomplete="off" ng-maxlength="90" class="form-control" required>
+        <div class="help-block" ng-messages="DataCollectorManagementForm.label.$error" ng-show="DataCollectorManagementForm.$submitted">
+          <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+        </div>
+      </div>
+    </div>
+    <!-- description -->
+    <div class="form-group" ng-class="{ 'has-error' : DataCollectorManagementForm.$submitted && DataCollectorManagementForm.description.$invalid }">
+      <label class="control-label" translate>FORM.LABELS.DETAILS</label>
+      <div>
```
jniles: Why do you need to wrap this text area in a `<div>`?

client/src/modules/data_collector_management/modals/data_collector_management.modals.html
```diff
@@ -0,0 +1,98 @@
...
+        <div class="help-block" ng-messages="DataCollectorManagementForm.description.$error" ng-show="DataCollectorManagementForm.$submitted">
+          <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+        </div>
+      </div>
+    </div>
+
+    <!-- version_number -->
+    <div class="form-group" ng-class="{ 'has-error' : DataCollectorManagementForm.$submitted && DataCollectorManagementForm.version_number.$invalid }">
+      <label class="control-label" translate>FORM.LABELS.VERSION</label>
+      <div>
```
jniles: It looks like you are wrapping a lot of elements in `<div>`s - is there a reason for this?

client/src/modules/data_collector_management/modals/data_collector_management.modals.js
```diff
@@ -0,0 +1,74 @@
...
+  '$state', 'DataCollectorManagementService', 'NotifyService', 'appcache', 'ColorService',
+];
+
+/**
+ * Data Collector Management Modal Controller
+ */
+
+function DataCollectorManagementModalController($state, DataCollectorManagement, Notify, AppCache, Color) {
+  const vm = this;
+  const cache = AppCache('AccountReferenceModal');
```
jniles: This shouldn't be `AccountReferenceModal`.

client/src/modules/survey_form/modals/survey_form.modals.js
```diff
@@ -0,0 +1,108 @@
...
+  '$state', 'SurveyFormService', 'NotifyService', 'appcache', 'DataCollectorManagementService',
+];
+
+/**
+ * SURVEY FORM Modal Controller
+ */
+
+function SurveyFormModalController($state, SurveyForm, Notify, AppCache, DataCollectorManagement) {
+  const vm = this;
+  const cache = AppCache('AccountReferenceModal');
```
jniles: This probably shouldn't be `AccountReferenceModal`.

client/src/modules/survey_form/modals/survey_form.modals.js
```diff
@@ -0,0 +1,108 @@
...
+    .then(surveyForms => {
+      vm.surveyForms = surveyForms;
+    })
+    .catch(Notify.handleError);
+
+  // submit the data to the server from all two forms (update, create)
+  function submit(surveyForm) {
+    vm.hasNoChange = surveyForm.$submitted && surveyForm.$pristine && !vm.isCreating;
+    if (surveyForm.$invalid) { return null; }
+    if (!surveyForm.$dirty) { return null; }
```
jniles: Instead of `!surveyForm.$dirty`, go ahead and use `surveyForm.$pristine`.  It is the same thing, but without a `!`.

client/src/modules/fill_form/modals/fill_form.modals.js
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  function clear(value) {
+    delete vm.form[value];
+  }
+
+  function closeModal() {
+    const transitionTo = vm.updateMode ? 'display_metadata' : 'fill_form';
+    if (!vm.stateParams.patient) {
+      $state.transitionTo(transitionTo);
```
jniles: Can you use `$state.go()` here?

client/src/modules/patients/record/patient-record.html
```diff
@@ -132,6 +132,15 @@
           </bh-patient-financial-activity>
         </div>
       </div>
+
+      <div class="row">
+        <!-- Medical Sheet -->
+        <div class="col-xs-12">
+          <bh-patient-medical-sheet patient-uuid="PatientRecordCtrl.patient.uuid">
```
jniles, line 139: I propose we name this something else - "medical sheet" makes it sound like it has something to do with the patient's medical information, but it is actually just data collections about the patient.

Let's call this `<bh-patient-data-collection-forms>` or something like that.

jniles (commented):
@lomamech Great work on making these changes.

I'm still working through most of the code, but I've made a few more comments.  Take a look :)

:+1:

client/src/modules/patients/record/patient-record.html
lomamech: the use of the term "medical sheet" simply wanted to mention that this area is dedicated to medical information of patients.

Logically, the forms that should appear on this zone should only be those collecting medical information

jeremielodi:
I propose that we use "Survey Form management" instead of "Survey Form" and
"Survey Form setting" instead of  "Data Collector Management", that way the module names will be more understandable

jeremielodi:
Make sure the user can see a warning message if he submit an invalid form.
Currently you return null in your JavaScript code for this problem.

lomamech:
@jniles @jeremielodi I wait another review

client/src/modules/choices_list_management/choices_list_management.service.js
```diff
@@ -0,0 +1,100 @@
...
+ * @class ChoicesListManagementService
+ * @extends PrototypeApiService
+ *
+ * @description
+ * Encapsulates common requests to the /choices_list_management/ URL.
+ */
+function ChoicesListManagementService(Api) {
+  const service = new Api('/choices_list_management/');
+
+  service.formatStore = formatStore;
```
jniles: Hmm... looks like we have identical code in the [AccountService](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/accounts/accounts.service.js#L115).  Is there any way we can move this into a shared service so that we don't have to duplicate code?

client/src/modules/data_collector_management/modals/data_collector_management.modals.html
```diff
@@ -0,0 +1,94 @@
...
+
+    <div class="checkbox">
+      <label>
+        <input
+          type="checkbox"
+          id="is_related_patient"
+          ng-true-value="1"
+          ng-false-value="0"
+          ng-model="DataCollectorManagementModalCtrl.dataCollector.is_related_patient">
+        <span translate>
```
jniles: This `translate` tag needs to be on the `<strong>` not the `<span>`.

client/src/js/components/bhPatientMedicalSheet.js
```diff
@@ -0,0 +1,108 @@
...
+      DataCollectorManagement.read(null, { is_related_patient : 1 })
+        .then((dataCollectorManagement) => {
+          $ctrl.gridOptions.data = dataCollectorManagement;
+        })
+        .catch(handleError)
+        .finally(toggleLoadingIndicator);
+    }
+
+    function toggleLoadingIndicator() {
+      $ctrl.loading = false;
```
jniles, line 98: Toggle loading indicator always sets "loading" to be "false".  Shouldn't it turn loading on and off?

client/src/js/components/bhPatientMedicalSheet.js
```diff
@@ -0,0 +1,108 @@
...
+ * to display the different medical sheet to record the Patients Medical Data
+ */
+function PatientMedicalSheetCtrl(DataCollectorManagement, Notify, uiGridConstants, $state) {
+  const $ctrl = this;
+
+  this.$onInit = () => {
+    $ctrl.gridApi = {};
+    $ctrl.filterEnabled = false;
+    $ctrl.toggleFilter = toggleFilter;
+    $ctrl.consult = consult;
```
jniles, line 28: Just FYI - for functions, you don't need to put them in the `$onInit` call.  The `$onInit()` is most for variables that are passed in and/or need to be initialized with each component.

client/src/modules/fill_form/fill_form.service.js
```diff
@@ -0,0 +1,86 @@
...
+  service.restoreImage = restoreImage;
+  service.formatConstraint = formatConstraint;
+
+  function formatData(form, data) {
+    const dataSurveyForm = {};
+    // Ici l'on verifie les occurences des donnees existantes
+    // S'il sont multiples afin de les affecter dans des tableaux
+    form.forEach(f => {
+      data.forEach(d => {
+        if ((f.id === d.survey_form_id) && (f.type === '4')) {
```
jniles, line 26: What is `type === '4'`?  Why not `type === '7'`?

client/src/modules/fill_form/fill_form.service.js
```diff
@@ -0,0 +1,86 @@
...
+          dataSurveyForm[f.name] = [];
+        }
+      });
+    });
+
+    form.forEach(f => {
```
jniles, line 32: You are doing two `.forEach()` loops here that you could combine.    Just move lines 26 through 28 to the start of the next loop and it should save you a whole loop.

client/src/modules/fill_form/fill_form.service.js
jniles, line 26: It looks like you use this check more than once.  Can you make a function that tells us what it is?  Like this:
```js
function isValidSurveyForm(form, data) {
  return form.id === data.survey_form_id && form.type === '4';
}
```

Then you can just do:
```js
form.forEach(f => {
  data.forEach(d => {
    if (isValidSurveyForm(f, d)) {
      // do something.
    }
  });
});

File: client/src/modules/fill_form/fill_form.service.js [hunk 41]
```diff
@@ -0,0 +1,86 @@
...
+      data.forEach(d => {
+        // Pour les identifiants des listes des choix,
+        // il est primordiale de le transormer en Integer
+        if ((f.id === d.survey_form_id) && (f.type === '4')) {
+          // Multiple choice list management
+          dataSurveyForm[f.name].push(parseInt(d.value, 10));
+        } else if ((f.id === d.survey_form_id) && (f.type !== '4') && (f.type !== '7')) {
+          // Single choice list management
+          d.value = (f.type === '1') ? parseFloat(d.value) : d.value;
```
jniles: Aha, so `f.type` gives the question type.  Could you set those as constants in this service or in the `bhConstants` constant?

In this way, it will be much more readable.  For example:
```js
d.value = (f.type === QUESTION_DATE_TYPE) ? new Date(d.value) : d.value;
d.value = (f.type === QUESTION_CHOICE_LIST_TYPE) ? coolfunction() : null;
```

client/src/modules/fill_form/fill_form.service.js
```diff
@@ -0,0 +1,86 @@
...
+          d.value = (f.type === '3') ? parseInt(d.value, 10) : d.value;
+
+          // format date
+          d.value = (f.type === '6') ? new Date(d.value) : d.value;
```
jniles: You are writing something like `d.value = (f.type === x) ? onething() : d.value` over and over again.  Let's  extract this code into a separate parser function that uses a [`switch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) statement.  Something like this:

```js
d.value = parseDataValue(d.value, f.type);


// somewhere 
function parseDataValue(value, type) {
  switch(type) {
    case '1':
      return parseInt(value, 10);
    case '4':
      return new Date(value);
    default:
      return value;
  }
}
```

If you make this a function _and_ you implement the constants suggestion above, you will get a really nice looking function:
```js
function parseDataValue(value, type) {
  switch(type) {
    case QUESTION_INTEGER_TYPE:
      return parseInt(value, 10);
    case QUESTION_DATE_TYPE:
      return new Date(value);
    // etc...
    default:
      return value;
  }
}
```
My syntax for the switch function may be wrong, be sure to test it out.

client/src/modules/fill_form/fill_form.service.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /*
+    * This function is used to format the value of the constraint
+    * to a valid angular expression, the value encapsulate in. {} Must be transformed into
+    * FillFormModalCtrl.form. and if the eval function
+    * is incorrectly evaluates the return function a null value
+  */
+
+  function formatConstraint(constraint) {
```
jniles, line 77: I like this function - putting code in a function like this makes it possible to write good tests for it.  Well done! 👍 

Did you write a couple unit tests for this function? 😄

client/src/modules/fill_form/modals/fill_form.modals.html
```diff
@@ -0,0 +1,250 @@
...
+    <ol class="headercrumb">
+      <li class="static" translate>TREE.DATA_COLLECTION</li>
+
+      <li class="title">
+        <span translate>TREE.FORMS_CONFIGURATION</span>
+      </li>
+    </ol>
+  </div>
+
+  <div style="background-color: {{ FillFormModalCtrl.dataCollector.color }}; border-radius : 25px ; text-align: center; text-decoration-color: #ffffff; text-shadow: rgb(255, 255, 255) 5px 3px 8px;">
```
jniles, line 12: Can you use [`ngStyle`](https://docs.angularjs.org/api/ng/directive/ngStyle) here instead of hacking in the style?

client/src/modules/fill_form/modals/fill_form.modals.html
```diff
@@ -0,0 +1,250 @@
...
+    <h3><strong>
+      <span translate>
+        {{ FillFormModalCtrl.dataCollector.label }}
+      </span>
+    </strong></h3>
+    <h4><strong ng-if="FillFormModalCtrl.patient.display_name"> {{ FillFormModalCtrl.patient.display_name }} / {{ FillFormModalCtrl.patient.reference}} ({{ FillFormModalCtrl.patient.sex }}) </strong></h4>
+  </div>
+
+  <div class="modal-body" popover-append-to-body="true" style="overflow-y: scroll; min-height:300px; max-height:600px; ">
```
jniles: Where did this "popover-append-to-body" come from?  What does it do?

client/src/modules/fill_form/modals/fill_form.modals.html
```diff
@@ -0,0 +1,250 @@
...
+    </strong></h3>
+    <h4><strong ng-if="FillFormModalCtrl.patient.display_name"> {{ FillFormModalCtrl.patient.display_name }} / {{ FillFormModalCtrl.patient.reference}} ({{ FillFormModalCtrl.patient.sex }}) </strong></h4>
+  </div>
+
+  <div class="modal-body" popover-append-to-body="true" style="overflow-y: scroll; min-height:300px; max-height:600px; ">    
+    <div ng-repeat="formItem in FillFormModalCtrl.formItems">
+
+      <div ng-if="!formItem.constraint || $eval(formItem.constraintCheck)">
+        <!-- Type text -->
+        <div ng-if="formItem.typeForm === 'text'" class="form-group" ng-class="{ 'has-error' : FillFormForm.$submitted && FillFormForm.{{ formItem.name }}.$invalid }">
```
jniles: As discussed together, I think that you can simply do `FillFormForm[formItem.name].$invalid` and save yourself a bit of parsing.

client/src/modules/fill_form/modals/fill_form.modals.html
```diff
@@ -0,0 +1,250 @@
...
+          <label class="control-label"> {{ formItem.label }} </label>
+          <span style="color: #cecece"><em> {{ formItem.hint }} </em></span>
+          <div>
```
jniles: Do you need this `<div>` here?

client/src/modules/fill_form/modals/fill_form.modals.html
```diff
@@ -0,0 +1,250 @@
...
+            <span ng-if="formItem.required">
+              <input name="{{ formItem.name }}" ng-model="FillFormModalCtrl.form[formItem.name]" autocomplete="off" ng-maxlength="90" class="form-control" required>
+            </span>
+            <span ng-if="!formItem.required">
```
jniles: Why not use [`ngRequired`](https://docs.angularjs.org/api/ng/directive/ngRequired)?

client/src/modules/fill_form/modals/fill_form.modals.html
```diff
@@ -0,0 +1,250 @@
...
+        </div>
+
+        <!-- Calculation -->
+        <div ng-if="formItem.typeForm === 'calculation'">
+          <span style="background-color: {{ FillFormModalCtrl.dataCollector.color }}; color: black"><em> <strong>{{ formItem.label }}</strong> : {{ formItem.calculation }}
+          <h5 ng-if="FillFormModalCtrl.updateMode">:{{ FillFormModalCtrl.form[formItem.name] }} </h5> </em></span> 
+        </div>
+
+        <!-- TextArea -->
+        <div ng-if="formItem.typeForm === 'text_area'">
```
jniles: Since you are doing a lot of `formItem.typeForm === "x"`, could you use an [`ngSwitch`](https://docs.angularjs.org/api/ng/directive/ngSwitch)?  It would probably make the code easier to read.

client/src/modules/reports/generate/dataKit/dataKit.config.js
```diff
@@ -0,0 +1,133 @@
...
+
+    if (vm.reportDetails.multipleChoice) {
+      const multipleChoiceLength = Object.keys(vm.reportDetails.multipleChoice).length;
+
+      if (multipleChoiceLength) {
+        Object.keys(vm.reportDetails.multipleChoice).forEach((key) => {
+          if (vm.reportDetails.multipleChoice[key].length) {
+            for (let i = 0; i < vm.reportDetails.multipleChoice[key].length; i++) {
+              vm.choicesLists.forEach(list => {
+                if (list.id === vm.reportDetails.multipleChoice[key][i]) {
```
jniles: So .. it looks like you are taking a list of ids and turning them into a list of `label`s.  So, you are trying to do:
```js
var ids = [3, 3, 5, 7, 1]
// do something
console.log(ids) // => ['formA', 'formA', 'formZ', 'formX', 'formT']
```

Let's try to clean this up a bit.  Here is what I think the basic structure could look like:
```js
angular.forEach(vm.reportDetails.mulitpleChoice, (values, key) => {
  // values is an array.
  values.forEach(value => {
    vm.choicesLists.forEach(list => {
      if (list.id === value) {
         // replace value with list.label
      }
    });
  });
});
```

Let's see if we can go one step further:
```js
// create a way to map id -> label for the choice list
let choiceListMap = vm.choicesList.reduce((map, item) => {
  map[item.id] = item.label;
}, {});

angular.forEach(vm.reportDetails.mulitpleChoice, (values, key) => {
  // unfortunately, angular.map() doesn't exist for object.  We have to use an ugly forEach
  vm.multipleChoice[key] = values.map(value => choiceListMap[value]);
});
```

I think that should do it.

First, we create an object to make choiceList ids map to their labels.  This takes up memory but removes the `choiceLists.forEach()` loop.

Then, we use `angular.forEach()` to loop through the object.  This removes all the `Object.keys()` calls and checks for `.length`.  

Try it out and see if I've missed anything.

jniles (commented):
@lomamech, a few more comments.  Your changes from last time look good.

client/src/modules/fill_form/fill_form.service.js
lomamech: image
It's form multiple value, look in bhima.sql in table survey_form_type

jniles: Can you instead use the label?  It will make it much easier for a developer to read.  Like this:
```js
f.label === 'FORM.LABELS.SELECT_MULTIPLE' // now all future developers knows what this means!
```

client/src/modules/fill_form/fill_form.service.js
lomamech: The reason is that in the first loop, we initialize a table for all the elements of the type 4 form, and in the second loop we simply push the elements in this array

client/src/modules/fill_form/fill_form.service.js
lomamech: @jniles, All type come from table survey_form_type
image

kwilu:
Heads up @lomamech, we just merged some commits that conflict with the changes your made in this pull request! You can review this repository's [recent commits](https://github.com/IMA-WorldHealth/bhima/commits/master) to see where the conflicts occur. Please rebase your feature branch against the `upstream/master` branch and [resolve](https://zulip.readthedocs.io/en/latest/git/troubleshooting.html#recover-from-a-git-rebase-failure) your pull request's merge conflicts accordingly.

lomamech:
@jniles @mbayopanda @jeremielodi can I have another Review

jniles:
bors try

lomamech:
@jniles  all test Pass

lomamech:
@jniles  @mbayopanda  After several attempts this old PR is ready to be merged

jniles:
I'm hoping we can just get this in.  Let's do it bors.

bors r+

jniles:
bors try

PR #3869, opened 2019-08-29
fix(accounts): order transactions chronologically

### LINKED ISSUE #3868: Account Statement does not always render writings in chronological order
We received this bug report:

> For the "Rapport de compte simple" it would be necessary to have the
opportunity to put all the writings in a chronological order. We
talked about that. My statement for 2019 starts in the middel of may,
what is confusing.

It is clear that sometimes the account statement report doesn't render writings in chronological order.

Here is an example:
BugInAccountOrdering
_Nice that 11/05 comes before 1/05_

jniles:
Orders first by the transaction date, then by the creation date.

Closes #3868.

Proof:
FixedAccountOrdering

### DIFF
--- a/server/controllers/finance/accounts/transactions.js
+++ b/server/controllers/finance/accounts/transactions.js
@@ -72,7 +72,7 @@ function getGeneralLedgerSQL(options) {
   `;
 
   filters.setGroup('GROUP BY record_uuid');
-  filters.setOrder('ORDER BY created_at ASC, trans_date ASC');
+  filters.setOrder('ORDER BY trans_date ASC, created_at ASC');
 
   const query = filters.applyQuery(sql);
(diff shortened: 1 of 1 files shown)


server/controllers/finance/accounts/transactions.js
```diff
@@ -72,7 +72,7 @@ function getGeneralLedgerSQL(options) {
   `;
 
   filters.setGroup('GROUP BY record_uuid');
-  filters.setOrder('ORDER BY created_at ASC, trans_date ASC');
+  filters.setOrder('ORDER BY trans_date ASC, created_at ASC');
```
mbayopanda, line 75: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3875, opened 2019-09-05
fix(cashflow): use Pivot() for byService results

### LINKED ISSUE #3837: The sum of columns in the Cashflow By Service report is not equal to the total
We have a bug in the SQL for the Cashflow by Service report.  The sum of the columns may not be equal to the total on the right hand side.  This is because the JOIN on cash_items only picks up a single cash item - if the cash payment was for more than one invoice, it will not be the right number.

We need to update the logic of the page so that it displays multiple cash_items on the same cash payment line, like a proper pivot table.  Perhaps we can even use the `Pivot()` stored procedure.

jniles:
Rewrites the cashflow by service report to use `Pivot()` instead of
manually executing queries.  The change fixes a bug where the columns
didn't sum to the cumsum.  There are two drawbacks with this method:
 1. The cumulative summation is done in JS.  Unfortunately, the Pivot()
 function doesn't support this, although we may be able to add it in the
 future.
 2. The results are not ordered.  Ideally they should be ordered by
 date.

Fixes #3837.

### DIFF
--- a/server/controllers/finance/reports/cashflow/index.js
+++ b/server/controllers/finance/reports/cashflow/index.js
@@ -26,12 +26,14 @@ const TEMPLATE_BY_SERVICE = './server/controllers/finance/reports/cashflow/repor
 exports.report = report;
 exports.byService = reportByService;
 exports.reporting = reporting;
+
 /**
- * This function creates a cashflow report by service, reporting the realized income
- * for the hospital services.
+ * @function reportByService
  *
- * @todo - factor in cash reversals.
- * @todo - factor in posting journal balances
+ * @description
+ * Called "Journal de Ventilation" in French.  Creates a pivot table of cash receipts
+ * divided out by the services that received in the income.  Rows are payments, columns
+ * are hospital service departments.
  */
 async function reportByService(req, res, next) {
   const dateFrom = new Date(req.query.dateFrom);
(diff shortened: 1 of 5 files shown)


jniles:
@mbayopanda, looks like this is ready for a review.

mbayopanda (commented):
Data in the report are correctly calculated, i just noticed that regardless of whether we select $ or Fc, we always have the same values but just the currency symbol change.

In Francs : 
cash franc


In Dollars : 
cash dollar

jniles:
@mbayopanda, great catch.  I just realized I need to be filtering by currency/account ID in my where statement.  I'll make the changes.

jniles:
@mbayopanda changes landed.

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -68,7 +68,7 @@ async function reportByService(req, res, next) {
   const whereQuery = `
     WHERE cash.is_caution = 0 AND cash.reversed = 0
       AND DATE(cash.date) >= DATE(?) AND DATE(cash.date) <= DATE(?)
-      AND cash.cashbox_id =  ?
+      AND cash.cashbox_id = ? AND cash.currency_id = ?
```
mbayopanda, line 71: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

mbayopanda:
@jniles could you update this branch with last changes of travis

mbayopanda:
bors try

PR #3883, opened 2019-09-11
fix: remove cache in production

### LINKED ISSUE #3881: Quit making users clear their browser cache to receive an upgrade in production
When we do an upgrade, it requires users to clear their browser cache.  We need to stop doing this.

The problem is likely that we cache the [index.html file indefinitely](https://github.com/IMA-WorldHealth/bhima/blob/master/server/config/express.js#L89) and we really should make an exception for the index.html file.  Since the index.html file isn't cache-busted, it fails to find all other files, and the user is presented a blank page.

jniles:
This commit removes the production cache so that assets are always loaded.  We can cache at the load balancer level.

Closes #3881

### DIFF
--- a/server/config/express.js
+++ b/server/config/express.js
@@ -21,7 +21,6 @@ const interceptors = require('./interceptors');
 const { Unauthorized } = require('../lib/errors');
 const uploads = require('../lib/uploader');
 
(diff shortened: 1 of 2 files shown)


mbayopanda (approved):
LGTM 👍 

bors r+

PR #3885, opened 2019-09-11
Use of xenial linux and xvfb as service for travis

mbayopanda:
- Change linux version from trusty to xenial
- Use of xvfb as service

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ node_js:
 
 before_script:
   - yarn build
-  - yarn webdriver-manager update --versions.chrome="$(google-chrome --version | cut -d ' ' -f 3)" --gecko false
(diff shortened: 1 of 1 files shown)


jniles:
@mbayopanda do we need `xvfb`?  I was under the impression we haven't used it in a long time.  I think only unpatched versions of `wkhtmltopdf` require it.

jniles:
So, I think what the documentation [here](https://docs.travis-ci.com/user/reference/xenial/#jvm-clojure-groovy-java-scala-support) is saying is to put in your travis.yml file:
```yaml

jdk:
  - oraclejdk8
```

[Some other people seem to have errors with it](https://travis-ci.community/t/oracle-jdk-11-and-10-are-pre-installed-not-the-openjdk-builds/785/18), but can you try it?

jniles:
Hey, looks like you got it!

jniles:
bors try

jniles:
Alright, I'll just merge this.  Great work @mbayopanda 

bors r+

PR #3894, opened 2019-09-18
chore(payroll): rm unnecessary db.transaction()

jniles:
We've replaced a db.transaction() with a Promise.all() call.  Since no
data updates were taking place (they are all SELECT queries), we don't
need the table locks of db.transaction().  It may also result in a minor
performance increase since queries are run in parallel.


jniles:
@lomamech, can you give me a review?

lomamech:
LGTM

bors r+

PR #3897, opened 2019-09-24
Realized cash from invoices

mbayopanda:
This report add a way to know from invoices made what cash was paid and what is remaining.


client/src/i18n/en/report.json
```diff
@@ -247,7 +247,13 @@
       "SHOW_OTHER_SUPPORT":"Show patient debts supported by organizations",
       "SHOW_OTHER_SUPPORT_HELP":"This option allows you to see the support provided by organizations that are not employees",
       "SHOW_DETAILS":"Show details",
-      "SHOW_DETAILS_HELP":"This option allows you to see more details"
+      "SHOW_DETAILS_HELP":"This option allows you to see more details",
+      "SHOW_REMAINING_DETAILS":"View details about remains (debts)",
```
jniles: "Remains" on its own is usually what you call a corpse ☠️ .  Go ahead and make this "View details about remaining debts".

client/src/i18n/en/tree.json
```diff
@@ -142,6 +142,7 @@
     },
     "VISITS_REPORT": "[Visits] Admission by Services",
     "CLIENT_DEBTS_REPORT": "[Clients] Debts Report",
-    "CLIENT_SUPPORT_REPORT": "[Clients] Support Report"
+    "CLIENT_SUPPORT_REPORT": "[Clients] Support Report",
+    "REALIZED_PROFIT_REPORT": "[Invoicing] Realized Profits"
```
jniles: _Technically speaking_, I think this should be "Realized Income" or "Realized Revenue".   The term "profit" is what you make after expenses are taken out.

server/controllers/finance/reports/realized_profit/index.js
```diff
@@ -0,0 +1,155 @@
...
+      moment(dateFrom).format('YYYY-MM-DD'), moment(dateTo).format('YYYY-MM-DD'),
+    ];
+
+    const transaction = db.transaction();
+    const tempTable = 'invoicePaymentsPivotTable';
+    const createTempTable = `
+      CREATE TEMPORARY TABLE ${tempTable} AS (${tableQuery.concat(groupByServiceAndDebtorGroup)});
+    `;
+
+    transaction.addQuery(`DROP TEMPORARY TABLE IF EXISTS ${tempTable};`);
```
jniles, line 105: Just FYI - I don't think you need to do a `DROP TABLE` query when you are using a transaction - transactions _should_ run in isolated memory environments and they should clean up any temporary tables at the end of execution.

server/controllers/finance/reports/realized_profit/index.js
```diff
@@ -0,0 +1,155 @@
...
+    `;
+
+    const parameters = [
+      moment(dateFrom).format('YYYY-MM-DD'), moment(dateTo).format('YYYY-MM-DD'),
+    ];
+
+    const transaction = db.transaction();
+    const tempTable = 'invoicePaymentsPivotTable';
+    const createTempTable = `
+      CREATE TEMPORARY TABLE ${tempTable} AS (${tableQuery.concat(groupByServiceAndDebtorGroup)});
```
jniles, line 102: Hmm.. it looks like you always put the `groupByServiceAndDebtorGroup` in the query.  Since that is the case, go ahead and write it as part of the `tableQuery`.  .e.g :
```sql
      SELECT
        SUM(w.paid) paid, SUM(w.invoiced) invoiced, SUM(w.invoiced - w.paid) remaining,
        w.debtorGroupName, w.serviceName, w.service_id, w.debtor_group_uuid 
      FROM (${globalQuery})w
      GROUP BY w.service_id, w.debtor_group_uuid;
```

jniles (approved):
@mbayopanda, this is really clean.  Your code is super easy to read and elegant.  I haven't tested it for performance issues, but we can always come back and try to improve it later.  Well done!

I've made a few minor comments to translation and variable declarations.  There isn't a need for a second review - once you make those changes, we can go ahead and merge the code.

server/controllers/finance/reports/realized_profit/index.js
mbayopanda: The `tableQuery` without `groupByServiceAndDebtorGroup` is used for getting totals.

server/controllers/finance/reports/realized_profit/index.js
mbayopanda: Without the `DROP TABLE` there is an error which is thrown

server/controllers/finance/reports/realized_profit/index.js
jniles: Aha, didn't see that.  No problem then.

server/controllers/finance/reports/realized_profit/index.js
jniles: :'(.  Everything I used to believe is a lie.  Thank you for correcting me!

jniles (approved):
Looks good to me.

bors r+

PR #3909, opened 2019-10-02
deps: upgrade old dependencies

jniles:
This mostly upgrades old dependencies, but I also caught a bug in the fiscal year's opening balance.  On errors, next() was called twice.

Closes #3906.


jniles:
bors try

mbayopanda (approved):
bors r+

PR #3910, opened 2019-10-03
test(csv): add unit tests for CSV renderer

### LINKED ISSUE #2651: Write unit tests for renderers/csv.js
We heavily depend on our [CSV renderer](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/server/lib/renderers/csv.js) for the downloads of almost all registries.  Writing unit tests will enable us to have confidence in the code.

**Tests to Write**
 1. The library should set three public properties: `extension`, `render` and `headers`.
 2. `render(data)` should return a promise.
 3. `render({ csv : [ /* some array */ ] })` should return a known string for a known input.
 5. `render()` should use a different key for finding if `options.csvKey` is set.
 6. `render()` should remove empty rows by default.
 7. `render()` should not remove empty rows is `suppressDefaultFiltering` is set.
 4. `render()` should convert dates if dates are passed in by default.
 8. `render()` should not convert dates if dates are passed with the `suppressDefaultFormatting` option.

jniles:
Adds in tests for the CSV renderer used on the server.  We also caught a typo in one of the options, fixed it, and fixed all known calls to the CSV renderer.

Closes #2651

### DIFF
--- a/server/controllers/finance/reports/account_statement/index.js
+++ b/server/controllers/finance/reports/account_statement/index.js
@@ -27,7 +27,7 @@ function report(req, res, next) {
     orientation : 'landscape',
     csvKey : 'rows',
     suppressDefaultFiltering : true,
-    suppressDefaultFormating : false,
+    suppressDefaultFormatting : false,
   });
 
   let rm;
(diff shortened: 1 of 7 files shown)


jeremielodi (approved):
Well done @jniles

PR #3913, opened 2019-10-03
Use of report key in cronEmailReport components

mbayopanda:
This commit replace the use of report_id by report_key in
cronEmailReport components, and also remove ids in report table during
insertion of reports in bhima.sql


server/models/bhima.sql
```diff
@@ -353,5 +353,4 @@ INSERT INTO `cron` (`label`, `value`) VALUES
   ('CRON.DAILY', '0 1 * * *'),
   ('CRON.WEEKLY', '0 1 * * 0'),
   ('CRON.MONTHLY', '0 1 30 * *'),
-  ('CRON.YEARLY', '0 1 31 12 *'),
-  ('CRON.EACH_MINUTE', '* * * * *');
```
jniles, line 357: :+1:

client/src/js/components/bhCronEmailReport/bhCronEmailReport.js
```diff
@@ -38,13 +39,19 @@ function bhCronEmailReportController(CronEmailReports, Notify, Session) {
...
+        $ctrl.cron = {
+          report_id : report.id,
+          has_dynamic_dates : 0,
+        };
+        $ctrl.isEmailFeatureEnabled = Session.enterprise.settings.enable_auto_email_report;
+        load(report.id);
+      });
+  }
+
+  function loadReportDetails(key) {
```
jniles, line 53: Just curious - why create a separate function for this?  Why not just call `BaseReport.requestKey()` in the `init()` function?

client/src/js/components/bhCronEmailReport/bhCronEmailReport.js
```diff
@@ -38,13 +39,19 @@ function bhCronEmailReportController(CronEmailReports, Notify, Session) {
...
-    };
-    $ctrl.isEmailFeatureEnabled = Session.enterprise.settings.enable_auto_email_report;
-    load($ctrl.reportId);
+    loadReportDetails($ctrl.reportKey)
+      .then(([report]) => {
+        $ctrl.cron = {
+          report_id : report.id,
+          has_dynamic_dates : 0,
+        };
+        $ctrl.isEmailFeatureEnabled = Session.enterprise.settings.enable_auto_email_report;
```
jniles, line 48: Since `$ctrl.isEmailFeatureEnabled` isn't asynchronous, let's remove it from the `loadReportDetails` callback.   We can put it as the first line in the `init()` function.

jniles (approved):
@mbayopanda, this is a good initiative.  I've made a few comments, but none of them worth blocking your work to standardize databases.  We can make the changes eventually if at all.

bors r+

PR #3918, opened 2019-10-07
Update imck and vanga units according master

mbayopanda:
This PR rewrite units of imck and vanga according the master branch


jniles:
@mbayopanda, can you break this into two files?  One for IMCK and one for Vanga?

I think it makes the most sense as we won't want to run the Vanga one at IMCK and vice versa.

jniles (approved):
These scripts seem to work for me.

bors r+

PR #3920, opened 2019-10-08
chore: clean up date logic in payroll config

jniles:
This PR creates an extra file "datelogic.js" in the payroll controller to deal with date logic.  The goal is to be able to unit test some of the date logic in the payroll controller.

Two functions are exposed:
 1. `isDateOnWeekend(date, weekendDateArray)` which returns a boolean value.  It is true if the date is on a weekend according to the weekendDateArray indices.
 2. `createDateRange()` which is a copy of the `getDates()` function that used to be in the `manageConfigurationData()` function.  I've moved it so I can write unit tests for it.

I've written a few simple unit tests, but we can get more complicated as more corner cases are discovered.

### DIFF
--- a/server/controllers/payroll/multiplePayroll/datelogic.js
+++ b/server/controllers/payroll/multiplePayroll/datelogic.js
@@ -9,14 +9,14 @@ function isDateOnWeekend(date, weekendDayIndex) {
 }
 
 /**
- * @function getDates
+ * @function createDateRange
  *
  * @description
  * Creates an array of dates, with each element being a date within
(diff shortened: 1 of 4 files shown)


jniles:

lomamech (commented):
I have just tested the functionality by configuring for periods of pay, several days of holidays, holidays during the weekend, employees with periods of vacancies between two months, but also several  vacancies in one month and all things work very well

lomamech:
All unit tests pass without problem,
image

PR #3925, opened 2019-10-09
feat(bhSearchComponent): implement search component

### LINKED ISSUE #3882: Create a `bhSearchButton` component
We have a ton of buttons that look like this:
image

The HTML code is always:
```html
 <button
          ng-click="SomeController.search()"
          data-method="search"
          class="btn btn-default">
          <span class="fa fa-search"></span> <span class="hidden-xs" translate>FORM.BUTTONS.SEARCH</span>
</button>
```

Since it is always the same, let's make a component to replace all instances of the search button.  The component will have a single method `&onClick` that will fire a callback when the button is clicked.

This will ensure consistency across all modules.

jniles:
Implements the search component to standardize the user interface of all search buttons.

Closes #3882

### DIFF
--- a/client/src/js/components/bhSearchButton.js
+++ b/client/src/js/components/bhSearchButton.js
@@ -1,0 +1,25 @@
+const template = `
+<div class="toolbar-item">
+  <button
+    ng-click="$ctrl.onClick()"
+    data-method="search"
+    class="btn btn-default">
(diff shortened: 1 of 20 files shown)


mbayopanda (approved):
Great job Jon,

I have just noticed that the `Inventory list` doesn't have change on its search button.

mbayopanda:
bors try

jniles:
bors try

Issue #3793, opened 2019-07-10
(Stock Loss) Restrict one lot by line in the grid


mbayopanda:
This issue was tackled with #3189 but the problem is : 
If the user fill the grid with many inventory without specifying lots : 

![image](https://user-images.githubusercontent.com/5445251/60947546-b8ea8380-a2e8-11e9-966d-67b7c6752af9.png)

The user can set the same lot on each line of the inventory

![image](https://user-images.githubusercontent.com/5445251/60947624-e33c4100-a2e8-11e9-9ef2-566b970406ae.png)

We have not to allow movement like that because it a source of many errors, so we have to block the user when he or she try to submit the movement by showing him a message or highlighting concerned rows.

jniles:
@mbayopanda did this get fixed?

mbayopanda:
@jniles yes on the stock exit module, not yet in the stock adjustment module

jeremielodi:
I think this issue can be closed

mbayopanda:
Fixed

PR #3926, opened 2019-10-09
Catch duplicated lots in stock adjustment

mbayopanda:
This PR fixes the issue #3793

### DIFF
--- a/client/src/modules/stock/StockForm.service.js
+++ b/client/src/modules/stock/StockForm.service.js
@@ -127,19 +127,18 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout, bhConst
   };
 
   /**
-   * @method catchDuplicatedLots
+   * @method hasDuplicatedLots
    *
    * @description
    * this method catch duplicated row and emit notification on the row
    */
-  StockForm.prototype.catchDuplicatedLots = function catchDuplicatedLots(_store) {
+  StockForm.prototype.hasDuplicatedLots = function hasDuplicatedLots() {
     const ROW_ERROR_FLAG = bhConstants.grid;
-    let selectedLots;
     let doublonDetectedLine;
 
-    if (hasDuplicatedLots(_store)) {
+    if (findDuplicatedLots(this.store)) {
       // notify on the concerned row
-      errorLineHighlight(doublonDetectedLine, _store);
+      errorLineHighlight(doublonDetectedLine, this.store);
       return true;
     }
 
@@ -155,16 +154,16 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout, bhConst
 
     // update the list of selected lots
     function refreshSelectedLotsList(store) {
-      selectedLots = store.data
+      return store.data
         .filter(item => item.lot && item.lot.uuid)
         .map(item => item.lot.uuid);
     }
 
     // detect the presence of duplicated lots
-    function hasDuplicatedLots(store) {
-      refreshSelectedLotsList(store);
-
+    function findDuplicatedLots(store) {
       let doubleIndex;
+      const selectedLots = refreshSelectedLotsList(store);
+
       const doublonDetected = selectedLots.some((lot, idx) => {
         const hasDoubles = selectedLots.lastIndexOf(lot) !== idx;
         if (hasDoubles) { doubleIndex = idx; }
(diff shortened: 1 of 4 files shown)


client/src/modules/stock/StockForm.service.js
```diff
@@ -125,5 +126,57 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
...
   };
 
+  /**
+   * @method catchDuplicatedLots
+   *
+   * @description
+   * this method catch duplicated row and emit notification on the row
+   */
+  StockForm.prototype.catchDuplicatedLots = function catchDuplicatedLots(_store) {
+    const ROW_ERROR_FLAG = bhConstants.grid;
```
jniles: I think this should be:
```js
const { ROW_ERROR_FLAG } = bhConstants.grid;
```

client/src/modules/stock/StockForm.service.js
```diff
@@ -125,5 +126,57 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
...
+    // update the list of selected lots
+    function refreshSelectedLotsList(store) {
+      selectedLots = store.data
+        .filter(item => item.lot && item.lot.uuid)
+        .map(item => item.lot.uuid);
+    }
+
+    // detect the presence of duplicated lots
+    function hasDuplicatedLots(store) {
+      refreshSelectedLotsList(store);
```
jniles: Instead of making a variable `selectedLots` in the upper scope, let's just make it a `const` in this scope   Like this:
```js

    function refreshSelectedLotsList(store) {
      return store.data
        .filter(item => item.lot && item.lot.uuid)
        .map(item => item.lot.uuid);
    }

 function hasDuplicatedLots(store) {
      const selectedLots = refreshSelectedLotsList(store);
      // etc...
 }
```

client/src/modules/stock/adjustment/adjustment.js
```diff
@@ -26,6 +26,7 @@ function StockAdjustmentController(
   // global variables
   vm.Stock = new StockForm('StockAdjustment');
   vm.movement = {};
+  vm.ROW_ERROR_FLAG = bhConstants.grid;
```
jniles: Same comment - this should be `vm.ROW_ERROR_FLAG = bhConstants.grid.ROW_ERROR_FLAG`.

client/src/modules/stock/exit/exit.js
```diff
@@ -470,17 +454,6 @@ function StockExitController(
       .finally(() => reinit(form));
   }
 
-  function errorLineHighlight(rowIdx) {
```
jniles, line 459: Excellent work refactoring this into a shared function. :+1:

client/src/modules/stock/StockForm.service.js
```diff
@@ -125,5 +126,57 @@ function StockFormService(StockItem, Store, AppCache, Session, $timeout) {
     return Object.keys(this.cache).length > 0;
   };
 
+  /**
+   * @method catchDuplicatedLots
+   *
+   * @description
+   * this method catch duplicated row and emit notification on the row
+   */
+  StockForm.prototype.catchDuplicatedLots = function catchDuplicatedLots(_store) {
```
jniles: Since this returns a boolean, can we call it `hasDuplicatedLots`?

jniles: Also, I think you can just use `this.store` instead of passing it in, no?

jniles (approved):
A few comments here - the spirit of the PR is beautiful (refactoring, refactoring, refactoring), but I think we can make the implementation nicer.  Let me know what your thoughts are.

client/src/modules/stock/StockForm.service.js
mbayopanda: I have tried with `this.store` it doesn't work curiously

client/src/modules/stock/StockForm.service.js
mbayopanda: Oops... it seems like the highlight works only with this `const ROW_ERROR_FLAG = bhConstants.grid;` syntax.

jniles: Did you also make this change in the controller?

See: https://github.com/IMA-WorldHealth/bhima/pull/3926/files#r333052727

client/src/modules/stock/StockForm.service.js
mbayopanda: It works 😆

jniles:
Looks good to me and tests pass.  Well done!

bors r+

PR #3930, opened 2019-10-14
feat(install) add password meter

### LINKED ISSUE #3922: Add password strength meter to installation master password
Here is the installation screen as presented in bhima:

image
_Fig 1: Configure user and enterprise_

It would be nice to have the `bhStrengthMeter` on this installation screen.

jeremielodi:
closes #3922

### DIFF
--- a/client/src/modules/users/UserEditPasswordModal.html
+++ b/client/src/modules/users/UserEditPasswordModal.html
@@ -14,7 +14,7 @@
         class="form-control"
         type="password"
         bh-password-meter
-        type="password" required>
+        required>
 
         <bh-strength-meter password="PasswordForm.password.$viewValue"></bh-strength-meter>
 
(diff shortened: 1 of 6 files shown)


client/src/modules/users/UserEditPasswordModal.html
```diff
@@ -14,7 +14,7 @@
         class="form-control"
         type="password"
         bh-password-meter
-        type="password" required>
+        required>
```
jniles, line 17: 💃

test/client-unit/services/PasswordMeterService.spec.js
```diff
@@ -75,12 +74,6 @@ describe('PasswordMeterService', () => {
     expect(validate).to.equal(true);
   });
 
-  it('#validate() should return true if the session is not set', () => {
```
jniles, line 78: Instead of deleting tests to make the test pass, can you alter this test to show the new behavior?

jniles (commented):
The logic looks fine to me.  Could you update the tests to show how the logic changed?

jniles (approved):
Looks fine to me.

bors r+

PR #3936, opened 2019-10-15
fix(stock): do not submit adjustments on btn

### LINKED ISSUE #3935: Bug: Stock Adjustment submits the form when you select "increase" or "decrease"
The form submit validation triggers when you click the "increase" or "decrease" buttons.  You can see it in action below.

dV3RovDGZi
_Fig 1: Form Validation_

There shouldn't be any red text when the user clicks those buttons.

jniles:
Adds type="button" to button inputs to avoid them defaulting to type="submit" and triggering validation.

Closes #3935.


mbayopanda (approved):
LGTM 👍 

bors r+

PR #3940, opened 2019-10-15
Add bh-change-depot component

mbayopanda:
This PR add the bhChangeDepot component and refactor : 
- Stock Adjustment
- Stock Entry
- Stock Import
- Stock Exit


jniles:
That was fast!  Let's see what the tests do, then I'll do a review. 💃

client/src/js/components/bhChangeDepot/bhChangeDepot.html
```diff
@@ -0,0 +1,12 @@
+<div uib-dropdown dropdown-append-to-body data-action="open-tools">
```
jniles: Would it be possible to make the component just the `<a href>` part?  That way, we can add more items in the future (like requisitions) to the dropdown menu.

Alternatively, you could include a `transclude` slot so that we can add items to the menu by putting them as sub-elements of the component.
```html
<bh-change-depot>
  <a href="" new-component></a>
</bh-change-depot>
```

Personally, I prefer the first one, but it is up to you.

client/src/js/components/bhChangeDepot/bhChangeDepot.js
```diff
@@ -0,0 +1,44 @@
...
+        $ctrl.onSelect({ depot });
+      })
+      .catch(Notify.handleError);
+  }
+
+  function changeDepot() {
+    // if requirement is true the modal cannot be canceled
+    const requirement = !cache.depotUuid;
+
+    return Depots.openSelectionModal($ctrl.currentDepot, requirement)
```
jniles: Why not use `cache.depotUuid`?  Then we can remove the required of passing in the `currentDepot`.

client/src/js/components/bhChangeDepot/bhChangeDepot.js
```diff
@@ -0,0 +1,44 @@
...
+      .then(depot => {
+        cache.depotUuid = depot.uuid;
+        $ctrl.onSelect({ depot });
```
jniles: 👍

jniles (commented):
@mbayopanda that was super quick!  I like the way this is heading.

I've proposed a few ideas to remove the `current-depot` attribute and make it so we can add other things to the menu in the future.  Let me know what you think!

mbayopanda:
@jniles i have switched to `Directives` since with `Components` we cannot replace the component tag name in case we want to keep certain order of tags for css styles (dropdown style with Bootstrap)

jniles (approved):
This is excellent work.  Behaves exactly as we want.

The code is good too.  Well done!

bors r+

jniles:
bors r+

For some reason, travis is really struggling today.

mbayopanda:
bors try

jniles:
bors try

jniles:
bors retry

PR #3945, opened 2019-10-17
feat(price list) import items from csv file

### LINKED ISSUE #3893: Import Price List from CSV
We need to have an import functionality to allow users to quickly set prices in the entire application.  This will ensure they can quickly and easily change prices without having to do a ton of clicking.  This feature is similar to both the "import accounts from csv" functionality (https://github.com/IMA-WorldHealth/bhima/pull/3061) and "import stock from csv" functionality (https://github.com/IMA-WorldHealth/bhima/issues/2571) that exist in the system.

The basic workflow would be:

1. The user goes to the price list page
2. From the menu in the top right, they are able to download a "price list change template" in CSV format.
3. The template has columns for uuid, code, inventory group name, inventory label, and an empty "price" column.  All are filled out except for the "price" column.
4. The user fills out the price column in the enterprise currency.
5. They can then "Import new price list from template", which gives them the ability to upload that CSV file.
6. The CSV file creates a new price list for the user based on the prices entered into the CSV template.

jeremielodi:
closes  #3893

### DIFF
--- a/client/src/modules/prices/prices.service.js
+++ b/client/src/modules/prices/prices.service.js
@@ -1,7 +1,7 @@
 angular.module('bhima.services')
   .service('PriceListService', PriceListService);
 
-PriceListService.$inject = ['PrototypeApiService'];
+PriceListService.$inject = ['PrototypeApiService', '$translate'];
 
 /**
  * @class PriceListService
@@ -10,7 +10,7 @@ PriceListService.$inject = ['PrototypeApiService'];
  * @description
  * Encapsulates common requests to the /prices/ URL.
  */
-function PriceListService(Api) {
+function PriceListService(Api, $translate) {
   const baseUrl = '/prices/';
   const service = new Api(baseUrl);
 
@@ -83,7 +83,7 @@ function PriceListService(Api) {
     const url = service.url.concat('download/template');
     return service.$http.get(url)
       .then(response => {
-        return service.util.download(response, 'Iventory item Template', 'csv');
+        return service.util.download(response, $translate.instant('PRICE_LIST.ITEM_TEMPLATE'), 'csv');
       });
   }
   return service;
--- a/client/src/modules/prices/prices.service.js
+++ b/client/src/modules/prices/prices.service.js
@@ -20,7 +20,7 @@ function PriceListService(Api) {
   service.details = details;
   service.deleteItem = deleteItem;
   service.download = download;
-
+  service.downloadTemplate = downloadTemplate;
   /**
    * @method create
    *
@@ -77,6 +77,14 @@ function PriceListService(Api) {
     const url = service.url.concat('download/list');
     return service.$http.get(url, params)
       .then(service.util.unwrapHttpResponse);
+  }
+
+  function downloadTemplate() {
+    const url = service.url.concat('download/template');
+    return service.$http.get(url)
+      .then(response => {
+        return service.util.download(response, 'Iventory item Template', 'csv');
+      });
   }
   return service;
 }
--- a/server/controllers/finance/priceList.js
+++ b/server/controllers/finance/priceList.js
@@ -13,11 +13,14 @@
  * DELETE /prices/:uuid
  */
 const path = require('path');
+const _ = require('lodash');
 const db = require('../../lib/db');
 const { uuid } = require('../../lib/util');
 
 const BadRequest = require('../../lib/errors/BadRequest');
 const util = require('../../lib/util');
+const csv = require('../../lib/renderers/csv');
+const ReportManager = require('../../lib/ReportManager');
 
 exports.lookup = lookup;
 /**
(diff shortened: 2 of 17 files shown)


server/controllers/finance/priceList.js
```diff
@@ -229,6 +236,92 @@ exports.createItem = function createItem(req, res, next) {
...
+};
+
+function removeEmptyLines(data) {
+  data.forEach((r, index) => {
+    if (!(r.code && r.price)) delete data[index];
+  });
+}
+
+function hasValidDataFormat(data) {
+  const invalids = data.filter(r => {
```
jniles: A more semantic way of doing this is using [Array.prototype.every()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every).  It returns a boolean if the condition is met for every element in an array.  In your case, it would be something like:
```js
return data.every(row => (row.code && row.price > 0));
```

jeremielodi:
@mbayopanda, you can give me a review to this PR once you have time. 
thanks

server/controllers/finance/priceList.js
```diff
@@ -229,6 +236,92 @@ exports.createItem = function createItem(req, res, next) {
...
+    res.sendStatus(200);
+  } catch (ex) {
+    next(ex);
+  }
+
+};
+
+function removeEmptyLines(data) {
+  data.forEach((r, index) => {
+    if (!(r.code && r.price)) delete data[index];
```
jniles: Did you know that when you delete an array element, the length isn't affected?  Try it:
```js
const array = [1, 2, 3, 4, 5];
console.log(array.length); // => 5
delete array[2];
console.log(array.length); // => 5
```

The `delete` operator also has some pretty strange behavior for prototype versus non prototype objects, strict vs non-strict mode, etc.  You can read all about it in [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete).

For arrays, I always prefer to simply `filter()`.  Like so:
```js
return data.filter(row => (row.code && r.price));
```

server/controllers/finance/priceList.js
```diff
@@ -229,6 +236,92 @@ exports.createItem = function createItem(req, res, next) {
...
+    const file = path.join(__dirname, '../../resources/templates/import-inventory-item-template.csv');
+    res.download(file);
+  } catch (error) {
+    next(error);
+  }
+};
+
+exports.downloadFilledTemplate = async (req, res, next) => {
+  try {
+    const pdfOptions = {
```
jniles, line 256: This is only available as CSV, right?  Or can we download it in other formats?

If only CSV, you can probably remove most of this code (for example, `orientation` doesn't apply to CSVs, just PDFs).

client/src/modules/prices/prices.service.js
```diff
@@ -78,5 +78,13 @@ function PriceListService(Api) {
     return service.$http.get(url, params)
       .then(service.util.unwrapHttpResponse);
   }
+
+  function downloadTemplate() {
+    const url = service.url.concat('download/template');
+    return service.$http.get(url)
+      .then(response => {
+        return service.util.download(response, 'Iventory item Template', 'csv');
```
jniles: This should be spelled 'Inventory Item Template'.

Can you use `$translate` to provide a custom filename for the user?

server/models/procedures/inventory.sql
```diff
@@ -74,4 +74,30 @@ BEGIN
...
+  DECLARE _inventory_uuid BINARY(16);
+  DECLARE isIventory tinyint(5);
+   DECLARE inventoryLabel VARCHAR(100);
+  
+  SELECT uuid, text,  count(uuid) 
+  INTO _inventory_uuid, inventoryLabel, isIventory
+  FROM inventory
+  WHERE code = _inventory_code;
+
+  IF isIventory = 1 THEN
```
jniles: This should be `isInventory`.

server/models/migrations/next/migrate.sql
```diff
@@ -5,3 +5,38 @@ ALTER TABLE `stock_consumption` MODIFY `period_id` MEDIUMINT(8) UNSIGNED NOT NUL
...
+ * DATABASE CHANGES FOR VERSION 1.6.0 TO 1.7.0 
+ */
+
+/*
+* Pricelist importation
+  by Jeremielodi
+  2019-10-16
+*/
+
+DROP PROCEDURE IF EXISTS importPriceListItem;
```
jniles, line 18: No need to add procedures to the migration script.  We rebuild all procedures when we do an upgrade anyway.

jniles (commented):
@jeremielodi, I know you asked @mbayopanda for a review, but I was already working on one, so I'm putting it here!

Client-side looks good!  No comments there.

I've made a few comments on the server-side code.  I haven't tested it yet, but I trust it works and appreciate the end to end tests you've written.

PR #3951, opened 2019-10-21
refactor(bhFiscalYearPeriodSelect): replace custom logic with two components

jniles:
Refactors the bhFiscalYearPeriodSelect to just be a combination of two other components.  Much less to maintain and test.

Note that tests will probably need to be updated.


mbayopanda (approved):
LGTM 👍

PR #3954, opened 2019-10-23
chore(ci): bump testing versions to latest LTS

jniles:
Node v12 just went into LTS.  I've changed our testing to use the latest LTS.


jniles:
Tests pass.  Wonderful.

PR #3957, opened 2019-10-23
Feat(trial balance) set credit value in red

### LINKED ISSUE #3907: Render credits as red values on Trial Balance
The trial balance doesn't render credits as red like most other places in the application:
image
_Fig 1: Credit balances as negative_

We should fix this to make the credit balances look like other places in the application.

jeremielodi:
closes #3907

### DIFF
--- a/client/src/modules/accounts/accounts.service.js
+++ b/client/src/modules/accounts/accounts.service.js
@@ -28,6 +28,7 @@ function AccountService(Api, bhConstants, HttpCache) {
 
   service.flatten = flatten;
   service.order = order;
+  service.redCreditCell = redCreditCell;
 
   /**
    * @method getOpeningBalance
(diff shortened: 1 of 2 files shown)


jniles:
Can you include a screen shot?  It's hard to tell what it looks like from just the code.

jeremielodi:
credit value

jniles:
@jeremielodi, there shouldn't be a "-" sign in the negative numbers, since the parens replaces that.  See the compte de resultat report, for example:
image

Could you remove the minus?

You could also try making these changes directly in the $currency filter - bhima has a fork of it that we maintained to support FC rendering. https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/filters/currency.js

kwilu:
Hello @jeremielodi, it seems like you have referenced #3907 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #3907.`.

```
Author: kwilu

    Fixes #51.
```

jniles (approved):
This solves the issue.  I think we will have better results in the long run by using the currency filter, since we won't need to rewrite a bunch of code.   However, this code is good and the PR solves the trial balance issue.

bors r+

PR #3959, opened 2019-10-24
fix: condition appcache on hostname

jniles:
Allows you to have multiple different BHIMA tabs open on different hosts without sharing the same appcache content.


jniles:
bors try

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3960, opened 2019-10-24
docs(stock): enrich stock management text

jniles:
This commit enriches the stock management documentation to clear up a misunderstanding reported by our users.


mbayopanda (commented):
:+1; So helpful

bors r+

PR #3961, opened 2019-10-25
fix(purchases): render enterprise currency

### LINKED ISSUE #3958: Purchase Order Receipt shows $USD, not the enterprise currency
The receipt for Purchase Orders doesn't take into account the enterprise currency.  It always shows $USD.  See below:

w633Q43T5u

jniles:
This commit fixes the purchase receipt to render the enterprise currency consitently.

Closes #3958


lomamech (commented):
I just checked the functionality with the Vanga database, and the problem is solved successfully
image

PR #3970, opened 2019-10-30
Fix(Optimal Purchase Order)

### LINKED ISSUE #3964: Fixed the optimal purchase order option
Fixed the optimal purchase order option
Due to numerous system updates, the optimal purchase order option no longer works, and the E2E test that verifies this functionality must be restored

lomamech:
- Correction of the error during the release of the optimal purchase
  orders
- Implement of the formatOptimalPurchase service allowing  for all
  existing repositories, Inventories that have reached a point of order,
  then calculates the total quantity ordered for each Inventory
- Implement E2E for Optimal Purchase Order

closes #3964

### DIFF
--- a/client/src/modules/purchases/create/createUpdate.js
+++ b/client/src/modules/purchases/create/createUpdate.js
@@ -3,11 +3,12 @@ angular.module('bhima.controllers')
 
 PurchaseOrderController.$inject = [
   'PurchaseOrderService', 'PurchaseOrderForm', 'NotifyService',
-  'SessionService', 'util', 'ReceiptModal', 'bhConstants', 'StockService',
+  'SessionService', 'util', 'ReceiptModal', 'bhConstants', 'StockService', 'InventoryService',
 ];
 
 
-function PurchaseOrderController(Purchases, PurchaseOrder, Notify, Session, util, Receipts, bhConstants, Stock) {
+function PurchaseOrderController(Purchases, PurchaseOrder, Notify,
+  Session, util, Receipts, bhConstants, Stock, Inventory) {
   const vm = this;
 
   // create a new purchase order form
@@ -82,6 +83,12 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Notify, Session, util
     data : vm.order.store.data,
   };
 
+  Inventory.read(null, { detailed : 1, locked : 0 })
+    .then((inventories) => {
+      vm.inventories = inventories;
+    })
+    .catch(Notify.handleError);
+
   // this function will be called whenever items change in the grid.
   function handleUIGridChange() {
     vm.order.digest();
@@ -133,7 +140,6 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Notify, Session, util
 
     return Purchases.create(order)
       .then((res) => {
-
         // open the receipt modal
         Receipts.purchase(res.uuid, true);
 
@@ -164,26 +170,36 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Notify, Session, util
 
   function optimalPurchase() {
     vm.optimalPO = true;
+    const filters = {
+      includeEmptyLot : 0,
+      period : 'allTime',
+      require_po : 1,
+    };
 
-    Stock.inventories.read(null, { require_po : 1 })
+    Stock.inventories.read(null, filters)
       .then(rows => {
         if (!rows.length) {
           return Notify.warn('FORM.INFO.NO_INVENTORY_PO');
         }
 
+        const optimalPurchaseData = Purchases.formatOptimalPurchase(vm.inventories, rows);
+
         // adding items.length line in the Order store, which will be reflected to the grid
-        if (rows.length > 1) {
-          vm.order.addItem(rows.length);
+        if (optimalPurchaseData > 1) {
+          vm.order.addItem(optimalPurchaseData.length);
         }
 
         vm.order.store.data.forEach((item, index) => {
-          item.code = rows[index].code;
-          item.inventory_uuid = rows[index].inventory_uuid;
-          item.description = rows[index].text;
-          item.quantity = rows[index].S_Q;
-          item.unit_price = 0;
-          item.unit = rows[index].unit_type;
-          item._initialised = true;
+          item.inventory_uuid = optimalPurchaseData[index].uuid;
+          item.code = optimalPurchaseData[index].code;
+          item.description = optimalPurchaseData[index].label;
+          item.quantity = optimalPurchaseData[index].quantity;
+          item.unit_price = optimalPurchaseData[index].unit_price;
+          item.unit = optimalPurchaseData[index].unit;
+          item._initialised = optimalPurchaseData[index]._initialised;
+          item._hasValidAccounts = optimalPurchaseData[index]._hasValidAccounts;
+          item._invalid = optimalPurchaseData[index]._invalid;
+          item._valid = optimalPurchaseData[index]._valid;
         });
 
         return 0;
--- a/client/src/modules/purchases/create/createUpdate.js
+++ b/client/src/modules/purchases/create/createUpdate.js
@@ -182,24 +182,12 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Notify,
           return Notify.warn('FORM.INFO.NO_INVENTORY_PO');
         }
 
-        const optimalPurchaseData = Purchases.formatOptimalPurchase(vm.inventories, rows);
+        const optimalPurchaseData = PurchaseOrder.formatOptimalPurchase(vm.inventories, rows);
 
-        // adding items.length line in the Order store, which will be reflected to the grid
-        if (optimalPurchaseData > 1) {
-          vm.order.addItem(optimalPurchaseData.length);
-        }
-
-        vm.order.store.data.forEach((item, index) => {
-          item.inventory_uuid = optimalPurchaseData[index].uuid;
-          item.code = optimalPurchaseData[index].code;
-          item.description = optimalPurchaseData[index].label;
-          item.quantity = optimalPurchaseData[index].quantity;
-          item.unit_price = optimalPurchaseData[index].unit_price;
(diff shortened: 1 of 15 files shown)


jniles:
bors try

jniles:
bors try

test/integration/stock/stock.js
```diff
@@ -152,8 +154,9 @@ describe('(/stock/) The Stock HTTP API', () => {
     returns remaining quantity of QUININE-A in Depot Principal (80pcs)`,
     () => agent.get(`/stock/lots/depots?lot_uuid=${shared.lotQuinineUuid}&depot_uuid=${shared.depotPrincipalUuid}`)
       .then((res) => {
+
         helpers.api.listed(res, 1);
-        expect(res.body[0].quantity).to.be.equal(100 - 20);
+        expect(res.body[0].quantity).to.be.equal(100 - 100);
```
jniles: Isn't this just 0?

lomamech:
@jniles  @mbayopanda  This PR is ready to merged

lomamech:
@mbayopanda  this PR need review

jniles:
I'll grab this one.

client/src/modules/purchases/purchases.service.js
```diff
@@ -157,5 +158,37 @@ function PurchaseOrderService(
...
+   * then calculates the overall quantity ordered for each Inventory
+   *
+   * @param {*} inventories - This is the list of all inventories
+   * @param {*} stock - it is the data of the inventories having reached their point of order
+   */
+  function formatOptimalPurchase(inventories, stock) {
+    inventories.forEach(inventory => {
+      inventory.unit_price = 0;
+      inventory._initialised = true;
+      inventory._hasValidAccounts = inventory.cogs_account;
```
jniles: It looks like this is supposed to be a `Boolean` - https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/purchases/create/PurchaseOrderItemService.js#L88

client/src/modules/purchases/purchases.service.js
```diff
@@ -157,5 +158,37 @@ function PurchaseOrderService(
...
+   * @method formatOptimalPurchase
+   *
+   * This Service analyzes for all existing repositories, Inventories that have reached a point of order,
+   * then calculates the overall quantity ordered for each Inventory
+   *
+   * @param {*} inventories - This is the list of all inventories
+   * @param {*} stock - it is the data of the inventories having reached their point of order
+   */
+  function formatOptimalPurchase(inventories, stock) {
+    inventories.forEach(inventory => {
```
jniles: This looks a lot more complicated than it needs to be - a lot of these properties and calculations are done already by the [`PurchaseOrderForm`](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/purchases/create/PurchaseOrderForm.js) and the [`PurchaseOrderItemService`](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/purchases/create/PurchaseOrderItemService.js).

Could you try using those?  I would proceed like this:

   1. Start by moving this function into the `PurchaseOrderFormService`, since it will only be used on the create page.
   2. Use the `PurchaseOrderItemService` for each item.  That would look like this, I think:
```js
inventories.forEach(inventory => {
  const row = new PurchaseOrderItem(inventory);
  // quantity is by default 1, so set it it 0
  row.quantity = 0;

  stock.forEach(item => {
    if (row.inventory_uuid === item.inventory_uuid) {
      row.quantity += item.S_Q;
    }
  });
```
   3. Then finally, add the custom logic to sort:
```js
     const optimized = rows.filter(inventory => inventory.quantity > 0); 
     const optimizedSorted = optimized.sort((a, b) => Number(b.quantity) - Number(a.quantity));
     return optimizedSorted;
```

server/controllers/finance/purchases.js
```diff
@@ -218,11 +218,13 @@ function create(req, res, next) {
 
           purchaseInterval = (((row.purchase_interval * (row.num_purchase - 1)) + durationMonth) / row.num_purchase);
         }
-
-        transactionWrapper.addQuery(
-          'UPDATE inventory SET purchase_interval = ?, last_purchase = ?, num_purchase = ?  WHERE uuid = ?',
-          [purchaseInterval, datePurchase, numPurchase, db.bid(row.inventory_uuid)]
-        );
+        // Just to prevent cases where the order interval is less than a month
```
jniles: What happens if the order interval is less than a month?  Why is that a bad thing?

server/controllers/stock/core.js
```diff
@@ -469,7 +469,7 @@ function getInventoryQuantityAndConsumption(params) {
       d.text AS depot_text, l.unit_cost, l.expiration_date,
       ROUND(DATEDIFF(l.expiration_date, CURRENT_DATE()) / 30.5) AS lifetime,
       BUID(l.inventory_uuid) AS inventory_uuid, BUID(l.origin_uuid) AS origin_uuid,
-      l.entry_date, i.code, i.text, BUID(m.depot_uuid) AS depot_uuid,
+      l.entry_date, BUID(i.uuid) AS inventory_uuid, i.code, i.text, BUID(m.depot_uuid) AS depot_uuid,
```
jniles, line 480: It looks like this property already exists in this SQL statement on the line directly above it:  `BUID(l.inventory_uuid) AS inventory_uuid`.  Did that not work?

client/src/modules/purchases/create/createUpdate.js
```diff
@@ -164,26 +170,36 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Notify, Session, util
 
   function optimalPurchase() {
     vm.optimalPO = true;
+    const filters = {
```
jniles, line 173: Nice :+1:

client/src/modules/purchases/create/createUpdate.js
```diff
@@ -164,26 +170,36 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Notify, Session, util
...
 
         vm.order.store.data.forEach((item, index) => {
-          item.code = rows[index].code;
-          item.inventory_uuid = rows[index].inventory_uuid;
-          item.description = rows[index].text;
-          item.quantity = rows[index].S_Q;
-          item.unit_price = 0;
-          item.unit = rows[index].unit_type;
-          item._initialised = true;
+          item.inventory_uuid = optimalPurchaseData[index].uuid;
```
jniles: It looks like, if there is data in the grid, it will replace the old data with new data.

I propose that you just go ahead and clear the grid with `vm.order.clear()` and then add the appropriate number of elements to the grid.

client/src/modules/purchases/purchases.service.js
jniles: If you do all this, you won't need to do a `forEach` loop in the `optimalPurchase()` function.  You can just push these into the stack.

```js
const optimalPurchaseData = Purchases.formatOptimalPurchase(vm.inventories, rows); 

// clear the grid as suggested above
vm.order.clear();

optimalPurchaseData.forEach(item => {
  vm.order.store.post(item);
});

vm.order.digest();
```

Note that if you make this a function of `PurchaserOrderForm`, instead of `vm.order`, you can just use `this`.

jniles (commented):
@lomamech, great work getting the end to end tests back up and runnning!

I've made a few comments here.  They are detailed, but if there is anything confusing, let me know and I'll clarify.

client/src/modules/purchases/purchases.service.js
lomamech: I initialized this property just because when you place a purchase order, the property _hasValidAccounts contains the identifier of an account,  image

jniles: Oh!  Than that is a bug.  It should return true if both a stock account and a cost of goods sold account exist.

lomamech:
@jniles  @mbayopanda  I finished with the code update,

kwilu:
Heads up @lomamech, we just merged some commits that conflict with the changes your made in this pull request! You can review this repository's [recent commits](https://github.com/IMA-WorldHealth/bhima/commits/master) to see where the conflicts occur. Please rebase your feature branch against the `upstream/master` branch and [resolve](https://zulip.readthedocs.io/en/latest/git/troubleshooting.html#recover-from-a-git-rebase-failure) your pull request's merge conflicts accordingly.

server/controllers/stock/core.js
jniles: What is the difference between `BUID(l.inventory_uuid) AS inventory_uuid` in the line above and `BUID(i.uuid) AS inventory_uuid` in this line? 

I worry we've changed the behavior of this SQL and that will have negative impacts somewhere.

server/controllers/stock/core.js
```diff
@@ -356,8 +357,14 @@ function stockManagementProcess(inventories, inventoryDelay = 1, purchaseInterva
       inventory.status = '';
     }
 
+    // Round
+    inventory.S_SEC = util.roundDecimal(inventory.S_SEC, 2);
```
jniles, line 361: @mbayopanda can you take a look at these? You know better how they should work.

jniles (approved):
Code looks fine to me.  I'd like @mbayopanda to look at the logic of the ROUND() calls, but otherwise it seems fine to me.

server/controllers/finance/purchases.js
```diff
@@ -386,7 +391,7 @@ function purchaseStatus(req, res, next) {
 
   const sqlPurchaseDelay = `
     SELECT
-      p.cost, CEIL(DATEDIFF(m.date, p.date)) AS delay
+      p.cost, ROUND(DATEDIFF(m.date, p.date), 2) AS delay
```
mbayopanda: Is it possible to use just `DATEDIFF(m.date, p.date)` since DATEDIFF returns number of days.

mbayopanda (approved):
LGTM :+1

lomamech:
@mbayopanda @jniles this PR is ready

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3973, opened 2019-10-31
refactor(Merge Weekend create and Confiuration)

### LINKED ISSUE #3783: Proposal: Merge weekend create and configuration modals
We currently have two modals to configure and create weekends.  I don't see any reason why we should need two modals.  You can see the separate versions below and my proposal following:
**Create Modal**
image

**Config Modal**
image

**Merged Proposal**
image

lomamech:
- Merge weekend create and configuration modals
- Remove not necessary Option and Test
- Update API in server side
- Updte Integration and E2E test

closes #3783


lomamech:
@jniles  @mbayopanda can you review this PR

client/src/modules/payroll/weekend_configuration/templates/action.tmpl.html
```diff
@@ -11,12 +11,6 @@
       </a>
     </li>
 
-    <li>
-      <a data-method="configure" ui-sref="configurationWeekend.config({id : row.entity.id})" href>
```
jniles, line 15: Can we remove the templates/controllers for this?  It seems like we can now delete whatever files make up the `configurationWeekend.config()` state.

server/config/routes.js
```diff
@@ -735,7 +735,6 @@ exports.configure = function configure(app) {
   app.post('/weekend_config', weekendConfig.create);
   app.put('/weekend_config/:id', weekendConfig.update);
   app.get('/weekend_config/:id/days', weekendConfig.listConfig);
-  app.post('/weekend_config/:id/days', weekendConfig.createConfig);
```
jniles, line 738: 👍

server/controllers/payroll/weekendConfig/index.js
```diff
@@ -52,10 +52,22 @@ function detail(req, res, next) {
 function create(req, res, next) {
   const sql = `INSERT INTO weekend_config SET ?`;
   const data = req.body;
+  const configuration = data.daysChecked;
+  let insertId;
+
+  delete data.daysChecked;
 
   db.exec(sql, [data])
```
jniles: Go ahead and make this function `async/await` so you don't have to use `let`.  Like this:
```js
const { insertId } = await db.exec(sql, [data]);
const dataConfigured = configuration.map(id => ([id, insertId]));
await db.exec('INSERT INTO .... VALUES ?', dataConfigured]);
res.status(201).jsoin({ id : insertId });
```

test/end-to-end/weekendConfig/weekend_config.spec.js
```diff
@@ -18,14 +18,6 @@ describe('Weekend Configuration Management', () => {
     await page.update(newWeekendConfigLabel, updateWeekendConfigLabel);
   });
 
-  it('successfully set week days in weekend configuration', async () => {
-    await page.setWeekendConfig(updateWeekendConfigLabel);
```
jniles, line 22: Is there somewhere that you can remove these functions?  I don't think they will ever be used again.

client/src/js/constants/bhConstants.js
```diff
@@ -139,13 +139,13 @@ function constantConfig() {
...
-      { id : 0, label : 'FORM.LABELS.WEEK_DAYS.SUNDAY' },
-      { id : 1, label : 'FORM.LABELS.WEEK_DAYS.MONDAY' },
-      { id : 2, label : 'FORM.LABELS.WEEK_DAYS.TUESDAY' },
-      { id : 3, label : 'FORM.LABELS.WEEK_DAYS.WEDNESDAY' },
-      { id : 4, label : 'FORM.LABELS.WEEK_DAYS.THURSDAY' },
-      { id : 5, label : 'FORM.LABELS.WEEK_DAYS.FRIDAY' },
-      { id : 6, label : 'FORM.LABELS.WEEK_DAYS.SATURDAY' },
+      { id : 0, label : 'FORM.LABELS.WEEK_DAYS.SUNDAY', checked : false },
+      { id : 1, label : 'FORM.LABELS.WEEK_DAYS.MONDAY', checked : false },
+      { id : 2, label : 'FORM.LABELS.WEEK_DAYS.TUESDAY', checked : false },
```
jniles, line 144: 👍

jniles (changes requested):
This is great!  It looks like there may be a lot of legacy code hanging around after this change.  Please remove all functions/controllers/templates that this change will cause to never be used again.

server/controllers/payroll/weekendConfig/index.js
```diff
@@ -49,24 +49,42 @@ function detail(req, res, next) {
 }
 
 // POST /WEEKEND_CONFIG
-function create(req, res, next) {
+async function create(req, res) {
```
jniles: This is better!  We still need to handle errors though.  To make sure these errors are handled, please add back in "next()' and wrap all your code in a try/catch block.  Like this:
```js
async function create(req, res, next) {
  try {
    // all your code
  } catch (e) {
    next(e);
  }
}
```

test/end-to-end/weekendConfig/weekend_config.page.js
```diff
@@ -25,28 +32,6 @@ class WeekendConfigPage {
     await row.dropdown().click();
     await row.edit().click();
     await FU.input('WeekendModalCtrl.weekend.label', newLabel);
-    await FU.modal.submit();
-    await notification.hasSuccess();
-  }
-
-  async setWeekendConfig(label) {
```
jniles, line 32: 👍

test/integration/weekEndConfig.js
```diff
@@ -77,37 +75,6 @@ describe('(/payroll/weekend_configuration) The /payroll/weekend_configuration  A
       .catch(helpers.handler);
   });
 
-  // INTEGRATION TEST FOR SETTING WEEKEND IN CONFIGURATION
-
-  it('POST /WEEKEND__CONFIG/:ID/DAYS should Set WeekEnds in Configuration', () => {
```
jniles, line 82: 👍

jniles (commented):
Awesome work.  Just one tiny try/catch and then I'll merge it.

lomamech:
@jniles I finished with the review

jniles (approved):

lomamech:
@jniles  I't done

jniles:
Excellent, responds to all my concerns.  Code looks good to me.

bors r+

PR #3979, opened 2019-11-04
Fix: add balances to entry exit cash report

### LINKED ISSUE #3727: Add balances to the footer of the entry/exit portion of the cash report
We've received this request from a client:

> Please put the sum of the entries and exits at the bottom of the cash report.

We need to make sure that all the values are in the same currency to be able to sum them.  If all values are in the same currency, we can simply sum the values and display the totals in the grid footer as normal.

If not, we should put a note on the bottom of the report that says something like "mixed currencies are used in this statement so totals are not computed."

Balances
_Fig 1: Where the balance totals should go_

jniles:
This PR adds the feature to show the balance at the foot of the cash report if all transactions within the range are in the same currency.  It also migrates the controller to async/await and corrects the column ordering that was previously incorrect.

Closes #3727


mbayopanda (approved):
LGTM 👍 

bors r+

PR #3980, opened 2019-11-04
fix(vouchers): fixes i18n on voucher item error

### LINKED ISSUE #3507: Voucher Error on negative values is not translated
The i18n key for negative values in the complex vouchers form is not translated.  It should read something like "Negative values are not allowed in a transaction."

negativevaluesnottranslated

jniles:
Fixes the i18n key for errors on negative amounts in the voucher interface.

Closes #3507.


jeremielodi (approved):
LGTM

PR #3981, opened 2019-11-04
fix(xls): export XLS files without currency symbol

### LINKED ISSUE #3764: Do not render currency in MS Excel reports
For most reports, if a report is rendered in MS Excel, we should not render the currency symbol.  This will allow users to easily export to MS Excel, and sum the values in the report.  Ideally, this would be an option during the "Save as MS Excel" option.

jniles:
Implements a feature asked for by our clients to remove the currency symbol from the HTML -> Excel render.  This allows users to sum the numbers are ordinary values rather than parse the currency symbol.

Closes #3764.

### DIFF
--- a/server/lib/renderers/html.js
+++ b/server/lib/renderers/html.js
@@ -16,6 +16,7 @@ const debug = require('debug')('renderers:html');
 const util = require('../util');
 const hbs = require('../template');
 const translateHelperFactory = require('../helpers/translate');
+const finance = require('../template/helpers/finance');
 
 const headers = {
   'Content-Type' : 'text/html',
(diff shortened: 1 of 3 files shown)


server/lib/renderers/html.js
```diff
@@ -39,14 +40,20 @@ function renderHTML(data, template, options = {}) {
...
   // factory code.
   const translate = translateHelperFactory(options.lang);
   hbs.helpers.translate = translate;
 
+  // NOTE(@jniles) - if the file is an XLS file that is being converted to
+  // excel, we don't do anything for the currency besides returning it.
+  if (options.skipCurrencyRendering) {
+    hbs.helpers.currency = finance.currencyWithoutSymbol;
+    hbs.helpers.debcred = finance.currencyWithoutSymbol;
+  }
```
mbayopanda, line 55: 👍 😄

mbayopanda (approved):

PR #3982, opened 2019-11-04
chore(xlsx): clean up renderer code

jniles:
Formats the code, comments it, and uses lodash instead of custom loops to reduce the surface area of code.  It also names the worksheet based on the report name.


jniles:
@jeremielodi can you give me a review?

jniles:
@jeremielodi ?

jeremielodi:
I'm giving you a review

jeremielodi (approved):
LGTM

PR #3983, opened 2019-11-05
Fix(client debts reports) fix form submit bug

### LINKED ISSUE #3934: Bug - Rapport de dettes de Clients does not show red highlight on error on radio buttons
This GIF demonstrates the issue:

RY6MUYcAv5
_GIF: Problem with the error highlight_

The error message says the problem is highlighted in red but it is not.  We should highlight the problem in red.

jeremielodi:
closes #3934


jniles:
This fixes the issue with this form.

bors r+

PR #3984, opened 2019-11-06
Feat: use puppeteer for pdf rendering

### LINKED ISSUE #3274: Replace WKHTMLTOPDF with Puppeteer
It looks like wkhtmltopdf project has stagnated and uses a pretty old rendering engine, preventing us from taking full advantage of the latest CSS technologies.  An alternative to this is [puppeteer](https://github.com/GoogleChrome/puppeteer) which embeds an entire Chromium Browser and uses that for rendering.  It's something we should consider as a replacement.

Pros:
 1. Latest and greatest CSS engine - our HTML versions could be identical to our PDFs.
 2. Potentially easier installations, since we won't need to mess with xvfb and other issues.
 3. Easy update process, first class support.

Cons:
 1. Bloat.  WKHTMLTOPDF is a lean engine, Chrome is a browser.

jniles:
Removes the dependency on WKHTMLTOPDF and replaces it with puppeteer.  This supercedes https://github.com/IMA-WorldHealth/bhima/pull/3388 and is a lot less complex because we rely on an external library to inline assets.  Hopefully this will also work reasonably well across Windows and Linux.

If tests pass, we'll look into updating the documentation and updating styles.  Currently the rendered PDF looks like this on my machine:

image

Closes #3388.  Closes #3274.

-----

**Updates**
I've taken the opportunity to run through the major PDF reports used on the daily basis in BHIMA.  This is what they look like on my machine:

tZxTMenHNb-min

The following sizes work correctly:
 1. Thermal (210mm x 80mm)
 2. A6
 3. A4

I've updated the installation documentation to reflect the lack of WKHTMLTOPDF.  I've also removed wkhtmltopdf from the CI configuration.  We'll need to update https://github.com/IMA-WorldHealth/bhima/pull/3987 to reflect the lack of wkhtmltopdf after it lands.

------

My proposal is that we should wait to merge this until after we've released 1.7.0 (#3986).  That way, we are sure we have a stable base to roll back onto.


mbayopanda (commented):
This feature requires to build lzma module, on Windows the lzma module is built correctly with Node 10 not with Node 13.

With Node10, the feature works fine there are some margin issues but in general the PDF is generated correctly

jniles:
@mbayopanda thanks for the feedback.  I'll continue working on the CSS then.

jniles:
@mbayopanda this is ready for a second review.

server/lib/ReportManager.js
```diff
@@ -223,16 +220,10 @@ class ReportManager {
...
 
-    fs.writeFile(link, this.stream, (err) => {
-      if (err) { return dfd.reject(err); }
-
-      return db.exec(SAVE_SQL, data)
-        .then(() => dfd.resolve({ uuid : reportId }))
-        .catch(dfd.reject)
-        .done();
-    });
+    await fs.promises.writeFile(link, stream);
```
mbayopanda, line 223: 👍 great this promise version

mbayopanda (commented):
LGTM 👍

jniles:
@mbayopanda do you have any final thoughts on this?

mbayopanda:
Its okay, i was looking just if last changes on stock receipts works properly

bors try

jniles:
@mbayopanda good idea.  You can merge it when you are comfortable.

mbayopanda:
@jniles this PR LGTM, i will merge it on the version 1.8.1 since it is a major work i would like to have a version which contains only the code of only this work.

jniles:
Sounds like a good plan!

mbayopanda:
Great job @jniles 

bors r+

PR #3987, opened 2019-11-11
Feat(deployment) cloud-unit config for running BHIMA on a linux server

### LINKED ISSUE #3620: Make a cloud-config for easy deployment of BHIMA
As a step toward #3439, we can use [cloud-config](https://cloudinit.readthedocs.io/en/latest/topics/examples.html) to set up the ideal deployment environment.  This would allow other developers (and even our installation team) to quickly spin up servers with the environment completely set up.  We basically have the barebones of this in [our installation instructions](https://docs.bhi.ma/en/for-developers/installing-bhima.html).

Basically, we would want the basics of a cloud configuration that:
 1. Downloads all server dependencies for BHIMA (mysql, redis, yarn, node, etc)
 2. Creates a non-root bhima user for running the application
 3. Installs BHIMA's dependencies in a given location (say, `/var/www/bhima/` or `/opt/bhima/`)
 4. Uses `pm2` to spin up the server so that it restart automatically.

This would be considered the minimum viable product for a cloud-config.  However, the following bonus additions would be welcome:
 1. Use nginx as a reverse proxy to listen on port 80 and then redirect to wherever the nodejs server is listneing.
 2. Set a hostname and use letsencrypt to get an SSL certificate for the hostname so that we get automagic HTTPS.

jeremielodi:
This PR make  the depoyment of BHIMA easy thanks to the cloud-unit configuration file that we can just copy and paste while creating a server on line.
We tested this script on DigitalOcean and it works flawlessly.

closes #3620


docs/en/install/index.md
```diff
@@ -0,0 +1,108 @@
...
+<ul class="left40">
+<li>
+Be sure to replace YOUR_EMAIL and YOUR_DOMAIN with your email address and your domain.
+</li>
+<li>
+Lets Encrypt uses the email you provide to send notifications about expiration of certificates.
+</li>
+</ul>
+
+# Launch BHIMA to the broser
```
mbayopanda: Launch BHIMA to the browser

docs/cloud-init.yml
```diff
@@ -0,0 +1,129 @@
+#cloud-config
+
+packages:
+  - curl
+  - nginx
+  - git
```
jniles: Looks like we don't use git anymore - see if we can remove this and it still pass successfully.

docs/cloud-init.yml
```diff
@@ -0,0 +1,129 @@
...
+  - ufw
+  - fail2ban
+  - build-essential
+  - redis-server
+  - redis-tools
+  - screen
```
jniles: I added screen during my testing, but it really isn't necessary.  Let's take it out.

docs/cloud-init.yml
```diff
@@ -0,0 +1,129 @@
...
+
+write_files:
+  - path: /tmp/bhima-mysql-user.sql
+    content: |
+      CREATE USER 'bhima'@'localhost' IDENTIFIED BY 'HISCongo2013';
+      GRANT ALL PRIVILEGES ON * . * TO 'bhima'@'localhost';
+      FLUSH PRIVILEGES;
+  - path: /tmp/bhima-latest-release.sh
+    content: |
+      wget -O bhima-latest $(curl -s https://api.github.com/repos/IMA-WorldHealth/bhima/releases/latest | jq -r '.zipball_url');
```
jniles: Beautiful :+1:

docs/en/install/index.md
```diff
@@ -0,0 +1,108 @@
...
+
+<i> Follow these steps below posted at https://docs.opendatakit.org/aggregate-digital-ocean</i>
+
+## Warning
+<p>
+To use this setup, you must able to link a domain name to the machine's IP address. If you don’t own a domain, services such as <a href='https://freedns.afraid.org/'>FreeDNS</a> offer free sub-domains under a range of domains.
+</p>
+
+### Tip
+If you have not already created a DigitalOcean account, use our referral link to do so: https://m.do.co/c/39937689124c.
```
jniles, line 30: Technically, this isn't our referral code link.  We can either put our own (generated by DO) or note that this is the Open Data Kit's referral link.  Can you do that?

docs/fr/install/index.md
```diff
@@ -0,0 +1,108 @@
...
+<ul class="left40">
+<li>
+Si vous n'avez pas le mot de passe root, cliquez sur le nom de votre droplet, sélectionnez Accès dans la navigation de gauche, puis choisissez Réinitialiser le mot de passe root pour qu'un mot de passe vous soit envoyé par courrier électronique.
+</li>
+<li>
+ Vous pouvez également vous connecter via SSH à l'aide de l'adresse ssh root@votre.domaine.
+</li>
+</ul>
+
+3. Une fois connecté, exécutez la commande ``` sudo certbot run --nginx --non-interactive --agree-tos -m YOUR_EMAIL --redirect -d YOUR_DOMAIN.```
```
jniles, line 94: Note that this will only work if they change the [`server_name`](https://github.com/IMA-WorldHealth/bhima/pull/3987/files#diff-aad866a9311fd8ed6d4c5e99baf24a1eR56) variable in the nginx configuration from `demo.bhi.ma` to whatever their domain name is.  We should probably write that somewhere.

jniles (commented):
Just a few comments here.

docs/cloud-init.yml
```diff
@@ -0,0 +1,129 @@
...
+  - bash /tmp/bhima-latest-release.sh
+  - cd bhima
+  - cp .env.development .env.production
+  - bash ./sh/build-init-database.sh
+
+  - yarn global add pm2@latest
+  - yarn --ignore-engines
+  - NODE_ENV=production yarn build
+  - cd bin
+  - NODE_ENV=production pm2 start server/app.js -i 1 --name "bhima"
```
mbayopanda: Could you add `pm2 startup` and `pm2 save` commands, to startup the application automatically when the computer reboot

mbayopanda (commented):
We need to handle startup on reboot with `pm2 startup` and `pm2 save`

jniles:
Interestingly, it looks like GZIP isn't working for the javascript or css files.  This is how I know:

image
_Fig 1: GZIP not applied to JavaScript files_

Let's go ahead and add the following rule to cover our bases:
```
gzip_types application/javascript application/rss+xml application/vnd.ms-fontobject application/x-font application/x-font-opentype application/x-font-otf application/x-font-truetype application/x-font-ttf application/x-javascript application/xhtml+xml application/xml font/opentype font/otf font/ttf image/svg+xml image/x-icon text/css text/javascript text/plain text/xml;
```

This is the source: https://www.keycdn.com/support/enable-gzip-compression

PR #3995, opened 2019-11-14
fix: do not crash an uninstalled server by cron

jniles:
Fixes a bug where cron reports aggressively tried to look up schedules. This caused the server to crash before the application could be installed.  Instead, we log such errors now and keep on going.

### DIFF
--- a/server/controllers/admin/cronEmailReport/index.js
+++ b/server/controllers/admin/cronEmailReport/index.js
@@ -2,7 +2,7 @@
  * HTTP END POINT
  * API controller for the table cron_email_report
  */
-const debug = require('debug')('app');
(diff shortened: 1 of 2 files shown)


server/controllers/admin/cronEmailReport/index.js
```diff
@@ -201,7 +200,7 @@ async function sendEmailReportDocument(record) {
       debug(`(${record.label}) report sent by email to ${contacts.length} contacts`);
     }
   } catch (e) {
-    throw e;
+    debug(e);
```
mbayopanda, line 203: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #3996, opened 2019-11-15
Fix filtering on the entry/exit column

### LINKED ISSUE #3969: Stock Movements registry cannot filter by entry/exit
We can't use the inline filter to filter by entry/exit.  See below:
WuWdqL5OI5

I think this is because we have this element wrapped in a ".label" tag.  If we just had the plain text and changed the background color, we could probably accomplish both.

mbayopanda:
Close #3969


kwilu:
Hello @mbayopanda, it seems like you have referenced #3969 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #3969.`.

```
Author: kwilu

    Fixes #51.
```

jniles (approved):
Works for me.  Great work.

bors r+

PR #4000, opened 2019-11-15
Add more details about stock exit on the voucher receipt

### LINKED ISSUE #3990: Improvements to Stock Exit to Patient receipt
This is what the body of a current stock exit receipt looks like:

image
_Fig 1: Stock exit to patient receipt_

A couple pieces of would be welcome:
 1. Each line concerns a medication/inventory item.  Could we put that in the description of the transaction?  For example, instead of "medicaments vendue aux malades", it would be awesome to have "Medicaments vendue aux malades - 20 comprimes de Quinine 500mg".
 2. It would be nice to include the patient and pharamacy information in the voucher description.  We do this for invoices, for example.  Like "Distribution to JNILES (PA.TPA.123) from PHARMACY123".
 3. If an invoice is linked, we should put the invoice identifier in the text of the description of the voucher so that we can more easily find it later.

mbayopanda:
This PR update description of the voucher and items of the voucher receipt.

voucher stock details

Close #3990


server/models/procedures/stock.sql
```diff
@@ -117,23 +123,47 @@ BEGIN
...
+
+    IF (v_depot_description IS NOT NULL AND v_service_description IS NOT NULL) THEN 
+      SET voucher_description = CONCAT('Distribution to ', v_service_description, ' from ', v_depot_description, ' : ', v_voucher_description);
+    END IF;
+
+    IF (v_depot_description IS NOT NULL AND v_patient_description IS NOT NULL) THEN 
+      SET voucher_description = CONCAT('Distribution to ', v_patient_description, ' from ', v_depot_description, ' : ', v_voucher_description);
+    END IF;
+    
+    IF (v_depot_description IS NOT NULL AND v_patient_description IS NOT NULL AND v_invoice_description IS NOT NULL) THEN
```
jniles: Oof.  This is super complicated.  Can't we find a way to do this in JavaScript to get regular translation support?  For example, this is how the patient invoice description is created: https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/i18n/en/patient_invoice.json#L15.

test/data.sql
```diff
@@ -755,7 +755,7 @@ INSERT INTO `voucher` (uuid, `date`, project_id, currency_id, amount, descriptio
   (@fourth_voucher, CURRENT_TIMESTAMP, 1, 1, 75, 'Fourth Voucher to be Posted', 1, 9);
 
 -- voucher items sample data
-INSERT INTO `voucher_item` VALUES
+INSERT INTO `voucher_item` (`uuid`, `account_id`, `debit`, `credit`, `voucher_uuid`, `document_uuid`, `entity_uuid`) VALUES
```
jniles, line 758: :+1:

server/models/schema.sql
```diff
@@ -1865,6 +1865,7 @@ CREATE TABLE IF NOT EXISTS `voucher_item` (
   `voucher_uuid`    BINARY(16) NOT NULL,
   `document_uuid`   binary(16) default null,
   `entity_uuid`     binary(16) default null,
+  `description`     TEXT NULL,
```
jniles, line 1868: :+1:

jniles (commented):
Technically, this solves the problem, but it can't be translated.  Did you look into creating the string in JS as we've done with other descriptions?  That would make the most sense to me, if it is possible.

server/models/procedures/stock.sql
mbayopanda: We must add a server side api end point for getting details of the description, and before to create a stock movement we must ask the API end point for having description details and after that we can create the stock movement on the client side.

client/src/i18n/en/stock.json
```diff
@@ -41,9 +41,12 @@
     "EXPIRATION"      : "Expiration",
     "EXPIRATION_RISK" : "Expiration Risk",
     "EXIT"            : "Stock Exit",
-    "EXIT_DEPOT"      : "Exit to a depot",
-    "EXIT_PATIENT"    : "Exit to a patient",
-    "EXIT_SERVICE"    : "Exit to a service",
+    "EXIT_DEPOT"      : "Distribution to a depot",
+    "EXIT_PATIENT"    : "Distribution to a patient",
+    "EXIT_PATIENT_ADVANCED" : "Distribution to the patient {{patient}} from depot {{depot}}",
```
jniles, line 46: 👍

server/controllers/admin/services.js
```diff
@@ -40,9 +41,17 @@ function list(req, res, next) {
       LEFT JOIN project AS p ON s.project_id = p.id`;
   }
 
-  sql += ' ORDER BY s.name;';
+  const params = db.convert(req.query, ['uuid']);
+  const filters = new FilterParser(params);
 
-  db.exec(sql)
+  filters.custom('uuid', 's.uuid = ?');
```
jniles: Does equals not work here?

```js
filters.equals('uuid', 'uuid', 's');
```

server/controllers/finance/reports/vouchers/receipt.handlebars
```diff
@@ -61,7 +61,13 @@
       {{#each items}}
         <tr>
           <td>{{number}}</td>
-          <td>{{label}}</td>
+          <td>
+            {{#if description}}
```
jniles, line 65: 👍

client/src/modules/stock/exit/exit.js
```diff
@@ -427,6 +429,53 @@ function StockExitController(
     };
   }
 
+  function buildDescription(entityUuid, invoiceUuid) {
+    const glb = {};
+    return PatientService.read(null, { uuid : entityUuid })
```
jniles: If these services are based on the `PrototypeAPIService`, you should be able to call `.read()` with the `uuid` directly to return a JSON object.  For example:
```js
PatientService.read(null, { uuid : entityUuid })
  .then(([patient]) => {});
```
becomes
```js
PatientService.read(entityUuid)
  .then(patient => {});
```

client/src/modules/stock/exit/exit.js
```diff
@@ -427,6 +429,53 @@ function StockExitController(
...
+      .then(([patient]) => {
```
jniles: I think you can probably combine a lot of these calls together to save space and time.  For example:
```js
// we use all settled here, since any of these might be undefined.
return Promise.allSettled([
  PatientService.read(entityUuid),
  PatientInvoiceService.read(invoiceUuid),
  ServiceService.read(entityUuid),
]).then(([patient, invoice, service]) => {
  // etc..
});
```

client/src/modules/stock/exit/exit.js
```diff
@@ -427,6 +429,53 @@ function StockExitController(
...
+        if (patient) {
+          glb.patientDescription = patient.display_name.concat(` (${patient.reference})`);
```
jniles: If we are mixing both $translate and String.concat(), it probably means we can write a better translation template.

I think what you are trying to achieve is to have a translation with something like this:
```
Distribution to the patient JNILES (PA.TBD.123) from ...
```
And your template is
```
Distribution to the patient {{patient}} from...
```

I propose you change your template to be:
```
Distribution to the patient {{patientName}} ({{patientReference}})
```
And then you could just provide the `patientName` and `patientReference` as two different translation keys.

client/src/modules/stock/exit/exit.js
```diff
@@ -427,6 +429,53 @@ function StockExitController(
...
+
+        if (vm.depot.text && glb.patientDescription) {
+          description = $translate.instant('STOCK.EXIT_PATIENT_ADVANCED', {
+            patient : glb.patientDescription,
+            depot : vm.depot.text,
+          });
+        }
+
+        if (vm.depot.text && glb.patientDescription && glb.invoiceDescription) {
+          description = $translate.instant('STOCK.EXIT_PATIENT_ADVANCED_WITH_INVOICE', {
```
jniles: It looks like your translation key object doesn't change all that much.  Let's put all the keys in a single object and use them for each translation.  Like this:
```js
const i18nKeys = {
  patient : glb.patientDescription,
  depot : vm.depot.text,
  invoice : glb.invoiceDescription
};

// later..
if (x && y) {
  description = $translate.instant('STOCK.EXIT_PATIENT_ADVANCED', i18nKeys);
}

if (y && z) {
 description = $translate.instant('STOCK.EXIT_PATIENT_ADVANCED_WITH_INVOICE', i18nKeys);
}
```

jniles (commented):
@mbayopanda, this is a lot better!   I've made a few comments to tighten up the code in the client controller.  Let me know what you think.

client/src/modules/stock/exit/exit.js
mbayopanda: The reason here is i don't want to get an error if there is nothing, i want to have an empty array instead.

mbayopanda:

jniles (approved):
@mbayopanda, this looks like a good first implementation to me. I would still like to see the stock movement description translated, but we can do that after this is merged.

Once the merge issues are fixed, we can "bors" it.

PR #4005, opened 2019-11-19
chore: remove mz from dependencies

### LINKED ISSUE #3993: Remove mz/fs and use fs.promises instead
I added mz/fs a few years ago to have a promise-based FS. Turns out NodeJS has moved to support promises with the [`fs.promises`](https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_fs_promises_api) API.  We could drop mz/fs (which is somewhat unmaintained) in favor of fs.promises.

jniles:
Removes mz/fs from dependencies and relies on only fs.promises.

Closes #3993.


mbayopanda (approved):
LGTM 👍 

bors r+

PR #4007, opened 2019-11-19
fix(cloud-init): set working directory

jniles:
We need to set the working directory to be able to find the `.env` file.

@mbayopanda can you confirm this solves the issue we ran into?

### DIFF
--- a/docs/cloud-init.yml
+++ b/docs/cloud-init.yml
@@ -60,6 +60,7 @@ write_files:
 
       [Service]
       Environment=NODE_ENV=production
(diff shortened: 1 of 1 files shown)


docs/cloud-init.yml
```diff
@@ -60,6 +60,7 @@ write_files:
 
       [Service]
       Environment=NODE_ENV=production
+      WorkingDirectory=/opt/bhima/bin/
```
mbayopanda, line 63: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4011, opened 2019-11-20
Expand depot text size

mbayopanda:
This PR extend the size of the `text` column in the depot table.


server/models/schema.sql
```diff
@@ -449,7 +449,7 @@ CREATE TABLE debtor_group_subsidy (
 DROP TABLE IF EXISTS `depot`;
 CREATE TABLE `depot` (
   `uuid` BINARY(16) NOT NULL,
-  `text` VARCHAR(50) NOT NULL,
+  `text` VARCHAR(127) NOT NULL,
```
jniles: I believe the max size we can have under `UTF8MB4` is VARCHAR(191).  Let's go ahead and make this VARCHAR(191) so that we don't run into this problem again.

docs/cloud-init.yml
```diff
@@ -63,6 +63,7 @@ write_files:
       WorkingDirectory=/opt/bhima/bin/
       Type=simple
       User=bhima
+      WorkingDirectory=/opt/bhima/bin
```
jniles, line 66: This is already in master :)

jniles (commented):
Nice!  Just some cleanup comments.

docs/cloud-init.yml
jniles: Can you remove it?

PR #4012, opened 2019-11-20
Fix(tests): wait for navigation on patient record

jniles:
We've been experiencing lots of crashes on master.  This may be in part because we don't wait for navigation on the Patient Record page.  This commit fixes that issue.

An example of this bug is shown in this log:
https://travis-ci.org/IMA-WorldHealth/bhima/builds/614450716#L1024


jniles:
@jeremielodi, can you give this a quick review?

jeremielodi (approved):
LGTM

PR #4015, opened 2019-11-22
Enhance stock depot registry

mbayopanda:
This PR : 
- Refactor the bhDepotSelect by using typeahead instead of ui-select
- Add filters in the depot registry
- Add search tool in the depot registry
- Add filter in the change depot modal tool


client/src/i18n/en/depot.json
```diff
@@ -28,7 +28,10 @@
       "NO_DEPOTS" : "Don't see any depots?",
       "SELECT_A_DEPOT" : "Select a depot",
       "NO_DEFAULT_DEPOT" : "You need to choose a depot.",
-      "SELECTION_EXPLANATION" : "It looks like you don't have a default depot set!  Please select your depot from the list below.  You will only need to do this once."
+      "SELECTION_EXPLANATION" : "It looks like you don't have a default depot set!  Please select your depot from the list below.  You will only need to do this once.",
+      "SEARCH":"Search",
+      "MORE_DEPOT_WARNING": "You have a large number of depots, please look for the depot by typing his name in the search input zone",
```
jniles: This isn't a big deal, but it is more natural in English to say "its name" rather than "his name" for objects.  English tend to only [anthropomorphize](https://en.wikipedia.org/wiki/Anthropomorphism) very significant objects like boats, hurricanes, and cups of tea. :tea:

client/src/js/components/bhDepotSelect/bhDepotSelect.js
```diff
@@ -1,6 +1,6 @@
 angular.module('bhima.components')
   .component('bhDepotSelect', {
-    templateUrl : 'modules/templates/bhDepotSelect.tmpl.html',
+    templateUrl : 'js/components/bhDepotSelect/bhDepotSelect.tmpl.html',
```
jniles, line 3: Good move 👍

client/src/js/components/bhDepotSelect/bhDepotSelect.tmpl.html
```diff
@@ -7,18 +7,21 @@
...
-    <ui-select
+    <!-- search by Name  -->
+    <input
+      type="text"
+      class="form-control"
       name="depot_uuid"
-      ng-model="$ctrl.depotUuid"
-      on-select="$ctrl.onSelect($item, $model)"
+      ng-model="$ctrl.depotText"
+      ng-keypress="$ctrl.onKeyPress($event)"
```
jniles: Do you have an `onKeyPress` callback?  I cannot see where it is defined.  Is it actually needed?

client/src/modules/depots/depots.js
```diff
@@ -37,6 +45,8 @@ function DepotManagementController(Depots, ModalService, Notify, uiGridConstants
         displayName : 'DEPOT.LABEL',
         headerCellFilter : 'translate',
         cellTemplate : '/modules/depots/templates/label.tmpl.html',
+        aggregationType : uiGridConstants.aggregationTypes.count,
+        aggregationHideLabel : true,
```
jniles, line 49: Nice 👍

client/src/modules/depots/depots.service.js
```diff
@@ -36,6 +37,15 @@ function DepotService(Api, Modal) {
     }).result;
   };
 
+  service.searchByName = function searchByName(options) {
+    const opts = angular.copy(options || {});
```
jniles: Just FYI - you can accomplish the same thing via [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters).  This is a simpler way of writing the same code:
```js
service.searchByName = function searchByName(options = {}) {
  // 🐤 
}
```
I also don't think you need to copy the opts - you aren't modifying the object in any way.  Normally, we write `angular.copy()` to prevent angular from changing the view value if we remove or add properties to the object.

client/src/modules/depots/modals/search.modal.js
```diff
@@ -0,0 +1,72 @@
...
+SearchDepotModalController.$inject = [
+  'data', '$uibModalInstance', 'Store', 'util', 'StockService',
+];
+
+function SearchDepotModalController(data, Instance, Store, util, Stock) {
+  const vm = this;
+  const displayValues = {};
+  const changes = new Store({ identifier : 'key' });
+
+  const searchQueryOptions = ['text'];
```
jniles: We might eventually want to add "entrepot" to the queries as well.

server/controllers/inventory/depots/index.js
```diff
@@ -152,6 +154,47 @@ function list(req, res, next) {
...
+      d.allow_entry_transfer, d.allow_exit_debtor, d.allow_exit_service,
+      d.allow_exit_transfer, d.allow_exit_loss, BUID(d.location_uuid) AS location_uuid,
+      v.name as village_name, s.name as sector_name, p.name as province_name, c.name as country_name
+    FROM depot d
+      LEFT JOIN village v ON v.uuid = d.location_uuid
+      LEFT JOIN sector s ON s.uuid = v.sector_uuid
+      LEFT JOIN province p ON p.uuid = s.province_uuid
+      LEFT JOIN country c ON c.uuid = p.country_uuid
+    WHERE LOWER(d.text) LIKE ?
+    LIMIT ${limit}
```
jniles: Can you make this use `FilterParser`? It should be best practice to avoid SQL injection attacks.

It will also allow us to expand this with additional search parameters in the future.

server/controllers/stock/core.js
```diff
@@ -388,6 +388,7 @@ function getStockConsumption(periodIds) {
  * @description
  * Algorithm to calculate the CMM (consommation moyenne mensuelle) or average stock consumption
  * over a period for each stock item that has been consumed.
+ * NOTA: A FISCAL YEAR MUST BE DEFINED FOR THE FEATURE WORK PROPERLY
```
jniles, line 391: This seems pretty important :)

client/src/js/components/bhDepotSelect/bhDepotSelect.tmpl.html
```diff
@@ -7,18 +7,21 @@
...
+      uib-typeahead="depot for depot in $ctrl.searchByName($viewValue)"
+      typeahead-loading="$ctrl.loading"
+      typeahead-template-url="/modules/templates/depotList.tmpl.html"
+      typeahead-on-select="$ctrl.onSelect($item, $model)"
+      typeahead-min-length="0"
+      autocomplete="off"
+      translate-attr="{ 'placeholder': 'FORM.SELECT.DEPOT' }"
```
jniles, line 24: This is cool :).

jniles (commented):
This is pretty much perfect.  The functionality is good, and the code is really readable.

One block change - could you use FilterParser to prevent SQL injections?  Even if this code doesn't suffer from it, it is a good habit to get into.  After that, LGTM.

kwilu:

server/controllers/stock/core.js
mbayopanda: Yes... consumption is stored according period ids.

PR #4019, opened 2019-11-25
fix(tests): attempt to fix flakey voucher tests

jniles:
This PR basically does two things:
 1. Removes skipped complex voucher tests and calls "modal.close()" instead of clicking the first close button (just in case).
 2. Ensures that the application waits for the modals to clear on the complex voucher page before trying to trigger the generic income/expense tests.

The hope is this will reduce the intermittent failures we get in the complex voucher module.


jniles:
bors try

jniles:
bors try

PR #4022, opened 2019-11-26
Harmonize stock receipt headers

mbayopanda:
This PR is about to use one code base for stock receipt headers.


jniles (approved):
LGTM.

bors r+

PR #4029, opened 2019-11-29
fix stock identifier for barcode

mbayopanda:
This PR fix an issue related to barcode of stock receipts.


jniles (approved):
Good catch!

bors r+

PR #4031, opened 2019-11-30
fix: limit LIMIT fields to positive numbers

### LINKED ISSUE #4020: Block negative numbers from the "LIMIT" search field.
Looks like I can break the LIMIT field by putting in a negative number (on the depots management search field).   We should try to fix this globally.  See below:

gBevMQxxoY
_GIF 1: I shouldn't be able to do this_

jniles:
This commit changes all the client-side limit fields to only allow positive numbers.

Closes #4020.


mbayopanda (approved):
LGTM 👍 

Good catch

bors r+

PR #4034, opened 2019-12-03
Replace bloquote by div

mbayopanda:
This PR fix stock receipt header size and replace bloquote by div


jniles:
Can you include a screenshot?  Not sure what this does.

mbayopanda:
h4 bloquote

Use of `h4` instead of `h3` because with puppeteer pdf, the h3 text is bigger than h3 with wkhtmltopdf.

Replace bloquote by div with border for have the same visual aspect for note and receipt details.

jniles:
Cheers, thanks.  Changes look good to me.

bors r+

PR #4036, opened 2019-12-03
Fix default label for depot select component

mbayopanda:
This PR fix the default label for the bhDepotSelect component.


jniles (approved):
Great catch!

bors r+

Issue #4014, opened 2019-11-21
Fix Debts Report


mbayopanda:
The Client debts report shows : 
- Double the amount found in the account report
- Display details always in $

![image](https://user-images.githubusercontent.com/896472/69338231-01d12200-0c63-11ea-8428-d8c1c2a0d953.png)

jniles:
I'll grab this one

kwilu:
Hello @jniles, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@kwilu claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

jniles:
@kwilu claim

jniles:
So, I'm not sure if this report makes a lot of sense.  A debt shouldn't be related to a date range - it should be for all time, I think?  I may go ahead and change this to consider debts for all time rather than for a particular time period.

PR #4037, opened 2019-12-04
fix(report): render client debt currencies

jniles:
Fixes a bug with the currency rendering of the client debt report.
Renders the enterprise currency at every level, not just the top level.

Partially addresses #4014.

image


mbayopanda (approved):
LGTM 

bors r+

PR #4039, opened 2019-12-04
Release 1.8.2

mbayopanda:
This commit will be used as reference for the bhima v1.8.2 release which contains some code fixing.

The cloud init file get the last release version, since v1.8.1 has some bugs building BHIMA with cloud-init  will build the application with these bugs (such as bhDepotSelect doesn't have label)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bhima",
-  "version": "1.8.1",
+  "version": "1.8.2",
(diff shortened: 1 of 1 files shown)


jniles (approved):
LGTM

bors r+

jniles:
@mbayopanda it doesn't look like this was [ever released](https://github.com/IMA-WorldHealth/bhima/releases).  Can you tag it?

PR #4056, opened 2019-12-30
fix(cron): adjust weekly unit test

jniles:
The weekly unit test assumed that you would never have the start and
the end of a week in two different months. This is clearly untrue.
We've removed this test and have not replaced it.


jniles:
bros ping

PR #4070, opened 2020-01-04
fix(stock): sort registries correctly

### LINKED ISSUE #4057: Stock Registry shows bold group headers when not grouped by that category
Like this:
image

The registry shouldn't have anything in the header row unless it is being grouped by that thing.

### LINKED ISSUE #4069: Stock Registries do not always sort numbers as numbers
There are multiple fields (quantity, unit cost, etc) in multiple stock registries that do not sort by number value, but instead sort alphabetically.  Further, all the number values should be right-aligned.  The registries I've noted are:

Lots en Stock:
 1. Cout Unitaire => text-right
 2. Quantity => text-right
 3. CMM => text-right

Movements in Stock
 1. Quantite -> Sort is alphabetic.
 
Articles en Stock
 1. Alphabetic Sort of numbers

...though I may have missed a few.

jniles:
Fixes the stock management so that the registries sort correctly instead of sorting alphabetically for numbers.

Closes #4069.  Closes https://github.com/IMA-WorldHealth/bhima/issues/4057.

### DIFF
--- a/client/src/modules/stock/lots/registry.service.js
+++ b/client/src/modules/stock/lots/registry.service.js
@@ -49,10 +49,13 @@ function LotsRegistryService(uiGridConstants, Session) {
     }, {
       field : 'quantity',
       displayName : 'STOCK.QUANTITY',
+      cellClass : 'text-right',
       headerCellFilter : 'translate',
+      type : 'number',
     }, {
       field : 'unit_cost',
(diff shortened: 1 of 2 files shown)


mbayopanda (approved):
LGTM 👍 

bors r+

PR #4075, opened 2020-01-04
fix(stock): hide hidden services from distribution

### LINKED ISSUE #4066: Services marked with the `hidden` flag should be hidden in stock distribution
When you distribute to a service, you have the option of selecting any service in the hospital.  It would be useful to hide certain services, using the functionality implemented in https://github.com/IMA-WorldHealth/bhima/pull/3708.  The Patient Invoice module currently does this, and we need the same functionality in stock distributions.

jniles:
Fixes the services selection modal to ensure that hidden services are not rendered in the stock exit to service.

Closes #4066.

### DIFF
--- a/client/src/modules/stock/exit/modals/findService.modal.js
+++ b/client/src/modules/stock/exit/modals/findService.modal.js
@@ -16,16 +16,8 @@ function StockFindServiceModalController(Instance, Service, Notify, Data, Stock)
 
   Service.read()
     .then(services => {
-      vm.services = services;
-
-      // set defined the previous selected service
-      if (Data.entity_uuid) {
-        const currentService = services.filter(item => {
-          return item.uuid === Data.entity_uuid;
-        });
-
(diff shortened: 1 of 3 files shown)


client/src/modules/stock/exit/modals/findService.modal.js
```diff
@@ -16,16 +16,8 @@ function StockFindServiceModalController(Instance, Service, Notify, Data, Stock)
...
-      // set defined the previous selected service
-      if (Data.entity_uuid) {
-        const currentService = services.filter(item => {
-          return item.uuid === Data.entity_uuid;
-        });
-
-        vm.selected = currentService.length > 0 ? currentService[0] : undefined;
-      }
+      vm.services = services.filter(s => !s.hidden);
+      [vm.selected] = services.filter(s => s.uuid === Data.entity_uuid);
```
mbayopanda, line 20: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4077, opened 2020-01-08
fix(payroll): salaries report fits on page

jniles:
With the change to puppeteer, the salaries report seems to have started
running off the page.  This fixes the report and allows other reports to
take advantage of specifying options.orientation as 'landscape' to set
the page rendering to A4 landscape mode.

_Fig 2: After - All columns fit_
image

### DIFF
--- a/server/controllers/payroll/reports/multipayroll.handlebars
+++ b/server/controllers/payroll/reports/multipayroll.handlebars
@@ -15,7 +15,7 @@
         <h4 class="text-center">
           <strong>
           <span class="text-capitalize">
-            {{translate this.displayName}} : {{this.value}} 
+            {{translate this.displayName}} : {{this.value}}
           </span
           </strong>
         </h4>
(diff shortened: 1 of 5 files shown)


lomamech (approved):
No comments for these improvements, I had difficulty exporting this report to Excel and CSV, would it be possible to add this option in this PR

lomamech:
Let me know if the option to download in Excel and / or CSV would be possible in this PR or in another

jniles:
@lomamech, I'll give it a try and see how much work it is.

jniles:
@lomamech, I'll give it a try and see how much work it is.

jniles:
@lomamech, I think it will probably be best to put this in another PR since it will take a bit of work.  We'll also want tests to accompany the work.  Could we merge this one first and then work on the next one?

lomamech:
Okay, no problem :white_check_mark:

PR #4084, opened 2020-01-13
refactor(stock): merge stock receipts

### LINKED ISSUE #4073: Combine all stock entry/exit receipts into a single URL call
Our stock receipts are too complicated.  Right now, we have to know what kind of stock receipt we need before requesting it.  [For example](https://github.com/IMA-WorldHealth/bhima/blob/master/server/config/routes.js#L833):

```js
  app.get('/receipts/stock/exit_patient/:document_uuid', stockReports.stockExitPatientReceipt);
  app.get('/receipts/stock/exit_service/:document_uuid', stockReports.stockExitServiceReceipt);
  app.get('/receipts/stock/exit_depot/:document_uuid', stockReports.stockExitDepotReceipt);
  app.get('/receipts/stock/exit_loss/:document_uuid', stockReports.stockExitLossReceipt);

  app.get('/receipts/stock/entry_depot/:document_uuid', stockReports.stockEntryDepotReceipt);
  app.get('/receipts/stock/entry_purchase/:document_uuid', stockReports.stockEntryPurchaseReceipt);
  app.get('/receipts/stock/entry_integration/:document_uuid', stockReports.stockEntryIntegrationReceipt);
  app.get('/receipts/stock/entry_donation/:document_uuid', stockReports.stockEntryDonationReceipt);
```

Let's let the server worry about figuring out which receipt to render - just have a single call to something like `/receipts/stock/:uuid?renderer=pdf`.  This would make it much easier to link to different/variable kinds of stock receipts.

jniles:
Merges most stock receipt into a single URL to allow the client to uniformly render the stock receipt with a single API call.

Closes #4073.

---

This PR now contains a breaking change to stock_movements that should probably result in a new release.

The way we calculate stock_movement references has changed.  Instead of a new `reference` for each line of the stock movement, references now behave like `trans_id` or cash/voucher/invoice references, being assigned one per stock_movement.  This allows us to map `document_uuid` one-to-one with the `stock_movement` reference.

The PRIMARY KEY is now the `uuid`, and `document_uuid` also has an INDEX added to allow quick lookups.

Finally, the calculation of `reference` has been moved to a trigger and the associated stored procedures have been removed.  We've also modified the `zRecomputeDocumentMap()` procedure to re-generate the document mappings from the `stock_movement` table.


server/controllers/stock/reports/index.js
```diff
@@ -26,24 +30,106 @@ const stockValue = require('./stock/value');
...
+
+    case Stock.flux.FROM_OTHER_DEPOT:
+      renderer = stockEntryDepotReceipt;
+      break;
+
+    case Stock.flux.FROM_ADJUSTMENT:
+      renderer = stockAdjustmentReceipt;
+      break;
+
+    case Stock.flux.FROM_INTEGRATION:
```
mbayopanda, line 84: There is no `break;` here

mbayopanda (commented):
@jniles i cannot see these pdf receipts : 
- stock exit to service receipt for default tests data
- stock exit to loss receipt for default tests data
- stock adjustment receipt




image

server/controllers/stock/reports/index.js
jniles: I assumed that FROM_INTEGRATION and TO_INTEGRATION used the same receipt, so I combined them.  If they don't, what receipt does stock exit via integration use?

jniles:
Hmm... it looks like we have multiple ways of referencing stock movements, and I'm not sure which is the best way or what the future way should be.

The identifiers created in the stored procedures were of the form `SM.${flux_id}.${reference}`.  However,  [there are also identifiers](https://github.com/IMA-WorldHealth/bhima/blob/master/server/config/identifiers.js#L59) for all kinds of different stock movements and sometimes they are even [displayed on receipts](https://github.com/IMA-WorldHealth/bhima/blob/master/server/controllers/stock/reports/stock_entry_integration.receipt.handlebars#L19).  Which format should we use going forward?

server/controllers/stock/reports/index.js
```diff
@@ -26,24 +30,107 @@ const stockValue = require('./stock/value');
...
 const stockRequisitionReceipt = require('../requisition/requisition.receipt');
 
+/**
+ * @function determineReceiptType
+ *
+ * @description
+ * Figures out the type of stock receipt from the document uuid.  This allows
+ * a uniform API for rendering all stock receipts.
+ */
+async function determineReceiptType(uuid, isDepotTransferExit = -1) {
```
mbayopanda, line 40: 👍 that a good idea to figure out the type of stock receipt just from the document uuid.

mbayopanda (approved):
LGTM 👍 

Great Job Jon

bors r+

jniles:
Looks like a fluke.

bors r+

PR #4086, opened 2020-01-13
Fix the "Sold out" filter in inventory registry

### LINKED ISSUE #4072: The "sold out" filter doesn't work on items in stock registry
**Problem:**  The "Articles in Stock" registry doesn't show items with quantity 0.

**Description:**
If you remove all stock lots via a stock exit, the items no longer show up in the "Items in Stock" registry, since they have quantity = 0.  They should show up when you filter on "sold out", but they do not, since nothing with quantity = 0 shows up.  We need to fix this to alert users of stock outs.

mbayopanda:
This PR : 
- Add a default filter option for including or not empty lots or inventory in the search modal

Closes #4072


client/src/modules/stock/inventories/registry.js
```diff
@@ -167,6 +168,15 @@ function StockInventoriesController(
     gridColumns.openConfigurationModal();
   }
 
+  function setDefaultFilters() {
+    const assignedKeys = Object.keys(stockInventoryFilters.formatHTTP());
+
+    // assign default includeEmptyLot filter
+    if (assignedKeys.indexOf('includeEmptyLot') === -1) {
```
jniles, line 175: I always prefer [Array.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) since it is more readable.

client/src/modules/stock/inventories/registry.js
```diff
@@ -228,6 +236,8 @@ function StockInventoriesController(
   }
 
   function startup() {
+    setDefaultFilters();
```
jniles, line 241: How do other registries set their default filters?  I think they do it in a service.  Can you move this logic to the service?

jniles (commented):
Looks like some tests aren't happy.  The code is clean, let's see if we can figure out what is breaking.

client/src/modules/stock/inventories/registry.js
mbayopanda: If we set it in the stock service, it will be used everywhere (Depot Management, Requisition, Items in stock, Stock Assignment, Stock Movement, Stock Lots) since we don't need it everywhere so i restrict it to this module only.

client/src/modules/stock/inventories/registry.js
jniles: FYI - the reason this failed before is because
```js
assignedKeys.indexOf('includeEmptyLot') === -1
```
is 
```js
!assignedKeys.includes('includeEmptyLot')
```

This is fine though :)

jniles:

bors r+

PR #4088, opened 2020-01-13
bug(Render Account Type)

### LINKED ISSUE #4043: Adding a new account does not render the account type
When a new account is added, the account type for that account is not rendered.  Here is an example illustrating that:

ZRhQ9uZ2cL

**Expected Result**: the account type column should be rendered.

lomamech:
- This option allows you to display the account type during account
  creation.
closes #4043


lomamech:
@jniles all check have passed

jniles (approved):
Looks good to me.  Great work!

bors r+

PR #4091, opened 2020-01-14
refactor(report): ameliorate Articles in Stock

### LINKED ISSUE #4079: Order "Articles in Stock" by inventory group, then inventory label
The articles in stock report looks like this:
image

I believe it is ordered by the inventory code.  A more sensible ordering would be to first order by the kind of inventory (to group medicines and materials separately) and then by the inventory label (to allow one to more quickly find inventory items.

### LINKED ISSUE #4090: Remove the "Delai de Livraison" and "Intervalle de commande" options from the "Articles in Stock"
After discussion with @lomamech and @mbayopanda, it turns out the "Delai de Livraison" and "Intervalle de commande" options are both legacy options that are not longer needed in the report.  We can remove them.  See below:

image

jniles:
Improves the articles in stock report by the following:
 1. Removes the options to add lead time and reorder interval.  These
    are automatically computed.
 2. Adds an option to filter lots based on if they have a quantity or
    not.
 3. Orders first by the inventory group, then the inventory name.

Finally, I've added some better English text description.

L6ELEQAch4

**Note** - the translations for the INCLUDE_EMPTY_LOT will come from https://github.com/IMA-WorldHealth/bhima/pull/4086 when it is merged :)

Closes #4079. Closes #4090.

### DIFF
--- a/client/src/modules/reports/generate/inventory_report/inventory_report.html
+++ b/client/src/modules/reports/generate/inventory_report/inventory_report.html
@@ -31,37 +31,12 @@
             on-change="ReportConfigCtrl.onSelectDate(date)">
           </bh-date-editor>
 
-          <!-- delay -->
-          <div class="form-group col-xs-12 col-sm-6"
-            ng-class="{ 'has-error' : ConfigForm.$submitted && ConfigForm.delay.$invalid }">
-
-            <label class="control-label" translate>FORM.LABELS.DELAY</label>
-            <div class="input-group">
-              <input class="form-control" type="number" bh-integer name="delay" ng-model="ReportConfigCtrl.delay">
-              <span class="input-group-addon" translate>FORM.LABELS.MONTH</span>
-            </div>
-
-            <div class="help-block" ng-messages="ConfigForm.delay.$error" ng-show="ConfigForm.$submitted">
-              <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
-            </div>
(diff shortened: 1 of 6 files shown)


client/src/modules/reports/generate/inventory_report/inventory_report.html
```diff
@@ -31,37 +31,12 @@ <h3 translate>REPORT.STOCK.TITLE</h3>
...
-            <div class="help-block" ng-messages="ConfigForm.purchase_interval.$error" ng-show="ConfigForm.$submitted">
-              <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
-            </div>
-          </div>
-
-          <div class="clearfix"></div>
+          <bh-yes-no-radios
+            label="STOCK.INCLUDE_EMPTY_LOT"
+            value="ReportConfigCtrl.includeEmptyLot"
+            help-text="STOCK.INCLUDE_EMPTY_LOT_HELP"
```
mbayopanda, line 37: These translations are missing

mbayopanda (approved):
LGTM 👍 

bors r+

client/src/modules/reports/generate/inventory_report/inventory_report.html
jniles: This is true - they should be added with https://github.com/IMA-WorldHealth/bhima/pull/4086

PR #4094, opened 2020-01-15
fix(inventory): display sellable flag on edit

jniles:
Fixes a bug reported by HEV where the sellable flag was not displayed on the client while editing the inventory item, leading to much confusion.

Here is the bug:
scTlAAr8g0

And here is the fix:
wp30dbSgVE


jniles:

mbayopanda (approved):
bors +

PR #4096, opened 2020-01-15
fix(stock): Remove patient reference after submit

### LINKED ISSUE #4074: Stock exit to patient should erase patient information after submission
The stock exit to a patient should clear the value of the patient after submission.  Since this does not happen by default, the pharmacist must clear the value of the patient every time before beginning a new search.  See below:
QsOQyrs2Pt
_FIG 1: must clear patient after a successfully submission_

The stock exit module should clear all local data after a successful submission to the server.  This will prevent caching errors and frustrated users in the future.

jniles:
We rarely want to distribute to the same patient twice in a row. Instead of making the user manually clear the patient, we simply remove it after a successful distribution.

Here is the fix in action:
PBqsjs6UcZ

Closes  #4074.

### DIFF
--- a/client/src/modules/stock/exit/exit.js
+++ b/client/src/modules/stock/exit/exit.js
@@ -414,6 +414,7 @@ function StockExitController(
     vm.reference = null;
     vm.displayName = null;
     vm.inventoryNotAvailable = [];
+    delete vm.selectedEntityUuid;
   }
 
   function submit(form) {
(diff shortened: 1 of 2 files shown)


jniles:
@lomamech can I get a review?

mbayopanda (approved):
bors r+

jniles:
bors try

jniles:
bors try

jniles:
bors try

PR #4100, opened 2020-01-15
chore: make POS PDF font size smaller

### LINKED ISSUE #4059: Make PDF receipt font smaller on Thermal Receipts
Apparently, with the transition to puppeteer, the font size increased largely and is using more paper than previously for thermal receipts.  Since once goal of thermal receipts is to keep the cost down, we should figure out how to approach the previous size of receipts and ensure we test these changes on a real thermal receipt printer in the future.

jniles:
Sets the font size smaller on the PDF point of sale receipt.

This is what it looks like on my machine.
image

Ideally, we should have this as a setting somewhere, but for expediency, this PR will solve our clients' issues.

Closes https://github.com/IMA-WorldHealth/bhima/issues/4059

### DIFF
--- a/server/lib/template/partials/point-of-sale-stylesheet.handlebars
+++ b/server/lib/template/partials/point-of-sale-stylesheet.handlebars
@@ -17,6 +17,7 @@
     padding : 0;
     width: 100%;
     height: 100%;
+    font-size: 8pt !important;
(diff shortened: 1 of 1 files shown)


jniles:
Bors r+

Issue #3967, opened 2019-10-30
Improvements to the push email reports


jniles:
The first round of push email reports went out from one of our institutions and there are a number of things we can improve:
 1. The text of each email is hard coded the Balance Report.  It says "Balance ${date}", instead of "Cashflow ${date}" for example.
 2. The filenames are similarly named balance, no matter what the actual report is.
 3. The text has a lot of padding.  This is likely because we used multi-line template strings in JavaScript and that remembers the whitespace.  We can probably fix this by switching to writing out emails as text files that are read into memory and templated before sending.  See below:
![image](https://user-images.githubusercontent.com/896472/67839301-9d072980-faf3-11e9-9bf7-85e866bc9b52.png)

 4. Send all emails via "BCC" tag at once to ensure that the emails are delivered to the entire list at once and reduce sending bandwidth.

jniles:
@kwilu claim

kwilu:
Hello @jniles, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@kwilu abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

PR #4106, opened 2020-01-20
fix(emails): retry sending, improve whitespace

jniles:
Improves the text of the email by including the institution name and a
sign off by the BHIMA team.  Removes unnecessary whitespace from the
email as well, to make it look more professional.

Additionally, we've improved the sending by only sending a single email
per cron report by putting all the contacts in the BCC field.  This
allows us to further make the refinements to retry sending the message
multiple times if internet timeout issues occur using the p-retry
module.

Finally, further logging has been added to assist in our debugging.

Partially addresses #3967.


mbayopanda (approved):
LGTM 👍 

bors r+

jniles:
bors ping

PR #4110, opened 2020-01-21
Feat(user grid) display assigned depots

### LINKED ISSUE #4021: Add "Caisse" and "Depot" columns to the user management grid.
We previously added a very useful column called "roles" to the user management grid.  This allows us to see, at a glance, which users have what roles, and filter on them.  Let's add two more columns:

 1. Depot - the depots each user has access to
 2. Caisse - the caisses each user has access to

It would go here:
image

jeremielodi:
closes #4021

### DIFF
--- a/client/src/modules/users/users.html
+++ b/client/src/modules/users/users.html
@@ -36,6 +36,8 @@
     <div id="users-grid"
       ui-grid="UsersCtrl.gridOptions"
       class="grid-full-height"
+      ui-grid-resize-columns
+      ui-grid-move-columns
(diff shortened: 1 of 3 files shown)


mbayopanda:
Is it possible to have resizable columns with the user grid

jniles (approved):
Looks good to me.  I think we probably will want to rewrite this in the future to use `GROUP_CONCAT()` to reduce complexity.  For the moment, this fulfills the issue though.

mbayopanda (approved):
bors r+

PR #4114, opened 2020-01-23
improvement(Stock Mouvement Document)

### LINKED ISSUE #2694: Stock Movement documents should have either a Voucher or Transaction Reference

lomamech:
Set Voucher reference to
- Adjustment Receipt
- Donation Receipt
- Integration Receipt
- Purchase Receipt
- Loss Receipt
- Patient Receipt

closes #2694

### DIFF
--- a/server/controllers/stock/reports/stock/adjustment_receipt.js
+++ b/server/controllers/stock/reports/stock/adjustment_receipt.js
@@ -34,9 +34,6 @@ async function stockAdjustmentReceipt(documentUuid, session, options) {
     getVoucherReferenceForStockMovement(documentUuid),
   ]);
 
-  console.log('PLUSSSSSSssssss');
-  console.log(results);
-
   const rows = results[0];
   const voucherReference = results[1][0].voucher_reference;
 
--- a/server/controllers/stock/reports/stock/adjustment_receipt.js
+++ b/server/controllers/stock/reports/stock/adjustment_receipt.js
@@ -1,5 +1,5 @@
 const {
-  _, ReportManager, Stock, NotFound, db, identifiers, barcode, STOCK_ADJUSTMENT_TEMPLATE,
+  _, ReportManager, Stock, NotFound, db, identifiers, barcode, STOCK_ADJUSTMENT_TEMPLATE, q,
 } = require('../common');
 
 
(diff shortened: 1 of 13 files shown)


lomamech:
@mbayopanda can you review this PR

server/controllers/stock/reports/stock/adjustment_receipt.js
```diff
@@ -28,7 +28,23 @@ async function stockAdjustmentReceipt(documentUuid, session, options) {
     WHERE m.flux_id IN (${Stock.flux.FROM_ADJUSTMENT}, ${Stock.flux.TO_ADJUSTMENT}) AND m.document_uuid = ?
   `;
 
-  const rows = await db.exec(sql, [db.bid(documentUuid)]);
+  const sqlGetVoucherReference = `
```
jniles: Does every receipt need `sqlGetVoucherReference`?  If so, I would make it a function and put it in the `stock/reports/common.js`.

Like this:
```js
function getVoucherReferenceForStockMovement(documentUuid) {
  const sql = `
    SELECT v.uuid, dm.text AS voucher_reference
    FROM voucher AS v
    JOIN voucher_item AS vi ON vi.voucher_uuid = v.uuid
    JOIN document_map AS dm ON dm.uuid = v.uuid
    WHERE vi.document_uuid = ?
    LIMIT 1;
  `;

  return db.exec(sql, [db.bid(documentUuid)]);
}

exports.getVoucherReferenceForStockMovement = getVoucherReferenceForStockMovement;
```

Then every stock module can import it and just use:
```js
const { getVoucherReferenceForStockMovement }  = require('../common');

// later ..

const results = await q.all([
  db.exec(sql, [db.bid(documentUuid)]), 
  getVoucherReferenceForStockMovement(documentUuid),
]);

// etc
```

This would be more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

server/controllers/stock/reports/stock/entry_purchase_receipt.js
```diff
@@ -38,8 +38,23 @@ function stockEntryPurchaseReceipt(documentUuid, session, options) {
...
+  const sqlGetVoucherReference = `
+    SELECT v.uuid, dm.text AS voucher_reference
+    FROM voucher AS v
+    JOIN voucher_item AS vi ON vi.voucher_uuid = v.uuid
+    JOIN document_map AS dm ON dm.uuid = v.uuid
+    WHERE vi.document_uuid = ?
+    LIMIT 1;
+  `;
+
+  return q.all([
```
jniles: By the way, you don't need to use `q` here.  You can just call the native `Promise.all()` directly.

jniles (commented):
Made a few comments on the code - haven't checked the functionality yet, but the idea is the right one!

lomamech:

jniles (approved):
Looks good to me.  Thanks a ton!

PR #4117, opened 2020-01-23
fix(reports): allow saving Items in Stock

### LINKED ISSUE #4109: "Articles in Stock" does not save as PDF
The Articles In Stock Report does not save as PDF properly.  Here is what it looks like:

8xlqVKLH5o

I'm sure I broke it when I changed the way the URL parameters were passed to the server.

jniles:
Fixes saving the "items in stock" report that was accidentally applying a filter based on the filename.

Closes #4109.

### DIFF
--- a/server/controllers/stock/reports/stock/inventories_report.js
+++ b/server/controllers/stock/reports/stock/inventories_report.js
@@ -1,5 +1,5 @@
 const {
-  _, ReportManager, Stock, formatFilters, pdfOptions, STOCK_INVENTORIES_REPORT_TEMPLATE,
+  _, ReportManager, Stock, formatFilters, STOCK_INVENTORIES_REPORT_TEMPLATE,
 } = require('../common');
 
 /**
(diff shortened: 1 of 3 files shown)


server/controllers/stock/reports/stock/inventories_report.js
```diff
@@ -1,5 +1,5 @@
 const {
-  _, ReportManager, Stock, formatFilters, pdfOptions, STOCK_INVENTORIES_REPORT_TEMPLATE,
```
mbayopanda, line 2: 👍

server/controllers/stock/reports/stock/inventories_report.js
```diff
@@ -34,6 +35,8 @@ async function stockInventoriesReport(req, res, next) {
       options = req.query;
     }
 
+    delete options.label;
```
mbayopanda, line 38: 👍

mbayopanda (approved):
bors r+

PR #4122, opened 2020-01-24
fix(stock): remove underline from stock receipts

jniles:
The stock POS receipts were broken due to the removal of the underline
partial which was an alias for `<hr />`.  I've replaced all the
references to this removed partial with `<hr />`.


mbayopanda (approved):
LGTM 👍 

bors r+

jniles:
Bors r+

PR #4124, opened 2020-01-24
chore(stock): do not require description

### LINKED ISSUE #4082: Do not require stock description if we automatically template them in
Thanks to @mbayopanda's PR https://github.com/IMA-WorldHealth/bhima/pull/4000, we now have automatic templating into our receipts all the relevant information needed for a transaction description.  We could do like invoicing and cash payments and remove the requirement for them to be entered by the user.  That way, only users who need to add additional information will be required to enter some information.

jniles:
Removes the requirement that a user type a description on entry/exit since we automagically template in that description.

Closes #4082.


mbayopanda (approved):
LGTM 👍

jniles:
Bors r+

PR #4125, opened 2020-01-24
Feat(Inventories in stock) considere filters while exporting

### LINKED ISSUE #4067: Articles in Stock Registry export to Excel does not respect the depot_uuid
When you export the Articles in Stock registry to Excel, all depots show up, even if you have filtered on a depot.  This is quite confusing, since the underlying export is different from the data you are looking at.  When you have a filter applied to the registry, only the filtered data should be available for download.

jeremielodi:
closes #4067


mbayopanda (approved):
bors r+

jniles:
Bors r+

PR #4138, opened 2020-01-27
Feat(xlsx export) report template

### LINKED ISSUE #4130: Make a standard XLSX export template for reports
Our users in Vanga are complaining that the HTML -> XLS is no longer working correctly for them on all reports.  We should probably just create a standard XLSX export template that uses either js-xlsx or the excel4node.  It would need to contain the enterprise header, the title, and then a grid to put the rows of data that need to be rendered into the grid.

jeremielodi:
closes #4130

### DIFF
--- a/server/controllers/stock/reports/stock/inventories_report.js
+++ b/server/controllers/stock/reports/stock/inventories_report.js
@@ -20,8 +20,10 @@ async function stockInventoriesReport(req, res, next) {
   const data = {};
 
   // optionReports
-  const optionReport = _.extend({}, req.query, {
+  const metadata = req.session;
+  const optionReport = _.extend({ metadata }, req.query, {
     filename : 'TREE.STOCK_INVENTORY',
+    title : 'TREE.STOCK_INVENTORY',
   });
 
 
--- a/server/controllers/stock/reports/stock/inventories_report.js
+++ b/server/controllers/stock/reports/stock/inventories_report.js
@@ -20,8 +20,8 @@ async function stockInventoriesReport(req, res, next) {
   const data = {};
 
   // optionReports
-  const metadata = req.session;
-  const optionReport = _.extend({ metadata }, req.query, {
+  // const metadata = req.session;
+  const optionReport = _.extend({}, req.query, {
     filename : 'TREE.STOCK_INVENTORY',
     title : 'TREE.STOCK_INVENTORY',
   });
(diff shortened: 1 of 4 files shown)


jniles:
What does this look like?  How can a developer use it in a new report?  What parameters do they need to apply?

jeremielodi:
On the client side, you can specify the renderer like this :

```js
  vm.downloadExcel = () => {
    const filterOpts = stockInventoryFilters.formatHTTP();
    const defaultOpts = {
      renderer : 'xlsxReport',
      lang : Languages.key,
      renameKeys : true,
      displayNames : gridColumns.getDisplayNames(),
    };
    // combine options
    const options = angular.merge(defaultOpts, filterOpts);
    // return  serialized options
  };
```
The most important thing is <b>renderer : 'xlsxReport'</b>

On the server side, as we use <b>ReportManager</b>
You can add the report title as follow:

```js
const optionReport = _.extend({ }, req.query, {
   filename : 'TREE.STOCK_INVENTORY',
   title : 'TREE.STOCK_INVENTORY',
 //subtitle : 'some text'
});
```

<b>The result</b>
excel-report-bhima

server/controllers/stock/reports/stock/inventories_report.js
```diff
@@ -20,8 +20,10 @@ async function stockInventoriesReport(req, res, next) {
   const data = {};
 
   // optionReports
-  const optionReport = _.extend({}, req.query, {
+  const metadata = req.session;
+  const optionReport = _.extend({ metadata }, req.query, {
     filename : 'TREE.STOCK_INVENTORY',
+    title : 'TREE.STOCK_INVENTORY',
```
jniles, line 26: Would it be possible to use the filename as the title if the title doesn't exist?

server/lib/renderers/xlsxReport.js
```diff
@@ -0,0 +1,160 @@
...
+      horizontal : ['center'],
+    },
+  };
+  const right = {
+    alignment : {
+      horizontal : ['right'],
+    },
+  };
+
+  const titleStyle = {
```
jniles, line 87: These styles are used in other renderers, right?  If so, can we move them into a new file `renderers/styles.js` and then import them into each renderer?

jniles (commented):
Looks good!  I've made a few comments.  Once this is merged, we should replace all our XLS renderers with this version.

PR #4143, opened 2020-01-28
feat(reports): add file type detection

jniles:
Adds file type detection to the reports archive page so that we are not
proposing to the user to print reports that are not printable (xlsx for
example).  It also adds features to provide the correct font awesome
icon depending on the extension to help a user determine which kind of
file they will be downloading.

**Example**
RZoqUH5TAk


jniles:
This will need to be merged after https://github.com/IMA-WorldHealth/bhima/pull/4142 to keep things clean.

PR #4157, opened 2020-02-05
chore(inventory): display message when no changes

### LINKED ISSUE #4156: If there are no changes to an inventry item, display a message
If there are no changes to an inventory item, we currently display a blank modal and keep the "download as Excel" option visible.  It looks like this:
image

I propose that we:
  1. Display a message that says "There have been no changes to this inventory item
  2. Disable to "download as Excel" button.

jniles:
Displays a "no changes" message when there are no changes to inventory items.  It also disables the MS Excel export button when there are no changes to download.

It looks like this:
C9Q3XObdUv

Closes #4156.


jniles:
bors try

jniles:
bors try

PR #4181, opened 2020-02-20
fix(tests): specify date & time precision

jniles:
Fixes a few timezone bugs in tests due to the lack of the time component in a few tests.


jniles:
bors try

jniles:
bors try

PR #4183, opened 2020-02-22
chore: rewrite price lists to use pre-filled template

### LINKED ISSUE #4144: Proposal: only download pre-filled template for price list items
I have a proposal to simplify the management of the price list item import and hopefully yield better results.

Currently, the price list module allows you to download either a "pre-filled" template or an empty template. 

I propose that we only allow users to download the "pre-filled" template.  This will make our code simpler, since we will be able to populate this pre-filled template with the inventory uuid.  Then, when the user imports the items, we can use the inventory uuid as the basis for linking up inventory items.  This should make our code cleaner and more efficient.

jniles:
Rewrite the import from CSV file no longer require a stored procedure.  The user is required to download a pre-filled template first, then complete the blank "price" section.  Finally, it improve the English language description of the price import procedure.

Closes #4144.

Here is what it looks like:
M8iHHKYwyL


lomamech (commented):
@jniles I just started the review and trying to reproduce what is done in the git animation, I see that the format uses CSV UTF-8 (Comma delimited) (*. csv) 
image

Is there a particular reason

lomamech:
I also noticed that with the price list that I saved in csv (comma delimited) format, when I import it, the price list is not modified but the message success appears ??


image

jniles:
> 
> 
> @jniles I just started the review and trying to reproduce what is done in the git animation, I see that the format uses CSV UTF-8 (Comma delimited) (*. csv)
> image
> 
> Is there a particular reason

@lomamech yes there is!  If it isn't UTF8, the French accents are broken.  So, to keep French accents,  I always save CSV files as UTF8.

lomamech:
@jniles Could you send me the CSV file to download, I also fear for my version of Excel which is in French and uses (,) instead of (.)

````
12.5 (en) = 12,5 (fr)
````

jniles:
> I also noticed that with the price list that I saved in csv (comma delimited) format, when I import it, the price list is not modified but the message success appears ??

I'll look into this - I think it might have been a pre-existing issue, but I should fix it in this PR even if that is the case.  Thanks for the thorough testing!

jniles:
@lomamech changes landed!

lomamech:
> changes landed!

I will retest it

lomamech:
I just tested with a file that I adapted, I send it to you by email, but I still have the same result, try to check with my file and see if it works

PR #4191, opened 2020-02-25
chore: unmask patient phone number on fiche

### LINKED ISSUE #4190: Print "ND" on the A4 patient fiche if the phone number is not available
On the A4 patient fiche, we currently hide the phone number field if it isn't available.  Our clients have asked us to instead add in the field and write "ND" in the field if it isn't available.

jniles:
Unmask the patient phone number on the fiche as requested by vanga.  If it doesn't exist, it will display "ND".

Closes #4190.


lomamech:
No comment

lomamech (commented):
:white_check_mark:

PR #4214, opened 2020-03-23
upgrade(depot component) use typehead in depot component

jeremielodi:
Optimizes depot component when there is a lot of depots registered

### DIFF
--- a/client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
+++ b/client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
@@ -1,0 +1,97 @@
+
+angular.module('bhima.components')
+  .component('bhDepotSearchSelect', {
+    templateUrl : 'js/components/bhDepotSelectSearch/bhDepotSelectSearch.html',
+    controller  : DepotSearchSelectController,
+    bindings    : {
+      depotsUuids     : '<',
+      label           : '@?',
+      id              : '@?',
+      onChange        : '&',
+      formName        : '@?',
+    },
+  });
+
+DepotSearchSelectController.$inject = [
+  'DepotService', 'uuid',
+];
+
+/**
+* Multiple Depot Selection Component
+*
+*/
+function DepotSearchSelectController(Depots, uuidService) {
+  const $ctrl = this;
+
+  $ctrl.$onInit = () => {
+  // label to display
+    $ctrl.label = $ctrl.label || 'STOCK.DEPOT';
+    // default for form name
+    $ctrl.formName = $ctrl.formName || 'DepotForm';
+    //
+    $ctrl.depotsSected = $ctrl.depotsSected || [];
+    // init the model
+    $ctrl.depotsUuids = $ctrl.depotsUuids || [];
+
+    $ctrl.componentId = $ctrl.id || uuidService().replace('-', '');
+    $ctrl.depots = [];
+
+    // on search text change
+    setTimeout(() => {
+      angular.element(document.getElementById($ctrl.componentId)).keyup(() => {
+        loadDepots($ctrl.depotSelected);
+      });
+    }, 200);
+  };
+
+  $ctrl.$onChanges = () => {
+    loadSelected($ctrl.depotsUuids);
+  };
+
+  // get depots by text
+  function loadDepots(text) {
+    Depots.read(null, {
+      limit : 10,
+      text,
+    }).then(depots => {
+      const selectedUuids = $ctrl.depotsSected.map(d => {
+        return d.uuid;
+      });
+      $ctrl.depots = depots.filter(d => {
+        return !selectedUuids.includes(d.uuid);
+      });
+    });
+  }
+
+  // on select a depot from the typehead
+  $ctrl.onSelect = (item) => {
+    $ctrl.depotsSected.push(angular.copy(item));
+    $ctrl.handleChange();
+    delete $ctrl.depotSelected;
+  };
+
+  // fires the onChange bound to the component boundary
+  $ctrl.handleChange = () => {
+    const depots = $ctrl.depotsSected.map(d => {
+      return d.uuid;
+    });
+    $ctrl.onChange({ depots });
+  };
+
+  // remove a selected depot
+  // eslint-disable-next-line no-shadow
+  $ctrl.remove = (uuid) => {
+    $ctrl.depotsSected = $ctrl.depotsSected.filter(depot => {
+      return depot.uuid !== uuid;
+    });
+    $ctrl.handleChange();
+  };
+
+
+  function loadSelected(depotsUuids) {
+    if (!(depotsUuids.length > 0)) return;
+    Depots.depotsByUuids(depotsUuids).then(depots => {
+      $ctrl.depotsSected = depots;
+    });
+  }
+}
--- a/client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
+++ b/client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
@@ -4,7 +4,8 @@ angular.module('bhima.components')
     templateUrl : 'js/components/bhDepotSelectSearch/bhDepotSelectSearch.html',
     controller  : DepotSearchSelectController,
     bindings    : {
-      depotsUuids     : '<',
+      depotsUuids     : '<?',
+      required        : '<',
       label           : '@?',
       id              : '@?',
       onChange        : '&',
@@ -27,21 +28,12 @@ function DepotSearchSelectController(Depots, uuidService) {
   // label to display
     $ctrl.label = $ctrl.label || 'STOCK.DEPOT';
     // default for form name
-    $ctrl.formName = $ctrl.formName || 'DepotForm';
+    $ctrl.formName = $ctrl.formName || 'DepotSelectForm';
     //
     $ctrl.depotsSected = $ctrl.depotsSected || [];
     // init the model
     $ctrl.depotsUuids = $ctrl.depotsUuids || [];
-
     $ctrl.componentId = $ctrl.id || uuidService().replace('-', '');
-    $ctrl.depots = [];
-
-    // on search text change
-    setTimeout(() => {
-      angular.element(document.getElementById($ctrl.componentId)).keyup(() => {
-        loadDepots($ctrl.depotSelected);
-      });
-    }, 200);
   };
 
   $ctrl.$onChanges = () => {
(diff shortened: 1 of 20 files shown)


jeremielodi:
do not merge this PR yet

client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
```diff
@@ -0,0 +1,97 @@
+
+angular.module('bhima.components')
+  .component('bhDepotSearchSelect', {
+    templateUrl : 'js/components/bhDepotSelectSearch/bhDepotSelectSearch.html',
+    controller  : DepotSearchSelectController,
+    bindings    : {
+      depotsUuids     : '<',
```
jniles: This should be optional `<?`.

client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
```diff
@@ -0,0 +1,97 @@
...
+    //
+    $ctrl.depotsSected = $ctrl.depotsSected || [];
+    // init the model
+    $ctrl.depotsUuids = $ctrl.depotsUuids || [];
+
+    $ctrl.componentId = $ctrl.id || uuidService().replace('-', '');
+    $ctrl.depots = [];
+
+    // on search text change
+    setTimeout(() => {
```
jniles: Just a couple of notes:
 1. When using angular, you need to make sure that angular is aware of changes you are making.  Remember that angular needs to recompute things in a $digest loop.
a. Instead of `setTimeout` use [`$timeout()`](https://docs.angularjs.org/api/ng/service/$timeout).  
b. Instead of `document`, use [`$document`](https://docs.angularjs.org/api/ng/service/$document).  This makes it easier to test components in unit tests.

 2. This `.keyup()` even is never removed.  If you set up an event in the `$onInit()`, you need to remove it in [`$onDestroy()`](https://docs.angularjs.org/guide/component#component-based-application-architecture).

client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
```diff
@@ -0,0 +1,97 @@
...
+function DepotSearchSelectController(Depots, uuidService) {
+  const $ctrl = this;
+
+  $ctrl.$onInit = () => {
+  // label to display
+    $ctrl.label = $ctrl.label || 'STOCK.DEPOT';
+    // default for form name
+    $ctrl.formName = $ctrl.formName || 'DepotForm';
+    //
+    $ctrl.depotsSected = $ctrl.depotsSected || [];
```
jniles, line 33: I think there is a typo here and this should be `depotsSelected`.

client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
jniles: However, you actually don't need any of this code - uib-typeahead will do all this for you.  All you need to do is pass a function that returns an `$http` request to the `uib-typeahead` directive.  We do this is in [bhFindPatient](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/templates/bhFindPatient.tmpl.html#L46).

It looks like this:
```html
 uib-typeahead="patient as patient.display_name for patient in $ctrl.searchByName($viewValue)"
```
where `$ctrl.searchByName` is [this function](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/components/bhFindPatient.js#L165):
```js
  function searchByName(text) {
    $ctrl.loadStatus = 'loading';

    // format query string parameters
    const options = {
      display_name : text.toLowerCase(),
      limit        : LIMIT,
    };

    return Patients.searchByName(options);
};
```

This means you can just set `loadDepots()` as the function to call with `$viewValue` and it will then load the required depots.

client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
```diff
@@ -0,0 +1,97 @@
...
+  // fires the onChange bound to the component boundary
+  $ctrl.handleChange = () => {
+    const depots = $ctrl.depotsSected.map(d => {
+      return d.uuid;
+    });
+    $ctrl.onChange({ depots });
+  };
+
+  // remove a selected depot
+  // eslint-disable-next-line no-shadow
```
jniles: It doesn't look like you need this eslint-disable directive.

server/config/routes.js
```diff
@@ -348,6 +348,8 @@ exports.configure = function configure(app) {
   /* Depot routes */
   app.get('/depots', depots.list);
   app.get('/depots/:uuid', depots.detail);
+  app.get('/depots/group/byUuids', depots.getDepotGroup);
```
jniles: Why write an entirely new route?  Can't you just add the `uuids` parameter to the filters in the [/depots](https://github.com/IMA-WorldHealth/bhima/blob/master/server/controllers/inventory/depots/index.js#L113) route?

Otherwise, you'll need to write a couple new integration tests for this new route.

client/src/js/components/bhDepotSelectSearch/bhDepotSelectSearch.html
```diff
@@ -0,0 +1,24 @@
+<div ng-form="$ctrl.formName" bh-multiple-depot-seach-select>
```
jniles: Looks like a typo here.  I think this should be `bh-multiple-depot-search-select` not `seach`

test/end-to-end/shared/components/bhMultipleDepotSearchSelect.js
```diff
@@ -0,0 +1,14 @@
+/* global by */
+
+module.exports = {
+  selector : '[bh-multiple-depot-seach-select]',
+  set      : async function set(depot, id) {
+
+    const root = $(this.selector);
+    const input = root.element(by.id(id));
+    input.clear();
```
jniles: Both of these `input` operations are asynchronous and need `await` before them.

server/controllers/inventory/depots/index.js
```diff
@@ -234,3 +234,30 @@ function detail(req, res, next) {
...
+// Get depots by their uuids
+function getDepotGroup(req, res, next) {
+  const { depotUuids } = req.query;
+  const uuidGroup = [].concat(depotUuids);
+
+
+  let uuids = '';
+  uuidGroup.forEach((depotUuid, index) => {
+    if (index === 0) {
+      uuids = `HUID('${depotUuid}')`;
```
jniles: Remember that passing in raw values always exposes you to SQL injection attacks.  You should always escape values before putting them into a string to give to the database.

server/controllers/inventory/depots/index.js
```diff
@@ -234,3 +234,30 @@ function detail(req, res, next) {
     .catch(next)
     .done();
 }
+
+// Get depots by their uuids
+function getDepotGroup(req, res, next) {
+  const { depotUuids } = req.query;
+  const uuidGroup = [].concat(depotUuids);
```
jniles: You can do this entire thing by added only two lines to the `list()` function.

The first is changing
```js
  const options = db.convert(req.query, ['uuid']);
```
to 
```js
  const options = db.convert(req.query, ['uuid', 'uuids']);
```

The second is adding a line below:
```js
  filters.equals('uuid', 'uuid', 'd');
```
add this line:
```js
  filters.equals('uuid', 'uuid', 'd');
  filters.equals('uuids', 'uuid', 'd', true);
```
Everything should now work as expected.

jniles (changes requested):
@jeremielodi, thanks for working on this.

I've made a few comments.  The two big issues to address here are:
  1. Using the uib-typeaheads functionality to send the HTTP request, rather than using setTimeout + keyup().  It should function the same way as bhFindPatient.
  2. Using the `/depots` route instead of creating your own.  It should result in only two extra lines of code.

Let me know when you need another review.

client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
jeremielodi: thanks

server/controllers/admin/users/depots.js
```diff
@@ -48,8 +48,9 @@ function create(req, res, next) {
     .addQuery('DELETE FROM depot_permission WHERE user_id = ?;', [req.params.id]);
 
   // if an array of permission has been sent, add them to an INSERT query
-  if (req.body.depots.length) {
-    const data = req.body.depots.map((uuid) => {
+  const depots = req.body.depots || [];
+  if (depots.length) {
```
jniles, line 52: Good fix 👍

client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
```diff
@@ -0,0 +1,89 @@
+
+angular.module('bhima.components')
+  .component('bhDepotSearchSelect', {
+    templateUrl : 'js/components/bhDepotSelectSearch/bhDepotSelectSearch.html',
+    controller  : DepotSearchSelectController,
+    bindings    : {
+      depotsUuids     : '<?',
+      required        : '<',
```
jniles, line 8: Is required optional?  If so, it should be labeled that way `<?`.

server/controllers/inventory/depots/index.js
```diff
@@ -156,6 +156,15 @@ function list(req, res, next) {
     .done();
 }
 
+function hasUuids(uuids, filters) {
```
jniles, line 159: If you don't have this function, what happens?  If you just put in 
```js
filters.equals('uuids', 'uuid', 'd', true)
```
doesn't that work?  I thought it should.

server/lib/filter.js
```diff
@@ -145,7 +145,7 @@ class FilterParser {
 
       if (isArray) { // search in a list of values, example : where id in (1,2,3)
         preparedStatement = `${tableString}${columnAlias} in (${valueString})`;
-      } else { // seach equals one value , example : where id = 2
+      } else { // search equals one value , example : where id = 2
```
jniles, line 148: 👍

test/end-to-end/staffingIndice/payroll-indice.spec.js
```diff
@@ -26,7 +26,7 @@ describe('Multipayroll (indice)', () => {
     currency    : 2,
   };
 
-  it(`should seach period`, async () => {
+  it(`should search period`, async () => {
```
jniles, line 29: 👍

jniles (commented):
Just one quick check on equals and this looks good.  I notice we don't have unit tests for the `filters.equals()` case, so I wouldn't be surprised if the functionality stopped working silently at some point.

server/controllers/inventory/depots/index.js
jeremielodi: An array can have several items but our filter class doesn't add the corresponding number of question marks <b>?</b>


Maybe we should update ```filters.equals```

jniles: It should add an SQL `IN ()` clause instead of an equals.  Isn't that what you need?  [Here is the code](https://github.com/IMA-WorldHealth/bhima/blob/8e8e8bf541ab3408a51cbec06ea8de7c379a68ea/server/lib/filter.js#L147).

mbayopanda:
- Is it possible to have the dropdown of depots list over the footer bar

image

mbayopanda:
It is very fast now 👍

server/controllers/inventory/depots/index.js
jeremielodi: In can see very well that the number of ```  '?'``` doesn't change if we have an array.
We can comeback to this issue in another PR I think.

jeremielodi:
Let me try to use append-to-body if it will work in this case

jeremielodi:
good-sele

jniles (approved):

mbayopanda (approved):
bors r+

Issue #4220, opened 2020-03-25
Remove multiple "id" matches in the end to end tests


jniles:
In our Multiple Payroll (Indice) tests, the following warning occurs:
![image](https://user-images.githubusercontent.com/896472/77530563-d7ea6f00-6e91-11ea-85a0-150feaf5ad96.png)
_Fig 1: Warning multiple id matches_

This is a problem for a couple reasons.
   1. It means that we have multiple equivalent `id` elements on a page.  This isn't supposed to be allowed in HTML.
   2. It means our tests aren't specific enough.  We should always be precise about the element we are selecting.

These tests or components should be rewritten to address the above two points: so that `id`s are unique and that we are sure we are selecting the correct element when we click in a test.

individual-it:
I've fixed the mentioned issue in #4237 but there are more similar ones

jniles:
Closed in #4237.

Thanks so much @individual-it!  If you find other instances, feel free to look into how to fix them :)

individual-it:
please reopen for `Role Management Tests`
![image](https://user-images.githubusercontent.com/2425577/78245948-0dcdca00-7508-11ea-99d1-2660d8d68422.png)

jniles:
@individual-it done!

individual-it:
fix in #4249

individual-it:
more id fixes in #4250

individual-it:
I think this issue can now be closed, there are more non-identical selectors in the UI tests, but I cannot see any where the reason is a not-unique html id

jniles:
Thanks again @individual-it.  Really good clean up work.

PR #4237, opened 2020-03-31
make id of multiple_payroll_indice inputs unique

individual-it:
This PR changes the HTML id of the payment forms to be not only the id that comes from the Database, but a text + the label from the database. This should make the id unique.

It fixes the example given in #4220 but there are many more places with similar issued

before:
image

now:
image


jniles:
Looks good to me!  Excellent work!

bors r+

PR #4241, opened 2020-04-01
test: be more specific when selecting dropdown

individual-it:
when selecting an item from the dropdown, be more specific about what to select
This PR solves 6 occurrences of ` W/element - more than one element found for locator`
e.g. before:
image

now:
image

There are more such cases


jeremielodi (approved):
Good work @individual-it

individual-it:
looks like the build failed because of a network error
image

PR #4249, opened 2020-04-03
make id of actions in RoleActionForm unique

individual-it:
This PR changes the HTML id of the RoleActionForm to be not only the id that comes from the Database, but a text + the description from the database. This should make the id unique.

It fixes second part of #4220 see https://github.com/IMA-WorldHealth/bhima/issues/4220#issuecomment-607797037


jniles (approved):
Looks good to me.

bors r+

PR #4250, opened 2020-04-03
fix: make ids of links in sidebar unique

individual-it:
The links in the sidebar had only a simple number (from DB) as the HTML id, this PR changes it, so that the ids communicate better what they stand for and the change to have duplicate ids is less (see #4220)


jniles (approved):
@individual-it, this looks great to me.  Really good clean up work, and clear namespacing.  Thank you so much!

bors r+

individual-it:
> @individual-it, this looks great to me. Really good clean up work, and clear namespacing. Thank you so much!

@jniles happy to do more, I've just sent you an email, hope it will not end up in the SPAM folder

PR #4255, opened 2020-04-03
Feat(Report directory) Define the report directory on the server

### LINKED ISSUE #3650: Use an environmental variable to set the reports save directory
The early builds of https://github.com/IMA-WorldHealth/bhima/pull/3649 are failing because I've omitted the `server/reports` directory from the build.  Since we remove it every single time we rebuild, it risks users accidentally wiping away their old reports.

Instead of combining this data directory in with the server files, it should really be in its own location, such as `/opt/reports/` or something.  Better yet, let's make it configurable.  We'll set a variable `REPORT_SAVE_DIR` in our environmental files and use that for the saving reports.

jeremielodi:
This PR helps to define the report directory in the environment  file ( dev or production) in other to avoid wiping away the old reports while rebuilding the application.

closes #3650

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,6 +1,7 @@
 # Server Config
 PORT=8080
 
+
 # database configuration
 DB_HOST='localhost'
 DB_USER='bhima'
@@ -13,6 +14,9 @@ SESS_SECRET='XopEn BlowFISH'
 DEBUG=app,errors
 
 UPLOAD_DIR='client/upload'
+
+# Report directory (Define where reports will be saved on the server)
+REPORT_DIR=''
 
 # Mailgun Creds
 MAILGUN_API_KEY="key-XXX"
--- a/server/lib/ReportManager.js
+++ b/server/lib/ReportManager.js
@@ -22,13 +22,14 @@
 
 const _ = require('lodash');
 const path = require('path');
+const tempy = require('tempy');
 const fs = require('fs');
+const db = require('./db');
+const util = require('./util');
 const translateHelper = require('./helpers/translate');
-const util = require('../lib/util');
-
 const BadRequest = require('./errors/BadRequest');
 const InternalServerError = require('./errors/InternalServerError');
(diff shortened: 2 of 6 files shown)


server/lib/ReportManager.js
```diff
@@ -52,7 +52,9 @@ const defaults = {
 };
 
 // Constants
-const SAVE_DIR = path.resolve(path.join(__dirname, '../reports/'));
+const SAVE_DIR = process.env.REPORT_DIR;
```
jniles: I just realized that defining this in `.env.development` will make it hard for developers to work on both linux and windows.  

I propose that we default to a temporary directory of the save directory isn't defined.  We can do it like this: 
```js
const SAVE)_DIR = process.env.REPORT_DIR || tempy.directory();
```
We have the [tempy module](https://github.com/IMA-WorldHealth/bhima/blob/master/package.json#L99) already for our tests. The documentation is [here](https://www.npmjs.com/package/tempy#tempydirectory).

We could also add a "debug()" statement that warns the developers that the `SAVE_DIR` is not defined, so it will use a temporary directory.

You would be able to remove the line from protractor's configuration as well.

.env.development
```diff
@@ -14,6 +15,9 @@ DEBUG=app,errors
 
 UPLOAD_DIR='client/upload'
 
+# Report directory (Define where reports willbe saved on the server)
+REPORT_DIR='C:\BHIMA\'
```
jniles: This is a great comment.  I propose we leave this blank for now:
```sh
REPORT_DIR=''
```

jniles (commented):
@jeremielodi this is exactly what was requested.  Good work.

There is one small change now that I've realized we might not be able to work well across windows/linux/mac with this change.  I've made a suggestion to deal with that.  Let me know what you think.

jniles (approved):
LGTM.

bors r+

PR #4262, opened 2020-04-07
Feature stock assignment inventory by code

lomamech:
feature(Stock Assignment Inventory select by Code)
- Use component for select inventory
- Update E2E test by the using of component inventorySelect


lomamech:
@jeremielodi @mbayopanda @jniles All checks have passed i need a review

test/end-to-end/stock/stock.assign.spec.js
```diff
@@ -128,4 +128,4 @@ function StockAssignTests() {
   });
 }
 
-describe('Stock Assign Module', StockAssignTests);
+describe.only('Stock Assign Module', StockAssignTests);
```
jniles: This needs to be removed - it is only testing these tests.

lomamech: Sorry

lomamech:
@jniles  I just fixed it

jniles (dismissed):
@lomamech I just realized this needs to use `$ctrl.availableInventories`.  It shouldn't use all inventory, but only those that are available.  `bhInventorySelect` might not be able to do that.

One thing you could do is make `bhInventorySelect` take an optional parameter `depot-uuid` that will then filter based on inventory in the depot.

client/src/modules/stock/assign/modals/action.modal.html
```diff
@@ -25,17 +25,17 @@
...
         ng-model="$ctrl.inventory_uuid"
         on-select="$ctrl.onSelectInventory($item)"
         ng-disabled="!$ctrl.availableInventories.length">
         <ui-select-match placeholder="{{ 'FORM.SELECT.INVENTORY' | translate }}">
-          <span>{{$select.selected.text}}</span>
+          <span>{{$select.selected.hrLabel}}</span>
         </ui-select-match>
-        <ui-select-choices ui-select-focus-patch repeat="i.uuid as i in $ctrl.availableInventories | filter: { 'text': $select.search }">
-          <span ng-bind-html="i.text | highlight:$select.search"></span>
+        <ui-select-choices ui-select-focus-patch repeat="i.uuid as i in $ctrl.availableInventories | filter: { 'hrLabel': $select.search }">
```
jniles, line 35: Excellent 👍

client/src/modules/stock/assign/modals/action.modal.html
```diff
@@ -76,10 +76,13 @@
 
 
     <!-- quantity -->
-    <div class="form-group">
+    <div class="form-group" ng-class="{ 'has-error' : ModalForm.$submitted && ModalForm.quantity.$invalid }">
```
jniles, line 79: Nice 👍  Good catch!

client/src/modules/templates/bhInventorySelect.tmpl.html
```diff
@@ -12,6 +12,7 @@
       name="inventory_uuid"
       ng-model="$ctrl.inventoryUuid"
       on-select="$ctrl.onSelect($item, $model)"
+      ng-disabled="$ctrl.ngDisabled"
```
jniles: Shouldn't this be removed?  Where did this come from?

jniles (commented):
@lomamech, this looks good to me.  I've made one comment that I think is just an accident.  You left something in `bhInventorySelect` even though you didn't use it.

lomamech:
@jniles  All checks have passed

jniles (approved):
Perfect :+1:

bors r+

PR #4267, opened 2020-04-08
test(filters): add equals() array test

### LINKED ISSUE #4236: FilterParser needs more unit tests
As referenced in https://github.com/IMA-WorldHealth/bhima/pull/4214, we only have a couple of tests for the [lib/filter.js](https://github.com/IMA-WorldHealth/bhima/blob/8e8e8bf541ab3408a51cbec06ea8de7c379a68ea/server/lib/filter.js#L139) library.  We need to ensure that the array functionality is also tested in a unit test.  Any other coverage issues should also be addressed.

jniles:
Adds an array unit test for the equals() method of the filter parser library.

Closes #4236.

### DIFF
--- a/server/lib/filter.js
+++ b/server/lib/filter.js
@@ -144,7 +144,7 @@ class FilterParser {
       let preparedStatement = '';
 
       if (isArray) { // search in a list of values, example : where id in (1,2,3)
-        preparedStatement = `${tableString}${columnAlias} in (${valueString})`;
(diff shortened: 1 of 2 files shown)


jniles:

jeremielodi:
okay, I'm working on it

jeremielodi (approved):
Tested and approved by @jeremielodi

PR #4268, opened 2020-04-08
chore(install): do not enforce password meter

### LINKED ISSUE #4243: Do not enforce password policy on installation page
We enforce a strict password policy via the [bhPasswordMeter](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/directives/bhPasswordMeter.js) component.  While this is good, the user doesn't have a chance to opt-out on installation.  We should make this rule opt-in, and only applicable after the server is installed.

jniles:
This removes the validation on strict passwords for the installation page.  It should make things less confusing for users getting up and running quickly.

Closes #4243.


lomamech (approved):
No comments, changes are minor

PR #4269, opened 2020-04-08
feature(Add stock movement in the patient financial Status)

### LINKED ISSUE #4141: Add stock movements to the Patient Financial Status report
We should add a section to the patient financial status panel in the Patient Record page that links to the stock movements distributed to the patient.  It would go right here:

image

lomamech:
- Implement API for get the stock movement related to the Patient
- (test) add integration test for get patients/:uuid/stock/movements
- Sanitaze code
closes #4141

### DIFF
--- a/client/src/js/components/bhPatientFinancialActivity.js
+++ b/client/src/js/components/bhPatientFinancialActivity.js
@@ -59,6 +59,16 @@ function PatientFinancialActivityCtrl(Patients, moment, Session, Constants) {
       .finally(() => {
         $ctrl.loading = false;
       });
+
+    Patients.getStockMovements($ctrl.patientUuid)
+      .then(data => {
+        $ctrl.dataMovement = data;
+        $ctrl.noStockMovement = data.length === 0;
+      })
+      .finally(() => {
+        $ctrl.loading = false;
+      });
+
   };
 
 
--- a/client/src/js/components/bhPatientFinancialActivity.js
+++ b/client/src/js/components/bhPatientFinancialActivity.js
@@ -33,15 +33,22 @@ function PatientFinancialActivityCtrl(Patients, moment, Session, Constants) {
 
     $ctrl.loading = true;
 
-    Patients.getFinancialActivity($ctrl.patientUuid)
-      .then(data => {
-        $ctrl.data = data;
+    Promise.all([
+      Patients.getFinancialActivity($ctrl.patientUuid),
+      Patients.getStockMovements($ctrl.patientUuid),
+    ])
+      .then(([financialData, stockData]) => {
+        $ctrl.data = financialData;
+        const dataMovement = [];
+
+        // limit this to the most recent stock movements items
+        const limitRecent = 5;
 
         // indicate that the patient does not have a financial history
-        $ctrl.noFinancialHistory = data.transactions.length === 0;
+        $ctrl.noFinancialHistory = financialData.transactions.length === 0;
 
-        $ctrl.status = calculateFinancialStatus(data.transactions, data.aggregates);
-        $ctrl.groups = groupFinancialRecords(data.transactions);
+        $ctrl.status = calculateFinancialStatus(financialData.transactions, financialData.aggregates);
+        $ctrl.groups = groupFinancialRecords(financialData.transactions);
 
         $ctrl.hasWarnings = (
           $ctrl.status.hasExcessiveDebt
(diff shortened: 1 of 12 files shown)


lomamech:
@jniles checks have passed

client/src/js/components/bhPatientFinancialActivity.js
```diff
@@ -59,6 +59,16 @@ function PatientFinancialActivityCtrl(Patients, moment, Session, Constants) {
       .finally(() => {
         $ctrl.loading = false;
       });
+
+    Patients.getStockMovements($ctrl.patientUuid)
```
jniles: @lomamech, could you combine the `Patient.getFinancialActivity()` and `Patient.getStockMovements()` call with a `Promise.all()` and end it with the `.finally()`?

Like this:
```js
Promise.all([
 Patients.getFinancialActivity($ctrl.patientUuid),
 Patients.getStockMovements($ctrl.patientUuid),
]).then([financialData, stockData] => {
  // ... etc ...
})
.finally(() =>{
  $ctrl.loading = false;
});
```

The reason for doing this is that we don't want to set `$ctrl.loading = false` too early.

client/src/js/components/bhPatientFinancialActivity.js
```diff
@@ -59,6 +59,16 @@ function PatientFinancialActivityCtrl(Patients, moment, Session, Constants) {
...
+      .then(data => {
+        $ctrl.dataMovement = data;
+        $ctrl.noStockMovement = data.length === 0;
```
jniles: If there is no financial activity (but there is stock activity), this component will be covered up with a message.  Here is an example:
image

There is stock data here, but no financial data.  Could you change [this line](https://github.com/IMA-WorldHealth/bhima/pull/4269/files#diff-8422e147a296cedc27c69a1710e1cd2aL21) from:
```html
<bh-panel-overlay ng-if="$ctrl.noFinancialHistory">
```
to
```html
<bh-panel-overlay ng-if="$ctrl.noFinancialHistory && $ctrl.noStockMovement">
```

client/src/modules/patients/patients.service.js
```diff
@@ -345,6 +346,19 @@ function PatientService(
...
+  /**
+   * @method getStockMovements()
+   *
+   * @description
+   * This function makes it possible to list the stock movements linked to the Patient
+   *
+   */
+  function getStockMovements(uuid) {
+    const path = 'patients/:uuid/stock/movements';
+    return service.$http.get(path.replace(':uuid', uuid))
```
jniles: Just FYI - you can also do this: 
```js
 const path = `patients/${uuid}/stock/movements`;
```
and skip the `.replace()` part.

client/src/modules/patients/record/bh-patient-financial-activity.html
```diff
@@ -79,6 +79,33 @@ <h5 ng-show="!$ctrl.status.isInGoodStanding" class="text-danger">
...
+          <table class="table table-responsive table-condensed">
+            <thead>
+              <tr>
+                <td translate>FORM.LABELS.NR</td>
+                <td translate>FORM.LABELS.REFERENCE</td>
+                <td translate>FORM.LABELS.DATE</td>
+              </tr>
+            </thead>
+            <tbody>
+              <tr ng-repeat="dataMovement in $ctrl.dataMovement">
```
jniles, line 101: We should probably limit this to the most recent ... 5 items?  Or something like that.  We don't want the grid to go on forever!

I would just implement this on the client side for the moment.

server/controllers/medical/patients/index.js
```diff
@@ -678,3 +679,28 @@ function getDebtorBalance(req, res, next) {
...
     .catch(next);
 }
+
+/**
+ * @function getStockMovements
+ *
+ * @description
+ * returns the stock Movements to the patient.
+ */
+function getStockMovements(req, res, next) {
```
jniles, line 689: Beautiful.  Does what it needs to do, and nothing else :+1:

client/src/modules/patients/record/bh-patient-financial-activity.html
```diff
@@ -79,6 +79,33 @@ <h5 ng-show="!$ctrl.status.isInGoodStanding" class="text-danger">
             (<span am-time-ago="$ctrl.data.aggregates.until"></span> <i class="fa fa-clock-o"></i>)
           </dd>
         </dl>
+
+        <hr/>
+        <h5 class="text-success" style="margin-top:0;">
```
jniles: Can you remove the `text-success` here?  I'm not sure it makes sense to use here.

image

jniles (commented):
@lomamech, this is great!  Well done!  I used it with Vanga's database, and it immediately gave me a lot more power to ask questions of the data.

There are a few corner cases that I've highlighted in the comments.  Could you take a look at those?  After that, it LGTM.  Great job!

jniles (approved):
Looks like everything as passed!  Excellent implementation.

Bors r+

lomamech:
@jniles All checks have passed

image

image

image

image

Issue #4076, opened 2020-01-05
Rewrite Stock Adjustments to take in a quantity and make the correct adjustments


jniles:
Our Stock Adjustments requires the user to:
 1. Pick augmentation or reduction
 2. List all articles that are increased or decreased by the _amount_ of the increase or decrease

This is not user friendly.  A better way, proposed by our partners in Vanga, would be to record the _amount_ of each lot when you do an inventory, and let the system figure out whether it increased or decreased and generate the final transaction.  We should implement this method to save time from our users.

Here is an example to gain inspiration from:
![Stock Inventory Adjustment](https://user-images.githubusercontent.com/896472/78895023-048fb080-7a66-11ea-8b32-633f7f4f4854.png)

kwilu:
Hello @mbayopanda, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@kwilu claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

jniles:
Closed in https://github.com/IMA-WorldHealth/bhima/pull/4369

PR #4276, opened 2020-04-09
Stock inventory feature

mbayopanda:
This PR add a new way of performing stock inventory : 
- By removing all previous stock of a the selected depot
- By adding new stock as the real stock


jniles:
@mbayopanda I took a look at this, and I don't think this the functionality Vanga wanted.  This is my fault - I didn't specify the issue #4076 well.

They are used to having a giant registry like the stock lots module that shows every lot in the depot.  It would look something like this:
Stock Inventory Adjustment

I've updated the issue to show this.

The idea is that it should look like the inventory fiche they use to calculate stock on the shelf.  No need for a date or even a description.  We can generate both of those automatically.

mbayopanda:
@jniles what if there are lots in the real depot and not in bhima ? How to add them during the inventory adjustment ?

jniles:
@mbayopanda, I would say we ignore that case for the moment.  We will still have the other adjustment module for a time.

kwilu:

mbayopanda:
There is new one #4369 which is a refactoring

PR #4277, opened 2020-04-09
Feature: use previous purchase price of medicine in Purchase Order

### LINKED ISSUE #4119: Use previous order price for next order price of medicines
*Brief Description*
When ordering medication supply have the system automatically use the last price of the product to calculate the price of the order.  

*Full Description*
When creating a purchase order there is a hard stop when creating a requisition unless you enter the Prix d'achat.  Having the latest price would reduce the time needed to look this price up resulting probably a 50% time saving in the time needed to produce a requisition.  When stock is received one must still verify the price of goods purchased so using the old price to order will also alert the seller if they have increased their price.  

*Severity*
  a. High - Having this would help make much easier to order medications.  

priority: High.   A valuable feature that would really help.

jniles:
Implements the feature described in #4119.

This feature uses the last purchase price of the inventory item as the price in a new purchase order instead of the inventory price.  This makes better sense, since the inventory price is often the sale price, not that purchase price.

Closes #4119.

### DIFF
--- a/client/src/modules/purchases/create/PurchaseOrderForm.js
+++ b/client/src/modules/purchases/create/PurchaseOrderForm.js
@@ -38,7 +38,7 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
     this.inventory = new Pool({ identifier : 'uuid', data : [] });
 
     // set up the inventory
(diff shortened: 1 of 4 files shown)


mbayopanda (approved):
LGTM 👍

PR #4278, opened 2020-04-09
feat(purchase): add barcode to purchase orders

### LINKED ISSUE #4272: Add barcodes to purchase order receipts
It appears we don't have barcodes on the purchase order receipts.  Here is an example:

image
_Fig 1: Purchase Order with missing barcode_

We should implement barcodes on the A4 purchase order receipts so that they can be referenced later (for example, on stock entry).

jniles:
This commit adds barcodes to the purchase order receipts.  I've also re-balanced the size of the fonts on the purchase receipt to be more in line with the other modules (invoice, cash, etc).

Closes #4272.

It looks like this:
image

Note that the yellow highlight is to indicate what is in the barcode to demonstrate the barcode is working, and this has been removed from the actual receipt.

### DIFF
--- a/server/controllers/finance/purchases.js
+++ b/server/controllers/finance/purchases.js
@@ -17,11 +17,14 @@ const moment = require('moment');
 
 const { uuid } = require('../../lib/util');
 const db = require('../../lib/db');
+const barcode = require('../../lib/barcode');
 const BadRequest = require('../../lib/errors/BadRequest');
 
 const identifiers = require('../../config/identifiers');
 const FilterParser = require('../../lib/filter');
 const util = require('../../lib/util');
+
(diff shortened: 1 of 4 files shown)


lomamech (commented):
I just tested this branch, but I have a small question concerning a small detail below the barcode

image

lomamech:
image

lomamech:
Let me know if it doesn't matter

jniles:
@lomamech, thanks for the review.  I removed that detail because it shouldn't appear.  I was just showing the information in the barcode for you to review.

lomamech:
> @lomamech, thanks for the review. I removed that detail because it shouldn't appear. I was just showing the information in the barcode for you to review.

Ok

PR #4284, opened 2020-04-10
Feat(voucher tool) patient fleg invoices

### LINKED ISSUE #3532: Need voucher tool to write off patient debt as an expense
We currently have voucher tools to do prise en charge of a patient's invoices and to do convention payments.  What we lack is a way to write off patient debt that will never be paid (.e.g. the patient fled).  This would give the hospitals a way to zero out invoices easily without going through the expensive complex vouchers process.

The basic tool would be identical to the convention payments voucher tool except:
 1. Instead of a cashbox, you must pick an expense account.
 2. The voucher type would be set to autre charge perte or something similar.

jeremielodi:
closes #3532

### DIFF
--- a/client/src/i18n/en/vouchers.json
+++ b/client/src/i18n/en/vouchers.json
@@ -32,6 +32,8 @@
       "TRANSFER_DESCRIPTION": "Transfer: From {{ fromAccount }} to {{ toAccount }} for {{ amount }} {{ symbol }}",
       "CONVENTION_PAYMENT_DESCRIPTION": "Convention Payment: Payment for {{ numInvoices }} invoice(s) by {{ debtorGroupName }}.",
       "CONVENTION_INVOICES":"Convention - Invoices payment",
+      "WRITE_OFF_PATIENT_INVOICES":"Write off patient invoices",
+      "WRITE_OFF_PATIENT_INVOICES_DESC":"Writing off a loss of an unrecoverable invoice",
       "SUPPORT_PAYMENT_DESCRIPTION":"Debt transfer from {{ patientName }} ({{ patientReference}}) for {{ numInvoices }} invoice(s) {{ invoiceReferences }}.",
       "SUPPORT_FORM":"Patients Support - Form",
       "SUPPORT_TEXT":"When the above two fields are filled in, we will look for open debts for the patient to assign to the \"Support Patient\" Account.",
--- a/client/src/modules/vouchers/complex-voucher.html
+++ b/client/src/modules/vouchers/complex-voucher.html
@@ -20,6 +20,13 @@
               </a>
             </li>
 
+            
+            <li role="menuitem">
+              <a href data-action="patient-fled" ng-click="ComplexVoucherCtrl.openPatientFledModal()">
+                <i class="fa fa-building"></i> <span translate>VOUCHERS.GLOBAL.WRITE_OFF_PATIENT_INVOICES</span>
+              </a>
+            </li>
+
             <li role="menuitem">
               <a href data-action="generic-expense" ng-click="ComplexVoucherCtrl.openGenericIncomeModal()">
                 <i class="fa fa-level-down"></i> <span translate>VOUCHERS.GLOBAL.GENERIC_INCOME</span>
--- a/client/src/modules/vouchers/toolkit.service.js
+++ b/client/src/modules/vouchers/toolkit.service.js
@@ -16,6 +16,11 @@ function VoucherToolkitService(Modal) {
       controller  : 'ConventionPaymentKitController',
       templateUrl : 'modules/vouchers/toolkit/convention_payment/convention_payment.modal.html',
     },
+    // this tool help to write off patient debt as an expense
+    patient_fled : {
+      controller  : 'PatientFledKitController',
+      templateUrl : 'modules/vouchers/toolkit/patient_fled/patient_fled.modal.html',
+    },
     generic_income : {
       controller  : 'GenericIncomeKitController',
       templateUrl : 'modules/vouchers/toolkit/generic_income/generic_income.html',
@@ -36,6 +41,14 @@ function VoucherToolkitService(Modal) {
       controller  : 'PaymentEmployeeKitController',
       templateUrl : 'modules/vouchers/toolkit/payment_employee/payment_employee.modal.html',
     },
+  };
+
+  service.openConventionPaymentModal = function openConventionPaymentModal() {
+    return open(service.tools.convention_payment);
+  };
+
+  service.openPatientFledModal = function openPatientFledModal() {
+    return open(service.tools.patient_fled);
   };
 
   service.openConventionPaymentModal = function openConventionPaymentModal() {
--- a/client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
+++ b/client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
@@ -1,0 +1,192 @@
+angular.module('bhima.controllers')
+  .controller('PatientFledKitController', PatientFledKitController);
+
+PatientFledKitController.$inject = [
+  '$uibModalInstance', 'DebtorService', 'NotifyService',
+  'SessionService', 'bhConstants', '$translate', 'VoucherToolkitService',
+];
+
+// Import transaction rows for writting off patient invoices
+function PatientFledKitController(
+  Instance, Debtor, Notify, Session, bhConstants, $translate, ToolKits,
+) {
+  const vm = this;
+
+  const { MAX_DECIMAL_PRECISION } = bhConstants.precision;
+
+  // global variables
+  vm.gridOptions = {};
+  vm.account = {};
+  vm.enterprise = Session.enterprise;
+
+  // expose to the view
+  vm.selectGroupInvoices = selectGroupInvoices;
+  vm.close = Instance.close;
+  vm.import = submit;
+
+  // Set up page elements data (debtor select data)
+  vm.onSelectPatient = onSelectPatient;
+  vm.onSelectAccount = onSelectAccount;
+
+  function onSelectPatient(patient) {
+    vm.patient = patient;
+    selectGroupInvoices(vm.patient);
+  }
+
+  function onSelectAccount(account) {
+    vm.account.id = account.id;
+  }
+
+
+  // get debtor group invoices
+  function selectGroupInvoices(patient) {
+    vm.loading = true;
+
+    Debtor.invoices(patient.debtor_uuid)
+      .then((invoices = []) => {
+        // total amount
+        const total = invoices.reduce(aggregate, 0);
+
+        invoices.forEach(invoice => {
+          invoice.date = new Date(invoice.date);
+        });
+
+        // sort in order
+        invoices.sort((a, b) => a.date > b.date);
+
+        // invoices
+        vm.gridOptions.data = invoices.filter(inv => {
+          return inv.balance > 0;
+        });
+
+        // make sure we are always within precision
+        vm.totalInvoices = Number.parseFloat(total.toFixed(MAX_DECIMAL_PRECISION));
+      })
+      .catch((err) => {
+        vm.hasError = true;
+        Notify.handleError(err);
+      })
+      .finally(() => {
+        vm.loading = false;
+      });
+  }
+
+  // generate transaction rows
+  function generateTransactionRows(result) {
+    const rows = [];
+
+    const accountId = vm.account.id;
+    const patientAccountId = result.patient.account_id;
+    const { invoices } = result;
+
+    const line = ToolKits.getBlankVoucherRow();
+    line.account_id = accountId;
+    line.debit = vm.totalSelected;
+    line.credit = 0;
+    rows.push(line);
+
+
+    // then loop through each selected item and credit it with the patient account
+    invoices.forEach(invoice => {
+      const row = ToolKits.getBlankVoucherRow();
+
+      row.account_id = patientAccountId;
+      row.reference_uuid = invoice.uuid;
+      row.entity_uuid = invoice.entity_uuid;
+      row.credit = invoice.balance;
+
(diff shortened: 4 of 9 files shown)


client/src/modules/vouchers/complex-voucher.html
```diff
@@ -20,6 +20,13 @@
               </a>
             </li>
 
+            
+            <li role="menuitem">
+              <a href data-action="convention-payment" ng-click="ComplexVoucherCtrl.openPatientFledModal()">
```
jniles: This probably shouldn't be `convention-payment`.  It should be called something else.

client/src/i18n/en/vouchers.json
```diff
@@ -32,6 +32,7 @@
       "TRANSFER_DESCRIPTION": "Transfer: From {{ fromAccount }} to {{ toAccount }} for {{ amount }} {{ symbol }}",
       "CONVENTION_PAYMENT_DESCRIPTION": "Convention Payment: Payment for {{ numInvoices }} invoice(s) by {{ debtorGroupName }}.",
       "CONVENTION_INVOICES":"Convention - Invoices payment",
+      "PATIENT_FLED":"Patient fled - Invoices payment",
```
jniles: Let's change this name to "Write off patient invoices" or something.

A patient fleeing the hospital isn't the only reason we may need this tool.  A patient may also die and have no family to pay their debt.  Or they may simply negotiate with the hospital to show they aren't able to pay and the hospital will absorb the debt.

client/src/modules/vouchers/toolkit.service.js
```diff
@@ -16,6 +16,11 @@ function VoucherToolkitService(Modal) {
       controller  : 'ConventionPaymentKitController',
       templateUrl : 'modules/vouchers/toolkit/convention_payment/convention_payment.modal.html',
     },
+    // this tool help to write off patient debt as an expense
+    patient_fled : {
+      controller  : 'PateintFledKitController',
```
jniles: The word "Patient" has a typo here.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,193 @@
...
+  function generateTransactionRows(result) {
+    const rows = [];
+
+    const cashboxAccountId = vm.account.id;
+    const conventionAccountId = result.convention.account_id;
+    const { invoices } = result;
+
+    // first, generate a cashbox row
+    const cashboxRow = ToolKits.getBlankVoucherRow();
+    cashboxRow.account_id = cashboxAccountId;
```
jniles: This isn't a cashbox.  It is supposed to be an expense account.  See point 1 in the issue.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,193 @@
...
+    cashboxRow.debit = vm.totalSelected;
+    cashboxRow.credit = 0;
+    rows.push(cashboxRow);
+
+
+    // then loop through each selected item and credit it with the convention account
+    invoices.forEach(invoice => {
+      const row = ToolKits.getBlankVoucherRow();
+
+      row.account_id = conventionAccountId;
```
jniles: This doesn't necessarily have to be a convention - any patient, including Patient Payant Cash should be able to be selected.  Please name these variables something else.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,193 @@
...
+  /* ================ End Invoice grid parameters ===================== */
+
+  // submission
+  function submit(form) {
+    if (form.$invalid) { return; }
+
+    const selected = vm.gridApi.selection.getSelectedRows();
+
+    const bundle = generateTransactionRows({
+      cashbox    : vm.cashbox,
```
jniles: There is no cashbox here.  This is mentioned in the issue:

> Instead of a cashbox, you must pick an expense account.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,193 @@
...
+      convention : vm.convention,
+      invoices   : selected,
+    });
+
+    Instance.close({
+      rows       : bundle,
+      convention : vm.convention,
+      type_id    : bhConstants.transactionType.CONVENTION_PAYMENT,
```
jniles: This isn't a CONVENTION_PAYMENT since it is an expense.  A better type would be [GENERIC_EXPENSE](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/constants/bhConstants.js#L70).

This is mentioned in the issue:
> The voucher type would be set to autre charge perte or something similar.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,193 @@
...
+      description : $translate.instant('VOUCHERS.GLOBAL.CONVENTION_PAYMENT_DESCRIPTION', {
```
jniles: This should have a description that matches what is happening. This isn't a CONVENTION_PAYMENT.  It is writing off a loss of an unrecoverable invoice.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.modal.html
```diff
@@ -0,0 +1,64 @@
...
+  </ol>
+</div>
+
+<form name="ToolForm" bh-submit="ToolCtrl.import(ToolForm)" novalidate>
+  <div class="modal-body">
+    
+    <bh-account-select
+      name="account_id"
+      account-id="ToolCtrl.account.id"
+      on-select-callback="ToolCtrl.onSelectAccount(account)">
```
jniles, line 15: This should be only expense accounts.  Can you use the [bhAccountSelect's `accountTypeId` option](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/components/bhAccountSelect.js#L11) to filter for only expense accounts?

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,193 @@
...
+  // expose to the view
+  vm.selectGroupInvoices = selectGroupInvoices;
+  vm.close = Instance.close;
+  vm.import = submit;
+
+  // Set up page elements data (debtor select data)
+  vm.onSelectDebtor = onSelectDebtor;
+  vm.onSelectAccount = onSelectAccount;
+
+  function onSelectDebtor(debtorGroup) {
```
jniles: I made a mistake in the issue - this shouldn't select a debtor group, but a debtor instead.  We will want to use `bhFindPatient` instead of convention, then show that patient's invoices.  I had forgotten that convention payment doesn't work this way, and instead has you select an entire debtor group.

jniles (changes requested):
This is about halfway there.  We need to update a lot of the text since it doesn't make sense.  The logic is good, but it doesn't actually provide the functionality we need yet

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,192 @@
...
+// Import transaction rows for a convention payment
+function PatientFledKitController(
+  Instance, DebtorGroup, Notify, Session, bhConstants, $translate, ToolKits,
+) {
+  const vm = this;
+
+  const { MAX_DECIMAL_PRECISION } = bhConstants.precision;
+
+  // global variables
+  vm.debtorGroupFilter = { is_convention : 1 };
```
jniles: Do you still need this?

client/src/i18n/en/vouchers.json
```diff
@@ -32,6 +32,8 @@
       "TRANSFER_DESCRIPTION": "Transfer: From {{ fromAccount }} to {{ toAccount }} for {{ amount }} {{ symbol }}",
       "CONVENTION_PAYMENT_DESCRIPTION": "Convention Payment: Payment for {{ numInvoices }} invoice(s) by {{ debtorGroupName }}.",
       "CONVENTION_INVOICES":"Convention - Invoices payment",
+      "WRITE_OFF_PATIENT_INVOICES":"Write off patient invoices",
```
jniles, line 35: Good text 👍

client/src/modules/vouchers/complex-voucher.html
```diff
@@ -20,6 +20,13 @@
               </a>
             </li>
 
+            
+            <li role="menuitem">
+              <a href data-action="patient-fled" ng-click="ComplexVoucherCtrl.openPatientFledModal()">
```
jniles, line 25: Do you use this "data-action" attribute?

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,192 @@
...
+  function onSelectAccount(account) {
+    vm.account.id = account.id;
+  }
+
+
+  // get debtor group invoices
+  function selectGroupInvoices(patient) {
+    vm.loading = true;
+
+    DebtorGroup.invoices(patient.debtor_uuid, { debtorByuuid : 1 })
```
jniles: It looks like we already have this functionality in the [DebtorService](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/debtors/debtor.service.js#L44).  Can you just use that instead of re-implementing it here?

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
jniles: Let's remove references of conventions and cashboxes from this PR.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
jniles: Can you remove references to cashbox/convention from this module?  It's really confusing to developers since you aren't actually using either cashboxes or conventions.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,192 @@
...
+
+  // submission
+  function submit(form) {
+    if (form.$invalid) { return; }
+
+    const selected = vm.gridApi.selection.getSelectedRows();
+
+    const bundle = generateTransactionRows({
+      cashbox    : vm.cashbox,
+      convention : vm.patient,
```
jniles: Same here.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
```diff
@@ -0,0 +1,192 @@
...
+      invoices   : selected,
+    });
+
+    Instance.close({
+      rows       : bundle,
+      convention : vm.patient,
+      type_id    : bhConstants.transactionType.GENERIC_EXPENSE,
+      description : $translate.instant('VOUCHERS.GLOBAL.WRITE_OFF_PATIENT_INVOICES_DESC', {
+        numInvoices : selected.length,
+        debtorGroupName : vm.patient.name,
```
jniles: This isn't a debtor group.  It's a patient.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.modal.html
```diff
@@ -0,0 +1,64 @@
...
+    <li class="title" translate>VOUCHERS.GLOBAL.IMPORT_TRANSACTION</li>
+  </ol>
+</div>
+
+<form name="ToolForm" bh-submit="ToolCtrl.import(ToolForm)" novalidate>
+  <div class="modal-body">
+    
+    <bh-account-select
+      name="account_id"
+      account-type-id="'5,6'"
```
jniles, line 13: Beautiful 👍

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.modal.html
```diff
@@ -0,0 +1,64 @@
...
+      account-id="ToolCtrl.account.id"
+      on-select-callback="ToolCtrl.onSelectAccount(account)">
+    </bh-account-select>
+
+    
+          <!--select the Patient -->
+    <bh-find-patient
+      suppress-reset="true"
+      on-search-complete="ToolCtrl.onSetlectPatient(patient)">
```
jniles: This has a typo - it should be "onSelectPatient(patient)"

server/controllers/finance/debtors/groups/index.js
```diff
@@ -276,9 +276,11 @@ function invoices(req, res, next) {
  */
 function loadInvoices(params) {
 
+  const { debtorByuuid } = params;
+  const key = debtorByuuid ? 'uuid' : 'group_uuid';
```
jniles: If you use the `DebtorService`, you won't need to do this and we won't need to write any more integration tests for this functionality.

jniles (commented):
@jeremielodi this is looking better.  You still use the words "convention" and "cashbox"  and "debtor group" everywhere in the code base, even though this module doesn't use them.  Can you change them please?

We also have a service for looking up debtor invoices by debtor uuid - it's the "DebtorService".  Go ahead and use that so you don't have to change the server.

client/src/modules/vouchers/toolkit/patient_fled/patient_fled.js
jeremielodi: I don't think so

client/src/modules/vouchers/complex-voucher.html
jeremielodi: It's will be important for the test

jniles (approved):
@jeremielodi, I've been looking for this functionality for a long time.  Great work.  I've tested it, and it works well.

bors r+

PR #4285, opened 2020-04-10
chore(inventory): default delay to 1

### LINKED ISSUE #4265: Set inventory delay to 1 by default
It looks like the basis of the CMM calculation relies on the `delay` to be a number great than 0.  [Here is the line of code](https://github.com/IMA-WorldHealth/bhima/blob/9ef833074037f8d200cf3926ae31d7c95f4b3f8a/server/controllers/stock/core.js#L349):
```js
    inventory.S_MAX = (CM * inventory.purchase_interval) + inventory.S_MIN; // stock maximum
    inventory.S_MONTH = Math.floor(inventory.quantity / CM_NOT_ZERO); // mois de stock
    inventory.S_Q = inventory.S_MAX - inventory.quantity; // Commande d'approvisionnement
    inventory.S_Q = inventory.S_Q > 0 ? parseInt(inventory.S_Q, 10) : 0;
    inventory.S_RP = inventory.quantity - (inventory.lifetime * CM); // risque peremption
```

By default, the inventory.delay is `0`, which means we will not have a CMM calculation.  I propose that we set the default value to `1`.  That way, we will have a CMM calculation for most inventory items, unless a user explicitly disables the value.

Basically we need to change [this line](https://github.com/IMA-WorldHealth/bhima/blob/master/server/models/schema.sql#L788):
```sql
  `delay` DECIMAL(10,4) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'Delivery time',
```
to
```sql
  `delay` DECIMAL(10,4) UNSIGNED NOT NULL DEFAULT 1 COMMENT 'Delivery time',
```
This would ensure that we always get a CMM calculation.

jniles:
Makes the default delay "1" instead of "0" to allow CMM calculations to be valid by default.

Closes #4265.

### DIFF
--- a/server/models/schema.sql
+++ b/server/models/schema.sql
@@ -785,7 +785,7 @@ CREATE TABLE `inventory` (
   `sellable` TINYINT(1)   NOT NULL DEFAULT 1,
   `note` text  NULL,
   `locked` TINYINT(1) NOT NULL DEFAULT 0,
-  `delay` DECIMAL(10,4) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'Delivery time',
+  `delay` DECIMAL(10,4) UNSIGNED NOT NULL DEFAULT 1 COMMENT 'Delivery time',
   `avg_consumption` DECIMAL(10,4) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'Average consumption' ,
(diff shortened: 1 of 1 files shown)


mbayopanda (approved):
LGTM 👍 

bors r+

PR #4288, opened 2020-04-10
fix(stock): movements report download crash

### LINKED ISSUE #4280: Stock Movements Registry download as XLSX/CSV crashes
I tried to download the stock movements registry as Excel and it crashed my server.  My only assumption is that it isn't respecting the filters.  Someone should investigate this on a live production database like HEV or IMCK.

_Edit:_ This might actually be because it is not respecting the filters and it is also rendering out as PDF.  The errors I get are puppeteer errors, strangely enough.

jniles:
This commit fixes the stock movements report download so that it doesn't crash anymore.  However, the renameKeys() functionality no longer works.

Closes #4280.


jeremielodi (approved):
@jeremielodi approved this PR

Issue #4233, opened 2020-03-30
The transaction edit modal doesn't handle delete errors


jniles:
This was observed in production on BHIMA version 1.10.1.

If a delete call fails, the transaction edit modal will not necessarily tell the user anything.  We should provide improved error handling so that the user knows what happened.

jniles:
@mbayopanda would you mind looking into this?

jniles:
Closed in #4291.

PR #4291, opened 2020-04-10
Handle error message on transaction deletion

mbayopanda:
This PR is about the issue #4233

### DIFF
--- a/client/src/js/components/bhModalNotify/bhModalNotify.html
+++ b/client/src/js/components/bhModalNotify/bhModalNotify.html
@@ -1,21 +1,4 @@
-<style>
-  .bh-modal-notify {
-    position: fixed;
-    top: 0px;
-    left: 0px;
-    width: 100%;
-    height: 53px;
-    display: flex;
-    align-items: center;
-    justify-content: space-between;
-    padding: 5px 10px;
-  }
-  .bh-modal-notify > i {
-    cursor: pointer;
-  }
-</style>
-
-<div ng-if="$ctrl.visible" class="bh-modal-notify {{$ctrl.style}}">
+<div bh-modal-notify  ng-if="$ctrl.visible" class="bh-modal-notify" ng-class="$ctrl.style">
   <span translate>{{ $ctrl.message }}</span>
   <i ng-click="$ctrl.close()" class="fa fa-times"></i>
 </div>
--- a/client/src/js/components/bhModalNotify/bhModalNotify.html
+++ b/client/src/js/components/bhModalNotify/bhModalNotify.html
@@ -1,4 +1,4 @@
-<div bh-modal-notify  ng-if="$ctrl.visible" class="bh-modal-notify" ng-class="$ctrl.style">
+<div ng-if="$ctrl.visible" class="bh-modal-notify" ng-class="$ctrl.style">
   <span translate>{{ $ctrl.message }}</span>
   <i ng-click="$ctrl.close()" class="fa fa-times"></i>
 </div>
--- a/client/src/js/components/bhModalNotify/bhModalNotify.html
+++ b/client/src/js/components/bhModalNotify/bhModalNotify.html
@@ -1,0 +1,21 @@
+<style>
+  .bh-modal-notify {
+    position: fixed;
+    top: 0px;
+    left: 0px;
+    width: 100%;
+    height: 53px;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    padding: 5px 10px;
+  }
+  .bh-modal-notify > i {
+    cursor: pointer;
+  }
+</style>
+
+<div ng-if="$ctrl.visible" class="bh-modal-notify {{$ctrl.style}}">
+  <span translate>{{ $ctrl.message }}</span>
+  <i ng-click="$ctrl.close()" class="fa fa-times"></i>
+</div>
--- a/client/src/js/components/bhModalNotify/bhModalNotify.js
+++ b/client/src/js/components/bhModalNotify/bhModalNotify.js
@@ -15,8 +15,9 @@ ModalNotifyController.$inject = ['$timeout'];
 
 function ModalNotifyController($timeout) {
   const $ctrl = this;
-  const TTL = $ctrl.ttl || 3000;
+  const DEFAULT_TTL = 3000;
   const ERR_TTL = 50000;
+  let timer;
 
   const options = {
     success : 'notification-success',
@@ -39,18 +40,19 @@ function ModalNotifyController($timeout) {
     $ctrl.style = $ctrl.error ? errorFormat : otherFormat;
     $ctrl.visible = true;
 
-    $timeout(() => {
+    timer = $timeout(() => {
       $ctrl.visible = false;
     }, ttl);
   }
 
   $ctrl.close = () => {
+    $timeout.cancel(timer);
     $ctrl.visible = false;
   };
 
   $ctrl.$onChanges = (changes) => {
     if (changes.value && changes.value.currentValue) {
-      setNotification(changes.value.currentValue, TTL);
+      setNotification(changes.value.currentValue, $ctrl.ttl || DEFAULT_TTL);
     }
 
     if (changes.error && changes.error.currentValue) {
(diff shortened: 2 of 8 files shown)


mbayopanda:
DO NOT MERGE THIS PR SINCE ITS BROKE THE DEFAULT BEHAVIOR OF THE GROWL COMPONENT

client/src/modules/journal/modals/editTransaction.modal.js
```diff
@@ -342,7 +344,8 @@ function JournalEditTransactionController(
             };
 
             Modal.close(deleteTransactionResult);
-          });
+          })
+          .catch(Notify.handleError);
```
jniles: Can you just log into a `<span class="text-danger"> SOMETHING BROKE!!!</span>`?

mbayopanda: I have created a customised one : 

image

mbayopanda:
@jniles it is done

mbayopanda:
@jniles this PR can be MERGE NOW

client/src/js/components/bhModalNotify/bhModalNotify.js
```diff
@@ -0,0 +1,60 @@
...
+      error : '<?',
+      ttl : '<?',
+    },
+  });
+
+ModalNotifyController.$inject = ['$timeout'];
+
+function ModalNotifyController($timeout) {
+  const $ctrl = this;
+  const TTL = $ctrl.ttl || 3000;
```
jniles: This should be in an `$onInit()` function.

client/src/js/components/bhModalNotify/bhModalNotify.html
```diff
@@ -0,0 +1,21 @@
...
+    align-items: center;
+    justify-content: space-between;
+    padding: 5px 10px;
+  }
+  .bh-modal-notify > i {
+    cursor: pointer;
+  }
+</style>
+
+<div ng-if="$ctrl.visible" class="bh-modal-notify {{$ctrl.style}}">
```
jniles: It's more the angularjs way to use `ngClass`.  Could you do that here?

client/src/js/components/bhModalNotify/bhModalNotify.html
```diff
@@ -0,0 +1,21 @@
+<style>
+  .bh-modal-notify {
```
jniles: I can tell you've been writing VueJS.  Can you just put this in our bhima-bootstrap stylesheet?

client/src/js/components/bhModalNotify/bhModalNotify.js
```diff
@@ -0,0 +1,60 @@
...
+    $ctrl.style = $ctrl.error ? errorFormat : otherFormat;
+    $ctrl.visible = true;
+
+    $timeout(() => {
+      $ctrl.visible = false;
+    }, ttl);
+  }
+
+  $ctrl.close = () => {
+    $ctrl.visible = false;
```
jniles, line 49: You could also cancel the `$timeout` here just to clean some things up.

jniles (commented):
I've tested this and it works.  I'm a little torn on introducing a new UI pattern.  Do you think it is a good idea @mbayopanda?

I've made some clean up comments, mostly to make this more AngularJS-like.  Would it be possible to write some basic unit tests for this component?

mbayopanda:
@jniles since we don't change the old behavior of Notify and the current notify component is made for modal first, do you think it is a new UI pattern ?

client/src/js/components/bhModalNotify/bhModalNotify.html
mbayopanda: I was thinking about that but ngClass allow value in this format : 
```html
ng-class="{'the_css_class': javascript_variable }"
```

Here in the `bhModalNotify` is not known in advance

jniles:
Yes, a user has never had to click the top of a modal before, not have they ever seen an error message appear in the top of a modal.  Normally error messages appear: 
 1. At the top of the whole web page
 2. Right below an input that has a problem.
 3. At the base of the form that has the problem.

I don't think we've ever displayed an error message anywhere else.

mbayopanda:
@jniles could i get review and indication for a better client unit test

client/src/js/components/bhModalNotify/bhModalNotify.js
```diff
@@ -0,0 +1,62 @@
+angular.module('bhima.components')
+  .component('bhModalNotify', {
+    templateUrl : 'js/components/bhModalNotify/bhModalNotify.html',
+    controller  : ModalNotifyController,
+    transclude  : true,
+    bindings    : {
+      value : '<?',
```
jniles, line 6: What does `value` do?

jniles (commented):
@mbayopanda I don't see any client unit tests at the moment.

The scenarios I would test are:
  1. Ensure that no error message is present if the `error` is undefined.
  2. Ensure that an error message is present when the `error` is defined.
  3. Ensure that the error message is removed when the "x" is clicked.
  4. Ensure that the error message is removed with the TTL expires.

client/src/js/components/bhModalNotify/bhModalNotify.js
```diff
@@ -0,0 +1,62 @@
+angular.module('bhima.components')
+  .component('bhModalNotify', {
+    templateUrl : 'js/components/bhModalNotify/bhModalNotify.html',
+    controller  : ModalNotifyController,
+    transclude  : true,
```
jniles: Do you need transclude?  It doesn't look like you actually use it.

jniles (commented):
@mbayopanda thanks for working with me on the unit tests.

As discussed, please look at the `type` binding.  I don't think it is used.  Similarly, `$ctrl.status` isn't defined.  I think you might have gotten confused and they should be the same thing.

Also, please remove the `transclude` if it is unused.

If you can think of any more unit tests to write, go for it!

mbayopanda:
@jniles could you give me a review please

client/src/js/components/bhModalNotify/bhModalNotify.js
jniles: This is good.  Well done!

karma.conf.js
```diff
@@ -18,6 +18,7 @@ module.exports = (config) => {
       'test/client-unit/mocks/*.js',
       `bin/client/js/bhima/bhima.min.js`,
       'bin/client/modules/**/*.html',
+      'bin/client/js/components/**/*.html',
```
jniles, line 21: 👍

test/client-unit/components/bhModalNotify.spec.js
```diff
@@ -0,0 +1,181 @@
...
+  beforeEach(module('pascalprecht.translate', 'templates', 'bhima.services', 'bhima.components', 'ui.bootstrap'));
+
+  let $scope;
+  let $compile;
+  let $translate;
+  let element;
+
+  let $flushPendingTasks;
+
+  const find = (elm, selector) => elm[0].querySelector(selector);
```
jniles, line 20: Since we've used this in a number of other unit tests, we could probably make a shared utility file to put these in.  I can't remember if we are allowed to `require()` in this context or not though.

test/client-unit/components/bhModalNotify.spec.js
```diff
@@ -0,0 +1,181 @@
...
+  it('shows danger message with danger style when the type is given', () => {
+    const componentTemplate = `
+      <bh-modal-notify type="danger" value="someValue"></bh-modal-nofify>
+    `;
+    const compiled = $compile(angular.element(componentTemplate))($scope);
+    $scope.someValue = 'Danger Message';
+    const translated = $translate.instant($scope.someValue);
+    $scope.$digest();
+
+    const notificationClass = find(compiled, '.notification-danger');
```
jniles, line 164: These tests are great!  I think they are fine as they are.  One potential problem is that we may never find the `.notification-danger` class if a bug occurs where danger becomes success.  A way to test for this would be to instead search for the `.notification` class, then expect it to also have `.notification-danger`, like we do for the `fa-times` class.

```js
const notification  = find(compiled, '.notification');
expect(angular.element(notification).hasClass('notification-danger')).to.equal(true);
```

jniles (approved):
Looks good to me!  Nice additional tests :)

bors r+

jniles:
Looks like Travis is down for maintenance.

jniles:
Looks like a fluke.

Bors r+

PR #4299, opened 2020-04-14
improvement(Stock Value Report)

### LINKED ISSUE #3963: Add an option to Stock Value Report to hide stock with 0 value
The stock value report shows 0 value for some items in a depot.  I'm not sure if this is useful or not.  I think we should add an option to hide stock with "0" value.  See an example below:

neL6Qrh9FZ

lomamech:
- Adding option to exclude inventory where stock value is 0

closes #3963


lomamech:

jniles (approved):
LGTM. 

bors r+

PR #4312, opened 2020-04-16
Improvements to inventory changes report

### LINKED ISSUE #4287: Improvements to the Inventory Changes Report
The inventory changes report implemented in https://github.com/IMA-WorldHealth/bhima/pull/4002 could use some updates to make it more usable.  Here is a screenshot of a production changelog.

image
_Fig 1: Changes in Production_

There are a couple improvements that could be made:

1. Items are not ordered chronologically.  They should list the oldest change first to the newest change last for each item.  That way, you can track the progression of changes easier.
2. "Designation" seems to be used a lot without modification - this is probably a bug.  For some reason we are recording changes it in when there aren't any actual changes.  It also looks like for real changes, this isn't working as expected.
3. This is a report to print - let's make all the fonts the same size and put it in a table.  There is a lot of wasted space.  See below:
image

jniles:
This Pull Request refactors the SQL and presentation of in the inventory changes report.  The presentation is now an HTML table that is more amenable to printing.  Here is an example:
image

The inventory items are first ordered by inventory group, then by the current label of the inventory item.  The number of changes made to the inventory is shown in the header for each item.

At the bottom of the page, we provide an overview of the users who made the most modifications.  This will help an administrator easily detect if someone who was not authorized to make a modification made it.

Here is what that looks like:
image

Closes #4287.

### DIFF
--- a/client/src/i18n/en/inventory.json
+++ b/client/src/i18n/en/inventory.json
@@ -39,6 +39,9 @@
         "INVENTORY_PRICE_INFO" : "inventory_price : the sale price of the inventory in the enterprise currency",
         "COLUMNS_INFO" : "Note about some columns",
         "SELLABLE" : "Sellable",
-        "VIEW_ARTICLES_IN_STOCK" : "View Items in Stock"
+        "VIEW_ARTICLES_IN_STOCK" : "View Items in Stock",
+        "PREVIOUS_VALUE" : "Previous Value",
+        "UPDATED_VALUE" : "Updated Value",
+        "CHANGES" : "Changes"
     }
 }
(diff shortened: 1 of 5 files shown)


jeremielodi:
try to the conflic so that I could give you a review

jniles:
@jeremielodi done

jniles:
@jeremielodi what is the word on this PR?

jeremielodi:
ok let me give you a review

jeremielodi (approved):
I have not been able to fetch your branch and test locally, I connection is bad today for me, but your code looks perfect.
Good work @jniles

PR #4331, opened 2020-04-20
feature(StockMovementToPatient)

### LINKED ISSUE #4309: Link the patient from the "Stock Movement Registry" if the entry is from a patient or the distribution is to a patient
It would be useful to link to a patient from the Stock Movement Registry if the flux is to a patient or from a patient. 

It would look like this:
image

lomamech:
- Add link from stock Movement to Patient

closes #4309

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -270,7 +270,7 @@ function getLotsMovements(depotUuid, params) {
       m.flux_id, BUID(m.entity_uuid) AS entity_uuid, m.unit_cost,
       f.label AS flux_label, i.delay,
       iu.text AS unit_type,
-      dm.text AS documentReference
+      dm.text AS documentReference, pa.display_name AS patient_name
     FROM stock_movement m
     JOIN lot l ON l.uuid = m.lot_uuid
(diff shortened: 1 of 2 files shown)


lomamech:
g0A7K9vTc3

lomamech:

server/controllers/stock/core.js
```diff
@@ -279,6 +279,7 @@ function getLotsMovements(depotUuid, params) {
     JOIN flux f ON f.id = m.flux_id
     LEFT JOIN document_map dm ON dm.uuid = m.document_uuid
     LEFT JOIN service AS serv ON serv.uuid = m.entity_uuid
+    LEFT JOIN patient AS pa ON pa.uuid = m.entity_uuid
```
jniles: Can you instead use the `entity_uuid` to filter on patient?  That would save a join.  I'm pretty sure the `entity_uuid` here is actually the `patient.uuid`.

lomamech: Okay

lomamech:
@jniles it's done

jniles (approved):
Beautiful. 

bors r+

PR #4336, opened 2020-04-21
Feat(Stock report) stock consumption graph report

### LINKED ISSUE #4330: Create a chart of stock consumed by inventory by period
Now that we have a lot of raw data in our stock module, we could start producing graphs based on it.  By downloading the latest Vanga dataset, I generated this graph in Excel:

image
_Fig 1: Stock Consumption of Amoxyciline by day for March 2020_

This graph shows the value of amoxyciline distributed to patients (Y-axis) by the day of the month (X-axis) for March 2020.  Therefore, you can see how much amoxyciline has been consumed over the month and easily spot outliers.

It would be nice to have this kind of report as a dashboard that you can configure on client.  Whoever wants to tackle this issue should propose the way it would look after studying the problem so we can come up with the best solution.

The things you would need to filter are:
  - What period?  Jan 2020, Feb 2020, etc..
  - What inventory item? 
  - (optionally) what depot?

jeremielodi:
closes #4330


jeremielodi:
stock_report

client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
```diff
@@ -0,0 +1,64 @@
...
+            inventory-uuid="ReportConfigCtrl.inventory.uuid"
+            on-select-callback="ReportConfigCtrl.onSelectInventory(inventory)"
+            required="true"
+            only-consumable="true">
+            <bh-clear on-clear="ReportConfigCtrl.clear('inventory_uuid')"></bh-clear>
+          </bh-inventory-select>
+
+
+          <!-- date interval -->
+          <bh-date-interval
```
jniles: Let's make this select an entire period instead of between two dates.  That will make the logic a lot easier to maintain and ensure that we always have the same size data.

Also, since most of the stock indicators are calculated based on monthly indicators, it makes sense in that respect.

gulpfile.js/client-static.js
```diff
@@ -13,13 +14,28 @@ const STATIC_PATHS = [
   '!client/src/i18n/{en,fr}/*.json',
 ];
 
+
+const CHART_JS = [
+  'node_modules/jquery/dist/jquery.min.js',
+  'node_modules/chart.js/dist/Chart.min.js',
```
jniles: Why do you need jquery?  It doesn't look like you ever use it.

gulpfile.js/client-static.js
```diff
@@ -13,13 +14,28 @@ const STATIC_PATHS = [
...
 // collects all static files from the client (BHIMA src and vendor files)
 // and moves them to the build folder, respecting folder structure
 function buildStatic() {
   return src(STATIC_PATHS)
     .pipe(dest(CLIENT_FOLDER));
 }
 
+// some report will need only chartjs and jquery
+function buildStaticChartjs() {
+  return src(CHART_JS)
```
jniles: IF we are only using this on the server, let's just link directly into the node_modules folder from the handlebars template.

Otherwise, if we plan to use this on both the client and server, let's put it in the regular JS paths.

server/controllers/stock/reports/stock/consumption_graph.js
```diff
@@ -0,0 +1,87 @@
+const {
+  _, db, util, ReportManager, pdfOptions, STOCK_CONSUMPTION_GRAPTH_TEMPLATE,
+} = require('../common');
+
+const FilterParser = require('../../../../lib/filter');
+/**
+   * @method stockEntryReport
```
jniles, line 8: This isn't the stockEntryReport.

server/controllers/stock/reports/stock/consumption_graph.js
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  // set up the report with report manager
+  try {
+    report = new ReportManager(STOCK_CONSUMPTION_GRAPTH_TEMPLATE, req.session, optionReport);
+  } catch (e) {
+    return next(e);
+  }
+
+  const sql = `
```
jniles: Would it be possible to just use the stock functions that already exist in our stock modules for finding lots and consumption?  I don't want there to be a different in logic between this SQL and those SQLs.  Plus, this doubles the chance of bugs.

server/controllers/stock/reports/stock/consumption_graph.js
```diff
@@ -0,0 +1,87 @@
...
+  const rqtSQl = filters.applyQuery(sql);
+  const rqtParams = filters.parameters();
+
+  const bundel = {
+    labels : [],
+    data : [],
+  };
+
+  let inventory = {};
+  return db.one(inventorySql, filterObtions.inventory_uuid).then(invent => {
```
jniles: Could you use async/await here to clean up this nested structure?

client/src/i18n/en/report.json
```diff
@@ -141,6 +141,10 @@
       "DESCRIPTION": "This report shows the current stock value for each inventory in a depot",
       "EXCLUDE_INVENTORIES_ZERO_VALUE": "Exclude inventories with zero stock value"
     },
+    "STOCK_CONSUMPTION_GRAPH_REPORT" : {
+      "TITLE" : "Stock Consumption's graph report",
```
jniles, line 145: This could be something like "Stock Consumption from {{depotName}} during {{periodLabel}}".

client/src/i18n/en/report.json
```diff
@@ -141,6 +141,10 @@
...
+      "DESCRIPTION" : "This report graphically shows the consumption of stock during a period for a depot or inventory"
```
jniles, line 146: Does it show it by quantity consumed?  Or value of the stock?  Or what is it referencing?

jniles (commented):
@jeremielodi, I've made some suggestions.  Thank you for the screenshot and taking a look at this.

As stated in the issue, this should select a period as the first option.  Here is where it states it:
> What period? Jan 2020, Feb 2020, etc..

It would be nice to have labels on the y-axis.  What is the y-axis?

I can't tell what depot you selected.  What depot is this?  Could you put it on the chart somehwere?  Like maybe in the title?

I'm not sure why you needed jquery.  If you don't use it, can you remove it?  It doesn't look like chart.js depends on it.

gulpfile.js/client-static.js
jeremielodi: Good Idea, from node_modules is very easy. I did'nt think about that

server/controllers/stock/reports/stock/consumption_graph.js
jeremielodi: No problem.

jeremielodi:
statis

server/models/bhima.sql
```diff
@@ -149,8 +149,9 @@ INSERT INTO unit VALUES
   (264, 'Configurable Analysis Report','TREE.CONFIGURABLE_ANALYSIS_REPORT','Configurable Analysis Report', 144,'/modules/reports/configurable_analysis_report','/reports/configurable_analysis_report'),
   (265, 'Purchase order analysis', 'TREE.PURCHASE_ORDER_ANALYSIS', 'Purchase order analysis', 144, '/modules/reports/purchaseOrderAnalysis', '/reports/purchaseOrderAnalysis'),
   (266, 'Inventory Changes Report', 'REPORT.INVENTORY_CHANGE.TITLE', 'Inventory Changes Report', 144, '/modules/reports/inventoryChanges', '/reports/inventoryChanges'),
-  (267, 'Monthly consumption report', 'TREE.MONTHLY_CONSUMPTION', 'Monthly consumption report', 144, '/modules/reports/monthlyConsumptionReport', '/reports/monthlyConsumptionReport');
-
+  (267, 'Monthly consumption report', 'TREE.MONTHLY_CONSUMPTION', 'Monthly consumption report', 144, '/modules/reports/monthlyConsumptionReport', '/reports/monthlyConsumptionReport'),
+  (268, '[Stock] Consumption graph','TREE.STOCK_CONSUMPTION_GRAPH_REPORT','Stock Consumption graph report', 144,'/modules/reports/generated/stock_consumption_graph_report','/reports/stock_consumption_graph_report');
```
mbayopanda, line 153: Could you add this entry in the migration file

server/models/bhima.sql
```diff
@@ -212,7 +213,8 @@ INSERT INTO `report` (`report_key`, `title_key`) VALUES
   ('configurable_analysis_report', 'REPORT.CONFIGURABLE_ANALYSIS_REPORT.TITLE'),
   ('purchaseOrderAnalysis', 'REPORT.PURCHASE_ORDER_ANALYSIS.TITLE'),
   ('inventoryChanges', 'REPORT.INVENTORY_CHANGE.TITLE'),
-  ('monthlyConsumptionReport', 'REPORT.MONTHLY_CONSUMPTION.TITLE');
+  ('monthlyConsumptionReport', 'REPORT.MONTHLY_CONSUMPTION.TITLE'),
+  ('stock_consumption_graph_report', 'REPORT.STOCK_CONSUMPTION_GRAPH_REPORT.TITLE');
```
mbayopanda, line 217: Could you add this one also in the migration file

jeremielodi:
@jniles and @mbayopanda  this PR is once again ready for review

jniles:
I think this was closed by accident.

client/src/i18n/en/report.json
jniles: Could you make this modification?  Or did you disagree with it?

client/src/i18n/en/report.json
jniles: Can you add in text about what is being grouped to this description?

client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
```diff
@@ -0,0 +1,64 @@
...
+            <bh-clear on-clear="ReportConfigCtrl.clear('inventory_uuid')"></bh-clear>
+          </bh-inventory-select>
+
+          
+          <bh-fiscal-year-select
+            fiscal-id="ReportConfigCtrl.reportDetails.fiscal_id"
+            on-select-fiscal-callback="ReportConfigCtrl.onSelectFiscalYear(fiscalYear)">
+          </bh-fiscal-year-select>
+
+          <bh-period-selection
```
jniles, line 49: Just FYI - we have the [`bhFiscalYearPeriodSelect`](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/components/bhFiscalYearPeriodSelect/bhFiscalYearPeriodSelect.js) that groups these two (fiscal year and period) selection. It might make your logic easier in the future to just use that.

server/controllers/stock/core.js
```diff
@@ -403,6 +403,50 @@ function getStockConsumption(periodIds) {
...
+    JOIN lot l ON l.uuid = m.lot_uuid
+    JOIN inventory i ON i.uuid = l.inventory_uuid
+    JOIN depot d ON d.uuid = m.depot_uuid
+  `;
+
+  filters.dateFrom('dateFrom', 'date');
+  filters.dateTo('dateTo', 'date');
+  filters.custom('depot_uuid', 'd.uuid=?');
+  filters.custom('inventory_uuid', 'i.uuid=?');
+  filters.custom('consumption', '(m.flux_id IN (9, 10) OR (m.flux_id = 8 AND d.is_warehouse = 1))');
```
jniles, line 442: This is cool 👍

jeremielodi:
Yeah, you're right

server/controllers/stock/core.js
```diff
@@ -403,6 +403,50 @@ function getStockConsumption(periodIds) {
...
+        d.text AS depot_name, d.uuid AS depot_uuid
+    FROM stock_movement m
+    JOIN lot l ON l.uuid = m.lot_uuid
+    JOIN inventory i ON i.uuid = l.inventory_uuid
+    JOIN depot d ON d.uuid = m.depot_uuid
+  `;
+
+  filters.dateFrom('dateFrom', 'date');
+  filters.dateTo('dateTo', 'date');
+  filters.custom('depot_uuid', 'd.uuid=?');
```
jniles: You actually don't need to use custom here - you can just use `equals`.  Like this:

```js
filters.equals('depot_uuid', 'uuid', 'd');
```

server/controllers/stock/reports/stock/consumption_graph.js
```diff
@@ -0,0 +1,73 @@
...
+   * GET /reports/stock/consumption_graph
+   */
+async function stockConsumptionGrathReport(req, res, next) {
+  try {
+
+    const params = _.clone(req.query);
+
+    const optionReport = _.extend(params, pdfOptions, {
+      filename : 'REPORT.STOCK_CONSUMPTION_GRAPH_REPORT.TITLE',
+      waitFor : 1500,
```
jniles: What is waitFor?

server/controllers/stock/reports/stock/consumption_graph.js
jniles: Please update these comments.  They are not true.

server/controllers/stock/core.js
```diff
@@ -403,6 +403,50 @@ function getStockConsumption(periodIds) {
...
+ *
+ * @description returns the daily (periodic) stock consumption (CM)
+ *
+ * @param {array} periodIds
+ */
+async function getDailyStockConsumption(params) {
+  params.consumption = true;
+
+  if (params.period_id) {
+    const period = await db.one('SELECT start_date,end_date FROM period WHERE id=?', params.period_id);
```
jniles, line 419: Oh la la.  We don't have the period_id in stock_movements?  That is a problem.

Made an issue: https://github.com/IMA-WorldHealth/bhima/issues/4347

server/controllers/stock/reports/stock_consumption_graph.handlebars
```diff
@@ -0,0 +1,80 @@
...
+                        backgroundColor: 'rgb(255, 99, 132)',
+                        borderColor: 'rgb(255, 99, 132)',
+                        data: {{{data}}}
+                    }]
+                },
+
+                // Configuration options go here
+                options: {
+                    animation: {
+                        duration : 0,
```
jniles: This is a good start. Let's disable all animations.  The documentation can be found [here](https://www.chartjs.org/docs/latest/general/performance.html#disable-animations):
```js
    options: {
        animation: {
            duration: 0 // general animation time
        },
        hover: {
            animationDuration: 0 // duration of animations when hovering an item
        },
        responsiveAnimationDuration: 0 // animation duration after a resize
    }
```

jniles (commented):
I've run this on the latest Vanga dataset.  Here is an example for March:
image

This is starting to look good.

It would be nice to have data labels above each bar.  To my knowledge, chartjs doesn't do this by default, but [there is a plugin for it](https://github.com/chartjs/chartjs-plugin-datalabels).  Could you try to get the plugin to work?  I am looking for labels above each bar, like this:  https://chartjs-plugin-datalabels.netlify.app/samples/advanced/custom-labels.html

There are quite a few comments from the last review that were ignored.  Please give those another look.

server/controllers/stock/reports/stock/consumption_graph.js
jeremielodi: It's useful when the chart as animation, it takes time for being drawn. as I set the animation duration to zero, we can skip this

jniles: That makes sense.   Where do you use it?  I can't find any other reference to it in your code.

client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
jeremielodi: As the depot selection is optional we can keep ```Stock Consumption's graph report``` as title

jeremielodi:
shart-label

client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
jniles: I think this response was to another comment.  That makes sense.  I guess I didn't realize the depot was optional, but that makes sense from the code now.

package.json
```diff
@@ -64,6 +64,8 @@
     "angular-ui-router": "angular-ui/angular-ui-router-bower#1.0.0-rc.1",
     "body-parser": "^1.18.3",
     "bootstrap": "^3.3.0",
+    "chart.js": "^2.9.3",
+    "chartjs-plugin-datalabels": "^0.7.0",
```
jniles, line 68: 👍

server/models/migrations/next/migrate.sql
```diff
@@ -0,0 +1,9 @@
+
+-- @jeremielodi 23-04-2020
```
jniles, line 2: 👍

jniles (approved):
This is a good initial implementation.  I think it is probably good enough to merge and start getting feedback from out clients about.  I like that you don't have to specify a depot.

Here is a screenshot with IMCK's data for Feb 2020:
image

bors r+

PR #4348, opened 2020-04-23
feature(Stock Movement Registry)

### LINKED ISSUE #4318: Link to Articles in Stock from the Stock Movements Registry
It would be really nice if we could check the quantity in stock in a depot by clicking on the movement link in the depot.  We currently do not have that option.  Let's put an option on the dropdown menu to allow us to link to the Articles in Stock.  It should filter on the lot and the depot.

image

lomamech:
- make link betwen stock Movement registry to Articles In stock

closes #4318

### DIFF
--- a/client/src/modules/stock/movements/templates/action.cell.html
+++ b/client/src/modules/stock/movements/templates/action.cell.html
@@ -15,8 +15,11 @@
     </li>
 
     <li>
-      <a data-method="view-patient" href
-        ui-sref="stockInventories({ filters : [{ key : 'period', value : 'allTime', includeEmptyLot: 0 }, { key : 'depot_uuid', value : row.entity.depot_uuid }, { key : 'inventory_uuid', value : row.entity.inventory_uuid }]})">
+      <a data-method="view-stock-lot" href
+        ui-sref="stockInventories({ filters : [
+          { key : 'period', value : 'allTime', includeEmptyLot: 0 },
+          { key : 'depot_uuid', value : row.entity.depot_uuid, displayValue: row.entity.depot_text },
+          { key : 'inventory_uuid', value : row.entity.inventory_uuid, displayValue: row.entity.text }]})">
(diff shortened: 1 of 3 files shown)


lomamech:
9E4HFjnwyL

lomamech:

client/src/modules/stock/movements/templates/action.cell.html
```diff
@@ -14,6 +14,13 @@
       </a>
     </li>
 
+    <li>
+      <a data-method="view-patient" href
```
jniles: This "data-method" should be "view-stock-lot" or "view-article" or something.

client/src/modules/stock/movements/templates/action.cell.html
```diff
@@ -14,6 +14,13 @@
...
+        ui-sref="stockInventories({ filters : [{ key : 'period', value : 'allTime', includeEmptyLot: 0 }, { key : 'depot_uuid', value : row.entity.depot_uuid }, { key : 'inventory_uuid', value : row.entity.inventory_uuid }]})">
```
jniles: Could you change:
```js
 { key : 'depot_uuid', value : row.entity.depot_uuid },
```
to:
```js
 { key : 'depot_uuid', value : row.entity.depot_uuid, displayValue: row.entity.depot_text },
```
This will make the filter show the depot name on the target page.

jniles: Could you also change:
```js
{ key : 'inventory_uuid', value : row.entity.inventory_uuid }
```
to:
```js
{ key : 'inventory_uuid', value : row.entity.inventory_uuid, displayValue: row.entity.text }
```
This will show the inventory label in the filter.

jniles (commented):
@lomamech behavior is good.

I've commented to make the filters show the label rather than the UUID on the target page.  It should improve the user experience.

lomamech:
@jniles it's done

jniles (approved):
Je confirme que ca marche correctement.

image


bors r+

PR #4352, opened 2020-04-23
bug(Lot in Stock)

### LINKED ISSUE #4345: Lots in Stock throws an error if no lots have been added to the database
The Lots in Stock registry throws an error if there are no lots found, rather than telling the user that there is no lots in stock found. This was reported by AG Neron.  See below:

Hlh3FE18WF
_GIF 1: Default Broken Lots Registry_

It should behave like the Stock Movements registry which shows "No records found" if there is no information found.

lomamech:
- Prevent error in Lot in stock, when db.one request query empty tables

closes #4345


lomamech:
lgLh5VCOee

lomamech:
@jniles it's done

jniles (approved):
Good enough for me.

bors r+

PR #4355, opened 2020-04-24
refractor(employee registry) fix link to patient data

### LINKED ISSUE #4313: "Hospital Data" link on the Employee Registry is broken
It looks like the "hospital data" link in the employee registry broke at some point.  See below:
6ueofiyZtp

I am pretty sure this is supposed to link to the patient registry.

jeremielodi:
closes #4313

### DIFF
--- a/client/src/js/components/bhPatientFinancialActivity.js
+++ b/client/src/js/components/bhPatientFinancialActivity.js
@@ -8,7 +8,7 @@ angular.module('bhima.components')
   });
 
 PatientFinancialActivityCtrl.$inject = [
-  'PatientService', 'moment', 'SessionService', 'bhConstants',
+  'PatientService', 'moment', 'SessionService', 'bhConstants', '$q',
 ];
 
 /**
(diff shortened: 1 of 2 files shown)


client/src/js/components/bhPatientFinancialActivity.js
```diff
@@ -71,8 +71,6 @@ function PatientFinancialActivityCtrl(Patients, moment, Session, Constants) {
 
         $ctrl.dataMovement = dataMovement;
         $ctrl.noStockMovement = stockData.length === 0;
-      })
-      .finally(() => {
```
jniles, line 75: Aah, I see you noticed the loading bug.  The actual problem isn't here.   If you want to fix it, change the `Promise.all()` at the beginning of this chain  to `$q.all()`.

AngularJS doesn't run the $digest loop after the `Promise()` but does after `$q()`.

jniles (commented):
Looks good!  Just one comment on the $loading bit.

jniles (approved):
Thanks @jeremielodi !  Looks great to me.

bors r+

PR #4358, opened 2020-04-25
perf(bhUnique): cache validator results for 15s

### LINKED ISSUE #3427: Cache unique validator results
The unique validator sends HTTP requests to check if a field is unique in the database.  This can be a performance hit on heavy tables (e.g. hospital_no).  We should cache the results of bhUniqueValidator, with a developer-configurable timeout so that we don't hit the patient table a million times.

Ideally, this would be another binding on the component border where we say `bh-unique-cache-timeout="12345"` to make it cache results for 12345.

If you want to work on this, you can probably use the `HttpCache` directly.

jniles:
Improves system performance by caching the results of the bhUnique async validator for 15 seconds to prevent sending too many HTTP queries.  Leverages the HttpCache service to do so.

Here is what it looks like in action.  Normally, every keystroke would result in a new HTTP request.  In this case, only the new unique values result in a new `$http` request as shown in the console.
vkhsPHvsNs

Closes #3427.

### DIFF
--- a/client/src/js/services/UniqueValidatorService.js
+++ b/client/src/js/services/UniqueValidatorService.js
@@ -1,7 +1,7 @@
 angular.module('bhima.services')
-.service('UniqueValidatorService', UniqueValidatorService);
+  .service('UniqueValidatorService', UniqueValidatorService);
 
-UniqueValidatorService.$inject = ['$http', 'util'];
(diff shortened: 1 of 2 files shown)


jeremielodi (approved):
LGTM

PR #4359, opened 2020-04-25
chore: remove unused inventory store.

jniles:
We don't use it, so I've removed it.


lomamech (approved):
No comment :white_check_mark:

PR #4364, opened 2020-04-26
fix(debtors): ensure longest description shown

### LINKED ISSUE #4361: Show full invoice description instead of "Rounding Exchange Rate" in Patient Financial Activity
I came across a confusing thing during my analysis of the IMCK database.  There was an invoice that had a rounding of 88$, which led me to believe we had a rounding bug.   Here is what it looks like:
image

This actually is not a rounding bug - if the patient is referenced multiple times in the same transaction, we randomly choose which text to show in the group by.

Here is the full transaction:
image

So, rather than saying **Rounding Exchange Rate**, it would be better to show **Facture de Vicky Tshiyoyo Kena (PA.PAX.3957) pour 17 items dans le service Dispensaire Pax**.  This would make our reports look much more intelligible.

jniles:
This commit ensures that the longest description available to the user is displayed on reports such as the patient financial activity.  The longest description usually has the most detail.

Closes #4361.

**Before**
image

**After**
image

### DIFF
--- a/server/controllers/finance/debtors/index.js
+++ b/server/controllers/finance/debtors/index.js
@@ -10,13 +10,11 @@
 *
 * @module controllers/finance/debtors
 *
-* @requires q
+* @requires lodash
 * @requires lib/db
 * @requires lib/errors/NotFound
-* @requires lib/errors/BadRequest
 */
 
-const q = require('q');
 const _ = require('lodash');
 const db = require('../../../lib/db');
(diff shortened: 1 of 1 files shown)


mbayopanda (approved):
@jniles the code looks good to me, maybe we need this feature also in the `Simple Account Report` to.

bors r+

mbayopanda:
bors try

jniles:
bors try

PR #4370, opened 2020-04-27
feat(Pateint record) add link to stock movement distributed to patient

### LINKED ISSUE #4300: Add link to view all stock movements of a patient to the "Financial Situation" area
After https://github.com/IMA-WorldHealth/bhima/pull/4269 landed, we now have a really cool dashboard to show stock distributed to the patient.  It looks like this:

image
_Fig: Stock Movements given to Patient_

We limit this the 5 most recent movements.  It would be nice to add a link directly to the dashboard so that we can see all the stock movements in the registry tied to patient.

jeremielodi:
closes #4300


jniles (approved):
LGTM.

bors r+

PR #4371, opened 2020-04-27
feat(employee componet) add employee reference

### LINKED ISSUE #4357: Add the employee reference to the Prise En Charge voucher tool
It would be helpful to have the reference of the employee on the prise en charge voucher tool.  That would make it easier to look them up from a report.

image

jeremielodi:
closes #4357

### DIFF
--- a/client/src/modules/templates/bhEmployeeSelect.tmpl.html
+++ b/client/src/modules/templates/bhEmployeeSelect.tmpl.html
@@ -19,7 +19,10 @@
       ng-required="$ctrl.required"
       on-select="$ctrl.onSelect($item)">
       <ui-select-match placeholder="{{ 'FORM.SELECT.EMPLOYEE' | translate }}"><span>{{$select.selected.display_name}}</span></ui-select-match>
-      <ui-select-choices ui-select-focus-patch repeat="employee.uuid as employee in $ctrl.employees | filter: { 'display_name': $select.search }">
+      <ui-select-choices 
+        ui-select-focus-patch 
+        repeat="employee.uuid as employee in $ctrl.employees | filter: { 'display_name': $select.search, 'reference' :  $select.search}">
+        <strong ng-bind-html="employee.reference | highlight:$select.search"></strong> - 
         <span ng-bind-html="employee.display_name | highlight:$select.search"></span>
       </ui-select-choices>
     </ui-select>


client/src/modules/templates/bhEmployeeSelect.tmpl.html
```diff
@@ -20,7 +20,7 @@
       on-select="$ctrl.onSelect($item)">
       <ui-select-match placeholder="{{ 'FORM.SELECT.EMPLOYEE' | translate }}"><span>{{$select.selected.display_name}}</span></ui-select-match>
       <ui-select-choices ui-select-focus-patch repeat="employee.uuid as employee in $ctrl.employees | filter: { 'display_name': $select.search }">
-        <span ng-bind-html="employee.display_name | highlight:$select.search"></span>
+        <span> ({{employee.reference}}) {{employee.display_name}}</span>
```
jniles: This removes the highlighting.  It also doesn't filter on employee reference as you type.

Could you instead make an `hrLabel` similar to how the bhAccountSelect works?  For reference, [here it is](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/templates/bhAccountSelect.tmpl.html#L25).

That way we preserve highlighting on the name and reference.

jniles (changes requested):

client/src/modules/templates/bhEmployeeSelect.tmpl.html
jniles: Does this filter when you type EM.IMCK.XYZ or just on the employee name?  It looks like it is just on employee name.

client/src/modules/templates/bhEmployeeSelect.tmpl.html
```diff
@@ -20,6 +20,7 @@
       on-select="$ctrl.onSelect($item)">
       <ui-select-match placeholder="{{ 'FORM.SELECT.EMPLOYEE' | translate }}"><span>{{$select.selected.display_name}}</span></ui-select-match>
       <ui-select-choices ui-select-focus-patch repeat="employee.uuid as employee in $ctrl.employees | filter: { 'display_name': $select.search }">
```
jniles: Please make this filter on the employee's reference and it will be merged.

jniles (approved):
Good enough.

bors r+

jniles:
bors r+

It's a fluke.

PR #4372, opened 2020-04-27
improvement(Stock Movement)

### LINKED ISSUE #4347: Add period_id to stock_movement
Most of the stock calculations are performed on a monthly basis.  Let's add a column to the stock_movement table for the `period_id` so that we can filter out data much faster than using the date ranges.

Someone will need to write an upgrade script for our different institutions as well to make sure this works correctly and effectively.

lomamech:
- Add period_id in stock movement
- Writing the script to upgrade stock_movement in all institutions

closes #4347

### DIFF
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -54,7 +54,7 @@ function createStock(req, res, next) {
 
   Fiscal.lookupFiscalYearByDate(params.date)
     .then(result => {
-      const periodId = result.id || null;
+      const periodId = result.id;
 
       const transaction = db.transaction();
       const document = {
@@ -142,7 +142,7 @@ function createIntegration(req, res, next) {
 
   Fiscal.lookupFiscalYearByDate(params.movement.date)
     .then(result => {
-      const periodId = result.id || null;
+      const periodId = result.id;
 
       const integration = {
         uuid : db.bid(identifier),
@@ -218,7 +218,7 @@ function createMovement(req, res, next) {
 
   Fiscal.lookupFiscalYearByDate(params.date)
     .then(result => {
-      params.period_id = result.id || null;
+      params.period_id = result.id;
 
       const metadata = {
         project : req.session.project,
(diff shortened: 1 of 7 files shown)


lomamech:
@jniles all check have passed

server/models/migrations/next/migrate.sql
```diff
@@ -6,4 +6,17 @@ INSERT INTO unit VALUES
 (268, '[Stock] Consumption graph','TREE.STOCK_CONSUMPTION_GRAPH_REPORT','Stock Consumption graph report', 144,'/modules/reports/generated/stock_consumption_graph_report','/reports/stock_consumption_graph_report');
   
 INSERT INTO `report` (`report_key`, `title_key`) VALUES
-('stock_consumption_graph_report', 'REPORT.STOCK_CONSUMPTION_GRAPH_REPORT.TITLE');
\ No newline at end of file
+('stock_consumption_graph_report', 'REPORT.STOCK_CONSUMPTION_GRAPH_REPORT.TITLE');
+
+
+--@lomamech 2020-04-25
+ALTER TABLE `stock_movement` ADD COLUMN `period_id` MEDIUMINT(8) UNSIGNED DEFAULT NULL;
```
jniles: I don't think this should ever be NULL.  It doesn't make sense to record a stock movement for a day that we don't have a period defined.

server/controllers/stock/index.js
```diff
@@ -131,56 +140,62 @@ function createIntegration(req, res, next) {
...
-
-  const postingParams = [db.bid(documentUuid), 0, req.session.project.id, req.session.enterprise.currency_id];
-
-  if (req.session.enterprise.settings.enable_auto_stock_accounting) {
-    transaction.addQuery('CALL PostStockMovement(?)', [postingParams]);
-  }
-
-  // execute all operations as one transaction
-  transaction.execute()
+  Fiscal.lookupFiscalYearByDate(params.movement.date)
```
jniles, line 143: This is a good approach.

jniles (commented):
@lomamech the code looks pretty good to me.  Is there any reason why the periodId should be NULL by default?

lomamech:

jniles:
LGTM.

bors r+

PR #4373, opened 2020-04-27
Feat: render display values on server

### LINKED ISSUE #2233: Rendering display values on the server
With https://github.com/IMA-WorldHealth/bhima-2.X/issues/2130, we began a refactor to show human-readable text on the client for every filter value passed in.  That means, instead of `User : 2`, you will see `User: Jniles` rendered in the filter bar.  The underlying value used in the database WHERE statement is called the `value` and the value displayed to the user is the `displayValue`.

We need to do a similar thing for the server.  We've introduced filters at the top of most registries via the [filterbar](https://github.com/IMA-WorldHealth/bhima-2.X/pull/2230) helper.  We now need to implement something like [the displayValue logic](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/FilterService.js#L189) on the server.  It is at a disadvantage, since the client usually knows the human readable text to use (presumably, that is how the user selected the item).

I propose that we send back a "display value map" in the [FilterService.prototype.formatHTTP()](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/client/src/js/services/FilterService.js#L104) method.  The basic format will be `?displayValues={column}:{displayValue}[,{column}:{displayValue}]`.  For example:
```
?user_id=1&display_name="Niles"&displayValues=user_id:jniles
```
This has the following nice properties:
 1. No extra SQL necessary on the server.
 2. It is backwards-compatible.  If the `displayValues` query string is missing, it will simply behave as it does currently.
 3. No changes are required to `lib/filter.js`, other than parsing the display values.

The display values could be parsed like this:
```js
// on the server, in lib/filter.js

const parseDisplayValues = (displayValueString) => 
  displayValueString.split(',')
    .reduce((displayValueMap, displayValue) => {
      const [key, value] = displayValue.split(':');
      displayValueMap[key] = value;
      return displayValueMap;
    }, {});

// parse a display value string
parseDisplayValues(req.query.displayValues);
```

The above algorithm assumes the `displayValues` is a string ... it probably will be automatically converted to an array by ExpressJS if we provide it as an array on the client side.

jniles:
Implements the ability to use the client's displayValues on the server for rendering HTML and PDF reports.  It implements the design discussed in #2233 to allow reports to be updated without any changes to their rendering functions or the HTTP query itself.

I have gone through and brushed up the code of most registries with async/await as part of this.

**Before**
image

**After**
image

Closes #2233.

### DIFF
--- a/client/src/js/services/FilterService.js
+++ b/client/src/js/services/FilterService.js
@@ -158,6 +158,14 @@ function FilterService() {
     if (clientTimestamp) {
       httpFilters.client_timestamp = (new Date()).toJSON();
     }
+
+    // make displayValue for client.
+    const displayValues = Object.entries(this.getDisplayValueMap())
+      .map(([key, value]) => `${key}:${value}`)
+      .join(',');
+
+    Object.assign(httpFilters, { displayValues });
+
     return httpFilters;
   };
 
--- a/server/controllers/finance/reports/cash/index.js
+++ b/server/controllers/finance/reports/cash/index.js
@@ -194,9 +194,10 @@ async function report(req, res, next) {
       data.amounts = await db.exec(costSql, [uuids]);
     }
 
-    const project = await (query.project_id
-      ? Projects.findDetails(query.project_id)
-      : Promise.resolve({}));
+    let project = {};
+    if (query.project_id) {
+      project = await Projects.findDetails(query.project_id);
+    }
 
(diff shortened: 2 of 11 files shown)


lomamech (commented):
I just checked it, it works well

image

PR #4389, opened 2020-04-28
fix(purchases): use prev price in optimal PO

### LINKED ISSUE #4326: Improvements to "Optimal Purchase Orders"
We have an "optimal purchase order" functionality in the purchase orders module that creates a purchase order based on your previous stock consumption.

Currently, all the price information for the optimal purchase order is blank.    We need to implement https://github.com/IMA-WorldHealth/bhima/issues/4119 for this optimal purchase order feature.  The prices should be pulled out of the most recent price information.  This will greatly assist our users who need to fill out multiple purchase orders.

There is also no loading indicator on the "optimal purchase order" button which makes it difficult to know what state the module is in on a slow connection.

jniles:
Adds the previous purchase price into the optimal purchase order
scenario. Also reduces the number of HTTP requests by removing the
Inventory service call from the purchase controller and simply using the
inventory items that are already in the PurchaseFormService.

It also fixes the migration script so that we can migrate IMCK/Vanga successfully.

Closes #4326.

Here is it in action:
bjqALdiEPF

### DIFF
--- a/client/src/js/services/Pool.js
+++ b/client/src/js/services/Pool.js
@@ -1,7 +1,7 @@
 angular.module('bhima.services')
   .service('Pool', PoolService);
 
-PoolService.$inject = [ 'Store' ];
+PoolService.$inject = ['Store'];
 
 function PoolService(Store) {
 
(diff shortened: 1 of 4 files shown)


jniles:
@lomamech can I get a review?

lomamech (approved):
@jniles it works well, would it be possible to display the articles in alphabetical order

jniles:
@lomamech, sure, will do.

jniles:
@lomamech, this is done.

FDosJYlFrm

PR #4390, opened 2020-04-28
refactor(Stock report) Improvements to the Stock Consumption Graph report

### LINKED ISSUE #4353: Improvements to the Stock Consumption Graph Report
We just merged a graphical report on stock consumption (https://github.com/IMA-WorldHealth/bhima/pull/4336).  While using it to investigate live data, there are a couple of improvements we can make to make it more user-friendly.

1. Reorder the selection options so that the fiscal year and period are first (they are the only two required).
2. If multiple inventory items are selected, show the bar graphs in different colors.
3. Make the chart dynamically resize to take up 100% of the width of the page.  The PDF print out doesn't look good at all - see the attached PDF report.
4. The clear buttons don't work on the inventory and depot select.
6. Allow a user to choose between the _value_ of the stock or the _quantity_ of the stock.
7. Do not show non-consumable inventory items.
5. Sometimes the numbers stick above the graph - we should make the Y-axis long enough to accommodate the numbers on the graph.  See below:
image

[Example.pdf](https://github.com/IMA-WorldHealth/bhima/files/4527715/Test.pdf)

jeremielodi:
closes #4353

### DIFF
--- a/client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
+++ b/client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
@@ -22,41 +22,56 @@
 
         <div class="panel-body">
 
-        <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate autocomplete="off">
+          <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate autocomplete="off">
 
-          
-          <!-- select depot -->
-          <bh-depot-select
-            depot-uuid="ReportConfigCtrl.depot.uuid"
-            on-select-callback="ReportConfigCtrl.onSelectDepot(depot)">
-            <bh-clear on-clear="ReportConfigCtrl.clear('depot_uuid')"></bh-clear>
-          </bh-depot-select>
+            
+            <bh-fiscal-year-select
+              fiscal-id="ReportConfigCtrl.reportDetails.fiscal_id"
+              on-select-fiscal-callback="ReportConfigCtrl.onSelectFiscalYear(fiscalYear)">
+            </bh-fiscal-year-select>
+            
+            <bh-period-selection
+              fiscal-year-id="ReportConfigCtrl.reportDetails.fiscal_id"
+              period-id="ReportConfigCtrl.reportDetails.period_id"
+              on-select-callback="ReportConfigCtrl.onSelectPeriod(period)">
+            </bh-period-selection>
 
-          <!-- select inventory -->
-          <bh-inventory-select
-            inventory-uuid="ReportConfigCtrl.inventory.uuid"
-            on-select-callback="ReportConfigCtrl.onSelectInventory(inventory)"
-            only-consumable="true">
-            <bh-clear on-clear="ReportConfigCtrl.clear('inventory_uuid')"></bh-clear>
-          </bh-inventory-select>
+            <!-- select depot -->
+            <bh-depot-select
+              depot-uuid="ReportConfigCtrl.reportDetails.depot_uuid"
+              on-select-callback="ReportConfigCtrl.onSelectDepot(depot)">
+              <bh-clear on-clear="ReportConfigCtrl.clear('depot_uuid')"></bh-clear>
+            </bh-depot-select>
 
-          
-          <bh-fiscal-year-select
-            fiscal-id="ReportConfigCtrl.reportDetails.fiscal_id"
-            on-select-fiscal-callback="ReportConfigCtrl.onSelectFiscalYear(fiscalYear)">
-          </bh-fiscal-year-select>
-
-          <bh-period-selection
-            fiscal-year-id="ReportConfigCtrl.reportDetails.fiscal_id"
-            period-id="ReportConfigCtrl.reportDetails.period_id"
-            on-select-callback="ReportConfigCtrl.onSelectPeriod(period)">
-          </bh-period-selection>            
-
-          <!-- preview -->
-          <bh-loading-button loading-state="ConfigForm.$loading">
-            <span translate>REPORT.UTIL.PREVIEW</span>
-          </bh-loading-button>
-        </form>
+            <!-- select inventory -->
+            <bh-inventory-select
+              inventory-uuid="ReportConfigCtrl.reportDetails.inventory_uuid"
+              on-select-callback="ReportConfigCtrl.onSelectInventory(inventory)"
+              only-consumable="true">
+              <bh-clear on-clear="ReportConfigCtrl.clear('inventory_uuid')"></bh-clear>
+            </bh-inventory-select>
+            
+            <div class="form-group">
+              <label translate>FORM.LABELS.TYPE</label>
+                          
+              <div class="radio">
+                <label>
+                  <input type="radio" name="quantity" value="quantity" ng-model="ReportConfigCtrl.reportDetails.reportType"/>
+                  <span translate>STOCK.TYPE_QUANTITY</span>
+                </label>
+              </div>
+              <div class="radio">
+                <label>
+                  <input type="radio" name="value" value="value" ng-model="ReportConfigCtrl.reportDetails.reportType"/>
+                  <span translate>STOCK.TYPE_VALUE</span>
+                </label>
+              </div>
+            </div>
+            <!-- preview -->
+            <bh-loading-button loading-state="ConfigForm.$loading">
+              <span translate>REPORT.UTIL.PREVIEW</span>
+            </bh-loading-button>
+          </form>
         </div>
       </div>
     </div>
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -149,7 +149,7 @@ function getLots(sqlQuery, parameters, finalClauseParameter) {
         l.expiration_date, BUID(l.inventory_uuid) AS inventory_uuid, i.delay, l.entry_date,
         i.code, i.text, BUID(m.depot_uuid) AS depot_uuid, d.text AS depot_text, iu.text AS unit_type,
         BUID(ig.uuid) AS group_uuid, ig.name AS group_name,
-        dm.text AS documentReference, ser.name AS service_name, m.period_id
+        dm.text AS documentReference, ser.name AS service_name
       FROM lot l
       JOIN inventory i ON i.uuid = l.inventory_uuid
       JOIN inventory_unit iu ON iu.id = i.unit_id
@@ -272,7 +272,7 @@ function getLotsMovements(depotUuid, params) {
       m.flux_id, BUID(m.entity_uuid) AS entity_uuid, m.unit_cost,
       f.label AS flux_label, i.delay,
       iu.text AS unit_type,
-      dm.text AS documentReference, m.period_id
+      dm.text AS documentReference
     FROM stock_movement m
     JOIN lot l ON l.uuid = m.lot_uuid
     JOIN inventory i ON i.uuid = l.inventory_uuid
@@ -304,7 +304,7 @@ function getLotsOrigins(depotUuid, params) {
     SELECT BUID(l.uuid) AS uuid, l.label, l.unit_cost, l.expiration_date,
         BUID(l.inventory_uuid) AS inventory_uuid, BUID(l.origin_uuid) AS origin_uuid,
         l.entry_date, i.code, i.text, origin.display_name, origin.reference,
-        BUID(m.document_uuid) AS document_uuid, m.flux_id, m.period_id,
+        BUID(m.document_uuid) AS document_uuid, m.flux_id,
         iu.text AS unit_type,
         dm.text AS documentReference
     FROM lot l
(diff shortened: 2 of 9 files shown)


jeremielodi:
muti-enventory

client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
```diff
@@ -22,41 +22,59 @@ <h3 translate>REPORT.STOCK_CONSUMPTION_GRAPH_REPORT.TITLE</h3>
...
-          <bh-depot-select
-            depot-uuid="ReportConfigCtrl.depot.uuid"
-            on-select-callback="ReportConfigCtrl.onSelectDepot(depot)">
-            <bh-clear on-clear="ReportConfigCtrl.clear('depot_uuid')"></bh-clear>
-          </bh-depot-select>
+            
+            <bh-fiscal-year-select
+              fiscal-id="ReportConfigCtrl.reportDetails.fiscal_id"
+              on-select-fiscal-callback="ReportConfigCtrl.onSelectFiscalYear(fiscalYear)">
+            </bh-fiscal-year-select>
```
jniles, line 31: Excellent.  This is much better ordering.  :+1:

client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
```diff
@@ -22,41 +22,59 @@ <h3 translate>REPORT.STOCK_CONSUMPTION_GRAPH_REPORT.TITLE</h3>
...
-          <bh-fiscal-year-select
-            fiscal-id="ReportConfigCtrl.reportDetails.fiscal_id"
-            on-select-fiscal-callback="ReportConfigCtrl.onSelectFiscalYear(fiscalYear)">
-          </bh-fiscal-year-select>
+            <!-- select inventory -->
+            <bh-inventory-select
+              inventory-uuid="ReportConfigCtrl.reportDetails.inventory_uuid"
+              on-select-callback="ReportConfigCtrl.onSelectInventory(inventory)"
+              only-consumable="true">
+              <bh-clear on-clear="ReportConfigCtrl.clear('inventory_uuid')"></bh-clear>
```
jniles, line 51: Looks like these work now!  👍

client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
```diff
@@ -22,41 +22,59 @@ <h3 translate>REPORT.STOCK_CONSUMPTION_GRAPH_REPORT.TITLE</h3>
...
+            <bh-fiscal-year-select
+              fiscal-id="ReportConfigCtrl.reportDetails.fiscal_id"
+              on-select-fiscal-callback="ReportConfigCtrl.onSelectFiscalYear(fiscalYear)">
+            </bh-fiscal-year-select>
+            
+            <bh-period-selection
+              fiscal-year-id="ReportConfigCtrl.reportDetails.fiscal_id"
+              period-id="ReportConfigCtrl.reportDetails.period_id"
+              on-select-callback="ReportConfigCtrl.onSelectPeriod(period)">
+            </bh-period-selection>
```
mbayopanda, line 37: Is it possible to set the label to something like "Choisir le mois" because it lets the user think that it will get data from start until the selected period

server/controllers/stock/core.js
```diff
@@ -428,9 +428,13 @@ async function getDailyStockConsumption(params) {
...
-    SELECT SUM(m.quantity) as quantity, DATE(m.date) as date, 
-        i.uuid AS inventoty_uuid, i.text AS inventory_name,
-        d.text AS depot_name, d.uuid AS depot_uuid, m.period_id
+    SELECT SUM(m.quantity) as quantity,
+      SUM(m.quantity*m.unit_cost) as value,
+      DATE(m.date) as date,
+      HEX(i.uuid) AS inventory_uuid,
+      i.text AS inventory_name,
+      d.text AS depot_name,
+      HEX(d.uuid) AS depot_uuid
```
jniles: Could you use `BUID()` here instead of `HEX()`?  They are the same thing at the moment, but it was a pretty painful migration back in 1.x when we had to change the format and everyone had written their own custom code.

server/controllers/stock/core.js
```diff
@@ -272,7 +272,7 @@ function getLotsMovements(depotUuid, params) {
       m.flux_id, BUID(m.entity_uuid) AS entity_uuid, m.unit_cost,
       f.label AS flux_label, i.delay,
       iu.text AS unit_type,
-      dm.text AS documentReference, m.period_id
```
jniles, line 275: I think @lomamech added these `period_id`s in.  Did you mean to remove them?  Is there something they screw up?

server/controllers/stock/reports/stock/consumption_graph.js
```diff
@@ -25,48 +25,41 @@ async function stockConsumptionGrathReport(req, res, next) {
...
     const reportResult = await report.render({
-      labels : JSON.stringify(chart.labels),
-      yAxesLabelString : JSON.stringify(i18n(options.lang)('FORM.LABELS.QUANTITY')),
-      xAxesLabelString : JSON.stringify(i18n(options.lang)('FORM.LABELS.DAYS')),
-      title :  JSON.stringify(inventory.text || ''),
-      data : JSON.stringify(chart.data),
       dateFrom : options.dateFrom,
       dateTo : options.dateTo,
       depot,
+      chartjs : chartjs.barChar({
```
jniles: Can we call this `barChart()` instead of `barChar()`?

server/lib/util.js
```diff
@@ -243,6 +244,14 @@ function createDirectory(dirPath) {
   }
 }
 
+
+function getRandomColor() {
+  const r = Math.floor(Math.random() * 255);
+  const g = Math.floor(Math.random() * 255);
+  const b = Math.floor(Math.random() * 255);
+  return `rgb(${r},${g},${b})`;
```
jniles, line 252: You may consider looking into CSS's [`hcl()`](https://www.w3schools.com/csSref/func_hsl.asp) for better color consistency.

jniles: [Another resource](http://hclwizard.org/color-scheme/) explaining the science behind it.

server/lib/chart.js
```diff
@@ -0,0 +1,127 @@
...
+  };
+  const serieMap = {};
+
+  const dailyData = _.groupBy(data, label);
+  chart.labels = Object.keys(dailyData).sort((a, b) => {
+    return a.localeCompare(b);
+  });
+
+  const items = Object.keys(_.groupBy(data, item.uuid));
+  chart.labels.forEach(day => {
```
jniles, line 89: What is this loop doing?  I've been reading this function over and over and I can't figure it out.

server/lib/chart.js
```diff
@@ -0,0 +1,127 @@
...
+    xAxesLabelString, canvasId,
+  } = params;
+
+  const chart = {
+    labels : [],
+    data : [],
+  };
+  const serieMap = {};
+
+  const dailyData = _.groupBy(data, label);
```
jniles: It seems to me that you probably want to either:
  1. Sum based if you are dealing with quantities (using [_.sumBy()](https://lodash.com/docs/4.17.15#sumBy))
  2. Count the number of records if you are looking at just records (using [_.countBy()](https://lodash.com/docs/4.17.15#countBy)).

But maybe I'm missing something.

server/lib/chart.js
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+function barChar(params) {
+  const {
+    label, data, item, yAxesLabelString,
+    xAxesLabelString, canvasId,
+  } = params;
+
+  const chart = {
+    labels : [],
```
jniles, line 52: Since we know this is always going to be for a period, can we just generate the values of 0 - 30 days for the labels?  Any days without data will be 0?

jniles (commented):
Since @lomamech added in period_ids, can you migrate [this section](https://github.com/IMA-WorldHealth/bhima/pull/4390/files#diff-adf2f9d2ae8e9da41ea2ad217e8c0bd1R420-R424) to using the period_id instead?  It should give a nice performance increase.

I am finding it hard to understand what the code is doing in your 3 nested loops of the `barChar` function.  Could you add some comments or variable names to explain what is going on?   Or maybe try to unroll some of the loops a bit?

server/controllers/stock/core.js
mbayopanda: Why did you use the `HEX` function ?

server/controllers/stock/core.js
```diff
@@ -441,8 +445,10 @@ async function getDailyStockConsumption(params) {
   filters.dateTo('dateTo', 'date');
   filters.equals('depot_uuid', 'uuid', 'd');
   filters.equals('inventory_uuid', 'uuid', 'i');
-  filters.custom('consumption', '(m.flux_id IN (9, 10) OR (m.flux_id = 8 AND d.is_warehouse = 1))');
+  filters.custom('consumption', `((m.flux_id IN (9,10) OR (m.flux_id=8 AND d.is_warehouse=1))) AND i.consumable=1`);
```
mbayopanda: Is it possible to have `constants` instead of integer value so that we can quickly figure out what the code have to do. for example : 
```js
const value = `
(
  (m.flux_id IN (${flux.TO_PATIENT}, ${flux.TO_SERVICE}) 
  OR 
  (m.flux_id=${flux.TO_OTHER_DEPOT} AND d.is_warehouse=1)
) AND i.consumable=1
`
filters.custom('consumption', value);
```

client/src/modules/reports/generate/stock_consumption_graph_report/stock_consumption_graph_report.html
```diff
@@ -22,41 +22,59 @@ <h3 translate>REPORT.STOCK_CONSUMPTION_GRAPH_REPORT.TITLE</h3>
...
+                  <span translate>STOCK.TYPE_QUANTITY</span>
+                </label>
+              </div>
+              <div class="col-sm-12">
+                <label class="radio-inline">
+                  <input type="radio" name="value" value="value" ng-model="ReportConfigCtrl.reportDetails.reportType"/>
+                  <span translate>STOCK.TYPE_VALUE</span>
+                </label>
+              </div>
+            </div>
```
mbayopanda, line 69: Is it possible to remove `row` and `col` classes, and just use `form-group` for example : 
```html
<div class="form-group">
  <label translate>FORM.LABELS.TYPE</label>
              
  <div class="radio">
    <label>
      <input type="radio" name="quantity" value="quantity" ng-model="ReportConfigCtrl.reportDetails.reportType"/>
      <span translate>STOCK.TYPE_QUANTITY</span>
    </label>
  </div>
  <div class="radio">
    <label>
      <input type="radio" name="value" value="value" ng-model="ReportConfigCtrl.reportDetails.reportType"/>
      <span translate>STOCK.TYPE_VALUE</span>
    </label>
  </div>
</div>
```

server/controllers/stock/reports/stock/consumption_graph.js
```diff
@@ -25,48 +25,41 @@ async function stockConsumptionGrathReport(req, res, next) {
...
+        data : result,
+        item : {
+          uuid : 'inventory_uuid',
+          name : 'inventory_name',
+          value : options.reportType || 'quantity',
+        },
+        yAxesLabelString : i18n(options.lang)(`FORM.LABELS.${reportType.toUpperCase()}`),
+        xAxesLabelString : i18n(options.lang)('FORM.LABELS.DAYS'),
+        canvasId : 'stockConsumptionChart',
+      }),
```
mbayopanda, line 63: 👍

server/lib/chart.js
mbayopanda: Could you add in the description of the function, the format of different variables (data, label, item, etc.) so that we could easily reuse it

server/controllers/stock/core.js
jeremielodi: okay

mbayopanda (commented):
@jeremielodi i made just some comments, the code is 👍

server/lib/chart.js
jeremielodi: This function will also be used for other purposes. I created a function to make it reusable.
While creating series for chart.js, all items must have a value for every label.
Imagine on march 2th, a user want to know the stock consumption. The series will have a lot value for nothing.

Let take an example

```js
labels = ['01','02'];
```

```js
items = ['inventory1', 'inventory2', 'inventory3']
```

the series (datasets) should be
```js
[{
   label : 'inventory1',
   data : [10, 3]
 },
{ 
  label : 'inventory2',
  data : [5, 45]
},
{
  label : 'inventory3',
  data : [20, 0]
}];
```

each dataset should have the an array(data) that contains the value for each label
so the array's length must be equals to label's length

server/controllers/stock/reports/stock/consumption_graph.js
```diff
@@ -25,48 +25,45 @@ async function stockConsumptionGrathReport(req, res, next) {
...
-    }
     if (options.depot_uuid) {
       depot = await db.one(depotSql, db.bid(options.depot_uuid));
     }
-
+    if (params.period_id) {
+      const period = await db.one('SELECT start_date,end_date FROM period WHERE id=?', params.period_id);
+      options.dateFrom = period.start_date;
+      options.dateTo = period.end_date;
+    }
```
jniles, line 42: I don't think the you need to add the dates to the options object anymore.

server/controllers/stock/reports/stock_consumption_graph.handlebars
```diff
@@ -5,7 +5,7 @@
 
 <body>
 
-  <div class="container">
+  <div class="content-fluid">
```
jniles: I don't think this class actually exists.  You can probably just remove it and the excess `<div>`.

server/controllers/stock/reports/stock/consumption_graph.js
jeremielodi: it's used 
```js

    const reportResult = await report.render({
      dateFrom : options.dateFrom,
      dateTo : options.dateTo,
      depot,
   ...

jniles: I see. Can you separate the SQL query options from the rendering options?  You are adding these as SQL filters which will defeat the period_id performance increase.

jniles (approved):
Good improvements.  I especially like the comments.  We should probably break out the data manipulation functions into their own library and add some unit tests in the future.

bors r+

PR #4400, opened 2020-04-30
chore(purchase): refactor receipt + queries

### LINKED ISSUE #4303: Make Purchase Orders receipt look like other receipts
Our purchase orders have a non-standard presentation.  The reference/identifier is on the center of the page, rather than on the top right corner, like all other receipts.  Here is an example of a voucher.

image

And here is the changes that need to be made to the Purchase Orders:
image

Basically:
 1. Move the text "Ordre d'Achat" to the top right corner.
 1. Move the reference "PO.XXX.123" directly beneath the "Ordre d'Achat" text in the top right corner.

This will also have the effect of giving us more room on the receipt to print text.

jniles:
Refactors the purchase receipt to look more like the other receipts in
the system. Also includes the inventory code for the medications used at
the hospital.  Finally, the queries have been migrated to depend more
thoroughly on document_map.

Closes #4303.

Here is what it looks like for IMCK:
image

Here is what it looks like for HEV:
image

Also, since I changed the `reference` query, here is proof that it works:
gPBOfSXQr4

### DIFF
--- a/server/controllers/finance/purchases.js
+++ b/server/controllers/finance/purchases.js
@@ -100,12 +100,12 @@ function lookupPurchaseOrder(uid) {
   let record;
 
   let sql = `
-    SELECT BUID(p.uuid) AS uuid,
-      CONCAT_WS('.', '${entityIdentifier}', pr.abbr, p.reference) AS reference,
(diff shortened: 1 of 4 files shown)


jeremielodi (approved):
LGTM

PR #4402, opened 2020-04-30
Refactor Employee and Multiple Payroll w/ Links and References

### LINKED ISSUE #4350: Link the employee from the Multiple Payroll page
It would be nice to be able to click a dropdown on the mulitple payroll page and go immediately to the Employee's page.  Right here:
image

This would let us quickly check information on the employee before committing to pay them.

### LINKED ISSUE #4351: The employee "reference" should appear on the multiple payroll page
Here is an example list of employees:
image

Each employee has a reference and this is used in the posting journal and the general ledger to link them.  For example:

image

We should also include this column on the multiple payroll page.  At the moment, it doesn't have this column.

image

jniles:
This PR refactors a lot of the employee registry and search modal, and adds in a link between the multiple payroll and the employee registry.

Below is a screenshot of multiple payroll, showing the following improvements:
  1. There is a link to the employee
  2. There is a reference column for the employee
  3. There is a header with the employee reference on the dropdown.
  4. The $$ columns are now right-aligned.
  5. The actions dropdown is right-aligned.
image

There is a demonstration of the link working:
SWf7VZvrO3

Finally, I've cleaned up the employee search modal by using proper HTML and removing extraneous imports/code.

Closes #4350.
Closes #4351.

### DIFF
--- a/client/src/i18n/en/employee.json
+++ b/client/src/i18n/en/employee.json
@@ -9,6 +9,7 @@
   "REGISTER" : "Register Employee",
   "REGISTRATION_FORM" : "Registration Form",
   "TITLE":"Employee Management",
-  "NO_GRADE":"No grade assigned"
+  "NO_GRADE":"No grade assigned",
+  "VIEW_EMPLOYEE" : "View Employee"
  }
 }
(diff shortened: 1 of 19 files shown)


lomamech (commented):
I just checked it, it works really well

jniles:
@lomamech, thanks.  I'm still struggling with the tests.  Once they pass, I'll bors it.

jniles:
Ah, I see you already did.  I don't think they will pass, but we'll see!

jniles:
Okay, this should be the final change here.  Finally got over my laziness to actually track employee IDs and make them the same as the previous employee IDs.

jniles:
@lomamech, thanks for your help debugging yesterday.  This is finally ready for a final review.

PR #4403, opened 2020-04-30
docs(journal): add note about showing full txns

### LINKED ISSUE #3224: Show Full Transactions mode overrides LIMIT in Posting Journal
In the Posting Journal, it is possible to load data out of bounds outside of the LIMIT the users specifies by toggling "show full transactions" to be true.  This is because "show full transactions" attempts to only show all the rows of a transaction, instead of cutting it in half.

At minimum, we should document this behavior.  The question is: should we change it?  See below:

jniles:
Adds documentation to about overriding the limit field by selecting full transactions.

Closes #3224.

-----
[View rendered docs/en/finance/ledgers/posting-journal.md](https://github.com/jniles/bhima-2.X/blob/docs-add-note-about-show-full-transactions-3224/docs/en/finance/ledgers/posting-journal.md)

### DIFF
--- a/docs/en/finance/ledgers/posting-journal.md
+++ b/docs/en/finance/ledgers/posting-journal.md
@@ -101,3 +101,10 @@ Every transaction in the system has a transaction type.  Transaction types facil
 * **Custom**
 
 A transaction can only have a single transaction type.  You can add your own transaction types through the Transaction Type module.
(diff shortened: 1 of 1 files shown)


mbayopanda (approved):
bors r+

PR #4415, opened 2020-05-04
feat(stock exit) show sold out inventory

### LINKED ISSUE #4381: Show "X items out of stock" on the Stock Exit page for a depot
We have a large empty space in the stock exit module.  Here it is:
image
_Fig 1: Stock Exit Module empty area_

I propose that we make better use of this area by adding in a list of items that are out of stock ordered by the most recent items that have gone out of stock.  It would look something like this:
image

Basically, list the 5 most recent stockouts, and then provide a link to the "articles in stock" for that depot filtered on items out of stock.  This will hopefully prompt our users to do something about their stockouts rather than simply sit with incorrect stock counts in their depot.

jeremielodi:
closes #4381

### DIFF
--- a/client/src/modules/stock/exit/exit.html
+++ b/client/src/modules/stock/exit/exit.html
@@ -86,6 +86,49 @@
             </div>
           </div>
         </div>
+
+        <div class="col-xs-12 col-md-6 panel-default">
+
+          <table
+            ng-show = "StockCtrl.soldOutInventories.length > 0"
+            class="table table-condensed table-bordered">
+            <thead>
+              <tr>
+                <th>
+                  <span class="text-danger" 
+                    translate-values="{ total : StockCtrl.soldOutInventories.length }"
+                    translate="{{ 'STOCK.SOLD_OUT_WARNING' }}">
+                  </span>
+                </th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr ng-repeat = "inventory in StockCtrl.soldOutInventories | limitTo : 5">
+                <td>{{inventory.text}} ({{inventory.stock_out_date}})</td>
+              </tr>
+
+              <tr>
+                <td>
+
+                  <span class="text-danger" 
+                    ng-show="(StockCtrl.soldOutInventories.length - 5) > 0"
+                    translate-values="{ left : (StockCtrl.soldOutInventories.length - 5)}"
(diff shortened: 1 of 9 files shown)


client/src/modules/stock/exit/exit.html
```diff
@@ -86,6 +86,49 @@
...
+        <div class="col-xs-12 col-md-6 panel-default">
+
+          <table
+            ng-show = "StockCtrl.soldOutInventories.length > 0"
+            class="table table-bordered">
+            <thead>
+              <tr>
+                <th>
+                  <span class="text-danger" 
+                    translate-values="{ total : StockCtrl.soldOutInventories.length }"
```
jniles, line 99: Nice :+1:

client/src/modules/stock/exit/exit.html
```diff
@@ -86,6 +86,49 @@
             </div>
           </div>
         </div>
+
+        <div class="col-xs-12 col-md-6 panel-default">
+
+          <table
+            ng-show = "StockCtrl.soldOutInventories.length > 0"
+            class="table table-bordered">
```
jniles: To save space, could you add "table-condensed" to these classes?

client/src/modules/stock/exit/exit.html
```diff
@@ -86,6 +86,49 @@
...
+                  <span class="text-danger" 
+                    ng-show="(StockCtrl.soldOutInventories.length - 5) > 0"
+                    translate-values="{ left : (StockCtrl.soldOutInventories.length - 5)}"
+                    translate="{{ 'STOCK.AND_MORE' }}">
+                  </span>
+
+                <a ui-sref= "stockInventories({ filters : [
+                  { key : 'period', value : 'allTime'},
+                  { key : 'includeEmptyLot', value : 1  },
+                  { key : 'depot_uuid', value : StockCtrl.depot.uuid, displayValue: StockCtrl.depot.text },
```
jniles: Could you add `cacheable:false` to both of these keys: `depot_uuid` and `status`?

jniles (approved):
This looks great to me.  Thanks @jeremielodi !

I've made two quick comments, they shouldn't be hard changes.

jniles:
LGTM.  Great job @jeremielodi !

bors r+

PR #4441, opened 2020-05-05
test: refactor fee center GridAction tests

jniles:
This commit refactors a lot of the tests in the Fee Center/Distribution Center modules so that they are more precise using GridRow instead of GridAction.  It should make our test more stable and predictable.


jniles:
@lomamech can I get a review?

lomamech (approved):
This PR not having disturbed the functioning, nor even the tests

LGTM

PR #4445, opened 2020-05-05
Improve db creation scripts

individual-it:
1. create DB on the given server
2. if something fails exit the script and show an error message

I ran into trouble because I've started the DB in docker and could not run the e2e tests, the DB creation failed but there hasn't been any error message, I had to debug first why its failing.

This should reduce confusion specially for new developers when first trying to run the scripts and something does not work


jniles (approved):
This is an awesome addition.  Thank you very much!

bors r+

PR #4446, opened 2020-05-05
chore(stock): replace await with MAX() query

jniles:
This commit reduces the amount of SQL trashing that happens in the stock exit module by reducing the number of queries sent.  The same result is produced at the end.  Also fixes a few typos.


jeremielodi (approved):
LGTM 👍

PR #4453, opened 2020-05-06
Test installation in CI

### LINKED ISSUE #3998: Run installation tests on a CI
We currently have a few tests for the installation page that aren't run as part of our regular CI builds.  We risk having the installation procedure crash without knowing it.  Let's figure out how to run our installation tests as part of the regular CI checks.  This could be on appveyor or an entirely new CI.

For reference, here is the test file: https://github.com/IMA-WorldHealth/bhima/blob/master/sh/install-tests.sh

individual-it:
fixes #3998


jniles (approved):
Woah, this is awesome.  Thanks so much @individual-it !

bors r+

PR #4454, opened 2020-05-06
Add loading indicator

### LINKED ISSUE #4434: Stock Entry from Purchase needs a loading indicator on grid
When you attempt to make a stock entry from a purchase order, you must select a purchase order from a grid of purchase orders.  There is no loading indicator on this grid though, so by default, on a slow connection, it shows "no records available".  This is a bug - it should show a loading indicator to tell the user that it doesn't have any results yet.  See below this in action.

vstDJEocgw

mbayopanda:
Close #4434


jniles (approved):
LGTM.  Tested and working.

bors r+

PR #4455, opened 2020-05-06
Fix self requisition

### LINKED ISSUE #4422: A depot should not be able to do a requisition from itself
So, this is possible.  And it shouldn't be:
image

We should filter out the depot that is currently selected from the depots that we can requisition from.

mbayopanda:
Close #4422

### DIFF
--- a/server/controllers/inventory/depots/index.js
+++ b/server/controllers/inventory/depots/index.js
@@ -116,6 +116,14 @@ function list(req, res, next) {
     options.user_id = req.session.user.id;
   }
 
+  if (options.exception && Array.isArray(options.exception)) {
+    options.exception = options.exception.map(depotUuid => {
+      return db.bid(depotUuid);
+    }).join(',');
+  } else if (options.exception && typeof options.exception === 'string') {
+    options.exception = db.bid(options.exception);
+  }
+
   options.enterprise_id = req.session.enterprise.id;
 
   const filters = new FilterParser(options, { tableAlias : 'd' });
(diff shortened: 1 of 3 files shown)


server/controllers/inventory/depots/index.js
```diff
@@ -116,6 +116,14 @@ function list(req, res, next) {
     options.user_id = req.session.user.id;
   }
 
+  if (options.exception && Array.isArray(options.exception)) {
```
jniles: Do you really need this?  What happens if you just pass `options.exception` to `db.convert()`?  Like so:

```js
db.convert(options, ['exception']);
```
From my reading of [the code](https://github.com/IMA-WorldHealth/bhima/blob/master/server/lib/db/index.js#L209) it should do all this for you automagically.

jniles (commented):
@mbayopanda, I've made one suggestion to try and clean up the code.  Give it a whirl and see if that helps.

server/controllers/inventory/depots/index.js
mbayopanda: Yes, for pepkits for example they asked me if it is possible to hide some depots which are not in a given health zone. Having an array of uuids will help us to do that

jniles: So, why does db.convert() not handle arrays?  Is there a bug in it?  [This code](https://github.com/IMA-WorldHealth/bhima/blob/master/server/lib/db/index.js#L221)
```js
      if (prop && _.isArray(prop)) {
        // Every item should be converted to binary
        data[key] = data[key].map(this.bid);
      }
```
should be doing exactly what you are doing.  Is there a reason by db.convert() is failing here?

mbayopanda: It works

jniles (approved):
This is a much cleaner fix.  Very nice :+1:

PR #4456, opened 2020-05-06
feat(Rapport caisse auxiliaire) display background colors

### LINKED ISSUE #4382: Colors are not rendered in the PDF report with puppeteer
I'm not sure why colors are not being rendered on puppeteer.  Here is the "analyse de caisse auxiliares" report:

image
_Fig 1: PDF version of the caisse auxiliares report_

Normally, these icons have colors behind them.  It should look like this:
image

Someone should investigate this and correct this issue.

jeremielodi:
closes #4382

### DIFF
--- a/server/lib/template/partials/head.handlebars
+++ b/server/lib/template/partials/head.handlebars
@@ -5,6 +5,11 @@
   <link href="{{absolutePath}}/css/bhima-bootstrap.css" rel="stylesheet">
   <script src="{{nodeModulesPath}}/jsbarcode/dist/JsBarcode.all.min.js"></script>
   <style>
+    
+    * {
+      -webkit-print-color-adjust: exact !important;
+    }
+
     thead {
       display : table-header-group;
     }
--- a/server/controllers/finance/reports/analysis_auxiliary_cashbox/report.handlebars
+++ b/server/controllers/finance/reports/analysis_auxiliary_cashbox/report.handlebars
@@ -15,19 +15,19 @@
 
           <table style="width: 25%; float: right; margin-bottom: 1.5%">
             <tr>
-              <td width="10%" style="background-color: #5cb85c; color:white; text-align:center"> ✔ </td>
+              <td width="10%" style="background-color: #5cb85c !important; color:white; text-align:center"> ✔ </td>
               <td width="90%">&nbsp;<strong>{{translate 'REPORT.ANALYSIS_AUX_CASHBOXES.STATUS.CORRECT'}}</strong></td>
             </tr>
             <tr>
-              <td width="10%" style="background-color: #777777; color:white; text-align:center"> X </td>
+              <td width="10%" style="background-color: #777777  !important; color:white; text-align:center"> X </td>
               <td width="90%">&nbsp;<strong>{{translate 'REPORT.ANALYSIS_AUX_CASHBOXES.STATUS.TRANSFERT_PENDING'}}</strong></td>
             </tr> 
             <tr>
-              <td width="10%" style="background-color: #f0ad4e; color:white; text-align:center"> ▲ </td>
+              <td width="10%" style="background-color: #f0ad4e  !important; color:white; text-align:center"> ▲ </td>
               <td width="90%">&nbsp;<strong>{{translate 'REPORT.ANALYSIS_AUX_CASHBOXES.STATUS.TRANSFERT_GREATER'}}</strong></td>
             </tr>
             <tr>
-              <td width="10%" style="background-color: #d9534f; color:white; text-align:center"> ▼ </td>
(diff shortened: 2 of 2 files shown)


jeremielodi:
Everything shows up correctily
pdf_bg

jniles:
@jeremielodi does this make all colors show up in all PDF reports?  https://github.com/IMA-WorldHealth/bhima/issues/4382 was only used the  Caisse Auxiliare report as an example.  If so, what is the change that makes them all work?

It would be a much better developer experience if we could just figure out what was causing this not to work correctly and override that rule or remove it.

jeremielodi:
I made some comments at #4382

jeremielodi:
Every thing should work, if there's a problem some where we must make every background <b>!important</b>

server/lib/template/partials/head.handlebars
```diff
@@ -5,6 +5,11 @@
   <link href="{{absolutePath}}/css/bhima-bootstrap.css" rel="stylesheet">
   <script src="{{nodeModulesPath}}/jsbarcode/dist/JsBarcode.all.min.js"></script>
   <style>
+    
+    * {
```
jniles, line 9: After this change, can you removed the `!important` from the handlebars file and it will work?

jniles: Or will we be required to go into each style tag and put `!important`..?

jeremielodi: As we dont't know where we set each background <b>transparent !important</b>, adding <b> !important will help

jniles: How long did you spend looking in the CSS to try and find it?  I would much prefer the solution where we just override that rule, remove it, or change it than asking each developer to always put !important.  That seems like a really silly fix.

jeremielodi: I'm trying to find that. If it's due to <b>background transparent !important </b> some where in our css I will remote it else we will momentarily add <b>!important</b>

jniles: Perfect, thank you.  Let me know if you need help finding it.

jeremielodi: I would like to know where this bellow code is written
```css
/*! Source: https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css */
@media print {
  *,
  *:before,
  *:after {
    color: #000 !important;
    text-shadow: none !important;
    background: transparent !important;
    box-shadow: none !important;
  }
```
I can only see it at http://localhost:8080/css/bhima-bootstrap.css

jniles: It is in bootstrap.less, which imports it from print.less in the node_modules folder.

The exact path is `node_modules/bootstrap/less/print.less`.

jniles: To override it, you can probably just do:
```less
@media print {
  *,
  *:before,
  *:after {
    color: inherit !important; // Black prints faster: h5bp.com/s
    background: initial !important;
  }
```
Or something similar.  Put that in the bhima-bootstrap.less and it should all work.

jeremielodi: Curiously nothing works after all those manipulations.

jniles: Alright, I'll give it a whirl and see what I can come up with.

jniles: So, the property I was hoping to use ([revert](https://developer.mozilla.org/en-US/docs/Web/CSS/revert)) is not supported in Chrome.  That is the main issue.

Further, since it is applied to all elements (`*`) we cannot make a more specific selector.

So, I think the solution will be to remove the @import bootstrap and create our own @import rules that essentially pull in everything but that `print.less`.  Then we can copy what we want from that print.less file into our own, modified version.

jniles: @jeremielodi Do you want to tackle this, or do you want me to try?

jeremielodi: @jniles you can try, but what I noticed is even though you remove all bootstrap less files, the problem remains

```js

/**
 * @function compileLess
 *
 * @description
 * Copiles the less paths into CSS file.  Must be called
 * before the CSS compilation step.
 */
function compileLess() {
  return src(LESS_PATH)
    // .pipe(gulpLess(LESS_CONFIG))
    .pipe(gulpif(isProduction, postcss([cssnano({ zindex : false })])))
    .pipe(dest(`${CLIENT_FOLDER}/css`));
}

jeremielodi:
@jniles I think this PR should be closed

jniles:
Go ahead and close it!

Issue #4335, opened 2020-04-21
Investigate flakey stock movement tests


jniles:
It looks like our stock movement tests seem to flake out about 50% of the time.  I'm not sure why.  Here is an example:

https://travis-ci.com/github/IMA-WorldHealth/bhima/builds/161203135#L1337

My assumption is that this has to do with the movements registry being grouped by depot and the test isn't waiting long enough for the grouping to occur and count the extra row.  Either that, or we have some randomness in how many stock entry movements get generated.

jniles:
@kwilu claim.

jniles:
I'm almost certain this is an ORDER BY issue.

individual-it:
seems to fail more often/always with MySQL 8 #4473 and MariaDB #4442

jniles:
Indeed - that's why I forked #4442 to make https://github.com/IMA-WorldHealth/bhima/pull/4469.

So far, it seems like ORDER BY has been the main cause.  I've gotten it to reliably move past that failure on MariaDB.

Now I'm working on just running over and over the tests to see what else we can shake out while we are on a roll.

jniles:
With any luck, we'll find that it passes on MySQL 8 with the same changes.

PR #4465, opened 2020-05-07
fix(stock): add ORDER BY to the stock movements query

jniles:
Adds an ORDER BY to the stock movements query to give more predictable results.  Ordering is by the movement date.

Addresses #4335.


jniles:
bors try

jniles:
bors try

jniles:
bors try

jniles:
bors try

jniles:
bors try

jniles:
bors try

jniles:
Well, we tried.  I'll keep hunting.

PR #4474, opened 2020-05-08
Fix: test on mariadb reliably

### LINKED ISSUE #3501: Officially support MariaDB
We could save ourselves a lot of gnashing of teeth during our install and new user onboarding if we could figure out a way to support MariaDB and MySQL simultaneously.  [There is a list of differences between the two RDBs on MariaDBs websites](https://mariadb.com/kb/en/library/incompatibilities-and-feature-differences-between-mariadb-102-and-mysql-57/).

I think the basic approach to getting MariaDB support would be through developing a test suite that runs in multiple environments.  The easiest way to do this would be to make a docker version of BHIMA (#3439), swap out the database, and run the full test suite against it.

### LINKED ISSUE #4335: Investigate flakey stock movement tests

jniles:
This PR inherits from the https://github.com/IMA-WorldHealth/bhima/pull/4442 to run tests on mariadb.   A huge thank you to @individual-it for helping with this debugging.

This PR cleans up tests to remove flakey tests or stochastic queries.  Most of the randomness came from ORDER BY queries that were under-specified.  For example, the stock queries often failed because the order was somewhat random - ordering by date wasn't enough because our test dataset had multiple rows created at the same second.  Similarly, vouchers were not ordered by anything.

Closes https://github.com/IMA-WorldHealth/bhima/issues/3501
Closes https://github.com/IMA-WorldHealth/bhima/issues/4335
Closes https://github.com/IMA-WorldHealth/bhima/pull/4442
Closes https://github.com/IMA-WorldHealth/bhima/pull/4465

Furthermore, this should unblock a lot of the https://github.com/IMA-WorldHealth/bhima/pull/4473 to get more experience with MySQL 8.


jniles:
bors try

server/controllers/stock/core.js
```diff
@@ -143,8 +143,7 @@ function getLotFilters(parameters) {
  * @param {object} parameters - A request query object
  * @param {string} finalClauseParameter - An optional final clause (GROUP BY, HAVING, ...) to add to query built
  */
-function getLots(sqlQuery, parameters, finalClauseParameter) {
-  const finalClause = finalClauseParameter;
+function getLots(sqlQuery, parameters, finalClause = '', orderBy) {
```
mbayopanda, line 146: 👍

.travis.yml
```diff
@@ -18,6 +18,11 @@ matrix:
       services:
         - mysql
         - redis
+    - env: TEST_TYPE=e2e
+      services:
+        - redis
+      addons:
+        mariadb: '10.5'
```
mbayopanda, line 25: 👍

.travis.yml
```diff
@@ -18,6 +18,11 @@ matrix:
       services:
         - mysql
         - redis
+    - env: TEST_TYPE=e2e
```
mbayopanda, line 21: Does it means Travis will handle only e2e tests ?

mbayopanda (approved):
LGTM 👍

PR #4475, opened 2020-05-08
feat(stock requisition) add status filter

### LINKED ISSUE #4427: Filtering by "state" on in Stock Requsition doesn't seem to work
The inline filter in Stock Requisitions doesn't seem to filter on the "state" column.  Here is a demonstration:
YaynvBAALg

We should make this actually filter by the state of the requisition.

jeremielodi:
closes #4427

### DIFF
--- a/client/src/modules/stock/requisition/registry.js
+++ b/client/src/modules/stock/requisition/registry.js
@@ -2,9 +2,10 @@ angular.module('bhima.controllers')
   .controller('StockRequisitionController', StockRequisitionController);
 
 StockRequisitionController.$inject = [
-  'StockService', 'NotifyService', 'ModalService', 'ReceiptModal',
-  'uiGridConstants', 'StockModalService', 'LanguageService',
-  'GridStateService', 'GridColumnService', '$state',
+  '$state', 'StockService', 'NotifyService',
+  'ModalService', 'ReceiptModal',
+  'uiGridConstants', 'StockModalService',
+  'GridStateService', 'GridColumnService',
 ];
 
 /**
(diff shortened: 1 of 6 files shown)


client/src/modules/stock/requisition/templates/status.filter.html
```diff
@@ -0,0 +1,10 @@
+
+<div class="ui-grid-filter-container" ng-repeat="colFilter in col.filters">
+  <select class="ui-grid-filter-select" ng-model="colFilter.term" ng-options="option.value as (option.label | translate) for option in colFilter.options"></select>
+  <span style="display:inline-block;" class="pull-right">
+    <a data-reset-input href ng-click="grid.appScope.onClearStatusFilter(colFilter)" tabindex="-1">
```
jniles, line 6: 👍

client/src/modules/stock/requisition/registry.js
```diff
@@ -244,6 +254,17 @@ function StockRequisitionController(
    * action to perform when the page is loaded
    */
   function startup() {
+
+    Status.read().then(rows => {
+      vm.status = rows.map(s => {
+        return {
+          value : s.status_key,
```
jniles, line 258: This is awesome!

I propose two improvements:

1. Could you do the $translate.instant() here? It will help performance a bit.
2. Could you order to alphabetically after the $translate call? You could do it here with a sort() or in the template with an | orderBy: filter.

server/config/routes.js
```diff
@@ -784,6 +784,8 @@ exports.configure = function configure(app) {
   // @todo - this should use the JSON renderer instead of it's own route!
   app.get('/finance/cashflow', financeReports.cashflow.report);
 
+  app.get('/status', stock.listStatus);
```
jniles: These are only stock status's , right?  Could you make this route `/stock/status` to reflect that?

jniles (commented):
@jeremielodi this is an awesome PR. I was not expecting an implementation like this and it look super slick and nice.  Well done!

I've made some small suggestions to improve the UX.  It shouldn't be too difficult.

client/src/modules/stock/requisition/registry.js
jeremielodi: I had troubles while trying to use $translate.instant() here, using an html template helped a lot.

jeremielodi:
```js

 Stock.status.read().then(rows => {
      vm.status = rows.map(s => {
        return {
          value : s.status_key,
          label : $translate.instant(s.title_key),
        };
      });

 });
```
iNFRfZy

client/src/modules/stock/requisition/registry.js
jniles: Alright, it isn't a bit enough deal to keep this out.  Thanks for trying!

jniles (approved):
bors r+

PR #4476, opened 2020-05-08
fix: print colors in PDFs

### LINKED ISSUE #4382: Colors are not rendered in the PDF report with puppeteer

jniles:
Applies a color fix suggested by @jeremielodi, and omits bootstrap's broken `print.css`.  Instead, we've created our own `print.css` that will render the colors.  These changes should ensure that all colorschemes should print in each report without user intervention.

Also fixes the unrendered title of the cashflow report.

Closes #4382
Closes #4456

Here is what it looks like:
q5pUWAmUbm

### DIFF
--- a/client/src/less/bhima-bootstrap.less
+++ b/client/src/less/bhima-bootstrap.less
@@ -1,4 +1,58 @@
-@import "bootstrap.less";
+// These imports customize bootstrap's LESS files found in
+// node_modules/bootstrap/less/*.less
+// See the gulpfile's build process for more information
+
+// Core variables and mixins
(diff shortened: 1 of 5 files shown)


jniles:

jeremielodi (approved):
LGTM 👍

PR #4477, opened 2020-05-08
Add enterprise setting to change CMM calculation

### LINKED ISSUE #4346: Adjusting for Stockouts in the CMM/AMC
This issue is related to https://github.com/IMA-WorldHealth/bhima/issues/4344 and https://github.com/IMA-WorldHealth/bhima/issues/4327.

According to my research on Stock Management, the CMM/AMC should be "adjusted for stock outs".  The way BHIMA does this currently is by adjusting for lack of _exits_. In other words, if there is stock in the depot, but no exit occurred during a month, we remove that from the calculation.  That was the main idea because this issue: https://github.com/IMA-WorldHealth/bhima/issues/4327.

However, I think the calculation would be more accurate if we did two things:
  1. We transitioned from counting missing stock consumption as missing data to considering _stock outs_ as missing data.
  2. We started calculating the consumption for a month based on the daily consumption rather than the monthly consumption average.

We actually can't do (1) easily yet - I'm not sure BHIMA tracks the quantity in stock by day.  This would be a rather intense calculation.  We might need a table like `period_total` to hold these values.

Let me give you an example:

Let's assume that we give 50 pills of Amoxy to each patient every day in January.  Let's assume that we give 1 pill on February 1st, and then we run out of stock for the rest of February.  Then we order one pill and give it out on the 1st of March and have stock outs for the rest of March.

In the current calculation of BHIMA, we would have a CMM of: `(50 + 1 + 1) / 3 = 17.34`.

If we calculated this value by day, we would instead have:
  1. 31 days of 50 pills / day.
  2. 1 day of 1 pill / day
  3. 1 day of 1 pill / day.
  4. And the rest we exclude because we had stockouts.

This gives:  `( (50 x 31) + 1 + 1) / (31 + 1 + 1) = 47.03`.  Notice how this is much closer to our actual demand when we have sufficient stock (50 pills / day). 

@mbayopanda @lomamech @jeremielodi what do you think?

mbayopanda:
This PR is related to #4346 

Close #4346


mbayopanda:
@jniles this is a first temptative to resolve the issue #4346 could you give me a review and some other suggestions as usual to enhance the work

server/controllers/stock/core.js
```diff
@@ -508,6 +510,40 @@ async function getStockConsumptionAverage(periodId, periodDate, monthAverageCons
...
+        i.uuid inventory_uuid, i.text, i.code,
+        d.uuid depot_uuid, d.text AS depot_text
+      FROM stock_movement sm 
+      JOIN lot l ON l.uuid = sm.lot_uuid
+      JOIN inventory i ON i.uuid = l.inventory_uuid
+      JOIN depot d ON d.uuid = sm.depot_uuid
+      WHERE 
+        (DATE(sm.date) BETWEEN ? AND ?) AND flux_id IN (${flux.TO_PATIENT}, ${flux.TO_SERVICE})
+      GROUP BY d.uuid, i.uuid, sm.date
+      HAVING consumption <> 0
```
jniles: This is better than the previous version (considering the entire month), but we still want to only consider days where quantity in stock > 0, rather than consumption <> 0.  

However, if you intend this to be a step in the right direction, I can see this being a small improvement to the calculation.  What do you think?

client/src/i18n/en/enterprise.json
```diff
@@ -27,7 +27,9 @@
       "ENABLE_AUTO_STOCK_ACCOUNTING_HELP_TEXT" : "Enabling this feature will write stock movement transactions into the posting journal in real time. It requires all inventory accounts to be correctly configured.",
       "ENABLE_AUTO_EMAIL_REPORT_LABEL" : "Enable sending automatic reports by email",
       "ENABLE_AUTO_EMAIL_REPORT_HELP_TEXT" : "Enabling this option allows users to configure schedules to automatically send reports to a list of email addresses.",
-      "ENABLE_INDEX_PAYMENT_SYSTEM" : "Enabling index payment system"
+      "ENABLE_INDEX_PAYMENT_SYSTEM" : "Enabling index payment system",
+      "ENABLE_DAILY_CONSUMPTION_LABEL" : "Enable CMM calculation based on days of consumption",
+      "ENABLE_DAILY_CONSUMPTION_HELP_TEXT" : "The CMM is calculated taking into account the quantities consumed over the number of days on which there was consumption"
```
jniles, line 32: It would be good to mention what the default is.  So, by default, only months will be considered.

jniles (commented):
So far, this looks good to me.  I haven't done any formal tests.  Let me know if you would like me to do so.

mbayopanda:
@jniles you could do some formal tests

server/controllers/stock/core.js
```diff
@@ -508,6 +510,40 @@ async function getStockConsumptionAverage(periodId, periodDate, monthAverageCons
...
+      z.depot_uuid, z.inventory_uuid, z.code
+    FROM (
+      SELECT 
+        count(*) consumption, SUM(sm.quantity) quantity,  sm.date,
+        i.uuid inventory_uuid, i.text, i.code,
+        d.uuid depot_uuid, d.text AS depot_text
+      FROM stock_movement sm 
+      JOIN lot l ON l.uuid = sm.lot_uuid
+      JOIN inventory i ON i.uuid = l.inventory_uuid
+      JOIN depot d ON d.uuid = sm.depot_uuid
```
jniles: I doesn't look like you need to have a JOIN on depot here.  Can you move it to the outer SQL?  That way, do you the SUM()s and GROUP BY using the `depot_uuid`, then later JOIN on depot to get the text.

server/controllers/stock/core.js
jniles: We'll address this in another PR.  This is a significant enough improvement that we should merge it, in my view.

server/controllers/stock/core.js
```diff
@@ -508,6 +510,40 @@ async function getStockConsumptionAverage(periodId, periodDate, monthAverageCons
...
   `;
 
+  // the value (w.quantity/w.days) is the CMM
+  // is mentionned here as quantity to fit into the expected values
+  // for returned values
+  const queryConsumptionByDays = `
+  SELECT
+    w.consumption AS counted_consumption,
+    w.quantity AS consumed_quantity,
+    ROUND(w.quantity/w.days) AS quantity,
```
jniles: To make this into CMM we need to multiply by 30.5 days.

jniles (commented):
@mbayopanda I've validated your calculations with Vanga's dataset.   It works well!  Great job!

I've made one suggestion to improve the SQL.  It should be a pretty minor improvement.

As noted on Skype, the algorithm isn't respected for the Articles in Stock registry.  Before merging, we should figure out why that is and correct it.

Otherwise, this is an awesome PR.  Thanks!  Let me know when you want me to take another look.

server/controllers/stock/core.js
mbayopanda: There is a group by d.uuid,... which is useful for the `COUNT(*) consumption`

jniles: But you can group by `depot_uuid`, right?  Isn't that the same thing?

It doesn't look like you need to JOIN on the depot table to group on the uuid, since it is already in the `stock_movement` table.

server/controllers/stock/core.js
mbayopanda: @jniles `consumption` is not a quantity consumed but just the number of consumption movements of a day, and if we replace `consumption <> 0` by `quantity > 0` i think we will have a different result

jniles: Yes, this is what we want.  We want to count days with no movements, but with quantity in stock.  Those days will have a CMM of 0.  Will `quantity > 0` do that for us?

server/controllers/stock/core.js
```diff
@@ -508,6 +510,39 @@ async function getStockConsumptionAverage(periodId, periodDate, monthAverageCons
...
+      SELECT 
+        count(*) consumption, SUM(sm.quantity) quantity,  sm.date,
+        i.uuid inventory_uuid, i.text, i.code,
+        sm.depot_uuid
+      FROM stock_movement sm 
+      JOIN lot l ON l.uuid = sm.lot_uuid
+      JOIN inventory i ON i.uuid = l.inventory_uuid
+      WHERE 
+        (DATE(sm.date) BETWEEN ? AND ?) AND flux_id IN (${flux.TO_PATIENT}, ${flux.TO_SERVICE})
+      GROUP BY sm.depot_uuid, i.uuid, sm.date
```
jniles: This should be `GROUP BY  sm.depot_uuid, i.uuid, DATE(sm.date)`.

jniles (approved):
@mbayopanda, this is an improvement and I feel comfortable moving forward.  You need to add the `DATE()` function to actually group by days - right now it is grouping on TIMESTAMPs.

This doesn't quite achieve the full scope - because we are JOINing on `stock_consumption`, we are not getting the number of days without a stock out.  Maybe @lomamech 's PR will solve this in the future and make it easier to obtain.

jniles:
bors try

PR #4486, opened 2020-05-11
fix(purchase): make DECIMALs compatible with PJ/GL

### LINKED ISSUE #4478: Store more digits in the stock/purchase cost tables
We received this notice from our partners in HEV:
> J'ai une question à vous poser concernant la réception de la facture dans Bhima. Avez vous déjà augmenter la distance dans Bhima ??? parce que, lui n'acceptait pas un gros montant venant de la Pharmacie Centrale d'où, cela me demandais de diviser la facture en deux ou trois partie soit plus. En effet, il y'a une commande de 14.400.000 FC soit 8000$ qui vient à bientôt qu'on a commander avant hier.l'enregistrement s'effectuera comment ???.

The complaint is that they cannot enter in a huge values ($8000) into the purchase order table and move it into stock.  They need to break it into multiple stock entries to make it all fit.

This would be caused by the size of DECIMAL values in the database.  We should look whether or not this is a problem in the current schema and/or if this is an HEV specific issues.

jniles:
Fixes the purchase table to make unit price and total DECIMAL columns compatible with the DECIMAL values stored in the general_ledger and posting_journal. The previous iteration made the values far too small.

Closes #4478.


jniles:
bors try

lomamech (commented):
No comment

PR #4487, opened 2020-05-11
fix(reports): fixes indent on operating account

### LINKED ISSUE #4484: Compte d'Exploitation should indent child accounts of parents
In the following image, the highlighted account (641399) is the parent of all the child accounts below it (6413XX).  
image

You would never know this from the report.  In the account management module, we indent the accounts to show the parent/child relationship.   We should do the same thing for this report.  WE should also put the parents in bold.

jniles:
Fixes the indentation on the operating account report.  It also makes all titles bold.

Closes #4484.

Here is what it looks like: 
image

### DIFF
--- a/server/controllers/finance/reports/operating/index.js
+++ b/server/controllers/finance/reports/operating/index.js
@@ -1,10 +1,5 @@
-/**
- * NOTE(@jniles) - this is the comptes d'exploitation in French.
- */
-
-const q = require('q');
 const _ = require('lodash');
-const Exchange = require('../../../finance/exchange');
+const Exchange = require('../../exchange');
 const db = require('../../../../lib/db');
 const util = require('../../../../lib/util');
 const Tree = require('../../../../lib/Tree');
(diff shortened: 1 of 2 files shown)


jniles:

server/controllers/finance/reports/operating/index.js
```diff
@@ -23,31 +18,22 @@ const INCOME_ACCOUNT_TYPE = 4;
...
+ * Renders the Compte d'Exploitation
+ *
  * @param {*} options the report options
  * @param {*} session the session
  */
-function reporting(opts, session) {
-  const params = opts;
-  let docReport;
-  const options = _.extend(opts, {
+async function reporting(params, session) {
```
jeremielodi, line 29: Do you think it's a good idea to use this async without a ```try catch```?

jeremielodi (approved):

server/controllers/finance/reports/operating/index.js
jniles: The try/catch occurs in the function that calls this one - on this line: https://github.com/IMA-WorldHealth/bhima/pull/4487/files#diff-4c6917cb3623fce5a08ff0fa7b067112R153

jeremielodi: okay

PR #4488, opened 2020-05-11
fix(stock):  standardize requisition receipt

### LINKED ISSUE #4421: Improvements to Requisitions Receipts
Here is what a receipt for a requisition looks like:
image

In my view, this should look closer to a purchase order.  For reference, here is what a Purchase Order looks like:
image

Here are some improvements we should make to the requisition receipts:
  1. Make the font smaller!  Everything takes up way too much room.
  2. Move the `reference` to the top right corner.  We don't need the depot name up there.
  3. Add the "produced by {{user}} on {{timestamp}}" part to the top right corner.
  4. Add the type of inventory to the table.
  5. Add the total number of records to the bottom of the receipt.
  6. Add a small margin between the barcode and the description field.

jniles:
Makes the requisition receipt appear a bit more standard in spacing and
layout.  Removes whitespace and adds in the inventory type column.

Closes #4421.

Here is what it looks like:

image

### DIFF
--- a/server/controllers/stock/requisition/requisition.js
+++ b/server/controllers/stock/requisition/requisition.js
@@ -7,10 +7,10 @@ const util = require('../../../lib/util');
 const FilterParser = require('../../../lib/filter');
 
 const SELECT_QUERY = `
-  SELECT 
+  SELECT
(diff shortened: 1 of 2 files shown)


mbayopanda (approved):
LGTM 👍 

bors r+

PR #4492, opened 2020-05-12
fix(reports): improve Aux Cashbox Analysis

### LINKED ISSUE #4479: Analyse de Caisse Auxiliares Report doesn't work with Pax Cashboxes
We have an "Analyse de Caisse Auxiliares" report that should show the transactions between the auxiliary cashbox, the transfer account, and the principal cashbox.  It appears that this doesn't work at PAX when trying to look at transactions from the Pax Aux cashbox into the Caisse Coordination.  Instead, it always shows the "Caisse Principal" of HBB.   It is likely that this is hard-coded somewhere and we should make it configurable for our clients to be able to track their cash transactions.

jniles:
Various improvements to the Auxiliary Cashbox Analysis report.

 - Removes the Accounts from the client.  Not sure why we were downloading them.
 - Makes the title appropriately translated.
 - Removes comments/strings related to break even report and adds in Aux Cash Analysis text.
 - Migrates to async/await.
 - Uses UNION ALL.
 - Removes magic numbers and names the transaction type ID explicitly.
 - Spelling transferts -> transfers

Closes #4479.

### DIFF
--- a/client/src/modules/reports/generate/analysisAuxiliaryCash/analysisAuxiliaryCash.config.js
+++ b/client/src/modules/reports/generate/analysisAuxiliaryCash/analysisAuxiliaryCash.config.js
@@ -3,24 +3,17 @@ angular.module('bhima.controllers')
 
 AnalysisAuxiliaryCashController.$inject = [
   '$sce', 'NotifyService', 'BaseReportService', 'AppCache',
-  'reportData', '$state', 'AccountService', 'FormatTreeDataService',
+  'reportData', '$state',
 ];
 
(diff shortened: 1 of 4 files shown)


lomamech (approved):
I just checked it with IMCK data

lomamech:
@jniles  there is a conflict with this PR, as soon as it's fixed, I'll mix it up

jniles:
@lomamech, thanks for the review.  Just fixed the issues.

jniles:
@lomamech all green!

PR #4497, opened 2020-05-12
fix(invoice): link patient via debtor_uuid

### LINKED ISSUE #4414: Change invoice link to to patient to use uuid, not display name
Here is the code that [links to the patient registry in the invoice registry](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/invoices/registry/templates/action.cell.html#L26):
```html
      <a data-method="view-patient" href ui-sref="patientRegistry({ filters : [{ key : 'period', value : 'allTime' }, { key : 'display_name', value : row.entity.patientName }]})">
        <i class="fa fa-user"></i> <span translate>REPORT.VIEW_PATIENT</span>
      </a>
```

Here, it filters on `display_name`, which picks up a lot of other patients if there is a common name.  Instead, let's filter on a `uuid` - either the `debtor_uuid` or the `patient_uuid`, whichever is available.

We should also set `cacheable:false` in the filters.

jniles:
Makes the patient link from the invoice use the `debtor_uuid`.  This means that the lookup is both more performant and also more specific. Previously we picked up any number of patients in our lookup.

Closes #4414.

### DIFF
--- a/client/src/modules/invoices/registry/templates/action.cell.html
+++ b/client/src/modules/invoices/registry/templates/action.cell.html
@@ -23,7 +23,7 @@
 
     <!-- view linked records -->
     <li>
-      <a data-method="view-patient" href ui-sref="patientRegistry({ filters : [{ key : 'period', value : 'allTime' }, { key : 'display_name', value : row.entity.patientName }]})">
+      <a data-method="view-patient" href ui-sref="patientRegistry({ filters : [{ key : 'period', value : 'allTime' }, { key : 'debtor_uuid', value: row.entity.debtor_uuid, displayValue : row.entity.patientName, cacheable:false }]})">
         <i class="fa fa-user"></i> <span translate>REPORT.VIEW_PATIENT</span>
       </a>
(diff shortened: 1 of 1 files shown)


client/src/modules/invoices/registry/templates/action.cell.html
```diff
@@ -23,7 +23,7 @@
 
     <!-- view linked records -->
     <li>
-      <a data-method="view-patient" href ui-sref="patientRegistry({ filters : [{ key : 'period', value : 'allTime' }, { key : 'display_name', value : row.entity.patientName }]})">
+      <a data-method="view-patient" href ui-sref="patientRegistry({ filters : [{ key : 'period', value : 'allTime' }, { key : 'debtor_uuid', value: row.entity.debtor_uuid, displayValue : row.entity.patientName, cacheable:false }]})">
```
mbayopanda, line 26: 👍

mbayopanda (approved):
bors r+

PR #4502, opened 2020-05-13
fix(stock): movement search flux displayValues

### LINKED ISSUE #4316: Stock Exit "reason" filter does not preserve display value when reloaded
The stock exit "reason" filter displays a nice display value when you search/filter the first time, but then does not show a nice filter any other time.  See below:
7SWgHpfyb1

You can see that it show "from loss" at first, but when you add another parameter (like extending the time range), it shows `[7]` instead.

We should improve this by preserving the display value even when you search multiple times.

jniles:
Fixes the display values of the stock movement registry's flux ids so that they render better and are saved between page reloads.

Closes #4316.

FicgkMFeYM

### DIFF
--- a/client/src/modules/stock/movements/modals/search.modal.js
+++ b/client/src/modules/stock/movements/modals/search.modal.js
@@ -2,11 +2,11 @@ angular.module('bhima.controllers')
   .controller('SearchMovementsModalController', SearchMovementsModalController);
 
 SearchMovementsModalController.$inject = [
-  'data', 'NotifyService', '$uibModalInstance', 'FluxService', '$translate',
+  'data', 'NotifyService', '$uibModalInstance',
   'PeriodService', 'Store', 'util', 'StockService',
 ];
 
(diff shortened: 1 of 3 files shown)


client/src/modules/stock/movements/modals/search.modal.js
```diff
@@ -70,19 +56,17 @@ function SearchMovementsModalController(data, Notify, Instance, Flux, $translate
...
-        }
-      });
-    });
+  vm.onFluxChange = function onFluxChange(fluxes) {
+    const searchValue = fluxes.map(f => f.id);
+
+    // concats with a comma, replaces last comma
+    const displayValue = fluxes
+      .reduce((aggstr, flux) => aggstr.concat(flux.plainText, ', '), '')
+      .replace(/, $/i, '')
```
mbayopanda, line 65: 👍 Good catch

mbayopanda (approved):
This is a smooth refactoring 👍 

bors r+

PR #4504, opened 2020-05-13
fix link to installation

individual-it:
just a small fix, because the test-server does not offer HTTPS

-----
[View rendered docs/en/for-developers/installing-bhima.md](https://github.com/individual-it/bhima/blob/patch-1/docs/en/for-developers/installing-bhima.md)


jniles (approved):
LGTM.

bors r+

PR #4509, opened 2020-05-13
feat(Distribution center) save search filters

### LINKED ISSUE #4436: Distribution Centers should remember the parameters of search
I was trying to locate some transactions in the distribution centers modal and kept having to put in the fiscal year and periods over and over again.  It wasn't the best user experience.

We should have the modal remember what filters were applied such as what happens on most other registries.  Here is an example of how the module currently behaves:
xZogo5urPk

Let's change this behavior so that the filters are pre-selected on the modal.  The user then will only need to make updates to the previous values as needed.

jeremielodi:
closes #4436

### DIFF
--- a/client/src/js/components/bhFiscalPeriodSelect.js
+++ b/client/src/js/components/bhFiscalPeriodSelect.js
@@ -1,6 +1,9 @@
 angular.module('bhima.components')
   .component('bhFiscalPeriodSelect', {
     bindings : {
+      fiscalId : '<',
+      periodFromId : '<',
+      periodToId : '<',
       onSelectPeriodFromCallback : '&',
       onSelectPeriodToCallback : '&',
       onSelectFiscalCallback : '&',
@@ -19,6 +22,15 @@ function FiscalPeriodSelect(Fiscal, moment) {
       .then(fiscals => {
         $ctrl.fiscals = fiscals;
       });
+  };
+
+  $ctrl.$onChanges = (changes) => {
+
+    if (changes.fiscalId && $ctrl.fiscalId) {
+      $ctrl.selectedFiscal = $ctrl.fiscalId;
+      loadPeriodsForFiscalYear($ctrl.fiscalId);
+    }
+
   };
 
   $ctrl.loadPeriod = fiscalId => {
@@ -47,6 +59,16 @@ function FiscalPeriodSelect(Fiscal, moment) {
             return p;
           });
 
+        $ctrl.periods.forEach(period => {
+          if (period.id === $ctrl.periodFromId) {
+            $ctrl.selectedPeriodFrom = period;
+            $ctrl.onSelectPeriodFrom(period);
+          }
+          if (period.id === $ctrl.periodToId) {
+            $ctrl.selectedPeriodTo = period;
+            $ctrl.onSelectPeriodTo(period);
+          }
+        });
         // sure the period are ordered in an ASC fashion
         $ctrl.periods.sort(sortDates);
       });
--- a/client/src/modules/distribution_center/modals/setting_distribution.modal.html
+++ b/client/src/modules/distribution_center/modals/setting_distribution.modal.html
@@ -1,4 +1,4 @@
-<form name="SettingDistributionForm" bh-submit="SettingDistributionModalCtrl.submit(SettingDistributionForm)" novalidate>
+<form name="SettingDistributionForm" bh-submit="ModalCtrl.submit(SettingDistributionForm)" novalidate>
   <div class="modal-header">
     <ol class="headercrumb">
       <li class="title">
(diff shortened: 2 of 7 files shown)


client/src/js/components/bhFiscalPeriodSelect.js
```diff
@@ -33,7 +55,7 @@ function FiscalPeriodSelect(Fiscal, moment) {
     return -1;
   }
 
-  function loadPeriodsForFiscalYear(fiscalId) {
+  function loadPeriodsForFiscalYear(fiscalId, fn) {
```
jniles: This style is a bit weird to read, since we are mixing callbacks and promises.  Generally, you think about using one or the other.

I propose you do this operation in two steps.  First, you return `Fiscal.getPeriods()` which will return a promise.  Then you chain these callbacks with a `.then()` method.  The periods will be available in the `$ctrl.periods` variable, or you can pass them back.  Like so:
```js
return Fiscal.getPeriods( /* etc */);
```
... in the `$onChanges()` function:
```js
loadPeriodsForFiscalYear($ctrl.fiscalId)
  .then(periods => {
     periods.forEach(period => {
        if (period.id === $ctrl.periodFrom) {
          // etc 
        }
     });
  });

File: client/src/js/components/bhFiscalPeriodSelect.js [hunk 2]
```diff
@@ -1,6 +1,9 @@
 angular.module('bhima.components')
   .component('bhFiscalPeriodSelect', {
     bindings : {
+      fiscalId : '<',
```
jniles, line 4: Nice :+1:

client/src/js/components/bhFiscalPeriodSelect.js
```diff
@@ -1,6 +1,9 @@
...
+      periodFrom : '<',
+      periodTo : '<',
```
jniles: These are IDs, right?  It might make more sense to call them `periodFromId`, etc, to make that explicit.

client/src/modules/distribution_center/modals/setting_distribution.modal.html
```diff
@@ -1,4 +1,4 @@
-<form name="SettingDistributionForm" bh-submit="SettingDistributionModalCtrl.submit(SettingDistributionForm)" novalidate>
+<form name="SettingDistributionForm" bh-submit="ModalCtrl.submit(SettingDistributionForm)" novalidate>
```
jniles, line 1: I like this change 👍 much easier to read.

client/src/modules/distribution_center/modals/setting_distribution.modal.html
```diff
@@ -10,26 +10,29 @@
...
     <bh-fiscal-period-select
-      on-select-fiscal-callback="SettingDistributionModalCtrl.onSelectFiscal(fiscal)"
-      on-select-period-from-callback="SettingDistributionModalCtrl.onSelectPeriodFrom(period)"
-      on-select-period-to-callback="SettingDistributionModalCtrl.onSelectPeriodTo(period)">
+      fiscal-id = "ModalCtrl.distribution.fiscal"
+      period-from = "ModalCtrl.distribution.periodFrom"
+      period-to = "ModalCtrl.distribution.periodTo"
+      on-select-fiscal-callback="ModalCtrl.onSelectFiscal(fiscal)"
+      on-select-period-from-callback="ModalCtrl.onSelectPeriodFrom(period)"
+      on-select-period-to-callback="ModalCtrl.onSelectPeriodTo(period)">
```
jniles, line 18: A big improvement!

server/controllers/finance/distributionFeeCenter/configuration.js
```diff
@@ -7,61 +7,60 @@
...
+      excludes_distributed : true,
+      account_id : query.account_id,
+      trans_id : query.trans_id,
+      hrRecord : query.hrRecord,
+      limit : 100,
+    };
 
-      refAccounts = accounts;
-      const accountsId = accounts.map(account => account.account_id);
+    if (query.periodFrom && query.periodTo) {
```
jniles, line 34: Do we need this check now?  I think this will turn our period IDs into dates, which are much less efficient to query in SQL.  Instead of changing them into dates, can you just add some more filters that will handle this?  Maybe a `filters.custom()` or something?

server/controllers/finance/distributionFeeCenter/getDistributed.js
```diff
@@ -3,52 +3,65 @@
...
+    const options = req.query;
+
+    const { periodFrom, periodTo } = options;
+    options.is_cost = options.typeFeeCenter;
+    options.limit = 100;
+
+    if (options.fee_center_id || options.trans_id) {
+      delete options.limit;
+    }
+    if (periodFrom && periodTo) {
```
jniles, line 24: Same comment here about making period IDs into dates.

jniles (commented):
@jeremielodi here is my review.  @lomamech should take a look too since he is more familiar with the logic.

server/controllers/finance/distributionFeeCenter/configuration.js
```diff
@@ -7,61 +7,61 @@
...
+      excludes_distributed : true,
+      account_id : query.account_id,
+      trans_id : query.trans_id,
+      hrRecord : query.hrRecord,
+      limit : 100,
+    };
 
-      refAccounts = accounts;
-      const accountsId = accounts.map(account => account.account_id);
+    // get max and min date from period ids.
```
jniles, line 33: Why not just use the period IDs for start and stop?  It would be a lot more efficient.

server/controllers/finance/distributionFeeCenter/getDistributed.js
jniles: This still stands.  Waiting for either a change or a reply.

lomamech (approved):
Nice improvement :sparkles:

PR #4473, opened 2020-05-08
make bhima MySQL8 compatible

individual-it:
by hacking a PASSWORD equivalent
fixes #4463

PR #4511, opened 2020-05-14
catch MySQL exception and rethrow with error in case of incorrect value

### LINKED ISSUE #4501: HTTP result codes different when running on MySQL 8 vs. 5.7
when sending a string to the API instead of a int in some cases the API responses with HTTP 404 on MySQL 5.7 and HTTP 400 on MySQL 8

see this discussion for details: https://github.com/IMA-WorldHealth/bhima/pull/4473#issuecomment-627096996

individual-it:
these changes make sure the HTTP return codes are the same on MySQL 5.7 & 8, when trying to request, update or delete a row but sending an invalid id type.

fixes #4501

Only tests have been added to endpoints where the tests have been changed in #4473, so endpoints or actions that have been untested before are still untested and might still react different on different versions on MySQL

### DIFF
--- a/server/controllers/payroll/accounts/index.js
+++ b/server/controllers/payroll/accounts/index.js
@@ -81,19 +81,10 @@ function update(req, res, next) {
 
 // DELETE /ACCOUNT_CONFIG /:ID
 function del(req, res, next) {
-  const sql = `DELETE FROM config_accounting WHERE id = ?;`;
-
-  db.exec(sql, [req.params.id])
-    .then((row) => {
-    // if nothing happened, let the client know via a 404 error
-      if (row.affectedRows === 0) {
-        throw new NotFound(`Could not find a Account Configuration with id ${req.params.id}`);
-      }
-
-      res.status(204).json();
-    })
-    .catch(next)
-    .done();
+  db.delete(
+    'config_accounting', 'id', req.params.id, res, next,
+    `Could not find a Account Configuration with id ${req.params.id}`,
+  );
 }
 
 // get list of Account
--- a/server/lib/db/index.js
+++ b/server/lib/db/index.js
@@ -236,6 +236,27 @@ class DatabaseConnector {
   escape(key) {
     return mysql.escape(key);
   }
+
+  delete(table, idKey, idValue, res, next, notFoundErrorMessage) {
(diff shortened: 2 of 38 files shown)


server/controllers/payroll/accounts/index.js
```diff
@@ -82,16 +82,23 @@ function update(req, res, next) {
...
+  const notFoundErrorMessage = `Could not find a Account Configuration with id ${req.params.id}`;
   db.exec(sql, [req.params.id])
     .then((row) => {
     // if nothing happened, let the client know via a 404 error
       if (row.affectedRows === 0) {
-        throw new NotFound(`Could not find a Account Configuration with id ${req.params.id}`);
+        throw new NotFound(notFoundErrorMessage);
       }
 
       res.status(204).json();
```
jeremielodi: You can change this ```res.status(204).json();``` to  ```res.sendStatus(204);``` if there is no json to send back

test/integration/accountConfig.js
```diff
@@ -73,6 +73,14 @@ describe('(/payroll/account_configuration) The /payroll/account_configuration  A
       .catch(helpers.handler);
   });
 
+  it('DELETE /ACCOUNT_CONFIG/:ID will send back a 404 if the Account Configuration does receive a string instead of a number', () => {
+    return agent.delete('/account_config/str')
+      .then((res) => {
+        helpers.api.errored(res, 404);
```
jeremielodi, line 87: 👍

jeremielodi (commented):
@individual-it I propose to change ```deleteFromWhereId``` by just ```delete```  and you can pass the table name, the id key and value as parameters. This will help if we face the same issue with another table primary key whitch can be different to ```id```.

What do you think?

individual-it:
@jeremielodi done, please review again

server/lib/db/index.js
```diff
@@ -236,6 +236,27 @@ class DatabaseConnector {
   escape(key) {
     return mysql.escape(key);
   }
+
+  delete(table, idKey, idValue, res, next, notFoundErrorMessage) {
```
jeremielodi, line 240: 👍

jeremielodi (approved):
@jeremielodi approve this PR. Well done @individual-it . Thanks

PR #4527, opened 2020-05-19
feat(depot registry) add users columns

### LINKED ISSUE #4419: Add users to the Depot Management page
Similar to how we have the cashboxes and depots listed on the user page, we should add the users who have the depots available to them to the Depot Management page.  It would look something like this:

image

jeremielodi:
closes #4419

users


jniles (approved):
LGTM.

bors r+

PR #4530, opened 2020-05-20
deps: CORAL_REUSE_BROWSER flag for speedy PDF startup

jniles:
Bumps the coral dependency and adds the `CORAL_REUSE_BROWSER` flag to try and speed up PDF rendering on low-end devices.


mbayopanda (approved):
bors r+

PR #4534, opened 2020-05-21
feat(distribution fee center) translate grid footer

### LINKED ISSUE #4438: Distribution Fee Center doesn't have translated footer
The footer of the Distribution Fee Center grid always shows up in English.  We should remove the aggregation label so that we don't have to translate it.

image

jeremielodi:
closes #4438


jniles (approved):
LGTM

bors r+

PR #4538, opened 2020-05-23
feat(article en stock) link to inventory registry

### LINKED ISSUE #4418: Link the inventory registry from the "articles in stock" registry
It would be super helpful to link the inventory from the "Articles in Stock" registry.  That way, you can quickly jump in and see the inventory changes or other properties if you need to make changes.  It would go here:

image

jeremielodi:
closes #4418

### DIFF
--- a/client/src/modules/stock/inventories/templates/inventory.action.html
+++ b/client/src/modules/stock/inventories/templates/inventory.action.html
@@ -14,5 +14,15 @@
         <span class="fa fa-file-text-o"></span> <span translate>TREE.STOCK_MOVEMENTS</span>
       </a>
     </li>
+  
+    <li>
+      <a data-method="inventories" href
+        ui-sref="inventory.list({
+          filters : [
+          { key : 'uuid', value : row.entity.inventory_uuid, displayValue: row.entity.text,  cacheable: false }
+        ]})">
+        <i class="fa fa-cart-plus"></i> <span translate>TREE.INVENTORY_LIST</span>
+      </a>
+    </li>
   </ul>
 </div>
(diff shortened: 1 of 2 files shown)


client/src/modules/stock/inventories/templates/inventory.action.html
```diff
@@ -14,5 +14,17 @@
         <span class="fa fa-file-text-o"></span> <span translate>TREE.STOCK_MOVEMENTS</span>
       </a>
     </li>
+  
+    <li>
+      <a data-method="inventories" href
+        ui-sref="inventory.list({ 
+          uuid :  row.entity.inventory_uuid,
```
jniles: You don't need the keys `uuid`, `created`, or `updated`.

client/src/modules/stock/inventories/templates/inventory.action.html
```diff
@@ -14,5 +14,17 @@
...
+          created : false,
+          updated : false,
+          filters : [
+          { key : 'text', value : row.entity.text }]})">
```
jniles: Let's search based on UUID rather than on text.  Like this:
```html
<a ui-sref="inventory.list({ filters : [{ key : 'uuid', value : row.entity.inventory_uuid, displayValue: row.entity.text, cacheable: false }]})">
  ...
</a>

````

jniles: Note that, you'll need to add `uuid` to the list of registered filters in the [InventoryService](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/inventory/inventory.service.js#L115).

jniles (changes requested):
Just a few changes here.  Since we have the UUID, we might as well filter on it.  It would be more precise and faster.

jniles (approved):

PR #4539, opened 2020-05-24
docs: improve cloud-init script

jniles:
Removes wkhtmltopdf from the cloud-init setup process and adds in global variables to define different parts of the system (hostname, etc).

### DIFF
--- a/docs/cloud-init.yml
+++ b/docs/cloud-init.yml
@@ -1,5 +1,13 @@
 #cloud-config
 
+# NOTE:
+# This script uses several custom variables configured at the top of the run_cmd block:
+#   BH_DOMAIN - domain name where the server will be deployed. Used in nginx and letsencrypt config
+#   NODE_VERSION - node version to install.  This should be an integer (e.g. 12, 13, 14)
+#   BH_DATABASE_NAME - the name of the BHIMA database.  Often tied to the domain name.
+#   BH_DATABASE_PASSWORD - the password of the database. Defaults to $(uuid).
+#
+
 packages:
   - curl
   - nginx
@@ -24,9 +32,27 @@ users:
 write_files:
   - path: /tmp/bhima-mysql-user.sql
     content: |
-      CREATE USER 'bhima'@'localhost' IDENTIFIED BY 'HISCongo2013';
+      CREATE USER 'bhima'@'localhost' IDENTIFIED BY 'HISCongo2020';
       GRANT ALL PRIVILEGES ON * . * TO 'bhima'@'localhost';
       FLUSH PRIVILEGES;
+
+  - path: /tmp/.my.cnf
+    content: |
+      [client]
+      user=root
+      password=PASSWORD
+
+      default-character-set=utf8mb4
+
+      [mysql]
+      default-character-set=utf8mb4
+
+
+      [mysqld]
+      collation-server = utf8mb4_unicode_ci
+      init-connect='SET NAMES utf8mb4'
+      character-set-server = utf8mb4
+
 
   - path: /tmp/bhima-latest-release.sh
     content: |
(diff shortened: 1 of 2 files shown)


docs/en/for-developers/installing-bhima.md
```diff
@@ -51,6 +51,23 @@ sudo apt-get update && sudo apt-get install yarn --no-install-recommends
...
+
+```bash
+# Install puppeteer dependencies on Ubuntu
+sudo apt-get install ca-certificates fonts-liberation gconf-service \
+  libappindicator1 libasound2 libatk-bridge2.0-0 libatk1.0-0 libc6 libcairo2  \
+  libcups2 libdbus-1-3 libexpat1 libfontconfig1 libgbm1 libgcc1 libgconf-2-4 \
+  libgdk-pixbuf2.0-0 libglib2.0-0 libgtk-3-0 libnspr4 libnss3 libpango-1.0-0 \
+  libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxcomposite1 \
+  libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 \
+  libxss1 libxtst6 lsb-release wget -y
```
mbayopanda, line 66: 👍 hahaha

docs/cloud-init.yml
```diff
@@ -24,10 +32,28 @@ users:
...
-      CREATE USER 'bhima'@'localhost' IDENTIFIED BY 'HISCongo2013';
+      CREATE USER 'bhima'@'localhost' IDENTIFIED BY 'HISCongo2020';
       GRANT ALL PRIVILEGES ON * . * TO 'bhima'@'localhost';
       FLUSH PRIVILEGES;
 
+  - path: /tmp/.my.cnf
+    content: |
+      [client]
+      user=root
+      password=PASSWORD
```
mbayopanda, line 44: Where does the `PASSWORD` get its value ?

docs/cloud-init.yml
```diff
@@ -94,6 +120,12 @@ write_files:
 
 runcmd:
 
+# configure "global" variables
+  - export BH_DOMAIN="demo.bhi.ma"
+  - export NODE_VERSION="12"
```
mbayopanda, line 127: Why do we choose the version 12 ? if the reason is the latest LTS it can be useful to have a comment which mention that

docs/cloud-init.yml
```diff
@@ -24,10 +32,28 @@ users:
 write_files:
   - path: /tmp/bhima-mysql-user.sql
     content: |
-      CREATE USER 'bhima'@'localhost' IDENTIFIED BY 'HISCongo2013';
+      CREATE USER 'bhima'@'localhost' IDENTIFIED BY 'HISCongo2020';
```
mbayopanda: I don't know if it is possible now but if we can have the user `bhima` and its password in a env variable it will be very nice

docs/en/for-developers/installing-bhima.md
jniles: 🤷

docs/cloud-init.yml
jniles: Yeah, I should leave a comment about it.  I'll do that.

docs/cloud-init.yml
jniles: Good question.  It's hidden on [this line](https://github.com/IMA-WorldHealth/bhima/pull/4539/files#diff-aad866a9311fd8ed6d4c5e99baf24a1eR144).

The command is
```sh
sed -i "s/PASSWORD/$BH_DATABASE_PASSWORD/g" /tmp/.my.cnf
```

To my knowledge, we can't use variables in the `content` part of this script.

docs/cloud-init.yml
jniles: Yeah, you are right.  I kind of started that and didn't actually finish.  I'll do that.

jniles (commented):
@mbayopanda, thanks for the review.  You caught all the things I had left unfinished 🙇 

I'll fix those and ping you back.

docs/en/for-developers/installing-bhima.md
mbayopanda: There are some issues on linux if we do not install all these staff

jniles:
@mbayopanda this is now finished.  I've tested it and it should work pretty well.

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4524, opened 2020-05-19
fix stock value validation and exit from a custom date

mbayopanda:
This PR is related to #4513 and #4512 

Close #4513 
Close #4512

PR #4547, opened 2020-05-26
 fix stock value validation and exit from a custom date

### LINKED ISSUE #4512: Do not accept negative values for CMM Calculation
This works:
image

It should.  We should restrict this to positive (> 0) values.

### LINKED ISSUE #4513: Stock Exit changing the date does not refresh the lot selection
Changing the date on the stock exit module does not refresh the lot selection in the grid - you must manually remove and re-add the line for the changes to take effect.  See below for an example:
riexJZ7Y0W

jniles:
This PR is related to #4513 and #4512 

Close #4513 
Close #4512 

Closes #4524

### DIFF
--- a/client/src/modules/enterprises/enterprises.html
+++ b/client/src/modules/enterprises/enterprises.html
@@ -13,295 +13,295 @@
 
 <div class="flex-content">
   <div class="container-fluid">
-    <div class="row">
+    <form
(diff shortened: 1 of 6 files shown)


jniles:
This is essentially unchanged from #4524 except the test has been fixed.

bors r+

PR #4548, opened 2020-05-26
feat(report printing) allow black & white printing

### LINKED ISSUE #4506: Allow users to save reports with/without colors
With the changes landed in https://github.com/IMA-WorldHealth/bhima/pull/4476, all reports show up in vibrant color.  This is good and bad news.

The good news is that our color schemes look really good in PDF reports.

The bad news is that it will waste ink.  Particular things like stripped tables.  Users may prefer that the colors be turned off.   We should support this feature with an option on the settings page.

Thankfully, this should be trivial to implement.  Just add a client-side option that sends something like `?disableColors=1` in the report query, check for that on the server, and then [disable this line](https://github.com/IMA-WorldHealth/bhima/pull/4476/files#diff-8be1013b55ee7cd9611fa989b42e8558R106) in the CSS.  That alone should turn off all colors.

jeremielodi:
close #4506

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -402,6 +402,7 @@
             "DIAGNOSIS": "Diagnosis",
             "DIFFERENCE" : "Difference",
             "DISABLED": "Disabled",
+            "DISABLE_COLORS": "Black & white",
             "DISPLAY_NAME" : "Display Name",
             "DISTRIBUTABLE": "Distributable",
             "DISTRIBUTED": "Distributed",
--- a/client/src/i18n/fr/form.json
+++ b/client/src/i18n/fr/form.json
@@ -406,6 +406,7 @@
             "DIAGNOSIS": "Diagnostic",
             "DIFFERENCE" : "Différence",
             "DISABLED": "Désactivée",
+            "DISABLE_COLORS": "Noir et blanc",
             "DISPLAY_NAME" : "Nom",
             "DISTRIBUTABLE": "Distribuable",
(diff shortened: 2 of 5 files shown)


client/src/i18n/en/form.json
```diff
@@ -402,6 +402,7 @@
             "DIAGNOSIS": "Diagnosis",
             "DIFFERENCE" : "Difference",
             "DISABLED": "Disabled",
+            "DISABLE_COLORS": "Blac & white",
```
jniles: This should be "Black".

client/src/i18n/fr/form.json
```diff
@@ -406,6 +406,7 @@
             "DIAGNOSIS": "Diagnostic",
             "DIFFERENCE" : "Différence",
             "DISABLED": "Désactivée",
+            "DISABLE_COLORS": "Noir et blank",
```
jniles: Ce n'est pas "Blanc"?

server/lib/template/partials/head.handlebars
```diff
@@ -103,9 +104,8 @@
 
     body {
       margin: 1em;
-      -webkit-print-color-adjust: exact;
+      -webkit-print-color-adjust: {{{printColorAjustment}}};
```
jniles, line 107: Cool :+1:

jniles (commented):
Just some typos, otherwise, LGTM.  Can you include a screenshot as well?

jniles (approved):
Tested locally and works well.  Great job @jeremielodi 

bors r+

PR #4549, opened 2020-05-26
fix(reports): ensure filenames are URI encoded

jniles:
When report names are sent back to the client they must be URI encoded to be properly interpreted.


jniles:

jeremielodi:
Okay

Le ven. 5 juin 2020 à 13:48, Jonathan Niles <notifications@github.com> a
écrit :

jeremielodi (approved):
Tested and approved.

PR #4551, opened 2020-05-27
feat(Purchase registry) add link to stock inventory

### LINKED ISSUE #4417: Link purchase orders from "Articles in Stock" Registry
It would be nice to add a link to purchase orders for stock items from the the articles in stock registry.  It would show up right here:
image

Whoever tackles this could also tackle https://github.com/IMA-WorldHealth/bhima/issues/4305 at the same time.

The link should be to either:
  1. The purchase order that created this lot (if possible) OR
  2. Filter the purchase orders on the inventory item of this lot.

jeremielodi:
closes  #4417

### DIFF
--- a/client/src/modules/purchases/templates/action.cell.html
+++ b/client/src/modules/purchases/templates/action.cell.html
@@ -28,6 +28,16 @@
         <i class="fa fa-arrows-h"></i> <span translate>PATIENT_REGISTRY.STOCK_MOVEMENTS</span>
       </a>
     </li>
+    
+    <li>
+      <a data-method="view-stock-lot" href
+        ui-sref="stockInventories({ filters : [
+          { key : 'period', value : 'allTime' },
+          { key : 'includeEmptyLot', value : 1 },
+          { key : 'purchase_uuid', value : row.entity.uuid, displayValue: row.entity.reference }]})">
+        <i class="fa fa-cart-plus"></i> <span translate>REPORT.VIEW_ITEM_STOCK</span>
+      </a>
+    </li>
 
     <li ng-if="!row.entity.hasStockMovement">
       <a


client/src/modules/stock/inventories/registry.js
```diff
@@ -212,6 +212,9 @@ function StockInventoriesController(
     vm.hasError = false;
     vm.loading = true;
 
+    if (filters.purchase_uuid) {
+      filters.includeEmptyLot = 1;
```
jniles: Can we just send this from the client side instead of putting it on the server?

jniles (commented):
This is actually the opposite of https://github.com/IMA-WorldHealth/bhima/issues/4417.  That issue was trying to go from stock -> purchase orders, but this PR goes from purchase -> stock.  It's still a good PR though.  It just doesn't close the issue.

client/src/modules/stock/inventories/registry.js
jeremielodi: okay

client/src/modules/purchases/templates/action.cell.html
```diff
@@ -28,6 +28,16 @@
         <i class="fa fa-arrows-h"></i> <span translate>PATIENT_REGISTRY.STOCK_MOVEMENTS</span>
       </a>
     </li>
+    
+    <li>
+      <a data-method="view-stock-lot" href
+        ui-sref="stockInventories({ filters : [
+          { key : 'period', value : 'allTime' },
+          { key : 'includeEmptyLot', value : 0 },
```
jniles: I think this should be `value : 1`.

server/controllers/stock/core.js
```diff
@@ -585,7 +585,9 @@ function getInventoryQuantityAndConsumption(params, monthAverageConsumption, ena
     _status = params.status;
     delete params.status;
   }
-
+  if (params.purchase_uuid) {
```
jniles: Could you remove this from the server?

jniles (commented):
@jeremielodi, looks like you got half-way there.  Just a few more changes.

Also, could you remove the "closes" from the commit?  It won't close that issue.  You can do it with `git commit --amend`.

client/src/modules/purchases/templates/action.cell.html
jeremielodi: okay

jniles (approved):
LGTM.

bors r+

PR #4552, opened 2020-05-27
refactor: remove `event` table

### LINKED ISSUE #4526: Remove all references to "events"
We have this `events` table that I thought was cute once.   The originally idea was to use an event stream to broadcast messages in real time to clients.  We've never actually encountered a production need for this and it is just more bloat.  Let's tear out all code that uses the `event` table and the `event` table itself.

jniles:
We've never used the event table.  This removes it from the schema.

Closes #4526.


jeremielodi (approved):
LGTM 👍

PR #4553, opened 2020-05-27
feat(Payroll config) set ORDER BY dateTo DESC

### LINKED ISSUE #4349: Order "Periode de Paiement" from the most recent to the oldest in Multiple Payroll
Just a suggestion - when you use Multiple Payroll, you almost always want to select the most recent pay period.  Let's order them from the most recent to the oldest.  Here is what they look like now:
image

I think it makes more sense to reverse them.

jeremielodi:
closes #4349


jniles (approved):
LGTM.

bors r+

PR #4556, opened 2020-05-27
bug(Stock Value report crashes server with too much data)

### LINKED ISSUE #4116: Stock Value report crashes server with too much data
As reported by our friends in Vanga, the stock value report appears to crash the server for the "pharmacie usage", the most trafficked depot in the institution.  On my local machine, MySQL pins the CPU at 100% and eventually the server dies.

It looks like the entire report is in a [stored procedure,](https://github.com/IMA-WorldHealth/bhima/blob/master/server/models/procedures/stock.sql#L275) which is good, but uses a for loop instead of JOINs/GROUP BYs, which is bad.  It appears this report is unusable until rewritten.

lomamech:
- Improvement and optimization of the stock value by deposit report

closes #4116


lomamech:
@jniles @mbayopanda  I managed to optimize the report, please do the review

migration-hev_mai_2020.sql
```diff
@@ -0,0 +1,9 @@
+SET names 'utf8mb4';
+SET character_set_database = 'utf8mb4';
+SET collation_database = 'utf8mb4_unicode_ci';
+SET CHARACTER SET utf8mb4, CHARACTER_SET_CONNECTION = utf8mb4;
+SET collation_connection = 'utf8mb4_unicode_ci';
```
jniles: Is this really necessary?  The SQL dump should be exporting with these defaults. Is that not the case?

server/controllers/stock/reports/stock/value.js
```diff
@@ -30,21 +30,109 @@ async function reporting(_options, session) {
...
+
+  /*
+   * The following query collects all the data related to the movement
+   * of stocks of an item in order to calculate the stock value
+   * With the aim of solving the problem, count the total number of
+   * entries into stock *quantityIn* as well as that of exit
+   * but also the cumulative value of entries *quantityCum* and exits *cumOut*
+   * and the quantity in stock after each movement of stock *quantityStock*
+  */
+  const stockValues = await db.exec(sqlGetInventories, [db.bid(options.depot_uuid), options.dateTo]);
```
jniles, line 56: Sp, if I am understanding this correctly, this code will generate `1 + N` SQL queries in a transaction, where `N` is the number of inventory items in a depot.  At Vanga, the total number of inventory items is currently 784.  At IMCK, the total number is 1302.  So we could potentially get hundreds of SQL queries launched from a single query here.

... I'm not sure how optimal this is in low-end hardware like a Raspberry Pi.

server/controllers/stock/reports/stock/value.js
```diff
@@ -30,21 +30,109 @@ async function reporting(_options, session) {
...
+        item.newCumValue = item.valueStock + lastCumValue;
+        item.stockUnitCost = item.newQuantityCum ? item.newCumValue / item.newQuantityCum : unitCost;
+
+        unitCost = item.stockUnitCost;
+        lastCumValue = item.newCumValue;
+      }
+    });
+  });
+
+  let stockTolal = 0;
```
jniles: This looks like a typo - it should be "stockTotal"

server/controllers/stock/reports/stock_value.report.handlebars
```diff
@@ -4,61 +4,56 @@
...
-        </tbody>
-
-      </table>
+          </thead>
+          <tbody>
+            {{#each stockValues}}
+              <tr>
+                <td style="width:50%">{{inventory_name}}</td>
+                <td class="text-right">{{stockQtt}}</td>
+                <td class="text-right">{{currency (multiply stockUnitCost ../rate) ../currency_id}}</td>
```
jniles: Can you do this multiplication before putting it in the template?

server/controllers/stock/reports/stock_value.report.handlebars
```diff
@@ -4,61 +4,56 @@
...
+                <td class="text-right">{{currency (multiply stockValue ../rate) ../currency_id}}</td>
+              </tr>
+            {{/each}}
+            <!-- no data -->
+            {{#if emptyResult}}
```
jniles: There is no need for this "emptyResult" check.  Just use `{{#each}}{{else}}{{/each}}` as discussed in WhatsApp.

So, it would look like this:
```hbs
{{#each stockValues}}
   {{! ... etc ... }}
{{else}}
  <tr><th class="text-center" colspan="4">{{translate 'STOCK.NO_DATA'}}</th></tr>
{{/each}}
```

server/controllers/stock/reports/stock/value.js
```diff
@@ -30,21 +30,109 @@ async function reporting(_options, session) {
 
   const report = new ReportManager(STOCK_VALUE_REPORT_TEMPLATE, session, optionReport);
 
+
   const options = (typeof (_options.params) === 'string') ? JSON.parse(_options.params) : _options.params;
   data.dateTo = options.dateTo;
   data.depot = await db.one('SELECT * FROM depot WHERE uuid=?', [db.bid(options.depot_uuid)]);
-  const stockValues = await db.exec('CALL stockValue(?,?,?);', [
```
jniles, line 36: If we are not longer using the `stockValue()` procedure, let's remove it from the stored procedures.

server/controllers/stock/reports/stock/value.js
```diff
@@ -30,21 +30,109 @@ async function reporting(_options, session) {
...
+        inventory.stockValue = lastMovement.newCumValue;
+
+        stockTolal += inventory.stockValue;
+      }
+    });
+  });
 
   const stockValueElements = options.exclude_zero_value
-    ? stockValues[0].filter(item => item.stockValue > 0) : stockValues[0];
+    ? stockValues.filter(item => item.stockValue > 0) : stockValues;
```
jniles, line 105: Is there anyway we can apply this filter higher in logic?  Preferably in the SQL.  It would probably improve performance.

jniles (commented):
I tried to test this, but it still crashes for me.  My virtual machine only has 4GB of RAM, but that should be more than enough.  After 2 minutes, the request times out and retries.  It also pegs my MySQL CPU usage at over 300%.

image

I'm not sure this solves the problem at all.  I still think, as mentioned in our previous discussion, that something like `period_total` is the way to go.  That way, we are doing very little computation in the actual report itself.

caKfGd60mv

lomamech:
> I tried to test this, but it still crashes for me. My virtual machine only has 4GB of RAM, but that should be more than enough. After 2 minutes, the request times out and retries. It also pegs my MySQL CPU usage at over 300%.
> 
> image
> 
> I'm not sure this solves the problem at all. I still think, as mentioned in our previous discussion, that something like `period_total` is the way to go. That way, we are doing very little computation in the actual report itself.
> 
> caKfGd60mv

With me it works really well

m0p7jfOYIa

migration-hev_mai_2020.sql
lomamech: > Is this really necessary? The SQL dump should be exporting with these defaults. Is that not the case?

This portion of code was added when I rebase with master, I haven't added it personally

server/controllers/stock/reports/stock/value.js
lomamech: the way I understood the logic of the stock value report came down to retrieving the last line of the stock card report, my idea was to launch several requests on the stock movement table in order to calculate for each unit cost weight and the stock value after each transaction

lomamech: I had given up on the idea of using an aggregation table because of the risk of having backdated stock movements

jniles:
That's because you have a machine with 16GB of RAM that cost many, many dollars.  We can't expect poor hospitals to buy these kinds of computers.

server/controllers/stock/reports/stock/value.js
lomamech: it will be difficult seeing that the products we see the stock shortage of a product after on the last line, after having noted all the entries and exits from Stock

lomamech:
> That's because you have a machine with 16GB of RAM that cost many, many dollars. We can't expect poor hospitals to buy these kinds of computers.

I think the problem stems from the way in which stock managers calculate the stock value of a product, I fear that even if we only had computers with large capacities, even after 10 or 15 years. use, that the stock value report then crashes out again

mbayopanda (commented):
image

it seems like stock_sheet report is not available

lomamech:
@jniles  i need a review

jniles (approved):
This is much better for performance.  The query takes 10 seconds on my machine (22 seconds total time).  It also doesn't peg MySQL at 300% CPU usage.  See below.
 
8wged7bEs0

I still think we can do better via `period_total` style table, but this fixes the bug in production to an acceptable level where the report is usable.  Great work!

bors r+

PR #4560, opened 2020-05-28
feat(stock lot) add link to stock inventory

### LINKED ISSUE #4493: Link the "Lots in Stock" from the dropdown of the "Articles in Stock"
We should add a link to the "Lots in Stock" from the "Articles in Stock" registry.  It would go here:
image

jeremielodi:
closes #4493


jniles (approved):
LGTM.

bors r+

PR #4561, opened 2020-05-28
refractor(fee distribution key registration) display rate width %

### LINKED ISSUE #4435: Taux/Rate column in Distribution Keys should have % symbol
The taux column doesn't tell you what the ratio is.  I think it is a percentage - @lomamech can confirm.  If it is a percentage, we should put % next to the numbers so that it doesn't confuse users.
image

jeremielodi:
closes #4435

### DIFF
--- a/client/src/modules/distribution_center/distribution_key/distribution_key.js
+++ b/client/src/modules/distribution_center/distribution_key/distribution_key.js
@@ -46,22 +46,15 @@ function DistributionKeyController(DistributionCenters, Notify, uiGridConstants,
       headerCellFilter : 'translate',
     }, {
       field : 'rate',
-      type : 'number',
       displayName : 'TABLE.COLUMNS.RATE',
       headerCellFilter : 'translate',
-      cellClass : 'text-right',
-      footerCellClass : 'text-right',
-      cellFilter : 'number: 2',
-      footerCellFilter : 'number:2',
-      enableFiltering : true,
+      cellTemplate : '/modules/distribution_center/templates/rate.tmp.html',
       treeAggregationType : uiGridGroupingConstants.aggregation.SUM,
-      customTreeAggregationFinalizerFn : (aggregation) => {
-        aggregation.rendered = aggregation.value;
-      },
     }, {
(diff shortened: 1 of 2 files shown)


client/src/modules/distribution_center/distribution_key/distribution_key.js
```diff
@@ -46,22 +46,14 @@ function DistributionKeyController(DistributionCenters, Notify, uiGridConstants,
...
       field : 'rate',
-      type : 'number',
       displayName : 'TABLE.COLUMNS.RATE',
       headerCellFilter : 'translate',
-      cellClass : 'text-right',
-      footerCellClass : 'text-right',
-      cellFilter : 'number: 2',
-      footerCellFilter : 'number:2',
-      enableFiltering : true,
-      treeAggregationType : uiGridGroupingConstants.aggregation.SUM,
```
lomamech, line 57: This change is cool, please only restore **treeAggregationType : uiGridGroupingConstants.aggregation.SUM,**

client/src/modules/distribution_center/distribution_key/distribution_key.js
```diff
@@ -46,22 +46,15 @@ function DistributionKeyController(DistributionCenters, Notify, uiGridConstants,
...
-      customTreeAggregationFinalizerFn : (aggregation) => {
-        aggregation.rendered = aggregation.value;
-      },
+      cellTemplate : '/modules/distribution_center/templates/rate.tmp.html',
     }, {
       field : 'user_name',
       displayName : 'TABLE.COLUMNS.RESPONSIBLE',
       headerCellFilter : 'translate',
+      enableFiltering : true,
+      treeAggregationType : uiGridGroupingConstants.aggregation.SUM,
```
jniles: This is on the wrong column.  It doesn't make sense to SUM the username.

jniles (changes requested):
So, because this uses a custom template, we now no longer have the aggregate function working.

image

I propose that instead of making a new template, we just use the angular `percentage` filter.  It should be about 3 lines of code.

PR #4564, opened 2020-06-01
refracor(Stock) rename "Sold out" to "Stock out"

### LINKED ISSUE #4443: Rename "Sold Out" to "Stock Out"
In general, we talk about stock outs rather than sold out, since stock consumption doesn't necessarily correspond to sales, and the "stock out" terminology is more prevalent in supply chain management.  Finally, our consumption also can be theft or expiry, which isn't the result of sales.

To reduce confusion, let's rename everything that says "Sold Out" to "Stock Out" or "Out of Stock".

jeremielodi:
closes #4443


jniles (approved):
LGTM.

Issue #4507, opened 2020-05-13
Improvements to Stock Entry Report


jniles:
Here is what our stock entry report looks like:
![image](https://user-images.githubusercontent.com/896472/81815461-e82dda80-9521-11ea-9e28-4686b54bae2f.png)

There are a couple of improvements that can be made:
  1. What is the date for?  Let's label it as "entry date".
  2. Let's put the inventory code as well as the label above the item.
  3. Let's make a separate column for the type of item, rather than putting it in parenthesis.  Either that, or put it in the title with the inventory code and label.

jniles:
Closed in https://github.com/IMA-WorldHealth/bhima/pull/4566

PR #4566, opened 2020-06-01
Enhance the stock entry report

mbayopanda:
Fix the issue #4507

### DIFF
--- a/server/controllers/stock/reports/stock_entry.report.handlebars
+++ b/server/controllers/stock/reports/stock_entry.report.handlebars
@@ -20,14 +20,21 @@
 
       <!-- stock entry from purchase -->
       {{#if includePurchaseEntry}}
-        <h3>{{translate 'STOCK.RECEIPT.ENTRY_PURCHASE'}}</h3>
-
         <table class="table table-condensed table-report">
+          <thead>
+            <tr style="background-color: #EFEFEF;">
+              <th colspan="3"><h3 style="margin: 0px;">{{translate 'STOCK.RECEIPT.ENTRY_PURCHASE'}}</h3></th>
+              <th>{{translate 'STOCK.ENTRY_DATE'}}</th>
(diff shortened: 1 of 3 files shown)


server/controllers/stock/reports/stock_entry.report.handlebars
```diff
@@ -161,48 +195,11 @@
...
-            <tr><th class="text-center">{{translate 'STOCK.NO_DATA'}}</th></tr>
+            <tr><th colspan="6" class="text-center">{{translate 'STOCK.NO_DATA'}}</th></tr>
           {{/if}}
         </table>
       {{/if}}
 
-      <br/>
-
-      <!-- stock entry as loss -->
-      {{#if includeLossEntry}}
```
jniles, line 172: 😆

server/controllers/stock/reports/stock_entry.report.handlebars
```diff
@@ -125,21 +149,30 @@
...
+      <br>
+
       <!-- stock entry to other depot -->
       {{#if includeTransferEntry}}
-        <h3>{{translate 'STOCK.RECEIPT.ENTRY_DEPOT'}}</h3>
-
         <table class="table table-condensed table-report">
+          <thead>
+            <tr style="background-color: #EFEFEF;">
+              <th colspan="3"><h3 style="margin: 0px;">{{translate 'STOCK.RECEIPT.ENTRY_DEPOT'}}</h3></th>
```
jniles, line 164: It looks like this key might not be defined correctly.

image


Shouldn't this say something about from other depot?

jniles (commented):
image


This is what it looks like if anyone else was wondering.

Looks really good!  Could you check  up on that one key?  Otherwise, LGTM.

jniles:
LGTM.

Bors r+

PR #4569, opened 2020-06-03
fix(bhYesNo)

lomamech:
- set property name in all case when it empty in using of
  bh-yes-no-radios

### DIFF
--- a/client/src/modules/enterprises/enterprises.js
+++ b/client/src/modules/enterprises/enterprises.js
@@ -106,9 +106,11 @@ function EnterpriseController(Enterprises, util, Notify, Projects, Modal, Scroll
     const creation = (vm.hasEnterprise === false);
     const changes = util.filterFormElements(form, true);
 
-    if (changes.month_average_consumption) {
-      delete changes.month_average_consumption;
-    }
+    Object.keys(vm.enterprise.settings).forEach(key => {
+      if (changes[key]) {
+        delete changes[key];
+      }
+    });
 
     changes.settings = angular.copy(vm.enterprise.settings);
 
--- a/client/src/modules/enterprises/enterprises.js
+++ b/client/src/modules/enterprises/enterprises.js
@@ -107,9 +107,7 @@ function EnterpriseController(Enterprises, util, Notify, Projects, Modal, Scroll
     const changes = util.filterFormElements(form, true);
 
     Object.keys(vm.enterprise.settings).forEach(key => {
-      if (changes[key]) {
-        delete changes[key];
-      }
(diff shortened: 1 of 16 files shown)


lomamech:
3h4C4ybmlm
@mbayopanda  it's done

mbayopanda (commented):
The behavior on the UI is fine now, just there is an error after changing some options some more data are sent into the enterprise table such as `enable_price_lock` and `enable_auto_email_report` which throw an error

image

lomamech:
> The behavior on the UI is fine now, just there is an error after changing some options some more data are sent into the enterprise table such as `enable_price_lock` and `enable_auto_email_report` which throw an error
> 
> image

@mbayopanda I just did the test, and it works well

mbayopanda:
Here is an error in the console : 
image

Here what i did : 
TIoRw1QQwc

mbayopanda:
@lomamech i suppose that with the fact that you add the name property, all these fields are included in enterprise object during the submission, maybe you can try to remove all these variables which are in `changes`. Maybe something like : 
```js
delete changes.enable_price_lock;
delete changes.enable_prepayments;
...
```
These variables are for enterprise settings and not the enterprise table, during the submission they are sent into the enterprise table

client/src/modules/enterprises/enterprises.js
```diff
@@ -106,9 +106,11 @@ function EnterpriseController(Enterprises, util, Notify, Projects, Modal, Scroll
...
     const changes = util.filterFormElements(form, true);
 
-    if (changes.month_average_consumption) {
-      delete changes.month_average_consumption;
-    }
+    Object.keys(vm.enterprise.settings).forEach(key => {
+      if (changes[key]) {
+        delete changes[key];
+      }
+    });
```
mbayopanda, line 111: I think you can just have inside the forEach loop : 
```js
delete changes[key];
```
and remove the if condition because `if (changes[key])` will ignore where `changes[key]==0`

client/src/modules/enterprises/enterprises.js
```diff
@@ -107,9 +107,7 @@ function EnterpriseController(Enterprises, util, Notify, Projects, Modal, Scroll
     const changes = util.filterFormElements(form, true);
 
     Object.keys(vm.enterprise.settings).forEach(key => {
-      if (changes[key]) {
-        delete changes[key];
-      }
+      delete changes[key];
```
mbayopanda, line 110: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4570, opened 2020-06-04
bh-depot-select : do not send empty search text parameter

### LINKED ISSUE #4557: Do not sent empty search params to /depots/search/name
While testing https://github.com/IMA-WorldHealth/bhima/pull/4556, I noticed that the depot search component sometimes sends an empty field to the server.  The query looks like this `/depots/search/name?text=`.  We should prevent the query from sending an empty field - either do not send the query component (`?text=`) or don't send an HTTP request at all, whichever behavior makes more sense in this use case.

mbayopanda:
Fix the issue #4557

Close #4557

### DIFF
--- a/client/src/js/components/bhDepotSelect/bhDepotSelect.js
+++ b/client/src/js/components/bhDepotSelect/bhDepotSelect.js
@@ -39,6 +39,10 @@ function DepotSelectController(Depots, Notify) {
   };
 
   $ctrl.searchByName = text => {
+    if (!text) {
+      return null;
+    }
+
     const options = {
       text : (text || '').toLowerCase(),
       exception : $ctrl.exception,


kwilu:
Hello @mbayopanda, it seems like you have referenced #4557 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #4557.`.

```
Author: kwilu

    Fixes #51.
```

jniles (approved):
This works!

Another way to do this is by changing:
```html
typeahead-min-length="0"
```
to
```html
typeahead-min-length="1"
```
in [the template](https://github.com/IMA-WorldHealth/bhima/blob/c57e6873231c0001f0cf29f0e637b1c910e2f154/client/src/js/components/bhDepotSelect/bhDepotSelect.tmpl.html#L22).

bors r+

PR #4571, opened 2020-06-04
bug(Selector Ipr tax)

### LINKED ISSUE #4559: Eliminate Duplicate Selectors found in IPR Tax Management
This somewhat worries me:
image

We are finding duplicate selectors somewhere in a dropdown menu, and I'm concerned that it will lead to unpredictable results in our testing.  Someone should look into this and provide further specificity.

lomamech:
- Add property id ip uib-dropdown-toggle in bhIprScale
- Use select by id in dropdown for the select of ipr_scale

closes #4559

### DIFF
--- a/client/src/modules/templates/bhIprScale.tmpl.html
+++ b/client/src/modules/templates/bhIprScale.tmpl.html
@@ -1,5 +1,5 @@
 <div class="btn-group" uib-dropdown dropdown-append-to-body>
-  <a class="btn btn-default" uib-dropdown-toggle>
+  <a id="ipr_scale" class="btn btn-default" uib-dropdown-toggle>
     <span translate>IPRTAX.IPR_SCALE</span>
   </a>
   <button style="min-width : 30px" class="btn btn-default">
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -33,11 +33,6 @@ const config = {
     },
   },
 
-  plugins : [{
-    package : 'protractor-console-plugin',
-    failOnError : false,
-  }],
-
   // this will log the user in to begin with
   onPrepare : () => {
     return q.fcall(async () => {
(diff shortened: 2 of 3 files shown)


lomamech:
@jniles  or @mbayopanda  can you if possible review this PR

protractor.conf.js
```diff
@@ -33,11 +33,6 @@ const config = {
     },
   },
 
-  plugins : [{
```
jniles, line 36: Good 👍

client/src/modules/templates/bhIprScale.tmpl.html
```diff
@@ -1,5 +1,5 @@
 <div class="btn-group" uib-dropdown dropdown-append-to-body>
-  <a class="btn btn-default" uib-dropdown-toggle>
+  <a id="ipr_scale" class="btn btn-default" uib-dropdown-toggle>
```
jniles, line 2: This won't solve the issue, since this also isn't unique.  Every dropdown will have `id="ipr_scale"`.  Also IDs are supposed to be unique in HTML.

Can you use something from the `row` object? Like `{{row.entity.id}}`?  Whatever the column is that uniquely identifies the IPR Tax Scale in the database should be used here to select this particular row from the grid.

Alternatively, if there is a unique text (like `display_name` or `label` or something) that would work too.  It just needs to uniquely identify this row.

jniles (changes requested):
This is the right idea, but we still need to make the selector unique.

lomamech:
@jniles according to my understanding the first click concerned the dropdown scale well before selecting the existing IPR scales

image

jniles (approved):

jniles:
Aha,  I see now.  You are right - this will solve the issue.  Apologies for the confusion!

bors r+

PR #4574, opened 2020-06-04
feat(voucher registry)Link the credit note from the reversed record

### LINKED ISSUE #4482: Link the credit note from the reversed record in the Voucher Registry
Here is an example voucher that has been reversed.  
image

It would be nice if we could link, in the dropdown, the voucher that reversed this one.  That way, we could quickly find the reversing record and any text that accompanies it.

jeremielodi:
closes #4482


jniles (approved):
This looks right to me!

bors r+

PR #4582, opened 2020-06-05
The version 1.13.0 of bhima

mbayopanda:
bhima 1.13.0


server/models/migrations/v1.12.0-v1.13.0/migrate.sql
```diff
@@ -0,0 +1,27 @@
+-- migrate next.sql
```
jniles, line 1: Let's make this v1.12.0 - v1.13.0.

It just makes it easier to read if you `cat *.sql > migrate.sql`

jniles:
Looks like you've already tagged this - https://github.com/IMA-WorldHealth/bhima/releases/tag/v1.13.0.   Also [it doesn't include the yes/no fix](https://github.com/IMA-WorldHealth/bhima/compare/v1.13.0...master)!  We should release 1.13.1 to make sure this bug gets fixed before pushing to our clients.

PR #4591, opened 2020-06-09
fix(Fee Centers): proper currency rendering

### LINKED ISSUE #4437: Distribution Fee Center should show the correct currency for the transaction
Here is an example transaction imported via the Distribution Fee Center module. 
image

These values are in (potentially different) currencies.  We should make this clear by putting the currency symbols on the values.

jniles:
Adds in the appropriate currency rendering for the columns of the Distribution Fee Center.

Closes #4437.

image

### DIFF
--- a/client/src/modules/distribution_center/distribution_center.js
+++ b/client/src/modules/distribution_center/distribution_center.js
@@ -3,7 +3,7 @@ angular.module('bhima.controllers')
 
 DistributionCenterController.$inject = [
   'DistributionCenterService', 'NotifyService', 'uiGridConstants', '$state', 'GridColumnService',
-  'GridStateService', 'util', 'bhConstants',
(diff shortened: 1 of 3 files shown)


client/src/modules/distribution_center/distribution_center.js
```diff
@@ -83,8 +84,8 @@ function DistributionCenterController(DistributionCenters, Notify, uiGridConstan
         headerCellFilter : 'translate',
         cellClass : 'text-right',
         footerCellClass : 'text-right',
-        cellFilter : 'number: 2',
-        footerCellFilter : 'number:2',
+        cellFilter : 'currency:row.entity.currency_id',
```
mbayopanda, line 87: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4592, opened 2020-06-09
chore(db): remove unused unit.url

jniles:
Removes the `unit.url` from the database because we never use it.


mbayopanda (approved):
LGTM 👍 

bors r+

Issue #1578, opened 2017-05-03
(design) Detect Duplicate Patient Records


jniles:
This is an advanced feature.

Duplicated records seem to be unavoidable.  They are especially bad when they involve duplicated debtors, because the medical/financial history of a patient is then sharded between two entities.  We can help correct this by providing a feature to detect duplicated records and then merge them.

I propose we either a report or a tool in the Patient Registry that will find the duplicate records in the database.  A simply detection algorithm is this:
```sql
SELECT COUNT(uuid), display_name FROM patient
GROUP BY LOWER(display_name) HAVING COUNT(uuid) > 1
ORDER BY COUNT(uuid) DESC;
```

We can then load these into a grid and let the user merge two records at their own choosing.  As an even more advanced feature, we could compute a "similarity score" that compares the two records to merge and gives a warning if they are not similar enough (but still allows a user to do it).

This would be very helpful in maintaining data integrity at institutions over the long term.

jniles:
Before doing this, we should probably implement a "Merge Patient" function.  That way, users can actually perform the critical step of merging duplicate records if necessary.

PR #4594, opened 2020-06-09
feat(patients): detect duplicate patients

jniles:
This PR adds the functionality to detect duplicate patients in the database using the naive algorithm proposed in #1578 (essentially `LOWER(display_name)`).  The idea is that users will be able to use this to detect duplicate patients in their database and merge their histories.

Multiple patients are allowed to be selected at once.  We could change this if we think it is too confusing.  I've also included the ability to change the sensitivity and limits, but these are currently hard-coded.  I'm not convinced it is actually necessary.  We could add this functionality later if that feeling changes.

Here is what it looks like with the Vanga database.
FXDTX2af0V

### DIFF
--- a/client/src/modules/patients/registry/modals/findDuplicatePatients.modal.html
+++ b/client/src/modules/patients/registry/modals/findDuplicatePatients.modal.html
@@ -1,0 +1,38 @@
+<bh-modal-notify error="ModalCtrl.errorValue"></bh-modal-notify>
+
+<div class="modal-header">
+  <ol class="headercrumb">
+    <li class="static" translate>TREE.HOSPITAL</li>
+    <li class="static" translate>PATIENT_REGISTRY.TITLE</li>
+    <li class="title" translate>PATIENT_REG.FIND_DUPLICATE_PATIENTS</li>
+  </ol>
+</div>
+
+<div class="modal-body">
+  <p class="text-success" ng-if="ModalCtrl.hasNoDuplicates" translate>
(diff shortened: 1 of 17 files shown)


client/src/modules/patients/registry/modals/findDuplicatePatients.modal.html
```diff
@@ -0,0 +1,38 @@
+<bh-modal-notify error="ModalCtrl.errorValue"></bh-modal-notify>
```
mbayopanda, line 1: :-)

client/src/modules/patients/registry/modals/findDuplicatePatients.modal.html
```diff
@@ -0,0 +1,38 @@
...
+<div class="modal-header">
+  <ol class="headercrumb">
+    <li class="static" translate>TREE.HOSPITAL</li>
+    <li class="static" translate>PATIENT_REGISTRY.TITLE</li>
+    <li class="title" translate>PATIENT_REG.FIND_DUPLICATE_PATIENTS</li>
+  </ol>
+</div>
+
+<div class="modal-body">
+  <p class="text-danger" ng-if="ModalCtrl.hasNoDuplicates" translate>
```
mbayopanda: No duplicated records can have a successful message i think

mbayopanda (commented):
I just made few comments but the feature works well

jniles:
@mbayopanda changes landed.

mbayopanda:
LGTM 👍 

bors r+

PR #4601, opened 2020-06-10
fix(tests): Promise.all() -> FU.series()

jniles:
Transitions many of our `Promise.all()` calls to `FormUtils.series()` calls to prevent race conditions in the end to end tests.

Cherry-picked from https://github.com/IMA-WorldHealth/bhima/pull/4542.


test/end-to-end/shared/FormUtils.js
```diff
@@ -307,4 +307,12 @@ module.exports = {
 
   // bindings for modal overlay forms
   modal,
+
+  // chains an array of promises and runs them in series.
+  series : async (array, callback) => {
+    return array.reduce(
+      (promise, element, index, array) => promise.then(() => callback(element, index, array)),
+      Promise.resolve(),
```
mbayopanda, line 315: Why do you prefer this model of chaining Promise then just use `Promise.all` ? Does `Promise.all` doesn't order content of the given array each time ?

test/end-to-end/employees/registration.page.js
```diff
@@ -114,10 +114,14 @@ class RegistrationPage {
...
-  setRubricPayroll(rubrics) {
+  async setRubricPayroll(rubrics) {
     const keys = Object.keys(rubrics);
-    const promises = keys.map(key => element(by.id(key)).sendKeys(rubrics[key]));
-    return Promise.all(promises);
+
+    // eslint-disable-next-line
+    for (const key of keys) {
+      // eslint-disable-next-line
+      await element(by.id(key)).sendKeys(rubrics[key]);
```
mbayopanda, line 123: it's really worth removing `Promise.all` for `FU.series` ?

mbayopanda (commented):
@jniles i made just some comments

test/end-to-end/shared/FormUtils.js
jniles: Mainly because Promise.all() doesn't make sense in these cases.  `Promise.all()` runs things in parallel.  But if a user is putting in a list of four accounts to the `bhAccountMultipleSelect` (for example), they do not get four keyboards and start typing all four accounts into the `<input>` at the same time.  Instead, they do the first one, select it, then the next one, then the next one... etc.  That's what `FU.series()` does.

I noticed that `Promise.all()` works locally, but when we went to BrowserStack it crashed each time because it couldn't load the information fast enough.  With FU.series() it gets past that point.

test/end-to-end/employees/registration.page.js
jniles: There are a few places `Promise.all()` makes sense - when you are reading information, but not doing any actions.  For example, the place it might makes sense is :
```js
const allTextInParagraphs = Promise.all($$('p').map(element => element.getText()));
```
Here, you are only reading the DOM and it doesn't matter what order things happen in.  However, if you are doing a `.click()` or a `.sendKeys()` or any of that, it probably matters what order you do things in.  It also makes more sense since a user can only do one thing at a time.

mbayopanda:
👍 bors r+

PR #4604, opened 2020-06-11
fix(distribution key): add percentage filter

### LINKED ISSUE #4435: Taux/Rate column in Distribution Keys should have % symbol

jniles:
Adds the percentage filter to the distribution key module.

Closes #4435.
Closes #4561.

Here is what it looks like:
image


mbayopanda (approved):
bors r+

jniles:
Bors r+

PR #4605, opened 2020-06-11
chore(filters): run eslint --fix on filters

jniles:
This commit runs `eslint --fix` on the custom filters.


jniles:
It seems like we may not have unit tests for a lot of our filters.  I'll try to add those before this should be merged.  That way we can ensure that functionality didn't change.

PR #4607, opened 2020-06-11
fix(bhCurrencySelect): pass back full currency object

### LINKED ISSUE #4589: Currency Select doesn't work in Posting Journal Search
Looks like the currency select no longer works in the Posting Journal Search modal.  See below:
Xx1fvG8c25

jniles:
Passes back the full currency object from the bhCurrencySelect.  It also fixes the Posting Journal's search by currency.

Closes #4589.


jniles:

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4608, opened 2020-06-11
fix(db): ensure HR modules are under HR

### LINKED ISSUE #4597: Move Professions and Grade Management into Human Resources.
Under "Administration" we currently have Grade Management and Profession Management.  See below:
image

These are tied to Human Resources.  We should go ahead and move them into the Human Resources folder.

jniles:
Closes #4597.


mbayopanda (approved):
👍 

bors r+

PR #4609, opened 2020-06-11
feat(patients): refresh debtor group history

### LINKED ISSUE #4481: Changing debtor groups should update debtor group history immediately
When you change a patient's debtor group, it would be nice if the history updated immediately.  You have to refresh the page currently.  Here is an example of what currently happens.

lReo5Evw08

jniles:
Adds a feature to the patient edit page to instantly refresh the debtor group history after modification.  Only 15% hacky.  Also adds a loading indicator.

Closes #4481.

Here is what it looks like:
kZr8KRtYjW


mbayopanda (approved):
The feature works fine

bors r+

PR #4610, opened 2020-06-11
Add stock exit report grouped by services

### LINKED ISSUE #4495: Optionally group Stock Exit Report by service instead of product
I was on a Skype call with the MD of Vanga yesterday and he requested that we adjust the report of "Sorties de Stocks" by grouping by service instead of by product.  To his mind, he rarely wants to see the amount consumed of a product, but would rather see the amount consumed by service.  For reference, the report looks like this:

image

I'm not sure if it makes more sense to either:
  1. Make another report for this.
  2. Change the entire report to group by service first for this section.
  3. Include an optional flag that lets the user switch the presentation between the two.

mbayopanda:
This PR is releated to the issue #4495

image

image

Close #4495


jniles (approved):
Tested and confirmed it works.  Thanks!

bors r+

PR #4620, opened 2020-06-16
Feature: add bh-checkbox-tree component for nested multi-selection

jniles:
Adds a new component `bhCheckboxTree` with the following signature:
```html
<bh-checkbox-tree
  data="$ctrl.data"
  on-change="$ctrl.onChangeCallback(data)">
</bh-checkbox-tree>
```
The component looks like this:
image

And here is an example of it in action:
YlgQHmM58P

**Motivation**
Originally, this component was supposed to be a generic, tested component for dealing with permissions to units assigned to roles.  It supports nesting up to three levels deep for that reason.  We had a number of different implementations that were sometimes buggy (`ngClick` behaves differently from `ngChange` on checkboxes) and I hoped to reduce those errors by just using a single tested component.

After implementing it, I discovered that with only a few changes, we could use this on lists, for example, while assigning cashboxes/roles to users.  Those changes were made with the `is-flat-tree` option.   Now it provides a better interface (in my view) than the `uiSelectMultiple`s we were previously using.

**Additional Parameters**
The underlying component uses the [`TreeService`](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/services/TreeService.js) which has been enhanced with `idKey` property to signify which field to use as an identifier.  The additional options you can pass to the `bhCheckboxTree` component are:

1. `is-flat-tree="true"` - tells the `bhCheckboxTree` to consider all elements as direct descendants of the root node.  This is used for flat lists rather than nested lists.
2. `id-key="uuid"` - tells the component to use the `uuid` field of the objects as the id key.  This is what is returned in the `onChange` callback.
3. `parent-key="parent"` - tells the component which field has the identifier of the parent node.
4. `label-key="label"` - tells the component which field to use as the label on the tree.
4. `checked-ids="$ctrl.array"` - an array of ids that should be pre-selected.

### DIFF
--- a/client/src/js/components/bhCheckboxTree/bhCheckboxTree.html
+++ b/client/src/js/components/bhCheckboxTree/bhCheckboxTree.html
@@ -1,0 +1,39 @@
+<div data-bh-checkbox-tree>
+  <div class="checkbox">
+    <label>
+      <input type="checkbox" ng-model="$ctrl.root._checked" ng-change="$ctrl.setNodeValue($ctrl.root.id, $ctrl.root._checked)" />
+      <strong class="text-capitalize" data-root-node translate>FORM.LABELS.CHECK_ALL</strong>
+    </label>
+  </div>
+
+  <div ng-repeat="node in $ctrl.root.children track by node.id">
+    <div class="checkbox">
+      <label>
+        <input type="checkbox" ng-model="node._checked" ng-change="$ctrl.setNodeValue(node.id, node._checked)" />
+        <span ng-class="{ 'text-bold' : node.children.length }" data-label="{{node._label}}" translate>{{node._label}}</span>
+      </label>
+    </div>
+
+    <ul style="margin-left: calc({{node.depth}} * 15px)" class="list-unstyled">
+      <li ng-repeat="child in node.children track by child.id">
+        <div class="checkbox">
+          <label>
+            <input type="checkbox" ng-model="child._checked" ng-change="$ctrl.setNodeValue(child.id, child._checked)" />
+            <span data-label="{{child._label}}" translate>{{child._label}}</span>
+          </label>
+        </div>
+
+        <ul style="margin-left: calc({{child.depth}} * 15px)" class="list-unstyled">
+          <li ng-repeat="grandchild in child.children track by grandchild.id">
+            <div class="checkbox">
+              <label>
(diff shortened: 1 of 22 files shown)


client/src/js/components/bhCheckboxTree/bhCheckboxTree.html
```diff
@@ -0,0 +1,39 @@
...
+    <label data-root-node>
+      <input type="checkbox" ng-model="$ctrl.root._checked" ng-change="$ctrl.setNodeValue($ctrl.root, $ctrl.root._checked)" />
+      <strong class="text-capitalize" translate>FORM.LABELS.CHECK_ALL</strong>
+    </label>
+  </div>
+
+  <div ng-repeat="node in $ctrl.root.children track by $ctrl.tree.id(node)">
+    <div class="checkbox">
+      <label data-label="{{node._label}}">
+        <input type="checkbox" ng-model="node._checked" ng-change="$ctrl.setNodeValue(node, node._checked)" />
```
mbayopanda, line 12: Is there a reason for using `_` underscore as prefix of the variable name (or property name) ?

client/src/js/components/bhCheckboxTree/bhCheckboxTree.js
```diff
@@ -0,0 +1,158 @@
...
+    // compute node depths
+    $ctrl.tree.walk(Tree.common.computeNodeDepth);
+
+    // ensure that checked ids are an array
+    if (!Array.isArray($ctrl.checkedIds) || data.length === 0) {
+      return;
+    }
+
+    // ensure that the mask is a cloned array
+    const mask = [...$ctrl.checkedIds];
```
mbayopanda, line 66: Is this a nowadays JSON.parse(JSON.stringify(array)) ? for having a new array which doesn't have the same reference to the old one

mbayopanda (approved):
LGTM 👍 

bors r+

client/src/js/components/bhCheckboxTree/bhCheckboxTree.js
jniles: Correct.  Although it is more like something like:
```js
var a = [1, 2, 3];
var b = a.slice();
console.log(a !== b);
console.log(a) // [1, 2, 3]
console.log(b) // [1, 2, 3]
```

PR #4626, opened 2020-06-19
Start bhima from mocha

### LINKED ISSUE #4617: Start the expressjs server from Mocha in integration tests
We are currently not starting and stopping our server from `mocha` in our integration tests.  I'm afraid out method of starting and stopping the server isn't very robust.  It would be nice to start our server from `mocha` as is [proposed here](https://medium.com/@asciidev/testing-a-node-express-application-with-mocha-chai-9592d41c0083).

Other potentially useful links:
  1. [Another example](https://stackoverflow.com/a/38223989) of starting from mocha.

mbayopanda:
This PR is related to the issue #4617 

Close #4617


kwilu:
Hello @mbayopanda, it seems like you have referenced #4617 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #4617.`.

```
Author: kwilu

    Fixes #51.
```

sh/integration-tests.sh
```diff
@@ -13,24 +13,12 @@ set +a
 
 echo "[test]"
 
-# set build timeout
-TIMEOUT=${BUILD_TIMEOUT:-8}
-
 echo "[test] Spawning server process..."
-# build and start the server
```
jniles, line 20: 👍

sh/integration-tests.sh
```diff
@@ -13,24 +13,12 @@ set +a
...
-echo "[test] Spawned node process."
-
-# make sure we have enough time for the server to start
-echo "[test] Sleeping for $TIMEOUT seconds."
-sleep "$TIMEOUT"
 
 echo "[test] running tests using mocha"
-
 # run the tests
-../node_modules/.bin/mocha --recursive --bail ../test/integration/
```
jniles, line 34: Why do we need to put `NODE_ENV` here?  I think it should be fine without it.

.env.development
```diff
@@ -11,7 +11,7 @@ DB_NAME='bhima_test'
 # session variables
 SESS_SECRET='XopEn BlowFISH'
 
-DEBUG=app,errors,ReportManager
+DEBUG=app
```
jniles: We should probably keep `errors` around at minumum.

server/app.js
```diff
@@ -67,7 +67,7 @@ function configureServer() {
 
   // create the server
   http.createServer(app)
-    .listen(process.env.PORT, () => {
+    .listen(port, () => {
```
jniles, line 70: Nice. :+1:

test/integration/setup.js
```diff
@@ -10,12 +10,14 @@
...
 const chai = require('chai');
 const chaiHttp = require('chai-http');
 const chaiDatetime = require('chai-datetime');
 
+// server
+const server = require('../../bin/server/app');
+
 // base URL for all tests
 const port = process.env.PORT || 8080;
 const baseUrl = `http://localhost:${port}`;
```
jniles: Let's remove all references to baseUrl.  Nothing is using it that I can tell.

test/integration/setup.js
```diff
@@ -32,17 +34,19 @@ before(() => {
...
 
   // trigger login
-  return (() => agent.post('/auth/login').send(user))();
+  return agent.post('/auth/login').send(user);
 });
 
 // runs after all tests are completed
-after(() => {
+after(done => {
+  delete require.cache[require.resolve('../../bin/server/app')];
```
jniles: I'm not sure I understand why you need to do this.  It also looks really hacky.  Can we remove it?

sh/integration-tests.sh
jniles: Also, something is keeping our server alive and it isn't exiting.  I've ruled out `lib/db` and `@imaworldhealth/coral`, but still can't figure it out.  Let's not deal with it in this PR.  Instead, let's add a TODO and pass the `--exit` flag.  

Then create an issue and we can work on it after this PR gets merged.

mbayopanda: The reason of `NODE_ENV` is without that mocha cannot read the environment file, so the server is still running in production mode

test/integration/setup.js
mbayopanda: You're totally right, that's the reason why i talked about that

sh/integration-tests.sh
jniles: It should always read the `.env.development` file.  The code is [here](https://github.com/IMA-WorldHealth/bhima/blob/master/sh/test.sh#L8).  At least it does on my machine and on Travis.  Does it not work for you?

jniles:
LGTM.  bors r+

PR #4627, opened 2020-06-19
chore(deps): remove mkdirp

jniles:
Turns out that recursive mkdir is supported in all nodejs versions we support.  Removes the `mkdirp` module.


lomamech (commented):
No comment :shirt:

PR #4628, opened 2020-06-19
fix(feeCenter): check correct affectedRows prop

### LINKED ISSUE #4521: delete unknown fee_center returns 204
noticed when working on #4511

this if statement seems not to do anything https://github.com/IMA-WorldHealth/bhima/blob/fcd30509b0d7e04f43a87e6be45bba5c45fdf0f4/server/controllers/finance/feeCenter.js#L222 
because rows is an array
```
{
  "rows": [
    {
      "fieldCount": 0,
      "affectedRows": 0,
      "insertId": 0,
      "serverStatus": 3,
      "warningCount": 0,
      "message": "",
      "protocol41": true,
      "changedRows": 0
    },
    {
      "fieldCount": 0,
      "affectedRows": 0,
      "insertId": 0,
      "serverStatus": 3,
      "warningCount": 0,
      "message": "",
      "protocol41": true,
      "changedRows": 0
    },
    {parts of the 
      "fieldCount": 0,
      "affectedRows": 0,
      "insertId": 0,
      "serverStatus": 3,
      "warningCount": 0,
      "message": "",
      "protocol41": true,
      "changedRows": 0
    }
  ]
}
```

I guess the if should test all 3 `affectedRows`.

I can fix this, but don't want to mangle it into #4511 so this is a note to not forget

jniles:
Ensures that the server returns 404 if the feeCenter doesn't exist on DELETE.

Closes #4521

### DIFF
--- a/server/controllers/finance/feeCenter.js
+++ b/server/controllers/finance/feeCenter.js
@@ -82,7 +82,6 @@ function detail(req, res, next) {
     .catch(next);
 }
 
-
 // POST /fee_center
 async function create(req, res, next) {
   try {
@@ -136,7 +135,6 @@ async function create(req, res, next) {
     next(err);
   }
 }
-
 
 // PUT /fee_center /:id
 async function update(req, res, next) {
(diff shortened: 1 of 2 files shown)


lomamech (approved):
No Comment

PR #4637, opened 2020-06-23
chore(reports): use singe formatFilters()

### LINKED ISSUE #4613: Combine reports/common.js formatFilters() and reports/shared.js formatFilters()
We have two functions that are similar, but has diverged.  They were meant to be one utility function.  Here they are:
  1. https://github.com/IMA-WorldHealth/bhima/blob/master/server/controllers/finance/reports/shared.js#L183
  2. https://github.com/IMA-WorldHealth/bhima/blob/master/server/controllers/stock/reports/common.js#L52

The most maintained one is in `reports/shared.js` as it has seen the most work recently.  We should remove the `common.js` one and replace all calls to calls of `shared.js`.

jniles:
Combines all instances of formatFilters() into a single instance of formatFilters().

Closes #4613.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,5 +47,7 @@ sh/build-database-test.sh
 docs/_site/
 docs/.sass-cache/
 
+client/upload/
+
 yarn-error.log
 *.log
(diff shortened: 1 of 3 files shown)


server/controllers/stock/reports/common.js
```diff
@@ -40,49 +38,11 @@ const db = require('../../../lib/db');
...
-      } else {
-        column.value = value;
-      }
-      return true;
-    }
-    return false;
-  });
-}
-
+const { formatFilters } = require('../../finance/reports/shared');
```
mbayopanda, line 45: 👍

mbayopanda (approved):
LGTM

.gitignore
```diff
@@ -47,5 +47,7 @@ sh/build-database-test.sh
 docs/_site/
 docs/.sass-cache/
 
+client/upload/
```
mbayopanda, line 50: :-)

mbayopanda (approved):
LGTM

PR #4638, opened 2020-06-23
fix(reports): fix filename on report download

jniles:
The HTTP spec requires that the filename in the Content-Disposition header is put in quotes. See:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

### DIFF
--- a/server/lib/ReportManager.js
+++ b/server/lib/ReportManager.js
@@ -185,7 +185,7 @@ class ReportManager {
     const renderHeaders = renderer.headers;
 
     if (fileName) {
-      renderHeaders['Content-Disposition'] = `filename=${fileName}`;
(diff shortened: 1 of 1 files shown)


server/lib/ReportManager.js
```diff
@@ -185,7 +185,7 @@ class ReportManager {
     const renderHeaders = renderer.headers;
 
     if (fileName) {
-      renderHeaders['Content-Disposition'] = `filename=${fileName}`;
+      renderHeaders['Content-Disposition'] = `filename="${fileName}"`;
```
mbayopanda, line 188: 👍

mbayopanda (approved):
bors r+

PR #4652, opened 2020-06-24
refactor(Payroll Process)

### LINKED ISSUE #4460: Look up employees on UUID, not reference
There are a few places where we use the employee `reference` column instead of the `uuid` column to look up employees in the payroll modules.  This is somewhat confusing because the primary key is the `uuid` column and in all other tables, `reference` is not guaranteed to be unique.

We should migrate these places to using the `uuid` column instead for finding employees.

lomamech:
- Set the look up employees on Uuid, not reference
- Fix the link between payroll registry and employee registry

closes #4460

### DIFF
--- a/client/src/modules/multiple_payroll/multiple_payroll.js
+++ b/client/src/modules/multiple_payroll/multiple_payroll.js
@@ -5,7 +5,7 @@ angular.module('bhima.controllers')
 MultiplePayrollController.$inject = [
   'MultiplePayrollService', 'NotifyService',
   'GridSortingService', 'GridColumnService', 'GridStateService', '$state',
-  'ModalService', 'ReceiptModal', 'uiGridConstants', 'SessionService',
+  'ReceiptModal', 'uiGridConstants', 'SessionService',
 ];
 
 /**
@@ -18,7 +18,7 @@ MultiplePayrollController.$inject = [
  */
 function MultiplePayrollController(
   MultiplePayroll, Notify, Sorting, Columns, GridState, $state,
-  Modals, Receipts, uiGridConstants, Session,
+  Receipts, uiGridConstants, Session,
 ) {
   const vm = this;
   const cacheKey = 'multiple-payroll-grid';
@@ -190,8 +190,8 @@ function MultiplePayrollController(
     vm.getSelectedEmployees = employees;
 
     if (employees.length) {
-      // get All Employees Reference
-      const employeesRef = employees.map(emp => emp.reference);
+      // get All Employees Uuid
+      const employeesUuid = employees.map(emp => emp.employee_uuid);
 
       // returns true If one employee who is not configured is selected
       const isNotConfigured = employee => parseInt(employee.status_id, 10) !== 2;
@@ -203,7 +203,7 @@ function MultiplePayrollController(
         vm.activePosting = false;
 
         const idPeriod = vm.latestViewFilters.defaultFilters[0]._value;
-        MultiplePayroll.paiementCommitment(idPeriod, employeesRef)
+        MultiplePayroll.paiementCommitment(idPeriod, employeesUuid)
           .then(() => {
             Notify.success('FORM.INFO.CONFIGURED_SUCCESSFULLY');
             vm.activePosting = true;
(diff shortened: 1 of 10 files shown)


server/controllers/payroll/reports/payslipGenerator.js
```diff
@@ -30,11 +30,20 @@ const DEFAULT_OPTS = {
 function build(req, res, next) {
   const options = _.clone(req.query);
 
+  if (Array.isArray(options.employees)) {
```
jniles: One way you might simplify this code up is to make `options.employees` always an array.  You can do this with [`Array.prototype.concat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Concat).  Like this:
```js
options.employees = [].concat(options.employees); // now options.employees is an array
```
Then you can simply use a `map()` to process the uuids to binary form.
```js
options.employees = options.employees.map(uid => db.bid(uid));
```

server/controllers/payroll/multiplePayroll/find.js
```diff
@@ -13,28 +13,29 @@ function find(options) {
...
-      payroll.status_id, payroll.status, (payroll.net_salary - payroll.amount_paid) AS balance, payroll.hrreference
+    SELECT BUID(payroll.employee_uuid) AS employee_uuid, payroll.reference, payroll.code, payroll.date_embauche,
+      payroll.nb_enfant, payroll.individual_salary, payroll.account_id, payroll.creditor_uuid, payroll.display_name,
+      payroll.sex, payroll.paiement_uuid, payroll.payroll_configuration_id, payroll.currency_id,
+      payroll.paiement_date, payroll.base_taxable, payroll.basic_salary, payroll.gross_salary, payroll.grade_salary,
+      payroll.text, payroll.net_salary, payroll.working_day, payroll.total_day, payroll.daily_salary,
+      payroll.amount_paid, payroll.status_id, payroll.status, (payroll.net_salary - payroll.amount_paid) AS balance,
+      payroll.hrreference
     FROM(
-      SELECT BUID(employee.uuid) AS employee_uuid, employee.reference, em.text AS hrreference, employee.code,
```
jniles, line 34: This is a good change. 👍

client/src/modules/multiple_payroll/templates/action.tmpl.html
```diff
@@ -20,7 +20,7 @@
     <li class="divider"></li>
 
     <li>
-      <a data-method="employee" ui-sref="employeeRegistry({ filters : [{ key : 'hrreference', value : row.entity.hrreference, cacheable:false }]})" href>
```
jniles, line 23: Hmm... I think I wrote this code and I think I wrote it wrong.  Sorry!  I think it should be:
```html
<a 
  data-method="employee"
  ui-sref="employeeRegistry({ filters : [{ key : 'uuid', value : row.entity.employee_uuid, displayValue: row.entity.hrreference, cacheable:false }]})"
  href> 
```

jniles (commented):
I know you are still fighting through tests, but figured I would point out an error that I think I made 🤷 .

This looks like it is on the right track!  Keep up the good work.

lomamech:

client/src/modules/multiple_payroll/multiple_payroll.js
```diff
@@ -5,7 +5,7 @@ angular.module('bhima.controllers')
 MultiplePayrollController.$inject = [
   'MultiplePayrollService', 'NotifyService',
   'GridSortingService', 'GridColumnService', 'GridStateService', '$state',
-  'ModalService', 'ReceiptModal', 'uiGridConstants', 'SessionService',
```
jniles, line 8: :+1:

server/controllers/payroll/configuration/index.js
```diff
@@ -140,28 +139,6 @@ function payrollReportElements(idPeriod, employees, employeesPaiementUuid) {
     WHERE offday_paiement.paiement_uuid IN (?)
   `;
 
-  const sqlAggregateRubrics = `
```
jniles, line 143: 🕺

jniles (approved):
This looks good to me.  Great work.

bors r+

PR #4653, opened 2020-06-25
Add the recovery capacity report

### LINKED ISSUE #4636: Implement a "Capacite de Recouvrement" Report
An additional financial report that would be helpful to have is a "Capacite de Recouvrement" report that shows the daily cash receipts compared to the invoices.  It would look something like this:
image
_Fig 1: Capacite de Recouvrement" Report_

It will contain the following columns.  The parentheses is the English translation:
  1. Date - the date for that row.
  2. Frequentation (Registrations)  - the number of patients that were registered that day.
  3. Factures (Invoices) - the sum of the value of invoices made that day.
  4. Cout Moyen (Average Cost) - the average cost _per patient_ of all invoices made.  This is column 3 divided by column 2.
  5. Paiements (Cash Payments) - the sum of all money that came into the auxiliary cashboxes that day.
  6. Capacite Recouvrement (Recovery Capacity) - column 5 divided by column 3.  It shows the percentage of invoices paid that day.  Note that it may be over 100%, if old bills are paid over and above the amount billed that day.

We should also include the following color information in the Recovery Capacity column:
  1. Green - RC > 70%
  2. Yellow - 70% >= RC > 60%
  3. Red - RC <= 60%

mbayopanda:
This PR add the report "Recovery Capacity" : 

recovery_report_config

recovery_report_view

Close #4636

### DIFF
--- a/client/src/modules/reports/generate/collectionCapacity/collectionCapacity.config.js
+++ b/client/src/modules/reports/generate/collectionCapacity/collectionCapacity.config.js
@@ -1,0 +1,60 @@
+angular.module('bhima.controllers')
+  .controller('collectionCapacityController', CollectionCapacityController);
+
+CollectionCapacityController.$inject = [
+  '$sce', 'NotifyService', 'BaseReportService',
+  'AppCache', 'reportData', '$state',
+];
+
+function CollectionCapacityController($sce, Notify, SavedReports, AppCache, reportData, $state) {
+  const vm = this;
+  const cache = new AppCache('collectionCapacity');
+  const reportUrl = 'reports/finance/collectionCapacity';
+
+  vm.previewGenerated = false;
+  vm.reportDetails = {};
+
+  vm.clearPreview = function clearPreview() {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.preview = function preview(form) {
+    if (form.$invalid) {
+      Notify.danger('FORM.ERRORS.RECORD_ERROR');
+      return 0;
+    }
+
+    // update cached configuration
+    cache.reportDetails = angular.copy(vm.reportDetails);
+
+    return SavedReports.requestPreview(reportUrl, reportData.id, angular.copy(vm.reportDetails))
+      .then(result => {
+        vm.previewGenerated = true;
+        vm.previewResult = $sce.trustAsHtml(result);
+      })
+      .catch(Notify.handleError);
+  };
+
+  vm.requestSaveAs = function requestSaveAs() {
+    const options = {
+      url : reportUrl,
+      report : reportData,
+      reportOptions : angular.copy(vm.reportDetails),
+    };
+
+    return SavedReports.saveAsModal(options)
+      .then(() => {
+        $state.go('reportsBase.reportsArchive', { key : options.report.report_key });
+      })
+      .catch(Notify.handleError);
+  };
+
+  checkCachedConfiguration();
+
+  function checkCachedConfiguration() {
+    if (cache.reportDetails) {
+      vm.reportDetails = angular.copy(cache.reportDetails);
+    }
+  }
+}
--- a/server/controllers/finance/reports/collection_capacity/index.js
+++ b/server/controllers/finance/reports/collection_capacity/index.js
@@ -90,7 +90,7 @@ async function report(req, res, next) {
       SELECT 
         d.date, IFNULL(pa.registrations, 0) registrations, i.total_invoiced, IFNULL(p.total_paid, 0) total_paid,
         IF(pa.registrations <> 0, (IFNULL(i.total_invoiced, 0) / pa.registrations), 0) avg_cost,
-        IF(i.total_invoiced <> 0, (IFNULL(p.total_paid, 0) / i.total_invoiced), 0) recovery_capacity
+        IF(i.total_invoiced <> 0, ROUND((IFNULL(p.total_paid, 0) / i.total_invoiced), 2), 0) recovery_capacity
       FROM (${dateRange}) d
       LEFT JOIN (${invoices}) i ON i.invoice_date = d.date
       LEFT JOIN (${payments}) p ON p.trans_date = d.date
--- a/server/controllers/finance/reports/collection_capacity/index.js
+++ b/server/controllers/finance/reports/collection_capacity/index.js
@@ -1,0 +1,132 @@
+/**
+ * @overview ./finance/reports/collection_capacity/
+ */
+
+const _ = require('lodash');
+const moment = require('moment');
+
+const ReportManager = require('../../../../lib/ReportManager');
+const db = require('../../../../lib/db');
+
+module.exports.report = report;
+
+// path to the template to render
+const TEMPLATE = './server/controllers/finance/reports/collection_capacity/report.handlebars';
+
+const DEFAULT_OPTIONS = {
+  orientation : 'portrait',
+};
+
+/**
+ * @method report
+ *
+ * @description
+ * The HTTP interface which actually creates the report.
+ */
+async function report(req, res, next) {
+  try {
+    const qs = _.extend(req.query, DEFAULT_OPTIONS);
+    const { dateFrom, dateTo } = req.query;
+    const metadata = _.clone(req.session);
+
+    const rpt = new ReportManager(TEMPLATE, metadata, qs);
+
+    const CASH_PAYMENT_TRANSACTION_TYPE = 2;
+    const CAUTION_TRANSACTION_TYPE = 19;
+    const INVOICE_TRANSACTION_TYPE = 11;
+    const INCOME_ACCOUNT_TYPE = 4;
+
+    /**
+     * credit to :
+     * https://www.shayanderson.com/mysql/generating-a-series-of-dates-in-mysql.htm
+     */
+    const dateRange = `
+      SELECT DATE(cal.date) date
+      FROM (
+            SELECT
+              SUBDATE(DATE(?), INTERVAL DATEDIFF(DATE(?), DATE(?)) DAY) + INTERVAL xc DAY AS date
+            FROM (
+                  SELECT @xi:=@xi+1 as xc from
+                  (SELECT 1 UNION SELECT 2 UNION SELECT 3 UNION SELECT 4) xc1,
+                  (SELECT 1 UNION SELECT 2 UNION SELECT 3 UNION SELECT 4) xc2,
+                  (SELECT 1 UNION SELECT 2 UNION SELECT 3 UNION SELECT 4) xc3,
+                  (SELECT 1 UNION SELECT 2 UNION SELECT 3 UNION SELECT 4) xc4,
+                  (SELECT @xi:=-1) xc0
+            ) xxc1
+      ) cal
+      WHERE cal.date <= DATE(?)
+      ORDER BY cal.date ASC
+    `;
+
+    const patients = `
+      SELECT DATE(p.registration_date) AS registration_date, COUNT(*) registrations
+      FROM patient p 
+      WHERE DATE(p.registration_date) BETWEEN DATE(?) AND DATE(?)
+      GROUP BY DATE(p.registration_date)
+    `;
+
+    const invoices = `
+      SELECT DATE(gl.trans_date) invoice_date, SUM(IFNULL(credit_equiv - debit_equiv, 0)) AS total_invoiced 
+      FROM general_ledger gl
+      JOIN invoice i ON i.uuid = gl.record_uuid
+      JOIN account a ON a.id = gl.account_id
+      WHERE gl.transaction_type_id IN (${INVOICE_TRANSACTION_TYPE})
+        AND (DATE(gl.trans_date) >= DATE(?) AND DATE(gl.trans_date) <= DATE(?))
+        AND i.reversed = 0 AND a.type_id = ${INCOME_ACCOUNT_TYPE}
+      GROUP BY DATE(gl.trans_date)
+    `;
+
+    const payments = `
+      SELECT DATE(gl.trans_date) trans_date, SUM(IFNULL(debit_equiv - credit_equiv, 0)) AS total_paid 
+      FROM general_ledger gl
+      JOIN cash c ON c.uuid = gl.record_uuid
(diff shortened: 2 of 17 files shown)


client/src/modules/reports/generate/collectionCapacity/collectionCapacity.config.js
```diff
@@ -0,0 +1,65 @@
...
+
+  vm.previewGenerated = false;
+  vm.reportDetails = {};
+
+  vm.clearPreview = function clearPreview() {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.onSelectCashbox = function onSelectCashbox(cashbox) {
```
jniles: Would it be possible to just use all cashboxes?  Specifically, all `is_auxiliary` cashboxes.  I think that will make the analysis more complete and more useful.

server/controllers/finance/reports/collection_capacity/index.js
```diff
@@ -0,0 +1,102 @@
...
+const db = require('../../../../lib/db');
+
+module.exports.report = report;
+
+// path to the template to render
+const TEMPLATE = './server/controllers/finance/reports/collection_capacity/report.handlebars';
+
+const DEFAULT_OPTIONS = {
+  orientation : 'portrait',
+  footerRight : '[page] / [toPage]',
```
jniles: Both the `footerRight` and the `footerFontSize` properties are no longer necessary.  These were from wkhtmltopdf and we are using puppeteer now.

server/controllers/finance/reports/collection_capacity/index.js
```diff
@@ -0,0 +1,102 @@
...
+    const rpt = new ReportManager(TEMPLATE, metadata, qs);
+
+    const CASH_PAYMENT_TRANSACTION_TYPE = 2;
+    const CAUTION_TRANSACTION_TYPE = 19;
+
+    const patients = `
+      SELECT DATE(p.registration_date) AS registration_date, COUNT(*) registrations
+      FROM patient p 
+      WHERE (DATE(p.registration_date) BETWEEN ? AND ?)
+        AND project_id = ?
```
jniles: I think you are filtering on project because you want to capture the difference between PAX and Tshikaji.  I don't think it makes sense to do the filtering on project though.   Nothing in BHIMA says that invoices made in one project _must_ be paid back in that same project.  With this filtering we will probably lose information.  Let's just look at the entire institution instead.

server/controllers/finance/reports/collection_capacity/index.js
```diff
@@ -0,0 +1,102 @@
...
+    `;
+
+    const query = `
+      SELECT 
+        i.invoice_date, pa.registrations, i.total_invoiced, IFNULL(p.total_paid, 0) total_paid,
+        IF(pa.registrations <> 0, (IFNULL(i.total_invoiced, 0) / pa.registrations), 0) avg_cost,
+        IF(i.total_invoiced <> 0, (IFNULL(p.total_paid, 0) / i.total_invoiced), 0) recovery_capacity
+      FROM (${invoices}) i
+      LEFT JOIN (${payments}) p ON p.trans_date = i.invoice_date
+      LEFT JOIN (${patients}) pa ON pa.registration_date = i.invoice_date
```
jniles: If I'm reading this right ... this will only show rows for which an invoice exists, correct?  So if we had a day with patients and cash payments, but no invoices, it won't show that day?

server/controllers/finance/reports/collection_capacity/index.js
```diff
@@ -0,0 +1,102 @@
...
+        AND reversed = 0 AND project_id = ?
+      GROUP BY DATE(i.date)
+    `;
+
+    const payments = `
+      SELECT DATE(gl.trans_date) trans_date, SUM(IFNULL(debit_equiv - credit_equiv, 0)) AS total_paid 
+      FROM general_ledger gl
+      JOIN cash c ON c.uuid = gl.record_uuid
+      WHERE gl.transaction_type_id IN (${CASH_PAYMENT_TRANSACTION_TYPE}, ${CAUTION_TRANSACTION_TYPE})
+        AND (DATE(gl.trans_date) >= ? AND DATE(gl.trans_date) <= ?)
```
jniles: Hmm....  This is different from the invoices query.  On the invoices query, you are using the `invoice.date` as the basis for the calculation.  Here, instead of `cash.date` you are using the `general_ledger.trans_date`.  In my view, you should either:
  1. Use `cash.date`  OR
  2.  Use `gl.trans_date` for the invoices query.

Personally, I prefer the idea of using `gl.trans_date` for the invoices since someone could have edited the dates in the posting_journal.

server/controllers/finance/reports/collection_capacity/report.handlebars
```diff
@@ -0,0 +1,54 @@
...
+  {{> header}}
+
+  <!-- body  -->
+  <div class="row">
+    <div class="col-xs-12">
+
+      <!-- page title  -->
+      <h3 class="text-center text-uppercase">
+        <strong>{{translate 'REPORT.COLLECTION_CAPACITY.TITLE'}}</strong> <br>
+        <small>{{account.number}} - {{account.label}}</small><br>
```
jniles: Let's remove the references to accounts - we want to see the balance in all cashboxes that day.

server/controllers/finance/reports/collection_capacity/report.handlebars
```diff
@@ -0,0 +1,54 @@
...
+          <tr>
+            <td>{{date row.invoice_date}}</td>
+            <td style="width: 10em;" class="text-right">{{row.registrations}}</td>
+            <td class="text-right">{{debcred row.total_invoiced metadata.enterprise.currency_id}}</td>
+            <td class="text-right">{{debcred row.avg_cost metadata.enterprise.currency_id}}</td>
+            <td class="text-right">{{debcred row.total_paid metadata.enterprise.currency_id}}</td>
+            <td 
+              style="width: 15em;"
+              {{#gt row.recovery_capacity 0.7}} class="bg-success text-success text-right" {{/gt}}
+              {{#between row.recovery_capacity 0.6 0.7}} class="bg-warning text-warning text-right" {{/between}}
```
jniles, line 57: It looks like all of these have `text-right` in them.  Let's just move it out of this chain.

server/lib/template/helpers/logic.js
```diff
@@ -17,6 +17,14 @@ function gt(a, b, options) {
   return options.inverse(this);
 }
 
+// great than
+function between(a, b, c, options) {
```
jniles, line 21: 👍

server/controllers/finance/reports/realized_profit/index.js
```diff
@@ -42,13 +42,13 @@ async function report(req, res, next) {
 
     const normalPayment = `
       SELECT reference_uuid, SUM(IFNULL(credit_equiv - debit_equiv, 0)) AS paid FROM general_ledger 
-      WHERE transaction_type_id = ${CAUTION_TRANSACTION_TYPE}
+      WHERE transaction_type_id = ${CASH_PAYMENT_TRANSACTION_TYPE}
```
jniles, line 45: Oh la la.  Good catch.

jniles (commented):
@mbayopanda, good work on this.

I've provided some thoughts in comments inline.  I'm not sure it is useful to try and designate a cashbox - we should look at all cash accounts (that are auxiliary cashboxes).  That way, we know for sure the exact amount of money that is being included from our cash paying clients.

mbayopanda:
@jniles changes landed

server/controllers/finance/reports/collection_capacity/index.js
```diff
@@ -0,0 +1,134 @@
...
+    const CAUTION_TRANSACTION_TYPE = 19;
+    const INVOICE_TRANSACTION_TYPE = 11;
+    const INCOME_ACCOUNT_TYPE = 4;
+
+    /**
+     * credit to :
+     * https://www.shayanderson.com/mysql/generating-a-series-of-dates-in-mysql.htm
+     */
+    const dateRange = `
+      SELECT DATE(cal.date) date
```
jniles, line 44: Wow, this is fancy.  Thanks for including a reference to the source!  

I wonder ... would it be possible to create function out of this?  It might be nice to explore that for a future revision.  We could use this kind of thing in other modules.

server/models/migrations/next/migrate.sql
```diff
@@ -16,4 +16,14 @@ INSERT INTO unit VALUES
  * @author: mbayopanda
  * @date: 2020-06-11
  */
-DELETE FROM unit WHERE id = 162;
\ No newline at end of file
+DELETE FROM unit WHERE id = 162;
```
jniles, line 13: I had to make some changes to get this migrate file to run on our production databases.  Could you just add them here?

First, put 
```sql
DELETE FROM role_unit WHERE unit_id = 162;
```
right before this line.  It fails if users have a role assigned to the old stock management module.

server/models/migrations/next/migrate.sql
```diff
@@ -16,4 +16,14 @@ INSERT INTO unit VALUES
...
+
+/**
+ * @author: mbayopanda
+ * @date: 2020-06-23
+ */
+INSERT INTO unit VALUES
+  (271, 'Collection Capacity Report', 'TREE.COLLECTION_CAPACITY_REPORT', 'Collection Capacity Report', 144, '/reports/collectionCapacity');
```
jniles, line 20: `unit` id 271 is already taken by `Stock Inline Movements`.  Could you assign this 272?

server/controllers/finance/reports/collection_capacity/index.js
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * @method report
+ *
+ * @description
+ * The HTTP interface which actually creates the report.
+ */
+async function report(req, res, next) {
+  try {
+    const qs = _.extend(req.query, DEFAULT_OPTIONS);
+    const { dateFrom, dateTo, accountId } = req.query;
```
jniles: The query for `accountId` needs to be removed.  The report is currently broken because we can't select an `accountId`.

server/controllers/finance/reports/collection_capacity/index.js
```diff
@@ -0,0 +1,134 @@
...
+    const formatedDateTo = moment(dateTo).format('YYYY-MM-DD');
+
+    const parameters = [
+      formatedDateTo, formatedDateTo, formatedDateFrom, formatedDateTo,
+      formatedDateFrom, formatedDateTo,
+      formatedDateFrom, formatedDateTo, auxiliaryCashboxesAccountIds,
+      formatedDateFrom, formatedDateTo,
+    ];
+
+    const account = await db.one('SELECT `number`, `label` FROM account WHERE id = ?;', [accountId]);
```
jniles: I am not sure this SQL query is necessary.

jniles (changes requested):
@mbayopanda it looks like this is in a broken state because of the missing `accountId`.  Can you look into that?

Also, the migrate file doesn't build with the production databases.  I've made some comments, to fix those, but there is also one more line that needs to be fixed that I cannot comment on inline.  Could you change

```sql
/**
 * @author: mbayopanda
 * @date: 2020-06-04
 */
 */

INSERT INTO unit VALUES
  (271, 'Stock Inline Movements', 'TREE.STOCK_INLINE_MOVEMENTS', 'The stock movements registry', 160, '/modules/stock/inline-movements', '/stock/inline-movements');
```
to 
```sql
INSERT INTO unit VALUES
  (271, 'Stock Inline Movements', 'TREE.STOCK_INLINE_MOVEMENTS', 'The stock movements registry', 160, '/stock/inline-movements');
```
since we dropped the URL column?

server/models/migrations/next/migrate.sql
mbayopanda: Oh... Since `Stock Inline Movements` has took the place of `Stock Movement` (also the id the unit table), we don't need to add 271 for Stock Inline Movement, we must remove this line from our migrate.sql and keep 271 for Collection capacity report

jniles: Aha, I see.  Yeah, make whatever updates are necessary to make this build 👍

jniles:
Looks like you accidentally included the migrate-imck script.  Can you do a `git commit --amend` to fix that?  Or rebase?

server/controllers/finance/reports/collection_capacity/report.handlebars
```diff
@@ -0,0 +1,53 @@
...
+            <th>{{translate 'REPORT.COLLECTION_CAPACITY.PAYMENTS'}}</th>
+            <th>{{translate 'REPORT.COLLECTION_CAPACITY.RECOVERY_CAPACITY'}}</th>
+          </tr>
+        </thead>
+        <tbody>
+          {{#each rows as | row |}}
+          <tr>
+            <td>{{date row.date}}</td>
+            <td style="width: 10em;" class="text-right">{{row.registrations}}</td>
+            <td class="text-right">{{debcred row.total_invoiced metadata.enterprise.currency_id}}</td>
```
jniles: This should be `../metadata.enterprise_currency_id` in all these locations.

server/controllers/finance/reports/collection_capacity/report.handlebars
```diff
@@ -0,0 +1,53 @@
...
+  <div class="row">
+    <div class="col-xs-12">
+
+      <!-- page title  -->
+      <h3 class="text-center text-uppercase">
+        <strong>{{translate 'REPORT.COLLECTION_CAPACITY.TITLE'}}</strong> <br>
+        <small>{{date dateFrom}} - {{date dateTo}}</small>
+      </h3>
+
+      <table class="table table-striped table-condensed table-report table-bordered">
```
jniles, line 35: Could you put an explanation at the top of this report to explain the colors?  We should mention the ranges for green, yellow, and red.

jniles (commented):
@mbayopanda this is really close!  It looks like the enterprise currency isn't handled properly.  I also think it is important to mention the colorscheme for users who are not experienced.

Otherwise, this report is awesome!  Great work!  I have some ideas for extending it, but will make a new issue since I want this merged first.

jniles (approved):

jniles (approved):
LGTM.  Great work!

bors r+

PR #4655, opened 2020-06-26
chore: remove unused bhMultipleDepotSelect

jniles:
Removes the multiple depot select component which is no longer used. It is superseded by the bh-multiple-depot-search-select component.


mbayopanda (approved):
bors r+

Issue #4656, opened 2020-06-26
Write unit tests for client angularjs filters


jniles:
I paused https://github.com/IMA-WorldHealth/bhima/pull/4605 because I realized we didn't have tests for our filters.  Let's write some tests so we can get that one merged.  We need to have unit tests for:

 - [x] `filters/bytes.js`
 - [ ] `filters/currency.js`
 - [x] `filters/percentage.js`
 - [x] `filters/telephone.js`

Once these are in, #4605 will need to pass those tests to succeed.

PR #4657, opened 2020-06-26
test(filters/bytes): adds unit tests for bytes.js

jniles:
Adds unit tests for the bytes filter.

Partially addresses #4656.


lomamech (approved):
image

All unit tests pass

PR #4658, opened 2020-06-26
test(filters/percentage): add percentage unit tests

jniles:
Adds unit tests for the client's percentage filter.

Partially addresses #4656.

### DIFF
--- a/test/client-unit/filters/percentage.spec.js
+++ b/test/client-unit/filters/percentage.spec.js
@@ -1,0 +1,32 @@
+/* global inject, expect */
+describe('Percentage Filter', () => {
+
+  let percentage;
+
+  beforeEach(module('pascalprecht.translate', 'bhima.filters'));
+
+  beforeEach(inject(($filter) => {
+    percentage = $filter('percentage');
+  }));
+
+  it('returns an empty string for undefined', () => {
+    expect(percentage()).to.equal('');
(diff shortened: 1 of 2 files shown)


test/client-unit/filters/percentage.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  beforeEach(inject(($filter) => {
+    percentage = $filter('percentage');
+  }));
+
+  it('returns an empty string for undefined', () => {
+    expect(percentage()).to.equal('');
+    expect(percentage(undefined)).to.equal('');
+  });
+
+  it('returns the value if the value is not a number', () => {
```
mbayopanda: Is there any case where the percentage filter can receive a string or an object ? Since they are not numeric values why not to just return an empty string

jniles: I can't remember who wrote the percentage filter, but I believe the idea probably was to let a developer know when they have made a mistake by printing out the object.  I'm not certain though.

What do you think?  Should I update the behavior of the percentage filter?

mbayopanda: It will be really good if you can do that

jniles: It has been done.

mbayopanda (approved):
bors r+

PR #4659, opened 2020-06-26
test(filters/telephone): add telephone unit tests

jniles:
Adds unit tests for the telephone filter.  Also observes a silly default value - we default to the USA as a country code.

Partially addresses #4656.


mbayopanda (approved):
bors r+

PR #4661, opened 2020-06-26
refactor(patients): improve stock movement table

jniles:
Improves the "stock movements" section of the `bhPatientFinancialActivity` component by removing borders, adding in the
depot name, and allowing users to click on the stock movement and see the receipt.

This is what it looks like now:
image

And here is what it looks like in action:
V6wehaT7qT


lomamech (approved):
Wonderful

PR #4667, opened 2020-06-28
fix(migrate): read from .env.production

### LINKED ISSUE #4665: Default to reading `.env.production` in setup-migration-script.sh
In our [migration script](https://github.com/IMA-WorldHealth/bhima/blob/master/sh/setup-migration-script.sh) we read `.env.development`.  This is almost never what I want when I am migrating a database - I always want to migrate the database in `.env.production` since that is where I do my testing.

We should probably just make it default to `.env.production` unless someone passes another variable in.

jniles:
Defaults to reading from `.env.production` from the migration setup script.

Closes #4665.

### DIFF
--- a/sh/setup-migration-script.sh
+++ b/sh/setup-migration-script.sh
@@ -14,8 +14,12 @@ set -uo pipefail
 
 echo "[migrate] Migrating BHIMA database"
 
+ENVIRONMENT=${ENVIRONMENT:-"production"}
+
+echo "[migrate] Reading from $ENVIRONMENT."
+
 set -a
-source .env.development
+source .env.$ENVIRONMENT
 set +a
 
 # Set up variables used for naming things


sh/setup-migration-script.sh
```diff
@@ -14,8 +14,12 @@ set -uo pipefail
 
 echo "[migrate] Migrating BHIMA database"
 
+ENVIRONMENT=${ENVIRONMENT:-"production"}
+
+echo "[migrate] Reading from $ENVIRONMENT."
+
 set -a
-source .env.development
+source .env.$ENVIRONMENT
```
mbayopanda, line 22: I would like to know where to set the `ENVIRONMENT` if we want to run in production or in development mode

jniles: On line 17, it is set to "production" by default.  However you can change this the same way you set any environmental variable in bash, by just passing it in.  Like this:
```sh
ENVIRONMENT="development" yarn migrate
# now ENVIRONMENT will be "development"
```
or 
```sh
exports ENVIRONMENT="development"
yarn migrate
```
etc.

jmcameron:
Does this impact the installation directions?

jniles:
No, not the installation directions, but it does have some impact on the [upgrading documentation](https://docs.bhi.ma/en/for-developers/upgrading-bhima.html).  We should probably mention that you can change the default environment.

In the long run, we really should be doing https://github.com/IMA-WorldHealth/bhima/issues/3146 though.

PR #4671, opened 2020-06-29
refactor(Employees Search)

### LINKED ISSUE #4650: Remove the employees/search route
It looks like we have repeated code in the `/employees/search` function.  This is identical to the `/employees` route.  https://github.com/IMA-WorldHealth/bhima/blob/b8a1494d47def61423f6800cff6bf1f52d758f42/server/controllers/payroll/employees/index.js#L40-L47

... versus ...

https://github.com/IMA-WorldHealth/bhima/blob/b8a1494d47def61423f6800cff6bf1f52d758f42/server/controllers/payroll/employees/index.js#L435-L442

Let's just get rid of the `employees/search` route and migrate all code & tests to using the plain `/employees` route.

lomamech:
- Remove the employees/search route

closes #4650

### DIFF
--- a/server/config/routes.js
+++ b/server/config/routes.js
@@ -617,7 +617,6 @@ exports.configure = function configure(app) {
   app.post('/enterprises/:id/logo', upload.middleware('pics', 'logo'), enterprises.uploadLogo);
 
   // employees
-  app.get('/employees/search', employees.search);
   app.get('/employees', employees.list);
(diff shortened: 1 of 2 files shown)


jniles:
bors try

jniles (approved):
Changes look good to me.  We may need to do a few bumps to make the tests properly trigger and pass.

jniles:
Bors r+

PR #4674, opened 2020-06-30
Added install note about problem with MySQL 8.0+ pasword encoding

jmcameron:
Add a trouble-shooting section at the end of the (English) install document.  Add one item I encountered about MySQL 8.0+ not handling password encoding from Node.js and how to fix it.

-----
[View rendered docs/en/for-developers/installing-bhima.md](https://github.com/jmcameron/bhima/blob/install_docs_update/docs/en/for-developers/installing-bhima.md)


jmcameron:
For more details on the MySQL 8 issue, see:  https://stackoverflow.com/questions/50093144/mysql-8-0-client-does-not-support-authentication-protocol-requested-by-server

jniles (approved):

bors r+

Issue #4663, opened 2020-06-27
Enhancements to the Capacite de Recouvrement Report


jniles:
We have a "capacite de recouvrement" report that is landing in https://github.com/IMA-WorldHealth/bhima/pull/4653.  This issue is about further extensions to that report to make it even better.

1. We should include a "totals" footer.  It should contain the same information, just totaled.  It should also tell use how many days the selection was made for.
2. We should include some text describing why numbers could be over 100%.  It should explain what each column does.
3. It would be nice to link each line to the registry to check it.  Importantly, these links should not be visible to the user or visible in the printed version.  I'm not quite sure how we would do this yet, but it would be super helpful.
4. We should optionally include a "Include Posting Journal" and add that information in.  We would then need to change the title to "Provisionary" and include a disclaimer like we do on the Account Statement.

PR #4675, opened 2020-07-01
Add totals and consider unposted transactions

mbayopanda:
This PR : 
- Add totals in the footer of the recovery capacity table
- Add option for not posted transaction

This work is related to #4663

### DIFF
--- a/client/src/modules/reports/generate/collectionCapacity/collectionCapacity.config.js
+++ b/client/src/modules/reports/generate/collectionCapacity/collectionCapacity.config.js
@@ -12,7 +12,7 @@ function CollectionCapacityController($sce, Notify, SavedReports, AppCache, repo
   const reportUrl = 'reports/finance/collectionCapacity';
 
   vm.previewGenerated = false;
-  vm.reportDetails = {};
+  vm.reportDetails = { includeUnpostedValues : 0 };
 
   vm.onChangeUnpostedValues = value => {
     vm.reportDetails.includeUnpostedValues = value;
--- a/client/src/modules/reports/generate/collectionCapacity/collectionCapacity.config.js
+++ b/client/src/modules/reports/generate/collectionCapacity/collectionCapacity.config.js
@@ -14,6 +14,10 @@ function CollectionCapacityController($sce, Notify, SavedReports, AppCache, repo
   vm.previewGenerated = false;
   vm.reportDetails = {};
 
+  vm.onChangeUnpostedValues = value => {
+    vm.reportDetails.includeUnpostedValues = value;
+  };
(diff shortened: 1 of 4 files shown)


server/controllers/finance/reports/collection_capacity/index.js
```diff
@@ -36,6 +36,26 @@ async function report(req, res, next) {
...
+    if (includeUnpostedValues) {
+      generalTable = `
+      (
+        SELECT trans_id, description, trans_date, debit_equiv, credit_equiv, currency_id, debit, credit,
+          account_id, record_uuid, reference_uuid, created_at, transaction_type_id
+        FROM posting_journal WHERE DATE(trans_date) BETWEEN DATE("${formatedDateFrom}") AND DATE("${formatedDateTo}")
+        UNION ALL 
+        SELECT trans_id, description, trans_date, debit_equiv, credit_equiv, currency_id, debit, credit,
+          account_id, record_uuid, reference_uuid, created_at, transaction_type_id
+        FROM general_ledger WHERE DATE(trans_date) BETWEEN DATE("${formatedDateFrom}") AND DATE("${formatedDateTo}")
```
jniles: I don't think it makes any performance difference, but do you really need all these columns?  For example, you do not use description, currency_id, debit, credit, etc.  It would probably improve readability if you omitted them.

server/controllers/finance/reports/collection_capacity/report.handlebars
```diff
@@ -12,6 +12,7 @@
       <!-- page title  -->
       <h3 class="text-center text-uppercase">
         <strong>{{translate 'REPORT.COLLECTION_CAPACITY.TITLE'}}</strong> <br>
+        {{#if includeUnpostedValues}}<strong>({{translate 'REPORT.PROVISIONARY'}})</strong><br>{{/if}}
```
jniles, line 15: Beautiful 👍

client/src/modules/reports/generate/collectionCapacity/collectionCapacity.config.js
```diff
@@ -14,6 +14,10 @@ function CollectionCapacityController($sce, Notify, SavedReports, AppCache, repo
   vm.previewGenerated = false;
   vm.reportDetails = {};
```
jniles: Can you put a default value for `includeUnpostedValues`?  Like:
```js
vm.reportDetails = { includeUnpostedValues : 0 };
```

jniles (approved):
For those wondering what this looks like:

image

This is awesome!  I've checked the totals and they are correct.

bors r+

PR #4678, opened 2020-07-02
Add starting date in the stock sheet report

### LINKED ISSUE #4401: Add a start date limit to fiche de stocks report
Here is the current presentation of the fiche de stocks:
image

It offers only an end date. The problem is that when you have a busy medicine (one with lots of entries and exits) the report gets really difficult to read.  Here is an example: 
KmjE7r65ZH

You can see how, as time goes on, this will get really difficult to manage.  I propose that we add a start date and end date so that we can produce more valuable reports.

mbayopanda:
image

image

Close #4401

### DIFF
--- a/client/src/modules/reports/generate/stock_sheet/stock_sheet.config.js
+++ b/client/src/modules/reports/generate/stock_sheet/stock_sheet.config.js
@@ -12,6 +12,7 @@ function StockSheetConfigController($sce, Notify, SavedReports, AppCache, report
   const reportUrl = 'reports/stock/sheet';
 
   vm.previewGenerated = false;
+  vm.havingDateInterval = 0;
 
   vm.reportDetails = {
     dateTo : new Date(),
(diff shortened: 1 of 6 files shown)


client/src/modules/reports/generate/stock_sheet/stock_sheet.config.js
```diff
@@ -12,6 +12,7 @@ function StockSheetConfigController($sce, Notify, SavedReports, AppCache, report
   const reportUrl = 'reports/stock/sheet';
 
   vm.previewGenerated = false;
+  vm.havingDateInterval = 0;
```
jmcameron: I think a more natural variable name is:  vm.hasDateInterval

jmcameron (commented):
It looks good to me (but I have not actually tried it).  I added one comment about the "having..." variable name.

jniles (commented):
I agree with @jmcameron's comments.  Otherwise, LGTM.

jniles:
LGTM

bors r+

PR #4679, opened 2020-07-03
Feat(Inventory group) add expiration and consumption track flag

### LINKED ISSUE #2683: Turn off consumption, and expiration tracking for groups of inventory items
We've mentioned this in various issues, but we haven't really ever tackled this formally.

I propose that we put flags on the `inventory_group` table that will determine a lot of the display properties of the stock in the depots.
 1. `should_track_consumption`
 2. `should_track_expiration`

These show columns would have the following impact:
 1. During the creation of the a `lot` for an inventory item with `should_track_expiration` set to `0`, the `ui-grid` would not have the `expiration_date` column in it, and the expiration date would be set to something very, very far in the future by default.
 2. During all calculations for expiration dates, the lot would be avoided.
 3. In the "stock lots" registry, the column for "expiration date" and "expiration" would display nothing if `should_track_expiration` is set to `0` for that lot/inventory/inventory_group.
 4. If `should_track_consumption` is `0`, the stock will not show up in any of the consumption reports, and will never be available for automatic re-order.

jeremielodi:
closes #2683


server/controllers/inventory/inventory/groups.js
```diff
@@ -59,7 +59,10 @@ function update(record, identifier) {
  */
 function getGroups(uid) {
   const sql = `
-    SELECT BUID(uuid) AS uuid, code, name, sales_account, cogs_account, stock_account, expires, unique_item
+    SELECT 
+      BUID(uuid) AS uuid, code, name, sales_account,
+      cogs_account, stock_account, expires, unique_item,
+      should_track_consumption, should_track_expiration
```
jmcameron: Here is a suggestion for the variable names:
   should_track_consumption ==> tracking_consumption
   should_track_expiration ==> tracking_expiration
The "should_" versions seem a bit long.  And the "tracking_" ones have the same boolean sense so only the name would be changed.

jmcameron (commented):
Looks good to me (in my limited understanding!)
I do not have enough insight into the system to understand whether this is a good thing to do, but I assume you guys do!

jniles (changes requested):
This is only one third of the work needed to close #2683.

The idea is to make users no longer required to put in expiration dates in the "Stock Entry" module, if we are not requiring the expiration date.  Can you add that logic?  It's what our users asked for.

The second part of this is to remove the values from the CMM calculation if we aren't tracking the consumption.  This should speed up our calculation on the server.  Can you also add that logic?

jeremielodi:
No problem.
Thanks

jeremielodi:
@jniles can you tell me more about the implication of tracking_consumption on the CMM feature?

jeremielodi:
I want to know where I shoud really make changes. Thanks

jniles:
@jeremielodi, sure.

**Expiration**

For expiration, we want to hide/turn off all features that require an expiration date.  For example, we don't want to have a date in the "expiration" column of the Lots in Stock.
image

Similarly, we shouldn't have the option to change the expiration date.
image

We shouldn't show the expiration date for those inventory items that are not tracking the expiration date:
image

.. I may have missed a place or two.  But that is basically the idea.  Anywhere that shows the expiration date should be blank, and anywhere that requires it should not require it.

This also applies to any reports that show the expiration date.

jniles:
A similar thing should happen for consumption, but I think you already have that in your PR.

jeremielodi:
I already solved the stock entry issue

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -7,16 +7,17 @@ StockDefineLotsModalController.$inject = [
...
   const vm = this;
 
   const cache = new AppCache('StockEntryModal');
 
+
   // initialize the form instance
   vm.form = new EntryForm({
     max_quantity : Data.stockLine.quantity,
-    expires : Data.stockLine.expires,
+    expires : Data.stockLine.tracking_expiration,
```
jniles: I think we need to take into account both of these variables.   The logic is this:

  1. If `tracking_expiration` is true, and `expires` is false, then it should be false.
  2. If `tracking_expiration` is false, then it is always false.
  3. If `tracking_expiration` is true and `expires` is true, then it is true.

client/src/modules/stock/lots/modals/edit.modal.html
```diff
@@ -27,6 +27,7 @@
         label="STOCK.EXPIRATION_DATE"
         date-value="$ctrl.model.expiration_date"
         on-change="$ctrl.onDateChange(date)"
+        ng-show="$ctrl.trackingExpiration"
```
jniles: Can you make this `ngIf` instead of `ngShow`?

client/src/modules/stock/lots/modals/edit.modal.js
```diff
@@ -3,25 +3,34 @@ angular.module('bhima.controllers')
...
   vm.cancel = Instance.dismiss;
   vm.submit = submit;
 
+  vm.trackingExpiration = true;
+
   function startup() {
     Lots.read(Data.uuid)
       .then(lot => {
         vm.model = lot;
+        getInventoryDetails(lot.inventory_uuid);
```
jniles: Can you add a `return` to this?  That way the error will be handed and you don't need the additional `catch()` statement.

jniles (commented):
I'll test this later in the day, but figured I'd make some minor comments here first.  It looks pretty good to me so far.

jniles (commented):
It looks like `expires` was already in place.  Let's just rename the column and make sure it is set to the appropriate default value.

image

It looks like when I add an item that does not expire to the stock, we still show an expiration date on the receipt:

image

Again:

image

I propose that we put "---" everywhere that we generally show the expiration date or when the stock should expire if we have turned off the tracking of expiration dates.

server/models/migrations/next/migrate.sql
```diff
@@ -79,3 +79,12 @@ ALTER TABLE `service` DROP COLUMN `id`;
 
 
 ALTER TABLE `service` ADD PRIMARY KEY (`uuid`);
+  ('collectionCapacity', 'REPORT.COLLECTION_CAPACITY.TITLE');
```
jniles, line 111: This is going to break the migration script.  Can you remove it?

server/controllers/stock/reports/stock/entry_purchase_receipt.js
```diff
@@ -74,7 +75,10 @@ async function stockEntryPurchaseReceipt(documentUuid, session, options) {
     voucher_reference     : voucherReference,
   };
 
-  data.rows = rows;
+  data.rows = rows.map(row => {
+    row.expires = row.expires === 1;
```
jniles: I think you are trying to make this into a boolean.  You can do that directly in the SQL:

```sql
IF(eg.expires, TRUE, FALSE) as expires
-- or
IF(eg.expires = 1, TRUE, FALSE) as expires
```

See if that works.  If it does, let's save `for` loop by doing this everywhere you need it.

server/models/schema.sql
```diff
@@ -801,6 +801,8 @@ CREATE TABLE `inventory_group` (
   `donation_account` mediumINT(8) UNSIGNED DEFAULT NULL,
   `expires` TINYINT(1) DEFAULT 1,
   `unique_item`  TINYINT(1) DEFAULT 0,
+  `tracking_consumption`  TINYINT(1) DEFAULT 1,
+  `tracking_expiration`  TINYINT(1) DEFAULT 1,
```
jniles, line 805: Can you combine the `tracking_expiration` and `expires` column?  They are both the same thing.

client/src/modules/stock/lots/modals/edit.modal.js
```diff
@@ -3,21 +3,26 @@ angular.module('bhima.controllers')
...
   vm.cancel = Instance.dismiss;
   vm.submit = submit;
 
+  vm.trackingExpiration = true;
+
   function startup() {
     Lots.read(Data.uuid)
       .then(lot => {
         vm.model = lot;
+        return Inventory.read(lot.inventory_uuid);
```
jniles, line 23: Better 👍

jniles (commented):
Just a few more comments.

client/src/modules/stock/entry/modals/lots.modal.js
jmcameron: Has this been done yet?

client/src/modules/inventory/configuration/groups/modals/actions.modal.js
```diff
@@ -87,7 +87,7 @@ function InventoryGroupsActionsModalController(InventoryGroups, Notify, Instance
 
     if (vm.isCreateState) {
       // by default all inventory (for a group) expires and doesn't have a unique item
-      vm.session.expires = 1;
+      vm.session.tracking_expiration = 1;
```
jniles, line 90: I think you need to add the `tracking_consumption` here too.  It looks like it might default to something in the database that is surprising.  See below.

2GAEeboaMd

jniles (changes requested):
It looks like there is still a bit to do here.

For some reason, setting expiration date is broken now when you receive items in stock.  For some reason it keeps getting set to a value 1000 years in the future.  This doesn't occur on master.  See below:

9WlaqIZeOu

Also, if you make a stock entry with some lots that have expiration dates and some that do not have expiration dates, the receipt doesn't show any expiration dates.  Not sure why.  See below:

WgoWJr5psk

server/models/schema.sql
```diff
@@ -801,6 +801,8 @@ CREATE TABLE `inventory_group` (
   `donation_account` mediumINT(8) UNSIGNED DEFAULT NULL,
   `expires` TINYINT(1) DEFAULT 1,
```
jniles: Can you delete this column?  It is a duplicate of the `tracking_expiration` column.

server/controllers/stock/reports/stock/entry_integration_receipt.js
```diff
@@ -70,7 +71,10 @@ async function stockEntryIntegrationReceipt(documentUuid, session, options) {
     voucher_reference     : voucherReference,
   };
 
-  data.rows = rows;
+  data.rows = rows.map(row => {
+    row.tracking_expiration = row.tracking_expiration === 1;
```
jniles: Did you try the SQL method I proposed earlier with `IF()`?  Did it work?

server/models/migrations/next/migrate.sql
jniles: This is still a problem.  You can't migrate a live database because this SQL script is invalid SQL.  Please fix it.

client/src/modules/stock/entry/modals/StockEntryModalForm.service.js
```diff
@@ -77,7 +77,7 @@ function StockEntryModalForm(uuid) {
     row._error = null;
 
     // if the stock does not expire/have an expiration date, generate a fake one
-    if (!this.opts.expires) {
+    if (!this.opts.tracking_expiration) {
```
jniles, line 86: This change breaks the expiration dates completely.  Can you please double-check this?  I reported it last review, but it wasn't changed.

jniles (changes requested):
I'm still getting problems with this PR related to expiration dates.  For example, I configured quinine, which should have an expiration date, but it doesn't show up on the receipt.  See below:

Vv92UtyOaB

client/src/modules/stock/entry/modals/StockEntryModalForm.service.js
jeremielodi: okay

server/controllers/stock/reports/stock/entry_purchase_receipt.js
jeremielodi: ah okay I didn't see this recommendation. thanks

jeremielodi:
@mbayopanda if you can review this PR to make sure thant the CMM calutation is not broken, thanks

mbayopanda (commented):
The calculation of the CMM is still correct for inventories where we track consumption
The CMM is not calculated for inventories where we do not track consumption
Inventories which don't expire have a default expiration date as programmed 

LGTM 👍

mbayopanda (approved):
LGTM 👍

jniles (approved):
@jeremielodi I think this PR has probably taken long enough.  Let's see if we can merge it into master today and then take care of any remaining issues in other PRs.

Could you update the migration file so that I can bors it?

PR #4680, opened 2020-07-03
Refactor to use IMA-WorldHealth/Tree instead of local copy in lib/Tree

### LINKED ISSUE #4634: Remove lib/tree in favor of @ima-worldhealth/tree
A long time ago, we broke out the `lib/tree` library into an NPM module that could be upgraded/tested separately.  But we never completed the work of migrating to it fully.  We should remove the `lib/tree` library and the unit tests for it, and instead use the `@ima-worldhealth/tree` library.  Whoever does this should run some tests and make sure they behave the same.

jmcameron:
Fixes #4634


kwilu:
Hello @jmcameron, it seems like you have referenced #4634 in your pull request description, but you have not referenced them in your commit message description(s). When you reference an issue in a commit message, it automatically closes the corresponding issue when the commit is merged.

Please run `git commit --amend` in your command line client to amend your commit message description with `Fixes #4634.`.

```
Author: kwilu

    Fixes #51.
```

jmcameron:
I'm going to squash the commits.  Please wait to merge.

package.json
```diff
@@ -51,6 +51,7 @@
   ],
   "dependencies": {
     "@ima-worldhealth/coral": "^2.1.1",
+    "@ima-worldhealth/tree": "^2.2.0",
```
jniles, line 54: 👍

jniles (commented):
This looks good to me.

There should be some more tests for `lib/tree` in the `test/server-unit` folder.  Ideally, we should just remove those tests since they would be consolidated in the [tree repository](https://github.com/IMA-WorldHealth/tree).  Could you remove those?

package.json
jmcameron: Done.

jniles (approved):
This is great!  Thanks!

bors r+

PR #4686, opened 2020-07-07
docs: add notes on AMC in stock exit

jniles:
Added some more formal documentation on stock exits in English and notes on the AMC calculation.  This documentation will be updated when we have all components of the stock management working correctly.

-----
[View rendered docs/en/stock-management/movement.exit.md](https://github.com/jniles/bhima/blob/docs-add-notes-on-AMC-for-stock-movements/docs/en/stock-management/movement.exit.md)
[View rendered docs/en/stock-management/movement.md](https://github.com/jniles/bhima/blob/docs-add-notes-on-AMC-for-stock-movements/docs/en/stock-management/movement.md)

### DIFF
--- a/docs/en/glossary.md
+++ b/docs/en/glossary.md
@@ -1,0 +1,17 @@
+# Glossary
+
+## A
+
+##### Average Monthly Consumption (AMC)
+The average monthly consumption is a measure of the consumption of a drug or medication by patients or hospital services.  Importantly, the AMC does _not_ include the loss of drugs due to expiry, theft, or other losses.  The AMC may be defined for the entire institution or by pharmacy in the institution.
+
+## I
+
+##### Integration
+Integration is a method of increasing or decreasing the quantity in stock of a particular stock lot.  It's used to perform adjustments to the stock in a depot/pharmacy after an inventory is performed and quantities need to be adjusted to match the real quantity in stock.
+
+
+## S
+
+##### Service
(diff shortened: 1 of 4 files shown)


jniles:
@jmcameron can you take a look at this for me?

jmcameron:
Yes, I'll look at this today.

jmcameron:
I like these documents as far as they go.  Obviously a work in progress (which is good).  But I love figures!
A couple of questions:
  - What is a 'Service'
  - What is an 'Integration'

Since we are talking about documentation potentially for green users (like me!) defining each of these terms would be good.  Most of the terms are pretty obvious (eg, patient), so you do not need a full glossary of every term here.  However, including a glossary of all technical terms somewhere might be useful.

jniles:
I like the idea of a glossary.  That's a good idea.

I'll update the documentation with an answer to your questions!

jniles:
@jmcameron changes landed!

jmcameron:
I like the glossary!  I'm wondering what we should do to let folks reading some part of the documentation know that there is a glossary?   Perhaps a footnote?   It would be nice not to have to add something to every single doc page mentioning the glossary (hyperlink).

jniles:
@jmcameron I agree ... though I'm not sure where exactly that would go.  I propose we let this build up more (and flesh out the docs more) and figure out what organizational structure makes sense as the docs organically evolve.  Unless you have a proposal, of course!

PR #4690, opened 2020-07-07
refactor(Merges Locations)

### LINKED ISSUE #4599: Merge Countries In Locations Management
Here is what our countries look like at one of our installation sites:
image

It would be nice to merge these countries similar to how we have merge patients.  That way we could clean up these records.

lomamech:
- Merges locations: Country, Province, Sector, Village
- Quick fixe in merge patient: prevent fatar error when no patient
  selected

closes #4599


lomamech:
@jniles  @mbayopanda can you review this PR

client/src/i18n/fr/form.json
```diff
@@ -225,6 +225,7 @@
             "SAVE_SUCCESS": "Enregistrement avec succès",
             "SEARCH": "Recherche",
             "SET_EXCHANGE_CONVERSION": "Définir le taux de change pour la conversion",
+            "SELECT_LOCATION_CONSIDER": "Sélectionnez l'emplacement à considérer",
```
jmcameron: I am not good at French, but I noticed that the word  "Emplacement" is used here but "Localisations" is used  on line 278.    Is there a reason why the terminology is not consistent?   I have not looked further, but you agree this is an inconsistency, it would be worthwhile reviewing the rest of the modifications for this issue.

client/src/modules/patients/registry/modals/mergePatients.modal.js
```diff
@@ -27,6 +27,11 @@ function MergePatientsModalController(Patients, $state, Notify, Instance) {
   function submit() {
     if (vm.allAreEmployees) { return null; }
 
+    // Quick fix prevent fatar error : 2020-07-07 By: lomamech
```
jmcameron, line 30: Typo: fatar -> fatal

server/controllers/admin/locations.js
```diff
@@ -549,3 +549,73 @@ exports.delete.village = (req, res, next) => {
...
+  const { selected, other, locationStatus } = req.body;
+
+  const transaction = db.transaction();
+
+  if (locationStatus === 'country') {
+    const replaceCountryInProvince = `
+      UPDATE province SET country_uuid = ? WHERE country_uuid = ?;`;
+
+    const removeOtherCountry = `
+      DELETE FROM country WHERE uuid IN (?);`;
```
jmcameron: Are you possibly deleting more than one country?   If we only delete one, perhaps this should use 'uuid = ?' rather than IN?

server/controllers/admin/locations.js
```diff
@@ -549,3 +549,73 @@ exports.delete.village = (req, res, next) => {
...
+    const removeOtherVillage = `
+      DELETE FROM village WHERE uuid IN (?);`;
+
+    transaction
+      .addQuery(replaceVillageInDebtorGroup, [db.bid(selected), db.bid(other)])
+      .addQuery(replaceVillageInEnterprise, [db.bid(selected), db.bid(other)])
+      .addQuery(replaceVillageInPatient1, [db.bid(selected), db.bid(other)])
+      .addQuery(replaceVillageInPatient2, [db.bid(selected), db.bid(other)])
+      .addQuery(removeOtherVillage, [db.bid(other)]);
+  }
```
jmcameron, line 616: I looked through this code for server 'merge()' while reviewing the SQL schema and it all looks okay to me (from a logic standpoint).

server/controllers/admin/locations.js
```diff
@@ -549,3 +549,73 @@ exports.delete.village = (req, res, next) => {
...
+    // enterprise
+    const replaceVillageInEnterprise = `
+      UPDATE enterprise SET location_id = ? WHERE location_id = ?;`;
+
+    // Patient
+    const replaceVillageInPatient1 = `
+      UPDATE patient SET current_location_id = ? WHERE current_location_id = ?;`;
+
+    const replaceVillageInPatient2 = `
+      UPDATE patient SET origin_location_id = ? WHERE origin_location_id = ?;`;
```
jmcameron, line 605: This comment is probably outside the scope of this issue and pull request, but if a patient's current_location_id/origin_location_id is always a village, shouldn't 'village' be part of the name instead of 'location'?  Similarly for the 'location_id' in the other tables.  I seem to recall this discussion from 2015 and recall it was not an easy question to resolve and hard to generalize to different regions of the country (vs big cities, etc).

jmcameron (commented):
Another pretty light review.  I only scanned the client code but tried to look over the server 'merge()' code more closely.  

At some point in the future, I'd like to hear more about how the team balances making the front end client code prevent bad data from going to the server versus having the back end be bullet proof as well.

client/src/i18n/fr/form.json
lomamech: Thanks i will use "localisation"

client/src/modules/patients/registry/modals/mergePatients.modal.js
lomamech: I realized that for the old module, merge of patiens, that when we do not select one and that validated, that abruptly stopped the server

lomamech: My English is not correct because i use Google Translate 
 image

jniles: I always put my French into Google translate to get the accents correct 😆

jniles (commented):
It looks like, if there are duplicate province names, merging the countries fails.

image


Steps to reproduce:
  1. Make two countries, one named "DRC" and the other "RDC"
  2. Make a province in each country called "jniles"
  3. Try to merge "DRC" and "RDC"
 
The error is a unique key erreur since (country_uuid, name) must be unique in the provinces.

jniles:
I'm pretty sure this will exist for all levels, since we have the same constraint on all levels of locations.

@lomamech since we can merge locations, what do you think about us just removing this UNIQUE KEY constraint?

lomamech:
> It looks like, if there are duplicate province names, merging the countries fails.
> 
> image
> 
> Steps to reproduce:
> 
> 1. Make two countries, one named "DRC" and the other "RDC"
> 2. Make a province in each country called "jniles"
> 3. Try to merge "DRC" and "RDC"
> 
> The error is a unique key erreur since (country_uuid, name) must be unique in the provinces.

I just found out the problems, there are the following constraints
````
-- In Province
  UNIQUE KEY `province_1` (`name`, `country_uuid`),
-- In Sector
  UNIQUE KEY `sector_1` (`name`, `province_uuid`),
-- In Village
  UNIQUE KEY `village_1` (`name`, `sector_uuid`),
````
This means that when, for example, two different countries are merged with provinces having the same names, there is the error message linked to the constraint

https://github.com/IMA-WorldHealth/bhima/pull/4690#pullrequestreview-444549532

**Comment by lomamech** [comment]
@jniles  The best way to do this is to mix the lower entities first

**Comment by jniles** [comment]
Alright, that sounds good to me. :+1:

File: server/controllers/admin/locations.js (hunk 5 above)
Comment (jniles) [reply server/controllers/admin/locations.js:605]: Yes, that is correct, and I agree it is confusing.  It might be worth opening an issue over, and I've talked to @lomamech about doing so.

**Comment by lomamech** [comment]
@jniles It's done

File: server/models/migrations/next/migrate.sql [hunk 6]
```diff
@@ -79,3 +79,15 @@ ALTER TABLE `service` DROP COLUMN `id`;
...
+  ('collectionCapacity', 'REPORT.COLLECTION_CAPACITY.TITLE');
+
+/**
+@author: lomamech:
+@date: 2020-07-09
+*/
+
+ALTER TABLE `province` DROP INDEX `province_1`;
+
+ALTER TABLE `sector` DROP INDEX `sector_1`;
```
jniles, line 90: These needs to drop the key, not just the index.  I am still running into errors: 

image

lomamech: With the Vanga database, I was able to mix all the occurrences of DR Congo, and we can also see that there are even several Kwilu Province
image

lomamech: @jniles  Try to check in the structure of the province table, with the error message it seems to me that the index province_1 is not delete
image

server/models/migrations/next/migrate.sql
```diff
@@ -79,3 +79,15 @@ ALTER TABLE `service` DROP COLUMN `id`;
 
 
 ALTER TABLE `service` ADD PRIMARY KEY (`uuid`);
+  ('collectionCapacity', 'REPORT.COLLECTION_CAPACITY.TITLE');
```
jniles: I discovered what the issue was.  This migration script doesn't compile.

Can you remove the `collectionCapacity` part?  It breaks the script.

lomamech: Okay

lomamech:
@jniles it's removed

jniles:
Thanks!  LGTM.

bors r+

PR #4691, opened 2020-07-07
Encapsulate search query logic into service

### LINKED ISSUE #4606: Encapsulate search query logic into a service
We use this kind of code a lot:

```js
    angular.forEach(vm.searchQueries, (value, key) => {
        const displayValue = displayValues[key] || lastDisplayValues[key] || value;
        changes.post({ key, value, displayValue });
      }
    });
```

For example, a quick [ripgrep](https://github.com/BurntSushi/ripgrep) shows that we use it in 21 controllers!

```sh
client/src/modules/account_reference/search.modal.js
82:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/invoices/registry/search.modal.js
97:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/journal/modals/search.modal.js
158:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/multiple_payroll/modals/search.modal.js
115:    angular.forEach(vm.searchQueries, (_value, _key) => {

client/src/modules/display_metadata/modals/search.modal.js
84:    angular.forEach(vm.searchQueries, (_value, _key) => {

client/src/modules/vouchers/modals/search.modal.js
139:    angular.forEach(vm.searchQueries, (_value, _key) => {

client/src/modules/patients/registry/search.modal.js
95:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/employees/registry/search.modal.js
93:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/purchases/modals/search.modal.js
116:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/multiple_payroll_indice/modals/search.modal.js
116:    angular.forEach(vm.searchQueries, (_value, _key) => {

client/src/modules/depots/modals/search.modal.js
57:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/stock/movements/modals/search.modal.js
106:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/stock/lots/modals/search.modal.js
87:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/stock/inventories/modals/search.modal.js
99:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/stock/requisition/modals/search.modal.js
61:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/stock/assign/modals/search.modal.js
87:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/patients/visits/modals/search.modal.js
84:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/dashboards/indicators_files_registry/modals/search.modal.js
97:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/cash/payments/templates/search.modal.js
122:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/inventory/list/modals/search.modal.js
77:    angular.forEach(vm.searchQueries, (value, key) => {

client/src/modules/payroll/staffing_indice/modal/search.modal.js
91:    angular.forEach(vm.searchQueries, (value, key) => {
```

Can we do better?  It would be nice if someone where able to go through and figure out if we could remove some of these lines and replace them with a single call to a service that deals with these values.

mbayopanda:
Close #4606


client/src/js/services/SearchModal.js
```diff
@@ -0,0 +1,33 @@
+angular.module('bhima.services')
+  .service('SearchModalUtilService', SearchModalUtilService);
+
+SearchModalUtilService.$inject = [];
+
+function SearchModalUtilService() {
+  const service = this;
+
+  service.submit = (instance, searchQueries, changes, displayValues, lastDisplayValues) => {
```
jniles: It looks like `service.submit()` is the same as `service.applyChanges()`, but it also closes the modal.

It also looks like you only wrote these two functions to account for one modal in the [DisplayMetadataSearchModalController](https://github.com/IMA-WorldHealth/bhima/pull/4691/files#diff-2afefe47ae8f9b1a1dc2e769bb8a5c2bL83).

I propose you make this one function `getChanges()` that returns `changes.getAll()`.  That way you don't have to pass in `instance` into the function, and the function isn't responsible for closing the modal.

client/src/modules/account_reference/search.modal.js
```diff
@@ -78,27 +79,7 @@ function ReferenceSearchModalController(ModalInstance, filters, Store, util, Pat
       displayValues.number = null;
     }
 
-    // push all searchQuery values into the changes array to be applied
-    angular.forEach(vm.searchQueries, (value, key) => {
-      if (angular.isDefined(value)) {
-
-        // To avoid overwriting a real display value, we first determine if the value changed in the current view.
```
jniles, line 85: I am a little worried there is a corner case we are missing here, but I cannot reproduce it.  I remember that we wrote this code for a reason.  

Still, I don't think it is blocking.

jniles (commented):
This is really good work!

I think it makes more sense to remove the `instance` from this service - there is no reason for it to know that we will close the modal afterwards.  Also, we may want to do something else before closing the modal anyway, as `DisplayMetaDataSearchModalController` does.

Otherwise, LGTM.

jniles (approved):
Excellent.  This is a really nice improvement. 👍 

bors r+

PR #4701, opened 2020-07-09
upgrade(inventory creation) put required fields on top

### LINKED ISSUE #4672: Move required inventory fields to the top of the create/update form
Here is the inventory creation and update form:

image

The fields marked in red are the required items.  Rather than making a user scroll to find them, let's move required fields to the top of the modal so they are easy to locate.

jeremielodi:
close #4672


jniles (approved):
LGTM.

bors r+

PR #4708, opened 2020-07-09
Added 'temp' directory to .gitignore

jmcameron:
Allow developers to create a local 'temp' directory for development files, etc, that git will ignore.


jniles (approved):
Seems like a useful change to me.

bors r+

Issue #4505, opened 2020-05-13
Improvements to "Purchase Order Analysis" Report


jniles:
The "Purchase Order Analysis" report provides a view into the items received in stock for a given purchase order.  However, it can be improved.  Here is how it currently looks:

![image](https://user-images.githubusercontent.com/896472/81811838-2d034280-951d-11ea-8ecc-5b755fb1e2a0.png)
_Fig 1: Example Purchase Order Analysis_

I would make the following suggestions:
  1. The values do not all need to be bold. Only the headers should be bold.
  2. Put the inventory code and label on the same line.  No need to take extra vertical space for that.
  3. Why is everything indented?  We don't need to indent by one cell the items in the table.  Let's move everything over to the left.
  4. The quantity received should be right-aligned.
  5. Something is going on with the quantities - they are all the same.  I think this may be a bug.

jniles:
Looks like this has been closed in #4714

PR #4714, opened 2020-07-13
Improvement to "Purchase Order Analysis" report

mbayopanda:
This report is related to issue #4505 

image

### DIFF
--- a/server/controllers/stock/reports/purchaseOrderAnalysis/report.handlebars
+++ b/server/controllers/stock/reports/purchaseOrderAnalysis/report.handlebars
@@ -55,8 +55,8 @@
           <tbody>
             {{#each this.inventoriesOrdered}}
               <tr>
-                <td><strong> {{ code }} </strong></td>
-                <td><strong> {{ inventory_text }} </strong> </td>
+                <td>{{ code }}</td>
+                <td>{{ inventory_text }}</td>
                 <td class="text-right">{{ quantity_ordered }}</td>
                 <td class="text-right">{{ quantity_inStock }}</td>
                 <td class="text-right"><strong>{{ quantity_difference }}</strong></td>
@@ -102,6 +102,8 @@
                   <td class="text-center" colspan="6">{{translate "TABLE.COLUMNS.EMPTY"}}</td>
                 </tr>
               {{/each}}
+            </tbody>
+            <tfoot>
               <tr>
                 <td colspan="6"></td>
               </tr>
@@ -132,7 +134,7 @@
                   <strong>{{translate status.text}}</strong>
                 </td>
               </tr>
-            </tbody>
+            </tfoot>
           </table>
           <br>
         {{/each}}
--- a/server/controllers/stock/reports/purchaseOrderAnalysis/report.handlebars
+++ b/server/controllers/stock/reports/purchaseOrderAnalysis/report.handlebars
@@ -2,7 +2,7 @@
 <html>
   {{> head title="REPORT.PURCHASE_ORDER_ANALYSIS.TITLE" }}
 
-  <body class="container-fluid">
+  <body class="container">
     {{> header}}
 
     <!-- body -->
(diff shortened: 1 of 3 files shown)


server/controllers/stock/reports/purchaseOrderAnalysis/report.handlebars
```diff
@@ -74,65 +74,61 @@
...
-          <table style="width: 90%; margin-right: auto; margin-left: auto" class="table table-striped table-condensed table-report table-bordered">
-
-              <tr>
-                <td width="15%" class="text-uppercase"> {{translate 'FORM.LABELS.CODE'}}  </td>
-                <td colspan="7" width="85%"><strong> {{ code }} </strong></td>
-              </tr>
-              <tr>
-                <td width="15%" class="text-uppercase"> {{translate 'FORM.LABELS.INVENTORY'}}  </td>
-                <td colspan="7" width="85%"><strong> {{ inventory_text }} </strong></td>
+          <table style="width: 90%; margin: auto;" class="table table-condensed table-report table-bordered">
```
jmcameron: I do not suggest this be fixed now, but in general the html for tables should use &lt;caption&gt;, &lt;thead&gt;, &lt;th&gt;, &lt;tbody&gt;, etc.  See https://alistapart.com/article/web-typography-tables/.    Using these extra elements allows more uniform and appropirate application of CSS rules.    If that is done, rather than embedding 'text-bold' in the <tr> for the header, simple CSS rules for <th> can used and changed later without affecting the HTML code.

jmcameron (commented):
The changes look good.   However, https://github.com/IMA-WorldHealth/bhima/issues/4505 mentions that there may have been a bug in the quantity column.   If that is true, it was not obvious what fixed that in these changes.

jniles:
@mbayopanda this is good.  I like the new presentation.

The problem with quantities persists.  The problem is that we aren't actually filtering on movements that come in from purchase orders.  Here is the SQL that makes this report:

```sql
SELECT BUID(p.uuid) AS purchase_uuid, BUID(iv.uuid) AS inventory_uuid, iv.text AS inventoryText,
  l.label AS lotLabel, p.project_id, p.reference, p.user_id, l.uuid AS lotUuid, l.quantity,
  l.expiration_date, l.entry_date, iv.code,
  d.text AS depotText, BUID(sm.uuid) AS stock_movement_uuid, dm1.text AS stock_movement_reference
  FROM purchase AS p
  JOIN lot AS l ON l.origin_uuid = p.uuid
  JOIN inventory AS iv ON iv.uuid = l.inventory_uuid
  JOIN stock_movement AS sm ON sm.lot_uuid = l.uuid
  JOIN depot AS d ON d.uuid = sm.depot_uuid
  JOIN document_map AS dm1 ON dm1.uuid = sm.document_uuid
  WHERE p.uuid = X'c6a44fc9045843b39ac9cfd2bea8f19f' AND sm.is_exit = 0
  ORDER BY iv.text ASC, l.entry_date ASC 
```

Notice that this will find _all stock movements_ that use this lot.  So, you will notice that here:
image
This is multiple movements that are entering the Pharmacie Usage.  But in reality, the only movement we should be showing is the one entering the pharmacy stock.

We should only be filtering on the type of movement "received from purchase".

server/controllers/stock/reports/purchaseOrderAnalysis/report.handlebars
```diff
@@ -11,51 +11,52 @@
         <!-- page title  -->
         <h3 class="text-center text-bold text-uppercase">{{translate 'REPORT.PURCHASE_ORDER_ANALYSIS.TITLE'}}</h3>
         <br>
-        <table style="width: 90%; margin-right: auto; margin-left: auto" class="table table-condensed table-report table-bordered">
+        <table class="table-condensed table-report">
```
jniles, line 14: You still need the class "table" here, I think.  It doesn't render on the whole page.

server/controllers/stock/reports/purchaseOrderAnalysis/report.handlebars
```diff
@@ -72,69 +73,70 @@
...
-
-              <tr>
-                <td width="15%" class="text-uppercase"> {{translate 'FORM.LABELS.CODE'}}  </td>
-                <td colspan="7" width="85%"><strong> {{ code }} </strong></td>
-              </tr>
-              <tr>
-                <td width="15%" class="text-uppercase"> {{translate 'FORM.LABELS.INVENTORY'}}  </td>
-                <td colspan="7" width="85%"><strong> {{ inventory_text }} </strong></td>
-              </tr>
+          <table class="table-condensed table-report table-bordered">
```
jniles: We need to have some sort of avoiding page break in here.   Otherwise, we lose information:

image

server/controllers/stock/reports/purchaseOrderAnalysis/report.handlebars
mbayopanda: The idea was to have a table width defined by the table content and not a table with 100% width

jniles: I think that makes sense for viewing, but not for printing.  When printing, we want to have a predictable layout for each report.  For example, page numbers should always be in the bottom left, the enterprise name should always be in the top left, etc...

server/controllers/stock/reports/purchaseOrderAnalysis/report.handlebars
mbayopanda: > We need to have some sort of avoiding page break in here. Otherwise, we lose information:
> 
> image

How do you think we can fix this behavior in this PR cause I don't know since we have a variable table number and variable rows for each table

jniles (approved):
LGTM.  Great work!

bors r+

PR #4715, opened 2020-07-13
Filter barcode

### LINKED ISSUE #4273: Filter purchase order registry on barcode
Similar to https://github.com/IMA-WorldHealth/bhima/issues/3396, https://github.com/IMA-WorldHealth/bhima/issues/3005, we should add the ability to rapidly filter the purchase order registry by barcode, just by scanning it.  The design should be identical to the other registries (stock, invoice, patient, cash, etc).  

**Design**
The basic idea would look like this:
  1. In the menu, select menu > Find by Barcode.
  2. The barcode component in a modal comes up to scan a barcode.
  3. The user scans the barcode.
  4. If a record matches the barcode, the modal disappears and the registry is filtered on that particular record.

Note that this issue depends on https://github.com/IMA-WorldHealth/bhima/issues/4272 being implemented first.

jmcameron:
Implemented searching purchase orders by barcode.  Fixes https://github.com/IMA-WorldHealth/bhima/issues/4273.
Thanks to Jonathan Niles for help!  I could squash further but it got hairy with git getting to this point, so I'd rather not.


jniles (approved):
@mbayopanda went on paternity leave, so I'll go ahead and approve this.  Great work!

bors r+

PR #4727, opened 2020-07-15
fix(stock): render inventory report for adjustment

### LINKED ISSUE #4724: Stock Inventory Adjustments appear broken
While testing https://github.com/IMA-WorldHealth/bhima/pull/4679, I came across a bug that does not appear to be related to that PR.  Here is what it looks like:

1kOoi8dciG

It looks like the document_uuid cannot be found for the receipt.  This can either be a bug with the receipt lookup or a bug with the generation of the document_uuid.  I'm guessing it is just a problem with the receipt modal.

Steps to reproduce:
  1. Run the integration tests
  2. Go to the Stock Adjustments module.
  3. Set any of the quantities to "0" 
  4. Type a description
  5. Submit.

jniles:
Changes the stock adjustments module to render the inventory "Articles
in Stock" report for the depot once the adjustment is made.  This way,
the overview of the stock in the depot is presented to the user and they
can print it out and sign off on it.

Closes #4724.

### DIFF
--- a/client/src/js/services/receipts/ReceiptModal.js
+++ b/client/src/js/services/receipts/ReceiptModal.js
@@ -168,7 +168,6 @@ function ReceiptModal(Modal, Receipts, Invoice, Cash, Voucher) {
     return instance.result;
   }
 
-
   /**
    * @method voucher
    *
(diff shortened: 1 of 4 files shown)


lomamech (commented):
After having analyzed this PR, and by checking with the master branch, I noticed that this PR allows to make adjustments with null values

lomamech:
@jniles  Let me know if this was the only problem solved

jniles:
@lomamech it shouldn't make any changes to the logic.  It should just render a different receipt.  Is that not the case?

lomamech:
> @lomamech it shouldn't make any changes to the logic. It should just render a different receipt. Is that not the case?

image

But compared to this image I got the impression that there was a problem when the value to adjust was null

lomamech:
On the master branch I had this error that I cannot find on your branch 
https://i.imgur.com/mD9EfrT.png

lomamech:
On the master branch I had this error that I cannot find on your branch 
image

jniles:
Ah, yes, the idea was to render a different receipt so that the receipt doesn't error :)

PR #4728, opened 2020-07-15
fix(purchase): add bh-integer for purchases

### LINKED ISSUE #4719: Decimal quantities in purchase orders
We allow decimal quantities in purchase orders, but we can't store them in the database.  See below for example:

ESTItkEMjZ

... but here is the definition:
https://github.com/IMA-WorldHealth/bhima/blob/48320993b2d1297da21fe578e11975c2362b0da5/server/models/schema.sql#L1399

This results in a buggy looking receipt.  We should block this on the client-side if we don't accept decimal points.

jniles:
Enforces purchase quantities being integers using `bhInteger`.

tVBZfWqhiv

Closes #4719

### DIFF
--- a/client/src/js/directives/bhInteger.js
+++ b/client/src/js/directives/bhInteger.js
@@ -10,7 +10,7 @@ IntegerValidatorDirective.$inject = [];
  * input with a valid ngModel.
  *
  * @example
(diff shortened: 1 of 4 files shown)


jmcameron (approved):
This looks good to me.

PR #4731, opened 2020-07-16
4687: feat add expired flag to lots in stock

### LINKED ISSUE #4687: Filter Articles in Stock by "Expired"
We don't have an option to filter the articles in stock by the "expired" status.  This status is for articles that have an expiration date in the past _and_ have quantity in still in stock.  Here is the current list of statuses that we can filter the registry on.

image

jniles:
This PR adds a filter on the lots registry to filter on expired stock.  See below for it in action.
LgtxJK8Set

Importantly, the "expiration" flag is set on the server time, not the client time.  Therefore, stock items will be marked as expired if their expiration dates are in the past according to the server.

Closes #4687

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,7 +1,6 @@
 # Server Config
 PORT=8080
 
-
 # database configuration
 DB_HOST='localhost'
 DB_USER='bhima'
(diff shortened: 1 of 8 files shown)


jeremielodi:
I'm reviewing this PR

jeremielodi:
Look good to me.

PR #4736, opened 2020-07-17
Expose database port in env

mbayopanda:
Exposing the database port in the environnement file for using bhima with MAMP or other system which doesn't run MySQL database on 3306


jniles (approved):
LGTM!  

bors r+

PR #4737, opened 2020-07-17
chore: add subfolders for reports

### LINKED ISSUE #4146: Break global "reports" folder into separate "reports" subfolders in each module
We currently have a global "reports" folder that is getting quite large.  We have a total of 41 reports in the `report` table.  Most of these reports are quite specialized - finance, stock, or medical.

To better manage our reports, we should break them up into separate folders for each component.  For example, we would have
```sh
- Finance
   - Reports
- Medical
   - Reports
- Inventory
   - Reports
```

This would also make it easier to manage permissions to the different reports, since we wouldn't have stock reports visible if there were no stock modules enabled.

jniles:
Adds subfolders for reports to each top-level module.

Closes #4146


mbayopanda (commented):
This is a great job 👍 

I have  noticed that : 

- Units are not sorted alphabetically 
image

- We can have multiple `Reports` folder in a module folder : 
image

We can merge them or rename them to avoid confusion

jniles:
@mbayopanda, what database did you upgrade?  Or is this the default bhima_test database?  You aren't supposed to have a `reports` folder anymore in the top-level.  I need to fix that if it exists for some database.

mbayopanda:
> @mbayopanda, what database did you upgrade? Or is this the default bhima_test database? You aren't supposed to have a `reports` folder anymore in the top-level. I need to fix that if it exists for some database.

I have used the default `bhima_test` database

mbayopanda:
It seems like changes are not yet landed

jniles:
@mbayopanda you are right, I haven't made any changes yet - apologies!  Been kind of busy.  I'll ping you when I need another review.  Thank you for your last review :)

jniles:
@mbayopanda, this is ready for another review.  I think I fixed both bugs.

jniles:

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4739, opened 2020-07-17
Highlight expired stock lots rows

### LINKED ISSUE #4730: Expired Stock should show up as red in the Stock Lots management
Here is some stock from our default data that is expired:

image

We should signal strongly that these lots are expired.  In my view, the best way would be to highlight the entire row in red, similar to what we do with reversed records:
image
_Reversed Records in the Invoice Registry_

jmcameron:
This should highlight in red the expired  lot lines of the 
    Stock > Stock Lots
table. 

NOTE: All the stock lots in the default bhima_test database have expired.  To see some stock lots that are not expired, you will need to manually change the expiration dates appropriately.

Closes: #4730

### DIFF
--- a/client/src/modules/stock/lots/registry.js
+++ b/client/src/modules/stock/lots/registry.js
@@ -32,6 +32,7 @@ function StockLotsController(
     showColumnFooter : true,
     fastWatch : true,
     flatEntityAccess : true,
+    rowTemplate : '/modules/stock/lots/templates/row.expired.html',
     showGridFooter : true,
     gridFooterTemplate : LotsRegistry.gridFooterTemplate,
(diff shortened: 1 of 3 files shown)


client/src/modules/stock/lots/registry.js
```diff
@@ -32,6 +32,7 @@ function StockLotsController(
     showColumnFooter : true,
     fastWatch : true,
     flatEntityAccess : true,
+    rowTemplate : '/modules/stock/lots/templates/row.expired.html',
```
mbayopanda, line 35: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4743, opened 2020-07-20
feat(Stock) Sold out compenent

### LINKED ISSUE #4723: Make "Sold Out Inventories" table in Stock Exit into a component.
We should make the "sold out inventories" table in the Stock Exit controller into a component.  Specifically, this:
image
should be a component.  That would take some of the pressure of the StockExitController.

We also need a loading indicator on that table.   If you change depot, the component does update, but it doesn't tell you when it is loading or not.

jeremielodi:
closes #4723


.env.development
```diff
@@ -6,7 +6,7 @@ DB_PORT=3306
 DB_HOST='localhost'
 DB_USER='bhima'
 DB_PASS='HISCongo2013'
-DB_NAME='bhima_test'
+DB_NAME='imck'
```
jniles: This should be reverted to "bhima_test".

client/src/js/components/bhSoldOut/bhSoldOut.html
```diff
@@ -0,0 +1,36 @@
...
+        translate-values="{ left : ($ctrl.soldOutInventories.length - 5)}"
+        translate="{{ 'STOCK.AND_MORE' }}">
+      </span>
+
+    <a ui-sref= "stockInventories({ filters : [
+      { key : 'period', value : 'allTime'},
+      { key : 'includeEmptyLot', value : 1  },
+      { key : 'depot_uuid', value : $ctrl.depot.uuid, displayValue: $ctrl.depot.text, cacheable:false },
+      { key : 'status', value : 'stock_out', displayValue: 'STOCK.STATUS.STOCK_OUT', cacheable:false }
+      ]})"translate>FORM.LABELS.SEE_ALL</a>
```
jniles: It looks like this translate is stuck on the ".  Can you put a space between the two?

client/src/js/components/bhSoldOut/bhSoldOut.js
```diff
@@ -0,0 +1,48 @@
+angular.module('bhima.components')
+  .component('bhSoldOut', {
```
jniles: Can you add the word "stock" to this?  "bhSoldOutStock"?  It would make it easier to look up.

client/src/js/components/bhSoldOut/bhSoldOut.js
```diff
@@ -0,0 +1,48 @@
...
+    stockOut();
+  };
+
+  /**
+   * @function stockOut
+   * get stock out inventories for a depot
+   */
+  function stockOut() {
+    const dateTo = $ctrl.date || new Date();
+    if (!$ctrl.depotUuid) return;
```
jniles: Can you include a loading indicator?  When the depot changes, it would be nice to show a "loading..." indicator.

jniles (commented):
Just a few nit-picks here.  Looks good otherwise!  Thanks!

.env.development
jeremielodi: okay

client/src/js/components/bhSoldOut/bhSoldOut.js
jeremielodi: Good idea

.env.development
jniles: This can't be merged until this is fixed.

jniles:
Bors r+

PR #4756, opened 2020-07-22
feat(lot in stock) add in_risk_of_expiration filter

### LINKED ISSUE #4699: Add a "At Risk of Expiration" status filter to the Articles in Stock Registry
This is similar to https://github.com/IMA-WorldHealth/bhima/issues/4687.

As opposed to the "expired" status, this status would focus on articles in stock that have articles in stock with an expiration date in the future.  The formula for stock at risk of expiration is if the date the quantity in stock will be consumed is past the expiration date.

Basically, in pseudo-code:
```js
const numMonthsOfStockLeft = (lot.quantity  / lot.CMM);  // how many months of stock left
const today = new Date();

// if we have more months of stock than the expiration date, then we'll need to label these are in risk of expiration
const isInRiskOfExpiration = lot.expiration_date < (today.addMonths(numMonthsOfStockLeft))
```

jeremielodi:
closes #4699

### DIFF
--- a/client/src/modules/stock/lots/registry.service.js
+++ b/client/src/modules/stock/lots/registry.service.js
@@ -121,7 +121,16 @@ function LotsRegistryService(uiGridConstants, Session) {
         direction : uiGridConstants.DESC,
         priority : 3,
       },
-    }, {
+    },
+    {
+      field : 'IS_IN_RISK_EXPIRATION',
+      displayName : 'STOCK.IS_IN_OF_RISK_EXPIRATION',
+      headerCellFilter : 'translate',
+      cellTemplate     : 'modules/stock/lots/templates/in_risk_of_expiration.html',
+      cellClass : 'text-right',
+      type : 'number',
+    },
+    {
       field : 'S_RISK_QUANTITY',
       displayName : 'STOCK.RISK_QUANTITY',
       headerCellFilter : 'translate',
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -731,6 +731,13 @@ function processMultipleLots(inventories) {
         lot.S_RISK_QUANTITY = Math.round(lot.S_RISK * lot.avg_consumption);
         lotLifetime += lot.S_LOT_LIFETIME;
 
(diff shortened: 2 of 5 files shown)


server/controllers/stock/core.js
```diff
@@ -731,6 +731,12 @@ function processMultipleLots(inventories) {
         lot.S_RISK_QUANTITY = Math.round(lot.S_RISK * lot.avg_consumption);
         lotLifetime += lot.S_LOT_LIFETIME;
 
+        const numMonthsOfStockLeft = (lot.quantity / lot.CMM); // how many months of stock left
+        const today = new Date();
+        // if we have more months of stock than the expiration date,
+        // then we'll need to label these are in risk of expiration
+        const isInRiskOfExpiration = lot.expiration_date < (today.setMonth(today.getMonth() + numMonthsOfStockLeft));
```
jniles: I know the issue had this pseudo-code in it, but now that I look at it, it might make more sense to make this calculate use days of stock, rather than months.  Stock expired by 2 days isn't able to be sold.

So, let's do something like:
```js
const numMonthsOfStockLeft = (lot.quantity / lot.CMM);
const numDaysOfStockLeft =  numMonthsOfStockLeft * 30.5;
const  isInRiskOfExpiration = lot.expiration_date <  moment(today).add(numDaysOfStockLeft, 'days').toDate();
```

client/src/modules/stock/lots/registry.service.js
```diff
@@ -121,7 +121,16 @@ function LotsRegistryService(uiGridConstants, Session) {
...
         priority : 3,
       },
-    }, {
+    },
+    {
+      field : 'IS_IN_RISK_EXPIRATION',
+      displayName : 'STOCK.IS_IN_OF_RISK_EXPIRATION',
+      headerCellFilter : 'translate',
+      cellTemplate     : 'modules/stock/lots/templates/in_risk_of_expiration.html',
+      cellClass : 'text-right',
```
jniles, line 130: It looks like you already have 'text-center' in the cell template, so this probably isn't necessary.

jniles (approved):
Tested and this looks like it works.  Very nice!

image


bors r+

PR #4757, opened 2020-07-22
feat(Stock) expiration report

### LINKED ISSUE #4700: Create a Stock Expiration Report
We need a report that combines the logic of #4699 and #4687 into a single report.

It would look something like this:

image

The basic idea is that it should show all stock that is either:
  1. Expired 
  2. At Risk of Expiration.

The following options would be available for filtering:
  1. Filter by Depot.
  2. Filter by Inventory Item.

jeremielodi:
closes #4700


jeremielodi:
This PR will be merged after #4756 will land

server/models/bhima.sql
```diff
@@ -231,7 +230,12 @@ INSERT INTO `report` (`report_key`, `title_key`) VALUES
   ('stock_consumption_graph_report', 'REPORT.STOCK_CONSUMPTION_GRAPH_REPORT.TITLE'),
   ('invoicedReceivedStock', 'REPORT.COMPARE_INVOICED_RECEIVED.TITLE'),
   ('invoiceRegistryReport', 'Invoice Registry as report'),
+<<<<<<< bc4c80f0220243ba26ee4f382b51b1ffe724d045
   ('stock_movement_report', 'Stock Movement Report');
+=======
+  ('stock_expiration_report', 'REPORT.STOCK_EXPIRATION_REPORT.TITLE');
+
+>>>>>>> feat(Stock) expiration report
```
mbayopanda: 😄 you committed this

mbayopanda (commented):
It is a good start, I have some issue about the report configuration page which doesn't help much the user, 
Is it possible to : 
- Have a date range or a date to (in place of the fiscal year period select) with a meaningful message which describes that date range is for lots which will expire within this period
- Since depot and inventory are optional, could you add radio buttons (or checkbox) for showing them or not (you can see the Article en stock report)

image

client/src/i18n/en/tree.json
```diff
@@ -147,6 +147,7 @@
     "STOCK_ENTRY_REPORT":"[Stock] Entries Report",
     "STOCK_EXIT_REPORT" : "[Stock] Exit Report",
     "STOCK_INVENTORY" : "Items in Stock",
+    "STOCK_EXPIRATION_REPORT": "[Stock] expiration report",
```
jmcameron: I suggest capitalizing Expiration and Report here to match the other entries.

jmcameron:
I was able to check this out locally and test it.  It worked fine with bhima_test.

Unfortunately, I did not see the Stock Expiration report when I switched to the (updated) IMCK database and 'yarn app'.  See:
image

So I think there are two things to be fixed before merging:
- No report with IMCK
- The capitalization issue in client/src/i18n/en/tree.json

jmcameron (changes requested):
See: https://github.com/IMA-WorldHealth/bhima/pull/4757#issuecomment-666662468

client/src/i18n/en/tree.json
jeremielodi: okay

jeremielodi:
Did you run the migration queries?

jmcameron:
Yes. I tried to run them.   I downloaded the latest IMCK data, installed it, did 'yarn migrate'.  I had to hack that file a bit on my system (MySQL 8.0) to add many 'drops'.   I tried to preserve the migration items at the end, but did have to do some manual fixes (which I cannot guarantee are right!).   I'll attach my updated migration.sql file.  Note that I had to comment out some messed up SQL at line 4016.

[migration-imck.sql.txt](https://github.com/IMA-WorldHealth/bhima/files/5003899/migration-imck.sql.txt)

jniles:
@jmcameron there is a good chance that you don't have permission to see the new report.  By default, we don't auto-assign permissions for new modules.

To give yourself permissions (on the production database), go into Admin > Roles and give the administrative user permissions to all modules in the tree.  That should make this report show up.

jmcameron:
> @jmcameron there is a good chance that you don't have permission to see the new report. By default, we don't auto-assign permissions for new modules. To give yourself permissions (on the production database), go into Admin > Roles and give the administrative user permissions to all modules in the tree. That should make this report show up.

@jniles That did not help.  I will try again later after (hopefully), I've straightened out the migration script issue.

jmcameron:
@jniles Actually, fixing the role permissions did work (after I did it for the right account!).

So my only remaining issue with this PR is the capitalization issue in the English labels.

jmcameron:
BTW, after I merged this branch (locally) to 'master' to get the recent updates to get the fixes to the migration scripts, downloading, installing, and updating the IMCK database went smoothly without problems.

jmcameron:
I think this PR is ready except for fixing line 150 of client/src/i18n/en/tree.json,  it should be capitalized in English like this:

   "STOCK_EXPIRATION_REPORT": "[Stock] Expiration Report",

jniles (approved):
Looks like the last change requested by @jmcameron has been merged.  Thanks for your reviews @jmcameron and for your work @jeremielodi .

bors r+

PR #4759, opened 2020-07-26
fix(reports): typo in system usage statistics

jniles:
Fixes a typo in the system usage statistics report.  Also uses Promise.all() instead of three awaits.


jniles:

jmcameron (approved):
These changes look good to me.

jeremielodi (approved):
LGTM

PR #4762, opened 2020-07-28
Lot tags feature

### LINKED ISSUE #2662: feat(stock): Adding a tag system for stock entry
This issue is about to have a mean of tagging the stock with any key words (project name).

This feature will be useful for knowing in a given depot, the stock which correspond to one or more tags.

mbayopanda:
This PR add tagging feature in lots registry

A tag module exists in /Administration/Tags : we have to create tags first then we can use those tags in the lots registry

Capture d’écran 2020-07-28 à 22 08 20
Capture d’écran 2020-07-28 à 22 08 29
Capture d’écran 2020-07-28 à 22 08 48
Capture d’écran 2020-07-28 à 22 08 59

Close #2662


jmcameron:
How will the tags be used?

- I see the Admin/Tags page that allows the creation of arbitrary tags.  I assume that tags will be used (eg, in Lots management) to facilitate searching.  However, it is confusing to me why we need a global pool of tags.   I suppose that is to prevent mistyping.   But that makes it difficult to create new tags when we are in the Lots Management page.   We may need a popup menu on the Lots page (and others as gets used in other pages) that will allow creating tags on the fly (rather than leaving and going to the Admin/Tags page).   
- A related question is why should we share the same pool of tags with everything.  Shouldn't the lots tags be distinct from all the other modules that need tags?  This could be done "under the hood" by associating a "module/user" with each tag (selected from a fixed list).  Then anytime we see tags in the Lots module, it would selected from the right kind of tags.   There are other ways to handle this.  Not sure it is worth the effort.   
- In the Admin/Tags management, is there any way to find out where specific tags are being used?   For instance, if we want to delete or rename a tag, it would be useful to know if that might cause problems by the items using that tag.

mbayopanda:
Thank you @jmcameron for your review : 

- The concept of tags is not limited to lots only, the idea is we can use tags anywhere I know that the Tags Module is too minimal but we can improve it

- I really like your idea to know where a tag is used, actually, it is not implemented but we can do that

mbayopanda:
@jmcameron I am open to all suggestions that you can make 👍

jmcameron:
If this PR is intended to resolve Issue https://github.com/IMA-WorldHealth/bhima/issues/2662, please link it.  If it is to partially address the issue, please mention this issue in the description at the top.

mbayopanda:
@jmcameron I have a question about how can we link the tags to where they are used? because the idea behind tags is we can use them not just with lots only, maybe it can require a new module or registry

jniles:
@mbayopanda what is the progress here?  When will you be ready for another review?

mbayopanda:
@jniles can I get a review, I skipped some tests which are now broken

jmcameron (approved):
I skimmed through the code and did not spot any problems.  I tried testing the basic functionality using bhima_test and it seems to work okay.    So it is ready to go, as far as I'm concerned but I do suggest you get one more review from someone else on the team.

jmcameron:
> I skipped some tests which are now broken

What does this mean?  It looks like there are some good tests you've added.  Do feel any additional tests need to be added?

jmcameron:
I would still like to get an explanation of how tags are going to be used.   I think this addition is fine (I did approve it), but it is not clear to me how the users of BHIMA will use this feature.

client/src/js/components/bhTagSelect/bhTagSelect.js
```diff
@@ -0,0 +1,52 @@
+angular.module('bhima.components')
+  .component('bhTagSelect', {
+    templateUrl : 'js/components/bhTagSelect/bhTagSelect.html',
+    controller  : TagSelectController,
+    transclude  : true,
+    bindings    : {
+      tagUuids         : '<',
```
jniles, line 7: Are these optional?  If so, can you make it `<?`?

client/src/js/components/bhTagSelect/bhTagSelect.js
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Tag selection component
+ */
+function TagSelectController($rootScope, Tags, Notify) {
+  const $ctrl = this;
+
+  $ctrl.createUpdateTagsModal = Tags.createUpdateTagsModal;
+
+  $rootScope.$on('TAGS_CHANGED', () => {
```
jniles, line 24: Why is this needed?  I don't see and `$broadcast` or `$emit` calls.

Generally, we should probably avoid the event emitter pattern.  We've used it a couple times (for example, I used it in the cash modal), but those are usually when we didn't know how to make a proper data model passing data back and forth.  In general, we should use bindings for inputs and callbacks for outputs.

client/src/js/components/bhTagSelect/bhTagSelect.js
```diff
@@ -0,0 +1,52 @@
...
+    $ctrl.onSelectCallback({ tags : $ctrl.tagUuids });
+    loadTags();
+  };
+
+  $ctrl.getTagColor = t => {
+    return t ? { color : t.color, 'font-size' : '14px' } : null;
+  };
+
+  function loadTags() {
+    Tags.read(null)
```
jniles: Why `read(null)`?

mbayopanda:
@jmcameron we can think about tag usage as for example if we have computers in our depot, and we want to separate computer destined to each service of the organization, this case can occur if in the depot all computers for all purchase orders but we can have some computers purchased just for a given service and we don't want to give them to other services.

Tags can help us to perform this kind of distinction

jmcameron:
> Tags can help us to perform this kind of distinction

Thank you @mbayopanda.  That was helpful.

client/src/i18n/en/form.json
```diff
@@ -882,6 +882,7 @@
             "QUANTITY": "Enter a quantity"
         },
         "SELECT": {
+            "TAGS":"Select tags to join",
```
jniles: I think "select tags to add" is a more natural way of saying this in English.

client/src/js/components/bhTagSelect/bhTagSelect.js
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  // fires the onSelectCallback bound to the component boundary
+  $ctrl.onChanges = () => {
+    $ctrl.onSelectCallback({ tags : $ctrl.tagUuids });
+    loadTags();
+  };
+
+  $ctrl.getTagColor = t => {
+    return t ? { color : t.color, 'font-size' : '14px' } : null;
```
jniles: Is `t` ever `null`?

jniles: Also, can you just put 'font-size' as a `style` on the element?

client/src/modules/debtors/groups.update.html
```diff
@@ -111,7 +111,8 @@
               </ui-select-match>
               <ui-select-choices repeat="val.value as val in GroupUpdateCtrl.colors | filter: {'name' :$select.search}">
                 <span><i class="fa fa-circle" style="color: {{val.value}}; font-sze:14px"></i> <span translate>{{ val.name }}</span></span>
-              </ui-select>
+              </ui-select-choices>
+            </ui-select>
```
jniles, line 115: 👍

client/src/modules/stock/lots/registry.js
```diff
@@ -146,6 +150,11 @@ function StockLotsController(
       .finally(toggleLoadingIndicator);
   }
 
+  // returns a style for a given tag
+  vm.getTagColor = t => {
+    return t ? { color : t.color, 'font-size' : '14px' } : null;
```
jniles: Since this is in `for loop` in a cell in a `uiGrid`, for performance reasons, let's pre-process this.  Stick it right on the tag like this:
```js
vm.gridOptions.data = lots.map(lot => {
  let tagNames = []
  lot.tags.forEach(tag => {
    tag.style = getTagColor(t);
  });
   return lot;
});
```

Now, in your cell, you can use `ng-style="tag.style"`.  That should make it not have as many $digest calls.

jniles: Also, does the `font-size` do anything here?

client/src/modules/tags/modal/createUpdate.js
```diff
@@ -2,12 +2,18 @@ angular.module('bhima.controllers')
...
+  '$uibModalInstance', '$rootScope', 'ColorService',
 ];
 
-function TagsModalController(data, $state, TagsService, Notify, Instance, $rootScope) {
+function TagsModalController(
+  data, TagsService, Notify, Instance, $rootScope, Colors,
+) {
   const vm = this;
+  vm.colors = Colors.list.map(e => {
+    e.iconColor = { color : e.value, 'font-size' : '14px' };
```
jniles: Since this is more than a `iconColor` (it has the font size in it), it might make sense to call it `iconStyle`.

client/src/modules/tags/modal/createUpdate.js
```diff
@@ -20,6 +26,14 @@ function TagsModalController(data, $state, TagsService, Notify, Instance, $rootS
       return false;
     }
 
+    if (typeof (vm.tags) === 'object') {
+      delete vm.tags.iconColor;
+    } else if (Array.isArray(vm.tags)) {
+      vm.tags.forEach(t => {
+        delete t.iconColor;
+      });
```
jniles: It might be cleaner to just turn this into an array.

```js
// delete all iconColors on all tags
[].concat(vm.tags).forEach(tag => {
  delete tag.iconColor;
});
```

client/src/modules/tags/tags.ctrl.js
```diff
@@ -41,7 +34,10 @@ function TagsController($uibModal, Tags, Modal,
     vm.errorState = false;
     Tags.read()
       .then(tags => {
-        vm.gridOptions.data = tags;
+        vm.gridOptions.data = tags.map(t => {
+          t.iconColor = { color : t.color, 'font-size' : '14px' };
```
jniles: Since you do this about 3+ times in this PR, I suggest you put this in a Service.  Like the `TagService`.  Or `ColorService`. 

```js
Tags.formatTagStyle = tag => {
  tag.style = { color : tag.color, 'font-size': '14px' };
};
```

client/src/modules/tags/templates/name.cell.html
```diff
@@ -0,0 +1,6 @@
+<div
+  class="ui-grid-cell-contents"
+  data-row="{{row.entity.name}}">
```
jniles: Do you need the `data-row` binding?  I can't tell where it is used.

client/src/modules/tags/tags.service.js
```diff
@@ -1,14 +1,30 @@
...
  */
-function TagsService(Api) {
-  const service = new Api('/tags/');
+function TagService(Api, util, $uibModal) {
+  const baseUrl = '/tags/';
+  const service = new Api(baseUrl);
+  service.types = new Api(baseUrl.concat('types/'));
+
+  const tagKeys = ['uuid', 'name', 'color'];
+  service.clean = tag => util.maskObjectFromKeys(tag, tagKeys);
```
jniles, line 20: Ooooh.  I had forgotten we had this service.  Excellent 👍

client/src/modules/tags/tags.service.js
```diff
@@ -1,14 +1,30 @@
...
+
+  service.createUpdateTagsModal = (tag) => {
+    $uibModal.open({
+      templateUrl : 'modules/tags/modal/createUpdate.html',
+      controller : 'TagsModalController as ModalCtrl',
+      resolve : { data : () => tag },
```
jniles, line 26: Beautiful 👍

server/controllers/stock/index.js
```diff
@@ -521,7 +522,7 @@ function listMovements(req, res, next) {
  * GET /stock/lots/depots/
  * returns list of each lots in each depots with their quantities
  */
-function listLotsDepot(req, res, next) {
+async function listLotsDepot(req, res, next) {
```
jniles, line 526: 👍

test/end-to-end/tags/tags.spec.js
```diff
@@ -13,33 +13,36 @@ function tagsManagementTests() {
...
   it('should add a third tags', async () => {
     await page.openCreateModal();
     await page.setName('Test tag');
+    await page.setColor('Vert');
     await page.submit();
     await components.notification.hasSuccess();
   });
 
-  it('should edit tags', async () => {
+  it.skip('should edit tags', async () => {
```
jniles: Do you know why these fail?  Any ideas?

server/models/schema.sql
```diff
@@ -2184,6 +2192,7 @@ CREATE TABLE `service_fee_center` (
 CREATE TABLE `tags`(
   `uuid` BINARY(16) NOT NULL PRIMARY KEY,
   `name` VARCHAR(100) NOT NULL,
+  `color` VARCHAR(20) NULL,
```
jniles: Since you are adding this property, I would go ahead and make it much longer.  VARCHAR(20) is super limiting.

jniles (commented):
@mbayopanda I haven't given this a proper test, but I've looked over the code.  It looks good!  I'm a bit concerned about the performance implications of these tags, especially in the JOIN and the ui-grid.  I've added a few comments to try and improve on the client-side grid performance and tidy up the code.

client/src/js/components/bhTagSelect/bhTagSelect.js
mbayopanda: The reason of that is when we call the `tag modal` via the `lot edit modal` and we add a new `tag` from this `lot edit modal`, we must add this listener if not the list of tags loaded in the `lot edit modal` will not be up to date

client/src/js/components/bhTagSelect/bhTagSelect.js
mbayopanda: No, they were required when I created that component

client/src/js/components/bhTagSelect/bhTagSelect.js
mbayopanda: yes `t` can be `null` if nothing is selected from the list of tags

client/src/modules/tags/templates/name.cell.html
mbayopanda: It was used for `GridRow` tool for e2e tests, it is used also with: `/modules/roles/templates/action.cell.html`

test/end-to-end/tags/tags.spec.js
mbayopanda: Just after adding the color picker in the tag modal, these tests started to fail

mbayopanda:
Thank you @jniles for the review

client/src/modules/tags/tags.ctrl.js
mbayopanda: @jniles we have these three variants, and i am truggling a bit : 
```js
$ctrl.getTagColor = t => {
  return t ? { color : t.color } : null;
};

function addColorStyle(tag) {
  tag.style = { color : tag.color };
}

function addIconStyle(item) {
  item.iconStyle = { color : item.value };
  return item;
}
```
Finally, the `font-size` doesn't matter.

client/src/js/components/bhTagSelect/bhTagSelect.js
```diff
@@ -0,0 +1,52 @@
...
+
+TagSelectController.$inject = ['$rootScope', 'TagService', 'NotifyService'];
+
+/**
+ * Tag selection component
+ */
+function TagSelectController($rootScope, Tags, Notify) {
+  const $ctrl = this;
+
+  $ctrl.createUpdateTagsModal = Tags.createUpdateTagsModal;
```
jniles, line 22: This returns a promise, right?  Instead of using `$rootScope.$on()`, let's just do this:
```js
$ctrl.createUpdateTasModal = () => {
  Tags.createUpdateTagsModal()
    .finally(() => loadTags());
};
```

Would that work to remove the event-emitter pattern?

(you might also use a `.then()` instead - not sure what the return values of the tags modal is)

client/src/js/components/bhTagSelect/bhTagSelect.js
```diff
@@ -0,0 +1,52 @@
...
+
+  $ctrl.getTagColor = t => {
+    return t ? { color : t.color } : null;
+  };
+
+  function loadTags() {
+    Tags.read()
+      .then(tags => {
+        const identifiers = $ctrl.tagUuids.map(t => t.uuid);
+        $ctrl.tags = $ctrl.tagUuids.length ? tags.filter(t => identifiers.includes(t.uuid) === false) : tags;
```
jniles: Minor nitpick - if `$ctrl.tagUuids.length === 0`, then `identifiers.length === 0`.  Since you are actually using `identifiers` on this line, I think it cleaner to write:
```js
$ctrl.tags = identifiers.length ? tags.filter(t => identifiers.includes(t.uuid) === false) : tags; 
```

client/src/modules/tags/templates/name.cell.html
jniles: `GridRow` is used for clicking on dropdown menu items in the `actions` cells of tests.  What do you need to the `data-row` selector for here?  What are you checking/clicking on?

server/controllers/stock/index.js
```diff
@@ -530,12 +531,19 @@ function listLotsDepot(req, res, next) {
     params.defaultPeriodEntry = params.defaultPeriod;
     delete params.defaultPeriod;
   }
-  core.getLotsDepot(null, params)
-    .then((rows) => {
-      res.status(200).json(rows);
-    })
-    .catch(next)
-    .done();
+  try {
```
jniles, line 535: So, if I am reading this code correctly, if we have N lots in the depot, we will issue N + 1 SQL requests:
```js
// first SQL request (num SQL requests = 1)


const rows = await Promise.all(data.map(async (line) => {
  // one SQL request per tag 
  const tags = await Lots.getLotTags(db.bid(line.uuid));
  line.tags = tags;
  return line;
}));

// Total SQL requests N + 1
```

I'm not sure something like a raspberry pi could handle this kind of a query.  Especially if no one uses tags, we'll be slowing down our queries for nothing.

I propose you either:
  1. Do something like `GROUP_CONCAT()` to get the tag information in a single SQL query.  This will be messy, but doable.
  2. Do two queries and stitch the results together in javascript.  You are already looping through each lot once, and it will save us from hammering our SQL server.  It would look something like this:
```js
const lotUuids = data.map(row => db.bid(row.uuid));
const tags = await db.exec('SELECT lot_uuid, tag.color /*, etc ... */ WHERE lt.lot_uuid IN (?)', [lotUuids]);


// make a lot_uuid -> tags map.
const tagMap = _.groupdBy(tags, 'lot_uuid');

data.forEach(lot => {
  lot.tags = tagMap[lot.uuid];
});

// Total SQL requests 2.
// Loops are O(N + M) where length(lots) is N and length(tags) is M.

```

test/end-to-end/tags/tags.page.js
```diff
@@ -7,22 +7,27 @@ function TagsPage() {
...
+  const gridId = 'tags-grid';
 
   page.gridId = gridId;
   page.setName = setName;
   page.editTags = editTags;
   page.deleteTags = deleteTags;
   page.openCreateModal = openCreateModal;
+  page.setColor = setColor;
 
-  const departmentName = element(by.model('ModalCtrl.tags.name'));
```
jniles, line 18: 🤣

server/controllers/stock/index.js
jniles: I suppose as an improvement, you can check if `tags.length === 0` and return early if that is the case.  It would save a for loop in the default case.

jniles (commented):
I'm still tracking down the issues with the end to end tests - they take a long time to run!

Overall, this is really good.  I made a proposal about restructuring your SQL requests so that we don't shoot off a huge number of small queries.

I have also made a proposal on the use of the event emitter.  Can you see if that works?  I think it's a lot less magical than the `$on()` events and makes it easier for us to go to vue/angular2/react if we ever do that.

client/src/modules/tags/templates/name.cell.html
jniles: I actually think this is breaking the end to end tests.  When I remove it, I think the end to end tests pass. 👍

client/src/js/components/bhTagSelect/bhTagSelect.js
mbayopanda: Sorry, it seems like it doesn't work

test/end-to-end/tags/tags.spec.js
jniles: Please remove these skips.

mbayopanda:
@jniles can i get a review please

jniles:
Why did you use grid action?  That is a step backwards. :/

mbayopanda:
GridRow still doesn't pass the test

client/src/js/components/bhTagSelect/bhTagSelect.js
jniles: Turns out it doesn't return a promise.  This isn't a blocking change.  I'll open a pull request after this lands to change the behavior.

jniles (approved):
This works for me!  Thanks @mbayopanda. All comments have been addressed.

bors r+

PR #4764, opened 2020-07-29
Added client unit test for PurchaseOrderForm

### LINKED ISSUE #4748: Write Unit Tests for the Purchase Order Form Service
We have unit tests for many of our client "form" services.   They are important to test form logic (totaling, error handling) separate and apart from the rendering of the form.   Here are some examples:

  1. https://github.com/IMA-WorldHealth/bhima/blob/master/test/client-unit/services/PatientInvoiceForm.spec.js
  2. https://github.com/IMA-WorldHealth/bhima/blob/master/test/client-unit/services/VoucherForm.spec.js
  3. https://github.com/IMA-WorldHealth/bhima/blob/master/test/client-unit/services/StockEntryModalForm.spec.js

We need this for the `PurchaseOrderFormService`.  It is in this file:  https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/purchases/create/PurchaseOrderForm.js

At minimum, we should verify in our tests that:
  1. The `setup()` method sets up default data.
  2. The `validate()` method returns and error for invalid data.
  3. The `digest()` method sets the totals.
  4. The `addItem()` method grows the internal store.
  5. The `removeItem()` method removes the correct item from the store.

This would at minimum be a good starting point.

jmcameron:
Thanks to @jniles for his help and suggestions on this.

Hopefully, this is close to good.   Minor questions I still have:

- In the #digest() test, I had to manually adjust the PO items for digest() to work.   Updating the quantity and unit_price seems okay to me since I wanted control of the amounts to make it easier to verify the correct calculations.  But having to manually set item._hasValidAccount seems clunky and I wonder if there is a better way to do this?
- Are any of the services included in the setup not necessary?


jmcameron:
Had to remerge from master when my rebase skipped all the previous merges with master that I had done in the branch.   I'll figure out git rebase eventually!

lomamech (approved):
image

Unit test successfully reviewed

test/client-unit/services/PurchaseOrderFormService.spec.js
```diff
@@ -0,0 +1,148 @@
...
+  beforeEach(module(
+    'bhima.services',
+    'angularMoment',
+    'ui.bootstrap',
+    'ui.router',
+    'ngStorage',
+    'pascalprecht.translate',
+    'tmh.dynamicLocale',
+    'bhima.StockMocks',
+    'bhima.mocks',
```
jniles, line 20: Since you asked in your comments - nope, this is the minimal set of all possible modules.  I tried to tweak it manually, but you need each one of these.

test/client-unit/services/PurchaseOrderFormService.spec.js
```diff
@@ -0,0 +1,148 @@
...
+    // First, verify that the initial cost is 0
+    expect(form.details.cost).to.equal(0);
+
+    // Add a PO item to the form
+    const item1 = form.addItem();
+    item1.inventory_uuid = uuid1;
+    form.configureItem(item1);
+    item1.quantity = 2;
+    item1.unit_price = 2.30;
+    item1._hasValidAccounts = true;
```
jniles, line 128: You are implicitly testing both the `PurchaseOrderForm` and the `PurchaseOrderFormItem` here.  Each item is actually a subclass to itself.  When you call `configure()` on it, it should validate that it has all the correct accounts set.

I think this is fine, even if a bit clunky.  The only way to make this better would be to improve the mock data .. but a better order of operations would be to write unit tests for the `PurchaseOrderItem` service (thereby forcing the development of that mock data) and then use the mock data from that to power the "correct" scenario here.

jniles (commented):
I see you tagged me for a review and had some questions.  I think these tests are really good.  I couldn't have written better ones myself.

test/client-unit/services/PurchaseOrderFormService.spec.js
jmcameron: @jniles Thanks for looking at these!

PR #4765, opened 2020-07-30
feat: use bhCheckboxTree for employee config

jniles:
Uses the bhCheckboxTree for the employee configuration to save configuration by employee.

Onm9f7x6q5


jmcameron (approved):
This looks good.   I checked it out and tried it and it works smoothly and correctly.

However, as a future enhancement, if there are a lot of people that should appear on the checkbox form, we should consider formatting the form into multiple columns, so that the length of the pop-up selection form is not so long.     I tried it with the IMCK database and I had to scroll a long ways to get to the bottom and the [Submit] button.

jniles:

Indeed, you are right that is a good future enhancement.  We've previously done this with ng-repeats, I wonder how straightforward it would be to do with CSS...

jniles:
bors r=jmcameron

PR #4770, opened 2020-07-31
fix(i18n): better translation on link to inventory

### LINKED ISSUE #4706: Improve French Translation of Articles in Stock link to the Inventory
This link links the inventory from the articles in stock registry:
image

It's not at all obvious from the word "liste".  We should improve it.

jniles:
Gives a better translation for the Articles in Stock link to the inventory.

Closes #4706.

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -137,7 +137,7 @@
       "EXIT_DEPOT"     : "Stock exit to depot",
       "EXIT_SERVICE"   : "Stock exit to service",
       "EXIT_SERVICES"   : "Stock exit to services",
-      "EXIT_LOSS"      : "Stock Exit - Loss"
+      "EXIT_LOSS"      : "Stock exit to loss"
     },
     "RECEIVER"        : "Receiver",
     "RECEPTION_TRANSFER" : "Transfer",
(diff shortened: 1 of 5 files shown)


client/src/i18n/en/stock.json
```diff
@@ -135,7 +135,6 @@
       "EXIT_PATIENT"   : "Stock exit to patient",
       "EXIT_PATIENTS"   : "Stock exit to patients",
       "EXIT_DEPOT"     : "Stock exit to depot",
-      "EXIT_DEPOTS"     : "Stock exit to depots",
       "EXIT_SERVICE"   : "Stock exit to service",
       "EXIT_SERVICES"   : "Stock exit to services",
       "EXIT_LOSS"      : "Stock Exit - Loss"
```
jmcameron: I noticed that this one is a bit different format from the previous ones.  On the Stock Exit report, they are all presented at the same level, etc, so it looks a bit odd for this one to be different.

jmcameron: The French version is consitent...

jmcameron (approved):
These changes look good.  I would suggest updating the "Stock Exit - Loss" one (English) to match the rest.   Actually they call could match EXIT_LOSS; they just should be consistent.

jniles:
@jmcameron thanks for the review!  Will make the updates.

jniles:
@jmcameron changes landed.  Let me know what you think!

(this might be a good time to try "bors" out to see if it works for you if you are happy with these pull request changes)

jmcameron:
Looks good to me.   I'm impressed by your expressive branch names!

jmcameron:
bors+

jniles:
@jmcameron the syntax is "bors r+".  Give it an updated try!  You'll need to use a new comment - I don't think it listens on edits.

jniles:
Yay!  Looks like that worked!

PR #4774, opened 2020-07-31
Update migration script to use mysql calls to use username/password

### LINKED ISSUE #4766: Update migration script to drop all function, triggers, etc
The current migration script used the production env but does not use the MYSQL user/password, so additions (such as dropping exiting triggers, etc) that require a mysql call fail and the absence of these drops cause failures when the migration script is used on MySQL 8.0.    Also, functions are not getting dropped now but should be.

Goal: Update the script so that all items that are replaced get DROPped first.

jmcameron:
And fixed errors in the next/migrate.sql file

- Disabled problematic drop village_1 index
- Fixed a problem probably due previous fixes for pull request merging problems on previous releases.

These fixes make the 'yarn migrate' and subsequent application of the migration file to work for me on my system with MySQL 8.0

Testing:  This has only been test with recent  (today) mysql dumps of IMCK data.

When testing, when you run 'yarn migrate', make sure there are no complaints!

NOTE: I was wrong about need additional code to drop functions; it was doing that already.

### DIFF
--- a/server/models/migrations/next/migrate.sql
+++ b/server/models/migrations/next/migrate.sql
@@ -89,7 +89,9 @@ ALTER TABLE `province` DROP INDEX `province_1`;
 
 ALTER TABLE `sector` DROP INDEX `sector_1`;
 
-ALTER TABLE `village` DROP INDEX `village_1`;
+/* ??? ALTER TABLE `village` DROP INDEX `village_1`; 
+JMC: Disabled since this fails when using migration script  2020-07-31
+*/
 
 /*
 @author jmcameron
@@ -106,9 +108,6 @@ ALTER TABLE `enterprise_setting` ADD COLUMN `enable_auto_purchase_order_confirma
 */
 
 ALTER TABLE `enterprise_setting` MODIFY COLUMN `enable_daily_consumption` TINYINT(1) NOT NULL DEFAULT 0;
-
-ALTER TABLE `village` DROP INDEX `village_1`
-  ('collectionCapacity', 'REPORT.COLLECTION_CAPACITY.TITLE');
 
 /**
  * @author: jeremielodi
--- a/sh/setup-migration-script.sh
+++ b/sh/setup-migration-script.sh
@@ -43,14 +43,14 @@ SET collation_connection = 'utf8mb4_unicode_ci';
 " > $MIGRATION_FILE
 
 echo "[migrate] Adding DROP TRIGGERS for $DATABASE."
-mysql -e "SELECT CONCAT('DROP TRIGGER IF EXISTS ', trigger_name, ';') FROM information_schema.triggers WHERE trigger_schema = '$DATABASE';" \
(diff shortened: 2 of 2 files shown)


jmcameron:
Updated next/migrate.sql based on @jniles PR https://github.com/IMA-WorldHealth/bhima/pull/4772.  Still have to disable DROPs of village_1 since these fail on my MySQL 8.0 setup.   Not sure why the DROP of sector_1 does not fail too.  But all DROPs of village_1 index must be removed for the migration-mysql.sql SQL code to work on my setup with MySQL 8.0.

Maybe we should guards on these kinds of DROPs (eg, if exists... DROP...).  I'll add an example I figured out in my next comment.

jniles:
@jmcameron I think you may be on a wild goose chase because of our silliness.  You aren't crazy - that same drop fails on my machine.  The drop village is actually installation/production database dependent.  Someone already did it manually on the IMCK database which is why it is failing you.  On Vanga, it doesn't fail because the key exists.

server/models/migrations/next/migrate.sql
```diff
@@ -89,7 +89,9 @@ ALTER TABLE `province` DROP INDEX `province_1`;
 
 ALTER TABLE `sector` DROP INDEX `sector_1`;
 
-ALTER TABLE `village` DROP INDEX `village_1`;
+/* ??? ALTER TABLE `village` DROP INDEX `village_1`; 
+JMC: Disabled since this fails when using migration script  2020-07-31
```
jniles, line 93: Fair enough!

sh/setup-migration-script.sh
```diff
@@ -43,14 +43,14 @@ SET collation_connection = 'utf8mb4_unicode_ci';
 " > $MIGRATION_FILE
 
 echo "[migrate] Adding DROP TRIGGERS for $DATABASE."
-mysql -e "SELECT CONCAT('DROP TRIGGER IF EXISTS ', trigger_name, ';') FROM information_schema.triggers WHERE trigger_schema = '$DATABASE';" \
+mysql -u $DB_USER --password=$DB_PASS -e "SELECT CONCAT('DROP TRIGGER IF EXISTS ', trigger_name, ';') FROM information_schema.triggers WHERE trigger_schema = '$DATABASE';" \
```
jniles, line 46: Thanks, this is helpful and explains a lot. 

It's clear to me now.  Why does it work on my machine?  Because I have a "superuser" in my ~/.my.cnf file that has access to all these databases.  So even though this doesn't have a user passed in, it looks up the user from my environmental file.  Whoops!  Thanks for catching this.

jniles (approved):
Looks good to me.  Great catch!

jmcameron:
I figured out how to get a procedure working to delete indexes if they exist.   I tried to get this to work as a simple SQL IF statement without luck.   In any case, we should consider adding this function to the top of the SQL includes so that we avoid these problems.  I would be very interested to know if the attached script (for IMCK) works with others with their setups, different versions of MySQL, etc.
[drop_index.sql.txt](https://github.com/IMA-WorldHealth/bhima/files/5008983/drop_index.sql.txt)

NOTE: If we decide to use this, it should be generalized to not hard-code the database name.

PR #4784, opened 2020-08-03
feat(SQL procedures) drop columns and indexes if exist

### LINKED ISSUE #4783: Lack of drop index or column if it exists procedure
While running the migration file, sometime we drop columns or indexes if needed but if it doesn't exist, the process will be stopped due to the Sql exception. So, we need procedures to fix that.

jeremielodi:
closes #4783

### DIFF
--- a/server/models/procedures.sql
+++ b/server/models/procedures.sql
@@ -75,4 +75,10 @@ SOURCE server/models/procedures/payroll.sql
   analysis procedures
 */
 SOURCE server/models/procedures/analysis.sql
+
+/*
+  migration process procedures
+*/
+SOURCE server/models/procedures/migration-process.sql
+
 DELIMITER ;
--- a/test/server-unit/mysql_functions.spec.js
+++ b/test/server-unit/mysql_functions.spec.js
@@ -1,0 +1,133 @@
+/* eslint global-require:off */
+const { expect } = require('chai');
+
+function DatabaseUnitTests() {
+  let db;
+  before(() => {
+    db = require('../../server/lib/db');
+  });
+
+  const tableName1 = 'my_test_table1';
+  const tableName2 = 'my_test_table2';
+  const constraint1 = 'constraint1';
+  const constraint2 = 'constraint2';
+
+  const createTable1Sql = `
+    CREATE TABLE IF NOT EXISTS ${tableName1} (
+      uuid BINARY(16) PRIMARY KEY,
+      name VARCHAR(40)
+    )ENGINE=InnoDB DEFAULT CHARSET=latin1;
+  `;
+
+  const createTable2Sql = `
+    CREATE TABLE IF NOT EXISTS ${tableName2} (
+      uuid BINARY(16) PRIMARY KEY,
+      type VARCHAR(40),
+      uuid_parent BINARY(16),
+      uuid_parent2 BINARY(16),
+      INDEX \`type\`(\`type\`),
+    CONSTRAINT \`${constraint1}\` FOREIGN KEY (\`uuid_parent\`)  
+    REFERENCES \`${tableName2}\` (\`uuid\`) ON DELETE CASCADE,
+
+    CONSTRAINT \`${constraint2}\` FOREIGN KEY (\`uuid_parent2\`)
+    REFERENCES \`${tableName2}\` (\`uuid\`) ON DELETE CASCADE
+    )ENGINE=InnoDB DEFAULT CHARSET=latin1;
+  `;
+
+  const checkIfConstraintExistSql = ` SELECT Constraint_exists("${tableName2}", "${constraint1}") as exist;`;
+
+  const checkIfConstraintIfExistSql = ` SELECT Constraint_exists("${tableName2}", "unknown") as exist;`;
+  const dropFKExistSql = ` SELECT Constraint_exists("${tableName2}",  "${constraint2}") as exist;`;
+  const dropForeignKeySql = ` CALL drop_foreign_key("${tableName2}", "${constraint2}");`;
+  const dropUnknownForeignKeySql = ` CALL drop_foreign_key("${tableName2}", "unknown");`;
+  const dropUnknownColumnSql = ` CALL drop_column_if_exists("${tableName2}", "unknown");`;
+  const dropColumnSql = ` CALL drop_column_if_exists("${tableName2}", "uuid_parent2");`;
+  const checkIfColumnExistSql = ` SELECT column_exists("${tableName2}", "uuid_parent2") as exist;`;
+
+  const checkIfIndexExistSql = ` SELECT index_exists("${tableName2}", "type") as exist;`;
+  const dropUnknownIndexSql = ` CALL drop_index_if_exists("${tableName2}", "unknown");`;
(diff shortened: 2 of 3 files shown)


jmcameron (changes requested):
This looks good!   But we really needs tests to confirm that the procedures work.  These would not be too hard to do with a series of SQL statements:  Delete non-existent items (without errors).  And create obscurely named items, call these procedures to delete them, and the verify they've been deleted.

I struggled with the one for indexes.  I'm not convinced the one from stackoverflow will actually work (I have not tried it).  However, the one I created from these examples and actually tested (and I'm pretty sure it works) is at:  https://github.com/IMA-WorldHealth/bhima/files/5008983/drop_index.sql.txt.   It only needs a little updating to generalize the database, etc.   My experience with stackoverflow is that it is often very helpful, but that some people post things that they think should work without actually trying them! Caveat Emptor.

jeremielodi:
I my self tested both of theme, they worked.

jeremielodi:
You can run this to see the result

```sql


CREATE TABLE IF NOT EXISTS test11 (

id INT PRIMARY KEY,
`name` VARCHAR(40) NOT NULL,
INDEX `name_index1`(`name`)
);

DESC test11;

call drop_column_if_exists('test11', 'name_test');
DESC test11;


call drop_column_if_exists('test11', 'name');
DESC test11;

-- drop_index_if_exists

call drop_index_if_exists('test11', 'name_');
DESC test11;


call drop_index_if_exists('test11', 'name_index1');
DESC test11;

```

jmcameron:
Could you create a server test (.spec.js) that does this?

jmcameron:
And also for the other functions too?

jeremielodi:
Okay I can try to do that

server/models/procedures.sql
```diff
@@ -75,4 +75,10 @@ SOURCE server/models/procedures/payroll.sql
   analysis procedures
 */
 SOURCE server/models/procedures/analysis.sql
+
+/*
+  migration process procedures
+*/
+SOURCE server/models/procedures/migration-process.sql
```
jniles, line 82: I don't think you need these here since they aren't useful in the actual code, they are only useful for migrations and this file gets built when we create the database the first time on an installation.

I wonder if it would be better to just change the bash file of the migration script to build this when you want to do a migration and then clean it up afterwards.

jeremielodi: Maybe the file name wasn't good but to my mind we will need to use those procedures in our migration files if we want to drop a column or index

test/server-unit/mysql_functions.spec.js
```diff
@@ -0,0 +1,225 @@
...
+  const dropUnknownForeignKeySql = ` CALL drop_foreign_key("${tableName2}", "unknown");`;
+  const dropUnknownColumnSql = ` CALL drop_column_if_exists("${tableName2}", "unknown");`;
+  const dropColumnSql = ` CALL drop_column_if_exists("${tableName2}", "uuid_parent2");`;
+  const checkIfColumnExistSql = ` SELECT column_exists("${tableName2}", "uuid_parent2") as exist;`;
+
+  const checkIfIndexExistSql = ` SELECT index_exists("${tableName2}", "type") as exist;`;
+  const dropUnknownIndexSql = ` CALL drop_index_if_exists("${tableName2}", "unknown");`;
+  const dropIndexSql = ` CALL drop_index_if_exists("${tableName2}", "type");`;
+
+  it('#should create the test table1', (done) => {
```
jniles: I think you can probably use async/await here without difficulties since Mocha should consume promises just fine.  See this article](https://zaiste.net/posts/modern-nodejs-async-await-testing-mocha-chai/).

So, you could instead write:
```js
it('#should create the test table1', async () => {
   await db.exec(createTable1Sql);
  expect(true).to.equal(true);
});
```

However, instead of checking that `true` is `true`, you could actually check to see if the table exists by executing a follow-up SQL:
```js
const [{ counter }] = await db.exec('SELECT COUNT(uuid) AS counter FROM table1');
expect(counter).to.equal(0);
```

test/server-unit/mysql_functions.spec.js
```diff
@@ -0,0 +1,225 @@
...
+    db.exec(checkIfIndexExistSql)
+      .then(res => {
+        expect(res.length).to.equal(1);
+        expect(res[0].exist).to.equal(0);
+        done();
+      })
+      .catch((error) => {
+        done(error);
+      });
+  });
```
jniles, line 129: Could this be an `after()` method that will delete these tables if they exist?  That way they don't stick around in the developer's database?

jniles (commented):
@jeremielodi this is on the right track.  I think you could simplify your code a lot by using async/await as proposed.

For many of these, you are simply asserting that `db.exec()` doesn't throw an error and that is it - you don't actually check to see if anything happened.  Can you see if you can come up with a way to make sure that, if the functions don't do anything at all, they don't still pass the test?

For example, you could try and use a unique index, remove it, then try to insert the same value twice and make sure that errors.

test/server-unit/mysql_functions.spec.js
jeremielodi: Okay, I also propose that we could update the schema.sql, for  every foreign key it is prudent to define it constraint name. That way, all our bhima instance will have the same database structure.

jeremielodi: You should know that in this case, testing the function is very easy than the procedure. because all those functions return a value. It's easy to test the returned value with mocha

test/server-unit/mysql_functions.spec.js
```diff
@@ -0,0 +1,133 @@
...
+  const dropColumnSql = ` CALL drop_column_if_exists("${tableName2}", "uuid_parent2");`;
+  const checkIfColumnExistSql = ` SELECT column_exists("${tableName2}", "uuid_parent2") as exist;`;
+
+  const checkIfIndexExistSql = ` SELECT index_exists("${tableName2}", "type") as exist;`;
+  const dropUnknownIndexSql = ` CALL drop_index_if_exists("${tableName2}", "unknown");`;
+  const dropIndexSql = ` CALL drop_index_if_exists("${tableName2}", "type");`;
+
+  it('#should create the test table1', async () => {
+    await db.exec(createTable1Sql);
+    expect(true).to.equal(true);
```
jmcameron: I don't understand how this test actually checks to see if the table was created. 
Couldn't you check the return value of 'await db.exec()' which should produce the error flag of the MySQL query call, correct?   So would this work?

  const result = await db.exec(createTable1Sql);
  expect(result).to.equal(true);

or something similar?

jmcameron: The additions look good to me!

jmcameron (approved):

test/server-unit/mysql_functions.spec.js
jmcameron: @jniles The changes look good to me and tests pass.   So I think it is ready to go.  Do you want to check on this last item about the after()  method?

jmcameron:
@jeremielodi Nice job!  Thanks for being patient with my suggestions!

jniles (approved):
This is a great starting point.  Thanks @jeremielodi !

bors r+

PR #4788, opened 2020-08-05
fix(stock): activate cancel button on depot modal

### LINKED ISSUE #4776: Stock Requisition popup cannot cancel
When you first launch the Stock Requisition page, it pops up a dialog forcing the selection of a Depot.  That part is fine.  
But it completely disables the [Cancel] button.   I'm sure this is to prevent  proceeding on the page without selecting a Depot.  

What if someone clicks on the [Requisition] button by accident and wants to back out?  The would normally click on the [Cancel] button.  But that is disabled.   

A secondary issue:  If the user only has access to one Depot, that depot will show in the drop-down menu of depots to select and it may be selected (or not).   Again, the [Cancel] button is disabled.  The behavior is a bit puzzling because it is not obvious that it is a Drop-down menu (you cannot select anything else).  Normally a drop-down menu has several options and it obvious what to do.

If we allowed a user to proceed to the page without selecting depot, perhaps the buttons on the top (eg, [New Requisition], etc) could force the pop-up to chose a depot if invoked and no depot is selected.

Also, If the user only has access to one depot, perhaps we should use that automatically without prompting for it?

jniles:
Ensures that the "cancel" button works on the depot selection modal.
This means that users can always cancel an operation and it will not
leave them in an undefined state.  If there is no default depot, it
returns them to the previous state.  If there is a defined depot, it
returns them to the current module.

Partially addresses #4776.

n1WcBYTwqQ

### DIFF
--- a/client/src/js/directives/bhChangeDepot.js
+++ b/client/src/js/directives/bhChangeDepot.js
@@ -41,7 +41,8 @@ function bhChangeDepotController(Depots, AppCache, Notify) {
       }
 
       function changeDepot() {
-        // if requirement is true the modal cannot be canceled
+        // if requirement is true the modal will use History.back() to
+        // cancel the modal.
         const requirement = !cache.depotUuid;
 
         return Depots.openSelectionModal($ctrl.currentDepot, requirement)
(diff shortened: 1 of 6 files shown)


jmcameron (approved):
I tested this and it works nicely.  Much more pleasant and intuitive.
I also tested this with the browser cache cleared so there was not default depot and it worked fine too.
Thanks!

PR #4791, opened 2020-08-05
feat(inventory): enforce account types on groups

### LINKED ISSUE #4703: Filter bhAccountSelects in Inventory Group create form on appropriate type
On the inventory group create/update form, we have the following accounts:
image

These accounts can be filtered:
  1. Income account should have the income account type.
  2. Stock account should have the asset account type.
  3. Expense account should have the expense account type.

Let's make this change.

jniles:
Enforce the account types for each type of account on the inventory group via the `bhAccountSelect` component.

Closes #4703.

Here is what it looks like:
u95E40ge8u

### DIFF
--- a/client/src/js/components/bhAccountSelect.js
+++ b/client/src/js/components/bhAccountSelect.js
@@ -1,7 +1,7 @@
 angular.module('bhima.components')
   .component('bhAccountSelect', {
     templateUrl : 'modules/templates/bhAccountSelect.tmpl.html',
-    controller  : AccountSelectController,
+    controller  : bhAccountSelectController,
     transclude  : true,
     bindings    : {
       accountId        : '<',
@@ -14,15 +14,17 @@ angular.module('bhima.components')
     },
   });
 
-AccountSelectController.$inject = [
+bhAccountSelectController.$inject = [
   'AccountService', 'FormatTreeDataService', 'bhConstants', '$scope',
 ];
 
 /**
  * Account selection component
  */
-function AccountSelectController(Accounts, FormatTreeData, bhConstants, $scope) {
+function bhAccountSelectController(Accounts, FormatTreeData, bhConstants, $scope) {
   const $ctrl = this;
+
+  const TITLE_ACCOUNT_TYPE_ID = bhConstants.accounts.TITLE;
 
   // fired at the beginning of the account select
   $ctrl.$onInit = function $onInit() {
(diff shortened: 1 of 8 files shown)


client/src/js/components/bhAccountSelect.js
```diff
@@ -44,16 +46,39 @@ function AccountSelectController(Accounts, FormatTreeData, bhConstants, $scope)
...
+      parsed = types.map(type => parseInt(type, 10));
+    } else if (typeof types === 'string') {
+      parsed = types.split(',')
+        .filter(type => type !== '')
+        .map(type => parseInt(type, 10));
+    } else if (typeof types === 'number') {
+      parsed = [types];
+    } else {
+      throw new Error('Cannot parse account types from '.concat(types));
+    }
```
mbayopanda, line 80: I guess we are not waiting to receive an object if it can occur maybe we can handle that also here

client/src/modules/inventory/configuration/groups/modals/actions.tmpl.html
```diff
@@ -41,6 +41,7 @@
       data-sales-account
       name="sales_account"
       account-id="$ctrl.session.sales_account"
+      account-type-id="$ctrl.INCOME_ACCOUNT_TYPE_ID"
```
mbayopanda, line 44: 👍

mbayopanda (approved):
LGTM 👍

client/src/js/components/bhAccountSelect.js
jniles: I actually think it's a good thing to throw an error if the developer doesn't respect the API.  I can't imagine a scenario in which an object should be accepted.

jmcameron (approved):
I tested this (with Bhima_test) and it worked fine.  I looked through the code and looked okay too.  
I was curious about two things:

- Why the Bhima Constants involved are INCOME/ASSET/EXPENSE yet the names used for these (in English) on the Add Inventory Group page are SALE/STOCK/COST?
- Why the 1x accounts (Durable Resources) are not included on this form.  Are the buildings not considered inventory?

jniles:

For the account names, I'm basing them off of what I've heard the accountants discuss.  For example, cost of goods sold (COGS) is a technical account name and it is _always_ an expense account.  But just saying "expense account" may be a little to generic.  I think of it as the difference between "limo garage" and "car garage".  Both are true, the former is much more contextual and specific to what the garage/account's function is.

You make a really good point about 1x accounts.  I think you are right - restricting to asset accounts only may be too restrictive.  Let me open this up a bit.

jniles:
I have two approvals, but I changed a few more files since these approvals landed.  @mbayopanda can you take another look for me?

client/src/modules/templates/bhAccountSelect.tmpl.html
```diff
@@ -1,4 +1,4 @@
-<div ng-form="AccountForm" bh-account-select ng-model-options="{ updateOn: 'default' }">
+<div ng-form="{{ $ctrl.name }}" bh-account-select ng-model-options="{ updateOn: 'default' }">
```
mbayopanda, line 1: 👍

mbayopanda (approved):
The behavior expected is working fine. LGTM 👍

PR #4792, opened 2020-08-05
Prevent selection of expired lots by the user in Stock-Exit (except for Loss)

### LINKED ISSUE #4749: Allow only stock exit for expired lots to loss
This issue is about to forbid all distribution of expired lots to patients, services, and depots.

Allow only expired lots to be distributed as a loss.

jmcameron:
For Stock-Exit to Patient, Service, or Depot, once the user has selected the inventory item, the drop-down menu to select the lot/batch will not show expired lots.  We also have a secondary check in the main submit function in the hopefully improbable case that expired lots come in via Invoices(Patient) or Requisitions(Service/Depot).

Suggested testing procedure using Bhima_test
 - ~Update the expiration date for Stock Lots for Quinine to make sure that one of the lots is not expired~.  <br>[This is no longer necessary due to the automatic updating of bhima_test data in PR https://github.com/IMA-WorldHealth/bhima/pull/4812]
 - Do Stock/Stock Exit, select Patient, Depot, or Service
 - Select the stock item for Quinine
 - Verify that only the unexpired lots are present in the Lots/Batches drop-down menu
 - Try the same procedure with Loss, the drop-down should show all the available lots including the expired ones.

Verify the French/English translations for the new error message:  ER_EXPIRED_STOCK_LOTS
This error message will be displayed as an error message (in red) at the top of the Stock Exit page when expired lots are found.

Note: There is no easy way to test for the error detection in the main submit button.  However if you modify `client/src/modules/stock/exit/exit.js` line 242 to be a copy of line 240 (temporarily in your local test sandbox), you should be able to test this by selecting an expired item (eg, in Patient) and then trying to Submit. You should see this error message in red at the top of the screen and the submit should fail.

### DIFF
--- a/client/src/modules/stock/exit/exit.js
+++ b/client/src/modules/stock/exit/exit.js
@@ -236,7 +236,11 @@ function StockExitController(
       dateTo : vm.movement.date,
     })
       .then(lots => {
-        item.lots = lots.filter(lot => !vm.selectedLots.includes(lot.uuid));
+        if (vm.movement.exit_type === 'loss') {
+          item.lots = lots.filter(lot => !vm.selectedLots.includes(lot.uuid));
+        } else {
+          item.lots = lots.filter(lot => !vm.selectedLots.includes(lot.uuid) && lot.lifetime > 0);
+        }
       })
       .catch(Notify.handleError);
   }
@@ -418,8 +422,21 @@ function StockExitController(
     delete vm.selectedEntityUuid;
   }
 
+  function expiredLots() {
+    return vm.selectedLots.some(lotId => vm.stockForm.store.data[0].lots
+      .some(lot => lot.uuid === lotId && lot.lifetime < 0));
+  }
+
   function submit(form) {
     if (form.$invalid) { return null; }
+
+    if (vm.movement.exit_type !== 'loss' && expiredLots()) {
+      // NOTE: This check may not be necessary, since the user cannot select
+      //       expired lots/batches directly.  But lots can also come in via
+      //       Invoices(Patient) or Requisions(Service/Depot), so it seems
+      //       prudent to check again here.
+      return Notify.danger('ERRORS.ER_EXPIRED_STOCK_LOTS');
+    }
 
     if (!vm.movement.entity.uuid && vm.movement.entity.type !== 'loss') {
       return Notify.danger('ERRORS.ER_NO_STOCK_DESTINATION');
--- a/client/src/i18n/en/errors.json
+++ b/client/src/i18n/en/errors.json
@@ -18,6 +18,7 @@
     "OVERPAID_INVOICE" : "You have overpaid the invoices.",
     "ER_DUP_KEY" : "A key collided in a unique database field.  Please retry your action.  If the problem persists, contact the developers.",
     "ER_DUP_ENTRY" : "You have duplicated a value in a unique field!  Please make sure that the necessary values are unique.",
+    "ER_EXPIRED_STOCK_LOTS" : "Some stock lots are expired",
     "ER_BAD_FIELD_ERROR" : "Column does not exist in database.",
     "ER_ROW_IS_REFERENCED" : "Cannot delete entity because entity is used in another table.",
     "ER_ROW_IS_REFERENCED_2" : "Cannot delete entity because entity is used in another table.",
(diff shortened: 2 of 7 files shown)


client/src/modules/stock/exit/exit.js
```diff
@@ -418,9 +423,28 @@ function StockExitController(
     delete vm.selectedEntityUuid;
   }
 
+  function expiredLots() {
+      for (const lotId of vm.selectedLots) {
```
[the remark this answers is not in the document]
jmcameron: What does hound want here?

client/src/modules/stock/exit/exit.js
```diff
@@ -236,7 +236,12 @@ function StockExitController(
       dateTo : vm.movement.date,
     })
       .then(lots => {
-        item.lots = lots.filter(lot => !vm.selectedLots.includes(lot.uuid));
+        if (vm.movement.exit_type == 'loss') {
+          item.lots = lots.filter(lot => !vm.selectedLots.includes(lot.uuid));
+        }
```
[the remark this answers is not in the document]
jmcameron: This warning is obscure (to me!)

client/src/modules/stock/exit/exit.js
```diff
@@ -418,9 +422,22 @@ function StockExitController(
...
+  function expiredLots() {
+    return vm.selectedLots.some(lotId => vm.stockForm.store.data[0].lots
+      .some(lot => lot.uuid === lotId && lot.lifetime < 0));
+  }
+
   function submit(form) {
     if (form.$invalid) { return null; }
 
+    if (vm.movement.exit_type !== 'loss' && expiredLots()) {
+      // NOTE: This check may not be necessary, since the user cannot select
```
jniles, line 434: 👍

client/src/modules/stock/exit/exit.js
```diff
@@ -418,9 +422,22 @@ function StockExitController(
     delete vm.selectedEntityUuid;
   }
 
+  function expiredLots() {
+    return vm.selectedLots.some(lotId => vm.stockForm.store.data[0].lots
+      .some(lot => lot.uuid === lotId && lot.lifetime < 0));
```
jniles, line 427: This is surprisingly pretty for O(n^2) time 😄

jniles (commented):
This is really good.  It looks like the tests are using expired lots which you've now blocked.

I wonder if there is a way to use [DATE_ADD()](https://www.w3schools.com/SQl/func_mysql_date_add.asp) or something to set the dates of stock expiration in the [default dataset](https://github.com/IMA-WorldHealth/bhima/blob/9cb4fba9f384b4f3bb9faf1fa2f7ca81426d4a15/test/data.sql#L2989).

Do you think you could tackle it?  It may also result in failing tests all over ... but I suppose those probably should be rethought or updated as well.

client/src/modules/stock/exit/exit.js
jmcameron: That sounds good.  But I'd like to push that to a new issue.  Let's go ahead and close this one.

client/src/modules/stock/exit/exit.js
jmcameron: I'm not convinced that the two for loops would be any slower.  It did not look like the prototype for .some() was particularly optimized.

jniles (approved):
So, this is good and I approve of the work done here.

The tests aren't passing yet, and we basically have two options:

1. Make the tests pass by updating the tests for patients, services, etc, to use unexpired items.
2. Update the items to use future expiration dates using DATE_ADD() and other tricks.

1 is easiest, but 2 is better on the long run.  It just may be too big for this PR.

jmcameron:
I'm pausing briefly on this pull request to implement https://github.com/IMA-WorldHealth/bhima/issues/4796 to update the expiration dates in the Bhima_test database so all the stock is not expired (no matter when the tests are run).  Once I have that done, this PR should be simpler to complete.

jniles:
@jmcameron thanks for getting the last pull request done on expiration dates.  Let's see if that has any effect on this PR.

jmcameron:
@jniles I think this one is ready to go!
I did a rebase to clean up the commits, which did not go smoothly (due to merges, the stock registry PR 4814 merge, etc).   I think I fixed up everything and the automated tests pass, so I think it is okay.   

But I will ask Bruce @mbayopanda  to double-check that I did not overwrite any of his changes.

mbayopanda (approved):
LGTM 👍 

The exception for stock loss is working so good

jniles (approved):
@jmcameron great work.  I know these controllers can be a bit hairy to work with.  I appreciate your efforts here and updating the tests is huge.

bors r=mbayopanda,jniles

PR #4799, opened 2020-08-07
fix(inventory): reverse COGS & asset accounts

### LINKED ISSUE #4798: Test database stock/expense accounts are backwards.
In our test database, we have our stock and expense accounts backwards.  See below:
image

The stock asset account is the 3* accounts and the stock expense accounts should be the 6* accounts.

jniles:
Fixes the test database by reversing the cost-of-goods-sold (COGS) expense accounts and the asset accounts.  Here is the fixed version:

image

Closes #4798.

### DIFF
--- a/test/data.sql
+++ b/test/data.sql
@@ -394,7 +394,7 @@ INSERT INTO cashbox_permission (user_id, cashbox_id) VALUES
   (@superUser, 2),
   (2, 1);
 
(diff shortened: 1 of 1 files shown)


mbayopanda (approved):
LGTM 👍

jniles:
@mbayopanda I need to update the test cases before this passes :)

PR #4801, opened 2020-08-07
fix: use correct CSP formulation in helmetjs

### LINKED ISSUE #4800: Fix CSP errors on fonts.gstatic.com
We hit a lot of CSP errors for our fonts:
image

I thought I had fixed this, but CSP is confusing in that sometimes it wants things wrapped as strings and sometimes it doesn't.  In this case, I think I got it backwards.

Here is the [MDN web docs link](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) that I've read three times and failed to actually implement properly...

jniles:
Uses the correct CSP rule form for fonts in helmetjs's configuration. This eliminates the CSP errors from the console for fonts.

Closes #4800.

### DIFF
--- a/server/config/express.js
+++ b/server/config/express.js
@@ -44,7 +44,7 @@ exports.configure = function configure(app) {
     contentSecurityPolicy : {
       directives : {
         defaultSrc : ['\'self\'', '\'unsafe-inline\'', 'blob:'],
-        fontSrc : ['\'self\'', '\'https://fonts.gstatic.com\''],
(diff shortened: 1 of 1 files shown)


jmcameron (approved):
Just tested this on my Windows 10 / WSL2 setup using the Windows Chrome browser (not one running inside the WSL).  It got rid of the font complaints.  I found other complaints but they they were related to Angular and not related to this issue.

PR #4810, opened 2020-08-11
feat(component) stock expired

### LINKED ISSUE #4785: Add a "expired stock" list next to the "out of stock" on stock exit.
Here is the "out of stock" component designed by @jeremielodi at work:

image

We need something nearly identical to this for "expired stock".  This is important to understand what stock is non-viable in your current depot so you can find it and get rid of it.  Basically, we would put it right next to this component, in nearly an identical layout.  It would list the inventory items and lot numbers of items that are expired, their expiration date, and the number of days they have been expired.

jeremielodi:
closes #4785


jeremielodi:
exit

client/src/js/components/bhStockExpired/bhStockExpired.js
```diff
@@ -0,0 +1,53 @@
...
+
+  $ctrl.$onChanges = () => {
+    expred();
+  };
+
+  /**
+   * @function expred
+   * get stock out inventories for a depot
+   */
+  function expred() {
```
jmcameron: 'exred()' ?  is this a typo?

jmcameron (changes requested):
I tested this on the IMCK data and it looked good.   I did not verify it was actually correctly listing only expired items.

However, when I tested it on the Bhima_test data, it did not report there were any expired stock items.  However, all the stock items (currently) start as expired -- and it did not list any.  When I edited one stock item to make it NOT expired, it showed up in the stock expired panel.  So I think there is a logic problem somewhere and the listing is inverted (shows NOT expired, not expired).

client/src/js/components/bhStockExpired/bhStockExpired.js
jeremielodi: maybe because I setted ```is_expired``` to true instead of ```1```

jmcameron:
@jeremielodi It still does not look right to me (with bhima_test).  I have these suggestions:

- Sync with master so that the expiration (lifetime) of the lots will be consistent.
- In the list of stock items that are expired, include the lot name (eg, QUININE-A).  
- It is not clear in the output of whether ALL quinine is expired or only a particular lot
- Check the end-to-end and integration tests and update as necessary.

Thanks!

jeremielodi:
The component I created makes a simple request for getting expired inventories, if the request composition is good that means we have never been able to retrieve expired inventories by using ```is_expired``` parameter.

jniles:
@jeremielodi I think the idea is to have a working component.  If that route doesn't work can you either

1. Confirm it doesn't work and open an issue OR
2. Fix it if it is a really quick fix and ensure that this component work correctly?

jeremielodi:
exit

jeremielodi:
About the component, I think it works perfectly. I'm trying to see if there is an issue on core.js or test data. But this PR is all set.

jmcameron (changes requested):
I just re-tested this with bhima_test and it seems to work correctly (not sure if I did my previous tests correctly).   However, I still suggest the changes I mentioned above:

- Sync with master so that the expiration (lifetime) of the lots will be consistent.
- In the list of stock items that are expired, include the lot name (eg, QUININE-A).
- Verify that after fresh 'yarn build:db', Stock Exits lists LOT Quinine-A expired. This should be the only expired item.
- Check the end-to-end and integration tests and update as necessary.

client/src/js/components/bhStockExpired/bhStockExpired.js
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * @function expred
+   * get stock out inventories for a depot
+   */
+  function expired() {
+    const dateTo = $ctrl.date || new Date();
+    if (!$ctrl.depotUuid) return;
+    $ctrl.loading = true;
+    Stock.inventories.read(null, {
+      is_expired : 1,
```
jniles, line 50: It might be good to filter out inventory lots with 0 quantities in stock as well.  We don't need to know a lot of expired if it has already been consumed!

image

client/src/js/components/bhStockExpired/bhStockExpired.js
```diff
@@ -0,0 +1,53 @@
+angular.module('bhima.components')
+  .component('bhStockExpired', {
+    templateUrl : 'js/components/bhStockExpired/bhStockExpired.html',
+    controller  : bhStockExpiredController,
+    transclude  : true,
+    bindings    : {
+      depotUuid : '<',
+      date : '<',
```
jniles: Is the date required?  If not, can you mark it as `<?`?

client/src/js/components/bhStockExpired/bhStockExpired.html
```diff
@@ -0,0 +1,49 @@
...
+      <span class="text-danger" 
+        ng-show="($ctrl.expiredInventories.length - 5) > 0"
+        translate-values="{ left : ($ctrl.expiredInventories.length - 5)}"
+        translate="{{ 'STOCK.AND_MORE' }}">
+      </span>
+
+    <a ui-sref= "stockInventories({ filters : [
+      { key : 'period', value : 'allTime'},
+      { key : 'includeEmptyLot', value : 1  },
+      { key : 'depot_uuid', value : $ctrl.depot.uuid, displayValue: $ctrl.depot.text, cacheable:false },
```
jniles, line 41: On the Vanga dataset, this didn't work for me.  When I click the link, the `stockInventories` registry isn't filtered on depot.  Can you check that this works on master? It may also affect the other component.

jniles: Now that I think about it, it actually is probably more useful to show the _lots_ that are expired, not the inventory items.  We want the user to be able to remove the lots that are expired from their stock.  So I think you should make this link go to the lots registry not the articles in stock registry.

This may have been what was giving @jmcameron confusion.

client/src/js/components/bhStockExpired/bhStockExpired.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+  });
+
+bhStockExpiredController.$inject = ['StockService', 'moment', 'NotifyService'];
+
+/**
+ * service or depot selection component
```
jniles: This isn't the service or depot selection component.  Can you update these comments, please?

client/src/js/components/bhStockExpired/bhStockExpired.html
```diff
@@ -0,0 +1,49 @@
...
+    </th>
+  </tr>
+</thead>
+<tbody>
+  <tr>
+    <th translate>FORM.LABELS.LOT</th>
+    <th translate>FORM.LABELS.INVENTORY</th>
+  </tr>
+  <tr ng-repeat = "inventory in $ctrl.expiredInventories | limitTo : 5">
+    <td>{{inventory.code}}</td><td>{{inventory.text}} ({{inventory.stock_out_date}})</td>
```
jniles: Can you make this expiration date human readable?   Here is what it currently looks like:

image

client/src/js/components/bhStockExpired/bhStockExpired.html
```diff
@@ -0,0 +1,49 @@
+<div class="row">
+  <bh-grid-loading-indicator
+  loading-state="$ctrl.loading"
+  error-state="$ctrl.hasError">
```
jniles: Is this ever set?  `hasError`?

jniles (commented):
@jeremielodi I've tested this out and made some comments below.

You also included a giant migration file.  Please remove it before a merge.

client/src/js/components/bhStockExpired/bhStockExpired.js
jeremielodi: okay

client/src/js/components/bhStockExpired/bhStockExpired.html
jeremielodi: perfect

jmcameron (changes requested):
Jérémie,
I tested this with Bhima_test and it looks good.  I still suggest two changes:

- It appears to be using the CODE name for the lot item instead of the name of the lot item itself.   In the case of Bhima_test, that should be "QUININE-A".   
- This is is a suggestion:  I wonder if it would be better to put the lot name on the right and the inventory name on the left?

client/src/js/components/bhStockExpired/bhStockExpired.html
```diff
@@ -0,0 +1,49 @@
...
+    <td colspan="2">
+      <span class="text-danger" 
+        ng-show="($ctrl.expiredInventories.length - 5) > 0"
+        translate-values="{ left : ($ctrl.expiredInventories.length - 5)}"
+        translate="{{ 'STOCK.AND_MORE' }}">
+      </span>
+
+    <a ui-sref= "stockLots({ filters : [
+      { key : 'period', value : 'allTime'},
+      { key : 'includeEmptyLot', value : 1  },
```
jniles: Let's make this `includeEmptyLot` to be `0`.  We don't want to see lots that have expired if they do not have stock in them.

client/src/js/components/bhStockExpired/bhStockExpired.js
```diff
@@ -0,0 +1,65 @@
...
+   * get expired inventories for a depot
+   */
+  function expired() {
+    const dateTo = $ctrl.date || new Date();
+    if (!$ctrl.depotUuid) return;
+    $ctrl.loading = true;
+    Stock.inventories.read(null, {
+      is_expired : 1,
+      depot_uuid : $ctrl.depotUuid,
+      includeEmptyLot : 0,
```
jniles, line 52: Excellent!

client/src/modules/stock/exit/exit.html
```diff
@@ -87,12 +87,18 @@
           </div>
         </div>
 
-        <div class="col-xs-12 col-md-6 panel-default">
+        <div class="col-xs-6 col-md-3 panel-default">
```
jniles, line 90: Cool 👍

client/src/js/components/bhStockExpired/bhStockExpired.html
```diff
@@ -0,0 +1,50 @@
...
+        ng-show="($ctrl.expiredInventories.length - 5) > 0"
+        translate-values="{ left : ($ctrl.expiredInventories.length - 5)}"
+        translate="{{ 'STOCK.AND_MORE' }}">
+      </span>
+
+    <a ui-sref= "stockLots({ filters : [
+      { key : 'period', value : 'allTime'},
+      { key : 'includeEmptyLot', value : 0  },
+      { key : 'depot_uuid', value : $ctrl.depot.uuid, displayValue: $ctrl.depot.text, cacheable:false },
+      { key : 'includeEmptyLot', value : 0 },
```
jniles, line 42: Minor nitpick - looks like you have `includeEmptyLot` twice here.

client/src/css/structure.css
```diff
@@ -996,3 +996,7 @@ a[disabled] {
   line-height: 100px;
   margin-bottom: 5px;
 }
+
+.lot-width {
+  width: 30%;
+}
```
jmcameron, line 1002: Thanks!  I was not sure where this should go.

jmcameron (approved):
Nice job, Jérémie.  I like the formatting.   An very small thing which you can ignore for now and tweak the next time we edit this function:  In the Inventory, desrciption, you've embedded the expiration info in parentheses.  Nice.  But you might insert: 'expired:' before it like this:    inventory, details, etc   (expired: 3 months ago).

jmcameron:
@jniles : this looks good to me.  Not sure if you want to take another look.  It seems like you have some unresolved comments.

jniles (approved):
@jmcameron I agree.  Thanks @jeremielodi and @jmcameron for your back and forth on this.  It looks good and works well.  Anything dangling we can put in other issues.

bors r+

PR #4814, opened 2020-08-12
Patch for v1.14.0 about stock movement registry

### LINKED ISSUE #4813: Stock Movements Unit has disappeared in the latest upgrade at both IMCK and HEV
We are missing the stock movements registry.  It might be worth making a hotfix patch for it.

mbayopanda:
Close #4813


jniles (approved):
This works for me!

bors r+

PR #4815, opened 2020-08-14
chore: document stock vars in comments

### LINKED ISSUE #4721: Expiration column in "Lots in Stock" gives the date differences in English
The "expiration" column always displays the date differences in English.   See below:

image

It should be like the other columns that show the differences in French.

jniles:
This PR adds comments to document in English on the stock lots calculations.  It also fixes a translation issue on the lots registry.

Closes #4721.

### DIFF
--- a/client/src/modules/stock/lots/registry.js
+++ b/client/src/modules/stock/lots/registry.js
@@ -138,7 +138,7 @@ function StockLotsController(
 
         lots.forEach((lot) => {
           const delay = moment(new Date(lot.expiration_date)).diff(current);
-          lot.delay_expiration = moment.duration(delay).humanize();
+          lot.delay_expiration = moment.duration(delay).humanize(true);
         });
 
         // FIXME(@jniles): we should do this ordering on the server via an ORDER BY
--- a/client/src/modules/stock/lots/registry.js
+++ b/client/src/modules/stock/lots/registry.js
@@ -4,6 +4,7 @@ angular.module('bhima.controllers')
 StockLotsController.$inject = [
   'StockService', 'NotifyService', 'uiGridConstants', 'StockModalService', 'LanguageService', 'GridGroupingService',
   'GridStateService', 'GridColumnService', '$state', '$httpParamSerializer', 'BarcodeService', 'LotsRegistryService',
+  'moment',
 ];
 
 /**
@@ -12,7 +13,7 @@ StockLotsController.$inject = [
  */
 function StockLotsController(
   Stock, Notify, uiGridConstants, Modal, Languages, Grouping,
-  GridState, Columns, $state, $httpParamSerializer, Barcode, LotsRegistry,
+  GridState, Columns, $state, $httpParamSerializer, Barcode, LotsRegistry, moment,
 ) {
   const vm = this;
   const cacheKey = 'lot-grid';
@@ -133,6 +134,12 @@ function StockLotsController(
 
     Stock.lots.read(null, filters)
       .then((lots) => {
+        const current = new Date();
+
+        lots.forEach((lot) => {
+          const delay = moment(new Date(lot.expiration_date)).diff(current);
+          lot.delay_expiration = moment.duration(delay).humanize();
+        });
 
         // FIXME(@jniles): we should do this ordering on the server via an ORDER BY
         lots.sort(LotsRegistry.orderByDepot);
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -393,13 +393,18 @@ function getLotsOrigins(depotUuid, params) {
  * DEFINITIONS:
  *   S_SEC: Security Stock - one month of stock on hand based on the average consumption.
  *   S_MIN: Minimum stock - twice the security stock.
- *
  *   S_RP: Risk of Expiration.
  */
 function stockManagementProcess(inventories) {
   let CM;
   let Q;
   let CM_NOT_ZERO;
+
+  // @const the minimum months of stock to keep on hand as security stock
+  // This affects the calculation of the minimum stock
+  // NOTE(@jniles) - should this be moved into some database variable or
+  // global constant?
+  const MIN_MONTHS_OF_SECURITY_STOCK = 2;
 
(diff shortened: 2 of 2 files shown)


server/controllers/stock/core.js
```diff
@@ -402,14 +408,38 @@ function stockManagementProcess(inventories) {
     Q = inventory.quantity; // the quantity
     CM = inventory.avg_consumption; // consommation mensuelle
     CM_NOT_ZERO = !CM ? 1 : CM;
+
```
jmcameron, line 421: Thanks for adding these comments!   I was looking at these lines yesterday and thought that we should beef up the comments!   Great addition!

server/controllers/stock/core.js
```diff
@@ -402,14 +408,38 @@ function stockManagementProcess(inventories) {
     Q = inventory.quantity; // the quantity
```
mbayopanda, line 413: We can also mention that this quantity is the available quantity of inventory for a given depot, the implementation is in : `/server/controllers/stock/core.js`

server/controllers/stock/core.js
```diff
@@ -402,14 +408,38 @@ function stockManagementProcess(inventories) {
...
     CM = inventory.avg_consumption; // consommation mensuelle
```
mbayopanda, line 419: The `avg_consumption` is the CMM or AMC which is calculated during stock consumption (stock exit to patient or to service, or to a depot if the depot is a warehouse), implementation is in `/server/controllers/stock/index.js`

mbayopanda (approved):
LGTM 👍 

I just add some comment about the quantity and the avg_consumption.

server/controllers/stock/core.js
jniles: Great idea.  Thank you!

jniles:
I've updated the PR with comments from @mbayopanda.  Thanks!

client/src/modules/stock/lots/registry.js
```diff
@@ -133,6 +134,12 @@ function StockLotsController(
 
     Stock.lots.read(null, filters)
       .then((lots) => {
+        const current = new Date();
+
+        lots.forEach((lot) => {
+          const delay = moment(new Date(lot.expiration_date)).diff(current);
+          lot.delay_expiration = moment.duration(delay).humanize();
+        });
```
jmcameron, line 143: Somehow this is losing the sign of the expiration when the lot is expired.  For expired items, in English, we should append "ago", and in French prepend "il y a"  (or something similar).  Or instead saying 'un mois' leave it as a digit and prepend a negative sign?  See below.

jmcameron: image

jmcameron: The translation of 'year' / 'ans', 'month' / 'mois' looks good.

server/controllers/stock/core.js
```diff
@@ -389,27 +389,62 @@ function getLotsOrigins(depotUuid, params) {
...
+    // Compute the Security Stock
+    // Security stock is defined by taking the average monthly consumption (AMC or CMM)
+    // and multiplying it by the Lead Time (inventory.delay).  The Lead Time is by default 1 month.
+    // This gives you a security stock quantity.
     inventory.S_SEC = CM * inventory.delay; // stock de securite
+
+    // Compute Minimum Stock
+    // The minumum of stock required is double the security stock.
     inventory.S_MIN = inventory.S_SEC * 2; // stock minimum
+
```
jmcameron, line 431: We should probably make that '2' a parameter.  And should it be possible to assign different values to different inventory items?

server/controllers/stock/core.js
```diff
@@ -389,27 +389,62 @@ function getLotsOrigins(depotUuid, params) {
...
+    // Compute the Risk of Stock Expiry
+    // The risk of stock expiry is the quantity of drugs that will expire before you
+    // can use them.
     inventory.S_RP = inventory.quantity - (inventory.lifetime * CM); // risque peremption
 
+    // compute the inventory status code
     if (Q <= 0) {
       inventory.status = 'stock_out';
       inventory.stock_out_date = inventory.last_movement_date;
```
jmcameron, line 455: I see that this section has been expanded in the current code to include other status values ('security_reached', 'minimum_reached', etc).   I wonder if these should be Bhima constants somewhere?

server/controllers/stock/core.js
```diff
@@ -389,27 +389,62 @@ function getLotsOrigins(depotUuid, params) {
...
     inventory.S_MONTH = Math.floor(inventory.quantity / CM_NOT_ZERO); // mois de stock
+
+    // Compute the Refill Quantity
+    // The refill quantity is the amount of stock needed to order to reach your maximum stock.
     inventory.S_Q = inventory.S_MAX - inventory.quantity; // Commande d'approvisionnement
     inventory.S_Q = inventory.S_Q > 0 ? parseInt(inventory.S_Q, 10) : 0;
+
+    // Compute the Risk of Stock Expiry
+    // The risk of stock expiry is the quantity of drugs that will expire before you
+    // can use them.
```
jmcameron, line 449: Thanks for adding the comments!

jmcameron (approved):
Once you fix the missing negative on the expiration of expired lots, this is ready to go.  The other comments are minor or could be handled  in a later issue.

server/controllers/stock/core.js
jniles: Yeah, this is probably true.  I'll make an issue.

client/src/modules/stock/lots/registry.js
jniles: I've added a direction ("in two months") and ("two months ago").  

image

What is your view on that presentation?

Also, I noticed that we have some bugs in one of our production databases where expired stock isn't showing up in red.  I'm not sure why that is the case, but I'll have to investigate.

jmcameron (approved):
It looks great!    Don't forget to add an issue for making 'stock_out', etc, into BHIMA constants (https://github.com/IMA-WorldHealth/bhima/blob/master/server/controllers/stock/core.js#L414).

PR #4821, opened 2020-08-17
fix collection capacity name to recovery capacity

### LINKED ISSUE #4809: Choose either the name "collection capacity" and "recovery capacity"
We have a recovery capacity report in BHIMA, but it is called multiple things.  In the Tree, it is "recovery capacity", but on the document itself.  See below:

image

We should choose one name, and stick to it.  This is probably a problem with multiple reports we have, since we have two names in the names for each report - one in the "reports" table and the "tree" table.

mbayopanda:
This PR is related to #4809 

Close #4809


jniles (approved):
This looks great to me!  If tests pass, we can merge it straight away.

jmcameron:
bors +r

PR #4836, opened 2020-08-20
chore: update test inventory group expiration flag

### LINKED ISSUE #4828: Set tracking_expiration and tracking_consumption sensibly for default data
We have some PNAM data as default data in our dataset.  We should set the default options for `tracking_expiration` and `tracking_consumption` appropriately to allow us to have nicer test datasets.

For example, we have a lot of equipment that isn't consumable and should not expire.  See below:
image

jniles:
Updates the default/test datasets inventory group expiration tracking flags.  The material's expiration dates are no longer tracked.

Closes #4828.


jmcameron (approved):
The changes look fine to me.   I assume the reasoning is that there are not any expiration dates for things like hard equipment.   However, I did notice that the item on line 429 (some type of water) will not expire either.  Even water generally has some kind of shelf-life due to the container, rust, etc.

jmcameron:
Also, the test pass.   I noticed that only data was changed, not tests.   What tests will use this change in the data?

jmcameron:
I retract the comment about Bhima_data (I was testing it incorrectly).   

I see that the changed tracking_expire flags are set per the updated test data SQL file.

However, I do not see any change in the display of the inventory configuration page, since the "Tracking Expiration" is not shown in the main display, even though it is visible when editing the inventory group.   Should column(s) be added for the two flags?

Also, why does  "3167969 Eau, Hygiène et Assainissement" not display Sale, Stock, and Cost?  Is this tap water?

lomamech (approved):
I just tried it out by adding a new inventory in a group that does not have expiration dates

PR #4839, opened 2020-08-21
Improve stock lot highlighting in Stock Lots registry

### LINKED ISSUE #4802: Nearly expired Stock should show up as yellow in the Stock Lots management
Like in Issue https://github.com/IMA-WorldHealth/bhima/issues/4730 (PR https://github.com/IMA-WorldHealth/bhima/pull/4739), we would like stock that is near to expiration to show in yellow in the stock lots management page (Stock > Stock Lots).  Exactly what to use as the time frame needs to be determined.

**NOTE:** Issue https://github.com/IMA-WorldHealth/bhima/issues/4344 should be resolved first since the determination of whether the stock is about to expire depends on the average consumption (CMM).   Currently CMM seems to be zero (at least in bhima_test).

jmcameron:
Updated stock lot highlighting.   No changes to underlying lot status logic.

**My goals are:**

- Clearly mark lots that are expired.   This is an important issue for stock management.
- Mark lots that are stock-out.  Note that the Stock Lots registry hides these unless the user selects an option in the "Search" filter.
- Mark lots that are at risk of expiration so they can be used first.  
   - NOTE: We should consider how to inform the user about this when they select lots!  Perhaps the lot selection drop-down menu could be highlighted?
- Mark lots that are "at risk" (about to run out or expire).   

The currently supported highlight colors are (all for the specific lot):https://github.com/jmcameron/bhima/blob/at_risk_highlighting/client/src/modules/stock/lots/templates/row.expired.html

- **Red** (css class=expired): When the stock lot quantity is not 0 but it is expired.
- **Purple** (css class=out-of-stock): When the stock lot quantity is 0 (stock out)
- **Orange** (css class=at-risk-of-expiring): When the stock lot quantity is not 0, is not expired, but flag in the "At risk of expiration" column  is true.  (I have not been able to test this because I have not found a case where this is true.  There may be a logic issue with this.)
- **Yellow** (css class=at-risk): When the stock lot quantity is not 0, is not expired, but will run out or expire soon (the value in the "Risk" column is <= 0).

The logic for these classes is in: https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/stock/lots/templates/row.expired.html

Although I have attempted to make these mutually exclusive, I would not be surprised if there are logical inconsistencies.    It is quite possible that these are not mutually exclusive cases.   In that case, we need to figure out how to handle it best and update the class selection logic appropriately.     It is not clear to me why we need three columns for Risk-related items.  They do not seem to me to tell a very clear story.

I welcome suggestions on color scheme, how to improve this, etc.

Also renamed translation label  'IS_IN_OF_RISK_EXPIRATION' to 'IS_IN_RISK_OF_EXPIRATION' for readability.

**Suggestions for testing**

- Use the IMCK database.  
- Go to the Stock > Stock Lots and observe the colors of the rows and how they correspond to the descriptions above.
- In the "Search" filter, scroll to the bottom and click on the "Included Exausted  Lots" at the very end.  Click [Submit] on the pop-up Search filter dialog.  You should see stock-out items in purple.

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -881,10 +881,21 @@ a[disabled] {
 }
 
 .expired {
-  /* text-decoration : line-through; */
-  background-color: #ffb3b3 !important;
+  background-color: #ff000083 !important;
 }
 
+.at-risk {
+  background-color: #ffff6686 !important; 
+}
+
+.at-risk-of-expiring {
+  background-color: #ff880191 !important; 
+}
+
+.out-of-stock {
+  /* text-decoration : line-through; */
+  background-color: #cc00ff8f !important; 
+}
 
 .dropdown-menu {
   z-index : 9999;
--- a/client/src/js/services/LotService.js
+++ b/client/src/js/services/LotService.js
@@ -19,5 +19,46 @@ function LotService(Api, $http, util) {
       .then(util.unwrapHttpResponse);
   };
 
+  /**
+  * @function computeLotWarningFlags()
+  *  @description
+  *   Takes in a Lot and attaches warning flags to the Lot.
+  *
+  *   The logic goes like this, in this order.
+  *   NOTE: Once a case is found to be true, all following cases are ignored.
+  *     1. If the stock is exhausted, warn about that.
+  *        (Recall that the user can choose to display exhausted lots in Lots Registry.)
+  *     2. If the Lot is expired, warn about that.
+  *     3. If the Lot is near expiration, warn about that.
+  *        NOTE that this assumes the CMM and that stock exits all come from this
+  *        lot exclusively.  But the CMM is based on not only on this lot, but on
+  *        an aggregate all lots of this inventory item, so there is no guarantee
+  *        that this will be correct.
+  *     4. If a Lot is at risk of running out, warn about that. Again this is
+  *        based on the aggregate CMM which may not work out exactly for this
+  *        Lot in practice.
+  *
+  * Based on this logic, only one of the warning flags should be set to true.
+  */
+  lots.computeLotWarningFlags = (lot) => {
+    lots.exhausted = false;
+    lots.expired = false;
+    lots.near_expiration = false;
+    lots.at_risk = false;
+    if (lot.quantity <= 0) {
+      lot.exhaused = true;
+    } else if (lot.lifetime < 0) {
+      // Equivalent to: lot.quantity > 0 && lot.lifetime < 0
+      lot.expired = true;
+    } else if (lot.IS_IN_RISK_EXPIRATION) {
+      // Equivalent to: lot.quantity > 0 && lot.lifetime >= 0 && lot.IS_IN_RISK_EXPIRATION
+      lot.near_expiration = true;
+    } else if (lot.S_RISK <= 0) {
+      // Equivalent to: lot.quantity > 0 && lot.lifetime >= 0 && lot.S_RISK <= 0
+      lot.at_risk = true;
+    }
+
+  };
+
   return lots;
 }
--- a/client/src/js/services/LotService.js
+++ b/client/src/js/services/LotService.js
@@ -46,7 +46,7 @@ function LotService(Api, $http, util) {
     lots.near_expiration = false;
     lots.at_risk = false;
     if (lot.quantity <= 0) {
-      lot.exhaused = true;
+      lot.exhausted = true;
     } else if (lot.lifetime < 0) {
       // Equivalent to: lot.quantity > 0 && lot.lifetime < 0
       lot.expired = true;
--- a/client/src/modules/stock/lots/templates/row.expired.html
+++ b/client/src/modules/stock/lots/templates/row.expired.html
@@ -4,7 +4,9 @@
   ng-repeat="(colRenderIndex, col) in colContainer.renderedColumns track by col.uid"
   ui-grid-one-bind-id-grid="rowRenderIndex + '-' + col.uid + '-cell'"
   class="ui-grid-cell"
-  ng-class="{ 'ui-grid-row-header-cell': col.isRowHeader, 'expired': row.entity.lifetime < 0 }"
+  ng-class="{ 'ui-grid-row-header-cell': col.isRowHeader, 'at-risk-of-expiring': row.entity.near_expiration,
+              'expired': row.entity.expired, 'at-risk': row.entity.at_risk, 'out-of-stock': row.entity.exhausted,
+            }"
   data-vals="{{col.isRowHeader}}"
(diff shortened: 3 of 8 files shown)


client/src/modules/stock/lots/templates/row.expired.html
```diff
@@ -4,7 +4,12 @@
   ng-repeat="(colRenderIndex, col) in colContainer.renderedColumns track by col.uid"
   ui-grid-one-bind-id-grid="rowRenderIndex + '-' + col.uid + '-cell'"
   class="ui-grid-cell"
-  ng-class="{ 'ui-grid-row-header-cell': col.isRowHeader, 'expired': row.entity.lifetime < 0 }"
+  ng-class="{ 'ui-grid-row-header-cell': col.isRowHeader, 
+              'at-risk-of-expiring': row.entity.quantity > 0 && row.entity.lifetime >= 0 && row.entity.IS_IN_RISK_EXPIRATION,
```
jniles: This is excellent, but the logic is hidden, and we probably need to use this logic elsewhere as well.  I propose moving this logic into a service's function, so we can document it.  You've explained in detail your decisions in your PR text, but it would be nice to have that inline for future devs.

Here's what I propose:

1. Let's map each of these states to constants in `bhConstants`.  There's already [a few](https://github.com/IMA-WorldHealth/bhima/blob/bc7af348a0fdca0a83156717cabbdf814bb798b8/client/src/js/constants/bhConstants.js#L98).  I'm going to call each of these states "flags".
2. Move this logic into the `StockLotsService`.  Call the function something like `LotsService.computeLotFlags()`.  This way, we'll be able to clearly document the logic of how we label the lots.  It will like like something like this:

```js
/**
* @function computeLotFlags()
*  @description
*   Takes in a lot and attaches a flag to the log in the property "flag".  The logic goes like this....
*/
service.computeLotFlags = function computeLotFlags(lot) {
  // pseudo-code.  The details will be filled in later.
  lot.flag = getFlagFromSwitchOrIfElseDecisionTree();
};
```

3. Now we've attached a `bhConstant` to the lot in the `flag` property.  We can next just change this logic to use those constants rather than the conditions documented here.
```html
<div
  ...
  ng-class="{ 'at-risk-of-expiring' : row.entity.lot.flag === grid.appScope.AT_RISK_OF_EXPIRING, ... }"
  >
```

The goal is to make sure that when someone wants to change the logic, they don't have to touch this template.  Also, we can test the service function under different conditions and trust that this just displays the right thing.

jniles: Finally, that function will be portable to the server and we can figure out where the best place is to display it.

client/src/css/structure.css
```diff
@@ -881,10 +881,21 @@ a[disabled] {
...
+  background-color: #ffff6686 !important; 
 }
 
+.at-risk-of-expiring {
+  background-color: #ff880191 !important; 
+}
+
+.out-of-stock {
+  /* text-decoration : line-through; */
+  background-color: #cc00ff8f !important;
```
jniles, line 897: I learned something today.  I never knew that there could be four hexadecimal numbers.  I had to look it up in the MDN Web Docs to try and figure out why four and not three.

[Source for anyone else who is confused](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#Syntax_2):

> Hexadecimal notation: #RRGGBB[AA]
    R (red), G (green), B (blue), and A (alpha) are hexadecimal characters (0–9, A–F). A is optional. For example, #ff0000 is equivalent to #ff0000ff.

client/src/css/structure.css
```diff
@@ -881,10 +881,21 @@ a[disabled] {
 }
 
 .expired {
-  /* text-decoration : line-through; */
-  background-color: #ffb3b3 !important;
+  background-color: #ff000083 !important;
+}
+
+.at-risk {
```
jniles, line 887: I find it slightly confusing the distinction between `at-risk` and `at-risk-of-expiration`.  It seems like there is a decision matrix here:


  | will run out of stock soon | will expire soon
-- | -- | --
**will run out of stock soon** | at-risk | ??
**will expire soon** | ?? | at-risk-of-expiry

It's clearly out of the scope of this PR, but we should make an issue to determine the correct logic flow and maybe name some of these variables more explicitly.

jniles (commented):
This looks good to me.  I've added some comments about moving the logic to a place where it is more portable.  Let me know what your thoughts are on that suggestion.

I'll make an issue about a unified colorscheme - it seems to me that we should pick something and go with it for all of our registries and reports that deal with stock expiry or stock out.

client/src/modules/stock/lots/templates/row.expired.html
jmcameron: Thanks for the suggestions.  I had wanted to use a function but had not quite figured it out this way.   When I figured out this short cut in the template I just went with it..    I'm glad to do it the way you suggest.

client/src/css/structure.css
jmcameron: Yes, it is confusing but lot variables for those are being computed around line 788 of https://github.com/IMA-WorldHealth/bhima/blob/master/server/controllers/stock/core.js#L788.   Once we have this approach and function set up, it will be simple to update once we clarify/improve the "risk" calculations.

client/src/modules/stock/lots/templates/row.expired.html
jmcameron: I implemented these changes.  I did go with the same logic I had before, so that will probably need to be addressed in the future.   In the lots registry, we may want display the lots with their parent inventory items by grouping lots under their respective inventory item and applying some highlighting to the parent inventory item and some to the child lots, as appropriate.

jniles (commented):
This looks largely good to me.  I can't get the purple to show up though.  I believe this lot should be purple, correct?

image

I'm using the test dataset.  I've also tried with the Vanga database.  Can you check to see what is going on here?

client/src/js/services/LotService.js
```diff
@@ -19,5 +19,46 @@ function LotService(Api, $http, util) {
...
+  *
+  * Based on this logic, only one of the warning flags should be set to true.
+  */
+  lots.computeLotWarningFlags = (lot) => {
+    lots.exhausted = false;
+    lots.expired = false;
+    lots.near_expiration = false;
+    lots.at_risk = false;
+    if (lot.quantity <= 0) {
+      lot.exhaused = true;
```
jniles: Typo - this should be `lot.exhausted`.  This would be why the purple isn't showing up 😆

jmcameron: Good catch!  I fixed this typo and now the purple highlighting works.

jniles (approved):
This works for me now.  Great work!

image

bors r+

PR #4857, opened 2020-08-26
Swapped inventory/lot labels on expiration warning block

jmcameron:
In the warning block for expired stock (in the Stock Lots registry) from PR https://github.com/IMA-WorldHealth/bhima/pull/4810, 
we missed that the label for Inventory vs Lot needed to be swapped.  This PR fixes that.

### DIFF
--- a/client/src/js/components/bhStockExpired/bhStockExpired.html
+++ b/client/src/js/components/bhStockExpired/bhStockExpired.html
@@ -19,8 +19,8 @@ class="table table-condensed table-bordered">
 </thead>
 <tbody>
   <tr>
-    <th translate>FORM.LABELS.LOT</th>
     <th translate>FORM.LABELS.INVENTORY</th>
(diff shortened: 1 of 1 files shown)


jmcameron:
I think the current implementation of the expired warning block is a bit odd.   If I have two lots of an inventory item that are expired and one that still has stock, the expired warning block does not catch that.  It only lists one expired lot.  However, when you click on "See all" you see both expired lots.   Since this warning block is for Stock Exits, it seems like we should be warning about expired "Lots", not expired "products".   And for each inventory, we can show a list of the expired lots (on the right) instead of the current one.  Also the count of expired "Lots" should be computed based on the total number of lots expired in the whole warning block. 

NOTE: Converted this to Issue https://github.com/IMA-WorldHealth/bhima/issues/4863

jmcameron:
@jeremielodi This is ready for a review now.

jeremielodi (approved):

PR #4865, opened 2020-08-28
Alert about expiration in stock registry

### LINKED ISSUE #4854: Add alert about lot expiration in the "Article en stock" registry
This issue is about to add : 
- A column with alert icon about lot expiration
- A click on the icon show a popup with more details about lots expirations of the inventory

The popup can contain : 
- Lots with their duration until the expiration (and quantity remaining)

mbayopanda:
This issue is related to #4854 

Add icon with tooltip for lots status in the "Article en stock" registry

image

Close #4854

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -205,7 +205,8 @@
     "CLICK_TO_SEE":"Click here to see lots",
     "SHOW_WARNINGS":"Click to show Warnings",
     "NO_RISK_OF_EXPIRATION":"No lots with expiration risk",
-    "RISK_OF_EXPIRATION":"{{quantity}} {{unit}} of the lot {{label}}, expiration : {{expiration}}"
+    "RISK_OF_EXPIRATION":"{{quantity}} {{unit}} of the lot {{label}}, expiration : {{expiration}}",
+    "SHOW_ONLY_RISKY":"Show only inventories with risky lots"
   },
   "MOVEMENTS" : {
     "REGISTRY" : "Movements Registry"
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -207,7 +207,7 @@
     "NO_RISK_OF_EXPIRATION":"No lots with expiration risk",
     "RISK_OF_EXPIRATION":"There are {{quantity}} {{unit}} at risk of expiration",
     "EXPIRED":"There are {{quantity}} {{unit}} expired",
-    "NEAR_EXPIRE":"There are {{quantity}} {{unit}} near to expire",
+    "NEAR_EXPIRE":"There are {{quantity}} {{unit}} expiring soon",
     "SHOW_ONLY_RISKY_LOTS":"Show only inventories with risky lots"
   },
   "MOVEMENTS" : {
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -201,7 +201,11 @@
     "ASSIGNMENT_HISTORIC":"Assignment Historic",
     "ASSIGNMENT_CURRENT":"Current",
     "NO_ASSIGNMENT":"No Assignment",
-    "INCLUDE_EXHAUSTED_LOTS":"Include Exhausted Lots"
+    "INCLUDE_EXHAUSTED_LOTS":"Include Exhausted Lots",
+    "CLICK_TO_SEE":"Click here to see lots",
+    "SHOW_WARNINGS":"Click to show Warnings",
+    "NO_RISK_OF_EXPIRATION":"No lots with expiration risk",
+    "RISK_OF_EXPIRATION":"{{quantity}} {{unit}} of the lot {{label}}, expiration : {{expiration}}"
   },
   "MOVEMENTS" : {
     "REGISTRY" : "Movements Registry"
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -205,8 +205,8 @@
     "CLICK_TO_SEE":"Click here to see lots",
     "SHOW_WARNINGS":"Click to show Warnings",
     "NO_RISK_OF_EXPIRATION":"No lots with expiration risk",
-    "RISK_OF_EXPIRATION":"{{quantity}} {{unit}} of the lot {{label}}, expiration : {{expiration}}",
-    "SHOW_ONLY_RISKY":"Show only inventories with risky lots"
+    "RISK_OF_EXPIRATION":"Expiration:{{expiration}} of {{quantity}} {{unit}}",
+    "SHOW_ONLY_RISKY_LOTS":"Show only inventories with risky lots"
   },
   "MOVEMENTS" : {
     "REGISTRY" : "Movements Registry"
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -205,7 +205,9 @@
     "CLICK_TO_SEE":"Click here to see lots",
     "SHOW_WARNINGS":"Click to show Warnings",
     "NO_RISK_OF_EXPIRATION":"No lots with expiration risk",
-    "RISK_OF_EXPIRATION":"Expiration:{{expiration}} of {{quantity}} {{unit}}",
+    "RISK_OF_EXPIRATION":"There are {{quantity}} {{unit}} at risk of expiration",
+    "EXPIRED":"There are {{quantity}} {{unit}} expired",
+    "NEAR_EXPIRE":"There are {{quantity}} {{unit}} near to expire",
     "SHOW_ONLY_RISKY_LOTS":"Show only inventories with risky lots"
   },
   "MOVEMENTS" : {
--- a/client/src/modules/stock/inventories/registry.js
+++ b/client/src/modules/stock/inventories/registry.js
@@ -6,6 +6,7 @@ StockInventoriesController.$inject = [
   'uiGridConstants', 'StockModalService', 'LanguageService', 'SessionService',
   'GridGroupingService', 'bhConstants', 'GridStateService',
   '$state', 'GridColumnService', '$httpParamSerializer', 'BarcodeService',
+  'moment',
 ];
 
 /**
@@ -15,7 +16,7 @@ StockInventoriesController.$inject = [
 function StockInventoriesController(
   Stock, Notify, uiGridConstants, Modal, Languages,
   Session, Grouping, bhConstants, GridState, $state, Columns,
-  $httpParamSerializer, Barcode,
+  $httpParamSerializer, Barcode, moment,
 ) {
   const vm = this;
   const cacheKey = 'stock-inventory-grid';
@@ -41,6 +42,12 @@ function StockInventoriesController(
     field            : 'group_name',
     displayName      : 'STOCK.INVENTORY_GROUP',
     headerCellFilter : 'translate',
+  }, {
+    field            : 'warning',
+    displayName      : 'STOCK.LOTS',
+    headerCellFilter : 'translate',
+    cellTemplate     : 'modules/stock/inventories/templates/warning.cell.html',
+    width            : 50,
   }, {
     field            : 'quantity',
     displayName      : 'STOCK.QUANTITY',
@@ -135,6 +142,29 @@ function StockInventoriesController(
   vm.gridApi = {};
   vm.saveGridState = state.saveGridState;
   vm.getQueryString = Stock.getQueryString;
+
+  vm.toggleTooltip = row => {
+    if (row.tooltipIsOpen) {
+      row.tooltipIsOpen = !row.tooltipIsOpen;
+      return;
+    }
+
+    Stock.lots.read(null, {
+      inventory_uuid : row.inventory_uuid,
+      depot_uuid : row.depot_uuid,
+    })
+      .then(lots => {
+        row.lots = lots
+          .map(item => {
+            const delay = moment(new Date(item.expiration_date)).diff(new Date());
+            item.delay_expiration = moment.duration(delay).humanize(true);
+            return item;
+          })
+          .filter(item => item.quantity > 0 && item.IS_IN_RISK_EXPIRATION === true);
+        row.tooltipIsOpen = true;
+      })
+      .catch(Notify.handleError);
+  };
 
   function clearGridState() {
     state.clearGridState();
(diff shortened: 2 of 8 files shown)


jmcameron (approved):
I tried this on the IMCK database and it works fine.  I skimmed through the code and it also seemed okay (given my limited experience).  

I think that this concept is a good one and it seems to work fine.  However I have a couple of reservations:

- Do we really need a separate "Lots" column for this flag?  This warning icon could be added to the "Status" column before  the "In stock", etc, colored tokens.
- Using "True" in the "Lots" field is something that will probably not be obvious to typical users.  Not sure what to do about that.

Beyond the scope of this PR is the fact that we need a more consistent, unified way of warning about problem conditions.  In the "Stock Lots Registry" we have gone to coloring whole lot rows based on status (based on suggestions from @jmcameron and @jniles.  Here we are using combinations of Status tokens as well as the new warning icon in this PR.   This should be sorted out after this PR is completed.

mbayopanda:
Thank you @jmcameron for your review, you are right the filter this kind of column with "true" is so weird but I don't have yet an idea about how to enhance that.

Also having lot alert in it own column was the first Idea I had, and its benefits are : 
- We can filter quickly in a column
- 😄 also it was easy to handle the popover 

But we can always enhance that

jmcameron:
>     Thank you @jmcameron for your review, you are right the filter this kind of column 
>     with "true" is so weird but I don't have yet an idea about how to enhance that.
 Since all we have is True/False, we could just use a small button with a question mark '?' on it.  When clicked it would should show items with warnings (like typing "True" in the current filter.  When in this mode the question mark '?' could change to a small red X.  This button would only be  a character or two wide.  Not sure how easy this is to fold into the extra cell in that column when filtering is enabled.

>     * We can filter quickly in a column
There is no filter on the "Status" column, so one could be added for this

>     * 😄 also it was easy to handle the popover
It should be possible to move the popover to an HTML span/div containing the warning icon so that hovering over it does the tooltip popup like it currently does (which is nice).

client/src/i18n/en/stock.json
```diff
@@ -201,7 +201,12 @@
     "ASSIGNMENT_HISTORIC":"Assignment Historic",
     "ASSIGNMENT_CURRENT":"Current",
     "NO_ASSIGNMENT":"No Assignment",
-    "INCLUDE_EXHAUSTED_LOTS":"Include Exhausted Lots"
+    "INCLUDE_EXHAUSTED_LOTS":"Include Exhausted Lots",
+    "CLICK_TO_SEE":"Click here to see lots",
+    "SHOW_WARNINGS":"Click to show Warnings",
+    "NO_RISK_OF_EXPIRATION":"No lots with expiration risk",
+    "RISK_OF_EXPIRATION":"{{quantity}} {{unit}} of the lot {{label}}, expiration : {{expiration}}",
```
jniles: This is cool, but I actually think it probably gives us too much information for the context.  If the user needs this much context, we'll just ask them to look it up in the stock lots registry.

I propose that we change this to be:  "There are {{quantity}} {{unit}} expired in stock."  This will make it less busy.

client/src/modules/stock/inventories/templates/warning.cell.html
```diff
@@ -0,0 +1,44 @@
+<div ng-if="!row.groupHeader" style="cursor: pointer;" ng-click="grid.appScope.toggleTooltip(row.entity)" translate-attr="{'title': 'LOTS.SHOW_WARNINGS' }" class="ui-grid-cell-contents text-center" translate>
+  <span 
+    uib-popover-template="'warningLotsMessage.html'" 
+    popover-trigger="none" 
+    popover-is-open="row.entity.tooltipIsOpen" 
+    popover-append-to-body="true"
```
jniles: Instead of a popover, can we use either a title (like normal HTML) or a [tooltip](https://angular-ui.github.io/bootstrap/#!#tooltip)?  Personally, I like that behavior a lot better.  It will probably also have better performance since angular doesn't have to think about the HTML in it.

client/src/modules/stock/inventories/templates/warning.cell.html
```diff
@@ -0,0 +1,44 @@
+<div ng-if="!row.groupHeader" style="cursor: pointer;" ng-click="grid.appScope.toggleTooltip(row.entity)" translate-attr="{'title': 'LOTS.SHOW_WARNINGS' }" class="ui-grid-cell-contents text-center" translate>
```
jniles: `translate-attr` is great! 👍

client/src/modules/stock/inventories/registry.js
```diff
@@ -136,6 +142,29 @@ function StockInventoriesController(
...
+      depot_uuid : row.depot_uuid,
+    })
+      .then(lots => {
+        row.lots = lots
+          .map(item => {
+            const delay = moment(new Date(item.expiration_date)).diff(new Date());
+            item.delay_expiration = moment.duration(delay).humanize(true);
+            return item;
+          })
+          .filter(item => item.quantity > 0 && item.IS_IN_RISK_EXPIRATION === true);
```
jniles: You could probably switch this `.map()` and `.filter()` statement for a slight performance increase.

jniles (commented):
@mbayopanda, this looks _awesome_.  I can right away see the inventory items that have problematic lots and that is fantastic.

I think the best user experience (and the easiest to maintain) is to make the user go to the lots registry to find out information about the lots.  Here, we'll just let them know that there are problematic lots and directly them to click "view lots in stock" to see the details.

I agree with @jmcameron that the filtering is a bit strange with the word "True".  I wouldn't use an inline filter for this at all - I would add something to the search modal.  But that is a minor point.

client/src/i18n/en/stock.json
```diff
@@ -201,7 +201,12 @@
...
+    "SHOW_ONLY_RISKY":"Show only inventories with risky lots"
```
jmcameron: I suggest changing this to:   Show only inventories with lots at risk
and change the token 'SHOW_ONLY_RISKY" to "SHOW_ONLY_RISKY_LOTS"

jmcameron (approved):
I tried your updates to the risky column and they look good to me.  I made a small suggestion about an English translation.  I think once you've addressed @jniles concerns, this is ready to go.

client/src/i18n/en/stock.json
mbayopanda: You right @jniles there are a lot of information, but some lots can be expired and others not yet expired

client/src/modules/stock/inventories/templates/warning.cell.html
mbayopanda: Since we have already a title, the proposal is to keep just an icon for lots status and remove the popover?

jniles: Nah, I like the idea of a tooltip.  It should be faster than the popover.

client/src/i18n/en/stock.json
mbayopanda: What if we have `Expiration:{{expiration}} of {{quantity}} {{unit}}`

jniles: The expiration date is tied to the lot number, right?  What do you show if more than one lot is expired?  Is `{{expiration}}` just the most recent expiration date?

mbayopanda: It is used in a loop to show details of the first three risky lots : 
image

jniles: Gotcha.  Okay, no problem.

jniles:
@mbayopanda as we've discussed in person, I think we should probably has something like this for the expiration states

color | icon | Interpretation / Text
:-- |:-- | --
.text-success | .fa-check | No items at risk of expiration
.text-warning | .fa-exclamation-circle | Items at risk of expiration (will not be consumed before expiring)
.text-danger | .fa-exclamation-triangle | Items expired

You don't have to do all these components in this particular pull request as we'll want to probably update this registry to use the `stock_movement_status` table down the line.

jmcameron:
> we should probably has something like this for the expiration states

I suggest calling the CSS classes
   .icon-success
   .icon-warning
   .icon-danger

jmcameron (approved):
I tested this PR again.   A few issues I noticed:

- When you hover over a danger icon, it says "Click to show warnings".  However, if you click it, the warning pop-up comes up but there is no way simple way to dismiss it (other than clicking on "Show the Lots" link or refreshing the page.  Normally, clicking elsewhere will dismiss such a popup. :
image

- In the figure above (in English), notice that the "See the lots" has not been translated.

- I see that you changed the translation token SHOW_ONLY_RISKY to SHOW_ONLY_RISKY_LOTS, per my suggestion. Thanks!  But I see that there is still a reference to SHOW_ONLY_RISKY in the StockFilterer service.  In the figure above, you can see that the filter label has not gotten translated (in purple, it says "LOTS.SHOW_ONLY_RISKY" instead of being translated).

mbayopanda:
@jniles changes are landed

client/src/modules/stock/inventories/templates/warning.cell.html
```diff
@@ -0,0 +1,46 @@
...
+    <i class="fa fa-warning text-warning"></i> 
+    <span class="text-warning" 
+      translate="LOTS.RISK_OF_EXPIRATION"
+      translate-values="{ quantity : row.entity.riskyLotsQuantity, unit : row.entity.unit_type }">
+    </span>
+  </div>
+
+  <!-- expired lots -->
+  <div ng-if="row.entity.nearExpireLotsQuantity">
+    <i class="fa fa-warning text-warning"></i>
```
jniles: I commented on the PR about the `.text-warning` etc, labels, but that was only for the icon class.  It makes it hard to read for the text.  Let's just make this plain text with no icon.

client/src/modules/stock/inventories/templates/warning.cell.html
```diff
@@ -0,0 +1,46 @@
+<div ng-if="!row.groupHeader" class="ui-grid-cell-contents text-center" translate>
+  <span 
+    uib-tooltip-template="'warningLotsMessage.html'" 
+    tooltip-append-to-body="true"
```
jniles, line 4: Can we put the tooltip position to the right as well?

client/src/modules/stock/inventories/templates/warning.cell.html
```diff
@@ -0,0 +1,46 @@
...
+    translate>
```
jniles: You don't need this `translate` directive.

client/src/i18n/en/stock.json
```diff
@@ -201,7 +201,14 @@
...
     "ASSIGNMENT_CURRENT":"Current",
     "NO_ASSIGNMENT":"No Assignment",
-    "INCLUDE_EXHAUSTED_LOTS":"Include Exhausted Lots"
+    "INCLUDE_EXHAUSTED_LOTS":"Include Exhausted Lots",
+    "CLICK_TO_SEE":"Click here to see lots",
+    "SHOW_WARNINGS":"Click to show Warnings",
+    "NO_RISK_OF_EXPIRATION":"No lots with expiration risk",
+    "RISK_OF_EXPIRATION":"There are {{quantity}} {{unit}} at risk of expiration",
+    "EXPIRED":"There are {{quantity}} {{unit}} expired",
+    "NEAR_EXPIRE":"There are {{quantity}} {{unit}} near to expire",
```
jniles: "Near" in English implies that something is close in _distance_.  To talk say that something is close in _time_, you would use the world "soon".

This text is better written:  "There are {{quantity}} {{unit}} expiring soon".

server/controllers/stock/index.js
```diff
@@ -622,15 +622,67 @@ async function listLotsDepot(req, res, next) {
...
+      let hasRiskyLots = false;
+      let hasExpiredLots = false;
+      let hasNearExpireLots = false;
+
+      let riskyLotsQuantity = 0;
+      let expiredLotsQuantity = 0;
+      let nearExpireLotsQuantity = 0;
+
+      for (let j = 0; j < lots.length; j++) {
+        const hasSameDepot = lots[j].depot_uuid === inventories[i].depot_uuid;
```
jniles, line 648: I like these variables 👍

server/controllers/stock/index.js
```diff
@@ -622,15 +622,67 @@ async function listLotsDepot(req, res, next) {
...
   const enableDailyConsumption = req.session.enterprise.settings.enable_daily_consumption;
 
-  core.getInventoryQuantityAndConsumption(params, monthAverageConsumption, enableDailyConsumption)
-    .then((rows) => res.status(200).json(rows))
-    .catch(next)
-    .done();
+  try {
+    const inventoriesParameters = [params, monthAverageConsumption, enableDailyConsumption];
+    const inventories = await core.getInventoryQuantityAndConsumption(...inventoriesParameters);
+    const lots = await core.getLotsDepot(null, params);
```
jniles: You could fire these as the same time:
```js
const [inventories, lots] = await Promise.all([
  core.getInventoryQuantityAndConsumption(),
  core.getLotsDepot(),
]);
````

File: server/controllers/stock/index.js [hunk 12]
```diff
@@ -622,15 +622,67 @@ async function listLotsDepot(req, res, next) {
...
+      let riskyLotsQuantity = 0;
+      let expiredLotsQuantity = 0;
+      let nearExpireLotsQuantity = 0;
+
+      for (let j = 0; j < lots.length; j++) {
+        const hasSameDepot = lots[j].depot_uuid === inventories[i].depot_uuid;
+        const hasSameInventory = lots[j].inventory_uuid === inventories[i].inventory_uuid;
+        if (hasSameDepot && hasSameInventory) {
+          const lot = lots[j];
+          if (lot.quantity <= 0) {
```
jniles, line 652: Is this equivalent to @jmcameron's code on the client?  If so, can you make it into a function of the same name here in the file?  That would make it easier to update/merge.

jniles (commented):
@mbayopanda looks good!  I've made a few comments, but this is pretty much ready to go.  Let me know when you've take a look at them.

server/controllers/stock/index.js
mbayopanda: It is almost the same, the difference is here we have the calculation of quantities

image

jniles (approved):
Changes look good to me.  I really like this feature.

The server side code will need to be retouched in time (it should probably use the `stock_movement_status` table), but this is good for now.  

Here is what it looks like for those curious:
O5Olvm2Nv9

Appveyor is giving you trouble, but I don't think it will block this.  It looks like a fluke and I can't re-trigger the build.


bors r+

PR #4871, opened 2020-08-31
Implemented enterprise and depot-specific min months security stock parameter

### LINKED ISSUE #4860: Add ability to set the "safety stock" per depot, rather than per medication
Our safety stock is currently calculated based on a hard-coded variable:
https://github.com/IMA-WorldHealth/bhima/blob/1bf6400fd99f7997aca0f221635a6f218becb1c9/server/controllers/stock/core.js#L412-L416

However, to account for differences between the health centers (which need a single month of stock on hand) and the health zone offices (which need four), we'll want to set the minimum number of months of stock by depot.

jmcameron:
This PR:

- Adds a default Minimum Months of Security Stock ("Safety Stock") to the Enterprise settings
- Adds a minimum months of security stock parameter for depots
- Updates CMM/stock security calculations to use the depot-specific Months of Security Stock parameter
- Adds a 2 stock exits for patients to Bhima_test so that CMM can be calculated

Suggested testing procedure (MMSS = Min Months of Security Stock):
  - Use bhima_test
  - Do:  yarn build:db   (there are new stock exits)
  - Go to Stock > Depot Management
    - Edit each depot and verify that MMSS is:
      - Depot Principal: 2
      - Depot Secondaire: 3
    - Start to create a depot
      - Notice that MMSS defaults to 2
      - Enter name of depot, delete the value of MMSS, try to submit. 
        It should fail with warning that MMSS is required.
      - Exit without creating the depot
  - Go to Administration > Enterprise settings
    - Change default MMSS to 3
    - Select another page
    - Edit enterprise settings again, and verify default MMSS changed to 3
  - Go back to Stock > Depot Management
    - Create a depot, notice that MMSS dfaults to 3
    - Change the MMSS to 1
    - Complete creating the depot
    - Edit it again to verify the MMSS is 1
  - Verify that the MMSS value is being used in stock calculations
    - Go to Stock > Report > Inventories, click Preview
      - Recall that MMSS for the Principal depot defaults to 2
      - Note that the MIN column (near right end) are:
        - Quinine: 20
        - Vitamins: 10
    - Edit the Principal Depot, change MMSS to 1
    - Go to Stock > Report > Inventories, click Preview
      - Note that the MIN column (near right end) are changed:
        - Quinine: 10
        - Vitamins: 5

### DIFF
--- a/client/src/modules/depots/modals/depot.modal.html
+++ b/client/src/modules/depots/modals/depot.modal.html
@@ -119,6 +119,18 @@
       </div>
     </div>
 
+    <div class="form-group">
+      <label class="control-label" translate>
+        FORM.LABELS.MIN_MONTHS_SECURITY_STOCK
+      </label>
+      <input name="min_months_security_stock" class="form-control" ng-model="DepotModalCtrl.depot.min_months_security_stock"
+        type="number" min="1" autocomplete="off" required bh-integer>
+      <div class="help-block form-warning" ng-show="DepotForm.$submitted"
+           ng-messages="DepotForm.min_months_security_stock.$error">
+        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      </div>
+    </div>
+
   </div>
 
   <div class="modal-footer">
--- a/client/src/modules/depots/modals/depot.modal.html
+++ b/client/src/modules/depots/modals/depot.modal.html
@@ -119,13 +119,13 @@
       </div>
     </div>
 
-    <div class="form-group">
+    <div class="form-group" ng-class="{ 'has-error' : DepotForm.$submitted && DepotForm.min_months_security_stock.$invalid }">
       <label class="control-label" translate>
         FORM.LABELS.MIN_MONTHS_SECURITY_STOCK
       </label>
       <input name="min_months_security_stock" class="form-control" ng-model="DepotModalCtrl.depot.min_months_security_stock"
         type="number" min="1" autocomplete="off" required bh-integer>
-      <div class="help-block form-warning" ng-show="DepotForm.$submitted"
+      <div class="help-block" ng-show="DepotForm.$submitted"
            ng-messages="DepotForm.min_months_security_stock.$error">
         <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
       </div>
--- a/client/src/modules/enterprises/enterprises.js
+++ b/client/src/modules/enterprises/enterprises.js
@@ -231,6 +231,9 @@ function EnterpriseController(Enterprises, util, Notify, Projects, Modal, Scroll
   vm.setMonthAverage = function setMonthAverage() {
     $touched = true;
   };
+  vm.setDefaultMinMonthsSecurityStock = function setDefaultMinMonthsSecurityStock() {
+    $touched = true;
+  };
 
   startup();
 }
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -226,6 +226,7 @@ function getLotsDepot(depotUuid, params, finalClause) {
       SUM(m.quantity * IF(m.is_exit = 1, -1, 1)) AS quantity,
       SUM(m.quantity) AS mvt_quantity,
       d.text AS depot_text, l.unit_cost, l.expiration_date,
+      d.min_months_security_stock,
       ROUND(DATEDIFF(l.expiration_date, CURRENT_DATE()) / 30.5) AS lifetime,
       BUID(l.inventory_uuid) AS inventory_uuid, BUID(l.origin_uuid) AS origin_uuid,
       i.code, i.text, BUID(m.depot_uuid) AS depot_uuid,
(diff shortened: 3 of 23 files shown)


jmcameron:
I'm having trouble fixing the Appveyor integration tests with this PR.  I added a couple of stock exits (to patients) so the CMMs would be calculated.  It is passing some of the configurations but not others:
image

However, if If fix the error for this test, 
image

it fails the first one!   Why would this be working differently on these systems.  The only thing I can think of is that in SQL I'm using DATE_FORMAT(), ADD_DATE(), and CURENT_DATE.  But the last two worked in previous MySQL updates I've done, so I don't think that could be the problem.

Suggestions?

server/controllers/stock/core.js
```diff
@@ -436,7 +432,7 @@ function stockManagementProcess(inventories) {
 
     // Compute Minimum Stock
     // The minumum of stock required is double the security stock.
-    inventory.S_MIN = inventory.S_SEC * MIN_MONTHS_OF_SECURITY_STOCK; // stock minimum
+    inventory.S_MIN = inventory.S_SEC * inventory.min_months_security_stock;
```
mbayopanda, line 435: 👍

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -119,6 +119,18 @@
...
+      <label class="control-label" translate>
+        FORM.LABELS.MIN_MONTHS_SECURITY_STOCK
+      </label>
+      <input name="min_months_security_stock" class="form-control" ng-model="DepotModalCtrl.depot.min_months_security_stock"
+        type="number" min="1" autocomplete="off" required bh-integer>
+      <div class="help-block form-warning" ng-show="DepotForm.$submitted"
+           ng-messages="DepotForm.min_months_security_stock.$error">
+        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      </div>
+    </div>
```
mbayopanda, line 132: 👍

client/src/modules/enterprises/enterprises.js
```diff
@@ -231,6 +231,9 @@ function EnterpriseController(Enterprises, util, Notify, Projects, Modal, Scroll
   vm.setMonthAverage = function setMonthAverage() {
     $touched = true;
   };
+  vm.setDefaultMinMonthsSecurityStock = function setDefaultMinMonthsSecurityStock() {
```
jniles, line 234: 👍

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -119,6 +119,18 @@
       </div>
     </div>
 
+    <div class="form-group">
```
jniles: Generally, the way we show/hide errors across the application is with 3 different steps.
 
 1. We add an `ng-class` directive to the `<div class="form-group">`.  This is the standard bootstrap (v3) way of doing things.  [Here is their docs](https://getbootstrap.com/docs/3.4/css/#forms-control-validation).  It generally looks like this (pulled from this modal, just above):
  ```html
    <div class="form-group" ng-class="{ 'has-error' : DepotForm.$submitted && DepotForm.text.$invalid }">
  ```
  What is this doing?  It's saying that you should only show the red "error" field when the form has been submitted and the particular `<input>` is also invalid.  (`DeportForm.text` is the `name` of that `<input>`).  In your case, it is `DepotForm.min_months_security_stock`.

2. Secondly, we add a "help block" dive underneath the form.  This is colored based on the above class.  If there is no class, it is grey.  [Here is the documentation](https://getbootstrap.com/docs/3.4/css/#forms-help-text).  Otherwise, it can be red, green, etc, depending on the state.  Here is an example from the same block  above:
```html
    <div class="help-block" ng-messages="DepotForm.text.$error" ng-show="DepotForm.$submitted">
    </div>
```
What's going on here?  Well, we are displaying a help block, only once the form is submitted.  This is the same idea as above.  We don't want to show any help blocks with error messages in them until the user has submitted the form and it gets blocked because of a constraint.

 3. The final component is using [ngMessages](https://docs.angularjs.org/api/ngMessages/directive/ngMessages) to show a text field based on the error condition.  This is what makes our text get translated on error.  I'll let you read about that in the documentation.

jniles: So, with all the context, I propose you make these changes to make it look like more standard BHIMA fare.

 1. Add `ng-class="{ 'has-error' : DepotForm.$submitted && DepotForm.min_months_security_stock.$invalid }"` to the `<div class="form-group">`.
 2. Remove the `form-warning` class from the `help-block`.

Here is what happens:

**Before**
image

**After**
image

server/models/migrations/next/migrate.sql
```diff
@@ -707,3 +707,11 @@ CREATE TABLE `lot_tag` (
   FOREIGN KEY (`lot_uuid`) REFERENCES `lot` (`uuid`),
   FOREIGN KEY (`tag_uuid`) REFERENCES `tags` (`uuid`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+/*
+ * @author: jmcameron
+ * @date: 2020-08-28
+ * @description: add min_months_security_stock to enterprise and depot tables
```
jniles, line 714: Beautiful 👍

jniles (commented):
This works great for me!  Awesome work!

I've made a suggestion and explanation about how we typically show errors to the client.  Could you make that change?

Afterwards, it looks fine to me.  The logic is sound.  I've tested it and it all works.

jmcameron:
@jniles   I think this is ready to go!

jniles (approved):
Looks good to me.

bors r+

PR #4875, opened 2020-09-01
fix(stock): skip tags query if empty result

### LINKED ISSUE #4873: Lot Tags breaks filtering on the Inventory Item in Stock Lots Registry
If you filter on an inventory item that doesn't have lots in stock, normally you get an empty lot registry.  It now appears to break after the lot_tags PR landed with this error:

image

Steps to reproduce:
  1. Navigate to `/#!/stock/lots`
  2. Search for an inventory item that doesn't have stock lots.
  3. Observe the error.

We should probably skip the tag query if we have no lots in stock.

jniles:
Prevents a crash if an empty set is returned from the stock routes.

Closes #4873.

### DIFF
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -532,24 +532,29 @@ async function listLotsDepot(req, res, next) {
     params.defaultPeriodEntry = params.defaultPeriod;
     delete params.defaultPeriod;
   }
+
   try {
     const data = await core.getLotsDepot(null, params);
 
     const queryTags = `
       SELECT BUID(t.uuid) uuid, t.name, t.color, BUID(lt.lot_uuid) lot_uuid
       FROM tags t
-      JOIN lot_tag lt ON lt.tag_uuid = t.uuid
+        JOIN lot_tag lt ON lt.tag_uuid = t.uuid
       WHERE lt.lot_uuid IN (?)
     `;
-    const lotUuids = data.map(row => db.bid(row.uuid));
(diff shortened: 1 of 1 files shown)


server/controllers/stock/index.js
```diff
@@ -532,24 +532,29 @@ async function listLotsDepot(req, res, next) {
     params.defaultPeriodEntry = params.defaultPeriod;
     delete params.defaultPeriod;
   }
+
   try {
     const data = await core.getLotsDepot(null, params);
 
     const queryTags = `
       SELECT BUID(t.uuid) uuid, t.name, t.color, BUID(lt.lot_uuid) lot_uuid
```
jmcameron, line 540: Should we be using the SQL keyword 'AS' to define column aliases here?  I know it is optional, but shouldn't we be consistent for readability?  I personally prefer adding the AS since it makes it more readable to me.

jmcameron (approved):

server/controllers/stock/index.js
jniles: I agree and  I generally try to.  Do you think it's worth trying to enforce such a style for PRs?

jmcameron: I personally think it is a good idea both for readability.   À mon avis!  However, we already enforce various Javascript niceties via hound.  If hound can sniff SQL and check this, then sure, I think announcing that and enforcing it via hound makes sense and is not too painful (I assume hound only checks changed files).   So that developers only need to change future behavior.

PR #486, opened 2016-06-13
Refactor exchange rates

jniles:
This commit continues to refactor exchange rates to 2.x code quality standards.  In particular, the following have been implemented:
1. The exchange rate service uses less complex branching logic since many functions return early if possible.
2. Unnecessary `id`s have been removed from rows and buttons.
3. The colors for buttons have been brought in line with 2.x colors.
4. The `bhLoadingButton` is used where appropriate.
5. Notifications are used to signal success and error states instead of `vm.feedback` state magic.

The core functionality and UI of the module have not really changed.  Only two significant UI changes are introduced:
1. The CREATE button is now in the header instead of a util bar below.
2. All glyphicons have been replaced with font awesome.

---

Thank you for contributing!

Before submitting this pull request, please verify that you have:
- [x] Run your code through JSHint.  [Check out our styleguide](https://github.com/IMA-WorldHealth/bhima-2.X/blob/master/docs/STYLEGUIDE.md).
- [x] Run integration tests.
- [x] Run end-to-end tests.
- [x] Accurately described the changes your are making in this pull request.

For a more detailed checklist, [see the official review checklist](https://docs.google.com/document/d/1nupLVLRXgSZJQo_acLgrwvPnN8RukfSiwRhSToj81uU/pub) that this PR will be evaluated against.

Ensuring that the above checkboxes are completed will help speed the review process and help build a stronger application.  Thanks!

PR #4881, opened 2020-09-02
feat(CMM aggregate data calculation) set in_quantity and out_quantiy when  start_date = end_date

jeremielodi:
closes #486


jniles (approved):
LGTM.  Thanks!

bors r+

PR #4882, opened 2020-09-02
fix(build): use .env for application config

### LINKED ISSUE #3146: Proposal: replace .env.* with .env and add it to .gitignore
In the `master` branch there is currently a `.env.development` file checked into the git repository. I proposed that we restructure our environment infrastructure so that we can remove this and follow the environment file best practices. 

### The problem: 
Our current development tasks rely specifically on `env.development`. Because this file is checked into git this file cannot be updated to your local environment variables without it being tracked as a code change. The purpose of environment files is to seperate the repository code from each environment, this includes development. 

### Proposal: 
* Remove `.env.development` from the version history - dotenv - [Should I commit my .env file?](https://github.com/motdotla/dotenv#should-i-commit-my-env-file)
* Remove any parsing of `.env.${environment}` from the server in favour of a single `.env` file - [dotenv - Should I have multiple .env files?](https://github.com/motdotla/dotenv#should-i-have-multiple-env-files) 
* Check in a `.env.sample` file into the repository demonstrating exactly how a file should be structured and what it should be contained 
* Update the 'Getting Started' and 'Installation' guides to clearly describe how to copy `.env.sample` into `.env` and customise it to fit your setup
* Update continuous integration services to correctly set the `.env` file according to their needs 
* Specify upgrade plan for current production servers using the `.env.production` method, this will most likely just involve copying `.env.production` to `.env`
* Bonus: Make sure the server understands if there's a missing environment file and gives a clear error with instructions on how to resolve this

jniles:
Removes the .env.* files and creates a single .env file that is _not_ checked in to git as recommended by 3146.  This makes it less likely that we will have faulty connections in production.

Closes #3146.

----

## Rationale
Previously, we had two files `.env.development` and `.env.production`.  In truth, you could have any `.env.*` files.  The idea was that based on the `NODE_ENV` variable, you could easily switch environments.

The problem with this was the same as the advantage - you could easily switch the environment away from a _production_ system to a _development_ system.  This happened more than I would like to admit in production.  By removing this contrast, it offloads the work onto individual developers to customize their environments as they wish (which is slightly more work) at the benefit of never accidentally screwing up a production configuration.

## Instructions for Migrating from Previous Versions
To migrate from a previous version, copy your current `.env.development` into an untracked file (such as `.env.development.bak`), pull in the latest changes, and rename the `.env.development.bak` to `.env`.  Here is a script:

```sh
#! /usr/bin/env bash

# Run this before merging these changes!
cp .env.development .env.development.bak

# merge the changes somehow (rebase, merge, ff etc)

mv .env.development.bak .env
```


jmcameron:
I suggest adding some operational instructions on how we should be handling .env for 'yarn dev' and 'yarn app'.  Thanks!

jniles:
Yeah, if you try to use it in it's current form, you'll have problems.  Once I get the tests to pass on CI, I'll send out some instructions.  The issue that it closes contains a lot of good reasoning for why we would want this.  Plus @mbayopanda and I realized today we had fixed the same production issue related to this about 3 times now and that is too many times.

jeremielodi:
Okay, I'm looking into this PR

jniles:
@jeremielodi thank you

jeremielodi (dismissed):
It looks good to me but I faced some errors on my windows.

jeremielodi:
errora

Did you change those commands too?

jniles:
Did you create the .env file as stated in the PR directions?

jeremielodi:
No, there is already .env.sample and .env.production. Should I create .env file?

jeremielodi:
I says ```do not break on missing .env file``` is .env required?

jniles:
Yes, as it says in the PR description:


## Instructions for Migrating from Previous Versions
To migrate from a previous version, copy your current `.env.development` into an untracked file (such as `.env.development.bak`), pull in the latest changes, and rename the `.env.development.bak` to `.env`.  Here is a script:

```sh
#! /usr/bin/env bash

# Run this before merging these changes!
cp .env.development .env.development.bak

# merge the changes somehow (rebase, merge, ff etc)

mv .env.development.bak .env
```

jniles:
The `.env` file is now the only `.env` file used by the application.  All others are removed.  The sample one is included for new contributors to learn from.

jeremielodi:
When this PR will be merged, are we going to follow all these steps to create the environment file?

jeremielodi:
When I am in your branch, I suppose that I'm in the master but I don't find the file ```.env.development```

jeremielodi:
If I have never installed bhima before, Am I going to have the default development file?

jniles:
@jeremielodi, yes to all your questions.

There isn't a "default" development file.  There's just a sample environment file to show what variables BHIMA will read from the environment.  Normally, we expect a new install to change these variables, since they should name their database whatever they want, use whatever user they want, make a password for themselves, etc.  They shouldn't really be using these defaults.

The PR forces that issue by removing the .env.development file.  This means that we'll never accidentally start Karawa, IMCK, HEV, pepkits, or CHK in "development" mode.  This is a good thing.

It is a bit more inconvenient during the install, but that is a one-time issue.

jeremielodi:
That make sens. Just a proposal, why don't you have just ```.env``` instead of  ```.env.sample``` then contributors will just edit one file.
You can merge if that is not a good idea.
Good work @jniles, thanks.

jniles:
@jeremielodi check out the issue this is closing for the reason not to commit the `.env` file.  Sfount gave a pretty good explanation:
https://github.com/IMA-WorldHealth/bhima/issues/3146

Here is also the linked rationale from the creator of `dotenv`: https://github.com/motdotla/dotenv#should-i-commit-my-env-file

jeremielodi:
Perfect!

jniles:
Alright, I think I finally got the environmental variables set properly on the appveyor interface.

bors r+

jmcameron (approved):
I tried this out and it worked fine for development.  I removed my old .env.production file so that all that is left is .env (and .env.sample).  I copied .env from my old .env.development file, per directions.   I was able to run 'yarn dev' and that works fine.  

I was also able to run 'yarn app' and that seemed to work too, which as mystifying to me.   It did invoke ... NODE_ENV=production, etc, which was good.  But I had no idea of what to do if 'dev' and 'app' needed different passwords, environment variables, etc.

In other words, I think more documentation is needed to say how to run bhima in production mode ('yarn app') and how to set up appropriate .env??? settings for that mode.

jniles:
Bors r+

jniles:
bors try

jniles:
For anyone wondering what is going on here, AppVeyor seems to be somewhat selective in how to specify environmental variables and the scope to which they apply.  So I'm just slamming it with different configurations to see what is going on.  The documentation isn't very clear.

jniles:
bors try

jniles:
Closing in favor of https://github.com/IMA-WorldHealth/bhima/pull/4909

jniles:
> 
> 
> I tried this out and it worked fine for development. I removed my old .env.production file so that all that is left is .env (and .env.sample). I copied .env from my old .env.development file, per directions. I was able to run 'yarn dev' and that works fine.
> 
> I was also able to run 'yarn app' and that seemed to work too, which as mystifying to me. It did invoke ... NODE_ENV=production, etc, which was good. But I had no idea of what to do if 'dev' and 'app' needed different passwords, environment variables, etc.
> 
> In other words, I think more documentation is needed to say how to run bhima in production mode ('yarn app') and how to set up appropriate .env??? settings for that mode.

Now that this is merged, switching between production/development doesn't actually have any impact on environmental variables besides the NODE_ENV variable.  This is important because it affects the way that some of the expressjs modules behave and also affects on if you are minifying your build or not.  But for things like the databases, these are no longer version controlled so you can run any database - production or development - in either build configuration.

What I do is put my databases all in my .env file commented out as so:
```sh
DB_NAME="bhima_test"
#DB_NAME="pepkits"
#DB_NAME="imck"
# etc...
```
And this lets me easy switch between them by commenting/uncommenting the database name.

jmcameron:
I upgraded to the latest master (with .env changes, etc.). Everything seems to work! But I have a suggestion: why not add multiple environment variables to the .env file and choose them based on what they are called. For example, in the .env file:

DB_NAME_DEV = 'bhima_test'
DB_NAME_IMCK = 'imck'
DB_NAME_VANGA = 'vanga'

etc.

Then 'yarn dev' would use _DEV versions, 'yarn app imck' (or 'yarn app DB = imck') could use 'DB_NAME_IMCK', etc.

Other environment variables could be done in the same way.

This way we could avoid editing our .env files just to run different databases.

jniles:
You can actually avoid editing the .env file by just passing in the database name as an environmental variable.  These take precedence over the `.env` file variables.  For example, in bash:
```sh
DB_NAME=imck yarn app
# or
DB_NAME=vanga yarn dev
# if you want to be fancy:
alias run-imck="DB_NAME=imck yarn app"
# etc..
```

In my view, this is _almost_ the same amount of work, but reduces confusion for developers who may or may not have access to certain production databases.

Given this, I feel like it would be more confusing than helpful, but I'm open to the idea!  What do you think?

PR #4891, opened 2020-09-05
Refactored handling of 'Include Exhausted Lots' flag in Stock Lots registry

### LINKED ISSUE #4862: Move "Include Exhausted Lots" to the "Default Filters" Category
The current default for "Include Exhausted Lots" is "No".  We treat it like a default filter.   Let's move it to the default filters category.

image

jmcameron:
Based on the discussion in Issue https://github.com/IMA-WorldHealth/bhima/issues/4862 I refactored how the 'Include Exhausted Lots' flag is handled:

- Changed the flag to a 3-value radio with very clear options (and no "clear" link)
- Moved the flag into the defaults tab for the Stock Lots registry and the Inventory in Stock registry.
- Added StockFilterExhaustedController to simplify using this flag.  Note that it includes some flexibility for the names of the options.   I thought we might possibly want different wording for the Items in stock vs lots registry.  If not we can remove that flexibility and hardcode the language tokens in the controller to make it even easier to use.
- Fixed the filtering so that it is possible to list ONLY exhausted lots (like the expired flag).
- Renamed language token STOCK.INCLUDE_EMPTY_LOT to STOCK.INCLUDE_EMPTY_LOTS for readability
- Added an exhausted lot in the Bhima_test data to simplify testing

Note that the flag now accepts 3 options:
image

**Testing:**

- Make sure to do yarn build:db
- Do the stock lots registry and observe the 3 options and test them.
- Try the same in the the inventory in stock listing too.


jmcameron:
@jniles @mbayopanda Please take a look at this update and let me know what you think.   

I added a stock in/stock out so I would have some out-of-stock meds for testing purposes.  The integration/end-to-end tests are broken  but I may have to wait until the MySQL 8 problem is fixed before fixing them.  I have the last test in stock.z3.inventory-adjustment-lots.spec.js mostly commented out because what it was doing made no sense to me (and something may be broken).    I will not release it until that is resolved but I thought you might as well look this over in case I need to make other fixes.

Also, how can I see something like the last test in stock.z3.inventory-adjustment-lots.spec.js tests in the normal GUI so I can diagnose what is going on.   Normally that last query returns 2 items but now it returns 19. which makes no sense to me.

client/src/js/components/bhStockFilterExhausted.js
```diff
@@ -0,0 +1,34 @@
+angular.module('bhima.components')
+  .component('bhStockFilterExhausted', {
+    bindings : {
+      value : '<',
+      label : '@',
+      name : '@',
+      excludeLabel : '@',
+      includeLabel : '@',
+      includeOnlyLabel : '@',
+      helpText : '@?',
```
mbayopanda, line 10: Is it possible to remove these labels : label, excludeLabel, includeLabel, includeOnlyLabel and helpText. And just insert these labels directly in the template file since these labels will not be changed often, I think.

client/src/i18n/en/stock.json
```diff
@@ -195,13 +195,15 @@
     "REGISTRY" : "Lots Registry",
     "ENABLE_FAST_INSERT": "Enable fast lots insertions",
     "ENABLE_FAST_INSERT_DESCRIPTION": "This option will automatically select the next blank row's serial/lot number in the entry table. If necessary, it adds a new row. Useful for quickly inserting with a barcode reader",
+    "EXCLUDE_EXHAUSTED_LOTS": "Exclude exausted lots",
```
jniles: Typo here- this should `exhausted` lots.

client/src/js/components/bhStockExpired/bhStockExpired.html
```diff
@@ -29,15 +29,14 @@
...
-      <span class="text-danger" 
+      <span class="text-danger"
         ng-show="($ctrl.expiredInventories.length - 5) > 0"
         translate-values="{ left : ($ctrl.expiredInventories.length - 5)}"
         translate="{{ 'STOCK.AND_MORE' }}">
       </span>
 
     <a ui-sref= "stockLots({ filters : [
       { key : 'period', value : 'allTime'},
-      { key : 'includeEmptyLot', value : 0  },
```
jniles, line 40: Excellent 👍

server/controllers/stock/core.js
```diff
@@ -16,6 +16,7 @@ const moment = require('moment');
 const db = require('../../lib/db');
 const FilterParser = require('../../lib/filter');
 const util = require('../../lib/util');
+// const { param } = require('jquery');  ??? Not sure why hound is complaining about this now
```
mbayopanda: You don't have to insert this line after local require `require('../../a/given/path')`

If you put it after `const moment = require('moment');` there will be no complain I think

client/src/modules/stock/lots/modals/search.modal.html
```diff
@@ -108,6 +95,17 @@
             on-select-callback="$ctrl.onSelectPeriod(period)">
           </bh-period-select>
 
+          <bh-stock-filter-exhausted
```
jniles, line 98: Nice 👍

mbayopanda (approved):
The feature is working so fine 👍 great job

There are just two minor comments but the work is LGTM

client/src/modules/stock/lots/modals/search.modal.js
```diff
@@ -17,7 +17,7 @@ function SearchLotsModalController(data, util, Store, Instance, Periods, Stock,
 
   const searchQueryOptions = [
     'depot_uuid', 'inventory_uuid', 'group_uuid', 'label', 'entry_date_from',
-    'entry_date_to', 'expiration_date_from', 'expiration_date_to', 'includeEmptyLot',
+    'entry_date_to', 'expiration_date_from', 'expiration_date_to',
```
jniles, line 20: I think this change should be reversed.  What this array does is tell the modal what values to load in from the `data` object passed in.  In practice, this is how we "remember" the previous values set in the modal and select them.  I worry with this change, we'll lose this functionality.

server/controllers/stock/core.js
jniles: What was this for?  `jquery` isn't a require-able module, I don't think.

server/controllers/stock/core.js
```diff
@@ -204,8 +204,7 @@ function getLots(sqlQuery, parameters, finalClause = '', orderBy) {
  */
 function getLotsDepot(depotUuid, params, finalClause) {
   let _status;
-  // token of query to add if only no empty lots should be returned
-  let excludeToken = '';
+  let emptyLotToken = ''; // query token to include/exclude empty lots
```
jniles, line 207: I like this variable name.

server/controllers/stock/core.js
```diff
@@ -216,9 +215,12 @@ function getLotsDepot(depotUuid, params, finalClause) {
...
   }
 
-  if (Number(params.includeEmptyLot) === 0) {
-    excludeToken = 'HAVING quantity > 0';
+  const includeEmptyLot = Number(params.includeEmptyLot);
+  if (includeEmptyLot === 0) {
+    emptyLotToken = 'HAVING quantity > 0';
     delete params.includeEmptyLot;
+  } else if (includeEmptyLot === 2) {
+    emptyLotToken = 'HAVING quantity=0';
```
jniles, line 223: So, technically this is correct, but I wonder if in practice we would want to know `HAVING quantity <= 0`.  I know that there have been times before where we have missed a flaw in our logic and that has caused negative values.  I am sure it will happen again in the future.  Would it be useful for users to know this?  If so, would it be appropriate for us to show it here?

test/end-to-end/cash/registry.spec.js
```diff
@@ -26,23 +26,22 @@ function CashPaymentsRegistryTests() {
...
     const DEFAULT_PAYMENTS_FOR_TODAY = 2;
     await modal.switchToDefaultFilterTab();
     await modal.setPeriod('today');
     await modal.submit();
     await GU.expectRowCount('payment-registry', DEFAULT_PAYMENTS_FOR_TODAY);
   });
 
-  it('finds one payments for this last year', async () => {
-    const DEFAULT_PAYMENTS_FOR_TODAY = 2;
+  it('finds 2 payments for this last year', async () => {
```
jniles, line 37: 💃

test/end-to-end/stock/stock.inventories.spec.js
```diff
@@ -60,16 +60,16 @@ function StockInventoriesRegistryTests() {
   it('find 2 inventory by state (security reached)', async () => {
```
jniles: With the changes to this test, we'll need to update this text to be 3, not 2.

test/end-to-end/stock/stock.movements.spec.js
```diff
@@ -30,23 +30,23 @@ function StockMovementsRegistryTests() {
     await modal.switchToDefaultFilterTab();
     await modal.setPeriod('allTime');
     await modal.submit();
-    await GU.expectRowCount(gridId, 18 + (2 * depotGroupingRow));
+    await GU.expectRowCount(gridId, 20 + (2 * depotGroupingRow));
```
jniles: Collisions are coming!  🗡️ 

The MySQL8 fixes are going to clobber all this, but hopefully we can sort it out and have a better body of tests afterwards.

test/integration/stock/shared.js
```diff
@@ -156,7 +156,7 @@ const movementDepot = {
 };
 
 const depotPrincipalUuid = 'F9CAEB16168443C5A6C447DBAC1DF296';
-const depotPrincipalMvt = 28; // 10 initial plus 3 distributions + 10 imported
+const depotPrincipalMvt = 30; // 12 initial plus 3 distributions + 10? imported
```
jniles, line 159: Good question.

test/integration/stock/stock.js
```diff
@@ -67,39 +67,39 @@ describe('(/stock/) The Stock HTTP API', () => {
   // list all stock exit relatives to 'Depot Principal'
   it(
     `GET /stock/lots/movements?is_exit=1&depot_uuid=...
-    returns exits for Depot Principal (5 OUT)`,
+    returns list of lot exits for Depot Principal`,
     async () => {
       const res = await agent.get(`/stock/lots/movements?is_exit=1&depot_uuid=${shared.depotPrincipalUuid}`);
```
jniles, line 72: Just as a quick note (no fix needed here) - an easier way to write these queries is to break them into the base url + a query object in `chai`.  It looks like this:

```js
const res  = await (agent.get('/stock/lots/movements')
  .query({ is_exit : 1, depot_uuid : shared.depotPrincipalUuid }));
```

It just makes parameters easier to edit/change and solves the line-length issue.

jniles (commented):
I've made some comments here, but haven't done a thorough test.  I'll do that after I get the MySQL 8 updates pushed in to confirm a few theories.

client/src/js/components/bhStockFilterExhausted.js
jmcameron: I agree; I had already thought of that.  But I put them in there in case we wanted to use different language for Stock > Lots vs Stock > Articles in Inventory.  If you think that is unnecessary, I'm happy to do what you suggest.  Note that using the same language in both right now was only for convenience.

client/src/i18n/en/stock.json
jmcameron: Good catch!

server/controllers/stock/core.js
jmcameron: This was already there but hound complained about it.  I commented it out and did not see any difference.  I plan to move it up as Bruce suggested and see if that works.

client/src/modules/stock/lots/modals/search.modal.js
jmcameron: I do not think this fix is necessary.   I struggled with this issue when implementing this feature.  I tried putting it back in but it does not work (changing the flag does not change the results).   I think this list is related to searchQueries, but includeEmptyLot is now in defaultQueries.    In any case, this option works as is and saves and recalls its value correctly.  So I don't think this needs fixing.

client/src/js/components/bhStockFilterExhausted.js
jmcameron: Ok, I did this.  We can always add it back in if we need to use different language later.

server/controllers/stock/core.js
jmcameron: This did not help.  I'm leaving it commented out.  I suppose we should track down why it is no longer needed or delete the commented out version.

client/src/i18n/en/stock.json
jmcameron: Fixed.

server/controllers/stock/core.js
jmcameron: Leaving it commented out.

jmcameron:
I still need resolve why  the last test in stock.z3.inventory-adjustment-lots.spec.js is broken.

jniles:
So I think I may have a solution for you.  I've opened a PR to this branch: https://github.com/jmcameron/bhima/pull/1

jmcameron:
bors try

jmcameron:
@lomamech Please look at this PR.  It is now passing everthing except editing an offDay.   My fixes in this PR have nothing to do with this, but I have merged master back into it.   Could you see what this is failing and suggest a fix?  Thanks.

Issue #4083, opened 2020-01-13
Make document_uuid unique to a stock movement


jniles:
I stumbled on this while working on #4073.

It looks like, in general, our `document_uuid`s are unique to the stock movement.  There is one case where this isn't true - for transfers between depots.  In this case, the document_uuid is repeated for both the movement into the depot and the exit out of the depot.

Here is a way to prove it to yourself if you have a live database:
```sql
SELECT BUID(document_uuid), COUNT(DISTINCT flux_id) n, GROUP_CONCAT(flux_id) a
FROM stock_movement 
GROUP BY document_uuid 
HAVING n > 1 ORDER BY COUNT(DISTINCT flux_id);
```

This will show that only flux_id 2 and 8 are used in reference to the same document_uuid.  These in FROM_OTHER_DEPOT and TO_OTHER_DEPOT, respectively.

I propose that we do not use the same `document_uuid` across these two different stock_movements.  This would make it more like the `record_uuid` field in the `posting_journal` and `general_ledger` and allow us to uniquely link to these transactions.

@mbayopanda, you did the most work here.  Do you know why two stock movements use the same document_uuid?  Can we change this behavior safely without breaking too many things?

jniles:
This has come up a number of times recently.  I'm bumping the priority up.

PR #4894, opened 2020-09-07
Fix stock movements grouping

jniles:
Stock movements are not uniquely identified by their document_uuid.
This is documented in #4083.  However, the tests made this assumption
because MySQL 5.7 had a stable sort on the GROUP BY.  In MySQL 8, this
is no longer stable.

To solve this issue, grouping on `document_uuid` now includes the `is_exit` column
as well.

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -285,7 +285,7 @@ async function getLotsMovements(depotUuid, params) {
   }
 
   if (params.groupByDocument === 1) {
-    finalClause = 'GROUP BY document_uuid';
+    finalClause = 'GROUP BY document_uuid, is_exit';
     delete params.groupByDocument;
   }
 
@@ -348,7 +348,7 @@ async function getMovements(depotUuid, params) {
   LEFT JOIN service AS serv ON serv.uuid = m.entity_uuid
   `;
 
-  const finalClause = 'GROUP BY document_uuid';
+  const finalClause = 'GROUP BY document_uuid, is_exit';
   const orderBy = 'ORDER BY d.text, m.date';
   const movements = await getLots(sql, params, finalClause, orderBy);
   return movements;
--- a/docs/cloud-init.yml
+++ b/docs/cloud-init.yml
@@ -75,7 +75,7 @@ write_files:
       collation-server=utf8mb4_unicode_ci
       init-connect='SET NAMES utf8mb4'
       character-set-server=utf8mb4
-      sql-mode="STRICT_ALL_TABLES"
+      sql-mode="STRICT_ALL_TABLES,NO_UNSIGNED_SUBTRACTION"
 
   - path: /tmp/bhima.service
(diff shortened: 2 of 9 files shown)


jniles:
Hmm... this is tricky. We are executing the same create statements between the mysql5.7 and mysql8.

I'm going to have to set up a local mysql8 version to try and figure out what is going on.  It may take a while.

jniles:
@jmcameron do these tests break for you locally?

~I've installed mysql8 in docker and the first test locally passed.  I'm curious if you are seeing breakage locally?~

EDIT: finally got a break.  Investigating...

jmcameron:
> @jmcameron do these tests break for you locally?

A couple of weeks ago, I tried getting the tests working on my Windows 10/WSL2 setup.   I can do the client and server unit tests, but the end-to-end and integration tests required selenium.  I've installed Selenium and it seems like it works, but I cannot get the e2e/integration tests to work on the laptop.  I've been relying on the online test runner

jniles:
@jmcameron thanks for the info.  What kinds of errors are you facing?  I did some debugging on a cloud VM and determined that I was fighting an uphill battle because of X11.  Using docker locally, I got tests running with mysql8.

It would be good to debug that sometime on your machine - it would just be useful to figure out if there is a component we aren't understanding and document it somewhere.

jmcameron:
On my old laptop, test:ends fails with this:
image

jmcameron:
On my old (MySQL 8.0) laptop, test:integration passes.

jmcameron:
BTW, I'm using 'master' on my old laptop for these tests.

jmcameron:
Forgot to mention that test:client-unit and test:server-unit both pass on my old laptop.

jmcameron:
> On my old laptop, test:ends fails with this:

I re-ran test:ends on my old laptop and got the same error creating a new Fiscal year.  So it does seem to be repeatable on this system.

jniles:
@jmcameron, try changing the timezone to WAT (GMT+1).  Does that cause the fiscal year test to pass?

jmcameron:
> @jmcameron, try changing the timezone to WAT (GMT+1). Does that cause the fiscal year test to pass?

Still fails

server/controllers/stock/core.js
```diff
@@ -285,7 +285,7 @@ async function getLotsMovements(depotUuid, params) {
   }
 
   if (params.groupByDocument === 1) {
-    finalClause = 'GROUP BY document_uuid';
+    finalClause = 'GROUP BY document_uuid, is_exit';
```
mbayopanda, line 288: 👍

mbayopanda (approved):
LGTM 👍

jniles:
@mbayopanda this should fix the tests on master.

mbayopanda (approved):
LGTM 👍 

I didn't found a strange behavior

bors r+

PR #4897, opened 2020-09-08
Add tracker and importance columns in the inventory table

### LINKED ISSUE #4858: Track important inventories for dashboards
This issue is about to add a new field in the inventory table which defines if an inventory is important for some analysis (`Molecule traceur`)

This details can help us to have a dashboard for this kind of inventories only

mbayopanda:
This PR is related to #4858 

Capture d’écran 2020-09-08 à 06 43 13

Close #4858


server/models/schema.sql
```diff
@@ -778,6 +778,8 @@ CREATE TABLE `inventory` (
   `last_purchase` DATE NULL COMMENT 'This element allows to store the date of the last purchase order of the product in order to allow the calculation without making much of the average ordering INTerval',
   `num_purchase` INT(10) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'Number of purchase orders' ,
   `num_delivery` INT(10) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'Number of stock delivery' ,
+  `tracker` TINYINT(1) NOT NULL DEFAULT 0 ,
+  `importance` VARCHAR(20) NULL COMMENT 'Inventory level of importance : LOW, MID, HIGH' ,
```
jniles: Can we make these integers?  That way, we can use mathematical operators for things like "Anything above mid importance".

Right now if we want to show "mid and higher", we have to do ` if (importance === "MID" || importance ===  "HIGH")`.  But it might be easier to say `if (importance > 2) { // do something }`.

jniles: You would then need to add this to the bhConstants like our Flux IDs.

mbayopanda:
@jniles could you give me review please

client/src/modules/inventory/list/modals/actions.tmpl.html
```diff
@@ -184,6 +184,30 @@
...
+      </label>
+    </div>
+
+    <div class="form-group">
+      <label class="control-label" translate>INVENTORY.IMPORTANCE_LEVEL</label>
+      <select
+        class="form-control"
+        name="importance"
+        ng-model="$ctrl.item.importance">
+        <option value="" disabled translate>INVENTORY.IMPORTANCE_LEVEL_SELECT</option>
```
jniles, line 195: In the issue https://github.com/IMA-WorldHealth/bhima/issues/4858, you mention that NULL is the default importance level and it means "do not track".  I propose that you remove the column `tracker` and just have this column with `NULL` being "not tracked".  What do you think?

jniles: You would then change this from "Select Importance Level" to "do not track inventory item", and then "priority: low", "priority: medium", etc.

server/models/migrations/next/migrate.sql
```diff
@@ -738,3 +738,11 @@ CREATE TABLE  `stock_movement_status` (
     KEY `inventory_uuid` (`inventory_uuid`),
     CONSTRAINT `stock_movment_status__inventory` FOREIGN KEY (`inventory_uuid`) REFERENCES `inventory` (`uuid`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+/*
+ * @author: mbayopanda
+ * @date: 2020-09-07
+ * @description: Add the importance level column in the inventory table
+ */
+ALTER TABLE inventory ADD COLUMN `importance` SMALLINT(5) NULL COMMENT 'Inventory level of importance : 1 -> LOW, 2 -> MID, 3 -> HIGH';
```
jniles: Nice comment 👍

server/models/schema.sql
```diff
@@ -774,10 +774,12 @@ CREATE TABLE `inventory` (
   `locked` TINYINT(1) NOT NULL DEFAULT 0,
   `delay` DECIMAL(10,4) UNSIGNED NOT NULL DEFAULT 1 COMMENT 'Delivery time',
   `avg_consumption` DECIMAL(10,4) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'Average consumption' ,
-  `purchase_INTerval` DECIMAL(10,4) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'Purchase Order INTerval' ,
+  `purchase_interval` DECIMAL(10,4) UNSIGNED NOT NULL DEFAULT 0 COMMENT 'Purchase Order Interval' ,
```
jniles, line 777: 😆

test/integration/inventory/metadata.js
```diff
@@ -107,6 +107,28 @@ describe('(/inventory/metadata) The inventory metadata http API', () => {
...
+        helpers.api.listed(res, 1);
+        const [item] = res.body;
+        expect(item.tracker).to.equal(1);
+      })
+      .catch(helpers.handler);
+  });
+
+  it('GET /inventory/metadata filters on the importance column', () => {
+    return agent.get('/inventory/metadata')
+      .query({ importance : 2 })
```
jniles: Can it filter by `null`?

jniles (commented):
@mbayopanda the code is fine.  I've proposed using a single column with `NULL` being the same as `tracker` = false, and `1`,`2`, or `3` being the same as `tracker` = true.  What do you think?

mbayopanda:
@jniles changes are landed

client/src/i18n/en/inventory.json
```diff
@@ -44,6 +44,13 @@
...
         "UPDATED_VALUE" : "Updated Value",
-        "CHANGES" : "Changes"
+        "CHANGES" : "Changes",
+        "IMPORTANCE_LEVEL":"Importance Level",
+        "IMPORTANCE_LEVEL_SELECT":"Select the importance level",
+        "LOW":"Low",
+        "MID":"Mid",
+        "HIGH":"High",
+        "NONE":"None",
+        "TRACKER":"Tracker"
```
jniles, line 54: It doesn't look like this key is needed anymore.

server/controllers/inventory/inventory/core.js
```diff
@@ -224,7 +229,8 @@ function getItemsMetadata(params) {
   filters.equals('label');
   filters.equals('sellable');
   filters.equals('note');
-
+  filters.equals('importance');
+  filters.custom('find_null_importance', 'inventory.importance IS NULL');
```
jniles, line 233: This is a solution :P

test/integration/inventory/metadata.js
```diff
@@ -107,6 +106,30 @@ describe('(/inventory/metadata) The inventory metadata http API', () => {
...
+        allItemCounted = res.body.length;
+        return agent.get('/inventory/metadata')
+          .query({ importance : 2 });
+      })
+      .then(res => {
+        helpers.api.listed(res, 1);
+        const [item] = res.body;
+        expect(item.importance).to.equal(2);
+        return agent.get('/inventory/metadata')
+          .query({ importance : null });
```
jniles, line 122: Nice!  Really great to see a test about this 👍

jniles (approved):
Looks good to me.  Great work @mbayopanda 

bors r+

jniles:
So ... this is stuck because the github status checks haven't completed, even though they did in reality (see: https://travis-ci.com/github/IMA-WorldHealth/bhima/builds/183662393). 

I'm going to just merge this myself.

PR #4898, opened 2020-09-08
chore(stock): combine credit lines in stock exit movement

### LINKED ISSUE #3962: Distibution to Patient's voucher writes a lot of lines to credit stock account
I'm poking through some of the live data and it looks like we are writing two lines line of stock distributed.  It would be useful to only have a single line leaving the depot (if possible).

This is what it looks like below:
image

Can we do better?

jniles:
This commit combines the credit lines in the voucher of the stock movement to ensure that we aren't recording many more lines in the posting journal than we need to.  You can see the difference below - instead of three lines for stock distribution crediting the stock account, we have a single line.

**Before**
image

**After**
image

Closes #3962.

### DIFF
--- a/server/models/procedures/stock.sql
+++ b/server/models/procedures/stock.sql
@@ -97,10 +97,10 @@ BEGIN
   WHERE ig.stock_account IS NULL AND ig.cogs_account IS NULL AND sm.document_uuid = documentUuid AND sm.is_exit = isExit;
 
   IF (v_has_invalid_accounts > 0) THEN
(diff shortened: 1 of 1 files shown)


jniles:
@lomamech can I get a review?

jniles:
@mbayopanda, since @lomamech isn't feeling well, could you take a look that this PR?

mbayopanda (approved):
It works as expected.

LGTM 👍 

bors r+

mbayopanda:
bors try

PR #4901, opened 2020-09-09
test: force HTML rendering in travis w/ flag

### LINKED ISSUE #4844: Turn PDF rendering off for end to end tests
In the settings, we allow users to select only HTML receipts.  To make our end to end tests more efficient (and less error prone), we could add an environmental variable `FORCE_HTML_RECEIPT_RENDERER` that we would set true for the end to end tests.  This would turn on the HTML renderer for all receipt rather than using the PDF renderer.

Advantages:
  1. Less flake in the end to end tests.
  2. Potentially faster tests (less taxing)
  3. Would allow us to test the actual values in the receipts via CSS selectors, which may open a whole realm of testing possibilities.

Disadvantages:
  1. We may accidentally set this in production.
  2. Would be a less accurate simulation of real life.

We can account for disadvantage 1 by being careful.  We can account for disadvantage 2 by testing the PDF renderer in our unit tests, which we already do.

@jmcameron what are your thoughts on this?

jniles:
Implements the FORCE_HTML_RECEIPT_RENDERER flag to override the user-provided choice of renderer in the end to end tests.

Closes #4844.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,7 +98,7 @@ before_script:
   - sh -c "if [ '$TEST_TYPE' = 'e2e' ]; then yarn webdriver-manager update --versions.standalone=3.141.59 --versions.chrome=${CHROMEDRIVER_VERSION} --gecko false; fi"
 
 script:
-  - sh -c "if [ '$TEST_TYPE' = 'e2e' ]; then yarn test:ends; fi"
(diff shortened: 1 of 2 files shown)


mbayopanda (approved):
LGTM 👍 

bors r+

Issue #4081, opened 2020-01-12
OHADA Stock Accounting


jniles:
It has come to my attention that we do not fully support OHADA stock accounting.  I'll give an overview of regular stock accounting, the discuss what we need to add.

Regular Stock Accounting:

**Purchases**
Supplier delivers stock to the pharmacy:
 1. *debit* - stock account for value of goods
 2. *credit* - supplier for value of goods
 
Invoice is paid to supplier:
 1. *debit* - supplier for the value of goods
 2. *credit* - cash/bank for value of goods.
 
The net result is that we have converted cash assets into stock assets via the supplier's account.
 
**Sales**
When stock is sold:
 1. *debit* - the debtor account for the sale price of the stock.
 2. *credit* - the income account for the sale price of the stock.
 
When the debtor pays:
 1. *debit* - the cash/bank for the sale price.
 2. *credit* - the debtor account for the sale price.
 
The net result is that we have credited our income account (recorded income) and put that money in our cash or bank account.

[Here is an overview of the normal accounting method](https://www.double-entry-bookkeeping.com/bookkeeping-basics/purchase-transaction-journal-entries/).

**Stock Consumption**
When the stock is consumed by the debtor:
 1. *debit* - the cost of goods sold account for the _purchase_ price of the stock.
 2. *credit* - the stock account for the purchase price of the stock.
 
Note that there are only three stock accounts at play here:
 1. Income Account for the goods.
 2. Cost of Goods Sold Account for the goods.
 3. Stock Account for the goods.
 
This is what we have in BHIMA currently.  In this model:
 1. The Cost of Goods Sold (COGS) account shows the value of the stock sold.
 2. The Income Account shows the revenue generated by the stock.
 3. The Income Account - COGS account shows the profit made by that stock.
 4. The Stock account shows the current value of the stock.

---

For OHADA, a different model occurs during the purchase of goods.  Two different transactions occur:

**OHADA Purchasing**
Supplier delivers stock to the pharmacy:
 1. *debit* - a stock _expense_ account for the value of the goods (601)
 2. *credit* - supplier for value of goods

**_Simultaneously, we need to hit the stock account and variation stock._**

Supplier delivers stock to the pharmacy:
 1. *credit* - a stock _variation_ account for the value of the goods (603)
 2. *debit* - the stock account for the value of the goods (31)

Invoice is paid to supplier:
 1. *debit* - supplier for the value of goods
 2. *credit* - cash/bank for value of goods.

[...]

PR #4902, opened 2020-09-09
fix(stock): credit supplier on entry from purchase

jniles:
This commit fixes the stock management accounting to credit the supplier's account when stock is received in the depot from a purchase order.

Importantly, this feature is configurable, so that it doesn't affect enterprises that do OHADA accounting and wish to pass these writings manually.

Fixes a bug our friends in Vanga are facing in production.

Partially addresses #4081.

This will definitely collide with https://github.com/IMA-WorldHealth/bhima/pull/4898.


jniles:
@mbayopanda this is now ready for a review.

jniles:
@mbayopanda I've added an enterprise setting to turn this behavior on and off, as requested.  Can you give it a test?

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4904, opened 2020-09-10
chore: separate stock entry and exit types ctrls

jniles:
Separates the stock entry and exit types controllers into two different  controllers.  This simplifies the entry and exit type functionality.

Tangentially related to https://github.com/IMA-WorldHealth/bhima/issues/4744.


jniles:
bors try

PR #4908, opened 2020-09-11
Show depots on registry only according depots permission

### LINKED ISSUE #4868: Filter Stock Registries by Depot Access
We have a problem when it comes to the stock registries for a country-wide implementation.  What users get access to _see_ what data? 

Currently, we only filter the users by who can _access_ the depot, but every user can see every stock movement or quantity in every depot.  This may be overwhelming for users when we try to install for an entire DPS.  Also, it may be the case that users will not want other health zones to know the CMM of their zone.

We may need to increase the complexity of our users and permissions to allow us limit the view of data in the registries to only certain depots.

mbayopanda:
This PR is related to #4868 

Close #4868


server/controllers/stock/core.js
```diff
@@ -669,10 +677,21 @@ async function getStockConsumptionAverage(periodId, periodDate, monthAverageCons
   return execStockConsumption;
 }
 
+/**
+ * setDepotPermissionCheck
+ * @param {*} params
+ */
+async function setDepotPermissionCheck(params) {
```
jniles: This looks like this function is called every time, right?

You can actually make this a lot simpler by just making a custom filter out of it.  Instead of using this function, just put in your custom filter:
```js
// old version:
filters.custom('depot_permission_check', 'd.uuid IN (?)', [params.depot_permission_check]); 

// new version
filters.custom(
  'user_id',
  'd.uuid IN (SELECT depot_uuid FROM depot_permission WHERE user_id = ?)'
); 
```

You then don't need the `setDepotPermissionCheck` function at all!  It should do the same thing.  All you need to do now is check if `user_id` is set or not.

server/controllers/stock/core.js
```diff
@@ -256,18 +261,17 @@ function getLotsDepot(depotUuid, params, finalClause) {
...
-    .then(stockManagementProcess)
-    .then(processMultipleLots)
-    .then((rows) => {
-      if (_status) {
-        return rows.filter(row => row.status === _status);
-      }
-      return rows;
-    });
+
+  const inventories = await db.exec(query, queryParameters);
```
jniles, line 266: Nice use of `await` 👯

jniles (commented):
@mbayopanda I've made a suggestion to simplify the filter.  Let me know what you think!

Is there a way to turn this behavior on and off, like in an enterprise setting?  I think it might make sense since our institutions probably don't expect this change.

server/controllers/stock/core.js
jniles: If you don't want to use `user_id` (for example, if it clashes with something else), just use another `param` name.

server/controllers/stock/index.js
```diff
@@ -558,6 +558,10 @@ async function listLots(req, res, next) {
 function listLotsMovements(req, res, next) {
   const params = req.query;
 
+  if (req.session.enterprise.settings.enable_strict_depot_permission) {
+    params.check_user_id = req.session.user.id;
```
jniles, line 562: Beautiful 👍

server/models/migrations/next/migrate.sql
```diff
@@ -759,3 +759,10 @@ ALTER TABLE inventory ADD COLUMN `importance` SMALLINT(5) NULL COMMENT 'Inventor
  * entry from purchase.
  */
 ALTER TABLE `enterprise_setting` ADD COLUMN `enable_supplier_credit` TINYINT(1) NOT NULL DEFAULT 0;
+
+/*
+ * @author: mbayopanda
+ * @date: 2020-09-16
+ * @desc: add enterprise setting for strict depot permission which will restrict stock registries to depot owner
+ */
+ ALTER TABLE `enterprise_setting` ADD COLUMN `enable_strict_depot_permission` TINYINT(1) NOT NULL DEFAULT 0;
```
jniles, line 768: Nice 👍

.env.development
```diff
@@ -0,0 +1,24 @@
+# Server Config
```
jniles: This should be removed 👯

jniles:
@mbayopanda, let me know when you need another review!

mbayopanda:
@jniles changes was landed, can I get a review please

jniles (commented):
Looks pretty good to me!  Changes address all my concerns.

Can you remove the `.env.development` file?

jniles (approved):
LGTM.

bors r+

PR #4914, opened 2020-09-12
Add SemaphoreCI for testing instead of AppVeyor

jniles:
This PR configures semaphore for running our integration tests and removes appveyor.

Well, I've bit the bullet and am working again on testing with a different service.  My rationale for moving away from AppVeyor is the following:

 1. We ran into multiple issues with AppVeyor and environmental variables resulting from AppVeyor confusing my personal account and the Github org account.
 2. AppVeyor did not support parallel builds.
 3. AppVeyor is built for windows and it was kind of cludgey getting the linux builds working correctly.
 4. They do not support deleting accounts.  You can't do it.  You have to open a support ticket.  This is a red flag I was not aware of when installing it.

Things that I have learned since moving:
  1. Semaphore CI is blazingly fast.  I'm super happy with it.
  2. The configuration files are intuitive and their web GUI is really nice.

I'm still working on seeing if bors will respect this, but at least it's a step.


jniles:
bors try

jniles:
bors try

jniles:
bors try-

PR #4915, opened 2020-09-12
chore: add installation tests to semaphore

jniles:
Adds installation tests to the semaphore testing regime. Removes them from Travis.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -4,15 +4,18 @@ agent:
   machine:
     type: e1-standard-2
     os_image: ubuntu1804
+
 blocks:
   - name: Update Dependencies
+    dependencies: []
     task:
       jobs:
         - name: Update Dependencies
           commands:
             - sudo apt-get -y update && sudo apt-get install mysql-client redis-tools
             - sudo apt autoremove -y && sudo apt autoclean -y
-  - name: Testing Matrix
+  - name: Integration Testing Matrix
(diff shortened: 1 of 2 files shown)


jniles:
bors try

jniles:
bors try

jniles:
Excellent. This should reduce the test time.

Overall test time for the Semaphore CI run is 3:58. This is an increase of slightly less than 2 minutes.

This guts the installation tests run on Travis so that Travis can focus entirely on End to End tests.   This should speed up overall test runs.  Travis typically spends ~2-3 minutes building and running these tests.  Transferring them to Semaphore makes sense since they will not hog up precious resources over at Travis.

jmcameron (approved):
This looks okay to me.  But I'm not familiar with the semaphore or travis ci settings, so I cannot make an accurate judgement.  And it seems like the way to test this is with PRs.   So my suggestion is to go ahead with it and monitor it closely for a few days until it is clear that it working correctly.   (Trial by fire!)

jniles:
Thanks @jmcameron.

As long as we trust the tests are doing the right thing, one way to verify that the runs are executing correctly is by doing "bors try".  What that does is merge master and the present branch, then run the tests and see if that returns a success code.

It would be good for us to walk through some of the testing infrastructure together at some point. I can see that being a real asset going forwards!

bors r+

jniles:
Looks like https://github.com/IMA-WorldHealth/bhima/issues/4919 is hitting this.

bors r+

PR #4920, opened 2020-09-14
fix(reports): stock expiration report

### LINKED ISSUE #4835: Improvements to the Stock Expiration Report
Here is what the stock expiration report looks like:

image

We can make a few improvements to this based on my conversations with Dr. Larry.

1. Let's add a section at the _top_ of this report that shows three large values:
   1. Current value of the stock
   2. Value of the stock in risk of expiry
   3. Value of the stock that is expired
image

2. The colors in the report are also hard to read.  We should use a standard color scheme, like we have in the cashbox analysis report or the recover report:
image

jniles:
This commit fixes the stock expiration report by doing the following
things:
  1. Removes the date range as these don't make any sense.
  2. Uses the same color scheme as used in the other reports for warning/danger text.
  3. Adds total summaries to the top of the report.
  4. Removes the inventory select.
  5. Simplifies the queries so that less for loops are done.
  6. Fixes the currency bug so things render in the enterprise currency.
  7. Adds the column for the unit type.

Closes #4835.

Here is what it looks like now:
image

----

I chose not to implement the "total value in stock" since this would be a rather expensive query and probably only makes sense if we are looking at a single depot at a time.

### DIFF
--- a/client/src/i18n/en/report.json
+++ b/client/src/i18n/en/report.json
@@ -149,8 +149,8 @@
       "EXCLUDE_INVENTORIES_ZERO_VALUE": "Exclude inventories with zero stock value"
     },
     "STOCK_EXPIRATION_REPORT" : {
-      "TITLE": "Stock expiration report",
-      "DESCRIPTION": ""
+      "TITLE": "Stock Expiration Report",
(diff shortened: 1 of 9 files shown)


lomamech (approved):
I just compared the results with the data from the batch log and everything is working fine

PR #4921, opened 2020-09-15
Feature stock requisition

### LINKED ISSUE #4849: Stock Requisition Features To Add
This issue is about some feature we can add in the stock requisition registry : 

- Forbid the deletion of stock requisition if the requisition has been implied in any process
- Add a link to depot movement receipt in case of depot movement from requisition
- Add a link to stock movement registry in case of depot movement from requisition
- Add more status:
    - Canceled
    - Movement initialized: in case a stock exit to a depot from requisition is made
    - Movement fully completed: in case all inventories of the requisition have been received by the destination
    - Movement partially completed: in case just some inventories not all of the requisition have been received by the destination

lomamech:
- Forbid the deletion of stock requisition if the requisition has been
  implied in any process in Client Side
- Adding a link to depot movement receipt in case of depot movement from
  requisition
-Movement fully completed: in case all inventories of the requisition have been received by the destination
- Adding a new component to select the requisition reference
- Addition of the link from the register of requisitions to stock
  movements
- Adding a new component to select the requisition reference
- Addition of the link from the register of requisitions to stock
  movements

- Finalization of the management of the various requisition status
- Restrict the delivery of stocks by requisition to requesting departments
  and depots and to target depots

- Display requisition reference in stock exit to service
- Update and improve SQL request for getting balance for movement from
  requisition
-Improved bh-depot-select component added filter option for user-defined
repositories
- Improvement of alert messages when out of stock by requisition
- Set data for build E2E test for exit stock by requisition
- Add requisition status excessive received
- Add stock_requisition uuid for document map
- Test: set data for E2E test and integration
- Test: Reset value for many E2E requisition test

closes #4849


test/integration/stock/requisition.js
```diff
@@ -11,7 +11,7 @@ describe('(/stock/requisition) The Stock Assign HTTP API', () => {
     'depot_requestor', 'reference', 'items', 'status_key', 'title_key',
   ];
 
-  // create new stock requisition on "Depot Principal" from a service
+  // create new stock requisition on "Depot Principal" from a servicedelete stock requisition
```
jniles, line 14: This looks like a typo.

lomamech:
@jniles  @mbayopanda  can you review this PR

client/src/i18n/en/form.json
```diff
@@ -219,6 +219,11 @@
             "PAYMENT": "Cash Payment",
             "PREVIOUS": "Previous",
             "RECORD_SAME":"No changes have been made to the form",
+            "REQUISITION": {
+              "CANCEL": "Cancel the inventory requisition request",
+              "DONE": "The requisition can no longer be completed, and will be considered done",
```
jniles: You don't need this comma here.  This can be written best as `The requisition can no longer be completed and will be considered done.`

client/src/i18n/en/form.json
```diff
@@ -219,6 +219,11 @@
...
+              "IN_PROGRESS": "Rehabilitated the requisition request"
```
jniles, line 226: What is this used for?  It doesn't make sense to me.

client/src/js/components/bhDepotSelect/bhDepotSelect.js
```diff
@@ -5,6 +5,7 @@ angular.module('bhima.components')
     transclude  : true,
     bindings    : {
       depotUuid        : '<',
+      user             : '<',
```
jniles: I think this should be optional `<?`.

client/src/js/components/bhHasRequisitionVoucher/bhHasRequisitionVoucher.js
```diff
@@ -4,8 +4,11 @@ angular.module('bhima.components')
     controller  : bhHasRequisitionVoucherController,
     bindings    : {
       message   : '<',
+      classeLabel : '<',
```
jniles, line 7: Why can't we just use the regular success/error/info class states for this component?  Why do we need to pass in the class?

client/src/js/components/bhRequisitionSelect.js
```diff
@@ -0,0 +1,50 @@
+angular.module('bhima.components')
+  .component('bhRequisitionSelect', {
+    templateUrl : 'modules/templates/bhRequisitionSelect.tmpl.html',
+    controller  : RequisitionSelectController,
+    transclude  : true,
+    bindings    : {
+      requisitionUuid : '<',
+      depotUuid : '<',
```
jniles: Are either of these optional?  Are both required?  What happens if someone passes in both a service-uuid and a depot-uuid?

client/src/js/components/bhRequisitionSelect.js
```diff
@@ -0,0 +1,50 @@
...
+  $ctrl.$onInit = function onInit() {
+    $ctrl.required = $ctrl.required || false;
+    const params = {};
+
+    if ($ctrl.depotUuid) {
+      params.requestor_uuid = $ctrl.depotUuid;
+    }
+
+    if ($ctrl.serviceUuid) {
+      params.requestor_uuid = $ctrl.serviceUuid;
```
jniles: Since both the `depotUuid` and the `serviceUuid` map to the `requestor_uuid`, I propose you change the `bindings` to just be:
```
requestorUuid: '<?'
```

client/src/js/components/bhRequisitionSelect.js
```diff
@@ -0,0 +1,50 @@
...
+  'StockService', 'NotifyService',
+];
+
+/**
+ * Requisition Select Controller
+ *
+ */
+function RequisitionSelectController(Stock, Notify) {
+  const $ctrl = this;
+  $ctrl.label = $ctrl.label || 'FORM.LABELS.REQUISITION_REFERENCE';
```
jniles: This label should be in the `$onInit` block.

server/controllers/stock/requisition/requisition.js
```diff
@@ -42,6 +42,43 @@ async function getDetails(identifier) {
...
+      JOIN lot AS l ON l.uuid = m.lot_uuid
+      JOIN inventory AS inv ON inv.uuid = l.inventory_uuid
+      JOIN stock_requisition_movement AS srm ON srm.document_uuid = m.document_uuid
+      WHERE srm.stock_requisition_uuid = ?
+      GROUP BY inv.uuid
+    ) AS mouv ON mouv.inventory_uuid = req.inventory_uuid
+    WHERE (req.quantity - IF(mouv.quantity, mouv.quantity, 0)) > 0;
+  `;
+
+  const requisition = await db.one(sqlRequisition, [uuid]);
```
jniles: You can actually run both of these at the same time:
```js
const [requisition, items] = await Promise.all([ db.one(sqlRequisition), db.exec(sql) ]);
```

client/src/js/components/bhServiceOrDepot/bhServiceOrDepot.html
```diff
@@ -41,6 +41,7 @@
   <div ng-if="$ctrl.requestorType.type_key === 'depot'">
     <bh-depot-select
       depot-uuid="$ctrl.uuid"
+      user="true"
```
jniles: What does this do?  I can't see where this is used in the `bhDepotSelect` component.

client/src/modules/stock/exit/exit.js
```diff
@@ -575,10 +575,18 @@ function StockExitController(
...
-        const COMPLETED_STATUS = 2;
-        return Stock.stockRequisition.update(vm.requisition.uuid, { status_id : COMPLETED_STATUS });
+        if (!vm.requisition.uuid) { return null; }
+
+        const movementRequisition = {
+          stock_requisition_uuid : vm.requisition.uuid,
+          document_uuid : documentUuid,
+        };
+
+        const COMPLETED_STATUS = 6;
```
jniles: Is there somewhere global that this can be defined?  Like `bhConstant` or something like that?  I'm not sure I understand why it changed from "2" to "6" since both numbers don't mean anything to me in this context.

client/src/modules/stock/exit/modals/findService.modal.js
```diff
@@ -44,17 +44,32 @@ function StockFindServiceModalController(Instance, Service, Notify, Data, Stock)
   }
 
   function requisitionDetails([requisition]) {
+    if (Data.depot.uuid !== requisition.depot_uuid) {
+      vm.requisitionMessage = 'REQUISITION.NOT_FOR_DEPOT';
+      vm.requisitionLabel = 'label label-warning';
+      throw new Error('The requisition is not for depots');
```
jniles: I think this might be better written "This requisition is not for this depot".

jniles (commented):
I haven't given this a proper test, but I've added some comments to try and tighten up the client-side code.

client/src/i18n/en/form.json
lomamech: This message only appears when a requisition has been canceled and it needs to be rehabilitated, **rehabilitated requirements have the status In Progress**

client/src/js/components/bhHasRequisitionVoucher/bhHasRequisitionVoucher.js
lomamech: I just wanted the class to be defined dynamically

client/src/js/components/bhRequisitionSelect.js
lomamech: In this case, I will use two variables, one to retrieve the Uuid and the other for the type of requirement depot or service

client/src/js/components/bhRequisitionSelect.js
lomamech: Cool

mbayopanda (commented):
Great job @lomamech I have just some comment about the use of.

- When out of stock via a requisition there is the status "Partially" which is assigned to the requisition; so that this status cannot lead to confusion can we have:
   - Partially sent
   - Partially received
   - Or just partially processed
  
- When leaving stock to a depot, the use of the requisition depends on the depot selected in the modal, is it intentional or not? otherwise, the deposit selection zone must be reset when choosing the requisition option

- The list of items not available for output via a requisition does not show any item just empty bullets.

client/src/js/components/bhServiceOrDepot/bhServiceOrDepot.html
lomamech:
````
    if ($ctrl.user) {
      options.only_user = true;
    }
````
**https://github.com/IMA-WorldHealth/bhima/blob/c6a3349a36a2009920c98f6d9320693b9e1d17d8/client/src/js/components/bhDepotSelect/bhDepotSelect.js**
it allows to filter only the repositories defined for the user

**Comment by lomamech** [comment]
> Great job @lomamech I have just some comment about the use of.
> 
> * When out of stock via a requisition there is the status "Partially" which is assigned to the requisition; so that this status cannot lead to confusion can we have:
>   
>   * Partially sent
>   * Partially received
>   * Or just partially processed
> * When leaving stock to a depot, the use of the requisition depends on the depot selected in the modal, is it intentional or not? otherwise, the deposit selection zone must be reset when choosing the requisition option
> * The list of items not available for output via a requisition does not show any item just empty bullets.

From the way I understood this module of the management of the requirements, the statuses which appear on the registers of the requirements, concern only the withdrawals of stocks and do not refer to the entry of stock

The status partially concerns only the sending (Out of stock)

But I will **not say no** so that we add a new column to the register of requirements for tracing the entries into stocks of requesting deposits

**Comment by lomamech** [comment]
The list of items not available for output via a requisition does not show any item just empty bullets.
**_I Will fix it_**

File: client/src/i18n/en/form.json [hunk 13]
```diff
@@ -823,12 +829,13 @@
...
-                "DRAFT":"Draft",
-                "CANCELLED":"Cancelled",
-                "DONE":"Done"
+              "CANCELLED":"Cancelled",
+              "COMPLETED":"Completed",
+              "DONE":"Done",
+              "DRAFT":"Draft",
+              "EXCESSIVE_RECEIVED_QUANTITY": "Received but with excessive quantity",
+              "IN_PROGRESS":"In Progress",
+              "PARTIALLY":"Partially"
```
mbayopanda: Is it possible to a bit explicit, for example: Partially executed or partially processed because when we read the status column and we see just "Partially" we would like to know partially what?

client/src/i18n/fr/form.json
```diff
@@ -219,6 +219,11 @@
             "PAYMENT": "Référence du Paiement",
             "PREVIOUS": "Précédent",
             "RECORD_SAME":"Aucun changement n'a été apporté sur le formulaire",
+            "REQUISITION": {
+              "CANCEL": "Annuler la demande de réquisition des stocks",
+              "DONE": "La requisition ne pourra plus être completée et sera consideré comme fait",
+              "IN_PROGRESS": "Rehabilité la demande de requisition"
```
mbayopanda, line 226: "IN_PROGRESS" is not "La requisition n'a pas encore été traitée" ?

mbayopanda: 👍 Gotcha... It would be intuitive if knew that it is for the description of requisition status.
```
"STATUS_IN_PROGRESS_DESCRIPTION":"Rehabilité la demande de requisition"
```

lomamech:
Following @mbayopanda  's comment, I propose to add another column, to differentiate the status of the requirements made for the deposits, so we could analyze the status when sending as well as that of reception.
image

@jniles @mbayopanda  Do you think this new option is necessary

client/src/modules/stock/exit/modals/findDepot.modal.js
```diff
@@ -17,7 +17,7 @@ function StockFindDepotModalController(Instance, Depot, Notify, Data, Stock) {
   vm.submit = submit;
   vm.cancel = cancel;
 
-  Depot.read()
+  Depot.read(null, { only_user : true })
```
mbayopanda: Having this restriction here will oblige to distribute stocks only to depots that the user manages. The user must have the possibility to distribute stock also to depot that he doesn't manage.

jniles:
@lomamech nice screenshot! 

Does it make sense for something to be partially received, but totally sent?  I guess I don't understand that part.  Otherwise, that looks fine by me.

lomamech:
> @lomamech nice screenshot!
> 
> Does it make sense for something to be partially received, but totally sent? I guess I don't understand that part. Otherwise, that looks fine by me.

It may be that for a requisition one requires **30 quantity** of an inventory, but that the main deposit only sends 20 (Partial) and that the requesting deposit receives the 20 (Total received),

lomamech:
But if we already have a report that allows us to track the status of inter-depot transfers, we could abandon this option.

server/controllers/stock/requisition/requisition.js
```diff
@@ -155,7 +196,85 @@ exports.update = async (req, res, next) => {
...
+          WHERE balance.quantity < 0;
+        `;
+
+      const excessiveStatus = await db.one(checkExcessiveBalance, [
+        dataMovementRequisition.stock_requisition_uuid, dataMovementRequisition.stock_requisition_uuid,
+      ]);
+
+      if (movementStatus.numberInventoryPartial > 0) {
+        // Partially
+        requisition.status_id = 3;
```
mbayopanda: Is it possible to use named constant, such as: 
```js
const PARTIALLY_PROCESSED = 3
...
requisition.status_id = PARTIALLY_PROCESSED;
```

server/controllers/stock/requisition/requisition.js
```diff
@@ -155,7 +196,85 @@ exports.update = async (req, res, next) => {
...
+      }
+
+      if (movementStatus.numberInventoryPartial === 0 && excessiveStatus.numberExcessive > 0) {
+        // Excessive
+        requisition.status_id = 7;
```
mbayopanda: Use constant also

mbayopanda (commented):
@lomamech once again this is great work 👍 

I have noticed some stuff to enhance the work so I commented them

client/src/modules/stock/exit/modals/findDepot.modal.js
lomamech: My fear is that in the context where the health zones will use BHIMA to manage stocks, the BCZs would risk having the possibility of having the possibility of distributing products (By mistake) to the depots that do not concern them

server/controllers/stock/requisition/requisition.js
lomamech: :white_check_mark:

server/controllers/stock/requisition/requisition.js
lomamech: :white_check_mark:

lomamech:
> @lomamech nice screenshot!
> 
> Does it make sense for something to be partially received, but totally sent? I guess I don't understand that part. Otherwise, that looks fine by me.

Instead of adding a column, I discussed with Mbayo, and I propose to add an issue for a new report which will aim to track the transfers of stocks between depots, to thus track those who are pending , complete, excessive, a bit like the cash register analysis reports

jniles:
@lomamech, sounds good.  Let me know when you need another review.

client/src/i18n/en/form.json
jniles: Hmm... I think we could find a better term for this, but I cannot think of one right away.

client/src/i18n/en/form.json
jniles: I agree with this.  It isn't blocking as we have had a few discussion about this so far and haven't settled on a good idea yet.

client/src/i18n/fr/form.json
jniles: Looks like this was addressed. 👍

client/src/js/components/bhDepotSelect/bhDepotSelect.js
```diff
@@ -5,6 +5,7 @@ angular.module('bhima.components')
     transclude  : true,
     bindings    : {
       depotUuid        : '<',
+      user             : '<?',
```
jniles: We could probably just make this `@?` instead of `<?` to save a listener.

I would also rename it to describe what it does.  It does not take a user object.   It filters the depots on the user's permissions.  So something like `filterByUserPermissions` or `filterOnUser` or something would make more sense to me.

client/src/js/components/bhHasRequisitionVoucher/bhHasRequisitionVoucher.html
```diff
@@ -2,22 +2,24 @@
...
+    <bh-requisition-select
+      requisition-uuid="$ctrl.requisitionReference"
+      requestor-uuid="$ctrl.requestor"
+      required="true"
+      on-select-callback="$ctrl.onSelectRequisition(requisition)">
+    </bh-requisition-select>
 
     <div class="help-block" ng-show="$ctrl.message">
-      <p translate>{{ $ctrl.message }}</p>
+      <p class="{{ $ctrl.classeLabel }}" translate>{{ $ctrl.message }}</p>
```
jniles: This should probably use [`ng-class`](https://docs.angularjs.org/api/ng/directive/ngClass) here.

client/src/modules/stock/exit/modals/findService.modal.js
```diff
@@ -44,17 +44,32 @@ function StockFindServiceModalController(Instance, Service, Notify, Data, Stock)
...
+
     if (!requisition || !requisition.uuid) {
       vm.requisitionMessage = 'REQUISITION.VOUCHER_NOT_FOUND';
+      vm.requisitionLabel = 'label label-warning';
       throw new Error('Requisition Not Found');
     }
 
-    if (requisition.status_key === 'done') {
+    if (requisition.status_key === 'done' || requisition.status_key === 'completed'
+      || requisition.status_key === 'excessive') {
```
jniles, line 60: It looks like you have this logic in two places - `findService.modal.js` and `findDepot.modal.js`.  They look really similar - in fact, you throw the error "not for this depot" here, even though this is the find service modal

I think a good next step would be to refactor this into a `RequisitionService.validateRequsitionResult()` and use that in both modals.

client/src/modules/stock/movements/templates/action.cell.html
```diff
@@ -18,6 +18,12 @@
       </a>
     </li>
 
+    <li ng-if="row.entity.stock_requisition_uuid">
+      <a data-method="view-requisition" href ui-sref="stockRequisition({ filters : [{ key : 'reference', value : row.entity.document_requisition, displayValue: displayValue}]})">
```
jniles: We should probably put "cacheable: false" here.

client/src/modules/stock/movements/templates/action.cell.html
```diff
@@ -18,6 +18,12 @@
...
+        <i class="fa fa-file-pdf-o"></i> <span translate>REPORT.VIEW_REQUISITION</span>
```
jniles: Does this open a PDF?  If not, we probably should not show the PDF icon.

client/src/modules/stock/movements/templates/action.cell.html
jniles: Also, we might want to add the `period` filter with a value 'allTime' to ensure it shows all time.

client/src/modules/stock/requisition/modals/status.tmpl.html
```diff
@@ -0,0 +1,54 @@
+<div class="modal-header">
+  <ol class="headercrumb">
+    <li class="static"><span translate>FORM.LABELS.STATUS</span> <span translate>REQUISITION.STOCK_REQUISITION</span></li>
+    <li class="title">{{ $ctrl.requisition.reference }}</li>
+  </ol>
+</div>
+
+<form 
+  name="ModalForm"
```
jniles, line 9: Why does this modal exist?  Shouldn't these be modified automatically?  Except for canceled which is an action, not a status.

client/src/modules/stock/requisition/modals/action.modal.html
```diff
@@ -50,6 +56,7 @@
     <div id="depot-supplier" class="form-group">
       <bh-depot-select
         depot-uuid="$ctrl.model.depot_uuid"
+        user="true"
```
jniles: This appears buggy:

image

server/controllers/inventory/depots/index.js
```diff
@@ -208,12 +213,20 @@ function searchByName(req, res, next) {
       LEFT JOIN sector s ON s.uuid = v.sector_uuid
       LEFT JOIN province p ON p.uuid = s.province_uuid
       LEFT JOIN country c ON c.uuid = p.country_uuid
+      LEFT JOIN depot_permission dp  ON dp.depot_uuid = d.uuid
```
jniles: I think you need to remove these.  There isn't a need do a left join here and only causes repeats in this list.

image

server/controllers/stock/core.js
```diff
@@ -108,6 +109,7 @@ function getLotFilters(parameters) {
   filters.equals('invoice_uuid', 'invoice_uuid', 'm');
   filters.equals('purchase_uuid', 'origin_uuid', 'l');
   filters.equals('tag_uuid', 'tags', 't');
+  filters.equals('stock_requisition_uuid', 'uuid', 'sr');
```
jniles: Let's make this a custom filter that looks up the value in requisition movements similar to how the `only_user` works:

```js

  filters.custom(
    'user_id',
    'd.uuid IN (SELECT depot_permission.depot_uuid FROM depot_permission WHERE depot_permission.user_id = ?)',
  );

```

client/src/modules/stock/movements/templates/action.cell.html
jniles: Also, we should be filtering on the UUID of the requisition, not on the text, I think.

server/controllers/stock/core.js
```diff
@@ -345,14 +347,18 @@ async function getMovements(depotUuid, params) {
...
-    LEFT JOIN service AS serv ON serv.uuid = m.entity_uuid
+  JOIN lot l ON l.uuid = m.lot_uuid
+  JOIN inventory i ON i.uuid = l.inventory_uuid
+  JOIN depot d ON d.uuid = m.depot_uuid
+  JOIN flux f ON f.id = m.flux_id
+  LEFT JOIN document_map dm ON dm.uuid = m.document_uuid
+  LEFT JOIN service AS serv ON serv.uuid = m.entity_uuid
+  LEFT JOIN stock_requisition_movement smr ON smr.document_uuid = m.document_uuid
+  LEFT JOIN stock_requisition sr ON sr.uuid = smr.stock_requisition_uuid
+  LEFT JOIN document_map sr_m ON sr_m.uuid = sr.uuid
```
jniles: You are adding a huge amount of JOINs to a query that is executed frequently!  Do we really need the requisition uuid with every movement, even if the depot rarely uses requisitions...?

I'm not really comfortable with this change.  Where is it that you need the stock requisition text?  Can you just look that up in a separate query?

server/controllers/stock/requisition/requisition.js
```diff
@@ -155,7 +199,85 @@ exports.update = async (req, res, next) => {
...
+      const checkExcessiveBalance = `
+      SELECT COUNT(balance.inventory_uuid) AS numberExcessive
+        FROM (
+        SELECT req.inventory_uuid, req.code, req.inventoryType,
+        (req.quantity - IF(mouv.quantity, mouv.quantity, 0)) AS quantity
+            FROM (
+              SELECT BUID(i.uuid) inventory_uuid, i.code, it.text as inventoryType, sri.quantity
+              FROM stock_requisition_item sri
+              JOIN inventory i ON i.uuid = sri.inventory_uuid
+              JOIN inventory_type it ON i.type_id = it.id
```
jniles: You don't need the JOINs on the inventory_type.  In fact, if all you need is the movement quantity, you probably don't need a lot of these JOINs.  Can you try to remove them?

server/controllers/stock/requisition/requisition.js
```diff
@@ -155,7 +199,85 @@ exports.update = async (req, res, next) => {
...
+      const checkRequisitionBalance = `
+      SELECT COUNT(balance.inventory_uuid) AS numberInventoryPartial
+        FROM (
+        SELECT req.inventory_uuid, req.code, req.inventoryType,
+        (req.quantity - IF(mouv.quantity, mouv.quantity, 0)) AS quantity
+            FROM (
+              SELECT BUID(i.uuid) inventory_uuid, i.code, it.text as inventoryType, sri.quantity
+              FROM stock_requisition_item sri
+              JOIN inventory i ON i.uuid = sri.inventory_uuid
+              JOIN inventory_type it ON i.type_id = it.id
```
jniles: Same comment about the inventory_type here.  It doesn't look like you need it.

server/models/admin.sql
```diff
@@ -63,6 +63,11 @@ BEGIN
     SELECT voucher.uuid, CONCAT_WS('.', 'VO', project.abbr, voucher.reference)
     FROM voucher JOIN project where project.id = voucher.project_id;
 
+  -- stock_requisition
+  INSERT INTO document_map
+    SELECT stock_requisition.uuid, CONCAT_WS('.', 'SREQ', stock_requisition.reference)
```
jniles: Normally, our document maps are composed of three different elements, but here you only use two.  Can you add in the project to make it into three?

Ex) Invoices -> `IV.HBB.123`
Ex) Stock Movements -> `SM.12.1234`

server/models/migrations/next/migrate.sql
```diff
@@ -777,13 +777,27 @@ INSERT INTO `actions`(`id`, `description`) VALUES
...
+UPDATE unit SET parent = 233 WHERE id = 238;
+
+/* @author: lomamech
+ * @date: 2020-09-15
+ * @description: Stock Requisition Features To Add # 4849
+ */
+DROP TABLE IF EXISTS `stock_requisition_movement`;
+CREATE TABLE `stock_requisition_movement` (
+  `stock_requisition_uuid` BINARY(16) NOT NULL,
+  `document_uuid`BINARY(16) NOT NULL,
```
jniles: I think you need a space between the document_uuid and the BINARY here.

jniles (commented):
Most of this is looks good.

I'm actually quite concerned about all the `JOIN`s added here.   We are adding a huge amount of overhead to frequently called routes.  Our stock routes are already slow (Kintambo was upset with us when we used a raspberry pi there).  I think we need to find a better way than introducing more JOINs into the stock routes.

I've made some proposals, but you know where you need them more than I do.

lomamech:
All Checks have passed @jniles @mbayopanda

mbayopanda (changes requested):
In the requisition form, the supplier depot must not be restricted to the user permission on the depot

client/src/modules/stock/exit/modals/findService.modal.js
lomamech: :+1:

client/src/modules/stock/requisition/modals/status.tmpl.html
lomamech: The modals can also display a small message, to remind the user of the scope of each action
image

server/controllers/stock/core.js
lomamech: here I did not understand very well

server/controllers/stock/core.js
lomamech: It was just because of this constraint where there was almost a link between the stock movements and the requirements.

image

jniles: Yes, but this doesn't mean there needs to be a JOIN.  This just means there needs to be a link.

server/controllers/stock/requisition/requisition.js
lomamech: :+1:

client/src/modules/stock/exit/modals/findDepot.modal.js
jniles: This doesn't really make sense though.  Because then each CDR will need to have access to every BCZ depot.  Are you sure that makes sense?

Additionally, to send a requisition to the BCZ, the CS will need to have access to change the stock in the BCZ...

lomamech:
@jniles  I just finished the treatment of the reviews that you had done

server/controllers/stock/core.js
lomamech: I propose to add the requisition_uuid column on the stock_movement table in the same way as there is the invoice_uuid column
image

lomamech:
@jniles  @mbayopanda  I need another review or approval

client/src/modules/stock/requisition/modals/action.modal.html
```diff
@@ -50,6 +56,7 @@
     <div id="depot-supplier" class="form-group">
       <bh-depot-select
         depot-uuid="$ctrl.model.depot_uuid"
+        filter-by-user-permission="true"
```
jniles, line 59: Nice 👍

client/src/modules/stock/requisition/modals/status.tmpl.html
jniles: Personally, I find having one radio button confusing. 

I think a better interface would use ModalAction for this - where the message is just displayed and the user chooses ok/cancel.  This would involve the user clicking "cancel", "confirm", etc on a dropdown menu in the 'actions' column first though.

client/src/modules/stock/requisition/templates/action.cell.html
```diff
@@ -1,20 +1,42 @@
-<div class="ui-grid-cell-contents text-right" uib-dropdown dropdown-append-to-body uib-dropdown-toggle ng-if="!row.groupHeader">
-  <a href>
+<div class="ui-grid-cell-contents text-right" uib-dropdown dropdown-append-to-body>
+  <a href uib-dropdown-toggle>
```
jniles, line 2: Nice 👍

client/src/modules/stock/requisition/templates/action.cell.html
```diff
@@ -1,20 +1,42 @@
...
     <span data-method="action" translate>FORM.BUTTONS.ACTIONS</span>
     <span class="caret"></span>
   </a>
 
-  <ul data-action="{{ rowRenderIndex}}"  class="dropdown-menu" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
+  <ul class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
```
jniles, line 7: Nice 👍

server/controllers/stock/requisition/requisition.js
```diff
@@ -78,11 +119,13 @@ function getStockRequisition(params) {
   filters.equals('service_requestor', 'name', 's');
   filters.equals('depot_requestor', 'text', 'dd');
   filters.equals('uuid', 'uuid', 'sr');
+  filters.equals('stock_requisition_uuid', 'uuid', 'sr');
   filters.equals('type_id', 'requestor_type_id', 'sr');
   filters.equals('status_id', 'status_id', 'sr');
   filters.equals('depot_uuid', 'depot_uuid', 'sr');
   filters.equals('requestor_uuid', 'requestor_uuid', 'sr');
   filters.equals('user_id', 'user_id', 'sr');
+  filters.equals('project_id', 'project_id', 'sr');
```
jniles, line 128: Nice 👍

server/models/schema.sql
```diff
@@ -1936,15 +1936,19 @@ CREATE TABLE `stock_requisition` (
...
   `date`                DATETIME NOT NULL,
   `user_id`             SMALLINT(5) UNSIGNED NOT NULL,
-  `reference`           INT(11) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `project_id`          SMALLINT(5) UNSIGNED NOT NULL,
+  `reference`           INT(11) UNSIGNED NOT NULL DEFAULT 0,
   `status_id`           TINYINT(3) UNSIGNED NOT NULL DEFAULT 1,
   `updated_at`          TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `created_at`          TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
-  PRIMARY KEY (`reference`),
+  PRIMARY KEY (`uuid`),
```
jniles, line 1944: 👍

server/models/schema.sql
```diff
@@ -1958,22 +1962,23 @@ CREATE TABLE `stock_requisition_item` (
...
+  `description`             TEXT NULL,
+  `flux_id`                 INT(11) NOT NULL,
+  `date`                    DATETIME NOT NULL,
+  `quantity`                INT(11) NOT NULL DEFAULT 0,
+  `unit_cost`               DECIMAL(19, 4) UNSIGNED NOT NULL,
+  `is_exit`                 TINYINT(1) NOT NULL,
+  `user_id`                 SMALLINT(5) UNSIGNED NOT NULL,
+  `reference`               INT(11) UNSIGNED NOT NULL,
+  `invoice_uuid`            BINARY(16) NULL,
+  `stock_requisition_uuid`  BINARY(16) NULL,
```
jniles, line 1979: In the long run, we could probably merge the columns for `invoice_uuid` and _`stock_requisition_uuid`.

jniles (approved):
This looks good to me.  All my previous comments have been addressed, and then some.  

Great work @lomamech!

bors r+

jniles:
Lgtm!  Thanks @lomamech .

Bors r+

jniles:
Congratulations @lomamech

PR #4923, opened 2020-09-15
chore: migrate bhChangeDepot to a component

### LINKED ISSUE #4787: Make bhChangeDepot into a component
We have a directive for changing depots called [`bhChangeDepot`](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/directives/bhChangeDepot.js).  In studying the code, there really isn't a good reason for it to be a directive instead of a component.  It would simplify the code (and make future migrations easier) if we went ahead and migrated this to a component.

jniles:
Changes the directive into a component for bhChangeDepot.

Closes #4787.


jeremielodi:
It seems that travis fails

jniles:
@jeremielodi yeah, I asked for a review too soon, I apologise.  I'll get this working and hit you up again.

jeremielodi (approved):
look_good_to_me

PR #4926, opened 2020-09-15
Feat(cmm) get value from stock_movement_status

### LINKED ISSUE #4879: Transition the "Items in Stock" registry to using the stock_movement_status table
Now that we've added the `stock_movement_status` table that provides the quantity in stock for each inventory item at a given time, we should transition the "items in stock" registry to use this table.  It should provide a fairly significant performance bump.

jeremielodi:
closes #4879

### DIFF
--- a/server/models/procedures/stock.sql
+++ b/server/models/procedures/stock.sql
@@ -375,7 +375,7 @@ CREATE PROCEDURE `computeStockQuantity` (
   SET _start_date = DATE(_start_date);
   SET @depot_number = 0;
 
-  DELETE FROM `temp_depot`;
+  DELETE FROM `temp_depot` WHERE 1;
 
   SET @filter_by_depot = (_depot_filter_uuid IS NOT NULL);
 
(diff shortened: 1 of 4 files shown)


server/models/procedures/stock.sql
```diff
@@ -528,70 +529,80 @@ CREATE PROCEDURE `getCMM` (
...
   JOIN inventory i ON m.inventory_uuid = i.uuid
   WHERE i.uuid = _inventory_uuid AND m.depot_uuid = _depot_uuid
   ORDER BY `end_date` DESC LIMIT 1;
 
 
-  SET @sum_consumed_quantity = 0;
-  SET @sum_stock_day = 0;
-  SET @sum_consumption_day =0;
-  SET @sum_stock_out_days = 0;
+  SET _sum_consumed_quantity = 0;
```
jniles, line 548: Nice - much better!

server/models/procedures/stock.sql
```diff
@@ -528,70 +529,80 @@ CREATE PROCEDURE `getCMM` (
...
+              ((m.start_date <= _start_date) AND  (m.end_date <= _end_date))
+					  )
           ORDER BY d.text
       ) AS `x`
   ) AS `cmm_data`;
 
 
 
-  SET @algo1 = ( @sum_consumed_quantity/(IF((@sum_stock_day = NULL) OR (@sum_stock_day = 0),1, @sum_stock_day)))*30.5;
+  SET @algo1 = ( _sum_consumed_quantity/(IF((_sum_stock_day = NULL) OR (_sum_stock_day = 0),1, _sum_stock_day)))*30.5;
```
jniles, line 590: Could you also replace these `@` variables with `_` variables?

server/models/procedures/stock.sql
```diff
@@ -528,70 +529,80 @@ CREATE PROCEDURE `getCMM` (
...
 
-  SET @algo2 = ((@sum_consumed_quantity)/(IF( (@sum_consumption_day = NULL) OR @sum_consumption_day = 0,1, @sum_consumption_day)))  * 30.5;
+  SET @algo2 = ((_sum_consumed_quantity)/(IF( (_sum_consumption_day = NULL) OR _sum_consumption_day = 0,1, _sum_consumption_day))) * 30.5;
```
jniles, line 592: I'm not sure `= NULL` works in SQL, but there is an [`IFNULL()`](https://www.w3schools.com/sql/func_mysql_ifnull.asp) function.

server/models/procedures/stock.sql
```diff
@@ -375,7 +375,8 @@ CREATE PROCEDURE `computeStockQuantity` (
   SET _start_date = DATE(_start_date);
   SET @depot_number = 0;
 
-  DELETE FROM `temp_depot`;
+  DROP TEMPORARY TABLE IF EXISTS `stock_movement_status`;
```
jniles: Do you need this in the live code, or with this a one-time fix for your machine?

If it is a one-time fix, I would prefer it be put in the `migrate.sql` file.

jniles (commented):
Looks good to me.  

I may be misunderstanding, but are we calculating the CMM twice?  First with the old way, then with the new way?  Do the values of the minimum security stock and things like that depend on the new or the old CMM?

server/models/procedures/stock.sql
jeremielodi: It works, even ```is NULL``` instead of ```= NULL``` works too

server/models/procedures/stock.sql
jeremielodi: I had an existing temporary table named ```stock_movement_status```.
this instraction is no important in production. I should remove this line. Thanks

server/models/procedures/stock.sql
jeremielodi: Not off hand. I declared them like variables before remplacing ```@``` by ```_```.

jeremielodi:
In this PR, I want first to get the CMM value, if it's okay then we can use it for other manipulations.

server/models/procedures/stock.sql
jniles: Interesting.  I suppose we will see in the tests if it behaves the same across different MySQL/MariaDB versions.

jniles (approved):
Looks good to me.   As mentioned in the review, we need to next update the other indicators to support this CMM calculation.

bors r+

PR #4927, opened 2020-09-18
feat(posting journal) unpost transactions

### LINKED ISSUE #4843: Implement zUnpostRecord() admin method
I was recently asked to correct some balances that were thoroughly messed up by a user.  It would be helpful if we had a `zUnpostRecord(record_uuid)` admin method hidden in the database to allow a (super)user to unpost records.  This is something that existed in SANRUTrakker to help clean up aberrant data.

It would be helpful to have this utility function kicking around for when we need to repair broken data in the future.

jeremielodi:
closes #4843

### DIFF
--- a/client/src/js/constants/bhConstants.js
+++ b/client/src/js/constants/bhConstants.js
@@ -12,6 +12,7 @@ function constantConfig() {
   return {
     actions : {
       CAN_EDIT_ROLES : 1,
+      CAN_UNPOST_TRANSACTIONS : 2,
     },
     accounts : {
       ROOT  : 0,
--- a/client/src/modules/journal/journal.html
+++ b/client/src/modules/journal/journal.html
@@ -66,7 +66,11 @@
             <i class="fa fa-balance-scale"></i> <span class="visible-lg-inline" translate>FORM.BUTTONS.TRIAL_BALANCE</span>
           </button>
         </div>
-
+        <div class="toolbar-item" ng-if="JournalCtrl.canUnpostTransactions">
+          <button class="btn btn-default" ng-click="JournalCtrl.unpostTransactions()"> <i class="fa fa-undo"></i>
+            <span translate>POSTING_JOURNAL.UNPOST</span>
+          </button>
+        </div>
         <bh-search-button on-click="JournalCtrl.openSearchModal()"></bh-search-button>
 
         <div class="toolbar-item">
@@ -85,6 +89,7 @@
             <i class="fa fa-object-group"></i> <span class="visible-lg-inline" translate>FORM.LABELS.TRANSACTION</span>
           </button>
         </div>
+
         <div class="toolbar-item">
           <button ng-class="{ 'btn-info' : JournalCtrl.gridOptions.enableFiltering }" class="btn btn-default" data-method="filter" ng-click="JournalCtrl.filtering.toggleInlineFiltering()">
             <i class="fa fa-filter"></i>
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -6,7 +6,7 @@ JournalController.$inject = [
   'SessionService', 'NotifyService', 'bhConstants', '$state', 'uiGridConstants', 'ModalService', 'LanguageService',
   'AppCache', 'Store', 'uiGridGroupingConstants', 'ExportService', '$filter', 'GridExportService',
   'GridStateService', 'GridSelectionService', 'TrialBalanceService', '$httpParamSerializer', 'TransactionService',
-  'util',
+  'util', 'RolesService',
 ];
 
 /**
@@ -33,7 +33,7 @@ function JournalController(
   Journal, Sorting, Grouping, Filtering, Columns, Session, Notify, bhConstants,
   $state, uiGridConstants, Modal, Languages, AppCache, Store, uiGridGroupingConstants,
   Export, $filter, GridExport, GridState, GridSelection, TrialBalance,
-  $httpParamSerializer, Transactions, util
+  $httpParamSerializer, Transactions, util, Roles,
 ) {
   // store journal data
   const journalStore = new Store({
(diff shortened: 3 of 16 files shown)


client/src/modules/journal/journal.html
```diff
@@ -66,7 +66,11 @@
             <i class="fa fa-balance-scale"></i> <span class="visible-lg-inline" translate>FORM.BUTTONS.TRIAL_BALANCE</span>
           </button>
         </div>
-
+        <div class="toolbar-item" ng-show="JournalCtrl.canUnpostTransactiosn">
```
jniles: Let's make this an `ng-if`.  It will not render the item in the DOM at all if it doesn't have the permission.  `ng-show` only hides it via CSS that can be undone by a smart user.

client/src/js/constants/bhConstants.js
```diff
@@ -12,6 +12,7 @@ function constantConfig() {
   return {
     actions : {
       CAN_EDIT_ROLES : 1,
+      CAN_UNPOST_TRANSACTIONS : 2,
```
jniles, line 15: Nice :+1:

client/src/modules/journal/journal.js
```diff
@@ -90,6 +91,14 @@ function JournalController(
       .catch(Notify.handleError);
   };
 
+  function checkUnpostTransactionsAllowability() {
+    Roles.userHasAction(bhConstants.actions.CAN_UNPOST_TRANSACTIONS)
+      .then(response => {
+        vm.canUnpostTransactiosn = response.data;
```
jniles: Let's set this to `false` at the beginning of the controller to be extra safe.

client/src/modules/journal/trial-balance/trial-balance.service.js
```diff
@@ -113,6 +114,15 @@ function TrialBalanceService(util, $http, Accounts, Constants) {
       });
   }
 
+  /**
+   * @function unpostTransactions
+   *
+   * @description
+   * This function attempts to post to the General Ledger by
```
jniles: I think this should read "attempts to unpost records from the general ledger to allow modification in the posting journal".

server/controllers/finance/trialBalance/index.js
```diff
@@ -97,3 +97,20 @@ exports.postToGeneralLedger = function postToGeneralLedger(req, res, next) {
     .catch(next)
     .done();
 };
+
+exports.unpostTransactions = (req, res, next) => {
+
+  const { recordUuids } = req.body;
```
jniles: Because this is a such a super user privilege, let's check the user privilege again here.  Look up the user ID and see if their role allows them to do this and send a HTTP error 403 if not.

server/models/migrations/next/migrate.sql
```diff
@@ -766,3 +766,11 @@ ALTER TABLE `enterprise_setting` ADD COLUMN `enable_supplier_credit` TINYINT(1)
  * @desc: add enterprise setting for strict depot permission which will restrict stock registries to depot owner
  */
  ALTER TABLE `enterprise_setting` ADD COLUMN `enable_strict_depot_permission` TINYINT(1) NOT NULL DEFAULT 0;
+
+/*
+ * @author: jeremielodi
+ * @date: 2020-09-18
+ * @desc: unpost transactions #4927
+ */
+INSERT INTO `actions`(`id`, `description`) VALUES
```
jniles, line 775: 👍

server/models/admin.sql
```diff
@@ -226,4 +226,18 @@ CREATE PROCEDURE zComputeAllInventoryStockQuantities (
...
+CREATE PROCEDURE zUnpostRecord(
+  IN _record_uuid BINARY(16)
+)
+BEGIN
+
+  INSERT INTO posting_journal 
+  SELECT * FROM general_ledger WHERE record_uuid = _record_uuid;
+
+  DELETE FROM general_ledger WHERE record_uuid = _record_uuid;
+
```
jniles, line 240: I discussed this with Mr. Wayne and he gave me a good suggestion- we should probably write this into a log somewhere.

We actually have a table for this, but it is only for editing.  It's called `transaction_history`.  [Here it is](https://github.com/IMA-WorldHealth/bhima/blob/master/server/models/schema.sql#L2063).  Let's enhance it by adding a column to tell us if we are editing or unposting.

jniles (commented):
@jeremielodi this is far more than I expected.  Great work!

I've made some comments on the code.  Let me know what you think of them.

client/src/modules/journal/journal.js
```diff
@@ -61,7 +61,9 @@ function JournalController(
   vm.enterprise = Session.enterprise;
   vm.languages = Languages;
   vm.gridApi = {};
+  vm.canUnpostTransactiosn = false;
```
jniles: This is good, thank you 👍 

There is a typo in the word: `canUnpostTransactiosn` -> `canUnpostTransactions`.

server/controllers/finance/trialBalance/index.js
```diff
@@ -97,3 +99,40 @@ exports.postToGeneralLedger = function postToGeneralLedger(req, res, next) {
...
+    const transaction = db.transaction();
+
+    recordUuids.forEach(recordUuid => {
+      transaction.addQuery('CALL zUnpostRecord(?)', db.bid(recordUuid));
+
+      transaction.addQuery('INSERT INTO transaction_history SET ?', {
+        uuid : db.uuid(),
+        record_uuid : db.bid(recordUuid),
+        user_id : req.session.user.id,
+        action : 'unpost',
```
jniles, line 116: Nice 👍

server/models/migrations/next/migrate.sql
```diff
@@ -766,3 +766,14 @@ ALTER TABLE `enterprise_setting` ADD COLUMN `enable_supplier_credit` TINYINT(1)
...
+  (2, 'FORM.LABELS.CAN_UNPOST_TRANSACTIONS');
+
+
+ALTER TABLE `transaction_history`  ADD COLUMN `action` VARCHAR(15) DEFAULT 'edit';
```
jniles, line 779: This is excellent.  Great work. 👍

jniles (approved):
@jeremielodi this looks fantastic.  I only have one comment about a typo.  After that, you can bors it.

jniles (approved):
LGTM.

bors r+

PR #4930, opened 2020-09-18
fix: upgrade ui-router

### LINKED ISSUE #2124: deps: Upgrade ui-router to the latest release
Our ui-router is stuck on a beta version of `v1.0.0`.  As of this issue, the [latest release of ui-router is v1.0.6](https://github.com/angular-ui/ui-router/releases).  It would be ideal if we could catch up with their versions, or at least put ourselves on a stable version.

jniles:
Upgrades ui-router to the latest version!

This provides significant advantages:
  1. The documentation for ui-router (ui-router.github.io) is now relevant.
  2. Removes the previous event-emitter pattern for state transitions.
  3. Greatly simplifies the authentication/install/login states - see client/app.js for the improvements.

Most of these changes are not major.  The vast majority of changes are in the routes that open a modal `onEnter`.  The newest version of UI router broke these routes because we use `$state.params` in those modals despite the fact that the state had not been reached.  In previous versions, this worked.  Now it doesn't.

To fix this, I migrated all routes that use modals to injecting a `params` resolve into the modal with the state parameters we are transitioning into.  It looks like of like this (pseudo-code):
```js
function onEnterModalOpen($modal, $transition) {

  $modal.open({
    /*stuff */
    resolve : { params : () => $transition.params('to');},
  });
}
```

I also took this opportunity to make the following global changes:
  1. I remove `creating`, `created`, `isCreating`, etc from states.  They are all now `isCreateState`. 
  2. I removed dead code that wasn't used.
  3. I added more comments and tightened up the `client/app.js` code.

I think this should land as soon so we can reap the benefits of the latest and greatest ui-router.

Closes https://github.com/IMA-WorldHealth/bhima/issues/2124
Closes https://github.com/IMA-WorldHealth/bhima/pull/4805


client/src/js/app.js
```diff
@@ -8,8 +8,7 @@ const bhima = angular.module('bhima', [
   'ui.grid.edit', 'ui.grid.grouping', 'ui.grid.treeView', 'ui.grid.cellNav',
   'ui.grid.pagination', 'ui.grid.moveColumns', 'ui.grid.exporter',
   'ui.grid.expandable', 'angularMoment', 'ngMessages',
-  'growlNotifications', 'ngAnimate', 'ngSanitize', 'ui.select', 'ngTouch',
-  'ui.router.state.events', 'webcam',
+  'growlNotifications', 'ngAnimate', 'ngSanitize', 'ui.select', 'ngTouch', 'webcam',
```
mbayopanda, line 11: 👍 no `ui.router.state.events`

mbayopanda (approved):
It was a big PR and the app pass tests without issues

Good job @jniles for the update

bors r+

PR #4934, opened 2020-09-19
Get starting with typescript in the serverside

mbayopanda:
This PR is a trial with typescript at server side of the BHIMA project

-----
[View rendered server_src/README.md](https://github.com/mbayopanda/bhima/blob/typescript-serverside-intro/server_src/README.md)


jniles:
@mbayopanda this looks like a great start!  As discussed in person, it would be ideal if we could get the typescript compiler to use the same configuration as the client compiler.

It would also be helpful if this were in the gulp script, similar to the client.  That way, we would have a common way of compiling between the server/client.

Finally, I don't know that we need to create a separate folder ... unless it is impossible to mix TS/JS together in the same folder.  I would much rather compile everything together and slowly replace one file by TS at a time.

mbayopanda:
@jniles changes landed

jniles:
@mbayopanda I fixed hound for you 😆

server/app.ts
```diff
@@ -38,8 +42,8 @@ const app = express();
  */
 function configureServer() {
   // destruct the environmental variables
-  const port = process.env.PORT;
-  const mode = process.env.NODE_ENV;
+  const port:string = process.env.PORT || '8080';
+  const mode:string = process.env.NODE_ENV || 'production';
```
jniles, line 44: Do these need to be `||` definitions?

I've made a small improvement where I moved the `dotenv` import to being `import dotenv/config` which automatically imports the variables according to [this link](https://github.com/motdotla/dotenv#how-do-i-use-dotenv-with-import).  Can we remove these default values?

jniles:
@mbayopanda can you switch this PR to a "draft" PR?

PR #4936, opened 2020-09-22
feat(inventory log) display correctly la chagement

### LINKED ISSUE #4933: Inventory Changes sometimes doesn't record the label
I'm not sure why, but the log of the inventory changes sometimes doesn't record the label correctly.  Here it is:

image

It would be useful to track down why this is.

jeremielodi:
closes #4933


jeremielodi:
changement

jniles (approved):
LGTM.

bors r+

PR #4944, opened 2020-09-22
Refactor depot management

### LINKED ISSUE #4922: Improvement of the depot management interface
With the option of being able to manage multiple depots, the interfaces that are currently in BHIMA were easy to use in this context.

Since we should manage several hyarachized repositories by province (CDR), BCZs and structures, this issue proposes to add a hierarchy of repositories, by adding the parent attribute to repositories, this will facilitate the display and management especially in the hypothesis or we will have more than 50 deposits to create

This advantage will also be necessary in the interface for assigning deposits to users, with this way of doing things, it will now be very easy to assign to a user, all the deposits of a health zone, of a province. unlike the way we do it now

lomamech:
- Adding property parent to depot
- Manage depots as a tree structure
- Using the tree structure to assign repositories to users
- Update migrate file

closes #4922


client/src/i18n/fr/depot.json
```diff
@@ -19,6 +20,7 @@
     "NO_DEPOT" : "Aucun dépôt",
     "NO_DEPOT_ASSIGNED_TO_USER" : "Vous n'avez pas la permission d'acceder à un dépôt ou aucun dépôt n'est enregistrer jusque là, veillez contacter l'administrateur en cas de besoin",
     "NOT_ALLOWED_ACTIONS" : "Ce dépôt est protégé, vous ne pouvez effectuer des modifications sur ce dépôt",
+    "SELECT_MAIN_DEPOT" : "Sélectionner le dépôt principale",
```
mbayopanda: `principal`

client/src/modules/depots/depots.html
```diff
@@ -29,7 +36,27 @@
...
+        empty-state="DepotCtrl.gridOptions.data.length === 0"
+        error-state="DepotCtrl.hasError">
+      </bh-grid-loading-indicator>
+    </div>
+  </div>
+
+  <div class="container-fluid">
+    <div 
+      id="depot-grid"
+      ng-if="DepotCtrl.listStructure"
```
mbayopanda: Is it necessary to have both grid ?

We could just keep the tree version of the grid

client/src/modules/depots/depots.js
```diff
@@ -90,6 +99,12 @@ function DepotManagementController(
     vm.gridApi.core.notifyDataChange(uiGridConstants.dataChange.COLUMN);
   }
 
+  function toggleTreeMode() {
+    vm.treeStructure = !vm.treeStructure;
+    vm.listStructure = !vm.listStructure;
```
mbayopanda: Since `vm.listStructure` is the reverse of `vm.treeStructure` could we just have `vm.treeStructure` and handle its reverse as `!vm.treeStructure`

server/models/schema.sql
```diff
@@ -461,6 +461,7 @@ CREATE TABLE `depot` (
   `allow_exit_loss` TINYINT(1) UNSIGNED NOT NULL DEFAULT 0,
   `location_uuid` BINARY(16) NULL,
   `min_months_security_stock` SMALLINT(5) NOT NULL DEFAULT 2,
+  `parent` BINARY(16) NULL,
```
mbayopanda: Is it possible to rename the column `parent_uuid` since its must store the uuid of the parent

test/end-to-end/depots/depots.page.js
```diff
@@ -39,6 +39,29 @@ class DepotPage {
...
   }
 
+  /**
+   * simulate the create depot button click to show the dialog of creation
+   * @param {string} item the name of the depot to create and parent {depot and parent}
+   * @param {boolean} hasWarehouse
+   * @param {boolean} hasLocation if true it will enable the option of adding location
+   * @param {array} location an array of location as [country_uuid, province_uuid, sector_uuid, village_uuid]
+   */
+  async createDepotByParent(item, hasWarehouse, hasLocation, location) {
```
mbayopanda:
```js
async createDepotByParent(item, isWarehouse, hasLocation, location) {
```

mbayopanda (commented):
The feature works well 👍 

I have just some minors comments

client/src/i18n/fr/depot.json
```diff
@@ -1,6 +1,7 @@
 {
   "DEPOT" : {
     "ADD_DEPOT" : "Ajouter un dépôt",
+    "ADD_DEPENDANT_DEPOT": "Ajouter un dépôt dépendant",
```
jmcameron: Correct spelling is:  DEPENDENT   (not -ANT)

jmcameron (approved):
I tested its  behavior and it seems to work fine.  However, I found the user interface a little bit odd.   I do not think you need a "Select Main Depot" checkbox.   Just have a box that says "Parent Depot".  It would default to "None" but clicking in it would allow the selection of a parent:

image

Also, while you're at it, I suggest changing the label "Join location" to "Set location" on the option above.   You will need to update the French label appropriately too.

jmcameron:
I read the Issue for this PR and I understand one of the initial goals is to all user access to a tree of Depots.  That sounds reasonable. However, I wonder if it could be used in other ways.  For instance to automatically generate some kind of composite report for a depot and all children?

jmcameron:
One further thought: Do we need to keep the "List" structure option?  Why not just make the Tree structure the default and only view?    If there are no parents, it will look like a list.

lomamech:
> One further thought: Do we need to keep the "List" structure option? Why not just make the Tree structure the default and only view? If there are no parents, it will look like a list.

The reason is as follows, when searching for a repository, only repositories that do not have parents appear, the algorithm that is used in the ````FormatTreeData.formatStore()````
 function first loads repositories whose parent attribute is 0

lomamech:
> > One further thought: Do we need to keep the "List" structure option? Why not just make the Tree structure the default and only view? If there are no parents, it will look like a list.
> 
> The reason is as follows, when searching for a repository, only repositories that do not have parents appear, the algorithm that is used in the `FormatTreeData.formatStore()`
> function first loads repositories whose parent attribute is 0

Here is an illustration with the database that I loaded with the repositories created from the DR Congo Health Pyramid

9ukWBQZ9LO

jmcameron:
> The reason is as follows, when searching for a repository, only repositories that do not have parents appear, 
> ...

Why should this be the case?    Even in a tree structure, don't you want to be able to search for a specific depot even it if is not "at the top"?   If this is a limitation of FormatTreeData, perhaps it needs to be updated to allow searching all?

My concern here is from the user interaction standpoint:  Why have two different listing styles when one might be enough?   When should a user use one vs the other?  I think having list/tree is confusing for users so I would encourage using just the tree format, if possible.

lomamech:
> > The reason is as follows, when searching for a repository, only repositories that do not have parents appear,
> > ...
> 
> Why should this be the case? Even in a tree structure, don't you want to be able to search for a specific depot even it if is not "at the top"? If this is a limitation of FormatTreeData, perhaps it needs to be updated to allow searching all?
> 
> My concern here is from the user interaction standpoint: Why have two different listing styles when one might be enough? When should a user use one vs the other? I think having list/tree is confusing for users so I would encourage using just the tree format, if possible.

I understood, I will keep the architecture in tree structure and I will add an issue, to resolve this limitation on the use of the tree structure when filtering an element of the tree

lomamech:
@jniles  @jmcameron  @mbayopanda  all checks have passed

jmcameron (approved):
I tested this and was pleased with the way it works now.    Thanks for making the changes.   I ran into one small issue:  When you have a Depot with a parent and you edit it, if you do the "Clear" for it, nothing happens.  I think it should remove the parent.  Then clicking the [Submit] button would save that change.

Also, a cosmetic suggestion:  I suggest (if possible) to make tree default to all branches/leaves open, not just showing the top level.   The user could always collapse the branches as desired.  I was not 100% sure about this suggestion.  Do most possible installations deal with a lot of depots?   If the answer is yes, then maybe the way it is now would be best.  

Finally, other than the minor issue with the "Clear" button, I think this is ready to go.  However, you may want to check with @jniles about whether to hold off on releasing this until after the major release that he is working on now.

lomamech:
> I tested this and was pleased with the way it works now. Thanks for making the changes. I ran into one small issue: When you have a Depot with a parent and you edit it, if you do the "Clear" for it, nothing happens. I think it should remove the parent. Then clicking the [Submit] button would save that change.
> 
> Also, a cosmetic suggestion: I suggest (if possible) to make tree default to all branches/leaves open, not just showing the top level. The user could always collapse the branches as desired. I was not 100% sure about this suggestion. Do most possible installations deal with a lot of depots? If the answer is yes, then maybe the way it is now would be best.
> 
> Finally, other than the minor issue with the "Clear" button, I think this is ready to go. However, you may want to check with @jniles about whether to hold off on releasing this until after the major release that he is working on now.

Thank you I will make the tree view open but also fixed the deleted button, it escaped me after the last update, now I will add a test for that

mbayopanda (approved):
The feature is working fine.

I have just noticed that building tree on the client with the limit filter has some bad user experience, if the limited loaded depots doesn't have root parent they will not be displayed in the depot registry.

client/src/modules/depots/templates/label.tmpl.html
```diff
@@ -5,6 +5,6 @@
       { key : 'period', value : 'allTime', displayValue : 'PERIODS.ALL_TIME', cacheable: false },
       { key : 'limit', value : 1000, cacheable: false }
     ]})" href>
-    <span>{{ row.entity.text }}</span>
+    <span style='padding-left : {{row.treeLevel * grid.appScope.indentTitleSpace}}px;' >{{ row.entity.text }}</span>
```
jniles, line 8: Could you use [`ngStyle`](https://docs.angularjs.org/api/ng/directive/ngStyle) here?

client/src/i18n/en/depot.json
```diff
@@ -1,6 +1,7 @@
 {
   "DEPOT" : {
     "ADD_DEPOT" : "Add a Depot",
+    "ADD_DEPENDENT_DEPOT": "Add dependent Depot",
```
jniles, line 4: A better phrase would be either "Add a sub depot" or "Add a child depot".

client/src/modules/depots/depots.js
```diff
@@ -121,13 +133,23 @@ function DepotManagementController(
...
+            item.parent = 0;
+          }
+
           item.location = item.location_uuid
             ? ''.concat(`${item.village_name} / ${item.sector_name} / ${item.province_name} `)
               .concat(`(${item.country_name})`) : '';
           return item;
         });
+
+        vm.gridOptions.data = FormatTreeData.formatStore(depotsData);
```
jniles, line 157: Cool 👍

client/src/modules/depots/templates/action.tmpl.html
```diff
@@ -22,6 +22,13 @@
         <i class="fa fa-th-list"></i> <span translate>INVENTORY.VIEW_ARTICLES_IN_STOCK</span>
       </a>
     </li>
+    <li ng-if="row.entity.is_warehouse">
```
jniles, line 25: Aha, this is interesting.  Only warehouses can have child depots.

... that makes sense to me.  I'll be curious to see if that remains true throughout the project :).

client/src/modules/users/UserDepotManagementModal.html
```diff
@@ -19,14 +19,37 @@ <h4><i> {{ UsersDepotModalCtrl.user.display_name }} </i></h4>
...
+      <div style="margin-left: calc({{depot.$$treeLevel}} * 15px)" class="list-unstyled">
+        <div class="checkbox">
+          <label data-label="{{child._label}}">
+            <input type="checkbox" id="depot_{{$index}}" ng-model="depot._checked" ng-change="UsersDepotModalCtrl.setNodeValue(depot.children, depot)" />
+            <span translate>{{depot.text}}</span>
+          </label>
+        </div>
+      </div>
+    </div>
+    <div ng-if="UsersDepotModalCtrl.loading" class="container text-center">
```
jniles: I don't think this should be `class="container"`.  Maybe you mean `row`?

jniles:
Bors try

client/src/modules/users/UserDepotManagementModal.html
```diff
@@ -19,14 +19,37 @@ <h4><i> {{ UsersDepotModalCtrl.user.display_name }} </i></h4>
...
     <span translate> FORM.INFO.USER_DEPOT </span>
+    <div class="checkbox">
+      <label data-root-node>
+        <input type="checkbox" ng-model="UsersDepotModalCtrl.all" ng-change="UsersDepotModalCtrl.setAllNodeValue(UsersDepotModalCtrl.depotsData, UsersDepotModalCtrl.all)" />
+        <strong class="text-capitalize" translate>FORM.LABELS.CHECK_ALL</strong>
+      </label>
+    </div>
+    
+    <div class="panel panel-default" style="margin-bottom : 0px;">
+      <div class="panel-heading clearfix" style="text-align : left">
```
jniles: Do you need the `clearfix` class here?

jniles: Also, instead of using both class and style, you can also use the class `text-left`.

client/src/modules/users/UserDepotManagementModal.html
```diff
@@ -19,14 +19,37 @@ <h4><i> {{ UsersDepotModalCtrl.user.display_name }} </i></h4>
...
-      on-change="UsersDepotModalCtrl.onDepotChange(depots)">
-    </bh-depot-search-select>
+      <input ng-model="UsersDepotModalCtrl.filter" ng-show="UsersDepotModalCtrl.filterActive" class="form-control" placeholder="{{'FORM.PLACEHOLDERS.FILTER_NAME' | translate }}" style="border-radius: 0"/>
+    </div>
 
+    <div ng-if="!UsersDepotModalCtrl.loading" ng-repeat="depot in UsersDepotModalCtrl.depotsData | filter : { text : UsersDepotModalCtrl.filter }">
+      <div style="margin-left: calc({{depot.$$treeLevel}} * 15px)" class="list-unstyled">
+        <div class="checkbox">
+          <label data-label="{{child._label}}">
+            <input type="checkbox" id="depot_{{$index}}" ng-model="depot._checked" ng-change="UsersDepotModalCtrl.setNodeValue(depot.children, depot)" />
```
jniles: Instead of `id=depot_{{$index}}`, why don't use use the depot uuid?  That way, it is guaranteed to be unique and you can refer to the depot by its UUID directly.

client/src/modules/users/UserDepotManagementModal.js
```diff
@@ -19,29 +20,97 @@ function UsersDepotManagementController($state, Users, Notify, AppCache, params)
...
   vm.submit = submit;
   vm.closeModal = closeModal;
+  vm.toggleFilter = toggleFilter;
+  vm.loading = true;
 
   vm.onDepotChange = (depots) => {
     vm.user.depots = depots;
   };
 
+  function setNodeValue(childrens, depot) {
```
jniles, line 36: Hmm... we have a lot of the same functionality in the `bhCheckboxTree`.  I know you tried to use that already, so you don't need to do anything now.

But in a future PR, we should probably add the functionality you are looking for to the `bhCheckboxTree`.  That way, we can test the component to make sure it works correctly.

server/controllers/inventory/depots/index.js
```diff
@@ -90,7 +105,7 @@ function update(req, res, next) {
         SELECT BUID(uuid) as uuid, text, enterprise_id, is_warehouse,
           allow_entry_purchase, allow_entry_donation, allow_entry_integration, allow_entry_transfer,
           allow_exit_debtor, allow_exit_service, allow_exit_transfer, allow_exit_loss,
-          min_months_security_stock
+          min_months_security_stock, IF(parent_uuid IS NULL, 0, BUID(parent_uuid)) as parent_uuid
```
jniles, line 108: Just FYI - there is actually a function for this: [`IFNULL(x, y)`](https://www.w3schools.com/sql/func_mysql_ifnull.asp).

server/models/schema.sql
```diff
@@ -461,6 +461,7 @@ CREATE TABLE `depot` (
   `allow_exit_loss` TINYINT(1) UNSIGNED NOT NULL DEFAULT 0,
   `location_uuid` BINARY(16) NULL,
   `min_months_security_stock` SMALLINT(5) NOT NULL DEFAULT 2,
+  `parent_uuid` BINARY(16) NULL,
```
jniles, line 464: Let's add an INDEX on this so allow us to make faster lookups in the future.

jniles (commented):
@lomamech, this looks pretty good.  I haven't tested the functionality, but made some suggestions to clean up the code.  Most of these comments are non-blocking, but a few could use some touching up.

Once this gets rebased and those comments addressed, we can probably merge it any time.

client/src/modules/depots/templates/label.tmpl.html
lomamech: it generates error messages, maybe because there are several concatenations

client/src/modules/users/UserDepotManagementModal.html
lomamech: It was only a trick with for the **end to end** test

client/src/modules/users/UserDepotManagementModal.js
lomamech: Ok

lomamech:

client/src/modules/depots/templates/action.tmpl.html
jmcameron: > Aha, this is interesting. Only warehouses can have child depots.

Is that true?  Don't even pharmacies in a hospital act as a small depot?  And a depot in a hospital would be the parent for that child pharmacy, but would not be a warehouse, right?

client/src/modules/users/UserDepotManagementModal.html
jniles: As mentioned in person, if you could use an identifier from the depot, rather than it's placement in the array, this will save a lot of test errors in the future.

jniles:
Everything here has been addressed.  Thanks @lomamech!

bors r+

PR #4946, opened 2020-09-22
Stock settings page

### LINKED ISSUE #4803: Make a configuration page for Stock Management variables
We have quite a few Stock Management specific enterprise variables.  Since we are going to look at breaking out Stock Management in the long run, I think it makes sense by putting these in their own configuration page.  That configuration page would go under the "Stock" folder in the tree.

Specifically, here are the settings we currently have that could migrate over:

1. (`enable_auto_stock_accounting`) Automatically create stock movement financial transactions.
2. (`month_average_consumption`) Number of days for the AMC / CMM calculation 
3. (`enable_daily_consumption`) Enable daily consumption calculation for the CMM
4. (`enable_auto_purchase_order_confirmation`) Enable automatic confirmation of purchase orders

Any new proposals (like #4802) would go under this heading as well.

We could either continue to store this information in the `enterprise_setting` table or make a custom table for stock settings.

jmcameron:
Convert stock-related settings in Enterprise settings to their own page under the Stock menu.

The settings that were moved are:
1.  month_average_consumption
2.  default_min_months_security_stock
3.  enable_auto_purchase_order_confirmation
4.  enable_auto_stock_accounting
5.  enable_daily_consumption
6.  enable_strict_depot_permission
7.  enable_supplier_credit

**Suggested testing sequence**
- Use bhima_test (edit .env as needed)
- yarn build:db
- yarn dev
- Verify that there are no stock-related settings in the Enterprise settings
- Go to the "Stock Settings" under the stock menu
  - Change all the values then press [Submit].  Notice the success indicator on
    the top of the page
  - Do a full refresh of the page.  Verify that the values have not changed.
  - Change all the values then press [Submit]. 
  - Do a full refresh of the page.  Verify that the values have not changed.
  - Note the numeric values for month_average_consumption and
    default_min_months_security_stock (MMSS below)
- Go to Stock > Depot Management
  - Edit each depot and verify that MMSS is:
     - Depot Principal: 2
     - Depot Secondaire: 3
  - Start to create a depot
     - Notice that MMSS defaults to the last value set above
	 - Change MMSS value in Stock Settings and repeat above

The update also supports creating default stock settings if no entry exists
already.  This might occur in new installations.   Verify that it works like
this:
  - Notice defaults for stock settings in server/model/schema.sql
     - Look around line 1860 for: CREATE TABLE `stock_setting`
     - Note the default for enable_auto_stock_accounting in the SQL table is 1,
	but is 0 in the Bhima_test database.
  - Edit test/bhima.sql and temporarily disable: line 3000
     - Disable line: -- INSERT INTO `stock_setting` ...
  - Do 'yarn build:db'
    - Run mysql and verify that the stock_setting table is empty
  - Run 'yarn dev'
  - Refresh browser with BHIMA session.
  - On the server console look for message:
	  app Created default stock_setting for enterprise 1! +0ms
  - Verify that the stock settings revert to defaults


client/src/modules/stock/settings/stock-settings.html
```diff
@@ -0,0 +1,78 @@
+<div class="flex-header">
+  <div class="bhima-title">
+    <ol class="headercrumb">
+      <li class="static" translate>TREE.STOCK</li>
+      <li class="title" translate>TREE.STOCK_SETTINGS</li>
+    </ol>
+    <div class="flex-content">
```
jniles: You need to close the `<div class="flex-header">` before you can put the flex content.  You'll get very interesting results otherwise!

server/models/schema.sql
```diff
@@ -1856,6 +1856,17 @@ CREATE TABLE IF NOT EXISTS `voucher_item` (
 
 -- stock tables
 
+DROP TABLE IF EXISTS `stock_setting`;
+CREATE TABLE `stock_setting` (
```
jniles, line 1853: Excellent! You'll need to add this to `migrate.sql` when you are ready.

jniles (commented):
Just left a few (hopefully helpful) comments.  This is super exciting!

client/src/i18n/en/form.json
```diff
@@ -398,7 +398,7 @@
             "DEFAULT_QUANTITY": "Default Quantity",
             "DEFAULTS": "Default Filters",
             "DEFINE_UNTIL_DATE": "Define a limit date",
-            "DEFINE_MONTH_AVERAGE_CONSUMPTION": "Number of months for calculating the average monthly consumption",
+            "DEFINE_MONTH_AVERAGE_CONSUMPTION": "??? Number of months for calculating the average monthly consumption",
```
jniles: Do these ??? mean you don't understand how it is used or was this just to track down a bug?

In case you don't understand how it is used, this function defines the window size we use to calculate the average.  If it is 2 months, it is only the last 2 months of consumption considered.  If it is 12, it's only the last year.  And so forth.

client/src/modules/stock/settings/stock-settings.html
```diff
@@ -0,0 +1,93 @@
...
+      <div class="panel panel-default">
+        <div class="panel-heading">
+          <i class="fa fa-cogs"></i> <span translate>TREE.STOCK_SETTINGS</span>
+        </div>
+        <div class="panel-body">
+
+          <div class="form-group"
+            ng-class="{ 'has-error' : StockSettingsForm.$submitted && StockSettingsForm.month_average_consumption.$invalid }">
+            <label class="control-label" translate>
+              SETTINGS.DEFINE_MONTH_AVERAGE_CONSUMPTION
```
jniles, line 23: I would still put these settings under the "STOCK" namespace.  So either: `STOCK.SETTINGS.XYZ` or `SETTINGS.STOCK.XYZ`.  I know that I'll be grep-ing for the word "stock" at some point looking for them and be surprised they are just in the settings which have typically been client-specific.

test/data.sql
```diff
@@ -2996,6 +2996,10 @@ INSERT INTO `lot` (`uuid`, `label`, `initial_quantity`, `quantity`, `unit_cost`,
   (HUID('ef24cf1a-d5b9-4846-b70c-520e601c1ea6'), 'QUININE-C', 50, 50, 2.0000, DATE_ADD(CURRENT_DATE, INTERVAL 2 YEAR), @quinine, HUID('e07ceadc-82cf-4ae2-958a-6f6a78c87588'), 0, '2017-02-02 11:09:25'),
   (HUID('aca917fe-5320-4c3c-bea6-590e48cfa26b'), 'ERYTHRO-A',	10, 0,	3.1800,	DATE_ADD(CURRENT_DATE, INTERVAL 3 MONTH), @erythromycine, HUID('1908da7a-7892-48d7-a924-be647e5215ef'),	0, '2020-09-04 15:20:22');
 
+-- stock settings (go with defaults)
+-- INSERT INTO `stock_setting` (`enterprise_id`) VALUES (1);
+-- ??? Disabled during testing
```
jniles: I have finally understood what the `???` means.   This is a cool system to find touched code 👍

server/models/bhima.sql
```diff
@@ -162,7 +162,8 @@ INSERT INTO unit VALUES
   (286,'Fee Center Reports','TREE.REPORTS','reports for the fee center modules',218,'/fee_center/reports'),
   (287,'Inventory Reports','TREE.REPORTS','reports for the inventory modules', 138,'/inventory/reports'),
   (288, '[Stock] Movement Report','TREE.STOCK_MOVEMENT_REPORT','Stock Movement Report', 282,'/reports/stock_movement_report'),
-  (289, '[Stock] Expiration report','TREE.STOCK_EXPIRATION_REPORT','Stock expiration report', 282,'/reports/stock_expiration_report');
+  (289, '[Stock] Expiration report','TREE.STOCK_EXPIRATION_REPORT','Stock expiration report', 282,'/reports/stock_expiration_report'),
+  (290, '[Stock] Settings', 'TREE.STOCK_SETTINGS', 'Stock Settings', 160, '/stock/setting');
```
jniles, line 166: Excellent 👍

server/controllers/stock/setting/index.js
```diff
@@ -0,0 +1,69 @@
...
+const NotFound = require('../../../lib/errors/NotFound');
+// ??? const BadRequest = require('../../../lib/errors/BadRequest');
+
+// GET /stock/setting
+//
+// Get the current stock settings for the Enterprise
+//    If req.query.enterprise_id is set, it will use that,
+//    otherwise it will look up the entry for Enterprise.id=1
+exports.list = function list(req, res, next) {
+  const enterpriseId = req.query.enterprise_id || '1';
```
jniles: So, we don't have a great way of handling this case, but most other places in the app just list all the records.  Ironically, because of our inexperience, quite a few of our production enterprise ids are 200, not one.

You can actually get the current enterprise id from `req.session.enterprise.id`.  I propose you do that.

Finally, you should escape this identifier.  Change `WHERE enterprise_id = ${enterpriseId} LIMIT 1;` to `WHERE enterprise_id = ? LIMIT 1;`.  Then pass it in to `db.exec(sql, [enterpriseId])` and that should do the trick.

client/src/modules/stock/stock.routes.js
```diff
@@ -96,6 +96,14 @@ angular.module('bhima.routes')
         },
         onEnter : ['$state', 'StockModalService', onEnterFactory('create', 'stockRequisition')],
         onExit : ['$uibModalStack', closeModals],
+      })
+
+      .state('stockSetting', {
+        url         : '/stock/setting',
+        controller  : 'StockSettingsController as StockSettingsCtrl',
+        templateUrl : 'modules/stock/settings/stock-settings.html',
```
jniles, line 104: Nice hyphenated naming 👍

server/controllers/stock/setting/index.js
```diff
@@ -0,0 +1,69 @@
...
+  db.exec(sql)
+    .then(rows => {
+      res.status(200).json(rows);
+    })
+    .catch(next)
+    .done();
+};
+
+// POST /stock/setting
+exports.create = function create(req, res, next) {
```
jniles: Interesting... for this, we probably don't have a create route, eh?  Do we even need this?

client/src/modules/stock/settings/stock-settings.js
```diff
@@ -0,0 +1,119 @@
...
+      return 0;
+    }
+
+    // make sure only fresh data is sent to the server.
+    if (form.$pristine && !$touched) {
+      Notify.warn('FORM.WARNINGS.NO_CHANGES');
+      return 0;
+    }
+
+    const changes = util.filterFormElements(form, true);
```
jniles, line 61: So ... this is magic.  It's magic even to me.  I don't think you need it.

I'll post more in detail about it later, when I've had a chance to refresh memory on what it does.  But because you are using `$touched` here, I would just remove it.

jniles: Yeah, I would just use `vm.settings` and send that back to the server to update the settings.

jniles (commented):
@jmcameron this is really great work.  I'm glad to see all the progress made!

I've left some guiding comments, but overall it is super well done.  Let me know if you feel like you need more guidance than this though and I'll leave some more comments!

client/src/modules/stock/settings/stock-settings.html
jmcameron: These are actually in STOCK.SETTINGS.xyz (in stock.json).  However, if I include the STOCK. prefix it cannot find them.  Not sure how to fix that.

test/data.sql
jmcameron: Yes.  I'll clean up all the ??? before the final push!

client/src/i18n/en/form.json
jmcameron: Just reminders of things to clean up.

server/controllers/stock/setting/index.js
jmcameron: Thanks.  I actually thought of that but could not decide if that was "canonical".

client/src/modules/stock/settings/stock-settings.js
jmcameron: This is essentially the code that was in the corresponding function for the Enterprise settings (enterprises.js, lines 107+).   Please point me to an example that does it the way you suggest.  I tried changing this per your suggestion and it broke things.

server/controllers/stock/setting/index.js
jmcameron: Route removed

server/controllers/stock/setting/index.js
jmcameron: I did what you said, but also supported passing the ID in the GET request.  If the parameter is not set, it uses the value from the session.

server/models/schema.sql
jmcameron: Done.

jmcameron:
@jniles I think this is almost ready.   On my system, test:integration runs fine.  I could not figure out why it is failing under semaphore.  I need some help on that!

jmcameron:
**Congratulations to everyone!**  The test suite may be a pain but it is very helpful!  

I thought I had this PR ready to go this morning, but the integration tests and end-to-end tests helped me find several things that I had overlooked!  It took a bit to track down the problems and fix them, but the failing tests really helped me narrow the problems down.   Also, some of the tests that failed did not seem to be related to the changes in this PR, but once I investigated more carefully, I realized that I had missed doing some needed updates to the session data.

jniles:
Excellent!  Glad to hear it.

jmcameron:
bors try

jmcameron:
I fixed the integration test and I got it to pass under semaphore.  I think we are now in the random failure mode (with MySQL 8) with this error:  "POST /journal/transactions posts the a transaction to general_ledger and removes it from the journal" which we also see in PR https://github.com/IMA-WorldHealth/bhima/pull/4965.

jmcameron:
Bors try

lomamech (approved):
I just tested the features and it works great, and have no comments, please do a rebase with the master branch in the hope that the tests pass, by the time the tests pass i will merged it

jmcameron:
> I just tested the features and it works great, and have no comments, please do a rebase 
> with the master branch in the hope that the tests pass, by the time the tests pass i will merged it

Thanks @lomamech .  I'm waiting to see what happens on the integration test.  Once that is done, I'll rebase, per your suggestion.  However, I suggest you consult with @jniles about releasing it.  I think he wanted to release another PR first.  If that is the case, I will wait until that one lands before I do the rebase with master.

jniles:
@jmcameron @lomamech 

This PR looks really good.  I suggest that we merge https://github.com/IMA-WorldHealth/bhima/pull/4921 first, if possible.  It is large and more complicated.  It will be easier to fix this PR after that one is merged, than vice versa.

jmcameron:
@lomamech .   @jniles wants to merge 4921 first.
Thanks anyway!

lomamech:
> @lomamech . @jniles wants to merge 4921 first.
> Thanks anyway!

Sorry, I have not well understood

jmcameron:
Ok. I see 4921 got merged. I'll update and rebase this one after lunch (here).

jniles:
Thanks so much @jmcameron

jniles:
LGTM.  Great work!

bors r+

PR #4949, opened 2020-09-23
feat(stock e2e tests) pull all tests in one block

### LINKED ISSUE #4948: Running stock end-to-end test in one block
In other to facilitate stock end-to-end test checking locally, I propose that we could combine all stock e2e test in one ```describe```.
That will also back the idea of running stock module separately.

jeremielodi:
closes #4948

### DIFF
--- a/test/end-to-end/stock/stock.entry.js
+++ b/test/end-to-end/stock/stock.entry.js
@@ -89,6 +89,7 @@ function StockEntryTests() {
 
     await page.setLots(0, lots, true);
 
+    await this.timeout(600000);
     // submit
     await page.submit();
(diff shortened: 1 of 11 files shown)


test/end-to-end/stock/stock.entry.js
```diff
@@ -89,6 +89,7 @@ function StockEntryTests() {
 
     await page.setLots(0, lots, true);
 
+    await this.timeout(600000);
```
jniles, line 92: This probably isn't required.

jniles (approved):
Looks like a good change to me!

bors r+

PR #4950, opened 2020-09-24
Fix(unit list) put indicator report in dashboard folder

### LINKED ISSUE #4941: Indicators Report should be under "Dashboards" folder in the tree
There is a bug in the unit table in that the "indicators report" should actually be classified under "dashboards".  It is currently in the reports field of the Fee Centers section.  We should put it where it belongs in the Dashboards folder.

jeremielodi:
closes #4941


jniles (approved):
LGTM.

bors r+

PR #4951, opened 2020-09-24
fix(stock): do not show icon for stock out

### LINKED ISSUE #4943: Stock Lots warning column should not show check mark for exhausted lots
On the Stock > Items in Stock, the new warning column (to the left of the "Status" column) shows a warning icon for nearly exhausted items, etc.   However, if the item is completely exhausted (quantity=0), a check mark is shown with a tooltip that is somewhat confusing ("No lots with expiration risk").  See below.  The check mark should be suppressed for exhausted items or converted to a warning icon.  We already have the red "Stock Out" label pill in this case, so perhaps a warning icon is not necessary (see the second image below).

image

image

jniles:
This commit changes the display value of the icon to no longer show a green check mark when there is a stock out.

Closes #4943.

This is what it looks like:
image

### DIFF
--- a/client/src/modules/stock/inventories/registry.js
+++ b/client/src/modules/stock/inventories/registry.js
@@ -198,7 +198,7 @@ function StockInventoriesController(
     item.alert = item.hasExpiredLots;
     item.warning = !item.hasExpiredLots && (item.hasNearExpireLots || item.hasRiskyLots);
 
-    item.isSoldOut = item.status === bhConstants.stockStatus.IS_STOCK_OUT;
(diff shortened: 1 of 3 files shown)


jniles:
@jeremielodi ou @jmcameron pouvez-vous jeter un a coup d’œil sur ça?

jmcameron (approved):
The code looks fine.  And I tested it and it works as advertised.  I think it is ready to release.

jniles:
Thanks @jmcameron 

bors r+

PR #4958, opened 2020-09-26
fix(vouchers): document_uuid on the payment voucher

jniles:
Adds the document_uuid of the payment to the payment voucher.

Fixes a bug in production.


jniles:
@lomamech can you confirm this works correctly?

lomamech (commented):
it works fine, i tried partial payment as well as full payment, it's ok

PR #4968, opened 2020-09-29
fix: allow multiple debtor invoice selections

jniles:
This commit fixes the cash invoice selection modal by using `$transition` instead of `$state` params.  For some reason, selecting the `$state` multiple times does not work.

It's bug that is hitting our production systems.  We should do a quick check to make sure this isn't affecting more modals.  It's a tricky bug since it works the first time (and so, passes tests) but doesn't work after that.

### DIFF
--- a/client/src/modules/cash/cash.routes.js
+++ b/client/src/modules/cash/cash.routes.js
@@ -46,7 +46,7 @@ angular.module('bhima.routes')
           debtor_uuid : { value : undefined },
           invoices    : { value : [] },
         },
-        onEnter : ['$state', '$uibModal', debtorInvoicesModal],
+        onEnter : ['$state', '$uibModal', '$transition$', debtorInvoicesModal],
         onExit  : ['$uibModalStack', closeModal],
       })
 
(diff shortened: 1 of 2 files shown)


client/src/modules/cash/cash.routes.js
```diff
@@ -46,7 +46,7 @@ angular.module('bhima.routes')
           debtor_uuid : { value : undefined },
           invoices    : { value : [] },
         },
-        onEnter : ['$state', '$uibModal', debtorInvoicesModal],
+        onEnter : ['$state', '$uibModal', '$transition$', debtorInvoicesModal],
```
mbayopanda, line 49: 👍

client/src/modules/cash/modals/invoice-modal.ctrl.js
```diff
@@ -33,12 +33,12 @@ function CashInvoiceModalController(Debtors, Session, $timeout, Notify, $state,
...
     flatEntityAccess  : true,
-    onRegisterApi     : onRegisterApi,
+    onRegisterApi,
     enableColumnMenus : false,
     columnDefs        : [
-      { name: 'reference' },
-      { name: 'balance', cellFilter: 'currency:' + Session.enterprise.currencyId},
-      { name: 'date', cellFilter: 'date' },
+      { name : 'reference' },
+      { name : 'balance', cellFilter : `currency:${Session.enterprise.currencyId}` },
```
mbayopanda, line 40: 👍

mbayopanda (approved):
LGTM 👍 

bors r+

PR #4975, opened 2020-10-01
ci: use latest chrome for E2E tests

jniles:
Updates travis CI to no longer filter on a particular chromedriver. Should use the latest chrome browser for tests.


jmcameron (approved):
This seems like a safe PR.  But I have no way to test it.  
I suggest pushing it through and then try doing 'bors try' on some of our pending PRs to see what happens.  We can always do another PR to undo this if necessary.

jniles:
bors try

jniles:
Bors try

jniles:
bors try

PR #4979, opened 2020-10-02
Fixed New Requistions modal cancel

### LINKED ISSUE #4974: Do not refresh requisitions grid if the modal for new requisition has been canceled
I noticed this while reviewing https://github.com/IMA-WorldHealth/bhima/pull/4921, but it didn't have to do with that PR.

When you click the "new requisition" button, a modal appears to create a new requisition.  If you click "cancel", the entire grid refreshes!  We don't need to redraw the grid - we just need to get rid of the modal.  This is likely because of an instruction in the $state reference to refresh the $state on transition via the cancel button.  We don't need this and it should be removed on cancel.  On submit, it makes sense, but on cancel, it does not.

Specifically, you'll be looking for something like this:
```js
         $state.go('requisition.list', null, { reload : true }); 
```
and you'll need to remove the `reload` property.

jmcameron:
When in: Stock > Requisitions
a list of requisitions is shown.  When you click on the "New Requisition" or "Other Requistion" a new modal pops up for creating the requisition.  If you click on the [Cancel] button, this fix prevents the parent "Requisition Registry" from reloading.

Testing:
Before checking out this code and trying it, first try this with 'master'.  The reload is barely noticeable with Bhima_test.  You may want to try one of larger datasets to make this lack of a reload more obvious.

Then try it with this PR fix to verify there is no data reload.

NOTE: This is likely an issue for other modals to create a new item for a list/grid.

### DIFF
--- a/client/src/modules/stock/requisition/modals/action.modal.js
+++ b/client/src/modules/stock/requisition/modals/action.modal.js
@@ -56,8 +56,8 @@ function ActionRequisitionModalController(
   vm.addItem = addItem;
   vm.removeItem = removeItem;
   vm.configureItem = configureItem;
-  vm.cancel = Modal.close;
   vm.autoSuggestInventories = autoSuggestInventories;
+  vm.cancel = cancel;
 
   vm.onSelectDepot = depot => {
     vm.model.depot_uuid = depot.uuid;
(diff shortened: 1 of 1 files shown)


client/src/modules/stock/requisition/modals/action.modal.js
```diff
@@ -170,5 +170,9 @@ function ActionRequisitionModalController(
     item.quantity = 0;
   }
 
+  function cancel() {
+    Modal.dismiss('cancel');
```
mbayopanda, line 174: 👍

mbayopanda (approved):
LGTM

bors r+

PR #4983, opened 2020-10-05
chore: use bhFilterToggle on Entry from Transfer

jniles:
This commit makes the findTransfer modal use the `bhFilterToggle` component for filtering.  It also disables the submit button until data is loaded and removes an unnecessary binding.

### DIFF
--- a/client/src/modules/stock/entry/modals/findTransfer.modal.js
+++ b/client/src/modules/stock/entry/modals/findTransfer.modal.js
@@ -8,49 +8,39 @@ StockFindTransferModalController.$inject = [
 
 function StockFindTransferModalController(
   Instance, StockService, Notify,
-  uiGridConstants, Filtering, Receipts, data, bhConstants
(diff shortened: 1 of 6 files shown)


client/src/modules/stock/entry/modals/findTransfer.modal.js
```diff
@@ -75,13 +65,12 @@ function StockFindTransferModalController(
   }
 
   function rowSelectionCallback(row) {
-    vm.selectedRow = row.entity;
+    selectedRow = row.entity;
```
mbayopanda, line 68: 👍

mbayopanda (commented):
I have a non blocking comment, is it possible that these buttons have the same size ?

Capture d’écran 2020-10-07 à 10 41 28

jniles:
Great idea.  Fixed:

image

mbayopanda (approved):
LGTM 👍

PR #4985, opened 2020-10-05
Support gear icon for menu settings items

### LINKED ISSUE #4973: Make settings menu items support the gear icon
Once we move the Stock Settings under the Stock menu, it would be visually helpful to show its icon as "gear" indicating it has settings, not a regular page.  This may involve modifications to general menu handling to support the appropriate "fa" icon.  Then apply it to the new stock settings item.

jmcameron:
This PR adds a feature that supports using a gear in the main navigation menu for items that are identified as being a "settings" page.   This is accomplished by inserting '[SETTINGS]' (upper case) in the name of a particular menu item.   To see how this works, look at the new Stock settings page in the navigation menu.   Also notice the corresponding entry in the server/models/bhima.sql default menu definitions in the `units` table:

   (290, '[SETTINGS] Settings', 'TREE.STOCK_SETTINGS', 'Stock Settings', 160, '/stock/setting');

Notice that the name field  contains '[SETTINGS]'.   I noticed that some entries have '[Stock]' at the beginning of the name fields but we do not seem to be doing anything with it, except perhaps as a cue/reminder for the developers.   These insertions are not actually used in the construction of the menu entries (as far as is visible to the user).  The actual name of the menu item comes from translating 'TREE.STOCK_SETTINGS'.   So it seemed safe to insert a token in the name to identify it as a "Settings" item.


client/src/js/components/bhNavigation.js
```diff
@@ -147,6 +147,10 @@ function NavigationController($location, $rootScope, Tree, AppCache, Notify, $tr
     units.forEach(unit => {
       unitsIndex.id[unit.id] = unit;
       unitsIndex.path[unit.path] = unit;
+
+      // See if it is a settings page and needs a "gear" icon
+      unit.settings = unit.name.indexOf('[SETTINGS]') >= 0;
```
jniles: This is a boolean, right?  I propose that you name is `isSettings` or `isSettingsNode` or something like that.  Typically we name boolean variables with `is` or `has` or `should` or something like that.

client/src/modules/templates/navigation.tmpl.html
```diff
@@ -40,13 +40,16 @@
...
+          <!-- otherwise, it is just a plain link -->
           <a
             ng-click="$ctrl.navigate(child)"
             ng-if="$ctrl.isChildNode(child)"
             data-unit-key="{{ ::child.key }}"
             id="SidebarNavigation_Child_{{::child.key}}">
-            <span class="fa fa-file-o"></span>
+            <span class="fa" ng-class="{
+              'fa-file-o' : !child.settings,
+              'fa-cog' : child.settings
```
jniles: :+1:

jniles (approved):
Looks fine to me.  Nice!

client/src/js/components/bhNavigation.js
jmcameron: Updated

PR #4989, opened 2020-10-06
fix(cmm data) Make sure that the calculation is done for all movements

### LINKED ISSUE #4988: Error in the procedure for calculating the average monthly consumption
It has been noticed that there are errors in the calculation of the monthly average consumption, and this issue is to correct this problem.

jeremielodi:
closes #4988


lomamech (commented):
I have just started my investigation, using an inventory with which I entered stock from a purchase order, and I made three movements, the observation is that the CMM of the different algorithms is equal to Zero

image

image

@jeremielodi  Try to reevaluate with the solution you proposed the other time, or test with a simulation of stock entry and exit

jniles:
We should add some tests for this.  An integration test or a unit test would be ideal so we don't encounter this bug in the future.

lomamech:
> We should add some tests for this. An integration test or a unit test would be ideal so we don't encounter this bug in the future.
I will close, and the tests to ensure the different calculation will be dealt with with the issue
https://github.com/IMA-WorldHealth/bhima/issues/4976

PR #4991, opened 2020-10-06
Removed 'Add Tag' from modal

### LINKED ISSUE #4953: Remove "add tag" from the lots search modal
Our search modals so far haven't had functionality to add/modify data.  I don't think it makes much sense to add that now.  We should remove the "add tag" option from the search modal for lots in stock.  See below:

image

jmcameron:
Added support for controlling whether the "Add Tag" link is shown when using this bhTagSelect modal.   

To test this, 
  - go to Stock > Stock Lots
  - click the [Search] button
  - verify that the "Add Tag" is not shown after the "Tags" label:
image

  - Now try editing one of the lots.  You should see the "Add Tag" link.


client/src/js/components/bhTagSelect/bhTagSelect.html
```diff
@@ -4,14 +4,13 @@
     ng-class="{ 'has-error' : TagForm.$submitted && TagForm.tagUuids.$invalid }">
 
     <label class="control-label">
-      <span translate>{{ $ctrl.label }}</span>&nbsp;
-      <a style="cursor: pointer;" ng-click="$ctrl.createUpdateTagsModal()" translate>TAG.ADD</a>
```
jniles, line 8: This is used in the "edit lots" modal to quickly add lots on the fly.  Removing it from the component isn't really what we want since it removes it in each instance of the component.

Instead, let's add a binding to the `bhTagSelect.js` controller to add this `<a>` tag or hide it using an `ngIf` statement.  The default case will be to hide it.  It will be something like this:
```js
bindings : {
  // bindings...
  'enableAddTag' : '<?',
  // etc..
}
```

Then you can have a `ng-if="$ctrl.enableAddTag"` on is HTML tag and that is probably the sufficient.

jniles (changes requested):
@jmcameron this change looks like it will be applied application-wide, but we have a few use-cases that require it.  Could you make this configurable on the component-level?

jmcameron:
bors try

jmcameron:
bors try-

jmcameron:
Merging the latest changes in master made the tests pass.
@jniles This PR is ready for review and release.

jniles (commented):
@jmcameron this is almost ready to go!

We just need to update one thing - the lots edit modal.  Here is the place that needs to be touched:
https://github.com/IMA-WorldHealth/bhima/blob/cc77a85751fca71f18077a3e4a3177db4228fc02/client/src/modules/stock/lots/modals/edit.modal.html#L42-L45

Please add the `enableAddTag` to this component so that it can add tags at will.  After that, we are done!

PR #4992, opened 2020-10-06
Depot permission for distribution

mbayopanda:
This PR add feature of specifying distribution depots : 
Capture d’écran 2020-10-06 à 16 41 09

Enable or not this feature in the stock settings page
Capture d’écran 2020-10-06 à 16 44 09

If the feature is enable, the depot to select in the stock exit page for distribution are restricted to distribution depots defined for the current depot

### DIFF
--- a/client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
+++ b/client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
@@ -1,4 +1,3 @@
-
 angular.module('bhima.components')
   .component('bhDepotSearchSelect', {
     templateUrl : 'js/components/bhDepotSelectSearch/bhDepotSelectSearch.html',
@@ -79,7 +78,7 @@ function DepotSearchSelectController(Depots, uuidService) {
   };
 
   function loadSelected(depotsUuids) {
-    if (!(depotsUuids.length > 0)) return;
+    if (!depotsUuids || !(depotsUuids.length > 0)) return;
     Depots.read(null, { uuids : depotsUuids }).then(depots => {
       $ctrl.depotsSected = depots;
     });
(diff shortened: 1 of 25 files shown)


client/src/js/components/bhDepotSelectSearch/bhDepotSelect.js
```diff
@@ -79,7 +78,7 @@ function DepotSearchSelectController(Depots, uuidService) {
   };
 
   function loadSelected(depotsUuids) {
-    if (!(depotsUuids.length > 0)) return;
+    if (!depotsUuids || !(depotsUuids.length > 0)) return;
```
jniles, line 81: Personally, I find it slightly easier to read the `&&`  version of this.
```js
if (!(depotsUuids && depotsUuids.length > 0) {
  // Do something
}

```

client/src/modules/depots/depots.js
```diff
@@ -164,12 +163,12 @@ function DepotManagementController(
   }
 
   // switch to delete warning mode
-  function deleteDepot(depot) {
+  function deleteDepot(uuid) {
```
jniles, line 166: :+1:

client/src/modules/depots/modals/depot.modal.js
```diff
@@ -7,25 +7,37 @@ DepotModalController.$inject = [
...
-    delete vm.depot.parent_uuid;
-  }
+  vm.submit = submit;
 
   if (params.parentId) {
     vm.depot.parent_uuid = params.parentId;
   }
 
+  if (!vm.isCreateState) {
+    if (!vm.identifier) { return; }
```
jniles, line 26: Why would we be in an update state without a UUID?  Shouldn't we throw an error in this case?

jniles (approved):
This looks good to me.  I've taken a look at it and it seems to work fine.  Great work!

bors r+

PR #4993, opened 2020-10-06
Parameter setting cmm algo

### LINKED ISSUE #4984: Parameter setting of the CMM calculation algorithm to be used
There are several algorithms for the calculation of the average monthly consumption, and with the BHIMA system we had developed 4 algorithms, this issue proposes to be able to define in stock setting the algorithm to use

lomamech:
- Add property average consumption algorith is stock setting
- Update form for manage stock setting
- Add define algorith to use for average monthly consuption to Use

closes #4984


lomamech:
@jniles @mbayopanda  @jmcameron  All checks have passed

client/src/i18n/en/stock.json
```diff
@@ -101,6 +101,16 @@
     "MONTH"           : "Stock Months",
     "MONTH_OF_STOCK"  : "Stock Months - Months until Stock Out",
     "MONTHLY_CONSUM"  : "Average Monthly Consumption",
+    "MONTHLY_CONSUMPTION" : {
+      "ALGO_1"          : "Algorithm 1",
+      "ALGO_1_COMMENT"  : "The average monthly consumption is obtained by dividing the quantity consumed during the period defined by the number of days of stock for said period, and by multiplying the result obtained by 30.5",
```
mbayopanda, line 106: Since you mentioned `monthly consumption` and `30.5` which are related to month, is it possible to change `period defined` by just `month`?
Ex. The average monthly consumption is obtained by dividing the quantity consumed during the month by the number of days of stock of the month, and by multiplying the result obtained by 30.5

mbayopanda (approved):
LGTM 👍

jniles:
@lomamech looks like this will need to have some merge fixes before it can be merged.

client/src/modules/stock/settings/stock-settings.html
```diff
@@ -49,37 +49,62 @@
           <bh-yes-no-radios label="SETTINGS.ENABLE_DAILY_CONSUMPTION_LABEL"
             value="StockSettingsCtrl.settings.enable_daily_consumption"
             name="enable_daily_consumption"
+            id="enable_daily_consumption"
```
jmcameron, line 52: Great idea adding these tags so we can access them easily in the tests!

client/src/modules/stock/settings/stock-settings.js
```diff
@@ -43,6 +43,14 @@ function StockSettingsController(
           .catch(Notify.handleError);
       })
       .catch(Notify.handleError);
+
+    // load algorithmes for Average Consumption
+    StockSettings.getAverageConsuptionAlgo()
```
jmcameron: I find it hard to keep track of which algorithm is 1, 2, etc.   I suggest appending the algorithm number on these functions:    getAvertageConsomptionAlgo1()     (add the correct number!)
That way is is easier to track what is happening.

test/end-to-end/stock/stock.setting.js
```diff
@@ -0,0 +1,48 @@
...
+    await page.defineMonthAverageConsumption(5);
+    await page.defaultMinMonthsSecurityStock(4);
+
+    await page.setRadio('yes', 'enable_daily_consumption');
+    await page.setRadio('yes', 'enable_auto_purchase_order_confirmation');
+    await page.setRadio('yes', 'enable_strict_depot_permission');
+    await page.setRadio('yes', 'enable_auto_stock_accounting');
+    await page.setRadio('yes', 'enable_supplier_credit');
+    await element(by.id('algo_msh')).click();
+
```
jmcameron, line 26: Can we verify that these items have actually changed?  (also in the next 'it').

server/models/schema.sql
```diff
@@ -1849,6 +1849,17 @@ CREATE TABLE IF NOT EXISTS `voucher_item` (
   CONSTRAINT `voucher_item__voucher` FOREIGN KEY (`voucher_uuid`) REFERENCES `voucher` (`uuid`) ON DELETE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
+DROP TABLE IF EXISTS `average_consumption_algo`;
+
+CREATE TABLE `average_consumption_algo` (
+  `id` TINYINT(3) UNSIGNED NOT NULL AUTO_INCREMENT,
```
jmcameron: It is not clear to me why we need a table for the algorithms?   They seem to be something that is strictly in the Javascript code.

jmcameron (approved):
Great job!   

I ran this on system ('yarn dev') and it seems to work fine.

The new tests ran fine ('yarn test:ends').
Note that the unrelated distribution keys failed on my system (see Issue https://github.com/IMA-WorldHealth/bhima/issues/5006).  Since it passes on the CI testing, this is nothing for you to worry about.

I'm unclear: I see that we can select the CMM algorithm, etc, but are we actually using that algorithm in the CMM calculations (eg on the Items in Stock page)?

Also, when we start using the selected algorithm in the items in stock page, I suggest adding a tool tip on the "CMM" column header that shows which algorithm is being used for the calculation (when you hover over it).

client/src/i18n/en/stock.json
lomamech: I had agreed with @mbayopanda, for the calculation of the CMM we do not use only one me but rather the period defined in the stock Setting 
```
month_average_consumption 
````

**Comment by lomamech** [comment]
> Great job!
> 
> I ran this on system ('yarn dev') and it seems to work fine.
> 
> The new tests ran fine ('yarn test:ends').
> Note that the unrelated distribution keys failed on my system (see Issue #5006). Since it passes on the CI testing, this is nothing for you to worry about.
> 
> I'm unclear: I see that we can select the CMM algorithm, etc, but are we actually using that algorithm in the CMM calculations (eg on the Items in Stock page)?
> 
> Also, when we start using the selected algorithm in the items in stock page, I suggest adding a tool tip on the "CMM" column header that shows which algorithm is being used for the calculation (when you hover over it).

We are not using it yet, and it will be dealt with in the following issue, 

https://github.com/IMA-WorldHealth/bhima/issues/4976

File: client/src/modules/stock/settings/stock-settings.js (hunk 3 above)
Comment (lomamech) [reply client/src/modules/stock/settings/stock-settings.js]: This function just allows you to obtain the different existing algorithms...
````
// GET '/stock/setting/average_consuption'
// Get the different algorithms for calculating the average monthly consumption
exports.averageConsuptionAlgo = function averageConsuptionAlgo(req, res, next) {
  const sql = `SELECT id, name, label, comment FROM average_consumption_algo;`;

  db.exec(sql)
    .then((rows) => {
      res.status(200).json(rows);
    })
    .catch(next)
    .done();
};

````
image

File: server/models/schema.sql (hunk 5 above)
Comment (lomamech) [reply server/models/schema.sql]: The reason we used a table is just because we had thought to keep the identifier of the algorithm to be used among the stock settings.

**Comment by lomamech** [comment]
@jniles @mbayopanda @mbayopanda this PR need another review or Approbation

File: client/src/modules/stock/settings/stock-settings.js (hunk 3 above)
Comment (jmcameron) [reply client/src/modules/stock/settings/stock-settings.js]: Ok.  Thx.

File: server/models/schema.sql (hunk 5 above)
Comment (jmcameron) [reply server/models/schema.sql]: That sounds fine.  Can't we use an enum for the algorithm and add a new field in the stock_settings table?  I'm still not clear why we need a new table for the algorithm info.

Comment (lomamech) [reply server/models/schema.sql]: In this case we could use the component **bhConstants** and modify the property
````
 `average_consumption_algo_id` TINYINT(3) UNSIGNED NOT NULL DEFAULT 3,
````
In VARCHAR with the following values 'algo1', 'algo2', 'algo' and 'algo_msh'

Comment (lomamech) [reply server/models/schema.sql]: I will make this modification

**Comment by lomamech** [comment]
@jmcameron @jniles @mbayopanda  All Checks have passed

File: client/src/i18n/en/stock.json (hunk 1 above)
Comment (jniles) [reply client/src/i18n/en/stock.json:106]: We should probably come up with a better name than "algo 1", "algo 2", etc.  Also, I'm not sure we should support anything other than algo 1 and 4.

File: client/src/i18n/fr/stock.json [hunk 6]
```diff
@@ -182,7 +192,7 @@
     "ENABLE_AUTO_STOCK_ACCOUNTING_LABEL" : "Activer la comptabilisation de stock en temps réel",
     "ENABLE_AUTO_STOCK_ACCOUNTING_HELP_TEXT" : "L'activation de cette option va écrire automatiquement et en temps réel dans le journal toutes les transactions liées aux mouvements de stock. Les comptes des inventaires et groupes d'inventaire doivent être bien configurés au préalable",
     "ENABLE_DAILY_CONSUMPTION_LABEL" : "Activer le calcul du CMM sur base des jours de consommation",
-    "ENABLE_DAILY_CONSUMPTION_HELP_TEXT" : "La Cons mation Moyenne Mensuelle (CMM) est calculée en tenant compte des quantités consommées sur le nombre de jours où il y a eu des consommation",
+    "ENABLE_DAILY_CONSUMPTION_HELP_TEXT" : "La Consommation Moyenne Mensuelle (CMM) est calculée en tenant compte des quantités consommées sur le nombre de jours où il y a eu des consommation",
```
jniles, line 195: :+1:

server/models/migrations/next/migrate.sql
```diff
@@ -73,4 +73,31 @@ DROP TABLE IF EXISTS `depot_distribution_permission`;
...
+
+
+-- Average Consumption Algorithm
+INSERT INTO `average_consumption_algo` VALUES
+  (1, 'algo1', 'STOCK.MONTHLY_CONSUMPTION.ALGO_1', 'STOCK.MONTHLY_CONSUMPTION.ALGO_1_COMMENT'),
+  (2, 'algo2','STOCK.MONTHLY_CONSUMPTION.ALGO_2', 'STOCK.MONTHLY_CONSUMPTION.ALGO_2_COMMENT'),
+  (3, 'algo3', 'STOCK.MONTHLY_CONSUMPTION.ALGO_3', 'STOCK.MONTHLY_CONSUMPTION.ALGO_3_COMMENT'),
+  (4, 'algo_msh', 'STOCK.MONTHLY_CONSUMPTION.ALGO_4', 'STOCK.MONTHLY_CONSUMPTION.ALGO_4_COMMENT');
+
+ALTER TABLE `stock_setting` ADD COLUMN `average_consumption_algo_id` TINYINT(3) UNSIGNED NOT NULL DEFAULT 3;
```
jniles: Awesome 👍  Thanks for adding these to the migration script.

test/end-to-end/stock/stock.setting.js
jniles: Looks like this was addressed.

jniles (approved):
@lomamech great work!  This is a nice and concise PR.

I agree with @jmcameron that we don't really need to store these in the database.  However, I imagine on the long term, we may end up moving a lot of our constants into the database.  You've done a good implementation, if it becomes obvious that they should or should not be in the database, we'll take a further action.

bors r+

server/models/migrations/next/migrate.sql
lomamech: Sorry, I must update the migrate filte

PR #5004, opened 2020-10-07
fix(stock): use days not lifetime for expiration

jniles:
Lifetime is calculated in months, not days.  This made products that appear in less than a month break since lifetime is 0.  We've worked around it by changing to a calculation based on days.

Here is how to test this:
  1. Add a lot via integration that expires in a week.
  2. Try to find it in the drop down.

_It will not appear on master_ but will appear in this PR.


jmcameron (approved):
Looks good.  I tried it and it works as described.

PR #5009, opened 2020-10-08
fix(stock): show empty lots filter on stock lots

### LINKED ISSUE #4954: Remove "include empty lots" in default filters for stock movements registry
In the default filters for the stock movement registry, we have include a filter for including empty lots.  This doesn't make sense for the stock movements registry.  Let's remove it from the stock movements registry.

image

jniles:
This commit ensures that the empty lots filter is only shown on stock lots.

Closes #4954.

### DIFF
--- a/client/src/js/services/StockService.js
+++ b/client/src/js/services/StockService.js
@@ -79,6 +79,18 @@ function StockService(Api, StockFilterer) {
     requisition : StockRequisitionFilters,
   };
 
+  function assignNoEmptyLotsDefaultFilter(service) {
+    // add in the default key for the stock lots filter
+    const assignedKeys = Object.keys(service._filters.formatHTTP());
+    // assign default includeEmptyLot filter
+    if (assignedKeys.indexOf('includeEmptyLot') === -1) {
+      service._filters.assignFilter('includeEmptyLot', 0);
+    }
+  }
+
+  // assign non empty lots filter to the stockLots Filterer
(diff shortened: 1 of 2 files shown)


jmcameron (approved):
I verified that this works.  I see that the [Include empty lots: 0] filter token does not appear on the stock movements page but does appear on the Items in Stock and Stock Lots pages.

However, looking at the code I do not see how the fix does that.  An explanation would be helpful, but certainly not necessary to release this PR.

jniles:
@jmcameron you are right, the fix is opaque, because you need [addition context just a few lines above](https://github.com/IMA-WorldHealth/bhima/pull/5009/files#diff-bec8ceb2902c2b64f090b97c2deecce0R63).

Basically, in the previous version, this filter was being applied to every single filter service defined here.  They are:
```js
  const stockFilter = {
    lot : StockLotFilters,
    stockAssign : StockAssignFilters,
    movement : StockMovementFilters,
    inlineMovement : StockInlineMovementFilters,
    inventory : StockInventoryFilters,
    depot : StockDepotFilters,
    requisition : StockRequisitionFilters,
  };
```

So, the fix was to remove this loop and assign it to the stock.lots filter service specifically.

jmcameron:
So this looks like it is ready to go?

jniles:
It should be!  I suppose you did give me an approval, so I'll just merge it.

bors r+

jniles:
Bors r+

PR #5013, opened 2020-10-08
fix(distributionKey): ui-grid grouping expansion

### LINKED ISSUE #5006: Problem unfolding all rows in Grid
When you run the page:  Fee Center Management > Distribution Keys Managment,
it is supposed to start with all rows unfolded (un-collapsed).  However it does not always work -- although opening and closing the centers in the browser works.  This causes occasional problems for the end2end tests.

The problem seems to be in the  GridGrouping.js file:

  function unfoldAllGroups(api = this.gridApi) {
    $timeout(api.treeBase.expandAllRows, 0, false);
  }

https://github.com/IMA-WorldHealth/bhima/blob/9e85213a55075b2e64e4086b4f735303d26ed8a8/client/src/js/services/grid/GridGrouping.js#L115

It appears that this treeBase.expandAllRows function is not working correctly.

UPDATE 2020-11-11: This issue was resolved for the Distribution Keys issue in PR https://github.com/IMA-WorldHealth/bhima/pull/5093.  The problem was that the grid was not being started in a consistent state (in terms of folded/unfolded children).  After some debugging, we added a call 

    vm.grouping.unfoldAllGroups();

in the main loading function for the page.   This resolved the problem for Distribution Keys and is how some of the other pages with grids operate.   There may be a bug in the Grid 'expand' functionality that should be addressed eventually.

jniles:
Fixes the ui-grid expansion on entering the distribution keys registry.

Closes #5006.

### DIFF
--- a/client/src/js/services/grid/GridGrouping.js
+++ b/client/src/js/services/grid/GridGrouping.js
@@ -112,7 +112,7 @@ function GridGroupingService(GridAggregators, uiGridGroupingConstants, Session,
   }
 
   function unfoldAllGroups(api = this.gridApi) {
-    $timeout(api.treeBase.expandAllRows, 0, false);
+    $timeout(() => api.treeBase.expandAllRows(), 0);
   }
 
   function unfoldGroup(row) {
--- a/client/src/modules/distribution_center/distribution_key/distribution_key.js
+++ b/client/src/modules/distribution_center/distribution_key/distribution_key.js
@@ -12,8 +12,7 @@ DistributionKeyController.$inject = [
  * This controller is about the updating Distribution Center module in the Finance zone
  * It's responsible for editing and updating a Distribution Center
  */
-function DistributionKeyController(DistributionCenters, Notify, uiGridConstants,
-  Grouping, uiGridGroupingConstants) {
+function DistributionKeyController(DistributionCenters, Notify, uiGridConstants, Grouping, uiGridGroupingConstants) {
   const vm = this;
 
   // bind methods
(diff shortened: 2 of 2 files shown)


jniles:
@jmcameron I can't test this locally.  I'm not sure this will fix it - if not we need to remove that "false" from the `$timeout()` call and see if that does it.

Can you pull this and run the tests and report back if they work?

jmcameron (approved):
It seems to be fixed.  When I go to the Distribution Keys Management page, it is indeed all open.  So that is good.  

However, if I refresh that page, they are collapsed.  Since that is not what the tests will do, that may not be an issue.  But I found it odd that going to the page (via the menu) would produce different results than refreshing it after you're on the page.

jniles:

I'm not sure why that happens either, honestly.  I could be a ui-grid issue.  I'll try and reproduce a couple scenarios locally and see if I can get the same behavior.

Do _all_ tests pass on your machine now?

jmcameron:
> Do _all_ tests pass on your machine now?

NO!  Sorry, forgot to run it before.  Same error:

  1) Distribution keys Management
       displays all distributions keys loaded from the database:

      AssertionError: expected 3 to equal 10
      + expected - actual

      -3
      +10

jniles:
@jmcameron can you give this one more shot?  I don't think I fixed it, but it would be worth knowing.

jmcameron:
I tested this again:  Same results:
  - 'yarn test:ends' fails with the 3/10 distribution keys test
  - 'yarn dev' and going to Distribution keys Mgt comes up collapse.  If I go to another menu item and come back, it is uncollapsed. If I do a refresh, it is collapsed.

jniles:
@jmcameron so... try as I might, I can't get this to fail or reproduce this error.  I've tried windows, Linux, hard refresh, normal refresh, and clearing cache, collapsing the grid before all the above, and still can't get this to fail in the way you are seeing.  If you want to dig into it, feel free, but I'm going to close this since I can't reproduce the issue.

Issue #5011, opened 2020-10-08
Install feature is broken


mbayopanda:
This issue is about we cannot go to the state `install` when we want to install a new instance of bhima; we are locked in the `login` state

<img width="1680" alt="Capture d’écran 2020-10-08 à 14 09 42" src="https://user-images.githubusercontent.com/5445251/95462888-02a09e80-0970-11eb-8cd1-1b317fcbcda0.png">

mbayopanda:
Closed by #5014

PR #5014, opened 2020-10-08
fix install routing

mbayopanda:
This PR is about the issue #5011


jniles (approved):
This is right.

bors r+

PR #5016, opened 2020-10-09
Added 'chromeOptions' from local environment variable

jmcameron:
When we run 'yarn test:ends' on our local computer, we each may need different options in the 'chromeOptions' part of the file.  You may need to run 'headless'; but some may not want to do that.   This mod allows you to specify an environment variable:

   CHROME_OPTIONS

which will be added to the protractor 'chromeOptions'.   For example if you want to run 'ends' headless, define:

   export CHROME_OPTIONS="--headless"

Depending on your local environment, you may need other options depending on whether you are running Chrome or Chromium.  Other options that can be used for Chromium are:
	'--no-sandbox',
	'--disable-dev-shm-usage',
	'--remote-debugging-port=9222',
	'--disable-extensions'

To find out more about these options, try searching for them on Google.

NOTE: You should be able to specify multiple options here (space separated), but I have not actually tested that.

Please make sure this works (without defining CHROME_OPTIONS) on your local computers.


protractor.conf.js
```diff
@@ -29,7 +29,7 @@ const config = {
   capabilities : {
     browserName : 'chrome',
     chromeOptions : {
-      args : ['--disable-gpu', '--window-size=1280,1024'],
+      args : ['--disable-gpu', '--window-size=1280,1024', process.env.CHROME_OPTIONS],
```
jniles: This is a good idea, however, it doesn't work on my machine.  I tweaked it a bit and this is what works:
```js
const config = { 
  // all the params
};

if (process.env.CHROME_OPTIONS) {
  config.capabilities.chromeOptions.args.push(process.env.CHROME_OPTIONS);
}
```

jniles (changes requested):
@jmcameron this doesn't work out of the box without `CHROME_OPTIONS` defined. I've proposed the fix.

protractor.conf.js
jmcameron: Good catch.  Should have tried it without an environment variable myself!

jmcameron:
bors try

lomamech (commented):
When I did the E2E test with Protractor I got the following error message
image

protractor.conf.js
jmcameron: Done

jniles:
bors try

jniles:
@lomamech can you re-test this?

jniles:
Or @mbayopanda .  I just would like to make sure this is not windows specific.

protractor.conf.js
```diff
@@ -78,5 +78,10 @@ if (process.env.CI) {
   delete config.plugins;
 }
 
+// Add custom Chrome options for local environments
+if (!process.env.CI && process.env.CHROME_OPTIONS) {
```
mbayopanda, line 82: The test e2e works fine, I would like just to know what kind of values `CHROME_OPTIONS` can take, also if it is possible, add them in the comment as an example

mbayopanda (approved):
LGTM 👍 

I just left a non-blocking comment about the type of use of `CHROME_OPTIONS`

protractor.conf.js
jmcameron: @mbayopanda I added a little more info on the PR description at the top. 
Thanks for testing it!

jniles (approved):

jniles:
This is excellent, thank you.

bors r+

jniles:
This works for me and @mbayopanda .

bors r+

PR #5022, opened 2020-10-13
Reload after changing the date

### LINKED ISSUE #4981: Changing the date on the stock ajustments page does not refresh the quantity in stock
I got a bug report from one of our production sites - the `/stock/inventory-adjustment` route does not change the old quantity in stock when the user changes the date.  This makes it hard to back-date an inventory.

**Current Behavior**: When the date changes, nothing happens.

**Expected Behavior**: When the date changes:
  1. We should disable the submit button (to prevent someone messing with it).
  2. We should transition the grid into "loading" mode.
  2. We should send a request back to the server to get the amount of value in stock on that particular date.
  4. When we get the answer, take the grid out of loading mode.
  3. We should re-enable the submit button.

With these steps, we would allow users to back-date these transactions to the correct date and reload the values for that date.

**Steps to tackle this**

Add to the `onDateChange()` function below the `loadInventories(vm.depot)` function.

https://github.com/IMA-WorldHealth/bhima/blob/fd4e7cb36ca06c03d4f793df8a02537183121531/client/src/modules/stock/inventory-adjustment/inventory-adjustment.js#L29-L31

And I think that should be it.  Please check that:
  1. The submit button is disabled.
  2. The grid transitions into the loading state.
  3. The data is refreshed based on the date you chose.

jmcameron:
In Issue https://github.com/IMA-WorldHealth/bhima/issues/4981, @jniles indicated the steps to be taken in the callback. However, it seems to me that all you need to do is change the date and reload the inventory items (I think the loadInventories() function does all that is needed).  I believe the grid will not get control back until the reload is complete.

I was not sure how to test this, since I'm a bit confused about what adjusting the inventory at some past date really means and how to test it.

-Jonathan

### DIFF
--- a/client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
+++ b/client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
@@ -28,6 +28,7 @@ function StockInventoryAdjustmentController(
 
   vm.onDateChange = date => {
     vm.movement.date = date;
+    loadInventories(vm.depot);
   };
 
   vm.onChangeDepot = depot => {


jniles (approved):
Excellent, this works:

8I3oFkWd3r


bors r+

PR #5023, opened 2020-10-14
Fix migration problem with production databases.

### LINKED ISSUE #5021: Migration for non-test databases is broken
We seem to have a problem with the typical migration procedure for the non-test databases.  I've confirmed this for both IMCK and Vanga. 

The following procedure fails in the last step.  These use the "fetch" and "build" scripts provided to me some time ago.  I'll attach the (slightly updated) versions I use here.  I modified the fetch script to put the files in specific location on my system.

Here is the procedure I used:
  - **DATABASE=imck bhima-fetch-latest-backup.sh**
     - runs fine with no prompts
  - **DATABASE=imck bhima-fetch-latest-backup.sh**
     - enter mysql root password twice, runs slow but without problems
  - Edit .env to change:  DB_NAME='imck'
  - **yarn migrate**
      - Runs fine, produces migration-imck.sql
  - **mysql -u bhima -p --database=imck < migration-imck.sql**
     - type in 'bhima' password
     - Produces this error:
          _ERROR 1054 (42S22) at line 222: Unknown column 'project_id' in 'NEW'_
     - Offending SQL:
          _CREATE TRIGGER stock_requisition_reference BEFORE INSERT ON stock_requisition_
          _FOR EACH ROW_
          _SET NEW.reference = (SELECT IF(NEW.reference, NEW.reference, IFNULL(MAX(stock_requisition.reference) + 1, 1)) FROM stock_requisition  WHERE stock_requisition.project_id = new.project_id);$$_

Note that I do not see this SQL in server/migrate/next/migrate.sql.  This seems to be something produced by the 'yarn migrate' script.

I see the same error when I follow this procedure for the Vanga database.

[bhima-build-backup.sh.txt](https://github.com/IMA-WorldHealth/bhima/files/5374184/bhima-build-backup.sh.txt)
[bhima-fetch-latest-backup.sh.txt](https://github.com/IMA-WorldHealth/bhima/files/5374185/bhima-fetch-latest-backup.sh.txt)

jmcameron:
Updated migration file to fix some issues in the migration of production databases (for testing).

- Fixed oversight in stock_settings PR that did not create the stock_setting table (if it does not exist). 
  Other minor updates in the same file  combining some of migration parts related to stock_setting.
- Created new SQL function: add_column_if_missing() in server/models/procedures/migration-process.sql
- Changed the order of creating migration files to define SQL functions and procedures before triggers.sql since they 
  may be needed there.   Similar fix in script for Bhima_test.
- Tested against the IMCK and Vanga databases
- Tested with Bhima_test database


mbayopanda (commented):
The idea of using a procedure to add a new column is good

I had just an error related to `collations` : 
image

jniles:
It also looks like this is breaking the install builds.

server/models/procedures/migration-process.sql
```diff
@@ -35,17 +35,36 @@ BEGIN
...
+DELIMITER $$
+CREATE PROCEDURE add_column_if_missing(
+  tname VARCHAR(64),
+  cname VARCHAR(64),
+  typeinfo VARCHAR(128)
+)
+BEGIN
+    IF NOT column_exists(tname, cname)
+    THEN
+      SET @add_column_if_missing = CONCAT('ALTER TABLE `', tname, '` ADD COLUMN `', cname, '` ', typeinfo);
```
jniles: For some reason, this is resulting in the following error on mysql 5.7.  I believe the default collations changed on MySQL 8 which is why this isn't failing.
```sh
ERROR 1267 (HY000) at line 118: Illegal mix of collations (utf8_general_ci,IMPLICIT) and (utf8_unicode_ci,IMPLICIT) for operation '='
```

jmcameron:
bors try

jniles:
Okay, so I think I got this working.  Here are the highlights:

  1. Our build-init-database.sh script [was not using the utf8mb4 character set and collation](https://github.com/IMA-WorldHealth/bhima/pull/5023/files#diff-9332ba4b6f520e1bd9a46224c87395eff9cd33227ec0b9e7be4a059df30f4669L20).
  2. I've cleaned up our ICD10 sql that dropped a table needless and set the collation automagically to utf8 using MySQL dump comments.  This wasn't the problem though.
  3. I added `yarn test:install` to run the installation tests.

My approach to debugging was as follows:
  1. I ran the `./sh/install-tests.sh` to check if that worked.  It didn't.
  2. I logged in to the MySQL command line and ran the CALL add_column_if_missing()` procedure.  That worked.
  3. I added COLLATIONS to all variables in the stored procedure.  No change.
  4. I googled the error and found out that stored procedures use the database collation.
  5. I checked the variables on the command line using `SHOW PROCEDURE STATUS LIKE "add%"`.  [docs](https://dev.mysql.com/doc/refman/8.0/en/show-procedure-status.html).  This revealed that the production database's `add%` procedures were in `utf8mb4_unicode` but the installed database uses `utf8`.  Interesting.
  5.  I checked the build-init-database.sh script's DROP SCHEMA and CREATE SCHEMA.
  6.  There was the issue - the init script wasn't using the proper collation.

jmcameron:
bors try

jniles:
Looks like we still have issues with MariaDB.  Gonna try and figure this out.

jniles:
So, the issue was that `COLUMN_EXISTS()` is a mariadb reserved function.  Apparently, it is defined in their dynamic columns feature. Renaming it solved the issue.

jmcameron:
bors try

jmcameron:
bors try

lomamech:
> bors r+

I also just noticed that there is missing the script to add the stock setting module in unit 
```    (290, '[SETTINGS] Settings', 'TREE.STOCK_SETTINGS', 'Stock Settings', 160, '/stock/setting');  ````

jniles:
@lomamech great observation.  Can you add it in a separate pull request?

lomamech:
> @lomamech great observation. Can you add it in a separate pull request?

I'll do

jmcameron:
> 
> 
> > bors r+
> 
> I also just noticed that there is missing the script to add the stock setting module in unit
> ``` (290, '[SETTINGS] Settings', 'TREE.STOCK_SETTINGS', 'Stock Settings', 160, '/stock/setting'); ````

That was in the original stock settings PR https://github.com/IMA-WorldHealth/bhima/pull/4946.  Perhaps that was lost when the tests were refactored?

Here is where it was in the PR: https://github.com/jmcameron/bhima/blob/ed6f37b0b2528442fe5fa05129b1516250c42e65/server/models/bhima.sql#L166

jniles:
@jmcameron, it still is in the schema file.  When we update the `unit` table, we generally add a row in the `migrate` file for this as well, since it is additional data that we'll need to add to our production databases.  @lomamech noticed it and correctly fixed it.  I didn't notice it in my review which is my fault 😞

jmcameron:
> @jmcameron, it still is in the schema file. When we update the `unit` table, we generally add a row in the `migrate` file for this 

No problem.  Glad it was caught.  It took me a bit of digging before realizing that I missed it in the original PR.

PR #5024, opened 2020-10-15
fix(fee centers): do not show uuids

### LINKED ISSUE #4938: Hide UUIDs from the user in Distribution Center registry
Here is what the distribution center update page looks like:

image

We should update this page to replace the UUID with the transaction ID and hide the uuid from the client.  We don't show UUIDs anywhere else in the application and it doesn't make sense to show them here.

jniles:
Fixes the client to group by transaction ID, not UUID.  Also removes the selection from the grid.  Finally, it also adds a "view transactions" feature to allow users to quickly browse the linked transactions in the posting journal.

Closes #4938.

### DIFF
--- a/client/src/modules/distribution_center/update/update_center.js
+++ b/client/src/modules/distribution_center/update/update_center.js
@@ -12,8 +12,10 @@ UpdateCenterController.$inject = [
  * This controller is about the updating Distribution Center module in the Finance zone
  * It's responsible for editing and updating a Distribution Center
  */
-function UpdateCenterController(DistributionUpdateCenters, DistributionCenters, ModalService, Notify, uiGridConstants,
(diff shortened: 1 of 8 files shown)


client/src/modules/distribution_center/update/update_center.js
```diff
@@ -116,15 +113,14 @@ function UpdateCenterController(DistributionUpdateCenters, DistributionCenters,
     }],
   };
 
-  vm.grouping = new Grouping(vm.gridOptions, true, 'row_uuid', vm.grouped, true);
+  vm.grouping = new Grouping(vm.gridOptions, true, 'trans_id');
```
mbayopanda, line 116: 👍

jniles:

mbayopanda (approved):
LGTM 👍 

bors r+

PR #5037, opened 2020-10-21
feat(sh): add script to download latest release

jniles:
Adds a bash script to download the latest BHIMA release. Useful for upgrading applications that do not use git for version control.

### DIFF
--- a/sh/download-latest-release.sh
+++ b/sh/download-latest-release.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# bash script mode
+set -eo pipefail
+
+OWNER="ima-worldhealth"
+REPO="bhima"
+
+echo "looking up latest version of $REPO"
+VERSION=$(curl -s "https://github.com/$OWNER/$REPO/releases/latest/download" 2>&1 | sed "s/^.*download\/\([^\"]*\).*/\1/")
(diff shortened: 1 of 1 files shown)


sh/download-latest-release.sh
```diff
@@ -0,0 +1,15 @@
...
+
+OWNER="ima-worldhealth"
+REPO="bhima"
+
+echo "looking up latest version of $REPO"
+VERSION=$(curl -s "https://github.com/$OWNER/$REPO/releases/latest/download" 2>&1 | sed "s/^.*download\/\([^\"]*\).*/\1/")
+echo "found $VERSION"
+
+echo "downloading $VERSION from codeload.github.com to /tmp/bhima.tar.gz"
+wget -O /tmp/bhima.tar.gz -q https://codeload.github.com/$OWNER/$REPO/legacy.tar.gz/$VERSION
```
mbayopanda, line 14: The script is only for Linux, we can have wget for Windows but there is not the /tmp folder

mbayopanda (commented):
The script seems to work but on Windows we cannot download the file in `/tmp`. If we change the foler `/tmp` to an existing folder on Windows the script work but the file has 0 bytes

image

jmcameron:
bors try

PR #5048, opened 2020-10-23
Tests: Stock CMM Tests

jniles:
Work done by @mbayopanda.  I'm just pulling this to leave comments and make my review.


server/controllers/stock/core.js
```diff
@@ -773,8 +779,13 @@ async function getInventoryQuantityAndConsumption(params, monthAverageConsumptio
     const setting = await db.one(settingsql, filteredRows[0].enterprise_id);
     const nbrMonth = setting.month_average_consumption;
 
-    const startDate = new Date();
-    startDate.setDate(startDate.getDate() - 30.5 * nbrMonth);
+    let startDate = new Date();
+    let clone = moment(startDate);
+    const months = (nbrMonth - 1 > -1 && nbrMonth) ? nbrMonth - 1 : 0;
+    clone = clone.subtract(months, 'month').toDate();
```
jniles, line 785: I'm curious why you didn't use [`moment.startOf()`](https://momentjs.com/docs/#/manipulating/start-of/) to set the month?  Was there a reason?  

This could be written: 

```js
const startDate = moment().subtract(months, 'month').startOf('month').toDate();
```

server/models/procedures/stock.sql
```diff
@@ -625,7 +625,7 @@ BEGIN
 
 
   SET _sum_days = DATEDIFF(_end_date,  _start_date) + 1;
-  SET _number_of_month = (DATEDIFF(_end_date,  _start_date) + 1)/30.5;
+  SET _number_of_month = ROUND(DATEDIFF(_end_date,  _start_date)/30.5);
```
jniles, line 628: We need to figure out a better way to do this.  If we have 0.45 months, this will say 0 months.  Does that make sense?  I would think it should be at least 1.  Maybe we can use `CIELING()`?  Or, set _number_of_month to be `1` if it is `0`?

server/models/procedures/stock.sql
```diff
@@ -650,9 +650,9 @@ BEGIN
 
           WHERE i.uuid = _inventory_uuid AND m.depot_uuid = _depot_uuid
              AND  (
-              ((m.start_date >= _start_date) AND  (m.end_date <= _end_date)) OR
-              ((m.start_date >= _start_date) AND  (m.end_date >= _end_date)) OR
-              ((m.start_date <= _start_date) AND  (m.end_date <= _end_date))
+              DATE(m.start_date) >= DATE(_start_date) AND DATE(m.end_date) <= DATE(_end_date)
```
jniles, line 653: @lomamech has more experience here.  I know there were some bugs fixed in the logic here and I'm curious if this undoes them.

server/models/procedures/stock.sql
```diff
@@ -665,12 +665,21 @@ BEGIN
...
-  SET @algo_msh = (_sum_consumed_quantity/(_number_of_month - ((_sum_stock_out_days + _days_before_consumption)/30.5) ));
+  SET @algo_msh = (_sum_consumed_quantity/(_number_of_month - (_sum_stock_out_days/30.5) ));
 
-  SELECT IFNULL(@algo1, 0) as algo1,
+  SELECT ROUND(IFNULL(@algo1, 0), 2) as algo1,
     ROUND(IFNULL(@algo2, 0), 2) as algo2,
     ROUND(IFNULL(@algo3, 0),2) as algo3,
-    ROUND(IFNULL(@algo_msh, 0), 2) as  algo_msh;
+    ROUND(IFNULL(@algo_msh, 0), 2) as  algo_msh,
+    _start_date as start_date,
```
jniles, line 674: This is excellent :+1:.  We should make this available on the client too.

test/data/inventories.sql
```diff
@@ -0,0 +1,2862 @@
...
+;
+
+-- 
+-- inventory
+-- 
+
+INSERT INTO inventory (enterprise_id,uuid,code,`text`,price,default_quantity,group_uuid,unit_id,unit_weight,unit_volume,stock,stock_max,stock_min,type_id,consumable,sellable,note,locked,delay,avg_consumption,purchase_interval,last_purchase,num_purchase,num_delivery,importance,created_at,updated_at) VALUES 
+(1,0x3CF8D982EEF911EA880100155D871701,'DORA_ERYT5T-_0','Erythromycine, 500mg, Tab, 500, Vrac',3.1800,1,0xD81D17A7727C11EA8241000C2997DDC0,18,0,0,0,0,0,1,1,1,NULL,0,1.0000,0.0000,0.0000,NULL,0,0,NULL,'2020-03-30 12:49:56',NULL)
+,(1,0x43F3DECBFCE9426E940ABC2150E62186,'DORA_QUIN1S-_0','Quinine Bichlorhydrate, sirop, 100mg base/5ml, 100ml, flacon, Unité',6.6300,1,0xD81D17A7727C11EA8241000C2997DDC0,18,0,0,0,0,0,1,1,1,NULL,0,1.0000,0.0000,0.0000,NULL,0,0,NULL,'2020-03-30 12:49:57',NULL)
+,(1,0x6B4825F14E6E47998A81860531281437,'DORA_PARA5TB_0','Paracetamol, 500mg, Tab, 1000, Blister',1.7400,1,0xD81D17A7727C11EA8241000C2997DDC0,18,0,0,0,0,0,1,1,1,NULL,0,1.0000,0.0000,0.0000,NULL,0,0,NULL,'2020-03-30 12:49:57',NULL)
```
jniles, line 67: We should use relative dates in our tests to make sure we have accurate tests going forward that do not break with time 😄

jniles (commented):
@mbayopanda, I've made a few comments.  This is really good work.  I like the separation of test data into separate files.  That is a great idea and makes it really easy to analyze.

PR #5053, opened 2020-10-27
Allow all depots to add sub-depots from action menu

### LINKED ISSUE #5005: Allow any depot to have children, not just warehouses
The depot tree described in https://github.com/IMA-WorldHealth/bhima/issues/4922 and implemented in https://github.com/IMA-WorldHealth/bhima/pull/4944 is about creating a permissions hierarchy.  However, the implementation requires depots to be warehouses to have children.

I propose we remove this logic.  It is complicated and isn't related to permissions, which is why we originally wrote this code.  Furthermore, you can make a child, then edit the parent to not longer be a warehouse, so there are logic bugs.

Let's just remove this restriction.

jmcameron:
To test, go to the Depot Management page and verify that the action menu for all depots have the "Add Sub Depot" item whether the depot is a warehouse or not.


jmcameron:
bors try

jmcameron:
@lomamech I think this is ready to review.  I fixed the CI e2e tests. I think the integration test failure is the same as for many of the other PRs (I'm doing bors try).  Thanks!

lomamech:
It work fine,

PR #5063, opened 2020-10-29
Fix inventory selection bug in stock exits

### LINKED ISSUE #5040: Do not present inventory for selection in stock exit if no viable lots available
If there are no viable lots for the inventory, we should not show it as an option to distribute stock to.  This can happen for a couple reasons:
  1. There were lots, but they are being used in a other lines above.
  2. There are lots, but they are expired.
  3. ?

By hiding these lots, we have a weird user interface.  Here I am reproducing it:

KcyRk3w4HS

Instead, we should remove the inventory from the pool of options, rather than letting the user select it and then finding nothing.

jmcameron:
**Testing:**
With Bhima_test, try doing Stock > Exit, by Services, and then try adding "Vitamin.." twice (VITAMIN-A, and then VITAMIN-B).  Now if you try to add a third item, "Vitamin..." is no longer available (since we have already used all available lots already in the form).

NOTE That this assumes that if you cannot add a specific lot more than once in a Stock exit form.

### DIFF
--- a/client/src/modules/stock/exit/exit.js
+++ b/client/src/modules/stock/exit/exit.js
@@ -28,6 +28,7 @@ function StockExitController(
   vm.movement = {};
   vm.gridApi = {};
   vm.selectedLots = [];
+  vm.selectableInventories = [];
   vm.reset = reset;
   vm.ROW_ERROR_FLAG = bhConstants.grid.ROW_ERROR_FLAG;
 
@@ -263,7 +264,6 @@ function StockExitController(
 
   function loadInventories(depot, dateTo = new Date()) {
     setupStock();
-
     Stock.inventories.read(null, { depot_uuid : depot.uuid, dateTo })
       .then(inventories => {
         vm.loading = false;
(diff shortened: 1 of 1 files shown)


client/src/modules/stock/exit/exit.js
```diff
@@ -282,14 +281,21 @@ function StockExitController(
...
   // update the list of selected lots
-  function refreshSelectedLotsList() {
+  function refreshSelectedLotsList(row) {
     vm.selectedLots = vm.stockForm.store.data
       .filter(item => item.lot && item.lot.uuid)
       .map(item => item.lot.uuid);
+
+    // If we have picked the last lot for this inventory item,
+    // we need to remove it from the selectableInventories list
+    if (row.lots.length === 1 && row.lots[0].inventory_uuid === row.lot.inventory_uuid) {
```
mbayopanda, line 295: 👍 Good catch

mbayopanda (approved):
LGTM 👍 

bors r+

PR #5074, opened 2020-11-02
Fix stock inventory test

mbayopanda:
This PR fixes stock inventory registry tests on master by adjusting values of test


jniles (approved):
Good enough!  Just make it work :+1:

LGTM.

bors r+

PR #5075, opened 2020-11-02
Add Helpdesk contact info banner

### LINKED ISSUE #5031: Add a banner to the BHIMA login page
As we will be deploying a BHIMA server nationwide, we will want to redirect users to helpdesks if they forget their password.  Here is how DHIS2 solves this problem:

image

This is a custom "banner" that you can add to the home page.  I propose we do the same thing in BHIMA.  This would be a `textarea`, part of the enterprise information.  It will be displayed on every login page.

If the `banner` section is empty in the enterprise information, don't show anything.  The behavior will be identical to the current behavior.

jmcameron:
Added a new Enterprise information item:  Helpdesk Contact Info  
If this parameter is set, the contents of that field will be inserted at the bottom of the login form.

**To Test:**
- Use bhima_test
- yarn build:db
- yarn dev
- Observe at the bottom of the login  form there is nothing below the [Log in] button
- Log in and go the Admin/Enterprise page
- Set the value of the new "Helpdesk Contact Info" field (at the bottom of the left column of Enterprise info)
- Log out
- Observe the new string now shows at the bottom of the login form

NOTE:  Please check the French translation!


jmcameron:
bors try

server/models/schema.sql
```diff
@@ -558,6 +558,7 @@ CREATE TABLE `enterprise` (
   `logo`            VARCHAR(100) DEFAULT NULL,
   `currency_id`     TINYINT(3) UNSIGNED NOT NULL,
   `po_box`          VARCHAR(30) DEFAULT NULL,
+  `helpdesk`        VARCHAR(512) DEFAULT NULL,
```
jniles: Let's make this TEXT.  From experience, I know we'll regret it one day if we don't.

server/controllers/admin/helpdesk.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * @function HelpdeskInfo
+ * @description return the helpdesk info string
+ * Should update to support a passed-in enterprise id
+ */
+
+function helpdeskInfo(req, res, next) {
+
+  const sql = `select helpdesk
```
jniles, line 21: Perfect :+1:

server/controllers/admin/helpdesk.js
```diff
@@ -0,0 +1,32 @@
...
+    FROM enterprise
+    LIMIT 1;
+  `;
+
+  db.exec(sql, [])
```
jniles: If you use `db.one()` this will spit out a JSON object, avoiding the need to do `db.exec(query, params).then(rows => rows[0])`.  `db.one()` is an alias for this operation, returning only the first row of the resulting query.

It also throws an error if any more or less than one row is returned.

client/src/modules/login/login.js
```diff
@@ -62,6 +63,15 @@ function LoginController(AppCache, Session, Languages, Projects, Notify, Install
     vm.languages = languages;
   }
 
+  Helpdesk.read()
+    .then(setHelpdeskInfo);
+
+  function setHelpdeskInfo(info) {
```
jniles, line 69: Nice 👍

client/src/modules/login/login.html
```diff
@@ -132,6 +132,9 @@ <h1 translate class="text-center">INSTALL.TITLE</h1>
             </form>
           </div>
         </div>
+        <div class="helpdesk-info" ng-if="LoginCtrl.helpdesk_info">
```
jniles: Let's put this in a `class="alert alert-info"` box to draw attention to it.  Doing so, makes it look like this:

image

jniles (commented):
Looks really good!  I've made a few comments on the code.


This is what it currently looks like, for those interested:
image

client/src/modules/login/login.html
jmcameron: I've done this.  This is okay for now.  However, when we actually start putting information here, the blue of the 'alert' area may not fit in with the visualization that we want.  We will probably need to develop custom CSS rules for this area--which we can do later.

server/models/schema.sql
jmcameron: Done

server/controllers/admin/helpdesk.js
jmcameron: Done

lomamech (commented):
I noticed that in your note that you had several ways to do this, there are still the tests which do not pass yet, perhaps it will be necessary to update with the branch master

PR #5080, opened 2020-11-05
fix(reports): NULL recovery capacity where no data

jniles:
Adds "NULL" for the recovery capacity where there is no data to draw attention to it and ensure it is not taken into account in the
calculation.

**Before**
ISSUE - Do not show 0% recovery if no data -BHIMA

**After**
ISSUE - FIXED - Do not show 0% recovery if no data -BHIMA


lomamech (commented):
I will continue the analysis with a database with a lot of data, but I thought we had to use translation to display NULL in French or in English.

image

jniles:
@lomamech can you help me with a suggestion for the french word to use?

lomamech:
> @lomamech can you help me with a suggestion for the french word to use?

Null (en) and Nulle(fr) just because value is a feminine word

jniles:
@lomamech, thank you 😄

lomamech:
> @lomamech, thank you 😄

image
I have a concern with this report on collection capacity.
I carried out a series of invoices, payments just to analyze the data of the reports and I have the impression that I did not understand the value of registrations which is equal to 5 while I made more than 5 invoices

jniles:
@lomamech registrations are the number of patients, not the number of invoices.  The "average cost" is the average cost per patient, not per invoice.

We should probably include that text on the report 👍 .  That's a good suggestion.

lomamech:
> @lomamech registrations are the number of patients, not the number of invoices. The "average cost" is the average cost per patient, not per invoice.
> 
> We should probably include that text on the report 👍 . That's a good suggestion.

This is a good proposition, the labels should really speak for ease of analysis of the report.

jniles:
@lomamech  here are the changes:

image

jniles:
@lomamech changes landed!

lomamech:
> @lomamech changes landed!

I will investigate to better understand the report before closing it

PR #5084, opened 2020-11-05
Allow default access to Stock Setting menu item for admin, superuser

jmcameron:
Updates the migration file so that the new Stock Settings menu item is available by default for Admin and superuser without having to edit their permissions.

Tested on IMCK production database.

To test:

- download and install a production database (with migration, etc)
- yarn dev
- Observe that the "Stock Settings" item is visible under "Stock" menu.


jmcameron:
@jniles Could you please test this with the Vanga database?
Also, I'm not sure where the hex numbers in my update to the migration files come from and whether the will always be the same.   If there is a better way to retrieve those binary IDs, I would be interested  in updating this PR to do that.

jmcameron:
@jniles I think this is ready for a review and a test with the Vanga database.

server/models/migrations/next/migrate.sql
```diff
@@ -1,3 +1,20 @@
 /**
  * Migration from the version 1.16.0
- */
\ No newline at end of file
+ */
+
+/*
+ * @author: jmcameron
+ * @date: 2020-11-05 (updated 2020-11-10)
+ * @pull-release: 5084
```
jniles, line 8: Oh, this is cool.  Hadn't thought of this.

server/models/migrations/next/migrate.sql
```diff
@@ -1,3 +1,20 @@
...
+
+-- INSERT INTO `permission` (unit_id, user_id)  VALUES
+-- (290, 1),     --admin
+-- (290, 1000);  --superuser
+
+INSERT INTO `role_unit` (uuid, role_uuid, unit_id) VALUES
+-- superuser
+(HUID('BA7E12D3D16A4091A78742FF6F841C73'), HUID('7B7DD0D692734955A703126FBD504B61'), 290),
+-- admin
+(HUID('EC3847EED01E4A9096BDBA80F71C2B4A'), HUID('129BFCACD4A311E88196230B165892D5'), 290);
```
jniles, line 20: So ... these users don't actually exist in any other install but IMCK.  We generally customize the install to the particular environment, but because we were at IMCK when we started, these are stuck there.  Indeed, there are some transactions that use them and that makes them impossible to remove or else we would have removed them long ago.

So all tat being said .. I don't think this PR makes sense.

jmcameron: Ok. I'll change it to fold it into the Bhima_test SQL code.

jniles: I believe that we already give the superuser permission to everything - if it is in the unit table, we've got access to it.  So that is likely unnecessary ... unless there is a different bug going on.

Here is what I am referring to: https://github.com/IMA-WorldHealth/bhima/blob/master/test/data.sql#L333

jmcameron: I just double-checked that and realized the same thing.  So this is an unnecessary PR.
But developers installing BHIMA on production sites need to add those permissions somehow -- or train users to add them when needed.

jmcameron:
Not needed.

jmcameron:
It seems to me that we still need to do this kind of thing as SQL scripts for specific sites.   I may reopen it with a fix for IMCK.

jmcameron:
Here SQL script as a separate file for updating IMCK

[imck-stock-settings-permissions.sql.txt](https://github.com/IMA-WorldHealth/bhima/files/5520315/imck-stock-settings-permissions.sql.txt)

PR #5085, opened 2020-11-06
5067 - Fix accounts list by adding a filter

### LINKED ISSUE #5067: Do not allow a title account to be its own parent
While messing around with the accounts, I discovered this interesting corner case:

image

This is the account 4111, a title account.  As a parent, you can actually select itself!  This is bad because it will effectively remove it from the account tree, never to be seen again.

We should filter the list of proposed accounts so that we don't propose the account id under edit for the title account.

mbayopanda:
This PR fixes issue #5067 

Close #5067

### DIFF
--- a/client/src/modules/accounts/edit/accounts.edit.js
+++ b/client/src/modules/accounts/edit/accounts.edit.js
@@ -77,7 +77,7 @@ function AccountEditController(
   function handleAccountStore(accounts) {
     accountStore = angular.copy(accounts);
     accountStore.post(vm.rootAccount);
-    vm.accounts = accountStore.data;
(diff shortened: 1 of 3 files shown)


client/src/modules/accounts/edit/accounts.edit.js
```diff
@@ -77,7 +77,7 @@ function AccountEditController(
   function handleAccountStore(accounts) {
     accountStore = angular.copy(accounts);
     accountStore.post(vm.rootAccount);
-    vm.accounts = accountStore.data;
+    vm.accounts = accountStore.data.filter(item => item.id !== parseInt(id, 10));
```
jniles, line 80: This is correct. Thanks!

jniles (approved):
Solve the issue!  Great work.

bors r+

PR #5091, opened 2020-11-09
fix(bhFindEmployee): make search work

jniles:
Searching by the employee name or reference often didn't work in the bhFindEmployee component.  Here, we use the `hrlabel` like in accounts to make that more seamless.

**Before**
vSgoltfCCu

**After**
DxyARyIsK4


jniles:

lomamech (approved):
The test is no problem, success

jniles:
Bors r+

PR #5093, opened 2020-11-09
Fix distribution keys folding

### LINKED ISSUE #5006: Problem unfolding all rows in Grid

jmcameron:
Fix a problem that was preventing the Distribution Keys page from opening up fully expanded by default on Chrome.  This was affecting the End2End tests.  

Debugged with help from @jniles!

### DIFF
--- a/client/src/modules/distribution_center/distribution_key/distribution_key.js
+++ b/client/src/modules/distribution_center/distribution_key/distribution_key.js
@@ -114,6 +114,7 @@ function DistributionKeyController(DistributionCenters, Notify, uiGridConstants,
     DistributionCenters.getDistributionKey()
       .then((data) => {
         vm.gridOptions.data = data;
+        vm.grouping.unfoldAllGroups();
       })
       .catch(errorHandler)
       .finally(toggleLoadingIndicator);


jniles:
I approve this message :+1:

PR #5096, opened 2020-11-10
Feat: add multiple account report link to account references

jniles:
This PR implements a link from "account references" to the multiple account statement, auto-populating the data in the multiple account select input.  We can use this mechanism to auto-populate other reports as well (such as stock).  Here is what it looks like in action:

rQM9i2kLQM

We also need something like this to filter our balance sheet, though that is probably best left for a totally different report (https://github.com/IMA-WorldHealth/bhima/issues/5094).

### DIFF
--- a/server/controllers/finance/accounts/references.js
+++ b/server/controllers/finance/accounts/references.js
@@ -287,6 +287,22 @@ async function lookupAccountReference(id) {
   return reference;
 }
 
+/**
+ * @function getAccountsForReferenceHTTP
+ *
+ * @description
+ * HTTP interface to compute the accounts associated with a particular reference.
+ */
+async function getAccountsForReferenceHTTP(req, res, next) {
+  const { id } = req.params;
+  try {
+    const accounts = await compute.getAccountsForReference(id);
+    res.status(200).json(accounts);
+  } catch (err) {
+    next(err);
+  }
+}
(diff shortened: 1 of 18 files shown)


server/controllers/finance/accounts/references.js
```diff
@@ -288,13 +287,30 @@ async function lookupAccountReference(id) {
   return reference;
 }
 
+/**
+ * @function getAccountsForReferenceHTTP
+ *
+ * @description
+ * HTTP interface to compute the accounts associated with a particular reference.
+ */
+async function getAccountsForReferenceHTTP(req, res, next) {
```
mbayopanda, line 296: Why did you add `HTTP` in the name of the function, is that a convention?

mbayopanda (approved):
LGTM 👍

server/controllers/finance/accounts/references.js
jniles: There is another function name almost the same thing here.  So, to make sure it wasn't confusing, I named this function that takes in `req`, `res`, `next` "HTTP".  It isn't really a convention - just lack of creativity on my part.

If you can think of a  better name, I would be open to it 😄

jniles:
@mbayopanda can this be merged?

PR #5100, opened 2020-11-10
fix(stock): make adjustments one HTTP request

### LINKED ISSUE #4982: Inventory Adjustments sends a GET request per inventory item
The Inventory Adjustments function sends a GET request for each item in stock.  You can see that for each stock item, we are calling `configureItem()`.
https://github.com/IMA-WorldHealth/bhima/blob/7d310cdd6dda9adf582b40a73a9dfdac74f9207c/client/src/modules/stock/inventory-adjustment/inventory-adjustment.js#L164-L179

But the `configureItem()` function sends an HTTP request.
https://github.com/IMA-WorldHealth/bhima/blob/7d310cdd6dda9adf582b40a73a9dfdac74f9207c/client/src/modules/stock/inventory-adjustment/inventory-adjustment.js#L130-L138

This has the makings of a DOS attack on our server.  If all computers simultaneously loaded this page, it would be `(num_computers + 1) * num_items_in_inventory` requests hitting the server.

We can do better.  Let's make this page send a single request to `GET depots/:uuid/lots` that returns all the information necessary.  This way, we won't need to send thousands of HTTP requests at a time.

jniles:
Makes stock adjustments load all data in a single HTTP query. Also removes and simplifies the templates for stock adjustments.

Closes #4982.

### DIFF
--- a/client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
+++ b/client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
@@ -41,9 +41,13 @@ function StockInventoryAdjustmentController(
   vm.maxDate = new Date();
 
   // bind methods
-  vm.configureItem = configureItem;
-  vm.checkValidity = checkValidity;
   vm.submit = submit;
+
+  const expirationDateCellTemplate = `
+    <div class="ui-grid-cell-contents">
+      <span am-time-ago="row.entity.expiration_date"></span>
+    </div>
+  `;
 
   // grid columns
   const columns = [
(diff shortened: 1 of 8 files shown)


client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
```diff
@@ -161,48 +144,46 @@ function StockInventoryAdjustmentController(
...
+        let i = 0;
+
+        while (i < n) {
+          const lot = lots[i];
+          const row = vm.Stock.addItems(1);
+
+          row.configure(lot);
+
+          // mirror current quantity as old quantity
+          row.old_quantity = row.quantity;
```
mbayopanda: Does this line is still necessary since you use `Object.assign(row, { old_quantity : row.quantity })` in the line below?

mbayopanda (approved):
The fix is LGTM 👍

client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
jniles: Oh, good catch.  Hangon, let me fix it.

jniles:
Okay, it should be fixed now.

jmcameron:
bors try

jmcameron:
You may want to adjust the English translation that appears on the top right of the page.  Currently is says:

  This inventory adjustment module allows you to define article stocks; after this operation all the lots which are not defined here for the selected articles will be emptied

I suggest changing "define/defined" to "set" here.    Why is the second part needed (regarding emptying).  When would that ever occur.  Wouldn't all possible non-empty lots be listed here?  Perhaps it means setting the "New Quantity" to 0 will empty it?

jmcameron (approved):
This fix seems to work.   The code looks okay.
I tried this with the IMCK database and it appeared to run in less than 1/2 second.   But it was not much slower on the 'master' branch.

jniles:
@jmcameron thanks for your review!  You are right, the text doesn't make as much sense as when the module was first implemented.  Our functionality evolved, but our descriptions have not!  I'll make the change to note that the quantities in stock will be changed, and remove of the confusing language.

PR #5102, opened 2020-11-11
refactor(stock): entry from purchase

jniles:
This commit refactors the entry from purchase modal to use the standard receipt renderer and clarify the status code.


lomamech (commented):
No problem with this branch

PR #5103, opened 2020-11-11
refactor(stock): improve find modals

jniles:
Consolidates code related to finding depots and services for the stock exit modal.  The logic is all contained in a single file now so that we have uniform error handling between the two different modals.


mbayopanda (commented):
The feature works well for depot and service, but we have some `js` error displayed on the UI

Capture d’écran 2020-11-14 à 07 01 01

We have this error when we are trying to select a service requisition from a `depot find modal`

Also, there is a tiny typo that we can fix : 
Capture d’écran 2020-11-14 à 07 00 34

The message must be "Le bon de requisition ne concerne pas le depot en cours"

jniles:
Bors r+

PR #5104, opened 2020-11-11
bug(Out Stock)

### LINKED ISSUE #5098: Problem with the out-of-stock module
We had detected errors with the use of the out-of-stock module, in fact the following message sometimes appears when there are products available

image

lomamech:
- Use of another variable to check if the repository has inventories in
  stock for distribution instead of the vm.selectableInventories variable
  which is used in multiple times

closes #5098

### DIFF
--- a/client/src/modules/stock/exit/exit.html
+++ b/client/src/modules/stock/exit/exit.html
@@ -153,7 +153,7 @@
         ui-grid-resize-columns>
         <bh-grid-loading-indicator
           loading-state="StockCtrl.loading"
-          empty-state="StockCtrl.selectableInventories.length === 0"
+          empty-state="StockCtrl.emptyStock"
           error-state="StockCtrl.hasError"
           empty-state-message="STOCK.EMPTY">
         </bh-grid-loading-indicator>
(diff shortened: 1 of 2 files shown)


jmcameron (approved):
These changes look good.

FYI, to remind you, for the Lots registry, we created the function computeLotWarningFlags() in the LotService that computes these flags.   That way the LotService has the responsibility of figuring out these flags for the lots, which seemed appropriate.   Would it make sense to put a similar flag-setting function in the StockService and then calling that function in the load() call within exit.js?

lomamech:
> These changes look good.
> 
> FYI, to remind you, for the Lots registry, we created the function computeLotWarningFlags() in the LotService that computes these flags. That way the LotService has the responsibility of figuring out these flags for the lots, which seemed appropriate. Would it make sense to put a similar flag-setting function in the StockService and then calling that function in the load() call within exit.js?

Unless I am misunderstanding on my part, I have the impression that Stock.inventories.read (null, {depot_uuid: depot.uuid, dateTo}) returns all the necessary information concerning the lots

image

jniles (approved):
@jmcameron, I'm not sure how stock labels apply here. This is just a fix for a bug on stock exit that was introduced because of an invalid check.  It shouldn't have anything to do with lot flags.

bors r+

PR #5106, opened 2020-11-11
feat(stock): receipt component stock movements

jniles:
Adds a component similar to `bhReceipt` specifically for stock movements.  We cannot use `bhReceipt` because stock movements have wildly different receipts depending on the kind of stock movement.


lomamech (approved):
The component works very well

jniles:
Bors r+

PR #5107, opened 2020-11-11
fix(build) ensure that the manifest is correctly parsed

jniles:
Ensures that the manifest file is correctly parsed during the production build.  Only affects the "production" version of the app.

How to test this:
  1. Switch to `master`.  Run `NODE_ENV=production yarn build` in the root of the repository.  It will break.
  2. Switch to this branch.  Repeat the same command.  It will build.

This is due to breaking changes in gulp-rev-rewrite that is only called when `NODE_ENV=production`.

### DIFF
--- a/gulpfile.js/index.js
+++ b/gulpfile.js/index.js
@@ -3,6 +3,8 @@ const {
   src, dest, series, parallel,
 } = require('gulp');
 
+const { readFileSync } = require('fs');
+
 const template = require('gulp-template');
 const revRewrite = require('gulp-rev-rewrite');
 const mergeJson = require('gulp-merge-json');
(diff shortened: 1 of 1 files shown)


gulpfile.js/index.js
```diff
@@ -36,7 +38,7 @@ function collectRevisionsIntoManifest(cb) {
 // rewrite source HTML files with build versioned files and assets
 // usually run as the final step linking the build together
 function templateHTMLForProduction() {
-  const manifest = src(`${CLIENT_FOLDER}/rev-manifest.json`);
+  const manifest = readFileSync(`${CLIENT_FOLDER}/rev-manifest.json`);
   return src('client/src/index.html')
```
jmcameron, line 42: I don't understand why the "src" of rev-manifest.json did not work, but the "src" of client/src/index.html did.

jmcameron (approved):
This fix works (based on the test directions).   I did not understand why the fix was necessary, but the changes seem harmless in any case, so I approve.

gulpfile.js/index.js
jniles: The technical explanation is that `src()` creates a NodeJS stream, whereas `readFileSync()` creates a Buffer.  All of gulp works by piping streams from a `src()` to a `dest()`.  However, their documentation recommends against plugins taking in a `src()` - all plugins should use JSON/Buffer/Strings/Numbers/etc as parameters.  `rev-revision` recently updated to be compliant and it breaks this.

Why no streams in plugins?  I have no idea.  I assume there is a good explanation for it, but I haven't bothered to check.

PR #5083, opened 2020-11-05
Updated terms for minimum/security stock status labels

jmcameron:
The goal is to make these status labels easier to understand for typical users.
"Minimum Stock Reached" -> Low Stock
"Security Stock" -> Critical Stock

The French terms need to be reviewed.   

I could not figure out where STOCK.STATUS.MAXIMUM was used, so I did not change it.  

NOTE: To see these in action, I suggest using one of the production databases.

PR #5078, opened 2020-11-04
fix(reports): prune empty balance accounts

jniles:
Fixes a bug in the balance report to ensure pruning accounts takes place.  Also fixes the print orientation of the page.

PR #5047, opened 2020-10-22
Use confirmation date for purchase order

lomamech:
- Use confirmation date instead purchase date created
- Refactor of the calculation of the order interval as well as the number
  taking into account only purchase orders confirmed, received in stock
  (partially, in full, or even in excess)
- Recalculate the order interval each time the status of a purchase order is updated

closes #4831

PR #5111, opened 2020-11-12
Release v1.16.1

mbayopanda:
This release is about : 

#5107: fix(build) ensure that the manifest is correctly parsed 
#5047: Use confirmation date for purchase order 
#5078: fix(reports): prune empty balance accounts 
#5101: Updated migration script to support site-specific migration files 
#5093: Fix distribution keys folding
#5083: Updated terms for minimum/security stock status labels

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bhima",
-  "version": "1.16.0",
+  "version": "1.16.1",
   "description": "A rural hospital information management system.",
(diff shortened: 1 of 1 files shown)


jniles:
This is good!  I think we have more than this in terms of changes.   Will you make a changelog similar to https://github.com/IMA-WorldHealth/bhima/releases/tag/v1.16.0 for this release?

PR #5113, opened 2020-11-12
feat(stock): implement stock changes report

### LINKED ISSUE #5108: Feature Request:  Report to Record Changes in Inventory Value
To support stock in health zones, we'll want the health centers to bring in the quantities in stock for each product in stock so that we can record the consumption against them.  Similarly, when doing an "inventory" in your warehouse, the easiest way is to print out a copy of the current inventory with a column to allow you to record the correct values.  *It should look nearly identical to the Inventory Adjustment module*.

So, let's implement a report where you can pick...
  1. A depot
  2. A date

...and it will produce a report that should looks like the stock adjustment module, but with blank cells for the "new quantity".  This will allow users to print it, record changing values, and then key it in to the software later.

jniles:
Implements the "stock changes" report to allow users to download the
current inventory in a depot as of the end of a period.  This should
allow them to fill out the details and quickly compute the quantities in
stock to then make stock adjustments.

Closes #5108.

image

### DIFF
--- a/client/src/i18n/en/report.json
+++ b/client/src/i18n/en/report.json
@@ -479,6 +479,10 @@
       "PAYMENTS" : "Payments",
       "AVERAGE_COST" : "Average Cost",
       "RECOVERY_CAPACITY" : "Recovery Capacity"
+    },
+    "STOCK_CHANGES" : {
+      "TITLE" : "Stock Changes Report",
+      "DESCRIPTION" : "This report allows you to record changes to the stock of a depot by comparing the amount in the system to the amount found on the shelves."
     }
   }
 }
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -15,6 +15,7 @@
     "DECREASE_HELP"   : "Adjust stock by decreasing the quantity of lots of products in stock",
     "DEPOT"           : "Depot",
     "DEPOT_RESPONSIBLE" : "Depot Manager",
+    "DAYS_OF_STOCK_OUT" : "# Days of Stock Out",
     "DESTINATION"     : "Destination",
     "DETAILS"         : "Details",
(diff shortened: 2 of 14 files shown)


jniles:
@lomamech can I get a review?

lomamech:
@jniles Keys are missing for the translation
image

lomamech (commented):
@jniles I emptied the Test database, and I carried out integrations for the month of January, suide consumption and losses, but I notice that the data on consumption and losses do not appear ?

image

client/src/i18n/en/report.json
```diff
@@ -479,6 +479,10 @@
       "PAYMENTS" : "Payments",
       "AVERAGE_COST" : "Average Cost",
       "RECOVERY_CAPACITY" : "Recovery Capacity"
+    },
```
lomamech, line 482: These keys did not exist in the (fr) file

lomamech:
:warning: :warning: :warning:Let me understand the purpose of this report is to print a document that will allow you to enter the situation in stock compared to the data of the month and by batches (Consumption, losses, days of breaks and remaining quantities) ????? :warning: :warning: :warning:

jniles:
@lomamech thank you for the review. Yes, that is correct - the report should be blank so that users can download the report, print it out, and then be able to fill in the correct values over a month.

lomamech:
> @lomamech thank you for the review. Yes, that is correct - the report should be blank so that users can download the report, print it out, and then be able to fill in the correct values over a month.

Add the keys for the missing translation to conclude this PR

client/src/i18n/en/report.json
jniles: Great catch!  Thanks!

jniles:
@lomamech, changes landed:

image

jniles:
Bors r+

PR #5118, opened 2020-11-13
Feature dashboard stock

lomamech:
- Add indicator panel
- Addition of a loading when downloading data
- Complete panel for Dashbord
- Add link to registry of stock or lot


mbayopanda:
@lomamech is it possible to add screenshots of the UI in the description of this PR?

client/src/css/structure.css
```diff
@@ -888,19 +888,35 @@ a[disabled] {
   background-color: #ff000083 !important;
 }
 
+.icon-expired {
+  color: #ff000083 !important;
```
mbayopanda: - Is it necessary to add `!important`?
- Why using the HEX8 (#XXXXXXXX) color notation in place of the well-known HEX color (#XXXXXX)?

server/controllers/stock/index.js
```diff
@@ -591,6 +594,167 @@ function listMovements(req, res, next) {
...
+
+  const getDepotsByUser = `
+    SELECT BUID(p.depot_uuid) AS depot_uuid, d.text AS depot_text
+    FROM depot_permission p
+      JOIN depot AS d
+      ON d.uuid = p.depot_uuid
+    WHERE p.user_id = ?
+    ORDER BY d.text ASC`;
+
+  db.exec(getDepotsByUser, [req.session.user.id])
```
mbayopanda, line 618: Interesting

client/src/modules/stock/dashboard/dashboard.js
```diff
@@ -0,0 +1,20 @@
...
+ * @function StockDashBoardController
+ *
+ * @description
+ * This controller binds the basic CRUD operations on the enterprise.
+ */
+function StockDashBoardController() {
+  // fired on startup
+  function startup() {
+
+  }
```
mbayopanda: ?

client/src/modules/stock/dashboard/dashboard.html
```diff
@@ -0,0 +1,74 @@
...
+      <div class="col-md-6">
+        <bh-dashboard-stock
+          label="STOCK.REQUIRES_PO"
+          status="require_po">
+        </bh-dashboard-stock>
+      </div>
+    </div>
+
+    <div class="row">
+      <div class="col-md-6">
```
mbayopanda: Est-ce possible d'essayer avec `col-md-4` pour avoir trois colonnes de cette maniere on pourrait gagner en hauteur

mbayopanda (commented):
The dashboard looks nice 👍 

I left some comments

jmcameron (commented):
I like the concept.  However, I suggest that you actually list the items out on the Dashboard (at least a limited number).  Having the link to see the details is great, but showing some abbreviated details here would be good.
image

Note that the "Stock Exit" page has similar lists which could be adapted for this dashboard.

I glanced over the code and it looks okay.

client/src/css/structure.css
lomamech: I discovered it for the first time in this file, it is in several classes, this will force that the defined values have priority over any other value that will be defined

client/src/modules/stock/dashboard/dashboard.html
lomamech: Let me try

lomamech:
> I like the concept. However, I suggest that you actually list the items out on the Dashboard (at least a limited number). Having the link to see the details is great, but showing some abbreviated details here would be good.
> image
> 
> Note that the "Stock Exit" page has similar lists which could be adapted for this dashboard.
> 
> I glanced over the code and it looks okay.
This option will be very easy, if there were not too many deposits in Tshikaji for example there are a lot of depots and with the projects of BCZs and Health Centers there will be many, if it will be necessary to list even 10 products per depot, the list may be too long

Here is an overview with the Tshikaji database
bhima

client/src/modules/stock/dashboard/dashboard.html
lomamech: image

client/src/modules/stock/dashboard/dashboard.js
lomamech: See that everything is done with the component, there is unfortunately nothing inside the model

client/src/css/structure.css
jmcameron: The last two hex digits in the long format control the transparency of the item.  So '83' means Transparency should be 83/256 (or 32% transparent).

jmcameron: See https://css-tricks.com/8-digit-hex-codes/

jmcameron:
> I suggest that you actually list the items out on the Dashboard

I withdraw my suggestion.  I did not realize that this for the multi health center/depot info.  Spelling out the details for each depot for multiple sites would be too much detail and not with the upper-level manager needs.

client/src/modules/stock/dashboard/dashboard.html
jmcameron: Thanks for including this image.  I did not run it with  the IMCK database, so it was helpful to see this tableau.

client/src/js/components/bhDashboardStock.js
```diff
@@ -0,0 +1,109 @@
+angular.module('bhima.components')
+  .component('bhDashboardStock', {
+    templateUrl : 'modules/templates/bhDashboardStock.tmpl.html',
+    controller  : DashboardStock,
+    transclude  : true,
```
jniles: Do you use transclude?  It doesn't look like it.  Could you remove this?

client/src/js/components/bhDashboardStock.js
```diff
@@ -0,0 +1,109 @@
...
+];
+
+/**
+ * Dashboard Stock Controller
+ *
+ */
+function DashboardStock(StockDashBoard, Notify) {
+  const $ctrl = this;
+
+  $ctrl.$onInit = function onInit() {
```
jniles: Hmm... instead of having this parsing, it probably makes better sense to break up a lot of these into separate, small components.  Like `bh-stock-panel-expired` and `bh-stock-panel-out-of-stock`.  That way, if there is a bug in one of these, we have an easier time fixing it and do not have to parse through a ton of other code.

client/src/js/components/bhDashboardStock.js
```diff
@@ -0,0 +1,109 @@
...
+    templateUrl : 'modules/templates/bhDashboardStock.tmpl.html',
+    controller  : DashboardStock,
+    transclude  : true,
+    bindings    : {
+      status  : '@?',
+      label   : '@?',
+    },
+  });
+
+DashboardStock.$inject = [
```
jniles: Can you add the word "Controller" to this?

client/src/js/components/bhDashboardStock.js
```diff
@@ -0,0 +1,109 @@
...
+      $ctrl.keyInventoryFilter = 'require_po';
+      $ctrl.label = 'STOCK.REQUIRES_PO';
+    } else if ($ctrl.status === 'minimum_reached') {
+      $ctrl.display = 'fa fa-battery-quarter text-info';
+      $ctrl.keyInventoryFilter = 'minimum_reached';
+      $ctrl.label = 'STOCK.STATUS.MINIMUM';
+
+    }
+
+    StockDashBoard.read({ status : $ctrl.status })
```
jniles: Let's call this "StockDashboard", since Dashboard is one word.

client/src/modules/stock/dashboard/dashboard.routes.js
```diff
@@ -0,0 +1,9 @@
+angular.module('bhima.routes')
+  .config(['$stateProvider', $stateProvider => {
+    $stateProvider
+      .state('/stock/dashboard', {
+        url : '/stock/dashboard',
+        controller : 'StockDashBoardController as StockDashBoardCtrl',
```
jniles, line 6: You don't need to have a controller for each view.  It looks like you invented a controller just to put it here.  If the view doesn't have a controller just put in the templateUrl and that should be fine.

server/controllers/stock/index.js
jniles, line 618: I like this.  In the long run, I can see this code being put in the users controller, then we can have `/users/:id/depots` as a route to get the depots associated with that user.

jniles (commented):
Just a few comments here.

client/src/js/components/bhDashboardStock.js
lomamech: no problem

jniles (approved):
LGTM.

bors r+

PR #5120, opened 2020-11-14
Fix stockImport procedure

mbayopanda:
This PR is about : 

- Update the `ImportStock` procedure to fix : 
  - MySQL data too long error when the `stockLotQuantity` and `inventoryUnitCost` are not in the good format
  - Error related to `inventory_uuid` is null during the importation
  - Error related to `inventory_unit` with its `abbr` column which must be unique during the importation

- Add inventories to `inventory_group` defined by its code instead of the inventory group name in the template file

NOTE:
- The CMM is not changed by the import since we are using the `stock_movement_status` table, the CMM defined in the template file is successfully saved in the `avg_consumption` column of the inventory.

### DIFF
--- a/server/controllers/stock/import.js
+++ b/server/controllers/stock/import.js
@@ -103,29 +103,83 @@ async function importStock(req, res, next) {
 function checkDataFormat(data = []) {
   for (let i = 0; i < data.length; i++) {
     const item = data[i];
-    const isUnitPriceDefined = typeof (item.inventory_unit_price) !== 'undefined';
+    const isInventoryGroupDefined = typeof (item.inventory_group_name) === 'string'
+      && item.inventory_group_name.length > 0;
+    const isInventoryTextDefined = typeof (item.inventory_text) === 'string' && item.inventory_text.length > 0;
+    const isInventoryTypeDefined = typeof (item.inventory_type) === 'string' && item.inventory_type.length > 0;
+    const isInventoryUnitDefined = typeof (item.inventory_unit) === 'string' && item.inventory_unit.length > 0;
+    const isStockLotLabelDefined = typeof (item.stock_lot_label) === 'string' && item.stock_lot_label.length > 0;
+    const isExpirationDefined = typeof (item.stock_lot_expiration) === 'string' && item.stock_lot_expiration.length > 0;
+
     const isUnitPriceNumber = !Number.isNaN(Number(item.inventory_unit_price));
     const isLotQuantityNumber = !Number.isNaN(Number(item.stock_lot_quantity));
-    const validity = item.inventory_group_name
-      && item.inventory_text && item.inventory_type && item.inventory_unit
-      && isUnitPriceDefined && item.stock_lot_label
-      && item.stock_lot_quantity && item.stock_lot_expiration
-      && isUnitPriceNumber && isLotQuantityNumber;
+
+    /**
+     * The key parameter of BadRequest must be properly translated for the user
+     */
+
+    if (!isInventoryGroupDefined) {
+      throw new BadRequest(
+        `[line : ${i + 2}] The inventory group ${item.inventory_group_name} must be a valid text`,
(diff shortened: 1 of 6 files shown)


server/controllers/stock/import.js
```diff
@@ -94,6 +94,28 @@ function importStock(req, res, next) {
...
+
+/**
+ * @function formatDecimal
+ * @param {string} quantity
+ * @description
+ * remove number formatting commas and convert into a decimal (float)
+ * the decimal separator accepted is . not ,
+ */
+function formatDecimal(quantity) {
+  return typeof (quantity) === 'string' ? parseFloat(quantity.replace(/,/g, '')) : quantity;
```
jniles: What happens if someone submits "10,00"?  Will this translate it into "1000" or "10.00"?  It looks like the former when it should be the second one.

jniles (commented):
I have just one question on the parsing of decimals/integers.  Otherwise, this is fine.  We can probably improve it a lot if we are going to rely on it.

server/controllers/stock/import.js
mbayopanda: 1000

mbayopanda: Maybe we must find a best way to handle that cause the user can set in the quantity column : "10,000.00" and "10,05" for example, the current implementation accepted the US formatting "." as a separator of decimal

jniles: Could we just throw an error if it is not in the USD formatting?  In other words, if it isn't a JS Number type, we just break?  That way, we don't accidentally make a mistake in our parsing.

server/controllers/stock/import.js
```diff
@@ -34,79 +34,98 @@ function downloadTemplate(req, res, next) {
...
+    const isLotQuantityNumber = !Number.isNaN(Number(item.stock_lot_quantity));
+    const validity = item.inventory_group_name
       && item.inventory_text && item.inventory_type && item.inventory_unit
       && isUnitPriceDefined && item.stock_lot_label
-      && item.stock_lot_quantity && item.stock_lot_expiration;
-  });
+      && item.stock_lot_quantity && item.stock_lot_expiration
+      && isUnitPriceNumber && isLotQuantityNumber;
+
+    if (!isUnitPriceNumber) {
```
jmcameron: I like this kind of check!

server/controllers/stock/import.js
```diff
@@ -34,79 +34,98 @@ function downloadTemplate(req, res, next) {
...
+    const isUnitPriceNumber = !Number.isNaN(Number(item.inventory_unit_price));
+    const isLotQuantityNumber = !Number.isNaN(Number(item.stock_lot_quantity));
+    const validity = item.inventory_group_name
       && item.inventory_text && item.inventory_type && item.inventory_unit
       && isUnitPriceDefined && item.stock_lot_label
-      && item.stock_lot_quantity && item.stock_lot_expiration;
-  });
+      && item.stock_lot_quantity && item.stock_lot_expiration
+      && isUnitPriceNumber && isLotQuantityNumber;
+
```
jmcameron, line 116: You should consider added more checks for each field:  checking that the text fields have text, that numbers have numbers (such as stock_lot_quantity, etc).  I know that writing the code to check each item is a bit tedious, but it will be a tremendous help to users down the road when you tell them exactly what is wrong with their file (instead of "bad data format").

jmcameron (approved):
Can we add some tests to make sure the data checking works?

jniles:
This may be helpful: https://github.com/validatorjs/validator.js

server/controllers/stock/import.js
jniles: This may be helpful: https://github.com/validatorjs/validator.js

jmcameron (approved):
Excellent change!   I was thinking about line numbers and voila! you have them.   Thanks for the update.

jniles:
Bors r+

PR #5121, opened 2020-11-17
improvement(Stocks lots page)

### LINKED ISSUE #5099: Need color legend for Stock Lots page
We need a color legend for the colors of the rows in the Stock Lots page. 
This includes:

- No color - Normal stock status
- Yellow - A risk (of stock out)
- Red - At risk of expiration
- Purple - Out of Stock
- Blue - Over maximum (perhaps?)

Note that we should also re-examine the color scheme here and the color of the stock status labels on the Items in Stock page to see if they can be made consistent.

lomamech:
- Add legend for the colors of the rows

closes #5099

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -897,7 +897,7 @@ a[disabled] {
 }
 
 .icon-expired {
-  color: #ff000083 !important;
+  color: #FF7B7B !important;
 }
 
 .at-risk {
@@ -909,7 +909,7 @@ a[disabled] {
 }
 
 .icon-at-risk {
-  color: #ffff6686 !important;
+  color: #FFFFA9 !important;
 }
 
 .at-risk-of-expiring {
@@ -921,7 +921,7 @@ a[disabled] {
 }
 
 .icon-at-risk-of-expiring {
-  color: #ff880191 !important;
+  color: #FFB766 !important;
 }
 
 .out-of-stock {
@@ -934,7 +934,7 @@ a[disabled] {
 }
 
 .icon-out-of-stock {
-  color: #cc00ff8f !important;
+  color: #E067FF !important;
 }
 
 .dropdown-menu {
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -896,12 +896,20 @@ a[disabled] {
   color: #FF7B7B !important;
 }
 
+.icon-expired {
+  color: #ff000083 !important;
+}
+
 .at-risk {
   background-color: #FFFFA9 !important;
 }
 
 .icon-at-risk {
   background-color:#FFFFA9 !important;
+}
+
+.icon-at-risk {
+  color: #ffff6686 !important;
 }
 
 .at-risk-of-expiring {
@@ -912,6 +920,10 @@ a[disabled] {
   color: #FFB766 !important;
 }
 
+.icon-at-risk-of-expiring {
+  color: #ff880191 !important;
+}
+
 .out-of-stock {
   /* text-decoration : line-through; */
   background-color: #E067FF !important;
@@ -919,6 +931,10 @@ a[disabled] {
 
 .icon-out-of-stock {
   color: #E067FF !important;
+}
+
+.icon-out-of-stock {
+  color: #cc00ff8f !important;
 }
 
 .dropdown-menu {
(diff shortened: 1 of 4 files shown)


lomamech:
image

client/src/css/structure.css
```diff
@@ -888,19 +888,35 @@ a[disabled] {
   background-color: #ff000083 !important;
 }
 
+.icon-expired {
+  color: #ff000083 !important;
```
jmcameron: I encourage you to just use the 6-digit version here.  You do not need or want any transparency in the legend!

client/src/css/structure.css
```diff
@@ -888,19 +888,35 @@ a[disabled] {
...
+}
+
 .at-risk {
   background-color: #ffff6686 !important;
 }
 
+.icon-at-risk {
+  color: #ffff6686 !important;
```
jmcameron: I encourage you to just use the 6-digit version here.  You do not need or want any transparency in the legend!

client/src/css/structure.css
```diff
@@ -888,19 +888,35 @@ a[disabled] {
...
+}
+
 .at-risk-of-expiring {
   background-color: #ff880191 !important;
 }
 
+.icon-at-risk-of-expiring {
+  color: #ff880191 !important;
```
jmcameron: I encourage you to just use the 6-digit version here.  You do not need or want any transparency in the legend!

client/src/css/structure.css
```diff
@@ -888,19 +888,35 @@ a[disabled] {
...
+}
+
 .out-of-stock {
   /* text-decoration : line-through; */
   background-color: #cc00ff8f !important;
 }
 
+.icon-out-of-stock {
+  color: #cc00ff8f !important;
```
jmcameron: I encourage you to just use the 6-digit version here.  You do not need or want any transparency in the legend!

jmcameron (approved):
I like the legend.  I have a couple of suggestions

- The reason that we (maybe me) used the 8-digit HEX codes for the colors for highlighting the rows is so that the transparency specified would make the row easier to read, but still retain the desired color.   However, for the icons in the legend, you should use the 6-digit HEX color code so that there is no transparency in order to make the legends easier to read.
- The legend is a bit small at the bottom of the screen.  Perhaps we can use a larger font.  Also perhaps putting it on its own row (a separate line) would allow you to put more whitespace above and below it to make it more visible.

jmcameron:
By the way, here is some CSS I was playing around with to try to put a black border around the fa-circle to make the yellow circle easier to see.  I do not suggest you use this since it may not work the same on all browsers and it is not quite perfect.  But it was fun to figure out (https://stackoverflow.com/questions/33277777/css-circle-with-border):

```cs
.icon-at-risk {
  color: #fdfd00 !important;
  background-color: #fdfd00;
  border:1px solid black;
  height:2.1ex;
  border-radius:50%;
  -moz-border-radius:50%;
  -webkit-border-radius:50%;
  width:2.1ex;
}
```

client/src/css/structure.css
jniles: I agree with this suggestion.

client/src/modules/stock/lots/registry.service.js
```diff
@@ -154,6 +154,21 @@ function LotsRegistryService(uiGridConstants, Session) {
     <div>
       <b>{{ grid.appScope.countGridRows() }}</b>
       <span translate>TABLE.AGGREGATES.ROWS</span>
+
+      <span class="fa fa-circle icon-expired"></span>
```
jniles: This is a really good idea @lomamech.  You found the perfect place to put this.

jniles (commented):
I like both of @jmcameron's recommendations - making the colors non-transparent and then also adding a black border.  Can you implement those changes? 

This looks really good.

client/src/modules/stock/lots/registry.service.js
jmcameron: See if the tests pass now

bors try

jmcameron:
bors try

See if tests pass now

jniles:
@lomamech it looks like this needs to be updated to reflect the changes in the dashboard PR that just landed.

lomamech:
> @lomamech it looks like this needs to be updated to reflect the changes in the dashboard PR that just landed.

I'm working on this, and I understood that I have to restore the old configuration of the CSS file which handled the colors

lomamech:
image

jniles (approved):
LGTM.

bors r+

PR #5122, opened 2020-11-17
improvement(Break Even Report)

### LINKED ISSUE #4940: Improve formatting in Break Even Report configuration page
Here is the configuration page for the breakeven analysis report:

image

We should eliminate the whitespace indicated by the red boxes.  We should also remove the `<hr>` element as they don't really add much.  Finally, all the bold checkboxes and radio buttons don't need to be bold.  We generally use bold for labels.

lomamech:
- Eliminate whitespace and not necessary `<hr>`
- Remove property bold on checkboxes and radios buttons

closes #4940


jniles (approved):
This is excellent.  Great work @lomamech 👍

jniles:
Bors r+

PR #5124, opened 2020-11-17
Fixed bugs in editing of user roles

### LINKED ISSUE #5123: Roles and permissions bug
**Two bugs:**

1. Under Administration > User and Permissions Management, when you edit the Roles via the "Assign Role(s)" item under the Action menu for the user, the dropdown is not pre-filled in with the user's current roles.

2. When you edit a user's Role(s) (as in 1 above) and you click the [Submit] button on the Roles modal, the list of roles on the main page is not refreshed.

jmcameron:
Fixed bugs that occurred when editing user roles.

**To Test:**

- Use any site (Bhima_test is fine)
- Go to Administration > User and Permissions Management
- Edit the Roles for a user by using the "Assign Role(s)" item under the Action menu for the user
  - Verify that the selected roles are filled out with the previous values
- Change the selection of roles
- Click the [Submit] button
- Verify that the new roles are updated properly on the main grid of users, roles, depots, and cashboxes

### DIFF
--- a/client/src/modules/roles/modal/userRole.html
+++ b/client/src/modules/roles/modal/userRole.html
@@ -10,6 +10,7 @@
     is-flat-tree="true"
     data="UsersRolesCtrl.roles"
     id-key="uuid"
+    checked-ids="UsersRolesCtrl.selected"
     on-change="UsersRolesCtrl.onChangeRoleSelection(data)">
   </bh-checkbox-tree>
 </div>
--- a/client/src/modules/roles/modal/userRole.js
+++ b/client/src/modules/roles/modal/userRole.js
@@ -18,6 +18,9 @@ function UsersRolesController(data, $uibModal, $uibModalInstance, RolesService, 
     RolesService.userRoles(vm.user.id)
       .then(roles => {
         vm.roles = roles;
+        vm.selected = roles
+          .filter(role => role.affected === 1)
+          .map(role => role.uuid);
       })
       .catch(Notify.handleError);
   }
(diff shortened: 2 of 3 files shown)


client/src/modules/roles/modal/userRole.html
```diff
@@ -10,6 +10,7 @@
     is-flat-tree="true"
     data="UsersRolesCtrl.roles"
     id-key="uuid"
+    checked-ids="UsersRolesCtrl.selected"
```
jniles, line 13: This is exactly right.  Well done!  :+1:

client/src/modules/roles/modal/userRole.js
```diff
@@ -18,6 +18,9 @@ function UsersRolesController(data, $uibModal, $uibModalInstance, RolesService,
     RolesService.userRoles(vm.user.id)
       .then(roles => {
         vm.roles = roles;
+        vm.selected = roles
+          .filter(role => role.affected === 1)
```
jniles, line 22: Just a note for a future date; nothing to do with this PR.

We have role "assignments", but sometimes we use the word "affected".  This is a mix of French and English.  We might want to change the names of "affected" variables to "assigned".

client/src/modules/users/users.js
```diff
@@ -161,7 +161,10 @@ function UsersController($state, $uibModal, Users, Notify, Modal, uiGridConstant
           return user;
         },
       },
+    }).result.then(() => {
+      loadGrid();
```
jniles, line 165: Looks good!  I prefer the grid refresh to trying to edit the data locally, so this is my preferred fix.  Thanks!

jniles (approved):
Looks good to me.  Great work!

PR #5130, opened 2020-11-19
4834 - Fix calculation and status about expired and at risk lots

### LINKED ISSUE #4834: Stock Expiration Report should warn about expired stock
The Stock Expiration report doesn't show stock that is expired as "expired".  Instead it shows "At Risk of Expiration".  We should change this behavior.

Here is an example from the test dataset:

image

mbayopanda:
This PR fixes the report "Stock Expiration Report" by : 

- Adding new columns: CMM and quantity_at_risk
- Fix the way of fetching risky lots
- Fix calculation of the values of lots at risk
- Correctly display expired and risky lots in the table
- Adding the date of the day in the report 

image

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -296,21 +296,21 @@ async function getLotsDepot(depotUuid, params, finalClause) {
   }
 
   const inventoriesWithManagementData = await stockManagementProcess(resultFromProcess);
-  const inventoriesWithLotsProcessed = await processMultipleLots(inventoriesWithManagementData);
+  let inventoriesWithLotsProcessed = await processMultipleLots(inventoriesWithManagementData);
 
   if (_status) {
-    return inventoriesWithLotsProcessed.filter(row => row.status === _status);
+    inventoriesWithLotsProcessed = inventoriesWithLotsProcessed.filter(row => row.status === _status);
   }
 
   // Since the status of a product risking expiry is only defined
   // after the comparison with the CMM, reason why the filtering
   // is not carried out with an SQL request
-  if (params.is_expiry_risk === '1') {
-    return inventoriesWithLotsProcessed.filter(item => (item.S_RISK < 0 && item.lifetime > 0));
+  if (parseInt(params.is_expiry_risk, 10) === 1) {
+    inventoriesWithLotsProcessed = inventoriesWithLotsProcessed.filter(item => (item.S_RISK < 0 && item.lifetime > 0));
   }
 
-  if (params.is_expiry_risk === '0') {
-    return inventoriesWithLotsProcessed.filter(item => (item.S_RISK >= 0 && item.lifetime > 0));
+  if (parseInt(params.is_expiry_risk, 10) === 0) {
+    inventoriesWithLotsProcessed = inventoriesWithLotsProcessed.filter(item => (item.S_RISK >= 0 && item.lifetime > 0));
   }
 
   return inventoriesWithLotsProcessed;
--- a/server/controllers/stock/reports/stock/expiration_report.js
+++ b/server/controllers/stock/reports/stock/expiration_report.js
@@ -93,7 +93,7 @@ async function stockExpirationReport(req, res, next) {
     });
 
     const reportResult = await report.render({
-      depot, result : values, totals,
+      depot, result : values, totals, today,
     });
 
     res.set(reportResult.headers).send(reportResult.report);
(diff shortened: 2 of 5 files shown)


server/controllers/stock/reports/stock_expiration_report.handlebars
```diff
@@ -58,7 +62,9 @@
           <th>{{translate 'STOCK.INVENTORY'}}</th>
           <th>{{translate 'STOCK.LOT'}}</th>
           <th>{{translate 'STOCK.EXPIRATION'}}</th>
-          <th>{{translate 'STOCK.QUANTITY'}}</th>
+          <th>{{translate 'STOCK.CMM'}}</th>
+          <th>{{translate 'STOCK.CURRENT_QUANTITY'}}</th>
+          <th>{{translate 'STOCK.RISK_QUANTITY'}}</th>
           <th>{{translate 'TABLE.COLUMNS.UNIT'}}</th>
           <th>{{translate 'REPORT.STOCK.UNIT_COST'}}</th>
           <th>{{translate 'FORM.LABELS.VALUE'}}</th>
```
jniles: Is this the value at risk?  If so, should we label it so?

server/controllers/stock/reports/stock_expiration_report.handlebars
```diff
@@ -72,7 +78,9 @@
             <td>{{lot.text}}</td>
             <td>{{lot.label}}</td>
             <td>{{date lot.expiration_date}}</td>
+            <td class="text-right">{{lot.cmms.algo_msh}}</td>
```
jniles: Can we use the defined  algorithm?  It looks like this will always be MSH, but it should be set in the settings.

server/controllers/stock/reports/stock/expiration_report.js
```diff
@@ -83,7 +93,7 @@ async function stockExpirationReport(req, res, next) {
     });
 
     const reportResult = await report.render({
-      depot, result : values, totals,
+      depot, result : values, totals, today,
```
jniles, line 96: Good change 👍

server/controllers/stock/core.js
```diff
@@ -296,21 +296,21 @@ async function getLotsDepot(depotUuid, params, finalClause) {
   }
 
   const inventoriesWithManagementData = await stockManagementProcess(resultFromProcess);
-  const inventoriesWithLotsProcessed = await processMultipleLots(inventoriesWithManagementData);
+  let inventoriesWithLotsProcessed = await processMultipleLots(inventoriesWithManagementData);
 
   if (_status) {
-    return inventoriesWithLotsProcessed.filter(row => row.status === _status);
+    inventoriesWithLotsProcessed = inventoriesWithLotsProcessed.filter(row => row.status === _status);
```
jniles, line 302: Oh la la.  Good catch.

jniles (commented):
Looks pretty good to me!  I made a suggestion on changing the text to make the quantity at risk explicit.

jmcameron (approved):
I tried testing this with the IMCK database.  I looked through all the expiration report and could not find any items that were mislabeled. 

However, I notice a lot of items that have a CMM=0 which are therefore marked "at risk of expiration".  That makes sense from a technical perspective. But in reality, that actually means that none have been dispensed in the last N(6) months.  This means one of the following:
 (a) It is normally used but not in the last 6 months.  At this rate it is really "at risk of expiration"
 (b) It is no longer needed (perhaps a new drug has replaced it)
 (c) doctors/nurses are not aware it is in stock or that it can be used for current needs

We are currently assuming (a) is the only possibility.   
However, if it is (b) it might be useful to be able to mark the stock as obsolete?  
If (c) is true, then there is a training issue to be dealt with.  Fixing that is more complicated; the BHIMA software could assist with suggesting alternate meds when one is requested that an alternate nearing expiration is available, but that is complicated and requires a significant amount of alternative drug info or training.

jmcameron:
@jniles : I suggest you read my comment above (about "at risk of expiration" stock where CMM=0).

jniles:
@jmcameron this is a really good analysis.  I would propose that we make cmm === 0 (or days of consumption  === 0) with a different flag.  We will call it "unused medication" since we don't know which scenarios (a, b, c) it fits in.

However, we might decide on a fixed window before the expiration date to flip it back to "at risk".  Something like 3 months is probably good.  Because at that point, the medication will expire even if you are not using it and that is likely not a good thing at all.

mbayopanda:
@jmcameron and @jniles with IMCK database it is important to run the sql query `CALL zComputeAllInventoryStockQuantities(DATE('2020-01-01'), 0x8A37045FED5D40239D2ED7A4A90BD2CE);` in  order to recalculate values in stock_movement_status.

And the uuid `0x8A37045FED5D40239D2ED7A4A90BD2CE)` is for a given depot

jniles (approved):
LGTM.  

bors r+

PR #5143, opened 2020-11-24
Fix stock performance issues

### LINKED ISSUE #4711: Optimize stock/inventories/depots route
I recently recovered the logs from a raspberry pi running in production, and here is what I found:

image
_Fig 1: Slow `/stock/inventories/depots` route_

It looks like GET requests to the `/stock/inventories/depots` route never complete.  Normally, there should be a number of milliseconds at the end of this file.  But there isn't, meaning the GET request never completed.

We should look for ways to optimize this.

jniles:
This PR addresses two performance issues:
 1. We cache the results of the stock out/expired stock components on the depot page to prevent multiple calls to that route from hitting the server simultaneously.
 2. We've removed calls to `stock_consumption` in the stock controller and merged the calls to `getCMM` into a single function. Then we optimized the function to only call `getCMM()` for unique depot/inventory pairs.

While doing so, I've refactored the CMM calculations in stock/core.js to make the more streamlined and rename generic sounding names to more meaningful names .  There are two basic paths through core.js:

1. `getBulkInventoryCMM()` -> `computeInventoryIndicators()`
2. `getBulkInventoryCMM()` -> `computeInventoryIndicators()` -> `computeLotIndicators()`

The first path computes S_MAX, S_MIN, and other _inventory_ indicators that relate to the change of stock out.  This is good enough for the "Items in Stock" registry and other reports that need to see the data on an inventory level.

The second path does everything in the first, but also computes the S_RISK and S_RISK_QUANTITY and other _lot_ indicators at are related to the risk of expiration of an individual lot.

Closes https://github.com/IMA-WorldHealth/bhima/issues/4711
Partially addresses https://github.com/IMA-WorldHealth/bhima/issues/5127
Partially addresses https://github.com/IMA-WorldHealth/bhima/issues/5073

### DIFF
--- a/client/src/js/components/bhStockSoldOut/bhStockSoldOut.html
+++ b/client/src/js/components/bhStockSoldOut/bhStockSoldOut.html
@@ -4,27 +4,27 @@
 
   <table
     class="table table-condensed table-bordered"
-    ng-show="$ctrl.soldOutInventories.length > 0">
+    ng-show="$ctrl.stockOutInventories.length > 0">
     <thead>
       <tr>
         <th>
           <span class="text-danger"
-            translate-values="{ total : $ctrl.soldOutInventories.length }"
+            translate-values="{ total : $ctrl.stockOutInventories.length }"
             translate="{{ 'STOCK.STOCK_OUT_WARNING' }}">
           </span>
         </th>
       </tr>
     </thead>
     <tbody>
-      <tr ng-repeat = "inventory in $ctrl.soldOutInventories | limitTo : 5">
-        <td>{{inventory.text}} (<u translate-attr="{ 'title' : inventory.stock_out_date_raw }">{{inventory.stock_out_date}}</u>)</td>
+      <tr ng-repeat = "inventory in $ctrl.stockOutInventories | limitTo : 5">
+        <td>{{inventory.text}} (<u translate-attr="{ 'title' : inventory.stock_out_date_raw }">{{inventory.stock_out_date_parsed}}</u>)</td>
       </tr>
 
       <tr>
         <td>
           <span class="text-danger"
-            ng-show="($ctrl.soldOutInventories.length - 5) > 0"
-            translate-values="{ left : ($ctrl.soldOutInventories.length - 5)}"
+            ng-show="($ctrl.stockOutInventories.length - 5) > 0"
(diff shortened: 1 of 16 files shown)


jniles:

I discovered that on master, the months of stock for the CMM was 5, according to the `getCMM()` function.  I think this was a bug - it should have been 6, since that is what we put in the database.  That is why I had to update a few tests.

jniles:
@mbayopanda can this be merged?

jmcameron:
I would like to review this PR, but I'm unclear how to evaluate its performance improvement.
What DB and Bhima page best shows the speed up?

jniles:
@jmcameron, stock exit is the best example of this.  Here are the things to look for:

 1. When the page is refreshed (F5), there was two HTTP requests sent for each route.   Now there are one.
 2. You'll observe a minor performance improvement in the queries, but this could just be noise as well.

jniles:
Another thing you can do, I suppose is to print the DB logs (`DEBUG=db*`) and see that we call `getCMM()` _much_ less frequently.

client/src/js/components/bhStockSoldOut/bhStockSoldOut.html
```diff
@@ -4,27 +4,27 @@
 
   <table
     class="table table-condensed table-bordered"
-    ng-show="$ctrl.soldOutInventories.length > 0">
+    ng-show="$ctrl.stockOutInventories.length > 0">
```
jmcameron, line 7: Better variable name!

jmcameron (approved):
Looks good.   My review of the code was more like a skim of changes, but the changes seem okay.  I tested the latest BHIMA and IMCK data with and without this PR.  The number of calls to getCMM() is definitely reduced with this PR.  But there are still possibly some duplicate calls.  I noticed a definite speed up.   With the IMCK data, the "loading data" part of the Stock exit page took 2-3 seconds with master and was less than a second with this PR.   So this is a worthwhile refactor.

jniles:
@jmcameron thanks for your review!  I'm going to include this in the quick fix release we do at IMCK today, since our production users will hammer it and let us know if there are any breaking issues.  Much better to find that our while we can work with them here than have them call us from afar...

PR #5150, opened 2020-11-25
bug(Stock Exit)

lomamech:
- Addition in the function reinit() of the call of the function
  loadInventories for a depot to restore to reload
the products available for the depot


lomamech:
@jniles it's done

jniles:
Lgtm. 

Bors r+

PR #5151, opened 2020-11-25
Add option to display negative quantities

mbayopanda:
Some of our clients have negative quantities in their stock, so the question was how to fix that ?

This PR is about to display also negative quantities in the adjustment module : 

image

All negative value can be set to 0, and once we submit all these negative values will be set to 0 to their final quantity.

N.B. More simulation and tests are needed for financial impact of this adjustment

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -237,6 +237,7 @@
     "ASSIGNMENT_CURRENT":"Current",
     "NO_ASSIGNMENT":"No Assignment",
     "INCLUDE_EXHAUSTED_LOTS":"Include exhausted lots",
+    "INCLUDE_EMPTY_OR_NEGATIVE_LOTS" : "Include exhausted or negative lots",
     "INCLUDE_ONLY_EXHAUSTED_LOTS": "Include only exhausted lots",
     "CLICK_TO_SEE":"Click here to see lots",
     "SHOW_WARNINGS":"Click to show Warnings",
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -315,7 +315,7 @@ async function createInventoryAdjustment(req, res, next) {
         lot_uuid : db.bid(lot.uuid),
         depot_uuid : db.bid(movement.depot_uuid),
         document_uuid : db.bid(positiveAdjustmentUuid),
-        quantity : lot.oldQuantity,
+        quantity : lot.oldQuantity * -1,
         unit_cost : lot.unit_cost,
         date : new Date(movement.date),
         entity_uuid : movement.entity_uuid,
--- a/client/src/i18n/fr/stock.json
+++ b/client/src/i18n/fr/stock.json
@@ -67,6 +67,7 @@
     "GROUPING"        : "Groupage",
     "JOIN_INVOICE"    : "Joindre une facture",
     "INCLUDE_EMPTY_LOTS" : "Inclure les lots épuisés",
+    "INCLUDE_EMPTY_OR_NEGATIVE_LOTS" : "Inclure les lots épuisés ou négatives",
     "INCLUDE_EMPTY_LOTS_HELP" : "Afficher les articles dont la quantité est zero",
     "INCREASE"        : "Augmentation",
     "INCREASE_HELP"   : "Ajustement de stock en augmentant la quantité des lots des produits en stock",
--- a/client/src/modules/stock/inventory-adjustment/inventory-adjustment.html
+++ b/client/src/modules/stock/inventory-adjustment/inventory-adjustment.html
@@ -68,6 +68,20 @@
           <bh-filter-toggle on-toggle="StockCtrl.toggleInlineFilter()">
             <span translate>FORM.LABELS.FILTER</span>
           </bh-filter-toggle>
+          <div class="toolbar-item">
+            <div class="checkbox">
+              <label>
+                <input
+                  type="checkbox"
+                  name="includeEmptyLot"
+                  ng-true-value="1"
+                  ng-false-value="0"
+                  ng-model="StockCtrl.includeEmptyLot"
+                  ng-change="StockCtrl.reloadInventories()">
+                <span translate>STOCK.INCLUDE_EMPTY_OR_NEGATIVE_LOTS</span>
+              </label>
+            </div>
+          </div>
         </div>
       </div>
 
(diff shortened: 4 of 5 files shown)


server/controllers/stock/index.js
```diff
@@ -315,7 +315,7 @@ async function createInventoryAdjustment(req, res, next) {
         lot_uuid : db.bid(lot.uuid),
         depot_uuid : db.bid(movement.depot_uuid),
         document_uuid : db.bid(positiveAdjustmentUuid),
-        quantity : lot.oldQuantity,
+        quantity : lot.oldQuantity * -1,
```
jniles, line 318: Was this a bug?

jmcameron:
Before doing a review on this, I'd like to know why this is necessary.   Specifically:

- What does a negative quantity mean?
- How does a negative quantity occur in the BHIMA software?

It seems to me that this could introduce more problems than it solves.  It may be better to address why these negatives occur and fix it so they cannot happen!

jniles:
@jmcameron we were surprised by this as well.  It's actually quite easy to get negative values and nearly unavoidable.  Here is how:

  1. Start with a balance of 100 pills at the beginning of this month.
  2. Distribution 80 pills today.
  3. Back date a distribution of 50 pills to be made yesterday.
  4. We are now -30 in the hole today, even through you _did_ have 80 pills today, and you _did_ have 50 pills yesterday.

To avoid this, we would need to some check that says "would making this distribution put the database in an invalid/undefined state"?  And then we would further need to figure out how to communicate that to the user such that they can avoid the issue.

Does that make sense?  Maybe you can help us think of a way to work around this issue.... currently, it's escaping me.

client/src/i18n/en/stock.json
```diff
@@ -237,6 +237,7 @@
     "ASSIGNMENT_CURRENT":"Current",
     "NO_ASSIGNMENT":"No Assignment",
     "INCLUDE_EXHAUSTED_LOTS":"Include exhausted lots",
+    "INCLUDE_EMPTY_OR_NEGATIVE_LOTS" : "Include exhausted or negative lots",
```
jniles, line 240: This shouldn't be under the `LOTS` key.  The English version isn't showing up.

jniles (commented):
Looks fine to me.  Just one issue with the english translation key not being in the right place.

jmcameron:
@jniles and @bruce : I suggest holding off on this PR (if possible).  I'd like to take some time to look at it next week to see if I can figure out a better fix.  I really think that the idea of allowing negative values for quantities of stock will be problematic.

jniles:
No problemo. We will be at IMCK next week and will be able to do any changes we need directly on their database or with other tools.  A better solution would be welcome!

jmcameron:
I have two ideas about how to deal with this problem:

- If an update will produce negative values, do not accept it; print out a very specific and detailed error message, something to the effect of:   This update will produce a illegal negative quantity for stock item XYZ.   Update the quantity in the form before submitting the form.
- Automatically update negative values to zero when the form is submitted.   Perhaps print an info/warning message that it got zeroed out.

It also seems like we need a utility to scan the database to see if there are any negative values and deal with them.

If it is not obvious, I still do like the idea of allowing negative quantities!!

jmcameron:
>     * If an update will produce negative values, do not accept it; print out a very specific and detailed error message, something to the effect of:   This update will produce a illegal negative quantity for stock item XYZ.   Update the quantity in the form before submitting the form.

We could also include a suggestion that it may be necessary to go back and correct previous updates.   It seems obvious that there is an error in a previous update or this update.

jniles:
@jmcameron thanks for sharing some ideas.  They are useful to discuss.

> If an update will produce negative values, do not accept it; print out a very specific and detailed error message, something to the effect of: This update will produce a illegal negative quantity for stock item XYZ. Update the quantity in the form before submitting the form.

So, I worry about the calculation that backs this "if .. will produce negative values".  This seems like a whole stock_movements table scan, recomputing balances all along.  Unless we have a more efficient way to store balances, this could get really computationally intensive.  I suppose we could cache the balances in a table that contains (lot_uuid, depot_uuid, current_balance) and decrement that on each exit, and if current_balance goes negative we know something is wrong?

jmcameron:
Here are couple more thoughts on this:  
- In the stock exit form, prevent selecting a date before the current date.  I realize this may not be practical, given delays in updating data, etc.   But if we did this, fixes to historical quantities could be done via the Stock Adjustment page.  
- The Stock Adjustment page should accept the total at a certain date.  I'm not sure if the Increase/Decrease options really make sense in typical operations (it seems like they have use the total to compute the increase/decrease anyway).   If the Stock Adjustment form was used to correct historical data, it would be for  one stock item at a time and it would not really matter if the computations are involved.  Should anyone expect Stock Adjustment (an unusual operation to my mind) to be fast?   Stock Exits on the other hand should be speedy since they will be done a lot.
- Back to the concern about computations involved to see if a changes produces any negative values:  We are only doing one stock item at a time in the stock exit.   And **the computations only need to be done if the date is back-dated** -- and I think we can just put up a "Checking totals..."  notification while the computations are being done..  If it is the current date, we can just make sure that we do not exit more than the existing quantity available (which I assume we already do).

jmcameron:
Oops:  **the computations only need to be done if the data is back-dated** should be **the computations only need to be done if the DATE is back-dated**

jniles:
The problem with using Stock Adjustments is that it breaks the calculation for CMM since adjustments are not considered consumption.  It has been kind of nice to have all CMM calculations flow from a single source (stock exists), so we can optimize that to death.

For (3), you are right - we only need to do the computation for back dating totals, and I think that is entirely reasonable.  It seems like the best proposal yet.  That way, we don't delay the distribution of stock to "regular" patients, in real time.

I think our CMM calculation also needs to be reviewed to make sure this logic is put in place.  I believe we make an attempt, but am not sure the success of it.

mbayopanda:
bors try

jniles:
So... I actually think we will always need this tool, even if we try our best to stop negative values.  For pragmatism (we need this at IMCK at the moment), I'll merge this and we can add additional restrictions later if necessary.

jmcameron:
> So... I actually think we will always need this tool, even if we try our best to stop negative values. For pragmatism (we need this at IMCK at the moment), I'll merge this and we can add additional restrictions later if necessary.

That makes sense.   I'll move some of these comments into a separate issue so we can deal with them later.

jniles:
Thanks @jmcameron!  I really appreciate all your feedback on this.

PR #5154, opened 2020-11-26
fix(credit note): do not reload entire registry

### LINKED ISSUE #5152: Submitting a credit note reloads the entire invoice registry
I ran into this bug with the IMCK database in a live demo.  If you submit the credit note, instead of respecting the filters, it reloads the entire invoice registry. 

Expected behavior: reload the registry with the currently applied filters. 
Actual behavior: reload all rows from the beginning of time.

jniles:
Fixes the credit note submit action so that it doesn't reload the entire registry on submit.

Closes #5152.


lomamech (commented):
I just did the study of the functionality, it works very well

PR #5155, opened 2020-11-26
chore(reports): clean up wkhtmltopdf references

jniles:
This commit cleans up some wkhtmltopdf options that are no longer relevant since we are using coral.  Basically, I removed `footerRight` and `footerFontSize` wherever I found it.  I also ran ESLint on a lot of these files.


server/controllers/finance/reports/configurable_analysis_report/index.js
```diff
@@ -5,7 +5,7 @@ const ReportManager = require('../../../../lib/ReportManager');
 const NotFound = require('../../../../lib/errors/NotFound');
 
 const TEMPLATE = './server/controllers/finance/reports/configurable_analysis_report/report.handlebars';
-const cashboxes = require('../../../finance/cashboxes');
```
mbayopanda, line 8: 👍

mbayopanda (approved):
LGTM 👍 

bors try

PR #5157, opened 2020-11-26
fix(reports): fix client support error coloration

### LINKED ISSUE #5146: Form bug in Clients Support Report page
If you submit the form without selecting the options, you get a warning (in red at the top of the page) saying that the problems are marked in red - But there is nothing highlighted in read.  Note that there are no default values for the options on the form.
If you select values for all the options, the form works fine.

jniles:
Adds the correct error coloration to the `bhYesNo` values when the report is submitted.

Closes #5146.


client/src/modules/templates/bhYesNoRadios.tmpl.html
```diff
@@ -1,7 +1,8 @@
-<div class="form-group" data-bh-yes-no-radios>
-  <label class="control-label" translate>
-    {{$ctrl.label}}
-  </label>
+<div class="form-group" ng-form="bhYesNoForm"
+  data-bh-yes-no-radios
+  ng-class="{ 'has-error' : bhYesNoForm.$submitted && bhYesNoForm.$invalid }">
+
+  <label class="control-label" translate>{{$ctrl.label}}</label>
```
jmcameron, line 5: Excellent!  It will apply to all Yes/No radios in all forms, right?

jmcameron (approved):
Looks great.  I tested it on my system and it worked perfectly.

client/src/modules/templates/bhYesNoRadios.tmpl.html
jniles: That is correct!

PR #5160, opened 2020-11-30
fix(feeCenters): fix name collision with report

jniles:
This commit fixes a fee center name collision bug that is breaking the fee center module.

Steps to reproduce:
  -> Click on the fee center module.  Observe everything is broken.

This also prevents the end to end tests from passing locally.


jniles:
Since this is a breaking issue, I'll merge it myself.

bors r+

PR #5167, opened 2020-12-01
fix(stock): order stock sheet properly

### LINKED ISSUE #5149: Stock Sheet : Order lot movement by creation date is broken
This issue is about this option : 

image

Which must order lots movement by creation date do not work as expected.

The reason of this option was to fix in the stock sheet report, exit which occurs before entry

image

jniles:
Restores the ability for the user to order the stock sheet according to
the creation date. It also adds a couple of columns to the sheet: flux
and lot to provide even better detailed resolution to the query.

Closes #5149.

Here is what it looks like in practice:
image

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -762,13 +762,14 @@ function getInventoryMovements(params) {
       BUID(l.inventory_uuid) AS inventory_uuid, BUID(l.origin_uuid) AS origin_uuid,
       l.entry_date, i.code, i.text, BUID(m.depot_uuid) AS depot_uuid,
       i.avg_consumption, i.purchase_interval, i.delay, iu.text AS unit_type,
-      dm.text AS documentReference
+      dm.text AS documentReference, flux.label as flux
     FROM stock_movement m
-    JOIN lot l ON l.uuid = m.lot_uuid
-    JOIN inventory i ON i.uuid = l.inventory_uuid
-    JOIN inventory_unit iu ON iu.id = i.unit_id
-    JOIN depot d ON d.uuid = m.depot_uuid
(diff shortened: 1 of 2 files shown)


server/controllers/stock/reports/stock/stock_sheet.js
```diff
@@ -32,7 +32,9 @@ async function stockSheetReport(req, res, next) {
 
     const data = { inventory, depot };
 
-    data.rows = rows.movements.sort((x, y) => x.date - y.date);
+    // data.rows = rows.movements.sort((x, y) => x.date - y.date);
```
jmcameron: I don't think you meant to comment out this line.  If I uncomment it, the sort works great.  If left commented out no sorting happens in the report regardless of which option I choose for the sorting.

jmcameron: I was able to verify that the sorting flag works (if uncommented).

jmcameron (approved):
Fix the commented out line and it looks great.

There is one item that I found a little confusing:  The last flag says:  "Sort lots movements by their true creation date".   But the report just lists "Date" for one of the columns.  And when I look at the SQL `inventory` table definition, I see "entry_date".   Are these all the same thing? I suggest adjusting the labels to make this a little clearer for users.

jniles:
@jmcameron thanks for pointing that out.  I thought that the sorting worked without that comment, but I see now that I actually haven't solved the issue.  I'll take a closer look at this before merging.

jniles:
Looks like @mbayopanda beat me to it!

jniles:
@mbayopanda I have updated this PR to work with yours.  Can you give me a review?

mbayopanda (approved):
LGTM 👍 

image

bors r+

PR #5169, opened 2020-12-01
fix: update migration scripts

jniles:
This commit fixes a few typos and the missing path updates in the unit table.


server/models/migrations/v1.16.0-v1.17.0/migrate.sql
```diff
@@ -29,7 +29,7 @@ UPDATE `report` set report_key='employee_standing' WHERE report_key='employeeSta
 UPDATE `report` set report_key='debtor_summary' WHERE report_key='debtorSummary';
 UPDATE `report` set report_key='client_debts' WHERE report_key='clientDebts';
 UPDATE `report` set report_key='client_support' WHERE report_key='clientSupport';
-UPDATE `report` set report_key='receovery_capacity' WHERE report_key='recoveryCapacity';
+UPDATE `report` set report_key='recovery_capacity' WHERE report_key='recoveryCapacity';
```
jmcameron, line 32: Good catch.

jmcameron (approved):
Good catch.   I should have tested this migration file more thoroughly with one of the production databases--I missed that I needed to fix the unit paths too.

PR #5172, opened 2020-12-01
bug(Profit and loss Report)

### LINKED ISSUE #5171: Bug Balance Sheet Report
there was a problem with the profit and lost (Compte de resultat) report, the introduction of period 13 did not take into account that the AccountReference.computeAllAccountReference malfunction when for a closed fiscal year the period referred to was the 13th and not the 12, but also due to the fact that period 13 does not have the end date, this disrupted the display of the Report

lomamech:
- Fix report for closed fiscal year
- Improve report display

closes #5171

### DIFF
--- a/server/controllers/finance/reports/ohada_profit_loss/index.js
+++ b/server/controllers/finance/reports/ohada_profit_loss/index.js
@@ -193,6 +193,7 @@ function reporting(options, session) {
       _.merge(context, { fiscalYear });
 
       const currentPeriodReferences = AccountReference.computeAllAccountReference(fiscalYear.current.period_id);
+
       const firstChoice = AccountReference.computeAllAccountReference(fiscalYear.previous.period_id);
       const previousPeriodReferences = fiscalYear.previous.period_id ? firstChoice : [];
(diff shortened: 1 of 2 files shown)


server/controllers/finance/reports/ohada_profit_loss/index.js
```diff
@@ -356,11 +356,14 @@ function getFiscalYearDetails(fiscalYearId) {
   // get fiscal year details and the last period id of the fiscal year
   const query = `
     SELECT
-      p.id AS period_id, p.end_date,
+      p.id AS period_id, fy.end_date,
       fy.id, fy.label, fy.previous_fiscal_year_id
     FROM fiscal_year fy
     JOIN period p ON p.fiscal_year_id = fy.id
-      AND p.number = (SELECT MAX(period.number) FROM period WHERE period.fiscal_year_id = ?)
+      AND p.number = (
```
jniles, line 363: This is the right solution :+1:

jniles (approved):
LGTM.  Great analysis and catch!

bors r+

PR #5176, opened 2020-12-02
fix: sort the stock inventories by CMM vars

jniles:
This commit allows the user to sort by CMM, MS, quantity, and other variables.

To test it, go to the "Articles in Stock" registry and click on the quantity column.  Note it will sort in the correct order.  Also click on the other CMM-derived columns and they should also sort.


client/src/modules/stock/inventories/registry.js
```diff
@@ -46,6 +46,7 @@ function StockInventoriesController(
     displayName      : 'STOCK.QUANTITY',
     headerCellFilter : 'translate',
     cellClass        : 'text-right',
+    type : 'number',
```
jmcameron, line 49: Need to add here: 

   enableSorting    : true,

Without it, specifying "type: 'number'" sorts alphabetically/lexically, not numerically (it seems to ignore the type spec).  When you add this, the sorting by quantity numerically works correctly.   This probably also applies to the other items that should be sorted numerically.

jmcameron (approved):
Other than the fix I suggested, this works fine.

However, I did spot a minor usability thing:  It is not obvious that the columns are sort-able.  When you click on a column header, it does show the up/down caret, which is fine.  But it is not obvious when columns are sort-able.  This is a common feature of the the grid, so I am not suggesting a fix is needed for this PR, but something we should consider for future UI improvements.  This could be to include a combo up-down caret (fa-sort) in light grey at the right end of the column.  This would be replaced by a up or down caret when the column is sorted.

client/src/modules/stock/inventories/registry.js
jniles: Are you sure?  I seem to be getting the correct sorting behavior:

TBPKGKGfoE

Tested on both Chrome and FF.

jmcameron: Tested this again and it works without my fix.  Not sure what happened yesterday!

jniles: Probably just a mix up of branches 😆

PR #5178, opened 2020-12-02
fix(reports): use serviceUuid instead of serviceId

### LINKED ISSUE #5153: The unpaid invoices report does not filter by service
I seem to recall we've had this bug multiple times now.  The unpaid invoices report does not filter by service, but it should.

Steps to reproduce:
  1. Go to the report.
  2. Set a valid date range
  3. Pick a debtor group that has data.
  4. Filter by service.

Note that the service filter has no effect.

jniles:
Fixes a bug in the unpaid_invoice_payments report where serviceId was not updated to be serviceUuid.

Closes #5153

### DIFF
--- a/server/controllers/finance/reports/unpaid_invoice_payments/index.js
+++ b/server/controllers/finance/reports/unpaid_invoice_payments/index.js
@@ -44,7 +44,6 @@ async function build(req, res, next) {
   res.set(compiled.headers).send(compiled.report);
 }
 
-
 async function reporting(options, session) {
   const qs = _.extend(options, DEFAULT_OPTIONS);
   let results;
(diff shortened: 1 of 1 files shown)


jmcameron (approved):
This fix works great.

One minor suggestion:  It would be nice to add the "Service" as a column in the table in the report.  Since you can filter on that, it would be great to see it in the table.

jniles:
I'm not sure it makes sense to add a column called "service" since each column is a different service, unless I'm misunderstanding your request.

jmcameron:
> I'm not sure it makes sense to add a column called "service" since each column is a different service, unless I'm misunderstanding your request.

Right you are.   I did not understand that when I tried this with bhima_test.  But when I tried it with IMCK, the multitude of columns made this very clear.   Disregard this suggestion!

PR #5179, opened 2020-12-02
fix: remove .env file from release archive

jniles:
This may leak secrets or collide with an on-site .env file.  Strip it from the release archive.


jmcameron (commented):
This seems like a common-sense update.  However, since I have not been involved in the release process or installing releases on production sites, I cannot say that this update will work correctly.   If @jniles  is convinced it will work, I think it should be released.  We will find out if there is a problem with it when we do our next updates on production servers.

jniles:
Thanks for your review @jmcameron!

bors r+

PR #5180, opened 2020-12-03
Fix utility tfcomp

### LINKED ISSUE #5140: Fix bug in the translation comparison utility
Most of the time, the `utilities/translation/tfcomp.js` utility works fine.  However it does have a bug in one corner case:

There is a bug when you have a token which is single item in one language but is has a subtree under it in the other language.  In that case `tfcomp.js` does not detect a difference and the items in the sub-tree that are missing in one language are not reported.

jmcameron:
Fixed bug that was preventing the translation comparison utility (tfcomp.js) from working correctly when:

EN:    . . . KEY1: "Val1"
FR:     . . . KEY1: {
                 CHILDKEY: "Val2"
               }

Previously, the difference here wound not have been noticed.  Now it will correctly indicate that the EN is missing KEY1.CHILDKEY.

Test locally by temporarily modifying the En and Fr translation files to duplicate the problem above and run the test utility.  See comment below for example tests cases.

This fix also fixes the odd cases where we got:   KEY.0, KEY.1, etc missing items.   Those "missing" keys were due to a closely related issue in the construction of lists of missing items.

Also, add a usage warning if the paths are missing.

NOTE: The order of the paths is important.  Note the USAGE statement and updated comment at the top of the file.

### DIFF
--- a/utilities/translation/tfcomp.js
+++ b/utilities/translation/tfcomp.js
@@ -39,14 +39,14 @@ jsonFiles.forEach(function (jsonFile) {
   enMissList = [];
   frMissList = [];
 
-  if(jsonFile.en && jsonFile.fr){
+  if (jsonFile.en && jsonFile.fr) {
 
     // load JSON files
     let enTranslateObject = require(jsonFile.en);
     let frTranslateObject = require(jsonFile.fr);
 
     checkSubDict(enTranslateObject, frTranslateObject, '');
-  }else{
+  } else {
     //add to the missed files list
     !jsonFile.en ? enFileMissList.push(jsonFile.fr) : frFileMissList.push(jsonFile.en);
   }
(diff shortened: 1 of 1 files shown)


jmcameron:
Example test cases.  Insert into corresponding En and Fr files (with the same filename):
[en.txt](https://github.com/IMA-WorldHealth/bhima/files/5633217/en.txt)
[fr.txt](https://github.com/IMA-WorldHealth/bhima/files/5633218/fr.txt)

utilities/translation/tfcomp.js
```diff
@@ -1,26 +1,32 @@
 // Detect missing translation items between translation files
-// USAGE:  node tfcomp.js path1 path2
-// Where path1 and path2 : path to a folder containing json files of translation
+// USAGE:  node tfcomp.js ath1Eng path2Fr
```
jniles, line 2: Looks like this is missing a p

utilities/translation/tfcomp.js
```diff
@@ -1,26 +1,32 @@
...
 'use strict';
 
 const path = require('path');
 const fs = require('fs');
+const process = require('process');
+const { exit } = require('process');
+
+// Make sure we have two paths
+if (process.argv.length < 4) {
+  /* eslint-disable no-console */
```
jniles, line 14: Nice :+1:

jniles (approved):
Confirmed this works:

image


Great work!

bors r+

PR #5190, opened 2020-12-03
fix(lots): edit modal min size

### LINKED ISSUE #5189: Allow smaller prices than normal currency in Edit Lot Modal
We've used bhCurrencyInput in the edit lot modal, which results in this issue when trying to update an expiration date:

image

We should change the minimum allowed to be reflective of the real minimum allowed.

jniles:
This commit changes the minimum price allowed on the lot edit modal to be the minimum allowed in the database.

Closes #5189

### DIFF
--- a/client/src/modules/stock/lots/modals/edit.modal.html
+++ b/client/src/modules/stock/lots/modals/edit.modal.html
@@ -35,6 +35,7 @@
       <bh-currency-input
         label="STOCK.UNIT_COST"
         data-lot-currency-input
+        min="0.00009"
         currency-id="$ctrl.enterprise.currency_id"
         model="$ctrl.model.unit_cost">
       </bh-currency-input>


client/src/modules/stock/lots/modals/edit.modal.html
```diff
@@ -35,6 +35,7 @@
       <bh-currency-input
         label="STOCK.UNIT_COST"
         data-lot-currency-input
+        min="0.00009"
```
jmcameron, line 38: Where did this number come from?   Doesn't it depend on the currency being used (eg CFs?) and the current exchange rate?  I presume this comes from the minimum bill/coin in CF that is accepted converted to dollars?

jmcameron (approved):
I tested this and it works fine.   If this is needed in the field, it should be released.  But we should consider how to generalize this.  This could even be an enterprise setting and use the current exchange rate to compute it, etc.  Also the warning message could print out a more detailed warning message that indicates the problem in terms in the local current currency.

client/src/modules/stock/lots/modals/edit.modal.html
jniles: No, it's more basic than that.  This comes from the fact that we only store up to four decimal places in the `unit_cost` column of the `lot` table.  I wasn't sure if "min" is `>` or `>=` so I went one below it and put `0.00009` instead of `0.0001`.

@lomamech and I discussed this to some extent.  We have a some dangling values that are $0.001 at IMCK for example, but we display them as $0.00 because we only show one two decimal places.  Ultimately, this should probably be user-customizable.  We'd need to update a lot of places though.

jniles:
Note: moved discussion over to https://github.com/IMA-WorldHealth/bhima/issues/5199.

jniles:
In order to get our expiration dates fixed (IMCK was registering them incorrectly), we'll need this in production.  Since we can already set prices to this precision in the lots entry modal, I think this is probably fine.

bors r+

client/src/modules/stock/lots/modals/edit.modal.html
jniles: By the way, with your line of inquiry, you are on the right track. This component normally forces you to pay the minimum currency value, as you are asking about.  I'm telling it to ignore that restriction for the purposes of setting per-pill cost which may be much, much smaller than the minimum amount possible to pay.

PR #5191, opened 2020-12-03
fix(cash): filter correctly on patient debtor_uuid

jniles:
This commit fixes the link from the cash payments registry to the patient registry, so that we filter by debtor_uuid instead of patient name.  This is much stricter, and results in finding the exact patient.

### DIFF
--- a/client/src/i18n/en/transactions.json
+++ b/client/src/i18n/en/transactions.json
@@ -1,7 +1,7 @@
 {
   "TRANSACTIONS":{
     "VERIFIED_TRANSACTIONS": "Verified Transactions",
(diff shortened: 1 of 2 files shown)


jmcameron (approved):
Looks good.  I spot-checked several of them and they seem to work.

A tiny cosmetic suggestion.  Note that second word in "View transactions" is not capitalized like the other action entries (see en/transactions.json:line 4):
image

jniles:
Thanks for the review @jmcameron!  I'll fix that and do a merge.

jniles:
Okay, fixed.

bors r+

PR #5192, opened 2020-12-03
chore(inventory): reorder menu items

jniles:
This commit reorders the menu items to be similar to all other modules - first the filter, then the search, finally the menu.


jmcameron (approved):
Looks good. I tested it and it seems fine.

PR #5196, opened 2020-12-03
fix(stock): update dashboard variable names

### LINKED ISSUE #5193: Allow depots to have 0 as min months security stock
At IMCK, we discovered they have a few "passthrough" depots.  These are constantly in stock out.  Instead of warning of constant stock outs, let's make it clear that these depots are "passthrough" depots by simply saying 0 months of stock is their minimum.  Here is what happens when I try to do that:

image

### LINKED ISSUE #5195: Latest stock performance changes broke the "risk of expiration" calculation on dashboard
Looks like the latest stock updates have broken the dashboard.  Here is what happens on the client:

image

And here is the server log:
```sh
| 2020-12-03T12:45:48.232Z errors #interceptor() Error: Cannot calculate the AMC without consumption parameters!
|     at getBulkInventoryCMM (/home/bhima/apps/bhima/bin/server/controllers/stock/core.js:305:11)
|     at Object.getLotsDepot (/home/bhima/apps/bhima/bin/server/controllers/stock/core.js:263:47)
|     at processTicksAndRejections (internal/process/task_queues.js:94:5)
|     at async Promise.all (index 1)
```

jniles:
Update the dashboard variable names to account for the CMM changes.

Closes #5195.

### DIFF
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -599,8 +599,8 @@ function listMovements(req, res, next) {
  * returns data for stock dashboard
  */
 function dashboard(req, res, next) {
-  const monthAverageConsumption = req.session.stock_settings.month_average_consumption;
-  const averageConsumptionAlgo = req.session.stock_settings.average_consumption_algo;
+  // eslint-disable-next-line
+  const { month_average_consumption, average_consumption_algo } = req.session.stock_settings;
 
   const dbPromises = [];
   let depotsByUser = [];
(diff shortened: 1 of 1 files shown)


jmcameron (approved):
Looks good. I verified that this works in the dashboard. 

I was frustrated at first because my test user did not have permissions to see the stock dashboard.  People at the receiving end of new release should be informed of the new pages and the necessity to enable permissions for appropriate users to see the pages.

jniles:
@jmcameron, you are right that is frustrating.  Since we structure permissions as "roles" and "permissions", it's something that we've always just informed the system administrator about.  We'll say "hey, the following report is now available" and it is up to them to see who gets access to it.

I'm not entirely sure what the best way is to ensure that only users who need/are authorized get access to the modules without it taking quite a long time to update any site.

PR #5197, opened 2020-12-03
fix(depots): store depot uuid in URL

### LINKED ISSUE #5194: Refreshing the page on Edit Depot loses the depot reference
The depot registry loses reference to the editing depot when you refresh the page.  This is because we don't use the $state to store the uuid in the URL.  Instead you must come from the depot registry.  We should fix this as editing without a depot defined is an undefined state.

jniles:
Stores the depot uuid in the URL so that we have a reference to it even if we refresh the page.  Also contains some typo fixes.

Closes #5194


mbayopanda (approved):
LGTM 👍 

bors r+

PR #5200, opened 2020-12-04
Fix lot movement ordering in the stock sheet report

### LINKED ISSUE #5149: Stock Sheet : Order lot movement by creation date is broken

mbayopanda:
This PR fixes the use of created_at in the ordering of stock movement in the stock sheet report

image

Now we can force to see the real order of movements

Close #5149


jniles (approved):
Looks good to me.  This is what I was trying to fix!

bors r+

PR #5201, opened 2020-12-04
Bug: Fix Display of data relating to agent holidays on payslips

lomamech:
Fix Display of data relating to agent holidays on payslips


lomamech:
image

jniles (approved):
LGTM.  Great catch!

PR #5202, opened 2020-12-04
Use of cross-env for release

mbayopanda:
This PR adds `cross-env` for `yarn release` multi-platform compatibility


jniles:
bors r+

Good catch!

PR #5203, opened 2020-12-04
fix: remove old bin before archive upload

jniles:
If your bin isn't clean, you upload old artifacts.  This cleans bin before running yarn build in production mode while releasing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "hooks": {
       "before:init": "node utilities/translation/tfcomp.js client/src/i18n/en client/src/i18n/fr",
       "after:bump": [
+        "rm bin",
         "cross-env NODE_ENV=production yarn build",
         "rm bin/.env",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -38,6 +38,7 @@
     "hooks": {
       "before:init": "node utilities/translation/tfcomp.js client/src/i18n/en client/src/i18n/fr",
       "after:bump": [
+        "rm bin",
```
mbayopanda, line 41: 👍 Good catch

package.json
```diff
@@ -38,6 +38,7 @@
...
         "cross-env NODE_ENV=production yarn build",
         "rm bin/.env",
         "GZIP=-9 tar -czvf ${version}.tar.gz bin/"
```
mbayopanda, line 44: On Windows, we have got an error during the release because of `bin/`, and I think we have to adapt this line to the release `v1.17.2`

mbayopanda (approved):
LGTM :+1

PR #5209, opened 2020-12-05
chore(payroll): weekend config w/ bhCheckboxTree

jniles:
Migrates the weekend configuration to use bhCheckboxTree.


jniles:
@lomamech can I get a review?

lomamech (approved):
I inspected the different modules that interact with the configuration of the weekens, everything works correctly

PR #5210, opened 2020-12-06
Fix the AMC requirements error in stock changes report

mbayopanda:
This PR fixes an error related to a missing parameter for AMC calculation in the stock changes report


jniles (approved):
LGTM.

bors r+

PR #5211, opened 2020-12-06
fix(employees): restore locked employee filter

jniles:
This commit fixes the "locked" employee feature on the employee
configuration page for payroll which was lost during the migration to
bh-checkbox-tree.  To do this, I added a new feature to bh-checkbox-tree
to allow disabling nodes based on their id, similar to how the
checked-ids functionality works.

In this case, disabled ids will retain the state they came in with,
being ignore in the "setNodeValue()"  function.  Therefore, whatever the
node came in with, it will leave with.  Additionally, the icon disables
the checkbox, makes the font italics, and puts a little lock icon in
front of the text value.

Thanks to @lomamech for pointing this out.

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -1044,3 +1044,19 @@ a[disabled] {
 .lot-width {
   width: 30%;
 }
+
+/* lock certain items of the checkbox tree when disabled */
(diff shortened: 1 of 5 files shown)


jniles:
@lomamech can I get a review?

lomamech (approved):
image
It works very well

PR #5220, opened 2020-12-09
bug(Define Orientation Report PDF)

lomamech:
- Correction of the operating error on the bhRenderOption component

- Taking into account the choice of the orientation parameter for different
report

### DIFF
--- a/client/src/js/components/bhRenderOption.js
+++ b/client/src/js/components/bhRenderOption.js
@@ -8,7 +8,6 @@ angular.module('bhima.components')
     },
   });
 
-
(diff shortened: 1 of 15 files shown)


lomamech:
@jniles @mbayopanda  can you check this PR

jniles (approved):
Confirmed this works.  Great fix!

It looks like a lot of our reports are not adapted for flipping their orientation.  For example, here are the margins on this report:

image


We should do a second pass (after this is merged) to ensure that all reports scale correctly.

bors r+

PR #5222, opened 2020-12-09
feat(users): link users from role management

### LINKED ISSUE #4598: Link to User Management registry from Role Management
It would be nice if we could link to the users who have a particular role from the Role Management registry.  Like this:

image

Basically, we would add an `a href` that links to the User Management registry and filters the users on those that have this role.

jniles:
Adds a link from the role management interface to the users, filtering the users on the role selected.  I've had this branch for a while, but just got around to completing it.

Closes #4598.

Here is an example in action:

xiUof9eaXi

### DIFF
--- a/client/src/i18n/en/users.json
+++ b/client/src/i18n/en/users.json
@@ -3,19 +3,20 @@
     "ADDING_USER" : "New User",
     "ADD_USER" : "Add a New User",
     "CREATE" : "Create",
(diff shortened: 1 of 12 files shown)


jmcameron (approved):
I tested this and it works perfectly.  Very handy!
I reviewed the code and it seems fine.

PR #5226, opened 2020-12-10
Lifetime calculation in days not months

jniles:
This pull request attempts to clean up our inventory/lot flag calculations so that we have a uniform way of presenting the stock out, stock expiration, at risk of expiration, and at risk of stock out states across the client.

Here is what it looks like:
image

We've standardized around four "flags":
  1. exhausted - the quantity is 0.
  2. expired - the stock lot is expired
  3. near_expiration - the stock will expire before it will be consumed.
  4. at_risk - the stock is at risk of running out soon.

I've updated many of the calculations so they take place on the server will commented documentation. 

### Consumption Schedule
This PR implements what I'll call the "consumption schedule" for lots.  Basically, the lifetimes of each lot are calculated as if the enterprise was optimally using the lots within their depot.  So, the depot will use the lot that will expire soonest, then the next soonest, etc.  This means that a lot may be in stock, but will not be used until much later.

This may produce some misleading figures in the registry, which is why it may be important to show some of these calculations to the client.  I'll propose an issue about that.

### Bug Fixes
This PR also fixes bugs in the coloration of the registry and adds a final aggregation to the bottom of the registry for all four flags.


jniles:
@mbayopanda I think this is finally ready for a review.  Apologies for taking so long.

@jmcameron if you want to take a look at this as well, I would appreciate it.  It's pretty dense though, so don't worry if you feel out of your depth taking a look.

server/controllers/stock/core.js
```diff
@@ -266,23 +266,22 @@ async function getLotsDepot(depotUuid, params, finalClause) {
...
   );
 
-  // FIXME(@jniles) - this step seems to mostly just change the ordering of lots.  Can we combine
-  // it with the getBulkInventoryCMM?
+  // add lot indicators to the inventory list.
   let inventoriesWithLotsProcessed = computeLotIndicators(inventoriesWithManagementData);
 
   if (_status) {
-    inventoriesWithLotsProcessed = inventoriesWithLotsProcessed.filter(row => row.status === _status);
+    inventoriesWithLotsProcessed = inventoriesWithLotsProcessed.filter(lot => lot.status === _status);
```
jmcameron, line 273: I like these variable name changes.  Makes it more readable!

server/controllers/stock/core.js
```diff
@@ -692,16 +692,35 @@ async function getInventoryQuantityAndConsumption(params) {
...
 
-  _.map(inventoryByDepots, (depotInventories) => {
+  const inventoryByDepots = _.groupBy(inventories, 'depot_uuid');
 
+  _.forEach(inventoryByDepots, (depotInventories) => {
     const inventoryLots = _.groupBy(depotInventories, 'inventory_uuid');
 
-    _.map(inventoryLots, (lots) => {
+    _.forEach(inventoryLots, (lots) => {
+
```
jmcameron, line 723: Typo on line 713/732: hight -> higher

server/controllers/stock/core.js
```diff
@@ -714,32 +733,80 @@ function computeLotIndicators(inventories) {
       orderedInventoryLots = _.orderBy(orderedInventoryLots, 'lifetime', 'asc');
 
       // compute the lot coefficient
```
jmcameron: coefficientS

server/controllers/stock/core.js
```diff
@@ -714,32 +733,80 @@ function computeLotIndicators(inventories) {
...
+          // calculate finally the remaining days of stock, assuming we use this stock in order
+          // maybe we will not ever get a chance to use it ... then it is 0.
+          lot.lifetime_lot = Math.max(0, numDaysOfStockLeft - runningLotLifetimes);
+          lot.min_stock_date = moment(new Date()).add(runningLotLifetimes).toDate();
+          lot.max_stock_date = maxStockDate;
+
+          // add to the running LotLifetimes so that the next product will be used after it.
+          runningLotLifetimes += lot.lifetime_lot;
+
+          // the usuable quantity remaining is the minimum of the stock actually available
```
jmcameron: Typo: usable

jmcameron (approved):
A minor suggestion.  We have several all-caps variables (eg, IS_IN_RISK_EXPIRATION, S_RISK, S_LOT_LIFETIME, etc).  In general, we usually reserve all-caps-with-underscores for constants or i18n language tags, not variables/flags.  I suggest making these lower-case (with underscores or camel case, as appropriate).  And maybe choosing more descriptive names as part of the change?

Question: Since the CMM is calculated by consumption over N months, it will not change very fast when stock consumption increases due to some short-term period with high usage.  However, the usage timeline depends on CMM and will not properly reflect how the usage pattern may need to change if the anomalous usage pattern continues.   We may want to consider how to address that.  I can imagine this scenario:   There is a flare up of Cholera locally.  The hospital dispenses meds for that at a higher than usual rate.   The admins start to wonder, when will we run out?  If they use the current code (as nice as it is), they may get a rosy/misleading picture.   It may be useful to provide a way to do these calculations with a more instantaneous estimate of the CMM to get a more accurate picture.  This could be done by providing a way to let the CMM period be temporarily overridden in the query.

server/controllers/stock/core.js
```diff
@@ -692,16 +692,35 @@ async function getInventoryQuantityAndConsumption(params) {
  * process multiple stock lots
  *
  * @description
+ * Computes the indicators on stock lots.  Sets the following flags:
+ *   1) expired - if the expiration date is in the past
+ *   2) exhausted - if the quantity is 0.
+ *   3) near_expiration - if the expiration is before
```
jniles: Whoops this also is an incomplete thought.

jniles:
@jmcameron thanks for you review. 

> A minor suggestion. We have several all-caps variables (eg, IS_IN_RISK_EXPIRATION, S_RISK, S_LOT_LIFETIME, etc). In general, we usually reserve all-caps-with-underscores for constants or i18n language tags, not variables/flags. I suggest making these lower-case (with underscores or camel case, as appropriate). And maybe choosing more descriptive names as part of the change?

I agree with this suggestion.  I'll let @mbayopanda decide if this should be implemented in this PR or not.  I worried that changing too many things at once may make it harder to follow.  Happy to do that here or in a subsequent PR that would be more focused on renaming rather than introducing new logic.

> Question: Since the CMM is calculated by consumption over N months, it will not change very fast when stock consumption increases due to some short-term period with high usage. However, the usage timeline depends on CMM and will not properly reflect how the usage pattern may need to change if the anomalous usage pattern continues. We may want to consider how to address that. I can imagine this scenario: There is a flare up of Cholera locally. The hospital dispenses meds for that at a higher than usual rate. The admins start to wonder, when will we run out? If they use the current code (as nice as it is), they may get a rosy/misleading picture. It may be useful to provide a way to do these calculations with a more instantaneous estimate of the CMM to get a more accurate picture. This could be done by providing a way to let the CMM period be temporarily overridden in the query.

I'm a bit torn about this, so I think we need to ask our pharmacist team.  I think the point of CMM is to smooth out these peaks/troughs when you do your ordering.  The user can always update the "number of months" used in the stock settings, so a smart user could gain this information currently.

It might be interesting to have a graduated CMM report with columns for 24 month CMM, 12 month CMM, 6 month CMM, 3 month CMM.  That way, you could see the trend line over a large period and judge for yourself when ordering.  I don't think the CMM calculation architecture is currently designed for this scenario, but we could always refactor it.

client/src/js/services/LotService.js
```diff
@@ -28,23 +28,24 @@ function LotService(Api, $http, util) {
...
   */
   lots.computeLotWarningFlags = (lot) => {
-    lots.exhausted = false;
-    lots.expired = false;
-    lots.near_expiration = false;
-    lots.at_risk = false;
+    lot.exhausted = false;
+    lot.expired = false;
+    lot.near_expiration = false;
+    lot.at_risk = false;
```
mbayopanda, line 64: 👍🏽

mbayopanda (commented):
@jniles having a lifespan of lots in days is giving our calculation accuracy, most of our flags (expired, risk of expiration, risk of stock out, stock out, etc.) coloration have an obvious understanding now than before.

My concern is about the `lot lifespan by lot`, by reading the registry we cannot understand (in an obvious way) how this column is calculated; there is no reason to have this column in the registry if the user can't understand how it is calculated.
lot lifespan

In the previous implementation, this column had a reason because it was helping us to understand some calculations just by reading the registry.

image

If our clients need that column we must have that in a way where they understand easily where it comes from and the usage just by reading the registry

If our clients cannot understand that, we must not have this column in the registry.

Or, we can add in the registry other columns for CMM in days and MS in days which can help to understand some calculations made per day; with this option, our registry will become bigger (horizontally) and less comfortable for users. With this option also if the user can't understand the columns displayed there is no reason to have these columns

jniles:
@jmcameron I've switched the base branch to the `refactor-stock-management` branch.  I've also included a fix for #5311.  Give this a quick look and see if it all still passes muster.  If so, feel free to merge it.

Here is what the final version looks like:
image

jmcameron (approved):
It looks good to me.

I really like the simplified "Lots in Stock" page.  It makes much more sense to me!

I decided to let you release this yourself since I was not sure if you wanted to rebase first.

PR #5227, opened 2020-12-10
Fix patient merge2

jmcameron:
Remove incorrect deletion of record in entity and entity_group_entity tables.  Per @jniles in PR https://github.com/IMA-WorldHealth/bhima/pull/5223, these are not needed:

> Despite the identical name, I don't think this entity table or the
> entity_group_entity table has anything to do with the patient/entity_map/debtor
> tables. I believe it is only concerned with emails.


jniles (approved):
LGTM. 

bors r+

PR #5231, opened 2020-12-10
Fixed bug in server /patient/duplicates database query function

jmcameron:
Fix a bug in the server query function.  The code was trying to get the parameters from req.params, which does not work (with express Router), but req.query does work.    Also update the code so the limit was not hardcoded in the SQL code but came from the client side via the parameters.

Note that this also fixes an integration test that was succeeding incorrectly because the parameters passed into the test calls to this function were being ignored.

### DIFF
--- a/server/controllers/medical/patients/merge.js
+++ b/server/controllers/medical/patients/merge.js
@@ -34,17 +34,19 @@ router.get('/count_employees', async (req, res, next) => {
 });
 
 router.get('/duplicates', async (req, res, next) => {
-  const sensitivity = req.params.sensitivity || 2;
+  const sensitivity = req.query.sensitivity || 2;
+  const limit = parseInt(req.query.limit, 10) || 25;
   const duplicateSQL = `
-    SELECT COUNT(p.uuid) AS num_patients, p.display_name, GROUP_CONCAT(CONCAT(BUID(p.uuid), ':', em.text)) AS others
+    SELECT COUNT(p.uuid) AS num_patients, p.display_name,
+    GROUP_CONCAT(CONCAT(BUID(p.uuid), ':', em.text)) AS others
     FROM patient p LEFT JOIN entity_map em ON p.uuid = em.uuid
     GROUP BY LOWER(p.display_name) HAVING COUNT(p.uuid) > ?
     ORDER BY COUNT(p.uuid) DESC
-    LIMIT 25;
+    LIMIT ?;
   `;
 
   try {
-    const patients = await db.exec(duplicateSQL, [sensitivity]);
+    const patients = await db.exec(duplicateSQL, [sensitivity, limit]);
     res.status(200).json(patients);
   } catch (e) {
     next(e);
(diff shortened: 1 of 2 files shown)


server/controllers/medical/patients/merge.js
```diff
@@ -34,17 +34,19 @@ router.get('/count_employees', async (req, res, next) => {
 });
 
 router.get('/duplicates', async (req, res, next) => {
-  const sensitivity = req.params.sensitivity || 2;
+  const sensitivity = req.query.sensitivity || 2;
+  const limit = parseInt(req.query.limit, 10) || 25;
```
lomamech, line 38: I am going to make a confession, this is the very first time that I discover this functionality, I have the following concern how do you define the limit of values

lomamech: image

I have the impression that it does not use the filter defined on the registry

jmcameron: This value of 'limit' refers to the number of potential duplicate patients that is shown in the modal (before going back to the patient registry grid).  The user then selects one and then those patients are shown in the main patient registry grid (with the limit filter that can be set in the "Search" modal).  

But the two limits are unrelated.   The number of potential duplicate patients shown in its modal probably should have a separate 'limit' value.   Currently it is hard-coded.  We could, in the future, provide a way to set it.  But I think we can ignore that for now.

I did figure out how to pass in the current patient registry grid 'limit' to the duplicate patients function, but I don't think it makes sense to do.

@jniles What do you think?

jniles: Yes, @jmcameron is right - this "limit" is not related to the patient shown in the registry.  That URL is `/patients`, while this one is `/patients/duplciates`.

The point of the limit was to make sure we don't overwhelm either the user or the server with large queries.  So far, that has not happened, but it could.  Thanks for fixing it @jmcameron .

test/integration/patients.js
```diff
@@ -199,7 +199,7 @@ describe('(/patients) Patients', () => {
           .query({ sensitivity : 0 });
       })
       .then((res) => {
-        helpers.api.listed(res, 0);
+        helpers.api.listed(res, 5);
```
jniles, line 202: Good fix. 👍

jniles (approved):
This looks good to me.  Thanks @jmcameron !

server/controllers/medical/patients/merge.js
lomamech: > This value of 'limit' refers to the number of potential duplicate patients that is shown in the modal (before going back to the patient registry grid). The user then selects one and then those patients are shown in the main patient registry grid (with the limit filter that can be set in the "Search" modal).
> 
> But the two limits are unrelated. The number of potential duplicate patients shown in its modal probably should have a separate 'limit' value. Currently it is hard-coded. We could, in the future, provide a way to set it. But I think we can ignore that for now.
> 
> I did figure out how to pass in the current patient registry grid 'limit' to the duplicate patients function, but I don't think it makes sense to do.
> 
> @jniles What do you think?

No problem

PR #5234, opened 2020-12-11
Fix filter displayValue in stock movement

### LINKED ISSUE #5181: Searching twice in Stock Movements causes the depot_uuid to appear
If you search for a depot once, you get the name of the depot.  If you search twice, the uuid is presented.  This isn't a good user experience.  See below:

image

We need to cache the "displayValue:" of the depot as well as the uuid.  This will cause it to be saved.

mbayopanda:
This PR fixes the `displayValue` in stock movement registry

Close #5181

### DIFF
--- a/client/src/js/services/FilterService.js
+++ b/client/src/js/services/FilterService.js
@@ -74,7 +74,6 @@ function FilterService() {
     this._customFilters = this._customFilters.concat(formattedFilters);
   };
 
-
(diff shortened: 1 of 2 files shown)


jniles (approved):
LGTM.

bors r+

jniles (approved):
LGTM.

bors r+

PR #5240, opened 2020-12-14
fix(reports): properly exchange cash report

### LINKED ISSUE #5239: Cash Report does not use enterprise currency for balances
The Cash report has a bug where it uses USD instead of the enterprise currency for the balances at the bottom of the report.  See below.  This account and the enterprise is in FC.

image

jniles:
Uses the correct currency_id in the report with separated income/expenses.

Closes #5239.

Demonstration:
MLnm1erhnn

### DIFF
--- a/server/controllers/finance/accounts/transactions.js
+++ b/server/controllers/finance/accounts/transactions.js
@@ -110,7 +110,6 @@ function getTableSubquery(options, table) {
   return { query, parameters };
 }
 
-
 /**
  * @function getSubquery
  *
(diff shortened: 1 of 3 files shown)


lomamech (commented):
I was looking at the FC cash register report, and I noticed that in most lines we can see the same motant in FC displayed in dollars
image

jniles:
@lomamech, thanks for the review.  You are right - this report is entirely broken.  I'll take a longer look.

jniles:
@lomamech this is finally ready for review.

jniles:
Hi @lomamech could you take a look at this?

lomamech:
image
it works really well and it really takes into account the current exchange rate at the time of the transaction

PR #5241, opened 2020-12-15
Improve add employee reference

lomamech:
- Adding employee reference to payslip
- Addition of the employee's name, on the description of the transaction
  in posting journal

closes #465

### DIFF
--- a/server/models/procedures/voucher.sql
+++ b/server/models/procedures/voucher.sql
@@ -80,10 +80,12 @@ BEGIN
     reference_uuid, comment, transaction_type_id, user_id)
   SELECT
     HUID(UUID()), v.project_id, fiscal_year_id, period_id, transaction_id, transIdNumberPart, v.date,
-    v.uuid, v.description, vi.account_id, vi.debit, vi.credit,
-    vi.debit * (1 / current_exchange_rate), vi.credit * (1 / current_exchange_rate), v.currency_id,
+    v.uuid, IF(p.display_name IS NOT NULL, CONCAT(v.description,' / ', p.display_name), v.description) description,
+    vi.account_id, vi.debit, vi.credit, vi.debit * (1 / current_exchange_rate), vi.credit * (1 / current_exchange_rate), v.currency_id,
     vi.entity_uuid, vi.document_uuid, NULL, v.type_id, v.user_id
   FROM voucher AS v JOIN voucher_item AS vi ON v.uuid = vi.voucher_uuid
+  LEFT JOIN employee AS emp ON emp.creditor_uuid = vi.entity_uuid
+  LEFT JOIN patient AS p ON p.uuid = emp.patient_uuid
   WHERE v.uuid = uuid;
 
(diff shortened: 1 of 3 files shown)


lomamech:
image

image

lomamech:
To solve the problem, that of displaying the name of the employee on the description in the entry log, I opted for the modification of the procedure

lomamech:
@jniles  can you review this PR, let me know if the way I wrote the modification of the PostVoucher procedure on the migrate file is correct

server/models/procedures/voucher.sql
```diff
@@ -80,10 +80,12 @@ BEGIN
     reference_uuid, comment, transaction_type_id, user_id)
   SELECT
     HUID(UUID()), v.project_id, fiscal_year_id, period_id, transaction_id, transIdNumberPart, v.date,
-    v.uuid, v.description, vi.account_id, vi.debit, vi.credit,
-    vi.debit * (1 / current_exchange_rate), vi.credit * (1 / current_exchange_rate), v.currency_id,
+    v.uuid, IF(p.display_name IS NOT NULL, CONCAT(v.description,' / ', p.display_name), v.description) description,
+    vi.account_id, vi.debit, vi.credit, vi.debit * (1 / current_exchange_rate), vi.credit * (1 / current_exchange_rate), v.currency_id,
     vi.entity_uuid, vi.document_uuid, NULL, v.type_id, v.user_id
   FROM voucher AS v JOIN voucher_item AS vi ON v.uuid = vi.voucher_uuid
+  LEFT JOIN employee AS emp ON emp.creditor_uuid = vi.entity_uuid
```
jniles, line 87: I don't think this is the right approach - for the majority of transactions, we won't need this code and it will just be extra work the computer is doing for nothing.

A better approach would be to make the voucher_item description be set to the employee name.  It looks like we set the voucher_item here:
https://github.com/IMA-WorldHealth/bhima/blob/b0ea5c829603044ecdf402675d705a3ac5c830ad/server/controllers/payroll/multiplePayroll/commitment.js#L203-L205

https://github.com/IMA-WorldHealth/bhima/blob/b0ea5c829603044ecdf402675d705a3ac5c830ad/server/controllers/payroll/multiplePayroll/commitment.js#L215-L217

https://github.com/IMA-WorldHealth/bhima/blob/b0ea5c829603044ecdf402675d705a3ac5c830ad/server/controllers/payroll/multiplePayroll/commitment.js#L229-L231

Which means we need to add the employee name to this array.  The array comes from "rubricsConfigs` which is passed in.

I believe this data comes from this SQL query in `makeCommitment.js`:

https://github.com/IMA-WorldHealth/bhima/blob/b0ea5c829603044ecdf402675d705a3ac5c830ad/server/controllers/payroll/multiplePayroll/makeCommitment.js#L56-L67

So, if you can add the lookup of the employee name to this SQL query, then put in the `voucher_item` a description with the name, I think that should do it.

jniles (changes requested):

PR #5244, opened 2020-12-17
Use of enterprise currency in cash report

### LINKED ISSUE #5239: Cash Report does not use enterprise currency for balances

mbayopanda:
Close #5239


jniles:
This is a duplicate of https://github.com/IMA-WorldHealth/bhima/pull/5240

PR #5248, opened 2020-12-22
Improve patient finding functionality

### LINKED ISSUE #5233: Enhance looking up returning patients
In the Hospital > Patient Registration page, there is a [Lookup Returning Patient] button (top right).  It currently allows looking up a returning patient by Reference or Hospital No.   We should support doing "smart" lookups by name, sex, and birthdate.  This could be done on a separate tab in the modal.

We should also consider looking up the Patient Reference or Hosptial No (fiche num) via a type-in lookup.

jmcameron:
Improve the functionality of finding patients based on name, and optionally gender and data of birth.   This is a general server functionality for patients.   It is implemented initially in the "Lookup Returning Patient" modal in the patient registration page.

**TESTING:**
- Use the IMCK database
- Note that this PR adds a yarn dependency (jaro-winkler).  Be sure to do 'yarn' when testing the PR!
- On the Hospital > Patient Registration page, click on the [Lookup Returning Patient] on the top right
- Click on the "By Name" tab and try some searches.  Once a non-unique search returns, a list of potential names will be given.  If any are the match, click on the name to get the Patient info (for admitting, for instance).
- Here are few search suggestions:
  - BUKUMBA BADIBANGA   (should produce 3 matches, one exact)
  - TUBANGE MUTEBA ARLETTE  (should produce 2 exact matches, eg dupes)
  - TUAKULE TSHIPAMBA (should produce 3 matches with different DOBs)

Note that there are tests in `test/integration/patientMatches.js`


server/controllers/medical/patients/index.js
```diff
@@ -423,6 +427,233 @@ function searchByName(req, res, next) {
...
+          data.forEach((row) => {
+            const [/* name */, mscore] = matches.find(mrow => { return mrow[0] === row.uuid; });
+            row.matchScore = mscore;
+          });
+          return res.status(200).json(data);
+        })
+        .catch(next)
+        .done();
+
+      return res.status(200).json([]); // Keep Hound from complaining
```
jniles: This is where your error is occurring.  You have already called res.status().send() above, in the .then() promise chain.  Since each HTTP request can only return one result, express is complaining as this .json(), will trigger the return, followed by the .send() from a few lines above it.

jniles: Here is my proposal to clean this up and avoid sending two http responses: let's make this entire function an async function!  It will be easier to read as well.

Basically, use async in front of the function call, then use await in front of the db.exec(), and await in front of the find() call.  Then you can return the res.status().json() as the last call in the function without thinking about the .then() promise chain since it will be unraveled.

It should also make the code a bit easier to read.

test/integration/patientMatches.js
```diff
@@ -0,0 +1,258 @@
...
+  it('Save starting number of patients', () => {
+    return agent.get('/patients')
+      .then((res) => {
+        initialNumPatients = res.body.length;
+      })
+      .catch(helpers.handler);
+  });
+
+  // Add all the mock patients for this test
+  mockPatients.forEach(p => {
```
jniles: I do not believe that async works properly in Array forEach statements.  Here is what I would do:

```js
const promises = await Promise.all(mockPatients.map(p => {
  return Promise.all([db.exec(), db.exec()]);
}));

// Now do your asserts
const isAllTrue = promises.every(val => !!val === true);
expect(isAllTrue).to.equal(true);
```

I wrote this on a smartphone, apologies for syntax errors :)

jniles (commented):
I will review this in further detail again, but it looks really good!  Can you add into your comments above the searching/matching functions a bit of explanation about the theory behind it?  Like what the overall function is that is being applied for the match calculation.  I might have missed it, but it will be helpful in debugging if we discover any issues in this code down the road.

server/controllers/medical/patients/index.js
jmcameron: Until we can chat tomorrow, I did a hack on this to return [] to keep hound happy.  This line will never execute, so it is isn't a problem.  But I would like to fix it as you suggested.   Have not been able to get to it today.

test/integration/patientMatches.js
jmcameron: I've tried multiple ways to fix this.   I cannot force the setup DB queries to reliably execute before running the tests that depend on them.  Tried about 10 different things without luck -- including a global before() function.  Still either run into SQL queries not completing early enough or DB LOCK problems (running multiple queries at the same time).  Look forward to chatting with you about this.

jmcameron: With help from @jniles, I implemented this fix.

server/controllers/medical/patients/index.js
jmcameron: With help from @jniles I implemented this fix.

server/controllers/medical/patients/index.js
```diff
@@ -423,6 +427,289 @@ function searchByName(req, res, next) {
...
+ * @method findBestNameMatches
+ *
+ * @description
+ * This method implements a patient search based on name and optionally
+ * gender and date of birth.  Note that parts of names that are a single
+ * letter are ignored. It tries to do 'fuzzy' search and can find names
+ * that have their name parts in different order and is tolerant of
+ * minor misspellings.
+ *
+ * This function supports these query parameters:
```
jniles, line 507: Nice documentation :+1:

server/controllers/medical/patients/index.js
```diff
@@ -423,6 +427,289 @@ function searchByName(req, res, next) {
...
+      // If there are no matches, return now
+      if (matches.length === 0) {
+        return res.status(200).json([]);
+      }
+
+      // Resort the matches
+      matches = matches.sort((a, b) => { return b[1] - a[1]; });
+
+      // console.log("NameMatches: ");
+      // nameMatches.forEach(([pid, sname, score]) => {
```
jniles: You mentioned this in your PR body, but if you are wanting to see debug logs, a better method is to use the `debug()` module.

test/integration/patientMatches.js
```diff
@@ -0,0 +1,244 @@
...
+        expect(matches[1].display_name).to.be.not.equals(testName);
+        expect(matches[1].matchScore).to.be.lt(matches[0].matchScore);
+      })
+      .catch(helpers.handler);
+  });
+
+  // -------------------------------------------------------------------------------------
+
+  // Delete the mock patients
+  after('clean up temporary patients', () => {
```
jniles, line 276: Excellent :+1:

server/controllers/medical/patients/index.js
```diff
@@ -423,6 +427,289 @@ function searchByName(req, res, next) {
...
+            }
+          }
+
+          // Check the dob
+          if ('dob' in options) {
+            let dob = new Date(options.dob);
+            if (options.dob.length === 4) {
+              // Arbitrarily choose the middle of the year to
+              // avoid problems with the time zone offsets causing
+              // the wrong year to be used later.
```
jniles, line 639: Good idea.

client/src/modules/patients/lookupReturningPatient.modal.html
```diff
@@ -1,99 +1,306 @@
...
-      <div class="alert alert-warning" ng-if="ModalCtrl.hasMultiplePatients">
-        <i class="fa fa-warning"></i>
-        <span translate>PATIENT_REG.HAS_MULTIPLE_PATIENTS</span>
-      </div>
-    </div>
+  <div class="modal-body search-modal">
+    <uib-tabset>
+
+      <uib-tab index="0" select="ModalCtrl.selectTab(0)"
+        data-default-filter-tab
```
jniles: You don't need the "data-default-filter-tab" here.  It's used for end to end tests of our search modals.

client/src/modules/patients/lookupReturningPatient.modal.html
```diff
@@ -1,99 +1,306 @@
...
+                    <th translate>FORM.LABELS.GENDER</th>
+                    <th translate>FORM.LABELS.DOB_ABBREV</th>
+                    <th translate>FORM.LABELS.AGE</th>
+                    <th translate>FORM.LABELS.ORIGIN</th>
+                    <th translate>FORM.LABELS.PATIENT_ID</th>
+                    <th translate>FORM.LABELS.HOSPITAL_FILE_NR</th>
+                  </thead>
+                  <tbody>
+                    <tr ng-repeat="pat in ModalCtrl.choices">
+                      <td><button ng-click="ModalCtrl.useChoice(pat.uuid);">{{pat.display_name}}</button></td>
```
jniles: This button is a little jarring from a UI perspective.  We don't have anything like this in the app.  Here is how I would make it better:

I propose that we instead make this into an `<a href=""></a>` link that would select the patient. This will make it much more readable.  You could even include a sideways caret (&#8250;) `&#8250;`.

server/controllers/medical/patients/index.js
```diff
@@ -423,6 +427,289 @@ function searchByName(req, res, next) {
...
+ *     year matches vs exact date matches.
+ *   - In any case, the total score for each potential match is
+ *     scaled to be between 0 and 1.
+ *   - Specifying the sex and DOB can help to improve the ranking
+ *     of potential matches.
+ *
+ * @returns array of potential matching patient objects.
+ *          Each row has a additional 'matchScore' value that
+ *          indicates the quality of the match (0-1).
+ */
```
jniles, line 567: I think we need to include length as a really (minor) criteria.  For example, I can search for "jean" and get many 100% matches (jean jacques niles would be a match).  This doesn't totally make sense to me - we aren't 100% confident that is a match.

If we could add a weighting to the name so that the more information (3 name parts > 2 name parts > 1 name) would increase the score, that would be helpful.  Do you think that is possible?

jniles (commented):
The functionality here is really good.  Great documentation and tests as well!

The UI here needs a bit more work before I'm comfortable merging it into master.  It confused me about what to do (I didn't recognize the buttons as buttons) and then later got myself into a corner case where I could clear and return.  I also found a single match that seemed to automatically select that patient, which surprised me. 

I'd like to take a crack at this to make sure the UI/UX is as smooth and optimized as the back end.  There is some really great work here, and I can't wait to show it off!

server/controllers/medical/patients/index.js
jmcameron: Converted these to debug() calls.

client/src/modules/patients/lookupReturningPatient.modal.html
jmcameron: Removed.

server/controllers/medical/patients/index.js
jmcameron: I've added this functionality, updated the tests, and added a new test for this feature.

client/src/modules/patients/lookupReturningPatient.modal.html
jmcameron: This still needs to be done.  I'm deferring to @jniles for this.

jniles:
@jmcameron great work here!  I'm very happy with these results.

I've made some user interface and navigation changes.  Here is what it looks like:
cnCZziYfOY

Please give these changes a sanity check and let me know what you think.  If all looks good, feel free to run bors on it.

jmcameron:
I like the UI fix that you did.  It is intuitive and selecting by a link on the Patient ID is very clear.

PR #5264, opened 2021-01-04
feat(users): add "last_login" functionality

### LINKED ISSUE #5216: Add a "last login" column to the users page
It would be interesting to see the users that haven't logged in ever or in a long time.  Let's track that (if we don't already) and add a "last login" column to the users grid.  This is what it currently looks like:

image

jniles:
This commit adds the functionality to actually display the last login date of the user.  This column had been playing the role of "updated_at" rather than "last_login" since it was only changed automatically when user information had been changed.  Now, the last_login it updated whenever a user logs in.

I've also reset all values of the last_login to NULL in the migration script.

Closes #5216.

### DIFF
--- a/client/src/modules/users/users.js
+++ b/client/src/modules/users/users.js
@@ -65,6 +65,14 @@ function UsersController($state, $uibModal, Users, Notify, Modal, uiGridConstant
         cellClass : muteDisabledCells,
       },
       {
+        field : 'lastLogin',
+        type : 'date',
+        displayName : 'USERS.LAST_LOGIN',
+        cellFilter : 'date:"dd/MM/yyyy HH:mm:ss"',
+        headerCellFilter : 'translate',
+        cellClass : muteDisabledCells,
+      },
+      {
         field : 'action',
         displayName : '',
         cellTemplate : '/modules/users/templates/grid/action.cell.html',
--- a/server/models/schema.sql
+++ b/server/models/schema.sql
@@ -1735,16 +1735,17 @@ CREATE TABLE `unit` (
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
 DROP TABLE IF EXISTS `user`;
-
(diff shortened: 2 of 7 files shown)


client/src/modules/users/users.js
```diff
@@ -64,6 +64,14 @@ function UsersController($state, $uibModal, Users, Notify, Modal, uiGridConstant
         enableFiltering : true,
         cellClass : muteDisabledCells,
       },
+      {
+        field : 'lastLogin',
+        type : 'date',
+        displayName : 'USERS.LAST_LOGIN',
+        cellFilter : 'date:"dd/MM/yyyy HH:mm:ss"',
```
jmcameron, line 71: I like including the time.  I do wonder whether the 24 hour time should be used or the 12 hour time +am/pm?  I think Congo/French-speaking users are used to 24 hour time, but perhaps in the future this should be a regional option?

server/models/schema.sql
```diff
@@ -1735,16 +1735,17 @@ CREATE TABLE `unit` (
...
 
 DROP TABLE IF EXISTS `user`;
-
 CREATE TABLE `user` (
   `id`            SMALLINT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
   `username`      VARCHAR(80) NOT NULL,
   `password`      VARCHAR(100) NOT NULL,
   `display_name`  TEXT NOT NULL,
-  `email`         VARCHAR(100) DEFAULT NULL,
+  `email`         VARCHAR(99) DEFAULT NULL,
```
jmcameron, line 1743: Why this change to the length of the email field?

jmcameron (approved):
This looks fine.  The code changes are good and I tested it with bhima_test and it works fine. However, I did not test the SQL migration code (with IMCK for example).  However the SQL migration code looks fine.

server/models/schema.sql
jniles: Good question.  The new year's palm wine must not have worn off when I wrote this :smile:

mbayopanda (approved):
It works as expected 👍 

bors r+

PR #5267, opened 2021-01-04
fix(stock-movements): filter by "today" by default

### LINKED ISSUE #5258: Stock Movement Registry does not limit to "today" by default
The default period filter for Stock Movements Registry doesn't work on a fresh install of the system, despite the period being a "default" filter.  This is a bug that should be corrected.

How to test:
  1. Dump your browser's cache.
  2. Navigate to the Stock Movements Registry.
  3. Observe that the default period is not visible in the filter.

jniles:
Filters the stock movements registry by "today" by default.  It also unifies the filters for inline movements.

Closes #5258.

### DIFF
--- a/client/src/js/services/StockService.js
+++ b/client/src/js/services/StockService.js
@@ -2,10 +2,10 @@ angular.module('bhima.services')
   .service('StockService', StockService);
 
 StockService.$inject = [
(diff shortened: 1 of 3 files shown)


lomamech (approved):
A very good observation
image

PR #5268, opened 2021-01-04
fix(stock): refresh inventory adjustment on submit

### LINKED ISSUE #5255: Automatically refresh the page after a stock adjustment has been made.
Our users are getting frustrated with the stock adjustment module as they don't have a way to refresh the data after a stock adjustment has been made.  Right now, they need to exit the module, and re-enter the module to get the data to refresh.

jniles:
Refreshes the inventory adjustments page on submit so that users don't have to click the refresh button again to get current values in stock.

Closes #5255.

### DIFF
--- a/client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
+++ b/client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
@@ -129,8 +129,9 @@ function StockInventoryAdjustmentController(
   function startup() {
     vm.movement = {
       date : new Date(),
-      entity : {},
     };
+
+    setupStock();
   }
 
   // ============================ Inventories ==========================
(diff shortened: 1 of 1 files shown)


client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
```diff
@@ -129,8 +129,9 @@ function StockInventoryAdjustmentController(
   function startup() {
     vm.movement = {
       date : new Date(),
-      entity : {},
```
jmcameron, line 132: Why was this removed?

jmcameron (approved):
Code changes look fine.
I tested this with bhima_test and it works as advertised.

client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
jniles: Good question.  It was never used or needed.  I think it is copy/paste from somewhere where it was needed.

mbayopanda (approved):
LGTM 👍 

bors r+

PR #5269, opened 2021-01-05
Fixed issue with upload directories

### LINKED ISSUE #5207: UPLOAD_DIR doesn't seem to accept absolute paths
We're having a rather interesting bug in production where the UPLOAD_DIR doesn't accept absolute paths, but rather wants to make everything relative to the `client/` folder.  This isn't good.  We want to be able to set the absolute upload path to a place where future upgrades will not disturb it, particularly since upgrades can sometimes be destructive to the bin folder.

jmcameron:
Fix issue with upload directory always being under the bin directory.  Now it obeys the .env setting for UPLOAD_DIR.

**NOTES!**
- Only relative paths within the Bhima installation directory may be used
- Currently it will reject any absolute paths or paths that start with '..' for UPLOAD_DIR.
- Although it is probably possible to support absolute paths outside the Bhima installation, it will require updates to the app's 'app.use' function calls.  Without those changes, paths outside of the Bhima directory cannot be accessed via HTTPS calls without ENOENT errors.
- There are at least 3 ways to avoid conflicts with installation updates:
   - Create a new 'data' (or similar) directory inside the Bhima installation folder and add it to the .gitignore file.
   - Create a directory elsewhere and set up a symbolic/hard link to it within the Bhima directory
   - Revisit this issue and refactor to allow absolute paths.

**TESTING:**
- In your .env file, try changing UPLOAD_DIR to an absolute path and then a path beginning with '..'.  These should both fail.
   - NOTE that these failures are not handled very gracefully because this is an error that should be handled by the installer, not an end-user.
- Test the current default for the upload directory (eg, 'client/uploads'
   - Try uploading a file (eg in the Patient Registry, chose a patient and upload a document)
   - Verify the upload worked and the file is downloadable.  Do not delete it.
- Test in a new directory:
   - Use 'bhima_test' 
   - Create a folder within the Bhima installation directory (eg, 'data')
   - Change UPLOAD_DIR to the relative path for the new subdirectory the Bhima installation directory (eg, 'data').  
   - Enable the debugging option in .env:   DEBUG=app,errors,app:uploader
   - Monitor the client console for errors
   - Run 'yarn dev', notice in the server console the paths match what was given.  Also notice that it is NOT in the bin directory!
   - Try uploading a file (eg in the Patient Registry, chose a patient and upload a document)
   - Verify the upload worked and the file is downloadable
   - Delete the new file
   - Exit 'yarn dev' and delete the temporary upload folder
- Revert to the default UPLOAD directory (eg, 'client/uploads') and verify it still works and that the file uploaded earlier is still accessible.  Delete the file, if desired.
- Undo the debug settings in .env

### DIFF
--- a/server/lib/uploader.js
+++ b/server/lib/uploader.js
@@ -41,7 +41,7 @@ const dir = process.env.UPLOAD_DIR || defaultDir;
 if (path.isAbsolute(dir) || dir.startsWith('..')) {
   throw new Error(`UPLOAD_DIR (${dir}) must be a relative path within the BHIMA software installation!`);
 }
-const rootDir = path.resolve(`__dirname/../..`);
+const rootDir = path.resolve(`${__dirname}/../..`);
 const fsdir = path.join(rootDir, dir); // global path
 debug('UPLOAD_DIR: ', dir);
 debug('UPLOAD_DIR Abs dir: ', fsdir);
@@ -79,14 +79,13 @@ function Uploader(prefix, fields) {
   // configure the storage space using multer's diskStorage.  This will allow
   const storage = multer.diskStorage({
     destination : async (req, file, cb) => {
-      const folder = path.join(dir, directory);
-      debug(`creating upload directory ${folder}.`);
 
       try {
         // NOTE: need absolute path here for mkdirp
-        const fullFolderPath = path.join(rootDir, directory);
+        const fullFolderPath = path.join(fsdir, directory);
+        debug(`creating upload directory ${fullFolderPath}.`);
         await mkdirp(fullFolderPath);
-        cb(null, directory);
+        cb(null, fullFolderPath);
       } catch (err) {
         cb(err);
       }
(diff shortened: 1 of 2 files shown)


server/lib/uploader.js
```diff
@@ -35,17 +35,24 @@ const { uuid } = require('./util');
...
-const defaultDir = 'uploads'; // NOTE: this must be a relative path
-const dir = process.env.UPLOAD_DIR || defaultDir; // relative path
-const fsdir = path.join(process.cwd(), dir); // global path
+const defaultDir = 'uploads';
+const dir = process.env.UPLOAD_DIR || defaultDir;
+// NOTE: 'dir' must be a relative path (for http requests to work)
+if (path.isAbsolute(dir) || dir.startsWith('..')) {
+  throw new Error(`UPLOAD_DIR (${dir}) must be a relative path within the BHIMA software installation!`);
+}
+const rootDir = path.normalize(`${process.cwd()}/..`);
```
jniles: This isn't guaranteed to be the root directory in all circumstances.  As explained in [this SO answer](https://stackoverflow.com/a/9874415), the `process.cwd()` is wherever the `node` call was made.  So, it can changed, depending on where the executing terminal is in the file system.

Instead, you probably want `__dirname`.   [Here is a really good tutorial explaining how it works](https://www.digitalocean.com/community/tutorials/nodejs-how-to-use__dirname).  This will never change, no matter who calls it or from where it is called.

jniles:
Sorry about the force push 😞

jniles:
Got it working.  Thanks so much for your help @jmcameron !

bors r+

PR #5271, opened 2021-01-06
Add better explanation of UPLOAD_DIR

### LINKED ISSUE #5207: UPLOAD_DIR doesn't seem to accept absolute paths

jmcameron:
Based on updates in PR https://github.com/IMA-WorldHealth/bhima/pull/5269.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -27,8 +27,10 @@ SESS_SECRET='XopEn BlowFISH'
 # define logging level for the npm debug module.
 DEBUG=app,errors
 
-# define the directory where we will upload reports to.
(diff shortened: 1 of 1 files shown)


jniles (approved):
LGTM.

bors r+

PR #5273, opened 2021-01-07
improvement(Employee payslip and description in journal)

### LINKED ISSUE #4654: Add employee reference to Bulletin de Paie
We have a very nice looking Bulletin de Paie (see below).  It would be even nicer looking if we have the BHIMA reference employee code on it.  This code is of the form `EM.XXX.123`.  Let's add it somewhere on the sheet so that we can keep track of these items.

image

lomamech:
- Addition of the employee's name, on the description of the transaction
  in posting journal

- Adding employee reference to payslip

closes #4654


lomamech:

jniles (approved):
This is exactly what I was thinking. Great work!  The code looks good to me.

bors r+

PR #5274, opened 2021-01-07
fix and enhance the distribution to service section

### LINKED ISSUE #5261: Show sum of medicines distributed to different services in Sorties de Stocks
In our report of Sorties de Stocks, it would be nice to see the quantity of stock consumed by service, grouped by stock item.  Currently, they are grouped by stock movement.  See below:

image

In this image, the service "Dispensaire" has consumed 6000 capsuls of medicines.  I obtained this by counting.  We should provide a nicer user interface see this grouped by inventory item with sums.

mbayopanda:
This PR is related to #5261 

image

Close #5261


jniles (commented):
Looks good!

One thing: the currencies don't seem to be respected on the grouping row.  Can you look into that?


Otherwise, looks good to me!

mbayopanda:
@jniles changes landed

jniles:
This works!  Thanks!

bors r+

PR #5275, opened 2021-01-07
Add default patient discharge types

### LINKED ISSUE #5245: Add default discharge types to production systems
Currently the default data for production sites in server/models/bhima.sql has several predefined discharge types and they are available in bhima_test.  However, the IMCK and Vanga databases do not currently have them defined.    IMCK has requested that we support patient visits and discharges, etc, in the near future so we need to fix this.

We need to add migration code that adds these default `discharge_types` to the production systems.

**A side issue here:**  Although the `discharge_type` is defined in the database, its use depends on data embedded in the Bhima javascript code.   The `discharge_type` values contain the i18n language tokens for each type item.   This means that adding a new discharge type is awkward:  Not only do you need a new entry in the `discharge_type` table, but you have to add it to the i18n language tokens as well.    Having the discharge type in the database has limited value.  Why not make it all in the code?    Note that this is not the only data that is implemented awkwardly like this; see many of the other `_type` definitions in the database schema.sql.   Even better, it could all be in the database so it could be added/edited by end-users -- but this is complicated too by the translation issue.

jmcameron:
Add the default patient `discharge_type`s to all sites.

**Suggestions for Testing**

- Use the IMCK database (which has the `discharge_type` table but not  does not have data)
- Edit .env appropriately
- Download the latest IMCK database
- Run the migration script
- Use MySQL to verify the additions
   - mysql -u bhima -p --database=imck
   - select * from discharge_type  ==> Should show 9 entries
- Run bhima (yarn dev)
- Go to the Hospital > Patient Registry
- In the patient registry, select a period so you see patients
- Click on a patient NAME (clicking on the Reference gives the patient ID document, which does not allow the next step)
- In the Patient Details, click on the [Admit Patient] button (bottom right)
   - Fill out the form and click [Submit]
   - The button on the bottom right of the Patient Details should change to [Discharge Patient]
- Click on the [Discharge Patient] button
   - You should see the form with 9 options for discharge type.  Click on, add diagnosis and [Submit]
   - The submission should be succesful
- Visit the Hospital > Visits Registry
   - You should see the  visit created above listed with the discharge type


jniles:
@lomamech, could you give this a review?

lomamech (approved):
We never noticed it just because so far, hospitals that use bhima do not yet use Patient Management

@jmcameron  Please resolve the conflict with the branch master

PR #5279, opened 2021-01-08
Moved tags to end of lots search form

### LINKED ISSUE #5184: Move "Tags" input lower in the search modal
We use this much less often than any of the other search options.  Let's move it lower in the search modal so that getting to the "expiration date" and other more frequently used options is easier.

image

jmcameron:
Cosmetic change to move Tags section to the end of the Lots search form.


jniles (approved):
LGTM.

bors r+

PR #5280, opened 2021-01-08
Cache entry date, expiration date, and tags in lots search

### LINKED ISSUE #5182: The expiration date search field is not restored for a second search in the lots in stock registry
Most search fields are restored when you search for a second time.  For example, the "depot" is restored successfully.  However, the date range for "expiration date" is not.  This is somewhat confusing.  See below:

s8vB9i4Hzj

Expected result:
The date range for expiration date should be restored and cached in the date range component for a second search.

jmcameron:
Make sure that the entry dates and expiration dates are remembered between uses of the lots search modal.

**Suggestions for testing**
- Test in the Stock > Lots Registry
- Use the [Search] button and test adding entry dates, expiration dates, and tags
-  Submit and do the searches again to verify that the previous values are recalled into the form
- Also for Tags (at the bottom of the form) try this stress test:
   - Clear the tags and [Submit]
   - Reopen the Search modal
   - Chose a tag.  Do NOT press [Submit].  Do NOT click on the "Clear" link above the right end of the Tags entry.
   - Use the [X] on the tags right end to delete the tag.  This should result in no tags left.
   - Now click the [Submit] button
   - This should work and not create any spurious empty tags:   'xyz: []'

### DIFF
--- a/client/src/modules/stock/lots/modals/search.modal.html
+++ b/client/src/modules/stock/lots/modals/search.modal.html
@@ -24,7 +24,7 @@
             <bh-clear on-clear="$ctrl.clear('depot_uuid')"></bh-clear>
           </bh-depot-select>
 
-          <!-- Inventaire -->
+          <!-- Inventory -->
           <bh-inventory-select
             inventory-uuid="$ctrl.searchQueries.inventory_uuid"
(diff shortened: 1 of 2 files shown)


jniles (approved):
Tested and working.  Thanks!

PR #5281, opened 2021-01-10
feat(stock): add difference column to inventory adjustment

### LINKED ISSUE #5259: Add column to Stock Adjustments to show difference between the old stock and new stock
Our friends in Vanga have requested that we add a column to the stock adjustments grid to automatically show the difference between the old stock and the new stock.  The module is https://hev.bhi.ma/#!/stock/inventory-adjustment.

Basically, we need to add an extra column here:

image

That will show the difference between the old quantity and the quantity the user enters.

Buzunda:
Adds difference column to inventory adjustment page/table

### DIFF
--- a/client/src/modules/stock/inventory-adjustment/templates/difference.tmpl.html
+++ b/client/src/modules/stock/inventory-adjustment/templates/difference.tmpl.html
@@ -1,0 +1,1 @@
+<div class="ui-grid-cell-contents">{{row.entity.quantity - row.entity.old_quantity}}</div>
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -303,6 +303,7 @@
   "INVENTORY_ADJUSTMENT":{
     "TITLE":"Stock Inventory Adjustment",
     "DESC":"The inventory adjustment module allows you to adjust the quantities in stock of all lots in a depot at once.  Lot quantities will be increased or decreased to match the new quantity specified.",
+    "DIFFERENCE":"Difference",
     "NEW_QUANTITY":"New Quantity",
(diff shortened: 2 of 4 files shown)


jmcameron:
Thanks for the PR!  I'll take a look at it tomorrow.  The rest of our team in on travel now or unavailable.

client/src/modules/stock/inventory-adjustment/templates/difference.tmpl.html
```diff
@@ -0,0 +1 @@
+<div class="ui-grid-cell-contents">{{row.entity.old_quantity - row.entity.quantity}}</div>
```
jmcameron: Not sure about this:  Shouldn't this be:  NEW - OLD?  That way positive values mean the inventory has increased and vice versa.

jmcameron (changes requested):
I suggested making the difference be NEW - OLD.  @jniles What do you think?

Otherwise it looks fine.  I've tested it and it seems to work.

client/src/modules/stock/inventory-adjustment/templates/difference.tmpl.html
jniles: Hey @buzunda, this is great!  I do agree with @jmcameron, it makes more sense to do `quantity - old_quantity`.  I believe this is what the user who requested the feature was looking for. 

Otherwise, perfect!

Buzunda: ok, I will change this soon

jmcameron:
Thanks for the fix.  Great job!

bors r+

jmcameron (approved):
Thanks!

PR #5283, opened 2021-01-12
Allow multi depot stock sheet report

### LINKED ISSUE #5254: Do not require a depot for the fiche de stock
Our friends at Vanga want to see the aggregated stock in _all_ depots, not simply a single depot.  They need this capability to satisfy a group of auditors for the World Bank's PBF project.

The idea is that we would remove the requirement on the Fiche de Stock report to select a depot.  We would allow the user to optionally specify the depot to see the fiche to stock for a single depot.  If the depot is not specified, we would show the stock for all depots.  Since the Fiche de Stock only shows the quantity and value of the batches of medications, it is straightforward to combine these values into a single report.

jmcameron:
Make the Depot optional for Stock Sheet (Fiche) reports.

Note that this PR also fixes a bug in the stock sheet dialog that was preventing the dateTo from being cached properly.

**Suggestions for Testing**

- Use one of the production sites
- yarn dev
- Use the Stock > Items in stock to identify an inventory item that is in stock and note the depot
- Stock > Report > Stock Sheet (Stock Fiche)
  - Select Depot and inventory item and preview the report
  - Cancel
  - Clear the Depot and try the preview again.
     - Note that there is a new column for the Depot that should show the depot for each stock item.  This column is not present when one Depot is selected.

**After an initial review and dealing with bugs and suggestions, I'll rebase before releasing.**

### DIFF
--- a/client/src/modules/reports/generate/stock_sheet/stock_sheet.config.js
+++ b/client/src/modules/reports/generate/stock_sheet/stock_sheet.config.js
@@ -22,7 +22,7 @@ function StockSheetConfigController($sce, Notify, SavedReports, AppCache, report
   // check cached configuration
   checkCachedConfiguration();
 
-  vm.changeHavingDateInterval = value => {
+  vm.changeHavingDateInterval = (value) => {
     vm.hasDateInterval = value;
     vm.reportDetails.dateFrom = value ? new Date() : undefined;
     vm.reportDetails.dateTo = new Date();
@@ -68,7 +68,6 @@ function StockSheetConfigController($sce, Notify, SavedReports, AppCache, report
     // format date for the server
     options.dateTo = moment(options.dateTo).format('YYYY-MM-DD');
     options.lang = Languages.key;
-
     return SavedReports.requestPreview(reportUrl, reportData.id, options)
       .then((result) => {
         vm.previewGenerated = true;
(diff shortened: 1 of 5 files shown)


lomamech (commented):
The first thing I think about doing is to present this Report to the IMA Pharmacist team for their reactions.

image

lomamech:
@jmcameron  the sum of the quantities in stocks of this two depots correspond to that of the stock card, I will only wait for the reaction of pharmacists
image

lomamech:
@jmcameron 
After contacting the pharmacists, they approved the idea of being able to consult the stock card for several depots, but --I would like to suggest
- that we can activate the filter of depots to which the user has the right of access
- but also that the option of selection of deposits can be multichoice, for example we could know the quantity of a product of the deposits of a health zone

jniles:
I would make the second point a separate issue.  It may be a lot more complicated to fix.  @jmcameron let us know if you would be able to tackle that in this issue or in a separate one.

jmcameron:
> I would make the second point a separate issue. It may be a lot more complicated to fix. @jmcameron let us know if you would be able to tackle that in this issue or in a separate one.

I agree.  Lets go ahead and push this one out.  It is usable as is.  I'll add a new Issue for the enhancements you mentioned (which make sense to me).    I also think that we should show the table sorted by Depot then date (perhaps as an option).   And Adding subtotals for each depot could also be useful.   have to think about how to do that cleanly (in terms of the UI).

jmcameron:
@lomamech, if you agree to push the extra functionality into a new issue, please go ahead and release this PR.  Thanks!

lomamech:
There is still the label **Do not merge**

jmcameron:
> There is still the label **Do not merge**

Oops!  I'll rebase it and then remove the label.

jmcameron:
> There is still the label **Do not merge**

@lomamech It is ready now.  Thanks.

PR #5284, opened 2021-01-12
feat: add ignore reversed record to vouchers

### LINKED ISSUE #5278: Add an option to ignore reversed vouchers to voucher search modal
In our cash and invoice registry search modals, we have a filter to "ignore reversed records".  We need to add this to the voucher registry as well.  Here is what the "ignore reversed invoices" looks like: 

image

We need to implement the same thing for the voucher registry.  If we want to do sums of different search queries, we'll want to exclude the annulled records.  It should be added to the voucher search modal:

image

jniles:
Implements #5278.

In this case, we provide three options as suggested by @mbayopanda.

1. Exclude all reversed records
2. Include only reversed records
3. Exclude all reversed records _and_ reversal records

This allows users to filter out both the records that are reversed on the credit/debit notes that are made against those records.

Closes #5278.

image

### DIFF
--- a/client/src/modules/vouchers/vouchers.service.js
+++ b/client/src/modules/vouchers/vouchers.service.js
@@ -47,7 +47,7 @@ function VoucherService(
     { key : 'uuid', label : 'FORM.LABELS.REFERENCE' },
     { key : 'user_id', label : 'FORM.LABELS.USER' },
     { key : 'reference', label : 'FORM.LABELS.REFERENCE' },
-    { key : 'reversed', label : 'FORM.INFO.ANNULLED' },
+    { key : 'reversed', label : 'VOUCHERS.GLOBAL.REVERSED_RECORDS' },
     { key : 'description', label : 'FORM.LABELS.DESCRIPTION' },
     { key : 'entity_uuid', label : 'FORM.LABELS.ENTITY' },
     { key : 'account_id', label : 'FORM.LABELS.ACCOUNT' },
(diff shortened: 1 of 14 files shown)


jniles:

client/src/modules/vouchers/vouchers.service.js
```diff
@@ -47,7 +47,7 @@ function VoucherService(
     { key : 'uuid', label : 'FORM.LABELS.REFERENCE' },
     { key : 'user_id', label : 'FORM.LABELS.USER' },
     { key : 'reference', label : 'FORM.LABELS.REFERENCE' },
-    { key : 'reversed', label : 'FORM.INFO.ANNULLED' },
+    { key : 'reversed', label : 'VOUCHERS.GLOBAL.REVERSED_RECORDS' },
```
mbayopanda, line 50: 👍🏽

mbayopanda (commented):
The feature is working fine, we have just a text not translated in the modal of reversal

image

And what about the credit note row added after a reversal? We clean the registry by removing reserved voucher but we still have new lines added by the reversal operation

image

jniles:
@mbayopanda this is ready for another review.

jniles:
@mbayopanda this is ready for another review.

jniles:
@mbayopanda what is the status of this PR?

jniles:
@mbayopanda is there anything left to be done on this PR?

mbayopanda (approved):
Excellent 👍🏽 

The feature works as expected

bors r+

PR #5285, opened 2021-01-12
Fix: critical stock bugs

jniles:
Fixes two critical bugs related to stock movements.  Both are due to variable changes that were not properly made.

1. The stock movement registry search modal would not open due to a broken reference to `Stock.filter.inlineMovement` which should be `Stock.filter.movements`.
2. The second is that no automatic stock accounting was taking place (!) due to an old `req.session.enterprise.auto_stock_accounting` reference that needed to point to the newer `stock_settings`.

To test these:
   1. Try to search for a stock movement in the registry.  The search button is inoperable.
   2. Make a stock movement (exit to service, for example).  Observe that there is no corresponding exists written to the database in the voucher registry.  On this branch, there will be.

### DIFF
--- a/client/src/modules/stock/movements/modals/search.modal.js
+++ b/client/src/modules/stock/movements/modals/search.modal.js
@@ -37,7 +37,7 @@ function SearchMovementsModalController(data, Instance, Periods, Store, util, St
   };
 
   // map key to last display value for lookup in loggedChange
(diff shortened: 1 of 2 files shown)


jmcameron (approved):
Looks good.  I tested this with the IMCK database and it works as described.

PR #5286, opened 2021-01-12
Add description and dhis2_uid to depot table

### LINKED ISSUE #5025: Add a description to the depot table
We should add a `TEXT` description to the depot table.  The main motivation is to stick DHIS2's `uid` in it for that structure.  This will make it easy for us to link the two in the bespoke implimentation of medicine tracking across the country.  It also makes sense to use a description for some depots in a hospital setting.

jmcameron:
Add support for `description` and `dhis2_uid` fields to the `depot` table and adds the ability to add/modify those fields when creating or editing depots.  Also update the migration file to add these new fields to production servers.

NOTE:   While editing the next/migration.sql file, I updated a previous entry to use the add_column_if_missing() function so that the migration file can be applied multiple times.

Testing:  Try adding/editing the Description and DHIS2 UID fields when creating or editing Depots.

### DIFF
--- a/client/src/i18n/en/depot.json
+++ b/client/src/i18n/en/depot.json
@@ -27,6 +27,7 @@
     "WAREHOUSE" : "Warehouse",
     "WAREHOUSE_INFO" : "This depot can distribute to other depots",
     "ALLOWED_DESTINATION_DEPOTS":"Allowed destination depots",
+    "DHIS2_UID": "DHIS2 UID",
     "MODAL" : {
       "SELECTION" : "Depot Selection",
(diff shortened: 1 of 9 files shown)


jmcameron:
@jniles Please take a look at why the integration test fails (works for me).  I added the new field to the Bhima_test DB.  Not sure why it is causing problems on the CI server. Thanks!

jniles:
@jmcameron, found it!  You need to make the changes to `test/data/stock.sql` as well.  We build depots there too and you changed the structure of the `depot` table.

jniles (approved):
Looks good to me!

bors r+

PR #5294, opened 2021-01-14
Add tooltips to lot registry column headers

jmcameron:
Added tooltips to column headers in the Stock > Lots Registry grid.
Although it is possible to hide columns, I thought this was a helpful way to deal with the compressed column headers without actually removing them.  Here is what it looks like when you hover over one of the column headers:

image


mbayopanda (approved):
bors r+

mbayopanda:
Good catch @jmcameron 👍🏽

PR #5299, opened 2021-01-15
refactor: include all release information in package.json

### LINKED ISSUE #5287: Updates to the release process
We've been making slow improvements to the release process.  We now bundle code during our release process to allow us to simply download a release and run it.

There are a couple of issues with the current way we are doing releases that need to be improved:
  1. Include the `package.json` in the tarball.  This means we have no reference for downloading node_modules!  We need to include the package.json when we make a release.
  2. Include the `.env.sample` in the release, probably renamed to `.env` to help save the user a step.  Basically, this will serve as the initial template for them to put their variables in and then go.
  3. We currently tar/zip the `bin` folder.  That is fine, except that you don't want to ship the `bin` folder.  Let's rename that folder to the current release before zipping so that we have some reference when we unzip/untar on the target machine.  Right now, if we unzip/untar two releases, they both clobber each other.  A nicer name would be `bhima-1.xx.yy` where `1.xx.yy` is the version number.

jniles:
Includes all the necessary information when releasing, including the `package.json`, `.env.sample`, and renaming the bhima folder to something more amenable to installing/upgrading.

Closes #5287.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,10 +41,16 @@
         "rm -rf bin",
         "cross-env NODE_ENV=production yarn build",
         "rm bin/.env",
-        "cross-env tar -czvf ${version}.tar.gz bin"
+        "cp package.json bin/",
+        "cp yarn.lock bin/",
+        "cp .env.sample bin/.env",
+        "mkdir bhima-${version}",
+        "mv bin bhima-${version}/",
+        "cross-env tar -czvf ${version}.tar.gz bhima-${version}"
       ],
       "after:release": [
         "rm ${version}.tar.gz",
+        "rm -rf bhima-${version}",
         "echo Successfully released ${name} v${version} to ${repo.repository}."
       ]
     },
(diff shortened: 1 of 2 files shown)


jniles:
@jmcameron it looks like @mbayopanda is out of power this evening.  Could you give me a review?  I would like to get a release in before my trip tomorrow and this is necessary before the release.

package.json
```diff
@@ -41,10 +41,16 @@
         "rm -rf bin",
         "cross-env NODE_ENV=production yarn build",
         "rm bin/.env",
```
jmcameron, line 43: This line is not necessary since you're overwriting it below

jmcameron (approved):
Looks good!

jmcameron:
A suggestion:  Put an INSTALL.txt/UPGRADE.txt file(s) into the release bin/ directory with a checklist of things to do during the installation/upgrade process.   I understand these things are currently done manually and that different developers have their own checklists.  It would be useful to have a consolidated and reviewed checklist to avoid installation/upgrade errors.

mbayopanda (approved):
The changes look good to me 👍 

bors r+

PR #5307, opened 2021-01-22
Enable merging lots

### LINKED ISSUE #5165: Feature Request: allow users to merge lots
We need a way to merge lots.  There are many lots with the same name at IMCK that should probably be the same lot.  This would probably avoid quite a few mistakes.

jmcameron:
This updates lets the user merge lots from the Lots Registry.  

This is a preliminary release to get some feedback on coding, UIs, etc.  I'm continuing to work on the release.  Once those issues have been addressed, I've added some integration tests, and it is ready to go, I'll remove the [Do No Merge] flag.

### Testing
- Use the IMCK database
- Go to the Lots Registry
- Update the search to find lots named:  arp
- Click on the action menu and select the last item (Find Duplicate Lots).
   This finds all lots that have matching labels and inventory_uuid-s. 
- Note that there is a column for the lot to keep (initially set to the current lot) and
   a column for 'Merge'.  
- The selected lot is marked in green.
- Select any lot you wish to merge with the selected lot.
- Click on the [Merge] button
- When you get back to the lots registry, try the merge operation on the same lot.  Notice that no matching will be found.
- Clear the search and try the merge operation on a random  lot.  Very likely not matching lots will be found.


client/src/js/services/LotService.js
```diff
@@ -1,9 +1,9 @@
 angular.module('bhima.services')
   .service('LotService', LotService);
 
-LotService.$inject = ['PrototypeApiService', '$http', 'util'];
+LotService.$inject = ['PrototypeApiService', '$http', '$httpParamSerializer', 'util'];
```
jniles: I do not see where you use `$httpParamSerializer`.  If you don't use it, please remove it.

client/src/js/services/LotService.js
```diff
@@ -14,6 +14,23 @@ function LotService(Api, $http, util) {
       });
   };
 
+  lots.dupes = (params) => {
```
jniles, line 17: Since this function sends a GET request to the server, could you name it some action?  Something like `lookupDupes()` or `findDupes()` or `getDupes()`?  That way, it makes more sense when reading from a controller without this context.

server/controllers/stock/lots.js
```diff
@@ -92,6 +94,85 @@ async function update(req, res, next) {
...
+ *      references to the lot to keep.
+ *   2. Delete the lot to be merged
+ *
+ * lot_tag : lot_uuid
+ * stock_assign : lot_uuid
+ * stock_movement :
+ */
+async function merge(req, res, next) {
+  const uuid = db.bid(req.params.uuid);
+  const lotsToMerge = req.params.lots_to_merge.split(',').map(db.bid);
```
jniles: The best way to handle this isn't to embed in the URL `lotsToMerge` as you've done here, but instead pass it in the body.  It will save all the parsing for you.  Unless there is a specific reason you've done this, my advice would be:

1. Make the URL into `/lots/:uuid/merge`
2. Pass the other uuids into the `body` of the POST request.  
3. Access them from here via:
```js
const lotsToMerge = req.body.lotsToMerge;
````

File: server/controllers/stock/lots.js [hunk 4]
```diff
@@ -92,6 +94,85 @@ async function update(req, res, next) {
...
+
+  const updateLotTags = 'UPDATE lot_tag SET lot_uuid = ?  WHERE lot_uuid = ?';
```
jniles, line 145: Nice :+1:

server/controllers/stock/lots.js
```diff
@@ -92,6 +94,85 @@ async function update(req, res, next) {
...
+    wheres.push(`l.initial_quantity = ${req.query.initial_quantity}`);
+  }
+  if (req.query.entry_date) {
+    wheres.push(`l.entry_date = ${req.query.expiration_date}`);
+  }
+  if (req.query.expiration_date) {
+    wheres.push(`l.expiration_date = ${req.query.expiration_date}`);
+  }
+  if (wheres.length === 0) {
+    throw new NotFound(`Search for duplicate lots failed.`);
```
jniles: This `throw` will cause issues.  Because you are in an expressjs router function, you'll want to do `return next(new NotFound())` to ensure that it is passed correctly to the error handler.

Alternatively, you can wrap the whole thing in a try/catch and call `next(err)` in the catch block.

server/controllers/stock/lots.js
```diff
@@ -92,6 +94,85 @@ async function update(req, res, next) {
...
+ * @description
+ * Returns all lots with the given label or matching field(s)
+ * inventory_uuid, initial_quantity, entry_date, expiration_date
+ *
+ * TODO: After getting this working purge unneeded params
+ *
+ */
+function dupes(req, res, next) {
+  const wheres = [];
+  if (req.query.label) {
```
jniles: By the way, we have a fairly elegant, if undocumented, way of composing these filters using [`lib/filter.js`](https://github.com/IMA-WorldHealth/bhima/blob/master/server/lib/filter.js).

In essence you set up your base query and pass it an options object, then tell it which options are valid to use as WHERE query parts.  As a bonus, it does SQL injection escapes.  Here is how you would implement it here:
```js

const FilterParser = require('path/to/lib/filter');

const detailsQuery = `SELECT * FROM lots `;

// make sure all UUIDS are correctly parsed
const options = db.convert(req.query, ['inventory_uuid', 'lot_uuid', 'depot_uuid', /* etc */ ]);

// pass all the req.query options to the filterparser
const filters = new FilterParser(options, { tableAlias : 'lots' });

// tell it that "label" property should be a LIKE query (full-text index)
filters.fullText('label');

// tell it that inventory_uuid should be exactly equal
filters.equals('inventory_uuid');

// quantity should also be equal
filters.equals('initial_quantity');

// finally, let's get the statement out:
const statement = filters.applyQuery(detailsQuery);
const params = filters.parameters();

// pass to db.exec to execute them
return db.exec(statement, params);
```

jniles: This will ensure that you are correctly escaping SQL injection attacks and is also a lot more flexible.  We use it everywhere - you can draw inspiration from [cash.js](https://github.com/IMA-WorldHealth/bhima/blob/master/server/controllers/finance/cash.js#L128) which has a pretty good set of filters.

jniles (commented):
@jmcameron thanks for working on this!  I've given you some guidance on the filters which are a good exercise to learn.  We should probably have a skype chat about it - it's easier to screenshare and explain that way.

client/src/js/services/LotService.js
jmcameron: Done

client/src/js/services/LotService.js
jmcameron: Done

server/controllers/stock/lots.js
jmcameron: Done.

server/controllers/stock/lots.js

server/controllers/stock/lots.js
jmcameron: Done. Thanks for the pointers!

test/integration/mergeLots.js
```diff
@@ -0,0 +1,240 @@
...
+const UUID = require('uuid').v4;
+const helpers = require('./helpers');
+const db = require('../../server/lib/db');
+
+function uuid() {
+  return UUID().toUpperCase().replace(/-/g, '');
+}
+
+const preTestInfo = [
+  { table: 'lot', count: 0 },
```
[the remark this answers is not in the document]
jmcameron: Not quite finished yet...

jmcameron:
@jniles This PR is now ready for a final review.  When you're done, I'll make whatever mods you suggest and then rebase so it can be released.

jmcameron:
I was pleased at how I set up the mock items in the test, so let me know what you think of that.

jmcameron:
A note on this pull request:  The user invokes the the merge lot function via the action menu in the lot registry.   I view this approach as a temporary measure.  My next step is to create a separate "Find Duplicate Lots" page that will list lots that have duplicates and the merge lots dialog/function would be invoked from there.   Invoking it from the Lot Registry not very user-friendly since there is not an easy way to determine which lots need merging from that interface.  Once the "Find Duplicate Lots" page is enabled, we can remove the action menu item from the lot registry.  That also means that we can control access to that functionality to admins who are specifically enabled to do such operations (instead of anyone with access to the lot registry).

client/src/modules/stock/lots/modals/duplicates.modal.html
```diff
@@ -0,0 +1,71 @@
...
+
+  <div class="modal-body">
+    <div ng-if="$ctrl.lots.length === 1">
+      <span translate>LOTS.NO_LOTS_FOUND</span>
+    </div>
+    <div ng-if="$ctrl.lots.length > 1">
+      <span translate translate-values="{'N':$ctrl.lots.length - 1}">LOTS.FOUND_N_LOTS</span>
+      <table class="table table-condensed table-bordered">
+        <thead>
+          <th>Keep</th>
```
jniles: Looks like these titles need to be translated.  Are you hoping for someone's help on that?

client/src/modules/stock/lots/modals/duplicates.modal.html
```diff
@@ -0,0 +1,71 @@
...
+      <li class="title" translate>FORM.LABELS.DUPLICATE_LOTS</li>
+    </ol>
+  </div>
+
+  <div class="modal-body">
+    <div ng-if="$ctrl.lots.length === 1">
+      <span translate>LOTS.NO_LOTS_FOUND</span>
+    </div>
+    <div ng-if="$ctrl.lots.length > 1">
+      <span translate translate-values="{'N':$ctrl.lots.length - 1}">LOTS.FOUND_N_LOTS</span>
```
jniles, line 19: Nice translate-values :+1:

client/src/modules/stock/lots/modals/duplicates.modal.html
```diff
@@ -0,0 +1,71 @@
...
+          <th>Initial Quantity</th>
+          <th>Unit Price</th>
+          <th>Entry Date</th>
+          <th>Expiration Date</th>
+        </thead>
+        <tbody>
+          <tr ng-repeat="lot in $ctrl.lots track by lot.uuid"
+              ng-class="{ 'selected-lot' : lot.selected }">
+            <td class="center">
+              <input type="radio" name="selected" ng-if="lot.uuid === $ctrl.selectedLot.uuid"
```
jniles: It looks like you are trying to show a checked radio button if the lot is the selected lot.  This is best accomplished with `ngModel` or `ngChecked`.  You are using `ngIf` which injects/removes HTML based on conditions.  While this works in practice, conceptually it isn't what you want to do.

Here is how I would suggest you improve this code.

1. Make a single input element that is a radio button.
2. Set ng-model of the radio button to `$ctrl.selectedLot.uuid`.
2. Change the `ngClick` to `ngChange`.  This is necessary to prevent a race condition with `ngModel` where the model with change, then the click will fire changing the model, etc.  I've forgotten the technical details of "why", but all I know if you will be much happier with `ngChange` as your callback.

client/src/modules/stock/lots/modals/duplicates.modal.html
```diff
@@ -0,0 +1,71 @@
...
+              <input type="checkbox" name="merge" ng-if="lot.uuid !== $ctrl.selectedLot.uuid"
+                ng-model="lot.merge">
+              <input type="checkbox" name="merge" ng-if="lot.uuid === $ctrl.selectedLot.uuid"
+                disabled="true"
+                ng-model="lot.merge">
+            </td>
+            <td>{{lot.label}}</td>
+            <td>{{lot.quantity}}</td>
+            <td>{{lot.initial_quantity}}</td>
+            <td>{{lot.unit_price}}</td>
```
jniles: I think this price should be in the enterprise currency. 

```js
{{ lot.unit_price | currency:$ctrl.enterprise.currency_id }}
```

client/src/modules/stock/lots/modals/duplicates.modal.html
```diff
@@ -0,0 +1,71 @@
...
+      </table>
+
+    </div>
+  </div>
+
+  <div class="modal-footer">
+    <button type="button" class="btn btn-default" ng-click="$ctrl.cancel()" data-method="cancel" translate>
+      FORM.BUTTONS.CANCEL
+    </button>
+    <bh-loading-button loading-state="ModalForm.$loading" ng-if="$ctrl.lots.length > 1">
```
jniles, line 65: Instead of `ng-if` here, could you use the `disabled` property?  To me, it is a better user experience preventing pop-in.  What do you think?

```html
     <bh-loading-button loading-state="ModalForm.$loading" disabled="$ctrl.lots.length > 1"> 
 ```

File: client/src/modules/stock/lots/modals/duplicates.modal.js [hunk 13]
```diff
@@ -0,0 +1,71 @@
+angular.module('bhima.controllers')
+  .controller('DuplicateLotsModalController', DuplicateLotsModalController);
+
+// dependencies injections
+DuplicateLotsModalController.$inject = [
+  'data', '$state', '$uibModalInstance', 'LotService', 'NotifyService', '$translate',
+  // 'util', 'Store', 'PeriodService', 'StockService',
+  // 'SearchModalUtilService',
```
jniles: If these aren't going to be used, go ahead and clean them up.

client/src/modules/stock/lots/modals/duplicates.modal.js
```diff
@@ -0,0 +1,71 @@
...
+  vm.selectLot = selectLot;
+  vm.cancel = cancel;
+  vm.submit = submit;
+  vm.selectedLot = null;
+
+  function startup() {
+    Lots.read(data.uuid)
+      .then(selectedLot => {
+        vm.selectedLot = selectedLot;
+        Lots.dupes({ label : vm.selectedLot.label, inventory_uuid : vm.selectedLot.inventory_uuid })
```
jniles, line 22: It looks like you are missing error handling on the `Lots.dupes()`.  You'll need a `.catch()` statement somewhere.  The most logical place would be to `return Lots.dupes();`, then move the `.then()` callback into the promise chain formed by `Lots.read()`.  Like this:

```js
function startup() {
    Lots.read(data.uuid)
      .then(selectedLot => {
        vm.selectedLot = selectedLot;
        return Lots.dupes({ label : vm.selectedLot.label, inventory_uuid : vm.selectedLot.inventory_uuid });
     }).then(lots => {
        lots.forEach(lot2 => {
          lot2.selected = lot2.uuid === vm.selectedLot.uuid;
          lot2.merge = false;
        });
        vm.lots = lots;
    })
    .catch(Notify.handleError);
```

server/config/routes.js
```diff
@@ -950,6 +950,8 @@ exports.configure = function configure(app) {
   app.get('/lots/:uuid', lots.details);
   app.put('/lots/:uuid', lots.update);
   app.get('/lots/:uuid/assignments/:depot_uuid', lots.assignments);
+  app.post('/lots/:uuid/merge', lots.merge);
+  app.get('/lot/dupes/:label?/:entry_date?/:expiration_date?/:initial_quantity?/:inventory_uuid?', lots.getDupes);
```
jniles, line 954: Now that you are using `req.query`, this represents the query string component of the URL (`?x=y&w=z`).  That means you no longer need any of these extra parameters after `/lots/dupes`.  This are _parameters_ found in `req.params`, not _query string options_ found in `req.query`.

server/controllers/stock/lots.js
```diff
@@ -92,6 +94,76 @@ async function update(req, res, next) {
...
+ *
+ * TODO: After getting this working purge unneeded params
+ *
+ */
+function getDupes(req, res, next) {
+  const options = db.convert(req.query, ['inventory_uuid']);
+  const filters = new FilterParser(options, { tableAlias : 'l' });
+  filters.fullText('label');
+  filters.equals('inventory_uuid');
+  filters.equals('initial_quantity');
```
jniles, line 112: Looks like you have put `initial_quantity` in twice.  Go ahead and remove it.

server/controllers/stock/lots.js
```diff
@@ -92,6 +94,76 @@ async function update(req, res, next) {
...
+  try {
+    const transaction = db.transaction();
+    lotsToMerge.forEach(rawUuid => {
+      const mergeLotUuid = db.bid(rawUuid);
+      transaction.addQuery(updateLotTags, [uuid, mergeLotUuid]);
+      transaction.addQuery(updateStockAssign, [uuid, mergeLotUuid]);
+      transaction.addQuery(updateStockMovement, [uuid, mergeLotUuid]);
+      transaction.addQuery(deleteLot, [mergeLotUuid]);
+    });
+    await transaction.execute();
```
jniles, line 159: Beautiful 👍

test/integration/mergeLots.js
```diff
@@ -0,0 +1,286 @@
...
+    return mockStockMovements.reduce((chain, p) => {
+      return chain
+        .then(() => db.exec(`DELETE FROM stock_movement WHERE uuid=0x${p[0]};`));
+    }, Promise.resolve());
+  });
+
+  after('Delete temporary lots', () => {
+    return mockLots.reduce((chain, p) => {
+      return chain
+        .then(() => db.exec(`DELETE FROM lot WHERE uuid=0x${p[0]};`));
```
jniles: This works fine.  I have no issue with it as written.

However, since it doesn't matter the order in which you delete things, a more readable version might by to compose a single SQL statement using `WHERE uuid IN(?)`.  In fact, `db.exec()` will do this for you automagically.

```js
after('Delete temporary stock movements', () => {
  const lotUuids = mockLots.map(lot => db.bid(lot[0]));
  return db.exec(`DELETE FROM lot WHERE uuid IN (?);`, lotUuids);
});
```

client/src/modules/stock/lots/modals/duplicates.modal.html
jniles: A similar idea can probably be applied to the `checkboxes`.

client/src/modules/stock/lots/modals/duplicates.modal.html
```diff
@@ -0,0 +1,71 @@
...
+                checked="true"
+                required>
+              <input type="radio" name="selected" ng-if="lot.uuid !== $ctrl.selectedLot.uuid"
+                ng-click="$ctrl.selectLot(lot)"
+                required>
+            </td>
+            <td class="center">
+              <input type="checkbox" name="merge" ng-if="lot.uuid !== $ctrl.selectedLot.uuid"
+                ng-model="lot.merge">
+              <input type="checkbox" name="merge" ng-if="lot.uuid === $ctrl.selectedLot.uuid"
```
jniles: As opposed to my comment up there, let's hide this on the current line rather than disabling it since it doesn't make sense in the context of merging something into itself.   I would use an `ngIf` for this.

client/src/modules/stock/lots/modals/duplicates.modal.html
```diff
@@ -0,0 +1,71 @@
...
+                disabled="true"
+                ng-model="lot.merge">
+            </td>
+            <td>{{lot.label}}</td>
+            <td>{{lot.quantity}}</td>
```
jniles: For all numbers/prices/dates, let's put `class="text-right"` so that they are right-aligned.  This makes it easier to compare them.

client/src/modules/stock/lots/modals/duplicates.modal.html
jniles: Finally, let's add `style="cursor:pointer;"` to these radio buttons.  I'm surprised the browser does not default to it.  It will help our users know when they are over a "clickable area".

jniles (commented):
Thanks for tackling this @jmcameron.

I really like the green highlight.  I also think the user interface is really clean.  Well done!

I've tested it with the imck and vanga databases and there seem to be some columns missing:
image

Is there something I should be doing extra to get the dates/numbers to appear?

I've made a number of comments to improve the code.  If you need help on the translation work, let me know!

client/src/modules/stock/lots/modals/duplicates.modal.html
jmcameron: Just missed it; doing it now

client/src/modules/stock/lots/modals/duplicates.modal.js
```diff
@@ -0,0 +1,71 @@
...
+    const lotsToMerge = [];
+    vm.lots.forEach(lot => {
+      if (lot.merge) {
+        lotsToMerge.push(lot.uuid);
+      }
+    });
+    if (lotsToMerge.length === 0) {
+      Notify.warn($translate.instant('LOTS.NO_LOTS_MERGED'));
+    } else {
+      Lots.merge(vm.selectedLot.uuid, lotsToMerge);
```
jniles: In order to ensure the registry reload after the merge, you'll want to wait until after this returns to close the instance.

So, you'll need to call `Instance.close()` after `Lots.merge()`.  Maybe something like:

```js
Lots.merge(/* params  */).then(() => Instance.close());
```

client/src/modules/stock/lots/modals/duplicates.modal.html
jmcameron: Done. Reviewers need to check the new French items.

client/src/modules/stock/lots/modals/duplicates.modal.html
jmcameron: Done

client/src/modules/stock/lots/modals/duplicates.modal.html
jmcameron: Figured out ngDisabled. Cleaned all this up.

server/controllers/stock/lots.js
jmcameron: Ooops. Fixed.

client/src/modules/stock/lots/modals/duplicates.modal.html
jmcameron: Fixed this.

jmcameron: Need to look at this again.

client/src/modules/stock/lots/modals/duplicates.modal.html
jmcameron: Done

client/src/modules/stock/lots/modals/duplicates.modal.js
jmcameron: Done and verified that it works.

test/integration/mergeLots.js
jmcameron: Will do this next.

client/src/modules/stock/lots/modals/duplicates.modal.html
jmcameron: Done.

test/integration/mergeLots.js
jmcameron: Done. Thanks for the suggestion!

client/src/modules/stock/lots/modals/duplicates.modal.html
jmcameron: Done

jniles (approved):
I approve this PR.  Great job @jmcameron!

Can you do a rebase, squashing all these commits?  Then you can do let bors loose.

jniles:
bors delegate+

PR #5308, opened 2021-01-24
test: add missing unit tests for bhSubmit directive

Buzunda:
Add missing unit tests for bhSubmit directive and changes the bhSubmit directive to better handle when the callback promise has a rejection.

### DIFF
--- a/client/src/js/directives/bhSubmit.js
+++ b/client/src/js/directives/bhSubmit.js
@@ -42,9 +42,9 @@ function bhSubmitDirective() {
         // the response is a promise, toggle the loading state on
         // fulfillment/rejection
         if (isPromise(response)) {
-          response.finally(function () {
-            FormController.$toggleLoading();
-          });
+          response
+            .then(() => FormController.$toggleLoading())
+            .catch(() => FormController.$toggleLoading());
 
         // otherwise, toggle the loading state off right away.
         } else {
--- a/test/client-unit/directives/bhSubmit.spec.js
+++ b/test/client-unit/directives/bhSubmit.spec.js
@@ -1,0 +1,89 @@
+/* global inject, expect, chai */
+describe('(directive) bhSubmit', () => {
+  let $scope;
+  let element;
+  let deferred;
+
+  beforeEach(module('bhima.directives'));
+  beforeEach(inject(($q, $rootScope) => {
+    $scope = $rootScope;
+    deferred = $q.defer();
+
+    $scope.models = {
+      formValue : null,
+      submit : chai.spy(() => deferred.promise),
+    };
+  }));
+
+  describe('when the directive is added to a form', () => {
+    beforeEach(inject(($compile) => {
+      element = angular.element(`
+        <form name="form" bh-submit="models.submit(form)" novalidate>
+          <input ng-model="models.formValue" name="formValue">
+        </form>
+      `);
+
+      $compile(element)($scope);
+      $scope.$digest();
+    }));
+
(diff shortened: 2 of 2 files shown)


client/src/js/directives/bhSubmit.js
```diff
@@ -42,9 +42,9 @@ function bhSubmitDirective() {
         // the response is a promise, toggle the loading state on
         // fulfillment/rejection
         if (isPromise(response)) {
-          response.finally(function () {
-            FormController.$toggleLoading();
-          });
+          response
```
Buzunda, line 45: Because the error is not being catched anywhere if I try to reject the promise on the uni test I get the following error:

Buzunda, line 45: Because the error is not being catched anywhere if I try to reject the promise on the unit test I get the following error:

Screen Shot 2021-01-24 at 08 50 59

Buzunda, line 45: Because the error wast not being catched anywhere, if I try to reject the promise on the unit test I was getting the following error:

Screen Shot 2021-01-24 at 14 19 30

test/client-unit/directives/bhSubmit.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      formValue : null,
+      submit : chai.spy(() => deferred.promise),
+    };
+  }));
+
+  describe('when the directive is added to a form', () => {
+    beforeEach(inject(($compile) => {
+      element = angular.element(`
+        <form name="form" bh-submit="models.submit(form)" novalidate>
+          <inpu id="row" ng-model="models.formValue" name="formValue">
```
jniles: Looks like there is a typo here - `<inpu` -> `<input`.

client/src/js/directives/bhSubmit.js
jniles: Hi @Buzunda, you are right - this code was not very testable as written.  This is a good solution.

Could you wrap these in a closure?  Basically, instead of `.then(FormController.$toggleLoading)`, write `.then(() => FormController.$toggleLoading())`.  This just makes sure the callback is called in isolation.

test/client-unit/directives/bhSubmit.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      $scope.$digest();
+    }));
+
+    it('sets $loading to false on the FormControl', () => {
+      expect($scope.form.$loading).to.equal(false);
+    });
+
+    describe('when the form is submitted', () => {
+      beforeEach(() => {
+        angular.element(element).trigger('submit');
```
jniles, line 36: Nice 👍   This is always tricky to figure out.  Well done!

test/client-unit/directives/bhSubmit.spec.js
```diff
@@ -0,0 +1,89 @@
...
+        });
+
+        it('sets $loading to false on the FormControl', () => {
+          expect($scope.form.$loading).to.equal(false);
+        });
+      });
+    });
+  });
+
+  describe('when the directive is not added to a form', () => {
```
jniles, line 71: Interesting test!  I hadn't thought of this.  👍

jniles (commented):
These tests are awesome!  Testing AngularJS with Karma is one of the trickiest testing environments I've worked with and these are really impressive.

The only blocking comment is the typo in `<input />`.  The rest are suggestions to make the code a bit more sound.

jniles (approved):
Excellent.  All changes addressed.  Thanks!

bors r+

PR #5312, opened 2021-01-25
Update karma to latest version

jniles:
Closes #5305.

Removes the unknown "concurrency" option.

### DIFF
--- a/sh/test.sh
+++ b/sh/test.sh
@@ -26,7 +26,7 @@ SUITE=${SUITE:-"ALL"}
 # run karma (client unit) tests
 if [ $SUITE = "client-unit" ] || [ $SUITE = "ALL" ] ; then
   startfold "Running Client Unit Tests..." "test-client-unit";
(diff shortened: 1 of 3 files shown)


sh/test.sh
```diff
@@ -26,7 +26,7 @@ SUITE=${SUITE:-"ALL"}
 # run karma (client unit) tests
 if [ $SUITE = "client-unit" ] || [ $SUITE = "ALL" ] ; then
   startfold "Running Client Unit Tests..." "test-client-unit";
-  ./node_modules/.bin/karma start --single-run --no-auto-watch --concurrency 1 karma.conf.js
+  ./node_modules/.bin/karma start --single-run --no-auto-watch karma.conf.js
```
jmcameron, line 29: I see from the Karma 6.0 documentation this configuration option is still supported -- although not through the CLI.  And that the value for 'concurrency' is set to 2 in the Bhima Karma.js config file.

jmcameron (approved):
Looks good.

PR #5316, opened 2021-01-26
feat(stock): highlight expired lots in adjustment

### LINKED ISSUE #5187: Signal to the user that products are expired in the "inventory adjustment" page
Let's put a little icon on the "inventory adjustment" page for stocks that have expired.  Typically, you should set them to 0 during the adjustment.  See below the items that are expired:

image

It would be nice to put an icon or a color to indicate to the user that these items should be examined with care.

If we _really_ wanted to force the issue, we could automatically set them to 0.

jniles:
Highlights the expired lots in the inventory adjustment module.

This is what it looks like:
image

Closes #5187.


jmcameron (approved):
The code looks fine.  I tested it on the IMCK db and bhima_test and the both worked and looked nice.  

However, there seems to be a glitch with the translation.  Note that the untranslated item:
image

jniles:
Fixed!

image

jmcameron (approved):
Retested and the translation tokens now work.  Looks goo.d

PR #5317, opened 2021-01-27
feat(stock): add option to view AMC calcs

### LINKED ISSUE #5291: Add option to view CMM calculation details to dropdown in "Articles in Stock"
We should add an option to the "Articles in Stock" registry to view the details of the CMM calculation.  It would be call "View Calculation Details" or something similar.  Here is where it would go:

image

Clicking this would bring up a large modal with the following information:

  1. The name of the inventory item (e.g. Quinine 500mg)
  2. The depot the CMM calculation is relevant for (e.g. Pharmacie PAX)
  3. The time period on which the calculation is relevant.  This would be the current date minus the months specified in the depot settings.  It would be in the format "Time Period: `${start_date} - ${end_date}`". Note that this is **not** the period of time we have data for.  It should be the full date range derived from the depot settings.
  4. The total number of items consumed that time.  (e.g. 50,000 comprimes)
  5. The total number of days where there were stock outs during that period. (e.g 7 jours de rupture de stock)
  6. The total number of days where there was stock available (.e.g 156 jours de stock)
  7. The full, written out CMM calculation, based on the algorithm chosen.  For example, for the MSH algorithm, we would have:
  
CMM = (∑consumed_quantity / ( #month − (∑stock_out_days) / 30.5)))
CMM = (50000 / (6 − (7 / 30.5)))
CMM = 8664.773

This would allow the user to have full transparency into the way the calculation was taking place.

jniles:
Adds the ability to view the AMC/CMM calculations in the Articles in Stock Registry.  A summary of the information is presented for the user to examine and the explicit formula used to calculate the AMC is shown for the user to validate.

Closes #5291.

**Example using Algorithm 1**
image

**Example using Algorithm 2**
image

**Example using Algorithm 3**
image

**Example using Algorithm 4 (MSH)**
image

**Example showing Algorithm (MSH) for 12 months**
image


jniles:
@jmcameron, this code isn't ready to be merged (I need to check translations and such), but could you give me a review on the user interface and user experience?   I've been looking at this too long to know if my ideas are any good or not.

It should "just work" with the latest IMCK/Vanga databases for testing.

jmcameron (approved):
I looked over the code and it looked good.
I tested this with the IMCK database it looks great.   I did spot-check several of the CMM calculations to make sure the numbers worked.    Everything looks great.

One small/cosmetic/ignorable suggestion:  Make the values for the CMM column to links to the same function the action menu invokes.  It could include a tooltip:  "Click here to see the details of this calculation.".  Ignore this if you think this would slow things down too much.

jniles:
@jmcameron thanks for your review!  Let me take a look at figuring out how to do that linking.

I'll ping you for a final review when that is done.

client/src/modules/stock/inventories/modals/amc.modal.js
```diff
@@ -0,0 +1,31 @@
...
+      vm.inventory = items.inventory;
+      vm.depot = items.depot;
+
+      vm.data.avg_consumption = vm.data[vm.settings.average_consumption_algo];
+
+      // nicer aliases to use in the HTML
+      vm.isAlgo1 = vm.settings.average_consumption_algo === 'algo1';
+      vm.isAlgo2 = vm.settings.average_consumption_algo === 'algo2';
+      vm.isAlgo3 = vm.settings.average_consumption_algo === 'algo3';
+      vm.isAlgo4 = vm.settings.average_consumption_algo === 'algo_msh';
```
mbayopanda, line 40: 👍🏽 Clean and nice

mbayopanda (commented):
Good job @jniles 

I have noticed that : 

- The calculation of the Algo 3 doesn't match with what is displayed as an example **30.5 × (200 / 1) <> 32.97**: 
image

- The calculation of Algo 2 doesn't match with the example **30.5 × (200 / 185) <> 6100**
image

It seems like the example for Algo 2 is given to Algo 3, and vice-versa

And the french translation

jniles:
Good catch @mbayopanda!  I'll switch those algorithm examples.

jmcameron (changes requested):
I could not get the double-clicking on the CMM value to work.  I suggest making it a link with tooltip, if you have time.

I went back and did some more spot-checks (which I should have done the first time!).   The calculations for Alg 1 and 4 seem fine.  But I'm seeing errors on 2 and 3.  See below.

image

image

jniles:
Okay, this confusion that @jmcameron pointed out and that @mbayopanda pointed out is because we change the names of our algorithms somewhere between the concept note, the translation, and the calculation.  I'll try to track it all down.

jniles:
@jmcameron can you give me another review?

I've integrated the french language translations and attempted to polish up the double-click experience.  I believe this should solve the issues you were having with double-clicking.

I would prefer not to put a label/title/tooltip on it for performance reasons.  I also don't think it enhances the experience that much after testing it myself, and may hurt a users ability copy/paste as it gets in the way.  If you feel strongly about it, I can put it in.

For completeness, here is what the current version looks like:

image


----

I've noticed that in Vanga's database, the min/max isn't being properly computed, making everything either "in stock" or "over maximum".  I'm taking a look at that.

jniles:
Actually, I just realized the screen shot I posted doesn't make any sense.  How can we have more days of stock consumption than we do of stock available?

Hmm....  I hope our CMM calculation is working.

jmcameron (approved):
I tried the CMM double-click and it works fine now.  
Your decision to not add a tool-tip makes sense.
Let me know if you want me to review it again after straightening out the CMM calculations.

BTW, on the last screen-shot you have of the CMM calculations for Algo 4:  The fact that algorithms 1 and 2 have much bigger numbers than 3 and 4 is concerning too.

jniles:
@jmcameron, thanks for your review.  I've updated the base branch so that we will merge this off the main branch.

PR #5334, opened 2021-02-01
feat(stock): default filter on depot for movements

### LINKED ISSUE #5327: Linking from the Articles in Stock registry to the Stock Movements registry should filter by depot by default
In the "action" menu of the "articles in stock" registry, when you click to view stock movements it shows you all stock movements for that inventory item, even though the data you may want to follow up on (CMM, quantity in stock) is partitioned by the depot you are currently in.

In my view, this should not be the default.  It's much easier to remove the depot filter than to add it, and in general, my follow-ups are to try and verify the information in the Articles in Stock registry.  Let's change this behavior to filtering by the depot by default as well.

jniles:
Adds a default filter on the link from the Articles in Stock registry to the stock movements registry to filter by the depot.

Closes #5327.

### DIFF
--- a/client/src/modules/stock/inventories/templates/inventory.action.html
+++ b/client/src/modules/stock/inventories/templates/inventory.action.html
@@ -10,7 +10,8 @@
       <a data-method="stock-movement" ui-sref="stockMovements({
         filters : [
           { key : 'period', value : 'allTime' },
-          { key : 'inventory_uuid', value : row.entity.inventory_uuid, displayValue : row.entity.text }
+          { key : 'inventory_uuid', value : row.entity.inventory_uuid, displayValue : row.entity.text },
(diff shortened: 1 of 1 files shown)


jmcameron (approved):
Code looks good.  I checked the code in the action menu HTML and verified that depot_uuid was there.  So somehow adding this item to the filters was all that was needed to fix this issue.

I tested with master and was able to duplicate the problem in the original issue. 
I tried this PR and it works properly.

Nice job!

jmcameron:
@jniles:  Note that the codacity test is failing...

jniles:
Yeah, codacy is annoyed that I'm not pulling into the main branch.  While I got it to be okay with future PRs (like yours), but getting it to rerun these tests seems like it isn't a thing.

With your LGTM, I'll merge it anyway.  If I'm wrong, I'll figure out a fix!

bors r+

PR #5335, opened 2021-02-01
Stock wording fixes

### LINKED ISSUE #5319: Make Stock dashboard "no inventory found" messages positive
Here is what the dashboard "no inventory found" message looks like when it doesn't find any expired products:

image

This should be _much_ more positive!  We should make this into a green "success" message, potentially with a thumbs up icon.  This will give users a clear visual indicator that they are doing well.

### LINKED ISSUE #5322: The "articles in stock" report should say "include items with no stock" rather than "lots"
The Articles in Stock report shows items in stock, their CMM, and relevant information.  However, one of the options reads "Include Empty Lots" which is misleading, since the report does not show lots!  We should rename this option to "Include items not in stock".

image

### LINKED ISSUE #5326: Unify the names of "Articles in Stock" and "Items in Stock"
In the English translation, we use "articles in stock" and "items in stock" interchangeably to mean any kind of inventory item that has stock.  We should choose one as it is confusing to have both.  I prefer "Articles in Stock" since "items" is pretty generic.  However, my opinion could be swayed.

@jmcameron what do you think?  Which would you pick?

jmcameron:
dThis PR fixes several cosmetic/wording problems in various stock pages and reports.  Specifically:

- Issue https://github.com/IMA-WorldHealth/bhima/issues/5319  ('no inventory found' wording on Stock Dashboard)
- Issue https://github.com/IMA-WorldHealth/bhima/issues/5322  (fix wording in Stock Exit Report)
- Issue https://github.com/IMA-WorldHealth/bhima/issues/5326  (stock 'items' -> 'articles')

Attached is what some of these indicators should look like on the Stock Dashboard (English):
image

**TESTING / REVIEWING:** 
- To see the success indicators on the Stock Dashboard, use the bhima_test database and edit lots and adjust inventory to eliminate all the expired products, at risk, etc.  In the picture, I did not try to eliminate all products over maximum, but I'm confident it will work.
- Please carefully review the translation updates, Especially the French ones, to make sure they make sense and use good French.  
- Note that the French terms on the Stock Dashboard are not consistent:  Some use "Products" and some use "Stock".   I'm open to suggestions on how to improve that.

-----
[View rendered docs/en/stock-management/inventory.md](https://github.com/jmcameron/bhima/blob/stock_wording_fixes/docs/en/stock-management/inventory.md)
[View rendered docs/en/stock-management/overview.md](https://github.com/jmcameron/bhima/blob/stock_wording_fixes/docs/en/stock-management/overview.md)


client/src/i18n/en/inventory.json
```diff
@@ -41,8 +41,8 @@
         "INVENTORY_PRICE_INFO" : "inventory_price : the sale price of the inventory in the enterprise currency",
         "COLUMNS_INFO" : "Note about some columns",
         "SELLABLE" : "Sellable",
-        "VIEW_ARTICLES_IN_STOCK" : "View Items in Stock",
-        "VIEW_ITEM_IN_INVENTORY" : "View Item in Inventory",
+        "VIEW_ARTICLES_IN_STOCK" : "View Articles in Stock",
+        "VIEW_ITEM_IN_INVENTORY" : "View Article in Inventory",
```
jniles: Can you even update this translation key?  "VIEW_ITEM" -> "VIEW_ARTICLE"?

client/src/i18n/en/report.json
```diff
@@ -176,7 +176,7 @@
     "VIEW_ACCOUNT_STATEMENT": "View in Account Statement",
     "VIEW_CREDIT_NOTE": "View Credit Note",
     "VIEW_INVOICE": "View Invoices",
-    "VIEW_ITEM_STOCK": "View item in stock",
+    "VIEW_ITEM_STOCK": "View article in stock",
```
jniles: Same comment.

client/src/i18n/fr/stock.json
```diff
@@ -74,9 +75,9 @@
     "FROM"            : "Provenance",
     "GROUPING"        : "Groupage",
     "JOIN_INVOICE"    : "Joindre une facture",
-    "INCLUDE_EMPTY_LOTS" : "Inclure les lots épuisés",
+    "INCLUDE_EMPTY_LOTS" : "Inclure les articles non en stock",
```
jniles: I believe this is is referring to https://github.com/IMA-WorldHealth/bhima/issues/5322 which it would address.  However, I'm not sure that it isn't used elsewhere to actually mean lots.

If you are fixing https://github.com/IMA-WorldHealth/bhima/issues/5322, could you create a new key for articles that are out of stock and use that?

client/src/modules/stock/dashboard/dashboard.html
```diff
@@ -56,7 +56,7 @@
     <div class="row">
       <div class="col-md-4">
         <bh-stock-panel-minimum-reached
-          label="STOCK.STATUS.MINIMUM">
+          label="STOCK.DASHBOARD.LOW_STOCK">
```
jniles, line 59: Hey!  Nice! :+1:

We already have LOW_STOCK translated in `STOCK.STATUS.MINIMUM` ([here](https://github.com/IMA-WorldHealth/bhima/blob/00bd25f303b883582881d84df65bca613213f3dc/client/src/i18n/en/stock.json#L182)).  Since this dashboard is basically just looking for stock lots that have a particular flag, it might make sense to just use that translation key here.

I would propose holding back on this change for another PR where we can tackle all references to the stock status flags together.  I'll make an issue on it.

client/src/modules/templates/bhStockPanelMinimumReached.tmpl.html
```diff
@@ -11,7 +11,7 @@
         </span>
       </div>
 
-      <div class="text-center" ng-if="!$ctrl.loading && $ctrl.stockNotFound">
+      <div class="text-center bg-success text-success" ng-if="!$ctrl.loading && $ctrl.stockNotFound">
```
jniles, line 14: 👍

jniles (commented):
The English i18n looks good to me.  I made a suggestion about changing the key names.

client/src/i18n/en/inventory.json
jmcameron: Done

client/src/i18n/en/report.json
jmcameron: Done

client/src/i18n/fr/stock.json
jmcameron: Done.  Although there is an issue here that should be fixed in another PR.  When I made the bhStockFilterExhausted component (see the Default filters in Stock Lots), There is a (probably left-over) reference to INCLUDE_EMPTY_LOTS_HELP which does not seem to be working.   See the end bhStockFilterExhausted.tmp.html.  If you concur that it is no longer needed (which is what I thin), I'll delete it and rebase this PR before releasing.

client/src/modules/stock/dashboard/dashboard.html
jmcameron: I did this deliberately because the "MINIMUM" one was used as a column header, but I wanted something less terse and matching the style of the other Dashboard panes.  So I do not plan to change this unless you really think it is necessary.

jniles (approved):
Looks good to me!

bors r+

client/src/i18n/fr/stock.json
jmcameron: > Although there is an issue here that should be fixed in another PR. When I made the bhStockFilterExhausted component (see the Default filters in Stock Lots), There is a (probably left-over) reference to INCLUDE_EMPTY_LOTS_HELP which does not seem to be working. See the end bhStockFilterExhausted.tmp.html. If you concur that it is no longer needed (which is what I thin), I'll delete it and rebase this PR before releasing.

I'll deal with this is a separate PR.

jniles:
Perfect :+1:

PR #5341, opened 2021-02-02
Refactor: Stock Management

### LINKED ISSUE #4304: The "cost" column in stock movements registry should have a sum at the bottom
It would be really nice to have the sum in the base of the "cost" column of the stock movement registry.  This will let administrators make really quick analyses on the fly.  See below:

Test

### LINKED ISSUE #5073: Remove the stock_consumption table and associated procedures
We don't have many references to the stock consumption table, and we don't really use it.  It has been deprecated in favor of stock_movement_status.  Let's remove it and all associated procedures.

### LINKED ISSUE #5144: Remove the inventory.avg_consumption column
We don't actually need the `inventory.avg_consumption` column anymore since we don't store the average consumption in there anymore.  Instead we calculate it.  I think we should remove it, but maybe I'm missing a good reason to keep it around.

@mbayopanda @lomamech what do you think?

### LINKED ISSUE #5183: Lots in Stock Registry key should say "expiré" instead of "expiration"
The color key at the bottom of the "Lots in Stock" registry should say "expiré" instead of "expiration".  These are lots that are already expired.

image

### LINKED ISSUE #5185: Lifetime should be calculated in days, not months
If a lot is both at risk of expiration and expired, the coloration is at risk of expiration.  This is because the lifetime is calculated in months, which makes the color unsure if it is expired or at risk of expiration.

Somewhat related to: #5008

See below:

image

These lots are already expired, yet they show up as "at risk".  I worry it may have something to do with the fact that we calculate `is_expired` by comparing dates:

https://github.com/IMA-WorldHealth/bhima/blob/a76216b572a71bd7aa75d05bd5bdf6dad8b8eaf9/server/controllers/stock/core.js#L117-L120

... yet we still calculate `IS_IN_RISK_EXPIRATION` in months:

https://github.com/IMA-WorldHealth/bhima/blob/a76216b572a71bd7aa75d05bd5bdf6dad8b8eaf9/server/controllers/stock/core.js#L734-L736

### LINKED ISSUE #5186: "Entry date" and "Period" (in filters by default) do the same thing in Stock Lots Search Modal
We have multiple options for searching for dates in the Lots Search modal.  Not all of them make sense.  It appears that the "Period" is the same thing as the "Entry Date".  I propose we change the text on the "Period" to be "Entry Period" or "Entry Date", default that to "allTime" and remove the "Entry Date" search.

image

image

### LINKED ISSUE #5291: Add option to view CMM calculation details to dropdown in "Articles in Stock"

### LINKED ISSUE #5311: Remove CMM related columns from the stock lots registry
In order to combat the problem of confusing information on the Stock Lots Registry (see @mbayopanda's comments on https://github.com/IMA-WorldHealth/bhima/pull/5226) and having registries too large (see https://github.com/IMA-WorldHealth/bhima/issues/5300), I propose we remove some of the columns from the stock lots registry.

Here is the registry that exists in #5226:
https://user-images.githubusercontent.com/896472/104612221-40efdd00-5686-11eb-89e2-e22a9854ec44.png

Let's remove all columns that do not directly relate to the _lot_ itself.  Concretely, those are:

1. CMM
2. MSD
3. Lifetime
4. Lifetime Lot
5. Risk
6. Risk Quantity

If we remove these columns, we'll be left with only columns that are readily intelligible to the user:

1. Quantity
2. Unit Cost
3. Expiration Date

We could even add "Total Cost" if we wanted which would be quantity * unit_cost.

-------

Removing these columns would require us to implement #5292 to provide better information to the user about how the expiration risk is being calculated.

### LINKED ISSUE #5319: Make Stock dashboard "no inventory found" messages positive

### LINKED ISSUE #5322: The "articles in stock" report should say "include items with no stock" rather than "lots"

### LINKED ISSUE #5325: "Exclude inventories with zero stock" is not cached in the Stock Value Report
A GIF is worth a thousand words:

mtTZw4LmjJ

The "exclude inventories with zero stock" option is not cached/saved between submit operations.  This is a different user experience from other reports.  We should be caching it between submit operations.

### LINKED ISSUE #5326: Unify the names of "Articles in Stock" and "Items in Stock"

### LINKED ISSUE #5327: Linking from the Articles in Stock registry to the Stock Movements registry should filter by depot by default

### LINKED ISSUE #5328: Add a link to the "Fiche de Stock" from the Articles in Stock registry
We should add a link between the Fiche de Stock report (Stock Sheet) and the Articles in Stock registry.  Because we don't have the details of the lot movements in the Stock Movements registry, we need some way of verifying the totals for stock movements.  This is found in the Stock Sheet report.  Let's add a link to it from the Articles in Stock Registry.

### LINKED ISSUE #5332: Audit the procedures that create CMM data
As mentioned in https://github.com/IMA-WorldHealth/bhima/pull/5317, it appears that we have some bugs in our CMM calculation, at least with respect to one of our production sites.  We should do an audit to see if we can make sure that the calculations are both correct and efficient.

### LINKED ISSUE #5333: Link the Purchase Order Analysis report on the Purchase Order Actions menu
We provide an easy way to analyze a purchase order through the Purchase Order Analysis report.  Let's add a link directly to the Purchase Order registry.  Here is where it would go:

image

### LINKED ISSUE #5349: Align the columns of the Stock Expiration Report
The stock expiration report shows the stock that is expired or at risk of expiration, grouped by depot.  The columns do not line up between the different tables:
image

This is a bit annoying when trying to quickly scan the tables.  Let's make these columns line up.

### LINKED ISSUE #5354: Add default labels to bhDateInterval entry fields
When the date interval select entry fields appear on a form, the left one corresponds to the start date and the right one to the end date.  However, before dates are selected, the fields appear as empty boxes and it is not immediately clear what they correspond to.  I suggest that we insert inactive "Start Date" and "End Date" labels in the boxes (respectively) when date values have not been selected.  This text will be semi-grayed out and will clearly indicate the purpose of each box.  This should be done in the bhDateInterval component so that automatically appears wherever it is used. 

Question: Is there any reason to support optional overrides for these two labels?  If so, the this change could be done with an optional binding which would default to "Start Date" and "End Date".

Also, I have not examined the bhDateInterval component to see if they already support this. 

NOTE: Examples of this appear as the 'Entry Date' and 'Expiration Date' in the "Stock Lots" search dialog.

### LINKED ISSUE #5360: Store type of movements in stock_movement_status
Currently, our `stock_movement_status` table looks like this:

| uuid                               | start_date | end_date  | quantity | in_quantity | out_quantity | inventory_uuid  | depot_uuid                         |
|:--- | --- |--- | ---:| ---:| ---:| --- | --- |
| 0x26629E1F53E811EB9D86F44D306E0330 | 2019-10-15 | 2019-10-15 | 2738.0000 |   2738.0000 |       0.0000 | 0x10B50BC37BEF45BA8162353130EB5F43 | 0x05121CFE23CA4E8780484F231F5DA590 |
| 0x0EA69B293E1C11EBB2D1F44D306E0330 | 2019-10-15 | 2019-10-15 |  221.0000 |    221.0000 |       0.0000 | 0x3789C631DB714D02AF3A9D2F745AD20B | 0x05121CFE23CA4E8780484F231F5DA590 |
| 0x2640F05F53E811EB9D86F44D306E0330 | 2019-10-15 | 2019-10-15 |   37.0000 |     37.0000 |       0.0000 | 0x3732BD081F24429D8AC3C81D834F220E | 0x05121CFE23CA4E8780484F231F5DA590 |
| 0x2616137B53E811EB9D86F44D306E0330 | 2019-10-15 | 2019-10-15 |  224.0000 |    224.0000 |       0.0000 | 0x259C5056AF304F968751F731351FD4AC | 0x05121CFE23CA4E8780484F231F5DA590 |
| 0x262C07C053E811EB9D86F44D306E0330 | 2019-10-15 | 2019-10-15 |  415.0000 |    415.0000 |       0.0000 | 0x716774C2826141F1B43A93C716C69154 | 0x05121CFE23CA4E8780484F231F5DA590 |

One issue is that we have to do extra work to figure out what _kind_ of movement these movements represent.  Not all exists are consumption events, and they depend on the type of movement and the depot settings.  

To ease our calculations, let's encode these directly into the `stock_movement_status` table.  Let's add the following columns:

1. `out_quantity_consumption` - the quantity exiting the depot via a _consumption_ event.
2. `out_quantity_exit` - the quantity exiting the depot via a normal exit.

These would replace the `out_quantity` amount, since you could recover that simply by summing.

It might also be useful to have the `start_quantity` and `end_quantity` just to be explicit.  Right now, it is unclear if the `quantity` refers to be or after the movement.

### LINKED ISSUE #5416: Enhancement to "Rapport graphique des mouvements de stock"
1. Typo:
    - Replace `Stock Movement Report` by `Rapport graphique des mouvements de stock` in breadcrumb of the report
    - Add a description of the report in the report config panel
2. Extend the date range by letting the user specify the start and the end of date range instead of just selecting data for a given month
3. Since movements are categorized by flux type, instead of a simple bar chart, it can be better to have multiple bar charts where each bar is for a given flux type. And add a legend for these bars, this feature can be an option too.
4. Stock movements are related to a given depot, but in this report, there is no distinction between depots

### LINKED ISSUE #5422: Improvements to the "Articles in Stock" registry
This issue is remarks made while working on https://github.com/IMA-WorldHealth/bhima/pull/5341.

### Bug: The "at risk of expiration" flag doesn't work
It appears that the "risky lots" check detects lots that are not actually at risk of expiry.  Here is an example:
image

And here is the CMM calculation:
image

And finally, the expiration date of these lots:
image

You'll notice that we have plenty of time to consume this medication, and will consume this medication at current CMM rates before they expire.  So this flag is not correct.

_This is a blocking issue and should be fixed before merging #5341._ 

### Proposal: add a minimum delay to the CMM calculations

This is already documented in https://github.com/IMA-WorldHealth/bhima/issues/5407.

### LINKED ISSUE #5430: Lot Registry : Edit lot doesn't show the inventory name
When we click on Edit lot in the registry we don't see the inventory name in the modal 

image

### LINKED ISSUE #5431: Stock : avg_consumption is still used in most of stock modules and registries
The `avg_consumption` column in the inventory table has been removed in the `refactor-stock` branch but in `master`, there are modules and registries which still use this column.

### LINKED ISSUE #5444: Stock movements without descriptions on some lines of their transactions
At bandunduville, we are running into a few stock movements that are missing descriptions on one line of their journal transaction.  This causes the rule that descriptions can't be NULL to fail.  Here is an example:
image

They appear to all be negative adjustments:
image

Clearly there is something broken.  It looks like the negative adjustments module at Bandunduville isn't blocking on an empty description and that is what is causing the issue.

### LINKED ISSUE #5447: Rename `at_risk` flag to `at_risk_of_stock_out`
As mentioned in https://github.com/IMA-WorldHealth/bhima/pull/5438#pullrequestreview-601581750, we need to rename the `at_risk` flag to `at_risk_of_stock_out` on the `refactor-stock-management` branch.  Yes, it is more wordy.  But it is also much clearer.

One issue we have on our stock routes is that sometimes routes with "lots" return "inventory articles", not lots.  Further, to calculate the risk of expiration of an inventory item, you need to know the risk of expiration of each lot, and the order in which they will be consumed (see https://github.com/IMA-WorldHealth/bhima/issues/5292).  So we often can't tell if we are dealing with lots or inventory items from the data returned.  This means we need to have really clear naming of our properties so that we avoid confusion in the future.

jniles:
This is the off-main-branch work that @jmcameron and @jniles are working on to refactor stock management.  It shouldn't be merged until we have finished all the major stock management upgrades and can test it as a whole.

The goal of this branch is to do a major upgrade to our stock management modules that may have far reaching implications and are integrated too tightly to be able to do progressively on `master`.  Basically, we'll relax our requirements to allow ourselves to "break things" temporarily on this branch, then have a single large merge into the main branch.

### Issues Closed by the PR

Closes #4304.    
Closes #5073.    
Closes #5144.    
Closes #5183.    
Closes #5185.    
Closes #5186.    
Closes #5291.    
Closes #5311.    
Closes #5319.    
Closes #5322.    
Closes #5325.    
Closes #5326.    
Closes #5327.    
Closes #5328.    
Closes #5332.    
Closes #5333.    
Closes #5349.    
Closes #5354.    
Closes #5360.    
Closes #5416.    
Closes #5422.    
Closes #5430.    
Closes #5431.    
Closes #5444.    
Closes #5447. 

 ### Things to Test

Report/Module | Assigned to | Issues Found | Works on master? (Y/N) | Fix Here? (Y/N) | Fixed? (Y/N)
-- | -- | -- | -- | -- | --
Rapports >   Articles en Stock | @mbayopanda | Status, CMM, etc broken | N  | Y  |  
Rapports >   Comparaison des articles factures sur les recus |  @jniles |  #5414 | Y | N |  
Rapports >   Consommations Menseulles | @jmcameron   |  SQL Error  | Y  | Y  |   Y (https://github.com/IMA-WorldHealth/bhima/pull/5436)
Rapports >   Entrees des stocks | @mbayopanda   |   | Y  | N  |  
Rapports >   Fiche de stock |  @jniles   | #5415  |  Y | Y  | Y (#5415)
Rapports >   Rapport d'expiration | @jmcameron   |   |  Y  | Y  |  Y (#5438)
Rapports >   Rapport graphique des mouvements |  @mbayopanda   |  #5416  | Y | N  |  
Rapports >   Sorties des stocks | @jniles  | #5419 | Y  |  N |  
Rapports >   Valeur de stock | @jmcameron   | OK  | Y | N  |  
Rapports >   Rapport graphique de la consommation de stock | @mbayopanda   | #5418  | Y  | N  |  
Rapports >   Rapport sur les variations de stock | @jniles  |  OK |  Y  | N  |  
Stock >   Ajustement |  @jmcameron  | OK  | Y | N |  
Stock >   Ajustement des stocks | @mbayopanda   | #5428  | Y | N  | N | 
Stock >   Articles en stock |  @jniles  | #5422  |  Y  | Y  | Y( #5432)
Stock >   Assignation de stock | @jmcameron   | Y | Y | N |  
Stock >   Entrée de stock | @mbayopanda   | #5429 is related to purchase analysis  | Y | N  | 
Stock >   Gestion de depots | @jniles  | #5423  | Y |  N |  
Stock >   Importation de stock |  @jmcameron | N (wrong num args) | N #5426 |  Y |  Y (https://github.com/IMA-WorldHealth/bhima/pull/5427)
Stock >   Lots en stock |  @mbayopanda | #5430, #5431  | Y, Y  | N, Y  | 
Stock >   Movements de stock | @jniles  | #5424  | Y  | Y |   Y(#5459)
Stock >   Parametres de stock | @jmcameron   | Seems OK  | Y | N |  
Stock >   Requisition | @mbayopanda  | Seems OK   | Y  | N  |  
Stock >   Sortie de stock | @jniles   | #5425 |  N | N  |  
Stock >   Tableau de bord | @jmcameron   |  https://github.com/IMA-WorldHealth/bhima/issues/5400#issuecomment-785245058  | Y | N |  

I've left @lomamech off this list because I believe he is working on a separate PR.  If he gets done with that, we'll move some of these over to him.

@mbayopanda can you review the titles of these modules?  I think the French may be wrong for some of them.  For example, I don't understand why we have "Ajustement des stocks" and "value de stock".  Shouldn't we use "des stocks" or "de stock" in every one?


jniles:
Okay, I've finally finished running through my modules.  There are only two blocking issues for this PR:

1. https://github.com/IMA-WorldHealth/bhima/issues/5415
2. https://github.com/IMA-WorldHealth/bhima/issues/5422

I'll reproduce the text here to make it easier to track.

---

### Bug: Date range should provide an opening balance (@jniles is working on this bug)

If you choose to set a date range for the Fiche de Stock, you limit the output of the report to a certain time frame. However, we do not set opening balances for the amount of stock at the start_date nor do we compute the value of the stock. This is wrong and misleading.

To get the value of the stock, we will need to implement the weighted average cost (WAC) algorithm for a given date. See discussion in #5401.

To get the quantity in stock at that point, we can simply look it up from the stock_movement_status table. It will be the sum_quantity of the most recent date directly before the start_date.

The opening balance should be the first line in this report.

-----

### Bug: The "at risk of expiration" flag doesn't work (fix in PR https://github.com/IMA-WorldHealth/bhima/pull/5432)
It appears that the "risky lots" check detects lots that are not actually at risk of expiry.  Here is an example:
image

And here is the CMM calculation:
image

And finally, the expiration date of these lots:
image

You'll notice that we have plenty of time to consume this medication, and will consume this medication at current CMM rates before they expire.  So this flag is not correct.

---

These issues should be addressed before merging this branch.

jmcameron:
I found several problems that need to be fixed on the refactor-stock-management branch before merging into master.

### Monthly Stock Consumption Report (Fixed in https://github.com/IMA-WorldHealth/bhima/pull/5436)
Fails with a SQL error.

### Stock Expiration Report (Fixed in https://github.com/IMA-WorldHealth/bhima/pull/5438)
Does not find any expired stock (even with IMCK data).

### Stock Import (Fixed in https://github.com/IMA-WorldHealth/bhima/pull/5427)
Fails with an error: Wrong number of arguments.  The same import .csv file worked in master.
Also: The inventory code should be quoted in the example template file since it might not be a number.

jniles:
I'm going to work on the SQL errors in `Rapports > Consommations Menseulles`.

jniles:
Okay!  All blocking issues have landed, I believe.  I'm going to attempt to merge this branch with master.

jmcameron:
Good luck!

PR #5342, opened 2021-02-02
refactor(stock): remove stock entry search range in search modal

### LINKED ISSUE #5186: "Entry date" and "Period" (in filters by default) do the same thing in Stock Lots Search Modal

jniles:
Removes the stock entry filter for the period filter instead.

Closes #5186.


jmcameron (changes requested):
Code looks good.

I tested this and I think it makes sense and simplifies things.

However, I was confused when I first looked at the search form because there are just two empty date boxes below "Date d'expiration".   If you think about it or cnce you try some of the given time ranges, it makes sense.  However, I wonder if the boxes could initially display greyed-out 'Start date' and 'End date', respectively.   That would make it really clear.

One problem I encountered:  When I clicked on "This week", I got the little orange indicator "This Week" by Expiration Date.  Nice.  However, when I clear it (link right end above 'end date' field) the indicator was not erased.   Same for "Today", etc.

jniles:
@jmcameron you are right.  This is an issue with the date range more generally.  Would you like me to address it in the PR or a different one?

jmcameron:
Let's make it a separate Issue since it will affect multiple pages.

-Jonathan

jmcameron (approved):
Looks good.  I'll add in issue about the "Start date" "end date" placeholders in the bhDateInterval component.

jniles:
Based on @jmcameron's approval, I am merging this.

bors r+

PR #5344, opened 2021-02-02
Fix: two MySQL & Test bugs

jniles:
This PR fixes two bugs in our MySQL stored procedures and tests.

First, it fixes an end-to-end test bug in the employee config where the bhCheckboxTree state was not being taken into account in the test.  Now, we check if it is all checked before clicking.

Second, it changes the PostToGeneralLedger() trial balance method to be more mysql-8 compatible.  I'm still not 100% sure that this works _every_ time, but I cannot reproduce the issue with these queries re-written.  I believe that MySQL is being fancy and doing some subquery optimizations, which is why we had flakey tests on MySQL8.

To be sure, I'm turning Semaphore's MySQL 8 test suite back on.

### DIFF
--- a/server/models/procedures/trial_balance.sql
+++ b/server/models/procedures/trial_balance.sql
@@ -319,22 +319,23 @@ BEGIN
   DELETE FROM posting_journal WHERE record_uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
 
   -- Let specify that this invoice or the cash payment is posted
-  SELECT COUNT(uuid) INTO isInvoice  FROM invoice  WHERE invoice.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
-  SELECT COUNT(uuid) INTO isCash  FROM cash  WHERE cash.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
-  SELECT COUNT(uuid) INTO isVoucher  FROM voucher  WHERE voucher.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+  SELECT COUNT(uuid) INTO isInvoice FROM invoice WHERE invoice.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+  SELECT COUNT(uuid) INTO isCash FROM cash WHERE cash.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+  SELECT COUNT(uuid) INTO isVoucher FROM voucher WHERE voucher.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
 
   IF isInvoice > 0 THEN
-    UPDATE invoice SET posted = 1 WHERE uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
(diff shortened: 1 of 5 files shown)


jniles:
@jmcameron, can you run the full test suite on your machine and see if it hangs anywhere?

jniles:
bors try

server/models/procedures/trial_balance.sql
```diff
@@ -319,22 +319,23 @@ BEGIN
...
-  SELECT COUNT(uuid) INTO isInvoice  FROM invoice  WHERE invoice.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
-  SELECT COUNT(uuid) INTO isCash  FROM cash  WHERE cash.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
-  SELECT COUNT(uuid) INTO isVoucher  FROM voucher  WHERE voucher.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+  SELECT COUNT(uuid) INTO isInvoice FROM invoice WHERE invoice.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+  SELECT COUNT(uuid) INTO isCash FROM cash WHERE cash.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+  SELECT COUNT(uuid) INTO isVoucher FROM voucher WHERE voucher.uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
 
   IF isInvoice > 0 THEN
-    UPDATE invoice SET posted = 1 WHERE uuid IN (SELECT record_uuid FROM stage_trial_balance_transaction);
+    UPDATE invoice SET posted = 1 WHERE EXISTS (SELECT 1 FROM stage_trial_balance_transaction t WHERE t.record_uuid = invoice.uuid);
```
jmcameron, line 327: SQL reminds me of COBOL.  There may be some folks that are completely fluent in it; but the huge majority of folks who dabble in it have to look up every syntactical thing and then still hack around!  (You might infer that I know COBOL from this; you would be wrong!)

I have no idea why you need to do it this way!  If this necessary in general going forward, I suggest sending the team an email with details.

jmcameron (approved):
APPLAUSE!!!   It runs great on my laptop with WSL2/ubnuntu 20.04 with MySQL 8.0.23.

jniles:
bors try

jniles:
bors try

jniles:
bors try

jniles:
So, I think we can determine that this doesn't solve the issue with MySQL8.

So, here is what I think is happening: In MySQL 8, the algorithm for optimizing `UPDATE table SET z = y WHERE x IN (SELECT ... )` has been improved to
> make use of a semijoin transformation or subquery materialization

and I have no idea what that means.  This was either introduced in `8.0.17` or `8.0.21`.  What this means is that MySQL sometimes figures out a way to optimize the query that invalidates it and causes the query to fail.

I'm going to test my theory on the version switch to confirm it's one of these two issues today by pulling various MySQL versions and seeing how it behaves.

I'll break these changes into two different PRs - one that addresses the End to End tests and a second that address the MySQL8 integration testing issue.

PR #5345, opened 2021-02-02
Stock exit report whitespace

### LINKED ISSUE #5320: The header of the "Stock Exit Report" has too much whitespace
The title of the "Stock Exit Report" is way too big.  It is surrounded by too much whitespace as well.  This wastes paper.  See below:

image

### LINKED ISSUE #5339: Remove brackets from report names
We previously had brackets on report names when we had all reports in a single folder (before https://github.com/IMA-WorldHealth/bhima/pull/4737 landed).  This allowed reports to be grouped in a single large folder.

We no longer need this behavior since we group reports in their own sub-folders.  Let's remove these brackets as they don't make a lot of sense anymore.

jmcameron:
Reformatted the Stock Exit Report (screenshot below).

- Put the title in the top right of the report
- Put "Produced on ..." in the bottom in a footer.  This seems much less important than the title and putting it at the bottom in a footer makes sense.
- Minor formatting cleanups.

Also removed "[Stock]" from the names of the stock report navigation menu titles.

NOTES:  This change should only affect the "Stock Exit" report.   Others can be adapted quickly to use this format and the new handle-bar items, if we decide to use it for all stock reports.

image


jmcameron:
@jniles I tried to get the title "box" to shift more to the right, but got tired of hacking it.  Please look that part over and let me know how I can shift it a bit to the right.  Thanks.

If you like this, I'll add an issue to do the other stock reports.

jniles:
@jmcameron the code is a lot cleaner than I thought it might be to accomplish this. Nice job! 

My one real concern is putting the "Created at..." timestamp on the bottom of the page.  It would be nice to have this on the top of the page as the last page gets lots easier in a busy hospital setting.  And for some of our reports, the creation date really matters since the data becomes out of date.

I think @mbayopanda and/or @lomamech should also give their opinions since they have worked more closely with IMCK on reports.

mbayopanda (commented):
The typo fixes are 👍🏽 

My comment is just if we decide to adopt this way of putting the main title in the right and the `produced by` in the bottom, we will be obliged to do that with all of our reports.

jniles:
@mbayopanda thanks for the review.  I think this is a proposal for all reports - do you think this is a good way to go?

I'm starting to like it more and more as I look at it...

lomamech (commented):
image
After analyzing the new presentation, I thought to myself that we had used the old presentation in all our reports apart from the stock ones, if we should use the new one, we will have to update either all the stock reports or all reports from the BHIMA system

PR #5355, opened 2021-02-03
Enable caching of excludeZeroValue flag in Stock Value Report generat…

### LINKED ISSUE #5325: "Exclude inventories with zero stock" is not cached in the Stock Value Report

jmcameron:
Fix the issue that the "Exclude Zero Value" flag in the Stock Value Report dialog box was not being cached.


jniles (approved):
This looks good to me.  Thanks!

bors r+

PR #5361, opened 2021-02-04
Add page for finding and merging duplicate lots

### LINKED ISSUE #5166: Feature Request: detect duplicate lots
Building off of https://github.com/IMA-WorldHealth/bhima/issues/5165, we probably want a way to detect duplicate lots to merge them.  This would be based on the lot label being the same for a product of the same inventory item.

jmcameron:
Create a page where operators can find and merge potential duplicate lots. 

**Note** that this PR adds a new page under the Stock navigation menu:  "Find Duplicate Lots".  In order to use or test this, be sure to edit the user role permissions and enable the new page (under the Stock section).

Closes #5166

**Testing**
- Use a production database such as IMCK
- Start Bhima
- In Adminstration > Role Management , enable user access to the "Find Duplicate Lots" page.
- Go to Stock > Find Duplicate Lots
   - Note the number of duplicates found (see the grid footer)
   - Click on the Action menu of any row and click on  "Find duplicate lots".  (I'm not 100% happy about the wording here; I'm open to suggestions).
   - A modal will pop up.  Note that form will show all lots that have matching lot labels AND inventory UUIDs.  Only one lot is shown for all the lots that match it.
   - Select the lot to keep and the lot(s) to merge
   - Click on the [Merge] button
   - When you return to the main grid, notice the offending row is now gone (since there is only one of them)

### DIFF
--- a/client/src/modules/stock/lots-duplicates/templates/action.cell.html
+++ b/client/src/modules/stock/lots-duplicates/templates/action.cell.html
@@ -1,0 +1,15 @@
+<div class="ui-grid-cell-contents text-right" uib-dropdown dropdown-append-to-body uib-dropdown-toggle ng-if="!row.groupHeader">
+  <a href>
+    <span data-method="action" translate>FORM.BUTTONS.ACTIONS</span>
+    <span class="caret"></span>
+  </a>
+
+  <ul class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
+    <li class="bh-dropdown-header">{{row.entity.label}} - {{row.entity.inventory_text}}</li>
+    <li>
+      <a data-method="find-duplicates" ng-click="grid.appScope.openDuplicatesModal(row.entity.uuid, row.entity.depot_uuid)" href>
+        <i class="fa fa-search"></i> <span translate>LOTS.MERGE_DUPLICATE_LOTS</span>
+      </a>
+    </li>
+  </ul>
+</div>
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -262,6 +262,7 @@
     "EXCLUDE_EXHAUSTED_LOTS": "Exclude exhausted lots",
     "DETAILS":"Lot Information",
     "FIND_DUPLICATE_LOTS" : "Find duplicate lots",
+    "MERGE_DUPLICATE_LOTS" : "Merge duplicate lots",
     "LOT_LABEL":"Lot label",
     "SUCCESSFULLY_EDITED":"Lot updated with success",
     "ASSIGNMENT_HISTORIC":"Assignment History",
--- a/client/src/i18n/en/table.json
+++ b/client/src/i18n/en/table.json
@@ -163,6 +163,7 @@
             "NR": "Nr",
             "NO_PAYMENTS": "No payments made against this invoice.",
             "NUM": "Number",
+            "NUM_DUPLICATE_LOTS": "# Duplicates",
             "NUMBER_CASES": "Number of cases",
             "OF": "of",
             "PASSIVE": "Passive",
--- a/client/src/i18n/en/tree.json
+++ b/client/src/i18n/en/tree.json
@@ -49,6 +49,7 @@
     "DISPLAY_METADATA": "Display Metadata",
     "DISTRIBUTION_KEYS": "Distribution Keys Management",
     "DITRIBUTION_AUX_FEES_CENTERS" : "Distribution of Auxiliary Fee Centers",
+    "DUPLICATE_LOTS"  : "Find Duplicate Lots",
     "EMPLOYEE" : "Employee",
     "EMPLOYEE_REGISTRY" : "Employee Registry",
     "EMPLOYEE_STANDING_REPORT" : "Employee Standing Report",
--- a/client/src/i18n/fr/stock.json
+++ b/client/src/i18n/fr/stock.json
@@ -262,6 +262,7 @@
     "EXCLUDE_EXHAUSTED_LOTS": "Exclure les lots de stocks épuisés",
     "DETAILS":"Information sur le lot",
     "FIND_DUPLICATE_LOTS" : "Trouver lots en double",
+    "MERGE_DUPLICATE_LOTS" : "fusionner des lots en double",
     "LOT_LABEL":"Designation du lot",
     "SUCCESSFULLY_EDITED":"Mise à jour du lot avec succès",
     "ASSIGNMENT_HISTORIC":"Historique d'assignation",
--- a/client/src/i18n/fr/table.json
+++ b/client/src/i18n/fr/table.json
@@ -163,6 +163,7 @@
             "NR": "N°",
             "NO_PAYMENTS": "Aucun paiement n'a été effectué sur cette facture.",
             "NUM": "Numéro",
+            "NUM_DUPLICATE_LOTS": "Nb de doublons",
             "NUMBER_CASES": "Nombre de cas",
             "OF": "sur",
             "PASSIVE": "Passive",
--- a/client/src/i18n/fr/tree.json
+++ b/client/src/i18n/fr/tree.json
@@ -49,6 +49,7 @@
     "DISPLAY_METADATA": "Afficher les métadonnées",
     "DISTRIBUTION_KEYS": "Gestion des clés de répartitions",
     "DITRIBUTION_AUX_FEES_CENTERS" : "Répartitions des centres de frais auxiliaires",
+    "DUPLICATE_LOTS"  : "Trouver des lots en double",
     "EMPLOYEE":"Gestion des Employés",
     "EMPLOYEE_REGISTRY" : "Registre des employés",
     "EMPLOYEE_STANDING_REPORT" : "Rapport des employés",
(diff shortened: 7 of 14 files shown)


client/src/modules/stock/lots-duplicates/duplicate-lots.html
```diff
@@ -0,0 +1,104 @@
...
+    filters="DupeLotsCtrl.latestViewFilters"
+    on-remove-filter="DupeLotsCtrl.onRemoveFilter(filter)">
+  </bh-filters>
+</div>
+-->
+<div class="flex-content">
+  <div class="container-fluid">
+    <div
+      id="duplicate-lots-grid"
+      class="grid-full-height-with-filters"
```
jniles: FYI - to make this grid full height, use the class `grid-full-height`.  This accounts for the "filter-bar".

jniles (commented):
@jmcameron since you asked for a review of the functionality, I won't dive into the code yet.

Overall, looks really good for a first cut.  Some suggestions:

  1. Instead of "Find Lots" in the action menu, let's just call it "Merge Lots".  To me, this is intuitive: instead of finding the lots, we've found them, now we are opening the modal to merge them.
  2. Can you put a loading indicator on the loading button when merging lots?  It seems like we should be waiting for that to be done on the server, then closing the modal, then reloading the page.  It dismisses almost immediately, which surprised me.  Most other modals that do something in the system (trial balance, credit note, etc) hang out until the action is performed with a loading indicator.
 
Here is what that looks like on a "good 2G network" that I emulated.
MOYdGh0e9j

Again, haven't taken a look at the code, but it surprised me how quick that was!

jmcameron:
@jniles This is now ready for a real review.   

**QUESTION**:  Should we remove the "Find Duplicate Lots" in the Lots registry action menu?

jniles:
> QUESTION: Should we remove the "Find Duplicate Lots" in the Lots registry action menu?

No, I do not think so.  Let's wait until a couple of releases, then we'll remove it.  I think that folks are much more inclined to absent-mindedly look at that registry (which they are using all the time) than they will be for this new module.

client/src/modules/stock/lots-duplicates/duplicate-lots.html
```diff
@@ -0,0 +1,59 @@
+<div class="flex-header">
```
jniles: A meta comment - I find it a little strange this file path is `lots-duplicates/duplicate-lots.html`.  Is a reason why the folder name and the file name are different?

client/src/modules/stock/lots-duplicates/duplicate-lots.html
jniles: Since we don't have any filters on this registry, you should include this change.

client/src/modules/stock/lots-duplicates/duplicate-lots.js
```diff
@@ -0,0 +1,165 @@
...
+ */
+function DuplicateLotsController(
+  Lots, Notify, uiGridConstants, Modal,
+  Session, GridState, $state, Columns,
+) {
+  const vm = this;
+  const cacheKey = 'duplicate-lots-grid';
+  vm.enterprise = Session.enterprise;
+
+  // const duplicateLotsFilters = Stock.filter.inventory;
```
jniles: Let's just remove this.

client/src/modules/stock/lots-duplicates/templates/action.cell.html
```diff
@@ -0,0 +1,15 @@
+<div class="ui-grid-cell-contents text-right" uib-dropdown dropdown-append-to-body uib-dropdown-toggle ng-if="!row.groupHeader">
+  <a href>
+    <span data-method="action" translate>FORM.BUTTONS.ACTIONS</span>
+    <span class="caret"></span>
+  </a>
+
+  <ul data-action="{{ rowRenderIndex}}"  class="dropdown-menu-right" bh-dropdown-menu-auto-dropup uib-dropdown-menu>
```
jniles: You don't need the `{{rowRenderIndex}}`.  This is a testing anti-pattern (click in row by it's position in the grid).  Since there isn't an end to end test for this yet, you can just remove this binding.

client/src/modules/stock/lots-duplicates/duplicate-lots.js
```diff
@@ -0,0 +1,165 @@
...
+  // load stock lots in the grid
+  function load() {
+    vm.hasError = false;
+    vm.loading = true;
+
+    Lots.dupes({ find_dupes : true })
+      .then((rows) => {
+        vm.gridOptions.data = rows;
+      })
+      .catch(Notify.handleError)
```
jniles: You'll want to set "hasError" to true here.
```js
.catch((err) => {
  hasError = true;
  Notify.handleError(err);
})
```

This tells the grid to show an error message.  `Notify.handleError` doesn't do that part automatically - it just makes the notification bar come down.

client/src/modules/stock/lots-duplicates/duplicate-lots.js
```diff
@@ -0,0 +1,165 @@
...
+
+  vm.gridApi = {};
+  vm.saveGridState = state.saveGridState;
+
+  function clearGridState() {
+    state.clearGridState();
+    $state.reload();
+  }
+
+  function onRegisterApi(gridApi) {
```
jniles: You can probably remove all references to `gridApi`.  I don' think you need it or use it anywhere.  I can't see a scenario where you would need it either.

client/src/modules/stock/lots-duplicates/duplicate-lots.js
```diff
@@ -0,0 +1,165 @@
...
+  .controller('DuplicateLotsController', DuplicateLotsController);
+
+DuplicateLotsController.$inject = [
+  'LotService', 'NotifyService', 'uiGridConstants', 'StockModalService',
+  'SessionService', 'GridStateService', '$state', 'GridColumnService',
+];
+
+/**
+ * Stock Inventory Controller
+ * This module is a registry page for stock inventories
```
jniles: These comments should probably be updated to reflect the true nature of the controller.

client/src/modules/stock/lots-duplicates/duplicate-lots.html
```diff
@@ -0,0 +1,59 @@
...
+<div class="flex-content">
+  <div class="container-fluid">
+    <div
+      id="duplicate-lots-grid"
+      class="grid-full-height-with-filters"
+      ui-grid="DupeLotsCtrl.gridOptions"
+      ui-grid-resize-columns
+      ui-grid-auto-resize
+      ui-grid-save-state
+      ui-grid-grouping>
```
jniles: Do you need the `ui-grid-grouping` here?

jniles (approved):
Functionally, this looks good to me.  I've made a few clean up comments, but none of them are serious.  Once addressed, this PR can be merged.

jniles:
bors delegate+

client/src/modules/stock/lots-duplicates/duplicate-lots.html
jmcameron: Done

client/src/modules/stock/lots-duplicates/duplicate-lots.js
jmcameron: I wanted this to sort next to the 'lots' file folder uner 'stock', hence the name.  I plan to rename the files to 'lots-duplicates.*' to match the directory name.   'duplicate-lots' would be more readable, but would make it a bit harder to find with the file explorer.

client/src/modules/stock/lots-duplicates/duplicate-lots.html
jmcameron: I wanted this to sort next to the 'lots' file folder uner 'stock', hence the name. I plan to rename the files to 'lots-duplicates.*' to match the directory name. 'duplicate-lots' would be more readable, but would make it a bit harder to find with the file explorer.

jmcameron: Files renamed.

client/src/modules/stock/lots-duplicates/duplicate-lots.js
jmcameron: Removed

client/src/modules/stock/lots-duplicates/templates/action.cell.html
jmcameron: Removed

client/src/modules/stock/lots-duplicates/duplicate-lots.js
jmcameron: Updated

client/src/modules/stock/lots-duplicates/duplicate-lots.html
jmcameron: Removed

client/src/modules/stock/lots-duplicates/duplicate-lots.js
jmcameron: Removed.

client/src/modules/stock/lots-duplicates/duplicate-lots.js
jmcameron: Updated.

PR #5365, opened 2021-02-05
Fixed bhStockFilterExhausted to support help text

jmcameron:
Fix bhStockFilterExhausted to enable callers to specify help text to be shown below the stock filter exhausted options.   Note that this was not correctly implemented before (the help text was hardcoded). 

**Testing**
- Run Bhima
- In the Stock Lots page, open the search modal and click on the defaults tab.  There should be no additional text below the 'include exhausted lots' options.
- Edit client/src/modules/stock/lots/modals/search.modal.html
- Search for bh-stock-exhausted (around line 108)
- Add this to the component invocation:   help-text="TEST TEXT"
- Run Bhima
- In the Stock Lots page, open the search modal and click on the defaults tab.  The additional text should appear below the 'include exhausted lots'  options.

It is not clear that this will ever be needed, but since it was there, I though we should fix it so that it will work properly.

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -75,7 +75,6 @@
     "GROUPING"        : "Grouping",
     "JOIN_INVOICE"    : "Join an invoice",
     "INCLUDE_EMPTY_LOTS" : "Include empty lots",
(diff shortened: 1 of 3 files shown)


jniles (approved):
Looks good to me!  Good catch!

bors r+

PR #5368, opened 2021-02-08
Display inventory name in merge lots modal

jmcameron:
Update merge lots modal to display the inventory name in the modal.

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -437,13 +437,12 @@ li.selected {
 
 /**
  * Temporary header breadcrumb bootstrap fork
- * @todo implement this in less using the boostrap element
+ * @todo implement this in less using the bootstrap element
  */
 .headercrumb {
   padding: 8px 0px;
   list-style: none;
   border-radius: 2px;
-
 }
 ol.headercrumb {
   margin-bottom : 0px !important;
--- a/client/src/modules/stock/lots/modals/duplicates.modal.html
+++ b/client/src/modules/stock/lots/modals/duplicates.modal.html
@@ -12,6 +12,9 @@
   </div>
 
   <div class="modal-body">
+    <div style="margin-bottom: 0.5em">
+      <span class="text-bold">{{$ctrl.selectedLot.inventory_code}} - {{$ctrl.selectedLot.inventory_text}}</span>
+    </div>
     <div ng-if="$ctrl.lots.length === 1">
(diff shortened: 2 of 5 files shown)


client/src/css/structure.css
```diff
@@ -437,13 +437,12 @@ li.selected {
 
 /**
  * Temporary header breadcrumb bootstrap fork
- * @todo implement this in less using the boostrap element
+ * @todo implement this in less using the bootstrap element
```
jniles, line 440: :+1:

client/src/modules/stock/lots/modals/duplicates.modal.html
```diff
@@ -12,6 +12,9 @@
   </div>
 
   <div class="modal-body">
+    <div style="margin-bottom: 0.5em">
+      <span class="text-bold">{{$ctrl.selectedLot.inventory_text}}</span>
```
jniles: This may seem nit-picky, but some of our users are more comfortable with the inventory code.  Could you add that before the text, separated by a hyphen?
```html
{{ inventory_code}} - {{inventory_text.}}

```

jniles (dismissed):
Changes look good to me.  If you could add the code, as suggested, it would be helpful.

Here is what it looks like with that Vanga database:
image

client/src/modules/stock/lots/modals/duplicates.modal.html
jmcameron: Done.

jmcameron:
@jniles Please take one more look at this.  I also added the code to the Find Duplicate Lots page (after adding it to the modal per your suggestion).  But it occurred to me that it might be good to reorder the first three columns (put the lot label third).  Please take a lot and let me know what you think.  Also since we have more inventory columns, perhaps I should change the lot column table from "Label" to "Lot"?   Let me know what you think.  Then I'll fix it, rebase, and release.

jmcameron:
> @jniles Please take one more look at this. I also added the code to the Find Duplicate Lots page (after adding it to the modal per your suggestion). But it occurred to me that it might be good to reorder the first three columns (put the lot label third). Please take a lot and let me know what you think. Also since we have more inventory columns, perhaps I should change the lot column table from "Label" to "Lot"? Let me know what you think. Then I'll fix it, rebase, and release.

Per discussion with @jniles, I implemented both these changes.

Issue #5364, opened 2021-02-05
Move to MySQL8


jniles:
For some of our advanced queries, it would b really helpful to move entirely to MySQL 8.  Specifically, it would be nice to have [window functions](https://dev.mysql.com/doc/refman/8.0/en/window-functions.html) and partitioning.

We have a few places where we hack this together for the "simpler" queries for running balance.  However, if we want to calculate multiple running balances (as we do in the CMM calculations), we'll want to keep track of multiple running balances at a time.  I can't figure out an easy way to do this in MySQL 5.7, but it is trivial in MySQL 8:
```sql
-- pseudocode
SELECT 
  quantity,
  sum(quantity) over (partition by inventory_uuid order by day) as cumulative_sum
FROM stock_movement_status
```

These benefits (plus the utf8mb4 by default, etc) should make us take a serious look at moving entirely to MySQL 8.

~~We do have issues in our trial balance tests for MySQL 8 that should give us pause ... but in theory it's a misunderstanding on our part and we can figure them out.~~ UPDATE: solved this.

jniles:
Just an update: I figured out the issue with MySQL8 tests.  It was a query optimizer that was over eager.

PR #5372, opened 2021-02-09
Rework stock_movement_status calculation

### LINKED ISSUE #5073: Remove the stock_consumption table and associated procedures

### LINKED ISSUE #5144: Remove the inventory.avg_consumption column

jniles:
This is a full rewrite of the `stock_movement_status` table and associated stored procedures.  The goal of this rewrite is to improve clarity, remove user-defined variables that may potentially collide across MySQL connections, and improve the performance of the `GetCMM()` calculation.

Closes #5073.
Closes #5144.
Closes https://github.com/IMA-WorldHealth/bhima/issues/5332.
Closes https://github.com/IMA-WorldHealth/bhima/issues/5360.

----
image

image

heidisql_0CKbDzKdzl

This PR contains many far reaching changes.  Here is a lot of the notable ones:

 1. I've removed the `avg_consumption` column from the inventory table.  It was never used, as far as I can tell, to do any reporting.  It was used in the stock import feature, but we don't use the `avg_consumption` to report on, so I just removed that code.
 2. I've removed the `stock_consumption` table and associated procedures.  It is used in several reports and they are now broken.  However, I would argue they were always broken since the `stock_consumption` did not record accurate information about stock consumption.
 3.  I've entirely changed the structure behind the `stock_movement_status` table for the following reasons:
     1. The [previous structure](https://github.com/IMA-WorldHealth/bhima/blob/e455a8011db69ad48839afc35977d3dd041d3241/server/models/schema.sql#L2016) was very hard to audit.  For example, it had `start_date`, `end_date`, and `quantity`.  Is that the quantity at the start?  The quantity at the end?  If I wanted to find the value on date X, I needed to do a BETWEEN query.
     2. The table did not distinguish between consumption information and exit information.  This made the `GetCMM()` calculation have to look up additional information from the `stock_movement` table.  A user would need to as well.
     3. In theory, we'll be calling `GetCMM()` more than we will be moving stock.  It makes sense to make the performance tradeoff to increase the speed of `GetCMM()` and decrease the speed of inserting stock.
 5. I've renamed the `GetCMM()` call to `GetAMC()` for clarity.  It also takes in a single date and uses the depot to find the correct value of the date range.  This prevents us from having inconsistent results because a developer didn't properly compute the date range.  If we need to support a custom date range, we'll just write another SP.
 6. I've slightly changed the API of `GetAMC()`:
     1. It returns `head_days` and `tail_days` instead of `days_before_stock_consumption`.  That is because `days_before_stock_consumption` is the real number of days before the first consumption.  Instead, it the days between the start of the window and the next record.  Similarly, `tail_days` gives you the number of days between the last consumption record and the end of the window.
     2. It returns the `quantity_in_stock` so that we can perform stock out date calculations.
     3. `first_inventory_movement_date` -> `min_date`
     4. `last_inventory_movement_date` -> `max_date`

    
------

It isn't all good.  In fact, initializing this is terrible.  On my machine (4GB of RAM):
```mysql
mysql> use vanga;
mysql> CALL zRecomputeStockMovementStatus();
Query OK, 0 rows affected (22 min 4.81 sec)

mysql> use imck;
mysql> CALL zRecomputeStockMovementStatus();
Query OK, 0 rows affected (4 min 20.31 sec)
```

This is _horrible_.  I did get a lot faster speeds when I was just testing on MySQL 8, but the syntax didn't work on MySQL 5.7 (#5364).  When I fixed it on MySQL5.7, I ran into issues on MySQL 8.  Basically, we got the worst performance of both worlds by trying to get it to work correctly on both worlds.

However, I still think it is worth merging because the performance of any individual transaction isn't bad.  It is only when rebuilding the database that things go horribly long.

----

### How to test

Download this PR and run the tests!  (`yarn test:integration` and `yarn test:integration:stock`).  If that works, try with a production database.  Note, as above, you'll need to `CALL zRecomputeStockMovementStatus()` to build the table.


jmcameron (commented):
I did not review the code changes.
Both test:integration and test:integration:stock ran fine on my laptop (WSL Ubuntu 20.04, MySQL 8).
I installed the IMCK db, AND migration files.
I ran CALL zRecomputeStockMovementStatus(); it took 4:40 on my laptop.
The CMM calculations in 'Articles in Stock' seem to be correct.  I spot-checked all 4 algorithms and the calculations were correct.  However Alg 2 still seems to produce very different (higher) estimates than the other 3 algorithms.

Based on direct feedback from @JNiles, I think there will be another review of this in the future; I'll review the code then.

jniles:
@jmcameron this is ready for a final review.

client/src/modules/stock/inventories/modals/amc.modal.js
```diff
@@ -12,32 +12,39 @@ function StockAMCModalController(Stock, Notify, Instance, data, moment, Constant
...
+        vm.data.avg_consumption = vm.data[vm.settings.average_consumption_algo];
+
+        // calculate date when the stock will run out at current consumption rate
+        const daysOfStockLeft = (vm.data.quantity_in_stock / vm.data.avg_consumption) * 30.5;
+
+        // NOTE: momentjs does not accept decimals as of 2.12.0
+        const stockOutDate = moment().add(daysOfStockLeft, 'days').toDate();
+
+        // provide this information to the view.
+        vm.data.stock_out_date = stockOutDate;
```
jmcameron: Is there any reason not to combine this with the line above?

server/controllers/inventory/inventory/core.js
```diff
@@ -259,7 +259,7 @@ function getItemsMetadataById(uid) {
       it.text AS type, ig.name AS groupName, BUID(ig.uuid) AS group_uuid,
       ig.unique_item, i.consumable, i.locked, i.stock_min, i.sellable,
       i.stock_max, i.created_at AS timestamp, i.type_id, i.unit_id, i.unit_weight, i.unit_volume,
-      ig.sales_account, i.default_quantity, i.avg_consumption, i.delay, i.purchase_interval,
+      ig.sales_account, i.default_quantity, i.delay, i.purchase_interval,
```
jmcameron, line 262: The name of this function reminds me that we need an issue to convert the 'items' terminology to 'articles' throughout.

server/controllers/stock/import.js
```diff
@@ -37,6 +37,7 @@ function downloadTemplate(req, res, next) {
 async function importStock(req, res, next) {
   let queryParams;
 
+  const operationDate = new Date(req.body.date);
```
jmcameron, line 40: I like being able to specify the date of the import.  It makes sense because the data could be collected at a date before the operator is able to import it.

server/controllers/stock/index.js
```diff
@@ -132,7 +132,7 @@ async function createStock(req, res, next) {
     });
 
     const isExit = 0;
-    const postingParams = [db.bid(documentUuid), isExit, req.session.project.id, req.session.enterprise.currency_id];
+    const postingParams = [db.bid(documentUuid), isExit, req.session.project.id];
```
jmcameron, line 135: Why don't we need the currency_id?

jmcameron (changes requested):
I looked at the stock import page and had some questions about the import file syntax.  When I downloaded the sample page I see:
image
I suggest changing the group name and inventory_unit so that they do not have any non-ASCII unicode characters.   Also, should the numbers be quoted?

On the importStock page: Shouldn't the import modal support the currency ID?  It does not seem safe to assume the incoming inventory_unit_price will be whatever is the current default for Bhima.

There is a problem with the wording of description for the calculation for Algorithm 4:
image
Notice that "difference of the number of months in the period minus the total number of days of stock out in the period" implies you are subtracting days from months -- which is a units no-no.  The calculation is correct, but the description has a problem.

I did a units check on algorithms 1, 3 and 4 and they made sense.  While the units do technically work out for algorithm 2, it still does not make sense because it implies that in the next month there would be stock consumed EVERY day at the same rate as it was for the few days stock was consumed during the previous period. Here is what algorithm 2 should be:

   Nsc = number of stock consumption days
   A = pills per stock consumption day
   B = avg number stock consumption days per month

  AMC = A*B = (23 / Nsc)*(Nsc / 184)*30.5
           = (23/3)*(3/184)*30.5 = 3.81

But since there no way to estimate number of days of stock consumed in the future, the estimate has to be based on the past.  Just using the average over the period simply leads Algorithm 3.

server/controllers/stock/index.js
jniles: We already know it.  Internally, we value the stock in the enterprise currency id and there isn't any option to change it.  Even if you purchase stock in a foreign currency, we do the internal conversion so that it shows up correctly on the balance sheet.

There isn't anywhere that a user can change the currency associated with stock that I am aware of.

client/src/modules/stock/inventories/modals/amc.modal.js
jniles: Nope!  I'll do that.

jniles:
@jmcameron Thanks for the review!

>  I suggest changing the group name and inventory_unit so that they do not have any non-ASCII unicode characters. Also, should the numbers be quoted?

This is somewhat out of our control.  If you have a notepad editor (for example, Notepad++), you'll notice that we exporting the CSV as UTF-8 since we support more languages than en-US.  The name you would like changed is user-defined text and they should be allowed to put French accents in their names.  The problem is that an American MS Excel installation assumes every CSV is Windows ISO encoding, not UTF8, and there is no way to tell it otherwise from the export itself.

As for the numbers being quoted, I do think that is an option of the export library.  I'd prefer to cover it in another PR if we want to change it.  I can see that being a pretty significant bunny trail of breaking tests. (#5398)

> On the importStock page: Shouldn't the import modal support the currency ID? It does not seem safe to assume the incoming inventory_unit_price will be whatever is the current default for Bhima.

This is a great observation.  I've made an issue about it (#5399).  Currently, the stock import is mainly targeted towards super users who know what they are doing with the stock.  I think our team is the only one that actually has used it to date, but this would make it more functional.

>  Notice that "difference of the number of months in the period minus the total number of days of stock out in the period" implies you are subtracting days from months -- which is a units no-no. The calculation is correct, but the description has a problem.

I've tried to fix this ... let me know what you think.  The issue is that the description will be identical as ALGO_1 if we use the "natural English" approach to it.

> I did a units check on algorithms 1, 3 and 4 and they made sense. While the units do technically work out for algorithm 2, it still does not make sense because it implies that in the next month there would be stock consumed EVERY day at the same rate as it was for the few days stock was consumed during the previous period. Here is what algorithm 2 should be:

See https://github.com/IMA-WorldHealth/bhima/issues/5343.  I agree we need to review our algorithms and make them sane.  I don't think this makes sense to do here.

client/src/modules/stock/stock.service.js
```diff
@@ -244,10 +244,12 @@ function StockModalService(Modal) {
 
   function openAMCCalculationModal(request) {
     const templateUrl = 'modules/stock/inventories/modals/amc.modal.html';
-    const controller = 'StockAMCModalController as ModalCtrl';
+    const controller = 'StockAMCModalController';
+
     const params = angular.extend(modalParameters, {
       templateUrl,
       controller,
+      controllerAs : 'ModalCtrl',
```
mbayopanda, line 252: 👍🏽

mbayopanda (approved):
LGTM 👍🏽

jmcameron:
> I suggest changing the group name and inventory_unit so that they do not have any non-ASCII unicode characters. Also, should the numbers be quoted?

Sorry, I was not clear here.  What I was suggesting is that the example download CSV file not have funky unicode characters.  The issue of what users provide is outside of our control.

jniles:
@jmcameron ah, I understand now.  I'll take another look at this.

jniles:
@jmcameron changes landed!

jniles:
@jmcameron I think this is waiting for you to pull the trigger.  @lomamech is working on a different PR that is important.  This is also gating future work.  I suggest we get any kinks you can find in it worked out, then we merge it in.

jmcameron:
Looks good.

Bors r+

PR #5387, opened 2021-02-15
Add stock entry typeahead support for selecting existing lots

### LINKED ISSUE #5257: Make lot numbers unique per inventory item
Our friends at HEV have requested that we make lot number unique per inventory item.  Currently, lot numbers do not need to be unique at all.

Basically, during lot creation, we would need to look up the lot_uuid based on the lot number, and substitute that lot number if it exists.  From a user interface perspective, it might be useful to have a way to auto-populate the expiration date if they have put in the lot already.

This functionality would ~~replace~~ render obsolete https://github.com/IMA-WorldHealth/bhima/issues/5165.  Instead of merging lots, just put the same lot number and they will be merged.

jmcameron:
Support entering lots in Stock Entry using typeahead to select among existing lots. 
     - If an existing lot is selected, its expiration date is inserted to the lot selection form line.
     - If the user types the label of an existing lot, it is selected immediately.  This should prevent users from using duplicate lot names.
     - If a lot label that is typed in is not one of the possible candidate lots for that inventory item, the code assumes that it is a new lot.  The user will then have to correct the expiration date.

### NOTES

- I did not add code to see if the typed-in lot label (not selected) lot exists.  Hopefully the features above make that unnecessary.   
- I was able to verify that this works for integration and donations.  It does not seem relevant to Transfer since those already have lot labels.  
- I fixed the code for Purchases/Transfers so that it knows about the candidate lots for each inventory item.
- This PR does not disable the expiration date field when set for the selected candidate stock.   This will require an update to the bhDatePicker component and should be done in a separate PR.
- **WARNING**: However the current code for candidate lots (for the typeahead) does not check on whether the lots are expired or empty.  I suggest fixing that in another PR.
- BTW, I noticed an unrelated error/problem in Transfers:  When you select a transfer, it lists the lots in the transfer, but they are all flagged as being okay.  But if you click on the lot(s) link, some of the lot items may be invalid (bad expiration date, etc).

### TESTING

- Use one of the production databases for testing so there are multiple lots available for inventory items.
- Try typing a new label and verify that it works like it used to.
- When adding new lots, try the typeahed.  Once you enter a full Lot label, it should automatically select it.
- When testing, please try all 4 means of entry
- For Integration and Dontations, lots can either be created or reused.  Creating new lots (type until no existing lots are shown by the typeahead drop-down for a new Lot) and Using existing lots (selecting an existing lot using the typeahead).   In either case, note the lot labels for new and existing lots.
   - After the integration/donation tests, be sure to try "Find Duplicate Lots" and verify that no duplicate lots were created (check the lot labels in the previous step).  Note that his can also be down via the Stock Lots registry action menu.

### DIFF
--- a/client/src/modules/stock/entry/modals/lots.modal.js
+++ b/client/src/modules/stock/entry/modals/lots.modal.js
@@ -3,11 +3,13 @@ angular.module('bhima.controllers')
 
 StockDefineLotsModalController.$inject = [
   'appcache', '$uibModalInstance', 'uiGridConstants', 'data',
-  'SessionService', 'bhConstants', 'StockEntryModalForm', 'focus',
+  'SessionService', 'bhConstants', 'StockEntryModalForm',
+  '$translate', 'focus',
 ];
 
 function StockDefineLotsModalController(
-  AppCache, Instance, uiGridConstants, Data, Session, bhConstants, EntryForm, Focus,
+  AppCache, Instance, uiGridConstants, Data, Session, bhConstants, EntryForm,
+  $translate, Focus,
 ) {
   const vm = this;
 
@@ -30,7 +32,6 @@ function StockDefineLotsModalController(
   vm.enterprise = Session.enterprise;
   vm.stockLine = angular.copy(Data.stockLine);
   vm.entryType = Data.entry_type;
-
   vm.isTransfer = (vm.entryType === 'transfer_reception');
 
   // exposing method to the view
@@ -42,6 +43,8 @@ function StockDefineLotsModalController(
   vm.onChangeQuantity = onChangeQuantity;
   vm.onChangeUnitCost = onChangeUnitCost;
   vm.onDateChange = onDateChange;
+
+  vm.onSelectLot = onSelectLot;
 
   vm.isCostEditable = (vm.entryType !== 'transfer_reception');
 
@@ -85,11 +88,13 @@ function StockDefineLotsModalController(
   vm.gridOptions = {
     appScopeProvider : vm,
     enableSorting : false,
+    enableColumnResize : true,
     enableColumnMenus : false,
     showColumnFooter : true,
     fastWatch : true,
     flatEntityAccess : true,
     data : vm.form.rows,
+    minRowsToShow : 4,
     columnDefs : cols,
     onRegisterApi,
   };
@@ -115,9 +120,12 @@ function StockDefineLotsModalController(
    * if the fast insert option is enable do this :
    * - add new row automatically on blur
    * - set the focus in the new row
-   * @param {string} rowLot the row.entity.lot string
+   * @param {string} rowLot the row.entity.lot string/object
    */
   function onLotBlur(rowLot) {
+    // NOTE: rowLot will be an object if an existed lot was
+    //       selected from the typeahead.  Otherwise it will
+    //       be the lot name string that was typed in.
     if (vm.enableFastInsert && rowLot) {
 
       const emptyLotRow = getFirstEmptyLot();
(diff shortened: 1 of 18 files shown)


server/config/routes.js
```diff
@@ -951,7 +951,8 @@ exports.configure = function configure(app) {
   app.put('/lots/:uuid', lots.update);
   app.get('/lots/:uuid/assignments/:depot_uuid', lots.assignments);
   app.post('/lots/:uuid/merge', lots.merge);
-  app.get('/lot/dupes/:label?/:entry_date?/:expiration_date?/:initial_quantity?/:inventory_uuid?', lots.getDupes);
+  app.get('/lots_candidates/:inventory_uuid', lots.getCandidates);
```
jniles: The root of this in an inventory item, right?  If so, I would prefer that we use the path `/inventory/:uuid/lot_candidates`.  It's more REST-y.

test/integration/mergeLots.js
```diff
@@ -139,15 +139,15 @@ describe('Test merging lots', () => {
   // ===========================================================================
   // Verify we have created the lots, tags, etc
   it('Verify we created lot1 and its dupes (lot3, lot4, lot5)', () => {
-    return agent.get('/lot/dupes')
+    return agent.get('/lots_dupes')
```
jniles, line 142: 👀

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -115,9 +120,12 @@ function StockDefineLotsModalController(
    * if the fast insert option is enable do this :
    * - add new row automatically on blur
    * - set the focus in the new row
-   * @param {string} rowLot the row.entity.lot string
+   * @param {string} rowLot the row.entity.lot string/object
    */
   function onLotBlur(rowLot) {
+    // NOTE: rowLot will be an object if an existed lot was
+    //       selected from the typeahead.  Otherwise it will
+    //       be the lot name string that was typed in.
```
jniles, line 128: I believe we need to deal with this at some place to disaggregate the object into it's component parts.  Where do we do that?

jniles (changes requested):
Awesome work!

I used one old and one new lot, and here is what I got on the receipt as a return: 

image
This should be 1924, a very famous lot at Vanga.  I believe that it's because we are selecting an object when we select one in the input and it should at some point get dis-aggregated into a UUID and label.

Code and behavior look pretty good.  I'll give this a second pass once the `[Object object]` issue gets addressed.

server/config/routes.js
jmcameron: I fixed the lot name issue and changed the URL per your suggestion.

jmcameron: If you think it makes sense, I can move the lot candidate function to the inventory controller and service.  I have now left it in the lots controller and service.

test/integration/mergeLots.js
jmcameron: I was having problems making a unique url path for it because of all the optional arguments it supports.  I'm open to suggestions.

client/src/modules/stock/entry/modals/lots.modal.js
jmcameron: I fixed this in the lots.modal.js submit() function.   In the past we only dealt with 'lot' as a string.  But now since we have the drop-down, 'lot' is actually an object during the typeahead.  I fixed the submit function to do the necessary conversion if an existing lot was selected.  I've tested it and it seems to work.

jmcameron:
Note: I've realized that I can still create duplicate lots with this addition.  I'll track down why and fix it.
@jniles I still request you give a good going over.   I think the fix for avoiding creating duplicate lots will be a relatively small fix (I seem to be creating a new UUID instead of using the one provided by the selection and not creating a new lot somewhere).

jniles:
@jmcameron my assumption is that you need to disaggregate all the information in the lot, including the uuid and the expiration date.  I _think_ it should then just magically work.

So, try adding that into your disaggregation function and see if that solves it!

jmcameron:
> Note: I've realized that I can still create duplicate lots with this addition. I'll track down why and fix it.
> I seem to be creating a new UUID instead of using the one provided by the selection and not creating a new lot somewhere.

Just fixed this issue.

jmcameron:
@JNiles I think this is now ready for a thorough review.

jmcameron:
@JNiles : Now it is ready for a thorough review.  Thanks.

server/controllers/stock/index.js
```diff
@@ -85,13 +85,14 @@ async function createStock(req, res, next) {
 
       if (lotUuid === null) {
         // Create new lot (if one it does not already exist)
+        lotUuid = uuid();
```
jniles, line 88: 😃

jniles (approved):
Works for me!  I've tested every case and it seems to work.

bors r+

PR #5396, opened 2021-02-18
Automatically merge duplicate lots

### LINKED ISSUE #5370: Implement function to automatically merge lots which can be merged without user decisions
In the production databases, we have many duplicate lots.   Some of these will require knowledgeable uses to determine which lots are the and merge them as appropriate.   However, there probably are some duplicate lots that could be merged automatically.   Taking care of these would reduce the work for users how need to make decisions about how to handle the other lot merges. 

Lots have these intrinsic characteristics:

- Same UUID
- Same lot label
- Same inventory_uuid
- Same expiration date

There are other fields in the 'lot' table, but some of them are problem the result of muddled thinking about lots.

If lots have the same values for all the intrinsic fields above, they can probably be merged automatically.

Once we clarify what constitutes identical lots, we should be able to write a function that will identify identical lots, we can create function to do that and put it in the main drop-down [Menu] in the 'Find Duplicate Lots' page.

@JNiles, please review this issue and let me know what you think about how to identify identical duplicate lots that can be merged automatically without user decisions.

jmcameron:
This PR adds a function to automatically merge duplicate lots (where unambiguous).   All lots that have the same lot label, inventory_uuid, and expiration date will be merged by this function.

Closed #5370

This function is in the main menu of the 'Stock > Find Duplicate Lots' page.

**WARNING:** This is a first cut.  Do not release it yet!
- It still needs some work on the autoMerge() function in server/controllers/stock/lots.js
- The function works but does not refresh the duplicate lots list.  I think this has to do with the way the individual merge() functions are being used.

**TESTING**
- Use one of the production sites (eg, IMCK)
- Be sure to enable the permissions for the test user to access the "Find Duplicate Lots" page (under the Stock menu).
- In the Stock > Find duplicate lots, note the number of duplicate lots found (at the bottom of the page)
- Try the "Automatically Merge Lots" in the main [Menu] button
- It should work, but you will to have to refresh the page to see the updated total.  With the IMCK db, this went from 423 down to 201 after executing this function.


server/controllers/stock/lots.js
```diff
@@ -198,12 +201,80 @@ async function merge(req, res, next) {
...
   } catch (error) {
     next(error);
   }
 }
 
+/**
+ * GET /lots/merge/auto
+ *
+ * @description
+ * Finds and merges all lots suitable for automatic merging
```
jniles, line 231: I like these descriptions :+1:

server/controllers/stock/lots.js
```diff
@@ -198,12 +201,80 @@ async function merge(req, res, next) {
...
+ *  - To qualify, the lots must have the same inventory_uuid,
+ *    label, and expiration date.
+ */
+function autoMerge(req, res, next) {
```
jniles, line 235: Put an `async` tag here for my future comments.

server/controllers/stock/lots.js
```diff
@@ -198,12 +201,80 @@ async function merge(req, res, next) {
...
+    FROM lot l
+    JOIN inventory i ON i.uuid = l.inventory_uuid
+    WHERE l.label=? AND i.uuid=? AND l.expiration_date=DATE(?)
+  `;
+  db.exec(query1, [])
+    .then((rows) => {
+      const numInventories = rows.length;
+      let numLots = rows.reduce((sum, row) => {
+        return sum + row.num_duplicates;
+      }, 0) - numInventories;
```
jniles, line 270: This is very cheeky.  😄

server/controllers/stock/lots.js
```diff
@@ -198,12 +201,80 @@ async function merge(req, res, next) {
...
+      rows.forEach(async (row) => {
+        db.exec(query2, [row.label, db.bid(row.inventory_uuid), row.expiration_date])
+          .then((lots) => {
+            // Arbitrarily pick first lot and merge the duplicates into it
+            const masterLotUuid = lots[0].uuid;
+            const lotUuids = lots.reduce((list, elt) => {
```
jniles, line 277: It looks like you are trying to `filter` all lots that are the master lot out of the array.  Let's do that!  Pretty sure this will give you the same result with clearer syntax:

```js
const lotUuids = lots.filter(elt => elt.uuid !== masterLotUuid);
```

If you are familiar with ES5, this is pretty easy to read.  "Filter out all `elt` who's `uuid` is not `masterLotUuid`".  If that isn't what you were going for, let me know!

This will return a new array without modifying the underlying `lots` array.

server/controllers/stock/lots.js
```diff
@@ -198,12 +201,80 @@ async function merge(req, res, next) {
...
+    JOIN inventory i ON i.uuid = l.inventory_uuid
+    WHERE l.label=? AND i.uuid=? AND l.expiration_date=DATE(?)
+  `;
+  db.exec(query1, [])
+    .then((rows) => {
+      const numInventories = rows.length;
+      let numLots = rows.reduce((sum, row) => {
+        return sum + row.num_duplicates;
+      }, 0) - numInventories;
+      rows.forEach(async (row) => {
```
jniles: Async functions do not work in `Array.prototype.*` functions.  Here are a [couple](https://medium.com/swlh/how-to-make-array-iterations-with-async-await-functions-in-node-js-e37504a64e8d) [articles](https://dev.to/puruvj/async-await-usage-and-pitfalls-in-array-prototype-map-and-chaining-3o6o) [about it](https://flaviocopes.com/javascript-async-await-array-map/) that will do a better job of show-and-tell than I can do here.

The way I tackle these thing is to ask the following: are the queries independent from each other or are they linked?  If independent, use `Promise.all()` and push your queries into an array.  If linked, use a traditional `for/while/do` loop.  In the second case, eslint may complain, but I'm not sure there is a better solution.

It _looks_ like these are independent, but they do modify the same tables, so let's just play it safe and use a traditional for loop:

```js
// you must have the `async` keyword on the top-level function
while (rows.length) {
  const row = rows.shift();

  const lots = await db.exec(query2, [row.label, db.bid(row.inventory_uuid), row.expiration_date]);
  const masterLotUuid = lots[0].uuid;

  const lotUuids = lots.reduce((list, elt) => {
    if (elt.uuid !== masterLotUuid) {
      list.push(elt.uuid);
    }
    return list;
  }, []);

  numLots = numLots + lotUuids.length - 1;

  await merge(/ * all your parameters */);
}
```

server/controllers/stock/lots.js
```diff
@@ -198,12 +201,80 @@ async function merge(req, res, next) {
...
+            const lotUuids = lots.reduce((list, elt) => {
+              if (elt.uuid !== masterLotUuid) {
+                list.push(elt.uuid);
+              }
+              return list;
+            }, []);
+            numLots = numLots + lotUuids.length - 1;
+
+            // Now merge this lot and its duplicates
+            merge({
```
jniles: Remember this is an asynchronous function, so you'll need `await` before it.

jniles: By the way, this wasn't obvious to me at first, but mixing Promises and async/await is totally fine.  In fact, async/await compiles to a Promise chain internally.

So, this works:
```js
async eleven() {

  // this variable will be 3
  const three = await Promise.resolve(3);
  
  // this will be 8
  const eight = await(Promise.resolve(5).then(five => five + three));
  
  // this will finally return 11
  return three + eight;
}
```

jniles: You can make cute things like `sleep()` functions this way:

```js
// sleeps n seconds
function sleep(n) {
  return new Promise((resolve /*, reject */) => {
    setTimeout(() => {
      resolve();
    }, n * 1000);
  });
}

// mix async/await with a promise function
(async () => {
  console.log('Started!', Date.now());
  await sleep(3);
  console.log('Ended!', Date.now());
})();
```

server/controllers/stock/lots.js
jniles: You would probably gain a lot by rewriting this as `async/await` in it's entirety.  If you want to do that in a pair coding session, I'm game.

server/controllers/stock/lots.js
```diff
@@ -127,7 +128,7 @@ function getCandidates(req, res, next) {
  * Returns all lots with the given label or matching field(s)
  * inventory_uuid, initial_quantity, entry_date, expiration_date
  *
- * TODO: After getting this working purge unneeded params
+ * TODO: After getting this working, purge unneeded params
```
jniles, line 131: 👍

server/config/routes.js
```diff
@@ -951,6 +951,7 @@ exports.configure = function configure(app) {
   app.put('/lots/:uuid', lots.update);
   app.get('/lots/:uuid/assignments/:depot_uuid', lots.assignments);
   app.post('/lots/:uuid/merge', lots.merge);
+  app.post('/lots/merge/auto', lots.autoMerge);
   app.get('/inventory/:uuid/lot_candidates', lots.getCandidates);
   app.get('/lots_dupes/:label?/:entry_date?/:expiration_date?/:initial_quantity?/:inventory_uuid?', lots.getDupes);
```
jniles, line 956: Let's discuss this on Skype sometime.  I actually think you can remove everything here and just leave:
```js
   app.get('/lots_dupes/', lots.getDupes); 
```

but I'm curious where the confusion is coming in.  Note that this may be outside the scope if this PR if it's not as straightforwards as I just mentioned.

server/controllers/stock/lots.js
```diff
@@ -198,12 +201,80 @@ async function merge(req, res, next) {
...
+              if (elt.uuid !== masterLotUuid) {
+                list.push(elt.uuid);
+              }
+              return list;
+            }, []);
+            numLots = numLots + lotUuids.length - 1;
+
+            // Now merge this lot and its duplicates
+            merge({
+              query : {},
```
jniles: You seem to be "faking" an HTTP request here (since `merge()` looks like a standard expressjs http middleware function).

I find it much simpler/more legible to simply move the "merge" logic into it's own function, then make two http interfaces that call it.

Basically, you can imagine functions:
```js
// this guy has the `db.transaction()` call and 90% of the code from the original `merge()` fn
function merge(masterLotUuid /* uuid */, duplicateLots /* array */) { /* ... */ }

// this guy wraps the above in a try/catch and breaks out the req.body/req.params 
// into the masterLotUuid and duplicateLots array
function mergeDuplicateLotsHTTP(req, res, next) { /* ... */ };

// this guy wraps the `merge()` and performs the relevant DB queries to find all duplicates for all lots.
function autoMergeDuplicateLotsHTTP(req, res, next) { /* ... */ }
```

Then it is the HTTP interfaces responsibility to:
  1) unpack the req.params/req.body to get the masterLotUuid and the duplicate lots
  2) do whatever preprocessing needed.
  3) wrap the call to `merge()` in try/catch or `.catch()` to handle errors.
  4) unpack whatever the return result is and send to appropriate status code to the client.
  
What do you think?

You can see this structure play out in our `patients` controller.  The [`find()` function](https://github.com/IMA-WorldHealth/bhima/blob/1819d8f7380f16590fce3fbbe3c900513d001c56/server/controllers/medical/patients/index.js#L742) was originally a `function (req, res, next) {}` expressjs middleware function.  But we realized we needed the same logic elsewhere, so we broke it out.  The HTTP interface for it is now called [`read()`](https://github.com/IMA-WorldHealth/bhima/blob/1819d8f7380f16590fce3fbbe3c900513d001c56/server/controllers/medical/patients/index.js#L850) and just does some basic logic switches and error handling.

jniles (commented):
@jmcameron I've left a lot of comments on this PR.  My understanding is that it isn't quite functional yet due to async/await issues, so I've tried to give some guidance there.

It might make sense to pair-program on this one, if you would like to.  Otherwise, I hope my comments are intelligible and give some idea of how to improve the code.

server/controllers/stock/lots.js
jmcameron: > It looks like you are trying to `filter` all lots that are the master lot out of the array. Let's do that! Pretty sure this will give you the same result with clearer syntax:

Actually, I'm trying to do two things at once:  Extract the lot uuid from the duplicate lots but only if they do not match the masterLotUuid.   The suggested code accumulates the lots, not the uuids (I tested it to verify).  So later I would have to do a map on the result.  The reduce allows me to do both at once.

jniles: Ah, I see that now.

server/controllers/stock/lots.js
```diff
@@ -166,42 +167,124 @@ function getDupes(req, res, next) {
...
+  `;
+  db.exec(query1, [])
+    .then((rows) => {
+      const numInventories = rows.length;
+      const numLots = rows.reduce((sum, row) => {
+        return sum + row.num_duplicates;
+      }, 0) - numInventories;
+      const dbPromises = [];
+      rows.forEach((row) => {
+        db.exec(query2, [row.label, db.bid(row.inventory_uuid), row.expiration_date])
```
jniles: This `db.exec()` is asynchronous, but this code doesn't treat it as such.  The `row.forEach()` is going to just keep spinning without ensuring the `db.exec()`s are done, either in parallel or in serial.  You really should account for this.

You're close with your `dbPromises` approach.  However, because you aren't waiting for these `db.exec()`s to return, you may careen right past it without actually waiting.

A thought exercise: imagine that each `db.exec()` here takes 15 seconds.  What will happens is that `rows.forEach()` will churn through the array before a single `.then()` is called.  That means that `dbPromises` will be empty after running through the `.forEach()` and the `Promise.all()` call on an empty array will successfully return nothing.

To prevent this, let's `return mergeLotsInternal(keepLotUuid, lotUuids))` at the end of the `.then()` call.  What does this do?  It links the `mergeLotsInternal` call into the promise chain.  To `node` it will look like you've just added another `.then()` to the chain.  After, let's push the _entire chain_ into the `dbPromises` array.  Effectively:
```js
rows.forEach(row => {
  const promise = db.exec(/* params */)
    .then((lots) => {
       // do all kinds of cool processing here
       return mergeLotsInternal(/* more params */);
    });
    
  // add the entire chain to the dbPromises array
  dbPromises.push(promise);
});

// now when we call Promise.all(), we are waiting for the entire promise chain to complete.  That means
// the db.exec() and every .then() after it, including the mergeLotsInternal() function
Promise.all(dbPromises)
  .then(() => { /* return http status code */ })
  // remember error handling!  This will handle errors for the entirety of the chain.
  .catch(err => next(err));
```

jniles (changes requested):
@jmcameron this looks much better!  Great work!

There is one remaining issue with asynchronous code that isn't being treated as such, and missing error handling for the promise array.  I've made those comments above.

server/controllers/stock/lots.js
jmcameron: Thanks!  I applied this fix and now it works!

server/controllers/stock/lots.js
jmcameron: Fixed with your suggestion.  Thanks!

server/controllers/stock/lots.js
jmcameron: Done

server/controllers/stock/lots.js
jmcameron: Thanks!

server/controllers/stock/lots.js
```diff
@@ -166,42 +167,125 @@ function getDupes(req, res, next) {
...
+              if (elt.uuid !== keepLotUuid) {
+                list.push(elt.uuid);
+              }
+              return list;
+            }, []);
+            return mergeLotsInternal(keepLotUuid, lotUuids);
+          });
+        dbPromises.push(promise);
+      });
+      Promise.all(dbPromises)
```
jniles: This promise doesn't have any error handling.  You can either:
 1) add "return" in front of the `Promise.all()` and move the `.then()` block into the parent promise chain OR
 2) add a `.catch()` to this chain.

I prefer the former, but either will work.

jniles (changes requested):

mbayopanda (commented):
Thank you @jmcameron for the feature, I have noticed some stuff that we can enhance or discuss : 
- In the module `Trouver des lots en double`, is it a way to check all duplicated lots once in the selected lot modal
- How could we merge lots which are not found by the algo (such as SD20-0039 and SD20-039) ? The proposal is to have in the lots registry a way of mergin them
- In the module `Trouver des lots en double`, when we click on `Fusionner automatiquement des lots` the feature doesn't merge auto all duplicated lots as expected.
- Add the expiration date in the lot label during a stock exit
- The account report doesn't match with the stock sheet report since the stock entry is made with the new entered cost of lots (this cost is shown in the receipt and it is inserted in the accounting) but in the stock sheet report we have cost of the existing lot as expected

In the accounting : 
image

In the stock sheet report : 
image

jmcameron:
>     * In the module `Trouver des lots en double`, is it a way to check all duplicated lots once in the selected lot modal

Good suggestion.   I'll add that.  --> DONE.

>     * How could we merge lots which are not found by the algo (such as SD20-0039 and SD20-039) ? The proposal is to have in the lots registry a way of mergin them

These lots can be merged using the action menu on each row.

>     * In the module `Trouver des lots en double`, when we click on `Fusionner automatiquement des lots` the feature doesn't merge auto all duplicated lots as expected.

It only merges the lots with exact matches of label, inventory_uuid, and expiration date.   Others will need to be merged manually since it is not obvious that the lots are actually for the same thing.

>     * Add the expiration date in the lot label during a stock exit

@mbayopanda I don't know what you mean by this.  Does this affect this PR?    I suggest you add an Issue for this.  --  @jniles pointed out this related issue:  https://github.com/IMA-WorldHealth/bhima/issues/5401.  I don't think this affects this PR.

>     * The account report doesn't match with the stock sheet report since the stock entry is made with the new entered cost of lots (this cost is shown in the receipt and it is inserted in the accounting) but in the stock sheet report we have cost of the existing lot as expected

I did not understand this problem.    How should we be handling the unit_price during lot merges?  I'll talk it over with @jniles to see what he thinks.  -- After chatting with @jniles he explained the unit_price column in the lot table is not actually used.  That information is only accessed via stock movements.   So differences in unit_cost can be ignored when merging lots.

server/controllers/stock/lots.js

jmcameron:
@jniles and @mbayopanda  :  Please take a look at the new tooltip for the "Automatically merge..." button and see what you think.  Also verify that the "Select all lots" checkbox works (from the Find Duplicate Lots action menus).  Look for one that has more than 2 duplicates to verify that all get selected.

@jniles : I did the promise fix you suggested, please take a look at it and verify.

@mbayopanda : Please check the French translations and suggest fixes or improvements.

jmcameron:
I've removed the "Do not merge" flag.  I think once the final wording tweaks are done, this PR is ready to merge.

client/src/modules/stock/lots-duplicates/lots-duplicates.js
```diff
@@ -124,6 +124,20 @@ function DuplicateLotsController(
...
 
+  // Call the server function to merge lots automatically
+  function autoMergeLots() {
+    vm.loading = true;
+    Lots.autoMerge()
+      .then((res) => {
+        const msg = $translate.instant('LOTS.MERGED_LOTS_AUTOMATICALLY', res);
+        load();
+        Notify.success(msg);
+        vm.loading = false;
```
jniles: Looks like you are going to have competition here between `vm.loading = false` and the first call of `load()` where `vm.loading = true`.

Can you move the `vm.loading = false` to above the `load()` call?

-- this is a non-blocking comment since I don't think it functionally makes a difference.

server/controllers/stock/lots.js
```diff
@@ -166,42 +167,130 @@ function getDupes(req, res, next) {
...
-      transaction.addQuery(deleteLot, [mergeLotUuid]);
-    });
-    await transaction.execute();
-    res.sendStatus(200);
-  } catch (error) {
-    next(error);
-  }
+  const transaction = db.transaction();
+
+  lotsToMerge.forEach(rawUuid => {
```
jniles, line 194: Good naming schemes here, by the way.  Very clean code. 👍

server/controllers/stock/lots.js
```diff
@@ -166,42 +167,130 @@ function getDupes(req, res, next) {
...
+ * This is a accomplished in two steps for each lot to merge:
+ *   1. Replace all references to the lot to be merged with
+ *      references to the lot to keep.
+ *   2. Delete the lot to be merged
+ */
+function merge(req, res, next) {
+  const { uuid } = req.params;
+  const lotsToMerge = req.body.lotsToMerge.map(db.bid);
+
+  mergeLotsInternal(uuid, lotsToMerge)
```
jniles, line 219: Perfect!

server/controllers/stock/lots.js
```diff
@@ -166,42 +167,130 @@ function getDupes(req, res, next) {
...
+              if (elt.uuid !== keepLotUuid) {
+                list.push(elt.uuid);
+              }
+              return list;
+            }, []);
+            return mergeLotsInternal(keepLotUuid, lotUuids);
+          });
+        dbPromises.push(promise);
+      });
+      return Promise.all(dbPromises);
```
jniles, line 287: Exactly.  All errors are now caught by the `catch(next)` statement below since they are returned to the promise chain.

jniles (approved):
Looks good to me!  Tested and works well.

client/src/modules/stock/lots-duplicates/lots-duplicates.js
jmcameron: Done. Rebased.

jmcameron:
@mbayopanda When you review this, if there are no problems, please go ahead and release it.   @jniles have been working through all the issues on it and I think it is ready to go. Thanks!

client/src/i18n/en/form.json
```diff
@@ -10,6 +10,8 @@
             "ADD_RUBRIC_CONFIGURATION" : "Add a Rubric configuration",
             "ADD_WEEKEND_CONFIGURATION": "Add Weekend Configuration",
             "AUTO_INVOICES_BREAKDOWN": "Automatic invoices Breakdown",
+            "AUTO_MERGE_LOTS": "Automatically Merge Exactly Matching Lots",
+            "AUTO_MERGE_LOTS_TOOLTIP": "Automatically merge all lots with exactly matching label, inventory_uuid, and expiration_date",
```
mbayopanda, line 14: 👍🏽 Exactly matching 😄 We are playing safe

mbayopanda (approved):
The Auto merge tool works as expected 👍🏽 

We can add filter in the `Trouver des lots en double` registry to help searching an item in the future

jmcameron:
Bors r+

PR #5404, opened 2021-02-19
Updated bhDateInterval to use placeholder labels

### LINKED ISSUE #5397: Extend bhDateInterval with default placeholder labels
Update the bhDateInterval component to support Adding "Start date" and "End date" placeholders in the two fields, respectively.  These should be done with optional parameters so custom values can be given by the user.

jmcameron:
Add start-date and end-date placeholder labels which default to FORMS.LABELS.START_DATE/END_DATE.  I think this helps make it easier for users to automatically understand what the form is asking for.   Here is what it looks like:
image

Note the circled placeholders. Once actual dates are selected, they disappear.  But when you clear the field they are restored.

**NOTES**  
  - This applies to all appearances of the bhDateInterval component.
  - Code using the bhDateInterval component can provide the 'startDatePlaceholder' and 'endDataPlaceholder' to override these labels.  These will be  translated so should be language tokens.

**TESTING**

- Use bhima_test or any DB.
- On the Stock > Lots page, bring up the Search dialog and scroll down to see the date interval fields.

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -712,6 +712,7 @@
             "STAFF_COST": "Staff",
             "STAFFING_BASE_INDICES" : "Basic indices for recruitment",
             "START": "Start",
+            "START_DATE": "Start date",
             "STATUS": "Status",
             "STOCK": "Stock",
             "SUBSIDIES": "Subsidies Applied",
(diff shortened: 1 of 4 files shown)


jniles (approved):
Looks good to me.

bors r+

PR #5412, opened 2021-02-23
Extend date picker to support disabling

### LINKED ISSUE #5388: Update bhDatePicker to allow it to be disabled
The bhDatePicker component should support a 'disabled' flag that will prevent it from being changed.  

Once implemented and PR https://github.com/IMA-WorldHealth/bhima/pull/5387 is merged,  the new disabling functionality should be used in client/src/modules/stock/entry/modals/lots.modal.js (near line 198) when changing the expiration date in onSelectLot() to disable the date field.

jmcameron:
This PR extends the bhDatePicker component to support disabling the field.   

Closes #5388 

To use this feature: when setting up the bh-date-picker component, add this parameter:

    disabled="obj.disabled"

Where "obj" is the input field for the date picker and 'disabled' is a variable in the object that evaluates to **true** when the field should be disabled.  This should be backwards compatible.

This is currently implemented for the Stock entry page (when selecting/creating lots).  You can see examples of its use in
- client/src/modules/stock/entry/modals/lots.modal.js (search for 'disabled')
- client/src/modules/stock/entry/modals/templates/lot.expiration.tmpl.html

**TESTING**
- Use any database (eg, bhima_test)
- Use Stock > Lots to find an inventory item that has lots that are not expired  (for bhima_test, try inventory Item: Vitamines B1+B6+B12
- Go to Stock > Entry
   - Use integration
   - Do "Add" and select the inventory item from the previous step
   - Click on the "Lots" field to get the lot selection/creation modal
   - On the first row, notice that the expiration date can be changed.
   - Select one of the existing lots using the lot selection typeahead
   - Notice that the expiration date has been overwriten and cannot be modified.

### DIFF
--- a/client/src/js/components/bhDatePicker.js
+++ b/client/src/js/components/bhDatePicker.js
@@ -7,6 +7,7 @@ angular.module('bhima.components')
       onChange : '&', // use a callback to notify for changes
       format   : '<?',
       mode     : '@?', // will this ever change?  If so, we can use '<'
+      disabled : '<?', // Optional flag to disable this field
       required : '<?',
     },
   });
@@ -43,6 +44,10 @@ function DatePickerController(Modal, bhConstants) {
   }
 
   function open() {
+    if (vm.disabled) {
+      // Do not open the date picker popup if the field is disabled
+      return;
+    }
     openDatePicker({ mode : vm.mode })
       .then(date => {
         // notify the parent controller of a date change via a callback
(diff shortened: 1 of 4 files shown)


jniles:
@jmcameron can you do a rebase?  Looks like you made the same blunder I did earlier where old commits show up in the PR.  In this case, it is very difficult to review though because they outweigh your changes and I can't figure out the extent of your changes.

jmcameron:
> @jmcameron can you do a rebase? Looks like you made the same blunder I did earlier where old commits show up in the PR. In this case, it is very difficult to review though because they outweigh your changes and I can't figure out the extent of your changes.

I did not need to refactor.  I just need to change the base for the PR from master to refactor-stock-management!  @jniles It should be ready now.

jniles (approved):
LGTM.

bors r+

PR #5420, opened 2021-02-23
Feature aggregate consumption

### LINKED ISSUE #4850: Allow users to record aggregate consumption information for a depot
Separate and apart from the designs proposed in the https://github.com/IMA-WorldHealth/bhima/issues/4848 and https://github.com/IMA-WorldHealth/bhima/issues/4811, we need a way to put in "aggregate" data into BHIMA.  This would be for the "Stock Management in Health Zones" push.

This information would be put in by depot for a chosen period.  Basically, the user would chose a period, choose a depot, and be presented with a table of (consumable) medicines that they can use to put in the consumption during that period as well as the number of days of stock out.

The database schema would look like this
Period ID | Inventory UUID | Lot Number | Consumption | Days of Stock Out
-- | -- | -- | --:| --:
202003 | Quinine 500MG | LA2030 | 700 | 2
202003 | Quinine 500MG | XXT2093 | 100 | --
202003 | Amoxy 250MG | NF33001A | 200 | 3

The interface would look something like this:
Aggregate Stock Consumption Page

Basically, the user would put in these two columns and hit "submit" and that would be it.

**How does this fit into the regular design of BHIMA?**

As @jmcameron pointed out, we actually can check this table for consumption information and add it to our regular algorithms (obtained either by https://github.com/IMA-WorldHealth/bhima/issues/4848 orhttps://github.com/IMA-WorldHealth/bhima/issues/4811).  This would make it much easier to design and test.  In a regular hospital setting, this will never be used and will always give 0.

lomamech:
````
You should reload database and run the yarn migrate command
````
closes #4850

### CODE REVIEW

**Comment by jmcameron** [comment]
Chris, could you add a section in the main description suggestions on how to test this PR?  I suggest adding a note that you need to reload the datbase and do 'yarn migrate' to set the new page.
Thanks!

**Comment by jmcameron** [comment]
It looks like the migration file defines the unit entry for the aggregated page twice.  Needs fixing!

**Review by jmcameron — COMMENTED** [review]
I did a crude test (with the IMCK db).  But when I clicked on the [Submit] button, nothing happened and I see this error:

`
vendor.min.js:15710 TypeError: Cannot read property 'length' of undefined
    at getCheckedNodes (bhima.min.js:31917)
    at bhCheckboxTreeController.$ctrl.$onChanges (bhima.min.js:31839)
`

This is on line 132 of bhCheckboxTree.js

Also noticed a problem in the migration file: It looks like the migration file defines the unit entry for the aggregated page twice.

**Comment by lomamech** [comment]

**Comment by lomamech** [comment]
> It looks like the migration file defines the unit entry for the aggregated page twice. Needs fixing!

I just fixed it

File: client/src/i18n/fr/report.json [hunk 1]
```diff
@@ -244,7 +244,9 @@
       "STOCKS" : "Stocks",
       "EXIT_REPORT" : "Rapport des sorties de stocks",
       "EXIT_REPORT_DESCRIPTION" : "Ce rapport affiche les sorties de stock qui ont eu lieu dans different dépôts",
-      "INCLUDE_INFORMATION" : "Information à inclure",
+      "INCLUDE_AGGREGATE_CONSUMPTION": "Inclure les consommations agrégées",
+	  "INCLUDE_INFORMATION" : "Information à inclure",
+	  "INCLUDE_INFORMATION" : "Information à inclure",
```
jniles: Looks like this is repeated twice.  Could you remove the duplicate?

client/src/i18n/fr/stock.json
```diff
@@ -221,6 +237,7 @@
     "ENABLE_STRICT_DEPOT_DISTRIBUTION_HELP_TEXT" : "Limite les dépôts qui peuvent recevoir du stock à partir d'un dépôt donné; lors de la sortie de stock vers un dépôt, seul les dépôts autorisés peuvent s'afficher pour le dépôt donné afin de recevoir du stock"
   },
   "STOCK_FLUX" : {
+    "AGGREGATE_CONSUMPTION" : "Consommation agrégée",
```
jniles, line 245: I agree with the decision to make a new flux_type.  Note that we'll need to update all our procedures that create CMM to say this is a "consumption" type, no matter the depot type.

server/controllers/stock/index.js
```diff
@@ -975,3 +977,221 @@ function getStockTransfers(req, res, next) {
...
+async function createAggregatedConsumption(req, res, next) {
+  try {
+    const movement = req.body;
+
+    if (!movement.depot_uuid) {
+      throw new Error('No defined depot');
+    }
+
+    const checkInvalid = movement.lots
+      .filter(l => ((l.quantity_consumed + l.quantity_lost) > l.oldQuantity));
```
jniles, line 995: I think this check should be `>=`, not `>`.

server/controllers/stock/index.js
```diff
@@ -975,3 +977,221 @@ function getStockTransfers(req, res, next) {
...
+
+    if (checkInvalid.length) {
+      throw new Error('Invalid data');
```
jniles: We could improve this error message:
```js
throw new Error('Invalid data!  Some lots have consumed or lost more stock than they originally had.'); 
```

client/src/modules/stock/aggregated_consumption/modals/consumption_lot.modal.html
```diff
@@ -0,0 +1,89 @@
...
+    <div style="margin-bottom: 5px;">
+      <div class="checkbox">
+        <label>
+          <input 
+            id="enableFastInsert" 
+            ng-model="$ctrl.enableFastInsert" 
+            type="checkbox"
+            ng-true-value="1"
+            ng-false-value="0"> 
+          <span translate>LOTS.ENABLE_FAST_INSERT</span>
```
jniles: This checkbox doesn't make sense here.  We aren't dealing with lots.  We either need to remove it, or change the text and potentially behavior to reflect what we are really doing.

client/src/modules/stock/aggregated_consumption/templates/days_stock_out.tmpl.html
```diff
@@ -0,0 +1,12 @@
+<div class="ui-grid-cell-contents">
+  <input
+    ng-if="row.groupHeader"
+    name="days_stock_out"
+    id="row.treeNode.children[0].row.entity.inventory_uuid"
```
jniles, line 5: Do you need this `id` field?  What is it used for?

client/src/modules/stock/entry/modals/AggregateConsumptionModalForm.service.js
```diff
@@ -0,0 +1,169 @@
...
+
+AggregateConsumptionModalForm.$inject = ['uuid'];
+
+/**
+ * @function AggregateConsumptionModalForm
+ *
+ * @description
+ * This function allows you to create a new instance to record
+ * the aggregate consumption of a batch by date. This function makes it possible to better track
+ * consumption to quickly detect periods of out of stock
```
jniles, line 12: Nice description!

client/src/modules/stock/aggregated_consumption/templates/days_stock_out.tmpl.html
```diff
@@ -0,0 +1,12 @@
+<div class="ui-grid-cell-contents">
+  <input
+    ng-if="row.groupHeader"
+    name="days_stock_out"
```
jniles, line 4: Let's make this default to "0".

server/controllers/stock/index.js
```diff
@@ -975,3 +977,221 @@ function getStockTransfers(req, res, next) {
...
+            uuid : db.bid(uuid()),
+            lot_uuid : db.bid(item.uuid),
+            depot_uuid : db.bid(movement.depot_uuid),
+            document_uuid : db.bid(lossUuid),
+            quantity : elt.quantity_lost,
+            unit_cost : item.unit_cost,
+            date : new Date(eltDate),
+            is_exit : 1,
+            flux_id : core.flux.TO_LOSS,
+            description : movement.description,
```
jniles: We should do what we do with the other stock movements and template in our own description as well.  Something like "Aggregated Stock Adjustment for Period X by User Y"

server/controllers/stock/reports/stock/exit_aggregate_consumption_receipt.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+async function stockExitAggregateConsumptionReceipt(documentUuid, session, options) {
+  const data = {};
+  const optionReport = _.extend(options, { filename : 'STOCK.RECEIPTS.AGGREGATE_CONSUMPTION' });
+
+  // set up the report with report manager
+  const report = new ReportManager(STOCK_AGGREGATE_CONSUMPTION_TEMPLATE, session, optionReport);
+
+  /**
+   * TODO consider the donor also in a donation transaction
```
jniles: Do you need this comment?  It looks irrelevant.

server/models/schema.sql
```diff
@@ -834,8 +834,8 @@ CREATE TABLE `inventory_type` (
 DROP TABLE IF EXISTS `inventory_unit`;
 CREATE TABLE `inventory_unit` (
   `id` SMALLINT(5) UNSIGNED NOT NULL AUTO_INCREMENT,
-  `abbr` VARCHAR(10) NOT NULL,
-  `text` VARCHAR(30) NOT NULL,
+  `abbr` VARCHAR(100) NOT NULL,
```
jniles: 👍

test/end-to-end/stock/stock.aggregate_consumption.page.js
```diff
@@ -0,0 +1,164 @@
...
+  page.setDescription = description => {
+    return FU.input('StockCtrl.movement.description', description);
+  };
+
+  /**
+   * @method setHeaderValue
+   */
+  page.setHeaderValue = async function setHeaderValue(rowIndex, columnIndex, value) {
+    const cell = await GU.getCell(gridId, rowIndex, columnIndex);
+    const inputOutStock = cell.element(by.id('row.treeNode.children[0].row.entity.inventory_uuid'));
```
jniles, line 37: This: `by.id('row.treeNode.children[0].row.entity.inventory_uuid')`  selector makes me think something is broken.  Shouldn't it be replaced by the inventory uuid?

I also suggest that you use the inventory text instead of the `id=inventory_uuid`.  Set something like `data-inventory-label="{{row.treeNode.children[0].row.entity.inventory_text}}" which will let you change this selector to:

```js
const inputOutStock = cell.$('[data-inventory-label="Amoxy ..."]')); 
```
which is much easier to read.

test/end-to-end/stock/stock.aggregate_consumption.page.js
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * @method setDetailled
+   */
+  page.setDetailled = async function setDetailled(rowIndex, columnIndex) {
+    const getCell = await GU.getCell(gridId, rowIndex, columnIndex);
+    await getCell.element(by.id('aggregated_details')).click();
+  };
+
+  page.setLotsDetailled = async function setLotsDetailled(lotsArray) {
```
jniles: `detailed` only has a single `l` in English :)

test/end-to-end/stock/stock.aggregate_consumption.page.js
```diff
@@ -0,0 +1,164 @@
...
+    const quantityConsumptionCell4 = await GU.getCell(modalGridId, 4, 2);
+    const quantityLostCell4 = await GU.getCell(modalGridId, 4, 3);
+    await components.datePicker.set(lotsArray[4].date, dateCell4);
+    await FU.input('row.entity.quantity_consumed', lotsArray[4].quantity_consumed, quantityConsumptionCell4);
+    await FU.input('row.entity.quantity_lost', lotsArray[4].quantity_lost, quantityLostCell4);
+
+    FU.modal.submit();
+  };
+
+  page.setLots2Detailled = async function setLots2Detailled(lotsArray) {
```
jniles: What is `setLots2`?  What is the actual action that is being performed here?

test/end-to-end/stock/stock.aggregate_consumption.page.js
```diff
@@ -0,0 +1,164 @@
...
+  page.setFiscalPeriod = async (fiscalYear, period) => {
+    await components.fiscalYearPeriodSelect.set(fiscalYear, period);
+  };
+
+  // depot select
+  page.setDepot = depot => {
+    return components.depotSelect.set(depot, 'depot-supplier');
+  };
+
+  page.setRows = async number => {
```
jniles: Let's call this `addRows` instead of `setRows`, since we are clicking the "add rows" button on the page.

test/end-to-end/stock/stock.aggregate_consumption.page.js
```diff
@@ -0,0 +1,164 @@
...
+
+  // change current depot
+  page.changeDepot = SharedStockPage.setDepot;
+
+  // the grid id
+  page.gridId = gridId;
+  page.modalGridId = modalGridId;
+
+  // create modal
+  page.showCreateModal = async isNewRequisition => {
```
jniles: Do we need the `isNewRequisition` flag here?

I think you may have copied this from somewhere.  Can you remove the methods you don't actually use from this page?

test/end-to-end/stock/stock.aggregate_consumption.js
```diff
@@ -0,0 +1,147 @@
...
+    const getLastDays = new Date(getMovementYear, getMovementMonth + 1, 0);
+
+    const month = ['Jan', 'Fév', 'Mars', 'Avr', 'Mai', 'Juin', 'Juill', 'Août', 'Sept', 'Oct', 'Nov', 'Déc'];
+
+    const fiscalYearLabel = `Fiscal Year ${getMovementYear}`;
+
+    await page.setFiscalPeriod(fiscalYearLabel, month[getMovementMonth]);
+    await page.setDescription(`Aggregate consumption from current depot ${DEPOT_TERTIAIRE}`);
+
+    await page.setHeaderValue(0, 7, 0);
```
jniles: In general, you should try not to refer to the index in the grid in your tests.  Instead, refer to the name of the row.  Something like
```js
await page.setHeaderValue("Quinine", 7);
```
or even better:
```js
const quinine = await page.getInventoryGroup('Quinine');
await quinine.setDaysOfStockOut(7);
await quinine.setLotQuantityLost("Lot X", 5);
await quinine.setLotQuantityConsumed("Lot X", 7);
```

This is much closer to what the user is actually doing.  It also means that other programmers can add lots/inventory items without needing to rewrite these tests.

test/end-to-end/stock/stock.aggregate_consumption.js
```diff
@@ -0,0 +1,147 @@
...
+  });
+
+  it(`Create a complexe aggregate consumption on current depot ${DEPOT_TERTIAIRE}`, async () => {
+    const getMovementDate = moment(new Date(), 'YYYY-MM-DD').subtract(60, 'days');
+    const getMovementMonth = moment(getMovementDate).month();
+    const getMovementYear = moment(getMovementDate).year();
+
+    const getLastDays = new Date(getMovementYear, getMovementMonth + 1, 0);
+
+    const month = ['Jan', 'Fév', 'Mars', 'Avr', 'Mai', 'Juin', 'Juill', 'Août', 'Sept', 'Oct', 'Nov', 'Déc'];
```
jniles: You use these months a couple of times. Can you move them to be one of the first parts in the `describe` block?  That way, each `it()` will share them.

client/src/modules/stock/aggregated_consumption/modals/consumption_lot.modal.js
```diff
@@ -0,0 +1,206 @@
...
+  'appcache', '$uibModalInstance', 'uiGridConstants', 'data',
+  'SessionService', 'bhConstants', 'AggregateConsumptionModalForm', 'focus',
+];
+
+function StockConsumptionLotsModalController(
+  AppCache, Instance, uiGridConstants, Data, Session, bhConstants, AggregateForm, Focus,
+) {
+  const vm = this;
+
+  const cache = new AppCache('StockEntryModal');
```
jniles: I think this `AppCache` needs to be updated.

client/src/modules/stock/aggregated_consumption/templates/lot_movement_date.tmpl.html
```diff
@@ -0,0 +1,7 @@
+<div class="ui-grid-cell-contents">
+  <bh-date-picker
```
jniles, line 2: I think this datepicker should have `min` and `max` dates within the period, right?

client/src/modules/stock/aggregated_consumption/modals/consumption_lot.modal.js
```diff
@@ -0,0 +1,206 @@
...
+    type : 'number',
+    width : 150,
+    displayName : 'STOCK.QUANTITY_LOST',
+    headerCellFilter : 'translate',
+    aggregationType : uiGridConstants.aggregationTypes.sum,
+    aggregationHideLabel : true,
+    footerCellClass : 'text-right',
+    cellTemplate : 'modules/stock/aggregated_consumption/templates/lost.tmpl.html',
+  }, {
+    field : 'actions',
```
jniles, line 89: It would be good to add `width: "10"` here, to limit the width of this field:

image

jniles (commented):
@lomamech this represents a lot of work!  It's a very complicated module, and you've done an admirable job implementing it.  Well done!

A couple of improvements I would make.

If there are no days of stock out, I wouldn't offer the detailed consumption modal.  Specifically, I would hide these buttons:
image

It's true that maybe someone could make a RUMER-like consumption report, but realistically no one will do that.

Similar to the above comment, we also should hide the "details" link if both values are 0:
image

There appears to be two binding problems on the receipt modal and the receipt itself:
image


Should we make the user choose a start date and end date on the `consumption_lot` modal instead of just a single date?  If the idea is that they can show multiple stock outs using it, then they should be able to show the start date, end date, and quantity consumed and lost over that period, no?

lomamech:
> It's true that maybe someone could make a RUMER-like consumption report, but realistically no one will do that.
> 
> Similar to the above comment, we also should hide the "details" link if both values are 0:

well noted

lomamech:
I agree well note

server/controllers/stock/index.js
lomamech: Here I check the invalid situation, and this case is very unlikely because on the client side prevents the validation of an invalid form

test/end-to-end/stock/stock.aggregate_consumption.js
jniles: By the way, if you need help with this, I can assist.  I know my descriptions aren't always the best 👍

server/controllers/stock/index.js
jniles: Oh!  You are right.  I misread this.  Thanks for the clarification. 👍

client/src/modules/stock/aggregated_consumption/templates/days_stock_out.tmpl.html
lomamech: Only during End to End test
````
  page.setHeaderValue = async function setHeaderValue(rowIndex, columnIndex, value) {
    const cell = await GU.getCell(gridId, rowIndex, columnIndex);
    const inputOutStock = cell.element(by.id('row.treeNode.children[0].row.entity.inventory_uuid'));
    inputOutStock.sendKeys(value);
  };
````

File: client/src/modules/stock/aggregated_consumption/templates/days_stock_out.tmpl.html (hunk 8 above)
Comment (lomamech) [reply client/src/modules/stock/aggregated_consumption/templates/days_stock_out.tmpl.html:4]: I wanted the user to write really zero instead of using a default value, this will prevent him from making the mistake of forgetting

File: test/end-to-end/stock/stock.aggregate_consumption.page.js (hunk 12 above)
Comment (lomamech) [reply test/end-to-end/stock/stock.aggregate_consumption.page.js:37]: If I made this modification, I will also be forced to modify the server part to do the reverse operation to find the Uuid from the name of the inventory

File: test/end-to-end/stock/stock.aggregate_consumption.page.js (hunk 14 above)
Comment (lomamech) [reply test/end-to-end/stock/stock.aggregate_consumption.page.js]: image
It's just to fill the modal of consumption detailed by date during E2E test

**Comment by jniles** [comment]
@lomamech let us know when this is ready for another review.

**Comment by lomamech** [comment]
> @lomamech let us know when this is ready for another review.

@jniles  I just finished it took me longer than expected, for some recommendation I think it will be preferable that it be settled in another Issue and Pull Request

````
- In general, you should try not to refer to the index in the grid in your tests. Instead, refer to the name of the row. Something like 
  -> await page.setHeaderValue("Quinine", 7);
- I think this datepicker should have min and max dates within the period, right?
````

File: server/controllers/stock/index.js [hunk 22]
```diff
@@ -65,6 +65,10 @@ async function createStock(req, res, next) {
 
     const period = await Fiscal.lookupFiscalYearByDate(params.date);
     const periodId = period.id;
+
+    console.log('FISCAL_YEAR: _ID');
```
jniles: It looks like you forgot some console.logs() here.

Curious that hound didn't catch them.

server/controllers/stock/index.js
```diff
@@ -994,16 +998,28 @@ async function createAggregatedConsumption(req, res, next) {
       .filter(l => ((l.quantity_consumed + l.quantity_lost) > l.oldQuantity));
 
     if (checkInvalid.length) {
-      throw new Error('Invalid data');
+      throw new Error('Invalid data!  Some lots have consumed or lost more stock than they originally had.');
```
jniles, line 998: Beautiful 👍

jniles (commented):
It looks like you did address the datepicker issue 👍 

Let me give it a test locally and see if the functionality works.  If so, I'll be happy to merge it.

I think the integration tests will fail, but we've put in fixes on the `refactor-stock-management` branch.  No need to work on them here.

server/controllers/stock/index.js
lomamech: It's Outdated

client/src/modules/stock/aggregated_consumption/modals/consumption_lot.modal.js
```diff
@@ -0,0 +1,212 @@
...
+      } else {
+        // add a new row
+        const newLotRow = vm.form.addItem();
+        // set the focus on the new row
+        Focus(newLotRow.identifier);
+      }
+    }
+  }
+
+  function getFirstEmptyLot() {
```
jniles, line 145: This is a great function name, by the way.  Super clear 👍

client/src/modules/stock/aggregated_consumption/modals/consumption_lot.modal.js
```diff
@@ -0,0 +1,212 @@
...
+   * @method onLotBlur
+   *
+   * @description
+   * if the fast insert option is enable do this :
+   * - add new row automatically on blur
+   * - set the focus in the new row
+   * @param {string} rowLot the row.entity.lot string
+   */
+  function onLotBlur(rowLot) {
+    if (vm.enableFastInsert && rowLot) {
```
jniles, line 129: Do you ever use `enableFastInsert`?  I assume this code must be copied from somewhere.  We should probably remove anything that isn't critical for these modules.

This can be done after merge.

client/src/modules/stock/entry/modals/AggregateConsumptionModalForm.service.js
```diff
@@ -0,0 +1,200 @@
...
+  };
+
+  /**
+   * @function totalConsumed
+   *
+   * @description
+   * This function computes the total quantity consumed in the form.
+   */
+  StockForm.prototype.totalConsumed = function totalConsumed() {
+    return this.rows.reduce((sum, row) => sum + row.quantity_consumed, 0);
```
jniles, line 82: Cool! 👍

server/controllers/stock/reports/stock/exit/exitAggregateConsumption.js
```diff
@@ -0,0 +1,39 @@
+const db = require('../../../../../lib/db');
+
+const IS_EXIT = 1;
+const EXIT_AGGREGATED_CONSUMPTION_ID = 16;
+
+/**
+ * @function fetch
+ * @description fetch stock exit for patient
```
jniles, line 8: I assume this description is out of date.  It should be updated.

jniles (approved):
@lomamech this is a much better step in the right direction.  I can see it coming together.

There are quite a few bugs that I discovered.  I don't think ti should block this - if we merge this into master, the whole team can dedicate their resources to helping fix them.

I'll list them here:

1. After setting the "details", you can update the consumed quantities.  I think the system will use the "detailed" version, and this can cause negative numbers!
2. It isn't entirely clear when the "details" link should appear.  This is partly my fault - I suggested you hide it.  I wonder if there is a way to disable it, and provide a tooltip to the user saying what needs to be done to enable it.
3. The validation of dates within the period work well, but it should limit the date picker to only those dates.  I screwed this up a couple of times.

Ultimately, this is a huge amount of work.  I'm really impressed and you should be proud of the work you've accomplished.

I'm going to manually begin the merging process.

PR #5427, opened 2021-02-24
Fixed import stock issue

### LINKED ISSUE #5426: Problems with Stock Import
When stock is imported successfully, the Stock Lots page is displayed with a filter for the depot that received the import.  Unfortunately, the UUID is shown for the depot in the filter label, not the depot name (in 'master').
image
Note that this filter label works correctly in the stock lots registry when selected by the [Search] filter there.

Also the example import file is broken (includes double-quotes within double-quotes, quoted numbers, etc.  These issues will be addressed by one of the fixes in the refactor-stock-management branch.

jmcameron:
Fixes import stock problem in Refactor Stock Management branch.

Note that Issue https://github.com/IMA-WorldHealth/bhima/issues/5426 still remains but can be dealt with after merging back to master.

### DIFF
--- a/server/controllers/stock/import.js
+++ b/server/controllers/stock/import.js
@@ -82,7 +82,7 @@ async function importStock(req, res, next) {
     });
 
     const isExit = 0;
-    const postingParams = [documentUuid, isExit, req.session.project.id, req.session.enterprise.currency_id];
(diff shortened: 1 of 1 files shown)


jniles (approved):
Tested and confirmed working.

PR #5432, opened 2021-02-25
fix(stock): make "at risk of expiration" flag work

### LINKED ISSUE #5422: Improvements to the "Articles in Stock" registry

jniles:
This fix restores the functionality of the "at risk of expiration" flag
on the articles in stock registry.  It does not affect the API, just the
internal calculations.

In addition, we also changed the behavior of the articles in stock
action dropdown menu so the link to the stock lots now filters on the
depot and time period in question.

image

Closes #5422.

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -1,4 +1,5 @@
-{ "STOCK" : {
+{
+  "STOCK" : {
     "ALERT" : "Alert",
     "ADJUSTMENT"      : "Adjustment",
     "ADJUSTMENT_TYPE" : "Adjustment Type",
--- a/client/src/modules/stock/StockFilterer.service.js
+++ b/client/src/modules/stock/StockFilterer.service.js
@@ -33,7 +33,7 @@ function StockFiltererService(Filters, AppCache, $httpParamSerializer, Languages
     { key : 'is_expiry_risk', label : 'STOCK.STATUS.IS_IN_RISK_OF_EXPIRATION' },
     { key : 'tag_uuid', label : 'TAG.LABEL' },
     { key : 'tags', label : 'TAG.LABEL' },
-    { key : 'show_only_risky', label : 'STOCK.LOTS.SHOW_ONLY_RISKY' },
+    { key : 'show_only_risky', label : 'LOTS.SHOW_ONLY_RISKY_LOTS' },
     { key : 'stock_requisition_uuid', label : 'FORM.LABELS.REQUISITION_REFERENCE' },
     {
       key : 'dateFrom', label : 'FORM.LABELS.DATE', comparitor : '>', valueFilter : 'date',
--- a/client/src/modules/stock/inventories/registry.js
+++ b/client/src/modules/stock/inventories/registry.js
@@ -196,6 +196,7 @@ function StockInventoriesController(
   }
 
   function setStatusFlag(item) {
+
     item.noAlert = !item.hasRiskyLots && !item.hasNearExpireLots && !item.hasExpiredLots;
     item.alert = item.hasExpiredLots;
     item.warning = !item.hasExpiredLots && (item.hasNearExpireLots || item.hasRiskyLots);
--- a/client/src/modules/stock/inventories/templates/inventory.action.html
+++ b/client/src/modules/stock/inventories/templates/inventory.action.html
@@ -31,6 +31,8 @@
       <a data-method="stock-lots" href
         ui-sref="stockLots({
           filters : [
+            { key : 'period', value : 'allTime', cacheable: false },
+            { key : 'depot_uuid', value : row.entity.depot_uuid, displayValue : row.entity.depot_text, cacheable: false},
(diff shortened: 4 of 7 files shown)


jmcameron (commented):
When I tested this with bhima_test and selected the "Show Only Inventories with Risky Lots", I noticed that the filter tag did not get translated:
image

Then I tested with the IMCK db (downloaded 2021-01-08), did yarn migrate, and CALL zRecomputeStockMovementStatus().

Again, it found no records and the filter tag is not translated.  
ALSO when I go back to the search filters, the flag for "Show only inventories with risky lots" already clear (not being cached).

BUT if I clear the filter and search for Inventory Item 'Vitamin B6..." (code 100118), it DOES show as at risk of expiration.
image

So there are still some problems here.

jmcameron:
Here is the fix for the untranslated filter label:

<pre>
diff --git a/client/src/modules/stock/StockFilterer.service.js b/client/src/modules/stock/StockFilterer.service.js
index bf103af4a..f06d26b1a 100644
--- a/client/src/modules/stock/StockFilterer.service.js
+++ b/client/src/modules/stock/StockFilterer.service.js
@@ -33,7 +33,7 @@ function StockFiltererService(Filters, AppCache, $httpParamSerializer, Languages
     { key : 'is_expiry_risk', label : 'STOCK.STATUS.IS_IN_RISK_OF_EXPIRATION' },
     { key : 'tag_uuid', label : 'TAG.LABEL' },
     { key : 'tags', label : 'TAG.LABEL' },
-    { key : 'show_only_risky', label : 'STOCK.LOTS.SHOW_ONLY_RISKY' },
+    { key : 'show_only_risky', label : 'LOTS.SHOW_ONLY_RISKY_LOTS' },
     { key : 'stock_requisition_uuid', label : 'FORM.LABELS.REQUISITION_REFERENCE' },
     {
       key : 'dateFrom', label : 'FORM.LABELS.DATE', compa
</pre>

jniles:
@jmcameron thanks for the review!  I didn't test the search functionality - great catch.  I'll look more into it.  And thank you for the translation fix!

jniles:
@jmcameron something is amiss.  The filter seems to work fine on my machine:

image

I noticed in the first screenshot, you don't have any modules/units in your tree.  This is typically a sign that you have changed out the backend server without logging out first, or opening a new tab.  Your session would then get screwed up (you would be implicitly saving your permissions from one server and trying to use them on the other).  If you ever get in that state, just open a new tab.

Can you retest with the IMCK database?

For the parameters of stock, I'm using:
  - the MSH algorithm
  - 6 months of stock for the AMC/CMM calculation
  - a minimum of 2 months of security stock required

With that, I get the screenshot above - at least a full page of risky lots.

-----

Here is another way to test with the test database:

  1. Build a fresh copy (`yarn test:integration`) of the database using the integration test suite.
  2. Refresh your application (hard refresh is best).
  2. You should have four lines in your "Articles in Stock Registry", like this:
  
image

  3. Check that the filter works by filtering on the risky lots.  You should now only have two rows.

jmcameron:
I wonder if the script generated by 'yarn migrate' should include 'CALL zRecomputeStockMovementStatus();' at its end?

jmcameron:
> I noticed in the first screenshot, you don't have any modules/units in your tree. This is typically a sign that you have changed out the backend server without logging out first, or opening a new tab. Your session would then get screwed up (you would be implicitly saving your permissions from one server and trying to use them on the other). If you ever get in that state, just open a new tab.
> 
> Can you retest with the IMCK database?
> 

I reloaded everything and now it works like you suggested.   There a few differences in the items listed but I'm sure that is because I used the IMCK database from a different date.

BTW, I did notice one thing.  I think the wording could be improved on the CMM analysis pop-up regarding the date the stock will run out at the current (CMM) rate:
image
It seems like "Next Stock Out" is a bit vague.  I suggest something clearer such as "Stock will be exhausted on" (or something similar).

jmcameron:
I also tested with bhima_test, as you suggested, and it worked exactly like you said it would.
Looks good to me!

jmcameron (approved):
Based on the updates, my second round of tests, and a quick review of the code, it looks good to me.

jniles:
Well, looks like we have a month of 28 days.  I'll do the fix here and release.

Thanks @jmcameron!

jniles:
Passes!

bors r+

PR #5433, opened 2021-02-25
Fix issue with stock import depot label

jmcameron:
This fixes a cosmetic problem with Stock Import: Once a successful stock import is performed, the result is shown in Stock Lots.  But the Depot filter was shown as the UUID. This PR fixes that issue.   

Note that this is a fix for the refactor-stock-management branch (but could be applied to 'master' after merging.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5426

TESTING

- Use any DB
- Perform a stock import (example file attached, rename to .csv).
- Notice that the depot label is now correct.  See the screenshot below:

image

[test.csv,txt](https://github.com/IMA-WorldHealth/bhima/files/6045090/test.csv.txt)

### DIFF
--- a/client/src/modules/stock/import/stockImport.js
+++ b/client/src/modules/stock/import/stockImport.js
@@ -63,7 +63,7 @@ function StockImportController(
     function handleSuccess() {
       Notify.success('STOCK.IMPORT.UPLOAD_SUCCESS');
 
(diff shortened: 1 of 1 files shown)


jniles (approved):
Looks good to me.

bors r+

PR #5437, opened 2021-02-26
Enhance chart of stock movement report

mbayopanda:
This PR is related to #5416 

Add date range component and the depot select component : 

image

Render the chart by movement flux

image


client/src/i18n/en/report.json
```diff
@@ -158,8 +158,8 @@
     },
 
     "STOCK_MOVEMENT_REPORT" : {
-      "TITLE" : "Stock Movement Graphical report",
-      "DESCRIPTION" : ""
+      "TITLE" : "Stock Movement Graph",
+      "DESCRIPTION" : "This report displays in graphic form the various stock movements made during a period"
```
jniles: This would be better written as "This report displays a graph of stock movements made during a period".

"in graphic form" in English typically means something is very detailed or very vulgar. 😅

server/lib/chart.js
```diff
@@ -120,73 +121,110 @@ function barChart(params) {
...
-          data: {
-            labels: ${JSON.stringify(chart.labels)},
-            datasets: series,
-          },
+          // The type of chart we want to create
+          type: '${type}',
+
+          // The data structure for our dataset
+          // https://www.chartjs.org/docs/master/general/data-structures/
+          data: ${JSON.stringify(data)},
```
jniles, line 223: Cool 👍

server/models/migrations/next/migrate.sql
```diff
@@ -103,3 +103,10 @@ INSERT INTO `discharge_type` (`id`, `label`) VALUES
  */
 CALL add_column_if_missing('depot', 'description', 'TEXT DEFAULT NULL AFTER `text`');
 CALL add_column_if_missing('depot', 'dhis2_uid', 'VARCHAR(150) DEFAULT NULL AFTER `parent_uuid`');
+
+/**
+  * @author: mbayopanda
+  * @date: 2021-02-25
+  * @desc: fix stock movement report title
+  */
+UPDATE report SET title_key = 'REPORT.STOCK_MOVEMENT_REPORT.TITLE' WHERE report_key = 'stock_movement_report';
```
jniles: Nice 👍

server/controllers/stock/reports/stock_movement_report.handlebars
```diff
@@ -11,19 +11,14 @@
...
-
-    <!-- page title  -->
-        <h2 class="text-center">
-          {{translate 'REPORT.STOCK_MOVEMENT_REPORT.TITLE'}}
-        </h2>
-
-        <h4 class="text-center">
-          {{date dateFrom}} - {{date dateTo}}
-        </h4>
+        <h2 class="text-center">{{translate 'REPORT.STOCK_MOVEMENT_REPORT.TITLE'}}</h2>
```
jniles, line 14: Could you also change the `{{> head TITLE="TREE.STOCK_MOVEMENT_REPORT" }}` to this title key as well?

client/src/modules/reports/generate/stock_movement_report/stock_movement_report.html
```diff
@@ -24,59 +24,64 @@ <h3 translate>REPORT.STOCK_MOVEMENT_REPORT.TITLE</h3>
...
-                <input type="radio" name="value" value="value" ng-model="ReportConfigCtrl.reportDetails.reportType"/>
-                <span translate>FORM.LABELS.MOVEMENT_VALUE</span>
-              </label>
+            <!-- movement or value -->
+            <div class="form-group">
+              <label translate>FORM.LABELS.TYPE</label>
+              <bh-clear on-clear="ReportConfigCtrl.clear('reportType')"></bh-clear>
+              <div class="radio">
+                <label>
+                  <input type="radio" name="movement_number" value="movement_number" ng-model="ReportConfigCtrl.reportDetails.reportType"/>
```
jniles: Can you make these values selected by default?  When I don't select anything, I get a valid report:
image

So, one of these options is selected by default, but I don't know which as a user until I click it.  Can you select them?

jniles: I would also remove the `clear()` button, since all that does is reset the default option.  It will either show one or the other, having these fields blank doesn't make sense.

jniles (commented):
@mbayopanda this looks awesome.  Great work on refactoring the code as well.

image
This is what Vanga's movements look like.

Would it be possible to color the bars by either `is_exit` or not?  That would allow me to see if they are moving more stock out or moving it in during this period.

Otherwise, I've made a few minor comments.  Let me know what you think!

mbayopanda:
@jniles changes landed

server/controllers/stock/core.js
```diff
@@ -35,9 +35,27 @@ const flux = {
   INVENTORY_ADJUSTMENT : 15,
 };
 
+const fluxLabel = {};
```
jniles: To save space, you could also write: 
```js
const fluxLabel = Object.entries(flux)
  .reduce((map, [key, value]) => { map[value] = key; return map; }, {});
```
To accomplish the same thing.  It would also prevent typos.

server/lib/chart.js
```diff
@@ -120,73 +121,110 @@ function barChart(params) {
...
 
+  const dataParameter = {
+    labels : chart.labels,
+    datasets : chart.data,
+  };
+
+  const options = {
+    chart_canvas_id : canvasId,
+    chart_data : dataParameter,
+    chart_x_axis_label : xAxesLabelString,
```
jniles, line 132: Is there someway to specify that the values should be rendered as currency values?

It would be really nice to render this in the enterprise currency of FC:
image

It is a bit hard to read as it is.

client/src/modules/reports/generate/stock_movement_report/stock_movement_report.html
```diff
@@ -24,59 +24,63 @@ <h3 translate>REPORT.STOCK_MOVEMENT_REPORT.TITLE</h3>
...
-                <input type="radio" name="value" value="value" ng-model="ReportConfigCtrl.reportDetails.reportType"/>
-                <span translate>FORM.LABELS.MOVEMENT_VALUE</span>
-              </label>
+            <!-- movement or value -->
+            <div class="form-group">
+              <label translate>FORM.LABELS.TYPE</label>
+              <div class="radio">
+                <label>
+                  <input type="radio" name="movement_number" value="movement_number" ng-model="ReportConfigCtrl.reportDetails.reportType"/>
+                  <span translate>FORM.LABELS.MOVEMENT_NUMBER</span>
```
jniles: I know you didn't write this code, but it would be good to fix this while you are here:  the English translation should read "Movement Count", not "Movement Number".  "Movement Number" sounds like each movement has a number associated with it, whereas "Movement Count" is the count() of all movements.

jniles (commented):
Most of the previous changes are addressed!  There are a few dangling ones.

@mbayopanda it would be really nice if you could color the entries one color (like blue) and the exits another color (like red).  This would let an administrator estimate immediately if they had more exits or entries simply by assessing how much color is on the screen.

It would also be useful to order the datasets by `is_exit`.  That way, all the _entries_ appear on top of the graph, and all the _exist_ appear below.

It would be helpful to put the currency symbol in the x-axis label ($ or FC).

mbayopanda:
@jniles thank you for the review

server/lib/chart.js
mbayopanda: it gave me a headache today 😄, you will see my progress in the commit below

jniles:
@mbayopanda I think you mean to merge this into `refactor-stock-management`, right?

jniles:
Okay, switched the base of the branch.  I'll give this another review quickly.

jniles:
@mbayopanda can you please:

  1. Remove the `migration-imck.sql` file that got in here?
  2. Rebase against `refactor-stock-management`?

It will allow me to review just your edits.

mbayopanda:
@jniles I changed the remote branch

server/controllers/stock/core.js
```diff
@@ -608,7 +615,15 @@ async function getDailyStockConsumption(params) {
     filters.setGroup('GROUP BY DATE(m.date)');
   }
 
-  filters.setOrder('ORDER BY m.date ');
+  if (params.group_by_flux) {
+    filters.setGroup('GROUP BY m.flux_id');
```
jniles: I think this needs to group by `m.flux_id` and also `m.document_uuid`.  Either that, or you need to only `COUNT(DISTINCT document_uuid)` somewhere.  Although due to https://github.com/IMA-WorldHealth/bhima/issues/4083, counting unique documents may not work in this case.

To illustrate, here is the stock movement registry for IMCK:
image

There are only 14 movements from purchase orders.

And here is the graphical report:
image

This thinks there is 56 movements.  My assumption is the issue is that we are counting the individual lines instead of the distinct documents.

jniles (commented):
The code looks pretty good.  I've caught a bug in the logic. 

Also, are these supposed to be colored different colors?  Or is that not possible with ChartJS?

We don't have to address it here if it takes too long, but it would be really nice to have.

jniles:
It would also be nice to have the enterprise currency on our values, similar to what was mentioned in the review of https://github.com/IMA-WorldHealth/bhima/pull/5441

mbayopanda:
@jniles coloration was already done

image

Having values in the enterprise currency is the only remaining task that I didn't succeed to do with the Chartjs api

mbayopanda:
If the need is to color each type of flux with different color, it can be done

jniles:
Ah, wonder why it didn't work for me.  I just re-tested it and the coloration works.

Looks good to me.  The fix for the movement count is now fixed in most cases.

PR #5438, opened 2021-02-26
Fixed expired stock report in refactor-stock-management branch

### LINKED ISSUE #5349: Align the columns of the Stock Expiration Report

jmcameron:
This PR fixes the Expired Stock Report.

The fix was to convert uses of lot.S_RISK (numerical flag) to lot.at_risk (boolean flag).   It seems to work with bhima_test, but I have not tested it with a production database.

Note the comment below for some questions about these variables (https://github.com/IMA-WorldHealth/bhima/pull/5438#issuecomment-786770857).

This PR was updated to address aligning columns between tables for each depot (see Issue https://github.com/IMA-WorldHealth/bhima/issues/5349)
Closes #5349

WARNING: This PR does not fix the totals for the stock at risk!  I'll add a note below about that (https://github.com/IMA-WorldHealth/bhima/pull/5438#issuecomment-786854885)

### DIFF
--- a/server/controllers/stock/reports/stock/expiration_report.js
+++ b/server/controllers/stock/reports/stock/expiration_report.js
@@ -46,11 +46,11 @@ async function stockExpirationReport(req, res, next) {
     // get the lots for this depot
     const lots = await stockCore.getLotsDepot(options.depot_uuid, options);
 
-    // get the lots that are "at risk"
-    const risky = lots.filter(lot => (lot.S_RISK < 0 && lot.lifetime > 0));
+    // get the lots that are "at risk" of expiring
+    const risky = lots.filter(lot => (lot.near_expiration && lot.lifetime > 0));
 
     // get expired lots
-    const expired = lots.filter(lot => (lot.S_RISK <= 0 && lot.expiration_date <= today));
+    const expired = lots.filter(lot => lot.flags.expired);
 
     // merge risky and expired
     const riskyAndExpiredLots = risky.concat(expired);
@@ -76,12 +76,12 @@ async function stockExpirationReport(req, res, next) {
           totals.expired.quantity += lot.mvt_quantity;
           total += lot.value;
         } else {
-          lot.quantity_at_risk = (lot.S_RISK_QUANTITY * -1);
+          lot.quantity_at_risk = lot.S_RISK_QUANTITY;
           lot.value = (lot.quantity_at_risk * lot.unit_cost);
           lot.statusKey = 'STOCK.STATUS.IS_IN_RISK_OF_EXPIRATION';
           lot.classKey = 'bg-warning text-warning';
           totals.at_risk.value += lot.value;
-          totals.at_risk.quantity += (lot.S_RISK_QUANTITY * -1);
+          totals.at_risk.quantity += lot.quantity_at_risk;
           total += lot.value;
         }
       });
--- a/server/controllers/stock/reports/stock_expiration_report.handlebars
+++ b/server/controllers/stock/reports/stock_expiration_report.handlebars
@@ -59,32 +59,32 @@
     <table class="table table-condensed table-bordered table-report">
       <thead>
         <tr>
-          <th>{{translate 'STOCK.INVENTORY'}}</th>
-          <th>{{translate 'STOCK.LOT'}}</th>
-          <th>{{translate 'STOCK.EXPIRATION'}}</th>
-          <th>{{translate 'STOCK.CMM'}}</th>
-          <th>{{translate 'STOCK.CURRENT_QUANTITY'}}</th>
-          <th>{{translate 'STOCK.RISK_QUANTITY'}}</th>
-          <th>{{translate 'TABLE.COLUMNS.UNIT'}}</th>
-          <th>{{translate 'REPORT.STOCK.UNIT_COST'}}</th>
-          <th>{{translate 'STOCK.RISK_VALUE'}}</th>
-          <th>{{translate 'STOCK.STATUS.LABEL'}}</th>
+          <th width="34%">{{translate 'STOCK.INVENTORY'}}</th>
+          <th width="12%">{{translate 'STOCK.LOT'}}</th>
+          <th width="8%">{{translate 'STOCK.EXPIRATION'}}</th>
+          <th width="6%">{{translate 'STOCK.CMM'}}</th>
+          <th width="6%">{{translate 'STOCK.CURRENT_QUANTITY'}}</th>
+          <th width="6%">{{translate 'STOCK.RISK_QUANTITY'}}</th>
+          <th width="6%">{{translate 'TABLE.COLUMNS.UNIT'}}</th>
+          <th width="5%">{{translate 'REPORT.STOCK.UNIT_COST'}}</th>
+          <th width="5%">{{translate 'STOCK.RISK_VALUE'}}</th>
+          <th width="12%">{{translate 'STOCK.STATUS.LABEL'}}</th>
         </tr>
       </thead>
 
       <tbody>
         {{#each depot.rows as |lot| }}
           <tr>
-            <td>{{lot.text}}</td>
-            <td>{{lot.label}}</td>
-            <td>{{date lot.expiration_date}}</td>
-            <td class="text-right">{{lot.avg_consumption}}</td>
-            <td class="text-right">{{lot.quantity}}</td>
-            <td class="text-right">{{lot.quantity_at_risk}}</td>
-            <td>{{lot.unit_type}}</td>
-            <td>{{currency lot.unit_cost ../../metadata.enterprise.currency_id}}</td>
-            <td class="text-right">{{currency lot.value ../../metadata.enterprise.currency_id}}</td>
-            <td class="{{lot.classKey}}">{{translate lot.statusKey}}</td>
+            <td width="34%">{{lot.text}}</td>
+            <td width="12%">{{lot.label}}</td>
+            <td width="8%">{{date lot.expiration_date}}</td>
+            <td width="6%" class="text-right">{{lot.avg_consumption}}</td>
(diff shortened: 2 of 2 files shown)


jmcameron:
When I print out the lots in server/controllers/stock/reports/stock/expiration_report.js  (around line 48), I noticed some oddness with the data in each lot.  I've attached a sample of the output.

- It looks like the flags **expired, near_expiration, exhausted, at_risk**  appear both in the top level of the lot object as well as in the '**flags**' sub-object.   This duplication should probably be cleaned up.

- Using lot.S_RISK (numeric) as a flag should be avoided, now that we have lot.at_risk (boolean)?  We should probably check the usage of S_RISK everywhere.

- I see that the S_RISK, and S_RISK_QUANTITY have reasonable numeric values in some cases, and NaNs in others.  We should probably avoid  NaNs and set them to 'undefined' or 'null' when the calculation would produce NaNs.

[lots.txt](https://github.com/IMA-WorldHealth/bhima/files/6051235/lots.txt)

jniles:
It might be too much to ask, but could you tackle https://github.com/IMA-WorldHealth/bhima/issues/5349 in this PR?

If you would prefer to keep this focused on the expired fix though, I understand.

jmcameron:
> It might be too much to ask, but could you tackle #5349 in this PR?

Ok.  I'll see what I can do.

jmcameron:
@jniles and @mbayopanda : Please do not release this PR before we fix the problem with the cost of at-risk stock not being handled.

@jniles:  Please look at server/controllers/stock/reports/stock/expiration_report.js around line 79+.
I tried changing lot.S_RISK_QUANTITY to lot.mvt_quantity (like the branch above).  It does produce results but I did not understand why they were marked as negatives and adding them to the value of expired stock did not make sense to me..  Please clarify how this should be handled!

mbayopanda (approved):
Thank @jmcameron for this PR as you said this work fixed column alignment.

The issue with totals of `at_risk` and if the `at_risk` status is correctly calculated must be addressed in another PR

jmcameron:
> The issue with totals of `at_risk` and if the `at_risk` status is correctly calculated must be addressed in another PR

That sounds fine.  Do we have an issue for it or do we need to create one?

jmcameron:
@jniles  I updated this (based on my best guess) to handle the value of the at-risk stock in the report.  Take a look at the code in server/controllers/stock/reports/stock/expiration_report.js around line 79+ see if that makes sense to you.  And try the report, obviously.  Your fixes in PR 5438 fixed the negatives issue.

server/controllers/stock/reports/stock/expiration_report.js
```diff
@@ -47,10 +47,10 @@ async function stockExpirationReport(req, res, next) {
     const lots = await stockCore.getLotsDepot(options.depot_uuid, options);
 
     // get the lots that are "at risk"
-    const risky = lots.filter(lot => (lot.S_RISK < 0 && lot.lifetime > 0));
+    const risky = lots.filter(lot => (lot.at_risk && lot.lifetime > 0));
```
jniles: So, I think this report is actually looking for "at risk of expiration".  This "at_risk" flag is "at_risk_of_stock_out".  The correct flag you are looking for is `lot.near_expiration` which will filter on lots that will expire before teh CMM says they will be consumed.  We could make that clearer.

server/controllers/stock/reports/stock/expiration_report.js
```diff
@@ -76,12 +76,12 @@ async function stockExpirationReport(req, res, next) {
           totals.expired.quantity += lot.mvt_quantity;
           total += lot.value;
         } else {
-          lot.quantity_at_risk = (lot.S_RISK_QUANTITY * -1);
+          lot.quantity_at_risk = lot.mvt_quantity;
```
jniles: The quantity at risk here should be the quantity that will expire at the current consumption rate.  The critical code is found here:

https://github.com/IMA-WorldHealth/bhima/blob/cdf85350d60f94f134f6e0ce7b0e0534ec88e8fb/server/controllers/stock/core.js#L787-L809

The value you are looking for is `S_RISK_QUANTITY`.  That should be the quantity of stock that will be lost at the current rate of consumption.  The amount of time that stock could have lasted, had it not expired, is in `S_RISK`.

server/controllers/stock/reports/stock/expiration_report.js
jniles: So, I think you can just change this to:
```js
 const risky = lots.filter(lot => lot.flags.near_expiration);
 ```

File: server/controllers/stock/reports/stock/expiration_report.js [hunk 3]
```diff
@@ -47,10 +47,10 @@ async function stockExpirationReport(req, res, next) {
     const lots = await stockCore.getLotsDepot(options.depot_uuid, options);
 
     // get the lots that are "at risk"
-    const risky = lots.filter(lot => (lot.S_RISK < 0 && lot.lifetime > 0));
+    const risky = lots.filter(lot => (lot.at_risk && lot.lifetime > 0));
 
     // get expired lots
-    const expired = lots.filter(lot => (lot.S_RISK <= 0 && lot.expiration_date <= today));
+    const expired = lots.filter(lot => (lot.expired && lot.expiration_date <= today));
```
jniles: Similarly, this can be:
```js
const expired = lots.filter(lot => lot.flags.expired);
```

jniles (commented):
@jmcameron the changes to the column sizes look good!  Much better alignment.

I believe that the `at_risk` here is being confused between "at risk of expiry" and "at risk of stock out".  I'll make an issue to propose renaming the `at_risk` flags to state specifically what they are.  I've made suggestions about how you can get the real values you are looking for.

server/controllers/stock/reports/stock_expiration_report.handlebars
```diff
@@ -59,32 +59,32 @@
...
-            <td class="text-right">{{currency lot.value ../../metadata.enterprise.currency_id}}</td>
-            <td class="{{lot.classKey}}">{{translate lot.statusKey}}</td>
+            <td width="34%">{{lot.text}}</td>
+            <td width="12%">{{lot.label}}</td>
+            <td width="8%">{{date lot.expiration_date}}</td>
+            <td width="6%" class="text-right">{{lot.avg_consumption}}</td>
+            <td width="6%" class="text-right">{{lot.quantity}}</td>
+            <td width="6%" class="text-right">{{lot.quantity_at_risk}}</td>
+            <td width="6%">{{lot.unit_type}}</td>
+            <td width="5%">{{currency lot.unit_cost ../../metadata.enterprise.currency_id}}</td>
```
jniles: Can you make this cost `class="text-right"` as well?

jniles (commented):
Just wanted to add one thing.  With these changes, here is what the Vanga report looks like:
image

server/controllers/stock/reports/stock_expiration_report.handlebars
jmcameron: Done

server/controllers/stock/reports/stock/expiration_report.js
jmcameron: Done.

However, I'm still seeing duplication of the flags in the lot object as well as in log.flags.

server/controllers/stock/reports/stock/expiration_report.js
jmcameron: Done.

server/controllers/stock/reports/stock/expiration_report.js
jmcameron: Done

jmcameron:
@jniles I did the updates you suggested and it seems to work correctly on bhima_test as well as IMCK (although I have not checked the numbers carefully there).  

Please take one last review.  If it is okay, let me know and I'll rebase it and release it.

jmcameron:
I could also implement Issue https://github.com/IMA-WorldHealth/bhima/issues/5447 (renaming at_risk) before rebasing.

jniles:
Let's hold off on the renaming one.  It will expand to multiple places.  I will give this a quick review this evening.

server/controllers/stock/reports/stock/expiration_report.js
```diff
@@ -46,11 +46,11 @@ async function stockExpirationReport(req, res, next) {
     // get the lots for this depot
     const lots = await stockCore.getLotsDepot(options.depot_uuid, options);
 
-    // get the lots that are "at risk"
-    const risky = lots.filter(lot => (lot.S_RISK < 0 && lot.lifetime > 0));
+    // get the lots that are "at risk" of expiring
+    const risky = lots.filter(lot => (lot.near_expiration && lot.lifetime > 0));
```
jniles, line 50: Nice 👍

server/controllers/stock/reports/stock/expiration_report.js
```diff
@@ -46,11 +46,11 @@ async function stockExpirationReport(req, res, next) {
...
 
     // get expired lots
-    const expired = lots.filter(lot => (lot.S_RISK <= 0 && lot.expiration_date <= today));
+    const expired = lots.filter(lot => (lot.flags.expired && lot.expiration_date <= today));
```
jniles: You probably don't need the expiration date check - the `expired` flag will already do that for you.

jniles (approved):
Approved.  Looks good to me.  I'll let you rebase and release at your leisure.

bors delegate+

server/controllers/stock/reports/stock/expiration_report.js
jmcameron: Fixed

Issue #5418, opened 2021-02-23
Enhancements to "Rapport graphique de la consommation de stock"


mbayopanda:
There are three types of consumption in BHIMA : 
- Distribution to patient
- Distribution to service
- Distribution to depot (if the depot that send stocks is a warehouse)

The chart can be more useful if we can distinguish each of these types of consumption, as proposal we can have a multi bar chart where each bar is for a type of consumption

PR #5441, opened 2021-03-01
Enhance chart of consumption report

mbayopanda:
This PR is related to #5418 

Here is the upgraded report config options : 
image

We have added the distribution option

Here is the upgraded graph : 
image

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -3,7 +3,6 @@
     "ALERT" : "Alert",
     "ADJUSTMENT"      : "Adjustment",
     "ADJUSTMENT_TYPE" : "Adjustment Type",
-    "ALL_DESTINATION" : "All Destination",
     "AMOUNT"          : "Amount",
     "APPROVISIONING"  : "Quantity to order",
     "AT_LEAST_ONE_CHECKED" : "At least one option must be checked",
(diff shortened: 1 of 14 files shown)


jmcameron (commented):
I took skimmed the code. (J'ai parcouri le code.) It seemed okay.
When I loaded the IMCK data and tried it, it looks great except for one issue.  Please note that long inventory names are being cut off on the left.
image

I think there should be some way to force it to wrap text so that the text on the left is not lost.  It is also cut off in the PDF generated.  See attached.
[test.pdf](https://github.com/IMA-WorldHealth/bhima/files/6065049/test.pdf)

jniles:
What is the relationship between this PR and https://github.com/IMA-WorldHealth/bhima/pull/5437?  It looks like some of the changes are identical.  Should one come before the other?

mbayopanda:
@jniles the  #5437 report group data by type of flux (purchase, integration, distribution to patient, etc), the current report group just by inventories and data are restricted to (distribution to patient, distribution to service and distribution to depot)

mbayopanda:
@jniles changes landed

jniles:
@mbayopanda have you rebased against `refactor-stock-management`?  It looks like you've pull in some commits from @jmcameron somehow and that is likely breaking the tests.  Could you take a look?

PR #5445, opened 2021-03-01
fix(stock): give descriptions to adjustments

### LINKED ISSUE #5444: Stock movements without descriptions on some lines of their transactions

jniles:
This commit creates templated descriptions to adjustments and makes
the description mandatory.

Closes #5444.

Here is what it looks like:
image

----

How to test:

1. Load any database off master or refactor-stock-management. 
2. Try to do a negative stock adjustment for any product, _without_ putting in a description.
3. Observe the receipt renders and the app accepts the submit, despite showing a red mark on the description box.
3. Notice that the receipt does not print with a "note" field.

Now, load this branch.

You should not be able to make an adjustment without a description.  And when the description renders, it should look something like what the screenshot I posted above.

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -69,6 +69,8 @@
     "EXIT_SERVICE"    : "Distribution to a service",
     "EXIT_SERVICE_ADVANCED" : "Distribution to the service {{service}} from {{depot}}",
     "EXIT_LOSS"       : "Loss",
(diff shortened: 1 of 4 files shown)


mbayopanda (approved):
LGTM 👍🏽 

bors r+

Issue #5414, opened 2021-02-23
Improvements to "Comparaison des articles factures sur les recus" report


jniles:
In reviewing https://github.com/IMA-WorldHealth/bhima/pull/5341, here are some improvements to the "Comparaison des articles factures sur les recus" report:

1. There isn't a reason to put the date on every line of the distribution - all dates are identical for the movement.  We could just put dates on the header and save on ink.  If we really wanted dates, more useful would be to put the expiration date of that lot.
2. The dates should be right-aligned for easier comparison.
3. https://github.com/IMA-WorldHealth/bhima/issues/5351 exists on this branch.

### Proposal: interleave stock movements with invoices ordered by date
Would it make sense to interleave the stock movements and the invoices in order of date?  Currently, we have them separated by those stock distributions that do not have an associated invoice in a separate section.  It might make sense to combine them into a single thread to better understand the relationship between one and another.  This is probably a question for our clients.

### Proposal: Display subsidies on invoices
This is a somewhat difficult problem to solve, but if there is a subsidy on an invoice, it isn't shown on the invoice and it probably should be.   For example, here are two invoices made at Vanga:
![image](https://user-images.githubusercontent.com/896472/108825048-ce4a1980-75c2-11eb-9b15-3334e317479f.png)

However, here is the Patient Financial Activity:
![image](https://user-images.githubusercontent.com/896472/108825399-4284bd00-75c3-11eb-9eed-b27bed45fa2d.png)

Note that the invoice numbers here are the same `IV.HEV.148773` and `IV.HEV.148843`, but the cost is different.  That is because the patient gets half off their invoice due to a subsidy on their debtor group.  See below:

![image](https://user-images.githubusercontent.com/896472/108825587-81b30e00-75c3-11eb-8ad8-ab09e2c1241b.png)

This is probably relevant information.  The subsidy doesn't have anything to do with the individual items in the invoice, but is rather a subsidy for the entire bill.  So, technically, the report is showing the _correct_ values, since we record that as income in the P&L Statement.  However, it may still be relevant to alert the user to that this invoice has a subsidy.  My preference would be to put it on the invoice header with a (!) exclamation point and an explanation that there are subsidies applied.

### Proposal: Put summary statistics table at top of report
We should include a summary statistics table at the top of the report to help give the user information about how to parse the report.  It's currently quite cluttered.  Here is my suggestion:

Invoices w/o Stock   Movements | Invoices w/ Stock Movements | Unreferenced Stock Movements
[...]

PR #5451, opened 2021-03-02
Add opening balance to the fiche de stock

jniles:
This PR addresses the blocking part of #5414.  It adds an opening balance calculation to the fiche de stock.  It also colors negative values red so they jump out.

Here is what it looks like:
image

How to test this:

1. Build any production database
2. Ensure you've rebuilt the stock_movement_status tables
3. Load the fiche de stock.
4. Pick a variety of date ranges.  You should see that the opening balance is always the end of the previous date range.

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -833,9 +833,69 @@ function computeLotIndicators(inventories) {
 }
 
 /**
- * Inventory Movement Report
+ * @function getOpeningStockBalanceForDepot
+ *
+ * @description
+ * This function gets the opening stock balance for a date, depot
+ * and inventory combination by looking up the most recent value from
+ * the stock_movement_status table.  If none is found, it returns 0s.
  */
-function getInventoryMovements(params) {
+async function getOpeningStockBalanceForDepot(date, depotUuid, inventoryUuid) {
+  const quantities = await db.exec(` 
+    SELECT sum_quantity AS quantity FROM stock_movement_status
+    WHERE date < ? AND depot_uuid = ? AND inventory_uuid = ?
+    ORDER BY date DESC
+    LIMIT 1;
+    `, [date, db.bid(depotUuid), db.bid(inventoryUuid)]);
+
+  // if nothing is found, return nothing
+  if (quantities.length === 0) {
+    return { quantity : 0 };
+  }
+
+  const [{ quantity }] = quantities;
+
+  return {
+    date,
+    quantity,
+  };
+}
+
+/**
+ * @function getOpeningStockBalance
+ *
+ * @description
+ * Gets the opening balance across the enterprise.  This is somewhat
+ * more complex since we keep stock values binned by depot_uuid, so we
+ * just need to get the most recent stock values in all depots. at that date.
+ */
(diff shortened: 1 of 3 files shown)


jniles:
I think this is the last blocking issue of my sets of issues from `refactor-stock-management`.  Tomorrow I'll do a quick run through and update those tracking to make sure we are good to merge when the time comes.

server/controllers/stock/core.js
```diff
@@ -816,10 +816,46 @@ function computeLotIndicators(inventories) {
...
+ *
+ * @description
+ * This function gets the opening stock balance for a date, depot
+ * and inventory combination by looking up the most recent value from
+ * the stock_movement_status table.  If none is found, it returns 0s.
+ */
+async function getOpeningStockBalance(date, depotUuid, inventoryUuid) {
+  const quantities = await db.exec(` 
+    SELECT sum_quantity AS quantity FROM stock_movement_status
+    WHERE date < ? AND depot_uuid = ? and inventory_uuid = ?
```
jmcameron: Shouldn't that be:  date <=

jmcameron (changes requested):
The code looks okay.  I made one suggestion about it that may be incorrect if the "opening balance" is to be the balance at the START of the specified date rather than its ending.

Suggestion: Capitalize the first letter of:  jusqu'à la date
in the Fiche de stock report setup dialog

A few non-blocking appearance suggestions (that could be handled later):  
- The "opening balance" line could easily be overlooked.  It should be in bold at a minimum, maybe a larger font too?   
- Also the edges of the table are not marked with borders, breaking the flow of the table (visually).   I suggest either fixing that or moving the opening balance outside the table.
- Why isn't there an opening balance total?

image

I think there may be a problem with the totals at the end of the report.  See the circled number at the bottom of the screenshot (it appears to be the last value, not a total):
image

Finally, when I tried to use a date interval in the stock sheet setup dialog, selecting the dates  worked okay, but when I click on [Preview], I see this error on the server side:
image

jniles:
Oh dear, the depot_uuid is not required is it?  I forgot about that change.  That's where the error is coming from.  I'll address this case.

server/controllers/stock/core.js
jniles: No, because we want to show the balance as of the most recent date.  The values from the current date will be displayed in the grid, so it will be double-counting if we put the equals.

jniles:
@jmcameron, changes landed.  See if you can reproduce a crash - that was due to me not remembering we no longer require a depot_uuid and it should crash with any combo of dates/depots/inventory items or lack thereof.


> The "opening balance" line could easily be overlooked. It should be in bold at a minimum, maybe a larger font too?

Done. I opted for bold.  

> Also the edges of the table are not marked with borders, breaking the flow of the table (visually). I suggest either fixing that or moving the opening balance outside the table.

Fixed.

> Why isn't there an opening balance total?

There is - this was just a layout issue because we add a column when there was a depot.  It should be fixed.

Here is what the new version looks like:

image

jmcameron (approved):
I just retested this and it looks good and I could not reproduce the date interval bug.

Issue #5425, opened 2021-02-24
Improvements to Sorties de Stocks


jniles:
This issue details suggestions/improvements/bugs found while reviewing https://github.com/IMA-WorldHealth/bhima/pull/5341.

### Bug: Submit button can be active while form is invalid
The submit button can be active when the form isn't valid.  This is a corner case that happens when the date is changed.  I think this is due to asynchronous validation that waits to block the submit button until the HTTP request is finished returning from the server.  It's pretty easy to reproduce though:

![Tk2thkHu5h](https://user-images.githubusercontent.com/896472/109011018-c4064900-76b0-11eb-93a6-0044387275f2.gif)

_This bug exists on master, so it does not block #5341.  It is a critical bug though and should be addressed as soon as possible_.

### Proposal: if only one lot exists for an inventory item, auto-select it
On the stock exit, when we have a single lot for an inventory item, let's select it by default.  There is a slight risk that users will start to be lazy and no longer check that the lot is the correct lot.  However, I think the benefit to power users outweighs the risk of laziness.

I've discussed this proposal with @lomamech in person and he agrees with it as well.

### Feature Request: Implement caching of the exit grid
In invoices, we have the ability to cache the grid when we submit, and restore it via a button.  From a user perspective, it would be useful to do this for the stock exit.  We would need to move a lot of our validation checks from the client to the server though, to prevent errors happening from stale data.

Motivation: for the Health Zone project, users often get to the very end of an exit document, realize they need to change the date and that erases all their work - sometimes hours of work.  This is worth implementing to avoid this scenario, despite the development complexity.

jniles:
I'm going to tackle the "invalid submit button" states part of this issue.

jniles:
Note the invalid submit button has been fixed.

PR #5454, opened 2021-03-03
fix(stock): submit button active only when valid on exit

jniles:
This commit ensures the submit button on the exit module is only active
when the form is valid for submission.  Previously, there was a corner
case where we would enable the submit button after a date change,
despite not having valid data.

Partially addresses #5425.

### DIFF
--- a/client/src/modules/stock/exit/exit.js
+++ b/client/src/modules/stock/exit/exit.js
@@ -38,6 +38,7 @@ function StockExitController(
       vm.dateMessageWarning = true;
     }
     loadInventories(vm.depot, date);
+    checkValidity();
   };
 
   vm.onChangeDepot = depot => {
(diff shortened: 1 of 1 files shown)


jmcameron (commented):
The code looks good.
I tested this and it seems to work.

I was confused about one thing:  When you change the date to one before the present or in the future, it deletes ALL the current inventory items.  I found this strange.  It probably should give you a warning that you'll need to reselect the inventory items.  This could be improving the current "you have chosen a past date" message to indicate that has deleted any inventory items and you need to reselect them -- or, when you click on the [Change Date] button, it could pop up a dialog that confirms that you'll need to reselect the inventory items.

lomamech:
@jniles  it's really curious I can't reproduce this error

lomamech (approved):
Although I did not manage to reproduce this error, the changes made will prevent the submit button from remaining usable when the grid is empty

Issue #5424, opened 2021-02-24
Improvements to Stock Movements Registry


jniles:
This issue concerns remarks/bugs/suggestions made following https://github.com/IMA-WorldHealth/bhima/pull/5341.  

### Suggestion: move the "reason" next to the "flow" direction.
The reason column is too far over.  Let's move it next to the flow direction since they go together.
![image](https://user-images.githubusercontent.com/896472/109004494-1cd1e380-76a9-11eb-992f-cb8ec46e0d0f.png)

### Suggestion: show more information in "reason"
Similar to https://github.com/IMA-WorldHealth/bhima/issues/5415, we should show more information in or next to the "reason" column.  For example, we could show:

1. Patient Reference (PA.TPA.1234)
2. Service (Surgery)
3. Requisition Number (SREQ.TPA.1)
4. Depot Name (Pharmacy Stock)

Some of these are easier to show than others.  It would take some reflection to figure out how we could show these items without too much SQL churn.

### Proposal: We need a detailed view of this same registry
Currently, we have no way of showing all the movements of one particular lot across the enterprise.  This makes research a bit difficult and cumbersome.  It would be nice to show the actual details of the stock movement so we can see the number of different times Lot X or Inventory Article Y was moved.  I'll make this another issue in the future.

In https://github.com/IMA-WorldHealth/bhima/pull/4572 we removed a registry with details of the stock movements.  We may want to reverse this decision.

PR #5459, opened 2021-03-03
Improvements to Stock Movement Registry

### LINKED ISSUE #4304: The "cost" column in stock movements registry should have a sum at the bottom

jniles:
This commit moves the flux column over to the I/O column so that a user can more easily determine the reason for the directionality of the flow.

Partially addresses #5424.

This commit adds a dynamic sum to the movements registry to allow us to quickly sum the outgoing medications.  Here is what that looks like:

UIZvL5Q21G

Closes #4304.

### DIFF
--- a/client/src/modules/stock/movements/registry.js
+++ b/client/src/modules/stock/movements/registry.js
@@ -56,6 +56,11 @@ function StockMovementsController(
       headerCellFilter : 'translate',
       cellTemplate : 'modules/stock/movements/templates/io.cell.html',
     }, {
+      field : 'flux_id',
+      displayName : 'STOCK.FLUX',
+      headerCellFilter : 'translate',
+      cellTemplate : 'modules/stock/movements/templates/flux.cell.html',
+    }, {
       field : 'cost',
       type : 'number',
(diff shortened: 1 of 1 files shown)


jniles:
@lomamech, could you give me a review of this PR?  I'm trying to close #4304 and I know you had some discussion about it.

What do you think about this implementation?  It should hide the quantities when the `is_exit` type is mixed.  We could even hide if all rows are not the same `flux_id`.  The goal is just to be able to say how much $$ the pharmacy moved in or out with some different criteria.

lomamech:
> @lomamech, could you give me a review of this PR? I'm trying to close #4304 and I know you had some discussion about it.
> 
> What do you think about this implementation? It should hide the quantities when the `is_exit` type is mixed. We could even hide if all rows are not the same `flux_id`. The goal is just to be able to say how much $$ the pharmacy moved in or out with some different criteria.

Well received

lomamech (approved):
No Comment

PR #5461, opened 2021-03-03
Rename lot at_risk flag to at_risk_of_stock_out

### LINKED ISSUE #5447: Rename `at_risk` flag to `at_risk_of_stock_out`

jmcameron:
Rename the lot flag for clarity.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5447

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -895,11 +895,11 @@ a[disabled] {
   color: #FF7B7B !important;
 }
 
-.at-risk {
+.at-risk-of-stock-out {
   background-color: #ffff6686 !important;
 }
 
-.icon-at-risk {
+.icon-at-risk-of-stock-out {
   color: #FFFFA9 !important;
 }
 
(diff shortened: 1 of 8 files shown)


jniles (approved):
Looks good to me.  Tested and working with IMCK database.

bors r+

PR #5464, opened 2021-03-04
Clean and sanitaze code

lomamech:
Juste for clean and sanitaze code from aggregate stock consumption


jniles (approved):
Looks good to me.  I like changes that delete things.

PR #5465, opened 2021-03-04
improve-chart-of-consumption

### LINKED ISSUE #5418: Enhancements to "Rapport graphique de la consommation de stock"

mbayopanda:
Close #5441 
Close #5418 

image

This PR contains only changes about stock consumption graph

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -241,6 +241,7 @@
     "ENABLE_STRICT_DEPOT_DISTRIBUTION_HELP_TEXT" : "Limits the depots a user can interact with to the depots that the user has access to. Thus, if a user cannot distribute stock from a depot, they also cannot transfer stock to that depot."
   },
   "STOCK_FLUX" : {
+    "ALL_DESTINATION"  : "All Destination",
     "FROM_PURCHASE"    : "From Purchase Order",
     "FROM_OTHER_DEPOT" : "From Depot",
     "FROM_ADJUSTMENT"  : "Adjustment (Positive)",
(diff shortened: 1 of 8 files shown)


client/src/i18n/en/stock.json
```diff
@@ -241,6 +241,7 @@
     "ENABLE_STRICT_DEPOT_DISTRIBUTION_HELP_TEXT" : "Limits the depots a user can interact with to the depots that the user has access to. Thus, if a user cannot distribute stock from a depot, they also cannot transfer stock to that depot."
   },
   "STOCK_FLUX" : {
+    "ALL_DESTINATION"  : "All Destination",
```
jniles: This is better written "All Destinations".

mbayopanda:
@jniles I create a new PR for the chart of consumption, the previous one was corrupted during a rebase

jniles (commented):
@mbayopanda, these changes look good.  As demonstrated together, we need to ensure that the caches of the browser are cleared, or that we remove the fiscal year/ period ids on the server.

jniles (approved):
This looks really nice!  Great work @mbayopanda.  I know the ChartJS API can be hairy - you did a good job making the code readable.

It would be nice to order by value from largest to smallest, but I will create an issue for that.  With graphical reports, we can always improve, improve, improve!

bors r+

PR #5469, opened 2021-03-05
fix(stock): rewrite monthly consumption report

jniles:
Restored from https://github.com/IMA-WorldHealth/bhima/pull/5436

This commit rewrites the monthly consumption report to use the new version of `stock_movement_status`.

Here is what it looks like:
image

I've also attempted to improve the layout somewhat so we don't have as much wrapping of text on large screens.

---

How to test:
  1. Load any database.  In this case, load the test one.  Observe that you will find stock movements for February 2021 in Depot Principal.
  2. Go to the stock exit and choose the Depot Principal as your depot
  3. Set the date to January 15 2021.
  4. Make a distribution of Quinine to the server Test Service.
  5. Return to the report.
  6. Observe that this movement now shows up in January.

### DIFF
--- a/server/controllers/stock/reports/monthly_consumption.report.handlebars
+++ b/server/controllers/stock/reports/monthly_consumption.report.handlebars
@@ -18,32 +18,34 @@
       <table class="table table-striped table-condensed table-report table-bordered">
         <thead>
           <tr style="background-color:#ddd;">
-            <th class="text-center" style="width: 3%;"> {{translate 'FORM.LABELS.NR'}}</th>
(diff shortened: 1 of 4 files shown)


lomamech (commented):
After analysis, it will be necessary to modify the StageInventoryForAMC () procedure so that we can take into account the consumption of aggregated stock.

jniles:
Thanks for the review!

I'll work on fixing this.  It should be pretty straightforward.

jniles:
@lomamech it is fixed.

PR #5470, opened 2021-03-05
fix(stock): fix currencies in invoice comp report

### LINKED ISSUE #5351: Report comparing invoiced items to the stock received renders incorrect currency
A picture is easier to describe:

image

This value should be the enterprise currency, which is FC.  In the "No Invoice Attributable" section, we don't detect the currency properly in the totals section.

jniles:
Restored from https://github.com/IMA-WorldHealth/bhima/pull/5456

Renders currencies correctly on the report comparing invoiced stock and received stock.  The report now uses the enterprise currency for comparison.

Closes #5351.

image


mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #5471, opened 2021-03-05
chore(stock): improve exit report formatting

### LINKED ISSUE #5419: Improvements to the "Rapport des sorties de stocks" report
These are observations/bugs found during the review of https://github.com/IMA-WorldHealth/bhima/pull/5341.

### Proposal: make the titles smaller
Here is what the current version of the report looks like:
image
It doesn't make sense to have the title one size and the total column a different size.  We should make them both the same size.  I propose we make them both small, since that is how most of our reports work.  Saves paper.

### Proposal: remove the parentheses from the item type
There isn't any reason to have the parentheses and italics from the item type.  Let's remove these decorations and just leave the units next to the value.

jniles:
Restored from https://github.com/IMA-WorldHealth/bhima/pull/5455

This commit improves the stock exit report by:

 1. making the titles smaller
 2. removing the italics and parens around the units

Closes #5419.

image

### DIFF
--- a/server/controllers/stock/reports/stock_exit.report.handlebars
+++ b/server/controllers/stock/reports/stock_exit.report.handlebars
@@ -27,7 +27,7 @@
         <table class="table table-condensed table-report">
           <thead>
             <tr style="background-color: #EFEFEF;">
-              <th colspan="5"><h3 style="margin: 0px;">{{translate 'STOCK.RECEIPT.EXIT_PATIENTS'}}</h3></th>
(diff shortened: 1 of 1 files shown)


mbayopanda (commented):
The changes are 👍🏽 

I think you have missed the `Consommation de stock agrégée` 

image

The size of this section is a bit bigger

jniles:
Aha, thank you!  I will update it.

jniles:
Fixed:

image

mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #5472, opened 2021-03-05
chore: clean up install tests

jniles:
This commit cleans up the install tests by using the server's app to spin up the server similar to the integration tests.

### DIFF
--- a/sh/install-tests.sh
+++ b/sh/install-tests.sh
@@ -3,37 +3,21 @@
 # bash script mode
 set -euo pipefail
 
-trap 'kill $(jobs -p)' EXIT
-
 set -a
 source .env || { echo '[install-tests.sh] could not load .env, using variables from environment.' ; }
 set +a
 
-./sh/build-init-database.sh
+./sh/build-init-database.sh || { echo 'failed to build DB' ; exit 1; }
 
 echo "[install test]"
 
-# set build timeout
-TIMEOUT=${BUILD_TIMEOUT:-8}
-
-echo "[install test] Spawning server process..."
+echo "[install test] building the server..."
 # build and start the server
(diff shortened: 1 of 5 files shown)


sh/install-tests.sh
```diff
@@ -3,37 +3,21 @@
...
 set -euo pipefail
 
-trap 'kill $(jobs -p)' EXIT
-
 set -a
 source .env || { echo '[install-tests.sh] could not load .env, using variables from environment.' ; }
 set +a
 
-./sh/build-init-database.sh
+./sh/build-init-database.sh || { echo 'failed to build DB' ; exit 1; }
```
jmcameron, line 10: Nice.  I had forgotten that you could to this scripts as well as individual commands.

jmcameron (approved):
I admit to NOT completely understanding what the install script actually does, but the code looked okay.

I ran 'yarn test:install' and it worked fine on my system.

BTW, a suggestion that could be done in this PR (or in another):  Would it be possible to update printouts when running 'yarn test' to clearly/loudly printout when each of the individual sub test scripts is running?   In 'sh/tests.sh', I see the command 'startfold "message"' used with each test script.  It would be helpful if that command printed out the message more clearly.  The definition of startfold looks like it prints out the message, but I don't see it when I look at the output from 'yarn test'.  So either it is not actually being printed, or we should make it stand out better (eg, print blank lines or lines of '===' above/below it  to make it stand out).

jniles:
The "startfold" stuff is tied to Travis and is probably legacy at this point.

Feel free to implement this!  I'm not entirely sure what you have in mind, but it sounds cool.  I've no issue with the idea.

jniles:
It's been approved, so I suppose... 

bors r+

PR #5473, opened 2021-03-05
feat(stock): add minimum delay to stock settings

### LINKED ISSUE #5407: Add "minimum delay" setting to the inventory items to ensure we have useful stock warnings
Our minimum value of stock is computed as follows:
https://github.com/IMA-WorldHealth/bhima/blob/1819d8f7380f16590fce3fbbe3c900513d001c56/server/controllers/stock/core.js#L506-L515

This is fine, except in select cases where the `delay` is very small.  At Vanga, they have a supplier on the same grounds as the hospital, which means their `delay` is close to 0.  The minimum stock is effectively 0.  This is mathematically correct, but administratively it is not useful since we will never have warning about potential stock outs until we are literally in a stock out.

We should allow users to set a "minimum delay" so that we have a useful value to show users.  The way this would work is to take the maximum of the computed delay and the minimum delay and use that as the delay in the calculation.  This would ensure that we don't fall below the minimum delay in our reporting.

jniles:
Adds a "minimum delay" setting to stock management to ensure that we are able to detect stock outs on time, even if the purchase data is messy. Previously, we relied on purchasing data solely for the delay, and users who put in a purchase order, then immediately delivered it, shoved the delay closer to 0.  This caused us to not warn users in time of stock outs.  The user can now configure the minimum number of months in decimal form on the stock settings page.

Closes #5407.

Bq45NpG2Zz

----

How to test:
  1. Load a production database.  Remember to have run the migration file and built the stock movement status tables.
  2. Go to the articles in stock registry.  You'll likely have a mix of states of stock.
  3. Go to the stock settings page.  Notice that the setting for minimum delay is empty. Set it to 500 months.
  4. Go back to the articles in stock registry.  All the stock with CMM should show a critical stock state.
  5. Return and set to something more reasonable such as 1.  You will likely have another mix of stock states.

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -266,6 +266,11 @@ async function getLotsDepot(depotUuid, params, finalClause) {
 
   const resultFromProcess = await db.exec(query, queryParameters);
 
+  // add minumum delay
+  resultFromProcess.forEach(row => {
+    row.min_delay = params.min_delay;
+  });
+
   // calulate the CMM and add inventory flags.
   const inventoriesWithManagementData = await getBulkInventoryCMM(
     resultFromProcess,
@@ -509,11 +514,16 @@ function computeInventoryIndicators(inventories) {
     // Signal that no consumption has occurred of the inventory items
     inventory.NO_CONSUMPTION = (CMM === 0);
 
+    // NOTE(@jniles) ensure that this value is set before going too far.
+    const delay = Math.max(inventory.min_delay, inventory.delay);
+
     // Compute the Security Stock
     // Security stock is defined by taking the average monthly consumption (AMC or CMM)
-    // and multiplying it by the Lead Time (inventory.delay).  The Lead Time is by default 1 month.
+    // and multiplying it by the delay factor.  Typically, this is the lead time (with
+    // a default of one month).  However, if the lead time is too small, the user can set
+    // a floor value using the stock setting's min_delay parameter.
     // This gives you a security stock quantity.
-    inventory.S_SEC = CMM * inventory.delay; // stock de securite
+    inventory.S_SEC = CMM * delay; // stock de securite
 
     // Compute Minimum Stock
     // The minumum of stock required is double the security stock.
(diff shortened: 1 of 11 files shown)


server/controllers/stock/core.js
```diff
@@ -509,11 +514,14 @@ function computeInventoryIndicators(inventories) {
     // Signal that no consumption has occurred of the inventory items
     inventory.NO_CONSUMPTION = (CMM === 0);
 
+    // NOTE(@jniles) ensure that this value is set before going too far.
+    const delay = (inventory.min_delay > inventory.delay) ? inventory.min_delay : inventory.delay;
```
jmcameron: This could also be done more cleanly with max(), right?

server/controllers/stock/core.js
```diff
@@ -509,11 +514,14 @@ function computeInventoryIndicators(inventories) {
...
+
     // Compute the Security Stock
     // Security stock is defined by taking the average monthly consumption (AMC or CMM)
     // and multiplying it by the Lead Time (inventory.delay).  The Lead Time is by default 1 month.
```
jmcameron: You may want to update this comment?

jmcameron (approved):
I tested it (with IMCK db) and it works as advertised, with one exception: When I set the delay to 500, all of the articles in stock lines that had non-zero CMM worked as advertised.  The ones that did have zero CMM values showed as "Over Maximu" status tags.  This actually seems correct, so I think it is ready to go.

server/controllers/stock/core.js
jniles: Yup.  That would be smarter.  Let me make that change

server/controllers/stock/core.js
jniles: Yeah.  This is no longer the whole truth.  Thanks!

jniles:
@jmcameron changes landed!

jmcameron (commented):
Looks good. I retested and it worked like before.

Issue #5458, opened 2021-03-03
Enhance "PostStockMovement" procedure to handle entry and exit in one document


mbayopanda:
Our `PostStockMovement` procudure require : 
- `uuid` of the stock movement document
- `is_exit` for an entry or an exit movement
- `project_id`

The `uuid` and `is_exit` do not allow us to use the procedure for one document which contains `entries` and `exits` operations such as a `stock adjustment`

jniles:
Interesting.  I can see this being really challenging.  A lot of our logic depends on it currently, but we could just remove the `is_exit` and use the `flux_id`, then make `flux_id` either an "entry", "exit" or "both", with the idea being that both could contain both increases and decreases.

jniles:
@mbayopanda Can this be closed?

PR #5475, opened 2021-03-05
New stock adjustment receipt

mbayopanda:
This PR is related to #5458 by using just one document during an adjustment of stock, the part about the procedure is not touched

image

### DIFF
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -298,6 +298,7 @@ async function createInventoryAdjustment(req, res, next) {
     // pass reverse operations
     const trx = db.transaction();
 
+    const uniqueAdjustmentUuid = uuid();
     const positiveAdjustmentUuid = uuid();
     const negativeAdjustmentUuid = uuid();
 
@@ -314,8 +315,8 @@ async function createInventoryAdjustment(req, res, next) {
         uuid : db.bid(uuid()),
         lot_uuid : db.bid(lot.uuid),
         depot_uuid : db.bid(movement.depot_uuid),
-        document_uuid : db.bid(negativeAdjustmentUuid),
-        quantity : lot.oldQuantity,
+        document_uuid : db.bid(uniqueAdjustmentUuid),
+        quantity : lot.oldQuantity + (lot.quantity - lot.oldQuantity),
         unit_cost : lot.unit_cost,
         date : new Date(movement.date),
         entity_uuid : movement.entity_uuid,
@@ -333,8 +334,8 @@ async function createInventoryAdjustment(req, res, next) {
         uuid : db.bid(uuid()),
         lot_uuid : db.bid(lot.uuid),
         depot_uuid : db.bid(movement.depot_uuid),
-        document_uuid : db.bid(positiveAdjustmentUuid),
-        quantity : lot.oldQuantity * -1,
+        document_uuid : db.bid(uniqueAdjustmentUuid),
+        quantity : lot.oldQuantity + (lot.quantity - lot.oldQuantity),
         unit_cost : lot.unit_cost,
         date : new Date(movement.date),
         entity_uuid : movement.entity_uuid,
@@ -369,25 +370,25 @@ async function createInventoryAdjustment(req, res, next) {
     await trx.execute();
 
     // pass inventory adjustment as new movement
-    const document = {
-      uuid : uuid(),
-      date : new Date(movement.date),
-      user : req.session.user.id,
-    };
-    const positiveLots = lots
-      .filter(lot => lot.quantity > 0)
-      .map(lot => {
-        delete lot.oldQuantity;
-        return lot;
-      });
+    // const document = {
+    //   uuid : uuid(),
+    //   date : new Date(movement.date),
+    //   user : req.session.user.id,
+    // };
+    // const positiveLots = lots
+    //   .filter(lot => lot.quantity > 0)
+    //   .map(lot => {
+    //     delete lot.oldQuantity;
+    //     return lot;
+    //   });
 
-    movement.is_exit = 0;
-    movement.flux_id = core.flux.INVENTORY_ADJUSTMENT;
-    movement.lots = positiveLots;
-    movement.period_id = periodId;
+    // movement.is_exit = 0;
+    // movement.flux_id = core.flux.INVENTORY_ADJUSTMENT;
+    // movement.lots = positiveLots;
+    // movement.period_id = periodId;
 
-    await normalMovement(document, movement, req.session);
-    res.status(201).json(document);
+    // await normalMovement(document, movement, req.session);
+    res.status(201).json({ uuid : uniqueAdjustmentUuid, date : new Date(movement.date), user : req.session.user.id });
   } catch (err) {
     next(err);
   }
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -299,93 +299,69 @@ async function createInventoryAdjustment(req, res, next) {
     const trx = db.transaction();
 
     const uniqueAdjustmentUuid = uuid();
-    const positiveAdjustmentUuid = uuid();
-    const negativeAdjustmentUuid = uuid();
 
-    // get all lots with positive quantities
-    const positiveQuantities = lots.filter(lot => lot.oldQuantity > 0);
+    let countNeedIncrease = 0;
+    let countNeedDecrease = 0;
 
-    // get all lots with negative quantities
-    // negative quantities occurs during some extra stock exit when quantity in stock
-    // is already under or equal to zero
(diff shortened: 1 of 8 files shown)


mbayopanda:
The total cost of each line must be based on the difference and not on the quantity

server/controllers/stock/index.js
```diff
@@ -298,96 +298,91 @@ async function createInventoryAdjustment(req, res, next) {
...
+          unit_cost : lot.unit_cost,
+          date : new Date(movement.date),
+          entity_uuid : movement.entity_uuid,
+          is_exit : 0,
+          flux_id : core.flux.INVENTORY_ADJUSTMENT,
+          description : movement.description,
+          user_id : req.session.user.id,
+          period_id : periodId,
+        };
+        adjustmentLog.push(_.assign({}, lot, { difference, is_exit : 0 }));
```
jniles: Just FYI, you can do the `_.assign()` operation with vanilla javascript.  Like this:
```js
var x = _.assign({}, lot, {difference, is_exit: 0 });
// is equivalent to 
var y = { ...lot, difference, is_exit : 0 };
```

server/controllers/stock/reports/stock_adjustment.receipt.handlebars
```diff
@@ -48,40 +48,128 @@
...
+            <td>{{label}}</td>
+            <td>{{date expiration_date}}</td>
+            <td class="text-right">{{oldQuantity}}</td>
+            <td class="text-right">{{quantity}}</td>
+            <td class="text-right">{{difference}}</td>
+            <td class="text-right">{{currency unit_cost ../enterprise.currency_id}}</td>
+            <td class="text-right">{{currency total ../enterprise.currency_id}}</td>
+          </tr>
+        {{else}}
+          {{> emptyTable columns=11}}
```
jniles: I think this `columns` needs to be 9.

server/controllers/stock/reports/stock_adjustment.receipt.handlebars
```diff
@@ -48,40 +48,128 @@
...
```
jniles: I also think this is 9 columns.

server/models/schema.sql
```diff
@@ -1978,6 +1978,17 @@ CREATE TABLE `stock_requisition_item` (
   CONSTRAINT `stock_req_item__stock_req_item` FOREIGN KEY (`requisition_uuid`) REFERENCES `stock_requisition` (`uuid`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
+DROP TABLE IF EXISTS `stock_adjustment_log`;
+CREATE TABLE `stock_adjustment_log` (
+  `document_uuid` BINARY(16) NOT NULL,
+  `log_timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
jniles: Is there a particular reason you'd want to call this `log_timestamp`?  If not, let's call it the usual `created_at`, like we have in a couple other tables here.

server/models/schema.sql
```diff
@@ -1978,6 +1978,17 @@ CREATE TABLE `stock_requisition_item` (
...
+  `text` JSON,
```
jniles: Let's call this `data`, since it's JSON format.  Or `changes` or something more description.  We use `text` a lot for labels and this isn't a label.

server/controllers/stock/index.js
jniles: I think this may end up being a lot of information to store in the JSON column.  Would it be possible to store only the `movement.uuid` and the `difference`?  Then, when you are rendering the receipt, you simply stitch them together with while displaying the receipt?

server/controllers/stock/index.js
```diff
@@ -298,96 +298,91 @@ async function createInventoryAdjustment(req, res, next) {
...
+          uuid : db.bid(uuid()),
+          lot_uuid : db.bid(lot.uuid),
+          depot_uuid : db.bid(movement.depot_uuid),
+          document_uuid : db.bid(uniqueAdjustmentUuid),
+          quantity : Math.abs(difference),
+          unit_cost : lot.unit_cost,
+          date : new Date(movement.date),
+          entity_uuid : movement.entity_uuid,
+          is_exit : 0,
+          flux_id : core.flux.INVENTORY_ADJUSTMENT,
```
jniles: Cool 👍

server/controllers/stock/index.js
```diff
@@ -298,96 +298,91 @@ async function createInventoryAdjustment(req, res, next) {
...
-        date : new Date(movement.date),
-        entity_uuid : movement.entity_uuid,
-        is_exit : 0,
-        flux_id : core.flux.INVENTORY_RESET,
-        description : movement.description,
-        user_id : req.session.user.id,
-        period_id : periodId,
-      };
-      trx.addQuery('INSERT INTO stock_movement SET ?', reverseMovementObject);
+      if (difference < 0) {
```
jniles, line 330: It looks like there is a lot of shared/repeated information between these two `if`s.  Could you unify them a bit?

```js
const movementObject = {
          uuid : db.bid(uuid()),
          lot_uuid : db.bid(lot.uuid),
          depot_uuid : db.bid(movement.depot_uuid),
          document_uuid : db.bid(uniqueAdjustmentUuid),
          unit_cost : lot.unit_cost,
          quantity: Math.abs(quantity),
          date : new Date(movement.date),
          entity_uuid : movement.entity_uuid,
          flux_id : core.flux.INVENTORY_ADJUSTMENT,
          description : movement.description,
          user_id : req.session.user.id,
          period_id : periodId,
};

if (difference < 0) {
        countNeedDecrease++;
       Object.assign(movementObject, { is_exit : 1 });
       adjustmentLog.push(_.assign({}, lot, { difference, is_exit : 1 }));
} else {
        countNeedIncrease++;
       Object.assign(movementObject, { is_exit : 0 });
       adjustmentLog.push(_.assign({}, lot, { difference, is_exit : 0 }));
}

trx.addQuery('INSERT INTO stock_movement SET ?', movementObject);

```

You could actually probably get it smaller than this by adopting my proposal to remove `_.assign()`.  I think the only thing that changes is `is_exit`, right?

server/controllers/stock/index.js
```diff
@@ -298,96 +298,91 @@ async function createInventoryAdjustment(req, res, next) {
...
+        trx.addQuery('CALL PostStockMovement(?)', [increaseParams]);
       }
     }
 
-    // reset all previous lots
-    await trx.execute();
+    const logParams = [
+      db.bid(uniqueAdjustmentUuid),
+      JSON.stringify(adjustmentLog),
+      req.session.user.id,
```
jniles: Do you need the user on both this table and the `stock_movement` table?  It seems like you probably only need the `document_uuid` and you could recover the user, right?

server/controllers/stock/index.js
```diff
@@ -298,96 +298,91 @@ async function createInventoryAdjustment(req, res, next) {
...
-    movement.is_exit = 0;
-    movement.flux_id = core.flux.INVENTORY_ADJUSTMENT;
-    movement.lots = positiveLots;
-    movement.period_id = periodId;
+    // reset all previous lots
+    await trx.execute();
 
-    await normalMovement(document, movement, req.session);
-    res.status(201).json(document);
+    // await normalMovement(document, movement, req.session);
```
jniles, line 366: 😆

jniles (commented):
@mbayopanda this is really cool!

I've made a few comments on the code and structure of the table.  I'm curious how much this may impact our other modules/registries, but it actually looks really clean.  Great job!

server/controllers/stock/index.js
jniles: Actually, if you did that, you wouldn't need a JSON column.  Just two columns - `row_uuid` and `difference`.  You could then do the `JOIN` in MySQL!

mbayopanda:
@jniles changes landed

server/controllers/stock/index.js
```diff
@@ -298,96 +298,73 @@ async function createInventoryAdjustment(req, res, next) {
...
-        unit_cost : lot.unit_cost,
-        date : new Date(movement.date),
-        entity_uuid : movement.entity_uuid,
-        is_exit : 0,
-        flux_id : core.flux.INVENTORY_RESET,
-        description : movement.description,
-        user_id : req.session.user.id,
-        period_id : periodId,
+      const log = {
+        movement_uuid : adjustmentMovement.uuid,
```
jniles, line 325: Beautiful this is much cleaner.  :+1:

jniles (approved):
Code looks good to me.  Tested and works with the vanga dataset.

@mbayopanda I see you still have the "do not merge" flag on here.  Go ahead and merge this when ready.

bors delegate+

PR #5479, opened 2021-03-05
fix(patients): remove duplicate stock movements in financial standing

### LINKED ISSUE #5350: Patient Standing Report duplicates stock distributions
If you include the stock distributed to the patient in the patient standing report, we get a lot more rows than one would expect.  For example:

image

These are duplicates of each other (see the SM.* number is identical).  This is a bug.  We should be only showing the stock movements once.

jniles:
Removes duplicate stock movements from the patient standing report. Also improves the SQL/JS code by removing extraneous values and grouping where it should be grouped.  Finally, we've added the invoice reference to the stock movement where possible.

Closes #5350.

Here is what it looks like, with the invoice reference highlighted:
image

### DIFF
--- a/server/controllers/finance/reports/financial.patient.handlebars
+++ b/server/controllers/finance/reports/financial.patient.handlebars
@@ -60,9 +60,9 @@
         </tr>
       </tfoot>
     </table>
+
     {{#if includeStockDistributed }}
-      <br>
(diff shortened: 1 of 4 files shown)


lomamech (approved):
Very good correction

image

PR #5483, opened 2021-03-08
fix: typo in old migration script

jniles:
This commit fixes a typo in an old migration script.  If we ever need to upgrade an old database, we'll be bit by this.


jmcameron (approved):
This is a perfect fix. 
I would go ahead and release it but you've requested a review by someone else.

lomamech (approved):

PR #5484, opened 2021-03-08
chore(roles): use bhCheckboxTree for actions

### LINKED ISSUE #5378: Make "Role Actions" into a bhCheckboxTree with a "check all" parameter
The Role Actions modal has a list of actions that users with this role are allowed or not allowed to do.  Here they are:

image

We should make this into a `bhCheckboxTree` component, like our permissions modal.  This would save on whitespace and allow us to have a "check all" parameter that quickly toggles all the actions on or off.

jniles:
Transitions the role actions module to using bhCheckboxTree instead of
rolling its own list structure.

I've also changed the behavior of bhCheckboxTree slightly to make it
more user friendly - onChange() now checks if the previous value is
equal to the current value to avoid excessive calls.

Additionally isFlatTree is now a boolean value.

Closes #5378.

### DIFF
--- a/client/src/js/components/bhCheckboxTree/bhCheckboxTree.js
+++ b/client/src/js/components/bhCheckboxTree/bhCheckboxTree.js
@@ -46,8 +46,10 @@ function bhCheckboxTreeController(Tree) {
     }
 
     if (changes.checkedIds && changes.checkedIds.currentValue) {
-      processCheckedIds();
-      getCheckedNodes();
+      if (!angular.equals(changes.checkedIds.currentValue, changes.checkedIds.previousValue)) {
+        processCheckedIds();
+        getCheckedNodes();
+      }
     }
   };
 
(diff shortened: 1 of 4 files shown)


jniles:
~~Putting "do not merge" on this because I'm going to try and use it to get end to end tests up and running.~~

Scratch that, I'll do this later.

jmcameron (approved):
It looks good. I tested it and it works. 

However, I had one non-blocking question.  The "breadcrumbs" on the actions dialog look different for me than for the image I see in the Issue.  This is what I have:
image

I assume this is correct.  I got this dialog by:
Administration > Role Management 
   Then click on the "Actions" item in the "Actions" menu on on the Superuser role.

jniles:
Yes, I made the change to the modal breadcrumb.  I wanted to make it clear where this modal resides.

bors r+

PR #5485, opened 2021-03-08
Restore default aspect ratio for chartjs in report

mbayopanda:
The current chart aspct ratio break the report : `Rapport graphique des mouvements`

image


jniles:
It looks straight forward enough.

Bors r+

PR #5488, opened 2021-03-09
Limit stock sheet link to data of the current year

### LINKED ISSUE #5476: Make linked stock sheet from articles in stock use current year
Currently, the stock sheet linked from the articles in stock registry doesn't limit any dates.  This is not exactly ideal.  It would be nice if it only loaded a limited dataset, so as not to ask too much from the server.  Let's limit that to a year for the moment.

mbayopanda:
This PR is related to #5476 

Close #5476

### DIFF
--- a/client/src/js/services/util.js
+++ b/client/src/js/services/util.js
@@ -39,7 +39,6 @@ function UtilService(moment) {
     hiddenElement.click();
   };
 
-
   /** @todo comments showing usage */
   service.filterFormElements = function filterFormElements(formDefinition, requireDirty) {
     const response = {};
@@ -129,7 +128,6 @@ function UtilService(moment) {
     };
   };
 
-
   /**
    * @function before
    *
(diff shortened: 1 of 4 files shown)


jmcameron (commented):
These changes look fine and I tested this with Bhima_test and works fine.

However, I wonder whether this is the right thing to do?   If you do the query in January, you only get a report for a few weeks.   It seems to me if someone does this report they would like it to cover a period they could specify.   So it seems to me that we should pass the default filter "Period Limit" to the report and just let the Stock Sheet report cover the same period that hte "Aritcles in Stock" page covers (as is set in the Default Filter under the Search modal)..

jniles:
@jmcameron could you pull this PR and try making those changes and play around with it?  My intuition tells me that the period will be a negative user experience, since the articles in stock registry doesn't really contain chronological information (as opposed to so many of our other registries).

The main goal of this PR and issue #5476 is a performance improvement.  If you think that we can also get a better UX while achieving the same goal, feel free to propose it.

jniles:
Closed in #5493.

PR #5491, opened 2021-03-10
fix(cash): correctly calculate opening balance

jniles:
This commit corrects a bug in the opening balance calculation of the cash report.  The opening balance also included current day, leading to some very hard to detect bugs.  This commit simply removes the `maxDate` option from this report.


mbayopanda (approved):
LGTM 👍🏽 

bors r+

jniles:
My assumption is that this failed because it was run at midnight.  Trying again:

bors r+

PR #5492, opened 2021-03-10
fix(stock): count initial days as stock out

### LINKED ISSUE #5482: Should we consider "before depot had stock" as stock out?
Here is an interesting case from Bandunduville:

image

You'll notice that the period of CMM is 6 months, yet they were only installed this year (2021).  The first stock movement, presumably an entry, was made in February 2021.  That means, we are treating about 4ish months where the depot wasn't operation as days with stock.

Interestingly, if I move the number of months down to 3 months for the CMM calculation:

image

This doubles the CMM by halfing the period.  This doesn't seem right to me.

Should we consider "days before the depot had stock" as "days of stock out"?  This would fix the above issue.  But it might also freak users out because they were not technically in stock out.  Or should we modify the CMM algorithm by MSH to account for this?  Or should we just tell users that their CMM might not be right because the period is too large?

jniles:
This commit changes the CMM calculation to ensure that _head_days is
added to the _sum_stock_out_day variable when there are no previous
records.  This can happen when a depot has just newly opened and the
date range used for calculating the CMM is greater than the number of
days the depot has been operating.

Closes #5482.

### DIFF
--- a/server/models/procedures/stock.sql
+++ b/server/models/procedures/stock.sql
@@ -598,13 +598,15 @@ CREATE PROCEDURE GetAMC(
     sms.sum_quantity INTO _initial_quantity
   FROM stock_movement_status AS sms WHERE
     -- get the record right before the start
-    sms.date = (SELECT MAX(date) FROM stock_movement_status ss WHERE ss.date <= _start_date AND ss.depot_uuid = _depot_uuid AND ss.inventory_uuid = _inventory_uuid) AND
+    sms.date < _start_date AND
     sms.depot_uuid = _depot_uuid AND
-    sms.inventory_uuid = _inventory_uuid;
+    sms.inventory_uuid = _inventory_uuid
+    ORDER BY sms.date DESC
+    LIMIT 1;
 
   SELECT DATEDIFF(_min_date, _start_date)  INTO _head_days;
 
-  IF (_initial_quantity = 0) THEN
+  IF (_initial_quantity = 0 OR _initial_quantity IS NULL) THEN
     -- started without stock. Add the "head" to  days to stock out days.
     SELECT _sum_stock_out_day + _head_days INTO _sum_stock_out_day;
   ELSE


server/models/procedures/stock.sql
```diff
@@ -598,13 +598,15 @@ CREATE PROCEDURE GetAMC(
     sms.sum_quantity INTO _initial_quantity
```
jmcameron, line 598: It looks like the previous comment could be tightened up.

server/models/procedures/stock.sql
```diff
@@ -598,13 +598,15 @@ CREATE PROCEDURE GetAMC(
...
   FROM stock_movement_status AS sms WHERE
     -- get the record right before the start
-    sms.date = (SELECT MAX(date) FROM stock_movement_status ss WHERE ss.date <= _start_date AND ss.depot_uuid = _depot_uuid AND ss.inventory_uuid = _inventory_uuid) AND
+    sms.date < _start_date AND
```
jmcameron, line 601: This query looks good.

server/models/procedures/stock.sql
```diff
@@ -598,13 +598,15 @@ CREATE PROCEDURE GetAMC(
...
     sms.depot_uuid = _depot_uuid AND
-    sms.inventory_uuid = _inventory_uuid;
+    sms.inventory_uuid = _inventory_uuid
+    ORDER BY sms.date DESC
+    LIMIT 1;
 
   SELECT DATEDIFF(_min_date, _start_date)  INTO _head_days;
 
-  IF (_initial_quantity = 0) THEN
+  IF (_initial_quantity = 0 OR _initial_quantity IS NULL) THEN
```
jmcameron, line 609: SQL question:  Does "LIMIT 1" produce an error if there are no rows to return?

jmcameron, line 609: When will _initial_quantity be NULL?

jmcameron (commented):
I was not sure how to test this.
The code looks good, though.

server/models/procedures/stock.sql
jniles: Yeah, that was the bug.  If no records match, even with an `IFNULL()` call, it still returns NULL.  So, if there is no data, it's NULL.

jniles: Nope!  Seems to behave fine.

jniles:
@jmcameron here is how to test this:

1. On `master`, run the integration tests (`yarn test:integration`).
2. Go the the Stock Settings module.  Set the date range for the CMM calculation to 1000 months.
2. Go the the Articles in Stock Registry.
3. Open the AMC/CMM calculations for `DDIS_IODP1S2_0` in the teritary depot.
4. Here is what it looks like:
image
5. Note that this is counting the days as days in stock.


------


1. On **this** branch, repeat all the above steps.
2. You should now get the following:
image

server/models/procedures/stock.sql
jmcameron: Good.  I was not sure.

jmcameron:
Not sure what is going on.  But the suggested test is not working for me.  Here is a screenshot of master (bhima_test after running 'yarn test:integration'. cmm period=10 mo):
image

Now with CMM=1000mo:
image

Switch to this PR with CMM=1000 mo.
image

It does not appear to have handled the "Stock Outs" days correctly.

jniles:
Did you rerun the integration tests between the branch switch?

I've changed the stored procedure so you either need to rerun the integration tests (which rebuilds the stored procedures again), or just drop the old `master` procedure and build this one's `GetAMC()` procedure.

jmcameron:
> Did you rerun the integration tests between the branch switch?

Uhh... No.  I'll try it again.

jmcameron:
> > Did you rerun the integration tests between the branch switch?
> 
> Uhh... No. I'll try it again.

That fixed it.  Looks good.  Here is the result:
image

Looks like it is ready to go.

PR #5493, opened 2021-03-10
Refactor stock sheet report to use the current period filter for its date range

jmcameron:
Refactor stock sheet report (fiche de stock) to use the current period filter for the date range of the report.

This PR absorbs PR https://github.com/IMA-WorldHealth/bhima/pull/5488 as suggested by @jniles in https://github.com/IMA-WorldHealth/bhima/pull/5488#issuecomment-794954182.

This update passes the value of the period filter (from the default filters) to specify the date range for the report.
This done as a separate PR so that it could be tested without modifying the previous PR.

**Notes**
- if the 'allTime' option is used, no start  or end dates are printed on the report.  This could be improved by including a message such as "All stock data until the present" or "All stock data until <todays-date>"  Open to suggestions on this.
- Since there may be some confusion, I added a tooltip for the Articles In Stock action menus for the "Stock Sheet" command.

**Testing**
- Please try all the options for the period (in the [Search] -> DefaultFilters tab) and verify that they all produce report dates and contents that make sense.

### DIFF
--- a/client/src/modules/stock/inventories/templates/inventory.action.html
+++ b/client/src/modules/stock/inventories/templates/inventory.action.html
@@ -42,13 +42,17 @@
     <li class="divider"></li>
 
     <li>
-      <a data-method="view-amc-calculations" href ng-click="grid.appScope.viewAMCCalculations(row.entity)">
+      <a data-method="view-amc-calculations" href
+        ng-click="grid.appScope.viewAMCCalculations(row.entity)">
         <i class="fa fa-newspaper-o"></i> <span translate>INVENTORY.OPEN_AMC_CALCULATION</span>
       </a>
     </li>
 
     <li>
-      <a ng-href="/reports/stock/sheet?{{ grid.appScope.openStockSheetReport(row.entity) }}" target="_blank">
+      <a uib-tooltip="{{ 'REPORT.STOCK.INVENTORY_REPORT_TOOLTIP' | translate }}"
+        tooltip-placement="left"
+        ng-href="/reports/stock/sheet?{{ grid.appScope.openStockSheetReport(row.entity) }}"
+        target="_blank">
         <span class="fa fa-file-pdf-o"></span> <span translate>REPORT.STOCK.INVENTORY_REPORT</span>
       </a>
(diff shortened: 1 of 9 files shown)


client/src/modules/stock/inventories/templates/inventory.action.html
```diff
@@ -42,13 +42,16 @@
...
-      <a data-method="view-amc-calculations" href ng-click="grid.appScope.viewAMCCalculations(row.entity)">
+      <a data-method="view-amc-calculations" href
+        ng-click="grid.appScope.viewAMCCalculations(row.entity)">
         <i class="fa fa-newspaper-o"></i> <span translate>INVENTORY.OPEN_AMC_CALCULATION</span>
       </a>
     </li>
 
     <li>
-      <a ng-href="/reports/stock/sheet?{{ grid.appScope.openStockSheetReport(row.entity) }}" target="_blank">
+      <a uib-tooltip="{{ 'REPORT.STOCK.INVENTORY_REPORT_TOOLTIP' | translate }}"
```
jniles, line 52: Can you make this jut off the left side instead of putting it over the menu?  `uib-tooltip-placement="left"`

image

jniles (changes requested):
Okay, so I've tested this and it wasn't as bad as I thought it would be.  An added advantage is that we can look at previous periods, such as "last month", which is something I didn't consider.

I'm happy with hiding the date range if "all time" is selected.

One issue pointed out with the tooltip, otherwise, I'm happy to merge this.

client/src/modules/stock/inventories/templates/inventory.action.html
jmcameron: Fixed

jmcameron: The fix was:   tooltip-placement="left"  (not uib-tooltip-placement)
and it worked nicely.    I liked it so much I went back and tweaked the menu for auto merging lots in the Find Duplicate Lots page!

jniles (approved):
Looks good to me!

bors r+

Issue #2158, opened 2017-09-27
design: Changing a Debtor's Debtor Group


jniles:
This is a corner case we haven't properly considered:  what happens when a patient in billed in one debtor group, but it is discovered that they are not supposed to be in it?  An example of this would be if a patient were invoiced as a "convention" patient (like https://github.com/IMA-WorldHealth/bhima-2.X/pull/2152 intends to implement), but they are later discovered to be a cash-paying patient?

**Changing a Debtor's Debtor Group**
In my view, the following should happen when a user tries to change the debtor group of a patient.
 - Check the balance on the Patient's account.  If the balance is 0, go ahead and change the debtor group.  If the balance <> 0, block the operation.  Only once a patient has no debt and/or they are not owed any money can their account be changed.
 
In the future, a cool feature might be to migrate all the debtor's financial history to reflect the change in their debtor group (and therefore their accounts).  This would have to be thoroughly designed first -- and if it resulted in changing the General Ledger this is not desirable.

jniles:
So, this has bit Vanga pretty hard and caused a lot of pain points.  I say we block the changing of groups of patients with a non-zero balance as a first step.

We should mention clearly to the user that we cannot change their group because they have a debt of X values on their account.

lomamech:
I have just read the issue and I think this option should be lifted if the Patient has to leave a debtor group (Conventioned) to a group that is not

PR #5496, opened 2021-03-11
feat(patients): group changes on financial status

### LINKED ISSUE #5487: Display changes in debtor group on the Patient Financial Activity
We should show when a patient changes debtor group in-line on the Patient Financial Activity report.

There was a particularly hairy analysis that I needed to perform for our partners in Vanga because of patients that had been invoiced in multiple debtor groups.  This was not at all evident from their financial activity history and it _looked_ like they had done all their transactions correctly in the debtor group's account.

Let's work around this by making it very obvious when the patient changes a debtor group by dedicating an entire `<tr>` to the fact that the debtor changed groups within that time.  It should be something like `old group` &#8594; `new group` on its own line.

jniles:
Adds the debtor group changes onto the patient financial activity as requested by our friends in Vanga.

Closes #5487.

Here is what it looks like:
image

### DIFF
--- a/server/controllers/finance/debtors/groups/index.js
+++ b/server/controllers/finance/debtors/groups/index.js
@@ -1,4 +1,3 @@
-
 /**
 * The Debtor Groups Controllers
 *
@@ -42,6 +41,7 @@ exports.invoices = invoices;
 exports.lookupDebtorGroup = lookupDebtorGroup;
 
 exports.history = history;
+exports.getDebtorGroupHistory = getDebtorGroupHistory;
 
 /**
  * Looks up a debtor group in the database by uuid.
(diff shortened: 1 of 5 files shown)


lomamech (commented):
image
I have just tested it by making several change of debtor group and it works well but I think that we could add in the patient's financial activity report the possibility of being able to select a debtor group

lomamech:
> image
> I have just tested it by making several change of debtor group and it works well but I think that we could add in the patient's financial activity report the possibility of being able to select a debtor group

@jniles  Because I tell myself that users might ask for it someday, what do you think

lomamech:
@jniles What do you think of the idea of updating the option to view the report by specifying the debtor group

jniles:
> @jniles What do you think of the idea of updating the option to view the report by specifying the debtor group

I don't think that makes sense.  What information would that tell you?

Ideally, when a debtor changes group they should have their debt written off and that old account closed.  See https://github.com/IMA-WorldHealth/bhima/issues/2158.  So, even if the debtor group changed back then, it doesn't garantee that any of that debt is still in their account.  I'm not sure what a report filtered by debtor group would tell you.

lomamech:
> > @jniles What do you think of the idea of updating the option to view the report by specifying the debtor group
> 
> I don't think that makes sense. What information would that tell you?
> 
> Ideally, when a debtor changes group they should have their debt written off and that old account closed. See #2158. So, even if the debtor group changed back then, it doesn't garantee that any of that debt is still in their account. I'm not sure what a report filtered by debtor group would tell you.

One day users might ask for it

PR #5497, opened 2021-03-11
Enable filtering on status in Articles in Stock

### LINKED ISSUE #5478: Articles in Stock status tags
In the articles in stock, the status tags (Over Maximum, Critical Stock, Stock Out, etc) are cut on the right in the narrow column for Status.  One way to address this (without changing column widths) is to add a tooltip on them so when you hover over them, you see the full text

Also, I notice that the grid filtering option does not allow filtering on the Status column.   If that is not too difficult to fix, it might be useful.

jmcameron:
This PR also updates the stock status label coloring per https://github.com/IMA-WorldHealth/bhima/issues/5478#issuecomment-796820000.   This aligns most of the coloring between the Articles in Stock page and the Lots registry page (except that red is used for stock-out in the Articles in Stock.  But in the Lots registry, red is used for expired stock and purple is used for stock-out.  That still needs to be aligned, but was not covered in this PR.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5478

**Testing**
- You may want to use a production database to see all of the possible values.
- Update default filters to allow showing exhausted lots
- In the Articles in Stock page, enable filtering and enter a value in the "Status" filter to observe that it does filter correctly.
- Try all options:  Stock Out (red), Critical Stock (orange), Low Stock (yellow),  Over Maximum (blue), In Stock (green)
- Try these in English as French (use the French status names)

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -900,7 +900,7 @@ a[disabled] {
 }
 
 .icon-at-risk-of-stock-out {
-  color: #FFFFA9 !important;
+  color: #FFFF00 !important;
 }
 
 .at-risk-of-expiring {
@@ -909,6 +909,21 @@ a[disabled] {
 
 .icon-at-risk-of-expiring {
   color: #FFB766 !important;
+}
+
+.out-of-stock-status {
+  color: black !important;
+  background-color: red !important;
+}
+
+.critical-stock-status {
+  color: black !important;;
+  background-color: #FFB766 !important;
+}
+
+.low-stock-status {
+  color: black !important;;
+  background-color: #FFFF00 !important;
 }
 
 .out-of-stock {
--- a/client/src/js/services/StockService.js
+++ b/client/src/js/services/StockService.js
@@ -77,7 +77,7 @@ function StockService(Api, StockFilterer, HttpCache, util, Periods) {
 
   // stock status label keys
   const stockStatusLabelKeys = {
-    stock_out          : 'STOCK.STATUS.STOCK_OUT',
+    stock_out         : 'STOCK.STATUS.STOCK_OUT',
     in_stock          : 'STOCK.STATUS.IN_STOCK',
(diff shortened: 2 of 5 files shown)


jniles (approved):
This looks good to me!  Tested with the Vanga database and works as described.  Great job!

bors r+

jniles:
Hmm... this is a fluke, but a fluke that should have been caught.

@jmcameron do you want me to help you track this down and fix it here, or would you prefer we merge this and tackle it in its own PR?  It's clearly not related to this PR.

Failure: https://ima-worldhealth.semaphoreci.com/jobs/a1259e53-c910-4a9c-abb9-9b52f841a844

jniles:
@jmcameron I think I got this fixed.  Take a glance at it.  The main issue was that promises were not being returned in the massive `.then()` chains (there should have been `return db.exec()`s) that we didn't catch the first time around.

Instead of doing that, I decided to make it into `async/await` which is easier to read with something that long and complicated.

Ball's in your court.

bors delegate+

lomamech (approved):
I just did the review, on display in the articles in stock but I did not see the tooltip on the status tags

lomamech:
@jmcameron let me know if this is not an omission

jmcameron:
> I just did the review, on display in the articles in stock but I did not see the tooltip on the status tags

That is not implemented in this PR.  But I think it should be done, so I'll create an issue for it.

-Jonathan

lomamech:
> That is not implemented in this PR. But I think it should be done, so I'll create an issue for it.
> 
> -Jonathan

Thank you for that clarification

jmcameron:
I just did another check on this after @jniles updates.  The functionality works the same, and the tests seem to work (other than a fail in the end-to-end tests that I've been having on my system recently that is not related to this PR).  So it looks good!

jmcameron:
> I just did the review, on display in the articles in stock but I did not see the tooltip on the status tags

Thanks for the review!

PR #5506, opened 2021-03-12
Prevent expired lots from being selected in stock entry

### LINKED ISSUE #5500: Do not propose expired lots during stock entry during
When we enter stock in BHIMA, we automatically detect the lots associated with a particular inventory item.  However, it sometimes exposes expired stock lots.  We should at minimum make this optional for entries from purchases.  Here is an example:

image

We don't want people accidentally selecting expired stock just because the system proposed it.

jmcameron:
When selecting lots via the typeahead entry field for  Stock Entry, prevent expired lots from being selected.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5500

**Testing**
- Use bhima_test
- Using master, go to Stock Entry (integration)
    - Chose the Depot Principal
    - Use stock code QUIN1 (should give only one choice, select it)
    - Click on the "Lots" field on that row to get the lot selection modal
    - Type 'q' into the "Lot" field.  Observe that it shows 'QUININE-A' and that it is expired
- Switch to this PR
    - Repeat the process above but note that 'QUININE-A' is no longer available
    - Try typing in QUININE-A and hitting return or changing to another field.  Notice that the form will not allow it to be used.
- Close the lot selection modal and go back to Stock Entry page
    - Change the entry date to something in 2017 (or earlier)
    - Go back to the lots selection page (Click "Lots" link)
    - Notice that the entry date at the top of the page is the new date
    - Now you should be able to select QUININE-A and everything should work
- Close the lot selection modal and change the entry date back to today
    - Go back to the lots selection page (Click "Lots" link)
    - Notice that the QUININE-A lot is now expired and the form will complain about it

### DIFF
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -131,7 +131,34 @@ function StockEntryController(
    * @description on change in bhDateEditor component update the date
    */
   function onDateChange(date) {
+    if (typeof date === 'undefined') {
+      // Ignore intermediate invalid date strings (eg, 15/03/201)
+      return;
+    }
     vm.movement.date = date;
+
+    // Check all stock lines and their lots to make sure that all
+    // the lot expiration statuses are still
+    vm.stockForm.store.data.forEach(stockLine => {
+
+      // Check the expiration status of all previously selected lots
+      if (stockLine.lots) {
+        stockLine.lots.forEach(lot => {
+          if (lot.expiration_date > vm.movement.date) {
+            lot.expired = true;
+            lot.isValid = false;
+            lot.isInvalid = true;
+            stockLine.isValid = false;
+          }
+        });
+      }
+
+      // Update expiration status of the available and candidate lots
+      stockLine.availableLots.forEach(lot => {
+        lot.expired = lot.expiration_date > vm.movement.data;
+      });
+      stockLine.candidateLots = stockLine.availableLots.filter(lot => !lot.expired);
+    });
   }
 
   /**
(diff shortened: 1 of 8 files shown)


jniles (commented):
@jmcameron how does this handle back-dating the movement to before the lot expired?  I think that it will filter out lots that expired yesterday even if we are trying to make a movement from a month ago.

Can you take a look at this case?

jmcameron:
> @jmcameron how does this handle back-dating the movement to before the lot expired? I think that it will filter out lots that expired yesterday even if we are trying to make a movement from a month ago.
> 
> Can you take a look at this case?

Good catch!   I'm working on this.  What happens and how should we handle the user changing the movement date after lots have been selected?

jmcameron:
> > @jmcameron how does this handle back-dating the movement to before the lot expired? I think that it will filter out lots that expired yesterday even if we are trying to make a movement from a month ago.
> > Can you take a look at this case?
> 
> Good catch! I'm working on this. What happens and how should we handle the user changing the movement date after lots have been selected?

I suggest clearing all the lots if the movement date is changed to avoid any mixed-date issues.   What do you think?

jniles:
@jmcameron that would work, but ideally we could refactor the check into it's own function and run validation on submit.  You could then just update the symbol on the right to signal which line has a problem.  Something like this:

image


Would that be possible?  It would be a better user experience.

jmcameron:
> @jmcameron that would work, but ideally we could refactor the check into it's own function and run validation on submit. You could then just update the symbol on the right to signal which line has a problem. Something like this: 
> Would that be possible? It would be a better user experience.

Ok. That sounds good.  I'll figure it out.

client/src/modules/stock/entry/entry.js
```diff
@@ -602,20 +635,22 @@ function StockEntryController(
...
   function buildStockLine(line) {
     const inventory = inventoryStore.get(line.inventory_uuid);
+    const entryDate = vm.movement.date || Date();
     line.code = inventory.code;
     line.label = inventory.label;
     line.unit_cost = inventory.price;
     line.quantity = 0;
     line.cost = line.quantity * line.unit_cost;
-    line.expiration_date = new Date();
+    line.expiration_date = entryDate;
```
jniles, line 644: 👍

jniles (approved):
The code looks fine to me, and I've tested the behavior.  I'm happy with these changes.

You had some questions about `tracking_expiration`, but I think this is fine.  The goal of `tracking_expiration` is just to ignore expiration dates in certain scenarios.   To get around having too much form logic, we just set it to a really far future date.  That way we won't fail any expiration rules.

We should probably do an extensive rewrite of a lot of these form logic components.  They should be encapsulated into a service that we can unit test.  It's getting a bit hairy to rationalize without that security.  That will obviously be for a future PR.

jniles:
Based on interaction in the BHIMA chat, I'm merging this.

PR #5509, opened 2021-03-14
feat(stock): add user to stock movement registry

### LINKED ISSUE #5502: Let's add the username to the stock movement registry
Here is what the stock movement registry looks like:
image

I want to know who did these movements.  We typically include the user's name somewhere on these registries.  Let's add them here so that a user can tell at a glance the movements made by each person.

jniles:
Adds the user to the stock movement registry so that we are better able to track which user performed which movements.

Closes #5502.

Here is what it looks like:
image

### DIFF
--- a/client/src/modules/stock/movements/registry.js
+++ b/client/src/modules/stock/movements/registry.js
@@ -112,6 +112,10 @@ function StockMovementsController(
       cellFilter : 'date',
       cellClass : 'text-right',
     }, {
+      field : 'userName',
+      displayName : 'FORM.LABELS.USER',
+      headerCellFilter : 'translate',
+    }, {
       field : 'action',
       displayName : '',
       enableFiltering : false,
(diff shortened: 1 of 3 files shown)


jmcameron (approved):
The code is good.

It works as advertised, but the PDF report of stock movements does not contain the "doing" username (CSV/Excel include it).  Do you want to do that as part of this PR or as a separate issue?

Otherwise it is ready to go.

jniles:
@jmcameron I've made the changes to the PDF report.  To do so, I've reworked it a little so everything fits on one page.  It used to be:

image

And then I tried:

image

But some rows overflowed, so I've settled on combining the entry/exit and the "reason" lines:

image


Let me know what you think.

jmcameron (approved):
I checked the PDF and your update is good.

One idea for a future issue (or to be tacked onto this one):  When you record a movement "To Depot", it might be useful to print the name of the Depot it was transferred to.
image

Your call.  It is ready to release is.

jniles:
Bors r+

jniles:
@jmcameron thanks for the review!  Indeed, it would be good to include more information in the reason column.  This is somewhat alluded to in https://github.com/IMA-WorldHealth/bhima/issues/5415.  It's an expensive query though, so I'm picking off low-hanging fruit until then.

PR #5511, opened 2021-03-14
Remove stock movement

### LINKED ISSUE #5435: Allow users to delete stock movements
We allow users to delete most kinds of financial movements.  It would be ideal if we could allow users to delete stock movements.  This is conceptually simpler than reversal since reversing would require both finance and stock calculations to be recomputed.

This is a placeholder issue until I have time to flesh out the proposal.

mbayopanda:
This PR is related to #5435 

Close #5435


client/src/modules/stock/movements/registry.js
```diff
@@ -265,6 +265,15 @@ function StockMovementsController(
     vm.latestViewFilters = stockMovementsFilters.formatView();
   }
 
+  vm.deleteMovement = documentUuid => {
+    return Stock.movements.delete(documentUuid)
```
jniles: We typically have a confirmation modal for removing records.  For example:
https://github.com/IMA-WorldHealth/bhima/blob/03ccc62db8b09b82a2a09652ca196c592bb17b69/client/src/modules/cash/payments/registry.js#L215-L222

https://github.com/IMA-WorldHealth/bhima/blob/03ccc62db8b09b82a2a09652ca196c592bb17b69/client/src/modules/invoices/registry/registry.js#L227-L233

Could you add a similar check here?  I can imagine people quite easily running into issues by accidentally deleting movements.

client/src/modules/stock/movements/templates/action.cell.html
```diff
@@ -29,5 +29,11 @@
         <span class="fa fa-file-text-o"></span> <span translate>REPORT.VIEW_INVOICE</span>
       </a>
     </li>
+
+    <li>
+      <a data-method="delete" class="text-danger" ng-click="grid.appScope.deleteMovement(row.entity.document_uuid)" href>
```
jniles: Remember `document_uuid` isn't unique to a movement (https://github.com/IMA-WorldHealth/bhima/issues/4083).  So we may end up deleting more than one movement here.

I think that the easiest way to address this is to simply pass the `is_exit` flag in the HTTP request for now.  Please put a TODO in the code to revisit this when we have a solution to https://github.com/IMA-WorldHealth/bhima/issues/4083.

server/controllers/stock/index.js
```diff
@@ -407,6 +409,63 @@ async function createMovement(req, res, next) {
...
+  const movementDeletionQuery = `
+    DELETE FROM stock_movement WHERE document_uuid = ?;
+  `;
+
+  // simulate movement deletion
+  tx.addQuery(movementDeletionQuery, [identifier]);
+
+  const deleteLots = `
+    DELETE FROM lot WHERE uuid IN (
+      SELECT uuid FROM stock_movement WHERE document_uuid = ? AND flux_id IN (1, 6, 13)
```
jniles: After changes by @jmcameron, we now "link" stock lots to stock movements if the lot already exists.  So, instead of deleting the `lot`, check first that they do not appear in the `stock_movement` table.  If they no longer are referenced, we assume this was the movement that created them and can safely remove them.

server/controllers/stock/index.js
```diff
@@ -407,6 +409,63 @@ async function createMovement(req, res, next) {
...
+    );
+  `;
+
+  // delete lots from (purchase, donation and integration)
+  tx.addQuery(deleteLots, [identifier]);
+
+  // recompute stock movement status
+  // NOTE : this operation can take a lot of minutes
+  // tx.addQuery('CALL zRecomputeStockMovementStatus();');
```
jniles: Yeah, this is the right idea, but the wrong function.  Instead, you'll want to look up all the inventory_uuids, the depot_uuid, and the date. You can then just use this function to recompute the StockMovementStatus():
https://github.com/IMA-WorldHealth/bhima/blob/cbda958b40451ff2ad5bae4302cf3893caffb96b/server/controllers/stock/index.js#L159-L186

server/controllers/stock/index.js
```diff
@@ -407,6 +409,63 @@ async function createMovement(req, res, next) {
...
+
+  try {
+    await tx.execute();
```
jniles, line 448: What happens if `enable_auto_stock_accounting`, but the data is already posted into the `general_ledger`? 

I think that this code will remove the quantities in stock, but not the stock value.  Instead, let's make this one transaction and break the whole thing if the data has already been posted.  It's not ideal ... but it will save a lot of pain later.

server/controllers/stock/index.js
```diff
@@ -407,6 +409,63 @@ async function createMovement(req, res, next) {
...
+      `;
+      const records = await db.exec(findTransactionInJournal, [identifier]);
+      const dbPromise = [];
+      records.forEach(item => {
+        const params = [
+          item.record_uuid,
+          req.session.user.id,
+          item.description,
+        ];
+        dbPromise.push(journal.reverseTransaction(...params));
```
jniles: Let's remove this, not reverse them.  Otherwise, it will impact our CMM, the compte d'exploitation, weighted average cost, etc ...

server/controllers/stock/index.js
```diff
@@ -407,6 +409,63 @@ async function createMovement(req, res, next) {
...
+      });
+      await Promise.all(dbPromise);
+    }
+
+    res.sendStatus(200);
```
jniles: This should return 201.

server/controllers/stock/index.js
```diff
@@ -407,6 +409,63 @@ async function createMovement(req, res, next) {
...
+
+    if (stockSettings.enable_auto_stock_accounting) {
+      // reverse coresponding vouchers
+      // NOTE : posted transaction in the general ledger are not reversed
+      const findTransactionInJournal = `
+      SELECT DISTINCT record_uuid, description FROM posting_journal WHERE reference_uuid = ?
+      `;
+      const records = await db.exec(findTransactionInJournal, [identifier]);
+      const dbPromise = [];
+      records.forEach(item => {
```
jniles: Remember, every time you have code of this format:
```js
const array = [];
records.forEach(row => {
  // do something
  array.push(row);
 });
 ```
 You are doing a `map`!  It's easier to read if you use a map:
 ```js
// this is equivalent to what you wrote above
const dbPromise = records
  .map(item => journal.reverseTransaction(item.record_uuid, req.session.user.id, item.description);
```

server/config/routes.js
```diff
@@ -813,6 +813,7 @@ exports.configure = function configure(app) {
 
   app.post('/stock/lots/movements', stock.createMovement);
   app.get('/stock/lots/movements', stock.listLotsMovements);
+  app.delete('/stock/lots/movements/:document_uuid', stock.deleteMovement);
```
jniles: We have a `/stock/movements` API.  Why put this under `stock/lots/movements`?

client/src/modules/stock/movements/registry.js
```diff
@@ -265,6 +265,15 @@ function StockMovementsController(
...
+      .then(() => {
+        Notify.success('STOCK.SUCCESSFULLY_DELETED');
+        startup();
```
jniles: You could just call `load(stockMovementsFilters.formatHTTP(true));` directly.  That's what `startup()` will do.

client/src/modules/stock/movements/templates/action.cell.html
jniles: You should check if record deletion is allowed before showing this option to users.

server/controllers/stock/index.js
```diff
@@ -407,6 +409,63 @@ async function createMovement(req, res, next) {
   }
 }
 
+/**
+ * @method deleteMovement
+ * @desc perform a fake stock movement deletion based on its document uuid
```
jniles: What is a fake movement deletion?

jniles (commented):
@mbayopanda awesome!  This is a good first step.  I've made quite a few comments - I know you changed the design half-way through, so a lot of my comments are due to that.

Here are the main points:
  1. Could you include some integration tests?
  2. I think we really should delete the voucher if possible, not just the posting_journal transaction.  You can use the `safelyDeleteVoucher` function exposed in the vouchers controller on the server.  It should already do all the hard work for you.
  3. We should ask for user confirmation before deleting something.  That way nothing happens by accident.

client/src/modules/stock/movements/templates/action.cell.html
jniles: Also, I just noticed this, but the `text-danger` needs to be on the `<span>` on the inside of the `<a>`.  Otherwise, it isn't red.

mbayopanda:
@jniles changes landed

server/controllers/stock/index.js
```diff
@@ -407,6 +409,72 @@ async function createMovement(req, res, next) {
...
+      // find transaction's record_uuid from journal
+      // safely delete voucher related to record_uuid found
+      const findTransactionInJournal = `
+      SELECT DISTINCT record_uuid
+      FROM posting_journal 
+      WHERE reference_uuid = ?
+      `;
+      const records = await db.exec(findTransactionInJournal, [identifier]);
+      const dbPromise = records.map(item => vouchers.safelyDeleteVoucher(item.record_uuid));
+      await Promise.all(dbPromise);
```
jniles, line 460: I am still a little uncomfortable with this.  Why can't we block whole transaction if the value is posted?  It seems to me we prevent people from deleting vouchers if the value is posted, why wouldn't we prevent them from deleting stock movements if the value is posted?

jniles (approved):
We need to investigate some issues with the stock movement status not being updated in some of these tests.  But the functionality described here works fine.

bors r+

Issue #5379, opened 2021-02-11
Make deleting records (invoices/cash payments/vouchers) into a Role Action


jniles:
We have certain "role actions" that allow the users in possession of this role to do super user things.  For example, unposting transactions.  Here is the presentation of the modal:
![image](https://user-images.githubusercontent.com/896472/107614402-8c2bea00-6c4a-11eb-9f76-d1939e1541f3.png)

I propose we make the current enterprise setting for `enable_delete_records` into a Role Action.  It is currently an enterprise setting:
https://github.com/IMA-WorldHealth/bhima/blob/1f0e637814c67cc04f9ac0ae99744e5a16d7f8b4/server/controllers/admin/enterprises.js#L47

This would do two things:

1. It would allow you to have some users (like an IMA user) have the permission to delete records _without_ turning it on for everyone else.  This may be very useful during training exercises, for example.
2. It would elevate the privileges required to toggle this setting from "anyone who has access to the enterprise module" to "anyone who has access to roles".  This is presumably a stricter setting.

PR #5513, opened 2021-03-14
feat(purchase): remove unconfirmed purchase orders

### LINKED ISSUE #5490: Allow users to delete unconfirmed purchase orders
Similar to https://github.com/IMA-WorldHealth/bhima/issues/5489, we should allow users to delete unconfirmed purchase orders.  This would help us clean up the junk that users create and never get executed.  It's important that this only applies to unconfirmed purchase orders - confirmed purchase orders are considered contracts that should be executed.

jniles:
Adds the ability to delete unconfirmed purchase orders if they are not confirmed.  This is only active if the general "allow record deletion" ability is turned on.  It will only be useful when #5379 is implemented.

I've also renamed "purchase/list" to "purchase/registry" since it is incredible confusing as named.

Closes #5490.

WtP4UuoSlc


server/controllers/finance/purchases.js
```diff
@@ -533,6 +535,31 @@ function purchaseBalance(req, res, next) {
...
+ * @decription
+ * This function allows users to delete purchase orders that are awaiting confirmation.
+ */
+async function remove(req, res, next) {
+  const pid = db.bid(req.params.uuid);
+
+  try {
+    const record = await db.one('SELECT * FROM purchase WHERE uuid = ?', pid);
+
+    const WAITING_CONFIRMATION = 1;
```
jmcameron, line 550: Good check!

server/controllers/finance/purchases.js
```diff
@@ -533,6 +535,31 @@ function purchaseBalance(req, res, next) {
...
+    if (record.status_id !== WAITING_CONFIRMATION) {
+      throw new BadRequest('Cannot remove a purchase order that is confirmed');
```
jmcameron: Since I presume the status_id could contain more than 2 values (otherwise it would be a boolean, right?), it seems like this bad request warning should be something like:  "Can only remove purchases that have not been confirmed".

jmcameron (approved):
Looks good.

I did verify that the cascading deleted all the purchase order items associated with this PR (via direct SQL queries).    I wonder if it is worth checking that in the integration test?

server/controllers/finance/purchases.js
jniles: Yeah, this a good suggestion.  I'll make the change.

jniles:
Correct my understanding, but I think the integration tests are geared for the express/database interface.  A test like that goes under "unit test" in my brain.  Is it alright with you if I add a unit test for that?

jmcameron:
> Correct my understanding, but I think the integration tests are geared for the express/database interface. A test like that goes under "unit test" in my brain. Is it alright with you if I add a unit test for that?

That sounds fine.  Thanks!

jniles:
@jmcameron changes landed.

server/config/routes.js
```diff
@@ -669,6 +669,7 @@ exports.configure = function configure(app) {
   app.get('/purchases/purchaseState', purchases.purchaseState);
   app.get('/purchases/:uuid', purchases.detail);
   app.put('/purchases/:uuid', purchases.update);
+  app.delete('/purchases/:uuid/', purchases.remove);
```
jmcameron, line 672: Do you need/want the trailing slash in the URL?

test/server-unit/purchases/purchases.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    expect(purchase.cost).to.equal(300);
+
+    // okay, now that we know our data is built, actually test if the data is deleted
+    // correctly.
+    await db.exec(`DELETE FROM purchase WHERE uuid = HUID('${puid}');`);
+
+    const purchases = await db.exec(`SELECT * FROM purchase WHERE uuid = HUID('${puid}');`);
+    expect(purchases).to.have.length(0);
+    const items = await db.exec(`SELECT * FROM purchase_item WHERE purchase_uuid = HUID('${puid}');`);
+    expect(items).to.have.length(0);
```
jmcameron, line 37: Nice!  Glad to see that you verify that the PR items also get deleted.   It may not be necessary because the SQL table definitions enforce that, but I can imaging someone modifying the table definitions in the future and messing this up.  So I think this unit test is worth doing.

jmcameron (approved):
Looks good.

I ran the tests on my PC and the server, client, and integration unit tests ran fine.   The end-to-end test failed in the same way that master does currently (deleting employee fails).

test/server-unit/purchases/purchases.spec.js
jniles: Yeah, I agree.  I didn't really want to do it, but the peace of mind is well worth it.

server/config/routes.js
jniles: Eh, it doesn't make any difference.  ExpressJS has an option to ignore them by default.

PR #5514, opened 2021-03-14
feat(stock): link voucher from stock movement

### LINKED ISSUE #5505: Link voucher from stock movement if auto stock accounting is enabled
Here is the stock movement registry:

image

If we have "auto stock accounting" enabled, it would be really nice to link the voucher behind this movement directly from this movement.  That way, we can easily see the financial impact of the movement.

jniles:
Links the voucher from the stock movement if the enterprise has "auto_stock_accounting" enabled.  This should make it easier to
compare the movements of stock with the financial movements.

Closes #5505.

Qg1pwMiF4c


mbayopanda (approved):
Good job @jniles 

Its LGTM 👍🏽 

bors r+

PR #5521, opened 2021-03-17
fix(stock): stock movement status on delete

jniles:
Corrects the computation of stock_movement_status on deletion of a stock movement.  The previous version had too many calls to string -> binary conversion that caused matching issues.  Additionally, the date wasn't set correctly.

### DIFF
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -420,7 +420,8 @@ async function deleteMovement(req, res, next) {
   try {
     // movement details for future quantity update
     const movementDetailsQuery = `
-      SELECT DISTINCT m.depot_uuid, l.inventory_uuid, m.is_exit
+      SELECT DISTINCT BUID(m.depot_uuid) AS depot_uuid, BUID(l.inventory_uuid) AS inventory_uuid,
+        m.is_exit, m.date
       FROM stock_movement m
       JOIN lot l ON l.uuid = m.lot_uuid
       WHERE m.document_uuid = ?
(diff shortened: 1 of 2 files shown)


jniles:
@mbayopanda this seems to fix all stock_movement_status issues.   From the best I can tell `quantity` is consistently equal to the `GetAMC()` `quantity_in_stock` value.

server/controllers/stock/index.js
```diff
@@ -420,7 +420,8 @@ async function deleteMovement(req, res, next) {
   try {
     // movement details for future quantity update
     const movementDetailsQuery = `
-      SELECT DISTINCT m.depot_uuid, l.inventory_uuid, m.is_exit
+      SELECT DISTINCT BUID(m.depot_uuid) AS depot_uuid, BUID(l.inventory_uuid) AS inventory_uuid,
```
mbayopanda, line 423: 👍🏽

server/controllers/stock/index.js
```diff
@@ -463,7 +468,7 @@ async function deleteMovement(req, res, next) {
     const inventoriesByDepots = _.groupBy(movementDetails, 'depot_uuid');
     const inventoriesToUpdates = Object.keys(inventoriesByDepots).map(depot => {
       const inventories = inventoriesByDepots[depot].map(item => item.inventory_uuid);
-      return updateQuantityInStockAfterMovement(inventories, new Date(), depot);
+      return updateQuantityInStockAfterMovement(inventories, new Date(oldestDate), depot);
```
mbayopanda, line 471: 👍🏽 good catch

mbayopanda (approved):
LGTM 👍🏽

jniles:
Thanks for identifying the issues @mbayopanda 👍

PR #5525, opened 2021-03-18
feat(purchase): update purchase orders

### LINKED ISSUE #5489: Allow users to edit unconfirmed purchase orders
A vanga, we are running into lots of errors with our purchase orders.  Mainly, no one is looking at them before they are executed.  But even if they were to look at them, the only choices are "confirm" or "cancel".  Let's add the ability to edit the (unconfirmed) purchase order!

Basically, it would take you back to the "create purchase" page and fill out all the items that we had previously.  This would give you the ability to change the quantities, prices, and supplier.  This would fit a lot better with our flow of telling users to only confirm purchase orders that are valid with correct pricing information.

jniles:
This commit implements the ability to update non-confirmed purchase orders.  If a purchase order has been confirmed, the only thing that should be updated is the status (canceled).  This is yet to be discussed.

To modify a purchase order, one must go to the purchase registry and select "edit" on the purchase orders that is non yet confirmed.  This will take you to the regular purchase order creation page with all values filled out.

Closes #5489.

Here is what it looks like:

sXzDxvsomU

Note that this implementation breaks updating the status.  I'd like some help figuring out which status we should allow.  I'll fix that before a merge.

### DIFF
--- a/client/src/js/services/Store.js
+++ b/client/src/js/services/Store.js
@@ -11,10 +11,8 @@ angular.module('bhima.services')
 function StoreService() {
 
   /** @constructor */
-  function Store(options) {
-
+  function Store(options = {}) {
     // default to empty options
-    options = options || {};
 
     // default index and data
     this.index = {};
@@ -53,18 +51,18 @@ function StoreService() {
    * matches.  Otherwise, it returns undefined.
    */
   Store.prototype.get = function get(id) {
-    var key = this.index[id];
-    if (angular.isUndefined(key)) { return; }
+    const key = this.index[id];
+    if (angular.isUndefined(key)) { return undefined; }
     return this.data[key];
   };
 
   // return only the latest (indexed) copy of each data element
   Store.prototype.getAll = function getAll() {
-    var keys = Object.keys(this.index);
-    return keys.map(function (key) {
+    const keys = Object.keys(this.index);
+    return keys.map((key) => {
       return this.data[this.index[key]];
-    }.bind(this));
-  }
+    });
+  };
 
   /**
    * @method post
@@ -76,19 +74,17 @@ function StoreService() {
    * @param {Object} object - an object to be inserted into the store.
    */
   Store.prototype.post = function post(object) {
-    var data = this.data;
-    var index = this.index;
-    var identifier = this.identifier;
+    const { data, index, identifier } = this;
 
     // default to an empty array if data not provided
     if (!data) { this.data = []; }
 
-    var id = object[identifier];
+    const id = object[identifier];
 
     if (angular.isUndefined(id)) {
       throw new Error(
-        'Trying to insert an object without the identity property "' + identifier + '".\n' +
-        'Failing object: ' + JSON.stringify(object)
+        `Trying to insert an object without the identity property "${identifier}".\n`
+        + `Failing object: ${JSON.stringify(object)}`,
       );
     }
 
@@ -103,9 +99,8 @@ function StoreService() {
    *
    * @param {Object} object - an object to be inserted into the store
    */
-   Store.prototype.remove = function remove(id) {
-    var data = this.data;
-    var index = this.index;
+  Store.prototype.remove = function remove(id) {
+    const { data, index } = this;
 
     if (id in index) {
       data.splice(index[id], 1);
@@ -146,11 +141,11 @@ function StoreService() {
    * Recalculates the stores index when data is added/removed via other methods.
    */
   Store.prototype.recalculateIndex = function recalculateIndex() {
-    var data = this.data;
-    var index = this.index = {};
-    var identifier = this.identifier;
+    const { data } = this;
+    this.index = {};
+    const { index, identifier } = this;
 
-    for (var i = 0, l = data.length; i < l; i += 1) {
+    for (let i = 0, l = data.length; i < l; i += 1) {
       index[data[i][identifier]] = i;
     }
   };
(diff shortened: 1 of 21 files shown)


client/src/js/services/Store.js
```diff
@@ -76,19 +74,19 @@ function StoreService() {
    * @param {Object} object - an object to be inserted into the store.
    */
   Store.prototype.post = function post(object) {
-    var data = this.data;
-    var index = this.index;
-    var identifier = this.identifier;
+    const { data } = this;
```
jmcameron: Can't these be combined?

jmcameron (commented):
I skimmed the code.  Hopefully @mbayopanda will take a closer look at the code than I did.

I tested it with bhima_test.  It works and is a useful addition.

I ran into a couple of minor non-blocking issues:
- The edited form will not allow submission until notes are added.   Why is the description required?    If it is required, we should update the Bhima_test SQL files to add a non-null note to the unfulfilled purchase requisition (which is dated in 2016).   If this gets modified, you may as well add a relative date on it.

- The purchase order report has a minor/trivial typo.  The 's' of Note(S) is a capital letter.  It should be lower case:
image

jniles:
@mbayopanda @jmcameron this is ready for a second round of reviews and testing.

Here are basic changes:

1. You can't un-cancel a purchase order anymore.
2. I've made it so you can use the same update route to cancel purchase orders.  Controversally, you can cancel a confirmed purchase order... but I don't see a good way around that for the moment.

I made the changes @jmcameron suggested.  Thanks!

test/data.sql
```diff
@@ -590,17 +590,18 @@ INSERT INTO `price_list_item` VALUES
 UPDATE debtor_group SET price_list_uuid = HUID('75e09694-dd5c-11e5-a8a2-6c29955775b0') WHERE uuid = HUID('4de0fe47-177f-4d30-b95f-cff8166400b4');
 
 SET @purchase_order = HUID('e07ceadc-82cf-4ae2-958a-6f6a78c87588');
-INSERT INTO `purchase` VALUES
-  (@purchase_order, 1, 1, 300, 2, HUID('3ac4e83c-65f2-45a1-8357-8b025003d793'), DATE_ADD(CURRENT_DATE, INTERVAL -1725 DAY), CURRENT_TIMESTAMP, 1, NULL, NULL, 1);
+INSERT INTO `purchase`
+  (uuid, project_id, reference, cost, currency_id, supplier_uuid, date, user_id, payment_method, status_id) VALUES
```
jmcameron, line 594: I suggest adding a note here (like the one below).

jmcameron (approved):
Looks good to me.  I had one question about test/data.sql but there may be a reason for leaving out the 'note' field on some of the purchases.  In any case we can deal with it later.

I ran all tests on my local PC and all but E2E ran fine.  (The issue with my failing E2E test is not related to this issue; it fails on master too).

jmcameron:
@mbayopanda I think if you take a good look at the code on this and it seems okay to you, we can release it.  Thanks!

mbayopanda (commented):
When we create a purchase order it's already confirmed, so we cannot test the edit feature

client/src/modules/purchases/modals/status.tmpl.html
```diff
@@ -5,32 +5,32 @@
   </ol>
 </div>
 
-<form 
+<form
   name="ModalForm"
-  ng-submit="$ctrl.submit($ctrl.status)"
+  bh-submit="$ctrl.submit($ctrl.status)"
```
mbayopanda, line 10: 👍🏽

server/controllers/finance/purchases.js
```diff
@@ -299,23 +301,78 @@ function detail(req, res, next) {
...
+    // FIXME(@jniles) - we need a better process for this.  Probably a whole new modal to have
+    // users confirm only validated purchase orders.
+    const isCancelRequest = Object.keys(req.body).length === 1
+      && req.body.status_id === PURCHASES_STATUS_CANCELLED
+      && statusCanCancel.includes(purchase.status_id);
+
+    // can't edit non-awaiting-confirmation records
+    if (!isCancelRequest && purchase.status_id !== PURCHASE_STATUS_WAITING_CONFIRMATION) {
+      throw new BadRequest('Can only modify purchase orders that are awaiting confirmation.');
+    }
```
mbayopanda, line 344: 👍🏽 these checks on the server-side are excellent

server/controllers/finance/purchases.js
```diff
@@ -299,23 +301,78 @@ function detail(req, res, next) {
...
+      delete item.uuid;
+      db.convert(item, ['inventory_uuid']);
+      txn.addQuery(itemSQL, [item, uid]);
+    });
+
+    // run the transaction
+    await txn.execute();
+
+    // we only need the second return value
+    const [, record] = await Promise.all([
```
mbayopanda, line 367: `[,` so sexy 😆

mbayopanda (commented):
The edit purchase feature works fine 👍🏽 the only issue I got is when we create a new purchase order it is already confirmed

jniles:
@mbayopanda, that is a stock settings option that you can toggle.  For the Vanga database, they are automatically confirmed.  For IMCK, they shouldn't be.

mbayopanda (approved):
LGTM 👍🏽

mbayopanda:
bors delegate+

jniles:
Thanks @mbayopanda and @jmcameron !

PR #5527, opened 2021-03-19
Add tooltips to status cells in articles in stock

### LINKED ISSUE #5478: Articles in Stock status tags

jmcameron:
Added tooltips for the status cells in the Articles in Stock page.   
(Tried to position to the right of the cell, but that did not work. So ended up on the left and updated the tooltips for the warning cells to be on the left too for consistency).

#closes https://github.com/IMA-WorldHealth/bhima/issues/5478

**Testing**
- You may want to use a production database to see all the possible status tags
- To see the expired stock, enable that in the Search > Default filters.

### DIFF
--- a/client/src/modules/stock/inventories/templates/status.cell.html
+++ b/client/src/modules/stock/inventories/templates/status.cell.html
@@ -1,21 +1,41 @@
 <div class="ui-grid-cell-contents">
-    <div ng-if="row.entity.hasStockOut" class="label out-of-stock-status" translate>
+    <div ng-if="row.entity.hasStockOut" class="label out-of-stock-status"
+      uib-tooltip="{{'STOCK.STATUS.STOCK_OUT' | translate}}"
+      tooltip-append-to-body="true"
+      tooltip-placement="left"
+      translate>
       STOCK.STATUS.STOCK_OUT
     </div>
 
-    <div ng-if="row.entity.hasSecurityWarning" class="label critical-stock-status" translate>
+    <div ng-if="row.entity.hasSecurityWarning" class="label critical-stock-status"
+      uib-tooltip="{{'STOCK.STATUS.SECURITY' | translate}}"
+      tooltip-append-to-body="true"
+      tooltip-placement="left"
+      translate>
       STOCK.STATUS.SECURITY
     </div>
 
-    <div ng-if="row.entity.hasMinimumWarning" class="label low-stock-status" translate>
+    <div ng-if="row.entity.hasMinimumWarning" class="label low-stock-status"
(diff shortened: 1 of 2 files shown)


client/src/modules/stock/inventories/templates/status.cell.html
```diff
@@ -1,21 +1,36 @@
 <div class="ui-grid-cell-contents">
-    <div ng-if="row.entity.hasStockOut" class="label out-of-stock-status" translate>
+    <div ng-if="row.entity.hasStockOut" class="label out-of-stock-status"
+      uib-tooltip="{{'STOCK.STATUS.STOCK_OUT' | translate}}"
+      tooltip-placement="left"
+      translate>
```
mbayopanda, line 6: 👍🏽

mbayopanda (commented):
The code looks good but I have got some issue with the production database when you scroll down till the bottom, you can see there is no tooltip while on the flag we can see the tooltip 

WZlTRcaHZ7

mbayopanda:
Also, there is a behavior with the scrollbar when you put the cursor on the status while you are at the bottom of the registry

30PznNOgXe

jmcameron:
Thanks!  I'll take a look.  But I doubt I can do anything about it.

If the behavior is simply that the tooltips don't show when you scroll to the bottom, I think we can live with that since it is not any different from the way it used to work.

jmcameron:
> Also, there is a behavior with the scrollbar when you put the cursor on the status while you are at the bottom of the registry

I've fixed the problem. Apparently we also needed add this tooltip attribute:
   tooltip-append-to-body="true"

client/src/modules/stock/inventories/templates/status.cell.html
```diff
@@ -1,21 +1,41 @@
...
+      tooltip-append-to-body="true"
+      tooltip-placement="left"
+      translate>
       STOCK.STATUS.STOCK_OUT
     </div>
 
-    <div ng-if="row.entity.hasSecurityWarning" class="label critical-stock-status" translate>
+    <div ng-if="row.entity.hasSecurityWarning" class="label critical-stock-status"
+      uib-tooltip="{{'STOCK.STATUS.SECURITY' | translate}}"
+      tooltip-append-to-body="true"
```
mbayopanda, line 12: 👍🏽

mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #5534, opened 2021-03-23
chore(stock): remove integration table

### LINKED ISSUE #5533: Remove `integration` table
After some discussion with @mbayopanda, I propose we remove the stock `integration` table.  Its only purpose was to generate a stock identifier that we do not use (`PROJECT.INT.123`).  We've since gotten used to using the stock movement `SM.XX.YY` type of identifiers.

Further, it expects that lots created by integration should reference a row in the `integration` table as their origin_uuid, but this doesn't take into account integrated lots that exist already in the application.  Instead of coming up with a work around, let's simply remove that table.

jniles:
This commit removes the integration table from the database and updates the integration receipt accordingly.

Closes #5533.

### DIFF
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -190,26 +190,14 @@ function updateQuantityInStockAfterMovement(inventoryUuids, mvmtDate, depotUuid)
  * @method insertNewStock
  * @param {object} session The session object
  * @param {object} params Request body params (req.body)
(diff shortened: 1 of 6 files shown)


jmcameron (approved):
The code changes look good.
I tested it manually and it works correctly.
I looked over the document (directly from the stock entry via integration) as well as via the stock movements registry) and it looks good.
I also ran 'yarn test:integration' on my PC and it worked.

PR #5539, opened 2021-03-25
Prevent negative quantity

### LINKED ISSUE #5205: Resolve negative stock quantites problems
The problem is that we are encountering negative quantities in some production sites.  We have implemented a work-around for now that lets us operate with the negative values but it would be good to try

- Resolve exactly how they happen
- How negative stock quantities can be avoided in the future
- How current negative stock quantities can be repaired.

Note that negative stock quantities have the potential of causing errors in the computation of CMM, etc.

---

**Extracted from comments in PR https://github.com/IMA-WorldHealth/bhima/pull/5151**

**Jonathan Cameron:**
Before allowing negative quantities, I'd like to know why this is necessary. Specifically:
   - What does a negative quantity mean?
   - How does a negative quantity occur in the BHIMA software?
It seems to me that this could introduce more problems than it solves. It may be better to address why these negatives occur and fix it so they cannot happen!

**Jonathan Niles**
@jmcameron we were surprised by this as well. It's actually quite easy to get negative values and nearly unavoidable. Here is how:

   - Start with a balance of 100 pills at the beginning of this month.
   - Distribution 80 pills today.
   - Back date a distribution of 50 pills to be made yesterday.
   - We are now -30 in the hole today, even through you did have 80 pills today, and you did have 50 pills yesterday.

To avoid this, we would need to some check that says "would making this distribution put the database in an invalid/undefined state"? And then we would further need to figure out how to communicate that to the user such that they can avoid the issue.

Does that make sense? Maybe you can help us think of a way to work around this issue.... currently, it's escaping me.

**Jonathan Cameron**
I have two ideas about how to deal with this problem:

   - If an update will produce negative values, do not accept it; print out a very specific and detailed error message, something to the effect of: This update will produce a illegal negative quantity for stock item XYZ. Update the quantity in the form before submitting the form.
   - Automatically update negative values to zero when the form is submitted. Perhaps print an info/warning message that it got zeroed out.

We could also include a suggestion that it may be necessary to go back and correct previous updates. It seems obvious that there is an error in a previous update or this update.

It also seems like we need a utility to scan the database to see if there are any negative values and deal with them.

If it is not obvious, I still do like the idea of allowing negative quantities!!

**Jonathan Niles**
So, I worry about the calculation that backs this "if .. will produce negative values". This seems like a whole stock_movements table scan, recomputing balances all along. Unless we have a more efficient way to store balances, this could get really computationally intensive. I suppose we could cache the balances in a table that contains (lot_uuid, depot_uuid, current_balance) and decrement that on each exit, and if current_balance goes negative we know something is wrong?

**Jonathan Cameron**
Here are couple more thoughts on this:

   - In the stock exit form, prevent selecting a date before the current date. I realize this may not be practical, given delays in updating data, etc. But if we did this, fixes to historical quantities could be done via the Stock Adjustment page.
   - The Stock Adjustment page should accept the total at a certain date. I'm not sure if the Increase/Decrease options really make sense in typical operations (it seems like they have use the total to compute the increase/decrease anyway). If the Stock Adjustment form was used to correct historical data, it would be for one stock item at a time and it would not really matter if the computations are involved. Should anyone expect Stock Adjustment (an unusual operation to my mind) to be fast? Stock Exits on the other hand should be speedy since they will be done a lot.
   - Back to the concern about computations involved to see if a changes produces any negative values: We are only doing one stock item at a time in the stock exit. And the involved computations only need to be done if the date is back-dated -- and I think we can just put up a "Checking totals..." notification while the computations are being done.. If it is the current date, we can just make sure that we do not exit more than the existing quantity available (which I assume we already do).

**Jonathan Niles**
The problem with using Stock Adjustments is that it breaks the calculation for CMM since adjustments are not considered consumption. It has been kind of nice to have all CMM calculations flow from a single source (stock exists), so we can optimize that to death.

For (3), you are right - we only need to do the computation for back dating totals, and I think that is entirely reasonable. It seems like the best proposal yet. That way, we don't delay the distribution of stock to "regular" patients, in real time.

I think our CMM calculation also needs to be reviewed to make sure this logic is put in place. I believe we make an attempt, but am not sure the success of it.

lomamech:
Prevent negative quantity in client side in
- All stock exit
- Inventory adjustement
- Aggregate stock consumption

test(Prevent Negative quantity)
- Add integration test for prevent negative quantity during distribution
  lots to patient, aggregate consumption and adjustement inventories
closes #5205


lomamech:
@mbayopanda and @jmcameron Can you review this Pull Request

client/src/i18n/en/errors.json
```diff
@@ -25,6 +25,7 @@
     "ER_BAD_FIELD_ERROR" : "Column does not exist in database.",
     "ER_ROW_IS_REFERENCED" : "Cannot delete entity because entity is used in another table.",
     "ER_ROW_IS_REFERENCED_2" : "Cannot delete entity because entity is used in another table.",
+    "ER_PREVENT_NEGATIVE_QUANTITY_IN_STOCK" : "There are stock movements, which may overconsume the quantity in stock and generate negative quantity in stock",
```
jmcameron: I see that this error message will be used in multiple contexts.  Ideally there would be a separate message for each place this issue can come up.  Eg, one for stock entry that starts with:   "The proposed stock entry may overconsume..."
and for stock adjustment:  "The proposed stock adjustment may overconsume..."
Also, if you feel it is appropriate, I suggest replacing "may" with "will".

Whether you decided to add more error messages or not, please remove the comma.
"There are stock movements which may..."

client/src/modules/stock/aggregated_consumption/aggregated_consumption.js
```diff
@@ -228,6 +233,21 @@ function StockAggregatedConsumptionController(
...
   }
 
+  function loadCurrentInventories(depot, dateTo = new Date()) {
+    vm.loading = true;
+    Stock.lots.read(null, { depot_uuid : depot.uuid, dateTo })
+      .then(lots => {
+        vm.currentInventories = lots.filter(item => item.quantity > 0);
+
+        // Here we check directly if a Depot has inventories in stock available in current date
+        vm.emptyCurrentStock = !vm.currentInventories.length;
```
jmcameron: I think something like this would more readable:

vm.emptyCurrentStock = vm.currentInventories.length <= 0;

client/src/modules/stock/aggregated_consumption/aggregated_consumption.js
```diff
@@ -249,6 +269,37 @@ function StockAggregatedConsumptionController(
...
+      vm.overconsumption.forEach(item => {
+        item.textI18n = {
+          text : item.text,
+          label : item.label,
+          quantityAvailable : item.quantityAvailable,
+          quantity : (item.quantity_consumed + item.quantity_lost),
+        };
+      });
+
+      Notify.danger('ERRORS.ER_PREVENT_NEGATIVE_QUANTITY_IN_STOCK');
```
jmcameron: This error message gives no clue regarding which stock item is causing the problem.  I think it would be helpful to give the stock item name/description in the error message.   Unless the next error message has those details?

client/src/modules/stock/exit/exit.js
```diff
@@ -29,21 +29,29 @@ function StockExitController(
...
+
+    vm.overconsumption = [];
     checkValidity();
   };
 
   vm.onChangeDepot = depot => {
     vm.depot = depot;
     loadInventories(vm.depot);
+    loadCurrentInventories(vm.depot);
+
```
jmcameron, line 53: How does this handle the case where the current inventories may already have negative values?

client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
```diff
@@ -27,14 +27,23 @@ function StockInventoryAdjustmentController(
...
+    loadCurrentInventories(vm.depot);
+
+    vm.overconsumption = [];
   };
 
   vm.onChangeDepot = depot => {
     vm.depot = depot;
     loadInventories(vm.depot);
+    loadCurrentInventories(vm.depot);
+
```
jmcameron, line 45: Same question as before: How does this handle cases where the already existing data has negatives?

server/controllers/stock/index.js
```diff
@@ -284,14 +297,39 @@ async function createInventoryAdjustment(req, res, next) {
...
+          });
+        }
+      }
+    });
+
+    filteredInvalidData = await lots.filter(l => l.outputQuantity > l.quantityAvailable);
+
+    if (filteredInvalidData.length) {
+      throw new BadRequest(
+        `There are stock movements,
```
jmcameron: Are these duplicate error messages?

jmcameron (commented):
It mostly looks good.  I was able to get the error message by doing a bogus stock inventory adjustment and I see the error message:
image

But when I scroll up, I see an unstranslated error message:
image

Also, please note that the Stock Inventory Adjustment page is too long.  I suggest making it shorter by default and require the user to scroll down only if many items need adjustment.  In production sites, you'll have to scroll anyway, so why make it long?

Also, what is the strategy for dealing with sites that already have negative quantities?

I was able to run the tests and they all ran fine except for the End-to-End tests (which is also failing for me on master).

jmcameron:
Would you please add a short description of how to run tests manually to reproduce this error in all the places it might occur (eg, stock adjustment, stock exit, etc).  Thanks!

client/src/modules/stock/aggregated_consumption/aggregated_consumption.js
lomamech: I remove **vm.emptyCurrentStock** this element, I did not use it anywhere

client/src/i18n/en/errors.json
lomamech: ```` Also, what is the strategy for dealing with sites that already have negative quantities? ````
In this case alone the positive adjustment is required, I just tested it with the database of IMCK and it works well

lomamech: ```` Also, what is the strategy for dealing with sites that already have negative quantities? ````
In this case alone the positive adjustment is required, I just tested it with the database of IMCK and it works well

server/controllers/stock/index.js
```diff
@@ -382,7 +422,43 @@ async function createMovement(req, res, next) {
...
+      if (stockAvailable) {
+        stockAvailable.forEach(stock => {
+          if (stock.uuid === lot.uuid) {
+            lot.quantityAvailable = stock.quantity;
+          }
+        });
+      }
+    });
+
+    filteredInvalidData = await params.lots.filter(l => l.quantity > l.quantityAvailable);
```
mbayopanda, line 446: These checks in the server side is excellent 👍🏽

server/controllers/stock/index.js
```diff
@@ -382,7 +422,43 @@ async function createMovement(req, res, next) {
...
+  }
+
   try {
+    if (filteredInvalidData.length) {
+      throw new BadRequest(
+        `There are stock movements,
+        which may overconsume the quantity in stock and generate negative quantity in stock`,
+        `ERRORS.ER_PREVENT_NEGATIVE_QUANTITY_IN_STOCK`,
+      );
+    }
```
mbayopanda, line 455: 👍🏽

mbayopanda (approved):
Handling over consumption at the server-side is a good idea 👍🏽 , so I am waiting for all @jmcameron questions to be addressed to merge this.

client/src/modules/stock/aggregated_consumption/aggregated_consumption.js
lomamech: Indeed there is a message that describes the products that are problematic
image

client/src/modules/stock/exit/exit.js
lomamech: For products with negative quantities, the system will assimilate to products out of stock, which is why only those with positive quantities are filtered.
````
 function loadCurrentInventories(depot, dateTo = new Date()) {
    vm.loading = true;
    Stock.lots.read(null, { depot_uuid : depot.uuid, dateTo })
      .then(lots => {
        vm.currentInventories = lots.filter(item => item.quantity > 0);
      })
      .catch(Notify.handleError)
      .finally(() => {
        vm.loading = false;
      });
  }
````

File: client/src/modules/stock/inventory-adjustment/inventory-adjustment.js (hunk 5 above)
Comment (lomamech) [reply client/src/modules/stock/inventory-adjustment/inventory-adjustment.js:45]: Products with negative quantities will be considered as out of stock products.
image

File: server/controllers/stock/index.js (hunk 6 above)
Comment (lomamech) [reply server/controllers/stock/index.js]: Yes, and it is so in several uses of  ````throw new BadRequest````
image

**Comment by lomamech** [comment]
> Would you please add a short description of how to run tests manually to reproduce this error in all the places it might occur (eg, stock adjustment, stock exit, etc). Thanks!

To be able to test manually, you must
````- Use the 'Tertiary Depot' deposit
- Carry out all the INTEGRATION test, during the integration test the Tertiary deposit will be out of stock
- Look at the stock movement registers and see the date corresponding to the entry of stock by integration into this depot,
- And tries to take out of stock, adjustment or consumption of aggregated stock by backdating the date for that of the entry of stock````
Another means this 
- to consume a product in any warehouse and backdated to consume for this product a quantity greater than that which is in stock`

**Comment by lomamech** [comment]
@jmcameron  I'm ready for another review 😊

**Comment by jmcameron** [comment]
> @jmcameron I'm ready for another review 😊

I'll try to do it this afternoon!

**Comment by lomamech** [comment]
> > @jmcameron I'm ready for another review 😊
> 
> I'll try to do it this afternoon!

👍

File: client/src/modules/stock/exit/exit.js (hunk 4 above)
Comment (jmcameron) [reply client/src/modules/stock/exit/exit.js:53]: ok!

File: client/src/modules/stock/inventory-adjustment/inventory-adjustment.js (hunk 5 above)
Comment (jmcameron) [reply client/src/modules/stock/inventory-adjustment/inventory-adjustment.js:45]: Do we have a way to fix the lots with negative quantities?
Since lots with negative quantities will not be displayed, how do we find them and fix them?

File: client/src/i18n/en/errors.json (hunk 1 above)
Comment (jmcameron) [reply client/src/i18n/en/errors.json]: Once the more detailed error message is display, I agree that we don't need more detail in the temporary warning at the top of the page.

File: client/src/modules/stock/aggregated_consumption/aggregated_consumption.js (hunk 3 above)
Comment (jmcameron) [reply client/src/modules/stock/aggregated_consumption/aggregated_consumption.js]: Good to see this message with the details.

File: client/src/modules/stock/aggregated_consumption/aggregated_consumption.js (hunk 2 above)
Comment (jmcameron) [reply client/src/modules/stock/aggregated_consumption/aggregated_consumption.js]: Nice!

File: client/src/modules/stock/exit/exit.js (hunk 4 above)
Comment (jmcameron) [reply client/src/modules/stock/exit/exit.js:53]: ok. I understand!

**Review by jmcameron — COMMENTED** [review]
I tested it (based on your instructions) and it works fine.  Thanks for answering and addressing my previous comments.

I have two concerns:

1.  The warning message about problematic adjustment is hard to understand
     
     image

     I would suggestion wording something like this (in English):

     Changing the quantity of [stock name] in lot label PL1 from 1000 to 0 will result in a negative inventory quantity.

     My guess is that the French text could also use improvement.

     I was not completely sure I understand what you are trying to say here, so take my advice with grain of salt (

2. When I tried doing a backdated Stock Exit, I ran into a problem when there was no (or negative) stock available on the specified date:
       
    image

    When I clicked on the [Add] button, nothing happened.  However when I look at the console, I see this error message:
    
    image

    We should handle this more gracefully.  I'm not sure whether this is related to the PR or not.  So if it makes sense to you we could make a new issue for it and handle it later.  Let me know what you want to do about this.

Also, I think that the length of time that the temporary warning message at the top of the screen says up is too short.   It is okay for success messages to stay up a short time, but warnings should stay up longer so people have time to read them!   I suggest something more than 5 seconds.

Finally, I think you are very close here, so I suggest re-basing after you do your final adjustments.

File: client/src/modules/stock/inventory-adjustment/inventory-adjustment.js (hunk 5 above)
Comment (lomamech) [reply client/src/modules/stock/inventory-adjustment/inventory-adjustment.js:45]: image
The stock adjustment module allows you to do this, you just have to check the option **Include exhausted or negative lots**

**Comment by lomamech** [comment]
> 2\. We should handle this more gracefully.  I'm not sure whether this is related to the PR or not.  So if it makes sense to you we could make a new issue for it and handle it later.  Let me know what you want to do about this.
> 
> Also, I think that the length of time that the temporary warning message at the top of the screen says up is too short. It is okay for success messages to stay up a short time, but warnings should stay up longer so people have time to read them! I suggest something more than 5 seconds.
> 
> Finally, I think you are very close here, so I suggest re-basing after you do your final adjustments.

image
I think we will have to set **[Add] button** disable

**Comment by lomamech** [comment]
@jmcameron The only way we had to correct negative inventory quantities is with the inventory adjustment module.  I just added an issue to solve this problem #5549

File: client/src/modules/stock/inventory-adjustment/inventory-adjustment.js (hunk 5 above)
Comment (jmcameron) [reply client/src/modules/stock/inventory-adjustment/inventory-adjustment.js:45]: Thanks for pointing this out.  I had forgotten about it.

**Review by jmcameron — APPROVED** [review]
I think you have dealt with all the issues I raised.  I suggest re-basing and then releasing it.

File: client/src/modules/stock/exit/exit.js [hunk 9]
```diff
@@ -676,8 +734,11 @@ function StockExitController(
 
     return Stock.movements.create(movement)
       .then(document => {
-        ReceiptModal.stockExitLossReceipt(document.uuid, bhConstants.flux.TO_LOSS);
-        reinit(form);
+
+        if (document.uuid) {
```
jniles, line 738: Why wouldn't there be a document uuid?

client/src/modules/stock/inventory-adjustment/inventory-adjustment.js
```diff
@@ -218,6 +239,36 @@ function StockInventoryAdjustmentController(
...
+          text : item.text,
+          label : item.label,
+          old_quantity : item.old_quantity,
+          quantityAvailable : item.quantityAvailable,
+          quantity : item.quantity,
+        };
+      });
+
+      Notify.danger('ERRORS.ER_PREVENT_NEGATIVE_QUANTITY_IN_ADJUSTMENT_STOCK');
+      vm.$loading = false;
```
jniles: I think you mean `vm.loading` here.  I can't find another reference to `$loading` in this file.

server/controllers/stock/index.js
```diff
@@ -379,7 +418,42 @@ async function createMovement(req, res, next) {
...
+
+  const paramsStock = {
+    dateTo : new Date(),
+    depot_uuid : params.depot_uuid,
+    includeEmptyLot : 0,
+    month_average_consumption : params.month_average_consumption,
+    average_consumption_algo : params.average_consumption_algo,
+  };
+
+  stockAvailable = await core.getLotsDepot(null, paramsStock);
```
jniles: If you only use this check on exit, let's move this call into the `if (params.is_exit) {` black.  That way, we won't wait for this db request on stock entries.

jniles (changes requested):
I can't get this to work.  See below:

meii5DGiss

Isn't this supposed to block me?

client/src/modules/stock/exit/exit.js
lomamech: This could happen if two users connect to the Stock Exit interface and the first user consumes the entire stock of a product before the second validates its consumption, and as the control is also carried out at server level, an error message will be returned to the client without there being a document.uuid

jniles (approved):
LGTM.  Great work!

bors r+

PR #5540, opened 2021-03-26
Delete delay, initial_quantity, quantity, entry_date fields from the lot table

### LINKED ISSUE #5366: Remove 'quantity' and 'initial_quantity' from lot table
The '**quantity**' field in the 'lot' table should be removed.  The quantity in a lot should be tracked by stock movements (see queries in server/controllers/stock/core.js).   Note that several queries in server/controllers/stock/lots.js will need to refactored since they include the 'quantity' field in the 'lot' table rather than computing it on the fly.   It is also likely, that some integration tests will need to be fixed for changes.

We also need to remove the '**initial_quantity**' and **unit_price** fields since the same issues apply to them.  

Other lot table columns that need to be reconsidered 

- delay?
- entry_date (applies to parts of a lot received, not the entire manufacturer's lot)
- is_assign?

- **description** (since it does not seem to be used anywhere.  Although it is possible that there more be information for this in the future if we connect into online manufacturer lot databases (eg, which plan produced it, etc).

**ALSO:** We may need to remove the 'quantity' from the lots registry page -- or make sure it is computed correctly.

**NOTE:** The stock entry page should still have entries for initial_quantity and entry_date.  These should be stored somewhere, but not in the 'lot' table.

**QUESTION:** How is initial_quantity related to the "Global Quantity" in the Stock Entry lot modal page? What is "Global Quantity" and where does it come from?  Is this the total number of items in the manufacturer's lot?  If so, how could possibly know that?  Perhaps it should be purged from the Stock Entry pages?

jmcameron:
This PR removes fields from the lot table.
- The **delay** field was not used and was purged.
- The **initial_quantity** removed from the lot table.  It was ambiguous and not necessary and did not make sense to be part of the lot table.
- The **quantity** field was removed from the table.  Various queries were modified to compute the current quantity for the lot using the stock_movement table.
- and **entry_date** field was removed from the lot table.   Various queries were modified to get the date of the earliest stock movement for this lot as the lot's entry date.  Although we considered purging it entirely, it did seem useful when deciding whether to merge lots.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5366

**Testing**
- Suggest using a production database.
- Try a few stock entries and exists
- Try editing a lot (from the lot registry).  Notice that there all these fields are no longer present.
- Look at the Lots registry, hover over the header cell of the "Entry Date" column and notice the expanded tooltip
- Try Stock > Find Duplicate Lots (you may need to enable permission for this page)
   - Notice that there is no 'entry_date' column.  Since each line represents several lots with (potentially) different entry_dates, it did not make sense to display it here.
   - On the action menu of one duplicate lots, use the "Merge duplicate lots" item.  Notice that the popup does have the entry date of each lot (actually the date of the first stock movement for this lot).  I decided to leave this in since it might help with decisions about merging lots.

NOTE: This PR did not remove 'unit_cost' from the lot table.  This should probably be done but requires a little more discussion.  I'll create a new issue for it.


server/controllers/stock/reports/purchase_order_analysis/index.js
```diff
@@ -85,20 +85,27 @@ async function report(req, res, next) {
...
+    //       See getLots() query in server/controllers/stock/core.js for an example query.
+    //       jmcameron 2021-03-26.
 
     // This request tracks the entries in stock of a purchase order, by date, by deposit but also by product
     const sqlInventoriesInStockDetailled = `
       SELECT BUID(p.uuid) AS purchase_uuid, BUID(iv.uuid) AS inventory_uuid, iv.text AS inventoryText,
         l.label AS lotLabel, p.project_id, p.reference, p.user_id, l.uuid AS lotUuid, l.quantity,
-        l.expiration_date, l.entry_date, iv.code,
+        l.expiration_date,
+        (SELECT MIN(sm.date) FROM stock_movement sm WHERE sm.lot_uuid = l.uuid) AS entry_date,
```
mbayopanda, line 107: It does the job 👍🏽

mbayopanda (approved):
@jmcameron this work is excellent, and it looks good to me 👍🏽 

bors r+

PR #5542, opened 2021-03-27
Fix purchase registry template path

mbayopanda:
This PR is a about the path template of the purchase registry which was not correct.

TGUfGIKq2V


jmcameron (approved):
Looks good.
It is not clear to me how this got broken, but this PR fixes it.

PR #5545, opened 2021-03-29
Remove lots flags structure

### LINKED ISSUE #5460: Duplicate lot flags in expiration report?
In the process of fixing the Stock Expiration Report in PR https://github.com/IMA-WorldHealth/bhima/pull/5438, I noticed that the expiration, at_risk, etc, flags in the lot appear both at the top level of each lot object as well as in its 'flags' sub-object.   

To see this,  edit server/controllers/stock/reports/stock/expiration_report.js and insert console.log(lot) inside the rows.forEach loop (about line 70) and observe the results when you execute the Stock Expiration Report.

This duplication should be cleaned up.

jmcameron:
Removed the 'flags' sub-object from lots.  This was an unnecessary level of indirection for the lot flags (eg, expiration, etc) and it was not consistently being used.  Removing it makes the logic cleaner.

**Testing**
This should be completely transparent to the user.  The main pages that might be affected are the lots registry page and the Stock expiration report.   If those work normally for you, then it is probably okay.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5460

### DIFF
--- a/client/src/modules/stock/lots/templates/lifetime.cell.html
+++ b/client/src/modules/stock/lots/templates/lifetime.cell.html
@@ -1,5 +1,5 @@
 <div class="ui-grid-cell-contents text-right">
-  <span ng-if="row.entity.hasExpirationDate && !row.entity.flags.expired">
+  <span ng-if="row.entity.hasExpirationDate && !row.entity.expired">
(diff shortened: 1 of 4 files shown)


lomamech (approved):
No comment, i approve

PR #5550, opened 2021-03-30
improvement(Set Add Button Disable)

### LINKED ISSUE #5549: Make the add rows button disable when there are any stock in the depot
It has been observed that the Add button remains in use even when there is no stock, This issue proposes to solve this problem
image

lomamech:
- Set Add button disable when there are any stock

closes #5549


jmcameron (approved):
I tested this and it works nicely.
Thanks for fixing it!

PR #5551, opened 2021-03-31
feature(Implement link voucher to stock movement)

### LINKED ISSUE #5504: Link stock movement receipt from Vouchers if record is a stock movement
If the record is a stock movement, we should provide a link to the stock movement from the voucher page.  It would go in the menu:

image

lomamech:
- Implement link stock movement receipt from Vouchers for all Stock
  movement

closes #5504

### DIFF
--- a/client/src/modules/vouchers/templates/action.cell.html
+++ b/client/src/modules/vouchers/templates/action.cell.html
@@ -39,7 +39,7 @@
         <span class="fa fa-file-text-o"></span> <span translate>REPORT.VIEW_VOUCHER</span>
       </a>
     </li>
-    <li ng-show="row.entity.stock_mov_document">
+    <li ng-show="row.entity.type_id === 12 || row.entity.type_id === 13 || row.entity.type_id === 14">
       <a data-method="view-stock-movements" ui-sref="stockMovements({ filters : [{ key : 'voucherReference', value : row.entity.reference, displayValue: row.entity.reference, cacheable:false }, { key : 'period', value : 'allTime', cacheable:false }]})" href>
         <span class="fa fa-list-alt"></span> <span translate>STOCK.STOCK_MOVEMENT</span>
       </a>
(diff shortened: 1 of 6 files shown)


lomamech:
@jniles  It's done  **Link stock movement receipt from Vouchers if record is a stock movement**  #5504

client/src/modules/vouchers/templates/action.cell.html
```diff
@@ -39,6 +39,11 @@
         <span class="fa fa-file-text-o"></span> <span translate>REPORT.VIEW_VOUCHER</span>
       </a>
     </li>
+    <li ng-show="row.entity.type_id === 12 || row.entity.type_id === 13 || row.entity.type_id === 14">
```
jniles: In the future, it would be cleaner (and probably faster) to just stick an `isStockMovement` to these rows and move this logic into the JS controller.  I'm thinking something like:
```js
const stockMovementTypes = [
  bhConstant.transactionTypes.STOCK_EXIT,
  bhConstant.transactionTypes.STOCK_ENTRY,
  // .. etc
];

rows.forEach(row => {
  row.isStockMovement = stockMovementTypes.includes(row.type_id);
});
```

This isn't a critical change though.

client/src/modules/vouchers/templates/action.cell.html
```diff
@@ -39,6 +39,11 @@
...
+      <a data-method="view-stock-movements" ui-sref="stockMovements({ filters : [{ key : 'voucherReference', value : row.entity.reference, displayValue: row.entity.reference, cacheable:false }, { key : 'period', value : 'allTime', cacheable:false }]})" href>
```
jniles: Could you move this and the `data-method="view-voucher"` up in the list?  It seems to be below the reverse record option and I can see users frequently mis-clicking.

image

jniles (commented):
This functionality works great!  Good job @lomamech 

The only blocking comment is to move the `<li>` higher in the list.  We don't want users accidentally reversing records when they are supposed to just be looking at the record.

jniles (approved):
Looks good to me!

bors r+

PR #5557, opened 2021-04-02
Fix: end to end test bugs

jniles:
Fixes a number of test bugs introduced by the transition to Jenkins and the changes we've made to Stock Management.


jniles:
This doesn't fix all the test bugs in the end to end tests, but it does fix a number of them.

Notably, we now create the appropriate fiscal year, we also removed writes to the `integration` table in importing stock, and we allow multiple flags to be passed into the chromedriver to support running on Ubuntu 20.04.

We still fail in the stock lots tests though, because of the changes to handling unique lots.  I'm at the point where I think we should just rewrite the stock tests with a coherent dataset, like we did for stock integration testing.

@lomamech can you try to run these and see how they work on your machine?

lomamech:
> This doesn't fix all the test bugs in the end to end tests, but it does fix a number of them.
> 
> Notably, we now create the appropriate fiscal year, we also removed writes to the `integration` table in importing stock, and we allow multiple flags to be passed into the chromedriver to support running on Ubuntu 20.04.
> 
> We still fail in the stock lots tests though, because of the changes to handling unique lots. I'm at the point where I think we should just rewrite the stock tests with a coherent dataset, like we did for stock integration testing.
> 
> @lomamech can you try to run these and see how they work on your machine?

I just did it and twice in a row the E2E fails with the test  
    **_Stock Lots Registry
         find lots by entry date:_**
NoSuchElementError: No element found using locator: By(css selector, [date-id="entry-date"])

image

jniles:
Thanks @lomamech!  That reflects what I found as well.

lomamech (commented):
Otherwise it would take a great investigation

jniles:
@lomamech I've confirmed that this works on my local machine and on @jmcameron's machine.  If you can confirm it works for you, we can get this merged.

jniles:
Looks like this finally works on everyone's machines.  We still have some flakey issues to work out, but this should solve a lot of issues.

bors r+

PR #5558, opened 2021-04-02
CDR Peremption Reporting Tool

mbayopanda:
This PR adds a module for reporting expired rate from CDR 

image

To test it, please import these files : 
- MouvementStock
- Article
- ArticleLot
- LigneLotDansDoc

[Article.xlsx](https://github.com/IMA-WorldHealth/bhima/files/6249077/Article.xlsx)
[ArticleLot.xlsx](https://github.com/IMA-WorldHealth/bhima/files/6249078/ArticleLot.xlsx)
[LigneLotDansDoc.xlsx](https://github.com/IMA-WorldHealth/bhima/files/6249079/LigneLotDansDoc.xlsx)
[MouvementStock.xlsx](https://github.com/IMA-WorldHealth/bhima/files/6249080/MouvementStock.xlsx)


client/src/modules/cdr_reporting/depots/templates/label.tmpl.html
```diff
@@ -0,0 +1,5 @@
+<div class="ui-grid-cell-contents">
```
jniles, line 1: What's the purpose of this template?

server/controllers/cdrReporting/peremption.handlebars
```diff
@@ -0,0 +1,212 @@
...
+          {{!-- <td class="first">{{date periodKey}}</td> --}}
+          <td class="first">Stocks</td>
+          <td>{{currency (look ../vars.totalStockAtPeriod periodKey)}}</td>
+          {{#each ../depots as |depot|}}
+            <td>{{currency (look ../../vars.stockAtPeriod periodKey depot.text)}}</td>
+          {{/each}}
+        </tr>
+        <tr class="highlighted delimitter">
+          {{!-- <td class="first">{{date periodKey}}</td> --}}
+          <td class="first">Tx de péremption</td>
```
jniles, line 112: Would it be possible to translate these values?

server/lib/uploader.js
```diff
@@ -104,6 +105,37 @@ function Uploader(prefix, fields) {
   return multer({ storage }).array(fields);
 }
 
+function multipleFields(prefix, fields) {
+  // format the upload directory.  Add a trailing slash for consistency
+  const hasTrailingSlash = (prefix[prefix.length - 1] === '/');
+  const directory = path.join(fsdir, hasTrailingSlash ? prefix : `${prefix}/`);
+
+  // configure the storage space using multer's diskStorage.  This will allow
+  const storage = multer.diskStorage({
```
jniles, line 114: Correct me if I am wrong, but this looks nearly identical to the uploader definition above.

Could you move the common items out into their own function/object?

I'm imagining:
```js
function setupDiskStorage(prefix, fields) {
  // format the upload directory.  Add a trailing slash for consistency
  const hasTrailingSlash = (prefix[prefix.length - 1] === '/');
  const directory = path.join(dir, hasTrailingSlash ? prefix : `${prefix}/`);
  // configure the storage space using multer's diskStorage.  This will allow
  const storage = multer.diskStorage({
    destination : async (req, file, cb) => {
      try {
        // NOTE: need absolute path here for mkdirp
        const fullFolderPath = path.join(fsdir, directory);
        debug(`creating upload directory ${fullFolderPath}.`);
        await mkdirp(fullFolderPath);
        cb(null, fullFolderPath);
      } catch (err) {
        cb(err);
      }
    },
    filename : (req, file, cb) => {
      const id = uuid();
      // ensure that a link is passed to the req.file object
      file.link = `${directory}${id}`;
      debug(`Storing file in ${file.link}.`);
      cb(null, id);
    },
  });
}

// then...

function Uploader(prefix, fields) {
  const storage  = setupDiskStorage(prefix, fields);
  return multer({ storage }).array(fields);
}

function multipleFields(prefix, fields) {
  const storage  = setupDiskStorage(prefix, fields);
  return multer({ storage }).fields(fields);
}
```

jniles (commented):
@mbayopanda I've made a few comments.  I wonder how we should deploy this.  In the short run, we can definitely put it on the online deployment, but it doesn't really make sense for hospital deployments.  Do you have any ideas?

jniles (commented):
You also have the migration script in here, could you remove it?

PR #5566, opened 2021-04-08
Added Euros and fixed exchange rate selection page

jmcameron:
Added support for Euros and added related currency and exchange rate tests.

**WARNINGS**
- This PR includes migration code.  When migrating production sites, you should probably update the default exchange rate for Euros to be a more current value.

**TESTING**
- Use bhima_test
- Go to page:  Administration / Exchange Rate
- Notice the new entry for Euros
- Click on the [Update Exchange Rate(s)] button to get the modal for changing the exchange rate
- Select Euros from the drop-down currency selector
- Notice that the current exchange rate is updated for Euros (just below the date field)
- Try changing the value and submitting
- Notice the new value is shown on the main page
- Go back into the modal, select Euros again and verify that they new value is shown

**NOTES**
- The process of adding a new currency requires multiple steps and should be documented in the Wiki.


client/src/modules/exchange/exchange.modal.html
```diff
@@ -17,13 +17,14 @@
       <ui-select
         name="currency"
         ng-model="ModalCtrl.rate.currency"
+        ng-change="ModalCtrl.selectCurrency()"
         required>
         <ui-select-match placeholder="{{ 'FORM.PLACEHOLDERS.CURRENCY' | translate }}">
```
jniles, line 22: Can you check to make sure this translation exists?  It doesn't look like it is rendering on my machine:

image

jniles (changes requested):
This works _surprisingly_ well.  Well done!

The only thing that doesn't work that I could find is the handlebars rendering for euro.  You'll find that in the file `server/lib/template/helpers/finance.js`.  You need to add the configuration for Euro, inc,luding the currency symbol.  The FC and USD are already defined there.  It will make our receipts render euro when they are paid in euro instead of the default $.

Here is an example of the logic working, but not the currency rendering:
image

client/src/i18n/currency/eur.json
```diff
@@ -0,0 +1,29 @@
+{
+  "CURRENCY_SYM": "€",
+  "DECIMAL_SEP": ".",
```
jniles: It looks like this should be a comma, right?

jniles (approved):
Tested and works well.  Thanks @jmcameron !

I made one comment about unifying the format of the euro on the client and server (handlebars).  I think they both should use commas for decimals and a thin non-breaking whitespace for the thousands space.

client/src/modules/exchange/exchange.modal.html
jmcameron: Fixed

client/src/i18n/currency/eur.json
jmcameron: Fixed

PR #5573, opened 2021-04-14
Feature: move delete records to a role_action instead of an enterprise_setting

### LINKED ISSUE #5379: Make deleting records (invoices/cash payments/vouchers) into a Role Action

jniles:
This commit moves the delete records value to a role_action instead of an enterprise_setting.  This change makes it so that the deletion ability is partitioned by user instead of just being globally "on" or "off" for an entire enterprise.

Further, we've also separated out each type of deletion so that an enterprise can choose to allow the deletion of only a certain type of transaction (for example, purchase orders) without allowing any others.  Alternatively, they can give some users some permissions but not all users all permissions.

Here is what it looks like:

rod4ARuPsJ

How to tests this:
  1. Build the test database.
  2. Sign in as the superuser
  3. Go to the invoice registry
  4. Observe that you have access to delete records in the dropdown menu.
  5. Navigate to the posting journal.
  6. Observe that you have access to the "unpost" button at the top of the journal.
  7. Navigate to the roles module. 
  8. Edit the superuser **actions** and uncheck the "Can unpost transactions" option and the "Delete Invoice" options.
  9. Submit.
  9. Navigate to both modules and observe that you no longer have access.
  10. Navigate back and check the options again
  11. You will now have access.

Closes  #5379.


client/src/js/constants/bhConstants.js
```diff
@@ -13,6 +13,11 @@ function constantConfig() {
     actions : {
       CAN_EDIT_ROLES : 1,
       CAN_UNPOST_TRANSACTIONS : 2,
+      DELETE_CASH_PAYMENT : 3,
+      DELETE_INVOICE : 4,
+      DELETE_PURCHASE_ORDER : 5,
+      DELETE_STOCK_MOVEMENT : 6,
+      DELETE_VOUCHER : 7,
```
jmcameron, line 20: I wonder whether we should consider using non-sequential numbers such as:
      CAN_EDIT_ROLES : 1
      CAN_UNPOST_TRANSACTIONS : 101,
      DELETE_CASH_PAYMENT : 201,
      DELETE_INVOICE : 202,
      DELETE_PURCHASE_ORDER : 203,
      DELETE_STOCK_MOVEMENT : 204,
      DELETE_VOUCHER : 205,

So that new items can be added in the future without requiring numbering or having strange numbering sequences.   Since we are just using these as constants, the fact that they are non-sequential should not matter.

On second thought... I noticed later that these are actually the record numbers in an SQL table, so it is probably better to keep them sequential.

client/src/modules/invoices/registry/templates/action.cell.html
```diff
@@ -65,7 +65,7 @@
       </a>
     </li>
 
-    <li ng-hide="row.entity.reversed" ng-if="grid.appScope.allowsRecordDeletion()">
+    <li ng-hide="row.entity.reversed" bh-has-permission="grid.appScope.bhConstants.actions.DELETE_INVOICE">
```
jmcameron, line 68: Just a stylistic question:
- Might it be better to keep the permissions function allowsRecordDeletion() and hide the bh-has-permission, bhConstants.actions, etc hidden from the code.  That way if we decide to change that in the future, we only have to change the function.  I presume the function would be available in the Session object so it would be shared by all potential users?

jmcameron (approved):
The code changes look good.

I ran tests, as suggested, and it works as advertised.

I had one question (in general regarding permissions, and not blocking this PR): It looks like the permissions mechanism prevents the non-permitted link/button from being displayed.  Is it still possible for someone to fake the request (eg, with injected client Javascript) and perform the action anyway?  In other words, is there a mechanism on the server side preventing non-permitted actions even if they slip by the client side?  (I skimmed the server side code changes and did see anything...)

client/src/js/constants/bhConstants.js
jniles: This is a good thought, but at this point there aren't enough people smashing in permissions to make it worth coming up with a system.  This would make a lot more sense for our `unit` table, actually.  The probably is that we'd need to deal with legacy installations...

client/src/modules/invoices/registry/templates/action.cell.html
jniles: This isn't just stylistic - it is encapsulation so we don't repeat ourselves.  What the `bhHasPermission` does is encapsulates the `Session` check so the parent controller (Invoice Registry, Cash Registry, etc) doesn't have to contain any permissions logic.  It only imports Constants to pass to the directive.  To use `allowsRecordDeletion()`, we would need to import the services that are requred to figure out if record deletion is allowed.

One way that we could make this "simpler" would be to assign each action a `data-*` attribute that we affix to the DOM node.  Or we could just pass a string to the directive.  We then we wouldn't even need import the `bhConstants` into the parent controller, and the parent controller can be only concerned with the Invoice Registry, not users and permissions.  However, I judged (and may be wrong) that this is more things to break if we decide to change logic.

Here is how that would work:
```html
<!-- using a data-* attribute -->
<a bh-has-permission data-delete-invoice>
  <!-- all the code -->
</a>

<!-- using a string on the directive-->
<a bh-has-permission="delete-invoice">
  <!-- all the code -->
</a>
```

To address your second point, the functionality shipped to the client (Session, etc) is only for that single user/machine.  So, it shouldn't be "shared" in the sense of sharing state across users and devices, but it is indeed shipped to all clients.

jniles:
@jmcameron yes, there are no server-side checks.  That is a good idea.  It shouldn't be hard to add them.  Let me do so before merging this.

jniles:
@jmcameron this is ready for a second look.

server/controllers/finance/purchases.js
```diff
@@ -600,6 +600,7 @@ function purchaseBalance(req, res, next) {
  */
 async function remove(req, res, next) {
   const pid = db.bid(req.params.uuid);
+  const DELETE_PURCHASE_ORDER = 5;
```
jmcameron: Don't you want to get this from bhConstants? (DRY: Don't Repeat Yourself)

server/controllers/finance/purchases.js
```diff
@@ -608,6 +609,10 @@ async function remove(req, res, next) {
       throw new BadRequest('Can only remove purchase orders that have been confirmed.');
     }
 
+    if (!req.session.actions.includes(DELETE_PURCHASE_ORDER)) {
+      throw new Unauthorized(`User ${req.session.user.username} is not allowed to delete purchase orders.`);
+    }
+
```
jmcameron, line 615: Nice!

server/controllers/finance/transactions.js
```diff
@@ -48,6 +47,30 @@ function parseDocumentMapString(text) {
...
+ * @description
+ * This function checks if the user is authorized to delete the record from their "actions"
+ * set by their user role.
+ *
+ * @returns {Boolean} true if the user is authorized, false if not.
+ */
+function isUserAuthorized(text, actions) {
+  const DELETE_CASH_PAYMENT = 3;
+  const DELETE_INVOICE = 4;
+  const DELETE_VOUCHER = 7;
```
jmcameron: Again, shouldn't these be from bhConstants?

server/controllers/finance/transactions.js
```diff
@@ -59,21 +82,30 @@ function parseDocumentMapString(text) {
...
+      debug(`#delete() resolved transaction ${uuid} to ${documentMap.text}.`);
+
       // route to do the correct safe deletion method.
       const safeDeleteFn = parseDocumentMapString(documentMap.text);
 
+      if (!isUserAuthorized(documentMap.text, actions)) {
+        debug(`#delete() user is not autorized to remove ${documentMap.text}!`);
+        throw new Unauthorized(`User is not authorized to remove ${documentMap.text}.`);
+      }
+
```
jmcameron, line 109: Nice!

server/controllers/stock/index.js
```diff
@@ -476,15 +476,25 @@ async function deleteMovement(req, res, next) {
   const identifier = db.bid(req.params.document_uuid);
   const stockSettings = req.session.stock_settings;
 
+  // delete stock movement.
+  const DELETE_STOCK_MOVEMENT = 6;
```
jmcameron: Here too...

server/models/bhima.sql
```diff
@@ -333,7 +333,13 @@ INSERT INTO `flux` VALUES
...
 INSERT INTO `actions`(`id`, `description`) VALUES
   (1, 'FORM.LABELS.CAN_EDIT_ROLES'),
-  (2, 'FORM.LABELS.CAN_UNPOST_TRANSACTIONS');
+  (2, 'FORM.LABELS.CAN_UNPOST_TRANSACTIONS'),
+  (3, 'USERS.ACTIONS.DELETE_CASH_PAYMENT' ),
+  (4, 'USERS.ACTIONS.DELETE_INVOICE' ),
+  (5, 'USERS.ACTIONS.DELETE_PURCHASE_ORDER'),
+  (6, 'USERS.ACTIONS.DELETE_STOCK_MOVEMENT'),
+  (7, 'USERS.ACTIONS.DELETE_VOUCHER');
+
```
jmcameron, line 342: May want to add a note that these 'id's must match the ones in bhConstants.

server/models/migrations/next/migrate.sql
```diff
@@ -66,4 +66,18 @@ ALTER TABLE depot ADD COLUMN `default_purchase_interval` SMALLINT(5) NOT NULL DE
...
+@author: jniles
+@date: 2021-04-14
+@desc: Add actions to allow users to delete individual record types.
+*/
+INSERT INTO `actions`(`id`, `description`) VALUES
+  (3, 'USERS.ACTIONS.DELETE_CASH_PAYMENT' ),
+  (4, 'USERS.ACTIONS.DELETE_INVOICE' ),
+  (5, 'USERS.ACTIONS.DELETE_PURCHASE_ORDER'),
+  (6, 'USERS.ACTIONS.DELETE_STOCK_MOVEMENT'),
+  (7, 'USERS.ACTIONS.DELETE_VOUCHER');
```
jmcameron, line 83: And here too.

jmcameron (commented):
I tested deleting an invoice.  The permissions worked perfectly.  However (when I had permission) and tried to delete an invoice it complained that a transaction was referencing it so it could not be deleted.  That was good and to be expected.  What I wanted to note is that the red warning banner on the top of the screen from this error stays there when I switch to other pages.  When I did a hard refresh it disappeared.   Also, the error message is rather vague about what transaction to delete.  Since I assume only experienced operators will be doing this, this may not be a problem since they would know how to figure that out.  But should we consider adding more details to the error message?

The posting journal "unpost" works as advertised, although I did not actually try it.

server/controllers/finance/transactions.js
jniles: bhConstants is a client-side script, so no.  We probably should assemble these from a constant script on the server.

https://github.com/IMA-WorldHealth/bhima/issues/1221 requests that we build out all our constants from a central location.  This will require a pretty significant refactor.  I'll make this change the start, but we can later go through and update account type ids, etc, that are hard-coded.

server/controllers/finance/transactions.js
```diff
@@ -4,17 +4,21 @@
...
-
+const debug = require('debug')('bhima:controller:transactions');
 const shared = require('./shared');
 const db = require('../../lib/db');
+const Unauthorized = require('../../lib/errors/Unauthorized');
+const {
+  DELETE_CASH_PAYMENT,
+  DELETE_INVOICE,
+  DELETE_VOUCHER,
+} = require('../../config/constants').actions;
```
jmcameron, line 21: I like it!

server/config/constants.js
```diff
@@ -0,0 +1,118 @@
+// shared constants identical to `bhConstants` on the client side
```
jmcameron, line 1: Excellent!

jmcameron (approved):
Looks great!
Tested and still works...

PR #5574, opened 2021-04-14
feat: remove mailgun and add nodemailer

### LINKED ISSUE #5290: Transition BHIMA to using custom SMTP server with NodeMailer
We currently request mailgun credentials that we do not use since MailGun quit their free tier.  Here they are in our `.env.sample` file:
https://github.com/IMA-WorldHealth/bhima/blob/d3583f63ecc01905d323354a0f9d8df8d4e505fe/.env.sample#L39-L43

Additionally, their node module is unmaintained.

Instead of tying ourselves to mailgun, let's move towards a more flexible solution by just using [nodemailer](https://www.npmjs.com/package/nodemailer) for delivery.  A good point of reference is our [Sunfish project](https://github.com/IMA-WorldHealth/sunfish) where I have already set it up:
https://github.com/IMA-WorldHealth/sunfish/blob/70965671491ea71dbd2884843f7c52eb1397284b/lib/mail.js#L47-L70

To complete this issue:
   1. Remove mailgun-js
   2. Add in nodemailer.
   3. Change `.env.sample` to use SMTP credentials as is found in Sunfish.
   4. Add in the nodemailer transport.
   
Eventually we could create a user interface to set the SMTP variables in a database, but it isn't necessary now.

jniles:
This commit removes mailgun-js which is unmaintained and has been
unusable for us for a while due to shutting down their free tier.
Instead, we replace it with nodemailer which lets you set your own SMTP
credentials as environmental variables.  In the long run, we should
probably provide these as configuration settings a user can set for the
application.

Closes #5290.


jmcameron (changes requested):
I tested this on my system at home and it failed. 
I created a cash report and then tried to email it to myself (using the actions menu).
I got this error:
image

And the server shows this error:
image

I'm ready to assist in debugging this whenever it is convenient.

jmcameron:
> I tested this on my system at home and it failed.
> I created a cash report and then tried to email it to myself (using the actions menu).
> I got this error:

This turned out to be partly a problem with my .env file.
Also in a debugging session with @jniles, we got the mailing of the report working but without the attachment.  To be fixed.

jniles:
@jmcameron I was missing an "s" in "attachments".  Not sure why I thought it had worked before.

I've added additional logging as well.  Can you give it another spin?

.env.sample
```diff
@@ -25,7 +25,7 @@ DB_NAME='bhima_test'
 SESS_SECRET='XopEn BlowFISH'
 
 # define logging level for the npm debug module.
-DEBUG=app,errors
+DEBUG=app,errors,mailer
```
jmcameron: Did you want this in the production version?
BTW, the juicy email messages occur with DEBUG set to ...,bhima:mailer

jmcameron (approved):
I tested this several times and it worked for all tests.  Looks good!

I would have just released it but I did see one oddity that probably has nothing to do with the Bhima SW:  Sometimes the email message (with attachment) gets delayed several (4) minutes before I receive it (in gmail).  When I send the same file again it goes through in seconds.  So I presume the delay is due to Google scanning the PDF for viruses, etc.  What are you seeing when you test this?  BTW, the debug log (bhima:mailer) shows that the message was sent immediately (from Bhima) with appropriate file size, so this reinforces the idea that this has nothing to do with Bhima itself.

.env.sample
jniles: Whoops.  Not sure how that happened.

jniles:
@jmcameron excellent!  Thanks for the quick review!

Yeah, on my connection it's hard to say anything about timing since my internet is ... variable bandwidth, let's say.  I do know that email is asynchronous and we probably shouldn't place too much confidence in it arriving in a certain timeframe (or at all).

It sounds like this all works for you, so I'm going to go ahead and release it. 

bors r+

PR #5575, opened 2021-04-15
feature(Define Purchase interval)

### LINKED ISSUE #5481: Is our calculation of minimum stock correct?
For the resolution of this, see https://github.com/IMA-WorldHealth/bhima/issues/5481#issuecomment-802825729.  This discussion is kept for historical purposes.

----

I've been writing down how the stock flags are calculated and I think the minimum stock is not actually correct, or, at least, it doesn't make sense.

Here is how we calculate security stock and minumum stock:
https://github.com/IMA-WorldHealth/bhima/blob/6e5077e5fb9771532af326ad10aabbbb0ca085ad/server/controllers/stock/core.js#L520-L531

I don't think this definition of minimum stock makes sense.  Shouldn't it be some amount of stock + the security stock?  I can imagine it being better expressed as:

```js
    inventory.S_MIN = inventory.S_SEC + (CMM * inventory.min_months_stock);
 ```
 
 I rewrote this code for clarity in the last year.  Did I make a mistake, is this a bug, or does our definition actually make sense?  The more I think about it, the more I feel like it should be the method I propose above, rather the our current method.

lomamech:
- Addition of the option to define the default order interval at the stock setting and depots level

-This PR requires the execution of the scripts which are in the migrate.sql file

closes #5481


lomamech:
@jniles  All Checks have passed

jniles (approved):
Looks good to me.  I've tested it with Vanga's database and it seems to work fine.

bors r+

PR #5578, opened 2021-04-19
Allow Purchase Requests to use different currencies

jmcameron:
This PR allows a purchase order (PO) to use any supported currency.  
NOTE: Only one currency can be used per purchase order.

**TESTING Suggestion**
- Use bhima_test (make sure to do 'yarn' and 'yarn build:db')
- Test creating POs:  Purchase > Purchase
   - Create PO with default FC currency
   - This should work like before
   - Note that the unit purchase price, amount, and total in the bottom should be FC
   - Add item, notes and submit
   - Examine the receipt and verify it is in FC
- Go to the purchase order registry:  Purchase > Purchase Registry
   - Notice the new PO is shown in FC
   - Try editing it, it should default to FC
   - Change the currency and save, reexamine the purchase registry
- Verify creating POs in all 3 currencies ($, FC, Euro)
   - Use inventory code:  DINJ_TRIB1A_0 (Vitamines B1+B6+B12...)
   - NOTE: After you do the first one, you need to do a hard refresh
                or go to another page to force the inventory to reload.
                Once an item is used, it assumes you will not want to
                order more of the same inventory item in the session.
                If this should be done, let's do it in a future PR.
   - The unit cost you should see for each is:
       - $ => 3.62
       - FC => 3366.6
       - Euro 3.04
    - Double-check these in the PO registry
- Verify stock entry in other currencies
    - Confirm one the of the POs above (eg, for Euros)
    - Do stock entry via Purchase Order. 
    - Note that when entering lots, the currency matches.

**Future Related Work**
- We need more graceful handling of new currencies without defined exchange rates.  This will require updating some of the exchange rate queries to handle currencies with missing exchange rates.   Currently even setting the exchange rate for a new currency via the Admin > Exchange rate fails because of the current limitations. (See https://github.com/IMA-WorldHealth/bhima/issues/5581)
- I did not test Stock entry via transfers.  This probably needs further work in a later PR.

### DIFF
--- a/client/src/less/bhima-bootstrap.less
+++ b/client/src/less/bhima-bootstrap.less
@@ -286,7 +286,7 @@ input[type=number] {-moz-appearance: textfield;}
 
 .table-report > tbody > tr > td,
 .table-report > tbody > tr > th {
-  border: 1px solid #ddd;
+  border: 1px solid black;
 }
 
 .table-report > thead > tr > th {
--- a/client/src/modules/purchases/create/PurchaseOrderForm.js
+++ b/client/src/modules/purchases/create/PurchaseOrderForm.js
@@ -31,6 +31,9 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
     // bind the cache key
     this.cache = AppCache(cacheKey);
 
+    // Save a default for the current exchange rate
+    this.currentExchangeRate = 1.0;
+
     // set up the inventory
     // this will be referred to as PurchaseOrderForm.inventory.available.data
     this.inventory = new Pool({ identifier : 'uuid', data : [] });
@@ -80,6 +83,7 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
 
     this.details.date = new Date(order.date);
     this.details.cost = order.cost;
+    this.details.currency_id = order.currency_id;
     this.details.note = order.note;
     this.details.supplier_uuid = order.supplier_uuid;
 
@@ -205,8 +209,7 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
 
       // do not sum the row if the row is invalid
       if (unit._invalid) { return value; }
-
-      unit.total = unit.unit_price * unit.quantity;
+      unit.total = unit.unit_price * unit.quantity; // NB: In selected currency
       return unit.total + value;
     }, 0);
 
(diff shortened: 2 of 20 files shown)


lomamech:
I just went through adding a new currency in BHIMA with multiple report and generally it works fine.

When I wanted to test with the IMCK Tshikaji Database, for years before, I found that since the new € currency was not defined, the default exchange rate is equal to 1, it will be it may be necessary to display a message to indicate that there is not a defined exchange rate for the period

When analyzing the modifications of the purchase order module, it would be preferable that the company currency be defined as the default currency, so that users may place their orders with the first currency on the list.
image

lomamech:
I have just made a purchase order in Congolese francs but when I try to enter stock from this purchase order, the purchase cost is only displayed against the currency of the 'business without converting

image

lomamech:
@jmcameron The rest of the work is wonderful

jmcameron:
Good catch!  Thanks.

jmcameron:
> When analyzing the modifications of the purchase order module, it would be preferable that the company currency be defined as the default currency, so that users may place their orders with the first currency on the list.

Done.

jmcameron:
> When I wanted to test with the IMCK Tshikaji Database, for years before, I found that since the new € currency was not defined, the default exchange rate is equal to 1, it will be it may be necessary to display a message to indicate that there is not a defined exchange rate for the period.

I've partially addressed this with warning messages in the PO form (and not allowing any currency other than the enterprise currency while ANY currencies have missing exchange rates.  This turned out to be a more pervasive problem that I suggest handling later.  Please see the note I added to the end of the main PR description above.

jmcameron:
Note: This PR also makes a small cosmetic change to all forms.  Formerly some of the table cell edges were light gray:
image

This PR fixes that:
image

client/src/less/bhima-bootstrap.less
```diff
@@ -286,7 +286,7 @@ input[type=number] {-moz-appearance: textfield;}
 
 .table-report > tbody > tr > td,
 .table-report > tbody > tr > th {
-  border: 1px solid #ddd;
+  border: 1px solid black;
```
jniles, line 289: Well, that was easy.

client/src/modules/purchases/create/PurchaseOrderForm.js
```diff
@@ -299,10 +302,51 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
...
 
+  /**
+   * @method setCurrencyId
+   *
+   * @description
+   * Set the currency ID for this order
+   *
+   * @param {number} currencyId
+   */
+  PurchaseOrderForm.prototype.setCurrencyId = function setCurrencyId(currencyId) {
```
jniles, line 322: Yes!  I really like the cleanliness of setting this in its own function.

client/src/modules/purchases/create/PurchaseOrderForm.js
```diff
@@ -299,10 +302,51 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
...
+   */
+  PurchaseOrderForm.prototype.setExchangeRate = function setExchangeRate(newExchangeRate) {
+    this.currentExchangeRate = newExchangeRate;
+
+    // Update unit prices for all order items
+    this.store.data.forEach((item) => {
+      if (!item._initialised) { return; }
+      // We have no way to tell if this form is being invoked when creating
+      // or editing a purchase, so do this work-around to get the inventory item
+      const avItem = this.inventory.unavailable.get(item.inventory_uuid);
```
jniles, line 342: Hmmm... if this ever comes up again, we may just add a function to the `Pool` that will do this for us.  This is a good solution here.

jniles (approved):
Thanks @lomamech for doing a previous review.

Code looks good to me.

Firstly, this is an awesome PR.  Your changes spanned from handlebars, SQL, node, client-side JS and translation files, and did is all correctly and seemlessly like a seasoned BHIMA dev.  Further, you were changing rather complex finance concepts (multi-currency, exchange rates, pricing, etc) and dealing with BHIMA's quirks (JSON files and FormService definitions).  It's really, really cool to see how easily it all came together and if you haven't taken a minute to pat yourself on the back, do so now!

Because we have broken master's stock entry from purchase orders, I'm going to merge this to focus on getting that fixed.  However, could you do a follow-up to this PR that updates the existing [client-side unit tests for purchase orders](https://github.com/IMA-WorldHealth/bhima/blob/master/test/client-unit/services/PurchaseOrderFormService.spec.js) with a scenario or two that involves changing the currency and exchange rate?  Specifically, I'd be looking for these tests:
  1. Calling `setExchangeRate()` updates the `unit_prices` of each item.
  2. Calling `setExchangeRate()` updates the total value of the form (ensuring the this.digest() is working with exchange rates).
  3. Calling `setExchangeRate()` updates `this.currentExchangeRate`
  4. `this.currentExchangeRate` defaults to 1.

Once again, great work!

bors r+

PR #5582, opened 2021-04-21
fix(stock): remove lot.origin_uuid

### LINKED ISSUE #5580: Stock entries of pre-defined lots are broken
As noted by @jmcameron in the BHIMA chat, the stock entry from purchase orders is broken when one uses an stock lot that is pre-existing.  This is because the delay calculation uses the `origin_uuid` of the lot.  Ideally, it shouldn't.  It should use the stock_movement that is associated with the entry from purchase order.  This is stored in the `entity_uuid` column.

jniles:
This commit fixes the stock entry from purchase by removing calculations that depend on lot.origin_uuid and replace them with
stock_movement.entity_uuid.  Since we can re-use lots, this breaks the logic of some of our reports and status checks.  I've removed `lot.origin_uuid` entirely as it no longer makes sense to associate lots with their origins since a single lot can be reused across integrations, purchase orders, and donations.

Here are some other changes landing in this PR:

1. I've added a feature to the stock movement registry to quickly look up the purchase order that the stock movement came from:
image
Note that this isn't going to work well with our test dataset because our test data for stock movements is incoherent.

2. I've updated the name of the `PurchaseListController` to `PurchaseRegistryController` bringing it in line with invoices, vouchers, and cash payments.

3. I've removed the `getLotOrigins` route entirely.  I couldn't find where it was used currently and am not sure it was ever used.

4. I've updated the logic of purchase interval calculations to use the stock movement's `entity_uuid` column (which contains the `purchase.uuid`) instead of the `lot.origin_uuid` for lookups.

These tests are hard to work with because of how linked they are and how incoherent the data is.  It is probably worth redoing all our stock tests with relative dates. for integration and end to end tests.

Closes #5580.


jmcameron (approved):
I looked over the code and it looks fine.  

I tested the new action menu item for showing a purchase order from the action menu for stock movements (for Purchase Orders) and it works fine.

All the tests pass (on my PC) except the End-to-End tests. It fails when creating a new Role.   When I ran Bhima and look at the role management page there was no way there (obvious) to create a new Role (yes, superuser has full permissions and is enabled for all actions).  I tried the manual process in master too and saw the same thing.  The E2E test that fails:
```
  542 passing (16m)
  33 pending
  1 failing

  1) Role Management Tests
       Role Management
         should add a new role:
     WebDriverError: element not interactable
  (Session info: headless chrome=87.0.4280.141)
Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:25:53'
System info: host: 'basil', ip: '127.0.1.1', os.name: 'Linux', os.arch: 'amd64', os.version: '4.19.104-microsoft-standard', java.version: '11.0.10'
Driver info: driver.version: unknown
```
I see the same failure on master.

Since this is probably a problem with master and not this PR, I'm going ahead and approving it.

lomamech (approved):
I had proceeded by investigating the bug on the Master branch, then I saw the correction and it works really well, I also identified the same error when running the E2E test
image

jniles:
Thanks @jmcameron @lomamech !

PR #5585, opened 2021-04-22
Fix permissions for editing roles actions

jmcameron:
This fixes an oversight in the recent refactoring of action permissions.  Now anyone with CAN_EDIT_ROLES should be able to add roles in the Role management interface.

**Testing**
- Run as superuser or other admin account
- Go to Administration > Role Management
- Note that the [Add] role button now appears on the top right of the form.
- You should be able to add, edit, and delete roles


jniles (approved):
Yay!  Tests pass!  Great catch!

bors r+

PR #5587, opened 2021-04-23
Fix currency labeling for stock movements from purchase order entries

jmcameron:
This PR fixes omissions in the recent selectable-currency purchase order updates whihc 
missed updating stock movements for purchase order entries. 

**TESTING**
- Use any database
- Pick a non-enterprise currency (below I use Euros, but it could be FC too).
- Use the Admin > Exchange rate  and make a note of the current Euro exchange rate
   - If the rate for Euros is not defined, define it now.
- Create a purchase order (PO) in Euros.
   - Confirm the PO if necessary
   - Notice all interactions and the Receipt clearly show Euros (once it is selected in the PO creation)
- Do stock entry for that PO.
- Go to Stock Movements and find just-completed stock movement for the PO
- Notice that the cost for the movement is shown in USD (the enterprise currency) and that the exchange rate has been applied correctly.
- Use the action menu for that movement and look at the report. The PO report should be in the original currency (Euros) and should be similar to the one for that PO (you can check using the PO registry, action menu...)

### DIFF
--- a/client/src/modules/stock/movements/registry.js
+++ b/client/src/modules/stock/movements/registry.js
@@ -2,10 +2,11 @@ angular.module('bhima.controllers')
   .controller('StockMovementsController', StockMovementsController);
 
 StockMovementsController.$inject = [
-  'StockService', 'NotifyService', 'uiGridConstants',
-  'StockModalService', 'LanguageService', 'SessionService', 'FluxService',
-  'ReceiptModal', 'GridGroupingService', '$state', 'GridColumnService', 'GridStateService', '$httpParamSerializer',
-  '$translate', 'bhConstants', 'ModalService',
+  'StockService', 'NotifyService', 'uiGridConstants', 'StockModalService',
+  'LanguageService', 'SessionService', 'FluxService', 'ReceiptModal',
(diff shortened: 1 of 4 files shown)


jniles:
I'm a bit at a loss.  I thought we did the currency conversion when we did the stock entry from whatever the purchase currency was into the currency of the enterprise?  If so, why do we need to know the purchase price when showing the stock movements if that is the case?  Or did I misunderstand?

We should do the currency exchange at stock entry so that we are storing `unit_cost` in the enterprise currency in the stock movements table.  If that isn't what is happening, that is what should be happening.  On the finance side, this is likely already happening (the _voucher_ generated creates a posting_journal transaction that has the values stored in the enterprise currency valued at that day's exchange rate).  We should continue to show those same numbers, even if the exchange rate changes.  The easiest way is to just always show stock in the enterprise currency.

jmcameron:
> We should do the currency exchange at stock entry so that we are storing `unit_cost` in the enterprise currency in the stock movements table. 

Good catch.  I was fixing the symptoms, not the real problem!  I'll see if I can finish that today.

PR #5591, opened 2021-04-26
feat(stock): link stock movement transaction

### LINKED ISSUE #5276: Link the associated transaction for a stock movement if it exists
On our stock movements registry, we provide a list of stock movements and associated values.  See below:

image

It would be nice to have a link to the transaction in the Journal if the movement resulted in a financial transaction.  All movements that are not transfers/requisitions between depots will have a financial impact (either increase or decrease the stock).  If we could add a link in the "actions" menu similar to what we have for our financial registries, that would be helpful:

image

jniles:
Adds the ability to link stock movements with the journal if the automated stock accounting is working.  This makes it easy to correlate the financial writings with the stock ones.

Closes #5276.

Here it is in action:

CEzSYAON20

### DIFF
--- a/server/controllers/finance/journal/index.js
+++ b/server/controllers/finance/journal/index.js
@@ -108,7 +108,9 @@ function naiveTransactionSearch(options, includeNonPosted) {
 // if posted ONLY return posted transactions
 // if not posted ONLY return non-posted transactions
 function buildTransactionQuery(options, posted) {
-  db.convert(options, ['uuid', 'record_uuid', 'uuids', 'record_uuids', 'reference_uuid']);
+  db.convert(options, [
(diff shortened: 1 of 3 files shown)


server/controllers/finance/journal/index.js
```diff
@@ -160,6 +162,8 @@ function buildTransactionQuery(options, posted) {
   filters.equals('hrEntity', 'text', 'em');
   filters.equals('hrRecord', 'text', 'dm1');
   filters.equals('hrReference', 'text', 'dm2');
+  filters.equals('entity_uuid');
+  filters.equals('stockReference', 'reference_uuid', 'p');
```
mbayopanda, line 166: 👍🏽

mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #5592, opened 2021-04-26
Save purchase order stock movements in enterprise currency

jmcameron:
This PR ensures that stock movements from purchase orders (POs) are saved in the enterprise currency.
The general approach is:
- Purchase orders and related PO items are saved in user-specified currency.  The lot unit price is based on the inventory unit prices with the correct exchange rate applied.
- When entries are done from POs, once the lots are selected and the final [Submit] button is pressed, the stock movement is saved in enterprise currency.

**TESTING**
- use bhima_test
- yarn build:db
- Use the Admin > Exchange rate and make a note of the current Euro, FC exchange rates
   - If the rate for Euros is not defined, define it now as 0.84
- Create three POs ($, FC, and Euros) for 10 of DINJ_TRIB1A2_0 (Vitamines B1+B6+B12...).
- Unit prices should be:
  - $ -> 3.62
  - FC -> 930*3.62 = 3366.6
  - Euro -> 0.84*3.62 = 3.048
- As you create each PO, verify the unit price and totals in the PO Receipt
- In the PO registry confirm each PO (if necessary)
  - Verify the totals for each PO in the registry
- Do stock entry for each of the three POs
  - Note the Search Purchase modal will show totals in the respective currencies.  Verify these are right.
  - For each PO
    - Select the full 10 items in the "Lots" selector
    - In the Lots selector, notice the unit cost is shown in the respective currency
    - When you press [Submit] check the Stock Entry - Purchase receipt.  It should show unit and total in the enterprise currency.
- Review the stock movements for these three entries.  They should be identical and in enterprise currency ($).

### DIFF
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -5,7 +5,8 @@ angular.module('bhima.controllers')
 StockEntryController.$inject = [
   'InventoryService', 'NotifyService', 'SessionService', 'util',
   'bhConstants', 'ReceiptModal', 'PurchaseOrderService',
-  'StockFormService', 'StockService', 'StockModalService', 'LotService',
+  'StockFormService', 'StockService', 'StockModalService',
+  'LotService', 'ExchangeRateService',
   'uiGridConstants', 'Store', 'uuid', '$translate',
 ];
 
@@ -16,8 +17,11 @@ StockEntryController.$inject = [
  * This controller is responsible to handle stock entry module.
  */
 function StockEntryController(
-  Inventory, Notify, Session, util, bhConstants, ReceiptModal, Purchase,
-  StockForm, Stock, StockModal, Lots, uiGridConstants, Store, Uuid, $translate,
+  Inventory, Notify, Session, util,
+  bhConstants, ReceiptModal, Purchase,
+  StockForm, Stock, StockModal,
+  Lots, Exchange,
+  uiGridConstants, Store, Uuid, $translate,
 ) {
   // variables
   let inventoryStore;
@@ -44,6 +48,7 @@ function StockEntryController(
   vm.submit = submit;
   vm.reset = reset;
   vm.onDateChange = onDateChange;
+  vm.$loading = false;
 
   vm.gridOptions = {
     appScopeProvider : vm,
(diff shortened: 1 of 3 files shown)


client/src/modules/stock/entry/entry.js
```diff
@@ -537,6 +552,14 @@ function StockEntryController(
 
     movement.lots = Stock.processLotsFromStore(vm.stockForm.store.data, vm.movement.entity.uuid);
 
+    // fix the unit price, if necessary
+    if (vm.currencyId !== vm.enterprise.currency_id) {
+      const exchangeRate = Exchange.getCurrentRate(vm.currencyId);
+      movement.lots.forEach(lot => {
+        lot.unit_cost /= exchangeRate;
+      });
+    }
```
mbayopanda, line 561: 👍🏽

client/src/modules/stock/entry/modals/findPurchase.modal.js
```diff
@@ -52,7 +53,7 @@ function StockFindPurchaseModalController(
       field            : 'cost',
       displayName      : 'STOCK.AMOUNT',
       headerCellFilter : 'translate',
-      cellFilter       : `currency:${Session.enterprise.currency_id}`,
+      cellFilter       : `currency:row.entity.currency_id`,
```
mbayopanda, line 56: 👍🏽

mbayopanda (approved):
The feature works as expected, movements (from PO in a given currency) are inserted in the enterprise currency 👍🏽 

LGTM

client/src/modules/stock/entry/entry.js
```diff
@@ -537,6 +552,14 @@ function StockEntryController(
 
     movement.lots = Stock.processLotsFromStore(vm.stockForm.store.data, vm.movement.entity.uuid);
 
+    // fix the unit price, if necessary
+    if (vm.currencyId !== vm.enterprise.currency_id) {
+      const exchangeRate = Exchange.getCurrentRate(vm.currencyId);
+      movement.lots.forEach(lot => {
+        lot.unit_cost /= exchangeRate;
```
jniles, line 559: This is good ... until an error occurs.  Not sure why an error would occur, but that would cause this code to doubly exchange.

We should reset the `unit_cost` if an error occurs.  I can make this change.

jniles (changes requested):
Looks pretty close.  I'm worried that an error (http timeout, etc) would cause a user to submit twice and therefore round twice.  If an error occurs, we'll want to reset the original values prior to the user submitting again.

jniles (approved):
Looks good to me.

bors r+

jmcameron:
Thanks @jniles !

client/src/modules/stock/entry/entry.js
jmcameron: Were you able to fix this or should this be noted as an Issue?

PR #5595, opened 2021-04-28
Add database port in the database build script

mbayopanda:
This PR adds the database port in the database build script which will help to switch easily to a different version of MySQL by its running port

### DIFF
--- a/sh/build-database.sh
+++ b/sh/build-database.sh
@@ -21,41 +21,44 @@ if [ "$1" == "debug" ]; then
     fout=/dev/tty
 fi
 
+# default database port
+DB_PORT=${DB_PORT:-3306}
+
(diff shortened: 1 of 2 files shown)


jniles:
Not a bad idea.  Can you add this to the `.env.sample`?  And change the `build-init-database.sh` script?

It might also be useful to have a straightforward default value.  You can do this in bash by putting:
```sh
DB_PORT=${DB_PORT:-3306}
```
at the top of the `.sh` scripts.

jniles:
Looks like it is just `.env.sample` that needs to be updated, then we can merge this.

mbayopanda:
Here is .env.sample : 
image

We already have: `DB_PORT` defined

jniles:
Oh!  Didn't realize it was already there.  My apologies.

bors r+

PR #5599, opened 2021-04-30
Rumer report

### LINKED ISSUE #5395: Implement the RUMER report
We need to implement a "Rumer" report that shows the consumption of different drugs by day for one month with a total at the end.  The X-axis is the number of days (28/30/31 days) and the Y-axis is each drug.  Like this:

  | Opening Balance | 01/01/2021 | 01/02/2021 | … | 31/01/2021 | Total
-- | -- | -- | -- | -- | -- | --
Quinine 500mg | x | x | x | … | x | x
Cipro 250MG | x | x | x | .. | x | x
Alcool 1L | x | x | x | .. | x | x

This report is a standard stock report used by pharmacists in the DRC.  While it is true that you can recover all this information from BHIMA, we keep getting asked for it.  Might as well implement it.

Here is an example from AG Neron: [RUMER.xlsx](https://github.com/IMA-WorldHealth/bhima/files/6009338/RUMER.xlsx)

lomamech:
- Implement the rumer report
Note: This PR requires the execution of updates in the migrate file

closes #5395

### DIFF
--- a/client/src/modules/reports/generate/rumer_report/rumer_report.config.js
+++ b/client/src/modules/reports/generate/rumer_report/rumer_report.config.js
@@ -1,0 +1,83 @@
+angular.module('bhima.controllers')
+  .controller('rumer_reportController', rumerReportController);
+
+rumerReportController.$inject = [
+  '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state',
+  'LanguageService',
+];
+
+function rumerReportController($sce, Notify, SavedReports, AppCache, reportData, $state, Languages) {
+  const vm = this;
+  const cache = new AppCache('rumer_report');
+  const reportUrl = 'reports/stock/rumer_report';
+
+  // default values
+  vm.reportDetails = {
+    includePurchaseEntry : 1,
+  };
+  vm.previewGenerated = false;
+
+  // check cached configuration
+  checkCachedConfiguration();
+
+  vm.onSelectDepot = depot => {
+    vm.reportDetails.depotUuid = depot.uuid;
+    vm.reportDetails.depot_text = depot.text;
+  };
+
+  vm.onSelectFiscalYear = (fiscalYear) => {
+    vm.reportDetails.fiscal_id = fiscalYear.id;
+    vm.reportDetails.fiscalYearStart = fiscalYear.start_date;
+  };
+
+  vm.onSelectPeriod = (period) => {
+    vm.reportDetails.period_id = period.id;
+    vm.reportDetails.end_date = period.end_date;
+    vm.reportDetails.start_date = period.start_date;
+  };
+
+  vm.clear = key => {
+    delete vm[key];
+  };
+
+  vm.clearPreview = () => {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.preview = form => {
+    if (form.$invalid) {
+      return 0;
+    }
+
+    // update cached configuration
+    cache.reportDetails = angular.copy(vm.reportDetails);
+    angular.extend(vm.reportDetails, { lang : Languages.key });
+
+    return SavedReports.requestPreview(reportUrl, reportData.id, angular.copy(vm.reportDetails))
+      .then((result) => {
+        vm.previewGenerated = true;
+        vm.previewResult = $sce.trustAsHtml(result);
+      })
+      .catch(Notify.handleError);
+  };
+
+  vm.requestSaveAs = function requestSaveAs() {
+    const options = {
+      url : reportUrl,
+      report : reportData,
+      reportOptions : angular.copy(vm.reportDetails),
+    };
+
+    return SavedReports.saveAsModal(options)
+      .then(() => {
+        $state.go('reportsBase.reportsArchive', { key : options.report.report_key });
+      })
+      .catch(Notify.handleError);
+  };
+
+  function checkCachedConfiguration() {
+    vm.reportDetails = angular.copy(cache.reportDetails || {});
+  }
+
+}
--- a/client/src/modules/reports/generate/rumer_report/rumer_report.config.js
+++ b/client/src/modules/reports/generate/rumer_report/rumer_report.config.js
@@ -11,10 +11,7 @@ function rumerReportController($sce, Notify, SavedReports, AppCache, reportData,
   const cache = new AppCache('rumer_report');
   const reportUrl = 'reports/stock/rumer_report';
 
-  // default values
-  vm.reportDetails = {
-    includePurchaseEntry : 1,
-  };
+  vm.reportDetails = { };
   vm.previewGenerated = false;
 
   // check cached configuration
--- a/client/src/modules/reports/generate/rumer_report/rumer_report.config.js
+++ b/client/src/modules/reports/generate/rumer_report/rumer_report.config.js
@@ -31,9 +31,12 @@ function rumerReportController($sce, Notify, SavedReports, AppCache, reportData,
   };
 
   vm.onSelectPeriod = (period) => {
+
     vm.reportDetails.period_id = period.id;
     vm.reportDetails.end_date = period.end_date;
     vm.reportDetails.start_date = period.start_date;
+    vm.reportDetails.translate_key = period.translate_key;
+    vm.reportDetails.year = period.year;
   };
 
   vm.clear = key => {
--- a/server/controllers/stock/reports/rumer.report.handlebars
+++ b/server/controllers/stock/reports/rumer.report.handlebars
@@ -25,7 +25,7 @@
             {{#each header as | key |}}
               <th style="min-width:20px;" class="text-center">{{key}}</th>
             {{/each}}
-            <th style="width: 7.5%;" class="text-center">{{translate 'FORM.LABELS.TOTAL'}} </th>
+            <th style="width: 7.5%;" class="text-center">{{translate 'REPORT.RUMER.TOTAL_EXITS'}} </th>
             <th style="width: 7.5%;" class="text-center">{{translate 'REPORT.RUMER.STOCK_END'}}</th>
           </tr>
         </thead>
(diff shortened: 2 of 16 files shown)


lomamech:
image

lomamech:
image

Rumer Report with stock_movement_status data

lomamech:
Rumer Report with stock_movement_status data

jniles:
@lomamech, can you rebase this to `master`?

server/controllers/stock/reports/rumer.report.handlebars
```diff
@@ -0,0 +1,54 @@
...
+            <th style="width: 7.5%;" class="text-center">{{translate 'FORM.LABELS.TOTAL'}} </th>
+            <th style="width: 7.5%;" class="text-center">{{translate 'REPORT.RUMER.STOCK_END'}}</th>
+          </tr>
+        </thead>
+
+          <tbody>
+            {{#each configurationData}}
+            <tr>
+              <td> {{ inventoryText }} </td>
+              <td><strong> {{ quantityOpening }} </strong></td>
```
jniles: These should be `class="text-right"` as well.

server/controllers/stock/reports/rumer.report.handlebars
```diff
@@ -0,0 +1,54 @@
...
+              <td> {{ quantityTotalEntry }} </td>
+              {{#each dailyConsumption }}
+                <td class="text-right">
+                {{#if value}}
+                  {{ value }}
+                {{/if}}  
+                </td>
+              {{/each}}
+              <td><strong> {{ quantityTotalExit }} </strong></td>
+              <td><strong> {{ quantityEnding }} </strong></td>
```
jniles: Both `quantityTotalExit` and `quantityEnding` should be `text-right` as well.

server/controllers/stock/reports/stock/rumer.js
```diff
@@ -0,0 +1,172 @@
...
+  };
+
+  _.defaults(params, DEFAULT_PARAMS);
+
+  try {
+    const reporting = new ReportManager(TEMPLATE, req.session, params);
+
+    const sqlDailyConsumption = `
+      SELECT BUID(sms.inventory_uuid) AS uuid, inv.code, inv.text AS inventory_text,
+      sms.out_quantity_consumption AS quantity, sms.date
```
jniles: I talked to AG Neron, and apparently this needs to have both the consumption and the exits due to loss.  So, this should be changed to
```sql
SELECT (sms.out_quantity_consumption + sms.out_quantity_exit) AS quantity
```

server/controllers/stock/reports/stock/rumer.js
```diff
@@ -0,0 +1,172 @@
...
+        FROM stock_movement_status AS sms
+        JOIN inventory AS inv ON inv.uuid = sms.inventory_uuid
+      WHERE sms.depot_uuid = ? AND DATE(sms.date) >= DATE(?) AND DATE(sms.date) <= DATE(?)
```
jniles: Because [the `sms.date` column is a `DATE`](https://github.com/IMA-WorldHealth/bhima/blob/97eabad91d26934b575744c0bc27f3067b9e376b/server/models/schema.sql#L2027) and not a `DATETIME` or` TIMESTAMP`, you don't need to wrap it in `DATE()`.   By wrapping it, you lose a performance gain from the INDEX.

You can change this code to:
```sql
WHERE sms.depot_uuid = ? AND sms.date >= DATE(?) AND sms.date <= DATE(?)
```

server/controllers/stock/reports/stock/rumer.js
```diff
@@ -0,0 +1,172 @@
...
+    `;
+
+    const sqlMonthlyConsumption = `
+      SELECT BUID(sms.inventory_uuid) AS uuid, inv.code, inv.text AS inventory_text,
+      SUM(sms.out_quantity_consumption) AS quantity, sms.date
+        FROM stock_movement_status AS sms
+        JOIN inventory AS inv ON inv.uuid = sms.inventory_uuid
+      WHERE sms.depot_uuid = ? AND DATE(sms.date) >= DATE(?) AND DATE(sms.date) <= DATE(?)
```
jniles: Same comment about `DATE` here and the `exit` + `consumption` calculation.

server/controllers/stock/reports/stock/rumer.js
```diff
@@ -0,0 +1,172 @@
...
+      GROUP BY inv.uuid;
+    `;
+
+    const sqlStockEntryMonth = `
+      SELECT BUID(inv.uuid) AS uuid, inv.code, inv.text AS inventory_text, SUM(sm.quantity) AS quantity,
+      sm.date, sm.is_exit
+        FROM stock_movement AS sm
```
jniles: You should be able to get this from `stock_movement_status` from the column [`in_quantity`](https://github.com/IMA-WorldHealth/bhima/blob/97eabad91d26934b575744c0bc27f3067b9e376b/server/models/schema.sql#L2029).

server/controllers/stock/reports/stock/rumer.js
```diff
@@ -0,0 +1,172 @@
...
+    const [inventoriesOpening, inventoriesConsumed,
+      inventoriesEntry, monthlyConsumption, inventoriesEnding] = await Promise.all([
+      core.getInventoryQuantityAndConsumption(parameterOpeningStock),
+      db.exec(sqlDailyConsumption, [db.bid(params.depotUuid), params.start_date, params.end_date]),
+      db.exec(sqlStockEntryMonth, [db.bid(params.depotUuid), params.start_date, params.end_date]),
+      db.exec(sqlMonthlyConsumption, [db.bid(params.depotUuid), params.start_date, params.end_date]),
+      core.getInventoryQuantityAndConsumption(parameterEndingStock),
+    ]);
+
+    inventoriesEnding.forEach(inventory => {
```
jniles, line 105: It might be helpful to add an option to remove the inventories with `0` movements.  Could you add that?

server/controllers/stock/reports/rumer.report.handlebars
```diff
@@ -0,0 +1,54 @@
...
+      <table class="table table-striped table-condensed table-report table-bordered">
+        <thead>
+          <tr style="background-color:#ddd;">
+            <th style="min-width:100px;">{{translate 'FORM.LABELS.INVENTORY'}}</th>
+            <th class="text-center"> {{translate 'REPORT.RUMER.STOCK_BEGINNING'}} </th>
+            <th class="text-center"> {{translate 'REPORT.RUMER.TOTAL_ENTRIES'}} </th>
+            {{#each header as | key |}}
+              <th style="min-width:20px;" class="text-center">{{key}}</th>
+            {{/each}}
+            <th style="width: 7.5%;" class="text-center">{{translate 'FORM.LABELS.TOTAL'}} </th>
```
jniles: Could you rename this column to "Total Exits"?  It would match the "Total Entries" at the beginning a lot better.


image

server/controllers/stock/reports/stock/rumer.js
```diff
@@ -0,0 +1,172 @@
...
+              d.value = consumed.quantity;
+            }
+          });
+        }
+      });
+
+      inventory.dailyConsumption = dailyConsumption;
+    });
+
+    if (inventoriesOpening.length) {
```
jniles: You are doing lots of these loops that look very similar.

Can you combine them into a single loop?  Instead of checking if `inventoriesOpening.length`, then doing a `configurationData.forEach()`, just add the `inventoriesOpening.forEach()` into the `configurationData.forEach()` loop.  Like this:

```js
configurationData.forEach(row => {

  // first, add the opening balances:
  inventoriesOpening.forEach(inv => {
    if (row.inventoryUuid === inv.inventory_uuid) { row.quantityOpening = inv.quantity; }
  });

  // next process the entries
  inventoriesEntry.forEach(inv => {
    if (row.inventoryUuid === inv.uuid) { row.quantityTotalEntry = inv.quantity; }
  });

  // finally process the exit quantities
  monthlyConsumption.forEach(inv => {
    if (row.inventoryUuid === inv.uuid) { row.quantityTotalExit = inv.quantity; }
  });
});
```

jniles (commented):
@lomamech just a few comments here.  Overall looks really good!

I think my inline comments should be clear enough.  Let me know if you have any questions.

server/controllers/stock/reports/stock/rumer.js
lomamech: A small precision, do you allude to the products whose quantity at the beginning of the month is null and have had no movement during the month

jniles: Yes, that is correct.

jniles: Basically allow the user to remove products where the beginning balance is 0, the movements are 0, and the ending balance are 0.

lomamech:
image

jniles:
bors try

client/src/modules/reports/generate/rumer_report/rumer_report.config.js
```diff
@@ -0,0 +1,86 @@
...
+];
+
+function rumerReportController($sce, Notify, SavedReports, AppCache, reportData, $state, Languages) {
+  const vm = this;
+  const cache = new AppCache('rumer_report');
+  const reportUrl = 'reports/stock/rumer_report';
+
+  // default values
+  vm.reportDetails = {
+    includePurchaseEntry : 1,
```
jniles: I think we can probably remove this. I don't think it is ever used.

server/controllers/stock/reports/stock/rumer.js
```diff
@@ -0,0 +1,212 @@
...
+    const sqlMonthlyConsumption = `
+      SELECT BUID(sms.inventory_uuid) AS uuid, inv.code, inv.text AS inventory_text,
+      SUM(sms.out_quantity_consumption + sms.out_quantity_exit) AS quantity, sms.date
+        FROM stock_movement_status AS sms
+        JOIN inventory AS inv ON inv.uuid = sms.inventory_uuid
+      WHERE sms.depot_uuid = ? AND sms.date >= DATE(?) AND sms.date <= DATE(?)
+      GROUP BY inv.uuid;
+    `;
+
+    const sqlStockEntryMonth = `
```
jniles: It looks like all these use the same query.  I think we can probably gain a significant performance boost by combining them.  I'll go ahead and make that change in a commit.

jniles (commented):
Everything here looks good to me.  I think I see an easy performance improvement, so I'll make that change in a commit myself.  After that, this is ready to roll.  Great job @lomamech!

PR #5610, opened 2021-05-05
fix(patients): properly count stock movements

### LINKED ISSUE #5443: Patient Financial Status always shows 5 stock movements for 5+ movements
This patient actually has hundreds of movements:

image

But since we only load 5 movements, we don't actually know the number of movements here.  We should either:

1. Remove the "5" value and just put "all" movements
2. Load the number of movements via an API call and display it.

jniles:
Properly count stock movements on the patient overview page.

Closes #5443.

Here is what it looked like before:

image

After:
image

### DIFF
--- a/client/src/js/components/bhPatientFinancialActivity.js
+++ b/client/src/js/components/bhPatientFinancialActivity.js
@@ -69,11 +69,13 @@ function PatientFinancialActivityCtrl(Patients, moment, Session, Constants, $q) 
           if (index < limitRecent) {
             dataMovement.push(item);
           }
+
           $ctrl.dataMovementTotalValue += item.value;
         });
 
         $ctrl.dataMovement = dataMovement;
         $ctrl.noStockMovement = stockData.length === 0;
+        $ctrl.totalMovementCount = stockData.length;
       }).finally(() => {
         $ctrl.loading = false;
       });
--- a/client/src/modules/patients/record/bh-patient-financial-activity.html
+++ b/client/src/modules/patients/record/bh-patient-financial-activity.html
@@ -120,7 +120,7 @@
                   ]
                 })">
                   <i class="fa fa-link"></i>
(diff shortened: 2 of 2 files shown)


jmcameron:
How do I get this report?

jmcameron (approved):
The code looks fine.
I tested this with the Vanga DB and found a good test case with patient reference=PA.HEV.31609. I see that while 5 stock movements are listed, there is a link to 18 -- meaning the PR works properly.

I spotted one "off-by-one" error on the next page.  With the patient above, if I click on the link to see "All 18 distributions to Patient", on the display (at the bottom), it lists 19 movements.  But there are only 18, per the previous page (and a hand count):
image
I think this should be fixed or a new Issue added to address it later; it should be a trivial fix.

The second item I noticed was the financial reports for some patients have financial activity but no stock movements.   I was not sure if this is incorrect or that financial items were occuring without stock movements.  Here is a snap:
image
I think I saw similar things with the IMCK db but I seem to recall some of them were drug related (could be wrong on this).   Since I'm guessing that this is not a problem, I'm approving anyway.

jniles:
@jmcameron, thanks for the review!

The off-by-one error is tricky, because _technically_ there are 19 rows in the grid - counting the grouping row.  This is a ui-grid "feature".  If you have multiple groups, they are also counted (off-by-n).  We could figure out a work around to this, and it affects every grid in the application that has grouping.

One advantage of this is that if the group is condensed, you get a count of the number of groups.  But in this case above, you are right, it is useless.

I'm on the fence about if this is important to address or not, but I've lived with it so long that I'm biased.  In your view, should we change this behavior?

jmcameron:
> I'm on the fence about if this is important to address or not, but I've lived with it so long that I'm biased. In your view, should we change this behavior?

I suggest we not worry about it.  If anyone complains, we can deal with it later.

PR #5615, opened 2021-05-06
Add 'unused stock' status warning token to Articles in Stock display

### LINKED ISSUE #5323: Add a new label/flag of "unused stock" to the stock registry
We currently have multiple labels for different stock items including "out of stock", "low stock", etc.  As suggested in https://github.com/IMA-WorldHealth/bhima/issues/5321, let's make a label for "unused stock."  The definition of "unused stock" is any inventory item that has not had any consumption from a depot (CMM = 0).

Why make a new label?  These are actually interesting stock items - why would they never have a consumption event?  We could use this label to track down stock that either (1) should not be consumable or (2) users are not properly tracking.  It also gives us an easy way to account for the issue described in #5321 and remove unused stock from the "risk of expiration" calculation.

@jmcameron what do you think of this proposal?

jmcameron:
This PR adds a 'Unused Stock' warking token to the status column in the Articles in Stock report.

I chose "magenta" for the color of the "Unused Stock" item, but could change it if anyone has a better suggestion.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5323

**TESTING**
- Suggest using a production database
- Try the Stock > Articles in Stock page
   - Notice that items with CMM=0 show the magenta "Unused Stock" token in the Status column
   - Change the default search to allow exhausted lots
   - Refresh the Articles in Stock page and notice that items with "stock out" show that token instead of the "Unused Stock" token, even if their CMM=0.   I think that "Unused Stock" implies that there is stock to be used but it was not used.  If there is no stock ("stock out") then it is not obvious why no stock was used (it could be used if it had been in stock, or it is obsolete).
- Try the Stock > Report > Stock Inventories report and verify that the "Unsed Stock" labels are shown in the report.

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -926,6 +926,11 @@ a[disabled] {
   background-color: #FFFF00 !important;
 }
 
+.unused-stock-status {
+  color: black !important;
+  background-color: #FF00FF !important;
+}
+
 .out-of-stock {
   /* text-decoration : line-through; */
   background-color: #cc00ff8f !important;
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -222,13 +222,14 @@
     "SERVICE"         : "Service",
     "STATUS"          : {
       "IN_STOCK"      : "In Stock",
+      "IS_IN_RISK_OF_EXPIRATION" : "At risk of expiration",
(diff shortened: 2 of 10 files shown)


jniles (approved):
Looks good to me.  Tested and confirmed working.

bors r+

PR #5633, opened 2021-05-10
fix(reports): select patient before submit

### LINKED ISSUE #5611: Patient Standing Report problem
In Hospital > Reports  > Patient Standing Report
if you click on [Preview], the field for Patient Name is bordered in red but the preview still works --> but fails with a 404 error.  
The form needs to be fixed so that the [Preview] button is disabled until a patient name is selected.

Probably a good idea to check for similar issues with the Employee Standing Report.

jniles:
Fixes the bug in the patient standing report where the report could be submitted even without a fix in the patient standing report.

Closes #5611.

### DIFF
--- a/client/src/modules/reports/generate/patient_standing/patient_standing.html
+++ b/client/src/modules/reports/generate/patient_standing/patient_standing.html
@@ -25,7 +25,9 @@
 
           <!--select the Patient -->
           <bh-find-patient
+            patient-uuid="ReportConfigCtrl.reportDetails.patientUuid"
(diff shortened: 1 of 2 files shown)


jmcameron (approved):
The code changes look good.

I tested it and it works correctly.

jmcameron:
Sorry, can't help.  These tests pass on my machine!

jniles:
bors try

jniles:
For some reason, I have the magic fingers!

bors r+

PR #5639, opened 2021-05-11
chore(stock): add zMergeDepots admin routine

jniles:
Adds a "merge depots" administrative routine to facilitate correction in the field.

### DIFF
--- a/server/models/admin.sql
+++ b/server/models/admin.sql
@@ -246,4 +246,19 @@ BEGIN
   DELETE FROM general_ledger WHERE record_uuid = _record_uuid;
 END$$
 
+DROP PROCEDURE IF EXISTS  zMergeDepots$$
+CREATE PROCEDURE zMergeDepots(
+  IN _old_uuid BINARY(16),
+  IN _new_uuid BINARY(16)
+) BEGIN
+  UPDATE stock_movement SET depot_uuid = _new_uuid WHERE depot_uuid = _old_uuid;
+  DELETE FROM depot_distribution_permission WHERE depot_uuid = _old_uuid;
+  UPDATE stock_assign SET depot_uuid = _new_uuid WHERE depot_uuid = _old_uuid;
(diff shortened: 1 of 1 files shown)


server/models/admin.sql
```diff
@@ -246,4 +246,19 @@ BEGIN
...
+  UPDATE stock_movement SET depot_uuid = _new_uuid WHERE depot_uuid = _old_uuid;
+  DELETE FROM depot_distribution_permission WHERE depot_uuid = _old_uuid;
+  UPDATE stock_assign SET depot_uuid = _new_uuid WHERE depot_uuid = _old_uuid;
+  UPDATE stock_requisition SET depot_uuid = _new_uuid WHERE depot_uuid = _old_uuid;
+  DELETE FROM stock_movement_status WHERE depot_uuid =  _old_uuid;
+  DELETE FROM depot_permission WHERE depot_uuid =  _old_uuid;
+  DELETE FROM depot WHERE uuid =  _old_uuid;
+END$$
+
+
```
jmcameron, line 263: I notice that the all of the info in the old depot (text, description, allow_* flags, location_uuid, etc) is being ignored in preference to the new depot info.   I assume this is deliberate.  Is there any reason to keep some of this information from the old depot and merge/replace it in the new depot?

jmcameron (approved):
Not sure how to test this, since there are no code (currently) uses it.
But it looks okay.

server/models/admin.sql
jniles: That's a good question. The idea here is just that you can't delete a depot that already has movements.  We ran into this case in production once by accident, because users refered to the same depot by two different names.

Instead of dealing with trying to merge properties, we'll just keep the "new one" and you can edit it via the regular modules.

No code currently uses this because it's just a handy tool to have in our switchblade when fixing up production databases.  It isn't ever supposed to be exposed to the users.  Most of the "z" stored procedures are that way.

PR #5644, opened 2021-05-11
chore: improve find transfer/purchase modals

jniles:
This PR tightens up the code of both transfer modals for stock entry.  Their behavior should be nearly identical.

Here are the improvements:
 1. Both use the headercrumb properly so their formatting looks like other modules.
 2. Both use the `bh-filter-toggle` directive for toggling the grid's filtering.
 3. Both have a submit button that shows a loading indicator while submitting.
 4. Neither listen on every grid change and update the selectedRow.  Instead, that information is only queried on submit.
 5. Both use receipt components for their receipts instead of importing the ReceiptModal service.

### DIFF
--- a/client/src/modules/stock/entry/modals/findPurchase.modal.js
+++ b/client/src/modules/stock/entry/modals/findPurchase.modal.js
@@ -9,12 +9,9 @@ StockFindPurchaseModalController.$inject = [
 
 function StockFindPurchaseModalController(
   Instance, Purchase, Notify, uiGridConstants, Filtering,
-  bhConstants, Session, Exchange
+  bhConstants, Session, Exchange,
 ) {
   const vm = this;
-
-  // global
-  vm.selectedRow = {};
 
   const {
     CONFIRMED,
@@ -22,7 +19,6 @@ function StockFindPurchaseModalController(
   } = bhConstants.purchaseStatus;
 
   /* ======================= Grid configurations ============================ */
-  vm.filterEnabled = false;
   vm.gridOptions = { appScopeProvider : vm };
 
   const filtering = new Filtering(vm.gridOptions);
@@ -62,7 +58,7 @@ function StockFindPurchaseModalController(
   vm.gridOptions.columnDefs = columns;
   vm.gridOptions.multiSelect = false;
   vm.gridOptions.enableColumnMenus = false;
-  vm.gridOptions.enableFiltering = vm.filterEnabled;
+  vm.gridOptions.enableFiltering = false;
   vm.gridOptions.onRegisterApi = onRegisterApi;
   vm.toggleFilter = toggleFilter;
 
@@ -72,17 +68,11 @@ function StockFindPurchaseModalController(
 
   function onRegisterApi(gridApi) {
     vm.gridApi = gridApi;
-    vm.gridApi.selection.on.rowSelectionChanged(null, rowSelectionCallback);
-  }
-
-  function rowSelectionCallback(row) {
-    vm.selectedRow = row.entity;
   }
 
   /** toggle filter */
   function toggleFilter() {
-    vm.filterEnabled = !vm.filterEnabled;
-    vm.gridOptions.enableFiltering = vm.filterEnabled;
+    vm.gridOptions.enableFiltering = !vm.gridOptions.enableFiltering;
     vm.gridApi.core.notifyDataChange(uiGridConstants.dataChange.COLUMN);
   }
 
@@ -90,9 +80,7 @@ function StockFindPurchaseModalController(
   function load() {
     vm.loading = true;
     Exchange.read()
-      .then(() => {
-        return Purchase.search({ status_id : [CONFIRMED, PARTIALLY_RECEIVED] });
-      })
+      .then(() => Purchase.search({ status_id : [CONFIRMED, PARTIALLY_RECEIVED] }))
       .then(purchases => {
         vm.gridOptions.data = purchases;
       })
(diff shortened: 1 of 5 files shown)


client/src/modules/stock/entry/modals/findPurchase.modal.js
```diff
@@ -104,11 +92,12 @@ function StockFindPurchaseModalController(
       });
   }
 
-  // submit
   function submit() {
-    if (!vm.selectedRow || (vm.selectedRow && !vm.selectedRow.uuid)) { return null; }
+    const [selectedRow] = vm.gridApi.selection.getSelectedRows();
```
jmcameron, line 96: Does this only get the first selected row?  Is that the desired behavior?  Can more than one row be selected?

client/src/modules/stock/entry/modals/findTransfer.modal.html
```diff
@@ -1,14 +1,16 @@
 <form
   name="FindForm"
-  ng-submit="$ctrl.submit(FindForm)"
+  bh-submit="$ctrl.submit(FindForm)"
```
jmcameron, line 3: I noticed this change in a previous release.  What is the difference between ng-submit and bh-submit?

client/src/modules/stock/entry/modals/findTransfer.modal.js
```diff
@@ -35,7 +43,7 @@ function StockFindTransferModalController(
     field : 'document_reference',
     displayName : 'FORM.LABELS.REFERENCE',
     headerCellFilter : 'translate',
-    cellTemplate : 'modules/stock/entry/modals/templates/document_reference.tmpl.html',
+    cellTemplate : documentReferenceTmpl,
```
jmcameron, line 46: I like the idea of inserting these html/angular code snippets inline like this and not using a template file.  Is this a general trend that we should be doing as we have opportunity?

jmcameron (commented):
I tested this and doing Stock entry from Purchase Orders works nicely.  
I was unable to test it using Transfers because I not sure how to do that.

I did notice one small oddity.   To reproduce
- In (fresh) Bhima_test, there are two purchase orders.  
- Edit the status and confirm the second (unconfirmed) one.
- Do stock entry and select the PO you just confirmed
- Enter the lots for second item (pred...) and increase the global number to 400 and do the entry for that number
- Back in the main entry form delete the first item (pretend that a partial delivery was sent).
- Submit the form to complete the entry
- Go back to the Purchase registry and notice the status.  
- It is correct but it hides the fact that the order is incomplete.
image

client/src/modules/stock/entry/modals/findTransfer.modal.html
jniles: `bhSubmit` is what makes the loading magic happens.  They are essentially identical, but if you have a `bhSubmit` sets the variable `$loading` that makes the loading button show a loading indicator.

client/src/modules/stock/entry/modals/findTransfer.modal.js
jniles: I personally am a fan of this method.  

The real reason here is that I wanted to use a component and not that template which calls a callback. It was more straightforward for me to do it inline.

client/src/modules/stock/entry/modals/findTransfer.modal.html
jmcameron: Thanks for the info.  I've wondered why the handling of 'loading' happened (or didn't happen!)

client/src/modules/stock/entry/modals/findTransfer.modal.js
jmcameron: Thanks

client/src/modules/stock/entry/modals/findPurchase.modal.js
jniles: No, only one row can be selected in this grid.  We only let you move in one purchase order or transfer at a time, for simplicity.  That is on [this line](
https://github.com/IMA-WorldHealth/bhima/pull/5644/files#diff-fb0b5281b40977642fa8ae04796a1a12bc2c2941756d8019c5f87f5c865fb175R59).

You are right to check for this.

jmcameron: Thanks for pointing out that setting.   I had already assumed it was only one.  But the function reads "getSelectedRows()" which implies to me that more than one might be selected.

jniles: Yeah, that is the only way to get rows from the ui-grid API.  It's beyond our control.

jniles:
I'll take a look at your scenarios a bit more tomorrow.  We should have migrated to a new location by then.  Thanks for your review!

client/src/modules/stock/entry/modals/findPurchase.modal.js
jmcameron: Okay!

jniles:
@jmcameron finally took a look at this.

You are right, this is confusing.  There are two simultaneous errors - one item wasn't received at all and the other one was received over-stock.  Yet, we only show the second case.

Perhaps, instead of trying to say "received excessively", we should put something more generic like "Received Incoherently" and just link the "Analysis of Purchase Order" report (note though, this this has errors in master - https://github.com/IMA-WorldHealth/bhima/issues/5429).

I'll create an issue for further discussion.

jniles:
Since this got approval and the discussion revolves around purchase orders, I'm merging this.

bors r+

PR #5648, opened 2021-05-12
chore(stock): hover for full exp date on stock exit

### LINKED ISSUE #5637: Stock Exit should show the full expiration date when hovering over the "expires" field
The stock exit module should show the full expiration date of a lot when the user hovers over the "Expires" field.  This is a user request that we could honor.

jniles:
Adds the ability to view the entire expiration date by hovering over the "expires" field.

Closes #5637.

Here is what it looks like:

image

### DIFF
--- a/client/src/modules/stock/exit/exit.js
+++ b/client/src/modules/stock/exit/exit.js
@@ -32,6 +32,7 @@ function StockExitController(
   vm.currentInventories = [];
   vm.reset = reset;
   vm.ROW_ERROR_FLAG = bhConstants.grid.ROW_ERROR_FLAG;
+  vm.DATE_FMT = bhConstants.dates.format;
   vm.overconsumption = [];
 
   vm.onDateChange = date => {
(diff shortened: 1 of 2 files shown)


client/src/modules/stock/exit/exit.js
```diff
@@ -32,6 +32,7 @@ function StockExitController(
   vm.currentInventories = [];
   vm.reset = reset;
   vm.ROW_ERROR_FLAG = bhConstants.grid.ROW_ERROR_FLAG;
+  vm.DATE_FMT = bhConstants.dates.format;
```
jmcameron, line 35: This is fine for now, while we have Bhima installed in only one country.  For the future we should think about allowing the admin to select the desired date format (or use default regional formats).

jmcameron (approved):
Looks good.  I tested it and it works.

client/src/modules/stock/exit/exit.js
jmcameron: > This is fine for now, while we have Bhima installed in only one country. For the future we should think about allowing the admin to select the desired date format (or use default regional formats).

@jniles Would you like me to add an issue for this?

PR #5650, opened 2021-05-12
fix(reports): ensure red background on RUMER lines

### LINKED ISSUE #5638: RUMER report line ending in a negative balance should color the line red
If the RUMER report ends in a negative balance, it should color the line red.  This makes it obvious that there is a problem in that stock item.

jniles:
The table-striped directive was overriding the RUMER's red warning class
for rows ending in negative values.  Thus, the red label would sometimes
not show up.  This commit fixes that issue by creating a custom
background class with !important CSS rules.

Closes #5638.

image

### DIFF
--- a/client/src/i18n/en/tree.json
+++ b/client/src/i18n/en/tree.json
@@ -139,7 +139,7 @@
     "ROLE_MANAGEMENT" : "Role Management",
     "ROOT" : "Root",
     "RUBRIC_MANAGEMENT" : "Rubrics Management",
-    "RUMER_REPORT" : "Rumer",
+    "RUMER_REPORT" : "RUMER",
     "SERVICE" : "Services",
     "SIMPLE_VOUCHER" : "Simple Voucher",
     "STAFFING_INDICES_MANAGEMENT" : "Staffing indices management",
--- a/client/src/i18n/fr/tree.json
+++ b/client/src/i18n/fr/tree.json
@@ -139,7 +139,7 @@
     "ROLE_MANAGEMENT":"Gestion des rôles",
     "ROOT":"Racine",
     "RUBRIC_MANAGEMENT":"Gestion Rubriques",
-    "RUMER_REPORT" : "Rumer",
+    "RUMER_REPORT" : "RUMER",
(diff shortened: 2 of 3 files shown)


jmcameron (commented):
This looks good.

I was not able to really test it.   With Bhima_test I could see the Rumer report but there were to bad entries to red-highlight.   I tried imck but the Rumer report was not an option.  So I'm not sure how to test this.

BTW: A suggestion.  I'm sure RUMER stands for something.  It would be helpful to have a more descriptive name in the main menu for those not familiar with that report.  I'm not necessarily suggesting spelling out the full acronym, but something extra would be good.

jniles:
@jmcameron, the easiest way to test this to is to use a production database (such as IMCK) and just update it to the latest version.  You can do that by:

1. Checking out this branch
2. Setting the `DB_NAME` variable to a production database (such as IMCK) in your .env
3. Run `yarn migrate`.  This generates the `migration-$DATABASE.sql` file.
4. Concatenate in the latest migration file into that one: `cat server/models/migrations/v1.18.0-v1.19.0/migrate >> migration-$DATABASE.sql`.
5. Then `mysql $DATABASE < migration-$DATABASE.sql`
6. And finally, sign in and give your role permission to the RUMER report.

jmcameron:
> @jmcameron, the easiest way to test this to is to use a production database (such as IMCK) and just update it to the latest version. You can do that by:

Excellent! I was able to follow this procedure and it worked.  The red highlighting on the RUMER report looks perfect.

So I'll approve, but I still wonder if the naming needs to be improved to be a little more user-friendly.

NOTE: "Rumer" is not all upper case in the main menu under Stock > Report.  That should probably be fixed.

jmcameron (approved):
It works fine. 

However, "Rumer" is not all upper case in the main menu under Stock > Report. That should probably be fixed.

jniles:
Fixed the RUMER capitalisation.  Thanks for your review @jmcameron .

I'm not sure what to call the RUMER report in english.  It is an acronym, but I'll have to do some research when we get back to nice internet land.

bors r+

PR #5663, opened 2021-05-20
Fix stock variation report

### LINKED ISSUE #5662: Improvements to the "Rapport sur les variations de stock"
Here is a copy of the current "Rapport sur les variations de stock" found on `master` with the Tshikaji Health Zone's dataset as a sample dataset.

image

Some improvements that we can make:
  1. While the page is landscape, the content doesn't reflow to fill the entire page.
  2. The "Number of days of Stock out" doesn't leave much space.  We should move it left a column to ensure we have enough room.
  3. Let's change "items" to "products" at the bottom of the page. 
  4. Let's order this report alphabetically.
  5. Change "Quantity" to "Quantity in Stock"

mbayopanda:
This PR fixes issue #5662 

image

Close #5662

### DIFF
--- a/server/controllers/stock/reports/index.js
+++ b/server/controllers/stock/reports/index.js
@@ -35,7 +35,7 @@ const stockAdjustmentReceipt = require('./stock/adjustment_receipt');
 const stockValue = require('./stock/value');
 const stockAssignReceipt = require('./stock/assign_receipt');
 const stockRequisitionReceipt = require('../requisition/requisition.receipt');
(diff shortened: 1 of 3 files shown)


jniles:
Could you include a screenshot?  I'm not in a position to be able to pull it and look at it.

Also, it looks like it might not be ordered alphabetically still.  Could you look into that?

jniles (approved):
Looks good to me.  Thanks!

Bors r+

PR #5680, opened 2021-05-23
Fix text on requisition depot

### LINKED ISSUE #5641: Clarify the text on the the requisition depot
The "requisition" modal has two fields for the depot.  This is somewhat confusing, since there are two depots to select.  We should distinguish via the label between the depot _serving_ the requisition and the depot _emitting_ the requisition.  We should do this by just changing the text of the label.

One proposal is to rename to "Depot Fourniseur" or "supplying depot" that will fulfill the requisition.

mbayopanda:
Close #5641


jniles (approved):
LGTM.

Bors r+

PR #5686, opened 2021-05-24
Improve aggregate stock consumption

### LINKED ISSUE #5668: Aggregate consumption is confusing
The aggregate consumption of stock module is confusing.  Here are some frequently asked questions:

1. What does the quantity represent?  Is it the quantity at the beginning of the month, or the sum of all distributions during the month?
2. How does one model the days of stock out during the time period?

I propose that we separate the "inital quantity" and display the "quantity delivered during the month".  This will help users understand why you can have consumption during a month that you started without any stock in the depot.

lomamech:
- Displaying quantity at the beginning of the month, and the sum of
  stocks exits and stocks entries

closes #5668

### DIFF
--- a/client/src/modules/stock/aggregated_consumption/aggregated_consumption.js
+++ b/client/src/modules/stock/aggregated_consumption/aggregated_consumption.js
@@ -90,6 +90,7 @@ function StockAggregatedConsumptionController(
   const columns = [
     {
       field : 'text',
+      width : 500,
       displayName : 'TABLE.COLUMNS.DESCRIPTION',
       headerCellFilter : 'translate',
       enableSorting : true,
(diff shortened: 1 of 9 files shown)


lomamech:
about the option to limit the date in relation to the chosen period it may be desirable to improve the bhDatePicker component in another issue

server/controllers/stock/index.js
```diff
@@ -909,6 +910,129 @@ async function listLotsDepot(req, res, next) {
...
+    month_average_consumption : req.session.stock_settings.month_average_consumption,
+    average_consumption_algo : req.session.stock_settings.average_consumption_algo,
+    min_delay : req.session.stock_settings.min_delay,
+    default_purchase_interval : req.session.stock_settings.default_purchase_interval,
+    check_user_id : params.check_user_id,
+  };
+
+  try {
+    const sqlGetMonthlyStockEntry = `
+      SELECT BUID(l.inventory_uuid) AS inventory_uuid, BUID(sm.lot_uuid) AS lot_uuid,
```
jniles, line 951: I'm going to try to combine these SQL queries for efficiency.

client/src/modules/stock/aggregated_consumption/aggregated_consumption.js
```diff
@@ -145,6 +150,11 @@ function StockAggregatedConsumptionController(
       displayName : 'TABLE.COLUMNS.CONSUMPTION',
       headerCellFilter : 'translate',
       cellTemplate : 'modules/stock/aggregated_consumption/templates/lot_aggregate.tmpl.html',
+    }, {
```
jniles, line 159: These look good!  I'm going to put `text-right` on them so they can be summed easily in a user's head.

jniles (commented):
These are good additions.  The UI is a bit cluttered now, but we need these values displayed.  We can make the UX nicer later.

PR #5693, opened 2021-05-25
feat(stock): add currency to stock exit report

### LINKED ISSUE #5348: Add a currency option to the Stock Exit Report
Most enterprises should (ideally) run in the national currency FC.  However, most large expenditures (salaries, stock purchases, etc) are billing in $USD. It would therefore be valuable to see these in $USD.

Here is the stock exit report as currently stands.
image

Let's add an option to show the value in a different currency, such as $USD.  This would use the exchange rate of the day the report was produced to provide an instantaneous conversion, similar to the Balance Sheet.

jniles:
Adds a currency picker to the stock exit report so that users can see the value of stock in whatever currency they desire at the exchange rate of today.

Here is what it looks like:

image

Closes #5348.


jniles:

mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #5695, opened 2021-05-25
fix(depots): filter before limit in depot select

### LINKED ISSUE #5694: "Change depot" filter should filter _all_ depots, but only show 10.
The change depot modal is annoying to use since you almost always have to click "view all" to find the depot you are looking for, even after filtering.  This is because the filter is being applied on the view values, and not on the entire list.  Let's reverse this and make it filter the entire list, but limit the view values after.

jniles:
This commit switches the order of filtering/limiting to ensure we filter before limiting the view.  This brings values into the view as needed instead of the previous behavior where you were forced to click "see more".

Here is the result:

nhdCNQBVCu

Closes #5694.

### DIFF
--- a/client/src/modules/stock/depot-selection.modal.html
+++ b/client/src/modules/stock/depot-selection.modal.html
@@ -35,7 +35,7 @@
     <ul style="margin-bottom: 2px;" class="list-group">
       <li
         class="list-group-item"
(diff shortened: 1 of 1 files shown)


mbayopanda (approved):
Working fine

LGTM 👍🏽 

bors r+

PR #5697, opened 2021-05-25
Fix: stock exit disappearing inventory items

### LINKED ISSUE #5696: Removing an item from stock exit form may disappear it forever
Removing an item from the stock exit form may cause it to disappear forever, or at least until you clear the form and start again.  This occurs when all lots of that item are used and is because the `selectableInventories` is not updated after removal.

Here is a shot of the bug in action.  We should find SRO once, remove it, and add it again.  Instead, it disappears forever.

PrtBvDx70B

jniles:
This commit fixes the "disappearing" inventory items bug by adding the inventory back into the pool if it should be selectable again.

Here is the fix in action:
zlwEMu3SFb

Closes #5696.

### DIFF
--- a/client/src/modules/stock/depot-selection.modal.html
+++ b/client/src/modules/stock/depot-selection.modal.html
@@ -35,7 +35,7 @@
     <ul style="margin-bottom: 2px;" class="list-group">
       <li
         class="list-group-item"
-        ng-repeat="depot in $ctrl.depots  | limitTo: $ctrl.displayLimit | filter: $ctrl.search track by depot.uuid"
+        ng-repeat="depot in $ctrl.depots | filter: $ctrl.search | limitTo: $ctrl.displayLimit track by depot.uuid"
         ng-class="{'active' : $ctrl.depot.uuid === depot.uuid}"
         ng-click="$ctrl.selectDepot(depot.uuid)"
         style="cursor:pointer;">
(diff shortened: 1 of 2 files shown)


client/src/modules/stock/depot-selection.modal.html
```diff
@@ -35,7 +35,7 @@
     <ul style="margin-bottom: 2px;" class="list-group">
       <li
         class="list-group-item"
-        ng-repeat="depot in $ctrl.depots  | limitTo: $ctrl.displayLimit | filter: $ctrl.search track by depot.uuid"
+        ng-repeat="depot in $ctrl.depots | filter: $ctrl.search | limitTo: $ctrl.displayLimit track by depot.uuid"
```
mbayopanda, line 38: 👍🏽

client/src/modules/stock/exit/exit.js
```diff
@@ -235,8 +235,17 @@ function StockExitController(
   // remove item
   function removeItem(item) {
     vm.stockForm.removeItem(item.id);
+
+    // restore the inventory to the selectableInventories list
+    // if there are no more copies.
+    if (item.inventory) {
+      const isInList = vm.selectableInventories.some(row => row.uuid === item.inventory.uuid);
+      if (!isInList) { vm.selectableInventories.push(item.inventory); }
```
mbayopanda, line 243: 👍🏽 Good catch

mbayopanda (approved):
LGTM 

bors r+

PR #5698, opened 2021-05-25
feat(stock): add currency to expiration report

### LINKED ISSUE #5474: Add currency to Expiration Report
We should add the ability to select a currency to render the Expiration Report in.  Here is what it look like currently:
image

To satisfy this issue, we would need to:

1. The `bhCurrencySelect` to the page.
2. Set the enterprise currency by default.
2. Add logic to convert the values on the page into the selected currency at the exchange rate of the day.
3. Add text to the report that mentions the current exchange rate, if values are not in the enterprise currency.

jniles:
Adds the currency to the stock expiration report so that users can see the value of the stock at the current exchange rate.

Closes #5474.

Here is what it looks like:

image

(Normally this should be in FC)

### DIFF
--- a/server/controllers/stock/reports/stock/expiration_report.js
+++ b/server/controllers/stock/reports/stock/expiration_report.js
@@ -4,6 +4,15 @@ const {
 
 const stockCore = require('../../core');
 
+function exchange(rows, exchangeRate) {
+
+  rows.forEach(row => {
+    row.unit_cost *= exchangeRate;
+  });
+
+  return rows;
+}
+
 /**
  * @method stockExpirationReport
  *
@@ -15,6 +24,7 @@ async function stockExpirationReport(req, res, next) {
 
   try {
     const options = { trackingExpiration : 1, includeEmptyLot : 0, ...req.query };
+    const currencyId = parseInt(req.query.currencyId, 10);
 
     const optionReport = _.extend(options, {
       filename : 'REPORT.STOCK_EXPIRATION_REPORT.TITLE',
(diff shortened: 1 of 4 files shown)


mbayopanda:
On the fresh bhima_test database I got this : 
I0JW0lYBow

- The currency doesn't change
- For $ as enterprise currency, the report gives zeros while for other currencies there are not zeros values

I ran : 
- yarn build:db
- yarn dev (I tested the report in the app)

jniles:
Thanks for the review!  Let me take a look at this and make sure it is behaving correctly.

jniles:
This PR had multiple problems and wasn't at all ready.   Sorry!  Now all the problems should be fixed.  Could you give it another look @mbayopanda ?

image

server/controllers/stock/reports/stock/expiration_report.js
```diff
@@ -34,6 +44,15 @@ async function stockExpirationReport(req, res, next) {
...
     }
 
+    const isEnterpriseCurrency = currencyId === req.session.enterprise.currency_id;
+
+    const [{ rate }] = await db.exec(
+      'SELECT GetExchangeRate(?, ?, NOW()) AS rate;',
+      [req.session.enterprise.id, currencyId],
+    );
+
+    const exchangeRate = isEnterpriseCurrency ? 1 : rate;
```
mbayopanda, line 54: 👍🏽

mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #5705, opened 2021-05-26
feat(depots): add loading icon for depot select

### LINKED ISSUE #5704: bhDepotSelect should show loading indicator while searching
The bhDepotSelect looks up depots by text in the database which may take a while over a slow connection.  Here is the code for reference:

https://github.com/IMA-WorldHealth/bhima/blob/688839dbaafb6c0e024650fdc745ec7ce38bdc37/client/src/js/components/bhDepotSelect/bhDepotSelect.tmpl.html#L12-L25

We should add in the same timeglass loading indicator that we used in the hospital number looking to show that the application is looking for the depot:

image

This would lead to a much better user experience.  Currently, there is no feedback when the application is searching and this leads to some confusion on the user's part, particularly over a slow connection.

jniles:
Adds a loading indicator (hourglass) to the bhDepotSelect so that users are aware that the component is loading, even in a slow connection.

Closes #5704.

Here is what it looks like:
3KBGXuPPNa

### DIFF
--- a/client/src/js/components/bhDepotSelect/bhDepotSelect.tmpl.html
+++ b/client/src/js/components/bhDepotSelect/bhDepotSelect.tmpl.html
@@ -1,6 +1,6 @@
 <div ng-form="DepotForm" bh-depot-select ng-model-options="{ updateOn: 'default' }">
   <div
-    class="form-group"
+    class="form-group has-feedback"
     ng-class="{ 'has-error' : DepotForm.$submitted && DepotForm.depot_uuid.$invalid }">
 
     <label class="control-label" translate>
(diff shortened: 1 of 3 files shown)


jniles:

client/src/js/components/bhDepotSelect/bhDepotSelect.tmpl.html
```diff
@@ -1,6 +1,6 @@
 <div ng-form="DepotForm" bh-depot-select ng-model-options="{ updateOn: 'default' }">
   <div
-    class="form-group"
+    class="form-group has-feedback"
```
jmcameron, line 3: Is this necessary?  it seems like we have some forms that use the $loading flag that do not use it.

jmcameron (commented):
The code looks good.

I was not able to test it thoroughly because the databases I used had a small enough number of depots the look was essentially instant and I never saw the hour-glass.

client/src/js/components/bhDepotSelect/bhDepotSelect.tmpl.html
```diff
@@ -16,14 +16,16 @@
...
-      typeahead-loading="$ctrl.loading"
+      typeahead-loading="$ctrl.$loading"
       typeahead-template-url="/modules/templates/depotList.tmpl.html"
       typeahead-on-select="$ctrl.onSelect($item, $model)"
       typeahead-min-length="0"
       autocomplete="off"
       translate-attr="{ 'placeholder': 'FORM.SELECT.DEPOT' }"
       ng-required="$ctrl.required">
 
+    <span ng-show="$ctrl.$loading" class="glyphicon glyphicon-hourglass form-control-feedback"></span>
```
mbayopanda, line 27: `glyphicon` 😄 ça fait très longtemps. 👍🏽

mbayopanda (approved):
image

LGTM 👍🏽

client/src/js/components/bhDepotSelect/bhDepotSelect.tmpl.html
jniles: Yes, to show the icon on the input, the `has-feedback` class does the magic.

PR #5707, opened 2021-05-26
Add tags for inventories

mbayopanda:
This PR adds : 
- A column for tags in the inventory registry

image

- Tags selector in the inventory action (create and edit) modal : 

image

- Tags selector in the inventory search modal : 

image

- Tags selector in the stock inventory search modal
- A column for tags in the stock inventory registry

### DIFF
--- a/server/controllers/inventory/inventory/core.js
+++ b/server/controllers/inventory/inventory/core.js
@@ -273,7 +273,7 @@ async function getItemsMetadata(params) {
   filters.equals('sellable');
   filters.equals('note');
   filters.equals('importance');
-  filters.custom('tags', 't.uuid IN (?)', params.tags);
+  filters.custom('tags', 't.uuid IN (?)', [params.tags]);
   filters.custom('find_null_importance', 'inventory.importance IS NULL');
   filters.custom('inventory_uuids', 'inventory.uuid IN (?)', params.inventory_uuids);
   filters.setOrder('ORDER BY inventory.code ASC');
--- a/server/controllers/inventory/inventory/core.js
+++ b/server/controllers/inventory/inventory/core.js
@@ -72,9 +72,12 @@ exports.inventoryColsMap = inventoryColsMap;
 * @return {Promise} Returns a database query promise
 */
 async function createItemsMetadata(record, session) {
+  let tags;
+
   const recordCopy = _.clone(record);
-  record.enterprise_id = session.enterprise.id;
   const recordUuid = record.uuid || uuid();
+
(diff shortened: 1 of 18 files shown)


server/controllers/inventory/inventory/core.js
```diff
@@ -84,8 +87,21 @@ async function createItemsMetadata(record, session) {
...
+  if (record.tags) {
+    tags = _.clone(record.tags);
+    delete record.tags;
+  }
+
   transaction.addQuery(sql, [record]);
 
+  if (tags && tags.length) {
+    tags.forEach(t => {
+      const binaryTagUuid = db.bid(t.uuid);
```
jniles, line 99: Nice variable naming 👍

server/controllers/stock/core.js
```diff
@@ -124,7 +124,7 @@ function getLotFilters(parameters) {
     'd.uuid IN (SELECT depot_uuid FROM depot_permission WHERE user_id = ?)',
   );
 
-  // tags
+  // lot tags
```
jniles, line 127: Good addition 👍

test/end-to-end/shared/components/bhTagSelect.js
```diff
@@ -4,6 +4,16 @@ module.exports = {
...
     const root = $(this.selector);
-    await FU.typeaheadAppended('$ctrl.$ctrl.tagUuids', tags, root);
+
+    if (Array.isArray(tags)) {
+      const stack = [];
+      for (let index = 0; index < tags.length; index++) {
+        const tag = tags[index];
+        stack.push(FU.uiSelect('$ctrl.tagUuids', tag, root));
+      }
+      await Promise.all(stack);
```
jniles: Speaking from experience, you should call `await` inside the for loop and remove `Promise.all`.  What you are doing here is an excellent performance increase by making the clicks _parallel_, but will cause random failures in the test since a user cannot actually select things in parallel.  We had all sorts of random crashes with the bhAccountSelect because of this.

Another way to think about it: the E2E test should emulate a user's behavior.  They are not going to click all the tags at once - they will go one tag at a time and click it.  Removing `Promise.all()` and using `await` in the loop will better simulate that experience.

jniles (commented):
@mbayopanda, this code looks really good.  Really clean, very nice variable names, and well laid out.  Great job!

My only comment that needs to be addressed is the `Promise.all()` call in the test component which will cause random test crashes down the line.

mbayopanda:
@jniles changes landed

jniles (approved):
LGTM.

bors r+

Issue #5699, opened 2021-05-26
Bhima crashes during depots loading in the user permission


mbayopanda:
![image](https://user-images.githubusercontent.com/5445251/119583410-7b991d00-bdbe-11eb-92b8-2fd33fe313c5.png)

Administration -> Users -> Depots Permission

jniles:
That looks like the pepkits database.  Is this bug on master or on the pepkits version of BHIMA?

Also, how many depots do we have in the pepkits database?

mbayopanda:
> That looks like the pepkits database. Is this bug on master or on the pepkits version of BHIMA?
> 
> Also, how many depots do we have in the pepkits database?

![image](https://user-images.githubusercontent.com/5445251/119682026-7679b380-be3a-11eb-9359-2cb426b9656e.png)

10815 depots

jniles:
Oh my goodness.  Well, I wonder what the solution to this is.  We can try to improve the performance of the component, but I'm not sure showing 10,000+ checkboxes is a good idea.

jniles:
I'm going to close this as a "won't fix".  I think the problem is likely that we shouldn't try to load in 10K+ depots.  If we wanted to, we should change this interface to be something else.

PR #5710, opened 2021-05-27
fix(depots): improve performance on permissions modal

jniles:
Marginally improves the performance of the depot permission modal by using track by and one-time bindings, as a best practice.  I tried measuring the result with my browser recording, but ultimately, I'm not sure this makes a huge difference in the initial paint (though it should during selection).  Ultimately, we probably should use a ng-repeat for this and either use a custom directive with better performance or change the way we do permissions.

Addesses #5699.

### DIFF
--- a/client/src/modules/users/UserDepotManagementModal.html
+++ b/client/src/modules/users/UserDepotManagementModal.html
@@ -13,9 +13,7 @@
 
     <div class="form-group" ng-class="{ 'has-error' : UserForm.$submitted && UserForm.display_name.$invalid }">
       <label class="control-label" translate>FORM.LABELS.USERNAME</label>
-      <div>
-        <h4><i> {{ UsersDepotModalCtrl.user.display_name }} </i></h4>
-      </div>
+      <p class="form-control-static">{{ UsersDepotModalCtrl.user.display_name }}<p>
     </div>
 
     <span translate> FORM.INFO.USER_DEPOT </span>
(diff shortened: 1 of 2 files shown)


client/src/modules/users/UserDepotManagementModal.html
```diff
@@ -36,12 +34,12 @@ <h4><i> {{ UsersDepotModalCtrl.user.display_name }} </i></h4>
       <input ng-model="UsersDepotModalCtrl.filter" ng-show="UsersDepotModalCtrl.filterActive" class="form-control" placeholder="{{'FORM.PLACEHOLDERS.FILTER_NAME' | translate }}" style="border-radius: 0"/>
     </div>
 
-    <div ng-if="!UsersDepotModalCtrl.loading" ng-repeat="depot in UsersDepotModalCtrl.depotsData | filter : { text : UsersDepotModalCtrl.filter }">
-      <div style="margin-left: calc({{depot.$$treeLevel}} * 15px)" class="list-unstyled">
+    <div ng-if="!UsersDepotModalCtrl.loading" ng-repeat="depot in UsersDepotModalCtrl.depotsData | filter : { text : UsersDepotModalCtrl.filter } track by depot.uuid">
```
mbayopanda, line 37: 👍🏽

mbayopanda (commented):
The issue is still there for a huge number of depots (10000), we can see the result after the crash message if you decide to wait.

image

jniles:
Okay, then we just can't use checkboxes.  I'm not sure that we should support this use case.

PR #5714, opened 2021-05-29
improvement(Depot)

### LINKED ISSUE #5713: Improved display of distribution deposits for a deposit
The display of depots of distributions for a depot is very ambiguous for depots which must make distributions to several depots this ambiguity. This issue proposes to list the deposits for a better display

image

lomamech:
- Improve display in Allowed Destination Depots

closes #5713

### DIFF
--- a/client/src/modules/depots/modals/depot.modal.html
+++ b/client/src/modules/depots/modals/depot.modal.html
@@ -153,25 +153,26 @@
 
     <div ng-if="DepotModalCtrl.enable_strict_depot_distribution">
       <label class="control-label" translate>DEPOT.ALLOWED_DESTINATION_DEPOTS</label>
-
       <div class="row">
-        <div class="col-md-6">
-          <div ng-repeat ="r in DepotModalCtrl.depotRight">
-            <div class="checkbox">
-              <label>
-                <input type="checkbox" ng-model="r.checked" ng-true-value="1" ng-false-value="0">
-                <span translate> {{ r.text }} </span>
-              </label>
+        <div ng-repeat ="d in DepotModalCtrl.depots">
+          <div class="col-md-6">
+            <div ng-if="$index %2 == 0">
+              <div class="checkbox">
+                <label>
+                  <input type="checkbox" ng-model="d.checked" ng-true-value="1" ng-false-value="0">
+                  <span translate> {{ d.text }} </span>
+                </label>
+              </div>
             </div>
           </div>
-        </div>
-        <div class="col-md-6">
-          <div ng-repeat ="l in DepotModalCtrl.depotLeft">
-            <div class="checkbox">
-              <label>
-                <input type="checkbox" ng-model="l.checked"  ng-true-value="1" ng-false-value="0">
-                <span translate> {{ l.text }} </span>
-              </label>
+          <div class="col-md-6">
+            <div ng-if="$index %2 != 0">
+              <div class="checkbox">
+                <label>
+                  <input type="checkbox" ng-model="d.checked" ng-true-value="1" ng-false-value="0">
+                  <span translate> {{ d.text }} </span>
+                </label>
+              </div>
             </div>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


lomamech:
image

client/src/modules/depots/modals/depot.modal.js
```diff
@@ -22,10 +25,43 @@ function DepotModalController($state, Depots, Notify, Session, params) {
     vm.depot.parent_uuid = params.parentUuid;
   }
 
+  Depots.read()
+    .then(depots => {
+      vm.depots = depots;
+
+      let index = 0;
```
jniles: FYI, you don't need to do this extra index.  The second argument of `Array.forEach()` is an `index` like you made.  So you can do this:

```js
depots.forEach((depot, index) => {

// everything else 

});
```

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -153,20 +153,37 @@
...
+        <div class="col-md-6">
+          <div ng-repeat ="l in DepotModalCtrl.depotLeft">
+            <div class="checkbox">
+              <label>
+                <input type="checkbox" ng-model="l.checked"  ng-true-value="1" ng-false-value="0">
+                <span translate> {{ l.text }} </span>
+              </label>
+            </div>
+          </div>
+        </div>
```
mbayopanda, line 178: If depotLeft and depotRight are just for displaying depots left and right, you could use just one loop such as : 
```html
<div ng-repeat ="r in DepotModalCtrl.depots" class="col-md-6">
  <div class="checkbox">
    <label>
      <input type="checkbox" ng-model="r.checked" ng-true-value="1" ng-false-value="0">
      <span translate> {{ r.text }} </span>
    </label>
  </div>
</div>
```

mbayopanda (commented):
Using checkbox and grid for allowed distribution depot is a good idea 👍🏽 it gives to the user much space to interact and it is more readable.

I just made a comment about the use of `depotLeft` and `depotRight`.

Is it still necessary to have `depotLeft` and `depotRight` while we use just one loop with `<div ng-repeat ="r in DepotModalCtrl.depots" class="col-md-6">`

client/src/modules/depots/modals/depot.modal.html
lomamech: image

I tried with this option, the display stays on one column, my idea on concerning the two-column display was just not to have a very long list

jmcameron: Take a look at the answer to this:  https://stackoverflow.com/questions/25655207/angularjs-how-to-dynamically-split-list-in-multiple-columns

lomamech:
@mbayopanda can you check again

lomamech:
> @mbayopanda can you check again

image
@jmcameron  I found another way to display on 2 columns
image

jmcameron:
> @jmcameron I found another way to display on 2 columns

Your approach looks pretty robust and automatic.  Nice!

client/src/modules/depots/modals/depot.modal.html
```diff
@@ -153,20 +153,38 @@
...
+      </div>
     </div>
 
     <div class="form-group" ng-class="{'has-error' : DepotForm.description.$invalid && DepotForm.$submitted}">
       <label class="control-label" translate>FORM.LABELS.DESCRIPTION</label>
       <textarea
         class="form-control"
         name="description"
-        ng-model="DepotModalCtrl.depot.description"
+        ng-model=" DepotModalCtrl.depot.description"
```
jmcameron: Looks like an extra leading space got into this string

jmcameron:
@lomamech How do I get to this display?

jniles:
@jmcameron you'll need to enable restricting distribution depots.  I think it is in stock settings, but it might be in the depot settings.

jmcameron (approved):
The update looks good and it works fine.   

Minor suggestion: since the depots will be listed in two columns, perhaps the create depot modal should be a bit wider?

jniles (approved):
LGTM.  bors r+

PR #5717, opened 2021-05-29
fix(stock): properly exchange stock value

jniles:
The stock value report was updated to include exchanged values so that
the report could be viewed in $USD at Vanga.  However, the exchange rate
was incorrectly applied, causing bugs in the Weighted Average Cost
algorithm.  To fix this, we apply only the rate to the `unit_cost` and
carry that forward via sums.


mbayopanda (approved):
The feature works as expected but it requires defining all currencies rate (Enterprise currency, Euro, Dollars, CDF), if one currency is omitted the report will be weird for this currency.

LGTM 👍🏽 

bors r+

jniles:
Thanks @mbayopanda !

bors r+

PR #5720, opened 2021-06-01
fix(purchase): wait for inventory before edit

### LINKED ISSUE #5716: Editing purchase orders does not save the modifed price
When we modify a purchase order, the old unit price that we modified does not get restored.

jniles:
Ensures the module waits for the inventory to load the prices before merging in the purchase order's additions.  The prices are now always correctly restored by avoiding the race condition.

Closes #5716.

### DIFF
--- a/client/src/modules/purchases/create/PurchaseOrderForm.js
+++ b/client/src/modules/purchases/create/PurchaseOrderForm.js
@@ -91,26 +91,32 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
 
     // TODO(@jniles) - what do we do about user_id?  Should we
     // modify it?  Keep a table of modifications to purchase orders?
+    // NOTE(@jniles) - Vanga would like to see who modified the purchase orders.
 
-    // add the expected number of items
-    order.items.forEach((prev) => {
-      // make the item uuids match between the previous and current records
-      const current = this.addItem();
-      current.uuid = prev.uuid;
+    // make sure we have our inventory loaded before applying the old order prices.
+    return this.ready().then(() => {
 
-      // get the inventory item to configure this row
-      const inventory = this.inventory.available.get(prev.inventory_uuid);
-      current.configure(inventory);
+      // add the expected number of items
+      order.items.forEach((prev) => {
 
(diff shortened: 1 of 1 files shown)


client/src/modules/purchases/create/PurchaseOrderForm.js
```diff
@@ -91,26 +91,32 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
 
     // TODO(@jniles) - what do we do about user_id?  Should we
     // modify it?  Keep a table of modifications to purchase orders?
+    // NOTE(@jniles) - Vanga would like to see who modified the purchase orders.
```
jmcameron, line 94: Is there an issue for this?

jniles: No, I didn't have internet at the time of writing.  Now I do and here is the issue:  https://github.com/IMA-WorldHealth/bhima/issues/5721

client/src/modules/purchases/create/PurchaseOrderForm.js
```diff
@@ -91,26 +91,32 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
...
 
-    // add the expected number of items
-    order.items.forEach((prev) => {
-      // make the item uuids match between the previous and current records
-      const current = this.addItem();
-      current.uuid = prev.uuid;
+    // make sure we have our inventory loaded before applying the old order prices.
+    return this.ready().then(() => {
```
jmcameron, line 97: This code looks good, but I wonder what happens to some of the other purchase order fields such as:
- project_id
- reference
- created_at
- update_at
- user_id
- edited
- status_id
(I may be missing where these are being updated)

jniles: Good question.  Most of these never change, or are changed in triggers in the database.   Not one of those columns is meant to be changed in this particular editing interface that the `PurchaseOrderForm` powers (createUpdate.html).

jmcameron:
@jniles I think this is ready now.  It should not interfere with PR to fix PO editing, so it should be okay to merge this one whenever desired.

PR #5722, opened 2021-06-01
feat(reports): add currency to stock sheet

### LINKED ISSUE #5702: Fiche de Stock should allow users to pick the currency
The fiche de stock should allow users to pick what currency they want to view the report in.  This is in keeping with the recent themes of showing the stock value in different values (#5474).

jniles:
Adds a currency selection to the stock sheet so that we can view the valuation in multiple currencies.

Closes #5702.

Here is what it looks like:
image

### DIFF
--- a/client/src/modules/reports/generate/stock_sheet/stock_sheet.config.js
+++ b/client/src/modules/reports/generate/stock_sheet/stock_sheet.config.js
@@ -3,10 +3,12 @@ angular.module('bhima.controllers')
 
 StockSheetConfigController.$inject = [
   '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state',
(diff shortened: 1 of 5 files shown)


jmcameron (approved):
Code looks good (other than a minor issue I'll point out below). 
It works fine.

One minor glitch is that currency does not default to anything.  The problem is the handlebars uses vm.currency_id but the controller uses vm.reportDetails.currencyId.  Therefore when you go to the page to create the report, no currency is selected.  If you then select one, it works fine.   But it should default to something--but does not.

jniles:
@jmcameron, this has been fixed.  Thanks for that catch!

jniles:
I got approval last time and fixed the bug.  Merging.

bors r+

PR #5723, opened 2021-06-01
Search requisitions by user

### LINKED ISSUE #5692: Add search by user to requisition registry
Another search filter the requisitions registry needs is a filter by user.  That way, each user can easily find which requisitions they made, if multiple users are making requisitions.

jmcameron:
Adds "User" to the search modal to limit listings of requisitions to the specified user.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5692

**TESTING**
- With bhima_test (with other DBs you may not need to hack the reqs)
- Modify one of the Purchase orders to be by a different user
  - In Mysql:
  - select uuid from stock_requisition where reference=3;
  - update stock_requisition set user_id = 2 where uuid=<UUID from previous query>;
- Run bhima,  stock > Requisition
- You should see that Req SREQ.TPA.1 was done by "Regular User"
- Try
   - Super User => 3 reqs
   - Regular User => 1 req
   - Any other user => 0 req

### DIFF
--- a/client/src/modules/purchases/registry/registry.js
+++ b/client/src/modules/purchases/registry/registry.js
@@ -111,7 +111,6 @@ function PurchaseRegistryController(
     fastWatch : true,
     onRegisterApi : (api) => { vm.gridApi = api; },
     columnDefs,
-
   };
 
   const columnConfig = new Columns(vm.uiGridOptions, cacheKey);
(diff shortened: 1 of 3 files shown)


jniles (approved):
This is perfect.  Tested and working as expected with the BCZ database.

bors r+

PR #5727, opened 2021-06-04
Implement aggregate stock consumption per patient report

### LINKED ISSUE #5448: Report for Aggregated Stock Consumption by Patients
I was contacted by a doctor at Vanga to produce the following report:

image

Basically, they want to know the average quantity of stock distributed per distribution and per patient.  This is a rather useful report since it shows if they are following best practices for medicine distributions.

We should implement this report.   It would take:

1. Start date
2. End date
3. inventory group filter

 The full excel is attached.

[HEV Medicament Consumations par malades - 2020.xlsx](https://github.com/IMA-WorldHealth/bhima/files/6068830/HEV.Medicament.Consumations.par.malades.-.2020.xlsx)

jmcameron:
This report shows the aggregate stock consumption to patients and services.

Note:  Columns 3-7 are combined data for distributions to patients or services.   The last 4 columns are for distributions for patients.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5448

It is in the Stock > Reports menu

**TESTING**
- Try it out a date range (without inventory group)
   - Notice the the right 4 columns are blank for "To Service" items.
- Try it out with an inventory group (such as 'Medicaments en comprimes" in IMCK db)
   - Notice new subtitle for this inventory group
   - Notice the change in numbers of results

To verify the tabulated results, pick an
 inventory code from the results above, then
- Run a stock sheet for that inventory item.  You may want to limit the time range to get a smaller set of data to compare.
- Run a Aggregated Stock Report for the same period
- You should be able to correlate the "To Service" and "To Patient" items with the data from the aggregated stock report to validate the new report

This SQL query might be useful:

[check_agg.sql.txt](https://github.com/IMA-WorldHealth/bhima/files/6598052/check_agg.sql.txt)


server/controllers/stock/reports/stock/aggregated_consumption_report.js
```diff
@@ -0,0 +1,115 @@
...
+} = require('../common');
+
+/**
+ * @method stockAggregatedConsumptionReport
+ *
+ * @description
+ * This method builds the aggregated stock consumption report as either a
+ * JSON, PDF, or HTML file to be sent to the client.
+ *
+ * GET /reports/stock/aggreated_consumption
```
jniles: Typo here - aggregated_consumption

server/controllers/stock/reports/stock/aggregated_consumption_report.js
```diff
@@ -0,0 +1,115 @@
...
+
+  const report = new ReportManager(STOCK_AGGREGATED_CONSUMPTION_REPORT_TEMPLATE, session, optionReport);
+
+  const options = (typeof (_options.params) === 'string') ? JSON.parse(_options.params) : _options.params;
+  data.dateFrom = options.dateFrom;
+  data.dateTo = options.dateTo;
+  data.inventoryGroupUuid = options.inventoryGroupUuid;
+  data.inventoryGroupName = null;
+  if (data.inventoryGroupUuid) {
+    const gnameQuery = `SELECT name FROM inventory_group WHERE uuid = 0x${data.inventoryGroupUuid}`;
```
jniles: These don't have any SQL injection escapes.  Could you use the `db` library to do the UUID conversion?  Like this:

```js
if (data.inventoryGroupUuid) {
  data.inventoryGroupName = await db.one('SELECT name FROM inventory_group WHERE uuid = ?', [data.inventoryGroupUuid]);
}
```

If you could update the others as well, that would be perfect.  For integration/unit tests, we can use the `0x${uuid}` shortcut, but for the live server we really need to ensure we are sanitizing our inputs.

jniles:
So, these numbers don't seem to match the fiche de stock numbers.  I'll try to dig to the heart of it to figure it out using Vanga's dataset - the reason wasn't evident.

jmcameron:
@jniles I think I've addressed your comments.  The numbers look okay to me.

jmcameron:
@lomamech Please look this PR over.   @jniles asked for this report (check the Issue https://github.com/IMA-WorldHealth/bhima/issues/5448).  I've worked with him to make sure the report is what he wants.  Please look the code over and suggest improvements.   I think the code is a bit convoluted.  It could probably be done more cleanly with a more sophisticated SQL query, but I'm pretty sure this works correctly as it is.   I'm not sure it makes sense trying to make it any more efficient since it is already fast enough (with IMCK).

jmcameron:
@jniles and @lomamech: With this PR we have two similarly named items:  
   Stock > Aggregated stock consumption
   Stock > Report > Aggregated stock consumption

I'm not quite sure what the first one does, but we probably need to rename one or the other to make the difference clear.  I'm open to suggestions.

lomamech (commented):
I just analyzed the branch,
- I would like to propose to rename the report, Condensed report of the distributions of products to Patients and Services
- The impression I have is that the following report gives additional information compared to the stock exit
image

lomamech:
@jmcameron  I just finished my review

jmcameron:
> @jmcameron I just finished my review

Thanks @chris!

lomamech:
> > @jmcameron I just finished my review
> 
> Thanks @chris!

@jmcameron Did you find another way to rename this report?

jmcameron:
I'm thinking about what to name it.

jmcameron:
> @jmcameron Did you find another way to rename this report?

I think I will leave it with the current title and wait until @jniles has a chance to look at it.  He interacted with the doctor who requested the report and hopefully he will be able to suggest a name for the report that will make sense to the doctor.

lomamech:
> > @jmcameron Did you find another way to rename this report?
> 
> I think I will leave it with the current title and wait until @jniles has a chance to look at it. He interacted with the doctor who requested the report and hopefully he will be able to suggest a name for the report that will make sense to the doctor.

I'll go ahead, and merge it,a label can be changed to any time

PR #5728, opened 2021-06-04
Journal History Module

### LINKED ISSUE #5712: Save all details of a transaction when it is deleted
After talking to AG Neron, he would like a report showing everything that has been deleted.  Currently, we do not save much information about the deletion and it is not accessible to an auditor - only a programmer.  At this point, BHIMA is large enough and active enough that he needs this report.  We are missing two things:

-  report that he can pull to find this information and 
- the full transaction information that was removed.

With these, he will feel comfortable allowing deletion of records.

mbayopanda:
This PR adds a module for tracking editions and deletions in the posting journal

image

Close #5712


jmcameron:
I took a casual look at this.   I wonder if the info tags (English) could be improved?   It seems like you want past-tense items:
  EDITED
  DELETED
  UPDATED
etc.

client/src/modules/journal/journal_log.html
```diff
@@ -0,0 +1,59 @@
...
+    on-remove-filter="JournalLogCtrl.onRemoveFilter(filter)">
+  </bh-filters>
+</div>
+
+<div class="flex-content">
+  <div class="container-fluid">
+    <div class="row">
+      <div class="col-md-12">
+
+        <table class="table-condensed">
```
jniles: Can you add the class "table" here?  This will make it take up the full screen.

jniles (approved):
@mbayopanda, this is excellent.  It's a really, really good pull request.

The only comment I have is to add the table class to make the table take up the full screen.  Otherwise, we can merge this.  Feel free to do so when you make that change.

jniles:
Note, I didn't test it in English - you should probably follow @jmcameron's advice as well while making the table update.

mbayopanda:
@jmcameron @jniles changes landed

jniles (approved):
LGTM.

bors r+

PR #5737, opened 2021-06-08
Fix editing of purchase orders.

### LINKED ISSUE #5736: Modifying a purchase order doesn't create new purchase_item
Our feature to edit purchase orders doesn't seem to create new purchase_items.  The vanga team attempted to add purchase items later to the order and this operation did not work.  We should also ensure that deleting purchase order items works as expected and removes the individual line items.

jmcameron:
This PR fixes issues with editing purchase orders.  In particular items added while editing are handled correctly.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5736

**TESTING**
- The following functions should be tested
- Creating a new purchase order
- Editing an existing purchase order
   - Deleting an existing item
   - Changing the quantity of an existing item
   - Adding a new item
   - Entering a duplicate of an existing item with the same inventory item, the form will complain.  Please double-check the French translation of the new error message.

### DIFF
--- a/client/src/modules/purchases/create/createUpdate.html
+++ b/client/src/modules/purchases/create/createUpdate.html
@@ -210,10 +210,19 @@
               type="button"
               class="btn btn-default"
               ng-click="PurchaseCtrl.clear(PurchaseOrderForm)"
-              data-method="clear" translate>
+              data-method="clear"
+              ng-show="!PurchaseCtrl.isUpdateState"
+              translate>
               FORM.BUTTONS.CLEAR
             </button>
-            <p
+            <button
+              type="button"
+              class="btn btn-default"
+              ng-click="PurchaseCtrl.cancel()"
+              data-method="cancel">
+              <span ng-show="PurchaseCtrl.isUpdateState" translate>FORM.BUTTONS.CANCEL_EDIT</span>
+              <span ng-show="!PurchaseCtrl.isUpdateState" translate>FORM.BUTTONS.CANCEL</span>
+            </button>            <p
               class="text-danger"
               ng-show="PurchaseCtrl.order._invalid && PurchaseOrderForm.$submitted">
               <span class="fa fa-exclamation-circle"></span>
--- a/client/src/i18n/en/errors.json
+++ b/client/src/i18n/en/errors.json
@@ -20,6 +20,7 @@
     "OVERPAID_INVOICE" : "You have overpaid the invoices.",
     "ER_DUP_KEY" : "A key collided in a unique database field.  Please retry your action.  If the problem persists, contact the developers.",
     "ER_DUP_ENTRY" : "You have duplicated a value in a unique field!  Please make sure that the necessary values are unique.",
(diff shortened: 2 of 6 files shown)


jmcameron:
@jniles Looks like this is ready for review.

jmcameron:
Note: This does not have the fix in https://github.com/IMA-WorldHealth/bhima/pull/5720.  But this pull and 5270 do not affect the same files, so there is no need to combine them.

client/src/modules/purchases/create/createUpdate.html
```diff
@@ -213,7 +213,24 @@ <h4 class="text-right" style="margin-right: 25px;">
...
+              ng-show="PurchaseCtrl.isUpdateState"
+              translate>
+              FORM.BUTTONS.CANCEL_EDIT
+            </button>
+            <button
+              type="button"
+              class="btn btn-default"
+              ng-click="PurchaseCtrl.cancel()"
+              data-method="cancel"
+              ng-show="!PurchaseCtrl.isUpdateState"
```
jniles: These look identical, except for the `isUpdateState` call.  That means that you probably just need to change the text.  I propose you do something like:
```html
<button
  type="button"
  class="btn btn-default"
  ng-click="PurchaseCtrl.cancel()"
  data-method="cancel">
  <span ng-show="PurchaseCtrl.isUpdateState">FORM.BUTTONS.CANCEL_EDIT</span>
  <span ng-hide="PurchaseCtrl.isUpdateState">FORM.BUTTONS.CANCEL</span>
</button>
```

Do we really need both "clear" and "cancel"?  I don't think we need both.  Let's remove the other button and just put this button.

image

jniles (commented):
Functionality works perfectly.  I've made a suggestion about the number of buttons.  Do you really think it makes sense to have all of them?

client/src/modules/purchases/create/createUpdate.html
jmcameron: I Think the [Clear] button is only there when creating a PO, not when editing.   That seems right to me.  It maintains the [Clear] functionality that was already there (when creating) but removes it for editing, since that does not make as much sense to me.   If they want to remove everything while editing, they can do that-- but why not delete the PO?

I think the [Cancel] / [Cancel Edit] button are really useful since it is not always obvious what to do if you decide to abort creating or editing a PO.,

Regarding the duplicate labels--I'll fix that tomorrow.   I considered doing that but never got to it.

jniles: This screenshot was taken during an edit - I think that the three buttons do show up when there is an edit.

I agree with the logic you described.

jmcameron: @jniles I've fixed the buttons to work like I wanted it to (see above) and incorporated the code cleanup you suggested.  I think it is ready to go now.   BTW, the reason I wanted the "cancel" button to read [Cancel Edit] when editing is to avoid confusion that that button would cancel the PO itself.

jniles:
Looks good to me.

bors r+

PR #5745, opened 2021-06-10
Feature edit requisitions

### LINKED ISSUE #5660: Allow users to edit requisitions
We should allow users to edit requisitions that are not yet delivered.  Requisitions seem to be an excellent way to allow users to experiment with creating orders to be filled later.  However, if the user makes a mistake, they need to cancel the requisition entirely and begin again.  This isn't a nice user experience.

My understanding of requisitions is that they do not have any dependencies - you can create one at any time, request from any depot, no matter how much stock is or isn't available in the depot(s).  If that is true, we should be able to easily allow users to edit requisitions with minimal risk.

lomamech:
- Allow users to edit requisitions
- Add E2E test for Update requisition

closes #5660

### DIFF
--- a/client/src/js/components/bhServiceOrDepot/bhServiceOrDepot.html
+++ b/client/src/js/components/bhServiceOrDepot/bhServiceOrDepot.html
@@ -31,6 +31,7 @@
   <!-- service  -->
   <div ng-if="$ctrl.requestorType.type_key === 'service'">
     <bh-service-select
+      service-uuid="$ctrl.uuid"
       label="REQUISITION.SERVICE_REQUESTOR"
       on-select-callback="$ctrl.onSelectRequestor(service)"
       required="true">
--- a/server/controllers/stock/requisition/requisition.js
+++ b/server/controllers/stock/requisition/requisition.js
@@ -198,6 +198,7 @@ exports.update = async (req, res, next) => {
     const transaction = db.transaction();
     const uuid = db.bid(req.params.uuid);
     const requisition = _.omit(req.body, 'items');
+
     const requisitionItems = _.pick(req.body, 'items').items;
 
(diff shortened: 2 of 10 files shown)


mbayopanda (commented):
The feature works well with requisition for depot, but I just noticed that for service, the concerned service field is not defined

image

lomamech:
> The feature works well with requisition for depot, but I just noticed that for service, the concerned service field is not defined
> 
> image

lomamech:
@mbayopanda it's done

server/controllers/stock/requisition/requisition.js
```diff
@@ -280,7 +285,7 @@ exports.update = async (req, res, next) => {
 
     if (requisitionItems && requisitionItems.length) {
       transaction
-        .addQuery('DELETE FROM stock_requisition_item WHERE requisition_uuid = ?;', [binarize(requisition), uuid]);
+        .addQuery('DELETE FROM stock_requisition_item WHERE requisition_uuid = ?;', [uuid]);
```
mbayopanda, line 288: 👍🏽

client/src/js/components/bhServiceOrDepot/bhServiceOrDepot.html
```diff
@@ -31,6 +31,7 @@
   <!-- service  -->
   <div ng-if="$ctrl.requestorType.type_key === 'service'">
     <bh-service-select
+      service-uuid="$ctrl.uuid"
```
mbayopanda, line 34: 👍🏽

mbayopanda (approved):
LGTM 👍🏽

jniles:
LGTM as well.

bors r+

PR #5748, opened 2021-06-11
Show voucher information on the stock receipt

### LINKED ISSUE #5510: If auto_stock_accounting is enabled, show voucher on stock receipt
We have a patchwork of voucher labels shown on the stock receipts.  For example, the stock adjustment receipt shows the voucher, but not the distribution to services.

We should be checking the `auto_stock_accounting` flag in stock settings before attempting to display the voucher on any receipt.  If it is enabled, we should show the voucher on every receipt.  This will help us in our analysis of stock versus finances.

mbayopanda:
This PR is related to #5510 and adds information about voucher on the stock receipt.

If the movement was made before the setup of stock accounting, and then the stock accounting has been activated we have now the message : 

image

For all other stock movement after the setup of stock accounting, we have the voucher reference on the receipt

image

This information is defined for : 
- stock exit to patient
- stock exit to service
- stock exit to loss
- stock entry from purchase
- stock entry from integration
- stock entry from donation
- stock adjustment

Close #5510

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -1,6 +1,7 @@
 {
   "STOCK" : {
     "ALERT" : "Alert",
+    "DO_NOT_EXIST_BEFORE_ACCOUNTING_SETUP":"Does not exist because the movement of stock was made before the implementation of stock accounting",
     "ADJUSTMENT"      : "Adjustment",
     "ADJUSTMENT_TYPE" : "Adjustment Type",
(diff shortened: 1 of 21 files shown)


jniles (approved):
Tested with both the BCZ database (without stock accounting enabled) and the HEV dataset (with stock account enabled).  Works perfectly in both scenarios.  Thanks @mbayopanda !

bors r+

PR #5752, opened 2021-06-14
improvemet(Aggregate consumption)

lomamech:
- Display remain quatity during editing Aggregate consumption

### DIFF
--- a/client/src/modules/stock/aggregated_consumption/templates/quantity_remain.tmpl.html
+++ b/client/src/modules/stock/aggregated_consumption/templates/quantity_remain.tmpl.html
@@ -1,0 +1,3 @@
+<div class="ui-grid-cell-contents">
+  {{ row.entity.old_quantity - row.entity.quantity_consumed - row.entity.quantity_lost }}
+</div>
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -6,14 +6,14 @@
     "AGGREGATED_STOCK_CONSUMPTION": {
       "AGGREGATED_MOVEMENTS": "Aggregated movements",
       "DETAILS": "Details",
-      "DESCRIPTION": "This aggregated stock consumption module makes it possible to define for a warehouse and for a period the quantity of stock consumed by product but especially by batch, while specifying for each product the number of days of out of stock",
+      "DESCRIPTION": "The compute Aggregate stock consumption, makes possible to define for a warehouse and for a period the quantity of stock consumed but especially by batch number, while specifying for each product the number of days of out of stock",
       "END_OF_CONSUMPTION": "End of consumption",
       "START_OF_CONSUMPTION": "Start of consumption",
       "STOCK_BEGINNING": "Beginning of the month",
       "STOCK_END": "End of the month",
       "TOTAL_EXITS": "Exits",
       "TOTAL_ENTRIES": "Entries",
-      "TITLE": "Aggregate stock Consumption"
+      "TITLE": "Compute Aggregate Stock Consumption"
     },
     "AMOUNT"          : "Amount",
     "APPROVISIONING"  : "Quantity to order",
--- a/client/src/i18n/en/tree.json
+++ b/client/src/i18n/en/tree.json
@@ -9,7 +9,7 @@
     "ADMISSION_REGISTRY": "Admission Registry",
     "AGED_CREDITORS" : "Aged Creditor Report",
     "AGED_DEBTORS" : "Aged Debtors Report",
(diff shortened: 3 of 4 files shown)


lomamech:
image
@mbayopanda , @jmcameron This is a pull request, to facilitate the user experience

lomamech:
image
@mbayopanda , @jmcameron This is a pull request, to facilitate the user experience

jmcameron:
Please describe what this form does and how one would use it.  I'm wondering if should rename this one or the similarly named report https://github.com/IMA-WorldHealth/bhima/pull/5727.

lomamech:
@jmcameron here is a description on the consumption form of aggregated stocks
**AGGREGATE STOCK CONSUMPTION**
For the campaign for the use of the BHIMA computer system at the health zone level, the BHIMA information system is installed in an online server and accessible via the internet at the Zone Office to perform the following operations

- Receive products from Regional Distribution Centers
- Distribute the products to the health centers as well as the health posts by establishing the dispatch notes and the reception slips
 Receipt vouchers will be issued at the BCZs just because the health centers and posts do not have the resources to directly use the BHIMA system. Because of this difficulty, it was essential to be able to set up a form that will allow the BCZs to simulate the consumption of monthly stock at the level of health centers and posts. Reason for which a new module had been added in the BHIMA system to provide for the **consumption of aggregated stock**

image

lomamech:
The difference between #5727 is that the latter is a report, while this one is a form to simulate monthly consumption in health centers and posts.

jmcameron:
@lomamech Thank you for the helpful description!    I suggest renaming this page to something like:
- Compute Aggregate Stock Consumption
- Estimate Aggregate Stock Consumption
- or something similar.
I suggest a name that captures they way it will be used to aid in distributions to other depots.

I think that the name for the similarly named report is okay as it is but I'm open to suggestions to help distinguish these pages/reports.

lomamech:
> @lomamech Thank you for the helpful description! I suggest renaming this page to something like:
> 
> * Compute Aggregate Stock Consumption
> * Estimate Aggregate Stock Consumption
> * or something similar.
>   I suggest a name that captures they way it will be used to aid in distributions to other depots.
> 
> I think that the name for the similarly named report is okay as it is but I'm open to suggestions to help distinguish these pages/reports.

I will update BY **Compute Aggregate Stock Consumption**

client/src/modules/stock/aggregated_consumption/templates/quantity_remain.tmpl.html
```diff
@@ -0,0 +1,3 @@
+<div class="ui-grid-cell-contents">
+  {{ row.entity.old_quantity - row.entity.quantity_consumed - row.entity.quantity_lost }}
```
jniles, line 2: What about the quantity received during the month?  Is that part of "old_quantity"?

lomamech: You are right

lomamech: @jniles  it's run well
image

lomamech: @jniles  it's run well
image

jmcameron (commented):
The code looks fine.
I tested it and it seems to work well.  
I have some concerns about the usability.   I suggest widening the columns for the columns where the user enters data.  It will simplify the UI for the user if he/she can see the header immediately and not have to guess which column is which.  See this figure:
image

I suggest setting widths for those cells.  I played around with this and found some setting that worked for me.  I've attached my modified file; you may find it useful.   The other option is to do line wrap within the header cells... but I could not figure out how to do that.

[aggregated_consumption.js.txt](https://github.com/IMA-WorldHealth/bhima/files/6649086/aggregated_consumption.js.txt)

jmcameron:
Also, the form is very long.  That may be the nature of things, but is it possible that we may want to do the computation for some subset of the all the entries?    Do all depots need ALL of them?  Perhaps we should make it so that each each depot has their own list of preferred articles of stock when can be expanded/reduced as desired.  Then the computations could be done only for that subset.

lomamech:
@jmcameron ,  I opted for your proposal
image

jmcameron:
Cool!

client/src/modules/stock/aggregated_consumption/templates/quantity_remain.tmpl.html
jniles: Thanks @lomamech !

jniles (approved):
This looks good to me.  I've tested it with the latest version of the BCZ database and it works as expected.

bors r+

PR #5756, opened 2021-06-15
Chore: remove testing on MySQL 5.7

### LINKED ISSUE #5364: Move to MySQL8

jniles:
This commit removes testing on MySQL 5.7 and also removes the documentation tied to MySQL 5.7.  It completes the transition to MySQL 8.

Closes #5364. 

-----
[View rendered docs/en/for-developers/installing-bhima.md](https://github.com/jniles/bhima/blob/chore-remove-testing-on-mysql-5.7/docs/en/for-developers/installing-bhima.md)
[View rendered docs/fr/for-developers/installing-bhima.md](https://github.com/jniles/bhima/blob/chore-remove-testing-on-mysql-5.7/docs/fr/for-developers/installing-bhima.md)

### DIFF
--- a/docs/en/for-developers/installing-bhima.md
+++ b/docs/en/for-developers/installing-bhima.md
@@ -10,7 +10,7 @@ This guide will get you up and running with bhima locally. Please note that bhim
 
 Before you begin the installation process, please make sure you have all the bhima dependencies installed locally. We only test on Linux, so your best bet is to use a Linux flavor you are familiar with. Please make sure you have recent version of:
 
-1. [MySQL](http://dev.mysql.com/downloads/) \(5.7 or 8.0\)
+1. [MySQL 8](http://dev.mysql.com/downloads/)
 2. [Redis](https://redis.io)
 3. [curl](https://curl.haxx.se/)
 4. [NodeJS](https://nodejs.org/en/) \(we recommend using [node version manager](https://github.com/creationix/nvm) on linux. Note that we only test on stable and edge\).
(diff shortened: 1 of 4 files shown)


docs/en/for-developers/installing-bhima.md
```diff
@@ -149,9 +149,7 @@ docker run --name mysql8 -p 3306:3306 \
   -e MYSQL_ROOT_PASSWORD=MyPassword \
   -d mysql:8 \
   --sql-mode='STRICT_ALL_TABLES,NO_UNSIGNED_SUBTRACTION' \
-  --default-authentication-plugin=mysql_native_password \
-  --character-set-server=utf8mb4 \
-  --collation-server=utf8mb4_unicode_ci
+  --default-authentication-plugin=mysql_native_password
```
jmcameron, line 152: Do we still need the trouble-shooting section at the end of this document?

jmcameron (approved):
The script changes and documentation changes look goo.d
Was not able to run it to really test it.

jniles:
@jmcameron that was a good catch.  I've removed the troubleshooting section.  It only applied to MySQL anyway.

PR #5757, opened 2021-06-15
Inform about stock available in requisition request

### LINKED ISSUE #5738: Show if article is available in depot during requisition creation
We need to have the ability to know if the articles being requisitioned are in stock or not during the requisition creation process.  We should add a UI element to show us if the selected product is in stock in the depot that will serve the requisition.

mbayopanda:
This PR adds information in each line about available stock in the requisition action modal : 

image

Close #5738

### DIFF
--- a/client/src/modules/stock/requisition/modals/action.modal.js
+++ b/client/src/modules/stock/requisition/modals/action.modal.js
@@ -168,11 +168,8 @@ function ActionRequisitionModalController(
       consumable : 1,
     })
       .then(inventories => {
-        vm.supplierInventoriesQuantities = {};
-        vm.availableSupplierInventories = inventories.map(i => {
-          vm.supplierInventoriesQuantities[i.inventory_uuid] = i.quantity;
-          return i.inventory_uuid;
-        });
+        vm.supplierInventoriesQuantities = new Map(inventories.map(i => ([i.inventory_uuid, i.quantity])));
+        vm.availableSupplierInventories = inventories.map(i => i.inventory_uuid);
         return {
           supplierInventoriesQuantities : vm.supplierInventoriesQuantities,
           availableSupplierInventories : vm.availableSupplierInventories,
@@ -193,7 +190,8 @@ function ActionRequisitionModalController(
       inventory.isAvailable = !!vm.availableSupplierInventories.includes(identifier);
 
       if (inventory.isAvailable) {
-        inventory.isEnough = !!(vm.supplierInventoriesQuantities[identifier] >= quantity);
+        inventory.hasEnoughQuantity = !!(vm.supplierInventoriesQuantities.get(identifier) >= quantity);
+        inventory.supplierAvailableQuantity = vm.supplierInventoriesQuantities.get(identifier);
       }
     }
 
--- a/client/src/modules/stock/requisition/modals/action.modal.js
+++ b/client/src/modules/stock/requisition/modals/action.modal.js
@@ -16,6 +16,12 @@ function ActionRequisitionModalController(
   const store = new Store({ data : [] });
   const columns = [
     {
+      field : 'available',
+      displayName : '',
+      width : 25,
+      cellTemplate : 'modules/stock/requisition/templates/available.cell.html',
+    },
+    {
       field : 'inventory_uuid',
       displayName : 'FORM.LABELS.INVENTORY',
       headerCellFilter : 'translate',
@@ -59,9 +65,18 @@ function ActionRequisitionModalController(
   vm.configureItem = configureItem;
   vm.autoSuggestInventories = autoSuggestInventories;
   vm.cancel = cancel;
+  vm.checkInventoryAvailability = checkInventoryAvailability;
 
   vm.onSelectDepot = depot => {
     vm.model.depot_uuid = depot.uuid;
+
+    loadAvailableInventories()
+      .then(values => {
+        vm.supplierInventoriesQuantities = values.supplierInventoriesQuantities;
+        vm.availableSupplierInventories = values.availableSupplierInventories;
+        handleAvailability();
+      })
+      .catch(Notify.handleError);
   };
 
   vm.onDateChange = date => {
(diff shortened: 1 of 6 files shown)


client/src/modules/stock/requisition/modals/action.modal.js
```diff
@@ -102,6 +117,55 @@ function ActionRequisitionModalController(
...
+  function loadAvailableInventories() {
+    if (!vm.model.depot_uuid) { return {}; }
+
+    return Stock.inventories.read(null, {
+      depot_uuid : vm.model.depot_uuid,
+      includeEmptyLot : 0,
+      consumable : 1,
+    })
+      .then(inventories => {
+        vm.supplierInventoriesQuantities = {};
```
jniles: This actually looks like you could benefit from a real [Map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).  You could do the following:

```js
vm.supplierInventoryQuantities = new Map(inventories.map(i => ([i.inventory_uuid, i.quantity])));
```

Then you can use `vm.supplierInventoryQuantities.get(uuid)` to return the quantity.

client/src/modules/stock/requisition/modals/action.modal.js
```diff
@@ -102,6 +117,55 @@ function ActionRequisitionModalController(
...
+        inventory.quantity = currentQuantity;
+      }
+
+      const identifier = vm.enableAutoSuggest ? inventory.inventory_uuid : inventory.uuid;
+      const quantity = vm.enableAutoSuggest ? inventory.S_Q : inventory.quantity;
+
+      inventory.isAvailable = !!vm.availableSupplierInventories.includes(identifier);
+
+      if (inventory.isAvailable) {
+        inventory.isEnough = !!(vm.supplierInventoriesQuantities[identifier] >= quantity);
```
jniles: Instead of `isEnough` here, let's call this variable `hasEnoughQuantity`.  It's longer, but it makes it slightly easier to read and understand in the future.

jmcameron (commented):
I have not looked at the code yet, but I did try it.  I ran into some difficulties (or confusion on how to use it on my part):

I used bhima_test and saw which items were in stock  for the primary depot:
image

I then went to "Stock > Requisitions" and tried the [New Requisition] button. Note that I had the Secondary Depot selected on this page.  I could not get that to work even when trying to create a requisition from the secondary depot from the primary depot (which has the 2 items in stock shown above).
image

So I tried the [Other Requistion] button and had partial luck.  I was able to set up the requisition for one of the items, but not the other.  I was not able to requisition the quinine even though it had stock.  I see in the Articles in Stock display that some of its supply was expired.    From what I saw, it seems like that was preventing the item from being selected at all.
image

Also, I wonder if the depot selector for the providing Depot should be immediately after (or before) the requisitioning Depot selector?

jmcameron:
By the way, I think this is a very useful feature!

client/src/modules/stock/requisition/templates/available.cell.html
```diff
@@ -0,0 +1,21 @@
...
+        class="fa fa-thumbs-up text-success">
+      </i>
+      <i
+        translate-attr="{title : 'REQUISITION.INVENTORY_AVAILABLE_WITH_NOT_ENOUGH_QUANTITY'}"
+        ng-if="!row.entity.inventory.isEnough"
+        class="fa fa-warning text-warning">
+      </i>
+    </span>
+    <i 
+      translate-attr="{title : 'REQUISITION.INVENTORY_NOT_AVAILABLE'}"
```
jniles: Could you include the amount that is available in the target depot using a `uib-tooltip?`

image

That would make it much easier to validate that the information is correct from a user's standpoint.

jniles (commented):
@mbayopanda I've tested it and this works as described.  I've mentioned a couple of code improvements as well as one improvement to the user experience.  Let me know what you think.

mbayopanda:
> I have not looked at the code yet, but I did try it. I ran into some difficulties (or confusion on how to use it on my part):
> 
> I used bhima_test and saw which items were in stock for the primary depot:
> image
> 
> I then went to "Stock > Requisitions" and tried the [New Requisition] button. Note that I had the Secondary Depot selected on this page. I could not get that to work even when trying to create a requisition from the secondary depot from the primary depot (which has the 2 items in stock shown above).
> image
> 
> So I tried the [Other Requistion] button and had partial luck. I was able to set up the requisition for one of the items, but not the other. I was not able to requisition the quinine even though it had stock. I see in the Articles in Stock display that some of its supply was expired. From what I saw, it seems like that was preventing the item from being selected at all.
> image
> 
> Also, I wonder if the depot selector for the providing Depot should be immediately after (or before) the requisitioning Depot selector?

Thank you @jmcameron for the review, on bhima_test database I have got the same result as you. I am investigating about that. Thank you

mbayopanda:
Thank you @jniles for your review, I applied your recommandation.

mbayopanda:
@jmcameron changes landed, could you review it again please, when it is possible. Thank you

jmcameron (approved):
I looked over the code and it looks good.
I ran the tests I previously tried and worked perfectly.  Good Job!
I was able to get either [New Requision] or [Other Requisition] both working properly.   
NOTE: I'm not clear how [Other Requistion] is different from [New Requisition].

I also realized that the complaint I had earlier: 
   "I was not able to requisition the quinine even though it had stock. 
    I see in the Articles in Stock display that some of its supply was expired."
was my error.  I was not searching with the right inventory code.  When I tried the right one (DORA_QUIN...), it worked fine. 

I tried hovering over the status indicator to see how many items were available and that works well and will be useful.

I also tried a whole cycle of creating a requisition, doing a stock exit (transfer) of for the supplying depot, and doing a stock (transfer entry) on the receiving depot.  It all worked cleanly. 

It looks great to me!

jniles (approved):
Looks good to me.  Approved by @jmcameron as well.

bors r+

PR #5758, opened 2021-06-15
refactor(stock): change "cost" to "value"

### LINKED ISSUE #5746: Change the word "cost" to "value" in the stock entry from purchase document
In the stock entry from purchase, we should update the text to say "value", not "cost", since this refers to the value of the stock, rather than the purchase price.

value-not-cost-entry-from-purchase

jniles:
This commit changes the wording of "cost" to "value" in the stock entry
form to better represent that the stock is _valued_ at that amount,
regardless of what the enterprise paid for it.

Closes #5746.


jmcameron (approved):
Looks good.  I tested it and works.

PR #5759, opened 2021-06-15
chore(purchase): zero out purchase price

### LINKED ISSUE #5734: Do not pre-populate the purchase price in the purchase order form
Our partners in Vanga have reversed their feature request from pre-populating the purchase order form with the previous purchase price (https://github.com/IMA-WorldHealth/bhima/issues/4119), to just putting in 0.  This is because their users keep putting in the wrong unit price information by just clicking "submit".  We can avoid this by simply putting 0 as the purchase price and forcing them to put in the real purchase price.

jniles:
This commit zeroes out the purchase price so users are forced to put in
their own purchase price.

Closes #5734.

image

### DIFF
--- a/client/src/modules/purchases/create/PurchaseOrderForm.js
+++ b/client/src/modules/purchases/create/PurchaseOrderForm.js
@@ -42,8 +42,15 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
     this._ready = $q.defer();
 
     // set up the inventory
-    Inventory.read(null, { locked : 0, use_previous_price : 1 })
+    Inventory.read(null, { locked : 0 })
       .then((data) => {
+
+        // zero out price so that users don't automatically use the sale price.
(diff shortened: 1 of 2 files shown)


client/src/modules/purchases/create/PurchaseOrderForm.js
```diff
@@ -42,8 +42,15 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
     this._ready = $q.defer();
 
     // set up the inventory
-    Inventory.read(null, { locked : 0, use_previous_price : 1 })
+    Inventory.read(null, { locked : 0 })
       .then((data) => {
+
+        // zero out price so that users don't automatically use the sale price.
+        data.forEach(row => {
```
mbayopanda, line 49: 👍🏽

mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #5762, opened 2021-06-17
Feature filter requisition

### LINKED ISSUE #5687: We need to filter on requisition status in the requisition registry
Here is the requisition registry:

image

For the health zone stock management, we'll need to implement filtering on the status code in the search modal.  This should be a search field option in the search modal.

lomamech:
- implement filtering requisition by status
- Resolve conflict with branch master
- Improve: search requisition by status
- implement test intergration for filter requisition by status

closes #5687

### DIFF
--- a/client/src/modules/stock/requisition/modals/search.modal.html
+++ b/client/src/modules/stock/requisition/modals/search.modal.html
@@ -38,6 +38,20 @@
             </div>
           </div>
 
+          <div>
+            <label class="control-label" translate>FORM.LABELS.STATUS</label>
+            <div class="row">
+              <div ng-repeat ="r in $ctrl.requisitionStatus | orderBy:'plainText'" class="col-md-6">
+                <div class="checkbox">
+                  <label>
+                    <input type="checkbox" ng-model="r.checked" ng-true-value="1" ng-false-value="0">
+                    <span ng-class="r.class_style" translate> {{ r.plainText }} </span>
+                  </label>
+                </div>
+              </div>
+            </div>
+          </div>   
+
           <!-- depot -->
           <bh-depot-select
             depot-uuid="$ctrl.searchQueries.depot_uuid"
@@ -53,6 +67,7 @@
             on-select-callback="$ctrl.onSelectUser(user)">
             <bh-clear on-clear="$ctrl.clear('user_id')"></bh-clear>
           </bh-user-select>
+         
         </div>
       </uib-tab>
       <uib-tab index="1" heading="{{ 'FORM.LABELS.DEFAULTS' | translate }}" data-default-filter-tab>
--- a/client/src/modules/stock/requisition/modals/search.modal.html
+++ b/client/src/modules/stock/requisition/modals/search.modal.html
@@ -38,11 +38,11 @@
             </div>
           </div>
 
-          <div>
+          <div style="margin-bottom: 10px">
             <label class="control-label" translate>FORM.LABELS.STATUS</label>
             <div class="row">
               <div ng-repeat ="r in $ctrl.requisitionStatus | orderBy:'plainText'" class="col-md-6">
-                <div class="checkbox">
+                <div class="checkbox" style="margin-top: 0; margin-bottom: 2px">
                   <label>
                     <input type="checkbox" ng-model="r.checked" ng-true-value="1" ng-false-value="0">
                     <span ng-class="r.class_style" translate> {{ r.plainText }} </span>
@@ -50,7 +50,7 @@
                 </div>
               </div>
             </div>
-          </div>   
+          </div>
 
           <!-- depot -->
           <bh-depot-select
--- a/server/models/migrations/next/migrate.sql
+++ b/server/models/migrations/next/migrate.sql
@@ -1,1 +1,56 @@
+<<<<<<< HEAD
 /* migration file from v1.20.0 to next */
+=======
+/**
+ * Inventory tags table
+ */
+DROP TABLE IF EXISTS `inventory_tag`;
+CREATE TABLE `inventory_tag` (
+  `inventory_uuid`          BINARY(16) NOT NULL,
+  `tag_uuid`          BINARY(16) NOT NULL,
+  FOREIGN KEY (`inventory_uuid`) REFERENCES `inventory` (`uuid`),
+  FOREIGN KEY (`tag_uuid`) REFERENCES `tags` (`uuid`)
+) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+
+/**
+ * @author: jmcameron
+ * @date: 2021-06-02
+ */
+INSERT IGNORE INTO `report` (`report_key`, `title_key`) VALUES
+('aggregated_stock_consumption', 'REPORT.AGGREGATED_STOCK_CONSUMPTION.TITLE');
+
+INSERT IGNORE INTO `unit` VALUES
+(296, 'Agg. Stock Consumption Report','REPORT.AGGREGATED_STOCK_CONSUMPTION.TITLE',
+ 'aggregated consumption', 282, '/reports/aggregated_stock_consumption');
+
+
+/**
+ * @author: mbayopanda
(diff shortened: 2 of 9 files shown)


lomamech:
image

server/models/migrations/next/migrate.sql
```diff
@@ -31,8 +31,32 @@ INSERT IGNORE INTO `unit` VALUES
 
 ALTER TABLE transaction_history ADD COLUMN `value` TEXT DEFAULT NULL;
 
+<<<<<<< HEAD
 /* author: jniles
  * @date: 2021-06-15
  */
 INSERT INTO `actions`(`id`, `description`) VALUES
   (8, 'USERS.ACTIONS.EDIT_LOT');
+=======
```
jniles: This has a broken merge in it.  Can you please remove these?

test/integration/stock/requisition.js
```diff
@@ -8,7 +8,7 @@ describe('(/stock/requisition) The Stock Assign HTTP API', () => {
   const keys = [
     'uuid', 'requestor_uuid', 'requestor_type_id', 'description', 'date',
     'user_id', 'project_id', 'user_display_name', 'depot_uuid', 'depot_text', 'service_requestor',
-    'depot_requestor', 'reference', 'items', 'status_key', 'title_key',
+    'depot_requestor', 'reference', 'items', 'status_key', 'title_key', 'class_style', 'status_id',
```
jniles, line 11: How did we get the class style previously?  Do we really need to store it in the database?

server/models/migrations/next/migrate.sql
```diff
@@ -31,8 +31,32 @@ INSERT IGNORE INTO `unit` VALUES
 
 ALTER TABLE transaction_history ADD COLUMN `value` TEXT DEFAULT NULL;
 
+<<<<<<< HEAD
```
mbayopanda: Could you fix this conflict

server/models/migrations/next/migrate.sql
```diff
@@ -31,8 +31,32 @@ INSERT IGNORE INTO `unit` VALUES
...
 INSERT INTO `actions`(`id`, `description`) VALUES
   (8, 'USERS.ACTIONS.EDIT_LOT');
+=======
+
+/**
+ * @author: lomamech
+ * @date: 2021-06-17
+ * Requisition status
+ */
+ALTER TABLE `status` ADD COLUMN `class_style` tinyint(1) unsigned NOT NULL DEFAULT 0;
```
mbayopanda: I don't think the type is `tinyint` according to lines below, could you fix that

client/src/modules/stock/requisition/modals/search.modal.html
```diff
@@ -38,6 +38,34 @@
...
+                      <label>
+                        <input type="checkbox" ng-model="r.checked" ng-true-value="1" ng-false-value="0">
+                        <span ng-class="r.class_style" translate> {{ r.plainText }} </span>
+                      </label>
+                    </div>
+                  </span>
+                </span>
+              </div>
+            </div>
+          </div>
```
mbayopanda: You can also write (if the even or odd doesn't matter) : 
```html
<div class="row">
  <div ng-repeat ="r in $ctrl.requisitionStatus | orderBy:'plainText'" class="col-md-6">
    <div class="checkbox">
      <label>
        <input type="checkbox" ng-model="r.checked" ng-true-value="1" ng-false-value="0">
        <span ng-class="r.class_style" translate> {{ r.plainText }} </span>
      </label>
    </div>
  </div>
</div>
```

mbayopanda (commented):
@lomamech the feature works as expected 👍🏽 

I made some comments about `html` and migration script

server/models/migrations/next/migrate.sql
lomamech: Good catch, I had resolved this conflict but it is as if the file had not been saved

client/src/modules/stock/requisition/modals/search.modal.html
lomamech: I will do

lomamech:
@mbayopanda  it's done

jmcameron (commented):
The code looks okay.

I tested it and it works, but I noticed a couple of odd things.
- When I enabled the "Canceled" status, it worked very well.
- But the display of Status flag is a bit odd:   [Status : /Cancelled / x]  (it seems like the label should have been cleaned/filtered.
- When you disable the Status flag (in the Search modal), it does clear it, but then the Status flag is also odd:  [Status : / x]  (I don't think the status flag should show at all if no status option is checked)
- Also, when you go back into the Search modal, the previous setting has been lost.   It should probably have been cached.

lomamech:
> The code looks okay.
> 
> I tested it and it works, but I noticed a couple of odd things.
> 
> * When I enabled the "Canceled" status, it worked very well.
> * But the display of Status flag is a bit odd:   [Status : /Cancelled / x]  (it seems like the label should have been cleaned/filtered.
> * When you disable the Status flag (in the Search modal), it does clear it, but then the Status flag is also odd:  [Status : / x]  (I don't think the status flag should show at all if no status option is checked)
> * Also, when you go back into the Search modal, the previous setting has been lost.   It should probably have been cached.

Ok thanks i will work on this

lomamech:
@jmcameron  It is done
image

client/src/modules/stock/requisition/modals/search.modal.html
```diff
@@ -38,6 +38,20 @@
             </div>
           </div>
 
+          <div>
+            <label class="control-label" translate>FORM.LABELS.STATUS</label>
+            <div class="row">
+              <div ng-repeat ="r in $ctrl.requisitionStatus | orderBy:'plainText'" class="col-md-6">
+                <div class="checkbox">
```
jmcameron: This is a minor, cosmetic suggestion:  You could reduce the vertical space between the status check boxes to reduce the excess space that this selection takes.  Even better, if it could go into three columns that would help even more. Here is a suggestion for addressing this (notice the two "style" additions).
>           <div style="margin-bottom: 10px">
>             <label class="control-label" translate>FORM.LABELS.STATUS</label>
>             <div class="row">
>               <div ng-repeat ="r in $ctrl.requisitionStatus | orderBy:'plainText'" class="col-md-6">
>                 <div class="checkbox" style="margin-top: 0; margin-bottom: 2px">
>                   <label>
>                     <input type="checkbox" ng-model="r.checked" ng-true-value="1" ng-false-value="0">
>                     <span ng-class="r.class_style" translate> {{ r.plainText }} </span>
>                   </label>
>                 </div>
>               </div>
>             </div>
>           </div>

It seems inconvenient to have to scroll down to see the last field and the Submit/Cancel buttons.  Is it possible to make the search modal taller when it comes up to avoid scrolling?

jmcameron (commented):
I tried it again and the status labeling works perfectly.  Nice!  
You addressed all the issues I noticed in https://github.com/IMA-WorldHealth/bhima/pull/5762#pullrequestreview-688381485.

I added one comment on cosmetics, that you may or may not want to address.  Other than that, I think it is ready to go!

client/src/modules/stock/requisition/modals/search.modal.html
lomamech: Very nice 
image

jmcameron: Looks good.

jmcameron (approved):

server/models/migrations/next/migrate.sql
```diff
@@ -1 +1,56 @@
+<<<<<<< HEAD
```
jniles: This is a broken merge that needs to be fixed.

Otherwise, this looks good to me.

jniles (changes requested):
Just one broken merge to fix.  Otherwise, @jmcameron and @mbayopanda have done a good job with this review.

server/models/migrations/next/migrate.sql
lomamech: Hooo

lomamech:
@jniles  it's done

jniles (approved):
LGTM.

bors r+

PR #5763, opened 2021-06-18
Automatically merge lots for inventory items with zero quantity in stock 

### LINKED ISSUE #5740: If quantities in stock of lots in duplicate lots list is 0, auto-merge based on furthest expiration date
If we don't have any quantities in stock, we cannot know what the real expiration date is for any lots with matching lot numbers.  Therefore, instead of asking a user to pick randomly, let's just merge them automatically, using the expiration date the farthest in the future.  This will clean up lots and save the user time.

jmcameron:
This PR adds a command to the main [Menu] under the "Stock > Find Duplicate Lots" page that allows automatic merging of all lots for (specific inventory items) with zero quantity in stock.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5740

**TESTING**
- Use a production database
- Make sure the 1.19 - 1.20 SQL changes are executed
- Go to Stock > Find Duplicate Lots
- Try the new function (in the top [Menu])
- It should reduce the number of duplicate lots significantly.

NOTE: After doing this, there will still be some lots with zero quantity in stock.    That is because the new function only automerges lots with zero quantity in stock over all lots for an inventory type.   Whereas the main display does it for combinations of inventory type AND lot labels.  

ALSO: This PR fixes a bug in the quantity-in-stock for the main display; it now should show the correct result (before it was showing the quantity in stock for whichever item the "GROUP BY" selected.


client/src/js/services/LotService.js
```diff
@@ -50,6 +50,18 @@ function LotService(Api, $http, util) {
       .then(util.unwrapHttpResponse);
   };
 
+  lots.allDupes = (params) => {
+    return $http.get('/lots_all_dupes', { params })
+      .then((res) => {
+        res.data.forEach((lot) => {
```
jniles: There is nothing wrong with this code, and it works as described.

My preference: I would chain after `unwrapHttpResponse`, rather than inserting this before it.  That way, you are only messing with the data that comes from the request rather than having access to all the headers/full response object.  It reduces the "what do I have to think about" surface area when reading this code.

server/controllers/stock/lots.js
```diff
@@ -173,6 +155,54 @@ function getDupes(req, res, next) {
...
+       WHERE lot2.inventory_uuid = lot1.inventory_uuid AND lot2.label = lot1.label) as num_duplicates
+    FROM lot lot1
+    JOIN inventory as inv ON lot1.inventory_uuid=inv.uuid
+    ORDER BY inventory_name, num_duplicates, label;
+  `;
+
+  // Then compute the max quantity-in-stock for all lots for each inventory_uuid, label pair
+  const query2 = `
+    SELECT
+      HEX(lot1.uuid) AS uuid, label, unit_cost, expiration_date, entry_date,
```
jniles: Instead of using `HEX()` can you use `BUID()`?  All it does is a `HEX()`, but it is what we have been using everywhere, and if we decide to change it in a future upgrade, we won't have a mixture of calls for converting uuids from binary to hex and back.

server/controllers/stock/lots.js
```diff
@@ -303,6 +333,91 @@ function autoMerge(req, res, next) {
...
+  return transaction.execute()
+    .then(txresults => {
+      const rows = txresults[2];
+      numLots = rows.length;
+      const dbPromises = [];
+      rows.forEach((row) => {
+        const keepLotUuid = row.uuid;
+        const promise = db.exec(getLotsSQL, [db.bid(row.inventory_uuid), db.bid(keepLotUuid)])
+          .then((lots) => {
+            const lotUuids = lots.reduce((list, elt) => {
```
jniles: Why does a `reduce` make sense here and not a `map`?  It looks like you are trying to convert a list of elements into a list of element uuids, which seems like a `map()`.

I think this is probably what you want:
```js
const lotUuids = lots.map(elt => elt.uuid);
```

server/controllers/stock/lots.js
```diff
@@ -303,6 +333,91 @@ function autoMerge(req, res, next) {
...
+  transaction.addQuery(query1);
+  transaction.addQuery(query2);
+  transaction.addQuery(query3);
+
+  return transaction.execute()
+    .then(txresults => {
+      const rows = txresults[2];
+      numLots = rows.length;
+      const dbPromises = [];
+      rows.forEach((row) => {
```
jniles: Instead of a `.forEach()`, let's do a `map()` here as well.  Ideally, you want to end with an array of promises, right?  And those promises are generated for each element in the array.  So we can probably be able to just do a `map()` on each element, returning a promise.

```js
// combining with the comment I made down below, and adding my own comments for clarity.
const dbPromises = rows.map(row => {
  const keepLotUuid = row.uuid;

  // get the lots associated with this inventory, excluding the lot to keep
  return db.exec(getLotsSQL, [db.bid(row.inventory_uuid), db.bid(keepLotUuid)])
    .then(lots => {

      // assemble the lot uuids as an array for the mergeLotsInternal() call
      const lotUuids = lots.map(elt => elt.uuid);

      // merge all lotUuids into keepLotUuid
      return mergeLotsInternal(keepLotUuid, lotsUuid);
    });
});
```

jniles: Then, it is business as usual:

```js
return Promise.all(dbPromises);
```

server/controllers/stock/lots.js
```diff
@@ -303,6 +333,91 @@ function autoMerge(req, res, next) {
...
+        const keepLotUuid = row.uuid;
+        const promise = db.exec(getLotsSQL, [db.bid(row.inventory_uuid), db.bid(keepLotUuid)])
+          .then((lots) => {
+            const lotUuids = lots.reduce((list, elt) => {
+              list.push(elt.uuid);
+              return list;
+            }, []);
+            return mergeLotsInternal(keepLotUuid, lotUuids);
```
jniles, line 406: Just to confirm - does `mergeLotsInternal` use a transaction?  I hope so.  If it doesn't, we need to stick all of this in the same transaction.

server/controllers/stock/lots.js
```diff
@@ -303,6 +333,91 @@ function autoMerge(req, res, next) {
...
+    ORDER BY inventory_uuid,expiration_date;
+  `;
+
+  // Create a temporary table with max quantity-in-stock for all lots for each inventory_uuid
+  const query2 = `
+    CREATE TEMPORARY TABLE tmp_max_quantities AS
+    SELECT
+      uuid, expiration_date,
+      code, inventory_name, inventory_uuid, num_duplicates,
+      MAX(quantity_in_stock) OVER (PARTITION BY inventory_uuid) AS max_quantity
```
jniles, line 363: Nice!  I like this.

jniles (commented):
I've tested this and it works as described.

I've made a few comments to try and clean up the code.  Let me know what you think.  Otherwise, the functionality looks good to me.

server/controllers/stock/lots.js
jmcameron: Yes, it does use a transaction.

server/controllers/stock/lots.js
jmcameron: Very clean!  Thanks!

client/src/js/services/LotService.js
jmcameron: Done

server/controllers/stock/lots.js

jmcameron:
@jniles After you re-review this, assuming you approve it, I'll rebase and release it.

jniles (approved):
Looks good to me. 👍 

@jmcameron I'll let you release it whenever.

server/controllers/stock/lots.js

PR #5768, opened 2021-06-21
feature(Stock Exit to Patient)

### LINKED ISSUE #5256: Show recent invoices on stock exit to patient modal
A feature request from our friends at HEV:

To link a patient + invoice to a stock distribution, BHIMA requires you to first select a patient, then search for an invoice.  See below:

image

As a user experience, this is a bit clunky.  Our users would like to have a list of recent invoices that they can pick from to rapidly select the correct invoice.  This would reduce the amount of work they do.

To make this work correctly, I propose we show the most recent 3 invoices.  We could also include some sort of limit on dates, such as those made in the last month.   If they cannot find the invoice from that list, they can look it up with the same invoice lookup we currently have.

lomamech:
- Show recent invoices on stock exit to patient modal

closes #5256


client/src/modules/stock/exit/modals/findPatient.modal.html
```diff
@@ -41,6 +42,21 @@
...
+            <input
+              type="radio"
+              ng-click="$ctrl.findDetailInvoice(invoice)"
+              name="invoice">
+            <span><strong>{{ invoice.reference }}</strong> - {{ invoice.date | date:'dd MMMM yyyy, hh:mm' }}</span>
+          </label>
+        </div>
+      </div>
+      <div class="text-warning" ng-if="!$ctrl.patientInvoices.length && $ctrl.joinInvoice">
+        <i class="fa fa-warning"></i> <span translate> FORM.INFO.THIS_PATIENT_NOT_BILLED </span>
```
jmcameron, line 59: FORM.INFO.THIS_PATIENT_NOT_BILLED does not appear to be translated into English.

jmcameron (commented):
The code looks good and seems to work well.
I noticed one translation item was not translated into English (when a patient is selected who has no invoices):
image

client/src/modules/stock/exit/modals/findPatient.modal.js
```diff
@@ -39,12 +44,36 @@ function StockFindPatientModalController(Instance, Patients, Notify, Data, AppCa
...
-
   // set patient
   function setPatient(patient) {
+    vm.joinInvoice = false;
     vm.selected = patient;
   }
 
+  function getLastInvoice() {
+    // load debtor invoices
+    Debtors.invoices(vm.selected.debtor_uuid, { descLimit5 : 1 })
```
mbayopanda, line 58: `descLimit5`?

server/controllers/finance/debtors/index.js
```diff
@@ -185,13 +185,20 @@ function invoiceBalances(debtorUuid, uuids, options = {}) {
...
   let balanced = '';
+  let orderBy = '';
 
   if (options.balanced === '1') {
     balanced = 'HAVING balance = 0';
   } else if (options.balanced === '0') {
     balanced = 'HAVING balance <> 0';
   }
 
+  if (options.descLimit5 === '1') {
```
mbayopanda, line 196: Ah okay

mbayopanda (commented):
I still have this message
image

jniles:
bors try

client/src/modules/stock/exit/modals/findPatient.modal.html
lomamech: I just added it, for the first time I didn't save the file

lomamech:
image

image

i just updated it

client/src/modules/stock/exit/modals/findPatient.modal.html
jmcameron: The latest update fixed this.

jmcameron (approved):
I think it is ready now.

jniles (commented):
@lomamech this looks good.  I'm happy with the code and @jmcameron is right, it is good to go.

One small change while you are working here - can you add the value of the invoice to the listing?  This would probably be between the reference, and the date.  That way, it would say "IV.HEV.123 - $12.54 - June 12, 2021".  It is much easier for users to recognize the price than the invoice ID, I think.

lomamech:
> @lomamech this looks good. I'm happy with the code and @jmcameron is right, it is good to go.
> 
> One small change while you are working here - can you add the value of the invoice to the listing? This would probably be between the reference, and the date. That way, it would say "IV.HEV.123 - $12.54 - June 12, 2021". It is much easier for users to recognize the price than the invoice ID, I think.

ok i will add it

lomamech:
image

jniles (approved):
Excellent job, @lomamech 

bors r+

PR #5769, opened 2021-06-21
Enable editing lot expiration dates during stock entry

### LINKED ISSUE #5741: We need a way to update expiration dates from stock entry, if the system proposed expiration date is wrong
In the stock entry form, the system provided expiration date may be wrong due to negligence or mistyping of a previous stock entry.  To prevent this error from propagating into the future, we should add a checkbox for modifying the expiration dates of previously created lots.  If the user checks this, they will be able to modify expiration dates and the expiration date they provide will be saved/updated for that lot.

Here is where they would have that option:

image

jmcameron:
Enable editing lot expiration dates during stock entry for users with the "EDIT_LOT" action permission.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5741

NOTE:  Rather than do this on a line by line basis, I made it for all lots selected.   It seemed cumbersome to do this on every line.  However I'm willing to reconsider this if reviewers feel that we need the extra protection on each line to avoid date entry errors.  However, anyone who does not have the EDIT_LOT action permission will not be able to edit the expiration dates during stock entry in any case.   So if we trust someone to edit any of the lot expiration dates, perhaps we can trust them not to accidentally edit lot expiration dates.

*TESTING*
- The following procedure assumes bhima_test, but production databases should work as well.  Just check the Lot registry for suitable lots.
- Make sure that the "Edit lot information" action is permitted for your user.
- Go to Stock > Entry
- Use the entry mode of your choice (I tested it with Integration, but it should not matter).
- Add DINJ_TRIB1A2_0 : Vitamines B1+B6+B12, 100+50+0.5mg/2ml, Amp, Unité
- Select lots
   - Note the new option to enable editing of expiration dates (do not click it yet)
   - Select one of the existing lots
   - Notice that you cannot change the expiration date
   - Now, click on the option above to enable editing of lot expiration dates
   - Change the date
   - Save the modal
   - Save the Entry
   - Notice the new date in the PDF report that comes up
- Try repeating the stock entry and do not change the date and notice the expiration date does not change
- Disable the EDIT_LOTS action and repeat the test above
   - Note that the option to enable editing lot expiration dates is no longer available.

### DIFF
--- a/client/src/modules/stock/entry/modals/lots.modal.html
+++ b/client/src/modules/stock/entry/modals/lots.modal.html
@@ -1,5 +1,6 @@
 <form
   name="FindForm"
+  bh-submit="$ctrl.submit(FindForm)"
   ng-model-options="{ 'debounce' : { 'default' : 200, 'blur' : 0 }}"
   novalidate>
 
@@ -17,7 +18,7 @@
       <div class="col-xs-12">
         <div>
           <span class="text-bold" translate>STOCK.ENTRY_DATE</span> :
-          {{ $ctrl.entryDate | date:$ctrl.Constants.dates.format }}
+          {{ $ctrl.entryDate | date:$ctrl.bhConstants.dates.format }}
         </div>
       </div>
     </div>
@@ -73,7 +74,22 @@
         <p class="help-block" translate>LOTS.ENABLE_FAST_INSERT_DESCRIPTION</p>
       </div>
 
-      <bh-add-item callback="$ctrl.form.addItem()" disable="true">
+      <div style="margin-bottom: 5px;" bh-has-permission="$ctrl.bhConstants.actions.EDIT_LOT">
+        <div class="checkbox">
+          <label>
+            <input
+              id="editExpirationDates"
+              type="checkbox"
+              ng-model="$ctrl.editExpirationDates"
+              ng-change="$ctrl.onExpDateEditable()"
+              ng-true-value="1"
+              ng-false-value="0">
+            <span translate>LOTS.ENABLE_CORRECTING_LOT_EXPIRATION_DATES</span>
+          </label>
+          <p class="help-block" translate>LOTS.ENABLE_CORRECTING_LOT_EXPIRATION_DATES_DESCRIPTION</p>
+        </div>
+
+        <bh-add-item callback="$ctrl.form.addItem()" disable="true">
       </bh-add-item>
     </div>
 
@@ -105,8 +121,8 @@
       FORM.BUTTONS.CLOSE
     </button>
 
-    <button type="submit" class="btn btn-primary" ng-click="$ctrl.submit(FindForm, $event)" data-method="submit" translate>
-      FORM.BUTTONS.SUBMIT
-    </button>
+    <bh-loading-button loading-state="FindForm.$loading">
+      <span translate>FORM.BUTTONS.SUBMIT</span>
+    </bh-loading-button>
   </div>
 </form>
--- a/client/src/modules/stock/entry/modals/lots.modal.js
+++ b/client/src/modules/stock/entry/modals/lots.modal.js
@@ -2,15 +2,15 @@ angular.module('bhima.controllers')
   .controller('StockDefineLotsModalController', StockDefineLotsModalController);
 
 StockDefineLotsModalController.$inject = [
-  'appcache', '$uibModalInstance', 'uiGridConstants', 'data', 'SessionService',
-  'CurrencyService', 'NotifyService', 'bhConstants', 'StockEntryModalForm',
-  '$translate', 'focus',
+  'appcache', '$uibModalInstance', 'uiGridConstants', 'data', 'LotService',
+  'SessionService', 'CurrencyService', 'NotifyService', 'StockEntryModalForm',
+  'bhConstants', '$translate', 'focus',
 ];
 
 function StockDefineLotsModalController(
-  AppCache, Instance, uiGridConstants, Data, Session,
-  Currencies, Notify, bhConstants, EntryForm,
-  $translate, Focus,
+  AppCache, Instance, uiGridConstants, Data, Lots,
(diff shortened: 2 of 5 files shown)


client/src/modules/stock/entry/modals/lots.modal.html
```diff
@@ -17,7 +17,7 @@
       <div class="col-xs-12">
         <div>
           <span class="text-bold" translate>STOCK.ENTRY_DATE</span> :
-          {{ $ctrl.entryDate | date:$ctrl.Constants.dates.format }}
+          {{ $ctrl.entryDate | date:$ctrl.bhConstants.dates.format }}
```
mbayopanda, line 21: 👍🏽

client/src/modules/stock/entry/modals/lots.modal.html
```diff
@@ -73,7 +73,22 @@
         <p class="help-block" translate>LOTS.ENABLE_FAST_INSERT_DESCRIPTION</p>
       </div>
 
-      <bh-add-item callback="$ctrl.form.addItem()" disable="true">
+      <div style="margin-bottom: 5px;" bh-has-permission="$ctrl.bhConstants.actions.EDIT_LOT">
```
mbayopanda, line 77: 👍🏽

mbayopanda (approved):
The feature works as described in issue #5741 

LGTM 👍🏽

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -141,6 +144,18 @@ function StockDefineLotsModalController(
...
 
+  /**
+   * @method getExistingLot
+   *
+   * @description
+   *  If the lot given is a string, it is a label, so look it up by label.
+   *  If the lot is not a string, it will contain the lots UUID, use it to look up the lot
+   * @param {lot} the 'lot' object from the lot selection popup modal form
+   */
+  function getExistingLot(lot) {
```
jniles, line 155: Nice 👍

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -309,12 +325,26 @@ function StockDefineLotsModalController(
...
-        quantity : vm.form.total(),
-      });
+
+      // Maybe update some lot expiration dates
+      const promises = [];
+      if (vm.editExpirationDates) {
+        vm.form.rows.forEach((row) => {
+          const existingLot = getExistingLot(row.lot);
+          if (existingLot && (row.expiration_date !== existingLot.expiration_date)) {
+            promises.push(Lots.update(existingLot.uuid, { expiration_date : row.expiration_date }));
```
jniles, line 335: Are these individual HTTP requests?  It might make sense to just do this when we submit data to the server, rather than try to individually send HTTP requests for each lot.

I think this is probably a performance optimization for later though.  We can merge this in and maybe make improvements if we feel the need to later.

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -309,12 +325,26 @@ function StockDefineLotsModalController(
...
+          }
+        });
+      }
+      Promise.all(promises)
```
jniles: We need to ensure that, if this delays the modal (imagine editing 1000 lots over a slow network), we have a loading indicator. 

Could you transition the `lots.modal.html` file to using `bh-submit` with a `bh-loading-button`?  Then you can just return this Promise and you'll get a loading indicator for free.

jniles (changes requested):
@jmcameron I've made one comment about a loading indicator that should be addressed.  Since we are doing asynchronous work with little guarantee on network performance, I can imagine users feeling frustrated without some level of indication that HTTP requests are being processed.

client/src/modules/stock/entry/modals/lots.modal.js
jmcameron: @jniles Please take a look at my fix for this.  If it okay and you think it is ready to release, let me know and I'll rebase and release.

jniles: You need to return this `Promise.all()` for it to have the intended effect.  The magic happens because the `submit()` method must return a promise.

client/src/modules/stock/entry/modals/lots.modal.html
```diff
@@ -105,8 +121,8 @@
       FORM.BUTTONS.CLOSE
     </button>
 
-    <button type="submit" class="btn btn-primary" ng-click="$ctrl.submit(FindForm, $event)" data-method="submit" translate>
-      FORM.BUTTONS.SUBMIT
-    </button>
+    <bh-loading-button loading-state="$ctrl.$loading">
```
jniles: This needs to be `FindForm.$loading`.  The loading is set on the form.

jniles (commented):
Just a couple of changes.

client/src/modules/stock/entry/modals/lots.modal.html
jmcameron: How can I return the Promise.all(promises) and still execute the saveSetting(),  Instance.close() afterward?

jniles: My language was imprecise. You need to return the promise chain that starts with Promise.all(). Just out a return statement in front of it.

jmcameron: I'm obviously missing something here!  I tried that but VS Code complained.  See below
image

Are you suggesting that I can get rid of the .then() afterwards?

jniles: No, the error here is because you have mixed `return` values.  It's a style thing, not a real error.

There is an earlier `return` above that returns nothing during an `if` check.  Just add `return 0;` to it, and ESLint will shut up.

jniles (approved):
Changes look good to me.  Tested and working on IMCK's database.

bors r+

PR #5770, opened 2021-06-23
Feature stock entry report currency

### LINKED ISSUE #5701: Stock Entry Report should allow you to set a currency
In keeping with the recent change to the expiration report (#5698) and stock exit report (#5693), we should allow users to pick a currency on the stock exit report.  In principle, all reports should provide the option to switch currencies on the fly.  The currency select would be one of the options below:

image

lomamech:
- Set currency in stock entry report
- Udpate Entry report display currency value

closes #5701

### DIFF
--- a/client/src/modules/reports/generate/stock_entry/stock_entry.config.js
+++ b/client/src/modules/reports/generate/stock_entry/stock_entry.config.js
@@ -22,7 +22,7 @@ function StockEntryConfigController($sce, Notify, SavedReports, AppCache, report
   vm.previewGenerated = false;
   vm.onEntryTypeChange = onEntryTypeChange;
 
-  vm.currency_id = Session.enterprise.currency_id;
+  vm.reportDetails.currencyId = Session.enterprise.currency_id;
 
   // check cached configuration
   checkCachedConfiguration();
--- a/client/src/modules/reports/generate/stock_entry/stock_entry.config.js
+++ b/client/src/modules/reports/generate/stock_entry/stock_entry.config.js
@@ -35,7 +35,7 @@ function StockEntryConfigController($sce, Notify, SavedReports, AppCache, report
   };
 
   vm.onSelectCurrency = currency => {
-    vm.reportDetails.currency_id = currency.id;
+    vm.reportDetails.currencyId = currency.id;
   };
 
   vm.clear = key => {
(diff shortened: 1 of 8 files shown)


lomamech:
image

client/src/modules/reports/generate/stock_entry/stock_entry.html
```diff
@@ -39,6 +39,11 @@ <h3 translate>REPORT.STOCK.ENTRY_REPORT</h3>
             required="true">
           </bh-date-interval>
 
+          <bh-currency-select
+            currency-id="ReportConfigCtrl.currency_id"
```
jniles: I think this needs to be `ReportConfigCtrl.reportDetails.currencyId`.  That way, it matches the saved value, and doesn't reset to the enterprise currency each time.

client/src/modules/reports/generate/stock_entry/stock_entry.config.js
```diff
@@ -22,6 +22,8 @@ function StockEntryConfigController($sce, Notify, SavedReports, AppCache, report
   vm.previewGenerated = false;
   vm.onEntryTypeChange = onEntryTypeChange;
 
+  vm.currency_id = Session.enterprise.currency_id;
```
jniles: Can you set this to 
```js
vm.reportDetails.currencyId = Session.enterprise.currency_id;
```
instead?  That way, the component is preloaded and can be overridden by a cached currency.

server/controllers/stock/reports/stock/entry_report.js
```diff
@@ -27,87 +27,103 @@ function stockEntryReport(req, res, next) {
...
   const collection = {};
 
-  // entry to purchase
+  // entryFromPurchase: [], entryFromIntegration: [], entryFromDonation: [], entryFromTransfer
+
+  // get stock entry from purchase
   collection.entryFromPurchase = formatAndCombine(entryCollection.entryFromPurchase);
 
-  // entry to integration
+  // get stock entry from integration
```
jniles, line 74: 👍

jniles (commented):
Code looks pretty good.  I haven't tested this on the live server, but just had some quick feedback about the `reportDetails` cache.

jniles (approved):
Looks good to me!

bors r+

PR #5774, opened 2021-06-25
Use of weighted average cost

### LINKED ISSUE #5401: Implemented Weighted Average Cost (WAC) / Cout Moyenne Pondere (CMP) algorithm for stock exits
Actually, the unit cost of lots during a stock exit is related to the lot unit cost which causes a mismatch with the accounting value of the stock when there is multiple unit cost for a unique lot.

The proposal is to have an average unit cost for lots with multiple values during the stock exit.

the goal is to have the same value in the stock value report and the account report each time

mbayopanda:
This PR allows to use the WAC value : 
- In the stock exit
- In the stock entry

In the stock entry for purchases operation, this PR adds a message to inform about WAC value

image

In the stock entry for integration or donation, if the WAC value exists, we will use it : 

image

In the stock entry for transfer movements, we are still using the value defined during the exit (in our case WAC value)

Feel free to inform me about other places to use WAC.

In order to support the same WAC on all depot, we introduce a concept of *cost_regulator* depot, a cost regulator depot is a depot in which WAC calculation are based which will represent the WAC of the enterprise, for all not cost regulator depot their stock sheet display in each line the cost based on the movement.

And we introduce this new line : 
image

Closes #5401


mbayopanda:
@jniles Semaphore doesn't run the up to date `build-database` file, here is the one that Semaphore runs : 
image

It's still running on a commit which doesn't exists

jniles:
It is running `./build-stock-database.sh` for the `yarn test:integration:stock` routines.  Is that what is confusing?

client/src/modules/stock/entry/entry.js
```diff
@@ -431,6 +431,15 @@ function StockEntryController(
         .then((lots) => {
           item.availableLots = lots;
           item.candidateLots = lots.filter(lot => !lot.expired);
+
+          return vm.depot && vm.depot.uuid ? Stock.inventories.read(null, {
```
jmcameron: Could you just get the single inventory item already retrieved on line 677?
> const inventory = inventoryStore.get(line.inventory_uuid);

client/src/modules/stock/entry/entry.js
```diff
@@ -658,24 +667,35 @@ function StockEntryController(
...
+        line.tracking_expiration = inventory.tracking_expiration;
+        setInitialized(line);
+
+        return vm.depot && vm.depot.uuid ? Stock.inventories.read(null, {
+          inventory_uuid : line.inventory_uuid,
+          depot_uuid : vm.depot.uuid,
+        }) : {};
+      })
+      .then(stockInventory => {
+        if (!stockInventory.length) { return; }
```
jmcameron: When would this occur?  And should we warn the user when it does?

client/src/modules/stock/exit/exit.js
```diff
@@ -556,7 +556,7 @@ function StockExitController(
       inventory_uuid : row.inventory.inventory_uuid, // needed for tracking consumption
       uuid : row.lot.uuid,
       quantity : row.quantity,
-      unit_cost : row.lot.unit_cost,
+      unit_cost : row.inventory.wac,
```
jmcameron, line 575: I wonder if we should use the WAC and also save the stated unit cost in case it is needed for audit and diagnostic purposes?

jmcameron (commented):
I looked through the code and made a few comments.  Basically the code looks fine. 
I tried running stock entry and exit and they seem to work---although I could not see where the WAC changed how it worked.

However, I suggest inserting a more detailed description into the main documentation (and the PR description) describing what the WAC is, how it is computed, and how it us used.

jmcameron:
By the way, I made some notes about WAC for myself a while back and I just reviewed them.  One thing that I noted (probably based a chat with @jniles) is that "Fiche de Stock that does AWC correctly".    I did not compare them, but I think that would be a good idea.

Also, adding some tests for this would be really helpful.  The test could even compare the computations with the Fiche de Stock calculations (some refactoring might be required).

mbayopanda:
Thank you @jmcameron for the review

client/src/modules/stock/entry/entry.js
mbayopanda: The `inventoryStore.get(line.inventory_uuid);` function allows only to get inventory metadata it means without consideration of stock information (quantity in depot or unit cost or wac); in our case we need to know stock information about that given inventory.

client/src/modules/stock/entry/entry.js
mbayopanda: The `Stock.inventories.read(null, { inventory_uuid : line.inventory_uuid, depot_uuid : vm.depot.uuid })` function returns an array; the case where `!stockInventory.length` can be true is when the inventory_uuid doesn't exist in the database or when there is nothing as stock in the database for the given depot_uuid; All these cases are corner cases because in normal usage we think these cases cannot happen.

I agree with the idea to warn the user when it happens

client/src/modules/stock/entry/entry.js
jmcameron: ok

mbayopanda:
@jniles @jmcameron changes landed

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -2,21 +2,25 @@ angular.module('bhima.controllers')
...
+  Currencies, Notify, bhConstants, EntryForm,
+  $translate, Focus, ExchangeRate, Lots,
 ) {
   const vm = this;
 
   const cache = new AppCache('StockEntryModal');
 
   // initialize the form instance
+  if (Data.stockLine.wac) {
+    Data.stockLine.unit_cost = Data.stockLine.wac;
```
jmcameron, line 21: I suggest we leave this for @jniles to review.  I think that we should be saving the unit_cost for reference purposes.  So I think we should add a new lot field for the WAC and not overriding the old unit_cost.

jmcameron (commented):
I looked over the changes.
I still think we should be saving the unit_cost in the lot for reference purposes.
I do not think this PR should be released until Jonathan Niles has had a chance to look over it and think about it. 
I admit I do not fully understand what this change is supposed to do.  That is part of the reason that I suggested adding some documentation about it somewhere.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,145 @@ CREATE PROCEDURE GetAMC(
...
+  DECLARE v_line_quantity INT(11);
+  DECLARE v_line_unit_cost DECIMAL(19, 4);
+  DECLARE v_line_is_exit TINYINT(1);
+
+  DECLARE cursor_all_movements CURSOR FOR
+    SELECT sm.quantity, sm.unit_cost, sm.is_exit
+    FROM stock_movement AS sm
+    JOIN lot AS l ON l.uuid = sm.lot_uuid
+    JOIN inventory AS inv ON inv.uuid = l.inventory_uuid
+    JOIN document_map AS map ON map.uuid = sm.document_uuid
```
jniles: I don't think you need the `document_map` here.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,145 @@ CREATE PROCEDURE GetAMC(
...
+  OPEN cursor_all_depots;
+
+  loop_cursor_all_depots : LOOP
+    FETCH cursor_all_depots INTO v_depot_uuid;
+
+    IF v_cursor_finished = 1 THEN
+      LEAVE loop_cursor_all_depots;
+    END IF;
+
+    IF _date IS NOT NULL THEN
```
jniles: Nice 👍

server/controllers/stock/index.js
```diff
@@ -1387,6 +1399,8 @@ async function createAggregatedConsumption(req, res, next) {
       });
     });
 
+    trx.addQuery('CALL RecomputeStockValue(NULL);');
```
jniles, line 1413: By calling this so many times, we have increased the time it takes to make a stock exit by at least 3X.  At Vanga, we go from ~3 seconds to exit stock to ~11 seconds.

Is there a way we can limit this to only the inventory used?   Or only the dates?  Why do we need to recompute stock value on every exit like this?

jniles (commented):
I've taken an initial look at this, and I think we need a more performant approach.  As I try to understand more about the algorithm, I'll be able to give better advice, but it looks like we may want to store the current value of stock someplace to speed up the calculation.  Right now, we seem to be recalculating it on every movement and it crashes performance.

mbayopanda:
> 
> 
> I've taken an initial look at this, and I think we need a more performant approach. As I try to understand more about the algorithm, I'll be able to give better advice, but it looks like we may want to store the current value of stock someplace to speed up the calculation. Right now, we seem to be recalculating it on every movement and it crashes performance.

The WAC is stored in this table:
```sql
CREATE TABLE  `stock_value` (
    `depot_uuid` BINARY(16) NOT NULL,
    `inventory_uuid` BINARY(16) NOT NULL,
    `date` DATE NOT NULL,
    `quantity` INT(11) NOT NULL,
    `wac` DECIMAL(19,4) NOT NULL,
    KEY `depot_uuid` (`depot_uuid`),
    KEY `inventory_uuid` (`inventory_uuid`),
    INDEX `depot_inventory` (`depot_uuid`, `inventory_uuid`),
    INDEX `date` (`date`),
    CONSTRAINT `stock_value__depot` FOREIGN KEY (`depot_uuid`) REFERENCES `depot` (`uuid`),
    CONSTRAINT `stock_value__inventory` FOREIGN KEY (`inventory_uuid`) REFERENCES `inventory` (`uuid`)
```

We can remove all these calls `CALL RecomputeStockValue(NULL);` for stock exit, however my concerns was about anterior date movements, we must have a way to recompute the stock value for an exit (custom anterior date) which occurs before a stock entry; Maybe I was overthinking about this case but it was the reason of these calls in stock exit

mbayopanda:
@jniles @jmcameron @lomamech when you'll have time, you can look at this PR.

All recommendations are welcome

server/models/migrations/next/migrate.sql
```diff
@@ -66,3 +66,28 @@ ALTER TABLE `purchase` MODIFY `cost` DECIMAL(19,8);
...
+ * @author: mbayopanda
+ * @date: 2021-06-23
+ */
+DROP TABLE IF EXISTS `stock_value`;
+CREATE TABLE  `stock_value` (
+    `inventory_uuid` BINARY(16) NOT NULL,
+    `date` DATE NOT NULL,
+    `quantity` INT(11) NOT NULL,
+    `wac` DECIMAL(19,4) NOT NULL,
+    KEY `inventory_uuid` (`inventory_uuid`),
```
jniles, line 301: Can you make this a foreign key on the `inventory` table?

client/src/modules/inventory/inventory.service.js
```diff
@@ -42,6 +42,12 @@ function InventoryService(
       .then(service.util.unwrapHttpResponse);
   };
 
+  // return the global WAC for the given inventory
+  service.inventoryWac = (uuid) => {
+    return service.$http.get(`/inventory/wac/${uuid}`)
```
jniles: Can you make this into `inventory/:uuid/wac`?  It's more REST-like.

client/src/modules/stock/entry/entry.js
```diff
@@ -432,6 +432,18 @@ function StockEntryController(
...
+        })
+        .then(inventoryWacDetails => {
+          if (!inventoryWacDetails) { return; }
+          item.wacValue = inventoryWacDetails.wac;
+
+          // in case of transfer reception
+          // we force the movement to be made with current WAC
+          // since there is no way to edit the unit cost in this case
+          if (vm.movement.entity.type === 'transfer_reception') {
+            item.unit_cost = inventoryWacDetails.wac;
```
jniles, line 445: Do you need to also recompute the `item.cost` too?

client/src/modules/stock/entry/entry.js
```diff
@@ -676,24 +688,32 @@ function StockEntryController(
...
+        line.label = inventory.label;
+        line.unit_cost = inventory.price;
+        line.quantity = 0;
+        line.cost = line.quantity * line.unit_cost;
+        line.expiration_date = entryDate;
+        line.unit = inventory.unit;
+        line.tracking_expiration = inventory.tracking_expiration;
+        setInitialized(line);
+
+        return Inventory.inventoryWac(line.inventory_uuid);
```
jniles, line 711: This looks like something that we should on the server before sending the lots or inventory items back to the client.

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -336,7 +342,7 @@ function StockDefineLotsModalController(
           }
         });
       }
-      return Promise.all(promises)
+      Promise.all(promises)
```
jniles: You might need to return this for the loading indicator to work, if there is a loading indicator on this modal.  If there isn't, we really should have one.

server/controllers/inventory/depots/extra.js
```diff
@@ -24,6 +24,50 @@ router.get('/movements/:fluxId', getDepotStockMovementsByFluxType);
...
+router.get('/inventories/:inventoryUuid/wac', getInventoryWac);
+router.get('/inventories/:inventoryUuid/sheet_wac', getInventorySheetWac);
+
+/**
+ * return inventory WAC from stock_value table
+ */
+async function getInventoryWac(req, res, next) {
+  try {
+    const binaryInventoryUuid = db.bid(req.params.inventoryUuid);
+    const queryRecompute = 'CALL RecomputeInventoryStockValue(?, ?);';
```
jniles: Do you need to call `Recompute` here?  It should be already computed, correct?

server/controllers/stock/core.js
```diff
@@ -138,8 +138,8 @@ function getLotFilters(parameters) {
   filters.custom('patientReference',
     'entity_uuid IN (SELECT uuid FROM entity_map WHERE text = ?)');
 
-  filters.period('defaultPeriod', 'date');
-  filters.period('period', 'date');
+  filters.period('defaultPeriod', 'date', 'm');
```
jniles, line 142: 👍

server/controllers/stock/core.js
```diff
@@ -1010,12 +1019,22 @@ async function getInventoryMovements(params) {
       const newQuantity = line.quantity + stockQuantity;
       // fix negative value disorder
       // ignoring negative stock value by setting them to movement value for exit
-      const newValue = (stockValue < 0)
-        ? (line.unit_cost * line.quantity)
-        : (line.unit_cost * line.quantity) + stockValue;
+      let newValue;
+      let newCost;
+      if (isCostRegulatorDepot) {
```
jniles: I'm not sure the cost regulator depot idea makes sense.  Can't we just use the stock entries from only DONATION, PURCHASE, and INTEGRATION to get the value?  Why do we need to consider the depot?

server/controllers/stock/index.js
```diff
@@ -143,6 +143,10 @@ async function createStock(req, res, next) {
       transaction.addQuery('CALL PostStockMovement(?)', [postingParams]);
     }
 
+    if (!isExit) {
+      transaction.addQuery('CALL RecomputeStockValue(NULL);');
```
jniles, line 147: I don't think we need to compute this for the stock entry from transfer.

We might need to look up the previous record to give the unit cost associated with that lot for the exit by transfer.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,125 @@ CREATE PROCEDURE GetAMC(
...
+  CLOSE cursor_all_movements;
+
+  /* write the line in the database */
+  DELETE FROM `stock_value` WHERE `inventory_uuid` = _inventory_uuid;
+  INSERT INTO `stock_value` VALUES (_inventory_uuid, _date, v_quantity_in_stock, v_wac);
+
+END $$
+
+DROP PROCEDURE IF EXISTS RecomputeDepotStockValue$$
+CREATE PROCEDURE RecomputeDepotStockValue(
```
jniles: Why is this by depot?  Can we just ignore the depot_uuid?

In fact, it doesn't look like you use this at all.  Can we just remove it?

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,125 @@ CREATE PROCEDURE GetAMC(
...
+      LEAVE loop_cursor_all_movements;
+    END IF;
+
+    IF v_line_is_exit <> 0 THEN 
+      SET v_is_exit = -1;
+    ELSE 
+      SET v_is_exit = 1;
+    END IF;
+
+    IF v_line_is_exit = 0 AND v_quantity_in_stock > 0 THEN
```
jniles, line 738: We should probably put some comments here describing how this works.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,125 @@ CREATE PROCEDURE GetAMC(
...
+    SELECT sm.quantity, sm.unit_cost, sm.is_exit
+    FROM stock_movement AS sm
+    JOIN lot AS l ON l.uuid = sm.lot_uuid
+    JOIN inventory AS inv ON inv.uuid = l.inventory_uuid
+    JOIN depot d ON d.uuid = sm.depot_uuid
+    WHERE
+      d.is_cost_regulator = 1
+      AND inv.uuid = _inventory_uuid
+      AND DATE(sm.date) <= DATE(_date)
+    ORDER BY inv.text, DATE(sm.date), sm.created_at ASC;
```
jniles: I don't see any reason to order by `inv.text`.  All we care about here are dates.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,125 @@ CREATE PROCEDURE GetAMC(
...
+
+  DECLARE v_line_quantity INT(11);
+  DECLARE v_line_unit_cost DECIMAL(19, 4);
+  DECLARE v_line_is_exit TINYINT(1);
+
+  DECLARE cursor_all_movements CURSOR FOR
+    SELECT sm.quantity, sm.unit_cost, sm.is_exit
+    FROM stock_movement AS sm
+    JOIN lot AS l ON l.uuid = sm.lot_uuid
+    JOIN inventory AS inv ON inv.uuid = l.inventory_uuid
```
jniles: You don't need to join on `inventory` at all.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,125 @@ CREATE PROCEDURE GetAMC(
...
+    IF v_line_is_exit <> 0 THEN 
+      SET v_is_exit = -1;
+    ELSE 
+      SET v_is_exit = 1;
+    END IF;
+
+    IF v_line_is_exit = 0 AND v_quantity_in_stock > 0 THEN 
+      SET v_wac = ((v_quantity_in_stock * v_wac) + (v_line_quantity * v_line_unit_cost)) / (v_line_quantity + v_quantity_in_stock);
+    ELSEIF v_line_is_exit = 0 AND v_quantity_in_stock = 0 THEN
+      SET v_wac = (v_line_unit_cost * 1); /* we must use the exchange rate */
```
jniles: It would be good to note that this only happens once, for the very first stock movement.

jniles (commented):
@mbayopanda I've made a series of comments.

It looks like your stored procedures have a lot of unused parts in them.  Could you clean them up a bit to help with reading?

It also seems to me that we shouldn't be recomputing the WAC as much as we are doing.  If I understand the algorithm, the only time we need to compute the WAC is when the `flux_id` is FROM_PURCHASE, FROM_DONATION, or FROM_ADJUSTEMENT.

If that is the case, then we can remove all references to the `depot` from these queries.  Do you agree?

client/src/modules/stock/entry/entry.js
mbayopanda: In case of transfer between depots, as discussed last time, we have to entry all stock from other depots in current WAC

server/models/migrations/next/migrate.sql
mbayopanda: And remove this line ? 
```sql
CONSTRAINT `stock_value__inventory` FOREIGN KEY (`inventory_uuid`) REFERENCES `inventory` (`uuid`)
```

server/controllers/stock/core.js
mbayopanda: The reason for that is restrict calculation of WAC to certain depots (WAC calculation is only based on entries PURCHASE, DONATION, INTEGRATION); if remove the depot filter (is_cost_regulator), in case a less important depot perform a stock entry (such as INTEGRATION) with bad cost it will affect the entire enterprise WAC value.

mbayopanda:
Thank you @jniles for the review 🙏

server/controllers/stock/core.js
jniles: I think I understand, but I'm not sure I agree.  Shouldn't the value of the stock from integration at any depot affect the cost of stock?  It seems like we'll just ignore certain depots and let users put whatever they want as the value of stock in those depots when it isn't true!

client/src/modules/stock/entry/entry.js
jniles: No, but it looks like `unit_cost` changes, but `item.cost` doesn't.  I think they both should, right?

server/models/migrations/next/migrate.sql
jniles: Ah, sorry.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,135 @@ CREATE PROCEDURE GetAMC(
     _initial_quantity AS quantity_at_beginning;
 END $$
 
+/* 
+ * RecomputeInventoryStockValue
+ * This procedure recompute the stock value for a given inventory
```
jmcameron: Typo:  recompute -> recomputes

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,135 @@ CREATE PROCEDURE GetAMC(
...
+    END IF;
+
+    /*
+      WAC calculation is performed for new entries
+
+      v_quantity_in_stock will contains cumulative quantity for our movements
+      in case of entry v_quantity_in_stock will be incremented else v_quantity_in_stock
+      will keep its last value, the v_quanitity_in_stock is initialized with 0
+
+      WAC = (current stock value + the value of the new entry) / the final quantity
```
jmcameron, line 732: ok

server/models/migrations/next/migrate.sql
```diff
@@ -0,0 +1,24 @@
...
+) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+/** Recompute stock values for all depots */
+/** All databases in production must run this script */
+CALL RecomputeStockValue(NULL);
+
+/* author: mbayopanda
+ * @date: 2021-08-05
+ */
+CALL add_column_if_missing('depot', 'is_cost_regulator', ' TINYINT(1) UNSIGNED NULL DEFAULT 0');
```
jmcameron: Shouldn't this be done before RecomputeStockValue(NULL)?

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,135 @@ CREATE PROCEDURE GetAMC(
...
+
+      Since all entry are made in enterprise currency we do not have to do
+      conversion here, so the wac is based on movement unit cose * 1
```
jmcameron: This comment could be a little clearer.  I don't understand the last part.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,135 @@ CREATE PROCEDURE GetAMC(
...
+    */
+    IF v_line_is_exit = 0 AND v_quantity_in_stock > 0 THEN 
+      SET v_wac = ((v_quantity_in_stock * v_wac) + (v_line_quantity * v_line_unit_cost)) / (v_line_quantity + v_quantity_in_stock);
+    ELSEIF v_line_is_exit = 0 AND v_quantity_in_stock = 0 THEN
+      SET v_wac = (v_line_unit_cost * 1);
```
jmcameron, line 741: A comment here would be helpful.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,135 @@ CREATE PROCEDURE GetAMC(
...
+      AND DATE(sm.date) <= DATE(_date)
+    ORDER BY DATE(sm.date), sm.created_at ASC;
+  
+  DECLARE CONTINUE HANDLER FOR NOT FOUND SET v_cursor_all_movements_finished = 1;
+
+  OPEN cursor_all_movements;
+
+  loop_cursor_all_movements : LOOP
+    FETCH cursor_all_movements INTO v_line_quantity, v_line_unit_cost, v_line_is_exit;
+
```
jmcameron, line 714: I suggest changing the variable 'v_is_exit' to 'v_is_entry' (and fix the logic, of course).  It would make the succeeding parts easier to understand.  For instance, line 714 would then read:

   SET v_quantity_in_stock = v_quantity_in_stock + (v_line_quantity * v_is_entry);

where v_is_entry = 1 for entries and then it is clear that items entering are increasing the quantity in stock.

server/models/procedures/stock.sql
```diff
@@ -643,4 +643,135 @@ CREATE PROCEDURE GetAMC(
...
+    END IF;
+
+    SET v_quantity_in_stock = v_quantity_in_stock + (v_line_quantity * v_is_exit);
+    SET v_line_quantity = v_quantity_in_stock;
+
+  END LOOP loop_cursor_all_movements;
+
+  CLOSE cursor_all_movements;
+
+  /* write the line in the database */
```
jmcameron: Change "Write" to "Update" ?

test/integration-stock/depots.js
```diff
@@ -6,6 +6,31 @@ const helpers = require('./helpers');
 describe('(/depots) The depots API ', () => {
   const { principal } = helpers.data.depots;
 
+  /**
```
jmcameron, line 9: Great to have tests for this!  Thanks!

jmcameron (commented):
I looked over the code and made a few cosmetic comments.

When I try the stock entry, I do not see the WAC value as is shown in image in the comment above
https://github.com/IMA-WorldHealth/bhima/pull/5774#issue-677825951

I tried this with various types of Entry types including "from purchase".   Was this feature removed after the initial PR description?

jniles:
@mbayopanda, when you get a chance, can you incorporate the changes suggested by @jmcameron?  It would be nice to get this merged at some point in the near future :+1:

jniles:
So, I would like to get this in.  I think it's been long enough, and we aren't making progress on it.  If we merge it, we'll be able to make progress as a team.  Especially if we merge this before https://github.com/IMA-WorldHealth/bhima/pull/6051, we'll be able to clean up our stock data quite effectively.

@jmcameron you gave this your review.  Can we merge it in before #6051?  It will force us to address all the shortcomings and do a proper computation of the WAC for the future.

jmcameron:
> @jmcameron you gave this your review. Can we merge it in before #6051? It will force us to address all the shortcomings and do a proper computation of the WAC for the future.

@jniles I'm fine with that.   Let me take one more look at it.

jmcameron (commented):
I'm finding it difficult to follow which suggestions have been applied to the code and which ones have not.   Please do the following:
- Go through the suggestions and add comment to each whether they have been done or not.  If they have not been done, please explain why.    I have no problem with ignoring some of the suggestions, if appropriate.
- After that, please rebase the PR so it is ready to merge

This is a really good PR and we would all like to get it merged soon!

client/src/modules/stock/entry/entry.js
mbayopanda: DONE

client/src/modules/inventory/inventory.service.js
mbayopanda: DONE

client/src/modules/stock/entry/entry.js
mbayopanda: You're right but currently not yet done at the server side

client/src/modules/stock/entry/modals/lots.modal.js
mbayopanda: DONE

server/controllers/inventory/depots/extra.js
mbayopanda: DONE

server/controllers/stock/core.js
mbayopanda: DONE

server/controllers/stock/index.js
mbayopanda: DONE, We are now using this condition: 
```js
if (!isExit && params.flux_id !== core.flux.FROM_OTHER_DEPOT) {
    transaction.addQuery('CALL RecomputeStockValue(NULL);');
}
```

server/models/procedures/stock.sql
mbayopanda: DONE

server/models/procedures/stock.sql
mbayopanda: DONE

server/models/procedures/stock.sql
mbayopanda: DONE

server/models/procedures/stock.sql
mbayopanda: DONE, currently on the line `679`

server/models/migrations/next/migrate.sql
mbayopanda: Currently we are not supporting this `is_cost_regulator` column, so we removed that

server/models/procedures/stock.sql
mbayopanda: DONE, added more details

server/models/procedures/stock.sql
mbayopanda: Commented above (before the IF)

server/models/procedures/stock.sql
mbayopanda: DONE

jniles (approved):
LGTM.  Thanks for your extremely hard work on this @mbayopanda !

bors r+

PR #5776, opened 2021-06-30
Add shipping and handling costs to purchase orders

### LINKED ISSUE #5225: Add "Shipping and Handling Costs" to Purchase Order
We need a way to add shipping and handling to the purchase order.  Shipping & Handling is important because it increases the value of the inventory (e.g. the cost of the inventory).  We can either calculate this as a percentage of the order value (i.e. 5% of total value) or a fixed cost (it costs $20 to deliver this inventory).

We'll need to add a section to the purchase order where the user can put in either a percentage or a fixed cost to the order.  This would not change the prices of the inventory items displayed, but would add a final section to the order "shipping and handling".  We would then divide this across the entire order.

This feature was requested by our friends at IMCK.

### LINKED ISSUE #5495: How do we account for transport/tax/fees on purchase?
We currently don't have a good way (in either OHADA or normal accounting) to account for transportation costs.  The transport costs should show up in the value of the stock and the cost of goods sold [source](https://www.accountingcoach.com/blog/transportation-in-cost).

For OHADA, this would presumably enter into all the stock movements.  The real question is where do we configure the accounts associated with these costs?  They shouldn't be billed to the supplier - they would presumably go into a payable account similar to that one though.

This issue is a placeholder until we can come up with a decent design for it.

jmcameron:
Add optional shipping and handling to the creation of purchase requests and for stock entry via purchase order.  
This update also supports using any enterprise-supported currency for both operations.  

The purchase order receipt document will include information about the shipping and handling costs, if present.  If no shipping and handling costs are specified, then the reports will look like they always have.

The stock entry via purchase order receipt document clearly spells out the original unit costs of each inventory item and the adjusted cost (if shipping and handling costs are non-zero).   This document will add an annotation if the order is not complete. If the purchase order is not in the enterprise currency, the equivalent total values for the original order and the received order are also shown in the enterprise currency.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5225, https://github.com/IMA-WorldHealth/bhima/issues/5495

If shipping and handling costs are included in the purchase order, that cost is distributed among the individual articles in the order.  For a description on how this is done, see the Issue: https://github.com/IMA-WorldHealth/bhima/issues/5225#issuecomment-869768580

*TESTING*
- Use whichever database you wish
- Since you will be creating multiple purchase orders, in the Stock > Stock Settings, set the purchase orders to be confirmed automatically.
- Note that there are several variations on the stock entry via purchase order report you should test, if you have time.  You will need new purchase orders for these as you go.   
   - A full order in enterprise currency without shipping and handling costs
   - A full order in enterprise currency with shipping and handling costs
   - A partial order in enterprise currency without shipping and handling costs
   - A partial order in enterprise currency with shipping and handling costs
   - A full order in some other currency without shipping and handling costs
   - A full order in some other currency with shipping and handling costs
   - A partial order in some other currency without shipping and handling costs
   - A partial order in some other currency with shipping and handling costs


client/src/modules/roles/roles.html
```diff
@@ -5,7 +5,7 @@
       <li class="title" translate>TREE.ROLE_MANAGEMENT</li>
     </ol>
 
-    <div class="toolbar" ng-show="bh-has-permission="grid.appScope.bhConstants.actions.EDIT_ROLES"">
+    <div class="toolbar" bh-has-permission="grid.appScope.bhConstants.actions.EDIT_ROLES">
```
mbayopanda, line 8: 👍🏽

client/src/modules/stock/entry/entry.js
```diff
@@ -553,9 +554,50 @@ function StockEntryController(
...
+      // console.log("LOTS: ", movement.lots);
+      // console.log("PO TOTAL COST: ", costPO + vm.originalPO[0].shipping_handling);
+      // console.log("DELIV TOTAL COST: ", totalCostDelivered);
+      // movement.lots.forEach(lot => {
+      //   console.log("ORIG lot unit_cost: ", lot.unit_cost);
+      // });
+
+      // Adjust the article unit price by distributing the shipping+handling costs
+      movement.lots.forEach(lot => {
+        lot.unit_cost = ((lot.unit_cost * lot.quantity) * (1.0 + SHRatio)) / lot.quantity;
```
mbayopanda, line 571: 👍🏽 I spent time with my calculator to figure out this algo against mine, and your algo is much better 😄 

Good repartition 👍🏽

mbayopanda (approved):
The feature works as expected 
image

Good job 👍🏽

client/src/modules/purchases/create/createUpdate.html
```diff
@@ -124,6 +124,25 @@
...
+                    ng-class="{'has-error' : PurchaseOrderForm.$submitted && PurchaseOrderForm.shipping_handling.$invalid }">
+                    <label class="control-label">
+                      <span translate>FORM.LABELS.SHIPPING_HANDLING_COST</span> ({{PurchaseCtrl.rate.currency.symbol}})</label>
+                    <input
+                      ng-model="PurchaseCtrl.order.details.shipping_handling"
+                      name="shipping_handling"
+                      class="form-control"
+                      type="number"
+                      min="0"
+                      onkeypress="return event.charCode != 45"
```
jniles: Why do you have this onkeypress event?

client/src/modules/stock/entry/entry.js
```diff
@@ -553,9 +554,50 @@ function StockEntryController(
...
+
+      // // ??? TEMPORARY DIAGNOSTIC CODE
+      // // Compute the cost of the order, as delivered (in PO currency), not including shipping+handling
+      // const costDelivered = movement.lots.reduce((accum, lot) => accum + lot.unit_cost * lot.quantity, 0.0);
+
+      // Compute the ratio of the original PO shipping+handling costs to the total cost of lots
+      // (Note that the total shipping_handling is copied to each lot in originalPO)
+      const SHRatio = vm.originalPO[0].shipping_handling / costPO;
+
+      // // ??? TEMPORARY DIAGNOSITIC CODE
```
jniles: Can this be removed?

client/src/modules/stock/entry/entry.js
```diff
@@ -553,9 +554,50 @@ function StockEntryController(
...
+      //   console.log("ORIG lot unit_cost: ", lot.unit_cost);
+      // });
+
+      // Adjust the article unit price by distributing the shipping+handling costs
+      movement.lots.forEach(lot => {
+        lot.unit_cost = ((lot.unit_cost * lot.quantity) * (1.0 + SHRatio)) / lot.quantity;
+      });
+
+      // // ??? TEMPORARY DIAGNOSITIC CODE
+      // movement.lots.forEach(lot => {
```
jniles: This as well?

client/src/modules/stock/entry/entry.js
```diff
@@ -553,9 +554,50 @@ function StockEntryController(
...
-    // fix the unit price, if necessary
+    const exchangeRate = Exchange.getCurrentRate(vm.currencyId);
+
+    // If there are shipping and handling charges, distribute them to the individual lot articles
+    if (vm.originalPO[0].shipping_handling) {
+
+      // Compute the cost of the original order (in PO currency), not including shipping+handling
+      const costPO = vm.originalPO.reduce((accum, lot) => accum + lot.unit_price * lot.quantity, 0.0);
+
+      // // ??? TEMPORARY DIAGNOSTIC CODE
```
jniles: It looks like you need to repass through this js file and just clean up the comments.

client/src/modules/stock/entry/modals/findPurchase.modal.js
```diff
@@ -82,6 +85,9 @@ function StockFindPurchaseModalController(
     Exchange.read()
       .then(() => Purchase.search({ status_id : [CONFIRMED, PARTIALLY_RECEIVED] }))
       .then(purchases => {
+        purchases.forEach(p => {
+          p.total_cost = p.cost + p.shipping_handling;
```
jniles, line 89: Good 👍

server/controllers/stock/reports/stock/entry_purchase_receipt.js
```diff
@@ -66,16 +88,77 @@ async function stockEntryPurchaseReceipt(documentUuid, session, options) {
...
     barcode               : barcode.generate(key, line.document_uuid),
     voucher_reference     : voucherReference,
     autoStockAccountingEnabled,
   };
+
+  // Set up flag for handlebars
+  data.not_in_enterprise_currency = session.enterprise.currency_id !== data.details.p_currency_id;
+
+  // // ??? TEMPORARY DIAGNOSITIC CODE
+  // console.log('====================================================================');
```
jniles: More comments to clean up.

server/controllers/stock/reports/stock/entry_purchase_receipt.js
```diff
@@ -42,22 +46,40 @@ async function stockEntryPurchaseReceipt(documentUuid, session, options) {
...
 
+  const poUuid = rows[0].po_uuid;
+  const voucherReference = results[1][0] ? results[1][0].voucher_reference : null;
+
+  // Get the information about the original PO lots
+  const poSql = `
+     SELECT BUID(inventory_uuid) as inventory_uuid, quantity, unit_price
+     FROM purchase_item WHERE purchase_uuid = ?;
+     `;
+  const poLots = await db.exec(poSql, [db.bid(poUuid)]);
```
jniles: Can we rename this variable?   Let's call it `poInventoryArticles`.  It took me a couple double takes thinking about lots rather than inventory articles, when these are actually inventory articles.

server/controllers/stock/reports/stock/entry_purchase_receipt.js
```diff
@@ -66,16 +88,77 @@ async function stockEntryPurchaseReceipt(documentUuid, session, options) {
...
+  // // ??? TEMPORARY DIAGNOSITIC CODE
+  // console.log("ROWS: ", rows);
+
+  // Compute the total cost of all delivered articles (in PO currency)
+  // (This also includes distributed shipping+handling costs)
+  data.details.total_cost_deliv = sumCost;
+
+  // Compute the total cost of the original PO including shipping+handling (in PO currency)
+  data.details.p_total_cost = data.details.p_full_cost + data.details.p_shipping_handling;
+  data.details.p_total_cost_ent = data.details.p_total_cost / exRate; // Enterprise currency
```
jniles: In the posting journal/general ledger, we call the enterprise currency version `debit_equiv` and `credit_equiv`.  For consistency, I propose you call this `p_total_cost_equiv` as well.

server/controllers/stock/reports/stock/entry_purchase_receipt.js
```diff
@@ -66,16 +88,77 @@ async function stockEntryPurchaseReceipt(documentUuid, session, options) {
...
+
+  data.details.total_cost_deliv_ent = data.details.total_cost_deliv / exRate; // Enterprise currency
+
+  data.details.p_shipping_handling_ent = data.details.p_shipping_handling / exRate; // Enterprise currency
```
jniles: Good comments.

jniles (commented):
@jmcameron I've taken a look at the code.  This is awesome!  Well done.  The code is well commented and easy to follow.

I've made some comments on old comments that should be cleaned up.  I've also proposed some variable renames.  Once that is complete, this can be merged without any issue.

client/src/modules/purchases/create/createUpdate.html
jmcameron: I added this to prevent entering an minus sign, thus preventing negative values for shipping and handling.

client/src/modules/stock/entry/entry.js
jmcameron: I will remove these commented-out diagnostic items once the PR has been reviewed for correct operation.

client/src/modules/stock/entry/entry.js
jmcameron: I will remove these commented-out diagnostic items once the PR has been reviewed for correct operation.

client/src/modules/stock/entry/entry.js
jmcameron: I will remove these commented-out diagnostic items once the PR has been reviewed for correct operation.

client/src/modules/stock/entry/entry.js
jmcameron: I'll remove these.

client/src/modules/purchases/create/createUpdate.html
jniles: Why doesn't the "min" work?  It should block submission if any of the fields are incorrect.

jmcameron: It does work, but only when the input element gets verified.  This prevents negative values during entry.

server/controllers/stock/reports/stock/entry_purchase_receipt.js
jmcameron: Done

server/controllers/stock/reports/stock/entry_purchase_receipt.js
jmcameron: Done

server/controllers/stock/reports/stock/entry_purchase_receipt.js
jmcameron: Done

client/src/modules/purchases/create/createUpdate.html
jniles: In my view, just sticking to `min` is the best solution.  Currently, someone might think their keyboard is broken, rather than thinking that the data is invalid.  If you want to validate the data (and show a red indicator during entry), just change the `ng-class` on the `<div class="form-group">` to make the `has-error` constantly active, rather than wait for `PurchaseOrderForm.$submitted`.  It will light up red when the user enters invalid data, rather than silently ignoring their input.

However, pretty much all our forms work in the old way, where we define a data entry phase and then a review phase after submitting where anything wrong is marked in red.  I find it confusing if this is different for this specific form.  Is it really that much of a value add to change the behavior for this specific input?

jmcameron: Ok. Removed.

mbayopanda (approved):
The feature is working as expected : 
- Create purchase works fine with the shipping value
- Edit purchase works fine with the shipping value
- Calculation of shipping value for inventories is done correctly and works
- Stock entry from purchase works fine and calculation of shipping cost is also correct

There is an issue but I think it doesn't concern this PR, the `Analyse de commande d'achat` report is buggy 

image

Values in red circles are wrong

mbayopanda:
@jniles the PR is okay for me, you can merge it

jniles:
Thanks @mbayopanda .  The Purchase Analysis report has been broken for a while https://github.com/IMA-WorldHealth/bhima/issues/5429

@jmcameron can you do a final rebase and we can get this merged?

jmcameron:
> @jmcameron can you do a final rebase and we can get this merged?

Yes.  I'll do it after lunch, and fix the merge conflict issues.

jmcameron:
> There is an issue but I think it doesn't concern this PR, the `Analyse de commande d'achat` report is buggy
> 
> image
> 
> Values in red circles are wrong

Bruce, please add an issue for this (if you have not already).

jmcameron:
@bruce, please do one last sanity check on this.  I had to revert the translation files and I think I got my new items back in, but I would appreciate a sanity check.  Also if you have suggestions for improving the french related to this PR, please suggest them. Thanks!   If all is okay, per @jniles, I think you can go ahead and release it.  Thanks.

mbayopanda:
LGTM 👍🏽 

The feature is working as expected, the stock sheet is using the lot unit cost (lot unit cost + shipping). For an existing lot, we are using the unit cost of that existing lot.

bors r+

PR #5784, opened 2021-07-07
Fix console error on openSearchModal

### LINKED ISSUE #5773: Fix openSearchModal() error non-handling everywhere`
Many uses of the openSearchModal() have a problem when the user opens the dialog, does nothing, and then closes it.
For instance, code looks like this:

>   function openSearchModal() {
>     const filtersSnapshot = Journal.filters.formatHTTP();
>     Journal.openSearchModal(filtersSnapshot)
>       .then((changes) => {
>         Journal.filters.replaceFilters(changes);

But 'changes' here will be 'undefined' if the user closes the Search dialog without changing any search options.  Unfortunately, the call to replaceFilters(changes) (or similar uses of 'changes') may not handle 'undefined' (eg, applying a forEach() on it).  This should be addressed everywhere by either exiting immediately if changes===undefined, or making sure replaceFilters handles values of 'unbound' as its argument (or similar changes in other places).

This error prints an error message to the console, but does not usually otherwise affect normal software operation, so it might not have been noticed before.

mbayopanda:
This PR is related to #5773 

Close #5773

### DIFF
--- a/client/src/modules/account_statement/account_statement.ctrl.js
+++ b/client/src/modules/account_statement/account_statement.ctrl.js
@@ -58,7 +58,6 @@ function AccountStatementController(
       <span ng-show="row.groupHeader">{{ COL_FIELD | currency: row.treeNode.children[0].row.entity.currency_id }}</span>
     </div>`;
 
-
   // columns definition
   const columns = [{
     field            : 'trans_id',
@@ -243,6 +242,8 @@ function AccountStatementController(
       hidePostedOption : true,
     })
       .then(changes => {
+        if (!changes) { return; }
+
         AccountStatement.filters.replaceFilters(changes);
 
         AccountStatement.cacheFilters();
@@ -263,12 +264,6 @@ function AccountStatementController(
   };
 
   // exports zone =====================================================================
-
-  // format parameters
-  function formatExportParameters(type) {
-    const filters = AccountStatement.filters.formatHTTP(true);
-    return angular.extend(filters, { renderer : type || 'pdf', lang : Languages.key });
-  }
 
   vm.download = function download(type) {
     const filterOpts = AccountStatement.filters.formatHTTP();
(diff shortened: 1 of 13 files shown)


client/src/modules/account_statement/account_statement.ctrl.js
```diff
@@ -264,12 +265,6 @@ function AccountStatementController(
 
   // exports zone =====================================================================
 
-  // format parameters
```
jmcameron, line 267: Was this unused code?

client/src/modules/multiple_payroll_indice/multiple_payroll_indice.js
```diff
@@ -84,7 +84,7 @@ function MultiplePayrollIndiceController(
 
     MultiplePayroll.openSearchModal(filtersSnapshot)
       .then((changes) => {
-        if (!changes.length) { return null; }
+        if (!changes || !changes.length) { return null; }
```
jmcameron, line 87: Why is this one different from the others?

jmcameron (commented):
The changes look good.
I did not actually try them since I fixed one of these myself earlier and was pretty confident that they are done correctly.

client/src/modules/account_statement/account_statement.ctrl.js
mbayopanda: Yes it was

client/src/modules/multiple_payroll_indice/multiple_payroll_indice.js
mbayopanda: There was a test for array `if (!changes.length)...` so I have just merge both if condition
```js
if (!changes) { return null; }
if (!changes.length) { return null; }
```
into 
```js
if (!changes || !changes.length) { return null; }
```
While we are talking about, maybe the previous code had an error and the good test is just `if (!changes) { return null; }`

jniles (approved):
Looks good to me.  I've spot-tested these and they seem to work fine.

bors r+

PR #5785, opened 2021-07-08
Add showing lot usage in lots registry and articles in stock

### LINKED ISSUE #5292: Add "View Consumption Schedule" modal to the "Articles in Stock" registry
This feature depends on https://github.com/IMA-WorldHealth/bhima/pull/5226.

We currently have a "Consumption Schedule" where we assume that a depot will consume the lots in order of expiration date from soonest expiration date to the longest.  It would be nice to display this to the user in a graphical format - something like a [Gantt chart](https://www.wincalendar.com/Gantt-Chart).  This would allow the user to see which lot (on the Y axis) will be used during which time (on the X axis).  We could include cutoffs for their expiration dates so the user can see why they are unable to use all the quantity they think they have in stock, even if they have a large quantity in stock.

This would be linked from the Articles in Stock registry.

Here is a good example of how it should look:

https://cdpn.io/tutsplus/fullembedgrid/ZEzerNB?animations=run&type=embed

jmcameron:
Adds a new function in the action menu for the Stock > Lots registry and the Stock > Articles in Stock registry.
The new action item will display a modal with the suggested order to expend lots for all lots of the same stock article.
The lot usage shows tabular information about each lot in the sequence they should be used.  It also displays a timeline showing part of the same information graphically.

Note that the placement of the lot usage rectangles in the lower, graphical part of the display is based on treating all months as if they were of equal length (30.5 days).  Therefore there may be some very small errors in positioning and width of the items.  For 30/31 day months, the error will be less than one pixel.  But for February it will be more, but probably not very notiecable.

#### TESTING
- Suggest using IMCK for inititial tests and other databases for further testing
- In the Stock > Lots registry 
  - Try the new action menu item (last menu item)
  - Note that the lot you used for the action menu will be shown in light blue
  - Hover over one of the lot usage rectangles in the lower, graphical display to see a message about the lot usage.
  - Note that the initial quantity is shown in black on the left end of the lot rectangle and the red number on the right end of the lot rectangle is what quantity will be left when the lot expires.
  - Note the suggested reorder date
  - Try a lot that has several lots with the same label to see the waterfall/gantt chart display
- In the Stock > Articles in Stock
  - try the same function to see a similar display (no lots will be in blue since this function invoked from an inventory item, not a specific lot).
  - Try it with an inventory item that is all expired (you will need to enable that via the Default search filter).
- Please carefully review the French translations and suggest improvements

For the IMCK database, here are a few Inventory codes that produce displays that show most of the features of this PR:
- Normal inventory items:
    - 002002 Artesuna...
    - 100008 Amoxy...
    - 100022 Ceflacor...
-   Inventory items with CMM=0
    - 100013 Asprin..
    - 100044 Doxy...
    - 100001 Acetazol..

This PR includes various improvements based on discussions with J. Niles.


client/src/modules/stock/StockModal.service.js
```diff
@@ -229,7 +244,7 @@ function StockModalService(Modal) {
   }
 
   /** search transfer  */
-  function openFindTansfer(request) {
+  function openFindTransfer(request) {
```
mbayopanda, line 247: Good typo catch 👍🏽

mbayopanda (commented):
Thank you @jmcameron this is a useful feature. My concern is about this : 
consumption calendar

In the case above, we must have less than one month of consumption to start consuming the next lot, in this summary, calculation are based on the expiration of the first lot

client/src/modules/stock/lots/modals/schedule.modal.html
```diff
@@ -0,0 +1,88 @@
+<form
+  name="ModalForm"
+  data-modal="stock-lots-schedule"
+  ondragstart="return false;" ondrop="return false;"
```
jniles: Why do we need this drag/drop aliases?

client/src/modules/stock/lots/modals/schedule.modal.html
```diff
@@ -0,0 +1,88 @@
...
+  novalidate>
+
+  <div class="modal-header">
+    <h2 translate>LOTS_SCHEDULE.TITLE</h2>
```
jniles: To make this more uniform, it would be nice to use the `.headercrumb` class that we generally use in our modals.  It would also give you more space to work with in the modal itself.

client/src/modules/stock/lots/modals/schedule.modal.js
```diff
@@ -0,0 +1,129 @@
...
+        vm.lots.forEach(lot => {
+          // Process the lots and determine sequential start/end dates and other info
+          lot.start_date = new Date(runningDate);
+
+          // Compute when the lot runs out (based on adjusted start date)
+          lot.exhausted_date = Moment(lot.start_date).add(lot.quantity / lot.avg_consumption, 'months').toDate();
+
+          // Compute the end date for this lot
+          lot.end_date = new Date(lot.expiration_date);
+          if (lot.exhausted_date < lot.expiration_date) {
```
jniles: This logic is backwards - if the `expiration_date` is before the `exhausted_date`, then we need to make the `exhausted_date` the `expiration_date`, since we cannot use expired stock.

However, I still think it is best to keep both around and show the `end_date` as the date you will run out, but with X quantity wasted.  The interpretation being that the `end_date` will be the last day you would consume at the current consumption schedule.

jniles: image

Once we make that correction to the logic, we need to make sure the scenario above is possible.  I've just edited it to show a hypothetical scenario where the stock expires before it runs out.  The overlap is _unusable stock_.  While the institution won't be consuming it, it shows the number of months that they are missing out on, if only they better managed their stock.

client/src/modules/stock/lots/modals/schedule.modal.js
```diff
@@ -0,0 +1,129 @@
...
+          });
+          month = Moment(month).add(1, 'months').toDate();
+        }
+
+        // Get sorted list of the years and the number of months in each year
+        const allYears = vm.months.map(obj => obj.year);
+        const years = Array.from(new Set(allYears)).sort();
+        vm.years = [];
+        years.forEach(yr => {
+          vm.years.push({
```
jniles: I'm always a fan of `.map()` instead of `.forEach() push()`.  

```js
vm.years = years.map(yr => ({ year : yr, count : allYears.reduce((n, year) => n + (year === yr), 0), }));
```

jniles (changes requested):
This is really cool!  The concept is spot-on, but we are still rough around the edges.

It would be idea if this scaled to fit the modal so we didn't have as much scrolling to do. I don't know if that is possible, but it would be cool!

image

It looks like end date and expiry date is always the same.  The end date should be the date the stock runs out, while the expiry date should be the day that the stock expires.  In this modal, it would be useful to present those different pieces of information.

I've made some more detailed comments inline.  I'll swing around for another review once we have those details sorted out.

client/src/modules/stock/lots/modals/schedule.modal.html
jmcameron: Removed copy/paste debris

client/src/modules/stock/lots/modals/schedule.modal.html
jmcameron: Done.

client/src/modules/stock/lots/modals/schedule.modal.js
jmcameron: Made these fixes.  I had thought of the red unusable residual block so I was happy to do this.  I also did some adjustments on the spacing to make the dates work on the graphical display.

client/src/modules/stock/lots/modals/schedule.modal.js
jmcameron: Thanks for the reminder!

jmcameron:
> It would be idea if this scaled to fit the modal so we didn't have as much scrolling to do. I don't know if that is possible, but it would be cool!

> I've made some more detailed comments inline. I'll swing around for another review once we have those details sorted out.

I also rescaled things, used vertical text for months, etc, to get it on one screen without scrolling.   Currently only shows 36 months before scrolling is necessary.

server/controllers/stock/index.js
```diff
@@ -884,15 +884,14 @@ async function listLotsDepot(req, res, next) {
...
-    const queryTags = `
-      SELECT BUID(t.uuid) uuid, t.name, t.color, BUID(lt.lot_uuid) lot_uuid
-      FROM tags t
-        JOIN lot_tag lt ON lt.tag_uuid = t.uuid
-      WHERE lt.lot_uuid IN (?)
-    `;
-
     // if we have an empty set, do not query tags.
     if (data.length !== 0) {
+      const queryTags = `
```
jniles, line 889: 👍

server/controllers/stock/core.js
```diff
@@ -785,7 +785,7 @@ function computeLotIndicators(inventories) {
           lot.near_expiration = lot.expiration_date <= stockOutDate;
 
           // here, we are attempting to calculate if this quantity will be at risk of expiry
-          // given that all inventory is consummed in order of their expiration dates.
+          // given that all inventory is consumed in order of their expiration dates.
```
jniles, line 788: Good catch 👍

client/src/modules/stock/lots/registry.js
```diff
@@ -105,6 +105,14 @@ function StockLotsController(
       });
   };
 
+  // lot schedule modal
+  vm.openLotScheduleModal = (uuid, inventoryUuid, depotUuid) => {
+    Modal.openLotScheduleModal({ uuid, inventoryUuid, depotUuid })
+      .then((/* res */) => {
```
jniles, line 111: No need to have this `.then()` at all if that is the case.

jniles:
Note that we discussed this in person and there are few more changes to make.  Let me know when those have been done.

jmcameron:
@jniles I think this is ready for a near-final review.   Once you approve it, I'll rebase it and release it myself.

Also, please recall that I asked you to verify that the sort on line 49 of schedule.modal.js was necessary.

Thx

jniles (commented):
@jmcameron I've made some changes inline.  I caught a bug in the `moment().add()` call which doesn't take decimals.  I also didn't see a difference in the `sort()` for expiration date, so I pruned it.

Please give this a look over and see if you can catch anything wrong with my code.

From the commit message:

> MomentJS does not accept decimal points in the .add() method, but
transparently rounds the number.  This caused buggy calculations in the
stock consumption data, showing wasted values if the CMM was less than
have a month of stock.

> Additional changes:
>  1) Wasted values are now in red.
>  2) Removed extraneous code (sorting by expiration_date, empty .then()).
>  3) Added an auto dropup on the dropdown menu.

client/src/modules/stock/lots/registry.js
jmcameron: Thanks!

jniles (approved):
Thanks @jmcameron, you are right - my changes didn't allow for the final portion of the red to show.

image

Looks fixed now and looks good to me.  This is awesome.

bors r+

Issue #5786, opened 2021-07-09
Bug in stock exit


jmcameron:
Try this:
- Use IMCK db
- Do stock exit to Patient [PA.PAX.8989] Anasthasie Ntumba Ndaye
- Add stock 110010 - multivitamine sirop 100 ml
- Select only available lot:  MUT166
- Notice it is expired and yet we could select it AND that it is accepted as a valid stock item (green check mark on the left).  This should be prevented!
- See:

![image](https://user-images.githubusercontent.com/62145/125063683-ae198380-e0a7-11eb-8bcf-daf86dfb5354.png)

jniles:
Did this get closed in #5790?

jniles:
Can no longer reproduce on master.

Issue #5706, opened 2021-05-26
If a medicine has only expired stock, we should consider it as "stock out"


jniles:
@lomamech and I had an interesting discussion this morning.  If a medicine is expired, and we have no lots in stock, we should consider that as "out of stock" of that medicine.  Since the medicine cannot be used, it is effectively that in practice.  It seems certain health centers are masking their out-of-stock status by keeping expired medicine on the books, which it should really be removed.

@mbayopanda @jmcameron what do you think?

jmcameron:
This sounds good to me.   But what will actually cause them to remove the expired stock from the shelves and is this a potential source of abuse (using/selling expired stock)?

jniles:
Hopefully, their hierarchy.  People seem to treat out-of-stock much more strictly than stock expiry.  So, if we just make it painfully obvious that the expired medication doesn't count, they will hopefully be encouraged to get in-date medications to give to their patients.

lomamech:
I am thinking, how to resolve this issue, I have the following idea
- Have a usable quantity column, and that we can use this value as the quantity in stock
- The other idea is that the PV of stock loss, be automatically created by BHIMA and that we have a register similar to the movement of stock which will keep all of the automatic stock losses.

![image](https://user-images.githubusercontent.com/8035588/123394242-d2993a00-d596-11eb-90be-99d8f24f4f2c.png)

lomamech:
@jniles @mbayopanda  @jmcameron  How do you find these ideas

jniles:
I don't like the idea of having a automatic PVs generated by BHIMA.  There is a good chance that the date on the medicine is wrong, and the users can just correct it.

I'm okay with doing a new column with "usable quantity" in it.  I think that should be the first approach, and we can make it better with more PRs.

lomamech:
In this case I will start with the first proposal, but we will invite the stock manager to remove the stock by loss and to archive the documents

PR #5790, opened 2021-07-13
Add warning clock for expired lot

mbayopanda:
This PR adds a warning during stock exit for expired lots : 

image

Expired lots are still loaded for exit as lost because in others cases (patient, service, depot) expired lots cannot be submited

#5786


client/src/modules/stock/exit/exit.js
```diff
@@ -320,6 +320,10 @@ function StockExitController(
   function onLotSelect(row) {
     if (!row.lot || !row.lot.uuid) { return; }
 
+    if (row.lot.lifetime < 0) {
+      row.isExpired = true;
+    }
```
jmcameron: I was surprised this was necessary.  Doesn't the server DB function Stock.lots.read() do that?  I think there is an 'expired' flag that is set on the lots in computeLotIndicators() that might be what you want.

jmcameron (commented):
I think it would make more sense to filter out any expired lots in loadCurrentInventories() so they can never be selected.  So the time-running-out icon is nice but not necessary.

I think this is what we are headed toward in Issue https://github.com/IMA-WorldHealth/bhima/issues/5706.

If you feel there is a reason to have an expired lot accessible, lets talk further about it; I would like to understand why that might be needed.

mbayopanda:
> I think it would make more sense to filter out any expired lots in loadCurrentInventories() so they can never be selected. So the time-running-out icon is nice but not necessary.
> 
> I think this is what we are headed toward in Issue #5706.
> 
> If you feel there is a reason to have an expired lot accessible, lets talk further about it; I would like to understand why that might be needed.

In case we want to exit these expired lots with the option exit as loss

jmcameron:
> In case we want to exit these expired lots with the option exit as loss

Good one!  So I think you can filter out the expired lots for all types of exit except to Loss.  So I think you need a little more logic.

Then you need to allow the lot to be selected (no logic to prevent it from being select or need for the red time-out clock, right?

-Jonathan

client/src/modules/stock/exit/exit.js
```diff
@@ -215,8 +215,12 @@ function StockExitController(
     vm.movement.description = $translate.instant(mapExit[exitType.label].description);
     vm.stockForm.store.clear();
     vm.resetEntryExitTypes = false;
-
     vm.overconsumption = [];
+
+    // reload inventories only if exit type has been selected previously
```
jmcameron: This comment does not seem to match what you do in the "if (vm.depot)" block?

client/src/modules/stock/exit/exit.js
jmcameron: I'm still wondering if this is necessary.  Have you checked the value of the lot's '.expired' field.  When I look at line 770 of core.js, I see:

    lot.expired = !lot.exhausted && (lot.expiration_date < today);

Which seems very similar to what you need.

jmcameron (commented):
I made a couple of comments on the code.   Please review my comments and see if you want to update the code.

I tested it and it works perfectly.

client/src/modules/stock/exit/exit.js
mbayopanda: Old code... 👍🏽

client/src/modules/stock/exit/exit.js
mbayopanda: We already have information about expired lot with `Stock.lots.read(...)`, however the code above is to assign a flag on the current row of inventory to inform about the fact that the lot is already expired.

image

Since we are filtering now expired lots when we are not in loss context, we can remove this flag if we don't want this clock warning status

client/src/modules/stock/exit/exit.js
```diff
@@ -431,14 +444,19 @@ function StockExitController(
...
-      const uniformEntity = Stock.uniformSelectedEntity(entity);
-      vm.reference = uniformEntity.reference;
-      vm.displayName = uniformEntity.displayName;
-      vm.selectedEntityUuid = uniformEntity.uuid;
-      vm.requisition = (entity && entity.requisition) || {};
-      loadRequisitions(entity);
+    if (!entity) {
+      vm.reference = undefined;
+      vm.displayName = undefined;
+      vm.selectedEntityUuid = undefined;
```
jmcameron, line 446: Is it possible that you also need to reset vm.requistion here too?

client/src/modules/stock/exit/exit.js
jmcameron: I looked at the code again and would still like a response to my questions on this.   I see that this addition makes the status.tmp.html code a little cleaner -- but maybe not enough to be worth duplicating existing variables.

jmcameron (commented):
Further code comments.

I did not test it again.  I'm pretty confident it continues to work.

mbayopanda:
@jmcameron changes landed

jmcameron (commented):
The code changes look good.

I test this carefully and realized that there is still a problem.  For exits to Patients, Services, and Depots, it seems like it is possible to select medications that are out of stock.  I'm using an an out-of-date IMCK db.  For instance see this stock exit:
image

For reference, item 110016 is out of stock here:
image

My guess is that we have a check against 0 somewhere, rather than <= 0.    Note that I could not select 110010 which is also out of stock but does not have negative quantities.

Please see if you can fix it!

mbayopanda (commented):
> 110016

It seems like the inventory appear because its have some lots in stock out status and other still available 

image

In the lot select dropdown we can only select the lot which is available 

image

image

jniles (changes requested):
image

I just tried this using the test dataset and it looks like the green checkmarks no longer show up.    It also doesn't look like the inventory text is properly being displayed - can you fix that while you are at it?

mbayopanda:
@jniles changes landed

jniles (approved):
Confirmed working.  Looks good to me.

image

bors r+

jniles:
This was a fluke.

bors r+

PR #5792, opened 2021-07-14
four digit for unit cost in stock value report

### LINKED ISSUE #5788: Show more significant figures for stock values
Per email from a user:

> One of the problems is the exported report valeur du stock is that it rounds the price which when viewed gives the 
> value of zero instead of for example Promethazine 25mg tab 0.01 and slbutamol 4 mg ces 0.004.  
> 
> Can the report show the value of stock to four places?  This will help in confirming the total value and for making 
> further calculations on the value of stock.

mbayopanda:
This PR is related to #5788 

Close #5788

### DIFF
--- a/server/lib/template/helpers/finance.js
+++ b/server/lib/template/helpers/finance.js
@@ -36,17 +36,17 @@ function currency(value = 0, currencyId) {
  * @todo merge this function with the currency function, since actually
  * setting a position as third param in the currency function cause 0 decimal
  * when nothing is defined as position we are using this currencyWithPosition
- * @param {*} value 
- * @param {*} currencyId 
- * @param {*} precision 
- * @returns 
+ * @param {number} value the value to format
+ * @param {number} currencyId the id of the currency
+ * @param {number} digit the precision needed
+ * @returns {string}
  */
-function currencyWithPosition(value = 0, currencyId, precision) {
+function currencyWithPosition(value = 0, currencyId, digit) {
   const fmtNum = Number(currencyId);
   // if currencyId is not defined, defaults to USD.
   const fmt = fmtNum > 0 ? FORMATS[fmtNum] : USD_FMT;
   const format = JSON.parse(JSON.stringify(fmt));
-  format.precision = precision;
+  format.precision = digit;
   return new Handlebars.SafeString(accountingjs.formatMoney(value, format));
 }
 
--- a/server/lib/template/helpers/finance.js
+++ b/server/lib/template/helpers/finance.js
@@ -23,31 +23,24 @@ const EUR_FMT = {
 const FORMATS = [null, FC_FMT, USD_FMT, EUR_FMT]; // NB: Indeces must match currency IDs in database
 
 /** @todo use the currency filter fork written for the client to perform the same behaviour here */
-function currency(value = 0, currencyId) {
+function currency(value = 0, currencyId, digit) {
+  let output;
+  const DEFAULT_CURRENCY_PARAMETERS = 3;
   const fmtNum = Number(currencyId);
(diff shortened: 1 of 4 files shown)


server/lib/template/helpers/finance.js
```diff
@@ -30,6 +30,26 @@ function currency(value = 0, currencyId) {
...
+ */
+function currencyWithPosition(value = 0, currencyId, digit) {
+  const fmtNum = Number(currencyId);
+  // if currencyId is not defined, defaults to USD.
+  const fmt = fmtNum > 0 ? FORMATS[fmtNum] : USD_FMT;
+  const format = JSON.parse(JSON.stringify(fmt));
+  format.precision = digit;
+  return new Handlebars.SafeString(accountingjs.formatMoney(value, format));
+}
+
```
jmcameron, line 45: I like this approach.  However, I wonder if this functionality already exists in the code somewhere -- OR whether it should be moved to some utility class so it can be used in other areas of Bhima?

server/lib/template/helpers/finance.js
```diff
@@ -30,6 +30,26 @@ function currency(value = 0, currencyId) {
   return new Handlebars.SafeString(accountingjs.formatMoney(value, fmt));
 }
 
+/**
+ * @function currencyWithPosition
```
jmcameron: I wonder if a different name might be more descriptive:  currencyWithPrecision, currencyPrecision, ...?

jmcameron (commented):
The code looks good.  Although note my comments about the new utility function currencyWithPosition.

I tested this and it works.  But I think the stock value (the right-most column only needs 2 digits after the decimal.   Showing the unit price at is what really needs more digits (as you have done).

jmcameron:
BTW, I just took a second look at the Stock Value report I did (for IMCK) and noticed that the report includes some stock items with negative quantities.   Should we prevent them from occurring in the report or not compute subtotals for them.  See attached.

[Stock Value Report Test.pdf](https://github.com/IMA-WorldHealth/bhima/files/6816271/Stock.Value.Report.Test.pdf)

server/lib/template/helpers/finance.js
mbayopanda: The function is like currency, i was thinking to add another parameter in `currency` but it imply to add precision anywhere we have used currency (I didn't succeed to set the default precision when we have 3 parameters)

jmcameron (approved):
Looks good!

server/lib/template/helpers/finance.js
jniles: I think this is the best place for this functionality, as it will be exposed in all our handlebars reports as a helper function.  Considering that we use `Handlebars.SafeString()`, it is tied to the handlebars renderer.

server/lib/template/helpers/finance.js
```diff
@@ -23,11 +23,24 @@ const EUR_FMT = {
...
-function currency(value = 0, currencyId) {
+function currency(value = 0, currencyId, digit) {
+  let output;
+  const DEFAULT_CURRENCY_PARAMETERS = 3;
   const fmtNum = Number(currencyId);
   // if currencyId is not defined, defaults to USD.
   const fmt = fmtNum > 0 ? FORMATS[fmtNum] : USD_FMT;
-  return new Handlebars.SafeString(accountingjs.formatMoney(value, fmt));
+
+  if (arguments.length === DEFAULT_CURRENCY_PARAMETERS) {
```
jniles, line 33: Nice - this is helpful.  👍

jniles (approved):
Looks good to me.  I had hoped we could reuse the same function and I am glad we figured out a way!

bors r+

PR #5797, opened 2021-07-21
Improve the journal log with grid

### LINKED ISSUE #5796: Improvements to Journal Log
Here is the current Journal Log for HEV:

image

There are a couple of things that make this a bit hard to use and improvements that I would like to see:

1. Let's make this a ui-grid.  It would make it easier to sort, filter, save views, count rows, etc, for free
     1. We should add in inline filtering on all columns at the same time.
     2. We should add sorting by clicking on the column headers for the date and transaction ID fields.
     3. In the grid footer, let's put the number of visible rows so that we know if we should increase our limit or not.
2. We should add a link (`bh-receipt`) to the receipt on the transaction ID.
3. It might also be useful to add an `actions` link that allows you to see the document in the journal, view the patient's record concerned, view the invoice, etc, like we do for the Journal.  These would link back to the correct registries or bring up the rendered PDF of the document.

mbayopanda:
This PR is related to #5796 

image

Close #5796

### DIFF
--- a/client/src/modules/journal/journal_log.html
+++ b/client/src/modules/journal/journal_log.html
@@ -10,6 +10,7 @@
     <div class="toolbar">
       <fieldset>
         <bh-search-button on-click="JournalLogCtrl.openSearchModal()"></bh-search-button>
+        <bh-filter-toggle on-toggle="JournalLogCtrl.toggleFilter()"></bh-filter-toggle>
       </fieldset>
     </div>
   </div>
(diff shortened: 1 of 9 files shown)


jniles (commented):
All this looks good to me.

One request - could we add the `VO.TBA.123` identifier to the deleted records?  That will make it easier for users to see what was deleted.

mbayopanda:
We have now information about the deleted record : 

image

client/src/modules/journal/journal_log.html
```diff
@@ -27,41 +28,20 @@
...
-              </td>
-              <td>
-                <small translate>FORM.LABELS.BY</small>
-                <small>{{ row.display_name}}</small>
-              </td>
-            </tr>
-          </tbody>
-        </table>
-        
+        <div
```
jniles, line 31: 👍

jniles (approved):
Looks good to me.  This is very useful 👍 


bors r+

PR #5804, opened 2021-07-27
Fix currency xls export

### LINKED ISSUE #5798: Currency function crash after exporting document in Excel
This issue is about the fact that the currency function in handlebars crashes when we export a generated report in Excel format.

How to reproduce : 
1. Go to stock -> reports -> stock value report
2. Generate a report
3. Save the report as Excel
4. Download the Excel file

You will notice that the currency doesn't work 

image

mbayopanda:
This PR fixes the currency symbol deleted after the export to Xls.

The deletion of the currency symbol has been made intentionally to allow calculation in Excel, the issue was that the currency symbol was removed entirely in the entire BHIMA after export to Excel. With this PR we set back the default currency after the render of the Excel file.

Close #5798

### DIFF
--- a/server/lib/renderers/html.js
+++ b/server/lib/renderers/html.js
@@ -86,6 +86,13 @@ async function renderHTML(data, template, options = {}) {
   debug(`rendering HTML file`);
 
   const html = await hbs.render(template, data);
+
+  // set back the default currency after the rendering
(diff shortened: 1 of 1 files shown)


jniles (approved):
LGTM :+1:

bors r+

PR #5808, opened 2021-07-29
feature(Consider expired stock like stock out)

### LINKED ISSUE #5706: If a medicine has only expired stock, we should consider it as "stock out"

lomamech:
- Addition of the option in stock_setting to allow the consideration of expired stock as out of stock
- Addition of a new column, on the item grid in stock for the quantity of stock available and usable
- Recalculation of stock indicators taking into account only the quantity of stock available
- Update of the stock board table on the out of stock section

closes #5706


lomamech:
image

client/src/modules/stock/inventories/registry.js
```diff
@@ -47,7 +49,16 @@ function StockInventoriesController(
...
+    width            : '10%',
+    cellClass        : 'text-right',
+    type : 'number',
+  }, {
+    field            : 'usableAvailableStock',
+    displayName      : 'STOCK.USABLE_AVAILABLE_STOCK',
+    headerCellFilter : 'translate',
+    width            : '10%',
     cellClass        : 'text-right',
+    visible          : displayEnableExpiredStockOut,
```
jniles, line 61: This is cool 👍

client/src/i18n/en/stock.json
```diff
@@ -266,6 +268,8 @@
     "ENABLE_STRICT_DEPOT_PERMISSION_HELP_TEXT" : "Limit the consultation of stock registries to users according to their depots",
     "ENABLE_SUPPLIER_CREDIT_ON_STOCK_ENTRY" : "Enable automatic crediting of supplier on stock entry",
     "ENABLE_SUPPLIER_CREDIT_ON_STOCK_ENTRY_HELP_TEXT" : "Automatically adds the transactions to credit the supplier for the value of stock received in a depot when entering stock from a purchase order.  This is only triggered if the automatic stock accounting is set as well.",
+    "ENABLE_CONSIDERATION_EXPIRED_STOCK_OUT": "Enabled the consideration of expired stocks as in stock out",
```
jmcameron: I suggest using simpler English:
Consider expired stocks as stock out (unusable)

client/src/i18n/en/stock.json
```diff
@@ -243,6 +244,7 @@
     "SUCCESSFULLY_DELETED" : "Successfully deleted",
     "TO"              : "Destination",
     "UNIT_COST"       : "Unit cost",
+    "USABLE_AVAILABLE_STOCK": "Usable available stock",
```
jmcameron: For the English here, I suggest:
Quantity Usable

jmcameron (commented):
The code looks good.  I made a few suggestions for improving the English labels.

I tested this and it works okay.  I had to backdate the expiration date on a lot in order to test it.

Something to consider for the future (not this PR):  If the last expiration date for the lots for an inventory item is in the future but before the stock is exhausted (based on CMM), should we indicate the probable "Quantity Usable" based on CMM?

jmcameron:
@lomamech Please update the migration file to use the add_column_if_missing() functions, etc.  I've attached a modified version of the file that does this but keeps commented out old versions of the commands.  Thanks.
[migrate.sql.txt](https://github.com/IMA-WorldHealth/bhima/files/6916978/migrate.sql.txt)

client/src/modules/stock/inventories/registry.js
```diff
@@ -25,6 +25,8 @@ function StockInventoriesController(
   vm.openBarcodeScanner = openBarcodeScanner;
   vm.openStockSheetReport = openStockSheetReport;
 
+  const displayEnableExpiredStockOut = (Session.stock_settings.enable_expired_stock_out) || false;
```
mbayopanda, line 28: You can also write this code in a short way: 
```js
const displayEnableExpiredStockOut = !!(Session.stock_settings.enable_expired_stock_out);
```

server/controllers/stock/core.js
```diff
@@ -473,8 +497,10 @@ function computeInventoryIndicators(inventories) {
   for (let i = 0; i < inventories.length; i++) {
     const inventory = inventories[i];
 
-    // the quantity of stock available in the given depot
-    const Q = inventory.quantity; // the quantity
+    const inventoryExpiredLotsQuantity = inventory.expiredLotsQuantity || 0;
+
+    const Q = inventoryExpiredLotsQuantity
+      ? inventory.quantity - inventoryExpiredLotsQuantity : inventory.quantity; // the quantity
```
mbayopanda, line 503: The value of `Q` in some case must be conditioned by the stock setting `enableExpiredStockOut` also, I think

client/src/i18n/fr/stock.json
```diff
@@ -243,6 +244,7 @@
     "SUCCESSFULLY_DELETED" : "Supprimé avec succès",
     "TO"              : "Destination",
     "UNIT_COST"       : "Coût unitaire",
+    "USABLE_AVAILABLE_STOCK": "Stock disponible utilisable",
```
mbayopanda: This column is too long, could we just name it : `Quantité utilisable` ?

mbayopanda (approved):
Good job @lomamech 👍🏽 

I made just some minor comments but the PR is LGTM

client/src/i18n/en/stock.json
lomamech: I think this is a very nice analysis, a deposit can end up out of stock (Stock not usable in one day) or in risk of breaking

server/controllers/stock/core.js
lomamech: @mbayopanda  I used expiredLotsQuantity comes from this function, it's a bit confusing since I reused the same name for the property
````
  // verification of the activation of the option enable expired stock_out,
  // search for expired quantities in order to calculate the stock management
  // indicators with the usable quantities
  if (stockParams && stockParams.enable_expired_stock_out) {
    const lotsAvailable = await getLotsDepot(null, stockParams);

    lots.forEach(inventory => {
      inventory.expiredLotsQuantity = 0;

      lotsAvailable.forEach(lot => {
        const hasSameDepot = lot.depot_uuid === inventory.depot_uuid;
        const hasSameInventory = lot.inventory_uuid === inventory.inventory_uuid;

        if (hasSameDepot && hasSameInventory) {
          if (lot.expired) {
            inventory.expiredLotsQuantity += lot.quantity;
          }
        }
      });
    });
  }
````
in server/controllers/stock/core.js  line **359**

lomamech:
@jmcameron @mbayopanda  I'm done with the first part

jniles:
@lomamech could you do a rebase here?  It looks like there are conflicts to resolve.

jmcameron:
> @lomamech could you do a rebase here? It looks like there are conflicts to resolve.

@lomamech Chris, I'm releasing the "Fix migration schemas" PR https://github.com/IMA-WorldHealth/bhima/pull/5818 now.  So I suggest you wait until after it lands to rebase vs master.  Thanks.

jniles:
bors try

lomamech:
@jniles  I think I have completed all of the suggestions

jmcameron:
> @lomamech Please update the migration file to use the add_column_if_missing() functions, etc. I've attached a modified version of the file that does this but keeps commented out old versions of the commands. Thanks.

NOTE: These changes have been done in master.  So all you need to do is to rebase.

lomamech:
@jniles it's done

jniles (approved):
Looks good to me.

bors r+

PR #5810, opened 2021-07-30
Fix missing rates

### LINKED ISSUE #5581: Better handling of new currencies without exchange rates
We need more graceful handling of new currencies without defined exchange rates. 

This will require updating some of the exchange rate queries to handle currencies with missing exchange rates.  Currently they throw errors if ANY currency does not have an exchange rate.  This means even setting the exchange rate for a new currency via the Admin > Exchange rate fails.

Until this is done, developers/installers will have to add at least one exchange rate for any new currency as part of the installation / update of Bhima software that includes new currencies.

Once this is complete, we should update Admin > Exchange rate setting interface to identify exactly which currency needs an initial exchange rate and allow setting through the normal UI.   We should also update the purchase order form similarly so that other currencies (not missing exchange rates) can still be used.

jmcameron:
If a new currency is added, urges the user to set it's exchange rate before doing anything.  It does this by popping up a modal urging them to add an exchange rate for the new currency/currencies.  It provides a like to the exchange page and adds a notice to the exchange setting modal about which currency needs to be added.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5581

*TESTING*
- Use bhima_test
- Edit test/data.sql and comment out/remove the line in "INSERT INTO `exchange_rate` VALUES" for Euros.  In other words, delete this line:   (3, 1, @EUR, 0.8400, NOW())
- Rebuild the database:   yarn build:db
- Run Bhima
- Go to the top/main page and see the new modal
- Follow the link to the Exchange page and then try setting exchange rates
- Note the hint about which one needs setting
- Set the rate
- Verify things work normally now.


client/src/modules/exchange/exchange.modal.html
```diff
@@ -4,6 +4,7 @@
       <li class="static" translate>EXCHANGE.TITLE</li>
       <li class="title" translate>EXCHANGE.ADDING_RATE</li>
     </ol>
+    <p ng-if="ModalCtrl.missingRates" class="missing-exchange-rate">{{ModalCtrl.missingRatesWarning}}</p>
```
jniles, line 7: There appears to be a logic bug here where `getMissingRates()` doesn't return the correct value after setting the exchange rate.  To reproduce, follow the steps in the test case, submit a new rate, and then try to set another rate and this message will persist.

image

client/src/js/services/ExchangeRateService.js
```diff
@@ -196,19 +213,40 @@ function ExchangeRateService($http, util, Currencies, Session) {
...
+    return null;
   }
 
   function del(id) {
     return $http.delete(`/exchange/${id}`)
       .then(util.unwrapHttpResponse);
   }
 
+  function getMissingExchangeRates() {
+    return service.missingRates;
```
jniles, line 254: This list needs to be refreshed on create/update/delete.

Let's move the logic for creating the `missingRates` into a separate helper function (`computeMissingRates()` or something) that can get called when the exchange rate is updated, deleted, or created.

jniles (commented):
Thanks for including how to test this!  I walked through it and it worked pretty closely to as described.  I discovered one bug when setting a second rate though.  I've made those comments inline.

client/src/js/services/ExchangeRateService.js
jmcameron: Did not think a separate function was really necessary since all the necessary logic is in read() and writing a stand-alone computeMissingRates() would have duplicated a lot of read().   So I just updated getMissingExchangeRates() to call read and fixed an oversight in read().

lomamech (commented):
@jmcameron  My first concern, did we have an interface in Bhima to add a new currency? Or will I have to add directly from the database

jmcameron:
> @jmcameron My first concern, did we have an interface in Bhima to add a new currency? Or will I have to add directly from the database

No adding a new currency requires adding it to the DB via SQL.

lomamech:
> > @jmcameron My first concern, did we have an interface in Bhima to add a new currency? Or will I have to add directly from the database
> 
> No adding a new currency requires adding it to the DB via SQL.

I found how to add it to test the PR

jmcameron:
@jniles, please check my mods to the create/update/del functions in ExchangeRateService.js to see if they are okay.  Thx.

lomamech:
image
I now understand the impact of PR I approve

lomamech:
image

I noticed that there is a message that is not very well formatted, **CURRENCY_NOT_SUPPORTED(null)**

jmcameron:
> I noticed that there is a message that is not very well formatted, **CURRENCY_NOT_SUPPORTED(null)**

I've added these missing message definitions.  However, I think they will probably only be seen by developers so that is why (I think) they were not defined before.

jmcameron:
> 
> 
> > I noticed that there is a message that is not very well formatted, **CURRENCY_NOT_SUPPORTED(null)**
> 
> I've added these missing message definitions. However, I think they will probably only be seen by developers so that is why (I think) they were not defined before.

@lomamech Please try to bring up this error message again and see if the new definitions work.  I was not sure how you got them to appear so I was not able to test them.

lomamech:
@jmcameron I had tried to add the Rouble (Russian), 
````
INSERT INTO `currency` (`id`, `name`, `format_key`, `symbol`, `note`, `min_monentary_unit`) VALUES (4, 'Rouble', 'RBL', '₽', NULL, 0.01);
````
in finance.js
const ROUBLE_FMT = {
  symbol : '₽',
  precision : 2,
  thousand : '&nbsp;', // thin non-breaking space
  decimal : ',',
  format : '%v&nbsp;%s', // value before symbol
};

const FORMATS = [null, FC_FMT, USD_FMT, EUR_FMT, ROUBLE_FMT]; // NB: Indeces must match currency IDs in database
````

lomamech:
And then, I tried to define the exchange rate for the Ruble,

image

jniles (approved):
Tested and working.  Thanks @jmcameron.

bors r+

lomamech:
> Tested and working. Thanks @jmcameron.
> 
> bors r+

There is a text that is not yet translated

lomamech:
@jniles  
image

jmcameron:
@jniles @lomamech I'll create a new PR to fix this.

lomamech:
> @jniles @lomamech I'll create a new PR to fix this.

Good

jniles:
Oh dear, I only tested in French.  Whoops!  Good catch @lomamech.

In the future, you can cancel bors by typing "bors r-" if I miss something.  Please do it!

PR #5815, opened 2021-08-02
Fixed bugs in merge lots interface

### LINKED ISSUE #5802: Bugs in Merge Duplicate Lots
There are a few bugs in the merge duplicate lots modal.

1. Sometimes, the "number of duplicates" is off.  For example, see below:
DuurLsfTWA

The correct value appears to be 3 lots, yet the grid reports 2.  The grid should report 3.

2. The modal sometimes gets confused about what to merge and fails silently.  This often happens when you move away from the default selection, then move back.  Here is an example:

B7ATU2VNgE

This is easily fixed by just reloading the modal, but is still pretty annoying.  My assumption is that the data model being used to store the selection can get in a undefined state and cause these issues.

jmcameron:
Fixed the bugs in the lots merge interface.
The first problem is because the display of duplicate lots is based on label equality, but the lots shown when the user wants to merge specific lots used SQL "LIKE".
To fix the second issue (hopefully) I whenever the selected lot is changed, I clear the 'merge' attribute for all lots.  That way the flags for 'select' and 'merge' are always compatible.  I was not able to duplicate the problem after this fix.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5802

### DIFF
--- a/client/src/modules/stock/lots/modals/duplicates.modal.js
+++ b/client/src/modules/stock/lots/modals/duplicates.modal.js
@@ -38,9 +38,7 @@ function DuplicateLotsModalController(data, Instance, Lots, Notify, $translate) 
     vm.selectedLot = lotSelected;
     vm.lots.forEach(lot => {
       lot.selected = lot.uuid === lotSelected.uuid;
-      if (lot.uuid === lotSelected.uuid) {
-        lot.merge = false;
-      }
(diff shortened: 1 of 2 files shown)


jniles (approved):
Tested and cannot reproduce the issues I reported on this branch.

bors r+

PR #5818, opened 2021-08-03
fix: update migration schemas

jniles:
Updates the migration schemas so that they don't crash on production databases.

To test this, pull any production database (version 1.20.0 or higher) and run `yarn migrate`.  Execute the migration script in that database and there should be no warning in the console.


jmcameron (approved):
These changes look great.  Thanks!

Issue #5807, opened 2021-07-28
Sorting by expiration date in Duplicate Lots Registry doesn't work


jniles:
The date sort in the duplicate lots registry ui-grid doesn't work.  As an example:

![image](https://user-images.githubusercontent.com/896472/127358073-a56e1567-83d2-437c-b74a-d32876fd50c1.png)

jniles:
@lomamech can you work on this?

lomamech:
@jniles I cannot reproduce the error
![image](https://user-images.githubusercontent.com/8035588/128163738-24a05ee4-72e1-460f-a6c4-f6c67ca67d9a.png)

![image](https://user-images.githubusercontent.com/8035588/128163876-5d153393-ead4-4de2-bb82-1b8b24724b90.png)

jniles:
Duplicate lots registry, not the lots in stock registry.  :+1:

PR #5819, opened 2021-08-04
Force translation of error message for missing exchange rate

jmcameron:
In PR https://github.com/IMA-WorldHealth/bhima/pull/5810, some error messages were not translated.  This PR fixes this issue.  

To test it, see https://github.com/IMA-WorldHealth/bhima/pull/5810#issuecomment-892494814
and try setting the exchange rate for Rubles.   The warning message is now translated but is still a bit strange.  I don't think the format is that important, because only developers should see this error in case.
#5807


lomamech:
image
Cool

lomamech (approved):

PR #5823, opened 2021-08-05
feat(journal): auto-resize journal log

### LINKED ISSUE #5821: Journal Log needs to auto-resize
The Journal Log should have the ability to autoresize - see below:

QHZvK1FLSe

This is how the rest of our modules work in BHIMA, so it is unexpected that this doesn't work the same way.

jniles:
Adds resizing to the journal log for a better UX.

Closes #5821.


jniles:
@mbayopanda can you give me a review ?

mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #5828, opened 2021-08-06
Adding search by action and exports

### LINKED ISSUE #5827: We need a way to export data from the Journal Log
In light of the recent fraud reports, we need a way export transactions by user from the journal history module.  This would help in creating fraud reports in the future.  Specifically, it would be nice to be able to download the data in CSV or Excel format.

mbayopanda:
This PR adds:
- Export option for CSV and Excel
- Search by type of action (delete or edit)

Close #5827

### DIFF
--- a/server/config/routes.js
+++ b/server/config/routes.js
@@ -433,6 +433,7 @@ exports.configure = function configure(app) {
   app.get('/reports/finance/account_report', financeReports.reportAccounts.document);
   app.get('/reports/finance/account_report_multiple', financeReports.reportAccountsMultiple.document);
   app.get('/reports/finance/journal', financeReports.journal.postingReport);
+  app.get('/reports/finance/journal/log', financeReports.journal.journalLogReport);
   app.get('/reports/finance/account_statement', financeReports.accountStatement.report);
   app.get('/reports/finance/general_ledger/', financeReports.generalLedger.report);
   app.get('/reports/finance/creditors/aged', financeReports.creditors.aged);
--- a/client/src/modules/journal/journal_log.html
+++ b/client/src/modules/journal/journal_log.html
@@ -21,7 +21,7 @@
                 </a>
               </li>
               <li role="menuitem">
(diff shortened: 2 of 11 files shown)


server/config/routes.js
```diff
@@ -433,6 +433,7 @@ exports.configure = function configure(app) {
   app.get('/reports/finance/account_report', financeReports.reportAccounts.document);
   app.get('/reports/finance/account_report_multiple', financeReports.reportAccountsMultiple.document);
   app.get('/reports/finance/journal', financeReports.journal.postingReport);
+  app.get('/reports/finance/journal/log', financeReports.journal.journalLogReport);
```
jniles, line 436: :+1:

jniles (commented):
The CSV export works well 👍 

Could you do some updates to the XLSX export to make it more useful?  Users will want access to the information in the `value` section, but it currently isn't usable to them.

jniles (changes requested):
Have you tested this with the vanga database?  It crashes as is, but when I fixed the crash, it doesn't properly display all the information:

image

Since they are the ones who will use this information, let's try to fix this before merge.

mbayopanda:
> 
> 
> Have you tested this with the vanga database? It crashes as is, but when I fixed the crash, it doesn't properly display all the information:
> 
> image
> 
> Since they are the ones who will use this information, let's try to fix this before merge.

Thank you @jniles for the review and the catch

mbayopanda:
@jniles changes landed

jniles:
Hmm... still not getting any descriptions from the Vanga database in the Excel export.  Let me see if I can work on it.

jniles:
Okay, got it working:

image

image


@mbayopanda take a look at my changes.  If they are good with you, we can merge this.

mbayopanda:
It's LGTM 👍🏽

jniles (approved):
LGTM.

Bors r+

PR #5830, opened 2021-08-06
Fix issues with the stock lots usage schedule

### LINKED ISSUE #5825: Stock Consumption Schedule seems to be buggy still
It seems like the stock consumption schedule is still buggy.  Here is a screenshot with Vanga's latest data:

image

There is a random amount of "wasted stock" displayed here, even though that stock is still usable.  It is probably a rounding error somewhere.

jmcameron:
This PR fixes/improves these issues:
- Prevents rounding errors from indicating articles being wasted when they are actually all used
- Start the lot usage schedule at 00:00AM of "today"
- Improve the message about the suggested date to order if the indicated order date is before today.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5825

Tested on Vanga and IMCK databases.

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -338,6 +338,7 @@
     "NEVER_EXHAUSTED": "Never",
     "NO_LOTS_FOUND": "No lots found that were not expired or exhausted!",
     "PROJECTION_BASIS": "This projection is based on current average consumption of {{avg_consumption}} {{unit_type}} per month.",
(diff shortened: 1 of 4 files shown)


jniles (approved):
Yay!  This looks right by my eye.  Proof that it works with the Vanga dataset:

image
_A stock that will run out in years from now_

The Batrim case is also resolved, by my internet isn't powerful enough to upload the screenshot at the moment.


bors r+

PR #5836, opened 2021-08-09
fix: remove exit to purchase from database

### LINKED ISSUE #5552: Test data has a stock exit from a purchase order
If you run our integration tests, then inspect the stock movement registry, you find this gem:

image

This may just be a bug in our test data.  But it would be useful to track it down and figure out where it is coming from.

jniles:
Closes #5552.

image


jniles:
@mbayopanda can I get a review of this PR?

mbayopanda (approved):
Good catch 👍🏽 

bors r+

PR #5839, opened 2021-08-09
feat(reports): purchase price report

### LINKED ISSUE #5519: Report on inventory (purchase) prices
Inspired by https://github.com/IMA-WorldHealth/bhima/issues/5401#issuecomment-783668386, it would be nice to have a report that shows the historical purchase price for medication.  For example, for Quinine 500mg, it would be nice to show each instance in which is was "purchased" or entered into the inventory from either a donation, a purchase, or an integration.

The following information would be interesting to know:
  1. What date was it entered?
  2. Who entered it?
  2. What was the reason? (purchase, integration, donation)
  3. How many items?
  4. What was the unit cost?
  4. How did this affect the average cost? (https://github.com/IMA-WorldHealth/bhima/issues/5401)
  5. What was the total value of the movement?

Finally, in summary: what was the min price and what was the max price?

This would give a lot more transparency to the purchasing process and the value of stock.

jniles:
This report shows the historical purchase price of an item since there is no easy way to get this information in BHIMA currently.

Close #5519.

image
_Current Presentation_


client/src/i18n/en/report.json
```diff
@@ -157,6 +157,10 @@
       "DESCRIPTION": "This report shows the difference between what we purchase with what we have entered in the stock.",
       "TITLE": "Purchase Order Analysis"
     },
+    "PURCHASE_PRICES": {
+      "DESCRIPTION": "This report shows the purchase value of a medicine over time, eachtime it was entered into stock.  It also shows the margin over the current inventory price.",
```
jmcameron: Typo:   eachtime -> each time

client/src/modules/reports/reports.routes.js
```diff
@@ -8,52 +8,53 @@ angular.module('bhima.routes')
       'account_report_multiple',
```
jmcameron, line 8: Thank you for sorting these!

server/models/bhima.sql
```diff
@@ -434,7 +435,7 @@ INSERT INTO `indicator_status`(`id`, `text`,`translate_key`)VALUES
 INSERT INTO `indicator_type`(`id`, `text`,`translate_key`)VALUES
   (1, 'hospitalization', 'DASHBOARD.HOSPITALIZATION'),
   (2, 'staff', 'DASHBOARD.STAFF'),
-  (3, 'fianance', 'DASHBOARD.FINANCE');
+  (3, 'finance', 'DASHBOARD.FINANCE');
```
jmcameron, line 444: Good catch!

jmcameron, line 444: However, I think you need to fix the same item in the migrate.sql file.

jmcameron (commented):
I took a casual look at the code and it seemed fine.
I ran test with the IMCK db.   Here are few comments about the report output:

- It is not clear what the $0.16 is.  Either label it or delete it.
- What does the pink line mean (are we getting paid to take drugs)?
- Rather than having the Margin and percent margin in one cell, I suggest breaking them into two columns.  I think it will make it cleaner and also make the Excel version more useful.
- Since the report is only for one inventory item, I suggest putting the inventory code/text as a subtitle (under "PURCHASE PRICE ANALYSIS" in a larger font and not have it as part of the table.
- I don't see min/max prices (as described in the PR)
- We recently increased the number of digits after the decimal for unit prices (due to a Vanga request, as I recall).   I suggest changing it from 2 to 4 (or making it an option).
- The report currently covers all time.   Does it make sense to allow selection of a range of dates?
- Should this be specific to a Depot?

image

jniles:
@jmcameron thanks for the review!  I'll respond inline for ease:

> It is not clear what the $0.16 is. Either label it or delete it.

The price at the top is the current sale price.  I agree that it wasn't well labeled and will label it better.

> What does the pink line mean (are we getting paid to take drugs)?

The margin is the difference between the value of the stock (as defined by the purchase price) and the current sale price.  In the long run, this doesn't make a ton of sense (we should weave in the price changes over time), but it is a good first blush since we might not have all the information available to weave in the price changes...

> Rather than having the Margin and percent margin in one cell, I suggest breaking them into two columns. I think it will make it cleaner and also make the Excel version more useful.

Good point.

> Since the report is only for one inventory item, I suggest putting the inventory code/text as a subtitle (under "PURCHASE PRICE ANALYSIS" in a larger font and not have it as part of the table.

Good point.

> I don't see min/max prices (as described in the PR).

Will add it to a summary table.

>  We recently increased the number of digits after the decimal for unit prices (due to a Vanga request, as I recall). I suggest changing it from 2 to 4 (or making it an option).

Good idea.  I forgot about that.  Will add it.

>  The report currently covers all time. Does it make sense to allow selection of a range of dates?

Hmm.....  Hmm....  I'm not sure.  Let me ask our users.

> Should this be specific to a Depot?

No, since this concerns purchasing and selling, neither of which are tied to a particular depot.

jniles:
@jmcameron changes landed.

jmcameron (approved):
Looks great.
One MINOR cosmetic thing (ignore if you wish):
image

I don't think the "Sale Price" needs to be underlined.   However, adding a little CSS margin above it would make it a little easier to read.

jmcameron:
I did a little more testing with the IMCK db and ran into an oddity, which may not be related to this PR.  I tried a report for a stock that appeared in the "Articles in Stock" and got an empty report, with a Sales price (I presume that is from the Inventory item).  When I tried finding that stock article in "Articles in Stock", I see the system thinks it is in stock, but there are the "Articles in Stock" registry shows nothing for that item when I narrow the search for that inventory item and include exhausted stock.  Nothing appears when I switch to "Lots in Stock" for this inventory item.

image

client/src/i18n/fr/report.json
```diff
@@ -162,6 +162,10 @@
       "DESCRIPTION": "Ce rapport permet de connaitre la différence entre ce la quantité commandée et la quantité entrée en stock",
       "TITLE": "Analyse de commande d'achat"
     },
+    "PURCHASE_PRICES": {
+      "DESCRIPTION": "Ce rapport indique la valeur d'achat d'un médicament au fil du temps, à chaque entrée en stock. Il montre également la marge par rapport au prix d'inventaire actuel.",
+      "TITLE": "Analyse du prix d'achat"
+    },
```
mbayopanda, line 182: 👍🏽

server/controllers/stock/reports/purchase_prices/index.js
```diff
@@ -0,0 +1,89 @@
...
+  const entriesSQL = `
+    SELECT
+      sm.date, dm.text AS hrReference,
+      sm.unit_cost, sm.quantity,
+      l.label AS lot_label,
+      flux.label AS reason,
+      (inv.price - sm.unit_cost) AS margin,
+      ((inv.price - sm.unit_cost) / sm.unit_cost) AS percent_margin,
+      ((inv.price - sm.unit_cost) < 0) AS negative,
+      SUM(sm.unit_cost * sm.quantity) OVER (ORDER BY sm.date) AS running_total,
```
mbayopanda: :-) window function

mbayopanda (commented):
The code is looking good to me, but I cannot understand where this value is coming from 

image

jniles:
@mbayopanda great catch!  It was summing by day, rather than by row.  I've fixed it now.

jmcameron (commented):
I tested this and the UI and result table seems to work okay, 

But I have a couple of questions to improve my understanding of what you are trying to display in this table.

- "Sales price" == inventory price, right?
- I think the margin should be per unit (not based on the total sale)
- I suggest moving the Total to the last column and and putting the margin/percent margin just after the unit price.  Also relabel "Margin" to "Unit Margin".
- What is **margin**?  When I hear the word margin, I think of what is received that is MORE than what it costs us.   So that would be:  Margin = Inventory Price - unit_price, right?  Doesn't this assume all stock sales will receive full inventory price and ignore discounts, patient group subsidies, etc, etc?
- I also wondered if the percentage margin be based on the inventory price.   Currently we are doing:  

      Percent Margin = 100% * (inventory_price - unit_price)/unit_price

   It seems to me it should be:

      Percent Margin = 100% * (inventory_price - unit_price) / inventory_price

   Since the unit price varies significantly, it distorts the margins/percent margins when it is small.   Using the inventory price seems more stable.

- What value is there in showing the running value and sum at the bottom in the "Total" column?  
- Shouldn't the value of the stock be based on WAC?

jniles:
Hey @jmcameron, thanks for your feedback!  I'll respond inline:

> But I have a couple of questions to improve my understanding of what you are trying to display in this table.
> 
>     * "Sales price" == inventory price, right?

Yes.  Sales price is the default unit price filled in for a patient when creating an invoice (assuming no special price list is applied).  It's set in the inventory.

>     * I think the margin should be per unit (not based on the total sale)

The `inventory.price` is the unit price the institution uses to sell to patients.  So, they both should be in the same units - if the inventory item is a box, the institution will buy boxes at a `unit_cost` and sell them at the `inventory.price`.

>     * I suggest moving the Total to the last column and and putting the margin/percent margin just after the unit price.  Also relabel "Margin" to "Unit Margin".

I can do that.

>     * What is **margin**?  When I hear the word margin, I think of what is received that is MORE than what it costs us.   So that would be:  Margin = Inventory Price - unit_price, right?  Doesn't this assume all stock sales will receive full inventory price and ignore discounts, patient group subsidies, etc, etc?

Yes, the `margin = inventory price - unit_price`.  The purpose of margin here is to let us know if we are selling our stock for more than we are purchasing it and by what percentage.

I think I know where you are coming from, and the MD of Vanga has a similar train of thought.  Here is the feedback from the medical director of Vanga:

> One question is how would this report look if the sales price changed?  For cipro this report looks back a long way over multiple purchases which gives a trend.   I'm thinking in the base settings we should be able to limit the length to the following possible options: Last purchase, All time, 1 year or a menu of months (9, 6 or 3 months).  It would be good to just give a snap shot summary of all the produces on one report.  This becomes a one page actionable report that says we are selling Talk at a loss and should we not increase the price and Cipro if at 100% and does the management team think this is enough to help support the institution or do we need to try more for 150%.  

Basically, this report is keeping one value constant (the current sales price), but letting the purchase price vary.  That doesn't make sense.  We should either have the historical purchase price and historical sale price, or just do the snapshot approach the MD is recommending above.  I think that is where you are going as well, correct?



> 
>     * I also wondered if the percentage margin be based on the inventory price.   Currently we are doing:
>       ```
>       Percent Margin = 100% * (inventory_price - unit_price)/unit_price
>       ```
>       It seems to me it should be:
>       ```
>       Percent Margin = 100% * (inventory_price - unit_price) / inventory_price
>       ```

I'm not sure this is true.  We want to know our profit margin.  It should be double or triple what we are purchasing it for.  By using the `inventory.price` it may indeed be more stable, but we would be showing the percentage difference between the two, rather the profit margin, no?  Maybe I'm just confused.

>         
>       Since the unit price varies significantly, it distorts the margins/percent margins when it is small.   Using the inventory price seems more stable.
> 
>     * What value is there in showing the running value and sum at the bottom in the "Total" column?

It's just the last value.  So the total value of the orders to date (how much you've spent on Quinine, for example).

>     * Shouldn't the value of the stock be based on WAC?

Not for this report - we are looking at purchase price and not on stock value.  The goal is to see if there were errors made during purchasing that we can try and catch.

mbayopanda (approved):
Calculation in the report are correct 👍🏽 

LGTM

jniles:
image

@jmcameron I've removed all reference to a margin.  This report is needed because there isn't a good way to find the historical purchase price of a good.  Instead of doing two things at once, this report now does a single thing.  For looking at margins, we'll make another report that is just a snapshot as recommended by Dr. Rice.

jmcameron:
> @jmcameron I've removed all reference to a margin. This report is needed because there isn't a good way to find the historical purchase price of a good. Instead of doing two things at once, this report now does a single thing. For looking at margins, we'll make another report that is just a snapshot as recommended by Dr. Rice.

Ok.  This is on my list to review next.

jmcameron (commented):
Looks good.    
A few of comments:
- Totals (sums) do not need 4 decimal places; 2 is plenty.   We do need more for the unit price as you have done.
- It would be helpful if the drop-down menu to select an inventory item would indicate items which are not in stock and have no purchase history.  Not sure if the list should be pruned (easy) or non-existing items greyed out (not so easy).
- It would be useful of the price report could be accessed through the action menus on the inventory registry and articles in stock registry.

jniles:
@jmcameron, I didn't do the filtering because I think there isn't an elegant solution to it.  But the rest is ready to go.

server/controllers/stock/reports/purchase_prices/report.handlebars
```diff
@@ -0,0 +1,92 @@
...
+    </div>
+
+    <br/>
+
+    <div class="row">
+      <div class="col-xs-4">
+        <table class="table table-condensed table-report table-bordered">
+          <tbody>
+            <tr>
+              <th>{{translate 'FORM.LABELS.MAX_PRICE'}}</th>
```
jmcameron: I suggest changing max/min price to:  Max unit price / Min unit price (to avoid confusion)

jmcameron (approved):
It looks great.  Very useful report.
One minor suggestion: at the end of the form, to avoid confusion, change
- Max price => Max unit price
- Min price => Min unit price

This PR is ready to be released.  If you want to make the change above, fine.  Otherwise I can fold it into the next PR I work on.

server/controllers/stock/reports/purchase_prices/report.handlebars
jniles: Good idea.  I'll do that and release.

PR #5845, opened 2021-08-11
Update stock value report

### LINKED ISSUE #5408: Add "per-item purchase price" and "per-item invoice price" to Articles in Stock Registry
We've received the following request from our partners in the field:

>  I would like an easy way to assess the value of items in stock and the cost of a requisition.  Is it possible to add two columns to this report in BHIMA?  Stock---Items in Stock.  One column would be per-item purchase price and the second column would be per item sales price.  

By only adding these two columns to this report we could reach the following three benefits.:
   1. We could then export this report and in Excel calculate the total price based on the number of items I want to order for the items in stock out.
   2. We could export this report in Excel and calculate the value of stock on hand by simply multiplying the quantity on hand by the purchase price and then be able to easily report the value of stock on hand.  
   3. We could export this report and calculate the difference between the purchase price and sales price to evaluate if items are being sold under purchase price and also calculate the profit margin.

jmcameron:
Add the inventory price to the Stock Value Report.  Renames some internal variables to improve readability.

This also fixes a bug in the client-side form to create the Stock Value Report that was preventing it from remembering the "Exclude inventories with zero stock value" checkbox flag if the preview was cancelled and then a new preview was requested.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5408

**TESTING**
- Use production database
- Run the Stock Value Report
- Notice the new column for Inventory Price for inventory articles.
- Try exporting to PDF
- NOTE:  When I try to export to Excel the produced file is not in Excel format.   This needs to be fixed (probably in another PR).
- Test the bug fix
   - Go back to Stock Value Report form
   - Reload the browser to start fresh
   - Select the "Exclude inventories with zero stock value" checkbox
   - Preview the report
   - Note that there are no items with zero quantity
   - Cancel the preview (use the button at the top left to do that)
   - Do NOT click on the "Exclude inventories with zero stock value" checkbox (it should still be checked)
   - Preview the report
   - Note that there are still no items with zero quantity


mbayopanda (commented):
This feature will be useful since there was many request for that, good job @jmcameron 

For the first review, I have noticed that we can fix colspan and adds more information if it is possible : 
- Add a column for value based on inventory selling price (price * quantity)
- Add sum in the footer of this new column

image

Columns can be : 
- Code,
- Inventory Text,
- Quantity
- Inventory Price,
- Value for the price
- Stock Unit Cost
- Value for the stock unit cost

Names of these columns that I mentioned above don't matter

jmcameron:
>     * Add a column for value based on inventory selling price (price * quantity)
> 
>     * Add sum in the footer of this new column

@jniles and I discussed this yesterday and we agree with you but decided that it would be better to add a new issue for that (which I have not done yet).  @mbayopanda I encourage you to take your notes above and create a new issue.

mbayopanda (approved):
The issue #5847 has been created, and this PR is LGTM 👍🏽 

bors r+

Issue #5803, opened 2021-07-27
Translate boolean values in filter bar


jniles:
We've gotten away with just using numbers for boolean values in the `bh-filter-bar` that goes above our registries so far.  Basically, instead of "yes" or "no", we have a `1` or `0`.  To programmers, this makes sense, but it doesn't to users.  Here is an example:

![image](https://user-images.githubusercontent.com/896472/127131126-f4ed01a6-21e5-49d5-8af4-fbba37b5a6f6.png)

We can improve this by explicitly telling BHIMA that a filter is a `boolean` value and then displaying a translated yes/no based on the truthyness or falsiness of the value.  We will probably want to put this behavior in the [FilterService](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/services/FilterService.js) or configure it from a service using the `FitlerService` similar to [our `ngFitler` parameter to determine how to display the value](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/modules/purchases/purchases.service.js#L50).

PR #5851, opened 2021-08-13
5553: sort stock movements in decreasing order

### LINKED ISSUE #5553: Sort stock movements in decreasing order in stock movement registry
In our cash payments, invoices and voucher registries, we sort the items in decreasing order.   That way, the most recent value is the first value in the list.  However, for stock movements, we don't do this.  Here is the default view:

image

We should sort these in decreasing order for the stock movements registry.  It would be a better user experience since they will already be used to seeing the most recent values first.

jniles:
This PR does two things:

 1. Sorts the stock movements in decreasing order
 2. Adds in the reason for the stock movement to the registry.  This has a performance penalty, but in my tests it has been negligible.  There are ways to speed it up, but it will make the code more difficult to parse and think about.  If this becomes an issue, I can propose it.
 
 Some screenshots:
 
image
_Finding a particular patient by their ID_

image
_All interactions (entries + exits) to pharmacies with PAX in the name_

image
_Defaults to most recent to most ancient_

Closes #5553.
Partially addresses #5424.

### DIFF
--- a/client/src/modules/stock/movements/registry.js
+++ b/client/src/modules/stock/movements/registry.js
@@ -93,7 +93,6 @@ function StockMovementsController(
       field : 'fluxName',
       displayName : 'STOCK.FLUX',
       headerCellFilter : 'translate',
-      // cellTemplate : 'modules/stock/movements/templates/flux.cell.html',
     }, {
       field : 'cost',
       type : 'number',
--- a/client/src/modules/stock/movements/registry.js
+++ b/client/src/modules/stock/movements/registry.js
@@ -90,10 +90,10 @@ function StockMovementsController(
       headerCellFilter : 'translate',
       cellTemplate : 'modules/stock/movements/templates/io.cell.html',
     }, {
-      field : 'flux_id',
+      field : 'fluxName',
       displayName : 'STOCK.FLUX',
       headerCellFilter : 'translate',
-      cellTemplate : 'modules/stock/movements/templates/flux.cell.html',
+      // cellTemplate : 'modules/stock/movements/templates/flux.cell.html',
     }, {
       field : 'cost',
       type : 'number',
@@ -228,8 +228,10 @@ function StockMovementsController(
     // the column must be filtered on the translated text
     row.io = $translate.instant(row.is_exit === 0 ? 'STOCK.INPUT' : 'STOCK.OUTPUT');
 
+    const fluxName = $translate.instant(getFluxName(row.flux_id));
     // compute the fluxName from its ID
-    row.fluxName = getFluxName(row.flux_id);
+    row.fluxName = fluxName.concat(row.target ? ` - ${row.target}` : '').trim();
+
   }
 
   function toggleLoading() {
--- a/client/src/modules/stock/movements/templates/flux.cell.html
+++ b/client/src/modules/stock/movements/templates/flux.cell.html
@@ -1,3 +1,0 @@
-<div class="ui-grid-cell-contents" ng-if="!row.groupHeader">
-    {{ row.entity.fluxName }}
-</div>
--- a/client/src/modules/stock/movements/templates/flux.cell.html
+++ b/client/src/modules/stock/movements/templates/flux.cell.html
@@ -1,3 +1,3 @@
-<div class="ui-grid-cell-contents" ng-if="!row.groupHeader" translate>
+<div class="ui-grid-cell-contents" ng-if="!row.groupHeader">
     {{ row.entity.fluxName }}
 </div>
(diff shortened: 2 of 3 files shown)


client/src/modules/stock/movements/registry.js
```diff
@@ -90,10 +90,10 @@ function StockMovementsController(
       headerCellFilter : 'translate',
       cellTemplate : 'modules/stock/movements/templates/io.cell.html',
     }, {
-      field : 'flux_id',
+      field : 'fluxName',
       displayName : 'STOCK.FLUX',
       headerCellFilter : 'translate',
-      cellTemplate : 'modules/stock/movements/templates/flux.cell.html',
+      // cellTemplate : 'modules/stock/movements/templates/flux.cell.html',
```
jmcameron: Are we using this template now?  If not, can we delete this line and the template file (flux.cell.html)?

client/src/modules/stock/movements/registry.js
```diff
@@ -228,8 +228,10 @@ function StockMovementsController(
     // the column must be filtered on the translated text
     row.io = $translate.instant(row.is_exit === 0 ? 'STOCK.INPUT' : 'STOCK.OUTPUT');
 
+    const fluxName = $translate.instant(getFluxName(row.flux_id));
     // compute the fluxName from its ID
-    row.fluxName = getFluxName(row.flux_id);
+    row.fluxName = fluxName.concat(row.target ? ` - ${row.target}` : '').trim();
```
jmcameron, line 232: I like having the destination information here.
Would it be possible to do this for purchase orders too?  (include the 'Reference' of the purchase order?)

jmcameron (commented):
The code looks good.

I tested with the IMCK db and it worked fine.

I have a few cosmetic comments about the output--particularly for the summary line at the bottom of the page.   Here are two screenshots for two difference searches.

**Mixed input/output:**
image
- What does the 5 refer to?  There seem to be 4 movements.  If it should be for movements, perhaps can be labled:  "4 stock movements"?

**Input-only Search**
image
- Again the number of movements is off and could use a label.
- The sum at the bottom (center) is fine since it is all inputs but it could use a label too: "Total Cost" (or similar).   In the previous screenshot, the costs are mixed (input/output), so not having a total at the bottom is fine.
- The option flag for "input-only" is not very user-friendly!

I think including labels for the items in the bottom summary line is important to make it clear what the numbers are about -- and because once the list of stock movements gets long enough, you will not be able to see the top row with the column headers so it makes guessing what those summary numbers mean more difficult.

client/src/modules/stock/movements/registry.js
jniles: Good idea.  Will do.  👍

client/src/modules/stock/movements/registry.js
jniles: Done.

jniles:
You've asked about the first consideration previously - it's a feature of `ui-grid` and shows the number of grid rows, not the number of records found.  We have it in all our grids.  I propose that we change it one and for all in all our grids if that is the decision, rather than piecemeal.  The code will likely be copy and paste.

I'm actually against putting the label for the cost summary - many of our screens we work on in the field are tiny and reducing clutter is a positive.  Just telling users that whenever they see a label on the bottom of a currency cell is the sum is easier.  Like having a windows icon instead of menu labeled "start".  Non-obvious to someone who has never used BHIMA, but should be fairly straightforward once you've been trained.

What do you mean by the input-only button isn't user-friendly?  Which button?

jniles:
Other changes landed!

jmcameron:
> What do you mean by the input-only button isn't user-friendly? Which button?

In the "Input-only Search" graphic above, notice the search option flag circled in red in  the top center of the image:   

   [Output: 0  x]

That is what I was referring to.

jmcameron (approved):
Nice fix.
Regarding the grid sums.  I agree they can be fixed later.

jniles:
> In the "Input-only Search" graphic above, notice the search option flag circled in red in the top center of the image:

Ah, this looks like https://github.com/IMA-WorldHealth/bhima/issues/5803.   Do you think that will sort it out, or do we need a specific one for this search field?

Feel free to work on it :)

jmcameron:
> > In the "Input-only Search" graphic above, notice the search option flag circled in red in the top center of the image:
> 
> Ah, this looks like #5803. Do you think that will sort it out, or do we need a specific one for this search field?

Definitely.  I thought of that when I made the comment.  We can defer fixing it in this PR .  I may take a look at the 5803 issue later.

PR #5857, opened 2021-08-16
fix: do not crash on no account references

jniles:
This commit prevents an invalid SQL request from being executed when no account references exist.


lomamech (approved):
@jniles What a coincidence, I had just corrected this error in my very last PR

image

PR #5859, opened 2021-08-17
Improve the appearance of boolean search filter flags

### LINKED ISSUE #5803: Translate boolean values in filter bar

jmcameron:
Implements an improvement in the presentation of the boolean search flags in various registries to make them more user-friendly. The boolean flags now show "Yes"/"Oui" for 1 and "No"/"Non" for 0.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5803

For example, notice the circled flags in the stock lots registry:
image

To make a filter definition work this way, add this to its definition:

   ..., valueFilter : 'boolean'.

This change affects the search flag displays for the following registries.
- Stock > Stock lots
- Stock > Articles in Stock
- Administration > Account References
- Finance > Cash payment registry
- Finance > Patient Invoice registry
- Finance > Voucher registry
- Human Resources > Employee registry
- Inventory > Inventory registry

I attempted to update as many of these boolean filter flags as possible, but I'm sure I missed some.


mbayopanda:
LGTM 👍🏽 

bors r+

PR #5860, opened 2021-08-17
fix(Distribution center)

### LINKED ISSUE #4939: Filter on distribution_center may be buggy
In the test dataset, I filtered on the current year and found records from 2016.  What is the registry filtering on?  Is this a bug or intended behavior?

image

@lomamech do you have any insight?

lomamech:
- Resolve ER_PARSE_ERROR in SQL Syntax, by the adding ' for label
  of service who have Apostrophe
- fix error in params for function getDateRangeFromPeriods
- Correction of the call to the getDateRangeFromPeriods function by
  changing the parameters
- Prevent processing in case the list of accounts is empty

closes #4939

### DIFF
--- a/server/controllers/finance/distributionFeeCenter/configuration.js
+++ b/server/controllers/finance/distributionFeeCenter/configuration.js
@@ -38,7 +38,12 @@ async function configuration(req, res, next) {
 
     // get max and min date from period ids.
     if (query.periodFrom && query.periodTo) {
-      const result = await fiscal.getDateRangeFromPeriods(query.periodFrom, query.periodTo);
+      const periods = {
+        periodFrom : query.periodFrom,
+        periodTo : query.periodTo,
+      }
+
+      const result = await fiscal.getDateRangeFromPeriods(periods);
       options.custom_period_start = result.dateFrom;
       options.custom_period_end = result.dateTo;
     }
--- a/server/controllers/finance/distributionFeeCenter/configuration.js
+++ b/server/controllers/finance/distributionFeeCenter/configuration.js
@@ -19,6 +19,7 @@ async function configuration(req, res, next) {
     };
 
     const accounts = await referenceAccount.auxilliary(params);
+
     const refAccounts = accounts;
     const accountsId = accounts.map(account => account.account_id);
 
--- a/server/controllers/finance/distributionFeeCenter/configuration.js
+++ b/server/controllers/finance/distributionFeeCenter/configuration.js
@@ -12,7 +12,6 @@ const fiscal = require('../fiscal');
 async function configuration(req, res, next) {
   try {
     const { query } = req;
-
     const params = {
       typeFeeCenter : query.typeFeeCenter,
       fee_center_id : query.fee_center_id,
--- a/server/controllers/finance/distributionFeeCenter/configuration.js
+++ b/server/controllers/finance/distributionFeeCenter/configuration.js
@@ -42,7 +42,7 @@ async function configuration(req, res, next) {
       const periods = {
         periodFrom : query.periodFrom,
         periodTo : query.periodTo,
-      }
+      };
 
       const result = await fiscal.getDateRangeFromPeriods(periods);
       options.custom_period_start = result.dateFrom;
--- a/server/controllers/finance/distributionFeeCenter/getDistributed.js
+++ b/server/controllers/finance/distributionFeeCenter/getDistributed.js
@@ -23,8 +23,8 @@ async function getDistributed(req, res, next) {
     // get max and min date from period ids.
     if (periodFrom && periodTo) {
       const periods = {
-        periodFrom : periodFrom,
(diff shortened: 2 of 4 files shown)


jniles:
@lomamech, go ahead and fix these conflicts.  LGTM otherwise.

jniles (approved):

jmcameron (commented):
As long as you're updating this, I suggest changing the English items related to cost centers (since that seems to be the common terminology these days):
- "Fee Center Management" => "Cost Center Management" (main menu)
- "Fee Center" => "Cost Center" (under Fee Center Management menu)
- and other places that "Fee Center" appears.

Perhaps this is for another Issue, but the "Fee Center Report" seems to be restricted to specific months.   Do we want the ability to use a more arbitrary period for this analysis, eg, the last 6 months/quarter/year?  It seems like that flexibility would be important to make better estimates than ones based on a single month, since data over longer periods would not be as volatile..

BTW, I was not sure how to test this.

lomamech:
> As long as you're updating this, I suggest changing the English items related to cost centers (since that seems to be the common terminology these days):
> 
> * "Fee Center Management" => "Cost Center Management" (main menu)
> * "Fee Center" => "Cost Center" (under Fee Center Management menu)
> * and other places that "Fee Center" appears.
> 
> Perhaps this is for another Issue, but the "Fee Center Report" seems to be restricted to specific months. Do we want the ability to use a more arbitrary period for this analysis, eg, the last 6 months/quarter/year? It seems like that flexibility would be important to make better estimates than ones based on a single month, since data over longer periods would not be as volatile..
> 
> BTW, I was not sure how to test this.

@jmcameron  In the French documentation, the notion of Fee Center (Centre des frais) includes the cost center **(centre de coût)** and the profit center **(Centre de profit)**

lomamech:
@jmcameron @jniles  I would only like to know in English the concept of cost centers (Concerns both the cost center and the profit center)

jmcameron:
> @jmcameron In the French documentation, the notion of Fee Center (Centre des frais) includes the cost center **(centre de coût)** and the profit center **(Centre de profit)**

I have to defer to @jniles on this.  But all the English literature I've seen uses "Cost Centers" very prominently for parts of the organization that cost money to operate and does not emphasize "Profit centers" except in the sense that the centers that are potentially generating profit ("Direct") by providing products for the consumers vs those that are "Overhead".   I also find it strange that "Frais" (eg, fee) is applied to profit.

jmcameron:
> As long as you're updating this, I suggest changing the English items related to cost centers (since that seems to be the common terminology these days):
> 
> Perhaps this is for another Issue, but the "Fee Center Report" seems to be restricted to specific months. Do we want the ability to use a more arbitrary period for this analysis, eg, the last 6 months/quarter/year? It seems like that flexibility would be important to make better estimates than ones based on a single month, since data over longer periods would not be as volatile..

These issues could be addressed later in a separate Issue/PR.

lomamech:
> I have to defer to @jniles on this. But all the English literature I've seen uses "Cost Centers" very prominently for parts of the organization that cost money to operate and does not emphasize "Profit centers" except in the sense that the centers that are potentially generating profit ("Direct") by providing products for the consumers vs those that are "Overhead". I also find it strange that "Frais" (eg, fee) is applied to profit.
For French speakers the concept of Fee center is correct, maybe it will only be necessary to use "Cost Center" for the English translation. @jniles  what is your point of view
https://www.beaboss.fr/Definitions-Glossaire/Centre-de-frais-239852.htm

jniles:
@jmcameron @lomamech the name "Fee Center" was a compromise made a long time ago.  @jmcameron is right that we don't have Fee Centers in English.  They are all cost centers in English.

Part of this is related to the differences between how OHADA treats costs and the general accepted accounting principles (GAAP) of the USA treats costs.  In OHADA/PCGC, the purchase of medications is a cost.  In GAAP, it is the consumption of the medications that is the cost.  In GAAP, the purchase of fixed assets is not a cost, but the depreciation of them is a cost.  I'm not sure how OHADA handles this.

I think Larry is probably looking for the GAAP definition of cost centers, so we'll have to make sure we get his approval on whatever we come up with before getting too far.

I haven't finished reading enough of the Cost Center book to figure out the most appropriate terms.  We will probably need to change some of the English around eventually.  I think the French terms are probably appropriate.

jniles:
In any case, this should be more a discussion for an issue, not this PR.  This PR fixes a couple of bugs related to APIs that changed without warning and caused the date search to mess up.  We will address translation issues as we tackle the whole of cost centers.

jniles:
I've folded some of this discussion into https://github.com/IMA-WorldHealth/bhima/issues/5861, please add anything that might be missing @lomamech @jmcameron .

jniles (approved):
Merging as this addresses the bugs reported in the issue and has clean code.

bors r+

PR #5867, opened 2021-08-19
Support selection of currency in Fee Center Report

### LINKED ISSUE #5862: Fee Center Report should allow users to select a currency
We should allow users to select a currency on the Fee Center report.  Further, it looks like the currency rendering is broken for the "Distribution by Profit Center" section:

image

jmcameron:
Updated the Fee Center Report to support the selection of the currency for the report.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5862

### DIFF
--- a/client/src/modules/reports/generate/fee_center/fee_center.config.js
+++ b/client/src/modules/reports/generate/fee_center/fee_center.config.js
@@ -2,18 +2,20 @@ angular.module('bhima.controllers')
   .controller('fee_centerController', FeeCenterConfigController);
 
 FeeCenterConfigController.$inject = [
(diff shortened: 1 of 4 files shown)


jniles:
This isn't quite accurate, since we need to do some exchange rate conversion of the values.  It should probably use today's date for the exchange rate conversion.  I believe all values are in the enterprise currency.

jmcameron:
> This isn't quite accurate, since we need to do some exchange rate conversion of the values. It should probably use today's date for the exchange rate conversion. I believe all values are in the enterprise currency.

Fixed.

jniles:
This all looks good to me.

bors r+

PR #5871, opened 2021-08-23
Fix tag handling in inventory changes report

### LINKED ISSUE #5868: Properly detect modifications to tags
In the inventory changes report, the tags are likely being compared with an equality `===` which doesn't work for JavaScript arrays.  It makes it look like something is changing when it isn't.  We should update this to not make it look like tags are changing when they are not.  See below.

image

jmcameron:
Fixes several issues with inventory tags editing and displaying history of inventory tag changes.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5868

This PR fixes these problems
- Fixes the display of changes to inventory tags in both the action menu for inventory items and for the Inventory Changes Report.
- Fixed editing of inventory tags:  Formerly it was possible to add only a single tag to an inventory item.  Once the code was fixed to address that problem, the Inventory registry would return multiple rows for a single inventory item if it had more than one tag (due to the LEFT JOIN).  This was changed so that a secondary query is done to add the tags information for inventory items.

NOTE: To reduce the number of secondary queries to get tags for inventory items I use a GROUP_CONCAT SQL function to return the list of tags and tag data as a string.  To parse the string on the client side, I use the semicolon (;) and vertical bar (|) characters to parse the tag data and construct a tag data.   Therefore, we do not allow those characters to be used in tags.

*TESTING*
- Use any database. It may be easier with bhima_test
- Select an inventory item (it may be easier to use the Search function to narrow the display to one item).
- Using the action menu, edit the inventory item.
   - Add a tag then save (verify the tag was added in the main registry display)
   - Try various combinations of adding and removing tags until you are satisfied that editing the tags works correctly.
   - Make sure you do at least one edit where you change both tags and something else (eg, price) in one edit.
   - In the action menu, use the "Changes" function to see the list of changes and make sure they make sense.
   - Verify that the "Changes" listing shows the most recent changes first.
- Verify that searching by tags works in the inventory registry
- Try the Inventory > Inventory Changes Report and verify that is correct.
   - Verify that the listing shows the most recent changes first.

### DIFF
--- a/client/src/modules/inventory/list/modals/log.modal.js
+++ b/client/src/modules/inventory/list/modals/log.modal.js
@@ -29,6 +29,18 @@ function InventoryLogModalController(data, Instance, Inventory, LanguageService)
       } else {
         const updatedKeys = Object.keys(current);
         updatedKeys.forEach(col => {
+          if (col === 'tags') {
+            const oldTags = last[col];
+            const newTags = current[col];
+            if (oldTags.length === newTags.length) {
+              if (newTags.length === 0) {
+                return;
+              }
+              if (JSON.stringify(oldTags) === JSON.stringify(newTags)) {
+                return;
+              }
+            }
+          }
           vm.logs.push({
             col : Inventory.columnsMap(col),
             value : getValue(last, current, col),
@@ -55,6 +67,12 @@ function InventoryLogModalController(data, Instance, Inventory, LanguageService)
 
   function getValue(last, current, key) {
     const result = {};
+
+    if (key === 'tags') {
+      result.from = last.tags.map(tag => tag.name) || [];
+      result.to = current.tags.map(tag => tag.name) || [];
+      return result;
+    }
 
     if (key === 'inventoryGroup') {
       result.from = last.groupName || '';
--- a/client/src/modules/tags/modal/createUpdate.js
+++ b/client/src/modules/tags/modal/createUpdate.js
@@ -28,6 +28,14 @@ function TagsModalController(
       return false;
     }
 
+    // Do not allow semicolons or vertical bars in tag names since they will be
+    // used to separate the parts of the tags when retrieving them from the back end.
+    if (vm.tags.name.includes(';') || vm.tags.name.includes('|')) {
+      form.name.$invalid = true;
+      form.name.$valid = false;
+      return false;
+    }
+
     [].concat(vm.tags).forEach(tag => {
       delete tag.style;
     });
--- a/server/controllers/inventory/inventory/core.js
+++ b/server/controllers/inventory/inventory/core.js
@@ -43,6 +43,7 @@ const inventoryColsMap = {
   group_uuid : 'FORM.LABELS.GROUP',
   inventoryGroup : 'FORM.LABELS.GROUP',
   label : 'FORM.LABELS.LABEL',
+  tags : 'TAG.TAGS',
   text : 'FORM.LABELS.LABEL',
   note : 'FORM.INFO.NOTE',
   price : 'FORM.LABELS.UNIT_PRICE',
@@ -132,6 +133,7 @@ async function loadGroupAndType(record) {
 
   return data;
 }
+
 /**
 * Update inventory metadata in the database
 *
@@ -167,21 +169,24 @@ async function updateItemsMetadata(record, identifier, session) {
     transaction.addQuery(sql, [record, db.bid(identifier)]);
   }
 
-  if (tags && tags.length) {
-    tags.forEach(t => {
-      const binaryTagUuid = db.bid(t.uuid);
-      const queryAddTags = 'INSERT INTO inventory_tag(inventory_uuid, tag_uuid) VALUES (?, ?);';
-      const queryClearPreviousTags = 'DELETE FROM inventory_tag WHERE inventory_uuid = ?';
-      transaction.addQuery(queryClearPreviousTags, [db.bid(identifier)]);
-      transaction.addQuery(queryAddTags, [db.bid(identifier), binaryTagUuid]);
-    });
+  if (tags) {
+    const inventoryUuid = db.bid(identifier);
+    const queryClearPreviousTags = 'DELETE FROM inventory_tag WHERE inventory_uuid = ?';
+    transaction.addQuery(queryClearPreviousTags, [inventoryUuid]);
+    if (tags.length) {
+      tags.forEach(t => {
+        const binaryTagUuid = db.bid(t.uuid);
+        const queryAddTags = 'INSERT INTO inventory_tag(inventory_uuid, tag_uuid) VALUES (?, ?);';
+        transaction.addQuery(queryAddTags, [inventoryUuid, binaryTagUuid]);
+      });
+    }
   }
 
   // write into inventory log all changes
-  if (!_.isEmpty(record) || tags.length) {
+  if (!_.isEmpty(record) || typeof tags !== 'undefined') {
     const lastInfo = await getItemsMetadataById(identifier);
-    const othersKyes = await loadGroupAndType(record);
-    const current = _.extend({}, recordCopy, othersKyes);
+    const othersKeys = await loadGroupAndType(record);
+    const current = _.extend({}, recordCopy, othersKeys);
 
     transaction.addQuery(inventoryLogSql, {
       uuid : db.uuid(),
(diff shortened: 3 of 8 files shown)


client/src/modules/inventory/list/modals/log.modal.js
```diff
@@ -29,6 +29,19 @@ function InventoryLogModalController(data, Instance, Inventory, LanguageService)
       } else {
         const updatedKeys = Object.keys(current);
         updatedKeys.forEach(col => {
+          if (col === 'tags') {
+            // Note: This is a work-around to deal with incorrectly logged tag changes
+            const oldTags = last[col];
+            const newTags = current[col];
+            if (oldTags.length === newTags.length) {
+              if (newTags.length === 0) {
+                return;
```
jniles, line 37: What happens when I remove a tag?  It looks like this isn't handled.

I think we should store an empty string for the new tags if their length is 0, but old tags are greater than 0. That should probably handle this case.

server/controllers/inventory/inventory/core.js
```diff
@@ -167,21 +168,24 @@ async function updateItemsMetadata(record, identifier, session) {
...
+      });
+    }
   }
 
   // write into inventory log all changes
   if (!_.isEmpty(record) || tags.length) {
     const lastInfo = await getItemsMetadataById(identifier);
-    const othersKyes = await loadGroupAndType(record);
-    const current = _.extend({}, recordCopy, othersKyes);
+    const othersKeys = await loadGroupAndType(record);
```
jniles, line 188: Good fix 👍

server/controllers/inventory/inventory/core.js
```diff
@@ -278,10 +281,36 @@ async function getItemsMetadata(params) {
...
   const query = filters.applyQuery(sql);
   const parameters = filters.parameters();
   const response = await db.exec(query, parameters);
 
+  // Construct the tags for each inventory item
+  const queryTags = `
+    SELECT BUID(t.uuid) AS uuid, t.name, t.color
+    FROM tags t
+    JOIN inventory_tag it ON it.tag_uuid = t.uuid
+    WHERE it.inventory_uuid = ?;
```
jniles: Hmm... is there any logic that depends on having the uuid and color for each tag?  It would be more efficient to use a `GROUP_CONCAT()`  instead of firing of N queries for N inventory items that have tags.

https://www.mysqltutorial.org/mysql-group_concat/

jniles (commented):
It doesn't look like the tag changes are saved to the log when tags are deleted.  We probably want this information.

Also, it looks like the query for inventory items may potentially spawn a lot of sub-queries.  I have suggested using `GROUP_CONCAT` for this.  I'm actually even okay with doing a couple `GROUP_CONCAT()`s and parsing the rest in JS, just to avoid all the round trips to the database.  So, if you need the color and name, just do two concats, then stitch them together in JS afterwards.

Otherwise, performs as expected!  Thanks!

client/src/modules/inventory/list/modals/log.modal.js
jmcameron: Fixed

server/controllers/inventory/inventory/core.js
jmcameron: Done.

jmcameron:
@jniles Please check the changes I did and see if they are okay.  If you give the go-ahead I'll get Bruce or Chris to review it.

client/src/modules/tags/modal/createUpdate.js
```diff
@@ -28,6 +28,14 @@ function TagsModalController(
       return false;
     }
 
+    // Do not allow semicolons or vertical bars in tag names since they will be
+    // used to separate the parts of the tags when retrieving them from the back end.
+    if (vm.tags.name.indexOf(';') > -1 || vm.tags.name.indexOf('|') > -1) {
```
jniles: Clever!

Just as an FYI you can also use [`includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) to detect a character in a string.  It makes it less C-like and more JS like.

server/controllers/inventory/inventory/core.js
```diff
@@ -167,21 +169,24 @@ async function updateItemsMetadata(record, identifier, session) {
...
-    tags.forEach(t => {
-      const binaryTagUuid = db.bid(t.uuid);
-      const queryAddTags = 'INSERT INTO inventory_tag(inventory_uuid, tag_uuid) VALUES (?, ?);';
-      const queryClearPreviousTags = 'DELETE FROM inventory_tag WHERE inventory_uuid = ?';
-      transaction.addQuery(queryClearPreviousTags, [db.bid(identifier)]);
-      transaction.addQuery(queryAddTags, [db.bid(identifier), binaryTagUuid]);
-    });
+  if (tags) {
+    const inventoryUuid = db.bid(identifier);
+    const queryClearPreviousTags = 'DELETE FROM inventory_tag WHERE inventory_uuid = ?';
```
jniles, line 174: Excellent catch 👍

server/controllers/inventory/inventory/core.js
```diff
@@ -250,15 +255,21 @@ async function getItemsMetadata(params) {
       ig.sales_account, ig.stock_account, ig.donation_account, inventory.sellable, inventory.note,
       inventory.unit_weight, inventory.unit_volume, ig.sales_account, ig.stock_account, ig.donation_account,
       ig.cogs_account, inventory.default_quantity, ig.tracking_consumption, ig.tracking_expiration,
-      inventory.importance, t.name AS tag_name, t.color AS tag_color,
+      inventory.importance,
+      (SELECT GROUP_CONCAT(BUID(t2.uuid), ';', t2.name, ';', t2.color
```
jniles: Do you really need a nested select?  What happens if you just call `GROUP_CONCAT()` in the regular select statement?

server/controllers/inventory/inventory/core.js
```diff
@@ -278,10 +289,29 @@ async function getItemsMetadata(params) {
...
+
   const query = filters.applyQuery(sql);
   const parameters = filters.parameters();
   const response = await db.exec(query, parameters);
 
+  // Decode the tags for each inventory item with tags
+  response.forEach(inv => {
+    if (inv.tag_details) {
+      inv.tags = [];
+      const rawTags = inv.tag_details.split('|');
```
jniles, line 302: I wonder if we could choose some sort of unprintable character that users wouldn't have access to - like a UTF8 Byte Order Mark or something.

server/controllers/inventory/reports/changes.js
```diff
@@ -113,6 +111,14 @@ function formatKeys(record) {
 function getValue(last, current, key) {
   const result = {};
 
+  if (key === 'tags') {
+    const lastTags = last.tags.map(tag => `"${tag.name}"`).join(',  ');
+    result.from = last.tags.length ? `[${lastTags}]` : '[]';
```
jniles: I think you may not need this length check.  It will be an empty string by default.
```js
    result.from =  `[${lastTags}]` 
 ```
 Should work fine.

jniles (commented):
@jmcameron just some questions, things look fine to merge for me.   Happy to kick it over to someone else for a review.

server/controllers/inventory/inventory/core.js
jmcameron: Good catch.  I was able to get this working without the nested select.

server/controllers/inventory/inventory/core.js
jmcameron: I would prefer not to do that.  It will make it more difficult for anyone doing manual mysql queries.   Non-printable characters can be tricking to handle in terminals.  Second, I personally do not feel that tags should have semicolons or vertical bars.  It seems to me that they should be simple tokens.  I decided not to filter out commas, since I could imagine situations where a comma could be useful in a tag name (and that is why I did not use them for the parsing separators).

client/src/modules/tags/modal/createUpdate.js
jmcameron: Done

server/controllers/inventory/reports/changes.js
jmcameron: Done.  And it works!

lomamech (approved):
image

I have just tested the change, and I must admit that it is for the very first time that I discover this functionality

lomamech:
LGTM
bors r+

server/controllers/inventory/inventory/core.js
jniles: It is worth noting the French excel uses semicolons instead of commas, so there may be folks who put semicolons in the tags, just because of language differences.

But indeed, I'm find leaving this as is until we get a feature request about it.

jmcameron: > It is worth noting the French excel uses semicolons instead of commas, so there may be folks who put semicolons in the tags, just because of language differences.

Good point.  I'll add an Issue about it.  However, picking the separator characters is tricky.

PR #5877, opened 2021-08-24
Use of decimal for stock minimum months

### LINKED ISSUE #5876: Allow decimal figures for reorder interval and min months of stock
The Vanga team needs to be able to reorder on a weekly basis.  We currently use months as our basis.  It's possible to make this into weeks using a decimal, and the math should work.  However, the database needs to be modified and the inputs modified in both the stock settings and depot settings to allow for this.  See below:

image

mbayopanda:
This PR is related to #5876

Close #5876


client/src/modules/depots/modals/depot.modal.html
```diff
@@ -129,8 +129,7 @@
         FORM.LABELS.MIN_MONTHS_SECURITY_STOCK
       </label>
       <input name="min_months_security_stock" class="form-control"
-        ng-model="DepotModalCtrl.depot.min_months_security_stock" type="number" min="1" autocomplete="off" required
-        bh-integer>
+        ng-model="DepotModalCtrl.depot.min_months_security_stock" type="number" min="0" autocomplete="off" required>
```
jmcameron: We should consider making min="0.00001" (or similar) to prevent acceptance of zero.

jmcameron (changes requested):
I think you need to also do similar input form modifications on Stock > Stock Settings  (client/src/modules/stock/settings/stock-settings.html).

jniles:
bors try

client/src/modules/stock/settings/stock-settings.html
```diff
@@ -24,7 +24,7 @@
             </label>
             <input type="number" class="form-control" autocomplete="off" name="month_average_consumption"
               ng-model="StockSettingsCtrl.settings.month_average_consumption"
-              ng-change="StockSettingsCtrl.setMonthAverage(value)" min="1" bh-integer>
+              ng-change="StockSettingsCtrl.setMonthAverage(value)" min="0.00001">
```
jmcameron, line 27: Not sure it matters, but should these include:  type="number" 
?

jmcameron (commented):
The code changes look good (see my notes).

I tested this with bhima_test.  Changing the values in stock settings works fine.  

However when I tried it with the Depot editor, it seems like the values are set properly, but when I edit again, I see that the value "Default Purchase order interval (in months)" is still 0 (I set it to 0.25).  See:
image

client/src/modules/stock/settings/stock-settings.html
jniles: They do - you missed it.  It's up above.

server/models/schema.sql
```diff
@@ -461,7 +461,7 @@ CREATE TABLE `depot` (
   `allow_exit_transfer` TINYINT(1) UNSIGNED NOT NULL DEFAULT 0,
   `allow_exit_loss` TINYINT(1) UNSIGNED NOT NULL DEFAULT 0,
   `location_uuid` BINARY(16) NULL,
-  `min_months_security_stock` SMALLINT(5) NOT NULL DEFAULT 2,
+  `min_months_security_stock` DECIMAL(19,4) NOT NULL DEFAULT 2,
   `parent_uuid` BINARY(16) NULL,
   `dhis2_uid` VARCHAR(150) DEFAULT NULL,
   `default_purchase_interval` SMALLINT(5) NOT NULL DEFAULT 0,
```
jniles: This needs to be DECIMAL as well.  That is the problem @jmcameron ran into.

mbayopanda: I had done changes only in migrate.js

jniles (approved):
This worked for me.

bors r+

PR #5885, opened 2021-08-26
Feat: add aggregate cost center table

### LINKED ISSUE #5880: Add `cost_center_id` and `principal_center_id` to the posting_journal and general_ledger
In order to rapidly associate costs with their correct cost centers and principal cost centers, we'll need two extra columns in the posting_journal and general_ledger to track the cost center associated the transaction.  This has several advantages:

1. Finding the costs/revenues directly with a center is as simple as `SELECT SUM(debit_equiv) FROM general_ledger WHERE cost_center_id = ?;` . We will not need to go through the `account_reference` tables to find the cost centers.
2. We can use our usual tools (editing transactions, editing journal vouchers) to associate these cost centers with a new transaction.
3. We will be able to search/filter directly on cost center in the posting journal.
4. We can create aggregate tables (like the `period_total` table) for the cost center values during the posting process to speed up calculations later.

This means we can remove the `fee_center_distribution` table, since it currently plays this role.

### LINKED ISSUE #5881: Create a `cost_center_aggregate` table to store the aggregated costs of cost centers
The overview of Dr. Larry's cost center analysis design is two steps:

1. Get total costs by cost center
2. Use the step-down method to allocate costs from auxiliary centers one by one onto the revenue-generating centers

The first step is partially solved by https://github.com/IMA-WorldHealth/bhima/issues/5880, but we need a way to aggregate the costs for the step-down method to work.

I propose we make a table called `cost_center_aggregate` that functions similar to BHIMA's `period_total` table and that looks like this:

period_id | debit | credit | cost_center_id | principal_center_id?
-- | -- | -- | -- | --
202108 | 15 | 0 | medicaments | maternite
202108 | 25 | 0 | medicaments | chirurgie
202108 | 70 | 0 | charges perte divers | chirurgie
202108 | 1500 | 0 | impots | NULL

The basic idea is that we will store the aggregated cost by period in the `cost_center_aggregate` table to speed up calculations.  This table will be populated during the posting process with values from #5880.

The `cost_center_id` will be required and cannot be NULL.   The `principal_center_id` can be NULL and represents indirect costs that must be allocated.  All other costs will not need to be allocated since they are already associated with their `principal_center_id` and can be directly absorbed.

jniles:
This PR adds the back-end tables for aggregating costs by cost center and the stored procedure for generating it.  It should serve as a reference for how to create the table for step-down analysis.

Closes #5880.
Closes #5881.

### DIFF
--- a/server/models/migrations/next/migrate.sql
+++ b/server/models/migrations/next/migrate.sql
@@ -1,0 +1,35 @@
+/* migrate v1.21.0 to next */
+
+/**
+author: @jniles
+date: 2021-08-30
+description: adds cost columns and indices to relevant tables
+*/
+
+DROP TABLE IF EXISTS `cost_center_aggregate`;
+CREATE TABLE `cost_center_aggregate` (
+  `period_id`       MEDIUMINT(8) UNSIGNED NOT NULL,
+  `debit`           DECIMAL(19,4) UNSIGNED NOT NULL DEFAULT 0.00,
+  `credit`          DECIMAL(19,4) UNSIGNED NOT NULL DEFAULT 0.00,
+  `cost_center_id`  MEDIUMINT(8) UNSIGNED NOT NULL,
+  `principal_center_id` MEDIUMINT(8) UNSIGNED NULL,
+  KEY `cost_center_id` (`cost_center_id`),
+  KEY `principal_center_id` (`principal_center_id`),
+  KEY `period_id` (`period_id`),
+  CONSTRAINT `cost_center_aggregate__period` FOREIGN KEY (`period_id`) REFERENCES `period` (`id`),
+  CONSTRAINT `cost_center_aggregate__cost_center_id` FOREIGN KEY (`cost_center_id`) REFERENCES `fee_center` (`id`),
+  CONSTRAINT `cost_center_aggregate__principal_center_id` FOREIGN KEY (`principal_center_id`) REFERENCES `fee_center` (`id`)
+) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+CALL add_column_if_missing('posting_journal', 'cost_center_id', 'MEDIUMINT(8) UNSIGNED NULL');
+CALL add_column_if_missing('posting_journal', 'principal_center_id', 'MEDIUMINT(8) UNSIGNED NULL');
+
+ALTER TABLE `posting_journal` ADD CONSTRAINT `pg__cost_center_1` FOREIGN KEY (`cost_center_id`) REFERENCES `fee_center` (`id`) ON UPDATE CASCADE;
+ALTER TABLE `posting_journal` ADD CONSTRAINT `pg__cost_center_2` FOREIGN KEY (`principal_center_id`) REFERENCES `fee_center` (`id`) ON UPDATE CASCADE;
+
+
+CALL add_column_if_missing('general_ledger', 'cost_center_id', 'MEDIUMINT(8) UNSIGNED NULL');
+CALL add_column_if_missing('general_ledger', 'principal_center_id', 'MEDIUMINT(8) UNSIGNED NULL');
+
+ALTER TABLE `general_ledger` ADD CONSTRAINT `general_ledger__cost_center_1` FOREIGN KEY (`cost_center_id`) REFERENCES `fee_center` (`id`) ON UPDATE CASCADE;
+ALTER TABLE `general_ledger` ADD CONSTRAINT `general_ledger__cost_center_2` FOREIGN KEY (`principal_center_id`) REFERENCES `fee_center` (`id`) ON UPDATE CASCADE;
--- a/server/models/schema.sql
+++ b/server/models/schema.sql
@@ -2580,7 +2580,7 @@ CREATE TABLE `cost_center_aggregate` (
   `debit`           DECIMAL(19,4) UNSIGNED NOT NULL DEFAULT 0.00,
   `credit`          DECIMAL(19,4) UNSIGNED NOT NULL DEFAULT 0.00,
   `cost_center_id`  MEDIUMINT(8) UNSIGNED NOT NULL,
-  `principal_center_id` MEDIUMINT(8) UNSIGNED NOT NULL,
+  `principal_center_id` MEDIUMINT(8) UNSIGNED NULL,
   KEY `cost_center_id` (`cost_center_id`),
   KEY `principal_center_id` (`principal_center_id`),
   KEY `period_id` (`period_id`),
--- a/server/models/schema.sql
+++ b/server/models/schema.sql
@@ -528,10 +528,10 @@ CREATE TABLE `employee` (
   KEY `creditor_uuid` (`creditor_uuid`),
   KEY `grade_uuid` (`grade_uuid`),
   KEY `patient_uuid` (`patient_uuid`),
-  CONSTRAINT  `employee__fonction` FOREIGN KEY (`fonction_id`) REFERENCES `fonction` (`id`),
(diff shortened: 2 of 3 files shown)


server/models/schema.sql
```diff
@@ -2561,4 +2573,20 @@ CREATE TABLE `configuration_analysis_tools` (
   CONSTRAINT `config_analysis_tools__analysis_tool_type` FOREIGN KEY (`analysis_tool_type_id`) REFERENCES `analysis_tool_type` (`id`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
+
+DROP TABLE IF EXISTS `cost_center_aggregate`;
```
lomamech, line 2577: Will it be possible to rename this table fee_center_aggregate or cost_profit_center_aggregate, Just because this way it will make it look like this table is not about profit centers

server/models/schema.sql
```diff
@@ -1315,6 +1323,8 @@ CREATE TABLE `posting_journal` (
   KEY `period_id` (`period_id`),
   KEY `currency_id` (`currency_id`),
   KEY `user_id` (`user_id`),
+  KEY `cost_center_id` (`cost_center_id`),
```
lomamech, line 1326: Will it be possible to rename this property fee_center_id or cost_profit_center_id

lomamech (commented):
I will add other comments during the Pull Request test

server/models/schema.sql
jniles: I was implementing the proposal from https://github.com/IMA-WorldHealth/bhima/issues/5861.  I think in English, all off the fee centers are called cost centers.  But I could be wrong.

@jmcameron has read more of the documentation than i have.  @jmcameron do you have a thought on what the best name is?

lomamech: Let's wait for his feedback

jmcameron: > @jmcameron has read more of the documentation than i have. @jmcameron do you have a thought on what the best name is?

I suggest "Cost Centers" for all services and cost/profit centers.  In a sense this is appropriate:  All services or centers have costs.   Some centers can generate revenue.  These are usually also called revenue centers.  But they are still cost centers.  Regarding the "Adminstration" center at IMCK:  This should probably be split up into a Registration Service which could be treated as revenue center.  But Administration is normally a cost center NOT a revenue center.

The term "Cost Center" is not explicitly used in the "Essentials of Healthcare Finance" book (we have its chapter 13).  But "Cost Center" is used in the "Annex B" handbook.   AND "Cost Center" is used in the list of things Larry wants us to do.  So he uses this terminology.    So I encourage us to use "Cost Center" as the general term for all services/cost/profit/revenue centers.

lomamech: image
Centre de frais = Cost Center

lomamech:
@jniles You forgot to put the changes in the migrate.sql file

jniles:
@lomamech @jmcameron changes landed.

lomamech:
image

After entering the cost centers manually, calling the function works fine

lomamech:
@jmcameron what do you think of this PR i think we can merge it up

server/models/migrations/next/migrate.sql
```diff
@@ -0,0 +1,36 @@
...
+  KEY `cost_center_id` (`cost_center_id`),
+  KEY `principal_center_id` (`principal_center_id`),
+  KEY `period_id` (`period_id`),
+  CONSTRAINT `cost_center_aggregate__period` FOREIGN KEY (`period_id`) REFERENCES `period` (`id`),
+  CONSTRAINT `cost_center_aggregate__cost_center_id` FOREIGN KEY (`cost_center_id`) REFERENCES `fee_center` (`id`),
+  CONSTRAINT `cost_center_aggregate__principal_center_id` FOREIGN KEY (`principal_center_id`) REFERENCES `fee_center` (`id`)
+) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+
+ALTER TABLE `posting_journal` ADD COLUMN `cost_center_id`      MEDIUMINT(8) UNSIGNED NULL;
```
jmcameron: I suggest you use the add_column_if_missing() function to add these new columns

server/models/schema.sql
```diff
@@ -676,6 +678,8 @@ CREATE TABLE `general_ledger` (
...
+  KEY `principal_center_id` (`principal_center_id`),
```
jmcameron, line 682: Do we really need a separate key for the principal center id?  I presume the const center ("fee center") table will include a flag as to whether it is principal or not.  So a simple join will give a flag that indicates whether it is a principal center, etc.

EDIT:  I went back and looked at https://github.com/IMA-WorldHealth/bhima/issues/5881.  I know understand why they are both there.

server/models/schema.sql
```diff
@@ -2561,4 +2573,20 @@ CREATE TABLE `configuration_analysis_tools` (
   CONSTRAINT `config_analysis_tools__analysis_tool_type` FOREIGN KEY (`analysis_tool_type_id`) REFERENCES `analysis_tool_type` (`id`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
+
+DROP TABLE IF EXISTS `cost_center_aggregate`;
+CREATE TABLE `cost_center_aggregate` (
+  `period_id`       MEDIUMINT(8) UNSIGNED NOT NULL,
```
jmcameron, line 2579: What "periods" are these?  Will we have the flexibility to do aggregates over months, quarters, years, and arbitrary date ranges?  I see in the Bhima interface that some places 'periods' refer to months of a fiscal year.   While months may be adequate, I could easily imagine wanting cost/revenue aggregates over more arbitrary periods including quarters, years, and the immediately preceeding 6 months, etc.

jmcameron (commented):
I wonder whether this is a good time to rename the 'fee_center" table to 'cost_center'?  It will only get more complicated as we add functionality?

server/models/schema.sql
jniles: You are correct - period is a month.  The goal of these tables (period_total, for example) is to reduce the number of records we need to sum over.  But we can still sum for greater periods than just a month.  If you need finer-grained resolution (e.g. you are mid-month), you must do the summing yourself on the general_ledger.  Some of this logic is shown in the [accounts/extra.js](https://github.com/IMA-WorldHealth/bhima/blob/39c32b0522f335ebcded814b70d69c660524a446/server/controllers/finance/accounts/extra.js#L123) file.

server/models/migrations/next/migrate.sql
jniles: Whoops, good point.

jniles:
@jmcameron changes landed.  I tested these on the vanga database.

For the renaming of the fee_center, let's get this one in first and then I'll work on renaming the fee center table.  I think that may take a lot longer than this.

server/models/schema.sql
jmcameron: Ok.

jmcameron:
> For the renaming of the fee_center, let's get this one in first and then I'll work on renaming the fee center table. I think that may take a lot longer than this.

Sounds good.

jmcameron (approved):
Looks good.

One question (not a blocker): Can 'IGNORE' be used with the ADD CONSTRAINT lines in the migrate.sql file (so it can be loaded more than once).  I see that we have an SQL function Constraint_exists() so it would not be hard to create a function like add_column_if_missing() function for constraints.

jniles:
There is no `IGNORE` for `ALTER TABLE` that I can tell:  https://dev.mysql.com/doc/refman/8.0/en/alter-table.html

This would therefore need to be a custom stored procedure.  I'll make an issue.

PR #5894, opened 2021-08-31
Add the stepdown algorithm in lib

mbayopanda:
This PR adds the step down algorithm in lib and adds a test for the algo


jmcameron (approved):
It looks good for now.
We will probably have to restructure it later when we actually start applying it to the BHIMA data.  For instance the report will probably have to include selection for the period of coverage, etc.

PR #5905, opened 2021-09-03
Fee center step down report

mbayopanda:
First implementation of the Fee Center Report with the step-down algorithm based on test data in `migrate.sql`

image

The report : 

image

Be free for proposals to fix : 
- Error due to missing data in the procedure `FeeCenterCostWithIndexes`
- Other issues found
- Report presentation
- Etc.

Thank you

### DIFF
--- a/client/src/modules/reports/generate/fee_center_step_down/fee_center_step_down.config.js
+++ b/client/src/modules/reports/generate/fee_center_step_down/fee_center_step_down.config.js
@@ -1,0 +1,77 @@
+angular.module('bhima.controllers')
+  .controller('fee_center_step_downController', FeeCenterStepdownReportConfigController);
+
+FeeCenterStepdownReportConfigController.$inject = [
+  '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state',
+];
+
+/**
+ * @function FeeCenterStepdownReportConfigController
+ *
+ * @description
+ * This function renders the fee_center_step_down report.
+ */
+function FeeCenterStepdownReportConfigController($sce, Notify, SavedReports, AppCache, reportData, $state) {
+  const vm = this;
+  const cache = new AppCache('FeeCenterStepdownReport');
+  const reportUrl = 'reports/finance/fee_center_step_down';
+
+  vm.previewGenerated = false;
+  vm.reportDetails = {};
+
+  vm.onSelectFiscalYear = (fiscalYear) => {
+    vm.reportDetails.fiscal_id = fiscalYear.id;
+  };
+
+  vm.onSelectPeriodFrom = function onSelectPeriodFrom(period) {
+    vm.reportDetails.periodFrom = period.id;
+  };
+
+  vm.onSelectPeriodTo = function onSelectPeriodTo(period) {
+    vm.reportDetails.periodTo = period.id;
+  };
+
+  vm.clearPreview = function clearPreview() {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.preview = function preview(form) {
+    if (form.$invalid) {
+      Notify.danger('FORM.ERRORS.RECORD_ERROR');
+      return 0;
+    }
+
+    // update cached configuration
+    cache.reportDetails = angular.copy(vm.reportDetails);
+
+    return SavedReports.requestPreview(reportUrl, reportData.id, angular.copy(vm.reportDetails))
+      .then(result => {
+        vm.previewGenerated = true;
+        vm.previewResult = $sce.trustAsHtml(result);
+      })
+      .catch(Notify.handleError);
+  };
+
+  vm.requestSaveAs = function requestSaveAs() {
+    const options = {
+      url : reportUrl,
+      report : reportData,
+      reportOptions : angular.copy(vm.reportDetails),
+    };
+
+    return SavedReports.saveAsModal(options)
+      .then(() => {
+        $state.go('reportsBase.reportsArchive', { key : options.report.report_key });
+      })
+      .catch(Notify.handleError);
+  };
+
+  checkCachedConfiguration();
+
+  function checkCachedConfiguration() {
+    if (cache.reportDetails) {
+      vm.reportDetails = angular.copy(cache.reportDetails);
+    }
+  }
+}
--- a/server/controllers/finance/reports/fee_center_step_down/index.js
+++ b/server/controllers/finance/reports/fee_center_step_down/index.js
@@ -1,0 +1,161 @@
+const _ = require('lodash');
+const db = require('../../../../lib/db');
+const Stepdown = require('../../../../lib/stepdown');
+const ReportManager = require('../../../../lib/ReportManager');
+const fiscal = require('../../fiscal');
+
+const TEMPLATE = './server/controllers/finance/reports/fee_center_step_down/report.handlebars';
+
+// REMOVE ME
+const MOCK = require('./mock_data');
+// END REMOVE ME
+
+// expose to the API
+exports.report = document;
+
+/**
+ * @function reporting
+ *
+ * @description
+ * Renders the fee center report
+ *
+ * @param {*} options the report options
+ * @param {*} session the session
+ */
+async function buildReport(params, session) {
+  // REMOVE ME
+  await MOCK.mount();
+  // END REMOVE ME
+
+  const options = _.extend(params, {
+    filename : 'TREE.FEE_CENTER_STEPDOWN',
+    csvKey : 'rows',
+    user : session.user,
+  });
+
+  const report = new ReportManager(TEMPLATE, session, options);
+  const enterpriseCurrencyId = session.enterprise.currency_id;
+
+  const periods = {
+    periodFrom : params.periodFrom,
+    periodTo : params.periodTo,
+  };
+
+  const range = await fiscal.getDateRangeFromPeriods(periods);
+  const query = 'CALL FeeCenterCostWithIndexes(?, ?);';
+  let [feeCenters] = await db.exec(query, [range.dateFrom, range.dateTo]);
+
+  if (feeCenters.length) {
+    const [single] = feeCenters;
+    feeCenters = single.error_message ? [] : feeCenters;
+  }
+
+  const formattedFeeCenters = feeCenters.map(item => {
+    item.principal = !!item.is_principal;
+    item.auxiliary = !item.principal;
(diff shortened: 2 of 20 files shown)


server/models/migrations/next/migrate.sql
```diff
@@ -42,3 +42,62 @@ CALL add_column_if_missing('general_ledger', 'principal_center_id', 'MEDIUMINT(8
...
 ALTER TABLE `general_ledger` ADD CONSTRAINT `general_ledger__cost_center_2` FOREIGN KEY (`principal_center_id`) REFERENCES `fee_center` (`id`) ON UPDATE CASCADE;
+
+
+/*
+ * @author: mbayopanda
+ * @date: 2021-09-02
+ * @desc: fee center report tables and test data
+ */
+ALTER TABLE `fee_center` ADD COLUMN `step_order` SMALLINT(5) NOT NULL DEFAULT 0;
+ALTER TABLE `fee_center` ADD COLUMN `default_fee_center_index_id` MEDIUMINT(8) UNSIGNED NULL;
```
jniles: What is the `default_fee_center_id` for?

server/models/migrations/next/migrate.sql
```diff
@@ -42,3 +42,62 @@ CALL add_column_if_missing('general_ledger', 'principal_center_id', 'MEDIUMINT(8
...
+  (14, 'Cutting', 1, 4, NULL),
+  (15, 'Assembling', 1, 5, NULL),
+  (16, 'Packaging', 1, 6, NULL);
+
+INSERT INTO `fee_center_index_value` VALUES 
+  (1, 10, 8), (1, 11, 5), (1, 12, 9), (1, 13, 4), (1, 14, 20), (1, 15, 50), (1, 16, 5),
+  (2, 10, 10), (2, 11, 2), (2, 12, 10), (2, 13, 5), (2, 14, 6), (2, 15, 4), (2, 16, 8),
+  (3, 10, 250000), (3, 11, 50000), (3, 12, 200000), (3, 13, 300000), (3, 14, 500000), (3, 15, 700000), (3, 16, 400000); 
+
+INSERT INTO `cost_center_aggregate` (`period_id`, `cost_center_id`, `principal_center_id`, `debit`, `credit`) VALUES
```
jniles: I think this belongs in a test data script, not in a migration file.  Could you move it into it's own `sql` file?

server/models/schema.sql
```diff
@@ -2182,10 +2182,31 @@ CREATE TABLE `fee_center` (
...
+  `step_order` SMALLINT(5) NOT NULL DEFAULT 100,
+  `default_fee_center_index_id` mediumINT(8) UNSIGNED NULL,
   PRIMARY KEY (`id`),
   UNIQUE KEY `fee_center_1` (`label`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
+DROP TABLE IF EXISTS `fee_center_index`;
+CREATE TABLE `fee_center_index` (
+  `id` mediumINT(8) UNSIGNED NOT NULL AUTO_INCREMENT,
+  `label` VARCHAR(100) NOT NULL,
```
jniles: Let's make this label longer than 100 - it will bite us one day for sure.

server/controllers/finance/reports/fee_center_step_down/index.js
```diff
@@ -0,0 +1,118 @@
...
+const TEMPLATE = './server/controllers/finance/reports/fee_center_step_down/report.handlebars';
+
+// expose to the API
+exports.report = document;
+
+/**
+ * @function reporting
+ *
+ * @description
+ * Renders the Compte d'Exploitation
```
jniles: I don't think this renders the compte d'explotation 😄

server/models/schema.sql
```diff
@@ -2182,10 +2182,31 @@ CREATE TABLE `fee_center` (
   `label` VARCHAR(100) NOT NULL,
   `is_principal` TINYINT(1) UNSIGNED DEFAULT 0,
   `project_id` SMALLINT(5) UNSIGNED NULL,
+  `step_order` SMALLINT(5) NOT NULL DEFAULT 100,
+  `default_fee_center_index_id` mediumINT(8) UNSIGNED NULL,
```
jniles: I assume this is the default ordering?

client/src/modules/reports/generate/fee_center_step_down/fee_center_step_down.config.js
```diff
@@ -0,0 +1,85 @@
...
+  vm.onSelectPeriodTo = function onSelectPeriodTo(period) {
+    vm.reportDetails.periodTo = period.id;
+  };
+
+  vm.clearPreview = function clearPreview() {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.onSelectCronReport = report => {
```
jniles: I don't think you ever use this.

client/src/modules/reports/generate/fee_center_step_down/fee_center_step_down.config.js
```diff
@@ -0,0 +1,85 @@
...
+
+  vm.onSelectFiscalYear = (fiscalYear) => {
+    vm.reportDetails.fiscal_id = fiscalYear.id;
+  };
+
+  vm.onSelectPeriodFrom = function onSelectPeriodFrom(period) {
+    vm.reportDetails.periodFrom = period.id;
+  };
+
+  vm.onSelectCurrency = (currency) => {
```
jniles: I don't think you use this.

jniles (commented):
@mbayopanda I've taken a look.  Looks pretty good!  I've made some code clean up comments.

I recommend that you put the percentages on the right hand side instead of the left hand side.  I was really confused for a while about how costs were being allocated.

I propose that we include the "units" of the allocation key.  So, if it is meters, we can show "meters" on the report..  If it is days, we can show "days" on the report.

Finally, I think the end result we are looking for is something like this:

image

Would it be possible to include the direct costs in a column to the left like in the above picture?  Could you also mark the principal cost centers in **bold** or something similar so that we know that they should be the only ones with unallocated costs?

server/models/migrations/next/migrate.sql
```diff
@@ -41,3 +41,60 @@ CALL add_column_if_missing('general_ledger', 'principal_center_id', 'MEDIUMINT(8
...
+  `fee_center_index_id` MEDIUMINT(8) UNSIGNED NOT NULL,
+  `fee_center_id` MEDIUMINT(8) UNSIGNED NOT NULL,
+  `value` decimal(19,4) UNSIGNED DEFAULT NULL,
+  KEY `fee_center_index_id` (`fee_center_index_id`),
+  KEY `fee_center_id` (`fee_center_id`),
+  CONSTRAINT `fc_index_value__fee_center_index_id`FOREIGN KEY (`fee_center_index_id`) REFERENCES `fee_center_index` (`id`),
+  CONSTRAINT `fc_index__fee_center`FOREIGN KEY (`fee_center_id`) REFERENCES `fee_center` (`id`)
+) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
+
+INSERT INTO `fee_center_index` (`id`, `label`) VALUES
```
jmcameron: Please add INSERT IGNORE into all of these 'INSERTS' to make the script more robust.

jmcameron (commented):
It was great seeing this table appear with this realistic data!  Good job!

I had trouble running this with bhima_test.  When I did not run the migration file, I got this error:

    errors   sqlMessage: "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'NULL' at line 1",
    errors   sqlState: '42000',
    errors   index: 0,
    errors   sql: "CALL FeeCenterCostWithIndexes('2021-01-01 00:00:00.000', '2021-12-31 00:00:00.000');"

So I tried running the migration file.  It failed because the final line tried to insert unit 298 again (which was already added in bhima.sql).  When I added 'INSERT IGNORE' on those lines, I was able to load the migration file and it worked.

Comments regarding the tabular display:
image
- I do not think the first % column is necessary, so delete it
- Reduce the number of digits in the percents to one after the point
- Replace the first % column with the direct cost.  I realized it appears in the next table but it is useful to track what is happening on each line.
- I'm not sure if this is possible, but I would like to see the % and amount column headers grouped together for each center.  I'll include a figure below suggesting what I have in mind.  Part of this could be done by using grid column cellClass and colSpan attributes and new CSS rules.  Note that we could do this formatting thing later. 
image

server/models/migrations/next/migrate.sql
mbayopanda: The `index` (distribution key) to use for this `fee_center` during the allocation operation

server/models/schema.sql
mbayopanda: Yes, it is supposed to be defined by the `step_order` column

mbayopanda:
image

Here is latest report

mbayopanda:
@jmcameron changes landed

mbayopanda:
@jmcameron changes landed

image

server/controllers/finance/reports/fee_center_step_down/report.handlebars
```diff
@@ -0,0 +1,102 @@
...
+                <th>{{ value }}</th>
+              {{/each}}
+            </tr>
+          </thead>
+          <tbody>
+            {{#each feeCenterIndexes as | fcIndex |}}
+              <tr>
+                <td>{{translate fcIndex.index }}</td>
+                {{#each fcIndex.distribution as | dist |}}
+                  <td>{{ dist.value }}</td>
```
jniles: Can we put `class="text-right"` on these?  It would make it easier to read!

mbayopanda:
@jmcameron changes landed

server/models/migrations/next/migrate.sql
jmcameron: To avoid having to deal with when/where to inject the SQL data, I suggested to Bruce that he include it as mock data in the step-down report function so he can easily inject the data in the DB, generate the report, and then purge the data.  That simplifies processes and means the regression tests will not be affected.

jmcameron (approved):
I approve after the name fee_center->name field is fixed in schema.sql.
Looks great!

mbayopanda:
> 
> 
> I approve after the name fee_center->name field is fixed in schema.sql.
> Looks great!

@jmcameron changes landed

PR #5906, opened 2021-09-06
Add constraint if missing

### LINKED ISSUE #5893: Add a "add_constraint_if_missing()" stored procedure to easy migrations
As mentioned in https://github.com/IMA-WorldHealth/bhima/pull/5885#pullrequestreview-742435153, it would be useful to add a new SQL procedure to our migration processes to add constraints if missing and skip that step if it exists.  This would allow use to write idempotent migration scripts.

A template is already available for us in the `drop_column_if_exists` function, which as a constraint check.

mbayopanda:
This PR is related to #5893 

Close #5893


jniles (approved):

lomamech (approved):
LGTM

PR #5907, opened 2021-09-06
feat: add users/stock movements to sys usage

### LINKED ISSUE #5848: Improvements to the System Usage Statistics Report
Inspired by https://github.com/IMA-WorldHealth/bhima/pull/5846, we could make the following improvements to the System Usage Statistics report:

1. Include a column number of stock movements
2. Include a column for the number of active users
3. Add totals at the foot of the report

image

jniles:
Adds active user counts and stock movements to the system usage report.

Closes #5848.

image

### DIFF
--- a/server/controllers/finance/reports/systemUsage/index.js
+++ b/server/controllers/finance/reports/systemUsage/index.js
@@ -19,22 +19,33 @@ async function document(req, res, next) {
     const { date } = req.query;
 
     const patientsNbr = `
-      SELECT HOUR(created_at) as 'hour', COUNT(patient.uuid) as 'number'
+      SELECT HOUR(created_at) as 'hour', COUNT(patient.uuid) as 'number',
+        GROUP_CONCAT(DISTINCT user_id ORDER BY user_id) AS users
       FROM patient
       WHERE DATE(created_at) = DATE(?)
       GROUP BY HOUR(created_at)
     `;
 
     const invoicesNbr = `
-      SELECT HOUR(created_at)  as 'hour', COUNT(invoice.uuid) as 'number'
+      SELECT HOUR(created_at) as 'hour', COUNT(invoice.uuid) as 'number',
+        GROUP_CONCAT(DISTINCT user_id ORDER BY user_id) AS users
       FROM invoice
       WHERE DATE(created_at) = DATE(?)
       GROUP BY HOUR(created_at)
     `;
 
     const cashNbr = `
-      SELECT HOUR(created_at)  as 'hour', COUNT(cash.uuid) as 'number'
+      SELECT HOUR(created_at) as 'hour', COUNT(cash.uuid) as 'number',
+        GROUP_CONCAT(DISTINCT user_id ORDER BY user_id) AS users
       FROM cash
+      WHERE DATE(created_at) = DATE(?)
+      GROUP BY HOUR(created_at)
+    `;
+
+    const stockMovementNbr = `
+      SELECT HOUR(created_at) as 'hour', COUNT(DISTINCT document_uuid) as 'number',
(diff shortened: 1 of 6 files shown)


server/models/migrations/next/migrate.sql
```diff
@@ -36,7 +36,6 @@ CALL add_column_if_missing('posting_journal', 'principal_center_id', 'MEDIUMINT(
 ALTER TABLE `posting_journal` ADD CONSTRAINT `pg__cost_center_1` FOREIGN KEY (`cost_center_id`) REFERENCES `fee_center` (`id`) ON UPDATE CASCADE;
 ALTER TABLE `posting_journal` ADD CONSTRAINT `pg__cost_center_2` FOREIGN KEY (`principal_center_id`) REFERENCES `fee_center` (`id`) ON UPDATE CASCADE;
```
jmcameron: The new add_constraint_if_missing(sp?) is now merged into master.  I encourage you to merge with master and use it.

server/controllers/finance/reports/systemUsage/index.js
```diff
@@ -19,68 +19,117 @@ async function document(req, res, next) {
...
 
       if (cashPaymentMap[index]) {
-        row.cashPayments = cashPaymentMap[index][0].number;
+        const { number, users } = cashPaymentMap[index][0];
+        row.cashPayments = number;
+        aggregateUsers += users;
       }
 
+      if (stockMovementsMap[index]) {
+        const { number, users } = cashPaymentMap[index][0];
```
jmcameron: I think there is a bug here.  I picked a random day and this line crashed.  You may need to make sure it is not empty before dereferencing it:

      errors #interceptor() TypeError: Cannot read property '0' of undefined
      errors     at document (/home/jmcameron/Bhima/bin/server/controllers/finance/reports/systemUsage/index.js:99:56)
      errors     at processTicksAndRejections (internal/process/task_queues.js:97:5) +0ms

jmcameron (approved):
This looks good and pretty vanila.   Once the crash item is fixed, I think it is ready to go.

server/controllers/finance/reports/systemUsage/index.js
jniles: Yeah, a pretty serious bug at that, actually.  I'm referencing cash payments instead of stock movements 😞

jniles:
@jmcameron this is fixed now.

server/controllers/finance/reports/systemUsage/index.js
```diff
@@ -19,68 +19,117 @@ async function document(req, res, next) {
...
+      row.users = users;
+
       pivot.push(row);
     }
 
-    const cleaned = pivot.filter(row => (row.patients + row.invoices + row.cashPayments > 0));
+    const cleaned = pivot.filter(row => (row.patients + row.invoices + row.cashPayments + row.stockMovements > 0));
+
+    // aggregate all relevant
+    const aggregates = cleaned.reduce((agg, row) => {
```
jmcameron, line 119: I don't understand why you need to do pivot.filter.  Wouldn't the reduce work with the pivot array itself, since the aggregates would just add 0 for all the items in the row that are zero?   Is this an efficiency mod that may not be necessary with this small a number of items from the queries?

jmcameron (approved):
I tested this with the IMCK database and it works nicely.
I re-read the code and it looks good.

server/controllers/finance/reports/systemUsage/index.js
jniles: Filter removes rows that are all zero - it is just a way to clean up rows that are 0 and don't have any data in them.

It's for UX - we don't want to show a user a row with 0s.  We'll only show rows when people were working.

Issue #5879, opened 2021-08-24
Create allocation keys based real-world measures, not percentages


jniles:
Currently, BHIMA stores allocation keys based on a percentage that is calculated by the accountant.  Here is how we model them right now:

https://github.com/IMA-WorldHealth/bhima/blob/233f9353f56a3d99bfa3ee2734bed47f8c26678d/server/models/schema.sql#L2242-L2255

However, storing based on a percentage isn't really helpful - these percentages are derived from something, such as number of cases, number of square foot, etc.  You can imagine a layout like this:

Centres | Nombre Metres Carre | Frequentation | Heures de travaille
-- | -- | -- | --
Chirurgie | 250 | 1500 | 1600
Maternite | 300 | 2000 | 800
Maintenance | 10 | 0 | 300
Pediatrie | 250 | 1500 | 700
Electricte | 0 | 0 | 0
Med. Interne | 150 | 1700 | 1200
Pav. Med | 500 | 250 | 500
Tax | 0 | 0 | 0

In this layout, we can calculate the percentage contribution of each center as it relates to each column.  This allows us to change the number of square footages of Maternite, for example, without having to recompute _all_ percentages.  It also allows users to see exactly what kind of index is being used to calculate the cost allocations.

jniles:
Note that these allocations keys will come in three forms:

1. Static keys (e.g. square footage)
2. Dynamic keys that BHIMA can look up (.e.g number of employees, number of cases, number of patients)
3. Dynamic keys that must be set (e.g. number of kilograms of laundry)

The dynamic keys are discussed in more detail in https://github.com/IMA-WorldHealth/bhima/issues/5875.

jmcameron:
**A word on the allocation bases:**  These need to be flexible and able to be defined by the site administrator.  I could see that they operate similarly to tags, but would have some type information and a value associated with them.  One of the nice things about the step-down allocation algorithm currently being implemented, is that it uses a user-defined basis label so it is very flexible.  For instance it is not hard to imagine a situation where an auxiliary center provides services that can be quantified in two different ways.  For instance the Laundry service could be done by "kg of laundry processed for each service" or "number of pieces processed for each service".   We will need the flexibility to create these basis items on the fly since we cannot predict all the possible types of bases.

(Moved here from Issue https://github.com/IMA-WorldHealth/bhima/issues/5888.)

jmcameron:
> Note that these allocations keys will come in three forms:
> 
[...]

PR #5908, opened 2021-09-06
Support showing and editing cost center allocation method and basis

jmcameron:
This PR adds the ability to show the allocation method for each cost center and its allocation method.  

This PR partially addresses https://github.com/IMA-WorldHealth/bhima/issues/5899, This is a first step in implementing Issue #5879, and https://github.com/IMA-WorldHealth/bhima/issues/5879.   As it is finalized (or succeeding PRs finish the incomplete parts of this PR), some of those Issues will need to be closed later.

WARNING: This is a partial implementation:  Although you can edit the allocation method, I have not yet implemented editing the allocation basis; it is currently being done by SQL initialization data.   My goal in getting this out now, is so that the cost center step-down report can be updated to use this data before releasing it.

This PR adds the SQL tables 'cost_center_basis' and 'cost_center_basis_value' tables.  That later is not being used yet.  It also adds two members to the fee_center table:  'allocation_method' and 'allocation_method_id'.

NOTE: The goal of the new 'allocation_basis' field is to allow developers to define common allocation basis items (as is done in bhima.sql) but also support hand-created custom basis items.   The registry has a function that determines if it is a translation token and translates it; if not it does not change it.   I'm thinking that as we get this a bit more mature, we can predefine a number of basis items (in Bhima.sql) and reduce the need for creating the same things over and over.   Also one I thought of was "Number of services by center" (and add an optional field that identifies which center it is, perhaps with a units field.  We will not to work though things to decide what the final form should be.

**TESTING**
- Test it using bhima_test
- Show the Fee Center (Cost Center) registry page.  Note the new fields
- Try editing a center and changing its allocation method.


server/models/migrations/next/migrate.sql
```diff
@@ -41,3 +41,27 @@ CALL add_column_if_missing('general_ledger', 'principal_center_id', 'MEDIUMINT(8
...
 CALL add_constraint_if_missing('general_ledger', 'general_ledger__cost_center_1', 'FOREIGN KEY (`cost_center_id`) REFERENCES `fee_center` (`id`) ON UPDATE CASCADE');
 CALL add_constraint_if_missing('general_ledger', 'general_ledger__cost_center_2', 'FOREIGN KEY (`principal_center_id`) REFERENCES `fee_center` (`id`) ON UPDATE CASCADE');
+
+/**
+author: @jmcameron
+date: 2021-09-01
+description: adds allocation basis data to cost centers
+*/
+DROP TABLE IF EXISTS `cost_center_basis`;
+CREATE TABLE `cost_center_basis` (
```
mbayopanda: Removing the `DROP TABLE` instruction, and use the `CREATE TABLE IF NOT EXISTS ` will prevent us from the error below when running the migration script.

image

mbayopanda (approved):
Names of new tables are as we discussed 👍🏽 

New columns are displayed successfully and the update of allocation method for a `fee_center` is working fine.

Adding a new `fee_center` returns a success message but nothing in the grid, if this point is not a blocking point I can approve this PR because it is not the final work (UI for populate the `cost_center_basis` and `cost_center_basis_value`) and others PR depends on this PR

server/models/migrations/next/migrate.sql
jmcameron: I'll push and update.

jmcameron:
> Adding a new `fee_center` returns a success message but nothing in the grid, if this point is not a blocking point I can approve this PR because it is not the final work (UI for populate the `cost_center_basis` and `cost_center_basis_value`) and others PR depends on this PR

I agree. Lets get this out there and fix some of these issues later.

PR #5913, opened 2021-09-13
feat: lookup cost center by service

jniles:
This commit adds supporting functions to look up the cost centers by the service UUID.  It also adds an integration test for this feature.

### DIFF
--- a/server/controllers/finance/accounts/index.js
+++ b/server/controllers/finance/accounts/index.js
@@ -65,6 +65,26 @@ function create(req, res, next) {
 }
 
 /**
+ * @function lookupCostCenterByAccountId
+ *
+ * @description
+ * Retrieves a cost center by account id.
+ */
+async function lookupCostCenterByAccountId(id) {
+  const { cc } = await db.one('SELECT GetCostCenterByAccountId(?) AS cc;', [id]);
+  return cc;
+}
+
+async function lookupCostCenter(req, res, next) {
+  try {
+    const id = await lookupCostCenterByAccountId(req.params.id);
+    res.status(200).json({ id });
+  } catch (e) {
+    next(e);
+  }
+}
+
+/**
  * @method update
  *
  * @description
@@ -88,7 +108,6 @@ function update(req, res, next) {
     .catch(next)
     .done();
 }
-
 
 /**
  * @method remove
@@ -121,7 +140,6 @@ function remove(req, res, next) {
     .catch(next)
     .done();
 }
-
 
 /**
  * @method list
@@ -325,7 +343,6 @@ function getOpeningBalanceForPeriod(req, res, next) {
     .done();
 }
 
-
 /**
  * @method processAccountDepth
  * @description get the depth of an account
@@ -398,4 +415,5 @@ exports.categories = categories;
 exports.references = references;
 exports.processAccountDepth = processAccountDepth;
 exports.importing = importing;
+exports.lookupCostCenter = lookupCostCenter;
 exports.getAnnualBalance = getAnnualBalance;
--- a/server/models/procedures/cost_centers.sql
+++ b/server/models/procedures/cost_centers.sql
@@ -1,0 +1,46 @@
+DELIMITER $$
+/*
+  This file contains code for linking cost centers to various activities throughout the app.
+*/
+
+
+/*
+@function GetCostCenterByAccountId()
+
+@description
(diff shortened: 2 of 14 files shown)


server/controllers/finance/accounts/index.js
```diff
@@ -398,4 +415,5 @@ exports.categories = categories;
 exports.references = references;
 exports.processAccountDepth = processAccountDepth;
 exports.importing = importing;
+exports.lookupCostCenter = lookupCostCenter;
```
jmcameron, line 418: It looks like you only exported one function here.  But you exported both for lookup by service.  Was there a reason for the difference?

server/models/procedures/cost_centers.sql
```diff
@@ -0,0 +1,46 @@
...
+
+@description
+Retrieves the cost center id for a service using its service_uuid.  Each service should have one
+and only one cost center assocaited with it.
+
+NOTE(@jniles) - Currently, BHIMA does not guarantee this, but when we do make guarantees,
+we will be able to modify this function to do it.
+*/
+CREATE FUNCTION GetCostCenterByServiceUuid(service_uuid BINARY(16))
+RETURNS MEDIUMINT(8) DETERMINISTIC
```
jmcameron, line 38: Why do we need a SQL function for this?  Wouldn't a normal server db query be fine for something this simple?  (like the ones in server/controllers/*/*.js)?

jmcameron (commented):
The code looks good.  I had a couple of questions but nothing that should block this PR. 
I tried it on my system and the unit tests and integration test all ran fine.

However, I suggest you wait until we get the Fee Center -> Cost Center rename PR released before doing final adjustments so it can be released.

server/models/procedures/cost_centers.sql
jniles: Two reasons:

1.  The invoicing and stock movements data calculation are doing in a stored procedure.  We've left JS at that point - it is all in SQL.  So, we need this to be available in SQL.

2. Making it a function instead of a straight query means we can change it easier when we rename the tables or change the way cost centers are linked to accounts/services.  For example, we may eventually link cost centers directly to accounts.

server/controllers/finance/accounts/index.js
jniles: No, not really.  I just didn't need one of them yet.  Eventually, I'll probably export both!

jniles:
@jmcameron this is ready for a second review.

server/models/procedures/cost_centers.sql
jmcameron: ok

jmcameron (changes requested):
I tested this on my computer.   All the tests passed except for the end-to-end tests.    Before the end-to-end tests would run, I had to make a couple of naming fixes (related to the Fee Center -> Cost Center renaming).  See the diffs file:
[pr5913.diffs.txt](https://github.com/IMA-WorldHealth/bhima/files/7184848/pr5913.diffs.txt)

When I ran the end-to-end tests, I saw this error:
```
  Distribution Auxiliary Cost Center
    1) set allocation by percentage

  34 passing (45s)
  1 pending
  1 failing

  1) Distribution Auxiliary Cost Center
       set allocation by percentage:
     NoSuchElementError: No element found using locator: By(css selector, *[id="allocation-center-grid"])
```
I presume this is related to the Fee Center -> Cost Center / Distribution -> Allocation renaming now merged into master.

jniles:
@jmcameron I think this happens for me on `master`.  My end to end tests haven't worked for some time.  I don't think it's worth fixing something we are going to be rewriting imminently, right?

jmcameron:
> @jmcameron I think this happens for me on `master`. My end to end tests haven't worked for some time. I don't think it's worth fixing something we are going to be rewriting imminently, right?

Ok.  I still suggest you go ahead and apply the diffs before releasing so that the end-to-end tests actually start (that way we fix these on master).  We can deal with actually fixing the end-to-end tests later when the dust has settled.

jmcameron:
> Ok. I still suggest you go ahead and apply the diffs before releasing so that the end-to-end tests actually start (that way we fix these on master). We can deal with actually fixing the end-to-end tests later when the dust has settled.

Or we could do it on a separate PR?

jniles:
@jmcameron changes landed.

PR #5917, opened 2021-09-15
Fix config allocation key

mbayopanda:
This PR fixes an issue on */allocation_center/allocation_key* which doesn't allow to edit allocation setting.


jmcameron (approved):
It works.  Thanks!

PR #5918, opened 2021-09-15
Fix lot usage schedule display problem

### LINKED ISSUE #5897: Stock Consumption Schedule rendering can sometimes get out of whack
Here is an example at Vanga:
image

There are two issues here:

1. The top table off the modal to the right.
2. The bottom calendar doesn't reflect that actual expiry date.

jmcameron:
Fixes occasional wrapping of lot usage schedule lines.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5897

This was tested on bhima_test and IMCK databases, but not Vanga.

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -1127,7 +1127,7 @@ td.center {
 ---------------------------------------------- */
 
 table.lot-schedule {
-  padding-left: 10px;
+  padding: 0 10px;
 }
 
 table.lot-schedule th.year {
@@ -1145,6 +1145,7 @@ table.lot-schedule th.month {
 
 table.lot-schedule th.reorder {
   background-color: #ffaaaa;
+  border-right: 0;
 }
 
 table.lot-schedule tr.lot-schedule th {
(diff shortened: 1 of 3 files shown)


jmcameron:
@jniles : I think  I have this working.  I encourage you to try it again.

jniles (commented):
The wrapping problem appears fixed:

image

There is still an issue where the columns do not line up.  In the screenshot below, the box should end in June but it ends in July:

image


We can merge this PR since the first part is fixed, alternatively, we can work on both bugs before merging.  Which would you prefer?

jmcameron:
> There is still an issue where the columns do not line up. In the screenshot below, the box should end in June but it ends in July:
> 
> image
> 
@jniles : We may need to pair program on this one.  It works correctly on my system:
image

Notice that the width of the months are narrower on your system than mine.  But I think that the lot schedule bar (green/red) is the same on both systems.  So my guess is there is some CSS tweak that we need to make the months work the same on both systems.

I'm open to releasing now and fixing later -- or waiting until we get a chance to figure this out.  Your call.

BTW, what browser were you using?  Anything odd about your system (display-wise)?   I'm running Chrome 93.0.4577.82.

jniles:
@jmcameron I'm running FF92 on the same machine as you.

I think this is going to be a tricky fix of `px` in CSS, probably with some hacks.  Let's release this and work on the rendering in a different PR.

bors r+

PR #5919, opened 2021-09-15
Fixed problem with migrate.sql file

jmcameron:
Fixed an oversight in the migrate.sql file.

Tested on IMCK database and it works but takes a long time to complete.


lomamech (approved):
I have just done the execution with the VANGA database, and the execution was successful
image

PR #5920, opened 2021-09-16
Fix currency in stock sheet report

### LINKED ISSUE #5898: Stock Sheet currency rendering doesn't work correctly anymore
Here is how the current `master` currency rendering looks for the Stock Sheet.  I'm trying to render it in FC, but it seems to choose $USD for some of the columns,

image

jmcameron:
Closes https://github.com/IMA-WorldHealth/bhima/issues/5898

**TESTING**
- Check doing Stock > Reports > Stock Sheet reports
- Try selecting all three currencies and make sure that the currencies are correct in all columns and the totals at the end


jniles (approved):
Tested and working.  Thanks!

bors r+

Issue #5888, opened 2021-08-29
Implement a grid display of cost center properties / registry


jmcameron:
Add a grid table for the cost centers that shows the status of each cost center.   This could be done by extending the current "Fee Center" registry or adding a new option (Cost Center Allocations).  This would look similar to the "Table 13-3 Cost Allocation Example" from the book:
![image](https://user-images.githubusercontent.com/62145/131244350-78c3d138-574d-4e72-a696-839e8e8f3987.png)

The table should include:
- Direct costs (looked up or computed by the centers model)
- Whether it is a auxiliary/indirect/overhead center or a principal/revenue center
- The value of the supported allocation bases for the center (number of employees, area, etc).
- The action menu should allow editing the properties of the cost center 

EDIT: Moved comments about allocation bases to Issue https://github.com/IMA-WorldHealth/bhima/issues/5879.

jmcameron:
The second table of the step-down analysis report is similar to what we want:
![image](https://user-images.githubusercontent.com/62145/132689816-29db253f-8b88-4ad6-9562-e6648144fc1e.png)

We need to think about what to list across the top (centers or properties).  I see that the book example is on way and the step-down report is the other.    It seems to me more likely to be more centers than the different types of quantities, so I would put the properties across the top and the centers on rows (like the Table 13-3 example).  That way the 'action' menu corresponds to the row (center), not the property!

jmcameron:
Closed in PR https://github.com/IMA-WorldHealth/bhima/pull/5924

PR #5924, opened 2021-09-17
Add registries for allocation keys and allocations to cost_center

mbayopanda:
This PR is related to #5888 

This PR adds two registries : 
- Allocation keys registry (DONE)
- Allocation registry (IN PROGRESS)

In this PR we clean up the `server\controllers\finance\reports\cost_center_step_down\index.js` file, and also expose to the client `cost_center_allocation_keys` and `cost_center_allocation_registry` data.

image

How to test:
- After `yarn build:db`, you can run the file `server\models\migrations\next\sample_cost_center_data.sql` in mysql
- Look at the report `Centre des couts -> Rapports -> Allocation des centres de couts`
- Look at the registry `Centre des couts -> Clé de répartition`
- Look at the registry `Centre des couts -> Registre des allocations`

This PR is IN PROGRESS


client/src/i18n/en/cost_center.json
```diff
@@ -0,0 +1,7 @@
+{
```
jmcameron, line 1: Good idea to put these into their own file!

client/src/modules/cost_center/allocation_keys/allocation_keys.js
```diff
@@ -0,0 +1,73 @@
...
+        }];
+
+        rows.costCenterList.forEach((item, idx) => {
+          if (idx >= 0) {
+            const column = {
+              field : `idx${idx}`,
+              displayName : item,
+              headerCellFilter : 'translate',
+            };
+            columnDefs.push(column);
```
jmcameron: Nice!

client/src/modules/cost_center/cost_center.service.js
```diff
@@ -27,5 +27,17 @@ function CostCenterService(Api, $uibModal) {
     });
   };
 
+  service.getAllocationKeys = () => {
```
jmcameron, line 30: Is this different from the AllocationBasisService->getAllocationBases()?  If not, then I suggest using the one in AllocationBasisService and not having the same thing it two places.

client/src/modules/fee_center_allocation/fee_center_allocation.service.js
```diff
@@ -0,0 +1,30 @@
...
+FeeCenterAllocationService.$inject = ['PrototypeApiService'];
+
+/**
+ * @class FeeCenterAllocationService
+ * @extends PrototypeApiService
+ *
+ * @description
+ * Encapsulates common requests to the /fee_center/ URL.
+ */
+function FeeCenterAllocationService(Api) {
```
jmcameron: Please rename FeeCenter* to CostCenter*

jmcameron (commented):
I'm getting confused between the old direct distribution functions and the new cost allocation basis functions because we renamed the 'distribution' items to 'allocation' items.    I suggest we revert the names of the old direct distribution functions back to 'distribution' to make things easier to understand.   Sorry, I did not notice this in the big renaming PR!

We are also confusing the terms "keys" and "basis items".  The old approach uses "Keys" but I think the new cost allocation functions/pages should use the term "basis item" terminology.  So I suggest changing the new "allocation_keys.js" to "allocation_bases.js (and similar renames for the html file and titles within it.   

Reminder: in English, the singular is "basis" but the plural is "bases".  I think the appropriate terms in French are "base" and "bases" (respectively).

Great start of the Allocation Bases table!  Per our discussion, I hope you will be able to add the ability to edit the amounts directly from the grid.
image

Please rotate the Allocation Keys table 90 degrees - Put the allocation basis items in the columns and cost centers in the rows.   Note that this is the way that the allocation data is listed in the table at the bottom of the Step Down report page. There are two reasons for this:  (1) For realistic cases, we are likely to have more cost centers than distinct allocation basis items, so putting the cost centers in rows will make the form less wide and more manageable for the users, and (2) this will allow us to put the total for each type of allocation basis item as a grid column total at the bottom.  This would be useful for general information and as cross-check on data entry. (To prevent us from claiming for of any basis item than we actually have across the organization.)

The English name and menu entry for "Cost Center Allocations" is okay, but the English title on the page (once opened) is "Allocation of Cost Centers".  I think we could improve that.  I think it should be something more descriptive like:  "Cost Allocation to Revenue Centers"

Please rename all "Fee Center" items to "Cost Center" items (including fixing filenames).

Based on your testing suggestions, I imported the SQL file.  The "Clés de répartition" page worked, the "Centre de Frais" page worked, and the "Allocations des centres de coût" report page worked.  But when I tried the "Registre des allocations" page, it was empty:
image

jmcameron (commented):
Great step forward!  Thanks for taking my suggestions.
It is looking good!

client/src/modules/cost_center/allocation_bases/allocation_bases.js
```diff
@@ -0,0 +1,107 @@
...
+function CostCenterAllocationBasesController(
+  CostCenters,
+  uiGridConstants,
+  Notify,
+) {
+  const vm = this;
+
+  // global variables
+  vm.gridApi = {};
+  vm.filterEnabled = false;
```
jniles: I don't think you use this line.  Could you either put an online filter on the grid, or remove it?

client/src/modules/cost_center/allocation_bases/allocation_bases.js
```diff
@@ -0,0 +1,107 @@
...
+          field : 'label',
+          displayName : 'COST_CENTER.TITLE',
+          headerCellFilter : 'translate',
+          cellFilter : 'translate',
+          aggregationHideLabel : true,
+          aggregationType : uiGridConstants.aggregationTypes.count,
+        }];
+
+        rows.costCenterIndexes.forEach((item, idx) => {
+          if (idx >= 0) {
```
jniles: If I am not mistaken, the index in forEach Will always be positive.  Why do you need this check?

client/src/modules/cost_center/allocation_bases/allocation_bases.js
```diff
@@ -0,0 +1,107 @@
...
+              type : 'number',
+              headerCellFilter : 'translate',
+              cellClass : 'text-right',
+              footerCellClass : 'text-right',
+              cellFilter : 'number: 2',
+              footerCellFilter : 'number:2',
+              aggregationHideLabel : true,
+              aggregationType : uiGridConstants.aggregationTypes.sum,
+            };
+            columnDefs.push(column);
```
jniles: It looks like you are doing an array.forEach() and an array.push() inside it.  Could you just do an array.map()?

client/src/modules/cost_center/allocation_bases/allocation_bases.js
```diff
@@ -0,0 +1,107 @@
...
+    enableSorting     : false,
+    onRegisterApi     : onRegisterApiFn,
+  };
+
+  function onRegisterApiFn(gridApi) {
+    vm.gridApi = gridApi;
+  }
+
+  function load() {
+    CostCenters.getAllocationKeys()
```
jniles, line 56: Would it be possible to show a loading indicator while you are doing this?

client/src/modules/cost_center/allocation_bases/allocation_bases.js
```diff
@@ -0,0 +1,107 @@
...
+            value[`idx${i}`] = el.value;
+            line[i].push(value);
+          }
+        }
+
+        const data = [];
+        for (let i = 0; i < rows.costCenterList.length; i++) {
+          let value = {};
+          for (let j = 0; j < line.length; j++) {
+            const element2 = line[j];
```
jniles: What is `element2`?  Can we give it a more meaningful name?

client/src/modules/cost_center/allocation_bases/allocation_bases.js
```diff
@@ -0,0 +1,107 @@
...
+            columnDefs.push(column);
+          }
+        });
+
+        const line = [];
+        for (let i = 0; i < rows.costCenterIndexes.length; i++) {
+          const base = rows.costCenterIndexes[i];
+          line[i] = [];
+          for (let j = 0; j < base.distribution.length; j++) {
+            const el = base.distribution[j];
```
jniles: What is `el`?  Can we give it a more meaningful name?

client/src/modules/cost_center/allocation_registry/allocation_registry.js
```diff
@@ -0,0 +1,45 @@
...
+    enableSorting     : false,
+    onRegisterApi     : onRegisterApiFn,
+  };
+
+  function onRegisterApiFn(gridApi) {
+    vm.gridApi = gridApi;
+  }
+
+  function load() {
+    CostCenters.getAllocationRegistry({ test : 'hello world' })
```
jniles: Hello world! :)

server/controllers/finance/cost_center_allocation_bases.js
```diff
@@ -0,0 +1,46 @@
+/**
+* Cost Center Allocation Keys
+*
+* This controller exposes an API to the client for reading allocation keys
+*/
+const _ = require('lodash');
+const db = require('../../lib/db');
+
+async function fetch() {
+  const queryCostCenterIndexesList = `
```
jniles, line 10: This is a great variable name!

server/controllers/finance/cost_center_allocation_bases.js
```diff
@@ -0,0 +1,46 @@
...
+    FROM cost_center fc 
+    JOIN cost_center_allocation_basis_value ccbv ON ccbv.cost_center_id = fc.id 
+    JOIN cost_center_allocation_basis ccb ON ccb.id = ccbv.basis_id
+    ORDER BY fc.step_order ASC;
+  `;
+  const costCenterIndexesList = await db.exec(queryCostCenterIndexesList);
+  const indexes = _.groupBy(costCenterIndexesList, 'cost_center_allocation_basis_label');
+  const costCenterList = [];
+  const costCenterIndexes = _.keys(indexes).map((index, i) => {
+    const fcIndex = _.sortBy(indexes[index], 'step_order');
```
jniles, line 25: Are they not already ordered by step order?  I thought the SQL would take care of that.

server/controllers/finance/cost_center_allocation_registry.js
```diff
@@ -0,0 +1,116 @@
+/**
+* Cost Center Allocation Keys
+*
+* This controller exposes an API to the client for reading allocation keys
+*/
+// const _ = require('lodash');
+const db = require('../../lib/db');
+const Stepdown = require('../../lib/stepdown');
```
jniles: Normally, capital letters indicate a class name that you will invoke with `new`.  Would it be possible to import this as just `stepdown`?

server/controllers/finance/cost_center_allocation_registry.js
```diff
@@ -0,0 +1,116 @@
...
+const fiscal = require('./fiscal');
+const ccAllocationKeys = require('./cost_center_allocation_bases');
+
+async function fetch(session, params) {
+  const defaultCurrentYear = new Date();
+  const enterpriseCurrencyId = session.enterprise.currency_id;
+
+  const periods = {
```
jniles, line 15: I don't think we should make these default.  If a programmer makes a typo, they will be very confused why their dates are not working.

Can we just require the user to submit a period?  And throw an error is no period is specified?

server/controllers/finance/cost_center_allocation_registry.js
```diff
@@ -0,0 +1,116 @@
...
+    periodFrom : params.periodFrom || `${defaultCurrentYear.getFullYear()}01`,
+    periodTo : params.periodTo || `${defaultCurrentYear.getFullYear()}12`,
+  };
+
+  const range = await fiscal.getDateRangeFromPeriods(periods);
+  const query = 'CALL ComputeCostCenterAllocationByIndex(?, ?);';
+  let [feeCenters] = await db.exec(query, [range.dateFrom, range.dateTo]);
+
+  if (feeCenters.length) {
```
jniles: I am confused by this.  What is it doing?  It looks like if a single result is returned, that means an weird happened?  Can we just throw the error and used a try/catch?

server/controllers/finance/reports/cost_center_step_down/report.handlebars
```diff
@@ -79,13 +79,13 @@
...
               <th>{{translate 'COST_CENTER.DISTRIBUTION_KEY'}}</th>
-              {{#each feeCenterList as | value |}}
+              {{#each costCenterList as | value |}}
                 <th>{{ value }}</th>
               {{/each}}
             </tr>
           </thead>
           <tbody>
-            {{#each feeCenterIndexes as | fcIndex |}}
+            {{#each costCenterIndexes as | fcIndex |}}
```
jniles, line 88: :+1:

jniles (commented):
Just a few comments from me to improve the code.  I only read the code, I haven't executed it as I am on a phone.  Apologies if something was obvious that I missed!

mbayopanda:
Thank you @jmcameron and @jniles for the review

client/src/modules/cost_center/cost_center.service.js
mbayopanda: They are different according returned data

server/controllers/finance/cost_center_allocation_bases.js
mbayopanda: In the SQL procedure it was not obvious to sort these data.

server/controllers/finance/cost_center_allocation_registry.js
mbayopanda: It was a solution for the issue that we have got with `ComputeCostCenterAllocationByIndex` procedure, which thrown an error when there is not data to process for building the select query.

For the user, when he try to display the report when there is not data, the server returns an error message instead of just showing nothing as data in the report. So, to fix that I did these kind of checks (any optimal solution are welcome)

jmcameron (commented):
It looks good and it works.   

We can do more work later to improve the number formatting and consolidating the usage of language tokens (such as ALLOCATION_BASIS_AREA_USED) so they are not duplicated in two places, etc.  But I would like to get this out as  soon as possible so we can move on to other necessary functionality for this task.

I think after fixing the integration test, we should go ahead and release it.   Once that is done, I'll put together a few simple issues to address some of the cosmetic things I mentioned.

server/controllers/finance/cost_center_allocation_registry.js
jniles: Why isn't this a try/catch?  It doesn't look like an error is being thrown anywhere.

mbayopanda:
@jmcameron @jniles the try_catch for handling procedure result doesn't produce a good user experience, since when there is no data (the procedure throw an error) caught by NodeJS and we cannot see the report page (whenever there is no data) the behavior is like a bug while the thing is just we do not have data.

jniles:
Looks like bors froze.

bors r-

jmcameron:
> bors r+

Thanks for the assist!

jmcameron:
@mbayopanda Great job getting this done!  Thanks!

PR #5925, opened 2021-09-20
Revert some renames for the direct distributions

jmcameron:
Our recent fee center -> cost center names went a bit too far, in my opinion.  As @mbayopanda is developing the grid display of cost centers and their allocation basis values, I realized that the names for the pages were confusing.  Here what I would like us to do:

(1) Rename the menu functions for the older direct distribution pages back to using "Distribution" instead of "Allocation".  That will avoid confusion between the two approaches.  We an decide what do with the two approaches later (perhaps sub-menus for each).  

(2) For the older direct distribution pages and related functions, keep using the term "key"  (distribution keys), etc.  However for the new Cost Allocation approach (step-down), use the term "Allocation basis" and "Allocation Bases".   That will make the distinction between the two approaches clearer and avoid confusion.

This PR addresses part of item (1) above.

**TESTING**
- Use bhima_test
- Since menu items are changing, be sure to do 'yarn; yarn build; yarn build:db" and clear your browser before testing
- When you get it running look in the "Cost Center Management" menu and verify that the older direct distribution pages use the "Distribution" term and the titles on the corresponding pages use similar terms.


jniles:
Are we going to keep around the old code?  My understanding was that we were just going to remove it entirely, since it doesn't do step down.  What is the purpose in preserving it?

lomamech (commented):
@jmcameron  I think that if we decide to keep * Center de frais * for the French translation, I think that our labels being in English, we should keep COST CENTER instead of FEE CENTER

jmcameron:
I've had second thoughts about this.   Perhaps is is time to remove the old direct distribution functions entirely?  @lomamech what do you think?  I know you and ? (Dedrick?) worked hard on it.   We could also leave it in but disable it, but that would still leave the code in a confusing state between the old and the new code.   (This is especially true since we the Fee Center -> Cost Center renaming process added to the confusion).

jniles:
I think that we just need to commit to the current course and rewrite.  I can't see us needing these changes.

PR #5928, opened 2021-09-22
chore: remove TravisCI config

jniles:
Removes the Travis CI from the repository since we no longer use it.


jmcameron (commented):
Looks good to me.   I assume that we have no other references to the deleted .travis.yml file in our scripts, etc.?

jniles:
We do not.

PR #5929, opened 2021-09-22
feat(invoices): automatically allocate costs

### LINKED ISSUE #5855: Ensure automatic income distribution works with Patient Invoices
To push through Cost Center Analysis, we need to ensure that the profit centers are being automatically assigned their profits during the invoicing event.  This means the following things need to be ensured:

1. When an invoice is made, the associated profits are assigned to profit centers via some algorithm.
2. Once the invoice is made, it's impact is felt in the cost center analysis report.  This can either be before or after posting the data to the general ledger.
3. There is a straightforward way to see how canceled invoices are handled and/or impact the cost centers.

jniles:
This commit implements the automatic cost allocation during invoicing. The logic is that any income/expense account that is associated with a cost center is assigned that cost center when writing to the posting_journal.  If the account is not assigned, we default to the service's cost center.

This applies to all invoice_items, invoicing_fees, and invoice_subsidies.

Note: I haven't implemented the calculation of the cost_center_aggregate because @lomamech did that in his PR.

Closes #5855.

Here is an example of the test dataset with cost centers allocated:
image

-----

@mbayopanda @lomamech @jmcameron this shows how we can automatically assign cost centers to transactions made during invoicing.  The same kind of logic will apply to stock exits and payroll.  Please review this code and let me know what your thoughts are.

I haven't used the `principal_center_id` here, and I'm less sure that we need it.  I think directly assigning the income/costs may be enough without the `principal_center_id`.

### DIFF
--- a/client/src/modules/journal/journal.js
+++ b/client/src/modules/journal/journal.js
@@ -318,6 +318,16 @@ function JournalController(
     headerCellFilter : 'translate',
     visible : false,
   }, {
+    field : 'costCenterLabel',
+    displayName : 'FORM.LABELS.COST_CENTER',
+    headerCellFilter : 'translate',
+    visible : false,
+  }, {
+    field : 'principalCenterLabel',
+    displayName : 'FORM.LABELS.PRINCIPAL_COST_CENTER',
+    headerCellFilter : 'translate',
+    visible : false,
+  }, {
     field : 'comment',
     displayName : 'FORM.LABELS.COMMENT',
     headerCellFilter : 'translate',
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -666,6 +666,7 @@
             "PRICE": "Price",
             "PRINCIPAL_2": "Principal",
             "PRINCIPAL": "Principal",
+            "PRINCIPAL_COST_CENTER" : "Principal Cost Center",
             "PRINT_ORIENTATION_LANDSCAPE":"Landscape",
             "PRINT_ORIENTATION_PORTRAIT":"Portrait",
             "PRINT_ORIENTATION":"Print Orientation",
--- a/client/src/i18n/fr/form.json
+++ b/client/src/i18n/fr/form.json
@@ -666,6 +666,7 @@
             "PRICE": "Prix",
             "PRINCIPAL_2": "Principal",
             "PRINCIPAL": "Principale",
+            "PRINCIPAL_COST_CENTER" : "Principale centre de coûts",
             "PRINT": "Imprimer",
             "PRINT_ORIENTATION_LANDSCAPE":"Paysage",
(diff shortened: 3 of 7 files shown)


client/src/modules/journal/journal.js
```diff
@@ -317,6 +317,16 @@ function JournalController(
     cellTemplate : exchangeRateCellTemplate,
     headerCellFilter : 'translate',
     visible : false,
+  }, {
+    field : 'costCenterLabel',
+    displayName : 'FORM.LABELS.COST_CENTER',
+    headerCellFilter : 'translate',
+    visible : false,
```
jmcameron, line 324: Why not display this?

jmcameron (commented):
It looks good to me.  But my understanding of what is going on here is limited, so take that with a grain of salt.   

I read through invoicing.sql and learned some SQL!   The comments are helpful, but a walk-through with someone who understands the code would be very helpful.   The changes introduced in this PR seem appropriate.   

in cost_center.sql, I would like to see a clearer description of what GetCostCenterByAccountId() GetCostCenterByServiceUuid() return if the account/service does not have a cost center.  I presume these will return NULL in that case, but it would be good to be explicit.  And when the accounts/services are updated, perhaps it should throw and error in such cases?

BTW, in the reference_cost_center table, I see:
      `account_reference_id` MEDIUMINT(8) UNSIGNED NOT NULL,
Is this a large enough integer?  Could we have more than 256 accounts?

Finally:  my understanding is these changes will apply to invoicing after this change is implemented.  How will past invoices be fixed?

client/src/modules/journal/journal.js
jniles: No reason.  This just sets the default behavior.  We can switch it if we so desire.  Presumably, you won't be editing cost centers commonly in the posting_journal which is why I default to hidden.

jniles:

I've updated the procedures with comments to reflect that `NULL` can be returned.

For your mediumint question, the size is much larger: https://dev.mysql.com/doc/refman/8.0/en/integer-types.html  The max value is 16777215.

As for past invoices ... that is somewhat tricky.  I think we can do some hackery in the databases of institutions, but their data is already posted and we really shouldn't change it.  We'll need to get in touch with each institution and see what their preference is.

mbayopanda (approved):
It looks LGTM 👍🏽 

The UI feature works fine 

image

As you mentioned that, you're not writing in the cost_center_aggregate yet

This PR can be merged

jmcameron:
> For your mediumint question, the size is much larger: https://dev.mysql.com/doc/refman/8.0/en/integer-types.html The max value is 16777215.

Ah... I did not realize that 8 referred to number of bytes.  I was thinking 8 bits.   So it is plenty large!

jmcameron:
I was able to run this and verify that it seems to work with bhima_test.

PR #5930, opened 2021-09-22
feat(stock): add sale value on stock value report

### LINKED ISSUE #5847: Stock Value Report: Add Selling Price Value
This issue is about to: 

- Add a column for value based on inventory selling price (price * quantity)
- Add sum in the footer of this new column
- Fix colspan on the bottom of stock values table

In final, we could have these :
- Code,
- Inventory Text,
- Quantity
- Inventory Price,
- Value for the price
- Stock Unit Cost
- Value for the stock unit cost

jniles:
Adds the stock's sale value on the stock value report.  It also displays a red warning if the sale value is less than the stock's value to alert users of potential mistakes.

Closes #5847.

Here is an example:
image

### DIFF
--- a/server/controllers/stock/reports/stock/value.js
+++ b/server/controllers/stock/reports/stock/value.js
@@ -74,6 +74,7 @@ async function reporting(_options, session) {
   });
 
   let stockTotalValue = 0;
+  let stockTotalSaleValue = 0;
   const exchangeRate = await Exchange.getExchangeRate(enterpriseId, options.currency_id, new Date());
   const rate = exchangeRate.rate || 1;
 
@@ -97,12 +98,19 @@ async function reporting(_options, session) {
       item.weightedAverageUnitCost = weightedAverageUnitCost;
     });
 
+    stock.inventoryPrice = stock.inventory_price * rate;
+
     stock.stockQtyPurchased = quantityInStock;
     stock.stockUnitCost = weightedAverageUnitCost;
     stock.stockValue = (quantityInStock * weightedAverageUnitCost);
-    stockTotalValue += stock.stockValue;
+    stock.saleValue = (quantityInStock * stock.inventoryPrice);
 
-    stock.inventoryPrice = stock.inventory_price * rate;
+    // warn the user if the stock sales price is equivalent or _lower_ than the value of the stock
+    // the enterprise will not make money on this stock.
+    stock.hasWarning = stock.inventoryPrice <= weightedAverageUnitCost;
+
+    stockTotalValue += stock.stockValue;
+    stockTotalSaleValue += stock.saleValue;
   });
 
   const stockValueElements = options.exclude_zero_value
@@ -111,10 +119,12 @@ async function reporting(_options, session) {
   data.stockValues = stockValueElements || [];
 
   data.stockTotalValue = stockTotalValue;
+  data.stockTotalSaleValue = stockTotalSaleValue;
   data.emptyResult = data.stockValues.length === 0;
 
   data.currency_id = options.currency_id;
   data.exclude_zero_value = options.exclude_zero_value;
+
   return report.render(data);
 }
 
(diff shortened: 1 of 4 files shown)


jniles:
@mbayopanda please give me feedback on the French for "Sale Value".  I'm not sure the most appropriate term to use here.

server/controllers/stock/reports/stock/value.js
```diff
@@ -97,12 +98,19 @@ async function reporting(_options, session) {
...
     stock.stockQtyPurchased = quantityInStock;
     stock.stockUnitCost = weightedAverageUnitCost;
     stock.stockValue = (quantityInStock * weightedAverageUnitCost);
-    stockTotalValue += stock.stockValue;
+    stock.saleValue = (quantityInStock * stock.inventoryPrice);
 
-    stock.inventoryPrice = stock.inventory_price * rate;
+    // warn the user if the stock sales price is equivalent or _lower_ than the value of the stock
+    // the enterprise will not make money on this stock.
+    stock.hasWarning = stock.inventoryPrice <= weightedAverageUnitCost;
```
jmcameron, line 110: I think this should be: <
not: <=
since you should not have a warning if the sale price = value (right?)

server/controllers/stock/reports/stock_value.report.handlebars
```diff
@@ -29,28 +29,32 @@
...
-              <tr>
+            <tr {{#if hasWarning}}class="text-danger bg-danger"{{/if}} >
                 <td>{{inventory_code}}</td>
                 <td style="width:50%">{{inventory_name}}</td>
-                <td class="text-right">{{currency inventoryPrice ../currency_id 4}}</td>
                 <td class="text-right">{{stockQtyPurchased}}</td>
                 <td class="text-right">{{currency stockUnitCost ../currency_id 4}}</td>
                 <td class="text-right">{{currency stockValue ../currency_id}}</td>
+                <td class="text-right">{{currency inventoryPrice ../currency_id 4}}</td>
+                <td class="text-right">{{currency saleValue ../currency_id 4}}</td>
```
jmcameron: Probably only need 2 fractional digits for this.

jmcameron (approved):
Looks great!  Code is clean.  I made a couple of small cosmetic suggestions.

I tested it with the IMCK db and bhima_test.   The IMCK report looks great.   The Bhima_test report seemed a little odd, but I'm pretty sure it must be about the data, not this PR.  Note the big difference between the big difference between the cost and sale value (caused by big differences between the unit cost and the inventory price).   These seem out of line and the Bhima test data should probably be updated to more realistic values (LATER).
image

server/controllers/stock/reports/stock/value.js
jniles: It can be either - I'm open to changing it.  My thought is that the most common mistake users will make is setting the sale price = to the purchase price or vice versa, since they don't understand the difference between what they paid for the stock versus what they will sell it for.

Do you think we should limit it to `<`?

server/controllers/stock/reports/stock_value.report.handlebars
jniles: For all of them, or just the `saleValue` total?

server/controllers/stock/reports/stock/value.js
jmcameron: That makes sense.  Just leave it as it is.

server/controllers/stock/reports/stock_value.report.handlebars
jmcameron: Just the total sale value. The total value is already 2 digits.  The unit prices/cost should be 4 digits.

jniles: Okay!  Thanks for the tip.  Will make the change.

jniles (commented):
Changes landed.

jmcameron:
> Changes landed.

I think you only fixed the totals row.  It number of digits should also be fixed on every row (for total sale value).

jniles:
Ah, I misunderstood your comment.  I'll change them all.

jniles:
Now have I gotten it?

jmcameron (approved):
Looks good.

Issue #5856, opened 2021-08-16
Ensure automatic cost allocation happens during stock consumption


jniles:
To push through Cost Center Analysis, we need to ensure that the costs are being correctly allocated to their cost centers.  That is, it is possible to create a cost center and immediately start using it by exiting stock who's cost will be allocated to the cost center.

1. When a stock exit is made to a service, the cost center is automatically looked up and costs are allocated to the cost center.
2. When a stock exit is made to a patient who was billing in a cost center, the cost center is looked up from the invoice and automatically attributed to that cost center.
3. Stock losses are automatically attributed to some depot.

jniles:
Closed in #5931.

PR #5931, opened 2021-09-22
Auto allocation of cost center for stock exit to service and patient

mbayopanda:
This PR is related to #5856 

image

The case for stock loss is not implemented yet

### DIFF
--- a/server/models/procedures/cost_centers.sql
+++ b/server/models/procedures/cost_centers.sql
@@ -45,4 +45,23 @@ BEGIN
   );
 END $$
 
+/*
+@function GetCostCenterByInvoiceUuid(invoice_uuid)
+
+@params invoice_uuid BINARY(16) the invoice uuid
+
+@description
+Returns the cost center id for the invoice using the service_uuid in the invoice table
+*/
+CREATE FUNCTION GetCostCenterByInvoiceUuid(invoice_uuid BINARY(16))
+RETURNS MEDIUMINT(8) DETERMINISTIC
+BEGIN
+  RETURN (
+    SELECT GetCostCenterByServiceUuid(i.service_uuid)
+	FROM invoice i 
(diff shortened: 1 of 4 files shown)


server/models/procedures/cost_centers.sql
```diff
@@ -45,4 +45,23 @@ BEGIN
...
 
+/*
+@function GetCostCenterByInvoiceUuid(invoice_uuid)
+
+@params invoice_uuid BINARY(16) the invoice uuid
+
+@description
+Returns the cost center id for the invoice using the service_uuid in the invoice table
+*/
+CREATE FUNCTION GetCostCenterByInvoiceUuid(invoice_uuid BINARY(16))
```
jniles, line 56: Nice :+1:

server/models/procedures/stock.sql
```diff
@@ -84,7 +87,7 @@ BEGIN
         projectId as project_id, currencyId as currency_id,
         CONCAT(ig.name, ' - ', m.quantity, ' ', iu.text, ' of ', i.text , ' (', l.label, ')') AS item_description,
         m.uuid, m.description, m.date, m.flux_id, m.is_exit, m.document_uuid, m.quantity, m.unit_cost, m.user_id,
-        ig.cogs_account, ig.stock_account
+        ig.cogs_account, ig.stock_account, m.invoice_uuid, m.entity_uuid
```
jniles, line 90: Nice 👍

jniles (approved):
This looks exactly right.  I've tested it and it works well.

For the stock loss part, we'll need to provide some default cost centers.  There is some discussion here: https://github.com/IMA-WorldHealth/bhima/issues/5883.

As @jmcameron pointed out, we should merge this PR before working on the default stock loss cost center.

image


bors r+

PR #5937, opened 2021-09-23
Add an auto cost_center_id during a simple voucher

### LINKED ISSUE #5935: Automatically associate cost centers tied to an income/expense account in Simple Vouchers
BHIMA has a simplistic interface for creating simple journal vouchers.  This is often used to create double-entry transactions that are simple income/expense receipts.

We need to enhance the server-side with `GetCostCenterByAccountId()` to look up the cost center associated with any accounts that are used and automatically assign them when the transaction is being written.  This will make it so that when someone pays an electricity invoice, the debit is automatically assigned to the electricity cost center.

mbayopanda:
This PR is related to #5935 

image

Close #5935

### DIFF
--- a/server/controllers/finance/vouchers.js
+++ b/server/controllers/finance/vouchers.js
@@ -14,8 +14,6 @@
  * @requires lib/errors/BadRequest
  */
 
-const _ = require('lodash');
-
 const util = require('../../lib/util');
 const db = require('../../lib/db');
 
@@ -267,7 +265,7 @@ function create(req, res, next) {
 }
 
 function createVoucher(voucherDetails, userId, projectId) {
-  let items = voucherDetails.items || [];
+  const items = voucherDetails.items || [];
 
   const voucherType = voucherDetails.type_id;
   const updatesPaiementData = [];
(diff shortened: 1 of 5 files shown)


server/controllers/finance/vouchers.js
```diff
@@ -328,23 +326,32 @@ function createVoucher(voucherDetails, userId, projectId) {
...
-    )
-    .addQuery('CALL PostVoucher(?);', [voucherDetails.uuid]);
+  transaction.addQuery('INSERT INTO voucher SET ?', [voucherDetails]);
+
+  items.forEach(item => {
+    // lookup for cost_center_id from account_id if not given
+    transaction.addQuery(
+      `INSERT INTO voucher_item (
+        uuid, account_id, debit, credit, voucher_uuid, document_uuid, entity_uuid, cost_center_id
+      ) VALUES (?, ?, ?, ?, ?, ?, ?, ${item.cost_center_id ? '?' : 'GetCostCenterByAccountId(?)'})`,
```
jniles, line 340: Can we just do this in SQL entirely?
```sql
IFNULL(?, GetCostCenterByAccountId(?))
```
It might be cleaner than having `?` in Javascript.

server/models/migrations/next/migrate.sql
```diff
@@ -169,3 +169,10 @@ CALL add_constraint_if_missing('voucher_item', 'voucher_item__cost_center_2', 'F
 
 -- Update label in table account_reference_type 
 UPDATE account_reference_type AS art SET art.label = 'FORM.LABELS.COST_CENTER' WHERE art.id = 1;
+
+/**
+ * @author: mbayopanda
+ * @desc: extend column for account_reference table
+ */
+ALTER TABLE `account_reference` MODIFY COLUMN `abbr` VARCHAR(200) NOT NULL;
```
jniles, line 177: Nice 👍

jniles (approved):
Looks pretty good!  Just made one comment, you can decide which form you like more.

server/controllers/finance/vouchers.js
mbayopanda: The thing is, we have to deal with `cost_center_id` and `account_id` for this column, when the `cost_center_id` is not defined we use the `account_id` else we use the given `cost_center_id`.

Using `IFNULL` imply that if the variable given is `null` use the function `GetCostCenterByAccountId` on this `null` variable

PR #5941, opened 2021-09-24
Auto allocation cc for stock loss

mbayopanda:
This PR a new stock setting for cost center to use in case of stock loss

image

When this setting is defined, all stock loss movements will use this cost center

image


jmcameron (commented):
The code looks good.

In order to try this I used bhima_test.  I tried to create a new stock lost center.  When I clicked on [Submit], I got an error message but it did create a center.  I'll attach a screen shot of the create dialog right after clicking submit and also the error message:
image

    errors   code: 'ER_DUP_ENTRY',
    errors   errno: 1062,
    errors   sqlMessage: "Duplicate entry '11' for key 'reference_cost_center.reference_cost_center_1'",
    errors   sqlState: '23000',
    errors   index: 0,
    errors   sql: '\n' +
    errors     '        INSERT INTO reference_cost_center\n' +
    errors     '        (cost_center_id, account_reference_id, is_cost, is_variable, is_turnover) VALUES (7, 11, 1, 1, 0), (7, 11, 1, 0, 0)'

This error is probably due to me chosing the same account for the fixed and variable parts.   When I choose difference accounts the center gets created without error.  But in both cases, new stock loss center is created.   However, when I try to edit the cost center, there is no way to change the accounts associated with it.  

Note that the error message is hidden behind the open modal, so the user cannot necessarily see the error message.  This should probably be fixed, perhaps with a pop-up modal with the error message.

However, when I tried to exit stock to loss, I did not see this posted in the Journal (along with the related cost center).  I'm not sure this part of the functionality has been done yet (or perhaps I've missed a step).

Setting the loss center in the stock settings page works fine.

mbayopanda:
@jmcameron cost center feature is related to stock accounting, so to see this effect on the posting journal you have to activate this : 
image

You're right, we cannot update accounts in the cost center modal. @lomamech must better know the reason of this behavior

jmcameron:
> @jmcameron cost center feature is related to stock accounting, so to see this effect on the posting journal you have to activate this :

After doing this step, it worked correctly and the journal showed the stock exit was assigned to the new stock loss center. 

Nice job!

jniles (approved):
LGTM.

bors r+

PR #5947, opened 2021-09-28
Create report for stock lost during inter-depot transfers.

### LINKED ISSUE #5915: Report on stock lost during transit
In a meeting with Dr. Susan, she asked me:

> Who is looking at the difference in the BL & PV de reception in the health zone dataset?  Is it flagged?

I don't think we have a comprehensive report on this, and we really need one.  It would make BHIMA much more marketable to use on a health zone level if we had this level of reporting.

jmcameron:
Creates a new report to tabulate the stock lost during transfers.
The creation dialog allows a depot to be selected and its role in the transfers (destination, source, or either).
The to/from dates apply to the stock entry to the destination depot.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5915

**TESTING**
- To see the lost stock report, you will need some lost stock!   So either use a production database where you know there is lost stock, or use bhima_test and construct the lost stock transfers yourself
- If you use bhima_test or a production database with no losses, try the report first (Stock > Reports > Lost Stock Report) and verify that an empty report is generated.
- To create lost stock incidents:
   - Suppose the destination is Depot B and the source is Depot A.
   - If a requisition does not already exist for the transfer, create it.
   - Exit the desired stock from Depot A (note the quantity)
   - Enter the desired stock into Depot B (with requisition) -- AND do not accept the full count.  This emulates some stock being lost during the transfer.
   - To be most useful for testing, do the same procedure going the opposite direction
- Check the report:   Stock > Reports > Lost Stock Report
- Select Depot B for the depot
- Select the depot role as "Destination"
- Choose to/from date, etc.
- View the report.
- Try the other roles (Source, Either)

**OPEN QUESTION**
- How to handle deliveries with more stock received than sent (ie, miscount at origin depot)

### DIFF
--- a/client/src/i18n/fr/report.json
+++ b/client/src/i18n/fr/report.json
@@ -136,7 +136,17 @@
       "TITLE" : "Rapport de changement d'inventaire",
       "DESCRIPTION": "Ce rapport montre les changements survenus sur les articles en stock au cours d'une période"
     },
-
+    "LOST_STOCK_REPORT": {
+      "TITLE": "Rapport sur les stocks perdus",
+      "DEPOT_ROLE": "Lister les pertes de transfert de stock pour ce dépôt comme",
+      "DESCRIPTION": "Ce rapport répertorie les stocks perdus lors du transfert de stock d'un dépôt à un autre.",
+      "DESTINATION": "Destination",
+      "DESTINATION_DEPOT" : "Dépôt de destination",
+      "EITHER": "L'un ou l'autre",
+      "LOSS": "Perte",
+      "SOURCE": "Source",
+      "SOURCE_DEPOT" : "Dépôt de source"
+    },
     "OPENING_BALANCE": "Balance d'ouverture",
     "ORDER": {
       "LAST_TRANSACTION": "Date de la dernière transaction",
@@ -197,6 +207,7 @@
     "SINCE": "Depuis",
     "SIXTY_TO_NINETY_DAYS": "60 a 90 jours",
     "THIRTY_TO_SIXTY_DAYS": "30 a 60 jours",
+    "TOTALS": "Totaux",
     "VIEW_ACCOUNT_STATEMENT": "Voir le Rélevé de Compte",
     "VIEW_CREDIT_NOTE": "Voir la note de credit",
     "VIEW_INVOICE": "Voir les factures",
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -49,6 +49,7 @@ module.exports = {
   getLotsDepot,
   getLotsMovements,
   listStatus,
+  listLostStock,
   // stock consumption
   getInventoryQuantityAndConsumption,
   getInventoryMovements,
@@ -469,8 +470,6 @@ async function getMovements(depotUuid, params) {
     LEFT JOIN depot AS dp ON dp.uuid = m.entity_uuid
     LEFT JOIN document_map dm2 ON dm2.uuid = m.entity_uuid
     LEFT JOIN document_map sr_m ON sr_m.uuid = m.stock_requisition_uuid
-
-
   `;
 
   const finalClause = 'GROUP BY document_uuid, is_exit';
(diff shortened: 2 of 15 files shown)


server/controllers/stock/core.js
```diff
@@ -480,6 +479,59 @@ async function getMovements(depotUuid, params) {
...
+
+  db.convert(params, ['depot_uuid']);
+
+  const { depotRole } = params;
+  delete params.depotRole;
+
+  params.fake_field = true;
+  const filters = new FilterParser(params, { tableAlias : 'dest' });
+
+  filters.custom('fake_field', 'dest.flux_id = (?) AND IFNULL((ex.quantity - dest.quantity), 0) <> 0',
```
jniles: You probably should add a comment describing what you are doing here.  I would also rename this from "fake_field" to `unbalanced_quantities` or something similar that is related to the thing being looked up.

server/models/migrations/next/migrate.sql
```diff
@@ -176,3 +178,14 @@ UPDATE account_reference_type AS art SET art.label = 'FORM.LABELS.COST_CENTER' W
...
+/**
+* author: @jmcameron
+* date: 2021-09-24
+* description: Add lost stock report menu item
+*/
+INSERT IGNORE INTO `unit` VALUES
+  (300, 'Lost Stock Report', 'TREE.LOST_STOCK_REPORT', 'Report on stock lost during depot transfers', 282, '/reports/lost_stock_report');
+
+INSERT IGNORE INTO `report` (`report_key`, `title_key`) VALUES
+   ('lost_stock_report', 'LOST_STOCK_REPORT');
```
jniles, line 197: This should be `TREE.LOST_STOCK_REPORT`.

jniles (commented):
image

I can confirm that this works!  Excellent job.

I've made a suggestion or two.  I would like @mbayopanda to take a look at the French.

jmcameron:
@mbayopanda Please review my french translation items and suggest improvements.  Thanks!

jniles:
@jmcameron you'll also need to rebase this.

jmcameron:
@jniles Rebased.

client/src/i18n/fr/report.json
```diff
@@ -136,7 +136,17 @@
       "TITLE" : "Rapport de changement d'inventaire",
       "DESCRIPTION": "Ce rapport montre les changements survenus sur les articles en stock au cours d'une période"
     },
-
+    "LOST_STOCK_REPORT": {
+      "TITLE": "Rapport sur les stocks perdus",
+      "DEPOT_ROLE": "List stock transfer losses for this depot as",
```
mbayopanda: In French you can say: Lister les pertes de transfert de stock pour ce dépôt comme

client/src/i18n/fr/report.json
```diff
@@ -136,7 +136,17 @@
...
+      "DESCRIPTION": "Ce rapport répertorie les stocks perdus lors du transfert de stock d'un dépôt à un autre.",
+      "DESTINATION": "Destination",
+      "DESTINATION_DEPOT" : "Dépôt de destination",
+      "EITHER": "Soit",
```
mbayopanda: image

If the context of either here is any, you can say: `L'un ou l'autre` since there are just two cases

mbayopanda (approved):
There is just some french translation to fix, but the PR is LGTM 👍🏽

client/src/i18n/fr/report.json
jmcameron: Fixed.  Thanks!

client/src/i18n/fr/report.json
jmcameron: Fixed. Thanks!

server/controllers/stock/core.js
jmcameron: Done.

server/models/migrations/next/migrate.sql
jmcameron: Fixed.

PR #5956, opened 2021-09-30
fix(cost centers): filter on reference type

jniles:
This commit fixes the cost center functions to ensure that we are filtering the account references by the cost center id.  I've also combined both of the functions for calculating cost centers into a single file.


jniles:
@lomamech can I get a review?

server/models/procedures/cost_centers.sql
```diff
@@ -20,7 +20,37 @@ RETURNS MEDIUMINT(8) DETERMINISTIC
 BEGIN
   RETURN (
     SELECT cost_center_id FROM reference_cost_center WHERE account_reference_id IN (
-      SELECT account_reference_id FROM account_reference_item WHERE account_reference_item.account_id = account_id
+      SELECT account_reference_id FROM account_reference_item JOIN account_reference
+        ON account_reference_item.account_reference_id = account_reference.id
+        WHERE account_reference_item.account_id = account_id
+          and account_reference.reference_type_id =  1 -- reference_type_id 1 is cost center
```
mbayopanda, line 26: 👍🏽

mbayopanda (approved):
Good refactoring and it LGTM 👍🏽

PR #5961, opened 2021-09-30
Fix(Role management) display the add button if necessary

### LINKED ISSUE #5958: Bug: Role management add button is broken
The "add role" button is permanently broken due to a bug found by @jeremielodi .

The `has-permission` property is incorrectly set.  It needs to be map to something in the current scope, but refers to the ui-grid's scope which it doesn't actually have access to.  See below:

https://github.com/IMA-WorldHealth/bhima/blob/3b3c4d0a55a273dba06d4119aed8200a233b9781/client/src/modules/roles/roles.html#L8

jeremielodi:
fixes #5958

### DIFF
--- a/client/src/modules/roles/roles.html
+++ b/client/src/modules/roles/roles.html
@@ -5,7 +5,7 @@
       <li class="title" translate>TREE.ROLE_MANAGEMENT</li>
     </ol>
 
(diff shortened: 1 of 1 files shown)


jniles:
Thanks @jeremielodi !

bors r+

PR #5973, opened 2021-10-05
feat(cost centers): add is_editable flag

### LINKED ISSUE #5964: Add is_editable flag to cost_center_allocation_basis table
Since some of the cost center bases are calculated by the system (direct cost, number of employees), etc, it should not be possible to edit them.  

This is not the same as is_predefined.  'is_predefined' bases could be either computed by the system (not editable) or entered manually (editable),.

jniles:
This commit adds an "is_editable" flag to the cost center allocation basis to indicate if the basis should be able to be modified by the user.

Closes #5964.

image


jmcameron (changes requested):
Please hold off on this PR until we discuss this.   Please see my note in the Issue: 
https://github.com/IMA-WorldHealth/bhima/issues/5964#issuecomment-934163537

jniles:
@jmcameron what is your thought on this?  Shall I close it into oblivion?

jmcameron:
I don't think it is necessary.   We can create a new issue if it becomes necessary in the future.

PR #5975, opened 2021-10-05
feat(cost centers): add grid save state

### LINKED ISSUE #5959: Save grid state on cost centers grid
We need to be able to save the grid state on the cost centers grid.  Because the grid reloads each time, it makes the work flow really hard to always need to scroll to the bottom of the grid to keep working.  Adding in the grid save state module will help with this by preserving filters and sort ordering.

YMz9mDe0yA

jniles:
Adds the save-state configuration to the `ui-grid` to make working with the grid a slight bit easier.

Closes #5959.

xriQOIuuK4

**Testing Instructions**
1. Load the cost center registry.
2. Change the sort order.
3. Use the dropdown menu in the top right to save the grid state.
4. Refresh the page
5. Observe that the sort order has been preserved
6. Use the dropdown menu to clear the grid state
7. Observe that the sort order has returned to default.

### DIFF
--- a/client/src/modules/cost_center/cost_center.html
+++ b/client/src/modules/cost_center/cost_center.html
@@ -21,6 +21,21 @@
                 <i class="fa fa-table"></i> <span translate>ALLOCATION_BASES</span>
               </a>
             </li>
+
+            <li role="separator" class="divider"></li>
+
+            <li role="menuitem">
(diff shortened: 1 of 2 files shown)


jmcameron (approved):
Code looks good.
I tested it and works as described.

PR #5976, opened 2021-10-05
feat(cost center): add link to edit for bases

### LINKED ISSUE #5957: Directly link allocation key(s) from cost center
Here is the cost center configuration page:

image

To configure the number of square meters or employees in in a cost center, I need to go to the allocation keys management page and link them from there.  Here is what that page looks like:

image

It would be really nice to add the option of directly modifying the allocation keys from the cost center dropdown itself.  This is probably the most typical use case anyway.

@jmcameron @mbayopanda is this a reasonable thing to do?

jniles:
Adds a link to the cost center allocation basis edit modal.

Closes #5957.


jniles:
@mbayopanda, I've fixed all the bugs in this.  Can you give me a review?

jmcameron (commented):
I looked this over.  It looks good to me.  I tested this with bhima_test and the sample data and ti seems to work well.

However, I found one thing disconcerting:  When you use the new "Edit allocation basis" for one of the cost centers, it does bring up the right edit modal.  However when you are finished with it, it leaves you on the cost_center/allocation_bases (Allocation Bases) page.  It seems to me that it should put you back on the Cost center registry page.

jniles:
@jmcameron this is a good suggestion.  I'll take a whack at it, but I have a feeling it will require a fair amount of work.  If it isn't an easy fix, I'll make an issue about it for later.

jmcameron:
> @jmcameron this is a good suggestion. I'll take a whack at it, but I have a feeling it will require a fair amount of work. If it isn't an easy fix, I'll make an issue about it for later.

Given the time crunch, just create an issue and release  the PR.  We can fix that later.

-Jonathan

jniles:
I think I got this working.  Test again?

jmcameron:
Works great!

jmcameron (approved):
Looks good and works fine.  
Nice fix with the cancel and submit going back to the Cost Center page.  That seems more natural to me.

PR #5977, opened 2021-10-05
Automatically add number of employees to cost centers in step-down cost allocation reports

### LINKED ISSUE #5962: Create automatic allocation basis calculation for "number of employees" basis
We should be able to automatically compute certain allocation bases instead of filling them out.  This was previously mentioned in https://github.com/IMA-WorldHealth/bhima/issues/5879#issuecomment-904703477 and discussed further in https://github.com/IMA-WorldHealth/bhima/issues/5875.

This issue describes creating an automatic calculation for the "number of employees" basis.

This is pretty straightforward.  To compute the number of employees in all the services, the following SQL query will work:
```sql
SELECT service.name, COUNT(employee.uuid), GetCostCenterByServiceUuid(service.uuid) 
FROM service JOIN employee ON service.uuid = employee.service_uuid 
GROUP BY service.uuid;
```
To filter by a single cost center, just add a WHERE component.
```sql
SELECT service.name, COUNT(employee.uuid), GetCostCenterByServiceUuid(service.uuid) 
FROM service JOIN employee ON service.uuid = employee.service_uuid 
WHERE GetCostCenterByServiceUuid(service.uuid) = ? 
GROUP BY service.uuid;
```

The first query returns something like this:
<pre>+------------------------+---------------+----------------+
| Service                | Num Employees | Cost Center ID |
+------------------------+---------------+----------------+
| Administration         |             1 |             15 |
| Aumonerie              |             5 |             34 |
| Dispensaire            |             7 |             13 |
| Maternité              |             8 |             25 |
| Gyneco-obstetrique     |             3 |             17 |
+------------------------+---------------+----------------+
</pre>

----

This will give us the ability to automatically compute the allocation bases for each of the principal cost centers, assuming that one service has a single principal cost center.

jmcameron:
This PR adds/updates the number of employees for each cost center based on the query proposed by @jniles in its issue:

Closes https://github.com/IMA-WorldHealth/bhima/issues/5962

I think I was able to verify that computes the correct number of employees for each cost center both in bhima_test and in the vanga dataset (after applying the employees and cost center update scripts from @jniles and @lomamech (respectively).

To use:  Go to the "Allocation Keys" page and click on the [Update] button.  That will update the cost_center_allocation_base_value table with the items that it can compute (currently only num employees).

This does not work completely yet, because 
- For bhima_test, the sample data does not set appropriate services and link them to cost centers for the new cost centers
- For vanga, the code to construct the data for the step-down cost allocation is producing empty data.  Further work is needed to connect all  the pieces.
- Did not test on imck, etc.

### DIFF
--- a/server/controllers/finance/stepDownCostAllocationQuantity.js
+++ b/server/controllers/finance/stepDownCostAllocationQuantity.js
@@ -1,5 +1,6 @@
 const db = require('../../lib/db');
 const FilterParser = require('../../lib/filter');
+const constants = require('../../config/constants');
 
 module.exports = {
   create,
@@ -10,6 +11,7 @@ module.exports = {
   bulkCreate,
   bulkDelete,
   bulkUpdate,
+  updateQuantities,
 };
 
 // add a new allocation basis quantity
@@ -33,7 +35,7 @@ function create(req, res, next) {
 //
 async function bulkDetails(req, res, next) {
   const sql = `
-    SELECT 
+    SELECT
       abv.id, abv.quantity, abv.cost_center_id, abv.basis_id,
       cc.label AS cost_center_label, ab.name AS allocation_basis_label, ab.units AS allocation_basis_units
     FROM cost_center_allocation_basis_value abv
@@ -117,7 +119,7 @@ function bulkDelete(req, res, next) {
 //
 function list(req, res, next) {
   const sql = `
-    SELECT 
+    SELECT
       abv.id, abv.quantity, abv.cost_center_id, abv.basis_id,
       cc.label AS cost_center_label, ab.name AS allocation_basis_label, ab.units AS allocation_basis_units
     FROM cost_center_allocation_basis_value abv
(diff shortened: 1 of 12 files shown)


server/controllers/finance/cost_center_allocation_registry.js
```diff
@@ -35,6 +38,35 @@ async function fetch(session, params) {
...
+
+  // Get the allocation bases that are computable
+  const cabQuery = 'SELECT id, name FROM cost_center_allocation_basis WHERE is_computed = 1';
+  const computables = await db.exec(cabQuery);
+
+  // Compute and set each computable allocation basis quantity
+  const missingCenters = [];
+  for (let i = 0; i < computables.length; i++) {
+    const basis = computables[i];
+    const data = await allocationQuantities(basis.id);
```
jmcameron: @jniles How do I deal with this?

FIXED.

jniles (commented):
It would be awesome if this showed up in the cost center allocation registry.  Can you get that working? 

image

It would also look good to put this on the modal, but in an uneditable state.

jmcameron:
> It would be awesome if this showed up in the cost center allocation registry. Can you get that working?
> 
> It would also look good to put this on the modal, but in an uneditable state.

Ok, I'll see what I can do.

-Jonathan

server/controllers/finance/cost_center_allocation_registry.js
```diff
@@ -111,5 +144,31 @@ async function list(req, res, next) {
...
+function allocationQuantities(allocationBasisId) {
+  let query = null;
+  if (allocationBasisId === constants.allocationBasis.ALLOCATION_BASIS_NUM_EMPLOYEES) {
+    // Set up the query for number of employees
+    query = `
+      SELECT BUID(service.uuid) AS service_uuid, service.name AS service_name,
+        COUNT(employee.uuid) AS ALLOCATION_BASIS_NUM_EMPLOYEES,
+        GetCostCenterByServiceUuid(service.uuid) as cost_center_id
+      FROM service JOIN employee ON service.uuid = employee.service_uuid
+      GROUP BY service.uuid;
```
mbayopanda: The query works very well 👍🏽 : 
image

Here are employees of `Buanderie` service
image

mbayopanda (commented):
The query for getting employees as cost center basis works well, and the concern is we cannot see this computed basis in the `allocation_bases` registry: 

image

server/controllers/finance/stepDownCostAllocationQuantity.js
```diff
@@ -166,3 +168,87 @@ function remove(req, res, next) {
...
+        // TODO:  Do we want to zero out the others here?
+        unUpdatedCenters.push({ cc });
+      }
+    }
+  }
+
+  // TODO: Do something with upUpdateCenters to inform the user
+  // console.log("Centers not updated: ", unUpdatedCenters);
+
+  // TODO: Handle errors better
```
jniles: I would wrap this whole thing in a `try`/`catch` and then use `next()` to send the error back to the client.

server/controllers/finance/stepDownCostAllocationQuantity.js
```diff
@@ -166,3 +168,87 @@ function remove(req, res, next) {
...
+    for (let k = 0; k < costCenters.length; k++) {
+      const cc = costCenters[k];
+
+      // Get the data for this cost center (if any)
+      const ccData = data.find(item => item.cost_center_id === cc.id);
+      if (ccData) {
+        const newQuantity = ccData[basis.name];
+        // See if there is an existing quantity
+        const qRecordId = await db.exec(findQRec, [cc.id, basis.id]);
+        if (qRecordId.length > 0) {
```
jniles: Can we just delete these values and rewrite?  I.e. if we are computing the number of employees, just clear out the old values and then do an INSERT for the rest?

server/controllers/finance/cost_center_allocation_registry.js
jmcameron: Thanks!

server/controllers/finance/stepDownCostAllocationQuantity.js
jmcameron: Done. Thx

server/controllers/finance/stepDownCostAllocationQuantity.js
jmcameron: I'm happy to do that after the demo.   By doing drop and insert, I can do it with transactions and clean it up.  Lets get this out so we can use it for the demo.

jniles (approved):
I've tested this with the Vanga database, and it works like a charm.  Thanks!

bors r+

PR #5978, opened 2021-10-05
Filter null cost center

### LINKED ISSUE #5970: Add search on cost center or none to the Journal
We should add the ability to search for a Cost Center to the posting journal.  To make this more useful, we'll want to have the ability to also search for "NULL", that way, we can detect transactions that are not assigned to cost centers and should be.  This would let us detect transactions that need to have costs assigned before posting.

mbayopanda:
This PR is related to #5970 

- Adds a way to search transactions with null cost_center_id or principal_center_id
- Adds a way to search by cost_center_id and principal_center_id
- Enhances the `bhCostCenterSelect` to correctly handle `is_required` property and transclude

Closes #5970.

### DIFF
--- a/client/src/js/components/bhCostCenterSelect/bhCostCenterSelect.js
+++ b/client/src/js/components/bhCostCenterSelect/bhCostCenterSelect.js
@@ -37,8 +37,9 @@ function CostCenterSelectController(CostCenters, Notify) {
   }
 
   $ctrl.$onInit = () => {
+    $ctrl.required = !!($ctrl.required);
     $ctrl.label = $ctrl.label || 'COST_CENTER.TITLE';
-    $ctrl.costCenterId = +$ctrl.costCenterId;
+    $ctrl.costCenterId = $ctrl.costCenterId ? +$ctrl.costCenterId : null;
     loadCostCenters();
   };
 
--- a/client/src/js/components/bhCostCenterSelect/bhCostCenterSelect.js
+++ b/client/src/js/components/bhCostCenterSelect/bhCostCenterSelect.js
@@ -12,10 +12,11 @@ angular.module('bhima.components')
       disabled         : '<?',
       label            : '@?',
       helpText         : '@?',
+      enableNull       : '<?',
     },
   });
 
-CostCenterSelectController.$inject = ['CostCenterService', 'NotifyService'];
+CostCenterSelectController.$inject = ['CostCenterService', 'NotifyService', '$translate'];
 
 /**
  * @function CostCenterSelectionController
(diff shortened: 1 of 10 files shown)


client/src/modules/journal/modals/search.modal.html
```diff
@@ -137,6 +137,37 @@
...
+        <div class="form-group">
+          <label class="control-label" translate>TRANSACTIONS.NO_COST_CENTER_TRANSACTION</label>
+          <bh-clear on-clear="ModalCtrl.clear('showOnlyNullCostCenter')"></bh-clear>
+          <div class="checkbox">
+            <p class="text-info">
+              <i class="fa fa-info-circle"></i>
+              <span translate>TRANSACTIONS.ONLY_EXPLOITATION_ACCOUNTS</span>
+            </p>
+            <label>
+              <input ng-model="ModalCtrl.searchQueries.showOnlyNullCostCenter" ng-true-value="1" ng-false-value="0" type="checkbox"> <span translate>TRANSACTIONS.SHOW_ONLY_TRANSACTION_WITH_NULL_COST_CENTER</span>
```
jniles, line 150: Instead of making this separate, can we just add a cost center to the `bh-cost-center-select` with id NULL?

Then we could choose where `principal_center_id` is 3, and `cost_center_id` is NULL.  Is that possible?

client/src/js/components/bhCostCenterSelect/bhCostCenterSelect.js
```diff
@@ -37,8 +37,9 @@ function CostCenterSelectController(CostCenters, Notify) {
   }
 
   $ctrl.$onInit = () => {
+    $ctrl.required = !!($ctrl.required);
     $ctrl.label = $ctrl.label || 'COST_CENTER.TITLE';
-    $ctrl.costCenterId = +$ctrl.costCenterId;
+    $ctrl.costCenterId = $ctrl.costCenterId ? +$ctrl.costCenterId : null;
```
jmcameron, line 51: What does the + do here?

jniles: Tries to coerce a string into a number, I assume.  It's not very clean code. 😃

client/src/modules/journal/modals/search.modal.html
mbayopanda: The `showOnlyNullCostCenter` filter is not only for `cost_center_id`, but for `cost_center_id` and `principal_center_id` and only for income/expense accounts. It allows to check if `cost_center_id` and `principal_center_id` are null.

mbayopanda: Your idea of having a cost center with id NULL will help to filter null value for `cost center` or `principal center` column

mbayopanda:
@jniles changes landed

jniles (approved):
I still think we could do better by rolling everything into two different `bhCostCenterSelect`, but we can discuss this over time.  Looks good to me.

bors r+

PR #5981, opened 2021-10-06
View transactions from cost center page

### LINKED ISSUE #5971: Add "View Transactions" link from cost centers to posting journal
Once we implement https://github.com/IMA-WorldHealth/bhima/issues/5970, we should add a link from the cost centers registry to the posted transactions in the journal, filtered on that cost center ID.  It would go in the dropdown menu below:

image

mbayopanda:
This PR is related to #5971 

Close #5971


jniles (approved):
This is exactly right!

bors r+

PR #5982, opened 2021-10-06
fix(cost center): uniquely show auxiliary cost centers in columns

### LINKED ISSUE #5979: Do not show principal cost centers in the columns of step-down cost allocation report
Here is the current step-down cost allocation report, we currently have _every_ cost center listed in the columns.   This doesn't make sense.  We should only have the cost centers that are being allocated in the columns. 

Here is the current report with fake data from the Vanga dataset:

image

Notice how many columns we have!  But we only have 3 auxiliary cost centers that are allocating their costs.  We should only have three columns.

Here is an example for Essentials of Healthcare Ch. 13:

image

See how there are four cost centers, but only two columns, since only two of the four cost centers are auxiliary cost centers.  This is what we should be doing.

jniles:
This commit fixes the presentation of the cost center report to only show the allocation ratios for the auxiliary cost centers in the columns.

Closes #5979.

image

### DIFF
--- a/server/controllers/finance/reports/cost_center_step_down/report.handlebars
+++ b/server/controllers/finance/reports/cost_center_step_down/report.handlebars
@@ -29,8 +29,10 @@
               <th>{{translate 'FORM.LABELS.SERVICE'}}</th>
               <th>{{translate 'COST_CENTER.DIRECT_COST'}}</th>
               {{#each data as | service |}}
+                {{#unless service.is_principal}}
                 <th>{{ service.cost_center_label }}</th>
                 <th style="width: 3em;">%</th>
+                {{/unless}}
               {{/each}}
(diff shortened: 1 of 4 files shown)


server/controllers/finance/reports/cost_center_step_down/report.handlebars
```diff
@@ -29,8 +29,10 @@
               <th>{{translate 'FORM.LABELS.SERVICE'}}</th>
               <th>{{translate 'COST_CENTER.DIRECT_COST'}}</th>
               {{#each data as | service |}}
+                {{#unless service.is_principal}}
```
mbayopanda, line 32: 👍🏽

mbayopanda (approved):
Only auxiliary cost centers columns are shown 👍🏽 

LGTM

bors r+

PR #5986, opened 2021-10-06
Fix cost center typos

mbayopanda:
This PR fixes cost center typos and adds a way to show the allocation bases table in the allocation keys module.

image


jmcameron (approved):
Looks good.  I tested it and it works fine.  I have one suggestion.  Please change the title  of the edit modal from "Edit allocation keys" to "Edit allocation quantities"  (we are not editing the keys/bases themselves, but we are editing their quantities/values.

Once you do that, please go ahead and release it.

jmcameron:
Thanks for the updates!

PR #5987, opened 2021-10-06
feat(cost center): include/exclude revenue centers

### LINKED ISSUE #5985: Add option to cost center step-down report to exclude revenue
Right now, we merge revenue and costs together, which makes the analysis a bit confusing if all we want to know is costs.  We should be able to distinguish between the revenue and the costs and only show one or the other.

jniles:
Adds the ability to toggle the revenue in the cost center step down report.

Closes #5985.


jmcameron (approved):
I looked at the code and it looks fine.

I tested it with bhima_test + sample data and the interface worked, but there was no difference in the step-down analysis.

With help from @jniles I was able to download an updated Vanga dataset and was able to see that the step-down analysis worked and showed different results with and without the "include revenue" flag. 

Based on that limited testing, I think it is okay.

PR #6002, opened 2021-10-08
Add currency select in the cc report

### LINKED ISSUE #5990: Add currency selection for step-down cost analysis report

mbayopanda:
This PR adds currency select in the cost center allocation report

image

Closes #5990.

### DIFF
--- a/client/src/modules/reports/generate/cost_center_step_down/cost_center_step_down.config.js
+++ b/client/src/modules/reports/generate/cost_center_step_down/cost_center_step_down.config.js
@@ -35,6 +35,10 @@ function CostCenterStepdownReportConfigController($sce, Notify, SavedReports, Ap
     vm.reportDetails.include_revenue = bool;
   };
 
+  vm.onSelectCurrency = (currency) => {
+    vm.reportDetails.currency_id = currency.id;
+  };
+
   vm.clearPreview = function clearPreview() {
     vm.previewGenerated = false;
     vm.previewResult = null;
--- a/server/controllers/finance/cost_center_allocation_registry.js
+++ b/server/controllers/finance/cost_center_allocation_registry.js
@@ -8,8 +8,10 @@ const db = require('../../lib/db');
 const stepdown = require('../../lib/stepdown');
 const fiscal = require('./fiscal');
 const ccAllocationKeys = require('./cost_center_allocation_bases');
+const Exchange = require('./exchange');
 
 async function fetch(session, params) {
+  const enterpriseId = session.enterprise.id;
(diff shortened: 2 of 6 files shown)


server/controllers/finance/cost_center_allocation_registry.js
```diff
@@ -18,8 +20,27 @@ async function fetch(session, params) {
...
+    firstCurrency = params.currency_id;
+    secondCurrency = enterpriseCurrencyId;
+  }
+
+  const query = 'CALL ComputeCostCenterAllocationByIndex(?, ?, ?, ?, ?);';
+  let [costCenters] = await db.exec(query, [
+    range.dateFrom,
+    range.dateTo,
+    params.include_revenue,
+    enterpriseId,
```
jniles: Let's just get this from the enterprise table, instead of passing it in.  We aren't planning on supporting multiple enterprises anytime soon.

jmcameron (commented):
The code looks good.

I tested this with bhima_test and the sample data.  The currency selection works and the resulting report produces correctly converted currencies.

Once the comment by @jniles is resolved, I think this is ready to release.

mbayopanda:
@jmcameron changes landed

client/src/modules/reports/generate/cost_center_step_down/cost_center_step_down.config.js
```diff
@@ -35,6 +35,10 @@ function CostCenterStepdownReportConfigController($sce, Notify, SavedReports, Ap
     vm.reportDetails.include_revenue = bool;
```
jmcameron, line 35: I suggest getting the default currency from the session.enterprise and assigning it here to vm.reportDetails.currency_id so that <bh-currency-select> in the html file will have a default value.  We do this in many other Bhima forms.

jmcameron (commented):
Looks good.  I've tested it and it works.   I added one suggestion.

Was not 100% what @jniles was asking, so I'm not sure whether you have satisfied his request.

mbayopanda:
> 
> 
> Looks good. I've tested it and it works. I added one suggestion.
> 
> Was not 100% what @jniles was asking, so I'm not sure whether you have satisfied his request.

@jniles asked for not sending enterpiseId as parameter of the procedure since we have just one enterprise and this enterpriseId will never change we have just to query directly from the enterprise table inside the stored procedure

jmcameron:
> @jniles asked for not sending enterpiseId as parameter of the procedure since we have just one enterprise and this enterpriseId will never change we have just to query directly from the enterprise table inside the stored procedure

Okay

PR #6008, opened 2021-10-11
feat(stock): add invoice description

### LINKED ISSUE #5998: Add invoice description to stock exit to patient modal
The MD of Vanga asks if we can add the invoice description to the stock exit to patient modal.

### LINKED ISSUE #5999: Do not allow stock distributions to patients without an invoice
BHIMA currently allows you to distribute to a patient with or without an invoice via the "join an invoice" option.  We must now require an invoice to make cost centers analysis work.

jniles:
This commit adds an invoice description to the stock exit modal.  It
also makes the invoice a required field.

Closes #5998.
Closes #5999.

image


client/src/modules/stock/exit/modals/findPatient.modal.html
```diff
@@ -14,65 +14,56 @@
...
-            ng-click="$ctrl.getLastInvoice()">
-          <span translate>STOCK.JOIN_INVOICE</span>
-        </label>
-      </div>
+    <fieldset ng-disabled="$ctrl.loading">
+      <bh-find-patient
+        patient-uuid="$ctrl.selected.uuid"
+        suppress-reset="false"
+        on-search-complete="$ctrl.setPatient(patient)">
+      </bh-find-patient>
```
mbayopanda, line 23: 👍🏽

mbayopanda (commented):
The description looks good 👍🏽 

I have just noticed that: 
- It seems like the search invoice input doesn't work well, when you click on `Rechercher` you get this error 
image

- Submit button it's still disabled 
image

Also, make that the invoice must be required is a good thing, I hope only that we will not encounter case of distribution to patient which doesn't require invoice

jniles:
@mbayopanda thanks for the review!  I've made the changes.  Please test this again.

mbayopanda (approved):
LGTM 👍🏽

jmcameron (commented):
I tested this using bhima_test and found one problem.  When you enter anything into the "Please enter invoice ID" search area and hit the [Search] button, there is an error.  See this screenshot.
image

Where the error occurred is apparently in bhFindPatient.js on line 61.  Apparently invoice.details is not available.   This needs to be fixed.

When I selected an invoice and clicked [Submit] it worked fine.  

**A couple of suggestions regarding the user interface**
- Since it is most likely that the user will select from the list of inventory items, I suggest moving that above the invoice ID search box.
- Since we are now requiring the user to select an invoice, we should add a title above the list of invoices and search box that says:  "Select an invoice"
- And it might be a little friendlier if the title above the invoice search box said: "Or search by invoice ID"

jmcameron:
One more comment about this PR.  Based on the discussion at the Demo(?), it seems like the information should include the name of some/all of the drugs in the invoice.  That way it is easier for the user to select the correct invoice.  This could be added as a simple list of the drug/stock names (without amounts, etc) and included as part of the description.

jniles:
@jmcameron I disagree that the list of drugs should be present, and I don't think that was the conclusion of the meeting.  IMCK regularly invoices for tens of drugs at a time.  My understanding of the conclusion of the discussion was that the description was what Dr. Rice explicitly requested.

jniles:
You are right about the error - I need to figure out how to show red on the search input on error.

jniles:
@jmcameron please test this again.  I think I have it fixed.

jniles:
Just ran it by the MD of Vanga.  He thinks this is fine.

client/src/modules/stock/exit/modals/findPatient.modal.js
```diff
@@ -9,51 +9,46 @@ StockFindPatientModalController.$inject = [
...
   // set patient
   function setPatient(patient) {
-    vm.joinInvoice = false;
-    vm.invoiceSelected = false;
     vm.selected = patient;
+    return loadRecentInvoices();
   }
 
-  function getLastInvoice() {
+  function loadRecentInvoices() {
```
jmcameron, line 51: 👍

jmcameron (approved):
Looks good to me.  

I still think the UI could be improved a little per my earlier suggestions:
- Since it is most likely that the user will select from the list of inventory items, I suggest moving that above the invoice ID search box.
- Since we are now requiring the user to select an invoice, we should add a title above the list of invoices and search box that says: "Select an invoice"
- And it might be a little friendlier if the title above the invoice search box said: "Or search by invoice ID"

jniles:
Landed all the changes that @jmcameron asked for except renaming the label on the invoice.  I'll do that in a separate PR.

bors r+

PR #6012, opened 2021-10-11
Fix setting computed cost center allocation basis quantities

### LINKED ISSUE #6011: Automatic bases computation should delete all previous values associated with the basis
The algorithm to automatically compute bases needs to delete all previous values in that basis.  This bit us during the presentation last week, where we had dangling values inflating our numbers.   We should ensure that the basis value is cleaned up before doing an INSERT.

jmcameron:
Redo setting computed cost center allocation basis quantities to delete previous values in order to avoid stale duplicate values.

This PR also updates the Cost Center registry page and the Allocation Bases page to list the cost centers alphabetically.

NOTE: This re-work zeros out the values of all allocation basis values (for all cost centers) that do not have computed values.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6011

Tested: 

- bhima_test (with sample data).  This setup only has one employee currently assigned to any cost center so is not very interesting.
- vanga (JNiles version of Oct 5, 2021).   The fix works with this dataset.  However before applying the update, the total number of employees is 232.  After the update, the total is 200.  I suspect this has to do with issues with the vanga dataset, not this update/computation.

**TESTING**
- Use the vanga dataset from JNiles (Oct 5 or newer).
- Try  Cost Centers registry.  Note the before and after values below.  Some centers with zero employees are now showing (because we set them to zero if the computation does not provide a value).  Some centers with had some employees. Doing the [Update] set them to zero.  That is probably due to an issue with the Vanga dataset.
   
image

### DIFF
--- a/server/controllers/finance/stepDownCostAllocationQuantity.js
+++ b/server/controllers/finance/stepDownCostAllocationQuantity.js
@@ -14,21 +14,6 @@ module.exports = {
   updateQuantities,
 };
 
-// add a new allocation basis quantity
-//
-// POST /cost_center_allocation_basis_quantity
-//
-function create(req, res, next) {
-  const sql = `INSERT INTO cost_center_allocation_basis_value SET ?`;
-  const data = req.body;
-  db.exec(sql, data)
-    .then(() => {
-      res.sendStatus(201);
-    })
-    .catch(next)
-    .done();
-}
-
 // get details of all allocation base quantities for a given cost center
 //
 // GET /cost_center_allocation_basis_quantity/bulk/:id
@@ -141,6 +126,21 @@ function list(req, res, next) {
     .done();
 }
 
+// add a new allocation basis quantity
+//
+// POST /cost_center_allocation_basis_quantity
+//
+function create(req, res, next) {
+  const sql = `INSERT INTO cost_center_allocation_basis_value SET ?`;
+  const data = req.body;
+  db.exec(sql, data)
+    .then(() => {
+      res.sendStatus(201);
+    })
+    .catch(next)
+    .done();
+}
+
 // update allocation basis quantity details
 //
 // PUT /cost_center_allocation_basis_quantity/:id
@@ -173,9 +173,7 @@ function remove(req, res, next) {
 //
 // UPDATE /cost_center_allocation_basis_quantities_update
 //
-
 async function updateQuantities(req, res, next) {
-
   try {
     // Get the full list of cost center IDs
     const costCentersQuery = 'SELECT id, label AS name from cost_center';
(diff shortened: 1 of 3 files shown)


client/src/modules/cost_center/cost_center.service.js
```diff
@@ -39,7 +39,7 @@ function CostCenterService(Api, $uibModal) {
       .then(service.util.unwrapHttpResponse);
   };
 
-  service.getAllocationQuantities = (options) => {
+  service.getcomputedAllocationQuantities = (options) => {
```
jniles: Can we make this `getComputedAllocationQuantities` with the "C" capitalized?

server/controllers/finance/stepDownCostAllocationQuantity.js
```diff
@@ -186,48 +184,36 @@ async function updateQuantities(req, res, next) {
...
-            await db.exec(updateQRec, [newQuantity, qRecordId[0].id]);
-          } else {
-            // console.log("Inserting: ", cc.id, basis.id, newQuantity);
-            await db.exec(insertQRec, [cc.id, basis.id, newQuantity]);
-          }
-        } else {
-          // TODO:  Do we want to zero out the others here?
-          unUpdatedCenters.push({ cc });
-        }
+        const newQuantity = ccData ? ccData[basis.name] : 0;
```
jniles, line 203: Excellent 👍

server/controllers/finance/stepDownCostAllocationQuantity.js
```diff
@@ -186,48 +184,36 @@ async function updateQuantities(req, res, next) {
...
 
-    // Compute and set each computable allocation basis quantity
-    const unUpdatedCenters = [];
-
     /* eslint-disable no-await-in-loop */
     for (let i = 0; i < computables.length; i++) {
+      const transaction = db.transaction();
       const basis = computables[i];
-      const data = await allocationQuantities(basis.id);
+      const data = await computedAllocationQuantities(basis.id);
```
jniles, line 196: Good rename 👍

jniles (commented):
Looks good to me.  Just a minor tweak to the name.

Go ahead and add in the name tweak you sent me as a patch.  That should work fine.

client/src/modules/cost_center/cost_center.service.js
jmcameron: I reverted to the old name.

NOTE That some function in the cost center service are not being used and should be purged if appropriate:
- getAllocationQuantities
- getAllocationRegistry (its http call is superseded by /cost_center/allocation_bases)

jniles (approved):
Looks good to me!

bors r+

PR #6014, opened 2021-10-12
Add CC link in the reverse voucher transaction

### LINKED ISSUE #6001: Ensure that reversing a transaction maintains the links to the cost centers
In BHIMA, we have a generic "reverse transaction" feature, that flips the debits and credits for a transaction.  We should ensure that the new columns tied to cost centers are also taken into account when flipping the transaction.

mbayopanda:
This PR adds cost_center_id and principal_center_id in the reversed transaction 

image

Close #6001

### DIFF
--- a/server/models/procedures/voucher.sql
+++ b/server/models/procedures/voucher.sql
@@ -106,7 +106,7 @@ invoice.
 Once the procedure has finished, the corresponding cash or invoice record will
 have the "reversed" column set to "1".
 */
-DROP PROCEDURE IF EXISTS ReverseTransaction$$
+DROP PROCEDURE IF EXISTS ReverseTransaction $$
 CREATE PROCEDURE ReverseTransaction(
   IN uuid BINARY(16),
   IN user_id INT,
@@ -160,13 +160,13 @@ BEGIN
     LIMIT 1;
 
   -- NOTE: the debits and credits are swapped on purpose here
-  INSERT INTO voucher_item (uuid, account_id, debit, credit, voucher_uuid, document_uuid, entity_uuid)
-    SELECT HUID(UUID()), zz.account_id, zz.credit_equiv, zz.debit_equiv, voucher_uuid, zz.reference_uuid, zz.entity_uuid
+  INSERT INTO voucher_item (uuid, account_id, debit, credit, voucher_uuid, document_uuid, entity_uuid, cost_center_id, principal_center_id)
+    SELECT HUID(UUID()), zz.account_id, zz.credit_equiv, zz.debit_equiv, voucher_uuid, zz.reference_uuid, zz.entity_uuid, zz.cost_center_id, zz.principal_center_id
     FROM (
-      SELECT pj.account_id, pj.credit_equiv, pj.debit_equiv, pj.reference_uuid, pj.entity_uuid
+      SELECT pj.account_id, pj.credit_equiv, pj.debit_equiv, pj.reference_uuid, pj.entity_uuid, pj.cost_center_id, pj.principal_center_id
       FROM posting_journal AS pj WHERE pj.record_uuid = uuid
       UNION ALL
-      SELECT gl.account_id, gl.credit_equiv, gl.debit_equiv, gl.reference_uuid, gl.entity_uuid
+      SELECT gl.account_id, gl.credit_equiv, gl.debit_equiv, gl.reference_uuid, gl.entity_uuid, gl.cost_center_id, gl.principal_center_id
       FROM general_ledger AS gl WHERE gl.record_uuid = uuid
     ) AS zz;
 


server/models/procedures/voucher.sql
```diff
@@ -105,6 +105,7 @@ invoice.
 Once the procedure has finished, the corresponding cash or invoice record will
 have the "reversed" column set to "1".
 */
+DROP PROCEDURE IF EXISTS ReverseTransaction $$
 CREATE PROCEDURE ReverseTransaction(
```
jmcameron, line 110: Not sure why you are passing in voucher_uuid to this SQL function.  Why not create it here and return the UUID as the result of the SQL procedure.  I see in function reverseTransaction() (server/controllers/finance/journal/index.js line 694) that we create the new voucher uuid there and pass it to the SQL call.  But it is not used in that function.  It is returned as a result of the function.  So it could easily be created by the SQL procedure and returned that way.  It would make the reverseTransaction() function a little simpler.  Not sure if this needs to be fixed but I wondered why we are doing it this way.

jmcameron (commented):
I read through the SQL queries.  The new code seems okay and seems like it should accomplish the desired purpose.

However, I tested it with bhima_test and I think it worked.  I reloaded the bhima_test db and set the cost_center_id, principal_center_id in the voucher_items for the item I was going to change (their default values in the bhima_test db is null).   I reversed the voucher and the new corrected voucher item has the correct cost_center_id and principal_center_id.

I noticed a problem with the currency_id.  You should assign the same currency_id to the copied correction as in the original.  Otherwise you will see mixed currency in the voucher registry:
image

I have two suggestions:
- It would be good to add a regression test for this
- How can we see the cost center associated with the voucher items?  This does not seem to be shown anywhere.   I had to verify they change in SQL.  It seemed like we should be able to see that somewhere in Bhima.

jmcameron:
>     * How can we see the cost center associated with the voucher items?  This does not seem to be shown anywhere.   I had to verify they change in SQL.  It seemed like we should be able to see that somewhere in Bhima.

I had forgotten to enable the cost center column in the Menu > Columns function.
I tried to see the page shown in your PR and I do not see the cost centers for the items I modified.   Even in the Voucher Registry > <item action menu> > Show Transactions (and showed the cost centers).   I verified that the new voucher_item that the reveral created did have the correct cost_center_id, etc.  So I'm confused.

mbayopanda:
> 
> 
> I read through the SQL queries. The new code seems okay and seems like it should accomplish the desired purpose.
> 
> However, I tested it with bhima_test and I think it worked. I reloaded the bhima_test db and set the cost_center_id, principal_center_id in the voucher_items for the item I was going to change (their default values in the bhima_test db is null). I reversed the voucher and the new corrected voucher item has the correct cost_center_id and principal_center_id.
> 
> I noticed a problem with the currency_id. You should assign the same currency_id to the copied correction as in the original. Otherwise you will see mixed currency in the voucher registry: image
> 
> I have two suggestions:
> 
>     * It would be good to add a regression test for this
> 
>     * How can we see the cost center associated with the voucher items?  This does not seem to be shown anywhere.   I had to verify they change in SQL.  It seemed like we should be able to see that somewhere in Bhima.

Good catch 👍🏽 

@jniles you have more experience on the reversing transaction feature, what do you think about the fact of using the transaction's currency in the reversed one ?

mbayopanda:
> 
> 
> > ```
> > * How can we see the cost center associated with the voucher items?  This does not seem to be shown anywhere.   I had to verify they change in SQL.  It seemed like we should be able to see that somewhere in Bhima.
> > ```
> 
> I had forgotten to enable the cost center column in the Menu > Columns function. I tried to see the page shown in your PR and I do not see the cost centers for the items I modified. Even in the Voucher Registry > > Show Transactions (and showed the cost centers). I verified that the new voucher_item that the reveral created did have the correct cost_center_id, etc. So I'm confused.

In the voucher registry you cannot see the `Cost center` or `Principal center` column, if you want that you have to go in the `Journal` and enable these columns

jniles:
@mbayopanda @jmcameron this is a good question.  The fear with using the original currency is that the exchange rate will have changed.  So, here is the decision that we needed to make (and I can't remember what we decided - the code will say):

1. Do we reverse the transaction at the exact moment that it was originally created?  If so, use the original currency, that isn't a problem.  However, if the fiscal year is closed, that could cause an issue.  

2. Or do we reverse the transaction at the current date?  If so, we cannot use the original currency, but instead the _value_ in the enterprise currency.  This makes it record a bit confusing as @jmcameron, but philosophically, it isn't an exchange of money, so that isn't a problem - as long as the value is faithfully represented to be the same.

So, which do we do?  Do we reverse the record at the same timestamp?  If so, (1) can be done.  If not, we must do (2).

server/models/procedures/voucher.sql
jniles: I'm not sure.  I know that our JS code generates UUIDv4s, whereas MySQL generates UUIDv1s.  We do use MySQL UUIDs rarely, but generally prefer the UUIDv4.

It may have originally been for testing - create the voucher with a particular UUID, and then check that it exists.  But I'm not sure.

jniles:
@mbayopanda this looks fine to me.  Can you do a rebase so we can get this merged?

jniles:
LGTM.

Bors r+

PR #6016, opened 2021-10-12
Improve cost center allocation basis table and editor

jmcameron:
Improve the user interface for cost center allocation basis tables and value editor
- Allocation Keys registry: Add units to column headers
- Allocation Keys registry: Handle number of decimal places for various values
- Allocation Keys registry: Show direct cost as currency (currently only in the enterprise currency)
- Allocation basis value editor:  Show units in item labels
- Allocation basis value editor:  When using the [Edit] button, when you select a Cost Center, the known basis values for that cost center are now filled in

**TESTING**
- Use bhima_test with the sample data
- Cost Centers > Allocation Keys
  -  Note the units in some of the header cells
  -  Note appropriate numbers of decimal places (eg, none for whole numbers)
  - Click on the [Edit] button and note the added units labels
    - Select a cost center (eg Daycare Center).  Note the existing values are filled in

This should also be tested with a production database to test the migration file (I did not do that).

### DIFF
--- a/client/src/modules/cost_center/allocation_bases/allocation_bases.js
+++ b/client/src/modules/cost_center/allocation_bases/allocation_bases.js
@@ -7,7 +7,9 @@ CostCenterAllocationBasesController.$inject = [
   'AllocationBasisQuantityService',
   'uiGridConstants',
   'NotifyService',
+  'SessionService',
   '$uibModal',
+  '$translate',
 ];
 
 function CostCenterAllocationBasesController(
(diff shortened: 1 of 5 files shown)


client/src/modules/cost_center/allocation_bases/allocation_bases.js
```diff
@@ -95,17 +101,24 @@ function CostCenterAllocationBasesController(
 
         // Build other columns for ui-grid
         const otherColumns = rows.costCenterIndexes.map((item) => {
+
+          let label = $translate.instant(item.index);
+          if (item.units) {
+            label += ` (${$translate.instant(item.units)})`;
+          }
```
mbayopanda, line 108: 👍🏽

mbayopanda (approved):
LGTM 👍🏽

PR #6017, opened 2021-10-12
feature(Set dynamic Base Index growth rate)

### LINKED ISSUE #6005: Enter the annual rate of increase of the configurable basic vanga indices
During the implementation of the basic indices of Vanga, the rate of increase had been fixed whereas this value should be variable and configurable, it is therefore a question of defining this value as a parameter of the company

lomamech:
- Manual entry of the rate of increase of the base index
- Improvement of the UpdateStaffingIndices () procedure
   by allowing only the selection of employees defined for the period
- Rounding of basic index calculation
closes #6005

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -330,6 +330,7 @@
             "BALANCE_SHEET": "Balance Sheet",
             "BANK_ACCOUNT": "Bank Account",
             "BANK": "Bank",
+            "BASE_INDEX_GROWTH_RATE": "Base index growth rate",
             "BASIC_SALARY": "Basic Salary",
             "BEGINNING_VIEW_BALANCE" : "Beginning View Balance",
             "BILLING_DATE": "Billing Date",
--- a/client/src/modules/enterprises/enterprises.html
+++ b/client/src/modules/enterprises/enterprises.html
@@ -299,6 +299,9 @@
                   <input class="form-control" value="0" min="0" type="number" name="base_index_growth_rate" ng-model="EnterpriseCtrl.enterprise.settings.base_index_growth_rate" required>
                   <span class="input-group-addon"> % </span>
                 </div>
+                <span class="help-block" translate>
+                  SETTINGS.BASE_INDEX_GROWTH_RATE_HELP_TEXT
+                </span>                
                 <div class="help-block" ng-messages="EnterpriseForm.base_index_growth_rate.$error" ng-show="EnterpriseForm.$submitted">
                   <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
                 </div>
--- a/client/src/modules/enterprises/enterprises.html
+++ b/client/src/modules/enterprises/enterprises.html
@@ -291,6 +291,18 @@
                 on-change-callback="EnterpriseCtrl.enableIndexPaymentSetting(value)">
               </bh-yes-no-radios>
 
+              <div class="form-group" ng-class="{ 'has-error' : EnterpriseForm.$submitted && EnterpriseForm.base_index_growth_rate.$invalid }">
+                <label class="control-label" translate>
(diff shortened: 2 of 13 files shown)


lomamech:
@mbayopanda @jmcameron I need a review for this PR

client/src/i18n/en/form.json
```diff
@@ -329,6 +329,7 @@
             "BALANCE_SHEET": "Balance Sheet",
             "BANK_ACCOUNT": "Bank Account",
             "BANK": "Bank",
+            "BASE_INDEX_GROWTH_RATE": "Base index growth rate",
```
jmcameron, line 333: What are the units for this?   %/year?

client/src/modules/enterprises/enterprises.html
```diff
@@ -291,6 +291,18 @@
                 on-change-callback="EnterpriseCtrl.enableIndexPaymentSetting(value)">
               </bh-yes-no-radios>
 
+              <div class="form-group" ng-class="{ 'has-error' : EnterpriseForm.$submitted && EnterpriseForm.base_index_growth_rate.$invalid }">
+                <label class="control-label" translate>
+                  FORM.LABELS.BASE_INDEX_GROWTH_RATE
+                </label>
```
jmcameron, line 297: I would really like to see a description of what this option is how it is used in the system (a _HELP_TEXT msg) like on the previous settings.

jmcameron (commented):
As I indicated in the code, I would like the enterprises settings page to have an exaplanation of what this new setting is and how it is used.    It is not obvious to me that this new setting is being used anywhere currently (must have missed it in the code).

I tested this with bhima_test and the user interface works for setting the value.

client/src/i18n/en/form.json
lomamech: Yes

lomamech: Yes

jmcameron: I suggest you include that in the label.  It see in the gui it has '%'.  Could this be update to be '%/year' or '%/an' with language tokens.

lomamech: Okay :white_check_mark:

client/src/modules/enterprises/enterprises.html
lomamech: :white_check_mark:

jniles (commented):
@lomamech, this PR looks good to me.  As @jmcameron noted, having the units would be nice - it isn't clear to me that the index grows annually (instead of monthly, etc).

At some point, we might want to make a "Payroll Settings" page, like we do for "Stock Settings".  That shouldn't be in this PR though.

lomamech:
Usually it's after each year, it's sort of a bonus for one year of seniority

jniles:
@lomamech right - add that to the text.  A user will never know that otherwise.

lomamech:
@jniles @jmcameron I think I ended up with this pull request, in my next request request I think adding the option Payroll Setting and I will need some elements of this Pull Request

lomamech:
@jniles @jmcameron I think I ended up with this pull request, in my next request request I think adding the option Payroll Setting and I will need some elements of this Pull Request

lomamech:
@jniles @jmcameron I think I ended up with this pull request, in my next request request I think adding the option Payroll Setting and I will need some elements of this Pull Request

jmcameron:
> @jniles @jmcameron I think I ended up with this pull request, in my next request request I think adding the option Payroll Setting and I will need some elements of this Pull Request

Sounds good.  merge your new PR with the old PR and when the new PR is successful close the old one.  In the meantime, I suggest marking your old PR as "do not merge".

jniles (approved):
LGTM.

bors r+

PR #6020, opened 2021-10-13
Filter journal by account type

mbayopanda:
This PR adds the option to filter journal also by account type

image

### DIFF
--- a/client/src/js/components/bhAccountTypeSelect/bhAccountTypeSelect.html
+++ b/client/src/js/components/bhAccountTypeSelect/bhAccountTypeSelect.html
@@ -1,4 +1,4 @@
-<div ng-form="AccountTypeForm" bh-cost-center-select ng-model-options="{ updateOn: 'default' }">
+<div ng-form="AccountTypeForm" bh-account-type-select ng-model-options="{ updateOn: 'default' }">
   <div
     class="form-group"
     ng-class="{ 'has-error' : AccountTypeForm.$submitted && AccountTypeForm.accountTypeId.$invalid }">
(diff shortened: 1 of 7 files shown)


client/src/js/components/bhAccountTypeSelect/bhAccountTypeSelect.html
```diff
@@ -0,0 +1,35 @@
+<div ng-form="AccountTypeForm" bh-cost-center-select ng-model-options="{ updateOn: 'default' }">
```
jmcameron: Thanks for adding a component for this!

jmcameron (commented):
The code looks good.
I tested with bhima_test and the user interface, selection, etc worked fine.  The result seems to fine too.  However I was confused.  I search for "Income" accounts.  I would have expected income to be all credit (or debit?) but it looks mixed!
image

jniles (approved):
LGTM.  No comments on my end.

jmcameron:
I'm still a bit confused as to why the "Income" account has both credits and debits.    I realize that reversing a voucher could do that, but that does not seem to be the case here.  Try this with bhima_test (I also included the sample data, although it probably does not matter).    Notice the first two entries in the Journal after filtering for "Income".    They appear to be the same kind of transaction to me.  And the records (VO.TPA.2 and VO.TPB.1) seem to both describe similar simple stock purchases.

I have not problem with accepting this PR as it is since I'm pretty sure the changes in this PR are fine.   My question here is probably related to my understanding -- and if there is a problem, it can be fixed in a separate issue.

jniles:
@jmcameron I'm pretty sure this is a problem in the test data.  Do you want to try and track it down?  You are right, it is nonsensical... but a lot of our test data is.

jmcameron:
> @jmcameron I'm pretty sure this is a problem in the test data. Do you want to try and track it down? You are right, it is nonsensical... but a lot of our test data is.

Ok, I'll fix it in a separate PR.

PR #6024, opened 2021-10-14
Add display and editing of cost center to service registry

### LINKED ISSUE #5995: Add filter to services registry to show services not associated with a cost center
This would be helpful for review as sites are being prepared for cost center analysis identify services that need to be associated with a cost center

jmcameron:
This PR adds a column for the cost center to the Service registry.  It also adds support for selecting/changing the cost center when editing/creating services.

Although I'm closing issue 5995 (adding a filter for services without cost centers), I feel like this PR adequately addresses that issue.  To see the services without cost centers, simply sort on the Cost Centers column (click header cell twice) and all the lines with blanks instead of cost center names are the ones that do not have cost centers.

**CHANGED:**  Now this requires selecting a cost center when creating or editing a service

Closes https://github.com/IMA-WorldHealth/bhima/issues/5995

**TESTING**
- Use any database.  It might be easier with bhima_test since it has services without cost centers
- If you use bhima_test, load the sample cost center data
- Go to:  Administration > Services  (the services registry)
- Try sorting by Cost Center to easily group services with not cost centers
- Test these 5 combinations
   - --Create a service without specifying a cost center--
   - Create a service and specify a cost center
   - Edit a service that does not have a cost center, change the project,  and try to save.  The form should force you select a cost center.
   - Edit a service with a cost center, change the project, and save.  The updated service should keep the old cost center value
   - Edit a service with a cost center and change the cost center.
   - NOTE: I tested the migration with the imck database.  It might be good to try it with another production database.  Note that updating the constraints to 'ON DELETE CASCADE' turned out to be tedious and tricky!

**Also check that the new 'ON DELETE CASCADE' for the 'service_cost_center' table is working:**
- Create two test cost centers (do not select services)
- Create a new service and select one of the new cost centers
- Create a second new service and select the other cost center
- Delete one of the new services.
- Go to the cost centers registry and verify that the link is gone for the service just deleted
- In the cost centers registry, delete the other new cost center
- Go back to the services registry and verify that the corresponding service is not connected to a cost center.
- You might want to check the service_cost_center table in SQL to verify the changes.

### DIFF
--- a/client/src/modules/services/modals/service.modal.html
+++ b/client/src/modules/services/modals/service.modal.html
@@ -21,6 +21,13 @@
       </div>
     </div>
 
+    <bh-cost-center-select
+      label="COST_CENTER.SELECT_COST_CENTER"
+      cost-center-id="ServiceModalCtrl.service.cost_center_id"
+      on-select-callback="ServiceModalCtrl.onCostCenterSelect(costCenter)"
+      required="true">
+    </bh-cost-center-select>
+
     <bh-project-select
       project-id="ServiceModalCtrl.service.project_id"
       on-select-callback="ServiceModalCtrl.onSelectProject(project)"
--- a/server/controllers/admin/services.js
+++ b/server/controllers/admin/services.js
@@ -35,10 +35,14 @@ function list(req, res, next) {
   if (req.query.full === '1') {
     sql = `
       SELECT s.name, s.enterprise_id, BUID(s.uuid) AS uuid, s.hidden,
-      e.name AS enterprise_name, e.abbr, p.id AS project_id, p.name AS project_name
+      e.name AS enterprise_name, e.abbr, p.id AS project_id, p.name AS project_name,
+      cc.id AS cost_center_id, cc.label AS cost_center_name
       FROM service AS s
       JOIN enterprise AS e ON s.enterprise_id = e.id
-      LEFT JOIN project AS p ON s.project_id = p.id`;
+      LEFT JOIN project AS p ON s.project_id = p.id
+      LEFT JOIN service_cost_center AS scc ON scc.service_uuid = s.uuid
+      LEFT JOIN cost_center AS cc ON cc.id = scc.cost_center_id
+      `;
   }
 
   const params = db.convert(req.query, ['uuid']);
@@ -83,6 +87,10 @@ function countServiceByProject(req, res, next) {
  */
 function create(req, res, next) {
   const record = req.body;
+  const costCenterId = record.cost_center_id;
+  delete record.cost_center_id;
+  delete record.cost_center_name;
+
   const sql = `INSERT INTO service SET ?`;
 
   // add contextual information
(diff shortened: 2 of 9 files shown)


server/controllers/admin/services.js
```diff
@@ -114,16 +128,25 @@ function create(req, res, next) {
...
+
+  const sql = `UPDATE service SET ? WHERE uuid = ?;`;
+
+  db.exec(sql, [queryData, serviceUuid])
     .then((result) => {
       if (!result.affectedRows) {
         throw new NotFound(`Could not find a service with uuid ${req.params.uuid}.`);
       }
-
+      if (costCenterId) {
```
jniles, line 144: This is really elegant.  Nice job!

server/controllers/admin/services.js
```diff
@@ -208,6 +231,27 @@ function lookupService(uid) {
...
+ * @returns {Promise} - returns the result of the database query
+ */
+function setServiceCostCenter(serviceUuid, costCenterId) {
+  const delOld = 'DELETE FROM service_cost_center WHERE service_uuid = ?';
+  const addNew = 'INSERT INTO service_cost_center (service_uuid, cost_center_id) VALUES (?, ?)';
+
+  const transaction = db.transaction();
+  transaction.addQuery(delOld, [serviceUuid]); // Always delete any old version
+  transaction.addQuery(addNew, [serviceUuid, costCenterId]);
+  return transaction.execute();
```
jniles: Just an FYI - you can do this in a fluent API, if that tickles your fancy.

```js
return db.transaction()
  .addQuery(delOld, [serviceUuid]) // Always delete any old version
  .addQuery(addNew, [serviceUuid, costCenterId])
  .execute();

File: client/src/modules/services/modals/service.modal.html [hunk 3]
```diff
@@ -21,6 +21,12 @@
       </div>
     </div>
 
+    <bh-cost-center-select
```
jniles, line 24: Is it possible to make this required?

image

jniles (changes requested):
This is nearly good.  I have two requests:

1. Can you make the cost center field required?
2. We currently cannot delete a service because it is used in the cost_center_service table.  Can you make deleting the service clean up that record?  You can either CASCADE in SQL or you can do it in JS.

mbayopanda (approved):
If the cost center is not required you can add away to clean the cost center field in case we want to set a service free of cost center; otherwise adding a required property as @jniles mentioned that will be good.

jniles:
If cost centers are not required on services, we'll need a way to unassociate the cost center, with something like a `clear()` button.  Personally, I think we should just go all in on cost centers. and make them required.

server/controllers/admin/services.js
jmcameron: 😄

server/controllers/admin/services.js
jmcameron: Done

client/src/modules/services/modals/service.modal.html
jmcameron: Per @jniles suggestion, I made this required.

jmcameron:
@jniles Note that I updated the testing directions and info at the top.

jniles (approved):
All my issues have been addressed.  Thanks!

bors r+

Issue #6023, opened 2021-10-14
Cost center aggregate table needs to distinguish costs and profits


mbayopanda:
This issue is related to the fact that our `cost_center_aggregate` table aggregates all data from the general_ledger based on their `cost_center_id` and `principal_center_id` without considering if these data are for expenses or incomes.

Assume that we have a cost_center : Administration which have values for expense (Class 6) and incomes (Class 7), in the `cost_center_aggregate` table we put just together all these data.

The proposal is to have a way to distinguish expenses and incomes in the `cost_center_aggregate` table by adding maybe a new column for that.

jniles:
Yeah, I noticed this as well.  I hope we can have as elegant a solution as just an extra column in the cost_center_aggregates table.  We'll need to then update the procedure that creates this table.

jniles:
Closed in #6026.

PR #6026, opened 2021-10-15
Distinguish income from expense in cost center aggregate

mbayopanda:
This PR is related to #6023 and refactors some logic while having distinction between income and expense in the `cost_center_aggregate` table.

In this PR: 
- We have changed a bit the logic of `include_revenu` of  `Cost Center Allocation Report` 
- We adds the `is_income` column in the `cost_center_aggregate` table
- Update the procedure `zRecalculateCostCenterAggregates` to fill also the `is_income` column
- Update the procedure `ComputeCostCenterAllocationByIndex` to handle `is_income` column also, and fix data aggregation query

Benefit:
- Journal data are the same with data produced by the `Cost Center Allocation Report` (We have a way to compare data)

Todo:
- The client side of the `Cost Center Allocation Report` is still raw but we have to enhance that and fix some typos

Question:
- How does the `Cost Center Allocation Report` must be presented for revenue?
- Does for revenue also we can use the step-down algo? (with this PR we did that, you can explore that)


client/src/i18n/en/cost_center.json
```diff
@@ -28,7 +28,9 @@
     "UPDATE_ALLOCATION_QUANTITIES_TOOLTIP": "Update system-computed cost center allocation quantities",
     "SET_ALLOCATION_QUANTITIES_TOOLTIP": "Define cost center allocation quantities",
     "INCLUDE_REVENUE_CENTERS" : "Include revenue with costs?",
-    "INCLUDE_REVENUE_CENTERS_HELP_TEXT" : "This will include the revenue generated by the revenue centers in addition to their costs."
+    "INCLUDE_REVENUE_CENTERS_HELP_TEXT" : "This will include the revenue generated by the revenue centers in addition to their costs.",
```
jmcameron, line 32: I'm confused:  This is not currently used anywhere.   It seems like currently USE_REVENUE_CENTER* is used but not INCLUDE_REVENUE_CENTER*.  If INCLUDE_REVENUE_CENTER* are obselete, lets remove them.

server/models/admin.sql
```diff
@@ -272,11 +272,18 @@ BEGIN
...
-    WHERE cost_center_id IS NOT NULL
-    GROUP BY cost_center_id, principal_center_id, period_id;
+  INSERT INTO cost_center_aggregate (period_id, credit, debit, cost_center_id, principal_center_id, is_income)
+    SELECT 
+      gl.period_id, 
+      SUM(gl.credit_equiv) AS credit, 
+      SUM(gl.debit_equiv) AS debit, 
+      gl.cost_center_id, 
+      gl.principal_center_id,
+      IF(a.type_id = 4, 1, 0) AS is_income
```
jmcameron, line 282: I did not realize you could do this.  I've been doing:   if (a.type_id IN (4, 1, 0))...

jmcameron (commented):
Looks okay to me.   I followed most of the code/SQL changes.
I tested this with bhima_test and it seemed okay.  I looked at the new values in the cost_center_aggregate table and see the new column  (is_income) and it looks correct (although I could not be sure because there were no non-is_income entries.

mbayopanda:
> 
> 
> Looks okay to me. I followed most of the code/SQL changes. I tested this with bhima_test and it seemed okay. I looked at the new values in the cost_center_aggregate table and see the new column (is_income) and it looks correct (although I could not be sure because there were no non-is_income entries.

Thank you @jmcameron for the review, with a production database you will see well the impact of `is_income` column in the `Cost Center Allocation Report`

jniles (approved):
Looks good to me.  As discussed offline, we need to make these both visible at the same time, in two different tables, as requested by Dr. Larry.

bors r+

jniles:
@mbayopanda asked me to cancel this to fix up some typos.

mbayopanda:
@jniles typo changes landed, for the expense and profit in the same report I suggest to do another PR for that

jniles:
Bors r+

PR #6029, opened 2021-10-15
bug(Employee Advantage and Payroll Indice)

### LINKED ISSUE #6028: Improvement of the payroll module with index
We noticed that in several registers the data is not displayed in alphabetical order,

We also noticed that the monetary data is not very well rounded against the company currency.

We also noticed a Bug when saving data resulting from the Payroll process with index to the normal payroll process

lomamech:
- Sort rubrics configuration
- Sort functions configuration registry
- Fix rubrics map when monetary_value is empty
- Fix error in configuration of payroll for the employee_advantage
- Using min_monentary_unit for Round monetary value

closes #6028


jniles (approved):
Changes look good to me.  Good improvements!

bors r+

Issue #6027, opened 2021-10-15
Optimize /stock/inventories/depots route


jniles:
We need to optimize the `/stock/inventories/depots` route.  It is slowing down our production systems.

```sh
127.0.0.1 - GET /stock/inventories/depots?dateTo=2021-10-15T09:40:48.096Z&depot_uuid=05121CFE23CA4E8780484F231F5DA590&is_expired=0 HTTP/1.1 200 366158 - 20228.733 ms
127.0.0.1 - GET /stock/inventories/depots?dateTo=2021-10-15T09:44:10.451Z&depot_uuid=05121CFE23CA4E8780484F231F5DA590&is_expired=0 HTTP/1.1 200 366166 - 10316.522 ms
```

I believe all this does is load the available stock in a depot at a given date.  This could use the `stock_movement_status` table to speed up performance.

PR #6031, opened 2021-10-15
perf(stock): add skip tags functionality

jniles:
Ensures that routes which do not need to query lot tags can opt out to give a little performance gain.

Related to #6027.

### DIFF
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -889,8 +889,9 @@ async function listLotsDepot(req, res, next) {
   try {
     const data = await core.getLotsDepot(null, params);
 
-    // if we have an empty set, do not query tags.
-    if (data.length !== 0) {
+    // if no data is returned or if the skipTags flag is set, we don't need to do any processing
+    // of tags.  Skip the SQL query and JS loops.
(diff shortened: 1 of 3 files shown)


server/controllers/stock/index.js
```diff
@@ -889,8 +889,8 @@ async function listLotsDepot(req, res, next) {
   try {
     const data = await core.getLotsDepot(null, params);
 
-    // if we have an empty set, do not query tags.
-    if (data.length !== 0) {
+    // if we have an empty set, do not query tags.  Alternatively, if skipTags is set
```
jmcameron: I found this comment confusing

jmcameron (approved):
Looks good.  The code seems correct.  
I tested it and it seems to work; I was able to filter the inventory registry (like before) but other uses of the inventory (Stock Exit) seemed to work fine.

I am curious how much a speedup this update achieved.

jniles:
Hard to say exactly in terms of timing - I don't have standard hardware to measure on.  What it does to is save an SQL query, but also a for loop, and that is good.

Yeah, that comment was a half-complete thought clearly.  Let me reword it.

PR #6032, opened 2021-10-15
perf(stock): improve bhStockExpired perf

jniles:
Improves the performance of the `bhStockExpired` performance by using a separate route for the `bhStockExpired` component to load its data without as many JOINs.


jmcameron (approved):
The code looks good.  
I did very simple tests with bhima_test.  The list of expired stock came up on the Stock exit page.
Looks good to me!

PR #6037, opened 2021-10-18
Added allocation basis to each auxiliary center in step-down cost analysis

### LINKED ISSUE #6033: Indicate allocation basis on step down report
Here is an example of the Step Down Report:
image

It isn't possible from this report to know what the allocation basis was for administration.  We should add that information somewhere so that it becomes obvious when the report is printed, how it was generated.

jmcameron:
Updated the step-down cost center allocation report to include a new column indicated which allocation basis is used for each auxiliary center.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6033

Tested with bhima_test with sample cost allocation data.  
It would be useful to test this PR with the Vanga data.

### DIFF
--- a/server/controllers/finance/cost_center_allocation_registry.js
+++ b/server/controllers/finance/cost_center_allocation_registry.js
@@ -90,7 +90,9 @@ async function fetch(session, params) {
     const row = {
       name : ei.cost_center_label,
       principal : ei.principal,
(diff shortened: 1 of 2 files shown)


jniles (approved):
Looks good to me.

image

bors r+

PR #6040, opened 2021-10-19
Add DROPs before all SQL CREATE PROCEDURE/FUNCTION definitions

### LINKED ISSUE #6036: Add DROP ... to the server/models/procedures/*.sql files
Occasionally, we need to reload one of the procedures *.sql files when dealing with updated or adapted databases.   Many of the SQL files in server/models/procedures do have have a "DROP FUNCTION IF EXISTS..." or "DROP PROCEDURE IF EXISTS ..." before defining their functions or procedures.   We should add them in.    Basically all SQL files that define Bhima functions or procedures should be 'idempotent', in other words we should able to load them multiple times without encountering errors.

jmcameron:
Update the Bhima SQL procedure files to make sure they can be reloaded without causing errors or warnings.  This is useful in at least two cases to avoid problems when reloading procedure files: (1) while creating new procedures/functions or modifying old procedures/functions, and (2) when  dealing with migrations, sometimes it is necessary to reload certain procedure files to update procedure/function definitions.

**Adds** 
- DROP PROCEDURE IF EXISTS before all CREATE PROCEDURE ...
- DROP FUNCTION IF EXISTS before all CREATE FUNCTION ...

Closes https://github.com/IMA-WorldHealth/bhima/issues/6036

**TESTING**
- Suggest testing with bhima_test
- Run:  yarn; yarn build
- Make sure unit and integration tests run okay
- Download this script: [stest.txt](https://github.com/IMA-WorldHealth/bhima/files/7371767/stest.txt)
- Put stest.txt into the top level of your Bhima installation
- Edit stest.txt to add in bhima_test db password
- Execute it:  bash stest.txt
- You should see complaints about using the db password on the command line, BUT NO OTHER errors or warnings
- Run the tests again.  They should pass like before
-


jniles (approved):
Looks good to me.  Tested and working.

bors r+

PR #6041, opened 2021-10-19
Cost center value by accounts

### LINKED ISSUE #6000: Add the ability to view account balances tied to a cost center
Since all our cost centers are tagged in the journal, we should be able to create a report that shows the values of the cost centers broken out by the different accounts that were hit.  For example, the "Surgery" cost center may actually have direct costs coming from maintenance, medicines, personnel taxes, etc, and be hitting all these different income/expense accounts.  We should present a report that allows a user to choose a cost center, and show all the accounts that were hit by that cost center.

mbayopanda:
This PR adds a new report that show cost center value by accounts

image

image

Close #6000


jmcameron:
Would it make sense to add an extra column for the type of account?

server/controllers/finance/reports/cost_center_step_down/accounts_report.js
```diff
@@ -0,0 +1,114 @@
...
+      JOIN account a ON a.id = gl.account_id 
+      JOIN cost_center cc ON cc.id = gl.principal_center_id
+      WHERE gl.trans_date >= DATE(?) AND gl.trans_date <= DATE(?) AND a.type_id = ?
+        AND gl.principal_center_id = ?
+      UNION ALL
+      SELECT cc.label, gl.debit_equiv , gl.credit_equiv , gl.cost_center_id AS ccId, gl.account_id 
+      FROM general_ledger gl 
+      JOIN cost_center cc ON cc.id = gl.cost_center_id AND gl.principal_center_id IS NULL 
+      JOIN account a ON a.id = gl.account_id 
+      WHERE gl.trans_date >= DATE(?) AND gl.trans_date <= DATE(?) AND a.type_id = ?
```
jniles: Would it be possible to use the `period_id` instead of the `trans_date`?  It should be faster since you are just comparing integers.

mbayopanda:
> 
> 
> Would it make sense to add an extra column for the type of account?

Excellent question, for this report we are looking only for expense (there is a restriction) but we can enhance the report to consider both income and expense accounts

mbayopanda:
@jmcameron here are changes I added

image

image

server/controllers/finance/currencies.js
```diff
@@ -47,3 +49,37 @@ exports.detail = function detail(req, res, next) {
     .catch(next)
     .done();
 };
+
+/** get currencies information related to exchange rate */
+exports.metadata = async (session, params) => {
```
jniles: Can you give this a more explicit name and parameters?  Maybe something like `getExchangeInformationForReports()` or something.

I would also remove the logic related to periods from this function.  I think it could be a really simple signature: `getExchangeInformationForReports(enterpriseId, currencyId, date)`.

server/controllers/finance/reports/cost_center_step_down/accounts_report.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+async function buildAccountsReport(params, session) {
+  const options = _.extend(params, {
+    filename : 'TREE.COST_CENTER_STEPDOWN',
+    csvKey : 'rows',
+    user : session.user,
+  });
+
+  // Account type 5 is for expense
+  const ACCOUNT_TYPE = +options.include_revenue ? [4, 5] : [5];
```
jniles: Why make this in all capital letters?  Also, since it is an array, let's use a plural word - accountTypes.

server/controllers/finance/reports/cost_center_step_down/accounts_report.js
```diff
@@ -0,0 +1,116 @@
...
+  const metadata = await currencies.metadata(session, params);
+
+  const {
+    enterpriseId,
+    firstCurrency,
+    secondCurrency,
+    lastRateUsed,
+  } = metadata;
+
+  const { dateFrom, dateTo } = metadata.range;
```
jniles: Can you just use the periods directly?  The posting_journal/general_ledger have a period column.  It's probably faster to use that and saves you an SQL query.

server/controllers/finance/reports/cost_center_step_down/accounts_report.js
```diff
@@ -0,0 +1,116 @@
...
+        AND gl.cost_center_id = ?
+    )z on z.account_id = a.id 
+`;
+
+  const query = `
+    ${queryTemplate} GROUP BY z.ccId, a.id ORDER BY a.number;
+  `;
+
+  const queryTotals = `
+    ${queryTemplate};
```
jniles: Why repeat yourself?  You can just name `queryTemplate` to `queryTotals` from the beginning and you are done.

jniles (commented):
Some clean up comments, mostly related to naming things.  I haven't tested the report itself yet.

server/controllers/finance/reports/cost_center_step_down/accounts_report.js
mbayopanda: I have used `period_id` in the query for optimization, but we cannot be free for this SQL query of date range from period since we have to get these information (dates) in the currencies section of the report 

image

server/controllers/finance/reports/cost_center_step_down/report.handlebars
```diff
@@ -2,7 +2,7 @@
 <html>
   {{> head title="REPORT.COST_CENTER_STEPDOWN.TITLE"}}
   <body>
-  <div class="fluid container">
+  <div class="fluid-container">
```
jniles, line 5: :+1:

server/controllers/finance/reports/cost_center_step_down/accounts_report.js
```diff
@@ -0,0 +1,113 @@
...
+      SUM(z.debit_equiv * IFNULL(GetExchangeRate(?, ?, ?), 1)) debit,
+      SUM(z.credit_equiv * IFNULL(GetExchangeRate(?, ?, ?), 1)) credit,
+      SUM((z.debit_equiv - z.credit_equiv) * IFNULL(GetExchangeRate(?, ?, ?), 1)) solde
+    FROM account a 
+    JOIN account_type at ON at.id = a.type_id
+    JOIN (
+      SELECT cc.label, gl.debit_equiv , gl.credit_equiv , gl.principal_center_id AS ccId, gl.account_id 
+      FROM general_ledger gl 
+      JOIN account a ON a.id = gl.account_id 
+      JOIN cost_center cc ON cc.id = gl.principal_center_id
```
jniles, line 61: I just tested this on a production database, and the performance is pretty poor.

Can we remove the JOIN on cost center from this inner query and move it to the outside query?  We don't need it here.

jniles (approved):
I left a comment, but I'm approving this.  If we adopt https://github.com/IMA-WorldHealth/bhima/issues/6043, we'll rework the query to speed this up greatly.

image

PR #6045, opened 2021-10-20
chore(cost centers): remove principal_center_id

### LINKED ISSUE #6043: Consider removing principal_center_id from all tables
I'm not sure we need both `cost_center_id` and `principal_center_id` from tables.  I propose we remove `principal_center_id`.

Here is the rational:

1. Dr. Larry believes that we should have "principal" costs centers configurable when running the report.  The issue for that is here: https://github.com/IMA-WorldHealth/bhima/issues/5994.  Thus, the distinction between principal/auxiliary isn't necessarily straightforward.
2. We've determined that payroll, medicine consumption, and invoicing are all direct costs.  Any other information, we can recover from the accounts hit (INSS, etc) by looking at the report implemented in https://github.com/IMA-WorldHealth/bhima/pull/6041.
3. All our queries would be simplified, removing UNIONs and a few custom functions/stored procedures.  It's best to do this before we deploy cost center analysis.

For these reasons, I propose that we just remove the `principal_center_id` and keep only the `cost_center_id` column.

jniles:
Removes the principal_center_id from the code and tests pass.  I've also run the migration script on both the IMCK and Vanga databases and confirmed they pass as well.

Closes #6043.


mbayopanda (approved):
This is a big refactoring 👍🏽 

I have tested with the test and the production database and we still have the same values as with `principal_center_id` 👍🏽

PR #6050, opened 2021-10-21
feat(journal): trial balance checks on cost centers

### LINKED ISSUE #6009: Detect transactions hitting income/expense accounts not associated with a cost center in Trial Balance
The Trial Balance is the last set of checks before a user commits to posting data into the general ledger.  We should add in a check that sees if there are transactions that hit income/expense accounts that are not associated with a cost center.  If an institution is running with cost center analysis, we'll want to ensure that all their transactions hitting income/expense accounts are associated with a cost center before posting.

jniles:
Implements trial balance checks on the cost centers  The following checks are implemented:

1. Every line in a transaction that concerns an income/expense account requires a cost center.
2. Every line in a transaction that has a cost center id should be an income/expense account.

Closes #6009.

image


mbayopanda (approved):
LGTM 👍🏽 

It works as expected

PR #6051, opened 2021-10-21
Added checking of unit prices on stock entry

### LINKED ISSUE #6052: Add sanity checks for stock article unit costs in
Add sanity checks in the creation of purchase order creation to make sure that unit prices in new purchase orders make sense compared to past median unit costs.

For reference, see the added tests in Stock Entry in [PR 6051](https://github.com/IMA-WorldHealth/bhima/pull/6051).

jmcameron:
One of the problems of dealing with the stock is dealing with inventory article units correctly during data entry.  It easy to make a mistake if grams are mistaken for kilograms, etc.   This PR adds checking during the data entry to make sure the the specified unit price makes sense.    Currently it will complain if the proposed unit cost is less than 50% of the median purchase unit price or more than twice the median purchase price.

This PR was also merged with a different branch that does very similar sanity checking while creating purchase orders.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6052

This PR also does some minor improvements to the purchase price analysis report.  It adds the median unit cost in the summary area at the bottom of the report.  It also fixes an error in the constants for the underlying SQL query.  (3 was used for flux FROM_INTEGRATION instead of the correct 13; Now this SQL query uses the server-side constants file for those values).

**TESTING**
- Suggest using the Vanga dataset
- Select inventory article with a history of purchases
   - Suggest:  107-094 Eau Distillée
   - Try its purchase price analysis (in the action menu).  Note the median unit cost at the summary at the bottom.
- Create a purchase order with this inventory article.  Use a unit cost similar to the median cost but not the same value
- In the purchase registry, change the status of the purchase order to "Confirmed"
- Go to the Stock Entry page
- Select a depot that supports stock entry from Purchase
- Select from Purchase and select the purchase order you just created
- Click on the "Lots" indicator and chose a lot
- On the unit cost line try entering a very small number.  When you exit the entry cell, you should receive a warning dialog pop-up.   If you Confirm, the new unit cost will be accepted.  If you Cancel, the previous unit costs will be restored and a warning will be added at the bottom of the modal.
- Try the same thing with a unit price more than double the median cost.

- Try a similar sequence where you try creating a purchase order with too large or too small a value.  
- Also try creating a purchase order for an inventory article with no prior purchase history, confirm the purchase, and try stock entry for it (high and low).


client/src/modules/inventory/inventory.service.js
```diff
@@ -195,6 +195,11 @@ function InventoryService(
     return $httpParamSerializer(options);
   };
 
+  service.getInventoryUnitCosts = function getInventoryUnitCosts(uuid) {
+    return service.$http.get(`/inventory/unit_cost/${uuid}`)
```
jniles: I propose you flip this URL around to make it more "REST-like": `/inventory/:uuid/unit_cost`.

client/src/modules/purchases/create/createUpdate.js
```diff
@@ -70,9 +70,10 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Notify,
     cellTemplate : 'modules/purchases/create/templates/quantity.tmpl.html',
   }, {
     field : 'unit_price',
-    width : 100,
-    displayName : 'TABLE.COLUMNS.PURCHASE_PRICE',
+    width : 105,
+    displayName : 'TABLE.COLUMNS.PURCHASE_UNIT_PRICE',
```
jniles, line 78: Great change 👍

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -267,6 +278,35 @@ function StockDefineLotsModalController(
   }
 
   function onChangeUnitCost() {
+    // Sanity check on new unit cost
+    const medianUnitCost = Number(vm.stockLine.stats.median_unit_cost);
+    const prevUnitCost = Number(vm.stockLine.prev_unit_cost);
+    const newUnitCost = Number(vm.stockLine.unit_cost);
+    const allowableMinChange = 0.5; // Warn about entries that are less than half the previous unit cost
+    const allowableMaxChange = 2.0; // Warn entries that more than double the previous unit cost
+    if ((newUnitCost < medianUnitCost * allowableMinChange)
```
jniles: Not to be picky, but this is hard to read.  I propose you break this into two different variables, like this:
```js
const lowerCostBound = (medianUnitCost * allowableMinChange);
const upperCostBound = (medianUnitCost * allowableMaxChange);
if (newUnitCost < lowerCostBound || newUnitCost > upperCostBound) { /* etc... */ }
```

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -267,6 +278,35 @@ function StockDefineLotsModalController(
...
+      || (newUnitCost > medianUnitCost * allowableMaxChange)) {
+      const msgParams = { newUnitCost, medianUnitCost, curr : vm.currency.symbol };
+      const errMsg1 = newUnitCost > medianUnitCost * allowableMaxChange
+        ? $translate.instant('WARNINGS.WARN_UNIT_COST_PROBLEM_TOO_HIGH', msgParams)
+        : $translate.instant('WARNINGS.WARN_UNIT_COST_PROBLEM_TOO_LOW', msgParams);
+      let errMsg = errMsg1.concat($translate.instant('WARNINGS.WARN_UNIT_COST_REASON'));
```
jniles: You can just chain these two:
```js
const errMsg  = errMsg1
  .concat($translate.instant('WARNINGS.WARN_UNIT_COST_REASON'))
  .concat($translate.instant('WARNINGS.WARN_UNIT_COST_CONFIRM'));
```

That way, you aren't mixing `concat()` and `+=`.

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -267,6 +278,35 @@ function StockDefineLotsModalController(
...
+    // Sanity check on new unit cost
+    const medianUnitCost = Number(vm.stockLine.stats.median_unit_cost);
+    const prevUnitCost = Number(vm.stockLine.prev_unit_cost);
+    const newUnitCost = Number(vm.stockLine.unit_cost);
+    const allowableMinChange = 0.5; // Warn about entries that are less than half the previous unit cost
+    const allowableMaxChange = 2.0; // Warn entries that more than double the previous unit cost
+    if ((newUnitCost < medianUnitCost * allowableMinChange)
+      || (newUnitCost > medianUnitCost * allowableMaxChange)) {
+      const msgParams = { newUnitCost, medianUnitCost, curr : vm.currency.symbol };
+      const errMsg1 = newUnitCost > medianUnitCost * allowableMaxChange
```
jniles: You can actually condense this a bit more and it might be more clear:
```js
const msgKey = (newUnitCost > medianUnitCost * allowableMaxChange)
  ? 'WARNINGS.WARN_UNIT_COST_PROBLEM_TOO_HIGH'
  : 'WARNINGS.WARN_UNIT_COST_PROBLEM_TOO_LOW';
  
const errMsg = $translate.instant(msgKey, msgParams)
  .concat($translate.instant('WARNINGS.WARN_UNIT_COST_REASON'))
  .concat($translate.instant('WARNINGS.WARN_UNIT_COST_CONFIRM'));
  
```

client/src/i18n/en/errors.json
```diff
@@ -59,5 +59,11 @@
     "ER_ACCESS_DENIED_ERROR" : "Access denied for the current user",
     "ER_NO_DB_ERROR" : "No database selected",
     "ER_TOO_MANY_FIELDS" : "There are many more columns"
-  }
+  },
+  "WARNINGS" : {
+    "WARN_UNIT_COST_PROBLEM_TOO_HIGH" : "The new unit cost of {{newUnitCost}} {{curr}} is much higher than the typical unit cost of {{medianUnitCost}} {{curr}}.",
```
jniles: I don't see the point of having `PROBLEM` in these key.  Can you just make it `WARN_UNIT_COST_TOO_HIGH`  and so on?

client/src/css/structure.css
```diff
@@ -1196,7 +1196,7 @@ table.lot-schedule .right {
   text-align: right;
 }
 
-.allocationBasisColHeader * div {
```
jniles, line 1199: 👍 

Did you do a quick scan to make sure this wasn't used somewhere else?

client/src/modules/stock/entry/modals/lots.modal.js
```diff
@@ -267,6 +278,35 @@ function StockDefineLotsModalController(
...
+        .then(() => {
+          // Accept the new unit cost into the form
+          vm.form.setUnitCost(vm.stockLine.unit_cost);
+          onChanges();
+        })
+        .catch(() => {
+          // Revert to the previous unit cost in the form and display the warning
+          vm.stockLine.unit_cost = prevUnitCost;
+          vm.form.setUnitCost(vm.stockLine.unit_cost);
+          vm.errors = [$translate.instant(errMsg1)];
```
jniles: This should already be translated.  Do you need to translate it again?

server/controllers/inventory/index.js
```diff
@@ -221,6 +227,63 @@ function getInventoryItemsById(req, res, next) {
...
+
+    Promise.all([
+      db.exec(costListQuery, [db.bid(uuid)]),
+      db.one(costStatsQuery, [db.bid(uuid)]),
+    ])
+      .then(([costs, stats]) => {
+        // Compute the median unit cost
+        const unitCosts = costs.map(item => Number(item.unit_cost)).sort();
+        const count = unitCosts.length;
+        const medianUnitCost = count % 2 === 0
```
jniles: Nice!  Let's make this into a utility function that we could document and test.

```js
const median = (array) => {
  const len = array.length;
  const center = (len / 2) - 1;
  if (len % 2 === 0) {
    return (array[center] + array[center+1]) / 2
  }  else {
    return array[center]
  }
};
```

This will give us a reusable function, and also give us the ability to right test cases for it.

server/controllers/stock/reports/purchase_prices/index.js
```diff
@@ -77,6 +77,14 @@ async function report(req, res, next) {
       db.exec(entriesSQL, [inventoryUuid]),
     ]);
 
+    // Compute the median unit cost
+    const unitCosts = entries.map(item => Number(item.unit_cost)).sort();
```
jniles: Yeah!  See, you could use the `median()` function here.

jniles (commented):
Looks really good to me.  I've made some clean-up comments, but the logic is sound.

client/src/modules/inventory/inventory.service.js
jmcameron: Done

client/src/modules/purchases/create/createUpdate.js
jmcameron: 🆗

client/src/modules/stock/entry/modals/lots.modal.js
jmcameron: Done

client/src/modules/stock/entry/modals/lots.modal.js
jmcameron: I could  not do this completely since I needed ErrMsg1 later.  So I did not change this.

client/src/css/structure.css
jmcameron: Did it when I did a global search and replace.

server/controllers/inventory/index.js
jmcameron: Done.  And added server unit tests.

server/controllers/stock/reports/purchase_prices/index.js
jmcameron: 😃

client/src/modules/stock/entry/modals/lots.modal.js
jmcameron: Oops. Fixed.

client/src/modules/stock/entry/modals/lots.modal.js
jmcameron: Done.  Good suggestion!

client/src/i18n/en/errors.json
jmcameron: Fixed this.  Thx.

jmcameron:
@jniles This is now ready for a review.  I've incorporated your suggests and updated with master (after the WAC PR landed).

jmcameron:
@jniles I've improved handling of the case where there is no purchase history.   I've updated this PR to also handle sanity checks during creation of purchase orders.  And I've updated suggestions for testing in the PR description.  It is now ready for review.

jmcameron:
> @jniles I've improved handling of the case where there is no purchase history. I've updated this PR to also handle sanity checks during creation of purchase orders. And I've updated suggestions for testing in the PR description. It is now ready for review.

Oops. Checking why the integration tests fail...

server/controllers/stock/reports/purchase_prices/index.js
```diff
@@ -77,6 +78,11 @@ async function report(req, res, next) {
       db.exec(entriesSQL, [inventoryUuid]),
     ]);
 
+    // Compute the median unit cost
+    // Note: the median_price will be 'null' if there is no purchase history
+    const unitCosts = entries.map(item => Number(item.unit_cost));
+    totals.median_price = util.median(unitCosts);
```
jniles, line 84: Nice 👍

server/lib/util.js
```diff
@@ -252,6 +255,40 @@ function getRandomColor() {
...
+ *
+ * @param {Array} arrayIn - the array of values (not changed)
+ *
+ * @return {number} returns the median value of the array
+ */
+function median(arrayIn) {
+  if (arrayIn.length === 0) {
+    return null;
+  }
+  const array = [...arrayIn].sort(); // Make a sorted of the input array
```
jniles, line 277: This is cool 👍

test/server-unit/util.spec.js
```diff
@@ -110,4 +110,26 @@ describe('util.js', () => {
         });
       });
   });
+
+  it('#median() should return the median of an array', () => {
```
jniles, line 114: Awesome! 👍

client/src/modules/purchases/create/createUpdate.js
```diff
@@ -112,6 +117,48 @@ function PurchaseOrderController(Purchases, PurchaseOrder, Notify,
...
+      // NOTE: If there are no previous purchases, there is no purchase history to rely on
+      // for sanity checks so accept whatever the value is entered
+      const medianUnitCost = Number(item.stats.median_unit_cost);
+      const newUnitCost = Number(item.unit_price);
+      const allowableMinChange = 0.5; // Warn about entries that are less than half the previous unit cost
+      const allowableMaxChange = 2.0; // Warn entries that more than double the previous unit cost
+      const lowerUnitCostBound = (medianUnitCost * allowableMinChange);
+      const upperUnitCostBound = (medianUnitCost * allowableMaxChange);
+      if ((newUnitCost < lowerUnitCostBound) || (newUnitCost > upperUnitCostBound)) {
+        const msgParams = { newUnitCost, medianUnitCost, curr : vm.rate.currency.symbol };
```
jniles, line 137: Can you round the median unit cost to `.toFixed(4)`?  

image

client/src/i18n/en/errors.json
```diff
@@ -59,5 +59,11 @@
     "ER_ACCESS_DENIED_ERROR" : "Access denied for the current user",
     "ER_NO_DB_ERROR" : "No database selected",
     "ER_TOO_MANY_FIELDS" : "There are many more columns"
-  }
+  },
+  "WARNINGS" : {
+    "WARN_UNIT_COST_TOO_HIGH" : "The new unit cost of {{newUnitCost}} {{curr}} is much higher than the typical unit cost of {{medianUnitCost}} {{curr}}.",
+    "WARN_UNIT_COST_TOO_LOW" : "The new unit cost of {{newUnitCost}} {{curr}} is much less than the typical unit cost of {{medianUnitCost}} {{curr}}.",
+    "WARN_UNIT_COST_REASON"  : " This may indicate a misunderstanding about the quantity or units of packaging for the purchase order.",
+    "WARN_UNIT_COST_CONFIRM" : " Press \"Confirm\" to accept this change or \"Cancel\" to correct the Entry"
```
jniles: Do we need "entry" capitalized here?  I would change this to "to correct the cost" or "correct the value".

jniles (commented):
This is nearly ready to go.  I've done some testing and it works pretty well!  I've caught some issues with rounding though - could you change the median costs shown to a rounded `.toFixed(4)`?

client/src/modules/purchases/create/createUpdate.js
jmcameron: Done (for median, not user entered value).

client/src/i18n/en/errors.json
jmcameron: Fixed.

jniles:
LGTM.

Bors r+

PR #6053, opened 2021-10-21
Transpose the table of allocation values in the cost center step-down report

### LINKED ISSUE #6039: Transpose the table of allocation bases in Cost Center Step Down Report
In the cost center step down report, we have a table of allocation bases at the bottom.  It looks like this:
image

I propose we transpose the columns/rows here.  We will probably always have more cost centers than we do bases, and it is better to have this go vertically than horizontally.

We may also want to have that final table as an optional inclusion.  On the report configuration page, we can say "Include values of allocation bases" or something with a yes/no selection so that users can hide it if they don't need it.

jmcameron:
Also adds option to suppress or display the table of allocation values.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6039

**TESTING**
- Use bhima_test with sample cost center data (select showing the table)
- Compare to original data.   Here is a screenshot of the un-transposed table:
image
- Try suppressing the table

### DIFF
--- a/server/controllers/finance/cost_center_allocation_registry.js
+++ b/server/controllers/finance/cost_center_allocation_registry.js
@@ -82,7 +82,7 @@ async function fetch(session, params) {
   const hView = [];
   let totalAfterAllocation = 0;
 
-  // heck to get the correct number of colums
+  // heck to get the correct number of columns
   const numAuxiliaryCenters = data.filter(value => !value.is_principal).length;
 
   for (let i = 0; i < data.length; i++) {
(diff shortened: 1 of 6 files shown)


server/controllers/finance/cost_center_allocation_registry.js
```diff
@@ -123,14 +123,25 @@ async function fetch(session, params) {
   // cost center allocation keys details
   const { costCenterList, costCenterIndexes } = await ccAllocationKeys.fetch();
 
+  // Transpose the allocation columns
+  const allocationColumns = costCenterIndexes.map(item => item.index);
+  const allocationRows = [];
+  costCenterList.forEach((cName, i) => {
+    allocationRows.push({
+      centerName : cName,
+      allocationValues : costCenterIndexes.map(item => item.distribution[i].value),
```
mbayopanda: In a production database I experienced that `item.distribution[i]` can be undefined sometime

image

Which thrown an error for `item.distribution[i].value`

mbayopanda (commented):
I have tested it with the test database and it succeed 👍🏽 

But with our production database I have got this error: 
image

When fixed this PR can be merged

server/controllers/finance/cost_center_allocation_registry.js
jmcameron: Fixed.  @mbayopanda Please try this again.

jniles (approved):
LGTM.

bors r+

PR #6057, opened 2021-10-25
Fix(Exchange rate) display correctly the exchange rate

### LINKED ISSUE #6056: Display correctly the exchange rate in the home page
As BHIMA supports multi-currency, the exchange rate disposition in the home page has been broken.

WhatsApp Image 2021-10-25 at 01 16 14

jeremielodi:
I propose to display the exchange rates as follow :

new_rate

closes #6056

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -1023,9 +1023,10 @@ td.center {
 /* -------------------------------------
     Segment Statistic Card
 --------------------------------------- */
-.ima-stat-card .ui.statistic > .value {
+.ima-stat-card .ui.statistic > .value, .exchange-rate {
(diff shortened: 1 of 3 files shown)


jmcameron (commented):
It seems to me that all currencies should be expressed in terms of the enterprise currency.  So for Bhima test, should it be something like:

  930 Fc = $1 USD
set on <date>

  0.84 eu = $1 USD
set on <date>

Similar to what you see on the exchange rate page...

jeremielodi:
Thanks @jmcameron.

jmcameron (approved):
Thanks!  Looks good.

jniles:
LGTM.

bors r+

PR #6059, opened 2021-10-26
perf(stock): use invoiceUuid instead of reference

### LINKED ISSUE #6007: Optimize /invoices/consumable API
To look up consumable items from an invoice, we currently hit the `invoices/consumable` route with parameters.  An example is below:
```sh
/invoices/consumable/?invoiceReference=IV.HEV.165950&patientUuid=0460066886AC4DDB817BD3A537BCAE26
```

We should be using the invoice UUID instead of the invoice reference.  This saves us a lookup on the server. Ideally, the API should look like this:
```sh
/invoices/{uuid}/consumable/?patientUuid=XXXX
```

since that is more REST-like.

jniles:
This commit uses the invoice uuid instead of the text reference for a minor performance bump.

Closes #6007.


lomamech:
LGTM

bors r+

PR #6067, opened 2021-11-04
Add global lot description and peremption

mbayopanda:
This PR adds global description and expiration date in the lot modal 

image

The need is in case we have to identify each box as a lot with a unique identifier, and we have also batch number for boxes.

We have to save two information:
- box unique identifier : lot label in bhima context
- the box lot : lot description in bhima context


jmcameron (commented):
It mostly looks good. 
I tried it and it works for the first item in the lots modal.  

However, if you click [Add] in the lots modal to enter a second lot, then the expiration date is not applied to the second item:
image

Also, should the "Activate global entry..." flag be remembered so when you re-open the lot modal it will start up in the same mode with the same global description and expiration date?  I'm not sure about the use case for this, but I thought I should ask.

mbayopanda:
Thank you @jmcameron for the review let me review it again

jmcameron (approved):
Looks fine and works nicely.
I think it is ready to go.  
However I suggest you rebase first and then release it yourself.

PR #6070, opened 2021-11-04
Enable stock entry and exit using barcodes

### LINKED ISSUE #6066: Stock exit using barcode scanner
This issue is about to add a way to fill the stock exit grid by reading barcodes.

The proposal is to have a modal in which you scan barcodes when you finish with the modal by submitting, the stock exit grid will be filled and you will have just to click submit.

The need is for fast entering data.

jmcameron:
Adds support for using barcodes for entering and exiting stock.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6066

**TESTING STOCK ENTRY**
- **Go to Stock > Stock Entry** 
- Select a source
- Optionally:  Select the Fast Insert mode
- Select the global description/expiration date mode and fix the global expiration date
- Click on the [Add] button
- Enter the inventory code
- Click on the Lots link.  The modal will have a new column for barcode entry:
   image
- Click on the new 'Scan' button. The Barcode modal will pop up.
- Scan the desired barcode.   It should populate the data for the lot.  You may need to update the expiration date and global quantity before you can submit the modal.
- You should be able to submit the lot or add new items and scan new lots before submitting the modal.
- Then submit the parent "Stock Entry" form to complete the stock entry.

**TESTING STOCK EXIT**
- **Go to Stock > Stock Exit**
- Select a destination
- Notice the new [Scan] button next to the [Add] button
  image
- Click on the [Scan] button and scan a barcode.  If it found and unique, it a new row will automatically be created for it (with quantity = 1 which can be adjusted).  If the barcode cannot be found or is not unique, error messages will be displayed.
- Additional lots can be added by using the [Scan] button again
- Submitting the form should work.
- Then the parent "Stock Exit" form can be submitted to complete the stock exit


client/src/modules/stock/exit/exit.js
```diff
@@ -334,6 +334,44 @@ function StockExitController(
...
+            dateTo : vm.movement.date,
+            includeEmptyLot : 0,
+          })
+            .then(lots => {
+              if (lots.length <= 0) {
+                Notify.danger('STOCK.LOT_NOT_FOUND', 20000);
+                return;
+              }
+              if (lots.length > 1) {
+                Notify.danger('STOCK.DUPLICATE_LOTS', 20000);
```
mbayopanda, line 353: 👍🏽

mbayopanda (commented):
Thank you @jmcameron for the review and you did a great job by thinking for the entry also 👍🏽 

There are some stuff that I have noticed, in the lot modal could you update the column to hide when we check this : 
image

The expiration date column must be hidden not the quantity column, the code is in this function 
image

I have also noticed that we can scan more than twice the same lot for a given inventory 
image

jmcameron:
> The expiration date column must be hidden not the quantity column, the code is in this function

Done.

> I have also noticed that we can scan more than twice the same lot for a given inventory

Done.  Although this only applies in "global" mode since.  In this mode it also make sure that the lot label does not already exist in the inventory (since incoming units must have unique lot labels / barcodes).

mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #6072, opened 2021-11-05
Feature different ways posting payroll data

lomamech:


lomamech:
@jmcameron @mbayopanda please review this Pull Request

client/src/i18n/en/payroll.json
```diff
@@ -51,5 +51,14 @@
     "GROSS_SALARY" : "Gross salary",
     "IS_NUMBER_OF_DAYS" : "This rubric is the number of days",
     "NUMBER_OF_DAYS" : "Number of days"
+  },
+  "PAYROLL_SETTINGS" : {
+    "DEFAULT" : "Default",
+    "DEFAULT_HELP_TEXT" : "With this function, transactions related to employee payment are done in bulk and require that each expense account be linked to a cost center",
```
jmcameron: Suggest changing "function" to "option" to match the other entries

server/models/migrations/next/migrate.sql
```diff
@@ -316,3 +316,14 @@ INSERT INTO `unit` VALUES
...
 INSERT INTO `report` (`report_key`, `title_key`) VALUES
   ('cost_center_income_and_expense', 'TREE.COST_CENTER_INCOME_EXPENSE_REPORT');
+
+/*
+ * @author: lomamech
+ * @date: 2021-10-19
+ * @desc: Payroll Settings
+ */
+INSERT IGNORE INTO `unit` VALUES
+  (304, '[SETTINGS] Settings', 'TREE.PAYROLL_SETTINGS', 'Payroll Settings', 57, '/stock/setting');
```
jmcameron: The path (last field) is wrong.  It should be '/payroll/setting'

client/src/i18n/en/tree.json
```diff
@@ -127,6 +127,7 @@
     "PAYROLL" : "Payroll",
     "PAYROLL_RUB_CONFIGURATION" : "Payroll Rubric Configuration",
     "PAYROLL_RUB_MANAGEMENT" : "Payroll Rubric Management",
+    "PAYROLL_SETTINGS" : "Payroll Setting",
```
jmcameron: I suggest making this "Payroll Settings"  (plural).  It reads better in English.

server/controllers/payroll/multiplePayroll/find.js
```diff
@@ -39,7 +39,7 @@ function find(options) {
         paiement.payroll_configuration_id,  paiement.currency_id, paiement.paiement_date, paiement.base_taxable,
```
jmcameron: Since we are using English for variable names, I suggest changing 'paiement' to 'payment' where practical.

jmcameron (commented):
Due to my limited knowledge of the Payroll system, I was only able to do a superficial review.   
I did review the code and it looked okay.
I ran Bhima with this PR and was able to see and test the settings page.
I was not able to do further testing.

I did find an error in the migrate.sql file and suggested a few other minor updates.

lomamech:
@jmcameron I think I'm done with the first part of the reviews

client/src/i18n/en/payroll.json
jmcameron: Thanks

server/models/migrations/next/migrate.sql
jmcameron: Fixed. Thx

client/src/i18n/en/tree.json
jmcameron: Thx

server/controllers/payroll/multiplePayroll/find.js
jmcameron: Wow!  Glad you did that globally!  Thanks

jmcameron (commented):
Thanks for implementing the changes I suggested.

lomamech:
For @jmcameron @mbayopanda 

To be able to test the different methods that I created, I modified the test database
I added three period of tested the three methods

- Payroll sample 1 (Default)
- Payroll sample 2 (Individually per employee)
- Payroll sample 1 (Group by cost centers)

To be able to test the first method, you should go to the Payroll Settings menu, and choose the default mode

image

After choosing, go to the **Multiple Payroll module**
Click on the search button and search for the pay period for which you want to be paid
image
Select all employees
image

And then click on the Menu and click on the Configure for payment button
_This operation makes it possible to calculate the agents' pay based on a pre-recorded database for each heading of the pay slip._
image

After having done this operation, choose all the employees again and click on the Menu and this time the button **Put on Waiting list** This operation will write the transactions in the posting journal

The same operation must be done for the other Payroll sample 2 and Payroll Sample 3 periods, while changing the mode each time

jmcameron (commented):
I tested using the new suggestions.
The interface seems to work very well.
I did review different payslips using the different payroll methods, however payroll is not my area of expertise so I had no way to evaluate the correctness of the payroll calculations.

I have a few suggestions regarding the payslips:
- In English, the term "non-taxable" is not hyphenated.  use:   nontaxble
- I see the English term "Net Non-Taxable" on the payroll payslips.  I think this conveys the wrong idea:  In typical financial term "net" means total minus something (eg net profit = income - expenses).  So I think the term should be "Total Nontaxble"
- On the payslips, when using the Enterprise currency, there is no need to print the Exchange rate line.  It will always be 1 !
- If there is any chance that the payroll method might change at a particular site, it might make sense to include a code for that on the Payslip.  It does not need to be understandable by the employee, but available for future reference for the site financial administrators.

lomamech:
_On the payslips, when using the Enterprise currency, there is no need to print the Exchange rate line. It will always be 1 !
If there is any chance that the payroll method might change at a particular site, it might make sense to include a code for that on the Payslip. It does not need to be understandable by the employee, but available for future reference for the site financial administrators._ @jmcameron **I am thinking of making this change in another PR, given the many changes I have made**

jmcameron:
> _On the payslips, when using the Enterprise currency, there is no need to print the Exchange rate line. It will always be 1 ! If there is any chance that the payroll method might change at a particular site, it might make sense to include a code for that on the Payslip. It does not need to be understandable by the employee, but available for future reference for the site financial administrators._ @jmcameron **I am thinking of making this change in another PR, given the many changes I have made**

Certainly!  That is a good idea.

jmcameron:
@lomamech I rebased this for you and changed "Payroll Bulletins" to "Payroll Payslips" (English only).,

PR #6076, opened 2021-11-05
Add lot description in the lot registry

mbayopanda:
This PR adds lot description in the lot registry


jmcameron (approved):
The code is great.
I tested it and it works fine.
Two suggestions:  
 - Make the description column hidden initially to simplify the display.  Users can use the "Columns" menu item to display it.
 - [PERHAPS in a separate PR?] Please add the lot description field to the stock entry and stock exit report documents.

jmcameron:
Also, since the Stock Lots page has a lot of columns, you may want to make the "Tags" column hidden initially too.

jmcameron:
Looks good!

bors r+

PR #6077, opened 2021-11-05
Add optional daily balances to the RUMER report

### LINKED ISSUE #6064: Create a report on stock balances by day for one month
Our friends at Vanga have requested we implement a new kind of report very similar to the RUMER, but which provides the _stock balances_ for each medication, rather than the stock consumption for each medication.  Specifically, they are trying to answer:

> Our goal is to never run out of essential medications.  How much of the month do we have a rupture of stock?  Or how are we doing at having 100% of our medications 100% of the time?

Additionally, they provided the sample file to demonstrate how they are currently calculating stock outs:
[Rumor Oct 2021 with percent calculations and daily totals. -1.xlsx](https://github.com/IMA-WorldHealth/bhima/files/7453136/Rumor.Oct.2021.with.percent.calculations.and.daily.totals.-1.xlsx)

As an explanation:

>  At the end of this month we have 42 items on this list that have a stock of zero at the end of this month.  That is 23% of the total items on the list.  For each item I calculated what percent of the month was this item in rupture. So for example, at the start of this month we were out of Poche de Sang (C.P.D.) 250 ml for the first 6 day for a total of 19% of the month.  The time in rupture shows up in red.  So Acide Folique 5 mg / Ces we ran out on 23rd of October for a total of 29% of the month.  

> This is a time consuming report to create by hand but being able to display the daily totals helps graphically display what is happening with the stock and then we can work on the goal of lowering the percentage of time we spend in rupture of stock from 15% to hopefully ZERO.  

-----

This sounds fairly straightforward to do - we'll need to just use the `stock_movement_status` to generate a report of stock balances rather than stock consumption.  We could even use the same RUMER report handlebars structure to create it.

jmcameron:
Add an option to the RUMER report to all adding extra rows to show the daily stock entries and balances.  

Added a column on the far right indicating the percent of days of stock out for each stock article.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6064

I tested this against the Vanga database and it seems to work correctly.  I spot-checked against the example spreadsheet sent to us by the Vanga medical director and it matched except in a few cases where the database data I downloaded earlier this week is different than the earlier data they had.   Also note that the example spreadsheet is done manually and it is possible there are some errors there as well.

The numbers made sense for the VANGA Stock Pharmacy.  However there are some cases in the Usage Pharmacy where the daily balances are negative.  This not the fault of this PR but is most likely related to units issues that they are having problems with there.

**TESTING**
- Use the Vanga dataset
- Stock > Reports > RUMER
   - Select: Stock Pharmacy
   - Select: 2021
   - Select:  October
   - Try the report with and without the new "Include daily balances" option.
   - The version without the daily balances should match the previous report

### DIFF
--- a/client/src/i18n/en/report.json
+++ b/client/src/i18n/en/report.json
@@ -177,9 +177,13 @@
       "TITLE": "Purchase Price Analysis"
     },
     "RUMER": {
+      "BALANCE": "Balance",
       "CONDENSED_REPORT": "Condensed report",
       "DESCRIPTION": "Drug Usage and Recipe Register",
+      "ENTRIES": "Entries",
       "EXCLUDE_OUT_STOCK_ITEMS": "Exclude inventories without movements and out of stock",
+      "INCLUDE_DAILY_BALANCES": "Include daily stock balances and out-of-stock percentage",
+      "PERCENT_STOCK_OUT": "% Stock Out",
       "STOCK_BEGINNING": "Stock at the beginning of the month",
       "STOCK_END": "Stock at the end of the month",
       "TITLE": "RUMER",
--- a/client/src/i18n/fr/report.json
+++ b/client/src/i18n/fr/report.json
@@ -161,9 +161,14 @@
       "TOTAL_ALL_STOCK_MOV": "Total de tous les mouvements de stocks"
     },
     "RUMER": {
+      "BALANCE": "Solde",
       "CONDENSED_REPORT": "Rapport condensé",
       "DESCRIPTION": "Registre d’utilisation des médicaments et recettes",
+      "ENTRIES": "Entrées",
       "EXCLUDE_OUT_STOCK_ITEMS": "Exclure les inventaires sans mouvements et en rupture de stock",
+      "INCLUDE_DAILY_BALANCES": "Inclure les soldes quotidiens des stocks et le pourcentage de rupture de stock",
+      "INCOMING": "Entrant",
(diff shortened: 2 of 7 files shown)


lomamech (commented):
After having analyzed the relationship with Mr. Neron, Mr. Neron suggests that he can also have if possible the number of days of rupture,

I suggest that it can also have the symbol **%** just because if we have several drugs in a repository, with a very long list, this symbol will be important

By analyzing the code the percentage is calculated on the basis of dailyConsumption.length, do you think that it will also be possible to display this information (this recommendation is optional)

````  inventory.percentStockOut = Number(100 * (numStockOutDays / dailyConsumption.length)).toFixed(1);  ````

lomamech:
@jmcameron It will also be necessary to update the condensed report 
image

jmcameron:
> After having analyzed the relationship with Mr. Neron, Mr. Neron suggests that he can also have if possible the number of days of rupture,

Done

> I suggest that it can also have the symbol **%** just because if we have several drugs in a repository, with a very long list, this symbol will be important

Done

jmcameron:
> @jmcameron It will also be necessary to update the condensed report 

Done.  Also added toggle so that only one of the checkboxes "Condensed Report" or "Daily Balances" can be selected at the same time.

mbayopanda (approved):
LGTM 👍🏽 

image

lomamech:
image
 It's very well

PR #6079, opened 2021-11-08
feature(Payslip for Payroll Systems with index)

### LINKED ISSUE #6075: Payslip for Payroll systems with index
It is a question in designing a pay slip model usable for the case of payroll systems with Index

lomamech:
- Implement payslip for Payroll systems with index

closes #6075


lomamech:
@mbayopanda @jmcameron   can you review this Pull Request
image

server/controllers/payroll/reports/payslipGeneratorIndex.handlebars
```diff
@@ -0,0 +1,259 @@
...
+                      {{ label }}
+                      {{#if ratePercentage}}
+                        ({{ ratePercentage }} %)
+                      {{/if}}
+                    </td>
+                    <td style="width: 50%;" class="text-right"> 
+                      {{currency (multiply result ../../payrollPeriod.exchangeRate) ../../payrollPeriod.currency }}
+                    </td>
+                  </tr>
+                {{/each}} --}}
```
mbayopanda: You can remove these commented lines

mbayopanda:
@lomamech how can we test this feature ?

lomamech:
> @lomamech how can we test this feature ?

@mbayopanda  To be able to test this functionality, if you are in the test database, after having executed all the integration tests, all you have to do is go to the **Administration /Enterprise** menu and modify the parameter

image
 
Then go to **Human Resources/Multiple Payroll** to find the payment period February 2018

image
image

And to see the payslip of all the employees, just select all the employees, then go to the Menu and choose the option Payroll Bulletin

image


image

jmcameron (commented):
The code looks good.
The resulting Payslip looks very nice!   I did not check the accuracy of the data in the form.

One suggestion.  I think the menu item "Payroll Bulletins" does not translate well into English.  I suggest "Payroll Payslips" so that it matches the term in the employee action menu.
image


One more thing.  This is a suggestion to consider.  Even it if it seems like a good idea, it could be handled in a separate PR.  It looks like these payslips are designed to print in "Landscape" mode, one per page.   However the payslip will work very nicely in "Portrait" mode.   So if you're printing one payslip (action menu), then the current approach is fine although "Portrait" would also work.  However, for multiple payslips, it would be good to print them in "Portrait" mode, 2 per page.  That would save paper at the production sites.  Perhaps a dotted line could be added in the middle of the page to show where to cut the two apart.

mbayopanda (approved):
One small comment but the code looks good to me

lomamech:
> The code looks good. The resulting Payslip looks very nice! I did not check the accuracy of the data in the form.
> 
> One suggestion. I think the menu item "Payroll Bulletins" does not translate well into English. I suggest "Payroll Payslips" so that it matches the term in the employee action menu. image
> 
> One more thing. This is a suggestion to consider. Even it if it seems like a good idea, it could be handled in a separate PR. It looks like these payslips are designed to print in "Landscape" mode, one per page. However the payslip will work very nicely in "Portrait" mode. So if you're printing one payslip (action menu), then the current approach is fine although "Portrait" would also work. However, for multiple payslips, it would be good to print them in "Portrait" mode, 2 per page. That would save paper at the production sites. Perhaps a dotted line could be added in the middle of the page to show where to cut the two apart.

At IMCK he chooses the option of printing two payslips on a single page thanks to the parameters of the printer, but I think we can configure it

jmcameron:
@lomamech I fixed the merge with master issue and rebased.

PR #6094, opened 2021-11-15
Set created_at dates in BHIMA test data (for System Usage Report)

jmcameron:
Updated the BHIMA test data to set the 'created_at' times for the 'cash' table and the 'stock_movements' table.  

Since the System Usage Report relies on 'created_at' times, this makes testing that report easier with the test data (bhima_test).

This change should not affect the regression tests because none of them use the 'created_at' time.

**TESTING**
- Use bhima_test and run 'yarn build:db'
- In Hospital > Report > System Usage Statistics Report, get the report for today (the day you test it)
   - Note that there are two entries.  There should be 1 cash and 1 invoice
   - Do a stock exit. 
   - View the report again and note the stock_movement
- Run yarn:integration
   - The tests should all pass
   - Look at the System Usage Report again
     -  There should be a total of 2 patients, 3 invoices, 3 cash payments, and 11 stock movements


mbayopanda:
Hi @jmcameron, I did the test but I cannot see any data in the report.

image

I build the database with the `yarn build:db` command.

jmcameron:
> Hi @jmcameron, I did the test but I cannot see any data in the report.
> I built the database with the `yarn build:db` command.

This is strange.   It is producing different results for you and for me.   If we could figure out why, perhaps we can make some progress on the problem (https://github.com/IMA-WorldHealth/bhima/issues/6062)

I suggest you try this again at different times of day and see if it makes any difference.

Unfortunately, I will not be able to work on this further until after my vacation.

jniles:
I think I have a fix @jmcameron. The issue was with timezone offsets, which we were truncating.  We really need to build a unified model of how these reports should interact with the server's timezone - I have a feeling many reports suffer from this.

Please test this and see if it works on your end.

This shouldn't cause a problem to a local hospital, but anyone running BHIMA on server and expecting the same results for every timezone will be surprised.  We should address this as we start using BHIMA on a more global level.

jmcameron:
> I think I have a fix @jmcameron. The issue was with timezone offsets, which we were truncating. We really need to build a unified model of how these reports should interact with the server's timezone - I have a feeling many reports suffer from this.
> 
> Please test this and see if it works on your end.
> 
> This shouldn't cause a problem to a local hospital, but anyone running BHIMA on server and expecting the same results for every timezone will be surprised. We should address this as we start using BHIMA on a more global level.

@jniles : I tested it and it works for me.   I'm looking at another issue.  Once I resolve it, I'll release it.

jmcameron:
> @jniles : I tested it and it works for me. I'm looking at another issue. Once I resolve it, I'll release it.

I was confused:  When I exited stock to a patient, it did not affect count of "Active Users".  When I looked at the code, it was clear that "Users" here refers to the staff who performed the stock movement, not to the total number of staff + patients involved.  Presuming that is correct, I think we should release this.

PR #6096, opened 2021-11-16
Average medication costs per patient report

### LINKED ISSUE #5993: Create new report giving the average cost of medications per patient/illness
Need some discussion on what this report should look like:

- Should there be breakdowns by service / cost center?
- Should there be breakdowns by exact medication or type of medication?
- Should the summary data include max/min/mean/median?

jmcameron:
New Stock report to compute the average medication costs per patient. 
The report shows the total medication costs for each Service and the number of unique patients receiving those medications.  At the bottom of the report, the grand total of medication costs, number of patients, and average medication cost per patient is shown.

**NOTE:** The total of the number of patients per service will be larger than the total number of patients shown on the final, grand total, line because some patients receive medication from more than one department.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5993

**TESTING**
- Suggest using a production database such as Vanga
- Enable access to the new report via Admin > Role Management
- Try the report at:   Stock > Reports > Average Medication Costs Per Patient
- Try some of the combinations of depot, service, date range, and currency


lomamech (commented):
After reviewing this PR, I noticed with the VANGA Database, that when displaying the report with the company currency the data is displayed correctly

image

But when displaying the report with another currency, the conversion only works for the total, at the level of details there is only the symbol of the currency which changes

image

jmcameron:
> After reviewing this PR, I noticed with the VANGA Database, that when displaying the report with the company currency the data is displayed correctly

I've fixed this issue and it is ready for a new review.

server/controllers/stock/reports/stock/avg_med_costs_per_patient_report.js
```diff
@@ -0,0 +1,137 @@
...
+
+        FROM stock_movement AS sm
+        JOIN patient AS pat ON pat.uuid = sm.entity_uuid
+        LEFT JOIN invoice AS inv ON inv.uuid = sm.invoice_uuid
+        WHERE
+          sm.flux_id = 9 AND sm.is_exit = 1
+          AND DATE(sm.date) BETWEEN DATE(?) AND DATE(?)
+          ${depotSql}
+          ${serviceSql}
+          -- AND sm.entity_uuid IN (
```
jniles: Can you remove these comments?

server/controllers/stock/reports/stock/avg_med_costs_per_patient_report.js
```diff
@@ -0,0 +1,137 @@
...
+
+          -- Computes: COUNT(DISTINCT sm.entity_uuid) OVER (PARTITION BY sm.depot_uuid, inv.service_uuid)
+          (DENSE_RANK() OVER (PARTITION BY sm.depot_uuid,
+              inv.service_uuid ORDER BY sm.entity_uuid ASC) +
+           DENSE_RANK() OVER (PARTITION BY sm.depot_uuid,
+              inv.service_uuid ORDER BY sm.entity_uuid DESC) - 1) AS srvPatCount,
+
+          SUM(sm.quantity * sm.unit_cost) OVER () AS totalMedCosts,
+
+          -- Computes: COUNT(DISTINCT sm.entity_uuid) OVER ()
```
jniles: Looks like you didn't complete your thought here.

server/controllers/stock/reports/stock/avg_med_costs_per_patient_report.js
```diff
@@ -0,0 +1,137 @@
...
+          sm.unit_cost,
+          sm.quantity,
+
+          SUM(sm.quantity * sm.unit_cost) OVER (PARTITION BY sm.depot_uuid, inv.service_uuid) AS srvTotal,
+
+          -- Computes: COUNT(DISTINCT sm.entity_uuid) OVER (PARTITION BY sm.depot_uuid, inv.service_uuid)
+          (DENSE_RANK() OVER (PARTITION BY sm.depot_uuid,
+              inv.service_uuid ORDER BY sm.entity_uuid ASC) +
+           DENSE_RANK() OVER (PARTITION BY sm.depot_uuid,
+              inv.service_uuid ORDER BY sm.entity_uuid DESC) - 1) AS srvPatCount,
```
jniles, line 67: Why the `-1`?  Are we implicitly adding one somewhere?

jniles (commented):
Just a few cosmetic comments, otherwise, it looks really good!  I sent it off to the MD of Vanga for their comments, but I'm otherwise happy with it.

lomamech (approved):
I approve of the change, Jonathan has to add some comment, once you get them processed I will merge this PR

server/controllers/stock/reports/stock/avg_med_costs_per_patient_report.js
jmcameron: That is the way this work-around works!  See https://stackoverflow.com/questions/11202878/partition-function-count-over-possible-using-distinct

server/controllers/stock/reports/stock/avg_med_costs_per_patient_report.js
jmcameron: Done

server/controllers/stock/reports/stock/avg_med_costs_per_patient_report.js
jmcameron: Done

server/controllers/stock/reports/stock/avg_med_costs_per_patient_report.js
jniles: I understand - if we can have NULL values, then this is required, presumably to remove the NULL from the count (since its grouped).

jniles (approved):
Looks good to me.

bors r+

jniles:
Howdy bors.  You there?

bors r+

PR #6106, opened 2021-11-25
bug(Fix Display Decimal point)

### LINKED ISSUE #6104: Problem with the display of a decimal point
It is found that in some reports for the case where the corporate currency is a **currency which does not provide for decimal digits** and the report is viewed with a currency which requires decimal digits, the we can see that there are no digits after the decimal point

lomamech:
- Removal of the call of an unnecessary and undefined parameter in
  helpers finance in function currency()

closes #6104

### DIFF
--- a/server/lib/template/helpers/finance.js
+++ b/server/lib/template/helpers/finance.js
@@ -36,7 +36,13 @@ function currency(value = 0, currencyId, digit) {
   } else {
     // usage with custom digits for precision
     const format = JSON.parse(JSON.stringify(fmt));
-    format.precision = digit;
+
+    /*
+      * lomamech : there had been a problem with the display of digit after the decimal
+      * point after investigation, I realized that the value "digit" was undefined
+    */
+    // format.precision = digit;
+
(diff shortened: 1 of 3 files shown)


lomamech:
image

To be able to test this PR, all you have to do is use the **Vanga Database** and view a report by choosing a currency other than that of the Enterpriser

server/lib/template/helpers/finance.js
```diff
@@ -36,7 +36,13 @@ function currency(value = 0, currencyId, digit) {
   } else {
     // usage with custom digits for precision
     const format = JSON.parse(JSON.stringify(fmt));
-    format.precision = digit;
+
+    /*
+      * lomamech : there had been a problem with the display of digit after the decimal
+      * point after investigation, I realized that the value "digit" was undefined
+    */
+    // format.precision = digit;
```
jniles: Can we just set this to be

```js
if (digit !== undefined) { format.precision = digit; }
```

I think it was originally there for a purpose.

jniles (approved):
LGTM

bors r+

PR #6188, opened 2021-12-14
Add optional indicator to many system selection components.

### LINKED ISSUE #6187: Automatically add 'optional' indicator to depot selection component
- If the depot is optional, update the bhDepotSelect component to indicate that like this: 
image 
- Adding the "(Optional)" indicator could also be done for other selector components such as bhInventoryGroupSelect.

This item was formerly in Issue https://github.com/IMA-WorldHealth/bhima/issues/6148.

jmcameron:
Updated many BHIMA components to show an '(optional)' indicator in the presentation of the selection component when required="false" (and the default label is use).  If a non-default label is give it will be used (regardless of whether the item is required or not).

The goal is to make it clearer which form elements are required and which are optional.

Also fixed a few minor bugs.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6187

**The modified components**
- bh-account-select (see: Account Statement Report (Single), Simple Voucher, vs Finance > Journal > Search)
- bh-cashbox-select (see: Cash report)
- bh-debtor-group-select (see: Client Debts vs Unpaid Invoices reports)
- bh-employee-select (see: Employee Standing vs Human Resources > Staffing indices management > Search)
- bh-fiscal-year-select (see: many reports, appearance should not change)
- bh-inventory-group-select (see: Aggregated Stock Consumption vs Stock>Stock Lots>Search
- bh-inventory-select (see: Stock Sheet vs Purchase Price Analysis)
- bh-period-select (see: Finance > Journal > Search)
- bh-period-selection (see: many Finance reports)
- bh-purchase-order-select (see: no changes, fixed Purchase Order Analysis)
- bh-service-select (see: Visit registry > Search)

**TESTING**
- Most components above indicate a place where the 'optional' indicator was added vs a place where they weren't
- Perhaps the easiest way is to use go through all the report requests
   - Check where the '(optional)' indicator is shown and where it is not.   
   - Then press submit without selecting anything to see what the form complains about.  It should not complain that any field was required when you see the optional tag for it.
   - Also try some of the various registries and click on the "Search" button and see if the placement of "(optional)" indicators makes sense.


client/src/i18n/fr/periods.json
```diff
@@ -13,6 +13,7 @@
     "START" : "Début",
     "END" : "Fin",
     "PERIOD_LIMIT" : "Limite de Période",
+    "PERIOD_LIMIT_OPTIONAL" : "Limite de Période (optinnel)",
```
jniles: This looks like a typo - optionnelle.

client/src/js/components/bhCashboxSelect.js
```diff
@@ -32,7 +32,8 @@ function CashboxSelectController(Cashbox, Notify, Session) {
     $ctrl.name = $ctrl.name || 'CashboxForm';
 
     // translated label for the form input
-    $ctrl.label = $ctrl.label || 'FORM.SELECT.CASHBOX';
+    $ctrl.label = $ctrl.label
```
jniles, line 35: Nice!

client/src/modules/reports/generate/purchase_order_analysis/purchase_order_analysis.html
```diff
@@ -23,8 +23,8 @@ <h3 class="text-capitalize" translate>REPORT.PURCHASE_ORDER_ANALYSIS.TITLE</h3>
           <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate>
             <bh-purchase-order-select
               purchase-uuid="ReportConfigCtrl.reportDetails.purchase_uuid"
-              on-select-callback="ReportConfigCtrl.onSelectPurchase(purchase)">
-              <bh-clear on-clear="ReportConfigCtrl.onClear(purchase_uuid)"></bh-clear>
```
jniles, line 27: Nice :+1:

jniles (commented):
This looks pretty good!  I caught one typo.  It's super helpful!

client/src/i18n/fr/periods.json
jmcameron: Fixed

PR #6190, opened 2021-12-14
chore(inventory): skip tags where possible

jniles:
This commit ensures that the `bhInventorySelect` doesn't trigger the tag search path, since it isn't needed.

To test:
On `master`
1. Build the default test database via the integration tests.
2. Add "DEBUG=db*" to the .env file
2. Navigate to the "stock sheet" report and notice that a lot of tags are queried on initial load.

On this branch:
1. Repeat the steps above.
2. Notice that the tags are not queried.

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -802,7 +802,7 @@
             "TURNOVER": "Turnover",
             "TYPE": "Type",
             "UNCONFIGURED": "Unconfigured",
-            "UNCONSUMABLE": "Non-consummable",
+            "UNCONSUMABLE": "Non-consumable",
             "UNDEFINED": "Undefined",
             "UNIT": "Unit",
             "UNITS": "Units",
--- a/client/src/js/components/bhInventorySelect.js
+++ b/client/src/js/components/bhInventorySelect.js
@@ -24,9 +24,11 @@ function InventorySelectController(Inventory, Notify) {
   $ctrl.$onInit = function onInit() {
     // fired when an inventory has been selected
 
-    const params = $ctrl.onlyConsumable
-      ? { consumable : 1 }
-      : {};
+    const params = { skipTags : 1 };
+
+    if ($ctrl.onlyConsumable) {
+      Object.assign(params, { consumable : 1 });
+    }
 
     // load all inventories
     Inventory.read(null, params)
--- a/client/src/js/components/bhInventorySelect.js
+++ b/client/src/js/components/bhInventorySelect.js
@@ -24,7 +24,7 @@ function InventorySelectController(Inventory, Notify) {
   $ctrl.$onInit = function onInit() {
     // fired when an inventory has been selected
 
-    const params = { skipTags : 1 };
+    const params = { skipTags : true };
 
     if ($ctrl.onlyConsumable) {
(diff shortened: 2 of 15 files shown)


jmcameron (commented):
This fix definitely speeds things up.  Nice!  However, I still see 'LEFT JOIN tags...' when going to to the Stock Sheet page --- which I presume is not necessary.

jniles:
This report reuses a common function `getInventoryItems()`, which is why it does the LEFT JOIN.  The upside is that this route is tested by our integration tests.   The downside is that it is more generic.  I think the upside outweighs the downside of rewriting our own route here.  It could be that the query optimizer just removes it, seeing that it is never used.

jmcameron:
> This report reuses a common function `getInventoryItems()`, which is why it does the LEFT JOIN. The upside is that this routes is tested by our integration tests. The downside is that it is more generic. I think the upside outweighs the downside of rewriting our own route here. It could be that the query optimizer just removes it, seeing that it is never used.

Okay.

client/src/js/components/bhInventorySelect.js
```diff
@@ -24,9 +24,11 @@ function InventorySelectController(Inventory, Notify) {
   $ctrl.$onInit = function onInit() {
     // fired when an inventory has been selected
 
-    const params = $ctrl.onlyConsumable
-      ? { consumable : 1 }
-      : {};
+    const params = { skipTags : 1 };
```
jmcameron: Since you are using skipTags as a boolean, why use '1' instead of 'true'?

jmcameron (commented):
It looks pretty good.  See my note in the code. 

But I think you need to review all uses of Stock.inventories.read() to make sure that skipTags is used appropriately.  For instance, I think skipTags=true is missing for creating purchase orders.

client/src/js/components/bhInventorySelect.js
jniles: Because of type coercion and javascript being a terrible language to work in.  Try this:

```js
// as strings of numbers
'1' == true // => true
'0' == false // => false

// as strings themselves
'true' == true // => false
'false' == false // =>true
```

Basically, even if the number are passed around as strings, they true/false properties are retained.  If "true" and "false" are passed around as strings, it changes the `if (bool) { }` behavior.

... that being said, it's much more readable to use booleans, so I'll change it and figure out where we have buggy code instead.

jniles:
@jmcameron changes landed.  I know you are on break.  What is your advice to me?  Can you take a brief look at it or shall I do my own review?

client/src/i18n/en/form.json
```diff
@@ -802,7 +802,7 @@
             "TURNOVER": "Turnover",
             "TYPE": "Type",
             "UNCONFIGURED": "Unconfigured",
-            "UNCONSUMABLE": "Non-consummable",
+            "UNCONSUMABLE": "Non-consumable",
```
jmcameron, line 805: !

jmcameron (approved):
I looked over the code changes and it looks good.  I did have time to test each of the user interfaces that could have been impacted by this change.   So I think this is ready, however, if you have not verified that the potentially affected user interfaces still work correctly, you may want to do that.

PR #6192, opened 2021-12-14
Remove selection badge when dates are hand edited

### LINKED ISSUE #6173: Cosmetic bug in date interval selection component
When you use a bhDateInterval component and select a predefined item such as "This month", it does that and marks it with the "This Month" badge.  But if you then edit the start of the interval, the "This Month" badge is not removed!  If either end of the interval is changed, the badge should be removed.

image

The second problem is when selecting a date range such as "This Year". Currently it sets the start date to the last day of the previous year.  While this does usually work correctly, the date ranges for the reports using this range show something like this:
- December 2020 - December 2021
Where the it really should read "January 2021 - December 2021".   Perhaps the date range needs to set the start dates to midnight (or one second past midnight) on the correct starting date.

jmcameron:
Fix a bug in the date interval selection component.   When a reconfigured date range is selected (eg, 'This Year') the component puts a little "This Year" badge, in orange, next to the component label.   However when either the start or end dates are hand-edited, the badge is not removed even though it is probably not still correct.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6173

**TESTING**
- Use any database
- Go to the Finance > Reports > Cash Report page
- Select a predefined date range (eg, "This Month")
- Notice the orange "This Month" badge appears
- Manually edit either the start or end dates and the badge should disappear. 
- Try the "Preview" to verify that the date range still works properly in the report generation.

Note: Even if you manually restore the date to the correct range for one of the predefined ranges, the corresponding badge will not appear.  Only click on the predefined date range links will produce the badge.

### DIFF
--- a/client/src/js/components/bhDateInterval.js
+++ b/client/src/js/components/bhDateInterval.js
@@ -1,9 +1,9 @@
 /**
  * @name bhDateInterval
  * @description
- * The `bhDateInterval` component provide a mean to select dates plage between
- * two dates. The dates values returned are send to dates models given in
- * date-from and date-to attributes.
+ * The `bhDateInterval` component provide a means to select a date range
+ * between two dates. The dates values returned are sent to dates models
+ * given in date-from and date-to attributes.
  *
  * An optional flag `limit-min-fiscal` can be provided that limits the from and
  * to date inputs to not allow dates before the start of the first enterprise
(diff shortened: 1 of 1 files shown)


client/src/js/components/bhDateInterval.js
```diff
@@ -48,6 +48,8 @@ function bhDateInterval(bhConstants, Fiscal, Session, PeriodService, $translate)
   // expose to the view
   $ctrl.search = search;
   $ctrl.clear = clear;
+  $ctrl.lastDateFrom = null;
```
jniles, line 51: Do you need to attach these variables to `$ctrl`?  I assume you could just keep them scoped to this component with `let lastDateFrom = null;` and so on, since you don't use them in the template.

This isn't a big deal, just an observation.

jniles (approved):
LGTM.

bors r+

PR #6193, opened 2021-12-14
Employee standing report

### LINKED ISSUE #6123: Financial situation of all employees
Added the ability to be able to see the financial situation of all employees, to allow managers to be able to see the overall situation

lomamech:
- Implement global situation of all employees in different mode

closes #6123


lomamech:
To test this functionality, notice the checkbox **See the overall situation of all employees**
image

Once this box is checked, the interface offers the different modes to view the report

image

**_Aggregate report_**
image
**_Synthetic report_**
image
**_Detailed report_**
image

jmcameron (commented):
Looks good.  I tried it and it seems to work well.  However, I have cosmetic and translation suggestions.
- I suggest NOT underlining the report title.   None of our other reports titles use underlining.  I think h3 + bold is enough.
- When selecting all employees, I see the term "Synthetic".  This is a little obscure in English.  I think what you are thinking of is a 'synthesis' or summary.  If that is correct, I suggest  simply using the term 'Summary' instead of 'Synthetic'.   Also, the term 'Aggregate' is a bit odd to my eye.  Not sure what to suggest, but the terms 'Normal Report', 'Full Report', 'Report for All Employees', etc.  I like the last option 'Detailed report'.   Finally, I suggest moving the "Synthetic" option to the last one in the list since the other two are more closely related.
- I suggest rewording the label for the last option from "View extract of the employee's standing by date range" to something simpler such as "Limit to time interval".   I know what you mean by "Extract" but I'm not sure everyone will.

lomamech:
> Looks good. I tried it and it seems to work well. However, I have cosmetic and translation suggestions.
> 
> * I suggest NOT underlining the report title.   None of our other reports titles use underlining.  I think h3 + bold is enough.
> * When selecting all employees, I see the term "Synthetic".  This is a little obscure in English.  I think what you are thinking of is a 'synthesis' or summary.  If that is correct, I suggest  simply using the term 'Summary' instead of 'Synthetic'.   Also, the term 'Aggregate' is a bit odd to my eye.  Not sure what to suggest, but the terms 'Normal Report', 'Full Report', 'Report for All Employees', etc.  I like the last option 'Detailed report'.   Finally, I suggest moving the "Synthetic" option to the last one in the list since the other two are more closely related.
> * I suggest rewording the label for the last option from "View extract of the employee's standing by date range" to something simpler such as "Limit to time interval".   I know what you mean by "Extract" but I'm not sure everyone will.

Well received

jmcameron (commented):
I like your changes.  One very minor cosmetic thing:  You have several 'hr' separators in the  client-side page.  I don't think they are necessary and add extra unneeded white space.

I noticed one thing that may be an issue.  What is the purpose of the "Limit to time interval" option?  If I select one employee and select various time intervals it does not seem to make a difference.  But it does make a difference for normal/summary reports.  And why would I want to use this option anyway?  Don't we always want to know the current standing?   And why chose and interval?   It seems it would make more sense to me to want to know the employee standing at some specific time in the past.   My understanding of what you're trying to do is limited so I encourage @jniles to review this PR.

jmcameron:
@lomamech since you are working on this report, do you want to look at the issues in https://github.com/IMA-WorldHealth/bhima/issues/6126?

lomamech:
@jmcameron i have finish with the updating,

lomamech:
@jmcameron i have finish with the updating,

client/src/modules/reports/generate/employee_standing/employee_standing.js
```diff
@@ -12,21 +12,28 @@ EmployeeStandingController.$inject = [
...
 
   const vm = this;
   const cache = new AppCache('configure_employee_standing');
-  const reportUrl = '/reports/finance/employee_standing';
 
   vm.reportDetails = {};
 
   checkCachedConfiguration();
 
+  vm.reportDetails.currency_id = Session.enterprise.currency_id;
```
jmcameron: Note:  Doing it this way forces the form to use the enterprise currency each time you enter the page.  The cache is ignored.  Look at the checkCachedConfiguration() function in client/src/modules/reports/generate/avg_med_costs_per_patient/avg_med_costs_per_patient.js for an example.

server/controllers/finance/reports/financial.employee.handlebars
```diff
@@ -14,13 +14,13 @@
   <h5>{{translate "FORM.LABELS.REGISTRATION_NUMBER"}} : <b class="text-uppercase"></b>{{ employee.code }}</b></h5>
   <h4 class="text-uppercase" style="background-color: #efefef;">
     <b>
-      {{translate balanceCreditorText}} = {{debcred ./creditorAggregates.balance metadata.enterprise.currency_id}}
+      {{translate balanceCreditorText}} = {{debcred (multiply ./creditorAggregates.balance exchangeRate) currencyId}}
```
jmcameron, line 17: Nice!

server/controllers/finance/reports/financial.employee.js
```diff
@@ -49,8 +58,10 @@ async function build(req, res, next) {
   }
 
   try {
-
     const data = {};
+    data.exchangeRate = exchange.rate || 1;
+    data.currencyId = options.currency_id;
```
jmcameron, line 59: See my previous note.

jmcameron (commented):
Thanks for adding the multi-currency support.  I made one suggestion in the code about caching the currency used in the report.

As far as I know, once you address the currency caching and rebase, it is ready to go.  But I do not understand how the time limits work (see my previous reviews above).  So I still suggest getting @jniles to review this before merging.

client/src/modules/reports/generate/employee_standing/employee_standing.html
```diff
@@ -22,22 +22,76 @@ <h3 class="text-capitalize" translate>REPORT.EMPLOYEE_STANDING.TITLE</h3>
...
+              <span translate>REPORT.EMPLOYEE_STANDING.SEE_SITUATION_ALL_EMPLOYEES</span>
+            </label>
+          </div>
+
+          <div ng-if="ReportConfigCtrl.reportDetails.allEmployee" class="panel-body">
+            <div class="radio">
+              <label class="radio-inline">
+                <input
+                  type="radio"
+                  name="modeRepport"
```
jniles: Can you make this name `modeReport`?  `Report` doesn't have two `pp`s in English 😄

client/src/modules/reports/generate/employee_standing/employee_standing.html
```diff
@@ -22,22 +22,76 @@ <h3 class="text-capitalize" translate>REPORT.EMPLOYEE_STANDING.TITLE</h3>
...
+                </span>
+              </label>
+            </div>            
+            <div class="radio">
+              <label class="radio-inline">
+                <input
+                  type="radio"
+                  name="modeRepport"
+                  value="normal"
+                  ng-model="ReportConfigCtrl.reportDetails.modeRepport"
```
jniles: Same with the `ngModel`

client/src/modules/reports/generate/employee_standing/employee_standing.html
```diff
@@ -22,22 +22,76 @@ <h3 class="text-capitalize" translate>REPORT.EMPLOYEE_STANDING.TITLE</h3>
...
+                  REPORT.EMPLOYEE_STANDING.NORMAL_REPORT
+                </span>
+              </label>
+            </div>
+            <div class="radio">
+              <label class="radio-inline">
+                <input
+                  type="radio"
+                  name="modeRepport"
+                  value="detailled"
```
jniles: `detailed` has only one `l` in English 😄

client/src/modules/reports/generate/employee_standing/employee_standing.js
jniles: Yes, the best way to do this would be to check if the `currency_id` is defined, and only set it if it isn't defined.

Like this:

```js
vm.reportDetails.currency_id = vm.reportDetails?.currency_id || Session.enterprise.currency_id;
```

See [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) for more information on how this works.

client/src/modules/reports/generate/employee_standing/employee_standing.js
```diff
@@ -41,12 +48,24 @@ function EmployeeStandingController($state, $sce, Notify, BaseReportService, App
...
   vm.preview = function preview(form) {
     if (form.$invalid) {
       Notify.danger('FORM.ERRORS.RECORD_ERROR');
       return 0;
     }
 
+    if (vm.reportDetails.allEmployee) {
+      delete vm.reportDetails.employee_uuid;
+      delete vm.reportDetails.includeMedicalCare;
+      reportUrl = '/reports/finance/employees_standing';
```
jniles: These two URLs are really close together!  I thought you had a typo at first.  Is it possible to either

1. Make them really different, so we don't accidentally correct this as a typo OR
2. Make the switch between all employees just a query string parameter and use the same base URL?

server/controllers/finance/reports/financial.employees.js
```diff
@@ -0,0 +1,206 @@
...
+  } else {
+    exchange = await Exchange.getExchangeRate(req.session.enterprise.id, Number(options.currency_id), new Date());
+  }
+
+  _.defaults(options, PDF_OPTIONS);
+
+  // set up the report with report manager
+  try {
+    report = new ReportManager(TEMPLATE, req.session, options);
+  } catch (e) {
```
jniles: Can you wrap this entire function block in a `try/catch` statement?  I don't see the need to have several different `try/catch` blocks, and if something goes wrong on the `await Exchange.getExhangeRate()` (like if MySQL crashes), you won't catch it.

server/controllers/finance/reports/financial.employees.js
```diff
@@ -0,0 +1,206 @@
...
+          JOIN entity_map AS map ON map.uuid = emp.creditor_uuid
+          ${filterBydateLegder}
+        ) AS aggr
+        GROUP BY aggr.employee_uuid
+        ORDER BY aggr.employee_name, aggr.number ASC
+      `;
+    }
+
+    const [financialData] = await Promise.all([
+      db.exec(sql),
```
jniles: You don't need `Promise.all()` for a single query.

jniles (commented):
@lomamech this looks pretty good!

If you show values in multiple currencies, you need to have a date and exchange rate shown on the report, as well as a warning when the user does not choose the enterprise exchange rate:

image

Otherwise, I agree with @jmcameron commentary.

client/src/modules/reports/generate/employee_standing/employee_standing.js
lomamech: Nice

client/src/modules/reports/generate/employee_standing/employee_standing.js
```diff
@@ -12,21 +12,29 @@ EmployeeStandingController.$inject = [
...
 
   const vm = this;
   const cache = new AppCache('configure_employee_standing');
-  const reportUrl = '/reports/finance/employee_standing';
 
   vm.reportDetails = {};
 
   checkCachedConfiguration();
 
+  vm.reportDetails.currency_id = vm.reportDetails?.currency_id || Session.enterprise.currency_id;
```
lomamech, line 24: @jniles  Optional chaining is not **Recognized**

jniles: Yeah, hound is probably wrong here.  Let me try updating the version of ESLint used by hound.

jniles:
Hmm... looks like I will have to investigate this more.

bors r+

PR #6196, opened 2021-12-15
Update the unpaid invoice report based on the issue #6184

### LINKED ISSUE #6184: Updates for the Unpaid Invoices Report
Finance > Report > Unpaid Invoices Report

**Suggestions**
- Increase the font size of the date range subtitle on the report
- If a Debtor Group is selected, the column for it should be removed and the name of the Debtor Group should be added as a subtitle
- The currency values are not shown with currency symbols
- Add multi-currency support
- Should the Service selector allow multiple services to be selected?

mbayopanda:
This PR is related to #6184 

image

image

image

Close #6184

### DIFF
--- a/client/src/modules/reports/generate/unpaid_invoice_payments/unpaid_invoice_payments.config.js
+++ b/client/src/modules/reports/generate/unpaid_invoice_payments/unpaid_invoice_payments.config.js
@@ -3,15 +3,18 @@ angular.module('bhima.controllers')
 
 UnbalancedInvoicePaymentsConfigController.$inject = [
   '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state',
+  'SessionService',
 ];
 
-function UnbalancedInvoicePaymentsConfigController($sce, Notify, SavedReports, AppCache, reportData, $state) {
+function UnbalancedInvoicePaymentsConfigController($sce, Notify, SavedReports, AppCache, reportData, $state, Session) {
   const vm = this;
   const cache = new AppCache('configure_unpaid_invoice_payments');
   const reportUrl = 'reports/finance/unpaid_invoice_payments';
 
   vm.previewGenerated = false;
-  vm.reportDetails = {};
+  vm.reportDetails = {
+    currencyId : Session.enterprise.currency_id,
+  };
 
   checkCachedConfiguration();
 
--- a/client/src/modules/reports/generate/unpaid_invoice_payments/unpaid_invoice_payments.config.js
+++ b/client/src/modules/reports/generate/unpaid_invoice_payments/unpaid_invoice_payments.config.js
@@ -24,6 +24,10 @@ function UnbalancedInvoicePaymentsConfigController($sce, Notify, SavedReports, A
     vm.reportDetails = angular.copy(report);
   };
 
+  vm.onSelectCurrency = currency => {
+    vm.reportDetails.currencyId = currency.id;
+  };
+
   vm.onClear = () => {
     delete vm.reportDetails.debtorGroupName;
     delete vm.reportDetails.debtorGroupUuid;
--- a/server/controllers/finance/reports/unpaid_invoice_payments/index.js
+++ b/server/controllers/finance/reports/unpaid_invoice_payments/index.js
@@ -18,6 +18,8 @@ async function build(req, res, next) {
 
   const metadata = _.clone(req.session);
 
+  qs.isEnterpriseCurrency = !!(parseInt(qs.currencyId, 10) === metadata.enterprise.currency_id);
+
   let report;
   let results;
 
--- a/server/controllers/finance/reports/unpaid_invoice_payments/index.js
+++ b/server/controllers/finance/reports/unpaid_invoice_payments/index.js
@@ -66,6 +66,7 @@ async function getUnbalancedInvoices(options) {
   const params = [
     new Date(options.dateFrom),
     new Date(options.dateTo),
+    +options.currencyId,
   ];
 
   const { debtorGroupName, serviceUuid } = options;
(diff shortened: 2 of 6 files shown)


jmcameron (commented):
It looks like you have addressed all the items in the issue.  Looks good!

A few minor issues
- The currency should default to the enterprise currency.  See client/src/modules/reports/generate/avg_med_costs_per_patient/avg_med_costs_per_patient.js for an example -- especially in checkCachedConfiguration().
- When you chose a debtor Group and put its name as subtitle, I suggest using a little larger font.
- When you select a Service, there is no need to display a separate column for it since the totals column is the same (since you are only showing one service).  However, you also need to add a subtitle for the service when you remove its column.

client/src/modules/reports/generate/unpaid_invoice_payments/unpaid_invoice_payments.config.js
```diff
@@ -24,6 +24,10 @@ function UnbalancedInvoicePaymentsConfigController($sce, Notify, SavedReports, A
     vm.reportDetails = angular.copy(report);
   };
 
+  vm.onSelectCurrency = currency => {
+    vm.reportDetails.currencyId = currency.id;
```
jniles, line 31: Could you make this default to the enterprise currency?

server/controllers/finance/reports/unpaid_invoice_payments/index.js
```diff
@@ -66,6 +66,7 @@ async function getUnbalancedInvoices(options) {
   const params = [
     new Date(options.dateFrom),
     new Date(options.dateTo),
+    +options.currencyId,
```
jniles: This is cool, but I really worry about this being `NaN`.  What happens then?  What does MySQL convert it to?

It would be much better to convert this explicitly to something.

jniles (commented):
This looks good so far!  I just made a few comments.

jniles:
@mbayopanda if you can get these issues sorted out, we can try and include this report in the next release.  I would like to do a release this Friday.

mbayopanda:
@jmcameron @jniles changes landed

client/src/modules/reports/generate/unpaid_invoice_payments/unpaid_invoice_payments.config.js
```diff
@@ -3,15 +3,18 @@ angular.module('bhima.controllers')
...
-function UnbalancedInvoicePaymentsConfigController($sce, Notify, SavedReports, AppCache, reportData, $state) {
+function UnbalancedInvoicePaymentsConfigController($sce, Notify, SavedReports, AppCache, reportData, $state, Session) {
   const vm = this;
   const cache = new AppCache('configure_unpaid_invoice_payments');
   const reportUrl = 'reports/finance/unpaid_invoice_payments';
 
   vm.previewGenerated = false;
-  vm.reportDetails = {};
+  vm.reportDetails = {
+    currencyId : Session.enterprise.currency_id,
```
jniles, line 16: :+1:

server/models/procedures/invoicing.sql
```diff
@@ -980,11 +984,13 @@ CREATE PROCEDURE UnbalancedInvoicePaymentsTable(
   CREATE TEMPORARY TABLE `unbalanced_invoices` AS (
     SELECT BUID(ivc.uuid) as invoice_uuid , em.text AS debtorReference, debtor.text AS debtorName,
       BUID(debtor.uuid) as debtorUuid,
-      balances.debit_equiv AS debit,
-      balances.credit_equiv AS credit, iv.date AS creation_date, balances.balance,
+      (balances.debit_equiv * IFNULL(GetExchangeRate(_enterpriseId, currencyId, dateTo), 1)) AS debit,
```
jniles: It looks like you can speed this up a lot by moving the `GetExchangeRate(_enterpriseId, currencyId, dateTo)` out of the SQL query.  Like this:

```sql
DECLARE exchangeRate DOUBLE;

SET exchangeRate = (SELECT IFNULL(GetExchangeRate(_enterpriseId, currencyId, dateTo), 1));

  CREATE TEMPORARY TABLE `unbalanced_invoices` AS (
    SELECT BUID(ivc.uuid) as invoice_uuid , em.text AS debtorReference, debtor.text AS debtorName,
      BUID(debtor.uuid) as debtorUuid,
      (balances.debit_equiv * exchangeRate) AS debit,
       /* etc ... */
```

This has the advantage of speeding up the query since the `GetExchangeRate()` function doesn't depend on the data in the table.  Instead of calling it 3X per row, we only call it a single time.

server/controllers/finance/reports/unpaid_invoice_payments/report.handlebars
```diff
@@ -9,47 +9,78 @@
...
+  {{/if}}
+
+  {{#if uniqueService}}
+    <h4 class="text-center"><strong>{{ uniqueService }}</strong></h4>
+  {{/if}}
+
+  <h4 style="margin:15px; font-weight:bold" class="text-center text-uppercase">
     {{date dateFrom "DD/MM/YYYY"}} - {{date dateTo "DD/MM/YYYY"}}
-  </h5>
+  </h4>
```
jniles, line 22: If the currency is not the enterprise currency, we should show a warning that this is only temporary using this warning:

https://github.com/IMA-WorldHealth/bhima/blob/c7dd1f09205a7cd009ff86318c04630828fe1052/server/controllers/stock/reports/stock_expiration_report.handlebars#L36-L40

jniles (commented):
Just two more comments:

1. If the currency isn't the enterprise currency, we should warn about that on the report. 
2. One minor performance win.

Once you fix those feel free to merge this PR.

server/models/procedures/invoicing.sql
```diff
@@ -976,15 +981,19 @@ CREATE PROCEDURE UnbalancedInvoicePaymentsTable(
   -- even though this column is called "balance", it is actually the amount remaining
   -- on the invoice.
 
+  SET exchangeRate = (SELECT IFNULL(GetExchangeRate(_enterpriseId, currencyId, dateTo), 1));
```
jniles, line 984: Beautiful :+1:

jniles (approved):
Looks good to me!

bors r+

PR #6197, opened 2021-12-15
Update server report CSS rules to simplify and unify report title format

jmcameron:
I've updated the default rules in the CSS used by reports to fix up the font sizes for the title lines.  To use these, simply  update the report title and subtitles to use simple 'h2', 'h3', 'h4', and 'h5' without additional 'underline' or 'small' font html tags.  

This PR partially addresses Issue https://github.com/IMA-WorldHealth/bhima/issues/6148.

Here is the way the titles should come out:

  'h2' - the main title  (will be bold)
  'h3' - most subtitles
  'h4' - for a smaller subtitle
  'h5' - for normal font size (eg, descriptions)

I've updated the Inventory > Report > Inventory Change Report to use this convention.  Here is what the code and resulting output look like:
image

image

### DIFF
--- a/server/controllers/inventory/reports/changes.handlebars
+++ b/server/controllers/inventory/reports/changes.handlebars
@@ -4,11 +4,11 @@
   <!-- header  -->
  {{> header}}
 
+  <h2 class="text-center">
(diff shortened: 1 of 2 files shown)


jniles (approved):
LGTM.

bors r+

PR #6198, opened 2021-12-15
Updates for Inventory Changes Report

### LINKED ISSUE #6127: Updates for the Inventory Changes Report
Inventory > Reports > Inventory Changes Report

Suggested updates
- Display currency values with appropriate currency symbols and decimals
- Show a summary row with the total number of changes performed and the sum total change in value
- Allow currency for report to be selected
- For each stock article, shade the row with the stock name with light gray to make the report easier to read

jmcameron:
Updated Inventory Changes Report based on Issue https://github.com/IMA-WorldHealth/bhima/issues/6127:
- Display currency values with appropriate currency symbols and decimals [DONE]
- Show a summary row with the total number of changes performed and the sum total change in value  [NOT DONE]
- Allow currency for report to be selected [NOT DONE, See notes]
- For each stock article, shade the row with the stock name with light gray to make the report easier to read [DONE]

Also did a bit of reformatting to make the report easier to understand.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6127

See:  Inventory > Reports > Inventory Changes Report

### DIFF
--- a/client/src/modules/reports/generate/inventory_changes/inventory_changes.config.js
+++ b/client/src/modules/reports/generate/inventory_changes/inventory_changes.config.js
@@ -2,17 +2,19 @@ angular.module('bhima.controllers')
   .controller('inventory_changesController', inventoryChangesController);
 
 inventoryChangesController.$inject = [
-  '$sce', 'NotifyService', 'BaseReportService',
-  'AppCache', 'reportData', '$state',
+  '$sce', 'NotifyService', 'BaseReportService', 'AppCache',
+  'reportData', '$state', 'SessionService',
 ];
 
-function inventoryChangesController($sce, Notify, SavedReports, AppCache, reportData, $state) {
+function inventoryChangesController($sce, Notify, SavedReports, AppCache, reportData, $state, Session) {
   const vm = this;
   const cache = new AppCache('inventory_changes_report');
   const reportUrl = 'reports/inventory/changes/';
 
   vm.previewGenerated = false;
-  vm.reportDetails = {};
+  vm.reportDetails = {
+    currencyId : Session.enterprise.currency_id,
+  };
 
   vm.clearPreview = function clearPreview() {
     vm.previewGenerated = false;
--- a/server/controllers/inventory/reports/changes.js
+++ b/server/controllers/inventory/reports/changes.js
@@ -44,6 +44,8 @@ async function inventoryChanges(req, res, next) {
   const params = _.clone(req.query);
   const metadata = _.clone(req.session);
 
+  const currencyId = req.session.enterprise.currency_id;
+
   try {
     const report = new ReportManager(TEMPLATE, metadata, params);
     const { dateFrom, dateTo } = params;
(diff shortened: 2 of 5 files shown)


server/controllers/inventory/reports/changes.js
```diff
@@ -44,6 +46,11 @@ async function inventoryChanges(req, res, next) {
   const params = _.clone(req.query);
   const metadata = _.clone(req.session);
 
+  const enterpriseId = req.session.enterprise.id;
+  const { currencyId } = params;
+  const exchangeRate = await Exchange.getExchangeRate(enterpriseId, currencyId, new Date());
```
jniles: I don't think it is appropriate to have an exchange rate in this report, unless you are going to exchange the values with the rate of the _date the value was changed_.  We do this on the account statement report:

image

We then have to mention the rate used for each date.  I'm not sure if it is worth having this feature though.   What do you think?

----

Alternatively, we could just compute the exchanged value when we store the change.  That would mean that anytime we change the value, we would have an exchange value computed to display.  The problem is that if we go back and change the exchange rate, this value would get out of sync with the real exchange rate.

server/lib/template/partials/head.handlebars
```diff
@@ -6,6 +6,30 @@
   <script src="{{nodeModulesPath}}/jsbarcode/dist/JsBarcode.all.min.js"></script>
   <style>
 
+    h2 {
+      font-size : 1.75em;
```
jniles: Is this overridden by #6197?

client/src/modules/reports/generate/inventory_changes/inventory_changes.config.js
```diff
@@ -56,5 +60,10 @@ function inventoryChangesController($sce, Notify, SavedReports, AppCache, report
     if (cache.reportDetails) {
       vm.reportDetails = angular.copy(cache.reportDetails);
     }
+
+    // Set the defaults
+    if (!angular.isDefined(vm.reportDetails.currencyId)) {
```
jniles: It's probably best to just put this on line 15.

jniles (commented):
I'm not sure how much it makes sense for this report to show multicurrency values.  I've made some comments around it - let me know what you think.

server/lib/template/partials/head.handlebars
jmcameron: Yes

server/controllers/inventory/reports/changes.js
jmcameron: Ok.  For now, I've remove multi-currency support.

client/src/modules/reports/generate/inventory_changes/inventory_changes.config.js
jmcameron: Ok

jmcameron:
Removed multi-currency support, per review feedback.

server/controllers/inventory/reports/changes.js
```diff
@@ -44,6 +44,8 @@ async function inventoryChanges(req, res, next) {
   const params = _.clone(req.query);
   const metadata = _.clone(req.session);
 
+  const { currencyId } = params;
```
jniles: This is no longer defined.  It needs to be set to `req.session.enterprise.currency_id` to make it work.  Otherwise, looks good to me.

jniles (changes requested):
@jmcameron one bug fix, then you can merge.

server/controllers/inventory/reports/changes.js
jmcameron: Fixed

jniles (approved):
LGTM.

bors r+

PR #6200, opened 2021-12-16
Increase the precision of the exchange rate

### LINKED ISSUE #6195: Exchange rate needs more significant figures
We are currently storing the exchange rate as a DECIMAL(19,8) UNSIGNED NOT NULL.  Unfortunately, for something like the exchange rate for USD to FC, we store:  1/(FC per USD).   That means there only 8 significant figures after the decimal.  That is not enough in some situations.  For instance, suppose FC is the enterprise currency and we set an exchange rate of 2030 FC/USD:

    1/2030 = 0.0004926108374384236...

When we truncate this to 19.8, we get:

     rate = 0.00049261

Now suppose a value in FC is 123.456.789,1 FC.

Convert to USD using the truncated rate:  $60.816,049

However, the correct value is

     123.456.789,1 / 2030 = 60.816,15

This is not a big error, but since we usually track USD down to cents, the exchange rate should be accurate enough to do that.  Note that the error about $0,10 which is about 200 FC!   To get the error down to the level that the error is less than the equivalent of 1 FC, we need 4 more significant figures.

I suggest converting the SQL definition to at least 22.12.  With that, the cents will be correct for USD conversions and the result will still produce errors less than the equivalent of 1 FC (for this example).  Even more significant figures will be necessary for larger amounts.

Not sure what happens when the numerical values with that precision come out of SQL into our Javascript code and whether that precision is retained in subsequent operations.

jmcameron:
This PR changes the data type for the exchange rate to DOUBLE precision.  This ensures that exchange rate conversions do not loose precision.

It also adds a hint to the exchange rate entry modal:
image

Closes https://github.com/IMA-WorldHealth/bhima/issues/6195

**TESTING**
- Use a database that uses FC as the enterprise currency (eg, Vanga - NOT bhima_test)
- Make sure to apply the migrations.sql in this PR!
- Set the exchange rate for USD :  2030 FC/$ (enter all digits of 1/2030 = 0.00049261084)
- Create a purchase order
   - Select USD as the currency 
   - Select any stock item
   - Enter the unit price as:   3.0
   - Set the quantity to 100000
   - Save the purchase order.  Note the total order is:  $300,000.00
- Confirm the purchase order (via the Purchase Order Registry action menu for that PO)
- Enter stock for for that PO.
  - When entering/creating the LOT, note that the full unit price is shown:  3
  - Accept all 100000 items
  - Verify the Entry document shows $300,000.00 as the total
- Go to the Stock Entries Report page
  - Check the 'Include entries from purchases' and 'Show details' items 
  - View the report in USD and verify the price for that order is $300,000.00
  - View the report in FC.  It should show:  300000 * 2030 = 6.090.000.000,00 FC

NOTE: Before this fix, the last value would have been incorrect because the fractional value of 1/2030 is limited to 8 digits (including leading zeros):

      300000 / 0.00049261 = 609.001.035.302 FC

NOTE: If you want to see the before/after effect of this change. 

- Use the Vanga dataset.  
- Reload the database and the migrations.sql.  Do NOT change the exchange rate yet.  The FC/USD rate should be 2030.   
- Follow the steps above (but not setting a new exchange rate first).  
- Then set the USD/FC exchange rate to its full value: 

     1 / 2030.0 = 0.00049261083744
- Repeat the procedure above.  Then in the Shoe Entries Report you should see both the values with and without this correction.


server/models/migrations/next/migrate.sql
```diff
@@ -25,3 +25,27 @@ DELETE FROM report WHERE report_key IN (
...
+
+/*
+ * @author: jmcameron
+ * @description: Increase precision of exchange rate
+ * @date: 2021-12-16
+ */
+ALTER TABLE `exchange_rate` MODIFY `rate` DOUBLE NOT NULL;
+
+DELIMITER $$
+DROP FUNCTION IF EXISTS `GetExchangeRate`$$
```
jniles: I don't think you need to put this code here - the migration script drops all functions and stored procedures by default and rebuilds them from source.

jniles (commented):
I don't see a downside here.  Tested this and it works.

One thing I would ask while you are working here:

image

Can you make the precision of the "current exchange rate" 4 digits?  It just says $0.00 for any value of FC that is reasonable.

server/models/migrations/next/migrate.sql
jmcameron: Removed.  Thanks for the heads-up.

jmcameron:
> Can you make the precision of the "current exchange rate" 4 digits? It just says $0.00 for any value of FC that is reasonable.

DONE.  It will now print out all available digits.  Minor cosmetic tweaks.

jniles (approved):
Looks good to me.

bors r+

PR #6204, opened 2021-12-17
Handle cashbox configuration error gracefully.

### LINKED ISSUE #6191: Bug in Cash Report getCashboxByAccountId function
The getCashboxByAccountId function in server/controllers/finance/reports/cashReport/index.js has a bug.

If two cashboxes refer to the same account, the query in this function returns two values but db.one() is used, expecting only one.  THis should be fixed or handled more gracefully.   This is not an issue of Cashbox permissions.

To reproduce this error
- Use the Vanga dataset
- Go to the Finance > Report > Cash Report page
- Set the date interval, select either Petite Casse
- Try the preview and see the error.  
- If you try this with one of the Caisse Principal, it works

Looking the SQL database, I see:
```
mysql> select * from cash_box_account_currency;
+----+-------------+-------------+------------+---------------------+
| id | currency_id | cash_box_id | account_id | transfer_account_id |
+----+-------------+-------------+------------+---------------------+
|  1 |           1 |           1 |       3734 |                3735 |
|  2 |           2 |           1 |       3734 |                3735 |
|  3 |           1 |           2 |       3733 |                3749 |
|  4 |           2 |           2 |       3899 |                3749 |
+----+-------------+-------------+------------+---------------------+
4 rows in set (0.00 sec)

mysql> select * from cash_box;
+----+-------------------+------------+--------------+
| id | label             | project_id | is_auxiliary |
+----+-------------------+------------+--------------+
|  1 | Petite Caisse     |          1 |            1 |
|  2 | Caisse Principale |          1 |            0 |
+----+-------------------+------------+--------------+
2 rows in set (0.00 sec)
```
Clearly both Petite caisses refer to the same account so the LEFT join returns more than one row.  

Even if this is a configuration error in the Vanga cashbox setup, it should be handled gracefully by this function and useful error message produced.

jmcameron:
Adds an error message when trying to use misconfigured Cash Boxes.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6191

TESTING
- Use the test procedure given in Issue https://github.com/IMA-WorldHealth/bhima/issues/6191.  This should provide an error message helpful enough to track down and fix the problem.

### DIFF
--- a/client/src/modules/reports/generate/cash_report/cash_report.config.js
+++ b/client/src/modules/reports/generate/cash_report/cash_report.config.js
@@ -2,10 +2,10 @@ angular.module('bhima.controllers')
   .controller('cash_reportController', CashReportConfigController);
 
 CashReportConfigController.$inject = [
-  '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state',
+  '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state', '$translate',
 ];
 
-function CashReportConfigController($sce, Notify, SavedReports, AppCache, reportData, $state) {
+function CashReportConfigController($sce, Notify, SavedReports, AppCache, reportData, $state, $translate) {
   const vm = this;
   const cache = new AppCache('configure_cash_report');
   const reportUrl = 'reports/finance/cash_report';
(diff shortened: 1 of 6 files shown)


client/src/modules/reports/generate/cash_report/cash_report.config.js
```diff
@@ -44,11 +44,19 @@ function CashReportConfigController($sce, Notify, SavedReports, AppCache, report
         vm.previewGenerated = true;
         vm.previewResult = $sce.trustAsHtml(result);
       })
-      .catch(Notify.handleError);
+      .catch(err => {
+        if (err.data.code && err.data.code === 'TOO_MANY_CASHBOXES_PER_ACCOUNT') {
```
jniles, line 48: Nice :+1:

server/controllers/finance/discounts.js
```diff
@@ -12,7 +12,7 @@ const BadRequest = require('../../lib/errors/BadRequest');
 const db = require('../../lib/db');
 
 /**
- * @desc Looks up a discount in the database by its id.  Throws a prmoise error
+ * @desc Looks up a discount in the database by its id.  Throws a promise error
```
jniles, line 15: Good catch 👍

jniles (approved):
Looks good to me.

bors r+

PR #6215, opened 2021-12-24
fix(journal): ensure cost_center_id exists

### LINKED ISSUE #6214: If cost_center_id is NULL, cannot save edits to transactions
A report from Karawa HGR describes an issue they are having with the cost centers reported below:

image

There are no cost centers anywhere here, but the check for cost centers doesn't take into account that these columns might be NULL, and just checks that they are not `undefined`.

https://github.com/IMA-WorldHealth/bhima/blob/4af07b6fbb1185070d22d166302472ef053aac69/client/src/js/services/TransactionService.js#L128-L132

We need to add in a NULL check here.

jniles:
This commit adds in a NULL check to the posting journal's cost center id checks.

Closes #6214.

### DIFF
--- a/client/src/js/services/TransactionService.js
+++ b/client/src/js/services/TransactionService.js
@@ -125,7 +125,7 @@ function TransactionService($http, util, Modal, Accounts) {
         return ERROR_INVALID_DEBITS_AND_CREDITS;
       }
 
-      const hasCostCenters = angular.isDefined(row.cost_center_id);
+      const hasCostCenters = angular.isDefined(row.cost_center_id) && row.cost_center_id !== null;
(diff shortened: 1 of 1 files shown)


client/src/js/services/TransactionService.js
```diff
@@ -125,7 +125,7 @@ function TransactionService($http, util, Modal, Accounts) {
         return ERROR_INVALID_DEBITS_AND_CREDITS;
       }
 
-      const hasCostCenters = angular.isDefined(row.cost_center_id);
+      const hasCostCenters = angular.isDefined(row.cost_center_id) && row.cost_center_id !== null;
```
jmcameron, line 128: Good catch.

jmcameron (approved):
Looks good to me.

PR #6223, opened 2021-12-31
fix: report names with brackets

### LINKED ISSUE #6124: Update many report names
Many report names (in the main menu) include a prefix in square brackets.  Eg:

- Stock > Reports > [Stock] Exit Report

We should remove the square brackets and adjust the report names as needed.   I'm not sure about the history of the square brackets in the report names, but we have some that have them and some that do not.   Here are a few suggested changes as an example:

- [Stock] Compare invoiced items to received stock  => Compare Invoiced Items to Received Stock  
- [Stock] Entries Report => Stock Entries Report
- [Stock] Exit Report => Stock Exits Report
- [Stock] Expiration Report => Stock Expiration Report
- [Stock] Graphical Movements Report => Stock Movements Graphical Report
- [Stock] Inventories => Inventories Report
- [Stock] Monthly consumption => Monthly Stock Consumption Report
- [Stock] Stock Sheet => Stock Sheet Report
- [Stock] Stock value Report => Stock Value Report
- Stock Consumption's graph report => Stock Consumption's Graphical Report

It would also be good to regularize the capitalization:  In English, all words capitalized.  In French, only the first word should be capitalized (unless it is an acronym).

Also, in general, the title of the report should match the corresponding menu item name (where practical)

A final item is whether or not to include the word 'Report' as part of each report name in the main menu.  I've marked this as **Needs Discussion** until we agree on that question.

jniles:
This commit goes through the tree and unit files to fix report names with brackets.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6124


server/models/bhima.sql
```diff
@@ -90,14 +90,14 @@ INSERT INTO unit VALUES
   (203,'Employees Configuration','TREE.PAYROLL_EMP_CONFIGURATION','',57,'/payroll/employee_configuration'),
   (204,'Exchange Rate','TREE.EXCHANGE','',1,'/exchange'),
   (205,'Account Reference Management','TREE.ACCOUNT_REFERENCE_MANAGEMENT','',1,'/account_reference'),
-  (206,'[OHADA] Bilan','TREE.OHADA_BALANCE_SHEET','',281,'/reports/ohada_balance_sheet_report'),
+  (206,'OHADA Bilan','TREE.OHADA_BALANCE_SHEET','',281,'/reports/ohada_balance_sheet_report'),
```
jmcameron, line 93: Don't these changes need to be added to the migrate.sql file too?
I realize that using the translation item (eg, 'TREE.OHADA_BALANCE_SHEET') fixes this in the UI automatically since it is changed in the *.json files.  But if the second item in the 'unit' rows are ignored, why do we have them?

jmcameron (approved):
Looks great! 
Once you address the item I noted in the code, feel free to merge this yourself.

server/models/bhima.sql
jniles: Good question.  We have them so that we know what they are in the database when we are doing `SELECT * FROM unit;`.  Otherwise, there isn't anything that tells us what that unit is.

jniles:
@jmcameron, changes landed.  Thanks for pointing that out!

bors r+

PR #6224, opened 2021-12-31
Add row count to Articles in Stock registry

### LINKED ISSUE #6220: Articles in Stock registry should show count of rows in grid footer
AG Neron has requested that we put the number of rows in the grid footer of the Articles in Stock Registry.  He trains users to check that area to see if they have indeed gotten all the rows they expect or if they need to increase their limit.  He suggests we need to add this data point to the registry for consistency.

jmcameron:
Added row count in the bottom of the Articles in Stock registry table.

Also did minor cosmetic tweaks to the legend at the bottom of the Stock Lots registry.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6220

### DIFF
--- a/client/src/modules/inventory/list/list.js
+++ b/client/src/modules/inventory/list/list.js
@@ -248,7 +248,7 @@ function InventoryListController(
   function openColumnConfigModal() {
     gridColumns.openConfigurationModal();
   }
-  // delete an invetory from the database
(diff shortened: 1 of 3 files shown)


jniles (approved):
image

This is what it looks like.   Well done!

bors r+

Issue #6062, opened 2021-10-28
Feedback on System Usage Statistics


jniles:
We have received the following feedback about our system usage statistics report:

Can we put the totals under each column and the current report?

The question I am asking of this data: Do we start registering and having 
patient pay prior to 8AM or does our team delay registering patient and 
receiving payments until after 8AM.  The delay at the registration and 
the payment window delays the rest of the work to see and treat our 
patients efficiently. 

7h | 3 | 4 | 6 | Dimanche 24 Octobre   2021
-- | -- | -- | -- | --

-----

I believe that we already have the totals on the report already.  So we'll just need to find a way to make this data more easily accessible to the MD.

jmcameron:
It is not clear what is desired here.   Here is an example System Usage Statistics output:
![image](https://user-images.githubusercontent.com/62145/140969361-c7556e5f-b2df-4516-8dab-efc27d028415.png)

Also, I found it very odd that there were stock movements from midnight through 6am.  Is these times real?  I suspect these are not the real times so why are these odd times occurring?

jmcameron:
> Also, I found it very odd that there were stock movements from midnight through 6am. Is these times real? I suspect these are not the real times so why are these odd times occurring?

I checked the code and the SQL for times for all this data is based on SQL table 'created_at' fields which are created automatically when the rows are created.  So these should depend only on the server computer system time.

jmcameron:
I did some tests with a local dev sandbox and had no problem with the created_at date/times corresponding to when the entries occurred and the System Usage Statistics page is correct also.    So it seems likely that this has to do with some issue of the computer system time on the server - or how the database archives are created that we can download and use for testing.

jmcameron:
For reasons I do not understand, the created_at timestamps in production data that I was able to download do not correspond to the same data at the production site.    I requested that the System Usage stats be generated on-site for the same day I did above (in https://github.com/IMA-WorldHealth/bhima/issues/6062#issuecomment-964346494) and received this:
![image](https://user-images.githubusercontent.com/62145/141167772-50bd9278-58c6-4694-8182-3a681a1b89b8.png)

Notice that there are no entries before 8am.   So perhaps the process of uploading production date from the site to the cloud is that the created_at data is not being transmitted but is being constructed on the "cloud" side.  The difference in the listed columns is because the production site is running an older version of BHIMA.

jmcameron:
[...]

PR #6225, opened 2021-12-31
Display and export created_at timestamps in Stock Movements

### LINKED ISSUE #6221: Export created_at timestamp in Excel and CSV logs
AG Neron was trying to look at the "promptitude" of stock movements.  The way that the BHIMA team did this is by using the `created_at` field it the database, but we never expose it to the user.  It probably doesn't make sense for a regular use case, but in Excel/CSV reports, we should provide it as the "server time" of the operation.  This will allow administrators to see if data has been back-dated.

jmcameron:
Added ability to show the created_at time (shown as "Server Date") in the Stock Movements registry.
Also added tooltips for the date cell and the Server Date (create_at) cell that shows the full time and date (not just the calendar day).

NOTE:  Try this with the Vanga database and notice the large difference between the date and the server date.  It may be necessary to export a more complete time string that includes the time zone.   Also, please review https://github.com/IMA-WorldHealth/bhima/issues/6062 and decide how we should handle this.  Note that this PR could be accepted as is and we can add another issue to fix the export date string (which would probably apply to more than this page).

Closes https://github.com/IMA-WorldHealth/bhima/issues/6221

**TESTING**
- Use any database
- Go to the Stock Movements page
- Note that the "Server Date" is not showing.  Use the Menu->Column modal to enable its display.
- Hover over a Date cell and a Server Date cell to see the full time.
- Try exporting the page and verify that the server date gets exported.

### DIFF
--- a/client/src/modules/stock/movements/registry.js
+++ b/client/src/modules/stock/movements/registry.js
@@ -109,8 +109,14 @@ function StockMovementsController(
       type : 'date',
       displayName : 'FORM.LABELS.DATE',
       headerCellFilter : 'translate',
-      cellFilter : 'date',
-      cellClass : 'text-right',
+      cellTemplate : 'modules/stock/movements/templates/date.cell.html',
+    }, {
+      field : 'created_at',
+      type : 'date',
+      displayName : 'FORM.LABELS.SERVER_DATE',
+      headerCellFilter : 'translate',
+      cellTemplate : 'modules/stock/movements/templates/created_at.cell.html',
+      visible : false,
     }, {
       field : 'userName',
       displayName : 'FORM.LABELS.USER',
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -726,6 +726,7 @@
             "SELECT_ONE": "Select One",
             "SELECT": "Select",
             "SENIORITY_BONUS": "seniority bonus",
+            "SERVER_DATE": "Server Date",
             "SERVICE": "Service",
             "SERVICE_OPTIONAL": "Service (optional)",
             "SERVICES": "Services",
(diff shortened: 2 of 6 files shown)


client/src/modules/stock/movements/registry.js
```diff
@@ -109,8 +109,14 @@ function StockMovementsController(
...
-      cellFilter : 'date',
-      cellClass : 'text-right',
+      cellTemplate : 'modules/stock/movements/templates/date.cell.html',
+    }, {
+      field : 'created_at',
+      type : 'date',
+      displayName : 'FORM.LABELS.SERVER_DATE',
+      headerCellFilter : 'translate',
+      cellTemplate : 'modules/stock/movements/templates/created_at.cell.html',
+      visible : false,
```
jniles, line 119: Nice :+1:  I am glad `visible` is false by default.  I think that makes the most sense here.

jniles (approved):
Tested and it looks like this is working.  Code looks good! 

LGTM.

bors r+

jniles:
Ah... I remember this.

jniles:
Regarding

> NOTE: Try this with the Vanga database and notice the large difference between the date and the server date. It may be necessary to export a more complete time string that includes the time zone. Also, please review #6062 and decide how we should handle this. Note that this PR could be accepted as is and we can add another issue to fix the export date string (which would probably apply to more than this page).

There are two factors that go into this - we record the date as a timestamp for when the stock movement was started, not submitted - that explains the minutes/seconds drift.  Secondly, the created_at timestamp seems to be stored in UTC, whereas the other one is UTC+1, so they are always off by an hour.  I'm not sure why that is.  It looks like MySQL may be storing UTC by default.

PR #6227, opened 2022-01-02
fix(reports): fix profit/loss statement

### LINKED ISSUE #6180: Updates for the Profit And Loss Statement report
Finances > Report > Profit And Loss Statement

Not sure if this report is working.  I tried the Vanga dataset and bhima_test with various period ranges and all the numbers are zero.  Not sure if this a bug or some configuration issue.  It should probably be tested with other production datasets and debugged if they show zeros too.

**Suggestions**
- Increase the font size of the date range on the report
- The format of the rest of the report looks fine

jniles:
This commit fixes the profit and loss statement by correctly using the Tree library.

Closes #6180.

### DIFF
--- a/server/controllers/finance/reports/income_expense/index.js
+++ b/server/controllers/finance/reports/income_expense/index.js
@@ -35,98 +35,96 @@ const DEFAULT_PARAMS = {
 // expose to the API
 exports.document = document;
 
-function document(req, res, next) {
-  let report;
-
+async function document(req, res, next) {
   const options = _.defaults(req.query, DEFAULT_PARAMS);
 
   try {
-    report = new ReportManager(TEMPLATE, req.session, options);
+    const report = new ReportManager(TEMPLATE, req.session, options);
+    const { periodFromId, periodToId, fiscalYearId } = options;
+    const data = {};
+
+    const [periodFrom, periodTo, fiscalYear] = await Promise.all([
+      getPeriodById(periodFromId),
+      getPeriodById(periodToId),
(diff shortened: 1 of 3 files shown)


server/controllers/finance/reports/income_expense/index.js
```diff
@@ -35,98 +35,96 @@ const DEFAULT_PARAMS = {
...
+    const { periodFromId, periodToId, fiscalYearId } = options;
+    const data = {};
+
+    const [periodFrom, periodTo, fiscalYear] = await Promise.all([
+      getPeriodById(periodFromId),
+      getPeriodById(periodToId),
+      Fiscal.lookupFiscalYear(fiscalYearId),
+    ]);
+
+    if (periodFrom.start_date > periodTo.start_date) {
```
jmcameron, line 52: I wonder if this check could be incorporated into bh-fiscal-period-select (in another PR)?

server/controllers/finance/reports/income_expense/report.handlebars
```diff
@@ -1,4 +1,4 @@
-{{> head title="REPORT.PROIFT_AND_LOSS" }}
+{{> head title="REPORT.PROFIT_AND_LOSS" }}
```
jmcameron, line 1: 👍

jmcameron (approved):
Looks great!

server/controllers/finance/reports/income_expense/index.js
jniles: This is a good idea.  I see you made an issue #6233

PR #6229, opened 2022-01-02
feat: initial docker-compose support

### LINKED ISSUE #3439: Deploy BHIMA using Docker
We are approaching a point where we will be doing a lot of ops work - maintaining servers, rolling out updates across multiple institutions, so on.  We could benefit from the ease of maintenance of docker, as well as the ease in deploying online.  New contributors could also spin up instances rather easily.

I've never used Docker before, and the initial setup is a bandwidth hog for the Congo context, but these problems are surmountable.  I say we give it a whirl and see what comes of it.

jniles:
Here, we have created an initial docker-compose file that links together MySQL, Redis, and builds BHIMA.  In order to get it all working, I needed to make the REDIS_HOST into an environmental variable.

There are still some things that needs to done:

1. We need to have a better way of building the MySQL database.  Currently, the user has to run `yarn build:docker` to create an initial database file to populate the database.
2. We should be using multi-stage builds to ensure that we only have build artifacts in a "build" image and then deploy the most minimal image one can think of.
3. We need documentation on how to deploy/run with `docker` & `docker-compose`.

Closes #3439.

----

To get this working, you will need to have both [`docker`](https://docs.docker.com/engine/install/) and [`docker-compose`](https://docs.docker.com/compose/install/) installed.  Please follow those instructions for your operating system.  I've tested this on Ubuntu 20.04.

Then, please execute the following commands in the root directory of the BHIMA repository.

1. Add the `REDIS_HOST` to your `.env` file.  You can set it to `127.0.0.1` for running locally.  Docker will automatically change this to the appropriate address based on the docker-compose file.
2. `yarn build:docker` - this command will concatenate all the database files into a single SQL file.
3. `docker-compose up --build` - this command will create the BHIMA application, bring up MySQL and Redis, and link them all together in their own internal network.

The above command will give you the logs from the containers.  You can add the `-d` flag to docker to detach if you want to run in the background.

Once the containers are all up and running, the application should be available on localhost on the `$PORT` you have in your `.env` file.

Voila!  You should have the BHIMA installation page available.

It is probably useful to read through [dockerizing a nodejs app](https://nodejs.org/en/docs/guides/nodejs-docker-webapp/) for context.


jniles:
Next steps after this PR gets reviewed, tested, and approved:

1. Build and push a BHIMA image to [Docker Hub](https://hub.docker.com/).  This will allow our compose file to simply pull from the BHIMA docker image.
2. Automate the building of BHIMA images on release.  See [this resource](https://docs.github.com/en/actions/publishing-packages/publishing-docker-images) for how to do it automatically with github actions.
2. Update documentation to show how to deploy with docker.
3. Create scripts to dump the mysql database from the docker image associated with docker compose to a local file, so that we can still use `syncthing` for managing our backups.
4. Use staged builds to shrink the size of the BHIMA image.  Potentially investigate using `node:alpine` or `node:slim` to run BHIMA without 
5. Create separate Dockerfiles for `development` and `production`.

jniles:
Hmm... one easy improvement actually would be to set `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` and install chromium with the rest of the OS dependencies.  This might actually be a better way since then docker can cache that easier than the `node_nodules` version.

jmcameron:
During installation of docker and docker-compose on my Ubuntu LTS 20.04 (under WSL on Windows 10), I ran int a problem when running 'docker-compose up --build'.  It complained that it could not find 'docker-credential-desktop.exe'.   After a bit of googling, I found a fix.   In ~/.docker/config.json, fix the spelling of 'credsStore' to 'credStore'.  Then the docker-compose build runs.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN yarn --frozen-lockfile
+
+# Copy all the source code from host machine to the container project directory
+COPY . .
+
+RUN yarn build
+
+COPY .env bin/
+
+# change directory to the bin diretory
```
jmcameron, line 32: Thanks for the documentation.  It is really important to document your Dockerfile to indicate what changes are happening outside the container vs inside the container!

jmcameron (commented):
I tested this in Ubuntu LTS 20.04 inside WSL on Windows 10 and it worked.  I had a minor problem with installing docker-composed (see https://github.com/IMA-WorldHealth/bhima/pull/6229#issuecomment-1004205177).  One I fixed that, running 'docker-compose up --build' worked (it took about 6 minutes).  Then I was able to access Bhima via web broser, but got the "Basic System Installation" setup page. 
- We need scripts and documentation on how to run Bhima with specific production site data (without doing the Base System Installation).
- Restarting the container was fast (10+ seconds)
- Stopping the container was difficult.  ^C stopped it partially, but I eventually had to suspend and manually kill the process.
- With default configurations, building a docker container for Bhima requires multiple GB of data saved to your local computer.  It is probably worthwhile adding notes about where those files are stored and how to prune/delete the docker containers and images involved.  On my system, it looks like it required over 4GB of docker data to create the docker images and containers.

jniles:
By the way, the "import default ohada accounts" errors out on the fresh docker install (I haven't tried yet with the regular install).

image

We might need to adjust our collations somewhere.

```bash
bhima_1  | 2022-01-04T08:25:09.007Z db:errors #interceptor(): [SQL-QUERY] CALL ImportAccount(1, 1021, 'Dotation initiale', 'title', 102, 1);
bhima_1  | 2022-01-04T08:25:09.007Z db:errors #interceptor(): [SQL-RESPONSE] ER_CANT_AGGREGATE_2COLLATIONS
bhima_1  | 2022-01-04T08:25:09.007Z db:errors #interceptor(): [SQL-MESSAGE] Illegal mix of collations (utf8mb4_unicode_ci,IMPLICIT) and (utf8mb4_0900_ai_ci,IMPLICIT) for operation '='
```

jniles:
Okay, looks like the OHADA account import works on the `master` branch, so this is a bug with the mysql docker configuration on this branch.

jniles:
@jmcameron can you give this another whirl?  You may need to first type:


```sh
docker-compose down -v
```

to clear out old data.

jmcameron (approved):
It looks good to me.   It ran okay and I was able to see the installation page in my browser.  However, I still had difficulty stopping the docker container.  I did not see the OHADA errors.   I'll attach my log for the run so you can take a look at it.  I noticed several warnings, but they seemed harmless.
[docker-jmc-2022-01-06.log](https://github.com/IMA-WorldHealth/bhima/files/7823201/docker-jmc-2022-01-06.log)

jniles:
Awesome, thanks for the feedback.  It does take a little while to kill the containers, but should still be under a minute.  It could be something strange with your WSL+Windows setup, but I'll save investigating that until we've gotten more fleshed out here.

Another thing of interest: [A tutorial on optimizing docker container size](https://contains.dev/blog/optimizing-docker-image-size).

jniles:
Okay, fixed the charset problem. :+1:

PR #6237, opened 2022-01-04
fix(csp): turn useDefaults off in CSP

jniles:
This commit turns the cross-site protection `useDefaults` directive to `false`.  It was flipped to being `true` by default with the last release of `helmet` and wreaks havoc with our app.

See [the helmetjs changelog](https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md) for more information.


jmcameron (approved):
Looks good!

This fixes the cross-site issue.

PR #6239, opened 2022-01-04
fix(reports): require depots for stock changes

### LINKED ISSUE #6154: Stock Changes Report is broken
Stock > Report > Stock Changes Report

BUG!   This report crashes for me with the server-side error message: 
- no such file or directory, open  /Bhima/node_modules/chart.js/dist/Chart.min.js'
- In the node_modules directory, the file name is chart.min.js (lower case)
- Fixing this case name in server/controllers/stock/reports/stock_movement_report.handlebars does helps but there is still an error:  TypeError: Cannot read property 'toLowerCase' of undefined

jniles:
Makes the depot field a required field for viewing the stock changes report.

Closes #6154.

image

### DIFF
--- a/client/src/modules/reports/generate/stock_changes/stock_changes.html
+++ b/client/src/modules/reports/generate/stock_changes/stock_changes.html
@@ -42,8 +42,8 @@
             <!-- select depot -->
             <bh-depot-select
               depot-uuid="ReportConfigCtrl.reportDetails.depot_uuid"
-              on-select-callback="ReportConfigCtrl.onSelectDepot(depot)">
-              <bh-clear on-clear="ReportConfigCtrl.clear('depot_uuid')"></bh-clear>
+              on-select-callback="ReportConfigCtrl.onSelectDepot(depot)"
+              required="true">
             </bh-depot-select>
 
             <!-- preview -->


client/src/modules/reports/generate/stock_changes/stock_changes.html
```diff
@@ -42,8 +42,8 @@ <h3 translate>REPORT.STOCK_CHANGES.TITLE</h3>
             <!-- select depot -->
             <bh-depot-select
               depot-uuid="ReportConfigCtrl.reportDetails.depot_uuid"
-              on-select-callback="ReportConfigCtrl.onSelectDepot(depot)">
-              <bh-clear on-clear="ReportConfigCtrl.clear('depot_uuid')"></bh-clear>
+              on-select-callback="ReportConfigCtrl.onSelectDepot(depot)"
+              required="true">
```
jmcameron, line 46: Good catch!

jmcameron (commented):
I tested this and it does now produce a report, so the fix helps.

However, when I use the Vanga dataset, and tried the "Stock Changes Report" FY 2021, December 2021, either depot), i see that these columns are empty:
- Quantity Consumed-
- Quantity Lost
- Quantity Remaining

So I'm not sure it is working correctly.

jniles:
So .. we need a better name for this.  It isn't actually a report.  It's a document.

There will never be data in this report - it is meant to be printed out and recorded manually with pen/paper, then later used as a template for data entry for the "Aggregate Stock Consumption" module.  So, what you are seeing is intended behavior.

I'm hoping we can sunset this completely soon because (1) it isn't working in production and (2) using ODK offline ( https://github.com/IMA-WorldHealth/bhima/issues/6235) seems like a much better solution.

jniles:
Given that explanation, I'll merge this now (to fix the bug) and if we want to look into renaming, we should do that in a separate PR.

bors r+

PR #6240, opened 2022-01-04
fix(stock sheet): fixup language and rendering

### LINKED ISSUE #6149: Updates for [Stock] Stock Sheet report
Stock > Report > [Stock] Stock Sheet

**Suggestions**
- ~~Fix name:  [Stock] Stock Sheet => Stock Sheet Report~~
- What is the purpose of the "Sort lots movements by their true creation date" option?  Is it necessary?  Does something need fixing to eliminate the need for the option.  (BTW, see Issue https://github.com/IMA-WorldHealth/bhima/issues/6062 and closed PR https://github.com/IMA-WorldHealth/bhima/pull/6094).
- On the resulting report, the legend in the summary line "End value of the stock with the unit cost of the company" should be right justified.  Similarly the legend "Opening Balance" in first row should be right justified.  Also, the term "Stock" in the right most group of columns is ambiguours.  I would suggest something like "Stock Available" or "Stock on hand"  image

jniles:
Right aligns certain values and rewords titles to make it more explicit.
Also reorders the options so that they flow from required to optional.

Closes #6149.

See below:

image

image

### DIFF
--- a/client/src/i18n/en/report.json
+++ b/client/src/i18n/en/report.json
@@ -305,7 +305,7 @@
       "VALUE": "Value",
       "ENTRIES": "Entries",
       "EXITS": "Exits",
-      "STOCKS": "Stocks",
+      "STOCK_ON_HAND": "Stock on Hand",
       "EXIT_REPORT": "Stock Exit Report",
       "EXIT_REPORT_DESCRIPTION": "This report shows stock exits which had occurred in different depots",
       "INCLUDE_AGGREGATE_CONSUMPTION": "Include aggregate consumption",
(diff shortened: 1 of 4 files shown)


client/src/i18n/en/report.json
```diff
@@ -305,7 +305,7 @@
       "VALUE": "Value",
       "ENTRIES": "Entries",
       "EXITS": "Exits",
-      "STOCKS": "Stocks",
+      "STOCK_ON_HAND": "Stock on Hand",
```
jmcameron, line 308: Thanks.  That is clearer

server/controllers/stock/reports/stock_sheet.report.handlebars
```diff
@@ -38,7 +38,7 @@
             <th colspan={{#if depot.text}}"4"{{else}}"5"{{/if}}></th>
             <th class="text-center" colspan="3">{{translate 'REPORT.STOCK.ENTRIES'}}</th>
             <th class="text-center" colspan="3">{{translate 'REPORT.STOCK.EXITS'}}</th>
-            <th class="text-center" colspan="3">{{translate 'REPORT.STOCK.STOCKS'}}</th>
+            <th class="text-center" colspan="3">{{translate 'REPORT.STOCK.STOCK_ON_HAND'}}</th>
```
jmcameron, line 41: 👍

jmcameron (approved):
It looks great. 

I tested it and the report is clearer to me.

PR #6241, opened 2022-01-04
fix(cash): fix incorrect error msg key

jniles:
Fixes the error message i18n key for the cash payments form when the user hasn't selected any invoices.  The previous key didn't point to anything.


jmcameron (approved):
Good catch!

jmcameron (approved):
Looks good.  

I tested it and I see the message now.

PR #6242, opened 2022-01-04
fix: remove USD cashbox from Vanga db

### LINKED ISSUE #6238: Remove redundant auxilliary cashboxes from production sites
Vanga only deals with FC in it's Caisse Auxiliaire,  At one time, it was configured with FC for the USD account, but this was never used. We should create a patch to remove it from the Vanga database.

jniles:
Adds a patch SQL file to remove the USD cashbox from the vanga database.

Closes #6238

### DIFF
--- a/server/models/migrations/next/fix-vanga-usd-aux-cashbox.sql
+++ b/server/models/migrations/next/fix-vanga-usd-aux-cashbox.sql
@@ -1,0 +1,6 @@
+/**
+author: @jniles
+date: 2022-01-04
+description: removes the USD Caisse Auxiliare from Vanga
+*/
+DELETE FROM cash_box_account_currency WHERE id = 2;


server/models/migrations/next/fix-vanga-usd-aux-cashbox.sql
```diff
@@ -0,0 +1,6 @@
+/**
+author: @jniles
+date: 2022-01-04
+description: removes the USD Caisse Auxiliare from Vanga
+*/
+DELETE FROM cash_box_account_currency WHERE id = 2;
```
jmcameron, line 6: 👍

jmcameron (approved):
I tested this with the Vanga dataset and it worked perfectly.  I was able to do the Cashbox report for the Pettite Caisse Fc (which currently fails in master).

jmcameron:
**NOTE:** This could have been put into server/models/migration/next/migrate.vanga.sql and been included automatically in the file produced by 'yarn migrate'.   See the next to last bullet under 'Migrating Test Environents for Production sites' in the Wiki (https://github.com/IMA-WorldHealth/bhima/wiki/Releases-and-Migration-Notes)

jniles:
I think it will .. doesn't the regex just do `*${database}*.sql`?  I didn't realize that the best format was `migrate.vanga.sql`  I'll do that next time!

jmcameron:
> I think it will .. doesn't the regex just do `*${database}*.sql`? I didn't realize that the best format was `migrate.vanga.sql` I'll do that next time!

You are right.  So what you did should work.   You may want to check the resulting migration file when upgrading Vanga to be sure.

PR #6247, opened 2022-01-05
Update RUMER report

### LINKED ISSUE #6153: Updates for RUMER report
Stock > Report > RUMER

**Suggestions**
- It seems like a more descriptive name would be helpful such as:  Drug Usage Report (RUMER).  In French "Registre d’utilisation des médicaments et recettes (RUMER)"?
- The font sizes for the title block need to be neatened up. 

image

- The English subtitle has the word "Recipes".  This is wrong.  The French subtitle is "**R**egistre d’**u**tilisation des **m**édicaments **e**t **r**ecettes" (which seems to be where RUMER comes from).  This should translate in English to something like:  "Medication Use Records and Receipts" (eg, MURAR!)
- The flag "Exclude inventories without movements and out of stock" does not prevent displaying stock with (erroneous) negative quantities.
- Suggest renaming the flag "Include daily stock balances and out-of-stock percentage" to "Include daily stock balances and out-of-stock percentage details"

jmcameron:
Updates the stock RUMER report

Closes https://github.com/IMA-WorldHealth/bhima/issues/6153

Includes these fixes:

- Updated English description to:  Drug Usage and Entries Report
- Modified report title to include "RUMER" and description on the same line
- Minor formatting adjustments
- Improved label wording for "Include daily balances" option
- Fixed logic for "Exclude inventories without movements and out of stock".  Now it will exclude any stock that has no movements for the month AND is out of stock.  Also notice that negative values are treated as zero now.

**TESTING**
- Use the Vanga dataset
- Stock > Reports > RUMER
- Try the various report options
- To verify the logic for the exclude option, try PHARMACIE USAGE, FY 2021, February 2021
   - Without the exclude option checked, look for entries for 
      - Captopril (Tablets USP) 25 mg Ces
      - Vitamine E de 200 mg Ces  (near the end)
   - Without the option, you should see both of these
   - With the exclude option checked, these should not appear


client/src/i18n/fr/report.json
```diff
@@ -181,7 +181,7 @@
       "DESCRIPTION": "Registre d’utilisation des médicaments et recettes",
       "ENTRIES": "Entrées",
       "EXCLUDE_OUT_STOCK_ITEMS": "Exclure les inventaires sans mouvements et en rupture de stock",
-      "INCLUDE_DAILY_BALANCES": "Inclure les soldes quotidiens des stocks et le pourcentage de rupture de stock",
+      "INCLUDE_DAILY_BALANCES": "Inclure les détails du bilan quotidien des stocks",
```
jniles, line 184: @mbayopanda, can you review this translation?

server/controllers/stock/reports/stock/rumer.js
```diff
@@ -21,8 +21,8 @@ const DEFAULT_PARAMS = {
  * @method report
  *
  * @description
- * This method builds the RUMER (Drug Usage and Recipe Register) report by month JSON, PDF, or HTML
- * file to be sent to the client.
+ * This method builds the RUMER (Registre d’Utilisation des Médicaments Et Recettes) report
```
jniles, line 24: 👍

jniles (approved):
Looks good to me.  I've tagged @mbayopanda to review the French since I'm not sure it looks right to me, but cannot make a proposal.

client/src/i18n/fr/report.json
mbayopanda: It's okay 👍🏽

mbayopanda (approved):
LGTM 👍🏽

PR #6249, opened 2022-01-05
Updates for the stock expiration report

### LINKED ISSUE #6144: Updates for the Stock Expiration Report
Stock > Report > Stock Expiration Report

**Suggestions**
- The currency should default to the enterprise currency
- Include the inventory codes
- It looks like the report shows the expiring stock in order by how soon it will expire.  This makes sense.  However, I wonder whether this should be an option:   Alphabetical vs Date of Expiration?
- When you chose a currency other than the enterprise currency, it adds this message to the report:  "Attention! You are using a currency other than the enterprise currency. Values on this report are subject to change with changes to the exchange rate and should not be considered final. Please save a report in the enterprise currency for long term reference."   I wonder if we should include this on other reports?  (i18n token:  WARN_CURRENCY)

jmcameron:
Updated the stock expiration report

Stock > Reports > Stock Expiration Report

Closes https://github.com/IMA-WorldHealth/bhima/issues/6144

**Updates**
- Now defaults to enterprise currency
- Added inventory code to the listing for reference
- Minor formatting changes

### DIFF
--- a/client/src/modules/reports/generate/stock_expiration_report/stock_expiration_report.config.js
+++ b/client/src/modules/reports/generate/stock_expiration_report/stock_expiration_report.config.js
@@ -12,7 +12,7 @@ function StockExpirationReportConfigCtrl($sce, Notify, SavedReports, AppCache, r
   const reportUrl = 'reports/stock/expiration_report';
 
   // default values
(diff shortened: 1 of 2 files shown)


mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #6254, opened 2022-01-07
Fixed missing code in previous update

jmcameron:
Fix an error in PR https://github.com/IMA-WorldHealth/bhima/pull/6243.
Somehow, code that was in the handlebars file got lost in the version that got released.  This fix adds that code back in.

**TESTING**
- Use production database
- Try the report: Finance > Report > Open Debtors
   - Try it in the enterprise currency (default), and note a few numbers including the final total at the bottom of the report.   Also, no exchange rate should be listed under the title.
   - Try it with some other currency and verify that the new numbers are correct.  The exchange rate used should be listed at the top of the report (under the title).


jniles (approved):
Great catch, thanks!

bors r+

PR #6266, opened 2022-01-11
improvement(Repport Analysis of Cashbox)

### LINKED ISSUE #6169: Updates for the Analysis Of Auxiliary Cashboxes report
Finance > Report > Analysis Of Auxiliary Cashboxes

**Suggestions**
- Fix name.  Since only one cashbox is shown at a time in the report, the report title should be singular:  
  Analysis of Auxiliary Cashbox  (leaving the menu item in plural is okay).
- Is it possible to move the legend above the title?  Note that items like this are usually labeled as "Legend".
image
- What goes into the "Target" column?  Is it necessary?
- Is it always the case that each cash box only handles one currency?  Is there any need for multiple currency support (eg, show the FC cashbox in $ which would need a currency selection in the client side).

lomamech:
- Use Singular for the name of report
- Set a legend above the title
- Update data.sql fix combinaison of account - currency and cashbox

closes #6169


lomamech:
image

With the OHADA system, it is necessary to be able to use the fund transfer account, wherever target would be null, this would mean that the funds from the auxiliary fund did not reach their destination, we can also see if the Accountant used the wrong account to collect funds

lomamech:
@jniles @jmcameron for review

jmcameron (changes requested):
~~I had difficulties with this PR.  I could not get the Analysis of Aux Cashboxes report to work.  It looks to me like the attempt to rename 'cashbox' to 'cashboxes' was incomplete.  Also verify that any necessary updates were done to the unit and report tables in bhima.sql and the migration.sql file.~~

Oops!  I forgot to rebuild the database.  It actually worked, only leaving the comment below

I did eventually fix enough things to get it working and saw the report.  Regarding the legend, I was hoping to get it on the top right.   If that is not desired, then leave the legend below the titles (as it was before).   I've attached a version of the report.handlebars template file that does what I suggest.  Please try it and use it if you wish:
[report.handlebars.txt](https://github.com/IMA-WorldHealth/bhima/files/7847853/report.handlebars.txt)
.

jmcameron (commented):
After rebuilding the database, it works in bhima_test.  Did not test with a production database.  Otherwise it looks good.  However, please see my comment about location of the legend in https://github.com/IMA-WorldHealth/bhima/pull/6266#pullrequestreview-849366638.

jmcameron (dismissed):

jmcameron (commented):
Trying to fix my previous review.  Please refer to my comments inhttps://github.com/IMA-WorldHealth/bhima/pull/6266#pullrequestreview-849388959.

lomamech:
image

Last version

client/src/modules/reports/reports.routes.js
```diff
@@ -9,7 +9,7 @@ angular.module('bhima.routes')
       'aged_creditors',
       'aged_debtors',
       'aggregated_stock_consumption',
-      'analysis_auxiliary_cashboxes',
+      'analysis_auxiliary_cashbox',
```
jniles, line 12: This breaks the report key at IMCK.  You'll need to write a migration script that updates the `report` table  and the key in the `unit` table:

image

jniles (changes requested):
Please add these two lines to the `migrate.sql` file:
```sql
update report set report_key = 'analysis_auxiliary_cashbox' where report_key = 'analysis_auxiliary_cashboxes';
update unit set path = '/reports/analysis_auxiliary_cashbox' where path = '/reports/analysis_auxiliary_cashboxes';
```

You'll also need to fix the translation keys. 

Is this report only for OHADA institutions?  It seems like Vanga doesn't have any transfers into the principal cashbox:

image

If so, we should include that in the description, that this report shows transfers out of the auxiliary into a tertiary account then into the principal cashbox.

client/src/modules/reports/reports.routes.js
lomamech: Thanks for noticing

lomamech:
> Please add these two lines to the `migrate.sql` file:
> 
> ```sql
> update report set report_key = 'analysis_auxiliary_cashbox' where report_key = 'analysis_auxiliary_cashboxes';
> update unit set path = '/reports/analysis_auxiliary_cashbox' where path = '/reports/analysis_auxiliary_cashboxes';
> ```
> 
> You'll also need to fix the translation keys.
> 
> Is this report only for OHADA institutions? It seems like Vanga doesn't have any transfers into the principal cashbox:
> 
> image
> 
> If so, we should include that in the description, that this report shows transfers out of the auxiliary into a tertiary account then into the principal cashbox.

After analyzing this report requires the use of a transfer account, it will not really be very useful in the event that one will not use the transfer account, For systems that do not use the transfer account, this report will only track the daily movements of the auxiliary fund.

jniles (approved):
This looks good to me.

bors r+

PR #6267, opened 2022-01-11
improvement(Voucher Registry)

### LINKED ISSUE #6262: Weird total in footer on vouchers registry
What do you suppose this total is for?  Are we summing the transaction type_ids of the vouchers?

image

If so, we should just remove this code.

lomamech:
- Remove not necessary summing in column type_id and amount

closes #6262


lomamech:
image

It is not necessary to sum the amount just because they are different in nature and are different currencies

jniles (approved):
LGTM.

bors r+

PR #6268, opened 2022-01-11
Add server-side function to nicely show exchange rate in reports

### LINKED ISSUE #6248: Create bhShowExchangeRate component to use in various reports
This component would display the specified exchange rate in a smart way to make it easy to display in various reports when the enterprise currency is not used.   For an example of how this might work, see:
- [server/controllers/finance/reports/debtors/openDebtors.js](https://github.com/IMA-WorldHealth/bhima/blob/c6b605c8fc42c032be83174a44f5123b4c610bc5/server/controllers/finance/reports/debtors/openDebtors.js)
- [server/controllers/finance/reports/debtors/openDebtors.handlebars](https://github.com/IMA-WorldHealth/bhima/blob/c6b605c8fc42c032be83174a44f5123b4c610bc5/server/controllers/finance/reports/debtors/openDebtors.handlebars#L15)

This could also be extended to automatically display the WARN_CURRENCY message.

Ideally this would only require the currency ID (and an optional date) as inputs.

jmcameron:
Added `exchangeRate` handlebars partial function to simplify nicely showing the current exchange rate and a warning message if it is not the enterprise currency.

Updated several reports to the new functions:
   - [ ] Cost Centers > Reports > Cost Centers Income and Expense
   - [ ] Finance > Reports > Annual Clients Report
   - [ ] Finance > Reports > Monthly Balance (monthly_balance)
   - [ ] Finance > Reports > Open Debtors
   - [ ] Finance > Reports > Unpaid Invoices Report *
   - [ ] Human Resources > Reports > Employee Standing Report (all employees)
   - [ ] Human Resources > Reports > Employee Standing Report (single employee)
   - [ ] Stock > Reports > Average Medication Costs per Patient
   - [ ] Stock > Reports > Lost Stock Report (lost_stock_report)
   - [ ] Stock > Reports > Stock Entries Report
   - [ ] Stock > Reports > Stock Exits Report
   - [ ] Stock > Reports > Stock Expiration Report
   - [ ] Stock > Reports > Stock Sheet
   - [ ] Stock > Reports > Stock Value Report (stock_value)

**Note:** You can use the check-mark boxes above to keep track on which ones you have tested! Click on one and see!

Other reports were not updated for various reasons, such as they already had related functionality.

This PR also makes cosmetic formatting changes to some of the reports above.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6248

**TESTING**
- Use bhima_test and then a production dataset that uses Fc as the enterprise currency (eg, Vanga)
- For each dataset, try each of the reports above with the enterprise currency and with a non-enterprise currency.
- Verify that the exchange rate is correct and "nice" in each case.


server/controllers/finance/exchange.js
```diff
@@ -22,11 +25,47 @@ function getExchangeRate(enterpriseId, currencyId, date) {
...
 }
 
-// gets a positive number for the exchange rate display.
+// gets a non-fractional number for the exchange rate display.
 function formatExchangeRateForDisplay(value) {
   return (value < 1) ? util.roundDecimal(1 / value, 2) : value;
 }
 
+// Construct the HTML string to show the exchange rate in a nice format
+function exchangeRateMsg(currencyId, rate, enterprise, lang, date) {
```
jniles: I think this approach is close, but not optimal in my view.  Would it be possible to make this into a [handlebars helper](https://handlebarsjs.com/guide/partials.html#basic-partials)?  That seems like a better approach.  

Here is why:

1. Within the handlebars templates, you already have access to certain metadata - the enterprise (and therefore the currency), the language, etc.  You would need to add target currency and date.
2. You can use the handlebars translate + currency rendering to get both those elements for free.
3. We have less HTML embedded in JS, which makes it hard to lint, read, and debug.
3. It should be much easier to integrate this into other reports.

I would propose that you break this out to a handlebars partial.  If you need to attach a common object to all reports (like we do already in `ReportManager` with the `metadata` tag), please do it.

https://github.com/IMA-WorldHealth/bhima/blob/f35a8dfdf0556bda0bc35da9e805c58641e45ce2/server/lib/ReportManager.js#L133-L135

jniles: Concretely, rather than adding:
```handlebars
{{{ exchangeRateMsg }}}
```

It would be better to have:
```handlebars
{{>exhangeRateMessage foreignCurrency=currency rate=rate}}
```

jniles (commented):
@jmcameron I've proposed a modification to your approach.  Let me know what you think.  I don't know if you tried it and it didn't work, or if this was the first brush.

Let me know if you have any questions!

server/controllers/finance/exchange.js
jmcameron: 👍 I'll work on it.  Thanks for the suggestion.   I was not sure how to do it as a handlebars helper...

jmcameron: Done

jniles (approved):
This looks good to me.  Tested with both Vanga and IMCK databases (with two different enterprise currencies) and it works well.

image

bors r+

Issue #6235, opened 2022-01-03
Officially support ODK Central for offline stock movements


jniles:
We released a tag [`v1.21.3-pcima`](https://github.com/IMA-WorldHealth/bhima/releases/tag/v1.21.3-pcima) with support for ODK Central.  This was to handle the PCIMA project's need to track stock offline with ODK forms.  I propose we formalize this support in BHIMA and merge the functionality into `master`.  It will make support easier going forwards.

We've released a module [`@ima-worldhealth/odk-central-api`](https://www.npmjs.com/package/@ima-worldhealth/odk-central-api) that should offload the ODK Central API functionality into its own tested module.  This means we can focus on rewriting the [`import.js`](https://github.com/IMA-WorldHealth/bhima/blob/v1.21.3-pcima/server/controllers/odk/import.js) functionality to bring stock movements into BHIMA from data collected in the ODK Central Server.

---

One thing to think about - we could migrate the pepkits project to using this version of BHIMA/Central and have another project using this functionality in addition to the PCIMA project.

jniles:
Similarly, we might be able to improve our aggregate stock consumption by providing telephones for users in the health centers to use to record their stock consumption information.  This would ensure that we have more accurate data from the FOSAs in our Health Zone stock tracking project.

Basically, the following projects would greatly benefit from this:
 - [x] OFDA Pepkits III
 - [x] ASSR Stock Movement Tracking in Tshikaji Health Zone
 - [x] UNICEF PCIMA

jniles:
# Design

This describes the rough design of the BHIMA ODK link.  It is useful to look over the [ODK Central API documentation](https://odkcentral.docs.apiary.io/#introduction/api-overview).

## A BHIMA enterprise is an ODK Central project

When you install BHIMA, you'll be allowed to optionally add the configuration for an ODK Central Server.  This would be the username, password, and server URL.  This user must be an administrative user to create new projects.  A project on ODK Central would be created for the enterprise, and the `odk_project_id` would be stored in some table, probably its own table aside.  Currently, we don't model multiple enterprises in BHIMA.

Information we need to store in BHIMA so far:

- `odk_central_url`
- `odk_admin_user`
- `odk_admin_password`
- `odk_project_id`

## A BHIMA user (with access to a depot) is an ODK Central user

[...]

PR #6285, opened 2022-01-17
feat: add odk_central_integration tables

jniles:
Adds odk central integration tables for the eventual link to ODK Central.  Currently adds them under Stock Settings, but we can move them easily enough later.

Related to #6235.


jniles:
Closing as this functionality is in #https://github.com/IMA-WorldHealth/bhima/pull/6289

PR #6293, opened 2022-01-19
Display lot barcode

### LINKED ISSUE #6290: Add lot barcodes to Lots in Stock Registry export
We need the following abilities:

1. We need to add the ability to download/print the barcode from the lot registry.
2. We need the ability to scan a barcode and find it in the Lot/Asset registry.
3. We need a column in the XLSX export of the lot registry called "barcode" that contains the barcode text so we can turn them into barcodes using external programs.

mbayopanda:
This PR adds : 
- A way for displaying and print individual lot barcode
image

image

- The barcode column in the lot registry
image

Close #6290


server/controllers/stock/reports/stock/lot_barcode/lot_barcode.js
```diff
@@ -0,0 +1,51 @@
...
+  _, ReportManager, db, identifiers, barcode, LOT_BARCODE_TEMPLATE,
+} = require('../../common');
+
+/**
+ * @method lotBarcodeReceipt
+ *
+ * @description
+ * This method displays the lot barcode
+ *
+ * GET /receipts/stock/lot_barcode/:uuid
```
jniles: This is nice!  I would make this URL `/receipts/stock/lots/:uuid/barcode` just to be more rest-like :)

server/controllers/stock/core.js
```diff
@@ -288,9 +290,11 @@ async function getLotsDepot(depotUuid, params, finalClause) {
 
   const resultFromProcess = await db.exec(query, queryParameters);
 
-  // add minumum delay
+  // add minumum delay and barcode
+  const { key } = identifiers.LOT;
```
jniles: Nice 👍

client/src/modules/stock/lots/templates/action.cell.html
```diff
@@ -12,7 +12,11 @@
         <i class="fa fa-edit"></i> <span translate>FORM.LABELS.EDIT</span>
       </a>
     </li>
-
+    <li>
```
jniles, line 16: Can you put `<li class="separators">`s here to logically group the different parts of this menu together?

Something like:
  1. Everything that is a receipt goes together
  2. Everything that is a link goes together.
  3. Everything that causes an action (edit), goes together.
  4. If it is destructive (delete, cancel) it should be red and in its own section.

server/controllers/stock/reports/stock/lot_barcode/lot_barcode.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+function lotBarcodeReceipt(req, res, next) {
+  let report;
+  const data = {};
+  const uuid = db.bid(req.params.uuid);
+  const options = {
+    filename : 'LOTS.BARCODE_FOR_LOT',
+    pageSize : 'A6',
```
jniles, line 19: Does A7 exist?  :smile:

client/src/modules/stock/lots/registry.service.js
```diff
@@ -46,6 +46,11 @@ function LotsRegistryService(uiGridConstants, Session) {
       displayName : 'TABLE.COLUMNS.INVENTORY_GROUP',
       headerTooltip : 'TABLE.COLUMNS.INVENTORY_GROUP',
       headerCellFilter : 'translate',
+    }, {
+      field : 'barcode',
+      displayName : 'LOTS.BARCODE',
+      headerTooltip : 'LOTS.BARCODE',
+      headerCellFilter : 'translate',
```
jniles, line 53: Can you put this in a cell template with a `bh-reference-link` that opens the barcode modal automatically when we click it?

jniles: Also, let's set `visible: false` by default.

jniles (changes requested):
I made some basic comments, but the most important thing is that the barcode renderer doesn't seem to work on my computer!   Please investigate this.

firefox_rae64ltoQM

jniles:
Nevermind, the barcode not rendering was an enterprise setting.  Thanks @mbayopanda for pointing this out!

server/controllers/stock/reports/stock/lot_barcode/lot_barcode.js
mbayopanda: No, it doesn't

jniles:
@mbayopanda when you are comfortable that you are done with this, go ahead and merge it.

PR #6298, opened 2022-01-19
Fix expiration warning for not tracked inventories

mbayopanda:
This PR fixes the expiration date warning in the lot registry about inventory which doesn't support expiration date.


jniles (approved):
LGTM.

PR #6304, opened 2022-01-19
Fixed Excel export issues for the Lots registry

jmcameron:
Fixed a few issues in exporting the Lots registry as an Excel file.  Added more (initially hidden) fields and purged other (unneded) fields from the exported Excel file.  This also fixes the errors when some of the fields were converted to Excel output (some date values were invalid; converted the invalid date to an empty string).

Closes https://github.com/IMA-WorldHealth/bhima/issues/6295
Closes https://github.com/IMA-WorldHealth/bhima/issues/6296

**TESTING**
- Start with bhima_test
- Try Stock > Stock Lots (for the lot registry)
   - Use Menu > Columns and enable some of the new fields including 'Status'.  Check that the data in those new columns in the grid are okay.
   - Use Menu > Export to Excel and examine the output file:  Are the column labels good? Is the content of the columns okay?
- Try doing the same with a production data set (in particular check 'status' for correct handling of different status values).


server/controllers/stock/reports/stock/lots_report.js
```diff
@@ -52,8 +55,48 @@ function stockLotsReport(req, res, next) {
...
+    security_reached  : 'STOCK.STATUS.SECURITY',
+    minimum_reached   : 'STOCK.STATUS.MINIMUM',
+    over_maximum      : 'STOCK.STATUS.OVER_MAX',
+    unused_stock      : 'STOCK.STATUS.UNUSED_STOCK',
+  };
+
   return Stock.getLotsDepot(null, options)
     .then((rows) => {
+      // Purge unneeded fields from the row
+      rows.forEach(row => {
```
jniles, line 84: Just FYI - lodash has a utility for this:
```js
const rows = _.map(rows, row => _.omit(keys, row));
```

jniles:
@jmcameron I'm going to merge this, but can you check that this includes all the registries listed here: https://github.com/IMA-WorldHealth/bhima/issues/6295 ?

server/controllers/stock/reports/stock/lots_report.js
jmcameron: Actually I don't think this will work.  _.omit() creates a copy of the object with the members removed.   I think the invocation as shown above needs to do the changes in place.  And the _.unset() function takes only one path, so you would have to do a two-level mapping.  I think the forEach code I have above is cleaner and more efficient than copying lots (omit) or by doing two-level maps (unset).

PR #6323, opened 2022-01-21
Do not show expiration date for lots that do not expire

### LINKED ISSUE #6321: We shouldn't show "expires" if "tracking_expiration" is false.
image

jmcameron:
Fixed the Stock Entry and Stock Exit forms to prevent displaying expiration dates for lots that do not expire.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6321

**TESTING**
- Use any data set
- Pick any article that is in stock
- Check the inventory group for that article of stock and determine if it is tracking expiration.  If it is, temporarily change it to not tracking (Inventory > Inventory Configuration > (select item) and click [Edit]).
- Do stock entry and stock exit forms and verify that when you select a lot it does not include the expiration date on the drop-down to select the lot.
- Change the tracking expiration status to track expiration dates.
- Repeat the entry/exit forms and verify that the lot selection drop-downs now have expiration dates
- Change the tracking expiration setting for the inventory group back to its original value.

### DIFF
--- a/client/src/modules/stock/entry/modals/templates/selectLot.tmpl.html
+++ b/client/src/modules/stock/entry/modals/templates/selectLot.tmpl.html
@@ -1,7 +1,7 @@
 <a id="lot-uuid-{{ match.model.uuid }}" href>
   <span ng-bind-html="match.model.label | uibTypeaheadHighlight:query"></span>
-  <span style="margin-left: 1em">[</span>
-  <span translate>INVENTORY.EXPIRES</span><span> </span>
-  <span ng-bind-html="match.model.expiration_date | date"></span>
-  <span>]</span>
+  <span ng-if="match.model.tracking_expiration" style="margin-left: 1em">[</span>
+  <span ng-if="match.model.tracking_expiration" translate>INVENTORY.EXPIRES</span>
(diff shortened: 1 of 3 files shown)


client/src/modules/stock/entry/modals/templates/selectLot.tmpl.html
```diff
@@ -1,7 +1,7 @@
 <a id="lot-uuid-{{ match.model.uuid }}" href>
   <span ng-bind-html="match.model.label | uibTypeaheadHighlight:query"></span>
-  <span style="margin-left: 1em">[</span>
-  <span translate>INVENTORY.EXPIRES</span><span> </span>
-  <span ng-bind-html="match.model.expiration_date | date"></span>
-  <span>]</span>
+  <span ng-if="match.model.tracking_expiration" style="margin-left: 1em">[</span>
```
jniles, line 3: I worry about some DOM trashing with this (`ngIf` creates/removes DOM nodes).  Maybe AngularJS is smart enough, but an easy win would be to reduce all these `ngIf`s to a single one by wrapping everything that needs to be displayed or hidden in a single `<span>` that is conditionally shown, and put the `ngIf` on that.  Could you do that?

jniles:
I'm switching this to `master`, since we are going to start trying to pull in parts of `sprint-prosani` to `master`.

PR #6331, opened 2022-01-24
bug(Creation and Update depot)

### LINKED ISSUE #6329: Disfunction on the creation of depots
- It is noticed that creating a depot from the parent repository no longer works very well
- But also that the location data is lost when trying to update a depot

lomamech:
- fix creation of depot from parent depot
- fix sql for gettig data from location of depot

closes #6329


jniles (approved):
LGTM.

bors r+

PR #6333, opened 2022-01-24
Stock Assignment registry export into PDF, CSV and Excel

mbayopanda:
This PR adds the stock assign registry export option for PDF, CSV and Excel.


jniles (commented):
For some reason, I get two "allTime" filters using the default database.

image

The rest looks good to me.

jniles (approved):
LGTM.

Bors r+

PR #6340, opened 2022-01-25
Fix XSL export problems

### LINKED ISSUE #6295: Translate all XLSX headers in the stock registries
We need to translate all the column headers of our XLSX exports.  The key registries to focus on are:

- [x] Stock Lots
- [ ] Stock Movements
- [ ] Articles in Stock
- [ ] Inventory

Here is how to get the XLSX export, for example:

firefox_E8BUZfud4r

As you can see, things like `mvt_quantity` are not translated.  We should have these translated for the presentation.

### LINKED ISSUE #6296: Stock Lots registry warns "object" sent to XLSX export
If you export the stock lots on master to XLSX, you get this warning:

image

We should probably look into it.  I imagine it is the `tags`.

jmcameron:
Fixed Excel export/download for
- Articles in Stock (Note: changed 'xlsxReport' to 'xlsx' until [Issue 6313](https://github.com/IMA-WorldHealth/bhima/issues/6313) is fixed)
- Stock Movements
- Stock Lots 

Added more hidden columns to these reports and purged some unneeded fields for these reports. This PR also fixes the errors when some of the fields were converted to Excel output (some date values were invalid; converted the invalid date to an empty string).

This is a reapplication of PR https://github.com/IMA-WorldHealth/bhima/pull/6304 and https://github.com/IMA-WorldHealth/bhima/pull/6314 to master (they were applied to sprint-prosani).

Closes https://github.com/IMA-WorldHealth/bhima/issues/6295
Closes https://github.com/IMA-WorldHealth/bhima/issues/6296

Note: This PR does not address XLS export from the Inventory Registry because it does not currently have an XLS export function.

**TESTING**
- Use any data set (but may be simpler with bhima_test)
- Try Stock Lots Registry, Articles in Stock, and Stock movements.  For each:
  - Try Menu > Columns and show all hidden columns
     - Verify that the columns make sense
  - Try Menu > Download as Excel
     - Open the downloaded file
       - Verify all column headers are translated (Capitalized)
       - Check data in all columns
- Try doing the same tests with a production data set


client/src/modules/stock/inventories/registry.js
```diff
@@ -78,52 +78,69 @@ function StockInventoriesController(
...
-  }, {
-    field            : 'S_MONTH',
-    displayName      : 'MS',
+    field            : 'avg_consumption',
+    displayName      : 'STOCK.CMM',
+    headerTooltip    : 'STOCK.MONTHLY_CONSUMPTION.AVERAGE_MONTHLY_CONSUMPTION',
+    headerCellFilter : 'translate',
     enableFiltering  : false,
-    enableSorting    : false,
+    enableSorting    : true,
```
jniles, line 86: 👍

jniles (approved):
This looks good to me.

bors r+

PR #6341, opened 2022-01-25
Clean up ngIfs in lot selection for stock entry

jmcameron:
Per the request in PR https://github.com/IMA-WorldHealth/bhima/pull/6323 (See https://github.com/IMA-WorldHealth/bhima/pull/6323#discussion_r790472623), this reduces the number of ngIfs in the lot selection dropdown code for Stock Entry.

**TESTING**
- Use any data set
- Pick any article that is in stock
- Check the inventory group for that article of stock and determine if it is tracking expiration.  If it is, temporarily change it to not tracking (Inventory > Inventory Configuration > (select item) and click [Edit]).
- Do stock entry and stock exit forms and verify that when you select a lot it does not include the expiration date on the drop-down to select the lot.
- Change the tracking expiration status to track expiration dates.
- Repeat the entry/exit forms and verify that the lot selection drop-downs now have expiration dates
- DON"T FORGET:  Change the tracking expiration setting for the inventory group back to its original value.


jniles (approved):
LGTM.

bors r+

PR #6347, opened 2022-01-26
Update recovery capacity report

### LINKED ISSUE #6182: Updates for the Recovery Capacity Report
Finance > Report > Recovery Capacity Report

**Suggestions**
- Increase the font size of the date range on the report
- Add multi-currency support
- If "Include Unposted Values" is set to "Yes", add a column indicating whether items are posted or not
- A summary would be useful at the bottom indicating % of interval where recover was above 70% (green), 60-70% (yellow), and less than 70% (red).
- A possible low-priority enhancement:  It is possible there is correlation between the days that have poor recovery capacity and the days of the week.   A possible formatting idea is show the day of the week for each row in the report too and possibly marking the rows that begin each week.  Obviously this should be investigated to see if there really is such a correlation before doing this!

lomamech:
- Increase the font size of the date range on the report
- Add multi currency support
- Add days of the week
- Add a summary at the bottom of percentage of recovery

closes #6182

### DIFF
--- a/client/src/modules/reports/generate/recovery_capacity/recovery_capacity.html
+++ b/client/src/modules/reports/generate/recovery_capacity/recovery_capacity.html
@@ -29,6 +29,11 @@
               limit-min-fiscal>
             </bh-date-interval>
 
+            <bh-currency-select
+              currency-id="ReportConfigCtrl.reportDetails.currencyId"
+              on-change="ReportConfigCtrl.onSelectCurrency(currency)">
+            </bh-currency-select>
+
             <bh-yes-no-radios
               value="ReportConfigCtrl.reportDetails.includeUnpostedValues"
               name="includeUnpostedValues"
--- a/server/controllers/finance/reports/recovery_capacity/index.js
+++ b/server/controllers/finance/reports/recovery_capacity/index.js
@@ -191,18 +191,23 @@ async function report(req, res, next) {
       const numDays = moment(row.date).day();
       row.dayOfWeek = daysWeek[numDays];
 
-      rowLength += 1;
+      if (row.total_invoiced > 0 || row.total_paid > 0) {
+        rowLength += 1;
+        row.active_day = 1;
 
-      if (row.recovery_capacity > 0.7) {
-        greenAbove70 += 1;
-      }
+        if (row.recovery_capacity > 0.7) {
+          greenAbove70 += 1;
+        }
 
-      if (row.recovery_capacity >= 0.6 && row.recovery_capacity <= 0.7) {
-        yellowBetwen60And70 += 1;
-      }
+        if (row.recovery_capacity >= 0.6 && row.recovery_capacity <= 0.7) {
+          yellowBetwen60And70 += 1;
+        }
 
-      if (row.recovery_capacity < 0.6) {
-        redLess60 += 1;
+        if (row.recovery_capacity < 0.6) {
+          redLess60 += 1;
+        }
+      } else {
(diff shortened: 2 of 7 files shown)


lomamech:
image

image

lomamech:
image
@jmcameron

client/src/modules/reports/generate/recovery_capacity/recovery_capacity.html
```diff
@@ -29,6 +29,11 @@ <h3 class="text-capitalize" translate>REPORT.RECOVERY_CAPACITY.TITLE</h3>
               limit-min-fiscal>
             </bh-date-interval>
 
+            <bh-currency-select
```
jmcameron, line 32: I suggest moving this down to right before the [Submit] button.  I think the "include unposted values" option affects the content of the report, so should be grouped with the other ones (eg, date rage).  However the currency option only affects the format of the data, so I would put it lower in the list.

server/controllers/finance/reports/recovery_capacity/index.js
```diff
@@ -39,6 +43,16 @@ async function report(req, res, next) {
     const formatedDateFrom = moment(dateFrom).format('YYYY-MM-DD');
     const formatedDateTo = moment(dateTo).format('YYYY-MM-DD');
 
+    const daysWeek = [
```
jmcameron, line 46: It seems like this should be in some common file for access by any report on the server?

jmcameron (commented):
It is looking good.  I had a few suggestions (see the inline code comments).  I like the summary table, but several things were confusing to me:
image
Specifically:

- We should not display "Null" to end users.  Not sure why that occurred.  If it means "0", we should set it to 0.  If it means "unknown", we should indicate that and add an extra row to the summary that indicates the number of occurrences/percentage for it.
- In the summary, the column header "Percentage" is not really percentage, is it?  Isn't it "Count" / "Occurrences" / number of days in this category?   It should be labeled appropriately.
- Lastly, just a question: What does the last number in the "Total" row mean?  Is it the average over all days?  Not sure how to make that clearer.   Perhaps changing "Total" to something else (not sure what to suggest).

Since this report could be very long, I suggest moving the summary to the top and centering it.

lomamech:
@jmcameron I think it will be preferable to leave the field empty when there is no activity but also to ignore the days when there is no activity when calculating the percentage

image

image

Notice the percentage difference between the two reports

jmcameron:
> @jmcameron I think it will be preferable to leave the field empty when there is no activity 

I'm okay with that.  Just don't put 'Null' !

> but also to ignore the days when there is no activity when calculating the percentage
> Notice the percentage difference between the two reports

I do not know about this.  I suggest consulting with @jniles and/or AG Neron for guidance on this.

jniles:
~~In my view, if there are patients or invoices, we should include the 0% in our calculations.  If there is no activity, we leave it blank and we skip it.~~

I have been persuaded that we should show 0% only if there are invoices and no payments.

jmcameron:
One more small thing:  Since we are showing the summary on the report, I suggest that you show the total number of days involved somewhere, preferably as part of the summary block.

lomamech:
Last update
image

jmcameron (approved):
It looks good.  Since the summary is based on the number of "active" days, I suggest changing the the label from "Number of days" to "Number of active days" (or whatever makes sense to you).

jniles (approved):
LGTM.

Bors r+

PR #6368, opened 2022-02-02
Assets barcode generator

mbayopanda:
This PR adds a tool for generating barcodes (tag numbers) for assets before entering them in the stock, so that barcodes will be printed and affixed on assets before scanning them with BHIMA.

In this tool, barcodes are generated in two format : 
- CSV : which can be used with specialized software such as Dymo
- PDV : which can be printed and affixed on assets

firefox_rRNJJ1NfIU


client/src/js/services/LotService.js
```diff
@@ -82,6 +82,10 @@ function LotService(Api, $http, util) {
       .then(util.unwrapHttpResponse);
   };
 
+  lots.generateTags = (number) => {
+    return window.open(`/lots/generate_barcodes/${number}`);
```
jniles: Can you make this `$window` instead of `window`?  It's best practice for AngularJS testing.

client/src/modules/stock/entry/modals/generateTags.modal.js
```diff
@@ -0,0 +1,23 @@
...
+  '$uibModalInstance', 'LotService',
+];
+
+function GenerateTagsModalController(Instance, Lots) {
+  const vm = this;
+
+  vm.cancel = cancel;
+
+  vm.submit = () => {
+    if (!vm.totalTags || vm.totalTags > 500) { return null; }
```
jniles: Can you just make this an `ng-max` rule, and then use the `form.$invalid` check?

package.json
```diff
@@ -115,6 +116,7 @@
     "express-handlebars": "^6.0.2",
     "express-session": "^1.17.2",
     "font-awesome": "^4.7.0",
+    "fs-extra": "^10.0.0",
```
jniles: I don't think you need `fs-extra`. Eveything can be done using `fs`.

package.json
```diff
@@ -134,6 +136,8 @@
     "nodemailer": "^6.7.1",
     "p-retry": "^4.6.1",
     "pako": "^2.0.4",
+    "pdf-merge": "^1.2.0",
```
jniles: Where do you use `pdf-merge`?

jniles (commented):
This looks pretty good.  I've made some inline comments.

Could you rename these `generateTag` functions to `generateAssetTag`?  The main issue is that we already have a table named `tag` and it is confusing that all these `tag` functions don't have anything to do with that.   We could also rename all `tag` related names to `barcode` or `serial` or something else.

jniles (approved):
Looks good to me!  Really cool!

Bors r+

server/controllers/stock/lots.js
```diff
@@ -442,3 +455,65 @@ function assignments(req, res, next) {
...
+ * GET /lots/generate_barcodes/:number
+ *
+ * @description
+ * Returns generated barcodes in a zip file
+ */
+async function generateBarcodes(req, res, next) {
+
+  try {
+    const totalBarcodes = req.params.number;
+    const { key } = identifiers.LOT;
```
jniles, line 469: We should probably add the ability to send our own identifier from the client to make this more flexible.

Issue #6349, opened 2022-01-26
Asset Management - Stock Lot changes


jmcameron:
### Database Changes
- add: ‘serial_number’ field (text, 40 chars?, defaults to NULL),   This should be called ‘Manufacturer’s serial number’ in the GUI.
- Change GUI text of “lot” to be “lot/batch number”

### Interface changes
- Update stock lots registry grid to support the new serial number field (default to hidden).
- Perhaps If any of the lots are assets (determined via inventory is_asset field), make the grid show the new serial number field and perhaps show (read-only) the asset-related inventory fields (make, model, equipment_type).
- Editing lots should support editing this field -- IF the lot is an asset (according to the inventory is_asset flag).

PR #6390, opened 2022-02-09
Feature asset inventory changes

### LINKED ISSUE #6348: Asset Management - Inventory Changes
Our goal is to use the current stock lots as assets.  In order to do this, we need to add some data to the 'inventory' table and add some related functionality.

### Database Changes
- add: ‘is_asset’ to inventory table (boolean, default to false)
- add: ‘external_reference_number’ to inventory table (TEXT, defaults to NULL).  
- To be used to refer to an external reference such as PROSANI inventory ID.
- add: ‘manufacturer_brand’ to inventory table (marque) field (TEXT, defaults to NULL)
- add: ‘manufacturer_model’ field to inventory table (TEXT, defaults to NULL)
- add: ‘equipment_type_id’ to inventory table (reference into an equipment type table)

**Create ‘equipment_type’ table including these fields:**
- id
- label (TEXT, raw string or i18n token)
- description (TEXT)
- is_predefined (if label is i18n token)

**Notes**
- ‘inventory group’ is not quite suitable for equipment types.
- The equipment_type ‘label’ field will be the i18n token for predefined entries but raw text in case entered manually at a site.  Note that something like this is done in the cost_center_allocation_basis table and related code.

### Interface changes
- In the Inventory registry, add the new fields to the grid.  They should be hidden by default.
- The form to edit inventory items should allow editing these new fields.
- Add a simple editor for the equipment_type table entries.  It should not be possible to edit the label for equipment_types with predefined tokens.
- Add the ability to export the inventory registry as XLSX.

lomamech:
- Add property _is_asset, external_reference_number, manufacturer_brand, manufacturer_model_, in table **inventory**
- Use table **inventory_type** instead creat table **equipement_type**
- Complete integration test
- And constaint in form of inventory creation, allow reference number,
  manufacturer brand and manufacturer model only of Asset

closes #6348


client/src/modules/inventory/configuration/types/modals/actions.modal.js
```diff
@@ -60,12 +66,10 @@ function InventoryTypeActionsModalController(InventoryType, Notify, Instance, Da
 
     if (vm.identifier) {
       InventoryType.read(vm.identifier)
-      .then(function (type) {
-        vm.session = type[0];
-      })
-      .catch(Notify.handleError);
+        .then((type) => {
+          [vm.session] = type;
```
jmcameron, line 70: 👍

client/src/modules/inventory/configuration/types/types.html
```diff
@@ -40,16 +40,19 @@
       <tbody>
         <tr ng-repeat="type in TypesCtrl.typeList | filter: TypesCtrl.searchValue | orderBy: 'text'">
           <td>{{ $index + 1 }}</td>
-          <td>{{ type.text }}</td>
           <td>
-            <a style="cursor:pointer; text-decoration:none;"
+            <span ng-if="type.is_predefined"><strong>{{ type.text }}</strong></span>
```
jmcameron, line 44: Shouldn't there be a 'translate' if it is predefined?

client/src/modules/inventory/configuration/types/types.js
```diff
@@ -73,6 +73,12 @@ function InventoryTypesController(InventoryType, Notify, Modal) {
     // get inventory types
     InventoryType.read()
       .then((list) => {
+        list.forEach(item => {
+          if (item.is_predefined) {
+            item.text = $translate.instant(item.text);
```
jmcameron, line 78: Ok.  Ignore my last comment.

client/src/modules/inventory/list/modals/actions.tmpl.html
```diff
@@ -44,6 +44,58 @@
...
     </div>
 
+    <div class="checkbox">
+      <label>
+        <input type="checkbox" name="is_asset" ng-true-value="1" ng-false-value="0" ng-model="$ctrl.item.is_asset">
+        <strong><span translate>INVENTORY.IS_IT_ASSET</span></strong>
+      </label>
+    </div>
+
+    <div ng-if="$ctrl.item.is_asset" class="form-group"
```
jmcameron, line 54: Nice!  I like that you only show these asset-related fields if it is an asset.

jmcameron (commented):
When I edit an inventory item (via its action menu 'Edit' link) and I save the modifications, the Inventory registry loses my columns definitions.   The columns definitions should be cached.

Regarding whether 'inventory_type' is okay to use rather than adding a new 'equipment_type' filed.  I think re-using 'inventory_type' is fine.  So I do not object to using 'inventory_type' for this purpose.  However, we should take another look at our current predefined items for this field:  Article, Assembly, Service (IMCK adds 'Honoraire mediccin', and 'SOJA').  I'm not sure how these are being used, put we probably need to add explanations about this change when we install a release at production sites with this change.

I think that VARCHAR(30) for inventory_type.text may be too small.  Since it can now contain a i18n token, it could easily be longer.  So I suggest increasing its size to VARCHAR(200) or 'text' (for reference see cost_center_allocation_basis).

Although the new 'external_reference_number' field for 'inventory' was intended for "External" references, it could potentially be used for internal references.  So I suggest changing the 'external_reference_number' field name to 'reference_number'.  I notice that the GUI never mentions 'external', so this seems like a simple change.

Otherwise it looks good.  I reviewed the code.   And I tested it with bhima_test and everything seemed to work okay.   I did not test the migration scripts on a production dataset.  

After making these changes, please ask me for a new review and I'll check the migration part.

client/src/modules/inventory/configuration/types/types.html
jmcameron: I see that you took care of this later

jmcameron:
I checked 'Stock Lots' after finishing the review.  Shouldn't the new Inventory fields be available in the Stock Lots registry (via Menu > Columns)?

lomamech:
> When I edit an inventory item (via its action menu 'Edit' link) and I save the modifications, the Inventory registry loses my columns definitions. The columns definitions should be cached.

To keep the columns visible, click on the menu then on the Column button
image

Then you have to choose the columns to display
image

And to save the configuration you just have to click on the **Save Grid Configuration** button
image

lomamech:
@jmcameron i need a new review, it took me longer because I had also anticipated with the issue #6349 **_Asset Management - Stock Lot changes_**

jmcameron:
> And to save the configuration you just have to click on the **Save Grid Configuration** button !

Okay.  I tested this and it works fine.  However, I still wonder why we need the "Save Grid Configuration" function instead of caching any changes to the grid immediately.   I think this is a broader "usability" issue and should not slow down this PR.

lomamech:
> > And to save the configuration you just have to click on the **Save Grid Configuration** button !
> 
> Okay. I tested this and it works fine. However, I still wonder why we need the "Save Grid Configuration" function instead of caching any changes to the grid immediately. I think this is a broader "usability" issue and should not slow down this PR.

This functionality exists on all the grids of the BHIMA system and each user can save the grid, the size, the visibility and the positioning of the columns

jmcameron (commented):
I tried doing the migration procedure and ran into several problems:

- In the migration.sql, file, the change of inventory_type.text cannot include the 'AFTER' clause.  That makes sense because it is modifying an existing field, not added a new field.
- The definition of inventory_type.text needs to be updated in schema.sql (to VARCHAR(200) like the migration change).
- In the migration file (line 57), 'external_reference_number' needs to be changed to 'reference_number'

- The definitions in admin.sql of zRecalculateCostCenterAggregates() and zRecomputeLotBarcodes() need to be preceded by DROP PROCEDURE IF EXISTS <name>

Please make these updates and I'll try the migration procedure again.

server/models/admin.sql
```diff
@@ -265,6 +265,7 @@ END$$
 
  Removes all data from the cost_center_aggregate table and rebuilds it.
 */
+DROP PROCEDURE IF EXISTS zRecalculateCostCenterAggregates$$
```
jmcameron, line 268: Sorry, my mistake.  This addition was not necessary.  This statement is already in the SQL files somewhere.  When you create the migration file, you can find it.

server/models/migrations/next/migrate.sql
```diff
@@ -54,13 +54,12 @@ INSERT IGNORE INTO unit VALUES
   * @date: 2022-02-07
   */
 CALL add_column_if_missing('inventory', 'is_asset', 'TINYINT(1) NOT NULL DEFAULT 0 AFTER `importance`');
-CALL add_column_if_missing('inventory', 'external_reference_number', 'TEXT NULL');
+CALL add_column_if_missing('inventory', 'reference_number', 'TEXT NULL');
```
jmcameron, line 57: 👍

jmcameron (commented):
I tested migrating an production dataset with this PR and it worked okay.   I also tested displaying and editing asset-related data in the inventory registry and that worked fine.

Before releasing I suggest eliminating the extra DROP PROCEDURE (see the SQL code comments) and adding a DROP PROCEDURE fro zRecomputeLotBarcodes() in admin.sql.

I think this PR is ready to go, but I encourage getting an review by @jniles or @mbayopanda before merging.

client/src/modules/inventory/list/modals/asset.cell.tmpl.html
```diff
@@ -0,0 +1,4 @@
+<div class="ui-grid-cell-contents text-center">
+  <i ng-show="row.entity.is_asset === 1" class="fa fa-check-circle-o"></i>
+  <i ng-show="row.entity.is_asset === 0" class="fa fa-times-circle-o"></i>
```
jniles, line 3: You could just do this with a `ng-class={}` directive on a single `i`.  I don't think this has any performance penalty though.

client/src/modules/inventory/list/templates/reference_number.cell.html
```diff
@@ -0,0 +1,5 @@
+<div class="ui-grid-cell-contents">
+  <em>
+    <span>{{ row.entity.reference_number }}</span>
```
jniles, line 3: Why does this need to be a template?  You can do this with just a cellStyle, right?

client/src/modules/inventory/list/list.html
```diff
@@ -88,6 +88,8 @@
       ui-grid-save-state
       ui-grid-exporter
       ui-grid-auto-resize
+      ui-grid-resize-columns
```
jniles, line 91: :+1:

jniles (approved):
This looks good to me.  Thanks @lomamech 

bors r+

PR #6392, opened 2022-02-10
Fix issue with partial transfers

### LINKED ISSUE #6391: Stock entry from a transfer allow multiple bad entries
In the stock entry from a transfer, when we select a movement already entered we still have the original lots (all lots concerned by the movement) instead of have only not entered lots with remaining quantities

chrome_7j7GBFBhTn

jmcameron:
Fix issue when doing multiple stock entries from a single transfer.

Also added warning if the user tries to use a transfer that has already been completely processed.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6391

**TESTING**
See the issue above for a test procedure.


mbayopanda (approved):
It works like a charm 🥳 

LGTM 👍🏽 

bors r+

PR #6403, opened 2022-02-15
Updated headercrumbs for User and Role management create/edit dialogs

### LINKED ISSUE #6317: Role Management Create form should use headercrumb class
The role create form looks like this:

image

The user create form, which uses the headercrumb, looks like this:

image

The second one is used more, and gives more context.  We should add in the headercrumb to the role creation modal.

jmcameron:
Updated the Role Management and User Management create/edit dialogs to improve their headercrumbs.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6317

**TESTING**
- Any database is fine
- Try creating and editing Roles and Users
   - Administration > Role Management
   - Administration > User Management
- Verify that the headercrumb at the top of the create/edit dialog is okay in all 4 cases.

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -887,6 +887,7 @@
             "DECEMBER": "December",
             "PERMISSION_ASSIGNED_SUCCESS":"Permissions successfully assigned",
             "EDIT_ROLE":"Assign role(s)",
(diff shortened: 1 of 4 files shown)


jniles (approved):
LGTM.

bors r+

PR #6409, opened 2022-02-17
Verify that end date >= start date for bhDateInterval component

### LINKED ISSUE #6233: Update date selection components to check that end date >= start date
These components should be updated to verify that the end date/period is >= the end date/period.  Users of these components should not need to double-check this.
- bhDateInterval
- bhFiscalPeriodSelect

jmcameron:
Verify that "end date" >= "start date" when selecting a date interval using the bhDateInterval component.
When using the calendar widget this is not needed.  However if one types in the date manually, it is possible to chose an end date that is before the start date.  This PR adds checks to prevent that.   Note that if this occurs, the end date is erased immediately an error message is displayed.   I could not think of how to modify the form controls to do this automatically (in other words force the form into an error state so you could not submit).

Closes https://github.com/IMA-WorldHealth/bhima/issues/6233

NOTE: It was not necessary to fix anything related to the bhFiscalPeriodSelect component since its drop-down behavior prevents this type of error and there is no way to type in a period number manually.

**TESTING**
- Try the date interval in Stock > Stock Exit
- Enter a start date (manually or via the calendar widget)
- Verify that you cannot enter a end date before the start date via the calendar widget
- TYPE in an end date before the start date and observe the error handling behavior

### DIFF
--- a/client/src/i18n/en/errors.json
+++ b/client/src/i18n/en/errors.json
@@ -18,6 +18,7 @@
     "UNAUTHORIZED" : "You have submitted a bad username or password combination.  Please login in with valid credentials to continue.",
     "UNKNOWN" : "An error occurred.",
     "OVERPAID_INVOICE" : "You have overpaid the invoices.",
+    "ER_DATE_RANGE" : "The end date cannot be before the start date!",
     "ER_DUP_KEY" : "A key collided in a unique database field.  Please retry your action.  If the problem persists, contact the developers.",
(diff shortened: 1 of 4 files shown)


jniles (approved):
This is a fine solution.  I tested it and it works.

Ideally, the best solution would be to make a custom validation directive using the `$validators` API.  See [this SO paste](https://stackoverflow.com/a/43840247/6251968) for one idea.

I'm fine merge this as is though. 

bors r+

PR #6437, opened 2022-03-03
Delete asset-related data from inventory item when it is not an asset

### LINKED ISSUE #6402: Assets Management - Inventory editing issue
AFTER PR https://github.com/IMA-WorldHealth/bhima/pull/6390 is merged!

The ability to edit asset-related inventory data implemented in PR https://github.com/IMA-WorldHealth/bhima/pull/6390 is fine.  However there is one small thing that should be fixed:   When you have an inventory item that has been marked as an asset (is_asset = true), and then you edit it and change it to NOT be an asset, the asset-related fields (reference number, manufacturers brand & model) should be deleted before saving the inventory item.

jmcameron:
When the user changes an inventory item to an asset and adds asset-specific data (manufacturer brand, etc) that data is saved in the inventory item.  If the user decides that it should not be an asset, edits the inventory item, and changes it to NOT an asset, this PR fixes things so that the asset-related data is removed.

**QUESTION:**  Do we want manufacturer brand and model names for non-asset stock?  If so, this PR should be updated and only the reference_number should be removed and the edit form should be changed to not hide the manufacturer data when it is not an asset.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6402

**TESTING**
- Use any data set
- Open Inventory > Inventory Registry
- In the [Menu] > Columns, display the is_asset, reference_number, manufacturer_brand, and manufacturer_model columns AND save the Column configuration.
- Using the action menu, edit an inventory item, change it to an asset, enter reference_number, manufacturer brand, etc, and save it.
- Notice the asset-related data you entered in the columns
- Edit the inventory item again and  change it to NOT an asset, save
- The asset-related data should now be gone.
-


jniles:
> QUESTION: Do we want manufacturer brand and model names for non-asset stock? If so, this PR should be updated and only the reference_number should be removed and the edit form should be changed to not hide the manufacturer data when it is not an asset.

I don't think so.  Until proven otherwise, let's just remove those fields.

jniles (approved):

client/src/modules/inventory/list/modals/actions.modal.js
```diff
@@ -49,11 +50,28 @@ function InventoryListActionsModalController(
     vm.item.tags = tags;
   }
 
+  // Immediately clear out asset-related data if we toggle the is_asset checkbox
+  // (prevent preserving this data when the toggle is clicked twice)
+  function onChangeIsAsset(isAsset) {
+    if (!isAsset) {
```
mbayopanda, line 56: 👍🏽

mbayopanda (approved):
bors r+

PR #6439, opened 2022-03-04
Move reference_number from inventory to lot

### LINKED ISSUE #6433: Move reference_number field from inventory table to lot table
We initially put 'reference_number' in the 'inventory' table.   The idea was that 'reference_number' would be used as an "External Reference Number" to associate assets with specific orders by external entities.   However, an inventory item is actually more general.   For instance, different specific assets (lots) of the exact same inventory item could be ordered and processed by two different external entities or separate orders by the same external entity.  Therefore it does not make sense that this number should be in 'inventory'.  Instead it should be in the 'lot' table.

jmcameron:
Moved the column 'reference_number' from the inventory table to the lot table (per discussion in the issue).

Closes https://github.com/IMA-WorldHealth/bhima/issues/6433

**TESTING**
- Try editing an inventory item and notice that 'reference_number' is no longer available.
- Go to Stock > Articles in stock and pick an inventory item that has lots
- Edit the inventory item and change it to an asset and set the manufacturer brand/model
- Find the lots for the inventory item and edit one of them.  Set the reference_number
- In the Stock > Lots registry change the column definitions so show the asset related fields (is_asset, manufacturer_brand, manufacturer_model, serial_number) and then SAVE the column definitions
- Notice the asset related fields are available
- Use the Search and enter the reference number just added and notice that only that lot is shown
- Clear the search filter to revert to showing all lots
- Do a similar search filter test for is_asset


server/controllers/stock/core.js
```diff
@@ -99,9 +99,12 @@ function getLotFilters(parameters) {
   filters.equals('lot_uuid', 'lot_uuid', 'm');
   filters.equals('inventory_uuid', 'uuid', 'i');
   filters.equals('consumable', 'consumable', 'i');
+  filters.equals('is_asset', 'is_asset', 'i');
   filters.equals('group_uuid', 'uuid', 'ig');
   filters.equals('text', 'text', 'i');
   filters.equals('label', 'label', 'l');
+  filters.equals('is_asset', 'is_asset', 'i');
```
jniles, line 105: It looks like you have `is_asset` here listed twice.

server/models/migrations/next/migrate.sql
```diff
@@ -63,9 +63,17 @@ CALL add_column_if_missing('inventory_type', 'is_predefined', 'TINYINT(1) NOT NU
...
+ */
 CALL add_column_if_missing('lot', 'serial_number', 'VARCHAR(40) NULL');
+
+/*
+ * Issue: Asset Management - Move reference_number to lot
+ * @author: jmcameron
+ * @date: 2022-03-04
+ */
+CALL drop_column_if_exists('inventory', 'reference_number');
+CALL add_column_if_missing('lot', 'reference_number', 'TEXT NULL');
```
jniles, line 79: Nice 👍

jniles (approved):
This looks good to me.  Just one small change and then feel free to merge it.

server/controllers/stock/core.js
jmcameron: Done

jmcameron:
Bors r+

PR #6449, opened 2022-03-10
Fix payroll indices

### LINKED ISSUE #6443: Recreate a new function replacing the updateIndices() procedure
A malfunction has been observed concerning the updateIndice() procedure, not having found the cause of the malfunction, this issue proposes to replace this procedure with a function

lomamech:
- Write new function for fix bug in procedure updateIndices()
- Implement test for staffing indices configuration and setting

closes #6443

### DIFF
--- a/server/controllers/payroll/multiplePayrollIndice/paramter.config.js
+++ b/server/controllers/payroll/multiplePayrollIndice/paramter.config.js
@@ -6,6 +6,7 @@
  * filter.
  */
 const db = require('../../../lib/db');
+const util = require('../../../lib/util');
 
 // get staffing indice parameters
 function detail(req, res, next) {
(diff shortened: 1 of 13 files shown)


lomamech:
@jniles  with the new way, there is almost no difference in BHIMA and EXCELL OF VANGA
image

jniles (approved):
@lomamech does this make any other code useless?  If so, can you remove it?

Otherwise, this looks fine to me.  Please walk one of the other members of the team through it (Bruce, AG Neron, or someone) to make sure they are comfortable with how it works.

lomamech:
> @lomamech does this make any other code useless? If so, can you remove it?
> 
> Otherwise, this looks fine to me. Please walk one of the other members of the team through it (Bruce, AG Neron, or someone) to make sure they are comfortable with how it works.

I'll ask AG Neron

lomamech:
I just finished with the corrections, I just have to do a demonstration to allow this Pull Request to be merged

server/controllers/payroll/multiplePayrollIndice/paramter.config.js
```diff
@@ -21,19 +22,251 @@ function detail(req, res, next) {
...
+    let rubricTotalDaysId;
+    let rubricReagisteredIndexId;
+    let rubricTotalCodeId;
+    let rubricDayIndexId;
+
+    emp.totalBase = 0;
+
+    transaction.addQuery(`DELETE FROM employee_advantage WHERE employee_uuid = ?`, [emp.employee_buid]);
+
+    paymentIndice.forEach(ind => {
```
jmcameron: This would be more efficient if you use .find()

server/controllers/payroll/multiplePayrollIndice/paramter.config.js
```diff
@@ -21,19 +22,251 @@ function detail(req, res, next) {
...
+      if (emp.employee_uuid === ind.employee_uuid) {
+        // Calcul Total days for rubrics where type is 'is_day_worked' and 'is_extra_day'
```
jmcameron: Should be 'or' instead of 'and' in the comment?

jmcameron (commented):
I looked through the code and made a few comments.

I have not tested this but I hope to get a demo soon.

server/controllers/payroll/multiplePayrollIndice/paramter.config.js
lomamech: I can't quite figure out which functionality I should replace with the find() function

jmcameron: It looks like you have one item and you're looping through an array  (for foreEach) until you find a match.  Then you do the computations on the matching item.  And then you loop through the rest of the items in the array.   I was suggesting, you search for the matching item with find() and then do the operation on it.   find() should be faster than looping through the entire array.   Maybe I misunderstood what you're doing here...

lomamech:
@jmcameron I have finish

PR #6462, opened 2022-03-21
Create assets inventory management

### LINKED ISSUE #6350: Asset Management - New Asset Management scan support
### Database Changes
- Create new ‘asset_scans’ table
  - uuid
  - asset_uuid (this is the **uuid** in lot table)
  - scanned by (person/name)
  - date (of scan)
  - location (depot)
  - condition_id (id in ‘equipment_condition’ table, similar sql, editing functions as the equipment_type table)
  - Purpose (defined in separate table: initial receipt, packing, in-transit, relocation, delivery, inventory)  Table should work like equipment_type table.
  - created_at (date)

**Question:** If we want the ability to have part of assets inventoried on one schedule and part on another schedule, we may need to add 'inventory_scan_due_date' to the lots table.   I was initially picturing a global "Asset inventory scan" deadline, but now I wonder it this does not need to be specific to each asset (lot).    I can picture an Assets Registry "Set next asset inventory scan date" function which updates only the assets displayed (using existing filters) which would then update the 'inventory_scan_due_date' for those assets.  We could add filters to the Assets Registry that shows assets needing scanning for a specified period -- or sorted by 'inventory_scan_due_date'.

### Interface Changes
- Add new “Asset Inventory Scans” registry menu item
    - Show grid of asset scans
      - Menu: Search options to limit time frame, type of asset, etc
      - Asset action menu: Edit scan entry (which fields should we allow editing?)
      - Asset action menu: Edit scan entry, eg: conditions (edit equipment_condition table)
  - Main [Menu] item to set the default location for scans.  Ideally, this location should be able to be entered via a barcode.
  - Big [Scan Asset] button to create new asset scans
    - Brings up a barcode scan interface
    - Once the barcode is scanned, a pop-up modal allows editing Purpose, condition, etc.  id, asset_id, scanned_by, location.  Note that the exact inventory item and asset (lot) will be identified by this scan.
- In the Assets registry, and an action menu item to show the scan history for the asset

### LINKED ISSUE #6465: Add component to select asset condition
Add an angular component, bhAssetConditionSelect() to simply editing assets.

jmcameron:
This PR does several things related to Assets management
- Moves assignment functions to the Assets Management > Assets Registry page
- Other improvements to the Assets Registry page
- Moves all asset (lot) assignment functions to the Assets Management > Assets Registry page
- Adds Asset Scan Registry page (under Assets Management)
  - Adds ability to add and edit asset scans (in action menu)
  - Adds top-level barcode scanning function to add a new asset scan
- Consolidates all asset-related translation tokens to 'assets.json' files
- Added sample inventory (for assets), assets, and asset scans to bhima_test.
- Adds integration tests for asset scans

Closes https://github.com/IMA-WorldHealth/bhima/issues/6350
Closes https://github.com/IMA-WorldHealth/bhima/issues/6465

Work remaining to be done:
 - Add ability to specify a required date for items to be scanned.  I have a plan for this but want to get this PR in the system before tackling it.
 - Add reports for lists of assets remaining to be scanned
 - Add reports for lists of all assets globally as well as in specific final destinations
 - Eventually refine the concept of locations for scans.  Perhaps use GPS coordinates? (Current "Location" is simply a village)

**TESTING**
- Suggest testing first with bhima_test and then with a production dataset
- If you are not using bhima_test:
  - Make sure there are Contacts (entity) in your database.  Use the Contact management page if necessary. Some production databases are not using Contacts, so this step may be necessary for some sites since Asset assignment is to an entity (contact) which can be a person or place, or department, etc.
  - Use the testing directions from https://github.com/IMA-WorldHealth/bhima/pull/6417 to add a new inventory
  - Add a couple of new assets by doing Stock Entry via Integration and be sure to select the asset inventory type created above.
- Go to the  Assets Management > Assets Registry page
  - Test the various action menu items, the grid functionality, etc
  - Assign one of them to an entity (you created above)
  - Pick an asset and use the Barcode function in the drop-down menu to print the barcode (or copy the code)
- Go to the Assets Management > Assets Scan Registry page
  - Use the big [+New Asset Scan] button and paste in the asset barcode (or scan the printed one).  Fill out the data and submit
  - Test the items in the action menu for the asset scan
  - Review all asset-related French translation items


mbayopanda (commented):
Great features @jmcameron, the assets scan and asset registry are working fine;

I have some suggestions: 
- Add search by date option in the search modal like with most of our registries
- Add a toggle filter button for the assets scan registry
- Add a way to print out the assets scan registry as a report

With the assets registry:
- Fix the quantity in the assignment modal because we can assign any quantity regardless the current quantity of the selected asset lot.

lomamech (commented):
- My concern regarding scanning is what will happen if the base code attached to the Asset becomes invisible or gets lost
- **_Add reports for lists of assets remaining to be scanned_** about this option, concerning unscanned items, I think it would be important to be able to generate with the out of stock module with Loss, the possibility of importing unscanned assets

lomamech:
I think I'm done with the review, Bruce has already suggested about adding the Search by Date option

> Assets management

@jmcameron  What is your reaction to my comment, do you plan to do it in another PR so that we can add another Issue

jmcameron:
> * My concern regarding scanning is what will happen if the base code attached to the Asset becomes invisible or gets lost

I think we can still look up items by serial numbers.  In any case, we will have to figure this out later.

jmcameron:
>     * **_Add reports for lists of assets remaining to be scanned_** about this option, concerning unscanned items, I think it would be important to be able to generate with the out of stock module with Loss, the possibility of importing unscanned assets

Please look at this issue and add these comments or other thoughts about this to it:  https://github.com/IMA-WorldHealth/bhima/issues/6466

lomamech:
> > * My concern regarding scanning is what will happen if the base code attached to the Asset becomes invisible or gets lost
> 
> I think we can still look up items by serial numbers. In any case, we will have to figure this out later.

@mbayopanda  had just explained to me how to reprint the bare code number in BHIMA

jmcameron:
>     * Fix the quantity in the assignment modal because we can assign any quantity regardless the current quantity of the selected asset lot.

Thanks!  I just fixed this.  Note that based on our original discussions about assets, an asset is a singular item, so quantity=1 always.  However if it is not a asset (regular stock), then quantity can be more than one, depending on availability.

So the modal will not let you change the quantity if it is an asset.   Also, I fixed the lot/asset lookup so expired (non-assets) will not appear as an option for lot/asset selection.

I realize that the Stock assignment in the Asset Registry is a bit of strange thing since you can also still assign stock (non-assets), which then do not show up in the assets registry.   I suggest we deal with this later.

lomamech:
> 

OK, no problem for me

PR #6483, opened 2022-03-30
Added asset registry action menu item to show scans for the asset

### LINKED ISSUE #6482: Add Asset Registry action menu item to display scan history
Add action menu item "Show assets scan history" for Assets Registry page that will show all scans for the specific asset (similar to the assignment history modal) -- or go to the Assets Scans page with filtering on that asset.

jmcameron:
Add an action menu item for the Assets Registry that will jump to the Assets Scan Registry and only show the scans for that asset.

Note that this PR removed a couple of unused translation items in the asset.json files.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6482

**TESTING**
- Use bhima_test
- Go to Asset Management > Asset Registry
- In the action menu for an asset, select the "Show Asset Scans"
- This should jump to the "Asset Scans Registry" page but filtered down to the specified asset

- Check the french translation for ASSET.SHOW_ASSET_SCANS


client/src/modules/assets/templates/action.cell.html
```diff
@@ -47,6 +47,12 @@
       </a>
     </li>
 
+    <li>
+      <a data-method="view-scan-history" ng-click="grid.appScope.showAssetScans(row.entity)" href>
```
jniles, line 51: Note that you can also do this with a `ui-sref` and save the need to have an `ng-click` with a scoped binding.  I don't know which is better in terms of performance though.

client/src/modules/assets/templates/action.cell.html
```diff
@@ -68,7 +74,7 @@
     </li>
 
     <li>
-      <a data-method="view-stock-history" ng-click="grid.appScope.openHistoricModal(row.entity.uuid, row.entity.depot_uuid)" href>
+      <a data-method="view-assignment-history" ng-click="grid.appScope.openHistoricModal(row.entity.uuid, row.entity.depot_uuid)" href>
```
jniles, line 77: :+1:

client/src/modules/assets/assets-registry.js
```diff
@@ -277,6 +277,20 @@ function AssetsRegistryController(
...
   };
 
+  /**
+   * Show all the asset scans for this asset
+   * @param {string} uuid - the asset UUID
+   */
+  vm.showAssetScans = (asset) => {
+    $state.go('stockAssetsScans', {
+      filters : [{
+        key : 'asset_uuid',
```
jniles, line 287: Can you add `cacheable : false` to this filter?  It will tell the filter service _not_ to remember this filter in the cache when you press the back button.  That way, a user doesn't have to manually clear it if they were just looking something up quickly.

jniles: You might also want to add a filter on the time period to make it `allTime` or something similar.

jniles (commented):
Just a few suggestions. 😄

client/src/modules/assets/templates/action.cell.html
jmcameron: Thanks.  I'll try it next time.

client/src/modules/assets/assets-registry.js
jmcameron: Added cacheable.  Did not add the period since we currently don't a period filter on the Asset registry.

jniles (approved):
LGTM.

Bors r+

PR #6484, opened 2022-03-31
Filters for required inventory scans registry

### LINKED ISSUE #6480: Add more search filters to Required Inventory Scans registry
Please add search filters for the Required Inventory Scans registry:
- Custom:  reference_number
- Default: period  (this to filter for the end date)

jmcameron:
Added custom filters to Required Inventory Scans registry
- reference number
- period (for end date)

Closes:  https://github.com/IMA-WorldHealth/bhima/issues/6480

Also registry action menu item in the Required Inventory Scans registry to go to the Assets Scans Page and show the scans that satisfy this required inventory.   This is mentioned as part of [Issue 6481](https://github.com/IMA-WorldHealth/bhima/issues/6481) but does not close it.

Changes the default period for the Asset Scans registry and the Required Inventory Scans registry to 'this year'.

**TESTING**
- Use bhima_test
- In Asset Management > Required Inventory Scans
   - Create several new scans with a variety of reference_number and end_dates
   - Use the custom filters to test them
   - Use the action item "Show scanned assets" item to test that function.   Note that the filtering involved handles the period (start of inventory scan to end of inventory scan) as well as the optional Depot.  Test various combinations.

**NOTE:** I had some problems with the Required Inventory Scans registry period default filter (for the end date).  But it seemed to work okay for Chris.   And it seems to work since I did more work on the PR.  But I encourage testing it thoroughly.


client/src/js/services/PeriodService.js
```diff
@@ -52,7 +52,7 @@ function PeriodService(Moment) {
     },
 
     // components will make an exception for all time - no period has to be selected
-    // on the server this simple removes the WHERE condition
+    // on the server this simply removes the WHERE condition
```
jniles, line 55: :+1:

server/controllers/stock/required_inventory_scans.js
```diff
@@ -47,9 +47,12 @@ function getFilters(parameters) {
   filters.equals('title');
   filters.equals('description');
   filters.equals('depot_uuid');
+  filters.equals('reference_number');
```
jniles: Question - should this be an exact match or a `LIKE` match?  I would almost prefer it if it were a `fullText()` search (using `LIKE`) since the column type is `TEXT`.  I can be persuaded otherwise though.

client/src/modules/required_inventory_scans/modals/search.modal.html
```diff
@@ -24,12 +24,27 @@
             <bh-clear on-clear="$ctrl.clear('depot_uuid')"></bh-clear>
           </bh-depot-select>
 
+          <div class="form-group">
+            <label class="control-label" translate>FORM.LABELS.REFERENCE_NUMBER_OPTIONAL</label>
+            <bh-clear on-clear="$ctrl.clear('reference_number')"></bh-clear>
+            <input ng-model="$ctrl.searchQueries.reference_number"
+              class="form-control"
+              type="text"
```
jniles, line 32: Should we put a max limit on this field?

server/config/routes.js
```diff
@@ -148,7 +148,7 @@ const period = require('../controllers/finance/period');
 const lots = require('../controllers/stock/lots');
 
 // assets
-const assets = require('../controllers/stock/asset');
+const assetScans = require('../controllers/stock/asset_scan');
```
jniles, line 151: Good renaming 👍

jniles (approved):
I tested this and it works fine.  Thank you for including testing instructions.

I added one comment.  Let me know what you think - I can be persuaded either way, but wanted to make the comment here since you are changing it.

server/controllers/stock/required_inventory_scans.js
jmcameron: Changed these to fullText filters.   See additional notes on this in the comments.

client/src/modules/required_inventory_scans/modals/search.modal.html
jmcameron: Not now.  We don't know how long these will be since they depend on what external partners want to use.  See more notes on this in the comments below.

jmcameron:
**Regarding the reference_number field:**  It is possible we may have to reconsider this field once we actually start working with external partners based on what they provide in terms of reference numbers/IDs for their purchase orders.    It is also possible that we may need to change these to UUIDs and create a separate SQL table with the textual reference number as the external partner wants it.   This will reduce typing errors when entering/choosing reference_numbers.    However this means some extra machinery to create/edit/manage these external numbers.  It is probably best to avoid doing this until we have a clearer picture of how this will work in practice.

jniles:
Looks good to me.

bors r+

PR #6488, opened 2022-04-01
Prevent expired stock from being added to shipments

### LINKED ISSUE #6471: Creating shipments allows expired stock to be selected
When creating shipments, doing [add] and selecting an inventory code allows lots/assets to be selected that are expired.  Only assets OR unexpired lots should be allowed in the "available lots" so that expired lots cannot be selected.

Reference notes in merged PR https://github.com/IMA-WorldHealth/bhima/pull/6394

jmcameron:
Prevent expired stock to be shipped anywhere from a Depot.  Expired stock must be exited as expired and then is no longer tracked by BHIMA.  Normally it would be disposed.   If for some extraordinary reason, it needs to be shipped somewhere that shipment should occur outside of BHIMA and not using the BHIMA shipment functions.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6471

**TESTING**
- Use bhima_test
- Shipment > Create new shipment
  - Select Principal Depot
  - Set destination depot to Secondary Depot
  - Click [Add] to add a new line for an inventory/lot item
  - Type 'quin' into the inventory code area until it gives you one option for quinine. Select it.
  - Verify that it is not possible to select lot QUININE-A (which is expired)
- Since StockExitFormService is used by both Shipments and Stock Exit, verify Stock Exit still works properly
  - Stock > Stock Exit
    - Click on [LOSS]
    - Click [Add] to add a new line for inventory/lot
    - Type 'quin' into the inventory code and select the quinine inventory item
    - Verify that you can now select QUININE-A even thought it is expired

### DIFF
--- a/client/src/modules/shipment/create-shipment.js
+++ b/client/src/modules/shipment/create-shipment.js
@@ -20,7 +20,9 @@ function CreateShipmentController(
 
   vm.existingShipmentUuid = existingShipmentUuid;
   vm.isCreateState = $state.params.isCreateState;
+
   vm.stockForm = new StockForm('ShipmentForm');
+  vm.stockForm.setAllowExpired(false);
 
   const gridFooterTemplate = `
     <div style="margin-left: 10px;">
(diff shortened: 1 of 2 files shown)


jniles (approved):
Tested this and it works.

bors r+

PR #6496, opened 2022-04-04
Fix form warning messages in shipment creation form

### LINKED ISSUE #6490: Shipment validation isn't always right
In `Shipment > New shipment`, the validation will say that it is ready for submission even if there are missing fields.  In particular, a label is required.  See below for a live example.

image

jmcameron:
Fixes the warning messages in the Shipment creation form to be more helpful.

Made the estimated delivery date required.

Removed the "Exit type set" info message since there is no need for that in the Shipment form since it is always a transfer to a depot.  It was a bit confusing before.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6490

**NOTE:** One thing that was not addressed in this PR is whether selecting a condition for each item is required before submitting.   If it is not set, the form can be submitted, but the condition for the item is null for the resulting shipment.   We should think about this.   And it is not clear that the condition is needed as part of the shipment itself since that could be tracked with asset scans.

TESTING
- bhima_test
- Shipments > Create Shipment
  - Try clicking on the [Submit] button immediately
  - Try different orders of creating the items and filling out the form and make sure the warnings make sense.
- Then try Stock Exit (which usese the same StockExit


client/src/modules/shipment/create-shipment.js
```diff
@@ -268,11 +269,24 @@ function CreateShipmentController(
...
+    if (form.$invalid) {
+      vm.validateItems();
+
+      // Prepend general form needs fixing message
+      const formMsg = {
+        type : 'warn',
+        text : 'STOCK.MESSAGES.WARN_MAIN_FORM_ERRORS',
+        keys : {},
+      };
+      vm.messages.unshift(formMsg);
```
jniles, line 282: This is a creative way of solving this issue

client/src/modules/stock/StockExitForm.service.js
```diff
@@ -67,6 +69,14 @@ function StockExitFormService(
     this.allowExpired = flag;
   };
 
+  /**
+   * Set flag that the exit type is predefined
+   * @param {boolean} flag
+   */
+  StockExitForm.prototype.setExitTypePredefined = function setExitTypePredefined(flag) {
```
jniles, line 76: Cool!

jniles (approved):
Looks good to me.

bors r+

PR #6502, opened 2022-04-06
fix(stock): remove adjustments

### LINKED ISSUE #6501: Stock > Adjustments are broken on master
The adjustments module depended on templates in the stock exit module that are now broken.  Therefore, we'll need rewrite this module to get it functional again.

However, I wonder if we still need stock adjustments. The early version of stock adjustments existed to increase or decrease the value in stock of inventory items.  This can be done with the `Stock > Stock Adjustments` module, both to decrease lots in stock and increase lots in stock.  I've tested both scenarios, and as long as we don't have a lot of lots, it works fine.

jniles:
This commit removes the Adjustments module that is superseded by the Stock Adjustments module.  All the operations  performed by this module are already performed by other modules.

Closes #6501.


server/models/migrations/next/migrate.sql
```diff
@@ -252,4 +252,12 @@ CREATE TABLE IF NOT EXISTS `required_inventory_scan` (
...
\ No newline at end of file
+  (313, 'Inventory Scans Management', 'TREE.REQUIRED_INVENTORY_SCANS', 'Inventory Scans Management', 307, '/required/inventory/scans');
+
+/**
+Issue: 6501
+@author: jniles
+@date: 2022-04-06
+*/
+DELETE FROM role WHERE unit_id = 167;
+DELETE FROM `unit` WHERE id = 167;
```
jmcameron, line 263: 👍

jmcameron (approved):
Code changes look good.
Tested it and it works well.   Glad you removed the old "Adjustments" page.  I think the "Inventory Adjustment" page is much more intuitive.

I plan to add an Issue to add a Search option to this page to allow filtering the items to be listed.  I imagine that the list is very long and may be difficult to use in some situations.

server/models/migrations/next/migrate.sql
```diff
@@ -252,4 +252,12 @@ CREATE TABLE IF NOT EXISTS `required_inventory_scan` (
...
-  (313, 'Inventory Scans Management', 'TREE.REQUIRED_INVENTORY_SCANS', 'Inventory Scans Management', 307, '/required/inventory/scans');
\ No newline at end of file
+  (313, 'Inventory Scans Management', 'TREE.REQUIRED_INVENTORY_SCANS', 'Inventory Scans Management', 307, '/required/inventory/scans');
+
+/**
+Issue: 6501
+@author: jniles
+@date: 2022-04-06
+*/
+DELETE FROM role WHERE unit_id = 167;
```
jmcameron, line 262: Shouldn't this be DELETE FROM permission WHERE unit_id = 167; ???

jniles: Oh, that is probably true.  Sorry.

PR #6503, opened 2022-04-06
feat(suppliers): add inline filter

### LINKED ISSUE #6400: We should add inline filters to the supplier module
image

This registry could use an inline filter.

jniles:
This commit adds an inlinef ilter to the supplier module.

Closes #6400.

image


jmcameron (approved):
Code looks good. 
Tested it and it works fine.

PR #6504, opened 2022-04-06
fix(stock): improve Stock Value Report

### LINKED ISSUE #6150: Updates for Stock Value Report
Stock > Report > [Stock] Stock Value Report

There seems to be a bug:  If you do the report in the enterprise currency, some stock articles are marked in pink.  But if you do the report in some other currency different rows are shaded pink!

**Suggestions**
- ~~Fix name:   [Stock] Stock Value Report => Stock Value Report~~
- The currency should default to the enterprise currency
- The report is very slow.  With the Vanga database and on my laptop, it takes nearly 20 seconds to generate the report.
- The depot and currency are not cached for this report, so they are cleared every time the report is revisited.
- It is not clear why some rows are highlighted in pink.  The report needs to be fixed or updated to make this clearer (perhaps add a column with status: near expiration/out of stock/etc).  Notice the first two rows have different colors but otherwise seem similar (eg, 0 count): 
image
- If the "Exclude inventories with zero stock value" option is selected, it might be a good idea to add a footnote at the end of the report indicating that.

jniles:
This PR fixes various components of the stock value report cited in issue #6150. 

Notably, the query now uses the `stock_movement_status` table to pull in the quantities in stock for a given depot and loads the
`stock_value` table to get the value of the items in stock.  This speeds the query up significantly.

Miscellaneous changes:

- The report now defaults to the enterprise currency.
- The majority of the values are now cached, except for the date that defaults to today.
- Performance improvements as cited above.
- I've removed the coloration entirely.  

Closes #6150.

### DIFF
--- a/client/src/modules/reports/generate/stock_value/stock_value.config.js
+++ b/client/src/modules/reports/generate/stock_value/stock_value.config.js
@@ -4,12 +4,12 @@ angular.module('bhima.controllers')
 StockValueConfigController.$inject = [
   '$sce', 'NotifyService', 'BaseReportService',
   'AppCache', 'reportData', '$state',
-  'LanguageService', 'moment',
+  'LanguageService', 'moment', 'SessionService',
 ];
 
 function StockValueConfigController(
   $sce, Notify, SavedReports,
-  AppCache, reportData, $state, Languages, moment,
+  AppCache, reportData, $state, Languages, moment, Session,
 ) {
 
   const vm = this;
@@ -19,6 +19,7 @@ function StockValueConfigController(
   vm.reportDetails = {
     dateTo : new Date(),
     excludeZeroValue : 0,
+    currency_id : Session.enterprise.currency_id,
   };
 
   // Default values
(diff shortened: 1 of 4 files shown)


client/src/modules/reports/generate/stock_value/stock_value.config.js
```diff
@@ -100,6 +88,9 @@ function StockValueConfigController($sce, Notify, SavedReports,
   };
 
   function checkCachedConfiguration() {
-    vm.reportDetails = angular.copy(cache.reportDetails || {});
+    if (cache.reportDetails) {
+      vm.reportDetails = angular.copy(cache.reportDetails);
+      vm.dateTo = new Date(); // always default to today
+    }
```
jmcameron, line 95: Need to add a default for the currency here.   See the checkCachedConfiguration() function in client/src/modules/reports/generate/avg_med_costs_per_patient/avg_med_costs_per_patient.js for an example.

jmcameron (changes requested):
Code looks good.
If you clear the browser cache and go to the page, the default for the currency is not set.    Suggested a fix in the code comments.

client/src/modules/reports/generate/stock_value/stock_value.config.js
jniles: Done.

jmcameron (approved):
Works fine now.

PR #6507, opened 2022-04-07
Removed condition from shipment items

### LINKED ISSUE #6497: Are conditions for items necessary in the Shipments?
As noted in [PR 6496](https://github.com/IMA-WorldHealth/bhima/pull/6496), when creating shipments, selecting a condition for each item is required before submitting. If it is not set, the form can be submitted, but the condition for the item is null for the resulting shipment. We should think about this. And it is not clear that the condition is needed as part of the shipment itself since that could be tracked with asset scans.

I suggest that we remove these from the shipment (item) table and shipment forms and rely on tracking asset conditions from by asset scans.   The current plan is to scan each item as it is placed in a shipment, so that should provide this information.

jmcameron:
Removed shipment_id from shipment items.  The condition of items will be tracked via asset scans.

Also fixed an issue in the shipment.json files that was preventing the shipment status options from being translated in Shipments registry Search function.

This PR also fixed some issues with previous merged PRs that were preventing migrations from working properly.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6497

**TESTING**
- Create a shipment.  Notice that the condition is no longer needed for inventory added to the shipment.
- Exercise the full cycle
  - Create  new shipment (from principle depot to secondary)
  - Add item(s)
  - Submit
  - Go to Shipments Registry and mark the shipment as ready to go.
  - Do stock exit (Depot --> Select Secondary depot) for the shipment items
  - Do Stock entry (change depot to secondary depot) from shipment
- Verify the operations above


jmcameron:
Per discussion with @jniles, I will take the unusual step of approving this one myself.
I did test it again, carefully with bhima_test and worked fine.
I also tested doing a migration and ran into some previously existing problems -- which are fixed in this PR.
-Jonathan Cameron

PR #6545, opened 2022-04-12
Convert lots registry custom search flag from Assets-only to Exclude-assets

### LINKED ISSUE #6532: Change Assets-only flag in Stock Lots registry
Change the logic in the Stock Lots registry custom search modal from "Assets Only" to "Exclude Assets"

-Jonathan

jmcameron:
Changes the logic of the "Assets only" custom search query flag to "Exclude Assets".   To look at only assets, the user should use the Assets Managements > Assets Registry page.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6532

**TESTING**
- use bhima_test
- Try Stock > Stock Lots and try the "Exclude assets" flag near the bottom of the search query modal.  When set, notice that the two assets (MOT1 and MOT2) are no longer listed.

### DIFF
--- a/client/src/i18n/en/asset.json
+++ b/client/src/i18n/en/asset.json
@@ -10,6 +10,7 @@
     "ASSIGN_TO" : "Assign to",
     "DETAILS" : "Asset Details",
     "DOCUMENT" : "Assignment Document",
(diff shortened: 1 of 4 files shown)


jniles (approved):
LGTM.

bors r+

PR #6549, opened 2022-04-13
Fix several problems in create shipment form

### LINKED ISSUE #6522: Creating Shipment should filter out current depot from Destination depot
When creating a shipment, the drop-down for the Destination Depot should filter out the current depot.

### LINKED ISSUE #6525: Shipment expected delivery date cannot be in the past
When creating/editing a shipment, cannot allow the destination date to be in the past.

### LINKED ISSUE #6526: Assets in shipments sould default to quantity = 1
When creating shipments, if an inventory item is an asset, the quantity for the lot should default to 1

jmcameron:
Fixes several minor usability problems in the form to create shipments.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6522  -- Filter out source depot from destination depot drop-down
Closes https://github.com/IMA-WorldHealth/bhima/issues/6525  -- Shipment expected delivery date cannot be in the past
Closes https://github.com/IMA-WorldHealth/bhima/issues/6526  -- Assets in shipments sould default to quantity = 1

Also fixed a bug that was preventing updating quantity from automatically validating the row in form.

**TESTTING**
- Use bhima_test
- Try creating a shipment
   -  Assets should default to quantity = 1 (non-assets to 0)
   - Try changing the quantity (for non-asset) and see the row validation work immediately
   - Verify that a expected delivery date in the past cannot be selected
   - Verify that the source depot does not appear in the drop-down list for selecting the destination depot
   - Choose the source depot, set the destination depot, and then change the source depot. Verify that the destination depot is cleared (since it may not be valid any longer).

### DIFF
--- a/client/src/modules/shipment/create-shipment.js
+++ b/client/src/modules/shipment/create-shipment.js
@@ -65,7 +65,7 @@ function CreateShipmentController(
         width : 150,
         displayName : 'TABLE.COLUMNS.QUANTITY',
         headerCellFilter : 'translate',
-        cellTemplate : 'modules/stock/exit/templates/quantity.tmpl.html',
+        cellTemplate : 'modules/shipment/templates/quantity.tmpl.html',
         aggregationType : uiGridConstants.aggregationTypes.sum,
       }, {
         field : 'unit_type',
@@ -104,7 +104,7 @@ function CreateShipmentController(
 
   vm.maxLength = util.maxLength;
   vm.enterprise = Session.enterprise;
-  vm.maxDate = new Date();
+  vm.today = new Date();
   vm.onChangeDepot = onChangeDepot;
   vm.getOverview = getOverview;
   vm.setReady = setReady;
@@ -127,6 +127,10 @@ function CreateShipmentController(
   };
 
   vm.configureItem = function configureItem(row, lot) {
+    if (lot.isAsset()) {
+      // Override default quantity for assets
+      lot.quantity = 1;
+    }
     vm.stockForm.configureItem(row, lot);
     vm.validateItems();
   };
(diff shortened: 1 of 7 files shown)


client/src/modules/shipment/create-shipment.js
```diff
@@ -65,7 +65,7 @@ function CreateShipmentController(
         width : 150,
         displayName : 'TABLE.COLUMNS.QUANTITY',
         headerCellFilter : 'translate',
-        cellTemplate : 'modules/stock/exit/templates/quantity.tmpl.html',
+        cellTemplate : 'modules/shipment/templates/quantity.tmpl.html',
```
jniles, line 68: Nice 👍

client/src/modules/shipment/create-shipment.js
```diff
@@ -127,6 +127,10 @@ function CreateShipmentController(
   };
 
   vm.configureItem = function configureItem(row, lot) {
+    if (lot.__is_asset === 1) {
```
jniles: Does `lot.isAsset()` exist?  It would be better to use that.

jniles (approved):
Tested and found working.  I've made one comment that you can take or leave.

Looks good to me.

client/src/modules/shipment/create-shipment.js
jmcameron: I had tried isAsset() and it did not work.  When I reexamined it, I realized that the isAsset() test in LotItemService was faulty because when we use the is_asset as a checkbox element in the html, its values are 1 and 0.  So the isAsset() test that did val === true did not work properly.  I changed it to use the !! truth value and it works now for any truthy value.  So I was able to change this to isAsset() and it works now.

PR #6553, opened 2022-04-15
Fix issues with stock exit shipment selection, improve entry/exit documents

### LINKED ISSUE #6528: Shipment Registry label
In the shipment registry, there is a column called "Stock Reference".   This is actually a "Stock Movement Reference" and should be renamed appropriately.

jmcameron:
Fix usability issues when selecting shipments for Stock Exits.
Added shipment information to stock entry and exit documents.
Minor fix to shipment registry column label

Closes https://github.com/IMA-WorldHealth/bhima/issues/6528

**TESTING**
- Use bhima_test
- Create a shipment from Primary to Secondary
- In the Shipment registry, mark the shipment as "Ready to Ship" using the action menu
- Go to Stock > ~Stock Entry~ Stock Exit
  - Make sure the Primary Depot is selected
  - Click on the [Depot] button to bring up the modal to select the depot
  - For the destination depot, chose the Tertiary Depot
  - Click on "Have shipment" - Yes.  Notice the Secondary depot as overridden the destination depot and it is disabled
  - Click on "Have shipment" - NO.  The destination depot is erased and can be changed again
  - Select the shipment again and save the Stock Exit.  
  - Notice the stock exit report includes the Shipment reference (right above the To/From depots)
- Go to Stock > Stock Entry
  - Change the depot to the Secondary Depot (via the Menu button)
  - Click on [Transfer] and select the shipment for the just-completed Stock exit
  - Complete the stock entry
  - Notice the stock entry report includes the Shipment reference (right above the To/From depots)
- Go to the Stock > Stock Movements page
  - Using the action menu on the stock movements just completed, verify that the reports include the shipment reference
- Go to Shipments > Shipment Registry.  Verify the label for the third column is now "Stock Movement Reference" instead of "Stock Reference"

**NOTE:**
- All stock exits to Depot and stock entry by Transfers create/use a shipment now, whether the exit created from a Shipment voucher or not.   You should do a basic stock exit and stock entry via depot/transfer (without explicitly using a shipment), then check the Shipment registry and Stock movements to see this.   I think this is make sense and is okay, but I was a little surprised to realize that this was happening behind the scenes for all stock transfers between depots.

### DIFF
--- a/client/src/js/components/bhHasRequisitionVoucher/bhHasRequisitionVoucher.js
+++ b/client/src/js/components/bhHasRequisitionVoucher/bhHasRequisitionVoucher.js
@@ -4,10 +4,11 @@ angular.module('bhima.components')
     controller  : bhHasRequisitionVoucherController,
     bindings    : {
       message   : '<',
-      classeLabel : '<',
+      classLabel : '<',
       reference : '<',
       onChange  : '&',
       requestor : '<?',
+      required : '<',
     },
   });
 
--- a/client/src/modules/stock/exit/modals/findDepot.modal.html
+++ b/client/src/modules/stock/exit/modals/findDepot.modal.html
@@ -15,7 +15,8 @@
   <div class="modal-body">
     <div class="form-group">
       <label class="control-label" translate>STOCK.DEPOT</label>
-      <ui-select name="depot" ng-model="$ctrl.selected">
+      <ui-select name="depot"
+        ng-disabled="$ctrl.shipment">
         <ui-select-match>
           <span>{{$select.selected.text}}</span>
         </ui-select-match>
@@ -27,7 +28,7 @@
 
     <bh-has-requisition-voucher
       message="$ctrl.requisitionMessage"
-      classe-label="$ctrl.requisitionLabel"
+      class-label="$ctrl.requisitionLabel"
       reference="$ctrl.reference"
       requestor="$ctrl.selected.uuid"
       on-change="$ctrl.onChangeReference(reference)">
@@ -35,7 +36,7 @@
 
     <bh-has-shipment-voucher
       message="$ctrl.requisitionMessage"
-      classe-label="$ctrl.requisitionLabel"
+      class-label="$ctrl.requisitionLabel"
       reference="$ctrl.shipmentReference"
       depot-uuid="$ctrl.depot.uuid"
       on-change="$ctrl.onChangeShipmentReference(shipment)">
@@ -47,7 +48,9 @@
       FORM.BUTTONS.CLOSE
     </button>
 
-    <bh-loading-button loading-state="FindForm.$loading">
+    <bh-loading-button
+      disabled="!$ctrl.depotSelected()"
+      loading-state="FindForm.$loading">
       <span translate>FORM.BUTTONS.SUBMIT</span>
     </bh-loading-button>
   </div>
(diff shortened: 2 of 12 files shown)


client/src/modules/stock/exit/modals/findDepot.modal.html
```diff
@@ -27,15 +29,15 @@
...
+      class-label="$ctrl.requisitionLabel"
       reference="$ctrl.reference"
       requestor="$ctrl.selected.uuid"
       on-change="$ctrl.onChangeReference(reference)">
     </bh-has-requisition-voucher>
 
     <bh-has-shipment-voucher
       message="$ctrl.requisitionMessage"
-      classe-label="$ctrl.requisitionLabel"
+      class-label="$ctrl.requisitionLabel"
```
jniles, line 39: I agree with this change, but you'll need to change the component definitions as well to fix the `classe` typo.  Can you do that?

jniles (commented):
Just a quick fix here.

jmcameron:
@jniles As well as fixing the classe-label renaming in the component, I also adjusted the findDepotModal to prevent submitting without selecting a modal (by disabling the Submit button).    Please try it and make sure you're okay with the UI change.

client/src/js/components/bhHasRequisitionVoucher/bhHasRequisitionVoucher.js
```diff
@@ -4,10 +4,11 @@ angular.module('bhima.components')
     controller  : bhHasRequisitionVoucherController,
     bindings    : {
       message   : '<',
-      classeLabel : '<',
+      classLabel : '<',
```
jniles, line 7: :+1:

jniles:
There is a weird issue with the descriptions importing from shipment that I will document here:

image

I don't think it is tied to your PR though.

I didn't select a Honda as part of the shipment.  So something is breaking in the "import from requisition" feature.  It also would be helpful for us to change the text to reference "shipments" instead of requisition.

server/controllers/stock/reports/stock_entry_depot.receipt.handlebars
```diff
@@ -14,6 +14,7 @@
 
     <!-- summary  -->
     <div class="row receipt-summary">
+      <h4 style="margin: 0 0 5px 0; text-align: center">{{translate 'SHIPMENT.TITLE'}}: {{entry.details.shipment_reference}}</h4>
```
jniles, line 17: As it is, this looks somewhat out of place.  Can you make it fit the same style as the other references?

image

My thought is to move it down below and do something like:

```html
<span class="text-capitalize">{{translate 'SHIPMENT.TITLE'}}</span>: <strong>{{entry.details.shipment_reference}}</strong> <br>
```

Note that this should only be active is a `shipment_reference` exists.

jniles (commented):
Just a few more minor corrections.

client/src/modules/stock/exit/modals/findDepot.modal.html
jmcameron: Fixed

jmcameron:
> There is a weird issue with the descriptions importing from shipment that I will document here:
> I don't think it is tied to your PR though.

I agree that it is not related to this PR.  I'll add an issue for it.

server/controllers/stock/reports/stock_entry_depot.receipt.handlebars
jmcameron: I added the #if for the shipment_reference per your suggestion.   You can see this in operation with bhima_test, by looking at the document for an older stock exit in Stock Movements.  Since no shipments were created for these stock exits, the "Shipment" label is suppressed.  But when a new stock exit is created (with or without explicit shipment), the document will include the shipment ID.

jniles (approved):
LGTM.  Bors r+

PR #6560, opened 2022-04-19
fix(stock): added creation success key

### LINKED ISSUE #6512: Text on stock movement creation not displaying
There should be "stock movement created successfully" written in the red box.

image

jniles:
Adds the creation success key to all stock movements.

Closes #6512.

image

### DIFF
--- a/client/src/i18n/fr/stock.json
+++ b/client/src/i18n/fr/stock.json
@@ -281,7 +281,7 @@
     "STOCK_VALUE_WITH_WAC" : "Valeur finale du stock avec le cout unitaire de l'entreprise",
     "STOCK_MOVEMENT"  : "Voir le mouvement de Stock",
     "STOCK_MOVEMENTS" : "Les Mouvements de Stock",
-    "SUCCESS"         : "Mouvement realisé avec succès",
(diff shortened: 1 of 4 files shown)


client/src/i18n/fr/stock.json
```diff
@@ -249,7 +249,8 @@
       "EXIT_DEPOT"     : "Transfer vers un dépot",
       "EXIT_SERVICE"   : "Distribution à un service",
       "EXIT_SERVICES"   : "Distributions aux services",
-      "EXIT_LOSS"      : "Perte de stock"
+      "EXIT_LOSS"      : "Perte de stock",
+      "SUCCESS" : "Mouvement créé avec succès"
```
jmcameron, line 253: So the French is a little different than the English for this.  Do you want to add "stock" in to make it clearer?

jmcameron (approved):
Looks good.  I tested it with bhima_test and exercised every stock entry and stock exit and the message was shown in every case.
I am approving.    I think the French translation could be improved but otherwise it is ready to go!

client/src/i18n/fr/stock.json
jmcameron: Fixed

jmcameron (approved):

PR #6561, opened 2022-04-19
Shipment registry action menu items

jmcameron:
Added shipment registry action menu items to jump to Stock Exit for the shipment (sending depot) and Stock Entry (receiving depot).

Implements the Stock Exit side of https://github.com/IMA-WorldHealth/bhima/issues/6529

NB: The action item menu for doing the Stock Entry is shown but has not been implemented yet, so it does nothing.  That will be done in the next PR.

Note that the stock exit action is only shown when the shipment is marked "Ready to ship":
image

Similarly, the stock entry action is only shown after the stock exit has been performed for the shipment:
image

**TESTING**
- use bhima_test, reload the data set
- Create a shipment with a couple of items
- Mark the shipment ready to ship
- Use the action menu to do the stock exit.  it will jump to the Stock Exit for that shipment, all prepared for final checks before completing the exit.
- Notice that there is no error message on the left
- Clear the Stock Exit (with the [Clear] button at the bottom).
- Do a Stock Loss and exhaust one of the items in the shipment
- Go back to the Shipment Registry and try the stock exit again.
- Notice on the Stock Exit page, one of the items in the shipment has been removed (because not stock was available for it). 
- Also note the warning message on the left

**NOTE**
- Also suggest doing some other types of stock exits to verify that the changes for shipments have not broken anything.


jmcameron:
bors try

jmcameron:
The integration tests pass on my computer.

jniles:
Node 18 did just get released, so that might be the cause.

jmcameron:
bors try

PR #6568, opened 2022-04-22
Fix editing shipments (updating packing list)

### LINKED ISSUE #6559: Updating shipment packing list doesn't preserve the destination depot
When you click "update packing list" on the shipments registry, the data is imported into the shipments module, but it doesn't save the destination depot_uuid.  This causes you to have to reset it and potentially reset the whole form if you are not careful!

jmcameron:
This PR fixes an issue with editing a shipment (eg, "update packing list" in the action menu of the Shipments registry") that was losing the destination depot in the editing process.

This was a difficult one to track down due to the interaction of the callback functions that were being invoked when the shipment info was loaded.   A part of the fix was to prevent the display of the editing page until the startup was completed (eg, when the shipment info had been fully installed).

Closes https://github.com/IMA-WorldHealth/bhima/issues/6559

**TESTING**
- Use bhima_test
- Create a shipment.  Do NOT mark it ready for shipping.
- In the shipments registry, use the action menu item "Updated Packing List" to edit the shipment.
- Verify that the Destination depot shown and is correct
- Edit the shipment and save it.
- Try editing it again to double-check.
- Create another shipment but do not mark it ready for shipping
- Log out of Bhima, clear the cache, and log back in
- Go to Shipments > New Shipment and notice the popup to select a origin Depot 
   - DO NOT Select a depot. Just cancel.
- Go to the Shipments > Shipments Registry page
- Use the action menu to edit the last shipment created.  Note that there is no prompt to select a depot, since that is determined from the shipment info

**Notes**
- changing the origin depot deletes all items since these may not be available in a different depot.  If needed, they should be re-added.  NOTE: Will add an issue to prevent changing the origin depot.
- Changing the destination depot does not delete the packing list items


jmcameron:
bors try

client/src/js/components/bhChangeDepot.js
```diff
@@ -5,7 +5,10 @@ angular.module('bhima.directives')
       <i class="fa fa-archive"></i> <span translate>DEPOT.CHANGE</span>
     </a>`,
     controller : bhChangeDepotController,
-    bindings    : { onSelect : '&' },
+    bindings    : {
+      onSelect : '&',
+      skipInitialLoad : '<?',
```
jniles, line 10: Why would one want to skip the initial load?  Wouldn't we just want to pass in the `depotUuid` and then use that to load the relevant information?

client/src/modules/shipment/create-shipment.js
```diff
@@ -182,11 +176,11 @@ function CreateShipmentController(
     vm.depot = depot;
     vm.stockForm.setDepot(depot);
 
-    // set the shipment origin
     vm.shipment.origin_depot_uuid = vm.depot.uuid;
 
-    // Delete the old destination depot
-    delete vm.shipment.destination_depot_uuid;
+    if (vm.isCreateState || (vm.shipment.destination_depot.uuid === depot.uuid)) {
```
jniles, line 186: I'm not sure I understand this logic.  Can you leave a comment to explain it?

client/src/modules/shipment/shipment.service.js
```diff
@@ -86,20 +64,5 @@ function ShipmentService(Api, $httpParamSerializer, Languages) {
     return $httpParamSerializer(options);
   };
 
-  service.readAll = (uuid, parameters) => {
```
jniles, line 89: Nice 👍

server/controllers/asset_management/shipment/shipment.js
```diff
@@ -550,9 +550,9 @@ async function lookup(identifier) {
       dm.text AS reference,
       dm2.text AS stock_reference,
       d.text AS origin_depot,
-      BUID(d.uuid) AS origin_depot_uuid,
+      BUID(sh.origin_depot_uuid) AS origin_depot_uuid,
```
jniles, line 554: Nice 👍

jniles (commented):
Tested this and works as advertised.

client/src/js/components/bhChangeDepot.js
jmcameron: The initial load does onChangeDepot() which creates other problems.  So I added this work-around to make it work nicely.   When we are editing an shipment, the shipment information has been loaded, including loading the origin depot and shipment items, so we don't want to do a fresh reload of the source depot.

client/src/modules/shipment/create-shipment.js
jmcameron: Done.  Note that I plan to add an issue to prevent changing the origin depot on a shipment.   Then this code should be revisited and cleaned up.

PR #6573, opened 2022-04-25
Fix select shipment dropdown

### LINKED ISSUE #6572: In Stock Exit, disable shipment drop-down when no shipments are available
When doing a stock exit and you click on the "Do you have a ready shipment?" option, if there are no shipments available, the selection drop-down should be disabled.  This should behave like "Do you have a requisition slip ?" when there are no requisitions available.

jmcameron:
In Stock Exit, when you click on the [Depot] item at the top, one of the options is to select a shipment.  This PR made it so that the drop-down for it is disabled if there are no shipments that are ready for Stock Exit.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6572

**TESTING**
- Use bhima_test and 'yarn build:db' to refresh the data
- Go to Stock > Stock Exit, Select the primary depot
- Click on the [Depot] button at the top
- Click on the Yes for "Do you have a shipment ready?" item
- Note that the drop-down for it is disabled
- Go to Shipments > New Shipment
  - Create a new shipment
  - Mark it ready to ship
- Go to Stock > Stock Exit again, click on the [Depot] button at the top
  - Note that you could have done this through the action menu in the shipment registry but you would not be able to check this issue since it automatically selects the shipment, etc.
- Click on the Yes for "Do you have a shipment ready?" item
- Note that the drop-down for it is now enabled and you can choose the shipment you just created
- Complete the stock exit


jmcameron:
I tested this again and it works correctly.

bors r+

PR #6574, opened 2022-04-25
Add barcode scans for stock assignment

### LINKED ISSUE #6534: Asset assignments should support barcodes to find lots
Update the assets registry to support a new button to allow using a barcode to identify the item to be assigned.  Part of the process of doing assignments this way would be to guarantee that the item is present (since it was physically scanned).   We could also automatically add a new asset scan record at the same time.

jmcameron:
Add barcode option for selecting the asset/lot to be assigned.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6534

**TESTING**
- Use bhima_test (reload it)
- Go to:  Asset Registry
- Use the action menu to get the barcode for MOT2, scan/print/note the barcode for later use
- Click on the [New Assignment] button to the left of the [Menu] button
- On the modal that comes up, click on [Barcode] button.  This prompts for a barcode to scan
- Scan the previously saved barcode
   - It should fill in the form except for the person/entity to assign it to
   - Click [Submit] to save the assignment
- In the asset registry, note that MOT2 has the new assignment
- Repeat the procedure above for MOT2. This time it will not allow you to make the assignment (since the asset is already assigned).  Read the warning message that comes up.

### DIFF
--- a/client/src/i18n/en/asset.json
+++ b/client/src/i18n/en/asset.json
@@ -45,6 +45,7 @@
     "TITLE" : "Assets",
     "VIEW_IN_INVENTORY" : "View Asset in Inventory",
     "VIEW_IN_STOCK" : "View Asset in Stock",
+    "WARN_ALREADY_ASSIGNED" : "This asset/lot is already assigned. Remove the assignment before reassigning it !",
     "CONDITION" : {
       "NEW" : "New",
       "GOOD" : "Good",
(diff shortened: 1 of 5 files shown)


jmcameron:
I tested this and it works correctly.

bors r+

jmcameron:
Sorry about the strange name for the branch...

PR #6576, opened 2022-04-26
fix(stock): only exit expired stock on loss

### LINKED ISSUE #6570: Stock Exit to Depot allows expired stock to be chosen
When you do a Stock Exit and choose an exit to a Depot and then [Add] an item, if the inventory has expired items, you can select them for stock exit.  This should not be possible.  But Stock Exit to Loss should allow expired items!

jniles:
Fixes the stock exit logic to only exit expired stock on loss.

Closes #6570.

### DIFF
--- a/client/src/modules/stock/StockExitForm.service.js
+++ b/client/src/modules/stock/StockExitForm.service.js
@@ -50,8 +50,10 @@ function StockExitFormService(
     this.cache = AppCache(cacheKey);
     this.details = { is_exit : 1 };
     this.store = new Store({ identifier : 'uuid', data : [] });
-    this.allowExpired = true;
     this.exitTypePredefined = false;
+
+    // this flag will filter out lots from the display that contain expired elements
+    this.allowExpired = true;
 
     // this variable is private and will contain the stock for the current depot.
     this._pool = new Pool('lot_uuid', []);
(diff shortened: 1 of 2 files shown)


client/src/modules/stock/StockExitForm.service.js
```diff
@@ -195,6 +193,7 @@ function StockExitFormService(
   StockExitForm.prototype.listLotsForInventory = function listLotsForInventory(inventoryUuid, lotUuid) {
     const available = this._pool.list()
       .filter(row => row.inventory_uuid === inventoryUuid)
+      .filter(row => (this.allowExpired ? true : !row.isExpired(this.details.date)))
```
jmcameron, line 196: 👍

client/src/modules/stock/exit/exit.js
```diff
@@ -182,6 +183,13 @@ function StockExitController(
       break;
     }
 
+    // only allow expired stock if we are exiting to stock loss
+    if (exitType.label === 'loss') {
+      vm.stockForm.setAllowExpired(true);
+    } else {
+      vm.stockForm.setAllowExpired(false);
+    }
+
```
jmcameron, line 192: Nice!

jmcameron (approved):
The code looks good.
The decision to only allow expired stock when doings exits to Loss makes sense to me.
I tested it and it works.

PR #6583, opened 2022-04-28
fix(shipments): filter out used assets

### LINKED ISSUE #6523: Logic bug in Shipment allows reuse of assets in other shipments
image

jniles:
This PR updates the shipments to ensure that used lots are not shows in the list, or reduced by the quantity used.  Further, I've taken the liberty to rename the "affected" to "assigned" on both the client and the server as well as fix some of the asynchronous logic.

Finally, the logic is updated to make sure we don't refresh the lots when we change the date.

Closes #6523.

### DIFF
--- a/client/src/js/components/bhDepotSelect/bhDepotSelect.js
+++ b/client/src/js/components/bhDepotSelect/bhDepotSelect.js
@@ -55,7 +55,7 @@ function DepotSelectController(Depots, Notify) {
 
     const options = {
       text : (text || '').toLowerCase(),
-      exception : $ctrl.exception,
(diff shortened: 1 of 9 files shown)


jmcameron (approved):
The code looks good.
I tested it and it does prevent one shipment from using items selected in previous shipments (when the Stock Exit has not been done for the first shipment).

However, this does not prevent A stock exit from using stock in unrelated Shipments (that have not done their Stock Exit yet).   Not sure if that is easy to fix.  But current code will prevent a Stock Exit for shipment from using unavailable stock so that issue is handled to some degree.

PR #6584, opened 2022-04-28
Improve shipment tracking logging

### LINKED ISSUE #6581: Add more shipment tracking log entries automatically
The shipment tracking log would be more useful if typical shipment actions inserted new tracking log entries automatically, such as:
- Any change in shipment status including
  - Stock Exit
  - Stock Entry

### LINKED ISSUE #6582: Add shipment registry action menu item to mark the shipment as complete
We currently have no way to indicate that a shipment is completed if something was lost in transit.   Once a shipment has been partially received, we should have a way to mark it as completed even if we have not received everything.

We should add a (conditional) action menu item to the shipment registry item action menus that would allow the shipment to be marked as complete.   Note that this option would only be available if the shipment is in "PARTIAL" state.

jmcameron:
Improve the shipment tracking logging
- Automatically adds a shipment tracking node when you create, edit, mark ready to ship, do stock exit, do stock entry, and complete a shipment.
- Enable the user to add shipment tracking notes at any time
- Adds option to mark a partially delivered shipment as complete (with appropriate tracking log note).  This is for cases where some items were lost in transit.
- When stock exits or entries are invoked from the shipping registry, we automatically return to the registry when the action is complete.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6582
Closes https://github.com/IMA-WorldHealth/bhima/issues/6581

**TESTING**
- Create a shipment and do each of these steps. After each step, check the "Tracking log" from the action menu to observe the notes
   - Create
   - Edit (update packing list)
   - Mark ready to ship
   - Stock Exit
   - Stock Entry - PARTIAL
   - Stock Entry - for the rest of the items.  This will mark the shipment as complete and add an entry in the tracking log
- Repeat the process above, but after doing a partial entry, mark the shipment as complete and check the tracking log


jmcameron:
The integration tests found a real problem.  Working on it.

client/src/modules/shipment/modals/shipment-completed.modal.html
```diff
@@ -0,0 +1,36 @@
+<form
+  name="ModalForm"
+  ng-submit="$ctrl.submit(ModalForm)"
+  data-modal="stock-inventories-search"
```
jniles, line 4: This text could probably be updated - I don't think this is the stock inventories search.

client/src/modules/stock/entry/entry.js
```diff
@@ -31,6 +31,8 @@ function StockEntryController(
   const mapEntry = initEntryMap();
   const { params } = $state;
 
+  vm.fromShipmentRegistry = !!params.shipment;
```
jniles: Here's a more elegant approach using ui-router:

First, inject the `$transition$` service.

Then:
```js
const isFromShipmentRegistry = $transition$.from().name === 'shipments';
```

This way, we don't have to worry about `params` and what they contain.

I would also encourage you to do this in the `submit` block, not here, since that is all you need it for.  You also don't need to attach it to the `vm` scope.

Relevant documentation: https://ui-router.github.io/ng1/docs/latest/modules/injectables.html#_transition_

client/src/modules/stock/entry/entry.js
```diff
@@ -741,12 +742,21 @@ function StockEntryController(
       user_id       : vm.stockForm.details.user_id,
     };
 
+    if (vm.shipment) {
+      movement.shipment_uuid = vm.shipment.uuid;
```
jniles, line 744: Good catch 👍

server/models/migrations/next/migrate.sql
```diff
@@ -202,8 +202,8 @@ CREATE TABLE IF NOT EXISTS `shipment_tracking` (
   `date`               TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `user_id`            SMALLINT(5) UNSIGNED NULL,
   PRIMARY KEY (`uuid`),
-  CONSTRAINT `shipment_location__shipment` FOREIGN KEY (`shipment_uuid`) REFERENCES `shipment` (`uuid`),
-  CONSTRAINT `shipment_location__user` FOREIGN KEY (`user_id`) REFERENCES `user` (`id`)
+  CONSTRAINT `shipment_tracking__shipment` FOREIGN KEY (`shipment_uuid`) REFERENCES `shipment` (`uuid`) ON DELETE CASCADE,
```
jniles, line 205: Good rename

server/models/schema.sql
```diff
@@ -2750,7 +2750,7 @@ CREATE TABLE `shipment_item` (
   `quantity_sent`      INT(11) UNSIGNED DEFAULT 0,
   `quantity_delivered` INT(11) UNSIGNED DEFAULT 0,
   PRIMARY KEY (`uuid`),
-  CONSTRAINT `shipment_item__shipment` FOREIGN KEY (`shipment_uuid`) REFERENCES `shipment` (`uuid`),
+  CONSTRAINT `shipment_item__shipment` FOREIGN KEY (`shipment_uuid`) REFERENCES `shipment` (`uuid`) ON DELETE CASCADE,
```
jniles, line 2753: 👍

jniles (commented):
This looks awesome!  I've made one single comment on code.

The other thing is that this warning shows up both before the shipment is shipped and after it is received.  We should probably just hide it, since we can add notes whenever.

image


Otherwise, looks good to me.

jmcameron:
> The other thing is that this warning shows up both before the shipment is shipped and after it is received. We should probably just hide it, since we can add notes whenever.

Fixed.

client/src/modules/stock/entry/entry.js
jmcameron: Done.  Although I still needed to check whether a shipment UUID was provided since if we are in the shipments registry and we click on the Stock Exit page, $transition$ thinks we came from 'shipments'.  So I made the check (in Submit now) check both.   That approach seems more robust to future changes (eg, suppose in the future we jump to Stock Exit page with a shipment UUID, but not from the shipment registry, then we would not want to jump back to the shipment registry).

jniles: That makes sense to me.

jniles:
Looks good to me.

Bors r+

PR #6593, opened 2022-05-03
feat(assets): add registry exports

### LINKED ISSUE #6580: Add PDF, Excel, CSV exports to Assets Registry
Add export of PDF, Excel, CSV to Assets Registry.

Make sure that Asset properties (Manufacturer, Model, Reference number, and serial numbers are exported.

Not that this will replace https://github.com/IMA-WorldHealth/bhima/issues/6514.

jniles:
Adds the asset management registry exports to the Assets module.  This
includes exporting to PDF, XLSX, and CSV.

Closes #6580.

image


jmcameron (approved):
The code looks good.
I tested this and it works fine for all three types of exports.   I also use the search to limit the listing in the asset registry and the outputs faithfully matched the registry.

PR #6597, opened 2022-05-05
Refactor asset condition to convert from SQL table to constants (in bhConstants)

### LINKED ISSUE #6596: Convert asset conditiongs from a database table to constants (eg in bhConstants.js)
Currently we are storing asset conditions in a separate table.   Based on our current experience implementing asset management, it seems like we do not need that flexibility.  So it would be good to refactor things and delete the asset_condition table and use constants in the front and back ends.

jmcameron:
Deleted SQL asset_condition table and put the asset condition info in bhConstants.hs in the variable `assetCondition`.
Updated regression tests and migration.sql file.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6596

**TESTING**
- The changes in this PR should not be obvious in normal operation
- Verify correct operation of asset conditions in
  - Asset Management > Asset Scan Registry
  - and editing asset scans in that registry
- The PR should be tested in a migration of a production site database


server/models/migrations/next/migrate.sql
```diff
@@ -303,6 +303,15 @@ INSERT IGNORE INTO `report` (`report_key`, `title_key`) VALUES
 CALL drop_foreign_key('shipment_item', 'shipment_item__condition');
 CALL drop_column_if_exists('shipment_item', 'condition_id');
 
+/**
+ * Convert asset_condition to constants
+ * @author: jmcameron
+ * @date: 2022-05-05
+ */
+
+DROP TABLE asset_scan;
```
jniles: Instead of adding, then dropping the table, let's just remove the add table from this migrate file.  Since the release hasn't been published yet, we don't need to add then drop.

It's found up here:

https://github.com/IMA-WorldHealth/bhima/pull/6597/files#diff-2a7be2572b29530351a6070a44834c3cdae1a97785db7ed3010d615cb36ec358L222

jniles (changes requested):

jniles:
Bors r+

PR #6600, opened 2022-05-09
Add option to asset scan registry to show only the latest scan for each asset

### LINKED ISSUE #6536: Add search filter for last scanned date to the Assets Registry page
Add search filter for last scanned date to the Assets Registry

Probably need to switch loading of assets to new getAssets() function in core.js

jmcameron:
This PR adds a custom search option to the asset scan registry to limit the display to only the latest (last) scan for each asset.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6536

**TESTING**
- Use bhima_test
- Go to:  Asset Management > Asset Scan Registry
- Notice that there is one scan for each of the two items
- Add more scans to those two items (or others as desired).  Be sure to add notes in the scan modal to help you know which one is the latest.
- The registry should show all scans
- In the Search modal select the checkbox option at the bottom "Show only last scan"
- Now notice that only the lastest/last scan of each asset is shown
- Notice the filter item for this.  When you delete it, the display should revert to the normal display


jniles (approved):
Works well.  Good work.

Bors r+

PR #6606, opened 2022-05-11
Reformat reports

### LINKED ISSUE #5320: The header of the "Stock Exit Report" has too much whitespace

jmcameron:
Reformat the whitespace in all reports to make the header more compact.  

This PR is a re-do of [closed PR 5345](https://github.com/IMA-WorldHealth/bhima/pull/5345) but is applied to all reports in BHIMA.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5320

Here is what the updated report looks like.  To see the "before" version, check the issue.
image

**TESTING**
- Generate:  Stock > Reports > Stock Exit Report
- Review the title, etc, on the top right of the resulting report.
- ALL REPORTS were updated, so spot-checking a variety of reports is suggested.


server/controllers/stock/reports/common.js
```diff
@@ -6,7 +6,7 @@ const path = require('path');
 const mkPath = name => path.join(BASE_PATH, name).concat('.handlebars');
 
 // receipts
-const STOCK_EXIT_PATIENT_TEMPLATE = mkPath('stock_exit_patient.receipt.handlebars');
```
jniles, line 9: Nice :+1:

server/controllers/stock/reports/stock/exit_report.js
```diff
@@ -24,7 +24,7 @@ async function stockExitReport(req, res, next) {
   const params = util.convertStringToNumber(req.query);
 
   const optionReport = _.extend(params, {
-    filename : 'REPORT.STOCK.EXIT_REPORT',
```
jniles: Note that this is probably used in the ReportManager in addition here, so you might want to add a new key instead of renaming.

jniles (approved):
Looks pretty good to me.

What do you think about right-aligning the title?

server/controllers/stock/reports/stock/exit_report.js
jmcameron: Good idea.  I'll verify.

jmcameron:
@jniles Thanks for the review.

jmcameron:
@jniles This is what it looks like right-justified:
image

jniles:
Let me know when you need another review here.

server/controllers/stock/reports/stock/exit_report.js
jmcameron: Went back to filename.

client/src/modules/reports/reports.routes.js
```diff
@@ -9,7 +9,7 @@ angular.module('bhima.routes')
       'aged_creditors',
       'aged_debtors',
       'aggregated_stock_consumption',
-      'analysis_auxiliary_cashbox',
+      'analysis_auxiliary_cashboxes',
```
jniles, line 12: This might need an update in the SQL via migrate.sql.  Let me check with a few production versions.

client/src/i18n/en/report.json
```diff
@@ -34,12 +34,15 @@
       "STATUS": {
         "CORRECT": "Correct",
         "TRANSFERT_GREATER": "The amount transferred is greater",
-        "TRANSFERT_LOWWER": "The amount transferred is lower",
+        "TRANSFERT_LOWER": "The amount transferred is lower",
```
jniles, line 37: 👍

client/src/modules/reports/reports.routes.js
jniles: Okay, tested and confirmed we need to add this to the migration script.  Can you do that?

jniles (approved):
I've looked it the majority of reports.  They look excellent.

This was a monumental task, and I'm so glad to have it done.  Thank you!

There is only one comment about the migration.sql script.  The rest are fine.

client/src/modules/reports/reports.routes.js
jmcameron: > Okay, tested and confirmed we need to add this to the migration script. Can you do that?

Done.

jniles (approved):
LGTM.

bors r+

PR #6628, opened 2022-05-23
Improve unpaid invoices report

### LINKED ISSUE #6623: Unpaid Invoices report improvements
In the Finance > Reports > Unpaid invoices report:

If you select a service, it is not displayed in the title block if there are no unpaid invoices for the report -- even though the body of the report will say "no records to display".  If a service is selected, it should be displayed in the title block regardless of whether any unpaid services are found for that service.

To duplicate this issue:
 - Use bhima_test
 - go to Finance > Report > Unpaid invoices report
   - Choose dates:  "This year"
   - Do not select a service, note the item listed
   - Repeat and select the service "Test Service", note that the report title block shows  "Test Service"
   - Repeat and select another service, note that the service is not shown
   - However, the Debtor group is always shown (if selected), which is good

jmcameron:
Improve UI for Unpaid Invoices report

Closes https://github.com/IMA-WorldHealth/bhima/issues/6623

**TESTING**
- Use bhima_test
- Go to:  Finances > Report > Unpaid Invoices Report
  - Choose dates: "This year"
  - Do not select a service, note the item listed
  - Repeat and select the service "Test Service", note that the report title block shows "Test Service"
  - Repeat and select another service.  Even though there are not items found, the service name
    will be shown in the title block.  This makes the report clearer.


server/controllers/finance/reports/unpaid_invoice_payments/index.js
```diff
@@ -45,6 +46,13 @@ async function build(req, res, next) {
     results = { dataset : [], totals : {}, services : [] };
   }
 
+  if (serviceUuid) {
+    // If the user selected a service, force it to be used as "uniqueService"
```
jniles, line 50: Good idea :+1:

jniles (approved):
LGTM.


bors r+

PR #6632, opened 2022-05-25
Fix account balance report

### LINKED ISSUE #6616: Form problems in Balance of Accounts report
There is a logic problem with the report

  Finance > Report > Balance of Accounts

If the "Include Closing Balances" option is set to Yes, you cannot select the period.  But the form can be submitted at this point resulting in a SQL error.

However, if you select No for "Include Closing Balances", select a period, then change the "Include Closing Balances" option to Yes, the period is not cleared but cannot be changed, and the form works...

jmcameron:
Fix the logic for handling the period in the Account Balance report.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6616

**TESTING**
- Use bhima_test
- Go to:  Finance > Reports > Balance of Accounts
  - If "Include Closing Balances" is yes
     - The "Period" selection should be hidden (and not required)
     - Verify that you can generate a report
  - If "Include Closing Balances" is No
    - The "Period" should be shown (and required)
    - Verify that you can generate a report
    - If you change the Fiscal Year, the period should be erased

### DIFF
--- a/client/src/modules/reports/generate/balance_report/balance_report.config.js
+++ b/client/src/modules/reports/generate/balance_report/balance_report.config.js
@@ -21,10 +21,11 @@ function BalanceReportConfigController($sce, Notify, SavedReports, AppCache, rep
 
   vm.onSelectFiscalYear = (fiscalYear) => {
     vm.reportDetails.fiscal_id = fiscalYear.id;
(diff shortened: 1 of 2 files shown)


jniles (approved):
LGTM.

bors r+

PR #6633, opened 2022-05-25
fix(stock): hide expiration date on asset entry

### LINKED ISSUE #6588: Hide expiration date for assets in stock entry
Assets have no expiration dates, so hide/blank out expiration dates for assets in during stock entry and the lots registry.

This was mentioned in [closed issue 6581](https://github.com/IMA-WorldHealth/bhima/issues/6518).

jniles:
This commit ensures that the expiration date is hidden on asset entry.

Closes #6588.

To test:
  1. Create or update an asset in the inventory list.
  2. Attempt a stock entry.
  3. Note that the expiration date column in the "Lots" modal is not visible.
  4. Attempt a stock entry with a regular lot.
  5. Note that the column is now visible.


jmcameron (approved):
I looked over the code changes and they seem fine.
I tested it with Bhima_test and it works as described.

PR #6634, opened 2022-05-25
Updated configurable analysis report

### LINKED ISSUE #6627: Use a fieldset to disable all elements of the "Configurable Analysis Report" if not usable
Certain configurations need to be made before you can use the configurable analysis report.  Here is an example of a report that has not been configured yet:

image

We should disable all these inputs using a fieldset with `disabled` on it.  We should also move the text to the top of the section so it is immediately obvious why the user can't use it.

jmcameron:
Updated the configurable analysis report to disable the form if the configuration has not been initialized.

Also updated the report to show "Provisionary" in the title block if the options for the report use "Include Unposted Values" is YES.   

Note: I was not sure how to run a test without the configuration initialized, but I tested the functionality by overriding the flag in the controller.  It should work correctly.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6627

**TESTING**
- Use any data set
- Go to:  Finances > Report > Configurable Analsysis Report
  - If the configuration has not been initialized, the warning will be shown clearly at the top and the form elements will be disabled and grayed out.
- Try setting the option "Include Unposted Values" to "Yes" and verify that "(Provisionary)" appears in the title block.

### DIFF
--- a/client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.html
+++ b/client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.html
@@ -8,7 +8,7 @@
 <div ng-show="!ReportConfigCtrl.previewGenerated">
   <div class="row">
     <div class="col-md-12">
-      <h3 class="text-capitalize" translate>REPORT.CONFIGURABLE_ANALYSIS_REPORT.TITLE</h3>
+      <h2 class="text-capitalize" translate>REPORT.CONFIGURABLE_ANALYSIS_REPORT.TITLE</h2>
     </div>
   </div>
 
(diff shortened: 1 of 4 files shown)


client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.config.js
```diff
@@ -14,6 +14,7 @@ function ConfigurableAnalysisReportController($sce, Notify, SavedReports, AppCac
   ConfigurationAnalysisTools.read()
     .then((configurationAnalysisTools) => {
       vm.configurationsAreMissing = (configurationAnalysisTools.length === 0);
+      vm.warningStyle = vm.configurationsAreMissing ? 'background-color: #BBB; color: #777;' : '';
```
jniles: I don't think this is necessary.  Can we just use the regular disabled styles?

client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.html
```diff
@@ -21,54 +21,64 @@ <h3 class="text-capitalize" translate>REPORT.CONFIGURABLE_ANALYSIS_REPORT.TITLE<
...
-            <bh-yes-no-radios
-              value="ReportConfigCtrl.reportDetails.includeUnpostedValues"
-              on-change-callback="ReportConfigCtrl.onChangeUnpostedValues(value)"
-              name="includeUnpostedValues"
-              label="REPORT.OPTIONS.INCLUDE_UNPOSTED_RECORDS"
-              help-text="REPORT.OPTIONS.INCLUDE_UNPOSTED_RECORDS_HELP">
-            </bh-yes-no-radios>
+            <fieldset
+              ng-disabled="ReportConfigCtrl.configurationsAreMissing"
+              style="{{ReportConfigCtrl.warningStyle}}">
```
jniles: If you use angularjs templates in styles, you should use `ng-style`.  But either way, I think the regular disabled style is cleaner.  It is correct that this is less obvious about what is going on, but I think the message should be improved, rather than making a bespoke style.

client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.html
```diff
@@ -21,54 +21,64 @@ <h3 class="text-capitalize" translate>REPORT.CONFIGURABLE_ANALYSIS_REPORT.TITLE<
...
-            </div>
-            <div class="checkbox">
-              <label>
-                <input type="checkbox" ng-model="ReportConfigCtrl.reportDetails.hide_details_types" ng-true-value="0" ng-false-value="1">
-                <span translate>REPORT.CONFIGURABLE_ANALYSIS_REPORT.HIDE_DETAILS_TYPES</span>
-              </label>
+            <div ng-if="ReportConfigCtrl.configurationsAreMissing" style="margin-bottom: 1em;">
+              <span class="label label-warning text-capitalize" style="padding: 6px;">
+                <i class="fa fa-exclamation-triangle"></i>
+                <span style="font-size: 1.75em;" translate> FORM.INFO.THE_NECESSARY_CONFIG_MISSING </span>
```
jniles: Let's make this into a proper `alert alert-danger` and change the text a bit.

For example:
image


We should say something like
"Warning"
"The necessary configuration elements are missing.  If you would like to set them now, please follow this link."

And link to the configuration analysis tools page:

image

jniles (commented):
I think this system can be improved, and have proposed an improvement.  Let me know if you think that it doesn't make sense.

client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.html
jmcameron: What about just hiding all form elements if it isn't configured?

jmcameron: > Let's make this into a proper `alert alert-danger` and change the text a bit.

Ok.

> We should say something like "Warning" "The necessary configuration elements are missing. If you would like to set them now, please follow this link."
> 
> And link to the configuration analysis tools page:

Good idea.  Will do.

client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.config.js
jmcameron: Decided to hide all form elements if the configuration is missing.

client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.html
jmcameron: Decided to hide all form elements if the configuration is missing.

client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.html
jmcameron: Done

client/src/modules/reports/generate/configurable_analysis_report/configurable_analysis_report.html
jniles: Sounds good.

jniles (approved):
Yes, this is a much better workflow.  Thanks!

Bors r+

PR #6635, opened 2022-05-25
fix(stock): ensure correct movement filter text 

### LINKED ISSUE #6445: Filter on Stock Movements Search is inverted
The translated text of the label on the "in/out" of the Stock Movements search is inverted.  When I search of all inputs, the label is "output: yes" and then only shows me inputs.

image

jniles:
This commit ensures that the entry/exit movement filter shows the correct values when searching for the movement direction.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6445.

### DIFF
--- a/client/src/modules/stock/movements/modals/search.modal.html
+++ b/client/src/modules/stock/movements/modals/search.modal.html
@@ -22,12 +22,12 @@
             <bh-clear on-clear="$ctrl.clear('is_exit')"></bh-clear>
             <div>
               <label class="radio-inline">
-                <input type="radio" name="movement_direction" id="entry" value="0" ng-model="$ctrl.searchQueries.is_exit">
+                <input type="radio" name="movement_direction" id="entry" ng-value="0" ng-model="$ctrl.searchQueries.is_exit">
                 <span translate>STOCK.INPUT</span>
               </label>
 
               <label class="radio-inline">
-                <input type="radio" name="movement_direction" id="exit" value="1" ng-model="$ctrl.searchQueries.is_exit">
+                <input type="radio" name="movement_direction" id="exit" ng-value="1" ng-model="$ctrl.searchQueries.is_exit">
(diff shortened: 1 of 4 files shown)


client/src/modules/stock/movements/modals/search.modal.html
```diff
@@ -22,12 +22,12 @@
             <bh-clear on-clear="$ctrl.clear('is_exit')"></bh-clear>
             <div>
               <label class="radio-inline">
-                <input type="radio" name="movement_direction" id="entry" value="0" ng-model="$ctrl.searchQueries.is_exit">
+                <input type="radio" name="movement_direction" id="entry" ng-value="0" ng-model="$ctrl.searchQueries.is_exit">
```
jmcameron, line 25: Well... That is tricky!

client/src/modules/stock/movements/templates/io.cell.html
```diff
@@ -1,8 +1,5 @@
 <div class="ui-grid-cell-contents">
-    <span ng-show="row.entity.is_exit === 1" class="label label-danger" translate>
-        STOCK.OUTPUT
+    <span class="label" ng-class="{ 'label-danger' : row.entity.is_exit === 1, 'label-success' :  row.entity.is_exit === 0 }">
```
jmcameron, line 2: Nice

jmcameron (changes requested):
The code looks okay.   I tested it and it works.
But I have one suggestion.   The attached filter setting is a bit convoluted to my eye:
image

I think the filter should show "Input=Yes" or "Output=Yes", not "Output=No" for input-only.

jniles:
I agree.  I was lazy here.

We may need to build tooling for that, since it isn't a use case we've explored before - not sure how long it will take me, but I will give it a whack.

jmcameron (approved):
Will make a new issue for the awkward filter text.

PR #6637, opened 2022-05-26
Disable changing origin depot when editing shipments

### LINKED ISSUE #6571: When editing a shipment, it should not be possible to change the origin depot
Some of the complicated logic when editing shipments is to accommodate the possibility of changing the origin depot while editing the shipment.   We should prevent that possibility by disabling the "Change Depot" option under the [Menu] button while editing shipments.  We can also clean up the related code in create-shipments.js.   This would also be an opportunity to rename create-shipments.js to something more appropriate (for both creating and editing).

jmcameron:
When editing a shipment, this PR prevents the user from changing the origin depot.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6571

**TESTING**
- create a new shipment
  - notice the breadcrumbs say: New Shipment
- in the shipment registry, edit a shipment
  - notice that the breadcrumbs say:  Edit Shipment
  - notice that the menu is no longer available so that there is no way to change the origin depot!
- Double-check the "Edit Shipment" translation in French

### DIFF
--- a/client/src/i18n/en/shipment.json
+++ b/client/src/i18n/en/shipment.json
@@ -19,6 +19,7 @@
     "DOES_SHIPMENT_EXIST": "Do you have a shipment ready?",
     "DOES_SHIPMENT_EXIST_HELP": "Please provide the reference of a shipment that is ready to ship.",
     "EDIT": "Edit",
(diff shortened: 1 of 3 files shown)


jniles (approved):
LGTM.

bors r+

PR #6640, opened 2022-05-27
Fix purchase price analysis report

### LINKED ISSUE #6626: Inventory is not optional in the Price Analysis Report
In the price analysis report, the inventory is marked as optional, but the report breaks if we leave the inventory as optional.

image

jmcameron:
Made Inventory required in Inventory selection for the report.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6626

**TESTING**
- Go to:  Purhcase Orders > Reports > Purchase Price Analysis
- Verify that the report cannot be generated without selecting an inventory item


jniles (approved):
Confirmed this works.

bors r+

PR #6641, opened 2022-05-27
Fix issues in Stock Inventories Report

### LINKED ISSUE #6611: Articles in Stock Report depot filter does not show depot name
When creating the "Stock Inventories" report (articles in stock), if a depot is selected the filter display is broken.   It looks like it is displaying the UUID instead of the depot name. 

If the inventory item selected its filter display has the same problem.

For example:

image

jmcameron:
This fixes a couple of issues with the Stock Inventories report:
- Fixes the filter displays for depot and inventory filters
- Small fix to the grid display to show "No records" if none are found.   The problem here was that the emptyTables partial was being invoked in the loop for each depot.   But due to to the way the data structured, if a depot has no articles in stock, it will not appear in the table.  So moved it out of the depot loop.
- Minor formatting improvements
- Also removed an unrelated diagnostic comment left over from a previous PR.

NOTE: The fix for the depot and inventory filter was a change to the `formatFilters()` function in `server/controllers/finance/reports/shared.js`.  This should address the same filter display in other reports that used the function.   BTW I was not sure if I needed to do anything extra for the depot/inventory db queries introduced in the `formatFilters()` function (eg, escaping, catch errors, etc) so I welcome suggestions for improving this code.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6611

**TESTING**
- Go to:  Stock > Reports > Stock Inventories
  - Try the report with and without selecting a depot and/or inventory and check the outputs of the report
  - Select an inventory that has no articles and verify that the table display in the report includes "There are no records to display." message

### DIFF
--- a/server/controllers/finance/reports/vouchers/receipt.handlebars
+++ b/server/controllers/finance/reports/vouchers/receipt.handlebars
@@ -10,7 +10,6 @@
       </div>
     {{/if}}
   {{/header}}
-  QQQ1
 
   <header>
     <!-- headings  -->
--- a/server/controllers/stock/reports/stock/inventories_report.js
+++ b/server/controllers/stock/reports/stock/inventories_report.js
@@ -1,5 +1,5 @@
 const {
-  _, ReportManager, formatFilters, Stock, STOCK_INVENTORIES_REPORT_TEMPLATE, stockStatusLabelKeys,
+  _, db, ReportManager, formatFilters, Stock, STOCK_INVENTORIES_REPORT_TEMPLATE, stockStatusLabelKeys,
 } = require('../common');
 
 const i18n = require('../../../../lib/helpers/translate');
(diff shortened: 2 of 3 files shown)


server/controllers/finance/reports/vouchers/receipt.handlebars
```diff
@@ -10,7 +10,6 @@
       </div>
     {{/if}}
   {{/header}}
-  QQQ1
```
jniles, line 13: Oh la la.  Who do you suppose uses vim around here?  😆

server/controllers/finance/reports/shared.js
```diff
@@ -221,6 +224,16 @@ function formatFilters(qs) {
       if (filter.isPeriod) {
         const service = new PeriodService(new Date());
         filter.value = service.periods[value].translateKey;
+      } else if (filter.isDepot) {
+        db.one('SELECT text FROM depot WHERE uuid = ?', db.bid(value))
```
jniles: This function isn't asynchronous.  This might work, sometimes, but it won't work all the time.

Typically what we do is send back the depotUuid and inventoryUuid as "displayValues" mapping.   This isn't well developed so it may not even work.

I think that to make this work properly we either need to:

1. Send the proper display value from the client (this is the half-hearted implementation we currently have)
2. Make a standard filter definition for anything that could be a uuid -> text label.  So, for example, we could have "lookupHumanReadableLabelByUuid()" which scans the document_map, the entity_map, the inventory, etc, to figure out what the label should be and then displays that.  This would be a server-side function and used in things like reporting.  We could then just change all these to `isUuid`, make this filter options `async`, and then have a standard way of looking up the labels by their uuid.

What are you thoughts?

jniles (changes requested):
Comments made inline.

server/controllers/finance/reports/vouchers/receipt.handlebars
jmcameron: No, this one was mine!  DEFINITELY unrelated to vim!

server/controllers/finance/reports/shared.js
jmcameron: My thoughts:
- Although db.one() is asynchronous, I knew the `formatFilters()` was not.  Perhaps it should be!  I notice that the server-side report JS file does awaits on other things.  So it could easily do an awit for `formatFilters()`.
- I don't like sending the display text with things that have uuids.   I think we should be looking the labels up on the server side.
- Your second option seems like overkill to me...

jniles (approved):
Looks good to me.

bors r+

PR #6643, opened 2022-05-27
fix(payroll): add loading indicators to selects

### LINKED ISSUE #6385: Hide error state on Pay Period Configuration modal until data is loaded in
The error state on the Pay Period Configuration modal is incorrectly displayed on a slow internet connection.  The error state tries to inform the user that no configurations exist, even though that is not true - they are simply loading in.  We should:

1) Display a loading indicator on the side of the select boxes when data is loading
2) Show the error state only when data has finished loading and we confirm there are no configuration items.

Example:
firefox_YqeuK7RjoE

jniles:
Adds loading indicators to a bunch of selects so they do not currently distinguish between the "data has not loaded yet" and "has no data" state, causing confusing error messages on first loading.

Closes #6385

Best way to reproduce this is with Vanga's database.
 1. Use vanga's database
 2. Go to `#!/payroll`
 3. Click on the "Set up a payment period option"
 4. You will no longer see the red shown in the issue.

You can confirm by slowing your network speed down.

### DIFF
--- a/client/src/js/components/bhAccountConfigSelect.js
+++ b/client/src/js/components/bhAccountConfigSelect.js
@@ -30,11 +30,16 @@ function AccountConfigSelectController(AccountConfig, Notify) {
       $ctrl.required = true;
     }
 
+    $ctrl.isLoading = true;
+
     AccountConfig.read()
       .then(accountConfigs => {
         $ctrl.accountConfigs = accountConfigs;
       })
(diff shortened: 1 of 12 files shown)


jmcameron (approved):
I looked through the code and it looks good.
I tested this with the Vanga dataset and it seems to work.  I had no issues with errors due to delayed loading for selection components.

PR #6650, opened 2022-05-31
Fix translation term error

jmcameron:
Fixes untranslated item in the title to Stock Exit to Patient receipt documents.


jniles (approved):
LGTM.

bors r+

PR #6653, opened 2022-05-31
chore(reports): remove unused P&L Statements

### LINKED ISSUE #6629: Remove profit and loss reports
We only need Operating Account report.  The profit and loss reports and redundant and are currently broken, so we should delete them.

jniles:
This commit removes the unused P&L Statements.

Closes #6629.


jmcameron (approved):
I reviewed the code and migration files and they all look good.
I tested it with bhima_test DB and it worked fine.
Did not try a migration (but the migration file looks correct).

PR #6654, opened 2022-05-31
fix(stock): preserve patient name in link

### LINKED ISSUE #5442: Link from stock movement -> patient doesn't preserve patient name
If you link from a stock movement to the patient, the filter shows the patient's uuid.  See below:
image

A better way would be to show the patient's reference.  I'm not sure if we can do something about this in the current construction of the movements registry, but I'm reporting it here to track it since it is not ideal.

jniles:
This commit preserves the patient name in the stock movement registry.

Closes #5442.

### DIFF
--- a/client/src/modules/stock/movements/templates/action.cell.html
+++ b/client/src/modules/stock/movements/templates/action.cell.html
@@ -13,7 +13,7 @@
     </li>
 
     <li ng-if="row.entity.flux_id === grid.appScope.flux.TO_PATIENT">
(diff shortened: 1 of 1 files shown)


jmcameron (approved):
The code changes look good.
Tested it on bhima_test and it works correctly.  When "View Patient" is select for an exit to patient (in the Stock Movements registry), the resuling patient view shows the ID of the patient, not the UUID.

PR #6655, opened 2022-05-31
fix(stock): improve Stock Sheet report

### LINKED ISSUE #5415: Improvements to the "Fiche de Stock" report
In reviewing https://github.com/IMA-WorldHealth/bhima/pull/5341, here are some bugs/observations.

### Bug: Date range should provide an opening balance
If you choose to set a date range for the Fiche de Stock, you limit the output of the report to a certain time frame.  However, we do not set opening balances for the amount of stock at the `start_date` nor do we compute the value of the stock.  This is wrong and misleading.

To get the value of the stock, we will need to implement the weighted average cost (WAC) algorithm for a given date.  See discussion in https://github.com/IMA-WorldHealth/bhima/issues/5401.

To get the quantity in stock at that point, we can simply look it up from the `stock_movement_status` table.  It will be the `sum_quantity` of the most recent date directly before the `start_date`.

The opening balance should be the first line in this report.

_This bug should be fixed before merging https://github.com/IMA-WorldHealth/bhima/pull/5341_.  

### Proposal: Color negative stock values in red
It should be very obvious when stock dips into the negative numbers.  Let's make it increasingly obvious by setting `text-danger` on the entire row when the stock goes into the negative.

### Proposal: Include even more information in the "reason" column
Here is an example of the fiche de stock from one of our production sites.

image

We added the "reason" column in https://github.com/IMA-WorldHealth/bhima/commit/e3891a940fe5e69b0e425d8ae295e214784e8419 to show the flux of stock.  However, this leaves me wanting even more information.  It would be nice to put:
  1. The patient who interacted with the stock (received or returned)
  2. The depot (received or returned)
  3. The service (received or returned)

This will allow us to more easily compare patients invoiced for medication X with those who received medication X on day Y.

### LINKED ISSUE #6652: Fiche de stock has unit cost errors in summary table
image

The unit cost does not agree between the summary of the Fiche de Stock and the final line.

I reran the `RecomputeStockValue(NULL)` algorithm to recompute the weighted average cost, and the WAC still comes back as $24, though this has never been used as a purchase cost.

jniles:
This commit fixes the rendering of the unit cost calculation on the
stock sheet report.  It also adds in the purchase order, patient
reference, or other references on the line as needed.

Closes #6652.
Closes #5415.

image

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -277,7 +277,7 @@
       "STOCK_OUT_NOT_USABLE" : "Stock Out (Not usable)",
       "UNUSED_STOCK"  : "Unused stock"
     },
-    "STOCK_VALUE_WITH_WAC" : "End value of the stock with the unit cost of the company",
+    "STOCK_VALUE_WITH_WAC" : "End value of stock with the final unit cost",
     "STOCK_MOVEMENT"  : "View Stock Movement",
     "STOCK_MOVEMENTS" : "View Stock Movements",
     "SUCCESS"         : "Successfully created a movement",
--- a/server/controllers/inventory/index.js
+++ b/server/controllers/inventory/index.js
@@ -116,13 +116,15 @@ async function getInventoryWac(req, res, next) {
 
 async function computeWac(inventoryUuid) {
   const binaryInventoryUuid = db.bid(inventoryUuid);
+
   const queryRecompute = 'CALL RecomputeInventoryStockValue(?, ?);';
+
   const querySelect = `
     SELECT
       BUID(sv.inventory_uuid) inventory_uuid,
       i.text, sv.date, sv.quantity, sv.wac
     FROM stock_value sv
-    JOIN inventory i ON i.uuid = sv.inventory_uuid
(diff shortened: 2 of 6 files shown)


jmcameron (commented):
I ran into a problem when testing this PR.  I got an error when I used this configuration with bhima_test:
image

Here is the server error:
```
  errors #interceptor() TypeError: Cannot read properties of undefined (reading 'value')
  errors     at stockSheetReport (/home/jmcameron/Bhima/bin/server/controllers/stock/reports/stock/stock_sheet.js:98:23)
  errors     at processTicksAndRejections (node:internal/process/task_queues:96:5) +0ms
```

jniles:
Thanks for the review @jmcameron !  I think this should take care of it.

jmcameron (approved):
I looked over the code and it looks fine.
Tested it with bhima_test and it worked fine.
One suggestion:  The wording on the last row of the table:
   "End value of the stock with the unit cost of the company"
is a little strange to me eyes (what does "of the company" mean?). Suggest:
   "End value of the stock with the final unit cost"
or something similar.

jniles:
Made the suggested change.

bors r+

PR #6656, opened 2022-05-31
fix(payroll); improve multipayroll grid columns

### LINKED ISSUE #6387: Improvements to Multiple Payroll (index-based)
There are a couple of improvements to make to "Multiple Payroll (index-based)".  

image

We shouldn't display this column above to hide.

image

We should add the ability to filter on any column.

jniles:
This commit adds the ability to filter on any of the rubrics in the multiple payroll by index grid.  It also removes the row selection
checkbox.

Closes #6387.

### DIFF
--- a/client/src/modules/multiple_payroll_indice/multiple_payroll_indice.js
+++ b/client/src/modules/multiple_payroll_indice/multiple_payroll_indice.js
@@ -99,27 +99,25 @@ function MultiplePayrollIndiceController(
 
     MultiplePayroll.read(null, filters)
       .then((result) => {
-        renameGidHeaders(result.rubrics);
+        renameGridHeaders(result.rubrics);
         vm.gridOptions.data = setGridData(result.employees);
       })
       .catch(errorHandler)
       .finally(toggleLoadingIndicator);
   }
 
-  function renameGidHeaders(rubrics) {
+  function renameGridHeaders(rubrics) {
     const actions = angular.copy(columnDefs[columnDefs.length - 1]);
     const newColumns = columnDefs.slice(0, 1);
 
     const header = {
       type : 'number',
-      enableFiltering  : false,
       headerCellFilter : 'translate',
       cellClass  : 'text-right',
       footerCellClass  : 'text-right',
       footerCellFilter : 'number:2',
       cellFilter : 'number:2',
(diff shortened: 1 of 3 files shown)


client/src/modules/multiple_payroll_indice/multiple_payroll_indice.js
```diff
@@ -99,27 +99,25 @@ function MultiplePayrollIndiceController(
...
-        renameGidHeaders(result.rubrics);
+        renameGridHeaders(result.rubrics);
         vm.gridOptions.data = setGridData(result.employees);
       })
       .catch(errorHandler)
       .finally(toggleLoadingIndicator);
   }
 
-  function renameGidHeaders(rubrics) {
+  function renameGridHeaders(rubrics) {
```
jmcameron, line 109: 👍

jmcameron:
I could not figure out how to get to the page shown at the  top of : https://github.com/IMA-WorldHealth/bhima/issues/6387#issue-1125674841

jniles:
Ah, go into the menu and click "Columns"

image

jniles:
It would be must useful to use Vanga's database as there is actual data in there.

jmcameron:
> Ah, go into the menu and click "Columns"

I do not see how to add these extra columns (eg, PS, CSC), and (probably why), my page title reads "Payroll / Multiple Payroll" NOT "Payroll / Multiple Payroll (index-based)".  Is there a setting for this?   I already enabled the "Enable index payment system"...

jmcameron (approved):
Code looks good.

After figuring where how to bring this page up, I was able to test it with bhima_test and a production data set.  It seems to work correctly.  I found one oddity with the production data set when using the filtering on TX.P:  All entries showed 1,355.30.  If I typed 1, 3, 5, or 0, all entries stayed.  If I type in 4 all entries were hidden.  But if I type in 9, all entries remained.  I suppose the actual text for the number must have more digits than are shown.

One other minor cosmetic thing (which can be ignored, fixed, or pushed to later): With so many columns, all we are seeing for the title of each column is one or two letters.  I suggest adding the grid column setting for 'headerTooltip' (in `function renameGridHeaders(rubrics)`) to be the same as the label name so that users can just hover over a column label to see what it is.

jniles:
Changes landed.  Great idea!

Bors r+

PR #6661, opened 2022-06-01
Updated balance of accounts report

### LINKED ISSUE #6171: Updates for the Balance of Accounts report
Finance > Report > Balance of Accounts

**Suggestions**
- It seems to me that the "Include Closing Balances" option should have a default.
- What does the last row, marked "Total", represent?  If it is the closing balance, it should probably say that or at least have a label that is more descriptive than "Total"
- Add multi-currency support

jmcameron:
Updates the balance of accounts report
- Makes  "Include Closing Balances" default to NO
- Adds multi-currency report

Closes https://github.com/IMA-WorldHealth/bhima/issues/6171

**TESTING**
- Finances > Reports > Balance of Accounts
- Suggest testing with production data so there is adequate data to check
- First, do a Balance of Accounts with master (defaults to enterprise currency).  save the data for comparison
- Switch to the pull request
- Repeat the test (with the default, which should be the enterprise currency) and compare to the results saved from the master
- Try the report with an alternate currency and spot check the calculations
- Try both the Single Period report (Include Closing Balances = NO) and the FY report (Include Closing Balances = YES)

### DIFF
--- a/client/src/modules/reports/generate/balance_report/balance_report.config.js
+++ b/client/src/modules/reports/generate/balance_report/balance_report.config.js
@@ -2,7 +2,7 @@ angular.module('bhima.controllers')
   .controller('balance_reportController', BalanceReportConfigController);
 
 BalanceReportConfigController.$inject = [
-  '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state',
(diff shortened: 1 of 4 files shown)


jniles (approved):
Tested and this works beautifully.

Great job!

bors r+

PR #6664, opened 2022-06-02
Reimplemented Current Assignments report

### LINKED ISSUE #6621: Report issues
Once [PR 6606](https://github.com/IMA-WorldHealth/bhima/pull/6606) lands, there are few a reports that should be checked.

These reports may be obsolete (or I could not figure out how to see these reports in BHIMA)
   - [x] finance/reports/break_even/report.handlebars
   - [x] finance/reports/break_even/report_project.handlebars
   - [x] finance/reports/break_even_cost_center/report.handlebars
   - [x] finance/reports/break_even_cost_center/report_project.handlebars
   - [x] finance/reports/cost_center/report.handlebars
   - [x] inventory/reports/items.handlebars
   - [x] stock/reports/stock/assignment/stock_assign.registry.handlebars
   - [x] stock/reports/stock_assign.receipt.handlebars

I could not figure out how to check these reports.    I tried to fix the formatting in them, but it is possible further tweaks may be needed:
   - [x] finance/reports/cash/receipt.pos.handlebars
   - [x] finance/reports/vouchers/report.pos.handlebars
   - [x] stock/reports/stock_adjustment.receipt.handlebars
   - [x] stock/reports/stock_aggregate_consumption.receipt.handlebars
   - [x] stock/reports/stock_movements.report.handlebars

jmcameron:
Reconnected old Assignments registry report.
- Does not implement a CSV version.  If that is desired, we can add it later.

Although "Assignments" can be applied to all lots, assigning/etc, is only supported via the Assets Management pages.  Hence this report is under that menu.   This is due to earlier decisions regarding handling assets differently than non-asset lots (although it could be changed).  We currently have no hints of assignment-related functionality in the Stock Lots registry.

Although this report can currently be accessed under the Assets Registry Menu, it could be moved under Assets Management > Reports.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6621

**TESTING**
- Use bhima_test
- Go to:  Asset Management > Assets Registry
   - This should show two assets MOT1, and MOT2.  Only MOT1 is assigned
   - Use the Menu > Current Assignments and examine the resulting report
      - It should show only the the assignment for MOT1
   - Using the action menu for MOT2, assign it to an entity
   - Repeat generating the report and verify that both assets are shown
   - Test adding an assignment to a non-asset (eg, drug) using the "+ New Assignment" above the Assets Registry.  
     Even though it will not be shown in the asset registry, it will be shown in the Current Assignments report.


jniles (approved):
LGTM.

bors r+

PR #6673, opened 2022-06-10
Close requistions when stock exits complete them

### LINKED ISSUE #6424: Process requisition flags on server, not in StockExitController
Currently, we process if a requisition is "completed" or "partially completed" in the stock exit clientside module.  Here is an example code reference:

https://github.com/IMA-WorldHealth/bhima/blob/944af46489c171a0fd5a8de1a29836301fde01ec/client/src/modules/stock/exit/exit.js#L720-L732

This is problematic because if the network hitches between the stock movement and the requisition update request, the requisition will never be updated.  A better way to do this would be for the server to update the requisition during the stock movement creation, that way only one HTTP request is sent and the server figure out how to process it as a single `db.transaction()`.

I propose that we re-write the requisition processing to mark requisitions as finalized on the server.

jmcameron:
Updated the Stock Exit handling so that if a stock exits completes a requisition, it will be marked as complete.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6424

**Notes**
- This will only mark a requisition as complete if all items in the requisition are handled by the stock exit.  If any items are not handled, this change will leave it marked as "Partial".
- Since it is possible for a requisition to be completed even if all items are not handled (eg, when items are lost in transit), we probably need a way to mark such requisitions as completed.   A separate issue should be created to address this issue.

**TESTING**
- Use bhima_test
- Go to:  Stock > Stock Exit
  - Make sure the source depot is: Depot Principal
  - Select from [Depot] and choose the destination depot: Depot Secondaire
  - Answer "Yes" to "Do you have a requisition slip?"
  - Select the requisition
  - Click [Submit]
  - In the main Stock Exit form, notice that some items in the order (Prednisolone) are not available
  - Click [Submit] to perform as much of the requisition as possible.
- Go to: Stock > Requisition
  - Notice that the requistion SREQ.TPA.2 is still marked Partial.
  - Edit the requistion
     - Change the quantity for the quinine to 5 (greater than 4)
     - Delete the Prednisolone
     - Click [Submit]
- Go back to: Stock > Exit
  - Repeat the process above (exiting the extra quinine to fulfill the requisition)
- Go back to:  Stock > Requisition
  - Notice the requisition is now marked Complete

### DIFF
--- a/client/src/js/components/bhRequisitionSelect.js
+++ b/client/src/js/components/bhRequisitionSelect.js
@@ -7,6 +7,7 @@ angular.module('bhima.components')
       requisitionUuid : '<',
       requestorUuid : '<?',
       onSelectCallback : '&',
+      disallowCompleted : '@?',
       required : '@?',
       label : '@?',
       disabled : '<?',
@@ -14,26 +15,31 @@ angular.module('bhima.components')
   });
 
 RequisitionSelectController.$inject = [
-  'StockService', 'NotifyService',
+  'StockService', 'NotifyService', 'bhConstants',
 ];
 
 /**
  * Requisition Select Controller
  *
  */
-function RequisitionSelectController(Stock, Notify) {
+function RequisitionSelectController(Stock, Notify, bhConstants) {
   const $ctrl = this;
 
   $ctrl.$onInit = function onInit() {
     $ctrl.label = $ctrl.label || 'FORM.LABELS.REQUISITION_REFERENCE';
     $ctrl.required = $ctrl.required || false;
+    $ctrl.disallowCompleted = $ctrl.disallowCompleted || false;
 
     const params = {};
     params.requestor_uuid = $ctrl.requestorUuid;
 
+    $ctrl.disallowCompleted = false;
+
     Stock.stockRequisition.read(null, params)
       .then((requisitions) => {
-        $ctrl.requisitions = requisitions;
+        const COMPLETED = bhConstants.stockRequisition.completed_status;
+        $ctrl.requisitions = $ctrl.disallowCompleted
+          ? requisitions.filter(req => req.status_id !== COMPLETED) : requisitions;
       })
       .catch(Notify.handleError);
   };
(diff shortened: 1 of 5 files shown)


client/src/js/components/bhRequisitionSelect.js
```diff
@@ -31,9 +32,14 @@ function RequisitionSelectController(Stock, Notify) {
     const params = {};
     params.requestor_uuid = $ctrl.requestorUuid;
 
+    $ctrl.allowCompleted = false;
+
     Stock.stockRequisition.read(null, params)
       .then((requisitions) => {
-        $ctrl.requisitions = requisitions;
+        const allowCompleted = $ctrl.allowCompleted !== 'false';
```
jniles: Why is `false` a boolean in line 35 and a string here?

Can we coerce one into the other?

server/controllers/stock/requisition/requisition.js
```diff
@@ -316,5 +320,21 @@ exports.deleteRequisition = async (req, res, next) => {
...
+/**
+ * @function updateStatus
+ * @description updates the status of a requisition based
+ * @param {string} uuid - The uuid of the requisition
+ */
+async function updateStatus(uuid) {
+  // First get the requisition details
+  const balance = await getDetailsBalance(uuid);
+  if (balance.items.length === 0) {
+    return db.exec('UPDATE stock_requisition SET status_id = ? WHERE uuid = ?', [
```
jniles: Since we are in an `async` context, we can avoid some of this complexity with return statements.  I propose you just do:

```js
const balance = await getDetailsBalance(uuid);
  if (balance.items.length === 0) {
    await db.exec('UPDATE stock_requisition SET status_id = ? WHERE uuid = ?', [
  REQUISITION_STATUS_COMPLETE, db.bid(uuid)]);
  }
````

jniles (commented):
This looks good to me!  Just some formatting questions on the code that I found confusing, but the logic looks correct to me.

client/src/js/components/bhRequisitionSelect.js
jmcameron: Changed the logic on the flag to clean this up.

server/controllers/stock/requisition/requisition.js
jmcameron: Done.  Thanks for the reminder!

jmcameron:
This update fixed a couple of issues I ran into when testing the updates.

jniles (approved):
LGTM.

Thanks so much!

bors r+

PR #6677, opened 2022-06-13
Enable marking a partial requisition as completed

### LINKED ISSUE #6674: Add ability to mark a partially filled requisition as complete
The current code (after [PR 6673](https://github.com/IMA-WorldHealth/bhima/pull/6673) is merged) will mark a requisition complete if it is 100% fulfilled.   However, there are some cases where a requisition cannot be completely fulfilled (eg, due to losses in transit), it would still be useful to be able to mark the requisition as completed.    

We should add a way to edit the status of a requisition as complete even when it is in a partial state.  This could be done as a separate action menu item or via the existing interface to edit the status of requisitions.   In either case, a we should show a pop-up confirmation to prompt the user to be sure they know what they are doing.

jmcameron:
This PR is a follow up of [PR 6673](https://github.com/IMA-WorldHealth/bhima/pull/6673) to allow requisitions that are partially completed to be marked as complete.  

Closes https://github.com/IMA-WorldHealth/bhima/issues/6674

**TESTING**
- Use bhima_test
- Go to the requisition registry:  Stock > Requisition
  - Use the action menu for a requisition that is in "In Progress" and edit its status.  Note the new option to mark the requisition as complete.   
  - When you select the "Mark as complete" option and click [Submit], a confirmation dialog will be displayed
  - You should be able to click [Cancel] on the confirmation dialog and go back to the requisition registry
  - If you repeat the process but click [Confirm] on the confirmation dialog, the status of the requisition should change to "Completed".

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -107,10 +107,11 @@
             "CONFIRM_ACTION": "Do you confirm this action?",
             "CONFIRMATION_NEEDED": "Confirmation Needed",
             "CONFIRM_ACTIVATION": "Do you really want to allow this user's system access",
+            "CONFIRM_MARKING_REQUISITION_COMPLETE": "Do you really want to mark this requisition as complete?  It is only partially fulfilled!",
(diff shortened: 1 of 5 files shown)


jniles (approved):
LGTM.

bors r+

PR #6679, opened 2022-06-14
Feat(remote access) using Json Web Token to handle the remote access

### LINKED ISSUE #6675: Allow remote access to bhima
As bhima will be used a lot for the next project, I think allowing other servers or applications to communicate with it will be very important. The remote access will make it easy to build other applications on top of bhima (having bhima as server for instance as DHIS2 is).

We currently need a session to access bhima APIs (private ones). It will be possible to use a <a href="https://jwt.io">token</a> for creating a session in bhima, so all APIs will be called by any application(mobile, nodes, php, etc...).

The token will not replace the session, but will be used only for remote access, it means that the current behavior will not be affected.

jeremielodi:
closes #6675

This snippet below describes how you can call bhima private APIs remotely:
```js

const axios = require('axios').default;
const server = 'http://localhost:8080';

axios.defaults.xsrfCookieName = 'csrftoken';
axios.defaults.xsrfHeaderName = 'X-CSRFToken';

async function getToken() {
  const endPoint = `${server}/auth/login`;
  const response = await axios.post(endPoint, {
    username: 'superuser',
    password: 'superuser',
    project: 1
  }, {
    responseType: 'json',
  });
  return response.data.token;
}

async function getDepots() {
  const endPoint = `${server}/depots`;
  const token = await getToken();
  const response = await axios.
    get(endPoint, {
      headers: {
        'x-access-token' : token,
      }
    });
  const depots = response.data;
  console.log(depots)
}

getDepots();

```

You can try the same thing using python

```python

import requests
import json


server = 'http://localhost:8080'
authAPI = server + '/auth/login'

# headers
headers = {
  'response-type': 'json',
}
# bhima requires the username, password and the project id for authentification
authData = {
  'username': 'superuser',
  'password': 'superuser',
  'project': 1
}
# let get the token for remote access
authResponse = requests.request(
  "POST", authAPI, headers=headers, json=authData
)

session = json.loads(authResponse.content)
token = session['token']

# accessing private route

depotAPI = server + '/depots'

# headers
headers = {
  'x-access-token': token,
  'response-type': 'json',
}

depotResponse = requests.request("GET", depotAPI, headers=headers)
depots = json.loads(depotResponse.content)
# let print the result
print(depots)

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
     "jquery": "^3.5.1",
     "jsbarcode": "^3.11.5",
     "json-2-csv": "^3.17.1",
+    "jsonwebtoken": "^8.5.1",
     "lodash": "^4.17.21",
     "moment": "^2.29.3",
     "morgan": "^1.9.1",
--- a/server/config/access.js
+++ b/server/config/access.js
@@ -14,6 +14,13 @@ const publicRoutes = [
   '/currencies',
 ];
 
+const assetRoutes = [
+  '/index.html',
+  '/css/',
+  '/js/',
+  '/assets/',
+];
+
 module.exports = (app) => {
   // eslint-disable-next-line consistent-return
   app.use((req, res, next) => {
@@ -21,14 +28,12 @@ module.exports = (app) => {
     let sessionExists = true;
 
     if (token) {
-      JWTConfig.verify(req, (decoded) => {
-        if (!decoded) {
+      JWTConfig.verify(req, (session) => {
+        if (!session) {
           sessionExists = false;
           return;
         }
-        req.session.enterprise = decoded.enterprise;
-        req.session.project = decoded.project;
-        req.session.user = decoded.user;
+        _.merge(req.session, session);
       }, (error) => {
         sessionExists = false;
         debug(error);
@@ -36,14 +41,14 @@ module.exports = (app) => {
     }
 
     if (sessionExists) {
-      if (_.isUndefined(req.session.user) && !within(req.path, publicRoutes)) {
+      if (_.isUndefined(req.session.user) && !within(req.path, publicRoutes) && !withinAsset(req.path, assetRoutes)) {
         debug(`Rejecting unauthorized access to ${req.path} from ${req.ip}`);
         next(new Unauthorized('You are not logged into the system.'));
       } else {
         next();
       }
     } else {
-      debug('token not valide');
+      debug('token not valid');
       res.status(401).json({
         msg : 'token not valid',
       });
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -124,6 +124,7 @@
     "jquery": "^3.5.1",
     "jsbarcode": "^3.11.5",
     "json-2-csv": "^3.17.1",
+    "jsonwebtoken": "^8.5.1",
```
jmcameron, line 127: I think you need to add axios too.

server/config/access.js
```diff
@@ -0,0 +1,64 @@
...
+
+    if (sessionExists) {
+      if (_.isUndefined(req.session.user) && !within(req.path, publicRoutes) && !withinAsset(req.path, assetRoutes)) {
+        debug(`Rejecting unauthorized access to ${req.path} from ${req.ip}`);
+        next(new Unauthorized('You are not logged into the system.'));
+      } else {
+        next();
+      }
+    } else {
+      debug('token not valide');
```
jmcameron: Spelling error (English):  valide -> valid

server/config/access.js
```diff
@@ -0,0 +1,64 @@
+const _ = require('lodash');
```
jmcameron, line 1: I like the reorganization of these functions into the access.js file.

test/integration/remoteAccess.js
```diff
@@ -0,0 +1,64 @@
+/* global expect, chai, */
```
jmcameron, line 1: Thanks for adding a test!

jmcameron (commented):
I looked over the code and it looks good.  
Thanks for adding a test!  Excellent!

I tested this and it works fine.  But axios needs to be added to the package list.

The test script should also log off!  And any applications that use this interface should log off when they complete their task.

One concern I had was that for our production sites we connect with http, not https, so the user credentials will be transmitted in the clear.  Since our regular web interface login has the same problem, this is not any less secure.  However, this is an issue that we may want to consider addressing in the future:  setting up and require https even in local LAN installations.

One comment:  I noticed that the token created this way is good for 24 hours.  That is probably the default for web interfaces.  However for this application, I wonder if a smaller period could be used.  If a program is accessing a Bhima site, wouldn't it do its operations fairly quickly and then log off?  In which case a shorter period could be used.

Finally, although I tested this and it seems to work well, I'm not familiar enough with the login authentication security implications and whether this might create future security problems.  So hopefully someone else will be able to review it and provide additional insight into the security issues.

package.json
jeremielodi: I used ```axios``` just for snippet I provided in comment. Any node request module can be used for testing the remote access.

server/config/access.js
jeremielodi: You're right

mbayopanda:
It works like a charm 👍🏽

mbayopanda (approved):
LGTM 👍🏽

PR #6697, opened 2022-06-20
Fix bug in editing asset serial number during stock entry

### LINKED ISSUE #6694: Serial Number not saved when editing asset on Stock Entry
When editing an asset on stock entry, if you add a serial number to the modal, then try to edit the stock entry again, the serial number is not restored.  This is a bug.

The expected behavior is that the serial number should be restored.

jmcameron:
Fixes bug when trying to update a serial number of an asset during stock entry.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6694

**TESTING**
- Use bhima_test
- Go to Stock > Stock Entry
   - Select Integration
   - Add a motorcycle (use inventory code MOT...)
   - Click on the "Lots" popup
     - Add a new label (eg, "MOT3") 
     - Add a value for the serial number
     - Click on [Submit]
   - Click on the "Lots" popup again.  The previously entered serial number should be there

### DIFF
--- a/client/src/modules/stock/entry/modals/StockEntryModalForm.service.js
+++ b/client/src/modules/stock/entry/modals/StockEntryModalForm.service.js
@@ -24,6 +24,7 @@ function StockEntryModalForm(uuid) {
     this.unit_cost = row.unit_cost || null;
     this.quantity = row.quantity || 1;
     this.lot = row.lot || null;
+    this.serial_number = row.serial_number || null;
     this.isInvalid = true;
     this.isValid = false;
     this.identifier = uuid();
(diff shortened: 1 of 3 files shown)


client/src/modules/stock/entry/modals/StockEntryModalForm.service.js
```diff
@@ -24,6 +24,7 @@ function StockEntryModalForm(uuid) {
     this.unit_cost = row.unit_cost || null;
     this.quantity = row.quantity || 1;
     this.lot = row.lot || null;
+    this.serial_number = row.serial_number || null;
```
mbayopanda, line 27: 👍🏽

mbayopanda (approved):
LGTM

jniles:
@mbayopanda has approved!

bors r+

PR #6699, opened 2022-06-21
Rename actif to immobilisation for Assets

mbayopanda:
This PR rename "Actif" to "Immobilisation" for Assets Management in French translation


mbayopanda (commented):
bors r+

PR #6704, opened 2022-06-24
Shipments updates

### LINKED ISSUE #6470: Add more shipment status states
Currently the shipment has a state for when it is ready to ship.  However, once the shipment is shipped, we have no way to change its status until you actually start unpacking it at the destination.  I would like to see another state.  In fact, I would like to see the following states
- EMPTY (new)
- PACKED ("COMPLETE", but still possible to change contents -- container not sealed).  This is current "AT_DEPOT").
- READY_FOR_SHIPMENT (packed and ready to go, and the box is sealed so you cannot change the contents)
- IN_TRANSIT (has left the source Depot but has not been received by the destination depot)
- AT_DESTINATION_DEPOT (received at the destination depot, but not unpacked).
- DELIVERED (completely unpacked regardless of whether any items were lost/stolen)
- LOST (if we are certain it will never arrive at the destination depot; implies that some followup is needed to do appropriate stock exits)

Some of the logic is already implemented in PR https://github.com/IMA-WorldHealth/bhima/pull/6394.   However I think the receiving depot needs to be able to mark that it has received it (AT_DESTINATION_DEPOT) before it unpacks it -- since it may take a while to unpack.     I would add another flag-setting option in the shipment action menu for the shipment, similar to how the "ready for shipment" one works.   Also, I would not allow stock entry from the shipment until it is AT_DESTINATION_DEPOT.   

I'm not sure we need a distinction between "Packed" and "Ready to go". What do you think?

And please add a boolean flag that all items in the shipment were received.   It would default to FALSE.  Once all the items are unpacked, the stock entry code would compare received items to the packing list and determine if all items were received and mark the final value of the flag appropriately.    That would reduce the overhead in queries to list shipments, etc.

It might be useful to have a %delivered field (based on number of items?) for the shipments registry.

jmcameron:
This PR does these updates to shipment handling
- Provides a way to mark a shipment as *delivered* to the destination depot (but not unpacked or stock entered).  Note that this  is optional but would be helpful to perform this immediately when a shipment arrives at the destination depot especially if it may take some time to unpack the shipment and enter it into the destination depot stock.
- Adds a [+New Shipment] button on the Shipments Registry for convenience
- Updates the Shipments documentation (docs/en/stock-management/movement.shipments.md
- Updates to the ready-to-ship and completed confirmation dialogs.  Added a new confirmation dialog for marking a shipment as delivered.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6470

**TESTING**
- Use bhima_test
- Go to: Shipments > New Shipment
  - Create  simple shipment with one item (eg, code=MOT...)
  - Go through these steps with this shipment
    - Mark it ready to ship
    - Do Stock Exit on it (use the Shipments Registry action menu to simplify this)
    - Mark the shipment as Delivered
    - Do the Stock Entry on it (use the Shipments Registry action menu)
    - The shipment will automatically be marked complete
    - Check the Tracking Log (action menu) to verify the actions taken
- Create a new shipment (use the [+New Shipment] button in the Shipments Registry
   - Add an item with quantity > 1 (eg quinine)
   - Do Stock Exit on it (use the Shipments Registry action menu to simplify this)
   - SKIP marking the shipment as Delivered
   - Do a partial Stock Entry on it (use the Shipments Registry action menu) 
      - But edit the lot to reduce the quantity
   - The shipment should now show a status of "Partial"
   - Use the action menu to mark the shipment as completed anyway
   - Check the Tracking Log (action menu) to verify the actions taken

-----
[View rendered docs/en/stock-management/movement.shipments.md](https://github.com/jmcameron/bhima/blob/shipments-updates/docs/en/stock-management/movement.shipments.md)


mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #6710, opened 2022-06-27
Fix issue with entry of multiple assets via Lots entry modal

### LINKED ISSUE #6709: Stock entry of assets limited with quantity 1
This issue is about the fact that during a stock entry of assets, when we click on the "Lots" link, the default and fixed quantity is set to one.

In case we want to enter 3 DELL XPS with serial number "S1", "S2" and "S3"; the normal process is: 

- Go to *Stock Entry* Module
  - Click on *Integration*
  - Set the date
  - Set the description
  - Click on the "Add" button for adding lines for inventories
    - Select an asset by typing its name or code
    - Click on the "Lots" link for adding different lots of the inventory in the line

In the modal for inventories which are assets, we have a fixed quantity of 1, in our example, we would like to set quantity to 3 for our three serial numbers of DELL XPS.

jmcameron:
When you enter lots (by integration) and you want to add multiple instances of an asset, this PR makes this work more easily from the Lots modal popup.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6709

**TESTING**
- use bhima_test
- Go to:  Stock > Stock Entry
  - Select [Integration]
  - Add a new line for an asset.  Use inventory code:  MOT.HCRF250RX  for a motorcycle
  - Click on the **Lots** link to bring up the Lots entry modal. 
  - You should now be able to change the global quantities
  - Since Assets must be unique with quantity===1, the quantity field is not even shown
  - Try added multiple lines (with different LOT ids)
  - Click [Submit] on the modal
  - Try clicking the **Lots** link again to verify that the info on the non-yet-saved lots is still there
  - Close the modal
  - Try adding a non-asset (eg, inventory code DINJ_TRIB1A2_0) and use the "*Lots** link and verify that the Lots modal works as normal for non-assets
  - Close the Lots entry modal
  - Click on the [Submit] button to enter the stock.   
  - Check the receipt document to verify correct enty
- Go to Asset Management > Assets Registry and verify that the new assets appear there.

### DIFF
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -565,8 +565,10 @@ function StockEntryController(
     stockLine.tracking_expiration = inventory.tracking_expiration;
     stockLine.unique_item = inventory.unique_item;
     stockLine.is_asset = inventory.is_asset;
-    if (inventory.is_asset) {
-      stockLine.quantity = 1;
+    if (stockLine.lots && stockLine.lots.length > 0) {
+      stockLine.quantity = stockLine.lots.reduce((n, row) => n + row.quantity, 0);
+    } else {
+      stockLine.quantity = inventory.is_asset ? 1 : 0;
     }
 
     StockModal.openDefineLots({
--- a/client/src/modules/stock/entry/modals/lots.modal.html
+++ b/client/src/modules/stock/entry/modals/lots.modal.html
@@ -32,7 +32,6 @@
 
       <input
         ng-if="$ctrl.isCostEditable"
-        ng-disabled="$ctrl.stockLine.is_asset"
         class="form-control"
         type="number"
         ng-change="$ctrl.onChangeQuantity()"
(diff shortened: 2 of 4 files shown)


mbayopanda (commented):
I do not think it is related to this PR, but I have noticed that the quantity is doubled in the receiver depot when we do the entry with this link in the "Expedition Registry": 

image

In the case below I have just received 3 Vehicule but it shows 6 instead.

We can see the quantity doubled in the "Lot Registry"

image

To reproduce that: 
- yarn build:db
- Do a stock entry operation in a the Depot Principal of 3 Assets (MOT.HCRF250RX)
- Do a stock exit operation to the Secondary Depot of 3 Assets
- Do the stock entry operation in the Secondary Depot by using the link in the "Expedition Registry"
- Consult the "Articles en stock" and "Lot en stock" registry to check quantities

mbayopanda:
Here is another observation for a "Stock Entry" by using the "Transfert" button in the "Stock Entry" module, when we select the shipment the grid is filled with the doubled quantity: 

image

PR #6712, opened 2022-06-28
Add cost center column to the employees registry

jmcameron:
In order to facilitate the implementation of Cost Centers, this PR adds a column for Cost Centers in the employee registry.  It also adds a custom search field in the Employee registry page so that only the employees in a specific Cost Center can be listed or to list employees who are not associated with any Cost Center.

This PR also does a minor renaming of a Cost Center-related report i18n translation token to be consistent with other similar tokens.

**TESTING**
- Use bhima_test
- Create a new employee but do NOT assign them to a service
- Go to Cost Center Management > Cost Centers
  - Use the [Menu] -> Columns and show Service and Cost Center columns
  - Do [Menu] -> Save grid configuration
  - Verify that the cost centers and services are shown for the employees.  Note that no service center will be shown for the employee you just created
  - Use the [Search] modal and scroll down to select a Cost Center.  Close the Search modal
  - Verify that only employees within the specified Cost Center are shown
  - Try deleting the filter (from the filter bar and verify its operation to get the full list again
  - Use the [Search] modal and scroll down to select **Cost Center Not Defined**
  - Verify that EM.TE.3 and the newly created employee are listed (who is not assigned to service and therefore is not assigned to any cost center)
  - Export the PDF version of the report and verify that the filter info at the top of the document is correct.

**NOTE**: The PDF export now shows service and cost center.  The CSV export does also EXCEPT when you select **Cost Center Not Defined**.  Then it suppresses the cost center column.


mbayopanda (approved):
The feature is working as expected.

bors r+

PR #6722, opened 2022-07-04
Fix lost asset serial numbers in stock entry for shipments

### LINKED ISSUE #6720: Fix bug about losing asset serial number in stock entry
When doing stock entry for assets via shipments, the serial number seems to be lost when you examine the incoming lots.

jmcameron:
Fixes a bug in transferring assets using the Shipments interface where the serial_number was being lost during the stock entry phase.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6720

**TESTING**
- Use bhima_test
- Go to the shipments registry:  Shipments > Shipment Registry
   - Note the lot name and serial number for the motorcycle to use in the next step
   - Create a shipment to transfer the selected motorcycle (code = MOT...) from the Primary Depot to the secondary Depot.
   - Use the action menu for the new shipment to mark is as ready to ship
   - Use the action menu for the new shipment to do the stock exit
   - Use the action menu for the new shipment to do the stock entry
      - BEFORE clicking [Submit] to complete the stock entry, click on the link to the **Lots** for the lot (asset)
      - Verify that the original serial number is still present
      - Close the lots update modal
      - Click on the main [Submit] to complete the stock entry
- In the shipments registry
   - Verify that the serial number of the selected asset is unchanged

### DIFF
--- a/client/src/modules/stock/entry/entry.js
+++ b/client/src/modules/stock/entry/entry.js
@@ -515,6 +515,7 @@ function StockEntryController(
           isValid : true,
           lot : items[index].label,
           quantity : item.quantity,
+          serial_number : items[index].serial_number,
           expiration_date : new Date(items[index].expiration_date),
           uuid : items[index].uuid,
         });
@@ -776,6 +777,7 @@ function StockEntryController(
 
         const inventory = inventoryStore.get(line.inventory_uuid);
         const entryDate = vm.movement.date || Date();
+
         line.code = inventory.code;
         line.label = inventory.label;
         line.unit_cost = inventory.price;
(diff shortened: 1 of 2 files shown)


client/src/modules/stock/entry/entry.js
```diff
@@ -515,6 +515,7 @@ function StockEntryController(
           isValid : true,
           lot : items[index].label,
           quantity : item.quantity,
+          serial_number : items[index].serial_number,
```
mbayopanda, line 518: 👍🏽

mbayopanda (approved):
Thank you @jmcameron 

bors r+

PR #6727, opened 2022-07-07
Add support for updating Cost Centers in the accounts data

### LINKED ISSUE #5983: Adding a cost center to an account does not refresh the grid
The account management grid does not refresh when we associate an account with a cost center.

jmcameron:
This PR adds a new function to the main menu under the Cost Centers registry to force updating of the cost center information for all accounts.   This function takes into account the exceptions, etc, in account references.  This is a necessary step for end-users to be able to implement cost centers effectively.

This function was necessary since currently there is no other way to update that information.  Eventually we should add a global SQL zUpdateAccountCostCenters() function that can be called any time we change account reference definitions or change which account references any cost center uses.

Closes https://github.com/IMA-WorldHealth/bhima/issues/5983

**TESTING**
- Use bhima_test
- Go to: Finance > Account Management
  - Note that only one account (60310016) has a Cost Center associated with it
- Go to:  Administration > Account References
  - Notice the new "Cost Center" column indicating which cost center uses the account reference (if any)
  - Edit p_admin and add the 101 account, click [Submit]
  - Verify that the 101 account was added
- Go to: Cost Center Management > Cost Centers 
  - Use the [Menu] > Update Accounts function
  - This will take you to the Accounts page
  - Verify that the accounts have been updated
- Double-check exclusions
  - Edit account reference p_admin again
    - Add 23, but exclude 24480040 (Mobiliers Hopital)
    - Do Cost Centers > [Menu] -> Update Accounts again and verify the correct changes


lomamech (approved):
It's very cool

server/controllers/finance/cost_center.js
```diff
@@ -337,6 +338,56 @@ function setAllocationStepOrder(req, res, next) {
...
+    const accRefs = await db.exec(accRefsql, [center.id]);
+
+    const updateAccSql = 'UPDATE account SET cost_center_id = ? WHERE id = ?';
+
+    // eslint-disable-next-line no-restricted-syntax
+    for (const ref of accRefs) {
+
+      // Get a list of accounts belonging to the cost center
+      // eslint-disable-next-line no-await-in-loop
+      const accounts = await accountReferences.getAccountsForReference(ref.abbr);
```
lomamech, line 373: :tada:

lomamech (commented):
It's very cool

lomamech (commented):
It's very cool

PR #6771, opened 2022-08-04
Use correct translation tokens for reports

jmcameron:
The recent update of many reports apparently used the incorrect/obsolete translation items for the report names.


jmcameron:
This is such a minor tweak, that I will go ahead and release it without anyone else reviewing it.

bors r+

PR #6804, opened 2022-08-31
Fix issue with stock exits without permissions for destination depot

### LINKED ISSUE #6800: Improve error handling for stock exits for shipments
If a user tries to do a stock exit for shipment there a couple of problems:

- Assume the "Activate the restriction of distribution depots" flag in the stock settings is set to YES, and the user does NOT have permissions to access the destination depot.  When that user tries to do a stock exit for that shipment, clicks on YES for the "Do you have a shipment ready" item in the depot selection modal, they will be able to select the shipment but the depot will not be filled in since they do not have permissions to access the destination depot..    This should not be necessary for a stock exit so it should be fixed.

See also https://github.com/IMA-WorldHealth/bhima/issues/6793

jmcameron:
This PR fixes an issue with Stock Exits when the Stock setting "Activate the restriction of distribution depots" flag is set to YES and the user does not have permissions to access the destination depot.  Now the stock exit will work regardless of whether the user has permissions to access the destination depot.  But it does add a check to verify that the user has permissions to access the origin depot (if the Activate the restriction of distribution depots" flag is set to YES).

Closes https://github.com/IMA-WorldHealth/bhima/issues/6800

**TESTING**
- In the Stock Settings, set "Activate the restriction of distribution depots" to YES
- Create a shipment and mark it ready to ship
- In the Administration > User Management > *action menu* > Depot Management, disable access to the destination depot in the shipment.
- Go to:  Stock > Stock Exit
  - Click on the [Depot] button and in the pop-up modal, click that you have a shipment ready
  - Select the just created shipment
  - The destination depot should appear in the top field of the modal
  - The [Submit] button should be active.
  - Click on the [Submit] button
  - Do NOT complete the stock exit
- Go to:  Shipments > Shipments Registry
  - In the action menu for the shipment created for this test, select the "Exit Stock for this shipment" 
  - This should take you to the Stock Exit form with everything set up to complete the stock exit


lomamech (approved):
I just checked the feature and it works really great, look Good

PR #6807, opened 2022-09-01
fix depot details and distribution list

### LINKED ISSUE #6806: Load depot details is broken
Loading information of depot is broken

image

image

mbayopanda:
This PR fixes the display of information about  a depot and refactor the way of selecting destination depots

image

Close #6806

### DIFF
--- a/client/src/modules/depots/modals/depot.modal.js
+++ b/client/src/modules/depots/modals/depot.modal.js
@@ -3,9 +3,10 @@ angular.module('bhima.controllers')
 
 DepotModalController.$inject = [
   '$state', 'DepotService', 'NotifyService', 'SessionService', 'params',
+  'FormatTreeDataService',
 ];
 
-function DepotModalController($state, Depots, Notify, Session, params) {
+function DepotModalController($state, Depots, Notify, Session, params, FormatTreeData) {
   const vm = this;
 
   vm.depot = {};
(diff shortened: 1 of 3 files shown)


client/src/modules/depots/modals/depot.modal.js
```diff
@@ -109,4 +129,39 @@ function DepotModalController($state, Depots, Notify, Session, params) {
       })
       .catch(Notify.handleError);
   }
+
+  /**
+   * destination depots handlers
+   */
+  function setNodeValue(childrens, depot) {
```
jniles, line 136: Why not just use the TreeService to be this data structure?

That might give you better guarantees (if I recall, it is tested) than copy and pasting code from elsewhere.

jniles: Alternatively, you could also use the `bhCheckboxTree` (https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/components/bhCheckboxTree/bhCheckboxTree.js) to create the checkbox tree.  This would reduce the amount of logic in the controller.

lomamech (approved):
The structure of the code is similar to that of assigning repositories to users, I tested it and it works well

client/src/modules/depots/modals/depot.modal.js
mbayopanda: @jniles the `bhCheckboxTree` is a great tool and easy to use, however, there are some things that must be handled on it : 
- The selection of parent implies the selection of children (we cannot select only a parent without selecting children)

image

- The component does not have a filter

I will add these points as issues but for now I will merge the current progress.

lomamech: I also think that we can write another issue to solve this problem, but also using the **bhCheckboxTree** here would also imply that we can modify the module for assigning deposits to users

PR #6821, opened 2022-09-20
feature(Manage Drug Packing)

### LINKED ISSUE #6811: Problem with drug packaging
The BHIMA system manages inventories (Medicines and other Pharmaceutical products) using the smallest unit

But the managers of stocks and depots take packaging and boxes into account.

This situation means that when managers use the BHIMA systems for stock entry and exit, they are forced to multiply for stock entries and divide for stock exits. without forgetting that the suppliers transmit the price of the articles by boxes and by units.

To overcome this difficulty, this issue proposes to add in the stock management parameters, the possibility of being able to provide information on the packaging of pharmaceutical products and to make certain calculations automatically,

when entering stock, for example, it would suffice to enter the number of boxes, for example 120, the packaging Box of 50, cost of the box 4.5 so that the BHIMA system automatically fills in the quantity at 6000 and the unit cost at 0.09

The same goes for the stock issue, the manager will only have to enter the number of boxes for the system to convert it into a smaller unit.

There will be no question of modifying the structure of the database, so as not to generate any error in the BHIMA database

lomamech:
Pharmaceutical packaging configuration management
- Setup
- Purchase order
- Shipping management
- Stock entries and exits

closes #6811


lomamech:
**User guide for the test**

STOCK MANAGEMENT WITH PACKAGING

We have found that in practice, warehouse managers are not used to managing inventory in terms of units but in terms of boxes, bundles and packaging. We also found that they are forced to perform arithmetic operations several times when using the BHIMA system.

This is why we have implemented packaging management in the BHIMA system, this new implementation does not modify the logic of stock management with regard to taking into account the smallest unit.

**I. SETUP**
To be able to configure the packaging management in BHIMA, 3 levels of settings are required, at the level of:
     • Configuration of stock settings parameters
     • The depot configuration
     • The inventory parameter

**_I.1. Inventory management parameters_**
To manage the packaging management configuration, you must first configure the packaging of pharmaceutical products.

image

**_I.2. Repository configuration_**
Regarding the depot, only depots configured as a warehouse can be configured for the management of pharmaceutical products by packaging.

image

I.3. Inventory Setup
Regarding inventories, not all inventories are presented with packaging, which is why this possibility should be configured for each inventory.

image

Only consumable inventories can be packaged by box

**II. PURCHASE ORDER**
When establishing a purchase order, if packaging management had been configured in stock settings, a new column will be added to the pharmaceutical product selection grid with a box icon next to it. of each inventory configured as packaged per box.

image

When clicking on the box icon, an interface will appear to allow the definition of the number of boxes, the size of the packaging as well as the unit cost of the box.

_If for example we set the number of boxes to 25, the size of the packaging to 50 and the unit cost of the box to $5.50 the system will prefill the grid with the quantity to 1250 and the unit cost to $0.11_

image

This feature will facilitate the work of managers and reduce the risk of making arithmetic errors.

**III. STOCK ENTRY MANAGEMENT**

For the management of stock entries, the definition of the packaging of the box, intervenes at the time of the definition of the batches, after having chosen selected the inventory (configured to be managed by box) but also in the condition where all the parameters are activated.
The batch definition interface looks like this.

image

New input fields to define the number of boxes, the size of the packaging, i.e. (TAB of 10, 20, 50, 100, 200, 500, 100 or others), the unit cost of the box and the system will calculate automatically the overall quantity as well as the unit cost and will save in the database after validation the size of the packaging for the batch concerned.


For stock entries from a purchase order
it is possible that during a delivery that the supplier serves the same product with two different packaging, suppose that during the purchase order that 2000 tablets of Chloramine T, 500mg, Tab were ordered and that the supplier delivered the same product with 3 different packaging
     • A box of 1000 (1000 tablets)
     • A box of 500 (500 tablets)
     • and five boxes of 100 (500 tablets)

Note that the BHIMA system deactivates the modification of the number of lines for the cases of stock entries from a purchase order.

image

To be able to register several packaging, you will have to check the option to enter the stock with another packaging each time, so that a new line is automatically added to the stock entry grid.

image

Noted : For stock entries from a purchase order, the unit cost cannot be modified.


**IV. STOCK OUTPUT MANAGEMENT**
For stock issue management, for depots that are configured for packaging management, the BHIMA system automatically adds a new column with a box icon, this icon allows the user to define the stock quantity based on the number of boxes and their packaging.

image

**V. SHIPPING MANAGEMENT**
Warehouse managers will be called upon to make shipments, which are generally done in terms of boxes, when creating a new shipment, the BHIMA system will first check whether packaging management has been configured at the level of the parameters but also at the deposit level, once these two conditions are met. The system will once again show a box icon next to each box-packable inventory.

image

For inventories whose batches have been defined in relation to their packaging, the system will automatically retrieve the size of their packaging.

lomamech:
@mbayopanda and @jmcameron this PR is ready for a review

jmcameron:
Nice documentation.  I encourage you to add it to the BHIMA docs (eg, in docs/en/stock-management/packaging.md with appropriate links).

client/src/modules/inventory/list/modals/actions.tmpl.html
```diff
@@ -153,14 +153,22 @@
...
 
-
     <div class="checkbox">
       <label>
         <input type="checkbox" name="consumable" ng-true-value="1" ng-false-value="0" ng-model="$ctrl.item.consumable">
         <span translate>FORM.LABELS.CONSUMABLE</span>
       </label>
     </div>
 
+    <div ng-if="$ctrl.item.consumable" class="checkbox">
```
jmcameron, line 163: Is it possible for "non consumable" items to be packaged too?

client/src/modules/stock/entry/modals/StockEntryModalForm.service.js
```diff
@@ -28,6 +28,8 @@ function StockEntryModalForm(uuid) {
     this.isInvalid = true;
     this.isValid = false;
     this.identifier = uuid();
+    this.number_box = row.number_box || null;
+    this.package_size = row.package_size || null;
```
jmcameron, line 32: It seems like the this uses different terms:  package vs box.  I suggest using one for both.  Eg:  this.number_packages and this.package_size. (or with "box").  "Package" may be a little better than "box" since "package" seems more generic.  I can imagine that meds might come in tubes or barrels containing multiple doses.  In that case "box" seems awkward.

jmcameron (commented):
I took a casual look at the code and it seems fine.  
But I did have a problem and was not able to get it to work for me.
I tried (1) enabling packaging in stock settings, (2) updating the depot to support packaging, (3) editing an inventory item to enable packaging.   Then I tried to create a purchase order and could not get the dialog to define the number of boxes, size of the packaging, and unit cost of each box to work.  I could not select anything for the "Pack Size" and the [Submit] button would not work (even though it was not greyed out).

Also, I encourage updating the purchase order document to spell out the packaging details for each item that is packaged in purchase order.

lomamech:
> Nice documentation. I encourage you to add it to the BHIMA docs (eg, in docs/en/stock-management/packaging.md with appropriate links).

I will do it

client/src/modules/inventory/list/modals/actions.tmpl.html
lomamech: I don't think this notion could apply to services and other medical acts

lomamech:
> I took a casual look at the code and it seems fine. But I did have a problem and was not able to get it to work for me. I tried (1) enabling packaging in stock settings, (2) updating the depot to support packaging, (3) editing an inventory item to enable packaging. Then I tried to create a purchase order and could not get the dialog to define the number of boxes, size of the packaging, and unit cost of each box to work. I could not select anything for the "Pack Size" and the [Submit] button would not work (even though it was not greyed out).
> 
> Also, I encourage updating the purchase order document to spell out the packaging details for each item that is packaged in purchase order.

@jmcameron If you tested with the bhima_test database, could you tell me the repository as well as the inventory you used for the test so that I can also check

client/src/modules/stock/entry/modals/StockEntryModalForm.service.js
lomamech: I will do it

lomamech:
@jmcameron  I just finished with the first series of reviews

jmcameron:
After figuring out what I did wrong, I was able to create a purchase order with packaging.
However when I entered a price that was too high for the box, I got the dialog to confirm this cost but it was comparing the cost per pox to the previous cost per pill.   So some adjustment of that dialog is needed when packaging is involved.

lomamech:
> After figuring out what I did wrong, I was able to create a purchase order with packaging. However when I entered a price that was too high for the box, I got the dialog to confirm this cost but it was comparing the cost per pox to the previous cost per pill. So some adjustment of that dialog is needed when packaging is involved.

The comparison is not made in relation to the unit cost per box but to the cost per pill

client/src/i18n/fr/form.json
```diff
@@ -297,6 +298,7 @@
             "AMOUNT": "Montant",
             "AMOUNT_PAID" : "Montant Payé",
             "ARTICLES":"Articles",
+            "PACKAGE_UNIT_COST": "Coût unitaire du pacquet",
```
mbayopanda: It must be `Coût unitaire du paquet`

client/src/i18n/fr/form.json
```diff
@@ -952,6 +957,7 @@
             "NAME": "Entrer le nom",
             "NOTES": "Commentaire",
             "ORIGIN": "Origine",
+            "PACKAGING": "Entrer le Conditionnement du pacquet",
```
mbayopanda: `Entrer le Conditionnement du paquet`

client/src/i18n/fr/inventory.json
```diff
@@ -48,6 +48,7 @@
         "INVENTORY_TYPE_INFO" : "inventory_type : le type d'inventaire tel que Article, Service, Assemblage ou Autre",
         "INVENTORY_UNIT_INFO" : "inventory_unit : les unités de l'inventaire telles que Piece, Tablet, Box, Bottle, etc",
         "IS_BROKEN" : "est endommagé",
+        "IS_COUNT_PER_CONTAINER" : "est conditionné par boîte",
```
mbayopanda: Since it is related to inventory and it is optional, you can set it as "Peut être conditionné par paquet"

mbayopanda (commented):
Great job @lomamech, the feature is really helpful and it is working fine.

I have some comments about: 
- Adding information (popover) in the lot modal for: "Nombre des paquets", "Taille de l'emballage", "Cout unitaire du paquet", "Quantité globale" et "Coût unitaire".
- Adding information about packing such as "Pack Number" et "Pack Size" in receipts for: Stock Entry, Stock Exit and Purchase Order

lomamech:
@mbayopanda all checks have passed

mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #6843, opened 2022-10-15
Bug problem with asset register

### LINKED ISSUE #6842: Problem with fixed asset register
We have noticed that when adjusting assets, transferring and removing stock as a loss, the asset register continues to display assets whose quantity in stock is equal to Zero, this issue proposes to solve this problem.

image

lomamech:
- Hide the asset where quantity in stock is 0
- Add HAVING filters in SQL Clause format
**_Bonus_**
- Set Serial Number not required in lot entry stock
- Dont display expiry information for non Expirable asset

closes https://github.com/IMA-WorldHealth/bhima/issues/6842

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -309,7 +309,8 @@ async function getAssets(params) {
       ) AS last_scan ON last_scan.asset_uuid = l.uuid
   `;
 
-  const groupByClause = ` GROUP BY l.uuid, m.depot_uuid ORDER BY i.code, l.label `;
+  const groupByClause = ` GROUP BY l.uuid, m.depot_uuid `;
+  const havingClause = ` HAVING quantity > 0 `;
 
   const filters = getLotFilters(params);
   if (['scanned', 'unscanned'].includes(params.scan_status)) {
@@ -318,6 +319,8 @@ async function getAssets(params) {
   }
   filters.setGroup(groupByClause);
 
+  filters.setHaving(havingClause);
+  filters.setOrder('ORDER BY i.code, l.label');
   const query = filters.applyQuery(sql);
   const queryParameters = filters.parameters();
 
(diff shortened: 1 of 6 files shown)


server/controllers/stock/core.js
```diff
@@ -318,6 +319,8 @@ async function getAssets(params) {
   }
   filters.setGroup(groupByClause);
 
+  filters.setHaving(havingClause);
```
jmcameron, line 322: 👍

server/lib/filter.js
```diff
@@ -215,8 +228,9 @@ class FilterParser {
     const conditionStatements = this._parseStatements();
     const order = this._order;
     const group = this._group;
+    const having = this._having;
 
-    return `${sql} WHERE ${conditionStatements} ${group} ${order} ${limitCondition}`;
+    return `${sql} WHERE ${conditionStatements} ${group} ${having}${order} ${limitCondition}`;
```
jmcameron: I think you may want to add a space between the ${having} and ${order} so the developer does not need to add a space when using both.

jmcameron (commented):
I looked over the code and pointed out a minor issue.

I tested (with bhima_test) and it all seems work work fine.

However, when I looked at the articles in stock and hovered over the warning triangle for Quinine, I noticed that the tooltip had not been fully translated.  See:
image

This can be fixed by editing client/src/modules/stock/inventories/registry.js and adding a line to translate the unit_type after line 275 so it looks like this:

        // set status flags
        rows.forEach(row => {
          setStatusFlag(row);
          row.status_translated = $translate.instant(Stock.statusLabelMap(row.status));
          row.unit_type = $translate.instant(row.unit_type);
        });

mbayopanda:
@lomamech could you consider @jmcameron suggestions

lomamech:
@mbayopanda @jmcameron  all checks have passed

PR #6844, opened 2022-10-16
fix integration test time-out

mbayopanda:
Increate test delay for integration test

### DIFF
--- a/test/integration-stock/setup.js
+++ b/test/integration-stock/setup.js
@@ -39,6 +39,11 @@ before(() => {
   return agent.post('/auth/login').send(user);
 });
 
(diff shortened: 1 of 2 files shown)


test/integration-stock/setup.js
```diff
@@ -39,6 +39,11 @@ before(() => {
   return agent.post('/auth/login').send(user);
 });
 
+beforeEach(done => {
+  this.timeout(10000);
```
jniles, line 43: This isn't defined in an arrow function :wink:

You'll need to use `function () {}` for this.

jniles:
bors try

PR #6852, opened 2022-10-28
Use of ui-select instead of select in inventory modal

mbayopanda:
This PR replace `select` tag with `ui-select` in the inventory modal.

image


lomamech (approved):
It works well

image

PR #6855, opened 2022-11-01
Use of tabs in lots modal

mbayopanda:
This PR split the form's inputs into tabs to optimize the height of the modal

image


lomamech (approved):
It works very well

PR #6868, opened 2022-11-16
bug(Report Needed inventory Scan)

### LINKED ISSUE #6867: Fixed scan report errors
- The needed inventory scans reports does not work correctly, this issue proposes to explore this problem

lomamech:
- Removal of the reference number parameter which makes a request by looking for the reference number element of the scan on the reference number column of the batch

- Addition of the Date function for the selection of the very last scan carried out for an asset

closes #6867

### DIFF
--- a/server/controllers/asset_management/assets/reports/needed_inventory_scans.js
+++ b/server/controllers/asset_management/assets/reports/needed_inventory_scans.js
@@ -39,7 +39,6 @@ async function neededInventoryScansReport(req, res, next) {
       is_asset : scan.is_asset,
       scan_start_date : new Date(scan.start_date),
       scan_end_date : new Date(scan.end_date),
-      reference_number : scan.reference_number,
       // Do not pass in scan_status, filter after the query
     };
 
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -255,11 +255,11 @@ async function getAssets(params) {
   let scanWhere = 'WHERE s2.uuid IS NULL';
   if ('scan_start_date' in params) {
     const startDate = moment(params.scan_start_date).format('YYYY-MM-DD');
-    scanWhere += ` AND scan.created_at >= DATE(${db.escape(startDate)})`;
+    scanWhere += ` AND DATE(scan.created_at) >= DATE(${db.escape(startDate)})`;
   }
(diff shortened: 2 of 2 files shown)


server/controllers/stock/core.js
```diff
@@ -255,11 +255,11 @@ async function getAssets(params) {
   let scanWhere = 'WHERE s2.uuid IS NULL';
   if ('scan_start_date' in params) {
     const startDate = moment(params.scan_start_date).format('YYYY-MM-DD');
-    scanWhere += ` AND scan.created_at >= DATE(${db.escape(startDate)})`;
+    scanWhere += ` AND DATE(scan.created_at) >= DATE(${db.escape(startDate)})`;
```
jmcameron, line 258: Good catch.

server/controllers/asset_management/assets/reports/needed_inventory_scans.js
```diff
@@ -39,7 +39,6 @@ async function neededInventoryScansReport(req, res, next) {
       is_asset : scan.is_asset,
       scan_start_date : new Date(scan.start_date),
       scan_end_date : new Date(scan.end_date),
```
jmcameron, line 41: I do not think you should remove this.   The purpose of the "reference_number" is to allow correlation of scans to some reference such as a purchase order, etc.   It is used when you create an "Needed Inventry Scan" and is saved in the 'required inventory scan".  It should be coming into this function via the database query in line 35. And it appears in the report itself, so deleting it here is a problem.  You should check the value of "scan" after the database query on line 35 and make sure reference_number is reasonable (or blank).  If it is coming in as NULL, you should probably make sure the database table has a default value (like an empty string) of that the assignment you removed from like 42 has a fallback/default value:     scan.reference_number || '',

Also, you should double-check to make sure we are using the "reference_number" correctly.  It occurs to me that we may have a logical issue:  It seems to me that each asset could have its own reference number (bought under different contracts, etc).  In that case 'reference_number' should be have a column of its own in the Required Inventory Report.

jmcameron (commented):
I reviewed the code.  The DATE() fixes look fine.  However I do not think removing the reference_number is correct and deserves a closer look before releasing it.

I tested this briefly with bhima_test and it seems to work correctly.

PR #6887, opened 2022-12-14
bug(Fix and Improve stock registry)

### LINKED ISSUE #6870: Improve and resolve proble of entry stock by transfert
- No longer displayed transfer movements for deposits to which the user does not have access
- Solve the problem of entry into stock by transfer when the restriction option Limit stock by user's depot permissions is active

lomamech:
- Correct the functionality which limits the stock entry by transfer to only deposits managed by the user in the case where the option management of deposits Limit stock by user's depot permissions is active

- Added the option to show or hide the display of transfers in transit

- Restricted the display of transfers in transit to users who have access to deposits

closes #6870


lomamech:
@mbayopanda need a review

mbayopanda (approved):
LGTM 👍🏽

PR #6892, opened 2022-12-21
Feature problem with creating requisition request

### LINKED ISSUE #6875: Problem with creating a requisition request
We had noticed that the option "Limit stock by user's depot permissions" is activated that it is impossible for a user to be able to make the requisition request to a depot to which he does not have access?

The solution that can be proposed is that of activating the option "Activate the restriction of distribution depots" with this option, a depot can only make the requisition to a depot for which it has the distribution right

lomamech:
- Restoring the Activate the restriction of distribution depots option
- For the management of requisitions, you can only make a requisition for a deposit for which you have the right to manage
- when the restrict distribution option is activated, the requisition request can only be made to a structure that is authorized to distribute to the requesting depot
- Cancel the supplier depot selection if the user changes the requesting
  depot
- Restrict the display of the requisition register to depots managed by the user either as a requestor or as a supplier
- Added the ability to select a parent repository without selecting child repositories
- Clean the supplier depot when the selection change

closes https://github.com/IMA-WorldHealth/bhima/issues/6875

### DIFF
--- a/client/src/i18n/en/stock.json
+++ b/client/src/i18n/en/stock.json
@@ -72,6 +72,7 @@
     "TRACK_CONSUMPTION":"Track consumption",
     "TRACK_EXPIRATION":"Track expiration",
     "ERRORS" : {
+      "DEPOT_NOT_ALLOWED_DISTRIBUTE": "ERROR! <b>'{{ depot }}'</b> depot is not allowed to distribute to any depot",
       "EXCESSIVE_QUANTITY": "Excessive Quantity",
       "INVALID_GLOBAL_QUANTITY" : "Invalid Global Lot's Quantity",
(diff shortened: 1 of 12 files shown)


lomamech:
@mbayopanda to check this PR it is necessary to activate these two options
- Limit stock by user's depot permissions
- Activate the restriction of distribution depots
and configure for each repository the repositories **allowed for distribution**

image

mbayopanda (approved):
@lomamech requisitions are created correctly, and the beneficiary depot is based on user permission.

I have noticed that the requisition registry still has problems with status which remain pending while the beneficiary receives a part from the requisition.

I will merge this PR, the issue with requisition registry can be done in another PR

PR #6910, opened 2023-01-17
fix(tests): ensure 2023 exists

jniles:
This commit adds the 2023 fiscal year to begin work on fixing the semaphore tests.


jniles:
bors try

jniles:
bors try

jniles:
bors try

jniles:
bors try

jniles:
Let's try again.

bors try

jniles:
@mbayopanda this is ready for review.

It appears that either semaphore got more limited hardware or the chrome browser became slower as the tests were timing out on PDF rendering and processing of several heavy MySQL queries.  I am guessing the former.

To fix, I increased the timeout on the tests (from 2 seconds to 20 seconds).  I also added 2023 as a fiscal year (which we have to do every year) and made sure we dropped support for old NodeJS versions.  Finally, I bumped the dependencies of several of the modules.

mbayopanda:
@jniles good catch for the timeout extension, it seems like you didn't change anything related to timeout on the Semaphore-CI side.

PR #6935, opened 2023-02-08
Fix stock value calculation

mbayopanda:
This PR changes the order of calling the procedure `ComputeStockValue` so that calculation is done after stock movement.

This PR also set the stock adjustment movements to use the WAC value.

### DIFF
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -349,7 +349,19 @@ async function createInventoryAdjustment(req, res, next) {
       );
     }
 
+    // get unit costs from stock_value
+    const inventoryUuids = lots.map(l => db.bid(l.inventory_uuid));
+    const unitCosts = await db.exec(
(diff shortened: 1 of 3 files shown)


lomamech (commented):
Do you think you can add the possibility of including Pharmacists' feedback in this PR or in another

- If this is the case, it will be necessary that the entries coming from the integrations as well as the purchase orders can be done at the defined cost or at the cost of the supplier
- The transfer from the source deposit must always go out at the time of purchase and be received with the same cost
- For other deposits, the exit must be made in relation to the weighted average cost or the weighted average cost fixed in the system

lomamech:
> Do you think you can add the possibility of including Pharmacists' feedback in this PR or in another
> 
> * If this is the case, it will be necessary that the entries coming from the integrations as well as the purchase orders can be done at the defined cost or at the cost of the supplier
> * The transfer from the source deposit must always go out at the time of purchase and be received with the same cost
> * For other deposits, the exit must be made in relation to the weighted average cost or the weighted average cost fixed in the system

@mbayopanda How to discuss the different points mentioned above will be treated in another Issue

PR #6940, opened 2023-02-15
feature(Shipment Management)

### LINKED ISSUE #6939: Restriction on shipping management
- Restrict the establishment of a shipment to only depots whose shipper depots are authorized to supply
- Restrict consultation of the shipping register to only depots for which the user has management or supervision rights

lomamech:
- Improve depot component add option for select only allowed distribution depot
- Improve filter register display only data of shipment for the depots managed and supervised by User

closes #6939


mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #6958, opened 2023-03-03
feature(Purchase modules Product Packaging)

### LINKED ISSUE #6956: Problem on the management of product packaging
We have noticed that at the warehouse level some products are ordered in terms of packages but are consumed in terms of units. This issue proposes to modify the appearance of purchase orders, to display data related to packages in terms of packages and maintain stock entries in terms of units

lomamech:
- Added a new column to keep the packaging size,
- Change the appearance of the order form to display information about packages or boxes
- Addition of information concerning the packaging on the registers of detailed purchases

closes #6956

### DIFF
--- a/client/src/i18n/en/table.json
+++ b/client/src/i18n/en/table.json
@@ -52,6 +52,7 @@
             "CHARGES": "costs",
             "CLIENT": "Client",
             "CODE": "Code",
+            "PACKAGED": "Packaged",
             "CONTRIBUTION_DEDUCTIONS": "Contributions and deductions",
             "CONSULT": "Consult",
             "CONSUMPTION": "Detailed consumption",
--- a/client/src/i18n/fr/table.json
+++ b/client/src/i18n/fr/table.json
@@ -52,6 +52,7 @@
             "CHARGES": "Charges",
             "CLIENT": "Client",
             "CODE": "Code",
+            "PACKAGED": "Conditionnement",
             "CONTRIBUTION_DEDUCTIONS": "Contributions et retenues",
             "CONSULT": "Consulter",
             "CONSUMPTION": "Consommation détaillée",
--- a/client/src/modules/purchases/create/PurchaseOrderForm.js
+++ b/client/src/modules/purchases/create/PurchaseOrderForm.js
@@ -130,6 +130,7 @@ function PurchaseOrderFormService(Inventory, AppCache, Store, Pool, PurchaseOrde
 
         current.quantity = prev.quantity;
         current.unit_price = prev.unit_price;
+        current.package_size = prev.package_size;
       });
 
       // update the store's uuid -> object mapping since we
--- a/client/src/modules/purchases/create/createUpdate.html
+++ b/client/src/modules/purchases/create/createUpdate.html
@@ -283,7 +283,7 @@
                 type="button"
                 ng-click="PurchaseCtrl.optimalPurchase()"
                 id="optimal_purchase"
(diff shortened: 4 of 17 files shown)


lomamech:
How to test
1. Packaging of pharmaceutical products
image

2. allow the management of one product per package
image

3. When establishing the purchase order, mention the *Number Package*, *Pack size* and *Package unit cost*
image

The BHIMA system will continue to calculate the unit cost of the smallest unit as well as the quantity
image

But for the product which have packaging the information on the order form will be for the package and not the unit
image

lomamech:
@mbayopanda this PR is ready for review

jmcameron:
I tested this PR.  It worked, but I have a couple of suggestions.

1. Change the formatting of the 'Units' column from box :(100) to just:   box (100) . I do not think the colon is necessary.

2. I followed the directions as posted, but notice that "Step 1" does not show enabling the stock packaging.   So I left it unenabled and did the rest of test and saving the purchase order failed.   I'll add the error message below.   If packaging is not allowed, we should (1) not allow editing inventory to enable the "is packaged per package" option, and (2) if stock packaging is not enabled we need to ignore the inventory definition "is packaged per package" option even if it is enabled.  That probably means that the database query will need to be updated.   When I did enable the stock packaging, the the PR worked fine.

Here is the error message I encountered when stock packaging was NOT enabled and the inventory definition had the "is packaged per package" option enabled when I tried to save the purchase request:
```
  errors   sqlMessage: "Column 'package_size' cannot be null",
  errors   sqlState: '23000',
  errors   index: 0,
  errors   sql: '\n' +
  errors     '    INSERT INTO purchase_item\n' +
  errors     '      (uuid, inventory_uuid, quantity, unit_price, total, purchase_uuid, package_size)\n' +
  errors     "    VALUES (X'20dd12f45079466ba4eda23dae1f2198', X'b7c0c9ee66ca11eba57e0242ac110002', 1, 1, 1, X'b724ccb86c2a446080a49dd59b3ccaf4', NULL);\n" +
  errors     '  '
  errors } +0ms
```

jmcameron (commented):
I tested this PR.  It worked, but I have a couple of suggestions.

1. Change the formatting of the 'Units' column from `box :(100)` to just:   `box (100)` . I do not think the colon is necessary.

2. I followed the directions as posted, but notice that "Step 1" does not show enabling the stock packaging.   So I left it unenabled and did the rest of test and saving the purchase order failed.   I'll add the error message below.   If packaging is not allowed, we should (1) not allow editing inventory to enable the "is packaged per package" option, and (2) if stock packaging is not enabled we need to ignore the inventory definition "is packaged per package" option even if it is enabled.  That probably means that the database query will need to be updated.   When I did enable the stock packaging, the the PR worked fine.

Here is the error message I encountered when stock packaging was NOT enabled and the inventory definition had the "is packaged per package" option enabled when I tried to save the purchase request:
```
  errors   sqlMessage: "Column 'package_size' cannot be null",
  errors   sqlState: '23000',
  errors   index: 0,
  errors   sql: '\n' +
  errors     '    INSERT INTO purchase_item\n' +
  errors     '      (uuid, inventory_uuid, quantity, unit_price, total, purchase_uuid, package_size)\n' +
  errors     "    VALUES (X'20dd12f45079466ba4eda23dae1f2198', X'b7c0c9ee66ca11eba57e0242ac110002', 1, 1, 1, X'b724ccb86c2a446080a49dd59b3ccaf4', NULL);\n" +
  errors     '  '
  errors } +0ms
```

lomamech:
image

PR #6960, opened 2023-03-06
Add pagination in the stock_movement end-point

mbayopanda:
This PR adds paging in the `stock_movment` end-point.

To perform queries with paging, you have to add: `?paging=true` in the HTTP request, which returns an object with two elements: 
```js
{
  rows: [], // records
  pager: {
    "total": 18, // total of records found in the database
    "page": 4, // the current page, pages are indexed from 1
    "page_size": 5, // the number of records returned by pages
    "page_min": 15, // the minimum counted record of the page
    "page_max": 20, // the maximum counted record of the page
    "page_count": 4 // the total of pages
  } // pagination information
}
```

To test the feature:

- Log into BHIMA (with browser or REST Tools)
- Go to this route: `stock/lots/movements/?paging=true` to have pager information
- Go to this route: `stock/lots/movements/?paging=true&page=2` to have page 2 of records
- Go to this route: `stock/lots/movements/?paging=true&page=2&limit=10` to have page 2 when the number of records on each page must be 10

This feature is enabled for these end-points: 
- `/stock/lots/depots/` : for getting the list of lots with their quantities (see Lots in stock registry)
- `/stock/lots/depotsDetailed/` : for getting the detailed list of lots (see Lots in stock registry)
- `/stock/lots/movements/` : for getting the list of stock movements made (see Stock movements registry)
- `/stock/inventories/depots/` : for getting the list of inventories without consideration of lots (see Article in stock registry)

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -978,25 +978,27 @@ async function getInventoryQuantityAndConsumption(params) {
   const clause = ` GROUP BY l.inventory_uuid, m.depot_uuid ${emptyLotToken} ORDER BY ig.name, i.text `;
 
   const filteredRows = await getLots(sql, params, clause);
-  let _filteredRows = params.paging ? filteredRows.rows : filteredRows;
+  let filteredRowsPaged = params.paging ? filteredRows.rows : filteredRows;
 
-  if (_filteredRows.length === 0) { return []; }
+  if (filteredRowsPaged.length === 0) {
+    return params.paging ? { ...filteredRows } : [];
+  }
 
   const settingsql = `
     SELECT month_average_consumption, average_consumption_algo, min_delay, default_purchase_interval
     FROM stock_setting WHERE enterprise_id = ?
   `;
 
-  const opts = await db.one(settingsql, _filteredRows[0].enterprise_id);
+  const opts = await db.one(settingsql, filteredRowsPaged[0].enterprise_id);
 
   // add the minimum delay to the rows
-  _filteredRows.forEach(row => {
+  filteredRowsPaged.forEach(row => {
     row.min_delay = opts.min_delay;
   });
 
   // add the CMM
-  _filteredRows = await getBulkInventoryCMM(
-    _filteredRows,
+  filteredRowsPaged = await getBulkInventoryCMM(
+    filteredRowsPaged,
     opts.month_average_consumption,
     opts.average_consumption_algo,
     opts.default_purchase_interval,
@@ -1004,14 +1006,14 @@ async function getInventoryQuantityAndConsumption(params) {
   );
 
   if (_status) {
-    _filteredRows = _filteredRows.filter(row => row.status === _status);
+    filteredRowsPaged = filteredRowsPaged.filter(row => row.status === _status);
   }
 
   if (requirePurchaseOrder) {
-    _filteredRows = _filteredRows.filter(row => row.S_Q > 0);
+    filteredRowsPaged = filteredRowsPaged.filter(row => row.S_Q > 0);
   }
 
-  return params.paging ? { ...filteredRows, rows : _filteredRows } : _filteredRows;
+  return params.paging ? { ...filteredRows, rows : filteredRowsPaged } : filteredRowsPaged;
 }
 
 /**
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -340,7 +340,6 @@ async function getAssets(params) {
       params.scan_status === 'scanned' ? 'last_scan.uuid IS NOT NULL' : 'last_scan.uuid IS NULL');
   }
   filters.setGroup(groupByClause);
-
   filters.setHaving(havingClause);
   filters.setOrder('ORDER BY i.code, l.label');
   const query = filters.applyQuery(sql);
@@ -434,7 +433,6 @@ async function getLotsDepot(depotUuid, params, finalClause) {
   `;
 
   const groupByClause = finalClause || ` GROUP BY l.uuid, m.depot_uuid ${emptyLotToken} ORDER BY i.code, l.label `;
-
   const filters = getLotFilters(params);
   filters.setGroup(groupByClause);
 
@@ -484,7 +482,7 @@ async function getLotsDepot(depotUuid, params, finalClause) {
 
   if (params.paging) {
     return {
-      pager : paginatedResults,
+      pager : paginatedResults.pager,
       rows : inventoriesWithLotsProcessed,
     };
   }
@@ -979,24 +977,26 @@ async function getInventoryQuantityAndConsumption(params) {
 
   const clause = ` GROUP BY l.inventory_uuid, m.depot_uuid ${emptyLotToken} ORDER BY ig.name, i.text `;
 
-  let filteredRows = await getLots(sql, params, clause);
-  if (filteredRows.length === 0) { return []; }
+  const filteredRows = await getLots(sql, params, clause);
+  let _filteredRows = params.paging ? filteredRows.rows : filteredRows;
+
+  if (_filteredRows.length === 0) { return []; }
 
   const settingsql = `
     SELECT month_average_consumption, average_consumption_algo, min_delay, default_purchase_interval
     FROM stock_setting WHERE enterprise_id = ?
   `;
 
-  const opts = await db.one(settingsql, filteredRows[0].enterprise_id);
+  const opts = await db.one(settingsql, _filteredRows[0].enterprise_id);
 
   // add the minimum delay to the rows
-  filteredRows.forEach(row => {
+  _filteredRows.forEach(row => {
     row.min_delay = opts.min_delay;
   });
 
   // add the CMM
-  filteredRows = await getBulkInventoryCMM(
-    filteredRows,
+  _filteredRows = await getBulkInventoryCMM(
+    _filteredRows,
     opts.month_average_consumption,
     opts.average_consumption_algo,
     opts.default_purchase_interval,
@@ -1004,14 +1004,14 @@ async function getInventoryQuantityAndConsumption(params) {
   );
 
   if (_status) {
-    filteredRows = filteredRows.filter(row => row.status === _status);
+    _filteredRows = _filteredRows.filter(row => row.status === _status);
   }
 
   if (requirePurchaseOrder) {
-    filteredRows = filteredRows.filter(row => row.S_Q > 0);
+    _filteredRows = _filteredRows.filter(row => row.S_Q > 0);
   }
 
-  return filteredRows;
+  return params.paging ? { ...filteredRows, rows : _filteredRows } : _filteredRows;
 }
 
 /**
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -241,6 +241,13 @@ function getLots(sqlQuery, parameters, finalClause = '', orderBy = '') {
     filters.setOrder(orderBy);
   }
 
+  if (parameters.paging) {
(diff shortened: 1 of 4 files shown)


jmcameron:
Please add a description of how to test this PR.

jniles:
I'm also curious what this is going to be used for.  It seems like it might be better to create it's own route.  I think all API routes in BHIMA that support filtering currently return the maximum amount of records that meet a user's filter criteria, and this breaks that API contract.

mbayopanda:
> I'm also curious what this is going to be used for. It seems like it might be better to create it's own route. I think all API routes in BHIMA that support filtering currently return the maximum amount of records that meet a user's filter criteria, and this breaks that API contract.

@jniles this can be used for server-side pagination, which gives valuable information to the user instead of just loading a limited number of records to not break the client, the pagination can be applied to any list of records (grid, table, etc.) or through API call directly.

Actually, the BHIMA API is huge, so we cannot apply this change to all routes but we can do that progressively without breaking the current behavior of BHIMA routes.

jmcameron (commented):
The code looks okay.  I tried the suggested tests and they seem to work fine.

It is hard to evaluate the utility of this change until it is applied to a report or register page, so I do not have a firm opinion on whether this is a good idea or not.   But I do not object to merging it since I think (guess) it will not affect any existing pages or reports.

jmcameron:
@mbayopanda What do you want to do next on this PR?  Do we need further discussion?

jmcameron:
According to @mbayopanda (on WhatsApp):  Currently there is the BAO System team for their collaboration in the SEMI project they would like to integrate the BHIMA data into their report; they have been using our current API and they have asked me if we can provide the data per page, which will allow not loading all the data at once but as needed. We can merge this PR as long as it doesn't break what's already there.

jmcameron (commented):
I looked at the code again and it seems fine.
I tested it with the implemented endpoints and they work nicely.

I had one question that you might consider:  What should the REST queries return for pages beyond the end of the data?   According to [some best practices](https://learn.microsoft.com/en-us/azure/architecture/best-practices/api-design#get-methods) perhaps we should return a HTML 204 response.   Currently a empty JSON "pager" object is returned.   For instance, calling 

`http://localhost:8080/stock/lots/depots/?paging=true&limit=4&page=4` 

Produces

`{"pager":{"total":18,"page":4,"page_size":4,"page_min":12,
"page_max":16,"page_count":5},"rows":[]}`

It is not clear that the receiving party will be able to interpret this correctly.
Notice that there is no indication that there are only 3 pages with data.  Also, what does the "page_max" mean here?  There are 11 items total.  Perhaps we should discuss this with the BAO folks and see what response would work best for them.

mbayopanda:
> I looked at the code again and it seems fine. I tested it with the implemented endpoints and they work nicely.
> 
> I had one question that you might consider: What should the REST queries return for pages beyond the end of the data? According to [some best practices](https://learn.microsoft.com/en-us/azure/architecture/best-practices/api-design#get-methods) perhaps we should return a HTML 204 response. Currently a empty JSON "pager" object is returned. For instance, calling
> 
> `http://localhost:8080/stock/lots/depots/?paging=true&limit=4&page=4`
> 
> Produces
> 
> `{"pager":{"total":18,"page":4,"page_size":4,"page_min":12, "page_max":16,"page_count":5},"rows":[]}`
> 
> It is not clear that the receiving party will be able to interpret this correctly. Notice that there is no indication that there are only 3 pages with data. Also, what does the "page_max" mean here? There are 11 items total. Perhaps we should discuss this with the BAO folks and see what response would work best for them.

@jmcameron this is a bug with the way we use `SELECT COUNT(*)` we have to use `SELECT COUNT(DISTINCT specific_column)` since the code is used for many context in this context the result is not correct;

A temporary solution is to use naive code and javascript to fix that.

mbayopanda:
@jmcameron could you review this PR please

server/controllers/stock/core.js
```diff
@@ -956,39 +977,41 @@ async function getInventoryQuantityAndConsumption(params) {
...
     SELECT month_average_consumption, average_consumption_algo, min_delay, default_purchase_interval
     FROM stock_setting WHERE enterprise_id = ?
   `;
 
-  const opts = await db.one(settingsql, filteredRows[0].enterprise_id);
+  const opts = await db.one(settingsql, _filteredRows[0].enterprise_id);
 
   // add the minimum delay to the rows
-  filteredRows.forEach(row => {
+  _filteredRows.forEach(row => {
```
jmcameron: I find using the _ prefix not very helpful for making the code easy to understand.  Rather than `_filteredRows`, I suggest using `filteredRowsPaged` or `filteredRowsPaging` (or something similar) which is more descriptive and easy to understand why `_filteredRows` is different from `filteredRows`.

server/controllers/stock/index.js
```diff
@@ -1169,18 +1173,21 @@ async function listLotsDepotDetailed(req, res, next) {
       db.exec(sqlGetMonthlyStockMovements, [db.bid(params.depot_uuid), params.startDate, params.dateTo]),
     ]);
 
-    data.forEach(current => {
+    const _data = !params.paging ? data : data.rows;
+    const _dataPreviousMonth = !params.paging ? dataPreviousMonth : dataPreviousMonth.rows;
```
jmcameron: Again, I suggest calling this `dataPreviousMonthPaged` or something similar to improve code readability.

server/controllers/stock/index.js
```diff
@@ -1169,18 +1173,21 @@ async function listLotsDepotDetailed(req, res, next) {
       db.exec(sqlGetMonthlyStockMovements, [db.bid(params.depot_uuid), params.startDate, params.dateTo]),
     ]);
 
-    data.forEach(current => {
+    const _data = !params.paging ? data : data.rows;
```
jmcameron: I suggest `dataPaged`

server/lib/db/index.js
```diff
@@ -289,6 +289,50 @@ class DatabaseConnector {
...
+      // fetch all data
+      const query = filters.applyQuery(sql.concat(' ', tables));
+      rows = await this.exec(query, queryParameters);
+    } else {
+      // paginated data
+
+      // FIXME: Performance issue, use SQL COUNT in a better way
+      const total = (await this.exec(filters.getAllResultQuery(sql.concat(' ', tables)), queryParameters)).length;
+      const page = params.page ? parseInt(params.page, 10) : 1;
+      const limit = params.limit ? parseInt(params.limit, 10) : 100;
```
jmcameron, line 321: I suggest computing `page_count` here.  And when computing `page_min` and `page_max` below, if `page` > `page_count`, set them to `null' since they are invalid for pages beyond the max.  And apply similar fixes to the FilterParser below.

jmcameron (commented):
I reviewed the code again and made some suggestions for improving variable names, etc.

I tested the end points using bhima_test:
 - `/stock/lots/depots/?paging=true` worked as expected
 - `/stock/lots/depotsDetailed/` failed with this message:   `{"status":500}`
 - `/stock/lots/movements/?paging=true` worked as expected
 - `/stock/inventories/depots/` worked as expected, but when I asked for a page beyond the limit, I got:  `[]` which is a different response than the other endpoints.

**Suggestions**

 - All endpoints should return data in the same format
 - When asking for a page beyond the limit, some endpoints return: 
   - `{"rows":[],"pager":{"total":18,"page":3,"page_size":10,"page_min":20, "page_max":30,"page_count":2}}`
   - I see that `page_count` has been corrected.   Thank you.
   - However, I would prefer to see `page_min` and `page_max` to be `null` or something to indicate that they are not valid for this page beyond the end of the data.  I'm happy to see that `rows = []` in this case, but I think that it would be clearer to whoever is dealing with this data to know that the `page_min` and `page_max` values are invalid in this response.

mbayopanda:
> I reviewed the code again and made some suggestions for improving variable names, etc.
> 
> I tested the end points using bhima_test:
> 
> * `/stock/lots/depots/?paging=true` worked as expected
> * `/stock/lots/depotsDetailed/` failed with this message:   `{"status":500}`
> * `/stock/lots/movements/?paging=true` worked as expected
> * `/stock/inventories/depots/` worked as expected, but when I asked for a page beyond the limit, I got:  `[]` which is a different response than the other endpoints.
> 
> **Suggestions**
> 
> * All endpoints should return data in the same format
> * When asking for a page beyond the limit, some endpoints return:
>   
>   * `{"rows":[],"pager":{"total":18,"page":3,"page_size":10,"page_min":20, "page_max":30,"page_count":2}}`
>   * I see that `page_count` has been corrected.   Thank you.
>   * However, I would prefer to see `page_min` and `page_max` to be `null` or something to indicate that they are not valid for this page beyond the end of the data.  I'm happy to see that `rows = []` in this case, but I think that it would be clearer to whoever is dealing with this data to know that the `page_min` and `page_max` values are invalid in this response.

Thank you @jmcameron for your review

server/lib/db/index.js
mbayopanda: We have to set `page_min` and `page_max` after the execution of the query since the `page_min` is sent to the SQL query in the `OFFSET` part.

mbayopanda:
@jmcameron can I get a review, please?

About this `/stock/lots/depotsDetailed/ failed with this message:  {"status":500}`, the end-point requires a `depot_uuid` in params, it's a wrong definition of a REST endpoint.

server/controllers/stock/core.js
```diff
@@ -956,39 +977,43 @@ async function getInventoryQuantityAndConsumption(params) {
 
   const clause = ` GROUP BY l.inventory_uuid, m.depot_uuid ${emptyLotToken} ORDER BY ig.name, i.text `;
 
-  let filteredRows = await getLots(sql, params, clause);
-  if (filteredRows.length === 0) { return []; }
+  const filteredRows = await getLots(sql, params, clause);
+  let filteredRowsPaged = params.paging ? filteredRows.rows : filteredRows;
```
jmcameron, line 981: 👍

server/lib/db/index.js
```diff
@@ -289,6 +289,57 @@ class DatabaseConnector {
...
+        page_min : (page - 1) * limit,
+        page_max : (page) * limit,
+        page_count : pageCount,
+      };
+      const paginatedQuery = filters.applyPaginationQuery(sql.concat(' ', tables), pager.page_size, pager.page_min);
+      rows = await this.exec(paginatedQuery, queryParameters);
+      if (rows.length === 0) {
+        // update page_min and page_max after the query
+        // in case of empty result
+        pager.page_min = null;
```
jmcameron, line 336: 👍

jmcameron (commented):
Thanks for the code improvements.

I tested the endpoints again:
-   /stock/lots/depots/?paging=true worked as expected
-   /stock/lots/movements/?paging=true worked as expected
-   /stock/inventories/depots/?paging=true worked as expected
-   /stock/lots/depotsDetailed/ seems to be working(?)

Regarding depotsDetailed.   When I use the URL below, I get no results, which I do not understand, since the UUID if for the primary depot.  It is likely that I have the format or data of the URL incorrect.
  - `/stock/lots/depotDetailed/?depot_uuid=F9CAEB16168443C5A6C447DBAC1DF296`
  - Produces:  `[]`

However, when I use paging:
  - `/stock/lots/depotDetailed/?depot_uuid=F9CAEB16168443C5A6C447DBAC1DF296&paging=true`
  - Produces:  `{"pager":{"total":0,"page":1,"page_size":100,"page_min":null,"page_max":null,"page_count":0},"rows":[]}`
  - Which looks like the paging part is working correctly

As far as I'm concerned, I think this is ready to go.

mbayopanda:
Thank you @jmcameron 

bors r+

Issue #6130, opened 2021-12-07
More flexible precision for currencies


jmcameron:
Currently, our reports are forced to use the same precision (number of digits after the decimal separator) for currency for reports regardless of which currency the report is in.  It would be good to make this more flexible.    Here is the suggested approach:
- Extend the 'currency' table with an additional field for the desired number of decimal digits to display.  It may be necessary to add two fields:  one for the number of decimal digits for sums and a different (higher precision) value to use for unit prices.
- Update the HTML/handlebars 'currency' operator to get the number of decimal digits for the specified currency from the database.   These will probably need an optional override of the number of decimal digits in for some situations. 
- In some places, we truncate the values before display using toFloat(), etc.  These will probably need to be revisited to determine if they are still needed.
- Add a new Administration page to support editing the number of decimal digits to display for each supported currency.    It is quite possible that different institutions would prefer to see different numbers of decimal digits (eg, 0 or 1 for Francs, etc).

For reference, please review these potentially related Issues:
- https://github.com/IMA-WorldHealth/bhima/issues/5199
- https://github.com/IMA-WorldHealth/bhima/issues/2933
- https://github.com/IMA-WorldHealth/bhima/issues/3150

jniles:
Thanks for all these issues @jmcameron.  I agree, we need more flexible way to handle currencies.  I also think that we need to unify the rendering of the currencies so that they use a common configuration between the client and server.  For example, we have red negatives in parentheses on the server, but we don't necessarily do that on the client.

If we stored a common configuration file, we would be able to make sure that the rendering was rational across all platforms and reports.

PR #6964, opened 2023-03-07
Stock values4decimal places

### LINKED ISSUE #6951: Displayed stock values with 4 decimal places
- It is difficult in BHIMA to correctly read the stock value of certain products such as $0.0025 will be displayed in BHIMA as zero value $0.00 this poses a problem for stock managers

lomamech:
- Display unit cost with 4 decimal places

closes #6951

### DIFF
--- a/client/src/js/filters/currency.js
+++ b/client/src/js/filters/currency.js
@@ -17,11 +17,10 @@ CurrencyFilter.$inject = [
   'currencyFormat', 'SessionService', '$translate',
 ];
 
-
 function CurrencyFilter(CurrencyFormat, Session, $translate) {
   var requireCurrencyDefinition = false;
 
-  function currencyFilter(amount, currencyId) {
+  function currencyFilter(amount, currencyId, decimalSetting) {
     var formatConfiguration;
     var amountUndefined = angular.isUndefined(amount) || angular === null;
 
@@ -58,7 +57,8 @@ function CurrencyFilter(CurrencyFormat, Session, $translate) {
       return formatError('EXCHANGE.CURRENCY_NOT_SUPPORTED', amount);
     }
 
-    return formatNumber(amount, formatConfiguration.PATTERNS[1], formatConfiguration.GROUP_SEP, formatConfiguration.DECIMAL_SEP)
+    return formatNumber(amount,
+      formatConfiguration.PATTERNS[1], formatConfiguration.GROUP_SEP, formatConfiguration.DECIMAL_SEP, decimalSetting)
       .replace(/\u00A4/g, formatConfiguration.CURRENCY_SYM);
   }
 
@@ -69,7 +69,7 @@ function CurrencyFilter(CurrencyFormat, Session, $translate) {
 
   // Formatting method directly from angular native filter - does not support BHIMA coding guidelines
   var DECIMAL_SEP = '.';
-  function formatNumber(number, pattern, groupSep, decimalSep, fractionSize) {
+  function formatNumber(number, pattern, groupSep, decimalSep, fractionSize, decimalSetting) {
     if (angular.isObject(number)) return '';
 
     var isNegative = number < 0;
@@ -101,6 +101,11 @@ function CurrencyFilter(CurrencyFormat, Session, $translate) {
       // determine fractionSize if it is not specified
       if (angular.isUndefined(fractionSize)) {
         fractionSize = Math.min(Math.max(pattern.minFrac, fractionLen), pattern.maxFrac);
+      }
+
+      // Just to use the number of decimals defined in the parameters
+      if (decimalSetting) {
+        fractionSize = decimalSetting;
       }
 
       // safely round numbers in JS without hitting imprecisions of floating-point arithmetics
--- a/client/src/js/filters/currency.js
+++ b/client/src/js/filters/currency.js
@@ -1,3 +1,5 @@
+/* eslint-disable prefer-template */
+/* eslint-disable no-useless-escape */
 /**
  * BHIMA fork of angular's native currency filter.
  * Allows currency to be defined for each filter individually.
@@ -18,21 +20,19 @@ CurrencyFilter.$inject = [
 ];
 
 function CurrencyFilter(CurrencyFormat, Session, $translate) {
-  var requireCurrencyDefinition = false;
+  const requireCurrencyDefinition = false;
 
-  function currencyFilter(amount, currencyId, decimalSetting) {
-    var formatConfiguration;
-    var amountUndefined = angular.isUndefined(amount) || angular === null;
+  function currencyFilter(amount, currencyId, numDecimals) {
+    const amountUndefined = angular.isUndefined(amount) || angular === null;
 
     if (angular.isUndefined(currencyId)) {
 
       if (requireCurrencyDefinition) {
         return formatError('EXCHANGE.INVALID_CURRENCY_DEFINITION', amount);
-      } else {
-
-        // Display enterprise currency unless otherwise specified
-        currencyId = Session.enterprise.currency_id;
       }
+
+      // eslint-disable-next-line no-param-reassign
+      currencyId = Session.enterprise.currency_id;
     }
 
     // Terminate early to reduce calculations for ill formed requests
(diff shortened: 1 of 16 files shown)


server/controllers/stock/reports/stock_entry_depot.receipt.handlebars
```diff
@@ -55,7 +55,7 @@
   <table class="table table-condensed table-bordered table-report">
     <thead>
       <tr>
-        <th>{{translate 'STOCK.CODE'}}</th>
+        <th>{{translate 'STOCK.CODE'}} BBB</th>
```
jmcameron, line 58: It looks like you still have some debugging code here.

client/src/js/filters/currency.js
```diff
@@ -17,11 +17,10 @@ CurrencyFilter.$inject = [
   'currencyFormat', 'SessionService', '$translate',
 ];
 
-
 function CurrencyFilter(CurrencyFormat, Session, $translate) {
   var requireCurrencyDefinition = false;
 
-  function currencyFilter(amount, currencyId) {
+  function currencyFilter(amount, currencyId, decimalSetting) {
```
jmcameron: Please add an @param comment in the function description above for the new decimalSetting parameter.

jmcameron: I suggest renaming 'decimalSetting' to 'numDecimals' which is more descriptive.

jmcameron (commented):
This looks good (see my review comments).  However, I noticed that this update was not done on several reports:  shipping document, shipping manifest, and stock movement report.   I suggest you go back through and review all the reports to make sure this was not missed anywhere else.

Also, it is not clear to me why we have fractionalSize and 'decimalSetting'.  Why do we need both?  Could this be simplified?

I tested this on a couple of reports and it seems to work fine.

I view this as an interim fix.  Please see Issue https://github.com/IMA-WorldHealth/bhima/issues/6130 for my thoughts on this.

mbayopanda (commented):
Chris, I think this feature must optional since it's not all organizations that will want to have 4 decimals, for example, a hospital with enterprise currency FC; In Fc, there is no need for decimals which will produce huge numbers to read when it is not necessary.

image

lomamech:
@jmcameron  I was not aware of this issue, otherwise I was not going to create this issue #6130 if that was the case I wasn't going to create the #6951

lomamech:
> Chris, I think this feature must optional since it's not all organizations that will want to have 4 decimals, for example, a hospital with enterprise currency FC; In Fc, there is no need for decimals which will produce huge numbers to read when it is not necessary.

Good observation, for that I will have to find a way to take currencies into account.

server/controllers/stock/reports/stock_entry_depot.receipt.handlebars
lomamech: THANKS

lomamech:
@mbayopanda @jmcameron   I have an idea, that of keeping the display of the 2 rows after the decimal point except for amounts less than $0.01 for this kind of amounts we will display 4 digits after the decimal point, how do you find this idea, I wait for your feedback

jmcameron:
> @mbayopanda @jmcameron I have an idea, that of keeping the display of the 2 rows after the decimal point except for amounts less than $0.01 for this kind of amounts we will display 4 digits after the decimal point, how do you find this idea, I wait for your feedback

I like this idea.  However I would also suggest that if the value is exactly 0, there is no reason to use 4 digits.  And this could be done easily in the currency function.   @lomamech What do you think?

lomamech:
@mbayopanda and @jmcameron  I have just updated the PR, by adding the following rule, we can only display 4 digits after the decimal point if the number is less than 0.01 (the smallest value if digital displayed with 2 digits with the decimal point)

image

lomamech:
I'have finish with this PR

client/src/js/filters/currency.js
```diff
@@ -68,26 +68,37 @@ function CurrencyFilter(CurrencyFormat, Session, $translate) {
...
+ * @function formatNumber
+ * @number is the ammount
+ * @numDecimals is the number of decimals allowed
+ */
+  function formatNumber(number, pattern, groupSep, decimalSep, fractionSize, numDecimals) {
+
     if (angular.isObject(number)) return '';
 
-    var isNegative = number < 0;
+    let isNegative = number < 0;
```
jmcameron: A minor comment:  isNegative is not used until line 172.  Therefor it could be computed as a 'const' on that line.

client/src/js/filters/currency.js
```diff
@@ -68,26 +68,37 @@ function CurrencyFilter(CurrencyFormat, Session, $translate) {
...
+    const isInfinity = number === Infinity;
+    // eslint-disable-next-line no-restricted-globals
     if (!isInfinity && !isFinite(number)) return '';
 
-    var numStr = number + '',
-        formatedText = '',
-        hasExponent = false,
-        parts = [];
+    const numStr = `${number} `;
+    let formatedText = '';
```
jmcameron: In English, this is spelled:  formattedText

jmcameron (commented):
I looked over the code and made some OPTIONAL suggestions.   I realize that the function that you edited is old-style and complicated.  So it makes sense not to rewrite it since I doubt we have tests for it that will assure us that a rewritten version works correctly.  So feel free to ignore the coding suggestions I made.

I tested it and it seems to work well.   I did notice one possible problem.  If unit prices are less than 0.0001 (eg 0.000023), then the current algorithm will report 0.0000 (zero).  However I doubt this will occur in reality since per-unit stock prices will probably always be more than that.

So I think this PR is ready to go.

client/src/js/filters/currency.js
lomamech: @jmcameron This file is really very old and it must date from the first version of BHIMA
````
    const isNegative = number < 0;
````
I just made the change

**Comment by lomamech** [comment]
@jmcameron I just finished with the integration of the suggestions

File: server/controllers/stock/reports/stock_adjustment.receipt.handlebars [hunk 5]
```diff
@@ -126,7 +126,7 @@
             <td class="text-right">{{old_quantity}}</td>
             <td class="text-right">{{new_quantity}}</td>
             <td class="text-right">{{difference}}</td>
-            <td class="text-right">{{currency unit_cost ../enterprise.currency_id}}</td>
+            <td class="text-right">{{currency unit_cost ../enterprise.currency_id 4}}</td>
```
jmcameron, line 129: Question: Do you need to insert the '4' here now that you're forcing 4 if the number is too small?

client/src/js/filters/currency.js
```diff
@@ -68,91 +68,110 @@ function CurrencyFilter(CurrencyFormat, Session, $translate) {
...
+    const isInfinity = number === Infinity;
+    // eslint-disable-next-line no-restricted-globals
     if (!isInfinity && !isFinite(number)) return '';
 
-    var numStr = number + '',
-        formatedText = '',
-        hasExponent = false,
-        parts = [];
+    const numStr = `${number} `;
+    let formattedText = '';
```
jmcameron, line 90: 👍

client/src/js/filters/currency.js
```diff
@@ -68,91 +68,110 @@ function CurrencyFilter(CurrencyFormat, Session, $translate) {
...
+    } else if (fractionSize > 0 && number < 1) {
+      formattedText = number.toFixed(fractionSize);
+      // eslint-disable-next-line no-param-reassign
+      number = parseFloat(formattedText);
     }
 
-    if (number === 0) {
-      isNegative = false;
-    }
+    const isNegative = number < 0;
```
jmcameron, line 170: 👍

jmcameron (commented):
The code looks good.  I added one question about the code you might want to consider.

I think this is ready to go if @mbayopanda is happy with it.

server/controllers/stock/reports/stock_adjustment.receipt.handlebars
lomamech: Indeed, you will also notice that the option to display four decimals has only been used in very specific cases in cases where one seeks to display the unit cost of an inventory I had not thought generalized this rule throughout the system unless deemed necessary to apply everywhere.

jmcameron: Sounds good!

PR #7024, opened 2023-04-18
[OBSOLETE] Migrate end-to-end tests to Playwright

jmcameron:
This PR enables end-to-end tests that have been migrated from Protractor to Playwright.  Note that only 23% of the end-to-end have been converted, but it seems appropriate to start using the tests that have been migrated now to avoid any regressions that they might catch.

Note that the old Protractor tests have not been deleted yet, since the obsolete files may still be useful in the migration process.  Also the original command `yarn test:ends` are unchanged so it remains broken.  Use the new command `yarn test:e2e` instead.

Note that a few of these tests still sporadically fail; I'm trying to resolve those problems.

See [Issue 6976](https://github.com/IMA-WorldHealth/bhima/issues/6976) for details about the migration.

### Testing
- Set your .env file for bhima_test
- Run `yarn` to update dependencies (you may want to run `yarn upgrade` too)
- Install playwright:
   - `npm install playwright`
   -  Verify the two versions match:
      - `grep '@playwright/test@' yarn.lock`
      - `npx playwright --version`
- Run the tests:  `yarn test:e2e`


jmcameron:
> Some files could not be reviewed due to errors:
> 
> Oops! Something went wrong! :(
> ESLint: 7.7.0
> ESLint couldn't find the plugin "eslint-plugin-jsdoc".
> (The package "eslint-plugin-jsdoc" was not found when loaded as a Node module from the directory "/tmp/d20230418-5-102ee8n".)
> It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:
>     npm install eslint-plugin-jsdoc@latest --save-dev
> The plugin "eslint-plugin-jsdoc" was referenced from the config file in ".eslintrc".
> If you still can't figure out the problem, please stop by https://eslint.org/chat/help to chat with the team.

Do not worry about this problem.  I'll deal with it later.

jmcameron:
bors try

jmcameron:
bors try

client/src/js/services/NotifyService.js
```diff
@@ -13,7 +13,7 @@ function NotifyService($translate) {
   const service = this;
 
   // default time to live of 3 seconds
```
jniles: You might want to update this comment as well 😄

jmcameron: Thanks. Just pushed the fix.

jniles (approved):
I took the liberty to test this on a fresh install of Debian 12 (KDE) running in VMWare.

Here is what I got as my final result of tests running:

image

Does this match your expected test failures/successes?

If so, I think this should be merged in so we can start refactoring our continuous integration environment to support these tests.  If we are also expecting to add more developers to the team, it would be important to bring this online before they start making meaningful contributions so we can be sure functionality doesn't break.

It would also be helpful for @mbayopanda and myself to use this testing framework to either fix some of the outstanding tests or write some new ones.

Just my 2 💵 .

Great work overall!

jmcameron:
> I took the liberty to test this on a fresh install of Debian 12 (KDE) running in VMWare.
> Here is what I got as my final result of tests running:
> Does this match your expected test failures/successes?


@jniles I disabled the tests that were failing on my system and yours.  Please try the tests again on Debian to see if the tests will all pass.

jmcameron:
bors try

jniles:
bors try

jmcameron:
Replaced by https://github.com/IMA-WorldHealth/bhima/pull/7215

PR #7078, opened 2023-05-17
Fix total calculation in the stock sheet report for exit

mbayopanda:
This PR fixes the `Total` column calculation in the stock sheet report for the exit section.

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -1285,7 +1285,7 @@ async function getInventoryMovements(params) {
       // exit
       movement.exit.quantity = line.quantity;
       movement.exit.unit_cost = stockUnitCost;
-      movement.exit.value = line.quantity * line.unit_cost;
+      movement.exit.value = line.quantity * stockUnitCost;
     } else {
       const newQuantity = line.quantity + stockQuantity;
       // fix negative value disorder


lomamech (commented):
Hello @mbayopanda , is it possible to demonstrate by a use case scenario the errors that have been noticed in this report (in the master branch) and the impact of the modification brought by this PR, because when we do correctly the entries and exits of stock we do not directly notice the error

mbayopanda:
@jmcameron here was the issue: the calculation was not correct

WhatsApp Image 2023-05-17 at 08 58 36

The fixing produces this: 
WhatsApp Image 2023-05-17 at 09 02 53

The thing is we cannot reproduce this error with the test database but the error come from Vanga database, so we have to be sure this fixing do not break any other calculation

server/controllers/stock/core.js
```diff
@@ -1285,7 +1285,7 @@ async function getInventoryMovements(params) {
       // exit
       movement.exit.quantity = line.quantity;
       movement.exit.unit_cost = stockUnitCost;
-      movement.exit.value = line.quantity * line.unit_cost;
+      movement.exit.value = line.quantity * stockUnitCost;
```
jmcameron, line 1288: This change seems consistent with the other calculations of StockValue in the surrounding code.

jmcameron (commented):
This seems like a reasonable fix since it matches the other related calculations in the same code block

I tested this with bhima_test and it seems correct.

I tested this with the downloaded Vanga database and it seems correct.  I tried reverting to the old calculation with the vanga database and did not notice a difference.

However, I'm concerned.  

- The 'entry' values are all done with `line.*` values (core.js lines 1305-1307).   But the 'exit' values are mixed (core.js lines 1286-1288).  Why not make the values used for exit just use `line.*` values similarly?

Questions:
- Why would line.unit_cost ever be different from `openingBalance.unit_cost`?
- Wouldn't different lots purchased at different times potentially have different unit costs?
- Is it possible that the sql data for some movement was modified?

Although I suspect this change would not hurt anything, I think it might be good to understand the deeper cause for this problem.

lomamech:
The problem being solved, we can close this PR

PR #7104, opened 2023-06-01
fix(payroll): remove the "total" prefix from MP

### LINKED ISSUE #7073: Render the total row in the correct currency on the Payroll Multiple Page
The payroll multiple page has two total columns for salaries.  Here they are:

Screenshot 2023-05-16 at 8 54 23 AM

I propose that we render the total row in the same currency of the grid instead of numbers as currently configured.  I also propose that we remove the word "Total" in font of the text.

jniles:
Removes the 'total' prefix from the multiple payroll aggregation grid.

Closes #7073.

image

### DIFF
--- a/test/server-unit/pdf.spec.js
+++ b/test/server-unit/pdf.spec.js
@@ -34,6 +34,8 @@ const data = {
 const fixturesPath = path.resolve('test/fixtures');
 
 function PDFRenderUnitTest() {
+  this.timeout(5000);
+
   it('#pdf.render() renders a valid PDF file', async () => {
     const htmlString = await fs.readFile(template, 'utf8');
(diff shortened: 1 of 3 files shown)


test/server-unit/pdf.spec.js
```diff
@@ -34,6 +34,8 @@ const data = {
 const fixturesPath = path.resolve('test/fixtures');
 
 function PDFRenderUnitTest() {
+  this.timeout(5000);
```
jmcameron, line 37: Good one!

jmcameron (approved):
Looks good.  I tested with bhima_test and it works properly.  The code changes are fine.

PR #7114, opened 2023-06-08
chore: fixup docker file

jniles:
This commit fixes up the docker and compose files for distributing BHIMA.  It gets BHIMA building using `podman` or `docker` on a system without mysql and node installed.  Note that this only supports the x86_64 architecture at the moment.

If you want to test this for yourself:
  1. Check out this branch 
  2. Create a `.env` with credentials if you don't have them already.
  3. Run `./sh/docker-init.sh`.
  4. Run either `docker compose up -d` or `podman-compose up -d`.  

The rest should be automatically downloaded and installed.

----

This isn't completely working yet, due to how BHIMA packages the database.  For this to work, the user needs to have the file generated by "./sh/docker-init.sh" in their `/temp` folder, which somewhat defeats the purpose.  It could be that we just include this as a pre-requisite step (as done in the above instructions).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,16 +2,16 @@
 node_modules
 .git
 
-# ignore loggin utilities
-npm-debug.log
-yarn-error.log
+# ignore logging utilities
+*.log
 
 # ignore documentation
 *.md
+Dockerfile
 
 # ignore unnecessary folders
 docs
-temp
+#temp
 test
 sh
 bin
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # define base image
-FROM node:lts-slim
+FROM node:lts-slim as build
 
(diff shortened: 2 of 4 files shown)


jniles:
So, I think to make this work effectively, we will need to standardize the way we do deployments.  I think we can just do the following:

  1. Because we can use docker as our deployment mechanism, we do not need to have separate database names.  Instead, we can just use `bhima` as the standard database name, no matter where the installation site is.
  2. We can include the full schema file (including all migrations) as part of the release.
  3. We can include the `docker-compose` file as part of the release as well.  This will read the database schema and compile it into the `/docker-entrypoint-initdb.d` folder of the docker-file.

jniles:
I'm not sure I'll have time to continue pursing this at this point.  Closing to unclog the PR queue.

PR #7130, opened 2023-06-29
Fix stock sheet reason

mbayopanda:
This PR fixes the reason in the stock sheet report.


server/controllers/stock/core.js
```diff
@@ -1226,6 +1226,7 @@ async function getInventoryMovements(params) {
       JOIN document_map dm ON dm.uuid = m.document_uuid
       LEFT JOIN entity_map em ON em.uuid = m.entity_uuid
       LEFT JOIN document_map dm2 ON dm2.uuid = m.entity_uuid
+      LEFT JOIN depot d2 ON d2.uuid = m.entity_uuid
```
jmcameron, line 1229: 👍

jmcameron (commented):
The code looks good.

I tested this with bhima_test and was able to verify the 'Reason' was changed appropriately (for depot-to-depot transfers).

BUT, there is a problem because it appears that the yarn.lock file has been deleted from this branch.  That needs to be fixed before this can be released.

PR #7148, opened 2023-07-14
Updated survey form

jmcameron:
Updated the survey creation/edit form to disable the submit button if there is an error in the name or a element type has not been selected.

Note that adding the ng-blur option to the 'name' field was necessary because filling the field in the End-to-End tests did not activate the 'ng-keyup' callback.

I did not test this very much.  Please verify that it works correctly for both creating and editing forms.

### DIFF
--- a/client/src/modules/survey_form/modals/survey_form.modals.html
+++ b/client/src/modules/survey_form/modals/survey_form.modals.html
@@ -84,7 +84,9 @@
           </span>
         </label>
       </div> -->
+
     </div>
+
     <!-- rank -->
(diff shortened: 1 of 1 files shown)


lomamech (commented):
image

I have just recreated the branch and I see that the submit button is disabled as long as the selection of the type and the name of the variable are not entered. I tried and it works fine.

Let me know if I can already merge this PR

lomamech:
I wait your FeedBack

jmcameron:
That sounds good. Please go ahead and release it.

PR #7160, opened 2023-07-23
chore(reports): remove stock changes report

### LINKED ISSUE #7103: Do we still need the stock changes report?
By default, BHIMA ships with a "stock changes" report that was a hack for recording the stock changes offline using a paper based report.  However, now we are using a mobile application for recording the stock changes offline.

Can we remove this report?

jniles:
This commit removes the stock_changes report, useful only for manual stock calculations and never implemented in practice.  The same functionality (keeping track of stock offline) has been implemented in the BHIMA Stock Android application.

Closes #7103.


jmcameron (commented):
The changes look fine.  I'm happy to merge this PR.  But I would like Bruce to give me a second opinion since I'm not familiar with this report or how it was used.

jmcameron (approved):
The code changes look good.

I tested this with bhima_test and it seems to work fine.  I did not test the migration SQL file.

PR #7189, opened 2023-08-10
chore(payroll): fix typo "payrall"

jniles:
A simple commit to fix the "payrall" typo.

### DIFF
--- a/server/controllers/payroll/configuration/index.js
+++ b/server/controllers/payroll/configuration/index.js
@@ -10,7 +10,7 @@ const db = require('../../../lib/db');
 // GET /PAYROLL_CONFIG
 function lookupPayrollConfig(id) {
   const sql = `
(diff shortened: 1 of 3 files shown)


jmcameron (commented):
The code changes look fine, but I have a question:  Should we be changing old SQL migration files? 
- The changes in `server/models/migrations/v1.14.1-v1.15.0/migrate.sql` seem fine since we are adding a new constraint
- I was not sure about the changes to `server/models/migrations/v1.5.0-v1.6.0/migrate.sql` 

Please review these changes to the SQL migration files and verify that they changes make sense.

I tested this and it works fine.

I'm ready to merge this once I understand whether the old SQL migration files should be updated.

jniles:
@jmcameron I have removed that file from the PR.

PR #7195, opened 2023-08-13
deps: remove body-parser

### LINKED ISSUE #7124: Add "Service" column to the Multiple Payroll Registry
I received the following email from our friends at Vanga:

> On the report Payroll Payroll Multiple(indice) would you be able to add a column called service?  This is the column that appears on the report Payroll Registre des employés.  It would be helpful to be able to sort employees by service and have each head of service validate that everyone getting a paycheck is in the correct service and is actually working.  (There is a rumor that there are some workers that get paychecks that are not actually working here.)  I doubt this but this is a way to double check. 

It seems simple enough - we might already have that information available to add to the Multiple Payroll Registry.

### LINKED ISSUE #7193: Remove `body-parser` in favor of `express.json()`
We use `body-parser` in our stack to parse the form bodies of incoming messages.  See here:

https://github.com/IMA-WorldHealth/bhima/blob/45231669186c2e3b0da4cfd24b194d527178e2a5/server/config/express.js#L9C14-L9C14

However, this functionality _may_ have been implemented into `expressjs` itself from v4.16.0 onwards.  See [this documentation](https://expressjs.com/en/api.html#express.json).  If so, we should investigate if we can replace the dependency on body-parser and further cut down our dependencies.

jniles:
The functionality of body-parser we were using was merged into expressjs.  This commit removes body-parser and replaces it with expressjs's version.

Closes #7193.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,6 @@
     "angular-translate-loader-url": "^2.19.0",
     "angular-ui-bootstrap": "^2.5.6",
     "angular-ui-grid": "^4.12.3",
-    "body-parser": "^1.20.2",
     "bootstrap": "^3.3.0",
     "chart.js": "^3.9.1",
     "chartjs-plugin-datalabels": "^2.1.0",
--- a/server/config/express.js
+++ b/server/config/express.js
@@ -6,7 +6,6 @@
  */
 
 const express = require('express');
-const bodyParser = require('body-parser');
 const session = require('express-session');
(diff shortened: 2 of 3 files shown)


jmcameron (approved):
I tested this and it works fine.   It does remove body-parser from package.json but there are still dependencies in yarn.lock for body-parser (eg, express and karma).  So, I'm not sure how much this really helps.

PR #7196, opened 2023-08-13
chore(procedures): merge procedures into one file

### LINKED ISSUE #7188: Merge all stored procedures from server/models/procedures/*.sql into server/models/procedures.sql
The current way we build the database is by building all the files in server/models/*.sql.  This includes procedures.sql, which uses the SQL `source` keyword to pull in files in a certain order from `server/models/procedures/*.sql`.  However, this `source` keyword requires a relative path from the root of the repository, and so if the repository structure changes (such as when you build this into a minimal docker image), it breaks.  

I propose we manually replace the `source`ing of the files with the contents of the files to overcome this limitation.  Instead, we would have one massive `procedures.sql` file, similar to how we have one massive `schema.sql` file, that would contain all the code for our stored procedures.

This solution would (1) overcome the limitations around using `source` in our SQL files, (2) make it easier to find stored procedures (they are all in one file) and (3) maintain execution order in the `procedures.sql` file.  None of our current build processes would need to change, since the "API" of calling `procedures.sql` would remain the same.

The downside is that we would have one massive `procedures.sql` file, but that seems like a reasonable trade-off.

jniles:
This commit merges all the procedures into a single file, and deletes the old procedures/*.sql files.  It also removes references to the different procedure files in our build steps.

Closes #7188.

This PR does not necessarily mean we need to go this direction the project maintainers do not wish to.  It would just make things a bit easier when it comes to builds and deployment (though maybe increase in difficulty to maintenance).


jmcameron (changes requested):
The code changes look fine.  I tested it and it works.

However, please update the procedures.sql to have a separator (eg, horizontal bar) between each "file" in the procedures.sql indicating the nature of the following SQL code.  For instance, if you look at line 690 or 1957 it is not clear what the description applies to.  When it was in a separate file, it was obvious.  But now that it is part of single file, it would be helpful to add a little header for each section with a separating line and a brief title about what the following section is about (possibly just a simple adaptation of the original SQL file name..)

Also, does anything need to be done in the SQL migration file?

jniles:
Nothing needs to be done with the migration files.

I'll add in the requested changes this week. 👍 .  I agree it would make it nicer.

jniles:
@jmcameron I believe the changes have been made!

jmcameron (approved):
Thanks for the updates.

PR #7198, opened 2023-08-13
feat(payroll): add service to multiple payroll

### LINKED ISSUE #7124: Add "Service" column to the Multiple Payroll Registry

jniles:
Adds the service column to the multiple payroll to allow managers to see which services have paid their staff on time.  It also adds it to the multiple payroll index page.

_Multiple Payroll_
image

_Multiple Payroll Index_
image

Closes #7124.

### DIFF
--- a/client/src/modules/multiple_payroll/multiple_payroll.js
+++ b/client/src/modules/multiple_payroll/multiple_payroll.js
@@ -48,6 +48,10 @@ function MultiplePayrollController(
     displayName : 'FORM.LABELS.EMPLOYEE_NAME',
     headerCellFilter : 'translate',
   }, {
+    field : 'service_name',
+    displayName : 'FORM.LABELS.SERVICE',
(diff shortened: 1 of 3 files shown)


jmcameron (approved):
Code changes look good.
I tested this (sort-of) and it seems to work.
I also did  a test merge with PR 7159 (HGR payroll) and there were no merge problems so it does not matter whether this is merged before the HGR payroll PR.

PR #7205, opened 2023-08-16
Inventory usage REST API

jmcameron:
Create a REST API to get the lots usage schedule for a particular inventory item.

NOTE: Do not merge until we have discussions with the end-users for this REST API.

The API is:   /inventory/:uuid/schedule/:depotUuid

where the `uuid` is the UUID for the inventory item.  The GET query also needs to provide these query items:
   - (optional) month_average_consumption
   - (optional) average_consumption_algo (which AMC algorithm to use, eg:  'algo_msg' or 'algo_def'
   - interval_num_months (the interval/window of time to consider, starting today)

Note: We currently require the user to provide the month_average_consumption but this can probably be computed from the inventory article.


mbayopanda (approved):
LGTM 👍🏽 

bors r+

PR #7227, opened 2023-08-25
feature(Set Is Medical from Employee Jobtitle)

### LINKED ISSUE #7170: Move `is_medical` property from employee to title
Based on  https://github.com/IMA-WorldHealth/bhima/pull/7159#discussion_r1271467056

> Can we move the is_medical to the title_employee table? It seems like is_medical has to do with the employee's job title, not the employee themselves, right?

This issue to be closed by removing the property from the employee table and putting it on the profession, title, or some other table where it is appropriate.

lomamech:
- Remove property is_medical from Employee
- Add property is_medical in table title_employee
- Update modules 

Closes #7170.


jmcameron (commented):
The code changes look okay.

I tested it with bhima_test and would like two fixes:
 - In the employee registry when I try to do a search by "Job Title", it does not work.
 - Please update the employee registry to it remembers the column selections.  Currently it forgets the column selections every time you leave the page.

I did not test this by doing a payroll. I encourage you to verify that the payroll process works correctly with this change.

jmcameron (approved):
Chris, thanks for your updates.  I tested them and they work okay.

PR #7229, opened 2023-08-26
Remove protractor tests

### LINKED ISSUE #7218: Clean out old Protractor E2E tests
With the merger of the new Playwright End-to-End tests, there is no longer any reason to keep the files related to the old Protractor E2E tests.  They should be deleted and dependencies on Protractor removed (as much as possible).  

Rename the 'test/end-to-end-playwright' to 'test/end-to-end' and update the tests scripts as necessary.

Note that essentially every old E2E test was migrated.  I some cases (such as https://github.com/IMA-WorldHealth/bhima/issues/6976) the tests were only partially completed and need to be reconstructed because of changes in the BHIMA code.  So any tests that are currently missing, will need to be recreated from scratch in any case.

jniles:
This PR wraps up the saga of protractor as our end to end test
framework.  It removes all the protractor code in favor of playwright.
Thanks to @jmcameron for all his hard work to make this possible.

Closes https://github.com/IMA-WorldHealth/bhima/issues/7218.


mbayopanda:
@jniles can you resolve conflicts so that I can merge this PR? Thanks

jmcameron (approved):
The changes look good.  I've tested it and it seems to work correctly.

PR #7256, opened 2023-09-15
Update setup-migration-script.sh

jniles:
This commit fixes the migration script to solve the break introduced by the stored procedure rework.

### DIFF
--- a/server/models/migrations/next/migrate.sql
+++ b/server/models/migrations/next/migrate.sql
@@ -1,9 +1,10 @@
 /* v1.28.x */
-
+DELETE FROM role_unit WHERE unit_id = (SELECT id FROM unit WHERE `path` = '/admin/odk-settings');
 DELETE FROM unit WHERE `path` = '/admin/odk-settings';
 DROP TABLE IF EXISTS `odk_central_integration`;
 
 -- remove references to stock_changes report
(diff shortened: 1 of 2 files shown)


jmcameron (changes requested):
Sorry to report that this was not enough. When trying to do a migration, I fetch the DB, build the DB, do 'npm run migrate' and then try to load the migration file and get this error:

```ERROR 1451 (23000) at line 5033: Cannot delete or update a parent row: a foreign key constraint fails (`semi`.`role_unit`, CONSTRAINT `role_unit__unit` FOREIGN KEY (`unit_id`) REFERENCES `unit` (`id`) ON UPDATE CASCADE)```

I'll attach a prior version of the migration file (from before this PR) that has lines commented out that eventually let me use it.

[migration-semi.sql.txt](https://github.com/IMA-WorldHealth/bhima/files/12645091/migration-semi.sql.txt)

jniles:
@jmcameron , as discussed on WhatsApp, this has more to do with the migration scripts than the shell file here.  I ran into a similar thing while testing with the HEV database.

It might be a quick fix (something missing from migrate.sql) or it might be a longer fix (something changed in MySQL's handing of FKs/triggers).  I don't have the bandwidth for this at present, so would be happy for someone else to look into it.

jmcameron:
Updated the next/migration.sql file to get this working.

PR #7262, opened 2023-09-20
Expose the RUMER data through the API

mbayopanda:
This PR is about the expose the route : `/api/stock/rumer`

This route requires as parameters
- depotUuid: the depot uuid
- start_date: the begining date of a given month
- end_date: the end date of a given month

This API route returns data like this: 
image

Where data `configurationData` contains information about stock of each inventory concerned: 
image

And `dailyConsumption` contains for each days of the given period, information about stock entry, exit and balance: 
image

Example:
GET <SERVER_ADDRESS>/api/stock/rumer?depotUuid=0918809809BA11ED9604061E16E213FC&start_date=2023-09-01&end_date=2023-09-30

image

The corresponding report in BHIMA is:
image

image


server/config/routes.js
```diff
@@ -909,6 +909,9 @@ exports.configure = function configure(app) {
   app.get('/reports/stock/assign', stockReports.stockAssignReport);
   app.get('/reports/stock/satisfaction_rate_report', stockReports.satisfactionRateReport);
 
+  // stock api for inter-operability
+  app.get('/api/stock/rumer', stockReports.rumerApi.getData);
```
jmcameron: It is good to add a route for this, but I suggest removing the /api/ part of the URL.   None of our other REST urls use it and it does not seem necessary.

jmcameron (commented):
I looked at the code and it looks fine.  Very nice to have the key functionality moved into a dedicated function so it can be used for the REST API and the Rumer report.

I tested it with bhima_test and was able to get it to work with the following URL:

```
http://localhost:8080/api/stock/rumer?depotUuid=F9CAEB16168443C5A6C447DBAC1DF296&start_date=2023-09-01&end_date=2023-09-31
```

Here are some changes I suggest:
   - Remove the /api/ from the REST URL for this query
   - Add an integration regression test for the new URL.  For an example, see `test/integration/inventory/lot_schedule.js` in [PR 7205 ](https://github.com/IMA-WorldHealth/bhima/pull/7205).

server/config/routes.js
mbayopanda: @jmcameron I am okay with your proposal to remove the part `/api` in the URL, do you have some suggestions since the route returns only data to be consumed by tiers clients?

jmcameron: Will `/stock/rumer ` work?

jniles: I would approach this from a different angle.  Instead of thinking of this as a separate API, I have always thought of every BHIMA route as part of the BHIMA API, and we have been trying to follow REST conventions for the most part.

From a REST perspective, I would ask "what is the required parameter?"  Since it is a `depotUuid` in this case, I would put that in the URL and call this `depots/:uuid/rumer`.  That way, we can think about is as "this is the RUMER report for the depot identified by `depotUuid`".

server/models/migrations/next/migrate.sql
```diff
@@ -20,11 +24,11 @@ CREATE TABLE `title_employee` (
   UNIQUE KEY `title_1` (`title_txt`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
-INSERT INTO unit VALUES
+INSERT IGNORE INTO unit VALUES
   (318, 'Job Titles Management','TREE.TITLE','',57, '/titles');
 
 CALL add_column_if_missing('employee', 'title_employee_id', 'TINYINT(1) NOT NULL DEFAULT 0');
-ALTER TABLE `employee` ADD CONSTRAINT `employee__title_employee` FOREIGN KEY (`title_employee_id`) REFERENCES `title_employee` (`id`);
+-- ALTER TABLE `employee` ADD CONSTRAINT `employee__title_employee` FOREIGN KEY (`title_employee_id`) REFERENCES `title_employee` (`id`);
```
jniles: I think this might be important later.. so we might at least want to add a TODO about trying to fix it during the next release.  I know it was one of the last changes @lomamech made.

server/models/migrations/next/migrate.sql
```diff
@@ -1,5 +1,9 @@
 /* v1.28.x */
 
+-- Fix the constraint fails
+DELETE ru FROM role_unit ru JOIN unit u ON u.id = ru.unit_id WHERE u.`path` = "/admin/odk-settings";
```
jniles: Great catch!

server/controllers/stock/functions/rumer.function.js
```diff
@@ -0,0 +1,192 @@
...
+  params.condensed_report = parseInt(params.condensed_report, 10);
+
+  const data = {};
+  const headerReport = [];
+  const configurationData = [];
+
+  params.start_date = moment(new Date(params.start_date)).format('YYYY-MM-DD');
+  params.end_date = moment(new Date(params.end_date)).format('YYYY-MM-DD');
+
+  const startDate = parseInt(moment(params.start_date).format('DD'), 10);
```
jniles, line 38: This is a bit complicated to read.  I propose you use `moment.diff()` to make it a bit simpler since you are using moment anyway.  Something like:

```js
const numDays = moment(params.end_date).diff(params.start_date, 'days');
for (let i = 0; i <= numDays; i++) {
  headerReport.push(i);
}
```

Or something similar (you might have to add or subtract 1 from `i`).

jniles (commented):
I don't have a lot of time to review the code changes, but I've done what I can.

Overall, I really like the direction of this PR - it reuses the same functions for multiple report and I think that is excellent.  I agree with @jmcameron's review that it should have some integration tests.

Great work!

server/models/migrations/next/migrate.sql
mbayopanda: Thank you @jniles

server/config/routes.js
mbayopanda: @jniles the function used in the API route is also used elsewhere so we do not have the freedom to make this change on the endpoint route definition without modifying other modules.

mbayopanda:
@jmcameron could you review my PR? Please

jmcameron:
> @jmcameron could you review my PR? Please

I'm testing it now.

server/controllers/stock/functions/rumer.function.js
```diff
@@ -0,0 +1,193 @@
...
+ * @method getRumer
+ *
+ * @description
+ * This method builds the RUMER (Registre d’Utilisation des Médicaments Et Recettes) report by month
+ *
+ * GET /stock/rumer
+ */
+async function getData(reqQuery) {
+  const params = reqQuery;
+  params.depotUuid = params.depotUuid || params.depot_uuid;
```
jmcameron, line 26: I suggest choosing one depot UUID parameter name or the other.  Since the dates are line `start_date`, I suggest allowing only `depot_uuid`.   Don't forget to modify the PR description with the updated parameter name.

server/controllers/stock/functions/rumer.function.js
```diff
@@ -0,0 +1,193 @@
...
+
+  for (let i = startDate; i <= endDate; i++) {
+    headerReport.push(i);
+  }
+
+  data.params = params;
+  data.header = headerReport;
+
+  const parameterOpeningStock = {
+    depot_uuid : params.depotUuid,
```
jmcameron, line 51: Switch to `params.depot_uuid`

server/controllers/stock/functions/rumer.function.js
```diff
@@ -0,0 +1,193 @@
...
+    dateTo : lastDayPreviousMonth,
+    includeEmptyLot : 1,
+  };
+
+  const parameterEndingStock = {
+    depot_uuid : params.depotUuid,
```
jmcameron, line 57: `params.depot_uuid`

server/models/migrations/next/migrate.sql
```diff
@@ -21,7 +21,7 @@ CREATE TABLE `title_employee` (
   UNIQUE KEY `title_1` (`title_txt`)
 ) ENGINE=InnoDB DEFAULT CHARACTER SET = utf8mb4 DEFAULT COLLATE = utf8mb4_unicode_ci;
 
-INSERT INTO unit VALUES
+INSERT IGNORE INTO unit VALUES
```
jmcameron, line 24: 👍

test/integration/stock/rumer.js
```diff
@@ -0,0 +1,89 @@
+/* global expect, agent */
```
jmcameron, line 1: Thanks for adding this!

jmcameron (commented):
Unless you have a strong reason for allowing both `depotUuid` and `depot_uuid`, I encourage you to use only `depot_uuid` to avoid confusion and simplify the code.

I tested with all the tests (including E2E tests) with bhima_test and it worked fine.  I also tested this API manually with a large production database and it seems to work correctly.

Other than the minor issue of the depot uuid parameter name, I think this is ready to merge.

server/controllers/stock/functions/rumer.function.js
mbayopanda: The function is already used by the RUMER report module, so I do not want to change currently codes on the client or some existing modules

PR #7273, opened 2023-09-26
Speed up /stock/lots/depot query

jmcameron:
Refactored the filtering for user permissions speed up the lot queries.  The lots routes that use these function `getLots()`, `getAssets()`, and `getLotsDepot()` were using the following filter definition (in `getLotFilters()`):

```
// depot permission check
filters.custom(
  'check_user_id',
  `d.uuid IN (
    SELECT DISTINCT d.depot_uuid
      FROM (
        SELECT dp.depot_uuid, dp.user_id FROM depot_permission AS dp
        UNION
        SELECT ds.depot_uuid, ds.user_id FROM depot_supervision AS ds
      ) AS d
    WHERE d.user_id = ?
  )`,
);
```
This means that this 'SELECT DISTINCT' query and subquery were being executed for each row of the results.  This is the primary cause of the slow performance of the `/stock/lots/depot` endpoint with large databases.

This PR changes the above filter to run the 'SELECT DISTINCT' query once and then encodes all of the allowable depots in a new custom filter, so that each row of the results is simply doing checking to see if the depot UUID is a predefined list of UUIDs, which is much more efficient than multiple subqueries.

*TESTING*
  - Use a large production database
  - First, try this in 'master': check the time required for: Stock > Lots 
  - Switch to this PR and try it again and notice the speed up.\
  - Verify it works with paging, use the following API endpoint:
     `/stock/lots/depots?includeEmptyLot=1&limit=1000&period=allTime&paging=true&page=2`

NOTE: This could be generalized by extending `filterParser` with an additional filter that would do a query first and then define a custom filter based on the output.

### DIFF
--- a/server/controllers/stock/core.js
+++ b/server/controllers/stock/core.js
@@ -60,14 +60,13 @@ module.exports = {
 };
 
 /**
- * @function getLotFilters
+ * getLotFilters
  *
- * @description
  * Groups all filtering functionality used in the different getLots* functions into
  * a single function.  The filterparser is returned so that any additional modifications
  * can be made in the function before execution.
  *
- * @param {Object} parameters - an object of filter params.
+ * @param {object} parameters - an object of filter params.
  */
 
 function getLotFilters(parameters) {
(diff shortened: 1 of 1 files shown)


server/controllers/stock/core.js
```diff
@@ -196,6 +179,33 @@ function getLotFilters(parameters) {
...
+        FROM depot_supervision AS ds
+      ) AS d
+      WHERE d.user_id = ?;
+    `;
+    const results = await db.exec(psql, [userId]);
+    const uuids = results.map(item => `'${item.uuid}'`);
+    filters.custom('check_user_id', `BUID(d.uuid) IN (${uuids})`);
+  }
+}
+
```
mbayopanda, line 208: 👍🏽

mbayopanda (approved):
LGTM 👍🏽

PR #7302, opened 2023-10-18
Add the enable external access flag to users

### LINKED ISSUE #7287: Add an "enable external access" field to user table
We have several sites that have historically asked for external access to their databases: notably, IMCK, HEV, and Bandunduville.  We hacked together a solution by building the database on a VM, disabling all BHIMA users, creating an IMA specific BHIMA user, and then sharing that user with the users who need external access.  That worked in the short run.

Here are the problems with that approach:
  1. We need to maintain scripts that create IMA users for each database that we synchronize and build.  Often, this is the same user between each site, with the same password.  That means, the MD of Bandunduville could get into the database of IMCK.
  2. By default, the IMA user is a super user.  To give access to a user with more restrictive roles, we have to add more lines to a shell script.
  3. If the MD of a hospital quits/is fired, they still have access to the institution's data online unless we update the password.
  4. We have to remember what the password was for each site.

To improve this, I propose we add a boolean column to the `user` table called `enable_external_access` (default: `false`).  This would be set from the user edit page, and would allow us to make our build scripts simply deactivate all users without `enable_external_access` being true.  Then, a user at any of the sites wishing external access would turn on the "enable external access" boolean, and they would have their shared account and shared password across the local instance and the global instance.

There are some downsides to this idea:
  1. Users may have really weak passwords, and expose that vulnerability on a globally available BHIMA instance.  However, I am comfortable with this trade off - if our IMA user password gets compromised, all our sites are compromised, which is arguably worse.
  2. Users may be confused that their changes on the "global" instance are not synchronized to the local instance.  However, we can solve this by simply adding a banner (https://github.com/IMA-WorldHealth/bhima/issues/5031) to the global instance informing the users that they are on the global instance.

jniles:
Implements the `enable_external_access` flag on the `user` table.  It is modifiable on the user edit modal.  There is not functionality linked in BHIMA for this flag, rather, it is designed to allow our backup scripts to determine which user accounts to lock and which to leave open.  See #7287 for a full explanation.

To test:
 1. Log in to the test database.
 2. Use the "action" dropdown menu to edit a user.
 3. Click the checkbox on the bottom of the user modal and confirm it updates when saving and editing.

Screenshots:
Screenshot 2023-10-17 at 8 57 15 PM

@mbayopanda, this French translation could probably be cleaned up a lot 🤷 😆 

Closes #7287.

### DIFF
--- a/client/src/i18n/en/users.json
+++ b/client/src/i18n/en/users.json
@@ -6,6 +6,8 @@
     "CREATE" : "Create",
     "CREATED" : "Created",
     "DESCRIPTION" : "User management unit and rights",
+    "ENABLE_EXTERNAL_ACCESS" : "Enable External Access",
+    "ENABLE_EXTERNAL_ACCESS_DESCRIPTION" : "Unlocks the user account on external sites that mirror the hospital database.  The user will be able to log in with their username and password on mirrored sites.",
     "PASSWORD_MISMATCH" : "Passwords entered are not the same",
     "PASSWORDPOLICY" : "To conform with our Strong Password policy, the password must be more than 7 characters. It must include a special character (e.g. @), one or more numerical digits, and both uppercase and lowercase letters.",
     "PERMISSIONS" : "Permissions",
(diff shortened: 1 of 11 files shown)


jmcameron (approved):
The code looks good. 
I tested this and it works correctly on the on a dev system and the regression tests pass.

PR #7303, opened 2023-10-18
fix(lots): remove incorrect header

### LINKED ISSUE #7294: The headers on stock lots "unit" column are unnecessary
On the Stock Lots page, there are column headers for the "units" column.  They shouldn't be there.

Screenshot 2023-10-09 at 4 03 42 PM

jniles:
This commit removes the unnecessary and incorrect header on the stock lots registry.

Closes #7294.

Screenshot:
Screenshot 2023-10-17 at 9 27 08 PM

### DIFF
--- a/client/src/modules/stock/inventories/templates/unit.tmpl.html
+++ b/client/src/modules/stock/inventories/templates/unit.tmpl.html
@@ -1,1 +1,1 @@
-<div class="ui-grid-cell-contents" translate>{{ row.entity.unit_type }}</div>
+<div class="ui-grid-cell-contents" ng-if="!row.groupHeader" translate>{{ row.entity.unit_type }}</div>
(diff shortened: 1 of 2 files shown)


client/src/modules/stock/inventories/templates/unit.tmpl.html
```diff
@@ -1 +1 @@
-<div class="ui-grid-cell-contents" translate>{{ row.entity.unit_type }}</div>
\ No newline at end of file
+<div class="ui-grid-cell-contents" ng-if="!row.groupHeader" translate>{{ row.entity.unit_type }}</div>
```
jmcameron, line 1: 👍

jmcameron (approved):
The code changes look good.
I did not notice this problem on master with bhima_test, but the change seems appropriate.

PR #7318, opened 2023-10-30
Fix json 2 csv

jmcameron:
Includes fix to support new version of json-2-csv.  Now csv.render() is synchronous and returns the data, not a promise.  Async calls still work but test that checks that the result is a promise was removed.


jmcameron:
Bors r+

PR #7321, opened 2023-11-01
bug(Payslip Generator with index)

### LINKED ISSUE #7320: Error when displaying several pay slips in the system with Index
When viewing several pay slips in the system with Index, certain elements of the first employee are reflected on all employees. This issue aims to correct the problem.

lomamech:
- Added data filtering on the allocation of incident data for each employee individually

closes #7320


lomamech:
This bug was slipped due to the fact that during implementation, we only did the test for a single employee
image

jmcameron (approved):
Looks good.

PR #7331, opened 2023-11-09
Fix a few bugs

jmcameron:
In the process of developing the Budget module, a couple of pre-existing bugs were discovered and fixed.  In order to make these more explicit and easier to track, these bugs are corrected in this PR.  There are two bugs that are fixed:

- CSV exports now are generated in a format that allows editors to recognize that they are UTF-8 files containing UTF-8 characters, accents, etc.
- There was a bug in generating Shipments reports

**TESTING**
- Use bhima_test
- Use the master branch
   - Go to the `Finance > Account Managment` page
     - In the drop-down [Menu] on the toolbar, click on "Download as CSV"
     - Open the file in an text editor and notice that some of the account labels have strange characters
   - Go to the `Shipments > Shipment Registry` page
     - If there are no shipments, create one (Using the 'New Shipment' page.  Note that available stock can be found on the `Stock > Articles in Stock` page.
     - In the drop-down [Menu] on the toolbar, click on any of the "Download as ..." options. Notice that no file is generated.  If you examine the server-side console, you will see error messages.
- Switch to the branch for this PR.  Repeat the tests above.
   - Now when you open an exported CSV file, the UTF-8 characters are retained
   - You can now generate shipment reports

### DIFF
--- a/server/lib/renderers/csv.js
+++ b/server/lib/renderers/csv.js
@@ -40,16 +40,20 @@ exports.render = renderCSV;
 exports.headers = headers;
 
 /**
- *
- * @param {Object} data     Object of keys/values for the data
- * @param {String} template Path to a handlebars template
- * @param {Object} options  The default options to be extended and passed to renderer
- * @returns {Promise}       Promise resolving in a rendered dataset (CSV)
+ * Render a csv file`
+ * @param {object} data - Object of keys/values for the data
+ * @param {string} template - Path to a handlebars template
+ * @param {object} options - The default options to be extended and passed to renderer
+ * @returns {Promise} Promise resolving in a rendered dataset (CSV)
  */
 function renderCSV(data, template, options = {}) {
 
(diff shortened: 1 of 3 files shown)


server/lib/renderers/csv.js
```diff
@@ -40,17 +40,21 @@ exports.render = renderCSV;
...
+ * @returns {Promise} Promise resolving in a rendered dataset (CSV)
  */
 function renderCSV(data, template, options = {}) {
 
   // allow different server routes to pass in csvOptions
   const csvOptions = _.defaults(options.csvOptions, defaults);
 
+  // Force addition of the excel BOM to enable the output file to be treated
+  // as UTF-8 so language-specific UTF-8 characters, accents, etc, are retained
+  csvOptions.excelBOM = true;
```
jniles, line 56: !!  Amazing!

Ebed-meleck (approved):

PR #7336, opened 2023-11-14
correct the account selection field is required

### LINKED ISSUE #7288: The "releve de compte" report has required fields marked "optional"
In the Releve de Compte report, we have required fields (default filters) marked at "optional".  They should not be optional, as they are required.  See below:

image

### LINKED ISSUE #7338: Error downloading the "Relevé de compte" PDF
`Finance/Rapports/Relevé de compte`

When the account registry is empty, it should at least download the file. This returns an error
bug-pdf-nodata

Something else too when filtering the register for an account, if the debit is 0 and the credit has a value the calculation returns a negative number. Returns an error
bug-pdf-report-releve

Ebed-meleck:
[#7288](https://github.com/IMA-WorldHealth/bhima/issues/7288)

`Finance>Rapports/Rélevé de compte`  Correction of the account field on the modal filter

And Modification of the report, since we are filtering only one account fix [7338](https://github.com/IMA-WorldHealth/bhima/issues/7338)

Closes #7288.

### DIFF
--- a/client/src/modules/journal/modals/search.modal.html
+++ b/client/src/modules/journal/modals/search.modal.html
@@ -190,7 +190,7 @@
               name="account_id"
               on-select-callback="ModalCtrl.onSelectAccount(account)"
               exclude-title-accounts="true"
-              required="false">
+              required="true">
             </bh-account-select>
           </div>
 
--- a/server/controllers/finance/reports/account_statement/index.js
+++ b/server/controllers/finance/reports/account_statement/index.js
@@ -39,10 +39,10 @@ async function report(req, res, next) {
     const aggregateSql = `
       SELECT SUM(debit_equiv) AS debit_equiv, SUM(credit_equiv) AS credit_equiv
       FROM general_ledger
-      WHERE uuid = ?;
+      WHERE uuid IN (?);
     `;
 
-    const [transactionUuids] = rows.map(row => db.bid(row.uuid));
+    const transactionUuids = rows.map(row => db.bid(row.uuid));
     const aggregateData = await db.one(aggregateSql, [transactionUuids]);
     let aggregate;
     if (aggregateData) {
--- a/server/controllers/finance/reports/account_statement/index.js
+++ b/server/controllers/finance/reports/account_statement/index.js
@@ -37,15 +37,22 @@ async function report(req, res, next) {
     const rows = await generalLedger.findTransactions(options);
 
     const aggregateSql = `
-      SELECT SUM(debit_equiv) AS debit_equiv, SUM(credit_equiv) AS credit_equiv,
-        SUM(debit_equiv - credit_equiv) AS balance
+      SELECT SUM(debit_equiv) AS debit_equiv, SUM(credit_equiv) AS credit_equiv
       FROM general_ledger
-      WHERE uuid IN (?);
+      WHERE uuid = ?;
     `;
 
-    const transactionUuids = rows.map(row => db.bid(row.uuid));
-    const aggregate = await db.one(aggregateSql, [transactionUuids]);
-
+    const [transactionUuids] = rows.map(row => db.bid(row.uuid));
+    const aggregateData = await db.one(aggregateSql, [transactionUuids]);
+    let aggregate;
+    if (aggregateData) {
+      aggregate = { ...aggregateData };
+      aggregate.balance = aggregateData.debit_equiv - aggregateData.credit_equiv;
+    } else {
+      aggregate.debit_equiv = 0;
+      aggregate.credit_equiv = 0;
+      aggregate.balance = 0;
+    }
     const result = await rm.render({ rows, aggregate, filters });
     res.set(result.headers).send(result.report);
   } catch (e) {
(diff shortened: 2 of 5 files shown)


client/src/modules/journal/modals/search.modal.html
```diff
@@ -190,7 +190,7 @@
               name="account_id"
               on-select-callback="ModalCtrl.onSelectAccount(account)"
               exclude-title-accounts="true"
-              required="false">
+              required="true">
```
jmcameron, line 193: Thanks!

test/end-to-end/account/accountStatement/account_statement.spec.js
```diff
@@ -47,8 +47,9 @@ test.describe('Account Statement Core', () => {
     expect(colTitle.trim()).toBe('Account');
   });
 
-  test(`comment the rows for account ${sample.account} with ${sample.comment}`, async () => {
-    await AccountStatement.setPeriod('allTime');
+  // @TODO: Fix. Works but fails because timeout exceeded
```
jmcameron, line 50: Ok.  I'll take a look at this when I get a chance.

test/end-to-end/account/accountStatement/account_statement.page.js
```diff
@@ -37,6 +37,25 @@ class AccountStatementCorePage {
     return TU.waitForSelector('[ui-grid-grid-footer]');
   }
 
+  async setPeriodAccount(periodName, accountNumber) {
```
jmcameron, line 40: 👍

jmcameron (commented):
The code looks okay.  I tested it.

I find the user interface very strange.  The first time you enter the page, nothing is listed.  When you use the [Search] option to change the period and select an account, you only see one account (as expected).  If it really is a single account, there should be no need to use the [Search] function to define it.  This should work like the "Simple Account Report" (change the start date to 2016 for comparison).   I suggest discussing this with Bruce.   We should probably also talk to Jonathan Niles about this since it's his bug report.  I think we'll probably end up removing this report.

Finally, there are two problems with the resulting report:  The account name/number is not in the report.  Also, it seems that the dates of some elements of the report are incorrect (compare them to the "simple" report).

server/controllers/finance/reports/account_statement/index.js
```diff
@@ -37,15 +37,22 @@ async function report(req, res, next) {
     const rows = await generalLedger.findTransactions(options);
 
     const aggregateSql = `
-      SELECT SUM(debit_equiv) AS debit_equiv, SUM(credit_equiv) AS credit_equiv,
-        SUM(debit_equiv - credit_equiv) AS balance
+      SELECT SUM(debit_equiv) AS debit_equiv, SUM(credit_equiv) AS credit_equiv
       FROM general_ledger
-      WHERE uuid IN (?);
+      WHERE uuid = ?;
```
jniles: This should be reverted to how it was originally.  It sums the `debit_equiv` and `credit_equiv` of all transactions in the report, so it needs to use all UUIDs.  Here, it only considers one UUID.  You can see the incorrect result below:

image


The red should have the result of each item in the debit column and the blue should have the result of everything in the credit column.

jniles (changes requested):
I finally got a chance to review this with real data.

The changes to the SQL are incorrect, but everything else looks good to me.  There are still a few bugs in the report (for example, the filters are repeated on the handlebars page, and the account name should be shown instead of the ID), but those are outside of the scope of this pull request.  Don't worry about them for this merge.

(An example of the bugs is this image:)
image


Ideally, if the account_id is not cached in AppCache, the report page should automatically open the filter search modal.  I thought this was the previous behavior, but I cannot find the code for it in our git logs, so clearly it was never implemented!  I'll make a new issue for bugs in this report.

Please revert the changes to the SQL and this will be good to merge in my view.

Great job!

jniles:
@Ebed-meleck, I'd be happy to work on the other bugs I found in this report after this PR is merged. 👍

server/controllers/finance/reports/account_statement/index.js
Ebed-meleck: I had made a mistake with little of the data that was in bhima_test.

Ebed-meleck:
@jmcameron  You can check the test, if you had updated semaphore. Because locally I don't see this error.

jmcameron:
Note that the CI test failure is due to a PDF generation problem.  Sometimes this goes away by running the tests again.  But now that we do not have Bors, this is not so easy to do.

Ebed-meleck:
> @Ebed-meleck, I'd be happy to work on the other bugs I found in this report after this PR is merged. 👍

Great.

jmcameron:
> > @Ebed-meleck, I'd be happy to work on the other bugs I found in this report after this PR is merged. 👍
> 
> Great.

I'm about to merge this PR.  I'll file an issue for the updates discussed above.

jmcameron (approved):
I tested this again and it seems to work fine now.

jmcameron:
I ran the tests on my local dev system and they all pass without issues.
I'll create a new issue for the remaining requests (per @jniles suggestion).

PR #7342, opened 2023-11-20
Adding the filter to the user page registry

Ebed-meleck:


jmcameron (commented):
The code looks fine and it works.

However, I suggest two improvements:
- Use the typeahead feature to show potential Username matches after typing a few characters.   For instance, see how "Stock > Stock Exit" to Patient works when entering a Patient name.
- It might also be useful to add a search option for "Caisse" (cash box).  That would list all users who have access to a specific caisse.

Ebed-meleck:
@jmcameron  Check the latest commit, if that meets your suggestion

jmcameron (commented):
The code looks good.  Nice job getting the typeahead selector for usernames and cashbox search working. 

However, I think there is a bug somewhere in the code for the cashbox search.  Sometimes when you select a cashbox in the search, it shows the results for the wrong cashbox:
image.
Maybe the onSelect callback is getting the wrong value?

Also, I was not sure if selecting the date (in the Search - Defaults tab) makes any sense.  Is it actully doing anything?  If not, I suggest removing it from the Search.

Ebed-meleck:
> The code looks good. Nice job getting the typeahead selector for usernames and cashbox search working.
> 
> However, I think there is a bug somewhere in the code for the cashbox search. Sometimes when you select a cashbox in the search, it shows the results for the wrong cashbox: image. Maybe the onSelect callback is getting the wrong value?
> 
> Also, I was not sure if selecting the date (in the Search - Defaults tab) makes any sense. Is it actully doing anything? If not, I suggest removing it from the Search.

 the latest commit fixes this problem

jmcameron (approved):
I tested this with the latest fixes.  It works correctly now.

Thanks!  Good job.

PR #7346, opened 2023-11-23
fix search by period

Ebed-meleck:
Fix search by time period and add the ability to choose role when searching

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -1075,6 +1075,7 @@
             "RENDERER" : "Select a Renderer",
             "REPORT_TYPE": "Select a Report Type",
             "RESULT_ACCOUNT_SCT": "Select a Result Account section",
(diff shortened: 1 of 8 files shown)


jmcameron (commented):
The code looks good.   I tested it and it works fine.
However, I have a suggestion:  Since the "Period" search is based on the user creation time, I susggest you add another column to display it as well as the "Last login" time.

jmcameron (commented):
The code looks good.  I tested it and it seems to work properly.

However, I have one further suggestion.   When you go to the Defaults tab in the Search dialog, it is not clear that "Period" refers to date range for when the accounts were created.  I suggest you update the label to something appropriate, such as "User created" or something similar:
image

With that minor change I think this PR will be ready to merge.

jmcameron (approved):
Thanks for the updated.

Issue #7315, opened 2023-10-27
Improve the documentation for running in production


jniles:
The BHIMA documentation for installation is pretty good and has seen several revisions.  It is found at https://docs.bhi.ma/en/for-developers/installing-bhima.html.

This documentation is mostly meant for getting developers up and running, and isn't ideal for running in production.  We should add a separate set of instructions for running in production that covers:
  1. Using `nginx` as a proxy so that users don't have to type the :`$PORT` after the url in the web browser.
  2. Setting a strong `SESS_SECRET` variable in the `.env`
  3. Setting up backups using `cron`.

Some optional stuff we might consider adding:
 1. Setting up `.my.cnf` so that system administration is easier.
 2. Encouraging `systemd` over `pm2`.

jmcameron:
@jniles Shouldn't this have been closed by your recent Debian set up PR?

jniles:
I suppose it could be.  I would like to keep working on these issues though.  It probably won't be as relevant if I can get the server to build with docker-compose.

jniles:
I did get the application working with docker-compose, so I'm closing this issue.

PR #7347, opened 2023-11-26
feat: add debian 12 installation script

jniles:
This commit adds an installation script to prepare, download, and set up BHIMA from a fresh install of Debian 12.  It uses docker for for mysql and redis and systemd for restart management.  Tested on an Intel NUC7 with a fresh install of Debian 12.

To use this script, simply download it to a freshly installed machine and execute it.  It assumes the user is named "bhima" and the user has super user privileges.

Partially addresses #7315.


jmcameron (approved):
Looks good!

Given the changes, this should not affect any current code.  I did run the regular tests and they worked.

PR #7349, opened 2023-11-28
Token mobile

Ebed-meleck:
Reduce the size of the token, because after authentication with the mobile application, it refuses to recover the data following the size of the token which was too large.

### DIFF
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -1075,6 +1075,7 @@
             "RENDERER" : "Select a Renderer",
             "REPORT_TYPE": "Select a Report Type",
             "RESULT_ACCOUNT_SCT": "Select a Result Account section",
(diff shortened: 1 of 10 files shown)


jmcameron (changes requested):
This PR seems to have combination of the token fixes and the fixes from https://github.com/IMA-WorldHealth/bhima/pull/7346 (fix search by period).   I suggest using `git rebase` to rebase it to include only the commits relevant to the mobile token updates.  Specifically it looks like only these commits are relevant:  https://github.com/IMA-WorldHealth/bhima/pull/7349/commits/0628ebe862cf61394c78428b95b29008ce8349ee, and https://github.com/IMA-WorldHealth/bhima/pull/7349/commits/1e9122e8fb8be141dfc19f996f16008949d065e9.   

Having only the changes for the token changes will make this PR easier to review and test.

PR #7357, opened 2023-12-02
User report

### LINKED ISSUE #7353: Add a "Created" date column to the Administration > User Management page
Now that we have the ability to search by the date the user was created (See https://github.com/IMA-WorldHealth/bhima/pull/7346), it would be helpful to display that in a column in the main user registry grid.

Ebed-meleck:
Adding pdf, excel, csv report.
And Addition #7353 

Closes https://github.com/IMA-WorldHealth/bhima/issues/7353

### DIFF
--- a/client/src/modules/users/users.js
+++ b/client/src/modules/users/users.js
@@ -3,16 +3,14 @@ angular.module('bhima.controllers')
 
 UsersController.$inject = [
   '$state', '$uibModal', 'UserService', 'NotifyService', 'ModalService', 'uiGridConstants',
-  'GridStateService', 'appcache', 'GridExportService', 'GridColumnService', 'LanguageService',
-  '$httpParamSerializer',
+  'GridStateService', 'appcache',
 ];
 
 /**
  * Users Controller
  * This module is responsible for handling the CRUD operation on the user
  */
-function UsersController($state, $uibModal, Users, Notify, Modal, uiGridConstants, GridState,
-  GridExport, Columns, Languages, $httpParamSerializer) {
+function UsersController($state, $uibModal, Users, Notify, Modal, uiGridConstants, GridState) {
   const vm = this;
   const cacheKey = 'usersGrid';
 
@@ -20,7 +18,6 @@ function UsersController($state, $uibModal, Users, Notify, Modal, uiGridConstant
   vm.toggleFilter = toggleFilter;
   vm.editRoles = editRoles;
   vm.search = search;
-
   // this function selectively applies the muted cell classes to
   // disabled user entities
   function muteDisabledCells(grid, row) {
@@ -101,8 +98,6 @@ function UsersController($state, $uibModal, Users, Notify, Modal, uiGridConstant
   };
 
   const state = new GridState(vm.gridOptions, cacheKey);
-  const columnConfig = new Columns(vm.gridOptions, cacheKey);
-  const exportation = new GridExport(vm.gridOptions, 'selected', 'visible');
   vm.saveGridState = state.saveGridState;
 
   function onRegisterApiFn(gridApi) {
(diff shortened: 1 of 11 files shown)


jmcameron (commented):
Thanks!  The report seems to work ok.   I have not reviewed the code in detail yet.  
I have three suggestions:

1. Add the "Created" column to the main http://localhost:8080/#!/users page.
2. The report seems correct.  It reflects the search parameters.  However, it will not be obvious to the users what search parameters were used in the report.  I suggest adding a line to the report including which search parameters are used for the report.  For example, see the Finances > Report > Account Statement (Relevé de compte) PDF output:
image
3. Users who are Deactivated (disabled) are shown in the online page crossed out.  I think the PDF output should do that too (using <strike></strike>).  The PDF should also include the "Deactivated" column too.  Crossing out entries in the CSV and Excel files may not be possible but that is okay because the "Deactivated" column is already included.

Also, I noticed a bug in the Excel export:  When I tried to access it, Excel complained about a problem with the "file content".  Not sure what the reason for that, but I noticed that hte "enable_external_access" column had no data.

Finally, in the Excel export, columns with boolean data and 1 for true/1 but nothing for false/0.   Note that the CSV version handles this correctly.

client/src/modules/users/users.js
```diff
@@ -80,6 +84,16 @@ function UsersController($state, $uibModal, Users, Notify, Modal, uiGridConstant
         displayName : 'USERS.LAST_LOGIN',
         cellFilter : 'date:"dd/MM/yyyy HH:mm:ss"',
         headerCellFilter : 'translate',
+        visible : true,
+        cellClass : muteDisabledCells,
+      },
+      {
+        field : 'created_at',
```
jmcameron, line 91: Thanks!

server/controllers/admin/users/reports/report.handlebars
```diff
@@ -0,0 +1,46 @@
+{{> head }}
+
+<body>
+
+  {{> header }}
+
+  <!-- body  -->
+  <!-- filters  -->
+  {{> filterbar filters=filters }}
```
jmcameron, line 9: 👍

server/controllers/admin/users/reports/report.handlebars
```diff
@@ -0,0 +1,46 @@
...
+        <th>{{translate 'FORM.LABELS.DEPOT'}}</th>
+        <th>{{translate 'FORM.LABELS.CASHBOXES'}}</th>
+        <th>{{translate 'USERS.LAST_LOGIN'}}</th>
+        <th>{{translate 'FORM.LABELS.DATE_CREATED'}}</th>
+      </tr>
+    </thead>
+    <tbody>
+      {{#each rows}}
+        <tr>
+          <td class="text-muted" {{#if deactivated}} style="text-decoration:line-through;" {{/if}}>{{display_name}}</td>
```
jmcameron, line 27: 👍

jmcameron (commented):
The changes look great!   Other than one minor problem, I think it is almost ready to merge.

Although some of the search options appear on the PDF report, not all of them do.  For instance, see these options on the Users page:

image

However, on the PDF report, the "Roles" is missing:

image

I checked some of the other search parameters in the PDF output:
- Cashbox works
- Depot works
- Role does NOT display in the PDF
- Username does NOT display in the PDF
- Last Login does NOT display in the PDF
- Date Created works in the PDF output but does not show in the "Filters" line in the User page

If you feel that this update is needed soon by any site, I am okay with merging this as is and submitting another Issue for these items missing from the PDF search filters line.

Ebed-meleck:
> The changes look great! Other than one minor problem, I think it is almost ready to merge.
> 
> Although some of the search options appear on the PDF report, not all of them do. For instance, see these options on the Users page:
> 
> image
> 
> However, on the PDF report, the "Roles" is missing:
> 
> image
> 
> I checked some of the other search parameters in the PDF output:
> 
> * Cashbox works
> * Depot works
> * Role does NOT display in the PDF
> * Username does NOT display in the PDF
> * Last Login does NOT display in the PDF
> * Date Created works in the PDF output but does not show in the "Filters" line in the User page
> 
> If you feel that this update is needed soon by any site, I am okay with merging this as is and submitting another Issue for these items missing from the PDF search filters line.


I just corrected it, you can check if it's good

jmcameron (approved):
Thanks for the updates! I think this is a useful update for administrators to track users!

Great job!

PR #7358, opened 2023-12-02
generate short token

Ebed-meleck:
Reduce the size of the token, because after authentication with the mobile application, it refuses to recover the data following the size of the token which was too large.

### DIFF
--- a/server/config/access.js
+++ b/server/config/access.js
@@ -2,6 +2,7 @@ const _ = require('lodash');
 const debug = require('debug')('app');
 const JWTConfig = require('./jwt');
 const { Unauthorized } = require('../lib/errors');
+const { loadSessionInformation } = require('../controllers/auth');
 
 const publicRoutes = [
   '/auth/login',
@@ -16,36 +17,26 @@ const publicRoutes = [
 
 module.exports = (app) => {
   // eslint-disable-next-line consistent-return
-  app.use((req, res, next) => {
+  app.use(async (req, res, next) => {
     const token = req.headers['x-access-token'];
-    let sessionExists = true;
+    try {
 
-    if (token) {
-      JWTConfig.verify(req, (session) => {
-        if (!session) {
-          sessionExists = false;
-          return;
-        }
+      if (token) {
+        const user = await JWTConfig.verify(token);
+        const session = await loadSessionInformation(user);
         _.merge(req.session, session);
-      }, (error) => {
-        sessionExists = false;
-        debug(error);
-      });
+      }
+
+      if (_.isUndefined(req.session.user) && !within(req.path, publicRoutes)) {
+        debug(`Rejecting unauthorized access to ${req.path} from ${req.ip}`);
+        next(new Unauthorized('You are not logged into the system.'));
+      }
+
+      // go to the next middleware
+      next();
+    } catch (err) {
+      next(err);
     }
-
-    if (!sessionExists) {
-      debug(`Rejecting due to invalid session token`);
-      next(new Unauthorized('The session token is not valid'));
-    }
-
-    if (_.isUndefined(req.session.user) && !within(req.path, publicRoutes)) {
-      debug(`Rejecting unauthorized access to ${req.path} from ${req.ip}`);
-      next(new Unauthorized('You are not logged into the system.'));
-    }
-
-    // go to the next middleware
-    next();
-
   });
 
 };
--- a/server/controllers/auth.js
+++ b/server/controllers/auth.js
@@ -35,7 +35,7 @@ function loginRoute(req, res, next) {
     .then(session => {
       // bind the session variables
       _.merge(req.session, session);
-      session.token = JWTConfig.create(session);
+      session.token = JWTConfig.create({ ...session.user });
       // send the session data back to the client
       res.status(200).json(session);
     })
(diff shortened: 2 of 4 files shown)


jniles:
What was the (maximum) size of the token before, and what is the (maximum allowable) size of the token now? 

It looks like the size depends on the session.user object, which contains things like the username that the user can set.  Can you please confirm that this will not generate a token too large if the username is the maximum length of 80 characters? 

https://github.com/IMA-WorldHealth/bhima/blob/748936412c63a714e30547392d17162d0c1e2f25/server/models/schema.sql#L1798

It might be good to generate the maximum size of a req.session.user possible and make sure that doesn't crash the mobile app.  If it does, we'll need to think of a different solution on the long term, since a user would be able to crash the app by choosing large inputs for their user information and that probably isn't good.

Ebed-meleck:
> What was the (maximum) size of the token before, and what is the (maximum allowable) size of the token now?
> 
> It looks like the size depends on the session.user object, which contains things like the username that the user can set. Can you please confirm that this will not generate a token too large if the username is the maximum length of 80 characters?
> 
> https://github.com/IMA-WorldHealth/bhima/blob/748936412c63a714e30547392d17162d0c1e2f25/server/models/schema.sql#L1798
> 
> It might be good to generate the maximum size of a req.session.user possible and make sure that doesn't crash the mobile app. If it does, we'll need to think of a different solution on the long term, since a user would be able to crash the app by choosing large inputs for their user information and that probably isn't good.



There is no limit but this is what is recommended: the size of the token must be <8kb. Previously, after generating the token, it had a length of 2447 characters, which was too much, hence the mobile side returned error 431.

Because the token was regenerated on several character strings which contained a lot of information (paths, actions, user, project). But by just reducing the information based on that of the user, I believe even if the username is more than 80 characters the size of the token will always be < 8kb.

server/config/access.js
```diff
@@ -16,36 +17,26 @@ const publicRoutes = [
...
 
-    if (token) {
-      JWTConfig.verify(req, (session) => {
-        if (!session) {
-          sessionExists = false;
-          return;
-        }
+      if (token) {
+        const user = await JWTConfig.verify(token);
+        const session = await loadSessionInformation(user);
```
jniles, line 26: Nice use of async/await 👍

server/config/access.js
```diff
@@ -16,36 +17,26 @@ const publicRoutes = [
...
-      });
-    }
+      }
 
-    if (!sessionExists) {
-      debug(`Rejecting due to invalid session token`);
-      next(new Unauthorized('The session token is not valid'));
-    }
+      if (_.isUndefined(req.session.user) && !within(req.path, publicRoutes)) {
+        debug(`Rejecting unauthorized access to ${req.path} from ${req.ip}`);
```
jniles, line 31: Nice error message!

server/controllers/auth.js
```diff
@@ -35,7 +35,7 @@ function loginRoute(req, res, next) {
     .then(session => {
       // bind the session variables
       _.merge(req.session, session);
-      session.token = JWTConfig.create(session);
+      session.token = JWTConfig.create({ ...session.user });
```
jniles, line 38: Instead of using the session here, it looks like the only information the application needs is the username, password, and user id.  Let's just make a new object with those parameters:
```js
session.token = JWTConfig.create({ id: session.user.id, username: session.user.username, password: session.user.password });
```

jniles (commented):
I'm still a little confused.  I assume you mean that NodeJS is sending error 431, right?  Not the app?

Digging around, it looks like 8KB used to be the limit, but [now it is 16Kb](https://nodejs.org/api/http.html#httpmaxheadersize).  However, it also looks configurable, and we could configure it if we need to using [this solution](https://github.com/expressjs/express/issues/4131).

Looking at [the BHIMA stock application](https://github.com/IMA-WorldHealth/bhima_stock/blob/master/lib/services/connect.dart#L38) it looks like only the id, username, and password is used.  If so, I propose just using the information in the token.  That way, we will have a predictable way to control the token size (we know the size of each of those database fields) and can make changes in the future if the size becomes too big.

server/controllers/auth.js
Ebed-meleck: Oh, that's good but otherwise we only create the information of the user object, this will make it easier after decoding to reload the information in the session.
file `server/config/access.js` line 26

Ebed-meleck:
> I'm still a little confused. I assume you mean that NodeJS is sending error 431, right? Not the app?
> 
> Digging around, it looks like 8KB used to be the limit, but [now it is 16Kb](https://nodejs.org/api/http.html#httpmaxheadersize). However, it also looks configurable, and we could configure it if we need to using [this solution](https://github.com/expressjs/express/issues/4131).
> 
> Looking at [the BHIMA stock application](https://github.com/IMA-WorldHealth/bhima_stock/blob/master/lib/services/connect.dart#L38) it looks like only the id, username, and password is used. If so, I propose just using the information in the token. That way, we will have a predictable way to control the token size (we know the size of each of those database fields) and can make changes in the future if the size becomes too big.

This error is not on the nodejs side but rather on the client side

jniles (approved):
@Ebed-meleck, I don't know Dart well enough to know what error might be reproduced on the client side.  I assumed you were referring to HTTP error 431.  If so, that would be sent by the HTTP server, which is NodeJS in this case.  Even if it is displayed on the client, it still would be generated by the sever.

I'm happy with this PR.  I still worry there may be a corner case where a long username/password combination will crash the client, but I guess we can worry about that if it happens in production.

Issue #7355, opened 2023-11-30
Update French BHIMA documentation


jmcameron:
Update the French versions of the docs.bha documentation. These Markdown files are in the directory:
- `docs/en`  (English version)
- `docs/fr`  (French version)

Various updates and additions have happened for the English versions, but not all of these changes have been applied to the French version.   It is likely that there are some pages and content that needs to added, created, or changed in the French version.  

There are also tools and editors that will allow you to visualize your *.md files locally.  For instance, Visual Studio has a preview mode for Markdown files.

When you make changes to the docs/ *.md files and they are merged into master, they are automatically available on 

- `https://docs.bhi.ma`

jmcameron:
Closed by https://github.com/IMA-WorldHealth/bhima/pull/7361

jmcameron:
Closed by https://github.com/IMA-WorldHealth/bhima/pull/7361

PR #7361, opened 2023-12-04
Documentation french

Ebed-meleck:
Updated documentation in French 
#7355


jmcameron (approved):
I did not read the updates carefully but they seem complete.  However, I think it is valuable to merge them now and have a more systematic review of documentation (both English and French) later.

PR #7374, opened 2023-12-13
fix warning bad format date after report user

Ebed-meleck:


jmcameron (commented):
This update does eliminate the server error messages when generating the PDF report.  Thank you!

There is still a problem with the PDF report.  Notice the display of filters works on the page:
image

But the Depot and Role displays are UUIDs, not human-readble.
image

Perhaps you should not have deleted 'options?.displayValues'?

Also, I did not check the CSV/Excel output.   Once you fix the PDF problem, I suggest you verify that the CSV/Excel exports are still working properly.

server/controllers/admin/users/reports/index.js
```diff
@@ -28,6 +28,7 @@ async function report(req, res, next) {
     suppressDefaultFiltering : true,
     suppressDefaultFormatting : false,
   });
+  delete options?.displayValues;
```
jmcameron: This line should be removed.

jmcameron (changes requested):
I tested this again by commenting out the  line `delete options?.displayValues;`.  It works correctly.   Here is what selecting all the custome options looks like in the PDF file:

image

jmcameron (approved):
Looks good and works now.

PR #7379, opened 2023-12-17
chore: update testing standards

jniles:
This PR does two things:

  1. Removes the legacy bors configuration from the repo.  
  2. Calls dotenv before running playwright tests to ensure that server is running on $PORT configured in the users dotenv, rather than hardcoded 8080.  Defaults back to 8080 for compatibility reasons.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -1,9 +1,0 @@
-status = [
-  "ci/semaphoreci/%"
-]
-
-delete_merged_branches = true
-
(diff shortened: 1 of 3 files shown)


jmcameron (approved):
Looks good.  
Thanks @jniles!

Note that the default CI/Semaphore tests run 'npm run test:e2e-account' which use Playwright.  So I am confident this is okay.

PR #7384, opened 2023-12-20
feat(payroll): add service to the excel export

### LINKED ISSUE #7383: Add "service" column to the Excel spreadsheet of the Multiple Indice Payroll report
We have received the following message from one of the BHIMA sites.

> Since I have time, I am preparing to do some department (service) based reports to review with AG and the CODI looking at trends over this year.  For example using the `multiple_payroll_indice` report how many total supplementary days does each service use each month?  I also think I can also figure out the average cost of these supplementary days.   The point is for every 26 supplementary days in a department there could be an additional person hired and might save money as entry salary is lower then the senior staff (but that would cut the salary of others since there is a single pay envelope divided amongst the group.)

> I see the Service on the multiple_payroll_indice table online but when the Excel report gets produced from BHIMA that column is not visible either on the online version or the backup here at my house.  How difficult would it be to fix this?

I believe the user would like to see the "service" column appear in the Excel print out of the Multiple Indice Payroll report.

jniles:
Adds the service to the Excel export of the Multiple Payroll Indice report.

In order to test, please navigate to the Multiple Payroll Indice grid:
  1. In the tree, `Payroll > Indexes > Multiple Payroll (Index-based)` (`#!/multiple_payroll_indice`).
  2. Configure the report for a payment period, if necessary.
  3. Go to the menu, and download the Excel report (`Menu > Download as Excel`).
  4. Notice the second column in the Excel report is `Service`.

Compare this to `master`, where no service column exists.

Closes #7383.

### DIFF
--- a/server/controllers/payroll/multiplePayrollIndice/index.js
+++ b/server/controllers/payroll/multiplePayrollIndice/index.js
@@ -14,7 +14,7 @@ exports.parameters = require('./paramter.config');
 exports.reports = require('./report');
 
 // retrieve indice's value for employee(s)
-async function read(req, res, next) {
+function read(req, res, next) {
   lookUp(req.query).then(rows => {
     res.status(200).json(rows);
   }).catch(next);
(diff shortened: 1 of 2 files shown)


jmcameron (commented):
The code looks good.  I tested this and it seemed to work.  However, when I went back to master and tried the same thing, the results looked the same to me.    

I suggest adding some instructions to the PR on how to test it.

jniles:
@jmcameron updated with testing instructions.

jmcameron (approved):
Looks good.   With the instructions, I was able to test this (I was testing the wrong Payroll page before).  The PR adds the Service column, per its description.  I was surprised there was not more data in the Excel file.

PR #7386, opened 2023-12-26
Improvements to complex vouchers

### LINKED ISSUE #2196: Improvements to Complex Vouchers
The following improvements need to be made the Complex Vouchers:
 1. ~~Use a typeahead for accounts instead of a `ui-select`.  This will improve the performance of the page and also be the same style as the transaction editing modal.~~  Fixed: #2226.
 2. Allow a user to link the debtor/creditors and references by typing, as in the Transaction Edit Modal.  Most users will probably prefer this to scrolling through millions of lines to find their particular transaction.  We could still include the modal behind a double click.
 3. The debtor/creditor selection shouldn't require a user to select "debtor/creditor".  It might be easiest to just have an `entity/` route that allows a typeahead to fill with both based on the input.  We would label each with a [D] or [C] depending on the type.
 4. Remove the "Recover Items" since it doesn't work.
 5. ~~The transaction type should automatically focus itself.  At the moment, it doesn't focus the input when a user clicks on it.~~
 6. ~~[BUG] submitting a form without any rows in the grid results in an empty error message.~~

jniles:
This commit refactors the Advanced Journal Vouchers to meet all the requirements stated in #2196.

### Why is this important?
As a reminder of the motivation behind #2196 which may not have been clearly articulated (or too long ago now to remember), the reason a refactor of this code was important was two-fold:
  1. We hoped to use the same or similar components to what was used in the transaction edit modal for journal vouchers to both reuse code between the two, and also reduce the differences in what users had to learn to make both work.  I partially achieved code reuse in this rewrite, but ended up removing redundancies more than reusing code.
  2. We also observed that as more transactions were added to the database, there was no reasonable way of creating a quick interface for linking entities or references to vouchers.  The old way required loading _every single transaction_ for cash, vouchers, and invoices, then searching for the specific one by reference, then clicking on it.  This was slow (at risk of crashing small servers in practice) and rarely done by both us the developers and users.  We never recommended it after we implemented it.

These reasons above were the background of #2196.

### What does this PR do?
In addition to satisfying the objectives laid out in the issue, this PR cleans up a lot of cruft in the complex vouchers form service.  Notably, the modals to find references and entities have been removed, and the voucher tools updated to deal solely with uuids instead of having to create debtor/creditor references.  This makes the complex vouchers more performant, easier to use, and code clearer.

Here is what it looks like:

image

### What is next?
This is a pretty significant change and may need some retraining of users at installation sites.  I'm posting this as a draft to get comments from the team before turning it into a full PR.  It's been 4 years in the making 😆 !

I still believe now as I did 4 years ago that this is the best way to make Complex Vouchers more usable.  The PR also perfectly sets up the codebase to implement https://github.com/IMA-WorldHealth/bhima/issues/2101 in a future PR without too much issue. Combined, these will make Complex Vouchers more useful for modeling complicated transactions.

It also paves the way for https://github.com/IMA-WorldHealth/bhima/issues/5936 to be implemented as well.

Closes #2196.


jniles:

jniles:
@mbayopanda when can I expect a response?

client/src/js/components/bhEntityTypeahead.js
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  $ctrl.lookupEntities = (text) => {
+    cancelInProgressRequests();
+    if (text.length < 3) { return null; }
+    return FindEntities.read(null, { text, limit : 10 }, { timeout : timer.promise });
+  };
+
+  // cancels all pending requests
+  function cancelInProgressRequests() {
```
mbayopanda, line 50: 👍🏽

client/src/modules/templates/bhAccountTypeaheadInline.html
```diff
@@ -2,7 +2,7 @@
   <div
     class="form-group"
     style="margin-bottom:0;"
-    ng-class="{ 'has-error' : AccountForm.account.$error && AccountForm.$$parentForm.$submitted }">
+    ng-class="{ 'has-error' : AccountForm.account.$error && AccountForm.$submitted }">
```
mbayopanda, line 5: 👍🏽

server/controllers/finance/vouchers.js
```diff
@@ -358,16 +389,14 @@ function createVoucher(voucherDetails, userId, projectId) {
   transaction.addQuery('CALL PostVoucher(?);', [voucherDetails.uuid]);
 
   // Only for Employee Salary Paiement
-  if (voucherType === 7) {
+  if (voucherTypeId === SALARY_PAYMENT_VOUCHER_TYPE_ID) {
```
mbayopanda, line 392: 👍🏽

mbayopanda (approved):
This is an improvement in the UX and in the code.

The UI and the feature are working as expected and fine, and also seeing the entity and reference in the receipt is another point of stability of the feature.

There are no comments on the code, it's excellent.

You have just to update the `test/data.sql` to add the fiscal year 2024

jmcameron:
@jniles, I had difficulties testing this PR.  
Please merge with master and verify `npm run test` and `npm run `test:e2e-3` run succesfully.  Thx.

jniles:
@jmcameron I have done both.  Note that the end to end test on master frequently time out of me on my hardware, and did time out for me on this PR as well.  However, not on the relevant tests (complex vouchers).

jmcameron (changes requested):
I have not read the code in detail yet.
Basic tests pass.  But the end-to-end tests related to vouchers fail (more on this below).
When creating a simple voucher, the dropdown for the "Voucher Type" does not seem to work.

To run just the voucher end-to-end tests, temporarily add the following line to package.json:
`"test:e2e-test": "cross-env SUITE=\"end-to-end\" TESTS=\"test/end-to-end/vouchers/\" TEST_NUM=8 ./sh/test-e2e.sh",`
and run the test with:  `npm run test:e2e-test`

A few notes on the end-to-end voucher tests:
- in `test/end-to-end/couchers/complex.spec.js`, 
   - the const `voucher` needs to be updated with suitable values for Debtor/creditor and Reference
- in `test/end-to-end/vouchers/voucherRowClass.js`
   - reference()` and `entity()` need to be updated since they no longer use popups.  I was not clear on how this was supposed to work.
- The simple voucher test may need updating based on how the dropdown issue is resolved.

jniles:
@jmcameron, thanks for providing steps to reproduce.  I can reproduce the issue and will fix it soon!

jniles:
@jmcameron I think this is ready for another set of tests on your machine.  Thanks for catching that; the voucher tests for be fixed, and simple vouchers was totally borked 😱 .   Now those are fixed as well!

jmcameron (changes requested):
There still seem to be problems on my local server.
- In Simple Voucher form, the "Transaction Type" drop-down does not work (for me).  It is a drop-down, correct?  not a type-ahead?  BTW, this works fine in master.
- The simple voucher end-to-end tests fail:
   - `test/end-to-end/vouchers/simple.spec.js:36:3 › Simple Vouchers › can create a simple voucher`
   - You can try this using the method suggested in https://github.com/IMA-WorldHealth/bhima/pull/7386#pullrequestreview-1826701649.  Or run `test:e2e-8`.

jniles:
@jmcameron, weird a cherry-picked commit didn't get pushed.  It should be fixed now.

test/end-to-end/shared/TestUtils.js
```diff
@@ -370,7 +370,7 @@ module.exports = {
     await this.input(model, label, anchor);
 
     // select the item of the dropdown menu matching the label
-    return anchor.locator(`.dropdown-menu > [role="option"] >> text="${label}"`).click();
+    await anchor.locator(`.dropdown-menu > [role="option"] >> text="${label}"`).click();
```
jmcameron: This  function is advertised as returning a promise.  It used to do that with the final 'return...'.   Now it does not do that.  Was there a reason for this change?

server/config/routes.js
```diff
@@ -1064,6 +1066,11 @@ exports.configure = function configure(app) {
   // on the fly tag numbers
   app.get('/lots/generate_barcodes/:number', lots.generateBarcodes);
 
+  app.get('/finance/entities', financeShared.lookupFinancialEntity);
+  app.get('/finance/entities/:uuid', financeShared.lookupFinancialEntityByUuid);
```
jmcameron, line 1070: Are the tests for the /finance/entity and /finance/record API endpoints?

jmcameron (changes requested):
Simple vouchers work now.
However, I am still confused about how the 'entity' and 'record' typeahead fields are supposed to work in complex vouchers.   When I try to into them, nothing happens.  Is that the correct behavior (with bhima_test) because no entity/records are available, or am I missing something?  Operator error?

The code looks good.   I made a few comments on things I noticed.

I noticed the a very minor 'nit':  In the voucher registry, the column header for the action menu should not appear in the Column selector.  This is due to setting the `displayName` for the action column to '...' instead of '' (see below).   If you don't mind, please do a global search and replace to change `displayName : '...'` to 'displayName : ''`.  I noticed 5 occurences for this and they are all for action items.  Note that these are not a big deal in the main grid displays, but using '...' means that they get added to the Column selector, which does not make a lot of sense since you always want them, and a checkbox with ... beside it is not very informative!
image

test/end-to-end/shared/TestUtils.js
jniles: Nope, fixed.

server/config/routes.js
jniles: Just added some!  This was a good catch too, because I found an unexpected behavior around LIMITs.

jniles:
@jmcameron this is ready for another review.

The way to test these routes is to put in a valid identifier to the entity or record section of the complex voucher.  For example "PA.TPA.2" might be one for a patient.  Similarly, "IV.TBA.2" might be another.

If you cannot find a known invoice, voucher, or cash payment... that is a bug.

In order to prevent spamming the server with queries that hit (potentially large) tables, we only start sending requests to populate the dropdown after 3 letter are typed.  That should get through the prefix ("IV." or "VO.").  There is also little point is querying the database at this point since IV. matches every single invoice.

Admittedly, it's not the greatest user experience, and could be improved upon.  I'm also wondering if we should accept free-form text and then do a match on the description of stock movements, invoices, etc, in that field.  So you might be able to type in "payment for quinine" and it bring up a series of CP.* that match that description.  This could also lead to _way_ too much noise for users in practice as well....  so not clear yet on what direction to take this.  Either way, that is for another issue and pull request.

jmcameron (approved):
It all seems to work now.  I suggest merging with master and rebasing.  Also, if you could fix the minor "nit" I mentioned at the end of [this previous comment](https://github.com/IMA-WorldHealth/bhima/pull/7386#pullrequestreview-1848784760), please do that, or add a new issue so we can fix it later.

Regarding the `entity` and `reference` fields in the complex voucher form:  I found it usable, once I realized that that you have to type in 3 characters to get it to show choices.  But it took some fumbling around to guess what the possible prefixes are.   I'm not sure how to improve that since you have to know what code prefixes are possible for each field before you start typing.  However, since the developer/database would usually(?) know what the possible prefix code options are, perhaps when you click in the input field you get a modal dialog with the possible options.  Click on one and a drop-down appears with the available items. Click on the one you want and it fills out the original input form.   Perhaps an issue for later?   Also, do we have any hints elsewhere to indicate when a field requires 3 characters to be typed in before anything happens that could be applied here?   Or do we assume that a person with training and experience would know what to do?

jniles:
@jmcameron thanks for your feedback and diligence in testing this.  I'm going to spin the nit into a different issue so we can address it on its own.

Ideally, a user should be well acquainted with the prefixes since they are printed on every receipt.  But I do think the usability and discoverability could be much improved.  Auto-prefixing is not a bad idea ... I'd have to think about how to implement it effectively within the frameworks we are using.

jniles:
@jmcameron this is ready to go anytime 👍

PR #7389, opened 2023-12-27
fix(vouchers): remove patientName column

jniles:
Removes the patientName column that was accidentally copied and pasted in by yours truly, two years ago.  This is the PR that introduced the bug: https://github.com/IMA-WorldHealth/bhima/pull/6246.

The code was accidentally lifted from the patient registry which does have a patientName column.


jmcameron (approved):
This fix looks good.
I tested it and it works fine.

PR #7392, opened 2023-12-29
Fix cashboxes e2e problem

jmcameron:
Testing fixes to E2E tests.


jmcameron:
Will resubmit based on new release.

PR #7394, opened 2023-12-30
Fix end to end test problems

jmcameron:
Try several fixes to improve the pass rate of end-to-end tests.

Note that this also fixes a cosmetic bug the Create Fiscal year page in the drop-down for the previous fiscal year.


jmcameron:
Will reconstruct and create a new PR

PR #7411, opened 2024-01-09
Improvements for more end-to-end test robustness

jmcameron:
This PR includes some improvements to the end-to-end tests to reduce the failures of some of the tests.


jmcameron:
This PR was tested over several days on the BHIMA test server and it did reduce the number of failures. After the most recent updates, there was a period of abut 3 day when the all the tests were run about 60 times. Of those 60, there were only 7 tests that had any failures and 4 of those were at midnight, meaning 53/60 times all the tests ran without any failures. 

I most cases, the fixes to the tests involved improving the way that Playwright was waiting for something to happen (eg, a modal dialog to appear, etc).

The problems with the remaining tests will be addressed in a later PR.

PR #7419, opened 2024-01-15
add column project_id in table stock_movement

Ebed-meleck:
Addition of the project_id column in the stock_movement table, in order to identify the movement by project

### DIFF
--- a/client/src/modules/stock/StockFilterer.service.js
+++ b/client/src/modules/stock/StockFilterer.service.js
@@ -59,6 +59,9 @@ function StockFiltererService(Filters, AppCache, $httpParamSerializer, Languages
     {
       key : 'expiration_date_to', label : 'STOCK.EXPIRATION_DATE', comparitor : '<', valueFilter : 'date',
     },
+    {
+      key : 'project_id', label : 'FORM.LABELS.PROJECT', valueFilter : 'translate',
+    },
   ];
 
   class StockFilterer {
--- a/client/src/modules/stock/lots/modals/search.modal.html
+++ b/client/src/modules/stock/lots/modals/search.modal.html
@@ -63,6 +63,13 @@
             </div>
           </div>
 
+          <!-- project  -->
+          <bh-project-select
+            on-select-callback="$ctrl.onSelectProject(project)"
+            project-id="$ctrl.searchQueries.project_id">
+            <bh-clear on-clear="$ctrl.clear('project_id')"></bh-clear>
+          </bh-project-select>
+
           <!-- expiration date -->
           <bh-date-interval
             label="STOCK.EXPIRATION_DATE"
--- a/client/src/modules/stock/lots/modals/search.modal.js
+++ b/client/src/modules/stock/lots/modals/search.modal.js
@@ -57,6 +57,12 @@ function SearchLotsModalController(data, util, Store, Instance, Periods, Stock, 
     displayValues.inventory_uuid = inventory.label;
   };
 
+  // select project
+  vm.onSelectProject = function onSelectProject(project) {
+    vm.searchQueries.project_id = project.id;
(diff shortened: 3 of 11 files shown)


jmcameron:
It looks like you need to add a default value for project_id to the SQL adding that column.   I suppose defaulting it to 0 would work since there do not seem to be any projects with project_id == 0 in the databases I checked.  But then you cannot add a constraint that the project_id must match an existing project.  I suppose the admin who upgrades a production site with this fix could ask what they want as a default and edit the migration file before installing.

However, I'm a little confused about this and how projects work.  When you log in, you select a project.   Once you are logged for a project, shouldn't all stock activity be associated with that project?

jmcameron:
> It looks like you need to add a default value for project_id to the SQL adding that column. I suppose defaulting it to 0 would work since there do not seem to be any projects with project_id == 0 in the databases I checked. But then you cannot add a constraint that the project_id must match an existing project. I suppose the admin who upgrades a production site with this fix could ask what they want as a default and edit the migration file before installing.
> 
> However, I'm a little confused about this and how projects work. When you log in, you select a project. Once you are logged for a project, shouldn't all stock activity be associated with that project?

Perhaps @jniles could describe how projects work?

jniles:
@jmcameron, I do not know the use case of this feature.  I asked about it in the original issue - https://github.com/IMA-WorldHealth/bhima/issues/7234.

The projects were conceived as a namespace for combining records between PAX and Tshikaji, when they were running two servers asynchronously.  That way, human readable labels wouldn't change, even with an offline sync.  However, we stopped doing that level of async support and I haven't seen a design for what we could use them for since then.

There was a proposal to make `lot`s linked to project_id (https://github.com/IMA-WorldHealth/bhima/issues/5219), but that was just for lots, not for movements.  The idea being that if a lot of items was for Global Fund, you would be able to track that.  I'm not sure what it means to have a movement linked to a project.

Ebed-meleck:
> It looks like you need to add a default value for project_id to the SQL adding that column. I suppose defaulting it to 0 would work since there do not seem to be any projects with project_id == 0 in the databases I checked. But then you cannot add a constraint that the project_id must match an existing project. I suppose the admin who upgrades a production site with this fix could ask what they want as a default and edit the migration file before installing.
> 
> However, I'm a little confused about this and how projects work. When you log in, you select a project. Once you are logged for a project, shouldn't all stock activity be associated with that project?

In financial entries this is different for each project, but for the movement of stock you do not know how to see which movement is linked to this project. @mbayopanda had mentioned this hypothesis during a discussion.

jniles:
@Ebed-meleck, what is the question being answered by the addition of a project identifier?  I think this is an important question, considering it _will break all_ currently running CBCO sites, plus IMCK.  As good developers, we should come up with a use case, a design, and a design criteria before we implement things.

Previously, it was a way to keep label a variety records generated at two different, offline locations.  For Stock Movements, I assume the depot_uuid would already identify these records (since each depot is a physical site, potentially offline).

Projects don't have balance sheets, accounts, specific users, etc... so all they are is a label.  So, my questions would be:
  1. What information would a user gain from having a project_id attached to a stock movement record that isn't already answered by the depot_uuid?
  2. What does a stock movement being attached to a project mean?  
       a. For example, if I see a stock movement called "SM.PAX.123", what does that mean?  The stock belongs to PAX?  The stock is destined for PAX?  The stock movement is created by a user at PAX?  What additional information do we get as part the movement label that isn't part of the `user_id`, `depot_uuid`, etc?

If the BHIMA team (or user requesting the feature) could detail a design specification for what questions are not able to be answered currently, it would make it easier to understand why we need to break all BHIMA sites for this feature.

Ebed-meleck:
okay I understand
thanks .

PR #7421, opened 2024-01-18
change the order to call the procedure in normal movement

Ebed-meleck:
change the order of calling the RecomputeStockValue procedure in normalmovement

### DIFF
--- a/server/controllers/stock/index.js
+++ b/server/controllers/stock/index.js
@@ -457,7 +457,7 @@ async function movementsFromMobile(params) {
 
     // filter known lots of the movement initial movement
     // check that the quantity sent is less or equal to
-    // the inital quantity
+    // the initial quantity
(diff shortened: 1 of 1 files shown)


jmcameron:
Please add a note describing the purpose for this change.

Ebed-meleck:
This change only concerns the execution order, the RecomputeStockValue procedure was added to the transaction after its execution. In some cases this is added or not.

Let me explain, I integrated the batch of drugs from the mobile application into a repository which contained assets, I sent the data to the server successfully, but in the batch register it did not appear, I do it for other FOSA it's good. I do the same operation in the same repository on the web, it works, and then it shows me the batches that do not appear. I tried the same thing again with the mobile, still not and then this time, I call the RecomputeStockValue procedure in the mysql console, the batches appear. When I changed the calling order, all operations work correctly.

Reason why I'm making this little change

Issue #5219, opened 2020-12-09
Optionally filter lots by project_id


jniles:
One thing pointed out to me by the logistics team at IMA - if we ever wanted to make BHIMA respond to IMA's needs, we would need to segregate medicines/lots by different projects.  For example, we would need to know that lot 123 is for "covid relief" and lot 126 is for "ASSR" and so on.  Then, the depot would only show the stock from that particular project.  This would allow users to use it in a multi-project environment such as IMA.

We could make this an optional feature.

PR #7436, opened 2024-01-24
lot (add column project_id)

Ebed-meleck:
Add column project_id in lot table, filter the lot by project, column project in report


jmcameron:
I would prefer to talk about this more before accepting this change.  Is it possible that two projects will be getting lots from the same depot (at the same location)?    If that could ever be true, then this change is not adequate, since lots could be used for different projects.  In that case, the project_id needs to be in the stock_movement itself.

Ebed-meleck:
> I would prefer to talk about this more before accepting this change. Is it possible that two projects will be getting lots from the same depot (at the same location)? If that could ever be true, then this change is not adequate, since lots could be used for different projects. In that case, the project_id needs to be in the stock_movement itself.

It's just one piece of information in the bunch. which can help to monitor the batches of different projects.

jniles:
Just a few questions:

 1. How do we migrate old lots at IMCK?  What project do we use?  Or do we not use a project?
 2. Can we default the `project_id` to NULL instead of 0?  `0` is not allowed as a project_id in the project table.
 3. Is it possible for a user without access to a given project, to assign the lot to that project?

lomamech:
The **lot** table should guarantee the uniqueness of the batch for a specific item, if we were to link a batch to a project there would be a risk of having the same information duplicated and how would the system behave if two batches of the same products from two separate projects ended up in a single depot

jmcameron:
[_Voir le français ci-dessous._] I still think adding a project_id to lots is not very general and could lead to problems later.   As Chris said (and I said above)   my concern is that it seems possible (and perhaps even likely) that two projects will have the the same type of lots in one physical depot.  It seems problematic to require the the actual, physical depot keep the the batches of lots separate so that the depot manager can select the correct lots for each project.  Therefore, associating lots with projects seems short-sighted.  Somehow we need to associate the project with the actual stock movement (entry/exit). 

Je continue de penser que l'ajout d'un identifiant de projet aux lots n'est pas très général et pourrait entraîner des problèmes ultérieurs.   Comme Chris l'a dit (et comme je l'ai dit plus haut), ce qui me préoccupe, c'est qu'il semble possible (et peut-être même probable) que deux projets aient le même type de lots dans un dépôt physique.  Il semble problématique d'exiger que le dépôt physique réel garde les lots séparés afin que le gestionnaire du dépôt puisse sélectionner les lots corrects pour chaque projet.  Par conséquent, l'association des lots aux projets semble être une solution à courte vue.  D'une manière ou d'une autre, nous devons associer le projet au mouvement de stock réel (entrée/sortie).

Ebed-meleck:
This PR (even if I have to rework it again, for old batches in edit I put the possibility to assign it to a project. In case seeing the project authorization to the user, this had me escaped . And then make it optional,, add a line in the stock parameter for example), this was to respond to this proposal #5219. @jmcameron  you come back to the idea of seeing it in the stock movement, I think @jniles  had mentioned something in #7419 that caught my attention in the sense of the SM label on the report, there has a SM.SEMI .124 "SEMI" which is the project.

@jmcameron if you find that the information is not relevant enough, I could reopen this PR #7419 and fix it.

Or  @jniles can tell us more.

jniles:
Hi @Ebed-meleck,

There are a lot of old issues in BHIMA that don't make a lot of sense to implement anymore.  I've tried to close a few, but do not feel comfortable closing many of them as I am no longer officially part of the IMA team.  The issue you are referencing was created (along with several others - https://github.com/IMA-WorldHealth/bhima/issues/4859, https://github.com/IMA-WorldHealth/bhima/issues/4869, https://github.com/IMA-WorldHealth/bhima/issues/4895) when the project was running out of funding and we were looking to pivot away from hospitals.  Thankfully, that never happened and BHIMA is being installed in hospitals and health centers as our main focus. 

With any old issue, particularly those that are over 2+ years old, it makes sense to first propose a design, then do the work of the coding.  For example, https://github.com/IMA-WorldHealth/bhima/issues/209 is an issue that we spent some time trying to figure out the design.  Here is another really detailed issue for a proposal to the way BHIMA handles billing: https://github.com/IMA-WorldHealth/bhima/issues/1770.

In all the cases cited above, we never implemented them because the perceived risk was more costly than the perceived reward.  

Here are some examples of designs that got merged: https://github.com/IMA-WorldHealth/bhima/issues/273, https://github.com/IMA-WorldHealth/bhima/issues/843, https://github.com/IMA-WorldHealth/bhima/issues/2687, https://github.com/IMA-WorldHealth/bhima/issues/3012.

And finally here is an example of a design we canceled because we no longer needed it: https://github.com/IMA-WorldHealth/bhima/issues/2180.  I personally feel like the `project_id` associated with stock falls into this category.

Normally, the BHIMA team should have a roadmap for what issues are being prioritized, what issues are under consideration, etc.  That way, we don't do a lot of coding work that doesn't get merged into the application.  Does the current BHIMA team have a roadmap of features that are being worked on in the next ~6 months?

jmcameron:
Couldn't this be done with tagging?  
Stock movements do not currently have tags but that could be added and might be useful in other ways too.
Lots do have tags, but my concern about the potential of lots being distributed to more than one project from the same depot make me wonder if that is a good approach.
But in general, the more I try to figure out what projects are (as implemented in BHIMA), the less I understand them.  So instead of using "Projects", perhaps tagging is a better approach?

Ebed-meleck:
> Cela ne pourrait-il pas être fait avec le marquage ? Les mouvements de stock n'ont actuellement pas de balises, mais cela pourrait être ajouté et pourrait également être utile à d'autres égards. Les lots ont des étiquettes, mais mes inquiétudes quant à la possibilité que les lots soient distribués à plusieurs projets à partir du même dépôt me font me demander si c'est une bonne approche. Mais en général, plus j'essaie de comprendre ce que sont les projets (tels qu'ils sont mis en œuvre dans BHIMA), moins je les comprends. Donc, au lieu d'utiliser « Projets », le balisage est peut-être une meilleure approche ?

Yes, that could be the case, from where I'll come back with a similar solution.
I close this PR

PR #7444, opened 2024-01-25
fix ui table group inventories

Ebed-meleck:
Fix column offset


jmcameron (approved):
Good catch!
I tested this with bhima_test in my browser and it is definitely a good fix!
All tests pass, including relevant end-to-end tests.

PR #7449, opened 2024-01-30
Prevent importing budget data for locked accounts

### LINKED ISSUE #7439: Prevent importing budget data for locked accounts
The Budget page will not display locked accounts.  However, if a locked account occurs in a CSV import file, it will be imported, although it will not be shown on the budget page.    This should not happen during the normal budget process, but is a corner-case that should be fixed.

This will need to be addressed during the import process, probably in the functions
- importBudget() in `server/controllers/finance/budget/index.js:774`
- SQL function InsertBudgetItem() in `server/models/procedures.sql:4257`

jmcameron:
Although PR https://github.com/IMA-WorldHealth/bhima/pull/7435 prevents the budget module from displaying locked accounts, it did not prevent budget imports from doing that.

This PR prevents importing budget data that includes accounts that are locked.  It errors out and prints an error message.

Closes https://github.com/IMA-WorldHealth/bhima/issues/7439

*TESTING*
- Use bhima_test, rebuild the database (`npm run build:db`)
- Use the **Finance > Budget Management** > Menu > Export YTD Budget Data (CSV)
- Edit the exported CSV and add a budget to several accounts such as 7011* (Vente...)
- In BHIMA, use the **Finance > Budget Management** > Menu > Import Bdget CSV file.  This should work.
- In the **Finance > Account Management** page, lock an account, such as 70112010 Vente d'actifs
- Return to the **Finance > Budget Management** page.  The locked account should not be visible.
- Try importing the same CSV file as before.  It should fail with a message showing which account is locked.  At this point, all the exiting budget data for the FY should be erased.
- Edit the CSV file and remove the line for the locked account.
- Repeat trying to import the BUDGET data.  It should work this time

### DIFF
--- a/client/src/i18n/en/budget.json
+++ b/client/src/i18n/en/budget.json
@@ -23,6 +23,7 @@
     "IMPORT_BUDGET_ERROR_ACCT_TYPE": "ERROR: Budget import failed!  Check for incorrect Type values (should be 'title', 'income', or 'expense').",
     "IMPORT_BUDGET_ERROR_ACCT_TYPE_INCORRECT": "ERROR: Budget import failed!  Account type does not match account in the database.",
     "IMPORT_BUDGET_ERROR_BAD_BUDGET_VALUE": "ERROR: Budget import failed!  Budget value is not a number.",
+    "IMPORT_BUDGET_ERROR_LOCKED_ACCOUNT": "ERROR: Budget import failed!  The budget import data includes a locked account. Please remove it and try again : ",
     "IMPORT_BUDGET_ERROR_NEGATIVE_BUDGET_VALUE": "ERROR: Budget import failed!  Budget value is negative.",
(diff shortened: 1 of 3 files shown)


mbayopanda (commented):
@jmcameron the feature works as expected, but it seems like we cannot import the account, even if its value is zero, so we can conclude the budget module does not display a locked account.

jmcameron:
> @jmcameron the feature works as expected, but it seems like we cannot import the account, even if its value is zero, so we can conclude the budget module does not display a locked account.

Right.  The main budget display does not show locked accounts or include them in the budget and actuals computations.  That was implemented in the last PR.   This budget prevents anyone from importing CSV budget data for locked accounts.   As I understand it, old obsolete accounts are never deleted, they are just marked as locked (and perhaps hidden).  So it does not make sense to allow them in the budget imports.

mbayopanda:
LGTM 👍🏽

Issue #7451, opened 2024-01-31
Cannot search by user on registries in BHIMA


jniles:
The user dropdown on registry search pages are broken for non-super users at all sites (IMCK, Vanga, etc).  You cannot search for invoices, vouchers, patients, or payments by user.   Here is a screenshot from Vanga:

![Untitled](https://github.com/IMA-WorldHealth/bhima/assets/896472/b5d52950-07c3-41bb-92d8-3a3c1b8a8f6d)

Normally, we expect a dropdown here that lists all users.  However, the database is reporting the following error:
```sh
Jan 31 16:44:40 bhima-NUC7i3BNK node[872]: 2024-01-31T15:44:40.786Z errors #interceptor() Forbidden [Error]
Jan 31 16:44:40 bhima-NUC7i3BNK node[872]:     at isAdmin (/home/bhima/apps/bhima-1.29.0/bin/server/controllers/admin/users/index.js:323:10)
Jan 31 16:44:40 bhima-NUC7i3BNK node[872]:     at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
Jan 31 16:44:40 bhima-NUC7i3BNK node[872]:   status: 403,
Jan 31 16:44:40 bhima-NUC7i3BNK node[872]:   code: 'ERRORS.FORBIDDEN',
Jan 31 16:44:40 bhima-NUC7i3BNK node[872]:   description: 'ERRORS.ER_ACCESS_DENIED_ERROR'
Jan 31 16:44:40 bhima-NUC7i3BNK node[872]: }
Jan 31 16:44:40 bhima-NUC7i3BNK node[872]: 2024-01-31T15:44:40.786Z http ::ffff:127.0.0.1 - GET /users/ HTTP/1.1 403 86 - 2.201 ms
```

It looks like the `isAdmin` permission is required to access the `/users` route, and that has broken all the registry search modals.  I propose we remove it and implement the same concept using `role`s.

jniles:
Closed in #7457

PR #7457, opened 2024-02-01
Remove the permission isAdmin in router /user

Ebed-meleck:
This change brings the fix for this  #7451

### DIFF
--- a/server/config/routes.js
+++ b/server/config/routes.js
@@ -584,7 +584,7 @@ exports.configure = function configure(app) {
   app.delete('/debtor_groups/:uuid', debtorGroups.delete);
 
   // users controller
-  app.get('/users', users.isAdmin, users.list);
+  app.get('/users', users.list);
   app.post('/users', users.isAdmin, users.create);
   app.get('/users/:id', users.detail);
   app.get('/users/:username/exists', users.isAdmin, users.exists);


jmcameron (commented):
It seems to me like this is a reasonable approach.   I checked to make sure that access to the `Administration > User Management` page is controlled by permissions so it is simple to control access to that page without requiring it for access to the /user REST paths.

However, please double-check all the `/users` REST targets to make sure that any of them that could modify users require `isAdmin`.  For instance, why does
`/users/:username/exists` require `isAdmin`?  At first glance the rest seem reasonable.   I welcome second opinions on this !

mbayopanda:
> It seems to me like this is a reasonable approach. I checked to make sure that access to the `Administration > User Management` page is controlled by permissions so it is simple to control access to that page without requiring it for access to the /user REST paths.
> 
> However, please double-check all the `/users` REST targets to make sure that any of them that could modify users require `isAdmin`. For instance, why does `/users/:username/exists` require `isAdmin`? At first glance the rest seem reasonable. I welcome second opinions on this !

Thank you @jmcameron for your feedback, the reason why we were obliged to restrict the `GET /users` endpoint was the fact that we have created an account for a partner, and they were able to get the list of all BHIMA users via the `GET /users` endpoint when we did not gave to them access to the `Administration -> User Management`, it was a kind of security issue.

So when we decide to allow `GET /users` only for `superuser`, we found logic to restrict also `/users/:username/exists`.

The reason why we have created this issue is the fact that some components used in some search `modals` are using the `GET /users` endpoints, with this restriction there is no way to get the list of users in these components.

mbayopanda (approved):
LGTM 👍🏽 

@Ebed-meleck could you update the branch so that I can merge it.

PR #7460, opened 2024-02-02
chore(payroll): fix typos/eslint fixes

jniles:
Fixes typos and misleading documentation in the controllers for the payroll modules.  Also fixes some eslint complaints about modules that were imported but never used.

No logic has been touched in this PR.

I also added some debug() calls to the multiplePayrollIndice file to help follow actual payments in the logs in the future.


jmcameron (approved):
Changes look good.  Thanks!
I tested it locally and the regular tests pass as well as all the end-to-end tests.

Issue #7461, opened 2024-02-02
Unable to display the item in stock and batch in stock register due to the error ER_DIVISION_BY_ZERO  


Ebed-meleck:
This error was produced during training, say after users added the batches. Once we go to the item in stock and lot in stock register. This error message appears.This error is due to the fact that a division by zero is reproduced when calling the GetAMC procedure. As shown in the console image
<img width="932" alt="Screen Shot 2024-02-02 at 19 03 10" src="https://github.com/IMA-WorldHealth/bhima/assets/63899331/4a5f870b-2ccc-4ed7-8b96-ab4f2b41e538">

This crashed both registers

jniles:
Just a note - https://github.com/IMA-WorldHealth/bhima/issues/6876 already reported some problems related to CMM - it might be worth looking more deeply at issues around calculating the CMM while you are doing your installation and making sure it works according to the design.

The design was proposed in https://github.com/IMA-WorldHealth/bhima/issues/4864.  More details related to stock outs were discussed in https://github.com/IMA-WorldHealth/bhima/issues/4346.

Ebed-meleck:
Closed via #7462

PR #7462, opened 2024-02-02
Added a condition to handle the ER_DIVISION_ZERO case in the GetAmc procedure

Ebed-meleck:
This change is the cause of #7461.

Added this condition to handle the case if this subtraction returns zero or null.


server/models/procedures.sql
```diff
@@ -3295,7 +3295,7 @@ CREATE PROCEDURE GetAMC(
   -- The average consumption is obtained by dividing the quantity consumed during the period by the difference of the
   -- number of months in the period minus the total number of days of stock out in the period. The MSH algorithm
   -- is recommended by the Management Sciences for Health organization (https://www.msh.org).
-  SET _algo_msh = (_sum_consumed_quantity / (_number_of_month - (_sum_stock_out_day / 30.5)));
+  SET _algo_msh = (_sum_consumed_quantity / IF((_number_of_month - (_sum_stock_out_day / 30.5)) = 0 OR NULL, 1, (_number_of_month - (_sum_stock_out_day / 30.5))));
```
jniles, line 3298: Seems reasonable to me.  We do the same work around for the regular algorithm.

jniles (approved):
LGTM.  👍 This is a good safeguard.

Do you know what kinds of scenarios would reproduce the issue you experienced during training?  Some steps to reproduce would be helpful so that we can make sure it doesn't happen in any production sites.

Issue #7464, opened 2024-02-03
Expiry date not showing on In Stock Batch Report 


Ebed-meleck:
On the pdf report produced,
The expiration date and delay expiration are not displayed even if the tracking_expiration of the inventory group is true.

Ebed-meleck:
Closed via #7465

PR #7465, opened 2024-02-03
fix the lot report expiration_date and delay_expiration

Ebed-meleck:
The change made the fix for this #7464

- Remove the tracking_expiration in the purgeKeys list.
- Added delay_expiration calculation

So that if this condition is true it displays the expiration date and the deadline in the report


jmcameron:
Please say what report this affects!

Ebed-meleck:
> Please say what report this affects!

The batch report in stock pdf

server/controllers/stock/reports/stock/lots_report.js
```diff
@@ -60,14 +61,16 @@ function stockLotsReport(req, res, next) {
...
+    'purchase_interval', 'tag_name', 'tracking_consumption', 'wac'];
 
   const dateKeys = ['min_stock_date', 'max_stock_date'];
 
   return Stock.getLotsDepot(null, options)
     .then((rows) => {
       rows.forEach(row => {
+        const current = new Date();
+        const delay = moment(new Date(row.expiration_date)).diff(current);
+        row.delay_expiration = moment.duration(delay).humanize(true);
```
jniles, line 73: If this is going to be printed, I would be careful about using `humanize()`.  We don't want an administrator to see an old report and think "this expires in two months" when the stock is already expired.  Always include the expiration date itself as well, for context.

Also, are we sure `humanize()` will respect the user's choice of language?

server/controllers/stock/reports/stock/lots_report.js
```diff
@@ -60,14 +61,16 @@ function stockLotsReport(req, res, next) {
...
-    'tracking_expiration', 'wac'];
+    'purchase_interval', 'tag_name', 'tracking_consumption', 'wac'];
 
   const dateKeys = ['min_stock_date', 'max_stock_date'];
 
   return Stock.getLotsDepot(null, options)
     .then((rows) => {
       rows.forEach(row => {
+        const current = new Date();
+        const delay = moment(new Date(row.expiration_date)).diff(current);
```
jniles, line 72: What is the difference between this `delay` and the one that is included in the `purgeKeys`?  Can we just use that one instead of rewriting the function?  

I worry that if we change the delay logic in the future that these reports will get out of sync.

server/controllers/stock/reports/stock/lots_report.js
Ebed-meleck: Ok I see I agree with you in the idea of keeping the context, but` humanize()` works very well with the user's choice of language

server/controllers/stock/reports/stock/lots_report.js
Ebed-meleck: The delay found in pureKeys is different because it is a decimal.

jmcameron (approved):
Looks good to me.
The report is okay.
All the tests pass on my dev system, including end-to-end tests.

Issue #7466, opened 2024-02-03
 Suggestion to modify the text sale responsible on the invoice


Ebed-meleck:
In the context of the hospital I find that it is better to change the sales responsible to responsible.

<img width="560" alt="Screen Shot 2024-02-03 at 10 26 56" src="https://github.com/IMA-WorldHealth/bhima/assets/63899331/447c163c-c3ce-4087-a4a1-4f444891eaf4">

to

<img width="765" alt="Screen Shot 2024-02-03 at 10 25 19" src="https://github.com/IMA-WorldHealth/bhima/assets/63899331/44a47c7e-a5ca-4019-b8d8-095d61ed8332">

Ebed-meleck:
Closed via #7467

PR #7467, opened 2024-02-03
Change text sale responsible to responsible

Ebed-meleck:
This change is for #7466.

Change Sales Manager to Manager


jmcameron:
Please say what page this affects!

Ebed-meleck:
> Please say what page this affects!

The receipt that is generated after patient billing

server/controllers/finance/reports/invoices/creditNote.handlebars
```diff
@@ -145,7 +145,7 @@
 
   <div class="row">
     <div class="col-xs-5">
-      {{translate 'FORM.LABELS.SALE_RESPONSIBLE'}} <br />
```
jniles, line 148: If `SALE_RESPONSIBLE` is no longer used in the application, could you remove it from the `i18n` JSON files?  I did a quick search and couldn't find it used anywhere else in the application.

Ebed-meleck: All right

jmcameron (approved):
Looks good.
I tested it locally and all tests pass including end-to-end tests.

PR #7468, opened 2024-02-05
Allow the value zero to be debited from the accounting entry for an invoice worth zero

### LINKED ISSUE #7463: Journal entry error when customer invoice has a value of zero
With the SEMI project, some hospitals do not have to charge customers for medicines, so the value of the invoice will be zero, and for these cases, we do not have the correct writings in the journal.

image

mbayopanda:
This PR updates the `CopyInvoiceToPostingJournal` procedure to allow the value zero to be debited from the accounting entry for an invoice worth zero

Close #7463


jmcameron (changes requested):
The change looks okay.
All the tests pass on my dev system (including all end-to-end tests).

However, it seems like the input field needs to be updated in 
`client/src/modules/invoices/templates/grid/unit_price.tmpl.html` or you cannot enter zero for unit costs in patient invoice entry form. For instance, if you created an inventory item for free service called "Free Surgery" with unit cost 0, you cannot create an invoice with that single item without updating this input field.

Also, it would very good to add an integration test for creating a patient invoice with zero total (in 1test/integration/patientInvoice.js`).   

In addition, it might be good to add a `@fifth_invoice` with zero balance to `test/data.sql` (around line540).  This will not test the modified SQL procedure, but it would verify adding items with zero balance in the `invoice` table.

jmcameron:
@mbayopanda Don't forget to do a `git pull` on your PR branch before updating it.

mbayopanda:
> The change looks okay. All the tests pass on my dev system (including all end-to-end tests).
> 
> However, it seems like the input field needs to be updated in `client/src/modules/invoices/templates/grid/unit_price.tmpl.html` or you cannot enter zero for unit costs in patient invoice entry form. For instance, if you created an inventory item for free service called "Free Surgery" with unit cost 0, you cannot create an invoice with that single item without updating this input field.
> 
> Also, it would very good to add an integration test for creating a patient invoice with zero total (in 1test/integration/patientInvoice.js`).
> 
> In addition, it might be good to add a `@fifth_invoice` with zero balance to `test/data.sql` (around line540). This will not test the modified SQL procedure, but it would verify adding items with zero balance in the `invoice` table.

@jmcameron The reason why we have this PR is that we are creating invoices with zero cost, everything is okay but in the posting journal we do not have a double entry, only one entry. This PR is just about to fix this behavior.

jmcameron:
> @jmcameron The reason why we have this PR is that we are creating invoices with zero cost, everything is okay but in the posting journal we do not have a double entry, only one entry. This PR is just about to fix this behavior.

I understand why you did it.  But when I tested this, I could NOT get a zero balance invoice to save because the client input field has a minimum of 0.001.  See my comment above starting with **"However, it seems like the input field needs to be updated**"...

jmcameron (approved):
We will deal with the input field issue in a separate issue.

PR #7469, opened 2024-02-05
Optionally display satisfaction rate by quantity delivered. 

mbayopanda:
This PR adds a way to display or not information about satisfaction rate by quantity delivered

image

And the summary: 

image


jmcameron (commented):
The code looks good.
All tests pass (including end-to-end tests).
I tested it and it seems to work.  But I have a couple of questions:
- In the report, there is no indication of what Depots are selected
- If you select more than one depot is the table separated for each depot or is it all merged into one table?

I did not remember how to generate non-zero data for test this.  But I looked back at the old PRs and found [this procedure](https://github.com/IMA-WorldHealth/bhima/pull/7040#issuecomment-1521730886).   Once these questions are answered, I'll try testing this PR again with this procedure.

mbayopanda:
@jmcameron you're right it's difficult to test when there is no data.

Depot names are displayed in sections of data, and if you have selected more than one depot, there is a kind of separator in data for each depot

jmcameron:
> @jmcameron you're right it's difficult to test when there is no data.
> 
> Depot names are displayed in sections of data, and if you have selected more than one depot, there is a kind of separator in data for each depot

@mbayopanda Thanks for clarifying that.

jmcameron (approved):
I tested this again.  Nice job, @mbayopanda !

PR #7478, opened 2024-02-11
fix: remove '...' from action column headers

### LINKED ISSUE #7477: Exclude "action" columns from the column selection modal
First reported [here](https://github.com/IMA-WorldHealth/bhima/pull/7386#pullrequestreview-1848784760).

The action column for several of our registries have "..." in the header, which makes them appear in the column selection modal.  This is a mistake as users may inadvertently hide the column and then be confused about where the action column went.  The solution is proposed in the reported issue

> In the voucher registry, the column header for the action menu should not appear in the Column selector. This is due to setting the displayName for the action column to '...' instead of '' (see below).  Please do a global search and replace to change displayName : '...' to 'displayName : ''`. I noticed 5 occurences for this and they are all for action items. Note that these are not a big deal in the main grid displays, but using '...' means that they get added to the Column selector, which does not make a lot of sense since you always want them, and a checkbox with ... beside it is not very informative!

jniles:
This commit removes the '...' from action column headers in ui-grids on registries.  This prevents them from appearing in column selection modals and users cannot hide them.

Closes #7477.

### DIFF
--- a/client/src/modules/functions/functions.js
+++ b/client/src/modules/functions/functions.js
@@ -36,7 +36,7 @@ function FunctionManagementController(Functions, Modals, Notify, uiGridConstants
     }, {
       field : 'action',
       width : 80,
-      displayName : '...',
+      displayName : '',
       cellTemplate : '/modules/functions/templates/action.tmpl.html',
       enableSorting : false,
       enableFiltering : false,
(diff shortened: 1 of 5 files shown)


jmcameron (approved):
Excellent.  Thanks!

PR #7480, opened 2024-02-11
Fix promise ohada account result 

### LINKED ISSUE #7479: OHADA income statement report crashes the application
`module Finance>Rapport>OHADA compte de résultat`

This error recurs in kanzala (production). After selecting the exercise, clicking on preview did not return the error message, I tried again, still the same thing. I open the pm2 log to see what message. What I realize, the application crashes then calls the other process (Luckily, pm2 was started in cluster mode, this is not at all remarkable) But locally the application crashes. I noticed that there was an error with the promise implementation. it retrieves information from the current fiscal year and if previous_fiscal_year_id is null it returns is not found, which is not handled very well with the promise. So I suggest a code refactoring with good promise management. So much so that when there is an error, the application does not crash.

production log image

WhatsApp Image 2024-02-10 at 18 54 27

local image log
Screen Shot 2024-02-11 at 20 40 43

Ebed-meleck:
`Finance Module > Report > OHADA Income Statement`

This commit fixes the handling of promise errors. to avoid crashing the application once an error occurs.

Closes #7479


jmcameron:
It was not clear how to reproduce the problem.  The issue says "After selecting the exercise", but I was not sure what that meant.   Please explain how to reproduce this fault using 'master'.   Thanks.

Ebed-meleck:
> It was not clear how to reproduce the problem. The issue says "After selecting the exercise", but I was not sure what that meant. Please explain how to reproduce this fault using 'master'. Thanks.

This problem reproduces like this: when you select the fiscal year of a given year, if it does not have a previous fiscal year, this error will recur. In the case of HGR kanzala or Tshikapa this year we have just installed so we do not have a previous exercise, and this report does not occur and crashes the application. You can try even with vanga on the first year exercise, you will get this error.

jmcameron:
Ebed, could you fix Issue https://github.com/IMA-WorldHealth/bhima/issues/7487 in this PR too?

server/controllers/finance/reports/ohada_profit_loss/index.js
```diff
@@ -174,129 +174,128 @@ profitLossTable.forEach(item => {
  * @param {*} options the report options
  * @param {*} session the session
  */
-function reporting(options, session) {
+async function reporting(options, session) {
```
jmcameron, line 177: 👍

server/controllers/finance/reports/ohada_profit_loss/index.js
```diff
@@ -306,7 +305,7 @@ function reporting(options, session) {
 function document(req, res, next) {
   reporting(req.query, req.session)
     .then(result => {
-      res.set(result.header).send(result.report);
+      res.set(result.headers).send(result.report);
```
jmcameron, line 308: 👍  Good catch.

jmcameron (commented):
The code looks good.  
I only did minimal testing through the GUI.  It seems to work.  
I ran the normal and end-to-end tests on my dev system and they all passed.

One possible improvement.  If the query for fiscal year data fails, perhaps it would be more helpful to generate a more descriptive error message (via `throw new BadRequest...`)?

However, I'm still seeing $NaN.undefined in the OHADA Bilan report.  Could that be fixed in this PR too?  This could be addressed in a separate PR, but since both issues are for OHADA reports, they could be done together.

jmcameron:
Thanks for fixing the NaN issue.

PR #7484, opened 2024-02-13
Patient billing Change the select to Selection with filter

### LINKED ISSUE #7483: Patient billing, change select to selection with filter
On the patient billing page, when selecting a service, if a list is long, it forces users to scroll through it. Therefore, I propose to change the `<select>` to selection with filter. This will help users instead of scrolling we can directly filter the list by the name we are looking for

Ebed-meleck:
Patient billing

This change made is for this #7483

And I added the debtor group in the patient information popup.
This will help the biller see which group the patient belongs to.

Because some approved (HGR TPA) do not go to the counter, we just give them the invoice to present in other services.

Closes #7483


jmcameron (approved):
The code changes look good.  Thanks Ebed!
I tested it with the GUI and it works as expected.
I ran the tests locally and all tests pass, including end-to-end tests.

PR #7486, opened 2024-02-14
Budget title subtotals

### LINKED ISSUE #7441: Add budget totals in title accounts
Compute and display subtotals for title accounts for both budget items and actuals. Totals Title accounts in the import CSV will continue to be ignored.

NOTE: This will only change the display on the Budget Page and will be recomputed each time the budget page is displayed/redisplayed.

jmcameron:
Compute the budget, actuals, etc for all title accounts.  Note that these are computed on the fly and are based entirely on child expense and income accounts, not child title accounts. 

Closes https://github.com/IMA-WorldHealth/bhima/issues/7441

*TESTING*
- Use bhima_test in your dev environment (do `npm run build; npm run build:db' first)
- If you do not see `Budget Management` under `Finances`, update your role permissions.
- In `Finances > BudgetManagement [MENU]' execute `Expor tYTD Budget Data (CSV)`
- Using Excel or a text editor, edit the CSV file and add budget amounts for various income and expense accounts.
   - Be sure to add some budget data to account `70111010 - Vente Médicaments en comprimes`
   - Any data entered in title accounts will be ignored
- Run BHIMA again
   - When you go to the `Finance > Budget Management` page, you will see the accounts but no budget or actuals data
   - To add Actuals expenses to help verify the totals do this
      - Go to `Finance > Simple Vouchers`
      - For Transaction Type, select "Purchases"
      - Add a description
      - For Debit, select account 60310014
      - For Credit, select account 31110011
      - Set an amount (such as 100)
      - Click on [Submit]
   - go to `Finance > Journal`
      - Change the Search to "All Time"
      - Select all transactions
      - Click on the 'Trial Balance' button (near the right end of button row at the top)
      - A confirmation dialog will appear, click on [Submit]
   - Return to the `Finance > Budget Management` page.  Now you will see account 7011010 has some actuals data
   - In the the `Finance > Budget Management [MENU]`, execute `Import Budget CSV file` and select the CSV file you modified earlier.   When it is imported, you should see data for title accounts (where the data is no zero). 
   - Click on the top-right collapse button to show only the top-level accounts.   This should make it simple to see if overall computations are correct.  Here is what this looks like for me (probably with different budget numbers that you will use; also I fixed missing items in the summary row after this screenshot): 
image
   - With all rows expended, verify the title computations.  Here is what the first part of the income section looks like for me: 
image
   - Once you have budget and actuals data, verify that the CSV output and the PDF report are correct.

- It would probably be a good idea to test this with a production database.


jmcameron:
I'm still working on this.  It turned out to be more complex than I expected.

lomamech (commented):
image
it really works very, well, I was curious to know how the class 8 account behaves: because in OHADA accounting, the class 8 account such as 81, 83, 85, 87, 89 (Are expense accounts ) and 82, 84, 86, 88 (are profit accounts)

image

​

lomamech:
I don't really have many comments, but only would it be possible to display the expense accounts without the minus sign, just because if we budget a class 82 account (Profit account) the account displays of class 8 does not carry out the Profit - Charge subtraction but seems to carry out an addition

image

but also there is the presence of the +- sign

lomamech (commented):
I just used the module and the results are correct,
here are my concerns
- the calculation of the total value regarding it as a class 8 title: should be the difference between profits - expenses

image

To verify this, you will need to add class 8 profit accounts for example.
- 82000000 : Production de session d'immobilier barrage Hydro Electrique
- 82000001 : Produits des sessions Bâtiments Administrative
I also notice negative values in certain places
image

lomamech:
@jmcameron let me know if negative values are normal and the Difference YTD column is obtained by doing the operation FY Actuals (YTD) - Budget (YTD)

jmcameron:
> @jmcameron let me know if negative values are normal and the Difference YTD column is obtained by doing the operation FY Actuals (YTD) - Budget (YTD)

Yes.  See line 783 of server/controllers/finance/budget/index.js:
   `acct.differenceYTD = acct.actuals && acct.actuals !== 0 ? acct.actuals - budgetYTD : null;`

lomamech (approved):
It was really an excellent idea not to display the main title of the class 8 accounts
image

PR #7491, opened 2024-02-16
Improvement(Payroll Indice Configuration)

### LINKED ISSUE #7454: Users report 5% pay code increase does not work
At Vanga, they recently paid their employees this year.  Several potentially relevant pieces of information:

  1. It is a new fiscal year
  2. They added ~8 employees to the organisation
  3. They had two employees that were absent the entire month, and attempted to model this by putting in negative "jours supplementaire" instead of just putting 0 days worked.

The last point caused havoc, as you might imagine (https://github.com/IMA-WorldHealth/bhima/issues/7452). 

However, we also received this feedback from the Vanga team:

> The other problem was the 5% pay code increase did not show up on the calculation.  I recommended we just pay today and work with the BHIMA team to see if this can be corrected prior to payday in February.

I believe that the 5% yearly increase is supposed to happen automatically.  We should investigate why it is not being applied and ensure that it isn't a bug in BHIMA.  We can probably improve our user interface so that users know if it is being applied or not as well.

lomamech:
- Get the data related to the minimum monetary unit and the percentage fixed bonus directly from the database and not from the session

closes #7454

### DIFF
--- a/server/controllers/payroll/multiplePayrollIndice/parameters.config.js
+++ b/server/controllers/payroll/multiplePayrollIndice/parameters.config.js
@@ -37,10 +37,24 @@ async function create(req, res, next) {
 
   const transaction = db.transaction();
 
-  // FIXME(@jniles) - refresh these from the database rather
-  // than obtaining them for the session information
-  const minMonentaryUnit = req.session.enterprise.min_monentary_unit;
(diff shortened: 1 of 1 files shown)


lomamech:
During my investigation, I did not find the cause of the malfunction of the increase in the base index

jniles (approved):
This is exactly the fix I was hoping for.  Looks good to me.

PR #7492, opened 2024-02-16
improvement(restric negative values)

### LINKED ISSUE #7452: Do not accept negative values for payroll indices
Vanga really tied themselves in knots by putting in negative values to indice fields.  We should add a client-side rule to block the ability to put negative values in the indice fields.

image

lomamech:
- Added the minimum value property to the bhTextImput component to restrict the season of negative values (less than Zero)

closes #7452


jniles (approved):
LGTM.

PR #7494, opened 2024-02-19
bug(Payroll indice crashes)

### LINKED ISSUE #7453: Removing employees from payroll configuration crashes the multiple payroll indice
The payroll module breaks when an employee is configured and then removed from the configuration.  This is due to a logic bug in the multiple payroll grid.   Specifically, this code is affected:

https://github.com/IMA-WorldHealth/bhima/blob/8e534a66f3501295fe5d6e4580d336f447da2d0d/server/controllers/payroll/multiplePayrollIndice/index.js#L122-L125

Because we are looping through staged employees, if an employee is unstaged, the page crashes.

The proper solution for this is probably to remove all employees that are staged for payment in that pay period when we configure the pay period.  We would need to make sure those employees had not already been "configured" from the multiple indice page, but otherwise, it seems like the easiest solution is to just purge the employees whenever we are editing the list and write the new set of selected employees in.

lomamech:
- Resolved system crash when an employee is removed from the configuration

closes #7453


jmcameron:
Chris, SVP veuillez ajouter une note sur la manière de tester cela.

server/controllers/payroll/multiplePayrollIndice/index.js
```diff
@@ -120,8 +120,11 @@ async function lookUp(options) {
   // let get rubric values for each employee
   const stagePaymentsParams = employeeUuid ? [payConfigId, db.bid(employeeUuid)] : [payConfigId];
   const stagePayments = await db.exec(stagePaymentIndiceSql, stagePaymentsParams);
+
   stagePayments.forEach(stagePay => {
-    employeesMap[stagePay.employee_uuid].rubrics.push(stagePay);
+    if (employeesMap[stagePay.employee_uuid]) {
```
jniles, line 125: This will work around the crash, but won't the employee remain configured forever in the database? 

Is there some way to remove the employee id from the stage payments table?

lomamech:
> Chris, SVP veuillez ajouter une note sur la manière de tester cela.

To see the bug in the master branch
you will first have to create a new employee configuration which we will name **Test config employee**
Here is the procedure to follow:
 - Access the **Payroll Employees Configuration** modules in the large Human Resources menu
image

 - Click on the **Add Employee Configuration** button
image

- After creating the new configuration, you will need to select the employees who will be part of this configuration, by clicking on the action button and selecting the configure option and this will bring up the list of employees
​
image

- For example we will select three employees
image

- After this step you will need to create a new pay period, to create a new pay period, simply click on the button **Set up a paiement period** 

image

for the test we will create the pay period in the following way, you will need to select the configuration of the employees that we have just created

image

After creating the pay period, you will first need to configure the data of each employee, to do this you will need to choose the option **Multiple Payroll (index-based)** 

image

image

To configure the employees, simply click on the action button and choose the configure option, and for the test database, the sections to configure are as follows:
- Autres_profit,
- Jours_prestes,
- Jours_supplementaires

image
_**Here is an example configuration**_

Once we have finished configuring the employees, when we return to the employee configuration that we had created and we remove one of the employees, the **Multiple Payroll (index-based)** modules display a error message in master branch, This PR solves this problem

image

@jmcameron Here is the procedure to do the test

server/controllers/payroll/multiplePayrollIndice/index.js
lomamech: > This will work around the crash, but won't the employee remain configured forever in the database?
> 
> Is there some way to remove the employee id from the stage payments table?

To resolve this problem, I suggest doing a cleaning when we define the pay envelope

lomamech:
@jniles @jmcameron I need another review

jmcameron (approved):
This looks good.   
I tested this via the GUI and it works.      Chris, thanks for providing the detailed procedure to reproduce the error in 'master'.
I ran the regression tests on my local dev system and they all pass, including end-to-end tests.

jmcameron:
@jniles Please take a quick look at the code added to remove employees deleted from the configuration at https://github.com/IMA-WorldHealth/bhima/pull/7494/files#diff-b7dd1e570ce822af431a4282e17fa44ba4ca5ac6af7bb58a089960b5e688e3f1.  It looks okay to me, but I would like a second opinion.  Thanks.

jniles (approved):
Looks good to me as well. 👍

PR #7500, opened 2024-02-22
improvement(Payroll process)

### LINKED ISSUE #7459: Add a confirmation modal before employees are put on the payment waiting list
In our [payroll documentation](https://docs.bhi.ma/fr/payroll/payroll_multiple), we have the following warning:

>  La mise en attente dans la liste d’attente est une opération irréversible, et les transactions issues de cette opération sont enregistrées dans le journal de saisie à la date de la fin de la période de paie. 

This refers to the option shown below:

image

After clicking on "put on waiting list", there is no going back.  We should provide a modal confirmation that asks if the user is ready to confirm those employees for payment.  It should include, at minimum:
  1. The number of employees the user is adding to the list.
  3. The payment period being added to (e.g. January 2024)
  4. The total value being confirmed in the enterprise currency.

It would also be helpful to have this kind of modal for the "confirm for payment" as well, but I'm not sure how much of this information is available at confirmation time.

lomamech:
- Add confirmation modal before employees are put on the payment waiting list
- Update E2E test for payroll_process

closes #7459


lomamech:
How to Test:
- To carry out the test, you must access the module, **Multiple Payroll**, and choose one of the pay periods,
image

- After selecting the pay period and company currency (required for payroll)

The list of employees configured for the pay period appears
image

- The first step is to configure the employees for payment, to do this we will use the automatic configuration by selecting all the employees
- Just click on the Menu and click on the “Configure for payment” option
​
image

After configuring the employees, the next step is to select the employees for whom you will need to be placed in the payment waiting list, to do this simply select the employees click on the menu and select the option *put on the waiting list*

image

And when we click on the option a dialog box appears and asks for confirmation of the operation, on the dialog box we find information on the pay period, the number of employees selected and the total employees. net salary

image

jmcameron (approved):
The code looks good.
I tested this locally and all the regular and end-to-end tests pass.  This includes an E2E test that was modified to handle the confirmation dialog.
Although this PR failed in the Semaphore tests, I believe that was a sporadic problem downloading the puppeteer dependency, probably due to some issue with their server and not a problem with the PR itself.

PR #7515, opened 2024-02-28
feature(Employee)

### LINKED ISSUE #7236: Auto insert the base grade salary for employee when "Grade" is selected
When registering or editing an employee, when the grade is selected, the base salary for that grade should be inserted into the "Salary" field (unless it already has data).  In this screen shot, when a grade is selected, the Base Salary from that grade should be inserted in the "Basic Salary" field on the right.   Note that Grade Management shows a "Base Salary" for each grade, so this seems like the correct amount to insert into the "Basic Salary" field for new employees.    If the adminstrator has already added a basic salary, it should not be overridden.   And of course the Administrator can edit the field and change it as desired.  But it seems like the basic salary for that grade is the correct value to insert as a default.

image

Obviously, this could be edited before saving/updating the employee.

lomamech:
- auto insert the base grade when Grade is Selected

closes #7236

### DIFF
--- a/client/src/modules/employees/registration/employees.js
+++ b/client/src/modules/employees/registration/employees.js
@@ -15,6 +15,7 @@ function EmployeeController(Employees, Services, Grades, Functions, Titles, Cred
   const { saveAsEmployee } = $state.params;
 
   vm.enterprise = Session.enterprise;
+  vm.onSelectGrade = onSelectGrade;
 
   vm.isUpdating = $state.params.uuid;
   vm.updateEditLabel = vm.isUpdating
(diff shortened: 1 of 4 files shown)


lomamech:
image

client/src/modules/employees/registration/employees.js
```diff
@@ -90,6 +91,10 @@ function EmployeeController(Employees, Services, Grades, Functions, Titles, Cred
     employee.displayAge = moment().diff(employee.dob, 'years');
   }
 
+  function onSelectGrade(element) {
+    vm.employee.individual_salary = element.basic_salary;
```
jmcameron: I think you should only override this `.individual_salary` if it does not already have a value. 
Someone could insert the base salary before setting the grade, and then selecting the grade loose what was input.

jmcameron (approved):
This works fine in the GUI now.
Basic and employee-related end-to-end tests pass.

PR #7519, opened 2024-02-29
Allow 0 for unit prices in patient invoices

### LINKED ISSUE #7518: Fix unit price input field to allow zero
Fix the input field `client/src/modules/invoices/templates/grid/unit_price.tmpl.html` to allow 0 values.

From https://github.com/IMA-WorldHealth/bhima/pull/7468, note:

> it seems like the input field needs to be updated in
> client/src/modules/invoices/templates/grid/unit_price.tmpl.html or you cannot enter zero for unit costs in patient invoice entry form. For instance, if you created an inventory item for free service called "Free Surgery" with unit cost 0, you cannot create an invoice with that single item without updating this input field.

jmcameron:
Updates the patient invoice form to allow items with unit price 0.
Note that there are some legitimate reasons why one might want a unit price for a service (or product) provided to a patient to be zero.  This ensures that is possible.

Closes https://github.com/IMA-WorldHealth/bhima/issues/7518

*TESTING*
- Use bhima_test
- Start with `master`
- Go to `Finance > Patient Invoice`
   - Select a patient
   - Add an item (if one is not already shown)
   - Chose a diagnostic test such as: DDGT_GLUC5K-_0  Glucotest, kit 500nm, Unité
   - Set the unit price to 0
   - Notice that you cannot save the invoice because the unit price cannot be zero.
- Switch to a branch for this pull request
- Restart the node server:  `npm run dev`
- Retry the procedure above.  It should now work.
- Once the patient invoice is completed, you can see that invoice in the `Finance > Invoice Registry` page.  Also, an Invoice Registry report will show the new invoice correctly.


jmcameron:
I asked @jniles about this.  Here are few comments he made:
- After thinking about it, I believe it is a reasonable thing to do.  It's more a stylistic choice in accounting, though with some information loss.  But I don't think it is an invalid idea or an unreasonable functionality to support.
- My only question is - do we have the test coverage for it?
Well, and 2nd, do we have a way to let administrators know that the person at billing at a production site has silently changed the medicines for his friend to being worth $0?
- Finally, it would be nice to talk through some of the corner cases.  Can the entire bill be $0?  If so, is that considered an "unpaid invoice"?

jmcameron:
> I asked @jniles about this. Here are few comments he made:
> 
>     * After thinking about it, I believe it is a reasonable thing to do.  It's more a stylistic choice in accounting, though with some information loss.  But I don't think it is an invalid idea or an unreasonable functionality to support.
> 
>     * My only question is - do we have the test coverage for it?
>       Well, and 2nd, do we have a way to let administrators know that the person at billing at a production site has silently changed the medicines for his friend to being worth $0?
> 
>     * Finally, it would be nice to talk through some of the corner cases.  Can the entire bill be $0?  If so, is that considered an "unpaid invoice"?

@mbayopanda : Please give me some feedback based on JNiles comments above.  Should we go ahead on this PR?

mbayopanda:
* My only question is - do we have the test coverage for it? **_no, we do not have test for that_**

* Do we have a way to let administrators know that the person at billing at a production site has silently changed the medicines for his friend to being worth $0? **_the person at the billing cannot change the price and the all medicines costs 0$_**

* Finally, it would be nice to talk through some of the corner cases.  Can the entire bill be $0?  If so, is that considered an "unpaid invoice"? **_yes the entire bill can be $0 if the bill is only constituted by medicines, when the value of the invoice is $0, those invoices do not appear in the "Payment modules"_**

client/src/modules/invoices/templates/grid/unit_price.tmpl.html
```diff
@@ -7,7 +7,7 @@
   </div>
 
   <input
-    min="0.0001"
+    min="0"
```
mbayopanda, line 10: 👍🏽

mbayopanda (commented):
LGTM 👍🏽 

bors r+

jniles:
Thanks for passing along the responses.  I'll probably not be available to work on BHIMA much in the coming months.

At Vanga, the person making the invoice does have access to change the price.  I would be surprised if the same wasn't the case at IMCK.  It's just something to think about how we can prevent fraud and abuse.  

Because BHIMA tracks a lot of this kind of thing, Vanga was able to provide evidence to let four staff members go who were defrauding the institution.  Since this makes defrauding easier (the invoices will not show up as being unpaid in any report), it might be helpful to consider making it easier to detect when someone has changed the price of medications, rather than just the inventory log.

What is BHIMA's current policy on testing?  Are they optional?

jmcameron:
@jniles If you have time, please add an issue about how we could make this feature more secure?  For example, what kind of logging would be necessary?

BTW, as far as the possibility of defrauding by setting the unit price to 0:  I don't think this is that much different to setting the unit price to 1 Fc, which I think was possible before this change.

jniles:
The cost won't be different, but the reporting will be.  It would show up in the unpaid invoices report or log as a cash payment of 50FC (the minimum currency), which is a curious payment.

PR #7520, opened 2024-02-29
feature(Payroll Message error)

### LINKED ISSUE #7455: Provide better error messages for payroll errors
**Scenario**

We had a case where two employee configurations were incorrect, and Vanga attempted to pay payroll.  The system crashed with a generic error message (a NOT NULL error).  After locating the issue with the employee in the database logs, we fixed that employee, but the users reported that the system still crashed. 

In reality, we had fixed one employee, but were missing the second one.  Had the users realized this, they might have been able to continue working by fixing the second employee.

**Proposal**

I propose we need better error messages for payroll crashes.  We currently have everything in a database transaction, which is good, but we still need to keep some sort of log that to say "system crashed when processing employee `jniles`".  That would provide some guidance to the user that something is wrong with employee `jniles` configuration.  If we had this error message (instead of a `NOT NULL` error), there is a chance the payroll issues at Vanga would have worked themselves out yesterday without as much manual intervention from the team.

### LINKED ISSUE #7456: Ignore employees who worked 0 days in payroll calculation
As a second safeguard against https://github.com/IMA-WorldHealth/bhima/issues/7452, we could consider ignoring employees who did not work any days that month (e.g `totalDays` is `0`).  In an ideal world, these employees would never have been configured for payment.  But we have noticed that our users sometimes do configure employees for payment who did not work.

The way this would work is as follows:
  1. The first rubrique computed is `totalDays`.
  2. After the `totalDays` computation, we filter all employees for `totalDays > 0`.  All employees that have `totalDays <= 0` will be withheld from further calculations, even if they are stage for payment.
  3. After the payments have been made, we present a "payroll report" to the user that indicates that X employees could not be paid due to them having worked 0 days that month.

If this is a mistake, the user can return to the journal vouchers and create a specific transaction for paying the missing employees.

lomamech:
- Prevent entry into the database of items whose values are not defined
- Display of the name of the employee for whom the Payroll data configuration was incorrect

closes #7455
closes https://github.com/IMA-WorldHealth/bhima/issues/7456

### DIFF
--- a/client/src/modules/multiple_payroll_indice/modals/parameter.modal.js
+++ b/client/src/modules/multiple_payroll_indice/modals/parameter.modal.js
@@ -35,7 +35,7 @@ function MultiPayrollIndiceParamModalController(Notify, MultiplePayroll, Instanc
       .catch((error) => {
         if (error.status === 400) {
           Notify.errorMessage(error.data.code);
-          return vm.close(true);
+          vm.close(true);
         }
 
         Notify.handleError(error);
--- a/client/src/modules/multiple_payroll_indice/modals/parameter.modal.js
+++ b/client/src/modules/multiple_payroll_indice/modals/parameter.modal.js
@@ -28,7 +28,7 @@ function MultiPayrollIndiceParamModalController(Notify, MultiplePayroll, Instanc
   vm.submit = (form) => {
     if (form.$invalid) { return 0; }
     vm.param.lang = Languages.key;
-    return MultiplePayroll.parameters.create(vm.param).then((res) => {
+    return MultiplePayroll.parameters.create(vm.param).then(() => {
       Notify.success('FORM.INFO.OPERATION_SUCCESS');
       return vm.close(true);
     })
--- a/client/src/modules/multiple_payroll_indice/modals/parameter.modal.js
+++ b/client/src/modules/multiple_payroll_indice/modals/parameter.modal.js
@@ -18,7 +18,10 @@ function MultiPayrollIndiceParamModalController(Notify, MultiplePayroll, Instanc
   vm.onSelectPayrollPeriod = (payrollConfig) => {
     vm.param.payroll_configuration_id = payrollConfig.id;
     MultiplePayroll.parameters.read(payrollConfig.id).then(parameter => {
-      vm.param = parameter;
+      if (parameter.length) {
+        const [param] = parameter;
+        vm.param = param;
+      }
     });
   };
 
--- a/client/src/modules/multiple_payroll_indice/modals/parameter.modal.js
+++ b/client/src/modules/multiple_payroll_indice/modals/parameter.modal.js
@@ -32,14 +32,14 @@ function MultiPayrollIndiceParamModalController(Notify, MultiplePayroll, Instanc
       Notify.success('FORM.INFO.OPERATION_SUCCESS');
       return vm.close(true);
     })
-    .catch((error) => {
-      if (error.status === 400) {
-        Notify.errorMessage(error.data.code);
-        return vm.close(true);
-      }
+      .catch((error) => {
+        if (error.status === 400) {
+          Notify.errorMessage(error.data.code);
+          return vm.close(true);
+        }
 
-      Notify.handleError(error);
-    });
+        Notify.handleError(error);
+      });
   };
 
 }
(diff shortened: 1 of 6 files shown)


lomamech:
image

lomamech:
*How to take the test*

To be able to do the test, you first need a payment period, with the following specificity
image

after creating the pay period, you must configure the jobs for the pay period, for this you must access the module *Payroll Multiple Payroll (index-based)*

image

once you access this module, you must select the pay period by clicking on the Search button a dialog box will appear, to select the pay period, after validating, the list of employees to configure appears

image

- For the first employee: you must do a normal configuration by entering all the necessary information
image

-  For the second employee, we will not enter information summarizing the case where an employee who had not worked was configured, for this case it will be necessary not to enter information on the number of days worked

image

Once the configuration is complete, the next step is to define the pay envelope. To do this, simply click on the Menu button then Configure for payment

image

The step of defining the pay envelope takes place as follows,
- Select pay period
- Define the pay envelope
- Then the number of days worked (This element allows you to calculate the ratio of the basic index to the number of days worked)

image

After validation, the system must detect a malfunction and display the identity of the employee who has not been properly configured

image

jmcameron (commented):
The code looks good.
I verifieds works in the GUI.

However, I did see a potential problem.  When I opened the modal dialog to do the Multiple Payroll (index-based), the modal appeared correctly.  However, when i selected the Period of payment, to the newly created payroll period (and before I filled the other fields or presssed the [Submit] button, I see this error in the server
```
  app configureServer(): Server started in mode development on port 8080. +0ms
  errors #interceptor() NotFound [Error]
  errors     at /home/jmcameron/Bhima/bin/server/lib/db/index.js:127:17
  errors     at _fulfilled (/home/jmcameron/Bhima/node_modules/q/q.js:854:54)
  errors     at /home/jmcameron/Bhima/node_modules/q/q.js:883:30
  errors     at Promise.promise.promiseDispatch (/home/jmcameron/Bhima/node_modules/q/q.js:816:13)
  errors     at /home/jmcameron/Bhima/node_modules/q/q.js:624:44
  errors     at runSingle (/home/jmcameron/Bhima/node_modules/q/q.js:137:13)
  errors     at flush (/home/jmcameron/Bhima/node_modules/q/q.js:125:13)
  errors     at process.processTicksAndRejections (node:internal/process/task_queues:77:11) {
  errors   status: 404,
  errors   code: 'ERRORS.NOT_FOUND',
  errors   description: 'Expected record to contain a single record with id undefined, but 0 were found!'
  errors } +0ms
```

This makes me think that there is something wrong with `onSelectPayrollPeriod()` in `client/src/modules/multiple_payroll_indices/modals/parameter.modal.js` or MultipleIndicesPayrollServer function that that it calls (in `server/controllers/payroll/multiplePayrollIndice/parameters.config.js`.

But the presence of this warning does not seem to affect the correct operation of the feature.   However, I do see a minor issue:  The Notify.error() message is hidden by the modal dialog.  So you may want to catch the error in the client, close the modal, and rethrow it.

jmcameron:
Is this related to Issue https://github.com/IMA-WorldHealth/bhima/issues/7456 ?

lomamech:
> The code looks good. I verifieds works in the GUI.
> 
> However, I did see a potential problem. When I opened the modal dialog to do the Multiple Payroll (index-based), the modal appeared correctly. However, when i selected the Period of payment, to the newly created payroll period (and before I filled the other fields or presssed the [Submit] button, I see this error in the server
> 
> ```
>   app configureServer(): Server started in mode development on port 8080. +0ms
> ```

After investigating, I realize that this message appears because of the db.one query which is supposed to return data but does not return any data during the first use
image

lomamech:
> Is this related to Issue #7456 ?

You are right, both issues revolve around the same problem, if the first is resolved, this one should also be closed

client/src/modules/multiple_payroll_indice/modals/parameter.modal.js
```diff
@@ -18,17 +18,28 @@ function MultiPayrollIndiceParamModalController(Notify, MultiplePayroll, Instanc
...
+    vm.param.lang = Languages.key;
     return MultiplePayroll.parameters.create(vm.param).then(() => {
       Notify.success('FORM.INFO.OPERATION_SUCCESS');
       return vm.close(true);
     })
-      .catch(Notify.handleError);
+      .catch((error) => {
+        if (error.status === 400) {
+          Notify.errorMessage(error.data.code);
+          vm.close(true);
```
jmcameron, line 38: 👍 Thanks!

jmcameron (approved):
Works well.  Thanks!
The basic and End-to-End tests run okay in my dev system.

PR #7526, opened 2024-03-01
fix(Problem of deductions from salaries)

### LINKED ISSUE #7525: Problem of deductions from salaries above the net salary (bonus)
With the Payroll system based on pay indices, the pay envelope is defined and this envelope is broken down in relation to the ratios of the sum of the indices equivalent to each employee, but it happens that the institution is forced to withhold certain amounts at source such as: The salary advance, the salary deposit and various loans and this subtraction can cause situations where the total withholding is greater than the net salary, this issue aims to prevent the salary configuration with negative net payables.
​

lomamech:
- Red coloring of employees for whom the net salary is negative
- Added an error message reporting the case of employees for whom the net salary is negative

closes #7525

### DIFF
--- a/client/src/css/structure.css
+++ b/client/src/css/structure.css
@@ -918,6 +918,10 @@ a[disabled] {
   background-color: #ffb3b3 !important;
 }
 
+.negative {
+  background-color: #ffb3b3 !important;
+}
+
 .expired {
   background-color: #ff000083 !important;
 }
--- a/client/src/i18n/en/form.json
+++ b/client/src/i18n/en/form.json
@@ -1142,6 +1142,7 @@
             "ONLY_TWO_SECTORS":"Please select two sectors only",
             "ONLY_TWO_VILLAGES":"Please select two villages only",
             "ATTENTION_CONFIGURED": "Warning! Only employees waiting for configuration can be configured",
+            "ATTENTION_NEGATIVE_VALUE": "Warning, there are employees for whom the totality of their deductions is greater than their gross salary, please correct this problem​",
             "ATTENTION_PAYSLIPS": "Warning! There are no payslips records for employees waiting for configuration",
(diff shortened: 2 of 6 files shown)


lomamech:
*How to take the test*
1/ The first step is to modify the Payroll payroll rubric: Accompte sur salaire
image

You will need to modify this rubric so that it appears when configuring the indices
You will have to click on the salary deposit:
- Is an Index
- is a monetary value
- is to be entered

image

**2/ How to create a new topic configuration**
The new configuration of sections
- will be named: Config Withdrale Rubrics
image

After creating the configuration of the rubrics, you must select the rubrics which will be part of this configuration,
image

For our case we will select all the sections with indices while ensuring that "Accompte sur salaire" is indeed selected
image

**3/ Creating the new pay period**

image

**4/Configuring Indices** 
to be able to configure the pay indices, you must launch Payroll Multiple Payroll (index-based) and select the pay period TEST MARCH 2024 in the search
image

for the first employee you will need to configure the data as follows
image

for the second employee you will need to configure the data as follows
image

The configuration of the pay envelope is defined as follows
image

**5/Configuration for payment**
Configuration for payment: Payroll Multiple Payroll
After selecting the pay period:
image

Employee configuration will consist of selecting all employees
image

Click on the menu and choose the **configure for payment** option
image

After validation, the BHIMA system will display in red the employees for whom the net salary is negative.
image

image
Here is an extract from his pay slip, we clearly see that his gross salary is $34.13 and his deduction is $50.00

to resolve this problem the human resources department should reduce the amount to be withheld,

Here is the procedure,
Detect employees with negative net salaries, click on the action button and choose the configure option

image

The amount to be withheld must always be less than total Gross Salary

image

image

**The BHIMA system will now prevent employees with negative salaries from being placed on hold for payment**
image

jmcameron (approved):
The code looks good.
I tested this with the GUI and it works as described.

PR #7529, opened 2024-03-03
Fix rendering of installation docs

jniles:
Fixes an issue in the rendering of the installation documentation that causes the verify step to look strange.

Previously:

image


jmcameron (commented):
Good catch, thanks!

Would you please update the RELEASE.md based on our recent discussion about doing a release.   I think we still need to add a line regarding pushing it back to `upstream`, eg:

`git push --set-upstream upstream release-next`

And then manually merging the 'release-next` branch in the the upstream/master.

It seemed convenient to do this in this PR, but if you prefer, we could do this in another issue/PR.

PR #7537, opened 2024-03-07
Fix loading status in button on satisfaction rate report

Ebed-meleck:
This PR brings the fix to the loading status button in the rate satisfaction report.
`/reports/satisfaction_rate_report`


jmcameron (approved):
Looks good.
The updates work in the GUI.

PR #7586, opened 2024-05-17
Feature: deploy BHIMA using docker and docker-compose

### LINKED ISSUE #6252: Add documentation for docker
We need to add additional documentation now that we've merged the initial support for docker (https://github.com/IMA-WorldHealth/bhima/pull/6229).  

There are still some things that needs to done:

1. We need to have a better way of building the MySQL database. Currently, the user has to run yarn build:docker to create an initial database file to populate the database.
2. We should be using multi-stage builds to ensure that we only have build artifacts in a "build" image and then deploy the most minimal image one can think of.
3. We need documentation on how to deploy/run with docker & docker-compose.

jniles:
This PR adds a working docker-compose script, as well as a working BHIMA Dockerfile.  Currently, the BHIMA docker build is hosted at [thirdculturesoftware/bhima](https://hub.docker.com/repository/docker/thirdculturesoftware/bhima/) since I don't have permissions for this github repository anymore.

Note: if this PR is not wanted, it can be closed.  I plan to continue formalizing docker as the main deployment method on a fork of BHIMA.

Here are the major changes:

### Rewrite the BHIMA Dockerfile 
I rewrote the `Dockerfile` to use npm, and removed all mention of yarn.  This is in line with https://github.com/IMA-WorldHealth/bhima/issues/7194.  It also simplified the Dockerfile in the process.  There are still future improvements that could be made, such as using a multistage build and ensuring that all the chromium dependencies are truly still needed.  The current container for `bhima` is ~1GB in size.

The dockerfile also uses it's own `.env.docker` to set variables for the docker container that the user does not need to change.

### Rework docker-compose.yml
I ran into several bugs trying to get the docker-compose script to work.  For the time being, mysql is pinned to version 8.3 since versions 8.4 and higher are not supported by mysqljs.  The team should consider migrating to mysql2 for the BHIMA driver.  ~~It is rumored to be a drop in replacement.~~. Update: I have shown that we can drop `mysql2` into BHIMA without significant change in https://github.com/Third-Culture-Software/bhima/pull/5.

Many of the variables now have sensible defaults so the user does not have to set them in a `.env` file.  Obviously for real deployments, the database password should be generated, not reused from this script.  In the future, I'll likely work on generating the MySQL password and the `SESS_SECRET` during the `docker compose up` call.

To build the system with docker, I added installation instructions with docker [here](https://github.com/Third-Culture-Software/bhima/blob/2c5a026afee2073a879879d01013669fdce99f75/docs/en/for-developers/installing-bhima-with-docker.md).  Closes https://github.com/IMA-WorldHealth/bhima/issues/6252.

### Rename SQL scripts to `${number}-${scriptname}`
In order to build the initial database with docker compose, the files are copied into an initial mysql data directory, then executed in alphabetical order.  For that reason, I renamed all SQL scripts needed for an initial database to start with a number that specifies their execution order.  I went through and updated all references to these scripts and tested by running the repository test scripts. If I missed something, let me know!

### Trying it out
If you want to test this out, it should be as easy as checking out this branch and running `docker compose up`.  Note that it might take a while.  When the application is all settled, navigate to the port on your local system and you should be greeted by the clean BHIMA install.


docs/en/for-developers/installing-bhima-with-docker.md
```diff
@@ -0,0 +1,43 @@
+# Installing BHIMA with Docker
+
+Using Docker and Docker Compose provides a simpler alternative to the traditional [Linux installation method](./installing-bhima.md).  To install Docker and Docker Compose, follow the instructions on [the official Docker website](https://docs.docker.com/engine/install/).
+
+### System Requirements
+
+Currently, BHIMA is compatible only with the x64 architecture. If you are using other architectures (such as ARM64 or x86), please refer to alternative installation instructions.
+
```
jmcameron, line 8: It may seem obvious, but it is probably a good idea to mention that you need to install 'docker' on your system to do any of the following docker operations!

test/server-unit/bhMoment.spec.js
```diff
@@ -19,7 +19,7 @@ function evaluate(startDate, endDate, dateBhMoment) {
   expect(formated).to.deep.equal(expected);
 }
 
-describe('test/sever-unit/bhMoment', () => {
```
jmcameron, line 22: 👍 Good catch.

jmcameron (changes requested):
The code changes look fine and will not interfere with normal Bhima installs.
Thanks for adding the Docker documentation.  
I tested the normal test procedure on my local system and it ran without any problems.
I did not try running with the Docker setup, but I had tried that before on another system.
Other than a minor update to the Docker documentation, I think this is ready to go.

jniles:
Changes made!

jmcameron (approved):
Thanks

PR #7590, opened 2024-05-23
Fix: replace mysqljs with mysql2

jniles:
This PR replaces `mysql` with `mysql2`.   Learn more about some of the benefits of using `mysql2` [here](https://sidorares.github.io/node-mysql2/docs/history-and-why-mysq2).

The reason I am making this change is because [MySQL 8.4](https://dev.mysql.com/doc/relnotes/mysql/8.4/en/news-8-4-0.html) completely removed compatibility for the default authentication method we were using previously (mysql_native_password).  If you tried to start the server with MySQL 8.4, you would receive this warning:

> unknown variable 'default-authentication-plugin=mysql_native_password'

The `mysql` npm package is de facto deprecated in lieu of the `mysql2` npm package. There has been an outstanding [pull request](https://github.com/mysqljs/mysql/pull/2233) to add in the new authentication methods, but it doesn't look like this has any chance of being merged anytime soon.

Note that, for docker, the plugin has been removed, but a new flag has been added `--mysql-native-password=ON` to enable the mysql-native password authentication.  We'll likely need to switch off this in a future version to their SHA2 caching plugin.  However, I'm not sure what that would involve at this point.  Using this flag is a stopgap while we figure that out.


jmcameron (approved):
The code changes look very reasonable.  And I feel that is an issue that needs to be addressed soon, so I'm happy to accept this PR.

I tested this branch on my local test setup and it works without any issues.  All regular and end-to-end tests pass.

I did not test the parts of this related to running with Docker compose.

PR #7593, opened 2024-05-24
build: replace yarn with npm

### LINKED ISSUE #7194: Switch from yarn to npm
When BHIMA started, we needed lockfiles to give predictive deployments, and `yarn` had them before `npm`.  Now, `npm` has them and so much more!  Additionally, yarn and npm have diverged in their handling of plug and play modules, and you need npm to install the latest yarn.

I don't see a need to yarn anymore.  It adds additional complexity (one more thing to install and maintain) and provides little added value if any.  I propose that we migrate from yarn to npm by:
  1. Removing yarn.lock and committing package-lock.json
  2. Updating our documentation that refers to yarn.
  3. Updating our tests to only install npm and not yarn.
  4. Updating our installation + release process to use npm instead of yarn.

jniles:
Changes documentation, commands, and procedures to work with `npm` instead of `yarn`.  This should simplify our installation procedures and remove an additional dependency that is no longer required.

Closes #7194.


jmcameron (approved):
I tested this and works fine.

PR #7594, opened 2024-05-24
chore(store): replace store with Map

### LINKED ISSUE #7387: Replace the Store() with a Javascript Map()
The first lines of the BHIMA `Store` starts with:

https://github.com/IMA-WorldHealth/bhima/blob/9e66f548538674242f0043f5a74ef7676e9a3e97/client/src/js/services/Store.js#L4-L10

The `Store` was lifted almost in its entirety from the [DojoToolkit](https://dojotoolkit.org/)'s `MemoryStore` implemented back in 2010.  Javascript has evolved since then, and built its own native version of a Store - the Map/WeakMap.  Indeed, the [MDN documentation for WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#why_weakmap) states:

> A map API could be implemented in JavaScript with two arrays (one for keys, one for values) shared by the four API methods. Setting elements on this map would involve pushing a key and value onto the end of each of those arrays simultaneously.

This is almost exactly what the BHIMA store does, except we use an object for our index, not an array.  Because of this, and the fact that we need to be able to iterate over the contents of a store, we should use a `Map()` instead of a `WeakMap()`.

I propose we keep all the methods of Store, and keep the Store service, but transparently replace the implementation with `Map()`.  The test should still pass, and no other code outside of the Store service will need to be changed.  It will be a silent upgrade to one of the oldest modules in the BHIMA library.

jniles:
This commit removes the previous `Store.js` internals and replaces them with a simple JavaScript `Map()` while retaining the same API.  It is likely more efficient than the previous version.

Closes #7387.


jmcameron (changes requested):
The idea of this PR is good and the code changes look fine.
However, when I tested this on my local server, I got the following errors when running client unit tests:
image

jniles:
Yep, this is an issue.  We have a few services that $watch `store.data`, which isn't expected behavior from the service API standpoint.  That dependency might even be a deal breaker for this change, not sure.  I'll look into it more after my comprehensive exams.

PR #7662, opened 2024-07-11
Import payroll configuration from excel

### LINKED ISSUE #7655: Importing Payroll Configurations from an Excel File
This task aims to allow the human resources department or personnel management to configure payroll elements directly from an Excel file.

The template for this Excel file must first be imported from the BHIMA system and correspond to the selected pay period. This module has several advantages, including reducing the error rate and eliminating double checking.

lomamech:
- Implementation of the functionality allowing the import of the configuration made from the csv file

closes https://github.com/IMA-WorldHealth/bhima/issues/7655


lomamech:
**Detailed Explanation**

1. **Enable Index Payment System Mode:**

- Navigate to Payroll Settings.
- Enable the Index Payment System mode.

2. **Modify Payroll Rubrics:**

Go to Payroll Rubric Management.
- Select and edit the rubrics **ACOMPTE SUR SALAIRES** and **FRAIS SCOLARITÉ.**
- Check the options This rubric **is an index** and This index **is to be entered.**

3. **Configure Rubrics in Payroll Rubric Configuration**:
   - Navigate to `Payroll Rubric Configuration`.
   - Edit `CONFIGURATION DES RUBRIQUES (INDICES)` and select `ACOMPTE SUR SALAIRES` and `FRAIS SCOLARITÉ`.
 
image


4. **Create New Pay Period**:
   - Create a new pay period for July 2024 by going to `Pay Period Configuration`.
image

5. **Access Multiple Payroll (Index-based) Module**:
   - Navigate to the `Indexes` folder.
   - Access the `Multiple Payroll (index-based)` module.
   - Select the newly created pay period to view a grid with the configured employees.
image

6. **Download Template File**:
   - Click on the menu and select the `Download template file` option.
   - This will download a CSV file with the columns and employees for the selected pay period.
image

7. **Upload the Filled Template File**:
   - Return to BHIMA and click on the menu again.
   - This time, select the `Upload payroll configuration` option to upload the filled template file.

image

image

 After selecting the CSV file, the configuration will be automatically processed and applied successfully.


image

lomamech:
@jmcameron @jniles You can review this PR

jmcameron (changes requested):
The code looks okay.
All regular and end-to-end tests pass.
When I did step 5, I noticed that search options indicators were not be handled correctly:
image

I had problems on step 6.  When I edited the downloaded template file, it looked like this
image

When I tried uploading it, I got this error:
image

Please rename the downloaded template file from "Download the template file 7_12_2024.csv" to something like "payroll_configuration_template_7_12_2024.csv".  The downloaded filename should be a noun, not a command.

Please delete the yarn.lock file from your branch, and use 'npm to install the dependencies'.  We no longer use yarn for installing dependencies.

Thanks

client/src/modules/multiple_payroll_indice/multiple_payroll_indice.js
```diff
@@ -276,5 +279,14 @@ function MultiplePayrollIndiceController(
     // return  serialized options
     return $httpParamSerializer(options);
   };
+
+  function importConfigCSV() {
+    MultiplePayroll.importConfigModal({ payroll_configuration_id : vm.payrollConfigurationId }).then(result => {
```
jniles: Should this have some sort of error handling?

server/controllers/payroll/multiplePayrollIndice/parameters.config.js
```diff
@@ -479,5 +481,183 @@ function stagePaymentIndice(payrollConfigurationId) {
...
+      const errMsg = `${i18n(lang)('ERRORS.MISSING_UPLOAD_FILES')}`;
+
+      throw new BadRequest('Expected at least one file upload but did not receive any files.',
+        errMsg);
+    }
+    const filePath = req.files[0].path;
+
+    const data = await util.formatCsvToJson(filePath);
+
+    data.forEach(item => {
```
jniles: You might consider using a `Array.prototype.map()` here.

```js
const arrayDataFormated = data.map(item => {

  // ... process as needed ...
  
  return dataElement;
});
```

server/controllers/payroll/multiplePayrollIndice/parameters.config.js
```diff
@@ -479,5 +481,183 @@ function stagePaymentIndice(payrollConfigurationId) {
...
+    `;
+
+    const transaction = db.transaction();
+
+    transaction
+      .addQuery(sqlGetRubs, [payrollConfigId])
+      .addQuery(sqlGetEmployees, [payrollConfigId])
+      .addQuery(sqlGetEditableRubrics, [payrollConfigId]);
+
+    transaction.execute()
```
jniles: You are already using `async/await` in this function, go ahead and use it here.  This code would then read:

```
try {
const record = await transaction.execute();

// etc 
} catch (e) {
  next(e)
}
```

server/controllers/payroll/multiplePayrollIndice/parameters.config.js
```diff
@@ -479,5 +481,183 @@ function stagePaymentIndice(payrollConfigurationId) {
...
+      .then((record) => {
+        const rubPayroll = record[0];
```
jniles: It might be nice to use destructing here so that we know what properties are in `record`. 

```js
transaction.execute()
  .then(([rubPayroll, empPayroll, rubPayrollConfig]) => {
    const checkColumnFormated = Object.keys(arrayDataFormated[0]).length;
    // etc...
  });
```

server/controllers/payroll/multiplePayrollIndice/parameters.config.js
```diff
@@ -479,5 +481,183 @@ function stagePaymentIndice(payrollConfigurationId) {
 
 }
 
+/**
+ * Import Payroll Configuration for a payroll configuration
```
jniles, line 485: This code looks good overall!

One thing that would help a lot in debugging production bugs is using the `debug` module.  Could you import `debug`, and give it a name like:
```js
cont debug = require('debug')('payroll:import');
```

And then use it to indicate what is happening in the loop?

For example, you could do something like:
```js
debug(`Received a CSV file with ${data.length} rows.`);

// later:

let checkValidEmployee = 0;
empPayroll.forEach(emp => {
  arrayDataFormated.forEach(dt => {
    if (emp.employee_display_name.toLowerCase() === dt.employee.toLowerCase()) {
       debug(`Configuring payroll for ${emp.employee_display_name}.`);
       checkValidEmployee++;
    }
  });
});
```

If you add in a few of these debug messages, particularly where you are looping through records, we will be able to diagnose production issues much easier.

server/controllers/payroll/multiplePayrollIndice/parameters.config.js
```diff
@@ -479,5 +481,183 @@ function stagePaymentIndice(payrollConfigurationId) {
...
+            stagePaymentIndiceData.payroll_configuration_id = payrollConfigId;
+            stagePaymentIndiceData.currency_id = req.session.enterprise.currency_id;
+            stagePaymentIndiceData.rubric_id = rub.id;
+            stagePaymentIndiceData.rubric_value = rubricValue;
+
+            transactionOperation.addQuery(insertStagePaymentIndice, [stagePaymentIndiceData]);
+          });
+        });
+
+        transactionOperation.execute()
```
jniles: This could also be `async/await` to clean up this `.then()` series of loops.

server/controllers/payroll/multiplePayrollIndice/parameters.config.js
```diff
@@ -479,5 +481,183 @@ function stagePaymentIndice(payrollConfigurationId) {
...
+          .then(() => {
+            res.status(200).json();
```
jniles: If you aren't returning any data, you should use `res.sendStatus(204)`, since HTTP 204 indicates [No Content](https://developer.mozilla.org/fr/docs/Web/HTTP/Status/204).

jniles (commented):
Great work on this @lomamech!  I added some clean up code comments.  The only two that I care about are:
 1. Adding in the debug module (this will be incredibly helpful when things break in production as they always do).
 2. Using consistent async/await.

I haven't run the code to test it yet, but I see that @jmcameron gave you that review.

lomamech:
> I noticed that search options indicators were not be handled correctly:

This error often appears randomly, I will try to reproduce it and see how to correct it

server/controllers/payroll/multiplePayrollIndice/parameters.config.js
lomamech: > One thing that would help a lot in debugging production bugs is using the `debug` module. Could you import `debug`, and give it a name like:

I don't really understand what I have to do in relation to debugging.

lomamech: @jniles I need an explanation

jmcameron (approved):
Normal and end-to-end tests pass.
Tested per directions and it works fine.
Ready for merger after rebasing/squashing commits.

PR #7668, opened 2024-07-22
Minor improvements to the Annual Clients Report.

### LINKED ISSUE #6170: Updates for the Annual Clients Report
Finance > Report > Annual Clients Report

Bug: Watch the javascript console when opening the report request page on the client-side.  I see errors (on master) related to line 47 of bhCronEmailReport.js.

**Suggestions**
- Make the subtitle fiscal year of the report a little larger.  Perhaps make it:  Fiscal 2021
- Since "Include Cash Clients" is an option, perhaps when that option is selected, add an extra column indicating what type of client each row is.

jmcameron:
This PR implements the following changes

- Minor improvement to the Debtor Group Management page
- Minor improvements to the Annual Clients Report

*TESTING* 

- Use the bhima_test or a production database for testing this part
- Debtor Group Management page
   - Go to Administration > Debtor Group Management
      - Edit one debtor group and add a note
      - The note will be shown for for each debtor group that has one

- Annual Clients Report
   - Use a production database to test this part
   - Go to the Finances > Report > Annual Clients Report page
   - Select various options and click on the [Preview] button
   - Try the various combinations of the check boxes
     table
	 
        | Hide Locked Clients | Include Cash Clients | Cash Col? | Locked Col? |
        | --------------------|----------------------|-----------|-------------|
        |  NO                 | No                   | No        | No*         |
        |  NO                 | Yes                  | Shown     | No*         |
        |  Yes                | No                   | No        | No          |
        |  Yes                | Yes                  | Show      | No          |
	 
	 * If you chose "NO" for the "Hide locked clients" option,
       note that if no locked accounts are shown, it will print
	   a message about that at the bottom of the page.

   - To see the normal "Locked" go the the Administrion > Debtor Groups
     Management page and lock the "NGO IMA World Health" group
   - Repeat the verification process and note that a "Locked" 
     column will appear if the locked debtor groups are shown and
	 no message will appear at the end since there is a locked 
	 debtor group.

        | Hide Locked Clients | Include Cash Clients | Cash Col? | Locked Col? |
        | --------------------|----------------------|-----------|-------------|
        |  NO                 | No                   | No        | Shown       |
        |  NO                 | Yes                  | Shown     | Shown       |
        |  Yes                | No                   | No        | No          |
        |  Yes                | Yes                  | Shown     | No          |

Closes https://github.com/IMA-WorldHealth/bhima/issues/6170


server/controllers/finance/reports/debtors/annual_clients_report.js
```diff
@@ -33,8 +34,26 @@ async function setupAnnualClientsReport(options, enterprise) {
     getTotalsFooter(fiscalYear.id, currencyId, rate, shouldHideLockedClients, shouldIncludeCashClients),
   ]);
 
+  // Computations for optional columns
+  const numLocked = rows.reduce((sum, row) => sum + (row.locked ? 1 : 0), 0);
```
jniles, line 38: Nice use of reduce 👍

server/controllers/finance/reports/debtors/annual_clients_report.js
```diff
@@ -92,6 +111,7 @@ function getDebtorGroupMovements(fiscalYearId, currencyId, rate, hideLockedClien
   const excludeCashClientsCondition = 'AND dg.is_convention = 1 ';
   const sql = `
     SELECT ac.number AS accountNumber, dg.name AS groupName,
+      dg.locked, dg.is_convention AS cashClient,
```
jniles: This flag feels backwards to me.  If the `is_convention` is ticked, shouldn't that indicate it is not a cash client?  Where does the inversion happen?

EDIT: Nevermind, I see that it is an if/unless statement.  Let's make this either `isConvention` or `IF(dg.is_convention, 0, 1) as cashClient` to clear up the confusion.

server/controllers/finance/reports/debtors/annual_clients_report.handlebars
```diff
@@ -16,39 +37,56 @@
...
         </thead>
         <tbody>
           <!-- print a row for each debtor account-->
           {{#each rows as | row |}}
             <tr>
               <td>{{row.accountNumber}}</td>
-              <td>{{row.groupName }}</td>
+              <td>{{row.groupName}}</td>
+              {{#if ../cashClients}}
+              <td class="center">{{#if row.cashClient}}{{else}}<strong>X</strong>{{/if}}</td>
```
jniles: It took me a while to parse that this was an inversion - blank if `row.cashClient` and X otherwise.  Let's bring this naming in line with what it actually intends, which I relayed in a previous comment.

jniles (changes requested):
The code is good, and the logic is sound.  Thanks for the detailed PR!

I worry that `cashClient` is confusing, since `cashClient` is actually a boolean indicating `nonCashClient`.  Could you clean up that naming as suggested below?  While it does no harm here, I can see it tripping up another consumer of this API.

server/controllers/finance/reports/debtors/annual_clients_report.handlebars
jmcameron: I fixed the naming issue and I think it is clearer.  But I think that is_convention is a poor name to start with, but I will not fix that in this PR since it appears in multiple places and I'm not sure of a better naming.

jmcameron:
@jniles Please review this again.  I fixed the naming issue and it is less  confusing, but not perfect.   But I'd appreciate you double-checking that the displays work correctly.   It seems okay to me, but a second check would be very helpful!

server/controllers/finance/reports/debtors/annual_clients_report.js
```diff
@@ -91,7 +110,8 @@ function getDebtorGroupMovements(fiscalYearId, currencyId, rate, hideLockedClien
   const hiddenClientsCondition = ' AND dg.locked = 0 ';
   const excludeCashClientsCondition = 'AND dg.is_convention = 1 ';
   const sql = `
-    SELECT ac.number AS accountNumber, dg.name AS groupName,
+    SELECT act.number AS accountNumber, dg.name AS groupName,
+      act.locked, dg.is_convention AS isConvention,
```
jniles: In the context of this report, I believe this should be `dg.locked`, not `act.locked`.  I expected the "locked" function to be on the client, not the account.

Arguably, one could that locking a debtor group should lock their account, but that isn't how BHIMA currently (or historically) has worked.  Locking the debtor groups just prevents debtors from being billed in it, not blocking transactions to the account.  I think this makes sense on the surface (maybe you still want to be able to annul debts, but don't want any debtors to join the group or anyone else to transact on the group).

jniles (changes requested):
@jmcameron the code is still good.

I believe that the "locked' should apply to a debtor group, not the account, and proposed that revision in the code.  Otherwise, it works as described.

server/controllers/finance/reports/debtors/annual_clients_report.js
jmcameron: I have fixed this and retested.

PR #7684, opened 2024-09-07
feature(Budget Report)

### LINKED ISSUE #7683: Budget Report
The current Budget Management and Configuration module requires enhancement to improve its analytical capabilities https://github.com/IMA-WorldHealth/bhima/issues/7676. After a thorough review of the existing budget report, it has been determined that adding new columns is necessary to provide users with more comprehensive data insights.

Proposed Enhancement:

To support more detailed budget tracking and enable comparative analysis across multiple fiscal years, a new budget report needs to be developed. This report will allow users to:

Monitor Budget Performance: The new report will include additional columns that provide deeper insights into budget performance indicators, such as variances, trends, and forecasts.

Conduct Comparative Analysis: The report will feature comparative analysis capabilities, allowing users to easily compare current fiscal data with that of previous years. This will help in identifying patterns, making informed decisions, and improving future budget planning.

Enhanced Configuration Options: The budget management interface will be updated to support configuration for these new reporting requirements, making it easier for users to customize and filter data according to their needs.

lomamech:
- Implementation of the budget report, allowing the selection of the fiscal year and the number of previous years for result analysis.
- Retrieving the values for the periods, excluding period 0, the closing period 13, as well as hidden and blocked accounts.
- Correction of the restriction during the compilation of securities accounts for non-budgeted accounts.

closes #7683

### DIFF
--- a/client/src/modules/reports/generate/budget_report/budget_report.config.js
+++ b/client/src/modules/reports/generate/budget_report/budget_report.config.js
@@ -1,0 +1,88 @@
+angular.module('bhima.controllers')
+  .controller('budget_reportController', BudgetReportController);
+
+BudgetReportController.$inject = [
+  '$sce', 'NotifyService', 'BaseReportService', 'AppCache', 'reportData', '$state', 'SessionService',
+];
+
+function BudgetReportController($sce, Notify, SavedReports, AppCache, reportData, $state, Session) {
+  const vm = this;
+  const cache = new AppCache('configure_budget_report');
+  const reportUrl = 'reports/finance/budget_report';
+
+  vm.reportDetails = {
+    currency_id : Session.enterprise.currency_id,
+    set_number_year : 1,
+    filter : 'default',
+  };
+
+  vm.previewGenerated = false;
+  checkCachedConfiguration();
+
+  vm.onSelectFiscalYear = (fiscalYear) => {
+    vm.reportDetails.fiscal_id = fiscalYear.id;
+  };
+
+  vm.onSelectCurrency = (currency) => {
+    vm.reportDetails.currency_id = currency.id;
+  };
+
+  vm.onSelectCronReport = report => {
+    vm.reportDetails = angular.copy(report);
+  };
+
+  vm.numberYears = [
+    { id : 1 }, { id : 2 }, { id : 3 }, { id : 4 }, { id : 5 },
+  ];
+
+  vm.preview = function preview(form) {
+    if (form.$invalid) { return null; }
+
+    // update cached configuration
+    cache.reportDetails = angular.copy(vm.reportDetails);
+
+    return SavedReports.requestPreview(reportUrl, reportData.id, angular.copy(vm.reportDetails))
+      .then((result) => {
+        vm.previewGenerated = true;
+        vm.previewResult = $sce.trustAsHtml(result);
+      })
+      .catch(Notify.handleError);
+  };
+
+  vm.clearPreview = function clearPreview() {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.clear = (value) => {
+    delete vm.reportDetails[value];
+  };
+
+  vm.filterBudget = [
+    { value : 'default', label : 'REPORT.BUDGET_REPORT.DISPLAY_ALL_ACCOUNTS' },
+    { value : 'hide_title', label : 'REPORT.BUDGET_REPORT.HIDE_TITLE_ACCOUNT' },
+    { value : 'show_title', label : 'REPORT.BUDGET_REPORT.SHOW_ONLY_TITLE_ACCOUNT' },
+  ];
+
+  vm.requestSaveAs = function requestSaveAs() {
+
+    const options = {
+      url : reportUrl,
+      report : reportData,
+      reportOptions : angular.copy(vm.reportDetails),
+    };
+
+    return SavedReports.saveAsModal(options)
+      .then(() => {
+        $state.go('reportsBase.reportsArchive', { key : options.report.report_key });
+      })
+      .catch(Notify.handleError);
+  };
+
+  function checkCachedConfiguration() {
+    if (cache.reportDetails) {
+      vm.reportDetails = angular.copy(cache.reportDetails);
+    }
+    vm.reportDetails.type = 1;
+  }
+}
--- a/client/src/modules/reports/generate/budget_report/budget_report.html
+++ b/client/src/modules/reports/generate/budget_report/budget_report.html
@@ -1,0 +1,96 @@
+<bh-report-preview
+  ng-if="ReportConfigCtrl.previewGenerated"
+  source-document="ReportConfigCtrl.previewResult"
+  on-clear-callback="ReportConfigCtrl.clearPreview()"
+  on-save-callback="ReportConfigCtrl.requestSaveAs()">
+</bh-report-preview>
+
+<div ng-show="!ReportConfigCtrl.previewGenerated">
+  <div class="row">
+    <div class="col-md-12">
+      <h3 class="text-capitalize" translate>REPORT.BUDGET_REPORT.TITLE</h3>
+      <p class="text-info" translate>REPORT.BUDGET_REPORT.DESCRIPTION</p>
+    </div>
+  </div>
+
+  <div class="row" style="margin-top : 10px">
+    <div class="col-md-6">
+      <div class="panel panel-default">
+        <div class="panel-heading">
+          <span translate>REPORT.UTIL.OPTIONS</span>
+        </div>
+
+        <div class="panel-body">
+          <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate>
+            <bh-fiscal-year-select
+              fiscal-id="ReportConfigCtrl.reportDetails.fiscal_id"
+              on-select-fiscal-callback="ReportConfigCtrl.onSelectFiscalYear(fiscalYear)"
+              required="true">
+            </bh-fiscal-year-select>
+
+            <div class="form-group" ng-class="{'has-error' : ConfigForm.set_number_year.$invalid && ConfigForm.$submitted}">
+              <label class="control-label" translate>REPORT.BUDGET_REPORT.SET_NUMBER_YEAR</label>
+              <bh-clear on-clear="ReportConfigCtrl.clear('id')"></bh-clear>
+              <ui-select
+                name="set_number_year"
+                ng-model="ReportConfigCtrl.reportDetails.set_number_year">
+        
+                <ui-select-match placeholder="{{ 'REPORT.BUDGET_REPORT.MAX_5_YEAR' | translate }}">
+                  <span>{{$select.selected.id}}</span>
+                </ui-select-match>
+        
+                <ui-select-choices repeat="year.id as year in ReportConfigCtrl.numberYears | filter:{id: $select.search}">
+                  <strong ng-bind-html="year.id | highlight:$select.search"></strong>
+                </ui-select-choices>
+              </ui-select>
+              <div class="help-block" ng-messages="ConfigForm.id.$error" ng-show="ConfigForm.$submitted">
+                <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+              </div>
+            </div>
+
+            <div class="form-group" ng-class="{ 'has-error' : ConfigForm.$submitted && ConfigForm.budgetFilter.$invalid }">
+              <div ng-repeat="budgetFilter in ReportConfigCtrl.filterBudget" class="radio">
+                <label>
+                  <input
+                    name="filter"
+                    type="radio"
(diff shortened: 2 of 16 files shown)


server/models/06-bhima.sql
```diff
@@ -177,9 +177,10 @@ INSERT INTO unit VALUES
   (316, 'Detailed record of purchases','TREE.PURCHASE_REGISTRY_DETAILED','The purchase registry detailed',154,'/purchases/detailed'),
   (317, 'Satisfaction Rate Report','TREE.SATISFACTION_RATE_REPORT','Satisfaction Rate Report',282,'/reports/satisfaction_rate_report'),
   (318, 'Job Titles Management','TREE.TITLE','',57, '/titles'),
-  (319, 'Budget Management', 'TREE.BUDGET', '', 5, '/budget');
+  (319, 'Budget Management', 'TREE.BUDGET', '', 5, '/budget'),
+  (320, 'Budget Report','TREE.BUDGET_REPORT','',281,'/reports/budget');
```
jniles: I believe this url should correspond with the report key in the reports table and the URL bar.  For example:

https://github.com/IMA-WorldHealth/bhima/pull/7684/files#diff-2cf4bd0b9cd21b1ec14877dfaabf9ea79b6bf4e3d20dbecb408fc2d7829575e8R11

So you should either change the `path` here to be `/reports/budget_report` or change the `budget_report.*` files to simply be `budget`.  Otherwise, it will not load for me.  I had to update the path manually to be able to test this:

```sql
update unit set `path` = "/reports/budget_report" where id = 320;
```

server/models/migrations/next/migrate.sql
```diff
@@ -5,4 +5,12 @@
...
-ALTER TABLE `debtor_group` ADD COLUMN `is_non_client_debtor_groups` TINYINT(1) NOT NULL DEFAULT 0 AFTER `is_insolvent`;
\ No newline at end of file
+ALTER TABLE `debtor_group` ADD COLUMN `is_non_client_debtor_groups` TINYINT(1) NOT NULL DEFAULT 0 AFTER `is_insolvent`;
+
+/*
+ * @author: lomamech
+ * @date: 2024-09-02
+ * @description: Budget Report #7683
+ */
+INSERT INTO unit values (320, 'Budget Report','TREE.BUDGET_REPORT','',281,'/reports/client_debts');
```
jniles: This `path` should correspond to the `path` in the `units` table in `01-schema.sql`.

client/src/modules/reports/generate/budget_report/budget_report.html
```diff
@@ -0,0 +1,92 @@
...
+                ng-model="ReportConfigCtrl.reportDetails.set_number_year"
+                autocomplete="off"
+                required="true">
+              <div class="help-block" ng-messages="ConfigForm.set_number_year.$error" data-date-error ng-if="ConfigForm.$submitted && ConfigForm.set_number_year.$invalid">
+                <span translate> REPORT.BUDGET_REPORT.MAX_5_YEAR </span>
+              </div>
+            </div>
+
+            <div class="form-group" ng-class="{ 'has-error' : ConfigForm.$submitted && ConfigForm.budgetFilter.$invalid }">
+              <div ng-repeat="budgetFilter in ReportConfigCtrl.filterBudget" class="radio-vertical">
```
jniles: There is no such class as "radio-veritcal" in BHIMA.  It should just be "radio".

client/src/modules/reports/generate/budget_report/budget_report.html
```diff
@@ -0,0 +1,92 @@
...
+          <form name="ConfigForm" bh-submit="ReportConfigCtrl.preview(ConfigForm)" novalidate>
+            <bh-fiscal-year-select
+              fiscal-id="ReportConfigCtrl.reportDetails.fiscal_id"
+              on-select-fiscal-callback="ReportConfigCtrl.onSelectFiscalYear(fiscalYear)"
+              required="true">
+            </bh-fiscal-year-select>
+
+            <div class="form-group" ng-class="{'has-error' : ConfigForm.set_number_year.$invalid && ConfigForm.$submitted}">
+              <label class="col-md-6 control-label" translate>REPORT.BUDGET_REPORT.SET_NUMBER_YEAR</label>
+              <input
```
jniles: This `<input>` would be better as a `<select>` with five options.  Like this:

image

The reason is that (1) you will always have a default case (say, 1), and the user won't do something silly like put 5000 years in.  You have the form logic to catch this case, but it feels silly allow the user to even try.

client/src/modules/reports/generate/budget_report/budget_report.config.js
```diff
@@ -0,0 +1,78 @@
...
+      .catch(Notify.handleError);
+  };
+
+  vm.clearPreview = function clearPreview() {
+    vm.previewGenerated = false;
+    vm.previewResult = null;
+  };
+
+  vm.filterBudget = [
+    { value : 'default', label : 'REPORT.BUDGET_REPORT.DEFAULT_SETTING' },
```
jniles: This value should be checked by default, if it is actually the default value. 

Also, what is the default setting?  Maybe call it "Display all accounts".

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -0,0 +1,235 @@
...
+
+    const fiscalYearNumber = fiscalsYear[0].year || '';
+    const lastFiscalYearNumber = fiscalsYear[1].year || '';
+
+    const dataFiscalsYear = await q.all(transaction);
+    let tabFiscalReport = [];
+    const uniqueSet = new Set();
+
+    dataFiscalsYear.forEach(subArray => {
+      subArray.forEach(item => {
```
jniles: Instead of doing a loop within a loop, you might be able to use [array.flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat).  I think it would improve readability.  Like this:

```js
dataFiscalsYear.flat().forEach(item => {
  // process each item
});
```

jniles: Another improvement would be to change the `if (!uniqueSet.has(key)) {` to just return early if it doesn't have the key.

```js
const key = `${item.id}-${item.number}`;
if (!uniqueSet.has(key)) {
  uniqueSet.add(key);
  const isTitle = (item.type_id === 6);
}

// ... becomes ...

const key = `${item.id}-${item.number}`;
if (uniqueSet.has(key)) { return ; }
uniqueSet.add(key);

const isTitle = (item.type_id === 6);
// etc...
```

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -0,0 +1,235 @@
...
+        const key = `${item.id}-${item.number}`;
+        if (!uniqueSet.has(key)) {
+          uniqueSet.add(key);
+          const isTitle = (item.type_id === 6);
+
+          if (item.type_id) {
```
jniles: Why would an account not have a `type_id`?  Or are you just filtering out the root account?

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -0,0 +1,235 @@
...
+        LIMIT ?
+    `;
+
+    const fiscalsYear = await db.exec(sqlGetPreviousFiscalYear, [fiscalYearId, setNumberYear]);
+    const reporting = new ReportManager(BUDGET_REPORT_TEMPLATE, req.session, optionReport);
+
+    fiscalsYear.forEach((fisc, index) => {
+      transaction.push(Budget.buildBudgetData(fisc.id));
+
+      if (index > 0) {
```
jniles, line 63: Nice trick!

server/controllers/finance/reports/budget_analytical/report.handlebars
```diff
@@ -0,0 +1,181 @@
...
+
+            <th class="text-right">{{currency totalVariationExpense currencyId}}</th>
+            <th class="text-right">{{percentage totalCompletionExpense}}</th>
+          </tr>          
+        </tbody>
+      </table>
+
+
+
+      {{!-- <table class="table table-striped table-condensed table-report table-bordered">
```
jniles: What is this?

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -0,0 +1,235 @@
...
+                label : item.label,
+                type_id : item.type_id,
+                isTitle,
+              });
+          }
+        }
+      });
+    });
+
+    dataFiscalsYear.forEach((fiscal, index) => {
```
jniles, line 96: I am not sure why you are treating index === 0 as different.  I would advise you to just break it into two different functions/blocks though, for clarity:

```js
const [firstYear, ...laterYears] = dataFiscalsYear;

tabFiscalreport.forEach(rep => {
  if (firstYear.number === rep.number) {
     // do whatever 
   } 
}

// later:
tabFiscalreport.forEach(rep => {
  let totalIncomeRealised = 0;
  let totalExpenseRealised = 0;
   /// etc
 })
```
This saves you from a big `.forEach()` loop that just calls `if` first thing.

client/src/modules/reports/generate/budget_report/budget_report.config.js
```diff
@@ -0,0 +1,78 @@
...
+    vm.reportDetails = angular.copy(report);
+  };
+
+  vm.preview = function preview(form) {
+    if (form.$invalid) { return null; }
+
+    // update cached configuration
+    cache.reportDetails = angular.copy(vm.reportDetails);
+
+    return SavedReports.requestPreview(reportUrl, reportData.id, angular.copy(vm.reportDetails))
```
jniles, line 44: This error handling doesn't seem to work if there is no budget for that year.  I tried it for the test dataset with FY 2019, and it failed with this error, but there was not signal given on the client.

image

It should probably let the user know that no budget was set for that fiscal year.

jniles (changes requested):
@lomamech, I wasn't able to fully test this as there are some problems in the code.  I've made suggestions below to improve the code clarity and the data.

Let me know when you have made the changes and please provide an example year that has data in it that I can check.

server/controllers/finance/reports/budget_analytical/report.handlebars
lomamech: It's a saving error, I had deleted but unfortunately the saving of the file had not taken effect

server/controllers/finance/reports/budget_analytical/index.js
lomamech: I didn't quite understand what I should do, please give more guidance.

client/src/modules/reports/generate/budget_report/budget_report.config.js
lomamech: > This error handling doesn't seem to work if there is no budget for that year. I tried it for the test dataset with FY 2019, and it failed with this error, but there was not signal given on the client.
> 
> image
> 
> It should probably let the user know that no budget was set for that fiscal year.

image
@jniles  **_I can't reproduce the error, I used the test database by first running ````npm run build:db````_**

lomamech: image

server/controllers/finance/reports/budget_analytical/report.handlebars
jmcameron: This obsolete table still needs to be removed.

jmcameron (changes requested):
I did some basic testing and it seems to work.  
However, before merging, please address the issues that @jniles brought up in his review.
Also update to the latest dependencies in master.
Also, a minor cosmetic thing.  
- In the budget report, please center or right justify the "Realization" in the column headers.
- In Englsh, "Realization" should be "Actuals"
- In English, "Revenus" should be "Income"

lomamech:
> * Income
it's done
image

server/controllers/finance/reports/budget_analytical/index.js
jmcameron: > I am not sure why you are treating index === 0 as different. I would advise you to just break it into two different functions/blocks though, for clarity:

FYI If you check the SQL definition of the `budget` table, it indicates that period==0 is for the FY budget total for the account.

client/src/modules/reports/generate/budget_report/budget_report.config.js
jmcameron: This seems to be resolved.   I tested several FY's and they all worked.  If there was no budget, 0 was used for all items.

server/controllers/finance/reports/budget_analytical/index.js
jmcameron: After discussion with @jniles, I suggest we leave cleaning this up for a later PR (@jniles will create the PR).

jmcameron (approved):
Except for the coding issue that @jniles pointed out, I think this PR is ready to merge.  I've tested it in the browser and it seems to work.  

There are some improvements I would still like to see to the report, but I will add Issues for them so we can do them later.

The tests all pass locally.

PR #7695, opened 2024-09-13
feat(reports): use a datauri for enterprise logo

### LINKED ISSUE #7679: Bug: Report renderer uses fixed substring for logopath, breaking custom logo paths.
The current report manager uses a fixed number (`7`) to parse the logo string.  IMCK uses a custom path for its logo, so it caused intermittent crashing when the reports could not load the logo into the reports.  Here is the code in question:

https://github.com/IMA-WorldHealth/bhima/blob/fd3b7e2d397a3e22546778a8e05038bed4795bc5/server/lib/ReportManager.js#L160-L165

And here is the error caused:
image

This file path should be `bhima-uploads/pics/...` but is truncated by the `substring()` call.  We should figure out a better solution for this code.

I can see two alternatives:
 1. We treat the logo path as absolute (e.g. it must start with `/home/bhima/example/pics/` or `C:\Users\bhima\example\pics`) and simply inject the logo directly into the HTML file.  This might require updating the `uploader` code.
 2. We treat the logo path as relative and merge with the application's current path. 

In both cases, we should probably check that the logo path is well defined using [`fs.access()`](https://nodejs.org/api/fs.html#fspromisesaccesspath-mode`).  For example:

```js
const { access, constants } = require('fs/promises');

try {
  await access(metadata.enterprise.logo, constants.R_OK)
  // logo is readable, proceed to render as normal.
} except (e) {
  // logo is not readable, use a fallback
  metadata.enterprise.logo = '/some/default/logo/that/ships/with/bhima'
  // OR
  // delete metadata.enterprise.logo
} 

await render(template, options);
```

jniles:
This commit uses a datauri encoding for the enterprise logo.  It fixes the logo rendering issue called by the gulp v5 update which created encoding issues during the build step.  This was reported by @lomamech in #7679.

Secondarily, it ensures the enterprise logo exists before using it, as well as standardizing the paths to be absolute paths.  For ease of updates, a new partial has been added for point-of-sale receipts to avoid repeat code usage.  I've tested this both with a regular logo, and with an uploaded logo.  Both seem to work.

Closes #7679.


jniles:

server/lib/ReportManager.js
```diff
@@ -157,11 +158,20 @@ class ReportManager {
     // sanitise save report option
     this.options.saveReport = Boolean(Number(this.options.saveReport));
 
-    if (metadata.enterprise && metadata.enterprise.logo) {
-      // use dynamic path for logo ie client doesn't need absolute path but pdf need it
-      // so we use {{absolutePath}} which add or not absolute path part into logo
-      // remove the client/ part from logo because it will be injected by {{absolutePath}}
-      metadata.enterprise.logopath = metadata.enterprise.logo.substring(7);
+    // check if the enterprise has a logo, and normalize the path to absolute
+    // Use the IMA application icon as the default logo if nothing else exists
```
jmcameron, line 162: 👍

jmcameron (approved):
Thanks for deleting output.pdf.  Not sure how it got in there!
The code looks good.
Thanks for the new partial 'point-of-sale-logo'.  That simplifies the code!
All tests pass.
The only change I suggest is showing the current (or default) logo on the Enterprises settings page.

PR #7725, opened 2024-09-20
Update puppeteer and coral

jmcameron:
This PR updates puppeteer and coral.

The puppeteer API has changed; its PDF functions return a Uint8Array instead of a Buffer.
This PR also changes the BHIMA pdf render function (renderPDF() in server/lib/renderers/pdf.js) to return a Buffer so that all BHIMA functionality remains unchanged.

**TESTING:**
- Make sure normal tests work
- Verify that generating PDF documents works (eg, reports, etc)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.33.0",
       "license": "GPL-2.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


jniles (approved):
Downloaded and tested this.  Everything works as expected.

PR #7745, opened 2024-10-03
Add Feature to Multiply Pay Indices by Frequency for Employee grade

lomamech:
Add Feature to Multiply Pay Indices by Frequency for Employee Categorization
- Added the `is_linked_to_grade` property in the `rubrics` table.
- Created the `rubric_grade_indice` table to store rubric configurations related to grades.
- Implemented functionality to define frequency for each rubric, which will be multiplied by the value corresponding to the employee's grade.


lomamech:
**TO TEST A PULL REQUEST (PR)**
In order to test this pull request, it will be necessary to create new rubrics that will be linked to the employee grades for configuration.

lomamech:
The first step is to create new rubrics in the Human Resources/Payroll Rubric Management module.
image

lomamech:
The context for these rubrics is as follows: we are in a situation where a special bonus is configured by grade. For grade G1, the value is 2, for grade G2, the value is 5, and for grade G3, the value is 10. During the payroll period, the payroll manager will enter the number or frequency corresponding to the payroll month, and the system will assign to the rubric the product of the configured value for the grade and the frequency.

First, "Enable index payment system" in Human Resources > Payroll Settings

To create rubrics linked to grades, simply click on the **"Add a Payroll Rubric"** button, then enter the following fields: 
- Designation and Abbreviation. 
- select that the rubric is an index
- check the option indicating that the rubric is not a monetary value.
- check the option "This index is to be entered", 
-then select "This rubric is linked to grade". Finally, choose the index type "This rubric is other profit".
image

lomamech:
You can thus create one or more rubrics linked to employee grades.

CONFIGURATION OF GRADE-LINKED RUBRICS

The configuration of grade-linked rubrics allows payroll parameters to be adjusted according to the hierarchical levels of employees. Each grade can be assigned a specific value for rubrics such as bonuses or other benefits. These rubrics enable flexible and automated management of payroll elements, ensuring better accuracy in salary calculations.

lomamech:
To proceed with the configuration, simply access the Indexes/Staffing Indices Management modules.
image

lomamech:
Here is the main interface. To access the configuration interface, simply click on the menu.
image

image

The grade-linked rubrics appear in the menu with a book icon. Simply click on it to access the configuration form.

lomamech:
The configuration form displays the heading Settings for Calculating Rubrics Based on Employee Grades, followed by the name of the rubric to be configured.
image
There is a field to select the grade and enter the corresponding value. It should be noted that the system restricts the configuration to multiple grades with different values. To modify a grade, simply delete it and recreate it.

lomamech:
Here is the interface with several grades configured.
image

lomamech:
**USE CASES**
For the use case, we will reuse the existing rubric configurations. To do this, we will go to the Human Resources / Payroll Rubric Configuration module and modify the configuration of the rubrics (indices) by adding the new rubrics linked to grades.

image

lomamech:
Next, it will be necessary to check the new rubrics linked to the grade.
image

lomamech:
It is now necessary to create a new payroll period with the following parameters: **Human Resources / Pay Period Configuration**
image

image

lomamech:
**The final step** involves configuring the payroll. To do this, simply go to the **Human Resources / Indexes** Multiple Payroll (index-based) module, and in the interface, select the newly created payroll period along with the list of configured employees.

image
To configure an employee, simply click on the Action area, then on the Configure button. The configuration form will then appear.
image

lomamech:
The configuration interface allows you to enter values for the configured rubrics. For the grade-linked rubrics, there are three areas: the first is for the frequency (editable), the second indicates the value that has been configured for the rubric, and the last shows the product (frequency multiplied by the value).

image

This new feature has been added to assist the payroll manager by allowing them to perform certain calculations outside the BHIMA system.

lomamech:
@jniles @jmcameron  I have just completed this latest PR with instructions on how to perform the test. If possible, you can proceed with the code review.

server/controllers/payroll/staffingIndices/gradeIndice.js
```diff
@@ -16,6 +25,33 @@ function list(req, res, next) {
     }).catch(next);
 }
 
+function listRubrics(req, res, next) {
```
jniles: This function name is confusing, since it doesn't list all rubrics.  Maybe you can rename it `listRubricsLinkedToGrade`?

Also, why can't this just be a query string parameter to the `list()` function?  Why do you need to make a whole other route?

I would assume that `GET /staffing_grade_indices/?is_linked_to_grade=1` should work, right?

server/controllers/payroll/staffingIndices/gradeIndice.js
```diff
@@ -94,3 +130,91 @@ function lookUp(options = {}) {
...
+  db.exec(sql, id)
+    .then(() => {
+      res.sendStatus(200);
+    })
+    .catch(next);
+}
+
+function lookUpRubricsGrade(options = {}) {
+  const sql = `
+    SELECT HEX(rgi.uuid) as uuid, rgi.rubric_id, rgi.value, HEX(rgi.grade_uuid) as grade_uuid,
```
jniles: You probably want `BUID()` here, not `HEX()`.

server/models/01-schema.sql
```diff
@@ -2569,6 +2570,20 @@ CREATE TABLE `staffing_grade_indice` (
...
+  `uuid` BINARY(16) NOT NULL,
+  `value`  DECIMAL(19,4) NOT NULL,
+  `grade_uuid` BINARY(16) NOT NULL,
+  `rubric_id` INT(10) UNSIGNED NOT NULL,
+  PRIMARY KEY (`uuid`),
+  KEY `rubric_id` (`rubric_id`),
+  KEY `grade_uuid` (`grade_uuid`),
+  CONSTRAINT `rubric_grade_indice__grade` FOREIGN KEY (`grade_uuid`) REFERENCES `grade` (`uuid`),
+  CONSTRAINT `rubric_grade_indice__rubric_payroll` FOREIGN KEY (`rubric_id`) REFERENCES `rubric_payroll` (`id`),
+  UNIQUE KEY `rubric_grade_indice_1` (`grade_uuid`, `rubric_id`)
```
jniles, line 2584: If I understand your explanation correctly, the relationship between grades and rubrics is many-to-one.  That is, multiple grades can have the same rubric, but each grade can only have a single rubric.

If that is true, do we really need a whole other table for this?  Why can't we make it part of the `grade` table?

jniles (commented):
@lomamech, I've asked some clarifying questions of the code.

In your explanation for testing, you say this:

> To create rubrics linked to grades, simply click on the "Add a Payroll Rubric" button, then enter the following fields:
>  Designation and Abbreviation.
    select that the rubric is an index
    check the option indicating that the rubric is not a monetary value.
    check the option "This index is to be entered",
    -then select "This rubric is linked to grade". Finally, choose the index type "This rubric is other profit".
    
    
I am curious: can I create a rubric linked to a grade that does not respect those rules?  For example, can I make it a monetary value?  Can I make it an different kind of index type?

I'm just trying to understand the impact if the user makes a mistake in configuring this, since it is not at all obvious.  

Personally, I would expect a bonus related to the grade to be configured on the grade page, not on the rubrics page.  I would also expect that if these options can only be applied in a certain way, we shouldn't let the user make a mistake by mis-applying the options.

lomamech:
> I am curious: can I create a rubric linked to a grade that does not respect those rules? For example, can I make it a monetary value? Can I make it an different kind of index type?
> 
> I'm just trying to understand the impact if the user makes a mistake in configuring this, since it is not at all obvious.
> 
> Personally, I would expect a bonus related to the grade to be configured on the grade page, not on the rubrics page. I would also expect that if these options can only be applied in a certain way, we shouldn't let the user make a mistake by mis-applying the options.

I completely understand you. It would have been possible, on another configuration interface besides the one for creating payroll items, to select indexed, non-monetary items with an index type such as "this item is another benefit". Yes, you're right, configuring payroll items is indeed a delicate task. Personally, I opted for the quickest solution to implement.


In my opinion, the ideal solution would be a new modeling of the payroll module to merge the old system with the index-based national system. Additionally, it would be relevant to introduce a hybrid system where some employees would be paid based on an index, while others would continue to receive a fixed salary. The current payroll module is the result of several implementation periods: 2014-2015, 2017, 2019, and 2022. A complete overhaul would be the best option.

server/models/01-schema.sql
lomamech: I’m afraid not. Regardless of the case, it will always be necessary to have another table because for the same grade, there will be specific values for multiple items. For instance, grade A1 will have a value for overtime, another for call bonus, and yet another for holiday pay. This results in a 1.N and 1.N cardinality, hence the need to create a new table. The restriction I’ve applied with UNIQUE ensures that there are no duplicate grade-item pairs, such as 
````{grade: 'A1', item: 'night', value: 0.2} and {grade: 'A1', item: 'night', value: 2}.````

File: server/controllers/payroll/staffingIndices/gradeIndice.js (hunk 1 above)
Comment (lomamech) [reply server/controllers/payroll/staffingIndices/gradeIndice.js]: Good catch, i have delete it and use Rubrics.read(null, { is_linked_to_grade : 1 }) and it works well

**Comment by jniles** [comment]
@lomamech would you be able to make an issue with some proposals about how to overhaul the payroll systems?

**Comment by lomamech** [comment]
> would you be able to make an issue with some proposals about how to overhaul the payroll systems?

All right

File: server/models/migrations/next/migrate.sql [hunk 4]
```diff
@@ -13,6 +13,7 @@ ALTER TABLE `debtor_group` ADD COLUMN `is_non_client_debtor_groups` TINYINT(1) N
  * @description: Budget Report #7683
  */
 INSERT INTO unit values (320, 'Budget Report','TREE.BUDGET_REPORT','',281,'/reports/budget_report');
+<<<<<<< HEAD
```
jniles: @lomamech , looks like this is a broken merge.   Please remove the merge related code.

client/src/modules/multiple_payroll_indice/modals/config.modal.html
```diff
@@ -21,16 +21,51 @@
...
+              type="number"
+              min="0"
+              class="form-control"
+              ng-model="ConfigPaiementModalCtrl.frequencyRubrics[rubConfigured.id]"
+              ng-keyup="ConfigPaiementModalCtrl.onUpdateRubricValue(rubConfigured.id)"
+              ng-required="true"
+              ng-key>
+          </div>
 
+          <bh-input-text
```
jniles, line 37: Why are there three separate inputs here?  What do they correspond to? 

I see the text of https://github.com/IMA-WorldHealth/bhima/pull/7745#issuecomment-2391665924 that there are three separate inputs for bonus by grade.  Would it be possible to explain what they do and what impact they have to the user?  Maybe in a help message under the input?

I stared at it for a while, but I couldn't understand the difference between "bonus_by_grade", "frequency: bonus_by_grade" and "value: bonus_by_grade".

lomamech: > Why are there three separate inputs here? What do they correspond to?
> 
> I see the text of [#7745 (comment)](https://github.com/IMA-WorldHealth/bhima/pull/7745#issuecomment-2391665924) that there are three separate inputs for bonus by grade. Would it be possible to explain what they do and what impact they have to the user? Maybe in a help message under the input?
> 
> I stared at it for a while, but I couldn't understand the difference between "bonus_by_grade", "frequency: bonus_by_grade" and "value: bonus_by_grade".

The explanation is simple: for an item such as "**Heure supplémentaire**" that is configured to be linked to a grade, once the grades have been set for each item, during payroll configuration, the system will check the employee's grade for items that are grade-linked. The "**Heure supplémentaire**" field will indicate the configured value for this item and for the corresponding grade. The "Frequency" field will allow the input of the number of overtime hours, and the system will multiply the frequency by the value to calculate the total amount to be assigned to the **Heure supplémentaire**.

image

image

image

lomamech: image
In this example, the employee has worked a total of **10 Heure supplémentaire**. The system will multiply these 10 hours by the configured value for their grade, which is 1.05 for overtime. Thus, 10 x 1.05 equals 10.5, and this value will be automatically assigned to the "Heure supplémentaire " rubric. Before this feature was introduced, payroll managers had to perform this calculation manually for each employee.

jmcameron (changes requested):
I tried the instructions in the PR and it seems to work fine.  However, I did have to add a reminder to enable index based payroll (see [this step](https://github.com/IMA-WorldHealth/bhima/pull/7745#issuecomment-2391523597)).  

Some of the end-to-end tests related to rubrics fail.   To see this error, run this:
```
npm install
npx update-browserslist-db@latest
npx playwright install chromium
npm run test:e2e-5
```
I only glanced at the code.  I will look at it more carefully in the next review.  

I encourage you to also resolve any changes the @jniles suggested.   

I may try to update dependencies in master, so you should verify/update your depdencies before force-pushing your updates.

lomamech:
> I tried the instructions in the PR and it seems to work fine. However, I did have to add a reminder to enable index based payroll (see [this step](https://github.com/IMA-WorldHealth/bhima/pull/7745#issuecomment-2391523597)).
> 
> Some of the end-to-end tests related to rubrics fail. To see this error, run this:
> 
> ```
> npm install
> npx update-browserslist-db@latest
> npx playwright install chromium
> npm run test:e2e-5
> ```
> 
> I only glanced at the code. I will look at it more carefully in the next review.
> 
> I encourage you to also resolve any changes the @jniles suggested.
> 
> I may try to update dependencies in master, so you should verify/update your depdencies before force-pushing your updates.

It's done, the test passes now

client/src/modules/multiple_payroll_indice/modals/config.modal.js
```diff
@@ -77,15 +87,36 @@ function ConfigIndicePaiementModalController(
...
           }
         });
+
+        return StaffingIndice.getRubricsGradeEmp({ employee_uuid : params.uuid });
+      })
+      .then(data => {
+        vm.rubricsGrade = data;
+
+        vm.rubrics.forEach(rb => {
+          vm.rubricsGrade.forEach(rub => {
```
jmcameron: Couldn't you just find the rubricsGrade using "find()" rather than looping through to find it?

Also, the choice of temporary variables is confusing.  "rb" is fine for rubric.  But "rub" should be something to tie it to rubricsGrade.  For instance, "rg" or "rubGrade" would be clearer; if you do this, you may want to change "rb" to "rub".

jmcameron (changes requested):
The `server/models/migrations/next/migrate.sql` file still has a broken merge.  It needs to be fixed.

I looked at the code; it looks okay but I did make a few comments.

All tests pass.  Thanks @lomamech !

lomamech:
> The `server/models/migrations/next/migrate.sql` file still has a broken merge. It needs to be fixed.
> 
> I looked at the code; it looks okay but I did make a few comments.
> 
> All tests pass. Thanks @lomamech !

Sorry I had fixed the problem, but when I saved the file with Ctrl + S it didn't take effect

lomamech:
@jmcameron  It's done

jmcameron:
@lomamech Thanks.  I'll try to look at it tomorrow.

Hopefully @jniles will have a chance to look it over again before we merge it.

lomamech:
> @lomamech Thanks. I'll try to look at it tomorrow.
> 
> Hopefully @jniles will have a chance to look it over again before we merge it.

Okay and thank you

server/models/01-schema.sql
jniles: I understand 👍

jniles:
LGTM.  All my changes have been addressed.

jmcameron (approved):
Looks like all the issues that @jniles and I had have been addresses.  Thanks @lomamech !

PR #7758, opened 2024-10-08
Refactor: Employee Payroll Configuration

### LINKED ISSUE #6386: Simplify Payroll Employee Configuration
There are a few times when we have a separate "label" and "configuration" items in a modals.  We should just combine these into the same modal.  For example:

image

These the first one leads to a modal with just a text box:
image

The second one leads to a modal with employee selections:
image

There isn't a good reason why we can't have both of these on the same modal, with one modal to set the name and determine the configuration of employees.

------

Additionally, it would be nice to have the number of employees configured displayed on the grid.  Like this:

image

This would help inform the user if they had more work to do.

### LINKED ISSUE #7755: List number of configured employees on the payroll configuration page
In addition to the changes requested in https://github.com/IMA-WorldHealth/bhima/issues/6386, it would be helpful to list the number of configured employees on the payroll configuration page.  For example, it would be helpful to have "150 / 158" in a column to indicate that 150 of the 158 employees available for selection were selected.

jniles:
This PR adds a column to the employee configuration page in the
human resources folder to indicate the number of configured employees
out of the total number of employees.  Here is what it looks like:

image

**Note**: the total number of employees is computed at the time the request
is made, not at the time that the configuration was created.   Thus,
employees hired after the fact will appear on the list, and may lead to
confusion.  I've included a recommendation for future improvements
in a comment.

This PR also merges the payroll employee configuration pages into a
single modal so that there is a unified view for labeling and modifying
the employee information.  It also changes the delete behavior to
accept deletions, even if there are associated employees.

Closes https://github.com/IMA-WorldHealth/bhima/issues/6386.
Closes https://github.com/IMA-WorldHealth/bhima/issues/7755.


jmcameron (changes requested):
The code looks good.
I tested it manually in BHIMA and it works, but I was confused by the count (using bhima_test).  I thought the right number would always be GE to the left number.
image

There are a couple of failures in the end-to-end tests.  To replicate, run

```
npm install
npx update-browserslist-db@latest
npx playwright install chromium
npm run test:e2e-2
```
In particular, these tests fail:
```
test/end-to-end/employeesConfig/employees_config.spec.js

Employees Configuration Management › successfully creates a new Employee Configuration
test/end-to-end/employeesConfig/employees_config.spec.js:32

Employees Configuration Management › successfully set Employees in Employee Configuration
test/end-to-end/employeesConfig/employees_config.spec.js:51

Employees Configuration Management › do not create when incorrect Employee
test/end-to-end/employeesConfig/employees_config.spec.js:55
```

jniles:
@jmcameron thanks for the review!  I can reproduce the test failures and will fix them shortly. 

I think the issue is with the test data not being realistic at all.  That may be something we should investigate fixing.

jniles:
@jmcameron, changes made and tests pass.

jmcameron (changes requested):
I'm still seeing one end-to-end failure:
```

test/end-to-end/employeesConfig/employees_config.spec.js

Employees Configuration Management › successfully set Employees in Employee Configuration
test/end-to-end/employeesConfig/employees_config.spec.js:51
```

Also, I'm still thinking about the 2/0 "Configured" value in
Human Resources > Payroll Employees Configuration
I agree the data may be messed up, but even so, it seems like the left number should never display larger than the right number.   So we either need to fix the data or fix the display.  For instance, instead of `2/0`, perhaps it should be blank, or `2/?`, `?/?` or `NA`?

jniles:
Hmmm ... Okay, so here is the meaning of the two numbers :

The numerator is the number of employees currently configured.  The denominator is the number of employees that could be configured.  Both of these numbers are based on the reality of the configuration, and should show the user that something weird happened (e.g. you locked an employee currently configured for payment).  Ideally this shouldn't happen.  But it at least tells you that it did happen.  Payroll is complex enough that it may just happen.

So I don't think the question marks make sense, since it isn't that we don't know this information.  Maybe I can just put a "Configuration error" text or something like that?

jniles:
Sorry, for some reason the last test seemed to have passed and I spoke to early.  Now it all should work.

jmcameron (changes requested):
Based on your comment that the issue with the 2/0 config displays are due to misconfigured test data, I took another look at test/data.sql.   I see that `employee.locked` is a 
`TINYINT(1) NULL DEFAULT 0,`
For the two test employees in the test data, `locked` is set to NULL.   I think it should be `0` (not locked).   When I do this, the config display changes to `2/2`.  I suggest you make this change in the test data:
```
$ git diff test/data.sql
diff --git a/test/data.sql b/test/data.sql
index 35cd6fe2e..72d73d83a 100644
--- a/test/data.sql
+++ b/test/data.sql
@@ -502,8 +502,8 @@ INSERT INTO `reference` VALUES
   (4, 0, 'XX', 'Deletable reference 1', 1, NULL, NULL);

 INSERT INTO `employee` (`uuid`, `code`, `date_embauche`, `grade_uuid`, `nb_spouse`, `nb_enfant`, `individual_salary`, `bank`, `bank_account`, `fonction_id`, `service_uuid`, `creditor_uuid`, `locked`, `patient_uuid`) VALUES
-  (HUID('75e09694-65f2-45a1-a8a2-8b025003d793'),'E1','2016-02-02 00:00:00',HUID('71e9f21c-d9b1-11e5-8ab7-78eb2f2a46e0'),1,3,500,'TMB','1201-3456-5423-03', 1, @medicineInterneService, HUID('42d3756a-7770-4bb8-a899-7953cd859892'), NULL,HUID('274c51ae-efcc-4238-98c6-f402bfb39866')),
-  (HUID('75e69409-562f-a2a8-45a1-3d7938b02500'), 'WWEFCB', '2016-01-01 01:00:00', HUID('9ee06e4a7b5948e6812cc0f8a00cf7d3'), 0, 0, 0, 'BCOL', '00-99-88-77', 1, @testService, HUID('18dcada5-f149-4eea-8267-19c346c2744f'), NULL, HUID('d1d7f856-d414-4400-8b94-8ba9445a2bc0'));
+  (HUID('75e09694-65f2-45a1-a8a2-8b025003d793'),'E1','2016-02-02 00:00:00',HUID('71e9f21c-d9b1-11e5-8ab7-78eb2f2a46e0'),1,3,500,'TMB','1201-3456-5423-03', 1, @medicineInterneService, HUID('42d3756a-7770-4bb8-a899-7953cd859892'), 0,HUID('274c51ae-efcc-4238-98c6-f402bfb39866')),
+  (HUID('75e69409-562f-a2a8-45a1-3d7938b02500'), 'WWEFCB', '2016-01-01 01:00:00', HUID('9ee06e4a7b5948e6812cc0f8a00cf7d3'), 0, 0, 0, 'BCOL', '00-99-88-77', 1, @testService, HUID('18dcada5-f149-4eea-8267-19c346c2744f'), 0, HUID('d1d7f856-d414-4400-8b94-8ba9445a2bc0'));
```
I made this change in my copy of your PR and all tests (including end-to-end tests) pass.
And the config display makes more sense to me.

jniles:
@jmcameron this is ready to go.

I've made the following updates:
 1. Converted the locked column to a "NOT NULL" and provided the appropriate migration and update scripts.
 2. Made sure all tests pass, and provided updates to the data and tests to take sure this is the case.

jmcameron (approved):
Excellent. Thanks @jniles !

jniles:
Thank you for all your reviews!

PR #7794, opened 2024-10-21
fix(general ledger): link to journal rather than account stmt

### LINKED ISSUE #7787: Make the "View Account Statement" link in the General Ledger link to a filtered journal
The following link used to point to the Account Statement, that was removed in https://github.com/IMA-WorldHealth/bhima/pull/7381.

image

It no longer works because the module was deleted.  Our friends in Vanga appreciated the dynamic view of the Account Statement, and propose that it should like to the Posting Journal to allow for drill-down analyses.

jniles:
This commit fixes a long broken link in the General Ledger that links to the account statement report. This report was removed some time ago.  It replaces this link with the filtered posting journal, which emulates the same behavior.  It allows users to have a dynamic look at the transactions affecting the account in question.

Closes #7787.


jmcameron (approved):
Looks good.  
Thanks for improving the ordering of the translation tokens.

I tested it manually and it works fine.

All tests pass.

PR #7795, opened 2024-10-21
docs: update docker installation instructions

jniles:
This commit updates the BHIMA installation instructions for the docker installation of MySQL, and bumps the supported version to 9.1.  It also updates the docker-compose file accordingly.


jmcameron (approved):
Based on feedback from @jniles, I will approve this PR although I have not tested it thoroughly.    It is pretty clear that it will only affect BHIMA operations when running in a docker container/VM.

PR #7818, opened 2024-10-23
chore(schema): standardize locked column definitions

### LINKED ISSUE #7756: Unify the `locked` column types across all schemas.
In BHIMA, we use `locked` columns quite a bit to signal records that should not be used for some reason or another.  It may be that an account is not longer useful, or an employee left the organization, or that a patient's account should be frozen.  Despite its ubiquity, we actually have multiple different definitions of the the column.   For example:

For the `account` table:
https://github.com/IMA-WorldHealth/bhima/blob/7359bb7cf3b096edfed4acee0124e0b1f5dec6ac/server/models/01-schema.sql#L17

For the `budget` table:
https://github.com/IMA-WorldHealth/bhima/blob/7359bb7cf3b096edfed4acee0124e0b1f5dec6ac/server/models/01-schema.sql#L117

For the `debtor_group` table:
https://github.com/IMA-WorldHealth/bhima/blob/7359bb7cf3b096edfed4acee0124e0b1f5dec6ac/server/models/01-schema.sql#L380

We treat them all the same, functionally in Javascript.  They are most often converted to a binary `0`/`1` and filtered on with `locked <> 1` or `locked =1` in SQL.  Given this, we should standardize the column definitions.  I propose the following:

```sql
`locked` TINYINT(1) NOT NULL DEFAULT 0
```

This way, we set the rules that (1) it must be defined and (2) it defaults to 0 on creation.  Also, we remove the confusion around `BOOLEAN` types between SQL and Javascript.

To complete this issue, the developer should:

 1. Find/replace all the `locked` columns in the `schema.sql` with the new type definition.
 3. Ensure tests pass.
 4. Write migration scripts in `migrate/next.sql` to ensure that old databases are updated.

jniles:
Standardizes the `locked` column definitions so that they all default to 0 (instead of NULL, as in the case of employees).  It should make the schema more predictable, and therefore more maintainable.

Closes #7756.

### DIFF
--- a/server/models/01-schema.sql
+++ b/server/models/01-schema.sql
@@ -14,8 +14,8 @@ CREATE TABLE `account` (
   `number` INT(11) UNSIGNED NOT NULL,
   `label` VARCHAR(191) NOT NULL,
   `parent` INT(10) UNSIGNED NOT NULL,
-  `locked` TINYINT(1) UNSIGNED DEFAULT 0,
-  `hidden` TINYINT(1) UNSIGNED DEFAULT 0,
+  `locked` TINYINT(1) NOT NULL DEFAULT 0,
+  `hidden` TINYINT(1) UNSIGNED NOT NULL DEFAULT 0,
   `created` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `reference_id` TINYINT(3) UNSIGNED DEFAULT NULL,
   `cost_center_id` MEDIUMINT(8) UNSIGNED NULL,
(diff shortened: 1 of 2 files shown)


server/models/01-schema.sql
```diff
@@ -14,8 +14,8 @@ CREATE TABLE `account` (
   `number` INT(11) UNSIGNED NOT NULL,
   `label` VARCHAR(191) NOT NULL,
   `parent` INT(10) UNSIGNED NOT NULL,
-  `locked` TINYINT(1) UNSIGNED DEFAULT 0,
-  `hidden` TINYINT(1) UNSIGNED DEFAULT 0,
+  `locked` TINYINT(1) NOT NULL DEFAULT 0,
+  `hidden` TINYINT(1) UNSIGNED NOT NULL DEFAULT 0,
```
jmcameron, line 18: Should this still have "UNSIGNED" (line 18)?

server/models/migrations/next/migrate.sql
```diff
@@ -46,5 +46,21 @@ CREATE TABLE `rubric_grade_indice` (
 
 DROP TABLE IF EXISTS odk_app_user;
 
+-- @jniles:
 -- make sure that employee locked values are always defined.
 UPDATE employee SET locked = 0 WHERE ISNULL(locked);
```
jmcameron, line 51: Do you need to do this on any of the other `locked` columns for other tables?  I checked and the bhima_test data looked okay, but I was thinking about the data in production sites.

jmcameron (approved):
The code looks fine although I had a couple of questions in the code.  

In particular, it was not clear to me whether you need more applications of

`UPDATE employee SET locked = 0 WHERE ISNULL(locked);`

in the migration code for other tables for the production sites.

The tests all pass.

server/models/01-schema.sql
jniles: I haven't really looked into how `hidden` is used in this case, so I'm not going to touch it.

server/models/migrations/next/migrate.sql
jniles: image

Yep, looks like I need to do the same for account.  The rest shouldn't be NULL ABLE.

Done, thanks for the catch!

PR #7819, opened 2024-10-23
chore(employee): add timestamps to employee table

### LINKED ISSUE #7817: Add `created_at` and `updated_at` columns to the employee table
Refactoring the Employee Configuration page  (https://github.com/Third-Culture-Software/bhima/pull/7758) went some way to improving the payroll configuration, but it isn't fully useful yet.  In order to make this feature fully useful, we need to know when `employees` were created, otherwise it may flag employees as available for payment in payment periods before they were hired by the institution (inflating the denominator)."

We should add `created_at` and `updated_at` columns to the employee table.  This would allow filtering to ensure that the denominator only includes employees that were available during the payment period.

jniles:
This commit adds created_at and updated_at timestamps to the employee table.

Closes #7817.

### DIFF
--- a/server/models/01-schema.sql
+++ b/server/models/01-schema.sql
@@ -529,6 +529,8 @@ CREATE TABLE `employee` (
   `patient_uuid`  BINARY(16) DEFAULT NULL,
   `reference`     SMALLINT(5) UNSIGNED DEFAULT NULL,
   `title_employee_id`   TINYINT(3) UNSIGNED DEFAULT NULL,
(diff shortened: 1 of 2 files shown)


jniles:
@jmcameron I fixed merge conflicts and defaulted the timestamps to the `date_embauche`, as suggested on a sidechannel.  That was a really good idea.

I won't update the FR -> EN, since it is found in lots of different modules and I don't want to mess with debugging that yet.  Here is the issue: #7821 .

jmcameron (approved):
Code looks good.

All tests pass.

PR #7822, opened 2024-10-25
fix(admissions): block transfer of discharged patients

### LINKED ISSUE #7809: You can transfer discharged patients to other beds in the Admissions Registry
Below, you can see that a user can still "transfer" a patient that has already been discharged.  This option should be removed.

image

jniles:
This commit fixes a bug in the admissions registry where you could transfer a patient that has already been discharged.  This shouldn't happen in practice as the patient has left the hospital.

Closes #7809.

To test this:
 1. Create a Ward, Room, and Bed via Ward Management > Settings.
 2. Admit a patient to that room (as hospitalization).
 3. Note that you have the option to "transfer" the patient in the menu in the Admissions registry.
 4. Discharge the patient.
 5. Note that you cannot transfer the patient anymore.

### DIFF
--- a/client/src/modules/patients/visits/registry.js
+++ b/client/src/modules/patients/visits/registry.js
@@ -9,7 +9,6 @@ AdmissionRegistryController.$inject = [
 /**
  * Admission Registry Controller
  *
- * *
  * This module is responsible for the management of Admission Registry.
  */
 function AdmissionRegistryController(
@@ -19,7 +18,7 @@ function AdmissionRegistryController(
   const vm = this;
   const cacheKey = 'AdmissionRegistry';
 
-  // the grid registry filterer
+  // the grid registry filter
   const { grid } = Visits;
 
   vm.loading = false;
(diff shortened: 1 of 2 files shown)


jmcameron (changes requested):
The code is fine.
It works as described.   The PR can be merged, but I would like a  few minor fixes.
- Cosmetic: Please change the menu icon for Ward > Settings to a 'gear' (see Stock > settings)
- In English, the `Hospital > Visits Registry` menu item takes you to the "Hospital > Admissions Registry`.  It would be good to make these consistent.  Not sure if it is an issue in French.  I see "Admissions" more in actual hospitals in the USA, so perhaps that is a better term than "Visits".
- When you go to the `Hospital > Visits` and click on actions > transfer, the Submit button should be deactivated if the selected transfer is not possible (eg, trying to transfer to a ward without open beds) -- or at least put up a warning dialog that no beds are available in that ward.  If you prefer not to do this in this PR, please file an issue to do it later.

jniles:
@jmcameron what do you think of moving the entire "Ward Management" module into the "Hospital" folder?  To me, it has more to do with patient management + hospital than being a top-level item.

If you think that is the way we should go, I'd address these in a different PR that tackles those items specifically.

jmcameron:
> @jmcameron what do you think of moving the entire "Ward Management" module into the "Hospital" folder? To me, it has more to do with patient management + hospital than being a top-level item.
> 
> If you think that is the way we should go, I'd address these in a different PR that tackles those items specifically.

That sounds great.

PR #7823, opened 2024-10-25
feat(admissions): discharge patient directly from registry

### LINKED ISSUE #7801: Add ability to discharge a patient from the Admissions Registry.
It would be ideal to link the patient discharge form from the Admissions Registry.

jniles:
This commit adds the feature to discharge patients directly from the patient admissions registry.

Closes #7801.

Partially addresses https://github.com/Third-Culture-Software/bhima/issues/5232.

Here is what it looks like in practice:

image

### DIFF
--- a/client/src/modules/patients/visits/registry.js
+++ b/client/src/modules/patients/visits/registry.js
@@ -31,6 +31,7 @@ function AdmissionRegistryController(
   vm.toggleInlineFilter = toggleInlineFilter;
   vm.openTransferModal = openTransferModal;
   vm.openVisitModal = openVisitModal;
+  vm.openPatientDischargeModal = openPatientDischargeModal;
 
   const patientCardTemplate = `
     <div class="ui-grid-cell-contents">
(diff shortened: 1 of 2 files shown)


lomamech:
The feature works, but nevertheless I will make a comment, although the bhima is not very experienced in patient management, I had admitted a patient as inpatient and I discharged him as deceased and just after I readmitted him as an outpatient, I think that in the future we should think about the patient declared deceased

lomamech:
image

image

image

lomamech:
I think the PR is good

lomamech (approved):
The feature works, but nevertheless I will make a comment, although the bhima is not very experienced in patient management, I had admitted a patient as inpatient and I discharged him as deceased and just after I readmitted him as an outpatient, I think that in the future we should think about the patient declared deceased

jniles:
@lomamech that is a good idea.  In practice, I assume it is unlikely to happen, since the only reason to record this information is to help the hospital know how it is doing, but it still makes sense to try and limit junk data.

I see you don't have the ability to merge - let me try to fix that.

PR #7850, opened 2024-11-01
fix(wards): allow removing services from wards

### LINKED ISSUE #7802: We should be able to remove a service from the ward update modal.
image

jniles:
This commit fixes the "service" section of the ward create/update controller to allow for removing the service.

You can test this by creating a ward with a service, then updating the ward to remove the service.  You can also add a service to the ward.

Closes #7802.

### DIFF
--- a/client/src/modules/ward/configuration/ward/modals/createUpdate.html
+++ b/client/src/modules/ward/configuration/ward/modals/createUpdate.html
@@ -24,6 +24,7 @@
     <bh-service-select 
       service-uuid="ModalCtrl.ward.service_uuid"
       on-select-callback="ModalCtrl.onSelectService(service)">
+      <bh-clear on-clear="ModalCtrl.clear('service_uuid')"></bh-clear>
     </bh-service-select>
 
     <div class="form-group" ng-class="{'has-error': ModalForm.$submitted && ModalForm.description.$invalid }">
@@ -32,8 +33,7 @@
         ng-model="ModalCtrl.ward.description"
         name="description"
         class="form-control"
-        rows="3"
-      >
+        rows="3">
       </textarea>
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


jniles:
@mbayopanda could I get a review?

client/src/modules/ward/configuration/ward/modals/createUpdate.html
```diff
@@ -24,6 +24,7 @@
     <bh-service-select 
       service-uuid="ModalCtrl.ward.service_uuid"
       on-select-callback="ModalCtrl.onSelectService(service)">
+      <bh-clear on-clear="ModalCtrl.clear('service_uuid')"></bh-clear>
```
mbayopanda, line 27: 👍🏽

server/controllers/medical/ward/ward.js
```diff
@@ -22,14 +22,19 @@ function create(req, res, next) {
...
   db.convert(data, ['service_uuid']);
-  const sql = `UPDATE ward SET ? WHERE uuid =?`;
 
-  db.exec(sql, [data, uuid]).then(() => {
-    res.sendStatus(200);
-  })
+  const uuid = db.bid(req.params.uuid);
+
+  // if no service is set back, make it null
+  if (!data.service_uuid) { data.service_uuid = null; }
```
mbayopanda, line 32: 👍🏽

mbayopanda (approved):
LGTM 👍🏽

jmcameron (changes requested):
All tests pass and the code works fine, but I did notice one thing I suggest you correct:
in `client/src/modules/ward/configuration/ward/modals/createUpdates.js`,
on line 10:
```
function CreateUpdateWardController(uuid, Session, Ward, ModalService, Notify, Instance, Service) {
```

Two arguments are not used:  Session and ModalService.   I suggest removing them.

I'm quite sure this will not affect the tests.  So once you make that change, do a simple test via the GUI to confirm nothing is broken, then feel free to merge it yourself.

jniles:
I think I fixed all the issues 👍

PR #7874, opened 2024-11-14
Remove q.js from package.json

### LINKED ISSUE #7871: Remove `q` npm module
The node module `q` introduced promise support to NodeJS before Promises were mainstream.  It's been deprecated and hasn't been maintained in years.  See https://www.npmjs.com/package/q.

It's high time we removed it in favor of a native promise interface.  It should help us also more easily transition to async/await in the future since we will remove all our `.done()` calls.

jniles:
This PR removes the deprecated q.js from the server package.json.  Note that `$q`, AngularJS's implementation of promises, remains untouched on the clientside.  This only affects the server-side compilation.

The majority of this PR consists of the following:
 1. Remove all `.done();` calls at the end of promise chains.
 2. Transitional all `q.all()` calls to `Promise.all()` calls.
 3. Replace `q.resolve()` or `q()` calls with `Promise.resolve()`.
 4. Replace q's `.spread()` operator with `.then(([..]) => {})` syntax.

More substantially, the `lib/db/*.js` library was migrated to `mysql2/promise`, since it previously depended on `q.js`.  The majority of work was in this rewrite.   Instead of using `q.defer()` for setting timeouts, we now use NodeJS's internal `node:timers/promises` module which configures `setTimeout()` to return a promise.  

I also added a single unit test on the server to make sure that the transaction behavior is the same during the refactor.

Closes https://github.com/Third-Culture-Software/bhima/issues/7871.

Note that, because we have now removed the non-standard `.done()` calls, it should be straightforwards to migrate to using `async`/`await` everywhere on the server (https://github.com/Third-Culture-Software/bhima/issues/2458).


jniles:
@jmcameron @mbayopanda , when you get a chance, can you pull the repository and run the tests on your machines?  All of them, if possible.

The integration tests pass on my machine, which should be enough for most functions touched here. I'm currently running the end-to-end tests.

It would also be useful to just poke around in the application and make sure things behave as intended.

jmcameron (approved):
Looks good.
All tests pass.
Did some cursory checks with the  GUI and they seem to work.

PR #7876, opened 2024-11-15
docs: add prepayment text to cash-payments

jniles:
Improve the documentation by adding an explanation of prepayments (cautions) to the cash documentation.


jmcameron (approved):
I read through it and it seems okay (grammatically).  Not that familiar with how the underlying code actually works.  Not testing is necessary since docs cannot affect any tests.

jniles:
Thanks @jmcameron !

PR #7881, opened 2024-11-19
Updates to prevent marking a male patient pregnant during admission

### LINKED ISSUE #7847: Do not admit male patients as pregnant
We currently do not parse the sex of the patient when admitting them, allowing them to be admitted as pregnant.  We should probably pop up a warning about this issue.  

I'm not 100% sure if we should block this case, since it could be that the patient's sex was registered incorrectly and should be updated instead... but I'm willing to hear proposals for best practice around this.  Ultimately, I assume it should be a fairly rare event.

jmcameron:
Prevent marking a male patient as pregnant during admission.

This PR prevents the "Is pregnant" yes/no selection from being displayed for male patients during admission.
It also:
 - Sets defaults for the "is referal" and "is pregnant" yes/no flags (both to NO)
 - Adds a "Sex" column for the Visits registry

Closes https://github.com/Third-Culture-Software/bhima/issues/7847

Note that I considered the comment in the issue:

>    "I'm not 100% sure if we should block this case, since it could be that the patient's sex was registered incorrectly 
>    and should be updated instead... but I'm willing to hear proposals for best practice around this. Ultimately, I 
>    assume it should be a fairly rare event."

I tried a "Danger" message, but had to close the new visit dialog for that to work, which was confusing.  I also tried a secondary Modal popup with a warning message, but had other issues.

Hopefully this error is rare and the data entry person will realize the problem if a woman cannot be marked as pregnant.
Perhaps one way to address this is to insert a line below the patient identifier line with patient sex, age, etc, when a patient is selected.  That should make the problem a little clearer to the data entry person.  That could be done as part of this PR or as a separate issue.

**TESTING**
- Use bhima_test
- Try admitting patient 'Test Patient 1'.  This is a female, and the pregnancy field should be visible and usable.
- Try admitting patient 'Test Patient 2'.  This is a male, and the pregnancy field should not be visible.

### DIFF
--- a/client/src/modules/patients/visits/modals/visits-modal.ctrl.js
+++ b/client/src/modules/patients/visits/modals/visits-modal.ctrl.js
@@ -26,6 +26,8 @@ function VisitsAdmissionController(ModalInstance, Patients, Visits, Notify,
     hospitalized : 0,
     inside_health_zone : 1,
     is_new_case : 1,
+    is_refered : 0,
+    is_pregnant : 0,
   };
 
   vm.onBedRoomSelect = bed => {
--- a/client/src/modules/patients/visits/modals/visits.modal.html
+++ b/client/src/modules/patients/visits/modals/visits.modal.html
@@ -98,14 +98,16 @@
         <!-- optionally include help-text -->
     </bh-yes-no-radios>
 
-    <bh-yes-no-radios
-      label ="PATIENT_RECORDS.VISITS.IS_PREGNANT"
-      id = "patient-is-pregnant"
-      value = "AdmitCtrl.visit.is_pregnant"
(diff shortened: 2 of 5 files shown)


client/src/modules/patients/visits/modals/visits-modal.ctrl.js
```diff
@@ -26,6 +26,8 @@ function VisitsAdmissionController(ModalInstance, Patients, Visits, Notify,
     hospitalized : 0,
     inside_health_zone : 1,
     is_new_case : 1,
+    is_refered : 0,
+    is_pregnant : 0,
```
jniles, line 30: 👍

jniles (approved):
Tested and works as described.  Great work!

PR #7896, opened 2024-11-26
refactor(visits); uses yes-no radios in search modal

### LINKED ISSUE #7815: Improve the search modal for Admissions Registry
The Admissions Registry search modal doesn't follow best practices.  See below:

image

A checkbox is ambiguous because the "no" and "untouched" choices seem identical.  Instead, we should have a `<bh-yes-no>` button that is able to be cleared.  That way, we clearly differentiate between the three states - filter only patients who meet the criteria, filter patients who do not meet the criteria, or do not include the criteria in the filter.

jniles:
Improves the search modal by using the yes-no-radios to ensure that no default case is assumed.

Closes #7815.


jmcameron (changes requested):
One of the end-to-end tests fails:

`test/end-to-end/visits/visits.spec.js:95:3 › Patient Visits › search pregnant visits
`
To duplicate:  `npm run test:e2e-8`

BTW, 'git pull' your PR, since I updated it based on a prior merge.

jniles:
@jmcameron yep, that is why I made it a PR draft instead of an actual PR 😆 .

Now I have corrected the test, and it should be ready for merge.

jmcameron (approved):
The code looks good.
All tests pass.
Tested manually with bhima_test using the normal interface and the search options work fine.

PR #7900, opened 2024-11-28
Fix cashflow report improve budget report

### LINKED ISSUE #7897: Fix and Update Cashflow Report and Budget Report
Description:
This issue addresses the following tasks for improving the Cashflow Report and Budget Report:

1. Cashflow Report:

- Correct a variable name that is causing an error in the report.

2. Budget Report:

- Add an analytical section to enhance the report with more detailed insights.

lomamech:
- Fixed display error in the Cashflow report.
- Added a new summary section to improve budget report.

closes #7897


lomamech:
Here is a PR correcting the Display error on the Cashflow report and adding the summary section on the Budget report

jniles:
Can you remove the yarn.lock file?

Also, your migration SQL file refers to tables that do not exist.

client/src/i18n/en/form.json
```diff
@@ -368,7 +368,7 @@
             "BEGINNING_VIEW_BALANCE" : "Beginning View Balance",
             "BILLING_DATE": "Billing Date",
             "BREAK_EVEN": "Break Even",
-            "BUDGET": "Budget",
```
jniles, line 371: This key is used in two additional files:

 1. `client/src/modules/budget/modal/editAccountBudget.html`
 2. `client/src/modules/budget/modal/import.html`
 
By changing it, you are breaking those translations.  You probably should just add a new key and use that.

client/src/modules/reports/generate/budget_report/budget_report.config.js
```diff
@@ -31,13 +33,31 @@ function BudgetReportController($sce, Notify, SavedReports, AppCache, reportData
...
   ];
 
   vm.preview = function preview(form) {
     if (form.$invalid) { return null; }
 
+    if (vm.include_summary_section === 0) {
+      vm.reportDetails.cashboxesIds = [];
+    }
+
+    vm.reportDetails.include_summary_section = vm.include_summary_section;
```
jniles: Why do you have both `vm.reportDetails.include_summary_section` and `vm.include_summary_section`?  Can you just use the `reportDetails` directly?

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -43,9 +47,98 @@ async function report(req, res, next) {
     const reportColumn = [];
     const reportFootColumIncome = [];
     const reportFootColumExpense = [];
+    let configurationReferences = [];
 
     const fiscalYear = await Fiscal.lookupFiscalYear(fiscalYearId);
 
+    if (includeSummarySection) {
+      const cashboxesIds = _.values(params.cashboxesIds);
+      const dataCashboxes = {};
```
jniles: Why does dataCashboxes need to be an object?  It seems like you could just work with the values directly.

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -43,9 +47,98 @@ async function report(req, res, next) {
...
+      dataCashboxes.cashLabelDetails = dataCashboxes.cashboxes.map(
+        cashbox => `${cashbox.account_number} - ${cashbox.account_label}`);
+
+      const sqlTotalIncomeCash = `
+        SELECT gl.trans_id, gl.trans_date, a.number, a.label, gl.debit_equiv, gl.credit_equiv, tt.text, tt.type,
+          SUM(gl.debit_equiv - gl.credit_equiv) AS balance
+        FROM general_ledger AS gl
+        JOIN account AS a ON a.id = gl.account_id
+        JOIN transaction_type AS tt ON tt.id = gl.transaction_type_id
+        WHERE gl.account_id IN ? AND gl.fiscal_year_id = 10 AND tt.type = 'income'
```
jniles: Why `gl.fiscal_year_id = 10`?

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -43,9 +47,98 @@ async function report(req, res, next) {
...
+        AND gl.transaction_type_id <> 10 AND gl.record_uuid NOT IN (
```
jniles: I don't think you need the filter on `transaction_type_id = 10`, since the Credit Note transaction type doesn't have the `income` type.

https://github.com/Third-Culture-Software/bhima/blob/master/server/models/06-bhima.sql#L340C17-L340C23

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -43,9 +47,98 @@ async function report(req, res, next) {
...
+          SELECT DISTINCT gl.record_uuid
+          FROM general_ledger AS gl
+          WHERE gl.record_uuid IN (
```
jniles, line 89: This is a common pattern in BHIMA.  I wonder if we can simplify this by adding a column `reversed` to the general_ledger and posting_journal, and set that to 1 whenever the relevant reversal transaction is created.  I'll make an issue about it.

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -43,9 +47,98 @@ async function report(req, res, next) {
...
+        fiscalYear.start_date,
+        fiscalYear.end_date,
+        fiscalYear.start_date,
+        fiscalYear.end_date,
+        fiscalYear.start_date,
+        fiscalYear.end_date,
+      ];
+
+      const totalIncomeCash = await db.exec(sqlTotalIncomeCash, paramsFilter);
+      totalFinancement = totalIncomeCash[0].balance;
```
jniles: If this returns only a single row, why not use `db.one()`?  Then you could do something like:

```js
const { totalFinancement : balance } = await db.exec(sqlTotalIncomeCash, paramsFilter);
```

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -43,9 +47,98 @@ async function report(req, res, next) {
...
+
+      const BUDGET_ANALYSIS = 8;
```
jniles: Maybe call this `BUDGET_ANALYSIS_REFERENCE_TYPE_ID`.

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -43,9 +47,98 @@ async function report(req, res, next) {
...
+      `;
+
+      configurationReferences = await db.exec(sqlReferences, [BUDGET_ANALYSIS]);
+
+      configurationReferences = configurationReferences.map(item => ({
+        ...item,
+        accounts_number_formated : item.accounts_number
+          .split(',')
+          .map(num => parseInt(num.trim(), 10))
+          .filter(num => typeof num === 'number'),
```
jniles: I prefer testing `Number.isInteger()` here, since it should be an integer (from the `parseInt()` above, and, in Javascript.
```js
type of NaN === "number"
```

Because JavaScript is just the worst.

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -173,6 +266,10 @@ async function report(req, res, next) {
       });
     });
 
+    const tabIncomeDataBudget = tabFiscalReport.filter(id => {
+      return (id.type_id === 4 || id.isIncomeTitle);
```
jniles: Remind me again what `4` is for?  Could you give it a nice name?

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -43,9 +47,98 @@ async function report(req, res, next) {
...
+        FROM account_reference AS ar
+        JOIN account_reference_item AS art ON art.account_reference_id = ar.id
+        JOIN account AS a ON a.id = art.account_id
+        WHERE ar.reference_type_id = ?
+        GROUP BY ar.id;
+      `;
+
+      configurationReferences = await db.exec(sqlReferences, [BUDGET_ANALYSIS]);
+
+      configurationReferences = configurationReferences.map(item => ({
```
jniles, line 130: I'm a bit confused about this code.  It looks like you use `GROUP_CONCAT()` to make the account numbers into a comma separated string, then in Javascript, you do some work to separate them and parse them into separate numbers.  Why not just remove the `GROUP_CONCAT()` call, and treat them all as separate numbers from the beginning?

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -201,6 +298,33 @@ async function report(req, res, next) {
 
     const totalCompletionIncome = totalRealisationIncome / totalBudgetIncome;
     const totalCompletionExpense = totalRealisationExpense / totalBudgetExpense;
+    const realisationIncomeExpense = totalRealisationIncome - totalRealisationExpense;
+
+    configurationReferences.forEach(item => {
+      item.value_account_number = 0;
+
+      item.accounts_number_formated.forEach(elt => {
```
jniles: If you are just using the `GROUP_CONCAT()` account numbers for filtering, why use account numbers?  Why not use `account.id`s?

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -201,6 +298,33 @@ async function report(req, res, next) {
...
+
+        item.value_account_number += filterData[0].realisation;
+      });
+
+    });
+
+    let firstIncomeConfigurationReferences = 0;
+    let secondIncomeConfigurationReferences = 0;
+
+    if (includeSummarySection) {
```
jniles, line 306: Can you help me understand this block of code?  Why are the first two lines being treated differently?

And where is the `ORDER BY` or `sort()` command that guarantees the order they come out in?  Because it could be random.

server/models/06-bhima.sql
```diff
@@ -332,7 +332,7 @@ INSERT IGNORE INTO `transaction_type` (`id`, `text`, `type`, `fixed`) VALUES
   (2, 'VOUCHERS.SIMPLE.CASH_PAYMENT', 'income', 1),
   (3, 'VOUCHERS.SIMPLE.CONVENTION_PAYMENT', 'income', 1),
   (4, 'VOUCHERS.SIMPLE.SUPPORT_INCOME', 'income', 1),
-  (5, 'VOUCHERS.SIMPLE.TRANSFER', 'other',  1),
+  (5, 'VOUCHERS.SIMPLE.TRANSFER', 'income',  1),
```
jniles: This is a breaking change, and we should talk about it.

Why are all transfers considered income?  I think the majority of account to account transfers at Vanga (for example, from the auxiliary cashbox to the primary cashbox) are marked as "transfers".  Same for transfers to and from the bank.

Why is a transfer an income?

server/models/migrations/next/migrate.sql
```diff
@@ -0,0 +1,11 @@
+/*
+ * @author: lomamech
+ * @date: 2024-11-27
+ * @description: Fix and Update Cashflow Report and Budget Report #7897
+ */
+UPDATE `new_imck_nov_2024`.`transaction_type` SET `type`='income' WHERE  `id`=5;
```
jniles: The database `new_imck_nov_2024` doesn't exist.  Please correct this line.

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -43,9 +47,98 @@ async function report(req, res, next) {
     const reportColumn = [];
     const reportFootColumIncome = [];
     const reportFootColumExpense = [];
+    let configurationReferences = [];
 
     const fiscalYear = await Fiscal.lookupFiscalYear(fiscalYearId);
 
+    if (includeSummarySection) {
+      const cashboxesIds = _.values(params.cashboxesIds);
```
jniles, line 61: Isn't this already an array?  What does `_.values()` add?

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -201,6 +298,33 @@ async function report(req, res, next) {
 
     const totalCompletionIncome = totalRealisationIncome / totalBudgetIncome;
     const totalCompletionExpense = totalRealisationExpense / totalBudgetExpense;
+    const realisationIncomeExpense = totalRealisationIncome - totalRealisationExpense;
+
+    configurationReferences.forEach(item => {
```
jniles, line 296: Anytime you are doing `.forEach()` and `+=`, you probably are doing a `.reduce()`.  Here is a rewrite that does a similar thing, but with fewer intermediate variables, to make it easier to reason about.  It also uses `.find()` instead of `.filter()` since you only need the first item in the filtered array.

```js
configurationReferences.forEach(item => {
  item.value_account_number = item.accounts_number_formated.reduce((total, accountNumber) => {
    const matchingIncome = tabIncomeDataBudget.find(income => income.number === accountNumber);
    return total + (matchingIncome ? matchingIncome.realisation : 0);
  }, 0);
});
```

jniles (changes requested):
I haven't tested this yet with any live data, but here is a review of the code.  

There are a few things that should be caught before submitting a PR, like the missing translation tokens, the removal of `yarn.lock`, and the migration file hard coding the name of a database.

There are a lot of decisions I didn't understand, such as changing the transaction type "TRANSFER" to `income`.  That shouldn't have been done without an issue explaining _why_ all the sites (Vanga, Sona Bata, Bandunduville) need to change their transfer transaction types to `income`.  Then we could all discuss why this makes sense or doesn't make sense, but I'm alarmed that it was just slipped in without even a mention in the PR description.

client/src/i18n/en/form.json
lomamech: This is an inadvertent mistake on my part.

client/src/modules/reports/generate/budget_report/budget_report.config.js
lomamech: **This is a hack** for reasons I couldn't figure out when I was using vm.reportDetails.include_summary_section the default should be false but I wasn't getting the result I was hoping for which is why I went with this solution

server/controllers/finance/reports/budget_analytical/index.js
lomamech: You are right

server/controllers/finance/reports/budget_analytical/index.js
lomamech: This is a fatal mistake on my part.

server/controllers/finance/reports/budget_analytical/index.js
lomamech: The Lodash function is important for this section this method is used to retrieve only the values, in the form of an array and the query uses the IN clause, here is the error we get when we do not use Lodash
image

server/controllers/finance/reports/budget_analytical/index.js
lomamech: Great, it works well

server/controllers/finance/reports/budget_analytical/index.js
lomamech: All right

server/controllers/finance/reports/budget_analytical/index.js
lomamech: const ACCOUNT_TYPES_INCOME = 4;
    const ACCOUNT_TYPES_EXPENSE = 5;

server/controllers/finance/reports/budget_analytical/index.js
lomamech: The reason why I used GROUP_CONCAT was indeed to have the data separated by commas and the processing was intended to have its different elements in a table as shown in the image in illustration
image

server/controllers/finance/reports/budget_analytical/index.js
lomamech: This section is really a hack, Summary section is presented as follows
image
This section is really a hack, Summary section is presented as follows, Result equals the difference between the total of Income Total - Expenses Total, Result Without Accounts: : 71, 75, 77 and Result Without Accounts: : 71 should evaluate the result without the sum of the balance of the following accounts 71, 75 and 77 but the challenge I had was to obtain the balance of these specific accounts, that's why I had opted for using the account references by adding the Budget Analysis type account references and when creating the references I made the first grouping 71, 75 and 77 and the second 71 (also equivalent to grants)

server/models/06-bhima.sql
lomamech: Regarding the change of transfer type transaction, I had chosen to modify the type to allow the analysis of transfers coming from the auxiliary cash registers of the receipts, but since this will have repercussions on the other production sites, **_I have the following proposal: add in the interface of the cashflow report the possibility of considering the transfers as being receipts and the default value will be false._**, @jniles @jmcameron What do you think about this proposal which would resolve the IMCK problem without disturbing other production sites?

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -18,11 +18,17 @@ exports.report = report;
...
+  params.include_summary_section = parseInt(params.include_summary_section, 10);
+
   const fiscalYearId = parseInt(params.fiscal_id, 10);
   const setNumberYear = parseInt(params.set_number_year, 10) + 1;
   const colspanValue = parseInt(params.set_number_year, 10) + 6;
   const hideUnused = parseInt(params.hide_unused, 10);
   const { enterprise } = req.session;
+  const includeSummarySection = params.include_summary_section;
+
+  const ACCOUNT_TYPES_INCOME = 4;
```
jniles, line 29: Nice 👍

server/controllers/finance/reports/budget_analytical/index.js
jniles: I am worried this will break in the future without a deterministic ordering.  Is there any way we can make sure this data is sorted in the correct display order before merging?

As an example, suppose that MySQL gets an update that makes the shorter `group_concat()` strings be put first.  Or suppose a production server crashes and we need to restore it from backup, and it reorders the data.  I think this should have some sorting/ordering to make sure that doesn't happen.

server/models/06-bhima.sql
jniles: I think I'm okay with that hack as a solution.  Alternatively, we could create a custom "cash transfer" transaction type for IMCK that has the type_id "income" that IMCK could start to use for caisse-to-caisse transfers instead and then it would show up here.

---

Ultimately, I think this speaks to a larger issue in BHIMA - we don't separate "cash" and "unrealized revenue" very well.  I thought we had an issue about this, but I cannot find it.  We probably should model this in the `account_type` or `account_category` which would have greater refinement of the type of transaction.  Instead of simply having "asset", "liability", etc, as our account types, we would have sub-categories, such as "bank", "credit card", "caisse", "mpesa", all of which are subcategories of the "cash" account.  Then, instead of analyzing transaction types, we would be looking at flows into and out of the cash accounts, which I think is likely the spirit of the cashflow report.

Of course, this would mean that all institutions would need to classify their chart of accounts correctly, but that seems like a good requirement to have.

server/models/migrations/next/migrate.sql
```diff
@@ -0,0 +1,11 @@
+/*
+ * @author: lomamech
+ * @date: 2024-11-27
+ * @description: Fix and Update Cashflow Report and Budget Report #7897
+ */
+UPDATE `transaction_type` SET `type`='income' WHERE  `id`=5;
```
jniles: If we add a special flag to the report, remember to undo these changes and the ones in the data.sql.

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -192,15 +287,39 @@ async function report(req, res, next) {
     }
 
     const tabFiscalIncomeData = tabFiscalReport.filter(id => {
-      return (id.type_id === 4 || id.isIncomeTitle);
+      return (id.type_id === ACCOUNT_TYPES_INCOME || id.isIncomeTitle);
```
jniles, line 285: Great, thanks!

server/controllers/finance/reports/budget_analytical/index.js
jniles: https://github.com/Third-Culture-Software/bhima/issues/7904

jniles (commented):
This is much better, thank you!

I am still concerned about the sorting issue.  Is there anyway we can guarantee the output sort?

server/controllers/finance/reports/budget_analytical/index.js
lomamech: image
To improve things, I will make the summary section dynamic, so that the analysis of the result will no longer be limited to two account references and but only in relation to the configuration of the account references

server/models/06-bhima.sql
lomamech: But regarding the cashflow report, I just think to add the option to consider the transfer in the income category, like that we will not need to add new types of transaction_type

jniles: Yes, I'm okay with this for IMCK.

server/controllers/finance/reports/budget_analytical/index.js
lomamech: > referen

@jniles After trying to make the summary section dynamic, I encountered the following constraint: how to distinguish the value of the subsidies represented by account 71 in the configuration?

Because in the current configuration I have configured the account references so that the subsidies are associated with the second reference, while accounts 71, 75 and 77 are grouped under the first reference. After creating a new transaction type, I make sure that this section remains compliant with the report template.

What do you think of this hack or trick, if not we will have to find another way to distinguish the subsidy accounts

image

lomamech:
image
@jniles @jmcameron 
I just made major changes in this PR, I just restored the transaction type transfer as **other** transaction and not as an income and I added the option **Consider transfer movements as revenue**. on the Cashflow report interface and it works very well

client/src/modules/reports/generate/budget_report/budget_report.config.js
jniles: You probably just needed to clear your browser cache, since the report loads the browser cache by default (line 22).  Not a big deal, just not very clean code.

server/controllers/finance/reports/budget_analytical/index.js
```diff
@@ -18,11 +18,17 @@ exports.report = report;
...
 async function report(req, res, next) {
   const params = req.query;
+  params.include_summary_section = parseInt(params.include_summary_section, 10);
+
   const fiscalYearId = parseInt(params.fiscal_id, 10);
   const setNumberYear = parseInt(params.set_number_year, 10) + 1;
   const colspanValue = parseInt(params.set_number_year, 10) + 6;
   const hideUnused = parseInt(params.hide_unused, 10);
   const { enterprise } = req.session;
+  const includeSummarySection = params.include_summary_section;
```
jniles: It looks like you have two variables for `includeSummarySection`.  Like this:

```js
params.include_summary_section = parseInt(params.include_summary_section, 10);
const includeSummarySection = params.include_summary_section;
```

In the future, just use one.

```js
const includeSummarySection = parseInt(params.include_summary_section, 10);
```

server/controllers/finance/reports/budget_analytical/index.js
jniles: The array `tabFiscalIncomeData` is the exact same as `tabIncomeDataBudget`.  Is there a reason they are identical?  Can you just reuse the `tabIncomeDataBudget` array?

If they should be identical, just make a copy of the array instead of another loop.
```js
// clone to use later for some reason.
const tabFiscalIncomeData = [...tabIncomeDataBudget];

File: server/controllers/finance/reports/budget_analytical/index.js [hunk 22]
```diff
@@ -43,9 +54,89 @@ async function report(req, res, next) {
...
+        fiscalYearId,
+        fiscalYear.start_date,
+        fiscalYear.end_date,
+        fiscalYear.start_date,
+        fiscalYear.end_date,
+        fiscalYear.start_date,
+        fiscalYear.end_date,
+      ];
+
+      const totalIncomeCash = await db.one(sqlTotalIncomeCash, paramsFilter);
```
jniles, line 113: This is a good variable name 👍

server/controllers/finance/reports/cashflow/index.js
```diff
@@ -381,6 +383,22 @@ function report(req, res, next) {
       return db.exec(queryRun, paramsRun);
     })
     .then(rows => {
+      // When the isTransferAsRevenue option is enabled,
```
jniles, line 388: Maybe add a FIXME that this is an IMCK-specific hack.

jniles (commented):
Code looks okay to me.  I'll wait for @jmcameron to provide his feedback as well, since he is more familiar with the budget work at IMCK.

jmcameron (changes requested):
@lomamech I encountered an error in basic tests (`npm run test` or `npm run build:db`).   While building the database, complains:

```
[build] default data
mysql: [Warning] Using a password on the command line interface can be insecure.
mysql: [Warning] Using a password on the command line interface can be insecure.
ERROR 1064 (42000) at line 351: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INSERT IGNORE INTO `flux` VALUES
  (1,  'STOCK_FLUX.FROM_PURCHASE'),
  (2,  'STO' at line 11
failed to import default data into DB 2/2
failed to build DB
```

I'm about to merge more PRs, so please merge with master again after you fix this issue.

Thanks

client/src/modules/reports/generate/budget_report/budget_report.config.js
lomamech: > You probably just needed to clear your browser cache, since the report loads the browser cache by default (line 22). Not a big deal, just not very clean code.

You are right, it works now.

lomamech:
> @lomamech I encountered an error in basic tests (`npm run test` or `npm run build:db`). While building the database, complains:
> 
> ```
> [build] default data
> mysql: [Warning] Using a password on the command line interface can be insecure.
> mysql: [Warning] Using a password on the command line interface can be insecure.
> ERROR 1064 (42000) at line 351: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INSERT IGNORE INTO `flux` VALUES
>   (1,  'STOCK_FLUX.FROM_PURCHASE'),
>   (2,  'STO' at line 11
> failed to import default data into DB 2/2
> failed to build DB
> ```
> 
> I'm about to merge more PRs, so please merge with master again after you fix this issue.
> 
> Thanks
@jmcameron I just encountered the same error, I will do an investigation
image

lomamech:
@jmcameron I just updated my PR

server/controllers/finance/reports/cashflow/index.js
jmcameron: @lomamech I encourage you to add an issue about this so we can generalize it later -- or base it on some environment variable that can be set for IMCK and ignored by other sites.

jmcameron (changes requested):
I've started working through the code.  I'll work on this more when the test issue below is resolved.

I encountered another testing problem.  One of the end-to-end tests fails because you added an account reference type but did not update a related test.  To reproduce the error, do this:
  `npx install playwright chromium`
  `npm run test:e2e-1`

lomamech:
> I encountered another testing problem. One of the end-to-end tests fails because you added an account reference type but did not update a related test. To reproduce the error, do this:

I just did it

jmcameron (approved):
I reviewed the code a couple of times.   I'm satisfied with it, but I request that @jniles take one more look at it to see if all the issues he brought up have been addressed.

I tested it with the IMCK database and the budget report seems to work well.   I was not completely sure how to reproduce the Cash flow report error, but I tested it and it seemed to work correctly.

All tests pass on my local test server.

@lomamech One suggestion for an improvement to the budget report.
At the end of the summary section, you list the Cashbox accounts but you use the phrase (English):  
"This report concerns these accounts"
Since many accounts are shown in the report, this is confusing.  I suggest you to change this to:
"This report concerns these cashbox accounts"

@lomamech I will be merging in other bump PRs, so I encourage you to merge with master once you have completed any changes.

client/src/modules/reports/generate/budget_report/budget_report.config.js
```diff
@@ -16,6 +16,8 @@ function BudgetReportController($sce, Notify, SavedReports, AppCache, reportData
     filter : 'default',
   };
 
+  vm.reportDetails.include_summary_section = 0;
```
jniles, line 19: 👍

client/src/modules/reports/generate/budget_report/budget_report.config.js
```diff
@@ -31,13 +33,29 @@ function BudgetReportController($sce, Notify, SavedReports, AppCache, reportData
...
+  };
+
   vm.numberYears = [
     { id : 1 }, { id : 2 }, { id : 3 }, { id : 4 }, { id : 5 },
   ];
 
   vm.preview = function preview(form) {
     if (form.$invalid) { return null; }
 
+    if (vm.reportDetails.include_summary_section === 0) {
```
jniles, line 55: 👍

server/controllers/finance/reports/budget_analytical/index.js
jniles: @lomamech that seems fine to me.

jniles (approved):
Changes look good to me.  Thanks for all your work on this @lomamech !

lomamech:
**How to Perform the Test with the BHIMA Test Database**  
1. Execute all integration tests: `npm run test:integration`.  
2. Post all the entries in the posting journal to the general ledger.  

To do this, you will need to unlock the following debtor groups:  
- Locked Debtor Group (Test)  
- Updated Debtor Group (Test)  

image

**After posting the entries to the general ledger, proceed as follows:**  

1. Navigate to the **Administration** menu and access the **Account References** module.  
2. Add the following two references:  

These references will enable the analysis of financial performance by calculating the total revenue minus the total expenses. Afterward, proceed with a specific analysis:  

1. **Results excluding revenue from non-medical activities**: This step provides a clear view of purely medical performance by isolating non-medical revenue.  
2. **Results excluding grants**: This evaluation offers insight into the hospital’s financial independence by excluding external contributions to measure its ability to generate local income.  

To integrate this evaluation into the BHIMA system, we have added a new reference titled "Account Type Reference – Budget Analysis".

This reference is configured to:

Identify and group revenue related to non-medical activities.
Analyze the relationship between local own revenue and grants while calculating the balance by subtracting the total expenses.

image

image

**Once the two account references are created and ordered such that the second account reference represents the total grants (this is a workaround), it enables the evaluation of local revenue excluding grants, the total grants, and the total revenue.**  

**The budget report interface is structured as follows:**  
(Provide the layout or additional details about the interface, such as sections, tables, or specific fields, if available.)  


image

**To view the summary section, simply click on "Yes".**  

This will allow you to select a cashbox, particularly the main cashbox.

image

An Overview of the Report

image

image

image

lomamech:
*To be able to test Cashflow*
These modules with the test database, it will first be necessary to create the grouping of account references related to the cash flow.

Accounting analysis:
The creation of this grouping is essential to ensure the consistency of financial data and to facilitate the analysis of cash flows. Accounts related to cash flow must be properly associated to guarantee an accurate overview of cash inflows and outflows. This also helps identify potential discrepancies in cash flow forecasts and effectively manage working capital needs.


*Simply go to the Administration Menu and configure the following account groupings.*

image
_For groupings with reference types **Expense Cashflow**_

1. Achat marchandises with accounts **601**
image

2. Charges Administratif with accounts **605, 61, 62, 65**
image

3. Charges personnel with accounts **66**
image

4.Impôt et taxes with accounts **64**
image

_For groupings with reference types **Income Cashflow**_
1. Ventes biens et services with accounts 701, 706, 58511010

image

2. Subventions with accounts **71**
image

3. Autre produits with accounts **75, 77**
image

After configuring the various account references, the groupings enable the visualization of the cash flow report with specific parameters, such as selecting "This Month." We have also added the option **"Consider transfer movements as revenue"**. This feature allows transactions of the type "Other" to be treated as revenue, facilitating the tracking of funds transferred from auxiliary cash registers to the main cash register.

image

image
- Here is a preview of the report without considering transfers as revenue.

image
- Here is a preview of the report with transfers considered as revenue.

lomamech:
@jniles @jmcameron I have just completed the description on how to test my latest PR using the Test Database. For this, I had to make several additions to the test database.

jniles:
It looks like this breaks the end to end tests for the posting journal edits,  unfortunately.  I'll try to look into this next week if you don't get to it before.

image

jmcameron:
@lomamech Sorry this PR is taking a while to merge!

I think that some of the added items have caused some of the end-to-end tests to fail.

---
**In `npm run test:e2e-2`:**
```
  1) [chromium] › test/end-to-end/editJournal/edit_journal.spec.js:72:3 › Edit Posting Journal › edits a transaction to change the value of debit and credit

    Error: Expected locator #validation-errored-alert to not exist.

    expect(received).toBe(expected) // Object.is equality

    Expected: false
    Received: true

       at test/end-to-end/shared/TestUtils.js:215

      213 |     return expect(count > 0,
      214 |       `Expected locator ${selector} to ${bool ? '' : 'not '}exist.`,
    > 215 |     ).toBe(bool);
          |       ^
      216 |   },
      217 |
      218 |   /**
        at Object.exists (/home/jmcameron/Bhima/test/end-to-end/shared/TestUtils.js:215:7)
        at /home/jmcameron/Bhima/test/end-to-end/editJournal/edit_journal.spec.js:85:5
```
---
**In `npm run test:e2e-3`:**
```
  1) [chromium] › test/end-to-end/journal/SearchModal.tests.js:59:3 › Posting Journal › Search Tests › finds 0 rows for transfer transaction type

    Error: expect(received).toBe(expected) // Object.is equality

    Expected: 0
    Received: 2

       at test/end-to-end/shared/GridUtils.js:122

      120 |     expect(number.includes(rows.length), message);
      121 |   } else {
    > 122 |     expect(rows.length, message).toBe(number);
          |                                  ^
      123 |   }
      124 |
      125 |   // ADAPTED FROM PROTRACTOR CODE:
        at Object.expectRowCount (/home/jmcameron/Bhima/test/end-to-end/shared/GridUtils.js:122:34)
        at /home/jmcameron/Bhima/test/end-to-end/journal/SearchModal.tests.js:62:5
```
---

The rest of the end-to-end tests pass.  Please fix these tests.

lomamech:
image
@jmcameron @jniles  all tests pass

jniles:
I am still getting test errors using `npm run test:e2e-all`.  I'll work on a fix for this PR today.

jniles:
Okay, I think I fixed it.

@jmcameron see if it works on your machine.

jmcameron (approved):
All looks good.
All test pass.
I tried the test using Bhime test DB and that worked fine.

PR #7906, opened 2024-11-30
Standardize email varchar length

jniles:
As I was writing some test scripts, I noticed that we have a really short length for the `email` field for patients.  I've standardized all `email` field lengths to `VARCHAR(150)`, which was our longest supported email field.


jmcameron (approved):
The changes look good.
I did not try the migration code, but it seems very straightforward and unlikely to be a problem

All regular tests pass.

PR #7933, opened 2024-12-14
chore: clean up bash scripts

jniles:
This commit removes unused or deprecated scripts in the BHIMA repository and also improves some of the other bash scripts after passing through shellcheck linting.  This should relieve some maintenance burden on the developers.

Maintenance scripts (such as the install script, etc) should be in a separate repository third-culture-software/bhima-scripts.

To test this, just run the regular tests, as well as create a migration file (`npm run migrate`).  All should complete fully and without error.


sh/setup-migration-script.sh
```diff
@@ -2,6 +2,14 @@
 
 # bash strict mode
 set -uo pipefail
+trap 'echo "Error: Command failed on line $LINENO"; exit 1' ERR
```
jmcameron, line 5: Nice!  👍

sh/test-e2e-all.sh
```diff
@@ -22,33 +22,26 @@ function startfold {
...
-#     echo -en "travis_fold:end:$1\r"
-#   fi
-# }
-
 SUITE=${SUITE:-"ALL"}
 
 # run end to end account tests with Playwright
 for i in {1..8}; do
   date
-  startfold "Running account End to End Tests $i ...";
```
jmcameron, line 36: Since we are no longer doing "Folds", perhaps it makes sense to rename this function to something more obvious.

sh/test-e2e-all.sh
```diff
@@ -22,33 +22,26 @@ function startfold {
...
 date
 
 # Delete left-over zombie server process
-procs=`netstat -tulpn |& grep 8080` || true
-proc=`echo $procs | sed -r 's/.* ([0-9]+)\/node$/\1/g'`
-if [[ ! -z "$proc" ]]
-then
-    echo "Deleting zombie node Bhima process $proc"
-    kill -9 $proc || true
+procs=$(netstat -tulpn |& grep 8080) || true
```
jmcameron, line 37: Just out of curiosity, why is "$(...)" preferred to "'...'" ?

jmcameron (approved):
The code changes look fine.
I had one minor comment about function names in the migration script.
All the tests pass.

sh/test-e2e-all.sh
jniles: The answer applicable to this line is that it was done to satisfy a shell linter (https://www.shellcheck.net/).

[I've been told](https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts) in general that `$()` is preferred due to the ability to more easily nest commands, though that isn't applicable here.

sh/test-e2e-all.sh
jniles: Done.

PR #7938, opened 2024-12-17
refactor(*): replace JSON.parse/JSON.stringify with structuredClone

jniles:
Replaces `JSON.parse`/`JSON.stringify` with `structuredClone()` native function for cloning.  Previously, we used a hacky workaround for cloning that destroyed JS dates.  Now, there is a supported native function for doing this operation in all versions of NodeJS > 17.

Ref: https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone.


jniles:
I am merging this myself since it shouldn't be controversial.

PR #7948, opened 2024-12-29
Minor UI tweaks to Employee Registration

jniles:
This PR does some minor UI tweaks to the employee registration page.  It adds a labels above the "Rubrics" section to clearly identify that they are rubrics, and moves the rubrics section to the bottom of the optional information in the form.

Additionally, I added an optional flag to the `bhCurrencyInput` to support `form-horizontal` cases.  It is fixed a `col-md-3` since that is what is used on this page.  I'm conflicted about if having the `form-horizontal` is even a good idea here, but we'll keep it until there is more data than just my own personal feeling.

Major code changes are the updates to the Currency Service to bring it in line with modern ECMAScript.

Here is what it looks like using the test database:

image

### DIFF
--- a/client/src/modules/employees/registration/employees.html
+++ b/client/src/modules/employees/registration/employees.html
@@ -292,7 +292,7 @@
                         name="dateEmbauche"
                         type="text"
                         class="form-control"
-                        ng-model="EmployeeCtrl.employee.date_embauche"
+                        ng-model="EmployeeCtrl.employee.hiring_date"
                         uib-datepicker-popup="{{EmployeeCtrl.datepickerOptions.format}}"
                         datepicker-options="EmployeeCtrl.datepickerOptions"
                         is-open="EmployeeCtrl.dateEmbauchePopupIsOpen"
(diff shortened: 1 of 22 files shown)


client/src/modules/employees/registration/employees.html
```diff
@@ -515,20 +504,23 @@ <h4 translate>FORM.LABELS.CURRENT_LOCATION</h4>
...
-                    <label class="control-label">
-                      <input
-                        type="checkbox"
-                        name="locked"
-                        ng-true-value="1"
-                        ng-false-value="0"
-                        ng-model="EmployeeCtrl.employee.locked">
-                        <span translate>FORM.LABELS.LOCKED</span>
-                    </label>
+                  <h4 class="text-capitalize" translate>FORM.LABELS.RUBRICS</h4>
```
jmcameron: I suggest making the "Rubrics" text bold to better separate the section.

jmcameron (approved):
It looks good.  I made one minor suggestion about formatting.
The tests all run okay.   And I tested it manually with bhima_test and works fine.
Thanks!

jniles:
Thanks for your review @jmcameron.

I've made the requested change:

image

PR #7952, opened 2024-12-31
Unify Rubric Configuration Modal

jniles:
The goal of this PR is to reduce complexity, both for users and developers.  It accomplishes this by removing unnecessary steps for the user, and by namespacing and reducing the amount of code for developers.

Update: it also fixes a bug in the IPR Config client code.  See the "Bug Fix" section below.

Specifically, the rubric  configuration modal UI and API have been updated to remove separate routes and interfaces for the `/setting` routes that configured `rubric_config_item`s.  This creates a single modal for the user to manage.  On the developer side, it namespaces the API route to behind the `payroll/` prefix and updates the data model in transit to make `rubric_config_item`s a property of the `rubric_config` option, similar to how invoices, vouchers, and cash payments work with their associated `_item`s.  

Here is the updated User Interface.
image

In the future, we might use the `bh-checkbox-tree` for the rubrics, but that relies on having a sane `classify()` function for the different types of rubrics.  At the moment, the rubric classification is not well enough understood by me to make those kinds of changes.

#### Bug Fix: IPR Config
In addition to the refactoring of the Rubric Configuration interface, I have also fixed a race condition in the IPR tax configuration modal.  In some cases the `vm.taxIprId` was not defined, which could lead to undefined behavior.  I added the `$q` module to do the following:
  1. Ensure that all promises are settled before the modal is usable.  Also ensure that variables used between HTTP connections are ordered correctly.
  2. Ensure that there is common error handling for all promises/HTTP requests.
  3. Clarify the control flow between the update/create states in the modal.
  
Additionally, I also refactored the URL handling to remove the duplicate `configuration` text in the URL.  Previously, the URL for editing a configuration was `/ipr_tax/configuration/1/configuration/4/edit`. Now, it is simply `/ipr_tax/configuration/1/edit/4`, which more closely follows standard REST conventions of `namespace/{id}` than previously.


jniles:
@jmcameron, I'm pretty sure this is finally able to be reviewed when you get a chance!

EDIT: Finally got the end to end tests stable here.

jniles:
I'm going to go ahead and merge this into the `refactor-payroll-management` mega branch.  The idea of having sub-branches was to make things easier to review, but it isn't an ideal time for the rest of the team to give me reviews.

PR #7954, opened 2025-01-01
fix(tests): add FY 2025

jniles:
Adds the fiscal year 2025 to the test database.


jniles:
I'm merging this since it should fix the integration tests.  It also removes moment from the test library! (https://github.com/Third-Culture-Software/bhima/issues/7872)

Issue #6097, opened 2021-11-17
Payroll cosmetic requests


jmcameron:
In the recent Payroll-related releases (https://github.com/IMA-WorldHealth/bhima/pull/6072 and https://github.com/IMA-WorldHealth/bhima/pull/6079) we ran out of time to perform some cosmetic improvements:
- Changing English "Payroll Bulletins" to "Payroll Payslips" has already been done
- In English "nontaxable" is a more current and proper spelling than "non-taxable".  See https://english.stackexchange.com/questions/458357/non-taxable-vs-non-taxable.
- It looks like these payslips are designed to print in "Landscape" mode, one per page. However the payslip will work very nicely in "Portrait" mode. So if you're printing one payslip (action menu), then the current approach is fine although "Portrait" would also work. However, for multiple payslips, it would be good to print them in "Portrait" mode, 2 per page. That would save paper at the production sites. Perhaps a dotted line could be added in the middle of the page to show where to cut the two apart.  
**RESPONSE:** At IMCK he chooses the option of printing two payslips on a single page thanks to the parameters of the printer, but I think we can configure it.   (From https://github.com/IMA-WorldHealth/bhima/pull/6079#issuecomment-968087024)
- A few suggestions regarding the payslips (see https://github.com/IMA-WorldHealth/bhima/pull/6079#issuecomment-968087024)
   - I see the English term "Net Non-Taxable" on the payroll payslips. I think this conveys the wrong idea: In typical financial term "net" means total minus something (eg net profit = income - expenses). So I think the term should be "Total Non-taxable"
   - On the payslips, when using the Enterprise currency, there is no need to print the Exchange rate line. It will always be 1 !
   - If there is any chance that the payroll method might change at a particular site, it might make sense to include a code for that on the Payslip. It does not need to be understandable by the employee, but available for future reference for the site financial administrators.

A small bug on the display of filters on the Multiple Payroll page.  Note that the filter terms are not being transformed into anything meaningful like for currency, it should show a monetary symbol, not the currency ID number
![image](https://user-images.githubusercontent.com/62145/142095310-d9b4578c-f7b3-47e4-bb82-2bc8ba7159e3.png)

jniles:
Noting down which of these have been completed (as of https://github.com/Third-Culture-Software/bhima/pull/7967)

 - [x] Changing English "Payroll Bulletins" to "Payroll Payslips" has already been done
[...]

PR #7967, opened 2025-01-12
chore(payroll): address cosmetic requests

jniles:
Addresses a number of cosmetic requests discussed in #6097.  Notably, the English syntax for "nontaxable" and "net" have been addressed. Additionally, the currency exchange rate is not rendered on the payslip if the currency is the enterprise currency.  However, there is a bug with the currency rendering on the payslip generator that does not allow the currency to be properly rendered.  This will be addressed in a future PR.

Partially addresses #6097.


jniles:
I'm going to go ahead and merge this into the `refactor-payroll-management` mega branch. The idea of having sub-branches was to make things easier to review, but it isn't an ideal time for the rest of the team to give me reviews.

PR #7979, opened 2025-01-20
Fix payroll currency bug

### LINKED ISSUE #7966: Multiple Payroll "Configuration" modal will increase the base salary indefinitely if the currency is switched too many times.
In the `Human Resources` > `Payroll Multiple` configuration modal for an individual employee, when a user changes the currency two unexpected things occur:

 1. The currency in the `bh-currency-input` does not update (it always shows the enterprise currency, I would assume).
 2. The "base salary" increases indefinitely by running the exchange rate algorithm multiple times and updated the base value.  This is a huge bug!  Instead, the original base salary should be cached and all exchanges should be run against the cached value.
 
 Here is a demonstration of the bug:
PayrollCurrencyBug

When fixing this issue, the following improvements could also be made:

 1. Instead of "Employee Indice:2" in the header of the modal, we should display the employee's name in the modal body.
 2. We could also put the payment period in the modal body as well.
 3. Instead of `disabled` inputs, let's just use [Static Controls](https://getbootstrap.com/docs/3.4/css/#forms-controls-static) to render the options that cannot be displayed.
 4. The "Jours Feries", etc, are all days.  Let's put the units on the inputs.  For the best user experience, it should say "Holidays" "None", or "1 day", "2 days", etc.  But "0 days" is also likely acceptable.
 5. The "Money" input should occur above the "Base Salary" input.

jniles:
This PR fixes the payroll configuration currency bug reported in #7966.  It does so be completely removing the currency selection (see discussions in https://github.com/Third-Culture-Software/bhima/issues/7966 and https://github.com/Third-Culture-Software/bhima/issues/4956#issuecomment-2602873015). Some small touches include making the translation and function keys all English values. 

## Routing Improvements
I've improved the ui-routing of the `multiple_payroll` module so that only minimal information is passed into the `config` modal and the both the parent and modal controllers are less reliant on parsing the `vm.latestViewValues`.  There is still some work to be done in this area to completely eliminate this dependency in the `multiple_payroll` controller, but the configuration modal is now state agnostic.

## UI Improvements
I've added additional contextual information to the `ConfigPaymentModal` for the user to orient them about which employee is being configured.  In particular, the employee name, grade, and number of dependents are all prominently displayed.  See below:

image

Closes #7966.

See  PR https://github.com/Third-Culture-Software/bhima/pull/7494 for a description on how to test a payroll with indeces.


client/src/modules/multiple_payroll/modals/config.modal.html
```diff
@@ -1,79 +1,124 @@
-<form name="ConfigPaiementForm" bh-submit="ConfigPaiementModalCtrl.submit(ConfigPaiementForm)" novalidate>
+<form name="ConfigPaymentForm" bh-submit="ConfigPaymentModalCtrl.submit(ConfigPaymentForm)" novalidate>
   <div class="modal-header">
     <ol class="headercrumb">
       <li class="static" translate>TREE.HUMAN_RESOURCES</li>
-      <li class="title" translate>TREE.MULTI_PAYROLL</li>
-      <label class="badge badge-success">{{ ConfigPaiementModalCtrl.employee.display_name }}</label>
-      <label class="badge badge-primary"><span translate> {{ ConfigPaiementModalCtrl.label }} </span> : {{ ConfigPaiementModalCtrl.displayValue }}</label>
+      <li class="static" translate>TREE.MULTI_PAYROLL</li>
+      <li class="title" translate>{{::ConfigPaymentModalCtrl.period.label}}</li>
```
jmcameron, line 6: This does not seem to work correctly.  Displays as '2', not USD, etc

client/src/modules/multiple_payroll/modals/config.modal.js
```diff
@@ -1,151 +1,138 @@
...
 ];
 
-function ConfigPaiementModalController(
-  $state, Notify, AppCache, Employees, MultiplePayroll, Configuration,
-  Exchange, Session, params,
+/**
+ * @function ConfigPaymentModalController
+ *
+ * @description
+ *  Configures an employee for payment, allowing the user to adjust the employee's remuneration
```
jmcameron, line 13: Thanks for adding function comments!

client/src/modules/multiple_payroll/modals/config.modal.js
```diff
@@ -1,151 +1,138 @@
...
   vm.config = {};
   vm.payroll = {};
 
   vm.enterprise = Session.enterprise;
-  vm.lastExchangeRate = {};
-
-  const cache = AppCache('multiple-payroll-grid');
 
-  if (params.isCreateState || params.uuid) {
+  if (params.employeeUuid) {
```
jmcameron, line 37: 👍

client/src/modules/multiple_payroll/multiple_payroll.routes.js
```diff
@@ -25,7 +23,7 @@ function configurationMultiplePayroll($modal, $transition) {
   $modal.open({
     size : 'lg',
     templateUrl : 'modules/multiple_payroll/modals/config.modal.html',
-    controller : 'ConfigPaiementModalController as ConfigPaiementModalCtrl',
+    controller : 'ConfigPaymentModalController as ConfigPaymentModalCtrl',
```
jmcameron, line 26: Thank you for the language fix!

jmcameron (changes requested):
I did a casual read-thru of the code -- which looks good.   I especially appreciated the new function comments!  Good job!

However, during testing, I ran into several questions and potential improvements.  See below.

The employee configuration form looks good (in terms of content), but it seems like there is excess white space (particularly between the label and the info for that label). It seems like most of these could be put onto the same line, eg,  "Label: <value>".  That would make the form more compact and reduce scrolling needed on smaller displays:
image


In Multiple Payroll, when you select the pay period, the pop-up still includes a selector for the currency.  Shouldn't it be deleted too?
image

Question: Does it make sense to include the hiring date on the employee configuration page (along with their name, function, etc)?

Minor Problem that it would be good to fix:  In the initial display for Multiple Payroll, the "Currency" search term is not translated properly.  Also note that the default period of payment seems to be cut off? But is okay when a new one is selected): 
image

Please update package.json to the latest on master so we can keep them in sync.  Thanks.

jniles:
@jmcameron, thanks for the review!

> The employee configuration form looks good (in terms of content), but it seems like there is excess white space (particularly between the label and the info for that label). It seems like most of these could be put onto the same line, eg, "Label: ". That would make the form more compact and reduce scrolling needed on smaller displays:

A good suggestion.  I'll take a look at reducing the whitespace here. 

>In Multiple Payroll, when you select the pay period, the pop-up still includes a selector for the currency. Shouldn't it be deleted too?

Maybe ... right now, this popup changes the displayed currency of the grid.  This doesn't change anything about the actual salaries or payment period, but is more for a user to view.  I'm not sure it is useful, and may end up removing it, but I need to do some more research about it before I make that leap.  You could potentially see its usefulness in allowing users to see the total value in USD of their payroll obligations, for example.

> Question: Does it make sense to include the hiring date on the employee configuration page (along with their name, function, etc)?

That's easy to add.  I could do that.

> Minor Problem that it would be good to fix: In the initial display for Multiple Payroll, the "Currency" search term is not translated properly. Also note that the default period of payment seems to be cut off? But is okay when a new one is selected):

This is likely a bit harder to address, so I am going to kick it down the road for a future PR.  Depends a bit on how we end up handling currencies in payroll in the long run, and we may end up removing this filter.  I'd rather do that first than invest time into something we may remove.

jniles:
I've tried to compress the whitespace a little bit.  I don't want to go too far outside of bootstrap's own style since that would add an additional maintenance burden and probably would result in some technical debt.  Here is what it looks like now:

image

client/src/modules/multiple_payroll/modals/config.modal.html
jniles: You are commenting on a line in the `config` modal, but I think you are referring to the $USD value in the filters for the `multiple_payroll` module.   If so, then that is a long-standing issue that won't be addressed in this PR. 

I'm not super concerned about it, since the user has other cues about the currency displayed.

jniles:
@jmcameron I've completed the fixes suggested.  Let me know what the next steps are.

Issue #7980, opened 2025-01-22
Limit the ability to incur debt on a specific account


jniles:
At Vanga, we have received the request to limit the invoicing of a patient if they are associated with a debtor group that has maxed out its debt limit.  That is, when we check to load the debtor group, the invoicing module should show warning that disables the submission of the debtor group has a balance over the limit.

### Design
To accomplish this, I'll rename the `max_credit` field on the `debtor_group` table to `max_debt`.  I'll change the translation keys around this to "Overdraft Limit", with the description talking about how this will prevent billing patients after the account balance surpasses the limit.

Then, I'll add two checks to the `PatientInvoiceFormService` when the patient is loaded.  First, it will check the account balance using the `AccountService.getBalance()` function. Next, it will call `DebtorGroup.read()` to check the `max_debt` field from the patient's debtor group.  If the `max_debt` value is non-zero, and the account balance (debit - credit) exceeds the `max_debt` value, the `PatientInvoiceFormService` will set an error flag informing the user that this patient's debtor group has a debt that exceeds the institution's policy, and the patient cannot be billed in this debtor group.

Finally, I'll add the same logic to the `patientInvoice.create.js` service-side controller to prevent the creation of invoices via the API if the same conditions aren't met.

### Considerations
Some implementation notes:

 1. We could limit credit/debit balances on the account itself, instead of the debtor group.  However, this would force us to have a good UI/UX to resolve issues across the entire application (purchasing, vouchers, transaction edit modal, etc).  That design is too big a challenge for this issue.
 2. This will not protect the debtor group from being "over drafted" in the traditional sense.  If the limit is $500 on a debtor group, they can still process a bill for $510.  However, the next invoice will fail.  This seems like a reasonable tradeoff to me.
 3. To prevent users from working around this restriction by never posting, the account balance will be the balance _including the posting journal_.  This means that a debtor group may exceed its max debt dynamically between two invoices.  It also means that clever accountants cannot work around the issue by simply never posting.

jniles:
It sounds like we will need to do another release with the final updates to reports from @lomamech.  I'll take this opportunity to add in this change requested by Vanga, since they are encountering administrative difficulties over account spending limits.

jniles:
Note: this will also need to be applied to the "Prise en charge" journal voucher method too.

PR #7987, opened 2025-01-26
Add account spend limit to debtor group

jniles:
This PR implements the client feature request in #7980.  It blocks the creation of invoices if the `max_debt` field has been surpassed on patient invoicing in the UI and in the server API level.

## Demo
Below, you can see this in action:
DebtorGroupOverdraftLimit

To test this PR, simply decrease the max_debt limit to something like $1 on a debtor group with patients, then try to bill any of those patients.  If the debtor group does not have any transactions, the first will succeed.  But any further invoices will fail.


jniles:
@jmcameron this is ready for your review.

jniles:
Tests pass locally and in the cloud.  Merging this code.

PR #8009, opened 2025-02-17
Add the support of funding source in stock management

mbayopanda:
This PR adds:

1. A funding source module where we can manage CRUD about funding source

image

2. Enable or not the support of funding source during stock entry 

image

image

3. Search stock lots or assets by funding source

image

image


jniles:
@mbayopanda thanks for the PR and the screenshots!

Can you describe what the intended use case for Funding Sources?  I'm not familiar with the concept or who will be using it.  It will help when reviewing the functionality.

Does the "Funding Source" concept apply only to assets?  Or also to consumable stock?  Or only to consumable stock?

I assume that we probably shouldn't allow the merging of lots that have two different funding sources, right?   Currently, I think the algorithm checks only the lot name and expiration date.

Lots are now uniquely identified by (`lot.label`, `funding_source.label`), since two of the same lot labels could have two different funding sources?  Or should we insist that the same lot label have the same funding source?  (I'm not sure how this works in practice.)

mbayopanda:
@jniles thank you for your feedback

The funding source is a way to categorize lots by project in the case of humanitarian organisations such as IMA, stock manager must have a way to find which lots (products) belongs to a given project (funding source)

The term `funding source` has been taken from Jericho.

You're right, we shouldn't allow the merging of lots that have two different funding sources when this feature is enabled.

client/src/js/components/bhFundingSourceSelect/bhFundingSourceSelect.js
```diff
@@ -0,0 +1,55 @@
+angular.module('bhima.components')
+  .component('bhFundingSourceSelect', {
+    templateUrl : 'js/components/bhFundingSourceSelect/bhFundingSourceSelect.html',
+    controller  : FundingSourceSelectController,
+    transclude  : true,
+    bindings    : {
+      fundingSourceUuid : '<',
```
jniles, line 7: If this parameter is optional (and it looks like it is on line 35), I would put a `?` on it.

client/src/js/components/bhFundingSourceSelect/bhFundingSourceSelect.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Funding source selection component
+ */
+function FundingSourceSelectController($rootScope, FundingSources, Notify) {
+  const $ctrl = this;
+
+  $ctrl.createUpdateFundingSourcesModal = FundingSources.createUpdateFundingSourcesModal;
+
+  $rootScope.$on('FUNDING_SOURCES_CHANGED', () => {
```
jniles: I'm not sure how I feel about this design.  We generally have been trying to avoid using event emitters in favor of component callbacks.

There's two issues I see with this:
 1. It's difficult to know what is subscribed to the event emitter, or how many different times things have subscribed to the same event emitter.  This makes the code harder to reason about (and maintain) as a developer.
 2.  It's hard to know how Angularjs will handle unsubscribe-ing.  For example, if page navigation changes, will it automatically remove this subscription?  Or will it still send `loadFundingSources()` whenever the funding source select component is updated (for example, in a select on a different page)?  
 
 The way we handle this in the `bhFindPatient` component is to offer an API to the parent controller to be able to call things like `reset()` when needed:
https://github.com/Third-Culture-Software/bhima/blob/3001aef0153ccda547dcd3fea1d5bd5ce387e65f/client/src/js/components/bhFindPatient.js#L78-L82

I recommend that solution here, since it makes it explicit who is using this API.

client/src/js/components/bhFundingSourceSelect/bhFundingSourceSelect.js
```diff
@@ -0,0 +1,55 @@
...
+  .component('bhFundingSourceSelect', {
+    templateUrl : 'js/components/bhFundingSourceSelect/bhFundingSourceSelect.html',
+    controller  : FundingSourceSelectController,
+    transclude  : true,
+    bindings    : {
+      fundingSourceUuid : '<',
+      onSelectCallback  : '&',
+      required          : '<?',
+      label             : '@?',
+      enableAdd         : '<?',
```
jniles: This is cool! 👍 

If you don't need to dynamically update this, I would recommend you make it a `@` binding instead of a `<` binding to remove a `$watch`-er on this component interface.

server/config/routes.js
```diff
@@ -1218,4 +1219,11 @@ exports.configure = function configure(app) {
   app.post('/shipment_containers', shipmentContainer.create);
   app.put('/shipment_containers/:uuid', shipmentContainer.update);
   app.delete('/shipment_containers/:uuid', shipmentContainer.deleteContainer);
+
+  // funding sources
+  app.get('/funding_sources', fundingSource.read);
```
jniles, line 1224: Nice API design!  Is there a vision to use these funding sources outside of stock management?

Otherwise, we might put them under the `/stock/` prefix.

client/src/js/components/bhFundingSourceSelect/bhFundingSourceSelect.html
```diff
@@ -0,0 +1,34 @@
+<div ng-form="FundingSourceForm" bh-funding-source-select ng-model-options="{ updateOn: 'default' }">
+  <div
+    class="form-group"
+    ng-class="{ 'has-error' : FundingSourceForm.$submitted && FundingSourceForm.fundingSourceUuid.$invalid }">
+
+    <label class="control-label">
+      <span translate>{{ $ctrl.label }}</span>
+      <span ng-if="$ctrl.enableAdd">&nbsp;&nbsp;<a style="cursor: pointer;" ng-click="$ctrl.createUpdateFundingSourcesModal()" translate>FORM.LABELS.ADD</a></span>
```
jniles: It doesn't look like you ever enable `enable-add` in your PR.  I did a quick check, and it is `false` everywhere.  I would recommend just dropping this functionality if it isn't used in the current version.

If you have a plan to enable it later in a specific module, I would still remove it from this PR and add it back in later when you are going to actually use it.

client/src/modules/assets/modals/search.modal.js
```diff
@@ -33,6 +34,12 @@ function AssetsSearchModalController(data, util, Store, Instance, Stock, SearchM
     vm.searchQueries.is_assigned = vm.searchQueries.is_assigned === 'ASSET.SHOW_ONLY_ASSIGNED' ? 1 : 2;
   }
 
+  // Hanlde the Funding Source
```
jniles: Small typo:

Hanlde => handle

client/src/modules/fundingSources/modal/createUpdate.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    const operation = vm.isCreation
+      ? FundingSourcesService.create(vm.data)
+      : FundingSourcesService.update(data.uuid, vm.data);
+
+    return operation
+      .then(() => {
+        Notify.success('FORM.INFO.OPERATION_SUCCESS');
+        $rootScope.$broadcast('FUNDING_SOURCES_CHANGED', true);
```
jniles: Instead of `$broadcast`, why don't you return the `uuid` of the Funding Source that was created or updated from the modal?

client/src/modules/fundingSources/fundingSources.service.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @description
+ * Encapsulates common requests to the /funding_sources/ URL.
+ */
+function FundingSourceService(Api, util, $uibModal) {
+  const baseUrl = '/funding_sources/';
+  const tagKeys = ['uuid', 'label', 'code'];
+  const service = new Api(baseUrl);
+
+  service.clean = tag => util.maskObjectFromKeys(tag, tagKeys);
```
jniles: It doesn't look like `clean()` is ever used. I would remove it.

client/src/modules/fundingSources/fundingSources.service.js
```diff
@@ -0,0 +1,29 @@
...
+
+  service.createUpdateFundingSourcesModal = (fundingSource) => {
+    $uibModal.open({
```
jniles: You should call `return $uibModal.open({ ... }).instance` here, so that the consumers of this API will be notified when the modal has closed.

client/src/modules/fundingSources/templates/action.cell.html
```diff
@@ -0,0 +1,29 @@
+<div
+  class="ui-grid-cell-contents text-right"
+  data-row="{{row.entity.name}}"
```
jniles: What is `row.entity.name` here?

server/controllers/admin/fundingSources.js
```diff
@@ -0,0 +1,73 @@
...
+};
+
+// add a new tag
+function create(req, res, next) {
+  const sql = `INSERT INTO funding_source SET ?`;
+  const data = req.body;
+  data.uuid = data.uuid ? db.bid(data.uuid) : db.uuid();
+  db.exec(sql, data)
+    .then(() => {
+      res.sendStatus(201);
```
jniles: Could you make these CRUD APIs return the same values as the rest of the BHIMA CRUD APIs?

https://github.com/Third-Culture-Software/bhima/wiki/Design:-Server-CRUD-APIs

Create should return the `id`/`uuid` of the resource created.

server/controllers/admin/fundingSources.js
```diff
@@ -0,0 +1,73 @@
...
+    }).catch(next);
+}
+
+// update tag information
```
jniles: These are not tags 😆

server/controllers/admin/fundingSources.js
```diff
@@ -0,0 +1,73 @@
...
+function update(req, res, next) {
+  const sql = `UPDATE funding_source SET ?  WHERE uuid =?`;
+  const data = req.body;
+  delete data.uuid;
+  const uuid = db.bid(req.params.uuid);
+
+  db.exec(sql, [data, uuid])
+    .then(() => {
+      res.sendStatus(200);
```
jniles: As above, the `update` route should return `200 (OK), 404 (NOT FOUND), 400 (BAD REQUEST)`, depending on the result.

The update function should return a JSON object with the full resource.

server/controllers/stock/reports/stock_report.report.handlebars
```diff
@@ -0,0 +1,68 @@
+{{> head }}
+
+<body class="container-fluid">
+  {{#> header}}
+    {{!-- <h3 class="text-right">{{translate 'REPORT.STOCK_REPORT.TITLE'}}</h3> --}}
```
jniles: If these aren't being used, please remove them.

server/models/migrations/next/migrate.sql
```diff
@@ -8,3 +8,29 @@
...
+
+-- @mbayopanda: Add entry in the unit table for funding sources
+INSERT INTO `unit` VALUES 
+  (321, 'Funding Source','TREE.FUNDING_SOURCES','',1,'/funding_sources');
+
+-- @mbayopanda: Update the lot table to support funding source for each lot
+CALL add_column_if_missing('lot', 'funding_source_uuid', 'BINARY(16) NULL');
+
+-- @mbayopanda: Update the stock_setting table to enable to funding source
+CALL add_column_if_missing('stock_setting', 'enable_funding_source', 'TINYINT(1) NOT NULL DEFAULT 0');
```
jniles, line 33: Do we need something about a `stock_report` in the reports here?

test/integration/fundingSources.js
```diff
@@ -0,0 +1,109 @@
...
+      .then((res) => {
+        expect(res).to.have.status(201);
+      })
+      .catch(helpers.handler);
+  });
+
+  it('GET /funding_sources returns a list of funding sources', () => {
+    return agent.get('/funding_sources')
+      .then((res) => {
+        expect(res).to.have.status(200);
```
jniles: Could you use the standard `helpers.api.listed()` style methods here?  That way, you'll know that your API follows the typical BHIMA API guidelines.

test/integration/fundingSources.js
```diff
@@ -0,0 +1,109 @@
...
+
+const helpers = require('./helpers');
+
+/*
+ * The /funding_sources API
+ *
+ * This test suite implements full CRUD on the /funding_sources API.
+ */
+describe('test/integration/funding_sources The funding_sources API', () => {
+  // project we will add during this test suite.
```
jniles: This is probably not the correct comment 😆

server/controllers/stock/functions/stock_report.function.js
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * @method getData
+ *
+ * @description
+ * This method builds the monthly stock report
+ *
+ * GET /stock/stock_report
+ */
+async function getData(reqQuery) {
```
jniles: Hmm... A lot of this code looks identical to the code in `rumer.function.js`.  What exactly is different?  Could you maybe abstract out the common code into a common file that is used by both the `rumer` and this report?

https://github.com/Third-Culture-Software/bhima/blob/master/server/controllers/stock/functions/rumer.function.js

server/models/migrations/next/migrate.sql
```diff
@@ -8,3 +8,29 @@
...
+
+-- @mbayopanda: Update the lot table to support funding source for each lot
+CALL add_column_if_missing('lot', 'funding_source_uuid', 'BINARY(16) NULL');
+
+-- @mbayopanda: Add entry in the unit table for funding sources
+INSERT INTO `unit` VALUES 
+  (321, 'Funding Source','TREE.FUNDING_SOURCES','',1,'/funding_sources');
+
+-- @mbayopanda: Update the lot table to support funding source for each lot
+CALL add_column_if_missing('lot', 'funding_source_uuid', 'BINARY(16) NULL');
```
jniles: It looks like you duplicated code here.  You already have this mentioned on line 26.

jniles (changes requested):
@mbayopanda I've gone through and reviewed the code.  I haven't tested the functionality.  

In general, it's pretty clean code.  Well done!

There are a couple of changes that would be good to make:
 1. Make sure your API corresponds to the typical API return codes that BHIMA is supposed to conform to.  I know not every API does, but new ones should, to make maintaining easier.
 2. You added in a new report that is almost identical to the RUMER.  You also didn't add a key for it in the `report` table, and didn't mention it in your PR description.  Was this additional report a `git add` error?  I didn't understand what it was for or why it was different from the RUMER.
 3. I would remove the event-emitter style from BHIMA (e.g. `$rootScope.$on()`).    We experimented with it early in BHIMA, but it is just hard to maintain.  Now that we have promises, it's pretty straightforward to pass promises across component/modal/service boundaries to accomplish the same goal of notifying parent modules when a child component has updated data.  If it isn't easy to do that, it probably means your design aren't well thought out.
 
Once these changes are made, I'd be happy to review it again.  👍

server/controllers/stock/functions/stock_report.function.js
mbayopanda: The report has not been implemented, so the code must not be there. My bad

mbayopanda:
@jniles thank you for comments, "Meticulous" as usual. Thx

server/config/routes.js
mbayopanda: Not only in stock

mbayopanda:
@jniles could you review this PR please.

client/src/modules/fundingSources/modal/createUpdate.html
```diff
@@ -0,0 +1,47 @@
...
+          <p translate> FORM.VALIDATION.FUNDING_SOURCE_NAME_ERROR </p>
+        </div>
+      </div>
+
+      <div class="form-group clearfix" ng-class="{ 'has-error' : fundingSourceForm.code.$invalid && fundingSourceForm.$submitted }">
+        <label class="control-label" translate>FORM.LABELS.CODE</label>
+        <input type="text" class="form-control col-md-9" name="code" ng-model="ModalCtrl.data.code" required>
+        <div class="help-block" ng-messages="fundingSourceForm.code.$error" ng-show="fundingSourceForm.$submitted">
+          <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+          <p translate> FORM.VALIDATION.FUNDING_SOURCE_CODE_ERROR </p>
```
jniles: image

Both of these are just grey text.  Could you update these to make fields with errors red?  Here is an example from patient group management:

image

client/src/modules/stock/entry/entry.html
```diff
@@ -73,6 +73,16 @@
               <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
             </div>
           </div>
+
+          <!-- funding source -->
+          <div ng-if="StockCtrl.stockSettings.enable_funding_source && StockCtrl.allowFundingSource">
```
jniles, line 78: Could you make the `clear()` functionality work for this new feature?

ClearNotWorking

As a user, I expect that clicking the clear button should either reset the dropdown to a default value, or clear the value entirely.

jniles: Similarly, when the value is `submit`-ed, it should clear this dropdown to prevent users from accidentally submitting the incorrect funding source.

test/integration/fundingSources.js
```diff
@@ -0,0 +1,104 @@
...
+        helpers.api.listed(res, 1 + FS_IN_TEST_DB);
+      })
+      .catch(helpers.handler);
+  });
+
+  it('POST /funding_sources add another funding source', () => {
+    return agent.post('/funding_sources')
+      .send(fundingSource2)
+      .then((res) => {
+        expect(res).to.have.status(201);
```
jniles, line 56: Could you use `helpers.api.created()` here to check for conformance to the API specification?

The inventory metadata tests are a good template to follow when creating integration tests:
https://github.com/Third-Culture-Software/bhima/blob/master/test/integration/inventory/metadata.js

test/integration/stock/import.js
```diff
@@ -75,9 +95,11 @@ describe.skip('test/integration/stock/import The stock import API', () => {
     return agent.post('/stock/import')
       .type('form')
       .field('depot_uuid', params.depot_uuid)
+      .field('date', formattedDate)
       .attach('file', fs.createReadStream(invalidFile))
       .then(res => {
-        helpers.api.errored(res, 400, 'ERRORS.BAD_DATA_FORMAT');
+        const errorMsg = `[line : 2] The inventory group undefined must be a valid text`;
```
jniles: This is an interesting error message.  Could you make it into something that could be translated into French for our French-speaking users?

test/integration/stock/import.js
```diff
@@ -1,20 +1,39 @@
+/* eslint-disable max-len */
 /* global expect, agent */
 
 const fs = require('fs');
 const helpers = require('../helpers');
 const shared = require('./shared');
 
-describe.skip('test/integration/stock/import The stock import API', () => {
```
jniles, line 7: Thanks for unskipping these!

test/integration/fundingSources.js
```diff
@@ -0,0 +1,104 @@
...
+        helpers.api.listed(res, 2 + FS_IN_TEST_DB);
+      })
+      .catch(helpers.handler);
+  });
+
+  it('POST /funding_sources add a new funding source with an existing code', () => {
+    return agent.post('/funding_sources')
+      .send(fundingSource1)
+      .then((res) => {
+        expect(res).to.have.status(400);
```
jniles, line 73: What is the expected error message?

test/integration/fundingSources.js
```diff
@@ -0,0 +1,104 @@
...
+      })
+      .catch(helpers.handler);
+  });
+
+  it('PUT /funding_sources update a funding source', () => {
+    return agent.put(`/funding_sources/${uuid}`)
+      .send(fundingSourceUpdate)
+      .then((res) => {
+        expect(res).to.have.status(200);
```
jniles, line 82: Does it return the updated object?

jniles (commented):
Thanks @mbayopanda, I have done another review, focused on the functionality this time.

There are only two changes I feel strongly about:  the color of the error messages, and the `clear()` behavior.  Both should be easy fixes.

Here are a few more ideas to improve the functionality:

Would it be possible to add the funding source to the stock entry receipt, if it exists?  I propose that it goes in the area circled in red.

image

I also couldn't remember how to edit lots in BHIMA, but editing funding source might be a good option to add to the lot edit modal.

The PR performs as described and the code looks good to me.  The tests could use some tightening up to make sure they don't silently break in the future.

You can decide at what point to merge this PR - just some suggestions from my end.

mbayopanda:
@jniles could you review my PR please ?

client/src/js/components/bhFundingSourceSelect/bhFundingSourceSelect.js
```diff
@@ -0,0 +1,50 @@
...
+    bindings    : {
+      fundingSourceUuid : '<',
+      onSelectCallback  : '&',
+      required          : '@?',
+      label             : '@?',
+    },
+  });
+
+FundingSourceSelectController.$inject = [
+  '$rootScope',
```
jniles, line 15: It doesn't look like you need to import `$rootScope` anymore.

server/controllers/stock/import.js
```diff
@@ -120,94 +122,75 @@ function checkDataFormat(data = []) {
...
+      throw new BadRequest(msgFormatted, msgFormatted);
     }
 
     if (!isInventoryTypeDefined) {
-      throw new BadRequest(
-        `[line : ${i + 2}] The inventory type ${item.inventory_type} must be a valid text`,
-        `[line : ${i + 2}] The inventory type ${item.inventory_type} must be a valid text`,
-        // 'ERRORS.NOT_A_TEXT',
-      );
+      msg = t('ERRORS.ER_BAD_INVENTORY_TYPE');
```
jniles, line 150: Cool!

jniles (approved):
Did another run through and it looks like my concerns were addressed.  Thanks!

PR #8018, opened 2025-02-20
Updated for cron 4.0.0

jmcameron:
Updated cron and fixed a regtest.


jniles (approved):
Thanks for this work @jmcameron .

The test failed to to package-lock getting out of date.  Fixed now.

PR #8057, opened 2025-03-18
Add logo for projects

mbayopanda:
This PR adds:
- Logo to project
- Adjust logo on the report's header

image

The logo of the project on the report's header and on the UI:
image

### DIFF
--- a/client/src/modules/templates/modals/project.modal.html
+++ b/client/src/modules/templates/modals/project.modal.html
@@ -25,6 +25,7 @@
         (<i translate>FORM.LABELS.PROJECT</i>)
       </label>
       <input type="text" class="form-control" name="name" ng-maxlength="$ctrl.maxLength" ng-model="$ctrl.project.name" required>
+
       <div class="help-block" ng-messages="ProjectForm.name.$error" ng-show="ProjectForm.$submitted">
         <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
       </div>
@@ -38,6 +39,48 @@
       <input type="text" class="form-control" name="abbr"ng-maxlength="$ctrl.length3" ng-model="$ctrl.project.abbr" required >
       <div class="help-block" ng-messages="ProjectForm.abbr.$error" ng-show="ProjectForm.$submitted">
         <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+      </div>
+    </div>
+
+    <!-- project logo -->
+    <div class="form-group">
+      <div class="text-center">
+        <div class="logo" ng-if="!$ctrl.project.logo">
+          <span ng-if="!$ctrl.hasThumbnail">
+            <i class="fa fa-3x fa-file-image-o"></i>
+          </span>
+          <span ng-if="$ctrl.hasThumbnail">
+            <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}" style="width: 100px !important; height: 100px !important;">
+          </span>
+        </div>
+        <div class="logo" ng-if="$ctrl.project.logo">
+          <span ng-if="!$ctrl.hasThumbnail">
+            <img ng-src="{{ $ctrl.project.logo }}" style="width: 100px !important; height: 100px !important;" onError="this.src='assets/empty.png';" >
+          </span>
+          <span ng-if="$ctrl.hasThumbnail">
+            <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}" style="width: 100px !important; height: 100px !important;">
+          </span>
+        </div>
+      </div>
+
+      <div class="text-center">
+        <label data-method="upload-logo" class="btn btn-default">
+          <i class="fa fa-edit"></i> <span translate>PROJECT.UPDATE_LOGO</span>
+          <input
+            type="file"
+            name="logo"
+            accept="image/*"
+            ng-model="$ctrl.file"
+            ngf-max-size="{{ $ctrl.maxLogoFileSize }}"
+            ngf-select="$ctrl.setThumbnail($file)"
+            style="display: none;">
+        </label>
+      </div>
+      <div class="help-block text-center form-warning"
+        ng-messages="ProjectForm.logo.$error"
+        ng-show="ProjectForm.logo.$error.maxSize">
+        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+        <span translate translate-values="{ limit : '{{ProjectCtrl.maxLogoFileSize}}' }">FORM.VALIDATION.FILE_SIZE_LIMIT</span>
       </div>
     </div>
 
--- a/client/src/modules/templates/modals/project.modal.html
+++ b/client/src/modules/templates/modals/project.modal.html
@@ -44,23 +44,20 @@
 
     <!-- project logo -->
     <div class="form-group">
-      <div class="text-center">
-        <div class="logo" ng-if="!$ctrl.project.logo">
-          <span ng-if="!$ctrl.hasThumbnail">
-            <i class="fa fa-3x fa-file-image-o"></i>
-          </span>
-          <span ng-if="$ctrl.hasThumbnail">
-            <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}" style="width: 100px !important; height: 100px !important;">
-          </span>
+      <div style="display: flex; flex-direction: column; justify-content: center; align-items:center; margin-bottom: 10px;">
+        <div ng-if="!$ctrl.project.logo && !$ctrl.hasThumbnail" class="square100" style="display:flex; flex-direction:column; justify-content:center; align-items:center; border: 1px dotted gray; background:#efefef;">
+          <i class="fa fa-3x fa-file-image-o"></i>
         </div>
-        <div class="logo" ng-if="$ctrl.project.logo">
-          <span ng-if="!$ctrl.hasThumbnail">
-            <img ng-src="{{ $ctrl.project.logo }}" style="width: 100px !important; height: 100px !important;" onError="this.src='assets/empty.png';" >
-          </span>
-          <span ng-if="$ctrl.hasThumbnail">
-            <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}" style="width: 100px !important; height: 100px !important;">
-          </span>
+        <div ng-if="!$ctrl.project.logo && $ctrl.hasThumbnail">
+          <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}" class="square100">
         </div>
+        <div ng-if="$ctrl.project.logo && !$ctrl.hasThumbnail">
+          <img ng-src="{{ $ctrl.project.logo }}" class="square100" onError="this.src='assets/empty.png';" >
+        </div>
+        <div ng-if="$ctrl.project.logo && $ctrl.hasThumbnail">
+          <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}" class="square100">
+        </div>
+        <small translate>PROJECT.LOGO_SIZE</small>
       </div>
 
       <div class="text-center">
@@ -75,6 +72,7 @@
             ngf-select="$ctrl.setThumbnail($file)"
             style="display: none;">
         </label>
+        <button type="button" ng-click="$ctrl.removeLogo()" class="btn btn-danger"><i class="fa fa-trash"></i></button>
       </div>
       <div class="help-block text-center form-warning"
         ng-messages="ProjectForm.logo.$error"
--- a/client/src/modules/templates/modals/project.modal.html
+++ b/client/src/modules/templates/modals/project.modal.html
@@ -44,18 +44,18 @@
 
     <!-- project logo -->
     <div class="form-group">
-      <div style="display: flex; flex-direction: column; justify-content: center; align-items:center; margin-bottom: 10px;">
-        <div ng-if="!$ctrl.project.logo && !$ctrl.hasThumbnail" class="square100" style="display:flex; flex-direction:column; justify-content:center; align-items:center; border: 1px dotted gray; background:#efefef;">
-          <i class="fa fa-3x fa-file-image-o"></i>
+      <div class="text-center" style="margin-bottom: 10px;">
+        <div>
+          <img ng-if="!$ctrl.project.logo && !$ctrl.hasThumbnail" src="assets/empty.png" alt="The logo of the project" class="img-thumbnail" height="100" width="100" />
         </div>
         <div ng-if="!$ctrl.project.logo && $ctrl.hasThumbnail">
-          <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}" class="square100">
+          <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}" class="img-thumbnail" height="100" width="100" />
         </div>
         <div ng-if="$ctrl.project.logo && !$ctrl.hasThumbnail">
-          <img ng-src="{{ $ctrl.project.logo }}" class="square100" onError="this.src='assets/empty.png';" >
+          <img ng-src="{{ $ctrl.project.logo }}" class="img-thumbnail" height="100" width="100" />
         </div>
         <div ng-if="$ctrl.project.logo && $ctrl.hasThumbnail">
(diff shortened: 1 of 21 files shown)


mbayopanda:

jniles:
This is related to https://github.com/Third-Culture-Software/bhima/issues/8025.

I am not sure what the use case is for this feature.  Can you explain it to me? 

I commented on the issue when it was created, suggesting a better fix would be to just run multiple versions of BHIMA than a code fix.  After all, we have `enterprise_id` everywhere since we thought we would handle changes like this at an enterprise level, but then dropped it in favor of just running multiple versions of BHIMA.  So... Why can't we just run multiple versions of BHIMA to get different logos?   How does this feature benefit a rural hospital running BHIMA?  What is the use case?

From a brief review of the code, it seems like the project logo is related to the current logged in session, not the underlying record.  This means that a record like a cash payment, for example, could be created at PAX (e.g. CP.PAX.123), but be printed with the logo of HBB, if the user is signed in as HBB.  I'm worried this would be confusing to users, since they would see the logo and assume the record was made at HBB, when the record really made at PAX. 

To demonstrate this, I created two different logos for HBB and PAX and set them up, using current IMCK data.  I logged in under both projects and printed out the record IV.HBB.117278.  Here is what they look like side by side:

SameInvoiceDifferentLogo-2

At Vanga we ran into lots of issues with projects.  If a patient was registered under one project, they wanted to re-register them under the other project and duplicate the patient record.  I'm sure adding a project-specific logo will add to this desire, since they will be convinced that the patient/invoice/cash payment is a different one since the header is entirely different.

The solution to this could be to reference the project_id from the record and use that to power the logo.  However, we don't have project IDs everywhere.  For example, project_ids are not attached to the `account` table nor the `stock` table.  We could put them in there, but at that point ... what is the point of having an `enterprise_id`?  We are just turning `project`s into `enterprise`s.

---

I suppose we could consider the case of the logo confusion to be rare enough in practice that its worth the risk of deploying it and just telling our hospital users "ignore the logo feature for projects" until we have a good use case for it that our users will understand.  But that seems like introducing a maintenance burden and possible data confusion for very little gain.

I personally feel like running two versions of BHIMA is easiest.  I currently run multiple versions for IMCK, HEV, etc on my local hardware, so I know it is possible.   I'm not sure why that isn't possible here.   

If the pain of setting up two versions of BHIMA is too high, then I suggest we work on getting the installation process to be easier.  I've been doing some work in https://github.com/Third-Culture-Software/bhima/pull/7994 to try to improve it, but it is far from perfect.

mbayopanda:
@jniles thank you for your feedback, the reason of this feature is the case of organizations such as IMA where for some project they need their logo on their documents.

We have received this request from the PMNS project where they need their logo on their documents such as stock report, stock sheet report, etc.

jniles:
Can you run a separate instance of bhima for PNMS with their logo as the enterprise logo?

mbayopanda:
> Can you run a separate instance of bhima for PNMS with their logo as the enterprise logo?

Unfortunately we cannot, since the goal is to centralize the stock management of all projects so that the logistic team can easily manage stock a given project while all stock are stored in the same depot.

jniles:
I see.  I'm not sure how to review this PR.  It's for a use case that I am not familiar with, and there isn't a user story/design/detailed description in an issue to help me decide if the PR fits the needs.  I don't have insider information into how IMA is using BHIMA to evaluate it based on my knowledge or experience.  I don't see how changing logos will benefit the hospitals that use BHIMA, and can see the potential for confusion in implementing this feature.

What are you looking for in a review from me?

mbayopanda:
> I see. I'm not sure how to review this PR. It's for a use case that I am not familiar with, and there isn't a user story/design/detailed description in an issue to help me decide if the PR fits the needs. I don't have insider information into how IMA is using BHIMA to evaluate it based on my knowledge or experience. I don't see how changing logos will benefit the hospitals that use BHIMA, and can see the potential for confusion in implementing this feature.
> 
> What are you looking for in a review from me?

I understand your concern since there is not a lot of organisation which are not hospitals that use BHIMA for this feature actually apart from IMA, but the concept of a project in BHIMA for stock management fits very well with organizations (not just hospitals) that have multiple branches, where documents need to have headers specific to these branches.

This PR does not require a logo for each project; if no logo is defined for a project, the company’s logo will be used instead.

It's will be good to know if the code is correct, and it's doesn't break any functional feature or BHIMA design principles.

Thank you

client/src/modules/templates/modals/project.modal.html
```diff
@@ -41,6 +42,46 @@
       </div>
     </div>
 
+    <!-- project logo -->
+    <div class="form-group">
+      <div style="display: flex; flex-direction: column; justify-content: center; align-items:center; margin-bottom: 10px;">
```
jniles: Instead of introducing custom `flex` styles, could you use bootstrap classes for these?  It will make it easier to upgrade in the future.

I think all you need are `row` classes with different `col-xs` and `col-md` classes.  I would imagine you could use the `text-alignment` classes for center the items (provided you used `<span>` and not `<div>`) ([link](https://getbootstrap.com/docs/3.4/css/#type-alignment)) or you might have better luck with `center-block` ([src](https://getbootstrap.com/docs/3.4/css/#images-responsive)).  Also, you could use `image-thumbnail` for the image ([link](https://getbootstrap.com/docs/3.4/css/#images-shapes)).

client/src/modules/templates/modals/project.modal.html
```diff
@@ -41,6 +42,46 @@
...
+        <div ng-if="!$ctrl.project.logo && !$ctrl.hasThumbnail" class="square100" style="display:flex; flex-direction:column; justify-content:center; align-items:center; border: 1px dotted gray; background:#efefef;">
+          <i class="fa fa-3x fa-file-image-o"></i>
+        </div>
+        <div ng-if="!$ctrl.project.logo && $ctrl.hasThumbnail">
+          <img ngf-thumbnail="{{ '$ctrl.thumbnail' }}" class="square100">
+        </div>
+        <div ng-if="$ctrl.project.logo && !$ctrl.hasThumbnail">
+          <img ng-src="{{ $ctrl.project.logo }}" class="square100" onError="this.src='assets/empty.png';" >
```
jniles: Can you move the `this.src` into a function?  Something like `$ctrl.loadDefaultLogo()` would be nice.

It's not clear to me what `this` is set to in the callback.  I assume the DOM element, but it's hard to know when you mix AngularJS and raw JS.

client/src/modules/templates/modals/project.modal.html
```diff
@@ -41,6 +42,46 @@
...
+            ngf-select="$ctrl.setThumbnail($file)"
+            style="display: none;">
+        </label>
+        <button type="button" ng-click="$ctrl.removeLogo()" class="btn btn-danger"><i class="fa fa-trash"></i></button>
+      </div>
+      <div class="help-block text-center form-warning"
+        ng-messages="ProjectForm.logo.$error"
+        ng-show="ProjectForm.logo.$error.maxSize">
+        <div ng-messages-include="modules/templates/messages.tmpl.html"></div>
+        <span translate translate-values="{ limit : '{{ProjectCtrl.maxLogoFileSize}}' }">FORM.VALIDATION.FILE_SIZE_LIMIT</span>
```
jniles, line 81: Nice 👍

client/src/modules/templates/modals/project.modal.js
```diff
@@ -3,22 +3,65 @@ angular.module('bhima.controllers')
...
+    if (!file) {
+      vm.documentError = true;
+      return;
+    }
+    const isImage = file.type.includes('image/');
+    vm.thumbnail = file;
+    vm.hasThumbnail = (vm.thumbnail && isImage);
+  }
+
+  function removeLogo() {
```
jniles, line 40: Nice 👍

server/controllers/admin/projects.js
```diff
@@ -98,16 +97,16 @@ exports.detail = function detail(req, res, next) {
...
-  const data = req.body;
-  const sql = `INSERT INTO project (name, abbr, enterprise_id, zs_id, locked) VALUES (?, ?, ?, ?, ?);`;
-
-  db.exec(sql, [data.name, data.abbr, data.enterprise_id, data.zs_id, data.locked])
-    .then((row) => {
-      res.status(201).send({ id : row.insertId });
-    })
-    .catch(next);
-
+exports.create = async function create(req, res, next) {
```
jniles, line 100: Nice transition to async/await 👍

server/controllers/admin/projects.js
```diff
@@ -117,19 +116,23 @@ exports.create = function create(req, res, next) {
  */
 exports.update = async function update(req, res, next) {
```
jniles, line 117: As a general comment on the design of BHIMA: when the user updates information about a project they are actively using, they must sign out to refresh the session information.  To avoid this, it would be nice to call the `loadSessionInforation()` from `auth` here ([link](https://github.com/Third-Culture-Software/bhima/blob/master/server/controllers/auth.js#L130)) if the project id of the user session is the same as the project id being updated.

You can choose to tackle this in this PR or we can put this off for a later PR.

server/lib/ReportManager.js
```diff
@@ -160,16 +160,33 @@ class ReportManager {
 
     // check if the enterprise has a logo, and normalize the path to absolute
     // Use the IMA application icon as the default logo if nothing else exists
-    const logoPath = metadata.enterprise?.logo || './client/assets/IMAicon.png';
+    const logoPath = metadata.enterprise?.logo || './client/assets/IMAicon.jpg';
```
jniles, line 163: 🐛 🥳 

Good catch!

server/models/migrations/next/migrate.sql
```diff
@@ -51,3 +51,10 @@ UPDATE unit SET `path` = 'TREE.HUMAN_RESOURCES' WHERE `path` = 'TREE.HUMANS_RESS
  * Change the name of the "hiring_date" column to "date_embauche" in the "employee" table
  */
 ALTER TABLE `employee` CHANGE COLUMN date_embauche hiring_date DATETIME DEFAULT NULL;
+
+/*
+ * @author: mbayopanda
+ * @date: 2025-01-05
+ * Add the logo column to the project table
+ */
+CALL add_column_if_missing('project', 'logo', 'VARCHAR(100) DEFAULT NULL');
```
jniles: Let's make this TEXT, since we don't know how long the filename path will be 😄

server/lib/template/partials/title.handlebars
```diff
@@ -1,6 +1,6 @@
 <!-- this partial requires binding the variable 'metadata' -->
 
-<div class="col-xs-7 col-auto">
+<div class="col-auto">
```
jniles: Hmm... we use `col-auto` several times in BHIMA, but I cannot figure out where it is defined.  I don't think we actually have a definition for it.  Maybe we just remove it?

image

test/integration/projects.js
```diff
@@ -104,8 +119,30 @@ describe('test/integration/projects The projects API', () => {
...
+      .field('locked', project.locked)
+      .attach('logo', fs.createReadStream(file), 'file.jpg')
+      .then(innerRes => {
+        expect(innerRes).to.have.status(201);
+        return agent.get(`/projects/${innerRes.body.id}`);
+      })
+      .then(innerRes => {
+        expect(innerRes).to.have.status(200);
+        expect(innerRes).to.be.json;
+        expect(innerRes.body.name).to.equal(project.name.concat(' with logo'));
```
jniles, line 138: Interesting test!  Is there a way to check if the file path is actually returned?

I'm thinking something like:
```js
expect(innerRes.logoPath).to.include('file.jpg'); 
```

Or something similar?

test/integration/projects.js
```diff
@@ -9,25 +9,33 @@ const helpers = require('./helpers');
...
+    abbr : 'TMP',
+    name : 'Temporary Project',
     enterprise_id : 1,
-    zs_id :         759,
-    locked :     0,
+    zs_id : 759,
+    locked : 0,
   };
 
+  const file = './test/fixtures/file.jpg';
```
jniles, line 21: 👍

server/lib/template/partials/enterpriseDetails.handlebars
```diff
@@ -1,34 +1,36 @@
 <!-- this partial requires binding the variable 'metadata' -->
 
-{{#if metadata.enterprise.logoDataURI}}
-  <div class="col-xs-1">
-    <img src="{{metadata.enterprise.logoDataURI}}" width="65" height="65" alt="">
-  </div>
-{{/if}}
+<div class="col-xs-6" style="display: flex; flex-direction: row; justify-content: start;">
```
jniles, line 3: I am okay with flex styles here since we control the chrome browser that renders this template. 👍

jniles (commented):
@mbayopanda, here you go!

Overall, it is a solid pull request.  Good job!  Thanks for adding in some tests!  I think the test should find a way to check for the presence of the `logo` or `logoPath` to make sure that functionality doesn't break in the future.

server/controllers/admin/projects.js
jmcameron: I agree with this concern.   I suggest adding the `loadSessionInformation()` before merging this PR.
This is especially true you plan to switch between projects regularly.

jmcameron (changes requested):
I skimmed the code.  I suggest responding to @jniles concerns before accepting this PR.  Particularly reloading the session info when editing the current project.   If it necessary to change projects regularly, one should not have to log out and log back in to see the changes!

The concept of projects has always been confusing to me.   For instance, the stock does not seem to be associated with any specific project.  So stock reports will be identical regardless of which project is being used.   That seems odd to me.

However, my understanding is that multiple projects would be part of one enterprise.   Therefore replacing the enterprise logo in reports seems wrong.   I suggest that the enterprise logo always shows in the reports (like before) but that a project logo, if available, would shown as well.  So in the reports, the enterprise logo would be on the left, but the project logo could be shown the blank space in the middle of the report header.

client/src/modules/templates/modals/project.modal.html
mbayopanda: I will find a way to remove these flex

server/lib/template/partials/title.handlebars
mbayopanda: Done

test/integration/projects.js
mbayopanda: The filename is changed on the server, I only consider that this attribute exists

jniles: But how do we know that the server succeeded in changing the file name if we don't test for it?

test/integration/projects.js
```diff
@@ -104,8 +119,34 @@ describe('test/integration/projects The projects API', () => {
...
+        return agent.get(`/projects/${innerRes.body.id}`);
+      })
+      .then(innerRes => {
+        expect(innerRes).to.have.status(200);
+        expect(innerRes).to.be.json;
+        expect(innerRes.body.name).to.equal(project.name.concat(' with logo'));
+        expect(innerRes.body.logo).to.exist; // we cannot check the filename since the filename is changed by the server
+        const filepath = 'client/upload/pics/';
+        expect(innerRes.body.logo.startsWith(filepath)).to.be.true;
+        expect(innerRes.body.logo.length).to.equal(filepath.length + 32); // the length of the filename is 32
```
jniles, line 142: This is better!  Thanks.

jniles:
@mbayopanda are you still hoping to get this merged?

If IMA plans to use BHIMA on an ongoing basis, but doesn't plan to use it in hospitals, I would suggest we fork BHIMA back to IMA.  Then you can rapidly modify the code without going through this review process and have to explain to @jmcameron and I the use case as well as get a review.  It seems like it might be easier on you!

mbayopanda:
@jniles there is no support from IMA actually, having a fork at IMA will be poor in contribution, we can just merge it and go on.

jniles:
@mbayopanda, if there is not support from IMA, who will maintain the IMA-specific patches we are merging in?

mbayopanda:
> @mbayopanda, if there is not support from IMA, who will maintain the IMA-specific patches we are merging in?

Actually no one

jniles:
I'm closing this, but willing to reopen if IMA is able to contribute developer resources to maintaining it.

